Issue #2807, opened 2015-04-21
local_advanced_indexing_crossentropy_onehot_grad optimization introduces faulty code


Alexis211:
Hi,

The following Theano code produces a bug that is introduced by the `local_advanced_indexing_crossentropy_onehot_grad` optimization:

```
xy = tensor.matrix("xy", dtype='int32')      ## our ngrams: first k-1 columns = input, last column = output
x = xy[:, :ngram_size-1]
y = xy[:, ngram_size-1]

# Get the embeddings -- input layer
l_in = embed_in[x.flatten()].reshape((y.shape[0], nin))

# First layer : linear, then RELU
j = tensor.dot(l_in, w1) + b1.dimshuffle('x', 0)
nnet_query = tensor.switch(j > 0, j, numpy.float32(0))

# Find best match, then softmax
p_y_given_x = tensor.dot(nnet_query, embed_out.T) + bias_out.dimshuffle('x', 0)
p_softmax = tensor.nnet.softmax(p_y_given_x)

# Calculate cross-entropy & perplexity
cross_entropy = -tensor.mean(tensor.log(p_softmax[tensor.arange(y.shape[0]), y]))

# Calculate the cost to minimize
reg_l1_cost = l1_reg * (abs(w1).sum() + 0.1 * abs(embed_in).sum() + 0.0 * abs(embed_out).sum())
reg_l2_cost = l2_reg * ((w1 ** 2).sum() + 0.1 * (embed_in ** 2).sum() + 0.0 * (embed_out ** 2).sum())
cost = cross_entropy + reg_l1_cost + reg_l2_cost

# Gradient descent learning
updates = [(var, var - learning_rate * tensor.grad(cost, var))
                for var in params]
```

I know for sure that the `local_advanced_indexing_crossentropy_onehot_grad` optimization is in cause because the error does not happen when I disable that single optimization. It also doesn't happen when I run the code on CPU.

Here is the computation graph that causes the error:

```
HostFromGpu [@A] ''   56
 |GpuElemwise{Composite{(((i0 / i1) + (i2 * (i3 + (i4 * i5)))) + (i6 * (i7 + (i4 * i8))))}}[(0, 0)] [@B] ''   55
   |GpuCAReduce{add}{1} [@C] ''   36
   | |GpuCrossentropySoftmaxArgmax1HotWithBias.0 [@D] ''   34
   |   |GpuDot22 [@E] ''   33
[...]

PR #2808, opened 2015-04-22
[REGRESSION, CRASH]Fix crash gpu

nouiz:
fix gh-2807

### DIFF
--- a/theano/sandbox/cuda/nnet.py
+++ b/theano/sandbox/cuda/nnet.py
@@ -247,31 +247,42 @@ class GpuCrossentropySoftmax1HotWithBiasDx(GpuOp):
 
     def c_code_cache_version(self):
         # return ()
-        return (6,)
+        return (7,)
 
     def c_code(self, node, nodename, inp, out, sub):
         dnll, sm, y_idx = inp
         dx, = out
         fail = sub['fail']
         return """
-        if ((CudaNdarray_NDIM(%(dnll)s) != 1)
+        // Get `dnll.shape[0]` or set it to zero if `dnll` is a scalar.
+        const npy_intp %(dnll)s_dims0 = (CudaNdarray_NDIM(%(dnll)s) > 0 ?
+                                         CudaNdarray_HOST_DIMS(%(dnll)s)[0] :
+                                         (npy_intp) 0);
+
+        // Get `dnll.strides[0]` and set it to zero if `dnll` is a scalar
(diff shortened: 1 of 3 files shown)


theano/sandbox/cuda/nnet.py
```diff
@@ -247,31 +247,42 @@ def make_node(self, dy, sm, y_idx):
...
                          "dnll.shape[0] == %%i, but sm.shape[0] == %%i",
-                         CudaNdarray_HOST_DIMS(%(dnll)s)[0],
+                         %(dnll)s_dims0,
                          CudaNdarray_HOST_DIMS(%(sm)s)[0]);
             %(fail)s;
         }
-        if (CudaNdarray_HOST_DIMS(%(dnll)s)[0] !=
-            CudaNdarray_HOST_DIMS(%(y_idx)s)[0])
+        if (%(dnll)s_dims0 !=
+            CudaNdarray_HOST_DIMS(%(y_idx)s)[0] && %(dnll)s_dims0 > 1)
```
lamblin: We should check that `sm.shape[0]` is equal to `y_idx.shape[0]` instead.
Before, we checked that both were equal to `dnll.shape[0]`, but since it can now be broadcasted, we no longer check that the other two are consistent.

theano/sandbox/gpuarray/nnet.py
```diff
@@ -317,24 +317,35 @@ def c_code(self, node, nodename, inp, out, sub):
...
                          "dnll.shape[0] == %%i, but sm.shape[0] == %%i",
-                         PyGpuArray_DIMS(%(dnll)s)[0],
+                         %(dnll)s_dims0,
                          PyGpuArray_DIMS(%(sm)s)[0]);
             %(fail)s;
         }
-        if (PyGpuArray_DIMS(%(dnll)s)[0] !=
-            PyGpuArray_DIMS(%(y_idx)s)[0])
+        if (%(dnll)s_dims0 !=
+            PyGpuArray_DIMS(%(y_idx)s)[0] && %(dnll)s_dims0 > 1)
```
lamblin: Same here.

lamblin:
I'm merging since if fixes an actual crash, but we should still address the points above soon.

theano/sandbox/cuda/nnet.py
nouiz: done in https://github.com/Theano/Theano/pull/2812

PR #2809, opened 2015-04-22
[CRASH] add check to ensure that first input is HostFromGpu in gpuarray/local_assert

carriepl:
The previous implementation would crash if `node.inputs[0].owner` was None (which happened during some scan unit tests).

### DIFF
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -526,8 +526,11 @@ def local_gpua_softmaxwithbias(node):
 @register_opt('fast_compile')
 @op_lifter([theano.tensor.opt.Assert])
 def local_assert(node):
-    return [host_from_gpu(node.op(node.inputs[0].owner.inputs[0],
-                                  *node.inputs[1:]))]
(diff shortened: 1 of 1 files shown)


theano/sandbox/gpuarray/opt.py
```diff
@@ -526,8 +526,9 @@ def local_gpua_softmaxwithbias(node):
 @register_opt('fast_compile')
 @op_lifter([theano.tensor.opt.Assert])
 def local_assert(node):
-    return [host_from_gpu(node.op(node.inputs[0].owner.inputs[0],
-                                  *node.inputs[1:]))]
+    if isinstance(node.inputs[0].owner, HostFromGpu):
```
abergeron: Didn't you want to check that there was an owner too?

carriepl: Yeah, plus I did my check wrong, looking at the apply node instead of the op. Don't know what I was thinking. I implemented the right check and did a rebase to keep the history clean.

abergeron:
This is good and GPU-only code.

PR #2810, opened 2015-04-22
Fix syntax error.

abergeron:
I fucked up when merging #2809 and this introduced a syntax error in master.  This fixes it.


carriepl:
Looks fine. Merging since it fixes syntax error in code that isn't used right now (new gpu backend)

abergeron:
Well the problem is that it's imported so it breaks importing theano (at least on travis).  Which is why it had to be merged fast.

carriepl:
Fully agree. My point was simply that since the code isn't executed, waiting for travis in anyway would have been useless and not at our advantage.

PR #2812, opened 2015-04-22
Mixed: more error detection, clean up, optimization

nouiz:


theano/sandbox/gpuarray/nnet.py
```diff
@@ -351,6 +351,13 @@ def c_code(self, node, nodename, inp, out, sub):
                             "dnll.shape[0] != y_idx.shape[0]");
             %(fail)s;
         }
+        if (PyGpuArray_DIMS(%(sm)s)[0] !=
+            PyGpuArray__DIMS(%(y_idx)s)[0])
```
abergeron: You have two underscores here.

theano/sandbox/gpuarray/tests/test_opt.py
```diff
@@ -7,6 +7,7 @@
 from theano.tensor.tests import test_basic
 
 import theano.sandbox.gpuarray
+from theano.sandbox.gpuarray import basic_ops
```
abergeron: I would rather you add the gpu_contiguous import to the ..basic_ops line below.

If you really want to import basic_op do it with a relative import `from .. import basic_ops`.

nouiz: I did both.

PR #2813, opened 2015-04-22
[CRASH new backend] Fix comparison between ssize_t and size_t

carriepl:
`*start` and `*stop` are `ssize_t` and `len` is `size_t`. When either `ssize_t` is below 0, is will automatically be cast to `size_t` for the comparison and will thus appear to be incredibly large, making the check pass even if it shouldn't.


abergeron:
There obviously was a problem before, and I think this is the right fix.

nouiz:
Was this discovered by crash in current tests? Or do we need new test?

carriepl:
It was discovered by a crash in one of Scan's tests so we don't need a new test case as log as we remember to eventually make scan's tests also run on the new gpu backend.

Issue #2613, opened 2015-03-13
T.concatenate([A, B], axis=-1) doesn't work on GPU


abergeron:
`axis=-1` is not documented to work on CPU but does.  On GPU it gives a confusing error message.

We should clarify whether this is supported or not and regularize the situation on GPU.

julianser:
I've been experimented with T.concatenate (where axis=-1) on CPU and GPU, and it seems to work fine. Arnaud or Xavier, do you have an example code where it doesn't work?

Here is the test that works fine for me:

import numpy as np

import theano
import theano.tensor as T
from theano import shared

rng = np.random.RandomState(22)
t1 = shared(np.asarray(rng.rand(5), theano.config.floatX))
t2 = shared(np.asarray(rng.rand(5), theano.config.floatX))
t = T.concatenate([t1, t2], axis=-1)

test = theano.function(
	inputs=[],
	outputs=t
)

print test()

bouthilx:
@julianser Do you get a 1D tensor with 5 values? This is even worse, there is no error message, but the output is wrong.

Here's what I get:

```python
# Numpy
[ 0.20846054  0.48168105  0.42053804  0.859182    0.17116155  0.33886397
  0.27053285  0.69104135  0.22040452  0.81195092]
# Theano
[ 0.33886397  0.27053285  0.69104135  0.22040452  0.81195092]
```

bouthilx:
@abergeron Was Theano supposed to raise an error or nobody reported such a behavior yet?

nouiz:
I never heard of negative axis with concatenation. So probably nobody
taught of that case before.

If we could support it, that is better, otherwise, we should raise an error.

On Thu, Apr 2, 2015 at 5:11 PM, Xavier Bouthillier <notifications@github.com

bouthilx:
@nouiz Should not be to complicated to support I guess. axis=-1 -> axis=ndims-1?

lamblin:
Indeed, the usual thing to do is `axis < 0` -> `axis += ndim`.

nouiz:
Exact. In fact, if axis <0 do axis+NDIM. That work for all negative axis.
Le 2 avr. 2015 17:38, "Xavier Bouthillier" <notifications@github.com> a
écrit :

julianser:
@bouthilx That's strange! With the current version of Theano I get the exact same results for Theano NumPy:

[...]

Issue #2747, opened 2015-04-09
Crash for T.concatenate([x, x], axis=-1) grad


tscohen:
Hi Theano devs, thanks for writing this great tool.

When I use the fast_run optimizer the following code raises an exception. It works when I use fast_compile, or when I replace the axis=-1 with axis=0 (or whatever the last axis of the argument may be). Since this works under one optimizer and not under another, I think its is a bug and not just a missing feature.

```python
import numpy as np
import theano
import theano.tensor as T

theano.config.exception_verbosity = 'high'
theano.config.optimizer = 'fast_run'
#theano.config.optimizer = 'fast_compile'

X = T.dvector()  # or dmatrix, dtensor3, ...
f = T.concatenate([X, X], axis=-1).sum()  # This fails with fast_run but succeeds with fast_compile
#f = T.concatenate([X, X], axis=0).sum()  # This always works
grad = T.grad(f, wrt=[X])
fg = theano.function([X], grad)
fg(np.random.randn(10))
```

The error I get is:
```
python concat_test.py 
Traceback (most recent call last):
  File "concat_test.py", line 15, in <module>
    fg(np.random.randn(10))
  File "/Users/tsc/anaconda/lib/python2.7/site-packages/theano/compile/function_module.py", line 606, in __call__
    storage_map=self.fn.storage_map)
  File "/Users/tsc/anaconda/lib/python2.7/site-packages/theano/compile/function_module.py", line 595, in __call__
    outputs = self.fn()
  File "/Users/tsc/anaconda/lib/python2.7/site-packages/theano/gof/op.py", line 768, in rval
    r = p(n, [x[0] for x in i], o)
  File "/Users/tsc/anaconda/lib/python2.7/site-packages/theano/tensor/basic.py", line 3205, in perform
    (numpy.sum(splits), len_along_axis))
ValueError: The splits sum to 20, expected 10
Apply node that caused the error: Split{2}(Alloc.0, TensorConstant{-1}, MakeVector.0)
Inputs types: [TensorType(float64, vector), TensorType(int8, scalar), TensorType(int64, vector)]
Inputs shapes: [(10,), (), (2,)]
Inputs strides: [(8,), (), (8,)]
Inputs values: ['not shown', array(-1, dtype=int8), array([10, 10])]

Backtrace when the node is created:
  File "/Users/tsc/anaconda/lib/python2.7/site-packages/theano/gradient.py", line 1089, in access_term_cache
[...]

PR #2814, opened 2015-04-22
Implemented fix for issue #2613: Concatenation bug for negative axes on ...

julianser:
#2613 

@bouthilx, everything seems to work. Let's see if unit tests pass too.

- Julian

### DIFF
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3044,6 +3044,24 @@ class GpuJoin(tensor.Join, GpuOp):
         as_tensor_variable_args = [as_cuda_ndarray_variable(x)
                                    for x in tensors]
 
+        # If axis is negative, we need to increment it with the number
+        # of dimensions of the smallest tensor variable
+        axis_int = int(axis.eval())
+        if (axis_int < 0):
+            # Find tensor with smallest number of dimensions
+            min_dim = -1
+            for cnda in tensors:
+	        if min_dim < 0 or min_dim > len(list(cnda.shape)):
+	            min_dim = len(list(cnda.shape))
+
+	    # Throw error if it's not safe to increment axis with the 
+	    # minimum dim. Normally, this error will be caught in the 
+	    # Join op class, but just to be sure we double check it here. 
+            if axis_int + min_dim < 0:
+                raise ValueError("Cannot join list of tensors at axis (%s) when tensor with smallest dim (%s) is smaller." % (axis_int, min_dim))
+
+	    axis = axis + min_dim
+
         output_maker = \
                 lambda bcast: CudaNdarrayType(broadcastable=bcast)()
 
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3044,21 +3044,25 @@ class GpuJoin(tensor.Join, GpuOp):
         as_tensor_variable_args = [as_cuda_ndarray_variable(x)
                                    for x in tensors]
 
-        # If axis is negative, we need to increment it with the number
-        # of dimensions of the smallest tensor variable
-        axis_int = int(axis.eval())
+        # Get joining axis as int
+        axis_int = 0
+        if not isinstance(axis, int):
+            try:
+                # Note : `get_scalar_constant_value` returns a ndarray not
+                # an int
+                axis_int = int(tensor.get_scalar_constant_value(axis))
+
+            except tensor.basic.NotScalarConstantError:
+                pass
+	else:
+	    axis_int = axis
+
         if (axis_int < 0):
             # Find tensor with smallest number of dimensions
             min_dim = -1
             for cnda in tensors:
-	        if min_dim < 0 or min_dim > len(list(cnda.shape)):
-	            min_dim = len(list(cnda.shape))
-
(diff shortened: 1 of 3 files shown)


julianser:
On another note:

1) I wasn't able to run the file test_opt.py, because I don't have CULA installed (it imports it at the beginning)... Should I install CULA? If so, how?
2) Not sure, but maybe we need to have similar tests for gpuarray?

- Julian

bouthilx:
@julianser Do you mean that an ```ImportError``` is raised during import or that tests dependant on cula fail/are skipped? 

What if you execute this?
```bash 
$ THEANO_FLAGS=device=gpu,floatX=float32,force_device=True nosetests test_opt.py:test_opt_gpujoin_joinvectors_negativeaxes
```

I'll take a look at your PR tomorrow after my exam. Thanks! :)

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3044,6 +3044,24 @@ def make_node(self, *axis_and_tensors):
         as_tensor_variable_args = [as_cuda_ndarray_variable(x)
                                    for x in tensors]
 
+        # If axis is negative, we need to increment it with the number
+        # of dimensions of the smallest tensor variable
+        axis_int = int(axis.eval())
```
bouthilx: As far as I know, all tensor must have the same dimension. See [here](https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L3473) in the CPU version. 

You could simply take first tensor dimension for setting the positive axis value. No need to check that all tensor have same dimension has it's done in ```tensor.Join._make_node_internal()``` just below.

bouthilx: Also, use ```X.ndim``` to get number of dimensions rather than ```len(list(X.shape))```. 

To get ```axis``` value, do something like [this](https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L3440).

bouthilx:
@julianser Few minor things to clear, then it will be ready for merge.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3044,6 +3044,28 @@ def make_node(self, *axis_and_tensors):
...
+                axis_int = int(tensor.get_scalar_constant_value(axis))
+
+            except tensor.basic.NotScalarConstantError:
+                pass
+	else:
+	    axis_int = axis
+
+        if (axis_int < 0):
+            # Find tensor with smallest number of dimensions
+            min_dim = -1
```
bouthilx: @julianser Can you point me where in doc it's said that different number of dimensions is supported or an example that support it? As far as I know, different shapes that are broadcastable are supported but not different number of dimensions.

julianser: Hmm. I don't know if different number of dimensions are supported, but they certaintly should be in the future.

I think as a rule-of-thumb we should always write the most general code, especially if it's in the "make_node" step which is not going to be the bottleneck anyway.

bouthilx: I agree general code could be better when well planed. However, in this case I do not see how we could support different number of dimensions, how would you concatenate this? Numpy, for instance, does not ([see doc](http://docs.scipy.org/doc/numpy/reference/generated/numpy.concatenate.html)) support it and theano's CPU version is based on numpy implementation.

julianser: Ah okay, I didn't realize that NumPy also did not support it. Alright, I've removed it :)

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3044,6 +3044,24 @@ def make_node(self, *axis_and_tensors):
         as_tensor_variable_args = [as_cuda_ndarray_variable(x)
                                    for x in tensors]
 
+        # Get joining axis as int
+        axis_int = 0
```
nouiz: This is not a complete fix. If the axis is a constant, it is the best fix I can think of. It will help merge operation in the graph.

But if the axis is symbolic, it will still crash. So you should add test for that case (small modif to the test T_Join_and_Split.test_join_matrixV in tensor/tests/test_basic.py, it is reused by the GPU in cuda/tests/test_basic_ops.py:T_Join_and_Split.test_join_matrixV). After the test is added, you need to modify the perform and the c_code of this op to handle that correctly.

lamblin:
Just for the record: that commit introduced tabs in the python file, which made python3 unhappy. 
Unfortunately, we do not check everything with Python 3 in Theano, so it went undetected.
#2859 should fix that, but in the future, this is something to keep in mind for an extended buildbot.

julianser:
I'm sorry about the white space issue! I will take more care in the future...

@bouthilx @nouiz I did not expect you to accept this commit directly, because it's still missing the tests for symbolic axis that Fred asked for. Do you want me to implement this test, or should I move on to my next CCW ticket?

bouthilx:
@julianser Tests would not pass, there is additional modifications needed to {Gpu}Join to support variable axis with negative values. I now agree with @nouiz that ticket #2613 and #2747 are tightly related and I believe both should be addressed at the same time to avoid duplicates.

bouthilx:
@julianser You can focus on your next CCW ticket (fuel) for now.

PR #2815, opened 2015-04-22
Merge conv

nouiz:


theano/gof/op.py
```diff
@@ -655,15 +655,22 @@ def do_constant_folding(self, node):
...
 
+        The way we find candidate merge is by checking common the
+        clients of inputs[0] for both node. If it isn't shared by
+        node1 and node2, they will not be merged.
+
         """
-        return True
+        inputs_match = all(node_in is cand_in
+                           for node_in, cand_in in zip(node1.inputs,
+                                                       node2.inputs))
```
lamblin: Check the number of inputs as well.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3354,7 +3354,7 @@ class GpuAlloc(GpuAllocEmpty):
     """
     __props__ = ('memset_0',)
 
-    def do_merge(self, node):
+    def do_merge(self, node1, node2):
         return True
```
lamblin: We should use the new default now, right? Otherwise nodes with different inputs will be merged together

theano/sandbox/cuda/dnn.py
```diff
@@ -368,6 +368,27 @@ def ensure_float(val, default, name):
     return val
 
 
+def do_merge_gpu_alloc_empty(node1, node2):
```
lamblin: Can you explain the logic in words here? I don't understand what we are trying to do.

theano/sandbox/cuda/basic_ops.py
abergeron: Yes this should call Op.do_merge()

theano/sandbox/cuda/dnn.py
abergeron: We are checking if two GpuDnnConv have all the same inputs except for the third one (the output buffer) where it must either be the same or be two GpuAllocEmpty.

abergeron:
This broke scan merging (for some reason).

theano/sandbox/cuda/dnn.py
lamblin: Should it check that the nodes are `GpuDnnConv`, then?

lamblin: Also, this should go in the docstring.

lamblin:
In the failing test, the two scan nodes compute the same thing, but are passed different modes for the inner function, if it helps debug things...

theano/gof/opt.py
```diff
@@ -530,10 +528,10 @@ def process_node(self, fgraph, node):
             if len(node.inputs) != len(candidate.inputs):
                 continue
 
-            inputs_match = all(node_in is cand_in
-                               for node_in, cand_in in zip(node.inputs,
-                                                           candidate.inputs))
-            if inputs_match and node.op == candidate.op:
+            if node.op != candidate.op:
+                continue
```
lamblin: Could it be possible that `not node.op == candidate.op` and `node.op != candidate.op` give different results (for scan in particular)?

lamblin: Indeed, in that particular case, `node.op == candidate.op` _and_ `node.op != candidate.op`.
The reason is that `__eq__` is defined, but `__ne__` is never defined to return `not __eq__`, for any Op.
This was discovered because of the test of scan, but I'm suspecting that all Ops with attributes that should not be used for comparison could be affected.

lamblin: Defining it in `PureOp` seems to work: https://github.com/nouiz/Theano/pull/36

theano/gof/op.py
nouiz: It is already done in the MergeFeature, but I agree, it is safer to do it there too.

theano/sandbox/cuda/dnn.py
nouiz: The check for the op equility is already done in MergeFeature. But I agree, it would be better to do it there too.

nouiz: I do not think we should put the check for op equility there. It is already done by MergeFeature before the first call to do_merge. But we call it again later before doing the swap. I suppose this test can be slow in some scan case. I don't want to slow down the optimization again. But I'll document that.

nouiz:
I updated the PR.

lamblin:
Apparently, my fix triggered other problems in scan, that should probably have happened earlier.
I'll have a look.

lamblin:
Triggering new merge after #2821

lamblin:
The remaining failing test is pep8: `/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/cuda/tests/test_dnn.py:63:5: F841 local variable 'out_shp' is assigned to but never used`

lamblin:
Regarding the idea itself behind the PR, I'm wondering if we should:
- extend it to other GpuAlloc, if they would compare equal, but not be merged (and maybe not have other clients?)
- extend it to other Ops than convolution as well.
We can discuss this later.

nouiz:
After discussion with @lamblin, we have thaught of a better way to do it. We won't disable the merge of GpuAllocEmpty. But we will make the optimization that make the op inplace on them create a new one when needed. This remove much complication in the core of Theano.

nouiz:
A new fix is done in https://github.com/Theano/Theano/pull/2829

Issue #2756, opened 2015-04-10
GPU scalar miss sign


nouiz:
See [[theano-dev] Possible missing sign bug when using GPU](https://groups.google.com/d/topic/theano-dev/LzHtP2OWeRE/discussion)

lamblin:
After a bit of investigation, it appears that there is a bug in CUDA that prevents casting by `(float)` to consistently work.
If I try to use `printf("%i, %f\n", %(z)s, (float) %(z)s)` in the c_code of `scalar.Neg`, it prints `-1, 1.000000`. It does print `-1.000000` if I initialize a new `npy_int8` variable to -1, though.

lamblin:
Converting with `__int2float_rd` instead of `(float)` seems to work, though.

lamblin:
Might be related to https://github.com/ComputationalRadiationPhysics/picongpu/issues/570

lamblin:
It looks like a bug in cuda 6.5 (at least), in any case it works for me with cuda 7.0.

ax3l:
great to see our issue tracker helped you guys.
we could confirm the issue for CUDA 6.5 only and in our case we spotted it with `int(bool)` casts. the official nvidia bug tracker number is 1604407 and it's fixed in nvcc/cuda 7.0 onwards.

https://github.com/ComputationalRadiationPhysics/picongpu/issues/652

PR #2817, opened 2015-04-22
Try to detect bug in nvcc.

lamblin:
Addresses #2756.

### DIFF
--- a/theano/sandbox/cuda/tests/test_driver.py
+++ b/theano/sandbox/cuda/tests/test_driver.py
@@ -69,5 +69,24 @@ def test_nvidia_driver3():
     assert any([isinstance(node.op, cuda.GpuElemwise) for node in topo])
     assert theano.sandbox.cuda.use.device_number is not None
 
+
+def test_nvcc_cast():
+    """Test that the casting behaviour is correct.
+
+    Some versions of nvcc, in particular the one in 6.5.14, return an incorrect
+    value in this case.
+
+    Reported by Zijung Zhang at
+    https://groups.google.com/d/topic/theano-dev/LzHtP2OWeRE/discussion
+    """
+    var = theano.tensor.fvector()
+    f = theano.function([var], -1. * (var > 0), mode=mode_with_gpu)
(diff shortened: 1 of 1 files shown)


lamblin:
I tried it on the same machine:
- the exception is raised with cuda 6.5.14
- it is not raised with cuda 7.0 (7.0.27)

theano/sandbox/cuda/tests/test_driver.py
```diff
@@ -69,5 +69,24 @@ def test_nvidia_driver3():
...
+    Some versions of nvcc, in particular the one in 6.5.14, return an incorrect
+    value in this case.
+
+    Reported by Zijung Zhang at
+    https://groups.google.com/d/topic/theano-dev/LzHtP2OWeRE/discussion
+    """
+    var = theano.tensor.fvector()
+    f = theano.function([var], -1. * (var > 0), mode=mode_with_gpu)
+    if not numpy.allclose(f([-1, 0, 1]), [0, 0, -1]):
+        raise Exception(
```
nouiz: Do we want to always raise? I think it would prevent to run too much frequently as cuda 6.5 is installed at many places and it work correctly most of the time. What about doing this change in GpuElemwise, when the intenal is a Composite, where some of the intermediate variable are not float?

lamblin: Without knowing the full extent of the bug, I'm not sure it would be safe that way, but forbidding cuda 6.5 is indeed a bit extreme.
We could also print a warning in that case, telling users to use DebugMode at least once to check, and only raise an Exception if there is a "cast" in a GpuElemwise.

lamblin: Updated version, since cuda 6.5 is getting kind of old anyway.
I'm just adding a test to the test suite, which will fail for the affected version.

nouiz:
I tested it on cuda 7.5 and the test pass. With cuda 6.5 it fail, so this is good to merge when travis pass.

PR #2820, opened 2015-04-23
Fix how some tests get a reference to scan nodes in a theano function

carriepl:
This would make test work on cpu (and also apparently on the old gpu backend) but not on the new backend because the last node in the toposort would be a HostFromGpu.


carriepl:
Added asserts as requested by @nouiz

PR #2769, opened 2015-04-15
Remove dependance on io_toposort in equal_computations()

carriepl:
Usage of toposort is replaced by parallel depth-first exploration of the two computational graphs. For Kratarth's use case (which admittedly uses huge scans) , it made calls to equal_computations roughly 200x faster.

I launched all the tests in scan_module/tests/, comparing the output between the old version of `equal_computations()` and the new. They agreed in all cases except in test case test_grad_multiple_outs_some_uncomputable on the graph : 

`[RandomFunction{uniform}(<RandomStateType>, TensorConstant{[]}, TensorConstant{-1.1}, TensorConstant{1.1}), Elemwise{mul,no_inplace}(TensorConstant{2}, u2[t]), RandomFunction{uniform}(<RandomStateType>, TensorConstant{[]}, TensorConstant{-1.1}, TensorConstant{1.1}), dot(u1[t], win_copy), DimShuffle{x}(RandomFunction{uniform}.1), Elemwise{add,no_inplace}(<TensorType(float64, (True,))>, dot.0), Elemwise{add,no_inplace}(Elemwise{add,no_inplace}.0, x0[t-1]), Elemwise{add,no_inplace}(Elemwise{add,no_inplace}.0, DimShuffle{x}.0)]`
and
`[RandomFunction{uniform}(<RandomStateType>, TensorConstant{[]}, TensorConstant{-1.1}, TensorConstant{1.1}), RandomFunction{uniform}(<RandomStateType>, TensorConstant{[]}, TensorConstant{-1.1}, TensorConstant{1.1}), Elemwise{mul,no_inplace}(TensorConstant{2}, u2[t]), dot(u1[t], win_copy), DimShuffle{x}(RandomFunction{uniform}.1), Elemwise{add,no_inplace}(<TensorType(float64, (True,))>, dot.0), Elemwise{add,no_inplace}(Elemwise{add,no_inplace}.0, x0[t-1]), Elemwise{add,no_inplace}(Elemwise{add,no_inplace}.0, DimShuffle{x}.0)]`

which the old implementation evaluated as non-equal (apparently, because the toposort returns the nodes in a different order) where as the new implementation says they are equal.

PR #2821, opened 2015-04-23
Fix condition in equal_computations

carriepl:
Adresses the comment raise by @lamblin  in #2769 about tests failing in #2815.


theano/scan_module/scan_utils.py
```diff
@@ -490,8 +490,8 @@ def compare_nodes(nd_x, nd_y):
 
     # Validate that each xs[i], ys[i] pair represents the same computation
     for i in range(len(xs)):
-        if xs[0].owner:
-            # The case where xs and ys don't both have an owner
+        if xs[i].owner:
```
nouiz: could it happend that xs[i] have an owner but not ys[i]?

carriepl: This is validated in the loop that starts at line 406

PR #2824, opened 2015-04-23
[BUG, TEST] Fix bug in numpy 1.9 and test in DebugMode

nouiz:


nouiz:
I updated the title to: [BUG, TEST] Fix bug in numpy 1.9 and test in DebugMode

theano/tensor/basic.py
```diff
@@ -5128,6 +5134,8 @@ class Diagonal(Op):
     """
 
     def __init__(self, offset=0, axis1=0, axis2=1):
+        if numpy_diagonal_return_view:
+            self.view_map = {0: [0]}
```
lamblin: I thought we were not supposed to have views before a certain point of the optimization in general?

nouiz: view do not cause problem. I taught that at some point, (and wrongly changed DimShuffle for this). The problem is to have inplace in the user graph.

theano/tensor/elemwise.py
```diff
@@ -870,6 +870,13 @@ def perform(self, node, inputs, output_storage):
             elif (not isinstance(variable, numpy.ndarray) or
                   variable.dtype != nout.dtype):
                 variable = numpy.asarray(variable, nout.dtype)
+                # The next line is needed for numpy 1.9. Otherwise
+                # there are tests that fail in DebugMode.
+                # Normally we would call theano.misc._asarray, but it
+                # is faster to inline the code. We know that the dtype
+                # are the same string, just different typenum.
+                if numpy.dtype(nout.dtype).num != variable.dtype.num:
+                    variable = variable.view(dtype=nout.dtype)
```
abergeron: I find the justification shaky.

Do you mean that you have something like nout.dtype == 'float32' and variable.dtype.num == NPY_FLOAT32 and they are somehow not equal?

If I misunderstood can you give me an example of a problem that fixes.

nouiz: This happen in a test in DebugMode with numpy 1.9.

This line show it ( I forgot which one in the batch caused this error).

THEANO_FLAGS=optimizer_excluding=local_useless_crossentropy_softmax_1hot_with_bias_dx_alloc:local_useless_inc_subtensor_alloc,on_shape_error=raise,on_opt_error=raise,,warn.argmax_pushdown_bug=False,warn.gpusum_01_011_0111_bug=False,warn.sum_sum_bug=False,warn.sum_div_dimshuffle_bug=False,warn.subtensor_merge_bug=False,,device=cpu,floatX=float64,unittests.rseed=11212,mode=DEBUG_MODE,DebugMode.check_strides=0,DebugMode.patience=3,DebugMode.check_preallocated_output=
theano/../bin/theano-nose
theano/tensor/tests/test_basic.py:{ErfcinvTester,ErfinvTester,TestInferShape,TestTensorInstanceMethods}
theano/tensor/tests/test_extra_ops.py:TestBinCountOp
theano/tensor/tests/test_nlinalg.py:test_qr_modes --pdb --pdb-failure -s

For elemwise of scalar, we receive an numpy scalar, so we call
numpy.asarray() on it with dtype 'int64' but it return a numtype of 9
(longlong) instead of 7 (int64). This show the problem. You can use my
software stack v4 in the lab.

>>> numpy.asarray([], dtype='longlong').dtype == 'int64'
True
>>> numpy.asarray([], dtype='longlong').dtype.num
9
>>> numpy.asarray([], dtype='int64').dtype.num
7

theano/tensor/basic.py
lamblin: OK.

theano/tensor/elemwise.py
abergeron: Ok then, but how are we certain that this won't do bad views (like int64 to uint64)?

Or put another way, how do we know that the dtype is the same string?

nouiz: I prefer not to add another check for the dtype. We already passed on line 872 the dtype we wanted, so it must be good. The extra check would slow down that computation and I already optimized it in the past as it was slow on some tests.

If you insist, I can add it, but I would prefer not to do it.

abergeron: Yes, I see it.  Ok.

PR #2829, opened 2015-04-24
Merge conv2

nouiz:
This PR replace gh-2815 as it have a better way to fix it.


theano/sandbox/cuda/basic_ops.py
```diff
@@ -3299,9 +3299,6 @@ def perform(self, node, inputs, out_):
             # XXX: We could implement and call CudaNdarray.empty(sh) instead.
             out[0] = cuda_ndarray.cuda_ndarray.CudaNdarray.zeros(sh)
 
-    def do_merge(self, node):
-        return False
-
```
abergeron: That's not ok, we need to explicitly disable the merge for GpuAllocEmpty nodes.

abergeron: Ignore this comment, I hadn't realized you used a different approach.

abergeron:
It doesn't pass flake8

nouiz:
fixed

PR #2830, opened 2015-04-24
Fix crash on Mac with curand op. We where using g++ extention that MSVC ...

nouiz:
...also supported

### DIFF
--- a/theano/sandbox/cuda/rng_curand.py
+++ b/theano/sandbox/cuda/rng_curand.py
@@ -167,7 +167,7 @@ class CURAND_Base(GpuOp):
         }
         for (int i = 0; i < %(ndim)s; ++i)
         {
-            odims[i] = ((npy_int32*)(PyArray_DATA(%(size)s) +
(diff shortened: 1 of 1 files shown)


abergeron:
What is the crash exactly?

abergeron:
Never mind I get it.  We where using pointer manipulation on void*, which is forbidden by the standard.

nouiz:
See the report of the error: https://groups.google.com/forum/#!topic/theano-users/yCa4CrPW_5w

abergeron:
But I would use `PyArray_GETPTR1(%(size)s, i)[0]` instead to make this cleaner

lamblin:
Should you update the version as well?

nouiz:
I used the more clear code. I didn't change the version as the code is equivalent and it just fix crash with clang. No need to recompile if it compiled correctly before.

abergeron:
You still need the cast to npy_int32

nouiz:
I amended the last commit. I ran the test locally for that file and they passed.

PR #2831, opened 2015-04-24
[TEST, BUILDBOT] Update tests since new optimization

nouiz:


abergeron:
Just for my curiosity what optimization removes the splits?

nouiz:
in tensor/opt.py local_useless_split

PR #2833, opened 2015-04-24
Change cache refresh patter for speed up

nouiz:
I do not think we need something more complicated for now, as for the CPU tests, with that PR, we now spend only 343s in refresh instead of 1500s.


abergeron:
I'm ok with this change, but I don't think it will make much of a difference for non-buildbot uses.

nouiz:
Agreed. So the frist to see travis pass can merge this PR.

Issue #2825, opened 2015-04-23
Make a meta op of SparseBlockDot and a cpu version


bouthilx:
We decided to break ticket #2501 in two tickets. We need a cpu version of SparseBlockDot to implement the hierarchical softmax for both CPU and GPU. 

```sparse_block_dot``` will stay the public interface and should now return a meta op node rather than the gpu implementation. 

CPU implementation doesn't have to be optimized yet. A simple implementation with python/numpy in ```perform()``` will be sufficient.

Also, SparseBlockDot is hard to understand only with textual description. A drawing should be added in the doc to help comprehension. @adbrebs has a nice sketch that he could use.

bouthilx:
@nouiz @abergeron Do you agree with the proposal?

nouiz:
Yes. The CPU op can be a graph. It don't need to be a new op.
Le 23 avr. 2015 17:48, "Xavier Bouthillier" <notifications@github.com> a
écrit :

abergeron:
This was done in #3311

PR #2835, opened 2015-04-26
Meta op of sparseblockdot and cpu version

adbrebs:
#2825 @bouthilx 
This is only a first draft for the forward graph. I don't know how to deal with the gradient.
In which file should we put the metaop and device optimizer?
I will add more precise documentation at the end.


theano/sandbox/cuda/nnet.py
```diff
@@ -739,3 +739,124 @@ def c_support_code_apply(self, node, nodename):
...
+@local_optimizer([SparseBlockDot])
+def local_sparse_block_dot_device(node):
+    is_running_on_gpu = False
+    for inp in node.inputs:
+        if isinstance(inp, CudaNdarrayType):
+            is_running_on_gpu = True
+            break
+
+    if is_running_on_gpu:
+        return [sparse_block_gemv_ss(*node.inputs)] # Todo: the name of this op should include gpu. What does gemv_ss stand for?
```
bouthilx: It stand for **ge**neral **m**atrix-**v**ector multiplication. ```ss``` stand for **s**parse **s**parse because @abergeron was planning to implement a version where one matrix is sparse and the other is dense. I would agree to add gpu in the name, but gemv and ss should stay there.

theano/sandbox/cuda/nnet.py
```diff
@@ -739,3 +739,124 @@ def c_support_code_apply(self, node, nodename):
...
+
+@local_optimizer([SparseBlockDot])
+def local_sparse_block_dot_device(node):
+    is_running_on_gpu = False
+    for inp in node.inputs:
+        if isinstance(inp, CudaNdarrayType):
+            is_running_on_gpu = True
+            break
+
+    if is_running_on_gpu:
```
bouthilx: What about using any? :)

adbrebs: any does more computations

theano/sandbox/cuda/nnet.py
adbrebs: ok thanks I agree!

theano/sandbox/cuda/nnet.py
```diff
@@ -739,3 +739,124 @@ def c_support_code_apply(self, node, nodename):
...
+    respectively.
+
+    This should not be directly called since the interface is subject
+    to change without notice.  Use the sparse_block_dot_SS() function
+    for a stable interface.
+    """
+    def __init__(self, inplace=False):
+        self.inplace = False
+
+    def make_node(self, o, W, h, inputIdx, outputIdx):
```
bouthilx: where is the bias?

theano/sandbox/cuda/nnet.py
```diff
@@ -739,3 +739,124 @@ def c_support_code_apply(self, node, nodename):
...
+        assert inputIdx.type.dtype in discrete_dtypes
+        assert outputIdx.type.dtype in discrete_dtypes
+
+        return Apply(self, [o, W, h, inputIdx, outputIdx], [o.type()])
+
+    def perform(self, node, inp, out_):
+        raise NotImplementedError('Optimization of SparseBlock failed.')
+
+
+def sparse_block_dot_cpu(o, W, h, inputIdx, outputIdx):
```
bouthilx: where is the bias?

theano/sandbox/cuda/nnet.py
bouthilx: not if using a generator I think

bouthilx: Also CudaNdarrayType would be sufficient for HierarchicalSoftmax but not SparseBlockDot. We should test for HostFromGpu too.

theano/sandbox/cuda/nnet.py
adbrebs: it is included there https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/blocksparse.py#L702

theano/sandbox/cuda/nnet.py
adbrebs: Right!

theano/sandbox/cuda/nnet.py
bouthilx: Right, and it's explained [here](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/blocksparse.py#L25). Then you should remove ```b``` for this method documentation.

theano/sandbox/cuda/nnet.py
bouthilx: OK

theano/sandbox/cuda/nnet.py
abergeron: Actually we can probably drop the `ss` part since it's more generic than I thought and can handle dense matrices.  I was just a bit lazy after I finally got it working.

theano/sandbox/cuda/nnet.py
abergeron: Please don't test for gpu things in other files than stuff under sandbox/cuda.  The design of the meta-opt should ensure that it is possible to "plug in" new implementations of that computation by just writing an optimizer.

bouthilx: @abergeron I understand two optimization functions should be written for 2 different implementations of the meta-op. However, how could you determine which one should be applied? If meta-op is replaced by cpu-version, gpu-version can no longer be used.

theano/sandbox/cuda/nnet.py
bouthilx: @abergeron Can we drop gemv for the MetaOp too?

abergeron: That kind of describes what the op does so it might be better to keep it.

PR #2841, opened 2015-04-28
[CRASH] Improve shared variable management in Scan

carriepl:
~~Those checks were added for the sake of completeness in a previous PR but they were not required for any use case known at that time.~~

~~This PR deactivates them for the time being because it appears these checks are causing issues with valid graphs (https://groups.google.com/forum/#!topic/theano-users/y7Nb8oYRUa0).~~

~~Since they are not required anywhere, they can be deactivated for the time being but we should go back eventually to figure out why.~~

Updated description:
The PR changes the way Scan handles some shared variables as recurrent outputs to make sure the associated and outputs are on the CPU. 

The current way it's implemented would mistakenly have MRG_RandomStreams with an input state on the CPU and the output state on GPU. This wasn't caught by anything but, eventually, through the optimization process, the graph was modified and made consistent so no one ever had any issue with this. The recently-added checks for the consistency of the inner graph caught the inconsistency and that is what caused the crash.

### DIFF
--- a/theano/scan_module/scan.py
+++ b/theano/scan_module/scan.py
@@ -870,7 +870,8 @@ def scan(fn,
                         tensor.unbroadcast(
                             tensor.shape_padleft(input.variable), 0),
                         actual_n_steps))
-                sit_sot_inner_outputs.append(input.update)
+                tensor_update = tensor.as_tensor_variable(input.update)
+                sit_sot_inner_outputs.append(tensor_update)
                 # Not that pos is not a negative index. The sign of pos is used
                 # as a flag to indicate if this output should be part of the
                 # update rules or part of the standard outputs of scan.
--- a/theano/scan_module/tests/test_scan.py
+++ b/theano/scan_module/tests/test_scan.py
@@ -1372,7 +1372,7 @@ class T_Scan(unittest.TestCase):
         if not cuda.cuda_available:
             raise SkipTest('Optional package cuda disabled')
 
-        rs = theano.sandbox.rng_mrg.MRG_RandomStreams()
+        rs = theano.sandbox.rng_mrg.MRG_RandomStreams(use_cuda=True)
         output, _ = theano.scan(lambda : rs.uniform((3,)), n_steps=3)
         cPickle.loads(cPickle.dumps(output))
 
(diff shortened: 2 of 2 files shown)


nouiz:
Can you add a scan test with the case that caused problem? That way, when we try to reintroduce it we will be sure to not break it again.

carriepl:
I updated the PR. It turns out that the checks were not the issue but rather the way Scan handled some shared variables. It allowed them to have input states on the CPU but output states on the GPU. This was fixed during the optimization process so no one ever had any issue related to this before but now that we have checks for the consistency of the inner graph, we could actually detect this invalid state.

udibr:
Just adding more info on problems I observed related to this issue:

I got the  error message:
TypeError: Inconsistency in the inner graph of scan 'scan_fn' : an input and an output are associated with the same recurrent state and should have the 
same type but have type 'TensorType(float32, vector)' and 'CudaNdarrayType(float32, vector)' respectively.

I found that the problem happens with any test value set and testing enabled:

<code>
theano.config.compute_test_value = 'warn'
x.tag.test_value = np.ones((max_steps,batch_size,3)).astype(np.float32)
</code>

or using np.random... to initalize test value for x
but when I remove the testing (comment out the theano.config.compute_test_value = 'warn') then the error disappers

nouiz:
Can you update the description of this issue?

I'm ok with this PR, but if @lamblin or @abergeron can confirm they are ok with this before merging, it would be better.

carriepl:
I updated the description.

nouiz:
Do you think it is useful to add a test that check if the code get optimized correctly on the GPU?

lamblin:
The fix seems to make sense, but I have two questions:
- What if `new_var` is actually not a tensor? Should we convert `input.update` to that type instead?
- I'm afraid this change will prevent the whole thing from moving to the GPU. It would be good to have a test checking that the optimized graph do not have useless transfers.

carriepl:
I added a check that the scan gets moved on the GPU. I think this should be enough and, since we're working on making scan work on the gpuarray backend, this makes the test work on both backends without having to check for transfer nodes from both backends.

PR #2843, opened 2015-04-28
Maxpool c code

yaoli:
all test passed with DebugMode

### DIFF
--- a/theano/tensor/signal/downsample.py
+++ b/theano/tensor/signal/downsample.py
@@ -634,6 +634,9 @@ class DownsampleFactorMaxGrad(Op):
           Py_XDECREF(%(gx)s);
           %(gx)s = (PyArrayObject*) PyArray_ZEROS(4, PyArray_DIMS(%(x)s), x_typenum,0);
         }
+        else {
+          PyArray_FILLWBYTE(%(gx)s, 0);
+        }
         int r_st, r_end, c_st, c_end; // used to index into the input img x
         dtype_%(z)s maximum; // temp var for maximum value in a region
         if (z_r && z_c)
@@ -650,11 +653,6 @@ class DownsampleFactorMaxGrad(Op):
                   r_st -= %(pd0)s;
                   r_end -= %(pd0)s;
 
-                  // handle the case where no padding, ignore border is True
-                  if (%(ignore_border)s)
-                  {
-                    r_end = r_end > r ? r : r_end;
-                  }
                   for(int j=0; j<z_c; j++){
                     c_st = j * %(st1)s;
                     c_end = c_st + %(ds1)s;
@@ -665,11 +663,6 @@ class DownsampleFactorMaxGrad(Op):
                     // change coordinates from padding_img space into img space
                     c_st -= %(pd1)s;
                     c_end -= %(pd1)s;
-                    // handle the case where no padding, ignore border is True
-                    if (%(ignore_border)s)
-                    {
-                      c_end = c_end > c ? c : c_end;
-                    }
                     // the maximum value
                     maximum = ((dtype_%(z)s*)(PyArray_GETPTR4(%(z)s,b,k,i,j)))[0];
                     // the gradient corresponding to this maximum value in z
@@ -697,8 +690,7 @@ class DownsampleFactorMaxGrad(Op):
         """ % locals()
 
     def c_code_cache_version(self):
-        return (0, 3)
-
+        return (0, 6)
 
 class DownsampleFactorMaxGradGrad(Op):
     __props__ = ('ds', 'ignore_border', 'st')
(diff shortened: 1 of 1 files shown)


lamblin:
Your changes make sense, but you will have to rebase on top of the latest master, because there are conflicts, and they are preventing Travis from running.
Also, while you rebase, it would be better to squash those two commits together.

theano/tensor/signal/downsample.py
```diff
@@ -635,44 +634,70 @@ def c_code(self, node, name, inp, out, sub):
...
-            }//for i
-
-            for(int i = x_shp0_usable; i < PyArray_DIMS(%(x)s)[2]; i++){
-                for (int j = 0; j < PyArray_DIMS(%(x)s)[3]; ++j) {
-                    dtype_%(gx)s * gxp = ((dtype_%(gx)s*)(PyArray_GETPTR4(%(gx)s,b,k,i,j)));
-                    gxp[0] = 0;
+                  // handle the case where no padding, ignore border is True
+                  if (%(ignore_border)s)
+                  {
+                    r_end = r_end > r ? r : r_end;
```
nouiz: I don't understand this condition. It don't seam right. If ignore border is True, r_end is already sure to be less then r (see line  648 and 650). What I'm missing? The tests pass... So I probably miss something or tests are missing.

theano/tensor/signal/downsample.py
```diff
@@ -635,44 +634,70 @@ def c_code(self, node, name, inp, out, sub):
...
+                    c_st = c_st < %(pd1)s ? %(pd1)s : c_st;
+                    c_end = c_end > (c - %(pd1)s) ? c - %(pd1)s : c_end;
+                    
+                    // change coordinates from padding_img space into img space
+                    c_st -= %(pd1)s;
+                    c_end -= %(pd1)s;
+                    // handle the case where no padding, ignore border is True
+                    if (%(ignore_border)s)
+                    {
+                      c_end = c_end > c ? c : c_end;
```
nouiz: idem

theano/tensor/signal/downsample.py
```diff
@@ -635,44 +634,70 @@ def c_code(self, node, name, inp, out, sub):
           Py_XDECREF(%(gx)s);
           %(gx)s = (PyArrayObject*) PyArray_ZEROS(4, PyArray_DIMS(%(x)s), x_typenum,0);
         }
```
nouiz: if the output is preallocated region is not initialized to 0., but the code request that.

The old code didn't needed the output to be zeros, as it was putting zeros itself. So it could have allocated not initialized memory. But the new code do not do that, so it should make sure the preallocated output is initialized to 0.

theano/tensor/signal/downsample.py
lamblin: These line may indeed be unnecessary.

nouiz: I checked the code and I also think it isn't used.

For the output preallocated, you will probably get an error when using those Theano flags:

mode=DebugMode,DebugMode.check_preallocated_output=ALL

theano/tensor/signal/downsample.py
yaoli: Do you mean we need to set gx to 0 even when the if in line 626 is not satisfied?

nouiz: I mean, you need to add an else clause to this if. In it, you init to 0.

yaoli: What's the C code I need to write to explicitly set an already existed gx
to all zero?

theano/tensor/signal/downsample.py
yaoli: I've removed the redundant codes and the tests passed. So I think it's OK
to remove them.

What's the C code that sets an existing 'gx' to zero? I've never written
this before.

nouiz: search there: http://docs.scipy.org/doc/numpy/reference/c-api.array.html

yaoli:
all comments addressed

nouiz:
I think we can go simple, if not c contiguous, do not reuse the perallocated output. Currently, we always preallocate with c contiguous memory and I do not see in the near future that changing.

yaoli:
OK. I guess there's an API in numpy to check if the memory is c contiguous.

BTW, why numpy API cares so much of memory being c contiguous or not? Is it
just a matter of looping through the array anyway for most of the APIs?

nouiz:
"PyArray_ISCONTIGUOUS(%s)"

lamblin:
@yaoli In general, it does not care about the memory layout, but for this particular call, since it translates to a `memset`, the memory needs to be contiguous.

yaoli:
OK, so do we go for Fred's suggestion or we simply do double for loop over
it?

nouiz:
no loop, if the preallocated output isn't c contiguous, do not reuse it and
reallocate it.

yaoli:
tests passed

nouiz:
thanks.

Issue #2836, opened 2015-04-27
Bug in the Alloc op grad


rizar:
Here is the code:

```python
import theano
from theano import tensor

x = tensor.vector()
xx = x.dimshuffle('x', 'x', 0)
y = tensor.alloc(xx, 1, 3, x.shape[0])
tensor.grad(y.sum(), x)
```

and here is how it crashes: 

```
Traceback (most recent call last):
  File "alloc.py", line 7, in <module>
    tensor.grad(y.sum(), x)
  File "/home/rizar/Dist/theano/theano/gradient.py", line 545, in grad
    grad_dict, wrt, cost_name)
  File "/home/rizar/Dist/theano/theano/gradient.py", line 1286, in _populate_grad_dict
    rval = [access_grad_cache(elem) for elem in wrt]
  File "/home/rizar/Dist/theano/theano/gradient.py", line 1244, in access_grad_cache
    term = access_term_cache(node)[idx]
  File "/home/rizar/Dist/theano/theano/gradient.py", line 951, in access_term_cache
    output_grads = [access_grad_cache(var) for var in node.outputs]
  File "/home/rizar/Dist/theano/theano/gradient.py", line 1244, in access_grad_cache
    term = access_term_cache(node)[idx]
  File "/home/rizar/Dist/theano/theano/gradient.py", line 1091, in access_term_cache
    input_grads = node.op.grad(inputs, new_output_grads)
  File "/home/rizar/Dist/theano/theano/tensor/basic.py", line 2527, in grad
    gx = gx.dimshuffle(new_order)
  File "/home/rizar/Dist/theano/theano/tensor/var.py", line 340, in dimshuffle
    pattern)
  File "/home/rizar/Dist/theano/theano/tensor/elemwise.py", line 158, in __init__
    (input_broadcastable, new_order))
ValueError: ('You cannot drop a non-broadcastable dimension.', ((True, False), ('x', 'x', 0)))
```

I think I know what the problem is and even tried to fix it, but somehow tests do not pass even in a freshly checked out master making it hard to for me to change anything. I ran `theano-nose theano/tensor/tests/test_basic.py` and got `FAILED (KNOWNFAIL=1, failures=5)` with the following failures:

```
======================================================================
FAIL: test_broadcastable_single_input_broadcastable_dimension (theano.tensor.tests.test_basic.T_Join_and_Split)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #2844, opened 2015-04-29
Fix crash in Alloc grad

nouiz:
fix gh-2836

This is a crash in alloc grad, when the inputs have a broadcastable dimension that isn't broadcasted by the node.

reported by Dmitry Bogdanov

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -2523,7 +2523,7 @@ class Alloc(gof.Op):
             new_order = list(x.broadcastable)
             idx = 0
             for i in range(x.ndim):
-                if not new_order[i]:
+                if i not in axis_broadcasted:
                     new_order[i] = idx
                     idx += 1
                 else:
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -2512,22 +2512,20 @@ class Alloc(gof.Op):
         axis = range(n_axes_to_sum)
         # The broadcasted dimensions
         axis_broadcasted = []
+        axis_kept = []
         for i, (ib, gb) in enumerate(
             zip(inputs[0].broadcastable,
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -2523,7 +2523,7 @@ def grad(self, inputs, grads):
             new_order = list(x.broadcastable)
             idx = 0
             for i in range(x.ndim):
-                if not new_order[i]:
+                if i not in axis_broadcasted:
```
rizar: I am not sure about this solution. What if `n_axes_to_sum` is greater then zero? Then indices in `axis_broadcasted` correspond to `gz` dimensions, not `x` dimensions.

Here is how I fixed it in my fork:

```diff
+        axis_kept = []
         for i, (ib, gb) in enumerate(
             zip(inputs[0].broadcastable,
                 # We need the dimensions corresponding to x
                 grads[0].broadcastable[-inputs[0].ndim:])):
             if ib and not gb:
                 axis_broadcasted.append(i + n_axes_to_sum)
+            else:
+                axis_kept.append(i)
         gx = gz.sum(axis=axis + axis_broadcasted)
         if axis_broadcasted:
-            new_order = list(x.broadcastable)
-            idx = 0
-            for i in range(x.ndim):
-                if not new_order[i]:
-                    new_order[i] = idx
-                    idx += 1
-                else:
-                    new_order[i] = 'x'
+            new_order = ['x'] * x.ndim
+            for idx, axis in enumerate(axis_kept):
+                new_order[axis] = idx
--

rizar:
up!

theano/tensor/basic.py
nouiz: You are right, I forgot a case. This is good, I added an extra tests to this PR for the missing case.

PR #2847, opened 2015-04-29
[MRG] softmax function that builds expression instead of using softmax op

harmdevries89:


nouiz:
There is this one you missed:

```
ERROR: test_grad_two_scans (theano.scan_module.tests.test_scan.T_Scan)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/scan_module/tests/test_scan.py", line 907, in test_grad_two_scans
    outputs_info=[None])
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/scan_module/scan.py", line 745, in scan
    condition, outputs, updates = scan_utils.get_updates_and_outputs(fn(*args))
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/scan_module/tests/test_scan.py", line 903, in forward_scanner
    y_t = tensor.nnet.softmax(a2_t)
AttributeError: 'module' object has no attribute 'softmax'

```

nouiz:
@carriepl what do you think? Should this be softmax op or a graph that represent the softmax for this scan test?

carriepl:
Seems like the test is only there to check if there is a crash when taking the grad through two scans. In that case, I think it doesn't really matter for the test in it's current state if you use the graph or the softmax op so it's probably better to put the graph in case, for instance, someone decides to revisit this test case and add checks when you take the second derivative (which is something the Softmax op doesn't support it memory serves right)

harmdevries89:
@nouiz @lamblin Just to let you know that I committed a working implementation (+test) of the gradient of SoftmaxGrad.

nouiz:
This test failed:

```

FAIL: theano.tensor.nnet.tests.test_nnet.test_argmax_pushdown
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tensor/nnet/tests/test_nnet.py", line 1117, in test_argmax_pushdown
    assert len(fgraph.toposort()) == 2  # an output_guard is second
AssertionError

```

theano/tensor/nnet/nnet.py
```diff
@@ -77,11 +77,16 @@ def perform(self, node, input_storage, output_storage):
...
+        # sm = numpy.zeros_like(x)
+        # for i in xrange(sm.shape[0]):
+            # row = x[i] + b
+            # sm[i] = numpy.exp(row - numpy.max(row))
+            # sm[i] *= 1.0 / numpy.sum(sm[i])
+        # output_storage[0][0] = sm
+
+        x_plus_b = x + b[None, :]
+        e_x = numpy.exp(x_plus_b - x_plus_b.max(axis=1)[:, None])
+        sm = e_x / e_x.sum(axis=1)[:, None]
```
nouiz: This is good change, but can you keep the way of doing with 1.0 / sum(), this was faster as division is much slower the multiplication.

sm = (e_x * (1.0 / e_x.sum(axis=1))[:, None]

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -984,7 +984,7 @@ def test_optimize_xent_vector4(self):
             try:
                 g = theano.function([x, b, y], T.grad(expr, x), mode=mode)
             finally:
-                config.warn.sum_div_dimshuffle_bug = backup
+                config.warn.sum_div_dimshuffle_qbug = backup
```
nouiz: typo introduced.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -999,7 +999,7 @@ def test_optimize_xent_vector4(self):
                 theano.printing.debugprint(g)
                 raise
 
-    def test_scale_cost(self):
+    def test_scrossentropy_softmax_1hot_with_bias_dxcale_cost(self):
```
nouiz: scross -> cross

nouiz:
I did just a few very quick comment. After that we can merge.

Thanks

harmdevries89:
@nouiz Ok, somehow I can not get rid of the assertion error in test_argmax_pushdown, even if I replace the softmax with softmax_op (so that we should be back to the old  code). I'll drop by tomorrow morning to see if you understand what the problem is.

nouiz:
I looked at the code. It seem you missed two conversion from softmax to softmax_op in the optimization local_argmax_pushdown

harmdevries89:
Thanks! I have fixed the issues.

nouiz:
Thanks. We can merge this PR after the python23 PR

PR #2848, opened 2015-04-30
Shape reallocation

royxue:
Rebase of gh-2795

Need to test it with those flags.
    One with flags vm.lazy=False,linker=vm
    One with flags vm.lazy=False,linker=vm,allow_gc=False.

### DIFF
--- a/theano/gof/tests/test_vm.py
+++ b/theano/gof/tests/test_vm.py
@@ -6,14 +6,12 @@ import unittest
 
 from nose.plugins.skip import SkipTest
 import numpy
+
 import theano
-
-
 from theano import function
 from theano.gof import vm
 from theano.gof import OpWiseCLinker
 from theano.compile import Mode
-
 from theano import tensor
 from theano.ifelse import ifelse
 from theano.tensor.var import TensorConstant
@@ -375,6 +373,7 @@ def test_reallocation():
         assert check_storage(storage_map)[0]
         assert len(set([id(v) for v in
                         storage_map.values()])) < len(storage_map)
+
 
 
 def test_shape_reallocation():
--- a/theano/gof/tests/test_vm.py
+++ b/theano/gof/tests/test_vm.py
@@ -375,47 +375,48 @@ def test_reallocation():
                         storage_map.values()])) < len(storage_map)
 
 
-
 def test_shape_reallocation():
         a, b, c, d = [tensor.dvector(n) for n in ['a', 'b', 'c', 'd']]
         z = 3*a + b
         x = c + 4*d
         y = z + x
 
-        m = theano.compile.get_mode(theano.Mode(linker='vm_nogc'))
-        m_1 = m.excluding('fusion', 'inplace')
-        m_2 = m_1.excluding('ShapeOpt')
+        for l in [vm.VM_Linker(allow_gc=False, lazy=False, use_cloop=False),
+                  vm.VM_Linker(allow_gc=True, lazy=False, use_cloop=False)]:
+            m = theano.compile.get_mode(theano.Mode(linker=l))
+            m_1 = m.excluding('fusion', 'inplace')
+            m_2 = m_1.excluding('ShapeOpt')
 
-        m_dict = {m_1: True, m_2: False}
+            m_dict = {m_1: True, m_2: False}
 
-        for m in m_dict.keys():
-            f = theano.function([a, b, c, d], y, name="test_reduce_memory",
-                                mode=m)
+            for m in m_dict.keys():
+                f = theano.function([a, b, c, d], y, name="test_reduce_memory",
+                                    mode=m)
 
-            output = f([1, 2], [3, 5], [9, 8], [4, 3])
-            assert output.any()
-            storage_map = f.fn.storage_map
+                output = f([1, 2], [3, 5], [9, 8], [4, 3])
+                assert output.any()
+                storage_map = f.fn.storage_map
 
-            def check_storage(storage_map):
-                for i in storage_map.keys():
-                    if not isinstance(i, TensorConstant):
-                        keys_copy = storage_map.keys()[:]
-                        keys_copy.remove(i)
-                        for o in keys_copy:
-                            if isinstance(
-                               storage_map[i][0] == storage_map[o][0],
-                                    numpy.ndarray):
-                                compare = (
-                                    storage_map[i][0] == storage_map[o][0]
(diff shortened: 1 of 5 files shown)


theano/gof/tests/test_vm.py
```diff
@@ -5,6 +5,8 @@
 
 from nose.plugins.skip import SkipTest
 import numpy
+import theano
+
```
nouiz: pep8 tell that we should have 3 blocs of import:
1) python related import
2) 3th party module import
3) this project import.

Move this import at line 15?

theano/gof/tests/test_vm.py
```diff
@@ -371,3 +374,48 @@ def check_storage(storage_map):
...
+            assert output.any()
+            storage_map = f.fn.storage_map
+
+            def check_storage(storage_map):
+                for i in storage_map.keys():
+                    if not isinstance(i, TensorConstant):
+                        keys_copy = storage_map.keys()[:]
+                        keys_copy.remove(i)
+                        for o in keys_copy:
+                            if isinstance(
```
nouiz: All this if/else can be simplified to this:

```
compare = all(storage_map[i][0] == storage_map[o][0])

If compare return True, False or ndarray, it will work. 
```

nouiz: My mistake, this isn't True. I think the full inner fo the for loop can be simplified to this:

 if storage_map[i][0] is not None and (storage_map[i][0] == storage_map[o][0]).all():
    return [True, storage_map[o][0]]

If the storage isn't none, in this test, it will always be an ndarray.

theano/gof/tests/test_vm.py
```diff
@@ -371,3 +374,48 @@ def check_storage(storage_map):
...
+        a, b, c, d = [tensor.dvector(n) for n in ['a', 'b', 'c', 'd']]
+        z = 3*a + b
+        x = c + 4*d
+        y = z + x
+
+        m = theano.compile.get_mode(theano.Mode(linker='vm_nogc'))
+        m_1 = m.excluding('fusion', 'inplace')
+        m_2 = m_1.excluding('ShapeOpt')
+
+        m_dict = {m_1: True, m_2: False}
```
nouiz: That is good. I think we should also test with linker "vm". Check the current implementation of test_reallocation. So we will end up with 4 element in the dict, the 2 actual + the same allow_gc=True. Don't use directly linker=vm, we need to explicitly disable the lazy with lazy=False. Otherwise, if people run the tests with the Theano flags vm.lazy=True, the test will fail.

nouiz:
I finished my round of review. I'll start the tests with the special flags here. I have a few very small comments, but they don't prevent me from running the tests now.

theano/gof/tests/test_vm.py
royxue: @nouiz 
(storage_map[i][0] == storage_map[o][0]).all():  I think this only works for when strorage_map[i][0] is a array, but when it is not array. storage_map[i][0] == storage_map[o][0] is True, False, boolean dont have all() method.

nouiz: You are right, that is why I made the check again None before calling this. If it isn't None, in this particular test, they must be ndarray, so it would be fine. But we can't reuse that code if the graph start to contain arbitrary object. This make the test code much clearning, it took me a few minutes to understand what was going on.

nouiz:
I checked the problem for the test_memory.py test failure with the special Theano flags. The failure isn't related to this PR, so when you fix the small changes and document the flag in Theano/doc/library/config.txt we can merge this PR.

I'll update the test_memory.py test to pass in a different PR.

royxue:
@nouiz 
Fred, I have updated this PR

theano/gof/tests/test_vm.py
```diff
@@ -368,6 +370,55 @@ def check_storage(storage_map):
                             return [True, storage_map[o][0]]
             return [False, None]
 
-        assert check_storage(storage_map)[0]
-        assert len(set([id(v) for v in
-                        storage_map.values()])) < len(storage_map)
+        if config.memory_realloc:
```
nouiz: The test should enable it only for the test. Otherwise, it won't test it by default. If you rebase this PR, you can use a decorator like this:
```
@theano.configparser.change_flags(**{'vm.lazy': True})
```

nouiz:
This PR need a rebase.

royxue:
@nouiz 
Sorry for late reply, Fred. I just went back to China from Japan recently. I will do these things asap

nouiz:
tests are failing. Can you look at them?

```


    self.test(*self.arg)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tests/test_flake8.py", line 322, in test_format_flake8
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:109:80: E501 line too long (87 > 79 characters)
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:110:21: W503 line break before binary operator
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:110:80: E501 line too long (86 > 79 characters)
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:111:21: W503 line break before binary operator
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:121:80: E501 line too long (81 > 79 characters)
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:122:37: W503 line break before binary operator
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:138:80: E501 line too long (85 > 79 characters)
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:139:41: W503 line break before binary operator
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:1018:80: E501 line too long (114 > 79 characters)
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/vm.py:1048:43: W503 line break before binary operator

```

nouiz:
can you rebase this branch? Don't forget to make a comment when you add new commits. Otherwise there is delays before I see them. I just found that you updated this PR.

thanks

PR #2851, opened 2015-04-30
Crash fix for tensor.concatenate with negative axis and fast_run

### LINKED ISSUE #2747: Crash for T.concatenate([x, x], axis=-1) grad

orhanf:
fix #2747

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3405,9 +3405,19 @@ class Join(Op):
         """
         axis, tensors = axis_and_tensors[0], axis_and_tensors[1:]
 
+        # Axis can also be a constant
+        if not isinstance(axis, int):
+            try:
+                # Note : `get_scalar_constant_value` returns a ndarray not
+                # an int
+                axis = int(get_scalar_constant_value(axis))
+
+            except NotScalarConstantError:
+                pass
+
         # Fix the axis if it is negative, since it crashes with fast_run
         if axis < 0:
-            axis += axis_and_tensors[1].ndim
+            axis = axis + tensors[0].ndim
 
         if not tensors:
             raise ValueError('Cannot join an empty list of tensors')
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3404,6 +3404,11 @@ class Join(Op):
 
         """
         axis, tensors = axis_and_tensors[0], axis_and_tensors[1:]
+
+        # Fix the axis if it is negative, since it crashes with fast_run
+        if axis < 0:
+            axis += axis_and_tensors[1].ndim
+
         if not tensors:
             raise ValueError('Cannot join an empty list of tensors')
         as_tensor_variable_args = [as_tensor_variable(x) for x in tensors]
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3405,19 +3405,29 @@ class Join(Op):
         """
         axis, tensors = axis_and_tensors[0], axis_and_tensors[1:]
 
-        # Axis can also be a constant
+        # Fix the axis if it is negative, since it crashes with fast_run
+        # Axis can be a constant
         if not isinstance(axis, int):
             try:
+
                 # Note : `get_scalar_constant_value` returns a ndarray not
                 # an int
-                axis = int(get_scalar_constant_value(axis))
+                if int(get_scalar_constant_value(axis)) < 0:
+                    axis += tensors[0].ndim
 
             except NotScalarConstantError:
-                pass
 
-        # Fix the axis if it is negative, since it crashes with fast_run
-        if axis < 0:
-            axis = axis + tensors[0].ndim
+                # Axis can be a tensor variable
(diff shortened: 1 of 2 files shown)


nouiz:
Add as a test the code from the issue in the T_Join_and_Split class in test_basic.py

Also do the fix in perform and c_code in case axis is symbolic. At the same time, check that the value of the axis is a good value. (and add a test to make sure the code raise an error)

theano/tensor/basic.py
```diff
@@ -3404,6 +3404,31 @@ def make_node(self, *axis_and_tensors):
 
         """
         axis, tensors = axis_and_tensors[0], axis_and_tensors[1:]
+
+        # Fix the axis if it is negative, since it crashes with fast_run
```
bouthilx: I believe we should move this to _make_node_internal. This would simplify GpuJoin code at the same time.

orhanf:
@nouiz i've added the tests, also what you say about moving it to `_make_node_internal`

bouthilx:
I made a PR to fix GpuJoin and it fixed Join at the same time (#2871). @nouiz was right, both tickets are tightly related, I now believe we should address both at the same time. 

@nouiz You might not agree with my use of assert_op. Can you take a look at the PR? I would like to know your opinion.

abergeron:
This was fixed with #2871

Issue #399, opened 2012-01-27
prod have no optimization: Reuse sum opt on prod when applicable.


nouiz:

kelvinxu:
The optimizations covered by the ticket seem to be the following:

local_sum_alloc
local_sum_mul_by_scalar
local_sum_div_dimshuffle
local_sum_all_to_none
local_sum_sum

kelvinxu:
The optimizations covered by the ticket seem to be the following:

local_sum_alloc
local_sum_mul_by_scalar
local_sum_div_dimshuffle
local_sum_all_to_none
local_sum_sum

nouiz:
I re-open as the opt  local_sum_div_dimshuffle wasn't converted.

carriepl:
Finished by #3290

PR #2852, opened 2015-04-30
Prod opts [WIP]

kelvinxu:
As a part of #399, @abergeron

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3890,7 +3890,7 @@ def local_sum_prod_mul_by_scalar(node):
                         return [T.mul(scalars[0], node.op(non_scalars[0]))]
                     else:
                         return [scalars[0]]
-        if node_inps.owner and node_inps.owner.op == T.neg:
+        if isinstance(node.op, T.Sum) and node_inps.owner and node_inps.owner.op == T.neg:
             return [T.neg(node.op(node_inps.owner.inputs[0]))]
 
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3865,7 +3865,7 @@ def local_sum_prod_mul_by_scalar(node):
     """
     # TODO: if the the thing inside the Sum is a division,
     # we should get at the numerator....
-    if isinstance(node.op, T.Sum) or isinstance(node.op, T.prod):
+    if isinstance(node.op, T.Sum) or isinstance(node.op, T.elemwise.Prod):
         node_inps, = node.inputs
         if node_inps.owner and node_inps.owner.op == T.mul:
             terms = node_inps.owner.inputs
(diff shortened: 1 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -3888,6 +3888,41 @@ def local_sum_mul_by_scalar(node):
...
 
+@register_specialize
+@gof.local_optimizer([T.elemwise.Prod])
+def local_prod_mul_by_scalar(node):
+    """
+    Convert a prod(scalar * smth) -> scalar * prod(smth)
+
+
+    Really, the logical of this opt is exactly the same as local_sum_mul_by_scalar,
+    and this is just code duplication
```
abergeron: Then why are you copying it?  It would be better to integrate the small changes in the opt above than to copy all the code.

theano/tensor/opt.py
```diff
@@ -4246,6 +4330,43 @@ def local_sum_alloc(node):
...
+                    pass
+            else:
+                try:
+                    val = get_scalar_constant_value(input)
+                    assert val.size == 1
+                    val = val.reshape(1)[0]
+                    to_prod = [shapes[i] for i in xrange(len(shapes))
+                               if i in node.op.axis]
+                    if to_prod:
+                        val *= T.mul(*to_prod)
```
abergeron: This is wrong.  In the case of prod, we need to do the exponent, not the multiplication.

theano/tensor/opt.py
```diff
@@ -4006,6 +4041,55 @@ def local_sum_all_to_none(node):
...
+    Prod(Prod()) -> Prod
+    """
+    if isinstance(node.op, T.elemwise.Prod):
+        prod_inps = node.inputs
+        out_dtype = node.op.dtype
+        # We manipulate the graph so this is done to make sure the opt
+        # doesn't affect other computations.
+        if len(prod_inps.clients) == 1:
+            if (prod_inps.owner and
+                    isinstance(prod_inputs.owner.op, T.elemwise.Prod):
```
abergeron: You need an extra `)` at the end here.

abergeron:
This seems to veer somewhat from the direction we discussed.  Copy-and-pasting the code of the optimizations was only to be done as a last resort if the code can't be made to work with Prod.

This is not the case for most of these optimisations.

kelvinxu:
I tried to reflect your comments, and reuse code. There are probably still some errors however.

abergeron:
This is much better.  There are a couple of nits I pointed out.

Also, you should extend the tests for those opts to check if they work correctly with Prod too (or write new tests).

kelvinxu:
slowly getting there..

kelvinxu:
So it is finally passing the tests. Let me know what changes I should make.

theano/tensor/opt.py
```diff
@@ -3885,8 +3890,8 @@ def local_sum_mul_by_scalar(node):
                         return [T.mul(scalars[0], node.op(non_scalars[0]))]
                     else:
                         return [scalars[0]]
-        if thing_summed.owner and thing_summed.owner.op == T.neg:
-            return [T.neg(node.op(thing_summed.owner.inputs[0]))]
+        if node_inps.owner and node_inps.owner.op == T.neg:
+            return [T.neg(node.op(node_inps.owner.inputs[0]))]
```
abergeron: Are you sure that -prod(smth) == prod(-smth).  I think the sign changes according to the number of elements, so we should skip this replacement for Prod.

abergeron:
I noted the final change, and then it's good.

theano/tensor/opt.py
kelvinxu: thanks for catching this

kelvinxu:
passing the tests now.

PR #2854, opened 2015-05-01
Fix warning reported on the mailing list

nouiz:


carriepl:
Should there be a test for this? At least to reproduce the situation described on the mailing list.

nouiz:
added

nouiz:
Don't merge now.

nouiz:
I think this PR is ready, it passed a subset of tests with the daily buidbot tests.

PR #2855, opened 2015-05-01
Scan memory reuse

carriepl:
This PR fixes two bug in the currently deactivated memory pre-allocation feature in Scan and activates it in both the Python and Cython backends. This means that is doesn't fix bugs that are currently occurring but rather it fixes the memory pre-allocation so that we can use it without introducing bugs.

The first bug it fixes comes from an optimization in `GpuDimshuffle` in the cuda gpu backend where, to avoid creating a new array, it can reuse the one it receives as input and alter its attributes, including the `gpudata` attribute. This caused an issue in the old code for the memory pre-allocation because it relied on the assumption that a given CudaNdarray object would always point to the same memory region and that if an Op needed to return a CudaNdarray pointing to a new memory region, it would create a new one and leave the original untouched.

The second bug comes from the SaveMem optimization and the fact that it tries (if it can) to keep a circular buffer only as small as the number of states required at any given iteration to compute the next state. This works without the memory reuse feature because all the outputs of the inner function are computed and then they are put in the buffers. With the memory reuse, since the outputs are stored in the buffers as they are computed by the inner function, the buffers need to be at least 1 element bigger so that the new states can be stored without overwriting the oldest state that is still used in the inner function.

Now, there are added checks to ensure that the pre-allocated array has been reused *and* that it still points to the same data.


carriepl:
I benchmarked the latest commit on the Deeplearning.net LSTM tutorial on a GTX470. 

Under the Cython backend, time per call for the fprop Scan drops from 0.026s to 0.013s (overhead % drops from 45% to 5%). For the bprop scan, time per call drops from 0.117s to 0.093s (overhead % drops from 25% to 13%). Globally, the time taken to train for one epoch drop from 11.26s to 7.63s.

theano/scan_module/scan_op.py
```diff
@@ -674,8 +674,8 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
                   self.n_mit_sot +
                   self.n_sit_sot +
                   self.n_nit_sot)
-        wrapped_inputs = [Param(x, borrow=True) for x in self.inputs]
-        wrapped_outputs = [Out(x, borrow=False) for x in
+        wrapped_inputs = [Param(x, borrow=False) for x in self.inputs]
```
nouiz: Why change this line?

theano/scan_module/scan_opt.py
```diff
@@ -1228,8 +1228,32 @@ def sanitize(x):
...
+                                           node.op.n_sit_sot - 1)
+                        if (i >= first_mitsot_idx and i <= last_sitsot_idx and
+                            len(node.op.outputs) > 1):
+
+                            pval = select_max(nw_steps - start + init_l[i],
+                                              init_l[i] + 1)
+                        else:
+                            pval = select_max(nw_steps - start + init_l[i],
+                                              init_l[i])
+
```
nouiz: I can't review this part.

theano/scan_module/scan_op.py
carriepl: To ensure there is not aliasing between the inputs and the outputs of the inner function. There are probably be cases where such aliasing would be without consequence but detecting these cases could be tricky. Worse case scenario when using this precaution is that some outputs of the inner function, that would otherwise be aliased, will have to be produced via DeepCopy but since we avoid the copy outside of the inner function, that doesn't really cost us anything.

nouiz:
I think it is good, but there is a part I can't review. Can you document all that at one place? Otherwise, we will all forget the details when we need to play with this next time.

carriepl:
What place do you think would be appropriate for documenting everything related to this feature?

nouiz:
The top of the file scan_op.py?

carriepl:
I added the documentation in the scan_op.py file.

@abergeron and/or @lamblin, can one of you take a look? The part in ScanSaveMem probably deserves even more attention than the others since Fred didn't review it.

nouiz:
What about making a Theano flag scan.prealloc_output that would disable this to not cause extra memory usage?
I discussed yesterday with @carriepl about adding this as an option in the info dict of scan, and make an opt that swap enable it to make DebugMode able to test this. But he wasn't sure if this would end-up as what happened with the gpua key in that dict. @abergeron what do you think?

nouiz:
Also, I asked a 3 persons to test it with big model to see if the memory increase cause problems. If so, it make it more important to allow to disable this.

abergeron:
I don't see why DebugMode should be able to mess with that.  It's a very op-specific setting.

Also, if we add a dict option it's very easy to forget to transfer it at one spot and have bugs that way.

theano/scan_module/scan_perform.pyx
```diff
@@ -366,9 +378,26 @@ def perform(
...
+
+                # The pre-allocated output is only considered as having
+                # been reused if it still points to the same data as it
+                # did before the execution of the inner function
+                if old_output_data[idx] is None:
+                    output_reused[idx] = False
+                else:
+                    if hasattr(new_var, 'gpudata'):
+                        output_reused[idx] = (new_var.gpudata ==
+                                              old_output_data[idx])
```
abergeron: As a bonus, this will work correctly for the new backend too!  Not that I reuse old python objects though.

theano/scan_module/tests/test_scan.py
```diff
@@ -3895,6 +3916,88 @@ def onestep(x, x_tm4):
...
+
+        out1 = out[0].flatten()
+        out2 = out[1].flatten()
+
+        fct = theano.function([input1, init], [out1, out2],
+                              mode=mode_with_gpu)
+
+        floatX = theano.config.floatX
+        output = fct(numpy.ones((2, 1, 1), dtype=floatX),
+                     numpy.ones((1, 1, 1), dtype=floatX))
```
abergeron: If you want to test with GpuDimShuffle, you should force the use of 'float32' so that it always runs on Gpu.

sebastien-j:
I tested this pull request for the largest NMT models I trained on a 3GB GPU (approximately 15k source and target words). The memory usage is fine. For the largest minibatches, it now takes ~2.03 seconds instead of ~2.095s.

theano/scan_module/tests/test_scan.py
carriepl: True. Truth is, I figured leaving it this way would also check the CPU implementation of Dimshuffle, in case anyone ever decided to apply the same optimization as done in the GPU version.

I should probably update the name and description of the test instead.

carriepl:
Modified the test to use float32

carriepl:
Added new flag (scan.allow_output_prealloc) to allow/disallow new optimization

carriepl:
Done

abergeron:
This looks fine but another set of eyes wouldn't be too much.

carriepl:
@lamblin, would you mind having a look?

theano/scan_module/scan_op.py
```diff
@@ -44,6 +86,11 @@
              "Allow/disallow gc inside of Scan (default: False)",
              BoolParam(False))
 
+AddConfigVar('scan.allow_output_prealloc',
+             "Allow/disallow memory preallocation for outputs inside of scan "
+             "(default: False)",
+             BoolParam(True))
```
lamblin: The default value is inconsistent with the doc.

theano/scan_module/scan_op.py
```diff
@@ -1100,9 +1159,19 @@ def execute(self, node, args, outs):
...
+            # the actual outputs of the inner function after its execution
+            for idx in xrange(len(output_storage)):
+
+                var = output_storage[idx].storage[0]
+                old_output_storage[idx] = var
+
+                if hasattr(var, 'gpudata'):
+                    old_output_data[idx] = var.gpudata
+                elif hasattr(var, 'data'):
+                    old_output_data[idx] = var.data
```
lamblin: What does it mean to have `data` or `gpudata`? Are you matching constants? Shared variables? Can you add a comment?

theano/scan_module/scan_opt.py
lamblin: That part makes sense to me.

theano/scan_module/scan_op.py
```diff
@@ -1100,9 +1159,19 @@ def execute(self, node, args, outs):
...
+                else:
+                    old_output_data[idx] = None
```
lamblin: Oh, `var` is actually a value (`ndarray` or `CudaNdarray`), not a Theano variable, and `data` or `gpudata` is the pointer to the memory. Is that right? If so, could you clarify why you are doing that?

carriepl:
Clarified comment for section 4.5 in both Scan backends and fixed the documentation for the flag scan.allow_output_prealloc.

PR #2856, opened 2015-05-01
Lstm fixes

abergeron:
This fixes the memory leak in the DLT LSTM example and allows training to complete in float32 with the new backend.

### DIFF
--- a/theano/sandbox/gpuarray/blas.py
+++ b/theano/sandbox/gpuarray/blas.py
@@ -29,14 +29,15 @@ class BlasOp(HideC):
 
     def c_support_code(self):
         return """
-PyGpuArray *gpublas_try_copy(PyGpuArray *out, PyGpuArray *y)
+PyGpuArrayObject *gpublas_try_copy(PyGpuArrayObject *out,
+                                   PyGpuArrayObject *y) {
   if (out &&
       GpuArray_CHKFLAGS(&out->ga, GA_CARRAY) &&
       theano_size_check(out, PyGpuArray_NDIM(y),
                         PyGpuArray_DIMS(y),
                         y->ga.typecode)) {
     if (pygpu_move(out, y)) {
-      Py_XDECREF(%(out)s)
+      Py_XDECREF(out);
       return NULL;
     }
   } else {
@@ -264,7 +265,7 @@ class GpuDot22(BlasOp, Dot22):
         dims[1] = PyGpuArray_DIMS(%(B)s)[1];
 
         if (theano_prep_output(&%(out)s, 2, dims, %(typecode)s, GA_C_ORDER,
-                              pygpu_default_context()))
+                              pygpu_default_context())) {
             %(fail)s
         }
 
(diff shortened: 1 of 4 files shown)


theano/sandbox/gpuarray/blas.py
```diff
@@ -239,6 +239,7 @@ def c_code(self, node, name, inputs, outputs, sub):
         dims[0] = PyGpuArray_DIMS(%(A)s)[0];
         dims[1] = PyGpuArray_DIMS(%(B)s)[1];
 
+        Py_XDECREF(%(out)s);
```
nouiz: make an issue to tell that this op do not reuse preallocated output. With scan, we start to do that more frequently.

Ideally, we should only allocate if no output or output wrong shape.

theano/sandbox/gpuarray/type.py
```diff
@@ -50,13 +55,20 @@ def filter(self, data, strict=False, allow_downcast=None):
...
+        elif (allow_downcast or
+              (allow_downcast is None and
+               type(data) == float and
+               self.dtype == config.floatX)):
             data = gpuarray.array(data, dtype=self.typecode, copy=False,
                                   ndmin=len(self.broadcastable))
         else:
+            if not hasattr(data, 'dtype'):
+                data = gpuarray.array(data, copy=False)
+
```
nouiz: why this extra call? If useful, add a comment that tell why.

theano/sandbox/gpuarray/blas.py
nouiz: In fact, it is so fast to implement it, can you do it instead of making a ticket?

abergeron:
I've made the change to GpuDot22 and I've also made a similar change to all the other blas ops so that they attempt to reuse the output buffer in some cases.

theano/sandbox/gpuarray/blas.py
```diff
@@ -262,19 +283,18 @@ def c_code(self, node, name, inputs, outputs, sub):
         return code
 
     def c_code_cache_version(self):
-        return (1,)
+        return (2,)
+
+    def c_header_dir(self):
+        ret = super(GpuDot22, self).c_header_dirs()
```
nouiz: This isn't good. It can raise MethodNotImplemented and loose the new information.

abergeron: No, because it inherits from BlasOp which does define c_header_dirs()

abergeron:
I fixed all the problems.

theano/sandbox/gpuarray/blas.py
```diff
@@ -262,19 +284,14 @@ def c_code(self, node, name, inputs, outputs, sub):
         return code
 
     def c_code_cache_version(self):
-        return (1,)
+        return (3,)
 
     def c_headers(self):
         ret = super(GpuDot22, self).c_headers()
-        return ret + ['<numpy_compat.h>']
+        return ret + ['<numpy_compat.h>', '"gpuarray_helper.h"']
```
nouiz: Now this change isn't needed anymore. Also, numpy_compat is now added by blas op, so it isn't useful now, probably old stuff. So you can remove this method completly I think.

abergeron: Fixed.

nouiz:
The tests failing in gpuarray where already there, so I merge.

PR #2857, opened 2015-05-01
Another fix to #2613 which should handle the case where axis is symbolic var

julianser:
@bouthilx @nouiz 

Here is a fix which disables optimization when the axis is symbolic variable (different from constant tensor). Can you double check everything is OK?

Xavier, sorry I didn't see your message until now.

fix gh-2613


bouthilx:
GpuJoin can be used on symbolic axis but perform() and c_code() methods must be modified. The host-to-gpu optimization should not be avoided on symbolic axis.

nouiz:
gh-2814 fixed this.

PR #2859, opened 2015-05-01
Fix use of tabs in python file

lamblin:


lamblin:
Auto-merging because it fixes a crash in pylearn2's buildbot, and it is only whitespace fixing.

PR #2860, opened 2015-05-01
Replace except Class, e -> except Class as e

dwf:
**NOTE:** Ideally #2860, #2862, #2864 and #2869 all get merged around the same time, to minimize the number of rebases. I can also consolidate them into one PR if need be.

Python 3 compatibility, old `except E, e` is unsupported in 3.x, but 'as' is supported >= 2.6, etc.

For the record, all it took was

```
find . -name '*.py' -exec sed -i '.bak' 's/^\( *\)except \([^\(]*\),/\1except \2 as/' {} \;
find . -name '*.py' -exec sed -i '.bak' 's/^\( *\)except \((.*)\),/\1except \2 as/' {} \;
```


abergeron:
This one is simple and ok.

PR #2861, opened 2015-05-01
first commit to Make equal_computations more efficient

### LINKED ISSUE #2799: Make equal_computations more efficient
This is a follow up to #2769. During review, a few points have been suggested to improve the function equal_computations() but they were not implemented since the PR was still an improvement over the previous version of the implementation and the additional points could be used as an introductory ticket for for the CCW and Scan. The improvements that were suggested (and can still be viewed in the original PR) are :
- [ ] Pass `common` as an argument to `compare_nodes()`
- [ ] Also have a cache for pairs of "not common" nodes 
- [ ] Check if the outputs are already in `common`

SinaHonari:
fixes #2799

### DIFF
--- a/theano/scan_module/scan_utils.py
+++ b/theano/scan_module/scan_utils.py
@@ -452,15 +452,13 @@ def equal_computations(xs, ys, in_xs=None, in_ys=None):
         elif len(nd_x.outputs) != len(nd_y.outputs):
             return False
         else:
+            all_in_common=True
             for dx, dy in izip(nd_x.outputs, nd_y.outputs):
                 if (dx, dy) in different:
                     return False
-
-            all_in_common=True
-            for dx, dy in izip(nd_x.outputs, nd_y.outputs):
                 if (dx, dy) not in common:
                     all_in_common = False
-                    break
+
             if all_in_common:
                 return True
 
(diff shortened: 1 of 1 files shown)


theano/scan_module/scan_utils.py
```diff
@@ -451,6 +452,18 @@ def compare_nodes(nd_x, nd_y):
...
+                if (dx, dy) in different:
+                    return False
+
+            all_in_common=True
+            for dx, dy in izip(nd_x.outputs, nd_y.outputs):
+                if (dx, dy) not in common:
+                    all_in_common = False
+                    break
+            if all_in_common:
+                return True
```
lamblin: I think you can avoid doing a second loop.

lamblin:
Otherwise, it looks good.
@carriepl it would be nice if you had a look, since you wrote the original code.

carriepl:
This looks good. However, since in many cases errors in this function (like always returning False, for instance) will only lead to a less efficient graph and would thus not be detectable by the current unit tests, can you run the Scan unit tests and validate that the result of your implementation of equal_computations is identical to the result of the current implementation?

You can do it simply by defining `equal_computations_old()`, containing the old implementation and `equal_computations_new()` containing the new and making `equal_computations()` call both and assert that the output is the same. Don't commit this code, though, simply use it to test your implementation. If this passes, then I am fine with this PR.

SinaHonari:
@carriepl: I compared what the two versions return in the unit tests and they are the same.

lamblin:
Great, thanks a lot!

nouiz:
How much speed up this give to theano.grad with scan?

PR #2862, opened 2015-05-02
Fix print -> print()

dwf:
**NOTE:** Ideally #2860, #2862, #2864 and #2869 all get merged around the same time, to minimize the number of rebases. I can also consolidate them into one PR if need be.

Even on commented code. Just in case.

Also add \__future__ imports.

### DIFF
--- a/bin/theano-cache
+++ b/bin/theano-cache
@@ -13,25 +13,25 @@ _logger = logging.getLogger('theano.bin.theano-cache')
 
 def print_help(exit_status):
     if exit_status:
-        print 'command "%s" not recognized' % (' '.join(sys.argv))
-    print 'Type "theano-cache" to print the cache location'
-    print 'Type "theano-cache help" to print this help'
-    print 'Type "theano-cache clear" to erase the cache'
-    print 'Type "theano-cache list" to print the cache content'
-    print 'Type "theano-cache unlock" to unlock the cache directory'
-    print 'Type "theano-cache cleanup" to delete keys in the old format/code version'
-    print 'Type "theano-cache purge" to force deletion of the cache directory'
-    print ('Type "theano-cache basecompiledir" '
-            'to print the parent of the cache directory')
-    print ('Type "theano-cache basecompiledir list" '
-            'to print the content of the base compile dir')
-    print ('Type "theano-cache basecompiledir purge" '
-            'to remove everything in the base compile dir, '
-            'that is, erase ALL cache directories')
+        print('command "%s" not recognized' % (' '.join(sys.argv)))
+    print('Type "theano-cache" to print the cache location')
+    print('Type "theano-cache help" to print this help')
+    print('Type "theano-cache clear" to erase the cache')
+    print('Type "theano-cache list" to print the cache content')
+    print('Type "theano-cache unlock" to unlock the cache directory')
+    print('Type "theano-cache cleanup" to delete keys in the old format/code version')
+    print('Type "theano-cache purge" to force deletion of the cache directory')
+    print('Type "theano-cache basecompiledir" ')
+           'to print the parent of the cache directory')
+    print('Type "theano-cache basecompiledir list" '
+           'to print the content of the base compile dir')
+    print('Type "theano-cache basecompiledir purge" '
+           'to remove everything in the base compile dir, '
+           'that is, erase ALL cache directories')
     sys.exit(exit_status)
 
 if len(sys.argv) == 1:
-    print config.compiledir
+    print(config.compiledir)
 elif len(sys.argv) == 2:
     if sys.argv[1] == 'help':
         print_help(exit_status=0)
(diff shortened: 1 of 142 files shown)


abergeron:
I don't think using automatic conversion tools is the right approach here.

It leads to silly code like under benchmarks where it tries to execute with python2.5 (via the sheebang) while print_function is only available in 2.6+

Also whatever tool you used somehow missed all the print statements in theano-nose/theano-cache/theano-test, but still added the \__future__ import, which breaks the code.

Finally a giant commit like this forces pretty much everyone to rebase their code after it's merged.

I would rather someone goes through the code file by file and does a careful conversion of all things python3, perhaps aided by tools.

dwf:
I actually added the `__future__` import by myself, and used 2to3 for the rest (since 2to3 doesn't add that import). 2to3 missed it because I was explicitly only running on .py files. I saw the Python 2.5 shebang and I left it, presuming that these benchmarks obviously hadn't been run in a while, and mentally filed it under "not my problem".

However, I'm really sick of waiting around for the right person to come along and do such careful work. It's 2015, nearly 7 years after 2.6 and 3.0 were released, and nobody's done it or been assigned to do it, so I took matters into my own hands. I'm using Python 3 now and the mess of 2to3-aided install means I can't use setup.py develop, which means I have to manually reinstall Theano into every environment I use every time I want to update it.

bartvm:
Just to say that I agree with @dwf. It's very frustrating not to be able to install Theano in development mode. In my opinion, the Herculean task of making a codebase as large as Theano Python 3-compliant manually is easily outweighed by a few people having to rebase their PRs (something that should be extremely straightforward considering the very local changes this PR introduces). I'm also not particularly hopeful that assigning this as CCW tickets for months to come is a better approach.

dwf:
I'll also add that these kinds of changes are pretty much noise in the commit log. While the necessity of rebase after a larger commit is annoying, I'd say there's an argument from a commit history navigation perspective for minimizing the number of nuisance commits in any way we can.

bin/theano-nose
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
bartvm: This one should go after the shebang.

bin/theano-test
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
bartvm: Same here.

dwf: 3958ed5

bin/theano-nose
dwf: 3958ed5

theano/misc/do_nightly_build_send
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
 #!/bin/env python
```
bartvm: Here too

theano/misc/hooks/reindent.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
bartvm: And here

bin/theano-cache
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
bartvm: And I think this is the last one :)

dwf: Oops. bf14992

bin/theano-cache
```diff
@@ -12,25 +13,25 @@ _logger = logging.getLogger('theano.bin.theano-cache')
...
-            'that is, erase ALL cache directories')
+        print('command "%s" not recognized' % (' '.join(sys.argv)))
+    print('Type "theano-cache" to print the cache location')
+    print('Type "theano-cache help" to print this help')
+    print('Type "theano-cache clear" to erase the cache')
+    print('Type "theano-cache list" to print the cache content')
+    print('Type "theano-cache unlock" to unlock the cache directory')
+    print('Type "theano-cache cleanup" to delete keys in the old format/code version')
+    print('Type "theano-cache purge" to force deletion of the cache directory')
+    print('Type "theano-cache basecompiledir" ')
```
bartvm: Parentheses were already used here for string concatenation, and it seems `2to3` got confused by that.

nouiz:
I don't think there is one alternative much better then the others. So if people do PR to work on this, I'll take them. But don't make a 1 big PR with everything. Split it, like 1 PR per feature/syntax fix, or group or files/folders.

To keep the code base working is it worthwhile to update the setup.py file to disable the corresponding conversion in 2to3? That way, new code will have travis fail with the old syntax.

abergeron:
With these 4 PRs is the code python3-ready?

Because if it is, it would significantly alter the balance of disturbance/usefulness.

dwf:
@abergeron Silly `__future__` stuff removed. PTAL.

theano/misc/hooks/reindent.py
abergeron: This is still at the wrong place

theano/misc/tests/test_pycuda_theano_simple.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
abergeron: Should this go after the docstring?  I'm not sure.

lamblin: It should go after the docstring, otherwise the string it is not recognized as a "module-level docstring".

abergeron: Then there area couple more to change.

dwf: I've moved it below any module-level docstring in all the files affected, and rebased again. PTAL.

abergeron:
This looks ok, I'll just wait for travis to pass and I'll merge.

dwf:
@abergeron A test was failing because of whitespace cleanup. Turns out the whitespace was part of a triple-quoted string literal. In 55f6912 I've reformatted so that it won't be mistaken for nuisance whitespace in the future. Though maybe somebody should make debugprint not emit useless trailing whitespace... Might be a good CCW beginner ticket.

PR #2865, opened 2015-05-02
small optimization of log likelihood function computation

warmspringwinds:
I think that it's not necessary to compute the logarithm of the whole matrix.
We can compute the logarithm only of classes from the labels array.


abergeron:
I don't think this makes any change in the final code since we have an optimization that lift subtensors though elemwise operations.

Have you made any timings and noticed a difference?

nouiz:
@abergeron is right, it should not make a difference in the execution speed of the compiled Theano function.

But as this could speed up a little bit the compilation, I'll merge it.

thanks

PR #2866, opened 2015-05-02
Fix for degenerate filters in conv3d2d

xeray:
Fix corresponding to:

https://groups.google.com/forum/#!topic/theano-dev/Q_6FXIpKJmY

I think a whole unit test is not neccessary here as the difference to previous implementation is minor and the effect of the fix can be surveyed directly.

Many thanks to the Theano Team for their great work!
Cheers
Marius


nouiz:
thanks for the PR. I think it sould still be a good idea to have a test that cover that case. It can be very simple, just try it and make sure it run and not crash.

Also, travis failed with this error:

```

ERROR: theano.tensor.nnet.tests.test_conv3d2d.test_conv3d
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tensor/nnet/tests/test_conv3d2d.py", line 120, in test_conv3d
    utt.verify_grad(conv3d, [signals, filters], eps=1e-1)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tests/unittest_tools.py", line 90, in verify_grad
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gradient.py", line 1618, in verify_grad
    o_output = fun(*tensor_pt)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tensor/nnet/conv3d2d.py", line 257, in conv3d
    if filters_shape[1]!=1:
TypeError: 'NoneType' object is unsubscriptable
```

Can you look at this?

xeray:
Hi Frédéric,

thanks for your reply. The trace is very usefull, I know the reason
already. I'm quite busy at the moment, but I'll not forget to create the
tests and push it.

Is it ok to, to have the new and the old version of the function in that
file and put a main section at the end that calls both versions on
random data and verifies that the outputs are the same and there is a
speedup w.r.t to the order?

Cheers
Marius

nouiz:
For the test, I would go to call it with shape that trigger this case and compare to a fixed expected results. Coping this code 2 time seem overkill I think.

nouiz:
Normally, we don't put speed up numbers in the tests. They tend to fail
frequently on user computer as they other stuff that run on the background
that influence the timming.

nouiz:
For the tests, instead of comparing to fixed expected results, you can
reuse the python function in the test file

test_conv3d2d.py to compare the output. So mostly, just add/make sure the
tests contain case that trigger this new code.

xeray:
In case anyone is interested: using filters of size 1 in Tf/z works in recent theano versions already without this fix. BUT for some reasons the gradient on GPU is about 1 magnitude faster using this fix (which skips the use of diagonal_subtensor ). I.e. using the convolution in a CNN, inference is equally fast whereas training is significantly faster!

nouiz:
The code is good, but we don't like merge commit. They make the history harder to follow. Can you rebase on top of master? If you don't have the time or have difficulty, I'll do it in a new PR.

thanks.

xeray:
I tried to rebase it but I have messed up my branch by doing this. If you could do the PR quickly it would be great. Otherwise I might spend quite some time trying to sort it out by myself...

nouiz:
Thanks. I rebased.

PR #2869, opened 2015-05-03
Python 2 and 3 in a single, unmodified codebase (Travis tests pass!)

dwf:
Probably worth rebasing out lots of little nuisance commits.


dwf:
@nouiz @abergeron As of 0be56a8 the only thing left is the dict fixes. Doing that now.

abergeron:
I'll resume review tomorrow once you've dealt with rebases/touch-ups.

dwf:
@abergeron I think that I've covered everything. The only ones that I'd ask you pay special attention to are 698e7e4 and 2864401 (and to a lesser extent, 1558e4a and 0be56a8).

For 698e7e4 and 2864401 I carefully replaced calls to `.items()` with `six.iteritems(...)` where it was appropriate. The only place it really isn't appropriate is if the dict is being modified (i.e keys added/deleted) in the loop iteration, and that happens quite rarely. I was careful to read the loop body but it's possible that I missed some.

dwf:
@abergeron I couldn't sleep so I did the gnarly rebase. I'll be on planes tomorrow but if you have any concerns I'll address them tomorrow night/Thursday.

abergeron:
One of the problems remaining is that you broke flake8 for theano/compat/\__init__.py (mostly by adding imported things that are not declared in \__all__

Also, there a couple of issues pointed out above.

dwf:
Okay, I can fix that shortly.

dwf:
@abergeron did I miss anything?

abergeron:
There are still some flake8 problems with theano/compat/__init__.py.  You have to remove these lines from the python 3 section:

```python
    from functools import partial
    from collections import defaultdict, deque
    from sys import maxsize
    from itertools import combinations, product
```
and fix a couple of blank lines issues.

Finally I tried running the code under python3 with this PR and removing the 2to3 conversion and this reveals a lot of remaining issues.  So this will need more polishing before it goes in.

It should at least pass a full buildbot run with a gpu before going in.  You can run the buildbot with theano/misc/do_nightly_buildbot

dwf:
@abergeron I've addressed some of the issues I've found in 0824f0b, fc58e5b, and 00bb5cf. I'm running the tests now with a GPU on ceylon and I'll fix what I find there, and finally run the buildbot.

Incidentally, to get rid of 2to3 conversion do I just delete the cmd_class and relevant imports in setup.py?

abergeron:
Yes to get rid of 2to3 you just remove the cmd_classes.

Also, I'm pretty sure there is another map problem in compile/debugmode.py that closely resemble the one in function_module

dwf:
@abergeron as is probably obvious, I haven't had time to finish this in light of NIPS, though I did try to fix what errors I could. The remaining Travis failures elude me; if you have any spare cycles to take a look it'd be appreciated.

theano/tensor/basic.py
```diff
@@ -2515,9 +2517,14 @@ def grad(self, inputs, grads):
...
-            for idx, axis in enumerate(axis_kept):
-                new_order[axis] = idx
+            new_order = list(x.broadcastable)
+            idx = 0
+            for i in xrange(x.ndim):
+                if not new_order[i]:
+                    new_order[i] = idx
+                    idx += 1
+                else:
+                    new_order[i] = 'x'
```
lamblin: I'm not sure why the logic was changed here, but it looks like the result is not the same and is the reason one of the tests are failing.
The new code does not look at `axis_kept` for instance.

lamblin:
Also, if you remove `six` from Theano, we should stop relying on `theano.compat.six` in Pylearn2 and add it as an explicit dependency, or it will break.

nouiz:
I won't remove six from Theano now. I think it is better that this is
separate. six was done to be embedded. There is problem as some os have old
version os six and having it embeded assure that we have a good version for
us.

theano/tensor/basic.py
dwf: Very strange, I don't remember changing this.

dwf:
Turns out that we _can't_ run pydotprint tests for Python 3.3 because pydot doesn't work on Python 3. Oh joy.

nouiz:
Here is a pydot version that seem to work with python 3

https://github.com/nlhepler/pydot/commits/master

It see a new maintained version of pydot. This is just a note, we do not need to use it for travis.

dwf:
@lamblin @abergeron @nouiz This passes all tests but one (run by Travis; I have enabled full runs for version 3.3) without any 2to3 conversion. The test that is failing is doing so intermittently: it is the garbage collector test [here](https://travis-ci.org/Theano/Theano/jobs/63437247#L3165). 

I am uncertain whether this is a legitimate test failure or if some behaviour with respect to pickle changed such that we need a higher tolerance for the length differences. 50 bytes difference (< 0.5%) does not seem significant, but I have no idea how to go about determining if this change is actually caused by a newly introduced bug.

nouiz:
David, can you update the threashold to 52? It would pass with that number.
This is due to float not ending up with the same number of caracter in its
string representation...

lamblin:
I also checked the number of tests executed, they are the same in the python2 and python3 versions.
Before merging, we should really launch the whole buildbot, including DebugMode, on a GPU machine, if it has not been done already.

dwf:
@lamblin I agree, if you know how please go for it; I had some trouble (there are some python -c statements that need the print function fixed, but I hit a stumbling block of some sort after that).

Opinions remain divided on the issue of six vs theano.compat.six. I'd rather just suck it in as a dependency but I can also just change it back.

dwf:
That flaky test now passes reliably with a limit of 55 bytes difference.

Can someone give me a hand by running the buildbot script (or instructing me how to do it) under Python 2 and again under Python 3?

P.S. The reason I do not like the idea of bundling six is that it becomes tempting to modify it (already happened once with `get_unbound_function`, which I had to fix in 9769b8e) which makes updating it a huge pain.

lamblin:
The buildbot script is in `theano/misc/do_nightly_build`. You will probably have to either change it a bit to be able to specify which python you want to use, or use separate virtualenvs for python 2 and 3, so there is only one "python" available in each.
If you can't, I'll try to get started on that tomorrow, but I don't have much experience in python3.

nouiz:
After some discussion with @lamblin we think we should keep six now. To
close to NIPS to change that. Also, this would break pylearn2 that expect
it there.

abergeron:
I don't think we should merge this PR before NIPS anyway.  There is way too much potential for breakage.

But after that, assuming it passes the buildbot, I'm all for it.

nouiz:
Agreed. But we should finish it fast after NIPS, to prevent too much rebase
problems...

dwf:
@nouiz @abergeron Now that NIPS is behind us, can we look at getting this merged? I'm running the `do_nightly_build --buildbot` (as best I can) under Python 3 on Boltzmann gpu2. Can someone else try running it on Python 2?

dwf:
(I'm getting really sick of rebasing this)

nouiz:
I agree, we should merge asap. I wanted to push on this, this morning:)

I'll launch now the tests for python2.

kastnerkyle:
I am also excited about this (and thanks for NIPS safety!).

dwf:
@nouiz I've run the buildbot a few times (most recently it died with "bus error" on Boltzmann) and I'm seeing a couple of weird failures in test_basic for Arcsin/Arccos. Also see some failures that are apparently still there on master...

nouiz:
I did a PR to your PR with some fixes

In python2, there is still some problems.

If you run do_nightly_build on theano/sandbox/test_multinomial.py, there is problems with the lock. Do you have time to check that?

How went your tests with python3?

nouiz:
This PR need another rebase. we merged some fix in the master that fixed some tests in the daily buidlbot, so there will be less case false errors when checking if this PR is ready to merge.

abergeron:
You have merge errors since there are `<<<<<< HEAD` tags in the files.

abergeron:
I'm doing a rebase + fixes and I'll post it when done.

dwf:
Weird, I must have done some of this stuff while I was not quite awake.
Thanks @abergeron, I appreciate it.

abergeron:
It's done.

PR #2871, opened 2015-05-04
(#2613, #2747) Add handling of negative axis for Join and GpuJoin

bouthilx:
(#2613, #2747)

Negative axis were automatically handled by numpy implementation inside
Join.{perform,c_code}. However, both Join.grad and
GpuJoin.{perform,c_code,grad} implementations were not thought to take care of
negative axis.

This commit adds convertion of axis when negative such that it is always
positive. Fixing it in Join._make_node_internal also fix GpuJoin at the
same time. Symbolic axis need to be taken care of by adding a
tensor.switch to increment axis when necessary. However, assert_op can
be removed by some optimizer modes and thus exceeding negative axis can
be undetected. Another option is to hardcode asserts in
{Gpu}Join.{perform, c_code} but that means maintaining a similar code
4 times...

### DIFF
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3041,24 +3041,6 @@ class GpuJoin(tensor.Join, GpuOp):
         as_tensor_variable_args = [as_cuda_ndarray_variable(x)
                                    for x in tensors]
 
-        # Get joining axis as int
-        axis_int = 0
-        if not isinstance(axis, int):
-            try:
-                # Note : `get_scalar_constant_value` returns a ndarray not
-                # an int
-                axis_int = int(tensor.get_scalar_constant_value(axis))
-
-            except tensor.basic.NotScalarConstantError:
-                pass
-        else:
-            axis_int = axis
-
-        if (axis_int < 0):
-            # Since all tensors must have the same number of dimensions,
-            # we simply add the number of dimensions for the first tensor
-            axis = axis + as_tensor_variable_args[0].ndim
-
         output_maker = \
                 lambda bcast: CudaNdarrayType(broadcastable=bcast)()
 
@@ -3071,6 +3053,12 @@ class GpuJoin(tensor.Join, GpuOp):
         axis, cndas = axis_and_tensors[0], axis_and_tensors[1:]
         # In case axis is numpy.int8 and has no __index__() method
         axis = int(axis)
+        ndim = tensors[0].ndim
+        if axis < -ndim:
+            raise IndexError("Join axis %d out of bounds [0, %d)" %
+                             (axis, ndim))
+        if axis < 0:
+            axis += ndim
 
         # compute size/shape
         width_sum = 0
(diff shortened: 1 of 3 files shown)


theano/tensor/basic.py
```diff
@@ -3471,6 +3479,12 @@ def _make_node_internal(self, axis, tensors,
                 bcastable = [False] * len(
                     as_tensor_variable_args[0].type.broadcastable)
 
+                axis = theano.tensor.switch(
```
nouiz: Up to now, we always do those check in the op, not in the graph. Even if we duplicate this check 4 times, I think it is better, it is not a natural place to put it in the graph.

theano/tensor/tests/test_basic.py
```diff
@@ -3587,6 +3587,81 @@ def test_join_matrixV(self):
         utt.verify_grad(lambda a, b: join(0, a, b), [v, 2 * v], mode=self.mode)
         utt.verify_grad(lambda a, b: join(1, a, b), [v, 2 * v], mode=self.mode)
 
+    def test_join_matrixV_negative_axis(self):
+        """variable join negative axis"""
+        v = numpy.array([[.1, .2, .3], [.4, .5, .6]], dtype=self.floatX)
+        a = self.shared(v.copy())
+        b = as_tensor_variable(v.copy())
```
nouiz: No need to copy in both place, a copy will be made by default.

bouthilx:
@nouiz I moved the checks in op methods.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3150,9 +3138,24 @@ def c_code(self, node, name, inputs, out_, sub):
 
         """ % locals()
 
+        # Test negative axis
+        str += """
+        if( axis < -nd ){
+            char error_message[50];
+            sprintf(error_message, "Join axis %%d out of bounds [0, %%d)",
+                    axis, nd);
+            PyErr_SetString(PyExc_IndexError, error_message);
```
nouiz: Can you use "PyErr_Format(PyExc_IndexError, "Join axis %%d out of bounds [0, %%d)", axis, nd);"

This is the normal way and is simpler.

theano/tensor/basic.py
```diff
@@ -3511,9 +3522,18 @@ def c_code(self, node, name, inputs, outputs, sub):
             """ % locals()
         code += """
         //PyObject* PyArray_Concatenate(PyObject* obj, int axis)
+        int axis = ((%(adtype)s *)PyArray_DATA(%(axis)s))[0];
+        int ndim = PyArray_NDIM(%(input_1)s);
+        if( axis < -ndim ){
+            char error_message[50];
+            sprintf(error_message, "Join axis %%d out of bounds [0, %%d)",
+                    axis, ndim);
+            PyErr_SetString(PyExc_IndexError, error_message);
```
nouiz: idem.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3041,24 +3041,6 @@ def make_node(self, *axis_and_tensors):
         as_tensor_variable_args = [as_cuda_ndarray_variable(x)
                                    for x in tensors]
 
-        # Get joining axis as int
```
nouiz: Keep that in the make_join_internal. This help canolicalize the ops and help the merge when there is constant.

bouthilx: You want constant axis to be set to 0 in make_node while symbolic axis can get to {perform,c_code} as a negative value?

bouthilx:
@nouiz Done.

PR #2872, opened 2015-05-04
Add new scalar and elemwise op erfcx

TimSalimans:
op for calculating the scaled complementary error function exp(x***2)**erfc(x) in a numerically stable way

This op is useful for calculating various things related to Gaussian distributions, such as truncated moments and expressions involving log(cdf(x)) for large negative values of x. The contributed code very closely follows that of the Erfinv op.

Tested on an EC2 GPU instance. Works for both CPU and GPU.

### DIFF
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -16,6 +16,7 @@ from nose.plugins.attrib import attr
 import numpy
 from numpy.testing import dec, assert_array_equal, assert_allclose
 from numpy.testing.noseclasses import KnownFailureTest
+from distutils.version import LooseVersion
 
 import theano
 from theano.compat import PY3, exc_message, operator_div
@@ -56,6 +57,7 @@ mode_no_scipy = get_default_mode()
 try:
     import scipy.special
     import scipy.stats
+    from scipy import __version__ as scipy_version
     imported_scipy_special = True
 except ImportError:
     if config.mode == "FAST_COMPILE":
@@ -1642,7 +1644,6 @@ ArctanhInplaceTester = makeBroadcastTester(
 if imported_scipy_special:
     expected_erf = scipy.special.erf
     expected_erfc = scipy.special.erfc
-    expected_erfcx = scipy.special.erfcx
     expected_erfinv = scipy.special.erfinv
     expected_erfcinv = scipy.special.erfcinv
     expected_gamma = scipy.special.gamma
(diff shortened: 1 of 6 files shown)


abergeron:
Is it possible that erfcx is very new in scipy?  Because it doesn't seem to be present in 0.11 which is the first version we support.

TimSalimans:
It was introduced in 0.12 according to http://docs.scipy.org/doc/scipy-dev/reference/generated/scipy.special.erfcx.html

theano/tensor/tests/test_basic.py
```diff
@@ -1696,6 +1698,24 @@ def _numpy_true_div(x, y):
...
     skip=skip_scipy)
 
+ErfcxTester = makeBroadcastTester(
+    op=tensor.erfcx,
+    expected=expected_erfcx,
+    good=_good_broadcast_unary_normal_float_no_complex,
+    grad=_grad_broadcast_unary_normal,
+    eps=2e-10,
+    mode=mode_no_scipy,
+    skip=skip_scipy)
```
nouiz: Here you must update skip to check for scipy version >= 0.12, This is probably what make travis fail.

nouiz:
If you want, you could add an opt in tensor/opt.py that would recognize the graph and introduce this stabilized op.

TimSalimans:
Having an automatic optimization would be nice, but difficult to do given the current framework: Whether you want to calculate exp(x***2)**erfc(x) or erfcx(x) depends on the value of x. I currently deal with this by using T.switch() a lot in my own code, but this is probably not a good solution to introduce as a general optimization.

nouiz:
Can describe in a docstring when each version should be used?

TimSalimans:
Travis CI is still failing on python 3.3. Any ideas why?

nouiz:
Could it be that different random number are used? Can you check the range of inputs to have them in a numerical stable region?

TimSalimans:
You were right. The build passed this time.

nouiz:
thanks

carriepl:
It seems like, after the merge of this PR, a few tests (test_basic.py:ErfcinvTester, for instance) started having issues for some people running them locally while being fine according to Travis. I'm guessing this is because some people are not using the same Scipy version as Travis.

nouiz:
What is the error they have? This could be just numerical stability caused by different random number that could be too close to the limit of those functions inputs.

carriepl:
```
GradientError: GradientError: numeric gradient and analytic gradient exceed tolerance:
        At position 5 of argument 0,
            abs. error = 2.057037,  abs. tolerance = 0.010000
            rel. error = 0.025507,  rel. tolerance = 0.010000
Exception args:
The error happened with the following inputs:, [array([[ 1.34068012,  1.47293711,  0.18350478],
       [ 0.53964251,  0.56349099,  1.99717915]], dtype=float32)],
The value of eps is:, None,
The out_type is:, None, Test Elemwise{erfcinv,no_inplace}::normal: Error occurred while computing the gradient on the following inputs: [array([[ 1.34068012,  1.47293711,  0.18350478],
```

Issue #2695, opened 2015-03-27
Accept symbolic `reps` in `tensor.tile`


lamblin:
Now that #2362 has been merged, there is no reason not to accept a symbolic vector as `reps`, or a list of symbolic scalars.
The checks need to be upgraded, and `ndim` will need to be used to access the individual elements of `reps` if it is a vector.
This would address #2434.

lamblin:
Partially addressed by #2734, but we should still add support for a vector of integers.

lamblin:
Also, support `uint` dtypes. The check should be done against `discrete_dtypes`, not an explicit list.

Saizheng:
Hi,
I guess the current version does not support:
1. rep is a integer.
2. rep is a theano scalar.
3. rep is a list of theano scalar, something like [T.scalar(), T.scalar()].
4. rep is a theano vector.

It seems that I need to rewrite the whole function of tile? Or I only need to focus on 3. and 4.?

lamblin:
3. should already be supported, since #2734 was merged.
There is no need to rewrite the whole function, you just have to build a new "reps" that is a list of scalar variables (or scalar constants or ints), and then the main part of the function (lines 4281 and following) does not have to change.
The more important point is 4., but it should not be hard to add 1 and 2 as well.

lamblin:
Btw, points 1 and 2 are mentioned in #2719, which is now assigned to you as well.

dwf:
@Saizheng let me know if you have any questions about this.

dwf:
@Saizheng any update?

Saizheng:
Hey David,

I finished the code for integer and T.scalar cases which are simple, but when reps is a T.vector whose length is not fixe , I have no idea about it yet.

Do you have time this afternoon? We can talk about it when you have time.

lamblin:
@Saizheng If `reps` is a vector, you have to simply make sure that ndim is defined.
Then, you have to pad `reps` with ones to the left, in order for it to have length `ndim`.
And then you simply redefine `reps` as `[reps[i] for i in range(ndim)]`.

Saizheng:
@lamblin you mean reps is a theano symbolic vector? I think in that case, because we don't know the length of a symbolic vector, it is hard to do any following operation like padding..

lamblin:
Fixed by #2875

PR #2875, opened 2015-05-05
TH-2695 Accept symbolic `reps` in `tensor.tile`

Saizheng:
https://github.com/Theano/Theano/issues/2695

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -1,4 +1,5 @@
 """A `Type` and `Op` classes to work with numpy.ndarrays symbolically."""
+import __builtin__
 import sys
 import warnings
 
@@ -4711,17 +4712,19 @@ def tile(x, reps, ndim=None):
 
     See the docstring of `numpy.tile` for details.
 
-    Currently, 'reps' can be constant integer (e.g. 3), constant vector(e.g. [2 3]),
+    'reps' can be constant integer (e.g. 3), constant vector(e.g. [2 3]),
     symbolic scalar (e.g. tensor.iscalar()), symbolic vector (e.g. tensor.ivector())
     or a list of symbolic scalar (e.g. [tensor.iscalar(), tensor.iscalar()]).
-    If reps is constant vector/list/symbolic vector, the length of such vector(list)
-    should be less or equal to x.ndim, and if specified, 'ndim' should be less or equal
-    to x.ndim.
 
-    TODO: expand this so that it could support cases when the length of reps (or 'ndim')
-    is larger than x.ndim, which is supported by numpy.
+    ndim is the number of the dimensions of the output, if it is provided, ndim
+    should be equal or larger than x.ndim and len(reps), otherwise, we will use
+    max(x.ndim, len(reps)) as ndim. If reps is symbolic vector, the ndim has to
+    be provided.
 
     """
+
+    if ndim is not None and ndim < x.ndim:
+        raise ValueError("ndim should be equal or larger than x.ndim")
 
     # if reps is tensor.scalar, integer or tensor.vector, we convert it to a list.
     if not isinstance(reps, (list, tuple)):
@@ -4740,9 +4743,6 @@ def tile(x, reps, ndim=None):
                 raise ValueError("if reps is tensor.vector, you should specify "
                                  "the ndim")
             else:
-                if ndim > x.ndim:
-                    raise ValueError("ndim > x.ndim not currently supported")
-
                 offset = ndim-reps.shape[0]
 
                 # assert that reps.shape[0] does not exceed ndim
@@ -4757,22 +4757,21 @@ def tile(x, reps, ndim=None):
         else:
             raise ValueError("the dimension of reps should not exceed 1")
     else:
-       if len(reps) > x.ndim:
-           raise ValueError("len(reps) > x.ndim not currently supported")
+       if ndim is not None and len(reps) > ndim:
+           raise ValueError("len(reps) should be equal or less than ndim")
        if not numpy.all([isinstance(r, (int, long)) or
            (isinstance(r, TensorVariable) and
             r.dtype in theano.tensor.discrete_dtypes) for r in reps]):
             raise ValueError("elements of reps must be scalars of integer dtype")
-
+       
     # if reps.ndim is less than x.ndim, we pad the reps with
     # "1" so that reps will have the same ndim as x.
-    reps = list(reps) 
-    if len(reps) < x.ndim:
-        reps = [1]*(x.ndim-len(reps)) + reps
-
-    if ndim is None:
-        ndim = len(reps)
     reps = list(reps)
+    if ndim is None:
+        ndim = __builtin__.max(len(reps), x.ndim) 
+    if len(reps) < ndim:
+        reps = [1]*(ndim-len(reps)) + reps
+
     shape = [x.shape[i] for i in xrange(ndim)]
     alloc_shape = reps + shape
     y = alloc(x, *alloc_shape)
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4718,16 +4718,52 @@ def tile(x, reps, ndim=None):
 
     """
 
+    # if reps is tensor.scalar, integer or tensor.vector, we convert it to a list.
+    if not isinstance(reps, (list, tuple)):
+        reps_astensor = as_tensor_variable(reps)
+        ndim_check = reps_astensor.ndim
+
+        # tensor.scalar/integer case
+        if ndim_check == 0:
+            reps_ = []
+            reps_.append(reps)
+            reps = reps_
+        # tesor.vector case
+        elif ndim_check == 1:
+            if ndim == None:
+                raise ValueError("if reps is tensor.vector, you should specify "
+                                 "the ndim")
+            else:
+                offset = ndim-reps.shape[0]
+                # assert that reps.shape[0] does not exceed ndim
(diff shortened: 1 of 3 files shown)


theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+    # if reps is tensor.scalar, integer or tensor.vector, we convert it to a list.
+    if not isinstance(reps, (list, tuple)):
+        reps_astensor = as_tensor_variable(reps)
+        ndim_check = reps_astensor.ndim
+
+        # tensor.scalar/integer case
+        if ndim_check == 0:
+            reps_ = []
+            reps_.append(reps)
+            reps = reps_
```
lamblin: You do not need 3 lines to do that, you can simply do `reps = [reps]`.

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+        # tesor.vector case
```
lamblin: "tensor.vector"

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+        elif ndim_check == 1:
+            if ndim == None:
```
lamblin: The best practice for comparison with `None` is to use `if ndim is None` instead.

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+
+    # if reps.ndim is less than x.ndim, we pad the reps with
+    # "1" so that reps will have the same ndim as x. 
+    if len(reps) < x.ndim:
+       reps = [1]*(x.ndim-len(reps)) + reps
+
     try:
         iter(reps)
     except TypeError:
         raise ValueError("reps must be iterable")
```
lamblin: You can remove that check. At this point, you've made sure that `reps` is either a list or a tuple.

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+        #other raise error
+        else:
+            raise ValueError("the dimension of reps should not exceed 1")
+    else:
+       ndim_check = None
+
+    # if reps.ndim is less than x.ndim, we pad the reps with
+    # "1" so that reps will have the same ndim as x. 
+    if len(reps) < x.ndim:
+       reps = [1]*(x.ndim-len(reps)) + reps
```
lamblin: This may not work if `reps` is a tuple. You can change it to a list in that case.

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+        if ndim_check == 0:
+            reps_ = []
+            reps_.append(reps)
+            reps = reps_
+        # tesor.vector case
+        elif ndim_check == 1:
+            if ndim == None:
+                raise ValueError("if reps is tensor.vector, you should specify "
+                                 "the ndim")
+            else:
```
lamblin: Add a comment explaining what we are trying to do here (dynamically add 1 to the left if `reps` is too short)

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
-            r.dtype in ["int8", "int16", "int32", "int64"]) for r in reps]):
-        raise ValueError("elements of reps must be scalars of integer dtype")
-    elif len(reps) != x.ndim:
+
+    # if reps is tensor.vector, this check should not go through
+    if not ndim_check == 1:
+        if not numpy.all([isinstance(r, (int, long)) or
+            (isinstance(r, TensorVariable) and
+                r.dtype in ["int8", "int16", "int32", "int64"]) for r in reps]):
+            raise ValueError("elements of reps must be scalars of integer dtype")
```
lamblin: You should accept unsigned integer dtypes as well. You could use `theano.tensor.discrete_dtypes` for the list of accepted dtypes.

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
...
+
+    if len(reps) != x.ndim:
         raise ValueError("len(reps) != x.ndim not currently supported")
```
lamblin: It _is_ supported now, that is what you are adding :)
You can make it an assert to catch errors, but there should be no case where those are different after the changes that were made to `reps`.

lamblin:
One of the tests for Tile is failing in Travis, can you check it out?

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
     TODO: expand this.
     """
 
+    # if reps is tensor.scalar, integer or tensor.vector, we convert it to a list.
+    if not isinstance(reps, (list, tuple)):
+        reps_astensor = as_tensor_variable(reps)
+        ndim_check = reps_astensor.ndim
```
lamblin: Check the dtype of `reps_astensor` here as well.

theano/tensor/tests/test_basic.py
```diff
@@ -4936,6 +4936,83 @@ def run_tile(x, x_, reps, use_symbolic_reps):
...
+    reps_ = 2
+    f = function([x, reps], tile(x, reps))
+    assert numpy.all( f(x_, reps_) == numpy.tile(x_, reps_))
+    #tensor.vector:
+    reps = ivector()
+    reps_ = [2, 3]
+    ndim_ = 4
+    f = function([x, reps], tile(x, reps, ndim_))
+    assert numpy.all( f(x_, reps_) == numpy.tile(x_, reps_))
+
```
lamblin: I don't have an exact solution for that, but can you think of a way of refactoring all those tests to avoid that much copy-pasting?

lamblin:
Also, new tests are needed to check that incorrect inputs fail as expected. For instance:
- if the length of `reps` is too long for `ndim`
- if the dtype of `reps` is not correct
- if `ndim` is not provided but the length of `reps` is not known.

You can use `assert_raises` for these tests.

theano/tensor/basic.py
```diff
@@ -4268,15 +4268,51 @@ def tile(x, reps, ndim=None):
     TODO: expand this.
```
lamblin: Update the docstring

lamblin:
We should have another look at the tests together as well.

lamblin:
Thanks for adding the tests.
You still need to:
- rebase another time on the latest master, since you _merged_ two branches into this one (which you should avoid)
- fix the flake8 issues reported by Travis

theano/tensor/tests/test_opt.py
```diff
@@ -3855,15 +3855,15 @@ def test_local_useless_tile(self):
         # But it isn't supported for now, so assert that we raise an
         # error.
```
lamblin: You can remove the comment saying it is not supported.

theano/tensor/tests/test_opt.py
```diff
@@ -3855,15 +3855,15 @@ def test_local_useless_tile(self):
...
 
-        self.assertRaises(ValueError, T.tile, v, (1,)*(v.ndim+1))
+        #self.assertRaises(ValueError, T.tile, v, (1,)*(v.ndim+1))
```
lamblin: Remove that line as well.

theano/tensor/tests/test_opt.py
```diff
@@ -3855,15 +3855,15 @@ def test_local_useless_tile(self):
...
         # If the repeat parameter is shorter then m.ndim, it should
         # pad tot he left the repeat patter with 1. It is not supported for now.
```
lamblin: Remove "It is not supported for now."

theano/tensor/tests/test_opt.py
lamblin: Oh, wait, it is still not supported by the Op.
In the end, I think we should just remove that part, until the end of the test.
In the future, we should add optimizations and tests for the reshape/dimshuffles introduced by the new T.tile, but that's not for now.

theano/tensor/tests/test_opt.py
Saizheng: There is another error which is not just Flake8 error:
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/tensor/basic.py:4762:47: F821 undefined name 'long'

This 'long' type is in the dtype check in tile, but it seems that this dtype is undefined in python 3.3?

lamblin: Right, `long` does not exist in Python 3 (all ints are long ints). You con use `six.integer_types` instead.

theano/tensor/basic.py
```diff
@@ -4705,32 +4706,67 @@ def tile(x, reps, ndim=None):
...
+                reps_ = [switch(i < offset, 1, reps[i - offset]) for i in range(ndim)]
+                reps = reps_
+
+        # other raise error
+        else:
+            raise ValueError("the dimension of reps should not exceed 1")
+    else:
+        if ndim is not None and len(reps) > ndim:
+            raise ValueError("len(reps) should be equal or less than ndim")
+        if not numpy.all([isinstance(r, (int, long)) or
```
lamblin: For instance, this would become `if not numpy.all([isinstance(r, integer_types) or`

lamblin:
Thanks a lot!

PR #2750, opened 2015-04-10
Tracking Ops with grad methods returning None

SinaHonari:

Issue #2863, opened 2015-05-02
Write test for bug introduced by #2750


dwf:
https://github.com/Theano/Theano/pull/2750/files#diff-3a81c280f7c4d50677c682af9ee70240L1158

Steps to fix (in this order):
* Write a test for the grad of this op.
* Verify that it fails.
* Fix the bug (add "self." in front of the "get_grad" on the above-mentioned line).
* Verify that this makes your test pass.

Ideally make a commit for each of these steps.

dwf:
@thomasmesnard This is your next ticket. It is fairly straightforward, let me know if you have any questions.

lamblin:
Should this be closed, or are there other steps remaining?

dwf:
Arnaud fixed the bug. @thomasmesnard is still assigned to write the test.

PR #2877, opened 2015-05-05
Fix nit spotted by @dwf

abergeron:
See: https://github.com/Theano/Theano/pull/2750/files#diff-3a81c280f7c4d50677c682af9ee70240R1171

This would be better with a test, but I don't have the time.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -1190,7 +1190,7 @@ class InRange(LogicalComparison):
         (gz,) = gout
         grads = []
         for elem in [x, low, hi]:
-            grads.append(get_grad(elem))
+            grads.append(self.get_grad(elem))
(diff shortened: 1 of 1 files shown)


theano/scalar/basic.py
```diff
@@ -1190,7 +1190,7 @@ def grad(self, inputs, gout):
         (gz,) = gout
         grads = []
         for elem in [x, low, hi]:
-            grads.append(get_grad(elem))
+            grads.append(self.get_grad(elem))
```
nouiz: This mean tests are missing. Otherwise, we would have caught it.

dwf: @nouiz See #2863, which @thomasmesnard is going to take care of.

PR #2886, opened 2015-05-07
Fix scan test for float64

carriepl:
With floatX=float64, rs.uniform would produce a float64 output which would prevent the Scan from being moved to the GPU.


nouiz:
we can merge when travis pass

carriepl:
I just changed the description of the issue because I realized I had written "float32" instead of "float64".

PR #2887, opened 2015-05-07
Fixed typo in custom pickler

dmitriy-serdyuk:


theano/misc/pkl_utils.py
```diff
@@ -166,7 +166,7 @@ def write_array(f):
                 zipadd(write_array, self.zip_file, name)
                 self.seen[id(obj)] = 'cuda_ndarray.{0}'.format(name)
             return self.seen[id(obj)]
-        super(PersistentCudaNdarrayID, self).__call__(obj)
+        return super(PersistentCudaNdarrayID, self).__call__(obj)
```
nouiz: What was the consequence of this error?

dmitriy-serdyuk: It used regular pickler in case of Cuda Ndarray instead of custom one.

nouiz: Can you add a test for this case?

dmitriy-serdyuk: Ok, I'll add it.

dmitriy-serdyuk:
@nouiz , seems that this test is more difficult than I expected. The problem that without a return statement arrays are saved twice: in an npy file and in a pkl file, but the pkl file is used for unpickling. So I cannot test it without changing a lot in internals of `Pickler` and `Unpickler`.

But if you insist, I'll do it.

nouiz:
Ok, we can merge like this.

thanks

PR #2888, opened 2015-05-07
Update some GPU tests

nouiz:


abergeron:
Ok when travis passes.

PR #2889, opened 2015-05-08
Try to fix autodoc generation

lamblin:


lamblin:
Combined with the update of NumPy on deeplearning.net, this should fix the doc generation for Theano.

nouiz:
Ok, so I merge. thanks

PR #2892, opened 2015-05-08
fix crash in debugmode

nouiz:


carriepl:
I'm fine with this once travis is done.

Issue #2378, opened 2015-01-07
Finish/merge in Theano batched gemm on GPU 


nouiz:
https://github.com/benanne/theano_wmf/blob/master/ops.py

which I found in this thread:
https://groups.google.com/forum/#!searchin/theano-users/batched$20dot/theano-users/NpO2uG0d9uo/GzFdaIdd7FIJ

- move to: theano/sandbox/cuda/blas.py
- make sure there is tests in test_blas.py for it
- remove make_thunk and make c code.
- later, remove gpu contiguous?

briancheung:
Just wanted to check in to see if there needs to be anything done for this. Batched gemm seems like a much better solution for batched_dot rather than using a scan.

benanne:
Afaik no work has been done for this so far (but the Theano devs should correct me if I'm wrong). A version of this op with C-code as part of Theano would be awesome! And it would speed up the FFT convolution a bit as well, probably :)

bouthilx:
It should be done in the next two weeks.

nouiz:
Module insertitute with NIPS deadline... But it is on our short term plan.

PR #2893, opened 2015-05-08
GPU batched gemm

sebastien-j:
@bouthilx. Addresses issue #2378. Work in progress.

- Moved batched gemm (by @benanne) to sandbox/cuda/blas
- Made unit test
- TODO: C code

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -14,6 +14,192 @@ from theano.sandbox.cuda.basic_ops import (as_cuda_ndarray_variable,
                                            gpu_contiguous)
 from theano.tensor import as_tensor_variable
 
+class BatchedDotOp(GpuOp):
+
+    __props__ = ()
+
+    def make_node(self, inp1, inp2):
+        inp1 = gpu_contiguous(as_cuda_ndarray_variable(inp1))
+        inp2 = gpu_contiguous(as_cuda_ndarray_variable(inp2))
+
+        assert inp1.dtype == "float32"
+        assert inp2.dtype == "float32"
+        assert inp1.ndim == 3 # (batch, a, b)
+        assert inp2.ndim == 3
+
+        return theano.Apply(self, [inp1, inp2],
+                           [self.output_type(inp1, inp2)()])
+
+    def output_type(self, inp1, inp2):
+        return CudaNdarrayType(
+            (inp1.type.broadcastable[0] or inp2.type.broadcastable[0],
+             inp1.type.broadcastable[1], inp2.type.broadcastable[2]))
+
+    def c_code(self, node, name, input_names, output_names, sub):
+        bx, by = input_names
+        bz, = output_names
+        fail = sub['fail']
+        return """
+        float alpha = 1.0;
+        float beta = 0.0;
+
+        int i, x_dim0, x_dim1, x_dim2, y_dim0, y_dim1, y_dim2;
+        int x_stride, y_stride, z_stride, total_size;
+        int ptr_array_size = 3 * CudaNdarray_HOST_DIMS(%(bx)s)[0] * sizeof(float *);
+        int out_dim[3];
+
+        cublasStatus_t err;
+        cudaError_t err1;
+
+        float **host_x = NULL;
+        float **host_z = NULL;
+        float **host_y = NULL;
+
+        float **gpu_x = NULL;
+        float **gpu_y = NULL;
+        float **gpu_z = NULL;
+
+        x_dim0 = CudaNdarray_HOST_DIMS(%(bx)s)[0];
+        x_dim1 = CudaNdarray_HOST_DIMS(%(bx)s)[1];
+        x_dim2 = CudaNdarray_HOST_DIMS(%(bx)s)[2];
+
+        y_dim0 = CudaNdarray_HOST_DIMS(%(by)s)[0];
+        y_dim1 = CudaNdarray_HOST_DIMS(%(by)s)[1];
+        y_dim2 = CudaNdarray_HOST_DIMS(%(by)s)[2];
+
+        if (x_dim0 != y_dim0)
+        {
+            PyErr_Format(PyExc_RuntimeError,
+                    "The batchsizes (%%d, %%d) don't match.\\n",
+                    x_dim0, x_dim1);
+            %(fail)s;
+        }
+
+        if (x_dim2 != y_dim1)
+        {
+            PyErr_Format(PyExc_RuntimeError,
+                    "Shape mismatch. (%%d, %%d, %%d) (%%d, %%d, %%d)\\n",
+                    x_dim0, x_dim1, x_dim2, y_dim0, y_dim1, y_dim2);
+            %(fail)s;
+        }
+
+        out_dim[0] = x_dim0;
+        out_dim[1] = x_dim1;
+        out_dim[2] = y_dim2;
+
+        if ( !(%(bz)s
+               && %(bz)s->nd==3
+               && CudaNdarray_is_c_contiguous(%(bz)s)
+               && CudaNdarray_HOST_DIMS(%(bz)s)[0]==out_dim[0]
+               && CudaNdarray_HOST_DIMS(%(bz)s)[1]==out_dim[1]
+               && CudaNdarray_HOST_DIMS(%(bz)s)[2]==out_dim[2]))
+        {
+            Py_XDECREF(%(bz)s);
+            %(bz)s = (CudaNdarray*)CudaNdarray_NewDims(3,out_dim);
+            if (NULL == %(bz)s)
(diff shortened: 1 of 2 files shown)


sebastien-j:
@bouthilx ,

I just wrote the C code.

To do:
- Write `c_code_cache_version` after review.
- Set seed in tests

Questions:
- Should the dimension checks be done inside the C code?
- Do I need additional INCREFs/DECREFs?
- Can we create the pointer array directly on GPU?
- Should I use strides instead of dimensions?

bouthilx:
@nouiz  Sébastien pointed to me some [code](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/cuda_ndarray.cu#L4023) in ```CudaNdarray_gemm``` where we handle axes of dimension 0. How could it be 0? Is it a synonym for 1? To me, an axe of scrict dimension 0 would just push down the tensor to size 0 (ex: shape = (10, 0, 1) => total size=10x0x1=0).

lamblin:
@bouthilx Yes, an axis of dimension 0 means that the array is empty, but it still has dimensions.
If we follow numpy, you can have things like:
```python
>>> x = np.zeros((0, 10))
>>> y = np.ones((0, 10))
>>> (x == y).all()
True
>>> (x == y).any()
False
>>> x.dot(y.T)
array([], shape=(0, 0), dtype=float64)
>>> x.T.dot(y)
array([[ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.]])
>>> x[:, :5] + y
ValueError: operands could not be broadcast together with shapes (0,5) (0,10)
```

bouthilx:
@lamblin Wow that is so weird. Is there any use case where this is usefull? 

Thank you for the answer!

lamblin:
I guess it can be useful for things not to break down when indexing with a list of indices that could be empty, or slicing after the end of an array for instance, like if you slice after the end of a list.

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,208 @@
...
+        {
+        CLEANUP();
+        PyErr_Format(PyExc_RuntimeError,
+                     "%%s", "malloc failure");
+                %(fail)s;
+        }
+
+        for (i = 0; i < out_dim[0]; i++)
+        {
+            host_x[i] = CudaNdarray_DEV_DATA(%(bx)s) + i * x_dim1 * x_dim2;
```
bouthilx: Should be done with stride no? You could do ```host_x[0] = CudaNdarray_DEV_DATA(%(bx)s)``` and then ```host_x[i] = host_x[i-1] + x_stride``` for the others.

bouthilx:
Looks good to me overall. I believe pointers could be computed more efficiently.

@sebastien-j Do you know if cublasSgemmBatched supports 0 dimension?

nouiz:
I would not suppose this. You can just not call it in that case and init
the memory the the value if needed.

sebastien-j:
@bouthilx , ready for review.

bouthilx:
Seems ready to be merged. @nouiz Do you want to take a final look before we merge?

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,220 @@
...
 
+class BatchedDotOp(GpuOp):
+
+    def __str__(self):
+        return 'BatchedDotOp'
+
+    def __eq__(self, other):
+        return type(self) == type(other)
+
+    def __hash__(self):
```
nouiz: Remove those 3 methods. This is the old way of doing things. Add this attribute:

`__props__ = ()`

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,220 @@
...
+
+        assert inp1.dtype == "float32"
+        assert inp2.dtype == "float32"
+        assert inp1.ndim == 3 # (batch, a, b)
+        assert inp2.ndim == 3
+
+        return theano.Apply(self, [inp1, inp2], [self.output_type(inp1)()])
+
+    def output_type(self, inp):
+        return CudaNdarrayType(broadcastable=[False] * inp.type.ndim)
```
nouiz: That is ok, but not the best. If inp1 and inp2 have broadcastable dimensions, we can infer some broadcastable dims for the outputs. If inp1.broad[0] or inp2.broad[0], then we know the output should be broadcastable in that dim.

If inp1.broad[1] is broad, out.broad[1] is broad. If inp2.broad[2] is broad, out.broad[2] is broad.

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,220 @@
...
+        cudaError_t err1, err2, err3;
+        cublasHandle_t handle = NULL;
+
+        float **host_x;
+        float **host_y;
+        float **host_z;
+
+        float **gpu_x;
+        float **gpu_y;
+        float **gpu_z;
```
nouiz: Init the ptr to NULL, otherwise, the if(host_x) could be troublesome. But your code is probably fine, this is just an extra safeguard.

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,220 @@
...
+
+            if (err1 != cudaSuccess || err2 != cudaSuccess || err3 != cudaSuccess)
+            {
+                CLEANUP();
+                PyErr_Format(PyExc_RuntimeError,
+                             "%%s", "cudaMemcpy failure");
+                %(fail)s;
+            }
+
+            err = cublasCreate(&handle);
```
nouiz: It should be already created. Do not create it again.

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,220 @@
...
+
+            if (CUBLAS_STATUS_SUCCESS != err)
+            {
+                PyErr_Format(PyExc_RuntimeError,
+                             "cublasSgemmBatched failed (%%i) %%s",
+                             err,  cublasGetErrorString(err));
+                %(fail)s;
+            }
+
+            err = cublasDestroy(handle);
```
nouiz: don't destroy it, as handle is global and already init.

theano/sandbox/cuda/blas.py
```diff
@@ -14,6 +14,220 @@
...
+            for (i = 1; i < out_dim[0]; i++)
+            {
+                host_x[i] = host_x[i-1] + x_stride;
+                host_y[i] = host_y[i-1] + y_stride;
+                host_z[i] = host_z[i-1] + z_stride;
+            }
+
+            err1 = cudaMalloc((void **)&gpu_x, ptr_array_size);
+            err2 = cudaMalloc((void **)&gpu_y, ptr_array_size);
+            err3 = cudaMalloc((void **)&gpu_z, ptr_array_size);
```
nouiz: You could do only 1 big malloc. This would be faster (malloc costly and 1 bigger transfer is faster then 1 smaller).

sebastien-j:
@bouthilx , I updated the code to address @nouiz 's comments.

theano/sandbox/cuda/blas.py
bouthilx: @sebastien-j What about doing 2 big malloc (one on host one on device) and then set {host,gpu}_{x,y,z} pointers accordingly. Not that it doesn't work with current modifications, but readability suffers a lot from it.

sebastien-j:
@bouthilx , ok now?

bouthilx:
@sebastien-j Looks good. Can you squash your commits before I merge? They are mostly fixup/undo/redo commits.

sebastien-j:
@bouthilx , done.

bouthilx:
GPU Tests pass locally. :shipit:

sebastien-j:
We may close #2378.

nouiz:
That is good. I made a follow up ticket: https://github.com/Theano/Theano/issues/2968

PR #2894, opened 2015-05-09
Fix problems with advsub in the gpuarray backend

abergeron:
There where problems when the data element had to be broadcasted in some cases.


nouiz:
Can you confirm that existing tests fails due to that and that we do not need new tests?

nouiz:
My review tell it is ok. I'm waiting for the confirmation about the tests. I'll also run the tests before merging.

PR #2898, opened 2015-05-12
Fix crash with GpuGemm.

abergeron:


nouiz:
I'm good with those 2 commits. Anyone can merge when travis pass.

abergeron:
I forgot to bump the c_code_version for GpuGemm.

PR #2899, opened 2015-05-12
Fix tests in new backend

carriepl:
This fixes a few failing tests in Scan for the gpuarray backend

For the clone=True in gpuarray/opt.py, `FunctionGraph` was previously raising exceptions saying the graph contained cached constants and that the graph should be cloned.

### DIFF
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -723,7 +723,7 @@ def local_scan_to_gpua(node):
     # __init__ does not know about the gpu and can not
     # handle graphs with inputs being on the gpu
     tmp_in, tmp_out = gpu_reconstruct_graph(scan_ins, scan_outs)
-    local_fgraph = gof.FunctionGraph(tmp_in, tmp_out, clone=False)
+    local_fgraph = gof.FunctionGraph(tmp_in, tmp_out, clone=True)
     _cmodule_key = gof.CLinker().cmodule_key_(local_fgraph, [])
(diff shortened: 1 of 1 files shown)


theano/scan_module/tests/test_scan.py
```diff
@@ -625,12 +625,11 @@ def f_rnn(u_t, x_tm1, W_in, W):
...
-        scan_node = [node for node in topo
-                     if isinstance(node.op, theano.scan_module.scan_op.Scan)]
-        assert len(scan_node) == 1
-        scan_node = scan_node[0]
-        assert scan_node.op.gpu
+        # Ensure f2 contains only one scan node and it's been moved to the GPU
+        scan_nodes = self.scan_nodes_from_fct(f2)
+        assert len(scan_nodes) == 1
+        assert (scan_nodes[0].op.info.get('gpu', False) or
+                scan_nodes[0].op.info.get('gpua', False))
```
abergeron: I would check for gpu only here since the function uses mode_with_gpu.  This shouldn't get moved to gpua.

carriepl: The goal was to make tests able to test both backends but after chatting with @nouiz, I'll proceed differently in a follow-up PR. The tests with the cuda backend will be moved to a different test class with attributes refering to the cuda backend and another test class will inherit from it and redefine the attributes to point to the GpuArray backend. 

I removed the contentious commit from this PR and only kept the one with which you didn't seem to have any issue.

PR #2901, opened 2015-05-12
Work to make the DLT LSTM work with float16.

abergeron:


theano/tensor/nnet/sigm.py
```diff
@@ -59,15 +59,8 @@ def c_code(self, node, name, inp, out, sub):
         # 0. The reset of the logic always generate 0 or 1 in those
         # cases. This is a speed optimization.
         # The constants were obtained by looking at the output of python commands like:
```
lamblin: Update the comment if you remove the code it is referring to.

nouiz:
In the ticket for f16 follow up, put all the op that you marked as float16 safe and tell we need to test them. Put as an item, make the list of op that isn't supported on CPU/GPU.

theano/compile/debugmode.py
```diff
@@ -1708,10 +1708,20 @@ def make_all(self, profiler=None, input_storage=None
...
+
+                # Don't try to test the C code for float16 if not
+                # tagged ok.
+                if (not getattr(node.op, '_f16_ok', False) and
+                        (any(getattr(i, 'dtype', '') == 'float16'
+                             for i in node.inputs) or
+                         any(getattr(o, 'dtype', '') == 'float16'
+                             for o in node.outputs))):
+                    raise utils.MethodNotDefined()
+
```
nouiz: This continue to add duplicated code here from Op.make_thunk(). Could you refactor that in a method make_c_thunk in the file gof/op.py?

theano/tensor/nnet/sigm.py
abergeron: I will undelete and reformat it.  It looked like dead code.

theano/compile/debugmode.py
abergeron: I'll also do it for python thunks.

abergeron:
I've made a series of changes that should address all comments.

abergeron:
I fixed the problem with DebugMode that was the source of failing tests.  I'm not 100% convinced that the change is the right one since I make the python and C thunk share the compute map.

If you have a better solution, I'm all ears.

kyunghyuncho:
@nouiz @abergeron Does it mean that any code can be run with float16 as long as it only includes those OPs used in DLT LSTM tutorial? Does simply setting floatX=float16 work?

nouiz:
Not yet.

PR #2902, opened 2015-05-13
Make gpuarray.GpuFromHost.r_Op return a list

carriepl:


abergeron:
This doesn't need a travis run.

PR #2903, opened 2015-05-13
Speed reduce 100

nouiz:
I saw this as being a significant time in the profile from one user.


abergeron:
This looks ok, but I haven't done a detailed review.

nouiz:
If you think it is good, merge it. Run the careduce tests on your comuters before.

As it reuse existing code, it don't need as much review, but if you can check that I reuse it correctly, that is good.

PR #2905, opened 2015-05-14
Fix test due to new error type.

nouiz:


lamblin:
A CPU test expects an IndexError (test_join_matrixC_negative_axis at https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_basic.py#L3681), so it is failing now. I'm not sure which error is better, but we should be consistent in the tests.

abergeron:
This changes IndexError to ValueError, so it can't fix the test you are talking about.

abergeron:
In fact I'd say the first commit broke the test.

nouiz:
I think the error should be ValueError. IndexError mean there is a problem during indexing. This is also what numpy does:

```
$python
Enthought Canopy Python 2.7.6 | 64-bit | (default, Apr 11 2014, 13:46:04) 
[GCC 4.1.2 20080704 (Red Hat 4.1.2-52)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
n>>> numpy.sum([1,2], axis=2)
  File "/Tmp/lisa/os_v3/canopy/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 1709, in sum
    out=out, keepdims=keepdims)
  File "/Tmp/lisa/os_v3/canopy/lib/python2.7/site-packages/numpy/core/_methods.py", line 25, in _sum
    out=out, keepdims=keepdims)
ValueError: 'axis' entry is out of bounds

```

I updated this PR to fix this tests.

abergeron:
You might have to change that too: https://github.com/Theano/Theano/blob/master/theano/sandbox/gpuarray/basic_ops.py#L862

nouiz:
done

bouthilx:
Actually, I looked at Numpy to decide it was an IndexError

nouiz:
done

bouthilx:
Actually, I looked at Numpy to decide it was an IndexError
```python
np.concatenate([np.zeros((20,20)),np.zeros((20,20))], axis=3)
IndexError: axis 3 out of bounds [0, 2)
```

nouiz:
How you tested it?

bouthilx:
@nouiz It seems numpy.concatenate doesn't give same error as numpy.sum

nouiz:
That seem what happened. But Join call numpy.concatenate. I don't like that
error type, but anyway, it is better to stay as numpy. I'll change it again.

nouiz:
I updated the PR again.

PR #2909, opened 2015-05-14
improving debugprint for scan ops

### LINKED ISSUE #2347: Better scan debugprint (show link with the outer graph)

aalmah:
- Fixes #2347. 
- I have tested this for 3 cases: one scan, two separate scans (added together), two nested scans.
- We need to find a good way of testing this other than looking at the debug print and making sure that references from the inner graph to the outer graph make sense.

### DIFF
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -521,7 +521,8 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
                file=sys.stdout, print_destroy_map=False,
                print_view_map=False, order=None, ids='CHAR',
                stop_on_name=False, prefix_child=None,
-               scan_ops=None, profile=None):
+               scan_ops=None, profile=None,
+               scan_inner_to_outer_inputs=None):
     """Print the graph leading to `r` to given depth.
 
     :param r: Variable instance
@@ -544,6 +545,8 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
                          we don't print anything below it.
     :param scan_ops: Scan ops in the graph will be added inside this list
                      for later printing purposes.
+    :param scan_inner_to_outer_inputs: a dictionary mapping a scan ops inner function
+    inputs to the scan op inputs (outer inputs) for printing purposes.
 
     """
     if depth == 0:
@@ -578,6 +581,7 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
         elif ids == "":
             id_str = ""
         done[obj] = id_str
+
         return id_str
 
     if hasattr(r.owner, 'op'):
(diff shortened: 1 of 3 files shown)


nouiz:
For test, we can hardcode an expected output of debugprint that we know to
be good.

I didn't checked the code

carriepl:
The tests for flake8 are failing.

theano/compile/debugmode.py
```diff
@@ -524,6 +525,8 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
                          we don't print anything below it.
     :param scan_ops: Scan ops in the graph will be added inside this list
                      for later printing purposes.
+    :param scan_inner_to_outer_inputs: a dictionary mapping a scan ops inner function
+    inputs to the scan op inputs (outer inputs) for printing purposes.
```
carriepl: indentation

theano/printing.py
```diff
@@ -156,17 +157,31 @@ def debugprint(obj, depth=-1, print_type=False,
...
+            if hasattr(s.owner.op, 'fn'):
+                # If the op was compiled, print the optimized version.
+                inner_inputs = s.owner.op.fn.maker.fgraph.inputs
+            else:
+                inner_inputs = s.owner.op.inputs
+            outer_inputs = s.owner.inputs
+            inner_to_outer_inputs = dict([(inner_inputs[i],outer_inputs[o])
+                                          for i,o in enumerate(
+                                                  s.owner.op.get_outer_iidx_from_inner_iidx_seq())])
+            #import pdb; pdb.set_trace()
```
carriepl: This comment can be removed.

carriepl:
For the tests, you can do @nouiz suggested and hardcode some outputs you know to be good. I think your test case(s) for this should include scans with sequence inputs, non-sequences inputs, non-recurrent states as well as recurrent states. It should also include scan nodes obtained by taking the gradient of user-defined scans because those types of scans can have a type of recurrent states (mitmots) that user-defined scans cannot.

theano/compile/debugmode.py
```diff
@@ -659,12 +663,25 @@ def get_id_str(obj):
...
+           r in scan_inner_to_outer_inputs:
+
+            id_str = get_id_str(r)
+            outer_r = scan_inner_to_outer_inputs[r]
+
+            if hasattr(outer_r.owner, 'op'):
+                outer_id_str = get_id_str(outer_r.owner)
+            else:
+                outer_id_str = get_id_str(outer_r)
+            print('%s%s %s%s -> %s' % (prefix, r, id_str, type_str, outer_id_str), file=file)
```
carriepl: An arrow might not be the most evident way to signal that one variable refers in a scan to another outside of the scan. Maybe something like "corresponds to" or else. Or maybe keep the array but print something after the debugprints of the scan function to explain the meaning of the arrow. @nouiz  and @lamblin : any opinion on this?

carriepl:
This looks good and it should be a very useful feature for debugging. Thanks a lot.

PR #2911, opened 2015-05-15
[CRASH] Scan index error

carriepl:
This is a fix for the segfault issue reported on the theano-users mailing list  (https://groups.google.com/forum/#!topic/theano-users/ryurT5G6G7M/discussion).

The issue was with the scanOp_pushout_seqs optimization which was trying to push out computation that did not depend on sequences and, therefore, should have been handled by scanOp_pushout_non_seqs.


carriepl:
@abergeron or @lamblin : @nouiz  had a look and said it looks fine but a second set of eyes wouldn't be a luxury. Can any of you spare a few minutes?

Issue #2913, opened 2015-05-15
A bug in 'local_gpu_lazy_ifelse'


rizar:
The bug is here: https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/opt.py#L490

When the condition variable (the first argument of `theano.ifelse.ifelse`) is an output of `HostFromGpu`, two other inputs are subjected to `GpuFromHost` (https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/opt.py#L503). But it might be that the condition variable is `float32` and others have different dtypes.

An example:

```python
import numpy
import theano
import theano.ifelse
from theano import tensor
from theano.printing import debugprint

c = theano.shared(numpy.float32(1.0))
x = theano.shared(numpy.zeros((3, 3)))
y = theano.ifelse.ifelse(c, x.mean(), x[0, 0])
theano.function([], [y])
```

This is quite a bad issue for Blocks, because we are using `ifelse` deep inside and the user should be utterly confused getting an optimization failure when they did nothing special.

nouiz:
I'm owkring on it. I'll have a fix shortly.

nouiz:
https://github.com/Theano/Theano/pull/2914

PR #2914, opened 2015-05-15
Ifelse remove warning and speed up

nouiz:
fix https://github.com/Theano/Theano/issues/2913
fix https://github.com/bartvm/blocks/issues/633

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
   - cd $(python -c 'import theano; import os; print(os.path.split(theano.__file__)[0])')
   - echo "$PART"
   - theano-nose -v $PART
-  - theano-cache list
+#  - theano-cache list
 
 #after_script:
 after_failure:
--- a/theano/ifelse.py
+++ b/theano/ifelse.py
@@ -250,7 +250,7 @@ class IfElse(PureOp):
                             if self.as_view:
                                 storage_map[out][0] = val
                             # Work around broken numpy deepcopy
-                            elif type(a) in (numpy.ndarray, numpy.memmap):
+                            elif type(val) in (numpy.ndarray, numpy.memmap):
                                 storage_map[out][0] = val.copy()
                             else:
                                 storage_map[out][0] = deepcopy(val)
--- a/theano/ifelse.py
+++ b/theano/ifelse.py
@@ -246,10 +246,9 @@ class IfElse(PureOp):
                         for out, outtype, t in izip(outputs, outtypes, ts):
                             compute_map[out][0] = 1
                             if self.as_view:
-                                oval = outtype.filter(storage_map[t][0])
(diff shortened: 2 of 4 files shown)


rizar:
The PR is definitely doing the right thing, but I am not proficient enough in Theano, especially in its CPU-GPU aspects, to review it.

lamblin:
Tests are failing with "ValueError: expected an ndarray", so maybe the filtering was necessary after all.

nouiz:
I updated this pr.

theano/ifelse.py
```diff
@@ -245,12 +246,14 @@ def thunk():
                     else:
                         for out, outtype, t in izip(outputs, outtypes, ts):
                             compute_map[out][0] = 1
+                            val = storage_map[t][0]
                             if self.as_view:
-                                oval = outtype.filter(storage_map[t][0])
+                                storage_map[out][0] = val
+                            # Work around broken numpy deepcopy
+                            elif type(a) in (numpy.ndarray, numpy.memmap):
```
lamblin: Do you mean `type(val)`?

lamblin: Also, can you simply use `_less_broken_deepcopy`? I think it is essentially doing the same thing.

nouiz: I don't want that. It add computation time (ifelse overhead is already big) and it do something bad for cdatatype (not copy them). Here we need a crash in case we can't deep copy it.

nouiz: I fixed the typo.

lamblin:
Could you reuse https://github.com/Theano/Theano/blob/57a52cab600dbac2b6e09a0b5c0e3ed93e27a5d7/theano/compile/debugmode.py#L918 ?

.travis.yml
```diff
@@ -49,7 +49,7 @@ script:
   - cd $(python -c 'import theano; import os; print(os.path.split(theano.__file__)[0])')
   - echo "$PART"
   - theano-nose -v $PART
-  - theano-cache list
+#  - theano-cache list
```
lamblin: Thanks :)

theano/ifelse.py
lamblin: Oh, OK, I missed your answer since the new push collapsed it.

theano/ifelse.py
```diff
@@ -245,12 +246,14 @@ def thunk():
                     else:
                         for out, outtype, t in izip(outputs, outtypes, ts):
```
abergeron: You don't need the outtype anymore, so maybe you can remove it from the iteration.

abergeron:
This looks good, although we might want to precompute the list of input and output storage cells in order to get even more speedup.  I haven't tried, so maybe it won't help.

PR #2915, opened 2015-05-15
Move gpu tests from T_Scan to ScanGpuTests

carriepl:
This PR refactors the gpu tests in test_scan.py to make them more easily usable to test both the cuda and the gpuarray backends.

The tests that can be made to work for both backends are moved to ScanGpuTests and are modified to rely on certain attributes/methods that will manage the interactions with the concrete gpu backend. The classes T_Scan_Cuda and T_Scan_Gpuarray inherit from ScanGpuTests and TestCase and implement these attributes/methods so that the tests will rely on their respective backends.

Some tests were highly cuda specific so they we moved to T_Scan_Cuda instead of ScanGpuTests.

### DIFF
--- a/theano/sandbox/gpuarray/__init__.py
+++ b/theano/sandbox/gpuarray/__init__.py
@@ -54,6 +54,14 @@ if pygpu:
             optdb.add_tags('gpuarray_opt', 'fast_run', 'fast_compile', 'inplace')
         elif config.gpuarray.init_device != '':
             init_dev(config.gpuarray.init_device)
+
+        from .basic_ops import (GpuAlloc, GpuContiguous, GpuEye, GpuFromHost,
+                                GpuJoin, GpuReshape, GpuSplit, HostFromGpu)
+        from .basic_ops import host_from_gpu, gpu_from_host
+        from .elemwise import GpuElemwise
+        from .subtensor import (GpuSubtensor, GpuIncSubtensor,
+                                GpuAdvancedIncSubtensor1)
+
     except Exception:
         error("Could not initialize pygpu, support disabled", exc_info=True)
 else:
--- a/theano/scan_module/tests/test_scan.py
+++ b/theano/scan_module/tests/test_scan.py
@@ -432,207 +432,6 @@ class T_Scan(unittest.TestCase):
         theano_values = f2(v_u, v_x0, W_in, W)
         utt.assert_allclose(theano_values, v_out)
 
-    # as test_one_sequence_one_output_weights, but on the gpu
-    # This first version test the first case in the optimizer to the gpu.
-    def test_one_sequence_one_output_weights_gpu1(self):
-        from theano.sandbox import cuda
-        if not cuda.cuda_available:
-            raise SkipTest('Optional package cuda disabled')
-
-        def f_rnn(u_t, x_tm1, W_in, W):
-            return u_t * W_in + x_tm1 * W
-
-        u = theano.tensor.fvector('u')
-        x0 = theano.tensor.fscalar('x0')
-        W_in = theano.tensor.fscalar('win')
-        W = theano.tensor.fscalar('w')
-
-        # The following line is needed to have the first case being used
-        # Otherwise, it is the second that is tested.
-        mode = mode_with_gpu.excluding('InputToGpuOptimizer')
-        output, updates = theano.scan(f_rnn,
-                                      u,
-                                      x0,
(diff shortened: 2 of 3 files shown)


theano/sandbox/gpuarray/__init__.py
```diff
@@ -54,6 +54,14 @@ def init_dev(dev):
             optdb.add_tags('gpuarray_opt', 'fast_run', 'fast_compile', 'inplace')
         elif config.gpuarray.init_device != '':
             init_dev(config.gpuarray.init_device)
+
+        from .basic_ops import (GpuAlloc, GpuContiguous, GpuEye, GpuFromHost,
+                                GpuJoin, GpuReshape, GpuSplit, HostFromGpu)
+        from .basic_ops import host_from_gpu, gpu_from_host
+        from .elemwise import GpuElemwise
+        from .subtensor import (GpuSubtensor, GpuIncSubtensor,
+                                GpuAdvancedIncSubtensor1)
```
abergeron: I don't like this part.  Making the exported list variable is a bad idea.

theano/scan_module/tests/test_scan.py
```diff
@@ -4417,6 +4019,434 @@ def _scan_loose(x):
...
+
+
+class T_Scan_Gpuarray(unittest.TestCase, ScanGpuTests):
+    """This class takes the gpu tests for scan that are defined in
+    class ScanGpuTests and runs them using the gpuarray backend.
+    """
+
+    def __init__(self, *args, **kwargs):
+        from theano.sandbox import gpuarray
+        self.gpu_backend = gpuarray
```
abergeron: To avoid the variable import you could make an object on which you set the attributes you need that you fetch from the various submodules of gpuarray.

theano/scan_module/tests/test_scan.py
```diff
@@ -4417,6 +4019,434 @@ def _scan_loose(x):
...
+        self.mode_with_gpu = mode_with_opt.including('gpuarray_opt', 'scan')
```
abergeron: I'm pretty sure the tag for gpuarray is 'gpua'.  Unless you mean something different here.

abergeron:
Other that my comments, I like this direction.  This could even mean that we could move the backend-specific classes into their respective subdirectories eventually.

theano/scan_module/tests/test_scan.py
carriepl: The only thing that worried me a bit about that approach is that this `gpu_backend` object could end up quite big and annoying as the tests require more and more attributes from the gpu backends where as the current approach would have the side effect of making the interfaces of the gpu backends more similar and compatible. For my own benefit, what are the negative points we are trying to avoid by not proceeding the way it is implemented in this PR?

theano/scan_module/tests/test_scan.py
carriepl: I'm a bit confused. Shoulf it be 'gpuarray_opt' (https://github.com/Theano/Theano/blob/master/theano/sandbox/gpuarray/__init__.py#L54, )? 'gpuarray' (https://github.com/Theano/Theano/blob/master/theano/sandbox/gpuarray/opt.py#L46) or 'gpua' (which I think is rather the name of the key in scan_op.info)

nouiz: It is not consistent. The most frequently used tag is gpuarray, but the scan opt use gpua. @carriepl can you change the scan opt to use gpuarray and use this one instead of gpuarray_opt.

theano/scan_module/tests/test_scan.py
nouiz: @abergeron I agree that normally, listing them is fine as we have a small finit set that will be used. But in this this, this is scan and this will trigger many stuff to be used and more to be added in the futur. As the code work now, I would keep it like this.

Also, I like the idea that the 2 modules have a similar interface by default. If we want to change that, we should have a reason.

theano/scan_module/tests/test_scan.py
carriepl: Done

theano/scan_module/tests/test_scan.py
carriepl: @abergeron, @nouiz : So what's the verdict on this one? I can implement either solution you feel is better suited but I'd like to do it soon because some of my upcoming pull requests would be much simpler if this was merged first.

PR #2916, opened 2015-05-15
In local_scan.to_gpua, don't transfer as_while condition on gpu

carriepl:


nouiz:
Do this need a test?

carriepl:
It already has tests, this modification makes the test_while0, test_while1, test_while2 and test_while_infershape tests pass in test_scan.py for the new gpu backend. Without this modification, the tests fail.

PR #2923, opened 2015-05-19
broadcastable option to tensor.alloc

kyunghyuncho:
Added a broadcastable option to tensor.alloc. When ``broadcatable=False``, no axis will be set broadcastable even when its dim is 1. The default value is ``True``. As an example,

```
In [2]: from theano import tensor
In [3]: a = tensor.alloc(1,1,10)
In [4]: print a.broadcastable
(True, False)
In [5]: b = tensor.alloc(1,1,10,broadcastable=False)
In [6]: print b.broadcastable
(False, False)
In [7]: c = tensor.alloc(1,1,10,broadcastable=True)
In [8]: print c.broadcastable
(True, False)
```

This becomes handy when tensor.alloc is used to initialize a variable passed to the inner scan function as a part of ``outputs_info``, since the broadcastable patterns affects the variable filter inside the scan op.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -2400,7 +2400,7 @@ class Alloc(gof.Op):
     """
     __props__ = ()
 
-    def make_node(self, value, *shape):
+    def make_node(self, value, *shape, **kwargs):
         v = as_tensor_variable(value)
         sh = [as_tensor_variable(s) for s in shape]
         bcast = []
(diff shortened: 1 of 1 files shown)


kyunghyuncho:
Work in progress: Do not merge.

nouiz:
Currently, we tell people to use tensor.unbroadcast(...) and this work. Why use a different mechanism?

I told you verbally, that this would need change to GpuAlloc  and its optimization, but I forgot that now this isn't needed anymore. unbroadcast would be added automatically by all optimization to unbroadcast if needed. So this could be workable. But I'm not sure I want to add an new interface for this. I think it make using Theano more complicated as this create an exception. Only 1 op will support this.

We could generalize this, by doing this change to Op.__call__() instead. @carriepl @lamblin @abergeron what do you think of this solution? This could simplify the most frequent case of adding Rebroadcast op in the graph. We could also extend this to allow the broadcastable parameter to be a tuple, the tuple with the pattern to use.

lamblin:
I would prefer for the `unbroadcast` to be explicitly in the graph, rather than have a variant of the Op itself. Is there a way of returning `unbroadcast(...)` from `make_node`?
In scan, what is the dimension that you do not wish to be broadcastable? Is it the time (number of steps), or another one?

kyunghyuncho:
I'm okay with using tensor.unbroadcast explicitly. When we pass outputs_info to scan, there should essentially be no broadcastable dimension.

nouiz:
I think scan should do it for you. Can you make a ticket that describe this?

kyunghyuncho:
Related issue #2985

PR #2925, opened 2015-05-19
Scan gpuarray memory usage

carriepl:
Without this modifications to traverse(), some HostFromGpu and GpuFromHost are introduced in the graph much too early in the optimization process and it prevents some Scan optimizations from doing their work.


carriepl:
Precision; this happens under the gpuarray backend. Under the cuda backend, all is fine. This is why traverse is modified to also handle the gpuarray backend.

PR #2929, opened 2015-05-20
Added documentation to explain that mode=DebugMode is only meant for CPU jobs.

julianser:
@lamblin .


lamblin:
I think there was a misunderstanding. DEBUG_MODE can be used with gpu, it is actually a good way to check if a GPU implementation is consistent with the CPU reference.
FAST_COMPILE, on the other hand, disables the generation of CUDA code, and makes it almost impossible to use with the GPU. However, using "optimizer=fast_compile" will run on GPU, but without some optimizations.

nouiz:
I agree with @lamblin. I think the doc update should be with mode=FAST_COMPILE.

nouiz:
@julianser will you do the fix shortly? Otherwise I'll close this PR.

thanks

PR #2930, opened 2015-05-20
Add opt for gpuarray.GpuElemwise so exponent of pow has same dtype as output

carriepl:
In the gpuarray backend, pow can only be done on inputs that are both float (producing a float output) or both double (producing a double output). Currently, some unit tests for Scan attempt to do `a_float ** an_int16` which raises a compilation error. This optimization adds  a cast to ensure that the exponent has the same dtype as the output to prevent this.

### DIFF
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -14,7 +14,7 @@ from theano.gof import (local_optimizer, EquilibriumDB,
                         SequenceDB, Optimizer, toolbox)
 from theano.gof.optdb import LocalGroupDB
 
-from theano.scalar.basic import Scalar, Pow, Cast, upcast
+from theano.scalar.basic import Scalar, Pow, Cast
 from theano.scan_module import scan_utils, scan_op, scan_opt
 
 from theano.tensor.nnet.conv import ConvOp
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -265,8 +265,8 @@ def local_gpu_elemwise(node):
         name = 'Gpu' + name
 
     res = GpuElemwise(scal_op, name=name,
-                          inplace_pattern=copy.copy(op.inplace_pattern),
-                          nfunc_spec=op.nfunc_spec)
+                      inplace_pattern=copy.copy(op.inplace_pattern),
+                      nfunc_spec=op.nfunc_spec)
 
     # If the elemwise operation is a pow, casts might be required on the
     # inputs and or outputs because only the (float, float)->float and
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -274,6 +274,12 @@ def local_gpu_elemwise(node):
     if isinstance(op.scalar_op, Pow):
         old_out_dtype = node.outputs[0].dtype
         old_inp_dtypes = [inp.dtype for inp in node.inputs]
+
+        # Upcast the input dtypes with 'float32' to obtain a floating-point
+        # dtype in which to do the computation.
+        # TODO : Currently, a bug in GpuElemwise prevents support for float16.
+        # It should be fixed and then the upcast below can use 'float16'
+        # instead of 'float32'
         new_out_dtype = upcast("float32", *old_inp_dtypes)
 
         # Transfer the inputs on the GPU and cast them to the right dtype
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -21,7 +21,8 @@ from theano.tensor.nnet.conv import ConvOp
 from theano.tests.breakpoint import PdbBreakpoint
 
 from .type import GpuArrayType, GpuArrayConstant
-from .basic_ops import (host_from_gpu, gpu_from_host,
+from .basic_ops import (as_gpuarray_variable,
+                        host_from_gpu, gpu_from_host,
                         HostFromGpu, GpuFromHost,
                         GpuSplit, GpuContiguous,
                         gpu_alloc, GpuAlloc, GpuReshape,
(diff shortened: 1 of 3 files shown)


theano/sandbox/gpuarray/opt.py
```diff
@@ -260,6 +261,25 @@ def local_gpu_elemwise(node):
     return res
 
 
+@register_opt()
+@gof.local_optimizer([GpuElemwise])
+def cast_elemwise_pow_exponent(node):
```
nouiz: Why this is needed? This also need a test.

carriepl: I just updated the description. I just wanted to create the PR as soon as possible so Travis would start faster. Let me know if anything is still needed

abergeron: It would still need a test.

carriepl: I added tests. Now, if the base is float32 or float64, the exponent can have any uint, int or float dtype and it works. However, there are still cases where it doesn't work. If both the base and the exponent are ints, for instance, the output dtype will also be int and it will not fit in the two patterns that the gpu implementation supports at this point in time (float32,float32)->(float32) and (float64,float64)->(float64).

I am not sure which way would be best to handle this.
We could add casts before the inputs and after the output when moving the Elemwise(pow) to the Gpu but that means in some cases we would have to downcast the output (say from float32 to int8). 
We could also not move the Elemwise(pow) nodes with unsupported dtypes to gpu but if the goal of the new backend was to support other dtypes than floating points, then that's probably not the best solution.
Any opinion on this?

carriepl: @lamblin, what do you think?

lamblin:
Summary of IRL discussion:
- Both numpy and the current C code for `power` seem to cast their arguments to float or double, then do the exponentiation and cast back, all of that implicitly. One way of exposing that is to compute `(-1) ** x` for a large odd `x` (for instance, `2**60 - 1`)
- So, explicitly doing these conversion on the GPU makes sense. We could do that by changing the C code, or adding casts in the graph, I think adding casts in the graph would be easier.
- If we want to make things better for integer-valued exponentiation, I think we should not use the `power` op when `**` is called on integers, but have a different Op for that. For instance, Python gives the right result if both sides of the exponent are integers, but rounds things if one of the arguments is a float:
```python
>>> (-1) ** (2**60 + 1)
-1
>>> (-1.) ** (2**60 + 1)
1.0
```

carriepl:
Following the discussion with @lamblin, I implemented the solution in which casts are added to the graph, if needed. I'm not 100% certain the way I did the casts on the GPU is the simplest/cleanest way  so if there's anything to improve there, don't hesitate.

theano/sandbox/gpuarray/opt.py
```diff
@@ -254,10 +255,43 @@ def local_gpu_elemwise(node):
     name = op.name
     if name:
         name = 'Gpu'+name
+
     res = GpuElemwise(scal_op, name=name,
-                      inplace_pattern=copy.copy(op.inplace_pattern),
-                      nfunc_spec=op.nfunc_spec)
-    return res
+                          inplace_pattern=copy.copy(op.inplace_pattern),
+                          nfunc_spec=op.nfunc_spec)
```
lamblin: Strange indentation here

theano/sandbox/gpuarray/opt.py
```diff
@@ -254,10 +255,43 @@ def local_gpu_elemwise(node):
...
+
+    # If the elemwise operation is a pow, casts might be required on the
+    # inputs and or outputs because only the (float, float)->float and
+    # (double, double)->double cases are implemented at the moment.
+    if isinstance(op.scalar_op, Pow):
+        old_out_dtype = node.outputs[0].dtype
+        old_inp_dtypes = [inp.dtype for inp in node.inputs]
+        new_out_dtype = upcast("float32", *old_inp_dtypes)
```
lamblin: It makes sense to me, I'm just wondering if we need a special case for float16.

carriepl: I'm not sure. @abergeron, what do you think?

theano/sandbox/gpuarray/opt.py
carriepl: Done.

theano/sandbox/gpuarray/opt.py
abergeron: I think you should just upcast all the argument to `upcast(*old_inp_dtypes)` and not include a "float32".

This way it should work for float16.

carriepl: The upcast was meant to ensure that the resulting dtype would be floating point and not an integer (which is not currently supported). Can't I simply upcast with 'float16' instead of 'float32'?

abergeron: Yes that would work.

theano/sandbox/gpuarray/tests/test_elemwise.py
```diff
@@ -46,6 +46,26 @@ def test_c_inplace(self):
...
+def test_elemwise_pow():
+    # Test that GpuElemwise(pow) can compile with any combination of integer
+    # or float input dtype.
+    dev = theano.sandbox.gpuarray.init_dev.device
+    if not dev.startswith('cuda'):
+        raise SkipTest("Cuda specific tests")
+
+    dtypes = ["uint8", "uint16", "uint32", "uint64",
+              "int8", "int16", "int32", "int64",
+              "float32", "float64"]
```
abergeron: Perhaps you can add "float16" to that list.

theano/sandbox/gpuarray/opt.py
lamblin: Or maybe use floatX, unless floatX is float64, in which case use float32.
I'm afraid that int8 ** int8 gets casted to float16 even in cases where float16 is not supported.

abergeron: float16 is always supported for Elemwise operations on the GPU.  It might not be optimal, but it will work.  And since we cast the result back to the previous type, it won't make much of a difference.

However using floatX is fine for me.

theano/sandbox/gpuarray/tests/test_elemwise.py
carriepl: Adding "float16" makes compilation crash without the kernels for float16 installed (the same error as with integer types before : "calling std::pow from a kernel function is forbidden and you shouldn't") so I'm not certain it's a good idea.

theano/sandbox/gpuarray/opt.py
carriepl: It doesn't appear to be always supported. Without the float16 kernels installed (since they're still an optional dependency), the compilation raises an error (described in the other comment thread below).

nouiz:
This need a rebase.

carriepl:
Added 'float16' to the list of dtypes to test and did the requested rebase.
Upcasting with 'float16' instead of 'float32' cannot be done for now because of a bug identified in GpuElemwise by @abergeron. I added a comment to document this.

abergeron:
This is ok (after travis passes).  I've made https://github.com/Theano/Theano/issues/3138 to keep track of the GpuElemwise bug.

theano/sandbox/gpuarray/opt.py
```diff
@@ -262,10 +263,49 @@ def local_gpu_elemwise(node):
...
+
+        # Transfer the inputs on the GPU and cast them to the right dtype
+        new_inputs = []
+        for inp in node.inputs:
+            if inp.dtype != new_out_dtype:
+                gpu_cast_op = GpuElemwise(Cast(Scalar(new_out_dtype)))
+                new_inputs.append(gpu_cast_op(gpu_from_host(inp)))
+            else:
+                new_inputs.append(gpu_from_host(inp))
```
nouiz: use as_gpuarray_variable not, gpu_from_host.

theano/sandbox/gpuarray/tests/test_elemwise.py
```diff
@@ -46,6 +46,26 @@ def test_c_inplace(self):
...
+              "int8", "int16", "int32", "int64",
+              "float16", "float32", "float64"]
+
+    for dtype_base in dtypes:
+        for dtype_exp in dtypes:
+            # Compile a gpu function with the specified dtypes
+            base = theano.tensor.vector(dtype=dtype_base)
+            exp = theano.tensor.vector(dtype=dtype_exp)
+            output = base ** exp
+            f = theano.function([base, exp], output)
```
nouiz: use the function to make sure you have the right answer.

carriepl:
Done. Should be all fine now.

PR #2932, opened 2015-05-20
c code for GpuAdvancedSubtensor1

nouiz:


nouiz:
It gave me a 1.87x speed up for that op in one case.

abergeron:
Also, this would benefit from a test.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -2614,6 +2618,52 @@ def perform(self, node, inp, out_):
...
+        out_ = (PyObject *) %(out)s;
+        if (out_ == NULL)
+            out_ = Py_None;
+        else
+            Py_INCREF(out_);
+        PyTuple_SetItem(args, 0, (PyObject *) %(idx)s);
+        PyTuple_SetItem(args, 1, zero);
+        PyTuple_SetItem(args, 2, out_);
+        PyTuple_SetItem(args, 3, raise);
+        PyTuple_SetItem(args, 4, max);
```
nouiz: There is already tests for that inside sandbox/cuda/test_basic_ops.py:T_subtensor

I added check and fixed a segfault (I forgot to incref idx before setitem)

nouiz:
I updated the PR.

caglar:
That is a really good news, because we used that kind of advanced indexing everywhere in language modelling and NLP. Specifically to get compute the cost at the output embedding.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -2614,6 +2618,75 @@ def perform(self, node, inp, out_):
...
+        if(PyTuple_SetItem(args, 3, raise)){
+            Py_DECREF(%(idx)s); Py_DECREF(zero); Py_DECREF(out_);
+            Py_DECREF(raise); Py_DECREF(max); Py_DECREF(out_);
+            %(fail)s;
+        }
+        if(PyTuple_SetItem(args, 4, max)){
+            Py_DECREF(%(idx)s); Py_DECREF(zero); Py_DECREF(out_);
+            Py_DECREF(raise); Py_DECREF(max); Py_DECREF(out_);
+            %(fail)s;
+        }
```
abergeron: I think this code would be clearer (and shorter) if you used [Py_BuildValue()](https://docs.python.org/2/c-api/arg.html#c.Py_BuildValue) like:
```
Py_BuildValue("OiOsi", %(idx)s, 0, %(out)s == NULL ? Py_None, %(out)s, "raise", max_threads);
```

nouiz: done

PR #2942, opened 2015-05-21
Add doc on optimizing scan's performance

carriepl:


doc/library/scan.txt
```diff
@@ -454,6 +457,73 @@ As a rule, scan always expects the condition to be the last thing returned
...
+
+Deactivating the garbage collection for Scan can allow it to reuse memory
+between executions instead of always having to allocate new memory. This can
+improve performance at the cost of increased memory usage.
+
+There are two ways to achieve this, using the Theano flag
+``config.scan.allow_gc`` and setting it to False, or using the argument
+``allow_gc`` of the function theano.scan() and set it to False (when a value
+is not provided for this argument, the value of the flag
+``config.scan.allow_gc`` is used).
```
nouiz: Tell that by default, we do allow_gc between iteration, but free the memory after the last iteration.

doc/library/scan.txt
```diff
@@ -454,6 +457,73 @@ As a rule, scan always expects the condition to be the last thing returned
...
+
+
+Graph optimizations
+^^^^^^^^^^^^^^^^^^^
+
+This one is simple but still worth pointing out. Theano is able to
+automatically recognize and optimize many computation patterns. However, it
+doesn't catch every case that could be optimized and it remains useful for
```
nouiz: I think it is useful that some optimization can't be done by Theano. Even if we would like, we would need to change the user interface. This is the case of the example bellow.

carriepl:
Done

carriepl:
@lamblin : @nouiz mentionned it would be good if you could have a look at this since you have prior exposure to the strict mode, etc.

doc/library/scan.txt
```diff
@@ -454,6 +457,78 @@ As a rule, scan always expects the condition to be the last thing returned
...
+
+
+Graph optimizations
+^^^^^^^^^^^^^^^^^^^
+
+This one is simple but still worth pointing out. Theano is able to
+automatically recognize and optimize many computation patterns. However, there
+are patterns that Theano doesn't optimize because doing so would change the
+user interface (such as merging shared variables together into a single one,
+for instance). Additionnaly, Theano doesn't catch every case that it could
```
lamblin: "Additionally"

doc/library/scan.txt
```diff
@@ -454,6 +457,78 @@ As a rule, scan always expects the condition to be the last thing returned
...
+The `LSTM tutorial <http://deeplearning.net/tutorial/lstm.html>`_ on
+`DeepLearning.net <http://deeplearning.net>`_ provides an example of an
+optimization that Theano cannot perform. Instead of performing many matrix
+multiplications between matrix :math:`x_t` and each of the shared matrices
+:math:`W_i`, :math:`W_c`, :math:`W_f` and :math:`W_o`, the matrices
+:math:`W_*`, are merged into a single shared matrix :math:`W` and the graph
+performs a single larger matrix multiplication between :math:`W` and
+:math:`x_t`. The resulting matrix is then sliced to obtain the results of that
+the small individual matrix multiplications would have produced. This
+optimization replaces many small and inefficient matrix multiplications but a
```
lamblin: I would say "replaces several" rather than "many" (4 is not that many), and "multiplications by a".

doc/library/scan.txt
```diff
@@ -454,6 +457,78 @@ As a rule, scan always expects the condition to be the last thing returned
...
+reduce the overhead introduces by using Scan.
+
+
+Passing shared variables as non-sequences
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is possible, inside of Scan, to use shared variables previously defined
+outside of the Scan without explicitly passing them as inputs to the Scan.
+However it is often more efficient to explicitly pass them as non-sequence
+inputs instead. Section :ref:`lib_scan_shared_variables` provides an
```
lamblin: Actually, passing the shared variables themselves as non-sequences will not help performance.
What can help is if a variable depending only on shared variables (and constants) is explicitly passed as a non-sequence, instead of its shared ancestors being (implicitly) the non-sequences.

carriepl: What about the state of a RNG?

carriepl: At any rate, you're right that this should be changed.

lamblin: Right, if it is a variable representing a random sample, or a shared variable with a `default_update` expression, then passing it explicitly as a non-sequence could help, otherwise that variable would be considered like an initial state, and the update expression be executed at each time step.

lamblin:
Works for me, thanks!

nouiz:
@lamblin is there a reason you didn't merged? If you are good with this, I think we can merge.

Issue #2779, opened 2015-04-16
CPU contiguous Op


lamblin:
In some rare cases, we may want to be able to force a copy into a C-contiguous memory layout even for CPU tensors.
In particular, it can be useful for performance if some Ops are slower when reading non-contiguous memory, and are not using a smart caching system.
In particular, for Elemwise, see https://groups.google.com/d/topic/theano-users/64GSAsbhqnk/discussion

lamblin:
Done in #2949.

PR #2949, opened 2015-05-22
Cpu Contiguous

kelvinxu:
Can you let me know if I am on the right track? @abergeron I wrote the following based upon my memory of our conversation.

Kelvin

### DIFF
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -11,6 +11,56 @@ from theano.gradient import DisconnectedType
 tensor = basic
 
 
+class CpuContiguous(theano.Op):
+    """
+    Check to see if the input is c-contiguous,
+    if it is, do nothing, else return a contiguous array
+    """
+    __props__ = ()
+    view_map = {0: [0]}
+    
+    def make_node(self, x):
+        x_ = theano.tensor.as_tensor_variable(x)
+        return theano.Apply(self, [x_], [x_.type()])
+    
+    def perform(self, node, inputs, output_storage):
+        x, = inputs
+        y = output_storage[0]        
+        # if the ouput is contiguous do nothing, else copy
+        # the input
+        if not x.flags['C_CONTIGUOUS']:
+            x = x.copy()
+        assert x.flags['C_CONTIGUOUS']
+        y[0] = x 
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        code = """
+            if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
+                // check to see if output is contiguous first 
+                if (%(y)s != NULL && PyArray_CHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS)){
+                    PyArray_CopyInto(%(y)s, %(x)s);
+                }
+                else{
+                    Py_XDECREF(%(y)s);
+                    Py_XINCREF(%(x)s);
+                    %(y)s = PyArray_GETCONTIGUOUS(%(x)s);
+                }
+            }
+            else{
+                Py_XINCREF(%(x)s);
+                Py_XDECREF(%(y)s);
+                %(y)s = %(x)s;
+            }
+            """ % locals()
+        return code
+         
+    def c_code_cache_version(self):
+        return (0,)
+
+cpu_contiguous = CpuContiguous()
+
 class CumsumOp(theano.Op):
     # See function cumsum for docstring
     def __init__(self, axis=None):
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -44,7 +44,6 @@ class CpuContiguous(theano.Op):
                 }
                 else{
                     Py_XDECREF(%(y)s);
-                    Py_XINCREF(%(x)s);
                     %(y)s = PyArray_GETCONTIGUOUS(%(x)s);
                 }
             }
--- a/theano/tensor/tests/test_extra_ops.py
+++ b/theano/tensor/tests/test_extra_ops.py
@@ -7,9 +7,9 @@ import theano
 from theano.tests import unittest_tools as utt
 
 from theano.tensor.extra_ops import (CumsumOp, cumsum, CumprodOp, cumprod,
-                                     BinCountOp, bincount, DiffOp, diff,
-                                     squeeze, compress, RepeatOp, repeat,
-                                     Bartlett, bartlett,
+                                     CpuContiguous, cpu_contiguous, BinCountOp,
+                                     bincount, DiffOp, diff, squeeze, compress,
+                                     RepeatOp, repeat, Bartlett, bartlett,
                                      FillDiagonal, fill_diagonal,
                                      FillDiagonalOffset, fill_diagonal_offset,
                                      to_one_hot)
@@ -19,6 +19,18 @@ from theano import config, tensor, function
 
 numpy_ver = [int(n) for n in numpy.__version__.split('.')[:2]]
 numpy_16 = bool(numpy_ver >= [1, 6])
+
+
+def test_cpu_contiguous():
+    a = T.fmatrix('a')
+    i = T.iscalar('i')
+    a_val = numpy.asarray(numpy.random.rand(4, 5), dtype='float32')
+    f = theano.function([a, i], cpu_contiguous(a.reshape((5,4))[::i]))
+    topo = f.maker.fgraph.toposort()
+    assert any([isinstance(node.op, CpuContiguous) for node in topo])
+    assert f(a_val, 1).flags['C_CONTIGUOUS']
+    assert f(a_val, 2).flags['C_CONTIGUOUS']
+    assert f(a_val, 3).flags['C_CONTIGUOUS']
 
 
 class TestCumsumOp(utt.InferShapeTester):


theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,43 @@
 tensor = basic
 
 
+class cpu_contiguous(theano.Op):
```
nouiz: Camel case name.

Remove eq and hash methods. Add the `__props__ = ()` attribute.

Set an attribute destroy_map = {0:0}

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,43 @@
...
+        assert x.is_c_contiguous()
+        y = x 
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        # check if output is continguous, if it is copy it in.
+        code = """
+            if (PyArrayCHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS){
+                PyArray_CopyInto(%(y)s, %(x)s)
```
nouiz: If the input is contiguous, do not copy it. The output isn't preallocated by default.

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,43 @@
...
+            }
+            else{
+            %(x)s = PyArray_GETCONTIGUOUS(%(x)s)
```
nouiz: Never change the inputs directly.

Do:
```
Py_DECREF(%(y)s);
%(y)s = PyArray_GETCONTIGUOUS(%(x)s)
```

nouiz:
this also need tests.

nouiz:
Why this op is needed? It is a good start.

abergeron:
This was a CCW request. (#2779)

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,43 @@
...
+    def __hash__(self, other):
+        return hash(type(self))
+    
+    def make_node(self, x):
+        x_ = theano.tensor.as_tensor_variable(x)
+        return theano.Apply(self, [x_], [x_.type()])
+    
+    def perform(self, node, inputs, output_storage):
+        x, = inputs
+        y = output_storage[0][0]
```
abergeron: This won't work.

You would have to say:
```python
y = output_storage[0]
# compute result
y[0] = x
```
Otherwise you will just overwrite the value of the local variable `y` and not change the storage list.

theano/tensor/extra_ops.py
kelvinxu: For my benefit, why do we need to decrement the reference count?

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,45 @@
...
+        if not x.is_c_contiguous():
+            x = x.copy()
+            y[0] = x 
+        assert x.is_c_contiguous()
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        code = """
+            if !PyArrayCHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS{
```
abergeron: I don't think this compiles.  It should be written like this:

```C
if (!PyArrayCHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)) {
```

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,45 @@
...
+                // check to see if output is contiguous first 
+                if (PyArrayCHKFLAGS(%(y)s), NPY_ARRAY_C_CONTIGUOUS){
```
abergeron: This also has parenthesis problems.

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,45 @@
...
+                    PyArray_CopyInto(%(y)s, %(x)s);
+                }
+                else{
+                    Py_XDECREF(%(y)s);
+                    %(y)s = PyArray_GETCONTIGUOUS(%(x)s);
+                }
+            }
+            """ % locals()
+        return code
+
```
abergeron: You might want to add a c_code_cache_version() too.

abergeron:
This would also still need a basic test.  You can probably look at the tests for GpuContiguous for inspiration.

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,47 @@
...
+            y[0] = x 
+        assert x.is_c_contiguous()
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        code = """
+            if (!PyArrayCHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
+                // check to see if output is contiguous first 
+                if (PyArrayCHKFLAGS(%(y)s), NPY_ARRAY_C_CONTIGUOUS){
```
abergeron: This line is still bad.  The parenthesis aren't in the right spot.

kelvinxu:
@abergeron, hey arnaud, is there something obviously wrong with the way the op is written :/

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,52 @@
 tensor = basic
 
 
+class CpuContiguous(theano.Op):
+    """
+    Check to see if the input is c-contiguous,
+    if it is, do nothing, else return a contiguous array
+    """
+    __props__ = ()
+
```
abergeron: This needs a view_map that declares the output is a view on the input `{0: [0]}`

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,52 @@
...
+                if (PyArray_CHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS)){
+                    PyArray_CopyInto(%(y)s, %(x)s);
+                }
+                else{
+                    Py_XDECREF(%(y)s);
+                    %(y)s = PyArray_GETCONTIGUOUS(%(x)s);
+                }
+            }
+            else{
+                %(y)s = %(x)s;
```
abergeron: You need to do proper reference counting here.

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,52 @@
...
+        assert x.flags['C_CONTIGUOUS']
+        y[0] = x 
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        code = """
+            if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
+                // check to see if output is contiguous first 
+                if (PyArray_CHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS)){
```
abergeron: Need to make sure that y is not NULL before checking the flags.

abergeron:
What I've indicated could easily be the cause of the crash.

kelvinxu:
Thanks!

theano/tensor/extra_ops.py
kelvinxu: If y is NULL, I'm guessing I need to allocate the memory?

abergeron: Yes.

kelvinxu:
Very oddly, when I added a few of the things your recommended. I received a seg fault. Are you around today/tomorrow?

nouiz:
@abergeron isn't avail this week. I'm busy today and tomorrow is holiday.
So come see me Thursday.

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,56 @@
...
+        assert x.flags['C_CONTIGUOUS']
+        y[0] = x 
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        code = """
+            if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
+                // check to see if output is contiguous first 
+                if ($(y)s != NULL && PyArray_CHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS)){
```
lamblin: It should be `%(y)s`, not `$(y)s`.

kelvinxu:
Let me know if I need to fix anything

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,56 @@
...
+        y, = onames
+        code = """
+            if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
+                // check to see if output is contiguous first 
+                if (%(y)s != NULL && PyArray_CHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS)){
+                    PyArray_CopyInto(%(y)s, %(x)s);
+                }
+                else{
+                    Py_XDECREF(%(y)s);
+                    Py_XINCREF(%(x)s);
```
abergeron: This adds an extra useless reference to `x`.  PyArray_GETCONTIGUOUS already returns a new reference.

kelvinxu:
Fixed

abergeron:
Looks good.

theano/tensor/extra_ops.py
```diff
@@ -11,6 +11,55 @@
...
+        assert x.flags['C_CONTIGUOUS']
+        y[0] = x 
+
+    def c_code(self, node, name, inames, onames, sub): 
+        x, = inames
+        y, = onames
+        code = """
+            if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
+                // check to see if output is contiguous first 
+                if (%(y)s != NULL && PyArray_CHKFLAGS(%(y)s, NPY_ARRAY_C_CONTIGUOUS)){
```
nouiz: you also need to check that they are of the same shape. Otherwise, it won't work. This can happen if the minibatch size change for the user.

lamblin: Right, I did not see that.
@kelvinxu can you make that fix, please? Thanks.

PR #2952, opened 2015-05-23
Flake8 work

abergeron:
I was bored and I did this.  Should not change anything functional*, but it does make test_flake8 a little less useless.

*: although it does fix one bug with detect_nan in compile/monitormode.py


nouiz:
This is good, but a bad timming. There is already a big PR for python3 that do much changes...

What do you think, which PR whould be merged first? Do they conflict?

nouiz:
python3 PR seem much bigger. Can you rebase your PR on it?

nouiz:
I checked the PR. I'm ok with it, but I have one comment on the code.

The only problems is related to the possible conflict with the python3 PR.

abergeron:
The python 3 PR doesn't work (at least when I tried it).  If it does work and will be merged, I can rebase my work on top of it.  It took only a couple of hours to do.

abergeron:
Also, it seems I used a slightly older version of pep8 so the extra commits fixes the flake8 errors that travis sees.  I'll make sure to use the right version for any future work.

PR #2953, opened 2015-05-23
Small

nouiz:


nouiz:
I'm not sure about the last commit. It would make multi-thread system that use different function on the same GPU troublesome. The missing optimization probably made it work more easily.

@abergeron what do you think? Do you know an easy way from python c api to know the current threads to have one per thread?

abergeron:
Multiple threads on a single GPU is not (and never was) supported with the old backend.  So I don't see anything changing with that.  We don't even support multiple functions on the CPU at the same time.  So I don't think that commit is going to be a problem.

PR #2958, opened 2015-05-26
Change default value of reoptimize_unpickled_function

nouiz:


nouiz:
Wait after NIPS to merge.

carriepl:
After IRL discussion with @nouiz; he says he should revisit this PR before merging so, even if this is after NIPS, this PR shouldn't be merged as it is.

nouiz:
I updated this PR. I'm not sure how to warn user about the interface change. What do you think? I think we can just email again on the mailing list. If you disagree, tell it.

nouiz:
@lamblin @abergeron what do you think of this PR?

lamblin:
OK for me.

PR #2965, opened 2015-05-27
Added small gradient calculation for batched_dot

briancheung:
Wrote a quick gradient for the new batched_dot op just to get the ball rolling. Let me know what else needs to be written.

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -199,6 +199,20 @@ class BatchedDotOp(GpuOp):
             } while (0)
         """
 
+    def grad(self, inp, grads):
+        x, y = inp
+        gz, = grads
+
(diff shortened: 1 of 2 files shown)


nouiz:
thanks. can you add test for the gradient? http://deeplearning.net/software/theano/tutorial/extending_theano.html#testing-the-gradient

briancheung:
Anything else needed for this PR?

nouiz:
Thanks. I did a PR to your PR to fix the test with mode=FAST_COMPILE

https://github.com/briancheung/Theano/pull/1

We run our daily buildbot in that mode. If you want to tests with all modes that our daily buildbot test, you can use theano/misc/do_nightly_build .... instead of nosetests ...

With that, we can merge your PR. thanks

Also, when you update a PR with new commit, we don't get notified. So if you want us to see it faster, just make a comment in the PR. That send us a notification.

nouiz:
thanks

PR #2967, opened 2015-05-27
Proper string representation of a ValueError on Softmax.make_node

dg2:
Very minor change to create an actual string representing the error, instead of a tuple of two strings


nouiz:
thanks

PR #2969, opened 2015-05-27
Function.copy() that shares memory

ChienliMa:
A draft of ```Function.copy()``` method which enable two functions share storages of intermediate computation.Test is not finished.I just decide to create a PR for review, to prevent myself from going to far if I am wrong.

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -517,6 +517,7 @@ returned directly?"""
 
         if not share_memory:
             return self.__copy__()
+
         else:
             # copy SymbolocKits
             ins, outs = copy.deepcopy([self.maker.inputs, self.maker.outputs])
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -501,7 +501,7 @@ returned directly?"""
                 there.data = here.data
         return cpy
 
-    def copy(self, share_memory = False):
+    def copy(self, share_memory=False):
         """
         Copy this function. Copied function will have separated maker and fgraph
         with original function. User can choose whether to separate storage by 
@@ -543,7 +543,8 @@ returned directly?"""
             # But to be safe for now as it isn't documented and we aren't sure 
             # it is well tested, we don't share the part of the storage_map.
             for key in storage_map.keys():
-                if key not in self.maker.fgraph.outputs:
+                if key not in self.maker.fgraph.outputs and \
+                    not isinstance(key, theano.tensor.Constant):
                     new_storage_map[memo[key]] = storage_map[key]
 
             # copy input storages if it's mutable
@@ -556,13 +557,10 @@ returned directly?"""
                 else:
                     input_storage.append( copy.deepcopy[storage])
 
-<<<<<<< HEAD
             new_func = new_maker.create(input_storage, storage_map=new_storage_map)
 
             return new_func
 
-=======
->>>>>>> 87f2056... Delete extra line
     def __call__(self, *args, **kwargs):
         profile = self.profile
         t0 = time.time()
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -545,13 +545,7 @@ returned directly?"""
             # But to be safe for now as it isn't documented and we aren't sure 
             # it is well tested, we don't share the part of the storage_map.
             for key in storage_map.keys():
-<<<<<<< HEAD
-                if key not in self.maker.fgraph.outputs and 
-                    not isinstance(key, theano.tensor.Constant):
-=======
-                # output_storages should not be shared
-                if key not in self.maker.fgraph.outputs and memo.has_key(key):
->>>>>>> 38ae402... Format Change; Modify inproper comment;Change constructor class;Add test for output storage.
+                if key not in self.maker.fgraph.outputs:
                     new_storage_map[memo[key]] = storage_map[key]
 
             # copy input storages if it's mutable
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -545,6 +545,7 @@ returned directly?"""
             # But to be safe for now as it isn't documented and we aren't sure 
             # it is well tested, we don't share the part of the storage_map.
             for key in storage_map.keys():
+<<<<<<< HEAD
                 if key not in self.maker.fgraph.outputs and 
                     not isinstance(key, theano.tensor.Constant):
                     new_storage_map[memo[key]] = storage_map[key]
@@ -560,6 +561,33 @@ returned directly?"""
                     input_storage.append( copy.deepcopy[storage])
 
             new_func = new_maker.create(input_storage, storage_map=new_storage_map)
+=======
+                # output_storages should not be shared
+                # if key not in self.maker.fgraph.outputs and \
+                #     memo.has_key(key):
+                new_storage_map[memo[key]] = storage_map[key]
+
+            # copy input storages if it's mutable
+            input_storage = [] 
+            for i in self.maker.inputs:
+                storage = getattr(i, 'value', None) 
+                if isinstance(i.variable, theano.tensor.Constant) or\
+                    not i.mutable:
+                    input_storage.append(storage )
+                else:
+                    input_storage.append( copy.deepcopy[storage])
+
+            new_func = new_maker.create(input_storage, \
+                                storage_map=new_storage_map)
+
+            # share immutable SharedVariable's storage
+            # for (input, _1, _2), here, there in zip(self.indices,
+            #                                         self.input_storage,
+            #                                         new_func.input_storage):
+            #     if isinstance(i.variable, theano.tensor.Constant) or \
+            #         not input.mutable:
+            #         there.data = here.data
+>>>>>>> 4acf04b... Add some docs to FunctionGraph and minor changes of function_module.copy()
 
             return new_func
 
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -523,6 +523,7 @@ returned directly?"""
 
             # copy fgraph and get memo
             maker = self.maker
+<<<<<<< HEAD
             fg_cpy, memo = maker.fgraph.clone_get_equiv(attach_feature=False)
 
             # use copied ins, outs and fgraph to init a maker
@@ -530,6 +531,13 @@ returned directly?"""
                                         fgraph=fg_cpy, profile=maker.profile,
                                         accept_inplace=maker.accept_inplace,
                                         function_builder=maker.function_builder,
+=======
+            new_maker = FunctionMaker(inputs=ins, outputs=outs, mode=maker.mode,
+                                        fgraph=new_fgraph, profile=maker.profile,
+                                        accept_inplace=maker.accept_inplace,
+                                        function_builder=maker.function_builder,
+                                        on_unused_input=maker.on_unused_input)
+>>>>>>> 6ba7c5d... Small modification to copy(). Testcase is finished. Start debuging.
 
             # construct new storage_map that map new variable to old storage
             # so that the ensuing function shares storage with the original one
@@ -543,6 +551,7 @@ returned directly?"""
             # But to be safe for now as it isn't documented and we aren't sure 
             # it is well tested, we don't share the part of the storage_map.
             for key in storage_map.keys():
+<<<<<<< HEAD
                 if key not in self.maker.fgraph.outputs and 
                     not isinstance(key, theano.tensor.Constant):
                     new_storage_map[memo[key]] = storage_map[key]
@@ -559,6 +568,24 @@ returned directly?"""
 
             new_func = new_maker.create(input_storage, storage_map=new_storage_map)
 
+=======
+                # output_storages should not be shared
+                if key not in self.maker.fgraph.outputs and \
+                    memo.has_key(key):
+                    new_storage_map[memo[key]] = storage_map[key]
+
+            # copy input storages and link function with new storage_map
+            input_storage = copy.copy([getattr(i, 'value', None) for i in ins])
+            new_func = new_maker.create(input_storage, storage_map=new_storage_map)
+
+            # share immutable SharedVariable's storage
+            for (input, _1, _2), here, there in zip(self.indices,
+                                                    self.input_storage,
+                                                    new_func.input_storage):
+                if not input.mutable:
+                    there.data = here.data
+
+>>>>>>> 6ba7c5d... Small modification to copy(). Testcase is finished. Start debuging.
             return new_func
 
     def __call__(self, *args, **kwargs):
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -501,7 +501,7 @@ returned directly?"""
                 there.data = here.data
         return cpy
 
-    def copy(self, share_memory=False):
+    def copy(self, share_memory = False):
         """
         Copy this function. Copied function will have separated maker and fgraph
         with original function. User can choose whether to separate storage by 
@@ -530,15 +530,20 @@ returned directly?"""
                                         fgraph=fg_cpy, profile=maker.profile,
                                         accept_inplace=maker.accept_inplace,
                                         function_builder=maker.function_builder,
-                                        on_unused_input=maker.on_unused_input)
 
             # construct new storage_map that map new variable to old storage
             # so that the ensuing function shares storage with the original one
             new_storage_map = {}
             storage_map = self.fn.storage_map
+            
+            # TODO: We could share the output storage, but we must make sure 
+            # 2 different function call won't override each other values. This 
+            # is already done elsewhere, so to reuse it the user would need to 
+            # use Out(var, borrow=True) and maybe the mutable=True flag too. 
+            # But to be safe for now as it isn't documented and we aren't sure 
+            # it is well tested, we don't share the part of the storage_map.
             for key in storage_map.keys():
-                # output_storages should not be shared
-                if key not in self.maker.fgraph.outputs and memo.has_key(key):
+                if key not in self.maker.fgraph.outputs:
                     new_storage_map[memo[key]] = storage_map[key]
 
             # copy input storages if it's mutable
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -523,7 +523,6 @@ returned directly?"""
 
             # copy fgraph and get memo
             maker = self.maker
-
             fg_cpy, memo = maker.fgraph.clone_get_equiv(attach_feature=False)
 
             # use copied ins, outs and fgraph to init a maker
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -518,11 +518,11 @@ returned directly?"""
         if not share_memory:
             return self.__copy__()
         else:
-            # copy SymbolocKits
-            ins, outs = copy.deepcopy([self.maker.inputs, self.maker.outputs])
+            maker = self.maker
+            # copy Ins, so that they have different storage as their value
+            ins = copy.deepcopy(maker.inputs)
 
             # copy fgraph and get memo
-            maker = self.maker
             fg_cpy, memo = maker.fgraph.clone_get_equiv(attach_feature=False)
 
             # construct new storage_map that map new variable to old storage
@@ -540,9 +540,17 @@ returned directly?"""
                 if key not in maker.fgraph.outputs:
                     new_storage_map[memo[key]] = storage_map[key]
 
-            # copy input storages if it's mutable
+            
             input_storage = []
             for in_ori, in_cpy in zip(maker.inputs, ins):
+                # Since we reuse original Out instances, the copied In
+                # instances should use the original variabls as their variables
+                # and updates. Otherwise the compilation will fail at function
+                # FunctionMaker._check_unused_inputs()
+                in_cpy.variable = in_ori.variable
+                in_cpy.update = in_ori.update
+
+                # share input storages if it's immutable
                 is_const = isinstance(in_ori.variable, theano.tensor.Constant)
                 if is_const or not in_ori.mutable:
                     storage = getattr(in_ori, 'value', None)
@@ -552,7 +560,8 @@ returned directly?"""
                 input_storage.append(storage)
 
             # reinitialize new maker and create new function
-            return maker.__class__(inputs=ins, outputs=outs, fgraph=fg_cpy,
+            return maker.__class__(inputs=ins, outputs=maker.outputs,
+                                   fgraph=fg_cpy,
                                    mode=maker.mode, profile=maker.profile,
                                    on_unused_input=maker.on_unused_input,
                                    function_builder=maker.function_builder,
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -503,8 +503,8 @@ returned directly?"""
     def copy(self, share_memory=False):
         """
         Copy this function. Copied function will have separated maker and
-        fgraph with original function. User can choose whether to separate
-        storage by changing the share_memory arguments
+        fgraph with original function if share_memory=True. User can choose
+        whether to separate storage by changing the share_memory arguments
         ---------------------
         Params:
             share_memory -- { boolean } Default is False. When True, two
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -552,7 +552,7 @@ returned directly?"""
             storage_map = self.fn.storage_map
             for key in storage_map.keys():
                 if not isinstance(key, theano.tensor.Constant) and \
-                    equiv.has_key(key):
+                    memo.has_key(key):
                     new_storage_map[memo[key]] = storage_map[key]
 
             # copy input storages and use new storage_map to link function
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -517,7 +517,6 @@ returned directly?"""
 
         if not share_memory:
             return self.__copy__()
-
         else:
             # copy SymbolocKits
             ins, outs = copy.deepcopy([self.maker.inputs, self.maker.outputs])
@@ -561,7 +560,6 @@ returned directly?"""
             input_storage = copy.copy([getattr(input, 'value', None) for input in ins])
             new_func = new_maker.create( input_storage, storage_map = new_storage_map )
             return new_func
-
 
     def __call__(self, *args, **kwargs):
         profile = self.profile
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -517,7 +517,6 @@ returned directly?"""
         Returns:
             func -- Copied theano.Function
         """
-
         if not share_memory:
             return self.__copy__()
         else:
@@ -528,26 +527,26 @@ returned directly?"""
             # copy fgraph and get memo
             fg_cpy, memo = maker.fgraph.clone_get_equiv(attach_feature=False)
 
-            # construct new storage_map that map new variable to old storage
+            # Construct new storage_map that map new variable to old storage,
             # so that the ensuing function shares storage with the original one
-            new_storage_map = {}
-            storage_map = self.fn.storage_map
-
             # TODO: We could share the output storage, but we must make sure
             # 2 different function call won't override each other values. This
             # is already done elsewhere, so to reuse it the user would need to
             # use Out(var, borrow=True) and maybe the mutable=True flag too.
             # But to be safe for now as it isn't documented and we aren't sure
             # it is well tested, we don't share the part of the storage_map.
+            storage_map = self.fn.storage_map
+            new_storage_map = {}
             for key in storage_map.keys():
                 if key not in maker.fgraph.outputs:
                     new_storage_map[memo[key]] = storage_map[key]
 
             input_storage = []
-            for in_ori, in_cpy in zip(maker.inputs, ins):
-                # Since we reuse original Out instances, the copied In
-                # instances should use the original variabls as their variables
-                # and updates. Otherwise the compilation will fail at function
+            assert len(ins) == len(fg_cpy.inputs)
+            for in_ori, in_cpy, in_v in zip(maker.inputs, ins, fg_cpy.inputs):
+                # Since we reuse original Out instances, copied In instances
+                # should use the original variabls as their variables and
+                # updates. Otherwise the compilation will fail at function
                 # FunctionMaker._check_unused_inputs()
                 in_cpy.variable = in_ori.variable
                 in_cpy.update = in_ori.update
@@ -561,12 +560,10 @@ returned directly?"""
                     storage = getattr(in_cpy, 'value', None)
                 input_storage.append(storage)
 
-            # pop out input_storage in storage_map and only use storage of In
-            # instances to initialize the make, so that we can avoid 
-            # storage conflictions in link.map_storage()
-            assert len(fg_cpy.inputs) == len(input_storage)
-            for in_var in fg_cpy.inputs:
-                new_storage_map.pop(in_var)
+                # pop out input_storage in storage_map and only use storage of
+                # In instances to initialize the make, to avoid storage
+                # conflictions in link.map_storage()
+                new_storage_map.pop(in_v)
 
             # reinitialize new maker and create new function
             return maker.__class__(inputs=ins, outputs=maker.outputs,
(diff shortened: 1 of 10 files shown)


theano/compile/function_module.py
```diff
@@ -483,6 +488,66 @@ def __copy__(self):
                 there.data = here.data
         return cpy
 
+    def copy(self, share_memory = False):
```
nouiz: pep8 tell no space around = in parameter.

theano/compile/function_module.py
```diff
@@ -483,6 +488,66 @@ def __copy__(self):
...
+        """
+
+        if not share_memory:
+            return self.__copy__()
+        else:
+            # copy SymbolocKits
+            ins, outs = copy.deepcopy([self.maker.inputs, self.maker.outputs])
+
+            # get copied input, output variables
+            in_vars = [ i.variabls for i in ins ]
```
nouiz: typo, variabls -> variable?

theano/compile/function_module.py
```diff
@@ -483,6 +488,66 @@ def __copy__(self):
...
+        ---------------------
+        Returns:
+            func -- Copied theano.Function
+        """
+
+        if not share_memory:
+            return self.__copy__()
+        else:
+            # copy SymbolocKits
+            ins, outs = copy.deepcopy([self.maker.inputs, self.maker.outputs])
```
nouiz: Instead of redoing all this, what about changing FunctionGraph.clone_get_equiv() to have a new parameter add_features that default to True and you set it to False? I think it would do what you need.

Another probably better option would be to add an attribute to Feature that need a new instance before being used in another fgraph. I would call it something like reinstanciate_on_reattach=True. You could set it to the DestroyHandler that caused you problems.

theano/compile/function_module.py
```diff
@@ -483,6 +488,66 @@ def __copy__(self):
...
+            new_fgraph = FunctionGraph(in_vars, out_vars)
+
+            # re-initialize new FunctionMaker
+            maker = self.maker
+            kwargs = dict(inputs=ins, outputs=outs, fgraph=new_fgraph,
+                            mode=maker.mode, profile=maker.profile,
+                            accept_inplace=maker.accept_inplace,
+                            function_builder=maker.function_builder,
+                            on_unused_input=maker.on_unused_input )
+            new_maker = FunctionMaker( kwargs )
```
nouiz: Set the kwargs directly as parameter instead of doing it separatly.

theano/compile/function_module.py
```diff
@@ -483,6 +488,66 @@ def __copy__(self):
...
+
+            # construct new storage_map that map new variable to old storage
+            # so that the ensuing function shares storage with the original one
+            new_storage_map = {}
+            storage_map = self.fn.storage_map
+            for key in storage_map.keys():
+                if not isinstance(key, theano.tensor.constant) and \
```
nouiz: type constant -> Constant.

theano/compile/function_module.py
```diff
@@ -483,6 +488,66 @@ def __copy__(self):
...
+                    equiv.has_key(key):
```
nouiz: when equiv could not have the key? If that should not happen, remove that condition.

theano/gof/tests/test_vm.py
```diff
@@ -358,17 +361,24 @@ def test_reallocation():
...
+
+                    for other_key in other_keys:
+                        if (storage_map[key][0] and
+                            storage_map[key][0] is storage_map[other_key][0]):
+                            return [True, storage_map[other_key][0]]
             return [False, None]
 
+        # assert there's same value in strage_map
         assert check_storage(storage_map)[0]
+        # assert there's same stoarge in storage_map
```
nouiz: Don't modify that test. Revert it to the original test. I pointed you to it as an example of how you could build your tests. Add your tests in the new test you added that tell todo.

nouiz:
good start.

theano/compile/function_module.py
ChienliMa: Good idea. But there's another problem that variables in In() instances be the same as thoes in fgraph.   It will be a little hard to rewarp In() instances or modify them with only an fgraph. So I decided to copy In()s first and then use variables inside to copy fgraph. I will try to find if there exist better ways.

theano/gof/tests/test_vm.py
ChienliMa: Sure.

theano/compile/function_module.py
nouiz: I don't understand why creating new In() and Out() after having cloned the
fgraph to be difficult. I can just call copy.copy() on it and swap the
variable.

ChienliMa: I am not sure about this. Initially I thought that variables in ```In()``` instances share same references with those in fgraph. However, it seem that it's not the case.
```
In [37]: func = theano.function([x],[x+1])
In [38]: ins = func.maker.inputs
In [39]: for i in ins:
    print i.variable in func.maker.fgraph.variables
   ....:     
False
```
The main problem is that matching two variable isn't easy, as now variables in ```In()```s aren't those in ```fgraph```, so memo mapping can't work. What's worse, as expression```new_var == old_var``` always return ```False``` even two var instances represent same things(at lease this happened in my experiments), I can hardly match variable in ```In()```s with those in fgraph. Also, after fgraph is copied, the order of variable is not guaranteed(right?), so I cannot loop over pairs. Apart from this, matching updates will also become a problem.  

All these problems occur only when variable in ```In()```s and fgraph need to share same reference, otherwise it will be easy. I don't even need to copy a variable.  

---------------------------------

Apart from this, I would like to know how ```In()``` affect the evaluation of function, like ```update```. Because I cannot find any relevant codes.

ChienliMa: Well, I come out with a new approach and I am working on it. But I still don't understand why variables in ```In()``` don't refer to same object to those in fgraph.  Does it matters?. Also I don't know how the update work -- I could not find the codes. I would like to know Any hints?

nouiz: In() containt reference to the variable used before creating the function
graph. When we create the fgraph, we clone the user graph as we do not want
to modify the user graph itself. But we never up to now made the In() point
to the cloned var. So you do not need to do this.

To know the mapping between maker.inputs and fgraph.inputs, the order is
the same. Both are lists.

There is the notion of extended inputs and extended outputs. They come by
appending the shared var used to the inputs and the update to the outputs.

The function graph only see the extended inputs and outputs. FunctionMaker
also only see the extended inputs and outputs.

When you call theano.function() if you use shared variable, this will call
pfunc(). It is him that will create the extended inputs list and call
orig_function() with it. orig_function, call std_fgraph() and it is him
that create the extended outputs that are passed to the function graph.

So f.maker.inputs contain the extended inputs, but not the extended outputs.

f.maker.fgraph.inputs/outputs contain a cloned copy of the extended inputs
and outputs.

Yes this is ugly. I needed to dig in the code to tell you this. A clean up
of that would be good, but not for now. Can you take notes of what would
need clean up and what could be done?

ChienliMa: Many thanks :)

I used to believe that the value of SharedVariable was updated by storage_map and ```In()```s, so I was struggling about this. Now I know it's not the case. I got the idea of how this should be done properly( I hope ) as I'll update the code tomorrow morning.

And, I'll take notes about possible improvement from now on.

ChienliMa:
Finished and test for "PerformLinker" pass. Please review my test case.
I should have push this earlier, but there's bug in ```FunctionGraph.clone_get_equiv()``` getting in my way, so I fix it by the way.

If you think this is OK, I'll add argument ```storage_map``` to other linkers. Just a small modification so that other test  can pass.

theano/compile/function_module.py
```diff
@@ -483,6 +488,70 @@ def __copy__(self):
...
+        else:
+            # copy SymbolocKits
+            ins, outs = copy.deepcopy([self.maker.inputs, self.maker.outputs])
+
+            # copy fgraph and get memo
+            maker = self.maker
+            fg_cpy, memo = maker.fgraph.clone_get_equiv(attach_feature=False)
+
+            # use copied ins, outs and fgraph to init a maker
+            new_maker = FunctionMaker(inputs=ins, outputs=outs, mode=maker.mode,
```
nouiz: use `maker.__class__` instead of FunctionMaker. DebugMode use a special FunctionMaker.

theano/compile/function_module.py
```diff
@@ -483,6 +488,70 @@ def __copy__(self):
...
+            input_storage = [] 
+            for i in self.maker.inputs:
+                storage = getattr(i, 'value', None) 
+                if isinstance(i.variable, theano.tensor.Constant) or\
+                    not i.mutable:
+                    input_storage.append(storage )
+                else:
+                    input_storage.append( copy.deepcopy[storage])
+
+            new_func = new_maker.create(input_storage, \
```
nouiz: pep8 tell that we should not use \ when possible. Here it is useless.

theano/compile/function_module.py
```diff
@@ -483,6 +488,70 @@ def __copy__(self):
...
+                                        on_unused_input=maker.on_unused_input)
+
+            # construct new storage_map that map new variable to old storage
+            # so that the ensuing function shares storage with the original one
+            new_storage_map = {}
+            storage_map = self.fn.storage_map
+            for key in storage_map.keys():
+                # output_storages should not be shared
+                # if key not in self.maker.fgraph.outputs and \
+                #     memo.has_key(key):
```
nouiz: remove this useless code and comments. Otherwise, tell why you think the comment is true.

theano/compile/function_module.py
```diff
@@ -1250,7 +1319,7 @@ def _check_unused_inputs(self, inputs, outputs, on_unused_input):
                         "valid values are 'raise', 'warn', and 'ignore'."
                         % on_unused_input))
 
-    def create(self, input_storage=None, trustme=False):
+    def create(self, input_storage=None, trustme=False, storage_map = None):
```
nouiz: pep8, no space around = in function signature and function call.

theano/compile/function_module.py
```diff
@@ -1323,7 +1392,7 @@ def create(self, input_storage=None, trustme=False):
         try:
             theano.config.traceback.limit = 0
             _fn, _i, _o = self.linker.make_thunk(
-                input_storage=input_storage_lists)
+                input_storage=input_storage_lists, storage_map = storage_map)
```
nouiz: idem

theano/compile/tests/test_function_module.py
```diff
@@ -239,6 +239,37 @@ def test_copy(self):
...
+        storage_map_ori = ori.fn.storage_map
+        storage_map_cpy = cpy.fn.storage_map
+        fgraph_ori = ori.maker.fgraph
+        fgraph_cpy = cpy.maker.fgraph
+
+        # assert intermediate and Constants storages are shared
+        i_o_variables = fgraph_cpy.inputs + fgraph_cpy.outputs
+        ori_storages = storage_map_ori.values()
+        for key in storage_map_cpy.keys():
+            if key not in i_o_variables or isinstance(key, theano.tensor.Constant):
```
nouiz: I think that inputs and outputs should have different storage except for Constant and SharedVariable. Why do you think otherwise?

theano/compile/tests/test_function_module.py
```diff
@@ -239,6 +239,37 @@ def test_copy(self):
...
+
+        # assert storages of SharedVariable without updates are shared
+        for (input, _1, _2), here, there in zip(ori.indices,
+                                                ori.input_storage,
+                                                cpy.input_storage):
+            if not input.mutable:
+                self.assertTrue(here.data is there.data)
+            else:
+                self.assertFalse(here.data is there.data)
+
```
nouiz: assert that variable that should not share storage do not share storage.

theano/compile/tests/test_function_module.py
ChienliMa: I think so. In line 526~529 in function_module.py  I share storage except for output storage.
In next for loop I share storage only when it's ```Constant``` or immutable. 
In this test, I first test that intermediate storage should be shared( line 259~260 ). Next I test if I deal with input storage properly. It seems that I forgot to test output storage and ```Constant```. I'll add corresponding test.

theano/compile/function_module.py
ChienliMa: Sorry have this. I forgot to remove useless codes and comment before commiting. I try to avoid this next time.

nouiz:
This PR need a rebase. This prevent travis-ci to run.

theano/gof/link.py
```diff
@@ -422,7 +422,7 @@ def __deepcopy__(self, memo):
         return r
 
 
-def map_storage(fgraph, order, input_storage, output_storage):
+def map_storage(fgraph, order, input_storage, output_storage, storage_map = None):
```
nouiz: pep8, no space around = in function signature.

ChienliMa:
This is the first time I use ```git rebase```. Well, so many duplicated commits...:(

theano/compile/function_module.py
```diff
@@ -483,6 +488,60 @@ def __copy__(self):
...
+                                        function_builder=maker.function_builder,
+                                        on_unused_input=maker.on_unused_input)
+
+            # construct new storage_map that map new variable to old storage
+            # so that the ensuing function shares storage with the original one
+            new_storage_map = {}
+            storage_map = self.fn.storage_map
+            for key in storage_map.keys():
+                # output_storages should not be shared
+                if key not in self.maker.fgraph.outputs and memo.has_key(key):
```
nouiz: Why memo would not have key in it? When this happen? I think it should never happen. If it is the case, remove the corresponding condition. Otherwise, put a comment that tell why this happen.

I don't understand why we should not share the output storage. If Theano gc is enabled, it will be emptied so we don't care (and this share_memory would be useless anyway). If Theano gc is disabled, sometimes we would like to reuse it. But currently, there is another mechanism that will prevent that, even without this.

But I think for now, it is safer to skip it like you do. But can add a comment like this, so that we have the information for later:

TODO: We could share the output storage, but we must make sure 2 different function call won't override each other values. This is already done elsewhere, so to reuse it the user would need to use Out(var, borrow=True) and maybe the mutable=True flag too. But to be safe for now as it isn't documented and we aren't sure it is well tested, we don't share the part of the storage_map.

theano/compile/function_module.py
nouiz: About ins/outs, now they are used only on line 516. I do not think there copy is still needed as they aren't shared with the Function graph. Can you try without the copy? If that work, just put a comment that it is fine that they are shared between Maker.

nouiz:
A rebase normaly shouldn't add extra commit. Try this procedure:

git checkout YOUR_BRANCH
git checkout -b YOUR_BRANCH_rb
git remote update central
git rebase -i central/master

this should remove the duplicate commits.

After that, check travis-ci failure.

ChienliMa:
Actually I agree with you. But I am not quite sure about whether such case exists. If it does, this will raise an exception and break the program. So I can add either  a ```try...except...``` or a condition. I choose the later one.

For the output storage, I separate it for two reasons. Firstly, I guess someone would like to reuse it, and I am not aware of GC. Secondly, later on we will need to modify updates and output.  Separate the storage might make coding easier. Although probably not ...

As simple this modification is, we can come back and share the output storage later on if we think this necessary.

nouiz:
For the output, can you just add my comments? Otherwise people will wonder
why it is like this if they read the code. I'm ok to let it like this for
now.

For the condition, if it should not happen, don't make special code for
this. If it happen, I want it to crash, as this should never happen. So if
this happen, it mean there is a big problem and working around it isn't
good I think. The user won't have what it asked.

ChienliMa:
OK. 
Now I've rebased my brach, and I have 3 branches 
```
Theano ChienliMa$ git branch
  Copy_shares_memory
* Copy_shares_memory_rb
  master
```
What should I do?
It's too late now. I leave some work to tomorrow.

nouiz:
You can push your local branch to your PR:

git push -f origin Copy_shares_memory_rb:Copy_shares_memory

The name before : is the local name, the one after is the remote name. The -f is needed as you need to force the push as this change the history.

ChienliMa:
I add some code to ```CLinker```, I don't know whether I understand it correctly. Please have a look.

theano/compile/function_module.py
ChienliMa: ```In.value```, ```storage_map[input_var]``` and storage in thunks shared the same reference of input storages. Therefore I think they should be copied. And the values should refer to same object(shared) if ```In.mutable == False and instance(In.variable, theano.tensor.Constant)```.

There was a bug in my code that I did not share the storage between new ```In``` and old ```In```. Now it's fixed.

nouiz: To my understanding, input_storage and output_storage should be different subsets of storage_map. Is that the case? If not, do you know why?

theano/compile/function_module.py
```diff
@@ -496,6 +500,65 @@ def __copy__(self):
...
+                if key not in maker.fgraph.outputs:
+                    new_storage_map[memo[key]] = storage_map[key]
+
+            # copy input storages if it's mutable
+            input_storage = []
+            for in_ori, in_cpy in zip(maker.inputs, ins):
+                is_const = isinstance(in_ori.variable, theano.tensor.Constant)
+                if is_const or not in_ori.mutable:
+                    storage = getattr(in_ori, 'value', None)
+                    in_cpy.value = in_ori.value
```
nouiz: I agree that if it is a Constant, we should not copy the data. But I think we should modify the class Constant, by adding to it a method `__deepcopy__` that do not copy the data. This would remove that line and would be cleaner. The class Container in the file link.py implement it to work around a numpy problem with deepcopy. Maybe it can help you.

theano/compile/function_module.py
nouiz: I still don't understand why we need to deepcopy maker.inputs and maker.outputs. Can you explain why and put this in a comment?

ChienliMa:
There's a bug in ```___copy__()```, there also is another bug in my code. I don't know whether it's a bug, but it just not work properly. My defence hold tomorrow morning, and I need to prepare for it. I'll try to explain my idea tomorrow.

However here's the case:
```
import theano
import theano.tensor as T
import copy

x = T.scalar()
y = theano.shared(value=12)
f = theano.function([x],[x+y+1],updates={y:y+1},mode='FAST_COMPILE')
f1 = copy.copy(f)

# execute different functions different times
f(1)
f1(1),f1(2)
```
and 
```
In [70]: for i1,i2 in zip(f.maker.inputs, f1.maker.inputs):
   ....:         print "\n %d st inputs" % (i) 
   ....:         print "f's input storage:",f.input_storage[i],"|f.maker.inputs.value:",i1.value
   ....:         print "f1's input storage:",f1.input_storage[i],"|f1.maker.inputs.value:",i2.value
   ....:         i+=1
   ....:     

 0 st inputs
f's input storage: <None> |f.maker.inputs.value: None
f1's input storage: <None> |f1.maker.inputs.value: None

 1 st inputs
f's input storage: <13> |f.maker.inputs.value: <13>
f1's input storage: <14> |f1.maker.inputs.value: <13>
```

nouiz:
I hope you have/will have a good defence. You are right, what you show demonstrate a problem. The shared variable end up with different storage in the storage_map

```
>>> id(f.fn.storage_map[f.maker.fgraph.inputs[1]])
139973909779520
>>> id(f1.fn.storage_map[f1.maker.fgraph.inputs[1]])
139973901903272
```

You can look at if after your defense. If you didn't found a solution Monday/Tuesday, I'll try to look at it then.

ChienliMa:
Thanks, Fred.
Here, the storage is designed to be separated. Because this shared variable has ```update```, and is mutable. Only immutable shared variables share data. 
```python
# codes in __copy__()
for (input, _1, _2), here, there in zip(self.indices,
                                        self.input_storage,
                                        cpy.input_storage):
    if input.mutable and here is not None:
        there.data = copy.copy(here.data)
    else:
        there.data = here.data
```

Or do you mean they should share storage even they have ```update```. ( In this case, the shared variable should have different value after different calls )

-----------

My idea is that, in the original ```__copy__()``` function, when the function is copied, the maker is not. Therefore the copied function still using the original maker, and those original ```In``` instances.  The value of ```In```( a container if it's shared variable ) still stores the value of original function. This is why I want to copy those ```In``` instances.

By copying ```In``` instances and use them to initialize a new maker, I can make sure that   ```copied_function.maker.inputs[i].value``` is the value of corresponding input in the copied function, not the original one. 

I push a new commit to solve this problem. Hope it will pass.

nouiz:
I need to review it again, but this PR need a rebase. Can you do it now?

ChienliMa:
wait a moment please.

nouiz:
Can you rebase instead of merging? Or it was too hard? We perfer to not have merge commit in PR. This make the history harder to understand later on.

ChienliMa:
Actually I need some help here. I'm not familiar with ```git rebase```. I just simply checked out new branch and used  ```git rebase```  and then ```git push -f ```, I don't know why there're so many duplicated commits. Should I use ```git reset``` to remove them?

theano/compile/function_module.py
```diff
@@ -496,6 +500,73 @@ def __copy__(self):
                 there.data = here.data
         return cpy
 
+    def copy(self, share_memory=False):
+        """
+        Copy this function. Copied function will have separated maker and
```
nouiz: If shared_memory is False, then the maker is shared.

theano/compile/function_module.py
```diff
@@ -496,6 +500,73 @@ def __copy__(self):
...
+        fgraph with original function. User can choose whether to separate
+        storage by changing the share_memory arguments
+        ---------------------
+        Params:
+            share_memory -- { boolean } Default is False. When True, two
+            function share intermediate storages(storages except input and
+            output storages)
```
nouiz: Tell that this is useful with allow_gc=False.

theano/gof/link.py
```diff
@@ -496,7 +496,10 @@ def map_storage(fgraph, order, input_storage, output_storage):
     else:
         assert len(fgraph.inputs) == len(input_storage)
 
-    storage_map = {}
+    if storage_map is None:
+        storage_map = {}
+
+    # add input storage into storage_map
     for r, storage in zip(fgraph.inputs, input_storage):
         storage_map[r] = storage
```
nouiz: This with the line 495 cause that if the storage_map is provided, but not input_storage, we will change the storage_map of inputs to [None]. Do we want that? If not, you can change line 495 to use the content of the storage_map if provided.

If both, storage_map and input_storage and provided, assert that they point to the same content. If not, raise an error.

theano/gof/link.py
```diff
@@ -504,11 +507,13 @@ def map_storage(fgraph, order, input_storage, output_storage):
 #             raise TypeError("Cannot link a graph with non-constant orphans.", orphan)
 #         storage_map[orphan] = [orphan.data]
 
+    # allocate output storage
     if output_storage is not None:
```
nouiz: We have the same problem here.

I'm not sure what we should do. Do you have an idea? If we provide output_storage or input_storage and storage_map, do they need to match or we overide the content of the storage_map? Whatever the final version is, it must be in the docstring.

nouiz: Can you test this case:

check that it input_storage and storage_map is provided or output_storage, assert that they are the same? If tests pass like that, I think we should keep the assert. This make the understanding easier I think.

ChienliMa:
I think this time I rebased correctly. Do I need to delete those duplicated commits?
And I am working on your comments now.

ChienliMa:
Please comment for this commit so that I can slightly modify the code at weekend.

nouiz:
This test is failing:
```

    self.test(*self.arg)

  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tests/test_flake8.py", line 322, in test_format_flake8

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/compile/debugmode.py:1678:5: E303 too many blank lines (2)

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/compile/function_module.py:565:69: W291 trailing whitespace

```

I'm starting a new round of review of this PR.

theano/compile/tests/test_function_module.py
```diff
@@ -239,6 +239,41 @@ def test_copy(self):
         f(1, 2)  # put them out of sync
         self.assertFalse(f(1, 2) == g(1, 2))  # they should not be equal anymore.
 
+    def test_copy_share_memory(self):
+        x = T.fscalar('x')
+        y = T.tanh((x+2)/(x-0.2)**2)
+
+        # test for PerformaLinker, will cover VM_linker later
+        ori = theano.function([x], [y], mode="FAST_COMPILE")
```
nouiz: Test all linker here.

theano/compile/tests/test_function_module.py
```diff
@@ -239,6 +239,41 @@ def test_copy(self):
...
+        # assert intermediate and Constants storages are shared.
+        # and output stoarges are not shared
+        i_o_variables = fgraph_cpy.inputs + fgraph_cpy.outputs
+        ori_storages = storage_map_ori.values()
+        for key in storage_map_cpy.keys():
+            storage = storage_map_cpy[key]
+            storage_is_shared = any([ storage is s for s in ori_storages])
+            if key not in i_o_variables or isinstance(key, theano.tensor.Constant):
+                self.assertTrue(storage_is_shared)
+            elif key in fgraph_cpy.outputs:
```
nouiz: inputs aren't shared now I think. So this should be an else close.

theano/compile/tests/test_function_module.py
```diff
@@ -239,6 +239,41 @@ def test_copy(self):
         f(1, 2)  # put them out of sync
         self.assertFalse(f(1, 2) == g(1, 2))  # they should not be equal anymore.
 
+    def test_copy_share_memory(self):
```
nouiz: You had a tests with shared variable. Add it in another test. We need 2 tests, one where we don't update the shared variable and one where we do.

theano/compile/function_module.py
```diff
@@ -496,6 +500,83 @@ def __copy__(self):
...
+                    in_cpy.value = in_ori.value
+                else:
+                    storage = getattr(in_cpy, 'value', None)
+                input_storage.append(storage)
+
+            # pop out input_storage in storage_map and only use storage of In
+            # instances to initialize the make, so that we can avoid 
+            # storage conflictions in link.map_storage()
+            assert len(fg_cpy.inputs) == len(input_storage)
+            for in_var in fg_cpy.inputs:
```
nouiz: To simplify the code a little, can you merge this loop with the loop in line 542

nouiz:
The PR seem good, but without the 2 tests with shared variable, we can't be sure there is nothing we missed. The core isn't easy to modify.

Good work.

ChienliMa:
OK, I will finish this at this weekend. 
Here's another problem.
Currently ```copy()``` call ```__copy__()``` when memory isn't shared. As I've explained, ```__copy__()``` might have some problem of storage in ```In``` instance. So I can:
1.    modify ```__copy__()```
2.    implement code in ```copy()``` for a separate storage case .
3.    leave it there.
What do you think?

You can review the code next Monday.

theano/compile/tests/test_function_module.py
ChienliMa: Part of input( SharedVariable without update ) are shared. Therefore here I only test the output. Input is test in the next for-loop.

theano/compile/tests/test_function_module.py
ChienliMa: Well, I just add two SharedVariable into the test case and give one of them an update. I can separate the test if you want, but that will produce repeated codes.

theano/compile/tests/test_function_module.py
ChienliMa: I added test for ```FAST_RUN``` and ```FAST_COMPILE``` modes, which corresponds to ```PerformLinker``` and ```VM_Linker```. ```ProfileMode``` is deprecated so I didn't test it.

-----------------------------

For ```DebugMode``` which use ```_Linker```, the test didn't passed. The reason is that ```variable```s in ```storage_map``` generated by ```_Linker``` are copied variables in ```fgraph```. Therefore I cannot map old variable to new one using memo.
#### Test and output
I set a break_point inside copy() and type the following code.
```python
(Pdb) storage_map=self.fn.storage_map
(Pdb) [key in memo for key in storage_map]
[False, True, False, False, False, False, False, False, False, False, False, False, False, False, True, True, False, False, False, False, False, False, False, False, False, True, False, True, False, True, False, False, False, True, False, False, False, False, True, False]
```

#### Reason
```python
# LINE 1689 in theano/compile/debugmode.py
fgraph_equiv = fgraph.equivalence_tracker
order_outputs = copy.copy(fgraph_equiv.all_variables_ever) 
# ^^^^     variables are copied in the upper line. ^^^^
del fgraph_equiv
order_outputs.reverse()
order = graph.io_toposort(fgraph.inputs, order_outputs)
# variables in order are used to init storage_map
```
So I drop the test for DebugMode.
To solve this problem, I can either delete the copy line or add a doc to inform users not to use copy() in DebugMode.
I only had a glance to debugmode.py, I hope I didn't make it wrong.

nouiz: I think the copy isn't needed. But in all cases, we don't need to uptimise this case as DebugMode handle the memory differently for verification. So not testing DebugMode is good.

nouiz:
I don't recall or find your explanation of why `__copy__()` have problems. Can you repeat or point me to it?

Otherwise I think this PR is ready to merge. Did I forget something to your knowledge? If change are needed for `__copy__`, this can be in another PR.

ChienliMa:
Just this one that I've shown to you:
```python
import theano
import theano.tensor as T
import copy

x = T.scalar()
y = theano.shared(value=12)
f = theano.function([x],[x+y+1],updates={y:y+1},mode='FAST_COMPILE')
f1 = copy.copy(f)

# execute different functions different times
f(1)
f1(1),f1(2)
```
and 
```python
In [39]: print f.input_storage, f1.input_storage
[<None>, <array(13)>] [<None>, <array(14)>]

In [40]: for i1,i2 in zip(f.maker.inputs, f1.maker.inputs):
   ....:         print i1.value, i2.value
   ....:     
None None
<13> <13>
```

As you can see ```f1.maker.inputs.value``` still share references with ```f.maker.inputs.value``` , instead of storage in storage_map(```f1.input_storage```). Therefore when the storage update, ```f1.maker.inputs.value``` did not follow. This is not a big problem, though. I can fix this in another PR.

nouiz:
I see what you mean. Can you fix it rapidly? If not, I'll merge this PR, otherwise it fit well in this PR.

theano/compile/function_module.py
nouiz: The second sentence isn't always True. Can you reread the full docstring and make sure it is up to date and exact?

ChienliMa:
I updated the docs and you can merge it as soon as Travis Ci build pass.
I will try to fix it in another PR.
However, as I need to implement next feature base on this PR, I don't want to stuck here for a long time.

nouiz:
Ok to fix in another PR. There is some trouble to merge this PR now. (merge
conflict with other PR). You can start a new branch based on your PR and
start now the next step.

thanks

ChienliMa:
Ok, I checkout a new branch in my laptop. So what's next? You can guide me on hangouts.

nouiz:
I made an issue to fix the `__copy__` inconsistency: https://github.com/Theano/Theano/issues/3049

ChienliMa:
I pushed a commit that fix that issue and merge two codes. 
Please review it and improve the test case if needed.

ChienliMa:
Hi Fred, do you know something about the following error. It just suddenly occured:
```
Tests if NanGuardMode is working by feeding in numpy.inf and numpy.nans ... ERROR (theano.compile.nanguardmode): Inf detected
ERROR (theano.compile.nanguardmode): Big value detected
ERROR (theano.compile.nanguardmode): In an input
ERROR (theano.compile.nanguardmode): Inputs: 
ERROR (theano.compile.nanguardmode): var
ERROR (theano.compile.nanguardmode): <TensorType(float64, matrix)>
ERROR (theano.compile.nanguardmode): A. <TensorType(float64, matrix)>
ERROR (theano.compile.nanguardmode): val
ERROR (theano.compile.nanguardmode): [array([[ inf,  inf,  inf,  inf,  inf],
```

nouiz:
For the error message you show, it is tests that print that (but catch the associated exception). We should remove that from the output. I'll check that, thanks for tell me.

nouiz:
This PR need a rebase.

nouiz:
gh-3055 is the continuation of this, so I close this PR.

PR #2970, opened 2015-05-27
[Bug] Revert dot22->gemm optimization back to dot22->dot22scalar

carriepl:
The previously done change from dot22->dot22scalar to dot22->gemm had the effect, under some not-fully-understood circumstances,  of duplicating certain Scan nodes and preventing certain optimizations (such as ScanSaveMem) on the duplicates. The result was a drastic increase in memory usage.

The issue should be investigated further and tests added to try and ensure this doesn't occur again but, in the mean time, this fixes this issue which is preventing from people from using the latest version.

### DIFF
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -570,8 +570,6 @@ def local_gpu_dot22(node):
 @local_optimizer([gpu_from_host, tensor.blas.Dot22Scalar])
 def local_gpu_dot22scalar(node):
     """
-    Deprecated : _dot22scalar has been replace by gemm
-    see Dot22scalar for more details
     gpu_from_host(dot22scalar) -> gpudot(gpu_from_host)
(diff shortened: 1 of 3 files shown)


nouiz:
gemm do not have an infer_shape. Could you try rapidly if adding it would solve the problem?

nouiz:
I'm ok with that PR. Anyone can merge when travis pass.

carriepl:
I currently testing with the infer_shape in gemm. So far it seems to work without any issue. I'll keep testing to make sure.

carriepl:
@nouiz : in the optimization that does the replacement of dot22 with dot22scalar, the comment?

nouiz:
for the comment, yes, in the opt.

carriepl:
Nevermind what I said about infer_shape. The issue is still there.
I've added the comment @nouiz requested.

PR #2972, opened 2015-05-27
Remove useless gpu_from_host(host_from_gpu(x)) op.

t13m:
Profiling output after optimization: https://gist.github.com/t13m/f6b2cc1f23229eeabbba#file-profile_after_opt0
Profiling output befoer optimization: https://gist.github.com/t13m/f6b2cc1f23229eeabbba#file-profile_before_opt0

### DIFF
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -34,10 +34,12 @@ _logger = logging.getLogger(_logger_name)
 
 
 def as_cuda_ndarray_variable(x):
+    if x.owner and isinstance(x.owner.op, HostFromGpu):
+        return x.owner.inputs[0]
     if \
-            x.owner is not None and \
+            x.owner and \
             isinstance(x.owner.op, GpuFromHost) and \
-            hasattr(x.owner.inputs[0].owner, 'op') and \
+            x.owner.inputs[0].owner and \
             isinstance(x.owner.inputs[0].owner.op, HostFromGpu):
         return x.owner.inputs[0].owner.inputs[0]
     if hasattr(x, '_as_CudaNdarrayVariable'):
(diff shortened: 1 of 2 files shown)


nouiz:
Can you give us here the optimizer profile before and after this PR?

t13m:
Thanks, the checking is fixed. Host_from_gpu case added. The new profiling output is here: https://gist.github.com/t13m/f6b2cc1f23229eeabbba#file-profile_after_opt0_1
It works, but seems not improve a lot for this profiling example.

I wonder if we can do more gpu_cut_transfer job in as_cuda_ndarray_variable or other opts called before gpu_cut_transfer. Like check_chain(node, gpu_from_host, host_from_gpu).

theano/sandbox/cuda/basic_ops.py
```diff
@@ -34,6 +34,14 @@
 
 
 def as_cuda_ndarray_variable(x):
+    if x.owner and isinstance(x.owner.op, HostFromGpu):
+        return x.owner.inputs[0]
+    if \
+            x.owner and \
```
nouiz: line 39 and 40 should be merged.

nouiz:
The gpu_cut_transfers opt still remove about 4k nodes. I checked rapidly the file cuda/opt.py and some optimization like local_gpu_elemwise_0 call gpu_from_host instead of as_cuda_ndarray_variable(). This do not use the new optimization in this PR. Can update the opt in that file for this. Do not check the opt in that file before local_gpu_elemwise_0, there use of gpu_from_host is normal.

t13m:
New profiling output: https://gist.github.com/t13m/f6b2cc1f23229eeabbba#file-profile_before_opt0_3

nouiz:
Thanks.

t13m:
Thank you. Backslashes removed in latest commit.

PR #2974, opened 2015-05-27
tutorial/using_gpu documentation fixes

lukemetz:
Small tweaks to code snippets to make them run.

### DIFF
--- a/doc/tutorial/using_gpu.txt
+++ b/doc/tutorial/using_gpu.txt
@@ -108,6 +108,7 @@ after the ``T.exp(x)`` is replaced by a GPU version of ``exp()``.
 .. code-block:: python
 
     from theano import function, config, shared, sandbox
+    import theano.sandbox.cuda.basic_ops
     import theano.tensor as T
     import numpy
     import time


nouiz:
Those change aren't needed. You must use the Theano flags device=gpu and floatX=float32 to have it run on the GPU. I think it is documented in that page. If it isn't clear enough, can you try to make it more clear?

lukemetz:
My bad. I had a `force_device=True` in my .theanorc which was giving me the error. Sorry for the inconvenience.

But, without the `import theano.sandbox.cuda.basic_ops` the demo doesn't  run: `THEANO_FLAGS=device=cpu,floatX=float32 python using_gpu.py`. I get `AttributeError: 'module' object has no attribute 'cuda'`. Pull request updated with just these 2 imports.

doc/tutorial/using_gpu.txt
```diff
@@ -108,6 +109,7 @@ after the ``T.exp(x)`` is replaced by a GPU version of ``exp()``.
 .. code-block:: python
 
     from theano import function, config, shared, sandbox
+    import theano.sandbox.cuda.basic_ops
```
nouiz: this one is right thanks.

doc/tutorial/using_gpu.txt
```diff
@@ -39,6 +39,7 @@ file and run it.
...
```
nouiz: this one isn't needed. We don't call cuda.basic_ops there. Can you remove it?
thanks

lukemetz:
Updated. Sorry about this whole pull request taking up more time than its worth.

nouiz:
thanks

PR #2978, opened 2015-05-28
Small doc and crash fix.

nouiz:


theano/compile/debugmode.py
```diff
@@ -1753,6 +1753,12 @@ def make_all(self, profiler=None, input_storage=None,
             else:
                 thunks_py.append(None)
 
+            if not self.maker.mode.check_c_code and thunks_py[-1] is None:
+                print("Op %s don't have a perform, forcing check of the c code" % node.op)
```
lamblin: This line is too long for flake8.

lamblin: Also, it should probably use `_logger` instead of `print`.

lamblin:
Otherwise, I'm OK with the change.

nouiz:
done.

PR #2980, opened 2015-05-29
Fix error and remove unnecessary constructor

bartvm:
The counter that checked whether an array with that name already existed in the zip file should count the original names, not the enumerated ones, otherwise it saves them as `array`, `array_2`, `array_2`, ... instead of `array`, `array_2`, `array_3`.

Also removed a constructor which seemed useless.


bartvm:
@dmitriy-serdyuk

nouiz:
Seem good to me. We can merge when travis pass.

PR #2982, opened 2015-05-29
[WIP] First partial working version of OpFromGraph op to GPU

nouiz:


theano/sandbox/cuda/opt.py
```diff
@@ -2313,3 +2313,73 @@ def local_gpu_allocempty(node):
...
+    if (isinstance(node.op, theano.compile.builders.OpFromGraph) and
+        any([(i.owner and isinstance(i.owner.op, HostFromGpu))
+             for i in node.inputs]) and
+        # If the output is on the GPU, we already moved it to the GPU.
+        not any([isinstance(o.type, CudaNdarrayType) for o in node.outputs])):
+            # Need to implement updates, givens, ...
+            if node.op.kwargs:
+                for k in node.op.kwargs.keys():
+                    if k not in ['mode']:
+                        return
```
ChienliMa: Kind of confused. Is this a feature? Or is uncompleted.

theano/sandbox/cuda/opt.py
```diff
@@ -2313,3 +2313,73 @@ def local_gpu_allocempty(node):
...
+                    new_inputs[i] = new_inputs[i].owner.inputs[0]
+            op = theano.compile.builders.OpFromGraph(
+                new_inputs, new_outputs, **node.op.kwargs)
+            ret = op(*new_outer_inp, **dict(return_list=True))
+            assert all([inp in node.fgraph.variables for inp in theano.gof.graph.inputs(ret)])
+
+            for i in range(len(ret)):
+                if getattr(node.op.outputs[i], 'dtype', None) == 'float32':
+                    ret[i] = theano.tensor.as_tensor_variable(ret[i])
+            return ret
```
ChienliMa: I went through the code(not the test) just now.  Following is my idea, please correct my mistake. I don't want to start on a wrong direction:

1. It seems this code convert every possible inputs&output into ```CudaNdarray```. 
2. Therefore in the next step ```rebuild_collection_shared()```, ```variable.filter_variable()``` will introduce the ```GpuFromHost``` OP. 
3.  Next, the code contains the ```HostFromGpu``` in front of the inputs. 
4. Finally it use the new inputs&outputs to init a new OpFromGraph, the Optimization process within the inner_function will therefore delete unnecessary copy of data.

-----------------------------------

So, is there any missing feature I need to implement? 
I saw there's a note ```Need to implement updates, givens, ...```, that mean we need to do something to updates&givens? I am not very familiar with optimization,  so you could show some "user case" while explaining to me what I need to handle.

theano/sandbox/cuda/opt.py
nouiz: Currently, the opt only know how to handle the mode parameter. So this is incomplete. I didn't taught what to do with the other param.

theano/sandbox/cuda/opt.py
nouiz: givens can add extra inputs. But I don't see any code to handle that correctly in OpFromGraph. I just check OpFromGraph init and updates aren't allowed. Can you go in OpFromGraph.__init__ and raise an error for givens to tell not implemented? This isn't a needed feature for OpFromGraph, so no trouble to not implement it for now.

ChienliMa: That's OK. Should I make a PR to this branch or theano/master?

nouiz: Theano master
Le 27 juil. 2015 08:17, "Chienli Ma(马千里)" <notifications@github.com> a
écrit :

PR #2984, opened 2015-05-29
Remove unsupported option for pip. Should fix Travis

lamblin:


lamblin:
Travis started running, so I'm merging now in order to fix the master.

PR #2987, opened 2015-06-01
Added tensor implementation of numpy's allclose function.

mbeissinger:
NEWS.txt: Add tensor.{allclose,isclose}

### DIFF
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -1197,6 +1197,35 @@ The six usual equality and inequality operators share the same interface.
 
     This is equivalent to ``numpy.isinf``.
 
+.. function:: isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)
+
+    Returns a symbolic ``'int8'`` tensor representing where two tensors are equal
+    within a tolerance.
+
+    The tolerance values are positive, typically very small numbers.
+    The relative difference `(rtol * abs(b))` and the absolute difference `atol` are
+    added together to compare against the absolute difference between `a` and `b`.
+
+    For finite values, isclose uses the following equation to test whether two
+    floating point values are equivalent:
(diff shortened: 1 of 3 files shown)


nouiz:
thanks. numpy.allclose handle correctly inf. Can you add support for it or document that it don't support it? Dev version of numpy have a new parameter for nan, so can you document that it don't support that? That way people won't get unexpected behavior.

Can you add it in the doc in this file doc/library/tensor/basic.txt?

thanks

mbeissinger:
Sure thing, I'll add the support and documentation.

mbeissinger:
Works the same as numpy, docs have been updated.

nouiz:
Can you add tests for nan and inf, to make sure they give the same answer as numpy?

mbeissinger:
They exist in the docstring for isclose(), where should I put them - in the theano.tensor.tests.test_basic module under test_comparison class?

nouiz:
Yes. That is the good place. Currently, we don't run doctests.

mbeissinger:
Tests have been added.

mbeissinger:
Ah it appears the travis build doesn't have numpy with the isclose() method - what should I do? The allclose() tests pass. Here was the error:

ERROR: test_isclose (theano.tensor.tests.test_basic.test_comparison)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tensor/tests/test_basic.py", line 4146, in test_isclose
    numpy.all(v1 == numpy.isclose(l, r, equal_nan=False)),
AttributeError: 'module' object has no attribute 'isclose'

mbeissinger:
I guess second best would be to hardcode what the answer is - the test just won't keep up to date if numpy changes. I'll do that now.

mbeissinger:
Looks all good from this end!

nouiz:
thanks

PR #2989, opened 2015-06-01
[MRG] Fix for using MonitorMode with scan

lamblin:
I'm not sure it is the cleanest way of solving that problem.
Also, it is missing a unit test.

### DIFF
--- a/theano/compile/monitormode.py
+++ b/theano/compile/monitormode.py
@@ -54,6 +54,16 @@ class MonitorMode(Mode):
 
         super(MonitorMode, self).__init__(wrap_linker, optimizer=optimizer)
 
+    def __getstate__(self):
+        lnk, opt = super(MonitorMode, self).__getstate__()
+        return (lnk, opt, self.pre_func, self.post_func)
+
+    def __setstate__(self, state):
+        lnk, opt, pre_func, post_func = state
+        self.pre_func = pre_func
+        self.post_func = post_func
+        super(MonitorMode, self).__setstate__((lnk, opt))
+
     def eval(self, i, node, fn):
         """
         The method that calls the thunk `fn`.
--- a/theano/compile/monitormode.py
+++ b/theano/compile/monitormode.py
@@ -85,7 +85,7 @@ class MonitorMode(Mode):
 
 def detect_nan(i, node, fn):
     for output in fn.outputs:
-        if (not isinstance(numpy.random.RandomState, output[0]) and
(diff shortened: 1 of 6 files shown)


nouiz:
It seem ok for me, except for the missing test. Could we make a clone method in the Mode as we do for the linker?

lamblin:
@nouiz Yes, that would probably be better.

nouiz:
I can merge it now if you prefer to put on the todo list that change. If so, just tell me.

lamblin:
Here are the changes to use `Mode.clone`.
I would prefer to wait for unit tests before merging, so I'm keeping the PR tagged as "WIP".

theano/compile/monitormode.py
```diff
@@ -82,10 +92,24 @@ def requiring(self, *tags):
         ret.post_func = self.post_func
         return ret
 
+    def clone(self, opt_kwargs=None, **kwargs):
+        """
+        Create a new instance of this Mode.
+
+        Keyword arguments can be provided for the linker,
+        but they will be ignored, because ProfileMode needs
```
nouiz: the param opt_kwargs is named link_kwargs in Mode. Can you make that consistent? Also kwargs isn't used in both class.

Can you use in both class just an link_kwargs parameter?

theano/compile/profilemode.py
```diff
@@ -695,6 +695,29 @@ def exp_float32_op(op):
         if not printed_tip:
             print("  Sorry, no tip for today.")
 
+    def clone(self, link_kwargs=None, message=None):
```
nouiz: instead of maintaining ProfileMode, why not delete it now? This would make testing this PR easier.

theano/compile/monitormode.py
lamblin: Yes, that's a mistake.

theano/compile/profilemode.py
lamblin: This part was just refactoring of existing code, so I would not have written new tests for it anyway, since it does not break existing ones.
We can remove ProfileMode, but I think we can wait a bit for that.

nouiz: ok.

lamblin:
I updated the PR with a test, it should be ready for final review.

nouiz:
Ok to merge after the python23 PR.

lamblin:
Trying to trigger a new merge with all the tests for python 3.

PR #2992, opened 2015-06-02
[CRASH] Subtensor list

nouiz:
Fix crash reported in "[theano-users] Slice indexing: tuple vs. list" and clean up

### DIFF
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -355,7 +355,10 @@ class _tensor_py_operators:
 
     # SLICING/INDEXING
     def __getitem__(self, args):
-        if not isinstance(args, tuple):
+        if (isinstance(args, list) and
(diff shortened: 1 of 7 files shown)


theano/tensor/var.py
```diff
@@ -355,7 +355,7 @@ def astype(self, dtype):
 
     # SLICING/INDEXING
     def __getitem__(self, args):
-        if not isinstance(args, tuple):
+        if not isinstance(args, (list, tuple)):
             args = args,
```
lamblin: I don't think this is correct, lots of tests are failing now.

lamblin: In fact, I think the only case where we should treat a list like a tuple is if the list has slices in it.

nouiz:
I amended the first commit with the good fix and added a last commit.

PR #2994, opened 2015-06-02
Opt EquilibiumOptimizer to do less pass

t13m:
When register a optimizer to a EquilibiumOptimizerDB instance, 'final_opt' can be used as argument to make it a final_optimizer in the same time.

### DIFF
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -482,7 +482,7 @@ class FunctionGraph(utils.object2):
             verbose = config.optimizer_verbose
         if verbose:
             print(reason, r, new_r)
-        if r.fgraph is not self:
+        if hasattr(r, 'fgraph') and r.fgraph is not self:
             raise Exception("Cannot replace %s because it does not belong "
                             "to this FunctionGraph" % r, str(reason))
         if r.type != new_r.type:
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -402,6 +402,7 @@ class FunctionGraph(utils.object2):
                  if var.clients]):
 
                 self.variables.remove(variable)
+                del variable.fgraph
 
     def __prune__(self, apply_node, reason=None):
         """Always called on owner of pruned variable from the graph.
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -402,6 +402,8 @@ class FunctionGraph(utils.object2):
                  if var.clients]):
 
                 self.variables.remove(variable)
+                # This allow to quickly know if a var is still in the fgraph
+                # or not.
                 del variable.fgraph
 
     def __prune__(self, apply_node, reason=None):
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -603,11 +603,11 @@ class MergeOptimizer(Optimizer):
                         not hasattr(candidate, 'fgraph')):
                     continue
                 if var.owner and candidate.owner:
-                    node = node.owner
+                    var = var.owner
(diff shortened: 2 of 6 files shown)


t13m:
The profiling output shows local_gpu_optimizers do 1 less pass now.

The profiling output with this optimization: https://gist.github.com/t13m/82ad66e8f2e57952f043#file-after_ver0
The profiling output before this optmization: https://gist.github.com/t13m/82ad66e8f2e57952f043#file-before

nouiz:
This can be used to speed up the canonicalize too. Add the canon_merge opt as a finalizer.

t13m:
Done

nouiz:
Can you do a new profile?

nouiz:
As we are so close to NIPS, I'll merge this Monday. But you can start a new branch for the next opt. thanks.

t13m:
OK. I'll fix this travis build error first.

nouiz:
travis passed.

t13m:
Profiling output:
https://gist.github.com/t13m/82ad66e8f2e57952f043#file-after_2

nouiz:
This new change didn't do what I expected. Can you time again after adding this diff:

```
diff --git a/theano/gof/fg.py b/theano/gof/fg.py
index b20ec78..2167a9e 100644
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -398,6 +398,7 @@ class FunctionGraph(utils.object2):
         for r in variables:
             if not r.clients and r in self.variables:
                 self.variables.remove(r)
+                del r.fgraph
 
     def __prune__(self, apply_node, reason=None):
         node = apply_node
diff --git a/theano/gof/opt.py b/theano/gof/opt.py
index a6f61cc..65b7431 100644
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -597,9 +597,12 @@ class MergeOptimizer(Optimizer):
                 # doing the full cycle check. The full cycle check is
                 # skipped by validate() if the graph don't contain
                 # destroyers.
-                node = pairs[0][0]
+                var = pairs[0][0]
                 candidate = pairs[0][1]
-                if node.owner and candidate.owner:
+                if (not hasattr(var, 'fgraph') or
+                        not hasattr(candidate, 'fgraph')):
+                    continue
+                if var.owner and candidate.owner:
                     node = node.owner
                     candidate = candidate.owner
                     inputs_match = all(node_in is cand_in
```

After that, start the next opt. Keep that second opt even if it didn't help this particular case. I'm sure it is safe and could help other case. I'll merge this PR Monday.

t13m:
The new profiling uotput: https://gist.github.com/t13m/82ad66e8f2e57952f043#file-after_3

theano/gof/fg.py
```diff
@@ -383,6 +383,7 @@ def __prune_r__(self, variables, reason=None):
         for r in variables:
             if not r.clients and r in self.variables:
                 self.variables.remove(r)
+                del r.fgraph
```
nouiz: Can you add this comment to help people understand why this is done:

```
# This allow to quickly know if a var is still in the fgraph or not
```

nouiz:
This PR need a rebase. When done, I can merge.

nouiz:
Make a comment when you add commit to a PR and you want us to check it. New commit do not trigger a notification. This help make the review happen faster. I'm checking it now.

nouiz:
I'm merging. thanks

t13m:
Sorry, I forgot to comment. Thanks for merging

PR #2996, opened 2015-06-02
Scan test values

lamblin:


nouiz:
If the user don't set the good dtype, do we want to cast the value blindly? What if the cast is the other way around: from float to int. The value could end up all being 0s.

I think test_value are more used for shape, so I'm ok with this PR, but I want to raise this question.

lamblin:
The only places I've seen this called is when computing gradients, to make sure the gradient is not upcasted wrt the original variable. I'm not even sure this interface is accessible by a user.

nouiz:
ok. So I merge.

theano/scan_module/tests/test_scan.py
```diff
@@ -4983,12 +4983,34 @@ def loss_ti(ti, sum_ti, mi, W):
...
 
 
+def test_compute_test_value_grad_cast():
+    # Test for test values when variables have to be casted
+    # Reported by Daniel Renshaw at
+    # https://groups.google.com/d/topic/theano-users/o4jK9xDe5WI/discussion
+    backup = theano.config.compute_test_value
+    theano.config.compute_test_value = 'raise'
+    try:
+        h = tensor.matrix('h')
```
nouiz: This test fail when floatX=float32. Is that expected? when floatX=float64, everything in this tests is in float64. So I don't understand how this test can test the change in this PR.

Do I miss something? Can we just use dmatrix to fix this test?

lamblin: Should be fixed in #3044.

PR #2998, opened 2015-06-03
Add MRG_RandomStreams.seed

nouiz:


nouiz:
I'll add more test for more cases. Don't merge yet.

lamblin:
The change makes sense to me, I have a couple of minor comments that can be addressed later.
I'll wait for the new test anyway.

theano/sandbox/rng_mrg.py
```diff
@@ -1077,28 +1077,77 @@ def __init__(self, seed=12345, use_cuda=None):
...
+        :rtype: None
+
+        """
+        if seed is None:
+            seed = self.default_instance_seed
+        self.set_rstate(seed)
+
+        for old_r, new_r, size, nstreams in self.state_updates:
+            if nstreams is None:
+                nstreams = self.n_streams(size)
```
lamblin: Why call `self.n_streams(size)` again on the original `size`, rather than saving the inferred size?

theano/sandbox/rng_mrg.py
```diff
@@ -1077,28 +1077,77 @@ def __init__(self, seed=12345, use_cuda=None):
...
+
+            if self.use_cuda and dtype == 'float32':
+                rstates = rstates.flatten()
+                # HACK - we use fact that int32 and float32 have same size to
+                # sneak ints into the CudaNdarray type.
+                # these *SHOULD NEVER BE USED AS FLOATS*
+                tmp_float_buf = numpy.frombuffer(rstates.data, dtype='float32')
+                assert tmp_float_buf.shape == rstates.shape
+                assert (tmp_float_buf.view('int32') == rstates).all()
+                rstate = tmp_float_buf
```
lamblin: I think we should factor out that logic in one function, rather than copy-pasting it.

theano/sandbox/test_rng_mrg.py
```diff
@@ -932,6 +932,27 @@ def test_multMatVect():
...
+        fn_val1 = fn()
+        assert not numpy.allclose(fn_val0, fn_val1)
+
+        random.seed(234)
+
+        fn_val2 = fn()
+        fn_val3 = fn()
+        assert numpy.allclose(fn_val0, fn_val2)
+        assert numpy.allclose(fn_val1, fn_val3)
+
```
lamblin: Can you also test that calling `random.seed()` resets it to the seed that was originally provided?

theano/sandbox/rng_mrg.py
nouiz: if the original size is a tuple of integer vs a Theano variable, we don't have the same behavior. old_r.inputs[-1] point to the size, but as a Theano variable. If it is Constant, we don't know if the user provided a constant or if it was a tuple that we converted to Constant.

I don't want to change the old logic, so to have exactly the same behavior, I reuse the original parameter.

theano/sandbox/test_rng_mrg.py
nouiz: done.

theano/sandbox/rng_mrg.py
nouiz: done

nouiz:
I think this PR is good. It pass test with the old gpu back-end.

theano/sandbox/rng_mrg.py
lamblin: OK.

lamblin:
I'm OK to merge if Travis passes.

PR #2999, opened 2015-06-03
[CRASH,TESTS] Fix crash recently introduced and fix tests

nouiz:


lamblin:
Seems reasonable, we can merge if Travis passes.

PR #3000, opened 2015-06-03
Do not decref descriptor, as PyArray_View steals the reference

lamblin:
This should fix the following errors:
```
*** Reference count error detected:
an attempt was made to deallocate 11 (f) ***
```

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -1064,7 +1064,7 @@ CudaNdarray_TakeFrom(CudaNdarray * self, PyObject *args){
     if(!CudaNdarray_is_c_contiguous(indices) != 0) {
         PyErr_SetString(PyExc_NotImplementedError,
                         "CudaNdarray_TakeFrom: The indices must be contiguous in memory.");
-        Py_DECREF(indices_obj);
+        Py_DECREF(indices);
         return NULL;
     }
     int nb_indices = CudaNdarray_SIZE((CudaNdarray *)indices) / 2;// int64 are 8 bytes, float32 are 4 bytes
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/cuda_ndarray.cu
```diff
@@ -1030,7 +1030,6 @@ CudaNdarray_TakeFrom(CudaNdarray * self, PyObject *args){
         PyObject * indices_float32 = NULL;
         indices_float32 = PyArray_View((PyArrayObject*)indices_obj,
                                                   float32_descr, NULL);
-        Py_DECREF(float32_descr);
```
nouiz: I checked numpy code and PyArray_DescrFromType do an incref. The question is, do PyArray_view decref it? I'm checking.

Did you tested caglar code without the c code of this op?

nouiz: line 1068 is:

Py_DECREF(indices_obj);

but should be:

Py_DECREF(indices);

But I would be surprised this is the problem.

lamblin: NumPy C-API [says](http://docs.scipy.org/doc/numpy/reference/c-api.array.html#data-type-descriptors):
> Functions that take PyArray_Descr * objects and return arrays steal references to the data-type their inputs unless otherwise noted. Therefore, you must own a reference to any data-type object used as input to such a function.

So I'm pretty sure we should not decref it after `PyArray_View` is called.
I tried to run Caglar's code with this PR and it seemed to work, without any sign of memory leak. I did not try without the C code at all.

I can also fix line 1068.

caglar:
It seems like this PR fixes the deallocation error that I was getting previously.

PR #3001, opened 2015-06-04
[WIP] Prevent destroymaps from affecting lazily-computed vars

lamblin:
This is a hacky solution, just to see if it would work.

### DIFF
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -945,6 +945,20 @@ class DestroyHandler(toolbox.Bookkeeper):
                 raise InconsistencyError("Attempting to destroy indestructible variables: %s" %
                         illegal_destroy)
 
+            # We want to compute a list of nodes that may never be computed,
+            # so that we do not force their computations (they may be uncomputeable).
+            # For that, we start by a list of lazy nodes (ifelse), and use io_toposort
+            # to find out which variables will always be computed, even if the lazy
+            # ones are not.
+            lazy_outputs = [var for var in fgraph.variables
+                            if var.owner and isinstance(var.owner, theano.ifelse.IfElse)]
+            always_computed = graph.io_toposort(fgraph.outputs, lazy_outputs)
+            not_always_computed = [var for var in fgraph.variables
+                                   if var not in always_computed]
+            for var in not_always_computed:
+                if var in droot:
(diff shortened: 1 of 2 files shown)


theano/gof/destroyhandler.py
```diff
@@ -945,6 +945,20 @@ def orderings(self, fgraph):
...
                         illegal_destroy)
 
+            # We want to compute a list of nodes that may never be computed,
+            # so that we do not force their computations (they may be uncomputeable).
+            # For that, we start by a list of lazy nodes (ifelse), and use io_toposort
+            # to find out which variables will always be computed, even if the lazy
+            # ones are not.
+            lazy_outputs = [var for var in fgraph.variables
+                            if var.owner and isinstance(var.owner, theano.ifelse.IfElse)]
+            always_computed = graph.io_toposort(fgraph.outputs, lazy_outputs)
```
nouiz: I don't understand why setting the fgraph.outputs as the inputs parameter is useful. I think fgraph.outputs should be fgraph.inputs. I also think that setting it as an empty list is equivalent and would be faster.

But I think this logic won't work if some of the ifelse depend on another one.

nouiz: I just don't understand this line. If there is only one ifelse, I think it will return all the node needed for the 3 inputs of the ifelse.

I think this PR "fix" the problem as the line 956 check if a variable is in the list of apply nodes. So not_always_computed contain all variable. So it mostly disable all node with destroy_map. So it disable all inplace.

lamblin: My idea was to not disable inplace Ops between the output of lazy operations and the output of the graph.
If all outputs are ifelse nodes, then it would disable it completely.
I know we have to be more clever than that, in particular:
- the "condition" input of ifelse is always computed if the ifelse is computed,
- if a chain of node is only used in one "branch" of an ifelse, then they can work inplace on one another,
- if a node is used in both branches of the same ifelse, then it can work inplace on an earlier one

The first and last points are specific to ifelse, and not other potential lazy ops, so that logic should probably be expressed in the IfElse Op instead, but I don't know how to specify all that.

nouiz:
Do we need this PR? I'm not sure, but I think it was a test and we have discussed to continue in another direction.

PR #3002, opened 2015-06-05
[CRASH] Scan index error

carriepl:
This fixes an IndexError that would occur when running the test case included in the PR. It also removes 3 methods in ScanOp, two of which were never used and one, ``get_output_pos()``, which was used in ``grad()``. 

The method ``get_output_pos()`` allowed to travel between inner inputs and inner outputs of recurrent states. I removed it because : 
- It only worked for variables representing recurrent states (I added another method which can deal with any inner input)
- The method's documentation implied that it could be used for any inner input even it that wasn't the case. This was confusing and just begged for the function to be misused by an unsuspecting user.
- The method WAS actually misused in ScanOp.grad() where it was, by mistake, called to travel between inner inputs and **outer** outputs. This didn't cause any error when taking the gradient of a user-defined Scan node because, in the absence of mitmots, the number of inner outputs associated with recurrent states is equal to the number of outer outputs associated with recurrent states. When taking the gradient of a grad-scan, this assumption breaks and leads to the crash in some cases.


lamblin:
I would have to take a deeper look before confirming it is OK.

carriepl:
You are right; it is possible that the taps list contains non-successive indices. It is also possible that it contains successive indices but out of order. However, I don't get why you say that taking the length in that case would be incorrect. The inner function only receives as inputs the taps it needs so taking fhe length of the taps associated with a state will return the number of inner inputs associated with that state.

carriepl:
This shouldn't be merged yet (I'm adding the tag 'WIP' to the PR so it'll be obvious). I found a few typos that need fixing and I am working on replacing the different functions to navigate between inner and outer inputs and outputs by a single centralized mapping that is computed only once. I found a bug which this will solve as well and it will also lower the possibility of future bugs by reducing the amount of code associated with navigating these variables.

carriepl:
Fixed the typo, implemented the centralized mapping and added a test case specifically for it.

carriepl:
This pull request is linked to the PR #2909 which uses the helper function ``get_outer_iidx_from_inner_iidx_seq()``. Whichever is not merged first will require a rebase and some small adjustments.

nouiz:
This is too deep in scan for me to efficiently review this. I'll let @lamblin or @abergeron do it.

Issue #2976, opened 2015-05-28
BUG IN FGRAPH.REPLACE OR A LISTENER (2)


memming:
<<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.AssertionError'>  local_shape_to_shape_i
ERROR (theano.gof.opt): Optimization failure due to: local_shape_to_shape_i
ERROR:theano.gof.opt:Optimization failure due to: local_shape_to_shape_i
ERROR (theano.gof.opt): TRACEBACK:
ERROR:theano.gof.opt:TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/theano/gof/opt.py", line 1527, in process_node

Full error message dump: http://pastebin.com/YTbueDUS

Theano 0.7.0, and also with bleeding edge

I dumped the [function with `function_dump`](http://www.memming.com/temp/bug_function_dropout.pkl).

nouiz:
I'm able to reproduce it. thanks.

A quick workaround is to disable one optimization with this Theano flag:
"optimizer_excluding=local_shape_to_shape_i"

nouiz:
You have a scan in your graph with random numbers. Did you pass both
outputs of the scan function to theano.function?

It happen that people forget to pass the second outputs.

nouiz:
Can I add your pickle in Theano for a permanant test? Replicating this
manually is hard.

I have a fix in gh-3003.

thanks for the report.

memming:
Sure. Feel free to add it to the test.

Thank you for fixing it.

PR #3003, opened 2015-06-05
FunctionGraph crash

nouiz:
fix gh-2976, need a test before merging.


nouiz:
I added a test. It should work with python3. The pickle load problem fix wasn't complicated.

lamblin:
The new test fails both with Python 3.6 and 3, with:
```
AttributeError: 'module' object has no attribute 'OrderedDict'
```

nouiz:
The problem is only with python 2.6 not with python 3. I skip it with python 2.6 and older.

nouiz:
travis passed.

nouiz:
Mereged wrong PR by mistake. But it already passed a review. @lamblin can you confirm in a comment you are good with this merge?

lamblin:
Yes, it's fine for me.

PR #3005, opened 2015-06-05
improve error message for grad of ProdWithoutZeros

jsalvatier:
PyMC3 users get cryptic errors from Theano if they use `prod` in their models. This aims to improve the situation.


nouiz:
@carriepl do you know if this is good? I think it should return that error message differently. I think you should use return the output of theano.gradient.grad_not_implemented()

lamblin:
I agree with @nouiz, explicitly returning grad_not_implemented with a meaningful message would probably be better.
You can see an example of how to use it in https://github.com/Theano/Theano/blob/3b24a1998da4a7a00925b83ae6843b79e0db3c00/theano/tensor/sort.py#L59

jsalvatier:
Done, I added a test too.

nouiz:
I'm ok with this, but I'll let @lamblin merge.

thanks

lamblin:
Thanks!

PR #3008, opened 2015-06-06
Cross product

marcocaccin:
Added cross product function.
Implementation via Levi-Civita symbol, uses scalar products.

See https://groups.google.com/d/topic/theano-users/yvJDccUWw3M/discussion

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3263,6 +3263,24 @@ def batched_tensordot(x, y, axes=2):
     return result
 
 
+def cross(x, y):
+    """
+    Warning: cross product for vectors of shape (3,) only
+    :param x: Tensor
+    :param y: Tensor
+    This function computes the cross product between the two tensors
+    using a scalar product with the 3D Levi-Civita symbol
+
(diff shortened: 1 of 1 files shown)


dwf:
This will need unit tests and probably more documentation.

theano/tensor/basic.py
```diff
@@ -3263,6 +3263,24 @@ def batched_tensordot(x, y, axes=2):
...
 
+def cross(x, y):
+    """
+    Warning: cross product for vectors of shape (3,) only
+    :param x: Tensor
+    :param y: Tensor
+    This function computes the cross product between the two tensors
+    using a scalar product with the 3D Levi-Civita symbol
+
+    Returns a tensor
```
nouiz: Can you tell the output shape at the same time. As David wrote, it will need tests in the file tensor/tests/test_basic.py. In the tests, compare the output again numpy.cross.

thanks

nouiz:
We try to clean up the PR. Do you intend to finish this? If not, we will close this PR.

PR #3011, opened 2015-06-09
Mixed stuff.

nouiz:


theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
```diff
@@ -410,6 +412,16 @@ def test_dnn_valid():
...
 
 
+def test_dnn_valid_err():
+    try:
+        _params_allgood((1, 2, 4, 4), (1, 1, 2, 2), 'valid',
+                        theano_mode=theano_mode.including("cudnn"),
+                        cls=DnnBase)
+        assert "Shape error not raised"
+    except ValueError:
+        pass
```
lamblin: Why not using `assert_raises` here?

nouiz: done

PR #3013, opened 2015-06-10
Add cuSOLVER QR factorization routine (GEQRF)

bartvm:
@lamblin @caglar My first Theano Op, so let me know if I'm doing it wrong

- [ ] Tests
- [ ] Implement `linalg.qr` interface

The next one I'll probably try to implement is `cusolverDnDormqr`.

### DIFF
--- a/theano/sandbox/cuda/cusolver.py
+++ b/theano/sandbox/cuda/cusolver.py
@@ -1,0 +1,87 @@
+import ctypes
+import theano
+from theano.sandbox.cuda import (CudaNdarrayType, as_cuda_ndarray_variable,
+                                 GpuOp, cuda_ndarray, CudaNdarray)
+from scikits.cuda.cusolver import (_libcusolver, cusolverDnCreate,
+                                   cusolverCheckStatus, cusolverDnDestroy)
+
+dimshuffle = cuda_ndarray.cuda_ndarray.dimshuffle
+
+
+class GpuGeqrf(GpuOp):
+    __props__ = ()
+
+    def output_type(self, inp):
+        return [CudaNdarrayType(broadcastable=[False] * inp.type.ndim)(),
+                CudaNdarrayType(broadcastable=[False])()]
+
+    def make_node(self, inp):
+        inp = as_cuda_ndarray_variable(inp)
+        assert inp.ndim == 2
+
+        return theano.Apply(self, [inp], self.output_type(inp))
+
+    def make_thunk(self, node, storage_map, _, no_recycling=[]):
+        inputs = [storage_map[v] for v in node.inputs]
+        outputs = [storage_map[v] for v in node.outputs]
+
+        def thunk():
+            A = inputs[0][0]
+            m, n = A.shape
+            # LAPACK routines expect Fortran order
+            A_T = dimshuffle(A, (1, 0)).reshape((m, n))
+            A_copy = A_T.copy()
+
+            lda = max(1, m)
+            TAU = CudaNdarray.zeros((min(m, n),))
+            handle = cusolverDnCreate()
+            Lwork = cusolverDnSgeqrf_bufferSize(handle, m, n, A_copy.gpudata,
+                                                lda)
+            Work = CudaNdarray.zeros((max(1, Lwork),))
+            devInfo = CudaNdarray.zeros((1,))
+            cusolverDnSgeqrf(handle, m, n, A_copy.gpudata, lda, TAU.gpudata,
+                             Work.gpudata, Lwork, devInfo.gpudata)
+            cusolverDnDestroy(handle)
+            outputs[0][0] = A_copy
+            outputs[1][0] = TAU
+
+        return thunk
+
+
+geqrf = GpuGeqrf()
+
+
+_libcusolver.cusolverDnSgeqrf.restype = int
+_libcusolver.cusolverDnSgeqrf.argtypes = [ctypes.c_void_p,
+                                          ctypes.c_int,
+                                          ctypes.c_int,
+                                          ctypes.c_void_p,
+                                          ctypes.c_int,
+                                          ctypes.c_void_p,
+                                          ctypes.c_void_p,
+                                          ctypes.c_int,
+                                          ctypes.c_void_p]
+
+
+def cusolverDnSgeqrf(handle, m, n, A, lda, TAU, Work, Lwork, devInfo):
+    status = _libcusolver.cusolverDnSgeqrf(handle, m, n, int(A), lda,
+                                           int(TAU), int(Work),
+                                           Lwork, int(devInfo))
+    cusolverCheckStatus(status)
+
+
+_libcusolver.cusolverDnSgeqrf_bufferSize.restype = int
+_libcusolver.cusolverDnSgeqrf_bufferSize.argtypes = [ctypes.c_void_p,
+                                                     ctypes.c_int,
+                                                     ctypes.c_int,
+                                                     ctypes.c_void_p,
+                                                     ctypes.c_int,
+                                                     ctypes.c_void_p]
+
+
+def cusolverDnSgeqrf_bufferSize(handle, m, n, A, lda):
+    Lwork = ctypes.c_int()
+    status = _libcusolver.cusolverDnSgeqrf_bufferSize(handle, m, n, int(A),
+                                                      lda, ctypes.byref(Lwork))
+    cusolverCheckStatus(status)
+    return Lwork.value


theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,92 @@
...
+            TAU = CudaNdarray.zeros((min(m, n),))
+            handle = cusolverDnCreate()
+            Lwork = cusolverDnSgeqrf_bufferSize(handle, m, n, A_copy.gpudata,
+                                                lda)
+            Work = CudaNdarray.zeros((max(1, Lwork),))
+            devInfo = CudaNdarray.zeros((1,))
+            cusolverDnSgeqrf(handle, m, n, A_copy.gpudata, lda, TAU.gpudata,
+                             Work.gpudata, Lwork, devInfo.gpudata)
+            cusolverDnDestroy(handle)
+            outputs[0][0] = reshape(A_copy, (n, m)).dimshuffle((1, 0))
```
bartvm: @lamblin @nouiz Quick question: Right now the compiled version of my op ends up returning `CudaNdarray` and `TensorVariable` objects instead of `ndarray`s. What am I doing wrong?

bartvm:
Question: I often end up using LAPACK routines directly, instead of the higher level SciPy and NumPy routines, and I think it would be nice if Theano could expose these as well. Is that something you think is a good idea? If so:

* What do we do about naming? GPU only has single precision right now, so do we assume that everything is simple precision and just call the method `geqrf`, or do we use `sgeqrf` and not implement the others?
* Do we try to mirror the `scipy.linalg.lapack` interface? It's more simplified than the LAPACK/cuSOLVER interface.

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,92 @@
...
+    def make_thunk(self, node, storage_map, _, no_recycling=[]):
+        inputs = [storage_map[v] for v in node.inputs]
+        outputs = [storage_map[v] for v in node.outputs]
+
+        def thunk():
+            A = inputs[0][0]
+            m, n = A.shape
+            # LAPACK routines expect Fortran order
+            A_T = dimshuffle(A, (1, 0)).reshape((m, n))
+            A_copy = A_T.copy()
```
nouiz: You can't assume inputs order is c order. I think this code expect that. The input can be strided or fortran order.

theano/sandbox/cuda/cusolver.py
nouiz: It is normal that it return CudaNdarray. This is gpu code.

The reshape on that line is the symbolic reshape. I think you want:
dimshuffle(a_copy, (1,0)).reshape((n,m))

as you have done above.

bartvm: Okay, so I'm assuming that usually some sort of optimization gets applied that moves things back to the host? Because if I set `device=gpu` and do `A.dot(B).eval({...})` the computation happens on GPU, but the result is still a NumPy array. In my case, if I do `gqerf(x).eval({...})` the answer is a symbolic variable.

`dimshuffle(a_copy, (1,0)).reshape((n,m))` does not give the correct result (tests fail). This reshape and dimshuffle is the inverse of the one above, which makes sense I guess.

nouiz:
For the naming, I would not put the dtype in the naming. The Op should call the right version for the corresponding input. Useless in old back-end (always float32), but this is what we do on CPU and the new back-end.

For the interface, ideally, we should have the user interface as scipy, numpy, but allow the full power of lapack. We can do this on a as needed basis.

theano/sandbox/cuda/cusolver.py
bartvm: Yeah, this code expects that, because I thought CUDA arrays were always in C order. What do you mean by strided? And if `A.is_c_contiguous()` is `False`, does that mean I can assume it's Fortran order?

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,92 @@
...
+            lda = max(1, m)
+            TAU = CudaNdarray.zeros((min(m, n),))
+            handle = cusolverDnCreate()
+            Lwork = cusolverDnSgeqrf_bufferSize(handle, m, n, A_copy.gpudata,
+                                                lda)
+            Work = CudaNdarray.zeros((max(1, Lwork),))
+            devInfo = CudaNdarray.zeros((1,))
+            cusolverDnSgeqrf(handle, m, n, A_copy.gpudata, lda, TAU.gpudata,
+                             Work.gpudata, Lwork, devInfo.gpudata)
+            cusolverDnDestroy(handle)
```
nouiz: creating and destroying the handle is time consuming I think. Can you check that? Is can you create it in the make_thunk and just use it in the thunk? This will cause that we never delete it. Better, can you create only one in the file and reuse it for all thunks for this op?

theano/sandbox/cuda/tests/test_cusolver.py
```diff
@@ -0,0 +1,28 @@
...
+from theano.sandbox.cuda.cusolver import geqrf
+from scipy.linalg.lapack import dgeqrf
+
+
+def test_geqrf():
+    rng = numpy.random.RandomState(1)
+
+    def assert_geqrf_allclose(m, n):
+        A = rng.rand(m, n).astype(theano.config.floatX)
+        A_tensor = tensor.matrix('A')
```
nouiz: use fmatrix.

theano/sandbox/cuda/tests/test_cusolver.py
```diff
@@ -0,0 +1,28 @@
...
+        dgeqrf_ = theano.function([A_tensor], geqrf(A_tensor))
+        qr_, tau_, work_, info_ = dgeqrf_(A)
+        qr, tau, work, info = dgeqrf(A)
+        assert_allclose(qr, qr_, 1e-3)
+        assert_allclose(tau, tau_, 1e-3)
+        print(m, n, 'correct')
```
nouiz: Why you don't compare work and info? Don't keep the print in the final version. 1e-3 seem high. Is there something lower that would work?

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,92 @@
...
+_libcusolver.cusolverDnSgeqrf_bufferSize.restype = int
+_libcusolver.cusolverDnSgeqrf_bufferSize.argtypes = [ctypes.c_void_p,
+                                                     ctypes.c_int,
+                                                     ctypes.c_int,
+                                                     ctypes.c_void_p,
+                                                     ctypes.c_int,
+                                                     ctypes.c_void_p]
+
+
+def cusolverDnSgeqrf_bufferSize(handle, m, n, A, lda):
```
nouiz: As this function is very small and currently used in only one place, why you didn't inline it in the thunk function?

bartvm:
Okay. For now this op always uses the single-precision version, `sgeqrf`, should I change anything to make it compatible with the other types in the future?

I'll try to stay close to the `scipy.linalg.lapack` interface then (which is slightly different from the real LAPACK interface), but with `{s,d,c,z}gqerf` replaced by a single `geqrf` function.

nouiz:
I have done some comments but overall, it seem good. travis failed, maybe you need to import test_basic_ops.py in the test fail to make sure we skip it when no GPU is available. Also make it skipped if gpu avail, but not scikits.cuda.

theano/sandbox/cuda/tests/test_cusolver.py
bartvm: Nope, on my computer it fails with `1e-4`

theano/sandbox/cuda/tests/test_cusolver.py
bartvm: 231e436

theano/sandbox/cuda/cusolver.py
bartvm: I guess it's a habit, just not a fan of code that is too nested. (And nested functions aren't picklable, so we avoid them in Blocks in most cases.) If you prefer, I can move it inline.

theano/sandbox/cuda/cusolver.py
bartvm: Correct, just timed it an 95% of the time is spent there. I have to say I'm not really sure what creating/destroying handles mean, but I'll just move it to the `__main__` namespace in the file if you say that's okay.

bartvm:
Yeah, I wanted to ask about skipping, thanks for the pointer to `test_basic_ops`.

theano/sandbox/cuda/cusolver.py
bartvm: Okay, so I moved the handle to outside the `thunk`, but now it spends 87% of the time in going from C to Fortran order and back... Any way around that, or no?

theano/sandbox/cuda/tests/test_cusolver.py
bartvm: `work` contains the working memory, so it didn't seem important to me that they are the same. `info` should be `0` and I'll ad a test for that.

bartvm:
Everything should import smoothly now even when cuSOLVER is not available, and tests should be skipped.

theano/sandbox/cuda/cusolver.py
nouiz: Normally, all inputs to a GPU node are already on the GPU. There is transfer op that take care of transfer.

gqerf(x).eval({...}) should not return a symbolic variable, but real data. (on the CPU as we will add transfer automatically)

theano/sandbox/cuda/cusolver.py
nouiz: if is_c_contiguous return False, it mean it can be f contig or c contig or strided. Here is an example that return a strided version:

a=matrix()
a[::2, ::2]

This return a view and it is strided (not c or f contig)

theano/sandbox/cuda/cusolver.py
nouiz: cublas function have some trans (for transpose) parameter. if cusolve have them, you can use them and it will accept an c contig input. So mostly, if c contig, use trans=True, if f contig use trans=False, if strided copy(this could be more optimized, but is a good start).

theano/sandbox/cuda/tests/test_cusolver.py
nouiz: Why return the working memory? Do it contain useful information? Can it be reused in other computation?

theano/sandbox/cuda/cusolver.py
nouiz: I agree, nested function have limitation. I was not refering to doing that. I was refering to taking the inside of this function and putting it directly where you call it. This remove a python function call and make the code more readable as it is all closer.

theano/sandbox/cuda/cusolver.py
bartvm: So how do I recognise a strided array? `a.strides[0] > 1`?

theano/sandbox/cuda/tests/test_cusolver.py
bartvm: That's to make sure we mirror `scipy.linalg.lapack.sgeqrf`. I think it
would be good if their interfaces are the same, and SciPy returns it,
although I don't think it's actually very useful.

On Wed, Jun 10, 2015 at 12:59 PM, Frédéric Bastien <notifications@github.com

theano/sandbox/cuda/cusolver.py
bartvm: It's mostly because this is the way it's done in scikits.cuda, and it makes
it clear that it's a wrapper of a cuSOLVER function.

theano/sandbox/cuda/cusolver.py
bartvm: I know this exists for some cuBLAS functions, but sadly not for cuSOLVER, which always assumed things are Fortran order. From the documentation: "The dense matrices are assumed to be stored in column-major order in memory."

theano/sandbox/cuda/cusolver.py
lamblin: I'm pretty sure @bartvm's code is correct, this is the same thing that @caglar did when wrapping the cula Op.
`A_T.copy()` will always produce a C-contiguous array, so it's transpose will be F-contiguous as intended.

nouiz: Do you think we can get rid of the copy? Otherwise, we will need to implement as op for this with code optimized for the different stride of the input and output.

theano/sandbox/cuda/tests/test_cusolver.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_geqrf():
+    rng = numpy.random.RandomState(1)
+
+    def assert_geqrf_allclose(m, n):
+        A = rng.rand(m, n).astype(theano.config.floatX)
+        A_ = theano.shared(CudaNdarray(A), borrow=True)
+        A_T = theano.shared(CudaNdarray(A.T), borrow=True)
+        theano.config.optimizer = 'None'
```
bartvm: Oops, need to remove this.

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class GpuGeqrf(CuSolverOp):
+    """Wrapper of cuSOLVER's implementation of the geqrf LAPACK method.
+
+    Notes
+    -----
+    Wrapping of the methods is done as in `scikits.cuda.cusolver`. The
+    interface is made to match that of `scipy.linalg.lapac. The interface
+    is intendend to mimick `scipy.linalg.lapack`.
```
lamblin: The last sentense seems duplicated.

theano/sandbox/cuda/cusolver.py
lamblin: The solver works inplace and destroys the contents of A_copy, so I think it is acceptable.
The only case were we could gain something is if A is already F-contiguous (or at least that each of the columns is contiguous) _and_ the Op is allowed to work inplace.

bartvm:
Now with ormqr op!

theano/sandbox/cuda/tests/test_cusolver.py
```diff
@@ -0,0 +1,40 @@
...
+from theano.sandbox.cuda import cuda_available, CudaNdarray
+
+if not cuda_available:
+    raise SkipTest('CUDA not available')
+if not cusolver_available:
+    raise SkipTest('cuSOLVER not available')
+
+
+def test_geqrf():
+    rng = numpy.random.RandomState(1)
```
lamblin: Could you use `theano.tests.unittest_tools.fetch_seed()` as seed instead of 1? That way it is easier to run the whole test suite with a different seed.

theano/sandbox/cuda/cusolver.py
bartvm: Fixed in a0f9969 (along with a new op for you to review!)

theano/sandbox/cuda/tests/test_cusolver.py
bartvm: cc87b06

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,221 @@
...
+            m, n = A.shape
+            # LAPACK routines expect Fortran order
+            A_T = dimshuffle(A, (1, 0)).reshape((m, n))
+            A_copy = A_T.copy()
+
+            lda = max(1, m)
+            TAU = CudaNdarray.zeros((min(m, n),))
+            Lwork = cusolverDnSgeqrf_bufferSize(handle, m, n, A_copy.gpudata,
+                                                lda)
+            Work = CudaNdarray.zeros((max(1, Lwork),))
```
lamblin: If these allocations become too inefficient, we could add these buffers as additional outputs. That way, if the garbage collector is disabled, the Op could try to re-use them.

theano/sandbox/cuda/cusolver.py
lamblin: Also, I just realized that only `A_copy.gpudata` is used, its dimensions and strides are not read by the calls to cusolver*.
In that case, we can probably get rid of the reshapes, which could remove an implicit memory copy during the first reshape (depending on the exact memory layout of A).

bartvm: Good call, after removing the reshapes tests still pass.

bartvm:
Removing the reshuffles sped things up a bit. Current profile:
```
Total time: 0.003677 s
File: /Users/bartvanmerrienboer/Theano/theano/sandbox/cuda/cusolver.py
Function: thunk at line 53

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    53                                                   @profile
    54                                                   def thunk():
    55         6           10      1.7      0.3              A = inputs[0][0]
    56         6            9      1.5      0.2              m, n = A.shape
    57                                                       # LAPACK routines expect Fortran order
    58         6          290     48.3      7.9              A_T = dimshuffle(A, (1, 0))
    59         6          236     39.3      6.4              A_copy = A_T.copy()
    60
    61         6            9      1.5      0.2              lda = max(1, m)
    62         6          199     33.2      5.4              TAU = CudaNdarray.zeros((min(m, n),))
    63         6            9      1.5      0.2              Lwork = cusolverDnSgeqrf_bufferSize(handle, m, n, A_copy.gpudata,
    64         6          106     17.7      2.9                                                  lda)
    65         6          223     37.2      6.1              Work = CudaNdarray.zeros((max(1, Lwork),))
    66         6          164     27.3      4.5              devInfo = CudaNdarray.zeros((1,))
    67         6            7      1.2      0.2              cusolverDnSgeqrf(handle, m, n, A_copy.gpudata, lda, TAU.gpudata,
    68         6          158     26.3      4.3                               Work.gpudata, Lwork, devInfo.gpudata)
    69         6         2242    373.7     61.0              outputs[0][0] = dimshuffle(A_copy, (1, 0))
    70         6            5      0.8      0.1              outputs[1][0] = TAU
    71         6            4      0.7      0.1              outputs[2][0] = Work
    72         6            6      1.0      0.2              outputs[3][0] = devInfo
```
```
Total time: 0.003307 s
File: /Users/bartvanmerrienboer/Theano/theano/sandbox/cuda/cusolver.py
Function: thunk at line 117

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   117                                                   @profile
   118                                                   def thunk():
   119        12           15      1.2      0.5              A = inputs[0][0]
   120        12           12      1.0      0.4              TAU = inputs[1][0]
   121        12            8      0.7      0.2              C = inputs[2][0]
   122        12           15      1.2      0.5              lda, k = A.shape
   123        12           12      1.0      0.4              m, n = C.shape
   124        12           16      1.3      0.5              ldc = max(1, m)
   125
   126                                                       # LAPACK routines expect Fortran order
   127        12          341     28.4     10.3              A_T = dimshuffle(A, (1, 0))
   128        12          402     33.5     12.2              A_copy = A_T.copy()
   129        12          207     17.2      6.3              C_T = dimshuffle(C, (1, 0))
   130        12          263     21.9      8.0              C_copy = C_T.copy()
   144        12          386     32.2     11.7              Work = CudaNdarray.zeros((self.Lwork,))
   145        12          328     27.3      9.9              devInfo = CudaNdarray.zeros((1,))
   146
   147        12           17      1.4      0.5              cusolverDnSormqr(handle, _CUBLAS_SIDE_MODE[self.side],
   148        12           12      1.0      0.4                               _CUBLAS_OP[self.trans], m, n, k,
   149        12           14      1.2      0.4                               A_copy.gpudata, lda, TAU.gpudata, C_copy.gpudata,
   150        12          926     77.2     28.0                               ldc, Work.gpudata, self.Lwork, devInfo.gpudata)
   151
   152        12          220     18.3      6.7              outputs[0][0] = dimshuffle(C_copy, (1, 0))
   153        12           12      1.0      0.4              outputs[1][0] = Work
   154        12            9      0.8      0.3              outputs[2][0] = devInfo
```

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,220 @@
...
+                             ldc, Work.gpudata, self.Lwork, devInfo.gpudata)
+
+            outputs[0][0] = dimshuffle(C_copy, (1, 0))
+            outputs[1][0] = Work
+            outputs[2][0] = devInfo
+
+        return thunk
+
+
+def gpu_ormqr(side, trans, a, tau, c, lwork):
```
caglar: A short documentation for the arguments that, this function requires might be useful.

bartvm:
This time with `CUDA_LAUNCH_BLOCKING=1` and with larger matrices. They look a lot better now:
```
Total time: 0.029486 s
File: /u/vanmerb/Theano/theano/sandbox/cuda/cusolver.py
Function: thunk at line 52

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    52                                                   @profile
    53                                                   def thunk():
    54         6            6      1.0      0.0              A = inputs[0][0]
    55         6            8      1.3      0.0              m, n = A.shape
    56                                                       # LAPACK routines expect Fortran order
    57         6          153     25.5      0.5              A_T = dimshuffle(A, (1, 0))
    58         6          492     82.0      1.7              A_copy = A_T.copy()
    59
    60         6           15      2.5      0.1              lda = max(1, m)
    61         6          327     54.5      1.1              TAU = CudaNdarray.zeros((min(m, n),))
    62         6           10      1.7      0.0              Lwork = cusolverDnSgeqrf_bufferSize(handle, m, n, A_copy.gpudata,
    63         6           82     13.7      0.3                                                  lda)
    64         6          241     40.2      0.8              Work = CudaNdarray.zeros((max(1, Lwork),))
    65         6          201     33.5      0.7              devInfo = CudaNdarray.zeros((1,))
    66         6           11      1.8      0.0              cusolverDnSgeqrf(handle, m, n, A_copy.gpudata, lda, TAU.gpudata,
    67         6        27779   4629.8     94.2                               Work.gpudata, Lwork, devInfo.gpudata)
    68         6          145     24.2      0.5              outputs[0][0] = dimshuffle(A_copy, (1, 0))
    69         6            7      1.2      0.0              outputs[1][0] = TAU
    70         6            4      0.7      0.0              outputs[2][0] = Work
    71         6            5      0.8      0.0              outputs[3][0] = devInfo
```
```
Total time: 0.038459 s
File: /u/vanmerb/Theano/theano/sandbox/cuda/cusolver.py
Function: thunk at line 116

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   116                                                   @profile
   117                                                   def thunk():
   118         8           10      1.2      0.0              A = inputs[0][0]
   119         8            8      1.0      0.0              TAU = inputs[1][0]
   120         8            5      0.6      0.0              C = inputs[2][0]
   121         8           12      1.5      0.0              lda, k = A.shape
   122         8            5      0.6      0.0              m, n = C.shape
   123         8           13      1.6      0.0              ldc = max(1, m)
   124
   125                                                       # LAPACK routines expect Fortran order
   126         8          251     31.4      0.7              A_T = dimshuffle(A, (1, 0))
   127         8          642     80.2      1.7              A_copy = A_T.copy()
   128         8          157     19.6      0.4              C_T = dimshuffle(C, (1, 0))
   129         8          312     39.0      0.8              C_copy = C_T.copy()
   143         8          318     39.8      0.8              Work = CudaNdarray.zeros((self.Lwork,))
   144         8          268     33.5      0.7              devInfo = CudaNdarray.zeros((1,))
   145
   146         8           14      1.8      0.0              cusolverDnSormqr(handle, _CUBLAS_SIDE_MODE[self.side],
   147         8            7      0.9      0.0                               _CUBLAS_OP[self.trans], m, n, k,
   148         8           11      1.4      0.0                               A_copy.gpudata, lda, TAU.gpudata, C_copy.gpudata,
   149         8        36153   4519.1     94.0                               ldc, Work.gpudata, self.Lwork, devInfo.gpudata)
   150
   151         8          183     22.9      0.5              outputs[0][0] = dimshuffle(C_copy, (1, 0))
   152         8            9      1.1      0.0              outputs[1][0] = Work
   153         8            8      1.0      0.0              outputs[2][0] = devInfo
```

PR #3015, opened 2015-06-10
Fix crash in some case of compilation of GPU code.

nouiz:


carriepl:
Looks fine. Just waiting on travis.

Issue #2962, opened 2015-05-26
Speed up inplace elemwise optimizer


nouiz:
Currently in the inplace_elemwise_optimizer opt in tensor/opt.py, we have a quick hack to try to make it faster. If the graph have move then 500 nodes, we only call validate for each group of 10 nodes replaced. If validate fail (a cycle was detected) we skip all those optimization.

We could call validate less frequently to make it faster, but we risk of loosing even more optimization. So here is another algo that would allow to call validate only log(n) when there is no cycle where n is the number of elemwise in the graph and won't skip optimization.

- elem_list = Make a list of all the elemwise node in the graph.
- then set the number of call to validate after log(len(elem_list))
- Then process as before, but if we fail a validate, set all the skipped nodes in a failed_list.
- After having passed on all node in elem_list, put all nodes from failed_list to elem_list and empty failed_list. and restart(set new threashold before calling validate, ...).
- When the threashold is 1, stop iterating. All the element in failed_list can't be applied.

I suppose the number of failed will be low, as we pre-select some of them to limit the number of failure. The only failure should be caused by cycle.

t13m:
In current code, there is a for loop `for node in list(graph.io_toposort(fgraph.inputs, fgraph.outputs)):`,  failed node will be skipped when executing this for loop. If we try to optimize failed nodes iteratively as described, the time cost can be large. What do you think?

nouiz:
There is 2 contradiction goal that we must reconsile.
1) We want optimization fast
2) We want the most optimized graph.

What is taking time in this optimization is the call to validate. So we want to call it the less frequently we can.

So to make the fastest, we could do all opt and call it only once at the end. But if this fail, then we loose all optimization, not just the last one.

So I describe an algo that I think would try to archive both goal most of the time. I think that the number of failure of validate will be low as we prevalidate many case manually before doing the optimization. This allow us to do few call to validate. But I still try to get all optimization when this fail. That is why there is an iteration on the failied node, but where we will do more frequently call to validate to allow part of the failed nodes to get optimized.

Your comment made me think of a small modificaiton to the algo, when validate fail, if there is only 1 node that we optimized, we should not try it again. There is no change that we can optimize it later.

Can you add a link here to your current code? I forget where it is and it isn't in a PR. Can you make a PR with your current code? That would be better.

t13m:
Thanks for explanation. PR is made here https://github.com/Theano/Theano/pull/3033

nouiz:
[...]

PR #3033, opened 2015-06-15
Speed up inplace elemwise optimizer

t13m:
This is a PR related to issue #2962 .
It now fails in compile test cases.

### DIFF
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -253,12 +253,11 @@ def _build_droot_impact(destroy_handler):
             input = app.inputs[input_idx]
 
             # This part is the inline version of original getroot function
-            r = input
-            try:
-                while True:
-                    r = destroy_handler.view_i[r]
-            except KeyError:
-                pass
+            view_i = destroy_handler.view_i
+            _r = input
+            while _r is not None:
+                r = _r
+                _r = view_i.get(r)
             input_root = r
 
             if input_root in droot:
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -13,6 +13,7 @@ from theano.compat import OrderedDict
 from theano.misc.ordered_set import OrderedSet
 
 from .fg import InconsistencyError
+from Queue import Queue
 
 
 class ProtocolError(Exception):
@@ -211,6 +212,10 @@ def get_impact(root, view_o):
 
 def add_impact(r, view_o, impact):
     """
+    This function has been integrated as an inline function. See function
+    _build_droot_impact. For the inline version, code without recursion is
+    used as the commented code.
+
     In opposition to getroot, which finds the variable that is viewed *by* r, this function
     returns all the variables that are views of r.
 
@@ -223,6 +228,13 @@ def add_impact(r, view_o, impact):
           IG thinks so, based on reading the code. It looks like get_impact
           does what this docstring said this function does.
     """
+    # queue = Queue()
+    # queue.put(r)
+    # while not queue.empty():
+    #     v = queue.get()
+    #     for n in view_o.get(v, []):
+    #         impact.add(n)
+    #         queue.put(n)
     for v in view_o.get(r, []):
         impact.add(v)
         add_impact(v, view_o, impact)
@@ -257,7 +269,14 @@ def _build_droot_impact(destroy_handler):
 
             # Here is the inline version of original get_impact function
             input_impact = OrderedSet()
-            add_impact(input_root, destroy_handler.view_o, input_impact)
+            # Here is the inline version of original add_impact function
+            queue = Queue()
+            queue.put(input_root)
+            while not queue.empty():
+                v = queue.get()
+                for n in destroy_handler.view_o.get(v, []):
+                    input_impact.add(n)
+                    queue.put(n)
 
             for v in input_impact:
                 assert v not in droot
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -182,64 +182,6 @@ def _contains_cycle(fgraph, orderings):
     return visited != len(parent_counts)
 
 
-def getroot(r, view_i):
-    """
-    This function has been integrated as an inline function. See function
-    _build_droot_impact
-
-    TODO: what is view_i ? based on add_impact's docstring, IG is guessing
-          it might be a dictionary mapping variables to views, but what is
-          a view? In these old docstrings I'm not sure if "view" always
-          means "view variable" or if it also sometimes means "viewing
-          pattern."
-    For views: Return non-view variable which is ultimatly viewed by r.
-    For non-views: return self.
-    """
-    try:
-        return getroot(view_i[r], view_i)
-    except KeyError:
-        return r
-
-def get_impact(root, view_o):
-    """
-    This function has been integrated as an inline function. See function
-    _build_droot_impact
-    """
-    impact = OrderedSet()
-    add_impact(root, view_o, impact)
-    return impact
-
-
-def add_impact(r, view_o, impact):
-    """
-    This function has been integrated as an inline function. See function
-    _build_droot_impact. For the inline version, code without recursion is
-    used as the commented code.
-
-    In opposition to getroot, which finds the variable that is viewed *by* r, this function
-    returns all the variables that are views of r.
-
-    :param impact: is a set of variables that are views of r
-    :param droot: a dictionary mapping views -> r
-
-    TODO: this docstring is hideously wrong, the function doesn't return anything.
-          has droot been renamed to view_o?
-          does it add things to the impact argument instead of returning them?
-          IG thinks so, based on reading the code. It looks like get_impact
-          does what this docstring said this function does.
-    """
-    # queue = Queue()
-    # queue.put(r)
-    # while not queue.empty():
-    #     v = queue.get()
-    #     for n in view_o.get(v, []):
-    #         impact.add(n)
-    #         queue.put(n)
-    for v in view_o.get(r, []):
-        impact.add(v)
-        add_impact(v, view_o, impact)
-
-
 def _build_droot_impact(destroy_handler):
     droot = {}   # destroyed view + nonview variables -> foundation
     impact = {}  # destroyed nonview variable -> it + all views of it
@@ -252,7 +194,7 @@ def _build_droot_impact(destroy_handler):
             input_idx = input_idx_list[0]
             input = app.inputs[input_idx]
 
-            # This part is the inline version of original getroot function
+            # Find non-view variable which is ultimatly viewed by input.
             view_i = destroy_handler.view_i
             _r = input
             while _r is not None:
@@ -266,9 +208,9 @@ def _build_droot_impact(destroy_handler):
             droot[input_root] = input_root
             root_destroyer[input_root] = app
 
-            # Here is the inline version of original get_impact function
+            # The code here add all the variables that are views of r into
+            # an OrderedSet input_impact
             input_impact = OrderedSet()
-            # Here is the inline version of original add_impact function
             queue = Queue()
             queue.put(input_root)
             while not queue.empty():
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -181,22 +181,6 @@ def _contains_cycle(fgraph, orderings):
     return visited != len(parent_counts)
 
 
-def getroot(r, view_i):
-    """
-    TODO: what is view_i ? based on add_impact's docstring, IG is guessing
-          it might be a dictionary mapping variables to views, but what is
-          a view? In these old docstrings I'm not sure if "view" always
-          means "view variable" or if it also sometimes means "viewing
-          pattern."
-    For views: Return non-view variable which is ultimatly viewed by r.
-    For non-views: return self.
-    """
-    try:
-        return getroot(view_i[r], view_i)
-    except KeyError:
-        return r
-
-
 def add_impact(r, view_o, impact):
     """
     In opposition to getroot, which finds the variable that is viewed *by* r, this function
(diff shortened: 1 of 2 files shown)


t13m:
travis build fails on test cases of inplace op and mutable=True/False. Trying to find out why they failed.

t13m:
Fixed.
Doing profiling now, I'll put output here later.

t13m:
The new profiling output: https://gist.github.com/t13m/32e3d9712be6d6a80cfb#file-after-opt-inplace-optimizer_0
```
370.541s for fgraph.validate()
...
1497.519540s - ('inplace_elemwise_optimizer', 'FromFunctionOptimizer', 33) - 315.055s
```

The output before this optimization: https://gist.github.com/t13m/a6931e342ef1be00d552#file-opt3_usercase_after_0
```
72.644s for fgraph.validate()
...
143.351832s - ('inplace_elemwise_optimizer', 'FromFunctionOptimizer', 34) - 30.064s
```

nouiz:
It seem the new version is now 10x slower. We can't merge it now. I'll do a
new review to try to understand what is going on.

Do you have an idea why it is slower?

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
+        while check_each_change > 0 and prev_len_elemlist > len(elemwise_nodelist):
+            prev_len_elemlist = len(elemwise_nodelist)
+            for node in elemwise_nodelist:
+                op = node.op
+                baseline = op.inplace_pattern
+                protected_inputs = [
+                    f.protected for f in node.fgraph._features if
+                    isinstance(f, theano.compile.function_module.Supervisor)]
+                protected_inputs = sum(protected_inputs, [])  # flatten the list
+                protected_inputs.extend(fgraph.outputs)
```
nouiz: The old version was like that, but I think we can move the protected_inputs computation out of the while loop. Just a small speed up.

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
-                        fgraph.revert(chk)
-                        continue
-                    candidate_inputs.remove(candidate_input)
-                    node = new_node
-                    baseline = inplace_pattern
-                    break
+                            new_node = new_outputs[0].owner
+
+                            # Add changed node into changed_node_since_validate
+                            if node not in changed_node_since_validate:
```
nouiz: Instead of doing all those check on a list, (that is costly) using a set would be better I think for changed_node_since_validate.

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
+                                print(e, file=sys.stderr)
+                                raised_warning = True
+                            fgraph.revert(chk)
+
+                            # Add changed node to failed_list
+                            if len(changed_node_since_validate) > 1:
+                                failed_list.extend(changed_node_since_validate)
+                                changed_node_since_validate = []
+                            # Add failed node to failed_list
+                            if node not in failed_list:
```
nouiz: This could be the cause of the big slow down. Here node was already in changed_node_since_validate. So this do much comparison for nothing as the condition will always be False.

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
+                                failed_list.append(node)
+                            continue
+                        candidate_inputs.remove(candidate_input)
+                        node = new_node
+                        baseline = inplace_pattern
+                        break
+            elemwise_nodelist = failed_list
+            failed_list = []
+            if len(elemwise_nodelist) == 0:
```
nouiz: this condition can be moved 2 line before to remove some work. You will need to check the size of failed_list.

nouiz:
I did a few comments that could speed up the computation.

t13m:
Hi, thanks for reviewing. I think what made it slower is the optimization iteratively applied on the failed list, which is very inefficient. For example, in the first iteration of the `While` loop, the failed list contains n nodes. In the next iteration, maybe optimization on one node applied successfully, but the failed list still contained n-1 nodes. The ending condition in current code is len(failed_list) == 0 || prev_len <= len(failed_list), which means when n is large, the optimizer can iterate for n times (worst case). Compared with original version of code, I think this is the most time consuming part.

If I understood correctly, there is a trade-off between the number of nodes optimized and the time optimizer takes. In the original version of code, there is no failed_list, so all nodes optimized failed will be skipped. In current code, we're trying to optimize as many nodes as we can by the `while` iterations. So maybe a better ending condition can be set to balance the trade-off?

nouiz:
Can you implemented the suggested modification and time again?

Maybe you are right, but i'm not sure. If that isn't enough, do you have
suggestion of a better trade off? Could you print some extra timing stats,
like the number of iteration? We could stop iterating when we have small
group. In the past, if the fgraph had more then 500 nodes, we did only 1
pass with group of 10 nodes. So we could do, if the original group size is
bigger then 10, we stop iterating when the group size is less then 10?

t13m:
Yes, that's what I'm doing. I'll put the new profiling here later.

I don't know if I'm right, either. But firstly I'll make it print more information, that will be helpful to know the reason and may show the right way.

theano/tensor/opt.py
t13m: Yes, to be a set is a good idea, i think  failed_list should also be a set

t13m:
The new profiling output: https://gist.github.com/t13m/32e3d9712be6d6a80cfb#file-after-opt-inplace-optimizer_1
It shows that new optimizations worked, but still slower than original version.

nouiz:
use line_profile. It is a python profiler that will profile one function and tell you how much time is spent on each line.

t13m:
The output of line profiler: https://gist.github.com/t13m/61fa1acefb2a27aaab59#file-inplace-optimizer-line-profiling

It seems in this user case the while loop is executed for 25 times. The most time consuming operation is validate() and fgraph.destroyers(node) in line 267. I don't understand what does fgraph.destroyers() do, could you briefly explain it? Thanks very much.

Then I made it print out some extra information like this,

```
        while check_each_change > 0 and prev_len_elemlist > len(elemwise_nodelist):
            print("\n=== Inplace Optimizer While Loop ===", file=sys.stderr)
            print("check_each_change: %d" % check_each_change)
            print("elemwise_nodelist: %d\n" % len(elemwise_nodelist), file=sys.stderr)
...
            node_count = 0
            for node in elemwise_nodelist:
                node_count += 1
                print("\n --- For Node %d ---\n" % node_count, file=sys.stderr)
...
                        try:
...
                        except (ValueError, TypeError, InconsistencyError) as e:
...
                            fgraph.revert(chk)

                            print("   Reverted!", file=sys.stderr)
                            print("    changed_node_since_validate: %d" % len(changed_node_since_validate), file=sys.stderr)
                            print("    failed_list: %d" % len(failed_list), file=sys.stderr)
```

Here is what I got https://gist.github.com/t13m/61fa1acefb2a27aaab59#file-inplace-optimizer-profiling
Through the while loop, length of elemwise_nodelist decreases as follows,
[2548, 1517, 986, 674, 462, 390, 286, 228, 183, 156, 128, 109, 87, 73, 64, 53, 44, 39, 37, 33, 132, 15, 9, 6, 2], I plotted a figure here https://drive.google.com/file/d/0B2b7Zo-aJOgwSVI0Y3NtMzdSRXc/view?usp=sharing

I think maybe the while loop should be stopped earlier, but i didn't measure the time took by every iteration yet.

t13m:
I measured the time of each iteration
```
164.944055 seconds
175.430905 seconds
147.706283 seconds
115.976406 seconds
95.490152 seconds
87.722477 seconds
63.806031 seconds
55.061313 seconds
43.343463 seconds
41.062584 seconds
31.758372 seconds
23.928888 seconds
21.226140 seconds
19.755656 seconds
15.677064 seconds
18.763028 seconds
21.131622 seconds
20.590929 seconds
0.000001 seconds
19.791509 seconds
2.747470 seconds
```

t13m:
The line profiler output shows that almost all nodes have been optimized, the number of nodes optimized is much larger than in the past.
Is it possible to make validate() run faster? or have a special version of validate() only deal with possible case of becoming invalid caused by the optimizer?

nouiz:
I think you can speed up fgraph.destroyers. Go in the file theano/gof/destroyhandler.py The stuff is complicated, you don't need to understand everything.  fgraph.destroyers correspond to the method get_destroyers_of() (there is copy of that method, not great, but anyway, it change nothing for the speed up, they are the exact same copy of the function at 2 places.

Those function call the function `self.refresh_droot_impact()`, that call `_build_droot_impact`. This is the function that you can optimize. This function call 2 other functions in that file: getroot and get_impact. Those 2 functions are very call frequently and are very short. The extra function call is costly in python. So you can inline those call in the function _build_droot_impact and delete them.

getroot and get_impact are used in 2 places in the file. But those 2 place are the exact same code! Can you refactor this at the same time to remove the code duplication. You can make refresh_droot_impact call _build_droot_impact.

That inlining will speed up the computation, but I don't know by how much, so reprofile after to know if this is still a significant botttleneck.

nouiz:
I would recommant you to do the first opt in the comment above first. This one is more well defined.

It is very hard to speedup validate. I don't want you to go in that now. This would be something separate.

I find the print output you gave very strange as the number of node at each iteration. It seem to be halved at each iteration. I do not expect such a clean change in the number of nodes at each iteration (even with the small strange value in the graph at the end). Can you try to understand why this happen?

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
+                                print((
+                                        "Some inplace optimization was not "
+                                        "performed due to unexpected error:"), file=sys.stderr)
+                                print(e, file=sys.stderr)
+                                raised_warning = True
+                            fgraph.revert(chk)
+
+                            # Add changed node to failed_list
+                            if len(changed_node_since_validate) > 1:
+                                failed_list.update(list(changed_node_since_validate))
```
nouiz: no need to call list() on this line.

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
+                op = node.op
+                baseline = op.inplace_pattern
+                candidate_outputs = [i for i in xrange(len(node.outputs))
+                                     if i not in baseline]
+                # node inputs that are Constant, already destroyed,
+                # fgraph protected inputs and fgraph outputs can't be used as
+                # inplace target.
+                # Remove here as faster.
+                candidate_inputs = [i for i in xrange(len(node.inputs))
+                                    if i not in baseline.values() \
```
nouiz: the "\" at the end of the line are not pep8 compliant and in this case, are not useful. Just remote them.

theano/tensor/opt.py
```diff
@@ -221,89 +221,124 @@ def inplace_elemwise_optimizer(fgraph):
...
+                                        and not isinstance(node.inputs[i],
+                                                           Constant)\
+                                        and not fgraph.destroyers(node.inputs[i])\
+                                        and node.inputs[i] not in protected_inputs]
```
nouiz: Move the destroyers condition at the end of the conditions. This could make it called a little bit less often.

nouiz:
If you compare the time per iteration, the first iteration take more time then the version before this PR. This is probably caused by the fact that we validated each 10 changes in the past, in this PR, we do it more frequently in the first iteration.

So doing more then 1 iteration don't make sense for now. I'm not sure what to do now, except the speed up of destroyers. Can you also change the initial check_each_change to be the max of the current one and the old one?

t13m:
I'm not quite sure about this. But validate() is called for every 8 changes at the beginning. When validate() fails, all nodes related to these 8 changes will be added into elemwise_nodelist in next iteration. When optimizing this user case, I find it fails frequently.
> I find the print output you gave very strange as the number of node at each iteration. It seem to
> be halved at each iteration. I do not expect such a clean change in the number of nodes at each
> iteration (even with the small strange value in the graph at the end). Can you try to understand why
> this happen?

It reminds me that if fgraph.checkpoint() is fast, it can be called more frequently than now, so that less iterations are needed. We can even have an incremental version of checkpoint() because all changes are known to us. What do you think about this ?

t13m:
The new profiling output is here: https://gist.github.com/t13m/61fa1acefb2a27aaab59#file-profile-destroyhandler-opt

nouiz:
I'm not sure if doing more frequent checkpoint would help. When it fail
fail after 8 changes, we could do a revert of the last 4 changes and
validate again. This could speed up the wrapping code, but not the number
of call to validate. This version of the code is slower as we call it more
frequently.

I think we should for now just accept that we need to do only 1 iteration
on all the elemwise and not try again them again in smaller batch. So we
will continue to loose some inplace opt. Can you that change while keeping
the rest of your code? When we have a faster validate later, we can easy
use this new version.

After that, do a new profileing, but also a new line profile. I want to
know how much time the destroyers is taking now.

t13m:
add_impact is now inline.

nouiz:
This PR need a rebase as we merge a big PR that make our code work for python2/3 with the same code base.

t13m:
Oh, it seems something went wrong when I rebase the code. I'll fix it.

t13m:
Travis build failed on new python3 test cases. Do you know how to fix that?

Here is the new profiling output: https://gist.github.com/t13m/32e3d9712be6d6a80cfb#file-after-opt-inplace-optimizer_2
It is still a little slower than before, especially the merge optimizer.

theano/gof/destroyhandler.py
```diff
@@ -13,6 +13,7 @@
 from theano.misc.ordered_set import OrderedSet
 
 from .fg import InconsistencyError
+from Queue import Queue
```
abergeron: This is your problem.  It has been renamed in python3.  Use `from six.moves.queue import Queue` instead.

t13m: Thanks very much

nouiz:
This PR need a rebase.

nouiz:
I'm not sure what to do with the change in the inplace elemwise optimization.

For now, can you cherry-pick the speed up done in the destroyhandler and make a new PR with this? That way, we can merge it this week.

nouiz:
Could you do the cherry-pick and the new PR with that result today? It should be fast and would allow to get the speed up in Theano rapidly.

t13m:
Hi, I just created PR #3200 for changes of destroyhandler.py only. Thanks.

nouiz:
Closing as the good part was merged in another PR and the rest didn't give the expected result. https://github.com/Theano/Theano/issues/4233 should be the right way to speed this up.

PR #3035, opened 2015-06-15
Add missing test.

thomasmesnard:
fix gh-2863

### DIFF
--- a/theano/scalar/tests/test_basic.py
+++ b/theano/scalar/tests/test_basic.py
@@ -21,7 +21,7 @@ from theano.scalar.basic import (floats, float32, float64,
                                  ints, int8, int32, complex64,
                                  ComplexError, IntDiv, TrueDiv,
                                  Composite, add, div_proxy,
-                                 and_, eq, neq, invert, mul, Scalar)
+                                 and_, eq, neq, invert, mul, Scalar, InRange)
 from theano.scalar.basic import (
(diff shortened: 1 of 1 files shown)


theano/scalar/tests/test_basic.py
```diff
@@ -412,6 +412,11 @@ def test_grad_identity():
     l = y.sum(dtype=theano.config.floatX)
     theano.gradient.grad(l, x)
```
carriepl: There should be two empty lines between these two functions (PEP8)

theano/scalar/tests/test_basic.py
```diff
@@ -412,6 +412,11 @@ def test_grad_identity():
...
 
+def test_grad_InRange():
+    op = InRange(True, True)
+    x = theano.tensor.fscalar('x')
+    out = op(x, 5, 12)
+    gx = theano.tensor.grad(out, x)
```
carriepl: Same here

carriepl:
Right now, the added test covers the ticket it needed to address. To ensure good coverage, it should also validate that the grad is valid for in a variety of cases, at least the following : 
-x is below the lower bound
-x is on the lower bound (test both with that bound included and excluded from the range
-x is between the lower and upper bound
-x is on the upper bound (test both with that bound included and excluded from the range
-x is higher than the upper bound

carriepl:
I think this is fine and can be merged as soon as Travis passes.

PR #3037, opened 2015-06-16
Python23

abergeron:
This a rebase + fixes of #2869 

This should fix the problem with Unlocker as well as a few semantic problems around longs (We do want to have longs in tests to make sure they actually work correctly).

This should be ready for merging and I'm running the full test suite to make sure.


dwf:
Right, I wasn't sure what to do about longs since the literals don't parse...

nouiz:
The new commits by @abergeron seem good. To be sure, before merge, we should run the daily buildbot in python2 and  python3 with a GPU to make sure all is good.

abergeron:
Aaaand this needs a rebase again!

abergeron:
I'd rather we merge early (since travis passes) and then fix the problems we face afterwards.

Otherwise, the near-constant rebases this will need mean that the buildbot run will always be testing an older version from a previous rebase (or almost always).

Also I got one test failure in cuda/test/test_memory.py because it appears that compiling a function uses less GPU memory than before.  I'm not sure why though.

lamblin:
Could you put pack `six` in `theano.compat` as well?
Pylearn2 depends on it currently.

abergeron:
I really rather we didn't have six bundled, but we can add it back temporarily to help pylearn2.

We should make pylearn2 use the real six library rather than theano's bundled copy.

lamblin:
@abergeron OK, I'll take care of that for Pylearn2, but in the mean time, let's put it back temporarily.
We will also have to update the requirements in the doc, and in setup.py.

abergeron:
I've re-bundled six for now, to be removed later.

nouiz:
travis fail again. I'll stop merging PR except fix until this one get in.

I think this make it clear that this PR is way too big. This always cause problems.

nouiz:
I don't like removing six. It was made to be bundled. This assert that we have the version we require. This is important, as old os bundle only an old version. It is easier to support of numpy version then six version. bundling it is just a very small inconvenience that didn't caused problems for years. I don't see any good reason to change that.

dwf:
@nouiz It caused inconvenience in the sense that people started modifying it, which should never be done for bundled code.

dwf:
Sorry guys, I should have broken this up into multiple PRs. I wanted to get everything else at once so that there wouldn't be time for people to start committing more incompatible code; sadly that wasn't quite realistic.

abergeron:
You can merge other things.  It only fails because I made an edit that doesn't respect flake8.

As for six bundling, I think it is a bad idea.  Having it be installed with theano (that's what setup.py does) is a much better idea.  We do specify that we need a certain version to work.

nouiz:
@abergeron do the nighly build in python 2 and python 3 passed with this PR?

abergeron:
In python 2, yes (minus a few minor errors like the memory test for cuda).

In python 3, it's running.

lamblin:
In the long term, I think it is better to get rid of the bundled `six`, and to rely on Python's dependency system instead.
However, we often have to specify `--no-deps` when updating Theano, so that could be troublesome if an older version of `six` is bundled with the OS.

dwf:
I think in the medium term, as a person who develops with Python, you have
to be mildly insane to rely on the OS package manager for your Python
packages when conda exists.

nouiz:
Many people still use the OS package manager.

I'm not sure I would spent time making old OS work, but I certainly won't
break them when this don't bring great benefit. We can't request all python
user to use conda or canopy or similar for now.

But I agree, if we could expect people to use conda/canopy, we could rely
on a good python package manager and remove six.

dwf:
Well there's always also `pip install --user`. Or virtualenv.

nouiz:
Are all six version back-ward compatible? I'm not sure of that. If not,
that isn't a good solution too as it can break other package installed by
the OS...

I'm also not sure it is good to just drop support of things that worked for
long time for no good reason. Especially, when this don't bring much
benefit.

pip install --user don't work in all case too (check my software stack v3
at MILA for example...)

I don't see the benefit of removing the file except removing a small file
from Theano. Do I miss something?

nouiz:
@abergeron do you have news about the python 3 tests?

abergeron:
There is a segfault with our convolution code in CudaNdarray.  There are also a bunch of gradient testing failures along with the problem about less memory being used on the GPU (that is already fixed).

Added to that, I didn't log the previous output (because I assumed that the script did it automatically) so I might have to re-run it to make sure there are no new problems.

nouiz:
I suppose you investigate the problems. If not, tell me.

abergeron:
I working on cudnn Pooling instead of investigating.  I also re-ran the buildbot with proper logging to a file so that future investigations will be easier.

abergeron:
I've found and fixed a problem with Arccos and Arcsin in python3 (basically our test code was less lucky in python3 and hit extremum values which made the numeric gradient go crazy).

There is also a problem of the buildbot segfaulting with this message in the DEBUG_MODE portion:
```
python: mod.cu:1584: int CudaNdarray_conv_valid(const CudaNdarray*, const CudaNdarray*, CudaNdarray*, int, int, int, int, int):
Assertion `CudaNdarray_HOST_DIMS(img)[1] == CudaNdarray_HOST_DIMS(kern)[1]' failed.
```
but that happens on both 2 and 3 so it might not be related to this PR.

We still have the Unlocker problem (even with my attempt at fixing it).  I've found that the problem is that internal function of the os and/or path module (whose name start with `_`) get set to None before our `__del__` method run.  Exception classes also disappear which is why we can't catch the resulting exceptions (and sometimes get an error about that in python3).

The proper fix for that would require a substantial reorganization of the compilelock module to avoid the need for the Unlocker object altogether, but that feels like too much change to tackle on this PR.

nouiz:
There is no segfaul/c assert in the master. So that is related to the PR.

Do you know why the unlocker get called at the end? Why we don't have the
problem with the master and have it there?

abergeron:
I've found which test is causing this and I do get an assert with the current master:
```
THEANO_FLAGS=mode=DEBUG_MODE nosetests -vs tests/test_conv_cuda_ndarray.py:test_dnn_valid_err
Using gpu device 0: GeForce GTX 750 Ti
theano.sandbox.cuda.tests.test_conv_cuda_ndarray.test_dnn_valid_err ... python: mod.cu:1584: int CudaNdarray_conv_valid(const CudaNdarray*, const CudaNdarray*, CudaNdarray*, int, int, int, int, int): Assertion `CudaNdarray_HOST_DIMS(img)[1] == CudaNdarray_HOST_DIMS(kern)[1]' failed.
Aborted (core dumped)
```

abergeron:
One last rebase (I hope) for a minor conflict.

nouiz:
So, is this PR ready to merge? I think so. The assert in DebugMode with cudnn test, I also have it with current master, but it do not show up in daily buildbot as the GPU do not support cudnn. We should fix it, but this is outside this PR.

PR #3044, opened 2015-06-17
Make test pass in float32

lamblin:


nouiz:
I understand that this PR make the test pass, but I don't undertand why using floatX everywhere test that we cast the test value at the right place. Can you explain?

lamblin:
I think the variable needing casting was the initial value, 0L. I checked that without the fix, the new version of the test failed both in float32 and in float64.

nouiz:
right. Ok to merge when travis pass.

PR #3047, opened 2015-06-18
Fix symbol visibility in cuda_ndarray on Windows

lamblin:
This is a first try, I've not been able to test it on Windows yet.

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cuh
+++ b/theano/sandbox/cuda/cuda_ndarray.cuh
@@ -3,8 +3,6 @@
 
 #include <algorithm>
 
-#include "theano_mod_helper.h"
-
 // Defines for Python 2/3 compatibility.
 #if PY_MAJOR_VERSION >= 3
 // Py3k treats all ints as longs. This one is not caught by npy_3kcompat.h.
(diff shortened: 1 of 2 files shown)


lamblin:
@abergeron feedback is welcome.

theano/sandbox/cuda/cuda_ndarray.cuh
```diff
@@ -48,16 +54,19 @@
 #include <cublas_v2.h>
 
 #ifdef _WIN32
+/*
 # ifdef _CUDA_NDARRAY_C
 #  define DllExport   __declspec( dllexport )
 # else
 #  define DllExport   __declspec( dllimport )
 # endif
+*/
```
nouiz: I would delete this comment to make the code cleaner. and the line 66. Otherwise, it seem good for me, but @abergeron know this more then me.

Did you tried it with a clean Theano cache on windows?

abergeron:
This looks good for cuda_ndarray, but as @nouiz said, I would remove the commented-out code.

The default to dllexport *should* be safe, but I'm not 100% sure of that.

lamblin:
I'm testing right now, I'll update when I have a working version.
Regarding the default, it was easier for me to do it that way, since it is not practical to add defines before the `#include` in all the automatically-generated C code, which do not have headers.
The alternative would be to have a second header that would set a define, and include the current one.

lamblin:
This version works on Windows, thanks to @abergeron's help.
I'll check that it did not break things on Linux.

lamblin:
It works on Linux.

nouiz:
So I merge now. Thanks.

Issue #2698, opened 2015-03-29
Relu different speed - optimization not working?


albertz:
Relu is a quite simple function and I would expect that different definitions would all be optimized to the fastest calculation and thus that all different definitions would have the same speed. However, that seems not the case.

1) `T.switch(x < 0., 0., x)`
2) `x * (x > 0)`
3) `T.maximum(0.0, x)`
4) `(T.sgn(x) + 1) * x * 0.5`

From [here](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/scripts/benchmark/time_relu.py) or [here](http://stackoverflow.com/questions/26497564/theano-hiddenlayer-activation-function), it seems as if `T.maximum` is really the slowest on CPU for both fprop and bprop. On GPU, they are similar in fprop but not in bprop where `T.maximum` is the fastest.

f0k:
In Lasagne, we found the last formulation to give the best results on GPU both for the forward and backward pass, because it gets compiled into a single GpuElemwise Op: https://github.com/benanne/Lasagne/pull/163#issuecomment-81806482

Theano can only optimize expressions it has an explicit optimization rule for. While it would be easy to introduce an optimization that replaces `T.maximum(0, x)` with `0.5 * (x + abs(x))`, that wouldn't help for the backward pass -- `theano.grad()` would still be called on the original expression, so it would need another optimization to replace the expression for the gradient of `T.maximum(0, x)` with the expression for the gradient of `0.5 * (x + abs(x))`. For leaky rectified linear units, things would get even more complicated.

An automatic optimization of this would be great, but for now it's probably easier to use a more efficient formulation manually if you're running experiments on a GPU.

By the way, does anybody know if cuDNN would be faster for this? I've had instances where the rectified nonlinearity took about the same time as the first convolution in a network, so it could be worth optimizing.

nouiz:
We could add in Theano a relu method that use the fastest implementation.
That way people do not need to benchmark it. Do you want to push it to
Theano?

If you find case where relu is significatively slow, tell us. We need to
know the input shape and strides. This could be the reason. I didn't plan
to use cudnn for that type of stuff as I do not expect it to give
significant speed up, but if you find place where it is much slower then
expected, this should be checked.

gforge:
[...]

PR #3048, opened 2015-06-18
Add theano.tensor.nnet.relu function

f0k:
As discussed in #2698, this adds a `theano.tensor.nnet.relu` function providing an optimized formulation of the linear rectifier and leaky linear rectifier at least for GPUs, with the potential to be extended with a different formulation for CPUs or even a ReLU Op in future.

/edit: I'm not sure about the docs, how can I test them?

### DIFF
--- a/theano/tensor/nnet/tests/test_nnet.py
+++ b/theano/tensor/nnet/tests/test_nnet.py
@@ -26,7 +26,8 @@ from theano.tensor.nnet import (categorical_crossentropy,
                                 softmax_grad,
                                 softmax_with_bias, SoftmaxGrad,
                                 Prepend_scalar_constant_to_each_row,
-                                Prepend_scalar_to_each_row)
+                                Prepend_scalar_to_each_row,
+                                relu)
 from theano.tensor import matrix, vector, lvector, scalar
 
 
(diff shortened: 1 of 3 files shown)


theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1394,5 +1395,29 @@ def test_stabilize_log_softmax():
     rng = numpy.random.RandomState([2012, 8, 22])
     f(numpy.cast[config.floatX](rng.randn(2, 3)))
 
+
+def test_relu():
+    x = matrix('x')
+    rng = numpy.random.RandomState([2015, 6, 18])
```
nouiz: Use this seed instead (this allow to change it with a Theano flags)

theano.tests.unittest_tools.fetch_seed()

nouiz:
can you import it in the file `theano/tensor/nnet/__init__.py` that way, if we move it people won't need to udpate there code. (and it is easier to use it)

otherwise, it is pretty good. thanks.

theano/tensor/nnet/tests/test_nnet.py
f0k: Okay. What about the test directly above? Shall I change that as well or leave it as is?

f0k:
> can you import it in the file `theano/tensor/nnet/__init__.py`

That file has `from nnet import *` already, and `nnet.py` does not have an `__all__` statement. So my function is already available as `theano.tensor.nnet.relu` (along with `theano.tensor.nnet.numpy`, `theano.tensor.nnet.DisconnectedType` and a lot of other cruft, but that's outside the scope of this PR).

I've changed the random seed for the test as you asked for, and corrected some markup for the documentation.

nouiz:
thanks. I'll merge when travis pass.

f0k:
Travis is happy.

PR #3051, opened 2015-06-19
Try to fix travis

nouiz:


carriepl:
This is to fix the "a not used" flake8 error we're getting these days in travis?

nouiz:
Yes.

Issue #2881, opened 2015-05-05
Documentation for config.assert_no_cpu_op


lamblin:
Now that #2791 has been merged, we should:
- document the config flag in https://github.com/Theano/Theano/blob/master/doc/library/config.txt
- explain how to use it when a GPU function seems to slow, either in https://github.com/Theano/Theano/blob/master/doc/tutorial/debug_faq.txt, in http://deeplearning.net/software/theano/tutorial/using_gpu.html#tips-for-improving-performance-on-gpu, or in a new "GPU FAQ" section
- Another important use case is when no operation are on the GPU, like when float64 was used by mistake (floatX not set to float32, but device=gpu). We could check if there is no GpuOp in the graph, then we do as the new flag tell.

This is a follow-up to #2471.

nouiz:
Another important use case is when no operation are on the GPU, like when float64 was used by mistake (floatX not set to float32, but device=gpu). I added this to the description.

nouiz:
For the third point, I think you need to make a global optimizer, that take the fgraph completly to test if no op are on the GPU. Should we make an exception if all op are deepcopy? I see from time to time this type of function in pylearn2. @lamblin what do you think?

lamblin:
@nouiz I can see two cases, that may not need a global optimizer:
- all shared variables, inputs, and outputs are on CPU, which would be the case of floatX=float64. We can detect that in theano.function.
- There are shared variables on the GPU, but they are only used by HostFromGpu. That one could be a local_optimizer running HostFromGpu, with the possible exclusion of the case where the client for that transfer is an output.
One thing to consider is how to make it easy (or easier) for a user to specify that some functions are meant to run on CPU, and other on GPU. If it is easy, then a user could do that for the functions consisting only of deepcopy.

lamblin:
The documentation is fixed in #3053.
I opened #3062 for the new case.

PR #3053, opened 2015-06-19
Adding the documentation for assert_no_cpu_op #2881.

caglar:
Started writing the documentation for the assert_no_cpu_op.

### DIFF
--- a/doc/tutorial/debug_faq.txt
+++ b/doc/tutorial/debug_faq.txt
@@ -313,7 +313,7 @@ It is possible to use this mode by providing the flag in THEANO_FLAGS, such as:
 ``THEANO_FLAGS="float32,device=gpu,assert_no_cpu_op='raise'" python test.py``
 
 But note that this optimization will not catch all the CPU Ops, it might miss some 
-errors.
+Ops.
 
 .. _faq_monitormode:
 
--- a/doc/tutorial/debug_faq.txt
+++ b/doc/tutorial/debug_faq.txt
@@ -295,6 +295,23 @@ Tips:
   of type *float64*.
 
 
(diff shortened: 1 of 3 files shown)


doc/tutorial/debug_faq.txt
```diff
@@ -295,6 +295,23 @@ Tips:
...
+is a CPU Op on your computational graph. assert_no_cpu_op can take the following one of the three 
+options:
+
+* ``warn``: Raise a warning
+* ``pdb``: Stop with a pdb in the computational graph during the compilation
+* ``raise``: Raise an error,
+if there is a CPU Op in the computational graph.
+
+It is possible to use this mode by providing the flag in THEANO_FLAGS, such as:
+``THEANO_FLAGS="float32,assert_no_cpu_op='raise'" python test.py``
```
nouiz: add device=gpu. Tell that we don't always print them, and I think we still miss some detection pattern.

doc/tutorial/debug_faq.txt
```diff
@@ -295,6 +295,23 @@ Tips:
   of type *float64*.
 
 
+"Why is my GPU function seems to be slow?"
+------------------------------------
```
lamblin: I would say "Why does my GPU function seem to be slow?"

lamblin: Also, you may need to add more "-" to be as wide as the title.

lamblin:
Otherwise, once you address Fred's comments, I think it is good. Thanks!

caglar:
OK I have fixed the parts you commented.

nouiz:
I killed travis as it isn't useful.

lamblin:
Thanks!

PR #3055, opened 2015-06-22
Swap SharedVariable

ChienliMa:
This PR allow users to swap sharedvariable when copy a function. Test included.
Please review this PR.

I will add another feature "delete update" in this PR later.

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -560,9 +560,9 @@ class Function(object):
             share_memory -- { boolean } Default is False. When True, two
             function share intermediate storages(storages except input and
             output storages). Otherwise two function will only share partial
-            storages( see method __copy__() ). If two functions share memory
-            and allow_gc=False, this will increase executing speed and save 
-            memory.
+            storages( see method __copy__() ) and same maker. If two functions
+            share memory and allow_gc=False, this will increase executing speed
+            and save memory.
         ---------------------
         Returns:
             func -- Copied theano.Function
@@ -610,6 +610,13 @@ class Function(object):
                 else:
                     storage = getattr(in_cpy, 'value', None)
                 input_storage.append(storage)
+
+            # pop out input_storage in storage_map and only use storage of In
+            # instances to initialize the make, so that we can avoid 
+            # storage conflictions in link.map_storage()
+            assert len(fg_cpy.inputs) == len(input_storage)
+            for in_var in fg_cpy.inputs:
+                new_storage_map.pop(in_var)
 
             # reinitialize new maker and create new function
             return maker.__class__(inputs=ins, outputs=maker.outputs,
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -580,7 +580,7 @@ class Function(object):
                 "Type of original SharedVariable:", sv_ori.type)
 
         maker = self.maker
-        
+
         # Copy Ins and their storage.
         # so that they have different storage as their value
         ins = [copy.copy(input) for input in maker.inputs]
@@ -629,13 +629,13 @@ class Function(object):
             # Swap SharedVairable in fgraph and ins
             for index, (i, in_v) in enumerate(zip(ins, fg_cpy.inputs)):
                 # Variables in maker.ipnuts are defined by user, therefore we
-                # use them to make comparision and do the mapping. 
+                # use them to make comparision and do the mapping.
                 # Otherwise we don't touch them.
                 swap_sv = maker.inputs[index].variable
 
                 if swap_sv in swap_svs_ori:
                     checkSV(i.variable, swap_sv)
-                    
+
                     # In the fgraph we use the cloned SharedVariable
                     swap_sv = swap_sv.clone()
 
@@ -676,7 +676,7 @@ class Function(object):
         for in_ori, in_cpy, ori, cpy in zip(maker.inputs, f_cpy.maker.inputs,
                                             self.input_storage,
                                             f_cpy.input_storage):
-            
+
             # Share immutable ShareVariable and constant input's storage
             swapped = swap is not None and in_ori.variable in swap_svs_ori
 
@@ -1387,7 +1387,6 @@ class FunctionMaker(object):
             need_opt = False
             updates = [spec.update for spec in inputs if spec.update]
             additional_outputs = map(SymbolicOutput, updates)
-            pass
 
         self.fgraph = fgraph
 
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -604,7 +604,7 @@ class Function(object):
 
         # Re initialize Outs and swap update and variable in Ins
         # By doing this, we can pass FunctionMaker._check_unused_inputs()
-        outs = map(SymbolicOutput, fg_cpy.outputs[:len(maker.outputs)])
+        outs = list(map(SymbolicOutput, fg_cpy.outputs[:len(maker.outputs)]))
 
         update_i = len(outs)
         for i, in_var in zip(ins, fg_cpy.inputs):
@@ -1386,7 +1386,7 @@ class FunctionMaker(object):
             # fgraph is already an optimized one
             need_opt = False
             updates = [spec.update for spec in inputs if spec.update]
-            additional_outputs = map(SymbolicOutput, updates)
+            additional_outputs = list(map(SymbolicOutput, updates))
 
         self.fgraph = fgraph
 
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -624,7 +624,7 @@ class Function(object):
 
             # Check if given ShareVariables exist
             for sv in iterkeys(swap):
-                if not exist_svs.has_key(sv):
+                if not sv in exist_svs:
                     raise ValueError("SharedVariable: %s not found" %
                                      (sv.name))
 
@@ -635,7 +635,7 @@ class Function(object):
                 # Otherwise we don't touch them.
                 var = maker.inputs[index].variable
 
-                if swap.has_key(var):
+                if var in swap:
                     swap_sv = swap[var]
                     checkSV(i.variable, swap_sv)
 
@@ -708,7 +708,7 @@ class Function(object):
                                             f_cpy.input_storage):
 
             # Share immutable ShareVariable and constant input's storage
-            swapped = swap is not None and swap.has_key(in_ori.variable)
+            swapped = swap is not None and in_ori.variable in swap
 
             # Using the original storage if SharedVariable will not be updated
             # and is not swapped
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -589,9 +589,6 @@ class Function(object):
         # Copy Ins and their storage.
         # so that they have different storage as their value
         ins = [copy.copy(input) for input in maker.inputs]
-        for in_cpy, in_ori in zip(ins, maker.inputs):
-            if in_ori.mutable:
-                in_cpy.value = copy.deepcopy(in_ori.value)
 
         # Delete update output in fgraph and updates In instances if needed
         if delete_updates:
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -545,7 +545,11 @@ class Function(object):
         """
         Copy this function. Copied function will have separated maker and
         fgraph with original function. User can choose whether to separate
-        storage by changing the share_memory arguments
+        storage by changing the share_memory arguments. 
+        Note:
+        We reuse original Out instances but In instances, therefore variables
+        in Ins and Outs are not in the same graph. Please avoid using variables
+        is Ins and Outs to reinitialize another Function or FunctionGraoh.
         ---------------------
         Params:
             share_memory -- { boolean } Default is False. When True, two
@@ -554,6 +558,10 @@ class Function(object):
             storages( see method __copy__() ) and same maker. If two functions
             share memory and allow_gc=False, this will increase executing speed
             and save memory.
+
+            swap -- { dict } Dictionary<String, theano.SharedVariable> that
+            map old SharedVariable's name to new SharedVariable. Default is
+            None.
         ---------------------
         Returns:
             func -- Copied theano.Function
@@ -607,7 +615,7 @@ class Function(object):
                                 accept_inplace=maker.accept_inplace).create(
                                 input_storage, storage_map=new_storage_map)
 
-        # Share immutable and constant input storage
+        # Share immutable and constant input's storage
         for in_ori, in_cpy, ori, cpy in zip(maker.inputs, f_cpy.maker.inputs,
                                             self.input_storage,
                                             f_cpy.input_storage):
@@ -615,6 +623,10 @@ class Function(object):
             if is_const or not in_ori.mutable:
                 cpy.data = ori.data
                 in_cpy.value = in_ori.value
+
+            # swap SharedVariable in In instances
+            if in_cpy.variable.name in swap.keys():
+                in_cpy.variable = swap[in_cpy.variable.name]
 
         return f_cpy
 
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -624,7 +624,7 @@ class Function(object):
 
             # Check if given ShareVariables exist
             for sv in iterkeys(swap):
-                if sv not in exist_svs:
+                if not exist_svs.has_key(sv):
                     raise ValueError("SharedVariable: %s not found" %
                                      (sv.name))
 
@@ -648,7 +648,6 @@ class Function(object):
 
                     # Swap SharedVariable in fgraph
                     # if inputs was replaced, change self.inputs
-
                     fg_cpy.inputs[index] = swap_sv
                     fg_cpy.replace(in_v, swap_sv, reason="Swap SV")
 
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -555,10 +555,9 @@ class Function(object):
         Params:
             share_memory -- { boolean } Default is False. When True, two
             function share intermediate storages(storages except input and
-            output storages). Otherwise two function will only share partial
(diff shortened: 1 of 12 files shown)


ChienliMa:
Hi Fred.
I add new feature `delete_update` and rewrite the code.
This version of code is more simplified.
Please review this. If you think this is OK, you can merge this instead of the prior one.

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+        """
+        return self.copy()
+
+    def copy(self, share_memory=False, swap=None, delete_updates=False):
+        """
+        Copy this function. Copied function will have separated maker and
+        fgraph with original function. User can choose whether to separate
+        storage by changing the share_memory arguments.
+        Note:
+        We reuse original Out instances but In instances, therefore variables
```
nouiz: "but In instance" isn't good.

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+        in Ins and Outs are not in the same graph. Please avoid using variables
+        is Ins and Outs to reinitialize another Function or FunctionGraoh.
```
nouiz: I don't understand the last sentence. Can we copy a function multiple time? Add a test for that case.

ChienliMa: Do you mean
```python
cpy = ori.copy()
cpy1 = cpy.copy()
```
or
```python
cpy=ori.copy()
cpy1=ori.copy()
```

theano/compile/function_module.py
ChienliMa: I forgor to update the docs. I will update them in the next commit.
In the new code. I copy both ```In```s and ```Out```s

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+        ---------------------
+        Params:
+            share_memory -- { boolean } Default is False. When True, two
+            function share intermediate storages(storages except input and
```
nouiz: "(storage" -> "("

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+            output storages). Otherwise two function will only share partial
```
nouiz: function -> functions

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+            storages( see method __copy__() ) and same maker. If two functions
```
nouiz: why the node "see method __copy__()"?

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+            share memory and allow_gc=False, this will increase executing speed
+            and save memory.
+
+            swap -- { dict } Dictionary<String, theano.SharedVariable> that
+            map old SharedVariable's name to new SharedVariable. Default is
+            None.
+
+            delete_updates -- { boolean } Default is False. If True, Copied
+            function will not have update.
```
nouiz: The sentense isn't clear for me. What about:

If True, the new function won't do any updates. This do not trigger a new optimization phase, so small graph difference can happen compared to recompiling instead of coping the function.

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+        ---------------------
+        Returns:
+            func -- Copied theano.Function
+        """
+        maker = self.maker
+        # Copy Ins, so that they have different storage as their value
+        ins, outs = copy.deepcopy([maker.inputs, maker.outputs])
```
nouiz: Doing a deep copy of outputs will cause a deep copy of all the original graph.

Can you do that after the condition bellow, to do less deep copy if we delete the outputs?

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
+            outs[i].variable = fg_cpy.outputs[i]
+
+        # Swap update and variable in Ins
+        # By doing this, we can pass FunctionMaker._check_unused_inputs()
+        update_i = len(outs)
+        for i, in_var in zip(ins, fg_cpy.inputs):
+            i.variable = in_var
+            if not delete_updates and i.update is not None:
+                i.update = fg_cpy.outputs[update_i]
+                update_i += 1
```
nouiz: if we have an update, but do not have that shared variable as inputs, is that working correctly? Can you make a test with that case?

theano/compile/function_module.py
```diff
@@ -485,16 +486,178 @@ def __setitem__(self, item, value):
...
-        return cpy
+                i.update = None
+
+        # swap SharedVariable
+        if swap is not None:
+            self.__swapSV(swap, ins, fg_cpy)
+            # the name of SV we swapped
+            swapped_sv = swap.keys()
+
+            print (fg_cpy)
```
nouiz: don't keep that in the final version.

theano/compile/function_module.py
```diff
@@ -485,16 +486,180 @@ def __setitem__(self, item, value):
...
+                "Type of given SharedVariable:", sv_rpl.type,
+                "Type of original SharedVariable:", sv_ori.type)
+
+        exist_names = [i.variable.name for i in ins]
+        swap_names = swap.keys()
+
+        # Check if given names exist
+        for name in swap_names:
+            if name not in exist_names:
+                warnings.warn("Given name: %s wasn't found" % (name))
```
nouiz: this should be an error. We should not swap by name, but by shared var instance. {shared_var: new_shared_var}

theano/compile/function_module.py
```diff
@@ -485,16 +486,180 @@ def __setitem__(self, item, value):
...
+        --------------------
+        Params:
+            swap -- { dict } The same as docs in copy()
+            ins -- { list } List of In instances to be modified.
+            fg_cpy -- { FounctionGraph } Copied FunctionGraph.
+        --------------------
+        Returns:
+            None
+        """
+        def checkSV(sv_ori, sv_rpl):
```
nouiz: as this function is used only once and is short, inline it where it is used.

theano/compile/function_module.py
```diff
@@ -485,16 +486,180 @@ def __setitem__(self, item, value):
...
+                warnings.warn("Given name: %s wasn't found" % (name))
+
+        # Swap SharedVairable in fgraph and ins
+        for index, (i, in_v) in enumerate(zip(ins, fg_cpy.inputs)):
+            assert i.variable.name == in_v.name
+
+            if in_v.name in swap_names:
+                # In the fgraph we use the cloned SharedVariable
+                swap_sv = swap[in_v.name].clone()
+                checkSV(in_v, swap_sv)
```
nouiz: the check can be done before the clone.

theano/compile/function_module.py
```diff
@@ -485,16 +486,180 @@ def __setitem__(self, item, value):
...
+
+                # Swap SharedVariable in fgraph
+                fg_cpy.inputs[index] = swap_sv
+                fg_cpy.replace(in_v, swap_sv, reason="Swap SV")
```
nouiz: SV -> SharedVariable.

nouiz:
I'll try to review this PR again Thursday. It is holyday tomorrow. If you have nothing to do, can you start on the next step?

theano/compile/function_module.py
ChienliMa: Would you mind explaining why this should be plural form ? I suspect this method only work for single function object.

theano/compile/function_module.py
ChienliMa: I should have deleted this. This is the docs for the old code, when ```copy()``` still called ```__copy__()``` when ``` share_memory=False```

theano/compile/function_module.py
ChienliMa: Oh, thanks for reminding. I've not expect that. I think I can re-initialize ```Out```s. However, separated ```In```s and ```Out```s are needed.

theano/compile/function_module.py
nouiz: "two function" two indicate plurials. But I'm not a native english. I think it should be

Otherwise the two functions ...

If you don't agree, I won't insist as I could be wrong.

nouiz:
Just to note that this PR need rebase. It is possible that the rebase isn't trivial as there is big changes merged in master.

theano/compile/function_module.py
ChienliMa: Oh I see, I looked at the wrong line:(

nouiz:
For your info, travis fail. Sorry, I don't have the time to look in detail in your last comment. About your last sentence, could we pass the storage to create() instead of the data? Do you think it is a good idea?

ChienliMa:
I would like to. But this need to modify many places in theano core, including all maker, linker and other related code. If it must be done, I think it's better to do it in another PR.

ChienliMa:
I'm sorry to say that this happen again even though I rebase the code:
```python
Tests if NanGuardMode is working by feeding in numpy.inf and numpy.nans ... ERROR (theano.compile.nanguardmode): Inf detected
ERROR (theano.compile.nanguardmode): Big value detected
ERROR (theano.compile.nanguardmode): In an input
ERROR (theano.compile.nanguardmode): Inputs: 
ERROR (theano.compile.nanguardmode): var
```
Only in some versions of tests.


---------------------------------------

Another failure is that, after I rebase the code, the linker seems to share some inner storage automatically. In some cases it share output storage with intermediate storages. We have a test that output storage should not be shared. But since output storage is shared with intermediate storage, and intermediate storages is shared, we can not promise that any more.  It that ok if we cancel that test?

ChienliMa:
Hi, Fred. Do I need to do something else to make this PR merged?

nouiz:
I need to review it when not too tired. This is not trivial changes and I prefer to be sure all is good. I'll try to do it tomorrow.

nouiz:
travis didn't completly finish, but have that errors:

```

/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/cc.py:1450:63: W291 trailing whitespace
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/cc.py:1602:21: E127 continuation line over-indented for visual indent
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/fg.py:760:67: W291 trailing whitespace
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/fg.py:770:1: W293 blank line contains whitespace
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/link.py:509:15: E121 continuation line under-indented for hanging indent
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/link.py:525:18: E121 continuation line under-indented for hanging indent
```

ChienliMa:
Fixing.

nouiz:
travis fail.

nouiz:
I think this is good, just a comments to update the tests. I didn't again with the DLT, this shoudl be done before merging.

theano/compile/function_module.py
```diff
@@ -488,16 +489,195 @@ def __setitem__(self, item, value):
...
+        """
+        # helper function
+        def checkSV(sv_ori, sv_rpl):
+            """
+            Assert two SharedVariable follow some restirctions:
+                1. same type
+                2. same shape or dim?
+            """
+            assert sv_ori.type == sv_rpl.type, (
+                "Type of given SharedVariable conflicts with origianl one",
```
abergeron: Small typo here: origianl -> original

theano/compile/function_module.py
```diff
@@ -488,16 +489,195 @@ def __setitem__(self, item, value):
...
+
+        maker = self.maker
+
+        # Copy Ins and their storage.
+        # so that they have different storage as their value
+        ins = [copy.copy(input) for input in maker.inputs]
+
+        # Delete update output in fgraph and updates In instances if needed
+        if delete_updates:
+            # The first len(maker.outputs) variabels are original variables.
```
abergeron: variabels -> variables

theano/compile/function_module.py
```diff
@@ -488,16 +489,195 @@ def __setitem__(self, item, value):
...
+                                     (sv.name))
+
+            # Swap SharedVariable in fgraph and In instances
+            for index, (i, in_v) in enumerate(zip(ins, fg_cpy.inputs)):
+                # Variables in maker.inputs are defined by user, therefore we
+                # use them to make comparision and do the mapping.
+                # Otherwise we don't touch them.
+                var = maker.inputs[index].variable
+
+                if var in swap_svs_ori:
```
abergeron: This will not work on python3.

theano/compile/function_module.py
```diff
@@ -488,16 +489,195 @@ def __setitem__(self, item, value):
...
+        for out_ori, out_cpy in zip(maker.outputs, outs):
+            out_cpy.borrow = out_ori.borrow
+
+        # swap SharedVariable
+        if swap is not None:
+            swap_svs_ori = swap.keys()
+            exist_svs = [i.variable for i in maker.inputs]
+
+            # Check if given ShareVariables exist
+            for sv in swap_svs_ori:
```
abergeron: You should use `iterkeys(swap)` (iterkeys comes from six).

theano/compile/function_module.py
abergeron: An alternative is to use `var in swap`.

theano/compile/function_module.py
```diff
@@ -488,16 +489,195 @@ def __setitem__(self, item, value):
...
+                    i.variable = swap_sv
+                    i.value = swap_sv.container
+
+                    # In the fgraph we use the cloned SharedVariable
+                    swap_sv = swap_sv.clone()
+
+                    # Swap SharedVariable in fgraph
+                    # if inputs was replaced, change self.inputs
+                    fg_cpy.inputs[index] = swap_sv
+                    fg_cpy.replace(in_v, swap_sv, reason="Swap SV")
```
abergeron: This seems to make it possible to replace an ordinary input by a shared variable through the copy.  I wonder if that is intended.

abergeron:
I've noted a couple of typos and a problem with python3.

Also, the DLT example still does not seem to learn correctly.  If I run with the copied test function the error is enormous, but if I run with the non-copied version then the error is about the same as for valid.  

The validation copied function seems to work fine which might indicate that there is a problem with the `delete_updates=True` case.

theano/compile/function_module.py
```diff
@@ -488,16 +489,195 @@ def __setitem__(self, item, value):
...
+                                            f_cpy.input_storage):
+
+            # Share immutable ShareVariable and constant input's storage
+            swapped = swap is not None and in_ori.variable in swap_svs_ori
+
+            # Using the original storage if SharedVariable will not be updated
+            # and is not swapped
+            if not in_ori.mutable and not swapped:
+                cpy.data = ori.data
+                in_cpy.value = in_ori.value
```
abergeron: SharedVariables should always use the same container, even if mutable. (Unless they have been swapped for another one).

So the condition should be `if not swapped`.

ChienliMa: condition ```not swapped``` is to make sure not to share container of swapped ```SharedVariable```. I do not copy the container of any ```SharedVariable``` so I think here it's ok to not enter the block for not swapped ```SharedVariable```.

--------------------

condition ```not in_ori.mutable``` is to make sure not to copy container of ```In``` instances. This is to support the original test ```test_copy()``` where the function was defined using ```In``` rather than variables.

ChienliMa:
I can repeat this error now, there is a difference in two definitions of test function. But I could not find any problem even though I set a  break point and followed it into the code.

Still working on it.

ChienliMa:
I think the difference of the error is because two test model are two different function, which cause the difference of fgraph.

```test_model``` is defined by 

```python
    test_model = theano.function(
        inputs=[index],
        outputs=classifier.errors(y),
        givens={
            x: test_set_x[index * batch_size: (index + 1) * batch_size],
            y: test_set_y[index * batch_size: (index + 1) * batch_size]
        }, name='test_model'
    )
```

while ```test_model2``` is defined by 
```python
    train_model = theano.function(
        inputs=[index],
        outputs=cost,
        updates=updates,
        givens={
            x: train_set_x[index * batch_size: (index + 1) * batch_size],
            y: train_set_y[index * batch_size: (index + 1) * batch_size]
        }, name='train_model'
    )

    test_model2 = train_model.copy(swap={train_set_x: test_set_x,
                                         train_set_y.owner.inputs[0]:
                                         test_set_y.owner.inputs[0]},
                                   delete_updates=True,
                                   name="test_model2")
```
Note that ```test_model2``` share the same input with ```train_model```, which is ```cost``` rather then ```error```. 
```validate_model``` doesn't have this problem because it modified from ```test_model``` which use error as output.

What do you think?

I'll fix those minor error at first.

abergeron:
I think you found the problem with test_model2.  If I make a copy of a train2 function that returns the right thing then it works.

When committing the python 3 changes you also added logistic_sgd and best_model.pkl, which I think is a mistake.  Please amend the commit and re-push to not have those in the history.

Also, it seems that the PR needs a rebase.

ChienliMa:
Working on that. I am going to take a flight. I will update later.

theano/compile/function_module.py
abergeron: ok

nouiz:
What about changing the default of copy parameter share_memory to True?
I think it would be better. What do you think @ChienliMa, @abergeron, @carriepl and @lamblin?

ChienliMa:
I think it's OK, as in most case users will not run programming
multi-threadedly.
However, I am not an intensive user of Theano now, so your opinion value
more.

2015-09-01 1:45 GMT-04:00 Frédéric Bastien <notifications@github.com>:

PR #3058, opened 2015-06-22
Don't assert for array sizes, just raise an exception.

abergeron:
This should stop the buildbot from aborting the DEBUG_MODE part.


nouiz:
do this also in the new gpu back-end

abergeron:
Done

nouiz:
I kill travis as it don't test that PR.

PR #3061, opened 2015-06-23
adding documentation for __props__

gyom:
Fixed an error in the existing doc.

Added an example for __props__ where an Op is created and it has two properties to be compared.


doc/tutorial/extending_theano.txt
```diff
@@ -297,6 +297,78 @@ You can try it as follows:
     print out
 
 
+Example for properties of a Op
+==========
```
abergeron: The `=====` line should be as long as the title above it.

doc/tutorial/extending_theano.txt
```diff
@@ -259,7 +259,7 @@ Op Example
 
         def make_node(self, x):
             # check that the theano version has support for __props__
-            assert hasattr(self, '_props')
+            assert hasattr(self, '__props__')
```
abergeron: This was added when \__props__ was very new to make sure that people that tried the example with an older theano had an error.  Now it should be fine to remove it since \__props__ is old enough.

Otherwise, you should leave it as it was before.

doc/tutorial/extending_theano.txt
```diff
@@ -297,6 +297,78 @@ You can try it as follows:
...
+
+        def infer_shape(self, node, i0_shapes):
+            return i0_shapes
+
+        def grad(self, inputs, output_grads):
+            return [a * output_grads[0] + b]
+
+
+The use of :attr:`__props__` saves
+the user the trouble of implementing :func:`__eq__` and :func:`__hash__` manually.
```
abergeron: It also generates a default `__str__` method that prints the attribute names and their values.

doc/tutorial/extending_theano.txt
```diff
@@ -297,6 +297,78 @@ You can try it as follows:
...
+    inp = numpy.random.rand(5, 4).astype(numpy.float32)
+    assert numpy.allclose(4 * inp + 5, f(inp))
+    assert numpy.allclose(2 * inp + 3, g(inp))
+
+
+
+
+
+
+
```
abergeron: This is too much whitespace.

doc/tutorial/extending_theano.txt
gyom: Well, the original reason why "_props" was there is not explained at the tutorial and it's a bit confusing. If it's fine to get rid of this, my opinion would be to do so.

I'm not the one making the judgment call about whether it's time or not to get rid of this. Which one do you want ?

abergeron: I'd have to confer with Fred to be sure if it's ok.  But I would be in favor.

lamblin:
Also, please use "git rebase" on top of the master, rather than "git merge" when there are conflicts. This makes a more straightforward history.

doc/tutorial/extending_theano.txt
gyom: Added your comment to the documentation.

doc/tutorial/extending_theano.txt
nouiz: @carriepl what do you think? I think we should keep the old one. The goal is to know if the Theano version is recent enough. Using self.__props__ won't do that, as the Op will have it implemented.

abergeron: My vote is to remove the check completely since \__props__ is old enough that not many people should have a theano without it and it was in the latest release.

But if we are to keep the check, then it should go back to _props, of course.

nouiz: I would keep it until the next release.

abergeron: You mean 0.7.1 or 0.8 sometime next year?

gyom:
Oops. I did a git-rebase and I think some things went wrong. Now it's including other files, it seems.

doc/tutorial/extending_theano.txt
nouiz: I think 1 year after 0.7. We'll see the number at that time

lamblin:
And the original "merge" commit is still there... we can have a look on monday.

Issue #2616, opened 2015-03-13
Debug flag that print all rejected optimization


nouiz:
Some optimization get rejected for "inplace errors" silently. This isn't always good. A debug flag that will make them displayed would help investigation.

t13m:
So we print out all optimizations rejected for any reason, but not for "inplace errors" only?

nouiz:
I would do it for all, but mark the inplace case as such to differentiate
it.

t13m:
ok. here's the plan, first add a config var named "list_rejected_optimizers", then add methods "get_rejected()" and "print_rejected()" to Optimizer class and LocalOptimizer class. So that information of rejected optimizers can be collected by top level optimizer and can be printed out, in imitation to profiling system.

The output information maybe like this,
```
Global optimizer 1 name
  Global optimizer 2 name: 
     (rejected n times, for reason xxx)
     (rejected m times, for reason yyy)
     ...
  Local optimizer 1 name:
     (rejected p times, for reason zzz)
     ...
...
```

What do you think about this?

t13m:
I'm not very sure that I understand whole mechanism of profiling system, but I think it will be the right way to imitate it. Or just let the profiling system print out rejected optimizers?

nouiz:
I was thinking of something much simpler. In the file gof/toolbox.py there
is a class ReplaceValidate.

This is what add the fgraph.replace_validate(var, new_var, reason) method
to the functiongraph.

I think you can go in the except after the call to validate() and print
there there "reason" and the exception.

PR #3066, opened 2015-06-25
Debug flag print rejected optimizations when validate fails

t13m:
fix gh-2616

### DIFF
--- a/theano/gof/toolbox.py
+++ b/theano/gof/toolbox.py
@@ -208,20 +208,14 @@ class Validator(Feature):
             uf = cf.f_back
             uf_info = inspect.getframeinfo(uf)
 
-            # If the caller is replace_all_validate, just raise the
-            # exception. replace_all_validate will print out the
-            # verbose output.
-            # Or it has to be done here before raise.
-            if uf_info.function == 'replace_all_validate':
-                raise
-            else:
-                verbose = uf.f_locals.get('verbose', False)
-                if verbose:
-                    r = uf.f_locals.get('r', "")
-                    reason = uf_info.function
-                    print("validate failed on node %s.\n Reason: %s, %s" %
-                          (r, reason, e))
-                raise
+            # Print if verbose, then raise
+            verbose = uf.f_locals.get('verbose', False)
+            if verbose:
+                r = uf.f_locals.get('r', "")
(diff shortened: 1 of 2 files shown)


nouiz:
Just a comment. In the description of the PR, if you add something like fix gh-ISSUE_NUMBER, when the PR get closed, it will automatically close the corresponding issue. THis can be in a comment I think:

fix gh-2616

t13m:
In the new commit, Validator.validate_(fgraph) itself will print out information of rejection.

theano/gof/toolbox.py
```diff
@@ -200,7 +201,27 @@ def on_detach(self, fgraph):
 
     def validate_(self, fgraph):
         t0 = time.time()
-        ret = fgraph.execute_callbacks('validate')
+        try:
```
nouiz: I think it would be simpler to just do the printing here and not at replace_validate. This would make the code much simpler.

But otherwise, it is working correctly.

t13m:
Done.

nouiz:
I checked and my suggested modification a worst error message. So having more complicated code is useful. Can you just remove the last commit from this PR? Then I'll merge.

t13m:
Done, but I'm not sure why do you think it is better to not use the simpler code, can you explain a bit more? Thanks

nouiz:
triger new travis build with current master.

lamblin:
Last build (16 days ago) succeeded.
@nouiz, do you think we should trigger a new one? Also, is it worth removing the last two commits, rather than have the last one revert the previous one?

theano/gof/toolbox.py
```diff
@@ -200,7 +201,27 @@ def on_detach(self, fgraph):
...
+            uf_info = inspect.getframeinfo(uf)
+
+            # If the caller is replace_all_validate, just raise the
+            # exception. replace_all_validate will print out the
+            # verbose output.
+            # Or it has to be done here before raise.
+            if uf_info.function == 'replace_all_validate':
+                raise
+            else:
+                verbose = uf.f_locals.get('verbose', False)
```
nouiz: I run all Theano tests with the flag and this else clause wasn't triggered. So it isn't tested, but anyway, I think it is ok, so I merge this PR.

PR #3068, opened 2015-06-25
Fix GPU test failure since cpu softmax change. 

nouiz:


abergeron:
Not tested on travis, so I'll merge.

PR #3069, opened 2015-06-25
added optimization for dividing by one

### LINKED ISSUE #2377: Missing Elemwise optimization
```
   |   | | | | |   | | | | | |   | | |Elemwise{floor_div,no_inplace} [@W] <TensorType(int64, scalar)> ''   2
   |   | | | | |   | | | | | |   | |   |Elemwise{add,no_inplace} [@X] <TensorType(int64, scalar)> ''   1
   |   | | | | |   | | | | | |   | |   | |TensorConstant{-1} [@Y] <TensorType(int64, scalar)>
   |   | | | | |   | | | | | |   | |   | |Shape_i{0} [@K] <TensorType(int64, scalar)> ''   0
   |   | | | | |   | | | | | |   | |   |TensorConstant{1} [@V] <TensorType(int64, scalar)>
```
The floor div can be remove.

This was found by this code: 

```          
import theano
import theano.tensor as T

y = T.tensor4('y')

f = theano.function([y], y)
theano.printing.debugprint(f)

f = theano.function([y], y[::-1])
theano.printing.debugprint(f)

f = theano.function([y], y[::-1][::-1])
theano.printing.debugprint(f, print_type=True)
```

On the gh-2160 branch. After doing it, check if other optimization are missing.

To help find missing optimization, run that code with those flags: THEANO_FLAGS=optimizer_excluding=fusion,optimizer_verbose=True

After this optimization, some other optimization will get executed.

aalmah:
- This fixes #2377.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4397,6 +4397,7 @@ def local_mul_to_sqr(node):
             if node.inputs[0] is node.inputs[1]:
                 return [T.sqr(node.inputs[0])]
 
+
 @register_canonicalize
 @gof.local_optimizer([T.int_div, T.floor_div])
 def local_intdiv_by_one(node):
@@ -4404,7 +4405,7 @@ def local_intdiv_by_one(node):
     """
     if node.op in [T.int_div]:
         if isinstance(node.inputs[1], T.TensorConstant) and \
-           node.inputs[1].value == 1:
+           numpy.all(node.inputs[1].value == 1):
             return [node.inputs[0]]
 
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -2429,7 +2429,6 @@ def local_subtensor_merge(node):
                 lambda x: isinstance(x, T.Variable))
             # Do not call make_node for test_value
             out = subtens(x, *sl_ins)
-
             return [out]
 
 
@@ -4397,6 +4396,17 @@ def local_mul_to_sqr(node):
         if len(node.inputs) == 2:
             if node.inputs[0] is node.inputs[1]:
                 return [T.sqr(node.inputs[0])]
+
+@register_specialize
+@gof.local_optimizer([T.int_div, T.floor_div])
+def local_div_by_one(node):
+    """x // 1 -> x
+    """
+    if node.op in [T.int_div]:
+        if len(node.inputs) == 2:
+            if isinstance(node.inputs[1], T.TensorConstant) and \
+               node.inputs[1].value == 1:
+                return [node.inputs[0]]
 
 
 @gof.local_optimizer([T.pow])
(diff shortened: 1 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -4398,6 +4397,17 @@ def local_mul_to_sqr(node):
             if node.inputs[0] is node.inputs[1]:
                 return [T.sqr(node.inputs[0])]
```
carriepl: There should be a second empty line before this optimization.

theano/tensor/opt.py
```diff
@@ -4398,6 +4397,17 @@ def local_mul_to_sqr(node):
...
 
+@register_specialize
```
carriepl: I don't think that this is a specialization optimization so this can be removed

theano/tensor/opt.py
```diff
@@ -4398,6 +4397,17 @@ def local_mul_to_sqr(node):
...
+@gof.local_optimizer([T.int_div, T.floor_div])
+def local_div_by_one(node):
```
carriepl: Please update the name to reflect that it's an optimization on int_div, not div

theano/tensor/opt.py
```diff
@@ -4398,6 +4397,17 @@ def local_mul_to_sqr(node):
...
+    """x // 1 -> x
+    """
+    if node.op in [T.int_div]:
+        if len(node.inputs) == 2:
```
carriepl: I'm 95% certain that the fact that the scalar_op is an IntDiv will guarantee that the node has exactly 2 inputs.

theano/tensor/opt.py
```diff
@@ -4398,6 +4397,17 @@ def local_mul_to_sqr(node):
...
+            if isinstance(node.inputs[1], T.TensorConstant) and \
+               node.inputs[1].value == 1:
```
carriepl: What happens if the constant contains more than one element? (for instance if you int-divide a vector by a constant vector of ones)?

theano/tensor/tests/test_opt.py
```diff
@@ -5393,6 +5393,32 @@ def test_assert_op_gradient():
     assert func(x_val) == 1
 
 
+class TestDivByOne(unittest.TestCase):
```
carriepl: Please update the name the name to contain IntDiv instead of Div to avoid confusion.

carriepl:
Have you validated that the test fails before your fix?

carriepl:
Finally, the tests are failing in Travis with FAST_COMPILE. This is because, in that mode, your  optimization is not executed. To fix this, you need to explicitly activate the optimization in the compilation mode you are using to compile your Theano function. This can be done in almost the same way as deactivating an optimization : 
`mode = mode.including('my_optimization')`

theano/tensor/opt.py
aalmah: This should work as well. I will add another test for this case.

aalmah:
Yes, I have verified the tests fail before this fix.

carriepl:
This looks very good and Travis is passing so I'm merging.

PR #3072, opened 2015-06-26
Remove an error when Theano process exit.

nouiz:


nouiz:
I'm working on a better fix.

nouiz:
I won't finish shortly the better fix. I think this don't introduce problem and at least fix it in my setup. So I think we should merge it now.

Issue #3018, opened 2015-06-11
Keep graph build stack trace during optimization


nouiz:
First, make sure the opt during fast_compile keep it.

If we can make it happen more frequently, that is better as we won't request user to rerun the experiments.

lamblin:
It is hard to determine a generic way of how the stack trace should be assigned in an optimization.
Alternative proposal:
- implement a `copy_stack_trace(from=..., to=...)`
- go through all optimizations, and add calls to `copy_stack_trace` for obvious matching nodes.

`copy_stack_trace` could implement some additional logic for ease of understanding, for instance:
- keep multiple stack traces if things are merged together, maybe check how Python 3 exceptions handle multiple stack traces
- keep a log of which optimizations were applied to that node

nouiz:
Why check python3? Do you have any indication that there is something for
multiple stack trace?

lamblin:
I think one new feature of Python3 is that, when an Exception is caught and raised again, the stack trace of all the points where it was raised is kept.

PR #3073, opened 2015-06-26
Initial work on issue #3018.

julianser:
I've submitted some preliminary code. I'm not very sure about my implementation so far, because many optimizations move around the graph is funky ways and call functions I don't understand.
Maybe we can talk about some of this.

Also, some optimizations, which reprocess outputs don't really need to have their stack traces copied over (I think). For example: 

- local_lift_transpose_through_dot
- local_dimshuffle_no_inplace_at_canonicalize
- local_fill_to_alloc (first two returns at least)

These take the inputs to the node and output another tensor variable. The stacktrace should be constructed appropriately through the rest of the code.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -62,6 +62,29 @@ theano.configparser.AddConfigVar('on_shape_error',
 
 # Utilities
 
+def copy_stack_trace(from_var, to_var):
+    """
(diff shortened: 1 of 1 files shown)


lamblin:
Superseded by #3077, which is already merged.

PR #3075, opened 2015-06-26
Make tensor.floor_div a reference to tensor.int_div

carriepl:


nouiz:
I think this is good. We can merge when travis pass.

carriepl:
Looks like some of the unit tests depend heavily on the exact class. I'm looking into it.

PR #3076, opened 2015-06-26
[TEST] fix tests

nouiz:


theano/sandbox/cuda/opt.py
```diff
@@ -1785,8 +1785,10 @@ def local_gpu_downsample_factor_max_grad(node):
...
     if isinstance(node.op, downsample.DownsampleFactorMaxGradGrad):
-        assert node.op.__props__ == ('ds', 'ignore_border', 'st')
-
+        assert node.op.__props__ == ('ds', 'ignore_border', 'st',
+                                     'padding', 'mode')
+        if node.op.padding != (0, 0) or node.op.mode != 'max':
+            return
         x, z, gx = node.inputs
         if (x.owner and isinstance(x.owner.op, HostFromGpu)):
             op = GpuDownsampleFactorMaxGradGrad(node.op.ds,
```
lamblin: We need to pass the new props here as well...

theano/sandbox/test_multinomial.py
```diff
@@ -18,6 +18,8 @@ def get_mode(gpu):
                               'local_gpu_multinomial')
     if isinstance(mode.linker, theano.gof.PerformLinker):
         mode.linker = predefined_linkers['c|py']
+    if not getattr(mode.linker, 'c_thunks', True):
+        mode.linker.c_thunks = True
```
lamblin: Do all linkers have a `c_thunks` attribute?

theano/tensor/tests/test_raw_random.py
```diff
@@ -1189,12 +1189,15 @@ def test_pkl(self):
...
+        if theano.config.mode in ["DEBUG_MODE", "DebugMode"]:
+            mode = 'FAST_COMPILE'
         post_bin_r, bin_sample = binomial(rng_r, (3, 5), 1, .3)
-        f = theano.function([rng_r], [post_bin_r, bin_sample])
+        f = theano.function([rng_r], [post_bin_r, bin_sample], mode=mode)
         pkl_f = pickle.dumps(f)
 
         post_int_r, int_sample = random_integers(rng_r, (3, 5), -1, 8)
-        g = theano.function([rng_r], [post_int_r, int_sample])
+        g = theano.function([rng_r], [post_int_r, int_sample], mode=mode)
```
lamblin: Right, I forgot about that case. Thanks.

theano/sandbox/cuda/opt.py
lamblin: Oh, I did not see that we abort the optimization for unsupported versions of padding and mode.
However, "st" should not be ignored.

theano/sandbox/test_multinomial.py
nouiz: no, only VM_Linker.

theano/sandbox/cuda/opt.py
nouiz: I updated the PR.

theano/sandbox/test_multinomial.py
lamblin: Then, maybe we should not set it to `True` if it did not exist.

nouiz: It is a copy, so we don't care, but I changed it (and added a small commit)

PR #3077, opened 2015-06-26
Further work on issue #3018.

julianser:
I implemented proper stack trace copying for a couple of functions, and changed the printing of the stack trace. Now, a stack trace is list of lists of tuples. 

Sorry, I had to create a new pull request. I completely messed up my other branch trying to do rebase myself.

### DIFF
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -333,13 +333,15 @@ class FunctionGraph(utils.object2):
                             assert path is not None
                             tr = getattr(r.tag, 'trace', None)
                             detailed_err_msg = ""
-                            if tr:
-                                sio = StringIO()
-                                traceback.print_list(tr, sio)
-                                tr = sio.getvalue()
+                            if len(tr) > 0:
                                 detailed_err_msg += "\nBacktrace when the variable is created:\n"
-                                detailed_err_msg += str(tr)
 
+                                # Print separate message for each element in 
+                                # the list of batcktraces
+                                sio = StringIO()
+                                for subtr in tr:
+                                    traceback.print_list(subtr, sio)
+                                    detailed_err_msg += str(sio.getvalue())
                             raise MissingInputError(
                                 'A variable that is an input to the graph was '
                                 'neither provided as an input to the function '
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -331,7 +331,7 @@ class FunctionGraph(utils.object2):
                             # if there is no path then r isn't really a graph input so we shouldn't be running error
                             # handler code in the first place
                             assert path is not None
-                            tr = getattr(r.tag, 'trace', None)
+                            tr = getattr(r.tag, 'trace', [])
                             detailed_err_msg = ""
                             if len(tr) > 0:
                                 detailed_err_msg += "\nBacktrace when the variable is created:\n"
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -336,7 +336,7 @@ class FunctionGraph(utils.object2):
                             if len(tr) > 0:
                                 detailed_err_msg += "\nBacktrace when the variable is created:\n"
 
-                                # Print separate message for each element in 
+                                # Print separate message for each element in
                                 # the list of batcktraces
                                 sio = StringIO()
                                 for subtr in tr:
--- a/theano/gof/link.py
+++ b/theano/gof/link.py
@@ -157,7 +157,7 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
             " have the inputs shapes and strides printed.")
 
     # Print node backtraces
-    tr = getattr(node.outputs[0].tag, 'trace', None)
+    tr = getattr(node.outputs[0].tag, 'trace', [])
     if len(tr) > 0:
         detailed_err_msg += "\nBacktrace when the node is created:\n"
         
--- a/theano/gof/link.py
+++ b/theano/gof/link.py
@@ -160,7 +160,7 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
     tr = getattr(node.outputs[0].tag, 'trace', [])
     if len(tr) > 0:
         detailed_err_msg += "\nBacktrace when the node is created:\n"
-        
+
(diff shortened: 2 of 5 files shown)


theano/gof/fg.py
```diff
@@ -334,13 +334,15 @@ def find_path_to(output_var, input_var):
                             assert path is not None
                             tr = getattr(r.tag, 'trace', None)
                             detailed_err_msg = ""
-                            if tr:
-                                sio = StringIO()
-                                traceback.print_list(tr, sio)
-                                tr = sio.getvalue()
+                            if len(tr) > 0:
```
lamblin: `tr` can be `None` here, so its length is not defined in that case.
The easiest fix would be to change l. 335 to something like `tr = getattr(r.tag, 'trace', [])`.

julianser:
@lamblin It would be nice if we could go through these tomorrow or early next week.

Especially, I'm not sure about local_track_shape_i and local_elemwise_alloc.

julianser:
The test fails because of some white space issues. But other than that everything seems to be OK...

lamblin:
It looks good and tests pass, so I'd like to merge this PR for the moment, then we can look into continuing that work for other optimizations.
Could you rebase on top of the current master to get rid of the "merge" commits in the history?

julianser:
Sounds good. I'll rebase Monday morning, or maybe even tomorrow if I get the time. 

Sorry, I havn't been active this week...

theano/tensor/opt.py
```diff
@@ -519,7 +571,12 @@ def local_lift_transpose_through_dot(node):
     x, y = node.inputs[0].owner.inputs
 
     if x.ndim == y.ndim == 2:
-        return [T.dot(y.T, x.T)]
+        # Output is dot product of transposed inputs in reverse order
+        ret = [T.dot(y.T, x.T)]
+
+        # Copy over stack trace to output from result of dot-product
+        copy_stack_trace(node.inputs[0], ret)
```
lamblin: I think we should actually copy the stack trace from node.outputs[0] instead.

nouiz:
travis have errors.

lamblin:
The errors in travis seem to be pep8 formatting, that should be easy to fix.

julianser:
@lamblin Pascal, can you restart the Scrutinizer? It seems to have crashed the other day...

I already have new commits, but I want you to merge this PR before I continue pushing.

- Julian

lamblin:
I don't seem to have the rights to do that yet, but I'll merge anyway. Thanks!

theano/gof/fg.py
```diff
@@ -331,15 +331,17 @@ def find_path_to(output_var, input_var):
...
                                 detailed_err_msg += "\nBacktrace when the variable is created:\n"
-                                detailed_err_msg += str(tr)
 
+                                # Print separate message for each element in
+                                # the list of batcktraces
+                                sio = StringIO()
+                                for subtr in tr:
+                                    traceback.print_list(subtr, sio)
+                                    detailed_err_msg += str(sio.getvalue())
```
nouiz: sio.getvalue() keep the old content. So this line 344, should be outside the loop.

theano/gof/link.py
```diff
@@ -156,14 +156,16 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
...
+    tr = getattr(node.outputs[0].tag, 'trace', [])
+    if len(tr) > 0:
         detailed_err_msg += "\nBacktrace when the node is created:\n"
-        detailed_err_msg += str(tr)
+
+        # Print separate message for each element in the list of batcktraces
+        sio = StringIO()
+        for subtr in tr:
+            traceback.print_list(subtr, sio)
+            detailed_err_msg += str(sio.getvalue())
```
nouiz: idem

theano/gof/op.py
```diff
@@ -458,14 +458,17 @@ def _get_test_value(cls, v):
...
+                if len(tr) > 0:
                     detailed_err_msg += (
                         " \nBacktrace when that variable is created:\n")
-                    detailed_err_msg += str(tr)
+                    # Print separate message for each element in the list 
+                    # of batcktraces
+                    sio = StringIO()
+                    for subtr in tr:
+                        traceback.print_list(subtr, sio)
+                        detailed_err_msg += str(sio.getvalue())
```
nouiz: idem. Should this be refactored to remove the duplicate code?

theano/tensor/opt.py
```diff
@@ -63,6 +63,54 @@
...
+    of tuples. Each tuple contains the filename, line number, function name
+    and so on. Each list of tuples contains the truples belonging to a
+    particular variable.
+    """
+
+    # Store stack traces from from_var
+    tr = []
+    if type(from_var) is list:
+        # If from_var is a list, store concatenated stack traces
+        if len(from_var) > 0:
```
nouiz: this if is useless. A loop on an empty list would cause 0 iteration of the loop.

theano/tensor/opt.py
```diff
@@ -63,6 +63,54 @@
...
+            for v in from_var:
+                if hasattr(v.tag, 'trace') and len(v.tag.trace) > 0:
```
nouiz: The second condition is not useful. A list + an empty list, return the original list.

theano/tensor/opt.py
```diff
@@ -63,6 +63,54 @@
...
+
+    # Copy over stack traces to to_var
+    if type(to_var) is list:
+        # Copy over stack traces from from_var to each variable in
+        # to_var, including the stack_trace of the to_var before
+        for v in to_var:
+            if hasattr(v.tag, 'trace'):
+                v.tag.trace = v.tag.trace + tr
+            else:
+                v.tag.trace = tr
```
nouiz: the 4 lines above can be simplified as:

`v.tag.trace = getattr(v.tag, 'trace', []) + tr`

theano/tensor/opt.py
nouiz: this comment didn't got answer and wasn't done. Is that normal? At first impression, I think the comment is right.

nouiz:
I didn't checked all change to opt, but I got a few comments.

thanks for this good new feature.

PR #3080, opened 2015-06-26
Port of the cudnn code for the gpuarray backend

abergeron:
This ports the current cudnn code we have for the cuda backend over to the gpuarray one.

I've made it so the first commit just copies all the code and then the subsequent ones adapt it.

Some of the surrounding code had to be fixed also so there is a bit more than just straight adapting.

All of tests/test_dnn.py passes locally.

### DIFF
--- a/theano/sandbox/gpuarray/dnn.py
+++ b/theano/sandbox/gpuarray/dnn.py
@@ -1,0 +1,1749 @@
+import os
+import numpy
+
+import theano
+from theano import Apply, gof, tensor, config, Variable
+from theano.scalar import as_scalar, constant
+from theano.gradient import DisconnectedType, grad_not_implemented
+from theano.gof import Optimizer, local_optimizer, COp
+from theano.gof.type import CDataType, Generic
+from theano.compile import optdb
+from theano.compile.ops import shape_i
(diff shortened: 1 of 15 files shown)


theano/sandbox/gpuarray/dnn.py
```diff
@@ -0,0 +1,1717 @@
...
+
+        top = gpu_contiguous(top)
+
+        d_img = GpuDnnConvGradI()(kerns, top, empty_like(img), desc)
+        d_kerns = GpuDnnConvGradW()(img, top, empty_like(kerns), desc)
+        d_alpha = grad_not_implemented(self, 4, alpha)
+        d_beta = grad_not_implemented(self, 5, beta)
+
+        return [d_img * alpha, d_kerns * alpha, top * beta,
+                DisconnectedType()(), d_alpha, d_beta]
```
nouiz: Why the output is disconected? If we work implace and one of alpha or beta isn't 0 (forgot which one), then it isn't disconnected. Why not return grad_not_implemented? Ok the old back-end also do that.

abergeron: The DisconnectedType is for the descriptor.

Issue #1902, opened 2014-06-05
Opt: Alloc(Alloc(...),...) -> Alloc(...)


nouiz:
We saw this in a graph, where the first alloc had only 1 client, the second alloc:

Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)

Here m was broadcastable on both dimensions, but the optimization can cover more cases.

We need to prove that the shape x is the same for both Alloc, otherwise, we could hide a shape mismatch.

lamblin:
If the shapes cannot be proven to be equal, we should add an Assert node, not simply give up.
This could be a follow-up ticket if you prefer.

nouiz:
done by gh-3132.

PR #3082, opened 2015-06-28
Opt: Alloc(Alloc(...),...) -> Alloc(...) #1902

yaoli:
#1902
local optimization
Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5909,6 +5909,8 @@ def local_merge_alloc(node):
     dims_outer_rev = dims_outer[::-1]
     dims_inner_rev = dims_inner[::-1]
     # check if the pattern of broadcasting is matched, in the reversed ordering
+    # The for loop below will run at most len(dims_inner_rev) times, sufficient
+    # to cover all the cases. 
     for dim_inner, dim_outer in zip(dims_inner_rev, dims_outer_rev):
         if dim_inner != dim_outer:
             if isinstance(dim_inner, Constant) and dim_inner.data == 1:
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5908,9 +5908,10 @@ def local_merge_alloc(node):
     dims_inner = inputs_inner[1:]
     dims_outer_rev = dims_outer[::-1]
     dims_inner_rev = dims_inner[::-1]
-    # check if the pattern of broadcasting is matched, in the reversed ordering
-    # The for loop below will run at most len(dims_inner_rev) times, sufficient
-    # to cover all the cases. 
+    # check if the pattern of broadcasting is matched, in the reversed ordering.
+    # The reverse ordering is needed when an Alloc add an implicit new
+    # broadcasted dimensions to its inputs[0]. Eg:
+    # Alloc(Alloc(m, y, 1, 1), x, y, z, w) -> Alloc(m, x, y, z, w)
     for dim_inner, dim_outer in zip(dims_inner_rev, dims_outer_rev):
         if dim_inner != dim_outer:
             if isinstance(dim_inner, Constant) and dim_inner.data == 1:
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5887,3 +5887,32 @@ register_canonicalize(gof.OpRemove(theano.gradient.disconnected_grad_),
 def local_grad_clip(node):
     if isinstance(node.op, theano.gradient.GradClip):
         return node.inputs
+
+
+@register_canonicalize
+@register_stabilize
+@register_specialize
+@gof.local_optimizer([T.Alloc])
+def local_merge_alloc(node):
+    # This opt takes care of several cases:
+    # Alloc(Alloc(m, x, 1, 1, 1), x, y, z, w) -> Alloc(m, x, y, z, w)
+    # Alloc(Alloc(m, y, 1, 1), x, y, z, w) -> Alloc(m, x, y, z, w)
+    if not isinstance(node.op, T.Alloc):
+        return False
+    if not node.inputs[0].owner or not isinstance(
+            node.inputs[0].owner.op, T.Alloc):
+        return False
+    inputs_outer = node.inputs
+    inputs_inner = node.inputs[0].owner.inputs
+    dims_outer = inputs_outer[1:]
+    dims_inner = inputs_inner[1:]
+    dims_outer_rev = dims_outer[::-1]
+    dims_inner_rev = dims_inner[::-1]
+    # check if the pattern of broadcasting is matched, in the reversed ordering
+    for dim_inner, dim_outer in zip(dims_inner_rev, dims_outer_rev):
+        if dim_inner != dim_outer:
+            if isinstance(dim_inner, Constant) and dim_inner.data == 1:
+                pass
+            else:
+                return False
+    return [T.alloc(inputs_inner[0], *dims_outer)]
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -5443,6 +5443,32 @@ class TestIntDivByOne(unittest.TestCase):
         assert len(divs) == 0
 
 
+def test_local_merge_alloc():
+    x = T.iscalar('x')
+    y = T.iscalar('y')
+    z = T.iscalar('z')
+    w = T.iscalar('w')
+    m = T.fscalar('m')
+    # case 1
+    # Alloc(Alloc(m, x, 1, 1, 1), x, y, z, w) -> Alloc(m, x, y, z, w)
+    output = T.alloc(T.alloc(m, 1, y, 1, 1), x, y, z, w)
+    f = theano.function([m, x, y, z, w], output)
+    topo = f.maker.fgraph.toposort()
+    assert len(topo) == 1
+    assert isinstance(topo[0].op, T.Alloc)
+    o = f(0., 1, 2, 3, 4)
+    assert o.shape == (1, 2, 3, 4)
+
+    # case 2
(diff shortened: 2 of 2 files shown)


nouiz:
You need to rebase this PR on top of central/master.

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,22 @@ def local_add_mul_fusion(node):
 def local_grad_clip(node):
     if isinstance(node.op, theano.gradient.GradClip):
         return node.inputs
+
+@register_canonicalize
+@register_stabilize
+@register_specialize
+@gof.local_optimizer([T.alloc])
```
nouiz: I think it is good to register it at the 3 places. But use [T.Alloc] instead of [T.alloc]

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,22 @@ def local_add_mul_fusion(node):
...
+def local_double_alloc(node):
```
nouiz: rename to local_useless_alloc()

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,22 @@ def local_add_mul_fusion(node):
...
+    # Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)
+    if node.op != T.alloc:
```
nouiz: use "if not isinstance(node.op, T.Alloc):", this is faster.

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,22 @@ def local_add_mul_fusion(node):
...
+        return False
+    if not node.inputs[0].owner or node.inputs[0].owner.op != T.alloc:
```
nouiz: idem, use isinstance.

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,22 @@ def local_add_mul_fusion(node):
...
+        return False
+    if len(node.inputs) != 3 or len(node.inputs[0].owner.inputs) != 3:
```
nouiz: The opt should not be specific to only matrix. Make it more general for any ndim and there can be one or more 1 and it can be at any dimensions.

theano/tensor/tests/test_opt.py
```diff
@@ -5299,6 +5299,15 @@ def test_assert_op_gradient():
...
 
 
+def test_alloc_alloc():
+    # test local optimization: Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)
+    x = T.iscalar()
+    y = T.iscalar()
+    output = T.alloc(T.alloc(0., x, 1), x, y)
+    fg = FunctionGraph([x,y], [output])
+    f = optimize(fg)
+    # TODO: how to compare the graph before and after???
```
nouiz: Compile a function to make sure the result is also the right one.

```
f=theano.function(...)

assert len(topo) == 1
assert isinstance(topo[0].op, T.alloc)
```

yaoli: improved code according to comments

theano/tensor/tests/test_opt.py
```diff
@@ -5299,6 +5299,21 @@ def test_assert_op_gradient():
...
+def test_useless_alloc():
+    # test local optimization: Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)
+    x = T.iscalar('x')
+    y = T.iscalar('y')
+    z = T.iscalar('z')
+    w = T.iscalar('w')
+    m = T.iscalar('m')
+    output = T.alloc(T.alloc(m, 1, y, 1, 1), x, y, z, w)
+    fg = FunctionGraph([m, x, y, z, w], [output])
+    optimize(fg)
```
nouiz: remove that line and the one above.

theano/tensor/tests/test_opt.py
```diff
@@ -5299,6 +5299,21 @@ def test_assert_op_gradient():
...
+    f=theano.function([m,x,y,z,w], output)
+    topo = f.maker.fgraph.toposort()
+    assert len(topo) == 1
+    assert isinstance(topo[0].op, T.Alloc)
```
nouiz: Execute the function and check that the output is the right one.

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,26 @@ def local_add_mul_fusion(node):
 def local_grad_clip(node):
     if isinstance(node.op, theano.gradient.GradClip):
         return node.inputs
+
+@register_canonicalize
+@register_stabilize
+@register_specialize
+@gof.local_optimizer([T.Alloc])
+def local_reduce_double_alloc(node):
```
nouiz: Rename local_useless_alloc

yaoli: The name is already taken

theano/tensor/opt.py
```diff
@@ -5841,3 +5841,26 @@ def local_add_mul_fusion(node):
...
+    # Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)
+    if not isinstance(node.op, T.Alloc):
+        return False
+    if not node.inputs[0].owner or not isinstance(node.inputs[0].owner.op, T.Alloc):
+        return False
+    inputs_outer = node.inputs
+    inputs_inner = node.inputs[0].owner.inputs
+    if len(inputs_outer) != len(inputs_inner):
+        return False
```
nouiz: This is too restrictive restriction. inputs_outer must be longer then inputs_inner. Add a test for that case and update the code bellow to handle it.  Mostly, test this case:

 T.alloc(T.alloc(m, y, 1, 1), x, y, z, w)

theano/tensor/opt.py
nouiz: reduce most of the time mean reduction. So I try to find something else. If you have idea, tell them.

What about: local_merge_alloc? Or local_fusion_alloc? I prefer the first, but this can be discussed.

theano/tensor/opt.py
yaoli: On Mon, Jun 29, 2015 at 3:39 PM, Frédéric Bastien <notifications@github.com>
wrote:

> In theano/tensor/opt.py
> <https://github.com/Theano/Theano/pull/3082#discussion_r33503259>:
>
> > @@ -5841,3 +5841,26 @@ def local_add_mul_fusion(node):
> >  def local_grad_clip(node):
> >      if isinstance(node.op, theano.gradient.GradClip):
> >          return node.inputs
> > +
> > +@register_canonicalize
> > +@register_stabilize
> > +@register_specialize
> > +@gof.local_optimizer([T.Alloc])
> > +def local_reduce_double_alloc(node):
> > +    # Alloc(Alloc(m, x, 1), x, y) -> Alloc(m, x, y)
> > +    if not isinstance(node.op, T.Alloc):
> > +        return False
> > +    if not node.inputs[0].owner or not isinstance(node.inputs[0].owner.op, T.Alloc):
> > +        return False
> > +    inputs_outer = node.inputs
> > +    inputs_inner = node.inputs[0].owner.inputs
> > +    if len(inputs_outer) != len(inputs_inner):
> > +        return False
>
> This is too restrictive restriction. inputs_outer must be longer then
> inputs_inner. Add a test for that case and update the code bellow to handle
> it. Mostly, test this case:
>
> T.alloc(T.alloc(m, y, 1, 1), x, y, z, w)
>
What are you expecting from this example, optimized or not optimized?

nouiz: optimized

yaoli: Do you mean T.alloc(T.alloc(m, x, y, 1, 1), x, y, z, w) instead of
T.alloc(T.alloc(m, y, 1, 1), x, y, z, w)? There is no x in inner alloc, so
how does it get broadcasted in the outer alloc?

nouiz:
This need a rebase.

theano/tensor/opt.py
nouiz: T.alloc(T.alloc(m, y, 1, 1), x, y, z, w) should get optimzed by this opt. The inner alloc broadcast on y, the second on x, z and w.

The if above will cause this opt do not cover this case. You also need to modify code bellow the if to make it work.

yaoli: code updated, and also a rebase to clear up old commits.

lamblin:
Can you fix the failing pep8 tests?

lamblin:
Also, before the PR is merged, you should do a rebase that squashes together the last two commits, so you do not add the conflicts and then remove them.

yaoli:
fixed pep8, and rebased

theano/tensor/tests/test_opt.py
```diff
@@ -5443,6 +5443,32 @@ def test3(self):
...
+    x = T.iscalar('x')
+    y = T.iscalar('y')
+    z = T.iscalar('z')
+    w = T.iscalar('w')
+    m = T.fscalar('m')
+    # case 1
+    # Alloc(Alloc(m, x, 1, 1, 1), x, y, z, w) -> Alloc(m, x, y, z, w)
+    output = T.alloc(T.alloc(m, 1, y, 1, 1), x, y, z, w)
+    f = theano.function([m, x, y, z, w], output)
+    topo = f.maker.fgraph.toposort()
```
lamblin: That test will fail if the optimization you are testing is not included, which is the case in FAST_COMPILE for instance.
Other tests in that file create a special mode that explicitly includes the necessary optimization, for instance:
```python
default_mode = theano.compile.mode.get_default_mode()
opt_mode = default_mode.including("local_...")
...
f = theano.function(..., mode=opt_mode)
...
```
Can you do that here? This should fix the failing test in Travis in FAST_COMPILE.

yaoli:
OK, fixed the incompatibility with FAST_COMPILE

yaoli:
added more code comments

yaoli:
yet more code comments

PR #3083, opened 2015-06-28
Speed up Scan optimizations

caglar:
@lamblin @carriepl @nouiz 
I have done some easy optimizations in the scan_opt.py. Most of the operations that can be done in constant time by using set or dictionary, previously was being done in linear time inside the loop. It was a bit slow because of that. 

There are a few other changes that I want to do as well. 

These changes speedup by more than 100 secs for a computational graph that was taking more than 900 secs to compile.

### DIFF
--- a/theano/scan_module/scan_opt.py
+++ b/theano/scan_module/scan_opt.py
@@ -249,14 +249,14 @@ class PushOutNonSeqScan(gof.Optimizer):
 
         to_remove_set = set()
         to_replace_set = set()
-        to_replace_add = to_replace_set.add
         to_replace_map = OrderedDict()
         nto_replace = 0
 
-        def add_to_replace(y, nto_replace):
-            to_replace_add(y)
-            to_replace_map[y] = nto_replace
-            return nto_replace + 1
+        def add_to_replace(y):
+            to_replace_set.add(y)
+            to_replace_map[y] = add_to_replace.n
+            add_to_replace.n +=1
+        add_to_replace.n = 0
 
         replace_with_in = []
         replace_with_out = []
@@ -317,7 +317,7 @@ class PushOutNonSeqScan(gof.Optimizer):
                 # Step 2. Create variables for replacements
                 for idx, y in enumerate(nd.outputs):
                     y_place_holder = scan_utils.safe_new(y, '_replace')
-                    nto_replace = add_to_replace(y, nto_replace)
+                    add_to_replace(y)
                     replace_with_in.append(y_place_holder)
                     assert isinstance(y, type(nw_outer_node.outputs[idx]))
                     replace_with_out.append(nw_outer_node.outputs[idx])
@@ -332,8 +332,10 @@ class PushOutNonSeqScan(gof.Optimizer):
         existent_nodes = [nd for nd in local_fgraph_topo
                           if nd not in to_remove_set]
         existent_nodes_set = set(existent_nodes)
-        to_keep = []; [to_keep.extend(nd.inputs) for nd in existent_nodes]
-        to_keep_set = set(to_keep)
+
+        to_keep_set = set([])
+        for nd in existent_nodes:
+            to_keep_set.update(nd.inputs)
 
         for out, idx in to_replace_map.items():
             if (# If types are different, conversion Op will be inserted,
@@ -377,7 +379,7 @@ class PushOutNonSeqScan(gof.Optimizer):
                 remove=[node],
                 reason='scanOp_pushout_nonseqs_ops')
             return True
-        elif not to_keep:
+        elif not to_keep_set:
             # Nothing in the inner graph should be kept
             replace_with = OrderedDict()
             for out, idx in to_replace_map.items():
@@ -441,14 +443,14 @@ class PushOutSeqScan(gof.Optimizer):
 
         to_remove_set = set()
         to_replace_set = set()
-        to_replace_add = to_replace_set.add
         to_replace_map = OrderedDict()
         nto_replace = 0
 
-        def add_to_replace(y, nto_replace):
-            to_replace_add(y)
-            to_replace_map[y] = nto_replace
-            return nto_replace + 1
+        def add_to_replace(y):
+            to_replace_set.add(y)
+            to_replace_map[y] = add_to_replace.n
+            add_to_replace.n += 1
+        add_to_replace.n = 0
 
         replace_with_in = []
         replace_with_out = []
@@ -516,7 +518,7 @@ class PushOutSeqScan(gof.Optimizer):
                 # Step 2. Create variables for replacements
                 for idx, y in enumerate(nd.outputs):
                     y_place_holder = scan_utils.safe_new(y, '_replace')
-                    nto_replace = add_to_replace(y, nto_replace)
+                    add_to_replace(y)
                     replace_with_in.append(y_place_holder)
                     replace_with_out.append(nw_outer_node.outputs[idx])
 
@@ -540,7 +542,7 @@ class PushOutSeqScan(gof.Optimizer):
                 new_outer = outside_ins.dimshuffle(new_ord)
                 y = nd.outputs[0]
                 y_place_holder = scan_utils.safe_new(y, '_replace')
-                nto_replace = add_to_replace(y, nto_replace)
+                add_to_replace(y)
                 replace_with_in.append(y_place_holder)
                 replace_with_out.append(new_outer)
 
@@ -561,8 +563,10 @@ class PushOutSeqScan(gof.Optimizer):
         existent_nodes = [nd for nd in local_fgraph_topo
                           if nd not in to_remove_set]
         existent_nodes_set = set(existent_nodes)
-        to_keep = []; [to_keep.extend(nd.inputs) for nd in existent_nodes]
-        to_keep_set = set(to_keep)
+
+        to_keep_set = set([])
+        for nd in existent_nodes:
+            to_keep_set.update(nd.inputs)
 
         for out, idx in to_replace_map.items():
             if (out in to_keep_set
@@ -608,7 +612,7 @@ class PushOutSeqScan(gof.Optimizer):
                 remove=[node],
                 reason='scanOp_pushout_seqs_ops')
             return True
-        elif (not to_keep and
+        elif (not to_keep_set and
               not op.as_while and
               not op.outer_mitmot(node)):
             # Nothing in the inner graph should be kept
--- a/theano/scan_module/scan_opt.py
+++ b/theano/scan_module/scan_opt.py
@@ -113,8 +113,8 @@ def remove_constants_and_unused_inputs_scan(node):
     op = node.op
     # We only need to take care of sequences and other arguments
     st = op.n_seqs
-    st += int(numpy.sum([len(x) for x in
-                         op.tap_array[:(op.n_mit_mot + op.n_mit_sot)]]))
+    st += int(sum([len(x) for x in
+                   op.tap_array[:(op.n_mit_mot + op.n_mit_sot)]]))
     st += op.n_sit_sot
     st += op.n_shared_outs
 
@@ -463,7 +463,7 @@ class PushOutSeqScan(gof.Optimizer):
                     depends_on_seqs = False
 
                     for x in nd.inputs:
-                        if x in inner_non_seqs:
+                        if x in inner_non_seqs_set:
                             _idx = inner_non_seqs_map[x]
                             outside_ins.append(outer_non_seqs[_idx])
                         elif x in inner_seqs_set:
@@ -609,7 +609,7 @@ class PushOutSeqScan(gof.Optimizer):
                     if out in op.inner_mitsot_outs(ls):
                         odx = op.inner_mitsot_outs(ls).index(out)
                         inp = op.outer_mitsot(node)[odx]
-                        st = abs(numpy.min(op.mitsot_taps()))
+                        st = abs(min(op.mitsot_taps()))
                         y = tensor.set_subtensor(inp[st:], _y)
                     elif out in op.inner_sitsot_outs(ls):
                         odx = op.inner_sitsot_outs(ls).index(out)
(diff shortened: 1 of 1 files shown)


caglar:
I don't know why TravisBot gives a syntax error. These changes run without any problem for me and they reduced the compilation time for one of my graph from 1150 secs to 860 secs(by about 300 secs). But still the bottleneck in compilation time is PushoutNonseqScanOpt's.

I wanted to make a PR for the easier optimizations to do first, to least have a temporary relieve on the pain I am having with the compilation right now :). I would like to discuss about a less hackier and more permanent solution that will completely remove the whole problem, in another PR.

nouiz:
For the syntax error, use this:
```
dict([(v,k) for k,v in enumerate(local_fgraph.outputs)])
```

carriepl:
As we discussed IRL, some of these sets/dicts would probably need to be ordered sets/dicts instead, either because the code is already iterating over them or because people might try to.

Also, you have some commits that do X, followed by commits that undo X. You can use an interactive rebase to remove these useless pairs of commits from the PR.

caglar:
@carriepl , you are right. But I have already made to_replace_map as an ordered dictionary. The reason is because I was iterating over that. But I decided to keep the rest to be the same. Because rest of them are not being iterated over. If I make them Ordered the compilation time was increasing noticable about 60 secs more for me and I put a notice on top of the process_node functions to change them to Ordered counterparts, if they want to iterate over these data structures. I will make that comment more noticeable.

I don't think the possibility of iterating over these data structures in the future is enough reason to wait 60 secs more for the compilation. I would rather prefer teach people to read the comments :p.

But if you want I can create a new safe dictionary and safe set classes inheriting from dict such that if the user wants to iterate over these data structures, it will raise a warning, for example:

     class SafeDict(dict):
         def __iter__(self):
              warnings.warn("If you want to iterate over that data structure, use OrderedDict")
          def items(self):
              warnings.warn("If you want to iterate over that data structure, use OrderedDict")
           .
           .
           .

What do you think?

nouiz:
Don't bother with a SafeDict. Few people will modify scan opt.

carriepl:
Fine for the comments then.
Also it seems like the changes in the condition ordering make some tests fail.

caglar:
@carriepl I have squashed the commits, you have mentioned also there were some flake8 errors the travis test was raising on the parts that I haven't touched! Anyways I have fixed some flake8 errors, I was getting when I ran flake8 locally.

theano/scan_module/scan_opt.py
```diff
@@ -393,143 +423,159 @@ def apply(self, fgraph):
...
+        to_remove_add = to_remove_set.add
+        to_replace_set = set()
+        to_replace_add = to_replace_set.add
+        to_replace_map = OrderedDict()
+        nto_replace = 0
+
+        def add_to_replace(y, nto_replace):
+            to_replace_add(y)
+            to_replace_map[y] = nto_replace
+            return nto_replace + 1
```
carriepl: This function should probably not take nto_replace as an input and return it as an output; it not needed and if someone forgets to use the output, he's in big trouble. The fonction can instead use nto_replace directly and modify it inplace. This will make the whole thing a lot easier to use.

theano/scan_module/scan_opt.py
```diff
@@ -393,143 +423,159 @@ def apply(self, fgraph):
...
-        to_replace = []
+        local_fgraph = gof.FunctionGraph(clean_inputs, clean_outputs,
+                                         clone=False)
+        local_fgraph_topo = local_fgraph.toposort()
+        local_fgraph_outs_set = set(local_fgraph.outputs)
+        local_fgraph_outs_map = dict([(v,k) for k,v in \
+                                     enumerate(local_fgraph.outputs)])
+
+        to_remove_set = set()
+        to_remove_add = to_remove_set.add
```
carriepl: This shorthand for to_remove_set.add should probably be removed. It's only ever used once, in the function 'add_to_replace()' and it's only marginally shorter than directly calling to_remove_set.add().

theano/scan_module/scan_opt.py
```diff
@@ -393,143 +423,159 @@ def apply(self, fgraph):
...
+        to_replace_set = set()
+        to_replace_add = to_replace_set.add
```
carriepl: This shorthand for to_replace_set.add should probably be removed. It's only marginally shorter than directly calling to_replace_set.add() and it doesn't really improve readability.

theano/scan_module/scan_opt.py
```diff
@@ -219,106 +228,123 @@ def apply(self, fgraph):
...
+                                         clean_outputs,
+                                         clone=False)
+
+        local_fgraph_topo = local_fgraph.toposort()
+        local_fgraph_outs_set = set(local_fgraph.outputs)
+        local_fgraph_outs_map = dict([(v, k) for k, v in \
+                enumerate(local_fgraph.outputs)])
+
+        to_remove_set = set()
+        to_remove_add = to_remove_set.add
```
carriepl: This shorthand for to_remove_set.add should probably be removed. It's only marginally shorter than directly calling to_remove_set.add() and it doesn't really improve readability.

theano/scan_module/scan_opt.py
```diff
@@ -219,106 +228,123 @@ def apply(self, fgraph):
...
+        to_replace_set = set()
+        to_replace_add = to_replace_set.add
```
carriepl: This shorthand for to_replace_set.add should probably be removed. It's only marginally shorter than directly calling to_replace_set.add() and it doesn't really improve readability.

theano/scan_module/scan_opt.py
```diff
@@ -219,106 +228,123 @@ def apply(self, fgraph):
...
+        to_replace_map = OrderedDict()
+        nto_replace = 0
+
+        def add_to_replace(y, nto_replace):
+            to_replace_add(y)
+            to_replace_map[y] = nto_replace
+            return nto_replace + 1
```
carriepl: This function should probably not take nto_replace as an input and return it as an output; it not needed and if someone forgets to use the output, he's in big trouble. The fonction can instead use nto_replace directly and modify it inplace. This will make the whole thing a lot easier to use.

theano/scan_module/scan_opt.py
```diff
@@ -219,106 +228,123 @@ def apply(self, fgraph):
...
-                    # If types are different, conversion Op will be inserted,
-                    # and it may trigger an infinite loop.
-                    and replace_with_in[idx].type == out.type):
-                clean_to_replace += [out]
-                clean_replace_with_in += [replace_with_in[idx]]
-                clean_replace_with_out += [replace_with_out[idx]]
+        existent_nodes = [nd for nd in local_fgraph_topo
+                          if nd not in to_remove_set]
+        existent_nodes_set = set(existent_nodes)
+        to_keep = []; [to_keep.extend(nd.inputs) for nd in existent_nodes]
```
carriepl: This way of doing it is less readable then putting the desired content in the list `to_keep` than using a for loop because you have multiple instructions on the same line and the approach is not intuitive (using a list comprehension not for the purpose of creating a list). Is this really saving a significant amount of time over a for loop?

theano/scan_module/scan_opt.py
```diff
@@ -351,14 +378,14 @@ def process_node(self, fgraph, node):
                 remove=[node],
                 reason='scanOp_pushout_nonseqs_ops')
             return True
-        elif to_keep == []:
+        elif not to_keep:
```
carriepl: This version reduces readability compared to the previous and I would be surprised if it made a significant impact on performance so it should probably be left as it was. The code for Scan seriously lacks readability so any small thing can make a big difference in that regard.

theano/scan_module/scan_opt.py
caglar: This one was done for speed, but it end up improving the compilation speed only a few secs for a CG that takes half an hour to compile.

theano/scan_module/scan_opt.py
caglar: I think that was recommended by pep8 or pylint(I don't remember which), that was why I changed it. I disagree with your argument for readability for me "not to_keep" is way more readable.

theano/scan_module/scan_opt.py
caglar: I will check it once more, if the speed difference is still only few seconds, I agree that this version is less readable.

carriepl:
Since PushOutSeqScan and PushOutNonSeqScan follow pretty much the same pattern. My comments in any one of them generally apply to both.

Also, did you make sure that both optimization only ever needed one pass on the fgraph and that any additional pass after that was useless?

carriepl:
Also, since this is Scan, it's usually a very good thing to have at least two people have a look. @nouiz or @lamblin can one of you give it a quick look?

caglar:
Thanks @carriepl I will try to check and fix rest of your comments tomorrow afternoon.

theano/scan_module/scan_opt.py
caglar: I agree.

lamblin: Since `nto_replace` is an int, I don't think you can modify it inplace from inside the function.

lamblin: Or you could do something like:
```python
def add_to_replace(y):
    to_replace_add(y)
    to_replace_map[y] = add_to_replace.n
    add_to_replace.n += 1
add_to_replace.n = 0
```

theano/scan_module/scan_opt.py
lamblin: I agree.
On top of that, you do not actually need to use `to_keep`, so you could simply build `to_keep_set` directly in a loop.

theano/scan_module/scan_opt.py
lamblin: @caglar is right, it is a recommendation of PEP8.
In that case, you could actually use `to_keep_set` here instead.

theano/scan_module/scan_opt.py
```diff
@@ -219,106 +228,123 @@ def apply(self, fgraph):
...
-                clean_to_replace += [out]
-                clean_replace_with_in += [replace_with_in[idx]]
-                clean_replace_with_out += [replace_with_out[idx]]
+        existent_nodes = [nd for nd in local_fgraph_topo
+                          if nd not in to_remove_set]
+        existent_nodes_set = set(existent_nodes)
+        to_keep = []; [to_keep.extend(nd.inputs) for nd in existent_nodes]
+        to_keep_set = set(to_keep)
+
+        for out, idx in to_replace_map.items():
```
lamblin: The indices are sequential here, right?
In that case, you can make `to_replace_map` an regular (un-ordered) dictionary, and also build the `to_replace` list at the same time, and iterate here over `enumerate(to_replace)` as before.
I'm not sure how big the speed benefit would be (if any) though.

lamblin:
Apart from that and @carriepl's comments, I think it's OK.

carriepl:
@caglar How is it going? Are you running into any issue?

caglar:
Hey PL,

Thanks I did some profiling on the code on last friday, I have updated some
parts of the code pointed during the discussion. Now I am trying to finish
my NIPS reviews. As soon as they are done, I will try to finish this
ticket, probably by tomorrow.

Thanks,

On Wed, Jul 22, 2015 at 1:10 PM, carriepl <notifications@github.com> wrote:

> @caglar <https://github.com/caglar> How is it going? Are you running into
> any issue?
>
> —
> <https://github.com/Theano/Theano/pull/3083#issuecomment-123794789>.
>



-- 
Caglar GULCEHRE

theano/scan_module/scan_opt.py
caglar: I do not think there will be a big speed up and this is not inside the big loop, hence this is not the main bottleneck. Plus we will have to keep track of and maintain a one more additional list just for that.

caglar:
Hi @lamblin @carriepl ,
So I have integrated both your and Pascal's suggestions. This version is only about 7 secs slower for a CG that takes 20 mins to compile.  I am redoing line_profile to see where the main bottleneck now.

carriepl:
Looks good. Did you make sure. like I asked, about the 2nd and ulterior pass through the graph not being required in PushOutSeqScan and PushOutNonSeqScan?

carriepl:
It really needs to be verified before this is merged

lamblin:
According to his earlier message https://github.com/Theano/Theano/pull/3083#issuecomment-122027973, he checked and only one pass was executed.

carriepl:
Thanks @lamblin, I hadn't seen that.

carriepl:
I guess it should be fine, then. Thanks.

theano/scan_module/scan_opt.py
```diff
@@ -205,7 +211,10 @@ def remove_constants_and_unused_inputs_scan(node):
 # This is a global opt for historical reason
 # It should be possible to change it to a local opt.
 class PushOutNonSeqScan(gof.Optimizer):
-
+    """
+    A global optimizer for pushing out the variables inside the scan that
+    are not used by the scan.
```
nouiz: I don't like this docstring. The computation can be used by other part of the inner scan graph. What about:

    A global optimizer for pushing out the variables inside the scan that depend only on non-sequence.

theano/scan_module/scan_opt.py
```diff
@@ -379,7 +407,10 @@ def process_node(self, fgraph, node):
 # This is a global opt for historical reason
 # It should be possible to change it to a local opt.
 class PushOutSeqScan(gof.Optimizer):
-
+    """
+    A global optimizer for pushing out the input variables that are not being
+    used inside the scan and provided in the sequences.
```
nouiz: Those variable are used by scan. What about this

    A global optimizer for pushing out some input sequences variables.

caglar: @nouiz I have fixed the both of the docstrings that you have pointed out, there:

https://github.com/Theano/Theano/pull/3216

Issue #2963, opened 2015-05-26
example Op.__props__ not empty


nouiz:
We could base it on the Choose clas in tensor/basic.py

lamblin:
Fixed in #3085

PR #3085, opened 2015-06-29
a proper rebased version of the modifications for my CCW modifications

gyom:
Issue #2963.

Added the modifications for the __props__ attribute of theano ops.

This should override what happened in https://github.com/Theano/Theano/pull/3061 (which should now be discarded).


doc/tutorial/extending_theano.txt
```diff
@@ -259,7 +259,7 @@ Op Example
 
         def make_node(self, x):
             # check that the theano version has support for __props__
-            assert hasattr(self, '_props')
+            assert hasattr(self, '__props__')
```
nouiz: Revert to the old version.

PR #2665, opened 2015-03-24
New conv2d interface (work in progress)

ballasn:
Add placeholder classes for 2d conv to avoid call to conv3d2d when using gpu implementation.

PR #3089, opened 2015-06-29
[WIP] CpuCorrMM closes #3026

### LINKED ISSUE #3026: Implement CpuCorrMM
We should have an implementation of CPU convolution (including gradient with subsampling) based on GEMM, like GpuCorrMM on GPU.
That way, when the abstract convolution op is finished, we will have a fast implementation of all cases on CPU.

JesseLivezey:
CPU implementation of the CorrMM style convolutions for #3026.

To make sure I understand all of the steps and don't do anything unnecessary, here is my current understanding of what I need to do.

General
- [x] Write real tests, current 'tests' are just copied from tensor.nnet, not sure if they make any sense
- [x] Get tests to pass
- [ ] Move code out of sandbox and into final location (tensor.nnet?)

C code
 - [x] write im2col and col2im for cpu
 - [x] write cormm for cpu
 - [x] float 32 and 64 versions

Theano
- [x] write ops
 - [x] BaseCpuCorrMM
 - [x] CpuCorrMM
 - [x] CpuCorrMM_gradWeights
 - [x] CpuCorrMM_gradInputs

Later PRs
- Write optimization to swap conv ops (waiting on #2665)
- Add 'same' mode for cpu and gpu versions


JesseLivezey:
@lamblin or @nouiz, could one of you comment on my high level todo list? Besides the im2col and col2im functions, the code is a mess, so don't bother looking at it yet.

lamblin:
The high-level plan looks good!
Regarding some specific points:
- Most of the existing CPU tests make sense in general, but many of them will be redundant in your case (testing different kinds of loop unrolling, for instance).
- For examples of use of s/dgemm, you can have a look at theano/tensor/blas.py and theano/tensor/blas_c.py. In particular, both get the blas headers from theano/tensor/blas_headers.py.
- For optimizations, it will probably be wiser to wait for #2665 to be merged, since it will have individual Ops to represent each of the 3 operations. Optimizations for the gradient operations (when the convolution is strided) are unfortunately unfeasible for the moment.

Thanks a lot!

JesseLivezey:
The current tests fail with
```
ERROR: Tests that basic convolutions work for odd and even
  File "/home/jesse/Development/Theano/theano/sandbox/cpu_corr/tests/test_conv.py", line 140, in test_basic
    self.validate((3, 2, 8, 8), (4, 2, 5, 5), 'valid', verify_grad=False)
  File "/home/jesse/Development/Theano/theano/sandbox/cpu_corr/tests/test_conv.py", line 69, in validate
    theano_conv = theano.function([input, filters], output, mode=self.mode)
  File "/home/jesse/Development/Theano/theano/compile/function.py", line 310, in function
    output_keys=output_keys)
  File "/home/jesse/Development/Theano/theano/compile/pfunc.py", line 523, in pfunc
    output_keys=output_keys)
  File "/home/jesse/Development/Theano/theano/compile/function_module.py", line 1526, in orig_function
    defaults)
  File "/home/jesse/Development/Theano/theano/compile/function_module.py", line 1390, in create
    input_storage=input_storage_lists)
  File "/home/jesse/Development/Theano/theano/gof/link.py", line 606, in make_thunk
    output_storage=output_storage)[:3]
  File "/home/jesse/Development/Theano/theano/gof/vm.py", line 1025, in make_all
    no_recycling))
  File "/home/jesse/Development/Theano/theano/gof/op.py", line 805, in make_thunk
    no_recycling)
  File "/home/jesse/Development/Theano/theano/gof/op.py", line 732, in make_c_thunk
    output_storage=node_output_storage)
  File "/home/jesse/Development/Theano/theano/gof/cc.py", line 1068, in make_thunk
    keep_lock=keep_lock)
  File "/home/jesse/Development/Theano/theano/gof/cc.py", line 1010, in __compile__
    keep_lock=keep_lock)
  File "/home/jesse/Development/Theano/theano/gof/cc.py", line 1437, in cthunk_factory
    key=key, lnk=self, keep_lock=keep_lock)
  File "/home/jesse/Development/Theano/theano/gof/cmodule.py", line 1055, in module_from_key
    module_hash = get_module_hash(src_code, key)
  File "/home/jesse/Development/Theano/theano/gof/cmodule.py", line 411, in get_module_hash
    return hash_from_code('\n'.join(to_hash))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 38: ordinal not in range(128)
```

I'm not sure what that means or how to fix it.

lamblin:
It looks like a non-ascii character in the source code. I would guess it is from "Frédéric" or "Schlüter" in the comments.
I'm not sure how the source file is encoded (it should be utf8 I think), but you could probably use the "decode" method of Python strings to get a unicode string. I'm not that familiar with python's new string/unicode conversions and good practices, though.
I doubt @nouiz or @f0k would take offence if they got credited as "Frederic Bastien" and "Jan Schlueter" respectively, though :)

nouiz:
We do that (remove accent) at a few place for that. This won't be new and I
don't take offence of it:) I do this myself at a few places.

f0k:
About the conversion, are you actually converting the code by hand? You can just get the CPU version from the same source (Caffe), they have everything implemented both for GPU and CPU. I've put some links to the respective GPU code in the GpuCorrMM implementation, I think you mostly just need to change `.cu` to `.c` to get to their CPU implementations.

theano/sandbox/cpu_corr/blas.py
```diff
@@ -0,0 +1,499 @@
...
+        assert inputs[1] == filters[1]
+        # nb mul and add by output pixel
+        flops = filters[2] * filters[3] * 2
+        # nb flops by output image
+        flops *= outputs[2] * outputs[3]
+        # nb patch multiplied
+        flops *= inputs[1] * filters[0] * inputs[0]
+        return flops
+
+    def c_support_code(self):
```
JesseLivezey: I basically just added everything from the original GpuCorrMM and blas.py from ```c_support_code``` down to ```c_code_helper```, but I'm not very familiar with what they all do. Am I missing anything obvious or including things it doesn't need.

Currently, I'm not using the [COp](http://deeplearning.net/software/theano/tutorial/extending_theano_c.html#alternate-way-of-defining-c-ops) interface since I'm trying to keep things similar to the GPU version. Is this reasonable?

theano/sandbox/cpu_corr/corr_gemm.cpp
```diff
@@ -0,0 +1,490 @@
...
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#undef _GLIBCXX_ATOMIC_BUILTINS
+
+
+// (borrowed from Caffe: https://github.com/BVLC/caffe/blob/master/src/caffe/util/im2col.cpp)
+// Loops for fast unfold + copy
+void im2col(const npy_float* data_im, const int channels,
```
JesseLivezey: I'm not sure what the correct type is for data_im and data_col. Should they be npy_floats types or just regular C floats?

Also, rather than writing two versions of this for double and float, maybe it makes more sense to load the text differently and do a few string substitutions. Could also do this for {d,s}gemm.

JesseLivezey:
@lamblin, removing those characters fixed it, although I'm curious what I'm doing differently since they are also in the GPUCorrMM files. I've added a few questions I have in the code comments.

@f0k, I did use the Caffe CPU code as a reference. Most of the work so far has been converting the code that wraps the im2col/col2im and blas calls.

theano/sandbox/cpu_corr/blas.py
```diff
@@ -0,0 +1,500 @@
...
+    if ( !(%(out)s
+           && PyArray_NDIM(%(out)s)==4
+           && PyArray_ISCONTIGUOUS(%(out)s)
+           && PyArray_DIMS(%(out)s)[0]==out_dim[0]
+           && PyArray_DIMS(%(out)s)[1]==out_dim[1]
+           && PyArray_DIMS(%(out)s)[2]==out_dim[2]
+           && PyArray_DIMS(%(out)s)[3]==out_dim[3]))
+    {
+        Py_XDECREF(%(out)s);
+        %(out)s = (PyArrayObject*)PyArray_EMPTY(4,
```
JesseLivezey: As is, I believe this allocation causes a segfault. I had some trouble figuring out how to declare out_dim correctly.

Any ideas?

theano/sandbox/cpu_corr/corr_gemm.cpp
JesseLivezey: I think I've figured this out. I just needed to get the pointer to the data from PyArray_DATA  in the corrMM function.

theano/sandbox/cpu_corr/blas.py
f0k: I didn't know the COp interface, I just continued from what @stencilman had started in #2002. Not sure if there would be an advantage in refactoring it to be based on COp, I think it's fine to just use about the same Python code as for GpuCorrMM.

What I was referring to was that https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/corr_gemm.cu is based on https://github.com/BVLC/caffe/blob/master/src/caffe/layers/conv_layer.cu, and when porting this to CPU you might want to grab some parts from https://github.com/BVLC/caffe/blob/master/src/caffe/layers/conv_layer.cpp instead of translating the kernels yourself (acknowledging the source and original license as I did).

theano/sandbox/cpu_corr/corr_gemm.cpp
nouiz: I would continue to use npy_float*, in this case for consistency. Otherwise, npy_intN is more clear and portable then other int type.

Agree, string substitution would be useful. Possible with the new COp interface.

theano/sandbox/cpu_corr/blas.py
nouiz: I think that out_dim is good. But The 3rd parameter should be PyArray_DTYPE(...)

theano/sandbox/cpu_corr/blas.py
nouiz: Cop interface is more recent the the GpuCorrMM. You do not need to use it, but you can. It is your choise. Not using it could be simpler as you base the new code on existing code that don't use it.

theano/sandbox/cpu_corr/corr_gemm.c
```diff
@@ -0,0 +1,417 @@
...
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#undef _GLIBCXX_ATOMIC_BUILTINS
```
JesseLivezey: I'm not sure what this does, but it was in the GPU code, so I left it.

theano/sandbox/cpu_corr/tests/test_conv.py
```diff
@@ -0,0 +1,388 @@
...
+            if not should_raise:
+                raise
+            return
+        else:
+            if should_raise:
+                raise Exception(
+                "ConvOp should have generated an error")
+
+        ############# REFERENCE IMPLEMENTATION ############
+        # Testing correlation, not convolution. Reverse filters.
```
JesseLivezey: Since this is a correlation and the test is for a convolution, I just flip the filter shapes here.

JesseLivezey:
Getting close to something that works. Currently, I think the grad ops are not working. When I run tests that check gradients, the ```all_close``` assertions fail and I get a bunch of:
```
*** Reference count error detected: 
an attempt was made to deallocate 11 (f) ***
```

theano/sandbox/cpu_corr/tests/test_conv.py
nouiz: good

nouiz:
this error indicate a problem with refcount. Mostly, you did Py_DECREF too frequently in this case. If you did Py_INCREF too frequetly, this would cause a memory leak.

theano/sandbox/cpu_corr/corr_gemm.c
nouiz: this isn't needed on the CPU I think.

nouiz:
On travis, the compilation is crashing as you use: PyArray_DTYPE, the good name is: PyArray_TYPE. The version you use only exist since numpy 1.7 and we support older version.

Then you need to change PyArray_Empty to PyArray_EMPTY, as _TYPE return a typenum and not the object.

For the ref count problem, I checked rapidly the code and it seem ok. So without seing more about the error you show, I can't tell more for now.

theano/sandbox/cpu_corr/blas.py
```diff
@@ -0,0 +1,501 @@
...
+           && PyArray_NDIM(%(out)s)==4
+           && PyArray_IS_C_CONTIGUOUS(%(out)s)
+           && PyArray_DIMS(%(out)s)[0]==out_dim[0]
+           && PyArray_DIMS(%(out)s)[1]==out_dim[1]
+           && PyArray_DIMS(%(out)s)[2]==out_dim[2]
+           && PyArray_DIMS(%(out)s)[3]==out_dim[3]))
+    {
+        Py_XDECREF(%(out)s);
+        if (direction != 1) {
+          dtype = PyArray_DTYPE(weights);
```
lamblin: I think you have to incref `dtype` here, as `PyArray_Empty` will steal a reference to it.
Doing that should fix the refcount error about deallocating "11 (f)".

theano/sandbox/cpu_corr/blas.py
```diff
@@ -0,0 +1,501 @@
...
+        }
+        else {
+          dtype = PyArray_DTYPE(bottom);
```
lamblin: and here as well.
Or just incref after the if/else block, actually.

JesseLivezey:
Changing to TYPE and EMPTY also fixed the refcount problem.

Currently, the forward and grad_weights directions seem to always work. The grad_inputs direction fails for certain shape inputs and filter, so I'm still trying to track down that bug.

JesseLivezey:
@nouiz and @lamblin, I'm at a stage where every test I understand is passing. The next thing I need to do is get both float32/64 variants working. Before I start that, I have a few questions.

1. I removed a bunch of tests that were checking the behavior if the shape at compile and runtime were the different. Since this op doesn't need to know the shapes up front, I removed these tests. Is that reasonable?
2. In the same vein, are there any sorts of tests that I'm obviously missing? It seems like there is good coverage of the CorrOp and the derivatives for different options.
3. The one test that is failing is this one:
```======================================================================
FAIL: test_infer_shape (theano.sandbox.cpu_corr.tests.test_conv.TestCorr2D)
  File "/Users/jesse/Development/Theano/theano/sandbox/cpu_corr/tests/test_conv.py", line 248, in test_infer_shape
    [adtens_val, bdtens_val], conv.CpuCorrMM)
  File "/Users/jesse/Development/Theano/theano/tests/unittest_tools.py", line 245, in _compile_and_check
    assert not any(isinstance(t.op, cls) for t in topo_shape)
AssertionError

```
I don't understand what is going on in ```_compile_and_check```, so I can't tell if there is still a bug or if the test is irrelevant.
4. I don't understand the difference between the blas and blas_c versions and the corresponding sgemm_ and cblas_sgemm. I'm currently using blas and sgemm_. Can I just use one, or do I need to check something and use one or the other?

f0k:
> I don't understand what is going on in `_compile_and_check`, so I can't tell if there is still a bug or if the test is irrelevant.

Not sure if it helps, but if you run `nosetests --pdb -s`, it will drop you into a `pdb` prompt on a failure so you can investigate the error more closely (e.g., do `l` to list the source code around the error, `bt` for a backtrace, `p topo_shape` to print `topo_shape`). Also have a look at the `infer_shape` test, just from the name it seems it tests for proper shape inference. I did a [shape inference implementation for ConvOp](https://github.com/Theano/Theano/blob/c2cd739bff9668e2061a011097bb7cf4d4423a9b/theano/tensor/nnet/conv.py#L253-L271) once, maybe you'd need to copy or reuse that?

JesseLivezey:
@f0k, that's helpful.

```topo_shape``` comes from ```shapes_function.maker.fgraph.toposort()``` and is
```
[CpuCorrMM{valid, (1, 1)}(<TensorType(float64, 4D)>, <TensorType(float64, 4D)>), Shape_i{3}(CpuCorrMM{valid, (1, 1)}.0), Shape_i{2}(CpuCorrMM{valid, (1, 1)}.0), Shape_i{1}(CpuCorrMM{valid, (1, 1)}.0), Shape_i{0}(CpuCorrMM{valid, (1, 1)}.0), MakeVector{dtype='int64'}(Shape_i{0}.0, Shape_i{1}.0, Shape_i{2}.0, Shape_i{3}.0)]
```.

I haven't figured out what the value of toposort() should be, where it gets computed, and why it's bad that CpuCorrMM is still in it (first element). Still working through it...

lamblin:
`toposort()` returns the list of Ops in a function graph, in a valid topological order (from inputs to outputs).
In this case, it is called on some function's `.maker.fgraph` attribute, which is the optimized function graph after the compilation of a function.
That function in particular computes the shape of the outputs of a given apply node, given the inputs of that node. This is what is defined in `_compile_and_check`. The goal of that test is to make sure that it is possible to compute the *shape* of `CpuCorrMM(...)` without having to compute `CpuCorrMM(...)` itself, if all we care about is the shape.
It is bad that `CpuCorrMM` is still in the graph of the compiled function, because it means that `shape(CpuCorrMM(...))` was not replaced by an expression depending only on the shapes of the inputs, and the "optimized" function still computes the whole convolution just to get the shape of its output.

nouiz:
So mostly, the op infer_shape method isn't implemented.

lamblin:
The only remaining failures of Travis seem to be:
- an import error on python 3 that prevents the import of `theano`, and
- a couple of pep8 / flake8 issues due to formatting.

The first one could mask more issues with python 3, though.
That looks really good, thanks!

JesseLivezey:
@lamblin, the commit history on this branch has gotten into a state I don't understand. The github commit history is different from my local one and I can't seem to reconcile them. I've started a new PR at #3285.

PR #3091, opened 2015-06-30
gpuarray in new back-end

nouiz:
This is a follow up of gh-3080.

What is still needed is refactor/move the misssing conv cudnn tests.


nouiz:
As this PR add stuff to the new gpuarray and this is converted code. I'll merge myself.

Issue #2971, opened 2015-05-27
GSoC: Compiling time optimizations to do


t13m:
This is a todo list of optimizations to make theano compile faster. Any idea and discussion is welcome! Thanks.
- [x] Remove unnecessary gpu_from_host(host_from_gpu(x)) ([ref](https://groups.google.com/d/msg/theano-dev/BiyppZg9ykA/IXpjNsCN7x0J))
- [x] Modify EquilibriumOptimizer to do less pass over the graph ([ref](https://groups.google.com/d/msg/theano-dev/BiyppZg9ykA/iPx7qLhwVm0J))
- [x] Make local_dimshuffle_list lift through many elemwise at once.
- [ ] Speed up inplace elemwise optimizer (issue #2962)
  - [x] Make some function inline
- [ ] This need discussion with other dev before being done. Make var.dimshuffle and var.reshape return the GPU op directly. This will remove work from the GPU. For this, go in the file cuda/var.py and create a new gpu_operator class that inherit from _operator and make CudaNdarrayVariable and CudaNdarrayConstant inherit from it, but not CudaNdarraySharedVariable.
- [x] Make local_fill_sink more efficient ([ref](https://groups.google.com/d/msg/theano-dev/BiyppZg9ykA/iPx7qLhwVm0J)) (not merged yet)
- [x] Merge nodes with input of assert node.
- [x] [Debug flag that print all rejected optimization](https://github.com/Theano/Theano/issues/2616)
- [x] Remove `merge3` (can't be done, it is needed)

later
- [ ] Find where the time spend during theano.function outside of linker and optimizer.

nouiz:
@lamblin @abergeron what do you think of the 5th point that I added? I still want the graph to be build on the GPU, so I don't want to change what is done on shared variable. But if the user or during optimization we do method directly on a gpu variable, I think we could return directly the gpu op.

To test this and make sure it do not cause unforeseen problem, I would test it with dimshuffle and reshape.

lamblin:
Do you mean that you "still want the graph to be built on the **CPU**" instead?
If so, that makes sense, although I'm not sure how much work is actually being done by the optimizer in that step.

nouiz:
I mean, currently we var.dimshuffle will always create a dimshuffle on the
CPU. Even if var is a CudaNdarrayVariable.

What I want is that var.dimshuffle is a GpuDimShuffle when var is a
CudaNdarrayVariable.

This make optimization that build some graph like that do the stuff
[...]

PR #3092, opened 2015-06-30
Optimize local_fill_sink

t13m:
make it call itself recursively.
see issue #2971

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1390,7 +1390,7 @@ def local_fill_sink(node):
             r = local_fill_sink.transform(new_client.owner)
             if r is False:
                 continue
-            replacements.pop(node.outputs[0], None)
+            # replacements.pop(node.outputs[0], None)
             replacements.update(r)
     return replacements
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1377,6 +1377,7 @@ def local_fill_sink(node):
     # The newly created node c doesn't has 'clients',
     # so this iteration is took place with node.outputs[0]
     replacements = {node.outputs[0]: c}
+    all_clients_replaced = True
     for client, cl_idx in node.outputs[0].clients:
         if (hasattr(client, 'op') and
                 isinstance(client.op, T.Elemwise) and
@@ -1388,10 +1389,14 @@ def local_fill_sink(node):
             # Add clients to new_client
             new_client.owner.outputs[0].clients = client.outputs[0].clients
             r = local_fill_sink.transform(new_client.owner)
-            if r is False:
+            if not r:
+                all_clients_replaced = False
                 continue
-            # replacements.pop(node.outputs[0], None)
             replacements.update(r)
+        else:
+            all_clients_replaced = False
+    if all_clients_replaced:
+        replacements.pop(node.outputs[0], None)
     return replacements
 
 register_canonicalize(local_fill_sink)
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3576,7 +3576,9 @@ def local_fill_sink(node):
 
     f need to be an elemwise
     """
-    if not isinstance(node.op, T.Elemwise) or node.op == T.fill:
+    if (not hasattr(node, 'op') or
+            not isinstance(node.op, T.Elemwise) or
+            node.op == T.fill):
         return False
     models = []
     inputs = []
@@ -3591,6 +3593,18 @@ def local_fill_sink(node):
     c = node.op(*inputs)
     for model in models:
         c = T.fill(model, c)
+
+    # The newly created node c doesn't has 'clients',
+    # so this iteration is took place with node.outputs[0]
+    for client, _ in node.outputs[0].clients:
+        if (hasattr(client, 'op') and
+                isinstance(client.op, T.Elemwise) and
+                not client.op == T.fill):
+            node_sub_c = client.op([v for v in client.inputs if v is not c])
+            # import ipdb; ipdb.set_trace()
+            r = local_fill_sink.transform(node_sub_c)
+            if r:
+                return {client: r}
     return [c]
 
 register_canonicalize(local_fill_sink)
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1349,6 +1349,54 @@ theano.compile.mode.optdb.register('ShapeOpt', ShapeOptimizer(),
                                    0.1, 'fast_run', 'fast_compile')
 
 
+@gof.local_optimizer([T.Elemwise])
+def local_fill_sink(node):
+    """
+    f(fill(a, b), fill(c, d), e) -> fill(a, fill(c, f(b, d, e)))
+
+    f need to be an elemwise
+    """
+    if (not hasattr(node, 'op') or
+            not isinstance(node.op, T.Elemwise) or
+            node.op == T.fill):
+        return False
+    models = []
+    inputs = []
+    for input in node.inputs:
+        if input.owner and input.owner.op == T.fill:
+            models.append(input.owner.inputs[0])
+            inputs.append(input.owner.inputs[1])
+        else:
+            inputs.append(input)
+    if not models:
+        return False
+    c = node.op(*inputs)
+    for model in models:
+        c = T.fill(model, c)
+
+    # The newly created node c doesn't has 'clients',
+    # so this iteration is took place with node.outputs[0]
+    replacements = {node.outputs[0]: c}
+    for client, cl_idx in node.outputs[0].clients:
+        if (hasattr(client, 'op') and
+                isinstance(client.op, T.Elemwise) and
+                not client.op == T.fill):
+            client_inputs = client.inputs[:]
+            client_inputs[cl_idx] = c
+            new_client = client.op(*client_inputs)
+
+            # Add clients to new_client
+            new_client.owner.outputs[0].clients = client.outputs[0].clients
+            r = local_fill_sink.transform(new_client.owner)
+            if r is False:
+                continue
+            replacements.pop(node.outputs[0], None)
+            replacements.update(r)
+    return replacements
+
+register_canonicalize(local_fill_sink)
+
+
 @register_specialize
 @register_stabilize
 @register_canonicalize
@@ -3568,53 +3616,6 @@ register_canonicalize(local_fill_cut)
 
 register_canonicalize(gof.OpRemove(T.tensor_copy), name='remove_tensor_copy')
 
-
-@gof.local_optimizer([T.Elemwise])
-def local_fill_sink(node):
-    """
-    f(fill(a, b), fill(c, d), e) -> fill(a, fill(c, f(b, d, e)))
-
-    f need to be an elemwise
-    """
-    if (not hasattr(node, 'op') or
-            not isinstance(node.op, T.Elemwise) or
-            node.op == T.fill):
-        return False
-    models = []
-    inputs = []
-    for input in node.inputs:
-        if input.owner and input.owner.op == T.fill:
-            models.append(input.owner.inputs[0])
-            inputs.append(input.owner.inputs[1])
-        else:
-            inputs.append(input)
-    if not models:
-        return False
-    c = node.op(*inputs)
-    for model in models:
-        c = T.fill(model, c)
-
-    # The newly created node c doesn't has 'clients',
-    # so this iteration is took place with node.outputs[0]
-    replacements = {node.outputs[0]: c}
-    for client, cl_idx in node.outputs[0].clients:
-        if (hasattr(client, 'op') and
-                isinstance(client.op, T.Elemwise) and
-                not client.op == T.fill):
-            client_inputs = client.inputs[:]
-            client_inputs[cl_idx] = c
-            new_client = client.op(*client_inputs)
-
-            # Add clients to new_client
-            new_client.owner.outputs[0].clients = client.outputs[0].clients
-            r = local_fill_sink.transform(new_client.owner)
-            if r is False:
-                continue
-            replacements.pop(node.outputs[0], None)
-            replacements.update(r)
-    return replacements
-
-register_canonicalize(local_fill_sink)
 
 
 ################
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3603,8 +3603,10 @@ def local_fill_sink(node):
             # node_sub_c = client.op([v for v in client.inputs if v is not c])
             # import ipdb; ipdb.set_trace()
             r = local_fill_sink.transform(client)
-            if r:
-                return {client: r}
+            if isinstance(r, list):
+                return {client.outputs[0]: r[0]}
+            elif isinstance(r, dict):
+                return r
     return [c]
 
 register_canonicalize(local_fill_sink)
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -2031,11 +2031,6 @@ local_elemwise_alloc = register_specialize(
         local_elemwise_alloc_op(T.Elemwise, T.Alloc, T.DimShuffle)),
     'local_alloc_elemwise')
 
-local_elemwise_alloc = register_canonicalize(
-    gof.local_optimizer([T.Elemwise])(
-        local_elemwise_alloc_op(T.Elemwise, T.Alloc, T.DimShuffle)),
-    'local_alloc_elemwise')
-
 theano.configparser.AddConfigVar('experimental.local_alloc_elemwise',
                                  "DEPRECATED: If True, enable the experimental"
                                  " optimization local_alloc_elemwise."
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3596,18 +3596,18 @@ def local_fill_sink(node):
 
     # The newly created node c doesn't has 'clients',
     # so this iteration is took place with node.outputs[0]
+    replacements = {node.outputs[0]: c}
     for client, _ in node.outputs[0].clients:
         if (hasattr(client, 'op') and
                 isinstance(client.op, T.Elemwise) and
                 not client.op == T.fill):
-            # node_sub_c = client.op([v for v in client.inputs if v is not c])
             # import ipdb; ipdb.set_trace()
             r = local_fill_sink.transform(client)
-            if isinstance(r, list):
-                return {client.outputs[0]: r[0]}
-            elif isinstance(r, dict):
-                return r
-    return [c]
+            if r is False:
+                continue
+            replacements.pop(node.outputs[0], None)
+            replacements.update(r)
+    return replacements
 
 register_canonicalize(local_fill_sink)
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3600,9 +3600,9 @@ def local_fill_sink(node):
         if (hasattr(client, 'op') and
                 isinstance(client.op, T.Elemwise) and
                 not client.op == T.fill):
-            node_sub_c = client.op([v for v in client.inputs if v is not c])
+            # node_sub_c = client.op([v for v in client.inputs if v is not c])
             # import ipdb; ipdb.set_trace()
-            r = local_fill_sink.transform(node_sub_c)
+            r = local_fill_sink.transform(client)
             if r:
                 return {client: r}
     return [c]
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1372,7 +1372,8 @@ def local_fill_sink(node):
         return False
     c = node.op(*inputs)
     for model in models:
-        c = T.fill(model, c)
+        if model.type != c.type:
+            c = T.fill(model, c)
 
     # The newly created node c doesn't has 'clients',
     # so this iteration is took place with node.outputs[0]
@@ -3623,11 +3624,10 @@ register_canonicalize(local_fill_cut)
 
 register_canonicalize(gof.OpRemove(T.tensor_copy), name='remove_tensor_copy')
 
-
-
 ################
 # Canonization #
 ################
+
 
 class Canonizer(gof.LocalOptimizer):
     """
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3597,12 +3597,17 @@ def local_fill_sink(node):
     # The newly created node c doesn't has 'clients',
     # so this iteration is took place with node.outputs[0]
     replacements = {node.outputs[0]: c}
-    for client, _ in node.outputs[0].clients:
+    for client, cl_idx in node.outputs[0].clients:
         if (hasattr(client, 'op') and
                 isinstance(client.op, T.Elemwise) and
                 not client.op == T.fill):
-            # import ipdb; ipdb.set_trace()
-            r = local_fill_sink.transform(client)
+            client_inputs = client.inputs[:]
+            client_inputs[cl_idx] = c
+            new_client = client.op(*client_inputs)
+
+            # Add clients to new_client
+            new_client.owner.outputs[0].clients = client.outputs[0].clients
+            r = local_fill_sink.transform(new_client.owner)
             if r is False:
                 continue
             replacements.pop(node.outputs[0], None)
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3293,6 +3293,7 @@ def local_remove_switch_const_cond(node):
     return False
 
 
+@register_specialize
 @register_canonicalize
 @gof.local_optimizer([T.mul])
 def local_mul_switch_sink(node):
@@ -3322,6 +3323,7 @@ def local_mul_switch_sink(node):
         return False
     for idx, i in enumerate(node.inputs):
         if i.owner and i.owner.op == T.switch:
+            # import ipdb;ipdb.set_trace()
             switch = i.owner
             try:
                 if (get_scalar_constant_value(
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -2031,6 +2031,11 @@ local_elemwise_alloc = register_specialize(
         local_elemwise_alloc_op(T.Elemwise, T.Alloc, T.DimShuffle)),
     'local_alloc_elemwise')
 
+local_elemwise_alloc = register_canonicalize(
+    gof.local_optimizer([T.Elemwise])(
+        local_elemwise_alloc_op(T.Elemwise, T.Alloc, T.DimShuffle)),
+    'local_alloc_elemwise')
+
 theano.configparser.AddConfigVar('experimental.local_alloc_elemwise',
                                  "DEPRECATED: If True, enable the experimental"
                                  " optimization local_alloc_elemwise."
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -2875,7 +2875,6 @@ class Test_local_elemwise_alloc(unittest.TestCase):
         )
 
     def _verify_assert_count(self, f, count):
-        theano.printing.debugprint(f.maker.fgraph)
         assert(
             sum([isinstance(elem.op, T.opt.Assert)
                  for elem in f.maker.fgraph.toposort()
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -2875,6 +2875,7 @@ class Test_local_elemwise_alloc(unittest.TestCase):
         )
 
     def _verify_assert_count(self, f, count):
+        theano.printing.debugprint(f.maker.fgraph)
         assert(
             sum([isinstance(elem.op, T.opt.Assert)
                  for elem in f.maker.fgraph.toposort()
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -4101,6 +4101,8 @@ class T_local_switch_sink(unittest.TestCase):
                             self.resm[idx][i])).sum() == self.resm[idx][i].size
                 else:
                     res = f(condition[1], x[1], -1)
+                    # theano.printing.debugprint(f.maker.fgraph.outputs[0])
+                    # import ipdb;ipdb.set_trace()
                     assert (res == numpy.asarray(self.
                         resm[idx])).sum() == self.resm[idx].size
                 idx += 1
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -4102,7 +4102,7 @@ class T_local_switch_sink(unittest.TestCase):
                             self.resm[idx][i])).sum() == self.resm[idx][i].size
                 else:
                     res = f(condition[1], x[1], -1)
-                    # theano.printing.debugprint(f.maker.fgraph.outputs[0])
+                    theano.printing.debugprint(f.maker.fgraph.outputs[0])
                     # import ipdb;ipdb.set_trace()
                     assert (res == numpy.asarray(self.
                         resm[idx])).sum() == self.resm[idx].size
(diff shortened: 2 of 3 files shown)


t13m:
The new profiling output is https://gist.github.com/t13m/0233c448e767aae9ba1e#file-opt_local_fill_sink

t13m:
I think a new profiling output without this PR's optimization should be done, I'll put it here later.

nouiz:
travis is failing

t13m:
travis build is now failing on flake8 checking of test_dnn.py, should I fix it?

this opt now is done recursively only for one of all its clients. I'm trying to make it to be applied on all the clients.

nouiz:
the travis problem have been fixed in master. I trigger a new build with the updated master.

For that opt, as told, I'm not sure this is the right approach. But I let you decide if you want to continue this one or try the other idea to make a global optimizer with that opt.

nouiz:
If you continue on your version, make sure that local_fill_sink is registered before local_fill_alloc in that file.

theano/tensor/opt.py
```diff
@@ -3421,7 +3423,21 @@ def local_fill_sink(node):
...
+
+    # The newly created node c doesn't has 'clients',
+    # so this iteration is took place with node.outputs[0]
+    replacements = {node.outputs[0]: c}
+    for client, _ in node.outputs[0].clients:
+        if (hasattr(client, 'op') and
+                isinstance(client.op, T.Elemwise) and
+                not client.op == T.fill):
+            # import ipdb; ipdb.set_trace()
+            r = local_fill_sink.transform(client)
```
lamblin: Oh, I think I understand. Here, one input of `client` is still `node.outputs[0]`, not `c`. So, inside that call to `local_fill_sing.transform`, its inputs are not `fill` nodes, and `transform` will just return `False`.
What you actually have to do is _rebuild_ a new node instead of `client`, that takes `c` as input instead of `node.outputs[0]`, and call `local_fill_sink.transform` on that "new" client.

For instance:
```python
for client, cl_idx in node.outputs[0].clients:
    if (...):  # same condition
        client_inputs = client.inputs[:]
        # Replace node.outputs[0]  by c in the client inputs
        client_inputs[cl_idx] = c
        new_client = client.node.op(*client_inputs)
        r = local_fill_sink.transform(new_client)
        ...
```
This is not tested, just off the top of my head.
Could you try to make it work, and check how many times that path of the code is actually called?

theano/tensor/opt.py
```diff
@@ -3421,7 +3423,21 @@ def local_fill_sink(node):
...
+            if r is False:
```
lamblin: `if not r:` is more robust in general, because optimizations sometimes return `None` or `[]` as well as `False`.

theano/tensor/opt.py
t13m: Ah yes, you're right, thanks very much, but it still seems not work as expected for now (always return `False`). I'm trying to find out why.

t13m:
The output of line_profile:
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  3402                                           @gof.local_optimizer([T.Elemwise])
  3403                                           @profile
  3404                                           def local_fill_sink(node):
  3405                                               """
  3406                                               f(fill(a, b), fill(c, d), e) -> fill(a, fill(c, f(b, d, e)))
  3407                                           
  3408                                               f need to be an elemwise
  3409                                               """
  3410     56380       110700      2.0      2.3      if (not hasattr(node, 'op') or
  3411     56380       103091      1.8      2.2              not isinstance(node.op, T.Elemwise) or
  3412     56380       593233     10.5     12.6              node.op == T.fill):
  3413       527          608      1.2      0.0          return False
  3414     55853        62208      1.1      1.3      models = []
  3415     55853        52670      0.9      1.1      inputs = []
  3416    173087       193259      1.1      4.1      for input in node.inputs:
  3417    117234       875226      7.5     18.6          if input.owner and input.owner.op == T.fill:
  3418      2402         5420      2.3      0.1              models.append(input.owner.inputs[0])
  3419      2402         4522      1.9      0.1              inputs.append(input.owner.inputs[1])
  3420                                                   else:
  3421    114832       171899      1.5      3.6              inputs.append(input)
  3422     55853        56562      1.0      1.2      if not models:
  3423     53451        51191      1.0      1.1          return False
  3424      2402       863235    359.4     18.3      c = node.op(*inputs)
  3425      4804         7093      1.5      0.2      for model in models:
  3426      2402       828773    345.0     17.6          c = T.fill(model, c)
  3427                                           
  3428                                               # The newly created node c doesn't has 'clients',
  3429                                               # so this iteration is took place with node.outputs[0]
  3430      2402         3816      1.6      0.1      replacements = {node.outputs[0]: c}
  3431      5599         9653      1.7      0.2      for client, cl_idx in node.outputs[0].clients:
  3432      3197         5394      1.7      0.1          if (hasattr(client, 'op') and
  3433      3197         5940      1.9      0.1                  isinstance(client.op, T.Elemwise) and
  3434      2225        25336     11.4      0.5                  not client.op == T.fill):
  3435      1900         3901      2.1      0.1              client_inputs = client.inputs[:]
  3436      1900         2240      1.2      0.0              client_inputs[cl_idx] = c
  3437      1900       648595    341.4     13.8              new_client = client.op(*client_inputs)
  3438                                           
  3439                                                       # Add clients to new_client
  3440      1900         6192      3.3      0.1              new_client.owner.outputs[0].clients = client.outputs[0].clients
  3441      1900         8320      4.4      0.2              r = local_fill_sink.transform(new_client.owner)
  3442      1900         2386      1.3      0.1              if r is False:
  3443                                                           continue
  3444      1900         3780      2.0      0.1              replacements.pop(node.outputs[0], None)
  3445      1900         4455      2.3      0.1              replacements.update(r)
  3446      2402         2156      0.9      0.0      return replacements
```

It shows that `local_fill_sink` can return something other than False now, and another profile shows that the time applied was decreased. But it fails on `T_local_switch_sink` test case. I'm trying to figure out why. Do have any idea about this?

t13m:
In the function `test_local_mul_switch_sink(self)`, I found
```
                y = T.mul(T.switch(condition[0] > 0, 1. * x[0], 0. * x[0]),
                          T.switch(condition[0] > 0,
                                   1. * x[0], T.log(c) * x[0]))
                f = theano.function([condition[0], x[0], c],
                                    [y], mode=self.mode)
                if type(condition[1]) is list:
                       # omitted
                else:
                    res = f(condition[1], x[1], -1)
                    # it fails here
                    assert (res == numpy.asarray(self.
                        resm[idx])).sum() == self.resm[idx].size
```
when function `f` is called, argument `c` equals to `-1`, so `y` will contain `NaN` as it calculates `T.log(-1)`. The values of `res` and `self.resm[0]` and `condition[1]` are listed as follows. I think `res` is the correct result and the `self.resm[0]` is wrong. Am I misunderstood anything?
```
ipdb> p res
[array([[  1.,  nan,   1.,  nan],
       [ nan,  nan,  nan,  nan],
       [  1.,   1.,   1.,   1.]])]

ipdb> p self.resm[0]
array([[1, 0, 1, 0],
       [0, 0, 0, 0],
       [1, 1, 1, 1]])

ipdb> p condition[1]
array([[ 0.1,  0. ,  1. , -1. ],
       [ 0. ,  0. ,  0. ,  0. ],
       [ 1. ,  1. ,  1. ,  1. ]])
```

lamblin:
The new profile seems promising: the time spent in `local_fill_sink` is down, and the number of iterations of `canonicalize` is down as well: from `time 166.830s for 12 passes` to `time 143.991s for 9 passes`.
However, and I don't know why, the number of nodes at the end of the canonicalize phase is up from 14614 to 14794. It seems to make subsequent optimization phases (in particular `specialize`, and then `gpu_opt`) slower because they have more nodes to go through.

Regarding the failing test, I'll have a look. It may be that the test needs to be upgraded, or that the new optimization is missing some cases, or doing something incorrect in corner cases.

lamblin:
Regarding the failing test, I think the problem appears because the order in which the optimizations are applied has changed, so the final compiled graph is different.
Could you use debugprint on `f` with your changes and without them, to check what the difference is?

lamblin:
Finally, as @nouiz mentionned in an earlier comment, can you check if registering `local_fill_sink` _before_ `local_fill_to_alloc` makes a difference?
Currently, `register_canonicalize(local_fill_sink)` is called on line 3416 of `tensor.opt.py`, but `@register_canonicalize` is called on `def local_fill_to_alloc` on line 1274.
Thanks!

theano/tensor/opt.py
```diff
@@ -3421,7 +3423,26 @@ def local_fill_sink(node):
...
+            client_inputs = client.inputs[:]
+            client_inputs[cl_idx] = c
+            new_client = client.op(*client_inputs)
+
+            # Add clients to new_client
+            new_client.owner.outputs[0].clients = client.outputs[0].clients
+            r = local_fill_sink.transform(new_client.owner)
+            if r is False:
+                continue
+            replacements.pop(node.outputs[0], None)
```
lamblin: I just realized: it may be a bad thing to remove `node.outputs[0]` from `replacements` in general, because some clients may not be replaced, and in that case, they would still depend on the un-optimized `node.outputs[0]` instead of `c`.
Can you try to keep `node.outputs[0]` in `replacements`?
If there is a problem, let us know, and we will try to find a way to remove it only if *all* of the clients have been replaced.

t13m: OK, I'll try. But I think that the if any one client of `node.outputs[0]` be replaced, the newly created node will have `node.outputs[0]` as input, when the replacement happening, I thought the node.outputs[0] will also be replaced, in that case, `node.outputs[0]` can be removed.

t13m: Besides, if `node.outputs[0]` is included in `replacements`, is it possible that there will be two versions after the replacement happen?

t13m:
About the failed test case
with these changes, debugprints:
```
Elemwise{Mul}[(0, 0)] [@A] ''   
 |Elemwise{switch,no_inplace} [@B] ''   
 | |Assert{msg='Theano Assert failed!'} [@C] ''   
 | | |Elemwise{gt,no_inplace} [@D] ''   
 | | | |cond [@E]
 | | | |TensorConstant{(1, 1) of 0} [@F]
 | | |Elemwise{eq,no_inplace} [@G] ''   
 | | | |Shape_i{0} [@H] ''   
 | | | | |x [@I]
 | | | |Shape_i{0} [@J] ''   
 | | |   |cond [@E]
 | | |Elemwise{eq,no_inplace} [@K] ''   
 | |   |Shape_i{1} [@L] ''   
 | |   | |x [@I]
 | |   |Shape_i{1} [@M] ''   
 | |     |cond [@E]
 | |x [@I]
 | |TensorConstant{(1, 1) of 0.0} [@N]
 |Elemwise{Switch}[(0, 2)] [@O] ''   
   |Elemwise{gt,no_inplace} [@D] ''   
   |x [@I]
   |Elemwise{mul,no_inplace} [@P] ''   
     |Elemwise{log,no_inplace} [@Q] ''   
     | |InplaceDimShuffle{x,x} [@R] ''   
     |   |<TensorType(float64, scalar)> [@S]
     |x [@I]
```

without the changes
```
Elemwise{Switch}[(0, 1)] [@A] ''   
 |Elemwise{gt,no_inplace} [@B] ''   
 | |cond [@C]
 | |TensorConstant{(1, 1) of 0} [@D]
 |Elemwise{Mul}[(0, 0)] [@E] ''   
 | |Elemwise{Switch}[(0, 2)] [@F] ''   
 | | |Elemwise{gt,no_inplace} [@B] ''   
 | | |x [@G]
 | | |Elemwise{mul,no_inplace} [@H] ''   
 | |   |Elemwise{log,no_inplace} [@I] ''   
 | |   | |InplaceDimShuffle{x,x} [@J] ''   
 | |   |   |<TensorType(float64, scalar)> [@K]
 | |   |x [@G]
 | |x [@G]
 |TensorConstant{(1, 1) of 0} [@D]
```
I checked `local_mul_switch_sink`, I think it fails because this optimizer is not applied, but don't know why. Does the registration order matter?

lamblin:
Thanks for the update!

I think the Assert op is introduced by some optimization removing `Alloc` node, which means `local_fill_to_alloc` may be applied earlier than before. Registration order can have an influence on that.
The Assert it actually disrupting the optimization that was tested: the two `Switch` Ops, labeled `[@B]` and `[@O]` in the first graph, have conditions that are equivalent, but one is working directly on the condition `[@D]`, and one is working on the result of an Assert on the same condition (`[@C]`).

Is the first debugprint after the changes you just added, or with the changes you added yesterday?

If changing the order of registration did not change the graph, there are different ways we can solve that particular test. One would be to go through asserts during `local_switch_mul_sink`, another would be to disable `local_fill_to_alloc` in that particular test, another would be to include `unsafe` optimization to get rid of the assert. This is not a big issue.

What would be more interesting is the result of your last changes on the profiling.

t13m:
The first debugprint under version `59e03da`, which is not the latest commit.

The profiling output with `local_fill_sink` before `fill_to_alloc`: https://gist.github.com/t13m/0233c448e767aae9ba1e#file-rec_all_after_fill_sink_before_alloc
for the `canonicalize`, it seems nothing changed.


The profiling output with `node.outputs[0]` not removed:https://gist.github.com/t13m/0233c448e767aae9ba1e#file-rec_all_after_not_remove_output0
this shows that the `canonicalize` optimizer is applied more passes than before. Maybe better to keep `node.outputs[0]` removed? I'll have a look at how replacement happens.

Thanks.

t13m:
The debugprint with `local_fill_sink` registered before `fill_to_alloc`:
```
Elemwise{Mul}[(0, 0)] [@A] ''   
 |Elemwise{switch,no_inplace} [@B] ''   
 | |Assert{msg='Theano Assert failed!'} [@C] ''   
 | | |Elemwise{gt,no_inplace} [@D] ''   
 | | | |cond [@E]
 | | | |TensorConstant{(1, 1) of 0} [@F]
 | | |Elemwise{eq,no_inplace} [@G] ''   
 | | | |Shape_i{0} [@H] ''   
 | | | | |x [@I]
 | | | |Shape_i{0} [@J] ''   
 | | |   |cond [@E]
 | | |Elemwise{eq,no_inplace} [@K] ''   
 | |   |Shape_i{1} [@L] ''   
 | |   | |x [@I]
 | |   |Shape_i{1} [@M] ''   
 | |     |cond [@E]
 | |x [@I]
 | |TensorConstant{(1, 1) of 0.0} [@N]
 |Elemwise{Switch}[(0, 2)] [@O] ''   
   |Elemwise{gt,no_inplace} [@D] ''   
   |x [@I]
   |Elemwise{mul,no_inplace} [@P] ''   
     |Elemwise{log,no_inplace} [@Q] ''   
     | |InplaceDimShuffle{x,x} [@R] ''   
     |   |<TensorType(float64, scalar)> [@S]
     |x [@I]
```

lamblin:
OK, so removing `node.outputs[0]` in all cases seems a bad idea.
Ideally, I think we should remove it if all its clients have been replaced, but not otherwise.
The code for that could look like:
```python
replacements = {node.outputs[0]: c}
all_clients_replaced = True
for client, cl_idx in node.outputs[0].clients:
    if (hasattr(client, 'op') and
            isinstance(client.op, T.Elemwise) and
            not client.op == T.fill):
        client_inputs = client.inputs[:]
        client_inputs[cl_idx] = c
        new_client = client.op(*client_inputs)

        # Add clients to new_client
        new_client.owner.outputs[0].clients = client.outputs[0].clients
        r = local_fill_sink.transform(new_client.owner)
        if not r:
            all_clients_replaced = False
            continue
        replacements.update(r)
    else:
        all_clients_replaced = False

if all_clients_replaced:
    replacements.pop(node.outputs[0], None)
```

t13m:
I think you are right. I checked `FunctionGraph.replace(r, new_r)`, it only deals with `r` and `r`'s clients, and doesn't care about how other nodes depend on `r`'s input (or input of input), which maybe also replaced.

The new profiling output: https://gist.github.com/t13m/0233c448e767aae9ba1e#file-rec_all_after_remove_onlyif_all_replaced

lamblin:
OK, thanks.
The new profile seems a bit worse than "after2", but I'm not sure if this is as bad as it looks, because some phases (for instance "specialize") seem to have exactly the same execution profile, but a higher execution time, so this could simply be that the machine was busier at that time.

Unfortunately, it means that this is not the solution to the slow-down of later optimization phases.
I suspect this has to do with either:
- other optimizations combining with "fill" along the way, which do not have a chance to get executed now, or
- some merging optimizations no longer being applied, although I'm not sure how that could happen.

Speaking of merge optimizations, I just realized that "merge3" is completely useless: all the modifications it is suggesting are rejected, and it takes 140s. Given that the comment above its registration says "just to be sure", I think we could really well get rid of it.

lamblin:
Things that could be investigated to make this work (to be discussed with @nouiz)
- make `local_fill_to_alloc` always return an `alloc`, right now not all cases are implemented
- check when the optimization replacing `alloc(alloc(...))` by `alloc(...)` is merged, maybe that will bring down the number of nodes
- check with `optimizer_including=unsafe`, maybe the `assert` being added by some `alloc` optimization (but not by `fill`) are disrupting other optimizations
- if that is the case, maybe add an optimization to make Ops work after the Assert nodes. For instance, if we have `y = f(x)` and `x` has an assert client (`x_`), then replace `y` by `f(x_)`. This would ensure that it would be merged with existing `f(x_)`. Or we could integrate that in MergeOptimizer.

t13m:
Hi,
Thanks very much.
For the failed test case, I checked `local_fill_to_alloc`, but the unimplemented part was not called. 

Also, I tested with `self.mode = theano.compile.mode.get_default_mode().including( 'canonicalize', 'fast_run', 'local_remove_all_assert').excluding('gpu', 'fusion')`. It failed too, although assert node is removed. Is it because that both the opt who inserts `Assert` and opt `local_mul_switch_sink` are in `canonicalize` optimizer, so that `local_remove_all_assert` become useless?

Is there a way to know who generated the `Assert` node? or Is there a way to know how the function graph changed during the whole optimization period?

It seems to me a better option to make op works after the `Assert` node. A new local optimizer can be added to do this job, and be applied only when an `Assert` node is inserted (may be implemented as a fgraph feature).

t13m:
I'm trying to remove the assert opt in MergeOptimizer, by adding a method `process_assert`. An assert node takes several conditions as inputs. What is `(x_)` stands for, do you mean a condition?

nouiz:
You seem to have well advanced while I was in vacation. Great work, but it seem the problem is harder then we taught!

1) I just merged a new opt related to alloc that merge two consecutive alloc: https://github.com/Theano/Theano/pull/3082

Can you rebase your PR on top of Theano and profile again with it. This could help with some of those timming.

2) After that, to correctly test without assert, modify the Assert op in tensor/opt.py. Add an `__call__` method to it that look like this:

```
def __call__(*inputs):
    return inputs[0]
```

This should remove all assert from the graph by not creating them. Do a new timing with this.

3) For the merge optimizer if we have 2 nodes in the graph: f(x) and f(assert(x, cond)), currently, they don't get merged. Just to be clear, with more detail:
```
import theano
x=theano.tensor.matrix()
y1=x+1
y2=theano.tensor.opt.assert_op(x, (x>0).all())+1
f=theano.function([x], [y1,y2])
theano.printing.debugprint(f)

Elemwise{add,no_inplace} [@A] ''   1
 |TensorConstant{(1, 1) of 1.0} [@B]
 |<TensorType(float64, matrix)> [@C]
Elemwise{add,no_inplace} [@D] ''   4
 |TensorConstant{(1, 1) of 1.0} [@B]
 |Assert{msg='Theano Assert failed!'} [@E] ''   3
   |<TensorType(float64, matrix)> [@C]
   |All [@F] ''   2
     |Elemwise{gt,no_inplace} [@G] ''   0
       |<TensorType(float64, matrix)> [@C]
       |TensorConstant{(1, 1) of 0} [@H]

```

Here we see that the node A and D aren't merged. They need to be merged. Mostly, we need to make a special case in MergeFeature.process_node, to merge A and D (and keep the constant).

t13m:
The profiling output of the rebased code: https://gist.github.com/t13m/2eb855fe4641e9f5d703#file-opt_rebased
The profiling output without this modification: https://gist.github.com/t13m/2eb855fe4641e9f5d703#file-trunk_bd168a5

The time took by `canonicalize` optimizer is less, but it end up with more nodes than before. I think this is also the reason why other optimizers take more time, like `gpu_opt`.

The profiling output after removing assert: https://gist.github.com/t13m/2eb855fe4641e9f5d703#file-remove-assert

t13m:
In the latest commit, there is a problem. For the example you provided, `MergeFeature.scheduled` contained pairs `[[[(Elemwise{add,no_inplace}.0, Elemwise{add,no_inplace}.0)]]]`, but the merging didn't happen. It ended up in no change. 
Another question is, what's the difference between `Elemwise{add,no_inplace}` and `Elemwise{Add}[(0, 1)]`? I see the `add` op with Assert once became `Elemwise{Add}` and then back to `Elemwise{add, no_inplace}`, guessing this is related to the merge issue.

nouiz:
The diff between "Elemwise{add,no_inplace}" and "Elemwise{Add}[(0, 1)]" is that the second is inplace. The output index 0 will be writen on the input index 1.

t13m:
Nodes with assert inputs can be merged now.
But for the code:
```python
import theano
x=theano.tensor.matrix()
y1=x+1
y2=theano.tensor.opt.assert_op(x, (x>0).all())+1
f=theano.function([x], [y1,y2])
theano.printing.debugprint(f)
print('\n')

# If inputs vars' order changed, 
# the behavior would change too
f_rev=theano.function([x], [y2,y1])  
theano.printing.debugprint(f_rev)
```

The output is
```
DeepCopyOp [@A] ''   4
 |Elemwise{add,no_inplace} [@B] ''   3
   |TensorConstant{(1, 1) of 1.0} [@C]
   |Assert{msg='Theano Assert failed!'} [@D] ''   2
     |<TensorType(float64, matrix)> [@E]
     |All [@F] ''   1
       |Elemwise{gt,no_inplace} [@G] ''   0
         |<TensorType(float64, matrix)> [@E]
         |TensorConstant{(1, 1) of 0} [@H]
Elemwise{add,no_inplace} [@B] ''   3

DeepCopyOp [@A] ''   1
 |Elemwise{add,no_inplace} [@B] ''   0
   |TensorConstant{(1, 1) of 1.0} [@C]
   |<TensorType(float64, matrix)> [@D]
Elemwise{add,no_inplace} [@B] ''   0
```
Is this behavior what we want? Or are we going to only replace node without assert with those with assert?

nouiz:
Both case should give the first graph.

t13m:
How to merge nodes like these: one is `f( x, assert(y, cond_y))`, the other is `f( assert(x, cond_x), y)`. I think a new node `f( assert(x, cond_x), assert(y, cond_y))` can be created. But should it be done here in the `MergeOptimizer`?

t13m:
Another condition is `f( assert(x, cond_1) )` and `f( assert(x, cond_2) )`. Should the two conditions be combined?

t13m:
In the new commit I try to merge nodes with assert input at different position, by creating a new node. The problem is that when `MergeOptimizer.apply()` is called, following conditions will be examined.
```python
                var = pairs[0][0]
                candidate = pairs[0][1]
                if (not hasattr(var, 'fgraph') or
                        not hasattr(candidate, 'fgraph')):
                    continue
```
But for a newly created node, there is no `fgraph` attribute. Can this checking be skipped?

Thanks,

nouiz:
For the first question: I think we should create the `f( assert(x, cond_x), assert(y, cond_y))` node in the merge optimizer.
For the second question, good new corner case. I also think they should be merged.
For the 3rd question, I think we should keep that check for the normal case, but I think you can make a different test for created nodes. You could in that case, create a triple instead of a pair, where the last element is a string "new_node". In that case, you only check if var have an fgraph and don't check the candidate.

nouiz:
By the way, there is some flake8 test failing in travis.

t13m:
I'm trying to combine assert nodes on the same var but with different conditions, is `theano.tensor.and_` the right method to use?

t13m:
For the failed test case `test_local_mul_switch_sink`, is it ok to modify `local_mul_switch_sink` to make it be able to go through asserts?

t13m:
In the new commit, the following code:
```
import theano
x1 = theano.tensor.matrix()
x2 = theano.tensor.matrix()
y1 = theano.tensor.opt.assert_op(x1, (x1 >= 0).all()) + theano.tensor.opt.assert_op(x2, T.and_((x2 > 0).all(), (x1 < 0).all())) + 1
y2 = theano.tensor.opt.assert_op(x1, (x1 > 0).all()) + x2 + 1
f = theano.function([x1, x2], [y1, y2])
theano.printing.debugprint(f)
```

Its output is as followed:
```
DeepCopyOp [@A] ''   10
 |Elemwise{add,no_inplace} [@B] ''   9
   |TensorConstant{(1, 1) of 1.0} [@C]
   |Assert{msg='Theano Assert failed!'} [@D] ''   6
   | |<TensorType(float64, matrix)> [@E]
   | |All [@F] ''   4
   |   |Elemwise{ge,no_inplace} [@G] ''   1
   |     |<TensorType(float64, matrix)> [@E]
   |     |TensorConstant{(1, 1) of 0} [@H]
   |Assert{msg='Theano Assert failed!'} [@I] ''   8
     |<TensorType(float64, matrix)> [@J]
     |Elemwise{AND}[(0, 0)] [@K] ''   7
       |All [@L] ''   5
       | |Elemwise{gt,no_inplace} [@M] ''   2
       |   |<TensorType(float64, matrix)> [@J]
       |   |TensorConstant{(1, 1) of 0} [@H]
       |All [@N] ''   3
         |Elemwise{lt,no_inplace} [@O] ''   0
           |<TensorType(float64, matrix)> [@E]
           |TensorConstant{(1, 1) of 0} [@H]
Elemwise{add,no_inplace} [@B] ''   9
```

The different assert inputs are merged

t13m:
For the case of "new_node", following condition checking cannot be passed ( in MergeOptimizer.apply() )
```python
                        inputs_match = all(node_in is cand_in
                                           for node_in, cand_in in
                                           zip(node_inputs_assert_removed,
                                               cand_inputs_assert_removed))
```
So I skip this checking.

t13m:
Hi,
I did some debugging job to get to know how the test case fails. 
The `Assert` node is introduced by `local_elemwise_alloc` opt, it can be removed by setting `theano.config.experimental.local_alloc_elemwise_assert = False`.
Thought, the `local_mul_switch_sink` opt is still not applied because the input of `switch` node is not constant but `alloc` node, which is introduced by opt `local_fill_to_alloc`.

opt `local_elemwise_alloc` removes `alloc`, but it is registered in `special`, and `local_mul_switch_sink` is registered in `canonicalize`. So I'm trying to have `local_elemwise_alloc` also registered in `canonicalize` to see if the problem solved.

t13m:
Don't know why the test case `Test_local_elemwise_alloc` failed, it passed on my one computer. And for the flake8 checking, can't find the position pointed out by travis build. How can I fix it?

nouiz:
I also do not understand why travis fail. I restart it.

Can you move to another PR the change related to merging alloc? That way, we can merge this part when ready and not wait.

t13m:
Ok. By "change related to merging alloc" do you mean changes related to merging nodes with assert inputs?

nouiz:
yes. I used alloc instead of assert in my sentence.

t13m:
PR：https://github.com/Theano/Theano/pull/3232

t13m:
the failed case seems can be passed if registering `local_mul_switch_sink` to `specialize`.
is there anyway to evaluate the travis's flake8 checking on my own computer?

Thanks

nouiz:
I think it is ok to register local_mul_switch_sink it specialize in addition to what is actually done.

You can run theano-nose theano/tests/test_flake8.py, but sometimes, there is still some difference compared to what travis give, but it is the best I know that be done in advance.

theano/tensor/opt.py
```diff
@@ -1269,6 +1269,59 @@ def apply(self, fgraph):
...
+        if input.owner and input.owner.op == T.fill:
+            models.append(input.owner.inputs[0])
+            inputs.append(input.owner.inputs[1])
+        else:
+            inputs.append(input)
+    if not models:
+        return False
+    c = node.op(*inputs)
+    for model in models:
+        c = T.fill(model, c)
```
nouiz: I think I understand what is missing. This line create more fill node. The opt local_fill_to_alloc also remove useless fill, not just the opt local_useless_fill. Instead of doing this ping/pong between this opt and local_fill_to_alloc, can you try to introduce here model only when it is not a useless fill?

replace the loop with this:
```
    for model in models:
        if model.type != c.type:
            c = T.fill(model, c)
```

Can you add this, and rebase this PR on top of 3232 to don't have them collide and time again?

t13m:
I think the test case can be pass this time. But still confused how the test result can be different in travis builder and my my own computer. > <

nouiz:
@t13m can you remove the merge of assert modification from this PR?

theano/tensor/opt.py
```diff
@@ -1269,6 +1269,60 @@ def apply(self, fgraph):
...
+        if input.owner and input.owner.op == T.fill:
+            models.append(input.owner.inputs[0])
+            inputs.append(input.owner.inputs[1])
+        else:
+            inputs.append(input)
+    if not models:
+        return False
+    c = node.op(*inputs)
+    for model in models:
+        if model.type != c.type:
```
nouiz: I think this is the right condition to help remove the ping/pong between this opt and another one.

theano/tensor/tests/test_opt.py
```diff
@@ -4100,6 +4100,8 @@ def test_local_mul_switch_sink(self):
                             self.resm[idx][i])).sum() == self.resm[idx][i].size
                 else:
                     res = f(condition[1], x[1], -1)
+                    theano.printing.debugprint(f.maker.fgraph.outputs[0])
```
nouiz: can you also remove this debugprint? We prefer test to not be verbose.

nouiz:
Sorry for the late review. I'm back recently from vacation. Can you help finish it? The change are easy todo, if you don't have much time, I can do them. But if you could make a new timming, it will be faster for you to do it them me I think. I forgot where the example is and how to run it!

thanks.

t13m:
Hi @nouiz , I'm sorry to reply so late. I can make change to it but can not profile it now (my laptop has no GPU which is needed by the case). Profiling can be done in next week after me get back to school, is that OK? :)

nouiz:
That is good. Thanks.

nouiz:
Could you time it at the begginging of this week? If not, I'll try to do
it. I think this is the only thing missing to merge this and I know many
people would love the speed up we think this can bring.

If you don't have the time, just tell us so that we know.

thanks

On Thu, Sep 3, 2015 at 1:33 AM, Frédéric Bastien <frederic.bastien@gmail.com
> wrote:

> That is good. Thanks.
>

t13m:
Hi, I thought that assert node functionality needed to be removed so I deleted the code. Ok, I'll remove those commits. And I'm sorry to be here so late, a little busy as a new semester just begins.

I've just done the profiling. 

profile with PR: https://gist.github.com/t13m/0e43627b890d39e80276#file-20150915_profile_with_opt_local_fill_sink

profile without this change: https://gist.github.com/t13m/0e43627b890d39e80276#file-20150915_profile_without_opt_local_sink

`21.508s - 2364 - 4728 - local_fill_sink` to `5.364s - 679 - 502 - local_fill_sink`

Thanks,

nouiz:
I'm merging it. Thanks!

nouiz:
I need to wait for travis to pass! But I think it is good to go.

t13m:
Those last commits removed. Is it ok to use `git reset --hard` and `git push --force`?
Thanks

nouiz:
Yes.

theano/tensor/tests/test_opt.py
nouiz: It seem this line come back. Can you remove it? It make the tests extra verbose and this isn't good.

nouiz:
thanks.

nouiz:
thanks.

PR #3099, opened 2015-07-02
Fix flake8 errors in travis that I don't have locally.

nouiz:


nouiz:
I merge now as this make the master fail.

PR #3100, opened 2015-07-02
Added Help! section to docs index page

danielrenshaw:


nouiz:
I stoped travis as it isn't needed (doc only)

It is good for me @carriepl can you check it? If good for you, you can merge.

doc/index.txt
```diff
@@ -28,7 +28,7 @@ News
 
 * Ian Goodfellow did a `12h class with exercises on Theano <https://github.com/goodfeli/theano_exercises>`_.
 
-* Theano 0.6 was released. Everybody is encouraged to update.
+* Theano 0.7 was released 26th March 2015. Everybody is encouraged to update.
```
lamblin: I would put that at the top of the list, this is actually the latest news.

doc/index.txt
```diff
@@ -168,6 +168,60 @@ Community
 
 * Come visit us in Montreal! Most developers are students in the LISA_ group at the `University of Montreal`_.
 
+
+.. _support:
+
+Help!
+=====
```
lamblin: That's a good section to have, thanks!
I think it would be worth having an explicit point about looking for the answer to similar questions before looking for help.

doc/index.txt
danielrenshaw: Done

doc/index.txt
danielrenshaw: Done

carriepl:
This looks great. Thanks!

Issue #2985, opened 2015-05-30
Handle broadcastable better in outputs_info of scan


kyunghyuncho:
Passing a tensor whose shape includes 1 as a part of ``outputs_info`` when calling ``theano.scan`` fails unless those axes of shape 1 are unbroadcasted manually using ``tensor.unbroadcast``. This is due to the different broadcasting pattern between the actual return from the inner function of ``scan`` and the corresponding one passed via ``outputs_info``. It'll make ``theano.scan`` more usable, if ``scan`` can handle this situation automatically.

A sample code reproducing this behaviour
```
import theano
from theano import tensor

def test0(n_samples):
    def _loop(x, y):
        x = x + y
        return x

    rval, updates = theano.scan(_loop, sequences=[],
                                outputs_info=[tensor.alloc(0., n_samples, 10)],
                                non_sequences=[tensor.alloc(1., n_samples, 10)],
                                n_steps=10)

    test_fn = theano.function([], rval[-1])

    print test_fn()

def test1(n_samples):
    def _loop(x, y):
        x = x + y
        return x

    rval, updates = theano.scan(_loop, sequences=[],
                                outputs_info=[tensor.unbroadcast(tensor.alloc(0., n_samples, 10),0)],
                                non_sequences=[tensor.alloc(1., n_samples, 10)],
                                n_steps=10)

    test_fn = theano.function([], rval[-1])

    print test_fn()

if __name__ == '__main__':
    print 'Working case: n_samples > 1'
    test0(2)

    print 'Working with n_samples == 1 if unbroadcasted'
    test1(1)

    print 'Not working with n_samples == 1'
    test0(1)
```

This is the result of running the code above
```
Working case: n_samples > 1
/u/chokyun/work/Theano/build/lib/theano/scan_module/scan_perform_ext.py:133: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility
  from scan_perform.scan_perform import *
[[ 10.  10.  10.  10.  10.  10.  10.  10.  10.  10.]
 [ 10.  10.  10.  10.  10.  10.  10.  10.  10.  10.]]
Working with n_samples == 1 if unbroadcasted
[[ 10.  10.  10.  10.  10.  10.  10.  10.  10.  10.]]
Not working with n_samples == 1
Traceback (most recent call last):
  File "./test.py", line 40, in <module>
    test0(1)
  File "./test.py", line 12, in test0
    n_steps=10)
[...]

PR #3102, opened 2015-07-02
Match broadcastable pattern of original variable

carriepl:
This is for the issue raised in : https://groups.google.com/forum/#!topic/theano-dev/3-tRoZcnMuI
I don't yet have a test case for it.

### DIFF
--- a/theano/scan_module/tests/test_scan.py
+++ b/theano/scan_module/tests/test_scan.py
@@ -2436,10 +2436,10 @@ class T_Scan(unittest.TestCase):
         # recurrent output of a scan node associated with a state with a
         # state with broadcastable dimensions
 
-        x = tensor.col()
-        seq = tensor.col()
+        x = tensor.dcol()
+        seq = tensor.dcol()
         outputs_info=[x, tensor.zeros_like(x)]
-        (out1, out2), updates = theano.scan(lambda a, b, c : (a + b, b + c),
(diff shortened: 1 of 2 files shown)


nouiz:
Seem ok to me, but need tests and if @lamblin or @abergeron could review this, it would be great.

carriepl:
Added a test case

theano/scan_module/tests/test_scan.py
```diff
@@ -2431,6 +2431,27 @@ def test_use_scan_direct_output(self):
...
+                                            outputs_info=outputs_info)
+
+        # Obtain a reference to the scan outputs before the subtensor and
+        # compile a function with them as outputs
+        assert isinstance(out1.owner.op, tensor.subtensor.Subtensor)
+        assert isinstance(out2.owner.op, tensor.subtensor.Subtensor)
+        out1_direct = out1.owner.inputs[0]
+        out2_direct = out2.owner.inputs[0]
+        fct = theano.function([x, seq],
+                              [out1_direct, out2_direct])
```
nouiz: Seem ok, but if you can call the compiled function, it would test the run time in that case. But I would be surprised this cause problems there.

carriepl:
Added a call to the compiled function. I stopped the Travis run on the previous commit.

carriepl:
@kyunghyuncho : Also, I'm not sure if it complete fixes the underlying issue, but it seems to solve the crash with the test code in #2985.

PR #3105, opened 2015-07-03
Add an optimisation for useless reshape of vectors (close #2752)

### LINKED ISSUE #2752: Opt: Reshape(arange(x)) -> arange(x) when this do not change the shapes, number of dimensions.
```
>>> import theano
>>> i=theano.tensor.iscalar()
>>> m=theano.tensor.mgrid[0:i,]
>>> f=theano.function([i], m)
>>> theano.printing.debugprint(f, print_type=True)
Reshape{1} [@A] <TensorType(int32, vector)> ''   3
 |ARange [@B] <TensorType(int32, vector)> ''   1
 | |TensorConstant{0} [@C] <TensorType(int8, scalar)>
 | |<TensorType(int32, scalar)> [@D] <TensorType(int32, scalar)>
 | |TensorConstant{1} [@E] <TensorType(int8, scalar)>
 |MakeVector [@F] <TensorType(int64, vector)> ''   2
   |Elemwise{Cast{int64}} [@G] <TensorType(int64, scalar)> ''   0
     |<TensorType(int32, scalar)> [@D] <TensorType(int32, scalar)>
```

This happen in gh-2572

Make this more general, in the sense that if the input is 1d and the output is 1d, what ever created it, remove it (add AssertOp if we can't assert that the shape are the same)

ejls:
I always replace the Reshape with Assert(s) when the input and output ndim are 1, if you had something else in mind, let me know.

CCW (@bouthilx)


theano/tensor/opt.py
```diff
@@ -3253,6 +3253,50 @@ def local_flatten_lift(node):
...
+        if node.op.ndim == 1 and x.ndim == 1:
+            out = x
+
+            new_shape_len = None
+            try:
+                new_shape_len = get_scalar_constant_value(shape.shape[0])
+            except NotScalarConstantError:
+                pass
+            if new_shape_len != 1:
+                out = assert_op(out, T.eq(shape.shape[0], 1))
```
nouiz: I don't understand why you have this logic for new_shape_len here. The if at like 3259 make sure that shape.shape[0] is 1.

theano/tensor/opt.py
```diff
@@ -3253,6 +3253,50 @@ def local_flatten_lift(node):
...
+
+            try:
+                new_shape = get_scalar_constant_value(shape[0])
+                if new_shape == -1:
+                    return [out]
+
+                if new_shape == 1:
+                    out = T.addbroadcast(out, 0)
```
nouiz: An optimization can't change the broadcast pattern of the node it use. The type of the new node must match the original node type. This include the broadcast patter. So you can't do that here, as if when we build the graph, we didn't got that information, the broadcast pattern must stay False.

theano/tensor/opt.py
```diff
@@ -3253,6 +3253,50 @@ def local_flatten_lift(node):
...
+
+                shape_of_x = node.fgraph.shape_feature.shape_of[x][0]
+                shape_of_x = get_scalar_constant_value(shape_of_x)
+                if shape_of_x == new_shape:
+                    return [out]
```
nouiz: That part is good. But you can detect case where this isn't constant with this pseudo code:
```
node.fgraph.shape_feature.same_shape(x, var_shape=shape)
```

Note, the var_shape parameter don't exist yet. You can easily add it yourself. In the file, tensor/opt.py, in the class ShapeFeature, modify the method same_shape to have this signature: `def same_shape(self, x, y=None, dim_x=None, dim_y=None, shape_var=None):`

Then, assert that one and only one of shape_var or y is None. if shape_var is not None, you use it for the variable sy in that method.

theano/tensor/opt.py
ejls: You mean 3269, the node.op.ndim check ?
It's to prevent errors like :
```
y = x.reshape(z, ndim=1)
f = function([x,z], y)
f([0,1,2,3], [2,2])
```

theano/tensor/opt.py
ejls: The thing is, the output of reshape can have a different broadcast pattern :
```
x = theano.tensor.vector('x')
y = x.reshape([1])
```
`y` broadcast is `(True,)`, whereas `x` is `(False,)`.
Should I just return and leave the original Reshape in this case ?

If I don't do this, I get an error because I replaced Reshape with a node whose output has different type.

theano/tensor/opt.py
nouiz: I was talking about line 3269. But after a new check, my comment wasn't good. You should not have a special case here for vector of size 1. What ever the size is, if we can't infer that the size is the same before and after, we must add an assert.

theano/tensor/opt.py
nouiz: You should handle that case, but not like that. You should compare the
node.outputs[0].type vs the one you plan to return, if they differ, fix it.
Otherwise, there is other case then having a constant of 1 that can lead to
a True broadcast... So your check currently won't handle all case.

lamblin: In that case, "fix it" means using the Rebroadcast Op, probably through the `broadcast_like` helper function, so that `x` is replaced with something of the same type.

theano/tensor/opt.py
```diff
@@ -3253,6 +3253,50 @@ def local_flatten_lift(node):
...
+            except (KeyError, AttributeError, NotScalarConstantError):
+                pass
```
ejls: Here is shape is a MakeVector, you can do

 if same_shape(x, shape_var=shape.owner.inputs): return [out]

in same_shape, shape_var is the shape_of[x] in that case.

nouiz:
this PR need a rebase.

Issue #2501, opened 2015-02-12
Hierarchical softmax: Make a wrapper for sparse_block_dot_SS


lamblin:
Make it easier to implement a hierarchical softmax model.
There should already be code by Stan & Cho.

PR #3109, opened 2015-07-03
Sparse block dot

adbrebs:
#2825 #2501

- creates a meta op for BlockSparseGemv and BlockSparseOuter (meta op has no perform implementation)
- Adds CPU implementations for BlockSparseGemv and BlockSparseOuter
- Adds functions to automatically register meta ops


bouthilx:
@nouiz @abergeron @lamblin Can one of you review the PR please? I've been too much involved in it to be "neutral".

bouthilx:
Honestly I don't understand why it says that 'cpu_sparse_block_gemv_opt' is already an existing tag. The only reason I can see is that opt.py is executed several times. Why then doesn't it happen to cuda/opt.py as it is imported many times as well? Normally imports only happen once in python, I can't understand why opt.py would be executed several times...

lamblin:
Sorry for not having time to review it yet.
The import of a file can happen several times if an `ImportError` was raised the first time (and caught), because then the module is not considered imported by Python.
I don't know why that would happen here.

Issue #3088, opened 2015-06-29
Weird error when exiting python/ipython after using GPU


escopeta:
Hi I am getting the following error:

Exception AttributeError: AttributeError("'NoneType' object has no attribute 'remove'",) in <bound method Unlocker.__del__ of <theano.gof.compilelock.Unlocker object at 0x7f6a05de4f90>> ignored

This happens after I execute the following commands:

import theano.sandbox.cuda
theano.sandbox.cuda.use('gpu0')

and then exit ipython or python. I have to mention I am able to successfully secure GPU for usage (I get Using gpu device 0: Tesla K40c) and use the GPU fine during my computations, but I am puzzled about this Exception message and would like to fix it on my side. The python I am using comes in anaconda (Anaconda-2.2.0-Linux-x86_64) and I installed theano following the "Bleeding-edge install instructions" from deeplearning.net. I would really appreciate any help on this matter. Thanks.

nouiz:
Your error message seem cut. What is between "in" and "ignored"? I have
this message here:

Exception TypeError: "'NoneType' object is not callable" in <bound method
Popen.__del__ of <subprocess.Popen object at 0x1baefd0>> ignored

Your isn't the same, so this could be comming from a different place.

nouiz:
I did a PR that fix my error. I get it with the same code as your example, but it isn't the same error. https://github.com/Theano/Theano/pull/3110

Can you try it and tell me if it fix it for you?

escopeta:
Hi @nouiz 

Sorry for the late response, I just upgraded to the latest commit and the error disappeared. Thank you very much!!!!

PR #3110, opened 2015-07-03
Fix error at exit. Do this also solve gh-3088?

nouiz:


nouiz:
Why are we having those problems now? Do someone have an idea? I'll start a git bisect to try to find when this happen.

nouiz:
The first commit that cause this problem: 030c7b969fef05da75abe14d19007612a60849d3

This is python23 PR. Mostly, it change the import of StringIO, copyreg, builtins and cPickle to use six and moved some six import to six.moves.

lamblin:
It fixes the error for @yaoli as well, so I'm merging.

nouiz:
The first commit that cause this problem: 030c7b969fef05da75abe14d19007612a60849d3

This is python23 PR. Mostly, it change the import of StringIO, copyreg, builtins and cPickle to use six and moved some six import to six.moves. Do you have any idea why change at that level could cause this problem?

Issue #3040, opened 2015-06-17
Clearer error messages for a common PyMC3 error


aflaxman:
There is a common error for people learning about theano through PyMC3, as exemplified in https://github.com/pymc-devs/pymc3/issues/754 ; in PyMC2, there was a common pattern where model parameters would be magically transformed into numpy arrays, and the user could manipulate them with familiar numpy functions, e.g. `np.exp(p)`.  Since PyMC3 is theano-based, translating is often as simple as switching from `np.exp` to `tt.exp`.  But it would be so helpful if there was an error message that said this. I think that the minimal example of this is the following:

```
import theano.tensor as tt

x = tt.dmatrix('x')
y = np.exp(x)
```

which raises the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-142-00dd09faf982> in <module>()
      3 x = tt.dmatrix('x')
----> 4 y = np.exp(x)

AttributeError: 'TensorVariable' object has no attribute 'exp'
```

Would it be possible to give a hint here, that perhaps `theano.tensor.exp` is what the user intended?

nouiz:
This particular case can be fixed by adding the exp function in the right
class (TensorVariable).

But this is related to how numpy execute its function on unknow inputs.
This changed in recent version of numpy and will continue to change to make
this easier. So this isn't only related to Theano.

The simplest is that each time we see such error reported, we check if
adding a new method would solve this. Otherwise, someone (I don't have the
time) could search in the np namespace the one that don't work and we could
work on that.

Can you help on that? If you can make a list of the method that don't work
and the error, I could test and add the missing method if that fix it.

aflaxman:
I think I know how to search the numpy namespace, so I'll try and report back here.  Can you point me to the numpy code that executes a function on unknown inputs?  I took a look in the code last night, but I didn't find it in the time I had.  Thanks!

nouiz:
That functionality in numpy is very complicated. I fixed one case in that
past, when you where doing a_numpy_ndarray + a_theano_tensor to call
a_theano_tensor + a_numpy_ndarray as Theano know how to handle that.

But it was hard to find and do. It took me 2 days to get a first version
and probably 2 other days to finish and get the fix merged in Theano.

There is discussion on that in numpy mailing list. I don't have time to
follow this. If you want to get in that, I would suggest you to find those
discussion first.

But in all case, this could need some change in Theano. numpy need to find
[...]

PR #3112, opened 2015-07-04
[WIP] Allow numpy.exp(theano_var) to build a graph.

nouiz:
This demonstrate how to fix gh-3040.

### DIFF
--- a/theano/tensor/tests/test_var.py
+++ b/theano/tensor/tests/test_var.py
@@ -10,7 +10,6 @@ def test_numpy_method():
     x = tt.dmatrix('x')
     data = np.random.rand(5, 5)
     for fct in [np.exp]:
-        print fct
         y = fct(x)
         f = theano.function([x], y)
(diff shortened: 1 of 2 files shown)


theano/tensor/tests/test_var.py
```diff
@@ -0,0 +1,16 @@
...
+import theano.tensor as tt
+import theano.tests.unittest_tools as utt
+
+
+def test_numpy_method():
+    # This type of code is used frequently by PyMC3 users
+    x = tt.dmatrix('x')
+    data = np.random.rand(5, 5)
+    for fct in [np.exp]:
+        print fct
```
lamblin: This is not valid in Python 3, which is why the tests are failiing.

nouiz: I updated the PR.

carriepl:
Seems fine to me. Plus @lamblin already had a look in the past so I'm merging.

PR #3114, opened 2015-07-05
Changed itertools.izip to izip for compatibility

vikkamath:
where izip is from theano.compat


lamblin:
Thanks!
This is obviously a good fix, but it means that this part of the code was not tested in python 3 (otherwise, that test would have been failing).
Would it be possible for you to share a minimal example that reproduces the problem, so we can add it as a unit test?

vikkamath:
A blocks [build](https://github.com/mila-udem/blocks/pull/748#issuecomment-118623964) failed because of this. We're still looking into the exact statements that raised the error, but sure, I'll keep you posted.

PR #3117, opened 2015-07-07
OpFromGraph.infer_shape()

ChienliMa:
Only a draft.  Simple test pass but have many warning.

```Python
WARNING (theano.tensor.opt): Failed to infer_shape from Op <theano.compile.builders.OpFromGraph object at 0x10c5bb590>.
Input shapes: [(Shape_i{0}.0, Shape_i{1}.0), (Shape_i{0}.0, Shape_i{1}.0), (Shape_i{0}.0, Shape_i{1}.0)]
Exception encountered during infer_shape: <type 'exceptions.TypeError'>
Exception message: ('Bad input argument to theano function with name "/Users/ChienliMa/Code/Theano/theano/tensor/utils.py:87"  at index 0(0-based)', 'Expected an array-like object, but found a Variable: maybe you are trying to call a function on a (possibly shared) variable instead of a numeric array?')
Traceback: Traceback (most recent call last):
  File "/Users/ChienliMa/Code/Theano/theano/tensor/opt.py", line 1079, in on_import
    [self.shape_of[r] for r in node.inputs])
  File "/Users/ChienliMa/Code/Theano/theano/compile/builders.py", line 144, in infer_shape
    all_shapes = theano.tensor.utils.shape_of_variables(fg, in_shapes)
  File "/Users/ChienliMa/Code/Theano/theano/tensor/utils.py", line 97, in shape_of_variables
    numeric_output_dims = compute_shapes(*numeric_input_dims)
  File "/Users/ChienliMa/Code/Theano/theano/compile/function_module.py", line 531, in __call__
    allow_downcast=s.allow_downcast)
  File "/Users/ChienliMa/Code/Theano/theano/tensor/type.py", line 77, in filter
    'Expected an array-like object, but found a Variable: '
TypeError: ('Bad input argument to theano function with name "/Users/ChienliMa/Code/Theano/theano/tensor/utils.py:87"  at index 0(0-based)', 'Expected an array-like object, but found a Variable: maybe you are trying to call a function on a (possibly shared) variable instead of a numeric array?')
```

### DIFF
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -145,6 +145,10 @@ class OpFromGraph(gof.Op):
         shape = theano.scan_module.scan_utils.infer_shape(self.new_outputs, 
                                                           self.new_inputs,
                                                           shapes)
+
+        import pdb
+        pdb.set_trace()
+
         return shape
 
     def grad(self, inputs, output_grads):
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -4,8 +4,12 @@ from theano.compat import izip
 from theano.compile.function_module import orig_function
 from theano.compile import SharedVariable, rebuild_collect_shared
 from theano.gof import ops_with_inner_function
+<<<<<<< HEAD
 from theano.gof.graph import io_connection_pattern
 
+=======
+from theano.gof import graph, FunctionGraph
+>>>>>>> draft of infer_shape
 
 class OpFromGraph(gof.Op):
     """This creates an `Op` from inputs and outputs lists of variables.
@@ -140,6 +144,17 @@ class OpFromGraph(gof.Op):
         Return connection pattern of subfgraph defined by inputs and outputs
         """
         return io_connection_pattern(self.new_inputs, self.new_outputs)
+
+    def infer_shape(self, node, shapes):
+        # Construct a new fgraph
+        equiv = graph.clone_get_equiv(self.new_inputs, self.new_outputs)
+        replacement = dict((equiv[var], var) for var in self.new_inputs)
+        out_cpy = theano.clone([equiv[var] for var in self.new_outputs], replace=replacement)
+        fg = FunctionGraph(self.new_inputs, out_cpy)
+
+        in_shapes = dict([(v, shp) for v, shp in zip(fg.inputs, shapes)])
+        all_shapes = theano.tensor.utils.shape_of_variables(fg, in_shapes)
+        return [all_shapes[var] for var in fg.outputs]
 
     def grad(self, inputs, output_grads):
         # OpFromGraph doesn't implement a connection_pattern, so for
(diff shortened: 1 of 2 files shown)


ChienliMa:
Sorry to update lately. I was stuck by  a small mistake.

ChienliMa:
I don't know why Travis CI didn't start building in this repo. But I think the code is ok now.

nouiz:
travis is overloaded. We need to make our tests faster.

theano/compile/tests/test_builders.py
```diff
@@ -151,6 +152,14 @@ def test_connection_pattern(self):
                          [True, False, True]]
         assert results == expect_result
 
+    def test_infer_shape(self):
+        x = T.matrix('x')
+        y = x+x
+        op_graph = OpFromGraph([x], [y], mode='FAST_RUN')
```
nouiz: why do you force FAST_RUN?

ChienliMa: I just follow the tests above. Actually this confused me too.

nouiz: try without it and run the test, by replacing theano-nose or nosetests by:

theano/misc/do_nightly_buildbot

this will test with many modes. Do this before updating the PR.

theano/compile/builders.py
```diff
@@ -141,6 +141,11 @@ def connection_pattern(self, node):
         """
         return io_connection_pattern(self.new_inputs, self.new_outputs)
 
+    def infer_shape(self, node, shapes):
+        return theano.scan_module.scan_utils.infer_shape(self.new_outputs,
+                                                         self.new_inputs,
+                                                         shapes)
```
carriepl: Using scan_utils.infer_shape returns symbolic shapes potentially computed using the inputs of the inner function but we need them to be computed using the inputs of the OpFromGraph in the outer function. This means that you need to do a clone (with replace) on the output of scan_utils.infer_shape to replace the inner inputs by the outer inputs.

lamblin:
FYI, Travis passed on this.

ChienliMa:
Sorry for not seeing your code block. I update the codes now.

nouiz:
@carriepl, I think this is good, can you review too?

carriepl:
I think this is fine. There aren't that many cases covered by the shape tests but since it's reusing a good chuck of Scan's infer_shape implementation which is more tested, it should be fine.

Since it's now used by both Scan and OpFromGraph, the infer_shape implementation in scan_utils should probably be move to a more appropriate location like the method to infer the connection pattern of a Theano graph was moved in an earlier PR.

ChienliMa:
I cannot move ```infer_shape``` from graph because it makes use of some high level feature like ```tensor``` and ```shape_feature```. What about putting it in ```tensor/utils```

nouiz:
I taught about this. If we try to move the dependencies too, it just get complicated and cause others problem. So don't move it.

thanks, I'll merge now.

PR #3118, opened 2015-07-08
OpFromGraph.connection_pattern()

ChienliMa:


theano/compile/builders.py
```diff
@@ -1,8 +1,10 @@
+import numpy
```
carriepl: There should be an empty line after this import (Pep8)

theano/compile/builders.py
```diff
@@ -1,8 +1,10 @@
...
 import theano
 from theano import gof
 from theano.compile.function_module import orig_function
 from theano.compile import SharedVariable, rebuild_collect_shared
 from theano.gof import ops_with_inner_function
+from theano.gof import FunctionGraph
```
carriepl: This import can be merged with the import on the previous line

theano/compile/tests/test_builders.py
```diff
@@ -107,6 +107,19 @@ def test_shared_grad(self):
         fn = function([x, y, z], f)
         assert numpy.allclose(15.0 + s.get_value(),
                               fn(xv, yv, zv))
+    
+    def test_connection_pattern(self):
+        import numpy
```
carriepl: Why is numpy imported here?

theano/compile/tests/test_builders.py
```diff
@@ -107,6 +107,19 @@ def test_shared_grad(self):
...
+        x, y, z = T.matrices('xyz')
+        out1 = x * y
+        out2 = y * z
+
+        op = OpFromGraph([x ,y, z], [out1, out2], moe='FAST_RUN')
```
carriepl: "moe" => "mode"

theano/compile/tests/test_builders.py
```diff
@@ -107,6 +107,19 @@ def test_shared_grad(self):
...
+        results = op.connection_pattern(None)
+        expect_result = [[True, False],
+                         [True, True],
+                         [False, True]]
+        assert results == expect_result
```
carriepl: This test is very basic. To offer better coverage, there are additional cases it should test, including; 
- Inner graph with ops that don't have a 'full' connection pattern (meaning not every output is connected to every input)
- Inner graph with ops that have multiple outputs (a nested OpFromGraph, for instance)
- Inner graph where some computation doesn't rely on explicit inputs to the OpFromGraph (constants and random number sampling)

theano/compile/builders.py
```diff
@@ -133,6 +135,59 @@ def perform(self, node, inputs, outputs):
             # we wont need this copy anymore
             output[0] = variable.copy()
 
+    def connection_pattern(self, node):
```
carriepl: This function appears to do the exact same thing (infer the connection pattern from a set of 'input variables' and a set of 'output variables') as the one in the class `Scan` (in scan_op.py). If this is indeed the case, we should avoid the code duplication and have both Scan and OpFromGraph rely on the same implementation which would be in a central location (perhaps somewhere in theano.gof since this is where the class FunctionGraph is defined).

carriepl:
This pull requests includes many merge commits that add nothing to the PR. These should not be here; the PR should only include the commit ' f058a1e'.

theano/compile/builders.py
ChienliMa: What I implemented is indeed the ```thunks``` that using dimensions of variables as inputs and outputs to calculate to shape. This allow to reuse the optimized fgraph and bypass the optimization. So it is similar to every ```make_thunk``` method in every ```Linker``` and ```ScanOp```, if you are talking about this. Maybe I can abstract the code that go over the fgraph and do the calculation, as I have similar operation to each node.

I cannot find any ``` connection_pattern``` in ```ScanOp```.

theano/compile/tests/test_builders.py
ChienliMa: I added another 2 test cases. Test is not passed.
Yet I don't know how theano deal with sharedvariable in connection_pattern().
If the ```expected_result``` in the last case right? If not, please provide the right one.

theano/compile/builders.py
carriepl: I am really not sure what you are referring to. The connection_pattern() method is not about computing shapes but rather knowing (mainly for gradient reasons) what inputs the value of each output depends on.

The connection pattern in ScanOp can be found here : https://github.com/Theano/Theano/blob/master/theano/scan_module/scan_op.py#L1539. The first thing it does is call inner_connection_pattern() which infers the connection pattern of the inner graph (exactly the same thing as you are doing).

ChienliMa: Sorry I mixed  two PR.  Buy they are quit the same.
I will have a look and see if i can merge two codes. Can you provide some suggestion for test case?

carriepl:
From the Travis output, is seems the code has issues with Python 3, flake8 errors and the expected output for one of the tests is not the output that you are getting for that test.

ChienliMa:
Can't fall asleep now. :(
I will try to merge the code into a function called pattern_from_graph
tomorrow.Don't worry about those errors,  i will fix them later.
I will reuse test case for scanop.
I didn't dive into those code yet, but please leave a comment here if you
have any idea for testcase.

2015-07-10 1:04 GMT+08:00 carriepl <notifications@github.com>:

theano/compile/builders.py
carriepl: Unless the existing implementation has issues, it should probably be simply reused; I don't think there's much point in changing code without a purpose.

nouiz:
thanks (reviewd by @carriepl)

PR #3120, opened 2015-07-08
Fix CLinker

lamblin:
Fix issue reported at https://groups.google.com/d/topic/theano-users/6dLaEqc2R6g/discussion
The test brings up a different problem.

### DIFF
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -559,8 +559,10 @@ class CLinker(link.Linker):
             #            what to do at the beginning of each run,
             #            what to do at the end of each run]]
             if variable in self.inputs:
-                # we need to extract the new inputs at each run
-                # they do not need to be relayed to Python, so we don't sync
+                # We need to extract the new inputs at each run
+                # they do not need to be relayed to Python, so we don't sync.
+                # If the variable is both an input and an output, there is
+                # no need to synchronize either, it is already up-to-date.
                 policy = [[get_nothing, get_nothing, get_nothing],
                           [get_c_declare, get_c_extract, get_c_cleanup]]
             elif variable in self.orphans:
@@ -972,6 +974,10 @@ class CLinker(link.Linker):
         if output_storage is None:
             map = {}
             output_storage = []
+            # Initialize the map with the inputs, as some outputs may
+            # be inputs as well.
+            for i, variable in enumerate(self.inputs):
+                map[variable] = input_storage[i]
             for variable in self.outputs:
                 if variable not in map:
                     map[variable] = [None]
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -282,10 +282,13 @@ def get_nothing(r, name, sub):
 
 def get_c_declare(r, name, sub):
     """Wrapper around c_declare that declares py_name"""
-
     if any([c != "output" and getattr(c.op, 'check_input',
            config.check_input) for (c, _) in r.clients]) or (
             r.owner and getattr(r.owner.op, 'check_input', True)):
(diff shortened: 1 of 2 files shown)


lamblin:
Fixes #3121 now.

nouiz:
this pr need a rebase.

lamblin:
@nouiz I tried implementing the change we talked about yesterday, and it made the number of compiled op go up 200 (from around 4800 to 5000). The Ops in question are below.

After re-reading the code and the original PR (#1884) , I'm pretty sure we do not need to have `check_inputs=True` for outputs. In fact, I think that if _clients_ need check_inputs, we should only call `c_declare` and `c_extract_out` with `check_inputs=True`, not `c_extract`. However, if _this_ op needs check_input, we should call `c_extract`, `c_extract_out` and `c_declare` with `check_inputs=True`, but currently we do not always...
I'll try that and see if it works.

```
# Op  before  after
<class 'theano.compile.ops.DeepCopyOp'> 3 15
<class 'theano.compile.ops.OutputGuard'> 1 2
<class 'theano.compile.ops.Rebroadcast'> 13 21
<class 'theano.tensor.basic.Flatten'> 6 9
<class 'theano.tensor.basic.Join'> 13 20
<class 'theano.tensor.basic.Reshape'> 21 33
<class 'theano.tensor.elemwise.DimShuffle'> 119 195
<class 'theano.tensor.opt.Assert'> 12 19
<class 'theano.tensor.subtensor.IncSubtensor'> 103 122
<class 'theano.tensor.subtensor.Subtensor'> 98 143
```

lamblin:
In the end, the only change I did was to use the config flag in `get_c_declare` as well. I still have the smaller number of Ops in the cache, which is comparable to the master.

theano/gof/cc.py
```diff
@@ -306,13 +307,15 @@ def get_c_init(r, name, sub):
...
+            for (c, _) in r.clients
+            if not isinstance(c, string_types)]):
         # check_broadcast is just an hack to easily remove just the
         # broadcast check on the old GPU back-end. This check isn't
         # done in the new GPU back-end or on the CPU.
-        if any([getattr(c.op, 'check_broadcast', True) for (c, _) in
-                r.clients]):
+        if any([getattr(c.op, 'check_broadcast', True)
+                for (c, _) in r.clients
+                if not isinstance(c, string_types)]):
```
nouiz: this is is useless, but no need to change it.

theano/gof/cc.py
```diff
@@ -306,13 +307,15 @@ def get_c_init(r, name, sub):
 
 def get_c_extract(r, name, sub):
     """Wrapper around c_extract that initializes py_name from storage."""
-    if any([getattr(c.op, 'check_input', config.check_input) for (c, _) in
-            r.clients]):
+    if any([getattr(c.op, 'check_input', config.check_input)
+            for (c, _) in r.clients
+            if not isinstance(c, string_types)]):
```
nouiz: Do you understand why there is no check like `r.owner and getattr(r.owner.op, 'check_input', True)` like in get_c_declare? This could cause some check to be skiped.

I think it would be better to comment why this is like this. It took both of us some times to understand what it need to be.

theano/gof/cc.py
lamblin: Why is it useless? In `r.clients`, there could be `'output'` and an op with `'check_input`'. In that case, there would be an AttributeError when trying to access `'output'.op`.

theano/gof/cc.py
lamblin: You are right, I'll add comments.

lamblin: Comments added.

PR #3125, opened 2015-07-09
Flake8 misc

harlouci:


nouiz:
travis is failing.

nouiz:
please, don't use description like v7, better use the subdirectory name as you did in other PR.

theano/misc/pycuda_example.py
```diff
@@ -349,8 +346,8 @@ def make_thunk(self, node, storage_map, _, _2):
         def thunk():
             z = outputs[0]
             if (z[0] is None or
-                z[0].shape != inputs[0][0].shape or
-                not z[0].is_c_contiguous()):
+                    z[0].shape != inputs[0][0].shape or
+                    not z[0].is_c_contiguous()):
                 z[0] = theano.sandbox.cuda.CudaNdarray.zeros(
                     inputs[0][0].shape)
             if inputs[0][0].shape != inputs[1][0].shape:
```
lamblin: Should that function end with `z[0] = out`?
Flake8 complains that `out` is unused, and it may be right...

theano/misc/pycuda_example.py
```diff
@@ -384,7 +381,7 @@ def local_pycuda_gpu_elemwise(node):
     """
     if isinstance(node.op, GpuElemwise):
         if (not any([any(i.type.broadcastable) for i in node.inputs]) and
-            all([i.ndim <= 2 for i in node.inputs])):
+                all([i.ndim <= 2 for i in node.inputs])):
             new_op = PycudaElemwiseSourceModuleOp(node.op.scalar_op,
                                                   node.op.inplace_pattern)(
                                                       *node.inputs)
```
lamblin: Line 401 below uses PycudaElemwiseKernelOp, but its definition is commented out.
Should we comment out that function as well?

lamblin:
In addition, the following errors show up with python3:
```
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/misc/check_blas.py:55:14: E211 whitespace before '('
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/misc/check_blas.py:263:14: E211 whitespace before '('
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/misc/cudamat_utils.py:97:24: F821 undefined name 'long'
/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/misc/gnumpy_utils.py:111:24: F821 undefined name 'long'
```

harlouci:
additional flake8 for check_blas.py

lamblin:
The same errors with python 3 are still there.
Python 3 is a little bit stricter than python 2, this is why you do not see those errors when you manually run `flake8` on your machine. However, you still need to fix them.

harlouci:
OK. I did a rebase.

When I run
"flake8 --ignore E501 theano/misc/check_blas.py" or
"theano-nose theano/tests/test_flake8.py",
I get OK in both cases.

In file theano/misc/check_blas.py, I can see the whitespaces before '(',
but I don't see the undefined names 'long'. Where do Travis see them ? and
how ?

2015-07-22 19:46 GMT-04:00 Pascal Lamblin <notifications@github.com>:

lamblin:
I know you get OK when you run flake8, because your Python is Python 2. However, Travis also tests with Python 3, and with Python 3 these are errors, so you still have to fix them.
Regarding "long", they do not appear in "check_blas.py", but in cudamat_utils.py and gnumpy_utils.py, as written in the error message:
- https://github.com/harlouci/Theano/blob/flake8_v7/theano/misc/cudamat_utils.py#L97
- https://github.com/harlouci/Theano/blob/flake8_v7/theano/misc/gnumpy_utils.py#L111

The problem is that `long` exists in Python 2 but not Python 3.
I think you can add, add the beginning of those two files (after the imports):
```python
if PY3:
    long = int
```
@nouiz, does that seem reasonable?

nouiz:
that seem good.

nouiz:
There is still flake8 errors.

harlouci:
Let see if importing six settle the errors.

harlouci:
*let's

nouiz:
I think it should work.

harlouci:
Additional corrections recently added.

nouiz:
when travis pass, we can merge.

PR #3128, opened 2015-07-10
Flake8 sparse

harlouci:


harlouci:
Additional corrections to sparse/basic.py

Issue #3115, opened 2015-07-06
No such file or directory: 


cancan101:
I have the following set:

```
optimizer_including = cudnn
assert_no_cpu_op=warn
```

and the cache directory is on an nfs mount.

I am getting the following error all though most of the time (not all of the time), the process continues. Further cuDNN should be available on the box(es) (error goes away when the cache dir path is no longer on NFS):

```
ERROR (theano.gof.opt): SeqOptimizer apply <theano.sandbox.cuda.dnn.NoCuDNNRaise object at 0x2b6ff3f58550>
ERROR (theano.gof.opt): Traceback:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/home/ubuntu/git/Theano/theano/gof/opt.py", line 193, in apply
    sub_prof = optimizer.optimize(fgraph)
  File "/home/ubuntu/git/Theano/theano/gof/opt.py", line 78, in optimize
    ret = self.apply(fgraph, *args, **kwargs)
  File "/home/ubuntu/git/Theano/theano/sandbox/cuda/dnn.py", line 1714, in apply
    if not dnn_available():
  File "/home/ubuntu/git/Theano/theano/sandbox/cuda/dnn.py", line 75, in dnn_available
    v = version()
  File "/home/ubuntu/git/Theano/theano/sandbox/cuda/dnn.py", line 207, in version
    profile=False)
  File "/home/ubuntu/git/Theano/theano/compile/function.py", line 310, in function
    output_keys=output_keys)
  File "/home/ubuntu/git/Theano/theano/compile/pfunc.py", line 523, in pfunc
    output_keys=output_keys)
  File "/home/ubuntu/git/Theano/theano/compile/function_module.py", line 1526, in orig_function
    defaults)
  File "/home/ubuntu/git/Theano/theano/compile/function_module.py", line 1390, in create
    input_storage=input_storage_lists)
  File "/home/ubuntu/git/Theano/theano/gof/link.py", line 606, in make_thunk
    output_storage=output_storage)[:3]
  File "/home/ubuntu/git/Theano/theano/gof/vm.py", line 1025, in make_all
    no_recycling))
  File "/home/ubuntu/git/Theano/theano/sandbox/cuda/__init__.py", line 260, in make_thunk
    compute_map, no_recycling)
  File "/home/ubuntu/git/Theano/theano/gof/op.py", line 805, in make_thunk
    no_recycling)
[...]

PR #3135, opened 2015-07-14
Don't remove compilelock we don't own

nouiz:
test fix for gh-3115


abergeron:
I wholeheartedly approve of these changes since we can get rid of \__del__.

nouiz:
I just updated this PR.

PR #3136, opened 2015-07-14
Props

harlouci:
Added __props__, removed __eq__ and __hash__ to tensor/extra_ops.py

+ a flake8 commit


nouiz:
This PR contain additional commit then props. remove them.

nouiz:
Otherwise, it seem good. Do this in for all Theano op.

PR #3137, opened 2015-07-14
Fix memory leak of descriptors in cudnn bindings.

abergeron:
This fixes the memory leak observed at https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/lasagne-users/8iXu7NcAxuY/eI72oHkslGcJ and another one involving pooling descriptors.

There seems to be another leak in cuda-convnet, which this does not fix.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -256,7 +256,8 @@ class GpuDnnConvDesc(GpuOp):
             raise TypeError('kern must be 1D shape tensor')
 
         return Apply(self, [img_shape, kern_shape],
-                     [CDataType("cudnnConvolutionDescriptor_t")()])
+                     [CDataType("cudnnConvolutionDescriptor_t",
+                                freefunc="cudnnDestroyConvolutionDescriptor")()])
 
     def c_code(self, node, name, inputs, outputs, sub):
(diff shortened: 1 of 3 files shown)


theano/sandbox/cuda/dnn.py
```diff
@@ -256,7 +256,8 @@ def make_node(self, img_shape, kern_shape):
             raise TypeError('kern must be 1D shape tensor')
 
         return Apply(self, [img_shape, kern_shape],
-                     [CDataType("cudnnConvolutionDescriptor_t")()])
+                     [CDataType("cudnnConvolutionDescriptor_t",
+                                freefunc="cudnnDestroyConvolutionDescriptor")()])
```
nouiz: this do not trigger a new compilation, so to have the fix, we need to delete the compiledir.

The easy fix, to bump those op c_code_cache_version. The better fix, to add in CDataType.c_code_cache_version, the cdata type and freefunc.

I made a PR to your branch: https://github.com/abergeron/Theano/pull/20

nouiz:
This fix is also needed in the new back-end.

abergeron:
Done.

nouiz:
merge now as travis don't test GPU.

Issue #2792, opened 2015-04-18
Follow up Average pooling


nouiz:
gh-2783 added average pooling on the CPU. Here is what is left:
- [x] Refactor to have the grad average in a different op that do not that the input of the forward as input (lower memory usage) (The refactor, rename the ops, have a common PoolGrad op that contain the common code for AveragePoolGrad and MaxPoolGrad)
- [x] the grad need support for pooling
- [x] need c code on the CPU.
- [x] no grad of grad (it return grad not implemented)
- [x] on the GPU, this is only supported by cudnn.
- [ ] average grad python code don't support padding (but c code does).

lamblin:
@nouiz I checked the box for "refactor" and "no grad of grad", but I'm not sure. Can you check when you are back?

dwf:
So, there's no average pooling on GPU yet? Is there no optimization that replaces CPU average pooling (via `max_pool_2d`) with CuDNN average pooling?

lamblin:
There is average pooling on GPU via CuDNN, in `theano.sandbox.cuda.dnn.dnn_pool(..., mode='average_inc_pad')` (or `exc_pad` if you prefer).
However, you are right, there is no optimization replacing the CPU op... I'm adding that to the list.

dwf:
Okay, and it should probably warn or flat out error if the replacement fails and your GPU graph is executing partly on CPU, since I think, the majority of the time, this is going to make a model intolerably slow.

lamblin:
Well, actually, I think I was mistaken because the CPU Op is confusingly named `DownsampleFactorMax`, even when performing average pooling... according to the unit tests, average pooling should be transfered to the GPU.
For your second point, you can use [`config.assert_no_cpu_op`](http://deeplearning.net/software/theano/library/config.html#config.assert_no_cpu_op) to detect at least some of these cases.

dwf:
Yes, and you very bizarrely must call max_pool_2d with a particular mode
argument to get average pooling. That probably isn't winning any API design
awards...

lamblin:
#3592

dwf:
@lamblin here's my debugprint:

https://gist.github.com/dwf/054b1ff9ec915a43dd10

Seems like it's not getting transferred even when it's max pooling. WTF.

fchollet:
Is it still the case that there is no Theano average pooling op for GPU?

dwf:
[...]

PR #3139, opened 2015-07-15
Average pooling

sebastien-j:
@nouiz, resolves issue #2792 (WIP).

- I used the names given in issue #2792, although I feel that they clash too much against the current nomenclature.

- There is probably too much code duplication.

- Should I write the C code for `AveragePoolGrad`?

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -12,7 +12,7 @@ from theano.compile.ops import shape_i
 from theano.configparser import AddConfigVar, EnumStr
 from theano.tensor.nnet import SoftmaxGrad
 from theano.tensor.signal.downsample import (
-    DownsampleFactorMax, DownsampleFactorMaxGrad)
+    DownsampleFactorMax, MaxPoolGrad, AveragePoolGrad)
 from theano.sandbox.cuda import GpuOp
 from theano.sandbox.cuda.basic_ops import (as_cuda_ndarray_variable,
                                            host_from_gpu,
@@ -1672,11 +1672,11 @@ if True:
                                      desc)]
 
     @register_opt('cudnn')
-    @local_optimizer([DownsampleFactorMaxGrad])
+    @local_optimizer([MaxPoolGrad])
     def local_pool_dnn_grad_stride(node):
         if not dnn_available():
             return
-        if isinstance(node.op, DownsampleFactorMaxGrad):
+        if isinstance(node.op, MaxPoolGrad):
             if not node.op.ignore_border:
                 return
             inp, out, inp_grad = node.inputs
@@ -1692,6 +1692,31 @@ if True:
                 desc = GpuDnnPoolDesc(ws=ds, stride=st, mode=mode, pad=pad)()
                 ret = GpuDnnPoolGrad()(gpu_contiguous(inp),
                                        gpu_contiguous(out),
+                                       gpu_contiguous(inp_grad),
+                                       desc)
+                return [host_from_gpu(ret)]
+
+    @register_opt('cudnn')
+    @local_optimizer([AveragePoolGrad])
+    def local_avgpool_dnn_grad_stride(node):
+        if not dnn_available():
+            return
+        if isinstance(node.op, AveragePoolGrad):
+            if not node.op.ignore_border:
+                return
+            inp, inp_grad = node.inputs
+            ds = node.op.ds
+            st = node.op.st
+            pad = node.op.padding
+            mode = node.op.mode
+
+            if ((inp.owner and isinstance(inp.owner.op, HostFromGpu)) or
+                (inp_grad.owner and isinstance(inp_grad.owner.op,
+                                               HostFromGpu))):
+                desc = GpuDnnPoolDesc(ws=ds, stride=st, mode=mode, pad=pad)()
+                ret = GpuDnnPoolGrad()(gpu_contiguous(inp),
+                                       gpu_contiguous(numpy.empty((1,1,1,1),
+                                           dtype=numpy.float32)),
                                        gpu_contiguous(inp_grad),
                                        desc)
                 return [host_from_gpu(ret)]
--- a/theano/tensor/signal/downsample.py
+++ b/theano/tensor/signal/downsample.py
@@ -313,8 +313,7 @@ class DownsampleFactorMax(Op):
             maxout = self(x)
             return [MaxPoolGrad(self.ds,
                                             ignore_border=self.ignore_border,
-                                            st=self.st, padding=self.padding,
-                                            mode=self.mode)(
+                                            st=self.st, padding=self.padding)(
                                                 x, maxout, gz)]
         else:
             return [AveragePoolGrad(self.ds,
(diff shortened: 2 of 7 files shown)


sebastien-j:
@nouiz ,

should I also remove 'mode' and related checks from `DownsampleFactorMaxGradGrad`?

theano/tensor/signal/downsample.py
```diff
@@ -834,7 +968,7 @@ def __init__(self, ds, ignore_border, st=None, padding=(0,0), mode='max'):
 
     def make_node(self, x, maxout, gz):
         # make_node should only be called by the grad function of
-        # DownsampleFactorMaxGrad, so these asserts should not fail.
+        # MaxPoolGrad, so these asserts should not fail.
         assert isinstance(x, Variable) and x.ndim == 4
         assert isinstance(maxout, Variable) and maxout.ndim == 4
         assert isinstance(gz, Variable) and gz.ndim == 4
```
nouiz: Can you move the assert after the call to as_tensor_variable?, Then you can only assert ndim, and remove the isinstance.

theano/tensor/signal/downsample.py
```diff
@@ -845,8 +979,6 @@ def make_node(self, x, maxout, gz):
         return Apply(self, [x, maxout, gz], [x.type()])
 
     def perform(self, node, inp, out):
-        if self.mode != 'max':
-            raise theano.gof.utils.MethodNotDefined()
```
nouiz: As the perform only support max, for safety, add an assert in the \__init__ as you removed it here.

theano/tensor/signal/tests/test_downsample.py
```diff
@@ -417,12 +417,35 @@ def test_DownsampleFactorMaxGrad_grad(self):
...
+                        imval.shape, avgpoolshp, ignore_border=ignore_border)
+                    grad_val = rng.rand(*grad_shape) * 10.0
+
+                    def mp(input, grad):
+                        grad_op = AveragePoolGrad(
+                            avgpoolshp, ignore_border=ignore_border, mode=mode)
+                        return grad_op(input, grad)
+
+                    utt.verify_grad(mp, [imval, grad_val], rng=rng)
+
```
nouiz: Good catch that there where missing tests.

theano/tensor/signal/downsample.py
```diff
@@ -518,6 +600,15 @@ def __init__(self, ds, ignore_border, st=None, padding=(0, 0), mode='max'):
                 " 'average_inc_pad' and 'average_exc_pad'. Got %s" % mode)
         self.mode = mode
 
+    def infer_shape(self, node, in_shapes):
+        return [in_shapes[0]]
+
+
+class MaxPoolGrad(PoolGrad):
```
nouiz: I think we need to be able to reload old pickled graph. I talked with @lamblin about this and this is the best way we found to do this:

DownsampleFactorMaxGradGrad = MaxPoolGrad

And allow in \__init__ and grad all the mode in that class, but not perform (add an assert in perform and c code to not compute bad value).

then make a quick optimization that would convert MaxPoolGrad to AveragePoolGrad. Register it in the canonizer phase.

Do you need more details? I forgot if you already did an optimization, the description is long, but the code is short.

thanks
bellow this class.

theano/sandbox/cuda/dnn.py
```diff
@@ -1697,6 +1697,31 @@ def local_pool_dnn_grad_stride(node):
...
+            pad = node.op.padding
+            mode = node.op.mode
+
+            if ((inp.owner and isinstance(inp.owner.op, HostFromGpu)) or
+                (inp_grad.owner and isinstance(inp_grad.owner.op,
+                                               HostFromGpu))):
+                desc = GpuDnnPoolDesc(ws=ds, stride=st, mode=mode, pad=pad)()
+                ret = GpuDnnPoolGrad()(gpu_contiguous(inp),
+                                       gpu_contiguous(numpy.empty((1,1,1,1),
+                                           dtype=numpy.float32)),
```
nouiz: Did you test that this work? @carriepl, this is a simpler solution for cudnn and I like it, if you didn't test what I told you, don't test it for now.

sebastien-j: The tests in `theano/sandbox/cuda/tests/test_dnn.py` passed.

nouiz: great. we like it that way. This bring the memory optimization to the GPU with very few modification to the code.

theano/tensor/signal/downsample.py
sebastien-j: I don't really follow and I never did an optimization before. Some extra details would be helpful.

theano/sandbox/cuda/dnn.py
sebastien-j: In the tests, mode 'sum' is excluded. Furthermore, for average pooling, only the case `pad=(0,0)` is tested.

I think these conditions should be in dnn.py. Am I right?

theano/tensor/signal/downsample.py
```diff
@@ -309,13 +309,18 @@ def infer_shape(self, node, in_shapes):
...
+            return [MaxPoolGrad(self.ds,
+                                            ignore_border=self.ignore_border,
+                                            st=self.st, padding=self.padding)(
+                                                x, maxout, gz)]
+        else:
+            return [AveragePoolGrad(self.ds,
+                                            ignore_border=self.ignore_border,
+                                            st=self.st, padding=self.padding,
+                                            mode=self.mode)(
+                                                x, gz)]
```
lamblin: pep8 will not like that indentation

theano/tensor/signal/downsample.py
lamblin: If I understood correctly, we would keep a `DownsampleFactorMaxGrad` with all the properties it currently has, including `mode`, but not necessarily C code, infer_shape, etc.
Then, we add an optimization to replace `DownsampleFactorMaxGrad{..., mode='max'}` by `MaxPoolGrad` and `DownsampleFactorMaxGrad{..., mode='sum/average...'}` by `AveragePoolGrad`.

There are different options:
- we could use the base `PoolGrad` class for that
- we could use `MaxPoolGrad` as @nouiz suggested, since it already has the right number of inputs
- or we could use a new, empty class (maybe inheriting from `PoolGrad`?).
I'm not sure which would be best.

theano/sandbox/cuda/dnn.py
nouiz: I'm not sure what you mean for sum. For average pooling, pad != 0,0, should be tested. The python code don't support it, but the c code does.

theano/tensor/signal/downsample.py
nouiz: As you aren't sure, I would go with my suggestion :)

@sebastien-j for optimization, the doc is there:

https://github.com/abergeron/ccw_tutorial_theano

in the advanced.pdf file.

pseudo code:

```
@register_canonicalize
@gof.local_optimizer([MaxPoolGrad])
def local_average_pool_grad(node):
    if not isinstance(node.op, T.MaxPoolGrad) or node.op.mode not in ['average_inc*' , 'aevrage_ex*']:
        return False
    return [AveragePoolGrad(ds=node.op.ds, ignore_border=node.op.ignore_border, st=node.op.st, padding=node.op.padding, mode=node.op.mode)(node.inputs[0], node.inputs[2])]
```

sebastien-j: @nouiz , I added the optimization. However, tests fail with mode=FAST_COMPILE, where the optimization isn't applied. I suppose this is normal behaviour. Should I just skip the test in that case, or is there something better to do?

nouiz: We pass a mode that force that optimization to run. Pass this parameter to
theano.function

mode=Theano.compile.get_default_mode().including('optname')

sebastien-j: Ok, this is done.

nouiz:
With that small change, we can merge after the summer school.

sebastien-j:
Done. Should I squash my commits before the merge?

lamblin:
I don't think you need to rebase, there aren't that many commits, and there is no merge.
I'm trusting @nouiz for the review, and merging.

PR #3140, opened 2015-07-15
add larger matrix for check_blas

JesseLivezey:
GPU gemm comparisons are getting too fast. Added a default 5000x5000x5000 test.

### DIFF
--- a/theano/misc/check_blas.py
+++ b/theano/misc/check_blas.py
@@ -276,12 +276,28 @@ if __name__ == "__main__":
         GTX 750 Ti        
         """)
 
+    if options.M == 0:
+        M = options.B
+    else:
+        M = options.M
+    if options.N == 0:
+        N = options.B
+    else:
+        N = options.N
+    if options.K == 0:
+        K = options.B
+    else:
+        K = options.K
+
     t, impl = execute(not options.print_only, not options.quiet,
-                      M=options.M, N=options.N, K=options.K,
-                      iters=options.iter, order=options.order)
+                      M=M, N=N, K=K, iters=options.iter,
+                      order=options.order)
 
-
-    def print_timing(M, N, K, t, impl):
+    if options.print_only:
+        pass
+    elif options.quiet:
+        print(t)
+    else:
         print()
         print("We executed", options.iter, end=' ')
         print("calls to gemm with a and b matrices of shapes", end=' ')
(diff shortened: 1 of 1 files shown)


theano/misc/check_blas.py
```diff
@@ -143,6 +143,12 @@ def test():
                   help="The numpy memory layout parameter used when creating"
                   " the numpy.ndarray objects. It accepts 'C' for C memory"
                   " order and 'F' for Fortran order (for all matrices).")
+parser.add_option('--small_only', action='store_true', dest='small_only',
+                  default=False,
+                  help="If true, additionally perform gemm on large matrices.")
+parser.add_option('-B', '--B', action='store', dest='B',
+                  default=5000, type="int",
```
nouiz: I like that. But I would make it work a little differently. Can you make M,N and K default to 0. If they are 0, use the value of B, otherwise, use the corresponding value. Then you don't need the small_only parameter.

theano/misc/check_blas.py
```diff
@@ -240,26 +246,88 @@ def test():
         GT 220                                                             3.80s
         GT 210                                                      6.35s
         8500 GT                                                                   10.68s
+
+        Results for larger matrices.
+        There were 10 executions of gemm in float32
+        with matrices of shape 5000x5000 (M=N=K=5000).
+        All memory layout was in C order.
+
+        cuda version      6.5    6.0    5.5    5.0    4.2    4.1    4.0    3.2    3.0   # note
```
nouiz: Add only the columns 7.5 (annonced), 7.0, and 6.5.

I think it isn't useful to keep to old version now and we have the new.

theano/misc/check_blas.py
```diff
@@ -240,26 +246,88 @@ def test():
...
+        GTX Titan(D15U-50)
+        GTX 780           
+        GTX 980           
+        GTX 970           
+        GTX 680           
+        GRID K520         
+        GTX 580           
+        GTX 480           
+        GTX 750 Ti        
+        GTX 470
```
nouiz: I think I would remove all the one after 750 Ti. They probably won't be filled.

theano/misc/check_blas.py
JesseLivezey: I'm not sure I understand. Is your suggestion the same as making the default values 5000 rather than 2000? Or do you want 2000 to still also run?

lamblin:
Thanks!

PR #3146, opened 2015-07-15
Don't use type.make_variable() to have more stack trace

nouiz:


abergeron:
Sure, yeah.  I have no idea why we where using make_variable() in the first place.

nouiz:
I think it was done at one place and people copied it over as they took it as example.

PR #3147, opened 2015-07-15
Props compile

harlouci:
All __props__ checked in folder theano/compile/

### DIFF
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -65,6 +65,7 @@ class OpFromGraph(gof.Op):
         fn = function([x, y, z], [e2])
 
     """
+    __props__ = ()
 
     def __init__(self, inputs, outputs, **kwargs):
         if not isinstance(outputs, list):
@@ -100,14 +101,6 @@ class OpFromGraph(gof.Op):
         self.kwargs = kwargs
         self.input_types = [input.type for input in inputs]
         self.output_types = [output.type for output in outputs]
-
-    def __eq__(self, other):
-        # TODO: recognize a copy
-        return self is other
-
-    def __hash__(self):
-        # TODO: use internal variables in hash
-        return hash(type(self))
 
     def make_node(self, *inputs):
         for input, type in zip(inputs, self.input_types):
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -65,7 +65,6 @@ class OpFromGraph(gof.Op):
         fn = function([x, y, z], [e2])
 
     """
-    __props__ = ()
 
     def __init__(self, inputs, outputs, **kwargs):
         if not isinstance(outputs, list):
@@ -101,6 +100,14 @@ class OpFromGraph(gof.Op):
         self.kwargs = kwargs
         self.input_types = [input.type for input in inputs]
         self.output_types = [output.type for output in outputs]
+
+    def __eq__(self, other):
+        # TODO: recognize a copy
+        return self is other
+
+    def __hash__(self):
+        # TODO: use internal variables in hash
+        return hash(type(self))
 
     def make_node(self, *inputs):
         for input, type in zip(inputs, self.input_types):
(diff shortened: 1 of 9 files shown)


theano/compile/builders.py
```diff
@@ -101,14 +102,6 @@ def __init__(self, inputs, outputs, **kwargs):
         self.input_types = [input.type for input in inputs]
         self.output_types = [output.type for output in outputs]
 
-    def __eq__(self, other):
-        # TODO: recognize a copy
-        return self is other
```
nouiz: Keep \__eq__, it don't do the same as the default. And keep the comment. Also keep \__hash__ and its todo. remove \__props__ for this op, it is a special op.

theano/compile/ops.py
```diff
@@ -623,6 +603,7 @@ class Rebroadcast(gof.Op):
     c_code_and_version = {}
 
     check_input = False
+    __props__ = ("axis")
 
     def __init__(self, *axis):
         self.axis = dict(axis)
```
nouiz: In \__hash__ axis was sorted. I think we should use an OrderedDict, that we create empty and add the axis to it in the sorted order. That would make this Op better and allow to use \__props__ without regression.

theano/compile/ops.py
```diff
@@ -488,13 +475,6 @@ def __init__(self, fn, itypes, otypes, infer_shape):
         if self.__infer_shape is not None:
             self.infer_shape = self._infer_shape
 
-    def __eq__(self, other):
-        return (type(self) == type(other) and
-                self.__fn == other.__fn)
-
-    def __hash__(self):
-        return hash(type(self)) ^ hash(self.__fn)
-
```
nouiz: This is a special op. itypes and otypes are only used when building the graph, so they aren't "props" like other props. For now, don't change this op. We are missing what is needed to make it work.

theano/compile/ops.py
harlouci: Could you be more explicit, please. I don't figure out the solution.

nouiz: In the init, do this:
```
# Sort them to make sure we merge all possible case.
items = sorted(iteritems(self.axis))
self.axis = OrderedDict(items)
```

and use this import `from theano.compat import OrderedDict`

nouiz:
travis fail.

harlouci:
Inserting

 # Sort them to make sure we merge all possible case.
        items = sorted(iteritems(self.axis))
        self.axis = OrderedDict(items)

was your idea.

What is wrong about that ?

2015-07-17 9:21 GMT-04:00 Frédéric Bastien <notifications@github.com>:

lamblin:
@harlouci `self.axis` has not been assigned yet. You would have to use `axis` instead on the first line.

theano/compile/ops.py
```diff
@@ -623,21 +604,16 @@ class Rebroadcast(gof.Op):
     c_code_and_version = {}
 
     check_input = False
+    __props__ = ("axis",)
 
     def __init__(self, *axis):
-        self.axis = dict(axis)
+        # Sort them to make sure we merge all possible case.
+        items = sorted(iteritems(axis))
```
lamblin: Actually, axis is a list of pairs, so you cannot call iteritems on it.
I think that should be:
```python
items = sorted(axis)
self.axis = OrderedDict(items)
```
but please try it first.

harlouci: theano-nose theano/compile/tests/test_ops.py

says OK

lamblin:
So OrderedDict is not hashable either, so it cannot be part of props.
One solution would be to have two attributes: one is a list of pairs, one an OrderedDict (or even a dict, like currently). Only the list of pairs would be part of `__props__`, but the OrderedDict would be used at the other places.

nouiz:
I used a different strategy, now I enabled the hash of dict in props. What do you think of this? I think it is more general and easier to do.

nouiz:
I updated the PR.

PR #3149, opened 2015-07-16
changing local_sum_div_dimshuffle to local_sum_prod_div_dimshuffle

mohammadpz:
Regarding #399 
@carriepl , could you please take a look at my PR?


theano/tensor/tests/test_opt.py
```diff
@@ -4898,8 +4898,67 @@ def test_local_sum_div_dimshuffle(self):
...
+            for i, s in enumerate(prods):
+                print(i)
+                f = theano.function([a, b, c, d], s, mode=self.mode,
+                        on_unused_input='ignore')
+                g = f.maker.fgraph.toposort()
+                assert isinstance(g[-1].op.scalar_op,
+                                  theano.scalar.basic.TrueDiv)
+                f(a_val, b_val, c_val, d_val)
+        finally:
+            raise Exception("\'test_local_prod_div_dimshuffle\' failed!")
```
mohammadpz: @carriepl , Is this the right way of raising test failure?

theano/tensor/opt.py
```diff
@@ -3923,24 +3923,28 @@ def local_elemwise_sub_zeros(node):
...
-        # print 'axis =', axis
-        thing_summed = node.inputs[0]
-        if thing_summed.owner and thing_summed.owner.op == T.true_div:
-            numerator, denominator = thing_summed.owner.inputs
+        node_input = node.inputs[0]
+        if node_input.owner and node_input.owner.op == T.true_div:
+            numerator, denominator = node_input.owner.inputs
 
             # Old, bugged logic, reproduced here only to warn users
             if config.warn.sum_div_dimshuffle_bug:
```
mohammadpz: @carriepl , there is no ```prod_div_dimshuffle_bug```. Right?

theano/tensor/opt.py
```diff
@@ -3923,24 +3923,28 @@ def local_elemwise_sub_zeros(node):
...
+      if dimension l of the DimShuffle is 'x'
+    '''
 
     # It does not make much sense now to extend it to the case where the
     # dimshuffle is in the numerator, since elemwise inversion of the
-    # denominator would still be needed before the summation.
+    # denominator would still be needed before the summation or production.
 
-    if isinstance(node.op, T.Sum):
+    if isinstance(node.op, T.Sum) or isinstance(node.op, T.elemwise.Prod):
```
nouiz: isinstnace(node.op, (T.Sum, T.elemwise.Prod)):, this is faster

theano/tensor/opt.py
nouiz: Right. So add in the condition that node.op is T.Sum.

theano/tensor/tests/test_opt.py
nouiz: What are you triing to test? THis make the test always fail: it fail if the assert fail and it fail if the assert pass

mohammadpz: Should I remove ```try```?

nouiz: I think so.

theano/tensor/opt.py
```diff
@@ -3923,27 +3923,32 @@ def local_elemwise_sub_zeros(node):
...
+      if dimension l of the DimShuffle is 'x'
+    '''
 
     # It does not make much sense now to extend it to the case where the
     # dimshuffle is in the numerator, since elemwise inversion of the
-    # denominator would still be needed before the summation.
+    # denominator would still be needed before the summation or production.
 
-    if isinstance(node.op, T.Sum):
+    if isinstance(node.op, (T.SumT.elemwise.Prod)):
```
carriepl: You need a comma here between T.Sum and T.elemwise.Prod

mohammadpz: Sorry for this mistake. I resolved it.

theano/tensor/opt.py
```diff
@@ -3923,27 +3923,32 @@ def local_elemwise_sub_zeros(node):
...
+def local_sum_prod_div_dimshuffle(node):
     '''sum(a / dimshuffle{...}(b), axis=l) -> sum(a, axis={...}) / b,
-    if dimension l of the DimShuffle is 'x'.'''
-    # TODO: extend it to product, and quotient of products
+       if dimension l of the DimShuffle is 'x'
+
+       or
+
+       prod(a / dimshuffle{...}(b), axis=l) -> prod(a, axis={...}) / b,
+      if dimension l of the DimShuffle is 'x'
```
carriepl: I'm not sure that's quite right. Shouldn't that be 
`prod(a / dimshuffle{...}(b), axis=l) -> prod(a, axis={...}) / b ** shape(a)[0]`?
For instance, if a is [[1, 2], [2, 4]], b is [1,2] and l=0 :  
`prod(a / dimshuffle{...}(b), axis=l)` will give vector [2, 2]
`prod(a, axis={...}) / b` will give vector [2, 4]

mohammadpz: I'm a bit confused. what is the dimension of ```b```?

carriepl: in my example, b is a vector with 2 values: [1, 2].
Actually, I think the right optimization would be  :
prod(a, axis={...}) / dimshuffle{same as before don't add a new axis l}(b) ** shape(a)[0]

mohammadpz: What happens in this case?
```python
x = T.dtensor3()
y = T.dvector()
z = T.prod(x / y.dimshuffle(0, 'x', 'x'), axis=(0, 1))
```

nouiz: I expect the opt to generate the equivalent of:

T.prod(x / y.dimshuffle(0, 'x'), axis=(0,))

mohammadpz: Based our discussions with @carriepl , in the above-mentioned case, the optimization will return:
```python
T.prod(T.prod(x, axis=1) / y.dimshuffle(0,  'x'), axis=0)
```
@nouiz , what do you think?

mohammadpz:
@carriepl , whenever we decide on current optimization, I'll write tests.

theano/tensor/opt.py
```diff
@@ -4129,12 +4143,27 @@ def local_sum_div_dimshuffle(node):
...
+                            numerator, axis=compatible_dims)
+                        div_op = T.true_div(
+                            op_on_compatible_dims,
+                            (optimized_dimshuffle **
+                                T.prod([numerator.shape[ax]
+                                        for ax in compatible_dims])))
+                        op_on_incompatible_dims = T.prod(
+                            div_op,
+                            axis=reordered_incompatible_dims)
+                    return [op_on_incompatible_dims]
```
mohammadpz: @carriepl , Let me explain what I'm doing here by an example:
Let's say we have this:
```python
x = T.dtensor3() 
y = T.dvector()
z = T.prod(x / y.dimshuffle('x', 0, 'x'), axis=(0, 1, 2))
```
There are 3 different things here:
1. ```compatible_dims``` which is ```[0, 2]```
2. ```in_compatible_dims``` which is ```[1]```
3. ```optimized_dimshuffle_order``` which is remaining dimensions in ```dimshuffle``` which has nothing to do with ```prod```. In this case it is ```('x', 0)```
Strategy is to:
1. apply ```prod``` on ```compatible_dims```.
2. divide result by new ```dimshuffle```. (since new order is ```('x', 0)``` it is equal to not applying ```dimshuffle``` so we don't)
3. apply outer ```prod``` on ```in_compatible_dims```. (Note that ```in_compatible_dims``` will change from ```[1]``` to ```[0]```) because of the first ```prod```.

Finally the optimized graph is:
```python
z = T.prod(T.prod(x, axis=(0, 2)) / y ** T.prod(list_of_shapes_of_compatible_dims), axis=0)
```

Here's the ```debugprint``` results before and after optimizations. Can you confirm that?
Before:
```python
Prod{axis=[0, 1, 2], acc_dtype=float64} [@A] ''   2
 |Elemwise{true_div,no_inplace} [@B] ''   1
   |<TensorType(float64, 3D)> [@C]
   |DimShuffle{x,0,x} [@D] ''   0
     |<TensorType(float64, vector)> [@E]
```
After:
```python
Prod{acc_dtype=float64} [@A] ''   7
 |Elemwise{Composite{(i0 / (i1 ** i2))}}[(0, 0)] [@B] ''   6
   |Prod{axis=[0, 2], acc_dtype=float64} [@C] ''   2
   | |<TensorType(float64, 3D)> [@D]
   |<TensorType(float64, vector)> [@E]
   |InplaceDimShuffle{x} [@F] ''   5
     |Prod{acc_dtype=int64} [@G] ''   4
       |MakeVector [@H] ''   3
         |Shape_i{0} [@I] ''   1
         | |<TensorType(float64, 3D)> [@D]
         |Shape_i{2} [@J] ''   0
           |<TensorType(float64, 3D)> [@D]
```

mohammadpz: I added numerical tests and apparently it's working fine, but I don't know how to check if optimization has been applied or not. Any idea? @carriepl

theano/tensor/opt.py
```diff
@@ -63,54 +63,6 @@
...
-                v.tag.trace = tr
-    else:
-        # Copy over stack traces from from_var to each variable to
-        # to_var, including the stack_trace of the to_var before
-        if hasattr(to_var.tag, 'trace'):
-            to_var.tag.trace = to_var.tag.trace + tr
-        else:
-            to_var.tag.trace = tr
-
-
```
mohammadpz: @carriepl , I don't know why these changes has been added. Maybe I have pull something wrong?

carriepl: You've merged the central/master of Theano into your branch. You shouldn't do this; if there is a conflict between your branch and the master, you should do a rebase of your branch over the master.

carriepl:
Also, some tests are now failing in Travis.

theano/tensor/opt.py
mohammadpz: Now how should I fix this?

mohammadpz:
@carriepl , The error is related to ```test_local_sum_prod_mul_by_scalar``` which it is changed because of my mistake on pull from upstream.

theano/tensor/opt.py
carriepl: Do an interactive rebase to remove the commit where you merge the master into your branch, then rebase your branch on top of the master. If you need help, you can come see me IRL.

carriepl:
What's the status on this?

mohammadpz:
Sorry for the very long delay. I'll fix it tonight.

nouiz:
Just to note that I don't see a fix and that this PR need a rebase.

lamblin:
Continued in #3259, which was merged.

PR #3150, opened 2015-07-16
added the shared variable zero fn.

caglar:
I was using this function for a while for myself. It is very similar to what torch.tensor:zero function does:
https://github.com/torch/torch7/blob/master/doc/tensor.md#self-zero

This is just a syntactic sugar. I find this more readable than doing get_value and then set_value to reset a shared variable to zero in my code. 

If that PR gets through, I have a few other functions like that which I can make PR for too.

### DIFF
--- a/theano/compile/sharedvalue.py
+++ b/theano/compile/sharedvalue.py
@@ -113,6 +113,22 @@ class SharedVariable(Variable):
         else:
             self.container.value = copy.deepcopy(new_value)
 
+    def zero(self, borrow=False):
+        """Set the values of a shared variable to 0.
+
+        :param borrow:
+            True to modify the value of a shared variable directly by using
+            its previous value. Potentially this can cause problems
+            regarding to the aliased memory.
+
+        Changes done with this function will be visible to all functions using
+        this SharedVariable.
+        """
+        if borrow:
+            self.container.value = 0 * self.container.value
+        else:
+            self.container.value = 0 * copy.deepcopy(self.container.value)
+
     def clone(self):
         cp = self.__class__(
             name=self.name,
--- a/theano/compile/sharedvalue.py
+++ b/theano/compile/sharedvalue.py
@@ -125,9 +125,9 @@ class SharedVariable(Variable):
         this SharedVariable.
         """
         if borrow:
-            self.container.value = 0 * self.container.value
+            self.container.value[...] = 0
         else:
-            self.container.value = 0 * copy.deepcopy(self.container.value)
+            self.container.value = 0 * self.container.value
 
     def clone(self):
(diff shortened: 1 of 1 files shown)


theano/compile/sharedvalue.py
```diff
@@ -113,6 +113,22 @@ def set_value(self, new_value, borrow=False):
...
+        :param borrow:
+            True to modify the value of a shared variable directly by using
+            its previous value. Potentially this can cause problems
+            regarding to the aliased memory.
+
+        Changes done with this function will be visible to all functions using
+        this SharedVariable.
+        """
+        if borrow:
+            self.container.value = 0 * self.container.value
```
nouiz: that do not borrow the value. to borrow the value, do something like:

self.container.value[...] = 0

I tested it and it work with cpu, and both gpu back-end.

theano/compile/sharedvalue.py
```diff
@@ -113,6 +113,22 @@ def set_value(self, new_value, borrow=False):
...
+        else:
+            self.container.value = 0 * copy.deepcopy(self.container.value)
```
nouiz: the deep copy is useless, the elemwise will always create new memory

nouiz:
What are the other one you want? I think it could be good to add support for [...] and maybe fill.

caglar:
I have one function for fill and the to get the shapes. But what do you mean by [...]?

theano/compile/sharedvalue.py
caglar: Ok thanks. Commited both of your suggestions.

nouiz:
I mean, allowing a_shared_var[...] = X do the same as a_numpy_ndarray[...] = X.

Mostly support on shared var the syntex you used in zeros.

Can you add the others functions in this PR? This would also need tests to make sure we don't break it later.

caglar:
OK will do that.

lamblin:
Travis failures are unrelated, and come from the master at the time it was merged.
They should go away when you update the PR.

nouiz:
I'll merge now as this add a good feature. thanks. I made gh-3218 for the follow up.

caglar:
Hmmm, I was planning to put some tests for it. I think it is ok. But just
for your information, I have only used this on CPU, and it was working.
However I have never tested it on GPU.

On Tue, Jul 28, 2015 at 1:36 PM, Frédéric Bastien <notifications@github.com>
wrote:

> Merged #3150 <https://github.com/Theano/Theano/pull/3150>.
>
> —
> <https://github.com/Theano/Theano/pull/3150#event-367492050>.
>



-- 
Caglar GULCEHRE

PR #3151, opened 2015-07-16
[BUG] Fix invalid optimization for product of multiplication

carriepl:
Optimization local_sum_prod_mul_by_scalar had behavior : 
prod(scalar * smth) -> scalar * prod(smth)
prod(-smth) -> -prod(smth)

when it should have been : 
prod(scalar * smth) -> scalar ** size(smth) * prod(smth)
prod(-smth) -> (-1) ** size(smth) * prod(smth)

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3907,7 +3907,7 @@ def local_sum_prod_mul_by_scalar(node):
             # raised to the power of the number of elements in the input
             # to the Prod
             if (isinstance(node.op, T.elemwise.Prod) and
-                new_op_input_nb_elements != 1):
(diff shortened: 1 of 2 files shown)


lamblin:
Travis seems to be overwhelmed.
Can you run the relevant tests with DebugMode locally? If that works, I can merge.

carriepl:
I ran them in DebugMode, both float32 and float64. Also, I had travis run the two jobs that it had skipped and they pass.

carriepl:
Done

nouiz:
we can merge when travis pass. It will be long, it is overloaded.

PR #3152, opened 2015-07-16
Don't push out the outputs of an 'as_while' Scan

carriepl:
this is preemtive fix. It could cause an crash or bad value. as_while isn't much used.

### DIFF
--- a/theano/scan_module/scan_opt.py
+++ b/theano/scan_module/scan_opt.py
@@ -612,8 +612,10 @@ class PushOutScanOutput(gof.Optimizer):
         fgraph.attach_feature(gof.toolbox.ReplaceValidate())
 
     def apply(self, fgraph):
-        # Don't perform the optimization on as_while scans since it's
-        # impossible to know for how many steps it will run.
+        # Don't perform the optimization on as_while scans. Because these scans
+        # don't run for a predetermined number of steps, handling them is
+        # more complicated and this optimization doesn't support it at the
+        # moment.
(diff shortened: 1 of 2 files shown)


theano/scan_module/scan_opt.py
```diff
@@ -612,8 +612,11 @@ def add_requirements(self, fgraph):
         fgraph.attach_feature(gof.toolbox.ReplaceValidate())
 
     def apply(self, fgraph):
+        # Don't perform the optimization on as_while scans since it's
+        # impossible to know for how many steps it will run.
         nodelist = [x for x in fgraph.toposort()
-                    if isinstance(x.op, scan_op.Scan)]
+                    if (isinstance(x.op, scan_op.Scan) and
+                        not x.op.as_while]
```
nouiz: This need a test. Seem ok for now, but I don't agree we can never push output of scan in while mode. After the scan, we know how many iteration was done. But we shouldn't push outputs used for by the condition. But we can let this for later.

nouiz: So just update the comment for this if you agree.

nouiz:
I think this is good, but @abergeron can you also review it?

lamblin:
OK for me as well.

carriepl:
Fixed an issue with running the new test in floatX=float32

nouiz:
as travis is overloaded, I killed it on the last commit. I'm sure it is good. I merge now.

PR #3155, opened 2015-07-16
Props gof tests

harlouci:
Some more __props__

### DIFF
--- a/theano/gof/tests/test_destroyhandler.py
+++ b/theano/gof/tests/test_destroyhandler.py
@@ -43,6 +43,10 @@ def MyConstant(data):
 
 class MyOp(Op):
 
+    __props__ = ("nin", "name", "vmap", "dmap", "nout",
+                 "destroyhandler_tolerate_same",
+                 "destroyhandler_tolerate_aliased")
+                 
     def __init__(self, nin, name, vmap=None, dmap=None, nout=1,
                  destroyhandler_tolerate_same=None,
                  destroyhandler_tolerate_aliased=None):
--- a/theano/gof/tests/test_link.py
+++ b/theano/gof/tests/test_link.py
@@ -31,6 +31,9 @@ def double(name):
 
 
 class MyOp(Op):
+
+    __props__ = ("nin", "name", "impl")
+    
     def __init__(self, nin, name, impl=None):
         self.nin = nin
         self.name = name
--- a/theano/gof/tests/test_op.py
+++ b/theano/gof/tests/test_op.py
@@ -56,6 +56,8 @@ class MyType(Type):
 
 class MyOp(Op):
 
+    __props__ = ()
+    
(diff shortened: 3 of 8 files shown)


theano/gof/tests/test_destroyhandler.py
```diff
@@ -43,6 +43,10 @@ def MyConstant(data):
 
 class MyOp(Op):
 
+    __props__ = ("nin", "name", "vmap", "dmap", "nout",
+                 "destroyhandler_tolerate_same",
+                 "destroyhandler_tolerate_aliased")
```
lamblin: For that one, given that lists and dictionaries are not hashable, and that no `__eq__` and `__hash__` were defined in the first place, it may be better not to define `__props__` at all.

theano/gof/tests/test_link.py
```diff
@@ -31,6 +31,9 @@ def double(name):
 
 
 class MyOp(Op):
+
+    __props__ = ("nin", "name", "impl")
+
```
lamblin: Remove the implementation of `__str__(self)` as well?

theano/gof/tests/test_vm.py
```diff
@@ -331,6 +331,9 @@ def time_linker(name, linker):
 
 
 class RunOnce(theano.Op):
+
+    __props__ = ()
+
```
lamblin: It does not really make a difference here, but I would still include `'nb_run'` in `__props__`.

harlouci: Is "return self.name" 
the same as "return self.__class__.__name__" (in test_link.py) ?

theano/gof/tests/test_destroyhandler.py
```diff
@@ -42,7 +42,7 @@ def MyConstant(data):
 
 
 class MyOp(Op):
-
+
```
lamblin: This whitespace should not be added.

theano/gof/tests/test_op.py
```diff
@@ -56,6 +56,8 @@ def may_share_memory(a, b):
 
 class MyOp(Op):
 
+    __props__ = ()
+
```
lamblin: test_op.py:60:1: W293 blank line contains whitespace

theano/gof/tests/test_vm.py
```diff
@@ -331,6 +331,9 @@ def time_linker(name, linker):
 
 
 class RunOnce(theano.Op):
+
+    __props__ = ("nb_run",)
+
```
lamblin: test_vm.py:336:1: W293 blank line contains whitespace

theano/gof/tests/test_vm.py
nouiz: no. self.name is most of the time a name given by the user.

PR #3157, opened 2015-07-16
Props tests

harlouci:
props

### DIFF
--- a/theano/tests/test_tutorial.py
+++ b/theano/tests/test_tutorial.py
@@ -139,15 +139,12 @@ class T_extending(unittest.TestCase):
 
         from theano import gof
         class BinaryDoubleOp(gof.Op):
+        
+            __props__ = ("name", "fn")
+            
             def __init__(self, name, fn):
                 self.name = name
                 self.fn = fn
-
-            def __eq__(self, other):
-                return type(self) == type(other) and (self.name == other.name) and (self.fn == other.fn)
-
-            def __hash__(self):
-                return hash(type(self)) ^ hash(self.name) ^ hash(self.fn)
 
             def make_node(self, x, y):
                 if isinstance(x, (int, float)):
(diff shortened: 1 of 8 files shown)


theano/tests/test_tutorial.py
```diff
@@ -139,16 +139,13 @@ def make_node(x, y):
 
         from theano import gof
         class BinaryDoubleOp(gof.Op):
+        
+            __props__ = ("name", "fn")
+
```
lamblin: These tests are supposed to be in sync with the tutorials (check the comment at the beginning of the class).
So if you modify them here, you should also update the tutorials in the documentation at the same time.

harlouci: Right. I'll remove __props__ then.

harlouci: And put back the methods.

harlouci: Does your suggestion apply to all the classes below ?

lamblin: I think it would be better to update the tutorials instead. The files are in `doc/extending/*.txt`.
This applies to this whole file.

harlouci: I didn't find the corresponding doc to  :         
- class PyCUDADoubleOp(theano.Op):  
- one out of two  class BinaryDoubleOp(gof.Op):

lamblin: - https://github.com/Theano/Theano/blob/master/doc/tutorial/using_gpu.txt#L683
- https://github.com/Theano/Theano/blob/master/doc/tutorial/gpu_data_convert.txt#L81
- https://github.com/Theano/Theano/blob/master/doc/extending/cop.txt#L302
- https://github.com/Theano/Theano/blob/master/doc/extending/op.txt#L676

PR #3158, opened 2015-07-16
[TESTS] this fix a test while having concurrent PR that force pep8.

nouiz:
This should fix travis that is failing on master:

https://travis-ci.org/Theano/Theano/builds/71255496


abergeron:
It's only trailing whitespace so I'll merge it right away.

PR #3159, opened 2015-07-16
Remove printing during test

nouiz:


theano/configdefaults.py
```diff
@@ -246,7 +246,7 @@ def __str__(self):
     'on_opt_error',
     ("What to do when an optimization crashes: warn and skip it, raise "
      "the exception, or fall into the pdb debugger."),
-    EnumStr('warn', 'raise', 'pdb'),
+    EnumStr('warn', 'raise', 'pdb', 'ignore'),
```
lamblin: Good idea. Can you update `doc/library/config.txt` as well?

lamblin:
Apart from that, merge when travis passes.

carriepl:
I'm fine with the modifications since @lamblin's review. No need to wait for the last few processes in travis because the last 3 commits only change the documentation.

Issue #1471, opened 2013-07-29
Convert old GPU ops to GpuArray ops


nouiz:
This is a fast conversion. We do not support OpenCL to don't maintain 2 code base for a long time. OpenCL is for later. We need help from the community for this.

How to convert.
0) see http://www.deeplearning.net/software/theano/tutorial/extending_theano.html to get an idea of how to make an op.
1) Install compyte (a new dependency for Theano). Use the version in abergeron account in the reorg branch: http://deeplearning.net/software/compyte/installation.html
1) Fast way(but no OpenCL support), copy the old Op from the CudaNdarray back-end under theano/sandbox/cuda/*.py and its tests to theano/sandbox/gpuarray[/tests]. See gh-1590 and gh-1634 for examples.
   Then modify it, 
      - by adding the c_compilers, c_headers and if needed c_support_code, c_init_code methods.
      - modify the make node to return an GpuArrayVariabe with GpuArrayType and make sure to call as_gpu_array_varible(input) on all input variable.
      - Modify the CudaNdarray macro the the compatible version: https://github.com/abergeron/compyte/blob/reorg/src/compyte/numpy_compat.h
      - The macro CudaNdarray_DEV_DATA is replaced by the fct call cuda_get_ptr() defined in c_support_code, the offset need to be added: (float *)(((char *)cuda_get_ptr(%(inp)s->ga.data))+%(inp)s->ga.offset)
      - Don't forget that when changing the macro CudaNdarray_HOST_STRIDES to PyGpuArray_STRIDES(), the unit change from number of element to number of bytes. So you need to divide it by 4 if you support only float32. Otherwise, you need to update the kernel code.
      - change NUM_VECTOR_OP_THREADS_PER_BLOCK to 256
      - change CNDA_THREAD_SYNC to """        if config.gpuarray.sync:
            code += "GpuArray_sync(&%(zz)s->ga);\n" % dict(zz=zz)"""
      - Make sure all tests pass. Commit.
      - Make a new optimizer function by reusing the new op_lifter.
   Add support for more dtype then float32 and modify tests for this.
1.1) Optional. Make the new op inherit from the CPU op to share code.
1.2) Optional. Make the CPU tests generic and merge it in the gpu tests. Reuse for the GPU tests.

note::
- It can happen that after nosetests results are printed that it have a segfault or another errors. This is due to the order where we close stuff. Just ignore it for now, we will take care of it. This is not related to Ops code.

Absolutely needed ops:
[...]

PR #3163, opened 2015-07-17
Performance improvement for GpuCrossentropySoftmaxArgmax1HotWithBias

mjwillson:
After a bit of profiling this op turned out to be a massive bottleneck in training a smaller feedforward network with a large-ish (10000 class) softmax output layer, taking 40 to 60% of the total time, much more than any of the dot22 / gemm linear algebra ops which really should be the bottleneck.

On investigation it turned out the GPU kernel for this op was only using a single thread, with a TODO to fix it. This commit implements that TODO. This op now only takes around 6% of the total time for me. It could probably be optimised further (e.g. via hierarchical / multi-stage reductions) but I think this is worth submitting.

Tagging @jaberg as it looks like this was your code originally. It's my first theano contribution and I'm relatively new to CUDA so let me know if you notice anything wrong. The relevant tests still pass at least.

Incidentally the CPU implementation of this op seems suspiciously slow too, although I didn't focus on optimising that because I'm mainly using GPU.

### DIFF
--- a/theano/sandbox/cuda/nnet.py
+++ b/theano/sandbox/cuda/nnet.py
@@ -36,61 +36,124 @@ class GpuCrossentropySoftmaxArgmax1HotWithBias(GpuOp):
         am = y_idx.type()
         return Apply(self, [x, b, y_idx], [nll, sm, am])
 
+    def c_headers(self):
+        return ['<float.h>']
+
     def c_support_code(self):
         return """
-        __global__ void k_xent_sm_1hot_bias(int M, int N,
-            const float * x_data, int xs0, int xs1,
-            const float * b, int bs0,
-            const float * y_idx_data, int y_idxs0,
-            float * nll_data, int nlls0,
-            float * sm_data, int sms0, int sms1,
-            float * am_data, int ams0)
-        {
-          for (int row = blockIdx.x; row < M; row += gridDim.x){
+__global__ void k_xent_sm_1hot_bias(const int M, const int N,
+    const float * x_data, const int xs0, const int xs1,
+    const float * b, const int bs0,
+    const float * y_idx_data, const int y_idxs0,
+    float * nll_data, const int nlls0,
+    float * sm_data, const int sms0, const int sms1,
+    float * am_data, const int ams0)
+{
+  for (int row = blockIdx.x; row < M; row += gridDim.x){
 
-            const float * x = x_data + xs0 * row;
-            const int y_idx = (int)y_idx_data[row * y_idxs0];
-            float * sm = sm_data + sms0 * row;
+    const float * x = x_data + xs0 * row;
+    float * sm = sm_data + sms0 * row;
 
-            float sum = 0.0;
-            int row_max_j = 0;
-            float row_max = x[0] + b[0];
-            for (int j = 1; j < N; ++j)
-            {
-                float row_ij = x[j*xs1] + b[j*bs0];
-                //todo: store to shared memory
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/nnet.py
```diff
@@ -38,59 +38,120 @@ def make_node(self, x, b, y_idx):
...
-                nll_data[row*nlls0] = - x[y_idx*xs1]
-                           - b[y_idx*bs0]
-                           + row_max
-                           + log(sum);
-            }
-            am_data[row*ams0] = row_max_j;
-          }
-        }
-
+#include <float.h>
```
nouiz: header must be declared elsewhere, remove that line and add that method:

```
    def c_headers(self):
        return ['<float.h>']


```

theano/sandbox/cuda/nnet.py
```diff
@@ -38,59 +38,120 @@ def make_node(self, x, b, y_idx):
...
+    float * nll_data, const int nlls0,
+    float * sm_data, const int sms0, const int sms1,
+    float * am_data, const int ams0)
+{
+  for (int row = blockIdx.x; row < M; row += gridDim.x){
+
+    const float * x = x_data + xs0 * row;
+    float * sm = sm_data + sms0 * row;
+
+    __shared__ float per_thread_values[NUM_VECTOR_OP_THREADS_PER_BLOCK];
```
nouiz: It is better to have a dynamic shared memory size. If there a reason you didn't used that except that it was easier? As your PR make it faster, we can still merge without change here.

Thanks for the PR, I'll let @abergeron finish the review in more details.

abergeron:
This does seem ok and passes the tests, so I'm all for it.

You would only need to address the comment from @nouiz and also perhaps port this speedup to the new backend in sandbox/gpuarray/nnet.py.  You can just copy and paste your kernel and adapt the calling code in the same way.

theano/sandbox/cuda/nnet.py
mjwillson: Done

theano/sandbox/cuda/nnet.py
mjwillson: Yeah no particular reason, just I hadn't used dynamic shared memory before and it seemed potentially more complicated for a fixed size buffer. Happy to change it though, see the revised commit.

mjwillson:
Great, thanks.
I just pushed a revised commit that addresses your issues @nouiz .

@abergeron Re porting to the gpuarray backend, is this now the recommended back-end to use? I just had a look and it doesn't look like a straight copy-paste, looks like the kernel code needs to be generic to different numpy dtypes. I'll have a go if I get time for it but leaving you with this for now...

nouiz:
The new back-end isn't used much, but we are close. We prefer to keep in sync the part that are in both to not have them diverge too much. For the multi-dtype, this isn't very hard, just use the inputs/outputs dtype in the c code via string substitution do it. If you prefer not doing it, can you move the code to the new back-end and I can take care of the dtype.

thanks

nouiz:
I added a note in https://github.com/Theano/Theano/issues/1471 to port this speed up to the new back-end so that we don't forget that.

nouiz:
By the way, thanks for the great speed up. It is always great when people update their version and push the change upstream.

thanks

mjwillson:
Thanks, no problem! Sorry I didn't get around to porting it to the other
back-end.

PR #3164, opened 2015-07-17
Fix crash, the amdlibm hack to remove it for nvcc change the hash./me…

nouiz:
…mtestCL --gpu 0 4000 10000 This also make use keep the lock for less time.


nouiz:
after discussion with @carriepl we agree to merge now. This should not wait for travis as it fix crash and travis won't detect that change.

The crash fix isn't evident. I'll make another PR that clean up the code and would cause a crash in all case if this problem happen again.

PR #3165, opened 2015-07-17
Fix broadcastable pattern of flatten output

carriepl:
This was the cause of an error in one of Scan's optimization that was reported on the theano-dev mailing list : https://groups.google.com/forum/#!topic/theano-users/PCg5KjskuN4


nouiz:
it seem good. We can merge when travis pass. flatten don't exist on the new gpu back-end. In the old gpu back-end, it reuse the CPU flatten broadcastable, so it should be good.

PR #3167, opened 2015-07-17
Rework amdlibm exclusion on GPU.

abergeron:


nouiz:
All seem good. Can you doc the change in the file corresponding to:

http://deeplearning.net/software/theano/extending/cop.html

lamblin:
There are still places calling `c_compile_args()` without passing the compiler.

abergeron:
Should be fixed

nouiz:

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/gof/cc.py:1345:9: F841 local variable 'compiler_name' is assigned to but never used

abergeron:
Now there should be no problems left.  I hope.

doc/extending/cop.txt
```diff
@@ -57,39 +57,63 @@ There are less methods to define for an Op than for a Type:
...
+    .. method:: c_libraries([c_compiler])
 
        Returns a list of library names that your op needs to link to.
        All ops are automatically linked with 'python' and the
        libraries their types require. (arguments to -l)
 
-    .. method:: c_lib_dirs()
+       The `c_compiler` [#2v]_ parameter is the C compiler that will
+       be used to compile the code for the node. You may get multiple
+       calls with different C compilers.
```
nouiz: why there can be multiple call with different compiler? I don't care about the multiple call, but I find it strange the different compiler.

abergeron: Because this can get called with either NVCC_compiler or GCC_compiler (for now).  Since it doesn't get a node parameter, then there is no way to distinguish between multiple calls for different nodes.

nouiz: If any of the inputs or output is a gpu variable, it should be called by nvcc, otherwise gcc. Why anything else could happen?

nouiz:
This PR need a rebase.

doc/extending/cop.txt
abergeron: This method is not apply-node specific, so it can be called for multiple different nodes.

abergeron:
Rebased.

doc/extending/cop.txt
nouiz: Can you explain that in the 2v section you link to?

nouiz:
I ran the tests. With the extra doc, we can merge.

doc/extending/cop.txt
abergeron: It makes a reference to the `[#2v]` "footnote" defined below.  This is because I didn't want to repeat myself in a lot of places so I move the common explanation there.

abergeron: Ok, I'll do that.

abergeron:
Ok, done.

Issue #3024, opened 2015-06-12
Implement missing cases for the grad of sort()


lamblin:
Currently, only a couple of cases implement the gradient, in terms of:
- axes
- number of input dimensions.
We should implement (and test) the missing cases.

lamblin:
Closed by #3168.

PR #3168, opened 2015-07-17
changing sort.grad to work with other ndim

SinaHonari:
fixes issue #3024

### DIFF
--- a/theano/tensor/sort.py
+++ b/theano/tensor/sort.py
@@ -62,13 +62,7 @@ class SortOp(theano.Op):
             self, 0, axis,
             "Currently, we only implement the gradient on sort for vector"
             " matrix (and axis is None or 0) and tensor3")
-        if a.ndim == 1:
-            idx = argsort(*inputs, kind=self.kind, order=self.order)
-            # rev_idx = numpy.where(idx[None, :]==numpy.arange(5)[:,None])[1]
-            rev_idx = theano.tensor.eq(idx[None, :],
-                                       arange(a.shape[0])[:, None]).nonzero()[1]
-            inp_grad = output_grads[0][rev_idx]
-        elif isinstance(axis, theano.Constant):
+        if isinstance(axis, theano.Constant):
             if isinstance(axis, theano.Constant) and axis.data is not None:
                 indices = self.__get_argsort_indices(a, axis)
                 inp_grad = output_grads[0][tuple(indices)]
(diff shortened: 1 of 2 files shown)


nouiz:
the pr need a rebase.

nouiz:
travis still have this error:
```
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tensor/sort.py:68:52: W503 line break before binary operator
```

theano/tensor/sort.py
```diff
@@ -62,39 +62,31 @@ def grad(self, inputs, output_grads):
             self, 0, axis,
             "Currently, we only implement the gradient on sort for vector"
             " matrix (and axis is None or 0) and tensor3")
```
lamblin: Now, I think you implemented all supported cases. So I would get rid of that default "not implemented" message, and maybe put it in an "else" block below, changing the message to say that this is not a supported case.

theano/tensor/sort.py
```diff
@@ -62,39 +62,31 @@ def grad(self, inputs, output_grads):
...
-                    (isinstance(axis, theano.Constant) and axis.data is None)):
-                rev_idx = self.__get_argsort_indices(a, axis)
-                inp_grad = output_grads[0][rev_idx].reshape(a.shape)
+        if ((isinstance(axis, theano.Constant) or (isinstance(axis,
+                                                   theano.tensor.
+                                                   TensorVariable) and
+                                                   axis.ndim == 0)) and
+                axis.data is not None):
+            indices = self.__get_argsort_indices(a, axis)
+            inp_grad = output_grads[0][tuple(indices)]
```
lamblin: Do you think you could call `tuple(...)` before returning the value from `__get_argsort_indices` when axis is not None? That way, these two lines could be combined with the `elif` block.

lamblin:
Apart from that, which may need a bit of refactoring, that looks great, thanks a lot!
So we can keep the symbolic axis, since everything is working with it.

lamblin:
Can you fix the pep8 issue? Then I think it is ready to merge.

SinaHonari:
how to check that specific test locally before committing PR?

nouiz:
nosetests theano/tests/test_flake8.py

But it don't cover exactly the same as in travis, not sure why...

lamblin:
That failure was in the master at the time, so I'll merge.

PR #3176, opened 2015-07-20
Remove the restriction on zero target dimensions when reshaping a CudaNdarray

### LINKED ISSUE #3175: GpuReshape breaks on zero target dimension
I'm working with variable-length input which is sometimes empty. Currently these empty tensors seem to break reshape operations, when using the cuda backend:

`Reshape has invalid dimension 0 (must be >0)`

No such restriction exists when using the CPU backend, and I can't see any obvious reason why a zero dimension couldn't be allowed?

Relevant code is here:
https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/cuda_ndarray.cu#L875
Example error:

ValueError: Reshape has invalid dimension 0 (must be >0)
Apply node that caused the error: GpuReshape{3}(GpuAdvancedSubtensor1.0, TensorConstant{[  0   2 128]})
Toposort index: 1
Inputs types: [CudaNdarrayType(float32, matrix), TensorType(int64, vector)]
Inputs shapes: [(0, 128), (3,)]
Inputs strides: [(128, 1), (8,)]
Inputs values: [<CudaNdarray object at 0x7fe68b3728f0>, array([  0,   2, 128])]
Outputs clients: [[HostFromGpu(GpuReshape{3}.0)]]

mjwillson:
Fixes #3175 for me.

I couldn't see any obvious problems with removing this restriction (which isn't applied by the CPU backend). But perhaps there was some cuda-specific reason for the stricter restriction here?

There was a commit (https://github.com/Theano/Theano/commit/23fa09f02d731ca0a5437772cf6118ca16c542c9) made to permit reshaping of zero-size arrays before, which looks like it fixed this issue but was partially undone again at some point in a merge.

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -871,8 +871,8 @@ PyObject * CudaNdarray_Reshape(CudaNdarray * self, PyObject * shape)
                 free(rval_dims);
                 return NULL;
             }
-            if(rval_dims[i]<=0){
-                PyErr_Format(PyExc_ValueError, "Reshape has invalid dimension %i (must be >0)",rval_dims[i]);
+            if(rval_dims[i]<0){
+                PyErr_Format(PyExc_ValueError, "Reshape has invalid dimension %i (must be >=0)",rval_dims[i]);
                 free(rval_dims);
                 return NULL;
             }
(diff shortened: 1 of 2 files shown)


nouiz:
Thanks for the PR. Can you add a tests in sandbox/cuda/test_basic_ops.py so that we don't break it again?

mjwillson:
Sure, just pushed.

nouiz:
thanks

PR #3178, opened 2015-07-20
props to theano/misc/pycuda_example.py

harlouci:
props.

### DIFF
--- a/theano/misc/pycuda_example.py
+++ b/theano/misc/pycuda_example.py
@@ -279,7 +279,7 @@ class PycudaElemwiseSourceModuleOp(GpuOp):
 class PycudaElemwiseSourceModuleMakeThunkOp(Op):
     nin = property(lambda self: self.scalar_op.nin)
     nout = property(lambda self: self.scalar_op.nout)
-    __props__ = ("scalar_op", "inplace_pattern", "name")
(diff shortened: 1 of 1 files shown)


theano/misc/pycuda_example.py
```diff
@@ -279,6 +279,7 @@ def perform(self, node, inputs, out):
 class PycudaElemwiseSourceModuleMakeThunkOp(Op):
     nin = property(lambda self: self.scalar_op.nin)
     nout = property(lambda self: self.scalar_op.nout)
+    __props__ = ("scalar_op", "inplace_pattern", "name")
```
nouiz: remove name from the props.

nouiz:
i canceled travis on this one.

nouiz:
@harlouci can you finish this one?

harlouci:
I removed 'name' from the props.

There is a problem in 'theano.misc.pycuda_init.py' with the value of variable 'pycuda_available' that set to False raises en exception at line 38 in 'theano/misc/pycuda_example.py' when I run theano-nose.

nouiz:
you didn't push the new version. It is normal that you can't run those example, as I think you don't have pycuda installed.

harlouci:
Sorry, it's seems indeed that I didn't push my last commit... Here it is.

PR #3183, opened 2015-07-21
Cpu Contiguous Dim Check

kelvinxu:
Adds a dimension check pull request at #2949

### DIFF
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -40,6 +40,7 @@ class CpuContiguous(theano.Op):
             if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
                 // check to see if output is contiguous first
                 if (%(y)s != NULL &&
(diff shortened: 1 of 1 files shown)


theano/tensor/extra_ops.py
```diff
@@ -40,6 +40,7 @@ def c_code(self, node, name, inames, onames, sub):
             if (!PyArray_CHKFLAGS(%(x)s, NPY_ARRAY_C_CONTIGUOUS)){
                 // check to see if output is contiguous first
                 if (%(y)s != NULL &&
+                    PyArray_CompareLists(PyArray_DIMS(%(y)s), PyArray_DIMS(%(x)s), PyArray_NDIMS(%(x)s)) &&
```
nouiz: You need to bump c_code_cache_version.

nouiz:

lamblin:
This is failing with an error saying: `'PyArray_NDIMS' was not declared in this scope`.
I did not check why that would happen yet.

nouiz:
I think it should be PyArray_NDIMS

nouiz:
I think it should be PyArray_NDIM

kelvinxu:
Amended

PR #3184, opened 2015-07-21
props to theano/gof/op.py

harlouci:
props

### DIFF
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -956,6 +956,7 @@ class OpenMPOp(Op):
     """
     True/False after we tested this.
     """
+    __props__ = ()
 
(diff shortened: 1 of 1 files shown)


nouiz:
those 2 class are parent class. We should not do that, as if the chield forget to overrided them, this can hide bug.

Also, as now you don't have much change, don't make 1 PR per directory, but wait to have at least 5 files changed in one PR. So a PR could be multiple directory. I'll close this PR as we won't take any of it.

harlouci:
What do I have to do with this branch : make a pull request again ?

nouiz:
Do nothing about this PR. DO not change those classes.

PR #3186, opened 2015-07-21
Mixed

nouiz:
a few mixed stuff

### DIFF
--- a/doc/tutorial/examples.txt
+++ b/doc/tutorial/examples.txt
@@ -476,7 +476,7 @@ work on the CPU and GPU. CURAND only work on the GPU.
 
         .. code-block:: python
 
-            from theano.sandbox.rng_mrg import MRG_RandomStreams
+            from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
 
 .. _logistic_regression:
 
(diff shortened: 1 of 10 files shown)


doc/tutorial/examples.txt
```diff
@@ -465,11 +465,19 @@ There are :ref:`other distributions implemented <libdoc_tensor_raw_random>`.
...
+ The RandomStream only work on the CPU, MRG31k3p
 work on the CPU and GPU. CURAND only work on the GPU.
 
+.. note::
+
+    To use you the MRG version easily, you can just change the import to:
+
+        .. code-block:: python
+
+            from theano.sandbox.rng_mrg import MRG_RandomStreams
```
lamblin: Why not `... as RandomStreams`?

theano/sandbox/cuda/opt_util.py
```diff
@@ -73,8 +73,21 @@ def opt(node):
                 if lr is None or targ is None:
                     return None
                 inputs = list(targ.inputs)
-                inputs[alpha_in] = lr * targ.inputs[alpha_in]
-                inputs[beta_in] = lr * targ.inputs[beta_in]
+                try:
+                    raise NotScalarConstantError()
```
lamblin: Is that supposed to be the final version? The code below will always be skipped.

nouiz:
i fixed both comment.

lamblin:
Thanks!
I'm OK to merge when Travis passes.

PR #3189, opened 2015-07-21
[BUG, TEST] Fix Sng handling of nan and fix related test

nouiz:
this fix 2 tests in the buildbot in debugmode

theano-nose theano/tensor/tests/test_basic.py:test_comparison.test_{isclose,allclose}


abergeron:
Does this work on GPU?  I know cuda has isnan() but I don't think it has the NAN define.

nouiz:
I added a commit to test that. This shown weakness in the new back-end in comparison with nan. I fixed them at the same time.

abergeron:
Ok with @lamblin 's comment.

nouiz:
done

Issue #3079, opened 2015-06-26
Memory leak in theano.scan


the-moliver:
I found that theano.scan leaked memory as used by keras:
https://github.com/fchollet/keras/pull/283

I fixed the implementation to use theano.dot, but the issue with scan should be fixed.

nouiz:
Can you provide us a script that reproduce this? It would help us investigate this. Also, do we just need to install keras with the version before your fix or there is other dependency?

I'm surprised that this cause a memory leak, tell us how you see the memory? Could this be that it just use more memory instead of leaking memory?

I think your fix should speed up compilation and maybe have the same final graph after optimization then without it. But maybe not and if that is the case, this would be a missing optimization in Theano.

nouiz:
@carriepl can you check the pr in keras and tell me if you think we most the dot out of the scan in that case?

watterm:
Hi,
I also noticed memory leakage when using scan, in my case it happened with batched_tensordot. Here's some code for reproduction:

    import numpy as np
    import theano.tensor as T
    from theano import function, config, scan, shared

    import gc
    import resource

    A = T.matrix()
    B = T.ones((1000, 2, 4))

    # Leaking:
    #AB = T.batched_tensordot(A, B, axes=[[1], [1]])
    fn = lambda xm, ym: T.tensordot(xm, ym, [[0], [0]])

    # Not Leaking:
    #AB = T.batched_dot(A, B)
    #fn = lambda xm, ym: T.dot(xm, ym)

    AB, update = scan(fn=fn,
                      outputs_info=None,
                      sequences=[A, B],
                      non_sequences=None)

    f = function([A], AB, on_unused_input='warn')

    a = np.random.rand(1000, 2).astype(config.floatX)
    for i in range(int(1e4)):
        gc.collect()
        zn = f(a)
        print(i, resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024.0)

This only happens, when I use the batched_tensordot (= tensordot in scan) on the GPU. It works fine on CPU and batched_dot (= dot in scan) is also correct. I'm not exactly sure what happens behind the scenes in scan, but I assume the same think happened for the keras bug.

It might be that this is somewhat related to the OrderedDict problem in [this Lasagne issue](https://github.com/Lasagne/Lasagne/issues/310). If I execute the code above with python3, everything's fine as well, so maybe the OrderedDicts in scan are not freed correctly. Hope this helps to track the problem down further.

[...]

PR #3193, opened 2015-07-22
Free new_shape before exiting in gpu_reshape

carriepl:
This fixes a memory leak discussed in #3079. 

fix gh-3079


nouiz:
I tested it locally and reshape tests on the GPU pass, so I merge now.

PR #3198, opened 2015-07-23
Add CNMeM in Theano to speed up CUDA allocation.

nouiz:
I think it is ok. @abergeron can you review the integration with Theano?

Also, when you have the time, after this PR is merged, you should check how to allow it do keep the hide_symbols to False.


doc/acknowledgement.txt
```diff
@@ -24,3 +24,4 @@ Acknowledgements
     P. L'Ecuyer and R. Touzin, `Fast Combined Multiple Recursive Generators with Multipliers of the form a = +/- 2^d +/- 2^e <http://www.informs-sim.org/wsc00papers/090.PDF>`_, Proceedings of the 2000 Winter Simulation Conference, Dec. 2000, 683--689.
 
   We were authorized by Pierre L'Ecuyer to copy/modify his Java implementation in the `SSJ <http://www.iro.umontreal.ca/~simardr/ssj/>`_ software and to relicense it under BSD 3-Clauses in Theano.
+* A better GPU memory allocator :attr:`CNMeM <config.lib.cnmem>` is included in Theano. It have the same license.
```
abergeron: It have -> It has

doc/library/config.txt
```diff
@@ -370,6 +370,33 @@ import theano and print the config variable, as in:
     `amdlibm <http://developer.amd.com/cpu/libraries/libm/>`__
     library, which is faster than the standard libm.
 
+.. attribute:: lib.cnmem
+
+    Float value: >= 0
```
abergeron: I'm not sure I like that this is a float value.  I would like it better to have two separate config vars to enable/disable and to set the allocated memory.

But I don't feel too strongly about this.

doc/library/config.txt
```diff
@@ -370,6 +370,33 @@ import theano and print the config variable, as in:
...
+        * 0: not enabled.
+        * 0 < N <= 1: % of the total GPU memory (clipped to .985 for driver memory)
+        * > 0: use that number of MB of memory.
+
+
+    Default 0 (but should change later)
+
+    .. note::
+
+        This could cause memory fragmentation. So if you have an
```
abergeron: if you have an -> if you have a

doc/library/config.txt
```diff
@@ -370,6 +370,33 @@ import theano and print the config variable, as in:
...
+        memory error while using cnmem, try to allocate more memory at
+        the start or without it. If you try this, report your result
```
abergeron: or without it -> or disable it

doc/library/config.txt
```diff
@@ -370,6 +370,33 @@ import theano and print the config variable, as in:
...
+
+    Float value: >= 0
+
+    Do we enable `CNMeM <https://github.com/NVIDIA/cnmem>`_ or not (a
+    faster CUDA memory allocator). In Theano dev version until 0.7.1
+    is released.
+
+    That library is included in Theano, you do not need to install it.
+
+    The value represent the start size (in MB or % of
```
abergeron: The value represent -> The value represents

Also is that only the initial size and it could grow bigger (as implied by the current text) or is that in fact the maximum size (which is allocated right away).

theano/sandbox/cuda/blocksparse.py
```diff
@@ -208,22 +208,28 @@ def c_support_code_apply(self, node, nodename):
...
-            if (cudaMalloc(&%(n)s_W_list, s*sizeof(float *)) != cudaSuccess) return -1;
+            device_free(%(n)s_inp_list);
+            device_free(%(n)s_out_list);
+            device_free(%(n)s_W_list);
+            %(n)s_inp_list = (const float **) device_malloc(s*sizeof(float *));
+            if (%(n)s_inp_list == NULL) return -1;
+            %(n)s_out_list = (float **) device_malloc(s*sizeof(float *));
+            if (%(n)s_out_list == NULL) return -1;
+            %(n)s_W_list = (const float **) device_malloc(s*sizeof(float *));
+            if (%(n)s_W_list == NULL) return -1;
```
abergeron: If we have fast malloc/free we can probably just make this always allocate new memory.

theano/sandbox/cuda/cuda_ndarray.cu
```diff
@@ -3030,18 +3130,22 @@ CudaNdarray_ptr_int_size(PyObject* _unused, PyObject* args)
...
 PyObject *
 CudaNdarray_gpu_init(PyObject* _unused, PyObject* args)
 {
     int card_nb = 0;
     int card_number_provided = 1;
-
-    PyArg_ParseTuple(args, "|i", &card_nb); // if we're given something wildly invalid, this will throw a TypeError
+    float cnmem = 0; // Theano flag lib.cnmem
+    // if we're given something wildly invalid, this will throw a TypeError
+    PyArg_ParseTuple(args, "|if", &card_nb, &cnmem);
```
abergeron: This does raise an exception, but you still have to check the return value.

theano/sandbox/cuda/cnmem.h
```diff
@@ -0,0 +1,259 @@
...
+#include "cuda_runtime_api.h"
+
+#if defined(_MSC_VER) || defined(WIN32)
+#ifdef CNMEM_DLLEXPORT
+#define CNMEM_API __declspec(dllexport)
+#else
+#define CNMEM_API __declspec(dllimport)
+#endif
+#else
+#define CNMEM_API
```
abergeron: For the visibility you should only have to change this line to this:

```C
#ifdef CNMEM_EXPORT
#define CNMEM_API __attribute__((visibility ("default")))
#else
#define CNMEM_API
#endif
```

theano/sandbox/cuda/cnmem.h
```diff
@@ -0,0 +1,259 @@
...
+ * CNMEM_STATUS_INVALID_ARGUMENT, if one of the argument is invalid. For example, used_mem == 0 
+ *                                or free_mem == 0,
+ * CNMEM_STATUS_CUDA_ERROR,       if an error happens in one of the CUDA functions.
+ */
+cnmemStatus_t CNMEM_API cnmemPrintMemoryState(FILE *file, cudaStream_t stream);
+
+/**
+ * \brief Converts a cnmemStatus_t value to a string.
+ */
+const char* cnmemGetErrorString(cnmemStatus_t status);
```
abergeron: This function should be tagged CNMEM_API too.

doc/library/config.txt
nouiz: Exact. I'll make it more clear.

theano/sandbox/cuda/blocksparse.py
nouiz: I'm not sure of the overhead difference. As we have the code, I'll keep it like this. Also, cnmem could be disabled.

theano/sandbox/cuda/cnmem.h
nouiz: What is CNMEM_EXPORT? I don't see it anywhere.

nouiz: Also, I don't understand the line 38. It is never defined in this PR.

theano/sandbox/cuda/blocksparse.py
abergeron: ok

theano/sandbox/cuda/cnmem.h
abergeron: Oh I meant `CNMEM_DLLEXPORT`

And if that is never defined, then it probably doesn't work on windows.

We can probably define it in cuda_ndarray.cu before including these files to have proper support.

nouiz:
I updated the PR. I killed travis as it won't test the last changes. @abergeron can you test it on Mac?

abergeron:
This should also work on mac and linux, but we should test it to make sure.

PR #3199, opened 2015-07-23
Fix bad change to gpuarray optimizations.

abergeron:


nouiz:
Can you explain how you fix what it tried to fix?

PR #3200, opened 2015-07-23
Inline optimization for destroyhandler.py

t13m:
It makes following methods inline.
 - getroot
 - add_impact
 - get_impact

This speed up local_elemwise_inplace optimization


nouiz:
thanks

PR #3202, opened 2015-07-23
Allow slicing of tensor variables to return symbolic shape refs

craffel:
This addresses https://groups.google.com/forum/#!topic/theano-users/AUuHNstlVYY.  It allows for this kind of stuff to work:

```Python
>>> import theano
>>> import numpy as np
>>> import theano.tensor as T
>>> x = T.tensor4('x')
>>> a, b = x.shape[:2]
>>> a.eval({x: np.zeros((5, 4, 3, 2), dtype=theano.config.floatX)})
array(5)
>>> a = x.shape[3:10]
>>> a.eval({x: np.zeros((5, 4, 3, 2), dtype=theano.config.floatX)})
array([2])
```

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5,6 +5,7 @@ import warnings
 
 import numpy
 from six.moves import xrange
+import numbers
 
 import theano
 from theano.compat import izip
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -3922,15 +3922,20 @@ def get_vector_length(v):
...
+         isinstance(v.owner.op.idx_list[0], slice))):
+        start = extract_constant(theano.tensor.subtensor.get_idx_list(
+            v.owner.inputs, v.owner.op.idx_list)[0].start)
+        stop = extract_constant(theano.tensor.subtensor.get_idx_list(
+            v.owner.inputs, v.owner.op.idx_list)[0].stop)
+        if start is None:
+            start = 0
+        if stop is None:
+            stop = 0
+        return stop - start
```
nouiz: In the past, this function always returned int value. This change allow it to return symbolic variable. Replace that line by:

```
if isinstance(stop, int) and isinstance(start, int):
    return stop - start
```

nouiz:
Can you add in the file tensor/tests/test_basic.py a test for the case in the mailing list? To be sure it don't get broken again. With the other small comments, it look good. thanks.

p.s. I'll kill travis on this commit, as it is overloaded and need to run with the modified version.

craffel:
Ok, added a test and added the small fix.

nouiz:
Thanks, I think it is good to merge when travis pass. As there is back-log, this will be merged probably tomorrow.

craffel:
> Thanks, I think it is good to merge when travis pass. As there is back-log, this will be merged probably tomorrow.

Cool, thanks!

nouiz:
The new test fail with python 3:

```

ERROR: theano.tensor.tests.test_basic.test_symbolic_slice

    self.test(*self.arg)

  File "/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/tensor/tests/test_basic.py", line 7735, in test_symbolic_slice

    a, b = x.shape[:2]

  File "/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/tensor/var.py", line 442, in __iter__

    for i in xrange(theano.tensor.basic.get_vector_length(self)):

  File "/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/tensor/basic.py", line 3940, in get_vector_length

    raise ValueError("length not known")

ValueError: length not known

```

Can you check that?

craffel:
Hah, looks like for some reason in Python3 `type(stop)` was `numpy.int64` so `isinstance(stop, int)` was `False`.  Changing to `isinstance(stop, numbers.Integral)` fixes that; the main issue is that `numbers` isn't present in Python 2.5.  Does `Theano` try to support 2.5?  If not, it should be good to go now.

nouiz:
travis failure is due to a problem in master that is fixed, so I merge now.

thanks.

nouiz:
For your question, Theano do not support python 2.5 anymore.

thanks

PR #3203, opened 2015-07-23
Cudnn 3d benchmark

yaoli:
task description:

What you need to do is make an optimization that isn't enabled by default, but if enable by the user, would convert the op GpuCorrMM3d to GpuDnnConv3d, and the same for the 2 grad ops. You need to test that by default, we don't introduce it and that if we include the opt mode.including('opt_name'), we do it. Also, double check the timming, currently, previous pre-release version had the cudnn version slower then the cormm version. That is why we don't want to enable it by default. But it use less memory, so we want to allow this option.
------------------------------------------------------------------
So far I've done the part of replacing GpuCorr3dMM with GpuDnnConv3d, not with the grads yet. I need to lean more about the difference between those grad ops in order to transform one to the other.


nouiz:
This PR should have been done on the branch in @carriepl repo, not here.

PR #3206, opened 2015-07-24
Continued work on #3018: Stacktrace fix

julianser:
@lamblin Can you have a look at the comments in some of the tests? For example, I'm not sure how to exclude all opts except for "local_subtensor_make_vector" in the class "test_local_subtensor_make_vector".

### DIFF
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -199,7 +199,7 @@ optdb.register('merge1', gof.MergeOptimizer(),
 
 # rearranges elemwise expressions
 optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
-               1, 'fast_run', 'fast_compile')
+               1, 'fast_run', 'fast_compile', 'canonicalize_db')
 # Register in the canonizer Equilibrium as a clean up opt the merge opt.
 # Without this, as the equilibrium have ignore_newtrees=False, we
 # won't merge all nodes if it is set as a global optimizer with
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -450,7 +450,7 @@ class FunctionGraph(utils.object2):
                             assert path is not None
                             tr = getattr(r.tag, 'trace', [])
                             detailed_err_msg = ""
-                            if len(tr) > 0:
+                            if type(tr) is list and len(tr) > 0:
                                 detailed_err_msg += "\nBacktrace when the variable is created:\n"
 
                                 # Print separate message for each element in
--- a/theano/gof/link.py
+++ b/theano/gof/link.py
@@ -168,7 +168,7 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
 
     # Print node backtraces
     tr = getattr(node.outputs[0].tag, 'trace', [])
-    if len(tr) > 0:
+    if type(tr) is list and len(tr) > 0:
         detailed_err_msg += "\nBacktrace when the node is created:\n"
 
         # Print separate message for each element in the list of batcktraces
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -542,7 +542,7 @@ class PureOp(object):
                     "For compute_test_value, one input test value does not"
                     " have the requested type.\n")
                 tr = getattr(v.tag, 'trace', [])
-                if len(tr) > 0:
+                if type(tr) is list and len(tr) > 0:
                     detailed_err_msg += (
                         " \nBacktrace when that variable is created:\n")
                     # Print separate message for each element in the list
--- a/theano/gof/utils.py
+++ b/theano/gof/utils.py
@@ -77,6 +77,7 @@ def add_tag_trace(thing, user_line=1):
     if limit == -1:
         limit = None
     tr = simple_extract_stack(limit=limit)[:-1]
+
     # Different python version use different sementic for
     # limit. python 2.7 include the call to extrack_stack. The -1 get
     # rid of it.
@@ -93,7 +94,11 @@ def add_tag_trace(thing, user_line=1):
                   "theano/sparse/", "theano\\sparse\\",
                   "theano/typed_list/", "theano\\typed_list\\",
                   ]:
-            if p in file_path:
+            # Julian: I added the 'tests' exception together with Arnaud.
+            # Otherwise, we'd lose the stack trace during in our test cases
+            # (e.g. in test_opt.py). We're not sure this is the right way to
+            # do it though.
+            if p in file_path and 'tests' not in file_path:
                 tr = tr[:-1]
                 rm = True
                 break
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1841,7 +1841,7 @@ def local_subtensor_make_vector(node):
     elif isinstance(idx, Variable):
         if idx.ndim == 0:
             # if it is a constant we can do something with it
-            try:
+            try:               
                 v = get_scalar_constant_value(idx)
                 if isinstance(v, numpy.integer):
                     # Python 2.4 wants to index only with Python integers
@@ -1851,6 +1851,8 @@ def local_subtensor_make_vector(node):
                     ret = [x.owner.inputs[v]]
                 except IndexError:
                     raise NotScalarConstantError("Bad user graph!")
+
+                # Copy over stack trace from previous output to new output
                 return ret
             except NotScalarConstantError:
                 pass
@@ -2960,8 +2962,20 @@ def local_subtensor_of_dot(node):
 
     a_sub = a.__getitem__(tuple(a_indices))
     b_sub = b.__getitem__(tuple(b_indices)) if b_indices else b
+    
+    # Copy over previous output stacktrace to a_sub and b_sub,
+    # because an error in the subtensor operation (e.g. an index error) 
+    # on either a or b must correspond to an error in the 
+    # subtensor operation on their dot product.
+    copy_stack_trace(node.outputs[0], [a_sub, b_sub])
+    
+    # Copy over previous output stacktrace and previous dot product stacktrace,
+    # because an error here may correspond to an either in either the original
+    # dot product, or in the dot product after the subtensor operation.
+    r = T.dot(a_sub, b_sub)
+    copy_stack_trace([node.outputs[0], node.inputs[0]], r)
 
-    return [T.dot(a_sub, b_sub)]
+    return [r]
 
 
 @register_canonicalize
@@ -3016,6 +3030,11 @@ def local_IncSubtensor_serialize(node):
             new_inputs = ([i for i in node.inputs if not movable(i)] +
                           [mi.owner.inputs[0] for mi in movable_inputs])
             new_add = T.add(*new_inputs)
+            
+            # Copy over stacktrace from original output, as an error 
+            # (e.g. an index error) in this add operation should 
+            # correspond to an error in the original add operation.
+            copy_stack_trace(node.outputs[0], new_add)
 
             # stack up the new incsubtensors
             tip = new_add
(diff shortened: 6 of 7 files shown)


theano/gof/fg.py
```diff
@@ -402,7 +402,7 @@ def find_path_to(output_var, input_var):
                             assert path is not None
                             tr = getattr(r.tag, 'trace', [])
                             detailed_err_msg = ""
-                            if len(tr) > 0:
+                            if type(tr) is list and len(tr) > 0:
```
lamblin: Why is it necessary to check the type of `tr`? Could it be done using `isinstance` instead?

theano/gof/utils.py
```diff
@@ -93,7 +94,11 @@ def add_tag_trace(thing, user_line=1):
                   "theano/sparse/", "theano\\sparse\\",
                   "theano/typed_list/", "theano\\typed_list\\",
                   ]:
-            if p in file_path:
+            # Julian: I added the 'tests' exception together with Arnaud.
+            # Otherwise, we'd lose the stack trace during in our test cases
+            # (e.g. in test_opt.py). We're not sure this is the right way to
+            # do it though.
```
lamblin: I think it will work for now, but maybe in the future we would want an option to keep the whole stack trace, then we can compare them.

theano/tensor/opt.py
```diff
@@ -1633,6 +1631,8 @@ def local_subtensor_make_vector(node):
                     ret = [x.owner.inputs[v]]
                 except IndexError:
                     raise NotScalarConstantError("Bad user graph!")
+
+                # Copy over stack trace from previous output to new output
```
lamblin: If I understand correctly, `ret` is a variable that was already in the graph, so we do not have to add a new stack trace. In fact, the code is not doing it, so I think we should remove that comment.

theano/tensor/opt.py
```diff
@@ -1781,7 +1784,7 @@ def local_alloc_unary(node):
             # Is it really necessary to copy over stack trace here?
             # after all, T.alloc and T.cast should preserve the stack trace from x,
             # but perhaps the trace is lost in "v = node.op(x)"?
```
lamblin: `alloc` and `cast` are not keeping the stack trace from `x`.
I think we should change the comment to say that it is not necessary for the moment to add stack traces to `v`, or the output of `T.cast`, but it could be revisited.

theano/tensor/opt.py
```diff
@@ -2747,6 +2797,7 @@ def local_subtensor_merge(node):
 @register_specialize
 @gof.local_optimizer([Subtensor])
 def local_subtensor_of_alloc(node):
+    #TODO Julian: Document this better!
```
lamblin: Do you want to address that during that PR?
Otherwise, I would remove the comment, as it may not make sense for someone else later.

theano/tensor/opt.py
```diff
@@ -2920,6 +2988,11 @@ def movable(i):
                 assert tip.type == o_type
                 assert tip.type == mi.owner.inputs[0].type
                 tip = mi.owner.op(tip, *mi.owner.inputs[1:])
+                # Copy over stacktrace from outputs of the original 
+                # "movable" operation to the new operation. 
+                # Julian: Do we want to also include the stacktace of the output (node.outputs[0])?
```
lamblin: Yes, I think the stack trace of `node.outputs[0]` should be included here as well, as it was for ǹew_add`.

theano/tensor/opt.py
```diff
@@ -2949,6 +3022,11 @@ def local_inplace_setsubtensor(node):
             set_instead_of_inc=node.op.set_instead_of_inc,
             destroyhandler_tolerate_aliased=dta)
         new_node = new_op(*node.inputs)
+        # Copy stacktrace from original outputs to new outputs.
+        # This should be sensible, because the new operation is the 
+        # same as the old one, but now with different attributes?
+        # Julian: Pascal, is this correct?
```
lamblin: That's exactly right.

theano/tensor/opt.py
```diff
@@ -2967,6 +3045,12 @@ def local_inplace_incsubtensor1(node):
     if isinstance(node.op, AdvancedIncSubtensor1) and not node.op.inplace:
         new_op = node.op.clone_inplace()
         new_node = new_op(*node.inputs)
+
+        # Copy stacktrace from original outputs to new outputs.
+        # This should be sensible, because the new operation is the 
+        # same as the old one, but now with different attributes?
+        # Julian: same as above, is this correct?
```
lamblin: Yes, the inplace optimizations are the simplest :)

theano/gof/fg.py
julianser: Good question. I don't remember why (it's been months since I did this), but maybe the trace is not always a list but sometimes a string?

Anyway, I will change the code to use isinstance() instead.

theano/gof/utils.py
julianser: Maybe this should be a new CCW task?

theano/tensor/opt.py
julianser: Agreed. That was a bogus comment.

theano/tensor/opt.py
julianser: I'll change the comment, but leave the actual code alone (I think copying it over like this is fine).

theano/tensor/opt.py
julianser: I don't know how to document it. I will remove the comment for now, but you could add it as an issue if you want?

theano/tensor/opt.py
```diff
@@ -3346,6 +3475,12 @@ def local_join_empty(node):
             assert ret.dtype == o.dtype
             assert ret.ndim == o.ndim
             ret = T.patternbroadcast(ret, node.outputs[0].broadcastable)
+
+        # Copy over stacktrace from previous output (after join op)
+        # to new output, because an error in the new op must be caused
+        # by an error in the old join op.
+        copy_stack_trace(node.outputs, ret)
```
lamblin: Here, I would add the stacktrace to `ret` before checking the type (when it is the output of the new `T.Join`, and then add it also to the output of `T.patternbroadcast`. Otherwise, if the new Join fails but there is a patternbroadcast afterwards, there will be no trace on the Join.

theano/tensor/opt.py
```diff
@@ -3401,22 +3546,29 @@ def local_useless_switch(node):
...
-                out = T.cast(out, node.outputs[0].dtype)
-            if out.type.broadcastable != node.outputs[0].type.broadcastable:
+            if correct_out.dtype != node.outputs[0].dtype:
+                out = T.cast(correct_out, node.outputs[0].dtype)
+            if correct_out.type.broadcastable != node.outputs[0].type.broadcastable:
                 # We need to copy data to the new dimensions during execution
-                out = T.alloc(out, *[node.outputs[0].shape[i] for i
-                                     in xrange(out.ndim)])
+                out = T.alloc(correct_out, *[node.outputs[0].shape[i] for i
+                                     in xrange(correct_out.ndim)])
```
lamblin: If we need to return something else than `correct_out` directly, maybe we should keep the trace of `node.outputs`.
For instance, let's say we have:
```python
x = T.scalar()
y = T.matrix()
z = T.switch(1, x, y)
```
Then `z` will be replace by `alloc(x, ...)`, which could fail because of a memory allocation error. In that case, I guess it would be confusing to report this `alloc` as being created by `T.scalar()`, but it becomes more helpful to trace it back to the `switch` with a large matrix `y`.

theano/tensor/opt.py
```diff
@@ -3401,22 +3546,29 @@ def local_useless_switch(node):
...
-            if out.dtype != node.outputs[0].dtype:
-                out = T.cast(out, node.outputs[0].dtype)
-            if out.type.broadcastable != node.outputs[0].type.broadcastable:
+            if correct_out.dtype != node.outputs[0].dtype:
+                out = T.cast(correct_out, node.outputs[0].dtype)
+            if correct_out.type.broadcastable != node.outputs[0].type.broadcastable:
                 # We need to copy data to the new dimensions during execution
-                out = T.alloc(out, *[node.outputs[0].shape[i] for i
-                                     in xrange(out.ndim)])
+                out = T.alloc(correct_out, *[node.outputs[0].shape[i] for i
```
lamblin: Also, I'm not sure if it is your code or introduced by the merge, but both `correct_out`'s dtype and broadcastable pattern are different, the first `out` gets ignored.

theano/tensor/opt.py
julianser: I see. So then I guess the solution is to copy over the stacktrace from both T.switch() and T.scalar().

theano/tensor/tests/test_opt.py
```diff
@@ -5571,6 +5753,20 @@ def test_local_join_empty():
     assert all([not isinstance(n.op, Join) or len(n.inputs) == 4
                 for n in e if isinstance(n.op, Join)])
     assert f.maker.fgraph.outputs[0].dtype == config.floatX
+    
+    # Julian: we can enable the following test, once we 
+    # remove default optimizations.
+    # When we set optimizer=None, no optimizations should be applied,
+    # but that's not the case now...
```
lamblin: I'm not sure I understand this comment. Let's remove that part from the committed code before merging.

theano/gof/utils.py
lamblin: Yes, that would make sense.

lamblin: Created as #3641.

theano/tensor/opt.py

theano/tensor/opt.py
lamblin: Created as #3642.

lamblin:
I finally finished the review, sorry about the delay.

theano/tensor/opt.py
julianser: Yes, there was a bug here. I will fix it.

lamblin:
Travis is failing because of flake8 formatting at least, I did not check all reports.

julianser:
@lamblin I think everything should be ready to merge now.

lamblin:
Great, thanks!

theano/compile/mode.py
```diff
@@ -199,7 +199,7 @@ def apply(self, fgraph):
 
 # rearranges elemwise expressions
 optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
-               1, 'fast_run', 'fast_compile')
+               1, 'fast_run', 'fast_compile', 'canonicalize_db')
```
nouiz: Good idea, but I would have named it canonicalize_eq. @lamblin what do you think? I can change this in another PR.

theano/gof/fg.py
nouiz: this change haven't been done.

theano/gof/link.py
```diff
@@ -168,7 +168,7 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
 
     # Print node backtraces
     tr = getattr(node.outputs[0].tag, 'trace', [])
-    if len(tr) > 0:
+    if type(tr) is list and len(tr) > 0:
```
nouiz: idem.

theano/gof/op.py
```diff
@@ -542,7 +542,7 @@ def _get_test_value(cls, v):
                     "For compute_test_value, one input test value does not"
                     " have the requested type.\n")
                 tr = getattr(v.tag, 'trace', [])
-                if len(tr) > 0:
+                if type(tr) is list and len(tr) > 0:
```
nouiz: idem

theano/compile/mode.py
lamblin: We named it `_db` because it includes only the `gof.DB` itself, not all the optimizations that usually go in it.
I'm not sure what you have in mind. If you want to clarify that it is an EquilibriumDB, then we could maybe tag it `canonicalize_eqdb`. But the point is that nothing else than this empty DB should use that tag.

PR #3207, opened 2015-07-24
pep8 fix in master

nouiz:


nouiz:
I merge now as this fix test_flake8 in master.

PR #3208, opened 2015-07-24
ENH: allow numpy.*(theano_var) to build a graph

aflaxman:
This PR follows the pattern demonstrated in #3112 to allow more numpy functions to build a graph in a manner that is convenient for PyMC3 users. 

This includes everything listed in https://github.com/Theano/Theano/issues/3040#issuecomment-114097347 except for the following: `degrees, fabs, radians, rint`.  I suspect that these will not be missed, since they all have simple alternatives.

I think that this can close issue #3040.


nouiz:
thanks.

PR #3213, opened 2015-07-27
Pydotprint, fix printing of reused output and updated shared var

nouiz:


nouiz:
@lamblin, I think I did all the change for the SS. Other idea?

nouiz:
@cangermueller what do you think about this PR? This will affect you. I did some change of how pydotprint work and I'm not sure how it will affect you. I changed some colors too.

theano/printing.py
```diff
@@ -878,45 +897,70 @@ def apply_name(node):
...
+                        label += " "
+                    label += str(var.type)
+                    if len(label) > max_label_size:
+                        label = label[:max_label_size - 3] + '...'
+                    param['label'] = label
+                g.add_edge(pd.Edge(astr, varstr, **param))
+                if high_contrast:
+                    g.add_node(pd.Node(varstr, shape=var_shape))
+                else:
+                    g.add_node(pd.Node(varstr, shape=var_shape))
```
lamblin: What is the difference between the "if" and "else" cases?

theano/printing.py
```diff
@@ -642,11 +642,20 @@ def pydotprint(fct, outfile=None,
     label each edge between an input and the Apply node with the
     input's index.
 
-    Green boxes are inputs variables to the graph,
-    blue boxes are outputs variables of the graph,
-    grey boxes are variables that are not outputs and are not used,
-    red ellipses are transfers from/to the gpu (ops with names GpuFromHost,
-    HostFromGpu).
+    Variable color code::
```
lamblin: I think this is the syntax for introducing a code-block. If I remember correctly, the "-" below should be aligned just below "Variable", and there should be only one ":" here.

lamblin:
The more I think about it, the more it would make sense to me to have different nodes on the graph for the different variables that are:
- the shared variable
- its update expression

and then maybe a special edge going from the update expression to the shared variable to denote the update.

One case where the current printing can be confusing is if you have something like:
```python
x = theano.shared(...)
new_x = x * 2
out = new_x - 1
f = theano.function([], out, updates=[(x, new_x)])
```
Then it is not clear from the debugprint if `out` is defined as `new_x - 1` or as `x - 1`.

cangermueller:
Sorry for the late reply. My implementation is now independent of `pydotprintvariable`. So you can go ahead from my side.

nouiz:
@cangermueller

The change I did include some fixes. Can you verify that the updates of shared variable and function when one output is also used in the graph show correctly? Here is some code to test:

```
import numpy as np

import theano
from theano.printing import debugprint, pydotprint
x = T.vector('x')
W = T.matrix('W')
b = T.vector('b')
dot = T.dot(x, W)
out = T.nnet.sigmoid(dot + b)
debugprint(dot)
compact = True
f = theano.function(inputs=[x, W], outputs=dot)
g = theano.function([x, W, b], out)
h = theano.function([x, W, b], [dot, out])
i = theano.function([x, W, b], [dot + b, out])
pydotprint(f, outfile='pydotprint_f.png')
pydotprint(g, outfile='pydotprint_g.png')
pydotprint(h, outfile='pydotprint_h.png', compact=compact)
pydotprint(i, outfile='pydotprint_i.png', compact=compact)



np.random.seed(42)
W_val = np.random.randn(4, 3)
x_val = np.random.rand(4)
b_val = np.ones(3)
x = T.vector('x')
y = T.vector('y')
W = theano.shared(W_val)
b = theano.shared(b_val)
dot = T.dot(x, W)
out = T.nnet.sigmoid(dot + b)
f = theano.function([x], dot)  # W is an implicit input
g = theano.function([x], out)  # W and b are implicit inputs
print(f(x_val))

C = ((out - y) ** 2).sum()
dC_dW, dC_db = theano.grad(C, [W, b])
upd_W = W - 0.1 * dC_dW
upd_b = b - 0.1 * dC_db

cost_and_perform_updates = theano.function(
        inputs=[x, y],
        outputs=C,
        updates=[(W, upd_W),
                 (b, upd_b)])
pydotprint(cost_and_perform_updates, outfile='pydotprint_cost_and_perform_updates.png', compact=compact)

```

theano/printing.py
```diff
@@ -1026,7 +1073,7 @@ def var_name(var):
         if len(varstr) > max_label_size:
             varstr = varstr[:max_label_size - 3] + '...'
         var_str[var] = varstr
-        return varstr
+        return varstr, varlabel
```
lamblin: Pyflakes complains that `varlabel` is undefined. And indeed, I don't see it defined...

cangermueller:
[I tried the example here](https://goo.gl/hGw781) but it does not work. It seems that an edge attribute in `param` passed to `pd.Edge(varid, aid, **param)` is an `int`, but should be a `str`.

nouiz:
I should be fixed both problems.

PR #3216, opened 2015-07-28
Fixed the docstring for the scan_opt.

caglar:
Fixes the docstring for the scan optimizations.


nouiz:
It seem good for me. @carriepl is it ok with you?

carriepl:
The first one is fine.
The second one should be something like : A global optimizer for pushing out the variables inside the scan that depend  only on constants and sequences.

nouiz:
@caglar can you finish this and rebase it?

nouiz:
Pop. This is small think to finish. Could you finish this? thanks.

caglar:
@nouiz, OK fixed it.

nouiz:
thanks

PR #3221, opened 2015-07-29
[ENH] Scan PushOutSeqScan

carriepl:
This fixes the optimization issue raised recently raised on the theano-users mailing list :
https://groups.google.com/forum/#!topic/theano-users/_GKQXt-WPOk

This make an opt applied more frequently and remove useless warning shown to the user.

### DIFF
--- a/theano/configparser.py
+++ b/theano/configparser.py
@@ -116,6 +116,10 @@ def change_flags(**kwargs):
                          if v.fullname == k]
                     assert len(l) == 1
                     l[0].__set__(None, old_val[k])
+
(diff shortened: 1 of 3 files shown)


carriepl:
Done.
Including a fix to said decorator, for good measure; otherwise the name of the decorated test is changed and nosetests doesn't run it anymore because the new name doesn't look like the name of a unit test.

nouiz:
seem good to me. We can merge when travis pass.

carriepl:
Merging since Travis passes and @nouiz has previously reviewed and given his benediction.

Issue #3177, opened 2015-07-20
Clone with replacement that works in Scan's inner graph


lamblin:
See https://groups.google.com/d/topic/theano-dev/Imx0zxKIqi8/discussion

lamblin:
A promising approach would be to re-use the graph optimization mechanism.
Replacement for variables would be specified by a function similar to the ones used to define local optimizers. The optimizer could also have a special case for Scan or OpFromGraph nodes (which have an internal graph), and replace those by a version built on the same optimizer applied to the inner graph.

Milestones would be:
- [ ] A system cloning a graph and replacing variables tagged in a certain way, from the inputs to the outputs (in topological order), ignoring the inner graph of scan for the moment. Each variable is either replaced by another one, or left alone: no variables are created or removed.
- [ ] An extension on the first system that applies itself recursively to the inner graph of scan (and opfromgraph?)
- [ ] An extension that allows to specify variables from the inner graph of scan to be added to scan's outputs, so they are made available to the outer graph.

For the first system, a couple of pointers:
- http://deeplearning.net/software/theano/extending/optimization.html : it goes through the process of defining an `Optimizer`, global then local, and how to apply it manually.
- `FunctionGraph`: https://github.com/Theano/Theano/blob/master/theano/gof/fg.py#L48
- `LocalOptimizer`: https://github.com/Theano/Theano/blob/master/theano/gof/opt.py#L769
- `TopoOptimizer`: https://github.com/Theano/Theano/blob/master/theano/gof/opt.py#L1546 to transform a local optimizer into a global one.

Pseudo-code could look like:
```python
fg = FunctionGraph(inputs, outputs, clone=True)

@local_optimizer(None)
def local_transform(node):
    outs = node.outputs
    new_outs = []
    for out in outs:
        if getattr(out.tag, 'my_tag', False):
            new_outs.append(...)
        else:
            new_outs.append(out)
    return new_outs

topo_transform = TopoOptimizer(local_transform, 'in_to_out')
topo_transform.optimize(fg)

# Get the cloned/replaced variables
new_outs = fg.outputs
new_vars = fg.variables

# Clean up
fg.disown()
```

lamblin:
@koningrobot expressed interest in this and has been assigned that ticket.

nouiz:
I think this part of the description should be updated:


   - .... Each variable is either replaced by another one, or left alone:
   no variables are created or removed.

I think it would be more clear to tell:



[...]

PR #3222, opened 2015-07-30
Introduce map_variables (work in progress, do not merge)

koningrobot:
This is to show my progress on issue #3177.


nouiz:
I killed travis as the commit new code isn't executed.

carriepl:
What happens when doing replacements such that some of the initial inputs are not used anymore? Does this cause any issue with the current implementation?

PR #3225, opened 2015-07-31
[WIP] Add scrutinizer config from blocks with small modif. Run flake8

nouiz:
Don't merge yet. This would allow to have flake8 run very rapidly on each PR when well configured.

It could also give some comments/feedback on the code quality.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,161 @@
+build:
+  dependencies:
+    override:
+      - pip install -r req-scrutinizer.txt
+  tests:
+    override:
+      - flake8 --filename=* bin
+#      - flake8 --ignore=E501 theano
+       - nosetests theano/tests/test_flake8.py
+#      - pep257 bin --numpy --ignore=D100,D101,D102,D103 --match='.*'
+#      - pep257 blocks --numpy --ignore=D100,D101,D102,D103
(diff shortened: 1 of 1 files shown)


.scrutinizer.yml
```diff
@@ -0,0 +1,154 @@
+build:
+  tests:
+    override:
+      - flake8 --filename=* bin
+       - nosetests theano/tests/test_flake8.py
```
bartvm: I think it's actually the extra space that is throwing Scrutinizer off, not the comments.

nouiz: That seem exact. A fix allowed to go further.

nouiz:
I think it is ready to merge.

abergeron:
This could very well work, when our source code actually passes flake8 globally, which is far from the case right now.  We still need that giant whitelist.

theano/tests/test_flake8.py
```diff
@@ -3,9 +3,14 @@
 """
 
 from nose.plugins.skip import SkipTest
-import os
+import os, sys
```
abergeron: This makes flake8 unhappy.

nouiz: Yes, but why scrutinizer didn't complain? That is the problem I have with this PR, it should have also failed! I must have a bug in my config. But I don't know where. @bartvm do you have an idea why it scrutinizer wouldn't have failed?

PR #3226, opened 2015-07-31
OpFromGraph.c_code()

ChienliMa:
Draft idea of OpFromGraph.c_code().
The basic idea is to pile all inner node's C code in one section of code.
And add some intermediate variable to link them.

Am I on the wrong direction?


ChienliMa:
Currently there is a problem that code generated by ```def make_loop```  was reported a error:
```
 std::vector< std::pair<int, int> >OutVar4448878032_loops(2);
 std::vector< std::pair<int, int> >::iterator OutVar4448878032_loops_it = OutVar4448878032_loops.begin();
```
get :
``` 
 error: no member named 'vector' in namespace 'std'
    std::vector< std::pair<int, int> > OutVar4448878032_loops(2);
    ~~~~~^
```
This only happen when initializing intermediate variable's iterator.
And I stuck here for two days.

nouiz:
Just to tell that travis have errors.

nouiz:
Did you try to add this include:

#include <vector>

theano/compile/builders.py
```diff
@@ -137,6 +137,46 @@ def perform(self, node, inputs, outputs):
             # we wont need this copy anymore
             output[0] = variable.copy()
 
+    def c_code(self, node, name, inputs, outputs, sub):
+        order = io_toposort(self.new_inputs, self.new_outputs)
+
+        # assert all inner nodes have c_code()
+        if not (self._op_use_c_code and
+                all([getattr(node.op, '_op_use_c_code', False) and
+                     hasattr(node.op, 'c_code') for node in order])):
```
nouiz: All op have a default c_code method that raise utils.MethodNotDefined

nouiz: Can you test that this way:

                thunk = node.op.make_c_thunk(node, storage_map, compute_map,
                                             no_recycling)

and provide dummy storage_map and compute_map. Set an empty list for no_recycling.

theano/compile/builders.py
```diff
@@ -137,6 +137,46 @@ def perform(self, node, inputs, outputs):
             # we wont need this copy anymore
             output[0] = variable.copy()
 
+    def c_code(self, node, name, inputs, outputs, sub):
+        order = io_toposort(self.new_inputs, self.new_outputs)
```
nouiz: This will make you use the not optimized function. use self.fn.inputs and self.fn.outputs.

theano/compile/builders.py
```diff
@@ -137,6 +137,46 @@ def perform(self, node, inputs, outputs):
...
+                for i, var in enumerate(inner_node.outputs):
+                    if var in self.new_outputs:
+                        out_names.append(outputs[i])
+                    else:
+                        new_name = 'OutVar' + str(id(var))
+                        out_names.append(new_name)
+                        local.setdefault(new_name, var.clone())
+                        c_code += var.type.c_declare(new_name, sub)
+
+                c_code += inner_node.op.c_code(inner_node, name,
```
nouiz: I don't think this is the good strategy. I should have gave you more guideline. Sorry for this.

There is already a linker that compile a function into a full thunk. Mode(linker='c') generate that. This use the CLinker class. I think it is better to not try to duplicate all that code.

I'm not sure how to reuse it correctly. @abergeron you started work on that for scan I think. Do you still have that around?

abergeron: I gave up before I got to the part where I would have had to compile the inner function of scan.  So I don't have any code that would help.

nouiz: It seem it was me that started that!

https://github.com/nouiz/Theano/tree/op_from_graph

Only the 4 first commit are usefull. The last 2 are now useless.

Can you check that? Maybe there is enough comments.

ChienliMa:
Thanks. I will have al look today.

nouiz:
I didn't had the time to look, but if you aren't sure how to continue you can look at the class Composite in scalar/basic.py. It take a graph of scalar operation. Maybe this could help you. I'm not sure of this, this need to be checked.

ChienliMa:
Ok. But currently I don't know what this feature is expected to be. So I am sort of lose direction.

Should I implement all methods related to C code? Or just make sure that make_thunk() can return a cthunk. 

In current code I just follow the commits you gave me and modified it to use optimized fgraph.

ChienliMa:
I Tried to reuse codes of ```basic.py```. It's like my first implementation.

There's two problem:
1.  All test can be passed except the last two, for there were repeated declaration. I will try to remove them
2.  If I compile a op more than twice without pruge the cache, there will be a run time error. However, the test still can be passed.

After the above problems solves I will try to remove duplicated codes.

```
ERROR (theano.gof.opt): Optimization failure due to: local_dot_to_dot22
  File "/Users/ChienliMa/Code/Theano/theano/gof/opt.py", line 1483, in process_node
    replacements = lopt.transform(node)
  File "/Users/ChienliMa/Code/Theano/theano/tensor/blas.py", line 1763, in local_dot_to_dot22
    return [_dot22(*node.inputs)]
  File "/Users/ChienliMa/Code/Theano/theano/gof/op.py", line 509, in __call__
  File "/Users/ChienliMa/Code/Theano/theano/tensor/blas.py", line 1675, in make_node
    outputs = [T.tensor(x.type.dtype, bz)]
  File "/Users/ChienliMa/Code/Theano/theano/tensor/basic.py", line 745, in tensor
    return TensorType(*args, **kwargs)(name=name)
  File "/Users/ChienliMa/Code/Theano/theano/gof/type.py", line 323, in __call__
    return utils.add_tag_trace(self.make_variable(name))
  File "/Users/ChienliMa/Code/Theano/theano/tensor/type.py", line 401, in make_variable
    return self.Variable(self, name=name)
RuntimeError: maximum recursion depth exceeded while calling a Python object
```

theano/compile/builders.py
```diff
@@ -97,19 +99,40 @@ def __init__(self, inputs, outputs, **kwargs):
...
+        # or module cache key here.
+        return (self._c_code == other._c_code)
 
     def __hash__(self):
         # TODO: use internal variables in hash
-        return hash(type(self))
+        return  hash((type(self),
+                      self.nin,
+                      self.nout,
+                      self._c_code))
```
abergeron: This (and \__eq__) could be replaced by `__props__ = ('nin', 'nout', '_c_code')`

theano/compile/builders.py
```diff
@@ -121,14 +144,110 @@ def make_node(self, *inputs):
...
-                                                  compute_map, no_recycling)
-        if not hasattr(self, "fn"):
-            self.fn = orig_function(self.new_inputs,
-                                    self.new_outputs,
-                                    **self.kwargs)
+        ret = super(OpFromGraph, self).make_thunk(node,
+                                                  storage_map,
+                                                  compute_map,
+                                                  no_recycling)
         return ret
```
abergeron: Why did you move the creation of `self.fn` to the constructor?

theano/compile/builders.py
```diff
@@ -121,14 +144,110 @@ def make_node(self, *inputs):
...
+                [subd[output] for output in node.outputs],
+                dict(fail="%(fail)s", id="%%(id)s_%i" % j))
+            _c_code += s
+            _c_code += "\n"
+        _c_code += "}\n"
+
+        for char in 'idflxsc':
+            _c_code = re.sub('\%'+char, "%%"+char, _c_code)
+
+        self._c_code = _c_code
```
abergeron: This should share code with CLinker.  Otherwise it's bound to become obsolete and not work anymore sometime soon.

theano/compile/builders.py
```diff
@@ -121,14 +144,110 @@ def make_node(self, *inputs):
...
+            xv = x.op.c_code_cache_version()
+            if xv:
+                rval.append(xv)
+            else:
+                return ()
+        return tuple(rval)
+
+    def c_support_code(self):
+        rval = []
+        for subnode in self.fgraph.toposort():
```
abergeron: You don't need a toposort here.  Just iterating over the nodes in the graph in any order should do fine.

theano/compile/builders.py
```diff
@@ -121,14 +144,110 @@ def make_node(self, *inputs):
...
+                if subnode_support_code:
+                    rval.append(subnode_support_code)
+            except gof.utils.MethodNotDefined:
+                pass
+        # there should be no need to remove duplicate code blocks because
+        # each block should have been specialized for the given nodename.
+        # Any block that isn't specialized should be returned via
+        # c_support_code instead of c_support_code_apply.
+        return "\n".join(rval)
+
```
abergeron: There are lots of methods that help C code generation other than c_support_code() and c_support_code_apply().

See http://deeplearning.net/software/theano/extending/cop.html#what-needs-to-be-defined to get the list.

theano/gof/cmodule.py
```diff
@@ -153,7 +153,7 @@ def __init__(self, name=None):
 
         self.support_code = []
         self.functions = []
-        self.includes = ["<Python.h>", "<iostream>", '"theano_mod_helper.h"']
+        self.includes = ["<Python.h>", "<iostream>", '"theano_mod_helper.h"',"<vector>"]
```
abergeron: Why do you need to do that?  If it is for a specific Op, add it to that op's c_include().

theano/gof/fg.py
```diff
@@ -739,11 +739,11 @@ def __repr__(self):
         return self.__str__()
 
     # clone #
-    def clone(self, check_integrity=True):
+    def clone(self, check_integrity=True, attach_feature=False):
```
abergeron: This changes the default behavior for all clients, which is a bad idea.  Make it default to True and call it with False when you need it.

Also I wonder why you need this change.  You don't seem to clone FunctionGraphs anywhere.

abergeron:
All of my comments are based on what direction the code seems to be taking, but from what I can infer there was discussion on just having make_thunk() return a C code thunk directly, which I think is a much better idea and is probably much simpler.

So if that is the direction you take you can probably ignore a lot of the comments I made.

theano/compile/builders.py
ChienliMa: Because I want to have the optimized fgraph earlier.

ChienliMa:
Yes, there's a version of code that directly return the c_thunk, which I think is better, too. And I restore the version now.

PR #3228, opened 2015-07-31
Doctest

harlouci:


doc/extending/cop.txt
```diff
@@ -256,7 +256,7 @@ We will be defining C code for the multiplication Op on doubles.
 .. If you modify this code, also change :
 .. theano/tests/test_tutorial.py:T_extending.test_extending_2
 
-.. code-block:: python
+.. testcode::
```
abergeron: This one here should not be testcode

doc/extending/ctype.txt
```diff
@@ -526,7 +530,7 @@ know how to generate C code.
 
 You can implement c_code for this op. You register it like this:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/ctype.txt
```diff
@@ -548,7 +552,7 @@ ViewOp to generate C code when working with this type, as
 otherwise it will use Python code instead. This is achieved by
 calling:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/ctype.txt
```diff
@@ -568,7 +572,7 @@ Theano Variable that has a shape attribute (Shape_i returns only one of
 the elements of the shape).
 
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/cop.txt
abergeron: Or rather there should be a declaration of `mul`like this:

```
mul = object()
```

doc/extending/ctype.txt
```diff
@@ -156,12 +156,16 @@ out:
 Defining the methods
 ====================
 
+.. testsetup:: 
+
+    import theano
+
```
abergeron: You should add a definition of double like this:

```python
class Double(gof.Type):

    def filter(self, x, strict=False, allow_downcast=None):
        if strict and not isinstance(x, float):
            raise TypeError('Expected a float!')
        return float(x)

    def values_eq_approx(self, x, y, tolerance=1e-4):
        return abs(x - y) / (x + y) < tolerance

    def __str__(self):
        return "double"

double = Double()
```

doc/extending/fibby.txt
```diff
@@ -143,7 +145,7 @@ the correct size for the output. This is essentially simulating the line
 ``y = x.copy()``.
 
 
-.. code-block:: python
+.. testcode::
```
abergeron: This is not python code, but C code.  Also should not be tested.

doc/extending/graphstructures.txt
```diff
@@ -153,13 +155,13 @@ All nodes in the graph must be instances of ``Apply`` or ``Result``, but
...
-.. code-block:: python
+.. testcode::
 
-    e = dscalar('x') + 1
+    e = T.dscalar('x') + 1
 
 builds the following graph:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This shouldn't be testcode

doc/extending/graphstructures.txt
```diff
@@ -399,7 +402,7 @@ In both types of pairs, the second element of the tuple is an index,
 such that: ``var.clients[*][0].inputs[index]`` or
 ``fgraph.outputs[index]`` is that variable.
 
-.. code-block:: python
+.. testcode::
```
abergeron: The comments below like:

```
# [(Sum(Elemwise{add,no_inplace}.0), 0)]
```
Are actually things that would be outputs were that entered at the python prompt.

Maybe the section should be converted to a `>>>` block.

doc/extending/inplace.txt
```diff
@@ -55,7 +55,7 @@ Suppose you had an Op which took ``x`` as input and returned
 purpose, you would set the ``view_map`` field as follows:
 
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/inplace.txt
```diff
@@ -66,7 +66,7 @@ inputs that are viewed by a given output, this feature is currently
 unsupported. Here are more examples:
 
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/inplace.txt
```diff
@@ -144,7 +147,7 @@ Theano needs to be notified of this fact. The syntax is similar to
 that of ``view_map``:
 
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/inplace.txt
```diff
@@ -153,7 +156,7 @@ What this means is that the first output (position 0) operates inplace on the
 first input (position 0).
 
 
-.. code-block:: python
+.. testcode::
```
abergeron: This should not be testcode

doc/extending/op.txt
```diff
@@ -598,6 +631,7 @@ Trying out our new Op
 
 In the following code, we use our new Op:
 
+>>> import theano
```
abergeron: You would need to add `.. doctest:: mul` to make the doctest block part of mul.

Same with the other doctest block below.

doc/extending/optimization.txt
```diff
@@ -221,8 +223,9 @@ computation, using the ``merge_optimizer`` defined in
 
 >>> from theano.gof.opt import merge_optimizer
 >>> merge_optimizer.optimize(e)
+(0, 0.0001430511474609375, None, None, {}, 1, 0)
```
abergeron: This include a timing (the second element of the tuple) that should be replaced with `...` and `#  +doctest: ELLIPSIS`

doc/extending/optimization.txt
```diff
@@ -295,18 +298,18 @@ subset of them) and applies one or several local optimizers on them.
...
+>>> y = float64('y')
+>>> z = float64('z')
+>>> a = add(z, mul(true_div(mul(y, x), y), true_div(z, x)))
 >>> e = gof.FunctionGraph([x, y, z], [a])
 >>> e
-[add(z, mul(div(mul(y, x), y), div(z, x)))]
+[add(z, mul(true_div(mul(y, x), y), true_div(z, x)))]
 >>> simplify = gof.TopoOptimizer(local_simplify)
 >>> simplify.optimize(e)
+(<theano.gof.opt.TopoOptimizer object at 0x7f3219787f90>, 1, 5, 3, 0.00017309188842773438, 0.00020599365234375, 6.4849853515625e-05)
```
abergeron: This includes varying timings so `...`

doc/extending/optimization.txt
```diff
@@ -435,7 +438,7 @@ Query
 
 A Query is built by the following call:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This is not testcode

doc/extending/other_ops.txt
```diff
@@ -90,7 +90,7 @@ and (like in SciPy) they do not support broadcasting operations by default
 formats for sparse type: ``csr`` and ``csc``. So in ``make_mode()``,
 you can create output variables like this:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This doesn't look like test code.

doc/extending/unittest.txt
```diff
@@ -115,7 +117,7 @@ built-in unittest module uses metaclasses to know about all the
 them all, printing '.' for passed tests, and a stack trace for
 exceptions. The standard footer code in theano's test files is:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This is not test code.

doc/extending/unittest.txt
```diff
@@ -238,7 +243,7 @@ Example:
 
 Avoid hard-coding variables, as in the following case:
 
-.. code-block:: python
+.. testcode:: writeUnitest
```
abergeron: This is not test code

doc/extending/unittest.txt
```diff
@@ -134,7 +136,7 @@ To run all the tests in one or more ``TestCase`` subclasses:
 
 To run just a single ``MyTestCase`` member test function called ``test0``:
 
-.. code-block:: python
+.. testcode::
```
abergeron: This is not test code

doc/tutorial/examples.txt
```diff
@@ -487,50 +490,490 @@ A Real Example: Logistic Regression
...
+    ...
+    Final model:
+    ...
+    ...
+    target values for D:
+    ...
+    prediction on D:
+    ...
+
+.. code-block:: none
```
abergeron: I don't think we actually want a giant matrix in the docs.  Just the testoutput block is fine.

abergeron:
Closing this since the follow-up is done in #3284

PR #3230, opened 2015-08-01
Stack trace fix (needed for the summer school)

nouiz:
It fix one condition created in gh-3077.

This also fix the small comment there.


nouiz:
thanks

PR #3232, opened 2015-08-01
make MergeOptimizer merge nodes with assert input

t13m:


t13m:
I don't know how this change will affect performance. I'll put the profiling here later.

t13m:
Profiling output:
 - with this change: https://gist.github.com/t13m/64383eec4aead5e1aea8#file-merge-assert
 - without this change: https://gist.github.com/t13m/64383eec4aead5e1aea8#file-note-merge-assert

Optimization time from 847s go down to 812s, and there is a small increase of number of merged node in `gpu_opt`.

nouiz:
I'm not sure the speed difference is significant. Did you had different load on the computer for both run time case? This could explain that. I'm not expecting this PR to speed up computation as it is the other one that add more assert and I think the current master won't benefit from the extra assert merge. But combined with the other PR, this would be useful. Also, it is good for other case to have this PR, not just for compilation speed.

I don't see the difference in gpu_opt you talk about, can you send url to the different lines?

theano/gof/opt.py
```diff
@@ -533,19 +544,64 @@ def process_node(self, fgraph, node):
...
+            for i in candidate.inputs:
+                if i.owner and isinstance(i.owner.op,
+                                          theano.tensor.opt.Assert):
+                    cand_has_assert = True
+                    cand_inputs_assert_removed.append(i.owner.inputs[0])
+                else:
+                    cand_inputs_assert_removed.append(i)
+
+            # if the current node has assert input, it should not be
+            # replaced with a candidate node which has no assert input
```
nouiz: In that case, why not replace the candidate with the new node that have the assert?

theano/gof/opt.py
```diff
@@ -533,19 +544,64 @@ def process_node(self, fgraph, node):
...
+                # replace node with candidate
+                if not (node_has_assert or cand_has_assert):
+                    # Schedule transfer of clients from node to candidate
+                    pairs = list(zip(node.outputs,
+                                     candidate.outputs,
+                                     ['merge'] * len(node.outputs)))
+                else:
+                    new_inputs = self.get_merged_assert_input(node, candidate)
+                    new_node = node.op(*new_inputs)
+                    pairs = list(zip(node.outputs,
```
nouiz: Just add a comment that we replace node and the candidate with a new node that will have the assert.

theano/gof/opt.py
```diff
@@ -558,6 +614,39 @@ def process_node(self, fgraph, node):
...
+                    # Step 2. Combine the two with T.and_(a, b)
+                    node_cond = node_i.owner.inputs[1]
+                    cand_cond = cand_i.owner.inputs[1]
+                    if node_cond.owner is cand_cond.owner:
+                        new_inputs.append(cand_i)
+                    else:
+                        new_inputs.append(
+                            theano.tensor.opt.assert_op(
+                                node_i.owner.inputs[0],
+                                theano.tensor.and_(node_cond, cand_cond)))
```
nouiz: Assert op can have many condition. I think it will be better to have multiple condition then one more complicated one.

You can do:
new_cond = node_i.owner.inputs[1:] + cand_i.owner.inputs[1:]
assert_op(node_i.owner.inputs[0], *new_cond)

This will handle assert with multiple condition and make the graph simpler.

nouiz:
Except for my small comments, it seem good. thanks.

t13m:
> In the other PR, you named many cases. Can you make a test for each of them?

Ok.

> I don't see the difference in gpu_opt you talk about, can you send url to the different lines?

It's on Line 238, 242, 246 of https://gist.github.com/t13m/64383eec4aead5e1aea8#file-merge-assert

theano/gof/opt.py
t13m: I thought that in every iteration on `node`, only `node` was the one that should be replaced. So it is ok to replace the `candidate`?

theano/gof/opt.py
t13m: Should the order of conditions be kept? I use `new_cond = list(set(node_i.owner.inputs[1:] + cand_i.owner.inputs[1:]))`, so the order can be changed in the set.

theano/gof/opt.py
nouiz: Yes it is OK to replace candidate. It will be new but OK.

theano/gof/opt.py
nouiz: The order can be changed. Anyway there is no reason to put node before
candidate or the other way around.

nouiz:
I found the diff you pointed at. But I don't understand why the line at 161-164 don't show corresponding difference. Do you understand?

nouiz:
I had a few small comments. Otherwise, I think it is close be merged.

nouiz:
travis failed with this:
```

FAIL: theano.gof.tests.test_opt.TestMergeOptimizer.test_both_assert_merge_1

    self.test(*self.arg)

  File "/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/tests/test_opt.py", line 414, in test_both_assert_merge_1

    assert strg == strref

AssertionError
```

Can you replace the assert line by 
    assert strg == strref, (strg, strref)

That way, we see the data with the error.

nouiz:
We can merge this after the deep learning summer school at montreal Thursday. I prefer to be extra safe to not introduce problem during that time.

But can you answer the question I have, do you understand why Line 238, 242, 246 changed but not line 161-164 in the output? Personally, I don't.

PR #3238, opened 2015-08-03
Fix outdated description of border_mode in GpuCorrMM constructor

f0k:
#2259 changed `BaseGpuCorrMM` to accept custom padding in `border_mode` and warn against using `pad`, but the docstring of `GpuCorrMM` was not updated accordingly. This PR updates the docstring to match the implementation. It removes the warning that `border_mode='full'` can be slower than `GpuCorrMM_gradInputs` because that would need to go with a lengthy explanation on how to use the latter instead, it's not *always* slower, and it's probably just confusing end users.

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -726,7 +726,7 @@ class BaseGpuCorrMM(GpuOp):
                 'do not use pad for BaseGpuCorrMM; please set padding in '
                 'border_mode parameter, see the docstring for more details')
             if border_mode != "valid":
-                raise ValueError("border_mode must be 'valid'")
+                raise ValueError("border_mode must be 'valid' if pad is given")
             border_mode = pad
         if isinstance(border_mode, int):
             border_mode = (border_mode, border_mode)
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/blas.py
```diff
@@ -995,20 +995,20 @@ def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width
...
-        simulated by setting `pad="full"` (at the cost of performance), or
-        by using `GpuCorrMM_gradInputs`
+    :param border_mode: the width of a border of implicit zeros to pad the
+        input with. Must be a tuple with 2 elements giving the numbers of rows
+        and columns to pad on each side, or a single integer to pad the same
+        on all sides, or a string shortcut setting the padding at runtime:
+        ``'valid'`` for ``(0, 0)`` (valid convolution, no padding), ``'full'``
+        for ``(kernel_rows - 1, kernel_columns - 1)`` (full convolution),
+        ``'half'`` for ``(kernel_rows // 2, kernel_columns // 2)`` (same
+        convolution for odd-sized kernels).
```
nouiz: thanks for the update. But I think this description isn't clear that the number given is used at the top and bottom. It is not the total split in 2. There was system with the other semantic.

I'm not sure how to make this clear, so if you don't know how to update it, I'll merge, but if you find how to clarify this, it would be great.

f0k:
> But I think this description isn't clear that the number given is used at the top and bottom.

I thought it's quite clear with "to pad on each side" and "to pad the same on all sides". But I can add a sentence in the end:
```
Note that the two widths are each applied twice, once per side (left and right, top and bottom).
```

f0k:
Okay, amended the commit accordingly.

nouiz:
thanks

f0k:
Closing/Reopening in the hope that this re-triggers Travis and Scrutinizer.

f0k:
Something's wrong with Scrutinizer, but it's not me. I've only changed strings here. Feel free to merge!

nouiz:
Forget scrutinizer, it don't work well. thanks.

PR #3242, opened 2015-08-03
Fix memmap test error on windows.

abergeron:


nouiz:
strange fix. But I'll merge. Do the temp file could be the same for multiple tests? Do you think that making sure the test file have different file name could have fixed this?

abergeron:
No the files are different, but the memmap object lingers after the tests in some conditions so the file is still "in use".  A gc collect flushes it out.

PR #3245, opened 2015-08-04
CuDNN v3

carriepl:
Add support for CuDNN V3.
This includes : 
-Wrapping the 3d convolution
-Supporting the new features for selecting the most efficient convolution implementation either by timing them or by guessing according to heuristics.
-Support for the new log-softmax implementation (integrated into existing op for CuDNN softmax and inserted via optimizations)
-Adding tests for the corresponding features


nouiz:
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/cuda/tests/test_dnn.py:1064:20: E126 continuation line over-indented for hanging indent

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/cuda/tests/test_dnn.py:1157:33: E126 continuation line over-indented for hanging indent

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/cuda/tests/test_dnn.py:1225:33: E126 continuation line over-indented for hanging indent

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/cuda/tests/test_dnn.py:1238:37: E126 continuation line over-indented for hanging indent

nouiz:
I merge now, even if there is small doc update to do. They can be done tomorrow.

benanne:
Sweet! You guys are on the ball :)

f0k:
Kudos from me as well for merging that quickly! Unfortunately, for the Tesla K40 I tried it on, only the algorithm for bprop wrt. weights seems to have changed, sometimes it's faster and sometimes it's slower than before. Activating `time_once` makes things terribly slow (is it really timed only once per Op instance? and does it use the shape of the first invocation for that?). `time_on_shape_change` and `guess*` choose the FFT algorithm for the forward pass, which turns out to be slower than the default one. On a Maxwell card, things will probably look different, though -- to bad I don't have one.

carriepl:
For ``time_once``, it will only be timed by apply node in the graph (if two different apply nodes of your graph were reusing the same op instance, they would each do their own timing at their first execution).
Is it possible that your input and kernel shapes often vary and the shapes at the first execution are not representative of the shapes at most executions? What are the parameters of your convolution (kernel shape, input shape, border_mode, padding, subsampling, ...)?

f0k:
> Is it possible that your input and kernel shapes often vary and the shapes at the first execution are not representative of the shapes at most executions?

No, they're constant throughout. I've just posted a way to reproduce the behaviour in https://github.com/Theano/Theano/pull/3252#issuecomment-128316370, feel free to have a look!

carriepl:
Thanks, I'll have a look right now.

PR #3246, opened 2015-08-04
Clearer error message for missing pydot/graphviz.

lucasb-eyer:
Very minor, but I was puzzled by the error message since I had `pydot` installed, but not `graphviz`.


nouiz:
thanks

PR #3248, opened 2015-08-05
fix dnn_conv() direction_hint check

f0k:
Commit 05dc20d introduced a check for `direction_hint` in `dnn_conv()` that is too strict. `dnn_conv()` is regularly called with `direction_hint='bprop inputs'`, and the meta-optimizer may also call it with `direction_hint='forward!'` (the latter was supposed to change, but in discussing it we decided to wait until #2665 is finished, which seemed to be within a few weeks at that time).


f0k:
Superseded by #3250. Will be closed automatically if the latter is accepted.

PR #3250, opened 2015-08-05
Fix GpuDnnConvGradI not being inserted automatically

### LINKED ISSUE #2644: Get rid of the hack in dnn_conv to use the gradient ops
Timings I did of the convolution indicate that the forward op is much faster when workmem='small' than the gradient ops.  And by much faster I mean 0.5 s vs 10 s for shapes of (64, 3, 227, 227) and (96, 3, 11, 11).

For tiny sizes, like (1, 3, 8, 8) and (4, 3, 2, 2), its 5.1e-4 vs 4.5e-4 which is a bit slower for the plain forward of, but not by much.

We should disable the hack that tries to use the gradient ops by default and let people that want max speed in all cases use the meta-optimizer for convolution.

### LINKED ISSUE #3249: cuDNN convolution strong performance regression
6a63844 renders the automatic introduction of the cuDNN convolution partly ineffective. It prevents `GpuDnnConvGradI` being introduced for the gradient wrt. inputs. See the performance difference before and after that commit, using cuDNN v2:

**Before:**
```
$ SKIP=legacy,fft,convnet,meta,gemm python pylearn2_benchmark.py 1
Using gpu device 0: Tesla K40c

CONFIG: input = 3 x 128 x 128 * ker = 3 x 96 x 11 x 11 ( bs = 128 , stride = 1 )
(auto) theano.sandbox.cuda.dnn.GpuDnnConv          ==> fprop         ==>      78
(auto) theano.sandbox.cuda.dnn.GpuDnnConv          ==> bprop inputs  ==>     111
(auto) theano.sandbox.cuda.dnn.GpuDnnConv          ==> bprop weights ==>     105

(manual conv) theano.sandbox.cuda.dnn.GpuDnnConv   ==> fprop         ==>      79
(manual conv) theano.sandbox.cuda.dnn.GpuDnnConv   ==> bprop inputs  ==>     111
(manual conv) theano.sandbox.cuda.dnn.GpuDnnConv   ==> bprop weights ==>     105
```

**After:**
```
$ SKIP=legacy,fft,convnet,meta,gemm python pylearn2_benchmark.py 1
Using gpu device 0: Tesla K40c (CNMeM is disabled)

CONFIG: input = 3 x 128 x 128 * ker = 3 x 96 x 11 x 11 ( bs = 128 , stride = 1 )
(auto) theano.sandbox.cuda.dnn.GpuDnnConv          ==> fprop         ==>      78
(auto) theano.sandbox.cuda.dnn.GpuDnnConv          ==> bprop inputs  ==>    1069
(auto) theano.sandbox.cuda.dnn.GpuDnnConv          ==> bprop weights ==>     106

(manual conv) theano.sandbox.cuda.dnn.GpuDnnConv   ==> fprop         ==>      79
(manual conv) theano.sandbox.cuda.dnn.GpuDnnConv   ==> bprop inputs  ==>     112
(manual conv) theano.sandbox.cuda.dnn.GpuDnnConv   ==> bprop weights ==>     106
```

Inspecting the graph shows that `GpuDnnConv` is being used for the backward pass after the commit, while `GpuDnnConvGradI` was used before that commit.

What was the rationale for changing it?

There are three possible solutions:

1. Removing the `version()` check again
2. Changing the `version()` check to include cuDNN v2
3. Changing the check to only use `GpuDnnConvGradI` when `direction_hint == 'bprop inputs'` (this was proposed before, but then we decided to wait until #2665 was finished, which seemed to be due shortly at that time)

f0k:
#3245 rendered part of the `local_conv_dnn()` graph optimizer inoperable (due to an assertion on `direction_hint` in `dnn_conv()` being too strict, as described in #3248) and slow (due to `GpuDnnConvGradI` not being used any more, as described in #3249). This PR fixes both in one go, going for the third solution outlined in #3249.
It also resolves #2644, via the first solution proposed in https://github.com/Theano/Theano/issues/2644#issuecomment-84032628.

Fixes #3249. Closes #3248.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -1115,7 +1115,7 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1),
         If border_mode is 'valid', subsample is (1,1) and direction_hint is
         'bprop weights', it will use GpuDnnConvGradW.
         If border_mode is 'full', subsample is (1,1) and direction_hint is
-        *not* 'forward!', it will use GpuDnnConvGradI.
+        'bprop inputs', it will use GpuDnnConvGradI.
         This parameter is used internally by graph optimizers and may be
         removed at any time without a deprecation period. You have been warned.
     :param workmem: *deprecated*, use param algo instead
@@ -1138,7 +1138,7 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1),
         algo = workmem
 
     # Ensure the value of direction_hint is supported
-    assert direction_hint in [None, 'bprop weights', 'forward']
+    assert direction_hint in [None, 'bprop weights', 'bprop inputs', 'forward']
 
     fgraph = getattr(img, 'fgraph', None) or getattr(kerns, 'fgraph', None)
     if (border_mode == 'valid' and subsample == (1, 1) and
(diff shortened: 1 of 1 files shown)


carriepl:
This looks good. Thanks.

PR #3251, opened 2015-08-05
Fix copy/paste bug for cuDNN time_on_shape_change mode

f0k:
`THEANO_FLAGS=dnn.conv.algo_bwd=time_on_shape_change` didn't work due to a copy/paste bug in `dnn.py`.


nouiz:
thanks

PR #3255, opened 2015-08-05
Exempt np.random.mtrand.RandomState from NaN/inf/big_number checking in NanGuardMode.

cvangysel:
NanGuardMode applies np.ndarray operations on RandomState, which causes a crash.

Traceback:
    File "/home/cvangysel/Theano/theano/compile/nanguardmode.py", line 59, in contains_nan return np.isnan(np.min(arr))
    TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''


nouiz:
thanks

PR #3256, opened 2015-08-06
enable dots in section names

mop:
Config parameters in CUDNN v3 have more than one hierarchy level
(i.e. dnn.conv.algo_fwd). At the moment these values cannot by changed by .theanorc.
This patch corrects these issue.


carriepl:
I tested it locally and can confirm that it works. Thanks a lot for the fix.

PR #3258, opened 2015-08-06
Early detection of device capability for proper cuda_ndarray compilation

ivannz:
Dear developers,

upon a fresh installation of **theano** I ran into the problem where there was a mismatch between the architecture on which **cuda_ndarray** library was compiled and the architecture setting of compiled apply nodes. Basically calling some kernel functions from **cuda_ndarray.cu** (specifically **k_elemwise_unary_rowmajor_copy**) resulted in "invalid device function" error. Due to hardware reasons I have to use CUDA driver 6.5, which still supports old devices like the one I have: GeForce 320M (SM version 1.2).

The problem hid in the fact that the code for device capability detection resides in **cuda_ndarray** module, which had to be compiled upon the very first import of theano. The compilation in **__ init __.py** of **theano/sandbox/cuda** invokes **nvcc_compiler.compile_str()**, which in turn tries to detect device capabilities using the yet nonexistent **cuda_ndarray** module, and obviously fails. Thus the compilation of **cuda_ndarray** defaulted to some architecture of nvcc compiler, which turned out to be backwards incompatible with sm12.

That is why is patched the **__ init __.py** code so that it first tries to compile a lightweight module for early device capability detection (**cuda_devquery.cu**), and then, upon success, uses the correct device dependent flags for **cuda_ndarray** compilation.

I had to modify some extra code in **__ init __.py**, specifically the parts where CUDA availability is detected and where the compiled module is checked for being correct and up-to-date (former **try_import()**).

Please consider this patch request and send your remarks and suggestions.

Ivan


ivannz:
I want to add that this patch addresses a specific issue mentioned in the comment in **NVCC_compiler.compile_args()** at line _183_ of **sandbox/cuda/nvcc_compiler.py**, and not just some problem with my own hardware.

Also I found out that the NVCC compiler detection routine **is_nvcc_available()** at line _83_ (same file) first attempts to check the version of nvcc compiler on **PATH** (environment variable) and only then, upon failure, does it try the path in **config.cuda.root** set either via user-specified **CUDA_ROOT** environment variable or through autodetection routine **find_cuda_root()**(both from **default_cuda_root()**). According to the purpose of the user defined path to the CUDA runtime (either through explicit config value of the CUDA_ROOT) this is undesired: it is easy to imagine a use case when two distinct version of CUDA runtime are installed either for testing or for compatibility purposes and one is on the path, while the other is the one to be used for **theano**.

Furthermore the static method **NVCC_compiler.try_compile_tmp()** does not perform the action it is designed to do due to inflexibility: it refers to a method **Compiler** class in **cmodule.py**, which is hardcoded to produce **.c** source code files which makes nvcc invoke the system default c compiler. I am sure this is not how this functionality is supposed to work.

I have refactored and abstracted some code in my local **nvcc_compiler.py** so that the logic becomes more straightforward and clear, but I have some questions concerning the usage of the compatibility fixer **b()** from **compat/six.py** with **output_subprocess_Popen**, and why further down in **NVCC_Compiler.compile_str()** the actual compilation is done via the pattern
**p = subprocess.Popen(...); out, err = decode_iter( p.communicate( )[ :2 ] )** and not with the help of **output_subprocess_Popen** (which does the same thing only in a more controlled fashion).

In addition I would wholeheartedly welcome any feedback regarding the usefulness and desirability (by the community) of this small refactoring effort.

I am looking forward to your feedback.

nouiz:
Thanks for the patch. Currently, we are hosting a deep learning summer school, so we don't have much time to review it in detail, but from a high level, it seem good.

We should check it next week.

ivannz:
Thank you!

I shall continue working on patching **nvcc_compiler.py**, and  when the patch is ready and sound I will bundle it with this pull request. This will take at most a day.

theano/sandbox/cuda/__init__.py
```diff
@@ -150,23 +132,92 @@ def try_import():
...
-    # compilation with -L and also adds -lcuda_ndarray when compiling
-    # modules.
-    nvcc_compiler.add_standard_rpath(cuda_ndarray_loc)
-    compile_cuda_ndarray = not try_import()
 
+# If $TMPDIR is defined, nvopencc wants it to exist
+if cuda_available and 'TMPDIR' in os.environ :
+    tmpdir = os.environ[ 'TMPDIR' ]
+    if not os.path.exists( tmpdir ) :
+        os.makedirs( tmpdir )
```
abergeron: I would say that this part is up to the user to fix.  We shouldn't attempt to create it ourselves.

theano/misc/windows.py
```diff
@@ -2,6 +2,14 @@
 import subprocess
 
 
+def __check_params(params, forbidden=['stdout', 'stderr', 'stdin']):
```
abergeron: Usually private functions start with a single underscore.

theano/misc/windows.py
```diff
@@ -42,38 +50,33 @@ def subprocess_Popen(command, **params):
...
-    # we need to use communicate to make sure we don't deadlock around
-    # the stdour/stderr pipe.
-    out = p.communicate()
-    return out + (p.returncode,)
+    # Communication with subprocesses should be done with proc.communicate()
+    # to avoid deadlocks around the stdour/stderr pipe.
+    proc = subprocess_Popen(command, **params)
+    _stdout, _stderr = proc.communicate()
+    _exit_code = proc.returncode
+    return _stdout, _stderr, _exit_code
```
abergeron: Local variables beginning with an underscore are just annoying to type.

theano/sandbox/cuda/__init__.py
```diff
@@ -104,45 +104,27 @@ def set_cuda_disabled():
...
-        # be in sys.path.
-        sys.path[0:0] = [config.compiledir]
-        import cuda_ndarray.cuda_ndarray
-        del sys.path[0]
-    except ImportError:
+# Add the compilation dir to the list of paths for module import
+if config.compiledir not in sys.path :
+    sys.path.insert( 0, config.compiledir )
+
+def check_module( module, library, source_files ):
```
abergeron: No whitespace between parentheses and parameters.

abergeron:
Overall I like this idea.  However I had a weird feeling while reading the code like something was wrong and I realized it's because the style is really different from the usual Theano style.

I made some comments pointing out the differences where I could see them, but you may want to try and mimic the surrounding code.

theano/sandbox/cuda/__init__.py
ivannz: This chunk of code had already been in **__ init __.py** prior to my pull request. I kept it to ensure compatibility.

ivannz:
I have resolved the issues you pointed out and have done my best at trying to maintain the usual coding style of Theano.

ivannz:
I have detected and resolved an issue, where **cuda_devquery** refused to be imported by python's **__ import __()** primitive. It turns out that upon its compilation the **nvcc** compiler masks the module initialization function.

The reason was, that in _NVCC_compiler.compile_str()_ (in **nvcc_compiler.py**)  a novel option **hide_symbols** had been introduced which defaulted to **True**. Basically, it forced the compiler to hide all symbols, which were defined in the **CU** source file. Unfortunately, this included the exported functions prefixed by **PyMODINIT_FUNC**, which are responsible for module registration within Python. Thus I explicitly set **hide_symbols = False** in the compilation of **cuda_devquery**.

The **cuda_ndarray** compilation is done via a similar call to **compile_str()** in **__ init __.py**, but with **hide_symbols** defaulting to **True**. Won't this hide the **cuda_ndarray** symbols as well?

abergeron:
The fix for hide_symbols is to include "theano_mod_helper.h" which will redefine PyMODINIT_FUNC so that it is still exported with `hide_symbols = True`.

hide_symbols was introduced to slightly speedup module loading time (due do decreased symbol table size) and also avoid conflicts on Mac between different cuda modules.

Also, could you perform a rebase to get rid of the multiple merges in your branch.

luopuya:
Hi, I meet **invalid device function** too (CudaNdarray_TakeFrom)
How could I solve it, should I install theano from GitHub?

----------------------------------
I use theano 0.7 in WinPython 2.7, Cuda 6.5 with GT 330M.
I try simple embedding function
```Python
import theano
import numpy
IN = theano.tensor.ivector('IN')
W = numpy.arange(3*5).reshape((3, 5)).astype('float32')
EMB_W = theano.shared(value=W, name='EMBW')
EMB_OUT = EMB_W[IN]
EMB_FUNC = theano.function([IN], EMB_OUT)
IN_V = numpy.asarray([0,1,1,2]).astype('int32')
EMB_FUNC(IN_V)
```

```Text
RuntimeError: Cuda error: CudaNdarray_TakeFrom: invalid device function .

Apply node that caused the error: GpuAdvancedSubtensor1(EMBW, IN)
Inputs types: [CudaNdarrayType(float32, matrix), TensorType(int32, vector)]
Inputs shapes: [(3, 5), (4L,)]
Inputs strides: [(5, 1), (4L,)]
Inputs values: ['not shown', array([0, 1, 1, 2])]

Debugprint of the apply node: 
GpuAdvancedSubtensor1 [@A] <CudaNdarrayType(float32, matrix)> ''   
   |EMBW [@B] <CudaNdarrayType(float32, matrix)>
   |IN [@C] <TensorType(int32, vector)
```

theano/sandbox/cuda/__init__.py
abergeron: ok

nouiz:
We are droping the old back-end, so I'll close this PR.

PR #3259, opened 2015-08-06
opencl build error: duplicate ga_ types

wtfrank:
libgpuarray declares these ga_* types in gpuarray_buffer_opencl.c / gpuarray_buffer_cuda.c
causes a build error in opencl

### DIFF
--- a/theano/sandbox/gpuarray/elemwise.py
+++ b/theano/sandbox/gpuarray/elemwise.py
@@ -165,22 +165,6 @@ class GpuElemwise(HideC, Elemwise):
 #include <stdint.h>
 #endif
 """
-        # Translate ga_ pseudo-types into their specific realizations
-        support_code += """
-#define ga_bool uint8_t
-#define ga_byte int8_t
-#define ga_ubyte uint8_t
-#define ga_short int16_t
-#define ga_ushort uint16_t
-#define ga_int int32_t
-#define ga_uint uint32_t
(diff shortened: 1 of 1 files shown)


abergeron:
I would also delete the section about types on windows just above.

abergeron:
Or rather, after futher inspection, keep the other section and make the section you removed part of the conditional above about 'opencl'.

wtfrank:
Those types are also defined in the cuda path (to the same values as in the code I removed here). So currently, the cuda path defines these types twice to the same values, while the opencl path defines them twice to different values. Its not great that they are defined twice - whether to the correct values or not.

As the definition of the types varies between cuda and opencl (and potentially in the future with any 3rd acceleration system that might be implented), I think they should be defined in the relevant opencl/cuda kernels and not defined here. But if there is a compelling reason to make the decision up here, then I'm happy to change it by removing the type definitions out of the kernels and only have it defined here with a conditional on opencl vs cuda.

Also, is there another execution path we need to consider that doesn't involve cuda or opencl? It seemed to work on the CPU as well so I assume CPU-only code doesn't involve compiling code here.

abergeron:
The theano path for cuda is not the same since we extract the kernel source and compile it ourselves which is why this code was there in the first place.

So we need to keep it for now.

abergeron:
This is no longer needed due to #3364

PR #3263, opened 2015-08-06
Mixed

nouiz:


doc/tutorial/examples.txt
```diff
@@ -500,8 +500,8 @@ It will be used repeatedly.
   training_steps = 10000
   
   # Declare Theano symbolic variables
-  x = T.matrix("x")
-  y = T.vector("y")
+  x = T.dmatrix("x")
+  y = T.dvector("y")
```
lamblin: If we want to make that example work with float32, it would be better to get the code from the test instead:
https://github.com/nouiz/Theano/blob/mixed/theano/tests/test_tutorial.py#L727-L776

theano/misc/elemwise_openmp_speedup.py
```diff
@@ -49,7 +49,12 @@ def runScript(N):
     else:
         costlySpeed = costlyTimeOpenmp / costlyTime
         costlySpeedstring = "slowdown"
+    print("Timmed with vector of %d elements" % options.N)
```
lamblin: "Timed"

theano/misc/pycuda_example.py
```diff
@@ -285,6 +285,11 @@ def __init__(self, scalar_op, inplace_pattern=None, name=None):
         self.scalar_op = scalar_op
         self.inplace_pattern = inplace_pattern
 
+    # As we have a dict in props, we need to implement __hash__
+    def __hash__(self):
+        return (hash(type(self)) ^ hash(self.scalar_op) ^
+                hash_from_dict(self.inplace_pattern))
```
lamblin: Is that the right way? Or should the default `__hash__` method be updated to call `hash_from_dict` instead?

theano/sandbox/cuda/opt.py
```diff
@@ -261,7 +261,8 @@ def local_gpu_elemwise_0(node):
                     # TODO: change this when fusion makes Elemwise with multiple
                     # outputs
```
lamblin: Should that TODO be removed now?

theano/sandbox/cuda/opt.py
```diff
@@ -1068,6 +1070,8 @@ def local_gpu_incsubtensor(node):
             incsubt = host_output.owner.op
             x, y = host_output.owner.inputs[0:2]
             coords = host_output.owner.inputs[2:]
+            if x.dtype != "float32" or y.dtype != "float32":
+                return
```
lamblin: I don't think that is right. If the value of the increment is not float32, it should be casted, as this is what happens inside the "set_subtensor" anyway.
This is what the comment below is about, btw.

lamblin:
I did not review the part about elemwise with multiple outputs, as I'm not familiar with it yet.

theano/misc/pycuda_example.py
nouiz: that isn't always safe as some int have collision....

doc/tutorial/examples.txt
nouiz: This is before we talk about floatX, so I taught that forcing float64 would
make it easier to understand that example.

nouiz: I fixed the other comments.

On Sat, Aug 8, 2015 at 6:51 PM, Frédéric Bastien <frederic.bastien@gmail.com

nouiz:
I updated this PR with the comments.

doc/tutorial/examples.txt
lamblin: OK.

theano/misc/pycuda_example.py
lamblin: I'm not sure I understand. I was suggesting to update the parent `__hash__` method so it calls `hash_from_dict` on the elements of `__props__` that are dicts. How do the hash collisions for ints relate to that? Also, collisions in hashes are not necessarily a problem.
I don't like the idea of just applying xor on the different hash elements in general, because xor it is commutative.

nouiz: I tried to make the parent \__hash__ work, but I wasn't able to make it work. So I fall back to the old code that was working.

I updated the PR to make one hash of a tuple to remove the xor.

doc/tutorial/examples.txt
lamblin: So then, when we move to doctests, we should drop the more complex test and only test the case with explicit double precision. I don't think it is worth updating the test for the moment.

lamblin:
The part that still has to be reviewed is the multi-output elemwise.

abergeron:
Travis seems to be failing on flake8

theano/misc/pycuda_example.py
```diff
@@ -285,6 +285,11 @@ def __init__(self, scalar_op, inplace_pattern=None, name=None):
         self.scalar_op = scalar_op
         self.inplace_pattern = inplace_pattern
 
+    # As we have a dict in props, we need to implement __hash__
+    def __hash__(self):
+        return hash(type(self), hash(self.scalar_op),
+                hash_from_dict(self.inplace_pattern))
```
lamblin: pep8 complains about "E128 continuation line under-indented for visual indent" here.

abergeron:
Follow up in #3288

Issue #3231, opened 2015-08-01
-m32 compilation error on Raspberry Pi


davikrehalt:
I tried building theano on the Raspberry Pi 1 B and got the following error 
Problem occurred during compilation with the command line below:
````
/usr/bin/g++ -shared -g -D NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -D NPY_ARRAY_ENSUREARRAY=NPY_ENSUREARRAY -D NPY_ARRAY_ENSURECOPY=NPY_ENSURECOPY -D NPY_ARRAY_ALIGNED=NPY_ALIGNED -D NPY_ARRAY_WRITEABLE=NPY_WRITEABLE -D NPY_ARRAY_UPDATE_ALL=NPY_UPDATE_ALL -D NPY_ARRAY_C_CONTIGUOUS=NPY_C_CONTIGUOUS -D NPY_ARRAY_F_CONTIGUOUS=NPY_F_CONTIGUOUS -m32 -fPIC -I/usr/lib/pymodules/python2.7/numpy/core/include -I/usr/include/python2.7 -I/home/andy/pyjects/Theano/theano/gof -fvisibility=hidden -o /root/.theano/compiledir_Linux-3.18.11+-armv6l-with-debian-7.8--2.7.3-32/lazylinker_ext/lazylinker_ext.so /root/.theano/compiledir_Linux-3.18.11+-armv6l-with-debian-7.8--2.7.3-32/lazylinker_ext/mod.cpp -L/usr/lib -lpython2.7
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "theano/__init__.py", line 63, in <module>
    from theano.compile import (
  File "theano/compile/__init__.py", line 9, in <module>
    from theano.compile.function_module import *
  File "theano/compile/function_module.py", line 19, in <module>
    import theano.compile.mode
  File "theano/compile/mode.py", line 10, in <module>
    import theano.gof.vm
  File "theano/gof/vm.py", line 666, in <module>
    from . import lazylinker_c
  File "theano/gof/lazylinker_c.py", line 126, in <module>
    preargs=args)
  File "theano/gof/cmodule.py", line 2059, in compile_str
    (status, compile_stderr.replace('\n', '. ')))
Exception: Compilation failed (return status=1): cc1plus: error: unrecognized command line option ‘-m32’. 
````

nouiz:
In the file theano/gof/cmodule.py, around line 1896, we add the -m32
[...]

PR #3266, opened 2015-08-07
made compatible with Raspberry Pi 1

davikrehalt:
The raspberry pi 1 uses armv6. This is detected in the script and the -m32 tag is removed in this case

fix gh-3231

### DIFF
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -1893,7 +1893,7 @@ class GCC_compiler(Compiler):
         # or 64 bit and compile accordingly. This step is ignored for ARM
         # architectures in order to make Theano compatible with the Raspberry
         # Pi, and Raspberry Pi 2.
-        if not any(['arm' in flag for flag in cxxflags]) and platform.machine() != 'armv7l':
+        if not any(['arm' in flag for flag in cxxflags]) and platform.machine() not in ['armv7l','armv6l']:
             n_bits = local_bitwidth()
             cxxflags.append('-m%d' % n_bits)
             _logger.debug("Compiling for %s bit architecture", n_bits)


theano/gof/cmodule.py
```diff
@@ -1893,7 +1893,7 @@ def join_options(init_part):
         # or 64 bit and compile accordingly. This step is ignored for ARM
         # architectures in order to make Theano compatible with the Raspberry
         # Pi, and Raspberry Pi 2.
-        if not any(['arm' in flag for flag in cxxflags]) and platform.machine() != 'armv7l':
+        if not any(['arm' in flag for flag in cxxflags]) and platform.machine() not in ['armv7l','armv6l']:
```
nouiz: thanks. I think I would make this more generic: `... and not "arm" in platform.machine():`. That way, we don't need to patch it for each arm processor.

From this page, gcc do not support -m32 or -m64 for arm:

https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html#Option-Summary

moslemk: @nouiz @davikrehalt This fix fails on arm 64 bit processors. On ARMv8 (i.e., ARM64) processors platform.machine() returns 'aarch64'. Hence, the check will not fail and it will include the '-m64' in the compile list of options which in turn results in a "g++: unrecognized compile option -m64" during the compile time.

nouiz: Can you make a PR that will add 'aarch64' in the list at the end of this line? Just test it before, I think it will fix it.

Otherwise, instead of making a black list that don't support -m64, this could be changed to a white list that support it. There is much less arch that support it.

PR #3267, opened 2015-08-07
Implement batched_tensordot in terms of batched_dot

koningrobot:
This will eventually allow us to exploit sandbox.cuda.blas.BatchedDotOp to perform batched tensor dot products.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5322,8 +5322,8 @@ def _tensordot_as_dot(a, b, axes, dot, batched):
         outbcast = a.broadcastable[a_outaxes] + b.broadcastable[b_outaxes]
         outndim = len(outbcast)
 
-        a_shape = [1]*2
-        b_shape = [1]*2
+        a_shape = [1] * 2
+        b_shape = [1] * 2
 
         # compute total size of summed axes
         for i in xrange(0, axes):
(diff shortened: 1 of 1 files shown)


theano/tensor/basic.py
```diff
@@ -5269,6 +5266,129 @@ def dot(a, b):
...
+    # of a, leading axes of b), we can just reshape and use dot.
+    elif numpy.isscalar(axes):
+        axes = int(axes)
+
+        for operand_name, operand in (("a", a), ("b", b)):
+            if axes > operand.ndim:
+                raise ValueError(
+                    'axes can not be larger than the dimension of %s '
+                    '(%s.ndim=%i, axes=%i)'
+                    % (operand_name, operand_name, operand.ndim, axes))
```
nouiz: You can use another python string fonctionality to make this more readable:

"a_string %(data)s %(data)s %(out)d" %{data="test", out=2}

This give name to the string substitution

theano/tensor/basic.py
```diff
@@ -5269,6 +5266,129 @@ def dot(a, b):
...
+              (less the first dimension and any dimensions that were summed
+              over).
+    :rtype: symbolic tensor
+    """
+    a, b = as_tensor_variable(a), as_tensor_variable(b)
+
+    if not numpy.isscalar(axes) and len(axes) != 2:
+        raise ValueError('Axes should be an integer or a '
+                         'list/tuple of len 2 (%s was provided)'
+                         % repr(axes))
```
nouiz: we use str(axes) everywhere else in Theano. For consistency continue to do this.

nouiz:
thanks. I merge.

For reference, can you give here the speed up you got with this on CPU and GPU?

PR #3268, opened 2015-08-07
python 3 fix in dnn.py

pallegro:
In python 3, map returns a map object iterator rather than a list, which here gets consumed by min, leaving pad_desc empty.
This bug was just introduced with the update to cudnn v3.


nouiz:
thanks

PR #3270, opened 2015-08-10
Also match Windows version of paths

lamblin:
This should fix unhelpful stack traces on Windows.


lamblin:
Not tested on Windows yet.

lamblin:
I've just tested it on Windows with the "debug" tutorial from https://github.com/mila-udem/summerschool2015, and it fixes the display problem.

PR #3275, opened 2015-08-12
Fix buildbot

abergeron:
This fixes most of the problems on the buildbot.

Two problems remain:
 - test_use_scan_direct_output2 (theano.scan_module.tests.test_scan.T_Scan),
 - test_grad_multiple_outs_some_disconnected (theano.scan_module.tests.test_scan.T_Scan)

both in debug mode.  I'm looking at the first one, but it seems it's not a simple fix.  I haven't looked at the second one yet.

### DIFF
--- a/theano/scan_module/tests/test_scan.py
+++ b/theano/scan_module/tests/test_scan.py
@@ -43,6 +43,11 @@ if theano.config.mode == 'FAST_COMPILE':
 else:
     mode_with_opt = theano.compile.mode.get_default_mode()
 mode_with_gpu = mode_with_opt.including('gpu', 'scan')
+if theano.config.mode in ('DEBUG_MODE', 'DebugMode'):
+    mode_nodebug = theano.compile.mode.get_mode('FAST_RUN')
+else:
+    mode_nodebug = mode_with_opt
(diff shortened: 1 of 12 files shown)


abergeron:
I've fixed test_use_scan_direct_output2 which turned out to be a bug in get_scalar_constant_value()

theano/scan_module/tests/test_scan.py
```diff
@@ -1772,8 +1777,16 @@ def reset_rng_fn(fn, *args):
         analytic_grad = reset_rng_grad_fn(v_u, v_x0, vW_in)
         utt.assert_allclose(analytic_grad[0][:2], numpy.zeros((2, 2)))
 
-    @attr('slow')
```
lamblin: I'm not sure what the threshold is to tag a test as "slow", so maybe we should keep it?

lamblin:
I'm OK with the change to `filter_variable`, and the addition of `allow_convert`. Let's see how Travis reacta

theano/scan_module/tests/test_scan.py
nouiz: 1s

nouiz:
Documenting filter_variable would be great.

PR #3278, opened 2015-08-12
Fix most warnings and errors in the doc.

abergeron:


carriepl:
Looks good. I had only 1 comment but it was resolved IRL with @abergeron so I'm merging.

PR #3280, opened 2015-08-12
AdvancedSubtensor should instantiate a Variable of Scalar type when indexing a 1-dim tensor.

cvangysel:
Currently when indexing a 1-dimensional tensor AdvancedSubtensor instantiates a Variable of a 0-dim TensorType as output. However, the underlying implementation generates a primitive value (i.e. np.<dtype>) instead of a np.ndarray.

This causes confusing behaviour if this value is used as input to another Op: during runtime the `ValueError: expected an ndarray` exception is raised.

The generation of a primitive is as expected (i.e. numpy has the same behaviour: `type(numpy.ones(5)[0]) == numpy.float64`), but the Op should create a variable of the correct type.

### DIFF
--- a/theano/tensor/subtensor.py
+++ b/theano/tensor/subtensor.py
@@ -2033,10 +2033,14 @@ class AdvancedSubtensor(Op):
 
         index = tuple(map(as_index_variable, index))
         bcast = adv_index_broadcastable_pattern(x, index)
-        return gof.Apply(self,
-                         (x,) + index,
-                         [theano.tensor.tensor(dtype=x.type.dtype,
-                                               broadcastable=bcast)])
+
+        if bcast:
+            variable = theano.tensor.tensor(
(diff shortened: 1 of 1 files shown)


nouiz:
Thanks for this PR. Can you add a test in tensor/tests/test_subtensor.py.

I'm not sure this is the right fix. The perform method of that op do that you describe, but the c code will generate an ndarray. So the right fix would be in the perform method, case the scalar to an ndarray in that case. Can you also update that?

thanks.

nouiz:
I'm checking this again and I don't think of a use case of AdvancedSubtensor that will generate a scalar. Do you know one? I can call it directly and generate this case, but by normal code, it should not happen. It is more efficient to use the other class then this one for all the case I think of.

Do you remember which code was generating this case?

PR #3285, opened 2015-08-15
[WIP] CpuCorrMM closes #3026 - redux

### LINKED ISSUE #3026: Implement CpuCorrMM

JesseLivezey:
First PR was in strange state. Local commit history was different than github history and I couldn't reconcile them. Trying again...

CPU implementation of the CorrMM style convolutions for #3026.

General
- [x] Write real tests, current 'tests' are just copied from tensor.nnet, not sure if they make any sense
- [x] Get tests to pass
- [x] Move code out of sandbox and into final location (tensor.nnet?)

C code
 - [x] write im2col and col2im for cpu
 - [x] write cormm for cpu
 - [x] float 32 and 64 versions

Theano
- [x] write ops
 - [x] BaseCpuCorrMM
 - [x] CpuCorrMM
 - [x] CpuCorrMM_gradWeights
 - [x] CpuCorrMM_gradInputs

Later PRs
- Write optimization to swap conv ops (waiting on #2665)

### DIFF
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -1,0 +1,619 @@
+import os
+import logging
+
+import theano
+from theano import Apply
+from theano import gof
+from theano.tensor import as_tensor_variable, TensorType
+from theano.tensor.blas_headers import blas_header_text
+from theano.tensor.blas import ldflags
+
+
+_logger = logging.getLogger(__name__)
+
+
+class BaseCorrMM(gof.Op):
+    """
+    Base class for `CorrMM`, `CorrMM_gradWeights` and
+    `CorrMM_gradInputs`. Cannot be used directly.
+    Parameters
+    ----------
+    border_mode : {'valid', 'full', 'half'}
+        Additionally, the padding size could be directly specified by an integer
+        or a pair of integers
+    subsample
+        Perform subsampling of the output (default: (1, 1)).
+
+    """
+    check_broadcast = False
+    __props__ = ('border_mode', 'subsample')
+
+    def __init__(self, border_mode="valid", subsample=(1, 1)):
+        if isinstance(border_mode, int):
+            if border_mode < 0:
+                raise ValueError(
+                    'invalid border_mode {}, which must be a '
+                    'non-negative integer'.format(border_mode))
+            border_mode = (border_mode, border_mode)
+        if isinstance(border_mode, tuple):
+            if len(border_mode) != 2 or border_mode[0] < 0 or border_mode[1] < 0:
+                raise ValueError(
+                    'invalid border_mode {}, which must be a '
+                    'pair of non-negative integers'.format(border_mode))
+            pad_h, pad_w = map(int, border_mode)
+            border_mode = (pad_h, pad_w)
+        if not ((isinstance(border_mode, tuple) and min(border_mode) >= 0) or
+                border_mode in ('valid', 'full', 'half')):
+            raise ValueError(
+                'invalid border_mode {}, which must be either '
+                '"valid", "full", "half", an integer or a pair of'
+                ' integers'.format(border_mode))
+        self.border_mode = border_mode
+        if len(subsample) != 2:
+            raise ValueError("subsample must have two elements")
+        self.subsample = subsample
+
+    @property
+    def pad(self):
+        if self.border_mode != 'valid':
+            return self.border_mode
+        return (0, 0)
+
+    def __str__(self):
+        return '%s{%s, %s}' % (
+            self.__class__.__name__,
+            self.border_mode,
+            str(self.subsample))
+
+    def c_support_code(self):
+        return blas_header_text()
+
+    def c_libraries(self):
+        return ldflags()
+
+    def c_compile_args(self):
+        return ldflags(libs=False, flags=True)
+
+    def c_lib_dirs(self):
+        return ldflags(libs=False, libs_dir=True)
+
+    def c_header_dirs(self):
+        return ldflags(libs=False, include_dir=True)
+
+    def c_headers(self):
+        return ['<stdio.h>']
+
+    def c_code_cache_version(self):
+        # raise this whenever modifying any of the support_code_files
+        return (1, 0)
+
+    def c_support_code_apply(self, node, nodename):
+        # REMEMBER TO RAISE c_code_cache_version when changing any of
+        # these files
+        sub = {}
+        dtype = str(node.__dict__['inputs'][0].dtype)
+        assert dtype in ('float32', 'float64')
+        if dtype == 'float32':
+            sub['gemm'] = 'sgemm_'
+            sub['float_type'] = 'npy_float'
+            sub['float_typenum'] = 'NPY_FLOAT'
+            sub['n_bytes'] = 4
+            sub['c_float_type'] = 'float'
+        else:
+            sub['gemm'] = 'dgemm_'
+            sub['float_type'] = 'npy_double'
+            sub['float_typenum'] = 'NPY_DOUBLE'
+            sub['n_bytes'] = 8
+            sub['c_float_type'] = 'double'
+        files = ['corr_gemm.c']
+        codes = [open(os.path.join(os.path.split(__file__)[0], f)).read()
+                 for f in files]
+        final_code = ''
+        for code in codes:
+            final_code += code
+        return final_code % sub
+
+    def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width=None):
+        """
+        This generates the C code for CorrMM (direction="forward"),
+        CorrMM_gradWeights (direction="backprop weights"), and
+        CorrMM_gradInputs (direction="backprop inputs").
+        Depending on the direction, one of bottom, weights, top will
+        receive the output, while the other two serve as inputs.
+
+        :param bottom: Variable name of the input images in the forward pass,
+            or the gradient of the input images in backprop wrt. inputs
+        :param weights: Variable name of the filters in the forward pass,
+            or the gradient of the filters in backprop wrt. weights
+        :param top: Variable name of the output images / feature maps in the
+            forward pass, or the gradient of the outputs in the backprop passes
+        :param direction: "forward" to correlate bottom with weights and store
+            results in top,
+            "backprop weights" to do a valid convolution of bottom with top
+            (swapping the first two dimensions) and store results in weights,
+            and "backprop inputs" to do a full convolution of top with weights
+            (swapping the first two dimensions) and store results in bottom.
+        :param sub: Dictionary of substitutions useable to help generating the
+            C code.
+        :param height: If self.subsample[0] != 1, a variable giving the height
+            of the filters for direction="backprop weights" or the height of
+            the input images for direction="backprop inputs".
+
+            If self.border_mode == 'half', a variable giving the height of the
+            filters for direction="backprop weights".  Ignored otherwise.
+        :param width: If self.subsample[1] != 1, a variable giving the width
+            of the filters for direction="backprop weights" or the width of the
+            input images for direction="backprop inputs".
+
+            If self.border_mode == 'half', a variable giving the width of the
+            filters for direction="backprop weights".  Ignored otherwise.
+        """
+        dH, dW = self.subsample
+        if self.border_mode == "half":
+            padH = padW = -1
+        elif self.border_mode == "full":
+            padH = padW = -2
+        elif isinstance(self.border_mode, tuple):
+            padH, padW = self.border_mode
+        else:
+            assert self.border_mode == "valid"
+            padH = padW = 0
+        if direction == "forward":
+            direction = 0
+            out = top
+        elif direction == "backprop weights":
+            direction = 1
+            out = weights
+        elif direction == "backprop inputs":
+            direction = 2
+            out = bottom
+        else:
+            raise ValueError("direction must be one of 'forward', "
+                             "'backprop weights', 'backprop inputs'")
+        # When subsampling, we cannot unambiguously infer the height and width
+        # of bottom and weights from top, so we require them to be given.
+        # Similarly, when border_mode="half", we cannot infer the weight size.
+        if ((direction != 0) and (dH != 1)) or ((direction == 1) and (padH == -1)):
+            if not height:
+                raise ValueError("height must be given for backprop with vertical sampling or border_mode='half'")
+            height = '(*(npy_int*)(PyArray_DATA(%s)))' % height
+        else:
+            height = 'NULL'
+        if ((direction != 0) and (dW != 1)) or ((direction == 1) and (padW == -1)):
+            if not width:
+                raise ValueError("width must be given for backprop with horizontal sampling or border_mode='half'")
+            width = '(*(npy_int*)(PyArray_DATA(%s)))' % width
+        else:
+            width = 'NULL'
+        sub = sub.copy()
+        sub.update(locals())
+
+        return """
+    // Mandatory args
+    int direction = %(direction)s;  // forward, bprop weights, bprop inputs
+
+    // Optional args
+    int dH = %(dH)s;
+    int dW = %(dW)s;
+    int padH = %(padH)s;
+    int padW = %(padW)s;
+
+    PyArrayObject * bottom = %(bottom)s;
+    PyArrayObject * weights = %(weights)s;
+    PyArrayObject * top = %(top)s;
+    PyArrayObject * out2 = NULL;
+
+    // Obtain or infer kernel width and height
+    // (we need to know it early to be able to handle auto-padding)
+    int kH, kW;
+    if (direction != 1) {
+        // weight is an input variable, we can just read its shape
+        kH = PyArray_DIMS(weights)[2];
+        kW = PyArray_DIMS(weights)[3];
+    }
+    else {
+        if ((dH != 1) || (padH == -1)) {
+            // vertical subsampling or half padding, kernel height is specified
+            kH = %(height)s;
+        }
+        else if (padH == -2) {
+            // vertical full padding, we can infer the kernel height
+            kH = 2 - PyArray_DIMS(bottom)[2] + (PyArray_DIMS(top)[2] - 1) * dH;
+        }
+        else {
+            // explicit padding, we can infer the kernel height
+            kH = PyArray_DIMS(bottom)[2] + 2*padH - (PyArray_DIMS(top)[2] - 1) * dH;
+        }
+        if ((dW != 1) || (padW == -1)) {
+            kW = %(width)s;
+        }
+        else if (padW == -2) {
+            kW = 2 - PyArray_DIMS(bottom)[3] + (PyArray_DIMS(top)[3] - 1) * dW;
+        }
+        else {
+            kW = PyArray_DIMS(bottom)[3] + 2*padW - (PyArray_DIMS(top)[3] - 1) * dW;
+        }
+    }
+
+    // Auto-padding if requested
+    if (padH == -1) {  // vertical half padding
+        padH = kH / 2;
+    }
+    else if (padH == -2) {  // vertical full padding
+        padH = kH - 1;
+    }
+    else if (padH < 0) {
+        PyErr_SetString(PyExc_ValueError, "BaseCorrMM: padH must be >= -2");
+        %(fail)s
+    }
+    if (padW == -1) {  // horizontal half padding
+        padW = kW / 2;
+    }
+    else if (padW == -2) {  // horizontal full padding
+        padW = kW - 1;
+    }
+    else if (padW < 0) {
+        PyErr_SetString(PyExc_ValueError, "BaseCorrMM: padW must be >= -2");
+        %(fail)s
+    }
+
+    // Infer output shape
+    npy_intp out_dim[4];
+    switch(direction) {
+    case 0:  // forward pass
+        // output is top: (batchsize, num_filters, height, width)
+        // height and width: top = (bottom + 2*pad - weight) / sample + 1
+        out_dim[0] = (npy_intp)PyArray_DIMS(bottom)[0];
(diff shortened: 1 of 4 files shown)


JesseLivezey:
@lamblin, where should I put these files? Right now they are in the sandbox. ```tensor.nnet```?

JesseLivezey:
Failing tests are all: ```/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/cpu_corr/__init__.py:1:1: F401 'CpuCorrMM' imported but unused```, so I think everything is currently working.

It would probably be a good idea to add another test for ```infer_shape``` for ```gradInputs``` and ```gradWeights``` since they are currently untested.

JesseLivezey:
@lamblin, tests for shape inference for ```gradWeights``` and ```gradInputs``` are there; waiting for travis to finish.

Two other problems that I need feedback on are:
 - Where should this code go?
 - Some tests currently fail since there is no ```perform``` method. Do I need to implement one?

One problem that I can fix with the current tests is that I'm using the same shapes as the forward direction, and so some of the outputs happen to have equal shapes along various dimensions and so ```_compile_and_check``` throws a bunch of warnings that it can't unambiguously check the shape. I think I'll refactor the shape inference test into three separate tests.

theano/sandbox/cpu_corr/tests/test_corr.py
```diff
@@ -0,0 +1,242 @@
...
+import theano
+import theano.tensor as T
+from theano.tests import unittest_tools as utt
+from theano.sandbox import cpu_corr
+from theano.tensor.nnet import conv
+from theano.tensor.basic import _allclose
+
+
+class TestCorr2D(utt.InferShapeTester):
+    mode = None
```
nouiz: use:

mode = "FAST_RUN" if theano.config.mode = "FAST_COMPILE" else None

Or implement a perform method. I would go with the fastest change (changing mode).

theano/sandbox/cpu_corr/corr.py
```diff
@@ -0,0 +1,631 @@
...
+    out2 = corrMM(%(bottom)s, %(weights)s, %(top)s, direction, dH, dW, padH, padW);
+    if (out2==NULL){
+       %(fail)s
+    }
+    assert (out2 == %(out)s);
+
+""" % sub
+
+
+class CpuCorrMM(BaseCpuCorrMM):
```
nouiz: remove CPU from the name of all classes.

theano/sandbox/cpu_corr/tests/test_corr.py
```diff
@@ -0,0 +1,242 @@
...
+                                icol:icol + N_filter_shape[3]] * filter2d[::-1, ::-1]
+                            ).sum()
+
+        self.assertTrue(_allclose(theano_output, ref_output))
+
+        # TEST GRADIENT
+        if verify_grad:
+            utt.verify_grad(sym_CpuCorrMM, [orig_image_data, filter_data])
+
+    def test_basic(self):
```
nouiz: I suppose many of those test_* method are just copy/paste from the Conv op test class. If so, why not make this test class inherit from it to remove duplicate code?

nouiz:
Thanks for pushing this. A few answers/comments to help get this finished:

- This should go in tensor/nnet/corr.py and the tests in tensor/nnet/tests/test_corr.py The flake8 travis error  will get removed with this.
- This implementation would need documentation in doc/library/tensor/nnet/conv.txt
- For the test failing with missing perform method, you don't need to implement one, but make sure the test work with a mode that don't request one. I did another comment that tell how to do this.

thanks, not much is left to finish this.

JesseLivezey:
@nouiz, I tried inheriting from TestConv2D (the class from conv op), but it seems like all of the tests from TestConv2D are run during the Corr tests in that case (many of which do not apply). It seems like there should really be a base conv/corr test from which specific ops can inherit and then possibly add more tests. This might be too much for this PR, but I could work on it next if this would be useful.

Or is there a clean way to inherit some of the methods without making a bunch of dummy tests that override the superclass test in the Corr class?

Also, the current infer_shape tests for the gradW and gradI ops give a bunch of warnings
```WARNING (theano.tests.unittest_tools): While testing the shape inference, we received an input with a shape that has some repeated values: (5, 6, 4, 5), like a square matrix. This makes it impossible to check if the values for these dimensions have been correctly used, or if they have been mixed up.
```
Do I need to fix these? I could write specific tests for gradW and gradI with their own input shapes rather than reusing the ones from the forward CorrMM.

Finally, I've been using ```sgemm_```, ```dgemm_```, and ```blas_header_text()``` rather than the cblas versions. Will this always work? Or do I need to be more careful?

nouiz:
The only way to inherit some method that I know is to override the one that you don't want or make a a new super class with only the methods that you want and make the 2 other class inherit from it.

If you would update the test to don,t have the warning, that would be good. Mostly, what this tell is due to the used input shape, we aren't able to do the infer_shape test.

I don't see a problem of using the fortran BLAS interface, so keep it like this.

thanks

JesseLivezey:
> or make a a new super class with only the methods that you want and make the 2 other class inherit from it.

It might be useful to do this for the conv/corr tests after #2665 is finished.

JesseLivezey:
@nouiz, it looks like everything is passing. Currently, the op checks to make sure the input arrays are c-contiguous and errors if they are not. Is that reasonable behavior or should something else happen?

nouiz:
There is 2 way to fix this:

* an old way, make the op copy to contiguous memory if needed.
* wrap the input in cpu_contiguous op.

I think the 1st is better and is needed. You can use:

https://docs.scipy.org/doc/numpy/reference/c-api.array.html#c.PyArray_GETCONTIGUOUS

Don't forget to always decref what this return. If it return a view, it will incref the input.

JesseLivezey:
@nouiz, if PyArray_GETCONTIGUOUS returns a view in the case where the array is already contiguous, is it okay to just always call it and not check with PyArray_IS_C_CONTIGUOUS. There shouldn't be any additional overhead in the case where the array is already contiguous, correct?

nouiz:
Doing that is good. The code is clean and the test probably very cheap.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,636 @@
...
+        assert inputs[1] == filters[1]
+        # nb mul and add by output pixel
+        flops = filters[2] * filters[3] * 2
+        # nb flops by output image
+        flops *= outputs[2] * outputs[3]
+        # nb patch multiplied
+        flops *= inputs[1] * filters[0] * inputs[0]
+        return flops
+
+    def c_support_code(self):
```
JesseLivezey: @nouiz, one more question. I'm not sure if all of the methods from ```c_support_code``` to ```c_headers``` are actually necessary. Everything seems to work currently, but there might be needless code.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -0,0 +1,313 @@
...
+            return rval
+
+        output = sym_CorrMM(input, filters)
+        output.name = 'CorrMM()(%s,%s)' % (input.name, filters.name)
+        theano_corr = theano.function([input, filters], output, mode=self.mode)
+
+        # initialize input and compute result
+        image_data = numpy.random.random(N_image_shape).astype(self.dtype)
+        filter_data = numpy.random.random(N_filter_shape).astype(self.dtype)
+        if non_contiguous:
```
JesseLivezey: @nouiz, this is my attempted at testing the non-contiguous case. Let me know if there is a better way. Added new test that uses this flag.

JesseLivezey:
@nouiz, tests are passing. I had one more question about some of the op's methods and it would be great if you could look at the contiguous changes I made. Let me know if there is anything else needed.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,636 @@
...
+    def __str__(self):
+        return '%s{%s, %s}' % (
+            self.__class__.__name__,
+            self.border_mode,
+            str(self.subsample))
+
+    def flops(self, inp, outp):
+        """ Useful with the hack in profilemode to print the MFlops"""
+        # if the output shape is correct, then this gives the correct
+        # flops for any direction, sampling, padding, and border mode
```
f0k: Something with this method was wrong for one of the gradients. I think the assertion fails, but I can't find the issue or mailing list post, and I don't know when it's being called at all, maybe some special profile mode? If you're well in the code now (it's been some time since I wrote it), you might want to have a look. If you find what's wrong, this would need to be fixed in the GPU version as well. The idea was to have a single implementation for all three Ops (forward, GradI and GradW), and maybe it's enough to just remove the assertion, but maybe something needs to be flipped...

nouiz: This is only needed to have theano profile compute the Gf/s and show it. If it don't work and you don't know how to fix it, just delete it unless someone else can look at it. I'm pretty busy right now.

theano/tensor/nnet/corr.py
nouiz: They are needed to compile with blas correctly. So keep them.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -0,0 +1,376 @@
...
+                bottom_diff + (*bottom)[i]->offset(n));
+          }
+        }
+        */
+    }
+    // Free temporary columns
+    Py_DECREF(col);
+    if (!bottom_contiguous) {
+	Py_DECREF(bottom);
+    }
```
nouiz: Why did you do it like this? Why you didn't used the numpy function I gave you? I think this work, but it make the code more complicated.

Otherwise the part that make the contig version and decref seem good as the test. I need to leave, was there other think missing? If not, I'll try to do a global last review and run tests locally next week and merge this.

thanks

theano/tensor/nnet/corr.py
JesseLivezey: OK, I'll remove it for now. I can look into it more later.

theano/tensor/nnet/corr_gemm.c
JesseLivezey: Maybe I didn't understand your suggestion. There are two ways I could imagine doing this. The first way is to have
```
    if (!PyArray_IS_C_CONTIGUOUS(bottom))
    {
        bottom_contiguous = 0;
        bottom = PyArray_GETCONTIGUOUS(bottom);
    }
```
at the beginning and 
```
    if (!bottom_contiguous) {
        Py_DECREF(bottom);
    }
```
at the end. (I thought this is what you suggested)

The other way is to just always call
```
bottom = PyArray_GETCONTIGUOUS(bottom);
```
at the beggining and then always call
```
Py_DECREF(bottom);
```
at the end.

JesseLivezey:
@nouiz, I've removed the flops method for now and I've changed the contiguous checks to the simpler way (I think this is what you are suggesting).

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -0,0 +1,313 @@
...
+        self.validate((3, 2, 8, 8), (4, 2, 5, 5), 'full', verify_grad=False)
+        self.validate((3, 2, 7, 5), (5, 2, 2, 3), 'full')
+        self.validate((1, 10, 213, 129), (46, 10, 212, 1), 'valid',
+                      verify_grad=False)
+
+    def test_img_kernel_same_shape(self):
+        self.validate((3, 2, 3, 3), (4, 2, 3, 3), 'full')
+        self.validate((3, 2, 3, 3), (4, 2, 3, 3), 'valid')
+
+    @attr('slow')
```
JesseLivezey: @nouiz, should I remove the 'slow'? I'm not sure what this does. Subsampling with this implementation should be faster.

nouiz: Can you find that test? If it take less then 1s you can remove the slow,
otherwise keep it.

nouiz: I mean, can you time that test.

JesseLivezey: This test takes less than 1 second (~.6s). The three shape inference tests take about 1.5s each. Should I add the 'slow' attribute to them?

nouiz: The slow attribute is per test. As each test take less then 1s do not add it.

thanks

nouiz:
Thanks.

Personnaly, I think we can merge it. @lamblin do you want to make another review pass?

Maybe it would be good to add in the doc that currently it is not moved to the GPU but we will change that with the new interface? Just to don't have people be surprised by this until we finish the new interface.

Also, I have an idea to make it faster when the inputs aren't c contiguous. Are you interrested in that? I think that if the input have the inner dimensions c contiguous, but not the outer dims, we are don't need the copy (ex: a[:2,:]) and that this probably won't cause slowdown. But I'm not sure of that. Did you think of this? I don't want to delay this PR for that.

thanks

theano/tensor/nnet/corr_gemm.c
```diff
@@ -0,0 +1,357 @@
...
+                top_diff + top[i]->offset(n) + top_offset * g, N_,
+                1.0,
+                weight_diff + weight_offset * g, K_);
+          }
+        }
+        */
+    }
+    else if (direction == 2) {  // backprop wrt. inputs
+        output = bottom;
+        PyArray_FILLWBYTE(bottom, 0);
```
f0k: What is that needed for? You're overwriting everything with the col2im calls below, aren't you? It's not needed in the GPU version either: https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/corr_gemm.cu#L416

f0k:
> Also, I have an idea to make it faster when the inputs aren't c contiguous. Are you interrested in that? I think that if the input have the inner dimensions c contiguous, but not the outer dims, we are don't need the copy (ex: a[:2,:]) and that this probably won't cause slowdown. But I'm not sure of that.

As far as I see, the batch dimension for `bottom` (the input image) and `top` (the output image) can have arbitrary stride, but I think everything else needs to be contiguous (otherwise im2col and col2im need to be stride-aware, and that might slow things down).

theano/tensor/nnet/corr_gemm.c
JesseLivezey: It's mostly just a result of the caffe col2im and the available calls in the numpy C-API.

The ```col2im_cpu``` function that caffe uses calls ```caffe_set``` on one slice of the output array:
https://github.com/BVLC/caffe/blob/4f8775c1e88604879edef0698acab7b0d8a2ab43/src/caffe/util/im2col.cpp#L53
I looked through the numpy C-API and there didn't seem to be an equivalent call.

And in the GPU version you worked on, the result at each output location is accumulated in each thread (```val```) and then the value in the output array is overwritten. To do this on CPU would require allocating an extra array and accumulating into it, which seems more complicated.

Since col2im won't be used outside of this file, setting the array to zero outside of col2im seemed reasonable. I'll add a comment though so it's clear why that is happening.

JesseLivezey:
@nouiz @f0k , the idea for arrays that are not c contiguous in the batch dimension seems reasonable. Maybe it would be good to do that in a separate PR? I can try it out after.

theano/tensor/nnet/corr_gemm.c
f0k: Ah, all right, thanks for the explanation! `PyArray_FILLWBYTE` seems to be the best option then. It requires `bottom` to be contiguous, though -- if we want to relax that, we should move it into `col2im`. `PyArray_FILLWBYTE` just calls `memset` under the hood, so you could do that in `col2im`. It might even be a little faster due to better locality.

f0k:
> Maybe it would be good to do that in a separate PR?

Yes, I guess that's something to tackle separately.

JesseLivezey:
@nouiz, one test is failing, but I don't think I've caused it.

nouiz:
It was a timing test. I restarted it to have a clean history.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,621 @@
...
+
+class BaseCorrMM(gof.Op):
+    """Base class for `CorrMM`, `CorrMM_gradWeights` and
+    `CorrMM_gradInputs`. Cannot be used directly.
+
+    :param border_mode: one of 'valid', 'full', 'half'; additionally, the
+        padding size could be directly specified by an integer or a pair of
+        integers
+    :param subsample: perform subsampling of the output (default: (1, 1))
+    :param pad: *deprecated*, now you should always use border_mode
```
lamblin: I think there is no need to have a parameter that is deprecated from the start, so I think we should remove it.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,621 @@
...
+    __props__ = ('border_mode', 'subsample')
+
+    def __init__(self, border_mode="valid", subsample=(1, 1), pad=(0, 0)):
+        if pad != (0, 0):
+            _logger.warning(
+                'do not use pad for BaseCorrMM; please set padding in '
+                'border_mode parameter, see the docstring for more details')
+            if border_mode != "valid":
+                raise ValueError("border_mode must be 'valid'")
+            border_mode = pad
```
lamblin: So that block can be removed as well.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,621 @@
...
+        if isinstance(border_mode, int):
+            border_mode = (border_mode, border_mode)
+        if isinstance(border_mode, tuple):
+            pad_h, pad_w = map(int, border_mode)
```
lamblin: We should probably check that the padding passed by the user is >= 0 here, otherwise a user could try passing (-1, -1), expecting the image to be cropped by 1 pixel on each side, but it would actually mean "half" convolution.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,621 @@
...
+        Should be a tuple with 2 elements.
+        `(sv, sh)` is equivalent to `CorrMM(...)(...)[:,:,::sv, ::sh]`,
+        but faster.
+        Set to `(1, 1)` to disable subsampling.
+    :param pad: the width of a border of implicit zeros to pad the input
+        image with. Should be a tuple with 2 elements giving the numbers of
+        rows and columns to pad on each side, or "half" to set the padding
+        to `(kernel_rows // 2, kernel_columns // 2)`, or "full" to set the
+        padding to `(kernel_rows - 1, kernel_columns - 1)` at runtime.
+        Set to `(0, 0)` to disable padding.
```
lamblin: So, is the use of `pad` actually encouraged in the derived class? In that case, the warning would still be raised in the base constructor, right?

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -0,0 +1,315 @@
...
+                      verify_grad=False)
+
+    def test_img_kernel_same_shape(self):
+        self.validate((3, 2, 3, 3), (4, 2, 3, 3), 'full')
+        self.validate((3, 2, 3, 3), (4, 2, 3, 3), 'valid')
+
+    def test_subsample(self):
+        """
+        Tests correlation where subsampling != (1,1)
+        TODO
```
lamblin: It looks like that "TODO" has been done.

theano/tensor/nnet/corr.py
```diff
@@ -0,0 +1,621 @@
...
+
+    :note: Currently, the Op requires the inputs, filters and outputs to be
+        C-contiguous.
+
+    :note: You can either enable the Theano flag `optimizer_including=conv_gemm`
+        to automatically replace all convolution operations with `CorrMM`
```
lamblin: Is part actually correct? I don't think we have the optimizations to actually insert these Ops automatically yet.

lamblin:
Thanks a lot for the work! I'm sorry I could not provide more feedback earlier.

lamblin:
Also, I just realized that there do not seem to be any tests for modes other than "valid" and "full", in particular for padding, although it is advertised as working in the docstring.
This is probably why the warning about "do not use pad" does not get displayed in travis's log.

theano/tensor/nnet/corr.py
f0k: No, the docstring is just plain wrong. In e8fe4ad, somebody changed the GPU version's base class without updating the docstrings. In fbf3dcd, I noticed and fixed it for the GPU version. The updated docstring should be copied over to the CPU version as well.

theano/tensor/nnet/corr.py
f0k: Good idea. This holds for the GPU version as well. It would be cool to support the string shortcuts within the tuple, though, so you could also say `border_mode=("valid", "full")` to have a valid convolution in one direction, but a full convolution in the other. The C code would support this, as far as I remember, the Python code would just need to translate the strings to `-1` or `-2`.

JesseLivezey:
@nouiz @lamblin @f0k I've changed things from the last code review. As a reminder, they were

- Remove ```pad``` argument and update documentations
- Check that ```border_mode``` is non-negative if it's a tuple of ints.
- Added tests for ```border_mode == 'half'``` and tuples of ints.

Tests are passing now. If everything looks okay, I can rebase and squash into one commit.

lamblin:
Except for those comments, it looks good to me, thanks!

JesseLivezey:
@lamblin made the suggested changes. If things look good, I can rebase and squash.

lamblin:
It looks great, thanks a lot! I'll let Travis finish, and maybe @nouiz or @f0k has some additional comments, but as far as I'm concerned, you can go ahead and rebase/squash.

JesseLivezey:
@nouiz @lamblin I rebased and squashed commits. Tests passed.

lamblin:
Great, thanks a lot!

PR #3289, opened 2015-08-18
Disable the optimization to use DnnSoftmaxGrad for cuDNN v3 since it …

abergeron:
…is broken for (n, c 1, 1) shapes at the moment.


lamblin:
Travis will not help, so I'm merging.

PR #3290, opened 2015-08-18
Prod dimshuffle opt

mohammadpz:
I made a new PR regarding #399 and #3149 .
@carriepl , I need to talk to you once more to make sure that everything is ok and tests are expressive enough.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4130,39 +4130,40 @@ def local_elemwise_sub_zeros(node):
 
 @register_canonicalize
 @register_specialize
-@gof.local_optimizer([T.Sum])
-def local_sum_div_dimshuffle(node):
+@gof.local_optimizer([T.Sum, T.elemwise.Prod])
+def local_sum_prod_div_dimshuffle(node):
     """
     sum(a / dimshuffle{...}(b), axis=l) -> sum(a, axis={...}) / b,
-    if dimension l of the DimShuffle is 'x'.
+    if dimension l of the DimShuffle is 'x'
 
+    or
+
+    prod(a / dimshuffle{...}(b), axis=l) ->
+    prod(a, axis={...}) / b ** a.shape[l],
+    if dimension l of the DimShuffle is 'x'
     """
-    # TODO: extend it to product, and quotient of products
 
     # It does not make much sense now to extend it to the case where the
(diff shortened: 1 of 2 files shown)


carriepl:
There are currently conflicts so this PR will require a rebase.

carriepl:
Also, according to Travis, some tests are failing with floatX=float32

theano/tensor/opt.py
```diff
@@ -4076,37 +4079,49 @@ def local_sum_div_dimshuffle(node):
                                      'To disable this warning, set the Theano'
                                      ' flag warn.sum_div_dimshuffle_bug to'
                                      ' False.')
+                        break
```
carriepl: The change to this part will change the behavior in some cases

theano/tensor/tests/test_opt.py
```diff
@@ -4987,6 +4988,77 @@ def test_local_sum_div_dimshuffle(self):
...
+        b_val = rng.randn(2).astype(config.floatX)
+        c_val = rng.randn(2, 2, 2).astype(config.floatX)
+        d_val = numpy.asarray(rng.randn(), config.floatX)
+
+        mode_with_opt = copy.copy(theano.compile.mode.get_default_mode())
+        mode_without_opt = copy.copy(theano.compile.mode.get_default_mode())
+        mode_with_opt._optimizer = mode_with_opt._optimizer.including(
+            'local_sum_prod_div_dimshuffle')
+        mode_without_opt._optimizer = mode_without_opt._optimizer.excluding(
+            'local_sum_prod_div_dimshuffle')
```
mohammadpz: @carriepl , It seems that in both cases optimization is applied. Am I using ```including``` and ```excluding``` in a wrong way?

carriepl:
The 'logical' tests currently fail in FAST_COMPILE. This is because, in this mode, elemwise fusion is not activated, hence there will be no composite elemwise. For the tests to pass, you can do one of the following : 
* always activate elemwise fusion in your mode_with_opt compilation mode
* always deactivate elemwise fusion and adapt expected outer operator to reflect that
* do nothing about the elemwise fusion and always accept what should be the outer operator with and without the fusion.

carriepl:
Looks good. Thanks.

mohammadpz:
@carriepl , Thanks for your kind assistance :+1:

PR #3304, opened 2015-08-18
Add some minimal tests for cudnn softmax.

abergeron:


theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -489,6 +489,25 @@ def test_log_softmax():
...
+    softmax_op = dnn.GpuDnnSoftmax('bc01', 'accurate', 'channel')
+
+    x_val = numpy.random.normal(0, 1, (3, 4, 1, 1)).astype('float32')
+
+    utt.verify_grad(softmax_op, [x_val])
+
+
+def test_dnn_softmax_grad_opt():
+    from theano.tensor.nnet import softmax_op
+    from theano.gradient import verify_grad
```
lamblin: flake8 says: `F401 'verify_grad' imported but unused`

abergeron:
Fixed

nouiz:
Why this is useful? This is already tested in test_nnet.py (it test our softmax and cudnn one).

nouiz:
Maybe we can move the cudnn one to test_dnn.py file

abergeron:
Because they are not on test_dnn.py I didn't find them.  Also, why haven't these tests caught that GpuDnnSoftmaxGrad (x, y, 1, 1) doesn't work?

nouiz:
I think it would be great to move the tests to test_dnn.py. It probably didn't tested that shape. Adding it would be good.

abergeron:
I've moved and consolidated the tests.

nouiz:
tests fail.

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -450,43 +452,132 @@ def test_pooling_opt():
...
+    def test_dnn_softmax_grad(self):
+        softmax_op = dnn.GpuDnnSoftmax('bc01', 'accurate', 'channel')
+
+        x_val = numpy.random.normal(0, 1, (3, 4, 2, 5)).astype('float32')
+        x_val2 = numpy.random.normal(0, 1, (3, 4, 1, 1)).astype('float32')
+
+        utt.verify_grad(softmax_op, [x_val])
+
+        # Gradient is broken for (n, c, 1, 1) in v3 rc1
+        if cuda.dnn.version() == (2000, 2000):
```
nouiz: I would do 

... != (3000, 3000), next version should work, so we don't need to update.

abergeron: Fixed.

abergeron:
What fails exactly in the tests?  I haven't seen any failures on my end.

PR #3305, opened 2015-08-18
Do not read infinite value from output in advanced setsubtensor1

### LINKED ISSUE #3291: bug with T.set_subtensor()
```python
def rec_area(R):
    '''
    x_min, x_max, y_min, y_max
    '''
    return (R[:,1] - R[:,0]) * (R[:,3] - R[:,2])

def max_mat(M1, M2):
    return M1 * (M1 > M2) + M2 * (M2 >= M1)

def min_mat(M1, M2):
    return M1 * (M1 < M2) + M2 * ( M2 <= M1)

def jaccard_box_score(R1, R2):
    '''
    x_min, x_max, y_min, y_max
    '''
    intercept = max_mat(T.zeros((R1.shape[0],)), min_mat(R1[:,1], R2[:,1]) - max_mat(R1[:,0], R2[:,0])) * \
                max_mat(T.zeros((R1.shape[0],)), min_mat(R1[:,3], R2[:,3]) - max_mat(R1[:,2], R2[:,2]))
    A1 = rec_area(R1)
    A2 = rec_area(R2)
    denom = A1 + A2 - intercept

    X = intercept / denom
    Z = T.set_subtensor(X[T.eq(denom,0).nonzero()], -1)

    return Z

def testing(R1, R2):
    y = R1 / R2
    X = T.set_subtensor(y[T.eq(R2,0).nonzero()], -1)
    return X

if __name__ == '__main__':
    R1 = T.matrix()
    R2 = T.matrix()
    f = theano.function([R1, R2], jaccard_box_score(R1, R2)) 
    print f(np.asarray([[1,2,2,3],[0,0,0,0],[1,2,3,4]]), np.asarray([[2,3,2,3],[0,0,0,0],[2,0,3,0]])) # returns nan is not set to -1

   f = theano.function([R1, R2], testing(R1, R2)) # returns nan is set to -1
```
The `nan` or `inf` produced by `testing` when the denominator is 0 is set successfully to -1 with using `X = T.set_subtensor(y[T.eq(R2,0).nonzero()], -1)`, however doing exactly the same thing in `jaccard_box_score` with `T.set_subtensor(X[T.eq(denom,0).nonzero()], -1)`, the `nan` remains as nan. I wonder why.

lamblin:
This fixes #3291 after a clean-up of the existing cutils_ext.
Is there a way to force its recompilation?

### DIFF
--- a/theano/gof/cutils.py
+++ b/theano/gof/cutils.py
@@ -41,13 +41,16 @@ def compile_cutils_code():
     #endif
     """
 
-    floatadd = ("((%(type)s*)mit->dataptr)[0] = inc_or_set * "
-                "((%(type)s*)mit->dataptr)[0] + ((%(type)s*)it->dataptr)[0];")
+    floatadd = ("((%(type)s*)mit->dataptr)[0] = "
+                "(inc_or_set ? ((%(type)s*)mit->dataptr)[0] : 0)"
+                " + ((%(type)s*)it->dataptr)[0];")
     complexadd = """
-    ((%(type)s*)mit->dataptr)[0].real = inc_or_set *
-       ((%(type)s*)mit->dataptr)[0].real + ((%(type)s*)it->dataptr)[0].real;
-    ((%(type)s*)mit->dataptr)[0].imag = inc_or_set *
-       ((%(type)s*)mit->dataptr)[0].imag + ((%(type)s*)it->dataptr)[0].imag;
+    ((%(type)s*)mit->dataptr)[0].real =
+        (inc_or_set ? ((%(type)s*)mit->dataptr)[0].real : 0)
+        + ((%(type)s*)it->dataptr)[0].real;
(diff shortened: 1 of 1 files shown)


abergeron:
Please add auto recompile for cutils_ext.  Otherwise we will probably have lots of problems with this in the future.

On another note why is that not there yet?

lamblin:
cutils_ext was not expected to change at all.
In fact, having those functions here is kind of a hack, since we did not have any other compiled module available from all Python ops (except for cuda_ndarray).

abergeron:
Even if only run_cthunk() was expected to be there it might need some fixing at some point.

Also can those map_increment and inplace_increment functions be moved into the op that uses them.  Or are there lots of ops that do so?

nouiz:
one or both of map_increment and inplace_increment are used by the perform of some op. That is why it is not here.

nouiz:
This also need a test

PR #3307, opened 2015-08-19
Correct config.lib.cnmem name in documentation

f0k:
It's a trivial change, but the CNMeM configuration option was named `lib.cnmem` instead of `config.lib.cnmem` in the documentation. As I hadn't heard about the `lib` prefix before, I thought it should mean `config.cnmem` and thus my first attempt was to use `THEANO_FLAGS=cnmem=.3`, which didn't work. This PR makes it unambiguous.


lamblin:
Thanks!

PR #3308, opened 2015-08-19
Update sandbox/cuda/dnn_base.c to work on Windows

ivdorelian:
Allocating the strides array dynamically lets it also compile on Windows. See https://groups.google.com/forum/#!topic/theano-users/fex4OpL9hB0 for code that errors on Windows without this.

### DIFF
--- a/theano/sandbox/cuda/dnn_base.c
+++ b/theano/sandbox/cuda/dnn_base.c
@@ -7,8 +7,9 @@ c_set_tensorNd(CudaNdarray *var, cudnnTensorDescriptor_t desc) {
 
 
   int dim = CudaNdarray_NDIM(var);
-  int strides[dim];
+  int *strides = (int *)malloc(dim * sizeof(int));
   int default_str = 1;
+  int return_value = 0;
 
   for (int i = dim-1; i >= 0; i--)
   {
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/dnn_base.c
```diff
@@ -7,8 +7,9 @@ c_set_tensorNd(CudaNdarray *var, cudnnTensorDescriptor_t desc) {
 
 
   int dim = CudaNdarray_NDIM(var);
-  int strides[dim];
+  int *strides = (int *)malloc(dim * sizeof(int));
```
abergeron: You need to check that the allocation worked.

ivdorelian: @abergeron ok, is there a pattern for what should happen if it fails? Just return -1?

abergeron: Set a python exception (probably MemoryError) and return -1.

ivdorelian: @abergeron done, please let me know if there's more to be done.

theano/sandbox/cuda/dnn_base.c
```diff
@@ -5,31 +5,44 @@ static cudnnHandle_t _handle = NULL;
 static int
 c_set_tensorNd(CudaNdarray *var, cudnnTensorDescriptor_t desc) {
 
-
   int dim = CudaNdarray_NDIM(var);
-  int strides[dim];
+  int *strides = (int *)malloc(dim * sizeof(int));
```
nouiz: Can we use calloc on windows? I think it is faster on linux.

Otherwise, thanks.

ivdorelian: I think it works. I'll make sure and change it this weekend when I'm back
home.

PR #3310, opened 2015-08-19
props to theano/gof/op.py

harlouci:
Not yet merged, seemingly.


abergeron:
This add \__props__ to "superclasses" which will make classes that inherit from those have an empty \__props__ attribute.  This is a bad idea.

PR #3311, opened 2015-08-20
Sparse block dot

bouthilx:
Cherry-picked commits from #3109 + an additional commit to fix optimization issues.


abergeron:
You have some flake8 errors + your tests fail in FAST_COMPILE.

bouthilx:
@abergeron I amended last commit. I do not optimize inplace for FAST_COMPILE so I modified the test accordingly.

bouthilx:
@abergeron Ready for final review/merge :)

theano/sandbox/blocksparse.py
```diff
@@ -0,0 +1,234 @@
...
+        for b in range(x.shape[0]):
+            for i in range(xIdx.shape[1]):
+                for j in range(yIdx.shape[1]):
+                    o[xIdx[b, i], yIdx[b, j]] += numpy.outer(x[b, i],
+                                                             y[b, j, :])
+        out_[0][0] = o
+
+    def grad(self, inputs, output_gradients):
+        raise NotImplementedError("SparseBlockOuter has no gradient "
+                                  "implemented")
```
abergeron: If there is no gradient, just not having the method at all is better.

theano/sandbox/cuda/opt.py
```diff
@@ -741,11 +749,11 @@ def local_gpu_gemm(node):
...
         if x_on_gpu or y_on_gpu or z_on_gpu:
-            return [host_from_gpu(gpu_gemm_no_inplace(as_cuda_ndarray_variable(z),
-                                                 a,
-                                                 as_cuda_ndarray_variable(x),
-                                                 as_cuda_ndarray_variable(y),
-                                                 b))]
+            return [host_from_gpu(gpu_gemm_no_inplace(gpu_from_host(z),
+                                                      a,
+                                                      gpu_from_host(x),
+                                                      gpu_from_host(y),
```
abergeron: This is a bad change.  as_cuda_ndarray_variable() will work on stuff that is already on the gpu whereas gpu_from_host() will not.

Also why do you need to change this?

theano/sandbox/cuda/opt.py
```diff
@@ -2420,4 +2429,131 @@ def local_gpu_allocempty(node):
...
+    """
+    if isinstance(node.op, GpuSparseBlockOuter) and not node.op.inplace:
+        new_node = gpu_sparse_block_outer_inplace(*node.inputs)
+        return [new_node]
+    return False
+compile.optdb.register('local_inplace_gpu_sparse_block_outer',
+                       TopoOptimizer(
+                           local_inplace_gpu_sparse_block_outer,
+                           failure_callback=TopoOptimizer.warn_inplace),
+                       60, 'fast_run', 'inplace', 'gpu')  # DEBUG
```
abergeron: You can probably bundle up your two inplace optimizers into one registration.

theano/sandbox/opt.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    if isinstance(node.op, SparseBlockOuter) and not node.op.inplace:
+        new_node = sparse_block_outer_inplace(*node.inputs)
+        return [new_node]
+    return False
+compile.optdb.register('local_inplace_sparse_block_outer',
+                       gof.TopoOptimizer(
+                           local_inplace_sparse_block_outer,
+                           failure_callback=gof.TopoOptimizer.warn_inplace),
+                       60, 'fast_run', 'inplace')  # DEBUG
```
abergeron: Is this file imported somewhere?

theano/sandbox/tests/test_multinomial.py
```diff
@@ -4,7 +4,7 @@
 
 import theano
 from theano import config, function, tensor
-from . import multinomial
+from theano.sandbox import multinomial
```
abergeron: You can also use `from .. import multinomial`

abergeron:
Looks good, all my comments are minor, except for the one about as_cudandarray_variable() which might break something.

theano/sandbox/cuda/opt.py
bouthilx: Oh my! That was only intended to be pep8 fixes. It's probably due to a bad merge I made where I didn't realize new master version was using as_cuda_ndarray_variable and not gpu_from_host. I'll fix this!

theano/sandbox/opt.py
bouthilx: It is in sandbox/__init__.py 

I could import it in sandbox/blocksparse.py to avoid optimizations when there is no SparseBlock{Gemv,Outer} but then I get a nasty import cycle. >_<

theano/sandbox/tests/test_multinomial.py
bouthilx: Is it better? I wanted to get rid of relative paths.

theano/sandbox/cuda/opt.py
bouthilx: You mean by placing both local optimizations inside the same optimizer? Like SeqOptimizer rather than TopoOptimizer?

theano/sandbox/tests/test_multinomial.py
abergeron: I happen to like relative paths.  But it's fine as it is.

theano/sandbox/cuda/opt.py
abergeron: Yes.  Or perhaps a Topo with two optimizers inside.  I'm not sure.

theano/sandbox/opt.py
abergeron: Ok.  It's fine like this.

lucasb-eyer:
I believe adding the `__init__.py` file in the `sandbox` folder as part of 5b604a4 has the side-effect of breaking existing code that directly used anything other than `sandbox.opt`, such as writing `theano.sandbox.rng_mrg.MRG_RandomStreams()` . I don't think it's too bad, just putting the information out there.

bouthilx:
@lucasb-eyer Thank you for letting us know. I will fix this today.

abergeron:
This PR did not add the file, it was there and empty before.  This is required by python to have imports from that submodule working at all.

lucasb-eyer:
Hmm, you are right I misinterpreted github's display. Then I'm not sure when exactly things like `theano.sandbox.rng_mrg.MRG_RandomStreams()` stopped working. But I'd say just ignore my comment, since everywhere in the documentation `from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams` is recommended, which always has and still does work.

PR #3312, opened 2015-08-20
remove 'merge3'

t13m:
For the user case I have (https://gist.github.com/t13m/64383eec4aead5e1aea8#file-merge-assert , line 148), optimizer `merge3` is useless (all its attempts to merge fail). 
So it is removed here.

### DIFF
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -226,9 +226,10 @@ optdb.register('add_no_output_from_inplace', AddNoOutputFromInplace(),
 optdb.register('add_destroy_handler', AddDestroyHandler(),
                49.5, 'fast_run', 'inplace')
 
+# 'merge3' seems useless
 # final pass just to make sure
-optdb.register('merge3', gof.MergeOptimizer(),
-               100, 'fast_run', 'merge')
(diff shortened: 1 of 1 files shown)


theano/compile/mode.py
```diff
@@ -226,9 +226,10 @@ def apply(self, fgraph):
 optdb.register('add_destroy_handler', AddDestroyHandler(),
                49.5, 'fast_run', 'inplace')
 
+# 'merge3' seems useless
```
nouiz: Can you also run all Theano tests with profile=True to be sure there isn't corner case where this is not needed?

thanks

nouiz:
I found a profile with this:

   16.164870s - ('merge3', 'MergeOptimizer', 40) - 16.122s
     MergeOptimizer
       nb fail=    0 merged=   95 constant=   95
       time replace=16.16 validate=16.12 callback=16.14
       callbacks_time

So it mean we still need it. So I'll close this ticket. Thanks for testing this.

I'll make another ticket to optimize this new profile. Mostly, it is merge of constant, so we don't need to do a validate.

PR #3314, opened 2015-08-20
map_variables

cooijmanstim:
(Work in progress on #3177 , do not merge)

Now the replacement descends into scan nodes. For some reason that I don't understand, I need to keep track of those scan nodes that I have already replaced. Otherwise the local optimizer will be called again and again on the replaced node. Changing the order from out_to_in to in_to_out doesn't change that.

I'll sort out the mess of imports later.

(The other pull request was from my master and got contaminated with unrelated commits; I'll use this one from now on.)


carriepl:
If the other pull request is not used anymore, you should close it.

carriepl:
Also, it's good that it now works for scan nodes but it should also work for OpFromGraph nodes (which is just a way of bundling a computational graph into a single op

theano/gof/graph.py
```diff
@@ -816,6 +816,92 @@ def clone_get_equiv(inputs, outputs, copy_inputs_and_orphans=True, memo=None):
...
+    fg = FunctionGraph(inputs_, graphs, clone=False)
+
+    nodes_seen = set()
+
+    @local_optimizer(None)
+    def local_transform(node):
+        if node in nodes_seen:
+            return False
+        # FIXME: replacing inputs won't work because they are not
+        # outputs of any Apply node
```
carriepl: If you want to support input replacements, you can do it after the optimization with your 'local_transform', loop over the inputs of the optimized graph and get their replacements if applicable and then perform a clone with replacement.

carriepl:
Some tests should also be implemented to test this feature.

carriepl:
Can you add a test for an input replacement inside either a Scan or OpFromGraph? Once that's done, I'm fine with merging as long as Travis is happy.

carriepl:
And travis seems unhappy again  : 
```
ERROR: theano.gof.tests.test_graph.TestMapVariables.test_opfromgraph

    self.test(*self.arg)

  File "/home/travis/miniconda/envs/pyenv/lib/python3.3/site-packages/theano/gof/tests/test_graph.py", line 199, in test_opfromgraph

    for m, n in itertools.combinations(xrange(10), 2):

NameError: global name 'xrange' is not defined
```

carriepl:
Ok, I'll wait for that modification too. You know where to put it?

carriepl:
I guess putting it in the same place as 'clone' would probably make the most sense, given that both have similar uses.

carriepl:
Travis is complaining because of a syntax error

carriepl:
Why was this moved to scan_utils? Was there something preventing it from being in the same place as clone? Since this is useful for more than Scan, I'm not sure that files only related to Scan is the best place for it.

theano/scan_module/scan_utils.py
```diff
@@ -256,6 +256,208 @@ def clone(output,
...
+        if not constants and not foreign_inputs:
+            return new_graph
+
+        replacements = []
+
+        # constants just need to be replaced by copies that the inner
+        # `fg` can take ownership of
+        for input_ in constants:
+            new_input = input_.clone()
+            new_input.name = "%s_copiedd" % new_input.name
```
carriepl: 'copiedd'

cooijmanstim:
Yes, I moved it to scan_utils because there is a "clone" there.  This clone is the one that does replacement.  Both would be better moved elsewhere if you ask me.

I'll fix the syntax error and the typos when I get back.

carriepl:
I talked with @nouiz and, for this PR, map_variables can be left in scan_utils. It will probably be moved eventually but it's not obvious to him, at the moment, where it should go.

theano/scan_module/tests/test_scan_utils.py
```diff
@@ -11,3 +15,146 @@ def test_equal_compuations():
...
+            s2, = map_variables(self.replacer, [s])
+        except NotImplementedError as e:
+            e = sys.exc_info()[1]
+            assert("introduces shared variable" in str(e))
+            return
+
+        # test failed
+        return 0
+
+    def test_scan_with_shared_update3(self):
```
carriepl: 3? Is there supposed to be a test_scan_with_shared_update2 before this test?

theano/scan_module/tests/test_scan_utils.py
```diff
@@ -11,3 +15,146 @@ def test_equal_compuations():
...
+                           outputs_info=[numpy.array(0.)])
+        try:
+            s2, = map_variables(self.replacer, [s])
+        except NotImplementedError as e:
+            e = sys.exc_info()[1]
+            assert("introduces shared variable" in str(e))
+            return
+
+        # test failed
+        return 0
```
carriepl: You can replace, from the try to the return, by a call to self.assertRaises(NotImplementedError, map_variables, self.replacer, [s]).
For self.assertRaises to be defined, your test class must inherit from unittest.TestCase

This comment also applies to the other test that follows the same pattern.

carriepl:
Originally, I asked for a test adding inputs to either a Scan or OpFromGraph but given the complexity that this PR has reached, I would prefer if we had both. You've already made a test for Scan, can you add a small one for OpFromGraph? After this and my other comment, I think this will be all for this PR.

theano/scan_module/tests/test_scan_utils.py
cooijmanstim: Good call; yes there was but I merged it into test_scan_with_shared_update.

cooijmanstim:
It took some time to get the new test_opfromgraph to pass, but it does now that I added additional_inputs=[outer, shared] to the calls to map_variables().  It's interesting that I didn't have to do this for the scan tests.

carriepl:
Travis is not done running but it already spotted a flake8 issue : 
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/scan_module/tests/test_scan_utils.py:1:1: F401 'sys' imported but unused

carriepl:
Good work. This will probably be expended and improved as it becomes more heavily used. For now, this seems solid and there's a good list of tests. I think this is good for merge.

PR #3318, opened 2015-08-21
Numpy-like interface for stack.

### LINKED ISSUE #3182: wrap numpy.stack
https://github.com/numpy/numpy/pull/5605 add it, we should make it available as a user interface and use our current op.

Thrandis:
Fixes #3182

### DIFF
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -583,6 +583,20 @@ dimensions, see :meth:`_tensor_py_operators.dimshuffle`.
     :type n_ones: int
     :type n_ones: number of dimension to be added to `x`
 
+
+.. function:: shape_padaxis(t, axis)
+
+    Reshape `t` by adding 1 at the dimension `axis`. Note that this new
+    dimension will be broadcastable. To make it non-broadcastable
+    see the :func:`unbroadcast`.
+
+    :type x: any TensorVariable (or compatible)
+    :param x: variable to be reshaped
+
+    :type axis: int
+    :param  axis: axis where to add the new dimension to `x`
+
+
 .. autofunction:: unbroadcast(x, *axes)
 
 .. autofunction:: addbroadcast(x, *axes)
@@ -677,6 +691,26 @@ Creating Tensor
     :returns: A tensor of same shape as `x` that is filled with 0s everywhere
               except for the main diagonal, whose values are equal to one. The output
               will have same dtype as `x`.
+
+
+.. function:: stack(tensors, axis=0)
+
+    Warning: The interface stack(*tensors) is deprecated!
+
+    Return a Tensor representing for the arguments all stacked up into a single Tensor.
+    (of 1 rank greater).
+
+    :param tensors: a list or a tuple of one or more tensors of the same rank.
+    :param axis: the axis along which the tensors will be stacked.
+    :returns: A tensor such that rval[0] == tensors[0], rval[1] == tensors[1], etc.
+
+    >>> x0 = T.scalar()
+    >>> x1 = T.scalar()
+    >>> x2 = T.scalar()
+    >>> x = T.stack(x0, x1, x2)
+    >>> x.ndim # x is a vector of length 3.
+    1
+
 
 .. function:: stack(*tensors)
 
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -701,13 +701,13 @@ Creating Tensor
     (of 1 rank greater).
 
     :param tensors: a list or a tuple of one or more tensors of the same rank.
-    :param axis: the axis along which the tensors will be stacked.
+    :param axis: the axis along which the tensors will be stacked. Default value is 0.
     :returns: A tensor such that rval[0] == tensors[0], rval[1] == tensors[1], etc.
 
     >>> x0 = T.scalar()
     >>> x1 = T.scalar()
     >>> x2 = T.scalar()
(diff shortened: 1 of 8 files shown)


Thrandis:
@bouthilx : Sorry for the mess! Hope it is allright now!

theano/tensor/basic.py
```diff
@@ -3963,16 +3963,53 @@ def shape_padright(t, n_ones=1):
...
 def stack(*tensors):
     """Insert the arguments as slices into a tensor of 1 rank greater.
 
     The size in dimension 0 of the result will be equal to the number
     of tensors passed.
 
+    Note: The interface stack(*tensors) is deprecated, you should use
+    stack(tensors, axis=0) insted.
+
+    :Parameters:
```
bouthilx: Parameters docstring format is somehow inconsistant in Theano. @lamblin Should we use 
```
"""
:Parameters: 
- `parameter_name` : type
    description
"""
```

or this

```
"""
Parameters
----------
parameter_name : type
    description
"""
```

lamblin: Ideally, the second one, but I'm not sure if it is already supported.
@abergeron, what is the status?

bouthilx:
@Thrandis Arnaud confirmed new docstring format is supported. You can modify to this style

```python
"""
Parameters
----------
parameter_name : type
    description
"""
```

bouthilx:
Oops. Looking at the new docstring I realized that using the function as  
```python 
stack(tensors=[A, B, C], axis=1)
``` 
would not work. We need to handle **kwargs during the deprecation phase. Sorry @Thrandis ! Please add a test case to verify this function signature.

Thrandis:
@bouthilx there you go!

bouthilx:
@Thrandis What if you add ```warnings.simplefilter('always', DeprecationWarning)``` at the beginning of your test? (Make sure you sure you use ```DeprecationWarning``` when you raise the message)

Thrandis:
I was using warnings.simplefilter('always')
Le 4 sept. 2015 11:50 AM, "Xavier Bouthillier" <notifications@github.com> a
écrit :

bouthilx:
Ok, try with DeprecationWarning. For me it didn't work without it.

Thrandis:
I'll try that just after Aaron's class

Thrandis:
@bouthilx I made the changes. 
The weird thing is that if the warning has been raised in the code before, then it wont't be raised again (even with `warnings.simplefilter('always', DeprecationWarning`). So I switched to the new interface for the whole test script.

bouthilx:
oh! You should change the stack() calls to the new interface everywhere.

Thrandis:
@bouthilx ahah damn you're right! I'll make those changes!

Thrandis:
@bouthilx 
I added stack([a, b], axis=1) and made the change in the whole theano code!
Last thing to do is to check for those negative axis!

Thrandis:
@bouthilx how should I handle an out-of-bounds case?

Right now, if the tensor shape is let's say 3, and we want to stack them on the axis 5, it will add a new dimension and stack them on that dimension (so the tensor will be of size 4 instead of 5), because `list.insert()` allows it.

Shall I test that `abs(axis)` is not bigger than `tensors[0].ndim + 1` and raise an error if it is?

bouthilx:
You should test it like in numpy ([check here](https://github.com/numpy/numpy/blob/master/numpy/core/shape_base.py#L342)).

Yep if you have tensors of dimension 3, axis=5 is out of bound.

Thrandis:
@bouthilx added negative axis. Let me know if there is other modifications to do!

Thrandis:
@bouthilx there you go :-)

doc/library/tensor/basic.txt
```diff
@@ -678,6 +692,26 @@ Creating Tensor
...
+    (of 1 rank greater).
+
+    :param tensors: a list or a tuple of one or more tensors of the same rank.
+    :param axis: the axis along which the tensors will be stacked.
+    :returns: A tensor such that rval[0] == tensors[0], rval[1] == tensors[1], etc.
+
+    >>> x0 = T.scalar()
+    >>> x1 = T.scalar()
+    >>> x2 = T.scalar()
+    >>> x = T.stack(x0, x1, x2)
```
bouthilx: Deprecated :)

theano/tensor/basic.py
```diff
@@ -3963,16 +3963,78 @@ def shape_padright(t, n_ones=1):
...
 
+    Note: The interface stack(*tensors) is deprecated, you should use
+    stack(tensors, axis=0) insted.
+
+    Parameters
+    ----------
+    tensors : list or tuple of tensors
+        A list of tensors to be stacked.
+    axis : int
+        The index of the new axis.
```
bouthilx: Default value is 0

bouthilx:
Sorry @Thrandis for the two late minor fixes.

bouthilx:
Thanks!

theano/tensor/basic.py
```diff
@@ -3963,16 +3963,78 @@ def shape_padright(t, n_ones=1):
 
 
 @constructor
-def stack(*tensors):
+def shape_padaxis(t, axis):
+    """Reshape `t` by adding 1 at the dimension `axis`.
```
nouiz: This need an example. It is not clear on which side of the axis the 1 is added. Both option are possible.

nouiz: This also need to be documented in the file doc/library/tensor/basic.txt after shape_padright

PR #3322, opened 2015-08-24
d3viz: Interactive visualization of Theano compute graphs

cangermueller:
`d3viz` extends Theano’s printing module
to interactively visualize compute graphs. Instead of creating a static
picture, it creates an HTML file, which can be opened with any
web-browser. `d3viz` allows
-  to zoom to different regions and to move graphs via drag and drop,
-  to position nodes both manually and automatically,
-  to retrieve additional information about nodes and edges such as
   their data type or definition in the source code,
-  to edit node labels,
-  to visualizing profiling information, and
-  to explore nested graphs such as OpFromGraph nodes.

This pull request includes the fully documented and tested `d3viz` module, as well as an user guide with examples.


benanne:
Awesome project :) kudos!

doc/library/d3viz/index.txt
```diff
@@ -0,0 +1,274 @@
...
+
+    pip install pydot
+
+Overview
+--------
+
+``d3viz`` extends Theano’s `printing
+module <http://deeplearning.net/software/theano/library/printing.html>`__
+to interactively visualize compute graphs. Instead of creating a static
+picture, it creates an HTML file, which can be opened with any
```
carriepl: Sorry to be picky about this but this should probably say "most recent [...]" instead of "any [...]". Using 'any' will lead some people to think they can just use IE6 and expect all of the Javascript and CSS to work flawlessly.

doc/library/d3viz/index.txt
```diff
@@ -0,0 +1,274 @@
...
+
+Different output formats
+------------------------
+
+Internally, ``d3viz`` represents a compute graph in the `Graphviz DOT
+language <http://www.graphviz.org/>`__, using the
+`pydot <https://pypi.python.org/pypi/pydot>`__ package, and defines a
+front-end based on the `d3.js <http://d3js.org/>`__ library to visualize
+it. However, any other Graphviz front-end can be used, which allows to
+export graps to different formats.
```
carriepl: graphs

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,351 @@
...
+                    edge_params['color'] = 'red'
+
+                edge_label = vparams['dtype']
+                if len(node.inputs) > 1:
+                    edge_label = str(id) + ' ' + edge_label
+                pdedge = pd.Edge(var_id, __node_id, label=edge_label,
+                                 **edge_params)
+                graph.add_edge(pdedge)
+
+            # Loop over ouput nodes
```
carriepl: output

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,351 @@
...
+                    pd_var = dict_to_pdnode(vparams)
+                    graph.add_node(pd_var)
+
+                    graph.add_edge(pd.Edge(__node_id, var_id,
+                                           label=vparams['dtype']))
+                elif var.name or not self.compact:
+                    graph.add_edge(pd.Edge(__node_id, var_id,
+                                           label=vparams['dtype']))
+
+            # Create sub-groph for OpFromGraph nodes
```
carriepl: sub-graph

theano/d3viz/html/template.html
```diff
@@ -0,0 +1,105 @@
...
+		
+		var forceLayout;
+		var isProfiled = false; // true if profiling information available
+		var useProfileColors = false;
+		var fixOnInit = true; // fix nodes on initialization
+		var maxProfilePer = 0;
+		var profileColors = ["#fff5f0","#fee0d2","#fcbba1","#fc9272","#fb6a4a","#ef3b2c","#cb181d","#a50f15"];
+		var pad = 10;
+		var isEditNode = false; // true if node is edited
+    	var menuItems = [
```
carriepl: indentation

carriepl:
Seems like what most of what Travis is complaining about is flake8 issues.

carriepl:
Would it be feasible to add at least one test where you compare the result of parsing a computation graph to a reference result?

carriepl:
Other than the points I raised, this looks pretty good. Since this is a pretty large pull request, however, I'd still like to have someone else take a quick pass at this. @lamblin or @abergeron, can one of you spare some time?

cangermueller:
@carriepl: I revisted the documentation as suggested, extended test case to compare d3viz output to reference output, and fixed Travis complaints.

carriepl:
The changes appear good but there are still a few flake8 errors to address.

theano/d3viz/d3viz.py
```diff
@@ -0,0 +1,99 @@
...
+
+Author: Christof Angermueller <cangermueller@gmail.com>
+"""
+
+import os
+import os.path as pt
+import shutil
+import re
+
+from formatting import PyDotFormatter
```
abergeron: Change this to `from .formatting import PyDotFormatter` for python 3 compat.

theano/d3viz/d3viz.py
```diff
@@ -0,0 +1,99 @@
...
+
+__path__ = pt.dirname(pt.realpath(__file__))
+
+
+def replace_patterns(x, replace):
+    """Replace patterns `replace` in x."""
+    for from_, to in replace.items():
```
abergeron: Use iteritems from six to iterate.

theano/d3viz/d3viz.py
```diff
@@ -0,0 +1,99 @@
+"""Dynamic visualization of Theano graphs.
+
+Author: Christof Angermueller <cangermueller@gmail.com>
+"""
+
+import os
+import os.path as pt
```
abergeron: Please don't rename like this it's just confusing and only saves two characters.

theano/d3viz/d3viz.py
```diff
@@ -0,0 +1,99 @@
...
+        '%% JS_DIR %%': pt.join(dst_deps, 'js'),
+        '%% CSS_DIR %%': pt.join(dst_deps, 'css'),
+        '%% DOT_GRAPH %%': pt.basename(dot_graph),
+    }
+    html = replace_patterns(template, replace)
+
+    # Write HTML file
+    if outfile is not None:
+        f = open(outfile, 'w')
+        f.write(html)
```
abergeron: This write could throw an exception.  Use this pattern to ensure the file is closed:

```
with open(outfile, 'w') as f:
    f.write(html)
```

theano/d3viz/d3viz.py
```diff
@@ -0,0 +1,99 @@
...
+    # Replace patterns in template
+    replace = {
+        '%% JS_DIR %%': pt.join(dst_deps, 'js'),
+        '%% CSS_DIR %%': pt.join(dst_deps, 'css'),
+        '%% DOT_GRAPH %%': pt.basename(dot_graph),
+    }
+    html = replace_patterns(template, replace)
+
+    # Write HTML file
+    if outfile is not None:
```
abergeron: This check is not useful since if outfile is None, it will crash at line 59.

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,343 @@
...
+    def __init__(self, compact=True):
+        """Converts compute to to `pydot` object.
+
+        :param compact: if True, will remove intermediate variables without
+                        name.
+        """
+
+        if not pydot_imported:
+            raise RuntimeError("Failed to import pydot. You must install pydot"
+                               " for `pydotprint` to work.")
```
abergeron: This will lead to a strange error fo people calling d3write() and getting an error about pydotprint().

theano/d3viz/d3viz.py
```diff
@@ -0,0 +1,99 @@
...
+    edited by selecting Edit from the context menu.
+
+    Input nodes are colored in green, output nodes in blue. Apply nodes are
+    ellipses, and colored depending on the type of operation they perform. Red
+    ellipses are transfers from/to the GPU (ops with names GpuFromHost,
+    HostFromGpu).
+
+    Edges are black by default. If a node returns a view of an
+    input, the input edge will be blue. If it returns a destroyed input, the
+    edge will be red.
```
abergeron: Docstring should be numpydoc style.  See https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,343 @@
...
+            return self.__add_node(node)
+
+    def __call__(self, fct, graph=None):
+        """Create pydot graph from function.
+
+        :param fct: a compiled Theano function, a Variable, an Apply or
+                    a list of Variable.
+        :param graph: `pydot` graph to which nodes are added. Creates new one
+                      if not given.
+        """
```
abergeron: Numpydoc style.

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,343 @@
...
+            __node_id = self.__node_id(node)
+            nparams['name'] = __node_id
+            nparams['label'] = apply_label(node)
+            nparams['profile'] = apply_profile(node, profile)
+            nparams['node_type'] = 'apply'
+            nparams['apply_op'] = nparams['label']
+            nparams['shape'] = self.shapes['apply']
+
+            use_color = None
+            for opName, color in self.apply_colors.items():
```
abergeron: iteritems from six

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,343 @@
...
+                    vparams['fillcolor'] = self.node_colors[
+                        vparams['node_type']]
+                    vparams['shape'] = self.shapes['input']
+                    pd_var = dict_to_pdnode(vparams)
+                    graph.add_node(pd_var)
+
+                edge_params = {}
+                if hasattr(node.op, 'view_map') and \
+                        id in reduce(list.__add__,
+                                     node.op.view_map.values(), []):
```
abergeron: itervalues from six

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,343 @@
...
+                    edge_params['color'] = self.node_colors['output']
+                elif hasattr(node.op, 'destroy_map') and \
+                        id in reduce(list.__add__,
+                                     node.op.destroy_map.values(), []):
```
abergeron: itervalues from six

theano/d3viz/formatting.py
```diff
@@ -0,0 +1,343 @@
...
+        s = str(t.dtype)
+    else:
+        s = dtype_to_char(t.dtype) + s
+    return s
+
+
+def dict_to_pdnode(d):
+    """Create pydot node from dict."""
+    e = dict()
+    for k, v in d.items():
```
abergeron: iteritems from six

theano/d3viz/tests/test_d3viz.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def check(self, f, reference=None, verbose=False):
+        tmp_dir = tempfile.mkdtemp()
+        html_file = pt.join(tmp_dir, 'index.html')
+        if verbose:
+            print(html_file)
+        d3v.d3viz(f, html_file)
+        assert pt.getsize(html_file) > 0
+        if reference:
+            assert filecmp.cmp(html_file, reference)
```
abergeron: From a discussion I had with @carriepl it was agreed that comparing against a reference file is not really useful, since the reference file would have to get updated any time we change a node's name for any reason.

Also I'm pretty sure this test will fail when not run with the default optimizer settings (since that changes the final graph).

abergeron:
For the flake8 failures, you should add d3viz/__init__.py to the whitelist and fix the problem in test_d3viz

carriepl:
From a discussion with @abergeron ; we probably shouldn't keep the full length versions of Javascript files from external libraries and instead only include and use the minimized versions.

cangermueller:
Thanks for your feedback @abergeron. I will submit a revisted version tomorrow.

cangermueller:
@carriepl: do you mean keeping dagre-d3.min.js and graphlib-dot.min.js, and removing dagre-d3.js and graphlib-dot.js from the js/ directory?

cangermueller:
@abergeron, you suggested to replace all `for x in .keys()`, `.values()`, `.items()` statements by `six.iterkeys(d)`, `six.itervalues(d)`, and `six.iteritems(d)`. Python3 will return an iterator for `.items()` statements and so on, which is what we want to have, whereas Python2 will return a list. Hence, my code works for Python3 and Python2, although it is somewhat less efficient for Python2. Considering that performance is not crucial for visualization, and everybody will hopefully use Python3 anyway in a couple of days, I suggest to leave the code as clean as it is, and not to introduce the `six` dependency. Or am I missing anything?

theano/d3viz/tests/test_d3viz.py
cangermueller: I agree, which is why I did not compare against a reference at the beginning, but only checked if `d3viz()` terminates with an output file. In `test_formatting`, I test the `pydot` graph which is used by `d3viz90` and later visualized by javascript. I will remove the comparison again!

theano/d3viz/d3viz.py
cangermueller: I agree, this was legacy code!

carriepl:
@cangermueller : yes, exactly

abergeron:
While I agree that this is not likely to cause problems in its current form, using six will help make the code look like the rest of the code and make sure that future changes will not break compatibility.

Also, I think python 2 will stay alive for a couple of years still.

cangermueller:
I removed full-fledged external javascript libraries, used six to iterate, updated docstrings, and hopefully fixed all flake errors.

carriepl:
There are still full versions of the external javascript libraries in the folder doc/library/d3viz/examples/d3viz/js/
Also, Travis seems unhappy but I don't have access to it's results at the moment.

carriepl:
Finally got access to Travis' results. It complains that, in FAST_COMPILE, the tests for the 3d visualization are failing with a message like "Failed to import pydot. Please install pydot."

nouiz:
pydot is an optional dependency. If it is not present, skip the test.

cangermueller:
@carriepl, which js library do you mean? There is no minimized version for `d3viz.js` and `d3-context-menu.js`, which I implemented myself.

cangermueller:
@nouiz, how can I skip the nose test? Alternative, we could add a requirements.txt files with pydot.

carriepl:
@cangermueller : doc/library/d3viz/examples/d3viz/js/dagre-d3.js and doc/library/d3viz/examples/d3viz/js/graphlib-dot.js

Also, to skip a test : 
```
from theano.tests.unittest_tools import SkipTest
def test_something():
    if my_condition:
        raise SkipTest("This is the reason for skipping the test")
    ...
```

cangermueller:
I removed the full version of dagre-d3.js and graphlib-dot.js [from my repository](https://github.com/cangermueller/Theano/tree/master/theano/d3viz/js). They should also disappear in the pull request.

Without pydot print, all tests in `d3viz/tests` fail. Hence, it does not make sense to skip a single tests. One solution would be to skip all d3viz tests if pydot is missing. What do you think?

carriepl:
This also makes sense. For instance, you can look at the file https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/tests/test_dnn.py. It raises a single SkipTest exception at the beginning of the file when the necessary libraries are not available.

carriepl:
Btw, the files you deleted in your repository still show up in your pull-request

cangermueller:
@carriepl, I added and removed the full javascript libraries. Now, they don't appear anymore in the [latested repository](https://github.com/Theano/Theano/tree/ad5f9c85727892e306e08cccd4c355f2d7f09bca/theano/d3viz/js).

cangermueller:
Finally, Travis does not complaint anymore!

carriepl:
@cangermueller Actually, you have a commit adding back the full libraries in theano/d3viz/js/ and then deleting them. But the full libraries in doc/library/d3viz/examples/d3viz/js/ are still there.

carriepl:
Addendum, these two additional commits that add and remove the same things should be simply removed. And, if possible, the original commits that added the full libraries can be altered (either via amend or interactive rebase) to prevent the full libraries from ever being in the commit history.

carriepl:
Good. All that remains is the addendum I mentioned to keep the history clean (and therefore the time to do a checkout reasonable).

cangermueller:
@carriepl, I removed the add/remove commits from the history. I did not alter the commit that adds the full javascript libraries in the first instance, since many following commits depend on it, and it would be a pain to change all of them. Travis does not complain now any more.

carriepl:
@cangermueller, Ok I think at this point it's ready for merge. @abergeron, @nouiz and I had a look and what we asked for was done and Travis is also happy with this PR.

Thanks a lot for all of this. It will be very handy to have in Theano.

cangermueller:
Cool, thanks!

Issue #3031, opened 2015-06-15
theano.tensor.constant([1]).reshape(()) fails


jsalvatier:
theano.tensor.constant([1]).reshape(()) fails with:

```
/Library/Python/2.7/site-packages/theano/tensor/basic.pyc in make_node(self, x, shp)
   3831         shp = as_tensor_variable(shp, ndim=1)
   3832         if not shp.dtype.startswith('int'):
-> 3833             raise TypeError("Shape must be integers", shp, shp.dtype)
   3834         assert shp.ndim == 1
   3835         if isinstance(shp, TensorConstant):

TypeError: ('Shape must be integers', TensorConstant{[]}, 'float64')
```

But np.array([1]).reshape(()) succeeds, so tensor.reshape should too probably.

lamblin:
Right, an empty shape should work.
I think we could either a special case for an empty vector here, or maybe add an option to `as_tensor_variable` to specify a default `dtype` if the original is empty and does not have one.

PR #3323, opened 2015-08-24
theano.tensor.constant reshape fix

SinaHonari:
fixes issue #3031

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4126,14 +4126,13 @@ class Reshape(Op):
         x = as_tensor_variable(x)
         shp_orig = shp
         shp = as_tensor_variable(shp, ndim=1)
-        if not shp.dtype.startswith('int') and not(isinstance(shp,
-                                                   TensorConstant) and
-                                                   shp.data.size == 0):
-            # if shp is an instace of TensorConstant and
-            # shp.data is empty, then shp.dtype is float64
+        if not (shp.dtype.startswith('int') or
+               (isinstance(shp, TensorConstant) and shp.data.size == 0)):
+            # It raises an error if shp is not of integer type,
+            # except when shp is constant and empty
+            # (in this case, shp.dtype does not matter anymore).
             raise TypeError("Shape must be integers", shp, shp.dtype)
-        if not(isinstance(shp, TensorConstant) and not shp.data.size == 0):
-            assert shp.ndim == 1
+        assert shp.ndim == 1
         if isinstance(shp, TensorConstant):
             bcast = [s == 1 for s in shp.data]
             return gof.Apply(self, [x, shp], [tensor(x.type.dtype, bcast)])
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4126,9 +4126,14 @@ class Reshape(Op):
         x = as_tensor_variable(x)
         shp_orig = shp
         shp = as_tensor_variable(shp, ndim=1)
-        if not shp.dtype.startswith('int'):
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -4126,9 +4126,14 @@ def make_node(self, x, shp):
...
-        if not shp.dtype.startswith('int'):
+        if not shp.dtype.startswith('int') and not(isinstance(shp,
+                                                   TensorConstant) and
+                                                   shp.data.size == 0):
+            # if shp is an instace of TensorConstant and
+            # shp.data is empty, then shp.dtype is float64
             raise TypeError("Shape must be integers", shp, shp.dtype)
-        assert shp.ndim == 1
+        if not(isinstance(shp, TensorConstant) and not shp.data.size == 0):
+            assert shp.ndim == 1
```
bouthilx: I was wrong, I verified and ```as_tensor_variable(()).ndim == 1```. You don't need to modify this.

theano/tensor/basic.py
```diff
@@ -4126,9 +4126,14 @@ def make_node(self, x, shp):
         x = as_tensor_variable(x)
         shp_orig = shp
         shp = as_tensor_variable(shp, ndim=1)
-        if not shp.dtype.startswith('int'):
+        if not shp.dtype.startswith('int') and not(isinstance(shp,
+                                                   TensorConstant) and
+                                                   shp.data.size == 0):
```
bouthilx: I think it would be more readable this way
```python
if (not shp.dtype.startswith('int') and
    not (isinstance(shp, TensorConstant) and shp.data.size == 0)):
```
And you could even factor out the negation
```python
if not (shp.dtype.startswith('int') or
        (isinstance(shp, TensorConstant) and shp.data.size == 0)):
```

theano/tensor/basic.py
```diff
@@ -4126,9 +4126,14 @@ def make_node(self, x, shp):
...
+            # if shp is an instace of TensorConstant and
```
bouthilx: instance*

theano/tensor/basic.py
```diff
@@ -4126,9 +4126,14 @@ def make_node(self, x, shp):
...
+            # shp.data is empty, then shp.dtype is float64
```
bouthilx: I would rephrase it as : 
> It raises an error if shp is not of integer type, except when shp is constant and empty (shp.dtype does not matter anymore).

bouthilx:
@SinaHonari 
Thanks for the quick PR!

You still need to add a test case to ensure this bug does not happen again. Add a new method ```test_empty_shp()``` at the end of this [class](https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_basic.py#L4914)

SinaHonari:
Boss your comments and concerns are all realized :)

bouthilx:
Thanks Sina! But you forgot about the unit-test. ;)

SinaHonari:
What should the unit test do?

bouthilx:
Test the basic case reported and make sure it doesn't fail. #3031 

```python
theano.tensor.constant([1]).reshape(())
```

As I said above

> You still need to add a test case to ensure this bug does not happen again. Add a new method ```test_empty_shp()``` at the end of this [class](https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_basic.py#L4914).

Should be quite simple. :)

SinaHonari:
You can check if the test is sufficient.

bouthilx:
Please compile the function and verify that output.shape == ()
That will be sufficient. :)

bouthilx:
Looks good to me, Thanks Sina!

PR #3337, opened 2015-08-28
Warn about pending ignore_border default value change.

nouiz:
We need to change it. The current default isn't useful and cause slower code as cudnn do not support it.

### DIFF
--- a/theano/tensor/signal/downsample.py
+++ b/theano/tensor/signal/downsample.py
@@ -9,6 +9,7 @@ from __future__ import print_function
 # This file should move along with conv.py
 from six.moves import xrange
 import six.moves.builtins as builtins
+import warnings
 
 import numpy
 
@@ -44,7 +45,7 @@ def max_pool_2d_same_size(input, patch_size):
     return outs
 
 
-def max_pool_2d(input, ds, ignore_border=False, st=None, padding=(0, 0),
+def max_pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
                 mode='max'):
     """
     Takes as input a N-D tensor, where N >= 2. It downscales the input image by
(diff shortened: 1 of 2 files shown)


theano/tensor/signal/downsample.py
```diff
@@ -58,7 +59,7 @@ def max_pool_2d(input, ds, ignore_border=False, st=None, padding=(0, 0),
     ds : tuple of length 2
         Factor by which to downscale (vertical ds, horizontal ds).
         (2,2) will halve the image in each dimension.
-    ignore_border : bool
+    ignore_border : bool (default None, mean print warning and set False)
```
JesseLivezey: I would change this to:
```
(default None, will print a warning and set to False)
```

theano/tensor/signal/downsample.py
```diff
@@ -77,6 +78,14 @@ def max_pool_2d(input, ds, ignore_border=False, st=None, padding=(0, 0),
     """
     if input.ndim < 2:
         raise NotImplementedError('max_pool_2d requires a dimension >= 2')
+    if ignore_border is None:
+        warnings.warn("max_pool_2d() will have the parameter ignore_border"
+                      " default value change to True (currently"
+                      " False). To be safe with all Theano versions,"
+                      " explicitly add the parameter ignore_border=True."
+                      " (this is also faster then ignore_border=False)",
+                      stacklevel=2)
```
JesseLivezey: And this to
```
warnings.warn("max_pool_2d() will have the parameter ignore_border"
                        " default value changed to True (currently"
                        " False). To have consistent behavior with all Theano versions,"
                        " explicitly add the parameter ignore_border=True."
                        " (this is also faster than ignore_border=False)",
                        stacklevel=2)
```

nouiz:
Thanks, I updated the PR.

PR #3339, opened 2015-08-29
[CRASH] fix crash on Raspberry Pi 1

nouiz:
close gh-3266


abergeron:
This fixes a number of things beside crashes on Rpi, but everything seems fine.

PR #3343, opened 2015-08-31
Add a 'name' argument to '.copy' method of Theano variables

yingzha:


yingzha:
@dwf I modified the arg in pprint.assign, I guess it will pass the Travis.

theano/tensor/basic.py
```diff
@@ -2815,9 +2815,11 @@ def do_constant_folding(self, node):
 pprint.assign(alloc, printing.FunctionPrinter('alloc'))
 
 
-"""Create a duplicate of `a` (with duplicated storage)"""
-tensor_copy = elemwise.Elemwise(scal.identity)
-pprint.assign(tensor_copy, printing.IgnorePrinter())
+def tensor_copy(input):
```
dwf: I don't think this change is necessary?

theano/tensor/var.py
```diff
@@ -520,8 +520,11 @@ def take(self, indices, axis=None, mode='raise'):
         return theano.tensor.subtensor.take(self, indices, axis, mode)
 
     # COPYING
-    def copy(self):
-        return theano.tensor.basic.tensor_copy(self)
+    def copy(self, name=None):
+        """Copy a variable and set a name to the copy."""
```
dwf: "and optionally assign a name"

dwf:
This should probably be rebased and I think one of the changes is irrelevant (the change to `tensor_copy`). Otherwise it is fine.

theano/tensor/basic.py
yingzha: OK, I'll rebate it

nouiz:
You don't need to create a new PR when you rebase. You can force the push
again your github branch.

yingzha:
@nouiz Since I made changes on master branch before, I'm not sure if I can rebate it directly. That's why I copied the changes to another branch and closed this one.

PR #3350, opened 2015-09-01
Embed theano

AlOa:
Compile Theano's functions into shared libraries and call them from c++ program using Python/Numpy C api


theano/gof/cc.py
```diff
@@ -287,21 +287,23 @@ def struct_gen(args, struct_builders, blocks, sub):
...
             %(struct_cleanup)s
             %(storage_decref)s
-        }
-        int run(void) {
+        } // end cleanup()
+        int run(void);
+    };
+    """ % sub
+    run_code = """
+        DllExport int %(name)s::run(void) {
```
abergeron: Can you make run() not always be exported.  We try to keep the namespace clean for performance reasons.

theano/gof/cc.py
```diff
@@ -1321,6 +1327,17 @@ def cmodule_key_(self, fgraph, no_recycling, compile_args=None,
...
+        # We append it only if we are c_callable to don't trash the
+        # old compiled dir.
+        if c_callable:
+            sig.append('c_callable: ' + str(self.c_callable))
+            constants_vals = ""
+            for var in self.orphans:
+                if isinstance(var, graph.Constant):
+                    constants_vals += str(var.data.flatten())
+            if constants_vals != '':
+                sig.append('costants: '+constants_vals)
```
abergeron: costants -> constants

theano/gof/cc.py
```diff
@@ -1445,22 +1464,116 @@ def compile_cmodule(self, location=None):
...
+                        preargs=preargs,
+                        shared=False, py_module=False,
+                        code_filename='exec.cpp',
+                        out_filename='exec')
+                    mod_exec.gen_header(os.path.join(location, 'exec.h'))
+                    if sys.platform == "win32" and False:
+                        # I don't know why it work now, but this was needed in the past.
+                        # As this is complicated to find how to do it, I keep it here
+                        # just in case.
+                        mt = r"C:\Program Files (x86)\Microsoft SDKs\Windows\v7.0A\Bin\mt.exe"
```
abergeron: This is not ok since it will only work for people that have the Windows 7 SDK (32-bits).  We try to be a bit more flexible.

theano/gof/cc.py
```diff
@@ -1512,16 +1622,40 @@ def get_dynamic_module(self):
...
+                        mod.add_support_code(self.struct_code)
+                    else:
+                        mod.add_header_code("""
+                                            #ifdef _WIN32
+                                            #define DllExport __declspec(dllexport)
+                                            #else
+                                            #define DllExport
+                                            #endif
+                                            """)
+                        mod.add_header_code(self.struct_code)
```
abergeron: I can't help but feel that both parts of this condition do the same thing.

theano/gof/cc.py
```diff
@@ -1613,7 +1918,16 @@ class _CThunk(object):
...
-    def __init__(self, cthunk, init_tasks, tasks, error_storage):
+    def __init__(self, cthunk, init_tasks, tasks, error_storage, filename):
+        """
+        Parameters
+        ----------
+        cthunk: the CObject pointer used by run_cthunk
+        init_tasks: WRITEME
+        tasks: WRITEME
+        error_storage: WRITEME
+        filename: the dynamic lib that where this thunk is compiled.
```
abergeron: This is not numpydoc format

theano/gof/cmodule.py
```diff
@@ -208,7 +209,7 @@ def print_init(self, stream):
         print("}", file=stream)
 
     def add_include(self, str):
-        assert not self.finalized
+        # assert not self.finalized
```
abergeron: Why are you removing that assert?

theano/gof/cmodule.py
```diff
@@ -2039,11 +2061,135 @@ def try_flags(flag_list, preambule="", body="",
...
+            makefile += "%s " % (arg)
+        makefile += "\n"
+        for flags in cxx_falgs:
+            makefile += "CPPFLAGS+=%s\n" % (flags)
+        for flags in ld_flags:
+            makefile += "LDFLAGS+=%s\n" % (flags)
+        for flags in lib_flags:
+            makefile += "LDLIBS+=%s\n" % (flags)
+        makefile += "LDLIBS+=%s\n" % (library_name)
+        return makefile
```
abergeron: I think that generating a makefile is going a little far.  We do want people to be able to call our code but I don't think we need to provide *everything*.

theano/gof/op.py
```diff
@@ -922,7 +922,7 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
         # condition: either there was no c_code, or it failed
         return self.make_py_thunk(node, storage_map, compute_map, no_recycling)
 
-
+
```
abergeron: Please remove extra whitespace.

theano/tensor/nnet/nnet.py
```diff
@@ -157,7 +157,7 @@ def c_code_template(dtype):
         {
             if (NULL != %(sm)s) Py_XDECREF(%(sm)s);
             %(sm)s = (PyArrayObject*)PyArray_SimpleNew(2, PyArray_DIMS(%(x)s),
-                                                       PyArray_TYPE((PyArrayObject*) py_%(x)s));
+                                                       PyArray_TYP(%(x)s));
```
abergeron: TYP -> TYPE

theano/tensor/nnet/nnet.py
```diff
@@ -999,7 +999,7 @@ def c_code_template(dtype):
         {
             if (NULL != %(nll)s) Py_XDECREF(%(nll)s);
             %(nll)s = (PyArrayObject*)PyArray_SimpleNew(1,
-                PyArray_DIMS(%(y_idx)s), PyArray_TYPE((PyArrayObject*) py_%(x)s));
+                PyArray_DIMS(%(y_idx)s), PyArray_TYPE(%(y_idx)s));
```
abergeron: Are you sure this change is ok?

theano/gof/cc.py
```diff
@@ -1445,22 +1464,116 @@ def compile_cmodule(self, location=None):
...
+                        py_dll = os.path.join(pp[-1], "python27.dll")
+                        manifest = os.path.join(location, "py_dll.manifest")
+                        exec_f = os.path.join(location, "exec.exe")
+                        call_subprocess_Popen('"' + mt + '"' +
+                                            " -inputresource:" + py_dll + ";#2 -out:" + manifest)
+                        call_subprocess_Popen('"' + mt + '"' +
+                            " -manifest " + manifest +
+                            " -outputresource:" + exec_f)
+
+            except Exception, e:
```
abergeron: `except Exception as e` for python 3

theano/gof/tests/test_shared_lib.py
```diff
@@ -0,0 +1,37 @@
...
+                      ((a, b), [a + b, a - b]),
+                  ]:
+        linker = theano.gof.CLinker(c_callable=True)
+        mode = theano.Mode(linker=linker)
+        f = theano.function(inps, out, mode=mode,
+                            on_unused_input='ignore')
+        theano.printing.debugprint(f, print_type=True)
+        #filename = f.fn.thunks[0].filename  # with linker=vm
+        filename = f.fn.filename  # with linker=c
+        print filename
```
abergeron: `print()` for python 3

doc/shared_library_howto/function.py
```diff
@@ -0,0 +1,11 @@
...
+import theano.tensor as T
+import numpy as np
+
+if __name__ == '__main__':
+    x = T.matrix('x')
+    y = T.vector('y')
+    linker = theano.gof.CLinker(c_callable=True)
+    mode = theano.Mode(linker=linker)
+    f = theano.function([x, y], [2 * x, T.dot(y, x)], mode=mode)
+    print f.fn.filename
```
abergeron: `print()` for python3

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+linker = theano.gof.CLinker(c_callable=True)
+mode = theano.Mode(linker=linker)
+f = theano.function(inputs, outputs , mode=mode)
+
+The shared library is build in a specific temp directory, in the Theano cache
+directory. To know in which directory, you can call print the fn.name
+property of the function
+
+print f.fn.filename
```
abergeron: If that is supposed to be a code block it needs to be indented.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
+
+Compiling and Using Theano functions as shared libraries
```
abergeron: Add capitals to the first letter of each word.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+========================================================
+
+First one important note: this functionality is experimental, so not every
+features works, and some may not work as you expected.
```
abergeron: This would be better served with a `.. note` section to draw attention to it.

Also the text would read better as:

This functionality is experimental.  Not every feature works and some may break in unexpected ways or behave differently.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+This functionality was only be tested on CPU code on a Linux machine.
```
abergeron: only be tested -> only tested

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+For now, shared variables not works.
```
abergeron: Rewrite as:

For now shared variables are not supported.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+Clearly every node of function that you want to use as shared library must
+have a C implementation.
```
abergeron: Clearly every -> Every

node of function -> node of a function

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+ 
+To compile a function as shared library, you must use the CLinker with the the
```
abergeron: as shared library -> as a shared library

must use the CLinker -> must use CLinker

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+parameter c_callable=True:
+
+linker = theano.gof.CLinker(c_callable=True)
+mode = theano.Mode(linker=linker)
+f = theano.function(inputs, outputs , mode=mode)
```
abergeron: Missing indentation for code block.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+The shared library is build in a specific temp directory, in the Theano cache
```
abergeron: is build in -> is built in

specific temp directory -> specific temporary directory

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+directory. To know in which directory, you can call print the fn.name
+property of the function
```
abergeron: Rewrite the second sentence as:

The name of the directory is stored in the name property of the function.

Also since the example below disagrees which one is right?  Is it `name` or `filename`

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+print f.fn.filename
+
+In this directory you find the .so library, the header file, a makefile and a
```
abergeron: you find the -> you will find the

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+exe.cpp which can be used as template to build your application. There is also
+a exec.h in which you find the command used to build the library.
```
abergeron: a exec.h -> an exec.h

you find -> you will find

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+Let's walk through an example to see how use the shared library.
```
abergeron: how use -> how to use

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+Build the shared library for the function in function.py:
+
+python function.py
```
abergeron: Indent for code block.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+the outputs show you where Theano puts the library. To see if everything
```
abergeron: the -> The

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+works correctly go in that directory and launch the exec executable. You must
+get:
```
abergeron: You must get -> You should get

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+After import numpy (hexvalue)
+after import_array1()
+after cinit()
+run() from the shared library returned=0
+[[  0.   2.   4.   6.]
+ [  8.  10.  12.  14.]
+ [ 16.  18.  20.  22.]]
+[ 20.  23.  26.  29.]
+main end, before Py_Finalize
```
abergeron: Once again ident for code block.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+Please note that, in general case, the exec can give some errors for generic
+functions because it makes some assumption on the data the can be wrong. Take
+it as a simple template.
```
abergeron: This might be better as a `.. note` section.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+after import_array1()
+after cinit()
+run() from the shared library returned=0
+[[  0.   2.   4.   6.]
+ [  8.  10.  12.  14.]
+ [ 16.  18.  20.  22.]]
+[ 20.  23.  26.  29.]
+main end, before Py_Finalize
+
+Please note that, in general case, the exec can give some errors for generic
```
abergeron: Please note that, in the general case, -> In general,

abergeron: the exec can -> the program can

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+functions because it makes some assumption on the data the can be wrong. Take
```
abergeron: makes some assumption -> makes some assumptions

the data the can -> the data that can

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+it as a simple template. 
+
+Theano compiles the function as a C++ class with variables mapped to internal
+variables. To run the function you initialized the class calling the cinit
```
abergeron: you initialized -> you initialize

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+function. Then pass the values of inputs variables as numpy array and call the
```
abergeron: function. Then pass -> function, then pass

as numpy array -> as numpy arrays

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+run method. If there is no errors you can get the outputs values from C++
```
abergeron: If there is no -> If there are no

from C++ class -> from the C++ class members.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+class.
+To C++ class variable must be set and get using the PyList_SetItem and
```
abergeron: I have no idea what this sentence is supposed to be, but it is wrong (syntactically).

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+PyList_GET_ITEM respectively.
+
+Open the exec.cpp with your preferred editor. After the includes  you find a
```
abergeron: Rewrite the first two sentences as:

In the exec.cpp file you will find after the includes a comment with the mapping between theano inputs and class members.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+comment with the mapping between the function variables and class variables
+that you can use to set inputs and get outputs.
+In main Python is initialized e numpy is imported. After the C++ is
```
abergeron: `main` needs some markup to signify that it's a noun.

is initialized e numpy -> is initialized and numpy

Rewrite the sentence that starts with `After the C++` to:

Then the theano library is initialized with a call to `cinit()`.  After that there is the definition of two numpy arrays and they are set as function inputs.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+initialized calling the  cinit function, you find the can find the defintion
+of two numpy arrays and how they are set as function inputs. Then the run
```
abergeron: Will it always be two numpy arrays or is that just for this example.  If it's just for the example you might want to explain that.

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+method is called and the outputs values are return if there are no errors.
```
abergeron: are return -> are returned

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+
+One thing you can do to better understand how code work, is to change the
```
abergeron: how code works -> how the code works

abergeron: Also delete the comma just after works

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
+
```
abergeron: Couldn't this be in the tutorial/ directory with the rest of the tutorials?

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+comment with the mapping between the function variables and class variables
+that you can use to set inputs and get outputs.
+In main Python is initialized e numpy is imported. After the C++ is
+initialized calling the  cinit function, you find the can find the defintion
+of two numpy arrays and how they are set as function inputs. Then the run
+method is called and the outputs values are return if there are no errors.
+
+One thing you can do to better understand how code work, is to change the
+inputs values. As an example you can delete from 35 to 52 and substitute them
+with;
```
abergeron: with; -> with:

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+Open the exec.cpp with your preferred editor. After the includes  you find a
+comment with the mapping between the function variables and class variables
+that you can use to set inputs and get outputs.
+In main Python is initialized e numpy is imported. After the C++ is
+initialized calling the  cinit function, you find the can find the defintion
+of two numpy arrays and how they are set as function inputs. Then the run
+method is called and the outputs values are return if there are no errors.
+
+One thing you can do to better understand how code work, is to change the
+inputs values. As an example you can delete from 35 to 52 and substitute them
```
abergeron: I will assume that 35 to 52 refers to lines and thus this:

delete from 35 to 52 -> delete from line 35 to line 52

doc/shared_library_howto/howto.txt
```diff
@@ -0,0 +1,107 @@
...
+There are differents way to intialize numpy vectors with yours values, using numpy C api (http://docs.scipy.org/doc/numpy/reference/c-api.html) or using something like BoostNumpy (https://github.com/ndarray/Boost.NumPy).
+
+
+
+
+
+
+  
+
+
```
abergeron: Remove extra newlines at the end.

theano/gof/cc.py
```diff
@@ -1445,22 +1464,116 @@ def compile_cmodule(self, location=None):
                 preargs.remove('-DREPLACE_WITH_AMDLIBM')
             if 'amdlibm' in libs:
                 libs.remove('amdlibm')
-        # We want to compute the code without the lock
+
+        if self.c_callable:
+            # Add the include filename with the placeholder, as the hash is not
+            # yet computer, but we need to add the include to compute the hash.
```
abergeron: computer -> computed

theano/gof/cc.py
```diff
@@ -1445,22 +1464,116 @@ def compile_cmodule(self, location=None):
...
+                    mod_exec.add_header_code(
+                        "//command line used to compile the shared lib: \n" +
+                        "//" + ' '.join(
+                            c_compiler.compile_command(
+                                module_name=mod.code_hash,
+                                location=location,
+                                include_dirs=self.header_dirs(),
+                                lib_dirs=self.lib_dirs(),
+                                libs=libs,
+                                preargs=preargs)[2]))
```
abergeron: Why do we do this?

theano/gof/cc.py
```diff
@@ -1445,22 +1464,116 @@ def compile_cmodule(self, location=None):
...
+                            c_compiler.compile_command(
+                                module_name=mod_exec.code_hash,
+                                location=location,
+                                include_dirs=self.header_dirs(),
+                                lib_dirs=self.lib_dirs(),
+                                libs=libs,
+                                preargs=preargs,
+                                shared=False, py_module=False,
+                                code_filename='exec.cpp',
+                                out_filename='exec')[2]))
```
abergeron: Why do we do this twice with a slight difference (the input/output file name)?

theano/gof/cc.py
nouiz: Windows isn't a prio. So if you prefer to note do it, we will still accept the PR. Just add a TODO here in that case.

theano/gof/cmodule.py
nouiz: I think we should keep the makefile generation. When I worked on that, the flags where hard to find manually.

theano/tensor/nnet/nnet.py
nouiz: the make_node() method tell that the nll output have the same dtype as x, not as y_idx. So just change y_idx to x in the part that you changed. Good catch.

theano/gof/cmodule.py
```diff
@@ -2039,11 +2061,135 @@ def try_flags(flag_list, preambule="", body="",
...
+            lib_dirs.append(python_lib)
+
+        cpp_filename = os.path.join(location, code_filename)
+
+        if shared:
+            assert out_filename is None
+            out_filename = os.path.join(location, '%s.%s' %
+                                        (module_name, get_lib_extension()))
+
+            _logger.debug('Generating shared lib %s', out_filename)
```
nouiz: The string don't match what this does. I would just remove it.

theano/gof/cmodule.py
```diff
@@ -2039,11 +2061,135 @@ def try_flags(flag_list, preambule="", body="",
...
+            cmd = [theano.config.cxx, get_gcc_shared_library_arg(), '-g']
+        else:
+            assert not py_module
+            if out_filename is None:
+                out_filename = os.path.join(location, module_name)
+            else:
+                out_filename = os.path.join(location, out_filename)
+            _logger.debug('Generating exec %s', out_filename)
```
nouiz: idem

theano/gof/cmodule.py
```diff
@@ -2039,11 +2061,135 @@ def try_flags(flag_list, preambule="", body="",
...
+            makefile += "CPPFLAGS+=%s\n" % (flags)
+        for flags in ld_flags:
+            makefile += "LDFLAGS+=%s\n" % (flags)
+        for flags in lib_flags:
+            makefile += "LDLIBS+=%s\n" % (flags)
+        makefile += "LDLIBS+=%s\n" % (library_name)
+        return makefile
+
+    @staticmethod
+    def compile_command(module_name, location=None,
```
nouiz: Since my original prototype, we extracted from compile_str some code that was moved to compile_args.

I think this new method should reuse it to move duplicate code.

nouiz: This would also remove possible maintenance problems.

Ideally, compile_str should reuse compile_command, but this can be let to later.

theano/gof/cmodule.py
```diff
@@ -2105,10 +2261,23 @@ def compile_str(module_name, src_code, location=None,
         if python_lib not in lib_dirs:
             lib_dirs.append(python_lib)
 
-        cppfilename = os.path.join(location, 'mod.cpp')
-        cppfile = open(cppfilename, 'w')
+        if shared:
+            hide_symbols = False
```
nouiz: I think that for now, this is good. But can you add a todo to tell that we should explicitly mark as visible the part that need that and don't override that?

theano/gof/cmodule.py
nouiz: compile_str already reuse it. But I think we should have compile_command call the new compile_args method.

nouiz:
About shared var. I think this could be in 2 steps.

1) compile the function as in this PR
2) add a method to the class Function that generate the pickle file with the shared variables.

Then the example executable could take a filename as parameter and use it to load the shared var data.

What do you think of this plan?

theano/gof/cc.py
AlOa: Yes, I think we can export the run method only in the case the function is compiled as shared library

theano/gof/cc.py
AlOa: Ok. I don't have a Windows machine so I can' test it.

theano/gof/cmodule.py
AlOa: The problem is that the module_from_key() first calls the DynamicModule code function that set the finalized flag, after that it compiles the code and in that stage the add_include function is called and the assert not self.finalized fails.

theano/gof/cmodule.py
AlOa: While I worked on the code generation I found the make file very useful.

theano/gof/cc.py
AlOa: If we generate the makefile, probably this can be removed

AlOa:
Your solution has the add that one can change the values of shared variables without recompile the code, which is useful.
So I think we can  follow this route.

theano/gof/cc.py
AlOa: I was wrong. 
I think we should keep it. Knowing the command and the flags used to compile the library can be useful.

theano/gof/cc.py
AlOa: For a c_callable function the declaration should be added to the header code of the shared library while for a 'normal' function it should be added directly to the *.cpp module.

theano/gof/cc.py
AlOa: I made a commit to remove the  DllExport for non c_callable function. But now I think that  is no enough . Probably we should make refactor the code so that for non c_callable function we can use a anonymous namespace.

theano/gof/cmodule.py
AlOa: It seems to me that it already use GCC_compiler.compile_args(). compile_command is called by compile_args which is called in lazylinker_c.py at line 126:

args = cmodule.GCC_compiler.compile_args()
cmodule.GCC_compiler.compile_str(dirname, code, location=loc, preargs=args)

we can make a refactor the code so that compile_args() will be called inside compile_commad.

AlOa:
I'm working on the C++ code for loading shared variables from files. I'm thinking to use the name of the C++ class internal variable connected to shared variable as  name of the file ( the mapping is printed at the beginning of the exec.cpp file). Do you have a better idea?

AlOa:
I added the management of shared variables.
The shared variables are loaded from a npy files that have the name of the C++ class internal variables connected to them (The mapping is printed at the beginning of the exec file)
The files must be in the same directory in which the executable file ru

AlOa:
Hi,
now everything seem to work correctly ( all test are passed ) but travis signals a syntax error on a dictionary comprehension that seems right. I missing something?

abergeron:
You have to put `k, v` in parentheses: `(k, v)`.

AlOa:
I made the correction.
Now I will correct the conflict with the last  head branch caused by the pull request #3694.
After that you will think that this pr can be merged?

AlOa:
I added the parentheses, but travis continues to signal a syntax error in dictionary comprehension.
Frankly I don't understand why.

theano/gof/cc.py
```diff
@@ -1648,6 +1796,205 @@ def instantiate_code(self, n_args):
...
+                   **locals())
+        # TODO, should struct_ptr.cleanup() cleanup the __ERROR structure
+        return main
+
+    def cinit_code(self):
+        code = StringIO()
+        n_args = len(self.args)
+        struct_name = self.struct_name
+        param = ','.join('PyObject * arg_%i' % n for n in range(n_args)), ');'
+        symbol2r = {v: k for (k, v) in self.r2symbol.items()}
```
nouiz: This isn't valid python 2 syntax.

I think this would work:

`dict([(v, k) for (k, v) in self.r2symbol.items()])`

theano/gof/cc.py
```diff
@@ -1565,16 +1689,39 @@ def get_dynamic_module(self):
...
-                mod.add_support_code(support_code)
-            mod.add_support_code(self.struct_code)
-            mod.add_support_code(static)
-            mod.add_function(instantiate)
-            for header in self.headers():
-                mod.add_include(header)
-            for init_code_block in self.init_code() + self.c_init_code_apply:
-                mod.add_init_code(init_code_block)
-            self._mod = mod
+                for support_code in self.support_code() + self.c_support_code_apply:
```
nouiz: I think all the new code in this section should be less indented by 1 level.

I think this should fix problems in the tests.

AlOa:
Now, It remains only a problem with test of the shared library.
I can't reproduce it on my machine, but looking at travis logs, it seems a linking  problem.

AlOa:
Hi,
I returned to work to this project and I merged with the last mater. The problem is still the test on shared library, that fails on travis. It seems that when It tries to compile the c++ code It can't find the include file,  I don't understand why.

nouiz:
There is a few flake8 error left:

```
/home/travis/build/Theano/Theano/theano/gof/cc.py:1628:25: F821 undefined name 'call_subprocess_Popen'
/home/travis/build/Theano/Theano/theano/gof/cc.py:1630:25: F821 undefined name 'call_subprocess_Popen'
/home/travis/build/Theano/Theano/theano/gof/cmodule.py:2141:9: F841 local variable 'cxx_args' is assigned to but never used
/home/travis/build/Theano/Theano/theano/gof/tests/test_shared_lib.py:1:1: FI10 __future__ import "division" missing
/home/travis/build/Theano/Theano/theano/gof/tests/test_shared_lib.py:1:1: FI11 __future__ import "absolute_import" missing
/home/travis/build/Theano/Theano/theano/gof/tests/test_shared_lib.py:1:1: FI13 __future__ import "print_function" missing

```

For the last 3, add this as the first import:

AlOa:
Now I get a sphinx error that, frankly, I do not understand.

lamblin:
If I understand correctly, sphinx complains that `shared_library_howto/howto.txt` is an orphan page, that is not part of any table of content, or is not linked from anywhere.
I'm not sure what the best place to link from should be, though.

nouiz:
I think I would move the file from
doc/tutorial/shared_library_howto/howto.txt to doc/tutorial/shared_lib.txt

Then in the file doc/tutorial/index.txt, as shared_lib at the last line.

thanks

AlOa:
I made as you said and now Travis pass.
Thanks

PR #3356, opened 2015-09-02
cuDNN v3 support for gpuarray

abergeron:


abergeron:
Could someone look at this?

abergeron:
This is currently broken due to the explicit context activation requirement.  I have code that fixes this, but I also need #3364 to test it properly.

abergeron:
Now it is unbroken and ready for cudnn v3 final release.

abergeron:
I did the required rebase.

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1783,6 +1395,22 @@ def local_softmax_dnn(node):
         return [out]
```
lamblin: Should that one check for the version, because of the bug when taking softmax on (1, 1) images?

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1783,6 +1395,22 @@ def local_softmax_dnn(node):
...
+    if (isinstance(node.op, GpuElemwise) and
+            isinstance(node.op.scalar_op, Log) and
+            node.inputs[0].owner and
+            isinstance(node.inputs[0].owner.op, GpuDnnSoftmax) and
+            len(node.inputs[0].clients) == 1):
+        softmax_node = node.inputs[0].owner
+        new_softmax = GpuDnnSoftmax(None, 'log', softmax_node.op.mode)
+        return [new_softmax(softmax_node.inputs[0])]
+
+
```
lamblin: Maybe this one as well, not sure. At least there should be tests.

theano/sandbox/gpuarray/tests/test_dnn.py
```diff
@@ -706,7 +692,7 @@ class test_SoftMax(test_nnet.test_SoftMax):
     mode = mode_with_gpu
 
     def test_softmax_shape_0(self):
-        raise SkipTest("Cudnn do not suport 0 shapes")
+        raise SkipTest("Cudnn doesn't suport 0 shapes")
```
lamblin: "support"

lamblin:
I've done a quick overview just to understand what is in there.
Do the unit tests run on the new back-end? without cudnn? with cudnn v2? v3?

theano/sandbox/gpuarray/dnn.py
```diff
@@ -288,105 +240,56 @@ def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv'):
...
-            bmode = 2
+            pad0 = str(self.border_mode[0])
+            pad1 = str(self.border_mode[1])
+            if len(self.border_mode) > 2:
+                pad2 = str(self.border_mode[2])
+            bmode = '2'
+        elif self.border_mode == "valid":
+            bmode = 1
+        elif self.border_mode == "full":
+            bmode = 0
```
lamblin: What's the semantics of `bmode`, and should it be a string everywhere?

theano/sandbox/gpuarray/dnn.py
```diff
@@ -288,105 +240,56 @@ def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv'):
...
-           pad_h_spec=pad_h_spec, pad_w_spec=pad_w_spec)
+        return [('NB_DIMS', str(len(self.subsample))),
+                ('BORDER_MODE', bmode),
+                ('PAD_0', pad0), ('PAD_1', pad1), ('PAD_2', pad2),
+                ('CONV_MODE', conv_flag),
+                ('SUB_0', sub0), ('SUB_1', sub1), ('SUB_2', sub2)]
 
     def c_code_cache_version(self):
-        return (1, version())
+        return (super(GpuDnnConvDesc, self).c_code_cache_version(), version())
```
lamblin: Should we have a number identifying the current version of C code for this Op? Something that would be obvious we should bump when changing it?

theano/sandbox/gpuarray/conv_desc.c
```diff
@@ -0,0 +1,34 @@
+#section support_code_apply
+
+int conv_desc(PyArrayObject *filt_shp,
+              cudnnConvolutionDescriptor_t *desc) {
```
lamblin: I'm confused. Is that function part of the support code for this apply node, or is it the actual function that will perform the thunk?

theano/sandbox/gpuarray/conv_desc.c
```diff
@@ -0,0 +1,34 @@
...
+  pad[1] = *(npy_int64 *)PyArray_GETPTR1(filt_shp, 3) - 1;
+#if NB_DIMS > 2
+  pad[2] = *(npy_int64 *)PyArray_GETPTR1(filt_shp, 4) - 1;
+#endif
+#endif
+
+  if (PyArray_DIM(filt_shp, 0) - 2 != NB_DIMS) {
+    PyErr_Format(PyExc_ValueError, "Filter shape has too many dimensions: "
+                 "expected %d, got %lld.", NB_DIMS,
+                 (long long)PyArray_DIM(filt_shp, 0));
```
lamblin: Should there be a call to `FAIL` when there is an error? I'm not sure when reading http://deeplearning.net/software/theano/tutorial/extending_theano_c.html, the examples do not use it, but the text mentions it.

theano/sandbox/gpuarray/cudnn_helper.h
```diff
@@ -4,193 +4,110 @@
 #include <cudnn.h>
 
 #ifndef CUDNN_VERSION
```
lamblin: Actually, should that file be shared by `cuda` and `gpuarray` back-ends?

theano/sandbox/gpuarray/dnn.py
abergeron: No that was only in v3 rc.  Since we only support final versions that is ok.

theano/sandbox/gpuarray/conv_desc.c
abergeron: It part of the support code, but is called by the c_code to perform the computation.

theano/sandbox/gpuarray/conv_desc.c
abergeron: FAIL only works in the init_code_struct, code and code_cleanup sections.  This is not one of those.

Also I do pass a function name to __init__ which places me under these conditions: http://deeplearning.net/software/theano/tutorial/extending_theano_c.html#main-function

theano/sandbox/gpuarray/cudnn_helper.h
abergeron: Probably, but the cuda backend contains lots of cudnnv1-compatible code that I removed from here.  I could do a cleanup of it too, but I decided against it.

theano/sandbox/gpuarray/dnn.py
abergeron: Yes it should be a string, and the semantics are the same as when it was an int, meaning it's defined in the code.

theano/sandbox/gpuarray/dnn.py
abergeron: Yes and COp takes care of that.  But it also implements c_code_cache_version_apply() which overrides this one and thus does not take into account the library version.

I'll have to fix that.

abergeron: Err, sorry I spoke too fast.  COp does implement c_code_cache_version() properly and since we reuse it here this will reflect any change we make to the code.

theano/sandbox/gpuarray/dnn.py
abergeron: I was certain I added tests for that, but apparently I didn't.

I'll copy the tests from the old backend.

abergeron:
I should've addressed all comments.

theano/sandbox/gpuarray/conv_desc.c
lamblin: So when you pass a function name to `__init__`, it automatically generates a `code` section that will call that function, and call `FAIL` if it does not return 0, is that right?

theano/sandbox/gpuarray/cudnn_helper.h
lamblin: OK.

theano/sandbox/gpuarray/dnn.py
lamblin: OK. I thought for a moment that there was still a `c_code` function somewhere that used `conv_desc` and that could change. But it is just using the one in the parent class, right?
But what if `get_op_params()` change? How would the cache be invalidated?

lamblin:
I'm not done yet, actually. I was just trying to understand what happened to the C code of GpuDnnConvDesc.

theano/sandbox/gpuarray/dnn.py
lamblin: In which code, where?

theano/sandbox/gpuarray/dnn.py
lamblin: OK.

abergeron:
Ok, then I should've addressed all comments to date (plus a few issues I spotted myself while doing so).

theano/sandbox/gpuarray/dnn_pool_grad.c
```diff
@@ -0,0 +1,114 @@
...
+
+#section init_code_struct
+
+APPLY_SPECIFIC(input) = NULL;
+APPLY_SPECIFIC(input_grad) = NULL;
+APPLY_SPECIFIC(output) = NULL;
+APPLY_SPECIFIC(output_grad) = NULL;
+
+{
+  cudnnStatus_t err;
```
lamblin: This one was APPLY_SPECIFIC before, is there a reason for that?

theano/sandbox/gpuarray/dnn_pool_grad.c
```diff
@@ -0,0 +1,114 @@
...
+  if ((err = cudnnCreateTensorDescriptor(&APPLY_SPECIFIC(input))) != CUDNN_STATUS_SUCCESS) {
+    PyErr_Format(PyExc_MemoryError,
+                 "could not allocate tensor descriptor (input): %s",
```
lamblin: Should those error messages include the name of the current Op/operation? They used to include "GpuDnnPoolGrad".

lamblin:
OK, thanks. Review still in progress.

theano/sandbox/gpuarray/dnn_pool.c
```diff
@@ -0,0 +1,93 @@
...
+  if (theano_prep_output(out, ndims+2, dims, img->ga.typecode,
+                         GA_C_ORDER, c) != 0)
+    return 1;
+
+  if (c_set_tensorNd(*out, APPLY_SPECIFIC(output)) != 0)
+    return 1;
+
+  {
+    const float alpha = 1;
+    const float beta = 0;
```
lamblin: Should be double if data is float64.

theano/sandbox/gpuarray/dnn_pool_grad.c
```diff
@@ -0,0 +1,114 @@
...
+  cudnnStatus_t err;
+  PyGpuContextObject *c = pygpu_default_context();
+
+  if (!GpuArray_IS_C_CONTIGUOUS(&inp->ga)) {
+    PyErr_SetString(PyExc_ValueError, "Only contiguous inputs are supported.");
+    return 1;
+  }
+
+  if (!GpuArray_IS_C_CONTIGUOUS(&out_grad->ga)) {
+    PyErr_SetString(PyExc_ValueError, "Only contiguous input gradients are supported.");
```
lamblin: "output gradients" instead?

lamblin:
If tests pass on cudnn v2, v3, and Travis, I'm OK to merge.

abergeron:
Tests passed on v2 and v3.  Travis had a flake8 problem which I just fixed.

PR #3357, opened 2015-09-02
Gpu reshape opt.

### LINKED ISSUE #3188: Opt generalize reshape(reshape(x, ...), shapes...) -> reshape(x, shapes) to the GPU.

Thrandis:
Fixes #3188

### DIFF
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -2445,6 +2445,8 @@ class GpuReshape(tensor.Reshape, GpuOp):
 
     # __hash__, __eq__, __str__ come from tensor.Subtensor
     def make_node(self, x, shp):
+        x = as_cuda_ndarray_variable(x)
+        shp = tensor.as_tensor_variable(shp)
         host_reshaped = host_from_gpu(x).reshape(shp, ndim=self.ndim)
         return Apply(self, [x, shp],
                      [CudaNdarrayType(host_reshaped.broadcastable)()])
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -944,6 +944,35 @@ def local_gpu_reshape(node):
     return False
 
 
+@local_optimizer([GpuReshape])
+def local_gpu_reshape_chain(node):
+    """
+    GuReshape(GpuReshape(shape1),shape2) -> GpuReshape(shape2)
+
+    """
+    if not tensor.opt.opt.check_chain(node, GpuReshape, GpuReshape):
+        return False
+
+    # TODO: this can permit a failing program to run by eliminating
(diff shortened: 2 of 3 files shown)


bouthilx:
@abergeron Can you please relaunch Scrutinizer?

theano/sandbox/cuda/basic_ops.py
```diff
@@ -2445,6 +2445,8 @@ class GpuReshape(tensor.Reshape, GpuOp):
 
     # __hash__, __eq__, __str__ come from tensor.Subtensor
     def make_node(self, x, shp):
+        x = as_cuda_ndarray_variable(x)
+        shp = tensor.as_tensor_variable(shp)
```
bouthilx: @abergeron Should we rather wrap x and shp inside the test? The only reason I could see for GpuReshape.make_node not being shielded against host/device variables is because GpuReshape is not in the public interface.

abergeron: This is fine like this.

theano/sandbox/cuda/opt.py
```diff
@@ -944,6 +944,35 @@ def local_gpu_reshape(node):
...
+    # but some other transformation replaced the shape by one for which this
+    # cannot be guessed.
+    # We should try to figure out why we lost the information about this
+    # constant value... but in the meantime, better not apply this
+    # optimization.
+    if rval.broadcastable == node.outputs[0].broadcastable:
+        return [rval]
+    else:
+        return False
+gpu_cut_copies.register('cut_local_gpu_reshape_chain',
```
nouiz: There is stuff not standard and missing on that registration. The good fix is to remove that manual registration and just register it with the register_opt decorator.

Also, can you add a comment in the original opt that this one exist and should probably be updated with as the first one? Another option is to refactor the first opt to be able to work on both, GpuReshape or Reshape.

bouthilx: Should we use ```register_opt(final_opt=True)```?

nouiz: This is a local optimizer. final opt are global optimizer. Also, this
should be in the equilibrium, no don't use any parameter.

On Thu, Sep 3, 2015 at 2:10 PM, Xavier Bouthillier <notifications@github.com

bouthilx: Oh ok. I though so because of this [line](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/opt.py#L114) where it uses argument ```final_opt=True``` for local optimization function ```tensor.opt.constant_folding```.

So we should use register_opt() which registers the local optimization inside a ```EquilibriumDB``` named ```gpu_optimizer```. There is no risk that some ```GpuReshape```s are added afterward?

PR #3360, opened 2015-09-03
Add some of the default ignores back.

abergeron:


nouiz:
Can merge when travis pass

PR #3361, opened 2015-09-03
Two-layer hierarchical softmax

adbrebs:
#2501 @bouthilx 

I haven't added any tests as the core sparse_block_dot is already tested.


nouiz:
this still need test. To at least that that it run and don't get erased, got typo by mistake.

I didn't review the code.

adbrebs:
Ok, I have added a test to check that it compiles well and that the dimensions are correct.

theano/sandbox/blocksparse.py
```diff
@@ -2,7 +2,6 @@
 
 import theano
 from theano import Op, Apply
-from theano import tensor
```
adbrebs: @bouthilx Having this line makes the code crash, I don't really understand why, probably an import cycle?

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1442,5 +1443,74 @@ def test_relu():
...
+    not.
+    """
+
+    #############
+    # Config
+    #############
+
+    input_size = 4
+    batch_size = 2
+    h_softmax_level1_size = 2
```
bouthilx: 5, just to make sure there is nothing interleaved with ```batch_size=2```

adbrebs: Done.

theano/tensor/nnet/nnet.py
```diff
@@ -29,6 +29,7 @@
 from theano.tensor.nnet.sigm import sigmoid, softplus
 from theano.gradient import DisconnectedType
 from theano.gradient import grad_not_implemented
+from theano.sandbox.blocksparse import sparse_block_dot
```
nouiz: tensor should not import sandbox.

Can we move blocksparse ton tensor/nnet/ directory?

bouthilx: I guess so. Is there any precise definition of what should be in sandbox?

nouiz: In theory, everything not well finished/tested/reviewed.

In practice, there is some stuff we should move out of sandbox... I think this part of the code is well tested/reviewed now, so it can be moved out of sandbox.

bouthilx: I agree, let's move blocksparse.py out of sandbox. But obviously sandbox/cuda/blocksparse.py will stay there.

adbrebs:
@bouthilx The tests of inplace optimization fail, I don't really understand why.
h_softmax tests fail whereas they run properly on my machine, I'm going to investigate.

bouthilx:
@adbrebs During the move, you erased theano/sandbox/__init__.py, which is correct, but did you added some equivalent import under theano/tensor/nnet/__init__.py? The optimizations needs to be imported in order to be added to optimization database.

adbrebs:
I tried to add the "import op" to  theano/tensor/nnet/init.py but I still have the same errors (code pushed).

adbrebs:
@bouthilx I tried to add the "import opt" to theano/tensor/nnet/init.py but I still have the same errors (code pushed).

bouthilx:
@adbrebs It did fix tests on jobs 7226.7 and 7226.8. I will take a look this afternoon for 7226.1 and 7226.4.

bouthilx:
@abergeron If I do ```from . import opt``` under ```tensor/nnet/``` I get ```opt.py``` from ```tensor/```. Do you understand why? I run the nosetests from ```Theano/``` root folder. I don't understand why it imports the opt.py file one level higher in the hierarchy. The only way I could fix it so far was with ```ìmport theano.tensor.nnet.opt```.

bouthilx:
Small update, it works if I add ```from . import opt``` on the first line of ```tensor/nnet/__init__.py``` rather than after ```from .nnet import *```. 

Furthermore, inside  ```tensor/nnet/__init__.py```, I have
```python
from . import opt as opt2
from .nnet import *
```
and when running it interactively I get
```python
>>> print opt.__file__
/path/to/Theano/theano/tensor/opt.py
>>> print opt2.__file__
/path/to/Theano/theano/tensor/nnet/opt.py
```

So ```opt``` comes from ```from .nnet import *```, but why would it be from ```theano.tensor``` rather than ```theano.tensor.nnet```!?

bouthilx:
I've found the culprit. In theano/tensor/nnet/nnet there is ```from theano.tensor import opt```. This *damn* ```opt``` bounces back in ```nnet/__init__.py``` because of ```from .nnet import *``` and take the place of ```nnet/opt.py``` in the namespace. I could either rename ```opt``` in ```nnet.py``` or use ```__all__``` to constrain the public interface of ```nnet.py```. As we are using ```import *``` all over the place, I'd be in favor of using ```__all__```.

@abergeron Any preference?

nouiz:
Import *isn't a good practice. We are slowly moving away from this. Could
you do it for that case?
Le 28 sept. 2015 12:05, "Xavier Bouthillier" <notifications@github.com> a
écrit :

bouthilx:
Thanks @nouiz. I'd be very happy to remove any ```import *```. :)

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,126 @@ def relu(x, alpha=0):
...
+        the bias vector of the second softmax layer.
+
+    target: tensor of shape either (batch_size,) or (batch_size, 1)
+        (optional, default None)
+        contains the indices of the targets for the minibatch
+        input x. For each input, the function computes the output for its
+        corresponding target. If target is None, then all the outputs are
+        computed for each input.
+
+    :note: n_outputs_per_class and n_classes do not need to be defined. If
```
bouthilx:
```python
Notes
-----
```

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,126 @@ def relu(x, alpha=0):
...
+        both are not defined, then they are set to the square root of the
+        number of outputs, which is the most computational efficient
+        configuration. If only one is defined
+
+    """
+
+    # In case one or both of n_outputs_per_class and n_classes are not defined
```
bouthilx: Do you have a unit-test for every possibilities?

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,126 @@ def relu(x, alpha=0):
...
+        # Second softmax that computes the output probabilities
+        activations = sparse_block_dot(
+            W2[None, :, :, :], x[:, None, :],
+            tensor.zeros((batch_size, 1), dtype='int32'), b2, class_ids)
+
+        output_probs = theano.tensor.nnet.softmax(
+            activations.reshape((-1, n_outputs_per_class)))
+        output_probs = output_probs.reshape((batch_size, n_classes, -1))
+        output_probs = class_probs[:, :, None] * output_probs
+        output_probs = output_probs.reshape((batch_size, -1))
```
bouthilx: ```:n_outputs``` because ```n_classes * n_outputs_per_class``` is not necessarily equal to ```n_outputs```?

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,126 @@ def relu(x, alpha=0):
...
+        output_probs = output_probs[:, :n_outputs]
```
bouthilx: ```:n_outputs``` because ```n_classes * n_outputs_per_class``` is not necessarily equal to ```n_outputs```?

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,126 @@ def relu(x, alpha=0):
...
+        (optional, default None)
+        contains the indices of the targets for the minibatch
+        input x. For each input, the function computes the output for its
+        corresponding target. If target is None, then all the outputs are
+        computed for each input.
+
+    :note: n_outputs_per_class and n_classes do not need to be defined. If
+        both are not defined, then they are set to the square root of the
+        number of outputs, which is the most computational efficient
+        configuration. If only one is defined
```
bouthilx: To be continued? :)

theano/tensor/nnet/nnet.py
adbrebs: Yes, I will add a comment.

theano/tensor/nnet/nnet.py
adbrebs: Hehe ;)

adbrebs:
Thanks @bouthilx, do you think it is necessary to add two more tests to test the cases in which the layer sizes are inferred?

bouthilx:
You could turn the unit-test into a class to reuse the setup. I see four different combinations with ```n_classes``` and ```n_outputs_per_class```. You test only one of them so far.

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,134 @@ def relu(x, alpha=0):
...
+    """
+
+    # In case one or both of n_outputs_per_class and n_classes are not defined
+    if not n_outputs_per_class and not n_classes:
+        n_outputs_per_class = numpy.ceil(numpy.sqrt(n_outputs))
+        n_classes = numpy.ceil(n_outputs / n_outputs_per_class)
+    elif n_outputs_per_class and not n_classes:
+        n_classes = numpy.ceil(n_outputs / n_outputs_per_class)
+    elif n_classes and not n_outputs_per_class:
+        n_outputs_per_class = numpy.ceil(n_outputs / n_classes)
```
adbrebs: @bouthilx Actually I realize that these cases don't make sense... The sizes are defined by ``W1`` and ``W2``, so it is the user who should be careful. I will update this tomorrow.

bouthilx: Unit-tests are so useful! Even if you do know write them down. ;)

bouthilx:
@adbrebs It needs rebase.

adbrebs:
@bouthilx Rebase done. The user should now specify both ``n_classes`` and ``n_outputs_per_class`` accordingly with the weight tensors.

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,121 @@ def relu(x, alpha=0):
         f1 = 0.5 * (1 + alpha)
         f2 = 0.5 * (1 - alpha)
         return f1 * x + f2 * abs(x)
+
+
+def h_softmax(x, batch_size, n_outputs, W1, b1, W2, b2,
+              n_classes, n_outputs_per_class, target=None):
```
bouthilx: One last very minor modification. Now that they are no longer optional, it would make more sense to place ```n_classes``` and ```n_outputs_per_class``` after ```n_outputs```. It's actually the current order in the docstring.

adbrebs: @bouthilx Done

bouthilx: @adbrebs Did you forget to adapt calls to ```h_softmax``` accordingly? ;)

bouthilx:
@adbrebs I almost forgot. You need to add the documentation in ```doc/library/tensor/nnet/nnet.txt```. You should also move ```sparse_block_dot()``` documentation.

adbrebs:
Thanks @bouthilx for reviewing, I have added the doc.

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,126 @@ def relu(x, alpha=0):
...
+        (optional, default None)
+        contains the indices of the targets for the minibatch
+        input x. For each input, the function computes the output for its
+        corresponding target. If target is None, then all the outputs are
+        computed for each input.
+
+    Returns
+    -------
+    tensor of shape (batch_size, n_outputs)
+        Output of the two-layer hierarchical softmax.
```
bouthilx: The output depends on target. The behavior should be made clear here too.

adbrebs: fixed

bouthilx:
@adbrebs I found a bug while trying to generate the moved documentation of ```SparseBlockDot```. The code for GPU under ```theano/sandbox/cuda/opt.py``` tries to import ```SparseBlockGemv``` and ```SparseBlockOuter``` from ```theano.sandbox.blocksparse```. 

I suggest you run again unit-tests for ```SparseBlockDot``` related ops on the GPU side. There might be a few other slips hiding in ```theano/sandbox/cuda``` .

Once you are done. Please verify that the doc is correct.

```bash
path/to/Theano$ python doc/scripts/docgen.py --nopdf
```

adbrebs:
@bouthilx 
I have updated the imports (I will change the name of the commits later).
The following test on GPU is failing: ``test_sparseblockgemv_grad`` in ``BlockSparse_Gemv_and_Outer``.
More precisely this line fails:
``utt.verify_grad(metaop, [b_val, h_val, W_val], mode=self.mode)``
I havent understood why.

Side question: when are the Theano GPU tests run?

bouthilx:
> I have updated the imports (I will change the name of the commits later).

Yes please! :)

> The following test on GPU is failing: test_sparseblockgemv_grad in BlockSparse_Gemv_and_Outer.
More precisely this line fails:
utt.verify_grad(metaop, [b_val, h_val, W_val], mode=self.mode)
I havent understood why.

I will take a look at it.

> Side question: when are the Theano GPU tests run?

GPU tests are run by our nightly build-bot at the lab. So if you comited asd-asdasd to see if travis tests pass, then it was useless. :)

adbrebs:
> GPU tests are run by our nightly build-bot at the lab. So if you comited asd-asdasd to see if travis tests pass, then it was useless. :)
I had just committed so that I could pull on the servers and try the GPU tests ;).

theano/tensor/nnet/nnet.py
```diff
@@ -2049,3 +2050,130 @@ def relu(x, alpha=0):
...
+
+def h_softmax(x, batch_size, n_outputs, n_classes, n_outputs_per_class,
+              W1, b1, W2, b2, target=None):
+    """ Two-level hierarchical softmax.
+
+    The architecture is composed of two softmax layers: the first predicts the
+    class of the input x while the second predicts the output of the input x in
+    the predicted class.
+    More explanations can be found in the original paper:
+    http://arxiv.org/abs/cs/0108006.
```
bouthilx: References should be formated as described in [NumPy/SciPy Documentation Guide](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections) (see subsection #11).

bouthilx:
Please rebase your PR and squash the two last commits. :)

bouthilx:
You should do something similar to ```theano/sandbox/conv.py``` for ```blocksparse.py``` to avoid breaking the interface.

adbrebs:
Rebase done!

theano/tensor/nnet/nnet.py
bouthilx: @nouiz It seems to me that sparse_block_dot is general enough to be in ```tensor``` rather ```tensor.nnet```. The function ```h_softmax``` however should definitely be in ```tensor.nnet```. Do you have any objection?

bouthilx:
@nouiz It seems to me that sparse_block_dot is general enough to be in tensor rather tensor.nnet. The function h_softmax however should definitely be in tensor.nnet. Do you have any objection?

nouiz:
That seem resonable, but I would like the opinion of @lamblin and/or @abergeron on this.

bouthilx:
@lamblin or @abergeron Any comment?

abergeron:
Yes, I think that it's general enough, but I wouldn't fight strongly to move it.

theano/sandbox/blocksparse.py
```diff
@@ -1,266 +1,11 @@
-import numpy
+from __future__ import print_function
+import sys
```
bouthilx: those two lines are useless

theano/sandbox/blocksparse.py
```diff
@@ -1,266 +1,11 @@
...
+import warnings
+from theano.tensor.nnet.blocksparse import (
+    SparseBlockGemv, SparseBlockOuter, sparse_block_dot, sparse_block_gemv,
+    sparse_block_gemv_inplace, sparse_block_outer, sparse_block_outer_inplace)
```
bouthilx: Add ```__all__ = [...]``` and flake8 won't complain anymore.

Issue #3162, opened 2015-07-17
Softmax+crossentropy can lead to nan in gradient for unbounded activations


jsusskin:
When using unbounded activation functions (e.g. Relu) the softmax function can saturate. This can lead to nan gradients when paired with categorical crossentropy cost. 
If the softmax function is replaced with a numerically stable version of log-softmax and this is used directly in the cost function, then the gradients don't blow up.
It seems that this could be implemented as a pattern to recognize (softmax paired with categorical crossentropy).
Here's a code snippet that illustrates the problem with the regular softmax versus doing the same thing with the numerically stable log-softmax, where the former gives nans in the gradient and the latter does not blow up. It's interesting because the experiment indicates that for the regular softmax case, the crossentropy loss is coming out numerically stable but not the gradient.

```
import theano
import theano.tensor as T
import numpy as np

x,y=T.matrices('xy')

# regular softmax and crossentropy
sm = T.nnet.softmax(x)
cm1=T.nnet.categorical_crossentropy(sm,y)
g1 = T.grad(cm1.mean(),x)

# numerically stable log-softmax with crossentropy
xdev = x-x.max(1,keepdims=True)
lsm = xdev - T.log(T.sum(T.exp(xdev),axis=1,keepdims=True))
sm2 = T.exp(lsm) # just used to show equivalence with sm
cm2=-T.sum(y*lsm,axis=1)
g2 = T.grad(cm2.mean(),x)


# create some inputs into a softmax that are large and labels
a=np.exp(10*np.random.rand(5,10).astype(theano.config.floatX))
# create some one-hot coded labels
b=np.eye(5,10).astype(theano.config.floatX)

# show equivalence of softmax and exponentiated numerically stable log-softmax
f1=theano.function([x],[sm, sm2])
sm1,sm2=f1(a)
print np.allclose(sm1,sm2)

# now show that the two versions result in the same crossentropy cost
# this indicates that the forward function does provide some numerical stability
f2=theano.function([x,y],[cm1,cm2])
c1,c2 = f2(a,b)
print np.allclose(c1,c2)

# now, show that in the standard softmax case the gradients blow up 
# while in the log-softmax case they don't
f3=theano.function([x,y],[g1,g2])
g1_,g2_ = f3(a,b)
print g1_
print g2_
```

OUTPUT:
True
True
[[ nan  nan  nan  nan  nan  nan  nan  nan  nan  nan]
[...]

PR #3363, opened 2015-09-03
LogSoftmax

fvisin:
Work in progress to fix https://github.com/Theano/Theano/issues/3162


nouiz:
I renamed the PR title, but didn't checked the code.

fvisin:
A list of what has been done (and tested) so far:
* [X] Implementation of the first and second gradient of the Logsoftmax
    * [X] python code
    * [ ] C code
* [ ] Optimization
    * [X] fprop
    * [ ] bprop

fvisin:
I am trying to understand how to write the C code, but it is more complicated than I expected and the documentation is not very clear (both Theano's and Numpy's I would say). I am referring to the c code of Softmax as a guideline. These are some of the lines I don't understand:
* https://github.com/fvisin/Theano/blob/b28c81c79ca44f0762f0ee5e6cbce2fefee81454/theano/tensor/nnet/nnet.py#L489-L501
    * what is sm?
    * what is this doing?
* https://github.com/fvisin/Theano/blob/b28c81c79ca44f0762f0ee5e6cbce2fefee81454/theano/tensor/nnet/nnet.py#L513-L514
    * If I am not wrong PyArray_BYTES should be the content of the ndarrray, but what is PyArray_STRIDES? I can only find tautological definitions "returns a pointer to the strides"...but I don't get what is a stride...

This would be a good start. I will keep reading and add new comments if there is something else that I don't get.

fvisin:
I sent the comment too soon, I found this definition: 
`An array of integers providing for each dimension the number of bytes that must be skipped to get to the next element in that dimension.` 

It is still not completely clear though:
* is it right that when the ndarray does not contain structured items (e.g., is an array of ints, it does not contain lists or other ndarrays) every element (dimension) of stride corresponds to the size in bytes of the dtype of the ndarray (in my example, `stride==dtype(int)`)?
* in the case of `[[1, 2, 3], [4]]`, what is the value of stride? If you consider `[1, 2, 3]` stride should be `3 * sizeof(int)`, but if you consider `[4]` it should be `sizeof(int)`? I guess I am missing something trivial, but I am tired and this new unreadable font is driving me crazy. I'll try to reverse engineer the code, it is probably easier.

abergeron:
The code looks ok, but we still have to debug why it doesn't work on Travis.  It works locally.

nouiz:
I think travis tests fail due when mode=FAST_COMPILE and when floatX=float32.

Is the opt registered in FAST_COMPILE? I think you can make the test run by forcing 'FAST_COMPILE, to be sure the opt will be applied.

abergeron:
Yes, the tests pass now!

fvisin:
Great!!

PR #3364, opened 2015-09-04
Use the libgpuarray APIs to manage GPU code compilation, execution, etc.

seanprime7:
Currently, the gpuarray back end of Theano uses a mixture of libgpuarray and nvcc, which is slow for online compilation.  We can speed up the GPU code compilation using libgpuarray API alone without nvcc, especially when libgpuarray is built with the nvrtc support (see https://github.com/Theano/libgpuarray/pull/72).

This hopefully helps managing GPU related code in Theano and libgpuarray by having vendor specific implementation only in libgpuarray.

### DIFF
--- a/theano/sandbox/gpuarray/basic_ops.py
+++ b/theano/sandbox/gpuarray/basic_ops.py
@@ -144,19 +144,21 @@ class GpuKernelBase(object):
     def _generate_kernel_vars(self, k):
         return """static GpuKernel %(kname)s;""" % dict(kname=k.objvar)
 
-    def c_support_code_apply(self, node, name):
-        ceil_intdiv = """
+    def c_support_code(self):
+        return """
         template <typename T>
         static T ceil_intdiv(T a, T b)
         {
             return (a/b) + ((a % b) ? 1: 0);
         }
         """
+
+    def c_support_code_apply(self, node, name):
         kernels = self.gpu_kernels(node, name)
         bins = '\n'.join(self._generate_kernel_bin(k) for k in kernels)
         codes = '\n'.join(self._generate_kernel_code(k) for k in kernels)
         vars = '\n'.join(self._generate_kernel_vars(k) for k in kernels)
-        return '\n'.join([ceil_intdiv, bins, codes, vars])
+        return '\n'.join([bins, codes, vars])
 
     def _generate_kernel_init(self, k, err):
         if PY3:
--- a/theano/sandbox/gpuarray/basic_ops.py
+++ b/theano/sandbox/gpuarray/basic_ops.py
@@ -145,11 +145,18 @@ class GpuKernelBase(object):
         return """static GpuKernel %(kname)s;""" % dict(kname=k.objvar)
 
     def c_support_code_apply(self, node, name):
+        ceil_intdiv = """
+        template <typename T>
+        static T ceil_intdiv(T a, T b)
+        {
+            return (a/b) + ((a % b) ? 1: 0);
+        }
+        """
         kernels = self.gpu_kernels(node, name)
         bins = '\n'.join(self._generate_kernel_bin(k) for k in kernels)
         codes = '\n'.join(self._generate_kernel_code(k) for k in kernels)
         vars = '\n'.join(self._generate_kernel_vars(k) for k in kernels)
-        return '\n'.join([bins, codes, vars])
+        return '\n'.join([ceil_intdiv, bins, codes, vars])
 
     def _generate_kernel_init(self, k, err):
         if PY3:
--- a/theano/sandbox/gpuarray/conv.cu
+++ b/theano/sandbox/gpuarray/conv.cu
@@ -203,11 +203,8 @@ PyGpuArray_conv_valid(const PyGpuArrayObject *img,
                   int, int, int, int,
                   int, int);
 
-#define CONV_PATCH_SPECIAL(kern_wid) \
-            if(threads.y==out_len) f=conv_patch<true,kern_wid,false>;\
-            else f=conv_patch<true,kern_wid,true>;
-
-        CONV_PATCH_SPECIAL(THEANO_KERN_WID);
+        if(threads.y==out_len) f=conv_patch_2;
+        else f=conv_patch_3;
 
          f<<< grid, threads, shared_size>>>
              (cuda_get_ptr(img), cuda_get_ptr(kern), cuda_get_ptr(out),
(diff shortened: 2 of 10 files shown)


nouiz:
That seem excellent idea. I'll let @abergeron review in more detail.

If all of libgpuarray do this (is that the case of there is part missing still missing?) this mean we don't need to install nvcc anymore? That would greatly help on windows, installeing VisualStudio with ncvv isn't trivial.

thanks.

seanprime7:
You will still have to install the CUDA toolkit as nvrtc is released as part of the toolkit.
And it also depends on whether/when Theano/libgpuarray#72 is pulled in.

But it still will be easier to use if libgpuarray is configured to use nvrtc.  As nvrtc is a library, not an offline tool, the environment setup is much simpler.  Even without Theano/libgpuarray#72, nvcc invocations happen only within libgpuarray, so it's a better way to abstract the use of nvcc, I believe.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -145,11 +145,18 @@ def _generate_kernel_vars(self, k):
         return """static GpuKernel %(kname)s;""" % dict(kname=k.objvar)
 
     def c_support_code_apply(self, node, name):
+        ceil_intdiv = """
+        template <typename T>
+        static T ceil_intdiv(T a, T b)
+        {
+            return (a/b) + ((a % b) ? 1: 0);
+        }
+        """
```
abergeron: This could go in c_support_code() since nothing is specific to the apply node.

Also, because I'm curious, why do you need this?

theano/sandbox/gpuarray/conv.py
```diff
@@ -3,13 +3,20 @@
...
 from six.moves import reduce
 from .comp import NVCC_compiler
 from .type import GpuArrayType
-from .basic_ops import as_gpuarray_variable
-
+from .basic_ops import (as_gpuarray_variable, GpuKernelBase, HideC, Kernel)
+from theano.gof import utils
 
-class GpuConv(gof.Op):
+class GpuConv(GpuKernelBase, HideC, gof.Op):
```
abergeron: You don't need HideC since this doesn't inherit from a CPU op.

theano/sandbox/gpuarray/conv.cu
```diff
@@ -195,46 +188,47 @@ PyGpuArray_conv_valid(const PyGpuArrayObject *img,
...
+        size_t threads_per_block[3] = {(size_t)out_wid,
+                               ceil_intdiv((size_t)out_len,(size_t)nb_split),
+                               (size_t)1};
+        size_t n_blocks[3] = {(size_t)nbatch, (size_t)nkern, (size_t)1};
+
+        size_t shmem_sz = (img_size + kern_size)*sizeof(float);
+
+        GpuKernel *k = NULL;
+        if(threads_per_block[1]==out_len) k=&conv_patch_2_node_<<<<HASH_PLACEHOLDER>>>>_0;
+        else k=&conv_patch_3_node_<<<<HASH_PLACEHOLDER>>>>_0;
```
abergeron: I will say that I am not too keen on directly using `<<<<<HASH_PLACEHOLDER>>>>>` even though it looks like there is no other solution here.

Maybe you could use the facilities of COp to remove its use? If that looks too complex, then we can keep it as-is for now.

theano/sandbox/gpuarray/conv.py
```diff
@@ -303,3 +310,266 @@ def c_code(self, node, nodename, inp, out_, sub):
...
+        {
+            return (a/b) + ((a %% b) ? 1: 0);
+        }
+        static const char conv_bcode[] = {%(bcode)s};
+        static const char *conv_code = "%(code)s";
+        """ % locals()
+        for k in kernels:
+            mod += "static GpuKernel " + k.name + '_' + name + ";\n"
+        mod += open(os.path.join(os.path.split(__file__)[0], "conv.cu")).read()
+        return mod
```
abergeron: This tells me that I should add the functionality to load multiple kernels from a single image in libgpuarray.  But it's a clever hack for now.

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -204,60 +173,64 @@ def generate_kernel(self, node, nodename):
...
-        return ['cuda.h', '<gpuarray/extension.h>', '<numpy_compat.h>',
-                '<gpuarray/ext_cuda.h>']
+        cuda_root = config.cuda.root
+        if cuda_root:
+            return [os.path.join(cuda_root, 'include')]
+        else:
+            return []
 
     def c_compiler(self):
+        return GCC_compiler
```
abergeron: This is the default, so you can probably just remove the method.

abergeron:
I'll have to say that I like this idea very much.  This was on the backburner of my todo list for a while, but now that you've done the work, I'm happy.

theano/sandbox/gpuarray/conv.py
nouiz: Can you make an issue for this and tell in it that to clean up this tmp hack? Just to be sure we don't forget to update that part.

theano/sandbox/gpuarray/conv.py
seanprime7: Okay.  I will make the change and push an update soon.

theano/sandbox/gpuarray/elemwise.py
seanprime7: Without it, GpuIncSubtensor.c_compiler() fails with

```
AttributeError: ('The following error happened while compiling the node', GpuIncSubtensor{InplaceSet;:int64:}(GpuAlloc.0, Rebroadcast{0}.0, Constant{1}), '\n', "'GpuElemwise' object has no attribute 'c_compiler'")
```
as it is defined as
```
def c_compiler(self):
    return self.iadd_node.op.c_compiler()
```

So, I think I still need this.  Do you think I am missing something?

abergeron: Ok, leave it for now, it shouldn't hurt anything.

abergeron:
This PR would fix a number of issues that cropped up due to requiring explicit context activation when performing gpu operations.

So if there isn't any movement very soon, I'll just merge it as-is and fix the few remaining issues later.

theano/sandbox/gpuarray/basic_ops.py
seanprime7: ceil_intdiv is needed because it is used by the host code to compute the grid/block configurations.

seanprime7:
* I have moved ```ceil_intdiv``` from ```c_support_code_apply``` to ```c_support_code```.
* I have not had time to find a way to remove ```<<<<<HASH_PLACEHOLDER>>>>>``` from .cu files.
* ```GpuConv``` no longer inherits ```HideC```.

I think it's good if you can merge it now and we can deal with the ```<<<<<HASH_PLACEHOLDER>>>>>``` issue later.

abergeron:
Wow, you're fast!  But thanks.

seanprime7:
I had the commits ready but was testing them with another run of unit tests, which finished on time. :)

PR #3367, opened 2015-09-08
Scan mitmot prealloc

carriepl:
This also includes a few things required for this to work : 
-Adding new tests
-Make the NoOutputFromInplace feature and optimization parameterizable
-Add a register() method to class Mode to allow including optimization instances (not only tags) to compilation modes.
-Allow usage of In() input wrappers without forcing the usage of old Theano functions
-Modify FunctionGraph to have knowledge of which of its outputs are updates for which inputs.
-Implementing a preferential ordering on the inputs when trying to make elemwise inplace (to favor output updates being inplace on the updates variables)


theano/compile/mode.py
```diff
@@ -323,6 +328,32 @@ def including(self, *tags):
...
+        -------
+        Mode
+            Copy of the current Mode which includes the provided
+            optimizations.
+        """
+
+        link, opt = self.get_linker_optimizer(self.provided_linker,
+                                              self.provided_optimizer)
+        return self.__class__(linker=link,
+                              optimizer=opt.register(*optimizations))
```
abergeron: Plz use clone instead of `__class__`.

theano/compile/function.py
```diff
@@ -280,7 +280,7 @@ def opt_log1p(node):
                                    (hasattr(i, 'mutable') and i.mutable))):
             check_for_aliased_inputs = True
```
nouiz: Add a quick test for this.

theano/compile/mode.py
```diff
@@ -170,9 +170,14 @@ class AddNoOutputFromInplace(gof.Optimizer):
     outputs
 
     """
+
+    def __init__(self, first_output_idx=0, last_output_idx=None):
+        self.feature = gof.NoOutputFromInplace(first_output_idx,
```
nouiz: What about a quick generalization of this. Make this opt take the feature as input and rename it to AddFeatuerOptimizer?

theano/gof/optdb.py
```diff
@@ -186,9 +192,9 @@ def __init__(self, include, require=None, exclude=None,
 
     def __str__(self):
         return ("Query{inc=%s,ex=%s,require=%s,subquery=%s,"
-                "position_cutoff=%d}" %
+                "position_cutoff=%d,extra_opts=%d}" %
```
nouiz: use %s and not %d

theano/gof/optdb.py
```diff
@@ -306,19 +321,32 @@ def query(self, *tags, **kwtags):
 
         position_cutoff = kwtags.pop('position_cutoff',
                                      config.optdb.position_cutoff)
+        position_dict = self.__position__.copy()
```
nouiz: don't make the copy here, but only in the if bellow to prevent useless copy.

theano/gof/toolbox.py
```diff
@@ -440,10 +440,21 @@ def on_change_input(self, fgraph, node, i, r, new_r, reason=None):
...
+        self.first_idx = first_output_idx
+        self.last_idx = last_output_idx
+
     def validate(self, fgraph):
         if not hasattr(fgraph, 'destroyers'):
             return True
-        for out in list(fgraph.outputs):
+
+        if self.last_idx is None:
+            outputs_to_validate = list(fgraph.outputs)[self.first_idx:]
```
nouiz: Always using the else condition is good. By default, None as the stop of a slice mean the full slice.

theano/tensor/opt.py
```diff
@@ -320,7 +320,50 @@ def inplace_elemwise_optimizer(fgraph):
...
+                # If the output of the node can be established as an update
+                # output of the fgraph, visit the candidate_inputs in an order
+                # that will improve the chances of making the node operate
+                # inplace on the input it's meant to update
+                candidate_out_var = node.outputs[candidate_output]
+                sorted_candidate_inputs = candidate_inputs
+                if fgraph.update_mapping:
+
+                    update_outs = [fgraph.outputs[i]
+                                   for i in fgraph.update_mapping]
```
nouiz: move those 2 lines before the 2 for loop.

theano/tensor/opt.py
```diff
@@ -320,7 +320,50 @@ def inplace_elemwise_optimizer(fgraph):
...
+                        vars_from_inplace = []
+                        other_vars = []
+                        for inp_idx in candidate_inputs:
+                            inp = node.inputs[inp_idx]
+                            if inp is updated_inp:
+                                updated_vars.append(inp_idx)
+                            elif (hasattr(fgraph, 'destroy_handler') and
+                                  inp.owner and
+                                  updated_inp in fgraph.destroy_handler.root_destroyer and
+                                  fgraph.destroy_handler.root_destroyer[updated_inp] == inp.owner):
```
nouiz: I think "is" would be better then "==" here.

theano/tensor/opt.py
```diff
@@ -320,7 +320,50 @@ def inplace_elemwise_optimizer(fgraph):
...
+
+                    update_outs = [fgraph.outputs[i]
+                                   for i in fgraph.update_mapping]
+
+                    if candidate_out_var in update_outs:
+
+                        # The candidate output is an update. Sort the
+                        # variables in candidate_inputs in the following order:
+                        # - Vars corresponding to the actual updated input
+                        # - Vars computed inplace on the updates input
```
nouiz: Those 2 lines are very short. It took me a few minutes to remember the second case. Can you explain more why this is needed? To help the next reader of that method.

nouiz:
I just reviewed the code that isn't scan. travis is failing.

carriepl:
I think this is it; I've done the changes requested and, at long last, Travis is not complaining anymore.

theano/tensor/opt.py
```diff
@@ -320,7 +325,56 @@ def inplace_elemwise_optimizer(fgraph):
...
+                    vars_from_inplace = []
+                    other_vars = []
+                    for inp_idx in candidate_inputs:
+                        inp = node.inputs[inp_idx]
+                        if inp in updated_inputs:
+                            updated_vars.append(inp_idx)
+                        elif (hasattr(fgraph, 'destroy_handler') and
+                              inp.owner and
+                              any([(up_inp in fgraph.destroy_handler.root_destroyer and
+                                    fgraph.destroy_handler.root_destroyer[up_inp] is inp.owner)
```
nouiz: can you comment those 2 lines? They are very hard to understand. But I think it is good. In particular, root_destroyer isn't well documented in destroy handler

This can be be re-writen for clarity like this:

fgraph.destroy_handler.root_destroyer.get(up_inp, None) is inp.owner

theano/compile/mode.py
```diff
@@ -333,7 +353,7 @@ def requiring(self, *tags):
                                               self.provided_optimizer)
         return self.__class__(linker=link, optimizer=opt.requiring(*tags))
 
-    def clone(self, link_kwargs=None, **kwargs):
+    def clone(self, link_kwargs={}, optimizer=None, **kwargs):
```
abergeron: Don't use a mutable structure as a default value for an argument.  This is asking for trouble.

theano/compile/mode.py
```diff
@@ -333,7 +353,7 @@ def requiring(self, *tags):
                                               self.provided_optimizer)
         return self.__class__(linker=link, optimizer=opt.requiring(*tags))
```
abergeron: While there, please update including() excluding() and requiring() to use clone().

theano/compile/mode.py
```diff
@@ -323,6 +319,30 @@ def including(self, *tags):
...
+            position at which to insert the optimization in the mode.
+
+        Returns
+        -------
+        Mode
+            Copy of the current Mode which includes the provided
+            optimizations.
+        """
+
+        new_optimizer = self.provided_optimizer.register(*optimizations)
```
abergeron: I'm pretty sure provided_optimizer can be a string which does not have the register() method.  You should probably use get_linker_optimizer()

theano/compile/mode.py
abergeron: Also, `None` is a valid value for optimizer.

theano/compile/monitormode.py
```diff
@@ -80,6 +80,12 @@ def including(self, *tags):
         ret.post_func = self.post_func
         return ret
 
+    def register(self, *optimizations):
+        ret = super(MonitorMode, self).register(*optimizations)
+        ret.pre_func = self.pre_func
+        ret.post_func = self.post_func
+        return ret
```
abergeron: This is something clone() should handle.

theano/gof/fg.py
```diff
@@ -109,7 +109,19 @@ class FunctionGraph(utils.object2):
...
+        """
+        Create an FunctionGraph which operates on the subgraph bound by the
+        inputs and outputs sets.
+
+        :param inputs: inputs nodes of the graph, usually declared by the user
+        :param outputs: outputs nodes of the graph.
+        :param clone: If true, we will clone the graph. This is
+        useful to remove the constant cache problem.
+        :param update_mapping: dictionnary describing the mapping between
+        the inputs with updates and the outputs corresponding to their updates
```
abergeron: Wrong format for doc.

theano/gof/optdb.py
```diff
@@ -173,12 +178,13 @@ class Query(object):
     """
 
     def __init__(self, include, require=None, exclude=None,
-                 subquery=None, position_cutoff=None):
+                 subquery=None, position_cutoff=None, extra_optimizations=[]):
```
abergeron: Do not use a mutable object as default value.

theano/gof/optdb.py
```diff
@@ -306,19 +326,34 @@ def query(self, *tags, **kwtags):
...
             # the call to super should have raise an error with a good message
             assert len(tags) == 1
             if getattr(tags[0], 'position_cutoff', None):
                 position_cutoff = tags[0].position_cutoff
 
-        opts = [o for o in opts if self.__position__[o.name] < position_cutoff]
+            # The Query instance might contain extra optimizations which need
+            # to be added the the sequence of optimizations (don't alter the
+            # original dictionary)
+            position_dict = position_dict.copy()
```
abergeron: This will always copy the dict even if we have nothing to add to it.

theano/gof/optdb.py
```diff
@@ -306,19 +326,34 @@ def query(self, *tags, **kwtags):
...
+                position_dict[opt.name] = position
+
+        opts = [o for o in opts if position_dict[o.name] < position_cutoff]
         # We want to sort by position and then if collision by name
         # for deterministic optimization.  Since Python 2.2, sort is
         # stable, so sort by name first, then by position. This give
         # the order we want.
         opts.sort(key=lambda obj: obj.name)
-        opts.sort(key=lambda obj: self.__position__[obj.name])
+        opts.sort(key=lambda obj: position_dict[obj.name])
```
abergeron: If you're doing key sort anyway, you can make the sort key a tuple of (position, name) and get the right order in one go.  This will also be faster.

theano/gof/optdb.py
```diff
@@ -306,19 +326,34 @@ def query(self, *tags, **kwtags):
...
+            # original dictionary)
+            position_dict = position_dict.copy()
+            for extra_opt in tags[0].extra_optimizations:
+                # Give a name to the extra optimization
+                opt, position = extra_opt
+                opt.name = str(opt.__class__)
+
+                # Add the extra optimization to the optimization sequence
+                opts.add(opt)
+                position_dict[opt.name] = position
```
abergeron: Is it acceptable to remap existing ops to a new position?  Because in case the class name clashes, this is what it'll do.

abergeron: As an optimization could you skip adding the opt if the position cutoff is smaller than the desired position.

theano/scan_module/scan_op.py
```diff
@@ -746,17 +745,91 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
...
+            self.mitmots_preallocated = [False] * self.n_mit_mot_outs
+
             wrapped_inputs = [Param(x, borrow=True) for x in
                               self.inputs]
             wrapped_outputs = [Out(x, borrow=False) for x in
                                self.outputs[:slices]]
-        wrapped_outputs += self.outputs[slices:]
+            wrapped_outputs += self.outputs[slices:]
+
+            compilation_mode = self.mode_instance
```
abergeron: Why store this here?

theano/compile/mode.py
carriepl: I just read how Python handles mutable default values for arguments. I think this is a really counter-intuitive and simply dangerous way of handling default values but, alas, those are the rules we must play by.

theano/scan_module/scan_op.py
carriepl: Why not? It matches the assignment on line 818 so when compiling the function, I don't need to add a condition and duplicate the call to function(), I can just pass the mode as `compilation_mode` whether some outputs have been preallocated or not.

abergeron: Ah, I had not seen the other assignment.  Ok then.

abergeron:
With the new changes I'm ok with merging this when travis passes.

carriepl:
I removed the comment and kill Travis on that latest commit since the only change is a comment.

nouiz:
To my knowledge, the only missing thing is the tests for IN.

theano/compile/tests/test_function.py
```diff
@@ -26,3 +28,160 @@ def test_function_dump():
...
+        # This is not a test of the In class per se, but the In class relies
+        # on the fact that shared variables cannot be explicit inputs
+        a = theano.shared(1.0)
+        self.assertRaises(TypeError, theano.function, [a], a + 1)
+
+    def test_in_shared_variable(self):
+        # Ensure that an error is raised if the In wrapped is used to wrap
+        # a shared variable
+        a = theano.shared(1.0)
+        a_wrapped = In(a, update=a + 1)
```
nouiz: also test that an update with a different type then the original variable raise an error.

nouiz:
I found one missing test to add. I'm ok with this. We need to wait for @lamblin to review this.

lamblin:
It makes sense to me. I did not review in detail, but all seems reasonable. I'm OK to merge if everything passes.

PR #3372, opened 2015-09-09
OpenCL support on Theano.

Nirvik-D:
Created a new OpenCL convolution kernel. Its optimized with local memory, float4, constant and 16x16 local workgroup size. The optimizations are derived at https://www.evl.uic.edu/kreda/gpu/image-convolution/ . Please check the site for more information. Also, kindly check whether any new changes can be made.

For now, the To do list is as follows.

- [ ] Initial work on cl_corr_gemm

      - [x] Port `im2col` and `col2im` to OpenCL.
      - [ ] Implement GEMM on OpenCL.

### DIFF
--- a/theano/sandbox/gpuarray/col2im_kernel.cl
+++ b/theano/sandbox/gpuarray/col2im_kernel.cl
@@ -1,0 +1,61 @@
+/*
+Copyright (c) 2014, The Regents of the University of California (Regents)
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met: 
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer. 
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution. 
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+__kernel void col2im_kernel ( const int num_kernels, __global float * data_col,
+    const size_t height, const size_t width, const size_t channels,
+    const size_t patch_h, const size_t patch_w,
+    const size_t pad_h, const size_t pad_w,
+    const ptrdiff_t stride_h, const ptrdiff_t stride_w,
+    const size_t height_col, const size_t width_col,
+    __global float * data_im )
+{
+    for (int index = get_global_id(0); index < num_kernels; index += get_global_size(0))
+    {
+        float val = (float) 0.0;
+    	int w = index % width + pad_w;
+    	int h = (index / width) % height + pad_h;
+    	int c = index / (width * height);
+    	// compute the start and end of the output
+    	int w_col_start = (w < patch_w) ? 0 : (w - patch_w) / stride_w + 1;
+    	int w_col_end = min(w / stride_w + 1, width_col);
+    	int h_col_start = (h < patch_h) ? 0 : (h - patch_h) / stride_h + 1;
+    	int h_col_end = min(h / stride_h + 1, height_col);
+	// equivalent implementation
+    	int offset =
+            (c * patch_h * patch_w + h * patch_w + w) * height_col * width_col;
+    	int coeff_h_col = (1 - stride_h * patch_w * height_col) * width_col;
+    	int coeff_w_col = (1 - stride_w * height_col * width_col);
+    	for (int h_col = h_col_start; h_col < h_col_end; ++h_col) 
+	{
+     	    for (int w_col = w_col_start; w_col < w_col_end; ++w_col) 
+	    {
+        	val += data_col[offset + h_col * coeff_h_col + w_col * coeff_w_col];
+      	    }
+   	}    	
+	data_im[index] = val;
+	barrier(CLK_GLOBAL_MEM_FENCE);
+    }
+}
+
--- a/theano/sandbox/gpuarray/im2col_kernel.cl
+++ b/theano/sandbox/gpuarray/im2col_kernel.cl
@@ -1,0 +1,66 @@
+/*
+Copyright (c) 2014, The Regents of the University of California (Regents)
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met: 
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer. 
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution. 
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+__kernel void im2col_kernel(const int num_kernels, __global float * data_im,
+    const size_t height, const size_t width, const size_t kernel_h, const size_t kernel_w,
+    const size_t pad_h, const size_t pad_w,
+    const ptrdiff_t stride_h, const ptrdiff_t stride_w,
+    const size_t height_col, const size_t width_col,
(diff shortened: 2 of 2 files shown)


Nirvik-D:
@nouiz @abergeron I created this pull request to maintain the code for `OpenCL` versions of `gpuarray Ops` under `sandbox`. Once, its done, I am going to transfer it to `libgpuarray`. Will it be okay to do that?

abergeron:
If the ultimate goal is to move this in libgpuarray, the PRs should be made against libgpuarray directly.

Also, I try (whenever possible) to maintain a single copy of kernel code by using the cluda macros defined at the top of the Cuda and OpenCL backend files.  Note that these are probably not complete and could certainly be extended to support new things (like perhaps local/shared memory).

Other than that, I am pleased that somebody is undertaking the task of beefing up OpenCL support.  I can't spend as much time as I would like on it so thanks for fill up the holes.

Nirvik-D:
@abergeron Since I have just worked partially on the OpenCL code, I am keeping the new codes inside sandbox. Once, the code is tested  and nearly complete, it would be a good idea to move the files over to libgpuarray and either close this PR or maybe merge it with the libgpuarray PR? What do you think?

Also, can you have a look at the kernel that I updated? If there are any specifics that I need to address, since the kernel is for a single image convolution only. For batch convolution, I would require guidance.

nouiz:
I'm in the bus without WiFi, so I can't check the code. For OpenCL
convolution we should use the gpucorrmm in cuda as a base. It will be
faster then porting the old conv. Also there is a pr that update the new
backbend port of our old convolution.

To help guide the work, can you do a to-do list that we comment on? That
way you can start each step at your time line but still have some guidance?

Nirvik-D:
@nouiz Ok, I will look up on gpucormm and the PR. I will let you know when its done.

Nirvik-D:
@abergeron @nouiz Have you tested ViennaCL for Theano? Seems like there is no available code for matrix to Toeplitz conversion in CL.

nouiz:
That conversion is done in the gpucorrmm ops itselfs. It isn't a separate
funciton.

Also, the web site you look to is old (2012) on the convolution
implementation. I don't know if it is faster then gpucorrmm, but my guess
is that gpucorrmm is faster. But without benchmark, it is hard to tell.

Nirvik-D:
@nouiz Ok thanks. I will look into it.

Nirvik-D:
@nouiz @abergeron 

Hi,
So far I am done porting im2col and col2im for the corr_gemm kernel to run on OpenCL. Now I am dealing with the convolution operation. Any hints on how I should go about applying GEMM?

nouiz:
Just use the libgpuarray blas gemm call? It will hide the CUDA/OpenCL
difference.

Also, those 3 news op should work for both cuda and opencl at the same
time. We have all we need for this I think.

Nirvik-D:
@nouiz  
I am ready to implement the libgpuarray GEMM call, but when I looked at corr_gemm.cu, I saw that you used CUDAndarray. Now since, there is no equivalent for CUDAndarray in OpenCL, so I am unsure as to what I should use for implementing the blas gemm. Also, can you please tell me which new Ops I can implement?

Thanks a lot.

Nirvik-D:
reopen

nouiz:
The class I pointed use the old back-end. So maybe the first step is to
move it to the new back-end with CUDA only support. Then update it again to
support OpenCL:

https://github.com/Theano/Theano/issues/1471


Those instruction aren't the competly updated, so don't follow them to
closely, but it give the directly.

Fred

Nirvik-D:
@nouiz 

Sounds like a good idea. But without Nvidia hardware, I wouldn't be able to go on much with it. Still, I will try to work through the Ops.

Nirvik-D:
@nouiz 

And by the way, by the old GPU back-end, which back-end are you talking about? I am still confused about it. I mean, is CUDAndarray the old gpu backend?

theano/sandbox/gpuarray/corr_gemm.cl
```diff
@@ -0,0 +1,132 @@
...
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#undef _GLIBCXX_ATOMIC_BUILTINS
```
abergeron: You shouldn't need this.

theano/sandbox/gpuarray/corr_gemm.cl
```diff
@@ -0,0 +1,132 @@
...
+        	*data_col_ptr = (h >= 0 && w >= 0 && h < height && w < width) ?
+            	data_im_ptr[i * width + j] : 0;
+        	data_col_ptr += height_col * width_col;
+      	    }
+    	}
+    	barrier(CLK_LOCAL_MEM_FENCE | CLK_GLOBAL_MEM_FENCE);
+    }
+}
+
+void im2col(const __global float4 * data_im, const int channels,
```
abergeron: If I understand correctly this is the entry point, so it should get the __kernel qualifier.

abergeron: Or rather, the porting will have to do something different since it appears you tried to use cuda-style inline launches which is not supported by OpenCL.  The code that calls the kernel has to be separate from the code that is the kernel.

theano/sandbox/gpuarray/corr_gemm.cl
```diff
@@ -0,0 +1,132 @@
...
+#undef _GLIBCXX_ATOMIC_BUILTINS
+
+__kernel void im2col_kernel(const int n, const __global float4 * data_im,
+    const int height, const int width, const int kernel_h, const int kernel_w,
+    const int pad_h, const int pad_w,
+    const int stride_h, const int stride_w,
+    const int height_col, const int width_col,
+    __global float4 * data_col) 
+{
+    for (int i = get_global_id(0); i < (n); i += get_global_size(0))
```
abergeron: I'm worried about what happens if the data is not float4-aligned (which is our usual case, we have float data).

Also what if the number of elements is not divisible by 4?

Using float4 makes reads and writes more efficient, but the rest of the code more complex to handle corner cases.  I think having a first version that works on plain floats is a better idea.  We can speed it up later.

theano/sandbox/gpuarray/corr_gemm.cl
```diff
@@ -0,0 +1,132 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#undef _GLIBCXX_ATOMIC_BUILTINS
+
+__kernel void im2col_kernel(const int n, const __global float4 * data_im,
+    const int height, const int width, const int kernel_h, const int kernel_w,
+    const int pad_h, const int pad_w,
```
abergeron: Please size_t/ptrdiff_t for the sizes/strides.  ints will overflow for bigger arrays.

theano/sandbox/gpuarray/corr_gemm.cl
Nirvik-D: @abergeron 
I had a doubt at that point as to whether write the kernel and the kernel calling function in separate files or not. So let me get this straight, we need to use the __kernel qualifier in the im2col and col2im functions as they will be invoked first, right? And the im2col_kernel and col2im_kernel needs to invoked as another kernel with the __kernel qualifier? So basically we are going to Dynamic parallelism right?

theano/sandbox/gpuarray/corr_gemm.cl
Nirvik-D: @abergeron 
Alright. I will switch them to float. Although, float32 and float64 could work, if we use __kernel ... <vec_type_hint(float4)>. I believe the float32 would be divided accordingly. But of course for now, float would be more suitable.

theano/sandbox/gpuarray/corr_gemm.cl
Nirvik-D: @abergeron 

Ok, I am going to get that removed.

Nirvik-D:
@abergeron 

It seems Dynamic Parallelism would only be applicable for OpenCL 2.0. So what we can do is in order to support earlier versions, we can simply implement the code of the im2col_kernel inside im2col function. That way, we needn't apply a kernel function call.

Nirvik-D:
@abergeron @nouiz 

Made the modifications. I am still confused about the get_global_id(0) addition. Can you have a look and verify?

abergeron:
I'm really sorry if this comes out as sort of rude it's hard to convey tone over the internet.  I appreciate all help we can get.

It looks like you have minimal or no experience doing GPU programming with OpenCL. So I will attempt to explain where you went wrong.

To do something in OpenCL you write what are called kernels which look a lot like C functions and are tagged with __kernel.  Then this code is placed into a string in your program. You compile it using the OpenCL api (or a wrapper, like libgpuarray) and then you obtain an object which you can use to make calls to that kernel.

The cuda model is a bit different since you can mixed kernel and host code in the same file and the compiler will separate the two and insert the appropriate calls to launch the kernels on the GPU.

The original code uses this facility and mixes host and kernel code which is not supported by OpenCL.  So you will have to adapt the way the code is structured to make this work. Which means most likely splitting the code into kernel and host parts.

In this specific example im2col_kernel was the kernel part and im2col was the host part, which includes code to launch the kernel.  Merging the two together is a bad idea.

Finally I encourage you to look at resources for programming in OpenCL if you still have some questions.  I can answer some questions, but I can do a whole training session on how to develop for OpenCL over the internet.

Nirvik-D:
@abergeron 

Got it. I am really thankful for your advice. I admit that I am still a novice at OpenCL and I am trying to learn as much as I can through the process, while building. I will continue working on the kernels, so that I can gain some more knowledge and code at the same time.

I am consulting as much online help as I can and I am sorry about giving you a tough time with the kernels. I will try my level best to work on them. Once again thanks for the advice.

Now, regarding the host code, is PyArray being used in place of CUDAndarray? I saw the PyArray statements in the blas.c file under libgpuarray. So, I was wondering if I can use the PyArray_STRIDES and PyArray_DIM in place of CUDAndarray_HOST_STRIDES and CUDAndarray_HOST_DIMS.

nouiz:
PyArray is the CPU ndarray. CUDAndarray is the old back-end object.

Somewhere, there is some macro similar to PyArray_STRIDES for the new
back-end like PyGpuArray_STRIDES. They are in the
filelibgpuarray/pygpu/numpy_compat.h.

Nirvik-D:
@nouiz 

Ok, Thank you. I will check it out.

Nirvik-D:
@abergeron @nouiz 

Hi,
I made some modifications to the `im2col` and `col2im` kernels. Please have a look at them. Also, the first commit is messed up. The proper changes are made in the second commit.

nouiz:
I think you push frequently modif to this PR. Can you do it at most once or twice per day? Each push, even forced push trigger a new test run by travis and this is slow. So if we queue to much change, it create a backlog. We need to find a way to make Theano tests faster, but in the mean time if we don't to too much update it is working fine.

I canceled the tests for commit that got amended.

nouiz:
You starting to learn all OpenCL in the hard way if you start with the gpucorrmm. I think it would easy your learning if you start by an op already in the new back-end but with only the CUDA code. What do you think of that?

in issue, I see that there is 2 candidates for this:

CrossentropySoftmax1HotWithBiasDx and CrossentropySoftmaxArgmax1HotWithBias

What do you think of that? If you port code from the old back-end you need to also learn how to read and understand the old back-end code. By this approch, you can concentrate on the OpenCL vs CUDA stuff. As you don't have a CUDA capable GPUs, I think it would be useful too.

Nirvik-D:
@nouiz 

I usually do very few pushes a day. But the last commits go horribly messed up, plus I also made some errors in the kernels. Well, to be more accurate, some memory management problems. But they are fixed for now. This is why, I had a lot of pushes yesterday.

And thank you for the information. It would certainly help to make things easier for learning at the same time. I will check them out.

nouiz:
As this PR isn't well advanced and is stale, I'll close it. If someone continu this PR, open a new PR.

PR #3374, opened 2015-09-09
Whitespace needed in theano_vs_c.txt

lucasb-eyer:
It screwed up the [rendered formatting](http://deeplearning.net/software/theano/extending/theano_vs_c.html).


nouiz:
thanks

PR #3375, opened 2015-09-09
Scan Speedup Compilation #2

### LINKED ISSUE #3346: Speed up 2 scan opt
PushOutSeqScan and PushOutNonSeqScan.

Don't go over the toposort.

Create a dequeue object and init it with local_fgraph.inputs.

When you move one node outside, append its client to that dequeue. This will make that optimization just check the start of the inner fgraph and not all of it. It should make this faster.

caglar:
Fixes #3346 .

### DIFF
--- a/theano/scan_module/scan_opt.py
+++ b/theano/scan_module/scan_opt.py
@@ -329,9 +329,9 @@ class PushOutNonSeqScan(gof.Optimizer):
 
                 # Add the nodes only if they are not already
                 # in the bookkeeper's list
-                for nd_cli in node_clients:
-                    if nd_cli not in local_fgraph_bookkeeper:
-                        local_fgraph_bookkeeper.append(nd_cli)
+                nd_cli_cleaned = filter(lambda x: x not in local_fgraph_bookkeeper,
+                                        node_clients)
+                local_fgraph_bookkeeper.extend(nd_cli_cleaned)
 
                 outside_ins = []
                 for x in nd.inputs:
@@ -363,7 +363,6 @@ class PushOutNonSeqScan(gof.Optimizer):
                     replace_with_in.append(y_place_holder)
                     assert isinstance(y, type(nw_outer_node.outputs[idx]))
                     replace_with_out.append(nw_outer_node.outputs[idx])
-
 
         # We need to check all candidate replacements and choose those that
         # make sense for us
@@ -485,7 +484,6 @@ class PushOutSeqScan(gof.Optimizer):
         local_fgraph_inps = local_fgraph.inputs
         local_fgraph_bookkeeper = deque(retrieve_clients_fromnodes(local_fgraph_inps))
 
-
         local_fgraph_outs_set = set(local_fgraph.outputs)
         local_fgraph_outs_map = dict([(v,k) for k,v in \
                                      enumerate(local_fgraph.outputs)])
@@ -519,8 +517,10 @@ class PushOutSeqScan(gof.Optimizer):
         assert len(inner_non_seqs) == len(outer_non_seqs)
         assert len(inner_seqs) == len(outer_seqs)
 
+
         while local_fgraph_bookkeeper:
             nd = local_fgraph_bookkeeper.pop()
+
             if (nd not in to_remove_set and
                 all([(x in inner_non_seqs_set) or
                      (x.owner in to_remove_set) or
@@ -560,14 +560,13 @@ class PushOutSeqScan(gof.Optimizer):
                     # scan.
                     continue
 
-
                 node_clients = retrieve_clients_fromnodes(nd)
 
                 # Add the nodes only if they are not already
                 # in the bookkeeper's list
-                for nd_cli in node_clients:
-                    if nd_cli not in local_fgraph_bookkeeper:
-                        local_fgraph_bookkeeper.append(nd_cli)
+                nd_cli_cleaned = filter(lambda x: x not in local_fgraph_bookkeeper,
+                                        node_clients)
+                local_fgraph_bookkeeper.extend(nd_cli_cleaned)
 
                 # Add the node to remove list
                 to_remove_set.add(nd)
@@ -634,7 +633,7 @@ class PushOutSeqScan(gof.Optimizer):
         clean_replace_with_out = []
 
         existent_nodes = [nd for nd in local_fgraph_topo
-                          if nd not in to_remove_set]
+                        if nd not in to_remove_set]
         existent_nodes_set = set(existent_nodes)
 
         to_keep_set = set([])
@@ -643,10 +642,10 @@ class PushOutSeqScan(gof.Optimizer):
 
         for out, idx in to_replace_map.items():
             if (out in to_keep_set
-               and out.owner not in existent_nodes_set
-               # If types are different, conversion Op will be inserted,
-               # and it may trigger an infinite loop.
-               and replace_with_in[idx].type == out.type):
+            and out.owner not in existent_nodes_set
+            # If types are different, conversion Op will be inserted,
+            # and it may trigger an infinite loop.
+            and replace_with_in[idx].type == out.type):
 
                 clean_to_replace.append(out)
                 clean_replace_with_in.append(replace_with_in[idx])
@@ -658,8 +657,8 @@ class PushOutSeqScan(gof.Optimizer):
             nw_outer = []
             nw_inner = []
             for to_repl, repl_in, repl_out in zip(clean_to_replace,
-                                                  clean_replace_with_in,
-                                                  clean_replace_with_out):
+                                                clean_replace_with_in,
+                                                clean_replace_with_out):
                 if isinstance(repl_out, theano.Constant):
                     repl_in = repl_out.clone()
                 else:
@@ -678,7 +677,7 @@ class PushOutSeqScan(gof.Optimizer):
             nwScan = scan_op.Scan(op_ins, op_outs, nw_info)
             # Do not call make_node for test_value
             nw_node = nwScan(*(node.inputs[:1] + nw_outer + node.inputs[1:]),
-                             **dict(return_list=True))[0].owner
+                            **dict(return_list=True))[0].owner
 
             fgraph.replace_all_validate_remove(
                 list(zip(node.outputs, nw_node.outputs)),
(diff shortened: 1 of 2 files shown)


caglar:
@carriepl and @nouiz, this is the initial version of the PR. Please let me know if there is anything that needs to change.

theano/scan_module/scan_opt.py
```diff
@@ -207,6 +209,32 @@ def remove_constants_and_unused_inputs_scan(node):
...
+        else:
+            for ncli in node.clients:
+                if isinstance(ncli[0], list) or isinstance(ncli[0], tuple):
+                    if ncli[0][0] != "output" and ncli[0][0] != node:
+                        clients.append(ncli[0][0])
+                else:
+                    if ncli[0] != "output" and ncli[0] != node:
+                        clients.append(ncli[0])
+
+    return clients
```
carriepl: This function can be made much simpler : 
- A node can never be a client of one of it's own inputs.
- A variable's clients is always a list of tuples. The first element of each tuple will be either 'output' or an apply node
- There is much code duplication; the code can be refactored for readability

theano/scan_module/scan_opt.py
```diff
@@ -246,6 +274,9 @@ def process_node(self, fgraph, node):
                                          clone=False)
 
         local_fgraph_topo = local_fgraph.toposort()
+        local_fgraph_inps = local_fgraph.inputs
+        local_fgraph_bookkeeper = deque(retrieve_clients_fromnodes(local_fgraph_inps))
+
```
carriepl: With this initialization, you can have duplicates in your deque

theano/scan_module/scan_opt.py
```diff
@@ -294,6 +326,14 @@ def add_to_replace(y):
                 # We have a candidate node to removable
                 # Step 1. Reconstruct it on outside
                 to_remove_set.add(nd)
+                node_clients = retrieve_clients_fromnodes(nd)
+
+                # Add the nodes only if they are not already
+                # in the bookkeeper's list
+                nd_cli_cleaned = filter(lambda x: x not in local_fgraph_bookkeeper,
+                                        node_clients)
+                local_fgraph_bookkeeper.extend(nd_cli_cleaned)
```
carriepl: if a client occurs more than once in node_clients, you will end up with those same duplicates in your deque

carriepl: Also, now you are adding to the right of your deque and then popping, also from the right. This amounts to doing depth-first search. You should add to the left and pop from the right => this will amount to breadth-first search

carriepl: Also, since your constantly checking whether a client is in local_fgraph_bookkeeper, it might be advantageous to consider also having a Set that contains the same elements as local_fgraph_bookkeeper to speed up these containement checks. In a deque, containement check is O(N), in a set it's o(1). To investigate.

theano/scan_module/scan_opt.py
```diff
@@ -443,6 +483,9 @@ def process_node(self, fgraph, node):
         local_fgraph = gof.FunctionGraph(clean_inputs, clean_outputs,
                                          clone=False)
         local_fgraph_topo = local_fgraph.toposort()
+        local_fgraph_inps = local_fgraph.inputs
+        local_fgraph_bookkeeper = deque(retrieve_clients_fromnodes(local_fgraph_inps))
+
```
carriepl: Same as in the other optimization

theano/scan_module/scan_opt.py
```diff
@@ -568,7 +635,7 @@ def add_to_replace(y):
         clean_replace_with_out = []
 
         existent_nodes = [nd for nd in local_fgraph_topo
-                          if nd not in to_remove_set]
+                        if nd not in to_remove_set]
```
carriepl: This change breaks the indentation

theano/scan_module/scan_opt.py
```diff
@@ -577,10 +644,10 @@ def add_to_replace(y):
...
         for out, idx in to_replace_map.items():
             if (out in to_keep_set
-               and out.owner not in existent_nodes_set
-               # If types are different, conversion Op will be inserted,
-               # and it may trigger an infinite loop.
-               and replace_with_in[idx].type == out.type):
+            and out.owner not in existent_nodes_set
+            # If types are different, conversion Op will be inserted,
+            # and it may trigger an infinite loop.
+            and replace_with_in[idx].type == out.type):
```
carriepl: This change breaks the indentation

theano/scan_module/scan_opt.py
```diff
@@ -592,8 +659,8 @@ def add_to_replace(y):
             nw_outer = []
             nw_inner = []
             for to_repl, repl_in, repl_out in zip(clean_to_replace,
-                                                  clean_replace_with_in,
-                                                  clean_replace_with_out):
+                                                clean_replace_with_in,
+                                                clean_replace_with_out):
```
carriepl: This change breaks the indentation

theano/scan_module/scan_opt.py
```diff
@@ -612,7 +679,7 @@ def add_to_replace(y):
             nwScan = scan_op.Scan(op_ins, op_outs, nw_info)
             # Do not call make_node for test_value
             nw_node = nwScan(*(node.inputs[:1] + nw_outer + node.inputs[1:]),
-                             **dict(return_list=True))[0].owner
+                            **dict(return_list=True))[0].owner
```
carriepl: This change breaks the indentation

theano/scan_module/scan_opt.py
```diff
@@ -620,8 +687,8 @@ def add_to_replace(y):
                 reason='scanOp_pushout_seqs_ops')
             return True
         elif (not to_keep_set and
-              not op.as_while and
-              not op.outer_mitmot(node)):
+            not op.as_while and
+            not op.outer_mitmot(node)):
```
carriepl: This change breaks the indentation

theano/scan_module/scan_opt.py
caglar: Yeah, but in that case I would need to use an OrderedSet(which doesn't exist in python itself, but there are other ways to implement it), which is not that efficient either. I think, the size of the bookkeeper shouldn't be very large, thus N in O(N) is not very large probably less than the constant in O(1). But probably worth trying out.

theano/scan_module/scan_opt.py
caglar: Hmm, I guess I mistakenly break those during the rebase.

theano/scan_module/scan_opt.py
nouiz: I think we have an ordered set in Theano

caglar: Hmm ok, I didn't know that, I will try that one. Thanks!

carriepl: Actually, I was talking about having a Set along with the deque but, if we already have an OrderedSet set in Theano, that's even better.

carriepl:
What's the drop in the time it takes with your latest version? Also, have you started the version using an OrderedSet? Finally, there are still the indentation errors I previously pointed out.

caglar:
@carriepl @nouiz it seems like some tests are failing, interestingly when I run those tests locally they do not fail. Do yo have any idea on what might have gone wrong?

carriepl:
I'm not 100% certain what the root cause is but it only seems to fail in python 3.3 (which would explain why it passes locally; we have python 2.7.10 installed locally, not 3.3)

theano/scan_module/scan_opt.py
```diff
@@ -565,20 +630,18 @@ def add_to_replace(y):
         clean_replace_with_in = []
         clean_replace_with_out = []
 
-        existent_nodes = [nd for nd in local_fgraph_topo
-                          if nd not in to_remove_set]
+        existent_nodes = filter(lambda n: n not in to_remove_set, local_fgraph_topo)
```
nouiz: Here, remove local_fgraph_topo from this opt, it isn't needed. Use local_fgraph.apply_nodes here.

caglar: OK done that.

caglar:
@carriepl the problem seems to be due to the fact that filter function in python 3 now returns a iterable python 3 object unlike python 2 which returns a list. I have just fixed this.

caglar:
@nouiz will that change work for you:

        elif isinstance(key, collections.Sequence):
            lkey = list(key)
            for l in reversed(key):
                self.__appendleft(l)
	elif isinstance(key, collections.Iterable):
	    raise ValueError
        else:
            self.__appendleft(key) 

So if an iterator is a Sequence it has to be deterministic, if it is not and iterable we raised an error. I like doing this check here. Because I think in terms of UX, it is more accessible. I don't have to worry about changing the output of the iterable to a list before adding to an OrderedSet.

nouiz:
ok

carriepl:
@caglar : do you have a profile output on your use case to see the difference in compilation and execution between this version and the old version (both optimization and execution are relevant in this case since the goal is also to see if there's any difference in the resulting graph)

caglar:
@carriepl I have just noticed your message, but I have sent my profile output on Friday by e-mail.

carriepl:
There's an issue in PushOutNonSeqScan; it starts its exploration from only the explicit inputs of the graph and not also the constants present in the graph. Because of this it might skip entire sections of the graph.
I fixed this in a local branch and then ran various tests. In none of them I saw a speed up compared to the implementation presently in the master branch of the central repository.

caglar:
@carriepl sorry last week and this week was pretty crazy. Moving out from Montreal...etc.

 First of all, what is the expected speedup that should be obtained from this ticket, second of all why do you think that it didn't achieve the expected speedup. Do you think that is because of the algorithm or something else? I don't think I will be able to work on this week, but will try to have a look during the weekend.

nouiz:
For now, don't work on it anymore. We aren't sure if it is useful to
continue in that direction.

Can you give us a way to reproduce your use cases where you saw speed up
from that PR?

carriepl:
I am not saying that it did not achieve the expected speedup, I'm saying that, with the bug fixed, I did not see any speedup in the tests that I ran, not even a few percent points.

Now, I did not have the use case you tried (I emailed you for it, you answered you were going to send it to me but, as far as I know, I received nothing), but I did have other use cases, including one provided by a student where much of the compilation time was spent in these two optimizations.

caglar:
@carriepl in the last few weeks was quiet hectic, as I mentioned earlier a few deadlines and moving out from Montreal and now work etc... I was planning to prepare a script for an old model that I was playing with for which compilation was taking about 4 hours. But I completely forgot about it after the e-mail.

I have just send you another e-mail about the model that I was testing the optimization before. Can you point me to your branch where you fixed the bug at? I might try it on some other code as well.

Issue #3377, opened 2015-09-10
Serious bug causes wrong computations in Keras, etc.


the-moliver:
Starting with this commit https://github.com/Theano/Theano/commit/41a8e89b63f5b4d4f9a0b942554ed5483e1ceab5 the training of various kinds of neural nets, and I presume other computations, is broken. I don't know the Theano codebase well enough to diagnose the exact cause, but I tracked it down to this commit. This issue has been confirmed by others here:
https://github.com/fchollet/keras/issues/664

nouiz:
Quick reply, I'm working on it.

the-moliver:
Thanks for the quick work!

PR #3378, opened 2015-09-10
fix gh-3377 bug introduced by 41a8e89b63f5 merged by 7320e1b. GpuElem…

nouiz:
…wise with non contiguous inputs give bad results


abergeron:
We should add a test so that the bug does not come back, but for now, this fixes it so I'm merging.

PR #3379, opened 2015-09-10
Enable merging of GpuDnnPoolDesc

lamblin:
if they are identical, as well as other Ops that do not take any input.
@aalmah, can you check that this removes one of the convolution/pooling from your graph?
@vdumoulin, this may also fix the issue you had with batch normalization and DNN convolutions. Can you check?


aalmah:
@lamblin yes, it removes one convolution/pooling from the graph. Thanks!

PR #3380, opened 2015-09-11
Mixed2

nouiz:
Add doc
better error
speed up opt by not making inplace ifelse/elemwise on scalar.
update to use cudnn v3 final
Update to NanGuardMode. Add flag NanGuardMode.action={raise,warn,pdb}


theano/compile/nanguardmode.py
```diff
@@ -24,6 +24,11 @@
              BoolParam(True),
              in_c_key=False)
 
+AddConfigVar('NanGuardMode.action',
+             "What NanGuardMode do when it find a problem",
```
abergeron: What NanGuardMode do -> What NanGuardMode does
when it find -> when it finds

theano/compile/nanguardmode.py
```diff
@@ -55,13 +60,14 @@ def flatten(l):
...
-def contains_nan(arr):
+def contains_nan(arr, nd=None):
     """
     Test whether a numpy.ndarray contains any `np.nan` values.
 
     Parameters
     ----------
-    arr : np.ndarray
+    arr : np.ndarray or output of any Theano op
+    nd : if the output of an Theano op, the node associated to it
```
abergeron: This doesn't have the right format for a parameter.  The format is
```
name : type
    descr
```
and what you have for `nd` is not a type.

`nd` makes me thinks about a number of dimensions so I think `node` is a much better name.

of an Theano -> of a Theano

Finally, I think it would be better design to have contains_nan be a method of Type.  I'm not sure what the best design would be to make exceptions like for GPU_mrg_uniform, though.

theano/compile/nanguardmode.py
```diff
@@ -80,17 +86,30 @@ def contains_nan(arr):
...
+def contains_inf(arr, nd=None):
     """
     Test whether a numpy.ndarray contains any `np.inf` values.
 
     Parameters
     ----------
-    arr : np.ndarray
-
+    arr : np.ndarray or output of any Theano op
+    nd : if the output of an Theano op, the node associated to it
```
abergeron: See comments about `contains_nan`.  They all apply here too.

theano/compile/nanguardmode.py
```diff
@@ -239,20 +271,24 @@ def do_check_on(var, nd, f, is_input):
...
-                logger.error(nd)
-                assert False
+                    logger.error("NanGuardMode found an error in the"
+                                 " input %d of this node.")
+                    logger.error('Node:')
+                    logger.error(nd)
+                    logger.error("The input variable that cause problem:")
+                    logger.error(theano.printing.debugprint(nd, file='str'))
+                if config.NanGuardMode.action == 'raise':
+                    assert False
```
abergeron: Raising some exception with the error message that is logged above is what I would expect for `"raise"`

theano/gof/link.py
```diff
@@ -302,8 +302,15 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
             "HINT: Use the Theano flag 'exception_verbosity=high'"
             " for a debugprint and storage map footprint of this apply node.")
 
-    exc_value = exc_type(str(exc_value) + detailed_err_msg +
-                         '\n' + '\n'.join(hints))
+    try:
+        exc_value = exc_type(str(exc_value) + detailed_err_msg +
+                             '\n' + '\n'.join(hints))
+    except TypeError:
+        print("WARNING: %s error do not allow us to add extra error message" %
```
abergeron: error do not -> error does not

theano/ifelse.py
```diff
@@ -395,7 +395,13 @@ def ifelse(condition, then_branch, else_branch, name=None):
 @gof.local_optimizer([IfElse])
 def cond_make_inplace(node):
     op = node.op
-    if isinstance(op, IfElse) and not op.as_view:
+    if (isinstance(op, IfElse) and
+        not op.as_view and
```
abergeron: Flake8 indentation.

abergeron: Or not since it's fine with the construct below.

theano/sandbox/cuda/dnn.py
```diff
@@ -2380,7 +2388,7 @@ def local_softmax_dnn_grad(node):
               isinstance(node.inputs[0].owner.op, HostFromGpu)) or
              (node.inputs[1].owner and
                  isinstance(node.inputs[1].owner.op, HostFromGpu)))):
-            if not dnn_available() or version() != (2000, 2000):
+            if not dnn_available():
                 # Softmax grad is broken in v3 rc1 for this case
```
abergeron: Remove the comment also

theano/sandbox/cuda/elemwise.py
```diff
@@ -66,7 +66,7 @@ class NaiveAlgo(object):
     def cache_version(self):
         ver = self.scalar_op.c_code_cache_version()
         if ver:
-            return (19, self.verbose, self.sync, ver)
+            return (20, self.verbose, self.sync, ver)
```
abergeron: Do you really need to bump the C code version if you are only adding comments?

theano/sandbox/cuda/opt.py
```diff
@@ -795,6 +795,11 @@ def local_gpu_careduce(node):
             replace = False
             if x.owner and isinstance(x.owner.op, HostFromGpu):
                 replace = True
+            # If this is a useless reduce, remove it as
+            # local_cut_useless_reduce.  This is needed as the code
+            # bellow do not support when x.ndim == 0.
```
abergeron: bellow -> below

abergeron:
The only change I'm not sure about is about contains_nan(), and contains_inf().  The rest is fine minus some small typos.

theano/compile/nanguardmode.py
nouiz: I'm not sure ether, so I'll let it like this for now.

theano/sandbox/cuda/elemwise.py
nouiz: I think it is better practice, in case I made mistake in my comment that make compilation crash.

nouiz:
I updated this PR with your comments. I also fixed var.owner test that wasn't good.

theano/sandbox/cuda/elemwise.py
abergeron: ok

abergeron:
Failure on python3 because of test/test_ifelse.py line 67 that is concatenating range() and list()

nouiz:
I fixed the test.

PR #3382, opened 2015-09-11
Scan inplace opt

carriepl:
This is work in progress to attempt to speed up the ScanInplaceOptimizer() since some users have been reporting a lot of time being spent in this optimization.

Instead of trying to make the scan compute every output inplace one by one, this starts by attempting to make scan compute all its outputs inplace. It is doesn't work then it goes back to trying every output one by one.


carriepl:
According to my informal testing (running the tests in test_scan.py:T_Scan), trying to first compute every recurrent state inplace works ~90% of the time. Only 10% of the time is the rest of the optimization used.

PR #3384, opened 2015-09-11
Fix a non-ascii character in a comment in test_flake8.py.

abergeron:
Apparently the python on travis doesn't care, but my version of python complains about this.


carriepl:
Modification is in a comment so I see no need to wait for travis to finish

PR #3385, opened 2015-09-12
Use shape_of instead of explicitly inserting Shape_i

lamblin:
These Shape_i nodes will not always be optimized away otherwise, and
they prevent local_shape_to_shape_i from being applied.
Instead, we manually apply the equivalent of local_shape_to_shape_i,
by using shape_of[...] directly.

A test is still missing.
@aalmah, this should fix the last remaining convolution in your case.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -924,7 +924,7 @@ class ShapeFeature(object):
             # worst case, we loop over shape_of and replace things
             raise NotImplementedError(s_i)
 
-        # s_i is x.shape[i], we change it to Shape_i.
+        # s_i is x.shape[i] for some x, we change it to shape_of[x][i]
         if (s_i.owner and
                 isinstance(s_i.owner.op, Subtensor) and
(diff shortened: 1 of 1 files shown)


abergeron:
I'm ok with this change, but I wonder why it block local_shape_to_shape_i otherwise.

lamblin:
@abergeron It does not prevent the optimization from running, it just removes the `Shape` from the expression, so `local_shape_to_shape_i` does not have a chance to replace it by the content of `shape_of`.

aalmah:
@lamblin I have tested it, and yes it removes the last extra convolution left. Thanks for the great work!

nouiz:
I think too this is good.

A test would be great. What about a pickle of Amjad graph? If you know how
to manually build it, that is better, but this type of graph are normally
hard to build.

I think we can merge without the new tests. Do both of you agree? If so, we
should do it I think.

abergeron:
If I wasn't clear, I'm ok with merging this change.  I'm just curious as to why things don't work as expected in that situation.

PR #3387, opened 2015-09-12
Allow dnn pooling ws, stride and pad to change at run time.

nouiz:
TODO:
- [ ] add test for the dynamic
- [ ] port this to gpuarray
- [ ] fix the todo about how to find the c int dtype correspondance in python to have this always work. Maybe it is currently ok, but i'm not sure.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -1306,7 +1306,8 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1),
 
 class GpuDnnPoolDesc(GpuOp):
     """
-    This Op builds a pooling descriptor for use in the other pooling operations.
+    This Op builds a pooling descriptor for use in the other pooling
+    operations.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


theano/sandbox/cuda/dnn.py
```diff
@@ -1784,9 +1844,9 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
         (padX, padY) padding information.
         padX is the size of the left and right borders,
         padY is the size of the top and bottom borders.
-    :param nd: dimensions of pooling, can be 2 or 3 for 2d or 3d pooling
-        If set to 3 all other params (except mode) must have an extra
-        dimension to match. 3 is only available for cudnn v3
+    ndim None, 2 or 3
```
abergeron: ndim needs a colon (:) after

abergeron:
Please hold off on porting to gpuarray.  I want to finish the multi-gpu first.

nouiz:
I wrote elsewhere that I won't work on this shortly. Someone else was
interrested to contribute the missing tests.

nouiz:
I close this PR as this only fix part of the problems. See gh-3499 for the follow up.

theano/sandbox/cuda/dnn.py
```diff
@@ -1340,41 +1343,95 @@ def c_compiler(self):
...
-        return Apply(self, [],
+    def make_node(self, ws, stride, pad):
+        ws = tensor.as_tensor_variable(ws)
+        stride = tensor.as_tensor_variable(stride)
+        pad = tensor.as_tensor_variable(pad)
+        assert ws.ndim == stride.ndim == pad.ndim == 1
+        ws = ws.astype(theano.scalar.upcast("int32", ws.dtype))
+        stride = stride.astype(theano.scalar.upcast("int32", stride.dtype))
+        pad = pad.astype(theano.scalar.upcast("int32", pad.dtype))
+        # TODO how to detect that int are int64 or int32?
```
abergeron: You could use the 'intp' dtype.  That is the int type which is the same size as ssize_t for numpy.

Issue #3096, opened 2015-07-01
Nose is loaded in production


cclamb:
Hey everybody, nose is loading in production, and it breaks interaction with Theano and ipdb in ipython notebooks (see https://github.com/gotcha/ipdb/issues/52).

When I load theano into ipython or the python REBL, I get this (or equivalent, this is in the standard python REPL):

```
Python 2.7.10 (default, Jun 10 2015, 19:42:47) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> msg = 'got nose!' if 'nose' in sys.modules.keys() else 'no nose.'
>>> print msg
no nose.
>>> import theano
>>> import sys
>>> msg = 'got nose!' if 'nose' in sys.modules.keys() else 'no nose.'
>>> print msg
got nose!
>>> 
```

I've hacked around this, but you might want to consider removing nose references from your distributed builds.

nouiz:
Do you want to work on it? We are pretty busy and won't work on this shortly.

If you can't work on it, can you paste the diff of your work around to be sure to guide the fix of that problem?

thanks for the report.

cclamb:
Sure, I'll see what I can do. The workaround is in the cited ticket over on the ipdb side.

lamblin:
The unit tests are part of the distributed builds, and they depend on `nose` for things like `SkipTest` for instance.
However, it is true that the tests should not be imported when simply importing `theano`.
Likely culprits are:
- `theano.tests.unittest_tools`, which may get imported from outside
- `theano.sandbox.cuda`, which imports `test_driver` from the `tests` subdirectory. `test_driver` should probably be moved somewhere else.

mohammadpz:
@abergeron , This happens at lines 92 to 100 of ```theano/__init__.py```:
```python
# This need to be before the init of GPU, as it add config variable
# needed during that phase.
import theano.tests
if hasattr(theano.tests, "TheanoNoseTester"):
    test = theano.tests.TheanoNoseTester().test
else:
    def test():
        raise ImportError("The nose module is not installed."
                          " It is needed for Theano tests.")
```
When we import ```import theano.tests```, it imports nose too.

Any idea on how to avoid importing tests or importing then whenever needed?

abergeron:
Maybe create a function `test()` which does the above imports/tests?

abergeron:
Fixed by #3389

PR #3389, opened 2015-09-13
test function is moved to another function

mohammadpz:
Dear @abergeron ,
Here's my PR regarding issue #3096 .

### DIFF
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -93,11 +93,11 @@ from theano.gradient import Rop, Lop, grad, subgraph_grad
 # needed during that phase.
 
 
-def test(*args):
+def test():
     import theano.tests
     if hasattr(theano.tests, "TheanoNoseTester"):
         test = theano.tests.TheanoNoseTester().test
-        test(*args)
+        return test()
     else:
         raise ImportError("The nose module is not installed.")
 
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -91,15 +91,13 @@ from theano.gradient import Rop, Lop, grad, subgraph_grad
 
 # This need to be before the init of GPU, as it add config variable
 # needed during that phase.
-
-
-def test():
-    import theano.tests
-    if hasattr(theano.tests, "TheanoNoseTester"):
-        test = theano.tests.TheanoNoseTester().test
-        return test()
-    else:
-        raise ImportError("The nose module is not installed.")
+import theano.tests
+if hasattr(theano.tests, "TheanoNoseTester"):
+    test = theano.tests.TheanoNoseTester().test
+else:
+    def test():
+        raise ImportError("The nose module is not installed."
+                          " It is needed for Theano tests.")
 
 if config.device.startswith('gpu') or config.init_gpu_device.startswith('gpu'):
     import theano.sandbox.cuda
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -91,13 +91,15 @@ from theano.gradient import Rop, Lop, grad, subgraph_grad
 
 # This need to be before the init of GPU, as it add config variable
 # needed during that phase.
-import theano.tests
-if hasattr(theano.tests, "TheanoNoseTester"):
-    test = theano.tests.TheanoNoseTester().test
-else:
-    def test():
-        raise ImportError("The nose module is not installed."
-                          " It is needed for Theano tests.")
+
+
+def test(*args):
+    import theano.tests
+    if hasattr(theano.tests, "TheanoNoseTester"):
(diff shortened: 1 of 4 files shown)


theano/__init__.py
```diff
@@ -91,13 +91,15 @@
...
-import theano.tests
-if hasattr(theano.tests, "TheanoNoseTester"):
-    test = theano.tests.TheanoNoseTester().test
-else:
-    def test():
-        raise ImportError("The nose module is not installed."
-                          " It is needed for Theano tests.")
+
+
+def test(*args):
```
mohammadpz: I used ```*args``` to return the results of the actual function, whenever this "fake" test function is called. Am I doing it in a right way?

mohammadpz:
@abergeron ,
I get this error:
```
 File "/Users/Mohammadpz/anaconda/lib/python2.7/site-packages/numpy/testing/nosetester.py", line 188, in __init__
raise AssertionError
```
Here's lines around line 188:
```python
def __init__(self, package=None, raise_warnings="release"):
        package_name = None
        # import ipdb; ipdb.set_trace()
        if package is None:
            f = sys._getframe(1)
            package_path = f.f_locals.get('__file__', None)
            if package_path is None:
                raise AssertionError
            package_path = os.path.dirname(package_path)
            package_name = f.f_locals.get('__name__', None)
        elif isinstance(package, type(os)):
            package_path = os.path.dirname(package.__file__)
            package_name = getattr(package, '__name__', None)
        else:
            package_path = str(package)
        self.package_path = package_path
```
Apparently ```package_path``` is ```None```.

theano/__init__.py
abergeron: `test()` doesn't take arguments.

theano/__init__.py
```diff
@@ -91,13 +91,15 @@
...
+    import theano.tests
+    if hasattr(theano.tests, "TheanoNoseTester"):
+        test = theano.tests.TheanoNoseTester().test
+        test(*args)
+    else:
+        raise ImportError("The nose module is not installed.")
```
abergeron: Maybe change this error to something else than ImportError.

abergeron:
I'll try it to see if I get the package_path error too.

abergeron:
Apparently this approach is a bad idea even if we fix the assertion error because running this inside a function causes all sorts of nasty side-effects.

Another approach would be to move the imports of nose in theano/tests/main.py to the function(s) that use them to make sure nose isn't imported.

mohammadpz:
Hi, I'm going to Toronto for 2 days. When I come back, I get back to this
and will try to find place where test() is used.

mohammadpz:
@abergeron ,
I put an ```import ipdb; ipdb.set_trace()``` in the ```test()``` function and then I run ```nosetests``` to realize where this function is called. But I didn't get anything. How should I realize where this function is used?

abergeron:
I don't mean to trace where test() is used.  I mean to restore the code like it was before and make sure that importing `TheanoNoseTester` doesn't import nose.

You can do that by either making `import theano.tests` not import nose or moving `TheanoNoseTester` elsewhere.

abergeron:
It seems that moving SkipTest breaks lots of tests that import it from there.  We could fix all of those to import directly from nose.  Since we now require 1.3.0 this would be ok.

Alternatively not importing unittest_tools automatically has less chances of breaking stuff.  But you would have to try both approaches and see which one is easier to fix.

Also, If you want feedback on some new changes you pushed, post a comment.  We get notifications for new comments, but not for new pushes.

mohammadpz:
Hi @abergeron , 
I tried option 1, and then I revert it back and then I tried option 2. Both seem working.
So what's you order, boss? :)

abergeron:
I like option 1 better, since it's less likely to break untested things.

Although for the final commit I would advise doing a rebase that would remove the unused commits.

mohammadpz:
@abergeron ,
I fetch from main_stream and then interactive rebase. I removed the line on option 2.
Is that fine?

abergeron:
The final changes are ok, but there are lots of commit that add and remove changes.  It would be good to remove those also.

Otherwise I'm ok to merge.

theano/tests/main.py
```diff
@@ -56,6 +51,7 @@ def prepare_test_args(self, verbose=1, extra_argv=None, coverage=False,
 
         Takes the same arguments as `test`.
         """
+        import nose.plugins.builtin
```
nouiz: Add a comment at each of the place you added import not at the top of the file.

Otherwise, in the futur someone will forget why this file isn't pep8 complient and will revert this commit.

PR #3397, opened 2015-09-16
automatic gpu selection no longer works in CUDA-TK 7.0

vesis84:
- fixing an issue introduced with CUDA-TK 7.0, when automatic selection
  of free gpu by os / library no longer works. Library always selects '0',
  which leads to crash in case of 'exclusive mode' when '0' is taken and 
  other gpus are free.
- this fix is inspired by Dan Povey's fix for Kaldi:
  kaldi-asr/kaldi@6548565
- it does a loop over gpus until a free gpu is taken.
- pycuda initialization was also affected by the bug from CUDA 7.0, so it is better to disable it.

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -399,9 +399,9 @@ def use(device,
 
         # Has PyCUDA already initialized the GPU context
         pycuda_init_dev = False
-        if config.pycuda.init:
-            import theano.misc.pycuda_init
-            pycuda_init_dev = theano.misc.pycuda_init.pycuda_available
+#        if config.pycuda.init:
+#            import theano.misc.pycuda_init
+#            pycuda_init_dev = theano.misc.pycuda_init.pycuda_available
 
         try:
             if (device != 'gpu') and not pycuda_init_dev:
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -399,9 +399,9 @@ def use(device,
 
         # Has PyCUDA already initialized the GPU context
         pycuda_init_dev = False
-#        if config.pycuda.init:
-#            import theano.misc.pycuda_init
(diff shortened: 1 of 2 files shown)


theano/sandbox/cuda/__init__.py
```diff
@@ -399,9 +399,9 @@ def use(device,
 
         # Has PyCUDA already initialized the GPU context
         pycuda_init_dev = False
-        if config.pycuda.init:
-            import theano.misc.pycuda_init
-            pycuda_init_dev = theano.misc.pycuda_init.pycuda_available
+#        if config.pycuda.init:
```
nouiz: Why this? By default pycuda.init is False, so by default it do nothing.

nouiz:
This all seem good. I tested it and we also had this problem, but never investigated. Thanks!

nouiz:
When travis pass, we can merge.

nouiz:
I forgot, what about the pycuda change? I don't understand why this is needed.

vesis84:
The pycuda change is not needed as long as the pycuda gpu selection is disabled by default. The problem will reappear only in the case if someone enables pycuda gpu selection and has CUDA 7.0. So there could be some warning message instead of disabling pycuda gpu selection?

vesis84:
Other reason to disable 'pycuda' is that we have seen a situation when 2GPUs were taken by theano. One was taken by pycuda and the other with our modified cuda_ndarray.cuda_ndarray.select_a_gpu().
It disappeard after commenting the pycuda initialization. Is pycuda something that many people use and depend on?

vesis84:
Hi, I changed it a bit. Now if the gpu is selected by pycuda, we don't call cuda_ndarray.select_a_gpu() anymore. While there is the initialization for cublas, etc... Cheers! K.

nouiz:
that seem the right fix. We can merge when travis pass.

thanks

Issue #3388, opened 2015-09-13
Docker images


Kaixhin:
I've created [CPU-backend](https://hub.docker.com/r/kaixhin/theano/) and [CUDA-backend](https://hub.docker.com/r/kaixhin/cuda-theano/) (v6.5, v7.0 and v7.5) Docker images for Theano. They may be worth adding to the [installation docs](http://www.deeplearning.net/software/theano/install.html#install) as an alternative for Ubuntu (like [Torch7](https://github.com/torch/torch7/wiki/Cheatsheet#docker-images) or [neon](http://neon.nervanasys.com/docs/0.9.0/installation.html#docker-images)).

nouiz:
Do you want to make a PR to Theano that add that information?

PR #3399, opened 2015-09-16
Add Docker images to the installation docs

Kaixhin:
This pull request adds an alternative way of running Theano to the docs. These Docker images have been in use successfully for some time for not just using Theano, but also as the basis for other libraries that build on top of Theano.

Closes https://github.com/Theano/Theano/issues/3388.


nouiz:
thanks

PR #3402, opened 2015-09-17
Fix GpuSoftmax and GpuSoftmaxWithBias for non-float32 operation.

abergeron:
This fixes a bug which causes crashes in float16 and wrong results in float64.

### DIFF
--- a/theano/sandbox/gpuarray/tests/test_nnet.py
+++ b/theano/sandbox/gpuarray/tests/test_nnet.py
@@ -162,6 +162,14 @@ def test_GpuCrossentropySoftmax1HotWithBiasDx():
             rtol, atol)
 
 
+def test_softmax_with_bias_float16():
+    softmax_with_bias_unittest_template(dtypeInput='float16',
+                                        dtypeBias='float32')
+    softmax_with_bias_unittest_template(dtypeInput='float16',
+                                        dtypeBias='float16')
+    softmax_with_bias_unittest_template(dtypeInput='float32',
+                                        dtypeBias='float16')
+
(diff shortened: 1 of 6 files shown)


nouiz:
Can you modify a current test to test float16 and float32?

abergeron:
Done.

theano/sandbox/gpuarray/tests/test_nnet.py
```diff
@@ -209,11 +207,7 @@ def softmax_with_bias_unittest_template(dtypeInput, dtypeBias):
...
     def cmp(n, m):
-        # print "test_softmax",n,m
-        if dtypeInput == 'float32':
-            data = numpy.arange(n * m, dtype='float32').reshape(n, m)
-        elif dtypeInput == 'float64':
-            data = numpy.arange(n * m, dtype='float64').reshape(n, m)
+        data = numpy.arange(n * m, dtype=dtypeInput).reshape(n, m)
 
         out = f(data)
         gout = f_gpu(data)
```
nouiz: The assert bellow is failing as the comparison is too strict. Use:

theano.tests.unittest_tools.assert_allclose(out, gout)

abergeron: Done.

nouiz: test fails if run locally with wrong value and low error and nan.

abergeron:
This should fix the nan problems with both tests.  I had to modify how they work a bit so make sure they still test what they should.

PR #3403, opened 2015-09-17
Faster merge

nouiz:
Also better error message, remove duplicate output and code simplification


abergeron:

nouiz:
I merge as @abergeron is ok and travis pass.

PR #3404, opened 2015-09-17
Add stack trace for nnet

kelvinxu:
#3018


theano/tensor/nnet/nnet.py
```diff
@@ -636,14 +638,17 @@ def local_softmax_with_bias(node):
                     vector_sum = tensor.add(*vectors)
                 else:
                     vector_sum = vectors[0]
+                copy_stack_trace(x_in, vector_sum)
```
lamblin: I would prefer to call `copy_stack_trace` only if a new `add` node is created, because otherwise `vectors[0]` already has a stack trace. Could you put it inside the `if len(...)` block?

theano/tensor/nnet/nnet.py
```diff
@@ -636,14 +638,17 @@ def local_softmax_with_bias(node):
...
 
                 if len(non_vectors) > 1:
                     non_vector_sum = tensor.add(*non_vectors)
                 else:
                     non_vector_sum = non_vectors[0]
+                copy_stack_trace(x_in, non_vector_sum)
```
lamblin: Same here.

theano/tensor/nnet/nnet.py
```diff
@@ -636,14 +638,17 @@ def local_softmax_with_bias(node):
...
 
                 try:
                     sm_bias = softmax_with_bias(non_vector_sum, vector_sum)
+                    copy_stack_trace(x_in, non_vector_sum)
```
lamblin: Probably a copy-paste error: here, you should actually copy the trace from `node.outputs[0]` to `sm_bias`.

theano/tensor/nnet/nnet.py
```diff
@@ -1485,13 +1383,18 @@ def local_argmax_pushdown(node):
         if x.owner and x.owner.op in (softmax_op, softplus, tensor.exp,
                                       tensor.log, tensor.tanh, sigmoid):
             pre_x, = x.owner.inputs
-            return tensor._max_and_argmax(pre_x, axis)
+            ret = tensor._max_and_argmax(pre_x, axis)
+            copy_stack_trace(pre_x, ret)
```
lamblin: So actually, I think here we should copy the stack trace of `x_max`.
We are replacing `argmax(increasing_op(pre_x))` by `argmax(pre_x)`, so I think what we want to keep track of is the creation of the original `argmax` (not of `pre_x`).

theano/tensor/nnet/nnet.py
```diff
@@ -1485,13 +1383,18 @@ def local_argmax_pushdown(node):
...
-            return tensor._max_and_argmax(pre_x +
-                                          tensor.DimShuffle(
-                                              pre_bias.broadcastable,
-                                              ('x', 0))(pre_bias), axis)
+            ret = tensor._max_and_argmax(pre_x +
+                                         tensor.DimShuffle(
+                                             pre_bias.broadcastable,
+                                             ('x', 0))(pre_bias), axis)
+            # copy both stack traces
+            copy_stack_trace([pre_x, pre_bias],  ret)
```
lamblin: Same here, we want to be able to trace back the creation of the original argmax.

theano/tensor/nnet/nnet.py
```diff
@@ -1585,9 +1488,11 @@ def local_advanced_indexing_crossentropy_onehot(node):
         # Check that rows == arange(labels.shape[0])
         if _check_rows_is_arange_len_labels(rows, labels):
             if labels.ndim == 1 and x_var.ndim == 2:
-                return [-crossentropy_softmax_argmax_1hot_with_bias(x_var,
-                                                                    b_var,
-                                                                    labels)[0]]
+                ret = -crossentropy_softmax_argmax_1hot_with_bias(x_var,
+                                                                  b_var,
+                                                                  labels)[0]
+                copy_stack_trace([x_var, b_var, labels], ret)
```
lamblin: This would actually copy the stack trace to the `neg` node that corresponds to the minus sign.
I'm not sure what would be best to assign to the output of `crossentropy_...bias`, but I think the stack of `node.outputs[0]` would make sense for both, even if they do not completely correspond to the computation being performed. Ideally, we would also have the intermediate steps of the computation as well (corresponding to `softmax(x)` and `log(softmax(x))` in the first case, and to `softmax(x)` and `softmax(x)[...]` in the second case), but it would require some bigger changes in the optimization itself.
So for the moment, I think `copy_stack_trace(node.outputs[0], [minus_ret, ret])` (where `minus_ret` would be `crossentropy_...(...)[0]`) would be enough.

theano/tensor/nnet/nnet.py
```diff
@@ -1809,11 +1714,14 @@ def local_advanced_indexing_crossentropy_onehot_grad(node):
 
     # Dimension check before substitution
     if labels.ndim == 1 and x_var.ndim == 2:
-        return [crossentropy_softmax_1hot_with_bias_dx(out_grad, sm, labels)]
+        ret = crossentropy_softmax_1hot_with_bias_dx(out_grad, sm, labels)
+        # The stack trace of output_grad, sm and labels are not added
+        # but may need to be added at a future point
```
lamblin: Actually, we may want to add _a_ stacktrace _to_ out_grad, sm, and labels at a future point.

theano/tensor/nnet/nnet.py
```diff
@@ -23,6 +23,7 @@
 from theano.tensor import subtensor
 from theano.tensor import elemwise
```
lamblin: The useless block of code you have deleted was the only thing using `elemwise`, that is why flake8 is now complaining.

theano/tensor/nnet/nnet.py
```diff
@@ -591,7 +593,7 @@ def softmax_graph(c):
 def softmax(c):
     return softmax_op(c)
 
-
+# seems like need to change softmax_with_bias
```
lamblin: You will have to remove that comment in the final version.

theano/tensor/nnet/nnet.py
```diff
@@ -592,6 +594,7 @@ def softmax(c):
     return softmax_op(c)
 
 
+# seems like need to change softmax_with_bias
```
lamblin: I'm not sure what this comment means. Should it be removed?

theano/tensor/nnet/nnet.py
```diff
@@ -1885,7 +1797,10 @@ def local_useless_crossentropy_softmax_1hot_with_bias_dx_alloc(node):
                 msg = '`sm` and `dy` do not have the same shape.'
                 dz = opt.Assert(msg)(dz, cond)
 
-            return [node.op(dz, sm, y_idx)]
+            ret = node.op(dz, sm, y_idx)
+            # copy node.outputs[0] to ret according to Pascal
```
lamblin: I don't think you really have to give me credit here :)

PR #3406, opened 2015-09-17
Add requirement specific to rtd

nouiz:
Try to fix readthedoc


nouiz:
no need to wait for travis

nouiz:
no need to wait for travis

PR #3410, opened 2015-09-18
Add batch normalization op

ballasn:


theano/tensor/nnet/bn.py
```diff
@@ -0,0 +1,56 @@
...
+        dmean = -(top*gamma) / var
+        dvar = -(top * gamma * (x - mean)) / (var*var)
+        dgamma = top*(x - mean) / var
+        return [dx, dmean, dvar, dgamma, top]
+
+
+def batch_normalization(inputs, gamma, beta, mean, variance, axis=0):
+    """
+    This function will build the symbolic graph for applying batch normalization
+    to a set of activations.
```
nouiz: Tell that this lower the memory usage during the back-prop, but could be a slower then if you build the corresponding Theano graph.

Tell that it work on the GPU.

nouiz:
$nosetests theano/tests/test_flake8.py
F
  File "/Tmp/lisa/os_v4/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/bastienf/repos/Theano/theano/tests/test_flake8.py", line 243, in test_format_flake8
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:14:43: E241 multiple spaces after ','
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:16:16: E225 missing whitespace around operator
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:22:18: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:23:22: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:24:50: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:25:21: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/bn.py:56:1: W391 blank line at end of file
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:11:15: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:11:18: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:12:17: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:12:19: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:29:36: E203 whitespace before ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:29:37: E231 missing whitespace after ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:30:40: E203 whitespace before ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:30:41: E231 missing whitespace after ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:49:15: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:49:18: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:50:17: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:50:19: E226 missing whitespace around arithmetic operator
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:65:5: E266 too many leading '#' for block comment
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:76:36: E203 whitespace before ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:76:37: E231 missing whitespace after ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:77:40: E203 whitespace before ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:77:41: E231 missing whitespace after ','
/u/bastienf/repos/Theano/theano/tensor/nnet/tests/test_bn.py:90:1: W391 blank line at end of file

theano/tensor/nnet/bn.py
```diff
@@ -0,0 +1,56 @@
...
+        the number of inputs channel
+    beta: symbolic vector
+        BN shift parameter, must be of same dimension that
+        the number of inputs channel
+    mean: symbolic tensor
+        inputs means
+    variance: symbolic tensor
+        inputs variance
+    axis: int
+        channel axis
```
f0k: What if there are multiple channel axes (i.e., axes I don't want to normalize over, but have separate means/variances for)? Would it be easy to support a list of axes here?

theano/tensor/nnet/bn.py
```diff
@@ -0,0 +1,56 @@
...
+    This function will build the symbolic graph for applying batch normalization
+    to a set of activations.
+
+    Parameters
+    ----------
+    inputs : symbolic tensor
+        Mini-batch of examples
+    gamma: symbolic vector
+        BN scale parameter, must be of same dimension that
+        the number of inputs channel
```
f0k: Actually you implemented it in a different way, it must be of the same dimensionality as `inputs` and broadcastable against it, as I see it in the tests. The implementation is good the way it is, I think, just the docstring would need to be adapted.

theano/tensor/nnet/bn.py
f0k: Ah wait, `axis` is not used at all, since you're requiring inputs to be broadcasted outside of that function. So you can either remove the `axis` argument completely, or move the broadcasting inside here, and allow `axis` to be a list, something like:
```python
if isinstance(axis, int):
    axis = [axis]
dims = range(gamma.ndim)
pattern = [next(dims) if dim in axis else 'x' for dim in range(inputs.ndim)]
gamma = gamma.dimshuffle(*pattern)
beta = beta.dimshuffle(*pattern)
mean = mean.dimshuffle(*pattern)
variance = variance.dimshuffle(*pattern)
```

But I think ultimately it gives a more flexible implementation if you just drop the `axis` argument and require everything to be broadcastable against `inputs` (i.e., just alter the docstring, keep the implementation as it is).

theano/tensor/nnet/bn.py
f0k: Why could it be slower?

ballasn: As no intermediate representations are stored for the back-propagation  it needs to redo some forward computations for the backprop.

theano/tensor/nnet/bn.py
ballasn: Thanks! I removed axis and updated the docstring accordingly

nouiz:
One test if failing:

ERROR: theano.tensor.nnet.tests.test_bn.test_bn
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/tensor/nnet/tests/test_bn.py", line 43, in test_bn
    utt.verify_grad(batch_normalization, [X, G, B, X.mean(axis=0, keepdims=True), X.std(axis=0, keepdims=True)])
TypeError: 'keepdims' is an invalid keyword argument for this function

theano/tensor/nnet/bn.py
```diff
@@ -0,0 +1,57 @@
...
+        return [dx, dmean, dstd, dgamma, top]
+
+
+def batch_normalization(inputs, gamma, beta, mean, std):
+    """
+    This function will build the symbolic graph for applying batch normalization
+    to a set of activations. As no intermediate representations are stored for the
+    back-propagation, this implementation lower the memory usage, however,
+    it is 5-10% slower than a naive theano implementation, as it redo
+    some foward computations for the backprop.
```
f0k: Then maybe there should be a `mode` argument or something that defaults to `"low_mem"` (where it uses this Op) but can also be set to `"fast"` (where it just directly returns `((inputs - mean) / std) * gamma + beta`)?

@nouiz: What about the optimizations you talked about in https://github.com/Lasagne/Lasagne/issues/141#issuecomment-129663016 and following -- is there a way to make use of them here in future? Or will they be used automatically when they are implemented along with an optimizer? I guess if there's a `batch_normalization()` function in Theano, people will expect it to be faster/better than writing down the expression themselves.

f0k:
> utt.verify_grad(batch_normalization, [X, G, B, X.mean(axis=0, keepdims=True), X.std(axis=0, keepdims=True)])
> TypeError: 'keepdims' is an invalid keyword argument for this function

The Python 2.6 tests run with numpy 1.6.2, but `keepdims` was only introduced in 1.7. (You'll need to use `X.mean(axis=0)[np.newaxis]` instead.)

theano/tensor/nnet/bn.py
nouiz: This PR is a continuation of the idea in lasagne comment. It is the same, a big Composite with a grad implemented. We changed how to implement the grad as it was easier/faster to implement that way.

The speed difference I timmed on a matrix of size 128,256 on the GPU was that the big op was 7% to 10% then the full graph during training. But it was using less memory.

As the quantity of memory on the GPU is a big bottleneck, I would default to the version that use less memory as it significatively less memory.

There won't be optimization that introduce it manually. If someone add it, it should not be enabled by default, but only with the low_mem tag is used.

Personnaly, as this version use much less memory, I think people would prefer that even if there is a small speed loos. If you think differently, tell it.

@ballasn didn't remart the speed different on a small model on mnist.

f0k: > Personnaly, as this version use much less memory, I think people would prefer that even if there is a small speed loos. If you think differently, tell it.

Not everybody -- my models currently fit easily within 12 GiB so I prefer speed over memory usage. So I'd be fine with having it default to low memory usage, but I still strongly encourage adding a `mode="low_mem"` keyword argument so people like me are made aware of and can easily switch between the two versions.

ballasn:
updated,

f0k:
Cool, looks good to me!

nouiz:
when travis pass, we can merge. I'll have another PR with some other related stuff (like to allow it to be inplace and doc)

theano/tensor/nnet/bn.py
```diff
@@ -0,0 +1,70 @@
...
+        it is 5-10% slower than 'high_mem' implementation. Note that 5-10% computation
+        time difference compare the batch_normalization operation only, time difference
+        between implementation is likely to be less important on the full model fprop/bprop.
+    """
+    if mode == 'low_mem':
+        elm_bn = theano.tensor.elemwise.Elemwise(scalar_op=BNComposite(dtype=inputs.dtype))
+        rval = elm_bn(inputs, mean, std, gamma, beta)
+    elif mode == 'high_mem':
+        rval = (inputs - mean) / std
+        rval = rval * gamma + beta
```
f0k: Noticed that too late, but wouldn't `rval = (inputs - mean) * (gamma / std) + beta` usually be a bit more efficient? `std` and `gamma` are usually smaller than `inputs`.

nouiz: I think you are right. But why you say too late? This is just changing the
order or a multiplication and should give the same value to eps. We do
optimization that change the eps, so I don't see a problem to change it
there. Here is the PR that do this:

https://github.com/Theano/Theano/pull/3536

thanks

f0k: "Too late" because I should have seen that while reviewing :) It's not too late to fix it, of course!

PR #3416, opened 2015-09-21
BN follow up

nouiz:
Doc, allow to be inplace, printing, var name, function name, fast_compile

All this found while reviewing/testing the BN code.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3010,7 +3010,7 @@ def mean(input, axis=None, dtype=None, op=False, keepdims=False,
 
     if dtype == 'float16' or (dtype is None and input.dtype == 'float16'):
         s = cast(s, 'float16')
-
+    s.name = 'mean'
     return s
 
 
@@ -3055,7 +3055,9 @@ def var(input, axis=None, keepdims=False):
     centered_input = input - mean_input
 
     # return the mean sqr
-    return mean((centered_input ** 2), axis, keepdims=keepdims)
+    v = mean((centered_input ** 2), axis, keepdims=keepdims)
+    v.name = 'var'
+    return v
 
 
 @constructor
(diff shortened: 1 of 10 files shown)


theano/tests/unittest_tools.py
```diff
@@ -311,8 +311,9 @@ def str_diagnostic(expected, value, rtol, atol):
         pass
     # Use the same formula as in _allclose to find the tolerance used
     narrow = 'float32', 'complex64'
-    if ((str(expected.dtype) in narrow) or
-        (str(value.dtype) in narrow)):
+    # If the user passed a list with tensor, dtype won't be defined
+    if ((str(getattr(expected, 'dtype', '')) in narrow) or
+        (str(getattr(value, 'dtype', '')) in narrow)):
```
abergeron: This will conflict with my changes for float16 since I remove this part to instead call the same code as in _allclose.

abergeron:
I'm ok with this after travis passes.

theano/tensor/basic.py
```diff
@@ -3055,7 +3055,9 @@ def var(input, axis=None, keepdims=False):
     centered_input = input - mean_input
 
     # return the mean sqr
-    return mean((centered_input ** 2), axis, keepdims=keepdims)
+    v = mean((centered_input ** 2), axis, keepdims=keepdims)
+    v.name = 'var'
+    return v
```
f0k: What about giving `std` a name as well (some lines below)? (And maybe fix the `std()` docstring, which says "It calls var() and var()".)

f0k:
LGTM.

theano/tensor/basic.py
nouiz: I'll add the name in another PR later. I don't understand the correction needed in the docstring.

f0k: > I don't understand the correction needed in the docstring.

The docstring for `std()` literally says "It calls var() and var()." That's obviously not what it does. Maybe it was meant to say "It calls var() and sqrt()." I just noticed that while checking if its result could also get a name.

PR #3421, opened 2015-09-21
More explicit description for axis insertion in shape_padaxis

bouthilx:
To fix #3318 docstrings


bouthilx:
@nouiz Is that ok?

nouiz:
excellent. thanks

PR #3432, opened 2015-09-22
crash fix and cgemv optimization

nouiz:
Work around crash in [[theano-users] Strange error with CLinker](https://groups.google.com/forum/#!msg/theano-users/6dLaEqc2R6g/51fH3TC6qf0J)

The cause of the crash isn't currently know or fixed. But I fixed the second error. This add an optimization for that use case that happen to fix the error.


abergeron:
Ok for this when travis passes.

Issue #3426, opened 2015-09-22
D3Vis not functional on windows due to carriage returns. 


sibleyd:
In testing the new D3Viz feature on windows I've encountered two issues, one windows specific and one that may be os independent. 

The main issue is that the generated HTML file contains no nodes to interact with. Inspecting the generated html reveals that for variable dotGraph the string provided to graphlibDot.read contains no information. This is the line that was generated for one of my graphs.

> var dotGraph = graphlibDot.read(""];}");

Looking into d3viz, the issue appears to be when the patterns are replaced in the template on line 106.

>'%% DOT_GRAPH %%': os.path.basename(dot_graph),

Passing dot_graph to os.path.basename results in almost all of the string being dropped, I'm not sure what the original intent of using os.path.basename was here. 

Modifying this line to 
>'%% DOT_GRAPH %%': dot_graph,

resolves this issue.

The second issue is windows specific, on line 78 when dot_graph is created it contains windows-style line breaks (\r\n), so the carriage returns are ultimately passed into the generated html. Inclusion of carriage returns causes a syntax error when attempting to parse the dotGraph in javascript. 

Modifying line 78 to 
> dot_graph = escape_quotes(graph.create_dot()).replace('\n', '').replace('\r','')

resolves this issue.

After making these two changes to d3viz I can successfully generate the visualization on windows 7x64.

nouiz:
Thanks for the investigation. Do you want to make a pr of this? That way
you will show up in the repo history.

sibleyd:
Pull should be submitted, I haven't done much with git so let me know if something has gone awry.

carriepl:
Closing the issue since the corresponding PR has been merged.

PR #3433, opened 2015-09-23
Remove Carriage returns and basename call in d3viz.

sibleyd:
Fix for issues described in #3426.


carriepl:
Thanks a lot for the bug report and the pull request. It seems good with the exception of a very small flake8 error in the added call to replace() (there's a missing whitespace after the comma). Would you mind fixing it?

Once this is done, I see no reason not to merge this.

carriepl:
Looks great. Thanks again for the report and the fix.

PR #3434, opened 2015-09-23
Don't crash with old nosetests

nouiz:
Hopefully, this will fix the doc on the web site


abergeron:
Ok when travis passes.

PR #3435, opened 2015-09-23
Scan grad speedup

carriepl:
NOTE : I'm still in the process of testing this change but it might still be good to start getting feedback. Since this changes the graphs that result from taking gradients of Scan nodes, it may mean that new optimizations that were previously not needed would now be more useful/important.

This PR modifies the way Scan takes the gradient of it's inner function. Instead of doing nb_inputs * nb_outputs calls to grad to get the gradient of its inner function, it now does only one such call. This makes grad() generally faster but even more so in the case of taking the gradient of nested scans in which the number of calls to grad() would increase exponentially with the number of levels of nested scans. Because of the reduced number of calls to grad() the resulting scan grads have simple graphs and less duplicate nodes so optimization is faster and (because we don't have optimizations for every single case) execution is sometimes faster. Also, execution with no/few optimizations (i.e. FAST_COMPILE), execution is much much faster.

On the LSTM tutorial, on my machine, the time to take the gradients goes from 0.30 to 0.15 seconds, the compilation time goes from 16s to 13s and the time to run the scan_grad drops by ~40% resulting in a global ~20% drop in execution time.

On the following (not-rigorous) graph using nested scans, the time to take the gradients goes from 6.9s to 0.7s seconds, the compilation time goes from 18s to 7s but the execution time stayed the same because the resulting Theano function was the same as without this change.
```
def test_grad_speed():
    a = tensor.fmatrix()
    b = tensor.fmatrix()
    c = tensor.fmatrix()
    d = tensor.fmatrix()
    e = tensor.fmatrix()

    def outer_step(a_1, b_1, c_1, d_1, e_1):
        def inner_step(a_2, b_2, c_2, d_2, e_2):
            total = a_2 + b_2 + c_2 + d_2 + e_2
            total = total + 1 + 1 + 1 + 1
            return total, total, total, total, total

        result_inner, _ = theano.scan(fn=inner_step,
                                      sequences=[a_1, b_1, c_1, d_1, e_1],
                                      name="inner_scan")

        total = (result_inner[0][-1] + result_inner[1][-1] +
                 result_inner[2][-1] + result_inner[3][-1] +
                 result_inner[4][-1])
        return total, total, total, total, total

    result_outer, _ = theano.scan(
        fn=outer_step,
        sequences=[a, b, c, d, e],
        name="outer_scan"
    )
```

### DIFF
--- a/theano/gradient.py
+++ b/theano/gradient.py
@@ -359,7 +359,8 @@ def Lop(f, wrt, eval_points, consider_constant=None,
 
 def grad(cost, wrt, consider_constant=None,
          disconnected_inputs='raise', add_names=True,
-         known_grads=None, return_disconnected='zero'):
+         known_grads=None, return_disconnected='zero',
+         null_gradients='raise'):
     """
     Return symbolic gradients for one or more variables with respect to some
     cost.
@@ -405,6 +406,12 @@ def grad(cost, wrt, consider_constant=None,
         - 'None' : If wrt[i] is disconnected, return value i will be
                    None
         - 'Disconnected' : returns variables of type DisconnectedType
+
+    :type null_gradients: string
+    :param null_gradients: Defines the behaviour if some of the variables in
+            ``wrt`` have a null gradient. The possibles values are :
+        - 'raise' : raise a NullTypeGradError exception
+        - 'return' : return the null gradients
 
     :rtype: variable or list/tuple of Variables (matching `wrt`)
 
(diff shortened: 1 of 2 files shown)


abergeron:
Overall I approve.  Would still have to look at the scan parts in detail before an ok though.

carriepl:
I performed some more tests:

First, @nouiz wanted me to verify that this doesn't cause a significant increase in memory consumption. I tested on the LSTM tutorial from deeplearning.net and found a 300 KB increase in memory consumption on a total of 52MB according to Theano's profile_memory output. I think that difference is pretty trivial.

I also tested on one of the larger use cases provided by a student from the lab. Without this change, time to take the gradients was 30 minutes and compilation time was 2 hours. With the change, the time to take the gradients is 45s and compilation time is 1 hour 30 minutes. Total time to get the Theano function thus went fro 2h30 to 1h30.

nouiz:
What about the run time?

carriepl:
@nouiz : Still haven't been able to run his script. Will ask the student if he can spare the time to test the  speed up himself.

@abergeron or @lamblin : can one of you have a quick look at this?

carriepl:
Oops, that was merged by accident. I made a PR to revert this.

PR #3440, opened 2015-09-25
Scan push out seq scan

### LINKED ISSUE #3282: Optimization failure (SeqOptimizer, PushOutNonSeqScan)
The following code generates an optimization failure.

    import theano
    outputs, _ = theano.scan(lambda x: (x * x, x), non_sequences=[1], n_steps=2)
    f = theano.function(inputs=[], outputs=outputs)

When run I see 6 occurrences of this warning/error sequence:

    WARNING: An optimization wanted to replace a Variable in the graph, but the replacement for it doesn't remove it. We disabled the optimization. Your function runs correctly, but it would be appreciated if you submit this problem to the mailing list theano-users so that we can fix it.
    scanOp_pushout_nonseqs_ops [(for{cpu,scan_fn}.0, Alloc.0)]
    ERROR (theano.gof.opt): SeqOptimizer apply <theano.scan_module.scan_opt.PushOutNonSeqScan object at 0x0000000004C33FD0>
      File "<snip>\theano\theano\gof\opt.py", line 194, in apply
      File "<snip>\theano\theano\gof\opt.py", line 79, in optimize
      File "<snip>\theano\theano\scan_module\scan_opt.py", line 228, in apply
        self.process_node(fgraph, node)
      File "<snip>\theano\theano\scan_module\scan_opt.py", line 401, in process_node
        reason='scanOp_pushout_nonseqs_ops')
      File "<snip>\theano\theano\gof\toolbox.py", line 324, in replace_all_validate_remove
        raise ReplacementDidntRemovedError()
    ReplacementDidntRemovedError

Then a single instance of:

    ERROR (theano.gof.opt): EquilibriumOptimizer max'ed out by 'all_pushout_opt'. You can safely raise the current threshold of 5.000000 with the theano flag 'optdb.max_use_ratio'.

A few observations:
* Setting `n_steps=1` avoids the problem.
* Returning just `x` from the lambda avoids the problem.
* Returning just `x * x` from the lambda avoids the problem.
* Returning `(x + x, x)` from the lambda also causes the problem.
* Returning `(x, x + x)` from the lambda also causes the problem.

carriepl:
Fixes #3282

### DIFF
--- a/theano/scan_module/scan_opt.py
+++ b/theano/scan_module/scan_opt.py
@@ -397,10 +397,20 @@ class PushOutNonSeqScan(gof.Optimizer):
             # because the scan op expects for a tensor3, to which an
             # subtensor is applied that takes only the last element
             if replace_with:
-                fgraph.replace_all_validate_remove(
-                    replace_with.items(),
-                    remove=[node],
-                    reason='scanOp_pushout_nonseqs_ops')
+                if len(node.outputs) == len(replace_with):
+                    # Every output of the node has a replacement, the Scan
+                    # node can be removed from the graph
+                    fgraph.replace_all_validate_remove(
+                        replace_with.items(),
+                        remove=[node],
+                        reason='scanOp_pushout_nonseqs_ops')
+                else:
+                    # The node has some outputs for which no replacement has
+                    # been established. The replacements can be performed but
+                    # the Scan node can't be removed at this point.
+                    fgraph.replace_all_validate(
+                        replace_with.items(),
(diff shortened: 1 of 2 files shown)


carriepl:
Improved the comment as requested by @nouiz.

nouiz:
I'm ok with this when tests pass and someone else review this

carriepl:
@lamblin or @abergeron : can one of you spare a few minutes?

carriepl:
(Travis was killed on the last commit because it's only a comment change)

PR #3442, opened 2015-09-25
update numpy version to 1.7.0

nouiz:
1.7.0 was released in 2012 or 2013.

This will make sure numpy support multiple axis on some reduction function.


carriepl:
Looks fine.

PR #3450, opened 2015-09-26
Now inc_subtensor is OK with Print

### LINKED ISSUE #3449: theano.printing.Print doens't work with inc_subtensor
If I do something like

```
T.subtensor(theano.printing.Print('slice')(M[3:5]), delta)
```
I get an error `TypeError: x must be the result of a subtensor operation`

eulerreich:
fixes #3449

### DIFF
--- a/theano/tensor/subtensor.py
+++ b/theano/tensor/subtensor.py
@@ -1081,7 +1081,9 @@ def inc_subtensor(x, y, inplace=False, set_instead_of_inc=False,
 
     if not x.owner:
         raise TypeError('x must be the result of a subtensor operation')
-
+        
+    if isinstance(x.owner.op, theano.printing.Print):
+        x = x.owner.inputs[0]
     # retrieve idx_list from x.owner
     if isinstance(x.owner.op, Subtensor):
         if tolerate_inplace_aliasing:
--- a/theano/tensor/subtensor.py
+++ b/theano/tensor/subtensor.py
@@ -1081,7 +1081,7 @@ def inc_subtensor(x, y, inplace=False, set_instead_of_inc=False,
 
     if not x.owner:
         raise TypeError('x must be the result of a subtensor operation')
-        
+
     if isinstance(x.owner.op, theano.printing.Print):
         x = x.owner.inputs[0]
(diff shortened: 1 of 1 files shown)


nouiz:
Thanks for the report and working on fixing this. This PR will remove the crash, but it completly remove the print at the same time. I think you need to keep it. So where you have code, we should keep the print in a variable and reuse it later.

Do you want to finish this fix? I'm not sure when I can work on it.

thanks

eulerreich:
Yea I'll take a look

eulerreich:
Looks like the checks passed. The CI error was just style. For me, the `Print` works as usual. Let me know if the PR still "complet[e]ly remove[s] the print" for you.

nouiz:
I updated gh-3449 example. It show the problem I tell that the Print get lost.

theano/tensor/subtensor.py
```diff
@@ -1082,6 +1082,8 @@ def inc_subtensor(x, y, inplace=False, set_instead_of_inc=False,
     if not x.owner:
         raise TypeError('x must be the result of a subtensor operation')
 
+    if isinstance(x.owner.op, theano.printing.Print):
+        x = x.owner.inputs[0]
```
lamblin: You would have to keep a reference to that `Print` Op, and apply it on the new Op later.
For instance, rebuild a Print op that gets executed after the AdvancedSubtensor, and prints the value of the original Subtensor.

nouiz:
I think this can't be done without changing the graph. The current Theano graph won't have the subtensor op in the graph. This will be internal to the IncSubtensor op. So this will need to change all *inc*subtensor op to implement the print.

I think we need to create a new subclass of ViewOp, that will take an extra input and do nothing with it, but force its computation. With that, we could instantiate that ViewSideEffect op on the input of IncSubtensor.

This need to be on the input of IncSubtensor, otherwise we will execute the print after the computation. It will be possible to trigger it for PdbBreakpoint op too. It will need opt to move it correctly to the GPU. This will execute the *subtensor* twice, but will do that the user want.

So this is not the right direction, I'll close this PR. I'll put this information on the original ticket.

PR #3453, opened 2015-09-28
Fix stack(tensors, axis=0) documentation

### LINKED ISSUE #3446: Deprecation warning for tensor.stack
In [the doc](http://deeplearning.net/software/theano/library/tensor/basic.html#theano.tensor.stack), the deprecation for `stack(*tensors)` is placed under the entry for `stack(tensors, axis=0)`, which is unclear whether there is a typo and `stack(tensors, axis=0)` is deprecated, or `stack(*tensors)` is deprecated but the warning is misplaced. I'm guessing it's the second one.

bouthilx:
Fix #3446.

The deprecation warning message was under the new stack() interface while it should be under the old stack() interface. 

I also added examples.

### DIFF
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -703,30 +703,60 @@ Creating Tensor
               except for the main diagonal, whose values are equal to one. The output
               will have same dtype as `x`.
 
-
 .. function:: stack(tensors, axis=0)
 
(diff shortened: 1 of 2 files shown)


nouiz:
tests are failing

bouthilx:
Thanks for letting me know. I amended the commit.

Issue #3451, opened 2015-09-27
tensor.prod arithmetic mistake


jsutton9:
There seems to be a bug relating to the way tensor.prod and tensor.sum are optimized.
```
import theano
import theano.tensor as T
import numpy as np
a = T.matrix("a")
b = T.prod(a, axis=1)
c = T.sum(b)

f = theano.function([a], [b, c])
f_bug = theano.function([a], c)

a_val = np.matrix([[0.1, 0.2], [0.3, 0.5]], dtype="float32")

b_val, c_val = f(a_val)
c_val_bug = f_bug(a_val)

print c_val     # this is (0.1*0.2) + (0.3*0.5) =~ 0.17
print c_val_bug # this is (0.1+0.2) + (0.3+0.5) =~ 1.10
```

nouiz:
thanks for the report. I have a fix here:

https://github.com/Theano/Theano/pull/3455

PR #3455, opened 2015-09-28
[BUG] in opt local_op_of_op

nouiz:
@carriepl can you review this?

Can you also check the other sum/prod opt changed at the same time if they have this problem? Just in case.

the opt local_op_of_op was merging 2 consecutive sum or prod. But it was also merging 1 sum followed by a prod (or the reverse)

 fix gh-3451


carriepl:
Looks good to me.

PR #3460, opened 2015-09-28
Revert "Scan grad speedup"

carriepl:
Reverts Theano/Theano#3435


carriepl:
Merging since this reverts the last PR to have been merged so it should put the master at the same state as it was before.

PR #3465, opened 2015-09-29
Fix arange to avoid overflow.

abergeron:


nouiz:
I'm ok to merge when travis pass.

nouiz:
travis fail.

nouiz:
travis fail.

abergeron:
Now fixed.

nouiz:
travis still fail

abergeron:
Now those pass locally and so should pass on travis.

nouiz:
@abergeron email theano-dev, theano-user to tell them about this interface change.

abergeron:
Ok, done.  But I am considering adding a new casting policy "simple" which would depend on two params:  intX and floatX.

The policy is: for ints always use intX and for floats always use floatX.

I would default intX to int64.  I think that would help a lot of cases that don't really need a fine-grained approach that tries to find the smallest applicable type.

nouiz:
For intX, it is not a new idea, but would be time consumming and could
cause bugs...

Currently, we explicitly made the shapes return int64 on all platform as it
was too complicated to use int32 on 32bit OS/computers.

Before working on this, we should take the time to plan and discuss what we
need. It isn't trivial. @delallea already speed about 2 weeks full time if
not more on this numpy+floatX approch. @lamblin also know well the
difficulty associated with this.

But we can start to think more about this and list the case that we want to
fix and the possible fix and the consequence they would have...

abergeron:
Any approach that tries to replicate the numpy behaviour is much more complex than what I propose.

It should only need a modification in scalar.upcast, without reaching out everywhere like numpy or numpy+floatX does.

nouiz:
I would be surprised that your change is only local to that method. There
is place where upcast isn't used... The fact that we don't have just a few
point where the dtype if selected is part of the difficulty.

A good refactor is to lower the number of such place...

But we should discuss that in an issue, not this PR. Can you open one with
your idea explained?

abergeron:
Ok

delallea:
About the numpy+floatX thingie, yes it turned out to be a lot trickier than expected. I don't remember the details, but I ran into a non-trivial blocking issue that I never got to tackle. I guess it would be cleaner to actually remove that code since it's not fully working.

nouiz:
That code is still useful to know where modification are me needed. But we
can remove it when we have something else.
Le 3 oct. 2015 15:46, "Olivier Delalleau" <notifications@github.com> a
écrit :

PR #3470, opened 2015-09-29
Use make_new_inplace mechanism to fix optimization with cast

lamblin:
To be tested as I don't have a case to reproduce it.
Also, this case should only be triggered when both input and output of the cast have the same dtype, so maybe this node should be removed altogether earlier on.


nouiz:
We can merge when travis pass and the users confirmed that it fix their problem. Ideally, we should test this.

nouiz:
The user who report that tell that this fix there problem.

PR #3471, opened 2015-09-30
Multi gpu operation for gpuarray

abergeron:
This enables support for multiple gpus in a single function in gpuarray.

Note that scan nodes have to run all their internal code on the same gpu for now.  This may get fixed later, but is too much trouble right now.


nouiz:
travis is failing.

abergeron:
I'll have to investigate a bit more to be sure, so I'll close this one for now.

PR #3475, opened 2015-09-30
Scan backend speedup

carriepl:


carriepl:
Some tests are failing in FAST_COMPILE. Currently looking at it.

carriepl:
Done and Travis passes.

theano/scan_module/scan_perform.pyx
```diff
@@ -256,9 +254,11 @@ def perform(
...
     other_args = args[offset:]
     input_storage = fnct.input_storage
-    len_input_storage = len(input_storage)
-    old_input_storage = [None] * len_input_storage
-    old_input_data = [None] * len_input_storage
+    nb_mitmot_in = 0
+    for idx in range(n_mit_mot):
+        nb_mitmot_in += tap_array_len[idx]
+    old_mitmot_input_storage = [None] * nb_mitmot_in
+    old_mitmot_input_data = [None] * nb_mitmot_in
```
abergeron: You might be able to go even faster by making these C arrays instead of python lists.

nouiz:
I merge now as this PR strictly bring Theano in a better state and it solve a regression in some cases.

The comments are future enhancement.

PR #3476, opened 2015-09-30
Multiple fixes preparing for multi-gpu

abergeron:
This has be split off to ease reviewing.


abergeron:
Changes done.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -312,254 +309,29 @@ def infer_shape(self, node, xshp):
 
     def c_code(self, node, name, inputs, outputs, sub):
         return """
+        PyGpuArrayObject *%(name)s_tmp;
+        %(name)s_tmp = PyArray_GETCONTIGUOUS(%(inp)s);
```
nouiz: Why this is here? Why it isn't in pygpu_fromhostdata?

abergeron: Because pygpu_fromhostdata() does not have access to the ndarray object and currently tries to copy the data as-is.  I've realized that it's not a good idea and made a ticket to change that in https://github.com/Theano/libgpuarray/issues/90

I made a ticket that might help at

nouiz: What do you mean by copy data as-is? It receive the stride. So it don't
need access to the ndarray object.

lamblin:
If the last commit is enough to make travis pass, I'm OK with merging.

theano/sandbox/gpuarray/basic_ops.py
abergeron: I mean by not trying to make it contiguous beforehand.  It finds the lowest and higest elements and copies all of the data in between (even if that data is not part of the array according to strides and dims).

I want to have libgpuarray make a contiguous version, but this PR can't do that since it's a separate repo, hence the issue.

Issue #3412, opened 2015-09-18
Extremely slow compilation / crash


ebenolson:
This issue has been discussed previously here: https://groups.google.com/forum/#!topic/lasagne-users/XR-jmU1pMWc

A test case is here: https://gist.github.com/ebenolson/4b2fe40ff677db7ca496

The problem seems to be related to inference of constant bounds for the lines

    input = conved[:, :, shift:input.shape[2] + shift]

When using `device=cpu`, the function compiles correctly, but each additional layer takes a far longer time to complete. Putting `None` into the `image_shape` removes the problem.

When using `device=gpu` and cuDNN, compilation fails after several minutes with an error:

    <<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.AssertionError'> ('Rebroadcast needs bool for new broadcast pattern. Got ', False) LocalOptGroup(local_dnn_conv_inplace,local_dnn_convgw_inplace,local_dnn_convgi_inplace)
    ERROR (theano.gof.opt): Optimization failure due to: LocalOptGroup(local_dnn_conv_inplace,local_dnn_convgw_inplace,local_dnn_convgi_inplace)
    ERROR:theano.gof.opt:Optimization failure due to: LocalOptGroup(local_dnn_conv_inplace,local_dnn_convgw_inplace,local_dnn_convgi_inplace)
    ERROR (theano.gof.opt): TRACEBACK:
    ERROR:theano.gof.opt:TRACEBACK:
    ERROR (theano.gof.opt): Traceback (most recent call last):
      File "/home/eben/projects/Theano/theano/gof/opt.py", line 1797, in process_node
        fgraph.replace_all_validate(repl_pairs, reason=lopt)
      File "/home/eben/projects/Theano/theano/gof/toolbox.py", line 295, in replace_all_validate
        fgraph.replace(r, new_r, reason=reason, verbose=False)
      File "/home/eben/projects/Theano/theano/gof/fg.py", line 545, in replace
        new_r2 = r.type.convert_variable(new_r)
      File "/home/eben/projects/Theano/theano/sandbox/cuda/type.py", line 268, in convert_variable
        return theano.tensor.patternbroadcast(var, self.broadcastable)
      File "/home/eben/projects/Theano/theano/tensor/basic.py", line 3598, in patternbroadcast
        for i in xrange(len(broadcastable))])(x)
      File "/home/eben/projects/Theano/theano/compile/ops.py", line 664, in __init__
        broad)
[...]

PR #3477, opened 2015-09-30
Crash gpu and opt speed up

nouiz:
This fix gh-3412. It fix the crash and the slow compilation time.

Should also close gh-3501 about the warnings

### DIFF
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -200,6 +200,12 @@ optdb.register('merge1', gof.MergeOptimizer(),
 # rearranges elemwise expressions
 optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
                1, 'fast_run', 'fast_compile')
+# Register in the canonizer Equilibrium as a local opt the merge opt.
+# Without this, as the equilibrium have ignore_newtrees=False, we
+# won't merge all nodes if it is set as a global optimizer with
+# final_opt=True.
+optdb['canonicalize'].register("merge", gof.opt.merge_optimizer, 'fast_run',
+                               "fast_compile", cleanup=True)
 
 optdb.register('merge1.2', gof.MergeOptimizer(),
                1.2, 'fast_run', 'fast_compile', 'merge')
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -204,7 +204,10 @@ optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
 # Without this, as the equilibrium have ignore_newtrees=False, we
 # won't merge all nodes if it is set as a global optimizer with
 # final_opt=True.
-optdb['canonicalize'].register("merge", gof.opt.merge_optimizer, 'fast_run',
+
+# We need a new instance of MergeOptimizer to don't have its name
+# changed by other usage of it.
+optdb['canonicalize'].register("merge", gof.opt.MergeOptimizer(), 'fast_run',
                                "fast_compile", cleanup=True)
 
 optdb.register('merge1.2', gof.MergeOptimizer(),
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -198,7 +198,7 @@ optdb.register('merge1', gof.MergeOptimizer(),
                0, 'fast_run', 'fast_compile', 'merge')
 
 # rearranges elemwise expressions
-optdb.register('canonicalize', gof.EquilibriumDB(),
+optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
                1, 'fast_run', 'fast_compile')
 
 optdb.register('merge1.2', gof.MergeOptimizer(),
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -200,7 +200,7 @@ optdb.register('merge1', gof.MergeOptimizer(),
 # rearranges elemwise expressions
 optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
                1, 'fast_run', 'fast_compile')
-# Register in the canonizer Equilibrium as a local opt the merge opt.
+# Register in the canonizer Equilibrium as a clean up opt the merge opt.
 # Without this, as the equilibrium have ignore_newtrees=False, we
 # won't merge all nodes if it is set as a global optimizer with
 # final_opt=True.
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -547,6 +547,7 @@ class CLinker(link.Linker):
         if no_recycling is None:
             no_recycling = []
         if self.fgraph is not None and self.fgraph is not fgraph:
+            # A linker can be tied to only one FunctionGraph.
             return type(self)(self.schedule).accept(fgraph, no_recycling)
         self.fgraph = fgraph
         self.fetch_variables()
@@ -1750,14 +1751,13 @@ class OpWiseCLinker(link.LocalLinker):
         if no_recycling is None:
             no_recycling = []
         if self.fgraph is not None and self.fgraph is not fgraph:
+            # A linker can be tied to only one FunctionGraph.
             return type(self)(
                 fallback_on_perform=self.fallback_on_perform,
                 allow_gc=self.allow_gc,
                 nice_errors=self.nice_errors,
                 schedule=self.schedule,
             ).accept(fgraph, no_recycling)
-            # raise Exception("Cannot accept from a Linker that is
-            # already tied to another FunctionGraph.")
         self.fgraph = fgraph
         self.no_recycling = no_recycling
         return self
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1981,13 +1981,22 @@ class EquilibriumOptimizer(NavigatorOptimizer):
 
     Parameters
     ----------
-    optimizers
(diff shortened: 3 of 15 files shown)


abergeron:
Should this be applied to gpuarray also?

nouiz:
I checked before the PR and it don't need this.

nouiz:
I added a speed optimization that go in one case from 93s to .3s! About 300x faster.

This fix gh-3412.

abergeron:
I wholeheartedly approve of the constant folding changes.

But I have some reservations about the loop at the end of local_fill_sink.  If @lamblin is fine with it, the I'm ok since I did not invest a large amount of time to try and understand it.

I'm ok with all the other changes.

nouiz:
I added another compilation crash to this PR.

nouiz:
In the GSoC case about optimization, the compilation was getting slower du to the canonizer doing more pass. It went from (total optimization time) 508s(9 canonizer pass) to  547s. (12 canonizer pass). But this also made the final graph smaller from 6233 nodes to 6208.

So I changed the canonizer to add in the current pass the new nodes. This bring back the opt time to 502s(7 canonizer pass) and 6195 nodes.

I think we can merge this PR.

This make the gh-3412 case go from original 93 to 0.3s to 0.4s. The small slow down in that case don't seem a problem. It could just be du to my computer node. So the speed up in that case is 230x.

lamblin:
What about removing the replaced variables without clients from the replacement list in `local_fill_sink`? Does that make a difference in the end?

abergeron:
I'm sorry, but this is not OK.

If you have some new changes to propose, please use a new PR for them.  Piling in tons of unrelated changes in a single PR makes this a nightmare to review and appending stuff all the time make the essential fixes contained at the beginning unnecessarily delayed.

Also, you broke something in tensor/nnet/tests/test_sigm.py, but I'm not sure what.

nouiz:
I updated this PR to fix travis. But I just taught that this would make some node not being optimized in the current pass of the EQ. So I'll change it again and killed travis.

nouiz:
Finally, it was fine due to an unknow properties. I added this information in a comment to tell that we need that properties.

theano/gof/opt.py
```diff
@@ -1770,6 +1771,8 @@ def process_node(self, fgraph, node, lopt=None):
                 raise
         if replacements is False or replacements is None:
             return False
+        if replacements is True:
+            return True
```
lamblin: What does that mean for `replacements` to be `True`? This is not documented in http://deeplearning.net/software/theano/extending/optimization.html#global-and-local-optimizations, nor in the docstring of `LocalOptimizer.transform` https://github.com/nouiz/Theano/blob/crash_gpu/theano/gof/opt.py#L1000

nouiz: It mostly mean, I did some change directly on the fgraph. I agree, this should be documented. But before that, we should agree that this is acceptable.

lamblin: You mean that a `LocalOptimizer` could change the graph directly, without going through `replace_all_validate` and all that? That does not seem correct to me. And I don't think the comments of a PR is the right place to discuss such a change.

theano/tensor/opt.py
```diff
@@ -502,9 +504,27 @@ def register(inner_lopt):
...
+# 2) after an local optimization being applied, if the
+#    current node is still in the graph, it will continue to the next
+#    local optimizer. So this won't trigger more iteration.
+def add_merge_feature(fgraph):
+    if not hasattr(fgraph, 'merge_feature'):
+        fgraph.attach_feature(theano.gof.opt.MergeFeature())
+
+
+@register_canonicalize('fast_compile', 'merge')
+@gof.local_optimizer(None, requirements=[add_merge_feature])
```
lamblin: Do we really need to change the signature of `gof.local_optimizer` for that? Could we use `AddFeatureOptimizer` (currently in `theano/compile/mode.py`) instead to make sure the feature is there?

theano/tensor/opt.py
```diff
@@ -502,9 +504,27 @@ def register(inner_lopt):
...
+def local_merge_optimizer(node):
+    if node.fgraph.merge_feature.scheduled:
+        ret = merge_optimizer(node.fgraph)
+        return ret[5] > 0
```
lamblin: What is at position 5? What does it mean?
Also, it does not seem right to have a global optimization pretending to be a local one.

theano/tensor/opt.py
```diff
@@ -1454,13 +1639,8 @@ def local_fill_sink(node):
             new_client.owner.outputs[0].clients = client.outputs[0].clients
             r = local_fill_sink.transform(new_client.owner)
             if not r:
-                all_clients_replaced = False
                 continue
             replacements.update(r)
-        else:
-            all_clients_replaced = False
-    if all_clients_replaced:
-        replacements.pop(node.outputs[0], None)
```
lamblin: Did you time the test cases without that modification?

theano/tensor/opt.py
lamblin: Actually, could we simply check if any of `node.outputs` are in `scheduled`, and just return the pair to be replaced? That would actually be a local optimizer.

theano/tensor/opt.py
nouiz: I don't see how to reuse AddFeatureOptimizer for that.

theano/gof/opt.py
nouiz: I have an idea to implement it in a cleaner way. I'll update this PR later with this. Mostly, add an cleanup tag to Equilibrium opt as we did for final_optimizers.

theano/tensor/opt.py
nouiz: that will get removed by my interface changed I have in mind. Mostly ret[5] is the number of node merged, so it was returning True if it merged something.

theano/compile/mode.py
```diff
@@ -196,8 +196,14 @@ def apply(self, fgraph):
                0, 'fast_run', 'fast_compile', 'merge')
 
 # rearranges elemwise expressions
-optdb.register('canonicalize', gof.EquilibriumDB(),
+optdb.register('canonicalize', gof.EquilibriumDB(ignore_newtrees=False),
                1, 'fast_run', 'fast_compile')
+# Register in the canonizer Equilibrium as a local opt the merge opt.
```
lamblin: Maybe say "a 'cleanup' opt" rather than "local".

theano/gof/cc.py
```diff
@@ -547,9 +547,7 @@ def accept(self, fgraph, no_recycling=None):
         if no_recycling is None:
             no_recycling = []
         if self.fgraph is not None and self.fgraph is not fgraph:
-            return type(self)().accept(fgraph, no_recycling)
-            # raise Exception("Cannot accept from a Linker that is already"
-            #                " tied to another FunctionGraph.")
+            return type(self)(self.schedule).accept(fgraph, no_recycling)
```
lamblin: Maybe add a comment explaining that we have to create a new `Linker` instance, because a Linker can only be tied to one FunctionGraph.

theano/gof/cc.py
```diff
@@ -1720,7 +1718,8 @@ def accept(self, fgraph, no_recycling=None):
             return type(self)(
                 fallback_on_perform=self.fallback_on_perform,
                 allow_gc=self.allow_gc,
-                nice_errors=self.nice_errors
+                nice_errors=self.nice_errors,
+                schedule=self.schedule,
             ).accept(fgraph, no_recycling)
             # raise Exception("Cannot accept from a Linker that is
             # already tied to another FunctionGraph.")
```
lamblin: Idem here, so we can remove the commented-out exception.

theano/gof/opt.py
```diff
@@ -1995,7 +1996,8 @@ def __init__(self,
                  failure_callback=None,
                  ignore_newtrees=True,
                  max_use_ratio=None,
-                 final_optimizers=None):
```
lamblin: Can you quickly document `final_optimizers` and `cleanup_optimizers` in the docstring?

theano/gof/opt.py
```diff
@@ -2134,12 +2154,18 @@ def pruner(node):
                         t_opt = time.time()
                         lopt_change = self.process_node(fgraph, node, lopt)
                         time_opts[lopt] += time.time() - t_opt
+                        # TODO: if not ...: continue
```
lamblin: We should probably do it now or file a ticket, but not add TODO in comments.

theano/tensor/opt.py
lamblin: To make it more explicit: if adding back those lines do not lead to a speed up, then it is better to remove them.

lamblin:
Also, a rebase will be needed.

theano/tensor/opt.py
nouiz: I timmed this again. If I add them back, it don't finisih compilation with the GSoC slow case in multiple hours. Otherwise, it finish in 471s.

nouiz:
I think this PR is ready to merge. In the GSoC use case, this PR give the same as the master 2928d02a.

For the slow case, it speed up the master commit at 75.17s to 0.428s. This give a speed up of 175x

theano/tensor/opt.py

theano/gof/opt.py
lamblin: Done.

Issue #3429, opened 2015-09-22
Python with statement, use it


nouiz:
We should modify Theano to use it isntead of the old python 2.4 syntax now.

Which grep statement would help for that?

piotrfrankowski:
Hi,
If you have installed pcregrep (pcre package), you can use command:
```find -name "*.py" | xargs pcregrep --color="auto" --buffer-size=1000k -M "try:[A-Za-z0-9\t ,./~\!\@\#\$%\^&\*\(\)\-=_+\[\]{};:\'\"\\|\n]*?finally:"``` 
I'm working on option "doesn't contain 'except' betweet 'try' and 'finally'".

**Comment by nouiz** [comment]
Thanks for working on this.
Le 29 sept. 2015 15:11, "Piotr Frankowski" <notifications@github.com> a
écrit :

**Comment by piotrfrankowski** [comment]
Hi,
Have you started working on this refactoring? Maybe I can start contributing to Theano?

**Comment by nouiz** [comment]
You are the only one working on this. It would be a good first contribution!

Just to help on getting the pr merged, don't make it to big and try to
split it in reasonable sized commit. The otherwise the review is more
complicated and this could need rebase and other related changes.

Thanks
Le 30 sept. 2015 14:50, "Piotr Frankowski" <notifications@github.com> a
écrit :

**Comment by piotrfrankowski** [comment]
Commit per file would be ok?

**Comment by piotrfrankowski** [comment]
Ok, I've started refactoring. First of all, I'm searching for
```
<variable> = open(<args>)
<do something with variable>
<variable>.close()
```
But I found some examples:
```
try:
     open(<some file>, 'w').close()
except Exception as e:
```
Should I refactor it?

nouiz:
Refactor it if this make the code more clear. If not, it isn't useful.

lamblin:
Done in #3478.

PR #3478, opened 2015-10-01
Issue  #3429 - refactor some code to python 'with' statement

piotrfrankowski:
Hi,
I've create pr for refactoring. @nouiz take a look. #3429 
The 'tests' module is little tricky than rest of pr.

### DIFF
--- a/theano/gof/compiledir.py
+++ b/theano/gof/compiledir.py
@@ -351,25 +351,28 @@ def print_compiledir_content():
     for dir in os.listdir(compiledir):
         filename = os.path.join(compiledir, dir, "key.pkl")
         with open(filename, 'rb') as file:
-            keydata = pickle.load(file)
-            ops = list(set([x for x in flatten(keydata.keys)
-                            if isinstance(x, theano.gof.Op)]))
-            if len(ops) == 0:
-                zeros_op += 1
-            elif len(ops) > 1:
-                more_than_one_ops += 1
-            else:
-                types = list(set([x for x in flatten(keydata.keys)
-                                  if isinstance(x, theano.gof.Type)]))
-                table.append((dir, ops[0], types))
+            try:
+                keydata = pickle.load(file)
+                ops = list(set([x for x in flatten(keydata.keys)
+                                if isinstance(x, theano.gof.Op)]))
+                if len(ops) == 0:
+                    zeros_op += 1
+                elif len(ops) > 1:
+                    more_than_one_ops += 1
+                else:
+                    types = list(set([x for x in flatten(keydata.keys)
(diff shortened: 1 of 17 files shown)


nouiz:
I didn't spot anything by review. We can merge when travis pass.

theano/gof/compiledir.py
```diff
@@ -349,35 +349,27 @@ def print_compiledir_content():
...
-
-                size = os.path.getsize(filename)
-                total_key_sizes += size
-                if size > max_key_file_size:
-                    big_key_files.append((dir, size, ops))
-
-                nb_keys.setdefault(len(keydata.keys), 0)
-                nb_keys[len(keydata.keys)] += 1
-            except IOError:
-                pass
```
lamblin: Here, we were ignoring `IOError`, but the new behaviour would not catch it.
@nouiz could it be an issue?

lamblin:
The other ones look fine to me, but that `try: ... except IOError: pass` should still be inside the `with` block.

theano/gof/compiledir.py
nouiz: I think it is ok. If not, it will pop again. This is not critical code, that is fine. It could be to handle broken pickled files... That is why I didn't say anything when I saw it.

But as there could be a use, @piotrfrankowski can you add it back?

abergeron: I'm pretty sure the except IOError case is to catch when the 'key.pkl' file does not exist.  With the new code you'll get an exception, which would break loading a cache that has some elements with no 'key.pkl'.

So the try..expect should go outside of the with statement because it applies to the open() which moved.

piotrfrankowski: I'll add condition at the begin, 
```
if os.isfile(filename):
   with...
```

lamblin: There would still be a race condition if the file exists during the first check, but not when you try to read it.

nouiz: Theano very rarely delete key.pkl files. It do this only when they are very old or when the user do this.

So yes, if would be ideal to prevent this race-condition, but in practice, it probably won't happen.

As it wasn't done, can you add the catch to be safe but add a comment to tell why it is like this? That way the next person will know why.

nouiz:
I think except for the small comment, this PR is close to merge. Could you finish this?

thanks.

piotrfrankowski:
I'll try to finish it in this week.

nouiz:
thanks.

piotrfrankowski:
Done, sorry for late response.

nouiz:
I don't know why we didn't get to merge this PR earlier. It is good, so merging. And happily, no conflict!

Issue #950, opened 2012-09-24
Refactor Flatten, missing opt?


nouiz:
We have the Flatten op. Why do we need it? We could reuse the reshape op. Having less op make the optimization easier.

We can reuse reshape(-1), this won't introduce bigger or uglier graph

TODO:
- [ ] Comment the Flatten op and use reshape(-1) where needed.
- [ ] Check optimization that would check for flatten, update them to check for the old Flatten and the new pattern. Maybe refactor this into a method?
- [ ] The user flatten interface should stay and should introduce reshape(-1)
- [ ] Then uncomment the Flatten op, change its description to tell it is deprecated (needed to unpickle old graph).

nouiz:
Just to node, I updated the description to tell more detail on what is needed at the end of the description

lamblin:
I believe this is closed by #3480, let's re-open it if something is missing.

Issue #3420, opened 2015-09-21
Rewrite Reshape.infer_shape and fix TestInferShape for flatten()


bouthilx:
[Reshape.infer_shape](https://github.com/bouthilx/Theano/blob/sparse_block_dot/theano/tensor/basic.py#L4018-L4066) is using Shape_i() which is adding a Reshape operation and makes [TestInferShape](https://github.com/bouthilx/Theano/blob/sparse_block_dot/theano/tensor/tests/test_basic.py#L6942-L6961) fail. It is not necessary to use Shape_i(), and the method should be rewritten to support -1 values in the new shape dimensions of Reshape.

There is two main cases

1. new shape dimensions are all constant
    1. new shape dimensions contain at most one -1
    2. new shape dimensions are all positive
2. new shape dimensions contain symbolic variables (switch is necessary)

For -1 dimensions, the value is inferred from the length of the array and remaining dimensions.

```newshape[i] == -1 ? prod(all inputs dim)/prod(all newshape dim but i) : newshape[i]```

SinaHonari:
Reading the code and the description above, the only part that is I see is not following the description is in case 2, where
a switch should be used. Is there any other part in infer_shape that is problematic?

The above description tells how Shape_i should be working, but I'm not sure if it also tells which part of it is problematic? The support for case 1 to support -1 is already there.

bouthilx:
Yes, Shape_i is used in case 2. It seems case 1 is already implemented properly except for l.4050 as it fails for scalars. 

There should be no Shape_i in ```infer_shape``` methods at all otherwise it reinserts the op in the shape graph. Case 2 should be implemented using a switch on ```node.inputs``` and ```ishapes``` only (not ```node.outputs```).

SinaHonari:
OK, in l.4050 when there is an scalar, [-1] should be returned or an empty list []?

bouthilx:
I  it would be ```[]``` but it seems to raised errors in #3480 so I'm not sure. We would need to investigate to understand what caused the errors. You can either start with this or by implementing cases 2 properly.

SinaHonari:
I'm using the following for case 2, but I get tuple index out of range error:

return [tuple([switch(eq(node.inputs[1][i], -1),
	                                  ishapes[0][i], node.inputs[1][i])
 	                           for i in xrange(self.ndim)])]

SinaHonari:
ishapes[0][i] apparently does not hold the right content.
It failing for test_basic.T_reshape
[...]

PR #3480, opened 2015-10-01
Refactor Flatten

SinaHonari:
Fixes issue #950

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -323,6 +323,11 @@ if cuda_available:
             ftensor3, ftensor4,
             scalar, vector, matrix, row, col,
             tensor3, tensor4)
+    warnings.warn(
+        "GpuFlatten class is deprecated, "
+        "please use gpu_flatten method instead.",
+        DeprecationWarning,
+        stacklevel=4)
     from .basic_ops import (host_from_gpu, gpu_from_host,
             as_cuda_array, as_cuda_ndarray_variable)
     import cuda_ndarray
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -6,6 +6,7 @@ import os
 import shutil
 import stat
 import sys
+import warnings
 
 import theano
 from theano.compat import get_unbound_function
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -318,7 +318,7 @@ if cuda_available:
             GpuDimShuffle, GpuCAReduce, GpuReshape, GpuContiguous,
             GpuSubtensor, GpuIncSubtensor,
             GpuAdvancedSubtensor1, GpuAdvancedIncSubtensor1,
-            GpuFlatten, GpuShape, GpuAlloc, GpuAllocEmpty, GpuSplit,
+            gpu_flatten, GpuFlatten, GpuShape, GpuAlloc, GpuAllocEmpty, GpuSplit,
             GpuJoin, fscalar, fvector, fmatrix, frow, fcol,
             ftensor3, ftensor4,
             scalar, vector, matrix, row, col,
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -324,11 +324,6 @@ if cuda_available:
             ftensor3, ftensor4,
             scalar, vector, matrix, row, col,
             tensor3, tensor4)
-    warnings.warn(
-        "GpuFlatten class is deprecated, "
-        "please use gpu_flatten method instead.",
-        DeprecationWarning,
-        stacklevel=4)
     from .basic_ops import (host_from_gpu, gpu_from_host,
             as_cuda_array, as_cuda_ndarray_variable)
     import cuda_ndarray
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3326,16 +3326,12 @@ class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
     """
     Implement Flatten on the gpu.
 
+    Note: The interface GpuFlatten is deprecated, you should use gpu_flatten
     """
 
     def make_node(self, x):
-        warnings.warn(
-            "GpuFlatten class is deprecated, "
-            "please use gpu_flatten method instead.",
-            DeprecationWarning,
-            stacklevel=4)
         assert isinstance(x.type, CudaNdarrayType)
-        rval = tensor.Reshape.make_node(self, x, [tensor.prod(x.shape)])
+        rval = flatten(x)
         host_out_broadcastable = rval.outputs[0].type.broadcastable
         out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
         return Apply(self, [x], [out_type()])
@@ -3343,9 +3339,12 @@ class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
 
 
 def gpu_flatten(x, outdim=1):
+    """
+    Implement flatten on the gpu.
+    """
     x = as_cuda_ndarray_variable(x)
     if outdim > 1:
-        dims = tuple(x.shape[:outdim-1])+(theano.tensor.prod(x.shape[outdim-1:]),)
+        dims = tuple(x.shape[:outdim-1])+(-1,)
     else:
         dims = (-1,)
     return  GpuReshape(outdim)(x, dims)
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3341,6 +3341,14 @@ class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
 def gpu_flatten(x, outdim=1):
     """
     Implement flatten on the gpu.
+
+    :param x: the variable that should be reshaped.
+    :type x: theano.tensor.var.TensorVariable
+
+    :param outdim: the number of dimensions of the returned variable
+    :type outdim: int
+
+    :returns: the flattend variable with dimensionality of outdim
     """
     x = as_cuda_ndarray_variable(x)
     if outdim > 1:
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3322,18 +3322,33 @@ class GpuIncSubtensor(tensor.IncSubtensor, GpuOp):
         return ()
 
 
-class GpuFlatten(gof.HideC, tensor.Flatten, GpuOp):
-    """
-    Implement Flatten on the gpu.
+#class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
+#    """
+#    Implement Flatten on the gpu.
+#
+#    """
+#
+#    def make_node(self, x):
+#        warnings.warn(
+#            "GpuFlatten class is deprecated, "
+#            "please use gpu_flatten method instead.",
+#            DeprecationWarning,
+#            stacklevel=4)
+#        assert isinstance(x.type, CudaNdarrayType)
+#        rval = tensor.Reshape.make_node(self, x, [tensor.prod(x.shape)])
+#        host_out_broadcastable = rval.outputs[0].type.broadcastable
+#        out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
+#        return Apply(self, [x], [out_type()])
 
-    """
 
-    def make_node(self, x):
-        assert isinstance(x.type, CudaNdarrayType)
-        rval = tensor.Flatten.make_node(self, x)
-        host_out_broadcastable = rval.outputs[0].type.broadcastable
-        out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
-        return Apply(self, [x], [out_type()])
+
+def gpu_flatten(x, outdim=1):
+    x = as_cuda_ndarray_variable(x)
+    if outdim > 1:
+        dims = tuple(x.shape[:outdim-1])+(theano.tensor.prod(x.shape[outdim-1:]),)
+    else:
+        dims = (-1,)
+    return  GpuReshape(outdim)(x, dims)
 
 
 class GpuShape(tensor.Shape, GpuOp):
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3322,23 +3322,23 @@ class GpuIncSubtensor(tensor.IncSubtensor, GpuOp):
         return ()
 
 
-#class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
-#    """
-#    Implement Flatten on the gpu.
-#
-#    """
-#
-#    def make_node(self, x):
-#        warnings.warn(
-#            "GpuFlatten class is deprecated, "
-#            "please use gpu_flatten method instead.",
-#            DeprecationWarning,
-#            stacklevel=4)
-#        assert isinstance(x.type, CudaNdarrayType)
-#        rval = tensor.Reshape.make_node(self, x, [tensor.prod(x.shape)])
-#        host_out_broadcastable = rval.outputs[0].type.broadcastable
-#        out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
-#        return Apply(self, [x], [out_type()])
+class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
+    """
+    Implement Flatten on the gpu.
+
+    """
+
+    def make_node(self, x):
+        warnings.warn(
+            "GpuFlatten class is deprecated, "
+            "please use gpu_flatten method instead.",
+            DeprecationWarning,
+            stacklevel=4)
+        assert isinstance(x.type, CudaNdarrayType)
+        rval = tensor.Reshape.make_node(self, x, [tensor.prod(x.shape)])
+        host_out_broadcastable = rval.outputs[0].type.broadcastable
+        out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
+        return Apply(self, [x], [out_type()])
 
 
 
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3322,16 +3322,22 @@ class GpuIncSubtensor(tensor.IncSubtensor, GpuOp):
         return ()
 
 
-class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
+class GpuFlatten(gof.HideC, tensor.Flatten, GpuOp):
     """
     Implement Flatten on the gpu.
 
-    Note: The interface GpuFlatten is deprecated, you should use gpu_flatten
+    .. note:: The interface GpuFlatten is deprecated, you should use gpu_flatten.
     """
+    def __init__(self):
+        warnings.warn(
+            "GpuFlatten class is deprecated, "
+            "please use gpu_flatten method instead.",
+            DeprecationWarning,
+            stacklevel=4)
 
     def make_node(self, x):
         assert isinstance(x.type, CudaNdarrayType)
-        rval = flatten(x)
+        rval = tensor.Flatten.make_node(self, x)
         host_out_broadcastable = rval.outputs[0].type.broadcastable
         out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
         return Apply(self, [x], [out_type()])
@@ -3341,14 +3347,23 @@ class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
 def gpu_flatten(x, outdim=1):
     """
     Implement flatten on the gpu.
+    Reshapes the variable x by keeping
+    the first outdim-1 dimension size(s) of x the same,
+    and making the last dimension size of x equal to
+    the multiplication of its remaining dimension size(s).
 
-    :param x: the variable that should be reshaped.
-    :type x: theano.tensor.var.TensorVariable
+    Parameters
+    ----------
+        x : theano.tensor.var.TensorVariable
+            the variable that should be reshaped.
 
-    :param outdim: the number of dimensions of the returned variable
-    :type outdim: int
+        outdim : int
+            the number of dimensions of the returned variable
 
-    :returns: the flattend variable with dimensionality of outdim
+    Returns
+    -------
+    theano.tensor.var.TensorVariable
+        the flattend variable with dimensionality of outdim
     """
     x = as_cuda_ndarray_variable(x)
     if outdim > 1:
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -24,7 +24,9 @@ from theano.sandbox.cuda.basic_ops import (
     gpu_eye, gpu_contiguous,
     gpu_from_host, host_from_gpu, GpuFromHost, HostFromGpu,
     GpuContiguous,
-    GpuElemwise, GpuDimShuffle, GpuReshape, GpuCAReduce, GpuFlatten,
+    GpuElemwise, GpuDimShuffle, GpuReshape, GpuCAReduce,
+    # GpuFlatten,
+    gpu_flatten, 
     GpuSubtensor, GpuAdvancedSubtensor1,
     GpuAdvancedIncSubtensor1, GpuAdvancedIncSubtensor1_dev20,
     GpuIncSubtensor, gpu_alloc, GpuAlloc, gpu_shape, GpuSplit, GpuAllocEmpty)
@@ -152,7 +154,7 @@ cpu_ops_moved_to_gpu = [
     tensor.elemwise.All, tensor.elemwise.Any,
     tensor.elemwise.CAReduceDtype, tensor.elemwise.Sum,
     tensor.elemwise.Prod, tensor.elemwise.ProdWithoutZeros,
-    tensor.Reshape, tensor.Flatten, tensor.Subtensor,
+    tensor.Reshape, tensor.flatten, tensor.Subtensor,
     tensor.AdvancedSubtensor1, tensor.AdvancedIncSubtensor1,
     tensor.IncSubtensor, tensor.Shape, tensor.Join,
     tensor.Alloc, tensor.Eye]
@@ -972,23 +974,23 @@ def local_gpu_reshape(node):
     return False
 
 
-@register_opt()
-@local_optimizer([gpu_from_host, tensor.Flatten])
-def local_gpu_flatten(node):
-    if isinstance(node.op, GpuFromHost):
-        host_input = node.inputs[0]
-        if host_input.owner and \
-           isinstance(host_input.owner.op, tensor.Flatten):
-            outdim = host_input.owner.op.outdim
-            return [GpuFlatten(outdim)(
-                as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
-    if isinstance(node.op, tensor.Flatten):
-        x, = node.inputs
-        outdim = node.op.outdim
-        if x.owner and isinstance(x.owner.op, HostFromGpu):
-            gpu_x, = x.owner.inputs
-            return [host_from_gpu(GpuFlatten(outdim)(gpu_x))]
-    return False
+#@register_opt()
+#@local_optimizer([gpu_from_host, tensor.Reshape])
+#def local_gpu_flatten(node):
+#    if isinstance(node.op, GpuFromHost):
+#        host_input = node.inputs[0]
+#        if host_input.owner and \
+#           isinstance(host_input.owner.op, tensor.Reshape):
+#            outdim = host_input.owner.op.outdim
+#            return [GpuFlatten(outdim)(
+#                as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
+#    if isinstance(node.op, tensor.Reshape):
+#        x, shp= node.inputs
+#        outdim = node.op.outdim
+#        if x.owner and isinstance(x.owner.op, HostFromGpu):
+#            gpu_x, = x.owner.inputs
+#            return [host_from_gpu(GpuFlatten(outdim)(gpu_x))]
+#    return False
 
 
 @register_opt()
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -25,8 +25,7 @@ from theano.sandbox.cuda.basic_ops import (
     gpu_from_host, host_from_gpu, GpuFromHost, HostFromGpu,
     GpuContiguous,
     GpuElemwise, GpuDimShuffle, GpuReshape, GpuCAReduce,
-    # GpuFlatten,
-    gpu_flatten, 
+    GpuFlatten, gpu_flatten,
     GpuSubtensor, GpuAdvancedSubtensor1,
     GpuAdvancedIncSubtensor1, GpuAdvancedIncSubtensor1_dev20,
     GpuIncSubtensor, gpu_alloc, GpuAlloc, gpu_shape, GpuSplit, GpuAllocEmpty)
@@ -974,23 +973,23 @@ def local_gpu_reshape(node):
     return False
 
 
-#@register_opt()
-#@local_optimizer([gpu_from_host, tensor.Reshape])
-#def local_gpu_flatten(node):
-#    if isinstance(node.op, GpuFromHost):
-#        host_input = node.inputs[0]
-#        if host_input.owner and \
-#           isinstance(host_input.owner.op, tensor.Reshape):
-#            outdim = host_input.owner.op.outdim
-#            return [GpuFlatten(outdim)(
-#                as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
-#    if isinstance(node.op, tensor.Reshape):
-#        x, shp= node.inputs
-#        outdim = node.op.outdim
-#        if x.owner and isinstance(x.owner.op, HostFromGpu):
-#            gpu_x, = x.owner.inputs
-#            return [host_from_gpu(GpuFlatten(outdim)(gpu_x))]
-#    return False
+@register_opt()
+@local_optimizer([gpu_from_host, tensor.Flatten])
+def local_gpu_flatten(node):
+    if isinstance(node.op, GpuFromHost):
+        host_input = node.inputs[0]
+        if host_input.owner and \
+           isinstance(host_input.owner.op, tensor.Flatten):
+            outdim = host_input.owner.op.outdim
+            return [gpu_flatten(outdim)(
+                as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
+    if isinstance(node.op, tensor.Flatten):
+        x, shp= node.inputs
+        outdim = node.op.outdim
+        if x.owner and isinstance(x.owner.op, HostFromGpu):
+            gpu_x, = x.owner.inputs
+            return [host_from_gpu(gpu_flatten(outdim)(gpu_x))]
+    return False
 
 
 @register_opt()
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -981,14 +981,14 @@ def local_gpu_flatten(node):
         if host_input.owner and \
            isinstance(host_input.owner.op, tensor.Flatten):
             outdim = host_input.owner.op.outdim
-            return [gpu_flatten(outdim)(
+            return [gpu_flatten(host_input.owner.inputs[0], outdim)(
                 as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
     if isinstance(node.op, tensor.Flatten):
         x, shp= node.inputs
         outdim = node.op.outdim
         if x.owner and isinstance(x.owner.op, HostFromGpu):
             gpu_x, = x.owner.inputs
-            return [host_from_gpu(gpu_flatten(outdim)(gpu_x))]
+            return [host_from_gpu(gpu_flatten(host_input.owner.inputs[0], outdim)(gpu_x))]
     return False
 
 
--- a/theano/sandbox/cuda/tests/test_basic_ops.py
+++ b/theano/sandbox/cuda/tests/test_basic_ops.py
@@ -308,6 +308,7 @@ def test_flatten():
     f = theano.function([x], x.flatten(), mode=mode_with_gpu)
     assert any([node for node in f.maker.fgraph.toposort()
                 if isinstance(node.op, B.GpuReshape)])
+    assert theano.tensor.is_flat(x.flatten())
     assert len(f([[0., 0.], [0., 0.]]).shape) == 1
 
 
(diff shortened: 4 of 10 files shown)


theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,35 @@ def c_code_cache_version(self):
...
 
     def make_node(self, x):
+        warnings.warn(
+            "GpuFlatten class is deprecated, "
+            "please use gpu_flatten method instead.",
+            DeprecationWarning,
+            stacklevel=4)
         assert isinstance(x.type, CudaNdarrayType)
-        rval = tensor.Flatten.make_node(self, x)
+        rval = tensor.Reshape.make_node(self, x, [tensor.prod(x.shape)])
```
bouthilx: No, ```GpuFlatten``` is deprecated as ```Flatten``` is. You should modify ```__init__.py``` to add a deprecation warning like in ```Flatten```.

bouthilx: Oh I stepped above the warning, looking at the bold ```Reshape```. You should move this warning to __init__.py. Also, the ```broadcastable``` attribute is not correct with the ```Reshape``` substitution, you would need to use ```flatten()```

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,35 @@ def c_code_cache_version(self):
...
         host_out_broadcastable = rval.outputs[0].type.broadcastable
         out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
         return Apply(self, [x], [out_type()])
 
 
+
+def gpu_flatten(x, outdim=1):
```
bouthilx: docstring :)

theano/sandbox/cuda/opt.py
```diff
@@ -952,23 +954,23 @@ def local_gpu_reshape(node):
...
-            return [GpuFlatten(outdim)(
-                as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
-    if isinstance(node.op, tensor.Flatten):
-        x, = node.inputs
-        outdim = node.op.outdim
-        if x.owner and isinstance(x.owner.op, HostFromGpu):
-            gpu_x, = x.owner.inputs
-            return [host_from_gpu(GpuFlatten(outdim)(gpu_x))]
-    return False
+#@register_opt()
```
bouthilx: Add a comment to explain why it is commented out and in which condition it should be brought back.

theano/tensor/basic.py
```diff
@@ -4447,6 +4447,11 @@ class Flatten(Op):
     __props__ = ("outdim",)
 
     def __init__(self, outdim=1):
+        warnings.warn(
```
bouthilx: You should add a deprecation warning in the class doctring.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,35 @@ def c_code_cache_version(self):
         return ()
 
 
-class GpuFlatten(gof.HideC, tensor.Flatten, GpuOp):
+class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
     """
     Implement Flatten on the gpu.
```
bouthilx: Deprecation

theano/tensor/basic.py
```diff
@@ -4584,9 +4589,34 @@ def c_code(self, node, name, inputs, outputs, sub):
         }
         """ % locals()
 
+def is_flatten(node, outdim=1):
```
bouthilx: docstring :)

theano/tensor/basic.py
```diff
@@ -4584,9 +4589,34 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    return isinstance(node.op, theano.tensor.Reshape) and\
+        node.inputs[1].ndim == outdim
+
 
 def flatten(x, outdim=1):
```
bouthilx: docstring :)

theano/tensor/basic.py
```diff
@@ -4584,9 +4589,34 @@ def c_code(self, node, name, inputs, outputs, sub):
...
-    return Flatten(outdim)(x)
+    outdim = int(outdim)
+    # an error is raised if outdim is not positive or
+    # of outdim is greater than one and also greater than x dimensionality.
+    # Any input variable can be flattened to have outdim of 1,
```
bouthilx: Great! I'd remove the first two lines, I find them a little cryptic, and I'd add something related after your second sentence, giving roughly
> Any input variable can be flattened to have `outdim` of 1, even if it's a scalar. Otherwise, `outdim` must be positive and smaller than `x.ndim`.

theano/tensor/basic.py
```diff
@@ -4584,9 +4589,34 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+        dims = tuple(x.shape[:outdim-1]) +\
+            (theano.tensor.prod(x.shape[outdim-1:]),)
+    else:
+        dims = (-1,)
+    x_reshaped = x.reshape(dims)
+    bcast_kept_dims = x.broadcastable[:outdim - 1]
+    bcast_new_dim = python_all(x.broadcastable[outdim - 1:])
+    broadcastable = bcast_kept_dims + (bcast_new_dim,)
+    broadcast_int = tuple([numpy.int(bc) for bc in broadcastable])
+    for dim, br in enumerate(broadcast_int):
```
bouthilx: Why do you need to cast broadcastable[i] to integers?

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -377,7 +377,7 @@ def test_log1msigm_to_softplus(self):
         f = theano.function([x], out, mode=self.m)
         topo = f.maker.fgraph.toposort()
         assert len(topo) == 3
-        assert isinstance(topo[0].op, T.Flatten)
+        tensor.is_flatten(topo[0])
```
bouthilx: assert

theano/tensor/opt.py
```diff
@@ -3597,13 +3597,13 @@ def local_useless_split(node):
 @register_stabilize
 @gof.local_optimizer([T.Flatten])
 def local_flatten_lift(node):
-    """
+    ""
```
bouthilx: ?

theano/tensor/opt.py
```diff
@@ -3597,13 +3597,13 @@ def local_useless_split(node):
...
     Flatten(UnaryElemwise(x)) -> UnaryElemwise(Flatten(x))
 
     This optimization is needed by optimization
     nnet/sigm.py:log1msigm_to_softplus to get applied when there is a flatten.
 
-    """
+    ""
```
bouthilx: ?

theano/tensor/tests/test_basic.py
```diff
@@ -5174,7 +5169,7 @@ def test_flatten_outdim2():
     f = inplace_func([a], c)
     assert numpy.all(f(a_val) == a_val)
 
-    utt.verify_grad(Flatten(2), [a_val])
+    utt.verify_grad(flatten, [a_val])
```
bouthilx: Oh, but ```Flatten(2)``` and ```flatten``` are not equivalent right? It will call ```flatten``` with default ```outdim=1``` while we wanted ```outdim=2```. You should use [functools.partial](https://docs.python.org/2/library/functools.html#functools.partial) to preset ```outdim=2```.

theano/tensor/tests/test_basic.py
```diff
@@ -5188,7 +5183,7 @@ def test_flatten_outdim2_of_3():
     f = inplace_func([a], c)
     assert numpy.all(f(a_val) == c_val)
 
-    utt.verify_grad(Flatten(2), [a_val])
+    utt.verify_grad(flatten, [a_val])
```
bouthilx: idem

theano/tensor/tests/test_basic.py
```diff
@@ -7000,24 +6995,27 @@ def test_infer_shape(self):
...
-        avec = vector()
-        avec_val = rand(4)
-        outdim = 1
-        self._compile_and_check([avec],
-                                [Flatten(outdim)(avec)],
-                                [avec_val], Flatten)
+        self._compile_and_check([atens3],
+                                [flatten(atens3, 1)],
+                                [atens3_val], Reshape)
+        #for outdim in (3, 2, 1):
```
bouthilx: Add a comment to explain why it is commented out and in which condition it should be brought back.

theano/tensor/tests/test_opt.py
```diff
@@ -60,6 +60,7 @@
 from theano.compile.mode import optdb
 from theano.compile import Mode
 from nose.plugins.attrib import attr
+from theano.tensor.basic import flatten, is_flatten
```
bouthilx: Why do you import ```flatten``` and ```is_flatten``` if you latter use ```tensor.{flatten, is_flatten}```?

theano/tensor/tests/test_opt.py
```diff
@@ -5691,18 +5692,19 @@ def test_local_useless_split():
...
-        f(numpy.random.rand(5, 4, 3, 2).astype(config.floatX))
+        x_np = numpy.random.rand(5, 4, 3, 2).astype(config.floatX)
+        out_np = f(x_np)
         topo = f.maker.fgraph.toposort()
-        assert len(topo) == 2
-        assert isinstance(topo[0].op, tensor.Flatten)
-        assert isinstance(topo[1].op, tensor.Elemwise)
+        shape_out_np = tuple(x_np.shape[:i-1])+(numpy.prod(x_np.shape[i-1:]),)
+        assert shape_out_np == out_np.shape
+        tensor.is_flatten(topo[0], outdim=i)
```
bouthilx: assert

bouthilx:
You should add a test for deprecation similar to [this].(https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_basic.py#L3486)

Test that ```flatten``` is not logging any warning (could be an ```op``` from ```Flatten```) and test that ```Flatten(i)(x)``` is logging warnings.

Also, is there any test for ```flatten()``` good/bad arguments? Same for ```gpu_flatten()```?

Add tests for ```is_flatten()```.

bouthilx:
All tests failed because of wrong string delimiters in ```theano/tensor/opt.py```, line 3600.

theano/sandbox/cuda/basic_ops.py
SinaHonari: Do you mean 
rval = tensor.Reshape.make_node(self, x, [tensor.prod(x.shape)])
should be replaced with
rval = flatten(x)
?

theano/sandbox/cuda/opt.py
SinaHonari: Actually I had forgotten to uncomment it. I will uncomment it and update the PR.

theano/tensor/basic.py
SinaHonari: Do you mean in addition to what's in here?

SinaHonari: And where is it? I couldn't find it.

theano/sandbox/cuda/basic_ops.py
SinaHonari: Apart from __init__.py where the deprecation message should go exactly?

theano/sandbox/cuda/basic_ops.py
bouthilx: exactly

theano/sandbox/cuda/basic_ops.py
bouthilx: There should be a note in the docstring. Like [here](https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L4047).

theano/sandbox/cuda/opt.py
bouthilx: Having a look at it, the optimization will fail. You can't replace Reshape by GpuFlatten in every cases. Leave Flatten there, theano already have an optimization for Reshape -> GpuReshape.

theano/tensor/basic.py
bouthilx: Yes, in addition. See comment for GpuFlatten.

The docstring is above at line 4437.

```python
class Flatten(Op):
     """
     Flatten a tensor.
 
     Flattens a tensor to `outdim` dimensions by preserving the leading
     outdim - 1 shape components.
 
     """
```

theano/tensor/tests/test_basic.py
SinaHonari: Do you mean replacing this line with:

    from functools import partial
    flatten_2 = partial(flatten, outdim=2)
    utt.verify_grad(flatten_2, [a_val])
?

bouthilx: Yep, you can pass it directly.

theano/tensor/tests/test_basic.py
SinaHonari: I don't remember the condition, it was you and Arnaud taking about a bug :)

theano/sandbox/cuda/opt.py
SinaHonari: OK, I commented it again, but in which case it should be brought back?

SinaHonari:
is_flatten() is tested in tensor/tests/test_opt.py in
test_local_flatten_lift method, which checks for 1 to 4 flatten dimensionality.

theano/tensor/tests/test_basic.py
bouthilx: It is related to issue #3420. ```Reshape.infershape``` is badly written and because of this all ```Reshapes``` are not removed from ```atens3.shape``` after compiling.

theano/sandbox/cuda/opt.py
bouthilx: Oh no, sorry if I wasn't clear. It should not be commented out, but it should transfer ```Flatten``` ops, not ```Reshape```. This is necessary to transfer ```Flatten``` in old pickled theano functions to GPU whenever possible. Optimize ```Flatten```, but use ```gpu_flatten```, not ```GpuFlatten```.

bouthilx:
I realize I made a bad name choice with ```is_flatten()```. It should be ```is_flat()```.

I'm not very comfortable with relying on ```test_local_flatten_lift()``` to assert ```is_flat()```'s validity. I believe ```flatten()``` and ```is_flat()``` are correct, but later on they could get both broken in such a way that the test still pass. 

I know ```is_flat()``` is very simple, but just to stay on the safe side, if ```test_is_flat()``` pass, then we know ```test_local_flatten_lift()``` can be trusted.

theano/tensor/basic.py
```diff
@@ -4584,9 +4589,34 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    # an error is raised if outdim is not positive or
+    # of outdim is greater than one and also greater than x dimensionality.
+    # Any input variable can be flattened to have outdim of 1,
+    # even if it's a scalar.
+    if outdim < 1 or (outdim > 1 and outdim > x.ndim):
+        raise ValueError('outdim %s out of bound [1, %d)' % (outdim, x.ndim+1))
+
+    if outdim > 1:
+        dims = tuple(x.shape[:outdim-1]) +\
+            (theano.tensor.prod(x.shape[outdim-1:]),)
```
bouthilx: Can we replace prod with -1? It would simplify graph complexity.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,35 @@ def c_code_cache_version(self):
...
         host_out_broadcastable = rval.outputs[0].type.broadcastable
         out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
         return Apply(self, [x], [out_type()])
 
 
+
+def gpu_flatten(x, outdim=1):
+    x = as_cuda_ndarray_variable(x)
+    if outdim > 1:
+        dims = tuple(x.shape[:outdim-1])+(theano.tensor.prod(x.shape[outdim-1:]),)
```
bouthilx: Same thing here for -1 index?

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -377,7 +377,7 @@ def test_log1msigm_to_softplus(self):
         f = theano.function([x], out, mode=self.m)
         topo = f.maker.fgraph.toposort()
         assert len(topo) == 3
-        assert isinstance(topo[0].op, T.Flatten)
+        assert tensor.is_flat(topo[0])
```
bouthilx: ```topo[0].outputs[0]``` maybe?

theano/tensor/tests/test_basic.py
```diff
@@ -5174,7 +5169,9 @@ def test_flatten_outdim2():
     f = inplace_func([a], c)
     assert numpy.all(f(a_val) == a_val)
 
-    utt.verify_grad(Flatten(2), [a_val])
+    from functools import partial
```
bouthilx: do not import it here. Import it at the top of the file.

theano/tensor/tests/test_basic.py
```diff
@@ -5188,7 +5185,9 @@ def test_flatten_outdim2_of_3():
     f = inplace_func([a], c)
     assert numpy.all(f(a_val) == c_val)
 
-    utt.verify_grad(Flatten(2), [a_val])
+    from functools import partial
```
bouthilx: idem

theano/sandbox/cuda/__init__.py
```diff
@@ -316,11 +317,16 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
...
+            GpuFlatten, gpu_flatten, GpuShape, GpuAlloc, GpuSplit,
             GpuJoin, fscalar, fvector, fmatrix, frow, fcol,
             ftensor3, ftensor4,
             scalar, vector, matrix, row, col,
             tensor3, tensor4)
+    warnings.warn(
+        "GpuFlatten class is deprecated, "
+        "please use gpu_flatten method instead.",
+        DeprecationWarning,
+        stacklevel=4)
```
bouthilx: Oh! You should not put the message in ```__init__.py``` but in GpuFlatten.__init__().

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,42 @@ def c_code_cache_version(self):
         return ()
 
 
-class GpuFlatten(gof.HideC, tensor.Flatten, GpuOp):
+class GpuFlatten(gof.HideC, tensor.Reshape, GpuOp):
```
bouthilx: Leave ```Flatten``` here. We want ```GpuFlatten``` and ```Flatten``` to stay bounded together. Actually, I would leave ```Flatten.make_node``` ```GpuFlatten.make_node``` just to make it clear. Sorry for changing my mind so often! ^_^

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,42 @@ def c_code_cache_version(self):
...
     """
     Implement Flatten on the gpu.
 
+    Note: The interface GpuFlatten is deprecated, you should use gpu_flatten
```
bouthilx: .. note:: The interface GpuFlatten is deprecated, you should use gpu_flatten.

This will render more nicely on the website. :)

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,42 @@ def c_code_cache_version(self):
...
+    """
+    Implement flatten on the gpu.
+
+    :param x: the variable that should be reshaped.
+    :type x: theano.tensor.var.TensorVariable
+
+    :param outdim: the number of dimensions of the returned variable
+    :type outdim: int
+
+    :returns: the flattend variable with dimensionality of outdim
```
bouthilx: This is the old docstring format. Please follow this [guide](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3329,20 +3329,42 @@ def c_code_cache_version(self):
...
+        rval = flatten(x)
         host_out_broadcastable = rval.outputs[0].type.broadcastable
         out_type = CudaNdarrayType(broadcastable=host_out_broadcastable)
         return Apply(self, [x], [out_type()])
 
 
+
+def gpu_flatten(x, outdim=1):
+    """
+    Implement flatten on the gpu.
```
bouthilx: It would be better to also explain what flatten is so the user doesn't have to look around for flatten to understand gpu_flatten.

theano/sandbox/cuda/opt.py
```diff
@@ -960,14 +961,14 @@ def local_gpu_flatten(node):
         if host_input.owner and \
            isinstance(host_input.owner.op, tensor.Flatten):
             outdim = host_input.owner.op.outdim
-            return [GpuFlatten(outdim)(
+            return [gpu_flatten(outdim)(
```
bouthilx: Wrong use of ```gpu_flatten```. Should be ```gpu_flatten(host_input.owner.inputs[0], outdim)```.

theano/sandbox/cuda/opt.py
```diff
@@ -960,14 +961,14 @@ def local_gpu_flatten(node):
...
                 as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
     if isinstance(node.op, tensor.Flatten):
-        x, = node.inputs
+        x, shp= node.inputs
         outdim = node.op.outdim
         if x.owner and isinstance(x.owner.op, HostFromGpu):
             gpu_x, = x.owner.inputs
-            return [host_from_gpu(GpuFlatten(outdim)(gpu_x))]
+            return [host_from_gpu(gpu_flatten(outdim)(gpu_x))]
```
bouthilx: idem

theano/sandbox/cuda/tests/test_basic_ops.py
```diff
@@ -307,7 +307,7 @@ def test_flatten():
     x = cuda.fmatrix('x')
     f = theano.function([x], x.flatten(), mode=mode_with_gpu)
     assert any([node for node in f.maker.fgraph.toposort()
-                if isinstance(node.op, B.GpuFlatten)])
+                if isinstance(node.op, B.GpuReshape)])
```
bouthilx: Would be nice to add a ```is_flat()``` test.

theano/tensor/basic.py
```diff
@@ -4440,6 +4440,7 @@ class Flatten(Op):
     Flattens a tensor to `outdim` dimensions by preserving the leading
     outdim - 1 shape components.
 
+    Note: The interface Flatten(Op) is deprecated, you should use flatten
```
bouthilx: .. note:: The interface Flatten is deprecated, you should use flatten

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
         """ % locals()
 
 
+def is_flat(node, outdim=1):
+    """
+    verifies the dimensionality of variable is correct.
```
bouthilx: Please explain here what is a flat theano variable. (like in ```flatten()```)

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+
+    :param node: the theano node on which the dimensionality is checked.
+    :type node: theano.tensor.var.TensorVariable
+
+    :param outdim: the expected dimensionality of node.
+    :type outdim: int
+
+    :returns: the comparison result of node's dim
+              and the expected outdim.
+    """
```
bouthilx: idem here to docstring format.

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    return node.ndim == outdim
+
+
 def flatten(x, outdim=1):
-    return Flatten(outdim)(x)
+    """
+    Reshapes the variable x by keeping
+    the first outdim-1 dimension(s) of x the same,
```
bouthilx: dimension(s) size

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    and making the last dimension of x equal to
```
bouthilx: dimension size

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    the multiplication of its remaining dimensions.
```
bouthilx: dimension sizes

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+
+    :param x: the theano variable that should be reshaped.
+    :type x: theano.tensor.var.TensorVariable
+
+    :param outdim: the number of dimensions of the returned variable
+    :type outdim: int
+
+    :returns: the flattend variable with dimensionality of outdim
```
bouthilx: idem for docstring format

theano/tensor/basic.py
```diff
@@ -4585,8 +4591,57 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    """
+    outdim = int(outdim)
```
bouthilx: Can you please remember me why we have to cast it to int?

SinaHonari: I added it to make sure that the outdim is a valid value, but if you think it's unnecessary I can remove it.

theano/sandbox/cuda/__init__.py
SinaHonari: GpuFlatten does not have __init__
so I added one there.

nouiz:
I didn't review, just to note that travis fail.

bouthilx:
Huh. :confused: I ran it locally and the test pass. I relaunched the two failling jobs just in case.

bouthilx:
@nouiz The optimization [```local_flatten_lift```](https://github.com/Theano/Theano/blob/master/theano/tensor/opt.py#L3652) is failing because there is no Flatten anymore. We could use ```is_flat```to detect flattened inputs, but we don't know about ```outdim```. In other words it can only work for ```outdim==1``` so far. I don't know if there is a way to store ```outdim``` information in the node, but even if there is that seams to be a very bad solution.

theano/tensor/basic.py
```diff
@@ -4372,7 +4372,7 @@ def infer_shape(self, node, ishapes):
                     if ele == -1:
                         requ[i] = missing
             elif crit == 1:  # we reshape to -1
-                requ = [mul(*ishapes[0])]
+                requ = [mul(*ishapes[0])] if ishapes[0] else [1]
```
nouiz: I think the else clause isn't good. Why this is needed?

theano/tensor/basic.py
```diff
@@ -4606,8 +4612,71 @@ def c_code(self, node, name, inputs, outputs, sub):
         """ % locals()
 
 
+def is_flat(node, outdim=1):
```
nouiz: rename node. node is more frequently used for Apply node. Use "var" or "x".

theano/tensor/basic.py
```diff
@@ -4606,8 +4612,71 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+        dims = tuple(x.shape[:outdim - 1]) + (-1,)
+    else:
+        dims = (-1,)
+    x_reshaped = x.reshape(dims)
+    bcast_kept_dims = x.broadcastable[:outdim - 1]
+    bcast_new_dim = python_all(x.broadcastable[outdim - 1:])
+    broadcastable = bcast_kept_dims + (bcast_new_dim,)
+    for dim, br in enumerate(broadcastable):
+        if br:
+            x_reshaped = theano.tensor.addbroadcast(x_reshaped, dim)
```
nouiz: if possible, try to make only one call to addbroadcast. This would make the graph smaller and more readable.

theano/tensor/tests/test_opt.py
```diff
@@ -5675,18 +5675,19 @@ def test_local_useless_split():
 
 def test_local_flatten_lift():
     for i in xrange(1, 4):
-        op = tensor.Flatten(i)
         x = tensor.tensor4()
-        out = op(T.exp(x))
+        out = tensor.flatten(T.exp(x), i)
         assert out.ndim == i
         mode = compile.mode.get_default_mode()
         mode = mode.including('local_flatten_lift')
```
nouiz: use this mode to fix the travis problem:
mode=mode.including('local_reshape_lift')

nouiz:
There is an local_reshape_lift optimization that should do the work. I did
a review of the PR and made a comment where this can be used to fix travis

On Mon, Nov 2, 2015 at 9:50 PM, Xavier Bouthillier <notifications@github.com

theano/tensor/tests/test_opt.py
bouthilx: Isn't test_local_reshape_lift already doing the job then? test_local_flatten_lift could be removed?

theano/tensor/basic.py
bouthilx: mul() fails when there is no argument, i.e when ishapes[0] is empty. Then we give it a scalar dimension of 1.

theano/tensor/basic.py
bouthilx: Something like
```python
x_reshaped = theano.tensor.addbroadcast(
    x_reshaped, filter(lambda i: broadcastable[i], range(outdim))
```

theano/tensor/tests/test_opt.py
nouiz: I would keep it if we reload pickled graph. But add a comment to it why we keep it and that it will be removed in the future. We should remove it at the same time as the Flatten class, so that old pickled files will just crash instead of running non optimized (or we should make an optimization to convert it to Reshape and remove that optimization?, I would go for the faster to implement for now)

theano/tensor/basic.py
nouiz: Ok. We could change the interface, to make it return a constant 1 of floatX dtype, but I'm not sure it is a good idea, so I won't push it. But if you think it is a good idea, tell it.

theano/tensor/tests/test_opt.py
bouthilx: I see no point to add an optimization that is going to be deprecated like the one we aim to remove. Like you propose, I would simply add a deprecation comment in the code to make sure the optimization is removed along with Flatten at the end of its deprecation phase.

theano/tensor/basic.py
bouthilx: By changing the interface you mean ishapes's interface or Reshape.infer_shape return's interface? Reshape.infer_shape is broken anyway and is Sina's next CCW ticket.

How does infer_shape usually process scalar dimensions in ishapes and in return?

nouiz: I was talking about changing mul() interface do not crash when it get no inputs.

Can you tell more on what in the Reshape.infer_shape is broken? If it is broken, maybe we should wait for it to be fixed before converting Flatten to Reshape.

I don't understand your questions. For scalar (ndim=0), We simply have an empty tuple.

bouthilx: >I was talking about changing mul() interface do not crash when it get no inputs.

Oh ok. I agree with you not to change mul's interface. It should fail when there is no argument, just as python's ```sum``` does for instance.

> Can you tell more on what in the Reshape.infer_shape is broken? If it is broken, maybe we should wait for it to be fixed before converting Flatten to Reshape.

It's using Shape_i() which makes TestInferShape fail as there is still some Reshapes in shape's graph after optimization. See issue #3420 

> I don't understand your questions. For scalar (ndim=0), We simply have an empty tuple.

Yep sure. I think I understand now why you say the else cause isn't good. It's because it should be ```[mul(*ishapes[0])] if ishapes[0] else []```, isn't? Otherwise we are turning a scalar into a 1 dimensional vector.

bouthilx: We can wait for #3420 to be fixed before merge. There is a test that Sina commented out because of this bug. We could put it back in before merge if #3420 is fixed before. Otherwise, everything seams to be done for the conversion of Flatten to Reshape.

nouiz: If the output is a scalar, then yes, it should return [[]].

If this PR need some tests to be commented and that gh-3420 would fix that, we should wait for gh-3420 before merging this PR.

bouthilx: OK. We'll do this. Thank you for your feedback!

theano/tensor/basic.py
SinaHonari: You correction works, however, the documentation on Theano website (http://deeplearning.net/software/theano/library/tensor/basic.html) seems wrong, since just passing a list or tuple throws this error:

list error: 
x = T.tensor4()
T.addbroadcast(x, [0,1,2])
TypeError: unhashable type: 'list'

tuple error: 
T.addbroadcast(x, (0,1,2))
AssertionError: ('Rebroadcast needs integer axes. Got ', (0, 1, 2))

The star(*) that unpacks list or tuple should also be passed to make it work

theano/tensor/tests/test_opt.py
SinaHonari: A comment should be written or a warning?

theano/tensor/basic.py
SinaHonari: So I changed it to this:
requ = [mul(*ishapes[0])] if ishapes[0] else []

theano/tensor/basic.py
SinaHonari: I changed it to \"var\"

theano/tensor/basic.py
SinaHonari: I keep the test commented for the moment, before putting it back.

theano/tensor/basic.py
nouiz: I checked the doc and it is file. *args mean that it take a variable number
of arguments, not that it take a list/tuple.

bouthilx: Yes this is why I added ```*``` before ```flatten```. You need to unpack the list/tuple.

theano/tensor/tests/test_opt.py
bouthilx: Nah, I don't think so. It's just that the FAST_COMPILE mode doesn't run the optimization unless we force it into config.mode. You can see that it is added in [test_local_reshape_lift](https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_opt.py#L5733-L5735)

theano/tensor/basic.py
bouthilx: I'd say leave it as is here, and fix it in #3420. You can rebase afterward.

theano/tensor/tests/test_opt.py
SinaHonari: I don't understand what you're talking about. Do you mean to change it to this?

        mode = compile.mode.get_default_mode()
        mode = mode.including('local_reshape_lift')
        mode = mode.including('FAST_COMPILE')

bouthilx: No. I'm explaining why travis fails on this test with mode=FAST_COMPILE. By doing ```mode.including('local_reshape_lift')``` you force the optimization in even if mode=FAST_COMPILE, so the test pass. It should be just as here [test_local_reshape_lift](https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_opt.py#L5733-L5735).

lamblin:
I'll close and reopen the pull request to see if the tests pass, now that #3420 has been closed.

bouthilx:
The [faulthy test](https://github.com/Theano/Theano/pull/3480/files#diff-bd238af9c7f5e76b67885294b7da00aaR7164) is still commented out.

lamblin:
@bouthilx: Thanks for the pointer.
@SinaHonari did tell me IRL that he would maybe need to uncomment the tests, so I stopped Travis.

SinaHonari:
I uncommented the test and updated the PR. It throws this error though:

  File "/data/lisatmp/honaris/pythonlibs/Theano/theano/tensor/tests/test_basic.py", line 7181, in test_infer_shape
    [avec_val], Reshape)
  File "/data/lisatmp/honaris/pythonlibs/Theano/theano/tests/unittest_tools.py", line 239, in _compile_and_check
    assert any(isinstance(t.op, cls) for t in topo_out)
AssertionError

lamblin:
So I had a quick look: `_compile_and_check` asserts that there _is_ a `Reshape` Op when you compute the complete value, and that it is not there when you only need the shape.
The thing is, here, when you reshape a vector into a vector, there is an optimization somewhere that removes that useless `Reshape` (which did not happen with the useless `Flatten` in the previous version of the test).
The solution would be to find which optimization that is (probably named something like "local_useless_reshape"), and to explicitly exclude it from that call to `_compile_and_check`, for instance:
```python
self._compile_and_check([avec],
                        [flatten(avec, outdim)],
                        [avec_val], Reshape,
                        excluding=['local_useless_reshape'])
```
```

**Comment by lamblin** [comment]
(Stopping Travis again since we know there is an issue to fix.)

**Comment by nouiz** [comment]
I confirm the opt name is local_useless_reshape

**Comment by SinaHonari** [comment]
Thanks Fred, I updated the PR.

File: theano/tensor/tests/test_opt.py [hunk 43]
```diff
@@ -5878,19 +5878,28 @@ def test_local_useless_split():
...
+        out_np = f(x_np)
         topo = f.maker.fgraph.toposort()
-        assert len(topo) == 2
-        assert isinstance(topo[0].op, tensor.Flatten)
-        assert isinstance(topo[1].op, tensor.Elemwise)
+        shape_out_np = tuple(x_np.shape[:i-1])+(numpy.prod(x_np.shape[i-1:]),)
+        assert shape_out_np == out_np.shape
+
+        reshape_nodes = filter(
+            lambda apply_node: isinstance(apply_node.op, tensor.Reshape), topo)
```
lamblin: Si, this does not work in Python 3, because `filter` does not return a list but a generator object, which has no length.
One option would be to explicitly call `list(...)` on the result, which would also work on Python 2.
Another option, which I find more "pythonic" than filter, would be to use the list comprehension syntax:
```
reshape_nodes = [n for n in topo if isinstance(n.op, tensor.Reshape)]
```

theano/tensor/tests/test_opt.py
```diff
@@ -5878,19 +5878,28 @@ def test_local_useless_split():
 
 
 def test_local_flatten_lift():
+    """
+    .. note:: The Flatten(Op) is deprecated, and this method
+        should be removed with Flatten.
```
lamblin: I don't think this is the right place for that note, since the test has been updated to use `tensor.flatten` rather than the `Flatten` Op. Is it supposed to go somewhere else?

lamblin:
Thanks, Travis passes and the code seems good.
There is still the issue of the docstring for test_local_flatten_lift: should that note be removed? You already added a similar one in the class docstring of `Flatten` and `GpuFlatten`, as requested by @bouthilx.

SinaHonari:
No problem. 
I think @bouthilx asked for that at some point, but I might be mistaken. If you think it's unnecessary, I can remove it.

lamblin:
I think it would be less confusing to remove that one, and then I'll merge.
Thanks!

SinaHonari:
Done. No problem.

PR #3481, opened 2015-10-02
Scan grad speedup bug

carriepl:


abergeron:
This fixes the problem with David's code I suppose?

carriepl:
Yes

carriepl:
I can't guarantee the test case I've included is 100% identical but it has the same symptom (IndexError in scan.grad()) and is solved by the same fix.

abergeron:
It looks ok to me.  I'm ok to merge even if it doesn't completely fix David's problem since it does fix a problem.

PR #3482, opened 2015-10-02
Multi-gpu support

abergeron:
This adds support for multi-gpu in gpuarray.

To the best of my knowledge it passes all tests and works correctly.


lamblin:
Flake8 errors in `theano/misc/check_multi_gpu.py`.
I think that is all what travis is complaining about.

nouiz:
travis fail

abergeron:
Hopefully not anymore.

theano/sandbox/gpuarray/__init__.py
```diff
@@ -53,10 +55,16 @@ def init_dev(dev):
         elif (config.init_gpu_device.startswith('cuda') or
               config.init_gpu_device.startswith('opencl')):
             init_dev(config.init_gpu_device)
+        if config.contexts != '':
+            for n, d in (c.split('->') for c in config.contexts.split(';')):
+                init_dev(d, n)
```
lamblin: call `strip()` on `d` and `n` to avoid spaces in the name of contexts or devices.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -19,27 +21,63 @@
...
+
+            # If x is the result of a transfer, try to dig through.
+            if getattr(x, 'owner', None):
+                if isinstance(x.owner.op, HostFromGpu):
+                    x = x.owner.inputs[0]
+                    continue
+                if isinstance(x.owner.op, GpuFromHost):
+                    x = x.owner.inputs[0]
+                    continue
+
```
lamblin: Dig through Gpu->Gpu transfers as well.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -309,29 +374,76 @@ def infer_shape(self, node, xshp):
...
-                                     PyArray_NDIM(%(name)s_tmp),
-                                     (size_t *)PyArray_DIMS(%(name)s_tmp),
-                                     (ssize_t *)PyArray_STRIDES(%(name)s_tmp),
-                                     pygpu_default_context(),
+        %(out)s = pygpu_fromhostdata(PyArray_DATA(%(inp)s),
+                                     get_typecode((PyObject *)PyArray_DESCR(%(inp)s)),
+                                     PyArray_NDIM(%(inp)s),
+                                     (size_t *)PyArray_DIMS(%(inp)s),
+                                     (ssize_t *)PyArray_STRIDES(%(inp)s),
+                                     %(ctx)s,
```
lamblin: Use `%(name)s_tmp` instead of `%(inp)s`, then decref it.

theano/sandbox/gpuarray/opt.py
```diff
@@ -779,15 +838,9 @@ def values_eq_approx(a, b):
...
         return
-    out = gpu_conv(gpu_from_host(img),
-                   gpu_from_host(kern))
-    # in some case the ConvOp broadcast the last 2 dimensions
-    # differently then the gpu ConvOp
-    out = tensor.patternbroadcast(
-        host_from_gpu(out),
-        node.outputs[0].broadcastable)
-    # op_lifter want the output on the GPU.
-    out = gpu_from_host(out)
```
lamblin: We should keep that, so that the gradient has the right broadcastable pattern.

theano/sandbox/gpuarray/__init__.py
```diff
@@ -52,11 +56,21 @@ def init_dev(dev):
             optdb.add_tags('gpuarray_opt', 'fast_run', 'fast_compile')
         elif (config.init_gpu_device.startswith('cuda') or
               config.init_gpu_device.startswith('opencl')):
+            if config.device != 'gpu':
```
nouiz: Why this restriction? It seem bad to me.

theano/sandbox/gpuarray/__init__.py
```diff
@@ -52,11 +56,21 @@ def init_dev(dev):
...
+            for n, d in (c.split('->') for c in config.contexts.split(';')):
+                init_dev(d.strip(), n.strip())
+            import theano.compile
+            theano.compile.shared_constructor(gpuarray_shared_constructor)
+            optdb.add_tags('gpuarray_opt', 'fast_run', 'fast_compile')
 
         from .basic_ops import (GpuAlloc, GpuContiguous, GpuEye, GpuFromHost,
                                 GpuJoin, GpuReshape, GpuSplit, HostFromGpu)
-        from .basic_ops import host_from_gpu, gpu_from_host
+        from .basic_ops import host_from_gpu, GpuFromHost
```
nouiz: why this change? Why remove gpu_from_gpu? this cause inconsistency in the user interface.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -259,57 +329,59 @@ def c_code_cache_version(self):
...
-        return [gpu_from_host(gz)]
+        return [GpuFromHost(inputs[0].type.context_name)(gz)]
 
     def R_op(self, inputs, eval_points):
         ev, = eval_points
-        if isinstance(ev, tensor.TensorType):
-            return [gpu_from_host(ev)]
-        else:
-            return [ev]
+        return self(ev)
```
nouiz: Why this? If needed, I think we should do the same in the old back-end for now to help keep them in sync.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -259,57 +329,59 @@ def c_code_cache_version(self):
...
         x, = inp
         z, = out
-        z[0] = gpuarray.array(x)
+        z[0] = gpuarray.array(x, context=ctx)
 
     def grad(self, inputs, grads):
         gz, = grads
-        return [host_from_gpu(as_gpuarray_variable(gz))]
+        return [host_from_gpu(as_gpuarray_variable(
+                gz, context_name=self.context_name))]
```
nouiz: I don't like this. But maybe optimization will clean that up? If the grad end up on a different GPU, then we will transfer it to self.context_name, and after transfer to the CPU... So there is one useless transfer.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -259,57 +329,59 @@ def c_code_cache_version(self):
...
 
     def R_op(self, inputs, eval_points):
         ev, = eval_points
-        if isinstance(ev, GpuArrayType):
-            return [host_from_gpu(ev)]
-        else:
-            return [ev]
+        return self(ev)
```
nouiz: idem, clean up the old back-end to keep them in sync.

theano/sandbox/gpuarray/__init__.py
abergeron: It's the same as https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/__init__.py#L539

theano/sandbox/gpuarray/__init__.py
abergeron: gpu_from_host can't be used directly since that op now needs to know to which context you want to transfer.

theano/sandbox/gpuarray/basic_ops.py
abergeron: There is an optimization what will cleanup useless transfers, but more importantly, if the grad ends up on a different gpu I would consider that a bug.

theano/sandbox/gpuarray/basic_ops.py
abergeron: @lamblin conviced me that this is the right way to implement R_op and that the previous way was bad and only worked ok since the jacobian is effectively the identity and we are patching up for things that are not in the expected location everywhere.

I can do a similar change in the old backend as well.

theano/sandbox/gpuarray/type.py
```diff
@@ -54,25 +119,28 @@ def filter(self, data, strict=False, allow_downcast=None):
                                 "got %d (dtype %s)." %
                                 (self, self.typecode, self.dtype,
                                  data.typecode, str(data.dtype)))
+            if self.context != data.context:
+                raise TypeError("data context does not match type context")
```
lamblin: Maybe include context names or device names in error message

theano/sandbox/gpuarray/type.py
```diff
@@ -90,8 +158,10 @@ def filter(self, data, strict=False, allow_downcast=None):
         return data
 
     def filter_variable(self, other, allow_convert=True):
```
lamblin: I think something is missing, `allow_convert` is not used at all.

theano/sandbox/gpuarray/__init__.py
nouiz: What you link assert that device is "cpu" not "gpu"

theano/sandbox/gpuarray/type.py
lamblin: Nevermind, I was fooled by GitHub's masking of unchanged lines.

theano/sandbox/gpuarray/type.py
```diff
@@ -120,7 +190,7 @@ def filter_variable(self, other, allow_convert=True):
                                  str(self.broadcastable)))
```
lamblin: (A couple of lines above) Why only check allow_convert if the broadcastable patterns are different? That could change in the future, this is why we have factored that out into `convert_variable`.

theano/sandbox/gpuarray/blas.py
```diff
@@ -116,11 +96,11 @@ class GpuGemm(BlasOp, Gemm):
     _f16_ok = True
 
     def make_node(self, C, alpha, A, B, beta):
-        alpha = as_tensor_variable(alpha)
-        beta = as_tensor_variable(beta)
-        A = as_gpuarray_variable(A)
-        B = as_gpuarray_variable(B)
-        C = as_gpuarray_variable(C)
+        ctx_name = infer_context_name(C, A, B)
+        Gemm.make_node(self, C, alpha, A, B, beta)
```
lamblin: Remove `Gemm.make_node`, and perform the checks about dtype and ndim in this function. I don't think the check about overlapping memory is needed.
Also re-add `as_tensor_variable` on `alpha` and `beta`.

theano/sandbox/gpuarray/blas.py
```diff
@@ -180,10 +159,11 @@ def c_code_cache_version(self):
 
 class GpuGer(BlasOp, Ger):
     def make_node(self, A, alpha, x, y):
+        ctx_name = infer_context_name(A, x, y)
         Ger.make_node(self, A, alpha, x, y)
```
lamblin: Same here.

theano/sandbox/gpuarray/blas.py
```diff
@@ -240,9 +220,10 @@ def c_code_cache_version(self):
 
 class GpuDot22(BlasOp, Dot22):
     def make_node(self, x, y):
+        ctx_name = infer_context_name(x, y)
         Dot22.make_node(self, x, y)
```
lamblin: Same here.

theano/sandbox/gpuarray/blas.py
```diff
@@ -180,10 +159,11 @@ def c_code_cache_version(self):
...
+        ctx_name = infer_context_name(A, x, y)
         Ger.make_node(self, A, alpha, x, y)
-        A = as_gpuarray_variable(A)
-        x = as_gpuarray_variable(x)
-        y = as_gpuarray_variable(y)
+        A = as_gpuarray_variable(A, ctx_name)
+        x = as_gpuarray_variable(x, ctx_name)
+        y = as_gpuarray_variable(y, ctx_name)
         assert A.dtype == x.dtype == y.dtype
         return Apply(self, [A, alpha, x, y], [A.type()])
```
lamblin: Should we pass the context to perform?

theano/sandbox/gpuarray/blas.py
```diff
@@ -231,7 +211,7 @@ def c_code(self, node, name, inp, out, sub):
         return code
```
lamblin: The C code does not seem to use the context either

theano/sandbox/gpuarray/blas.py
```diff
@@ -295,7 +276,12 @@ def c_code_cache_version(self):
 @local_optimizer([gpugemv_no_inplace], inplace=True)
 def local_inplace_gpuagemv(node):
     if node.op == gpugemv_no_inplace:
-        return [gpugemv_inplace(*node.inputs)]
+        inputs = list(node.inputs)
+        y = inputs[0]
+        if (y.owner and isinstance(y.owner.op, GpuAllocEmpty) and
+                len(y.clients) > 1):
+            inputs[0] = y.owner.op(*y.owner.inputs)
```
lamblin: Didn't you write  a decorator to avoid copy/pasting this code all over?

theano/sandbox/gpuarray/conv.py
```diff
@@ -202,22 +163,8 @@ def flops(self, inputs, outputs):
     def make_thunk(self, node, storage_map, compute_map, no_recycling):
         node_ = copy.copy(node)
```
lamblin: Why do we need to copy `node`?

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1002,7 +1031,7 @@ class GpuDnnPoolGrad(DnnBase):
         The input of the pooling.
     out
         The output of the pooling in the forward.
-    inp_grad
+    out_grad
```
lamblin: Right! I must have missed that last time.

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1363,27 +1394,30 @@ def local_avg_pool_dnn_grad_stride(node):
...
-        # No log-softmax before cudnn v3
-        return
+    # This looks for GpuDnnSoftmax so we know that we have cudnn.
     if (isinstance(node.op, GpuElemwise) and
             isinstance(node.op.scalar_op, Log) and
             node.inputs[0].owner and
             isinstance(node.inputs[0].owner.op, GpuDnnSoftmax) and
             len(node.inputs[0].clients) == 1):
+        # Don't move this call to version outside the condition, it
+        # needs to be here.
```
lamblin: Why?

theano/sandbox/gpuarray/type.py
abergeron: I do this for now since convert_variable repeats a lot of the checks that filter_variable does.  In order to avoid paying that cost most of the time, I only call it when I know it could be needed.

But this could be refactored to call convert_variable() when allow_convert is True and make all the checks otherwise.  I'd like to make that a ticket and not part of this PR though.

theano/sandbox/gpuarray/__init__.py
abergeron: Yes, that what I meant.  But I somehow mistyped in two spots.

theano/sandbox/gpuarray/opt.py
```diff
@@ -146,35 +166,81 @@ def apply(self, fgraph):
...
+            ctx_name = getattr(input.tag, 'context_name', None)
             try:
-                new_input = host_from_gpu(gpu_from_host(input))
+                new_input = host_from_gpu(GpuFromHost(ctx_name)(input))
                 fgraph.replace_validate(input, new_input,
                                         "InputToGpuOptimizer")
             except TypeError:
                 # This could fail if the inputs are not TensorTypes
                 pass
+            except ValueError:
```
lamblin: We should probably use a different kind of error, maybe ContextNotDefined or something like that, instead of a generic ValueError.

theano/sandbox/gpuarray/blas.py
abergeron: No, because these ops don't implement get_context() and so are not recognized as "context" nodes.

theano/sandbox/gpuarray/blas.py
abergeron: Same thing.

theano/sandbox/gpuarray/blas.py
abergeron: Yes.  I'll use it there.

theano/sandbox/gpuarray/conv.py
abergeron: Because this sets max_threads_dim0 which is then use in c_code() or some other place.

theano/sandbox/gpuarray/dnn.py
abergeron: Because you need to have called dnn_available() at least once before calling version().

theano/sandbox/gpuarray/blas.py
```diff
@@ -28,34 +29,22 @@ def c_header_dirs(self):
...
-    out = pygpu_copy(y, GA_ANY_ORDER);
-  }
-  return out;
-}
-"""
-
-
-class GpuGemv(BlasOp, Gemv):
+
+class GpuGemv(BlasOp):
```
lamblin: Why are you removing the inheritance from `Gemv`? Now it gives an error when calling `GpuGemv(inplace=...)` below.

abergeron:
Ok done.

lamblin:
OK, this looks good to me, provided that all GPU tests (old and new back-ends) pass when executed locally.

abergeron:
The new backend tests pass with the last commit, I'm still running the old one.

abergeron:
One test fails with the old back-end:

test_dnn.py:test_conv3d_bwd((8, 1, 2, 3, 1030), (5, 1, 1, 2, 1025), (1, 1, 1), 'full', 'cross')

This is an odd failure since all the values are off by 0.25 exactly.  I don't think that it is because anything I did.  I'll try with master to make sure.

lamblin:
Some of these tests seem to be failing at random in the master as well, and I think it is just a question of tolerance, added to the non-deterministic behaviour of backward convolutions.
In the case I managed to reproduce (test_conv3d_bwd((8, 1, 1030, 3, 4), (5, 1, 1025, 1, 1), (1, 1, 1), 'full', 'conv')), all the elements of the output have the same value, and the relative error is small:
- val1: all equal to 49368.566
- val2: all equal to 49369.145
- absolute difference: 0.578125
- relative difference: 5.85516e-06

In another one (theano.sandbox.cuda.tests.test_dnn.test_conv3d_bwd((8, 1, 1030, 3, 4), (5, 1, 1025, 1, 1), (1, 1, 1), 'full', 'conv')), the difference was not constant, but the max relative difference was 5.60961e-06.

This looks like rounding error to me.

lamblin:
Also, the relative difference that gets printed in the message is actually _half_ of what gets used during the checking, which is why even with `rtol = 1e-05`, it fails.

theano/sandbox/gpuarray/__init__.py
```diff
@@ -52,11 +56,21 @@ def init_dev(dev):
             optdb.add_tags('gpuarray_opt', 'fast_run', 'fast_compile')
         elif (config.init_gpu_device.startswith('cuda') or
               config.init_gpu_device.startswith('opencl')):
+            if config.device != 'cpu':
```
nouiz: This condition isn't good. It must be == as in the old back-end.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -19,27 +21,83 @@
...
+                if isinstance(x.owner.op, GpuToGpu):
+                    x = x.owner.inputs[0]
+                    continue
+
+            # If none of the conditions where met, then continue with
+            # the rest of the body
+            break
+
+        # If we couldn't deal with transfers, then maybe it's a tensor
+        if isinstance(x.type, tensor.TensorType):
```
nouiz: Why special case TensorType? The generic version should work. If not, add a comment to tell why.

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -319,19 +398,67 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+    def R_op(self, inputs, eval_points):
+        return self(eval_points[0])
+
+    def infer_shape(self, node, xshp):
+        return xshp
+
+    def c_code(self, node, name, inputs, outputs, sub):
+        return """
+        Py_XDECREF(%(out)s);
+        %(out)s = pygpu_transfer(%(inp)s, %(ctx)s, 0);
```
nouiz: It would be useful to make pygpu_transfer take an output array for the destination. This would make allow_gc=False Speed up things here.

theano/sandbox/gpuarray/opt.py
```diff
@@ -102,28 +97,53 @@ def op_lifter(OP, cuda_only=False):
...
+                new_op = maker(node, context_name)
+                # This is needed as sometimes new_op inherits from OP.
+                if new_op and new_op != node.op:
+                    if isinstance(new_op, theano.Op):
+                        # tag the inputs with the context in case
+                        # the context was derived from the outputs
+                        def tag(i, ctx):
+                            i.tag.context_name = ctx
+                            return i
+                        inputs = [tag(i, context_name) for i in node.inputs]
```
nouiz: This could be False. I mean, if one node have inputs on different context, then this would set the tag.context_name of some of them to the place we will read, not where it is. I don't think we should do that.

theano/sandbox/gpuarray/opt.py
nouiz: Add that in the a follow up ticket.

theano/sandbox/gpuarray/opt.py
```diff
@@ -146,35 +166,81 @@ def apply(self, fgraph):
...
+            node.inputs[0].owner and
+            node.inputs[0].owner.op == host_from_gpu):
+        other = node.inputs[0].owner.inputs[0]
+        if node.op.context_name == other.type.context_name:
+            return [other]
+        else:
+            return [GpuToGpu(node.op.context_name)(other)]
+
+    # ? -> gpua -> host
+    elif (node.op == host_from_gpu and
```
nouiz: use isinstance for consistency and this is faster.

theano/sandbox/gpuarray/basic_ops.py
lamblin: What generic version? `TensorType` does not have `as_gpuArray_variable` or anything like that.

theano/sandbox/gpuarray/__init__.py
lamblin: What do you mean? If `init_gpu_device` is specified and `device` is _not_ `cpu`, then it is an error.
The old backend has `assert device == 'cpu'`, which is equivalent.

theano/sandbox/gpuarray/opt.py
lamblin: If I understood correctly, currently it is required that all inputs have the same context.

abergeron: The tag can disagree with the type, it's a hint for the op about in which context it should execute.  It doesn't matter if the inputs are not actually in that context.

There may be cases where the inputs are not on the same context and in that case, they will all get transferred to the chosen context.

theano/sandbox/gpuarray/opt.py
abergeron: I'm pretty sure I did that, actually.  I may have forgot some places where it should be changed though, like here.  But I made the new exception inherit from ValueError to avoid problems in the meantime.

theano/sandbox/gpuarray/opt.py
abergeron: Err, it seems to have been lost somewhere.

theano/sandbox/gpuarray/basic_ops.py
nouiz: right, but for the old back-end, I taught that we duck-type one to TensorType. But I checked and I don't find it anymore. So this is good.

theano/sandbox/gpuarray/__init__.py
nouiz: ok. I compared too fast that part of the code

PR #3486, opened 2015-10-02
Fix daily buildbot tests

nouiz:


lamblin:
OK for me if Travis passes.

PR #3492, opened 2015-10-05
Device name in compiledir (Illegal instruction)

fmder:
When using both cpu and gpu on the same computer (or device with nfs file system) an Illegal instruction error comes up if the compile dir served to compile cpu/gpu trees when using a gpu/cpu configuration (note the inversion). The resulting `compiledir`s would now be

compiledir_Darwin-15.0.0-x86_64-i386-64bit-i386-2.7.10-64-cpu
compiledir_Darwin-15.0.0-x86_64-i386-64bit-i386-2.7.10-64-gpu

just by changing the device in the config.

### DIFF
--- a/theano/gof/compiledir.py
+++ b/theano/gof/compiledir.py
@@ -65,7 +65,8 @@ compiledir_format_dict = {
     "theano_version": theano.__version__,
     "numpy_version": numpy.__version__,
     "gxx_version": gcc_version_str.replace(" ", "_"),
-    "hostname": socket.gethostname()}
+    "hostname": socket.gethostname(),
+    "device": config.device}
 
 
 def short_platform(r=None, p=None):
@@ -146,7 +147,7 @@ def short_platform(r=None, p=None):
 compiledir_format_dict['short_platform'] = short_platform()
 compiledir_format_keys = ", ".join(sorted(compiledir_format_dict.keys()))
 default_compiledir_format = ("compiledir_%(short_platform)s-%(processor)s-"
-                             "%(python_version)s-%(python_bitwidth)s")
+                             "%(python_version)s-%(python_bitwidth)s-%(device)s")
 
 AddConfigVar("compiledir_format",
              textwrap.fill(textwrap.dedent("""\


theano/gof/compiledir.py
```diff
@@ -65,7 +65,8 @@ def python_int_bitwidth():
     "theano_version": theano.__version__,
     "numpy_version": numpy.__version__,
     "gxx_version": gcc_version_str.replace(" ", "_"),
-    "hostname": socket.gethostname()}
+    "hostname": socket.gethostname(),
+    "device": config.device}
```
nouiz: that is fine.

theano/gof/compiledir.py
```diff
@@ -146,7 +147,7 @@ def short_platform(r=None, p=None):
 compiledir_format_dict['short_platform'] = short_platform()
 compiledir_format_keys = ", ".join(sorted(compiledir_format_dict.keys()))
 default_compiledir_format = ("compiledir_%(short_platform)s-%(processor)s-"
-                             "%(python_version)s-%(python_bitwidth)s")
+                             "%(python_version)s-%(python_bitwidth)s-%(device)s")
```
nouiz: That isn't good. I understand that this fix your problem, but this isn't the right fix. I'm ok to allow the user to use the device in the compiledir, but we should not change the default version. We should fix the underling problem.

Most of the time, this is a bad detection by g++ of the processor. We already have a few work around g++ bug in Theano. What if you processor?

Could try with the normal compiledir and with this flag:

gxx.cxxflags = -march=corei7

What is your g++ vesrion? Your exact CPU error?

Another user reported similar problem recently:

https://groups.google.com/forum/#!topic/theano-users/w-OIYeLSlZk

nouiz:
Can you give the full error you get?

fmder:
Here is the procedure I use to recreate the error. I use two clusters with
shared nfs, one is GPU and one is CPU. The GPU cluster have Xeon Ivy Bridge
E5-2670 v2 processors and K20/K80 Teslas. The CPU cluster have Intel Xeon
X5560. I tried repeatedly with gcc 4.5, 4.8 and 4.9.

First I run Theano with the following on the GPU cluster

export THEANO_FLAGS="device=gpu,base_compiledir=$HOME/theano-test-dir"

python -c 'import theano; print "success"'

No error, the sucess appears in the out file

Then on the CPU cluster I run:

export
THEANO_FLAGS="device=cpu,base_compiledir=$HOME/theano-test-dir,gxx.cxxflags=-march=corei7"

python -c 'import theano; print "success"'


The out file shows nothing, and the err file shows only this:

/var/spool/torque/mom_priv/jobs/1572431.SC: line 43: 14127 Illegal
instruction     python -c 'import theano; print "success"'

If I delete the compiledir everything runs fine on the CPU cluster.

Best regards,
François-Michel

2015-10-05 12:40 GMT-04:00 Frédéric Bastien <notifications@github.com>:

nouiz:
that give good indication. I did this on one computer and Theano compiled 2
files. They are global files that don't have versioning. That is why you
have this problem. The long term solution would be to version them. But
this will take some times to implement.

A good work around would be do disable the instruction set that cause
problem for those 2 files. They aren't needed and shouldn't affect much
performance.

Can you go in the file theano/gof/cmodule.py, in the method compile_args(),
Can you print the output it return at the end of this method? Can you do
this on both cluter and give me both output? With that, I can make another
fix.

thanks

nouiz:
I'll close this PR as there is no update for long. I think you have resolved your problem. If not, tell us.

As said in a comment, we should not add the "device" flag by default in the compiledir, but if you want to be able to do this, I'm fine to add this possibility.

PR #3493, opened 2015-10-05
Scan fixes and speedup

carriepl:
1 - Speed up ScanInplaceOptimizer by prefiltering the list of outputs to attempt to compute inplace.
2 - Give In() variables with updates a default value (to avoid warnings with mode=DebugMode)


carriepl:
Note for the first point : on Daniel's use case, compilation time drops from ~1h23m to ~30m.

carriepl:
Did the changes you asked for and still have the speedup for Daniel's case.

nouiz:
Just a small comment on what was found. The mit-mot PR used the IN interface. This cause a reference to be kept for them between execution of apply node. @carriepl will work on this.

carriepl:
Just a small comment : after investigation, the reference that was kept was already cleaned up by the Scan backend so there is nothing so correct here.

PR #3494, opened 2015-10-05
Convert double recursion to simple recursion

carriepl:
This is done by in-lining one method in the other one.

### DIFF
--- a/theano/compile/pfunc.py
+++ b/theano/compile/pfunc.py
@@ -83,7 +83,17 @@ def rebuild_collect_shared(outputs,
         if v in clone_d:
             return clone_d[v]
         if v.owner:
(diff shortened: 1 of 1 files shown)


carriepl:
This allows to support graphs up to 2x bigger but now, if we want to go bigger, we need to change every place in Theano where something is recursive in the number of nodes/variables in a Theano graph... and there's quite a few of those.

nouiz:
travis fail

nouiz:
It is pep8.

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/compile/pfunc.py:92:45: E126 continuation line over-indented for hanging indent

carriepl:
I did Ye Olde Flake8, was there anything else?

PR #3500, opened 2015-10-06
map_variables: deal with cached constants

cooijmanstim:
map_variables fails when expressions that are not simply inputs are replaced by expressions that contain cached constants. This PR adds a failing test and a patch to fix the problem. Also, the argument additional_inputs is no longer needed as newly introduced inputs are automatically added using FunctionGraph.add_input.

### DIFF
--- a/theano/scan_module/scan_utils.py
+++ b/theano/scan_module/scan_utils.py
@@ -317,8 +317,6 @@ def map_variables(replacer, graphs):
         new_graph = replacer(graph)
         # deal with cached constants and new inputs in the replacement
         if new_graph is not graph:
-            # FIXME: all this cloning breaks the graphs_seen cache,
-            # leading to exponentially many replacements
             [new_graph] = without_cached_constants([new_graph])
             for input_ in gof.graph.inputs([new_graph]):
                 # if the input is already owned by one of the outer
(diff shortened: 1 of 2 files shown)


nouiz:
@lamblin do you have time to view this? You reviewed the original version.

lamblin:
I did not review the first version, @carriepl did.

cooijmanstim:
I've been using map_variables myself and find that using it properly (to avoid ballooning the graph with cloned subgraphs) requires a fair bit of support code. Maybe I should put that code and map_variables into a separate repository. It doesn't really touch theano internals anyway. What do you think?

lamblin:
I would prefer that to be part of Theano, actually, since that can be a useful feature to expose to people using Theano at the graph-manipulation level. Not sure what the others are thinging.

nouiz:
I would go in that direction except if the code dump is very big.

cooijmanstim:
The cache is now cleared after every call.

lamblin:

lamblin:
Is this ready, or is something still blocking this PR?

nouiz:
This PR is looking for someone to review it. Who reviewed that map_variable at first?

cooijmanstim:
@carriepl used to review these.

PR #3504, opened 2015-10-07
Fix dnn_conv and dnn_pool docs

### LINKED ISSUE #3503: dnn_pool padding documentation is wrong
> **pad** – (padX, padY) padding information. padX is the size of the left and right borders, padY is the size of the top and bottom borders.

But this is clearly not the case:

```
In [42]: from theano.sandbox.cuda.dnn import dnn_pool

In [43]: from theano.tensor import tensor4

In [44]: a = tensor4()

In [45]: q = dnn_pool(a, (2, 2), stride=(2, 2), pad=(4, 1))

In [46]: image = numpy.random.normal(size=(128, 3, 32, 32)).astype('float32')

In [47]: numpy.array(f(image)).shape
Out[47]: (128, 3, 20, 17)
```

dwf:
Make them clearer and also correct the inverted x and y coordinates.

Fixes #3503.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -1290,8 +1290,10 @@ class GpuDnnPoolDesc(GpuOp):
         The old deprecated name 'average' correspond to 'average_inc_pad'.
     pad
         (pad_h, pad_w) padding information.
-        pad_h is the number of zero-valued pixels added to the top and bottom borders.
-        pad_w is the number of zero-valued pixels added to the left and right borders.
+        pad_h is the number of zero-valued pixels added to each of the top and
+        bottom borders.
+        pad_w is the number of zero-valued pixels added to each of the left and
+        right borders.
 
     """
 
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -1105,8 +1105,11 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1),
     kerns
         Convolution filters.
     border_mode
-        One of 'valid', 'full'; additionally, the padding size could be directly
-        specified by an integer or a pair of integers.
+        One of 'valid', 'full'; additionally, the padding size can be
+        directly specified by an integer or a pair of integers (as a tuple),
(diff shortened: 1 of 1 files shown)


nouiz:
thanks

abergeron:
Just a note here that you've pushed a branch onto the main repo.  Please don't do that.

dwf:
I actually did the "edit in the browser" thing as I didn't have a lot of
time on my hands, and for some reason GitHub idiotically puts it on the
main repo fork rather than my own fork.

dwf:
As soon as it's merged I'll delete this branch, though.

abergeron:
Maybe it's because you did the edit thing while browsing the main repo?  I don't know I've never tried it.

theano/sandbox/cuda/dnn.py
```diff
@@ -1752,10 +1760,13 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
...
-        padX is the size of the left and right borders,
-        padY is the size of the top and bottom borders.
-    :param nd: dimensions of pooling, can be 2 or 3 for 2d or 3d pooling
+    	(pad_h, pad_w) padding information.
+        pad_h is the number of zero-valued pixels added to each of the top and
+        bottom borders.
+        pad_w is the number of zero-valued pixels added to each of the left
+        and right borders.
+    nd
+    	Number of dimensions of pooling, can be 2 or 3 for 2d or 3d pooling
```
abergeron: dnn_pool doesn't take an `nd` parameter.  Maybe it should, but it doesn't currently.

nouiz:
I merge. thanks

nouiz:
I deleted the branch

theano/sandbox/cuda/dnn.py
dwf: @abergeron re: the nd parameter, that was already there, but using a
different doc syntax. It was rendering weird in the doc. Maybe it should be
removed then.

abergeron: Yes, I see it now.  It's better this way for now.

nouiz: I fixed this in another pr.

Issue #2968, opened 2015-05-27
CPU/AbstractOp batched gemm and GPU optimization


nouiz:
This is a follow up from https://github.com/Theano/Theano/pull/2893.

- [x] add grad https://github.com/Theano/Theano/pull/2965
- [x] add infer_shape
- [x] Make a CPU op for it (instead of a graph with scan) or make an AbstractOp that get specialisez for the right device. As the specialization will be done after scan optimization, I think making a CPU op is better. [started](https://github.com/Theano/Theano/pull/1698/files#diff-d600105fbdb446c0fc639dd21274fd40)
   - [x] Make opt to convert it to the existing GPU op.
- [ ] The making of the vector of indices can be made by CUDA code to speed it up. What is done in libgpuarray?

PR #3508, opened 2015-10-09
WIP: BatchedDotOp

cooijmanstim:
Work in progress on #2968


theano/sandbox/cuda/tests/test_blas.py
```diff
@@ -115,6 +115,19 @@ def test_batched_dot_gradient(self):
...
+    def test_infer_shape(self):
+        # only matrix/matrix is supported
+        admat = tensor.ftensor3()
+        bdmat = tensor.ftensor3()
+        admat_val = my_rand(7, 4, 5)
+        bdmat_val = my_rand(7, 5, 3)
+        self._compile_and_check([admat, bdmat],
+                                [BatchedDotOp()(admat, bdmat)],
+                                [admat_val, bdmat_val],
+                                (BatchedDotOp))
```
nouiz: There is extra () here.

cooijmanstim:
I implemented the CPU op theano.tensor.BatchedDot based on Dot and a test using makeTester. Does this also test the gradient?

nouiz:
this do not test the grad. For this, you need to pass a grad parameter to makeTester. I didn't do a full review, just very quick to answer your answer.

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
     return ret
 
 
-def batched_dot(x, y):
+class BatchedDot(Op):
     """
-    This function computes the dot product between the two tensors, by
-    iterating over the first dimension using scan.
+    Dot but Batched
```
carriepl: Go on....

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
-    tensor
-        A tensor of size e.g. if it is 3D: (dim1, dim3, dim4).
+    Notes
+    -----
+    Matrix-matrix products are sometimes optimized to Dot22 or Gemm ops
+    (see tensor.blas).
+    Vector-vector products are sometimes optimized to Ger or CGer (see
+    tensor.blas).
+    Matrix-vector products are sometimes optimized to Gemv, CGemv (see
+    tensor.blas).
```
carriepl: This part and the one before seem copy pasted from the Dot op and don't quite fit with the BatchedDot op. You can start from the docstring of the method batched_dot() that you removed.

theano/tensor/tests/test_basic.py
```diff
@@ -1930,6 +1930,46 @@ def _numpy_true_div(x, y):
                         bad_build=dict(),
                         bad_runtime=dict(bad1=(rand(5, 7), rand(5, 7)),
                                          bad2=(rand(5, 7), rand(8, 3))))
+BatchedDotTester = makeTester(
```
carriepl: This part I will have to look in more details because I am not familiar with makeTester()

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
     -----
-    This is a subset of numpy.einsum, but we do not provide it for now.
-    But numpy einsum is slower than dot or tensordot:
-    http://mail.scipy.org/pipermail/numpy-discussion/2012-October/064259.html
+    Matrix-matrix products are sometimes optimized to Dot22 or Gemm ops
+    (see tensor.blas).
+    Vector-vector products are sometimes optimized to Ger or CGer (see
+    tensor.blas).
+    Matrix-vector products are sometimes optimized to Gemv, CGemv (see
+    tensor.blas).
```
carriepl: Those comments don't seem quite right. They seem taken from some dot() function or something similar. They should be adapted to reflect what batched_dot will do : take the two elements, iterate over the first dimension and call dot on every pair of sub-elements.

theano/tensor/tests/test_basic.py
carriepl: This being said, there seems to be insufficient test cases with involving tensors with less or more than 3 dimensions. Ex :  what happens if you call batched_dot with two vectors? two matrices? a scalar and something else? a 4d/5d tensor with something else?

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+
+    def grad(self, inp, grads):
+        x, y = inp
+        gz, = grads
+        xdim, ydim, gdim = x.type.ndim, y.type.ndim, gz.type.ndim
+
+        # grad is scalar, so x is vector and y is vector
+        if gdim == 0:
+            xgrad = gz.dimshuffle(0, 'x') * y
+            ygrad = gz.dimshuffle(0, 'x') * x
```
carriepl: How can this case occur if you are forcing x and y to have either 2 or 3 dimensions?

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+        # x is matrix, y is vector, grad is vector
+        elif xdim == 3 and ydim == 2:
+            xgrad = gz.dimshuffle(0, 1, 'x') * y.dimshuffle(0, 'x', 1)
+            ygrad = batched_dot(x.dimshuffle(0, 2, 1), gz)
+
+        # x is matrix, y is matrix, grad is matrix
+        elif xdim == ydim == 3:
+            xgrad = batched_dot(gz, y.dimshuffle(0, 2, 1))
+            ygrad = batched_dot(x.dimshuffle(0, 2, 1), gz)
+
```
carriepl: What if both x and y have 2 dimensions?

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+                'theano.tensor.batched_dot instead.' % inputs[0].ndim)
+        if inputs[1].ndim not in (2, 3):
+            raise TypeError(
+                'theano.tensor.BatchedDot: input 1 (0-indexed) must have ndim'
+                'of 2 or 3, %d given. Consider calling '
+                'theano.tensor.batched_dot instead.' % inputs[1].ndim)
+
+        dtype = scal.upcast(*[input.type.dtype for input in inputs])
+        broadcastable = (inputs[0].type.broadcastable[:-1] +
+                         inputs[1].type.broadcastable[2:])
```
carriepl: If at least one input is broadcastable along the first dimension, the output should be broadcastable along that dimension (currently, if the first input is not but the second is, the output will not be broadcastable along the first dimension).

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+        # This cause problem down the road. See gh-1461.
+        if xgrad.broadcastable != x.broadcastable:
+            xgrad = patternbroadcast(xgrad, x.broadcastable)
+        if ygrad.broadcastable != y.broadcastable:
+            ygrad = patternbroadcast(ygrad, y.broadcastable)
+
+        rval = xgrad, ygrad
+
+        for elem in rval:
+            assert elem.dtype.find('float') != -1
```
carriepl: This is probably not needed, the grad() method already ensures that the `gz` you receive will be a floating point value and since `gz` is involve in computing both `xgrad` and `ygrad`, it means that they will also be floating point.
Plus, I think if you return non-float gradients, tensor.grad() will already complain.

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+
+        return rval
+
+    def R_op(self, inputs, eval_points):
+        # R_op for a \dot b evaluted at c for a and d for b is
+        # simply c \dot b + a \dot d
```
carriepl: The R_op is now really something I understand really well. @lamblin, would you mind having a look at this method?

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+
+        assert len(inputs) == 2
+        assert len(eval_points) == 2
+        if eval_points[0] is None and eval_points[1] is None:
+            return [None]
+
+        debugger_available = config.compute_test_value != 'off'
```
carriepl: Most ops don't do anything special related to the test_values. Why is this needed here?

theano/tensor/basic.py
```diff
@@ -3340,43 +3340,236 @@ def transpose(x, axes=None):
...
+
+    def infer_shape(self, node, shapes):
+        for shape_ in shapes:
+            if len(shape_) not in (2, 3):
+                raise NotImplementedError()
+        xshp, yshp = shapes
+        return [xshp[:-1] + yshp[2:]]
+
+    def __str__(self):
+        return "batched_dot"
```
carriepl: Since you are defining the `__props__` attribute, you do not need to define this method, it should be automatically handled.

theano/tensor/basic.py
lamblin: It looks like it is simply to provide better error messages in case of shape mismatch, in particular which variables are involved, and what they mean (first or second input or eval point).
This is not necessary to keep that section, but if we do, we should update the strings that keep refering to `Dot`.

theano/tensor/basic.py
lamblin: It looks good to me, batched dot is a linear operation, as dot, so the same logic applies.

theano/tensor/basic.py
cooijmanstim: The condition here should have been "gdim == 1", which resolves both this comment and the next.

theano/tensor/basic.py
cooijmanstim: I hadn't thought about this, but now that I have, that doesn't seem right. The way I understand it, a dimension should be broadcastable only if it was broadcastable in both inputs. Otherwise the output will have a definite (possibly >1) size in that dimension.

theano/tensor/basic.py
cooijmanstim: That sounds right, I will remove it. It came from the Dot code and was introduced by Ian Goodfellow in 7966b05cc34fb3acdb3157c11efbdbd115988dcb.

cooijmanstim:
Please take a look at the optimization; I had to put in some special code to handle the cases where either or both operands are matrices (i.e. batches of vectors). I pad their shapes so they are both third-order tensors and then unpad the result.

Maybe it's better to simplify the BatchedDot op to only take third-order tensors in the first place, and do this padding/unpadding in the batched_dot function, or to modify the GPU op's make_node to handle the padding/unpadding. It'd be nice to have the same interface for both ops so the optimization can be simple.

theano/tensor/basic.py
lamblin: As far as I understand, batched dot should not perform broadcasting, so both dimensions should have the same length, which is also the length along the corresponding output dimension.
So, if one of the inputs is broadcastable, then the other input should also have a length of 1 at runtime, and the output should be broadcastable as well.

theano/tensor/basic.py
```diff
@@ -3340,52 +3340,214 @@ def transpose(x, axes=None):
...
+        # This cause problem down the road. See gh-1461.
+        if xgrad.broadcastable != x.broadcastable:
+            xgrad = patternbroadcast(xgrad, x.broadcastable)
+        if ygrad.broadcastable != y.broadcastable:
+            ygrad = patternbroadcast(ygrad, y.broadcastable)
+
+        return xgrad, ygrad
+
+    def R_op(self, inputs, eval_points):
+        # R_op for a \dot b evaluted at c for a and d for b is
```
carriepl: This still refers to Dot

cooijmanstim:
Pushing tentative C implementation for feedback. Still in progress (it segfaults).

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,499 @@ def transpose(x, axes=None):
...
+        assert len(z_dims) == z_ndim
+
+        z_shape_correct = " && ".join("PyArray_DIMS(%s)[%i] == %s"
+                                      % (_z, i, dim) for i, dim in enumerate(z_dims))
+        z_shape = ", ".join(z_dims)
+        allocate = """
+            if ((NULL == %(_z)s) || !(%(z_shape_correct)s))
+            {
+                npy_intp dims[3] = {%(z_shape)s};
+                if (NULL != %(_z)s) Py_XDECREF(%(_z)s);
```
carriepl: Py_XDECREF will already not decrease the refcount if `_z` is NULL so there's no need for the `if`.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,499 @@ def transpose(x, axes=None):
...
+                %(_z)s = (PyArrayObject*)PyArray_SimpleNew(
+                    %(z_ndim)s, dims, PyArray_TYPE(%(_x)s));
```
carriepl: Does PyArray_SimpleNew allocate F-contiguous, C-contiguous or possibly-non contiguous arrays? I think, in general, if we need to allocate outputs, we should favor C-contiguous one so, if PyArray_SimpleNew doesn't do that, you can call PyArray_EMPTY instead.

carriepl: The NumPy C-API documentation doesn't seem to give much information regarding this question : http://docs.scipy.org/doc/numpy-1.10.0/reference/c-api.array.html#c.PyArray_SimpleNew

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,499 @@ def transpose(x, axes=None):
...
+            contiguate.append("""
+                if (%(not_contiguous)s) {
+                    PyArrayObject * _copy = (PyArrayObject *) PyArray_Copy(%(var)s);
+                    if (!_copy)
+                        %(fail)s
+                    Py_XDECREF(%(var)s);
+                    %(var)s = _copy;
+                }
+            """ % locals())
+        contiguate = "\n".join(contiguate)
```
carriepl: There will probably be something to adjust here depending on the answer to my question about PyArray_SimpleNew. If SimpleNew only allocates C-contiguous arrays, then it would be more efficient to  change the check earlier to see if `_z` is NULL or not the right shape or not C-contiguous and reallocate it if needed. If SimpleNew doesn't always return C-contiguous arrays, it should be changed to something that does and then the earlier check should be changed as I said in the last sentence. In both cases, the `contiguate` part of the C code should be changed to match.

theano/tensor/basic.py
lamblin: The default is C-contiguous, and a quick look at the source (https://github.com/numpy/numpy/blob/60d1bc626b73c45735f23bf7dd37539fa67d4ddd/numpy/core/include/numpy/ndarrayobject.h#L134) confirms that explicit strides are not passed, and that the flags are the default ones.

theano/tensor/basic.py
cooijmanstim: I'm not sure where the inefficiency is. SimpleNew allocates C-contiguously[1]. It won't reallocate `_z` twice; after the first (re)allocation `_z` is contiguous and so the second reallocation won't be triggered.

[1] By @lamblin's comment https://github.com/Theano/Theano/pull/3508#discussion_r49207229

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,499 @@ def transpose(x, axes=None):
...
+                    {
+                        case 0x000: sgemm_(&N, &N, &Nz2, &Nz1, &Nx2, &a, y, &sy_1, x, &sx_1, &b, z, &sz_1); break;
+                        case 0x100: sgemm_(&N, &T, &Nz2, &Nz1, &Nx2, &a, y, &sy_1, x, &sx_2, &b, z, &sz_1); break;
+                        case 0x010: sgemm_(&T, &N, &Nz2, &Nz1, &Nx2, &a, y, &sy_2, x, &sx_1, &b, z, &sz_1); break;
+                        case 0x110: sgemm_(&T, &T, &Nz2, &Nz1, &Nx2, &a, y, &sy_2, x, &sx_2, &b, z, &sz_1); break;
+                        case 0x001: sgemm_(&T, &T, &Nz1, &Nz2, &Nx2, &a, x, &sx_1, y, &sy_1, &b, z, &sz_2); break;
+                        case 0x101: sgemm_(&N, &T, &Nz1, &Nz2, &Nx2, &a, x, &sx_2, y, &sy_1, &b, z, &sz_2); break;
+                        case 0x011: sgemm_(&T, &N, &Nz1, &Nz2, &Nx2, &a, x, &sx_1, y, &sy_2, &b, z, &sz_2); break;
+                        case 0x111: sgemm_(&N, &N, &Nz1, &Nz2, &Nx2, &a, x, &sx_2, y, &sy_2, &b, z, &sz_2); break;
+                        default: PyErr_SetString(PyExc_ValueError, "some matrix has no unit stride"); %(fail)s;
```
carriepl: Since the exact gemm call is dependant on the stride structure, the tests should make sure to test each structure at least once. Is it currently the case?

carriepl:
I've only done a quick review so far since Travis are still failing and you mentioned running the tests locally could result in segfaults.

theano/tensor/basic.py
cooijmanstim: I think it is tested for `x` and `y` (by the cases in `BatchedDotTester` where one or more of the axes are singletons), and I think consequently for `z` as well. This code comes from `Dot` and the `BatchedDot` tests are a superset of the `Dot` tests, so if this is not tested then neither is the corresponding the `Dot` code.

cooijmanstim:
I've found the bug; it turns out that numpy strides count bytes rather than elements. It's fixed now. Tomorrow I'll deal with mixed types and double-check what happens with higher-order inputs. I should be able to finalize it this week.

cooijmanstim:
By the way, the code is easily generalized to complex dtypes, but when I tried to I ran into the issue that the `cgemm_` and `zgemm_` prototypes from `blas_headers.py` take `float*` and `double*` rather than complex types. Not sure if that's intended or not; in any case I decided not to waste my time on it now.

carriepl:
Flake8 is not happy

carriepl:
I'll have a look so I can answer the broadcasting question

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
-        non_sequences=None)
-    return result
+    Compute the batched dot product of two variables:
+
+        batched_dot(a, b)[i] = dot(a[i], b[i])
+
+    Note that this batched_dot function does one of three things, in the
+    following sequence:
+
+        1.  If either a or b is a vector, it returns the batched elementwise
```
nouiz: would it be more clear if we replace batched by broadcasted in this sentence? Or add broadcasted in paranthesis next to batched?

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+                'theano.tensor.BatchedDot: inputs [%s] must have the same size'
+                'in axis 0, but have sizes [%s].'
+                % (", ".join(map(str, inp)),
+                   ", ".join([str(i.shape[0]) for i in inp])))
+
+        shape = self.infer_shape(node, [i.shape for i in inp])[0]
+        dtype = node.outputs[0].dtype
+        z[0] = numpy.empty(shape, dtype=dtype)
+        for i in xrange(z[0].shape[0]):
+            z[0][i] = numpy.dot(x[i], y[i])
```
nouiz: doing z[0] at each iteration is "costly". So to speed this up, make something like:

z[0] = ...
z_ = z[0]
# use z_ instead of z[0]

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
     return ret
 
 
-def batched_dot(x, y):
+class BatchedDot(Op):
```
nouiz: I would move that op to tensor/blas.py.
I would then use theano.tensor.blas.BatchedDot inside the batched_dot method in this file.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+                y += Sy[0] / type_size;
+                z += Sz[0] / type_size;
+            }
+
+            return 0;
+        }
+        """
+        return blas_header_text() + batch_gemm_defn
+
+    def c_libraries(self):
```
nouiz: What about making an BlasOp inside the file blas.py and make all the op that duplicate this code inherit from it? This is a nice to have, not mandatory to do for this PR.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+                    PyErr_SetString(PyExc_MemoryError,
+                                    "failed to alloc BatchedDot output");
+                    %(fail)s
+                }
+            }
+        """ % locals()
+
+        # generate code to reallocate inputs/output contiguously if necessary
+        contiguate = []
+        for var, ndim in [(_x, x_ndim), (_y, y_ndim), (_z, z_ndim)]:
```
nouiz: z should not be here I think. I think that is z isn't c contiguous, you can just in the allocate block, reallocate a new one.
It will be overwritten in all cases, we don't want to copy it. You could keep it here (so we could preallocate it and reuse it with some strides pattern, but make sure to don't copy it here.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+            strides = "PyArray_STRIDES(%s)" % var
+            not_contiguous = " || ".join([
+                " || ".join("{strides}[{i}] < 1 || {strides}[{i}] % type_size"
+                            .format(strides=strides, i=i) for i in range(ndim)),
+                "(%s)" % " && ".join("{strides}[{i}] != type_size"
```
nouiz: why this part of the condition is needed (line 3573 and 3574)? Above, there is already  {strides}[{i}] % type_size above.

theano/tensor/basic.py
nouiz: If the output is provided, but not contiguous, then this will copy the output instead of just allocating new memory. copy is costly then allocation for the same size. I made a similar comment.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+            && (PyArray_DESCR(xs)->type_num != NPY_FLOAT))
+        {PyErr_SetString(PyExc_NotImplementedError, "type(x) is not double or float"); %(fail)s;}
+
+        if ((PyArray_DESCR(ys)->type_num != NPY_DOUBLE)
+            && (PyArray_DESCR(ys)->type_num != NPY_FLOAT))
+        {PyErr_SetString(PyExc_NotImplementedError, "type(y) is not double or float"); %(fail)s;}
+
+        if ((PyArray_DESCR(zs)->type_num != NPY_DOUBLE)
+            && (PyArray_DESCR(zs)->type_num != NPY_FLOAT))
+        {PyErr_SetString(PyExc_NotImplementedError, "type(z) is not double or float"); %(fail)s;}
```
nouiz: you always create zs yourself. So this check isn't useful.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+
+        if ((PyArray_DESCR(xs)->type_num != PyArray_DESCR(ys)->type_num)
+            ||(PyArray_DESCR(xs)->type_num != PyArray_DESCR(zs)->type_num))
```
nouiz: idem, zs is allocated here, so no need to check its dtype.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+            case NPY_DOUBLE:
+            if (batch_gemm<double>(dgemm_, type_size, xs, ys, zs)) {
+                %(fail)s;
+            }
+            break;
+        }
+        """ % locals()
+
+    def c_code_cache_version(self):
+        return None
```
nouiz: before merging, don't forget to add one version.

theano/tensor/tests/test_basic.py
```diff
@@ -1932,6 +1932,61 @@ def _numpy_true_div(x, y):
                         bad_runtime=dict(bad1=(rand(5, 7), rand(5, 7)),
                                          bad2=(rand(5, 7), rand(8, 3))))
 
+BatchedDotTester = makeTester(
+    name='BatchedDotTester',
+    op=batched_dot,
+    expected=(lambda xs, ys:
```
nouiz: this is hard to read. Can you make a function numpy_batched_dot above BatchedDotTester and reuse it here?

theano/tensor/tests/test_basic.py
```diff
@@ -1932,6 +1932,61 @@ def _numpy_true_div(x, y):
...
+        numpy.asarray(
+            list(x * y if x.ndim == 0 or y.ndim == 0
+                 else (numpy.dot(x, y) if y.ndim == 1
```
nouiz: In fact, I would do 

if x.ndim == 0 or y.ndim == 0:
 ...
else: numpy.dot(x, y)

the call to tensordot bellow could be buggy and is equivalent in this case. So this make simpler test.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,483 @@ def transpose(x, axes=None):
...
+    if a.ndim == 0:
+        raise TypeError("a must have at least one (batch) axis")
+    elif b.ndim == 0:
+        raise TypeError("b must have at least one (batch) axis")
+    elif a.ndim == 1:
+        return a.dimshuffle(*([0] + ["x"] * (b.ndim - 1))) * b
+    elif b.ndim == 1:
+        return a * b.dimshuffle(*([0] + ["x"] * (a.ndim - 1)))
+    elif a.ndim > 3 or b.ndim > 3:
+        return batched_tensordot(
```
nouiz: I don't see any tests for that case. in test_basic.py, I see tests for batched_dot with inputs have 3 dims. Can you add one where both inputs have 4 dims in the test test_batched_dot? Also make it compare for the good value, not just the shape.

nouiz:
I finished my review. I just got small comments, so you are pretty close! thanks.

theano/tensor/basic.py
cooijmanstim: I don't understand. `PyArray_Copy` is a macro that calls `PyArray_NewCopy`, which allocates new memory and makes a contiguous copy: http://docs.scipy.org/doc/numpy-1.10.0/reference/c-api.array.html#c.PyArray_NewCopy

In any case I'm doing the same as the code for Dot22 at https://github.com/Theano/Theano/blob/master/theano/tensor/blas.py#L821

theano/tensor/basic.py
cooijmanstim: I don't like batched, but I don't like broadcasted either. I was hoping the equation below it would clear things up.

theano/tensor/basic.py
cooijmanstim: There is already a `GemmRelated` class in blas.py, which I used as a starting point. It's interface didn't really work for my purposes and I didn't want to refactor it. Maybe `GemmRelated` can call this function to reduce the duplication.

theano/tensor/basic.py
cooijmanstim: I don't create zs myself if it already has the right shape and is contiguous. Right?

theano/tensor/basic.py
cooijmanstim: Oh I see what you're saying now. All I need is new memory; I don't need it to copy the data over. That's true for Dot22 as well though, right?

theano/tensor/basic.py
cooijmanstim: That condition requires that at least one of the axes has stride 1. It's from https://github.com/Theano/Theano/blob/master/theano/tensor/blas.py#L799 .

theano/tensor/basic.py
cooijmanstim: There is BatchedDotTester in test_basic.py that holds most of the tests, and it does have cases with 4-dim tensors.

theano/tensor/tests/test_basic.py
cooijmanstim: Good catch.

theano/tensor/basic.py
cooijmanstim: Yup. d8538e4

theano/tensor/basic.py
cooijmanstim: 001ddc3

theano/tensor/tests/test_basic.py
cooijmanstim: 0fde9a4

theano/tensor/basic.py
cooijmanstim: c539a57

theano/tensor/basic.py
cooijmanstim: 914ebf7

theano/tensor/tests/test_basic.py
cooijmanstim: I'm hoping it's readable enough after the simplification.

theano/tensor/basic.py
```diff
@@ -3372,52 +3372,53 @@ def transpose(x, axes=None):
...
+            a, b, [[a.ndim - 1], [numpy.maximum(1, b.ndim - 2)]])
+    else:
+        # avoid circular import
+        from blas import BatchedDot
```
carriepl: This import fails in travis.

theano/tensor/tests/test_basic.py
carriepl: I think the new version is easily understood

carriepl:
@cooijmanstim You seem to have adressed @nouiz. The only think left is the import error that Travis is reporting.

carriepl:
This was the last thing and Travis is happy so I'm merging. Thanks.

PR #3510, opened 2015-10-10
Add n>1 experiment in multinomial sampling on CPU

### LINKED ISSUE #3359: Support more than one experiment in multinomial sampling on CPU
This involves :
* Adapting the Python code
* Adapting the C code
* Adding tests for this new feature
* Update the opt that move it to the GPU to not move it if n != 1.

aalmah:
- Fixes #3359.
- Probably needs more tests, especially with very large number of outputs. The tests I added so far seem to work with very large outputs, but we might still need to implement more numerically stable summation [Kahan summation] (https://en.wikipedia.org/wiki/Kahan_summation_algorithm)


carriepl:
Travis complains about a few little flake8 errors:
- /home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/multinomial.py:16:1: E302 expected 2 blank lines, found 1
- /home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/multinomial.py:111:1: W293 blank line contains whitespace
- /home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/multinomial.py:171:1: W293 blank line contains whitespace

theano/sandbox/multinomial.py
```diff
@@ -45,18 +46,18 @@ def make_node(self, pvals, unis):
...
 
     def grad(self, ins, outgrads):
-        pvals, unis = ins
+        pvals, unis, n = ins
         (gz,) = outgrads
         return [T.zeros_like(x) for x in ins]
 
     def c_code_cache_version(self):
-        return (5,)
+        return None
```
carriepl: Setting it to 'None' is fine for development but you need to set if back to a reasonnable value like '(6,)' or else your PR will deactivate the C code caching for this op.

carriepl:
L95 in multinomial.py : You can allocate that memory using PyArray_EMPTY instead of PyArray_ZEROS since you'll overwrite the initial values anyway.

theano/sandbox/rng_mrg.py
```diff
@@ -1312,9 +1312,10 @@ def binomial(self, size=None, n=1, p=0.5, ndim=None, dtype='int64',
     def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
                     nstreams=None):
         """
-        Sample `n` (currently `n` needs to be 1) times from a multinomial
+        Sample `n` (currently `n` needs to be > 1) times from a multinomial
```
carriepl: That's not true, `n` needs to be >= 1

theano/sandbox/rng_mrg.py
```diff
@@ -1312,9 +1312,10 @@ def binomial(self, size=None, n=1, p=0.5, ndim=None, dtype='int64',
...
         distribution defined by probabilities pvals.
 
+        TODO: MODIFY_ME
```
carriepl: If this comment needs a modification because of the changes you did in this PR, you should update this comment yourself.

theano/sandbox/rng_mrg.py
```diff
@@ -1339,25 +1340,23 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        if ndim is not None:
+            raise ValueError("Provided an ndim argument to "
+                             "MRG_RandomStreams.multinomial, which does not use "
+                             "the ndim argument.")
+        if pvals.ndim == 2:
+                size = pvals[:,0].shape
+                unis = self.uniform(size=size, ndim=1, nstreams=nstreams)
+                op = multinomial.MultinomialFromUniform(dtype)
+                n_samples = as_tensor_variable(n)
+                return op(pvals, unis, n_samples)
```
carriepl: Indentation for this whole block of code

theano/sandbox/rng_mrg.py
```diff
@@ -1339,25 +1340,23 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        if size is not None:
+            raise ValueError("Provided a size argument to "
+                             "MRG_RandomStreams.multinomial, which does not use "
+                             "the size argument.")
+        if ndim is not None:
+            raise ValueError("Provided an ndim argument to "
+                             "MRG_RandomStreams.multinomial, which does not use "
+                             "the ndim argument.")
+        if pvals.ndim == 2:
+                size = pvals[:,0].shape
```
carriepl: That doesn't seem right, doesn't using this size completely disregard the 'n' parameter?

theano/sandbox/rng_mrg.py
```diff
@@ -19,7 +19,7 @@
 from theano.tensor import sqrt, log, sin, cos, join, prod
 from theano.compile import optdb
 from theano.gof import local_optimizer
-
+from theano.scalar import constant
```
carriepl: Where is this import used in your changes?

theano/sandbox/tests/test_multinomial.py
```diff
@@ -29,18 +30,58 @@ def run_with_c(f, gpu=False):
...
     # This tests the MultinomialFromUniform Op directly, not going through the
     # multinomial() call in GPU random generation.
 
     p = tensor.fmatrix()
     u = tensor.fvector()
-
-    m = multinomial.MultinomialFromUniform('auto')(p, u)
+    n = constant(1)
+    
+    m = multinomial.MultinomialFromUniform('auto')(p, u, n)
```
carriepl: Doesn't it work if you simply put `1` instead of `n`? The op should support this.

carriepl:
You have tests for MultinomialFromUniform directly but you also need tests that go through the method multinomial to ensure the way it handles the arguments is valid. Once of the issues I raised in a previous comment would have been caught by such tests.

carriepl:
Also, last week you mentionned your tests we failing. What did you since change to make them pass?

theano/sandbox/rng_mrg.py
aalmah: yeah I forgot to do it before I made the PR :)

aalmah:
There was a small bug in my C code. That was making it fail.

aalmah:
@carriepl Can you check it now. Everything seems to pass.

carriepl:
Your modified/added tests have asserts to ensure that the sampled elements kinda follow the expected probability distribution. Did you test with different random seeds to see how often thoses tests would fail due to simple chance?

Testing stochastic code is difficult... on one hand, you want to be able to validate the results... on the other hand, it's stochastic so anything is possible.

@nouiz, do you have any advice for this? Maybe use the AttemptManyTimes decorator (ex : https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/tests/test_mlp.py#L130)?

aalmah:
The thing is, I didn't find that other stochastic code (e.g. same multinomial but for n=1, or shared random streams multinomial) is tested more than this.. We might as well reconsider tests for other ops..

carriepl:
After IRL discussion with Fred and Pascal, their take on this is : if you set the rtol such that the test will work for 1000 different random seeds. Then the probability of the test failing should be low enough to be acceptable.

theano/sandbox/tests/test_multinomial.py
```diff
@@ -143,7 +182,7 @@ def test_gpu_opt():
 
     # Test with a row, it was failing in the past.
     r = tensor.frow()
-    m = multinomial.MultinomialFromUniform('auto')(r, u)
+    m = multinomial.MultinomialFromUniform('auto')(r, u, n)
```
nouiz: where is n defined?

theano/sandbox/multinomial.py
```diff
@@ -33,7 +35,7 @@ def __setstate__(self, dct):
         except AttributeError:
             self.odtype = 'auto'
 
-    def make_node(self, pvals, unis):
+    def make_node(self, pvals, unis, n):
```
nouiz: to don't break old code use `n=1` to be compatible with old code.

theano/sandbox/rng_mrg.py
```diff
@@ -1312,11 +1311,11 @@ def binomial(self, size=None, n=1, p=0.5, ndim=None, dtype='int64',
     def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
                     nstreams=None):
         """
-        Sample `n` (currently `n` needs to be 1) times from a multinomial
+        Sample `n` (`n` needs to be >= 1) times from a multinomial
         distribution defined by probabilities pvals.
 
-        Example : pvals = [[.98, .01, .01], [.01, .98, .01]] will
-        probably result in [[1,0,0],[0,1,0]].
+        Example : pvals = [[.98, .01, .01], [.01, .49, .50]] and n=2 will
```
nouiz: I would keep both example.

theano/sandbox/multinomial.py
```diff
@@ -136,12 +151,12 @@ def c_code(self, node, name, ins, outs, sub):
         """ % locals()
 
     def perform(self, node, ins, outs):
-        (pvals, unis) = ins
+        (pvals, unis, n_samples) = ins
```
nouiz: To support old pickled graph
```
 if len(ins) == 2:
  (pvals, unis) = ins
    make n_samples = 1
else
 ... 
```

theano/sandbox/multinomial.py
```diff
@@ -45,18 +47,18 @@ def make_node(self, pvals, unis):
...
         (gz,) = outgrads
         return [T.zeros_like(x) for x in ins]
 
     def c_code_cache_version(self):
-        return (5,)
+        return (6,)
 
     def c_code(self, node, name, ins, outs, sub):
-        (pvals, unis) = ins
+        (pvals, unis, n) = ins
```
nouiz: As n is a scalar.Scalar and not a tensor.scalar, to support old pickled graph, doing the same as in perform() should work. Just test it.

nouiz:
For the tests, currently he use a fixed test. I think this is better then doing what we planned as it won't randomly crash a few times a year. So I would let this like that with a fixed seed.

carriepl:
Any progress on this?

aalmah:
@carriepl just did push now

carriepl:
@nouiz , can you advise?

theano/sandbox/tests/test_multinomial.py
```diff
@@ -29,18 +32,80 @@ def run_with_c(f, gpu=False):
...
+    X = T.matrix('X')
+    pvals = T.exp(X)
+    pvals = pvals / pvals.sum(axis=1, keepdims=True)
+    samples = th_rng.multinomial(pvals=pvals)
+    pickle.dump([X, samples], open("multinomial_test_graph.pkl", "w"))
+    """
+    folder = os.path.dirname(os.path.abspath(__file__))
+    with open(os.path.join(folder, "multinomial_test_graph.pkl")) as pkl_file:
+        if PY3:
+            u = CompatUnpickler(pkl_file, encoding="latin1")
```
nouiz: The error is strange. Maybe you pickled with a recent enough numpy version that don't need that work around? If you use the normal pickle with that file, do this work with python 3?

carriepl: @nouiz : It doesn't work, this is the reason he tried the CompatUnpickler.

aalmah:
@carriepl This is now passing all checks. Can you please take a final look at it?

theano/sandbox/multinomial.py
```diff
@@ -346,7 +377,12 @@ def c_code(self, node, name, ins, outs, sub):
 @local_optimizer([MultinomialFromUniform])
 def local_gpu_multinomial(node):
     if type(node.op) is MultinomialFromUniform:
-        p, u = node.inputs
+        p, u, n_samples = node.inputs
```
nouiz: If we unpickle an old op, there will be only 2 inputs. Make sure to don't crash here.

theano/sandbox/multinomial.py
```diff
@@ -354,16 +390,21 @@ def local_gpu_multinomial(node):
...
             gpu_op = GpuMultinomialFromUniform(node.op.odtype)
             return [host_from_gpu(gpu_op(*[gpu_from_host(i)
-                                           for i in node.inputs])).T]
+                                           for i in [p, u]])).T]
     if (isinstance(node.op, theano.sandbox.cuda.GpuFromHost) and
             node.inputs[0].owner and
             type(node.inputs[0].owner.op) is MultinomialFromUniform):
         multi = node.inputs[0].owner
-        p, u = multi.inputs
+        p, u, n_samples = multi.inputs
```
nouiz: idem

nouiz:
I did just a quick high-level overview without checking the tests or the computation. I had one small comments.

aalmah:
@nouiz Just added those changes

carriepl:
Looks good so I'm merging. Thanks a lot.

nouiz:
I think this need documentation.

aalmah:
@nouiz what are you referring to?

theano/sandbox/multinomial.py
```diff
@@ -149,18 +174,24 @@ def perform(self, node, ins, outs):
...
-                if (waiting and (cummul > unis_n)):
-                    z[0][n, m] = 1
-                    waiting = False
-                else:
-                    z[0][n, m] = 0
+        for c in range(n_samples):
+            for n in range(nb_multi):
+                waiting = True
+                cummul = 0
+                unis_n = unis[n]
```
aalmah: @carriepl This is a bug. It should be unis[c*nb_multi + n]. Should I create a new pull request, or can we re-open this one?

theano/sandbox/multinomial.py
```diff
@@ -45,18 +47,23 @@ def make_node(self, pvals, unis):
         else:
             odtype = self.odtype
         out = T.tensor(dtype=odtype, broadcastable=pvals.type.broadcastable)
-        return Apply(self, [pvals, unis], [out])
+        return Apply(self, [pvals, unis, as_scalar(n)], [out])
 
     def grad(self, ins, outgrads):
-        pvals, unis = ins
+        pvals, unis, n = ins
```
aalmah: @nouiz Don't we need also to handle here old graphs that have two inputs only?

theano/sandbox/multinomial.py
nouiz: We can't reopen a PR, so make a new one and add test.

theano/sandbox/multinomial.py
nouiz: yes. good catch.

nouiz:
For the doc, I only found this, is it up to date:

http://deeplearning.net/software/theano/library/sandbox/rng_mrg.html?highlight=multinomial#theano.sandbox.rng_mrg.MRG_RandomStreams.multinomial

?

PR #3511, opened 2015-10-12
Remove useless reshape

### LINKED ISSUE #2752: Opt: Reshape(arange(x)) -> arange(x) when this do not change the shapes, number of dimensions.

yingzha:
Fixes #2752

### DIFF
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -3135,7 +3135,7 @@ def test_local_fill_useless():
     assert T.Alloc in ops
     f(m_, x_)
 
-    
+
 class Test_local_useless_elemwise_comparison(unittest.TestCase):
     def test_local_useless_elemwise_comparison(self):
         # TODO: test each case individually.
@@ -3171,7 +3171,7 @@ class Test_local_useless_elemwise_comparison(unittest.TestCase):
          >Sum{acc_dtype=float64} [@L] ''
          > |X[t] [@M] -> [@I]
         """
-        
+
         mode = theano.compile.get_default_mode().excluding('fusion')
         f = theano.function([X, Y], Z, mode=mode)
         theano.printing.debugprint(f, print_type=True)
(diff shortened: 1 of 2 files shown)


theano/tensor/tests/test_opt.py
```diff
@@ -5719,6 +5719,16 @@ def test_local_reshape(self):
         assert sum(isinstance(node.op, self.op) for node in topo) == 1
 
 
+def test_local_useless_reshape():
+    mode = theano.compile.get_default_mode().including(
+            'local_useless_reshape')
+    i = T.iscalar('i')
+    m=theano.tensor.mgrid[0:i,]
```
abergeron: Small nit: there should be spaces around the `=`

yingzha: Oh, I forgot it.

abergeron:
You should probably check the broadcastable pattern of the input and output and leave the reshape if it's not the same.  This causes problems in some cases otherwise.

nouiz:
I think if the broadcast pattern change, then we need to introduce a Rebroadcast op. This can be done easily by using the method patternbroadcast()

yingzha:
@abergeron , I just added a judgement to compare the broadcastable pattern, I'm not sure if it's wise nor not to do so.

abergeron:
Code is fine, all that is left to do is to remove the merge commits.

yingzha:
@abergeron done.

abergeron:

PR #3513, opened 2015-10-13
conv2d support uint

### LINKED ISSUE #3329: theano.tensor.nnet.conv.conv2d don't support uint
Why is it the case that in the conv2d function, line 127 the following assert is there:

assert str(image_shape[i].dtype).startswith('int')

this does not allow uint shape and also allows negative shapes (if not checked before), right? Is there a reason you guys did startswith() and not something like 'int' in str(image_shape[i].dtype)?

mronian:
Modified check for datatype of image and filter shape. Now it allows for int and uint both.

fix gh-3329

### DIFF
--- a/theano/tensor/nnet/conv.py
+++ b/theano/tensor/nnet/conv.py
@@ -105,7 +105,7 @@ def conv2d(input, filters, image_shape=None, filter_shape=None,
                         " information are constant values. We got"
                         " %s for the image_shape parameter" %
                         image_shape[i])
-                assert str(image_shape[i].dtype).startswith('int')
+                assert "int" in str(image_shape[i].dtype)
                 image_shape[i] = int(image_shape[i])
     if filter_shape is not None:
         filter_shape = list(filter_shape)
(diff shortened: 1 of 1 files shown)


nouiz:
thanks. Can you add a new tests that will tests with the uint? We need the test in Theano.

lamblin:
Thanks for the PR. In addition to the test requested by @nouiz, could you also:
- rebase on top of the 'master' branch, rather than merging with it, and
- use `theano.tensor.discrete_dtypes`, rather than string matching? Ideally, we should have changed that everywhere already, but since you are changing those lines anyway... For instance `assert image_shape[i].dtype in discrete_dtypes`

mronian:
Will do so. I am looking up the docs for how to write tests for Theano since I am new to this library.

nouiz:
I'll merge, as the code seem good, but having the test would be good.

mronian:
I am writing the tests for this. As I understand, the tests required are : 
self.assertRaises(AssertionError, self.validate, (2.2, 3.4, 4.6, 5.4), (2, 3, 2, 2), 'valid', verify_grad=False)

self.assertRaises(AssertionError, self.validate, (2, 3, 4, 5), (2.4, 3.5, 2.6, 2.6), 'valid', verify_grad=False)

Anything else I should add?

mronian:
I also changed the check for datatype to use theano.tensor.discrete_dtypes.

nouiz:
I'm not sure I understand correctly what you wrote. Can you make a new PR with what you have?

To be sure your tests do what is needed, they must fail without the fix in this PR.

mronian:
I misunderstood the tests which were required. Instead of tests for uint I was testing for discrete datatypes which was already being checked. I've added tests for uint now. I am submitting a PR(#3784) with those.

PR #3515, opened 2015-10-14
Changed printf with %zd specifier to %llu for portability (MSVC2013)

### LINKED ISSUE #3514: Usage of printf %zd on Visual Studio 2013 and earlier
The file `cuda_ndarray.cu` uses `fprintf(stderr, ...)` with the `%zd` format specifier in several places to display error messages related to memory allocations.

Unfortunately `%z` isn't supported on Visual Studio 2013 and earlier, and causes those and subsequent values not to be printed properly. E.g. messages look something like this on my machine 
```
Error allocating zd bytes of device memory (gßf=-ÇB>ñ+F+L!_+Pt+=x?n;+ p+çM&?¦+ZÇ
'+ëëZ+¦q~=Vë?+¦s++ß?=+HU-+áë(;?=
=?¦<+n+?+å1++Y`1+»-P+t÷?=Üb-++¦à+æ
*E+·Öí+(M?+F=X=?}¶>?¦ÿ+,e). Driver report zd bytes free and zd bytes total
```

The corresponding format specifier would be `%I` (i.e. `%Id` or `%Iu`). Visual Studio 2015 is supposed to support `%z`, but last time I checked it was not yet supported by Cuda Toolkit.

Maybe one solution could be to add a helper function which converts the numerical value in bytes to a more human readable string in bytes, kilobytes, megabytes, gigabytes, .., depending on the value's magnitude and just use `%s` in the final message.

wuaalb:
Changed all `printf()` calls formatting `size_t` variables with the `%zd` format specifier to `%llu` and corresponding `unsigned long long` casts, for compatibility with Visual Studio 2013 and before.
In some of these strings `%d` was used to print `size_t`; those were also changed.

Fixes #3514 on my machine when one of the error messages was triggered (have not tried other compilers).

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -142,8 +142,8 @@ void * device_malloc(size_t size, int verbose)
         status = cnmemMalloc(&rval, size, NULL);
         if(status != CNMEM_STATUS_SUCCESS) {
             PyErr_Format(PyExc_MemoryError,
-                         "Error allocating %zd bytes of device memory (%s).",
-                         size, cnmemGetErrorString(status));
+                         "Error allocating %llu bytes of device memory (%s).",
+                         (unsigned long long)size, cnmemGetErrorString(status));
             return NULL;
         }
     }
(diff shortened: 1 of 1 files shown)


nouiz:
I'm ok to merge when travis pass. thanks.

PR #3516, opened 2015-10-14
Various scan fixes and speedups

carriepl:


nouiz:
I'm good with the PR. We can merge when travis pass.

For history, can you write here the small speed up you got with the combination of all those change?

carriepl:
The compilation time in Daniel's use case dropped from ~1930s to ~1880s on my machine.

PR #3519, opened 2015-10-15
Add imports to keep old sandbox interface

bouthilx:
According to [lucasb-eyer](https://github.com/Theano/Theano/pull/3311#issuecomment-148326471), the PR #3311 broke the interface of ```sandbox``` by adding ```__init__.py``` which turned it into a python module. 

>I believe adding the ```__init__.py``` file in the sandbox folder as part of 5b604a4 has the side-effect of breaking existing code that directly used anything other than sandbox.opt, such as writing ```theano.sandbox.rng_mrg.MRG_RandomStreams()``` . I don't think it's too bad, just putting the information out there.

I added imports as in ```theano/tensor/__init__.py``` but I'm worried some of ```sandbox```'s modules were not built to be imported automatically. Would it introduce some bugs or a useless and significant import overhead?

So far I only tested with ```device=cpu``` and only problem I see is this deprecation warning
```python 
DEPRECATION: theano.sandbox.conv no longer provides conv.  They have been moved to theano.tensor.nnet.conv
```
Because ```theano.tensor.nnet.conv``` was not automatically imported previously. We would soon have the same problem with ```blocksparse``` once #3361 is merged.


nouiz:
we should not import everything by default.

bouthilx:
How would you then keep both old interface and automatic import of ```sandbox.opt```? I'd like not to break it if possible.

abergeron:
What do you mean adding __init__.py?  It has always been there otherwise nothing could be imported from sandbox.  It was just empty before.

Also
```
import theano
theano.sandbox.rng_mrg.RandomStreams()
```
Has never worked before.  You have always needed to add `import theano.sandbox.rng_mrg`
somewhere.

So I am against this change since it has never worked like that.

For reference see the file's history:
https://github.com/Theano/Theano/commits/aeb8c035895cf13df42e125fbf784d9e7a913e14/theano/sandbox/__init__.py

bouthilx:
Oh right. I guess @lucasb-eyer's code was working previously because he imported rng_mrg somewhere and removed that import. Thanks @abergeron!

lucasb-eyer:
I just tried with an older Theano and you are right. Sorry for the wasted time!

PR #3520, opened 2015-10-16
Add optdb.DB __contains__ operator

julienr:
This adds a `__contains__` operator to DB. Since DB is supposed to act like a dict, I wanted to be able to use `in` to test if an optimizer is already registered.


nouiz:
Good idea. Can you add a test in the file gof/tests/test_optdb.py to be sure it don't get broken in the futur?

julienr:
@nouiz done

nouiz:
thanks. We can merge when travis pass. Just to let you know, we prefer to not have merge commit in a PR, it make reading the history harder, but as this PR is simple, it is fine.

thanks

julienr:
Ok, thanks for the review and comment. I missed the merge commit, will keep that in mind for future contributions

PR #3521, opened 2015-10-16
Scan replace Allocs with AllocEmpty

carriepl:


theano/tensor/basic.py
```diff
@@ -6093,3 +6093,12 @@ def c_code_cache_version(self):
...
         return False
+
+    def connection_pattern(self, node):
+        return [[False] for i in node.inputs]
+
+    def grad(self, inputs, grads):
+        return [DisconnectedType()() for i in inputs]
+
+    def R_op(self, inputs, eval_points):
+        return [zeros(inputs, self.dtype)]
```
carriepl: @lamblin : I know the gradients have restrictions on their dtypes (they can't be int, if I recall correctly). Do the outputs of the R_op function have the same restrictions? If yes, I will need to change this implementation which would be incorrect if AllocEmpty was used with an integer dtype.

nouiz: I can't comment on that. Can @lamblin  or @abergeron confirm this is good?

lamblin: @carriepl I don't think that this restriction exists, or if it does in theory, then it has never been a problem that [`Alloc.R_op`](https://github.com/carriepl/Theano/blob/scan_speedup_cgt/theano/tensor/basic.py#L2805-L2808) does not follow it.
Btw, looking at `Alloc.R_op`, I would be tempted to just return `[None]` for `AllocEmpty.R_op`, since `AllocEmpty` has no differentiable input, at least if it works. If it does not work, then we should probably fix `Alloc.R_op` to return zeros instead of None in the first case.

carriepl: I already tried returning [None] (like Alloc does) but I got numerical errors in test_scan.py:T_Scan.test_rop2 and I did not understand why. I haven't tried modifying Alloc.R_op, though.

lamblin: This is probably because `IncSubtensor.R_op` returns `None` if _either_ differentiable input is `None`, whereas it should only do so if _both_ are.
Since there may be other Ops doing the same error, I think we should be on the side of correct results, rather than maximum speed. So I agree with your current implementation.

nouiz: There is few ops that implement rop. I think it would be quick to check if
there is other op that don't implement it correctly. Would it be useful?

lamblin: For the moment, I don't think it is a priority, and it would still be a bit of work to make sure the fix is correct, and to add the required tests.

nouiz: ok. @carriepl can you make an issue about this?

carriepl: It's created.

nouiz:
I'm ok with this, but IRL discussion with @lamblin he taught that we should run some real experiments that use big memory, just to be sure this don't raise the memory usage. Maybe scan save mem optimization would not always apply now.

carriepl:
@nouiz and @lamblin : according to the test I ran based on the LSTM tutorial, the memory usage is the same (the difference is less than 1 meg on a total memory usage of 1972 megs ) after and before my modifications.

nouiz:
ok. thanks I merge.

PR #3522, opened 2015-10-16
Add warning if profile_optimizer=True but profile=False

julienr:
It took me some time to figure out why `config.profile_optimizer` wasn't working (config.profile was False). This adds a small note about this in the optimization doc page and shows a warning if the user forgets to set config.profile to True as well.


nouiz:
thanks. Ok to merge when travis pass.

PR #3524, opened 2015-10-16
Scan inplace opt

carriepl:
Modify ScanInplaceOptimizer so Scan can be inplace on more of its inputs (instead of copying the contents of said inputs at execution time).


nouiz:
This seem ok. Do this change optimization time or execution time for some cases that you saw? Or this is just potential optimization that you can't time right now?

carriepl:
This is a potential optimization. It *should* help but I did not yet time it.

PR #3532, opened 2015-10-19
[CRASH] when Elemwise have more then 31 inputs

nouiz:


twiecki:
That PR fixes the pymc3 issue for me.

PR #3533, opened 2015-10-19
Multi gpu followup

abergeron:
Some interface changes to allow for easier use of the multi-gpu and some small bug fixes for unusual circumstances.

### DIFF
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -427,7 +427,8 @@ TensorVariable
     you'll want to call.
 
 
-.. class:: _tensor_py_operators(object)
+.. autoclass:: _tensor_py_operators
+   :members:
 
     This mix-in class adds convenient attributes, methods, and support
     to TensorVariable, TensorConstant and TensorSharedVariable for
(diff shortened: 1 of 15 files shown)


nouiz:
travis is failing.

doc/library/tensor/basic.txt
```diff
@@ -427,7 +427,8 @@ TensorVariable
     you'll want to call.
 
 
-.. class:: _tensor_py_operators(object)
+.. autoclass:: _tensor_py_operators
+   :members:
```
nouiz: Why this change? I don't see a change in the doc generated. I don't have objection to this change.

abergeron: This should add the documentation of transfer() which is only in its docstring.  We can move the rest later.

nouiz: ok

abergeron:
I've restarted the build because it seems to have failed due to a problem with the installation of six.

nouiz:
I have started the tests on my machine for sandbox/gpuarray. If they pass, I'll merge.

PR #3534, opened 2015-10-20
[WIP] Add R_op for sparse.basic.dot

chaitan3:
Just added a R_op function for sparse.basic.Dot Op. Should be functional. Still have to add a test for the R_op in sparse/tests/test_basic.py


lamblin:
Given the current state of things, the best way would be to treat None as zeros, and only return None if all the eval points are None.

PR #3536, opened 2015-10-20
[ENH] Speed up bn high_mem, infer_shape, trac more stack trace.

nouiz:


theano/configdefaults.py
```diff
@@ -446,8 +446,12 @@ def safe_no_home(home):
     "The number of stack to trace. -1 mean all.",
     # We default to 6 to be able to know where v1 + v2 is created in the
     # user script. The bigger this number is, the more run time it takes.
-    # We need to default to 7 to support theano.tensor.tensor(...).
-    IntParam(7),
+    # We need to default to 8 to support theano.tensor.tensor(...).
```
lamblin: Update the comment, to remove "We default to 6"...

theano/configdefaults.py
```diff
@@ -446,8 +446,12 @@ def safe_no_home(home):
...
+    # import theano, numpy
+    # X = theano.tensor.matrix()
+    # y = X.reshape((5,3,1))
+    # assert y.tag.trace
```
lamblin: Why is that code in comment here?

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -550,9 +547,9 @@ def test_softmax_grad_optimizations(self):
                 has_softmax = True
             if node.op == softmax_grad:
                 has_softmaxdx = True
-        assert has_cx1hot
+        assert not has_cx1hot
         assert has_cx1hotdx
-        assert not has_softmax
+        assert has_softmax
```
lamblin: That looks like a regression. Why is a softmax still here?

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -592,9 +587,9 @@ def test_softmax_grad_optimizations_vector(self):
...
```
lamblin: Idem here.

theano/configdefaults.py
nouiz: To give the case where 8 is used.

theano/tensor/nnet/tests/test_nnet.py
nouiz: We change a crossentropy_softmax_argmax_1hot_with_bias to just a softmax. So we replace a complicated op with multiple output where we use only 1 output by a simpler op that do just the part we do. We where already taking the output that compute the softmax.

Idem for the other case.

nouiz:
I updated the comment. The failed travis is strange...

theano/tensor/nnet/bn.py
```diff
@@ -66,8 +66,7 @@ def batch_normalization(inputs, gamma, beta, mean, std,
         elm_bn = theano.tensor.elemwise.Elemwise(scalar_op=BNComposite(dtype=inputs.dtype))
         rval = elm_bn(inputs, mean, std, gamma, beta)
     elif mode == 'high_mem':
-        rval = (inputs - mean) / std
-        rval = rval * gamma + beta
+        rval = (inputs - mean) * (gamma / std) + beta
```
f0k: :+1:

nouiz:
I restarted travis and it passed today. Probably due to the problem they got yesterday.

nouiz:
@lamblin when you have time, can you check again this PR? I think it is ready to merge, but maybe you won't agree with my reply to your comment.

PR #3539, opened 2015-10-21
changing `assert` to `AssertionError`

jotterbach:
This `assert` statement does not provide any error message making debugging this case hard. Moreover, `assert` is eliminated when using optimized python code (https://docs.python.org/2.7/reference/simple_stmts.html#the-assert-statement), introducing a potential bug into the system. Explicitly checking and raising an AssertionError addresses both issues.


nouiz:
thanks. Do you use optimized python without assert? If so, what speed up do you see?

PR #3548, opened 2015-10-23
[ENH,REGRESSION FIX] Fix regression that introduce extra node related to infer shape

nouiz:
This should fix https://github.com/fchollet/keras/issues/866#issuecomment-150024775

This is a forward looking change that try to fix the fondamental issue. We need to add recursion that that fix for a complete fix and need to have other optimization call get_shape instead of using shape_of direction. I let this to a later PR to fix quickly the regression. The full fix will also need optimization speed testing to not have it slow down too much compilation.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -955,6 +955,34 @@ class ShapeFeature(object):
 
         return o_shapes
 
+    def get_shape(self, var, idx):
+        """ Optimization can call this to get the current shape_i
+
+        It is better to call this then use directly shape_of[var][idx]
+        as this method should update shape_of if needed.
+
+        TODO: Up to now, we don't update it in all cases. Update in all cases.
+        """
+        r = self.shape_of[var][idx]
+        if (r.owner and
+                isinstance(r.owner.op, Shape_i) and
+                r.owner.inputs[0] not in var.fgraph.variables):
+            assert var.owner
+            inp = var
+            node = inp.owner
+# TODO recur on inputs
+# Need to time this to don't have it too slow.
+# Make sure to handle the case of (shape_i(x)+1)
+#            for v in node.inputs:
+#                for idx in range(v.ndim):
+#                    self.get_shape(v, idx)
+            o_shapes = self.get_node_infer_shape(node)
+            assert len(o_shapes) == len(node.outputs)
+            for shps, out in zip(o_shapes, node.outputs):
+                self.set_shape(out, shps, override=True)
+            r = o_shapes[node.outputs.index(inp)][r.owner.op.i]
+        return r
+
     def shape_ir(self, i, r):
         """Return symbolic r.shape[i] for tensor variable r, int i."""
         if hasattr(r.type, "broadcastable") and r.type.broadcastable[i]:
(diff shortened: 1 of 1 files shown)


nouiz:
I timmed the optimization time and this didn't slow down the case the regression fixed: gh-3412. This do not slow down the GSoC case.

lamblin:
I did not review it yet, but there seems to be an issue in some tests of scan.

nouiz:
I amended the last commit. The difference is in local_elemwise_alloc. I didn't used the right idx.

theano/tensor/opt.py
```diff
@@ -921,6 +921,67 @@ class ShapeFeature(object):
...
+                r.owner.inputs[0] not in var.fgraph.variables):
+            assert var.owner
+            inp = var
+            node = inp.owner
+# TODO recur on inputs
+# Need to time this to don't have it too slow.
+# Make sure to handle the case of (shape_i(x)+1)
+#            for v in node.inputs:
+#                for idx in range(v.ndim):
+#                    self.get_shape(v, idx)
```
lamblin: Put that in a ticket, and reference the ticket from here.

theano/tensor/opt.py
```diff
@@ -1017,7 +1078,7 @@ def unpack(self, s_i):
             raise TypeError('Unsupported shape element',
                             s_i, type(s_i), getattr(s_i, 'type', None))
 
-    def set_shape(self, r, s):
+    def set_shape(self, r, s, override=False):
```
lamblin: Add that new parameter in the docstring, and document it.

theano/tensor/opt.py
```diff
@@ -921,6 +921,67 @@ class ShapeFeature(object):
...
+            node = inp.owner
+# TODO recur on inputs
+# Need to time this to don't have it too slow.
+# Make sure to handle the case of (shape_i(x)+1)
+#            for v in node.inputs:
+#                for idx in range(v.ndim):
+#                    self.get_shape(v, idx)
+            o_shapes = self.get_node_infer_shape(node)
+            assert len(o_shapes) == len(node.outputs)
+            for shps, out in zip(o_shapes, node.outputs):
```
lamblin: I don't think we should override the shape for _all_ outputs of `node`, I think we should only do it for `idx`.

lamblin: I mean, only override it for `out = node.outputs.index(inp)`, and keep the initial `shps[_i]` for `_i != idx`.

nouiz: If we do it only for that var and that index, as we will call get_shape on the other idx, it would trigger duplicate computation.

But I think there is a middle ground that is better: for all output variable and for all index, only update those that need update (the same condition as line 968 and 969).

lamblin:
Right, that makes sense.
Thanks!

nouiz:
Travis passed. Can you merge?

lamblin:
I'm still running the tests on GPU to make sure there is no issue there. I'll try to merge tonight.

carriepl:
@nouiz, @lamblin : Considering the number of times an issue of this nature (a lot more Scans in the compiled graph than there should be) has popped up, could you add the test case you used in a follow-up PR?

PR #3549, opened 2015-10-23
Fix infer_shape of Join when axis < 0

lamblin:
Also add tests


lamblin:
Updated with a comment.

PR #3550, opened 2015-10-24
Fix unpickling of Gemm and GpuGemm Ops

### LINKED ISSUE #3535: Attribute Error Loading pickled Functions using GpuGemm
While trying to load a pickled theano function I get an AttributeError (traceback printed below). I'm using HEAD. The following patch fixes this problem (although I doubt it's the correct fix):
```
diff --git a/theano/sandbox/cuda/blas.py b/theano/sandbox/cuda/blas.py
index b109e39..86c8322 100644
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -400,6 +400,7 @@ class GpuGemm(GpuOp):
     implement the gemm on the gpu.
 
     """
+    _op_use_c_code = theano.config.cxx
     def __init__(self, inplace):
         self.__setstate__({'inpfull lace': inplace})
```
Traceback:
```
  File "/users/nboley/src/TF_binding/rSeqDNN/KerasModel.py", line 115, in compile
    self.model = pickle.load(fp)
  File "/users/nboley/src/Theano/theano/compile/function_module.py", line 1018, in _constructor_Function
    f = maker.create(input_storage, trustme=True)
  File "/users/nboley/src/Theano/theano/compile/function_module.py", line 1642, in create
  File "/users/nboley/src/Theano/theano/gof/link.py", line 690, in make_thunk
    storage_map=storage_map)[:3]
  File "/users/nboley/src/Theano/theano/gof/vm.py", line 1037, in make_all
    no_recycling))
  File "/users/nboley/src/Theano/theano/sandbox/cuda/__init__.py", line 284, in make_thunk
    compute_map, no_recycling)
  File "/users/nboley/src/Theano/theano/gof/op.py", line 915, in make_thunk
    if self._op_use_c_code:
AttributeError: (AttributeError('The following error happened while compiling the node', GpuGemm{no_inplace}(<CudaNdarrayType(float32, matrix)>, TensorConstant{0.20000000298}, GpuElemwise{mul,no_inplace}.0
, <CudaNdarrayType(float32, matrix)>, TensorConstant{0.20000000298}), '\n', "'GpuGemm' object has no attribute '_op_use_c_code'"), <function _constructor_Function at 0x7f0febfa1cf8>, (<theano.compile.funct
ion_module.FunctionMaker object at 0x7f0f503aae50>, [<None>, <None>, <None>, <None>, <None>, <None>, <None>, <None>], [None, None, None, None, None, None, None, None]))
nboley@nandi:/srv/scratch/nboley/tmp/test_new_relaxed_peaks$ THEANO_FLAGS='mode=FAST_RUN,device=gpu2,floatX=float32,optimizer=fast_compile,warn_float64=warn,reoptimize_unpickled_function=False' python ~/sr
c/TF_binding/rSeqDNN/TrainModel.py --annotation-id 1 --tf-id T011266_1.02 --model-prefix MAX.tmp --only-test-one-fold --half-peak-width 500 --max-num-peaks-per-sample 500
 ```

### PR DESCRIPTION
**Pull request by lamblin**
Closes #3535.

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -401,7 +401,9 @@ class GpuGemm(GpuOp):
 
     """
     def __init__(self, inplace):
-        self.__setstate__({'inplace': inplace})
+        self.inplace = inplace
+        if self.inplace:
+            self.destroy_map = {0: [0]}
 
     def __str__(self):
         if self.inplace:
@@ -417,13 +419,14 @@ class GpuGemm(GpuOp):
         return hash(type(self)) ^ hash(self.inplace)
 
     def __setstate__(self, dct):
-        inplace = dct.get('inplace', True)
-        if inplace:
-            self.destroy_map = {0: [0]}
-        self.inplace = inplace
+        self.__dict__.update(dct)
 
-    def __getstate__(self):
-        return dict(inplace=self.inplace)
+        # Correctly reload older pickles where _op_use_c_code and
+        # destroy_map were not saved
+        if '_op_use_c_code' not in self.__dict__:
+            self._op_use_c_code = theano.config.cxx
+        if 'destroy_map' not in self.__dict__ and self.inplace:
+            self.destroy_map = {0: [0]}
 
     def make_node(self, z, a, x, y, b):
         # the more complicated error checking performed by tensor.gemm
@@ -518,7 +521,9 @@ class GpuGemv(GpuOp):
 
     """
     def __init__(self, inplace):
-        self.__setstate__({'inplace': inplace})
+        self.inplace = inplace
+        if self.inplace:
+            self.destroy_map = {0: [0]}
 
     def __str__(self):
         if self.inplace:
@@ -534,13 +539,14 @@ class GpuGemv(GpuOp):
         return hash(type(self)) ^ hash(self.inplace)
 
     def __setstate__(self, dct):
-        inplace = dct.get('inplace', True)
-        if inplace:
-            self.destroy_map = {0: [0]}
-        self.inplace = inplace
+        self.__dict__.update(dct)
 
-    def __getstate__(self):
-        return dict(inplace=self.inplace)
+        # Correctly reload older pickles where _op_use_c_code and
+        # destroy_map were not saved
+        if '_op_use_c_code' not in self.__dict__:
+            self._op_use_c_code = theano.config.cxx
+        if 'destroy_map' not in self.__dict__ and self.inplace:
+            self.destroy_map = {0: [0]}
 
     def make_node(self, z, a, x, y, b):
         # the more complicated error checking performed by tensor.gemv
@@ -615,7 +621,9 @@ class GpuGer(GpuOp):
 
     """
     def __init__(self, inplace):
-        self.__setstate__({'inplace': inplace})
+        self.inplace = inplace
+        if self.inplace:
+            self.destroy_map = {0: [0]}
 
     def __str__(self):
         if self.inplace:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: theano/sandbox/cuda/blas.py [hunk 1]
```diff
@@ -417,13 +417,19 @@ def __hash__(self):
...
+
+        # Correctly reload older pickles where _op_use_c_code was not saved
+        if '_op_use_c_code' not in self.__dict__:
+            self._op_use_c_code = True
 
     def __getstate__(self):
-        return dict(inplace=self.inplace)
+        rval = self.__dict__.copy()
+        # Do not serialize the destroy_map
+        rval.pop('destroy_map', None)
```
abergeron: Why don't we want to serialize destroy_map?  There shouldn't be any problems with pickle.

lamblin: Right, I'll change that for the sake of simplicity.

lamblin:
Updated, rebased, and squashed.

theano/sandbox/cuda/blas.py
```diff
@@ -417,13 +419,14 @@ def __hash__(self):
         return hash(type(self)) ^ hash(self.inplace)
 
     def __setstate__(self, dct):
-        inplace = dct.get('inplace', True)
-        if inplace:
-            self.destroy_map = {0: [0]}
-        self.inplace = inplace
+        self.__dict__.update(dct)
```
nouiz: I think we should keep the old code. Some old pickled file with this op would have only inplace in dct and would not have destroy_map. So add it back with a comment that we need to keep it to unpickled correctly old graph.

theano/sandbox/cuda/blas.py
```diff
@@ -417,13 +419,14 @@ def __hash__(self):
...
 
-    def __getstate__(self):
-        return dict(inplace=self.inplace)
+        # Correctly reload older pickles where _op_use_c_code and
+        # destroy_map were not saved
```
nouiz: this is getstate method. It get called when we pickled, not when we unpickle. Also this method need to return something and it don't.

theano/sandbox/cuda/blas.py
```diff
@@ -417,13 +419,14 @@ def __hash__(self):
...
+        if '_op_use_c_code' not in self.__dict__:
+            self._op_use_c_code = theano.config.cxx
```
nouiz: This would go in setstate. But this is an optional parameter. So I think we should fix where this is accessed.

theano/sandbox/cuda/blas.py
```diff
@@ -534,13 +539,14 @@ def __hash__(self):
         return hash(type(self)) ^ hash(self.inplace)
 
     def __setstate__(self, dct):
-        inplace = dct.get('inplace', True)
```
nouiz: idem for the 2 other op in that file.

theano/tensor/blas.py
```diff
@@ -962,7 +962,12 @@ class Gemm(GemmRelated):
     E_float = 'gemm requires floating-point dtypes'
 
     def __init__(self, inplace):
-        self.__setstate__({'inplace': inplace})
```
nouiz: I think this op is a good example to follow for the other op changed in this PR, except maybe for the _op_use_c_code.

theano/sandbox/cuda/blas.py
lamblin: Yes, this is what is done below.

theano/sandbox/cuda/blas.py
lamblin: I removed the `__getstate__` method, so we use the default one.
GitHub may display the diff in a confusing way, but that code is in `__setstate__`.

theano/sandbox/cuda/blas.py
lamblin: According to [the comments in `Op.__new__`](https://github.com/Theano/Theano/blob/master/theano/gof/op.py#L768-L774), this is not optional and all Ops should have it.

lamblin: In fact, all Ops have it at creation time, so I don't think they should discard it when pickling. If we silently set a default for Ops missing `_op_use_c_code`, we would not detect the issue and potentially keep doing something incorrect.

nouiz: ok. But I think it is not good to create the graph and hardcode in the the value of config.cxx. I think that by default, it should be None and that mean the current value of cxx. If it is set otherwise, use that value.

Anyway, that isn't a current concern. Continue as we do for consistency for now.

theano/sandbox/cuda/blas.py
nouiz: github diff is good, but strange. I didn't understand it correctly first time. It is good. I'm merging.

PR #3559, opened 2015-10-26
Multi gpu doc

abergeron:
This adds documentation for the new gpu backend to the library doc and add a tutorial section about how to use multi-gpu.

This also gets rid of some useless sections of the doc.


nouiz:
travis fail:
```
/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/gpuarray/type.py:519:17: E265 block comment should start with '# '

/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/gpuarray/type.py:520:17: E265 block comment should start with '# '

```

theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -26,6 +26,20 @@
 
 
 def as_gpuarray_variable(x, context_name):
+    """
+    This will attempt to convert `x` into a variable on the GPU.
+
+    It can take either a value of aother variable.  If `x` is already
```
nouiz: aother -> another

theano/sandbox/gpuarray/kernel_codegen.py
```diff
@@ -161,14 +170,10 @@ def inline_softmax(N, buf, buf2, threadPos, threadCount, dtype="float32"):
     dtype
         Dtype of the softmax's output.
 
-    :Precondition: buf and buf2 contain two identical copies of the input
-        to softmax
-    :Postcondition: buf contains the softmax, buf2 contains un-normalized
-        softmax
-
```
nouiz: why remove that information? Should it be moved to a note section?

theano/sandbox/gpuarray/kernel_codegen.py
```diff
@@ -230,20 +238,19 @@ def inline_reduce_fixed_shared(N, buf, x, stride_x, load_x, pos, count,
...
-        A function that accepts strings of arguments a and b, and returns c code
-        for their reduction.
-        Example: return "%(a)s + %(b)s" for a sum reduction.
-    manner_init
-        A function that accepts strings of arguments a and return c code for its
-        initialization.
+        A function that accepts strings of arguments a and b, and
+        returns c code for their reduction.
 
-    :postcondition:
```
nouiz: idem

theano/sandbox/gpuarray/kernel_codegen.py
```diff
@@ -352,13 +363,9 @@ def inline_softmax_fixed_shared(N, buf, x, stride_x, load_x,
     dtype
         Optional, the dtype of the softmax's output if not float32.
 
-    :Precondition: buf is empty
-    :Postcondition: buf[0] contains the softmax, buf2 contains un-normalized
```
nouiz: idem

theano/sandbox/gpuarray/type.py
```diff
@@ -95,6 +140,21 @@ def __init__(self, dtype, broadcastable, context_name=None, name=None):
                             (self.__class__.__name__, self.dtype))
 
     def clone(self, dtype=None, broadcastable=None):
+        """
```
nouiz: why document in the chield class all the method that are in the parent? I think it should be documented in the parent.

theano/sandbox/gpuarray/type.py
```diff
@@ -112,6 +177,31 @@ def __repr__(self):
                                              self.broadcastable)
 
     def filter(self, data, strict=False, allow_downcast=None):
+        """
```
nouiz: idem and at other places in this file.

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
+
+.. _using_multi_gpu:
```
nouiz: We make the tutorial section have a prefix, use "tut_using_multi_gpu"

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+referer to what we call context names.  These are then mapped to a
+device using the theano configuration.  This allows portability of
+models between machines.
+
+.. warning::
+
+   The code is rather new and is still considered experimental at this
+   point.  It has been tested and seems to perform correctly in all
+   cases observed, but make sure to double-check your results before
+   publishing a paper or anything of the sort.
```
nouiz: Also tell that currently there is extra sync that remove all speed up. It should be fixed in a few weeks.

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+    dev0->cuda0;dev1->cuda1
+
+Let's break it down.  First there is a list of mappings.  Each of
+these mappings is separeted by a semicolon ';'.  There can be any
+number of such mappings, but in the example above we have two of them:
+`dev0->cuda0` and `dev1->cuda1`.
+
+The mappings themselves are composed of a context name followed by the
+two characters '->' and the device name.  The context name is a simple
+string which oes not have any special meaning for Theano.  For parsing
```
nouiz: oes -> does

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+reasons, the context name cannot contain the sequence '->'.  The
```
nouiz: also, it can't start with some other restriction like cuda, ....

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+device name is a device in the form that gpuarray expects like 'cuda0'
+or 'opencl0:0'.
+
+If you don't have enough GPUs for a certain model, you can assign the
+same device to more than one name. You can also assign extra names
+that a model doesn't need to some other devices.  However, a
+proliferation of names is not always a good idea since theano often
+assumes that different context names will be on different devices and
+will optimize accordingly.  So you may get faster performance for a
+single name and a single device.
```
nouiz: How hard would it be to make sure we don't add extra transfer? Is there other case where this could cause slow down? Can you make a ticket to fix them and list the one you know. I think it is useful to allow using a multi-gpu program on a single GPU without slowdown and copy and extra memory usage.

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+number of such mappings, but in the example above we have two of them:
+`dev0->cuda0` and `dev1->cuda1`.
+
+The mappings themselves are composed of a context name followed by the
+two characters '->' and the device name.  The context name is a simple
+string which oes not have any special meaning for Theano.  For parsing
+reasons, the context name cannot contain the sequence '->'.  The
+device name is a device in the form that gpuarray expects like 'cuda0'
+or 'opencl0:0'.
+
```
nouiz: Add a note that if they define the flag in a shell, they most quote it:

THEANO_FLAGS="context=dev0->cuda0;dev1->cuda1"

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+The following simple program works on two GPUs.  It builds a function
+which perform two dot products on two different GPUs.
+
+.. code-block:: python
+
+   import numpy
+   import theano
+
+   v01 = theano.shared(numpy.random.random(1024, 1024).astype('float32'),
```
nouiz: random((1024, 1024)) for all cases.

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,127 @@
...
+                       target='dev0')
+   v11 = theano.shared(numpy.random.random(1024, 1024).astype('float32'),
+                       target='dev1')
+   v12 = theano.shared(numpy.random.random(1024, 1024).astype('float32'),
+                       target='dev1')
+
+   f = theano.function([], [theano.tensor.dot(v01, v02),
+                            theano.tensor.dot(v11, v12)])
+
+   f()
```
nouiz: Give an example of the output.

theano/sandbox/gpuarray/kernel_codegen.py
abergeron: The same information is also in the description above.

theano/sandbox/gpuarray/kernel_codegen.py
abergeron: Also in the description above.

theano/sandbox/gpuarray/kernel_codegen.py
abergeron: I've added the part about the postcondition, the other is already above.

theano/sandbox/gpuarray/type.py
abergeron: I agree, but the documentation does not propagate for reimplemented functions.  I could just drop the docstrings for those and sphinx show the base classes, however I personally dislike having to play a hide and seek with documentation just to find what a certain class can do.

We could however automate the transmission of docstrings for certain methods, which could alleviate this problem.

doc/tutorial/using_multi_gpu.txt
abergeron: I would say that this is very inconsistent since more than half of the files in tutorial/ don't have any sort of prefix and those that do don't all have the same form (it can be one of 'tut', 'tut_', 'tutorial', 'tutorial_') or have it somewhere in the middle.

Is this something we want to address?

doc/tutorial/using_multi_gpu.txt
abergeron: I do this later.  At line 92 to be precise.

doc/tutorial/using_multi_gpu.txt
abergeron: This is something that would be very hard (if not impossible) to fix if we want to keep support for pickled compiled functions.

The current setup favors multi-gpu performance at the cost of a potential slowdown (but mostly more memory) when on a single device.

I don't think we should spend any effort to fix that unless there is a clear use case which requires it.  The code *will* run on a single device as is.

doc/tutorial/using_multi_gpu.txt
abergeron: It doesn't have any output.

abergeron:
I've made the changes.

doc/tutorial/using_multi_gpu.txt
nouiz: That is too late. Move it here.

doc/tutorial/using_multi_gpu.txt
nouiz: we did a ticket for this: https://github.com/Theano/Theano/issues/3565

doc/tutorial/using_multi_gpu.txt
nouiz: I have this:

```
$ PYTHONPATH=.:$PYTHONPATH THEANO_FLAGS=floatX=float32,"contexts=dev0->cuda0;dev1->cuda1" python test_multi.py
Mapped name dev0 to device cuda0: GeForce GTX TITAN X
Mapped name dev1 to device cuda1: GeForce GTX TITAN X
```
This is to show what user should expect. Could you add timming to the example to show more?

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,137 @@
...
+   import numpy
+   import theano
+
+   v01 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
+                       target='dev0')
+   v02 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
+                       target='dev0')
+   v11 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
+                       target='dev1')
+   v12 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
```
nouiz: random((1024, 1024)) -> random(1024, 1024)

abergeron: No, this is the right way to make the call.

avostryakov:
1. How I understand for this multi-GPU I have to install libgpuarray library. Right?

2. If it is true, when I installed libgpuarray, I noticed that I have to install clBLAS and couldn't do it on Ubuntu. Do I need to install clBLAS for calculations on GPU too? If it so, can you add all of this information with installation libgpuarray and clBLAS in documentation?

doc/tutorial/using_multi_gpu.txt
nouiz: I needed to fix this in the file I created with this code. But my file currently have the same thing as this comment. Not sure what happened...

nouiz:
1) Yes
2) The instruction are here: http://deeplearning.net/software/libgpuarray/installation.html

clBLAS are optional and only needed for OpenCL. We are missing some important pieces before OpenCL is really usable. So for now, you should use CUDA.

Don't forget some of the warning, currently, there is no speed up. We know how to fix this, but it will probably take 2-3 weeks before being merged. But you can start using bigger model now.

doc/tutorial/using_multi_gpu.txt
abergeron: ok

doc/tutorial/using_multi_gpu.txt
abergeron: I changed this because @carriepl made a comment about it.

doc/tutorial/using_multi_gpu.txt
abergeron: For timings I would prefer to direct people to theano/misc/check_multi_gpu.py, since otherwise I feel it would complicate the code and make the example less clear.

abergeron: Also, for the printout I've added a section about what it should show in the "Defining a context map" section.

nouiz: that seem good.

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,149 @@
...
+these mappings is separeted by a semicolon ';'.  There can be any
+number of such mappings, but in the example above we have two of them:
+`dev0->cuda0` and `dev1->cuda1`.
+
+The mappings themselves are composed of a context name followed by the
+two characters '->' and the device name.  The context name is a simple
+string which does not have any special meaning for Theano.  For
+parsing reasons, the context name cannot contain the sequence '->' or
+';'.  To avoid confusion context names that begin with 'cuda' or
+'opencl' as disallowed.  The device name is a device in the form that
```
nouiz: as -> are

doc/tutorial/using_multi_gpu.txt
abergeron: Does this mean the current state is ok?

nouiz: for that point, yes.

nouiz:
I'm good to merge.

Just before, do we want to merge now as it won't have speed up and this could cause more questions/support request and make it harder for you to implement cnmem?

@lamblin what do you think?

nouiz:
closing/reopening to have travis run with current master. I think this would fix travis error.

lamblin:
That branch now needs a rebase.

abergeron:
I'll do it when we are ready to merge otherwise I'll have to do it a lot of times (probably).

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,149 @@
...
+Using multiple GPUs
+===================
+
+Theano has a feature to allow the use of multiple GPUs at the same
+time in one function.  The multiple gpu feature requires the use of
+the :ref:`gpuarray` backend, so make sure that works correctly.
+
+In order to keep a reasonably high level of abstraction you do not
+refer to device names directly for multiple-gpu use.  You instead
+referer to what we call context names.  These are then mapped to a
```
rizar: referer -> refer

doc/tutorial/using_multi_gpu.txt
```diff
@@ -0,0 +1,149 @@
...
+   point.  It has been tested and seems to perform correctly in all
+   cases observed, but make sure to double-check your results before
+   publishing a paper or anything of the sort.
+
+.. warning::
+
+   Due to some implementation issues, models using multiple GPUs will,
+   in most cases not exhibit any speedup over running on a single GPU.
+   These issues are being worked on and should be solved within 2-3
+   weeks.  We do not expect any interface change, so you can still
```
rizar: Is it still relevant?

nouiz: no. It wasn't updated, but this is fixed.

abergeron:
This should now be ready to merge.

abergeron:
Rebased again.  Please merge this unless there is a good reason not to.

avostryakov:
Does it mean an official support of a new multi-gpu GpuArray/libgpuarray backend? I mean, does it work correctly or you continue to tweak or improve it?

lamblin:
It is working correctly for a wide variety of tests, and we tried different models which seem to work correctly on it. However, it has not been tested on a large scale yet, so tweaks and improvement are certainly going to be necessary.

PR #3562, opened 2015-10-26
Small tutorial for op params

abergeron:
This renames op context to op params and adds a small tutorial for it.

I felt the rename was warranted beause it's not used much right now and params is a much better name for the new generic usage we are thinking about.

Testing with a couple of simple multi-gpu programs assures me that the rename went well.

### DIFF
--- a/doc/extending/using_params.txt
+++ b/doc/extending/using_params.txt
@@ -8,6 +8,10 @@ The Op params is a facility to pass some runtime parameters to the
 code of an op without modifying it.  It can enable a single instance
 of C code to serve different needs and therefore reduce compilation.
 
+The code enables you to pass a single object, but it can be a struct
+or python object with multiple values if you have more than one value
+to pass.
+
 We will first introduce the parts involved in actually using this
 functionality and then present a simple working example.
 
@@ -15,23 +19,19 @@ The params type
 ----------------
 
 You can either reuse an existing type such as :class:`Generic` or
-:class:`CDataType`, or create your own.
+create your own.
 
 Using a python object for your op parameters (:class:`Generic`) can be
 annoying to access from C code since you would have to go through the
 Python-C API for all accesses.
 
-Conversly, using a C struct (:class:`CDataType`) makes it complicated
-to set values through python code, forcing you to go through ctypes or
-other similar module.
-
 Making a purpose-built class may require more upfront work, but can
-pay off if you reuse the type for a lot of Ops, by making it easier to
-manipulate from python and C.
+pay off if you reuse the type for a lot of Ops, by not having to re-do
+all of the python manipulation.
 
 
 Defining a params type
-~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~
 
 .. note::
 
@@ -67,16 +67,16 @@ attribute :attr:`params_type` to an instance of your params Type.
 .. note::
 
    If you want to have multiple parameters you have to bundle those
-   inside a single object (or C struct) and use that as the params
-   type.  Multiple types are not supported.
+   inside a single object and use that as the params type.  Multiple
+   types are not supported.
 
-For example if we decide to use a pointer to an int as the params the
-following would be appropriate:
+For example if we decide to use an int as the params the following
+would be appropriate:
 
 .. code-block:: python
 
    class MyOp(Op):
-       params_type = CDataType('int')
+       params_type = Generic()
 
 After that you need to define a :meth:`get_params` method on your
 class with the following signature:
@@ -138,7 +138,7 @@ the params type.
    class MulOp(Op):
        params_type = Generic()
        __props__ = ('mul',)
-   
+
       def __init__(self, mul):
           self.mul = float(mul)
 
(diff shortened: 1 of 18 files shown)


nouiz:
@carriepl I let you do a review of this.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+===============
+Using Op params
+===============
+
+The Op params is a facility to pass some runtime parameters to the
+code of an op without modifying it.  It can enable a single instance
+of C code to serve different needs and therefore reduce compilation.
+
+I will first introduce the parts involved in actually using this
+functionality and then present a simple working example.
```
carriepl: This type of documentation should usually be impersonal and avoid 'I's.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+
+Additionaly if you want to use your params with C code, you need the
+following methods:
+
+  - :meth:`c_declare <CLinkerType.c_declare>`
+  - :meth:`c_init <CLinkerType.c_init>`
+  - :meth:`c_extract <CLinkerType.c_extract>`
+  - :meth:`c_cleanup <CLinkerType.c_cleanup>`
+
+You can also define other convinience methods such as
```
carriepl: convenience

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+:meth:`c_headers <CLinkerType.c_headers>` if you need any special things.
+
+
+Registering the params with your Op
+-----------------------------------
+
+To declare that your op uses params you have to set the class
+attribute :attr:`params_type` to an instance of your params Type.
```
carriepl: What if I want my op to have multiple params? Instinctively, I would set the attribute `params_type` to a list of instances but it's probably better to just be clear and  mention it.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+   class MyOp(Op):
+       params_type = CDataType('int')
+
+After that you need to define a :meth:`get_params` method on your
+class with the following signature:
+
+.. code-block:: python
+
+   def get_params(self, node):
+       pass
```
carriepl: Some people are going to read only half of what you wrote, see this and go 'Oh, my method just needs to exist and doesn't need to do anything'. Either give a fake method content consistent with your int example so far or really just show the signature of the method without any implementation.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+
+This methods must return a valid object for your Type (an object that
```
carriepl: method (no 's')

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+
+    Since this will change the expected signature of a few methods, it
+    is strongly discouraged to have your :meth:`get_params` method
+    return None.
+
+
+Signature changes from having params
+------------------------------------
+
+Having declared a params for you Op will affect the expected signature
```
carriepl: 'your Op'

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+of C code to serve different needs and therefore reduce compilation.
+
+I will first introduce the parts involved in actually using this
+functionality and then present a simple working example.
+
+The params type
+----------------
+
+You can either reuse an existing type such as :class:`Generic` or
+:class:`CDataType`, or create your own.
```
carriepl: I feel there should be a bit more information here to help a user reading the tutorial decide which of these 3 options is right for him/her.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,140 @@
...
+This is true for all methods that recieve a `sub` parameter, so this
+means that you can use your params in the :meth:`c_init_code_struct
+<Op.c_init_code_struct>` method.
+
+A simple example
+----------------
+
+This is a simple example which uses a params object to pass a value.
+The value in this case is a python float, hence the choice of Generic
+as the params type.
```
carriepl: State the functionality implemented by this example, either here or in a class comment.

theano/sandbox/rng_mrg.py
```diff
@@ -1014,7 +1014,7 @@ def c_code(self, node, nodename, inp, out, sub):
         """ % locals()
 
     def c_code_cache_version(self):
-        return (7, self.GpuKernelBase_version)
+        return (7,)
```
carriepl: Why is `self.GpuKernelBase_version` no longer required here?

carriepl:
Those are the points that I saw. Other than that, it looks ok to me.

theano/sandbox/rng_mrg.py
abergeron: This was something that was forgotten when doing the conversion to multi-gpu.  I just noticed it during the rename.

doc/extending/using_params.txt
abergeron: I've added some text to help decide.

abergeron:
I should have done all the changes.

abergeron:
This should be ok now.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,214 @@
...
+.. warning::
+
+   If you do not account for this extra parameter, the code will fail
+   at runtime if it tries to run the python version.
+
+Also, for the C code, the `sub` dictionary will contain an extra entry
+`'params'` which will map to the variable name of the params object.
+This is true for all methods that recieve a `sub` parameter, so this
+means that you can use your params in the :meth:`c_init_code_struct
+<Op.c_init_code_struct>` method.
```
nouiz: add c_code at the start of this list of method that accept it.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,214 @@
...
+       def make_node(self, x, y):
+           x = as_scalar(x)
+           y = as_scalar(y)
+           return Apply(self, [x, y], [x.type()]
+
+       def c_code(self, node, name, inputs, outputs, sub):
+           return """
+   PyObject *tmp;
+   double a, b;
+   tmp = PyObject_GetAttrString(%(p)s, "alpha");
```
nouiz: Can you move those param access to c_init_code_struct to show this can be done only once to speed up execution?

doc/extending/using_params.txt
```diff
@@ -0,0 +1,214 @@
...
+   if (tmp == NULL)
+     %(fail)s
```
nouiz: put {} around fail. We don't know in the futur if it will change and contain multiple statement. So we are better to give good example.

doc/extending/using_params.txt
```diff
@@ -0,0 +1,214 @@
...
+   a = PyFloat_AsDouble(tmp);
+   Py_DECREF(tmp);
+   tmp = PyObject_GetAttrString(%(p)s, "beta");
+   if (tmp == NULL)
+     %(fail)s
```
nouiz: idem.

doc/extending/using_params.txt
carriepl: On the other hand, this would make this tutorial inconsistent with others that don't use {}.
(for instance  : http://deeplearning.net/software/theano/tutorial/extending_theano_c.html#c_code). In order to avoid confusing people as to why it is done like that, the tutorials should probably agree on one way to do this, whether the current one or the one suggested.

nouiz: I would propose my suggestion, but I won't delay this PR for that.

If someone get confused by this, they probably don't know C and so probably
shoudn't try to do that... But I agree, being consistent is good.

carriepl:
I'm OK with how the points I have raised have been addressed. @nouiz : once the points you have raised are resolved to your satisfaction, you have my blessing to merge.

doc/extending/using_params.txt
abergeron: Fail currently contains multiple statements but wraps them in braces so it's safe.

nouiz: Seem like a great solution:)

nouiz:
How this work with the c linker? If we start to use it with other op, would it still work?

doc/extending/using_params.txt
abergeron: ok

doc/extending/using_params.txt
abergeron: ok

abergeron:
I'm pretty sure this works with multiple ops in CLinker.  I haven't tried though.

PR #3563, opened 2015-10-26
Fix doc

abergeron:
This fixes a lot of syntax errors, sphinx warnings and plain weird rendering in the docs.


nouiz:
I'm ok to merge when travis pass.

bouthilx:
Blocksparse were indeed "fugly" but the new ones are misleading. If you want to leave it to me I'll reformat it properly with correct comments.

abergeron:
I would merge this since it changes very little in the contents itself.  Then you are more than welcome to clarify the docs.

bouthilx:
Alright!

PR #3567, opened 2015-10-27
Re-enable mkl for python 2.6 since then aren't numpy pacakges of 1.7 with openblas.

abergeron:


abergeron:
Since this fixes the master I will merge right away.

PR #3568, opened 2015-10-27
Python 3 doesn't have numpy packages with openblas, apparently.

abergeron:


abergeron:
This should completely fix the master, we're back to using mkl.

From my investigations, numpy is linked with openblas starting with 1.9.2 on python 2.7.  For python 3.3 there are no numpy packages with openblas, so we're going to be stuck with mkl for a while.

Issue #3560, opened 2015-10-26
Make ShapeFeature.get_shape() update recurively


nouiz:
See the comment in that method.

PR #3570, opened 2015-10-28
Fix extra scan due to old information in infer shape.

nouiz:
fix gh-3560

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -969,10 +969,10 @@ class ShapeFeature(object):
                 r.owner.inputs[0] not in var.fgraph.variables):
             assert var.owner
             node = var.owner
-# TODO recur on inputs
-# Need to time this to don't have it too slow.
-# Make sure to handle the case of (shape_i(x)+1)
-# see https://github.com/Theano/Theano/issues/3560
+            # recur on inputs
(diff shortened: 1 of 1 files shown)


abergeron:
That looks ok to me but someone should have a second look.

carriepl:
This also looks good to me. Based on this and the result of Arnaud's review, I'll be merging.

nouiz:
It is good that it is merged, but it is possible that this slow down
compilation. I'm still investigating this.

carriepl:
I mentionned this IRL to @nouiz but I prefer to also have a written trace : there is a use case from a user (@caglar) which took ~830 seconds to compile on my machine before the introduction of the bug that this PR meant to resolve but now takes ~575 seconds. So it is very possible this PR might slow down compilation but we have at least one data point where it is still very useful.

As @nouiz pointed out, there is a good chance that this improvement comes from the changes that were done in the same PR as the bug was introduced.

nouiz:
I compared the compilation time for gh-3412 and it didn't change. I also checked for the GSoC case, and it did raise a little, but I think this is due to noise and not a real problem. Like the CPU being not overclocked the same way on both computer where I did it.

So I'm pretty sure the compilation time didn't got slowed down.

PR #3572, opened 2015-10-28
Fix index errors

carriepl:
In all 6 modified files, the code was supposed to look at the spatial dimensions of the inputs and the filters in the case of 2d convolution.

These should be dimensions with 2 and 3 since the input dimensions are (batch, channel, 0, 1) and the filter dimensions are (filter, channel, 0, 1). However, in both backends, the indices were wrong. This meant that the FFT implementation was not used in many cases where it could have been used.


abergeron:
Yep, ok.

Issue #3381, opened 2015-09-11
function_dump should be able to clean tags of the variables


rizar:
There is a function [function_dump](https://github.com/Theano/Theano/blob/752f1f73363554351e39d070914a2dbc354ba4eb/theano/compile/function.py#L23) which serves debugging purposes. For instance, when one wants to ask Theano developers to see why his function compiles too long, they can replace `theano.function` with `function_dump` call, specify the file where the dump should be saved, and send the file to the developers.

This does not work well when the computation graph was created by Blocks, because the variables may contain references to used-defined objects in their `.tag` attributes. This makes the dump unpicklable. 

The task of this ticket is to add cleaning variable tags to `function_dump`. [StripPickler](https://github.com/Theano/Theano/blob/414300e78041ef55cb0c419a96add960caa42121/theano/misc/pkl_utils.py#L43) can be used for this purpose, which however has to be first slightly improved to support a list of attributes to be cleaned (currently `.trace` only). `function_dump` in its turn would accept a list of attributes as a keyword argument and pass it to `StripPickler`.

nouiz:
I would at the same time also remove test_value by default in StripPickler.

On Fri, Sep 11, 2015 at 11:01 AM, Dzmitry Bahdanau <notifications@github.com

rizar:
@nouiz , you told this is being done, but I see no activity here. Can you please tell me if it has been assigned to anybody?

nouiz:
@abergeron where this is?

I have a quick work around here: https://github.com/Theano/Theano/pull/3576

PR #3576, opened 2015-10-29
[ENG] Doc, blas timming, better error, doc to use pydot-ng, function_dump now strip some stuff.

nouiz:


lamblin:
A rebase is needed. I'll try to review what I can.

theano/misc/pkl_utils.py
```diff
@@ -60,7 +60,17 @@ def save(self, obj):
         if isinstance(obj, theano.gof.utils.scratchpad):
             if hasattr(obj, 'trace'):
                 del obj.trace
-
+            if hasattr(obj, 'test_value'):
+                del obj.test_value
+            # The next 4 items are from Blocks
+            if hasattr(obj, 'annontations'):
```
lamblin: Should that be 'annotations' instead?

theano/misc/pkl_utils.py
```diff
@@ -60,7 +60,17 @@ def save(self, obj):
...
+                del obj.annontations
+            if hasattr(obj, 'replacement_of'):
+                del obj.replacement_of
+            if hasattr(obj, 'aggregation_scheme'):
+                del obj.aggregation_scheme
+            if hasattr(obj, 'rolesc'):
+                del obj.rolesc
```
lamblin: Should we remove the whole content of the scratchpad, instead of hardcoding names from another framework?

theano/sandbox/cuda/__init__.py
```diff
@@ -429,6 +429,10 @@ def use(device,
                 # query the active GPU. If we check the active GPU before
                 # the device is initialized we will always receive 0
                 # event if another device is selected later.
+                if not hasattr(cuda_ndarray.cuda_ndarray, 'select_a_gpu'):
+                    raise Exception(
+                        "Delete your Theano cache. The automatic"
+                        " recompilation didn't worked.")
```
lamblin: "did not work"

theano/misc/pkl_utils.py
nouiz: I don't think so, there is a mixture of stuff there. I think we should revisit this by taking our time and make sure what we want to do. We need a way to keep some information during pickling and right now it isn't possible, documented.

nouiz:
I updated the PR following the code review.

nouiz:
I rebased this PR.

theano/misc/pkl_utils.py
lamblin: I'm still uncomfortable hardcoding names from another framework. What if a different framework wants to pickle `tag.annotations`? That should maybe go in a Pickler from Blocks instead.

lamblin: Or a list of things to strip from `tag` could be passed to `StripPickler.__init__` for instance.

abergeron: Or `config.pickle.strip_tags`.

nouiz:
I updated this PR, this finish completly gh-3381

theano/tests/breakpoint.py
```diff
@@ -12,8 +12,8 @@ class PdbBreakpoint(Op):
     conditional breakpoint, inside a theano function, based on a symbolic
     scalar condition.
 
-    @type name: String
-    @param name: name of the conditional breakpoint. To be printed when the
+    :type name: String
+    :param name: name of the conditional breakpoint. To be printed when the
```
abergeron: Why go from one undesired format to another?

nouiz: From memory, the first just didn't render and this one does. This PR make Theano better then the current version and do not bring long term problem. So I won't get bugged by this. We didn't call for a rewrite of all the docstring to a specific format.

abergeron: We did agree to have numpydoc as the format for all docstrings.

If that changed, I was not aware.

lamblin: Yes, numpydoc everywhere is the goal. It does not mean that every time we touch a docstring, we have to make it numpydoc-compatible for the PR to be merged.
This formatting change does not reach the goal (numpydoc) but is strictly better than not changing it (now it displays correctly, before it did not), which is why I merged.

PR #3580, opened 2015-10-30
[ENH] Infer more things in Scan.infer_shape()

carriepl:


theano/scan_module/tests/test_scan.py
```diff
@@ -3967,8 +3971,16 @@ def onestep(x, x_tm4):
...
         f = theano.function([seq], results[1], updates=updates)
+
+        # This fails if scan uses wrongly the borrow flag
         assert numpy.all(exp_out == f(inp))
 
+        # This fails if Scan's infer_shape() is unable to remove the Scan
+        # node from the graph.
+        f_infershape = theano.function([seq], results[1].shape)
+        scan_nodes_infershape = scan_nodes_from_fct(f_infershape)
+        assert(len(scan_nodes_infershape) == 0)
```
nouiz: This fail in FAST_COMPIILE. I think we should enable the local_shape_to_shape_i in FAST_COMPILE. This is fast to execute, the feature is already registered and this can remove much computation.

in tensor/opt.py, register local_shape_to_shape_i  like  `@register_canonicalize('fast_compile')

nouiz: This will fix the travis error, they are in FAST_COMPILE

nouiz:
I'm ok with this, but it would be good if @lamblin or @abergeron also check this PR.

Also, maybe we are better to merge this Monday then today.

carriepl:
It seems registering local_shape_to_shape_i in FAST_COMPILE makes a bunch of tests fail. So either
a) we don't keep it in FAST_COMPILE
b) we keep it in FAST_COMPILE but I adapt the tests.
As @nouiz, this depends on what @lamblin says about registering it in FAST_COMPILE in the first place.

theano/scan_module/tests/test_scan.py
lamblin: Is the ShapeFeature registered in FAST_COMPILE? If not, adding it could slow down the whole compilation phase. We should profile that.
I don't think we expect shape inference to be performed in FAST_COMPILE in general, right? So we could just force the mode here for the tests.

lamblin: In fact, in `InferShapeTester`, we [add 'canonicalize'](https://github.com/Theano/Theano/blob/master/theano/tests/unittest_tools.py#L184) because of that.
If we want to include `local_shape_to_shape_i` in `'fast_compile'`, that could be reasonable, but it seems to be an orthogonal change, and we should also change `InferShapeTester` to be consistent.

carriepl:
I forced the new test to use FAST_RUN instead.

carriepl:
Travis has blessed the result. I think every comment has been addressed now.

PR #3418, opened 2015-09-21
Implement DownsampleFactorMax.R_op 

rodrigob:
This is a PR to attack issue #305.
Imported what seemed relevant from pascanur:downsampling_rop branch (PR #622).
When comparing the two code bases, the pascanur branch seems quite archaic.

I am not familiar with the Theano code base, so I did my best pythonic guesses. I most certainly introduced bugs. This code will need heavy testing and/or upgrades.

`test_rop.py` includes a test for the new feature, and it currently fails.

```
....SES..................
======================================================================
ERROR: test_downsample (theano.tests.test_rop.test_RopLop)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/benenson/code/references/Theano/theano/tests/test_rop.py", line 274, in test_downsample
    yv = tensor.Rop(out, input, eval_p)
  File "/home/benenson/code/references/Theano/theano/gradient.py", line 293, in Rop
    _traverse(out.owner)
  File "/home/benenson/code/references/Theano/theano/gradient.py", line 256, in _traverse
    _traverse(inp.owner)
  File "/home/benenson/code/references/Theano/theano/gradient.py", line 288, in _traverse
    seen_nodes[node] = op.R_op(node.inputs, same_type_eval_points)
  File "/home/benenson/code/references/Theano/theano/tensor/signal/downsample.py", line 346, in R_op
    st=self.st, padding=self.padding)
TypeError: __init__() got an unexpected keyword argument 'padding'

----------------------------------------------------------------------
```

I had added `st=self.st, padding=self.padding` because this are new arguments present on the other downsampling operators. Not sure how/if should be present in Rop.

Anyone interested on helping for this PR is most welcome. Theano internals look quite gloomy.

PR #3587, opened 2015-11-02
Downsample factor max.r op

nouiz:
This is a rebase/update to the current interface of work done by Razvan P. The rebase was started in gh-3418

@lamblin can you very how the new ops are tested? Otherwise, I think it is good.

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -233,6 +233,9 @@ class GpuDot22(GpuOp):
         return hash(type(self))
 
     def make_node(self, x, y):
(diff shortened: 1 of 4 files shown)


nouiz:
Before merging, we need to run the GPU tests for the new functionality.

lamblin:
This looks reasonable. The tests are OK, although the test for L_Op is missing, and it could probably be refactored into a test_rop_lop_tensor4 or something like that.

lamblin:
Are there tests for the new GPU Op, or should they be written?

nouiz:
closing as continued in gh-4412

PR #3590, opened 2015-11-02
Cast Py_None to PyTypeObject * to satisfy g++ 4.9.

abergeron:
Otherwise there are a bunch of compile errors in the new backend.

### DIFF
--- a/theano/sandbox/gpuarray/basic_ops.py
+++ b/theano/sandbox/gpuarray/basic_ops.py
@@ -423,7 +423,7 @@ class GpuFromHost(Op):
                                        (size_t *)PyArray_DIMS(%(name)s_tmp),
                                        (ssize_t *)PyArray_STRIDES(%(name)s_tmp),
                                        %(ctx)s,
-                                       Py_None);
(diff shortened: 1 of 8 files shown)


abergeron:
If this seems too weird I have another option which is to change the type of all the APIs in cython to take `object` instead of `type`.  That also works.

nouiz:
It is not very user friendly. Can you explain why g++ 4.9 have problem?
What is the consequence of using object instead of type? Less type checking
from cython when we call the function?

lamblin:
From my perspective, the issue is that `None` is not a type, so `Py_None` is not a `PyTypeObject *`.
If Cython is doing some checks on the passed value, expecting a `PyTypeObject *`, those checks would be meaningless on a casted `Py_None`. If there are some optimizations done by the compiler, then they may be wrong.

abergeron:
Ok I'll roll the changes in the allocation PR for libgpuarray.

nouiz:
I didn't saw this in current PR and I didn't saw a new PR.

PR #3593, opened 2015-11-03
Fix a crash when using Softmax with the CLinker

AlOa:
See https://groups.google.com/forum/#!topic/theano-users/33dqPAIQIHE


nouiz:
I think this is good. When test we can merge.

Did you check if this happen in other file? Maybe we could fix that in the code generator if we continue to find some of them. But I like the idea of this PR to be more consistency and use the same way to access variable everywhere.

PR #3595, opened 2015-11-03
Fix return value of R_op in HostFromGpu/GpuFromHost

lamblin:
This was introduced in #3482, and reported in https://groups.google.com/d/topic/theano-users/DFlcQ-992Ss/discussion


nouiz:
We can merge when travis pass. But we should add tests or make an issue to add tests.

nouiz:
close/reopen for travis fix

PR #3599, opened 2015-11-04
allow symbolic input for max pooling

hycis:


nouiz:
It seem we forgot this PR. Do you want to continue this? Now that the ICML deadline is finished, I'll be able to review and comment it.

thanks

hycis:
Hi Fred,

Thanks for still keeping track of this issue, definitely I will like to continue. My previous pull request seems to fail the testing. Which I’m not sure why. If you can provide some guidance and hints, that will be very helpful. Thanks.

nouiz:
You will need to rebase this PR. There have been change in the master. The op have been renamed. Instead of a rebase, maybe starting from scratch and coping the code from this PR would be useful.

For the error in travis, they are caused by the fact that you didn't update the perform() method and the c_code() method. Now the apply node take more inputs then before, so this question adjustment to the code. You can start by fixing the python code. To help you, use the flag "cxx=", this will disable the usage of c code.

Issue #3341, opened 2015-08-30
[BUG IN FGRAPH.REPLACE OR A LISTENER] maximum recursion depth exceeded


pcyin:
Hi,

My theano function fails to compile with the following error, which starts with <<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>>. The function compiles fine in `FAST_COMPILE` mode.

I tried to increase maximum recursion depth using `sys.setrecursionlimit(100000)`, then the compilation never ends and memory usage explodes (I don't think my model is such complex).

The function dump is [here](http://i.cs.hku.hk/~pcyin/func.bin). Reminiscent of #3192 (UPDATE: this is different cause), is it possible to circumvent this issue using similar trick?

Thanks in advance!

```
<<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.RuntimeError'> maximum recursion depth exceeded in __instancecheck__ local_fill_to_alloc
ERROR: clients intersect!
  RCLIENTS of Elemwise{second,no_inplace}.0 [(Elemwise{second,no_inplace}(Elemwise{second,no_inplace}.0, Elemwise{second,no_inplace}.0), 0, <class 'theano.gof.graph.Apply'>, 39491230672)]
  NCLIENTS of Elemwise{second,no_inplace}.0 [(Elemwise{second,no_inplace}(Elemwise{second,no_inplace}.0, Elemwise{second,no_inplace}.0), 0, <class 'theano.gof.graph.Apply'>, 39491230672)]
ERROR (theano.gof.opt): Optimization failure due to: local_fill_to_alloc
08/29/2015 10:06:04 [ERROR] opt: Optimization failure due to: local_fill_to_alloc
ERROR (theano.gof.opt): TRACEBACK:
08/29/2015 10:06:04 [ERROR] opt: TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/Users/yinpengcheng/Research/DeepLearning/Theano/theano/gof/opt.py", line 1771, in process_node
    fgraph.replace_all_validate(repl_pairs, reason=lopt)
  File "/Users/yinpengcheng/Research/DeepLearning/Theano/theano/gof/toolbox.py", line 306, in replace_all_validate
    fgraph.revert(chk)
  File "/Users/yinpengcheng/Research/DeepLearning/Theano/theano/gof/toolbox.py", line 192, in revert
    f()
[...]

PR #3604, opened 2015-11-04
[ENH] Lower triple recursion to simple recursion

nouiz:
this fix gh-3341

This inline 3 functions inside only 1. This should also speed up compilation time for big graph.

### DIFF
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -454,9 +454,20 @@ class FunctionGraph(utils.object2):
         the variable will stay in fgraph.variables.
 
         """
-        # Prunes the owners of the variables.
+        # Prunes the owners(Apply Node) of the variables.
         if variable.owner:
-            self.__prune__(variable.owner, reason)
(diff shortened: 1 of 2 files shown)


theano/gof/fg.py
```diff
@@ -289,16 +293,60 @@ def __remove_clients__(self, r, clients_to_remove,
...
         clients_to_remove
             List of (op, i) pairs such that node.inputs[i] is not r anymore.
+        prune
+            If prune is True, it remove r from this fgraph if it don't
+            have clients left.
+
+        Returns
+        -------
+
+        True if r is still in the fgraph and need to be pruned
```
abergeron: There shouldn't be a blank line here.  Also the type is mandatory.

nouiz:
I updated the PR.

nouiz:
I fixed the docstring.

abergeron:
The next to last commit passed travis and the last one only modifies a docstring, so I'm merging now.

PR #3605, opened 2015-11-04
Add print_test_value_by_default option

fvisin:
Add a configuration option to override `Variable.__str__` and print the `tag.test_value` after the variable name by default. By default the option is disabled.

@lamblin Thank you very much for the assistance!

### DIFF
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -764,7 +764,7 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
                 outer_id_str = get_id_str(outer_r.owner)
             else:
                 outer_id_str = get_id_str(outer_r)
-            print('%s%s %s%s -> %s' % (prefix, r, id_str, type_str,
+            print('%s%s %s%s -> %s' % (prefix, r.__str_name__(), id_str, type_str,
                                        outer_id_str), file=file)
         else:
             # this is an input variable
@@ -772,7 +772,7 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
             if smap:
                 data = " " + str(smap.get(r, ''))
             id_str = get_id_str(r)
-            print('%s%s %s%s%s' % (prefix, r, id_str,
+            print('%s%s %s%s%s' % (prefix, r.__str_name__(), id_str,
                                    type_str, data),
                   file=file)
 
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -764,7 +764,7 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
                 outer_id_str = get_id_str(outer_r.owner)
             else:
                 outer_id_str = get_id_str(outer_r)
-            print('%s%s %s%s -> %s' % (prefix, r.__str_name__(), id_str, type_str,
+            print('%s%s %s%s -> %s' % (prefix, r, id_str, type_str,
                                        outer_id_str), file=file)
         else:
             # this is an input variable
@@ -772,7 +772,7 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
             if smap:
                 data = " " + str(smap.get(r, ''))
             id_str = get_id_str(r)
-            print('%s%s %s%s%s' % (prefix, r.__str_name__(), id_str,
+            print('%s%s %s%s%s' % (prefix, r, id_str,
                                    type_str, data),
                   file=file)
 
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -645,7 +645,7 @@ AddConfigVar(
 
 
 AddConfigVar(
-    'print_test_value_by_default',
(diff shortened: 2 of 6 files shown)


theano/configdefaults.py
```diff
@@ -640,6 +640,14 @@ def warn_default(version):
     in_c_key=False)
 
 
+AddConfigVar(
+    'print_test_value_by_default',
```
nouiz: also document it in Theano/doc/library/config.txt

nouiz: I would name it print_test_value

It is shorter and to me, it is as much informative. I currently don't see possible futur colision of flag name with that. If you see some, tell me. We can discuss this.

theano/gof/graph.py
```diff
@@ -395,6 +396,17 @@ def __str__(self):
         WRITEME
 
         """
+        if config.print_test_value_by_default:
+            if self.__get_test_value__() is not None:
```
nouiz: Can this call recurse? self.__get_test_value() call repr that call str...

theano/gof/graph.py
```diff
@@ -395,6 +396,17 @@ def __str__(self):
...
+                return '\n'.join([self.__str_name__(),
+                                 self.__get_test_value__()])
```
nouiz: idem

theano/gof/graph.py
```diff
@@ -406,6 +418,12 @@ def __str__(self):
         else:
             return "<%s>" % str(self.type)
 
+    def __get_test_value__(self):
```
nouiz: There is already such implementation in theano.gof.op.get_test_value(). Also, the name isn't very good, it make me thing it return the test value, not string representation.

I would delete this method and refactor to don't need it and call theano.gof.op.get_test_value() to be consistent in how we handle that. This would also make it print constant and shared variable value. Is that good?

nouiz:
Do you know that debugprint have a parameter print_storage?

http://deeplearning.net/software/theano/library/printing.html#theano.printing.debugprint

It is not the same, I'm just telling this in case you don't know about it. I think this PR is complementary.

theano/configdefaults.py
fvisin: Agreed.

fvisin:
I didn't know about print_storage, which is indeed very useful. Good to know, thank you for pointing that out.

fvisin:
I updated the code, now it reuses `theano.gof.op.get_test_value()` and is much cleaner. I also added some code for `cudaNdArray`, but I didn't have a free GPU to test it.

For what concerns the potential infinite recursion, I moved the code in `_repr_`, which should solve the most of the problems. `theano.gof.op.get_test_value()` doesn't call `_repr_` directly (https://github.com/Theano/Theano/blob/ecd9dc0ba6fa70b7f65c6de620c21e502c414c5e/theano/gof/op.py#L528-L566), unless it is called indirectly we should be fine.

theano/compile/debugmode.py
```diff
@@ -764,7 +764,7 @@ def get_id_str(obj):
                 outer_id_str = get_id_str(outer_r.owner)
             else:
                 outer_id_str = get_id_str(outer_r)
-            print('%s%s %s%s -> %s' % (prefix, r, id_str, type_str,
+            print('%s%s %s%s -> %s' % (prefix, r.__str_name__(), id_str, type_str,
```
nouiz: need rename, this make tests fail.

theano/compile/debugmode.py
```diff
@@ -772,7 +772,7 @@ def get_id_str(obj):
             if smap:
                 data = " " + str(smap.get(r, ''))
             id_str = get_id_str(r)
-            print('%s%s %s%s%s' % (prefix, r, id_str,
+            print('%s%s %s%s%s' % (prefix, r.__str_name__(), id_str,
```
nouiz: idem

theano/gof/graph.py
```diff
@@ -406,8 +406,32 @@ def __str__(self):
...
+        Return a printable representation of the test value. It can be
+        overridden by classes with non printable test_value to provide a
+        suitable representation of the test_value.
+        """
+        try:
+            return repr(theano.gof.op.get_test_value(self))
+        except:
+            raise
+
+    def __repr__(self, firstPass=True):
```
nouiz: I think it is the __str__ that must be redefined, not __repr__. Is there a reason you did the reverse?

Also, never make an try: ... except:

Always as least catch Exception. Here, catch only AttributeError

theano/sandbox/cuda/var.py
```diff
@@ -43,6 +43,13 @@ def _as_CudaNdarrayVariable(self):
 
 class CudaNdarrayVariable(_operators, Variable):
     pass
+
+    # override default
+    def __str_test_value__(self):
+        try:
+            return repr(numpy.array(theano.gof.op.get_test_value(self)))
+        except:
```
nouiz: idem

PR #3618, opened 2015-11-10
CorrMM optimizations

JesseLivezey:
TODOs
- [x] write optimizations
- [x] register optimizations
- [x] write tests
- [x] Documentation for the user in doc/library/tensor/nnet/conv.text


nouiz:
that is a good start. Thanks. If you have questions, ask us.

JesseLivezey:
@nouiz @lamblin , I think I have the optimizations written, but I have three questions.

1. What are the correct arguments for registering the optimizations?
2. When should the normal CPU conv op be used?
3. What should a test look like? Or can you point me to a test for an existing optimization?

theano/tensor/nnet/opt.py
```diff
@@ -0,0 +1,88 @@
...
+                                                           shape)
+    return [rval]
+
+# Register CPU convolution implementation
+# They are tried in a specific order so we can control
+# which ones take precedence over others.
+abstractconv_groupopt = theano.gof.optdb.LocalGroupDB()
+abstractconv_groupopt.__name__ = "abstractconv_opts"
+#TODO all below
+register_specialize_device(abstractconv_groupopt, 'gpu', 'fast_compile')
```
lamblin: I think 'gpu' should be removed.

theano/tensor/nnet/opt.py
```diff
@@ -0,0 +1,88 @@
...
+
+# The GEMM-based convolution comes last to catch all remaining cases.
+# It can be disabled by excluding 'conv_gemm'.
+conv_groupopt.register('local_abstractconv_gemm', local_abstractconv_gemm, 30,
```
lamblin: The groupopt here should have the same name that you created above. I have no real preference.

lamblin:
1. This looks like the right way in principle, modulo the line comments I added.
2. I would be tempted to say "never", but it could still be useful in corner cases where CPU memory is limited. I think there are two options:
  - Do not introduce the legacy convolution. In that case, the optimizations at https://github.com/JesseLivezey/Theano/blob/cormm_opt/theano/tensor/nnet/abstract_conv2d.py#L331 should be removed
  - Allow the legacy convolution to be introduced, but only if 'conv_gemm' is explicitly excluded. In that case, these optimizations should be registered in the same `LocalGroupDB` as yours, but with a lower priority (let's say a rank of 40 or so). I think the easier would be to move them to `nnet/opt.py` as well. I would favour this option.

3. With the current state of Theano, I would add a new series of tests in https://github.com/JesseLivezey/Theano/blob/cormm_opt/theano/sandbox/cuda/tests/test_abstractconv.py. It would be modelled after `def test_cormm_conv(self)` and `def test_cpu_conv(self)`. This has the advantage of checking a wide variety of cases against a reference, and checking the gradients by finite-differences as well. The drawback is that the reference is CuDNN, so the tests can only be run on a machine with CuDNN.

JesseLivezey:
@lamblin @nouiz, tests pass (although the numerical tests don't run on travis). I can squash the commits after review.

theano/tensor/nnet/opt.py
```diff
@@ -41,3 +55,299 @@ def local_inplace_sparse_block_outer(node):
...
+                       'conv_gemm', 'fast_compile', 'fast_run')
+# Legacy convolution
+conv_groupopt.register('local_conv2d_gpu', local_conv2d_cpu, 40,
+                       'fast_compile', 'fast_run')
+conv_groupopt.register('local_conv2d_gradweight_cpu',
+                       local_conv2d_gradweight_cpu, 40,
+                       'fast_compile', 'fast_run')
+conv_groupopt.register('local_conv2d_gradinputs_cpu',
+                       local_conv2d_gradinputs_cpu, 40,
+                       'fast_compile', 'fast_run')
```
lamblin: I feel like `fast_run` is not needed if we already have `fast_compile`. @nouiz can you confirm?

lamblin:
I'll try to launch the numerical tests locally tomorrow. Otherwise, it looks good, but maybe @nouiz has more comments.

jonhoo:
I decided to give adding OpenMP support to `CorrMM` a go. With the patch below, I see about a 5x speed increase during forward operation in my particular network on an 80-core machine:

```diff
diff --git a/theano/tensor/nnet/corr.py b/theano/tensor/nnet/corr.py
index 8b704f8..876aa4d 100644
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -6,13 +6,13 @@ from theano import Apply
 from theano import gof
 from theano.tensor import as_tensor_variable, TensorType
 from theano.tensor.blas_headers import blas_header_text
-from theano.tensor.blas import ldflags
+from theano.tensor.blas import ldflags, blas_header_version
 
 
 _logger = logging.getLogger(__name__)
 
 
-class BaseCorrMM(gof.Op):
+class BaseCorrMM(gof.OpenMPOp):
     """
     Base class for `CorrMM`, `CorrMM_gradWeights` and
     `CorrMM_gradInputs`. Cannot be used directly.
@@ -28,7 +28,8 @@ class BaseCorrMM(gof.Op):
     check_broadcast = False
     __props__ = ('border_mode', 'subsample')
 
-    def __init__(self, border_mode="valid", subsample=(1, 1)):
+    def __init__(self, border_mode="valid", subsample=(1, 1), openmp=None):
+        super(BaseCorrMM, self).__init__(openmp=openmp)
         if isinstance(border_mode, int):
             if border_mode < 0:
                 raise ValueError(
@@ -72,7 +73,11 @@ class BaseCorrMM(gof.Op):
         return ldflags()
 
     def c_compile_args(self):
-        return ldflags(libs=False, flags=True)
+        ret = ldflags(libs=False, flags=True)
+        # Add the -fopenmp flags
+        ret += super(BaseCorrMM, self).c_compile_args()
+
+        return ret
 
     def c_lib_dirs(self):
         return ldflags(libs=False, libs_dir=True)
@@ -85,7 +90,7 @@ class BaseCorrMM(gof.Op):
 
     def c_code_cache_version(self):
         # raise this whenever modifying any of the support_code_files
-        return (1, 0)
+        return (2, self.openmp, blas_header_version())
 
     def c_support_code_apply(self, node, nodename):
         # REMEMBER TO RAISE c_code_cache_version when changing any of
@@ -353,8 +358,8 @@ class CorrMM(BaseCorrMM):
         Set to `(1, 1)` to disable subsampling.
 
     """
-    def __init__(self, border_mode="valid", subsample=(1, 1)):
-        super(CorrMM, self).__init__(border_mode, subsample)
+    def __init__(self, border_mode="valid", subsample=(1, 1), openmp=None):
+        super(CorrMM, self).__init__(border_mode, subsample, openmp=openmp)
 
     def make_node(self, img, kern):
         img = as_tensor_variable(img)
@@ -432,8 +437,8 @@ class CorrMM_gradWeights(BaseCorrMM):
 
     """
 
-    def __init__(self, border_mode="valid", subsample=(1, 1)):
-        super(CorrMM_gradWeights, self).__init__(border_mode, subsample)
+    def __init__(self, border_mode="valid", subsample=(1, 1), openmp=None):
+        super(CorrMM_gradWeights, self).__init__(border_mode, subsample, openmp=openmp)
 
     def make_node(self, img, topgrad, shape=None):
         img = as_tensor_variable(img)
@@ -529,8 +534,8 @@ class CorrMM_gradInputs(BaseCorrMM):
 
     """
 
-    def __init__(self, border_mode="valid", subsample=(1, 1)):
-        super(CorrMM_gradInputs, self).__init__(border_mode, subsample)
+    def __init__(self, border_mode="valid", subsample=(1, 1), openmp=None):
+        super(CorrMM_gradInputs, self).__init__(border_mode, subsample, openmp=openmp)
 
     def make_node(self, kern, topgrad, shape=None):
         kern = as_tensor_variable(kern)
diff --git a/theano/tensor/nnet/corr_gemm.c b/theano/tensor/nnet/corr_gemm.c
index 3c58bb8..c2ffee7 100644
--- a/theano/tensor/nnet/corr_gemm.c
+++ b/theano/tensor/nnet/corr_gemm.c
@@ -203,6 +203,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
         output = top;
         // valid correlation: im2col, then gemm
         // Iterate over batch
+        #pragma omp parallel for schedule(static)
         for (int n = 0; n < batchSize; n++) {
             // First, im2col
             im2col((%(float_type)s*)PyArray_DATA(bottom) + n * bottom_stride, nChannels, bottomHeight,
@@ -252,6 +253,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
         output = weight;
         // valid convolution: im2col, then gemm
         // Iterate over batch
+        #pragma omp parallel for schedule(static)
         for (int n = 0; n < batchSize; n++) {
             // First, im2col
             im2col((%(float_type)s*)PyArray_DATA(bottom) + n * bottom_stride, nChannels, bottomHeight,
@@ -303,6 +305,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
         PyArray_FILLWBYTE(bottom, 0);
         // full convolution: gemm, then col2im
         // Iterate over batch
+        #pragma omp parallel for schedule(static)
         for (int n = 0; n < batchSize; n++) {
             // gemm into columns
             %(gemm)s(&NTrans, &Trans,
```

jonhoo:
FWIW, there's further discussion about that patch in #3647. I'd be happy to open a new PR for the patch above if it's something you might consider merging?

theano/tensor/nnet/opt.py
nouiz: all tags are independent. There is at least one optimization that we enable just in fast_compile. So we need to keep both of them.

nouiz:
@jonhoo that is great. Can you make another PR for that? thanks.

theano/tensor/nnet/opt.py
```diff
@@ -41,3 +55,299 @@ def local_inplace_sparse_block_outer(node):
...
+                 direction_hint='bprop inputs')
+    din = din(topgrad, filters)
+    din = theano.tensor.patternbroadcast(din, node.outputs[0].broadcastable)
+    return [din]
+
+
+# Register Cpu Optmization
+conv_groupopt = theano.gof.optdb.LocalGroupDB()
+conv_groupopt.__name__ = "conv_opts"
+register_specialize_device(conv_groupopt, 'fast_compile')
```
nouiz: add fast_run here too.

nouiz:
It seem ok. I let @lamblin run the tests. One thing is missing is the doc for the user. Added that in the todo list in the ticket description with the path to the file.

thanks

lamblin:
So, `test_cormm_conv` seems to be passing, but `test_cpu_conv` is now failing because of a mismatch in the number of channels on one example (full error below). Since the test aborts at the first failure, there may be more. I'll have a quick look and update.
```
ERROR: test_cpu_conv (theano.sandbox.cuda.tests.test_abstractconv.TestConv2d)
  File "/u/lamblinp/code/Theano_git/theano/sandbox/cuda/tests/test_abstractconv.py", line 310, in test_cpu_conv
    filter_flip=flip)
  File "/u/lamblinp/code/Theano_git/theano/sandbox/cuda/tests/test_abstractconv.py", line 95, in run_fwd
    mode=mode)
  File "/u/lamblinp/code/Theano_git/theano/tests/unittest_tools.py", line 99, in verify_grad
  File "/u/lamblinp/code/Theano_git/theano/gradient.py", line 1697, in verify_grad
    analytic_grad = grad_fn(*[p.copy() for p in pt])
  File "/u/lamblinp/code/Theano_git/theano/compile/function_module.py", line 871, in __call__
  File "/u/lamblinp/code/Theano_git/theano/gof/link.py", line 314, in raise_with_op
  File "/u/lamblinp/code/Theano_git/theano/compile/function_module.py", line 859, in __call__
    outputs = self.fn()
ValueError: ("the filter stack size (8) and image stack size (5) differ\nApply node that caused the error: ConvOp{('imshp', (None, None, None)),('kshp', (None, None)),('nkern', None),('bsize', None),('dx', 1),('dy', 1),('out_mode', 'full'),('unroll_batch', None),('unroll_kern', None),('unroll_patch', True),('imshp_logical', (None, None, None)),('kshp_logical', (None, None)),('kshp_logical_top_aligned', True)}(Elemwise{Cast{float32}}.0, Subtensor{::, ::, ::int64, ::int64}.0)\nToposort index: 5\nInputs types: [TensorType(float32, 4D), TensorType(float32, (True, False, False, False))]\nInputs shapes: [(8, 5, 11, 11), (1, 8, 12, 12)]\nInputs strides: [(2420, 484, 44, 4), (4608, 576, -48, -4)]\nInputs values: ['not shown', 'not shown']\nOutputs clients: [['output']]\n\nHINT: Re-running with most Theano optimization disabled could give you a back-trace of when this node was created. This can be done with by setting the Theano flag 'optimizer=fast_compile'. If that does not work, Theano optimizations can be disabled with 'optimizer=None'.\nHINT: Use the Theano flag 'exception_verbosity=high' for a debugprint and storage map footprint of this apply node.", '\nThe error happened with the following inputs:', [array([[[[  7.12963641e-01,   6.90243542e-01,   1.03643179e-01, ...,
```

theano/tensor/nnet/opt.py
```diff
@@ -41,3 +55,299 @@ def local_inplace_sparse_block_outer(node):
...
+conv_groupopt.register('local_abstractconv_gemm', local_abstractconv_gemm, 30,
+                       'conv_gemm', 'fast_compile', 'fast_run')
+conv_groupopt.register('local_abstractconv_gradweight_gemm',
+                       local_abstractconv_gradweight_gemm, 30,
+                       'conv_gemm', 'fast_compile', 'fast_run')
+conv_groupopt.register('local_abstractconv_gradinputs_gemm',
+                       local_abstractconv_gradinputs_gemm, 30,
+                       'conv_gemm', 'fast_compile', 'fast_run')
+# Legacy convolution
+conv_groupopt.register('local_conv2d_gpu', local_conv2d_cpu, 40,
```
lamblin: That tag should be 'local_conv2d_cpu'

theano/tensor/nnet/opt.py
```diff
@@ -41,3 +55,299 @@ def local_inplace_sparse_block_outer(node):
...
+    rval = conv2d(img, kern,
+                  node.op.imshp, node.op.kshp,
+                  border_mode=node.op.border_mode,
+                  subsample=node.op.subsample)
+    return [rval]
+
+
+@local_optimizer([AbstractConv2d_gradWeights])
+def local_conv2d_gradweight_cpu(node):
+
```
lamblin: I think I understand where the issue comes from. Here, we should still check that `node` is a `..._gradWeights`, otherwise it could get applied to another node. When I ran the tests on the master branch, the optimizations were applied in the right order so that did not matter, but now that bug is uncovered.

theano/tensor/nnet/opt.py
```diff
@@ -41,3 +55,299 @@ def local_inplace_sparse_block_outer(node):
...
+    if node.op.border_mode == 'valid':
+        res = res.dimshuffle((1, 0, 2, 3))
+        res = res[:, :, ::-1, ::-1]
+
+    res = theano.tensor.patternbroadcast(res, node.outputs[0].broadcastable)
+    return [res]
+
+
+@local_optimizer([AbstractConv2d_gradInputs])
+def local_conv2d_gradinputs_cpu(node):
```
lamblin: Same here.

lamblin:
I'll do that change and check if it actually fixes the tests, and if so, submit a PR to your branch.

lamblin:
That looks great, thanks!

Issue #3614, opened 2015-11-08
switch shape failure due to optimization


danielrenshaw:
An optimization appears to be changing the output of `switch` expressions incorrectly. Possibly occurs only when the true condition and false condition values are the same.

Reproduced in latest Github version of Theano as of 7:30am GMT, Sunday 8th November 2015.

Reproduction code:

import theano
import theano.tensor as tt

    theano.config.compute_test_value = 'raise'
    x = tt.lvector()
    x.tag.test_value = [1, 2]
    y = tt.switch(x > 0, 1, 1)
    f = theano.function([x], y)
    print y.tag.test_value
    print f(x.tag.test_value)
    theano.printing.debugprint(y)
    theano.printing.debugprint(f)

Prints

    [1 1]
    [1]
    Elemwise{switch,no_inplace} [@A] ''   
     |Elemwise{gt,no_inplace} [@B] ''   
     | |<TensorType(int64, vector)> [@C]
     | |DimShuffle{x} [@D] ''   
     |   |TensorConstant{0} [@E]
     |DimShuffle{x} [@F] ''   
     | |TensorConstant{1} [@G]
     |DimShuffle{x} [@H] ''   
       |TensorConstant{1} [@G]
    DeepCopyOp [@A] ''   0
     |TensorConstant{(1L,) of 1} [@B]

Changing `y = tt.switch(x > 0, 1, 1)` to something like `y = tt.switch(x > 0, 1, 2)` yields expected results.

Originally reported by Dong-Hyun Lee on theano-dev mailing list: https://groups.google.com/forum/#!topic/theano-dev/gHeI49PqjfY

Original report:

    I might have found a bug in function theano.tensor.switch. 
	 

	<code>

	import theano, cPickle, time, os
	import theano.tensor as T
	import numpy as np

	input = np.array([[1.,0.,1.],[0.,1.,0.]], dtype='float32')
	X = T.fmatrix()
	ss = theano.function( [X], T.switch( X > 0 , 0.5 + 0.*X, 0.5 - 0.*X ) )
	ss(input)

	<output>
	array([[ 0.5]], dtype=float32)

	but it should be 2x3 matrix. If i modify the code, it's output is correct.

	<code>
	input = np.array([[1.,0.,1.],[0.,1.,0.]], dtype='float32')
	X = T.fmatrix()
	ss = theano.function( [X], T.switch( X > 0 , 0.499999 + 0.*X, 0.5 - 0.*X ) )
	ss(input)

[...]

PR #3619, opened 2015-11-10
fix gh-3614, fix switch opt when both branch are the same and we broa…

nouiz:
…dcast the condition. Implement it in an optimized way

### DIFF
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -4589,7 +4589,7 @@ class T_local_erfc(unittest.TestCase):
         print(t1 - t0, t2 - t1)
 
 
-class test_local_remove_switch_const_cond(unittest.TestCase):
+class test_local_useless_switch(unittest.TestCase):
     def setUp(self):
         self.mode = mode_opt.excluding('constant_folding')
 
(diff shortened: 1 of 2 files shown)


theano/tensor/tests/test_opt.py
```diff
@@ -4717,6 +4717,23 @@ def test_broadcast2(self):
...
+
+        # This case is not optimized for now.
+        x = theano.tensor.matrix('x', dtype='int32')
+        y = theano.tensor.vector('y', dtype='int64')
+        z = theano.tensor.switch(x, y, y)
+        f = theano.function([x, y], z, mode=self.mode)
+        vx = numpy.array([[0, 1], [1, 0]], dtype='int32')
+        vy = numpy.array([7, 8], dtype='int64')
+        utt.assert_allclose(f(vx, vy), numpy.where(vx, vy, vy))#[[7, 8], [8, 7]])
+        theano.printing.debugprint(f)
```
carriepl: That debugprint should be removed

nouiz:
I amended the commit.

carriepl:
The latest version seems fine to me. Once Travis is happy, we can merge

PR #3633, opened 2015-11-11
Fix pydotprint issue for Python 3

itdxer:
The main problem is that `zip` function in Python 3 is an iterator and `reversed` function doesn't work with an iterators. 

Previously I got `TypeError` exception: 

```
  File "theano/printing.py", line 847, in pydotprint
    fct.maker.fgraph.inputs)):
TypeError: argument to reversed() must be a sequence
```

That fix works fine for me in Python 3.4


nouiz:
thanks. We can merge when travis pass.

PR #3634, opened 2015-11-11
Better error in Scan

nouiz:


carriepl:
This looks correct. I'm okay with merging after Travis is happy

PR #3636, opened 2015-11-12
Don't check output of [Gpu]AllocEmpty

nouiz:


abergeron:
I am against a list of ops like this.  Add an attribute to the op classes and check for it in those functions.

PR #3640, opened 2015-11-13
Scan with inplace in inner function

albertz:
The initial discussion has been [here](https://groups.google.com/forum/#!topic/theano-dev/bmlu0d4Zs5Y).


carriepl:
The 3 commits which add code to force usage of the python version and eventually remove it should be rebased away to keep the commit history clean.

albertz:
Yea sure, I was going to rebase all this and clean it up. When I do that now, will that remove your older comments to the old commits? So maybe I do at only at the very end to leave your comments in here for easier discussion?

carriepl:
You're absolutely right. I fully agree.

albertz:
I see that there is already a merge conflict now. I guess this is because of the recent commit which increases the scan_perform.pyx version number. Maybe, when we disable the new option by default (prefer_inplace), we can already merge it in, once I rebased it and cleaned it up, to avoid further merge conflicts? prefer_inplace and allow_output_prealloc are also not mutually exclusive anymore.

carriepl:
It's not only the version number; some changes have been made to move checks from the execution to the compilation in order to speed up the execution. In the future, however, there should be some time before we need to change anything in that part of Scan so the likelihood of future merge conflicts is pretty low.

2 more things which I should have pointed out earlier : 
- Usually, since Scan is a complicated module, we require that 2 people review before it is merged instead of just one (the other person will most likely be @lamblin, but might also be @abergeron).
- Even before the merge conflicts, there were issues in Travis where some tests were failing. Before even considering merging anything, it needs to be established whether the tests failed because of the change in default behavior or because of an invalid behavior related to some changes in this PR.

albertz:
I rebased everything now and also cleaned it up a bit. The recompile of the Cython version is missing for now but I guess my `_same_data_and_repr` and `_assign_if_needed` can be improved anyway (in both the pure Python and Cython version).

PR #3643, opened 2015-11-14
Rewrite Reshape.infer_shape

SinaHonari:
issue #3420

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4400,6 +4400,11 @@ class Reshape(Op):
         # Here, we only simplify if the shape (node.inputs[1]) is a constant,
         # ideally it would suffice to check that it is always non-negative.
 
+        # If current variable is a scalar and its dimensionality should
+        # change to self.ndim, then use size 1 for all new dimensions.
+        if len(ishapes[0]) == 0:
+            return [(1,) * self.ndim]
+
         requ = node.inputs[1]
         if isinstance(requ, theano.tensor.TensorConstant):
             requ = list(requ.data)
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4417,17 +4417,12 @@ class Reshape(Op):
                                  ' must have at most one entry equal to -1')
             return [requ]
         else:
-            oshape = []
-            for i in xrange(self.ndim):
-                default_os_i = theano.tensor.opt.Shape_i(i)(node.outputs[0])
-                try:
-                    os_i = get_scalar_constant_value(node.inputs[1][i]).item()
-                    if os_i == -1:
-                        os_i = default_os_i
-                except NotScalarConstantError:
-                    os_i = default_os_i
-                oshape.append(os_i)
-            return [tuple(oshape)]
+            new_dims = [node.inputs[1][i] for i in xrange(self.ndim)]
+            return [tuple([switch(eq(new_dims[i], -1),
+                                  theano.tensor.mul(*ishapes[0]) /
+                                  (-theano.tensor.mul(*new_dims)),
+                                  new_dims[i])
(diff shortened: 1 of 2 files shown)


SinaHonari:
There were some errors in Travis, among which one was related.
A couple of errors did could not be reproduced locally (locally it passed) and some errors was not related to the part I had changed.

I updated test_bad_shape to fix it when bad shapes was used.

theano/tensor/basic.py
```diff
@@ -4417,17 +4417,12 @@ def infer_shape(self, node, ishapes):
...
-                    if os_i == -1:
-                        os_i = default_os_i
-                except NotScalarConstantError:
-                    os_i = default_os_i
-                oshape.append(os_i)
-            return [tuple(oshape)]
+            new_dims = [node.inputs[1][i] for i in xrange(self.ndim)]
+            return [tuple([switch(eq(new_dims[i], -1),
+                                  theano.tensor.mul(*ishapes[0]) /
+                                  (-theano.tensor.mul(*new_dims)),
```
lamblin: At least some failures in travis are coming from the fact that one of these `mul(*...)` is called with no arguments, meaning that `new_dims` or `ishapes[0]` is empty.
A way to fix that would be to replace those with 1.

theano/tensor/basic.py
```diff
@@ -4417,17 +4422,12 @@ def infer_shape(self, node, ishapes):
...
-                    os_i = get_scalar_constant_value(node.inputs[1][i]).item()
-                    if os_i == -1:
-                        os_i = default_os_i
-                except NotScalarConstantError:
-                    os_i = default_os_i
-                oshape.append(os_i)
-            return [tuple(oshape)]
+            new_dims = [node.inputs[1][i] for i in xrange(self.ndim)]
+            return [tuple([switch(eq(new_dims[i], -1),
+                                  theano.tensor.mul(*ishapes[0]) /
```
lamblin: Oh, this should be explicit integer division, `//` instead of `/`. I think it would solve the issue with python 3.

lamblin:
Cool, thanks!

theano/tensor/basic.py
```diff
@@ -4417,17 +4422,12 @@ def infer_shape(self, node, ishapes):
...
-                    if os_i == -1:
-                        os_i = default_os_i
-                except NotScalarConstantError:
-                    os_i = default_os_i
-                oshape.append(os_i)
-            return [tuple(oshape)]
+            new_dims = [node.inputs[1][i] for i in xrange(self.ndim)]
+            return [tuple([switch(eq(new_dims[i], -1),
+                                  theano.tensor.mul(*ishapes[0]) //
+                                  (-theano.tensor.mul(*new_dims)),
```
nouiz: There is a neg in this line. Why? Shapes returned by infer_shape shouldn't be negative I think.

So this probably mean there is missing infer_shape test.

SinaHonari: negative is correct, because one of the dims is negative and by getting the negative of multiplication result
we make the result positive.

lamblin: Could you add a comment explaining that in the code?

SinaHonari: Sure, since it PR is already merged I made a new PR (#3738) and linked to this one

bouthilx: Maybe you should test that there is not more than one -1 in new_dims as it is done just above in the TensorConstant case.

lamblin: It is quite hard to do symbolically, especially without adding assert nodes that could disrupt optimizations, so we decided not to do it. I suppose that could go in a comment as well.

SinaHonari: I added a comment.

Issue #3582, opened 2015-10-30
Unify shape inference for convolutions


lamblin:
We should have one implementation of shape inference for each type of convolution Op, I think it belongs in `theano/tensor/nnet/abstract_conv2d`.
Then, each implementation of the Conv operators, including the abstract one, should implement `infer_shape`.
Currently, we have such computation for CorrMM, DNN conv ops, the legacy CPU Conv Op, and GpuCorrMM.

lamblin:
Closed by #3644.

lamblin:
Still missing for DNN conv ops in the old back-end.

lamblin:
Should we also deprecate the static `get_out_shape` in `GpuDnnConv` Ops, and call `get_conv_output_shape` directly [here](https://github.com/laurent-dinh/Theano/blob/conv_infer_shape/theano/sandbox/cuda/dnn.py#L1151-L1153), [here](https://github.com/laurent-dinh/Theano/blob/conv_infer_shape/theano/sandbox/cuda/dnn.py#L1234-L1236), and [here](https://github.com/laurent-dinh/Theano/blob/conv_infer_shape/theano/sandbox/gpuarray/dnn.py#L815-L817)?

nouiz:
I think so.

nouiz:
@lamblin should we close this issue?

lamblin:
Yes, I think we're done.

PR #3644, opened 2015-11-15
Factoring inference of convolution output shape

laurent-dinh:
Tackling [this ticket](https://github.com/Theano/Theano/issues/3582).

### DIFF
--- a/theano/sandbox/gpuarray/dnn.py
+++ b/theano/sandbox/gpuarray/dnn.py
@@ -12,6 +12,7 @@ from theano.gof.type import CDataType, Generic
 from theano.compile import optdb
 from theano.compile.ops import shape_i
 from theano.tensor.nnet import SoftmaxGrad
+from theano.tensor.nnet.abstract_conv2d import get_conv_output_shape
 from theano.tensor.signal.downsample import (
     DownsampleFactorMax, MaxPoolGrad, AveragePoolGrad)
 
@@ -473,48 +474,11 @@ class GpuDnnConv(DnnBase):
         or scalar.
 
         """
-        b = ishape[0]  # Number of inputs
-        h = ishape[2]  # Height of input feature maps
-        w = ishape[3]  # Width of input feature maps
-        nb = kshape[0]  # Number of output feature maps
-        kh = kshape[2]  # Height of each filter
-        kw = kshape[3]  # Width of each filter
-
-        nd = len(subsample)
-
-        if nd > 2:
-            d = ishape[4]
-            kd = ishape[4]
-
-        sh = subsample[0]
-        sw = subsample[1]
-        if nd > 2:
-            sd = subsample[2]
-
-        if border_mode == 'full':
-            padh = kh - 1
-            padw = kw - 1
-            if nd > 4:
-                padd = kd - 1
-        elif isinstance(border_mode, tuple):
-            padh = border_mode[0]
-            padw = border_mode[1]
-            if nd > 2:
-                padd = border_mode[2]
-        else:
-            assert border_mode == 'valid'
-            padh = 0
-            padw = 0
-            padd = 0
-
-        res = [b, nb,
-               (h + 2 * padh - kh) // sh + 1,
-               (w + 2 * padw - kw) // sw + 1]
-
-        if nd > 2:
-            res.append(d + 2 * padd - kd // sd + 1)
-
-        return res
+        return get_conv_output_shape(
+            ishape,
+            kshape,
+            border_mode,
+            subsample)
 
     def infer_shape(self, node, shape):
         return [shape[2]]
--- a/theano/tensor/nnet/conv.py
+++ b/theano/tensor/nnet/conv.py
@@ -14,12 +14,14 @@ import logging
 
 import numpy
 from six.moves import xrange
+import warnings
 
 import theano
 from theano import OpenMPOp
 from theano.tensor import (as_tensor_variable, blas, get_scalar_constant_value,
                            patternbroadcast, NotScalarConstantError)
 from theano.gof import Apply
+from theano.tensor.nnet.abstract_conv2d import get_conv_output_shape
 
 try:
     # TODO: move these back out to global scope when they no longer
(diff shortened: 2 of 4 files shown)


theano/sandbox/gpuarray/dnn.py
```diff
@@ -473,48 +474,11 @@ def get_out_shape(ishape, kshape, border_mode, subsample):
...
-
-        if nd > 2:
-            res.append(d + 2 * padd - kd // sd + 1)
-
-        return res
+        return get_conv2d_output_shape(
+            ishape,
+            kshape,
+            border_mode,
+            subsample)
```
lamblin: DnnConv also supports 3D convolution, so it is a bit confusing to call it "conv2d" here.

theano/tensor/nnet/conv.py
```diff
@@ -34,6 +34,87 @@
...
+    image_shape: tuple of int corresponding to the input image shape. Its four
+        element must correspond respectively to: batch size, number of input
+        channels, height and width of the image. None where undefined.
+    kernel_shape: tuple of int corresponding to the kernel shape. Its four
+        elements must correspond respectively to: number of output channels,
+        number of input channels, height and width of the kernel. None
+        where undefined.
+    border_mode: string or tuple of int. If it is a string, it must be
+        'valid', 'half' or 'full'. If it is a tuple, its two elements
+        respectively correspond to the padding on height and width axis.
```
lamblin: You also support one integer, but it is not documented.

theano/tensor/nnet/conv.py
```diff
@@ -34,6 +34,87 @@
...
+    subsample: tuple of int. Its two elements espectively correspond to the
+        subsampling on height and width axis.
+
+    Returns
+    -------
+    output_shape: tuple of int corresponding to the output image shape. Its
+        four element must correspond respectively to: batch size, number of
+        output channels, height and width of the image. None where undefined.
```
lamblin: Actually, these functions are supposed to work with Theano scalar variables of integer dtype, not only Python ints. In fact, when they are called through `infer_shape`, their inputs will be symbolic scalars.

theano/tensor/nnet/conv.py
```diff
@@ -511,12 +592,16 @@ def __init__(self, imshp=None, kshp=None, nkern=None, bsize=None,
                 _logger.warn(warnstr, self.unroll_kern, self.nkern, new)
                 self.unroll_kern = new
 
-        self.outshp = ConvOp.getOutputShape(self.imshp_logical[1:],
-                                            self.kshp_logical, (dx, dy),
-                                            output_mode)
```
lamblin: While you're at it, could you change `ConvOp.getOutputShape` itself, so it calls `get_conv2d_output_shape`, and maybe emits a DeprecationWarning?

lamblin:
You will have to rebase on top of the latest master. Presumably that will get rid of the first (unneeded) commit.

theano/tensor/nnet/conv.py
```diff
@@ -511,12 +598,16 @@ def __init__(self, imshp=None, kshp=None, nkern=None, bsize=None,
...
-        self.fulloutshp = ConvOp.getOutputShape(self.imshp_logical[1:],
-                                                self.kshp_logical, (1, 1),
-                                                output_mode)
+        self.outshp = get_conv2d_output_shape(
+            (None,) + self.imshp_logical,
+            (None, None,) + self.kshp_logical,
+            output_mode,
+            (dx, dy))[2:]
```
lamblin: Well, now it is called `get_conv_output_shape`...

lamblin:
That seems mostly good. Now, I think we should discuss where that code should be. I would be tempted to put it in `abstract_conv2d.py` since it does not depend on any particular implementation, but it also works for 3D.
In any case, I think it should be in `theano.tensor.nnet` and be imported directly from there, regardless of where the implementation is.

lamblin:
Finally, we decided to put the method in `abstract_conv2d.py`, since it will be renamed.

lamblin:
Thanks!

f0k:
Wait, what about `theano.sandbox.cuda.*`? There are some other convolution ops that have a custom `infer_shape()` or lack one.

lamblin:
Ops missing infer_shape in `theano.sandbox.cuda` is not a big problem in general, since they are introduced by optimizations, as long as the initial Op (here, the abstract one) has an `infer_shape`. It would be nice to have, but was out of the scope for that ticket.
The conv3d Ops in `theano.sandbox.cuda` should be removed soon IMO, I don't think they even compile anymore.
There is also C code computing output shapes, but I think that's another piece of work entirely.
Did you have a specific one in mind?

f0k:
> Did you have a specific one in mind?

Just the ones that are sometimes used directly (if just for the lack of a fully-featured conv2d interface): the `GpuDnnConv*` and `GpuCorrMM*`. Even if there are no practical implications, it would probably be good to unify everything to make it more maintainable.

> The conv3d Ops in `theano.sandbox.cuda` should be removed soon IMO, I don't think they even compile anymore.

You mean `GpuConv3D` et al., right? `GpuCorr3dMM` should be alive and prosper.

lamblin:
Right, `cuda.dnn.GpuDnnConv{3D,}.get_output_shape` should be replaced as well, I did not realize it had only been done in the new back-end.
For `GpuCorrMM`, if they are used directly, then `infer_shape` should be implemented, see #3680

lamblin:
And yes, I meant the ones in `GpuConv*3D.py`.

theano/tensor/nnet/conv.py
```diff
@@ -363,10 +365,13 @@ def getOutputShape(inshp, kshp, stride=(1, 1), mode='valid'):
...
-                     else ((i - k) // d + 1) if mode == 'valid'
-                     else ((i + k + d - 2) // d)
-                     for i, k, d in zip(inshp, kshp, stride))
+        warnings.warn("The method `getOutputShape` is deprecated use"
+                      "`get_conv_output_shape` instead.")
+        return get_conv_output_shape(
+            image_shape=(None, None, inshp[0], inshp[1]),
+            kernel_shape=(None, None, kshp[0], kshp[1]),
+            border_mode=mode,
+            subsample=stride)
```
lamblin: That method should only return `(rows, cols)` for the output image, not the 4-tuple returned by `get_conv_output_shape`. I can see 2 options here:
- call `get_conv_output_shape` and return only the last 2 elements, or
- call `get_conv_shape_1axis` twice (once for each dimension instead

I would favor the second one.

dwf: The fact that this got merged suggests that unit tests for getOutputShape were lacking. Hopefully there's better coverage for get_conv_output_shape?

f0k: > I would favor the second one.

I'd favor the first one, otherwise you'll need to replicate the distinction between `mode` being a string and `mode` being a tuple of int here.

lamblin: @dwf There are no tests for these specific methods to my knowledge, they are tested indirectly via the tests for `infer_shape`. I guess we could add that, #3686.
@f0k `getOutputShape` has never supported `mode` being a tuple of ints, only "full" or "valid", and I don't think we want to add that.

f0k: > `getOutputShape` has never supported `mode` being a tuple of ints, only "full" or "valid"

Ah, sorry, forgot that this was the legacy `ConvOp`. I had assumed the docstring was outdated. Using `get_conv_shape_1axis()` seems good then.

PR #3651, opened 2015-11-17
Port the speedup from #3163 to the new backend.

abergeron:
port gh-3163 to the new back-end


nouiz:
The failure is from scrutinizer, we can ignore it.

PR #3652, opened 2015-11-18
Better error message.

nouiz:


theano/tensor/raw_random.py
```diff
@@ -379,10 +379,14 @@ def _infer_ndim_bcast(ndim, shape, *args):
...
-            ndim = tensor.get_vector_length(v_shape)
-        bcast = [False] * ndim
 
+    if v_shape.ndim != 1:
+        raise TypeError("shape must be a vector or list of scalar, got '%s'" %
+                        v_shape)
+
+    if ndim is None:
+        ndim = tensor.get_vector_length(v_shape)
+        bcast = [False] * ndim
```
lamblin: This line should be outside of the `if` block, otherwise `bcast` does not get defined.

nouiz: bcast have already been defined if this if is False. The logic isn't clear, but it prevent duplicate code. In the other if/ifelse then the else clause that I changed, bast is defined.

lamblin: Well, some tests are failing because `bcast` is used before being assigned.

nouiz:
@lamblin do you have the time to review this?

PR #3654, opened 2015-11-18
Better detect if we can use -lblas and more reuse of anaconda blas.

nouiz:


abergeron:
This looks ok to merge when travis passes.

theano/tensor/blas.py
```diff
@@ -312,8 +320,31 @@ def default_blas_ldflags():
...
+
+            if ((r - 30.f) > 1e-6 || (r - 30.f) < -1e-6)
+            {
+                return -1;
+            }
+            return 0;
+        }
+        """)
+    flags = ['-lblas']
+    flags.extend('-L'.join(theano.gof.cmodule.std_lib_dirs()))
```
f0k: Wait, this line is completely wrong. Just by coincidence it will work if `std_lib_dirs()` is empty. If there are one or more entries, then a) they are concatenated to a single large string with some `-L` in between, and b) they are appended to `flags` character by character. I'll send a PR.

PR #3656, opened 2015-11-19
Scan - Move costly checks from runtime to compilation

carriepl:


nouiz:
I'm merging now as @carriepl won't be allow to do the small clean up rename rapidly. But this should be done later.

avostryakov:
Can you say how much does this PR speed up scan method?

nouiz:
between 0 and 10% speed up in the case we timmed.

PR #3662, opened 2015-11-20
Allow optimizer_excluding=inplace to keep the move to GPU.

nouiz:


lamblin:
I cancelled Travis because the changes only affect the GPU.

PR #3664, opened 2015-11-21
Fixes cl.exe not being seen on Windows

and-colonna:
On Windows, in `theano.sandbox.cuda.dnn.dnn_available()`, `comp`, `out` and `err` evaluate to:

```python
False
b"nvcc fatal   : Cannot find compiler 'cl.exe' in PATH\r\n"
b''
```
So I found it necessary to add the `--compiler-bindir` flag to the source, in order to actually use CuDNN.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -69,6 +69,8 @@ if ((err = cudnnCreate(&_handle)) != CUDNN_STATUS_SUCCESS) {
                 params.append("-I" + config.dnn.include_path)
             if config.dnn.library_path:
                 params.append("-L" + config.dnn.library_path)
+            if config.nvcc.compiler_bindir:
+                params.extend(['--compiler-bindir', config.nvcc.compiler_bindir])
             # Do not run here the test program. It would run on the
             # default gpu, not the one selected by the user. If mixed
             # GPU are installed or if the GPUs are configured in


nouiz:
Thanks for the PR. To be sure, you add yourself the Theano flags to have this working? How did you got this value?

and-colonna:
Yes. I add some more information to provide some context.

Whenever I tried to use CuDNN (e.g. by using the `optimizer_including=cudnn` flag) I would get error messages like:

```
[…]
File “C:\Python34\lib\site-packages\theano\sandbox\cuda\dnn.py”, line 1688, in apply
dnn_available.msg)
AssertionError: cuDNN optimization was enabled, but Theano was not able to use it. We got this error:
Theano can not compile with cuDNN. We got this error:
b”
```

So I went and started looking into the code and found out that for some reason `dnn_available.avail` evaluated to `False`. I was mostly puzzled by that `nvcc fatal   : Cannot find compiler 'cl.exe' in PATH` message, as on my computer Theano can make use of nvcc without any issues. I thought that for some reason `theano.sandbox.cuda.dnn.dnn_available()` ignores the `compiler_bindir` flag. I looked at the code in `theano.sandbox.cuda.nvcc_compiler` where the `compile_str` method is defined for the `NVCC_compiler` class and found these lines:

```
if config.nvcc.compiler_bindir:
            cmd.extend(['--compiler-bindir', config.nvcc.compiler_bindir])
```

so I added very similar lines to the source as in the PR and then  `dnn_available.avail` started returning `True`.

nouiz:
thanks. I'll port this to the new GPU back-end in another PR later.

PR #3667, opened 2015-11-23
Enable cudnn for AbstractConv2d

nouiz:


theano/sandbox/cuda/dnn.py
```diff
@@ -2440,24 +2437,26 @@ def local_softmax_dnn_grad(node):
...
 def local_abstractconv_cudnn(node):
     inp1 = node.inputs[0]
     inp2 = node.inputs[1]
 
-    if ((not isinstance(node.op, AbstractConv2d) or
-         not isinstance(node.op, AbstractConv2d_gradWeights) or
-         not isinstance(node.op, AbstractConv2d_gradInputs))):
+    if (not isinstance(node.op, (AbstractConv2d,
+                                 AbstractConv2d_gradWeights,
+                                 AbstractConv2d_gradInputs))):
```
f0k: Oh, that's a fun one :)

f0k:
The fix looks good! I didn't check the two pep8 commits.

carriepl:
Looks ok.

PR #3673, opened 2015-11-23
Fix #3190

### LINKED ISSUE #3190: Ops creating variables without stack trace
This ticket is here to centralize a list of functions or Ops creating symbolic variables without a stack trace.
- `DimShuffle`. None of the following assign a stack trace to the output's `tag.trace`:
  - `x.T`
  - `x.dimshuffle(1, 0)`
  - `theano.tensor.transpose(x)`

Related to #3018.

Sentient07:
Fix #3190

### DIFF
--- a/theano/tensor/elemwise.py
+++ b/theano/tensor/elemwise.py
@@ -210,7 +210,7 @@ class DimShuffle(Op):
                 ob.append(ib[value])
 
         output = TensorType(dtype=input.type.dtype,
-                            broadcastable=ob).make_variable()
+                            broadcastable=ob)()
 
         return Apply(self, [input], [output])
 


nouiz:
The diff seem good. Did you check that it allow to see the stack trace?

Sentient07:
@nouiz Yeah, i've checked the error. The error is same as what i get in the master branch. The traceback : http://dpaste.com/0Q0Z2G0

nouiz:
I was talking about checking that using the method in that ticket now give a trace. I did this check like this just in case it is useful to know how to check this:
```
import theano
theano.tensor.matrix().T.tag.trace
```
thanks

PR #3676, opened 2015-11-23
Gpuarray abstractconv

abergeron:
This is still missing tests, but I'd like feedback on the code.


theano/sandbox/gpuarray/opt.py
```diff
@@ -854,6 +858,44 @@ def values_eq_approx(a, b):
...
+                    as_gpuarray_variable(node.inputs[0],
+                                         context_name=context_name),
+                    node.inputs[2])]
+
+
+# This will deal with ops that don't have an explicit transfer but
+# have one of their inputs on the GPU already and the other not on the
+# GPU (to avoid endlessly replacing things).
+@register_opt()
+@local_optimizer()([AbstractConv2d])
```
lamblin: Travis says "invalid syntax"

abergeron: There is indeed an extra set of parentheses, but that should be legal syntax I think.

Anyway, I'll fix it.

abergeron:
I fixed some issues and did the required rebase.

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1188,57 +1215,50 @@ def make_node(self, dy, sm):
...
+    if isinstance(node.op, AbstractConv2d_gradInputs):
+        shape = (inp2.shape[0], inp1.shape[1],
+                 node.inputs[2][0], node.inputs[2][1])
+        rval = dnn_gradinput(inp1, inp2, shape,
+                             border_mode=node.op.border_mode,
+                             subsample=node.op.subsample,
+                             conv_mode=conv_mode)
+    return [rval]
+
+conv_groupopt.register(local_abstactconv_cudnn)
```
lamblin: "abstract"

theano/sandbox/gpuarray/opt.py
```diff
@@ -780,77 +785,48 @@ def local_assert(node, context_name):
...
-    out.values_eq_approx = values_eq_approx
-    return [out]
-
-# Register this here so that it goes after 'local_gpu_conv'
+def local_error_convop(node, context_name):
+    assert False, """
+ConvOp does not work with the gpuarray backend.
+
+Use the new convolution interface to have GPU convolution working:
+theano.tensor.nnet.abstract_conv2d.conv2d()
```
lamblin: I really don't like this, "abstract_conv2d" is going to be renamed "abstract_conv" soon, and we should not have to support that backward compatibility.
I really think we should wait for `theano.tensor.nnet.conv2d` to point to the abstract conv before merging this.

theano/sandbox/gpuarray/opt.py
```diff
@@ -780,77 +785,48 @@ def local_assert(node, context_name):
...
+"""
+
+# This deals with any abstract convs that have a transfer somewhere
+@register_opt('fast_compile')
```
lamblin: We want that in `'fast_run'` as well, otherwise the graph does not run at all.

theano/sandbox/gpuarray/opt.py
```diff
@@ -780,77 +785,48 @@ def local_assert(node, context_name):
...
+                                   context_name=context_name)
+    inps[1] = as_gpuarray_variable(node.inputs[1],
+                                   context_name=context_name)
+    return [node.op(*inps)]
+
+
+# This will deal with ops that don't have an explicit transfer but
+# have one of their inputs on the GPU already and the other not on the
+# GPU (to avoid endlessly replacing things).
+@register_opt('fast_compile')
```
lamblin: Same here.

theano/sandbox/gpuarray/tests/test_abstractconv.py
```diff
@@ -0,0 +1,298 @@
...
+                                output_shape=o, subsample=s,
+                                verify_grad=True, mode=mode, device='gpu',
+                                provide_shape=provide_shape, border_mode=b,
+                                filter_flip=flip)
+            self.run_gradinput(inputs_shape=i, filters_shape=f,
+                               output_shape=o, subsample=s,
+                               verify_grad=True, mode=mode, device='gpu',
+                               provide_shape=provide_shape, border_mode=b,
+                               filter_flip=flip)
+
```
lamblin: The CPU CorrMM test should be there as well.

lamblin:
Otherwise, it looks reasonable, provided the tests run on a GPU with CUDNN.

nouiz:
This should fix a bug in the new back-end, so I would not make it wait for
another PR... That other PR could change what is in this PR.

lamblin:
Where is a bug fixed?
My problem is that I don't want users to follow these instructions, use `nnet.abstract_conv2d.conv2d` like the message tells, only for that to fail the next day because the file has been renamed, without any indication of what they should have done.

theano/sandbox/gpuarray/opt.py
abergeron: 'fast_run' and 'gpuarray' are defaults in this file.  I only specify 'fast_compile' because we want the lifting to work in that case.

theano/sandbox/gpuarray/tests/test_abstractconv.py
abergeron: I don't want to test that twice.  We should move the tests in sandbox/cuda to tensor/nnet and have the CpuCorrMM tests there.

abergeron:
There is a problem with convolution currently where it basically doesn't work if you don't use cudnn convolution directly.

This will make it somewhat work.

I don't mind waiting a bit for this, but the other PR that renames abstract_conv file (or at least the one that changes the interface of tensor.nnet.conv2d) should be done soon so that I can recommend something stable.

theano/sandbox/gpuarray/opt.py
nouiz: right.

abergeron:
This will need to wait until we merge the PR that changes the interface for theano.tensor.nnet.conv2d()

theano/sandbox/gpuarray/tests/test_abstractconv.py
lamblin: Then why is `test_cpu_conv` here?
Since the reference is currently CuDNN, we cannot move the tests outside of sandbox/cuda or sandbox/gpuarray yet.
If you really do not want to duplicate the CPU CorrMM tests, then let's move them from sandbox/cuda to here, and remove test_cpu_conv from sandbox/cuda as well.

abergeron: Alright I moved the CPU tests over.

lamblin:
For the record, the PR changing nnet.conv2d is #3721.

lamblin:
#3721 has been merged.

nouiz:
This PR need a rebase.

lamblin:
The optimizations are not properly registered, so all tests are failing because `AbstractConv2d` is still in the graph, even after a rebase.

theano/sandbox/gpuarray/tests/test_abstractconv.py
```diff
@@ -0,0 +1,326 @@
...
+    def test_cormm_conv(self):
+        if not dnn_available(test_ctx_name):
+            raise SkipTest(cuda.dnn.dnn_available.msg)
```
lamblin: That message should probably be imported from elsewhere.

lamblin:
Continued in #3737.

Issue #3659, opened 2015-11-20
Gradient on dnn.GpuDnnConv3d parameter raises an error that the input tensor should be 4D


jgwak:
I'm facing a strange issue regarding gradient on convolutional layer parameter using `GpuDnnConv3d`.

I took gradient of parameters using `T.grad(cost, W)` where `W` is a parameter of `GpuDnnConv3d`.
I expect theano to use `GpuDnnConv3dGradW` as defined in here:
https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/dnn.py#L638
I also expect the input to be 5D: (b, c, 0, 1, 2)

However, I get the following error:

```
ERROR (theano.gof.opt): Optimization failure due to: local_dnn_convw_output_merge
ERROR (theano.gof.opt): node: GpuElemwise{add,no_inplace}(GpuDnnConv3dGradW{algo='none', inplace=False}.0, GpuElemwise{mul,no_inplace}.0)
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/home/jgwak/.pyv/theano/local/lib/python2.7/site-packages/Theano-0.7.0-py2.7.egg/theano/gof/opt.py", line 1787, in process_node
    replacements = lopt.transform(node)
  File "/home/jgwak/.pyv/theano/local/lib/python2.7/site-packages/Theano-0.7.0-py2.7.egg/theano/sandbox/cuda/opt_util.py", line 121, in opt
    return maker(targ, *inputs)
  File "/home/jgwak/.pyv/theano/local/lib/python2.7/site-packages/Theano-0.7.0-py2.7.egg/theano/sandbox/cuda/dnn.py", line 2266, in local_dnn_convw_output_merge
    return [GpuDnnConvGradW()(*inputs)]
  File "/home/jgwak/.pyv/theano/local/lib/python2.7/site-packages/Theano-0.7.0-py2.7.egg/theano/gof/op.py", line 600, in __call__
    node = self.make_node(*inputs, **kwargs)
  File "/home/jgwak/.pyv/theano/local/lib/python2.7/site-packages/Theano-0.7.0-py2.7.egg/theano/sandbox/cuda/dnn.py", line 801, in make_node
    raise TypeError('img must be 4D tensor')
TypeError: img must be 4D tensor
```

As you can see in the error message, It did use `GpuDnnConv3dGradW` as expected.

[...]

PR #3678, opened 2015-11-24
Conv3d

nouiz:
fix gh-3659

This need to be ported to the new back-end, but I have difficulty as conv3d don't work well in it and the merge of alpha and output node don't seem to work well.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -33,7 +33,6 @@ from theano.tensor.nnet.abstract_conv2d import (AbstractConv2d,
                                                 AbstractConv2d_gradWeights,
                                                 AbstractConv2d_gradInputs)
 
-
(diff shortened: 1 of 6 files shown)


lamblin:
It seems OK to me, but I'd like @abergeron to take a look at the change removing `nd` from `alpha_merge` and `output_merge`.

bartvm:
I'm not entirely sure how to reproduce the crash when there is no g++ compiler. It happened for me when I hadn't accepted Xcode's T&C yet, but now that I have, I don't think there's a way to un-accept them... Either way, I think it's fine, and if it doesn't fix it, Travis for Blocks will let us know.

lamblin:
Nevermind, the same change had already been done in `gpuarray`, so this should be correct.

lamblin:
The issue with conv3d and alpha/beta may have been solved by #3726

PR #3679, opened 2015-11-24
Renamed max_pool_2d to pool_2d and  DownsampleFactorMax to Pool

### LINKED ISSUE #3592: Rename pooling functions and Ops on CPU
This is way too confusing to have to call `max_pool_2d(..., mode='average_inc_pad')` to perform average pooling.
It is also confusing to have an Op called `DownsampleFactorMax` in the resulting graph, that performs average pooling.

Rename them to pool_2d() and Pool for the Op. Do this for the grad op and 2nd grad.

in the file theano/tensor/signal/downsample.py and rename the file to pool.py

keep old name working to allow unpickle with warning for functions.

shabanian:
Fixed #3592 .


nouiz:
just a quick comment to tell that travis is failing.

shabanian:
Fixed #3592.

shabanian:
@vdumoulin

vdumoulin:
@shabanian Some tests are still failing.

shabanian:
Fixed #3592.

shabanian:
@vdumoulin

vdumoulin:
`theano/sandbox/cuda/dnn.py` still tries to import `downsample`, which is why tests are failing.

vdumoulin:
Alright, the tests pass! @nouiz, would you mind having a quick look just to confirm that the refactoring is like you intended it to be?

nouiz:
This seem good, but we need to run the GPU tests and to make it backward compatible.

To make it back-ward compatible, make file downsample.py and import the stuff from the new file there. Import them as the old name. Print a warning at the import of this file that tell to use the new path/name.

shabanian:
Hi,
I could not get why it is failing!

vdumoulin:
@nouiz would you mind having a quick look?

theano/sandbox/cuda/opt.py
```diff
@@ -1890,9 +1890,9 @@ def local_gpu_downsample_factor_max_grad(node):
 
 
 @register_opt()
-@local_optimizer([downsample.DownsampleFactorMaxGradGrad])
+@local_optimizer([pool.DownsampleFactorMaxGradGrad])
```
nouiz: I would rename DownsampleFactorMaxGradGrad to PoolGradGrad too.

theano/tensor/signal/downsample.py
```diff
@@ -1,1088 +1,57 @@
...
-        st0, st1 = self.st
-        pad_h = self.padding[0]
-        pad_w = self.padding[1]
-        img_rows = x.shape[-2] + 2 * pad_h
-        img_cols = x.shape[-1] + 2 * pad_w
-        inc_pad = self.mode == 'average_inc_pad'
-        sum_mode = self.mode == 'sum'
+    warnings.warn("function 'max_pool2D' is now located "
+                  "in 'pool'.")
+    return pool.max_pool2D(*args, **kwargs)
```
nouiz: Also make the old class name available here, to allow unpickling them. Example:

DownsampleFactorMax = Pool

Also raise a warning when this file is imported and tell people the should import the pool file.

theano/tensor/signal/downsample.py
```diff
@@ -1,1088 +1,57 @@
...
-        x = tensor.as_tensor_variable(x)
-        maxout = tensor.as_tensor_variable(maxout)
-        gz = tensor.as_tensor_variable(gz)
-        assert x.ndim == 4
-        assert maxout.ndim == 4
-        assert gz.ndim == 4
+def DownsampleFactorMax(*args, **kwargs):
+    warnings.warn("Class 'DownsampleFactorMax' has been renamed and is now located "
+                  "in 'pool.Pool'.")
+    return pool.Pool(*args, **kwargs)
```
nouiz: Do this work when unpickling old pickled file? I never used that pattern.

theano/sandbox/cuda/opt.py
nouiz: It currently only support max. So finaly, don't rename it now.

nouiz:
So mostly, raise a warning when the old file is imported. And we need to verify that the way the old class are supported work when unpickling old files. Did you test that?

when this is done, we need someone to run all Theano tests on a computer with a GPU to make sure the GPU tests also pass. Then we can merge.

thanks

theano/tensor/signal/downsample.py
vdumoulin: That's a good point, I'm not sure...

vdumoulin: So I guess the easiest solution to make sure everything works would be to raise a warning when the module is imported, not when the function/constructor is called, and to alias the function and class names.

lamblin: I don't think it would, since the `__init__` method is not called when unpickling, but other methods of the object are called, which do not exist in a function.

nouiz: Doing it when the class is useful to help user know where in there code
they do this. But this would also polute the output when unpickling. I
think you are right. I also prefer to do it only on import.

vdumoulin:
@shabanian Are you comfortable making the requested changes? I'll handle testing on GPU.

shabanian:
Sure! as a quick question, should I finish it today?

vdumoulin:
If you can, I'd appreciate it.

vdumoulin:
Looks good. Let's wait for the tests to pass. I'll test on GPU afterwards.

shabanian:
Thanks :)

vdumoulin:
The tests pass, I'll merge. Thank you very much! :)

Issue #3665, opened 2015-11-22
Error message in conv.py


dwf:
https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/conv.py#L871-L872

Two suggestions and a question:

* [ ] This is only raised when border_mode is not 'valid'; the error message should make mention of that so it's easier to debug
* [ ] "different from 1 or 2" would be better phrased as "greater than 2", assuming that dx and dy are elsewhere checked for > 0 (because <= 0 makes no sense whatsoever).

Finally, does this mean that one still cannot use stride > 2 convolution if you need to backprop?

JesseLivezey:
> Finally, does this mean that one still cannot use stride > 2 convolution if you need to backprop?

I think the cuDNN and gemm implementations should allow stride > 2. On cpu, there is a gemm implementation available through the abstract_conv interface/optimizations.

nouiz:
I have a PR: https://github.com/Theano/Theano/pull/3682

It mostly need link to the not yet done doc on the web site about
abstract_conv.

PR #3682, opened 2015-11-25
Better error message and docstring

nouiz:
fix gh-3665

I think this PR need an update to the error message to point to abstract_conv2d.

But I think we can merge now, people will search for it until this get done.


lamblin:
If pep8 is fixed, can you remove the file from the whitelist?

nouiz:
It wasn't in the whitelist. I fixed pep8 that the test tolerate like line too long.

nouiz:
docgen.py --test

git rm doc/crei2013/logreg_profile.txt
            extraopts = ['-W']

array(188.4...) # doctest: ELLIPSIS

>>> numpy.allclose(f(94.2), 188.4)
True

nouiz:
I updated this PR and I think it can be merged after review. I don't expect confict with other change.

doc/library/tensor/basic.txt
```diff
@@ -510,7 +510,7 @@ TensorVariable
     .. method:: diagonal(offset=0, axis1=0, axis2=1)
     .. method:: astype(dtype)
     .. method:: take(indices, axis=None, mode='raise')
-    .. method:: copy()
+    .. method:: copy() Return an new symbolic variable that copy the variable. Do not copy the tag.
```
abergeron: Do not copy -> Does not copy

abergeron: Return an new -> Return a new
that copy -> that is a copy of

theano/tensor/var.py
```diff
@@ -535,7 +535,10 @@ def take(self, indices, axis=None, mode='raise'):
 
     # COPYING
     def copy(self, name=None):
-        """Copy a variable and optionally assign a name."""
+        """Return a symbolic copy and optionally assign a name.
+
+        Do not copy the tags.
```
abergeron: Do not copy -> Does not copy

abergeron:
Looks good to merge except for a few typos.

nouiz:
I updated the PR.

PR #3695, opened 2015-11-26
Fix call to gpu_from_host

carriepl:


abergeron:
This makes things work, so ok.

Issue #3687, opened 2015-11-25
Test DownSampleFactorMax.out_shape as other framework use it


nouiz:
see: [theano-dev] third regression in 24 hours

Sentient07:
@nouiz , can I give it a try ? A little more details about this ticket please ?

nouiz:
You can take it. thanks!

In the file theano/tensor/signal/downsample.py there is a class
DownSampleFactorMax. It have a method out_shape. This ticket is about
adding test in the file .../signal/tests/test_downsample.py to test that
function. Is is used to compute the output of shape max pooling operation.
So mostly test that it compute the current result.

We are pretty sure it compute the value correctly, the goal is to also make
sure we don't break that interface as there is code outside Theano that use
it.

Sentient07:
@nouiz , i went through those two files. A quick confirmation, should the test be similar to how "DownsampleFactorMaxGradGrad" class is tested  ?

nouiz:
We don't want to test the op itself. This is already done. We want to test
just that method. If that don't answer the question, can you ask it
differently? I'm not sure what you want to know.

Sentient07:
Yeah, i think i get what you say. I need to make a call to the "out_shape" method and do an assert_allclose on what it returns ?

nouiz:
Yes. If you can test a few different schedule at the same time, it would be
great.

Le jeu. 26 nov. 2015 17:19, Ramana Subramanyam <notifications@github.com> a
écrit :

Sentient07:
@nouiz, i'm comparing the shape of numpy_output_val(https://github.com/Theano/Theano/blob/master/theano/tensor/signal/tests/test_downsample.py#L185 ) and the output of out_shape method. But the test seems to fail.. Shall I make a PR of the failed test so that it'd be easier for you to have a look at ?

nouiz:
If you want me to look, make a PR.

PR #3697, opened 2015-11-27
Testing the output shape

Sentient07:
The test gives an error while comparing the two shapes.

fix gh-3687

### DIFF
--- a/theano/tensor/signal/tests/test_downsample.py
+++ b/theano/tensor/signal/tests/test_downsample.py
@@ -193,8 +193,8 @@ class TestDownsampleFactorMax(utt.InferShapeTester):
 
                 # DownsampleFactorMax op
                 maxpool_op = DownsampleFactorMax(maxpoolshp,
-                                                ignore_border=ignore_border,
-                                                mode=mode)(images)
+                                                 ignore_border=ignore_border,
+                                                 mode=mode)(images)
 
                 output_shape = DownsampleFactorMax.out_shape(imval, maxpoolshp,
                                                         ignore_border=ignore_border)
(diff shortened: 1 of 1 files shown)


Sentient07:
#3687

theano/tensor/signal/tests/test_downsample.py
```diff
@@ -193,12 +193,21 @@ def test_DownsampleFactorMax(self):
 
                 # DownsampleFactorMax op
                 maxpool_op = DownsampleFactorMax(maxpoolshp,
-                                                 ignore_border=ignore_border,
-                                                 mode=mode)(images)
+                                                ignore_border=ignore_border,
+                                                mode=mode)(images)
```
nouiz: You changed the alignment of this line and the one above. This break the pep8 compliance and this checked by a test. So just undo the change on those 2 lines to fix one test failure.

Sentient07: Yeah, will do that. (Actually, i just ran this particular test alone, and since it failed, didn't run the entire test..)

theano/tensor/signal/tests/test_downsample.py
```diff
@@ -195,6 +195,14 @@ def test_DownsampleFactorMax(self):
                 maxpool_op = DownsampleFactorMax(maxpoolshp,
                                                  ignore_border=ignore_border,
                                                  mode=mode)(images)
+
+                output_shape = DownsampleFactorMax.out_shape(imval, maxpoolshp,
```
nouiz: I think you need to replace imval, by imval.shape.

Sentient07: Yeah, I got it, thanks!  :)

nouiz:
thanks

Issue #3663, opened 2015-11-21
BUG IN FGRAPH.REPLACE OR A LISTENER <type 'exceptions.Exception'> Constant{0} is already owned by another fgraph local_shape_to_shape_i


bourbakimember:
Hi,

Incurring this error when attempting to compile theano function. Function compiles with optimizer=fast_compile.

The function implements a for loop where i assume this error originates.

python version : Python 2.7.9 
theano version : 0.7.0.dev-6ca7b2b65000e371f009b617d41bc5a90f022d38

similar to ticket : https://github.com/Theano/Theano/issues/2976, although work around didn't seem to circumvent the issue.

Error begins:

<<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.Exception'> Constant\{0\} is already owned by another fgraph local_shape_to_shape_i\
ERROR (theano.gof.opt): Optimization failure due to: local_shape_to_shape_i\
ERROR (theano.gof.opt): node: Shape(for\{cpu,grad_of_scan_fn\}.61)\
ERROR (theano.gof.opt): TRACEBACK:\
ERROR (theano.gof.opt): Traceback (most recent call last):\
  File "/Users/Theano/theano/gof/opt.py", line 1822, in process_node\
    fgraph.replace_all_validate(repl_pairs, reason=lopt)\
  File "/Users/Theano/theano/gof/toolbox.py", line 306, in replace_all_validate\
    fgraph.replace(r, new_r, reason=reason, verbose=False)\
  File "/Users/Theano/theano/gof/fg.py", line 592, in replace\
    self.change_input(node, i, new_r, reason=reason)\
  File "/Users/Theano/theano/gof/fg.py", line 530, in change_input\
    self.__import_r__(new_r, reason=reason)\
  File "/Users/Theano/theano/gof/fg.py", line 363, in __import_r__\
    self.__import__(variable.owner, reason=reason)\
  File "/Users/Theano/theano/gof/fg.py", line 394, in __import__\
    raise Exception("%s is already owned by another fgraph" % r)\
Exception: Constant\{0\} is already owned by another fgraph\

Please find entire exception: [error.txt](https://github.com/Theano/Theano/files/40653/error.txt)

Thanks in advance.

nouiz:
Thanks for the report. Can you give us a way to reproduce this? This would help you generate such case:

http://www.deeplearning.net/software/theano/library/compile/function.html#theano.compile.function.function_dump

[...]

PR #3709, opened 2015-12-01
Fix BUG IN FGRAPH.REPLACE OR A LISTENER

nouiz:
Don't merge now, but it can be reviewed.

I need to check if there isn't extra scan in the graph. I'm not sure about that.

fix gh-3663

### DIFF
--- a/theano/scan_module/scan_op.py
+++ b/theano/scan_module/scan_op.py
@@ -1607,28 +1607,32 @@ class Scan(PureOp):
         inner_ins_shapes = []
         out_equivalent = OrderedDict()
 
+        # The two following blocks are commented as it cause in some
+        # cases extra scans in the graph. See gh-XXX for the
+        # investigation.
+
         # We skip the first outer input as it is the total or current number
         # of iterations.
         # sequences
         seqs_shape = [x[1:] for x in input_shapes[1:1 + self.n_seqs]]
-        inner_seqs = self.inputs[:self.n_seqs]
-        outer_seqs = node.inputs[1:1 + self.n_seqs]
-        for in_s, out_s in izip(inner_seqs, outer_seqs):
-            out_equivalent[in_s] = out_s[0]
+        # inner_seqs = self.inputs[:self.n_seqs]
+        # outer_seqs = node.inputs[1:1 + self.n_seqs]
+        # for in_s, out_s in izip(inner_seqs, outer_seqs):
+        #     out_equivalent[in_s] = out_s[0]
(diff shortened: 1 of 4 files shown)


carriepl:
I'm not familiar enough with this part of theano to review. @abergeron or @lamblin, can you have a look?

nouiz:
I did some check. This PR fix the crash, but will have 4 scan in my test
case instead of 2 at commit 146fbae8c4

@carriepl can you check why while the tests run with your modif to
DebugMode? Can you make a PR with your DebugMode change?

carriepl:
My changes to debugmode are in PR #3713. It might not be 100% ready to merge but it what I'm currently running to see the ops that would need changing to remove the checks in Scan.

nouiz:
I updated this PR.

lamblin:
OK to merge if Travis passes.

theano/scan_module/scan_op.py
```diff
@@ -1607,14 +1607,18 @@ def infer_shape(self, node, input_shapes):
...
         # of iterations.
         # sequences
-        seqs_shape = [x[1:] for x in input_shapes[1:1 + self.n_seqs]]
-        inner_seqs = self.inputs[:self.n_seqs]
-        outer_seqs = node.inputs[1:1 + self.n_seqs]
-        for in_s, out_s in izip(inner_seqs, outer_seqs):
-            out_equivalent[in_s] = out_s[0]
+        # seqs_shape = [x[1:] for x in input_shapes[1:1 + self.n_seqs]]
+        # inner_seqs = self.inputs[:self.n_seqs]
+        # outer_seqs = node.inputs[1:1 + self.n_seqs]
```
lamblin: These variables are still used below, so maybe there is more to comment out, or we should not comment out all of that.

nouiz:
Thanks for the comment. I updated this PR.

nouiz:
I amended the last commit.

nouiz:
I updated the tests in the PR to match the new behavior.

This trigger what I think is another not related bug in Scan that generate a crash. So I commented the execution of the problematic case. I think we should merge this PR as it leave the master in a better state then the current state of Theano and make an issue for the crash.

@carriepl you talked about a bug in scan that need fixing while working on the problematic PR. Was it this case?

nouiz:
IRL, @carriepl told me we can merge this PR. The failure of the new tests was there before and if we return not just the shape, but the real data, it crash. I'll create another issue for that.

PR #3712, opened 2015-12-01
Fix compilation crash

nouiz:


abergeron:

PR #3713, opened 2015-12-01
Add debug mode checks for reused outputs but altered data pointers

carriepl:


theano/compile/debugmode.py
```diff
@@ -1537,6 +1558,23 @@ def _check_preallocated_output(node, thunk, prealloc_modes, def_val,
...
+                    if init_output_vars[i] is var:
+                        if hasattr(var, 'data'):
+                            data = var.data
+                        elif hasattr(var, 'gpudata'):
+                            data = var.gpudata
+                        else:
+                            data = None
+
+                        if data != init_output_data[i]:
+                            _logger.warning("The node %s reuses its "
```
nouiz: We want to make this an error in another commit in this PR while we change the op that behave like this in the same PR. This will be prohibited from now on. It won't be just something that isn't optimal or acceptable but not optimal.

It is a good idea to help find node that do this to just warn about it. Just in case there is many of them in the same graph.

theano/compile/debugmode.py
```diff
@@ -2781,6 +2828,9 @@ def __init__(self,
             # Copy to avoid sharing the same list across different instances
             self.check_preallocated_output = check_preallocated_output[:]
 
+        if check_overwrite_preallocated_data is not None:
+            self.check_overwrite_preallocated_data = check_overwrite_preallocated_data
```
nouiz: Add an assert, if check_overwrite_preallocated_data is True, then check_preallocated_output must be activated.

theano/compile/debugmode.py
```diff
@@ -1523,13 +1531,26 @@ def _check_preallocated_output(node, thunk, prealloc_modes, def_val,
...
+            init_output_vars = []
+            init_output_data = []
             for r in node.outputs:
-                storage_map[r][0] = out_map.get(r, None)
+                var = out_map.get(r, None)
+                storage_map[r][0] = var
+
+                # Keep a reference to the initial output objects
+                # and data pointers for later comparison.
+                init_output_vars.append(var)
```
nouiz: Add this section in an "if check_data_ptr"

nouiz:
Except for my small comment, it seem good. Should we add in the same PR the change to remove that feature completly? I think we should wait for the result of the run before deciding this.

theano/compile/debugmode.py
```diff
@@ -1537,6 +1558,23 @@ def _check_preallocated_output(node, thunk, prealloc_modes, def_val,
...
+                    var = storage_map[r][0]
+                    if init_output_vars[i] is var:
+                        if hasattr(var, 'data'):
+                            data = var.data
+                        elif hasattr(var, 'gpudata'):
+                            data = var.gpudata
+                        else:
+                            data = None
+
+                        if data != init_output_data[i]:
```
nouiz: This should be ==

PR #3716, opened 2015-12-01
Fix DnnConv and its gradients to work with zero batch size. Fixes #3715

### LINKED ISSUE #3715: dnn_conv and gradients fail for zero image batch dimension
It seems cuDNN doesn't like the image batch dimension to be zero, meaning you get CUDNN_STATUS_BAD_PARAM errors. Note that a convolution (and its gradients) on an empty batch is perfectly well-defined.

If you're working with actual images then I imagine an empty batch rarely makes sense, but in our case we're using convolutions to do some NLP stuff, where we need to be able to handle zero-length "batches" in some cases without crashing.

```
RuntimeError: Could not set tensorNd descriptor: CUDNN_STATUS_BAD_PARAMdim=4
Apply node that caused the error: GpuDnnConv{algo='small', inplace=True}(GpuContiguous.0, GpuContiguous.0, GpuAllocEmpty.0, GpuDnnConvDesc{border_mode='valid', subsample=(1, 1), conv_mode='conv'}.0, Constant{1.0}, Constant{0.0})
Toposort index: 18
Inputs types: [CudaNdarrayType(float32, 4D), CudaNdarrayType(float32, 4D), CudaNdarrayType(float32, 4D), <theano.gof.type.CDataType object at 0x7fbb95d01c50>, Scalar(float32), Scalar(float32)]
Inputs shapes: [(0, 5, 6, 8), (3, 5, 5, 6), (0, 3, 2, 3), 'No shapes', (), ()]
Inputs strides: [(240, 48, 8, 1), (150, 30, 6, 1), (18, 6, 3, 1), 'No strides', (), ()]
Inputs values: [<CudaNdarray object at 0x7fbb95f99170>, 'not shown', <CudaNdarray object at 0x7fbb95fb7bb0>, <capsule object NULL at 0x7fbb95eeb870>, 1.0, 0.0]
Inputs name: ('image', 'kernel', 'output', 'descriptor', 'alpha', 'beta')

Outputs clients: [['output']]
```

mjwillson:
This checks for zero batch size and returns the appropriate result (zero-size and/or filled with zeros) in this case without calling cuDNN.

It seems cuDNN doesn't like zero dimensions in general, the call to `cudnnSetTensorNdDescriptor` via  `c_set_tensorNd` triggers the error, this is why some of the `c_set_tensorNd` calls needed to be moved further down.

fix gh-3715

### DIFF
--- a/theano/sandbox/cuda/dnn_gw.c
+++ b/theano/sandbox/cuda/dnn_gw.c
@@ -12,11 +12,6 @@ APPLY_SPECIFIC(conv_gw)(CudaNdarray *input, CudaNdarray *output,
     return 1;
   }
 
-  if (c_set_tensorNd(input, APPLY_SPECIFIC(input)) == -1)
-    return 1;
-  if (c_set_tensorNd(output, APPLY_SPECIFIC(output)) == -1)
-    return 1;
-
   int nb_dim = CudaNdarray_NDIM(output);
 
 #ifdef CONV_INPLACE
@@ -30,6 +25,20 @@ APPLY_SPECIFIC(conv_gw)(CudaNdarray *input, CudaNdarray *output,
     return 1;
 #endif
 
+
+  if (CudaNdarray_HOST_DIMS(input)[0] == 0) {
+    // zero batch size. Return a zero gradient without calling cuDNN (which will fail on zero batch size)
+    Py_XDECREF(*kerns);
+    *kerns = (CudaNdarray *) CudaNdarray_ZEROS(CudaNdarray_NDIM((*kerns)), (int *) CudaNdarray_HOST_DIMS((*kerns)));
+    if (*kerns == NULL) return 1;
+    return 0;
+  }
+
+
+  if (c_set_tensorNd(input, APPLY_SPECIFIC(input)) == -1)
+    return 1;
+  if (c_set_tensorNd(output, APPLY_SPECIFIC(output)) == -1)
+    return 1;
   if (c_set_filterNd(*kerns, APPLY_SPECIFIC(kerns)) == -1)
     return 1;
 
--- a/theano/sandbox/cuda/dnn_fwd.c
+++ b/theano/sandbox/cuda/dnn_fwd.c
@@ -12,8 +12,6 @@ APPLY_SPECIFIC(conv_fwd)(CudaNdarray *input, CudaNdarray *kerns,
     return 1;
   }
 
-  if (c_set_tensorNd(input, APPLY_SPECIFIC(input)) == -1)
-    return 1;
   if (c_set_filterNd(kerns, APPLY_SPECIFIC(kerns)) == -1)
     return 1;
 
@@ -30,7 +28,17 @@ APPLY_SPECIFIC(conv_fwd)(CudaNdarray *input, CudaNdarray *kerns,
     return 1;
 #endif
 
-   if (c_set_tensorNd(*output, APPLY_SPECIFIC(output)) == -1)
+  if (CudaNdarray_HOST_DIMS(input)[0] == 0) {
+    // if input batch is empty, we return the empty output without calling cuDNN
+    // (which will fail on zero batch size)
+    return 0;
+  }
+
+  if (c_set_tensorNd(input, APPLY_SPECIFIC(input)) == -1)
+    return 1;
+
+
+  if (c_set_tensorNd(*output, APPLY_SPECIFIC(output)) == -1)
      return 1;
 
   {
--- a/theano/sandbox/cuda/dnn_gi.c
+++ b/theano/sandbox/cuda/dnn_gi.c
@@ -12,8 +12,6 @@ APPLY_SPECIFIC(conv_gi)(CudaNdarray *kerns, CudaNdarray *output,
     return 1;
   }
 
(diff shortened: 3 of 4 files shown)


theano/sandbox/cuda/dnn_gw.c
```diff
@@ -30,6 +25,21 @@ APPLY_SPECIFIC(conv_gw)(CudaNdarray *input, CudaNdarray *output,
     return 1;
 #endif
 
+
+  if (CudaNdarray_HOST_DIMS(input)[0] == 0) {
+    // zero batch size. Return a zero gradient without calling cuDNN (which will fail on zero batch size)
+    Py_XDECREF(*kerns);
+    *kerns = (CudaNdarray *) CudaNdarray_ZEROS(CudaNdarray_NDIM((*kerns)), (int *) CudaNdarray_HOST_DIMS((*kerns)));
+    if (*kerns == NULL) return 1;
+    Py_INCREF(*kerns);
```
mjwillson: I wasn't sure whether I need the Py_INCREF here -- assumed I did given similar INCREF above on line 20, but wanted to check.

I also wasn't sure if it's OK to return a new zeros array in the case where CONV_INPLACE is specified, hopefully it is as it looked quite a pain in general to zero out an existing array in-place.

nouiz:
The code seem good. Can you add a test in theano/sandbox/cuda/tests/test_dnn.py to be sure we dno't break it in the futur?

mjwillson:
Thanks.  I did include a test this time!

theano/sandbox/cuda/dnn_gw.c
abergeron: The INCREF is not needed since you are creating a new object.  The one above is needed since we make a new reference to an existing object.

abergeron: Also to zero the array isn't there a memset() method for CudaNdarray.  I'm not sure though.

abergeron:
It would be really nice if you could do similar changes in the gpuarray backend. (under sandbox/gpuarray/).

theano/sandbox/cuda/dnn_gw.c
mjwillson: Aha, OK I'll remove the INCREF.

The nearest I saw to a memset method was `CudaNdarray_setitem` which can be called with "a python scalar with value 0 when the left side part is c contiguous", but fails otherwise. If I have to have an alternative branch anyway for the case where the existing in-place array isn't C-contiguous then I thought easier just to return a fresh array

mjwillson:
Just fixed the unneeded INCREF.

Sorry about lack of gpuarray equivalent, it's just a bit tricky to get up to speed with two different backends to make a fix when it's blocking research and you need to move on. Looking forward to when gpuarray can take over as the main GPU backend though!

mjwillson:
After rebasing this recently, I started getting a floating point exception segfault in the zero-batch case. I'm guessing something else changed in the relevant C code that clashes with this patch -- needs more investigation before merging.

nouiz:
thanks for the update.

lamblin:

nouiz:
@mjwillson do you still have segfault? Do you think this PR is ready?

mjwillson:
I'm afraid I haven't looked at it since I tried rebasing that last time. I was getting segfaults then and so I reverted the branch back to where it was. I'm not sure the segfault showed up in the test either :(

I'm now working on a custom kernel for my original use case (small max-over-time 1D convolutions) as CuDNN doesn't seem very optimised for these NLP use cases. So can't make any promises I'm afraid on when I'll get around to this

nouiz:
Did you tried cudnn v4rc1? It have optimization for some cases and maybe it
could covers yours.

If you end up with faster GPU code, it would be great to have them in
Theano.

thanks.

mjwillson:
I've not tried that no, is it supported in Theano now?
That said our code is a bit more specialised -- it takes advantage of the fact that the gradient from the max is sparse to avoid doing the full gradient computation for the convolution. It's also limited to particular ranges of input sizes. I'll see if we can share in some form if it ends up being useful though.

nouiz:
Theano support cudnn v4.

I'm not sure if we can use the information that the grad is sparse for
small windows and use it to speed up computation. I would think this make
the computation less regular, so harder to optimise. But I would love to be
prouved wrong:) Also, don't only compare speed again cudnn, as it use
assembler code that you probably don't do, so this isn't a completly fair
comparison.

mjwillson:
Yeah I'm hoping to prove you wrong :) Results so far are a lot faster, although I'm not sure how much of that's due to cudnn vs theano's max / max.grad being slow.

The memory access is less regular but not completely sparse, it's still able to read a bunch of contiguous chunks via coalesced reads for the filters in their argmax positions. I'm also making use of shared memory accumulators for gradients w.r.t. smaller inputs. Will see how it goes.

nouiz:
I think this is old stuff. So closing. If someone think it is still useful, tell why.

nouiz:
I tried the test case without the fix and the problem still exist. So reopening.

nouiz:
jenkins test this.

gvtulder:
The zero-batch-size problem should be fixed by PR #5267.

nouiz:
fixed by gh-5267

PR #3717, opened 2015-12-01
Misc fixes

abergeron:
Fix infer problems with optimizations.


nouiz:

nouiz:

PR #3718, opened 2015-12-01
Added config variable to disable initial testing of nvidia drivers

Erotemic:
Performing module tests on import causes a small overhead. It seems that running these tests is not required for Theano to properly initialize on systems where the tests would have passed. I added a configuration option to disable these tests which cuts import time by 20-30% for small test cases.

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -146,6 +146,13 @@ AddConfigVar(
     in_c_key=False)
 
 
+AddConfigVar(
+    'enable_initial_driver_test',
+    "Tests the nvidia driver when a GPU device is initialized.",
+    BoolParam(True, allow_override=False),
+    in_c_key=False)
+
+
 def default_cuda_root():
     v = os.getenv('CUDA_ROOT', "")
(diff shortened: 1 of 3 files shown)


nouiz:
Good idea. Can you add documentation about it in this file:

doc/library/config.txt

That way, people can find it.

Erotemic:
I went ahead and manually added that.

nouiz:
thanks

theano/configdefaults.py
```diff
@@ -146,6 +146,13 @@ def filter(val):
     in_c_key=False)
 
 
+AddConfigVar(
+    'enable_initial_driver_test',
+    "Tests the nvidia driver when a GPU device is initialized.",
+    BoolParam(True, allow_override=False),
+    in_c_key=False)
+
+
```
lamblin: Thinking about it, I would probably have put the option in the `cuda` section of the configuration, rather than at the root.
If you're willing to make that change before the flag gets used by other people, that would be cool, otherwise it's not that important.

Issue #3675, opened 2015-11-23
Abstract conv2d - Prepare transition


lamblin:
Here is a transition plan:
- rename `theano/tensor/nnet/abstract_conv2d.py` to `abstract_conv.py`
- Add a function `theano.tensor.nnet.conv2d`:
  - with the same prototype as `theano.tensor.nnet.abstract_conv.conv2d`
  - that also accepts the same parameters as `theano.tensor.nnet.conv.conv2d`, but with a warning when it differs from `abstract_conv.conv2d`
  - that calls `abstract_conv.conv2d`

Update the web documentation to make it clear that the new entry point is `nnet.conv2d`. That function could be defined in `__init__.py`, or in `abstract_conv.py` with a different name, not sure what is best.

f0k:
> Update the web documentation to make it clear that the new entry point is nnet.conv2d. That function could be defined in `__init__.py`, or in `abstract_conv.py` with a different name, not sure what is best.

*Maybe* we/you should even go one step further and create a `conv` subdirectory for the abstract convolution (`conv/abstract.py`), the gemm-based convolution and the legacy convolution, and the 3D convolution, ... there's quite a lot that's currently spread out over `blas.py`, `conv3d2d.py`, `Conv3D.py` and so on.
`conv/__init__.py` could house the entry points `conv2d` and `conv3d` then, which would be imported directly into the `theano.tensor.nnet` namespace (plus the others that need to live there for backward compatibility).

> that also accepts the same parameters as `theano.tensor.nnet.conv.conv2d`, but with a warning when it differs from `abstract_conv.conv2d`

You could also add those parameters and warnings to `abstract_conv.conv2d` (or `conv.abstract.conv2d`) and import this into the `T.nnet` namespace directly. What's the advantage of adding a separate wrapper instead of extending the current one? (Not a rhetorical question, I'd be interested in pros/cons.)

lamblin:
Regarding a `conv` subdirectory, that would be a good way of reorganizing the convolutions. However, I'm not sure if we want to do it right now, or start by replacing legacy convolution ops by better ones and deprecate the old ones first. That way, the burden of backward compatibility would be lower. Also:
- `theano.tensor.nnet.conv` currently points to where the legacy conv2d Op is defined, so there is a potential for confusion there
- most of `theano.tensor.nnet` is convolution stuff (except for sigmoid and softmax-related Ops), so it may be quite empty without `conv`

[...]

PR #3721, opened 2015-12-01
Cond2d interface update

ballasn:
Prepare the transition to abstract_conv


theano/tensor/nnet/__init__.py
```diff
@@ -30,3 +30,87 @@
...
+        are not flipped and the operation is referred to as a cross-correlation.
+
+    :rtype: symbolic 4D tensor
+    :return: set of feature maps generated by convolutional layer. Tensor is
+        of shape (batch size, output channels, output rows, output columns)
+    """
+
+    if len(kargs.keys()) > 0:
+        warnings.warn(str(kargs.keys()) +
+                      " are not deprecared in "
```
lamblin: "are now deprecated"

theano/tensor/nnet/__init__.py
```diff
@@ -30,3 +30,87 @@
...
+                      "`tensor.nnet.abstract_conv.conv2d` interface")
```
lamblin: Maybe add "and will be ignored".

theano/tensor/nnet/__init__.py
```diff
@@ -30,3 +30,87 @@
...
+
+    if image_shape is not None:
+        warnings.warn("image_shape is not supported in "
+                      "`tensor.nnet.abstract_conv.conv2d` interface"
+                      " use input_shape instead.")
+        if input_shape is None:
+            input_shape = image_shape
```
lamblin: if input_shape is not None, that should be an error (we should not provide both `image_shape` and `input_shape`)

theano/tensor/nnet/__init__.py
```diff
@@ -30,3 +30,87 @@
...
+    """
+
+    if len(kargs.keys()) > 0:
+        warnings.warn(str(kargs.keys()) +
+                      " are not deprecared in "
+                      "`tensor.nnet.abstract_conv.conv2d` interface")
+
+    if image_shape is not None:
+        warnings.warn("image_shape is not supported in "
+                      "`tensor.nnet.abstract_conv.conv2d` interface"
```
lamblin: Maybe "is no longer supported in `tensor.nnet.conv2d`".

lamblin:
Apart from that, that seems good, thanks!

lamblin:
Fixes #3675.

ballasn:
updated,

lamblin:
I cancelled the build for the last commit since it is only a change in a string, and the previous build was already well advanced.

theano/tensor/nnet/__init__.py
```diff
@@ -30,3 +30,91 @@
                    scalar_sigmoid, ultra_fast_sigmoid,
                    hard_sigmoid)
 from .bn import batch_normalization
+
+
+import warnings
+from .abstract_conv import conv2d as abstract_conv2d
+
+def conv2d(input, filters, input_shape=None, image_shape=None, filter_shape=None,
+           border_mode='valid', subsample=(1, 1), filter_flip=True,
```
f0k: This is not good, the order of arguments should be as for `conv2d()`. I.e., `image_shape` should be added in the end or scraped out of `**kwargs`, not interspersed with the others. Otherwise, existing code using positional arguments (such as `conv2d(x, y, x_shape, y_shape)`) will fail.

theano/tensor/nnet/__init__.py
```diff
@@ -30,3 +30,91 @@
...
+                      "`tensor.nnet.abstract_conv.conv2d` interface"
+                      " use input_shape instead.")
+        if input_shape is None:
+            input_shape = image_shape
+        else:
+            raise ValueError("input_shape and image_shape should not"
+                             " be provided at the same time.")
+
+    return abstract_conv2d(input, filters, input_shape, filter_shape,
+                           border_mode, subsample=(1, 1))
```
f0k: This is totally wrong. I'll file a PR.

lamblin: o.O Apparently, as we say in french, I didn't have my eyes aligned with my sockets when I reviewed this... Thanks so much for the fix.

f0k: It happens! There should have been a test noticing that strided convolutions with `conv2d()` stopped working, though.

Issue #3722, opened 2015-12-02
Wrong result when iterating over symbolic shapes


lamblin:
Some slices are not correctly handled when we try to extract all scalar elements of a symbolic shape after a subtensor:
```python
>>> x = theano.shared(np.zeros((2,3,4,5), dtype='float32'))
>>> list(x.shape)  # OK
[Subtensor{int64}.0,
 Subtensor{int64}.0,
 Subtensor{int64}.0,
 Subtensor{int64}.0]
>>> list(x.shape[2:4])  # OK
[Subtensor{int64}.0, Subtensor{int64}.0]
>>> list(x.shape[2:])  # BUG
[]
```
I think the previous behaviour was to raise an Exception in the last 2 cases.

lamblin:
Actually, Python 3 still raises an exception in the last 2 cases, because it does not consider `np.int64` as `numpy.Integrals`.

nouiz:
I have a PR: https://github.com/Theano/Theano/pull/3727

PR #3727, opened 2015-12-03
[BUG] fix get_vector length. fix gh-3722

nouiz:


theano/tensor/basic.py
```diff
@@ -4243,15 +4243,20 @@ def get_vector_length(v):
...
         start = extract_constant(theano.tensor.subtensor.get_idx_list(
             v.owner.inputs, v.owner.op.idx_list)[0].start)
         stop = extract_constant(theano.tensor.subtensor.get_idx_list(
             v.owner.inputs, v.owner.op.idx_list)[0].stop)
         if start is None:
             start = 0
+        ndim = v.owner.inputs[0].owner.inputs[0].ndim
         if stop is None:
-            stop = 0
+            stop = ndim
```
lamblin: This is still ignoring `step`, which may even be negative.

theano/tensor/basic.py
```diff
@@ -4243,15 +4243,20 @@ def get_vector_length(v):
...
+        elif isinstance(stop, numbers.Integral) and stop > ndim :
+            stop = ndim
         if ((isinstance(stop, numbers.Integral) and
              isinstance(start, numbers.Integral))):
```
lamblin: That condition does not work on python 3, since numpy.int64 is not in `numbers.Integral`.

theano/tensor/basic.py
```diff
@@ -4243,15 +4243,20 @@ def get_vector_length(v):
...
             return stop - start
```
lamblin: This is not correct if start or stop is negative.

nouiz:
I updated this PR by fixing and implementing the missing case.

theano/tensor/basic.py
```diff
@@ -4243,18 +4243,36 @@ def get_vector_length(v):
...
             v.owner.inputs, v.owner.op.idx_list)[0].stop)
+        step = extract_constant(theano.tensor.subtensor.get_idx_list(
+            v.owner.inputs, v.owner.op.idx_list)[0].step)
+
+        ndim = v.owner.inputs[0].owner.inputs[0].ndim
+        types = (numbers.Integral, numpy.integer)
         if start is None:
             start = 0
+        elif isinstance(start, types) and start < 0:
+            start += ndim
```
lamblin: If `start` is still `< 0` after that, we should set it to 0.

theano/tensor/basic.py
```diff
@@ -4243,18 +4243,36 @@ def get_vector_length(v):
...
+        elif isinstance(stop, types):
+            if stop > ndim:
+                stop = ndim
+            elif stop < 0:
+                stop += ndim
+        if step is None:
+            step = 1
+
+        if (isinstance(stop, types) and
+                isinstance(start, types) and
```
lamblin: and `isinstance(step, types)`

theano/tensor/basic.py
```diff
@@ -4243,18 +4243,36 @@ def get_vector_length(v):
...
+                start >= 0 and stop >= 0):
+            return (stop - start - 1) // step + 1
```
lamblin: Actually, I don't think that would work for `step < 0`, or when `start > stop`.

lamblin: There is some logic in https://github.com/Theano/Theano/blob/master/theano/tensor/subtensor.py#L107 but it is probably too complex if we just need the length.

nouiz:
I updated this PR

lamblin:
There are still cases not considered, but at least an error will be raised in those cases, so that is OK.

PR #3729, opened 2015-12-03
Fix errors in conv2d() -> abstract_conv.conv2d() wrapper

f0k:
Fixes errors recently merged with #3721:

1. The previous conv2d() argument order changed without reason
2. Arguments `subsample` and `filter_flip` were not passed on at all

I don't know how the previous version could have passed the tests, is `T.nnet.conv2d()` tested at all?


nouiz:
This look good. We can merge when travis pass.

lamblin:
I guess the old tests use `T.nnet.conv.conv2d` directly, and the new ones use `T.nnet.abstract_conv.conv2d`. We should have explicit tests, #3732.

PR #3734, opened 2015-12-03
Fix index error in scan backend

carriepl:


nouiz:
ok to merge when travis pass.

PR #3736, opened 2015-12-03
Update test for recent scipy.

lamblin:
This fixes a test when run with a recent version of scipy.
We still support the old cases, even if scipy dropped them.


nouiz:
ok to merge when travis pass.

PR #3737, opened 2015-12-04
Gpuarray abstractconv

lamblin:
Rebased version of #3676 with some additional fixes.
I've run the tests in `gpuarray/tests/test_abstractconv.py` locally on a TitanX and they pass, but I'll run them on the whole `gpuarray` overnight to be sure.


nouiz:
travis is failing:

```
  File "/home/travis/miniconda2/envs/pyenv/lib/python2.6/site-packages/theano/sandbox/gpuarray/opt.py", line 19, in <module>

    from theano.tensor.nnet.abstract_conv2d import (BaseAbstractConv2d,

ImportError: No module named abstract_conv2d
```

lamblin:
I think it is ready now.

theano/sandbox/gpuarray/opt.py
```diff
@@ -777,77 +782,49 @@ def local_assert(node, context_name):
...
+@register_opt('fast_compile')
+@local_optimizer([AbstractConv2d,
+                  AbstractConv2d_gradWeights,
+                  AbstractConv2d_gradInputs])
+def local_gpu_abstractconv2d(node):
+    if isinstance(node.op, BaseAbstractConv2d):
+        if ((isinstance(node.inputs[0].type, GpuArrayType) or
+             isinstance(node.inputs[1].type, GpuArrayType)) and
+            not (isinstance(node.inputs[0].type, GpuArrayType) or
+                 isinstance(node.inputs[1].type, GpuArrayType))):
```
nouiz: The second part of the condition " not (....)" seem useless. I think the total if will always be False. Did you wanted the not to be like this:

(not isinstance(...) or not isinstance(...))

abergeron: The second part prevents this optimization from always applying (or more specifically, from applying in a loop).

It looks for an abstractconv that has one input on the GPU and one that is not.  If both are on the GPU we don't want to do anything.

abergeron:
I also think this is ready, but I'll leave some time for @nouiz to respond.

PR #3696, opened 2015-11-27
Fix mask 101 GPU support in GpuCAReduce

adbrebs:
#2254 @abergeron @bouthilx

PR #3743, opened 2015-12-04
Reduce 101 newbackend

nouiz:
see also for the old back-end https://github.com/Theano/Theano/pull/3696

I did it in the op itself.


abergeron:
I'm not opposed to doing this, but I'll probably delete GpuCAReduceCuda soonish.

Anyway, the fix about params for perform() is reason enough to merge (as soon as travis passes).

There is a flake8 error in gpuarray/tests/test_elemwise.py.

nouiz:
I updated this PR.

PR #3747, opened 2015-12-05
Quick python op with gradient

### LINKED ISSUE #3094: Quick python op with gradient is difficult
For pymc3 it is important to easily specify gradients for custom ops. Currently this is a little complex. 

`as_op` is very nice but can't specify gradient. Inheriting from Op is good but requires specifying `make_node` which novice users  won't understand.

Would be lovely to be able to specify `itypes` and `otypes` when inheriting from Op. 

More here:
https://groups.google.com/forum/#!searchin/theano-users/simplest$20op$20with$20gradient$20function/theano-users/ByysGhDmep8/NOFR0uE6UL4J

Sentient07:
fix #3094

### DIFF
--- a/doc/tutorial/extending_theano.txt
+++ b/doc/tutorial/extending_theano.txt
@@ -67,8 +67,8 @@ possibilities you may encounter or need.  For that refer to
         #itypes and otypes attributes are
         #compulsory if make_node method is not defined.
         #They're the type of input and output respectively
-        itypes = []
-        otypes = []
+        itypes = None
+        otypes = None
 
         #Compulsory if itypes and otypes are not defined
         def make_node(self, *inputs):
@@ -265,7 +265,10 @@ Op Example
 
     import theano
 
-    class DoubleOp(theano.Op):
+    #Using make_node
+
+
+    class DoubleOp1(theano.Op):
         __props__ = ()
 
         def make_node(self, x):
@@ -297,11 +300,10 @@ Op Example
                 return eval_points
             return self.grad(inputs, eval_points)
 
-.. testcode:: example (using itypes and otypes)
+    #Using itypes and otypes
 
-    import theano
 
-    class DoubleOp(theano.Op):
+    class DoubleOp2(theano.Op):
         __props__ = ()
 
         itypes = [theano.tensor.dmatrix]
@@ -329,16 +331,48 @@ Op Example
 
 You can try it as follows:
 
-.. testcode:: example
+.. testcode:: example(Using make_node)
 
     x = theano.tensor.matrix()
-    f = theano.function([x], DoubleOp()(x))
+    f = theano.function([x], DoubleOp1()(x))
     import numpy
     inp = numpy.random.rand(5, 4)
     out = f(inp)
     assert numpy.allclose(inp * 2, out)
     print(inp)
     print(out)
+
+.. testoutput:: example
+   :hide:
+   :options: +ELLIPSIS
+
+   ...
+   ...
+
+.. code-block:: none
+
+    [[ 0.08257206  0.34308357  0.5288043   0.06582951]
+     [ 0.65977826  0.10040307  0.5402353   0.55472296]
+     [ 0.82358552  0.29502171  0.97387481  0.0080757 ]
+     [ 0.77327215  0.65401857  0.76562992  0.94145702]
+     [ 0.8452076   0.30500101  0.88430501  0.95818655]]
+    [[ 0.16514411  0.68616713  1.0576086   0.13165902]
+     [ 1.31955651  0.20080613  1.08047061  1.10944593]
+     [ 1.64717104  0.59004341  1.94774962  0.0161514 ]
+     [ 1.5465443   1.30803715  1.53125983  1.88291403]
+     [ 1.6904152   0.61000201  1.76861002  1.9163731 ]]
+
+.. testcode:: example (Using itypes and otypes)
+
+    x = theano.tensor.matrix()
+    f = theano.function([x], DoubleOp2()(x))
+    import numpy
+    inp = numpy.random.rand(5, 4)
+    out = f(inp)
+    assert numpy.allclose(inp * 2, out)
+    print(inp)
+    print(out)
+
 
 .. testoutput:: example
    :hide:
--- a/doc/tutorial/extending_theano.txt
+++ b/doc/tutorial/extending_theano.txt
@@ -64,6 +64,13 @@ possibilities you may encounter or need.  For that refer to
         # Properties attribute
         __props__ = ()
 
+        #itypes and otypes attributes are
+        #compulsory if make_node method is not defined.
+        #They're the type of input and output respectively
+        itypes = []
+        otypes = []
+
+        #Compulsory if itypes and otypes are not defined
         def make_node(self, *inputs):
             pass
 
@@ -98,7 +105,7 @@ possibilities you may encounter or need.  For that refer to
 .. ../extending/op.txt
 
 An op has to implement some methods defined in the the interface of
-:class:`gof.Op`. More specifically, it is mandatory for an op to define the method :func:`make_node` and one of the implementation methods, either :func:`perform`, :meth:`Op.c_code` or :func:`make_thunk`.
+:class:`gof.Op`. More specifically, it is mandatory for an op to define either the method :func:`make_node` or :attr:`itypes`, :attr:`otypes` and one of the implementation methods, either :func:`perform`, :meth:`Op.c_code` or :func:`make_thunk`.
 
   :func:`make_node` method creates an Apply node representing the application
   of the op on the inputs provided. This method is reponsible for three things:
@@ -174,6 +181,10 @@ An op has to implement some methods defined in the the interface of
   If :func:`make_thunk()` is defined by an op, it will be used by Theano
   to obtain the op's implementation.
   :func:`perform` and :meth:`Op.c_code` will be ignored.
+
+  If :func:`make_node` is not defined, the :attr:`itypes` and :attr:`otypes` 
+  are used by the Op's :func:`make_node` method to implement the functionality
+  of :func:`make_node` method mentioned above.
 
 Other methods can be optionally defined by the op.
 
@@ -265,6 +276,36 @@ Op Example
             assert hasattr(self, '_props'), "Your version of theano is too old to support __props__."
             x = theano.tensor.as_tensor_variable(x)
             return theano.Apply(self, [x], [x.type()])
+
+        def perform(self, node, inputs, output_storage):
+            x = inputs[0]
+            z = output_storage[0]
+            z[0] = x * 2
+
+        def infer_shape(self, node, i0_shapes):
+            return i0_shapes
+
+        def grad(self, inputs, output_grads):
+            return [output_grads[0] * 2]
+
+        def R_op(self, inputs, eval_points):
+            # R_op can receive None as eval_points.
+            # That mean there is no diferientiable path through that input
+            # If this imply that you cannot compute some outputs,
+            # return None for those.
+            if eval_points[0] is None:
+                return eval_points
+            return self.grad(inputs, eval_points)
+
+.. testcode:: example (using itypes and otypes)
+
+    import theano
+
+    class DoubleOp(theano.Op):
+        __props__ = ()
+
+        itypes = [theano.tensor.dmatrix]
+        otypes = [theano.tensor.dmatrix]
 
         def perform(self, node, inputs, output_storage):
             x = inputs[0]
--- a/theano/compile/ops.py
+++ b/theano/compile/ops.py
@@ -517,17 +517,7 @@ class FromFunctionOp(gof.Op):
 
     def __str__(self):
         return 'FromFunctionOp{%s}' % self.__fn.__name__
-
-    def make_node(self, *inputs):
-        if len(inputs) != len(self.itypes):
-            raise ValueError("We expected %d inputs but got %d." %
-                             (len(self.itypes), len(inputs)))
-        if not all(inp.type == it for inp, it in zip(inputs, self.itypes)):
-            raise TypeError(
-                "We expected inputs of types '%s' but got types '%s' " %
-                (str([inp.type for inp in inputs]), str(self.itypes)))
-        return theano.Apply(self, inputs, [o() for o in self.otypes])
-
+        
     def perform(self, node, inputs, outputs):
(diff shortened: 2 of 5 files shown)


nouiz:
It is the Op.make_node method that must be changed. We want to be able to have this work:
```
class MyOp(Op):
itypes = ...
otypes = ...
def perform(...):
   ....
def grad(...):
   ....
```

The problem is that implementing the make_node isn't trivial for everybody. So we want to allow some simple and frequence case to be done easily.

Sentient07:
@nouiz yeah, so that implementation which i've made in make_node and whatever has been done in that method should be moved into OP(theano/gof/op.py) right ? In that case, test should be added to theano/gof/tests/test_op.py making use of the as_op method in theano/compile/ops.py ?

Sentient07:
@nouiz also by this, are we making the make_node non-compulsory ?

nouiz:
That is exact. Now subclass of Op will have 2 choices, add itypes/otypes or make_node.

Sentient07:
Yet to add tests. Will add that in a while

Sentient07:
@nouiz have added tests. (sorry for the delay, was in-between my finals)

nouiz:
travis is failing. Can you check the problems?

theano/compile/ops.py
```diff
@@ -509,7 +509,7 @@ def __init__(self, fn, itypes, otypes, infer_shape):
             self.infer_shape = self._infer_shape
 
     def __eq__(self, other):
-        return (type(self) == type(other) and
+        return (isinstance(self, type(other)) and
```
nouiz: rever this change. this isn't equivalent and this is probably the cause of the crash in travis. We need the old behavior.

theano/gof/op.py
```diff
@@ -810,7 +835,8 @@ def __str__(self):
 
     def __eq__(self, other):
         if hasattr(self, '__props__'):
-            return (type(self) == type(other) and self._props() == other._props())
+            return (isinstance(self, type(other))
```
nouiz: idem revert.

theano/gof/tests/test_op.py
```diff
@@ -28,7 +29,7 @@ def __init__(self, thingy):
         self.thingy = thingy
 
     def __eq__(self, other):
-        return type(other) == type(self) and other.thingy == self.thingy
+        return isinstance(other, type(self)) and other.thingy == self.thingy
```
nouiz: revert.

theano/gof/tests/test_op.py
```diff
@@ -233,6 +235,28 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+
+            def perform(self, node, inputs, outputs):
+                input, = inputs
+                output, = outputs
+                output[0] = input + 1
+
+        x_input = T.fmatrix('x')
+        o = IncOne()(x_input)
+
+        # Confirming that make_node method is implemented
```
nouiz: I'm not sure this is needed. I don't understand what this help to test. Can you remove that test and replace it by compiling a Theano function that will output o and execute it to be sure that all the pipeline work?

nouiz:
This would also need an update to the doc in the file:

doc/tutorial/extending_theano.txt

thanks

theano/gof/tests/test_op.py
Sentient07: @nouiz , since this functionality should be working, i've written the test this way, without implementing the "make_node" method and just defining the itypes and otypes.
So, I should instead write the test to check if the operation performed on the "perform" method has updated the value in "output" variable ?

Sentient07:
@nouiz , a quick confirmation, in the documentation, I should be updating that make_node method is not compulsory any more and give an example of how extending is done without make_node method by just defining itypes and otypes, right ?

Sentient07:
@nouiz , a quick confirmation, in the documentation, I should be updating that make_node method is not compulsory any more and give an example of how extending is done without make_node method by just defining itypes and otypes, right ?

Sentient07:
@nouiz , Updated tutorial, fixed travis. Now, a merge conflict i guess, will check that.

doc/tutorial/extending_theano.txt
```diff
@@ -64,6 +64,13 @@ possibilities you may encounter or need.  For that refer to
         # Properties attribute
         __props__ = ()
 
+        #itypes and otypes attributes are
+        #compulsory if make_node method is not defined.
+        #They're the type of input and output respectively
+        itypes = []
+        otypes = []
```
nouiz: Make them default to None, to make it more clear that they aren't defined. There is some cases where you could have one op with no inputs and some outputs. Who know if we allow ops with no outputs in the futur (they could have have side effect for example)

theano/compile/ops.py
```diff
@@ -519,6 +519,13 @@ def __str__(self):
         return 'FromFunctionOp{%s}' % self.__fn.__name__
 
     def make_node(self, *inputs):
+
+        if not self.itypes:
+            raise NotImplementedError("itypes not defined")
```
nouiz: NotImplementedError normaly mean that Theano didn't implement that case. As this is a user error, I would think that returning an RuntimeError would be better. Also, I would give a longer error message like:

itypes not defined and make_nodes() not overrieded. In Op subclass, you must override itypes and otypes or overried make_node()

theano/compile/ops.py
```diff
@@ -519,6 +519,13 @@ def __str__(self):
...
+
+        if not self.otypes:
+            raise NotImplementedError("otypes not defined")
```
nouiz: idem

theano/gof/op.py
```diff
@@ -933,6 +961,23 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
         # condition: either there was no c_code, or it failed
         return self.make_py_thunk(node, storage_map, compute_map, no_recycling)
 
+    def make_node(self, *inputs):
+
+        if not hasattr(self, 'itypes'):
+            raise NotImplementedError("itypes not defined")
+
+        if not hasattr(self, 'otypes'):
```
nouiz: idem. I think the class FromFunctionOp have the exat same implementation as this one. As FromFunctionOp inherit from this op, you could remove the implementation in the class FromFunctionOp

theano/gof/tests/test_op.py
nouiz: The comment do not check if make_node is implemented. I don't understand what this try except is testing. The only missing tests I think of is to compile a Theano function and execute it and make sure it return the right value.

Sentient07: @nouiz , so i just need to assert if the output returned by the call to IncOne is the right value ?

Sentient07: @nouiz Earlier, since "make_node" method was compulsory to be implemented by all the class that inherits Op, I wrote a test that checks it's not compulsory any more if "itypes" and "otypes" are defined.

nouiz: Ok. But I still think you should compile a function with that op and verify that it compute correctly.

theano/gof/op.py
Sentient07: Done. Also, added test with compiling function and asserting.

nouiz: I mean, delete the make_node in the file compile/ops.py in the class FromFunctionOp. They can share the same code and this remove duplicate code.

thanks

Sentient07:
@nouiz  Yeah, I just saw that in travis. But I ran the test locally before pushing it and i wasn't able to see that error? The tests passes locally without changing it..

Sentient07:
@nouiz , yeah, logically dmatrix seemed correct. But i'm confused how it passes on my machine

nouiz:
travis test theano with many different configuration, not just the default configuration. To test of all them (and the extra one in our daily buildbot) use the script:

theano/misc/do_nightly_build 

in replacement of the command nosetests or theano-nose.

You will see, it run the tests in many combinations like with floatX=float32 and with mode=FAST_COMPILE.

nouiz:
The test fail due to the pep8:

/home/travis/build/Theano/Theano/theano/compile/ops.py:520:1: W293 blank line contains whitespace

Sentient07:
@nouiz I'm sorry, i forgot to run the flake test. Fixed that now

theano/gof/op.py
nouiz: review note, this is done.

doc/tutorial/extending_theano.txt
nouiz: Sorry for the long delay to review this. I got distracted by NIPS and vacations:)

I think this comment is the only thing left. Can you do it?

thanks

nouiz:
@abergeron @lamblin can you review this? I'm pretty sure it is good, it is mostly for your information and just in case I forgot something. This change the user interface.

doc/tutorial/extending_theano.txt
Sentient07: Sure. i was of the thought this was also addressed.

doc/tutorial/extending_theano.txt
```diff
@@ -286,6 +297,36 @@ Op Example
...
+
+        def R_op(self, inputs, eval_points):
+            # R_op can receive None as eval_points.
+            # That mean there is no diferientiable path through that input
+            # If this imply that you cannot compute some outputs,
+            # return None for those.
+            if eval_points[0] is None:
+                return eval_points
+            return self.grad(inputs, eval_points)
+
```
abergeron: testcode has to be paired with testoutput (like below).

theano/gof/op.py
```diff
@@ -714,7 +738,7 @@ def R_op(self, inputs, eval_points):
             "own op, implement the R_op method." %
             (self, self.__class__.__name__))
 
-    def perform(self, node, inputs, output_storage):
+    def perform(self, node, inputs, output_storage, params=None):
```
abergeron: Perhaps add the new parameter to the doc for the function?

theano/gof/op.py
```diff
@@ -512,7 +528,8 @@ def make_node(self, *inputs):
         MethodNotDefined : the subclass does not override this method.
```
abergeron: In the docstring above it says make_node is mandatory,  but this is no longer the case.

theano/gof/op.py
```diff
@@ -1360,8 +1408,9 @@ def get_c_macros(self, node, name, check_input=None):
         if check_input:
             # Extract the various properties of the input and output variables
             variables = node.inputs + node.outputs
-            variable_names = (["INPUT_%i" % i for i in range(len(node.inputs))] +
-                              ["OUTPUT_%i" % i for i in range(len(node.inputs))])
+            variable_names = (["INPUT_%i" %
+                               i for i in range(len(node.inputs))] + ["OUTPUT_%i" %
+                                                                      i for i in range(len(node.inputs))])
```
abergeron: I fail to see how this is formatted any better.

abergeron:
This sort of functionality pushes me more and more to have a metaclass on Op which would inject the required implementations if the specified attributes are present.

Otherwise it is too easy to try to use this functionality in a subclass of some Op and have it fail because it doesn't call the right make_node.

But anyway, that can be done in a later PR since it wouldn't change the user interface.

theano/gof/op.py
```diff
@@ -933,6 +961,23 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
         # condition: either there was no c_code, or it failed
         return self.make_py_thunk(node, storage_map, compute_map, no_recycling)
 
+    def make_node(self, *inputs):
+
+        if not hasattr(self, 'itypes'):
+            raise NotImplementedError("itypes not defined")
```
lamblin: Maybe mention in the error message that the user can either define `itypes` and `otypes`, or implement `make_node`.
I would be confused it I defined `def makenode(...)` instead of def `make_node(...)` and got an error message about `itypes` instead.

theano/gof/op.py
lamblin: Agreed, I think it is more confusing for several reasons.

lamblin:
Apart from that, it seems good. Thanks!

doc/tutorial/extending_theano.txt
Sentient07: @abergeron I didn't get you ? This and the code below are linked, and together yield the testoutput that's given below..

theano/gof/op.py
Sentient07: @abergeron I'm sorry, not able to recollect why i added this parameter. 
cc : @nouiz

theano/gof/op.py
Sentient07: @abergeron @lamblin I had used autopep8 for pep8 compliance. I'll revert them

theano/gof/op.py
Sentient07: Yeah, makes sense to me. Will change it

Sentient07:
@lamblin @abergeron @nouiz Have committed the changes. Couple of more clarifications (above) and i'll push it. Thanks :)

theano/gof/op.py
nouiz: It is probably a left over from rebase? Anyway, it is like this in the current master. If you rebase, it should go away.

doc/tutorial/extending_theano.txt
nouiz: @abergeron This is just a second example of the one above.
@Sentient07 I think that both case need to be tested. So can you duplicate the testoutput code?

theano/gof/op.py
Sentient07: @nouiz Yep, i guess I had once rebased because of a merge conflict and this could've emerged because of that.

doc/tutorial/extending_theano.txt
Sentient07: @nouiz Clarification, both case as in, one without itypes and otypes(with make_node) and one with itypes and otypes ?

nouiz: Yes, both testcode section need to get tested. So mostly,

testcode define op version one
testcode example
testoutput
testcode define op version 2
testcode example that use it
testoutput

Or it can be that you define the 2 class with different class name and test them together? This would show more clearly that it is equivalent I think.

Sentient07: @nouiz Yeah, the latter idea would be more efficient. Will make the changes

Sentient07: @nouiz Made the suggested changes, have a look at your convenience :)

nouiz:
@abergeron I'm ok with this. Can you double check the last commit? If good with you, we can merge.

PR #3750, opened 2015-12-05
added comments to logistic regression example in the documentation

andreh7:
I'm proposing a few additional comments in the logistic regression example (affecting theano/tutorial/examples.html ). No code changes.

### DIFF
--- a/doc/tutorial/examples.txt
+++ b/doc/tutorial/examples.txt
@@ -507,7 +507,7 @@ It will be used repeatedly.
     # between training iterations (updates)
     w = theano.shared(rng.randn(feats), name="w")
 
-    # initialize the bias (constant) term
+    # initialize the bias term
     b = theano.shared(0., name="b")
 
     print("Initial model:")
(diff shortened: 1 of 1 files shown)


doc/tutorial/examples.txt
```diff
@@ -489,16 +489,27 @@ It will be used repeatedly.
...
     y = T.vector("y")
+
+    # initialize the weight vector w randomly
+    #
+    # this and the following bias variable b
+    # are shared so they keep their values
+    # between training iterations (updates)
     w = theano.shared(rng.randn(feats), name="w")
+
+    # initialize the bias (constant) term
```
nouiz: Why did you put "(constant)" ? Otherwise, this PR is good to merge. thanks.

andreh7:
updated the branch, removed "(constant)" in commit 6253c08 . 

Originally I had put 'constant' since the result of the bias term is not proportional to any input, similar to the constant term in a polynomial. But I agree, the way it was written was a bit misleading, one could think that the bias term is fixed and not allowed to vary during training.

nouiz:
thanks

PR #3758, opened 2015-12-07
Add optimizer merging two switches of same condition

### LINKED ISSUE #3742: Opt: merge switch(cond, a, b) + switch(cond, c, d)
While playing with ELUs, I found that the gradient is a bit silly:
```python
import theano
T = theano.tensor
x = T.matrix('x')
# define forward expression
elu = T.switch(x > 0, x, T.exp(x) - 1)
# define backward expression
delta = T.matrix('delta')
grelu = theano.grad(None, x, known_grads={elu: delta})
# compile backward expression
fgrelu = theano.function([x, delta], grelu)
# print graph
from theano.printing import debugprint, pp
debugprint(fgrelu)
```
```
HostFromGpu [id A] ''   4
 |GpuElemwise{Composite{(Switch(i0, i1, i2) + Switch(i0, i2, (exp(i3) * i1)))}}[(0, 0)] [id B] ''   3
   |GpuElemwise{Composite{Cast{float32}(GT(i0, i1))},no_inplace} [id C] ''   2
   | |GpuFromHost [id D] ''   0
   | | |x [id E]
   | |CudaNdarrayConstant{[[ 0.]]} [id F]
   |GpuFromHost [id G] ''   1
   | |delta [id H]
   |CudaNdarrayConstant{[[ 0.]]} [id F]
   |GpuFromHost [id D] ''   0
```
Written a bit more concisely, it is:
```
switch(x > 0, delta, 0) + switch(x > 0, 0, exp(x) * delta)
```
Note that the switches share the same condition.

The following optimizer merges it nicely:
```python
local_merge_added_switch = theano.gof.PatternSub(
	in_pattern=(T.add, (T.switch, 'x', 'a1', 'b1'), (T.switch, 'x', 'a2', 'b2')),
	out_pattern=(T.switch, 'x', (T.add, 'a1', 'a2'), (T.add, 'b1', 'b2')),
	name='Merge addition of switch with same cond.', allow_multiple_clients=True)
T.opt.register_canonicalize(local_merge_added_switch, 'fast_run',
    name='Merge addition of switch with same cond.')
```

The gradient now reduces to:
```
HostFromGpu [id A] ''   4
 |GpuElemwise{Composite{Switch(i0, i1, (exp(i2) * i1))}}[(0, 0)] [id B] ''   3
   |GpuElemwise{Composite{Cast{float32}(GT(i0, i1))},no_inplace} [id C] ''   2
   | |GpuFromHost [id D] ''   0
   | | |x [id E]
   | |CudaNdarrayConstant{[[ 0.]]} [id F]
   |GpuFromHost [id G] ''   1
   | |delta [id H]
   |GpuFromHost [id D] ''   0
```
i.e.
```
switch(x > 0, delta, exp(x) * delta)
```

Should this be added to Theano? If so, where?

f0k:
This adds an optimizer merging switch(c, a, b) + switch(c, x, y) into switch(c, a+x, b+y), for all elemwise Ops of two inputs I could think of. Probably many of them are not practically relevant, but do you see any potential performance problems from implementing it this way? I.e., should we reduce this to a few relevant Ops, or implement it such that it looks for a `T.switch` first?

Closes #3742.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3806,6 +3806,29 @@ def local_div_switch_sink(node):
     return False
 
 
+# Merge add/sub/mul/div/minimum/maximum of two switches sharing the same
+# condition, to enable further simplification of their branches
+# Example: switch(c, a, b) + switch(c, x, y) -> switch(c, a+x, b+y)
+# TODO: generalize to all elemwise Ops? generalize to Ops with 3+ inputs?
+for _op_name in ('add', 'sub', 'mul', 'true_div', 'int_div', 'floor_div',
+                 'minimum', 'maximum', 'gt', 'lt', 'ge', 'le', 'eq', 'neq',
+                 'and_', 'or_', 'xor',
+                 'bitwise_and', 'bitwise_or', 'bitwise_xor', 'pow'):
+    _op = getattr(T, _op_name)
+    _opt_name = 'Merge %s of switch with same condition' % _op_name
+    _opt = gof.PatternSub(
+        in_pattern=(_op,
+                    (T.switch, 'c', 'a1', 'b1'),
+                    (T.switch, 'c', 'a2', 'b2')),
+        out_pattern=(T.switch, 'c',
+                     (_op, 'a1', 'a2'),
+                     (_op, 'b1', 'b2')),
+        name=_opt_name,
+        allow_multiple_clients=True)
+    register_canonicalize(_opt, 'fast_run', name=_opt_name)
+del _op_name, _op, _opt_name, _opt
+
+
 #############
 # Tile Opts #
 #############
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3806,27 +3806,30 @@ def local_div_switch_sink(node):
     return False
 
 
-# Merge add/sub/mul/div/minimum/maximum of two switches sharing the same
+# Merge add/sub/mul/div/minimum/maximum/... of switches sharing the same
 # condition, to enable further simplification of their branches
 # Example: switch(c, a, b) + switch(c, x, y) -> switch(c, a+x, b+y)
-# TODO: generalize to all elemwise Ops? generalize to Ops with 3+ inputs?
-for _op_name in ('add', 'sub', 'mul', 'true_div', 'int_div', 'floor_div',
-                 'minimum', 'maximum', 'gt', 'lt', 'ge', 'le', 'eq', 'neq',
-                 'and_', 'or_', 'xor',
-                 'bitwise_and', 'bitwise_or', 'bitwise_xor', 'pow'):
-    _op = getattr(T, _op_name)
-    _opt_name = 'Merge %s of switch with same condition' % _op_name
-    _opt = gof.PatternSub(
-        in_pattern=(_op,
-                    (T.switch, 'c', 'a1', 'b1'),
-                    (T.switch, 'c', 'a2', 'b2')),
-        out_pattern=(T.switch, 'c',
-                     (_op, 'a1', 'a2'),
-                     (_op, 'b1', 'b2')),
-        name=_opt_name,
-        allow_multiple_clients=True)
-    register_canonicalize(_opt, 'fast_run', name=_opt_name)
-del _op_name, _op, _opt_name, _opt
+@register_canonicalize
+@gof.local_optimizer([T.Elemwise])
+def local_merge_switch_same_cond(node):
+    scal = theano.scalar
+    # node must be binary elemwise or add or mul
+    if not isinstance(node.op, T.Elemwise) or not isinstance(
+            node.op.scalar_op, (scal.BinaryScalarOp, scal.Add, scal.Mul)):
+        return
+    # all inputs must be switch
+    if not all(s.owner and isinstance(s.owner.op, T.Elemwise) and
+               isinstance(s.owner.op.scalar_op, scal.Switch)
+               for s in node.inputs):
+        return
+    # all switch conditions must be the same
+    cond = node.inputs[0].owner.inputs[0]
+    if not all(s.owner.inputs[0] is cond for s in node.inputs[1:]):
+        return
+    # pull out switch
+    return [T.switch(cond,
+                     node.op(*[s.owner.inputs[1] for s in node.inputs]),
+                     node.op(*[s.owner.inputs[2] for s in node.inputs]))]
 
 
 #############
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -4914,6 +4914,29 @@ class test_local_useless_switch(unittest.TestCase):
                     isinstance(node.op, theano.tensor.Elemwise)]) == 0
 
 
+class test_merge_switch_same_cond(unittest.TestCase):
+    def test_elemwise(self):
+        # float Ops
+        mats = theano.tensor.matrices('cabxy')
+        c, a, b, x, y = mats
+        s1 = T.switch(c, a, b)
+        s2 = T.switch(c, x, y)
+        for op in (T.add, T.sub, T.mul, T.true_div, T.int_div, T.floor_div,
+                   T.minimum, T.maximum, T.gt, T.lt, T.ge, T.le, T.eq, T.neq,
+                   T.pow):
+            g = optimize(FunctionGraph(mats, [op(s1, s2)]))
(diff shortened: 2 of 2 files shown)


f0k:
You seem to have an old version of pep8 that's still reporting `E126 continuation line over-indented for hanging indent`. I've reduced the hanging indent by one level to circumvent it. Should pass this time.

f0k:
The optimization makes `test_local_div_switch_sink` fail (on `... resm[idx])).sum() == self.resm[idx].size`). Can't look into it right now.

f0k:
Okay, had a brief look. From the description of `local_div_switch_sink()` at https://github.com/f0k/Theano/blob/merge-switch-opt/theano/tensor/opt.py#L3745-L3749:
```
    This optimization makes the folowing changes in the graph:
    T.div(T.switch(cond,0,iff),A) -->  T.switch(cond,0,T.div(iff,A))
    T.div(T.switch(cond,ift,0),A) -->  T.switch(cond,T.div(ift,A),0)
    A being a symbolic variable.
```

But the test (https://github.com/f0k/Theano/blob/merge-switch-opt/theano/tensor/tests/test_opt.py#L4482-L4483) sets up a graph like this:
```
                y = T.true_div(T.switch(condition[0] > 0, 1. *
                    x[0], 0.*x[0]), T.switch(condition[0] > 0, 1.*x[0], T.log(c)*x[0]))
```
This will probably be simplified by my optimization rather than `local_div_switch_sink()`. I can see that the tests were originally done by @glorotxa in 5065772. What is the rationale of building a `div(switch(c,1,0), switch(c,1,A))` graph for `test_local_div_switch_sink()`, rather than some more direct `div(switch(c,1,0), A)`? Does anybody know?

f0k:
Any advice on this? What do we do about `test_local_div_switch_sink()`?

lamblin:
My guess is that the more complex test was closer to the real-life case that helped diagnose it. But I'm really not sure that the behaviour expected from that test is more correct than something else.
Changing that test case to something simpler (that would not trigger your optimization) is an option.
Disabling your optimization for that specific test would be another one.

f0k:
> My guess is that the more complex test was closer to the real-life case that helped diagnose it.

The question is whether my optimization would make things worse for that real-life case then. It's hard to tell what that was because the changes were committed directly to master (probably before Theano was even on github?).

I guess the test currently expects `switch(c,1,0) / switch(c,1,A)` to become `switch(c, 1 / switch(c,1,A), 0)`, while my optimization transforms it to `switch(c, 1/1, 0/A)`.

> Changing that test case to something simpler (that would not trigger your optimization) is an option.

Seems like the best way to me.

Apart from that, does my implementation look good? Is it okay to register a bunch of PatternSubs or should this become a single optimizer? I don't know enough about the optimization stage to see if this would cause any performance issues.

lamblin:
On Tue, Jan 26, 2016, Jan Schlüter wrote:
> The question is whether my optimization would make things worse for that real-life case then. It's hard to tell what that was because the changes were committed directly to master (probably before Theano was even on github?).

Indeed, the installation [instructions](https://github.com/Theano/Theano/blob/50657728b47be2466377efb12c31912b7bf5ad98/doc/install.txt#L68) from that time mention Assembla instead...

> I guess the test currently expects `switch(c,1,0) / switch(c,1,A)` to become `switch(c, 1 / switch(c,1,A), 0)`, while my optimization transforms it to `switch(c, 1/1, 0/A)`.

Right. In that case, this is not up to you to fix it: either 0/A should
also be replaced by 0 by another optimization, or we should remove
`local_div_switch_sink`.

My guess is that you did not get any error from `local_mul_switch_sink`
(which is really similar) because 0*A gets replaced by 0, but 0/A does
not.

> > Changing that test case to something simpler (that would not trigger your optimization) is an option.
> 
> Seems like the best way to me.

Let's do that then.

> Apart from that, does my implementation look good? Is it okay to register a bunch of PatternSubs or should this become a single optimizer? I don't know enough about the optimization stage to see if this would cause any performance issues.

Honestly, I have no idea how that would affect performance. Maybe @nouiz
has experience to share in that regard, but you may very well be the
first one to attempt that.
I hope using `profile=True,profile_optimizer=True` can help...

-- 
Pascal

nouiz:
Sorry for the long delay to reply. PatternSub is a good idea in theory, but the current execution engine is way to slow. I already did some work around for some current optimization. For this optimization, as it apply on Elemwise, I think not using PatternSub is better. Otherwise, it will cause too much extra overhead with the current implementation of PatternSub. I'll try to write a quick version in the diff of this PR.

theano/tensor/opt.py
```diff
@@ -3806,6 +3806,29 @@ def local_div_switch_sink(node):
     return False
 
 
+# Merge add/sub/mul/div/minimum/maximum of two switches sharing the same
+# condition, to enable further simplification of their branches
+# Example: switch(c, a, b) + switch(c, x, y) -> switch(c, a+x, b+y)
+# TODO: generalize to all elemwise Ops? generalize to Ops with 3+ inputs?
+for _op_name in ('add', 'sub', 'mul', 'true_div', 'int_div', 'floor_div',
```
nouiz: Code written to try to replace this without PatternSub:

```
@register_canonicalize
@gof.local_optimizer([T.Elemwise])
def local_merge_switch(node):
    if not isinstance(node.op T.Elemwise) or not isinstance(node.op.scalar_op, (scal.Add, scal.Sub, scal.Mul, scal.TrueDiv, scal.IntDiv, scal.FloorDir, ....):
        return
# all inputs must be switch
    if not all([i.owner and isinstance(i.owner.op, T.Elemwise) and isinstance(i.owner.op.scalar_op, scal.Switch) for i in node.inputs]):
        return
# all switch cond must be the same
    cond = node.inputs[0].owner.inputs[0]
    if not all([i is cond for i in node.inputs[1:]]):
        return
    return T.switch(cond,
                            node.op(*[s.owner.inputs[1] for s in node.inputs]),
                            node.op(*[s.owner.inputs[2] for s in node.inputs]))]
```

I think this version work for more then 2 inputs.

f0k:
> I'll try to write a quick version in the diff of this PR.

Thanks! I'll take your version and credit you if it works ;)

Pascal made an important observation though:

> Right. In that case, this is not up to you to fix it: either 0/A should also be replaced by 0 by another optimization, or we should remove `local_div_switch_sink`. My guess is that you did not get any error from `local_mul_switch_sink` (which is really similar) because 0*A gets replaced by 0, but 0/A does not.

Seems this is indeed the case. The test fails because it the `A` in `0/A` is `log(-1)`, so we get NaNs for my optimization, but not for the original one. There must be a specific reason the `local_div_switch` produces `0` instead of `0/A`, i.e., some use case that had to be covered.

Is there any reason why `0*A` should become `0`, but `0/A` should not? Note that `0*T.inv(A)` *does* become `0` in Theano. Otherwise I'd propose to add a `0/A -> 0` optimization. Quick test:
```
>>> import theano
>>> import theano.tensor as T
>>> x = T.scalar('x')
>>> theano.printing.debugprint(theano.function([x], 0*x))
DeepCopyOp [id A] ''   0
 |TensorConstant{0.0} [id B]
>>> theano.printing.debugprint(theano.function([x], 0/x))
HostFromGpu [id A] ''   2
 |GpuElemwise{TrueDiv}[(0, 1)] [id B] ''   1
   |CudaNdarrayConstant{0.0} [id C]
   |GpuFromHost [id D] ''   0
     |x [id E]
>>> theano.printing.debugprint(theano.function([x], 0*T.inv(x)))
DeepCopyOp [id A] ''   0
 |TensorConstant{0.0} [id B]
```

f0k:
> I'll try to write a quick version in the diff of this PR.

Works with minor fixes. Updated the code and test accordingly.

If someone adds a `0/x -> x` optimization, I can rebase and merge. Otherwise we'd still need to do something about `local_div_switch_sink`. Simply replacing the test by something that does not trigger my optimizer sounds dangerous to me, as this would remove a special case of `0/x -> x` that might have been important.

lamblin:
@f0k I agree with you, `0/x -> 0` should be added, and it would also help with #4086.
The timing is not great since we are between the RC and the final release, though, so I don't think we would be able to do that next week.

lamblin:
Closing and reopening to trigger a merge with the master after #4216 has been merged.

f0k:
Travis seems happy.

lamblin:
Then I'm happy too!

nouiz:
DLT tests pass.

PR #3769, opened 2015-12-11
Add __matmul__ method to tensor_py_operators 

emillynge:
in 3.5 the @ operator was added. This operator is specifically added to make matrix multiplications more clear. 
<http://legacy.python.org/dev/peps/pep-0465/>

This branch merely adds the `__matmul__` and `__rmatmul__` methods to the `_tensor_py_operators` class such that a two tensors can be be dotted with
```
AB = A @ B
BA = B @ A
```
instead of 
```
from tensor import dot
AB = dot(A, B)
BA = dot(B, A)
```
Hopefully improving code clarity

in accordance with the observed implementation of `__mul__`  the `__matmul__` method is supplied by  `tensor.dot` whilst `__rmatmul__` is supplied by `tensor.basic.dot`

### DIFF
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -158,7 +158,7 @@ class _tensor_py_operators(object):
         # See explanation in __add__ for the error catched
         # and the return value in that case
         try:
-            return theano.tensor.dot(self, other)
+            return theano.tensor.matmul(self, other)
         except (NotImplementedError, AsTensorError):
             return NotImplemented
 
@@ -244,7 +244,7 @@ class _tensor_py_operators(object):
         return theano.tensor.basic.sub(other, self)
 
     def __rmatmul__(self, other):
-        return theano.tensor.basic.dot(other, self)
+        return theano.tensor.basic.matmul(other, self)
 
     def __rmul__(self, other):
         return theano.tensor.basic.mul(other, self)
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -154,6 +154,14 @@ class _tensor_py_operators(object):
         except (NotImplementedError, AsTensorError):
             return NotImplemented
 
+    def __matmul__(self, other):
+        # See explanation in __add__ for the error catched
+        # and the return value in that case
+        try:
+            return theano.tensor.dot(self, other)
+        except (NotImplementedError, AsTensorError):
+            return NotImplemented
+
     def __mul__(self, other):
         # See explanation in __add__ for the error catched
         # and the return value in that case
@@ -234,6 +242,9 @@ class _tensor_py_operators(object):
 
     def __rsub__(self, other):
         return theano.tensor.basic.sub(other, self)
+
+    def __rmatmul__(self, other):
+        return theano.tensor.basic.dot(other, self)
 
     def __rmul__(self, other):
         return theano.tensor.basic.mul(other, self)
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5536,9 +5536,48 @@ def matmul(a, b):
     """
     Computes the matrix multiplication two tensor variables.
 
+    If any operands are scalar an error is raised.
+    If number of dimensions do not match and one operand is a vector, the vector
+    is promoted to a matrix before applying operation.
+        Note: As the exact shape of the operands cannot be determined at
+        compilation time, it is not possible to infer whether the vector should
+        be treated as a row- or column vector. Therefore _column_ vectors will
+        be assumed, if any inference must be made. such promoted vector will
+        have shape = [1, m] as is the default transform from 1D array to
+        np.matrix in the numpy library. *note that this is not strictly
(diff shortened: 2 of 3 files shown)


theano/tensor/var.py
```diff
@@ -154,6 +154,14 @@ def __sub__(self, other):
         except (NotImplementedError, AsTensorError):
             return NotImplemented
 
+    def __matmul__(self, other):
+        # See explanation in __add__ for the error catched
+        # and the return value in that case
+        try:
+            return theano.tensor.dot(self, other)
```
nouiz: I don't think this is right. I think numpy do something a little different. Can you add tests for many type of ndim inputs and compare that to that operator in numpy? The tests would run only in python 3.5.

emillynge: Hmmm...  It seems you are right, tensor does indeed have different behaviour than intended in PEP 465.
the @ operator covers the matmult operator in numpy, an operator that for all i know does not exist in theano atm.

from http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.matmul.html
```
Matrix product of two arrays.

The behavior depends on the arguments in the following way.

If both arguments are 2-D they are multiplied like conventional matrices.
If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.
If the first argument is 1-D, it is promoted to a matrix by prepending a 1 to its dimensions. After matrix multiplication the prepended 1 is removed.
If the second argument is 1-D, it is promoted to a matrix by appending a 1 to its dimensions. After matrix multiplication the appended 1 is removed.
Multiplication by a scalar is not allowed, use * instead. Note that multiplying a stack of matrices with a vector will result in a stack of vectors, but matmul will not recognize it as such.

matmul differs from dot in two important ways.

Multiplication by scalars is not allowed.
Stacks of matrices are broadcast together as if the matrices were elements.
```

There is however implemented functionality for N-D, N <= 3 as `basic.batched_dot` will work for N == 3
I will look into an implementation of tensor.basic.matmult that mirrors the numpy one

nouiz: thanks.

emillynge:
Update PR with matmul method defined in `tensor.basic` and changer the infix operator targets. Tests for the infix operator still pending.

emillynge:
So `test_good (theano.tensor.tests.test_basic.MatmulTester)` is failing. This is caused by expected function not being present in the numpy version.
the implemented method `tensor.basic.matmul` is expected to behave as `numpy.matmul`. `numpy.matmul` however was introduced in numpy `1.10` and since the python 3.3 tests are using numpy `1.9.1` the test cannot be run.

note that `tensor.basic.matmul` does not rely on any numpy  `1.10` functionality per se, only the test.

emillynge:
Just looked at this again. There might be a bit of work left.
the operation should work even for operands with different number of dimensions, which atm. raises an error.
numpy uses these shapes for matmul test cases:
```
            ((1, 1), (2, 1, 1)),     # broadcast first argument
            ((2, 1, 1), (1, 1)),     # broadcast second argument
            ((2, 1, 1), (2, 1, 1)),  # matrix stack sizes match
```
in effect first and second case, we repmat the 2D matrix to a 3D tensor to make the stack size match
Unfortunately i cannot seem to figure out whether this should also work also:
`            ((2, 1, 1), (4, 2, 1, 1)),     # broadcast first argument`

nouiz:
I didn't saw your last commit. When you do a new commit, this don't make a notification, so make a comment to let us know. Otherwise, it delay the review.

For your question in your last comment, the shape (2,1,1) and (4,2,1,1) should be broadcasted.

Theano broadcasting isn't the same as numpy. We do it based on the type of the variable when we build the graph, not based on the shape. This is probably caused you the problems with shapes you had.

thanks

nouiz:
Are you interrested to finish this? Many people start to use Python 3.5, so it would be good to have this. This PR need a rebase.

thanks

emillynge:
Yes, I will look into it again in the coming month. 

I have been busy finishing my masters thesis so the PR has been "shelved" for the time being. After i have defended the thesis, I should have some free time to rebase the PR and look into your comments from january.

BarclayII:
Wonder if someone is still maintaining this pull request, but the list of imports in the master `test_basic.py` conflicts that in the forked branch (`from theano.tensor import ...`).

This should be a quick fix.  Wondering how I could fix that.

emillynge:
Hi 

I have since gotten a job which takes up my time, and quite a bit of my mental capacity for coding. 
I would like to see this pr completed, but I can't even remember what outstanding tasks needs to be done. 

I think the snag was making sure that the matmul implementation mirrors numpys 100% which is a bit difficult bc it has so many 'modes' depending on the input, that even deciding the correct mode proved mind bending...

I can rebase the branch, but I will need help with completing the pr.

emillynge:
Hmmmm... It seems my rebase caused the PR to take ownership of a shit-ton of commits from upstream master :/

Should I do anything to fix this?

PR #3770, opened 2015-12-11
Fix the buildbot related script

nouiz:


nouiz:
killing and merging now as travis won't test the change in this PR.

PR #3772, opened 2015-12-12
Import pydot-ng with pydot as fallback.

gw0:
Added support for importing `pydot-ng` with `pydot` as fallback everywhere, shortened and standardized handling of this.

### DIFF
--- a/theano/d3viz/__init__.py
+++ b/theano/d3viz/__init__.py
@@ -1,11 +1,1 @@
 from theano.d3viz.d3viz import d3viz, d3write
-
-has_requirements = False
-try:
-    # pydot-ng is a fork of pydot that is better maintained
-    import pydot_ng as pd
-except ImportError:
-    # fall back on pydot if necessary
-    import pydot as pd
-if pd.find_graphviz():
-    has_requirements = True
--- a/theano/d3viz/__init__.py
+++ b/theano/d3viz/__init__.py
@@ -2,8 +2,10 @@ from theano.d3viz.d3viz import d3viz, d3write
 
 has_requirements = False
 try:
-    import pydot as pd
-    if pd.find_graphviz():
-        has_requirements = True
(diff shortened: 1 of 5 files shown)


theano/d3viz/__init__.py
```diff
@@ -2,8 +2,10 @@
...
-    if pd.find_graphviz():
-        has_requirements = True
+    # pydot-ng is a fork of pydot that is better maintained
+    import pydot_ng as pd
 except ImportError:
-    pass
+    # fall back on pydot if necessary
+    import pydot as pd
+if pd.find_graphviz():
+    has_requirements = True
```
nouiz: while cleaning this, could you rename has_requirement as the other pydot_imported for consistency?

nouiz:
thanks for the PR. Travis is failing in Python 3. I don't know why. Do you have an idea?

@abergeron do you have an idea why this PR cause problem in python 3? I'm restarting one of the run in python 3, just to be sure this isn't a random/temporary problem.

theano/d3viz/__init__.py
```diff
@@ -2,8 +2,10 @@
...
 try:
-    import pydot as pd
-    if pd.find_graphviz():
-        has_requirements = True
+    # pydot-ng is a fork of pydot that is better maintained
+    import pydot_ng as pd
 except ImportError:
-    pass
+    # fall back on pydot if necessary
+    import pydot as pd
```
nouiz: I think I just found the problem. pydot is an optional dependency. So this one can fail and need to be in another try...except. This is what make python 3 tests fail, as it don't have pydot of installed.

Also, I grepped in Theano and this version of the import isn't used. d3viz use the one in the file d3viz/formatting.py. So you can just delete the import of pydot from this file.

gw0:
You are right, this was an issue. I updated the PR.

nouiz:
thanks

PR #3773, opened 2015-12-12
Add support for CuDNN V4 convolution

carriepl:
This adds support for the new convolutions in CuDNN V4 but drops support for CuDNN V2


nouiz:
rebased at https://github.com/Theano/Theano/pull/3788

PR #3786, opened 2015-12-15
Abstract conv: fix tests and add infer_shape

### LINKED ISSUE #3732: Update some convolution tests to use `theano.tensor.nnet.conv2d`
Since `theano.tensor.nnet.conv2d` should be the main entry point for convolutions, we should have tests using it directly, and not only `theano.tensor.nnet.conv.conv2d` or `theano.tensor.nnet.abstract_conv.conv2d`.

lamblin:
This should actually finish #3582, and fix #3732.


nouiz:
I think we can merge this. @lamblin what is the result of the tests on the GPU?

lamblin:
@nouiz I did not think of saving the logs into a file, so the results were lost in the great reboot of last night. I launched them again.

nouiz:
On my tests, there is only problem with the new back-end. So I merge now.

Issue #3771, opened 2015-12-12
nvcc compilation problem


fayeshine:
In file included from /home/wenjian/anaconda2/include/python2.7/Python.h:8:0,
                 from mod.cu:1:
/home/wenjian/anaconda2/include/python2.7/pyconfig.h:1194:0: warning: "_POSIX_C_SOURCE" redefined [enabled by default]
 #define _POSIX_C_SOURCE 200112L
 ^
In file included from /usr/local/cuda/bin/..//include/host_config.h:161:0,
                 from /usr/local/cuda/bin/..//include/cuda_runtime.h:76,
                 from <command-line>:0:
/usr/include/features.h:228:0: note: this is the location of the previous definition
 # define _POSIX_C_SOURCE 200809L
 ^
In file included from /home/wenjian/anaconda2/include/python2.7/Python.h:8:0,
                 from mod.cu:1:
/home/wenjian/anaconda2/include/python2.7/pyconfig.h:1216:0: warning: "_XOPEN_SOURCE" redefined [enabled by default]
 #define _XOPEN_SOURCE 600
 ^
In file included from /usr/local/cuda/bin/..//include/host_config.h:161:0,
                 from /usr/local/cuda/bin/..//include/cuda_runtime.h:76,
                 from <command-line>:0:
/usr/include/features.h:169:0: note: this is the location of the previous definition
 # define _XOPEN_SOURCE 700
 ^
mod.cu(305): warning: subscript out of range
mod.cu(314): error: too few arguments in function call
1 error detected in the compilation of "/tmp/tmpxft_00000a04_00000000-9_mod.cpp1.ii".

[...]

PR #3788, opened 2015-12-16
Rebased cudnn v4

nouiz:
This is a rebase of gh-3773 on top of gh-3786.

The goal is to test all this together. I have started the GPU tests locally.

fix gh-3771


nouiz:
The tests are passing. I got some problems when fixing the  algo to deterministic, but this shouldn't not prevent to merge this PR.

PR #3790, opened 2015-12-16
set_subtensor GPU realizations

hidasib:


theano/sandbox/cuda/basic_ops.py
```diff
@@ -2710,7 +2710,7 @@ def make_node(self, x, ilist):
             raise TypeError('cannot index into a scalar')
 
         # c code suppose it is int64
-        if x.ndim in [2, 3] and ilist_.dtype in [
+        if x.ndim in [1, 2, 3] and ilist_.dtype in [
```
nouiz: review note, this is already tested. It don't request new tests.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3035,8 +3038,7 @@ def c_code_cache_version(self):
     def c_code(self, node, name, inputs, outputs, sub):
```
nouiz: As the c code was changed, we normally raise the value returned by c_code_cache_version. Can you do that? In the current state, it is to not do this, but it bring some inconsistency in the cache and I would prefer to not have this.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -3047,6 +3049,7 @@ def c_code(self, node, name, inputs, outputs, sub):
         out = outputs[0]
         fail = sub['fail']
         inplace = int(self.inplace)
+        set_instead_of_inc = int(self.set_instead_of_inc)
```
nouiz: I don't fins tests for this set_subtensor on the gpu. Could you add a test in the file sandbox/cuda/tests/test_basic_ops.py that test this?

Otherwise, this PR look great.
thanks.

hidasib:
I increased the c code cahce version from 6 to 7.

I added two tests. One is for the general case (ndim!=2) that uses the GpuAdvancedIncSubtensor1 operator. I tested this on a vector. The other test is for the ndim=2 when GpuAdvancedIncSubtensor1_dev20 is used. 

I ran these new tests and everything seems to be fine. (Also, I get the exact same results on CPU and GPU when running a complex code that uses these ops.)

nouiz:
thanks

PR #3792, opened 2015-12-17
Gpuarray prealloc

abergeron:
This adds an option to preallocate some memory on gpuarray devices.

It also fixes a leak of the handle for all cudnn ops.  This was causing out of memory error in the tests on smaller GPUs since each handle uses about 1M of GPU memory.

### DIFF
--- a/theano/sandbox/gpuarray/__init__.py
+++ b/theano/sandbox/gpuarray/__init__.py
@@ -39,11 +39,14 @@ register_transfer(transfer)
 
 
 def init_dev(dev, name=None):
-    if pygpu.gpuarray.api_version() != (-10000, 0):
-        raise RuntimeError("Wrong API version for gpuarray:",
-                           pygpu.gpuarray.api_version(),
+    v = pygpu.gpuarray.api_version()
+    if v[0] != -10000:
+        raise RuntimeError("Wrong major API version for gpuarray:", v[0],
                            "Make sure Theano and libgpuarray/pygpu "
                            "are in sync.")
+    if v[1] < 0:
+        raise RuntimeError("Wrong minor API version for gpuarray:", v[1],
+                           "Please update libgpuarray/pygpu.")
     global pygpu_activated
     if dev not in init_dev.devmap:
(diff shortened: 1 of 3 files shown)


abergeron:
Now with more Rebase™!

theano/sandbox/gpuarray/__init__.py
```diff
@@ -39,14 +39,26 @@ def transfer(x, target):
...
+    if v[1] < 0:
+        raise RuntimeError("Wrong minor API version for gpuarray:", v[1],
+                           "Please update libgpuarray/pygpu.")
     global pygpu_activated
     if dev not in init_dev.devmap:
-        init_dev.devmap[dev] = pygpu.init(dev)
+        ctx = pygpu.init(dev)
+        init_dev.devmap[dev] = ctx
+        if config.gpuarray.preallocate != 0:
+            if config.gpuarray.preallocate < 1:
```
nouiz: In the old back-end, if the value is between 0.98 and 1, I cap it to 0.98, as the driver need some spaces. I think it would be good to do the same here.

nouiz:
I'm merging now as this fix a significant problem for the tests. But when you have time, I think my comment should be fixed or discussed if you don't agree.

Also, some people needed to use 0.95 to have it work, but 0.98 work for most people.

gaebor:
Please document the `gpuarray.preallocate` configuration variable! It took me hours to get here and find out why `lib.cnmem` didn't work.

lamblin:
It is documented at http://deeplearning.net/software/theano_versions/dev/library/config.html#config.config.gpuarray.preallocate, but it could be made more clear.
See #5186.

Sentient07:
Also, the preallocate is only in the development version of the docs not the release version. Could that also be changed ?

nouiz:
Can you make a PR to Theano branch 0.8.X with that? As it was merged before the last release, we can include it there.

Sentient07:
I will try to today, if fails, will do it by tomorrow.

PR #3795, opened 2015-12-17
Sample multinomial with replacement

### LINKED ISSUE #3143: Add a new op for sampling elements randomly without replacement

aalmah:
This is intended to fix #3143

### DIFF
--- a/theano/sandbox/multinomial.py
+++ b/theano/sandbox/multinomial.py
@@ -6,13 +6,13 @@ import theano.tensor as T
 from theano.gof import local_optimizer
 from theano.tensor import NotScalarConstantError, get_scalar_constant_value
 from theano.scalar import as_scalar
+import sys
 
 from theano.sandbox.cuda import cuda_available, GpuOp
 if cuda_available:
     from theano.sandbox.cuda import CudaNdarrayType
     from theano.sandbox.cuda.basic_ops import host_from_gpu, gpu_from_host
     from theano.sandbox.cuda.opt import register_opt
-import sys
 
 
 class MultinomialFromUniform(Op):
--- a/theano/sandbox/multinomial.py
+++ b/theano/sandbox/multinomial.py
@@ -12,6 +12,7 @@ if cuda_available:
     from theano.sandbox.cuda import CudaNdarrayType
     from theano.sandbox.cuda.basic_ops import host_from_gpu, gpu_from_host
     from theano.sandbox.cuda.opt import register_opt
+import sys
 
 
 class MultinomialFromUniform(Op):
@@ -218,7 +219,10 @@ class WeightedSelectionFromUniform(Op):
         if unis.ndim != 1:
             raise NotImplementedError('unis ndim should be 1', unis.ndim)
         if self.odtype == 'auto':
-            odtype = pvals.dtype
+            if sys.maxsize > 2**32:
+                odtype = 'int64'
+            else:
+                odtype = 'int32'
         else:
             odtype = self.odtype
         out = T.tensor(dtype=odtype, broadcastable=pvals.type.broadcastable)
@@ -241,8 +245,12 @@ class WeightedSelectionFromUniform(Op):
             raise ValueError("unis.shape[0] != pvals.shape[0] * n_samples",
                              unis.shape[0], pvals.shape[0], n_samples)
 
+        if sys.maxsize > 2**32:
+            odtype = 'int64'
+        else:
+            odtype = 'int32'
         if z[0] is None or not numpy.all(z[0].shape == [pvals.shape[0], n_samples]):
-            z[0] = -1 * numpy.ones((pvals.shape[0], n_samples), dtype='int')
+            z[0] = -1 * numpy.ones((pvals.shape[0], n_samples), dtype=odtype)
 
         nb_multi = pvals.shape[0]
         nb_outcomes = pvals.shape[1]
--- a/theano/sandbox/multinomial.py
+++ b/theano/sandbox/multinomial.py
@@ -208,10 +208,6 @@ class WeightedSelectionFromUniform(Op):
 
     def __setstate__(self, dct):
         self.__dict__.update(dct)
-        try:
-            self.odtype
-        except AttributeError:
-            self.odtype = 'auto'
 
     def make_node(self, pvals, unis, n=1):
         pvals = T.as_tensor_variable(pvals)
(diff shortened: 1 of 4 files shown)


aalmah:
This is a preliminary version. I still need to add C code and some tests.

theano/sandbox/multinomial.py
```diff
@@ -193,6 +193,78 @@ def perform(self, node, ins, outs):
...
+
+    def __str__(self):
+        return '%s{%s}' % (self.__class__.__name__, self.odtype)
+
+    def __setstate__(self, dct):
+        self.__dict__.update(dct)
+        try:
+            self.odtype
+        except AttributeError:
+            self.odtype = 'auto'
```
carriepl: This try/except is not needed. All it does is ensure that the Op has an attribute `odtype` which is guaranteed by the fact that the `__init__()` method created that attribute and there has never been a version of this op without that attribute (since this op is new from your PR)

theano/sandbox/rng_mrg.py
```diff
@@ -1363,6 +1363,50 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
                                        " implemented for pvals.ndim = 2"))
 
+    def weighted_selection(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
+                           nstreams=None):
+        """
+        Sample `n` times *WITHOUT replacement* from a multinomial distribution
+        defined by probabilities pvals. `n` needs to be in [1, m], where m is the number of
+        elements to select from, i.e. m == pvals.shape[1]. By default n = 1.
+        
+        Example : WRITEME
```
carriepl: I agree there should be an example. I'm just putting a comment so I don't forget to look for the example in the final version of your PR.

theano/sandbox/rng_mrg.py
```diff
@@ -1363,6 +1363,50 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        
+        Notes
+        -----
+        -`size` and `ndim` are only there keep the same signature as other
+        uniform, binomial, normal, etc.
+        TODO : adapt multinomial to take that into account
```
carriepl: @nouiz : I'm guessing that multinomial() was never adapter in order to maintain the interface. Should we take this opportunity to not have these arguments in weighted_selection() or is it more important to maintain consistency with the signatures of uniform, binomial, normal, ...

theano/sandbox/tests/test_weighted_select.py
```diff
@@ -0,0 +1,70 @@
...
+    for rep in range(1000):
+        uni = numpy.random.rand(n_selected).astype(config.floatX)
+        res = f(pvals, uni, n_selected)
+        res = numpy.squeeze(res)
+        # print res
+        avg_pvals[res] += 1
+    avg_pvals /= avg_pvals.sum()
+    
+    print avg_pvals
+    print numpy.squeeze(pvals)
```
carriepl: These print statements are fine in Python 2.x but generate syntax errors in python 3.x (where print is now a function and can't simply be used as a keyword)

theano/sandbox/tests/test_weighted_select.py
```diff
@@ -0,0 +1,70 @@
...
+    numpy.random.seed(12345)
+    pvals = numpy.random.randint(1,100,(1,n_elements)).astype(config.floatX)
+    pvals /= pvals.sum(1)
+    avg_pvals = numpy.zeros((n_elements,))
+    
+    for rep in range(1000):
+        uni = numpy.random.rand(n_selected).astype(config.floatX)
+        res = f(pvals, uni, n_selected)
+        res = numpy.squeeze(res)
+        # print res
```
carriepl: this comment can be removed

theano/sandbox/tests/test_weighted_select.py
```diff
@@ -0,0 +1,70 @@
+import numpy
+from theano import config, function, tensor
+from theano.sandbox import multinomial
+
+def test_select_distinct():
+    p = tensor.fmatrix()
```
carriepl: Please add a quick comment at the beginning of each test simply to state what behavior the test seeks to validate.

theano/sandbox/tests/test_weighted_select.py
```diff
@@ -0,0 +1,70 @@
...
+        uni = numpy.random.rand(n_selected).astype(config.floatX)
+        res = f(pvals, uni, n_selected)
+        res = numpy.squeeze(res)
+        # print res
+        avg_pvals[res] += 1
+    avg_pvals /= avg_pvals.sum()
+    
+    print avg_pvals
+    print numpy.squeeze(pvals)
+    assert numpy.mean(abs(avg_pvals - pvals)) < mean_rtol
```
carriepl: (@aalmah  You already made a mention of this in your original PR comment). I just want to add a bit of precision to it)
This PR will need tests where the function weighted_selection() itself is tested to make sure that we don't get the case where the Op works but the function not.

theano/sandbox/tests/test_weighted_select.py
nouiz: It is bad practice to have print in tests. So if you can remove them this would be better.

carriepl: You're right. At first I thought it made sense them to have them because the tests are non-deterministic and, if they fail in travis/buildbot, we would want to know by how much... but they aren't deterministic; numpy.random is seeded in each test.

theano/sandbox/rng_mrg.py
nouiz: As currently all the method have this signature (and I think some of them also keep the extra param too from memory without using them) I would keep it like this.

theano/sandbox/tests/test_weighted_select.py
nouiz: If we want extra debug info, add it in the assert:

assert cond, extra_info

extra_info must be only 1 element, but can be a tuple of elements.

aalmah:
@carriepl fixed issues you raised. Please let me know if you can think of more tests

carriepl:
Travis complains about Flake8 issues

carriepl:
Plus, there are the issues raised by @nouiz in #3950. Some of these issues are not related to the C implementation so they are relevant in the current PR.

carriepl:
Looks good to me. I am merging even without the C implementation since we agreed it should be added in a separate pull request (#3950)

Issue #3789, opened 2015-12-16
str() needed around self.out_mode in conv.py


dwf:
```
  File "/home/dl/anaconda2/lib/python2.7/site-packages/theano/tensor/nnet/conv.py", line 149, in conv2d
    imshp=imshp, kshp=kshp, nkern=nkern, bsize=bsize, **kargs)
  File "/home/dl/anaconda2/lib/python2.7/site-packages/theano/tensor/nnet/conv.py", line 531, in __init__
    raise Exception("Mode %s not implemented" % self.out_mode)
TypeError: not all arguments converted during string formatting
```

nouiz:
Normally, self.out_mode should be a string. So that ConvOp instance got bad
parameter when it was created.

Yes, the error should get thrown correctly. But I think we should also fix
this upstream. Can you give us a way to reproduce this? The full stack
trace could also help.

dwf:
It's from someone attempting to use a tuple as the out_mode.

nouiz:
This is only supported in the new interface, not the old interface.

Do you call theano.tensor.nnet.conv2d() or theano.tensor.nnet.conv.conv2d()?

Do you still have the full stack trace? Was this when building the graph or
during optimization?

Issue #3765, opened 2015-12-10
More scan inter_shape (reapply it and investigate why extra scan get created)


nouiz:
gh-3580 introduced some scan enhancement that make it infer more shape.
gh-3709 fix a latent bug that now happened with gh-3580. But the use case for that now fixed bug caused extra scan.

This ticket is to enable what was reverted in gh-3709, use the use case (@nouiz have the file, crash_gh_3663_example2.pkl) that triggered the crash fixed by gh-3709 and investigate/fix why there is an extra scan. Maybe this extra scan is normal, we don't know.

PR #3797, opened 2015-12-18
Fix scan pushout opt. A GPU tests is already failing

nouiz:
fix gh-3789 also and the test

theano/scan_module/tests/test_scan.py:T_Scan_Cuda.test_gpu_memory_usage

The scan extra memory bug was created by commit https://github.com/Theano/Theano/commit/d3e30815459686 introduced by PR https://github.com/Theano/Theano/pull/3709

### DIFF
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -154,6 +154,7 @@ class FunctionGraph(utils.object2):
         self.inputs = list(inputs)
         self.outputs = outputs
 
+        # Why add feature before import of the graph?
         for f in features:
             self.attach_feature(f)
         self.attach_feature(toolbox.ReplaceValidate())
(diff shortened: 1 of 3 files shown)


carriepl:
I approve of this commit. If Travis is happy, so am I.

lamblin:

nouiz:
I updated the PR.

nouiz:
Don't merge, this don't fix anymore the gpu memory test.

nouiz:
I updated the PR with a different fix. I think this fix all known problems, the extra memory reported by a GPU test and the Constant in another FunctionGraph if we enable github.com/Theano/Theano/issues/3765.

This don't enable it gh-3765 as it need more tests. As this fix current problem, I think we can merge it if travis pass and a local checkout pass all Theano tests.

carriepl:
Some scan tests are failing in Travis

abergeron:
There is a working fix for the scan test is in #4125.  However, the other fixes that this contains aren't in there.

nouiz:
The good part have been ported to this PR, https://github.com/Theano/Theano/pull/4129

PR #3801, opened 2015-12-21
Doc clarification and skip flake8 tests on auto saved file

nouiz:


theano/tensor/signal/downsample.py
```diff
@@ -80,13 +80,16 @@ def max_pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
...
-                      stacklevel=2)
+        warnings.warn(
+            "max_pool_2d() will have the parameter ignore_border"
+            " default value changed to True (currently"
+            " False). To have consistent behavior with all Theano"
+            " version, explicitly add the parameter ignore_border=True."
+            " On the GPU, using ignore_border=False allow to use cudnn."
+            " Without ignore_border=True and cudnn, the only GPU combination"
+            " supported is when the `ds == st and padding == (0, 0) and"
+            " mode == 'max'`",
```
carriepl: On the GPU, using ignore_border=False is needed to use CuDNN. When using ignore_border=False and not using CuDNN, the only GPU [...]. Otherwise, the convolution will be executed on CPU.

nouiz: done.

carriepl:
I agree with the changes but now it's Travis that is unhappy (flake8 tests are failing). As soon as it's fixed and Travis is happy, anyone can merge.

nouiz:
I updated the PR.

PR #3802, opened 2015-12-21
Fix fallback in BLAS detection

f0k:
Fixes broken implementation of #3654.
Before, it would try `"-lblas f i r s t / d i r e c t o r y - L s e c o n d / d i r e c t o r y"`.
With this fix, it will try `"-lblas -Lfirst/directory -Lsecond/directory"`.
Should work better as a fallback...

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -340,7 +340,7 @@ SOMEPATH/Canopy_64bit/User/lib/python2.7/site-packages/numpy/distutils/system_in
         }
         """)
     flags = ['-lblas']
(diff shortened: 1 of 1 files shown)


f0k:
It works, BLAS is detected correctly again for our Lasagne test suite (both under Python 2.7 and 3.4, if it matters). Good to merge from my side. Maybe it will also speed up your Travis runs if you experienced any degradation after #3654.

nouiz:
It seem good to me. We can merge when travis fail.

We use miniconda in travis and not the OS blas. That is why we didn't got problem with this but you had problems.

thanks

nouiz:
I'm merging most of travis passed. thanks

PR #3803, opened 2015-12-21
Enable cudnn by default when avaialable in the new back-end

nouiz:
Otherwise, changing device=gpu to device=cuda disable the convolution.


carriepl:
This looks fine. Merging.

PR #3804, opened 2015-12-21
[BUG] in python code and tests fix

nouiz:
This fix a bunch of problem in the daily buildbot in DebugMode. The bug isn't expected in the normal user case and this is only for a new feature that you probably don't use.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -199,7 +199,11 @@ class Scalar(Type):
                 type(data), data, self.dtype), e)
 
     def values_eq_approx(self, a, b, tolerance=1e-4):
-        return abs(a - b) <= ((abs(a) + abs(b)) * tolerance)
+        # The addition have risk of overflow especially with [u]int8
+        diff = a - b
+        if diff == 0:
+            return True
+        return abs(diff) <= ((abs(a) + abs(b)) * tolerance)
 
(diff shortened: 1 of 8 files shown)


theano/scalar/basic.py
```diff
@@ -199,7 +199,11 @@ def filter(self, data, strict=False, allow_downcast=None):
                 type(data), data, self.dtype), e)
 
     def values_eq_approx(self, a, b, tolerance=1e-4):
-        return abs(a - b) <= ((abs(a) + abs(b)) * tolerance)
+        # The addition have risk of overflow especially with [u]int8
+        diff = a - b
+        if diff == 0:
+            return True
+        return abs(diff) <= ((abs(a) + abs(b)) * tolerance)
```
carriepl: There might still be an overflow if the difference is non-zero. What about : 
return abs(diff) <= (abs(a) * tolerence) + (abs(b) * tolerance)
Since tolerance is likely to be floating point, I think this should be fine.

carriepl:
Other than the point I raised, I think this is fine.

theano/scalar/basic.py
nouiz: I like that, I updated the PR.

carriepl:
Once Travis is happy, we can merge.

PR #3817, opened 2015-12-25
Check home_dir writing permission and assign temporary directory

viirya:
I ran Theano on a distributed environment. On worker nodes, it might possibly we don't have writing permission for the returned dir of `get_home_dir`. In this case, the running will be failed. I think we can assign it to temporary directory in this case.

### DIFF
--- a/theano/gof/compiledir.py
+++ b/theano/gof/compiledir.py
@@ -239,6 +239,10 @@ else:
     home_dir = get_home_dir()
     if not os.access(home_dir, os.W_OK):
         home_dir = '/tmp'
+        user = os.getenv('USER')
+        if user is not None:
+            home_dir = os.path.join(home_dir, user)
+
     default_base_compiledir = os.path.join(home_dir, '.theano')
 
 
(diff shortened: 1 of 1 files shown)


theano/gof/compiledir.py
```diff
@@ -236,7 +236,10 @@ def get_home_dir():
 if sys.platform == 'win32' and os.getenv('LOCALAPPDATA') is not None:
     default_base_compiledir = os.path.join(os.getenv('LOCALAPPDATA'), 'Theano')
 else:
-    default_base_compiledir = os.path.join(get_home_dir(), '.theano')
+    home_dir = get_home_dir()
+    if not os.access(home_dir, os.W_OK):
+        home_dir = '/tmp'
```
nouiz: I would use /tmp/$USER. Otherwise, there will be problems if there is multiple user on the same computer. The cache logic suppose the current user own the cache.

Otherwise, it seem a good idea.

viirya: Thanks. I've updated to use $USER if possibly.

nouiz:
Ok for me. Waiting for travis to pass, just in case.

PR #3057, opened 2015-06-22
Add exception when using an old version of pydot

diogo149:
addressing the issue here: https://github.com/Theano/Theano/issues/2988

PR #3819, opened 2015-12-27
Fix Python3 regression in ``pydotprint``

superbobry:
In Python3 ``map`` returns a map-object (and not a list). This makes the check
  introduced in #3057 fail with "TypeError: unorderable types: map() < list()"


nouiz:
thanks

PR #3822, opened 2015-12-28
Checks compiledir flag for prohibited characters. Fixes #3711

### LINKED ISSUE #3711: libcuda_ndarray.so: cannot open shared object file
I was trying to get the theano up and running on computation grid and I always got following error after import

    >>> import theano
        ....
        libcuda_ndarray.so: cannot open shared object file 

After some googling I found following thread on google groups:
https://groups.google.com/forum/#!msg/theano-users/F0AuGH3g1CY/Sm0Rphqbp_8J

The issue is on Linux if ```config.compiledir``` flag contains colon (:) character. The compiler then cannot find the correct location of the compiled library. In my case it was really obscure bug, as the ```%(platform)s-%(processor)s``` substitution keys from ```config.compiledir_format``` contained colon as part of the grid's platform description.

The workaround is easy - set the compiledir to some safe value.
The fix should be easy as well, just sanitize the compiledir path.

mronian:
fix gh-3711

### DIFF
--- a/theano/gof/compiledir.py
+++ b/theano/gof/compiledir.py
@@ -142,6 +142,11 @@ def short_platform(r=None, p=None):
     sr = '-'.join(sp)
     p = p.replace(r, sr)
 
+    p = p.replace(':', '-')
+    p = p.replace('?', '-')
+    p = p.replace('*', '-')
+    p = p.replace('/', '-')
+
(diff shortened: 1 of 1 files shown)


nouiz:
thanks

PR #3824, opened 2015-12-29
Added test for get_conv_output_shape method

### LINKED ISSUE #3686: Add unit tests for get_conv_output_shape
`theano.tensor.nnet.abstract_conv.get_conv_output_shape` is only tested indirectly, via the functions that call it. We should also have direct tests.

Sentient07:
fix #3686

### DIFF
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -1,0 +1,18 @@
+import unittest
+from theano.tensor.nnet.abstract_conv import get_conv_output_shape
+
+
+class TestGetConvOutShape(unittest.TestCase):
+
+    def test_basic(self):
+        image_shape, kernel_shape = (3, 2, 8, 8), (4, 2, 5, 5)
+        sub_sample = (1, 2)
+        test1_params = get_conv_output_shape(
+            image_shape, kernel_shape, 'valid', sub_sample)
+        test2_params = get_conv_output_shape(
+            image_shape, kernel_shape, 'half', sub_sample)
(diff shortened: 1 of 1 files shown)


theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+from theano.tensor.nnet.abstract_conv import get_conv_output_shape
+
+
+class TestGetConvOutShape(unittest.TestCase):
+
+    def test_basic(self):
+        image_shape, kernel_shape = (3, 2, 8, 8), (4, 2, 5, 5)
```
nouiz: This is a good start. To make the test coverage better, test with non square image and filter (change the last digit in the 2 shapes above).

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -0,0 +1,18 @@
...
+        sub_sample = (1, 2)
+        test1_params = get_conv_output_shape(
+            image_shape, kernel_shape, 'valid', sub_sample)
+        test2_params = get_conv_output_shape(
+            image_shape, kernel_shape, 'half', sub_sample)
+        test3_params = get_conv_output_shape(
+            image_shape, kernel_shape, 'full', sub_sample)
```
nouiz: To increase test coverage, add a test case when the 3rd parameter is a tuple of int and a test case with just an int.

With both of those change, I think it is good. thanks

Sentient07:
@nouiz have made the suggested changes

PR #3830, opened 2016-01-01
New year 

Sentient07:
Happy new year! :)

### DIFF
--- a/doc/LICENSE.txt
+++ b/doc/LICENSE.txt
@@ -10,7 +10,7 @@ Contains code from NumPy, Copyright (c) 2005-2011, NumPy Developers.
 All rights reserved.
 
 Contain CnMeM under the same license with this copyright:
-Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
+Copyright (c) 2015, NVIDIA CORPORATION. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
--- a/doc/LICENSE.txt
+++ b/doc/LICENSE.txt
@@ -3,14 +3,14 @@
 LICENSE
 =======
 
-Copyright (c) 2008--2015, Theano Development Team
+Copyright (c) 2008--2016, Theano Development Team
 All rights reserved.
 
 Contains code from NumPy, Copyright (c) 2005-2011, NumPy Developers.
 All rights reserved.
 
 Contain CnMeM under the same license with this copyright:
-Copyright (c) 2015, NVIDIA CORPORATION. All rights reserved.
+Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
--- a/theano/tests/test_flake8.py
+++ b/theano/tests/test_flake8.py
@@ -14,7 +14,7 @@ except ImportError:
     flake8_available = False
 
 __authors__ = ("Saizheng Zhang")
-__copyright__ = "(c) 2015, Universite de Montreal"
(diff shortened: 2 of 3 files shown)


nouiz:
thanks. if I grep for 2015 in Theano I find others. Do you want to also update them?

Sentient07:
@nouiz Yeah, sure! Actually, i was about to do that today.

Sentient07:
@nouiz I've ```grep```'ed all 2015 and changed the necessary.

doc/LICENSE.txt
```diff
@@ -3,14 +3,14 @@
...
-Copyright (c) 2008--2015, Theano Development Team
+Copyright (c) 2008--2016, Theano Development Team
 All rights reserved.
 
 Contains code from NumPy, Copyright (c) 2005-2011, NumPy Developers.
 All rights reserved.
 
 Contain CnMeM under the same license with this copyright:
-Copyright (c) 2015, NVIDIA CORPORATION. All rights reserved.
+Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
```
abergeron: There is still no new code for this, so the copyright year remains 2015.  We should update it when we next do a sync.

theano/compat/six.py
```diff
@@ -1,6 +1,6 @@
 """Utilities for writing code that runs on Python 2 and 3"""
 
-# Copyright (c) 2010-2015 Benjamin Peterson
+# Copyright (c) 2010-2016 Benjamin Peterson
```
abergeron: This isn't our code so the year should only change when (if) we import a new version of six.

theano/sandbox/cuda/cnmem.cpp
```diff
@@ -1,5 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////////////////////////
-// Copyright (c) 2015, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
```
abergeron: Not our code again

theano/sandbox/cuda/cnmem.h
```diff
@@ -1,5 +1,5 @@
 /* ********************************************************************** 
- * Copyright (c) 2015, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
```
abergeron: Idem

theano/tests/test_flake8.py
```diff
@@ -14,7 +14,7 @@
     flake8_available = False
 
 __authors__ = ("Saizheng Zhang")
-__copyright__ = "(c) 2015, Universite de Montreal"
+__copyright__ = "(c) 2016, Universite de Montreal"
```
abergeron: I'm not sure about this one ...

nouiz: About this one, we wanted to do as in this file everywhere in Theano. But
we didn't do it. But as this is our license, I would update it. I'm not
sure if we should keep it, so by not being sure, I would just update it.

doc/LICENSE.txt
Sentient07: Not sure, but since we use CUDA, i had thought i should update NVIDIA license as well. Will revert

abergeron: That copyright is for the CnMeM code that we embed.  CUDA has a separate license.

theano/tests/test_flake8.py
abergeron: Ok then, makes sense to update this.

Sentient07:
@nouiz @abergeron have made the changes

abergeron:
Ok with me (but still waiting for travis to be sure).

PR #3831, opened 2016-01-02
Assorted Python 3 fixes

kmike:
* ProfileStats.summary() is fixed for Python 3;
* Python 2/3 compatible examples in tutorial and library reference;
* less deprecation warnings in tests in Python 3


nouiz:
thanks

PR #3836, opened 2016-01-06
WIP Remove assertions for dimensionality.

twiecki:
nlinalg is forcing all ndims == 2. However, the numpy functions work just fine for higher dimensional inputs. For example:

```python
np.linalg.det(np.stack([np.eye(3), np.eye(3)]))
```
yields:
```
array([ 1.,  1.])
```

This PR is not functional yet as there seems to be a problem with `infer_shape`. Any pointers on how those would need to be updated? E.g. the one for `det`?

### DIFF
--- a/theano/tensor/nlinalg.py
+++ b/theano/tensor/nlinalg.py
@@ -38,7 +38,6 @@ class MatrixPinv(Op):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim == 2
         return Apply(self, [x], [x.type()])
 
     def perform(self, node, inputs, outputs):
@@ -70,7 +69,6 @@ class MatrixInverse(Op):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim == 2
         return Apply(self, [x], [x.type()])
 
     def perform(self, node, inputs, outputs):
@@ -273,7 +271,6 @@ class Det(Op):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim == 2
         o = theano.tensor.scalar(dtype=x.dtype)
         return Apply(self, [x], [o])
 
@@ -292,7 +289,7 @@ class Det(Op):
         return [gz * self(x) * matrix_inverse(x).T]
 
     def infer_shape(self, node, shapes):
-        return [()]
+        return [shapes[0]]
 
     def __str__(self):
         return "Det"
@@ -310,7 +307,6 @@ class Eig(Op):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim == 2
         w = theano.tensor.vector(dtype=x.dtype)
         v = theano.tensor.matrix(dtype=x.dtype)
         return Apply(self, [x], [w, v])
@@ -342,7 +338,6 @@ class Eigh(Eig):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim == 2
         # Numpy's linalg.eigh may return either double or single
         # presision eigenvalues depending on installed version of
         # LAPACK.  Rather than trying to reproduce the (rather
@@ -419,11 +414,7 @@ class EighGrad(Op):
 
     def make_node(self, x, w, v, gw, gv):
         x, w, v, gw, gv = map(as_tensor_variable, (x, w, v, gw, gv))
-        assert x.ndim == 2
-        assert w.ndim == 1
-        assert v.ndim == 2
-        assert gw.ndim == 1
-        assert gv.ndim == 2
+        assert 2 * x.ndim ==  w.ndim == 2 * v.ndim == gw.ndim == 2 * gv.ndim
         out_dtype = theano.scalar.upcast(x.dtype, w.dtype, v.dtype,
                                          gw.dtype, gv.dtype)
         out = theano.tensor.matrix(dtype=out_dtype)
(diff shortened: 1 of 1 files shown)


theano/tensor/nlinalg.py
```diff
@@ -292,7 +289,7 @@ def grad(self, inputs, g_outputs):
         return [gz * self(x) * matrix_inverse(x).T]
 
     def infer_shape(self, node, shapes):
-        return [()]
+        return [shapes[0]]
```
nouiz: In the old case when x.ndim == 2, this don't return the right answer.

What could be done:
```
if node.inputs[0].ndim == 2:
   return [()]
raise NotImplementedError()
```

This would make infer_shape not implemented for other ndim. I don't know what is det(nd tensor), so maybe something more generic can be done.

nouiz:
We also need to test those cases.

thanks

twiecki:
Trying:

```python
import numpy as np
import theano.tensor as T
1 / T.nlinalg.det(np.stack([np.eye(3), np.eye(3)]))
```

Which raises:
```
<ipython-input-4-f50e335bc855> in <module>()
      1 import numpy as np
      2 import theano.tensor as T
----> 3 1 / T.nlinalg.det(np.stack([np.eye(3), np.eye(3)]))

/home/wiecki/tools/theano_fork/theano/tensor/var.py in __rtruediv__(self, other)
    207 
    208     def __rtruediv__(self, other):
--> 209         return theano.tensor.basic.true_div(other, self)
    210 
    211     def __rfloordiv__(self, other):

/home/wiecki/tools/theano_fork/theano/gof/op.py in __call__(self, *inputs, **kwargs)
    608             for i, ins in enumerate(node.inputs):
    609                 try:
--> 610                     storage_map[ins] = [self._get_test_value(ins)]
    611                     compute_map[ins] = [True]
    612                 except AttributeError:

/home/wiecki/tools/theano_fork/theano/gof/op.py in _get_test_value(cls, v)
    536             # ensure that the test value is correct
    537             try:
--> 538                 ret = v.type.filter(v.tag.test_value)
    539             except Exception as e:
    540                 # Better error message.

/home/wiecki/tools/theano_fork/theano/tensor/type.py in filter(self, data, strict, allow_downcast)
    175             raise TypeError("Wrong number of dimensions: expected %s,"
    176                             " got %s with shape %s." % (self.ndim, data.ndim,
--> 177                                                         data.shape))
    178         if not data.flags.aligned:
    179             try:

TypeError: For compute_test_value, one input test value does not have the requested type.
 
Backtrace when that variable is created:
  File "<ipython-input-4-f50e335bc855>", line 3, in <module>
    1 / T.nlinalg.det(np.stack([np.eye(3), np.eye(3)]))

The error when converting the test value to that variable type:
Wrong number of dimensions: expected 0, got 1 with shape (2,).
```

What's confusing me is that I tried to set a break point into `infer_shape` but it's not triggering it. @nouiz Any idea?

nouiz:
test value don't use infer_shape. So this is normal the break point don't get triggered.

To trigger it, you can do:

```
>>> import numpy as np
>>> import theano
>>> import theano.tensor as T
>>> 1 / T.nlinalg.det(np.stack([np.eye(3), np.eye(3)]))
Elemwise{true_div,no_inplace}.0
>>> d=1 / T.nlinalg.det(np.stack([np.eye(3), np.eye(3)]))
>>> f=theano.function([], d.shape)
  File "theano/compile/function.py", line 317, in function
    output_keys=output_keys)
  File "theano/compile/pfunc.py", line 526, in pfunc
    output_keys=output_keys)
  File "theano/compile/function_module.py", line 1777, in orig_function
    output_keys=output_keys).create(
  File "theano/compile/function_module.py", line 1457, in __init__
    optimizer_profile = optimizer(fgraph)
  File "theano/gof/opt.py", line 101, in __call__
    return self.optimize(fgraph)
  File "theano/gof/opt.py", line 89, in optimize
  File "theano/gof/opt.py", line 230, in apply
  File "theano/gof/opt.py", line 85, in optimize
    self.add_requirements(fgraph)
  File "theano/tensor/opt.py", line 1451, in add_requirements
    fgraph.attach_feature(ShapeFeature())
  File "theano/gof/fg.py", line 624, in attach_feature
    attach(self)
  File "theano/tensor/opt.py", line 1275, in on_attach
    self.on_import(fgraph, node, reason='on_attach')
  File "theano/tensor/opt.py", line 1328, in on_import
    self.set_shape(r, s)
  File "theano/tensor/opt.py", line 1119, in set_shape
    len(s), r.ndim, sio.getvalue()))
AssertionError: Something inferred a shape with 3 dimensions for a variable with 0 dimensions for the variable:
Det [id A] <TensorType(float64, scalar)> ''   
 |TensorConstant{[[[ 1.  0... 0.  1.]]]} [id B] <TensorType(float64, 3D)>
```

theano/tensor/nlinalg.py
CzB404: Hello, I recently bumped into this problem when trying to translate an algorithm implementation from Numpy to Theano, and I see there is a bit of a misunderstanding here.

Numpy's linear algebra package considers any N D array as an N-2 D array of 2D matrices. So for an array of shape N1×N2×N3×...×Nn×M×M the numpy determinant function would return the results for each M×M matrix in an array of size N1×N2×N3×...×Nn.

Link to the relevant numpy documentation: https://docs.scipy.org/doc/numpy-1.12.0/reference/routines.linalg.html#linear-algebra-on-several-matrices-at-once

Issue #3837, opened 2016-01-06
Can't import register_opt on CPU


TiRune:
Working on bleeding edge Theano, get this issue:

======================================================================
ERROR: Failure: ImportError (cannot import name as_cuda_ndarray_variable)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/local/lib/python2.7/dist-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/local/lib/python2.7/dist-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/mnt/data/Projects/scyfernn/tests/layers/test_activation_layer.py", line 5, in <module>
    from tests.utils import func_fact, elicit_error, \
  File "/mnt/data/Projects/scyfernn/tests/utils.py", line 7, in <module>
    from NN.layers.identity_layer import IdentityLayer
  File "/mnt/data/Projects/scyfernn/NN/layers/__init__.py", line 20, in <module>
    from pooling_layer import PoolingLayer
  File "/mnt/data/Projects/scyfernn/NN/layers/pooling_layer.py", line 1, in <module>
    import pooling_ops
  File "/mnt/data/Projects/scyfernn/NN/layers/pooling_ops.py", line 698, in <module>
    from theano.sandbox.cuda.opt import register_opt
  File "/usr/local/lib/python2.7/dist-packages/Theano-0.7.0-py2.7.egg/theano/sandbox/cuda/opt.py", line 22, in <module>
    from theano.sandbox.cuda import as_cuda_ndarray_variable
ImportError: cannot import name as_cuda_ndarray_variable
-------------------- >> begin captured logging << --------------------
theano.sandbox.cuda: WARNING: CUDA is installed, but device gpu0 is not available  (error: Unable to get the number of gpus available: no CUDA-capable device is detected)
--------------------- >> end captured logging << ---------------------

I realize that this is fixed when I turn on my gpu, but I want to have my code work for CPU as well. Fixed this in Theano, and making a pull request soon for this.

Quick question, would it be better form to only include optimization in the DB by using register_opt if cuda is installed by checking cuda_available, or is it better form to just register the optimizations and check inside the optimization if cuda is available? I believe something can be said for both.

[...]

PR #3838, opened 2016-01-06
Pull request for #3837

TiRune:
Error occured when importing register_opt when running CPU code. 
Error is because:
- as_cuda_ndarray_variable was in theano.sandbox.cuda.basic_ops, and not theano.sandbox.cuda
- extra_ops has an import of register_opt only if if cuda_available, however, the optimization is registered without the same check.

fix gh-3837

### DIFF
--- a/theano/sandbox/cuda/extra_ops.py
+++ b/theano/sandbox/cuda/extra_ops.py
@@ -5,11 +5,12 @@ from theano.gof import local_optimizer
 from theano.sandbox.cuda import cuda_available, GpuOp
 from theano.sandbox.cuda.basic_ops import gpu_flatten
 from theano.tensor.extra_ops import CumsumOp
+from theano.sandbox.cuda import register_opt as register_opt
 
 if cuda_available:
     from theano.sandbox.cuda import CudaNdarrayType
     from theano.sandbox.cuda.basic_ops import host_from_gpu, gpu_from_host, HostFromGpu
-from theano.sandbox.cuda import register_opt as register_gpu_opt
+
 
 
 class GpuCumsum(CumsumOp, GpuOp):
@@ -436,7 +437,7 @@ def values_eq_approx_high_tol(a, b):
     return CudaNdarrayType.values_eq_approx(a, b, rtol=rtol)
 
 
-@register_gpu_opt()
+@register_opt()
 @local_optimizer([CumsumOp])
 def use_gpu_cumsum(node):
     if type(node.op) is CumsumOp \
--- a/theano/sandbox/cuda/extra_ops.py
+++ b/theano/sandbox/cuda/extra_ops.py
@@ -9,7 +9,7 @@ from theano.tensor.extra_ops import CumsumOp
 if cuda_available:
     from theano.sandbox.cuda import CudaNdarrayType
     from theano.sandbox.cuda.basic_ops import host_from_gpu, gpu_from_host, HostFromGpu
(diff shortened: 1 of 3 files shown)


nouiz:
I killed travis on the current commit. I'll let it run with the final version with the line uncommented.

nouiz:

/home/travis/build/Theano/Theano/theano/sandbox/cuda/extra_ops.py:16:1: E303 too many blank lines (3)

TiRune:
@nouiz Perhaps you are right, and that's not the place to change things. Here's the issue: On CPU fallback, an error is thrown in that spot, because in cuda.__init__, GpuAlloc is only added if cuda_available.

To fix the error, you can either add an explicit cuda_available check to the optimizer itself. Or only register the optimization in optdb when cuda_available. It is added in sandbox.cuda.opt.

So I'm wondering what the design philosophy is for optdb. Should all potential optimizations be added, and the optimizations themselves check if the gpu is available, or should be optimizations never be registered to optdb if the gpu is not available?

If I fix it in either spot, CPU fallback works without any errors, only warnings. Seems like a good thing to have :).

nouiz:
I think I understand the problem. theano.sandbox.cuda.opt got imported. But currently, it should only be imported if theano.sandbox.cuda is important and if cuda_available is true.

Do you have an idea why cuda.opt got imported? To allow importing this file when cuda_available is False Maybe this diff would allow to import cuda.opt when cuda_available is False:

diff --git a/theano/sandbox/cuda/opt.py b/theano/sandbox/cuda/opt.py
index 20fe061..eeeb4f4 100644
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2501,7 +2501,7 @@ optdb.register('gpu_scanOp_make_inplace',
                                              gpu_flag=True),
                75,
                'gpu',
-               'fast_run',
+#               'fast_run',
                'inplace',
                'scan')

Can you try it? There could be other place where change is needed to make this fully working.

nouiz:
It will also need this extra diff. So the full diff to try:

```
diff --git a/theano/sandbox/cuda/opt.py b/theano/sandbox/cuda/opt.py
index 20fe061..eeeb4f4 100644
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2501,7 +2501,7 @@ optdb.register('gpu_scanOp_make_inplace',
                                              gpu_flag=True),
                75,
                'gpu',
-               'fast_run',
+#               'fast_run',
                'inplace',
                'scan')
diff --git a/theano/sandbox/cuda/__init__.py b/theano/sandbox/cuda/__init__.py
index f8cdb00..dd93f13 100644
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -492,6 +492,8 @@ def use(device,
                        'fast_run')
         optdb.add_tags('gpu_after_fusion',
                        'fast_run')
+        optdb.add_tags('gpu_scanOp_make_inplace',
+                       'fast_run')
 
     if force:
         try:

```

TiRune:
Seems that with the most recent version of Theano, and the fix in our code:

```python
+from theano.sandbox.cuda import register_opt
-from theano.sandbox.cuda.opt import register_opt
```

This is fixed! Can ignore this PR now.

nouiz:
ok. So I'm closing.

PR #3845, opened 2016-01-07
Add border_mode="half" to cuDNN convolutions

f0k:
The new `conv2d()` interface supports `border_mode="half"` for same convolutions. However, the cudnn convolution interface in Theano does not support it. `local_abstractconv_cudnn()` does not take care of this. Thus, the following fails when cuDNN is available:
```python
In [1]: import theano
Using gpu device 0: Tesla K40c (CNMeM is enabled)

In [2]: x, w = theano.tensor.tensor4s('xw')

In [3]: c = theano.tensor.nnet.conv2d(x, w, border_mode='half')

In [4]: f = theano.function([x, w], c)

ERROR (theano.gof.opt): Optimization failure due to: LocalOptGroup(local_abstractconv_cudnn,local_conv_dnn,local_abstractconv_gemm,local_abstractconv_gradinputs_gemm,local_abstractconv_gradweight_gemm,local_conv_gemm)
ERROR (theano.gof.opt): node: AbstractConv2d{border_mode='half', subsample=(1, 1), filter_flip=True, imshp=None, kshp=None}(GpuFromHost.0, GpuFromHost.0)
  [...]
ValueError: invalid border_mode half, which must be either "valid", "full", an integer or a pair of integers
```

An easy fix would be modifying `local_abstractconv_cudnn()` to catch `border_mode="half"` and replace it with a suitable tuple of integers. It's difficult to adapt `test_abstractconv.py` accordingly, though, because it uses the cuDNN convolution as a reference. Plus it would only work for convolutions that have a filter size specified at compile time.

So I opted for changing `theano.sandbox.cuda.dnn` to include support for the "half" border mode, which wasn't very difficult either. `test_abstractconv.py` fails for me, but it also fails for the latest master.
`test_dnn.py` still needs updates for the shape inference tests (to not give repeated values).

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -259,10 +259,10 @@ class GpuDnnConvDesc(GpuOp):
             assert len(border_mode) == len(subsample)
             border_mode = tuple(map(int, border_mode))
         if not ((isinstance(border_mode, tuple) and min(border_mode) >= 0) or
-                border_mode in ('valid', 'full')):
+                border_mode in ('valid', 'full', 'half')):
             raise ValueError(
                 'invalid border_mode {}, which must be either '
-                '"valid", "full", an integer or a pair of'
+                '"valid", "full", "half", an integer or a pair of'
                 ' integers'.format(border_mode))
         self.border_mode = border_mode
         assert len(subsample) in [2, 3]
@@ -292,12 +292,14 @@ class GpuDnnConvDesc(GpuOp):
         if isinstance(self.border_mode, tuple):
             pad_desc = tuple(map(int, self.border_mode))
             assert min(pad_desc) >= 0
-            bmode = 2
+            bmode = 1
         else:
             pad_desc = [0] * nb_dim
 
             if self.border_mode == "valid":
                 bmode = 1
+            elif self.border_mode == "half":
+                bmode = 2
             else:
                 assert self.border_mode == "full"
                 bmode = 0
@@ -343,6 +345,14 @@ class GpuDnnConvDesc(GpuOp):
         pad[2] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 4) - 1;
     }
   }
+  // Adjust padding values if using half convolution
+  else if (%(bmode)d == 2) {
+    pad[0] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 2) / 2;
+    pad[1] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 3) / 2;
+    if (%(nb_dim)d >= 3) {
+        pad[2] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 4) / 2;
+    }
+  }
 
   err = cudnnSetConvolutionNdDescriptor_v3(
   %(desc)s,
@@ -365,7 +375,7 @@ class GpuDnnConvDesc(GpuOp):
            upscale_str=upscale_str, nb_dim=nb_dim, precision=precision)
 
     def c_code_cache_version(self):
-        return (3, version())
+        return (4, version())
 
 # scalar constants
 _zero = constant(numpy.asarray(0.0, dtype='float32'))
@@ -1097,7 +1107,7 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1),
     kerns
         Convolution filters.
     border_mode
-        One of 'valid', 'full'; additionally, the padding size can be
+        One of 'valid', 'full', 'half'; additionally, the padding size can be
         directly specified by an integer or a pair of integers (as a tuple),
         specifying the amount of zero padding added to _both_ the top and
         bottom (first entry) and left and right (second entry) sides of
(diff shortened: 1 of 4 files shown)


nouiz:
What errors do you have with test_abstractconv.py? I had errors with them, but only without the default config flag. If you change cudnn flags, to change the convolution to one that isn't deterministic, this will make the gradient tests fail due to op not being deterministic. This is expected.

Otherwise, by code review it look good. I'll try to run the tests on it today.

f0k:
The end of the output for `nosetests theano/sandbox/cuda/tests/test_abstractconv.py` is:
```
  Max Abs Diff:  1.77389
  Mean Abs Diff:  0.32628
  Median Abs Diff:  0.271882
  Std Abs Diff:  0.252909
  Max Rel Diff:  0.9874
  Mean Rel Diff:  0.203226
  Median Rel Diff:  0.161352
  Std Rel Diff:  0.168004

  rtol, atol: 1e-05 1e-05


Ran 2 tests in 84.711s

FAILED (errors=2)
```
My configuration has floatX=float32, device=gpu, lib.cnmem=700. Nothing else.

> Otherwise, by code review it look good. I'll try to run the tests on it today.

With `nosetests theano/sandbox/cuda/tests/test_dnn.py` I only get a failure for `test_DnnSoftMax.test_softmax` that's also present in master. However, I get a bunch of warnings like the following:
```
WARNING (theano.tests.unittest_tools): While testing the shape inference, we received an input with a shape that has some repeated values: (7, 8, 7, 4), like a square matrix. This makes it impossible to check if the values for these dimensions have been correctly used, or if they have been mixed up.
```
I need to figure out suitable shapes for `TestDnnInferShapes` that also work for `border_mode="half"` in addition to the modes already tested.

f0k:
All right, modified the shapes for the shape inference tests. Good to merge from my side!

(Fun fact: github displays the commits in chronological order, not in the order they would be applied to... "More informative warning" is applied first. Integration testing picked the correct one.)

ballasn:
Code looks good to me,

f0k:
Code looks good to Travis as well.

nouiz:
I tried it here and got problems too. But it was with the driver 352.63 . We discovered that this is a bugged driver. We are updating to 352.68. The master pass the tests with this driver. Which driver do you use?

f0k:
> I tried it here and got problems too.

But are they related to my code or independent? I only got problems that were not caused by my changes anyway.

> Which driver do you use?

This was with 352.39 and CUDA 7.5.

f0k:
> We are updating to 352.68. The master pass the tests with this driver.

Anything left to do for this PR then? I can't update the drivers right now, so I still get errors in `nosetests theano/sandbox/cuda/tests/test_abstractconv.py` both with my PR and with the latest Theano master. Let me know if you've been able to run the tests on your machine and if anything needs to be changed! That bug is a bit annoying.

craffel:
:+1:

theano/sandbox/cuda/dnn.py
```diff
@@ -343,6 +345,14 @@ def c_code(self, node, name, inputs, outputs, sub):
...
     }
   }
+  // Adjust padding values if using half convolution
+  else if (%(bmode)d == 2) {
+    pad[0] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 2) / 2;
+    pad[1] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 3) / 2;
+    if (%(nb_dim)d >= 3) {
+        pad[2] = *(npy_int64 *)PyArray_GETPTR1(%(kern_shape)s, 4) / 2;
+    }
+  }
```
lamblin: I think you would need to bump `c_code_cache_version`.

lamblin:
I'm trying to run the tests with a clean cache.
If that works, and then I'm OK to merge after you update the `c_code_cache_version`.

lamblin:
So I have issues with `theano.sandbox.cuda.tests.test_dnn.test_conv3d*`: some wrong values, some `RuntimeError: GpuDnnConvGradI: error getting worksize: CUDNN_STATUS_BAD_PARAM`).
I'm running with the latest master to make sure where these are coming from.

lamblin:
I can confirm that I have a total of `FAILED (errors=174)` with a checkout of this branch when running the tests on `theano/sandbox/cuda/tests/test_dnn.py`, but no failure with the lastest master.

lamblin:
Actually, the tests seem to run correctly if the cache is full from a previous run with the master, but if I clear it, then I encounter the errors.
I also tried rebasing this branch on top of the latest master, and I have the same behaviour.

xxxzhi:
I get this traceback when I use 'half' border_mode

    theano.tensor.nnet.conv2d(..., border_mode='half')

    File "/home/houzhi/Theano/theano/tensor/nnet/conv.py", line 531, in __init__
        raise Exception("Mode %s not implemented" % str(self.out_mode))
    Exception: Mode half not implemented

f0k:
Sorry for the late reply, I was sick. Found the problem: line 295 set `bmode=2` for custom padding, and I was inadvertently using `bmode=2` for half padding as well. I've changed line 295 to set `bmode=1` (matching the "valid" mode, which leaves the padding values as they are), and bumped the C code version. Should work now, fingers crossed -- I'll try to test it locally.

lamblin:
Looks good, thanks. I just ran `test_dnn.py` locally without issues, I'm running `test_abstractconv.py` now, and will merge if it succeeds.

lamblin:
Thanks!

f0k:
Great, thanks! Had some server problems and couldn't test myself yesterday.

PR #3846, opened 2016-01-07
DOC removed incorrect statement from compatibility docs

kmike:
In Python 2.x encode/decode on incorrect type is not no-op - Python 2 tries to convert data implicitly first. E.g. if you call `b'hello'.encode('utf8')`, Python 2 will first convert `b'hello'` to unicode by decoding it using `sys.getdefaultencoding()`, and then encode the result to utf8. It is almost always a wrong thing to do.


nouiz:
thanks

PR #3847, opened 2016-01-08
Update opt.py

jojolalpin:
StructuredDotCSC.c_code : typo correction: ) is missing before ; on two lines

### DIFF
--- a/theano/sparse/opt.py
+++ b/theano/sparse/opt.py
@@ -348,8 +348,8 @@ class StructuredDotCSC(gof.Op):
             // pointers to access actual data in the arrays passed as params.
             dtype_%(z)s*     __restrict__ Dz   = (dtype_%(z)s*)PyArray_DATA(%(z)s);
             const dtype_%(a_val)s* __restrict__ Dval = (dtype_%(a_val)s*)PyArray_DATA(%(a_val)s);
-            const npy_int32 * __restrict__ Dind = (npy_int32*)PyArray_DATA(%(a_ind)s;
-            const npy_int32 * __restrict__ Dptr = (npy_int32*)PyArray_DATA(%(a_ptr)s;
+            const npy_int32 * __restrict__ Dind = (npy_int32*)PyArray_DATA(%(a_ind)s);
+            const npy_int32 * __restrict__ Dptr = (npy_int32*)PyArray_DATA(%(a_ptr)s);
 
             //npy_intp nnz = PyArray_DIMS(%(a_ind)s)[0];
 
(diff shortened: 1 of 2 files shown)


jojolalpin:
about structured_dot_basic operator (simple modification of theano.sparse.structured_dot) :
"""
    Structured Dot Basic is like Structured Dot, except that it works
    on Sparse matrix csm properties + format ('csr': 0, 'csc': '1')
    It computes gradients only on a_data and b

    The output is presumed to be a dense matrix, and is represented by a
    TensorType instance.

    Main purpose of this operator (based on theano.sparse.structured_dot)
    is to enable gradient computing through scan with respect to sparse matrix.
    With scan, a sparse matrix not having broadcastable attribute,
    grad_of_scan fails.

    Parameters (tensors)
    ----------
    a_data
        A sparse matrix data vector
    a_indices
        A sparse matrix indices vector
    a_indptr
        A sparse matrix data index pointers
    a_shape
        A sparse matrix shape vector (not tuple)
    a_format:
        0 for "csr" or 1 for "csc"
    b
        A dense matrix.

    Returns
    -------
    A dense matrix
        The dot product of `a` and `b`.

    Notes
    -----
    The grad implemented is regular for a_data and b 
        
    """

nouiz:
Thanks. Ideally, we should have this tested. I don't know anyone with the times to add tests for this. Do you want to add some? Just tell me if you want or not. If you won't do it, I'll merge this PR.

The idea of the tests is to have this error detected, so to not break it again.

jojolalpin:
The optimizer is disabled so it won't do anything.
It works in my code and before correction compiler is clear, it says the ')' is missing before ;.
Without this I would not have caught it.
To make the test run, optimiser must be enabled and it is written that scipy is as fast. I need it for a custom version of structured_dot running on sparse properties instead of the matrix to enable grad of scan.
Tell me if you still want a test.

jojolalpin:
As I do not know how to include correctly a test in the project, here is just the code : 

import numpy as np
import scipy
import theano
from theano.sparse.opt import sd_csc

def test_sd_csc():

    A = scipy.sparse.rand(4, 5, density=0.60, format='csc', dtype=np.float32)
    b = np.random.rand(5,2).astype(np.float32)
    target = A*b
    
    a_val = theano.tensor.as_tensor_variable(A.data)
    a_ind = theano.tensor.as_tensor_variable(A.indices)
    a_ptr = theano.tensor.as_tensor_variable(A.indptr)
    nrows = theano.tensor.as_tensor_variable(np.int32(A.shape[0]))
    b = theano.tensor.as_tensor_variable(b)
    
    res = sd_csc(a_val, a_ind, a_ptr, nrows, b).eval()

    assert (res==target).all()

nouiz:
this is good. Thank!

PR #3853, opened 2016-01-09
Fix the lifter for abstractconv2d.

abergeron:
The abstractconv tests should already test for that.

### DIFF
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -806,6 +806,9 @@ theano.tensor.nnet.conv2d()
             AbstractConv2d_gradWeights,
             AbstractConv2d_gradInputs])
 def local_lift_abstractconv2d(node, context_name):
+    if isinstance(node.outputs[0].type, GpuArrayType):
+        # Don't handle this node here.
+        return
     inps = list(node.inputs)
     inps[0] = as_gpuarray_variable(node.inputs[0],
                                    context_name=context_name)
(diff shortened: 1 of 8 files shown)


theano/sandbox/gpuarray/opt.py
```diff
@@ -811,7 +811,7 @@ def local_lift_abstractconv2d(node, context_name):
                                    context_name=context_name)
     inps[1] = as_gpuarray_variable(node.inputs[1],
                                    context_name=context_name)
-    return [node.op(*inps)]
```
nouiz: what was the consequence of this? I suppose an optimization warning and the op not being lifted? Is that detected by a test?

abergeron:
I've made a different fix since the problems comes from elsewhere.

lamblin:
I'm OK with the new fix. Is it possible to make a small test, where the shapes of a convolution gradient are coming from the GPU?

theano/sandbox/gpuarray/opt.py
```diff
@@ -806,6 +806,9 @@ def local_error_convop(node, context_name):
             AbstractConv2d_gradWeights,
             AbstractConv2d_gradInputs])
 def local_lift_abstractconv2d(node, context_name):
+    if isinstance(node.outputs[0].type, GpuArrayType):
+        # Don't handle this node here.
```
nouiz: As this need a test, can you change this comment to tell:

# the op is already on the GPU

PR #3855, opened 2016-01-11
Update 2D convolution documentation

ballasn:


doc/library/tensor/nnet/conv.txt
```diff
@@ -24,8 +24,14 @@
...
-    As of October 21st, 2014, the default GPU image convolution
-    changed: By default, if :ref:`cuDNN <libdoc_cuda_dnn>`
+    As of December 2015, a new conv2d interface has been introduced.
+    :func:`nnet.conv2d <theano.tensor.nnet.conv2d>` defines an
+    abstract theano graph convolution operation
+    (:func:`nnet.abstract_conv.AbstractConv2d <theano.tensor.nnet.abstract_conv.AbstractConv2d>`)
+    that will be replaced by an actual convolution implementation during
+    the optimization phase.
+
+    By default, if :ref:`cuDNN <libdoc_cuda_dnn>`
```
lamblin: Below, I would say "Either CuDNN or the gemm version can be disabled", since the abstract Op does not get optimized to the legacy GPU convolution.
Maybe you can mention (later?) that this is still possible to get the legacy gpuconv2d using the old nnet.conv2d.conv2d() call.

lamblin:
The changes look reasonable, but I have to give another look to check if there are other things to update elsewhere in the doc.
Also, do you think it would be possible for you to convert the documentation format to NumPyDoc for these files, while you are at it?
Thanks.

lamblin:
Also, could you add something saying that since the abstract Op does not have any implementation, it will prevent computations in the un-optimized graph, and cause problems with DebugMode, test values, and when compiling with `optimizer=None`?

ballasn:
updated,

PR #3856, opened 2016-01-11
Don't transfer int inputs to the GPU by default

abergeron:
This was not the old behavior so it introduces new and exciting failures.

I left a way to force an input on the CPU/GPU if you really want to.

I also renamed tag.context_name to tag.target and documented it for inputs.


lamblin:
I discussed that change with @abergeron IRL, so I'm OK with it.
However I think we need @nouiz's approval before changing the behaviour.

nouiz:
ok to merge when travis pass.

Issue #3744, opened 2015-12-04
Factor out testing of blas flags, use in all cases


lamblin:
#3654 introduced a better test for blas flags, including compiling and launching something that actually tests the flags.
However, it is only used to test the last-resort case, `-lblas`, not for the flags gotten from numpy, anaconda, or provided by the user.
We should factor out that into one helper function or method of the compiler, and call it at the various places we want to test the flags.

PR #3858, opened 2016-01-12
auxiliary function 'try_blas_flags' added.

mohammadpz:
Regarding issue #3744 and PR #3654 .
@nouiz , could you please take a look?

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -349,7 +349,7 @@ def try_blas_flag(flags):
     # res[0]: shows successful compilation
     # res[1]: shows successful execution
     if res and res[0] and res[1]:
-        return "-lblas"
+        return flags
     else:
         return ""
 
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -288,8 +288,7 @@ SOMEPATH/Canopy_64bit/User/lib/python2.7/site-packages/numpy/distutils/system_in
                 flags += ['-l%s' % l for l in ["mkl_core",
                                                "mkl_intel_thread",
                                                "mkl_rt"]]
-                if GCC_compiler.try_flags(flags) and try_blas_flag():
-                    flags += ['-lblas']
+                if try_blas_flag(flags):
                     return ' '.join(flags)
 
         # If numpy was linked with library that are not installed or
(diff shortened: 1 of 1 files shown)


mohammadpz:
In the line 2082 of ```theano/gof/cmodule```, in the function ```try_flags```, the argument ```try_run``` is False. While we need to ```try_run``` for blas. One way is to leave this as is while another way is to override the function ```try_flags```.

theano/tensor/blas.py
```diff
@@ -344,6 +350,9 @@ def default_blas_ldflags():
     res = GCC_compiler.try_compile_tmp(
         test_code, tmp_prefix='try_blas_',
         flags=flags, try_run=True)
+    # res[0]: shows successful compilation
+    # res[1]: shows successful execution
+    print(res)
```
nouiz: remove the print.

theano/tensor/blas.py
```diff
@@ -304,14 +305,16 @@ def default_blas_ldflags():
...
+                ret += ['-lblas']
                 return ' '.join(ret)
             # Try to add the anaconda lib directory to runtime loading of lib.
             # This fix some case with Anaconda 2.3 on Linux.
             if "Anaconda" in sys.version and "linux" in sys.platform:
                 lib_path = os.path.join(sys.prefix, 'lib')
                 ret.append('-Wl,-rpath,' + lib_path)
-                if GCC_compiler.try_flags(ret):
+                if GCC_compiler.try_flags(ret) and try_blas_flag():
+                    ret += ['-lblas']
```
nouiz: remove this addition to the flags. We already found the flags we think should be used, don't modify them. Idem elsewhere.

theano/tensor/blas.py
```diff
@@ -304,14 +305,16 @@ def default_blas_ldflags():
                 ['-L%s' % l for l in blas_info['library_dirs']] +
                 ['-l%s' % l for l in blas_info['libraries']] +
                 [])
-            if GCC_compiler.try_flags(ret):
+            if GCC_compiler.try_flags(ret) and try_blas_flag():
```
nouiz: Doing those 2 tests are do duplicate test. Just call try_blas_flag, is cover what try_flags was doing.

But modify try_blas_flag() to take ret as a parameter and use it.

nouiz:
thanks for the PR. I got 2 small comments.

theano/tensor/blas.py
mohammadpz: @nouiz , For ```blas```, we also check the execution while for other flags we only check the compilation. Maybe we can also check execution for other flags too?

theano/tensor/blas.py
```diff
@@ -322,7 +322,10 @@ def default_blas_ldflags():
     # readily available. We try to see if that's the case, rather
     # than disable blas. To test it correctly, we must load a program.
     # Otherwise, there could be problem in the LD_LIBRARY_PATH.
+    return try_blas_flag()
```
nouiz: this should be `return try_blas_flags(['-lblas'])

nouiz:
Try it with the script I told you. Do you still get the same result?

mohammadpz:
By running
```python
python theano/misc/check_blas.py -B 500
```
It leads to the same results.

Issue #3854, opened 2016-01-11
Version number not updated


gladiator8072:
Hi Just download  theano package while i found the version number is still keep the same as 0.7.0
'''
VERSION
    0.7.0.dev-f279798ba7d7af72be74f7b48f1cf40f41bb163c
'''

PR #3860, opened 2016-01-12
update version to 0.7.0rc1 for #3854

evelynmitchell:
Fix version number issue #3854

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ PLATFORMS           = ["Windows", "Linux", "Solaris", "Mac OS-X", "Unix"]
 MAJOR               = 0
 MINOR               = 7
 MICRO               = 0
-SUFFIX              = ""  # Should be blank except for rc's, betas, etc.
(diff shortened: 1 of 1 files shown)


nouiz:
This isn't the good fix. This bring back in the past the version. This would be the good fix and use the new python pep 0440: https://www.python.org/dev/peps/pep-0440/#developmental-releases

Can you confirm the other PR fix your problem too?

setup.py
```diff
@@ -53,7 +53,7 @@
 MAJOR               = 0
 MINOR               = 7
 MICRO               = 0
-SUFFIX              = ""  # Should be blank except for rc's, betas, etc.
```
nouiz: I think a development version is a pre alpha version. So we should also the version number. I did what I think it should be in this PR: https://github.com/Theano/Theano/pull/3861

So I'll close this PR. If you think we should not bump the version number at the same time. Tell us.

thanks for pushing this.

PR #3861, opened 2016-01-12
Bump version to 0.8.0.dev0

nouiz:
this replace https://github.com/Theano/Theano/pull/3860 and close gh-3854

Use postfix .dev0 and bump to 0.8.0.dev0. This follow this new pep: https://www.python.org/dev/peps/pep-0440/#developmental-releases

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,9 +51,9 @@ AUTHOR              = "LISA laboratory, University of Montreal"
 AUTHOR_EMAIL        = "theano-dev@googlegroups.com"
 PLATFORMS           = ["Windows", "Linux", "Solaris", "Mac OS-X", "Unix"]
 MAJOR               = 0
-MINOR               = 7
+MINOR               = 8
 MICRO               = 0
-SUFFIX              = ""  # Should be blank except for rc's, betas, etc.
+SUFFIX              = ".dev0"  # Should be blank except for rc's, betas, etc.
 ISRELEASED          = False
 
 VERSION             = '%d.%d.%d%s' % (MAJOR, MINOR, MICRO, SUFFIX)


nouiz:
@f0k Can you confirm this would help tracking Theano development version for you?

We can bump the .dev0 when we add a new user interface feature. This would allow to force update Theano dependency version.

f0k:
> Can you confirm this would help tracking Theano development version for you?

Yes, we can include this in our `requirements.txt` for Lasagne and in the version check we do on importing Lasagne.

> We can bump the .dev0 when we add a new user interface feature.

Sometimes we pin the dependency for Lasagne to a Theano version that fixed a particular bug (without changing the user interface). For example, the current dependency is 54186290, after the merge of https://github.com/Theano/Theano/pull/3802 (which fixes BLAS detection for Travis). We could send you a PR bumping Theano's `dev` version whenever we need it to pin a dependency.

nouiz:
I'm fine with frequent bump of this number. If we see it happen too
frequently, we can reavaluate. But I agree, bugfix is also a good reason to
bump this number.

taion:
What's the benefit of going from e.g. v0.8.0.dev0 to v0.8.0.dev1 instead of just going from v0.8.0 to v0.8.1 on bugfixes?

The latter is much closer to how semver-managed packages handle things pre-v1.0.0 (in practice, I mean – semver doesn't strictly require anything pre-v1.0.0 but many package maintainers follow these conventions). I think this kind of scheme would just be simpler to reason about. There's not much risk of running out of numbers, after all.

f0k:
> What's the benefit of going from e.g. v0.8.0.dev0 to v0.8.0.dev1 instead of just going from v0.8.0 to v0.8.1 on bugfixes?

v0.8.0 implies a final release: https://www.python.org/dev/peps/pep-0440/#final-releases. Users will expect this to be published on PyPI.
v0.8.0.dev0 to devN are pre-release development versions: https://www.python.org/dev/peps/pep-0440/#developmental-releases. Those can be freely used to mark versions in between releases, and that's what we're dealing with here (although it would surely be nice to release to PyPI more often).

taion:
Oops, misunderstood that part.

Still, these seem in effect to just be pre-releases. It seems like it'd be easier for everyone involved with respect to version management if the pre-releases were all published to PyPI.

I'd much rather specify a dependency on e.g. `Theano==0.8.0a53` than have to go to GitHub.

nouiz:
@f0k, so you know if we can put dev on pypi? With recent pip version, by
default, it will install only final release, so we could update alpha there
and people would use them. But I don't know about dev version.

taion:
I just tested this out on the test PyPI:

PyPI accepts whatever: https://testpypi.python.org/pypi/Flask-RESTy/0.5.3.dev0

`pip` will ignore pre-releaes, but by default PyPI will show you the page for the pre-release, which might be annoying: https://testpypi.python.org/pypi/Flask-RESTy

abergeron:
I'm ok with this version scheme.

PR #3870, opened 2016-01-14
move print_tips to profiling.py

Saizheng:


theano/compile/profiling.py
```diff
@@ -1302,6 +1302,122 @@ def summary(self, file=sys.stderr, n_ops_to_print=20,
             print("-----------------", file=file)
             self.optimizer_profile[0].print_profile(file,
                                                     self.optimizer_profile[1])
+        self.print_tips()
+
+    def print_tips(self):
```
lamblin: Like for `summary_function`, `summary_globals`, and so on, we should be able to specify a destination file where the tips are printed. Can you add a `file` argument to `print_tips`, and then change all the `print(...)` calls to `print(..., file=file)`?

theano/compile/profiling.py
```diff
@@ -1302,6 +1302,122 @@ def summary(self, file=sys.stderr, n_ops_to_print=20,
...
+        print("""Here are tips to potentially make your code run faster
+                 (if you think of new ones, suggest them on the mailing list).
+                 Test them first, as they are not guaranteed to always provide a speedup.""")
+        from theano import tensor as T
+        from theano.tensor.raw_random import RandomFunction
+        import theano
+        import theano.scalar as scal
```
lamblin: The imports should be at the beginning of the file.

theano/compile/profiling.py
```diff
@@ -1302,6 +1302,122 @@ def summary(self, file=sys.stderr, n_ops_to_print=20,
...
+
+        printed_tip = False
+        # tip 1
+        if config.floatX == 'float64':
+            print("  - Try the Theano flag floatX=float32")
+            printed_tip = True
+
+        # tip 2
+        if not config.lib.amdlibm and any([amdlibm_speed_up(a.op) for i, a
+                                           in self.apply_time]):
```
lamblin: So, `self.apply_time` used to be a list of pairs, apparently, but now it is a dictionary, mapping from apply node to timing.
So you need to replace `for i, a in self.apply_time` with an iteration on a dictionary instead.
Here, you only want the Apply node `a`, so you can simply use `for a in self.apply_time`.
If you need both the Apply node `a` and the time `t`, you can use `for a, t in iteritems(self.apply_time)`.

It should be the same for the rest of the function.

lamblin:
Thanks for the first version!
Can you take care of the comments above, and check that the unit tests on Travis pass?
Thanks,

Saizheng:
I merged wrong things, I will rebase it.

Saizheng:
rebased.

kmike:
Sorry, what's the `:w` file in this PR?

Saizheng:
@kmike  Sorry, this is a mistake, I will delete it.
@lamblin Please check this issue again when I pushed the new version.

Issue #3681, opened 2015-11-24
strided-conv-{transpose,adjoint} in abstract-conv


dwf:
Having more direct access to the conv-transpose op (specifically, conv-transpose with stride in the output) would be a useful addition. This is useful for generative networks.

The current way of doing it (with CuDNN only) requires some acrobatics:

https://github.com/Newmu/dcgan_code/blob/master/lib/ops.py#L85-L96

lamblin:
Right.
This interface would instantiate `AbstractConv2d_gradInputs`, then the rest will be taken care of by existing optimizations.
Note that, unlike `conv2d`, you will have to provide the shape for the output of the operation, since it cannot be automatically determined when the strides are not 1.

nouiz:
@Sentient07 this one is also interresting. Mostly move the code from that repo to Theano, but make it use the new theano.tensor.nnet.abstract_conv.AbstractConv2d_gradInputs op instead to make it work on the CPU and add tests.

I would put it in the file theano.tensor.nnet.__init__.py

Sentient07:
@nouiz Yeah, seems doable. Should the test be at ``` theano.tensor.nnet.tests.test_conv.TestConv2D ``` ?

nouiz:
no, it should be in tensor/tests/test_abstract_conv.py. It currently don't
exist for some complex reason, but normally, it should be there.

Sentient07:
@nouiz Sorry for the delayed reply, i had gone on a vaccation. Should i be using only the AbstractConv2d_gradInputs op ? I'm having this confusion as i can see the GPU version using 3 different Ops..

Sentient07:
Hi, its almost a year and I just am realizing more about this issue. Can we have a GPUOp for this? caffe has an implementation. https://github.com/BVLC/caffe/blob/master/src/caffe/layers/deconv_layer.cu

lamblin:
The Ops have been in Theano for a while, there is one based on caffe-like convolutions and their gradients (GpuCorrMM), and one based on cuDNN.
There are also helper functions named `conv{2,3}d_grad_wrt_{inputs,weights}`.
The only part missing would be an interface named differently, to convey the fact that it is a transposed convolution (as well as gradient wrt inputs), which is what #3989 is about, so I'll close this ticket.

PR #3872, opened 2016-01-14
deconvolution interface

SinaHonari:
PR for issue #3681


lamblin:
@dwf, since you opened the original issue, your feedback would be welcome.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +177,187 @@ def conv2d(input,
...
+        You can give ``None`` for any element of the list to specify that this
+        element is not known at compile time.
+
+    :type border_mode: str, int or tuple of two int
+    :param border_mode: Either of the following:
+        * ``'valid'``: apply filter wherever it completely overlaps with the
+          input. Generates output of shape: input shape - filter shape + 1
+        * ``'full'``: apply filter wherever it partly overlaps with the input.
+          Generates output of shape: input shape + filter shape - 1
+        * ``'half'``: pad input with a symmetric border of ``filter rows // 2``
```
fvisin: The "traditional" border modes are usually `full`, `valid` and `same` (see e.g. [matlab's conv](http://www.mathworks.com/help/matlab/ref/conv.html#input_argument_shape)). 
I suggest to implement `same`, maybe in place of `half` which seems almost the same.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +177,187 @@ def conv2d(input,
...
+           input_shape=None,
+           output_shape=None,
+           filter_shape=None,
+           border_mode='valid',
+           subsample=(1, 1),
+           filter_flip=True):
+    """This function will build the symbolic graph for deconvoloving a
+    mini-batch of a stack of 2D inputs with a set of 2D filters (using
+    gradient of convolution), such that it upsamples the input to the
+    desired output resolution.
```
fvisin: We struggled to find a reference for this when we wrote the paper, but we did find a couple of articles. If you want to cite some literature and/or use our description of the operation, if you like it, here is the paper (ref. to sec. 2.2.4): http://arxiv.org/pdf/1511.07053v2.pdf

dwf: I think it's officially quite wrong to call it "deconv", as deconvolution
is something completely different (namely, trying to reverse the effect of
a convolution to recover the original signal).

fvisin: I definitely agree!! In ReSeg we called it *Gradient-based upscaling layer*, which is not very informative but more accurate.

theano/tensor/nnet/abstract_conv.py
abergeron: I would just add the 'same' mode, but maybe as part of another PR.

This is just exposing an interface and "half" is already there.

theano/tensor/nnet/abstract_conv.py
abergeron: Maybe we can rename the function to gradInputs().

theano/tensor/nnet/abstract_conv.py
lamblin: `same` is tricky, because it may mean uneven padding on both sides, which is not implemented in some back-ends (if not all), or taking a subtensor (which makes the tensor non-contiguous).
This is why we have `half`, which is equivalent to `same` when `same` is achievable with symmetric padding.

theano/tensor/nnet/abstract_conv.py
lamblin: OK, so maybe we should name the functions `conv2d_grad_inputs` and `conv2d_grad_weights`, and mention in the docstring of `conv2d_grad_inputs` that this is what Zeiler et al. call "deconvolution".
I think it is important to keep the "conv2d" part in the name, and we do not use `camelCase` in function names.

theano/tensor/nnet/abstract_conv.py
fvisin: I didn't know that, let's stick with `half` then.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +177,187 @@ def conv2d(input,
...
+def gradWeights(input,
+           output,
+           input_shape=None,
+           output_shape=None,
+           filter_shape=None,
+           border_mode='valid',
+           subsample=(1, 1),
+           filter_flip=True):
+    """This function will build the symbolic graph for getting the
+    gradient of the output of a convolution layer w.r.t its wights.
```
fvisin: Typo.

theano/tensor/nnet/abstract_conv.py
abergeron: I still think we should add 'same', but not in this PR.

theano/tensor/nnet/abstract_conv.py
abergeron: Ok with your proposed names.

fvisin: I think that naming them `conv2d_grad_wrt_inputs` and `conv2d_grad_wrt_weights` would be even clearer, but I would agree if you'd find them too long. Actually, to underline that it is not just the gradient of the conv2d op, it might be better to name them  `gradconv2d_wrt_inputs` and `gradconv2d_wrt_weights`.

SinaHonari:
Pascal and I checked Zeiler's deconvolutional networks papers (https://www.cs.nyu.edu/~gwtaylor/publications/zeilertaylorfergus_iccv2011.pdf, http://www.matthewzeiler.com/pubs/cvpr2010/cvpr2010.pdf). Apparently, what they do is using alternating layers of convolutional sparse coding instead of taking the gradient of conv outputs w.r.t its inputs.

lamblin:
So, the discussion at TensorFlow for this functionality is at https://github.com/tensorflow/tensorflow/issues/256. Interesting points:
- Matt Zeiler calls the entire stack Deconvolution Networks, but only ever refers to this operator as a 'projection operator'
- Caffe calls it Deconvolution

TensorFlow settled on `conv2d_transpose`, in addition to the underlying `conv2d_backprop_input`.
Maybe it would make sense to also add an alias with that name, that would also be consistent with the class in Blocks (https://github.com/mila-udem/blocks/blob/master/blocks/bricks/conv.py#L183-L235).

For the moment, I would be OK with merging this PR, and open a new ticket to add a `conv2d_transpose` later. Is that OK for you @nouiz @dwf @vdumoulin @abergeron ?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +177,195 @@ def conv2d(input,
...
+                           border_mode='valid',
+                           subsample=(1, 1),
+                           filter_flip=True):
+    """This function builds the symbolic graph for getting the
+    gradient of the output of a convolution (namely output_grad)
+    w.r.t the input of the convolution, given a set of 2D filters
+    used by the convolution, such that the output_grad is upsampled
+    to the input shape.
+
+    :type output_grad: symbolic 4D tensor.
```
fvisin: IMO it's quite confusing to call the input of the function `output_grad`. I understand why it was called like this, but probably something like `input_tensor` or 'to_be_upsampled_tensor' would be better.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +177,195 @@ def conv2d(input,
...
+    :note: If CuDNN is available, it will be used on the
+        GPU. Otherwise, it is the *CorrMM* convolution that will be used
+        "caffe style convolution".
+
+    :note: This is only supported in Theano 0.8 or the development
+        version until it is released.
+
+    """
+
+    deconv_op = AbstractConv2d_gradInputs(imshp=input_shape,
```
vdumoulin: I know it's internal to this function's implementation, but I think we'd be better off naming this `grad_input_op` in analogy to `conv2d_grad_wrt_weights`'s implementation, avoiding the whole "deconv vs transpose" debate (team _transpose_, by the way).

lamblin:
OK, so we agreed with @vdumoulin, @fvisin, and @nouiz to have a `conv2d_transpose` later. So for now I'm going to comment the last points that need to be addressed for that PR, and then I'll merge when they are addressed.

theano/tensor/nnet/abstract_conv.py
lamblin: Let's keep `output_grad` here for now.

theano/tensor/nnet/abstract_conv.py
lamblin: I agree, rename `deconv_op` into `grad_input_op`.

lamblin:
So in the end, there is only one small change, and then we can merge. Thanks!

SinaHonari:
I applied the change.

lamblin:
Thanks!

PR #3875, opened 2016-01-15
Fix buildbot

abergeron:
This should fix one of the problems with the buildbot.

Also it should get rid of the majority of deprecation warnings which should help readability of the output.

I'll work on the other problems, but this should be merged as soon as possible to help future work.

### DIFF
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -938,7 +938,7 @@ class Op(utils.object2, PureOp, CLinkerOp):
         """
         logger = logging.getLogger('theano.gof.op.Op')
 
-        new_node = self.prepare_node(self, node)
+        new_node = self.prepare_node(node)
         if new_node is not None:
             node = new_node
 
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -821,6 +821,14 @@ class Op(utils.object2, PureOp, CLinkerOp):
         else:
             return NotImplemented
 
+    def prepare_node(self, node):
+        """
+        Make any special modifications that the Op needs before doing
+        make_thunk().
+
+        """
+        pass
+
     def make_c_thunk(self, node, storage_map, compute_map, no_recycling):
         """
         Like make_thunk, but will only try to make a C thunk.
@@ -929,6 +937,10 @@ class Op(utils.object2, PureOp, CLinkerOp):
 
         """
         logger = logging.getLogger('theano.gof.op.Op')
+
+        new_node = self.prepare_node(self, node)
+        if new_node is not None:
+            node = new_node
 
         if self._op_use_c_code:
             try:
@@ -1166,10 +1178,8 @@ int main( int argc, const char* argv[] )
                 self.openmp = False
                 theano.config.openmp = False
 
-    def make_thunk(self, node, storage_map, compute_map, no_recycling):
+    def prepare_node(self, node):
         self.update_self_openmp()
-        return super(OpenMPOp, self).make_thunk(node, storage_map,
-                                                compute_map, no_recycling)
 
 
 def simple_meth(tag):
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -826,6 +826,8 @@ class Op(utils.object2, PureOp, CLinkerOp):
         Make any special modifications that the Op needs before doing
         make_thunk().
 
+        This can either modify the node inplace or return a new one.
+
         """
         pass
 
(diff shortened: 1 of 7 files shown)


lamblin:
`//` would have worked instead of `floor`, but not for `ceil`.
The `ufunc` change should fix the precision issue when computing floating point values from an int8 vector (for instance, `log(int8(2))`), where numpy now does the computation in float16, whereas the Theano C code does it in float32 since the output variable is float32 by default.

nouiz:
So the ufunc change should fix the daily buildbot error in debutmode?

Just add a comment in the code why this dtype value is needed. Otherwise people won't remember. Tell that without this, tests fail in debugmode. Otherwise we can easily think by removing it and having tests pass, that all is good. This is tricky code not very well tested. (Or add a test that force debugMode and would fail if you remove this line)

abergeron:
It's not only that tests fail in debugmode without this, but also that perform() gives wrong results without this.

nouiz:
So why travis pass? Can you add a test that without this perform fail? Travis run in FAST_COMPILE. So we should have seen it.

abergeron:
I don't know.  I have seen the tests pass in FAST_COMPILE too, but I just wonder if it's somehow not using the same codepath as DebugMode is for the python code.  I know this shouldn't be the case, but clearly DebugMode is doing something different here.

nouiz:
Can you check? Or add a test that would fail with the normal mode?

abergeron:
I've added a test for the elemwise DebugMode problem and also fixed another bug with MultinomialFromUniform.

nouiz:
Ok to merge when travis pass.

lamblin:
Doing the computation in float16 does not give completely wrong results, just a drop in precision, which is probably why it is only detected when explicitly comparing against a higher-precision computation, which is done by DebugMode.

abergeron:
This fix doesn't work (since it breaks other things).  I'll have to do something else.

abergeron:
It seems the problem is because DebugMode doesn't call make_thunk from Elemwise and this sets special hints so that the computation is performed with the right dtype.

nouiz:
So we should change how DebugMode work. I'm not sure how to change this.
There is a few cases to take care:
1) The op implement only a perform
2) The op implement only c code
3) The op implement c code and perform
4) The op implement only make_thunk
5) The op implement make_thunk AND perform or c code.

It is the 5) case that isn't well supported by DebugMode.

abergeron:
And that is because someone registered the make_thunk of Elemwise as a default_make_thunk back in the day, but now it is required and skipped by DebugMode.

If we remove the registration the tests work, but don't compare the C and Python code.

Maybe it's time to introduce a prepare_node() method that should be called from make_thunk and does the side-processing.  We have enough Ops that use that pattern that we should build special support for it.

nouiz:
Good idea. Can you do that for Elemwise to fix this problem in the
buildbot? We can see later to port this to the other places.

nouiz:
To make this work with DebugMode, it should be in make_py_thunk and
make_c_thunk, not in make_thunk.

abergeron:
I've implemented the proposed solution.

theano/gof/op.py
```diff
@@ -1166,10 +1178,8 @@ def update_self_openmp(self):
                 self.openmp = False
                 theano.config.openmp = False
 
-    def make_thunk(self, node, storage_map, compute_map, no_recycling):
+    def prepare_node(self, node):
         self.update_self_openmp()
```
nouiz: need to return it.

theano/tensor/elemwise.py
```diff
@@ -830,8 +829,6 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
             char = numpy.sctype2char(out_dtype)
             sig = char * node.nin + '->' + char * node.nout
             node.tag.sig = sig
-        return super(Elemwise, node_.op).make_thunk(node_, storage_map,
```
nouiz: need to return it.

theano/gof/op.py
```diff
@@ -821,6 +821,14 @@ def __eq__(self, other):
         else:
             return NotImplemented
 
+    def prepare_node(self, node):
+        """
+        Make any special modifications that the Op needs before doing
+        make_thunk().
```
nouiz: Tell that is if it can modify in place this node if this don't change its semantic. Like adding tag and precomputed stuff.

But tell that it can return a new node to use it this is needed.

abergeron: Ok.

nouiz:
Ok to merge when travis pass.

In this PR or another, we should document prepare_node in the user doc of new op.

abergeron:
This passed travis so I'll merge.

PR #3877, opened 2016-01-17
Fix segfault

abergeron:
This fixes a segfault when deleting ops that use cudnn in the new backend.

Since it reverts the fix for the context problem I also did an alternate fix for that.

### DIFF
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -820,8 +820,7 @@ class CLinker(link.Linker):
                 pass
 
             try:
-                struct_cleanup = op.c_cleanup_code_struct(node, name,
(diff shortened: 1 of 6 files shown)


nouiz:
It seem ok. I'm running tests in Theano.

How much space a dnn handle take? Do this take memory on the GPU? Having one per context is probably not ideal. If we want to change this, it would go in an issue for later, we should first fix the current bug. But if you have information on this, we can start the discussion here (or a new issue)

nouiz:
gpuarray tests passed.

abergeron:
A dnn handle is slightly below 1M.

Issue #3878, opened 2016-01-17
type of replacement doesn't match with original variable


namitkewat:
I am trying Theano on my Ubuntu laptop with AMD GPU. I was following documentations; on the page [Using the GPU](http://deeplearning.net/software/theano/tutorial/using_gpu.html), i faced one issue with 'check2.py' example that `'gpu_from_host' is not defined`. Then i tried 
```python
gpu_from_host = sandbox.gpuarray.basic_ops.GpuFromHost(None)
f = function([], gpu_from_host(tensor.exp(x)))
``` 
and it worked. its output was:
```bash
(py27)namitkewat@swagat:~$ python Documents/ANNs\ examples/check2.py 
Mapped name None to device opencl0:0: Hainan
[GpuElemwise{exp,no_inplace}(<GpuArrayType<None>(float64, (False,))>)]
Looping 1000 times took 0.188476 seconds
Result is [ 1.23178032  1.61879341  1.52278065 ...,  2.20771815  2.29967753
  1.62323285]
Used the gpu
```
i have also noticed the massive speed-up with above change in `check2.py` example(`check1.py` with device as `cpu` was taking around 8 seconds, and device as `opencl0:0` was taking around 2 seconds, and with above changes `check2.py` is taking around 0.20 seconds(around 40-45x faster than with `cpu` as device in `check1.py`; which also matches with the results shown in the documentations)

After looking at above results; i have added `gpu_from_host = GpuFromHost(None)` line in `Theano/theano/sandbox/gpuarray/basic_ops.py` file. then i tried [logistics regression](http://deeplearning.net/software/theano/tutorial/examples.html#a-real-example-logistic-regression) example, and now i am getting error as:

```bash
<<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.TypeError'> ('The type of the replacement must be compatible with the type of the original Variable.', Dot22.0, HostFromGpu(gpuarray).0, TensorType(float64, col), TensorType(float64, matrix), 'local_gpua_dot22') local_gpua_dot22
ERROR (theano.gof.opt): Optimization failure due to: local_gpua_dot22
ERROR (theano.gof.opt): node: Dot22(x.T, DimShuffle{0,x}.0)
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
[...]

PR #3881, opened 2016-01-18
Fix opt warning/not moving to GPU some node.

nouiz:
fix gh-3878, the user tested it and it was working(after adding the missing import) that I amended in the commit.

### DIFF
--- a/theano/sandbox/gpuarray/blas.py
+++ b/theano/sandbox/gpuarray/blas.py
@@ -8,8 +8,8 @@ from theano.tensor.basic import as_tensor_variable
 from theano.tensor.opt import in2out
 
 from .basic_ops import as_gpuarray_variable, infer_context_name
-
 from .opt_util import inplace_allocempty
+from .type import GpuArrayType
 
 try:
     import pygpu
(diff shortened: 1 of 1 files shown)


theano/sandbox/gpuarray/blas.py
```diff
@@ -281,7 +281,12 @@ def make_node(self, x, y):
         assert x.ndim == 2
         assert y.ndim == 2
         assert x.dtype == y.dtype
-        return Apply(self, [x, y], [x.type()])
+        otype = GpuArrayType(
+            dtype=x.dtype,
+            broadcastable=(x.type.broadcastable[0], y.type.broadcastable[1]),
+            # Here I suppose both inputs are on the same context.
+            context_name=x.type.context_name)
```
abergeron: This would probably be better by using `x.type.clone(broadcastable=...)`

nouiz: done

theano/sandbox/gpuarray/blas.py
```diff
@@ -8,8 +8,8 @@
 from theano.tensor.opt import in2out
 
 from .basic_ops import as_gpuarray_variable, infer_context_name
-
 from .opt_util import inplace_allocempty
+from .type import GpuArrayType
```
abergeron: You should also remove this since it's not used.

nouiz: done

PR #3883, opened 2016-01-18
[WIP] add bessel0

skaae:
Adding Scipy Bessel function. 

I have a few questions:

1) The PR adds both bessel0 and bessel1. Bessel1 is only added because I don't know how to call 
scipy.special.j1 when calculating the gradietn of bessel0?

2) Is `bessel0` ok for name? Alternatively i can call it `j0` to match scipy?

3) Not sure how tests work. I think i should follow https://github.com/jsalvatier/Theano-1/commit/b18e04323d91c29918341ffdb14501399a6f2acf and add them to ``theano/tensor/tests/test_basic.py
`` but i'm not sure how theano testing works.

### DIFF
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -402,8 +402,11 @@ class J1(UnaryScalarOp):
     """
 
     @staticmethod
-    def impl(x):
+    def st_impl(x):
         return scipy.special.j1(x)
+
+    def impl(self, x):
+        return self.st_impl(x)
 
     def grad(self, inp, grads):
         raise NotImplementedError()
@@ -422,8 +425,11 @@ class J0(UnaryScalarOp):
     """
 
     @staticmethod
-    def impl(x):
+    def st_impl(x):
         return scipy.special.j0(x)
+
+    def impl(self, x):
+        return self.st_impl(x)
 
     def grad(self, inp, grads):
         x, = inp
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -406,10 +406,21 @@ class J1(UnaryScalarOp):
         return scipy.special.j1(x)
 
     def impl(self, x):
-        return self.st_impl(x)
+        if imported_scipy_special:
+            return self.st_impl(x)
+        else:
+            super(J1, self).impl(x)
 
     def grad(self, inp, grads):
         raise NotImplementedError()
+
+    def c_code(self, node, name, inp, out, sub):
+        x, = inp
+        z, = out
+        if node.inputs[0].type in float_types:
+            return """%(z)s =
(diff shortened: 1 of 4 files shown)


abergeron:
1) does not seem like a question.  Anyway, adding a new op is the way to go when a gradient needs to call something else so it's ok.

2) I can't comment on whether `bessel0` or `j0` is the best name.

3) Yes, that is a good model for the tests.

skaae:
@abergeron: Thanks for the comments

I added tests and adressed your comments + squashed the commits.

theano/scalar/basic_scipy.py
```diff
@@ -394,3 +394,73 @@ def __hash__(self):
...
+
+    @staticmethod
+    def st_impl(x):
+        return scipy.special.j1(x)
+
+    def impl(self, x):
+        if imported_scipy_special:
+            return self.st_impl(x)
+        else:
+            super(J1, self).impl(x)
```
diogo149: could this be missing a `return`?

skaae: I mostly copy pasted from the psi class just above j0 and j1. It doesn't have a return either.

nouiz: Super will raise an error, so the return isn't needed.

theano/scalar/basic_scipy.py
```diff
@@ -394,3 +394,73 @@ def __hash__(self):
...
+
+    def c_code(self, node, name, inp, out, sub):
+        x, = inp
+        z, = out
+        if node.inputs[0].type in float_types:
+            return """%(z)s =
+                j0(%(x)s);""" % locals()
+        raise NotImplementedError('only floating point is implemented')
+
+    def __eq__(self, other):
```
nouiz: I looked in the code and yes, the original had this, but it shouldn't. Can you remove \__hash__ and \__eq__ in all classes in this file? They probably didn't caused problems, but they shouldn't be there.

theano/tensor/tests/test_basic.py
```diff
@@ -1677,6 +1677,8 @@ def _numpy_true_div(x, y):
     expected_gammaln = scipy.special.gammaln
     expected_psi = scipy.special.psi
     expected_chi2sf = lambda x, df: scipy.stats.chi2.sf(x, df).astype(x.dtype)
+    expected_j0 = scipy.special.j0
+    expected_j1 = scipy.special.j1
```
nouiz: You must add those variable in the else close around line 1700. Otherwise, there will be a crash is scipy isn't installed.

nouiz:
I made 2 small comments. After that, this PR is ready to be merged. Thanks.

skaae:
@nouiz: I adressed your comments. Thanks for the feedback.

nouiz:
thanks

PR #3884, opened 2016-01-18
Fix blas detection on mac.

abergeron:


nouiz:
Ok to merge when travis pass.

Issue #3395, opened 2015-09-16
delete the Param class


nouiz:
That class is very similar to In and Out. Try to understand why it was there. Is it older then the In and Out class or it was there to merge both of them?

The In class was extended but not Param. So maybe the simplest is just to delete Param.

lamblin:
Closed by #3890

PR #3890, opened 2016-01-19
Param class in now replaced by In

mohammadpz:
Hi @nouiz ,
This PR is for solving #3395 .


doc/library/compile/function.txt
```diff
@@ -50,7 +50,7 @@ Reference
 
         Initialize attributes from arguments.
 
-.. class:: Param
```
nouiz: The list of attribute isn't in the same order and the In class have more attribute. Look at the docstring of the In class in theano/compile/io.py and its method `__init__`

doc/library/compile/function.txt
```diff
@@ -27,7 +27,7 @@ array(8.0)
 The idea here is that we've compiled the symbolic graph (``2*x``) into a function that can be called on a number and will do some computations.
 
 The behaviour of function can be controlled in several ways, such as
-:class:`Param`, ``mode``, ``updates``, and ``givens``.  These are covered
+:class:`In`, ``mode``, ``updates``, and ``givens``.  These are covered
```
nouiz: add the Out to the list and add the class Out after the class In from line 53.

nouiz:
Other then the current comment, all seem good. thanks

doc/library/compile/function.txt
mohammadpz: For the new arguments, I wrote a shorter version of the description in ```__init__```. Could you please check if that is right?

lamblin:
Thanks!

theano/__init__.py
```diff
@@ -67,7 +67,7 @@
     predefined_modes, predefined_linkers, predefined_optimizers,
     FunctionMaker, function, function_dump, OpFromGraph,
     ProfileMode, ProfileStats,
-    Param, shared, as_op)
```
lamblin: Actually, this is breaking something in the deep learning tutorials.
Should we keep an alias `Param = In`, or a dummy `Param` class that calls `In` with a deprecation warning?

nouiz: What about making paran a sub class or in and adding the missing, changed
name parameter to init?

Le mar. 26 janv. 2016 14:34, Pascal Lamblin <notifications@github.com> a
écrit :

mohammadpz: Please let me know if you make a decision. Thanks.

PR #3891, opened 2016-01-19
Fix broken blas flags on Anaconda with openblas.

abergeron:
Also fix broken default blas fallback.

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -346,21 +346,15 @@ def try_blas_flag(flags):
             return 0;
         }
         """)
+    cflags = flags + ['-L' + d for d in theano.gof.cmodule.std_lib_dirs()]
     res = GCC_compiler.try_compile_tmp(
         test_code, tmp_prefix='try_blas_',
-        flags=flags, try_run=True)
+        flags=cflags, try_run=True)
     # res[0]: shows successful compilation
     # res[1]: shows successful execution
     if res and res[0] and res[1]:
         return ' '.join(flags)
     else:
-        # Retry adding '-L' flags which may be required
-        flags = flags + ['-L' + d for d in theano.gof.cmodule.std_lib_dirs()]
-        res = GCC_compiler.try_compile_tmp(
-            test_code, tmp_prefix='try_blas_',
-            flags=flags, try_run=True)
-        if res and res[0] and res[1]:
-            return ' '.join(flags)
         return ""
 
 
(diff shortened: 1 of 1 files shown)


lamblin:

theano/tensor/blas.py
```diff
@@ -319,7 +322,9 @@ def default_blas_ldflags():
     # readily available. We try to see if that's the case, rather
     # than disable blas. To test it correctly, we must load a program.
     # Otherwise, there could be problem in the LD_LIBRARY_PATH.
-    return try_blas_flag(['-lblas'])
+    ret = try_blas_flag(['-lblas'])
+    if ret:
+        return ' '.join(ret)
```
nouiz: Here we should return "-lblas". This probably explain travis failure.

abergeron: Which is what this code does.  But it doesn't explain the failure because the previous code never worked.

nouiz: What I propose isn't equivalent. try_blas_flags modify what it return. We do not want that. In the past, the code was really returning "-lblas". I think we should continue to do that.

abergeron:
Apparently it is impossible to correctly link with mkl and therefore we always end up in the "default" case on travis.  I reworked that case to be more solid.

theano/tensor/blas.py
abergeron: It add a '-L' flag, which might be required for proper -lblas usage.  Since we don't test without it we don't know if it will work.

So I modified it to try without that first and only add it if it doesn't work.

nouiz:
travis still fail. You are changing what was being done in the past. This isn't good and can cause many unforeseen problems in non frequently tested case. I don't want that now. It isn't a good time for this.

The flag added in try_blas_flags is already added elsewhere during normal compilation. As we don't use normal compilation path, we need to manually add it there. The problem can be that now it is not added in the right order or is added at multiple places.

abergeron:
Travis didn't fail, scrutinizer did.  And by timing out.  Travis hasn't started yet.

I am not changing the past, I am unbreaking the commit that introduced try_blas_flags.  This is what changed the past.

Also, adding -L multiple times is a no-op so it can't introduce a problem.  Still I removed it since it will slow down testing to compile stuff two times.

nouiz:
travis fail with flake8 error:

```
/home/travis/build/Theano/Theano/theano/tensor/blas.py:314:13: E129 visually indented line with same indent as next logical line

/home/travis/build/Theano/Theano/theano/tensor/blas.py:314:13: W503 line break before binary operator
```

ok to merge with this small change and when travis pass.

abergeron:
done.

abergeron:
This now passes

PR #3893, opened 2016-01-20
Remove more warnings from the buildbot

abergeron:
This should remove about 50% of the remaining warning messages in the tests.

The rest is a case of 80% of the effort for 20% of results.  I still might get to it at some point.


lamblin:
OK when Travis passes.

Issue #3894, opened 2016-01-20
Import bug in tensor\signal\downsample.py


daweileng:
current Line 2 in downsample.py

    import pool
should be

    from . import pool

dwf:
Introduced in 372939b.

jfsantos:
I submitted a PR (#3897) fixing this.

tebzito:
Hey there. I still can't figure out the problem. First off, I can't find the import pool in downsample.py. And when I just add the 'from . import pool', the problem persists. Even after changing the 'DownsampleFactorMax' class name to Pool, still nothing. HELP!!!

lamblin:
The most likely explanation is that you have an old version of Theano, maybe you have several versions of Theano at the same time.
Can you uninstall _all_ the versions of Theano and try with 0.8.2?

PR #3897, opened 2016-01-20
Fixed import of local pool submodule

jfsantos:
`pool` is a local submodule, so it should be imported as `from . import pool` instead of `import pool`. This fixes a bug when importing the `downsample` module (at least on Python 3).

fix gh-3894


nouiz:
ok to merge when travis pass.

nouiz:
thanks

PR #3899, opened 2016-01-20
Add optimizations to transfer new LogSoftmax op to the GPU as CuDNN softmax ops

carriepl:
fix gh-3882

### DIFF
--- a/theano/sandbox/cuda/tests/test_dnn.py
+++ b/theano/sandbox/cuda/tests/test_dnn.py
@@ -612,6 +612,35 @@ class test_DnnSoftMax(test_nnet.test_SoftMax):
 
             utt.assert_allclose(out, expected_out)
 
+    def test_log_softmax2(self):
+        # Test that the op LogSoftmax is correctly replaced by the op
+        # DnnSoftmax with the 'log' mode.
+
+        # This is a test for an optimization that depends on CuDNN v3 or
+        # more recent. Don't test if the CuDNN version is too old.
+        if cuda.dnn.version() < (3000, 3000):
+            raise SkipTest("Log-softmax is only in cudnn v3+")
+
+        # Build the first graph and ensure that the optimization is applied
+        x = T.fmatrix()
+        log_softmax_out = T.nnet.LogSoftmax()(x)
+        f = theano.function([x], log_softmax_out, mode=mode_with_gpu)
+
+        dnn_softmax_nodes = [n for n in f.maker.fgraph.toposort() if
+                             isinstance(n.op, cuda.dnn.GpuDnnSoftmax)]
+        assert len(dnn_softmax_nodes) == 1
+        assert dnn_softmax_nodes[0].op.algo == "log"
+
+        # Build the first graph and ensure that the optimization is applied
+        x = T.fmatrix()
+        log_softmax_out = T.log(T.nnet.Softmax()(x))
+        f = theano.function([x], log_softmax_out, mode=mode_with_gpu)
+
+        dnn_softmax_nodes = [n for n in f.maker.fgraph.toposort() if
+                             isinstance(n.op, cuda.dnn.GpuDnnSoftmax)]
+        assert len(dnn_softmax_nodes) == 1
+        assert dnn_softmax_nodes[0].op.algo == "log"
+
 
 def test_dnn_tag():
     """
--- a/theano/sandbox/cuda/tests/test_dnn.py
+++ b/theano/sandbox/cuda/tests/test_dnn.py
@@ -616,13 +616,12 @@ class test_DnnSoftMax(test_nnet.test_SoftMax):
         # Test that the op LogSoftmax is correctly replaced by the op
         # DnnSoftmax with the 'log' mode.
 
-        # This is a test for an optimization that depends on CuDNN v3 or
-        # more recent. Don't test if the CuDNN version is too old.
-        if cuda.dnn.version() < (3000, 3000):
-            raise SkipTest("Log-softmax is only in cudnn v3+")
+        # Compile a reference function, on the CPU, to be used to validate the
+        # results of the other function.
+        x = T.fmatrix()
+        f_ref = theano.function([x], T.nnet.LogSoftmax()(x))
 
         # Build the first graph and ensure that the optimization is applied
-        x = T.fmatrix()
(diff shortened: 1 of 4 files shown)


theano/sandbox/gpuarray/dnn.py
```diff
@@ -1471,6 +1474,19 @@ def local_log_softmax_dnn(node):
...
+        # Transform the input in the format expected by GpuDnnSoftmax
+        inp = node.inputs[0].owner.inputs[0]
+        if inp.ndim != 2:
+            return
+        inp = inp.dimshuffle(0, 1, 'x', 'x')
+
+        # Apply GpuDnnSoftmax and return the result
+        out = GpuDnnSoftmax('log', 'channel')(gpu_contiguous(inp))
+        return [out.dimshuffle(0, 1)]
+
```
abergeron: This part might be better as an independent op_lifter for LogSoftmax.

abergeron:
The rest is ok for me.

carriepl:
Done.

carriepl:
(I killed Travis on the old commit, it still hadn't started)

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -612,6 +612,35 @@ def test_log_softmax(self):
...
+
+        # Build the first graph and ensure that the optimization is applied
+        x = T.fmatrix()
+        log_softmax_out = T.log(T.nnet.Softmax()(x))
+        f = theano.function([x], log_softmax_out, mode=mode_with_gpu)
+
+        dnn_softmax_nodes = [n for n in f.maker.fgraph.toposort() if
+                             isinstance(n.op, cuda.dnn.GpuDnnSoftmax)]
+        assert len(dnn_softmax_nodes) == 1
+        assert dnn_softmax_nodes[0].op.algo == "log"
```
nouiz: execute both function once to make sure it work.

carriepl: re-done

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1472,6 +1475,27 @@ def local_log_softmax_dnn(node):
...
 
+@register_opt('cudnn')
+@op_lifter([LogSoftmax])
+def local_logsoftmax_to_dnn(node, ctx_name):
+    if not dnn_available(ctx_name) or version() < 3:
+        # No log-softmax before cudnn v3
+        return
+
+    if (isinstance(node.op, LogSoftmax) and node.inputs[0].owner and
+            isinstance(node.inputs[0].owner.op, HostFromGpu)):
```
abergeron: You can remove that test since it's part of the op_lifter code.

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1472,6 +1475,27 @@ def local_log_softmax_dnn(node):
...
+
+        # Transform the input in the format expected by GpuDnnSoftmax
+        inp = node.inputs[0].owner.inputs[0]
+        if inp.ndim != 2:
+            return
+        inp = inp.dimshuffle(0, 1, 'x', 'x')
+
+        # Apply GpuDnnSoftmax and return the result
+        out = GpuDnnSoftmax('log', 'channel')(gpu_contiguous(inp))
```
abergeron: You might want to tag inp with ctx_name to make sure it works in all cases.

theano/sandbox/gpuarray/dnn.py
carriepl: Which one, the whole `if` or only some of its conditions?

theano/sandbox/gpuarray/dnn.py
carriepl: Is that done with `as_gpuarray_variable(inp, ctx_name)` or some other way?

theano/sandbox/gpuarray/dnn.py
nouiz: op_lifter will decide when this op need to be moved to the GPU. So yo don't need any of those condition. node will be LogSoftmax and we will want to move it to the GPU. (so the input or the output is on the GPU)

theano/sandbox/gpuarray/dnn.py
nouiz: inp.tag.context_name = ctx_name

theano/sandbox/gpuarray/dnn.py
```diff
@@ -1472,6 +1475,27 @@ def local_log_softmax_dnn(node):
...
+def local_logsoftmax_to_dnn(node, ctx_name):
+    if not dnn_available(ctx_name) or version() < 3:
+        # No log-softmax before cudnn v3
+        return
+
+    if (isinstance(node.op, LogSoftmax) and node.inputs[0].owner and
+            isinstance(node.inputs[0].owner.op, HostFromGpu)):
+
+        # Transform the input in the format expected by GpuDnnSoftmax
+        inp = node.inputs[0].owner.inputs[0]
```
nouiz: use inp=node.inputs[0]

then you won't need to tag for context as this will be done by the op_lifter.

nouiz: I got it wrong about the context, but stil use the new inp version.

theano/sandbox/gpuarray/dnn.py
nouiz: The context should be done in the inp just before the line out, so on the result of the dimshuffle.

theano/sandbox/gpuarray/dnn.py
abergeron: Yeah, so the whole if.

theano/sandbox/gpuarray/dnn.py
abergeron: Yes, you should use `node.inputs[0]` only because there is no guarantee that your input is a HostFromGPU (except that you check for that above).

theano/sandbox/gpuarray/dnn.py
abergeron: as_gpuarray_variable is acceptable too, since we know that the input should be on the GPU.

carriepl:
Done

abergeron:
Ok when travis pass.

Issue #3876, opened 2016-01-15
Print cuDNN Version On Device "using"


cancan101:
Right now when theano loads a device is prints out wether or not `CNMeM` is enabled. I suggest also adding whether or not cuDNN is in use (and which version as well). BOth of these libraries are important for good performance.
Currently:
```
Using gpu device 0: GRID K520 (CNMeM is enabled)
```
I suggest:
```
Using gpu device 0: GRID K520 (CNMeM is enabled; cuDNN v4)
```

nouiz:
I like this suggestion, do you want to make a PR? To get cudnn information,
you can call the function theano.sandbox.cuda.dnn.version() to get cudnn
version.

PR #3901, opened 2016-01-20
Added CuDNN version output, should it be compiled and reachable

claymcleod:
Directed at issue #3876. 

NOTE: I have not been able to reproduce this, I need someone with a properly configured GPU to confirm that this works as intended.

However, it's a relatively small change, and I inferred types based on what the C function returns.

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -453,7 +453,8 @@ def use(device,
                 import theano.sandbox.cuda.dnn
                 cudnn_version = ""
                 try:
-                    cudnn_version = ", cuDNN v%i" % (theano.sandbox.cuda.dnn.version())
+                    (hdr_v, runtime_v) = theano.sandbox.cuda.dnn.version()
+                    cudnn_version = ", cuDNN v%i" % (runtime_v)
                 except:
                     pass
                 cnmem_enabled = "enabled" if config.lib.cnmem else "disabled"
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -450,9 +450,15 @@ def use(device,
                                  " this property")
 
             if config.print_active_device:
+                import theano.sandbox.cuda.dnn
+                cudnn_version = ""
+                try:
+                    cudnn_version = ", cuDNN v%i" % (theano.sandbox.cuda.dnn.version())
+                except:
+                    pass
                 cnmem_enabled = "enabled" if config.lib.cnmem else "disabled"
-                print("Using gpu device %d: %s (CNMeM is %s)" % (
-                        active_device_number(), active_device_name(), cnmem_enabled), file=sys.stderr)
+                print("Using gpu device %d: %s (CNMeM is %s%s)" % (
+                        active_device_number(), active_device_name(), cnmem_enabled, cudnn_version), file=sys.stderr)
             if device_properties(use.device_number)['regsPerBlock'] < 16384:
                 # We will try to use too much register per bloc at many places
                 # when there is only 8k register per multi-processor.


abergeron:
version() returns a tuple of (<header version>, <runtime version>) or the inverse.

Maybe it should check that they match?
Or print both?

Anyway, your format won't work.

claymcleod:
So something more like this probably? Again, I can't test it right now, so I'm depending on someone else to. Thanks for the input.

claymcleod:
I'm thinking checking that the header version against the runtime version is probably out of scope for the Theano library.

abergeron:
That should work, yes.  But I haven't tested it either.

lamblin:
This would look like `(CNMeM is enabled(4004, 4004))`, which looks rather confusing. I think "v4" would be enough for what gets automatically printed.
We would also need to support the case when cudnn is not installed, or not available on that device, since `version()` would raise an exception in that case.
Finally, it would be nice to add the same thing in the new back-end (`theano/sandbox/gpuarray/dnn.py`), but the semantics of `version()` there is a bit different, and it does in fact check the header version against the linked version, and returns only one number.

nouiz:
I the string would be better as " (CNMeM is enabled, cudnn=4004),". Using
v4 only it too strick. Sometimes we need to know if this is a rc, private
release, final version... But we could use a hardcoded mapping 4004 ->
v4rc1, 4XXX -> v4 when released. When unknow, give the exact same number.

claymcleod:
@lamblin I assume after my second commit, the output would now be ```(CNMeM is enabled, cuDNN v4004)```. This raises the question of what is the best way to format it --- keeping it as it is, adding a decimal like ```(CNMeM is enabled, cuDNN v4.004)```, or just truncating the minor number all together like ```(CNMeM is enabled, cuDNN v4)```.

With regards to exception handling, I'm not terribly versed in how the sandbox module works, but my understanding is that this would achieve the desired effect. If the version is not able to be found for any reason at all and an error occurs, the part of the string that report the cuDNN version will simply remain blank.

theano/sandbox/cuda/__init__.py
```diff
@@ -450,9 +450,16 @@ def use(device,
                                  " this property")
 
             if config.print_active_device:
+                import theano.sandbox.cuda.dnn
+                cudnn_version = ""
+                try:
+                    (hdr_v, runtime_v) = theano.sandbox.cuda.dnn.version()
+                    cudnn_version = ", cuDNN v%i" % (runtime_v)
+                except:
```
abergeron: I just noticed, but this should catch Exception at least.  Not everything.

abergeron:
The current code should display v4004 as @claymcleod said.  It will correctly handle the case when cudnn is not present or can't be linked to for any reason.  In that respect it is correct.

However, I have no opinion however on what format should we use to display the version.

abergeron:
Please leave the title alone.  The version you display is the cudnn version not the CUDA version.  So it was incorrect.

nouiz:
I would accept v4004. But if someone want to make the mapping clear, that
would be better.

nouiz:
if you want to add why cudnn isn't there when there is an exception, you
can get the str to show at:

...cuda.dnn.dnn_available.msg

claymcleod:
@abergeron My fault on the title, wasn't paying attention and didn't see you changed it originally. 

Should the exception thrown by the version command be a RuntimeError like the rest of the errors in dnn.py, [such as here](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/dnn.py#L94)? Narrowing down the scope of what exceptions are ignored might be good since this code would run every time there is a GPU job.

abergeron:
It's actually either a RuntimeError from the line you pointed or a plain Exception from there:https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/dnn.py#L215

Also, I don't think it is appropriate to display the availability message in the gpu init code.

nouiz:
I would change to have both exception be the same. Can you do it at the same time?

lamblin:
OK, I think I missed the try/catch for some reason.
I think it may be better to have "CuDNN disabled" or "CuDNN not available" rather than a blank string, but a blank string is OK.

claymcleod:
After doing some more research on this, I have determined that this is not possible with the current setup due to a "race condition" type of problem. Determining the version in ```theano.sandbox.cuda.dnn.version()``` requires code that executes in ```theano.sandbox.cuda.__init__```. Taking it a step deeper, I tried to eliminate the dependencies in ```version()``` by doing the following:

```
def version(ignore_exceptions=False):
    """Return the current cuDNN version we compile with.

    This returns a tuple with the header version and the library
    version we link with. For older cudnn version without version
    information, we return -1.

    """
    if not ignore_exceptions and not dnn_available():
        raise Exception(
            "We can't determine the cudnn version as it is not available",
            dnn_available.msg)
                                                                
    if version.v is None:                                       
        f = theano.function([], DnnVersion()(),                 
                            theano.Mode(optimizer=None),        
                            profile=False)                      
        version.v = f()                                         
    return version.v  
```

However, executing the theano function also depends on code in __init__. Therefore, either the code that reports the version numbers needs to be moved out of __init__, or this isn't possible.

nouiz:
This can be fixed by moving the version code inside the init file. I did it
in this PR:

https://github.com/Theano/Theano/pull/4000

I wasn't able to start my PR based on yours, as you deleted your Theano
fork. Thanks for pushing this and having identified the problems.

PR #3903, opened 2016-01-21
R_op for AbstractConv2d

### LINKED ISSUE #3888: AbstractConv2d does not implement R_op
While trying to calculate a Hessian vector product of a loss function involving convolutions, I came across the following exception which told me to report this to theano:

    NotImplementedError: AbstractConv2d{border_mode=(0, 0), subsample=(1, 1),
    filter_flip=True, imshp=(None, 1, 28, 28), kshp=(32, 1, 5, 5)} of class
    AbstractConv2d does not implement R_op. If this is a theano op, write to the
    theano-dev mailing list for assistance. If it is your own op, implement the R_op method.

I will be happy to help resolving this, however I are not familiar with theanos convolutions.

matthias-k:
The implementation is exactly as in Conv2d.

Fixes #3888.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -308,6 +308,17 @@ class AbstractConv2d(BaseAbstractConv2d):
             'Did you exclude both "conv_dnn" and "conv_gemm" from '
             'the optimizer? Is cudnn available and does the GPU support it?')
 
+    def R_op(self, inputs, eval_points):
+        rval = None
+        if eval_points[0] is not None:
+            rval = self.make_node(eval_points[0], inputs[1]).outputs[0]
+        if eval_points[1] is not None:
+            if rval is None:
+                rval = self.make_node(inputs[0], eval_points[1]).outputs[0]
+            else:
+                rval += self.make_node(inputs[0], eval_points[1]).outputs[0]
+        return [rval]
+
     def grad(self, inp, grads):
         bottom, weights = inp
         top, = grads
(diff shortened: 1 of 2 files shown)


theano/tensor/nnet/abstract_conv.py
```diff
@@ -308,6 +308,17 @@ def perform(self, node, inp, out_):
             'Did you exclude both "conv_dnn" and "conv_gemm" from '
             'the optimizer? Is cudnn available and does the GPU support it?')
 
+    def R_op(self, inputs, eval_points):
```
nouiz: This would also need tests. rop tests are in theano/tests/test_rop.py grep for conv, you can probably copy/paste/modify (or better, generalize it to test both)

nouiz:
killing travis as it won't help this PR to advance.

matthias-k:
I changed the test in theano/tests/test_rop.py to test both `tensor.nnet.conv.conv2d` and `tensor.nnet.conv2d`. The test fails without the new implemention in `AbstractConv2d` and passes with it.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -308,6 +308,17 @@ def perform(self, node, inp, out_):
...
+        rval = None
+        if eval_points[0] is not None:
+            rval = self.make_node(eval_points[0], inputs[1]).outputs[0]
+        if eval_points[1] is not None:
+            if rval is None:
+                rval = self.make_node(inputs[0], eval_points[1]).outputs[0]
+            else:
+                rval += self.make_node(inputs[0], eval_points[1]).outputs[0]
+        return [rval]
```
lamblin: There has been some confusion regarding whether the output should be `None` if *either* `eval_point` is missing, or if *both* are missing.
Since this is the same as `ConvOp`'s implementation, I think it is good for now.

lamblin:
OK to merge if Travis passes.

PR #3905, opened 2016-01-21
Don't use CorrMM when there is no c compiler (no perform, it will fal…

nouiz:
…lback to the old conv that have perform)


vdumoulin:
@nouiz the build has passed.

lamblin:
We should do the same if there is no usable blas, otherwise it will crash during compilation of the C code.

PR #3911, opened 2016-01-21
Add an "as_buffer" member function to CudaNdarray object

hma02:
Similar to the pycuda.gpuarray.GPUArray.gpudata.as_buffer function, user can use this function to get the memory of a CudaNdarray as a Python Buffer object. This function can be commonly used in GPU memory transferring based on MPI function in multi-GPU training.

See this issue on Google Groups:
https://groups.google.com/forum/#!searchin/theano-dev/theano.misc.pycuda_utils.to_gpuarray()/theano-dev/4hhFsteAq2I/beynb2UNAQAJ

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -657,6 +657,35 @@ PyObject * CudaNdarray_As_Buffer(CudaNdarray * self, PyObject *args)
 
     //printf("offset = %d size = %d\n", offset,size);
 
+    //check unified addressing  
+    
+    int device;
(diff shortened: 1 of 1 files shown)


abergeron:
This can only work on GPUs which support unified addressing.  You should at least check for that before returning the buffer.

Also, we should make sure it is enabled.

hma02:
That's right. So I just added a simple check of unified addressing. However, I think this check should be done outside the function and be read as something like a global variable. But I'm not sure where should that go. Any hint?

nouiz:
This was in the old GPU back-end that is removed. So closing.

PR #3913, opened 2016-01-22
Make sure imshp and kshp are constant

lamblin:
Otherwise, if they are symbolic variables, they could be
used during shape inference, which could introduce dependencies
on variables outside of the graph.

I'm not sure what the best way of raising the exception again with a different error message should be.

### DIFF
--- a/theano/sandbox/gpuarray/tests/test_abstractconv.py
+++ b/theano/sandbox/gpuarray/tests/test_abstractconv.py
@@ -1,10 +1,13 @@
 from nose.plugins.skip import SkipTest
 
+import numpy
+
 from theano.tensor.nnet.tests import test_abstract_conv
-from ..type import GpuArrayType, gpuarray_shared_constructor
+from ..type import GpuArrayType, gpuarray_shared_constructor, get_context
 from ..dnn import dnn_available, GpuDnnConv, GpuDnnConvGradW, GpuDnnConvGradI
 
 from .config import mode_with_gpu, test_ctx_name
+from pygpu import gpuarray
 
 gpu_ftensor4 = GpuArrayType(dtype='float32', broadcastable=(False,) * 4)
 
@@ -43,3 +46,6 @@ class TestDnnConvTypes(test_abstract_conv.TestConvTypes):
         self.input = gpu_ftensor4()
         self.filters = gpu_ftensor4()
         self.topgrad = gpu_ftensor4()
+        self.constant_tensor = gpuarray.array(
+            numpy.zeros((3, 5, 7, 11), dtype='float32'),
+            context=get_context(test_ctx_name))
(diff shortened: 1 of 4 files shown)


lamblin:
It does not work yet in all cases, and some clean-up is needed.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -246,7 +249,25 @@ def __init__(self,
                 ' integers'.format(border_mode))
 
         self.imshp = tuple(imshp) if imshp else None
+        for imshp_i in self.imshp:
```
nouiz: This would crash is imshp is None.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -246,7 +249,25 @@ def __init__(self,
...
+            if imshp_i is not None:
+                # Components of imshp should be constant or ints
+                try:
+                    get_scalar_constant_value(imshp_i)
+                except NotScalarConstantError:
+                    _logger.error("imshp should be None or "
+                                  "a tuple of constant int values")
+                    raise
         self.kshp = tuple(kshp) if kshp else None
+        for kshp_i in self.kshp:
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -246,7 +249,25 @@ def __init__(self,
                 ' integers'.format(border_mode))
 
         self.imshp = tuple(imshp) if imshp else None
+        for imshp_i in self.imshp:
+            if imshp_i is not None:
+                # Components of imshp should be constant or ints
+                try:
+                    get_scalar_constant_value(imshp_i)
+                except NotScalarConstantError:
+                    _logger.error("imshp should be None or "
```
nouiz: We use this type of code elsewhere to add to the error message:

```
        try:
            z[0] = numpy.asarray(numpy.dot(x, y))
        except ValueError as e:
            # The error raised by numpy has no shape information, we mean to 
            # add that
            e.args = e.args + (x.shape, y.shape)
            raise
```

lamblin:
Updated and rebased.

lamblin:
Also makes it possible to pass a `numpy.ndarray`, `CudaNdarray` or `gpuarray.array` directly.
Tests in `theano/sandbox/{cuda,gpuarray}/tests/test_abstractconv.py` passed locally.

theano/sandbox/gpuarray/tests/test_abstractconv.py
```diff
@@ -1,7 +1,10 @@
 from nose.plugins.skip import SkipTest
 
+import numpy
+from pygpu import gpuarray
```
nouiz: The pygpu import should be after line 10, otherwise, it make tests crash when pygpu isn't there like on travis.

The other travis errors are fixed by https://github.com/Theano/Theano/pull/4176

lamblin: OK.

lamblin:
Updated & rebased.

PR #3915, opened 2016-01-22
remove the mistakenly copied file ':w'

Saizheng:
@lamblin  This is to fix the problem in https://github.com/Theano/Theano/pull/3870


abergeron:
I don't think we need to wait on travis for this one.

PR #3916, opened 2016-01-22
Option for passing additional libraries for the GCC.

machpavel:
When using Theano on windows, I ran into a problem with linking - I needed to link package libpython.
So in .theanorc, I added

> [gcc]
> cxxflags = -Lc:\Anaconda3\libs -llibpython34

but that did not solve the issue.
Finally, I find out that it is because cxxflags are passed as the first arguments for the GCC - but GCC only accepts -l switch when it is passed AFTER source files.
Therefore, I modified python code that generates the GCC command, so it can accept additional libraries from the configuration. Now, my GCC section of .theanorc looks like this and the problem is solved:

> [gcc]
> cxxflags = -shared -Ic:\mingw\mingw32\include -Lc:\Anaconda3\libs -DMS_WIN64
> libraries = libpython34

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -191,6 +191,11 @@ def filter_nvcc_flags(s):
             " nvcc.flags value is '%s'" % s)
     return ' '.join(flags)
 
+AddConfigVar('gcc.libraries',
+             "Extra libraries to link with",
+             StrParam(""),
+             in_c_key=False)
+
 AddConfigVar('nvcc.flags',
              "Extra compiler flags for nvcc",
              ConfigParam("", filter_nvcc_flags),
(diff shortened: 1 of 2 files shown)


lamblin:
If I remember correctly, `cxxflags` already get split between different types of options when parsed.
I don't think there is a need to add a new option, I think it should be possible to pass the arguments gotten from `cxxflags` in the order expected by `gcc` instead.
I'll try to have a look.

lamblin:
OK, so we don't actually do that for `cxxflags` apparently, but we have the logic somewhere and we apply it to `blas.ldflags`, so we could generalize it to other similar flags.

machpavel:
OK, I undestand that similar problem arises elsewhere. What do you suggest as a solution to my problem? Should I implement some parsing of `cxxflags`?

nouiz:
You can call theano.tensor.blas._ldflags() To split the parameter. There is no need to generalize the code, it was already done. You just need to use it.

thanks

nouiz:
No news for too long and I didn't see anyone request such stuff. So closing.

PR #3918, opened 2016-01-23
Doc installation on Tegra TX1 board.

nouiz:


nouiz:
@lamblin, can you check this PR? It is pretty quick.

theano/misc/check_blas.py
```diff
@@ -272,6 +272,7 @@ def test():
         GTX 660           2.32s  2.32s
         GTX 580           2.42s         2.47s
         GTX 480           2.87s         2.88s
+        TX1                      7.6s
```
titusnicolae: is this comparison legitimate since TX1 does FP16 not FP32?

nouiz: This is in float32, not float16. Theano did not implement computation in
float16.

titusnicolae: isn't TX1 designed to work well with half-precision FP16? it might mislead people that it's slow. anyway

PR #3926, opened 2016-01-25
Alloc pickling of GpuArray object

nouiz:
I didn't test with context, but the code maybe support it. At least, it won't cause problem with old pickle due to interface change.

We can't put it easily in pygpu as there is a Theano flags that control the behavior of what should happen.


theano/sandbox/gpuarray/type.py
```diff
@@ -751,3 +754,34 @@ def c_code_cache_version(self):
...
+        # directly return numpy array
+        warnings.warn(
+            "config.experimental.unpickle_gpu_on_cpu is set to True. "
+            "Unpickling GpuArray as numpy.ndarray")
+        return npa
+    elif pygpu:
+        ctx = get_context(ctx_name)
+        return pygpu.gpuarray.array(npa, copy=True, context=ctx)
+    else:
+        raise ImportError("Cuda not found. Cannot unpickle GpuArray")
```
abergeron: Maybe the error message need to not reference cuda.

theano/sandbox/gpuarray/type.py
```diff
@@ -12,7 +15,7 @@
     from pygpu import gpuarray
     from pygpu.elemwise import compare, elemwise2
 except ImportError:
-    pass
+    pygpu = None
```
abergeron: Why are you adding this?

theano/sandbox/gpuarray/tests/test_type.py
```diff
@@ -1,12 +1,22 @@
...
 from theano.compile import DeepCopyOp
+from theano.misc.pkl_utils import CompatUnpickler
 
 from .config import test_ctx_name
 from .test_basic_ops import rand_gpuarray
 
 from ..type import GpuArrayType
+from .. import pygpu_activated
+
+if pygpu_activated:
```
abergeron: Since you import `.config`, you can just import pygpu directly.

The `.config` import will raise an exception if pygpu is not available.

theano/sandbox/gpuarray/tests/GpuArray.pkl
```diff
@@ -0,0 +1,41 @@
...
+I-1
+I0
+tp13
+bI00
+S'\x00\x00(\xc2'
+p14
+tp15
+bNtp16
+Rp17
+.
```
abergeron: You pickle is saved with protocol 0, which almost nobody uses.

theano/sandbox/gpuarray/type.py
nouiz: to detect when pygpu isn't there around line 770 in this file.

nouiz:
done

theano/sandbox/gpuarray/tests/GpuArray.pkl
nouiz: This is for test. More safe. This is a very small file anyway.

abergeron: ok

nouiz:
I separted the test that must always run to another file as you suggested.

This made me uncover another problem with a test, it wasn't testing correctly and when I enable it correctly, it was failing. So I implemented the same casting rule in the new back-end as the old back-end and the CPU.

nouiz:
done

theano/sandbox/gpuarray/type.py
```diff
@@ -751,3 +758,34 @@ def c_code_cache_version(self):
 declaration of an operation.
 """
 gpu_context_type = GpuContextType()
+
+
+# THIS WORKS But CudaNdarray instances don't compare equal to one
```
lamblin: Maybe update the comment to say "GpuArray", and mention that this is not a new issue as it is already present in CudaNdarray?

abergeron: Also, numpy arrays do the same thing.  So I'm pretty sure the cmodule cache has tricks to not save arrrays.

nouiz: I'm not sure this is still a useful comment. I don't want to go further in this PR as this work and leave Theano in a better state.

If you think it is important to understand if this comment is still true and investigate, we should open an issue about this. But I would consider this as non important as it work and I don't see a problem with this. But I could be wrong.

I updated the PR and made it skip travis that is overloaded.

lamblin:
OK.

Issue #3752, opened 2015-12-06
h_softmax is unreasonably slow when target is None


shawnjhenry:
I gave the new nnet h_softmax function  a whirl and it's super fast for training, but it seems like it's unreasonably slow when computing the full probability (>1min for 5000x512 with 200 classes of size 200).  I can post code if necessary, but the same computation with scan instead of sparse_block_dot it is more than 10x faster for large pieces of data.  Does theano not have a batch matrix-vector product?

bouthilx:
```sparse_block_dot``` is probably not optimal for such dense matrix multiplications. We used it because the implementation/semantic was very similar to ```h_sotmax(..., target != None)```. You are welcome to make a pull request to improve the speed, it would also be great if you can profile the old and new ```h_softmax``` version for a few different sizes of targets/batch.

**Comment by shawnjhenry** [comment]
I think I've found a good solution: replace sparse_block_dot with tensordot.  The following code shows the difference for a very large batch size, although tensordot is always faster:  
```
import numpy
import theano
import theano.tensor as T
from theano import function, shared

a = numpy.random.uniform(-0.08, 0.08, (5000, 512))

x = T.matrix()
batch_size = 5000
n_outputs = 40000
n_classes = 200
n_outputs_per_class = 200
W1 = shared(numpy.random.uniform(-0.08, 0.08, (512, 200)))
b1 = shared(numpy.random.uniform(-0.08, 0.08, (200)))
W2 = shared(numpy.random.uniform(-0.08, 0.08, (200, 512, 200)))
b2 = shared(numpy.random.uniform(-0.08, 0.08, (200, 200)))

class_ids = T.tile(T.arange(n_classes, dtype="int32")[None, :], (batch_size, 1))

activations1 = T.nnet.blocksparse.sparse_block_dot(W2[None, :, :, :], x[:, None, :], T.zeros((batch_size, 1), dtype='int32'), b2, class_ids)
activations2 = (T.tensordot(x, W2, (1, 1))+b2)

f1 = function([x], activations1, profile = True)
f2 = function([x], activations2, profile = True)

out1 =  f1(a)
out2 =  f2(a)
print(numpy.linalg.norm(out1-out2))  #sanity check that these are computing the same thing
```
I'll spare you the whole profile, but I get the following:
```
Time in 1 calls to Function.__call__: 7.311755e+01s
Time in 1 calls to Function.__call__: 2.521929e+00s 
5.40312422957e-13
```
[...]

PR #3885, opened 2016-01-19
Faster travis

abergeron:
Since travis upped their job length restriction to 2h, we can merge some parts together.  This will save a tiny bit on setup time, but also possibly enable us to test a PR with a single set of jobs that run concurrently.

Here I'm adapting travis.yml by steps to check the timings.

PR #3931, opened 2016-01-25
Speed up h_softmax when full output is requested.

adbrebs:
cf #3752, thanks @shawnjhenry!

In h_softmax, replace sparse_block_dot with tensor_block when full output is requested.

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -2293,14 +2293,8 @@ def h_softmax(x, batch_size, n_outputs, n_classes, n_outputs_per_class,
 
     if target is None:  # Computes the probabilites of all the outputs
 
-        class_ids = tensor.tile(
-            tensor.arange(n_classes, dtype="int32")[None, :], (batch_size, 1))
-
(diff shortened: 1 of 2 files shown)


lamblin:
I guess there should be a test for that.

nouiz:
In theano/tensor/nnet/test_nnet.py, there is already a test that execute with target=None. But it only check the shape of the result, not the value. Can you add a check that the value is what is expected?

 Can you confirm that what this do is a speed up and nothing else?

thanks

adbrebs:
> In theano/tensor/nnet/test_nnet.py, there is already a test that execute with target=None. But it only check the shape of the result, not the value. Can you add a check that the value is what is expected?

Ok I will do it next week.

> Can you confirm that what this do is a speed up and nothing else?
Yes, this yields a 2-3x speedup in my experiments.

lamblin:

adbrebs:
I have added a test suggested by @lamblin. It checks if un_output_tg and fun_output computes the same thing.

nouiz:
thanks

PR #3933, opened 2016-01-26
profile_memory, print the expected peak memory usage if excluding inplace opt.

nouiz:
Also code clean up, refactoring and better output of the information.


lamblin:
This breaks a test that expects the old messages to be printed: https://github.com/nouiz/Theano/blob/profile/theano/compile/tests/test_profiling.py#L52-L65

nouiz:
I update the test

lamblin:
I'm pretty sure the lines in the `else:` block should be updated as well.

nouiz:
done

PR #3935, opened 2016-01-27
add missing return to GpuFromHost R_op

poolio:
Add missing `return` to the `R_op` for `GpuFromHost`.


lamblin:
Thanks!

PR #3936, opened 2016-01-27
Mostly tests fix for 32bit python

nouiz:
Also remove an old restriction that bincount can't work with uint64

### DIFF
--- a/theano/tensor/tests/test_extra_ops.py
+++ b/theano/tensor/tests/test_extra_ops.py
@@ -156,13 +156,25 @@ class TestBinCountOp(utt.InferShapeTester):
                 f1 = theano.function([x], bincount(x))
                 f2 = theano.function([x, w], bincount(x, weights=w))
 
-                assert (np.bincount(a) == f1(a)).all()
-                assert np.allclose(np.bincount(a, weights=weights),
-                                   f2(a, weights))
-                f3 = theano.function([x], bincount(x, minlength=23))
+                def ref(data, w=None, minlength=None):
+                    size = data.max() + 1
+                    if minlength:
+                        size = max(size, minlength)
+                    if w:
+                        out = np.zeros(size, dtype=weights.dtype)
+                        for i in range(data.shape[0]):
+                            out[data[i]] += weights[i]
+                    else:
+                        out = np.zeros(size, dtype=a.dtype)
+                        for i in range(data.shape[0]):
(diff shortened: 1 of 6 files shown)


theano/tensor/tests/test_extra_ops.py
```diff
@@ -138,36 +138,36 @@ def test_bincountFn(self):
...
+            f2 = theano.function([x, w], bincount(x, weights=w))
+
+            def ref(data, w=None, minlength=None):
+                size = data.max() + 1
+                if minlength:
+                    size = max(size, minlength)
+                if w:
+                    out = np.zeros(size, dtype=weights.dtype)
+                    for i in range(data.shape[0]):
+                        out[data[i]] += weights[i]
```
lamblin: This block should use `w` instead of `weights`, I think.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -138,36 +138,36 @@ def test_bincountFn(self):
...
+                else:
+                    out = np.zeros(size, dtype=a.dtype)
+                    for i in range(data.shape[0]):
+                        out[data[i]] += 1
+                return out
+            assert (ref(a) == f1(a)).all()
+            assert np.allclose(ref(a, w), f2(a, weights))
```
lamblin: And here we should call `ref(a, weights)` or `ref(a, w=weights)`.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -424,6 +424,7 @@ def test_repeatOp(self):
                     r_var = T.scalar(dtype=dtype)
                     r = numpy.asarray(3, dtype=dtype)
                     if dtype in self.numpy_unsupported_dtypes:
+                        r_var = T.vector(dtype=dtype)
```
lamblin: Why that change?

lamblin: If the error is  not raised for a `T.scalar`, then we should test that the value is right, like in the `else:` block.

nouiz:
I fixed the comment and added 2 other fix in commit after last comments of @lamblin

carriepl:
Looks mostly good to me but I did not look much at the part that had already been reviewed by @lamblin.

@lamblin : if you are satisfied with how the points you raised were addressed, you have my blessing to merge.

PR #3937, opened 2016-01-27
ACDC op (GPU version)

alancucki:
Port of the ACDC layer (M.Moczulski, M. Denil, J. Appleyard, N. de Freitas, "ACDC: A Structured Efficient Linear Layer", http://arxiv.org/abs/1511.05946 ) from https://github.com/mdenil/acdc-torch for the GPU.

### DIFF
--- a/theano/sandbox/cuda/acdc.py
+++ b/theano/sandbox/cuda/acdc.py
@@ -6,33 +6,18 @@
 import os
 
 import theano
-import copy
-from theano.sandbox.cuda import cuda_available, GpuOp
-from theano.sandbox.cuda.basic_ops import (as_cuda_ndarray_variable, gpu_contiguous)
-from theano.sandbox.cuda.nvcc_compiler import NVCC_compiler
+from theano.sandbox.cuda import GpuOp
+from theano.sandbox.cuda.basic_ops import as_cuda_ndarray_variable, gpu_contiguous
 
 import cuda_ndarray
 
-if cuda_available:
-    from theano.sandbox.cuda import CudaNdarrayType
-    from theano.sandbox.cuda.basic_ops import host_from_gpu, gpu_from_host, HostFromGpu
-    from theano.sandbox.cuda import register_opt as register_gpu_opt
-
 class FastACDC(GpuOp):
 
-    def __init__(self):
-        self.max_threads_dim0 = None
-        self.max_grid_size1 = None
-        self.max_grid_size2 = None
-
-    def c_compiler(self):
-        return NVCC_compiler
-
     def c_headers(self):
-        return ['<stdio.h>', '<cuda.h>', '<cufft.h>', 'cuda_ndarray.cuh']
+        return ['<stdio.h>', '<cufft.h>']
 
     def c_libraries(self):
-        return ['cuda', 'cufft']
+        return ['cufft']
 
     def make_node(self, x, A, Ab, D, Db):
         x_ = as_cuda_ndarray_variable(x)
@@ -40,38 +25,15 @@ class FastACDC(GpuOp):
         Ab_ = as_cuda_ndarray_variable(Ab)
         D_ = as_cuda_ndarray_variable(D)
         Db_ = as_cuda_ndarray_variable(Db)
-
-        assert x_.dtype == 'float32'
+        for t in [x_, A_, Ab_, D_, Db_]:
+            assert t.dtype == 'float32'
         return theano.Apply(self, [x_, A_, Ab_, D_, Db_], [x_.type()])
 
-    def make_thunk(self, node, storage_map, compute_map, no_recycling):
-        node_ = copy.copy(node)
-        assert node.op is node_.op
-        if node_.op.max_threads_dim0 is None or node_.op.max_grid_size1 is None or node_.op.max_grid_size2 is None:
-            cuda = theano.sandbox.cuda
-            device_id = cuda.use.device_number
-            if device_id is None:
-                cuda.use("gpu",
-                         force=False,
-                         default_to_move_computation_to_gpu=False,
-                         move_shared_float32_to_gpu=False,
-                         enable_cuda=False,
-                         test_driver=True)
-                device_id = cuda.use.device_number
-            cuda_ndarray = theano.sandbox.cuda.cuda_ndarray.cuda_ndarray
-            prop = cuda_ndarray.device_properties(device_id)
-            node_.op.max_threads_dim0 = prop['maxThreadsDim0']
-            node_.op.max_grid_size1 = prop['maxGridSize1']
-            node_.op.max_grid_size2 = prop['maxGridSize2']
-
-        return super(FastACDC, node_.op).make_thunk(node_, storage_map,
-                                                      compute_map, no_recycling)
     def grad(self, inputs, cost_grad):
         inp, A, Ab, D, Db = inputs
         gradOutput, = cost_grad
         gradOutput = gpu_contiguous(gradOutput)
-        fast_acdc_grad = FastACDCGrad()
-        return fast_acdc_grad(inp, gradOutput, A, Ab, D, Db)
+        return FastACDCGrad()(inp, gradOutput, A, Ab, D, Db)
 
     def c_support_code_apply(self, node, nodename):
         # Based on a similar mechanism in sandbox/cuda/blas.py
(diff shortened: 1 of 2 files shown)


theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+        return ['cufft']
+
+    def make_node(self, x, A, Ab, D, Db):
+        x_ = gpu_contiguous(as_cuda_ndarray_variable(x))
+        A_ = gpu_contiguous(as_cuda_ndarray_variable(A))
+        Ab_ = gpu_contiguous(as_cuda_ndarray_variable(Ab))
+        D_ = gpu_contiguous(as_cuda_ndarray_variable(D))
+        Db_ = gpu_contiguous(as_cuda_ndarray_variable(Db))
+        for t in [x_, A_, Ab_, D_, Db_]:
+            assert t.dtype == 'float32'
```
nouiz: In the old back-end, only the float32 dtypes is possible. So no need of this check.

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+        return theano.Apply(self, [x_, A_, Ab_, D_, Db_], [x_.type()])
```
nouiz: If the output shape is the same as x, I would add this method in the code:
```
    return [shape[0]]
```
Otherwise, what is the shape relation of the output again the input?

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def c_support_code_apply(self, node, nodename):
+        # Based on a similar mechanism in sandbox/cuda/blas.py
+        #
+        # REMEMBER TO RAISE c_code_cache_version when changing any of
+        # these files
+        files = ['acdc.cu']
+        codes = [open(os.path.join(os.path.split(__file__)[0], f)).read()
+                for f in files]
+        return reduce(str.__add__, codes)
```
nouiz: This use the old inferface. A better one that don't request such hack but request some refactoring would be ideal, but currently this is working fine. If you want to do the refactoring, tell me and I can guide you. We didn't update all code in Theano as this is working fine.

You should add the c_code_cache_version() method:
```
    return (1,)
```
Otherwise, it will get recompiled for each process.

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+        Db_ = gpu_contiguous(as_cuda_ndarray_variable(Db))
+        for t in [x_, A_, Ab_, D_, Db_]:
+            assert t.dtype == 'float32'
+        return theano.Apply(self, [x_, A_, Ab_, D_, Db_], [x_.type()])
+
+    def grad(self, inputs, cost_grad):
+        inp, A, Ab, D, Db = inputs
+        gradOutput, = cost_grad
+        gradOutput = gpu_contiguous(gradOutput)
+        return FastACDCGrad()(inp, gradOutput, A, Ab, D, Db)
```
nouiz: building  a python object is a little costly, so we use the pattern at many place. This also show the user a better interface that don't request them to instantiate an op.

class MyOp(Op):
    ...
my_op = MyOp()

then people can use my_op(some_theano_variables, ...)

You can do this for both op and reuse the fast_acdc_grad instance here.

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def c_code(self, node, name, inputs, outputs, sub):
+        x, A, Ab, D, Db = inputs
+        z, = outputs
+        fail = sub['fail']
+        return """
+{
+    // Validate that the output storage exists.
+    if (NULL == %(z)s ||
+        CudaNdarray_NDIM(%(z)s) != CudaNdarray_NDIM(%(x)s))
```
nouiz: The output can be of different shape as the user can call Theano function with inputs of different shape to different call. If the output have the same shape as the input as in this case, you can use this method CudaNdarray_prep_output() that will do all that for you.

theano/sandbox/cuda/acdc.py
nouiz: Adding some assert about ndim is always a good idea.

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+    CudaNdarray* tmp2 = (CudaNdarray*)CudaNdarray_NewDims(1, &tmp_len);
+
+    // Verify.
+    if (!%(x)s || !%(z)s || !%(A)s || !%(Ab)s || !%(D)s || !%(Db)s || !tmp1 || !tmp2) {
+        PyErr_Format(PyExc_ValueError, "Could not allocate storage");
+        %(fail)s;
+    }
+
+    // cudaStream_t stream = THCState_getCurrentStream(state);
+    cudaStream_t stream = 0; // Assign the default stream?
```
nouiz: Theano current back-end use only one stream, so yes, this is what is needed to do.

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def make_node(self, inp, gradOutput, A, Ab, D, Db):
+        inp_ = gpu_contiguous(as_cuda_ndarray_variable(inp))
+        gradOutput_ = gpu_contiguous(as_cuda_ndarray_variable(gradOutput))
+        A_ = gpu_contiguous(as_cuda_ndarray_variable(A))
+        Ab_ = gpu_contiguous(as_cuda_ndarray_variable(Ab))
+        D_ = gpu_contiguous(as_cuda_ndarray_variable(D))
+        Db_ = gpu_contiguous(as_cuda_ndarray_variable(Db))
+        for t in [inp_, gradOutput_, A_, Ab_, D_, Db_]:
+            assert t.dtype == 'float32'
```
nouiz: idem, no need to assert float32, but assert ndim.

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def c_code(self, node, name, inputs, outputs, sub):
+        inp, gradOutput, A, Ab, D, Db = inputs
+        gradInput, gradA, gradAb, gradD, gradDb = outputs
+        fail = sub['fail']
+        return """
+{
+    // Validate output storage.
+    if (NULL == %(gradInput)s ||
+        CudaNdarray_NDIM(%(gradInput)s) != CudaNdarray_NDIM(%(gradOutput)s))
```
nouiz: idem, use CudaNdarray_prep_output

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+    // cudaStream_t stream = THCState_getCurrentStream(state);
+    cudaStream_t stream = 0; // Assign the default stream?
+
+    // It should be up to the calling framework to provide plans.
+    // This is a hack to make it work, though will be very slow if the
+    // length/batchSize/groupSize change frequently.
+    static cufftHandle planR2C;
+    static cufftHandle planC2C;
+    static int planLength = -1;
+    static int planBatchSize = -1;
```
nouiz: Those will be shared by each op instance that get compiled. I think we want one cache per op instance. This would allow to not recompute it if we change the batch size for the test and train. This can be done by defining those methods instead of those static declaration:

```
    def c_support_code_struct(self, node, name):
        return """
cufftHandle planR2C%(name)s;
cufftHandle planC2C%(name)s;
int planLength%(name)s = -1;
int planBatchSize%(name)s = -1;
""" % dict(name=name)

    def c_init_code_struct(self, node, name, sub):
        return """
cufftHandle planR2C%(name)s = NULL;
cufftHandle planC2C%(name)s = NULL;
""" % dict(name=name)

    def c_cleanup_code_struct(self, node, name):
        return """
          cufftDestroy%(name)s(planR2C);
          cufftDestroy%(name)s(planC2C);
""" % dict(name=name)
```

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+
+    // Arrays grad{A,Ab,D,Db} have to be initialized to zeros. See FastACDC.lua.
+    size_t total_size = input_size * sizeof(real);
+    cudaMemset(CudaNdarray_DEV_DATA(%(gradA)s), 0, total_size);
+    cudaMemset(CudaNdarray_DEV_DATA(%(gradAb)s), 0, total_size);
+    cudaMemset(CudaNdarray_DEV_DATA(%(gradD)s), 0, total_size);
+    cudaMemset(CudaNdarray_DEV_DATA(%(gradDb)s), 0, total_size);
+
+    // cudaStream_t stream = THCState_getCurrentStream(state);
+    cudaStream_t stream = 0; // Assign the default stream?
```
nouiz: idem

theano/sandbox/cuda/acdc.py
```diff
@@ -0,0 +1,309 @@
...
+
+    // It should be up to the calling framework to provide plans.
+    // This is a hack to make it work, though will be very slow if the
+    // length/batchSize/groupSize change frequently.
+    static cufftHandle planR2C;
+    static cufftHandle planC2C;
+    static int planLength = -1;
+    static int planBatchSize = -1;
```
nouiz: idem

theano/sandbox/cuda/acdc.py
nouiz: Also, this change the name of the variable to end with %(name)s, don't forget to update it elsewhere. I just updated the comment above, so take the version in github, not in the email you got.

PR #3940, opened 2016-01-27
Fix most of the remaining DeprecationWarnings in the tests.

abergeron:


lamblin:
OK when travis passes.

PR #3942, opened 2016-01-28
change shape dimensions int32 to int64

shabanian:


theano/tensor/raw_random.py
```diff
@@ -520,13 +520,8 @@ def binomial(random_state, size=None, n=1, p=0.5, ndim=None,
...
     p = tensor.as_tensor_variable(p)
     ndim, size, bcast = _infer_ndim_bcast(ndim, size, n, p)
-    if n.dtype == 'int64':
-        # THIS WORKS AROUND A NUMPY BUG on 32bit machine
-        # Erase when the following works on a 32bit machine:
-        # numpy.random.binomial(
-        #          n=numpy.asarray([2,3,4], dtype='int64'),
-        #          p=numpy.asarray([.1, .2, .3], dtype='float64'))
-        n = tensor.cast(n, 'int32')
+    # THIS WORKS AROUND A NUMPY BUG on 32bit machine
```
lamblin: Actually, there is an issue here. I checked the case in the comment, and there is still an issue on 32-bits installations of Python. So this change would probably break things...

lamblin: So, the other changes in the file are OK, but this one (and the comment) should be reverted for now.
We should decide whether we want to cast to int32 on 32-bit machines only, or always.

theano/tensor/raw_random.py
```diff
@@ -520,13 +520,8 @@ def binomial(random_state, size=None, n=1, p=0.5, ndim=None,
     n = tensor.as_tensor_variable(n)
     p = tensor.as_tensor_variable(p)
     ndim, size, bcast = _infer_ndim_bcast(ndim, size, n, p)
-    if n.dtype == 'int64':
-        # THIS WORKS AROUND A NUMPY BUG on 32bit machine
-        # Erase when the following works on a 32bit machine:
```
nouiz: This should be in a test, we run the daily buildbot with an python 32 bit, so we would catch regression about this.

theano/tensor/raw_random.py
lamblin: @shabanian Can you revert this part, please?

theano/tensor/raw_random.py
lamblin: @nouiz I'm not sure what you mean by that. What should be in a test?
As far as I understand it, the comment is not about a regression, it is about something that we hoped would be changed in the future.

nouiz: I would put that code in the test. In fact, depending of the consequence of the bug (error vs segfault) we could detect it here and apply the work around only if we need it. I just check it with numpy 1.8.0 and this raise an error. So I would change all this to:


```
if n.dtype == 'int64':
   try:
       numpy.random.binomial(n=numpy.asarray([2,3,4], dtype='int64'), p=numpy.asarray([.1, .2, .3], dtype='float64'))
   except TypeError:
      n = tensor.cast(n, 'int32')
```
So we will apply the fix only when needed. So we won't risk a silent bug with python 64 bit.

Something even better would be to modify RandomFunction to include a hack to check if all the int64 values fit in a int32 values, but this is probably too much.

What do you think?

lamblin: Oh, OK. Yes, we could do that.

shabanian:
Done!

shabanian:
@adri-romsor

theano/tensor/raw_random.py
```diff
@@ -521,12 +521,12 @@ def binomial(random_state, size=None, n=1, p=0.5, ndim=None,
...
-        # numpy.random.binomial(
-        #          n=numpy.asarray([2,3,4], dtype='int64'),
-        #          p=numpy.asarray([.1, .2, .3], dtype='float64'))
-        n = tensor.cast(n, 'int32')
+        try:
+            numpy.random.binomial(n=numpy.asarray([2, 3, 4], dtype='int64'), p=numpy.asarray([.1, .2, .3], dtype='float64'))
+        except TypeError:
+            n = tensor.cast(n, 'int32')
+    # THIS WORKS AROUND A NUMPY BUG on 32bit machine
```
lamblin: I think this section needs to be reorganized. The comment needs to go before the "try". Also, currently the code casts to int32, and then casts to int64, which defeats the purpose. What we want is for `n` to be casted in int64, except when a TypeError is raised, in which case it should be int32.
There are different ways of achieving that.

PR #3943, opened 2016-01-28
Fix path for BatchedDot

lamblin:
This fixes an ImportError when importing `theano.sandbox.cuda`.


carriepl:
This fix is in sandbox cuda so it won't be tested by Travis (so waiting for Travis to complete is superfluous) and it fixes a current issue in the master so I'm merging.

PR #3950, opened 2016-01-29
C implementation of sample without replacement OP

aalmah:
Addressing #3143. One test is still failing but I made a PR for reference.

### DIFF
--- a/theano/sandbox/multinomial.py
+++ b/theano/sandbox/multinomial.py
@@ -215,6 +215,112 @@ class MultinomialWOReplacementFromUniform(MultinomialFromUniform):
         out = T.tensor(dtype=odtype, broadcastable=pvals.type.broadcastable)
         return Apply(self, [pvals, unis, as_scalar(n)], [out])
 
+    def c_code_cache_version(self):
+        return None
+
+    def c_code(self, node, name, ins, outs, sub):
+        (pvals, unis, n) = ins
+        (z,) = outs
+        if self.odtype == 'auto':
+            t = "NPY_INTP"
+        else:
+            t = theano.scalar.Scalar(self.odtype).dtype_specs()[1]
+            if t.startswith('theano_complex'):
+                t = t.replace('theano_complex', 'NPY_COMPLEX')
+            else:
+                t = t.upper()
+        fail = sub['fail']
+        return """
+        if (PyArray_NDIM(%(pvals)s) != 2)
+        {
+            PyErr_Format(PyExc_TypeError, "pvals wrong rank");
+            %(fail)s;
+        }
+        if (PyArray_NDIM(%(unis)s) != 1)
+        {
+            PyErr_Format(PyExc_TypeError, "unis wrong rank");
+            %(fail)s;
+        }
+
+        if ( %(n)s > (PyArray_DIMS(%(pvals)s)[1]) )
+        {
+            PyErr_Format(PyExc_ValueError, "n > pvals.shape[1]");
+            %(fail)s;
+        }
+
+        if (PyArray_DIMS(%(unis)s)[0] != (PyArray_DIMS(%(pvals)s)[0] * %(n)s))
+        {
+            PyErr_Format(PyExc_ValueError, "unis.shape[0] != pvals.shape[0] * n");
+            %(fail)s;
+        }
+
+        if ((NULL == %(z)s)
+            || ((PyArray_DIMS(%(z)s))[0] != (PyArray_DIMS(%(pvals)s))[0])
+            || ((PyArray_DIMS(%(z)s))[1] != %(n)s)
+        )
+        {
+            Py_XDECREF(%(z)s);
+            npy_intp dims[2];
+            dims[0] = PyArray_DIMS(%(pvals)s)[0];
+            dims[1] = %(n)s;
+            %(z)s = (PyArrayObject*) PyArray_EMPTY(2,
+                dims,
+                %(t)s,
+                -1);
+            if (!%(z)s)
+            {
+                PyErr_SetString(PyExc_MemoryError, "failed to alloc z output");
+                %(fail)s;
+            }
+        }
+
+        { // NESTED SCOPE
+
+        const int nb_multi = PyArray_DIMS(%(pvals)s)[0];
+        const int nb_outcomes = PyArray_DIMS(%(pvals)s)[1];
+        const int n_samples = %(n)s;
+
+        //
+        // For each multinomial, loop over each possible outcome,
+        // and set selected pval to 0 after being selected
+        //
+        for (int c = 0; c < n_samples; ++c){
+            for (int n = 0; n < nb_multi; ++n)
+            {
+                double cummul = 0.;
+                const dtype_%(unis)s* unis_n = (dtype_%(unis)s*)PyArray_GETPTR1(%(unis)s, c*nb_multi + n);
+                for (int m = 0; m < nb_outcomes; ++m)
+                {
+                    dtype_%(z)s* z_nc = (dtype_%(z)s*)PyArray_GETPTR2(%(z)s, n, c);
+                    dtype_%(pvals)s* pvals_nm = (dtype_%(pvals)s*)PyArray_GETPTR2(%(pvals)s, n, m);
+                    cummul += *pvals_nm;
+                    if (cummul > *unis_n)
+                    {
+                        *z_nc = m;
+                        *pvals_nm = 0.;
+                        // renormalize the nth row of pvals
+                        dtype_%(pvals)s sum = 0.;
+                        dtype_%(pvals)s* pvals_n = (dtype_%(pvals)s*)PyArray_GETPTR2(%(pvals)s, n, 0);
+                        for (int k = 0; k < nb_outcomes; ++k)
+                        {
+                            sum = sum + *pvals_n;
+                            pvals_n++;
+                        }
+                        
+                        pvals_n = (dtype_%(pvals)s*)PyArray_GETPTR2(%(pvals)s, n, 0);
+                        for (int k = 0; k < nb_outcomes; ++k)
+                        {
+                            *pvals_n = *pvals_n / sum;
+                            pvals_n++;
+                        }
+                        break;
+                    }
+                }
+            }
+        }
+        } // END NESTED SCOPE
+        """ % locals()
+
     def perform(self, node, ins, outs):
         (pvals, unis, n_samples) = ins
         # make a copy so we do not overwrite the input
@@ -253,12 +359,6 @@ class MultinomialWOReplacementFromUniform(MultinomialFromUniform):
                         pvals[n, m] = 0.
                         pvals[n] /= pvals[n].sum()
                         break
-
-    def c_code_cache_version(self):
-        return None
-
-    def c_code(self, node, name, ins, outs, sub):
-        raise NotImplementedError('no C implementation yet!')
 
 
 class GpuMultinomialFromUniform(MultinomialFromUniform, GpuOp):
--- a/theano/sandbox/multinomial.py
+++ b/theano/sandbox/multinomial.py
@@ -290,7 +290,7 @@ class MultinomialWOReplacementFromUniform(MultinomialFromUniform):
         }
 
         { // NESTED SCOPE
-        
+
         const int nb_multi = PyArray_DIMS(%(pvals)s)[0];
         const int nb_outcomes = PyArray_DIMS(%(pvals)s)[1];
         const int n_samples = %(n)s;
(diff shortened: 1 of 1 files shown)


theano/sandbox/multinomial.py
```diff
@@ -192,6 +192,181 @@ def perform(self, node, ins, outs):
                             z[0][n, m] += 1
                             break
 
+class WeightedSelectionFromUniform(Op):
+    """
+    Converts samples from a uniform into sample from a multinomial.
```
nouiz: update the comment. it isn't a multinomial anymore.

theano/sandbox/multinomial.py
```diff
@@ -192,6 +192,181 @@ def perform(self, node, ins, outs):
...
+
+    __props__ = ("odtype",)
+
+    def __init__(self, odtype):
+        self.odtype = odtype
+
+    def __str__(self):
+        return '%s{%s}' % (self.__class__.__name__, self.odtype)
+
+    def __setstate__(self, dct):
```
nouiz: delete this method, it isn't useful. Idem for `__str__`.

theano/sandbox/multinomial.py
```diff
@@ -192,6 +192,181 @@ def perform(self, node, ins, outs):
...
+            odtype = pvals.dtype
+        else:
+            odtype = self.odtype
+        out = T.tensor(dtype=odtype, broadcastable=pvals.type.broadcastable)
+        return Apply(self, [pvals, unis, as_scalar(n)], [out])
+
+    def grad(self, ins, outgrads):
+        pvals, unis, n = ins
+        (gz,) = outgrads
+        return [T.zeros_like(x) for x in ins]
```
nouiz: I'm not sure of that one, mrg return grad not defined.

theano/sandbox/multinomial.py
```diff
@@ -192,6 +192,181 @@ def perform(self, node, ins, outs):
...
+            t = "NPY_INTP"
+        else:
+            t = theano.scalar.Scalar(self.odtype).dtype_specs()[1]
+            if t.startswith('theano_complex'):
+                t = t.replace('theano_complex', 'NPY_COMPLEX')
+            else:
+                t = t.upper()
+        fail = sub['fail']
+        return """
+        if (PyArray_NDIM(%(pvals)s) != 2)
```
nouiz: I have the feeling that this is much duplicate code from MultinomialFromUniform. Would it be better to refactor MutiminalFromUniform to remove duplicate code?

theano/sandbox/multinomial.py
carriepl: @lamblin : being the grad expert, what's your take on this. should the gradient wrt to the samples from the multinomial be zeros or not defined? Personally, I'm leaning towards zeros but I'd like your opinion.

theano/sandbox/multinomial.py
```diff
@@ -192,6 +192,181 @@ def perform(self, node, ins, outs):
...
+                        {
+                            *pvals_n = *pvals_n / sum;
+                            pvals_n++;
+                        }
+                        break;
+                    }
+                }
+            }
+        }
+        } // END NESTED SCOPE
```
carriepl: During our discussion with @abergeron, didn't we settle on a different, more efficient implementation or am I misremembering?

carriepl: If this is the implementation we considered, then I think we should consider the refactoring that @nouiz mentionned.

theano/sandbox/multinomial.py
lamblin: - gradient wrt `n` should be a `DisconnectedType()` (for `MultinomialFromUniform` as well)
- for `pvals` and `unis`, I think there are arguments for both. I think that the Op itself should return `zeros`, as it is purely deterministic, and flat almost everywhere.
- However, I think it would be better if the wrapper method `multinomial` or `weighted_selection` would also add "undefined_grad" nodes, so that the whole process of "getting a sample from a multinomial" would not be derivable. We are not doing that yet in MRG, so I think it is fine for now.

carriepl:
The two links that we discussed IRL. They have some useful info.
http://docs.scipy.org/doc/numpy-1.10.0/reference/c-api.array.html
http://deeplearning.net/software/theano/tutorial/extending_theano_c.html

aalmah:
@bouthilx All problems I had earlier are fixed now

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+            else:
+                t = t.upper()
+        fail = sub['fail']
+        return """
+        // create a copy of pvals matrix
+        PyArrayObject* pvals_copy = NULL;
+
+        if (PyArray_NDIM(%(pvals)s) != 2)
+        {
+            PyErr_Format(PyExc_TypeError, "pvals wrong rank");
```
bouthilx: I understand you took hit from MultinomialFromUniform but I think the error message from make_node is more explicit. If you agree could you please use 'pvals ndim should be 2' and modify it in MultinomialFromUniform.c_code at the same time.

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+            %(fail)s;
+        }
+        if (PyArray_NDIM(%(unis)s) != 1)
+        {
+            PyErr_Format(PyExc_TypeError, "unis wrong rank");
```
bouthilx: Same here

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+            %(fail)s;
+        }
+
+        if ( %(n)s > (PyArray_DIMS(%(pvals)s)[1]) )
+        {
+            PyErr_Format(PyExc_ValueError, "n > pvals.shape[1]");
```
bouthilx: I would use the same error message as in `perform`.

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+            }
+        }
+
+        { // NESTED SCOPE
+
+        const int nb_multi = PyArray_DIMS(%(pvals)s)[0];
+        const int nb_outcomes = PyArray_DIMS(%(pvals)s)[1];
+        const int n_samples = %(n)s;
+
+        // for (int n = 0; n < nb_multi; ++n)
```
bouthilx: An artefact?

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+        //
+        // For each multinomial, loop over each possible outcome,
+        // and set selected pval to 0 after being selected
+        //
+        for (int c = 0; c < n_samples; ++c){
+            for (int n = 0; n < nb_multi; ++n)
+            {
+                double cummul = 0.;
+                const dtype_%(unis)s* unis_n = (dtype_%(unis)s*)PyArray_GETPTR1(%(unis)s, c*nb_multi + n);
+                dtype_%(z)s* z_nc = (dtype_%(z)s*)PyArray_GETPTR2(%(z)s, n, c);
```
bouthilx: This line seams useless

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+                for (int m = 0; m < nb_outcomes; ++m)
+                {
+                    dtype_%(z)s* z_nc = (dtype_%(z)s*)PyArray_GETPTR2(%(z)s, n, c);
```
bouthilx: Could be moved inside the if-condition?

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
...
+                    dtype_%(pvals)s* pvals_nm = (dtype_%(pvals)s*)PyArray_GETPTR2(pvals_copy, n, m);
+                    cummul += *pvals_nm;
+                    if (cummul > *unis_n)
+                    {
+                        *z_nc = m;
+                        *pvals_nm = 0.;
+                        // renormalize the nth row of pvals
+                        dtype_%(pvals)s sum = 0.;
+                        dtype_%(pvals)s* pvals_n = (dtype_%(pvals)s*)PyArray_GETPTR2(pvals_copy, n, 0);
+                        for (int k = 0; k < nb_outcomes; ++k)
```
bouthilx: This might not be much more efficient but you could avoid summing over pvals again. You already accumulated `cummul` up to `m` so:

```c
dtype_%(pvals)s sum = cummul - pvals_nm;
if ( m + 1 < nb_outcomes ) {
    dtype_%(pvals)s* pvals_n = (dtype_%(pvals)s*)PyArray_GETPTR2(pvals_copy, n, m + 1);
    for (int k = m + 1; k < nb_outcomes; ++k)
    {
        sum = sum + *pvals_n;
        pvals_n++;
    }
}
```

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,146 @@ def make_node(self, pvals, unis, n=1):
         out = T.tensor(dtype=odtype, broadcastable=pvals.type.broadcastable)
         return Apply(self, [pvals, unis, as_scalar(n)], [out])
 
+    def c_code_cache_version(self):
+        return None
```
bouthilx: Should be set to 1 on the last commit.

theano/sandbox/multinomial.py
```diff
@@ -79,12 +79,12 @@ def c_code(self, node, name, ins, outs, sub):
         return """
         if (PyArray_NDIM(%(pvals)s) != 2)
         {
-            PyErr_Format(PyExc_TypeError, "pvals wrong rank");
+            PyErr_Format(PyExc_TypeError, "pvals ndim should be 2");
```
bouthilx: ndim of what? ^_^

theano/sandbox/multinomial.py
```diff
@@ -215,6 +215,131 @@ def make_node(self, pvals, unis, n=1):
...
+                for (int m = 0; m < nb_outcomes; ++m)
+                {
+                    dtype_%(pvals)s* pvals_nm = (dtype_%(pvals)s*)PyArray_GETPTR2(pvals_copy, n, m);
+                    cummul += *pvals_nm;
+                    if (cummul > *unis_n)
+                    {
+                        *z_nc = m;
+                        // renormalize the nth row of pvals, reuse (cummul-*pvals_nm) to initialize the sum
+                        dtype_%(pvals)s sum = cummul - *pvals_nm;
+                        dtype_%(pvals)s* pvals_n = (dtype_%(pvals)s*)PyArray_GETPTR2(pvals_copy, n, m+1);
```
bouthilx: I'm afraid m+1 might get out of bound, am I wrong?

theano/sandbox/multinomial.py
aalmah: of pvlas?

theano/sandbox/multinomial.py
aalmah: I thought about that, and decided to change it but forgot. I will do it now.

theano/sandbox/multinomial.py
bouthilx: Oh my! :sweat: Sorry I didn't realized the dark green region was not the entire string. My bad!

bouthilx:
@nouiz Does travis start from scratch every time, so that the code is compiled even if `c_code_cache_version` doesn't change from a run to another?

nouiz:
The compiledir is always empty in travis. So yes, it get recompiled.

bouthilx:
@aalmah I'm confused. There was no out of bound and every test completed smoothly... And last commit didn't trigger any test. Did you change commit order with a rebase before pushing? Maybe github got confused with the order.

aalmah:
I am not sure what you mean, but I didn't add a test for out of bound case. I just changed the code to be safe if this happens.

bouthilx:
Yes I know. The things is, if I was right, then some tests would have probably failed unless they ran on the last version you submitted. That's why I'm wondering if the commit order is messed up on github. Otherwise, I was probably wrong. I would like to understand if commit [#14241a7](https://github.com/aalmah/Theano/commit/14241a7d865a2a1b5a49f26b0b09278e8741d957) really is necessary. The tests seams alright but I want to be sure. I'll test it myself next monday morning.

lamblin:
With rebased / amended commits, it happens that GitHub displays them in the order of modification, not of dependencies.
Indeed, 2c77122 lists 14241a7 as its parent, even though that parent appears last in the list.

bouthilx:
Thanks Pascal for the info! I'd like to run f408152 locally before merging just to be sure.

bouthilx:
OK it turns out the out of bound values were always initialized to 0 thus not affecting the normalization. This is why the tests would not fail even with the out of bound bug.

Issue #3928, opened 2016-01-25
Bilinear interpolation for upsampling images by an int factor


lamblin:
Upsampling images by an int factor (doubling, tripling, ...) can be done by computing `conv2d_grad_wrt_inputs` with a specially-crafted kernel (and taking care of the borders).
With `conv2d_grad_wrt_...` is part of `theano.nnet`, we could add the wrapper to make it available easily.

For large upsampling factors, it is also faster to decouple horizontal and vertical upsampling, and apply the operation twice, with a row kernel and a column kernel.

Sentient07:
@lamblin can i take this ?

lamblin:
Sorry, @SinaHonari is already assigned to this, since he's been working on that already.

fvisin:
#3951 closes this ticket

Sentient07:
I am trying to use this as an alternative to Transpose Convolution and I want to ideally decrease the number of filters/kernels used by this compared to the previous layer. The current implementation seems to retain the input channel dimension as the output channel dimension. Is this the desired behavior ?

nouiz:
What you want is not interpolation of the image, but a convolution I think.

Le mar. 2 mai 2017 13:54, Ramana Subramanyam <notifications@github.com> a
écrit :

Sentient07:
This is using transpose convolutional operation to scale up the images. So, if I rather take multiple kernels(which is the desired number of output filters), perform a normal transpose convolution operation and then apply weighted average of nearest cells, isn't that also the bilinear interpolated convolution operation? I'm not sure if it is correct, but this is the intuitive idea that I have from the distil.pub article

Sentient07:
@vdumoulin as you're an author of that post, I'd also like to have your input :)

PR #3951, opened 2016-01-29
Bilinear interpolation

SinaHonari:
First commit for issue #3928, bilinear interpolation.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -2,6 +2,7 @@
 Abstract conv interface
 """
 
+import numpy as np
 import logging
 from six import reraise
 import sys
@@ -125,6 +126,8 @@ def conv2d(input,
     Refer to :func:`nnet.conv2d <theano.tensor.nnet.conv2d>` for a more detailed documentation.
     """
 
+    input = as_tensor_variable(input)
+    filters = as_tensor_variable(filters)
     conv_op = AbstractConv2d(imshp=input_shape,
                              kshp=filter_shape,
                              border_mode=border_mode,
@@ -135,17 +138,18 @@ def conv2d(input,
 
 def conv2d_grad_wrt_inputs(output_grad,
                            filters,
-                           output_grad_shape=None,
-                           input_shape=None,
+                           input_shape,
                            filter_shape=None,
                            border_mode='valid',
                            subsample=(1, 1),
                            filter_flip=True):
-    """This function builds the symbolic graph for getting the
+    """Compute conv output gradient w.r.t its inputs
+
+    This function builds the symbolic graph for getting the
     gradient of the output of a convolution (namely output_grad)
     w.r.t the input of the convolution, given a set of 2D filters
     used by the convolution, such that the output_grad is upsampled
-    to the input shape.
+    to the input_shape.
 
     Parameters
     ----------
@@ -154,27 +158,24 @@ def conv2d_grad_wrt_inputs(output_grad,
         channels, input rows, input columns).  This is the tensor that
         will be upsampled or the output gradient of the convolution
         whose gradient will be taken with respect to the input of the
-        convolution.  See the optional parameter
-        ``output_grad_shape``.
+        convolution.
     filters : symbolic 4D tensor
         set of filters used in CNN layer of shape (output channels,
         input channels, filter rows, filter columns).  See the
         optional parameter ``filter_shape``.
-    output_grad_shape : list of 4 symbolic or real ints
-        The shape of the output_grad parameter.  Optional, possibly
-        used to choose an optimal implementation.  You can give
-        ``None`` for any element of the list to specify that this
-        element is not known at compile time.
-    input_shape : list of 2 symbolic or real ints
-        The shape (row and column size) of the input (upsampled)
-        parameter.  Not Optional, since given the output_grad_shape
+    input_shape : [None/int/Constant] * 2 + [Tensor/int/Constant] * 2
+        The shape of the input (upsampled) parameter.
+        A tuple/list of len 4, with the first two dimensions
+        being None or int or Constant and the last two dimensions being
+        Tensor or int or Constant.
+        Not Optional, since given the output_grad shape
         and the subsample values, multiple input_shape may be
         plausible.
-    filter_shape : list of 4 symbolic or real ints
-        The shape of the filters parameter.  Optional, possibly used
-        to choose an optimal implementation.  You can give ``None``
-        for any element of the list to specify that this element is
-        not known at compile time.
+    filter_shape : None or [None/int/Constant] * 4
+        The shape of the filters parameter. None or a tuple/list of len 4.
+        Optional, possibly used  to choose an optimal implementation.
+        You can give ``None`` for any element of the list to specify that
+        this element is not known at compile time.
     border_mode : str, int or tuple of two int
         Either of the following:
 
@@ -192,7 +193,7 @@ def conv2d_grad_wrt_inputs(output_grad,
             rows and ``filter columns // 2`` columns, then perform a
             valid convolution. For filters with an odd number of rows
             and columns, this leads to the output shape being equal to
-            the input shape.
+            the input shape. It is known as 'same' elsewhere.
 
           ``int``
             pad input with a symmetric border of zeros of the given
@@ -231,24 +232,50 @@ def conv2d_grad_wrt_inputs(output_grad,
 
     """
 
-    grad_input_op = AbstractConv2d_gradInputs(imshp=input_shape,
+    filters = as_tensor_variable(filters)
+    output_grad = as_tensor_variable(output_grad)
+
+    # checking the type of input_shape
+    for dim in [0, 1]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorConstant,
+                                             int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorVariable,
+                                             theano.tensor.TensorConstant,
+                                             int))
+
+    # checking the type of filter_shape
+    if filter_shape is not None:
+        for dim in [0, 1, 2, 3]:
+            assert isinstance(filter_shape[dim], (theano.tensor.TensorConstant,
+                                                  int, type(None)))
+
+    # setting the last two dimensions of input_shape to None, if
+    # the type of these dimensions is TensorVariable.
+    numerical_input_shape = list(input_shape)
+    for dim in [2, 3]:
+        if isinstance(input_shape[dim], theano.tensor.TensorVariable):
+            numerical_input_shape[dim] = None
+
+    grad_input_op = AbstractConv2d_gradInputs(imshp=numerical_input_shape,
                                               kshp=filter_shape,
                                               border_mode=border_mode,
                                               subsample=subsample,
                                               filter_flip=filter_flip)
 
-    return grad_input_op(filters, output_grad, input_shape)
+    return grad_input_op(filters, output_grad, input_shape[-2:])
 
 
 def conv2d_grad_wrt_weights(input,
                             output_grad,
+                            filter_shape,
                             input_shape=None,
-                            output_grad_shape=None,
-                            filter_shape=None,
                             border_mode='valid',
                             subsample=(1, 1),
                             filter_flip=True):
-    """This function will build the symbolic graph for getting the
+    """Compute conv output gradient w.r.t its weights
+
+    This function will build the symbolic graph for getting the
     gradient of the output of a convolution (output_grad) w.r.t its wights.
 
     Parameters
@@ -261,25 +288,17 @@ def conv2d_grad_wrt_weights(input,
         mini-batch of feature map stacks, of shape (batch size, input
         channels, input rows, input columns).  This is the gradient of
         the output of convolution.
-    filters : symbolic 4D tensor.
-        set of filters used in CNN layer of shape (output channels,
-        input channels, filter rows, filter columns).  See the
-        optional parameter ``filter_shape``.
-    output_grad_shape : list of 4 ints or Constant variables
-        The shape of the input parameter.  Optional, possibly used to
-        choose an optimal implementation.  You can give ``None`` for
-        any element of the list to specify that this element is not
-        known at compile time.
-    input_shape : list of 2 ints or Constant variables
-        The shape of the input parameter.  This parameter indicates
-        the row and column size of the input in the forward pass.
+    filter_shape : [None/int/Constant] * 2 + [Tensor/int/Constant] * 2
+        The shape of the filter parameter.  A tuple/list of len 4, with the
+        first two dimensions being None or int or Constant and the last two
+        dimensions being Tensor or int or Constant.
+        Not Optional, since given the output_grad shape and
+        the input_shape, multiple filter_shape may be plausible.
+    input_shape : None or [None/int/Constant] * 4
+        The shape of the input parameter. None or a tuple/list of len 4.
         Optional, possibly used to choose an optimal implementation.
         You can give ``None`` for any element of the list to specify
         that this element is not known at compile time.
-    filter_shape : list of 4 ints or Constant variables
-        The shape of the filters parameter.  Not Optional, since given
-        the output_grad_shape and the input_shape, multiple
-        filter_shape may be plausible.
     border_mode : str, int or tuple of two ints
         Either of the following:
 
@@ -297,7 +316,7 @@ def conv2d_grad_wrt_weights(input,
             rows and ``filter columns // 2`` columns, then perform a
             valid convolution. For filters with an odd number of rows
             and columns, this leads to the output shape being equal to
-            the input shape.
+            the input shape. It is known as 'same' elsewhere.
 
           ``int``
             pad input with a symmetric border of zeros of the given
@@ -335,13 +354,217 @@ def conv2d_grad_wrt_weights(input,
         version until it is released.
 
     """
+
+    input = as_tensor_variable(input)
+    output_grad = as_tensor_variable(output_grad)
+
+    # checking the type of filter_shape
+    for dim in [0, 1]:
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorConstant,
+                                              int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorVariable,
+                                              theano.tensor.TensorConstant,
+                                              int))
+
+    # checking the type of input_shape
+    if input_shape is not None:
+        for dim in [0, 1, 2, 3]:
+            assert isinstance(input_shape[dim], (theano.tensor.TensorConstant,
+                                                 int, type(None)))
+
+    # setting the last two dimensions of filter_shape to None, if
+    # the type of these dimensions is TensorVariable.
+    numerical_filter_shape = list(filter_shape)
+    for dim in [2, 3]:
+        if isinstance(filter_shape[dim], theano.tensor.TensorVariable):
+            numerical_filter_shape[dim] = None
+
     gradWeight_op = AbstractConv2d_gradWeights(imshp=input_shape,
-                                               kshp=filter_shape,
+                                               kshp=numerical_filter_shape,
                                                border_mode=border_mode,
                                                subsample=subsample,
                                                filter_flip=filter_flip)
 
-    return gradWeight_op(input, output_grad, filter_shape)
+    return gradWeight_op(input, output_grad, filter_shape[:-2])
+
+
+def bilinear_kernel_2D(ratio, normalize=True):
+    """Compute 2D kernel for bilinear upsampling
+
+    This function builds the 2D kernel that can be used to upsample
+    a tensor by the given ratio using bilinear interpolation.
+
+    Parameters
+    ----------
+    ratio: int or Constant/Scalar Theano tensor of int* dtype
+        the ratio by which an image will be upsampled by the returned filter
+        in the 2D space.
+
+    normalize: bool
+        param normalize: indicates whether to normalize the kernel or not.
+        Default is True.
+
+    Returns
+    -------
+    symbolic 2D tensor
+        the 2D kernels that can be applied to any given image to upsample it
+        by the indicated ratio using bilinear interpolation in two dimensions.
+
+    """
+
+    hkern = bilinear_kernel_1D(ratio=ratio, normalize=normalize).dimshuffle('x', 0)
+    vkern = bilinear_kernel_1D(ratio=ratio, normalize=normalize).dimshuffle(0, 'x')
+    kern = hkern * vkern
+    return kern
+
+
+def bilinear_kernel_1D(ratio, normalize=True):
+    """Compute 1D kernel for bilinear upsampling
+
+    This function builds the 1D kernel that can be used to upsample
+    a tensor by the given ratio using bilinear interpolation.
+
+    Parameters
+    ----------
+    ratio: int or Constant/Scalar Theano tensor of int* dtype
+        the ratio by which an image will be upsampled by the returned filter
+        in the 2D space.
+
+    normalize: bool
+        param normalize: indicates whether to normalize the kernel or not.
+        Default is True.
+
+    Returns
+    -------
+    symbolic 1D tensor
+        the 1D kernels that can be applied to any given image to upsample it
+        by the indicated ratio using bilinear interpolation in one dimension.
+
+    """
+
+    T = theano.tensor
+    half_kern = T.arange(1, ratio + 1, dtype=theano.config.floatX)
+    kern = T.concatenate([half_kern, half_kern[-2::-1]])
+
+    if normalize:
+        kern /= ratio
+    return kern
+
+
+def bilinear_upsampling(input,
+                        ratio,
+                        batch_size=None,
+                        num_input_channels=None,
+                        use_1D_kernel=True):
+    """Compute bilinear upsampling
+
+    This function will build the symbolic graph for upsampling
+    a tensor by the given ratio using bilinear interpolation.
+
+    Parameters
+    ----------
+    input: symbolic 4D tensor
+        mini-batch of feature map stacks, of shape (batch size,
+        input channels, input rows, input columns) that will be upsampled.
+
+    ratio: int or Constant or Scalar Tensor of int* dtype
+        the ratio by which the input is upsampled in the 2D space (row and
+        col size).
+
+    batch_size: None, int or Constant variable
+        The size of the first dimension of the input variable.
+        Optional, possibly used to choose an optimal implementation.
+        batch_size will be used only if num_input_channels is not None.
+
+    num_input_channels: None, int or Constant variable
+        The size of the second dimension of the input variable.
+        Optional, possibly used to choose an optimal implementation.
+        num_input_channels will be used only if batch_size is not None.
+
+    use_1D_kernel: bool
+        if set to true, row and column will be upsampled seperately by 1D
+        kernels, otherwise they are upsampled together using a 2D kernel. The
+        final result is the same, only the speed can differ, given factors such
+        as upsampling ratio.
+
+    Returns
+    -------
+    symbolic 4D tensor
+        set of feature maps generated by bilinear upsampling. Tensor
+        is of shape (batch size, num_input_channels, input row size * ratio,
+        input column size * ratio)
+
+    Notes
+    -----
+
+    :note: The kernel used for bilinear interpolation is fixed (not learned).
+
+    :note: When the upsampling ratio is even, the last row and column is
+        repeated one extra time compared to the first row and column which makes
+        the upsampled tensor asymmetrical on both sides. This does not happen when
+        the upsampling ratio is odd.
+
+    """
+
+    T = theano.tensor
+    try:
+        up_bs = batch_size * num_input_channels
+    except TypeError:
+        up_bs = None
+    row, col = input.shape[2:]
+    up_input = input.reshape((-1, 1, row, col))
+
+    # concatenating the first and last row and column
+    # first and last row
+    concat_mat = T.concatenate((up_input[:, :, :1, :], up_input,
+                                up_input[:, :, -1:, :]), axis=2)
+    # first and last col
+    concat_mat = T.concatenate((concat_mat[:, :, :, :1], concat_mat,
+                                concat_mat[:, :, :, -1:]), axis=3)
+
+    pad = 2 * ratio - (ratio - 1) // 2 - 1
+
+    if use_1D_kernel:
+        kern = bilinear_kernel_1D(ratio=ratio, normalize=True)
+        # upsampling rows
+        upsampled_row = conv2d_grad_wrt_inputs(output_grad=concat_mat,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis, :,
+                                                            np.newaxis],
+                                               input_shape=(up_bs, 1,
+                                                            row * ratio, col),
+                                               filter_shape=(1, 1, None, 1),
+                                               border_mode=(pad, 0),
+                                               subsample=(ratio, 1),
+                                               filter_flip=True)
+        # upsampling cols
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=upsampled_row,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis,
+                                                            np.newaxis, :],
+                                               input_shape=(up_bs, 1,
+                                                            row * ratio,
+                                                            col * ratio),
+                                               filter_shape=(1, 1, 1, None),
+                                               border_mode=(0, pad),
+                                               subsample=(1, ratio),
+                                               filter_flip=True)
+    else:
+        kern = bilinear_kernel_2D(ratio=ratio, normalize=True)
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=concat_mat,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis, :, :],
+                                               input_shape=(up_bs, 1,
+                                                            row * ratio,
+                                                            col * ratio),
+                                               filter_shape=(1, 1, None, None),
+                                               border_mode=(pad, pad),
+                                               subsample=(ratio, ratio),
+                                               filter_flip=True)
+
+    return upsampled_mat.reshape((batch_size, num_input_channels,
+                                  row * ratio, col * ratio))
 
 
 class BaseAbstractConv2d(Op):
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -1,4 +1,5 @@
 import numpy
+import numpy as np
 import unittest
 
 from nose.plugins.skip import SkipTest
@@ -12,6 +13,9 @@ from theano.tensor.nnet.abstract_conv import get_conv_output_shape
 from theano.tensor.nnet.abstract_conv import AbstractConv2d
 from theano.tensor.nnet.abstract_conv import AbstractConv2d_gradInputs
 from theano.tensor.nnet.abstract_conv import AbstractConv2d_gradWeights
+from theano.tensor.nnet.abstract_conv import bilinear_kernel_1D
+from theano.tensor.nnet.abstract_conv import bilinear_kernel_2D
+from theano.tensor.nnet.abstract_conv import bilinear_upsampling
 from theano.tensor.nnet.conv import ConvOp
 from theano.tensor.nnet.corr import (CorrMM, CorrMM_gradWeights,
                                      CorrMM_gradInputs)
@@ -529,3 +533,180 @@ class TestConvTypes(unittest.TestCase):
         grad_filters = theano.grad(grad_input.sum(), wrt=filters)
         assert grad_filters.type == filters.type, (
             grad_filters, grad_filters.type, filters, filters.type)
+
+
+class TestBilinearUpsampling(unittest.TestCase):
+
+    def numerical_kernel_1D(self, ratio):
+        """Gets numerical 1D kernel for bilinear upsampling"""
+        return np.array(list(range(1, ratio + 1)) +
+                        list(range(ratio - 1, 0, -1)))
+
+    def numerical_kernel_2D(self, ratio):
+        """Gets numerical 2D kernel for bilinear upsampling"""
+        return np.array([i * j for i in self.numerical_kernel_1D(ratio) for j
+                         in self.numerical_kernel_1D(ratio)]).\
+            reshape(2 * ratio - 1, 2 * ratio - 1)
+
+    def test_bilinear_kernel_2D(self):
+        """Test 2D kernels used in bilinear upsampling
+
+        This method tests the correctness of the
+        2D kernel values used in bilinear upsampling
+        for some upsampling ratios.
+
+        """
+        for ratio in [2, 3, 4, 5, 6, 7, 8, 9]:
+            # getting the un-normalized kernel
+            kernel = bilinear_kernel_2D(ratio=ratio, normalize=False)
+            f = theano.function([], kernel)
+            kernel_2D = self.numerical_kernel_2D(ratio)
+            np.testing.assert_allclose(kernel_2D, f())
+
+            # getting the normalized kernel
+            kernel = bilinear_kernel_2D(ratio=ratio, normalize=True)
+            f = theano.function([], kernel)
+            kernel_2D = kernel_2D / float(ratio**2)
+            np.testing.assert_allclose(kernel_2D, f())
+
+    def test_bilinear_kernel_1D(self):
+        """Test 1D kernels used in bilinear upsampling
+
+        This method tests the correctness of the
+        1D kernel values used in bilinear upsampling
+        for some upsampling ratios.
+
+        """
+        rat = tensor.iscalar()
+        kernel_ten = bilinear_kernel_1D(ratio=rat, normalize=False)
+        f_ten = theano.function([rat], kernel_ten)
+
+        kernel_ten_norm = bilinear_kernel_1D(ratio=rat, normalize=True)
+        f_ten_norm = theano.function([rat], kernel_ten_norm)
+
+        for ratio in [2, 3, 4, 5, 6, 7, 8, 9]:
+            # getting the un-normalized kernel
+            kernel = bilinear_kernel_1D(ratio=ratio, normalize=False)
+            f = theano.function([], kernel)
+            kernel_1D = self.numerical_kernel_1D(ratio)
+            np.testing.assert_allclose(kernel_1D, f())
+            np.testing.assert_allclose(kernel_1D, f_ten(ratio))
+
+            # getting the normalized kernel
+            kernel = bilinear_kernel_1D(ratio=ratio, normalize=True)
+            f = theano.function([], kernel)
+            kernel_1D = kernel_1D / float(ratio)
+            np.testing.assert_allclose(kernel_1D, f())
+            np.testing.assert_allclose(kernel_1D, f_ten_norm(ratio))
+
+    def numerical_upsampling_multiplier(self, ratio):
+        """Compute upsampling multiplier
+
+        This method computes the multipliers of an array
+        that will be upsampled using bilinear interpolation.
+
+        Parameters
+        ----------
+        ratio: int
+            the ratio by which the array will be upsampled.
+
+        Returns
+        -------
+        1D numpy array
+            The multiplers that can be used in bilinear interpolation
+            to upsample an array.
+
+        int
+            The size of the multipliers array
+
+        """
+        kern = np.arange(ratio + 1)
+        return kern, kern.shape[0]
+
+    def get_upsampled_twobytwo_mat(self, two_by_two, ratio):
+        """Upsample 4D array with two rows and two columns
+
+        This method gets a 4D numpy array with two rows and two columns
+        and computes its upsampled array by using bilinear interpolation
+
+        Parameters
+        ----------
+        two_by_two: numpy 4D array
+            The array that will be upsampled by bilinear interpolation.
+            Array is of shape (batch size, num channels, 2, 2)
+
+        ratio: int
+            The ratio by which two_by_two's last
+            two dimensions (row and col) will be upsampled.
+
+        Returns
+        -------
+        4D numpy array
+            The array upsampled by using bilinear interpolation. Array
+            is of shape (batch size, num channels, 2*ratio, 2*ratio).
+
+        """
+        kern, shp = self.numerical_upsampling_multiplier(ratio)
+        up_1D = two_by_two[:, :, :, :1] * kern[::-1] + \
+            two_by_two[:, :, :, 1:] * kern
+        up_2D = up_1D[:, :, :1, :] * kern[::-1][:, np.newaxis] + \
+            up_1D[:, :, 1:, :] * kern[:, np.newaxis]
+        num_concat = (ratio - 1) // 2
+        for i in range(num_concat):
+            up_2D = np.concatenate([up_2D[:, :, :1, :], up_2D], axis=2)
+            up_2D = np.concatenate([up_2D, up_2D[:, :, -1:, :]], axis=2)
+            up_2D = np.concatenate([up_2D[:, :, :, :1], up_2D], axis=3)
+            up_2D = np.concatenate([up_2D, up_2D[:, :, :, -1:]], axis=3)
+        if ratio % 2 == 0:
+            up_2D = np.concatenate([up_2D, up_2D[:, :, -1:, :]], axis=2)
+            up_2D = np.concatenate([up_2D, up_2D[:, :, :, -1:]], axis=3)
+        return up_2D / float(ratio)**2
+
+    def test_bilinear_upsampling_1D(self):
+        """Test bilinear upsampling using 1D kernels
+
+        This method tests the bilinear_upsampling method
+        when using 1D kernels for some upsampling ratios.
+
+        """
+        # upsampling for a ratio of two
+        input_x = np.array([[[[1, 2], [3, 4]]]], dtype=theano.config.floatX)
+
+        for ratio in [2, 3, 4, 5, 6, 7, 8, 9]:
+            bilin_mat = bilinear_upsampling(input=input_x, ratio=ratio,
+                                            batch_size=1, num_input_channels=1,
+                                            use_1D_kernel=True)
+            f = theano.function([], bilin_mat)
+            up_mat_2d = self.get_upsampled_twobytwo_mat(input_x, ratio)
+            np.testing.assert_allclose(f(), up_mat_2d, rtol=1e-06)
+
+    def test_compare_1D_and_2D_upsampling_values(self):
+        """Compare 1D and 2D upsampling
+
+        This method verifies the bilinear upsampling done by using
+        1D and 2D kernels will generate the same result.
+
+        """
+        # checking upsampling with ratio 5
+        input_x = np.random.rand(5, 4, 6, 7).astype(theano.config.floatX)
+        mat_1D = bilinear_upsampling(input=input_x, ratio=5,
+                                     batch_size=5, num_input_channels=4,
+                                     use_1D_kernel=True)
+        mat_2D = bilinear_upsampling(input=input_x, ratio=5,
+                                     batch_size=5, num_input_channels=4,
+                                     use_1D_kernel=False)
+        f_1D = theano.function([], mat_1D)
+        f_2D = theano.function([], mat_2D)
+        np.testing.assert_allclose(f_1D(), f_2D(), rtol=1e-06)
+
+        # checking upsampling with ratio 8
+        input_x = np.random.rand(12, 11, 10, 7).astype(theano.config.floatX)
+        mat_1D = bilinear_upsampling(input=input_x, ratio=8,
+                                     batch_size=12, num_input_channels=11,
+                                     use_1D_kernel=True)
+        mat_2D = bilinear_upsampling(input=input_x, ratio=8,
+                                     batch_size=12, num_input_channels=11,
+                                     use_1D_kernel=False)
+        f_1D = theano.function([], mat_1D)
+        f_2D = theano.function([], mat_2D)
+        np.testing.assert_allclose(f_1D(), f_2D(), rtol=1e-06)


fvisin:
Please let me know when you are done and I can check.

nouiz:
travis is failing. I didn't check the code, but I updated the PR title. Use a title that tell what it is in it, not an issue number.

SinaHonari:
There is a link that connects it to the correct issue (which has the explanation). I have also mentioned bilinear interpolation in the above comment.

SinaHonari:
@fvisin: I'm already done. Travis has issues with importing some of the libraries, which I don't know why it's happening. Please review the code and let me know of any issues (either why the importing is problematic or comments on the implementation of the code itself). Thanks.

fvisin:
Great, thank you @SinaHonari. I will investigate why the import is failing in Travis and check your implementation. 

For the next time, I support Fred's request: it is desirable to have an informative title for the PR so that it is not necessary to open it to know what it refers to.

andreh7:
I tried import theano with this pull request (in a environment where I don't have another theano installed).

Putting printouts at the beginning and end of theano/__init__.py and theano/tensor/nnet/abstract_conv.py one can see that by the time theano/tensor/nnet/abstract_conv.py imports theano.tensor, the 'global' import of theano hasn't completed yet (i.e. this looks like a circular dependency to me).

moving the 'import theano.tensor as T' into the function where it is used (inside the function bilinear_upsampling(..) ) and therefore delaying the import solves the problem for me.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                               subsample=subsample,
+                                               filter_flip=filter_flip)
+
+    return gradWeight_op(input, output_grad, input_shape)
+
+
+def bilinear_kernel_2D(ratio, normalize=True):
+    """ This function gets a ratio
+    and returns the normalized 2D filters that can
+    be applied to any image to upsample it by the ratio
```
fvisin: by *that* ratio

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+    using bilinear interpolation.
+
+    :type ratio: integer
+    : param ratio: the ratio by which an image is upsampled
```
fvisin: I'd say `the ratio by which an image will be upsampled by the returned filters"

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+        in the 2D space.
+
+    :type normalize: bool
+    : param normalize: indicates whether to normalize the
+    kernel or not. Default is True.
+
+    returns: the 2D kernels that can be applied to
+             any given image to upsample it by indicated ratio
```
fvisin: by *the* indicated ratio

fvisin:
Thank you @andreh7 for looking into this. I confirm there is a circular import: 
`theano.tensor` imports `theano.tensor.nnet`, which in turn imports `abstract_conv.py`, which imports `theano.tensor` again (before it is completely initialized).

``` python
----> 1 import theano

/home/francesco/exp/theano/theano/theano/__init__.pyc in <module>()
     74 from theano.printing import pprint, pp
     75 
---> 76 from theano.scan_module import scan, map, reduce, foldl, foldr, clone
     77 
     78 from theano.updates import OrderedUpdates

/home/francesco/exp/theano/theano/theano/scan_module/__init__.py in <module>()
     38 __contact__ = "Razvan Pascanu <r.pascanu@gmail>"
     39 
---> 40 from theano.scan_module import scan_opt
     41 from theano.scan_module.scan import scan
     42 from theano.scan_module.scan_views import map, reduce, foldl, foldr

/home/francesco/exp/theano/theano/theano/scan_module/scan_opt.py in <module>()
     57 
     58 import theano
---> 59 from theano import tensor
     60 from theano.tensor import opt, get_scalar_constant_value, Alloc, AllocEmpty
     61 from theano import gof

/home/francesco/exp/theano/theano/theano/tensor/__init__.py in <module>()
     34 from theano.tensor.io import *
     35 
---> 36 from theano.tensor import nnet  # used for softmax, sigmoid, etc.
     37 
     38 from theano.gradient import Rop, Lop, grad, numeric_grad, verify_grad, \

/home/francesco/exp/theano/theano/theano/tensor/nnet/__init__.py in <module>()
     22     prepend_scalar_to_each_row, relu, softmax, softmax_grad, softmax_graph,
     23     softmax_op, softmax_simplifier, softmax_with_bias)
---> 24 from . import opt
     25 from .conv import ConvOp
     26 from .Conv3D import *

/home/francesco/exp/theano/theano/theano/tensor/nnet/opt.py in <module>()
      7 from theano.gof import local_optimizer
      8 
----> 9 from theano.tensor.nnet.corr import (
     10     CorrMM, CorrMM_gradInputs, CorrMM_gradWeights)
     11 from theano.tensor.nnet.blocksparse import (

/home/francesco/exp/theano/theano/theano/tensor/nnet/corr.py in <module>()
      6 from theano import gof
      7 from theano.tensor import as_tensor_variable, TensorType
----> 8 from theano.tensor.nnet.abstract_conv import get_conv_output_shape
      9 from theano.tensor.blas_headers import blas_header_text
     10 from theano.tensor.blas import ldflags

/home/francesco/exp/theano/theano/theano/tensor/nnet/abstract_conv.py in <module>()
      6 import logging
      7 import theano
----> 8 import theano.tensor as T
      9 
     10 from theano.tensor import as_tensor_variable, patternbroadcast

AttributeError: 'module' object has no attribute 'tensor'
```

Moving `import theano.tensor as T` inside `bilinear_upsampling(..)` solves the problem and seems a good solution to me, but I'd like to know the opinion of @lamblin on this.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+            kern[ii + ratio - 1, jj + ratio - 1] = (ratio - abs(ii)) * (ratio - abs(jj))
+    if normalize:
+        kern /= ratio**2
+    return kern
+
+
+def bilinear_kernel_1D(ratio, normalize=True):
+    """ This function gets a ratio
+    and returns the normalized 1D filters that can
+    be applied to any image to upsample it by the ratio
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+    :type ratio: integer
+    : param ratio: the ratio by which an image is upsampled
+        in the 1D space.
+
+    :type normalize: bool
+    : param normalize: indicates whether to normalize the
+    kernel or not. Default is True.
+
+    returns: the 1D kernels that can be applied to
+             any given image to upsample it by indicated ratio
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+
+
+def bilinear_upsampling(input,
+                        ratio,
+                        input_shape=None,
+                        use_1D_kernel=True,
+                        repeat_first_row=True,
+                        repeat_first_col=True):
+    """This function takes a stack of 2D inputs and upsamples it (in 2D space)
+    by the ratio. The subsample is assumed to be one and
```
fvisin: The end of the sentence is missing

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+        in the 2D space.
+
+    :type input_shape: None, tuple/list of len 4 of int or
+        Constant variable.
+    :param input_shape: The shape of the input variable.
+        Optional, possibly used to choose an optimal implementation.
+        You can give ``None`` for any element of the list to specify that this
+        element is not known at compile time.
+
+    :type use_1D_kernel: bool
```
fvisin: I don't think we want to let the user choose which strategy to apply. The optimization should go on under the hood without the user being aware of it. When the filter is big doing the two upsampling separately should be faster. 

Please benchmark the two implementations with a combination of values for image size, kernel size and upsampling ratio to see if it's possible to find an heuristic to determine when to apply one or the other strategy.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+             in the 2D space, each feature is upsampled by the given ratio.
+    """
+
+    bch, ch, row, col = input.shape
+    up_input = input.reshape(-1, 1, row, col)
+
+    # indicating whether the output dimension
+    # is one dimension less (in the case of even ratio)
+    # or not (in the case of odd ratio)
+    if ratio // 2 == 0:
```
fvisin: I don't think this does what you want. This should work: `sub = 0 if ratio % 2 else 1`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+    kern_size = ratio * 2 - 1
+    kern = np.zeros((kern_size))
+    for ii in range(-ratio + 1, ratio):
+        kern[ii + ratio - 1] = ratio - abs(ii)
+    if normalize:
+        kern /= ratio
+    return kern
+
+
+def bilinear_upsampling(input,
```
fvisin: Please use a different name, as this will redefine the standard python function [`input`](https://docs.python.org/3/library/functions.html#input).

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+        this variable to true copies the first column (making it appear twice
+        in the final upsampled tensor), on the other hand a false value
+        copies the last column.
+
+    returns: the upsampled 4D tensor, upsampled using bilinear
+             interpolation with fixed (not learned) kernels such that
+             in the 2D space, each feature is upsampled by the given ratio.
+    """
+
+    bch, ch, row, col = input.shape
```
fvisin: `bch` is an uncommon short name for "batch size". `bs` might be more clear.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+        sub = 1
+    else:
+        sub = 0
+
+    if use_1D_kernel:
+        kern = bilinear_kernel_1D(ratio=ratio, normalize=True)
+        kern_size = kern.shape[0]
+        # upsampling rows
+        upsampled_row = conv2d_grad_wrt_inputs(output_grad=up_input,
+                                               filters=kern.reshape(1, 1,
```
fvisin: Please use `kern[np.newaxis, np.newaxis, :, np.newaxis]`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                                            row * ratio - sub,
+                                                            col),
+                                               filters_shape=(1, 1, kern_size,
+                                                              1),
+                                               border_mode='valid',
+                                               subsample=(1, 1),
+                                               filter_flip=True)
+        # upsampling cols
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=upsampled_row,
+                                               filters=kern.reshape(1, 1, 1,
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                               filters_shape=(1, 1, 1,
+                                                              kern_size),
+                                               border_mode='valid',
+                                               subsample=(1, 1),
+                                               filter_flip=True)
+    else:
+        kern = bilinear_kernel_2D(ratio=ratio, normalize=True)
+        kern_row, kern_col = kern.shape
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=up_input,
+                                               filters=kern.reshape(1, 1,
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                                              kern_col),
+                                               border_mode='valid',
+                                               subsample=(1, 1),
+                                               filter_flip=True)
+
+    if sub:
+        # ratio is even
+        up_bch, up_ch, up_row, up_col = upsampled_mat.shape
+        if repeat_first_row:
+            upsampled_mat = T.concatenate((upsampled_mat[:, :, 0, :].reshape((
```
fvisin: ditto

fvisin: Please use the much cleaner `T.concatenate((upsampled_mat[:, :, 0:1, :], upsampled_mat), axis=2)`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                           up_bch, up_ch, 1, up_col)),
+                                           upsampled_mat), axis=2)
+        else:
+            # repeating last row
+            upsampled_mat = T.concatenate((upsampled_mat,
+                                           upsampled_mat[:, :, -1, :].reshape((
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                               up_bch, up_ch, 1, up_col))),
+                                          axis=2)
+        if repeat_first_col:
+            upsampled_mat = T.concatenate((upsampled_mat[:, :, :, 0].reshape((
+                                           up_bch, up_ch, up_row + 1, 1)),
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                           upsampled_mat), axis=3)
+        else:
+            # repeating last col
+            upsampled_mat = T.concatenate((upsampled_mat,
+                                           upsampled_mat[:, :, :, -1].reshape((
+                                               up_bch, up_ch, up_row + 1, 1))),
```
fvisin: ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+        in the final upsampled tensor), on the other hand a false value
+        copies the last column.
+
+    returns: the upsampled 4D tensor, upsampled using bilinear
+             interpolation with fixed (not learned) kernels such that
+             in the 2D space, each feature is upsampled by the given ratio.
+    """
+
+    bch, ch, row, col = input.shape
+    up_input = input.reshape(-1, 1, row, col)
```
fvisin: Should be `up_input = input.reshape((-1, 1, row, col))`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+        # upsampling rows
+        upsampled_row = conv2d_grad_wrt_inputs(output_grad=up_input,
+                                               filters=kern.reshape(1, 1,
+                                                                    kern_size,
+                                                                    1),
+                                               output_grad_shape=up_input.shape,
+                                               input_shape=(bch * ch, 1,
+                                                            row * ratio - sub,
+                                                            col),
+                                               filters_shape=(1, 1, kern_size,
```
fvisin: should be `filter_shape`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -177,6 +179,383 @@ def conv2d(input,
...
+                                               filter_flip=True)
+        # upsampling cols
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=upsampled_row,
+                                               filters=kern.reshape(1, 1, 1,
+                                                                    kern_size),
+                                               output_grad_shape=upsampled_row.shape,
+                                               input_shape=(bch * ch, 1,
+                                                            row * ratio - sub,
+                                                            col * ratio - sub),
+                                               filters_shape=(1, 1, 1,
```
fvisin: ditto

fvisin:
@SinaHonari please fix the issues I pointed out and also please add a unit test to check the correct behaviour of all the functions.

theano/tensor/nnet/abstract_conv.py
lamblin: I think this can be done later, and is out of the scope of this ticket. Moreover, the heuristic could depend on the type or generation of GPU as well. For the moment, I think we should let the user benchmark and choose, then we can add a new value, for instance `'auto'` when we have a heuristic.

fvisin: Ok, fair enough.

theano/tensor/nnet/abstract_conv.py
SinaHonari: Thanks Francesco for the feedback. I will apply the changes. @lamblin: Is it fine to import theano.tensor inside bilinear_upsampling as proposed by Francesco?

fvisin:
I spoke with @lamblin, the best way to solve the import loop is to remove `import theano.tensor as T` from the import section and then define `T = theano.tensor` in the functions when you use it. This should fix the import loop without the need to import `theano.tensor` inside the function.

theano/tensor/nnet/abstract_conv.py
SinaHonari: Other methods use input. It would make them different if I change this one.

theano/tensor/nnet/abstract_conv.py
SinaHonari: What's the difference with using a reshape and providing the dimension?

lamblin: Using indexing with `newaxis` adds `DimShuffle` operations, which can always be done inplace, and which are easier to analyze during the optimization phase (compared to `Reshape`).

theano/tensor/nnet/abstract_conv.py
```diff
@@ -309,9 +314,6 @@ def conv2d_grad_wrt_weights(input,
     :param input_shape: The shape of the input parameter.
         This parameter indicates the row and column size of the input
         in the forward pass.
-        Optional, possibly used to choose an optimal implementation.
-        You can give ``None`` for any element of the list to specify that this
-        element is not known at compile time.
```
nouiz: Why did you remove this? Bad rebase?

SinaHonari: There was an error when rebasing, which made me do force push. I will fix it.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -366,6 +374,215 @@ def conv2d_grad_wrt_weights(input,
...
+    # last row
+    concat_mat = T.concatenate((concat_mat, concat_mat[:, :, -1:, :]), axis=2)
+    # first col
+    concat_mat = T.concatenate((concat_mat[:, :, :, 0:1], concat_mat), axis=3)
+    # last col
+    concat_mat = T.concatenate((concat_mat, concat_mat[:, :, :, -1:]), axis=3)
+
+    cc_bs, cc_ch, cc_row, cc_col = concat_mat.shape
+
+    pad = ratio // 2
```
lamblin: That is not the formula I used in the notebook. I think the correct one is "f - (R - 1) // 2", or, in other words, "pad = 2 * ratio - (ratio - 1) // 2 - 1". I think it can be simplified, but that seems to work.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -366,6 +374,215 @@ def conv2d_grad_wrt_weights(input,
...
+        # upsampling rows
+        upsampled_row = conv2d_grad_wrt_inputs(output_grad=concat_mat,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis,
+                                                            :, np.newaxis],
+                                               output_grad_shape=concat_mat.
+                                               shape,
+                                               input_shape=(cc_bs, cc_ch,
+                                                            cc_row * ratio,
+                                                            cc_col),
```
lamblin: Input shape should not use `cc_row` or `cc_col`, but `row` and `col`.
There seems to be an additional issue: we would like `row` and `col` to be symbolic variables, but `conv2d_grad_wrt_inputs` needs then to be constants.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -366,6 +374,215 @@ def conv2d_grad_wrt_weights(input,
...
+    if ratio % 2 == 0:
+        sub = 1
+    else:
+        sub = 0
+
+    # concatenating the first and last row and column
+    # first row
+    concat_mat = T.concatenate((up_input[:, :, 0:1, :], up_input), axis=2)
+    # last row
+    concat_mat = T.concatenate((concat_mat, concat_mat[:, :, -1:, :]), axis=2)
```
lamblin: Did you check that these two `concatenate` operations were merged into only one in the final graph?
If not, then we should have only one concatenation on axis 2, and one on axis 3, not 4 in total.

theano/tensor/nnet/abstract_conv.py
lamblin: The rationale is that `2 * ratio - 1` is equivalent to the gradient of a full convolution, but we want an additional `(ratio -1) // 2` pixels on each side.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -366,6 +374,215 @@ def conv2d_grad_wrt_weights(input,
...
+             using bilinear interpolation in one dimension.
+    """
+    assert ratio > 1
+    kern_size = ratio * 2 - 1
+    kern = np.zeros((kern_size), dtype=theano.config.floatX)
+    for ii in range(-ratio + 1, ratio):
+        kern[ii + ratio - 1] = ratio - abs(ii)
+    if normalize:
+        kern /= float(ratio)
+    return kern
```
lamblin: Actually, I think this would be quite easily done in pure Theano. I think the following would work:
```
half_kern = T.arange(1, ratio + 1, dtype=floatX)  # Actually, half plus the middle element
kern = T.concatenate(half_kern, half_kern[-2::-1])
```
And then you can normalize if required.

Then you can express the 2D kernel as the elemwise product of two 1D kernels, one as a row and the other as a column. This makes it possible to use different ratios for rows and columns as well.

fvisin: Out of curiosity, is it more convenient (or is there any difference) to go for pure Theano or standard list comprehensions, e.g., `[i for i in range(1, ratio+1) + range(ratio-1, 0, -1)]`?

theano/tensor/nnet/abstract_conv.py
fvisin: If I am not wrong it would also be more memory-efficient to allocate all the space at once at the beginning and then modify the corresponding subtensors.

theano/tensor/nnet/abstract_conv.py
fvisin: Equivalently the 2D kernel could be obtained with:
```python
np.array([i*j for i in bilinear_kernel_1D(ratio, False) for j in bilinear_kernel_1D(ratio, False)]).reshape(2*ratio-1, 2*ratio-1)
```

Which solution is preferable?

lamblin: The pure Theano makes it possible to have the upsampling ratio depend on the shape of some variables, and change at runtime.

theano/tensor/nnet/abstract_conv.py
SinaHonari: This is the graph for concat_mat after two concat operations, is it correct?

Join [@A] ''   
 |TensorConstant{2} [@B]
 |Join [@C] ''   
 | |TensorConstant{2} [@B]
 | |Subtensor{::, ::, int64:int64:, ::} [@D] ''   
 | | |Reshape{4} [@E] ''   
 | | | |<TensorType(float64, 4D)> [@F]
 | | | |MakeVector{dtype='int64'} [@G] ''   
 | | |   |Elemwise{Cast{int64}} [@H] ''   
 | | |   | |TensorConstant{-1} [@I]
 | | |   |Elemwise{Cast{int64}} [@J] ''   
 | | |   | |TensorConstant{1} [@K]
 | | |   |Subtensor{int64} [@L] ''   
 | | |   | |Shape [@M] ''   
 | | |   | | |<TensorType(float64, 4D)> [@F]
 | | |   | |Constant{2} [@N]
 | | |   |Subtensor{int64} [@O] ''   
 | | |     |Shape [@M] ''   
 | | |     |Constant{3} [@P]
 | | |Constant{0} [@Q]
 | | |Constant{1} [@R]
 | |Reshape{4} [@E] ''   
 |Subtensor{::, ::, int64::, ::} [@S] ''   
   |Join [@C] ''   
   |Constant{-1} [@T]

fvisin: I would have expected two `Join` operation to be enough. I'll ask @lamblin 's opinion on this.

@SinaHonari can you confirm this is the graph of:
```python
 concat_mat = T.concatenate((up_input[:, :, :1, :], up_input, concat_mat[:, :, -1:, :]), axis=2)
 concat_mat = T.concatenate((concat_mat[:, :, :, :1], concat_mat, concat_mat[:, :, :, -1:]), axis=3)
```
which, BTW, would be more readable as:
```python
 concat_mat = T.concatenate((up_input[:, :, 0, :], up_input, concat_mat[:, :, -1, :]), axis=2)
 concat_mat = T.concatenate((concat_mat[:, :, :, 0], concat_mat, concat_mat[:, :, :, -1]), axis=3)
```

lamblin: @SinaHonari This looks like the un-optimized graph of the concatenation on only one axis (axis=2), when done in two steps. What I wanted to check is if the two joins in here would be optimized by only one (one join per axis), so for that I tried to compile it:
```python
>>> up_input = T.tensor4('up_input')
>>> concat_mat = T.concatenate((up_input[:, :, 0:1, :], up_input), axis=2)
>>> concat_mat = T.concatenate((concat_mat, concat_mat[:, :, -1:, :]), axis=2)
>>> f = theano.function([up_input], concat_mat)
>>> theano.printing.debugprint(f)
Join [id A] ''   3
 |TensorConstant{2} [id B]
 |Join [id C] ''   1
 | |TensorConstant{2} [id B]
 | |Subtensor{::, ::, int64:int64:} [id D] ''   0
 | | |up_input [id E]
 | | |Constant{0} [id F]
 | | |Constant{1} [id G]
 | |up_input [id E]
 |Subtensor{::, ::, int64::} [id H] ''   2
   |Join [id C] ''   1
   |Constant{-1} [id I]
```
Which shows us there are still 2 joins for axis=2, which is wasteful.

So, given that information, you should replace those two joins by only one, like @fvisin suggested, and do the same for axis=3 as well:
```python
concat_mat = T.concatenate((up_input[:, :, 0:1, :], up_input, up_input[:, :, -1:, :]), axis=2)
concat_mat = T.concatenate((concat_mat[:, :, :, 0:1], concat_mat, concat_mat[:, :, :, -1:]), axis=3)
```

@fvisin Your second block would not work, as indexing by `0` or `-1` removes a dimension, and makes it impossible to concatenate, unless you want to explicitly add a `newaxis` or a dimshuffle, which would make readability even worse.
```

Comment (fvisin) [inline theano/tensor/nnet/abstract_conv.py]: @lamblin: Oh right! Not enough coffee for me yet I guess! :)

BTW, please verify if you agree with what I wrote to Sina via email.

File: theano/tensor/nnet/abstract_conv.py (hunk 25 above)
Comment (SinaHonari) [inline theano/tensor/nnet/abstract_conv.py]: I tried your 2D theano version, it does not work.

Comment (fvisin) [inline theano/tensor/nnet/abstract_conv.py]: As the error message tells you (`"The 'tensors' argument must be either a tuple or a list`), the argument of concatenate should be a list:
```python
half_kern = T.arange(1, ratio + 1, dtype=floatX)  # Actually, half plus the middle element
kern = T.concatenate([half_kern, half_kern[-2::-1]])
```
Please try to solve these easy errors on your own. :)

SinaHonari: I already resolved the errors of 1D case, I was asking about 2D not the 1D :)

fvisin: What are you doing for the 2D kernel? Can you write the code here?

fvisin: Anyway, this should work:
```python
# kern1D = the 1D version of the kernel
theano.tensor.stacklists([kern1D[i] * kern1D[j] for i in range(2*ratio - 1) for j in range(2*ratio-1)]).reshape([2*ratio-1, 2*ratio-1])
```

lamblin: What I had in mind for the 2D kernel was something like:
```
hkern = bilinear_kernel_1D(hratio).dimshuffle('x', 0)
vkern = bilinear_kernel_1D(vratio).dimshuffle(0, 'x')
kern = hkern * vkern
```

lamblin: The issue with the `for` loop is that it needs `ratio` to be an int instead of a symbolic variable again.

SinaHonari: why hratio and vratio instead of just ratio?

SinaHonari: The 2D version works.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -204,18 +206,13 @@ def conv2d_grad_wrt_inputs(output_grad,
...
-    :type input_shape: tuple/list of len 2 of int or Constant variable.
+    :type input_shape: tuple/list of len 4, with the first two dimensions
+        being None or int or Constant and the last two dimensions being
+        Tensor or int or Constant.
     :param input_shape: The shape (row and column size) of the
         input (upsampled) parameter.
-        Not Optional, since given the output_grad_shape and the subsample values,
-        multiple input_shape may be plausible.
+        Not Optional, since given the output_grad_shape and the subsample
+        values, multiple input_shape may be plausible.
```
fvisin: If I am not wrong in the case `subsample` == 1 there is no ambiguity.
@lamblin do we want to make `input_shape` optional in that case or we don't care?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -263,19 +260,43 @@ def conv2d_grad_wrt_inputs(output_grad,
...
+        assert isinstance(input_shape[dim], (theano.tensor.TensorConstant,
+                                             int, type(None)))
+    for dim in xrange(2, 4):
+        assert isinstance(input_shape[dim], (theano.tensor.TensorVariable,
+                                             theano.tensor.TensorConstant,
+                                             int))
+
+    # setting the last two dimensions of input_shape to None, if
+    # the type of these dimensions is TensorType.
+    input_shape_no_tensor = input_shape[:2]
```
fvisin: Let's call this `numerical_input_shape`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -263,19 +260,43 @@ def conv2d_grad_wrt_inputs(output_grad,
...
+    if isinstance(input_shape[2], theano.tensor.TensorVariable):
+        input_shape_no_tensor = input_shape_no_tensor + (None, )
+    else:
+        input_shape_no_tensor = input_shape_no_tensor + (input_shape[2], )
+
+    if isinstance(input_shape[3], theano.tensor.TensorVariable):
+        input_shape_no_tensor = input_shape_no_tensor + (None, )
+    else:
+        input_shape_no_tensor = input_shape_no_tensor + (input_shape[3], )
```
fvisin: This should have the same behavior (please check it), but is more readable:
```python
numerical_input_shape = input_shape
for i in [2,3]:
    if isinstance(input_shape[i], theano.tensor.TensorVariable):
        input_shape[i] = None
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -313,7 +334,9 @@ def conv2d_grad_wrt_weights(input,
         You can give ``None`` for any element of the list to specify that this
         element is not known at compile time.
 
-    :type filter_shape: None, tuple/list of len 4 of int or Constant variable.
+    :type filter_shape: tuple/list of len 4 with the first two dimensions
+        being None or Constant and the last two dimensions being Tensor
+        or constant.
```
fvisin: **C**onstant, with uppercase C

theano/tensor/nnet/abstract_conv.py
```diff
@@ -204,20 +206,15 @@ def conv2d_grad_wrt_inputs(output_grad,
...
-    :param output_grad_shape: The shape of the output_grad parameter.
-        Optional, possibly used to choose an optimal implementation.
-        You can give ``None`` for any element of the list to specify that this
-        element is not known at compile time.
-
-    :type input_shape: tuple/list of len 2 of int or Constant variable.
+    :type input_shape: tuple/list of len 4, with the first two dimensions
+        being None or int or Constant and the last two dimensions being
+        Tensor or int or Constant.
     :param input_shape: The shape (row and column size) of the
```
fvisin: Extend this description to all the dimensions (that are now 4 instead of 2)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -263,19 +260,37 @@ def conv2d_grad_wrt_inputs(output_grad,
...
+
+    # checking the type of input_shape
+    for dim in [0, 1]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorConstant,
+                                             int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorVariable,
+                                             theano.tensor.TensorConstant,
+                                             int))
+
```
fvisin: Please check also `filter_shape`'s dtype

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+    # checking the type of filter_shape
+    for dim in [0, 1]:
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorConstant,
+                                              int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorVariable,
+                                              theano.tensor.TensorConstant,
+                                              int))
+
```
fvisin: Check also `input_shape`'s type

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+    """
+
+    hkern = bilinear_kernel_1D(ratio=ratio, normalize=normalize).dimshuffle('x', 0)
+    vkern = bilinear_kernel_1D(ratio=ratio, normalize=normalize).dimshuffle(0, 'x')
+    kern = hkern * vkern
+    return kern
+
+
+def bilinear_kernel_1D(ratio, normalize=True):
+    """ This function gets a ratio
```
fvisin: This docstring has a weird formatting. 
Also, according to [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings), the multi-line docstring should consist of a (**i.e., one**) summary line, followed by a blank line, followed by a more elaborate description.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+    and returns the normalized 1D filters that can
+    be applied to any image to upsample it by that ratio
+    using bilinear interpolation.
+
+    :type ratio: python integer or symbolic theano
```
fvisin: int or Constant/Scalar Theano tensor

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+    if normalize:
+        kern /= float(ratio)
+    return kern
+
+
+def bilinear_upsampling(input,
+                        ratio,
+                        input_shape=None,
+                        use_1D_kernel=True):
+    """This function takes a stack of 2D inputs and upsamples it (in 2D space)
```
fvisin: Ditto

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+    repeated one extra time compared to the first row and column which makes
+    the upsampled tensor asymmetrical on both sides. This does not happen when
+    the upsampling ratio is odd.
+
+    :type input: symbolic 4D tensor
+    :param input: mini-batch of feature map stacks, of shape
+        (batch size, input channels, input rows, input columns).
+        See the optional parameter ``input_shape`` that will be upsampled.
+
+    :type ratio: integer
```
fvisin: int or Constant/Scalar Theano tensor

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+    by the given ratio using bilinear interpolation with fixed (not learned)
+    kernels. When the upsampling ratio is even, the last row and colum is
+    repeated one extra time compared to the first row and column which makes
+    the upsampled tensor asymmetrical on both sides. This does not happen when
+    the upsampling ratio is odd.
+
+    :type input: symbolic 4D tensor
+    :param input: mini-batch of feature map stacks, of shape
+        (batch size, input channels, input rows, input columns).
+        See the optional parameter ``input_shape`` that will be upsampled.
```
fvisin: "that will be upsampled" was probably supposed to be part of the previous sentence?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+
+    :type ratio: integer
+    : param ratio: the ratio by which the input is upsampled
+        in the 2D space.
+
+    :type input_shape: None, tuple/list of len 4 of int or
```
fvisin: Looking into it more carefully, this is not even used at the moment! 
There is no need for the user to specify the upsampled shape, as we can easily compute it and can be passed as tensor. We can benefit from the user specifying the `batch_size` and the number of `input channels` though. Remove this argument and add two optional arguments for `batch_size` and `num_input_channels`. Check their type (None, int or Constant), default to None.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+        be upsampled seperately by 1D kernels, otherwise they are
+        upsampled together using a 2D kernel. The final result is the
+        same, only the speed can differ, given factors such as
+        upsampling ratio.
+    """
+
+    T = theano.tensor
+    bs, ch, row, col = input.shape
+    up_input = input.reshape((-1, 1, row, col))
+    up_bs, up_ch, up_row, up_col = up_input.shape
```
fvisin: We need up_bs as a numerical variable. This line can be removed. At its place you can exploit the new `batch_size` and `num_input_channels` arguments (if defined) to compute the `up_bs` as a product of the two.

fvisin: Note that `up_bs` should be `None` if one or both are `None`.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+
+    if use_1D_kernel:
+        kern = bilinear_kernel_1D(ratio=ratio, normalize=True)
+        kern_size = kern.shape[0]
+        # upsampling rows
+        upsampled_row = conv2d_grad_wrt_inputs(output_grad=concat_mat,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis,
+                                                            :, np.newaxis],
+                                               input_shape=(up_bs, up_ch,
```
fvisin: `up_ch` is always 1, no need to use a variable.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+                                                             1),
+                                               border_mode=(pad, 0),
+                                               subsample=(ratio, 1),
+                                               filter_flip=True)
+        # upsampling cols
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=upsampled_row,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis,
+                                                            np.newaxis, :],
+                                               input_shape=(up_bs, up_ch,
```
fvisin: Ditto.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -357,13 +374,185 @@ def conv2d_grad_wrt_weights(input,
...
+                                               border_mode=(0, pad),
+                                               subsample=(1, ratio),
+                                               filter_flip=True)
+    else:
+        kern = bilinear_kernel_2D(ratio=ratio, normalize=True)
+        kern_row, kern_col = kern.shape
+        upsampled_mat = conv2d_grad_wrt_inputs(output_grad=concat_mat,
+                                               filters=kern[np.newaxis,
+                                                            np.newaxis, :, :],
+                                               input_shape=(up_bs, up_ch,
```
fvisin: Ditto.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -263,19 +260,37 @@ def conv2d_grad_wrt_inputs(output_grad,
...
+        assert isinstance(input_shape[dim], (theano.tensor.TensorConstant,
+                                             int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorVariable,
+                                             theano.tensor.TensorConstant,
+                                             int))
+
+    # setting the last two dimensions of input_shape to None, if
+    # the type of these dimensions is TensorVariable.
+    numerical_input_shape = input_shape
```
fvisin: `numerical_input_shape` should be a list, or you won't be able to modify it.

fvisin:
I didn't check the tests as there were many errors in the code. 

From now on please run the tests locally *before* submitting a new version. Your own test failed and the majority of the errors I pointed out could have been found by you just by either executing your code at least once or running the tests. Please incorporate the changes I suggested, verify that the tests pass and when you are done push a new version.

theano/tensor/nnet/abstract_conv.py
lamblin: I think we do not care for now.

theano/tensor/nnet/abstract_conv.py
fvisin: Errata corrige: let's keep input_shape, which should be [Constant / int / None] * 4

theano/tensor/nnet/abstract_conv.py
fvisin: Here use input_shape to compute the upscaled shape. Check for None with:
```python
try:
    up_bs = input_shape[0] * input_shape[1]
except TypeError:                                                                                                                                                                                                                            
    up_bs = None
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +340,185 @@ def conv2d_grad_wrt_weights(input,
...
+    :type use_1D_kernel: bool
+    :param use_1D_kernel: if set to true, row and column will
+        be upsampled seperately by 1D kernels, otherwise they are
+        upsampled together using a 2D kernel. The final result is the
+        same, only the speed can differ, given factors such as
+        upsampling ratio.
+    """
+
+    T = theano.tensor
+    bs, ch, row, col = input.shape
```
fvisin: Here you probably just need `row, col = input.shape[2:]`

SinaHonari: I think I understand what you meant at the time by taking only batch_size and number_of_channels, since the last two dimensions can be taken as symbolic variables from input itself. However, the first two dims are still required. Given also what you have written here, we just need to take from the user only the first two dims.

fvisin: You are definitely right! I got fooled by the old docstring twice. Good job, thanks!

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -406,3 +410,270 @@ def test_grad_types(self):
...
+            is of shape (batch size, num channels, 2*ratio, 2*ratio).
+
+        """
+        kern, shp = self.make_1D_multiplier(ratio)
+        up_1D = TwoByTwoMat[:, :, :, :1] * kern[::-1] + \
+            TwoByTwoMat[:, :, :, 1:] * kern
+        up_2D = up_1D[:, :, :1, :] * kern[::-1][:, np.newaxis] + \
+            up_1D[:, :, 1:, :] * kern[:, np.newaxis]
+        num_concat = (ratio - 1) // 2
+        for i in xrange(num_concat):
```
fvisin: Use `range` here instead of `xrange`. This should fix the test.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -139,20 +143,17 @@ def conv2d_grad_wrt_inputs(output_grad,
...
-        parameter.  Not Optional, since given the output_grad_shape
+    input_shape : [None/int/Constant] * 2 + [Tensor/int/Constant] * 2
+        The shape of the input (upsampled) parameter.
+        A tuple/list of len 4, with the first two dimensions
+        being None or int or Constant and the last two dimensions being
+        Tensor or int or Constant.
+        Not Optional, since given the output_grad shape
         and the subsample values, multiple input_shape may be
         plausible.
     filter_shape : list of 4 symbolic or real ints
```
fvisin: Please modify the description of `filter_shape` to follow the previous convention i.e. use as type `[None/int/Const] * 4` and add in the description that it's a tuple/list of len 4.
+
Also please add to the `border_mode` docstring in the `half` description that it's also known as `same` elsewhere.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,229 @@ def conv2d_grad_wrt_weights(input,
...
+def bilinear_kernel_2D(ratio, normalize=True):
+    """Compute 2D kernel for bilinear upsampling
+
+    This function builds the 2D kernel that can be used to upsample
+    a tensor by the given ratio using bilinear interpolation.
+
+    Parameters
+    ----------
+    ratio: int or Constant/Scalar Theano tensor
+        the ratio by which an image will
```
fvisin: Can you please fix the weird line breaking here by using the line length as much as possible?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,229 @@ def conv2d_grad_wrt_weights(input,
...
+    a tensor by the given ratio using bilinear interpolation.
+
+    Parameters
+    ----------
+    ratio: int or Constant/Scalar Theano tensor
+        the ratio by which an image is upsampled
+        by the returned filter in the 1D space.
+
+    normalize: bool
+        indicates whether to normalize the
```
fvisin: Ditto.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,229 @@ def conv2d_grad_wrt_weights(input,
...
+
+    Parameters
+    ----------
+    input: symbolic 4D tensor
+        mini-batch of feature map stacks, of shape
+        (batch size, input channels, input rows, input columns)
+        that will be upsampled.
+
+    ratio: integer
+        the ratio by which the input is upsampled
```
fvisin: Ditto.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,229 @@ def conv2d_grad_wrt_weights(input,
...
+        in the 2D space (row and col size).
+
+    batch_size: None, int or Constant variable
+        The size of the first dimension of the input variable.
+        Optional, possibly used to choose an optimal implementation.
+        You can give ``None`` for any element of the list to specify that this
```
fvisin: It's not a list. Remove these two lines. Specify instead that `batch_size` will be used only if `num_input_channels` is not None

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -406,3 +410,270 @@ def test_grad_types(self):
...
+        kernel_1D = np.array([1, 2, 3, 4, 3, 2, 1])
+        np.testing.assert_allclose(kernel_1D, f())
+
+        # checking normalized filters
+        kernel = bilinear_kernel_1D(ratio=4, normalize=True)
+        f = theano.function([], kernel)
+        kernel_1D = kernel_1D / 4.
+        np.testing.assert_allclose(kernel_1D, f())
+
+    def make_1D_multiplier(self, ratio):
```
fvisin: The name of this function is not clear to me, `numerical_1D_kernel` seems to be more immediately comprehensible.

fvisin: Wait...it's not computing a kernel. I should look into it more carefully.

BTW, for the numerical kernels you can create two functions that return respectively:
```python
[i for i in range(1, ratio+1) + range(ratio-1, 0, -1)]  # 1D kernel
np.array([i*j for i in bilinear_kernel_1D(ratio, False) for j in bilinear_kernel_1D(ratio, False)]).reshape(2*ratio-1, 2*ratio-1) # 2D kernel
```

fvisin: Ok I looked into your code more carefully, disregard my previous comment on the name.

For what concerns the second part, you can define these functions:
```python
def numerical_kernel_1D(ratio):
    return [i for i in range(1, ratio+1) + range(ratio-1, 0, -1)] 
def numerical_kernel_2D(ratio):
    return np.array([i*j for i in numerical_kernel_1D(ratio) for j in numerical_kernel_1D(ratio)]).reshape(2*ratio-1, 2*ratio-1) 
```
to get the kernels to be used in all the tests of `test_bilinear_kernel_1D` and `test_bilinear_kernel_2D`

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -406,3 +410,270 @@ def test_grad_types(self):
...
+            to upsample an array.
+
+        int
+            The size of the multipliers array
+
+        """
+        kern = np.arange(ratio + 1)
+        return kern, kern.shape[0]
+
+    def get_upsampled_twobytwo_mat(self, TwoByTwoMat, ratio):
```
fvisin: By [PEP8 convention](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) variables should begin with a lowercase letter, as opposed to classes.

fvisin: Also, it's formally a tensor and not a matrix, so it's probably better not to have `Mat` in the name.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -406,3 +410,270 @@ def test_grad_types(self):
...
+        np.testing.assert_allclose(f(), up_mat_2d, rtol=1e-06)
+
+    def test_compare_1D_and_2D_upsampling_values(self):
+        """Compare 1D and 2D upsampling
+
+        This method verifies the bilinear upsampling done by using
+        1D and 2D kernels will generate the same result.
+
+        """
+        # checking upsamplign of 5
```
fvisin: typo

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -406,3 +410,270 @@ def test_grad_types(self):
...
+                                     batch_size=5, num_input_channels=4,
+                                     use_1D_kernel=True)
+        mat_2D = bilinear_upsampling(input=input_x, ratio=5,
+                                     batch_size=5, num_input_channels=4,
+                                     use_1D_kernel=False)
+        f_1D = theano.function([], mat_1D)
+        f_2D = theano.function([], mat_2D)
+        np.testing.assert_allclose(f_1D(), f_2D(), rtol=1e-06)
+
+        # checking upsamplign of 8
```
fvisin: typo

theano/tensor/nnet/abstract_conv.py
```diff
@@ -216,24 +217,50 @@ def conv2d_grad_wrt_inputs(output_grad,
...
+        assert isinstance(input_shape[dim], (theano.tensor.TensorConstant,
+                                             int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorVariable,
+                                             theano.tensor.TensorConstant,
+                                             int))
+
+    # checking the type of filter_shape
+    for dim in [0, 1, 2, 3]:
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorVariable,
```
fvisin: `filter_shape` should be either None, int or Constant if I'm not wrong.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,229 @@ def conv2d_grad_wrt_weights(input,
...
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorConstant,
+                                              int, type(None)))
+    for dim in [2, 3]:
+        assert isinstance(filter_shape[dim], (theano.tensor.TensorVariable,
+                                              theano.tensor.TensorConstant,
+                                              int))
+
+    # checking the type of input_shape
+    for dim in [0, 1, 2, 3]:
+        assert isinstance(input_shape[dim], (theano.tensor.TensorVariable,
```
fvisin: `input_shape` should be either None, int or Constant if I'm not wrong.

theano/tensor/nnet/abstract_conv.py
SinaHonari: In my code in bilinear_upsampling code, for one of the dimensions of filter_dim, I'm using a tensor variable and it is working fine. I think tensor_variable can be used, but not sure.

Another comment, if the user just passes None (as the default value), I guess this line would fail since it's neither a list nor a tuple.

SinaHonari: Either I should skip this test if filter_shape is None, or I make it a list of 4 None values: [None, None, None, None]. Which one is better?

fvisin: Just for reference, as we discussed IRL a few minutes ago `filter_shape` should be either `None` or a 4-elements Iterable whose elements can be either `None`, `int` or `Constant`.

theano/tensor/nnet/abstract_conv.py
fvisin: Remember to add the `None` option to the documentation as well.

SinaHonari:
I just pushed the changes

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,215 @@ def conv2d_grad_wrt_weights(input,
...
+        by the indicated ratio using bilinear interpolation in one dimension.
+
+    """
+
+    T = theano.tensor
+    half_kern = T.arange(1, ratio + 1, dtype=theano.config.floatX)
+    kern = T.concatenate([half_kern, half_kern[-2::-1]])
+
+    if normalize:
+        kern /= float(ratio)
```
fvisin: This will not work when ratio is a tensor. Since `kern` is a floatX we don't need it though, so please remove the float cast. We should also test `bilinear_kernel_1D` for tensor ratio.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -320,13 +339,215 @@ def conv2d_grad_wrt_weights(input,
...
+
+def bilinear_kernel_1D(ratio, normalize=True):
+    """Compute 1D kernel for bilinear upsampling
+
+    This function builds the 1D kernel that can be used to upsample
+    a tensor by the given ratio using bilinear interpolation.
+
+    Parameters
+    ----------
+    ratio: int or Constant/Scalar Theano tensor the ratio by which an image
```
fvisin: Please specify that it has to be of int* dtype

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -406,3 +410,171 @@ def test_grad_types(self):
...
+
+    def test_bilinear_kernel_1D(self):
+        """Test 1D kernels used in bilinear upsampling
+
+        This method tests the correctness of the
+        1D kernel values used in bilinear upsampling
+        for some upsampling ratios.
+
+        """
+        for ratio in [2, 3, 4, 5, 6, 7, 8, 9]:
```
fvisin: add `ratio = as_tensor_variable(ratio)`, so that we check that it works for tensors

fvisin:
@SinaHonari thank you for your work! Please fix my last comments and squash the commits together. 
@lamblin The rest LGTM, once Travis passes we can merge.

SinaHonari:
I updated the PR. Do you want all of the commits to get squashed into one (from the beginning of the ticket)?

fvisin:
Yes please.
Il 04 mar 2016 6:28 PM, "Sina Honari" <notifications@github.com> ha scritto:

SinaHonari:
I already suqashed the last 4 commits to make Travis work only on one version. Is it needed to squash all?

fvisin:
Yes please, after we fix the last error (Travis is not passing) squash all of them together. There is no reason to merge all those commits to theano, as I don't think anybody will ever want to check out a WIP version of this op. 

For the moment wait for news from our side. I will look into this Travis's error message tomorrow with @lamblin:
```
AssertionError: AbstractConv2d_gradInputs Theano optimization failed: there is no implementation available supporting the requested options. Did you exclude both "conv_dnn" and "conv_gemm" from the optimizer? If on GPU, is cuDNN available and does the GPU support it? If on CPU, do you have a BLAS library installed Theano can link against?
```

lamblin:
Looks good to me, thanks!
Once you rebase and squash, Travis should restart, and the issue should be fixed.

SinaHonari:
No problem. 
Just updated the PR.

fvisin:
Great @SinaHonari thanks! 
I will merge when Travis passes.

SinaHonari:
@fvisin You're welcome.

fvisin:
@SinaHonari there is a [flake8 error](https://travis-ci.org/Theano/Theano/jobs/114665022#L1236). Please fix it (and check for other flake8 errors running `flake8 abstract_conv.py` and `flake8 test_abstract_conv.py`), squash the commits and push again.

SinaHonari:
I updated the PR. Hopefully, this time it passes :)

fvisin:
Thanks! I'll keep my finger crossed, I can't wait to merge it! :)

fvisin:
@lamblin Travis failed again, it might be time to raise the tolerance of the test and/or to look into it. I will relaunch the test in the meanwhile.

fvisin:
@SinaHonari the test still fails, this time for [python3 compatibility](https://travis-ci.org/Theano/Theano/jobs/114846212#L3076) reasons. Please follow [this suggestion](http://stackoverflow.com/a/13318111) to fix it and squash and push again as usual. Let's hope for the last time! :)

SinaHonari:
There are many ways to do this. I can do the following:
_kern = range(1, ratio + 1)
kern.extend(range(ratio - 1, 0, -1))
return np.array(kern)_
Is it fine with you?

fvisin:
What I was suggesting is to add an explicit cast to list to every range, e.g. change `range(1, ratio + 1)` into  `list(range(1, ratio + 1))`. Can you try to do that and if the tests pass locally (i.e. it works for python 2.7) push it?

SinaHonari:
OK, I will replace it with 
```np.array(list(range(1, ratio + 1)) + list(range(ratio - 1, 0, -1)))```
which also passes locally

lamblin:
@fvisin #4203 has been merged.

SinaHonari:
Just updated the PR. I get my finger crossed :)

fvisin:
I almost can't believe it, it's finally merged!!
Thank you @SinaHonari and @lamblin for the hard work

SinaHonari:
Great. Finally :)

shampool:
Hello, I got the error with bi_linear_upsampling, any suggestions?
https://github.com/Theano/Theano/issues/4351

PR #3952, opened 2016-01-30
Fix doc gen

nouiz:


nouiz:
as this should fix the docgen on the web site, I merge it now.

PR #3954, opened 2016-01-30
LogAbsDet

harpone:
Hi! Go the LogAbsDet op done with gradient, but no GPU code so far... if anyone wants to help/ point me to PyCuda/CUDA implementation of SVD, let me know!


harpone:
Umm I have *no* idea whatsoever why the Travis CI failed... help??

harpone:
Also, maybe this should be in the sandbox instead??

harpone:
Ah OK, fails with float32... on it

PR #3957, opened 2016-01-31
LogAbsDet

harpone:


harpone:
Considered putting this into the sandbox...

PR #3958, opened 2016-01-31
LogABsDet

harpone:
Travis CI corrections (Python 3.3 compatibility fixes)


nouiz:
You don't need to close a PR to update it. You can just push on the branch that was used to create this PR on github and the PR will be updated.

harpone:
@nouiz damn... I figured I could probably do something like that :P 

Anyway thanks, learning lots of new stuff here :)

Issue #150, opened 2011-10-24
Op/optimization TODO: log determinant


dwf:
Now that we have a determinant Op we should probably look into this.

There are numerically stable ways to compute the natural logarithm of the determinant, a quantity which is often needed in evaluating the log probability of a multivariate Gaussian distribution. NumPy provides this as [`numpy.slogdet`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.slogdet.html) for "sign log determinant" (it computes `log(abs(det(X))`) and also returns the sign of the determinant).

The derivative of the log determinant also has a particularly simple form as `inv(X).T == inv(X.T)`.

harpone:
Just out of curiosity, why hasn't this been done yet? I haven't coded any Theano ops, but I could give it a shot... maybe just a log abs det, since that's probably the most common case (in MLE estimation etc)?

FYI I'm hitting the stability issue pretty badly...

lamblin:
I'm not sure why this has not been done yet, or maybe some people ended up implementing that in their own repositories.
If you want to give it a shot, it would be welcome. There should be a section on how to extend theano in the online doc. As long as you use python/numpy for the implementation and do not need a GPU version, it should be quite straightforward.
Please let us know if you have any issues or questions, either in a pull request or at `theano-dev`.

harpone:
OK I'll give it a try!

harpone:
Found a couple of implementations on Github. [This one](https://github.com/FlorianSeidel/GOL/blob/9f7716e78a78fbdd7ae8ecac5a58b701f6a4da41/gol/goal/constraints.py) does an SVD first and takes the log of the diagonal (not sure why it's diagonal squared... need to check the math; also has a funny expression for the gradient... it's correct but computationally expensive).

[This one](https://github.com/kolia/subunits/blob/05f40cbc45214d80567b132c95f16913ca2a0cc7/kolia_theano.py#L12) uses numpy's slogdet.

Any preferences? SVD on GPU should be a simple copy/paste job I think, but I haven't found any CUDA/PyCUDA/gnumpy code for the slogdet...

nouiz:
You could check inside numpy code how slogdet is implemented. Maybe it does
it with an svd? Can you compate the output of both op to see if the result
is the same?

There is svd code on the GPU somewhere, so with that, it wouldn't be too
hard to make a GPU implementation

harpone:
[...]

PR #3959, opened 2016-01-31
LogAbsDet

harpone:
Travis CI corrections (Python 3.3 compatibility fixes; PEP8 fixes)


theano/tensor/nlinalg.py
```diff
@@ -726,3 +726,37 @@ def norm(x, ord):
...
+        x = theano.tensor.as_tensor_variable(x)
+        o = theano.tensor.scalar(dtype=x.dtype)
+        return Apply(self, [x], [o])
+
+    def perform(self, node, inputs, outputs):
+        try:
+            (x,) = inputs
+            (z,) = outputs
+            s = numpy.linalg.svd(x, compute_uv=False)
+            log_abs_det = numpy.sum(numpy.log(numpy.abs(s)))
```
nouiz: I didn't check the math. But this can be done by reusing current op. The svd op and the elemwise op. This will be better as the elemwise op will get optimized by Theano and will work on the GPU (for the elemwise).

Then the only think missing to make it work on the GPU would be to have an svn op on the GPU. There is one implementation here that won't request c code: http://scikit-cuda.readthedocs.org/en/latest/generated/skcuda.linalg.svd.html

theano/tensor/tests/test_nlinalg.py
```diff
@@ -516,3 +518,29 @@ def test_numpy_compare(self):
...
+        utt.seed_rng()
+        self.op_class = LogAbsDet
+        self.op = logabsdet
+
+    def validate(self, input_mat):
+        x = theano.tensor.matrix()
+        f = theano.function([x], self.op(x))
+        out = f(input_mat)
+        svd_diag = numpy.linalg.svd(input_mat, compute_uv=False)
+        numpy_out = numpy.sum(numpy.log(numpy.abs(svd_diag)))
```
nouiz: I think it would be better to compare again the numpy function you want to implement.

theano/tensor/nlinalg.py
nouiz: In fact, I think it is a good idea to keep this as an op. It will lower memory usage during the gradient on the CPU and will provide a gradient (the currrent svd op don't have a grad implemented).

but an easy way to have this on the GPU would be to make an optimization that will convert this op to a graph that reuse the elemwise on the GPU and make an svd op. For the gradient, MatrixInverse will also need a GPU op to have this work on the GPU. But the first step, is to get this working!

theano/tensor/tests/test_nlinalg.py
harpone: That would be the slogdet function then, but it probably does something a bit different since the results are differ (relatively) by about 1E-8, as I discussed in the original issue (here)[https://github.com/Theano/Theano/issues/150]. Maybe I should compare against the slogdet but increase the tolerance of the allclose?

theano/tensor/nlinalg.py
harpone: Yeah, I have no idea what the gradient of SVD would be :) grad of log abs det is very simple on the other hand, so maybe it's better to leave this out of the theano optimizations...

OK didn't know there's no GPU op for MatrixInverse... so step 1 would now be to get LogAbsDet work on GPU? Is it possible to use the skcuda.linalg.svd directly? (I have no idea how the make_thunk works...). So not sure if I should try to do the GPU SVD here directly or in the SVD op? Damn it's still a bit confusing :)

lamblin:

theano/tensor/tests/test_nlinalg.py
nouiz: Having 1e-8 difference isn't a problem. You can have that type of error just by changing the order of addition when you add a vector of floats.

If the utt.assert_allclose pass, the I would compare again numpy.slogdet. If not, we need to understand why as they are supported to be equivalent. Maybe the fix would be to raise the comparison tolerance.

theano/tensor/nlinalg.py
nouiz: We didn't document much making new gpu op as we are in transition to a new gpu back-end...

For a first pass, you can make one GPU op that have its gradient on the CPU and use it directly. Then after, you can add an optimization that will convert the CPU op to the GPU op (I can help, it is very easy when you know how to do it).

To make a GPU op, I would use as an example the code of CuFFTOp in sandbox/cuda/fftconv.py. This use scikits.cuda for computation, so you can reuse it for the svd implementation.

harpone: OK thanks, I'll probably get that done some time soon (been a bit busy lately)!

theano/tensor/tests/test_nlinalg.py
harpone: I think my test cases are a bit silly, as the svd values are pretty big, about 1e+2 or so.. so maybe I should just rescale the test matrices (by e.g. 1 / sqrt(size)), then the error will probably be below the tolerance.

nouiz: Did you looked into this? I'm checking if we can merge this PR.

harpone: Sorry, haven't had the time to look into it as I just started a new day job recently... but it's definitely on my to-do list, just need to find a window for it

twiecki:
This might be a relevant PR for pymc3 too. Any way that the theano simplification would change a manual `log(det())` to use this `logdet` directly?

theano/tensor/nlinalg.py
```diff
@@ -726,3 +726,37 @@ def norm(x, ord):
...
+            z[0] = numpy.asarray(log_abs_det, dtype=x.dtype)
+        except Exception:
+            print('Failed to compute logabsdet of {}.'.format(x))
+            raise
+
+    def grad(self, inputs, g_outputs):
+        gz, = g_outputs
```
superbobry: The comma here can be hard to spot, why not use single-element list syntax instead?

```
[gz] = g_outputs
[x] = input
```

superbobry:
This looks great! Is it possible to automatically optimize `log(det(...))` to `logdet(...)`?

theano/tensor/nlinalg.py
```diff
@@ -726,3 +726,37 @@ def norm(x, ord):
...
+        return Apply(self, [x], [o])
+
+    def perform(self, node, inputs, outputs):
+        try:
+            (x,) = inputs
+            (z,) = outputs
+            s = numpy.linalg.svd(x, compute_uv=False)
+            log_abs_det = numpy.sum(numpy.log(numpy.abs(s)))
+            z[0] = numpy.asarray(log_abs_det, dtype=x.dtype)
+        except Exception:
```
superbobry: Is it possible to be more specific about the exception type here?

theano/tensor/nlinalg.py
harpone: > [gz] = g_outputs
yeah sounds good...

> This might be a relevant PR for pymc3 too. Any way that the theano simplification would change a manual log(det()) to use this logdet directly?

there could be some subtleties with that, for example if the matrix is not positive semi-definite => det is negative... not sure how to handle that

theano/tensor/nlinalg.py
harpone: > Is it possible to be more specific about the exception type here?
 
hmm I guess one should look inside numpy.linalg.svd for similar clauses... or the tests??

Actually I'm pretty swamped with other stuff at the moment, so I'd appreciate if someone else finished this stuff :)

theano/tensor/nlinalg.py
harpone: > Any way that the theano simplification would change a manual log(det()) to use this logdet directly?

There may be some problems with that if positivity of det is not enforced... not sure how to handle that...

theano/tensor/nlinalg.py
harpone: > This looks great! Is it possible to automatically optimize log(det(...)) to logdet(...)?

see [here](https://github.com/Theano/Theano/pull/3959#discussion_r55125239)

harpone:
Fixed the tests and did the minor change as suggested by @superbobry but Travis CI seems to be failing... don't know why and I have no time to check now

abergeron:
There was some breakage in the travis setup.  A the new run should go fine (assuming there are no problems in this PR).

harpone:
@nouiz I think this is done now (except for the GPU version)

lamblin:
Now that the release is over, is there anything preventing this from being merged?

nouiz:
I have comments that haven't been done. Mostly, the author started a new jobs, so he have difficulty to find time to finish this.

harpone:
It should be working now. I fixed the tests too 12 days ago. The only thing missing is the CUDA code, which is of course a major omission... prolly can't find time for that any time soon...

springcoil:
Is it possible to get this merged in without the CUDA code, and seperate that to another PR? 
Due to it's interest for PyMC3

twiecki:
@nouiz What would adding GPU support entail? Writing CUDA? Presumably SVD already exists for GPU.

PR #3961, opened 2016-02-01
Fix a test on GPU with mode=FAST_COMPILE

nouiz:
also fix a test with python 3


abergeron:
Ok when travis passes.

Issue #3431, opened 2015-09-22
Move ProfileMode Hint to the new profiler


nouiz:
Useful for user and would clean up code by removing old "if 0" ed code.
- [x] Add hint feature to new profiler (#3870)
- [ ] Clean-up the part in "if 0:"

lamblin:
So, the deprecated [`ProfileMode`](https://github.com/Theano/Theano/blob/master/theano/compile/profilemode.py#L112) had a feature printing some hints or tips for the user when some cases were detected, for instance:
- using amdlibm for faster elemwise operations
- setting floatX=float32
- when dot product between one float32 and one float64 matrix are computed (not optimized to use gemm)

That code starts at https://github.com/Theano/Theano/blob/master/theano/compile/profilemode.py#L607, there is another commented-out version at https://github.com/Theano/Theano/blob/master/theano/compile/profiling.py#L1307

Now, we use the regular `Mode`, but print a profile when `theano.config.profile == True`. However, we are missing the hints. This is the last thing preventing the complete removal of `ProfileMode`.

So, this ticket is about moving that code from `ProfileMode` to `theano.compile.profiling`, maybe in a new method `ProfileStats.print_tips()`, that would be called at the end of `ProfileStats.summary()`.
When this is done, the commented-out code in `profiling.py` can be removed.

Saizheng:
https://github.com/Theano/Theano/pull/3870

nouiz:
done by https://github.com/Theano/Theano/pull/3870

PR #3964, opened 2016-02-01
Cleaunup profileMode deprecation in docs

### LINKED ISSUE #3929: Officially deprecate `ProfileMode` in the docs
Also finish the clean-up `theano.compile.profiling.py` (#3431)

memimo:
fixes #3929 
@carriepl


carriepl:
- In doc/library/compile/mode.txt : ``'PROFILE_MODE'``: Deprecated. Use the string ProfileMode.
- In doc/tutorial/using_gpu.txt : 2 references to ProfileMode which need to be updated
This is what I found using 'git grep'. I think this is all that remains to be done.

memimo:
I added the changes (forced push so you might have not got notification.)

doc/library/compile/mode.txt
```diff
@@ -23,7 +23,7 @@ Theano defines the following modes by name:
 - ``'DebugMode'``: A mode for debugging. See :ref:`DebugMode <debugmode>` for details.
 - ``'ProfileMode'``: Deprecated, use the Theano flag :attr:`config.profile`.
 - ``'DEBUG_MODE'``: Deprecated. Use the string DebugMode.
-- ``'PROFILE_MODE'``: Deprecated. Use the string ProfileMode.
+- ``'PROFILE_MODE'``: Deprecated. Use the flag profile.
```
carriepl: Put a reference to the profile flag (just like 2 lines above, simply copy that whole description and paste it here)

memimo: fixed.

doc/cifarSC2011/advanced_theano.txt
```diff
@@ -179,96 +180,189 @@ Inplace optimization
...
-            <created/inplace/view> <Apply node>
-        <created/inplace/view> is taked from the op declaration, not ...
-             2508800B  [2508800] v InplaceDimShuffle{1,0}(x)
-                6272B  [6272] i Gemv{inplace}(w, ...)
-                3200B  [3200] c Elemwise{Composite{...}}(y, ...)
-
-    Here are tips to potentially make your code run faster (if you think of new ones, suggest them on the mailing list).
-    Test them first, as they are not guaranteed to always provide a speedup.
-      - Try the Theano flag floatX=float32
+    Function profiling
```
lamblin: Is that really the new profile for the same function?

doc/tutorial/using_gpu.txt
```diff
@@ -213,8 +213,8 @@ Tips for Improving Performance on GPU
   frequently-accessed data (see :func:`shared()<shared.shared>`).  When using
   the GPU, *float32* tensor ``shared`` variables are stored on the GPU by default to
   eliminate transfer time for GPU ops using those variables.
-* If you aren't happy with the performance you see, try building your functions with
-  ``mode='ProfileMode'``. This should print some timing information at program
+* If you aren't happy with the performance you see, try running your script with
+  ``profil=True`` flag. This should print some timing information at program
```
lamblin: "profile=True"

doc/cifarSC2011/advanced_theano.txt
carriepl: That's what Mehdi told me. It seems that now the data for the memory profile is all weird (it says 0KB memory usage everywhere) but that would be a bug with the memory profile rather than with this PR. What do you think, @lamblin ?

lamblin: I think it really strange that there was a `Gemv` in the original profile, but that it disappeared from the new one.
Also, the first ones says "10001 Theano fct call", while the new one reports only 1 call.
The 0KB probably comes from the fact that "All Apply nodes have output sizes that take less than 1024B". There seems to be some really small memory usage: "remaining 20 Apply account for  171B/171B ((100.00%)) of the Apply with dense outputs sizes", which is again not consistent with the previous profile.

Actually, I think the previous profile was for another function, one that is more interesting to profile than this one (which computes only one loop on scalars).
I think it would be better to do the profiling exercise on something more interesting, so we can show something where the memory is bigger than 1kB, and the bottleneck does not appear to be `Shape_i`.

memimo: There was no mention that the log was profile of which script so I just profiled one of the examples above. 
I take another look to see if can figure out the current log belongs to which script.

carriepl: Ok, thanks

lamblin:
Can you also remove the code in `if False:` [here](https://github.com/Theano/Theano/blob/master/theano/compile/profiling.py#L1398-L1710)? That is the second point in #3431.
Thanks,

carriepl:
@memimo : any news on this?

memimo:
Sorry, just start working on it now.

memimo:
So, I can't really find the original script for in the profile example. Instead I can take the logistis_sgd.py from DeepLearningTutorials and put the proifle for one of the frist function call?

carriepl:
Yes, good idea.

nouiz:
Just tell it in the doc that this is the code you used.

nouiz:
Why not use:
http://deeplearning.net/software/theano/tutorial/examples.html#a-real-example-logistic-regression

It won't request data from another repo for reader of the code. You could
just link to it.

On Mon, Feb 8, 2016 at 3:18 PM, Frédéric Bastien <frederic.bastien@gmail.com
> wrote:

> Just tell it in the doc that this is the code you used.
>

doc/cifarSC2011/advanced_theano.txt
```diff
@@ -179,96 +180,200 @@ Inplace optimization
...
+              1600B  [(400,)] c GpuFromHost(y)
+       ... (remaining 7 Apply account for  448B/3800192B ((0.01%)) of the Apply with dense outputs sizes)
+
+        <created/inplace/view> is taken from the Op's declaration.
+        Apply nodes marked 'inplace' or 'view' may actually allocate memory, this is not reported here. If you use DebugMode, warnings will be emitted in those cases.
+
+    Here are tips to potentially make your code run faster
+                     (if you think of new ones, suggest them on the mailing list).
+                     Test them first, as they are not guaranteed to always provide a speedup.
+      Sorry, no tip for today.
```
lamblin: I think we should investigate that. This example keeps a `dot` in the graph, because of mixed `float32`/`float64` dtype, and that is definitely something that should be warned about in the tips, I thought we did check for that.

doc/cifarSC2011/advanced_theano.txt
```diff
@@ -179,96 +180,200 @@ Inplace optimization
 Profiling
 ---------
 
-- To replace the default mode with this mode, use the Theano flags ``mode=ProfileMode``
+- To replace the default mode with this mode, use the Theano flags ``profile=True``
 
-- To enable the memory profiling use the flags ``ProfileMode.profile_memory=True``
+- To enable the memory profiling use the flags ``profile=True,profile_memory=True``
 
-Theano output:
```
abergeron: This is basically an archival copy of the docs.  Please don't modify it.

carriepl: There is nothing in the file that seems to mark it as such. Meaning that someone could find it and try to replicate what's in there. Is it that important to keep invalid documentation around to confuse users? The presentation was given in 2011. That's roughly 5 years ago!

nouiz: Now we don't put in Theano presentation, so this problem don't happen anymore. But I think as it is in Theano, we can update it. Or we should put a warning at the top that tell it isn't updated and point elsewhere.

As it was updated, I would keep it like this.

doc/cifarSC2011/advanced_theano.txt
carriepl: No objection but I think this falls outside of the scope of this CCW ticket. I'll create a ticket for it and link to this ticket for explanation.

doc/cifarSC2011/advanced_theano.txt
carriepl: Works for me (keeping it as it is).

doc/cifarSC2011/advanced_theano.txt
carriepl: TIcket was created : #4035

carriepl:
The 2 issues that remained have been addressed so this should be fine for merge.

PR #3588, opened 2015-11-02
Gpu dnn pool

harmdevries89:
fix gh-3417

PR #3965, opened 2016-02-01
gpu dnn pool takes tensor variables

harmdevries89:
rebase of gh-3588


lamblin:
Can you check pep8 and python3 syntax? This causes at least some tests to fail.

nouiz:
travis is failing in python 3. You need to use the new print interface:

print(....)

This is the place that have problems.

  File "/home/travis/build/Theano/Theano/theano/sandbox/cuda/dnn.py", line 1810
    print mode_flag

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -343,7 +348,7 @@ def fn(x):
                 assert any([isinstance(node.op, AveragePoolGrad)
                             for node in fc.maker.fgraph.toposort()])
             c_out = fc(data)
-            utt.assert_allclose(c_out, g_out)
```
nouiz: revert back that change. The old code is equivalent, but give a better error message.

theano/sandbox/cuda/dnn.py
```diff
@@ -1656,35 +1694,42 @@ class GpuDnnPoolGrad(DnnBase):
...
+        if mode == 'average':
+            mode = 'average_inc_pad'
+        assert mode in ('max', 'average_inc_pad', 'average_exc_pad')
+        self.mode = mode
 
+    def make_node(self, inp, out, inp_grad, ws, stride, pad):
         inp = as_cuda_ndarray_variable(inp)
+        assert (inp.ndim in [4, 5])
         inp_grad = as_cuda_ndarray_variable(inp_grad)
+        assert (inp_grad.ndim in [4, 5])
```
nouiz: assert that inp_grad.ndim == inp.ndim

theano/sandbox/cuda/dnn.py
```diff
@@ -1656,35 +1694,42 @@ class GpuDnnPoolGrad(DnnBase):
...
         out = as_cuda_ndarray_variable(out)
+        assert(out.ndim in [4, 5])
```
nouiz: idem

theano/sandbox/cuda/dnn.py
```diff
@@ -1544,20 +1554,41 @@ def c_init_code_struct(self, node, name, sub):
...
+        elif self.mode == "average_exc_pad":
+            mode_flag = 'CUDNN_POOLING_AVERAGE_COUNT_EXCLUDE_PADDING'
+            if version() == -1:
+                raise Exception("cudnn v1 do not support average_exc_pad")
+        else:
+            raise NotImplementedError("Unsupported pooling model.")
+
         return """
-cudnnStatus_t err%(name)s;
+fprintf(stderr, "test_forward\\n");
```
nouiz: remove this print

theano/sandbox/cuda/dnn.py
```diff
@@ -1734,17 +1787,30 @@ def c_cleanup_code_struct(self, node, name):
...
+            mode_flag = 'CUDNN_POOLING_MAX'
+        elif self.mode == "average_inc_pad":
+            mode_flag = 'CUDNN_POOLING_AVERAGE_COUNT_INCLUDE_PADDING'
+        elif self.mode == "average_exc_pad":
+            mode_flag = 'CUDNN_POOLING_AVERAGE_COUNT_EXCLUDE_PADDING'
+            if version() == -1:
+                raise Exception("cudnn v1 do not support average_exc_pad")
+        else:
+            raise NotImplementedError("Unsupported pooling model.")
+        print mode_flag
```
nouiz: remove this print

theano/sandbox/cuda/dnn.py
```diff
@@ -1734,17 +1787,30 @@ def c_cleanup_code_struct(self, node, name):
...
         return """
 cudnnStatus_t err%(name)s;
+//raise(SIGINT);
```
nouiz: remove

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -314,11 +317,12 @@ def fn(x):
             # Test the GPU grad + GPU implementation
             def fn(x):
                 dnn_op = cuda.dnn.dnn_pool(
-                    x, ws=(ws, ws),
-                    stride=(stride, stride),
-                    pad=pad,
+                    x, ws=ws,
+                    stride=stride,
+                    pad=pad_,
```
nouiz: Why this change? Do it work with the old code? It should work to not break too much compatibility.

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -285,22 +286,24 @@ def test_pooling():
...
 
         # Test the grad
         for shp in [(1, 1, 2, 2),
                     (1, 1, 3, 3)]:
             data = numpy.random.normal(0, 1, shp).astype("float32") * 10
 
-            ws = 2
-            stride = 2
-            if pad[0] > stride or pad[1] > stride:
+            ws = theano.shared(numpy.array([2, 2]))
```
nouiz: Why put in shared var all those changes?

theano/sandbox/cuda/dnn.py
```diff
@@ -1544,20 +1554,41 @@ def c_init_code_struct(self, node, name, sub):
...
 if (input%(name)s != NULL) { cudnnDestroyTensorDescriptor(input%(name)s); }
 if (output%(name)s != NULL) { cudnnDestroyTensorDescriptor(output%(name)s); }
+if (pool%(name)s != NULL) { cudnnDestroyPoolingDescriptor(pool%(name)s); }
 """ % dict(name=name)
 
     def c_code(self, node, name, inputs, outputs, sub):
-        desc = inputs[1]
+        ws = inputs[1]
+        stride = inputs[2]
+        pad = inputs[3]
```
nouiz: This change the interface. If user have pickled graph, it won't work. They will get a strange error.

I'm not sure what to do. I see 2 options:

1) if len(inputs) != 2, give a better error message like, "Theano DnnPool internal changed, rebuild your graph"
2) You could try to get the data from the descriptor

2) is better as it allow to unpickle old graph. Can you try to do it? I think it isn't very hard. To test it, pickle a small graph with the master of Theano that have both pool op in it. Then unpickle it and compile it with an empty cache.

nouiz:
in test_dnn.py, remove this failing test: test_dnn_pool_desc_merge it isn't useful anymore.

theano/sandbox/cuda/tests/test_dnn.py
harmdevries89: I want to test if the op accepts tensor variables. @lamblin suggested to use shared variables when you pass it to verify grad.

theano/sandbox/cuda/tests/test_dnn.py
harmdevries89: Ohh you mean don't change this test, and make a separate test to verify if it works with tensor variables?

theano/sandbox/cuda/dnn.py
harmdevries89: Ok, I understand. Will try option 2).

nouiz:

/home/travis/build/Theano/Theano/theano/sandbox/cuda/dnn.py:1730:1: W293 blank line contains whitespace

theano/sandbox/cuda/tests/test_dnn.py
nouiz: We need both to be tested. But I would go to make this test that is more complete test the new code and make a new test, that test only 1 case with the old interface. We just need to make sure the old interface work.

theano/sandbox/cuda/tests/test_dnn.py
nouiz: ok, that sound good.

nouiz:
also add the pickle file.

nouiz:
There is at least this test that fail:

FAIL: theano.sandbox.cuda.tests.test_dnn.test_dnn_pool_desc_merge
  File "/Tmp/lisa/os_v5/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/bastienf/repos/Theano/theano/sandbox/cuda/tests/test_dnn.py", line 85, in test_dnn_pool_desc_merge
    assert len(descs) == 1, f.maker.fgraph
AssertionError: [DeepCopyOp(*1 -> GpuDnnPool{mode='max'}(GpuContiguous(GpuFromHost(x)), TensorConstant{(2,) of 2}, TensorConstant{(2,) of 1}, TensorConstant{(2,) of 0})), *1]
-------------------- >> begin captured logging << --------------------
theano.gof.cmodule: WARNING: not all the following op(s) implement c_code_cache_version(). This makes them recompiled for each process.[<theano.sandbox.cuda.DnnVersion object at 0x7f1883b9f350>]
--------------------- >> end captured logging << ---------------------

Change the assert to == 0.

nouiz:
There was no other errors in test_dnn.py
thanks

theano/sandbox/cuda/dnn.py
```diff
@@ -1358,40 +1358,66 @@ class GpuDnnPool(DnnBase):
...
+            pad = theano.tensor.constant(node.inputs[1].owner.op.pad)
+            node.inputs[1] = ws
+            node.inputs.append(st)
+            node.inputs.append(pad)
+            storage_map[ws] = [None]
+            storage_map[st] = [None]
+            storage_map[pad] = [None]
+            compute_map[ws] = [False]
+            compute_map[st] = [False]
+            compute_map[pad] = [False]
```
abergeron: This is probably not right for constants.  You should put the value in the storage_map and set compute_map to True.

abergeron:
I would also delete the test about merged descriptors since we don't use external descriptors anymore.  If we change the assert to 0 it doesn't test much.

theano/sandbox/cuda/dnn.py
harmdevries89: @nouiz Do you agree?

nouiz: Yes. Check if they are instance of theano.Constant and if so, set True
instead of False.

harmdevries89: Ok, but we explicitly cast them to theano constants, right? So I can just the value in storage_map and set computer_map to true?

nouiz: They won't always be Constant. What is there is good when it isn't constant. I don't know why it work they it is constant, but I think we should fix this inconsistency. Normally, constant will have the .data member of the constant in the storage_map and have the compute_map to True.

So mostly, split this in 2 cases. If not Constant, do as now, if Constant, do the change I described above.

nouiz:
thanks. Don't forget to make a comment when you update a PR and want us to see it. Otherwise we don't get notification and this delay the review.

Issue #3967, opened 2016-02-01
NaNguard only on CPU?


jeanpijon:
Hi I am trying to use NaNguard on GPU, it runs, but speed is very low and also GPU's are not loaded (CPU's are on 100% though) is it a feature or bug? I am using the same code as runs normally for GPU.
Thanks
Jean

nouiz:
Here is a quick PR with a fix:

https://github.com/Theano/Theano/pull/3968

But there will be another problem on the GPU if you use cudnn. Just disable
cudnn and you should be fine for convolution on the GPU for now. The
problem exist for scan too. Here is the PR that should fix this problem:

https://github.com/Theano/Theano/pull/3768

jeanpijon:
Wow, that was fast... I'll give it a try. thanks

jeanpijon:
Unfortunatelly, still doesn't work:
```
AssertionError: NaN detected
NanGuardMode found an error in an input of this node.
Node:
GpuGemv{inplace}(GpuAllocEmpty.0, TensorConstant{1.0}, <CudaNdarrayType(float32, matrix)>, GpuDimShuffle{1}.0, TensorConstant{0.0})
...


Apply node that caused the error: GpuGemv{inplace}(GpuAllocEmpty.0, TensorConstant{1.0}, <CudaNdarrayType(float32, matrix)>, GpuDimShuffle{1}.0, TensorConstant{0.0})
Toposort index: 287
Inputs types: [CudaNdarrayType(float32, vector), TensorType(float32, scalar), CudaNdarrayType(float32, matrix), CudaNdarrayType(float32, vector), TensorType(float32, scalar)]
Inputs shapes: [(20,), (), (20, 20), (20,), ()]
Inputs strides: [(1,), (), (20, 1), (1,), ()]
Inputs values: ['not shown', array(1.0, dtype=float32), 'not shown', 'not shown', array(0.0, dtype=float32)]
Outputs clients: [[GpuDimShuffle{x,0}(GpuGemv{inplace}.0)]]
```

After applying #3968 and #3768 and disabling cuDNN (at least I think i managed to disable it - via dnn.library_path=/dev/null)

Cheers
Jan

lamblin:
Right, the AllocEmpty should be ignored by NaNGuardMode, but is not at the moment.
I don't remember if we already have a mechanism for that.

nouiz:
I rebased the other PR to have both fix:
https://github.com/Theano/Theano/pull/3768

[...]

PR #3968, opened 2016-02-01
Enable opt in NanGuardMode. Make it use the default optimizer.

nouiz:
Fix gh-3967.


lamblin:
OK to merge if Travis passes.

PR #3970, opened 2016-02-01
Fix the thunk differences in DebugMode

abergeron:
The workaround with signatures should also be applied for unsigned integers.


nouiz:
ok to merge when travis pass.

PR #3973, opened 2016-02-02
Fix problems in tests in DebugMode.

abergeron:


theano/typed_list/tests/test_basic.py
```diff
@@ -579,5 +579,5 @@ def test_correct_answer(self):
         z = make_list((x, y))
         fc = theano.function([a, b], c)
         fz = theano.function([x, y], z)
-        self.assertTrue(fc(A, B) == [A, B])
-        self.assertTrue(fz(X, Y) == [X, Y])
+        self.assertTrue((m == n).all() for m, n in zip(fc(A, B), [A, B]))
```
nouiz: why this change is needed? It weaken the test. Now we aren't sure both version return an object of the same type. fc could return a list and fz could return a tuple.

I'm not opposed to this change, but I don't understand why it is needed.

lamblin: That test was failing because of the way the comparison between lists works: it was raising an exception when trying to compare that had the same value but were not the same object. You can try the following:

    >>> [np.zeros(2)] == [np.zeros(2)]
    ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

The test was only passing (but not in DebugMode) because the arrays in the returned list were exactly the same object:

    >>> z = np.zeros
    >>> [z] == [z]
    True

PR #3982, opened 2016-02-02
Fix Gpu......WithBias to work properly for float64.

abergeron:


nouiz:
What was the consequence? Crash? result with less precission?

abergeron:
It would not compile with float64

nouiz:
That mean a crash due to an exception. thanks.

Issue #3979, opened 2016-02-02
Did the Param disappear/change in the bleeding edge , Theano-0.8.0.dev0


gammaguy:
I installed the Theano-0.8.0.dev0 to get Keras Relu working.

Now I can't import Param

1 import theano.tensor as T
2 from theano import Param
3 from theano import function
4 x, y = T.dscalars('x', 'y')
5 z = x + y
6 f = function([x, Param(y, default=1)], z)

I get:
/home/paul/anaconda2/bin/python /home/paul/.PyCharm50/config/scratches/scratch_6
Using gpu device 0: GeForce GTX TITAN X (CNMeM is disabled)
Traceback (most recent call last):
  File "/home/paul/.PyCharm50/config/scratches/scratch_6", line 2, in <module>
    from theano import Param
ImportError: cannot import name Param

Process finished with exit code 1

lamblin:
Yes, `Param` has been removed. We thought it was only used internally, and now `In` supports all the features of `Param`. We should put it back with a deprecation warning, I guess.
The solution is to replace `Param(y, default=1)` with `In(y, value=1)`, like was done at https://github.com/lisa-lab/DeepLearningTutorials/pull/133

gammaguy:
TY for answering so quickly.

I got it from the Theano tutorial section.

Setting a Default Value for an Argument section  from http://deeplearning.net/software/theano/tutorial/examples.html#computing-more-than-one-thing-at-the-same-time

lamblin:
OK, I did not remember about that. That's another good reason to add back `Param` with a deprecation warning.
On top of it, this section of the doc has been changed in the repository (it now tells to use `In()`), but the online doc does not seem to have been updated. I'll try to investigate that.

nouiz:
The problem with the only doc is that now warning are consided as error.
I'm working on a PR to add back Param.

nouiz:
Here is the PR: https://github.com/Theano/Theano/pull/3984

gammaguy:
There is still a problem as you changed the Param --> In but forgot to change the default --> value as per 
"The solution is to replace Param(y, default=1) with In(y, value=1), like was done at lisa-lab/DeepLearningTutorials#133"

In the Setting a Default Value for an Argument section of: http://deeplearning.net/software/theano/tutorial/examples.html 

>>> f = function([x, In(y, **default**=1)], z)
and
>>> f = function([x, In(y, default=1), In(w, **default**=2, name='w_by_name')], z)

[...]

PR #3984, opened 2016-02-03
Add back Param and make it deprecated

nouiz:
fix gh-3979.


lamblin:
This is deprecated, so I think we do not need a test, as long as it was manually tested once.
OK when Travis passes, if it has been manually tested.

nouiz:
I tested it manually on an old DLT that still used Param. I got the warning printed.

PR #3986, opened 2016-02-03
Fix test when mode!=DebugMode on the GPU.

nouiz:
This fix the failing test in the daily buildbot.


theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
```diff
@@ -835,6 +835,7 @@ def conv_grad(mode, bs, ch, nf, rImg1, rImg2, rFlt1, rFlt2, subsample, op):
     outputs.extend([corr_op_dik, conv_op_dik,
                     corr_op_dki, conv_op_dki])
 
+    theano_mode = None
```
abergeron: This should instead be `global theano_mode`.  It's declared at the top of the file.

nouiz:
I updated the PR.

abergeron:
Ok for me when travis passes.

PR #3987, opened 2016-02-03
fixing make_node param passing for conv_grad (conv_transpose)

SinaHonari:
related to issues #3872 and #3681


nouiz:
Is there tests failing without this fix? If not, we need to add tests.

SinaHonari:
No test is failing, but without this fix there would be errors.
Yes, a new ticket should be opened to test these methods.

lamblin:

PR #3991, opened 2016-02-04
keep stack trace in optimizations of nnet folder

### LINKED ISSUE #3927: Keep stack trace in opt: theano/tensor/nnet
Sub-ticket of #3018.
Use `copy_stack_trace` for optimizations defined in `theano/tensor/nnet/`, and add checks that the stack trace is still there during tests.
- optimizations in `nnet.py` should be already covered, but tests are missing
- optimizations for abstract conv in `opt.py` are already covered, tests including

vmichals:
fix #3927

### DIFF
--- a/theano/tensor/nnet/conv3d2d.py
+++ b/theano/tensor/nnet/conv3d2d.py
@@ -329,9 +329,11 @@ def make_gpu_optimizer(op, to_gpu):
                 new_inp = list(node.inputs)
                 for idx in to_gpu:
                     new_inp[idx] = cuda.gpu_from_host(new_inp[idx])
-                new_node = cuda.host_from_gpu(op()(*new_inp))
-                copy_stack_trace(node.outputs[0], new_node)
-                return [new_node]
+                result_node = op()(*new_inp)
+                copy_stack_trace(node.outputs[0], result_node)
+                transfer_node = cuda.host_from_gpu(result_node)
+                copy_stack_trace(node.outputs[0], transfer_node)
+                return [transfer_node]
         if node.op == cuda.gpu_from_host:
             # gpu_from_host(op) -> op(gpu_from_host)
             host_input = node.inputs[0]
@@ -342,7 +344,7 @@ def make_gpu_optimizer(op, to_gpu):
                 for idx in to_gpu:
                     new_inp[idx] = cuda.gpu_from_host(new_inp[idx])
                 new_node = op()(*new_inp)
-                copy_stack_trace(node.outputs[0], new_node)
+                copy_stack_trace(host_input, new_node)
                 return [new_node]
         return False
     local_to_gpu.__name__ = "local_to_gpu_" + op.__name__
--- a/theano/tensor/nnet/conv3d2d.py
+++ b/theano/tensor/nnet/conv3d2d.py
@@ -3,6 +3,7 @@ from theano.gradient import DisconnectedType
 from theano.gof import Op, Apply, TopoOptimizer
 from theano import tensor
 import theano.sandbox.cuda as cuda
+from theano.tensor.opt import copy_stack_trace
 
 
 def get_diagonal_subtensor_view(x, i0, i1):
@@ -328,7 +329,9 @@ def make_gpu_optimizer(op, to_gpu):
                 new_inp = list(node.inputs)
                 for idx in to_gpu:
                     new_inp[idx] = cuda.gpu_from_host(new_inp[idx])
-                return [cuda.host_from_gpu(op()(*new_inp))]
+                new_node = cuda.host_from_gpu(op()(*new_inp))
+                copy_stack_trace(node.outputs[0], new_node)
+                return [new_node]
         if node.op == cuda.gpu_from_host:
             # gpu_from_host(op) -> op(gpu_from_host)
             host_input = node.inputs[0]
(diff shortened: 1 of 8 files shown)


theano/tensor/nnet/conv3d2d.py
```diff
@@ -328,7 +329,9 @@ def local_to_gpu(node):
                 new_inp = list(node.inputs)
                 for idx in to_gpu:
                     new_inp[idx] = cuda.gpu_from_host(new_inp[idx])
-                return [cuda.host_from_gpu(op()(*new_inp))]
+                new_node = cuda.host_from_gpu(op()(*new_inp))
+                copy_stack_trace(node.outputs[0], new_node)
```
nouiz: I think we should also put the stack trace on the node created by `op()(*new_inp)`. If the computation have problem, it will be with that, not with the transfer.

nouiz: Here, I would put the stack trace on both variable.

nouiz:

/home/travis/build/Theano/Theano/theano/tensor/nnet/sigm.py:29:1: E302 expected 2 blank lines, found 1

theano/tensor/nnet/conv3d2d.py
lamblin: That's right, the variable returned by `op()(*new_inp)` should get the stack trace as well.

theano/tensor/nnet/conv3d2d.py
```diff
@@ -338,7 +341,9 @@ def local_to_gpu(node):
                 new_inp = list(op_node.inputs)
                 for idx in to_gpu:
                     new_inp[idx] = cuda.gpu_from_host(new_inp[idx])
-                return [op()(*new_inp)]
+                new_node = op()(*new_inp)
+                copy_stack_trace(node.outputs[0], new_node)
```
lamblin: The stacktrace here should come from `host_input` (`node.inputs[0]`), which is the stack trace from when the actual `op` was instantiated, not when the `gpu_from_host` was.

theano/tensor/nnet/nnet.py
```diff
@@ -751,7 +751,9 @@ def local_logsoftmax(node):
             isinstance(node.inputs[0].owner.op, Softmax)):
         inVars = node.inputs[0].owner.inputs[0]
         new_op = LogSoftmax()
-        return [new_op(inVars)]
+        new_node = new_op(inVars)
+        copy_stack_trace(node.outputs[0], new_node)
```
lamblin: In that case, we should add both the stack trace of the log (from `node.outputs[0]`) and the one from the softmax (from `node.inputs[0]`), because if logsoftmax raises an error, there are chances the issue was in the original softmax.
You can do `copy_stack_trace([node.inputs[0], node.outputs[0]], new_node).

theano/tensor/nnet/nnet.py
```diff
@@ -862,6 +865,7 @@ def local_softmax_with_bias(node):
                 if sm_bias.type == node.outputs[0].type:
                     # This condition is not always true. See the test
                     # nnet/tests/test_nnet.py:T_SoftmaxWithBias.test_broadcast
+                    copy_stack_trace(node.outputs[0], sm_bias)
```
lamblin: That is actually not needed, as `sm_bias` already has a stack trace.

theano/tensor/nnet/sigm.py
```diff
@@ -955,10 +960,13 @@ def local_inv_1_plus_exp(node):
             if len(nonconsts) == 1:
                 if nonconsts[0].owner and nonconsts[0].owner.op == tensor.exp:
                     if scalars and numpy.allclose(numpy.sum(scalars), 1):
-                        return opt._fill_chain(
+                        out = opt._fill_chain(
                             sigmoid(
                                 tensor.neg(nonconsts[0].owner.inputs[0])),
                             scalar_inputs)
+                        # keep stack trace
+                        copy_stack_trace(node.outputs[0], out)
```
lamblin: Here, we should combine the stack traces of `exp(x)`, `1 + exp(x)`, and `1/(1 + exp(x))`, respectively `nonconsts[0]`, `inv_arg` and `node.outputs[0]`. Can you add that with a comment?

theano/tensor/nnet/sigm.py
```diff
@@ -979,7 +987,9 @@ def local_1msigmoid(node):
             except Exception:
                 return
             if numpy.allclose(numpy.sum(val_l), 1):
-                return [sigmoid(-sub_r.owner.inputs[0])]
+                out = sigmoid(-sub_r.owner.inputs[0])
+                copy_stack_trace(node.outputs[0], out)
```
lamblin: I would tend to also add the stack trace of the initial sigmoid, `sub_r`.
I think it is OK not to add a stack trace to `-sub_r.owner.inputs[0]`.

lamblin:
I'm done with my review of these changes.
Are there other optimizations that are remaining?
Then, the next step is adding the tests.

vmichals:
Thanks for the Feedback!
I applied all the changes.
I think I covered all the optimizations in the folder.
I'll try to add the unit tests next.

lamblin:
Thanks!
Please let us know when you start adding the unit tests.

vmichals:
I think the edits in the last commit should contain the last missing unit tests for the nnet folder

lamblin:
That looks good for now, thanks!
We will probably have to revisit the tests when we have a helper function, for instance to check that the sigmoid specifically has a trace, not only the output of the compiled function, but it is good to have them in the mean time.

PR #3993, opened 2016-02-04
Fix last DLT problem(s) with the new backend.

abergeron:
This fixes a number of problems.

The main one is that the lifter for Pool was removing operations from the graph in certain circumstances.

Others are:
 - The lifter for Print was mostly broken
 - AbstractConv could stay unoptimized in certain cases of mixed input types.

There is now code in the AbstractConv ops to convert all inputs to be of the same type as the first one, which avoids the second problem.

### DIFF
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -451,7 +451,7 @@ def gpu_print_wrapper(op, cnda):
 @op_lifter([tensor.printing.Print])
 def local_gpu_print_op(node, context_name):
     x, = node.inputs
-    gpu_x, = x.owner.inputs
+    gpu_x = as_gpuarray_variable(x, context_name=context_name)
     new_op = node.op.__class__(global_fn=gpu_print_wrapper)
     new_op.old_op = node.op
     return new_op(gpu_x)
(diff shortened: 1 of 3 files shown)


theano/sandbox/gpuarray/opt.py
```diff
@@ -786,10 +783,9 @@ def local_gpua_softmaxwithbias(node, context_name):
 @register_opt('fast_compile')
 @op_lifter([theano.tensor.opt.Assert])
 def local_assert(node, context_name):
-    if (node.inputs[0].owner and
-            isinstance(node.inputs[0].owner.op, HostFromGpu)):
-        return [host_from_gpu(node.op(node.inputs[0].owner.inputs[0],
-                                      *node.inputs[1:]))]
+    return [host_from_gpu(node.op(as_gpuarray_variable(node.inputs[0],
+                                                       context_name),
+                                  node.inputs[1:]))]
```
lamblin: Is the removal of the `*` here intended? I don't think it makes sense.

nouiz:
travis have this error:

/home/travis/build/Theano/Theano/theano/tensor/nnet/abstract_conv.py:641:32: E128 continuation line under-indented for visual indent

nouiz:
Ok to merge when the flake8 test is fixed and you have run locally the tests under sandbox/gpuarray

abergeron:
Locally, the everfailing scan test fails, but everything else passes (Although I did not test with the flake8 correction).

nouiz:

PR #3994, opened 2016-02-04
Changes to GpuSolve op to improve compatibility with slingalg Solve op

matt-graham:
## Reason for PR

The current CULA `GpuSolve` op implementation is not entirely consistent in functionality with the `scipy.linalg` based `Solve` op. In particular it:

  1. Only allows two dimensional `b` second inputs (while `Solve` correctly handles either solving for a vector or matrix `b`).
  2. Does not allow specification of any information about the structure of the matrix `A` first input (while `Solve` takes an argument `A_structure` which can be used to specify whether the matrix has some special structure which may be useful for applying a more efficient and/or numerically stable solve implementation).

As the `local_gpu_solve` optimization replaces `Solve` with `GpuSolve` where possible, the first of the above issues can lead to cases where a graph will compile and run successfully when running on a CPU but fail when optimized to run on a GPU when a `Solve` instance with a one-dimensional `b` input is replaced with a `GpuSolve` instance.

The second issue does not cause any direct problems at the moment but it means that the `local_gpu_solve` optimization is lossy - information about any structure in the `A` input is lost on transforming from a `Solve` to `GpuSolve` op, preventing a further optimizer from potentially exploiting this structure to substitute a more specialized GPU based solver op.

This PR is a first attempt at fixing both these issues. 

I have added a `A_structure` property to the `GpuSolve` op which can be optionally specified in `__init__` (with string checked against same tuple of valid options as in `slingalg.Solve` op). At the moment nothing is done with this but it allows the information to at least be retained. I have also changed the `local_gpu_solve` optimization so that the `A_structure` property value is passed through when transforming from a `Solve` op to a `GpuSolve` instance.

I have also added some logic to allow for one-dimensional `b` second inputs, with a reshaped 2D view created if a 1D input is provided before passing to the CULA based solver implementation and the 2D output reshaped back to 1D before returning if appropriate. I have added some further test cases for 1D second inputs and some test cases to check that exceptions are raised if a one- or three-dimensional first input is passed to `GpuSolve`.

## Questions

  * Currently the boolean `lower` property in `Solve` is not passed through to `GpuSolve` as this seems to be redundant as `A_structure` can only be specifically specified as `upper_triangular` or `lower_triangular` which already encodes this information (assuming `lower` is a flag to specify if lower or upper triangular). Would it be worth also having this property retained by `GpuSolve`?
  * Most of the existing GPU optimizations seem to use aliased ops (e.g. `gpu_solve` as opposed to `GpuSolve`) in the optimization definition. Because of the need to pass the `A_structure` property through when instantiating the op currently `GpuSolve` is used directly. Is this problematic?
  * When implementing the unit test cases I noticed the current tests manually set the `numpy` random seed rather than using `unittest_tools.seed_rng()` as recommended in the [developer guide](http://deeplearning.net/software/theano/extending/unittest.html). Given I was adding further tests using random inputs I implemented these to follow the guidelines and changed the existing tests to be consistent with this. Were these special cases which should have been left with manual seeding?

Cheers

### DIFF
--- a/theano/sandbox/cuda/tests/test_cula.py
+++ b/theano/sandbox/cuda/tests/test_cula.py
@@ -32,13 +32,22 @@ class TestCula(unittest.TestCase):
     def run_gpu_solve(self, A_val, x_val):
         b_val = numpy.dot(A_val, x_val)
         A = theano.tensor.matrix("A", dtype="float32")
-        b = theano.tensor.matrix("b", dtype="float32")
-
+        if x_val.ndim == 1:
+            b = theano.tensor.vector("b", dtype="float32")
+        else:
+            b = theano.tensor.matrix("b", dtype="float32")
         solver = cula.gpu_solve(A, b)
         fn = theano.function([A, b], [solver])
         res = fn(A_val, b_val)
         x_res = numpy.array(res[0])
         utt.assert_allclose(x_res, x_val)
+
+    def test_invalid_input_fail_3D(self):
+        def invalid_input_func():
+            A = theano.tensor.tensor3("A", dtype="float32")
+            b = theano.tensor.matrix("b", dtype="float32")
+            solve = cula.gpu_solve(A, b)
+        self.assertRaises(AssertionError, invalid_input_func)
 
     def test_diag_solve(self):
         """ Diagonal solve test case with 1D vector as 2D matrix RHS. """
--- a/theano/sandbox/cuda/tests/test_cula.py
+++ b/theano/sandbox/cuda/tests/test_cula.py
@@ -26,6 +26,9 @@ else:
 
 
 class TestCula(unittest.TestCase):
+    def setUp(self):
+        utt.seed_rng()
+
     def run_gpu_solve(self, A_val, x_val):
         b_val = numpy.dot(A_val, x_val)
         A = theano.tensor.matrix("A", dtype="float32")
@@ -39,7 +42,6 @@ class TestCula(unittest.TestCase):
 
     def test_diag_solve(self):
         """ Diagonal solve test case with 1D vector as 2D matrix RHS. """
-        numpy.random.seed(1)
         A_val = numpy.asarray([[2, 0, 0], [0, 1, 0], [0, 0, 1]],
                               dtype="float32")
         x_val = numpy.random.uniform(-0.4, 0.4, (A_val.shape[1],
@@ -48,7 +50,6 @@ class TestCula(unittest.TestCase):
 
     def test_sym_solve(self):
         """ Symmetric solve test case with 1D vector as 2D matrix RHS. """
-        numpy.random.seed(1)
         A_val = numpy.random.uniform(-0.4, 0.4, (5, 5)).astype("float32")
         A_sym = (A_val + A_val.T) / 2.0
         x_val = numpy.random.uniform(-0.4, 0.4, (A_val.shape[1],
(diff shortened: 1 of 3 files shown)


theano/sandbox/cuda/tests/test_cula.py
```diff
@@ -26,44 +26,96 @@
...
+
     def run_gpu_solve(self, A_val, x_val):
         b_val = numpy.dot(A_val, x_val)
         A = theano.tensor.matrix("A", dtype="float32")
-        b = theano.tensor.matrix("b", dtype="float32")
-
+        if x_val.ndim == 1:
+            b = theano.tensor.vector("b", dtype="float32")
+        else:
+            b = theano.tensor.matrix("b", dtype="float32")
```
nouiz: You can use this more generic code:
b = theano.tensor.TensorType(dtype="float32", broadcastable=[False,]*x_val.ndim)()

theano/sandbox/cuda/tests/test_cula.py
```diff
@@ -26,44 +26,96 @@
...
+        def invalid_input_func():
+            A = theano.tensor.tensor3("A", dtype="float32")
+            b = theano.tensor.matrix("b", dtype="float32")
+            cula.gpu_solve(A, b)
+        self.assertRaises(AssertionError, invalid_input_func)
+
+    def test_invalid_input_fail_3d(self):
+        """ Invalid solve input test case with 3D tensor as first input. """
+        def invalid_input_func():
+            A = theano.tensor.vector("A", dtype="float32")
```
nouiz: this should be tensor3()

theano/sandbox/cuda/tests/test_cula.py
```diff
@@ -26,44 +26,96 @@
...
         res = fn(A_val, b_val)
         x_res = numpy.array(res[0])
         utt.assert_allclose(x_res, x_val)
 
-    def test_diag_solve(self):
-        numpy.random.seed(1)
+    def test_invalid_input_fail_1d(self):
+        """ Invalid solve input test case with 1D vector as first input. """
+        def invalid_input_func():
+            A = theano.tensor.tensor3("A", dtype="float32")
```
nouiz: this should be vector

nouiz:
For your questions:
1) I'm not sure, so you can let it as now.
2) This depend of each optimization. But normally, we should keep them. Which optimization don't keep the information?
3) You did the right thing. We didn't pass on all test to use the prefered way. But changing them is good.

thanks

matt-graham:
@nouiz : Updated in response to comments.

With regards to my second question above, from the comment by **abergeron** on my similar question in #4006  I've now realised that there is no problem with instantiating the ops directly in the optimization as required. So this question no longer is relevant.

Thanks for the responses to my other questions.

lamblin:
@nouiz Now that he release is over, are we waiting for something specific, or is there a discussion to have about the interface?

nouiz:
This need a full review. I forgot too much about this PR. I don't have time now.

matt-graham:
@lamblin @nouiz Anything in particular I should work on with this to move things forward (if you have time now to review)? Cheers

lamblin:
Sorry for the delay, I'll try to get a bit of time to take care of that.
The first thing is that we stopped working on the `sandbox/cuda` back-end, and are trying to push to the `gpuarray` back-end. We did not port `GpuSolve` there, though, so we could incorporate the updated interface directly without worrying about backward compatibility.

If you have time to work on the port to the new back-end, I think that would be really great. I don't think we have example of Ops using CULA there, but #4535 introduces Ops using `scikits.cuda`, so that could be a starting point. #4584 extends the documentation on the new back-end, but I don't think it covers that point in particular.

There may also be a discussion to have about the implementation, maybe using cuSOLVER instead of CULA, but that could independent.

Issue #3953, opened 2016-01-30
square of negative number is NA


rosstang:
```
x0 = theano.shared(np.array([[-1, 1], [3,4]], dtype=np.float32))
f1 = theano.function([], x0**2)
f1()


Out[6]: array([[ nan,   1.],
       [  9.,  16.]], dtype=float32)
```

I am using the latest theano.

AnishShah:
I'm using the latest theano, as in, the development version and I'm getting the correct answer.
Which version are you using?

![image](https://cloud.githubusercontent.com/assets/3175743/12695064/ce8ac49e-c767-11e5-9b76-abda78206fce.png)

rosstang:
```
In [34]: print theano.__version__
0.8.0.dev0.dev-18319b8f426e99fa209c4910af7208c0d51c41a6
```

I am using windows though.

nouiz:
I also have the good answer. Can you execute this and give me the output:

theano.printing.debugprint(f1, print_type=True)

rosstang:
```
In [755]: theano.printing.debugprint(f1, print_type=True)
Out[755]: HostFromGpu [id A] <TensorType(float32, matrix)> ''   1
 |GpuElemwise{pow,no_inplace} [id B] <CudaNdarrayType(float32, matrix)> ''   0
   |<CudaNdarrayType(float32, matrix)> [id C] <CudaNdarrayType(float32, matrix)>
   |CudaNdarrayConstant{[[ 2.]]} [id D] <CudaNdarrayType(float32, (True, True))>
```

nouiz:
Can you try while changing the place of the -1? Otherwise, it is probably a
problem with your setup, a bug in one of the software we reuse like
nvcc(CUDA compiler).

rosstang:
```
In [778]: x0 = theano.shared(np.array([[1, -1], [3,4]], dtype=np.float32))
f1 = theano.function([], x0**2)
f1()

Out[778]: 
array([[  1.,  nan],
       [  9.,  16.]], dtype=float32)

In [779]: x0 = theano.shared(np.array([[1, 3], [-1,4]], dtype=np.float32))
f1 = theano.function([], x0**2)
f1()

Out[779]: 
array([[  1.,   9.],
       [ nan,  16.]], dtype=float32)
```

Still the same. It looks like the 2 is converted into CudaNdarrayConstant{[[ 2.]]}  (which is a float?)

nouiz:
[...]

PR #3988, opened 2016-02-03
A few small stuff

nouiz:

PR #3995, opened 2016-02-05
Update documentation of fastmath config

rosstang:
Add issue https://github.com/Theano/Theano/issues/3953 to documentation to warn people when using fastmath and fast_compile together.


nouiz:
As I didn't got a fast reply and didn't wanted other people to have the same problem (you aren't the first one to have this problem) I did it in https://github.com/Theano/Theano/pull/3988, but I forgot to update the issue gh-3953. Sorry about that.

I also wanted to not use it in the example, so to not have many people try it.

thanks for the PR.

PR #3997, opened 2016-02-05
Really skip the flake8 part in "everything but tensor".

abergeron:


nouiz:
To be sure it work, can you compare the number of test ran in travis when it finish?

abergeron:
I did a check manually to make sure that this is the proper syntax.  I can also look at travis' numbers if you want.

lamblin:
The number of tests went from 992 to 991, so I'm merging.

PR #3998, opened 2016-02-05
Ignore division by zero in expmgrad when numerator is zero.

computer-whisperer:
This fixes an annoying RuntimeError I was running into when running slinalg.expm with zero values.


theano/tensor/slinalg.py
```diff
@@ -413,7 +413,12 @@ def perform(self, node, inputs, outputs):
         U = scipy.linalg.inv(V).T
 
         exp_w = numpy.exp(w)
-        X = numpy.subtract.outer(exp_w, exp_w) / numpy.subtract.outer(w, w)
+        exp_diff = numpy.subtract.outer(exp_w, exp_w)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", RuntimeWarning)
+            X = exp_diff/numpy.subtract.outer(w, w)
+        X[exp_diff == 0] = 0
```
nouiz: This line change the results. Can you add a test that cover this case? It will make sure it work well and don't get broken in the futur.

thanks

nouiz:
There is too much commit in this PR. Can you rebase this PR on top of Theano master to get rid of the other commit already in master? Otherwise, I can't see what you changed in Theano.

Otherwise, you can start a new branch on master and cherry-pick your commit.

thanks

computer-whisperer:
Fixed -- sorry about that.
I'll see if I can put together a test for this.

computer-whisperer:
If these tests pass, I will rebase these into one commit.

computer-whisperer:
Ok, I added a test and got it to validate. Ready to merge?

lamblin:
I'd be ready to merge, but should we wait for after the release?

nouiz:
I don't like changing behavior between an rc and a final release. One reason to change the behavior is if you have sure this is good change and that this fix a bug/problem. As this convert NaN to 0 and this is inside what is generated by theano.grad and people would exect 0 and non NaN to be returned, I think we can merge it. But as I didn't check what is expm and its grad, I let the final call to merge or not to @lamblin . If you think what I wrote is good and we should merge, you can merge.

theano/tensor/slinalg.py
```diff
@@ -413,7 +413,12 @@ def perform(self, node, inputs, outputs):
         U = scipy.linalg.inv(V).T
 
         exp_w = numpy.exp(w)
-        X = numpy.subtract.outer(exp_w, exp_w) / numpy.subtract.outer(w, w)
+        exp_diff = numpy.subtract.outer(exp_w, exp_w)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", RuntimeWarning)
+            X = exp_diff / numpy.subtract.outer(w, w)
+        X[exp_diff == 0] = 1
```
lamblin: I took a closer look, and I'm not actually sure this is correct.
If we have, for instance, `w[0] == -41` and `w[1] == -42`, then `exp_diff[0, 1] == 0`, but `subtract.outer(w, w)[0, 1] != 0`. Should `X[0, 1]` really be 1 then? Or should we do this instead?

```
diff = numpy.subtract.outer(w, w)
with warnings...:
    X = exp_diff / diff
X[diff == 0] = 1
```

tvercaut: I am not a Theano user and stumbled on this when looking for an implementation of the derivative of the exponential.

The division by zero arises in the computation of
```
(np.exp(x)-np.exp(y))/(x-y)
```
When `x` is very close to `y` we can use the Taylor expansion, e.g.
```
delta = x-y
np.exp(x)*(1 + delta/2 + delta*delta/6)
```
This removes the need for imposing `X[exp_diff == 0] = 1`

Overall, my current changes look like this
```
w, V = scipy.linalg.eig(A, right=True)
U = scipy.linalg.inv(V).T

def myloewner(x,y):
    delta = x-y
    if( abs(delta) < 1e-6 ):
        return np.exp(x)*(1 + delta/2 + delta*delta/6)
    return (np.exp(x)-np.exp(y))/(x-y)

myufunc = np.frompyfunc(myloewner, 2, 1)
ratio = myufunc.outer(w,w).astype(w.dtype)
Y = np.real( U.dot( np.multiply( V.T.dot(gA).dot(U), ratio ) ).dot(V.T) )
```

That being said, it looks like this function is just replicating the functionalities of [`scipy.linalg.expm_frechet`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.expm_frechet.html). Is this intentional to improve the speed or accuracy? Note that `scipy.linalg.expm_frechet` should also work for matrices that are not diagonalisable.

The reason I investigated this piece of code was because I wanted to have the complete derivative and not only the Fréchet derivative. This code allows to compute it fairly easily by using
```
np.kron(V, U).dot( np.diag(ratio.ravel()) ).dot( np.kron(U.T,V.T) ) 
```
but again this will only support matrices that are diagonalisable. If anyone has a good implementation of the complete derivative, I'd be quite interested. The following paper is a nice reference

> Najfeld and T. F. Havel, “Derivatives of the matrix exponential and their computation,” Advances in Applied Mathematics, vol. 16, no. 3, 1995. [Online]. Available: http: //dx.doi.org/10.1006/aama.1995.1017

but my implementation of it doesn't seem to be working very well (copied here for archiving purposes)
```
expA = np.expm(A)
IkeA = np.kron(np.eye(3), expA)
adA = np.kron(np.eye(3), A) - np.kron(A.T, np.eye(3))
funadA = scipy.linalg.funm(adA, lambda x: (1-np.exp(-x))/x )
DexpA = np.dot(IkeA, funadA)
```

tvercaut: To answer my own previous question, similarly fixing the division by zero with a Taylor expansion seems to work for the complete derivative:
```
def my_func(x):
    out = np.divide( (1-np.exp(-x)), x )
    idx = (abs(x) < 1e-6)
    x2idx = np.square(x[idx])
    out[idx] = 1 - x[idx]/2 + x2idx/6
    return out

expA = np.expm(A)
IkeA = np.kron(np.eye(3), expA)
adA = np.kron(np.eye(3), A) - np.kron(A.T, np.eye(3))
funadA = scipy.linalg.funm(adA, my_func )
DexpA = np.dot(IkeA, funadA)
```

PR #3983, opened 2016-02-03
Added required __future__ imports

Sentient07:
fix #3900

PR #3999, opened 2016-02-05
extension of #3983

Sentient07:
gh-3983

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -21,7 +21,7 @@ import numpy
 from six.moves import xrange
 
 import theano
-from theano.compat import PY3, imap, izip
+from theano.compat import imap, izip
 from theano import gof, printing
 from theano.gof import (Op, utils, Variable, Constant, Type, Apply,
                         FunctionGraph)
@@ -604,12 +604,11 @@ class _scalar_py_operators:
     def __mul__(self, other):
         return mul(self, other)
 
-    if PY3:
-        def __truediv__(self, other):
-            return div_proxy(self, other)
-    else:
-        def __div__(self, other):
-            return div_proxy(self, other)
+    def __truediv__(self, other):
+        return div_proxy(self, other)
+
+    def __div__(self, other):
+        return div_proxy(self, other)
 
     def __floordiv__(self, other):
         return int_div(self, other)
--- a/theano/tensor/xlogx.py
+++ b/theano/tensor/xlogx.py
@@ -1,5 +1,5 @@
+from __future__ import division
 import numpy
-
 from theano.tensor.elemwise import Elemwise
 from theano import scalar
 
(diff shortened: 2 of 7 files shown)


theano/tensor/xlogx.py
```diff
@@ -1,5 +1,5 @@
+from __future__ import division
 import numpy
-
```
nouiz: keep the empty line. Pep8 tell us that we need to separate the import from out project of the others. In fact, there should be 3 import section import from python default library, import of 3th party python library, import from this library.

Sentient07: @nouiz Made that change and will keep this in mind. :+1:

theano/scalar/basic.py
```diff
@@ -604,12 +604,11 @@ def __sub__(self, other):
     def __mul__(self, other):
         return mul(self, other)
 
-    if PY3:
-        def __truediv__(self, other):
-            return div_proxy(self, other)
-    else:
-        def __div__(self, other):
-            return div_proxy(self, other)
```
lamblin: @nouiz Is that actually correct? What happens if someone calls `a / b` on int tensors, from outside of Theano, without having imported `division` from `__future__`?

Sentient07: @lamblin @nouiz , I had tried this from outside theano without importing ```__future__``` , 

`>>> c = theano.function([a,b], b/a)
>>> c(2,5)
array(2, dtype=int32)
>>> c = theano.function([a,b], b//a)
>>> c(3,5)
array(1, dtype=int32)
>>> c(3,6)
array(2, dtype=int32)
>>> c = theano.function([a,b], b/a)
>>> c(3,6)
array(2, dtype=int32)
`

nouiz: in tensor, \__div__ is always div_proxy. But \__truediv__ is always true_div. There is code that make truediv be div, but it is overrided later.
Without the futur import.
python 2, int / int give and int // int give int.
Python 3, int / int give float and int // int give int.
numpy 1.10 behave like python 2.

With futur import,
python 2 do as python 3
numpy int also do as python 3.

div_proxy always return int_div by default. This is a theano flag operation.

I think we should do as numpy 1.10, we do as python 2/3 depending of the version used and if futur is imported, we behave like python 3. 

It is possible that older version of numpy do differently.

Can you test that with this code in python 2, in a file that just import theano, we keep the python 2 semantic, but in a file where the futur import is there, we use the python 3 semantic?

Sentient07: @nouiz, I've to test ```//``` and ```/``` on a tensor with and without future in python2 and test if the one with future behaves as python3 ?

nouiz: I think we need to add 2 tests file:
- theano/scalar/tests/test_div_no_futur.py
- theano/scalar/tests/test_div_futur.py

They should be mostly the same, except that one do the import about futur. They should check that we do the correct choise for the give python that is currently running.

This way, we will be sure we do it correctly and that it don't get broken in the futur.

Sentient07: Yeah, sure. Will add the tests

Sentient07: @nouiz Added the tests.

Sentient07:
@nouiz ready for merge ? I'll fix those merge conflicts in #3983 by tonight

Sentient07:
@lamblin ping. I guess this is ready for merge ?

lamblin:
Sure, thanks.

Issue #3996, opened 2016-02-05
Warn if cuDNN is too new?


f0k:
Some Lasagne users get trapped by trying to use cuDNN v4 with Theano 0.7. Maybe Theano should issue a warning if cuDNN is of a later version than expected, just something like: "Your cuDNN version is newer than what your version of Theano was written for. If something does not work as expected, try upgrading Theano to the bleeding-edge version or downgrading cuDNN to version XYZ."

nouiz:
Can you point to a discussion where this caused problem? Just to see the type of error the user currently have.

nouiz:
Done: https://github.com/Theano/Theano/pull/4000

f0k:
> Can you point to a discussion where this caused problem?

Sorry, was away over the weekend. One example is this: https://groups.google.com/forum/#!topic/lasagne-users/CV2Gx8NZPQY
The user got something cryptic:
```
mod.cu(147): error: argument of type "void *" is incompatible with
parameter of type "cudnnConvolutionBwdDataAlgo_t"
mod.cu(148): error: argument of type "float *" is incompatible with
parameter of type "size_t"
mod.cu(148): error: too few arguments in function call
3 errors detected in the compilation of
"/tmp/tmpxft_000040a5_00000000-9_mod.cpp1.ii".
```

PR #4000, opened 2016-02-05
Cudnn version, print it, warn when too new, make mandatory in the new back-end.

nouiz:
This PR do a few things about the cudnn version.

- It finish gh-3901(so fix gh-3876), print the cudnn version when we init the back-end.
- It fix gh-3996 by warning user if there cudnn version is too recent
- It make cudnn mandatory in the new back-end.

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -591,7 +591,14 @@ def use(device,
                                  " this property")
 
             if config.print_active_device:
-                cnmem_enabled = "enabled" if config.lib.cnmem else "disabled"
+                if config.lib.cnmem:
+                    if config.lib.cnmem > 1:
+                        cnmem_enabled = "enabled with initial size: %d MB" % config.lib.cnmem
+                    else:
+                        cnmem = min(config.lib.cnmem, 0.98)
+                        cnmem_enabled = "enabled with initial size: %.2f%% of memory" % cnmem
+                else:
+                    cnmem_enabled = "disabled"
                 cudnn_version = "not available"
                 warn = None
                 try:
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -593,9 +593,15 @@ def use(device,
             if config.print_active_device:
                 cnmem_enabled = "enabled" if config.lib.cnmem else "disabled"
                 cudnn_version = "not available"
+                warn = None
                 try:
                     (hdr_v, runtime_v) = dnn_version()
-                    cudnn_version = "%i" % (runtime_v)
+                    cudnn_version = runtime_v
+                    # 4100 should not print warning with cudnn 4 final.
+                    if cudnn_version > 4100:
+                        warn = ("Your CuDNN version is more recent then Theano."
+                                " If you see problems, try updating Theano or"
+                                " downgrading CuDNN to version 4.")
                 except Exception, e:
                     pass
                 print("Using gpu device %d: %s (CNMeM is %s, CuDNN %s)" % (
@@ -604,6 +610,10 @@ def use(device,
                     cnmem_enabled,
                     cudnn_version,),
                       file=sys.stderr)
+                if warn:
+                    import warnings
+                    warnings.warn(warn)
+
             if device_properties(use.device_number)['regsPerBlock'] < 16384:
                 # We will try to use too much register per bloc at many places
                 # when there is only 8k register per multi-processor.
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -278,12 +278,12 @@ class GpuOp(theano.gof.Op):
     """
 
     def make_thunk(self, node, storage_map, compute_map, no_recycling):
-        if theano.sandbox.cuda.use.device_number is None:
-            theano.sandbox.cuda.use("gpu",
-                                    force=True,
-                                    default_to_move_computation_to_gpu=False,
-                                    move_shared_float32_to_gpu=False,
-                                    enable_cuda=False)
+        if use.device_number is None:
+            use("gpu",
+                force=True,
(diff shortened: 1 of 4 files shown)


nouiz:
For commit 0b8a67f, travis didn't ran, I don't know why.

theano/sandbox/gpuarray/__init__.py
```diff
@@ -64,8 +65,24 @@ def init_dev(dev, name=None):
...
+            try:
+                cudnn_version = dnn.version()
+                # 4100 should not print warning with cudnn 4 final.
+                if cudnn_version > 4100:
+                    warn = ("Your CuDNN version is more recent then Theano."
+                            " If you see problems, try updating Theano or"
+                            " downgrading CuDNN to version 4.")
+                cudnn_version = " (CuDNN version %s)" % cudnn_version
+            except Exception:
+                raise RuntimeError("CuDNN is mandatory with the GpuArray back-end.")
```
abergeron: That I disagree with.

lamblin: I'm OK with not supporting convolutions without cudnn in the new back-end for a while, but elemwise, gemm, and so on should still be able to work.

nouiz: The other option is to raise an error only when pooling or convolution is
used. If we don't, the user will get very slow execution. We can't expect
all user to see the warning and understand the consequence. Some software
generate too much output, a simple warning will get lost.

I don't like the idea to spent time making this selective error... We do
already enough stuff. Why do you want us to suport that case?

abergeron: We could also spend the one or two days required to pour CorrMM and then there would be a not-CPU fallback.

That would be the ideal situation.

lamblin: I would really like people being able to run, say, fully-connected nets or RNNs on 580s, for instance. Also, if OpenCL supports gets better, then requiring CuDNN for OpenCL would be a bit silly.
I agree with @abergeron, porting the CorrMM code to the gpuarray backend would be ideal.

theano/sandbox/cuda/__init__.py
```diff
@@ -452,8 +592,28 @@ def use(device,
...
-                print("Using gpu device %d: %s (CNMeM is %s)" % (
-                        active_device_number(), active_device_name(), cnmem_enabled), file=sys.stderr)
+                cudnn_version = "not available"
+                warn = None
+                try:
+                    (hdr_v, runtime_v) = dnn_version()
+                    cudnn_version = runtime_v
+                    # 4100 should not print warning with cudnn 4 final.
+                    if cudnn_version > 4100:
+                        warn = ("Your CuDNN version is more recent then Theano."
```
f0k: *than

theano/sandbox/cuda/__init__.py
```diff
@@ -452,8 +592,28 @@ def use(device,
...
+                                " If you see problems, try updating Theano or"
+                                " downgrading CuDNN to version 4.")
+                except Exception, e:
+                    pass
+                print("Using gpu device %d: %s (CNMeM is %s, CuDNN %s)" % (
+                    active_device_number(),
+                    active_device_name(),
+                    cnmem_enabled,
+                    cudnn_version,),
+                      file=sys.stderr)
```
f0k: I think it would be useful to mention how much memory CNMeM is configured to claim, at least when it's enabled by default (it still isn't, right?). Otherwise people will wonder why their simple model already takes so much GPU memory in Theano compared to other libraries. Something like:
`Using gpu device 0: Geforce GT 640 (CNMeM enabled: 45% of memory; CuDNN 4004)`
(If not in this PR, I can file an issue.)

nouiz: Good idea, I make something similar in this PR.

theano/sandbox/cuda/__init__.py
nouiz: ok, fixed.

nouiz:
I now raise an error if convolution of pooling or logsoftmax are used in the new back-end and cudnn isn't there.

I also rebased as it was needed.

theano/sandbox/gpuarray/__init__.py
```diff
@@ -64,8 +65,25 @@ def init_dev(dev, name=None):
...
     if config.print_active_device:
-        print("Mapped name %s to device %s: %s" % (name, dev, context.devname),
+        warn = None
+        cudnn_version = ""
+        if dev.startswith('cuda'):
+            cudnn_version = " (CuDNN not available)"
+            try:
+                cudnn_version = dnn.version()
+                # 4100 should not print warning with cudnn 4 final.
+                if cudnn_version > 4100:
```
abergeron: How about `>= 5000`?  But otherwise I agree that 4100 is probably safe if NVIDIA follows previous numbering standards.

theano/sandbox/gpuarray/dnn.py
```diff
@@ -32,6 +32,10 @@
 from .opt_util import alpha_merge, output_merge, inplace_allocempty
 
 
+def raise_no_cudnn(msg="CuDNN is required for convolution and pooling"):
+    raise RuntimeError(msg)
```
abergeron: Isn't that caught by the optimizer safeguard?

abergeron:
Assuming that my comments are not an issue, I'm ok with this.

theano/sandbox/gpuarray/__init__.py
nouiz: I did that just in case they do a 4.1

theano/sandbox/gpuarray/dnn.py
nouiz: No, I tested it and it make the DLT convolution crash with the error.

nouiz:
As the comments don't request change and @abergeron agree with this PR, I merge.

PR #4005, opened 2016-02-05
Add tp_repr to CudaNdarray. 

nouiz:
This make DebugMode error print the value instead of <CudaNdarray object at 0x7f229c2f6eb0>

It will help debug the daily buildbot error with DebugMode.

### DIFF
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -2255,8 +2255,8 @@ class _Linker(gof.link.LocalLinker):
                     for r in node.outputs:
                         if r not in r_vals:
                             idx = order.index(node)
-                            assert thunks_py[idx] is None
-                            assert thunks_c[idx] is None
+                            assert thunks_py[idx] is None, node
+                            assert thunks_c[idx] is None, node
                             raise Exception("No code run for %s" % node)
 
                 if False:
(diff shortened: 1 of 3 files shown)


lamblin:
- Does that mean that we can now get rid of the special case for representing CudaNdarrays in https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/var.py#L48 ?
- Also, is it already done for GpuArray, or should we add that as well?

lamblin:
I'm OK with the change, and I'd be ready to merge if you can confirm that the GPU-related tests pass when run locally.

nouiz:
1) no, this is not related. What you point to print the test value. We could remove the call to numpy.asarray in that method, but we need to keep the method.

Note, that this make "print a", str(a) and repr(a) work for CudaNdarray.

I checked and GpuArray don't have it work. This need to be done in libgpuarray.

I'm starting the tests. I will see the result only Monday.

nouiz:
@abergeron I'll let you do the change for GpuArray. This isn't urgent, so it can wait. You know more cython then me.

lamblin:
@nouiz If we remove the call to `numpy.asarray`, then the implementation becomes identical to the one in the parent class (https://github.com/Theano/Theano/blob/master/theano/gof/graph.py#L409-L416). So we can remove it and the inherited version will be called, right?

nouiz:
right. I updated the PR.

PR #2059, opened 2014-08-27
initial working version of SolveCholesky and MatrixInverseCholesky

karlnapf:
Hi

Here is a working implementation of a Cholesky based linear solver and a matrix inversion based on a Cholesky factor. This allow for effectively precomputing the solutions of linear systems (with psd matrix) without the need of ever storing a matrix inverse.

Currently, there is no gradient for the Cholesky inversion. While writing tests for my implementation, I discovered a possible bug in gradients of matrix inversions, where the matrix is computed from a Cholesky, see #2066 

I included a couple of unit tests. I am already using this code in a research project of mine and it works fine. It is much faster than always re-computing the linear solve, and storing the inverted matrix is not numerically stable in my example.

Let me know what you think, and what is needed to merge this.

PR #4006, opened 2016-02-05
GPU Cholesky decomposition op

matt-graham:
## Reason for PR

There is currently no GPU Cholesky decomposition op to correspond to the `scipy.linalg` based `slingalg.Cholesky` op, and no existing PR implementing one as far as I can see (though an implementation of one might be useful for #2059). 

Cholesky decompositions are among other things useful in allowing more numerically stable and efficient solving of positive definite matrix systems (particularly if solving multiple systems sequentially as it allows an initial expensive decomposition to be reused without the normal numerical issues involved in precomputing a direct matrix inverse) and for use in generating samples from a multivariate Gaussian with a specified covariance matrix. Both of these uses crop up quite a lot in machine learning and statistics problems, and the O(D<sup>3</sup>) cost of computing the Cholesky decomposition means that such operations can often be a dominant cost in an algorithm so having a GPU accelerated operation can potentially give big gains.

This PR adds a CULA based GPU Cholesky decomposition op `GpuCholesky` which complements the exisiting CULA based `GpuSolve` op. 

For consistency with the existing `slinalg.Cholesky` op this returns a matrix in which only the entries below or above the diagonal (depending on value of `lower` parameter) are non-zero. This involves some further calls to `pycuda` based functions in `skcuda.linalg` for setting the above / below diagonal elements to zero as the CULA Cholesky decomposition function `culaDeviceSpotrf` acts destructively on the matrix it is decomposing and assigns the decomposition to the lower or upper triangular elements while leaving the existing values in the other elements in place (this is also the case for other LAPACK `*potrf` based Cholesky implementations e.g. `scipy.linalg.cho_factor`).

As well as the `GpuCholesky` op a corresponding `local_gpu_cholesky` optimization has been added which replaces `slinalg.Cholesky` op instances with `GpuCholesky` were possible. Unit tests for both the new op and optimization have also been added.

## Questions

  * The op currently creates a copy of the input matrix before performing the decomposition due to the destructive behaviour of the CULA `spotrf` function described above. In some cases maybe the original matrix does not need to be retained so an inplace / destructive operation would be permissible.   Would it be worth implementing this? Currently the corresponding `slinalg.Cholesky` op has a flag for this but it seems to not be implemented. If it was worthwhile adding this, are there any other requirements than defining a `destroy_map` attribute for the op as described [here](http://deeplearning.net/software/theano/extending/inplace.html) to make this work correctly (as well as obviously implementing the logic to do the operation in place)?

### DIFF
--- a/theano/sandbox/cuda/cula.py
+++ b/theano/sandbox/cuda/cula.py
@@ -191,8 +191,8 @@ class GpuCholesky(GpuOp):
             cula.culaInitialize()
             cula_initialized = True
 
-        # import util functions here to avoid circular import errors
-        from theano.misc.pycuda_utils import to_gpuarray, to_cudandarray
+        # import util function here to avoid circular import errors
+        from theano.misc.pycuda_utils import to_gpuarray
 
         inputs = [storage_map[v] for v in node.inputs]
         outputs = [storage_map[v] for v in node.outputs]
@@ -257,18 +257,18 @@ class GpuCholesky(GpuOp):
             # function to convert CudaNdArray -> GPUArray, applying this
             # before any transpose operation to ensure A_cpy buffer is still
             # C-contiguous as so can be shared with GPUArray object without
-            # any copy required.
+            # any copy required this meaning triu/tril operations occur in
+            # place on A_cpy buffer.
             if self.lower:
                 # extract only upper triangle in C-ordering i.e. lower triangle
                 # in F-ordering
-                A_copy_tri = linalg.triu(to_gpuarray(A_cpy), overwrite=True)
+                linalg.triu(to_gpuarray(A_cpy), overwrite=True)
             else:
                 # extract only lower triangle in C-ordering i.e. upper triangle
                 # in F-ordering
-                A_copy_tri = linalg.tril(to_gpuarray(A_cpy), overwrite=True)
-            # Convert back to CudaNdArray and assign output as transposed array
-            # to move from F to C ordering.
-            A_chol[0] = to_cudandarray(A_copy_tri).T
+                linalg.tril(to_gpuarray(A_cpy), overwrite=True)
+            # Assign output as transposed array to move from F to C ordering.
+            A_chol[0] = dimshuffle(A_cpy, (1, 0))
 
         thunk.inputs = inputs
         thunk.outputs = outputs
--- a/theano/sandbox/cuda/cula.py
+++ b/theano/sandbox/cuda/cula.py
@@ -4,6 +4,7 @@ import theano
 from theano.sandbox.cuda.type import CudaNdarrayType
 from theano.sandbox.cuda import GpuOp
 from theano.sandbox.cuda.basic_ops import as_cuda_ndarray_variable
+from theano.misc.pycuda_init import pycuda_available
 
 try:
     from theano.sandbox.cuda import cuda_ndarray
@@ -185,6 +186,10 @@ class GpuCholesky(GpuOp):
 
         if not cula_available:
             raise RuntimeError('Cula is not available and '
+                               'GpuCholesky Op can not be constructed.')
+
+        if not pycuda_available:
+            raise RuntimeError('pycuda is not available and '
                                'GpuCholesky Op can not be constructed.')
 
         if not cula_initialized:
--- a/theano/sandbox/cuda/cula.py
+++ b/theano/sandbox/cuda/cula.py
@@ -4,7 +4,6 @@ import theano
 from theano.sandbox.cuda.type import CudaNdarrayType
 from theano.sandbox.cuda import GpuOp
 from theano.sandbox.cuda.basic_ops import as_cuda_ndarray_variable
-from theano.misc.pycuda_utils import to_gpuarray, to_cudandarray
 
 try:
     from theano.sandbox.cuda import cuda_ndarray
@@ -191,6 +190,9 @@ class GpuCholesky(GpuOp):
         if not cula_initialized:
             cula.culaInitialize()
             cula_initialized = True
+
+        # import util functions here to avoid circular import errors
+        from theano.misc.pycuda_utils import to_gpuarray, to_cudandarray
 
         inputs = [storage_map[v] for v in node.inputs]
         outputs = [storage_map[v] for v in node.outputs]
--- a/theano/sandbox/cuda/cula.py
+++ b/theano/sandbox/cuda/cula.py
@@ -140,3 +140,100 @@ class GpuSolve(GpuOp):
         return thunk
 
 gpu_solve = GpuSolve()
+
+
+class GpuCholesky(GpuOp):
+    """
+    CULA GPU Choleksy factorisation op.
+
+    Given a real positive definite matrix `A` returns either a lower
+    triangular matrix `L` such that `A == dot(L, L.T)` if `lower == True`
+    else returns an upper triangular matrix `U` such that `A == dot(U.T, U)`
+    if `lower == False`.
+
+    Parameters
+    ----------
+    lower
+        Whether to return a lower rather than upper triangular decomposition.
+
+    """
+
+    __props__ = ('lower',)
+
+    def __init__(self, lower=True):
+        self.lower = True
+        super(GpuCholesky, self).__init__()
+
+    def output_type(self, inp):
+        return CudaNdarrayType(broadcastable=[False] * inp.type.ndim)
+
+    def make_node(self, inp):
+        inp = as_cuda_ndarray_variable(inp)
+
+        assert inp.ndim == 2
+        return theano.Apply(self, [inp], [self.output_type(inp)()])
+
+    def make_thunk(self,
+                   node,
+                   storage_map, _,
+                   no_recycling=[]):
+
+        # Initialize CULA the first time it is needed
+        global cula_initialized
+
+        if not cula_available:
+            raise RuntimeError('Cula is not available and '
+                               'GpuCholesky Op can not be constructed.')
+
+        if not cula_initialized:
+            cula.culaInitialize()
+            cula_initialized = True
+
+        inputs = [storage_map[v] for v in node.inputs]
+        outputs = [storage_map[v] for v in node.outputs]
+
+        def thunk():
+            # size of the matrices to invert
+            z = outputs[0]
+
+            # Matrix
+            A = inputs[0][0]
+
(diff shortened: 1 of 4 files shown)


nouiz:
The inplace version would be useful to remove the copy. So it would speed it up a bit. how much depend of the rest of the code.

This need the parameter and small code change, but this also need an optimization that would convert the non-inplace version of the op to the inplace version of the op.

theano/sandbox/cuda/opt.py
```diff
@@ -710,6 +710,34 @@ def local_gpu_solve(node):
...
+        if (host_input.owner and
+            isinstance(host_input.owner.op,
+                       slinalg.Cholesky)):
+            A = host_input.owner.inputs[0]
+            lower = host_input.owner.op.lower
+            return [gpu_cholesky(as_cuda_ndarray_variable(A), lower)]
+
+    if isinstance(node.op, slinalg.Cholesky):
+        if any([i.owner and isinstance(i.owner.op, HostFromGpu)
+                for i in node.inputs]):
```
nouiz: As the op always have only 1 input, you can remove the list comprehension and do the check only on node.inputs[0]. This make the code more clear.

theano/sandbox/cuda/cula.py
```diff
@@ -140,3 +143,142 @@ def cula_gpu_solve(A_, b_, trans='T'):
...
+        super(GpuCholesky, self).__init__()
+
+    def output_type(self, inp):
+        return CudaNdarrayType(broadcastable=[False] * inp.type.ndim)
+
+    def make_node(self, inp):
+        inp = as_cuda_ndarray_variable(inp)
+
+        assert inp.ndim == 2
+        return theano.Apply(self, [inp], [self.output_type(inp)()])
```
nouiz: I would remove the output_type method and use use inp.type() for the output. So replace [self.output...] by [inp.type()]

theano/sandbox/cuda/cula.py
```diff
@@ -140,3 +143,142 @@ def cula_gpu_solve(A_, b_, trans='T'):
...
+            A = inputs[0][0]
+
+            # Cholesky decomposition of matrix is first (and only) output
+            A_chol = outputs[0]
+
+            # CULA Cholesky function is destructive as it assigns calculated
+            # decomposition to lower / upper triangle of input array therefore
+            # force copy of array with allocation of a new C-contiguous buffer.
+            # CULA operation assumes Fortran ordering however input matrix must
+            # be symmetric therefore no need to alter order.
```
nouiz: Why inputs must be symmetric? I don't think this is right. I could easily add a parameter order that default to C, but could be F as numpy. This would make the copy a fortran order version.

nouiz:
Other then my comments, all looks pretty good. thanks.

theano/sandbox/cuda/cula.py
matt-graham: The Choleksy decomposition is only defined for positive definite matrices, which for real matrices are a subset of the symmetric matrices. If a non-symmetric and so non-positive-definite input is passed the underlying CULA function will raise an exception as the Cholesky decomposition algorithm will fail. Given an exception will be raised anyway without any additional checking overhead for non-symmetric inputs it didn't seem worth adding any explicit checks for symmetry. 

I guess the comment is not strictly correct however so I'll edit to make it more clear - perhaps it would be better to say any valid input will be symmetric? It is possible a non-symmetric matrix input will be passed to the op however it does not matter whether it uses C or Fortan ordering here as in either case the matrix will remain non-symmetric and end up raising an exception when it is attempted to calculate the decomposition. 

I'll also add an unit test to ensure non-symmetric inputs cause an exception to be raised in addition to the current one for a non-positive-definite but symmetric input.

theano/sandbox/cuda/cula.py
matt-graham: Thanks, will change, I hadn't fully understood what this code was doing.

theano/sandbox/cuda/opt.py
matt-graham: Cheers will update, bad case of copying and pasting code from another op without really considering what it was doing!

matt-graham:
With regards to the inplace version:

  * The `skcuda.cula` `culaDeviceSpotrf` and the `skcuda.linalg` `triu` and `tril` functions both require / assume contiguous input arrays I believe. If I use `basic_ops.gpu_contiguous` to ensure a contiguous copy is made if necessary when `inplace = True` will this cause some sort of inconsistency as the op in no longer actually in place? Or alternatively is there anyway to force no inplace op to be used in these settings? 

  * As there is also a `lower` property for the op, (how) should I define op instances? For example should I just add one for each of the possible property values e.g `gpu_cholesky_lower_no_inplace` etc. - it seems this would get a bit unmanageable for a large number of properties however I assume there is some overhead to creating instances on the fly using e.g. `GpuCholesky(lower=True, inplace=False)`? And would it make sense therefore to define an in place optimisation for each of the different `lower` values? For now I have defined instances for each `lower` / `inplace` combination and a separate inplace optimisation for each `lower` value.

  * Is there a standard way to define tests on inplace ops? My attempts caused a series of `TypeError("Graph must not contain inplace operations")` exceptions to be raised so I have removed them for now.

abergeron:
If you wrap your input in gpu_contiguous, it will not cause problems for inplace since that is just another op that will execute before yours.

You don't need to define op instances.  Even if lot of the ops do, it's just for quick access.  If it doesn't make sense for yours, then don't do it.  There is a minimal overhead to creating new instances, but it should be trivial and is also not paid during execution.

You can pass accept_inplace=True to function() for testing purposes.  This is not documented because we don't want people to use it in general as you may provide unsafe graphs with this.

A better way would be to generate some small graph where the operation should be inplace and verify in the compiled graph that it is inplace.

matt-graham:
Now updated to remove op instance definitions and add in tests of inplace op and the associated optimization.

I defined a small graph to verify the inplace operation is applied successfully as suggested (in `tests/test_opt.py`) but also used `accept_inplace=True` to check the inplace operations output was consistent with Numpy in `tests/test_cula.py` as it seemed cleaner to test the validity of the operation implementation and optimization separately.

I wasn't sure quite how to handle registering the inplace optimization  - I added a new `optdb.register` call for CULA inplace ops rather than wrapping this in to the existing call registering the inplace BLAS ops. In particular I wasn't sure how to choose the priority so just used the same value (70.0) as for the BLAS ops.

theano/sandbox/cuda/opt.py
```diff
@@ -710,6 +711,35 @@ def local_gpu_solve(node):
...
+    CpuCholesky(host_from_gpu) -> host_from_gpu(GpuCholesky{no_inplace})
+
+    """
+    if isinstance(node.op, GpuFromHost):
+        host_input = node.inputs[0]
+        if (host_input.owner and
+            isinstance(host_input.owner.op,
+                       slinalg.Cholesky)):
+            A = host_input.owner.inputs[0]
+            A_ndarray = as_cuda_ndarray_variable(A)
```
nouiz: ndarray make me think it is on the CPU. What about using A_gpu?

theano/sandbox/cuda/opt.py
```diff
@@ -2032,6 +2062,16 @@ def local_inplace_ger(node):
     if node.op == gpu_ger_no_inplace:
         return [gpu_ger_inplace(*node.inputs)]
 
+
+@local_optimizer([GpuCholesky(lower=True, inplace=False),
+                  GpuCholesky(lower=False, inplace=False)], inplace=True)
+def local_inplace_gpu_cholesky(node):
+    if node.op == GpuCholesky(lower=True, inplace=False):
```
nouiz: The full code of this method can be simplified:

if isinstance(node.op, GpuCholesky) and not node.op.inplace:
    return [GpuCholesky(node.op.lower, inplace=True)]

Constructing object just to compare is time consuming and would make optimization slower. So this code will be simpler and faster.

theano/sandbox/cuda/tests/test_cula.py
```diff
@@ -67,3 +67,111 @@ def test_uni_rand_solve(self):
...
+        chol_A_res = numpy.array(res)
+        utt.assert_allclose(chol_A_res, chol_A_val)
+
+    def test_invalid_input_fail_non_square(self):
+        """ Invalid Cholesky input test with non-square matrix as input. """
+        def invalid_input_func():
+            A_val = numpy.random.normal(size=(3, 2)).astype("float32")
+            fn = self.get_gpu_cholesky_func(True, False)
+            fn(A_val)
+        self.assertRaises(ValueError, invalid_input_func)
```
nouiz: This can be simplified. You don't need to make a new method. So just do something like:

def test_...(self):
    A_val = ...
   fn = ...
   self.assertRaises(ValueError, fn, A_val).

idem for the other tests.

nouiz:
thanks PR seem to advance well. I just got a few comments. I killed some of travis run, as it would have taken it too much time to test all of them.

thanks.

matt-graham:
Some further updates in response to comments.

@nouiz : I wasn't sure what your comment

> You can just add GpuCholesky. This will work for all instance.

 was referring to? Is this in reference to the `tracks` argument being passed to the `local_optimizer` decorator which currently adds two separate `GpuCholesky` instances?

Apologies about the large amounts of commits and so Travis runs; I am using a remote GPU server to run the tests as I don't have any CUDA compatible devices on my development machine, so I've being using git to sync the code to the remote server. Is there a simple way for me to switch to manually updating the PR after I make a series of pushes to my fork branch rather than it tracking every new push? Sorry if this is a stupid question, my Google searches didn't turn up an obvious answer.

lamblin:
Regarding the PR updates, the simplest way I can think of would be for you to have two remote branches on GitHub: one being tracked by the PR, and one for synchronization between your machines, say `gpu_cholesky_sync`.
You can still have only one branch on your machines, but decide to push to one or the other, for instance `git push origin gpu_cholesky:gpu_cholesky_sync` if you want to push the local branch `gpu_cholesky` to remote branch `gpu_cholesky_sync`. When you want to update `gpu_cholesky` being tracked by the PR, then you can push to it.

matt-graham:
@lamblin Thanks for the advice, I've set up two separate branches as you suggest.

nouiz:
For your other question, yes, you understood correctly my comment.

thanks

matt-graham:
@nouiz Now updated the `local_optimizer` decorator.

PR #4007, opened 2016-02-06
GPU triangular solve op

matt-graham:
## Reason for PR

This in some way extends on from #3994 which makes some proposed changes to the general CULA based `GpuSolve` op with the aim of making it more compatible with the `slinalg.Solve` op. 

The current `slinalg.Solve` op has an `A_structure` property which can be used to provide information about special structure in the matrix system being solved. Currently as well as the normal general dense solver the `slinalg.Solve` op also uses a specific solver for triangular systems if `A_structure` is one of `lower_triangular` or `upper_triangular`. Having a GPU triangular solve op would make the GPU based functionality equivalent and potentially give significant performance gains over using a general solver when the system is known to be triangular given the O(D<sup>2</sup>) cost compared to O(D<sup>3</sup>) cost for a general solver (this is ignoring however possible differences in degree of parallelism and so GPU speed up).

Combined with #4006 this PR would also allow efficient and numerically stable GPU-based solving of positive definite linear systems via the Cholesky decomposition of the matrix.

A new CUBLAS based `GpuTriangularSolve` op is defined in a new module `cublas.py`, this mainly just a wrapper around the `skcuda.cublas.cublasStrsm` single-precision triangular linear system solver function and very closely following the structure of the existing CULA based `GpuSolve` op.

The `local_gpu_solve` optimization is also updated to replace `slinalg.Solve` ops in which the `A_structure` property is set to one of `lower_triangular` or `upper_triangular` with instances of the new `GpuTriangularSolve` op, with the current `GpuSolve` op used as the fallback for other `A_structure` values.

Tests for the new op and edited optimization are also defined.
## Questions
- Is there a more logical existing module for the `GpuTriangularSolve` op to exist in rather than creating a new `cublas.py` module? One possibility seems to be the existing [sandbox/cuda/blas.py](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/blas.py) however this currently only contains ops directly implemented with CUDA / CUBLAS C-code as opposed to this op which just wraps a `skcuda.cublas` function. Another option would be to combine with the op(s) in [sandbox/cuda/cula.py](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/cula.py) to group all the `skcuda` based ops together.
- I am not sure if the `atexit` based method of making sure the `cublas.cublasDestroy` function is called on any initialized CUBLAS handle before exiting is the best approach?

### DIFF
--- a/theano/sandbox/cuda/cublas.py
+++ b/theano/sandbox/cuda/cublas.py
@@ -155,12 +155,14 @@ class GpuTriangularSolve(GpuOp):
                 A_ptr = A_.gpudata
                 b_ptr = b_.gpudata
 
-                alpha = 1.0  # unit scalar used for multiplicatio
-                side = 'l'  # indicates matrix A is on right of B
-                uplo = 'l' if lower else 'u'  # set whether upper or lower
-                                              # part of matrix A stored
-                diag = 'n'  # indicates elements on diagonal of matrix A may
-                            # not be unity
+                # unit scalar used for multiplication
+                alpha = 1.0
+                # indicates matrix A is on left of B
+                side = 'l'
+                # set whether upper or lower part of matrix A stored
+                uplo = 'l' if lower else 'u'
+                # indicates elements on diagonal of matrix A may not be unity
+                diag = 'n'
 
                 cublas.cublasStrsm(cublas_handle, side, uplo, trans, diag,
                                    n, m, alpha, A_ptr, lda, b_ptr, ldb)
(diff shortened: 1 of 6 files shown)


abergeron:
If you only use cublas (and not CULA) as seems to be the case here.  You should write C code instead of going through the scikits-cuda binding.

If you don't feel comfortable with C, then python would be fine, but it will be significantly slower.

abergeron:
This is for the old backend.  A port of this for the new backend is at #6182

PR #4009, opened 2016-02-06
Exponential Linear Units (ELUs)

AnishShah:


theano/tensor/nnet/nnet.py
```diff
@@ -2335,3 +2335,30 @@ class of the input x while the second predicts the output of the input x in
...
+    symbolic tensor
+        Element-wise exponential linear activation function applied to `x`.
+
+    References
+    -----
+    .. [1] Djork-Arne Clevert,  Thomas Unterthiner, Sepp Hochreiter
+        "Fast and Accurate Deep Network Learning by
+        Exponential Linear Units (ELUs)" <http://arxiv.org/abs/1511.07289>`.
+    """
+    return alpha * (tensor.exp(0.5 * (x - abs(x))) - 1) + 0.5 * (x + abs(x))
```
nouiz: The formula here seem more complicated then the one used in lasagne:
https://github.com/Lasagne/Lasagne/blob/master/lasagne/nonlinearities.py#L238

Is there a reason? Do you have an idea why they didn't included alpha? Can you confirm they are equivalent when alpha is 1 and if so, make an in here and use there simpler code for that special case? I think it will generate better code.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1617,3 +1618,19 @@ def test_h_softmax():
...
+    X = rng.randn(20, 30).astype(config.floatX)
+
+    # test the base case, without custom alpha value
+    y = elu(x).eval({x: X})
+    assert numpy.allclose(y, numpy.maximum(X, 0) + (numpy.exp(numpy.minimum(X, 0))-1))
+
+    # test for different constant alpha values
+    for alpha in 0, 1.5, 2, -1, -1.5, -2:
+        y = elu(x, alpha).eval({x: X})
+        assert numpy.allclose(y, numpy.maximum(X, 0) + alpha * (numpy.exp(numpy.minimum(X, 0))-1))
```
nouiz: use

utt.assert_allclose(y, ...)

this will generate an assert with a better error message in case of problems.

theano/tensor/nnet/nnet.py
nouiz: Is there a reason, why you didn't used the same algo as the version in the test in numpy?

thanks for the PR. I finished this round of review.

AnishShah: > Do you have an idea why they didn't included alpha? 

The formula in the research paper includes alpha (alpha is used to saturate negative net inputs). But, all the experiments mentioned in the paper used alpha = 1.  

> Is there a reason, why you didn't used the same algo as the version in the test in numpy?

I used `tensor.exp` because while reading the source code, I came across functions using `tensor.exp`. So, I didn't use numpy. 
But, I will change this formula to a simpler one which is in Lasagne, as you said.

AnishShah: @nouiz Also, if you look at the tests, they have `alpha` var. https://github.com/Lasagne/Lasagne/blob/master/lasagne/tests/test_nonlinearities.py#L19

AnishShah:
@nouiz I have updated the PR. Can you please review this? My comments got minimized, because of my changes.

nouiz:
I'm killing travis as it won't help and it is overloaded.

AnishShah:
@nouiz I've updated the PR

carriepl:
Looks good to me so I'm merging. Also, after IRL discussion with @nouiz, he says he also thinks this is good for merge.
Thanks a lot for the pull request.

Issue #3573, opened 2015-10-28
Assert to never reintroduce Apply node in fgraph.


nouiz:
Some optimization call replace_validate_remove.

I think we should tag the node to remove with a tag like node.tag.never_reuse=True. Then in FunctionGraph, when we add an apply node in the graph, we check that it don't have that tag.

That way we can be sure to not reintroduce by mistake some node.

Should we also do that for local optimizer on the original node if after substitution, it isn't in the fgraph anymore?

carriepl:
How costly would it be to do it every time a local optimizer requests a substitution?
Even if it ends up being too costly to do it for local optimizers, it would probably still have benefits to do it only in the cases or replace_validate_remove()

nouiz:
What about when we remove a node from fgraph, we alsays tag it as not to
reintroduce, until we start to do inplace optimization? Before inplace opt,
we shouldn't revert change in the graph I think. After that, some revert
could reintroduce it. Could we allow to reintroduce only for revert?

Sentient07:
@nouiz , is this ticket open, shall i take it ?

nouiz:
It is open. It need to investigate in the core of Theano. It isn't very well documented, so not the most trivial. But if you are able to investiage unknow code, you can take it.

Mostly, go in the file theano/gof/fg.py, then find __remove_clients__

Sentient07:
@nouiz So, i'd need to create a new variable in the class ```FunctionGraph```(probably a list) that contains the information of all the removed nodes (from ```__remove_clients__``` method) and prevent them from getting added to the graph from inside the ```__import__``` method, right ?

Sentient07:
@nouiz Also, according to your initial comment, I should use a variable ```never_reuse``` to check if apply nodes could be added back to graph, right ?

Sentient07:
@nouiz Ping

nouiz:
I think this is good. It is possible that this is too strict and cause
problems. I don't think so, but I just want to let you know about this. If
this end up being too strict this can be done in a less strict way.

thanks

Sentient07:
@nouiz , I've a question. While adding back ```apply_nodes``` that has once been removed, do i need to add the output and input of that node to the variable ? (This is a slight confusion I'm having as to how nodes and variables go in hand..)
P.S: This issue seems challenging and interesting! :)

nouiz:
This is in the core of Theano, so it is normal that this isn't as clear as
the user interface. I suggest that you revisit this page that help
[...]

PR #4013, opened 2016-02-07
Assert to never reintroduce Apply node in fgraph

Sentient07:
fix for #3573

### DIFF
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -153,6 +153,7 @@ class FunctionGraph(utils.object2):
 
         self.inputs = list(inputs)
         self.outputs = outputs
+        self._removed_nodes = set()
 
         for f in features:
             self.attach_feature(f)
@@ -320,13 +321,15 @@ class FunctionGraph(utils.object2):
                               if output.clients or output in self.outputs]
             # If the apply node is not used and is not an output
             if not used_or_output:
-                self.apply_nodes.remove(apply_node)
-                self.variables.difference_update(apply_node.outputs)
-                self.execute_callbacks('on_prune', apply_node, reason)
-
-                for i, input in enumerate(apply_node.inputs):
-                    self.__remove_clients__(input, [(apply_node, i)],
-                                            reason=reason)
+                if apply_node in self.apply_nodes:
+                    #keeping track of removed apply node
+                    self.apply_nodes.remove(apply_node)
+                    self.variables.difference_update(apply_node.outputs)
+                    self.execute_callbacks('on_prune', apply_node, reason)
+                    self._removed_nodes.add(apply_node)
+                    for i, input in enumerate(apply_node.inputs):
+                        self.__remove_clients__(input, [(apply_node, i)],
+                                                reason=reason)
         # variable should not have any clients.
         # assert not variable.clients
 
@@ -478,8 +481,14 @@ class FunctionGraph(utils.object2):
 
         for node in new_nodes:
             assert node not in self.apply_nodes
-            self.__setup_node__(node)
-            self.apply_nodes.add(node)
+            prevent_addition = False
+            for n in self._removed_nodes :
+                if node is n :
+                    prevent_addition = True
+                    
+            if not prevent_addition : 
+                self.__setup_node__(node)
+                self.apply_nodes.add(node)
             for output in node.outputs:
                 self.__setup_r__(output)
                 self.variables.add(output)
@@ -488,8 +497,9 @@ class FunctionGraph(utils.object2):
                     self.__setup_r__(input)
                     self.variables.add(input)
                 self.__add_clients__(input, [(node, i)])
-            assert node.fgraph is self
-            self.execute_callbacks('on_import', node, reason)
+            if not prevent_addition :
+                assert node.fgraph is self
+                self.execute_callbacks('on_import', node, reason)
 
     # change input #
     def change_input(self, node, i, new_r, reason=None):
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -481,14 +481,13 @@ class FunctionGraph(utils.object2):
 
         for node in new_nodes:
             assert node not in self.apply_nodes
-            prevent_addition = False
-            for n in self._removed_nodes :
-                if node is n :
-                    prevent_addition = True
+            prevent_addition = node in self._removed_nodes
                     
-            if not prevent_addition : 
-                self.__setup_node__(node)
-                self.apply_nodes.add(node)
+            if prevent_addition : 
+                raise InconsistencyError
+                ("Trying to reintroduce an old nodes in the graph. This should not happen")
+            self.__setup_node__(node)
+            self.apply_nodes.add(node)
             for output in node.outputs:
                 self.__setup_r__(output)
                 self.variables.add(output)
@@ -497,9 +496,8 @@ class FunctionGraph(utils.object2):
                     self.__setup_r__(input)
                     self.variables.add(input)
                 self.__add_clients__(input, [(node, i)])
-            if not prevent_addition :
-                assert node.fgraph is self
-                self.execute_callbacks('on_import', node, reason)
+            assert node.fgraph is self
+            self.execute_callbacks('on_import', node, reason)
 
     # change input #
     def change_input(self, node, i, new_r, reason=None):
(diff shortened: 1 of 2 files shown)


Sentient07:
@nouiz there are 7 Failures, which are raised from ```__add_clients__``` method, because of intersection of existing clients and new incoming clients. I'm not understanding this failure. Could it be because tests are trying to ```replace``` nodes in the graph ?

nouiz:
Thanks for the PR, I'm looking at it now. Can you try to give a better title? It allow us to more easily see what is going in this PR. We don't recall all issue number and what they mean. I changed this PR title.

thanks.

theano/gof/fg.py
```diff
@@ -478,8 +481,14 @@ def find_path_to(output_var, input_var):
 
         for node in new_nodes:
             assert node not in self.apply_nodes
-            self.__setup_node__(node)
-            self.apply_nodes.add(node)
+            prevent_addition = False
+            for n in self._removed_nodes :
+                if node is n :
+                    prevent_addition = True
```
nouiz: Why do a loop here? This is pretty not efficient. You can just do:
prevent_addition = node in self._removed_nodes

this will be faster as it won't loop over the old nodes. It will use the set code that use hashing for this. So the cost will be mostly constant vs having a cost that is linear in the number of old nodes.

theano/gof/fg.py
```diff
@@ -488,8 +497,9 @@ def find_path_to(output_var, input_var):
                     self.__setup_r__(input)
                     self.variables.add(input)
                 self.__add_clients__(input, [(node, i)])
-            assert node.fgraph is self
-            self.execute_callbacks('on_import', node, reason)
+            if not prevent_addition :
```
nouiz: I would remove this change and keep the old code.

theano/gof/fg.py
```diff
@@ -478,8 +481,14 @@ def find_path_to(output_var, input_var):
...
         for node in new_nodes:
             assert node not in self.apply_nodes
-            self.__setup_node__(node)
-            self.apply_nodes.add(node)
+            prevent_addition = False
+            for n in self._removed_nodes :
+                if node is n :
+                    prevent_addition = True
+                    
+            if not prevent_addition :
```
nouiz: Here I would keep the old code and if prevent_addiontion is true, I would raise an error of type InconsistencyError() with a messge like "Tring to reintroduce an old nodes in the graph. This should not happen"

I'm not sure what you are tring to do when we introduce an old node, but I think we should make the optimization that try that fail.

theano/gof/fg.py
```diff
@@ -320,13 +321,15 @@ def __remove_clients__(self, r, clients_to_remove,
                               if output.clients or output in self.outputs]
             # If the apply node is not used and is not an output
             if not used_or_output:
-                self.apply_nodes.remove(apply_node)
```
nouiz: Why did you added this if? The if above on line 322 should be enough. If your if fail, it mean there is a bug problem and we want to crash, the graph isn't in a good condition.

The other change you did here seem good.

nouiz:
I finished this round of review. If I wasn't clear, don't hesitate to ask questions.

thanks

theano/gof/fg.py
Sentient07: @nouiz I used ```is``` for comparing the exact instance of node object. AFAIK, ```in``` would do a value comparison and not the exact comparison.

nouiz: Apply class don't implement \__hash__ and \__eq__, so by default, a set will do the same instance comparison in python.

I just checked and this work with python 3.3 too.

Sentient07: Oh, sure. In that case i'd remove the loop and use ```in``` comparator.

theano/gof/fg.py
Sentient07: @nouiz , if the errors are raised when an added node is trying to get added, wouldn't that cause some of the existing tests to fail(I was a little skeptical about this) ? I'm trying to add the nodes back only if they've not been added before. That's why ```if``` statement there. I'll give try what you've said and get back to you on what's happening

theano/gof/fg.py
Sentient07: I'm trying to prevent the once removed ```apply_node``` from getting added back to the graph. Should I raise an error in this case ?

theano/gof/fg.py
Sentient07: @nouiz you mean this line is redundant, right ? Even i feel the same, not sure why added that ```if``` statement.
https://github.com/Theano/Theano/pull/4013/files#diff-8850ac78ebd71ffd0376fa6c58a80a69R324

Sentient07:
@nouiz Have replied wherever i've questions. Would make the suggested changes, and hopefully tests would pass :)

theano/gof/fg.py
nouiz: There is a checkpoint mechanism in the optimization. So if an optimization cause a problem, we revert to the state before and skip it. So yes, we can raise an error here. This is used for inplace optimization, so this is a used, tested and working system in Theano.

theano/gof/fg.py
nouiz: Not here, you need to raise the error above, in the part where you did the other change.

theano/gof/fg.py
nouiz: semantically redundant yes. Not trivial to guess without understanding what is going it.

Sentient07:
@nouiz made the changes you suggested, it's errors=3 and failures=6 now.

nouiz:
Due to the error we get, we can't do it like this except if we want to disable some currently allowed practice. I don't think it is a good timing for this and I'm not sure it is a good change to do.

So we need to fall back to the other implementation. So instead of disallowing all old node, we should only disallow nodes marked as to be always removed. There is the function replace_all_validate_remove, that have a remove parameter. So can you only add in the set those nodes? This can be done in that method.

To implement this cleanly, you need to start from scratch. In the class ReplaceValidate, add the set of nodes to remove. In this method replace_all_validate_remove, add the nodes to remove in that set. Then implement a method called on_import() (grep to know its signature). Then in the method check if the new node is in that set.

Sentient07:
@nouiz yeah, understood. How should i test this functionality ?

theano/gof/fg.py
Sentient07: @nouiz, sorry, I had missed out this comment while addressing the rest. Would make this along with the change that you've suggested now.

nouiz:
I'm not sure how to test this easily. The only way I think is to make optimizations that is bad and would generate this case. So unless you are motivated to make some complicated test, I think we can go without test for this and just code review.

If you want to test this, mostly, you need to make 2 optimizations. One that remove an op while calling replace_all_variables_remove and put the removed node in a list and have the other opt reintroduce this node.

Sentient07:
@nouiz , the tests are still failing. I've implemented what i've understood from your comment, could you please have a look at the PR and correct me where i was wrong?

nouiz:
It don't like that we raise the error during the call to replace. This need to be raised during validate or we should change that restriction.

I think the simplest would be to use a new self.fail_validate variable that is initialized to False. In on_import, make set it to True when we want to fail. Then in that class, add an validate() method (grep that file for an example). If fail_validate it True, set it back to False and then raise the error.

Sentient07:
@nouiz yeah, i've understood that. Didn't think of that validate method would help. Hope this is correct.

theano/gof/toolbox.py
```diff
@@ -273,6 +273,8 @@ def on_attach(self, fgraph):
             if hasattr(fgraph, attr):
                 raise AlreadyThere("ReplaceValidate feature is already present"
                                    " or in conflict with another plugin.")
+        self._nodes_removed = set()
```
nouiz: In the method on_detach, do "self._nodes_removed" to help make sure we loose refcount to some nodes. There is some cases where we keep to much of them, so this is just to help make sure we don't make those cases more troublesome.

After that I think the PR is ready to merge. Thanks

Sentient07: @nouiz , I have added that. Check and merge at your convenience. Thanks :smile:

nouiz:
@abergeron you can merge when travis pass.

Issue #3557, opened 2015-10-26
Follow-up and clean-up for abstract conv2d


lamblin:
This is a follow-up to the abstract conv2d op mentioned in #2505
Things that remain to be done after #2665:
- [x] Add online documentation about the abstract Op (#3779)
- [ ] Simplify the optimizations introducing GpuCorrMM or CuDNN Ops, by removing the special cases mapping one type of convolution operation to a different one (for instance, mapping gradW to forward) (edit: only when we are guessing the direction, not when it was deliberate for speed, at least for CorrMM)
- [x] Prepare transition from `theano.tensor.nnet.conv.conv2d` to `theano.tensor.nnet.abstract_conv.conv2d` as the default (#3675)
- [x] Port optimizations to the new GPU back-end (~~#3676~~ #3737)
- [x] When #3285 is merged, add optimizations to insert these Ops
- [x] Execute only a subset of tests by default, so the time is not too large (use a numpy RNG seeded with `unittest_tools.fetch_seed()`)
- [x] Update tests to check that the right implementation is inserted by the optimizations (#3668)
- [x] Update (some) tests to call `theano.tensor.nnet.conv2d` (#3732)
- [x] Centralize and unify the implementations of shape inference (#3582)
- [x] Find a way to make test values (and possibly DebugMode) work, but still not enable them in a compiled (optimized) function (#3598)
- [x] Better document "shape" arguments, add runtime checks (#4048)
- [ ] Check what to do about the meta opt?

JesseLivezey:
Would the optimization for the CPU CorrMM look like the ones here?
https://github.com/Theano/Theano/blob/5849db2111a57dbfe6511cc0a12708845aec5daf/theano/sandbox/cuda/opt.py#L2690

I should have time to add them in the next week.

lamblin:
Yes, exactly. You do not have to add the special cases that convert AbstractConv2d to gradI or gradW, in fact I'm arguing for their removal from the existing optimization.
They should simply check for TensorType instead of CudaNdarrayType.
Then, you can register them in a new `LocalGroupDB` along with the existing (legacy) convolution, and register that in `specialize_device` as well. We can give you a hand with that part if needed.
Thanks!

fvisin:
The output's shape in https://github.com/Theano/Theano/blob/3b3b2a4e1d4e932ddd15928d82881524b6f5da01/theano/tensor/nnet/abstract_conv2d.py#L59-L62 should also consider `subsample`

[...]

PR #4018, opened 2016-02-09
Move the AbstractConv tests with the implementation

abergeron:
This also tests the CPU implementations even if there is no GPU.

### DIFF
--- a/theano/tensor/nnet/corr_gemm.c
+++ b/theano/tensor/nnet/corr_gemm.c
@@ -162,7 +162,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
                 "CorrMM shape inconsistency:\n"
                 "  bottom shape: %%d %%d %%d %%d\n"
                 "  weight shape: %%d %%d %%d %%d\n"
-                "  top shape: %%d %%d %%d %%d (expected %%d %%d %%d %%d)\n",
+                "  top shape: %%ld %%ld %%ld %%ld (expected %%d %%d %%d %%d)\n",
                 batchSize, nChannels, bottomHeight, bottomWidth,
                 nFilters, nChannels, kH, kW,
                 PyArray_DIMS(top)[0], PyArray_DIMS(top)[1],
@@ -182,7 +182,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
     if (NULL == col)
     {
         PyErr_Format(PyExc_RuntimeError,
-                "CorrMM failed to allocate working memory of %%d x %%d\n",
+                "CorrMM failed to allocate working memory of %%ld x %%ld\n",
                 col_dim[0], col_dim[1]);
         return NULL;
     }
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -1,5 +1,38 @@
+import numpy
 import unittest
-from theano.tensor.nnet.abstract_conv import get_conv_output_shape
+import itertools
+
+import theano
+from theano import tensor
+from theano.tests import unittests_tools as utt
+from theano.tensor.nnet.abstract_conv import conv, get_conv_output_shape
+from theano.tensor.nnet import corr
+from theano.tensor.nnet.corr import (CorrMM, CorrMM_gradWeights,
+                                     CorrMM_gradInputs)
+from theano.tensor.nnet.conv import ConvOp
+from theano.tensor.nnet.ConvGrad3D import ConvGrad3D
+from theano.tensor.nnet.ConvTransp3D import ConvTransp3D
+
+
+def conv_corr(inputs, filters, border_mode="valid", subsample=(1, 1),
+              conv_mode='conv'):
+    if conv_mode == 'conv':
+        filters = filters[:, :, ::-1, ::-1]
+    return corr.CorrMM(border_mode, subsample)(inputs, filters)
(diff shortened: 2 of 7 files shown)


nouiz:
travis is failing.

nouiz:
Other then travis, it look ok to me, but I'll give time to @lamblin to review.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1,5 +1,41 @@
...
+                                     CorrMM_gradInputs)
+from theano.tensor.nnet.ConvGrad3D import ConvGrad3D
+from theano.tensor.nnet.ConvTransp3D import ConvTransp3D
+
+
+def conv_corr(inputs, filters, border_mode="valid", subsample=(1, 1),
+              conv_mode='conv'):
+    if conv_mode == 'conv':
+        filters = filters[:, :, ::-1, ::-1]
+    return corr.CorrMM(border_mode, subsample)(inputs, filters)
```
lamblin: I don't really like that the reference function consists in more than one single op depending on `conv_mode`, since there is a bigger risk of getting the gradient wrong.
However, since it has already been tested in the past with the single-op cudnn, I guess the advantage or being able to run the tests without a GPU (or without cudnn) is bigger.

lamblin:
Thanks for the refactoring.
Should we do that in `theano/sandbox/cuda/tests/test_abstract_conv.py` as well?

theano/tensor/nnet/corr_gemm.c
```diff
@@ -162,7 +162,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
                 "CorrMM shape inconsistency:\n"
                 "  bottom shape: %%d %%d %%d %%d\n"
                 "  weight shape: %%d %%d %%d %%d\n"
-                "  top shape: %%d %%d %%d %%d (expected %%d %%d %%d %%d)\n",
+                "  top shape: %%ld %%ld %%ld %%ld (expected %%d %%d %%d %%d)\n",
```
lamblin: Why not change the other `%%d` as well?
And should we cast the arguments into `(long)`?

abergeron: Because the ones I changed are already `long`.  The others are ints (that should maybe be changed to longs).

abergeron:
I'll do the cuda tests too.

theano/tensor/nnet/corr_gemm.c
abergeron: Actually they shouldn't be ints because they interact with the BLAS api which takes ints.

theano/tensor/nnet/tests/test_abstract_conv.py
abergeron: I think that the chance of all implementations getting their gradient wrong in the same way is very small.

And since we compare against all of them, this should ferret out if we ever break the gradient on CorrMM.

Also, just because CuDNN is a single op does not shield it from gradient errors...

abergeron:
I've modified the cuda tests and fixed the problems in the current tests.

The cuda and gpuarray tests pass locally.

abergeron:
There is a problem with these tests now: they are too long and make travis timeout.  I figure we could cut on the number of shapes or border_modes we try.  Or we can try to get smarter about how the combinations are tried and less than the cross product of everything to reduce the total time.

I'll try to come up with a reasonable combination.

lamblin:
@abergeron there is a point about that in #3557. I think the best way is to generate the `product()` once, and select a couple of them randomly (based on `rseed`). I don't think we should cut the number of shapes or border_modes.

lamblin:
I don't like the idea of never testing combinations of non-default strides and borders, for instance, but I guess a better solution can always be implemented later.

abergeron:
I didn't cut anything, but I avoided the giant product of everything.

lamblin:
That's what I'm saying: we never try non-default strides and non-default border at the same time anymore.
There are calls to `self.tcase(i, f, s, db, ...)`, `(i, f, ds, b, ...)`, but never to `(i, f, s, b)` where both `s != ds` _and_ `b != db`.

abergeron:
I've made border_mode and strides tested together, but kept the rest separate.

abergeron:
I've fixed the travis problems.

nouiz:
there is still travis problems.

abergeron:
Now it should be fixed (again).

abergeron:
And now subsample will be a tuple so it won't crash anymore.

abergeron:
I made it inherit from TestCase again and just skipped the test in the base class.  Now it should hopefully pass.

abergeron:
I'm kind of stuck on this one since it passes locally.  I'll try to more closely reproduce the travis environment to see if I can get the error.

abergeron:
I am a bit stumped on this because the tests pass locally (on two different machines) whereas they don't on travis.

I suspect a BLAS issue preventing the replacement with CorrMM is the problem.

nouiz:
It is possible that that config.blas.ldflags is empty? Can you check if Theano link to blas directly in travis?

abergeron:
I'll check.

abergeron:
This should be fixed once #4095 is merged

nouiz:
The tests should be skipped if blas isn't there and they don't run without blas.

abergeron:
It skips the CorrMM test is there is no blas now.

Issue #3103, opened 2015-07-02
Add the doc generation in travis


nouiz:
- [x] Add in the file .travis.yml a new test run that would generate Theano doc with the script doc/scripts/docgen.py. Configure it such that if there is an error in the doc, it make travis fail. (so mostly the process must return an error.)
- [x] If there is current error in the doc, fix them.
- [x] Make it also run doctest (docgen.py --test)
- [ ] Make travis fail when there is an error in doc generation
- [ ] Make travis fail when there is an error in doc test

hantek:
doc/scripts/docgen.py --nopdf

abergeron:
There is also `doc/scripts/docgen.py --test` that will run the code blocks embedded inside the docs.  This will make sure that the examples stay unbroken.

nouiz:
The only think left is to make travis fail when the doc test or doc generation.

nouiz:
done by gh-4030

PR #4030, opened 2016-02-11
solving doc test failures

hantek:
Solves test failures in the doc tests. Fix for #3103 .

### DIFF
--- a/doc/tutorial/profiling.txt
+++ b/doc/tutorial/profiling.txt
@@ -33,7 +33,7 @@ functions using either of the following two options:
     - You can also combine the profile of many functions:
 
       .. testcode::
-
+          :hide:
           profile = theano.compile.ProfileStats()
           f = theano.function(..., profile=profile)
           g = theano.function(..., profile=profile)
--- a/doc/tutorial/profiling.txt
+++ b/doc/tutorial/profiling.txt
@@ -30,14 +30,15 @@ functions using either of the following two options:
 2. Pass the argument :attr:`profile=True` to the function :func:`theano.function <function.function>`. And then call :attr:`f.profile.print_summary()` for a single function.
     - Use this option when you want to profile not all the
       functions but one or more specific function(s).
-    - You can also combine the profile of many functions:
-
-      .. testcode::
+    - You can also combine the profile of many functions: 
+    
+      .. doctest::
           :hide:
+
           profile = theano.compile.ProfileStats()
-          f = theano.function(..., profile=profile)
-          g = theano.function(..., profile=profile)
-          ...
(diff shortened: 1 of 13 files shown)


doc/tutorial/profiling.txt
```diff
@@ -33,7 +33,7 @@ functions using either of the following two options:
     - You can also combine the profile of many functions: 
     
       .. testcode::
-
+          :hide:
```
nouiz: This cause error in travis. I think you need an empty line after hide.

doc/tutorial/profiling.txt
```diff
@@ -33,11 +33,12 @@ functions using either of the following two options:
...
       .. testcode::
+          :hide:
 
           profile = theano.compile.ProfileStats()
-          f = theano.function(..., profile=profile)
-          g = theano.function(..., profile=profile)
-          ...
+          f = theano.function(..., profile=profile)  # doctest: +SKIP
+          g = theano.function(..., profile=profile)  # doctest: +SKIP
+          ...  # doctest: +SKIP
```
hantek: Adding a blank line still doesn't solve the problem. What we need here is to have a testcode block that doesn't actually get executed. I also tried adding "# doctest: +skip" at the end of each line, but it didn't work too.

hantek: I changed the block type to "doctest" instead, which made "# doctest: + SKIP" work. The layout of "doctest" in html seems to be the same to "testcode."

nouiz:
You need to modify the .travis.yml file to uncomment the execution of tests. line 69 in master.

hantek:
Hi Fred, Travis has passed, it looks ready for merge now.

hantek:
There is some problem in python 3.3, like it can't import cPickle, can't recognize "file," etc.  What did we  usually do on these?

nouiz:
In python 3, xrange don't exist. Just use range instead.

In Python 3; "print 3" don't exist. You should use "print(3)" instead. print is now a function.

Fix the import of cPickle by using this import "from six.moves import cPickle"

For the function file, I think you can replace it with open()

hantek:
Yes but what confused me is that, why these test got passed in python3.3 previously but starts to fail from the last commit? Nothing relating to that was changed.

nouiz:
In the commit  b5d75ef, you added the test back in travis, but travis didn't ran for this commit. I don't know why, this is why it got marked as passed (travis didn't run, so it didn't fail).

hantek:
Hi Fred, I updated the last commit correspondingly.

nouiz:

nouiz:
thanks

Issue #4024, opened 2016-02-09
Yet another Failed to compile cuda_ndarray.cu:


sygi:
1. I checked cuda by running nvidia-smi and deviceQuery and it works
2. Running (on the current github master of theano):
```
THEANO_FLAGS=device=gpu python -c "import theano; print(theano.sandbox.cuda.device_properties(0))"
```
gives me long error code and after it:
```
/usr/bin/ld: cannot find -lpython2.7.a
collect2: ld returned 1 exit status

['nvcc', '-shared', '-O3', '-m64', '-Xcompiler', '-DCUDA_NDARRAY_CUH=c72d035fdf91890f3b36710688069b2e,-DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION,-fPIC,-fvisibility=hidden', '-Xlinker', '-rpath,/icm/home/sygnowsk/.theano/compiledir_Linux-2.6-v1-x86_64-with-redhat-6.6-Carbon-x86_64-2.7.5-64/cuda_ndarray', '-I/icm/home/sygnowsk/dqn/deep_q_rl/venv/lib/python2.7/site-packages/theano/sandbox/cuda', '-I/icm/home/sygnowsk/dqn/deep_q_rl/venv/lib/python2.7/site-packages/numpy/core/include', '-I/mnt/software/common/python/2.7.5/include/python2.7', '-I/icm/home/sygnowsk/dqn/deep_q_rl/venv/lib/python2.7/site-packages/theano/gof', '-o', '/icm/home/sygnowsk/.theano/compiledir_Linux-2.6-v1-x86_64-with-redhat-6.6-Carbon-x86_64-2.7.5-64/cuda_ndarray/cuda_ndarray.so', 'mod.cu', '-L/mnt/software/common/python/2.7.5/lib', '-lpython2.7.a', '-lcublas', '-lcudart']
[...]

PR #4032, opened 2016-02-11
(#4024) cutting the *.a extension in ldlibrary file

sygi:
Small change to resolve problems when LDLIBRARY points to a *.a file instead of *.so.

### DIFF
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -1661,6 +1661,8 @@ def std_lib_dirs_and_libs():
         # remove extension if present
         if libname.endswith(".so"):
             libname = libname[:-3]
+        elif libname.endswith(".a"):
+            libname = libname[:-2]
 
         libdir = distutils.sysconfig.get_config_var("LIBDIR")
 


nouiz:
Do this allow you to compile correctly? Which library do you wanted to link
to that had only a .a extension?

sygi:
Yes, it helped me to make it work.
It was the python2.7 library. I had both *.so and *.a (and also *.so.1.0) in the directory, but for some reason LDLIBRARY config variable was pointing to the libpython2.7.a file.

nouiz:
What do you mean by the LDLIBRARY config variable? I don't remember that configuration variable. Do you mean the env variable LD_LIBRARY_PATH?

What OS do you have?

sygi:
I mean the output of [distutils.sysconfig.get_config_var("LDLIBRARY")](https://github.com/Theano/Theano/blob/master/theano/gof/cmodule.py#L1656), which, as far as I understand, points to a shared library of the python installation.

The system is my university cluster, which runs on CentOS6. I agree that it has some freaky architecture which I don't understand. I can try to provide more information if you'll ask more questions, though.

nouiz:
I just saw that this was discussed at gh-4024. I'm good with the PR. thanks.

nouiz:
I just saw that this was discussed in gh-4024. thanks.

nouiz:
As it fix completly the problem for you and make it work. I think we can
merge it. If this was just a partial fix, I would have waited for the
complete fix.

thanks

PR #4033, opened 2016-02-11
Fix CudaNdarray repr in python 3 and a crash.

nouiz:
This fix daily buildbot error in python 3 and a crash during optimization.


lamblin:
Looks good to me, OK to merge if everything passes locally.

nouiz:
local tests under sandbox/cuda passed, so merging.

Issue #3466, opened 2015-09-29
Use contexts to reduce the number C code instances we compile


abergeron:
Currently the only plan for contexts is to use them for multi-gpu, but we could also make structs and pass them through that to give ops theirs node params.

Since it would reduce the diversity of C code compiled we could avoid repeated compilations in a number of cases.

nouiz:
We can put the destroy_map, view_map in it to lower the number of compiled op in addition to the op attribute.

nouiz:
We should list all the ops that still need that.
Subtensor probably need a refactoring, so don't do it now.

notoraptor:
I try to maintain a list here: https://github.com/Theano/Theano/issues/5696#issuecomment-286161436

Maybe should this list be moved in another place ?

PR #4038, opened 2016-02-11
[WIP] Compiled c instances reduction by adding the destroy map as a Param to Elemwise Op

Hengjean:
Referencing to this issue : https://github.com/Theano/Theano/issues/3466

I'm not sure if I implemented that correctly. For now, I'm only able to have Ops with destroy maps of same length compiled together. 

I also still haven't looked at GPUElemwise as mentioned in the issue.


theano/tensor/elemwise.py
```diff
@@ -1059,35 +1076,110 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+        olength = len(_onames)
+        outArray = """
+        PyArrayObject **outRef[%(olength)s];
+        """ % locals()
+        for oname, j in izip(_onames, range(len(onames))):
+            outArray += """
+            outRef[%(j)i]=&%(oname)s;
+            """ % locals()
+
+        alloc += outArray + inArray
```
nouiz: I would do "inArray + outArray" just to keep the same order as they where created.

theano/tensor/elemwise.py
```diff
@@ -1059,35 +1076,110 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+
+        # Declaration and initialization of the hardcoded olv_index
+        # used by the code generated in elemwise_cgen
+        alloc += """
+        int olv_index;
+        olv_index = %(j)i;
+        """ % locals()
+
+        # If the node has a non empty destroy map
+        if(len(self.inplace_pattern.keys()) > 0):
```
nouiz: You can remove the .keys() call. len(a_dict) return the number of element in the dict.

theano/tensor/elemwise.py
```diff
@@ -1059,35 +1076,110 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+            alloc += """
+            int outDm[%(dml)s];
+            int inDm[%(dml)s];
+            PyObject *tmpOut, *tmpIn;
+            Py_ssize_t pos;
+            pos = 0;
+            PyObject* destroyMap;
+            destroyMap =PyObject_GetAttrString(%(p)s,"inplace_pattern");
+            if( destroyMap == NULL){
+                //Should place code to fail here
```
nouiz: Normally, it is just "%(fail)s". "fail" is in sub.

theano/tensor/elemwise.py
```diff
@@ -1059,35 +1076,110 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+            }
+            """ % dict(dml=len(self.inplace_pattern.keys()), p=sub['params'],ilength=ilength)
```
nouiz: no need to call keys(). ilength isn't needed.

theano/tensor/elemwise.py
```diff
@@ -1059,35 +1076,110 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+
+            # We no longer need the reference to the original Python object
+            # containing the destroy map
+            alloc += """
+            Py_XDECREF(destroyMap);
+            """
 
         # We loop over the "aliased" outputs, i.e., those that are
         # inplace (overwrite the contents of one of the inputs) and
         # make the output pointers point to theur corresponding input
```
nouiz: theur -> their

theano/tensor/elemwise.py
```diff
@@ -1059,35 +1076,110 @@ def _c_all(self, node, nodename, inames, onames, sub):
                                      fortran=alloc_fortran)
             alloc += cgen.make_checks([list(range(nnested))], [odtype],
                                       dict(sub, lv0=oname))
-        olv_index = i  # index of the last output
+        # index of the last output
```
nouiz: All the inplace work is done after the alloc of the outputs was done. So mostly, we alloc an output, then we free it. Why not reverse this, set reference of outputs that are destroy of inputs, then alloc the others?

The call make_alloc will only allocate the output if the isn't NULL. In that case, he call resize on it that will do nothing as it will be of the good size.

nouiz:
I have done a first round of review. thanks!

theano/tensor/elemwise.py
Hengjean: I'm not sure I understand.

If I'm not mistaken the allocated outputs are already only the "real" outputs not present in the destroy map.

Line 1014 of this file reads : 
        # These are the outputs that we will need to allocate
        # (output, name, name of the c type), transposed
        real = list(zip(*[(r, s, r.type.dtype_specs()[1])
                          for r, s in izip(node.outputs, onames)
                          if r not in dmap]))

to declare the array then used line 1068 to allocate the outputs. From my understanding of this, no allocated outputs are freed by the inplace work.

Hengjean:
I made a few fixes in the code and its formatting. However Travis seems to fail compilation on some tests without clearly indicating what went wrong. I tried replicating these errors locally but all tests are passing without compilation errors when running "theano-test".

theano/tensor/elemwise.py
nouiz: line 1075, put in the "alloc" bloc a call to cgen.make_alloc(). I was thinking it was called for all output, but you are right, they are called only for output that we aren't inplace on.

But this cause problems I think. we want to have all the elemwise version independent of the inplace use the same c code. So we should not check which output is inplace or not. So I think we should change the logic to something like:

- set all output that are inplace.
- then check all outputs that have not been set and allocated them if they aren't preallocated.

nouiz:
The travis error you see are mostly with python3. Can you easily install it and test with it? You can use anaconda to install it relatively easily.

Hengjean:
I just tried to reproduce these errors with python 3.4.3 (version of python3 packaged with the ubuntu version I am using) but everything compiled successfully. Should I try again with python 3.3 (which seems to be the version travis is using) ? Isn't there a way to make Travis print the error even though the log is over 4 Mb ?

theano/tensor/elemwise.py
Hengjean: Understood, I'll see what I can do about that.

nouiz:
About Python 3, you can wait after the other part. To replicate travis setup, you can do it with conda. Chexk the travis file, check the section before_install and install:

https://github.com/Theano/Theano/blob/master/.travis.yml#L14-L38

Hengjean:
My last two commits introduced a few errors in the tests. A key error on params for some tests and Gradient errors for some others. 

The gradient error is surely due to the quite naive change I made in my last commit as I only switched places of two codes block, surely more error checking is needed in the C code. I'm working on it.
However, the KeyError on param is puzzling me. I think it's because I began using param on Elemwise nodes without a destroy map but no errors are thrown when I compile simple operations such as add_inplace without specifying a destroy map in its parameters.


The C code is beginning to look quite similar for nodes with destroy map of different sizes. On the add_inplace operation I use as an example, only the define used to alias the inplace variable and some parts of the code generated in make_reordered_loop are left different.

Hengjean:
I found the reason of the keyError, c_all in CarReduce call c_all in Elemwise passing it its own sub variable which does not contain the param key. Committing the fix asap.

nouiz:
travis have errors. I'll review the recent commit.

nouiz:
I'm too tired to review this. I'll try next week.

thanks.

Hengjean:
I have still one last big commit in the works, so please wait until I push it before reviewing. If everything goes well it should be over today or tomorrow and it would finish making the C_Codes for the same elemwise ops the same independent of their destroy map. 

Right now, I found a solution for the variable aliasing problem, all that's left is the code generated in make_reordered_loop() in elemwise_cgen.py. I'm struggling a little to understand it but I have a general idea on how I should proceed.

Once that's over, I'll go on bug hunting. Right now the errors generated by Travis are generally either a Python3 related error (the log mentions "PyIntAsLong not function not found" quite frequently) or a gradient error over the acceptable error threshold.

Hengjean:
Sorry for being late with the commit. I was able to make the c code completely independent of the destroyMap and it's working with simple ops but some tests are generating a segfault I'm trying to fix before committing.

nouiz:
thanks for the update.

theano/tensor/elemwise.py
```diff
@@ -1187,12 +1189,12 @@ def _c_all(self, node, nodename, inames, onames, sub):
             for output, oname, i in izip(node.outputs, onames,
                                          range(len(onames))):
                 idtype = output.type.dtype_specs()[1]
+
                 # We alias the scalar variables
                 defines += """
-                #define %(oname)s_i *(%(idtype)s*) (refVar[i%(oname)s])
-                int i%(oname)s;
-                i%(oname)s = outIndexes[%(i)i];
+                #define %(oname)s_i *(t%(i)i*) (refVar[outIndexes[%(i)i]])
```
Hengjean: Basically, we would be able to retrieve the type of the variable where the refVar array points to and typedef it to t%(i) so that the casting is properly done (else it will lead to segfaults). However I was not able to find a solution relying only on C code. I think I'm thinking too much object oriented and not enough in C.

That or find an other implementation than macro to alias the inplace variables.

In order to make the test pass I've committed a temporary hack that will be removed from history once a solution is found.

theano/tensor/elemwise.py
```diff
@@ -1221,8 +1223,9 @@ def _c_all(self, node, nodename, inames, onames, sub):
         }
         """ % locals()
 
-        loop_orders = orders + [list(range(nnested))] * len(real_onames)
-        dtypes = (idtypes + list(real_odtypes))
+        loop_orders = orders + [list(range(nnested))] * len(onames)
```
Hengjean: In order to make the c code independent of the destroy maps, I'm giving all outputs even those inplace to the loop_orders variable.
To be honest I'm not sure if that's a good idea without any modifications on the functions make_loop and make_loop_reordered in elemwise_cgen.py. But the code there is a little cryptic to me and I've yet to understand it good enough to understand what problems that might generate.

Hengjean:
Sorry again for being so late.

I made the commit, however I was not able to fix the segfault without relying on the python side of things. More details on the line notes in the commit.

Moreover, even though I was able to make the c_code() function of elemwise generate the exact same code for the same ops with destroy maps of different sizes, there was still two different folders generated in the cache. Making a diff on the mod.cpp of those two folders revealed one variable more declared in one file. I will look more into it once the Elemwise code is fixed.

Lastly, I begin to ask myself about the efficiency of the code I submitted. That's quite a lot more operations I added in the C_code, won't that slow things down so much at execution time that the time gained at compilation won't be worth it ? Perhaps it would be better to make this optimization optional depending on the user needs.

theano/tensor/elemwise.py
```diff
@@ -1045,11 +1062,96 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+
+        alloc += inArray + outArray
+
+        # Declaration and initialization of the hardcoded olv_index
+        # used by the code generated in elemwise_cgen
+        alloc += """
+        int olv_index;
+        olv_index = %(j)i;
+        """ % locals()
+        if 'params' in sub:
```
Hengjean: CAReduce call the c_all function of Elemwise without properly generating the sub. That seemed to be quite a corner case and always called on node without a destroy map so I chose to fix it like that rather than search how to make CAReduce generate properly the sub with params in it.

theano/tensor/elemwise.py
```diff
@@ -1059,35 +1161,60 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+                if(i in self.inplace_pattern):
+                    name = _inames[self.inplace_pattern[i]]
+                else:
+                    name = _onames[i]
+                defines += """
+                    typedef typeof(%(name)s_i) t%(i)i;
+                """ % locals()
+                # We alias the scalar variables
+                defines += """
+                #define %(oname)s_i *(t%(i)i*) (refVar[outIndexes[%(i)i]])
```
Hengjean: Basically, we would be able to retrieve the type of the variable where the refVar array points to and typedef it to t%(i) so that the casting is properly done (else it will lead to segfaults). However I was not able to find a solution relying only on C code. I think I'm thinking too much object oriented and not enough in C.

That or find an other implementation than macro to alias the inplace variables.

In order to make the test pass I've committed a temporary hack that will be removed from history once a solution is found.

Hengjean:
Quick and dirty rebase for me to study the changes made since last year to Theano. Code and history will be cleaned once I wrap my head around everything that need to be done.

Hengjean:
Sorry for the ugly commit history and the slow work. 
I ran into an internal python error when trying to run a code as simple as 
x = T.dscalar('x')
y = T.dscalar('y')
z = x + y
f = function([x, y], z)
print(f(2,3))

with my pull request.

The error log is as follow : 
  File "basic.py", line 8, in <module>
    print(f(2,3))
  File "/home/heng/git/TheanoDev/Theano/theano/compile/function_module.py", line 915, in __call__
    for c in self.input_storage:
SystemError: ../Objects/dictobject.c:1704: bad argument to internal function

I want to see if Travis throws the same error or if it is somehow just a problem with my own installation.

nouiz:
@notoraptor do you have an idea?

theano/tensor/elemwise.py
```diff
@@ -995,11 +1024,96 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+               }
+               """ % dict(p=sub['params'], fail=sub['fail'])
+            # Iterate over the python object to fill the destroy map arrays
+            # and make the output point to the corresponding input and
+            # decrease the reference of whatever the output contained
+            # prior to this
+
+            alloc += """
+               int lenDM;
+               lenDM = PyDict_Size(destroyMap);
```
notoraptor: Hello! The problem seems to come from here. `inplace_pattern` (here the `destroyMap`) is a `frozendict`, not a dict (and `frozendict` class is not even a subclass of `dict`). I suggest you modify  `ElemwiseParams` to add a real dict field that you will then get in C code.
for example:
```python
class ElemwiseParams(object):
    def __init__(self, inplace_pattern):
        self._inplace_pattern = inplace_pattern  # frozendict is still useful for hashing method.
        self.inplace_pattern = self._inplace_pattern._dict  # Now `inplace_pattern` is a real dict.

    def __hash__(self):
        return hash(self._inplace_pattern)  # Using the frozendict for hashing.

```

Hengjean: Thank you, that was the problem.

theano/tensor/elemwise.py
```diff
@@ -455,16 +472,18 @@ class Elemwise(OpenMPOp):
...
     def __init__(self, scalar_op, inplace_pattern=None, name=None,
                  nfunc_spec=None, openmp=None):
+
         if inplace_pattern is None:
             inplace_pattern = frozendict({})
+
         self.name = name
         self.scalar_op = scalar_op
-        self.inplace_pattern = frozendict(inplace_pattern)
-        self.destroy_map = dict((o, [i]) for o, i in self.inplace_pattern.items())
```
nouiz: You must keep the destroy_map for the Theano execution engine. Otherwise, it won't execute the nodes in a good order.

Hengjean: That explains the bugs that appeared and had nothing to do with the c code. Thank you very much I will try to push a commit with that and other things by the end of the week.

nouiz:
thanks for continuing this.

Hengjean:
No problem. I'm sorry to be so slow, I try to work regularly on this pull request but I can only manage to squeeze a few hours a week most of which spent relearning some basics about Theano architecture or trying to understand my own code, so I'm not being the most efficient.

nouiz:
There is probably just one or two error left. This make most part of travis fail, but in each part, there is few errors. I think this one is a good one to investigate (probaly easier to understand what is going on:

https://travis-ci.org/Theano/Theano/jobs/251059712#L2185

Running with mode=DebugMode will probably tell you which Elemwise op is cause the wrong value.

thanks

Hengjean:
Update on current debug progress:

Thanks for the pointer, however in debug mode, some tests no longer fails (including test_grad you linked to). Test in Scan seems to get stuck in an infinite loop as they no longer returns even after waiting for 10 mins.

The only test I was able to get an helpful log was this one : 

THEANO_FLAGS="mode=DebugMode" theano-nose theano/tests/test_rop.py:test_RopLop.test_conv        
/home/heng/.local/lib/python2.7/site-packages/nose_parameterized/__init__.py:7: UserWarning: The 'nose-parameterized' package has been renamed 'parameterized'. For the two step migration instructions, see: https://github.com/wolever/parameterized#migrating-from-nose-parameterized-to-parameterized (set NOSE_PARAMETERIZED_NO_WARN=1 to suppress this warning)
  "The 'nose-parameterized' package has been renamed 'parameterized'. "
/home/heng/.local/lib/python2.7/site-packages/numpy/lib/function_base.py:4011: RuntimeWarning: Invalid value encountered in median
  r = func(a, **kwargs)
E
ERROR: test_conv (theano.tests.test_rop.test_RopLop)
  File "/home/heng/git/TheanoDev/Theano/theano/tests/test_rop.py", line 345, in test_conv
    v2 = scan_f(image_data, filter_data, ev_image_data, ev_filter_data)
  File "/home/heng/git/TheanoDev/Theano/theano/compile/function_module.py", line 884, in __call__
  File "/home/heng/git/TheanoDev/Theano/theano/compile/debugmode.py", line 2161, in deco
    return f()
  File "/home/heng/git/TheanoDev/Theano/theano/compile/debugmode.py", line 1887, in f
    new_e = e.__class__(msg)
TypeError: __init__() takes at least 6 arguments (2 given)
-------------------- >> begin captured logging << --------------------
theano.compile.debugmode: WARNING: Optimization Warning: input idx 0 marked as destroyed was not changed for node 'Elemwise{Mul}[(0, 0)](Subtensor{::, ::, ::int64, ::int64}.0, ev_filters_copy)'
--------------------- >> end captured logging << ---------------------

Ran 1 test in 18.618s

which seems to indicate that the destroy map is not working correctly. I have not yet tested the debug mode for all failing tests.

I think I will try to reproduce the bug with the minimal amount of code and debug from there. I also have to read again some parts of this pull request to make sure everything is correct and needed.

Hengjean:
Update on my progress.

I was able to obtain a better error log on the test mentioned in the previous comment. The actual exception is : 
```
An optimization (probably scanOp_make_inplace) inserted an apply node that raise an error.
The information we have about this optimizations is:for{cpu,scan_fn}.0
  for{cpu,scan_fn} [id A] <TensorType(float64, vector)> ''   
   |Elemwise{Mul}[(0, 0)] [id B] <TensorType(int64, scalar)> ''   
   | |Shape_i{0} [id C] <TensorType(int64, scalar)> ''   
   | | |input [id D] <TensorType(float64, 4D)>
   | |Shape_i{0} [id E] <TensorType(int64, scalar)> ''   
   | | |filter [id F] <TensorType(float64, 4D)>
   | |Elemwise{Composite{((i0 + i1) - i2)}}[(0, 1)] [id G] <TensorType(int64, scalar)> ''   
   | | |TensorConstant{1} [id H] <TensorType(int64, scalar)>
   | | |Shape_i{2} [id I] <TensorType(int64, scalar)> ''   
   | | | |input [id D] <TensorType(float64, 4D)>
   | | |Shape_i{2} [id J] <TensorType(int64, scalar)> ''   
   | |   |filter [id F] <TensorType(float64, 4D)>
   | |Elemwise{Composite{((i0 + i1) - i2)}}[(0, 1)] [id K] <TensorType(int64, scalar)> ''   
   |   |TensorConstant{1} [id H] <TensorType(int64, scalar)>
   |   |Shape_i{3} [id L] <TensorType(int64, scalar)> ''   
   |   | |input [id D] <TensorType(float64, 4D)>
   |   |Shape_i{3} [id M] <TensorType(int64, scalar)> ''   
   |     |filter [id F] <TensorType(float64, 4D)>
   |Subtensor{int64:int64:int8} [id N] <TensorType(int64, vector)> ''   
   | |ARange{dtype='int64'} [id O] <TensorType(int64, vector)> ''   
   | | |TensorConstant{0} [id P] <TensorType(int8, scalar)>
   | | |Elemwise{Mul}[(0, 0)] [id B] <TensorType(int64, scalar)> ''   
   | | |TensorConstant{1} [id Q] <TensorType(int8, scalar)>
   | |ScalarFromTensor [id R] <int64> ''   
   | | |Elemwise{Composite{Switch(i0, i1, minimum(i2, i3))}} [id S] <TensorType(int64, scalar)> ''   
   | |   |Elemwise{le,no_inplace} [id T] <TensorType(bool, scalar)> ''   
   | |   | |Elemwise{Composite{Switch(LT(i0, i1), Switch(LT((i0 + i0), i1), i1, (i0 + i0)), i0)}} [id U] <TensorType(int64, scalar)> ''   
   | |   | |TensorConstant{0} [id P] <TensorType(int8, scalar)>
   | |   |TensorConstant{0} [id P] <TensorType(int8, scalar)>
   | |   |TensorConstant{0} [id V] <TensorType(int64, scalar)>
   | |   |Elemwise{Mul}[(0, 0)] [id B] <TensorType(int64, scalar)> ''   
   | |ScalarFromTensor [id W] <int64> ''   
   | | |Elemwise{Composite{Switch(i0, i1, minimum(i2, i3))}}[(0, 2)] [id X] <TensorType(int64, scalar)> ''   
   | |   |Elemwise{le,no_inplace} [id T] <TensorType(bool, scalar)> ''   
   | |   |TensorConstant{0} [id P] <TensorType(int8, scalar)>
   | |   |Elemwise{Composite{Switch(LT(i0, i1), Switch(LT((i0 + i0), i1), i1, (i0 + i0)), i0)}} [id U] <TensorType(int64, scalar)> ''   
   | |   |Elemwise{Mul}[(0, 0)] [id B] <TensorType(int64, scalar)> ''   
   | |Constant{1} [id Y] <int8>
   |Elemwise{Mul}[(0, 0)] [id B] <TensorType(int64, scalar)> ''   
   |Reshape{1} [id Z] <TensorType(float64, vector)> ''   
   | |ConvOp{('imshp', (None, None, None)),('kshp', (None, None)),('nkern', None),('bsize', None),('dx', 1),('dy', 1),('out_mode', 'valid'),('unroll_batch', None),('unroll_kern', None),('unroll_patch', True),('imshp_logical', (None, None, None)),('kshp_logical', (None, None)),('kshp_logical_top_aligned', True)} [id BA] <TensorType(float64, 4D)> ''   
   | | |input [id D] <TensorType(float64, 4D)>
   | | |filter [id F] <TensorType(float64, 4D)>
   | |TensorConstant{(1,) of -1} [id BB] <TensorType(int64, (True,))>
   |ev_input [id BC] <TensorType(float64, 4D)>
   |ev_filters [id BD] <TensorType(float64, 4D)>
   |MakeVector{dtype='int64'} [id BE] <TensorType(int64, vector)> ''   
   | |Shape_i{0} [id C] <TensorType(int64, scalar)> ''   
   | |Shape_i{0} [id E] <TensorType(int64, scalar)> ''   
   | |Elemwise{Composite{((i0 + i1) - i2)}}[(0, 1)] [id G] <TensorType(int64, scalar)> ''   
   | |Elemwise{Composite{((i0 + i1) - i2)}}[(0, 1)] [id K] <TensorType(int64, scalar)> ''   
   |Alloc [id BF] <TensorType(float64, vector)> ''   
   | |TensorConstant{(1,) of 0.0} [id BG] <TensorType(float64, (True,))>
   | |Elemwise{Mul}[(0, 0)] [id B] <TensorType(int64, scalar)> ''   
   |InplaceDimShuffle{1,0,2,3} [id BH] <TensorType(float64, 4D)> ''   
   | |input [id D] <TensorType(float64, 4D)>
   |Subtensor{::, ::, ::int64, ::int64} [id BI] <TensorType(float64, 4D)> ''   
     |InplaceDimShuffle{1,0,2,3} [id BJ] <TensorType(float64, 4D)> ''   
     | |filter [id F] <TensorType(float64, 4D)>
     |Constant{-1} [id BK] <int64>
     |Constant{-1} [id BK] <int64>


The original exception: 
BadThunkOutput
  Apply   : Elemwise{Mul}[(0, 0)](Subtensor{::, ::, ::int64, ::int64}.0, ev_filters_copy)
  op      : <class 'theano.tensor.elemwise.Elemwise'>
  Outputs Type: TensorType(float64, 4D)
  Outputs Shape: (2, 2, 2, 3)
  Outputs Strides: (48, 96, 24, 8)
  Inputs Type : [TensorType(float64, 4D), TensorType(float64, 4D)]
  Inputs Shape: [None, (2, 2, 2, 3)]
  Inputs Strides: [None, (96, 48, 24, 8)]
  Inputs values: [<theano.compile.debugmode.DataDestroyed instance at 0x7f423d81bcb0>, 'not shown']
  Bad Variable: Elemwise{Mul}[(0, 0)].0
  thunk1  : perform
  thunk2  : c_code
           : shape, dtype, strides, min, max, n_inf, n_nan:
  Expected : (2, 2, 2, 3) float64 (48, 96, 24, 8) 0.0 0.825087804204 0 0 
  Value    : (2, 2, 2, 3) float64 (96, 48, 24, 8) 0.0 9.78346985015e-06 0 0 
  expected    : [[[[ 0.02210236  0.05846494  0.00432005]
   [ 0.15799687  0.42837768  0.29937713]]

  [[ 0.13901708  0.46305027  0.8250878 ]
   [ 0.01406407  0.0563943   0.00855108]]]


 [[[ 0.          0.          0.        ]
   [ 0.          0.          0.        ]]

  [[ 0.          0.          0.        ]
   [ 0.          0.          0.        ]]]]
  value    : [[[[  4.88127938e-07   4.13587699e-06   1.27031970e-07]
   [  6.76514336e-06   8.44186643e-06   7.00437122e-06]]

  [[  0.00000000e+00   0.00000000e+00   0.00000000e+00]
   [  0.00000000e+00   0.00000000e+00   0.00000000e+00]]]


 [[[  2.59131849e-06   5.32802544e-06   9.78346985e-06]
   [  3.40034942e-06   7.27321154e-06   2.32362992e-07]]

  [[  0.00000000e+00   0.00000000e+00   0.00000000e+00]
   [  0.00000000e+00   0.00000000e+00   0.00000000e+00]]]]
  Max Abs Diff:  0.825087804204
  Mean Abs Diff:  0.103200219615
  Median Abs Diff:  0.00216485457354
  Std Abs Diff:  0.200454061922
  Max Rel Diff:  nan
  Mean Rel Diff:  nan
  Median Rel Diff:  nan
  Std Rel Diff:  nan

  rtol, atol: 1e-05 1e-08
```
It seems there is a minor bug line 1888 of debugmode.py which caused the previous log. The catch code there is not made to work with BadThunkOutput exception. That can be easily fixed but that might be outside the scope of this pull request.

As for the actual error, it seems a destroyed value is used as input. Seeing as this seems linked to scan, the value might be destroyed while it is used. Output stride dimension 0 is also inverted with dimension 1 which, I think, causes the BadThunkOutput itself.

Next I will look at the code scanOp_make_inplace to see what might be going wrong. 

Again sorry for being so slow.

PR #4040, opened 2016-02-11
Fixes the multinomial grad dtype bug

caglar:


theano/sandbox/multinomial.py
```diff
@@ -52,7 +52,11 @@ def make_node(self, pvals, unis, n=1):
     def grad(self, ins, outgrads):
         pvals, unis, n = ins
         (gz,) = outgrads
-        return [T.zeros_like(x) for x in ins]
+        if x.dtype in T.discrete_dtypes:
```
nouiz: You can't do it like this. x isn't defined here. You need the if to be inside the loop. Each component much be checked individually.

caglar: Opps sorry pushed the wrong version by mistake!

caglar:
@nouiz @lamblin seems like tests are passing.

nouiz:
thanks

PR #3794, opened 2015-12-17
Moving flag declarations

martinarjovsky:
Flags are now on configdefaults except from the ones on tensor/blas.py (uses a lot of local variables and changes the logger) and tests/test_config.py .

Issue #3021, opened 2015-06-11
Warn Theano unknow Theano flags


nouiz:
Currently, we just ignore them. But we can't easily just warn about them as other lib like pylearn2 define there own flags that get know about later. Doing it at the end of the process would help, but if the user not always use the other lib, we still have the warning and having it at the end isn't great.

block solved this by (https://github.com/mila-udem/blocks/pull/712) by using a diff and if the key is missing, but close to known ones, print the warning at start up. I think this would also be good for Theano.

nouiz:
IRL discussion with @lamblin We think that it is better to move most if not all config declaration in the file configdefault.py. This will make sure that when the user define a config that we don't know about, we don't raise false warning (as it could be defined later).

abergeron:
Fixed by #4042

PR #4042, opened 2016-02-12
[WIP] Move flags declarations and warn user if unknown flags

adbrebs:
cf #3021 @abergeron 

I re-used commits of #3794 but there are circular imports that I did not manage to solve.
Basically, compiledir import configdefaults.py and vice versa


theano/configdefaults.py
```diff
@@ -860,3 +868,535 @@ def good_seed_param(seed):
...
+    ConfigParam(
+        default_compiledir,
+        filter=filter_compiledir,
+        allow_override=False),
+    in_c_key=False)
+
+# Check if there are remaining flags provided by the user through THEANO_FLAGS.
+if THEANO_FLAGS_DICT:
+    raise ValueError('theano does not recognise this flag: {}'
+                     .format(THEANO_FLAGS_DICT.keys()[0]))
```
abergeron: This should be a warning, not an error.

theano/configdefaults.py
```diff
@@ -1,11 +1,19 @@
...
+import re
+import socket
 
 import theano
 from theano.configparser import (AddConfigVar, BoolParam, ConfigParam, EnumStr,
                                  FloatParam, IntParam, StrParam,
-                                 TheanoConfigParser)
+                                 TheanoConfigParser, THEANO_FLAGS_DICT)
+from theano.gof.compiledir import (local_bitwidth, python_int_bitwidth,
+                                   gcc_version_str)
```
abergeron: You should probably move those in rather than import them.

Since they are used elsewhere, you can either import them in their original location from here, or change the imports.

adbrebs: Ok

adbrebs: I did what you suggested, which solves this circular import but there seems to be many more circular import issues. With the last commit, I know have:

  File "/Users/adeb/code/Theano/theano/__init__.py", line 52, in <module>
    from theano.gof import (
  File "/Users/adeb/code/Theano/theano/gof/__init__.py", line 38, in <module>
    from theano.gof.cc import \
  File "/Users/adeb/code/Theano/theano/gof/cc.py", line 29, in <module>
    from theano.gof import graph
ImportError: cannot import name graph

adbrebs:
@abergeron 
My last commits should fix all the remaining problems.

theano/configdefaults.py
```diff
@@ -860,3 +868,570 @@ def good_seed_param(seed):
...
+    ConfigParam(
+        default_compiledir,
+        filter=filter_compiledir,
+        allow_override=False),
+    in_c_key=False)
+
+# Check if there are remaining flags provided by the user through THEANO_FLAGS.
+if THEANO_FLAGS_DICT:
+    raise Exception('theano does not recognise this flag: {}'
+                    .format(THEANO_FLAGS_DICT.keys()[0]))
```
abergeron: This should not raise an execption, but print a warning.  It should also do so for every remaining key.

abergeron:
Apart from my comment, there are a bunch of flake8 errors, mostly related to now-useless AddConfigVar imports.

theano/configdefaults.py
adbrebs: done

adbrebs:
flake8 errors have been corrected.

Issue #4044, opened 2016-02-12
CpuCorrMM uses BLAS without checking availability


f0k:
CpuCorrMM uses _sgemm() or _dgemm(), but doesn't check if a BLAS library is actually available. This leads to cryptic error messages for Windows users (https://groups.google.com/forum/#!searchin/lasagne-users/undefined$20reference/lasagne-users/IQjKYgoBWr4/77813wVNFQAJ, https://github.com/Lasagne/Lasagne/issues/607).

Is there any way to prevent CpuCorrMM from being used in that case, or to use Theano's gemm fallback, or to give a clear error message telling the user to install a BLAS library (preferrably linking to corresponding documentation on the web)?

nouiz:
https://github.com/Theano/Theano/pull/4045

thanks

PR #4045, opened 2016-02-12
Don't use CpuCorrMM when there is no CPU BLAS

nouiz:
fix gh-4044

### DIFF
--- a/theano/tensor/nnet/opt.py
+++ b/theano/tensor/nnet/opt.py
@@ -62,7 +62,7 @@ compile.optdb.register('local_inplace_sparse_block_outer',
 # Conv opts
 @local_optimizer([AbstractConv2d])
 def local_abstractconv_gemm(node):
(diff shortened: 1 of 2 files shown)


theano/tensor/nnet/opt.py
```diff
@@ -62,7 +62,7 @@ def local_inplace_sparse_block_outer(node):
 # Conv opts
 @local_optimizer([AbstractConv2d])
 def local_abstractconv_gemm(node):
-    if theano.config.cxx == "":
+    if theano.config.cxx == "" or theano.config.blas.ldflags:
         return
```
f0k: Shouldn't this be `or not theano.config.blas.ldflags`? Otherwise you're skipping it exactly when BLAS is available.

nouiz: right, I updated the PR.

abergeron:
Ok when travis passes.

PR #4046, opened 2016-02-12
tip for mixed dtypes

### LINKED ISSUE #4035: Add tip in regarding mixed dtypes to profiling output.
See @lamblin's comment in #3964.

Context : the logistic regression example from the deep learning tutorials was used to provide an example of the output of Theano's profiling mode. However, in the profile output, it seems like one of the dot products has been kept in the graph because of float32/float64 dtypes (otherwise it could have been removed).

AnishShah:
Fixes #4035

### DIFF
--- a/theano/compile/profiling.py
+++ b/theano/compile/profiling.py
@@ -1391,6 +1391,16 @@ class ProfileStats(object):
                           " generator supported on the GPU.", file = file)
                 break
 
+        # tip 6
+        for a in self.apply_time:
+            node = a
+            if (isinstance(node.op, T.Dot) and
+                len(set(i.dtype for i in node.inputs)) != 1):
+                print("  - You have a dot operation that has different dtype "
+                      " for inputs (%s). Make sure that the inputs have same "
+                      " dtype." % [i.type for i in node.inputs], file = file)
+                printed_tip = True
+
         if not printed_tip:
             print("  Sorry, no tip for today.", file = file)
 


theano/compile/profiling.py
```diff
@@ -1391,6 +1391,16 @@ def exp_float32_op(op):
                           " generator supported on the GPU.", file = file)
                 break
 
+        # tip 6
+        for a in self.apply_time:
+            node = a
+            if (isinstance(node.op, T.Dot) and
+                len(set(i.dtype for i in node.inputs)) == 1):
```
nouiz: I think it should be == 2.

nouiz: Or better, != 1

AnishShah:
@nouiz I have updated the PR

theano/compile/profiling.py
```diff
@@ -1391,6 +1391,18 @@ def exp_float32_op(op):
                           " generator supported on the GPU.", file = file)
                 break
 
+        # tip 6
+        for a in self.apply_time:
+            node = a
+            print("dtype", [i.dtype for i in node.inputs])
+            print("dim", [i.ndim for i in node.inputsg])
```
nouiz: inputsg -> input

This is what make travis fail. In fact, remove those 2 print. We dont, want them printed.

AnishShah: Oops. Very sorry. I was trying to run the profiling example

AnishShah:
@nouiz Updated the PR

nouiz:
thanks

Issue #2943, opened 2015-05-21
Remove useless alloc/dimshuffle/reshape caused by tensor.tile


lamblin:
The new implementation of `tensor.tile` as a function, rather than an Op, makes the previously-existing optimization useless.
Moreover, the new implementation supports more cases that could give rise to useless operations.
A start could be removing dimshuffles that move (or add or remove) broadcastable dimensions if it is inside a Reshape operation, or simplify dimshuffles inside a Reshape.

Currently identified optimizations:
- [x] Remove useless dimshuffle (#4071)
- [x] Replace some Alloc by DimShuffle (#4072)
- [x] Replace some Reshape by DimShuffle (#4073)
- [x] Replace some Reshape(Dimshuffle) by Reshape (#4580)
- [x] Uncanonicalize Dimshuffle inside Alloc, Reshape and Subtensor (#4647)

nouiz:
Can you give an example that have this problem? There is already many
dimshuffle optimization (but none related to reshape I think)

How to dimshuffle inside a reshape? a reshape do not change the order of
dimensions. I don't see how this can be done.

lamblin:
For an example, see the existing test [test_local_useless_tile](https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_opt.py#L3616).
The optimization I have in mind would be something like:
`reshape(dimshuffle{'x',0,'x',1,2,3}(var), shape) => reshape(var, shape))` since adding broadcastable dimensions did not change anything.
We could probably do even more if `var` has some broadcastable dimensions that are moved around by the dimshuffle. For instance `reshape(dimshuffle{1,0}(var), shape)` is equivalent to `reshape(var, shape)` if `var` is a `row` or a `col`.
That way, we can also get rid of the `reshape` itself if the `shape` corresponds to `var.shape`.

mohammadpz:
An example that fails is when we call ```tile(var, (1,) * var.ndim)```. The operation is useless but the computational graph looks like:
```python
 |InplaceDimShuffle{0,1} [id B] <TensorType(float32, row)> ''   3
 | |Alloc [id C] <TensorType(float32, row)> ''   2
 |   |<TensorType(float32, vector)> [id D] <TensorType(float32, vector)>
 |   |TensorConstant{1} [id E] <TensorType(int8, scalar)>
 |   |Shape_i{0} [id F] <TensorType(int64, scalar)> ''   0
[...]

PR #4062, opened 2016-02-15
test using theano tile added

mohammadpz:
Hi @lamblin ,
Regarding issue #2943, I replaced ```Tile``` op with theano ```tile```. The test seems to fail.
So, where should I change?

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -560,7 +560,7 @@ def apply_local_dimshuffle_lift(var):
 #   1 - dimshuffle all dimensions in order.
 #   2 - dimshuffle a broadcastable dimension.
 def is_dimshuffle_useless(new_order, input):
-    is_useless = False
+    is_useless = True
     if len(new_order) == input.type.ndim:
         all_broadcastable_dims = [i for (i, is_broadcastable)
                                   in enumerate(input.type.broadcastable)
@@ -573,6 +573,8 @@ def is_dimshuffle_useless(new_order, input):
             else:
                 is_useless = False
                 break
+    else:
+        is_useless = False
     return is_useless
 
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -556,6 +556,26 @@ def apply_local_dimshuffle_lift(var):
     return var
 
 
+# Checks for two types of useless dimshuffles:
+#   1 - dimshuffle all dimensions in order.
+#   2 - dimshuffle a broadcastable dimension.
+def is_dimshuffle_useless(new_order, input):
+    is_useless = False
+    if len(new_order) == input.type.ndim:
+        all_broadcastable_dims = [i for (i, is_broadcastable)
+                                  in enumerate(input.type.broadcastable)
+                                  if is_broadcastable] + ['x']
+        for i in range(input.type.ndim):
+            if (new_order[i] == i or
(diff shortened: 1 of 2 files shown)


theano/tensor/tests/test_opt.py
```diff
@@ -4016,17 +4017,15 @@ def test_local_useless_tile(self):
...
-                assert isinstance(topo[0].op, compile.DeepCopyOp)
-                f(data)
-                
-                # Check that stacktrace is copied over
-                self.assertTrue(hasattr(f.outputs[0].variable.tag, 'trace'))
-                self.assertTrue(len(f.outputs[0].variable.tag.trace)>0)
+            f_tile_op = theano.function([var], Tile(var.ndim)(var, (1,) * var.ndim), mode=mode)
+            f_tile_th = theano.function([var], tile(var, (1,) * var.ndim), mode=mode)
+            topo = f_tile_th.maker.fgraph.toposort()
+            assert len(topo) == 1
```
nouiz: The test is failing du to this assert. Now with the call to tile, this assert isn't true, as we build a graph. Just remove it.

Also, you removed the for ndim loop. Why?

lamblin: The point of that test is that we want, ultimately, to add the optimizations so that it passes.
We will take care of the for loop later.

lamblin:
The fix in the second commit seems OK, but it should be in a separate PR, so that we can merge it without the changes to the tests of `tile` that is still failing.

theano/tensor/opt.py
```diff
@@ -618,6 +618,26 @@ def local_dimshuffle_lift(node):
             copy_stack_trace(node.outputs[0], ret)
             return [ret]
 
+    # remove useless dimshuffle in general
+    # covers two types of useless dimshuffle:
+    #   1 - dimshuffle all dimensions in order
+    #   2 - dimshuffle a broadcastable dimension
```
lamblin: This looks correct, thanks!
I think it would be better if we also did it for dimshuffles caused by merging, above. We can refactor that into one function that you would call in both places.

lamblin:
Also, it would be better if the history of commits did not have some commits undoing each other.
When you are available, we can rebase to have a cleaner history.

lamblin:
flake8 says:
/home/travis/build/Theano/Theano/theano/tensor/opt.py:583:1: E303 too many blank lines (3)

There is an issue with an existing test that is now failing. Can you have a look if we can figure out what the issue is, and if the test should be updated, or if the optimization has a problem?

mohammadpz:
Actually, there is nothing wrong in the latest commit. It seems like Travis is testing something else. Is that possible?

lamblin:
The line numbers are off, probably because of merging with a newer master, but you do have 3 blank lines, starting on line 626.

lamblin:
You have one failing test, probably due to the fact that your optimization how applies in a different test, changing the resulting graph.
Can you look at that test, and try to see if it should be updated?

```
FAIL: test6 (theano.tensor.tests.test_opt.test_local_subtensor_lift)
  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_opt.py", line 2042, in test6
    assert isinstance(prog[0].op, tensor.Subtensor)
AssertionError
```

mohammadpz:
I think this PR seems to be working!

lamblin:
Cool, thanks!

Issue #129, opened 2011-10-20
Don't downcast shape dimensions to int32 in random.


nouiz:
I found that we always cast to int32 the shape in code related to ramdom. This prevent one dimensions to be higher then 2e32. This will also be silent error. Below is the email that discuss this finding, but I saw more case like that. We should find a way to cast to ~PySizet or cast to int32 only on 32 bits python.

> There is one question I have about this graph: why there is a cast to
> int32 on the shape before doing to the RandomFunction? James do you
> know the answer? That could cause problem if the shape don't fix in
> int32 in one of the dimensions.
>
> From the file raw_random.py around line 450, this was added by you
> James. I think we should modify this to do the fix only on 32 bits
> python installation. What do you think? On 32 bits computer
> Py_SizeT(or similar) is 32 bits I think, so the fix is safe, but I
> think that on python 64 bits, this Py_SizeT is 64 bits.
>
Sure, 64 bit sizes on 64 bit computers would be better.

nouiz:
Use int64 in that case. This is what we use for the infer_shape.

For safety, we should use run Theano tests on windows for that PR.

lamblin:
For safety, test on windows and with a 32-bit python

abergeron:
This issue should be closed as we don't do this anymore.

nouiz:
Why did you close this? We still do this. Just grep the file tensor/raw_random.py for int32.

lamblin:
@nouiz is right, we are still doing that for the provided `shape` argument (not for the `infer_shape` mechanism).

abergeron:
The cast there https://github.com/Theano/Theano/blob/master/theano/tensor/raw_random.py#L187 is only done when shape == ().

Unless you mean something else.

lamblin:
- https://github.com/Theano/Theano/blob/575c7a97c305b09783218a8827f1bd7f2eb0ab0a/theano/tensor/raw_random.py#L397
- https://github.com/Theano/Theano/blob/575c7a97c305b09783218a8827f1bd7f2eb0ab0a/theano/tensor/raw_random.py#L521

abergeron:
Oh, ok.

lamblin:
[...]

PR #4064, opened 2016-02-15
fixed rng_mrg int32 overflow, just throw out the error when it overflows

yaoli:
ticket  #129


yaoli:
Am I understanding correctly that all the errors are pep8 related?

bouthilx:
Almost, those two are not pep8 
https://travis-ci.org/Theano/Theano/jobs/109459069
https://travis-ci.org/Theano/Theano/jobs/109459072

yaoli:
Any idea what went wrong here?

ERROR: This test checks if the new change to MultinomialFromUniform is still compatible
  File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 859, in __call__
    outputs = self.fn()
ValueError: size must be int64
During handling of the above exception, another exception occurred:
  File "/home/travis/miniconda2/envs/pyenv/lib/python3.3/site-packages/nose-1.3.4-py3.3.egg/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/Theano/Theano/theano/sandbox/tests/test_multinomial.py", line 96, in test_n_samples_compatibility
    res = f(numpy.random.randn(20, 10))
  File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 871, in __call__
  File "/home/travis/build/Theano/Theano/theano/gof/link.py", line 314, in raise_with_op
  File "/home/travis/miniconda2/envs/pyenv/lib/python3.3/site-packages/six.py", line 658, in reraise
    raise value.with_traceback(tb)
  File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 859, in __call__
    outputs = self.fn()
ValueError: size must be int64
Apply node that caused the error: mrg_uniform{TensorType(float64, vector),inplace}(<TensorType(int32, matrix)>, Elemwise{Cast{int32}}.0)
Toposort index: 4
Inputs types: [TensorType(int32, matrix), TensorType(int32, vector)]
Inputs shapes: [(15360, 6), (1,)]
Inputs strides: [(24, 4), (4,)]
Inputs values: ['not shown', array([20], dtype=int32)]
Outputs clients: [['output'], [MultinomialFromUniform{int64}(Softmax.0, mrg_uniform{TensorType(float64, vector),inplace}.1)]]

nouiz:
The input variable of the node are in int32, but the c code seem to request
int64? Maybe in the make_node you need to upcast input to int64? I just
checked the error message.

bouthilx:
Shouldn't the size always be int64 in the graph? I thought we could assume so, maybe I got confused with `shape`?

nouiz:
shape are always of dtype int64. So I think we can assume that. But for
some reason, this didn't happened.

yaoli:
Does this failure come from testing python, or old gpu backend or the new one? I have already explicitly converted shape in the first two, but not for the third yet.

yaoli:
e.g., https://github.com/Theano/Theano/pull/4064/files#diff-2c5b25b52b0351acdefc8608171eaf6aL334

bouthilx:
The error is from this test. https://github.com/Theano/Theano/blob/master/theano/sandbox/tests/test_multinomial.py#L75

There is no GPU on travis so it can only be a failing test on the CPU.

lamblin:
Can you check what the value of `NPY_INT64` is, and the value of `PyArray_DESCR(%(size)s)->type_num`? It may be that there are two different kinds of 64-bit ints in C (for instance `long` and `long long`), which give different dtypes in NumPy.
There may be examples of C code dealing with those inconsistencies.

lamblin:
So the issue happens when loading a graph that was created with a cast to int32. Since unpickling does not call `__new__` again, there is no cast to int64 inserted, and the input is actually int32.
@nouiz what do you think about dropping compatibility and removing this test? The other option would be to accept both int32 and int64, and update the C code. I think accepting both would be better.

bouthilx:
A middle ground would be a deprecation phase for int32 upcasting. 

It's weird to not support int32 while the actual internal computations are only done for sizes small enough for int32, but at the same time if the sizes should always be int64 in the graph then I think we should be consistent and not allow int32.

nouiz:
Accepting both in the c code is good with me. Just make new graph use int64
Le 17 févr. 2016 17:34, "Xavier Bouthillier" <notifications@github.com> a
écrit :

yaoli:
The recent commit address the failure caused in unpickling graph containing int32

nouiz:
this PR have travis fail.

bouthilx:
@yaoli I don't see any size that is too large [here](https://github.com/Theano/Theano/blob/master/theano/sandbox/tests/test_rng_mrg.py#L1007-L1015). Can you debug your c code and see what are the sizes you receive and why it raises an error.

yaoli:
I left a question on github, Pascal L. points out the missing casting in C
code, I'm not sure how to do it.

yaoli:
fixed the casting issue in c code, removed casting in perform()

yaoli:
new commit, with fixes in both backends

yaoli:
buildbot timed out on one of the tests, any idea what happened?

bouthilx:
I'll restart it. Let's see if it stalls again.

yaoli:
Finally!

theano/sandbox/rng_mrg.py
```diff
@@ -651,9 +656,9 @@ def c_code(self, node, nodename, inp, out, sub):
         SYNC = "CNDA_THREAD_SYNC"
         return """
         //////// <code generated by mrg_uniform>
-
-        int odims[%(ndim)s];
-        int n_elements = 1;
+        npy_int64 max_value_supported = 2147483647;      //2^31 - 1
```
bouthilx: I find this name more explicit but I would use M1 for coherence.

bouthilx:
Great! Could you please add a test with sizes casted to int32 to ensure we still support both ([see comment](https://github.com/Theano/Theano/pull/4064#issuecomment-185455137)). I know there is a test using pickled graphs containing size of type int32 but an explicit test would be... well... more explicit. ^_^

bouthilx:
@yaoli /home/travis/build/Theano/Theano/theano/sandbox/tests/test_rng_mrg.py:1067:28: E226 missing whitespace around arithmetic operator

yaoli:
What happened to the tests? One pep8 failed almost all the tests?

bouthilx:
I know there is at least two tests failing because of flake8. For the other I can't find the error. Let's fix flake8 first and we'll see next.

yaoli:
what's the command to run pep8 test offline? I only see one failure at test_rng_mrg.py:1067.

bouthilx:
pep8 filename.py

yaoli:
pep8, great!

abergeron:
There are some small flak8 problems but there are also segmentation faults on travis in the rng tests.  That should also be fixed.

bouthilx:
@abergeron Could you please tell me where you see this. I could only find this line telling the test was killed, but I hardly understand why.

> theano.sandbox.tests.test_rng_mrg.test_overflow_cpu_support_int32 ... /home/travis/build.sh: line 45:  9924 Killed                  theano-nose -v $PART

abergeron:
That line is a clear indicator that something went wrong.  It means that the python process crashed while running the tests.

bouthilx:
So you inferred a seg fault from this line?

yaoli:
I just did a new PL, and maybe let's wait for this one, and see what
happens?

yaoli:
PR, not PL, lol

bouthilx:
I ran it locally for debugging. ```test_overflow_cpu_support_int32``` takes up to 4GB of RAM, that's way too much for nothing. 

There is multiple tests for `rng_mrg_overflow` and some redundancies in the code. It would be better if they are all set in a single class with general methods to avoid duplication. int32 test for instance could share the same as an int64 version of the test (which is more or `less test_overflow_cpu` right now). You can then inherit from the same class to make a GPU version of the tests.

abergeron:
Ok so maybe it's the OOM that triggered on travis rather than a segfault.  But I agree that 4G is too much.  If should not go over about 1G.

yaoli:
2**30 float32 numbers roughly takes about 4G, and we are suppose to test the case where it overflows. So kindof impossible to shrink mem.

bouthilx:
If I understand correctly, too big shapes will trigger a ValueError right before the samples would be allocated in memory. Thus, we don't have a memory problem with too large values, but those small enough yet still very large are allocated and make Travis very angry.

yaoli:
the test is failing again due to some other perks:

Warning, treated as error:
../NEWS.txt:1: WARNING: Duplicate explicit target name: "news".

bouthilx:
@lamblin Could it be because of this [commit]( https://github.com/Theano/Theano/commit/4e7f550d28c010b996a157c8239546aecaca1e9f)?

lamblin:
@bouthilx probably, I'll check.

lamblin:
I have a fix in #4152, hopefully that will be enough.

bouthilx:
Thanks a lot Pascal!

bouthilx:
PR was merge. I relaunched the failing tests.

bouthilx:
Great! @yaoli Could you please reorganize the test functions inside two classes, one for CPU one for GPU. I'll make sure the GPU tests pass and the PR will be ready for merge.

yaoli:
Why not just put them into a single class?

On Tue, Mar 1, 2016 at 5:22 PM, Xavier Bouthillier <notifications@github.com

yaoli:
And frankly, I don't see the point of wrapping these into classes, as I'd
prefer things as simple and as independent as possible.

bouthilx:
You can use functions if you insist. There is no particular advantages for classes here beside aesthetic organization. Nevertheless there is several code duplications that could be avoided. There is a similar loop 5 times, it could be 

```python
def loop_rng_mrg_overflow(fct, sizes, should_overflow):
    for size in sizes:
        f = theano.function([], fct(size=x), mode=mode)
        if should_overflow:
            assert_raises(ValueError, f)
        else:
            f()
```

with the use of ```functools.partial()``` for ```GPUA_mrg_uniform.new()```.

yaoli:
refactored the tests, local tests passed, waiting the buildbots.

bouthilx:
I find most of the tests clearer now.  

Why did you add the new_gpu code inside the helper function? Was there a problem with using `functools.partial(rng.GPUA_mrg_uniform.new, rstate=rstate, ndim=None, dtype='float32')`?

It would simplify things if you pass a function rather than `rng` to `rng_mrg_overflow`.

bouthilx:
@yaoli There is one flake8 error.

yaoli:
I think I'm using a lower version of pep8, which passes all tests locally.
Let me try to fix that.

On Wed, Mar 2, 2016 at 4:12 PM, Xavier Bouthillier <notifications@github.com

bouthilx:
@yaoli I just noticed it touches the new_gpu code I was talking about. We should talk about this before fixing the code if you end up modifying it again. Have you seen this [comment](https://github.com/Theano/Theano/pull/4064#issuecomment-191337578)?

yaoli:
I did see that. I don't understand how 'partial' works, so I went for the
current solution.

Still I updated my current pep8 to 1.7.0, and it does not give any local
error. What's the version of pep8 running on github?

On Wed, Mar 2, 2016 at 4:20 PM, Xavier Bouthillier <notifications@github.com

bouthilx:
I see. `functools.partial` acts like a python decorator ([doc](https://docs.python.org/2/library/functools.html#functools.partial)). It saves the given arguments so that only the missing ones need to be given to the function. Here's the code

```python
def partial(func, *args, **keywords):
    """New function with partial application of the given arguments
    and keywords.
    """
    if hasattr(func, 'func'):
        args = func.args + args
        tmpkw = func.keywords.copy()
        tmpkw.update(keywords)
        keywords = tmpkw
        del tmpkw
        func = func.func

    def newfunc(*fargs, **fkeywords):
        newkeywords = keywords.copy()
        newkeywords.update(fkeywords)
        return func(*(args + fargs), **newkeywords)
    newfunc.func = func
    newfunc.args = args
    newfunc.keywords = keywords
    return newfunc
```

In our case we would use it this way

```python
# for rng.uniform
fct = rng.uniform  # No need for partial
rng_mrg_overflow(fct, sizes, mode, should_raise_error)

# for rng.GPUA_mrg_uniform
fct = functools.partial(rng.GPUA_mrg_uniform.new, rstate=rstate, ndim=None, dtype='float32')
rng_mrg_overflow(fct, sizes, mode, should_raise_error)
```

Please tell me if anything is unclear so that I can help you beforehand!

bouthilx:
pep8 version seams to be 1.6.2 (see [here](https://github.com/Theano/Theano/blob/master/.travis.yml#L34-L35)).

bouthilx:
You can configure pep8 to ignore some codes. Do you have something in `~/.config/pep8` or `~/.pep8`?

yaoli:
I don't have anything in ~/.config/pep8 or ~/.pep8, I installed pep8 1.6.2,
but still no error reported locally.

On Wed, Mar 2, 2016 at 4:50 PM, Xavier Bouthillier <notifications@github.com

bouthilx:
What if you try `pep8 --select=E121 yourfilename.py`?

yaoli:
yaoli@eos15:~/Lib/Theano/theano/sandbox/tests$ pep8 --select=E121
test_rng_mrg.py
test_rng_mrg.py:1050:15: E121 continuation line under-indented for hanging
indent

What the hell? So by default, it ignores E121?

On Wed, Mar 2, 2016 at 5:09 PM, Xavier Bouthillier <notifications@github.com

bouthilx:
Seams to. Some errors that aren't very consensual are ignored by default.

yaoli:
OK, all fixed, hope this will go through.

On Wed, Mar 2, 2016 at 5:16 PM, Xavier Bouthillier <notifications@github.com

yaoli:
All is well it seems.

bouthilx:
@yaoli Sorry for the late final review. You need to rebase it before I can merge. 

@lamblin Would you like to take a final look?

lamblin:
Looks OK to me, thanks!

yaoli:
just rebased

bouthilx:
@yaoli Did you rebase against your repo's master or Theano's master? Github says there is still conflicts.

yaoli:
github shows there are some conflicts after I pushed the rebased one, don't
know what's going on there

yaoli:
based on theano master

yaoli:
Can you take a look at the conflicts? I cannot see them.

bouthilx:
OK I'll take a look as soon as possible this afternoon. A power outage in half of my apartment is slowing me down. -_-

theano/sandbox/rng_mrg.py
```diff
@@ -540,7 +545,7 @@ def new(cls, rstate, ndim, dtype, size):
         if ndim is None:
             ndim = get_vector_length(v_size)
         op = cls(CudaNdarrayType((False,)*ndim))
```
bouthilx:
```python
<<<<<<< HEAD
        op = cls(CudaNdarrayType((False,) * ndim))
        return op(rstate, cast(v_size, 'int32'))
=======
        op = cls(CudaNdarrayType((False,)*ndim))
        return op(rstate, v_size)
>>>>>>> e5ac0e8... fixed rng_mrg int32 overflow, just throw out the error when it overflows
```

theano/sandbox/rng_mrg.py
```diff
@@ -332,7 +332,7 @@ def new(cls, rstate, ndim, dtype, size):
         if ndim is None:
             ndim = get_vector_length(v_size)
         op = cls(TensorType(dtype, (False,)*ndim))
```
bouthilx:
```python
<<<<<<< HEAD
        op = cls(TensorType(dtype, (False,) * ndim))
        return op(rstate, cast(v_size, 'int32'))
=======
        op = cls(TensorType(dtype, (False,)*ndim))
        return op(rstate, v_size)
>>>>>>> e5ac0e8... fixed rng_mrg int32 overflow, just throw out the error when it overflows
```

theano/sandbox/rng_mrg.py
```diff
@@ -782,7 +792,7 @@ def new(cls, rstate, ndim, dtype, size):
         if ndim is None:
             ndim = get_vector_length(v_size)
         op = cls(GpuArrayType(dtype, (False,)*ndim))
```
bouthilx:
```python
<<<<<<< HEAD
        op = cls(GpuArrayType(dtype, (False,) * ndim))
        return op(rstate, cast(v_size, 'int32'))
=======
        op = cls(GpuArrayType(dtype, (False,)*ndim))
        return op(rstate, v_size)
>>>>>>> e5ac0e8... fixed rng_mrg int32 overflow, just throw out the error when it overflows
```

bouthilx:
I identified 3 conflicts. You didn't have it when you rebased because your local Theano's master was behind by a few commits.

bouthilx:
@yaoli Woh! :open_mouth:  What happened? Did you really edited that much files to be able to rebase or you accidentally included many commits from master to your branch?

yaoli:
damn it, I made a mistake, now it's fixed.

PR #4066, opened 2016-02-15
GpuBatchedDot: streams implementation (WIP)

### LINKED ISSUE #3512: Merge cuda batched gemm and faster version for big size
In https://github.com/Theano/Theano/pull/1698/files#diff-43102a08f907f49ba6fbb467e08e04d8

we have another implementation of batched gemm that is faster for big matrices.

This ticket is about:

1) merging that in the batched gemm from cublas. Have a parameter that default to 128, if that size of smaller, use the cublas version, if bigger, use the other implementation. Do some speed tests to make sure 128 is the right number. This could change depending of cuda version. Test at least with 7.90 and 7.5.

cooijmanstim:
Fixes #3512. I switch between the two implementations based on the number of multiplications per `cublasSgemm` call; if that is large then the streams implementation is used. The next step would be to tweak the threshold according to measurements.

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -18,6 +18,9 @@ from theano.tensor import as_tensor_variable
 class GpuBatchedDot(GpuOp):
     __props__ = ()
 
+    def __init__(self, stream_threshold=128):
+        self.stream_threshold = stream_threshold
+
     def make_node(self, inp1, inp2):
         inp1 = gpu_contiguous(as_cuda_ndarray_variable(inp1))
         inp2 = gpu_contiguous(as_cuda_ndarray_variable(inp2))
@@ -39,6 +42,7 @@ class GpuBatchedDot(GpuOp):
         bx, by = input_names
         bz, = output_names
         fail = sub['fail']
+        threshold = self.stream_threshold
         return ("""
         float alpha = 1.0;
         float beta = 0.0;
@@ -59,7 +63,7 @@ class GpuBatchedDot(GpuOp):
         y_dim2 = CudaNdarray_HOST_DIMS(%(by)s)[2];
 
         // use parallel cublasSgemm calls rather than cublasSgemmBatched for large products
-        bool use_cublas_sgemm_batched = x_dim1 * x_dim2 * y_dim2 < 128 * 128 * 128;
+        bool use_cublas_sgemm_batched = x_dim1 * x_dim2 * y_dim2 < %(threshold)s * %(threshold)s * %(threshold)s;
 
         if (x_dim0 != y_dim0)
         {
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -53,7 +53,7 @@ class GpuBatchedDot(GpuOp):
         // use parallel cublasSgemm calls rather than cublasSgemmBatched for large products
         // (compute products in double because they can be large and we don't need to be exact)
         bool use_cublas_sgemm_batched = (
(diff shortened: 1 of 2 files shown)


theano/sandbox/cuda/blas.py
```diff
@@ -171,19 +187,130 @@ def c_code(self, node, name, input_names, output_names, sub):
...
+                                 status,  cublasGetErrorString(status), unit, N,
+                                 Nx[0], Nx[1], Nx[2], Sx[0], Sx[1], Sx[2],
+                                 Ny[0], Ny[1], Ny[2], Sy[0], Sy[1], Sy[2],
+                                 Nz[0], Nz[1], Nz[2], Sz[0], Sz[1], Sz[2]);
+                    %(fail)s;
+                }
+
+                x += Sx[0]; y += Sy[0]; z += Sz[0];
+            };
+            CNDA_THREAD_SYNC;
```
abergeron: Since you synchronize the streams just after, this may not be necessary.

abergeron:
Apart from my small comment, what do you think is missing?  It seems complete to me.

Also the tests pass locally.

cooijmanstim:
I wasn't sure if the comment meant I should remove the thread synchronization or not ("may not be necessary" sounds like it needs confirmation).

Also, in #3512 Fred said:
> Do some speed tests to make sure 128 is the right number. This could change depending of cuda version. Test at least with 7.90 and 7.5.
Not sure how to go about timing -- should I use timeit or the Theano profiler or something else?

cooijmanstim:
During timing I discovered that the switching based on the threshold didn't always work. I've fixed two problems there and now the tests segfault. So there's some work left to do on this PR.

nouiz:
For the speed test I would do it only with 7.5. I don't think it is super
useful to spend time on older verion.

cooijmanstim:
Measurements:
linear
I ran these with a Titan X, not sure what cuda version. If you want me to rerun with a specific version please tell me how to configure it. Based on the results I'll put the threshold at 650.
For the record here's the code I used: https://gist.github.com/cooijmanstim/4352afc91e7e1681a979

cooijmanstim:
Also, the segfault was due to not switching back to the default (NULL) stream.

abergeron:
I'm ok to merge this when the build is ok.

theano/sandbox/cuda/blas.py
```diff
@@ -39,79 +42,83 @@ def c_code(self, node, name, input_names, output_names, sub):
...
+        const int* Ny = CudaNdarray_HOST_DIMS(%(by)s);
+        int Nz[3] = {0};
 
-        y_dim0 = CudaNdarray_HOST_DIMS(%(by)s)[0];
-        y_dim1 = CudaNdarray_HOST_DIMS(%(by)s)[1];
-        y_dim2 = CudaNdarray_HOST_DIMS(%(by)s)[2];
+        // use parallel cublasSgemm calls rather than cublasSgemmBatched for large products
+        // (compute products in double because they can be large and we don't need to be exact)
+        bool use_cublas_sgemm_batched = (
+            double(Nx[1]) * double(Nx[2]) * double(Nx[2]) <
```
abergeron: The last NX[2] should probably be Nx[3].

abergeron: Or maybe that should be Nx[0], Nx[1], Nx[2].

cooijmanstim: Actually I intended Nx[1], Nx[2], Ny[2]. It's the number of multiplications per instance in the batch.

PR #4067, opened 2016-02-16
Make DebugMode handle a special version of perform.

abergeron:
Also use this to have AllocEmpty return memory with a fixed value to avoid tripping up checks by accident.


abergeron:
I've added documentation.

abergeron:
I did the paragraph rewrite.

abergeron:
This passes travis now.

Issue #4047, opened 2016-02-12
Make docstring of AssertOp appear in the generated doc


lamblin:
Follow-up to #2720.
Use `autodoc` or something similar so that `AssertOp`'s documentation appears somewhere on http://deeplearning.net/software/theano/library/tensor/

nouiz:
The assert op is in the file tensor/opt.py. Add a reference in its docstring to the pdbbreakpoint op that people probably want to use instead.

hantek:
Hi Fred, I have created a PR at #4069

nouiz:
done.

PR #4069, opened 2016-02-16
Add doc for opt.py

hantek:
a fix for #4047


theano/tensor/opt.py
```diff
@@ -2071,6 +2073,7 @@ class Assert(T.Op):
 
     Examples
     --------
+    import theano
     T = theano.tensor
```
nouiz: This example don't show well on the generated doc. It show  all the code on a single line.

nouiz:
ok to merge when travis pass.

PR #4070, opened 2016-02-16
Fix the LogSoftmax tests in DebugMode.

abergeron:


theano/tensor/nnet/nnet.py
```diff
@@ -751,7 +751,9 @@ def local_logsoftmax(node):
             isinstance(node.inputs[0].owner.op, Softmax)):
         inVars = node.inputs[0].owner.inputs[0]
         new_op = LogSoftmax()
-        return [new_op(inVars)]
+        ret = new_op(inVars)
+        ret.values_eq_approx = values_eq_approx_remove_inf
```
nouiz: I think it is better to set it under ret.tag.values_eq_approx. Tag are keep when cloning. Also this prevent to cluter the var itself. The interface you use is the old one.

theano/tensor/nnet/nnet.py
```diff
@@ -784,7 +786,9 @@ def local_logsoftmax_grad(node):
         if grads.broadcastable[1] and not sm.broadcastable[1]:
             grads = tensor.alloc(grads, grads.shape[0], sm.shape[1])
 
-        return [grads - tensor.sum(grads, axis=1, keepdims=True) * sm]
+        ret = grads - tensor.sum(grads, axis=1, keepdims=True) * sm
+        ret.values_eq_approx = values_eq_approx_remove_nan
```
nouiz: idem

nouiz:
ok to merge when travis pass.

abergeron:
That was mostly copy-and-pasted from some other test.  I'll update it to the new interface.

nouiz:
ok to merge when travis pass.

PR #4076, opened 2016-02-17
Don't call eval

nouiz:


abergeron:
I approve.

Ok when travis passes.

abergeron:
Also ok.

nouiz:
@abergeron I amended the last commit, can you review again? Now travis pass.

PR #4083, opened 2016-02-17
Fix the cnmem print percent message

caglar:
cnmem usage message was wrong.


nouiz:
ok to merge when travis pass.

caglar:
fyi @nouiz tests seem to pass.

nouiz:
thanks

PR #4084, opened 2016-02-17
cuDNN header / library paths fix

### LINKED ISSUE #4082: Inconsistent sourcing of cuDNN header and library files in CUDA backend
The `config.dnn.include_path` and `config.dnn.library_path` options do not seem to be currently getting passed to the `nvcc` compiler when compiling cuDNN ops using the CUDA backend. In particular the `DnnVersion` op in `cuda/__init__.py` and `DnnBase` op in `cuda/dnn.py` do not add `config.dnn.include_path` to the list of paths returned by the `c_header_dirs` method or `config.dnn.library_path` to the list of paths returned by `c_lib_dirs` unlike the corresponding ops in `gpuarray/dnn.py`.

I'm not sure if this is intentional, i.e. the `config.dnn.*_path` options are only intended for use with the GpuArray backend? 

At the moment however the `dnn_available` function in `cuda/__init__.py` does add `config.dnn.include_path`  and `config.dnn.library_path` to the compile string used to check if cuDNN is available. This means if the `config.dnn.*_path` options are set to valid paths including the requisite header/library files, `dnn_available.avail` is set to `True` when initially checking the cuDNN version with `dnn_version` in `use`. However later attempts to compile cuDNN functions are not passed these directories, therefore will potentially either fail to compile completely, or if there is a different cuDNN library version available in a system wide directory can create unexpected results. 

For example in my case an older v2 release of cuDNN is installed in the `/usr/include` / `usr/lib64` while I have a newer v3 release in a local user directory which `config.dnn.library_path` and `config.include_path` point to - when compiling graphs containing CUDA backend cuDNN ops this causes a series of `ImportError: ('The following error happened while compiling the node  ... : undefined symbol: cudnnGetVersion')` exceptions to be raised as the cuDNN v2 files in `/usr/*` being linked against do not define `cudnnGetVersion` but the check in `dnn_available` to make sure the version number is at least 3007 is done with the up to date v3 files in the local user directory.

To fix this issue it seems that as well as adding (or extending the existing) `c_header_dirs` and `c_lib_dirs` methods in the `DnnBase` and `DnnVersion` ops for the CUDA backend it is also necessary to edit how the compiler command string is constructed in `nvcc_compiler.py` by changing

```
lib_dirs = std_lib_dirs() + lib_dirs
```

to 

```
lib_dirs = lib_dirs + std_lib_dirs()
```

so that any library directories passed in `lib_dirs` are checked before those returned by `std_lib_dirs()`. This precedence of user over standard options is already used for the `include_dirs`. However I'm not sure if the library directory order was intentionally set the opposite way around for some reason and so if switching the order like this will break other things? 

I can submit a PR which makes the suggested changes but thought it would be best to check if any of this behaviour is intentional first.

matt-graham:
## Reason for PR

This is a proposed fix #4082.

The PR causes  `config.dnn.include_path` and `config.dnn.library_path` to be included in the compile commands for CUDA backend cuDNN ops by adding them to the lists returned by the `c_header_dirs` and `c_lib_dirs` methods of the cuDNN related ops in `sandbox/cuda/__init__.py` and `sandbox/cuda/dnn.py`.

To make sure the user specified library paths are given greater precedence by the compiler it also switches the order in which the `lib_dirs` list is constructed in `NVCC_compiler.compile_str` in `sandbox/cuda/nvcc_compiler`, with the user specified library paths put at the start of the list (rather than being added after the standard libraries returned by `std_lib_dirs()` as currently). This is already the behaviour for the `include_dirs`. For consistency I also switched the order for `libs` analagously as well though this isn't actually needed to fix #4082.

### DIFF
--- a/theano/sandbox/cuda/nvcc_compiler.py
+++ b/theano/sandbox/cuda/nvcc_compiler.py
@@ -214,7 +214,7 @@ class NVCC_compiler(Compiler):
         if os.path.abspath(os.path.split(__file__)[0]) not in include_dirs:
             include_dirs.append(os.path.abspath(os.path.split(__file__)[0]))
 
-        libs = std_libs() + libs
+        libs = libs + std_libs()
         if 'cudart' not in libs:
             libs.append('cudart')
 
--- a/theano/sandbox/cuda/nvcc_compiler.py
+++ b/theano/sandbox/cuda/nvcc_compiler.py
@@ -218,7 +218,7 @@ class NVCC_compiler(Compiler):
         if 'cudart' not in libs:
             libs.append('cudart')
 
-        lib_dirs = std_lib_dirs() + lib_dirs
+        lib_dirs = lib_dirs + std_lib_dirs()
         if any(ld == os.path.join(cuda_root, 'lib') or
                ld == os.path.join(cuda_root, 'lib64') for ld in lib_dirs):
             warnings.warn("You have the cuda library directory in your "
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -2122,8 +2122,8 @@ class GCC_compiler(Compiler):
             preargs = list(preargs)
 
         include_dirs = include_dirs + std_include_dirs()
-        libs = std_libs() + libs
-        lib_dirs = std_lib_dirs() + lib_dirs
+        libs = libs + std_libs()
+        lib_dirs = lib_dirs + std_lib_dirs()
 
         cppfilename = os.path.join(location, 'mod.cpp')
         with open(cppfilename, 'w') as cppfile:
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -355,8 +355,14 @@ class DnnVersion(GpuOp):
     def c_headers(self):
         return ['cudnn.h']
 
+    def c_header_dirs(self):
+        return [config.dnn.include_path]
+
     def c_libraries(self):
         return ['cudnn']
+
+    def c_lib_dirs(self):
+        return [config.dnn.library_path]
 
     def c_support_code(self):
         return """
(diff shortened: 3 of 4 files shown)


nouiz:
Just self note, this is already done like this in the new GPU back-end.

thanks

matt-graham:
@nouiz Also updated `gof/cmodule.py`.

nouiz:

ebenolson:
Hi, after this change when I `import theano` I am getting `UserWarning: You have the cuda library directory in your lib_dirs. This has been known to cause problems and should not be done.`

I am not sure what this warning means - can someone help me determine if this is a bug or I have something configured wrong?

matt-graham:
Hi @ebenolson,

Sorry for the delay in responding.

I think this is happening as the default for `config.dnn.library_path` is `[CUDA root]/lib64` so if `config.dnn.library_path` is not specified, when any cuDNN based ops are compiled (including the `DnnVersion` op called during `import theano` to check installed cuDNN version is compatible) the subsequent check to see if the CUDA root directory appears in `lib_dirs` in `nvcc_compiler.py` ends up flagging up the warning above. 

If `NVCC_compiler` automatically checks for libraries in the CUDA root `lib[64]` directory then possibly a way to prevent this warning being raised would be to just set `config.dnn.library_path` to be `None` or an empty string by default. Otherwise not sure what the best fix is however as if this is not the case the CUDA root path does need to be specified as a library path in the compile string to make sure the cuDNN library is found there if present.

@nouiz  @lamblin Any ideas as to whether this could be the cause and if removing the default would work as a fix?

Thanks

theano/sandbox/cuda/nvcc_compiler.py
```diff
@@ -214,11 +214,11 @@ def compile_str(
         if os.path.abspath(os.path.split(__file__)[0]) not in include_dirs:
             include_dirs.append(os.path.abspath(os.path.split(__file__)[0]))
 
-        libs = std_libs() + libs
+        libs = libs + std_libs()
         if 'cudart' not in libs:
             libs.append('cudart')
 
-        lib_dirs = std_lib_dirs() + lib_dirs
```
nouiz: I think the real solution is to remove empty string from the list of header dir and library dir list. So don't merge this PR now.

PR #4088, opened 2016-02-18
fix potential nan

dementrock:
The `local_dot22_to_ger_or_gemv` optimizer uses AllocEmpty for some cases. Usually this is fine, as it is multiplied by zero. However, on rare cases the allocated memory might contain nan values, in which case the product will be nan instead of zero. I've somehow encountered this rare case in my experiments.

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -1918,19 +1918,19 @@ def local_dot22_to_ger_or_gemv(node):
             # x and y are both vectors so this qualifies for a sdot / ddot
             # TODO: Theano doesn't have a sdot, but gemv is better than _dot22
             xv = x.dimshuffle(1)
-            zeros = T.AllocEmpty(x.dtype)(1)
+            zeros = T.zeros([1], x.dtype)
             rval = gemv_no_inplace(zeros, one, y.T, xv, zero)
             return [rval.dimshuffle('x', 0)]
         if xb[0] and not yb[0] and not yb[1]:
             # x is vector, y is matrix so try gemv
             xv = x.dimshuffle(1)
-            zeros = T.AllocEmpty(x.dtype)(y.shape[1])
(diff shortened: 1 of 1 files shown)


theano/tensor/blas.py
```diff
@@ -1918,19 +1918,19 @@ def local_dot22_to_ger_or_gemv(node):
             # x and y are both vectors so this qualifies for a sdot / ddot
             # TODO: Theano doesn't have a sdot, but gemv is better than _dot22
             xv = x.dimshuffle(1)
-            zeros = T.AllocEmpty(x.dtype)(1)
+            zeros = T.zeros([1], x.dtype)
```
nouiz: This will slow down some computation. The problem happen only for some blas implementation. Others won't do the mul by 0, so they don't have the problem.

We have a work around that will detect if the blas used need this 0s of the the input or not. This allow to use the slow version only if needed. So a better fix would be to reuse that. It isn't very hard, can you do it?

in the file tensor/blas_c.py, there is a function check_force_gemv_init(). Can you move it to this file and use it here to choose to zeros or not?

Also, can you test it on your side? It is hard for us to reproduice this and test it locally.

thanks

abergeron:
The current gemv code already checks for blas library that use the memory with beta=0 and works around it.  I don't see what is there to do here.

nouiz:
I think this should stay open. That was what my comment was triing to tell. We missing one case. We do this just for the CGemv op, not for the Gemv op. The numpy binding can have the same problem the our c binding.

abergeron:
Then we should fix the python code to do the same kind of check and fill the buffer with zeros if required.

abergeron:
That problem should be addressed by #4150

nouiz:
@dementrock can you try that PR to make sure it fix your problem? We think it is good, but it is hard for us to test it in real life as we would need a very specific setup.

PR #4090, opened 2016-02-18
Try blas flags with rpath if they don't work.

nouiz:
This is needed on Helios cluster.


nouiz:
fixed

abergeron:
Ok when travis passes.

PR #4093, opened 2016-02-19
Check blas flags on travis [DO NOT MERGE]

abergeron:


nouiz:
I would add this to the output of normal travis. So that in the futur, if
we need the information, it will be easily available.

abergeron:
Ok, but for now, I'll just try to see if it's empty and fix it if it is.

abergeron:
There is indeed no valid blas on travis right now.  I've fixed the problem in #4095

PR #4094, opened 2016-02-19
Add lifter for IfElse.

abergeron:


theano/sandbox/gpuarray/opt.py
```diff
@@ -525,6 +525,16 @@ def local_gpu_pdbbreakpoint_op(node):
 
 
 @register_opt('fast_compile')
+@op_lifter([IfElse])
+def local_gpua_lazy_ifelse(node, context_name):
```
nouiz: this need test. So I'm killing travis

theano/sandbox/gpuarray/opt.py
```diff
@@ -525,6 +525,16 @@ def local_gpu_pdbbreakpoint_op(node):
...
+    if node.op.gpu:
+        return
+    c = nodes.inputs[0]
+    outs = [as_gpuarray_variable(v, context_name) for v in node.inputs[1:]]
```
nouiz: strange name outs, they are inputs. inps or gpu_inps would be better.

nouiz:
ok to merge when travis pass.

PR #4095, opened 2016-02-19
Move leftover flags

abergeron:
It seems a few flags were forgotten in #4042.

As a bonus this fixes linking with mkl which apparently never worked before because it require a strange hack.  This might speedup travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,7 @@ script:
   - cd $(python -c 'import theano; import os; print(os.path.split(theano.__file__)[0])')
   - echo "$PART"
   - cd -; cd Theano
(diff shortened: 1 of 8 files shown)


nouiz:
Also, can you add in travis a print of Theano flags? This will help make sure we support it.

abergeron:
I've added a printout of theano.config.

nouiz:
I have an idea why it was working in the past and not now. My guess is that now, we test if we link with guessed blas before using it. the -lm wasn't in it, but was requested, so now the test was failing without this PR. But before, as we didn't test, when we used it, we linked again python that already link with -lm, so it was probably included by default.

PR #4097, opened 2016-02-19
GpuAdvancedSubtensor created

### LINKED ISSUE #3166: Convert AdvancedSubtensor to GpuAdvancedSubtensor1 on the GPU
@lamblin you told me you know how to make this. Can you describe it here so that someone else have a change to make it?

yingzha:
Fixes #3166.

@abergeron,  there some issues according to this op.
1. It doesn't support `None`(np.newaxis) to be included in the index list and I don't figure it out why.
2. As I wrote in the test case, when I tried to call f(idx1_val, idx2_val), it got a TypeError saying `Variables do not support boolean operations`. I traceback the error but have no idea right now.
Could you take a look at my code?

### DIFF
--- a/theano/sandbox/gpuarray/tests/test_subtensor.py
+++ b/theano/sandbox/gpuarray/tests/test_subtensor.py
@@ -7,6 +7,7 @@ from theano.tensor.tests import test_subtensor
 
 from ..basic_ops import HostFromGpu, GpuFromHost
 from ..subtensor import (GpuIncSubtensor, GpuSubtensor,
+                         GpuAdvancedSubtensor,
                          GpuAdvancedSubtensor1,
                          GpuAdvancedIncSubtensor1)
 from ..type import gpuarray_shared_constructor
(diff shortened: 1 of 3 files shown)


nouiz:
I didn't look at the code. I let @abergeron do this, but I just want to tell that travis is failing.

theano/sandbox/gpuarray/tests/test_subtensor.py
```diff
@@ -55,3 +56,22 @@ def test_advinc_subtensor1():
         rep = xval.copy()
         rep[[0, 2]] += yval
         assert numpy.allclose(rval, rep)
+
+
+def test_adv_subtensor():
```
nouiz: To have more complete test by having the CPU tests run on the GPU add this:

```
class G_subtensor(test_subtensor.TestAdvancedSubtensor):
     def shortDescription(self):
        return None

    def __init__(self, name):
        test_subtensor.T_subtensor.__init__(
            self, name,
            shared=gpuarray_shared_constructor,
            sub=GpuAdvancedSubtensor,
            inc_sub=GpuIncSubtensor,
            adv_sub1=GpuAdvancedSubtensor1,
            adv_incsub1=tensor.AdvancedIncSubtensor,
            mode=mode_with_gpu,
            # avoid errors with limited devices
            dtype='float32',
            ignore_topo=(HostFromGpu, GpuFromHost,
                         DeepCopyOp))
        # GPU opt can't run in fast_compile only.
        assert self.sub == GpuAdvancedSubtensor
```

abergeron:
Superseeded by #4396

Issue #3637, opened 2015-11-12
Confusion matrix


nouiz:
This question happen frequently on the mailing list, maybe we should add it in Theano directly:

[[theano-users] Show the confusion matrix](https://groups.google.com/d/topic/theano-users/MMuXRzYPePA/discussion)

meetshah1995:
@nouiz - I went through the google thread , is anyone working on it ? If not I'd like to add a PR with the implementation, example and tests.

nouiz:
Nobody work on this to my knowledge.

thanks

meetshah1995:
@nouiz in which file do I add the method to show the confusion matrix ? Sorry for being silly, I am new with this.

lamblin:
Closed by #4810.

Issue #1916, opened 2014-06-13
Make theano.tensor.* interface more similar to numpy.*


nouiz:
Todo list, what need to be done.
Incomplete list of missing implementation.:
- [x] unique gh-3009
- [ ] pad gh-1216, gh-1957
- [ ] argwhere
- [ ] indices (requested by a user gh-2433) 
- [ ] einsum: this URL could help: http://stackoverflow.com/questions/30305891/convert-einsum-computation-to-dot-product-to-be-used-in-theano We should build the corresponding Theano graph and not wrap the numpy method.

vdumoulin:
Here's the same list, with some false positives removed:
```
numpy.*                   theano.tensor.* interface compatible
abs                       no              no
absolute                  no              no
add                       yes             lacks the parameter ‘out’
alen                      no              no
all                       yes             lacks the parameter ‘out’
allclose                  no              no
alltrue                   no              no
alterdot                  no              no
amax                      no              no
amin                      no              no
angle                     yes             lacks the parameter ‘deg’
any                       yes             lacks the parameter ‘out’
append                    no              no
apply_along_axis          no              no
apply_over_axes           no              no
arange                    yes             yes
arccos                    yes             lacks the parameter ‘out’
arccosh                   yes             lacks the parameter ‘out’
arcsin                    yes             lacks the parameter ‘out’
arcsinh                   yes             lacks the parameter ‘out’
arctan                    yes             lacks the parameter ‘out’
arctan2                   yes             lacks the parameter ‘out’
arctanh                   yes             lacks the parameter ‘out’
argmax                    yes             added the param ‘keepdims’
argmin                    yes             added the param ‘keepdims’
argpartition              no              no
argsort                   yes             yes
argwhere                  no              no
around                    no              no
array                     no              no
array_equal               no              no
array_equiv               no              no
array_split               no              no
asmatrix                  no              no
asscalar                  no              no
atleast_1d                no              no
atleast_2d                no              no
atleast_3d                no              no
average                   no              no
bartlett                  yes             yes
bincount                  yes             yes
bitwise_and               yes             lacks the parameter ‘out’
bitwise_not               yes             lacks the parameter ‘out’
bitwise_or                yes             lacks the parameter ‘out’
bitwise_xor               yes             lacks the parameter ‘out’
blackman                  no              no
bmat                      no              no
broadcast                 no              no
broadcast_arrays          no              no
byte                      no              no
byte_bounds               no              no
bytes_                    no              no
[...]

PR #4100, opened 2016-02-20
Added confusion matrix op, issue #3637

RadhikaG:
Hi, I've added the confusion matrix operation from https://github.com/Theano/Theano/issues/3637 to `theano/tensor/nnet/nnet.py`, since it's frequently used in Machine Learning and I assumed this to be the correct file. Please let me know if I need to change the location of the operation, and I'll accordingly add tests for the op to the appropriate file.

Thanks!

### DIFF
--- a/theano/tensor/nnet/__init__.py
+++ b/theano/tensor/nnet/__init__.py
@@ -16,7 +16,8 @@ from .nnet import (
     graph_merge_softmax_with_crossentropy_softmax, h_softmax,
     logsoftmax, logsoftmax_op, prepend_0_to_each_row, prepend_1_to_each_row,
     prepend_scalar_to_each_row, relu, softmax, softmax_grad, softmax_graph,
-    softmax_op, softmax_simplifier, softmax_with_bias, elu)
+    softmax_op, softmax_simplifier, softmax_with_bias, elu,
+    confusion_matrix)
 from . import opt
 from .conv import ConvOp
 from .Conv3D import *
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -2381,7 +2381,7 @@ def confusion_matrix(actual, pred):
        Actual  |
                |
 
-    order : Order of entries in terms of original data
+    order : 1-d array of order of entries in rows and columns
 
     Examples
     --------
@@ -2398,12 +2398,10 @@ def confusion_matrix(actual, pred):
             [2, 1, 0],
             [0, 0, 1]]), array([ 0.,  1.,  2.])]
     """
-    if actual.type.ndim != 1:
+    if actual.ndim != 1:
         raise ValueError('actual must be 1-d tensor variable')
-    if pred.type.ndim != 1:
+    if pred.ndim != 1:
         raise ValueError('pred must be 1-d tensor variable')
-    if actual.shape[0] != pred.shape[0]:
-        raise ValueError('actual and pred must have the same length')
 
     order = extra_ops.Unique(False, False, False)(tensor.concatenate([actual, pred]))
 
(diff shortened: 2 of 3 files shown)


nouiz:
Thanks for the PR. I think we should not make a new OP for this, but build a Theano graph of existing ops. This would allow them to be optimized by Theano (like using c code) and to work on GPU. The GPU isn't for the speed, but to remove the transfer to/from GPU.

I think you can do so with mostly similar code then what is in the perform of this op. Do you want to try that?

RadhikaG:
No problem. I'll implement just a function for the confusion matrix using existing Theano ops. I assume the location of the confusion matrix function is alright?

RadhikaG:
Also, I noticed that there is no union1d operation for tensors. Shall I add on the union1d op in a separate PR as a continuation of https://github.com/Theano/Theano/issues/1916?

RadhikaG:
I've refactored the confusion matrix op to a function, and added unit tests. Please let me know if I need to make any more changes.

RadhikaG:
One Travis [job](https://travis-ci.org/Theano/Theano/jobs/111837509) is still failing, even though it passes on my local install. Could you please retest it once?

nouiz:
this travis problem happen randomly. I restarted it.

trungnt13:
Is it better to support also 2-d tensor as input (softmax matrix from the network)? 
For example:
```
    if pred.ndim == 2:
        pred = tensor.argmax(pred, axis=-1)
    elif pred.ndim != 1:
        raise ValueError('pred must be 1-d or 2-d tensor variable')
```

nouiz:
I think the argmax should be outside this function. But as always, we can
discuss this.

Why do you think it should be inside?

nouiz:
I like the idea of having the same interface as scikit.learn. But there isn't a big difference.

I also think the problem of small batch is important. As the classes much be integer starting from 0 with the current code, what about passing the parameter n_classes that is the number of classes. I think it could be a mandatory parameter. What do you think?

What scikit.learn do when classes isn't passed?

trungnt13:
This is the interface from sklearn:

```
y_true = ["cat", "ant", "cat", "cat", "ant", "bird"]
y_pred = ["ant", "ant", "cat", "cat", "ant", "cat"]
confusion_matrix(y_true, y_pred, labels=["ant", "bird", "cat"])
# or
y_true = [2, 0, 2, 2, 0, 1]
y_pred = [0, 0, 2, 2, 0, 2]
confusion_matrix(y_true, y_pred) # labels=[0,1,2] in this case
```

However, I think in theano case, labels should be a number (nb_classes) or a list of number represent class index (in case users can constraint the order of row and column of confusion matrix). 

Frankly speaking, I prefer a mandatory arguments _nb_classes_ as this is my only use case.

For the _argmax_ idea, just for convenient, I cannot think about the case users accidentally give 2D input to _confusion_matrix_ that they expect the methods do something else instead of argmax.

nouiz:
As Theano don't handle string, I think we must request numbers for inputs.
I would suggest that we requeset inputs between 0 and n classe -1. With
this interface, the 3rd parameter can be n_classes.

What do you think of that? It is mostly the conversion to number of that
interface. We could use another definition of classe number, like start at
0, but I always  see peple start from 0 in deep learning.

trungnt13:
I agree with your interface, so the final interface should be:
> confusion_matrix(pred, actual, nb_class) # tensor.extra_ops.to_one_hot also have nb_class argument

I prefer restrict all inputs must start from 0 to (nb_class -1). 

The most important thing is we have the funciton in theano as soon as possible.

nouiz:
So we agree on the interface, that is great:)
This branch also need a rebase before being merged.

thanks.

nouiz:
Does someone have time to finish this PR? It would be great to get this merged.

xoltar:
@nouiz I rebased and updated docstring example, new PR is  #4810 .

nouiz:
closing as continued in gh-4810

PR #4105, opened 2016-02-23
Revert "cuDNN header / library paths fix"

nouiz:
Reverts Theano/Theano#4084

This cause many problems. I prefer to revert as not having the PR seem better until we make it work correctly.


nouiz:
see gh-4107

PR #4116, opened 2016-02-24
Add debugmode implementation for abstractconv

ballasn:


theano/tensor/nnet/abstract_conv.py
```diff
@@ -9,6 +9,17 @@
 from theano.gof import Apply, Op
 
 
+import numpy
+try:
+    # TODO: move these back out to global scope when they no longer
+    # cause an atexit error
```
lamblin: If there is no error when exiting, we can remove that comment.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -430,6 +441,35 @@ def do_constant_folding(self, node):
...
+        """
+        Basic slow python implementatation for DebugMode
+        """
+
+        if not imported_scipy_signal:
+            raise theano.gof.utils.MethodNotDefined(
+                "c_headers", type(self), self.__class__.__name__,
+                "Need the python package for scipy.signal to be installed "
+                "for the python implementation. You can use the C"
+                " implementation instead.")
```
lamblin: I don't think that is the right exception type to raise, and the message should be updated (is there a C implementation?)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -465,10 +505,30 @@ def make_node(self, img, kern):
...
+            pad_h, pad_w = map(int, mode)
+            mode = "valid"
+            new_img = numpy.zeros((img.shape[0], img.shape[1],
+                                   img.shape[2] + 2 * pad_h,
+                                   img.shape[3] + 2 * pad_w), dtype=img.dtype)
+            new_img[:, :, pad_h:img.shape[2] + pad_h, pad_w:img.shape[3] + pad_w] = img
+            img = new_img
+        if not self.filter_flip:
+            kern = kern[:, :, ::-1, ::-1]
+        conv_out = self.corr2d(img, kern, mode)
```
lamblin: At this point, `mode` is always "valid", right?
So maybe we can just get rid of that keyword here, or call "valid" explicitly.
Also, we should check that `mode` is either "full", "half", "valid" or a pair of int, because we may accept new strings in the future.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -564,10 +624,42 @@ def make_node(self, img, topgrad, shape):
...
+            new_shape = (topgrad.shape[0], topgrad.shape[1],
+                         img.shape[2] - shape[0] + 1,
+                         img.shape[3] - shape[1] + 1)
+            new_topgrad = numpy.zeros((new_shape), dtype=topgrad.dtype)
+            new_topgrad[:, :, ::self.subsample[0], ::self.subsample[1]] = topgrad
+            topgrad = new_topgrad
+
+        topgrad = topgrad.transpose(1, 0, 2, 3)[:, :, ::-1, ::-1]
+        img = img.transpose(1, 0, 2, 3)
+        kern = self.corr2d(img, topgrad, mode)
```
lamblin: Idem, `mode` should always be "valid" here.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -656,10 +748,37 @@ def make_node(self, kern, topgrad, shape):
...
+        mode = self.border_mode
+        pad_h, pad_w = 0, 0
+
+        if mode == "full":
+            pad_h, pad_w = (kern.shape[2] - 1, kern.shape[3] - 1)
+        elif mode == "half":
+            pad_h, pad_w = (kern.shape[2] // 2, kern.shape[3] // 2)
+        elif isinstance(mode, tuple):
+            pad_h, pad_w = map(int, self.border_mode)
+        mode = "valid"
```
lamblin: I don't think `mode` is used afterwards.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -364,6 +364,30 @@ def tcase(self, i, f, s, b, flip, provide_shape):
...
+                            verify_grad=True, mode=mode,
+                            provide_shape=provide_shape, border_mode=b,
+                            filter_flip=flip, target_op=None)
+        self.run_gradinput(inputs_shape=i, filters_shape=f,
+                           output_shape=o, subsample=s,
+                           verify_grad=True, mode=mode,
+                           provide_shape=provide_shape, border_mode=b,
+                           filter_flip=flip, target_op=None)
+
+
```
lamblin: I don't think we need a special test for that, just running the regular tests when the default mode is DebugMode should test everything.

lamblin:
Also, we would still like the perform to fail when running an optimized graph, so users do not run the slow version by mistake. There are different ways to achieve that, I discussed some with @abergeron.

One way is to always have `perform()` raising an Exception, but define `debug_perform` instead, which will be used only by `DebugMode`. The issues with it are:
- it will not work with test values, unless we define a new mechanism for it
- it will not work with `optimizer=None`, which we advise to use in some error messages (unless, again, we define a new mechanism)
- there may be other cases where it would not run (maybe constant folding)

Another way is to add an optimizer doing a final pass on the graph to either:
- raise an Exception if there are still Abstract Ops in the graph, or
- set some flag in the Abstract Ops that will make the perform raise an Exception.

That optimizer would be registered whenever optimizations replacing Abstract Ops would be registered (so 'fast_compile' and 'fast_run', but not 'None').

nouiz:
I like the optimization. It also make the error happen during compilation
instead of during execution.

theano/tensor/nnet/abstract_conv.py
ballasn: This is already check in BaseAbstractConv2d,

Do you think we need to do another check in the perform method of each subclass?

ballasn:
Updated, but the current tests will failed as perform will run the slow_convolution instead of raising NotImplementedError for the TestCpuConvolution

ballasn:
I added the optimization to raise an error when the abstract nodes are still present in the graph, running tensor.nnet.test_abstract_conv in debugmode takes some thought (~3000sec)

theano/tensor/nnet/abstract_conv.py
lamblin: Yes, I just want to make sure that, if we decide to add a new mode (for instance "same"), we do not fall back automatically to the "valid" case because it is neither "full" nor "half".

ballasn:
I added the border_mode verification in the perform methods,

theano/tensor/nnet/abstract_conv.py
```diff
@@ -430,6 +439,32 @@ def do_constant_folding(self, node):
...
+                'invalid mode {}, which must be either '
+                '"valid" or "full"'.format(mode))
+
+        out_shape = get_conv_output_shape(img.shape, kern.shape, mode, [1, 1])
+        out = numpy.zeros(out_shape, dtype=img.dtype)
+        val = _valfrommode(mode)
+        bval = _bvalfromboundary('fill')
+        for b in xrange(img.shape[0]):
+            for n in xrange(kern.shape[0]):
+                for im0 in xrange(img.shape[1]):
```
lamblin: `xrange` is not defined in python 3, so Travis is failing.
You can import it from `six.moves` to be compatible with both.

ballasn: fixed, thanks

PR #4118, opened 2016-02-24
We need to import gof.compiledir in theano-cache.

abergeron:
I don't know when that changed, but it is now necessary.


nouiz:
I got that locally in my repo. We can merge when travis pass.

abergeron:
Travis passed.

PR #4125, opened 2016-02-25
Fix scan pushout

### LINKED ISSUE #4117: Scan GPU test failure: theano.scan_module.tests.test_scan.T_Scan_Gpuarray
In the daily buildbot:
```
FAIL: test_gpu_memory_usage (theano.scan_module.tests.test_scan.T_Scan_Gpuarray)
  File "/tmp/nightly_build_gpuarray/Theano/theano/scan_module/tests/test_scan.py", line 4812, in test_gpu_memory_usage
    assert len(grad_scan_node.outputs) == 2
AssertionError
```

I'm tring to resume the problem here. Mostly this is due to the optimization PushOutScanOutput. There is probably also interaction with Scan.infer_shape!

I did a PR that tried to fix this, but it have other tests failing: https://github.com/Theano/Theano/pull/3797/files. Not sure why those tests fail. This PR change some code in Scan.infer_shape, but not change the behavior. It just make it easier to change the behavior.

https://github.com/Theano/Theano/pull/4013 will add a mechanism that make sure we don't reintroduce old scan in the graph. When merged, we should rebase the fix on top of it before running our tests.

I'll give more information later.

abergeron:
This fixes #4117


nouiz:
Other then my comment, I think it is good to merge.

Issue #4060, opened 2016-02-15
flake8 compile/tests/*.py


lamblin:

fvisin:
This ticket is to fix all the PEP8 errors. This includes making sure that all the functions and classes have comprehensible and exhaustive docstrings.

PR #4126, opened 2016-02-25
flake8 compile/tests/*.py

olimastro:
Fixed the PEP8 Errors.

Did not touch the F403 errors (as reported by flake8). They identify "from ... import * " as "unable to detect undefined names" but in some cases functions from this importation are used.

Also test_function_module.py was modified at 
`def test_swap_SharedVariable_with_given`
it seemed to not do the intended tests.

fix gh-4060


theano/compile/tests/test_function_module.py
```diff
@@ -19,7 +19,10 @@
 
 import numpy as N
 
-PatternOptimizer = lambda p1, p2, ign=True: gof.OpKeyOptimizer(gof.PatternSub(p1, p2), ignore_newtrees=ign)
+
+# PatternOptimizer = lambda p1, p2, ign=True: gof.OpKeyOptimizer(gof.PatternSub(p1, p2), ignore_newtrees=ign)
```
abergeron: Remove the comment

theano/compile/tests/test_function_module.py
```diff
@@ -329,35 +342,37 @@ def test_swap_SharedVariable(self):
...
+        #                        givens={x: train_x[i], y: train_y[i]},
+        #                        updates={train_x: train_x + 0.1})
 
-        test_def = theano.function([i], out, givens={x:test_x[i], y:test_y[i]})
-        test_cpy = train.copy(swap={train_x:test_x, train_y:test_y},
-                              delete_updates=True)
+        test_def = theano.function([i], out, givens={x: test_x[i], y: test_y[i]})
+        test_def2 = theano.function([i], out2, givens={x: test_x[i], y: test_y[i]})
+        # test_cpy = train.copy(swap={train_x: test_x, train_y: test_y},
+        #                      delete_updates=True)
```
abergeron: Instead of commenting this, can you just not assign the result.  Because otherwise, this disables the test of the copy

abergeron:
You also need to remove the fixed files from the tests/test_flake8.py whitelist.

theano/compile/tests/test_function_module.py
```diff
@@ -60,79 +63,88 @@ def test_missing_inputs(self):
         def fn():
             x, s = T.scalars('xs')
             fn = function([], [x])
+            return fn
         checkfor(self, fn, MissingInputError)
```
fvisin: After a discussion with @lamblin on this, it is better not to return anything here as we just want to check that the function call raises an error. The way to remove the flake8 error is then not to assign the function to anything, i.e.
instead of this:
``` python
fn = function([], [x])
return fn
```
having this:
``` python
function([], [x])
```

Please also update the subsequent functions accordingly.

theano/compile/tests/test_function_module.py
```diff
@@ -246,17 +259,17 @@ def test_copy_share_memory(self):
...
-            ori = theano.function([x], [out], mode=mode,updates={z:z+1})
+        for mode in ["FAST_RUN", "FAST_COMPILE"]:
+            ori = theano.function([x], [out], mode=mode, updates={z: z + 1})
             cpy = ori.copy(share_memory=True)
 
             # Test if memories shared
             storage_map_ori = ori.fn.storage_map
             storage_map_cpy = cpy.fn.storage_map
-            fgraph_ori = ori.maker.fgraph
+            # fgraph_ori = ori.maker.fgraph
```
fvisin: You can remove this line.

theano/compile/tests/test_function_module.py
```diff
@@ -266,7 +279,7 @@ def test_copy_share_memory(self):
             for key in storage_map_cpy.keys():
                 storage = storage_map_cpy[key]
                 if key not in i_o_variables or isinstance(key, theano.tensor.Constant):
-                    self.assertTrue(any([ storage is s for s in ori_storages]))
+                    self.assertTrue(any([storage is s for s in ori_storages]))
```
fvisin: If I'm not wrong this is just a very very weird way to do:
```python
self.assertTrue(storage in ori_storages)
```

More generally, with @lamblin or @abergeron benediction, I would rewrite [this whole for loop](https://github.com/olimastro/Theano/blob/0440ea6e2ec912b0c54adddaefcc6d67036e18bc/theano/compile/tests/test_function_module.py#L279-L282) as:
```python
l = [val for key,val in storage_map_cpy.items() 
     if key not in i_o_variables or isinstance(key, theano.tensor.Constant)]
assert all([val in ori_storages for val in l])
```

theano/compile/tests/test_function_module.py
```diff
@@ -382,8 +397,8 @@ def test_shared_state0(self):
         a = T.scalar()  # the a is for 'anonymous' (un-named).
         x, s = T.scalars('xs')
 
-        f = function([x, In(a, value=1.0, name='a'), In(s, value=0.0, update=s+a*x, mutable=True)], s+a*x)
-        g = function([x, In(a, value=1.0, name='a'), In(s, value=f.container[s], update=s-a*x, mutable=True)], s+a*x)
+        f = function([x, In(a, value=1.0, name='a'), In(s, value=0.0, update=s + a * x, mutable=True)], s + a * x)
```
fvisin: These two lines seem too long to me. You almost certainly have to break them.

theano/compile/tests/test_function_module.py
```diff
@@ -396,8 +411,8 @@ def test_shared_state1(self):
         a = T.scalar()  # the a is for 'anonymous' (un-named).
         x, s = T.scalars('xs')
 
-        f = function([x, In(a, value=1.0, name='a'), In(s, value=0.0, update=s+a*x, mutable=True)], s+a*x)
-        g = function([x, In(a, value=1.0, name='a'), In(s, value=f.container[s])], s+a*x)
+        f = function([x, In(a, value=1.0, name='a'), In(s, value=0.0, update=s + a * x, mutable=True)], s + a * x)
```
fvisin: Ditto.

fvisin:
We need to remove this file from the whitelist to make sure Travis actually checks for PEP8 errors in this file. To do this please remove `theano/compile` from the `-e test_flake8.py` list in `.travis.yml` (in theano's root).

theano/compile/tests/test_function_module.py
fvisin: edit: this is probably better
```python
l = [val for key,val in storage_map_cpy.items() 
     if key not in i_o_variables or isinstance(key, theano.tensor.Constant)]
self.assertTrue(all([val in ori_storages for val in l]))
```

theano/compile/tests/test_function_module.py
```diff
@@ -777,16 +792,18 @@ def test_multiple_functions(self):
...
-                return len(saves)-1
+                return len(saves) - 1
             else:
                 return None
+
         def pers_load(id):
             return saves[id]
 
-        a = numpy.random.rand(4, 5)
+        # a = numpy.random.rand(4, 5)
```
fvisin: You can remove this line.

theano/compile/tests/test_function_module.py
```diff
@@ -809,7 +826,7 @@ def pers_load(id):
         fp.close()
         p = pickle.Unpickler(fp2)
         p.persistent_load = pers_load
-        f2 = p.load()
+        # f2 = p.load()
```
fvisin: This should not be commented out, but rather called without assigning it to `f2`.

theano/compile/tests/test_function_module.py
```diff
@@ -845,9 +862,9 @@ def __init__(self):
 
         self.e = a * x + s
 
-        self.f1 = function([x, In(a, value=1.0, name='a'), In(s, value=0.0, update=s+a*x, mutable=True)], s+a*x)
+        self.f1 = function([x, In(a, value=1.0, name='a'), In(s, value=0.0, update=s + a * x, mutable=True)], s + a * x)
```
fvisin: These lines are definitely too long!

theano/compile/tests/test_misc.py
```diff
@@ -56,5 +57,7 @@ def test_nnet(self):
             # print 'Mean cost at epoch %s: %s' % (epoch, mean_cost)
         self.assertTrue(abs(mean_cost - 0.20588975452) < 1e-6)
         # Just call functions to make sure they do not crash.
-        out = nnet.compute_output(input)
-        out = nnet.output_from_hidden(numpy.ones(10))
+        # out = nnet.compute_output(input)
```
fvisin: Remove the commented code.

theano/compile/tests/test_pfunc.py
```diff
@@ -810,15 +810,15 @@ def test_partial_input_aliasing_affecting_inplace_operations(self):
...
-                    + theano.dot((z * 4), m3)))
+            [theano.In(x, mutable=True),
+                theano.In(y, mutable=True),
+                theano.In(z, mutable=True),
+                theano.In(m1, mutable=True),
+                theano.In(m2, mutable=True),
+                theano.In(m3, mutable=True)],
+            (theano.dot((x * 2), m1) +
+                theano.dot((y * 3), m2) +
+                theano.dot((z * 4), m3)))
```
fvisin: The lines inside the list should not be further indented and you can remove the parenthesis in the second block and avoid one line break:
```python
[theano.In(x, mutable=True),
 theano.In(y, mutable=True),
 theano.In(z, mutable=True),
 theano.In(m1, mutable=True),
 theano.In(m2, mutable=True),
 theano.In(m3, mutable=True)],
(theano.dot((x * 2), m1) + theano.dot((y * 3), m2) +
 theano.dot((z * 4), m3)))
```

theano/compile/tests/test_profiling.py
```diff
@@ -101,7 +100,8 @@ def test_ifelse(self):
             big_mat1 = 10
             big_mat2 = 11
 
-            out = f_ifelse(val1, val2, big_mat1, big_mat2)
```
fvisin: You are removing the test by commenting this out. You should rather call it without assigning it to `out` so that flake8 doesn't complain.

fvisin:
That's it! 
Please comment when you are done so that I can check it one last time before merging.

abergeron:
Just to note that the whitelist is not maintained in .travis.yml, but in theano/tests/test_flake8.py.  Please don't edit .travis.yml.

theano/compile/tests/test_function_module.py
abergeron: Yes I this looks ok, but I don't have a strong opinion, as long as it works.

lamblin: `in` compares by value (like `==`), whereas the existing code compares by identity (with `is`).
I don't know the details of how `storage.__eq__` is implemented so that may be equivalent, but I would not do the change without being certain that it is.

olimastro:
All the comments above should be fixed!

theano/compile/tests/test_function_module.py
fvisin: Agreed.

fvisin:
Great, thank you for your work! 
I will review it one more time and merge it if everything is ok.

theano/compile/tests/test_function_module.py
```diff
@@ -246,27 +251,25 @@ def test_copy_share_memory(self):
...
             # and output stoarges are not shared
             i_o_variables = fgraph_cpy.inputs + fgraph_cpy.outputs
             ori_storages = storage_map_ori.values()
-            for key in storage_map_cpy.keys():
-                storage = storage_map_cpy[key]
-                if key not in i_o_variables or isinstance(key, theano.tensor.Constant):
-                    self.assertTrue(any([ storage is s for s in ori_storages]))
+            l = [val for key, val in storage_map_cpy.items()
+                 if key not in i_o_variables or isinstance(key, theano.tensor.Constant)]
+            self.assertTrue(all([val in ori_storages for val in l]))
```
fvisin: Following @lamblin comment, let's make this line
```python
self.assertTrue(any([val is s for val in l for s in ori_storages]))
```

fvisin:
Once my last comment is addressed, this can be merged.
Thank you for your great work!

olimastro:
The line is changed

fvisin:
Great job, thank you!

fvisin:
Closing and reopening to fix Travis.

nouiz:
travis failure are due to a change in miniconda that trigger a latent crash
bug. This should be fixed by:

https://github.com/Theano/Theano/pull/4176

lamblin:
@nouiz has to double check the changes to the tests.

theano/compile/tests/test_function_module.py
```diff
@@ -246,27 +251,25 @@ def test_copy_share_memory(self):
...
             # and output stoarges are not shared
             i_o_variables = fgraph_cpy.inputs + fgraph_cpy.outputs
             ori_storages = storage_map_ori.values()
-            for key in storage_map_cpy.keys():
-                storage = storage_map_cpy[key]
-                if key not in i_o_variables or isinstance(key, theano.tensor.Constant):
-                    self.assertTrue(any([ storage is s for s in ori_storages]))
+            l = [val for key, val in storage_map_cpy.items()
+                 if key not in i_o_variables or isinstance(key, theano.tensor.Constant)]
+            self.assertTrue(any([val is s for val in l for s in ori_storages]))
```
nouiz: double list comprehension is hard to read. It is not clear which of the for loop is the top level loop. Why did you changed that? Also, this does only one assert of one any. In the past, there was many assert of one smaller any call. So this version of the test is less strict. This isn't good.

fvisin: I proposed the change because it is usually faster and to me is easier to read! :) 
I guess it's a matter of taste.

You are right that the two codes are not doing the same thing, the `any` should have been in the list comprehension. To fix this and address your readability concerns too, we can change it to:
```python
l = [val for key, val in storage_map_cpy.items()
      if key not in i_o_variables or isinstance(key, theano.tensor.Constant)]
for storage in l:
    self.assertTrue(any([storage is s for s in ori_storages]))
```

nouiz: that see ok.

fvisin: @olimastro can you please fix this, squash the commits together and push again?

fvisin:
If the rest is ok this can be merged.

olimastro:
It should be

fvisin:
@nouiz can we merge this?

PR #4127, opened 2016-02-25
Fix small buildbot problems

abergeron:
The flake8 thing is just annoying while the other is an error.


nouiz:
ok to merge when travis pass.

PR #4140, opened 2016-02-29
Check for valid combination of padding and strides

sisp:
This PR fixes a bug where the combination of padding and strides is checked. Before, the padding and downscaling factor were compared while actually padding and strides should have been checked.

I'm not sure if these parameter incompatibilities are actually tested. At least, I couldn't see any tests. Would you like me to just add a new test case for this scenario?

### DIFF
--- a/theano/sandbox/gpuarray/opt.py
+++ b/theano/sandbox/gpuarray/opt.py
@@ -10,6 +10,7 @@ from theano.compile.ops import shape_i
 from theano.gof import (local_optimizer, EquilibriumDB,
                         SequenceDB, Optimizer, toolbox)
 from theano.gof.optdb import LocalGroupDB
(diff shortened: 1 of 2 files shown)


sisp:
Existing tests are passing.

abergeron:
I'm not sure but I think the original code was ok and the comment was wrong.

It does make sense to have a padding of (4, 4) and a stride of (2, 2), whereas it doesn't really make sense to have a padding of (4, 4) and a window size of (2, 2).

sisp:
Yes, you're right. I'll undo the changes and fix the message.

sisp:
Not sure what happened. I had reset the my local branch to the commit before the one in this PR and pushed the changes to my fork.

PR #4145, opened 2016-03-01
Improve code misleading tiny issues for Loop of Tutorial.

myme5261314:
The Loop(scan) chapter is a little misleading for reader (at least for me). The `outputs=[results]`, `outputs=results`, `print xxx_fun(...)[0]` and `print xxx_fun(...)` should be in one style to avoid missing reader guess if it's the special situation of the special difficult content `scan` that result in the with/without `[0]`.

Just for curiosity, is the `scan` function which justify **list of rows** into numpy ndarray? Here's the example in the tutorial.
```python
# define tensor variables
X = T.matrix("X")
W = T.matrix("W")
b_sym = T.vector("b_sym")

# define shared random stream
trng = T.shared_randomstreams.RandomStreams(1234)
d=trng.binomial(size=W[1].shape)

results, updates = theano.scan(lambda v: T.tanh(T.dot(v, W) + b_sym) * d, sequences=X)
compute_with_bnoise = theano.function(inputs=[X, W, b_sym], outputs=[results],
                          updates=updates, allow_input_downcast=True)
x = np.eye(10, 2, dtype=theano.config.floatX)
w = np.ones((2, 2), dtype=theano.config.floatX)
b = np.ones((2), dtype=theano.config.floatX)

print(compute_with_bnoise(x, w, b))
```
The result is 
```python
[array([[ 0.96402758,  0.        ],
       [ 0.        ,  0.96402758],
       [ 0.        ,  0.        ],
       [ 0.76159416,  0.76159416],
       [ 0.76159416,  0.        ],
       [ 0.        ,  0.76159416],
       [ 0.        ,  0.76159416],
       [ 0.        ,  0.76159416],
       [ 0.        ,  0.        ],
       [ 0.76159416,  0.76159416]])]
```
I mean each iter of `X` is a row vector of `X`, right? if the `results` is actually only the list of the outputs, the above result should be a list of **row vector** right? but it seems to be a numpy ndarray. I want to know if this is caused by `scan` function internally.


lamblin:
You would have to update the expected results in the doctests (which make Travis fail): using `f = theano.function(..., outputs=x)`, then calling `f(...)` will return one `ndarray`, but if you use `f = theano.function(..., outputs=[x])`, then calling `f(...)` will return a list containing one `ndarray`.

Regarding scan, scan will return all slices concatenated in one `ndarray` with one additional dimension, so all rows will be concatenated into one matrix.

myme5261314:
@lamblin Thanks for your remind. I've finished that, hope the Travis test will pass since I copy the result from the failed job log.

nouiz:
thanks

Issue #207, opened 2011-11-14
Implement gradient of Cholesky op in C


dwf:
The Cholesky Op has a `grad` method implemented that returns a `CholeskyGrad` op, but this op only has a `perform` method and no C code. The algorithm for computing this gradient is very serial and will be quite slow in Python.

This would be a good task for someone who wants to get started writing Ops with C code. See the reference below, section 2.3.1, 5th page (the "F" matrix):
1. S. P. Smith. "Differentiation of the Cholesky Algorithm". _Journal of Computational and Graphical Statistics_, Vol. 4, No. 2 (Jun., 1995), pp. 134-147 http://www.jstor.org/stable/1390762

Or, just look at my `perform` method in `CholeskyGrad`.

juancamilog:
The following article should be of interest to anyone planning to address this issue: http://arxiv.org/pdf/1602.07527v1.pdf

PR #4148, opened 2016-03-01
Symbolic Cholesky gradient implementation

matt-graham:
## Reason for PR

The existing `CholeskyGrad` implementation is implemented with triply nested Python loops which is very slow as noted in #2058, #207 and #208. It also does not have a `grad` method specified meaning it is not possible to differentiate through a `Cholesky` op multiple times. Given it is an atomic op defined directly in python it will also not be optimised to run on the GPU.

There has recently been a [note put on arXiv](http://arxiv.org/abs/1602.07527) giving a simple symbolic expression for the reverse-mode gradient updates for the Cholesky decomposition, which has been used to update the Cholesky gradient implementations in [autograd](https://github.com/HIPS/autograd/commit/024909c58f53b84f5a1c1ab60488c4668284898c#diff-580a144ef545fa15f38dd963feb9aa80) (just for full disclosure, the author of the note, Iain Murray @imurray, is an academic in the department I am PhD student in and this PR in partially in response to discussions with him).

This symbolic definition of the Cholesky gradient can be implemented entirely using existing Theano atomic ops with this PR implementing a first attempt at this. This version already gives a significant speed up on the CPU over the existing implementation for larger matrices (mean of >250× speedup for 200×200 inputs) and no obvious penalty even for small inputs:

profiling_results

It's probable that further performance gains could be achieved some adding some graph optimisations, adding an in-place mode to the `slinalg.Cholesky` op and/or defining a new atomic `CholeskyGrad` op which uses the blocked algorithms also discussed in the arXiv note for larger inputs (there are existing reference python implementations for these by @imurray [here](https://github.com/imurray/chol-rev/blob/master/python/chol_diff.py).

Currently with this alternative `grad` implementation the `Cholesky` op still cannot be differentiated through multiple times as the current `Solve` op does not have a `grad` method implemented. This should be easy to address however (I will add a separate PR for this).

To enable evaluation of gradients of functions involving the Cholesky Op on the GPU should also be much easier with this implementation - most of the atomic ops used in the symbolic expression should already optimise to GPU equivalents, with the major exception being the Cholesky decomposition itself though there has been some work done implementing this in #4006. Currently there is also only a general dense GPU solve op with information about triangular structure ignored however there is a start at attempting to deal with this in #4007 and #3994.

## Description of changes

Rather than creating a `CholeskyGrad` op instance in the `Cholesky.grad` method, here this is replaced instead with a direct symbolic implementation of the gradient updates following the method described in the arXiv note.

Additionally `solve_lower_triangular` and `solve_upper_triangular` op instances are defined to make the calls to the `slinalg.Solve` op more legible.

## Questions

  * I've currently left the existing `CholeskyGrad` implementation in place and just defined the symbolic reverse-mode gradient updates directly in the `Cholesky.grad` method. Is there any way to wrap a symbolic expression as an op so that a new `CholeskyGrad` op could be defined from this (assuming some separate `CholeskyGrad` op should be provided for backwards compatibility)? 
  * Rather than defining the `solve_lower_triangular` and `solve_upper_triangular`  instances globally I could just define them locally inside the method if preferred?
  * I'm not sure if I am using the most efficient method for halving the diagonal of a matrix - I couldn't see any obvious better solution though as `fill_diagonal` can only be used with a scalar and it doesn't seem to be feasible to use `inc_subtensor`. Possibly an optimisation could be added for patterns like `mtx_1 + scalar * tensor.diag(tensor.diagonal(mtx_2))` to make this an in place update to the diagonal were possible? Or would this already be dealt with?

### DIFF
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -62,8 +62,27 @@ class Cholesky(Op):
         z[0] = scipy.linalg.cholesky(x, lower=self.lower).astype(x.dtype)
 
     def grad(self, inputs, gradients):
-        return [CholeskyGrad(self.lower)(inputs[0], self(inputs[0]),
-                                         gradients[0])]
+
+        x = inputs[0]
+        dz = gradients[0]
+        chol_x = self(x)
+        chol_x = theano.printing.Print('Cholesky:')(chol_x)
+
+        def tril_and_halve_diagonal(mtx):
+            """Extracts lower triangle of square matrix and halves diagonal."""
+            return tensor.tril(mtx) - tensor.diag(tensor.diagonal(mtx) / 2.)
+
+        def conjugate_solve_triangular(outer, inner):
+            """Computes P^{-T} Q P^{-1} for lower-triangular P."""
+            return solve_upper_triangular(
+                outer.T, solve_upper_triangular(outer.T, inner.T).T)
+
+        s = conjugate_solve_triangular(
+            chol_x, tril_and_halve_diagonal(chol_x.T.dot(dz)))
+
+        s = theano.printing.Print('S:')(s)
+
+        return [tensor.tril(s + s.T) - tensor.diag(tensor.diagonal(s))]
 
 cholesky = Cholesky()
 
@@ -194,8 +213,9 @@ class Solve(Op):
             return [(rows, cols)]
 
 solve = Solve()  # general solve
-
-# TODO : SolveTriangular
+# lower and upper triangular solves
+solve_lower_triangular = Solve(A_structure='lower_triangular', lower=True)
+solve_upper_triangular = Solve(A_structure='upper_triangular', lower=False)
 
 # TODO: Optimizations to replace multiplication by matrix inverse
 #      with solve() Op (still unwritten)
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -65,11 +65,11 @@ class Cholesky(Op):
         """
         Cholesky decomposition reverse-mode gradient update.
 
-        Symbolic expression for reverse-mode Cholesky gradient taken from [1]_
+        Symbolic expression for reverse-mode Cholesky gradient taken from [0]_
 
         References
         ----------
-        .. [1] I. Murray, "Differentiation of the Cholesky decomposition",
+        .. [0] I. Murray, "Differentiation of the Cholesky decomposition",
            http://arxiv.org/abs/1602.07527
 
         """
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -62,6 +62,16 @@ class Cholesky(Op):
         z[0] = scipy.linalg.cholesky(x, lower=self.lower).astype(x.dtype)
 
     def grad(self, inputs, gradients):
+        """
+        Cholesky decomposition reverse-mode gradient update.
+
+        Notes
+        -----
+        Symbolic expression for reverse-mode Cholesky gradient taken from [1]_
+
+        [1] I. Murray, "Differentiation of the Cholesky decomposition",
+            http://arxiv.org/abs/1602.07527
+        """
 
         x = inputs[0]
         dz = gradients[0]
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -66,7 +66,6 @@ class Cholesky(Op):
         x = inputs[0]
         dz = gradients[0]
         chol_x = self(x)
-        chol_x = theano.printing.Print('Cholesky:')(chol_x)
 
         def tril_and_halve_diagonal(mtx):
             """Extracts lower triangle of square matrix and halves diagonal."""
@@ -79,8 +78,6 @@ class Cholesky(Op):
 
         s = conjugate_solve_triangular(
             chol_x, tril_and_halve_diagonal(chol_x.T.dot(dz)))
-
-        s = theano.printing.Print('S:')(s)
 
         return [tensor.tril(s + s.T) - tensor.diag(tensor.diagonal(s))]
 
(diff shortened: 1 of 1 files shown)


juancamilog:
This PR does result in a considerable speed up for computing the gradients through a Cholesky decomposition. However, I have an use case where it is faster to use existing Theano functionality which produces equivalent results, [see here](https://gist.github.com/juancamilog/f7ba342435c767f2dacd).

The script in the link compares computing the negative log marginal likelihood of a Gaussian process using the matrix_inverse Op, vs using the cholesky Op. Both approaches produce numerically equivalent results, since the graph optimization converts the MatrixInverse nodes into triangular solves via with the Cholesky decomposition. But computing the gradients using the Cholesky decomposition code from this PR is about 25% slower.  

Using the old version of the CholeskyGrad op (int the current version of Theano) is incredibly slower, so I did not include it in the comparison.

juancamilog:
I just saw your PR for the gradient of a Solve Op. I re did the test using solve_lower_triangular and it is faster than the code using the MatrixInverse Op. Thanks for this work!

imurray:
@juancamilog not every computation can avoid pushing the gradient through the Cholesky. For example, if I wanted to do Hamiltonian Monte Carlo using [some parameterizations that use the Cholesky](http://homepages.inf.ed.ac.uk/imurray2/pub/10hypers/). So this PR does serve a purpose.

Incidentally, I've managed to make the Cholesky derivative approach work just as fast or faster for GPs in [a Matlab demo](https://github.com/imurray/chol-rev/blob/master/matlab/chol_rev_gp_demo.m). I guess that using the blocked algorithm in my arXiv note may claw back that 25% for 1000x1000 matrices in Theano too. However, if I understand things correctly(?), this PR wouldn't make any existing code 25% slower (quite the contrary!). I think Matt's been sensible to go for a relatively simple algorithm and a smaller patch first, which may have advantages for getting multiple derivatives working easily, and is already ~25000% faster than what we currently have for this op.

juancamilog:
@imurray I noticed the improvement when I changed the following line in my example:

`Yc = matrix_inverse(cholK).dot(Y) # this will be transformed into a triangular solve by Theano`

to

`Yc = solve_upper_triangular(cholK.T,  solve_lower_triangular(cholK,Y))`

and included the changes from #4153, which adds the implementation of the gradient of the solve Op. That way, using the the Cholesky parametrization is about 15% faster. The reason I didn't explicitly use the solve Ops in the initial example, was that I did not know about #4153 at the moment.

matt-graham:
@juancamilog Thanks for your comments and clarification. No reason you should have known about the `Solve` gradient PR as I only submitted it shortly before you posted your original comment!

Looking at the optimised compiled graph in your original script it seems that though like you indicated the `matrix_inverse(cholK).dot(Y)` is optimised to a triangular solve for the forward pass evaluation of the NLML in the Cholesky formulation, in the gradient calculation graph there is a general `Solve` instance being applied to a `Cholesky` output which hasn't been optimised away for some reason - the following appears in the optimised graph print out

```
Solve{A_structure='general', lower=False, overwrite_A=False, overwrite_b=False}
(Cholesky{lower=True, destructive=False}.0, InplaceDimShuffle{1,0}.0)
```

I'm not quite sure why this is occurring as the optimization does seem to be successfully being applied to other  `Cholesky -> Solve` subgraphs, but it would explain the slower original performance and why with the `Solve.grad` implementation things improve.

matt-graham:
Got the following error on the Travis build

```
Warning, treated as error:
/home/travis/build/Theano/Theano/theano/tensor/slinalg.py:docstring of theano.tensor.slinalg.CholeskyGrad.perform:6: WARNING: Duplicate explicit target name: "1".
```

Wasn't entirely sure what the cause was but guessed it was possibly due to me defining a reference link previously to this as `[1]_` which then conflicted with the other `[1]_` reference in the `CholeskyGrad.perform` docstring, so I changed this to `[0]_` in the `Cholesky.grad` docstring. Apologies for the slightly incomprehensible commit message, forgot to inclued a 'fix' in there...

lamblin:
This looks OK to me, although I think there should be some tests for the `solve_*_triangular` functions.
Regarding your questions:
- The `CholeskyGrad` Op can be kept around for a bit for compatibility with pickled graphs, but should be deprecated. There is no way of replacing an Op with a symbolic expression when unpickling a graph (even though an optimization could do that), except by using `OpFromGraph`, which I would not advise.
- I think defining `solve_*_triangular` to make them available publically is a good thing, and those method should ideally be documented in the online doc.
- An optimized operation to set or increment the elements of the a matrix's diagonal does not exist AFAIK, but would be a great idea. Currently, the way you are doing it is correct, the other option would be to use `T.set_subtensor` with advanced indexing, for instance `T.inc_subtensor(mtx_1[T.arange(mtx1.shape[0]), T.arange(mtx1.shape[0]]), scalar * mtx_2)`.

@nouiz, @dwf, any objections or remarks?

nouiz:
I don't see problems. We could make a mecanism that would unpickled a CholeschyGrad to a graph, but I don't think it is worth it.

dwf:
No objections here: I saw @imurray's arXiv note in passing and waa thinking to myself that it would be nice to replace the clunky thing I wrote once upon a time with the method he describes.

dwf:
Pickled graphs are not really expected to be all that portable across versions. It's nice to support when possible, but I think the Cholesky op is sufficiently niche (and the existing implementation so godawful slow) that I'd be surprised if we heard about it. I doubt the gradient is being used by anyone and if it is, they will surely appreciate the speedup.

juancamilog:
Any news on this PR? Is there anything missing before merging it? I've been using it for quite a while without any problems.

lamblin:
Sorry about the delay. Now seems a good time to merge.
Thanks a lot!

juancamilog:
Thanks! There's also #4153 which implements the gradient for the Solve op, which I have also been using without issues.

PR #4150, opened 2016-03-01
Make the python code for Gemv also check for the need to initialize the output with beta=0

abergeron:
This addresses the problem raised in #4088 in a clean way.

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -183,6 +183,7 @@ except ImportError as e:
                         str(e))
 
 
+# If check_init_y() == True we need to initialize y when beta == 0.
 def check_init_y():
     if check_init_y._result is None:
         if not have_fblas:
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -188,14 +188,16 @@ def check_init_y():
         if not have_fblas:
             check_init_y._result = False
 
-        gemv = _blas_gemv_fns['float64']
         y = float('NaN') * numpy.ones((2,))
         x = numpy.ones((2,))
         A = numpy.ones((2, 2))
+        gemv = _blas_gemv_fns[y.dtype]
         gemv(1.0, A.T, x, 0.0, y, overwrite_y=True, trans=True)
         check_init_y._result = numpy.isnan(y).any()
 
     return check_init_y._result
+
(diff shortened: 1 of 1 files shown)


nouiz:
travis is failing with what look like an infinit loop.

theano/tensor/blas.py
```diff
@@ -182,6 +182,68 @@
                         'dot(vector, vector) (%s)',
                         str(e))
 
+def check_force_gemv_init_core(opt):
```
nouiz: I think we should change opt to pass the op we want to test and pass opt None, to make sure we test the right op.

This would also fix the current recursion.

abergeron:
I did the fix you suggested.

abergeron:
My previous approach obviously doesn't work, so I went in the alternate direction of testing the numpy binding directly.

abergeron:
This passes! Yay!

theano/tensor/blas.py
```diff
@@ -250,10 +270,11 @@ def perform(self, node, inputs, out_storage):
...
                 out *= alpha
-            if beta != 1:
-                out += beta * y
-            else:
-                out += y
+            if beta != 0:
+                if beta != 1:
+                    out += beta * y
+                else:
+                    out += y
```
nouiz: This have the inconvenience of hiding nan. We want to prevent false nan, but we don't want to hide real nan.

I think it is fine to hide them if blas isn't good. But here, we don't do that. I would just remove this change.

abergeron: This hides nan when beta=0 which is what we should do.  They will appear otherwise.

nouiz: IRL discussion with @abergeron This duplicate the BLAS interface that is beta is 0, the multiplication isn't done. So this is consistent with the c code. So this is good.

nouiz:
@abergeron @lamblin comments was to add your comment as a comment in the code. I agree it isn't 100% clear what check mean. After that, we can merge.

theano/tensor/blas.py
lamblin: @abergeron Can you put that explanation in the docstring of `check_init_y`?

nouiz:
ok to merge when travis pass.

Issue #817, opened 2012-08-03
make stabilization optimization to insert expm1 op


nouiz:
start at slide 42 of this presentation for a tutorial on how to make an optimization
https://github.com/abergeron/ccw_tutorial_theano/blob/master/advanced.pdf

Global doc on how to do opt: http://deeplearning.net/software/theano/extending/optimization.html

Register is in register_canonizer, register_stabilize and register_specialize.

The optimization should detect exp(a)-1 and convert this to expm1(a).
opt that will detect first the x-1, then check that x is exp(a).

nouiz:
done.

PR #4151, opened 2016-03-01
Opt

shabanian:
#817

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3576,6 +3576,38 @@ def local_join_make_vector(node):
         return [ret]
 
 
+#################
+# Exp stability #
+#################
+@register_stabilize
+@register_specialize
+@register_canonicalize
+@gof.local_optimizer([T.Elemwise])
+def local_expm1(node):
+    """ 
+    This optimization detects exp(a)-1 and converts this to expm1(a).
+    """
+    if (isinstance(node.op, T.Elemwise) and
+            isinstance(node.op.scalar_op, theano.scalar.basic.Sub)):
+        in1, in2 = node.inputs
+        out = node.outputs[0]
+        #import pdb; pdb.set_trace()
+
+        if (in1.owner and isinstance(in1.owner.op, T.Elemwise) and isinstance(in1.owner.op.scalar_op, theano.scalar.basic.Exp) and 
+                T.extract_constant(in2) == 1):
+            in11 = in1.owner.inputs[0]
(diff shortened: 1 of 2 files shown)


shabanian:
@adri-romsor

nouiz:
travis is failing:

```

FAIL: theano.tensor.tests.test_opt.test_local_expm1

  File "/home/travis/miniconda2/envs/pyenv/lib/python2.6/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_opt.py", line 6143, in test_local_expm1

    assert numpy.all(f_val == f_test(x_val))

AssertionError
```

It fail when mode=FAST_COMPILE. You can manually include the opt in the test to make sure it get applied.

theano/tensor/tests/test_opt.py
```diff
@@ -6122,6 +6123,38 @@ def test3(self):
...
+    t = T.exp(x) - x
+    s = T.exp(u) - numpy.ones((4, 3)).astype(config.floatX)
+
+    f = function([x], y)
+    g = function([x], z)
+    h = function([x], t)
+    r = function([u], s)
+    x_val = numpy.random.rand(4, 3).astype(config.floatX)
+    f_val = f(x_val)
+    f_test = function([x], T.expm1(x))
```
nouiz: pass mode=theano.compile.get_default_mode().including('local_expm1')

It should fix travis.

theano/tensor/opt.py
```diff
@@ -3576,10 +3576,37 @@ def local_join_make_vector(node):
...
+    """
+    This optimization detects exp(a)-1 and converts this to expm1(a).
+    """
+    if (isinstance(node.op, T.Elemwise) and
+            isinstance(node.op.scalar_op, theano.scalar.basic.Sub)):
+        in1, in2 = node.inputs
+        out = node.outputs[0]
+
+        if (in1.owner and isinstance(in1.owner.op, T.Elemwise) and isinstance(in1.owner.op.scalar_op, theano.scalar.basic.Exp) and
+                T.extract_constant(in2) == 1):
```
nouiz: Add the parameter only_process_constants=True to extract_constant().

shabanian:
We set it to False so that it would dig through the dimshuffles, in case we have an array of ones (either as part of the initial call or as a result of a broadcastable dimension). Is that correct or are we missing something?

PR #4152, opened 2016-03-01
Fix doc build

lamblin:


abergeron:
Did you test this locally, because if so, I would merge without waiting for travis.

abergeron:
Also, please don't create branches on the central repo, because that causes travis to build even more stuff.

abergeron:
I did test locally and it works, so I'll merge it.

lamblin:
@abergeron Thanks, that was a configuration mistake on that computer (the central repo was aliased to `origin`), but I killed one of the two instances of Travis before it started.

Issue #1567, opened 2013-10-18
matrix Solve doesn't seem to have gradients


jsalvatier:
The [matrix cookbook](http://www2.imm.dtu.dk/pubdb/views/publication_details.php?id=3274) has the gradient for matrix inverse and matrix multiplication

```
from theano import *
from theano.sandbox.linalg import * 
from theano.tensor import * 
config.compute_test_value = 'ignore'

m = dmatrix('m')
b = dvector('b')

x = grad(sum(solve(m, b)), m)
f = function([m,b], [x])
```

PR #4153, opened 2016-03-01
Gradient implementation for Solve op

matt-graham:
## Reason for PR

Current `tensor.slinalg.Solve` has no `grad` method implemented as noted in #1567 and #4148. This in particular in response to #4148 as implementing the reverse-mode gradient updates for `Solve` is needed to allow arbitrary order differentiation through the `Cholesky` op as noted there.

This implements the reverse-mode gradient updates for the `Solve` op using the symbolic expressions given in [Giles (2008)](http://eprints.maths.ox.ac.uk/1079/). It exploits triangular structure in the input matrix (first input) if specified by the `A_structure` argument, or defaults to a general dense implementation for other `A_structure` values.

A unit test performing a series of `verify_grad` checks for different input dimensions and matrix structures is also defined.

### DIFF
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -193,6 +193,40 @@ class Solve(Op):
             cols = Bshape[1]  # b is a Matrix
             return [(rows, cols)]
 
+    def grad(self, inputs, output_gradients):
+        """
+        Reverse-mode gradient updates for matrix solve operation c = A \ b.
+
+        Symbolic expression for updates taken from [1]_.
+
+        References
+        ----------
+        ..[1] M. B. Giles, "An extended collection of matrix derivative results
+          for forward and reverse mode automatic differentiation",
+          http://eprints.maths.ox.ac.uk/1079/
+
+        """
+        A, b = inputs
+        c = self(A, b)
+        c_bar = output_gradients[0]
+        trans_map = {
+            'lower_triangular': 'upper_triangular',
+            'upper_triangular': 'lower_triangular'
+        }
+        trans_solve_op = Solve(
+            # update A_structure and lower to account for a transpose operation
+            A_structure=trans_map.get(self.A_structure, self.A_structure),
+            lower=not self.lower
+        )
+        b_bar = trans_solve_op(A.T, c_bar)
+        # force outer product if vector second input
+        A_bar = -tensor.outer(b_bar, c) if c.ndim == 1 else -b_bar.dot(c.T)
+        if self.A_structure == 'lower_triangular':
+            A_bar = tensor.tril(A_bar)
+        elif self.A_structure == 'upper_triangular':
+            A_bar = tensor.triu(A_bar)
+        return [A_bar, b_bar]
+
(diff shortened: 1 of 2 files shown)


juancamilog:
Any idea why the travis check failed? I've been using the code from this PR, for about a month, without any issue.

matt-graham:
No I'm a bit lost as to what the issue is, the tests stalled in `theano.sandbox.tests.test_multinomial.test_n_samples_2` which as far as I can see should not have been affected at all by the changes made here.

lamblin:
It was probably a problem with Travis itself, I relaunched the failing test and it passes now.
Unfortunately, I don't have time to review yet.

theano/tensor/slinalg.py
```diff
@@ -193,6 +193,40 @@ def infer_shape(self, node, shapes):
...
+        c = self(A, b)
+        c_bar = output_gradients[0]
+        trans_map = {
+            'lower_triangular': 'upper_triangular',
+            'upper_triangular': 'lower_triangular'
+        }
+        trans_solve_op = Solve(
+            # update A_structure and lower to account for a transpose operation
+            A_structure=trans_map.get(self.A_structure, self.A_structure),
+            lower=not self.lower
```
lamblin: This is probably OK, although self.lower does not seem to be used at all, so maybe we should end up removing it altogether. The information is actually in `A_structure`.
That would be for later.

theano/tensor/tests/test_slinalg.py
```diff
@@ -228,6 +228,36 @@ def test_solve_correctness(self):
...
+    def test_solve_grad(self):
+        if not imported_scipy:
+            raise SkipTest("Scipy needed for the Solve op.")
+        rng = numpy.random.RandomState(utt.fetch_seed())
+        structures = ['general', 'lower_triangular', 'upper_triangular']
+        for A_structure in structures:
+            lower = (A_structure == 'lower_triangular')
+            self.verify_solve_grad(5, None, A_structure, lower, rng)
+            self.verify_solve_grad(6, 1, A_structure, lower, rng)
+            self.verify_solve_grad(4, 3, A_structure, lower, rng)
```
lamblin: The only thing that is bothering me is that a 'general' structure will give "lower=False", but the trans_solve_op will also have a "general" structure with "lower=True", which seems strange.
It seems unused, so it is probably not an issue, but maybe we should put a comment about that.

lamblin: Or maybe just call verify_solve_grad(..., 'general', lower=True) in the test, so that we are confident that the gradient of the gradient is still consistent.

matt-graham:
@lamblin  I've added an extra test case for `A_structure == 'general'` and `lower == True`. 

I agree that the `lower` property seems redundant; I can add a comment to the `Solve` op explaining that its value is currently unused but I guess this might be more appropriate in a separate pull-request / discussion of whether to remove/deprecate?

lamblin:
Sure, a separate discussion and PR would be good.
Thanks!

Issue #4131, opened 2016-02-26
GPU tests fail on current master


Sentient07:
When i ran the nose tests on theano/gof/tests, the tests failed with the following traceback:
Full traceback : 

```
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 1772, in process_node
    replacements = lopt.transform(node)
  File "/Users/Ramana/projects/theano/Theano/theano/sandbox/cuda/opt.py", line 1026, in local_gpu_flatten
    x, shp= node.inputs
ValueError: need more than 1 value to unpack

ERROR (theano.gof.opt): Optimization failure due to: local_gpu_flatten
ERROR (theano.gof.opt): node: Flatten{1}(Dot22.0)
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 1772, in process_node
    replacements = lopt.transform(node)
  File "/Users/Ramana/projects/theano/Theano/theano/sandbox/cuda/opt.py", line 1026, in local_gpu_flatten
    x, shp= node.inputs
ValueError: need more than 1 value to unpack

ERROR (theano.gof.opt): Optimization failure due to: local_gpu_flatten
ERROR (theano.gof.opt): node: Flatten{1}(Alloc.0)
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 1772, in process_node
    replacements = lopt.transform(node)
  File "/Users/Ramana/projects/theano/Theano/theano/sandbox/cuda/opt.py", line 1026, in local_gpu_flatten
    x, shp= node.inputs
ValueError: need more than 1 value to unpack

ERROR (theano.gof.opt): Optimization failure due to: local_gpu_flatten
ERROR (theano.gof.opt): node: Flatten{1}(Alloc.0)
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 1772, in process_node
    replacements = lopt.transform(node)
[...]

PR #4161, opened 2016-03-02
[OPT FAILURE] Fix gh-4131

nouiz:
Fix gh-4131

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ script:
   - cd $(python -c 'import theano; import os; print(os.path.split(theano.__file__)[0])')
   - echo "$PART"
   - cd -; cd Theano
-  - python -c 'import theano; print theano.config.__str__(print_doc=False)'
+  - python -c 'import theano; print(theano.config.__str__(print_doc=False))'
   - python -c 'import theano; assert(theano.config.blas.ldflags != "")'
   - theano-nose -v $PART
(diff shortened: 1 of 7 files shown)


theano/sandbox/cuda/opt.py
```diff
@@ -1023,11 +1023,11 @@ def local_gpu_flatten(node):
             return [gpu_flatten(host_input.owner.inputs[0], outdim)(
                 as_cuda_ndarray_variable(host_input.owner.inputs[0]))]
     if isinstance(node.op, tensor.Flatten):
-        x, shp= node.inputs
-        outdim = node.op.outdim
+        x = node.inputs
```
abergeron: Perhaps you forgot a comma after the `x`to actually do unpacking?

nouiz:
Fixed, added another small change in travis and rebased.

.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
   - cd $(python -c 'import theano; import os; print(os.path.split(theano.__file__)[0])')
   - echo "$PART"
   - cd -; cd Theano
-  - python -c 'import theano; print(theano.config)'
+  - python -c 'import theano; print theano.config.__str__(print_doc=False)'
```
abergeron: This will break on python 3 because you removed the parentheses around the argument to `print`

nouiz:
I fixed python 3.

Issue #1221, opened 2013-02-12
Duplicate class/function tensor.Diag?


nouiz:
In PR https://github.com/Theano/Theano/pull/1088 and https://github.com/Theano/Theano/pull/1119 @abalkin added  tensor_var.diagonal and theano.tensor.{diag,diagonal,Diag,Diagonal}. I think without too much investigation that there is too much Op(4) for that as there also exist theano.sandbox.linalg.heano.sandbox.linalg.ops.{diag,ExtractDiag,AllocDiag}.
- [ ] Check exactly what each op/function do and document why we need all of them
- [ ] If some of them are not needed refactor/remove them. In particular, it is possible that the class theano.tensor.Diag is a duplicate of theano.sandbox.linalg.ops.AllocDiag

abergeron:
After discussion, we should keep Diagonal (with support for arbitrary axis and offset) and remove all the others.  Diag can be implemented with set_subtensor with no loss of functionality.

abergeron:
Diagonal and ExtractDiag have been merged into a single ExtractDiag class that has all the features.

Diag and AllocDiag still left to merge.

abergeron:
Diag and AllocDiag merged in https://github.com/Theano/Theano/pull/5524

PR #4169, opened 2016-03-04
add docs and deprecation warning about AllocDiag and ExtractDiag

hantek:
Fix for #1221 , not finished yet.

### DIFF
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2385,11 +2385,13 @@ def local_gpu_extract_diagonal(node):
         isinstance(node.inputs[0].type, theano.tensor.TensorType)):
         inp = node.inputs[0]
         if (isinstance(node.op, theano.tensor.Diagonal) and
-           node.op.has_default_props()):
+           not node.op.has_default_props()):
             warnings.warn("Diagonal Op has no GPU implementation for"
                           "non-default properties. It will run on CPU.")
             return False
         if inp.owner and isinstance(inp.owner.op, HostFromGpu):
+            import pdb
+            pdb.set_trace()
             return [host_from_gpu(theano.tensor.diagonal(
                 as_cuda_ndarray_variable(inp)))]
     if isinstance(node.op, GpuFromHost):
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2390,8 +2390,8 @@ def local_gpu_extract_diagonal(node):
                           "non-default properties. It will run on CPU.")
             return False
         if inp.owner and isinstance(inp.owner.op, HostFromGpu):
-            import pdb
-            pdb.set_trace()
+            #import pdb
+            # pdb.set_trace()
             return [host_from_gpu(theano.tensor.diagonal(
                 as_cuda_ndarray_variable(inp)))]
     if isinstance(node.op, GpuFromHost):
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2399,6 +2399,34 @@ def local_gpu_extract_diagonal(node):
     return False
 
 
+@register_opt()
+@local_optimizer([theano.tensor.Diagonal])
+def local_gpu_diagonal(node):
+    """
+    diagonal(host_from_gpu()) -> host_from_gpu(diagonal)
+
+    gpu_from_host(diagonal) -> diagonal(gpu_from_host)
+
+    """
+    if (isinstance(node.op, theano.tensor.Diagonal) and
+        isinstance(node.inputs[0].type,
+                   theano.tensor.TensorType)):
+        inp = node.inputs[0]
+        if inp.owner and isinstance(inp.owner.op, HostFromGpu):
+            return [host_from_gpu(theano.tensor.diagonal(
+                as_cuda_ndarray_variable(inp)))]
+    if isinstance(node.op, GpuFromHost):
+        host_input = node.inputs[0]
+        if (host_input.owner and
+            isinstance(host_input.owner.op, theano.tensor.Diagonal) and
+            isinstance(host_input.owner.inputs[0].type,
+                       theano.tensor.TensorType)):
+            diag_node = host_input.owner
+            return [theano.tensor.diagonal(
+                as_cuda_ndarray_variable(diag_node.inputs[0]))]
+    return False
+
+
 def typeConstructor(broadcastable, dtype):
     if dtype == 'float32':
         return CudaNdarrayType(broadcastable=broadcastable)
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2381,9 +2381,12 @@ def local_gpu_extract_diagonal(node):
 
     """
     if (isinstance(node.op, nlinalg.ExtractDiag) and
-        isinstance(node.inputs[0].type,
-                   theano.tensor.TensorType)):
+        isinstance(node.inputs[0].type, theano.tensor.TensorType)):
         inp = node.inputs[0]
+        if (isinstance(node.op, theano.tensor.Diagonal) and
+            (node.op.offset, node.op.axis1, node.op.axis2) != (0, 0, 1)):
+            warnings.warn("Diagonal Op does not non-default properties.")
+            return False
         if inp.owner and isinstance(inp.owner.op, HostFromGpu):
             return [host_from_gpu(nlinalg.extract_diag(
                 as_cuda_ndarray_variable(inp)))]
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2385,8 +2385,9 @@ def local_gpu_extract_diagonal(node):
         isinstance(node.inputs[0].type, theano.tensor.TensorType)):
         inp = node.inputs[0]
         if (isinstance(node.op, theano.tensor.Diagonal) and
-            (node.op.offset, node.op.axis1, node.op.axis2) != (0, 0, 1)):
-            warnings.warn("Diagonal Op does not non-default properties.")
+           node.op.has_default_props()):
+            warnings.warn("Diagonal Op has no GPU implementation for"
+                          "non-default properties. It will run on CPU.")
             return False
         if inp.owner and isinstance(inp.owner.op, HostFromGpu):
             return [host_from_gpu(theano.tensor.diagonal(
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -2372,7 +2372,7 @@ def tensor_to_cuda(x):
 
 
 @register_opt()
-@local_optimizer([nlinalg.ExtractDiag])
+@local_optimizer([nlinalg.ExtractDiag, theano.tensor.Diagonal])
 def local_gpu_extract_diagonal(node):
     """
     extract_diagonal(host_from_gpu()) -> host_from_gpu(extract_diagonal)
@@ -2380,7 +2380,8 @@ def local_gpu_extract_diagonal(node):
     gpu_from_host(extract_diagonal) -> extract_diagonal(gpu_from_host)
 
     """
-    if (isinstance(node.op, nlinalg.ExtractDiag) and
+    if ((isinstance(node.op, nlinalg.ExtractDiag) or
+         isinstance(node.op, theano.tensor.Diagonal)) and
         isinstance(node.inputs[0].type, theano.tensor.TensorType)):
         inp = node.inputs[0]
         if (isinstance(node.op, theano.tensor.Diagonal) and
@@ -2388,40 +2389,13 @@ def local_gpu_extract_diagonal(node):
             warnings.warn("Diagonal Op does not non-default properties.")
             return False
         if inp.owner and isinstance(inp.owner.op, HostFromGpu):
-            return [host_from_gpu(nlinalg.extract_diag(
-                as_cuda_ndarray_variable(inp)))]
-    if isinstance(node.op, GpuFromHost):
-        host_input = node.inputs[0]
-        if (host_input.owner and
-            isinstance(host_input.owner.op, nlinalg.ExtractDiag) and
-            isinstance(host_input.owner.inputs[0].type,
-                       theano.tensor.TensorType)):
-            diag_node = host_input.owner
-            return [nlinalg.extract_diag(
-                as_cuda_ndarray_variable(diag_node.inputs[0]))]
-    return False
-
-
-@register_opt()
-@local_optimizer([theano.tensor.Diagonal])
-def local_gpu_diagonal(node):
-    """
-    diagonal(host_from_gpu()) -> host_from_gpu(diagonal)
-
-    gpu_from_host(diagonal) -> diagonal(gpu_from_host)
-
-    """
-    if (isinstance(node.op, theano.tensor.Diagonal) and
-        isinstance(node.inputs[0].type,
-                   theano.tensor.TensorType)):
-        inp = node.inputs[0]
-        if inp.owner and isinstance(inp.owner.op, HostFromGpu):
             return [host_from_gpu(theano.tensor.diagonal(
                 as_cuda_ndarray_variable(inp)))]
     if isinstance(node.op, GpuFromHost):
         host_input = node.inputs[0]
         if (host_input.owner and
-            isinstance(host_input.owner.op, theano.tensor.Diagonal) and
+            (isinstance(host_input.owner.op, nlinalg.ExtractDiag) or
+             isinstance(host_input.owner.op, theano.tensor.Diagonal)) and
             isinstance(host_input.owner.inputs[0].type,
                        theano.tensor.TensorType)):
             diag_node = host_input.owner
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5939,9 +5939,14 @@ class Diagonal(Op):
             raise ValueError('Currently Diagonal doesn\'t support non-default'
                              'offset and axis values.')
 
+
         # zero-dimensional matrices ...
         if numpy.min(x.shape) == 0:
-            z[0] = node.outputs[0].type.value_zeros((0,))
+            out_shape = [d for i, d in enumerate(x.shape)
+                         if i not in (self.axis1, self.axis2)]
+            diag_size = numpy.min((x.shape[self.axis1], x.shape[self.axis2]))
+            out_shape.append(diag_size)
+            z[0] = node.outputs[0].type.value_zeros(tuple(out_shape))
             return
 
         if x.shape[self.axis1] < x.shape[self.axis2]:
@@ -5987,12 +5992,9 @@ class Diagonal(Op):
         return [tuple(out_shape)]
 
     def has_default_props(self):
-        if (self.offset == self.default_offset and
-           self.axis1 == self.default_axis1 and
-           self.axis2 == self.default_axis2):
-            return True
-        else:
-            return False
+        return  (self.offset == self.default_offset and
+                 self.axis1 == self.default_axis1 and
+                 self.axis2 == self.default_axis2)
 
 
 def diagonal(a, offset=Diagonal.default_offset,
@@ -6083,10 +6085,7 @@ class Diag(Op):
         return [(shapes[0][0],) * 2]
 
     def has_default_props(self):
-        if self.offset == self.default_offset:
-            return True
-        else:
-            return False
+        return  self.offset == self.default_offset
 
 
 def diag(v, k=0):
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5890,23 +5890,23 @@ class Diagonal(Op):
 
     Usage: Diagonal(offset, axis1, axis2)(x)
 
-    "x" has to be a tensor with 2 or more dimensions.
+    `x` has to be a tensor with 2 or more dimensions.
 
     Parameters
     ----------
-    offset
-        Indicates which diagonal to extract, "0" as main diagonal, positive as
+    offset : int
+        Indicates which diagonal to extract, `0` as main diagonal, positive as
         diagonals in the upper triangle, and vice versa.
 
-    axis1, axis2
+    axis1, axis2 : int
         Indicates on which dimensions to extract the diagonals.
 
-    x
+    x: symbolic tensor
         A tensor variable with x.ndim >= 2.
 
     Returns
     -------
-    vector
+    vector : symbolic vector
         A vector representing the diagonal elements.
 
     """
@@ -5921,7 +5921,9 @@ class Diagonal(Op):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim >= 2
+        if x.ndim < 2:
+            raise TypeError('Diagonal needs an input with 2 or more '
+                            'dimensions', x)
         return Apply(self, [x], [tensor(dtype=x.dtype,
                                         broadcastable=[False] * (x.ndim - 1))])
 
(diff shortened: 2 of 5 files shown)


theano/tensor/nlinalg.py
```diff
@@ -144,6 +144,12 @@ class AllocDiag(Op):
 
     __props__ = ()
 
+    def __init__(self):
+        warnings.warn("DeprecationWarning: theano.tensor.nlinalg.AllocDiag"
+                      "is deprecated, please use theano.tensor.Diag"
```
nouiz: I would put the warning inside the make_node, as otherwise, this will always warn, as this file instantiate one such object. Idem for the other warning.

theano/tensor/nlinalg.py
```diff
@@ -179,6 +185,10 @@ class ExtractDiag(Op):
     __props__ = ("view",)
 
     def __init__(self, view=False):
+        warnings.warn("DeprecationWarning: theano.tensor.nlinalg.ExtractDiag"
```
nouiz: ExtractDiag have an optimization in theano/sandbox/cuda/opt.py. Double check that this opt also work for Diagonal.

Make sure that Diag and Diagonal have the grad and the infershape implemented.

theano/tensor/basic.py
```diff
@@ -5888,8 +5888,19 @@ def all(x, axis=None, keepdims=False):
 class Diagonal(Op):
     """Return specified diagonals.
 
+    Usage: Diagonal(offset, axis1, axis2)(x)
+
+    "x" has to be a tensor with 2 or more dimensions.
```
bouthilx: `x`

theano/tensor/basic.py
```diff
@@ -5888,8 +5888,19 @@ def all(x, axis=None, keepdims=False):
...
+
     Parameters
     ----------
+    offset
```
bouthilx: int

theano/tensor/basic.py
```diff
@@ -5888,8 +5888,19 @@ def all(x, axis=None, keepdims=False):
...
+        Indicates which diagonal to extract, "0" as main diagonal, positive as
```
bouthilx: ``` `0` ```

theano/tensor/basic.py
```diff
@@ -5888,8 +5888,19 @@ def all(x, axis=None, keepdims=False):
...
+        diagonals in the upper triangle, and vice versa.
+
+    axis1, axis2
```
bouthilx: : int

theano/tensor/basic.py
```diff
@@ -5888,8 +5888,19 @@ def all(x, axis=None, keepdims=False):
...
+        Indicates on which dimensions to extract the diagonals.
+
     x
```
bouthilx: : symbolic tensor

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
 
 
 def diagonal(a, offset=0, axis1=0, axis2=1):
+    """
+    A helper function for Diagonal Op.
```
bouthilx: I would be more explicit here. Look at how detailed [numpy's documentation](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.diagonal.html) is.

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+
+    Parameters
+    ----------
+    a
```
bouthilx: : symbolic tensor

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+        A tensor variable with x.ndim >= 2.
+
+    offset
```
bouthilx: : int

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+        Indicates which diagonal to extract, "0" as main diagonal, positive as
```
bouthilx: ``` `0` ```

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+        diagonals in the upper triangle, and vice versa.
+
+    axis1, axis2
```
bouthilx: : int

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+        A vector representing the diagonal elements.
+    """
     if (offset, axis1, axis2) == (0, 0, 1):
         return theano.tensor.nlinalg.extract_diag(a)
     return Diagonal(offset, axis1, axis2)(a)
 
 
 class Diag(Op):
+    """
+    An op that transfers a vector into a matrix containing values in that
```
bouthilx: What about: Copy a vector to the diagonal of an empty matrix (all zeros)?

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+    vector as its diagonal. It does the inverse of Diagonal.
 
+    Usage: T.diag()(x)
+
+    x should be a tensor vector. The parenthesis in the front should indicate
+    which main diagonal the vector value goes into. By default it is set to (
+    "0", which corresponds to setting the values of x to the main diagonal in
+    the returned matrix. Currently it only accepts "0", as other diagonals
+    hasn't been implemented for now.
+
```
bouthilx: Missing Parameters section

theano/tensor/basic.py
```diff
@@ -5943,13 +5954,45 @@ def infer_shape(self, node, shapes):
...
+    An op that transfers a vector into a matrix containing values in that
+    vector as its diagonal. It does the inverse of Diagonal.
 
+    Usage: T.diag()(x)
+
+    x should be a tensor vector. The parenthesis in the front should indicate
+    which main diagonal the vector value goes into. By default it is set to (
+    "0", which corresponds to setting the values of x to the main diagonal in
+    the returned matrix. Currently it only accepts "0", as other diagonals
+    hasn't been implemented for now.
```
bouthilx: Do you intend to modify the Op? So far I don't see anything about axis diagonal set to 0 by default.

theano/tensor/basic.py
```diff
@@ -5972,6 +6015,21 @@ def infer_shape(self, nodes, shapes):
 
 
 def diag(v, k=0):
+    """
+    A helper function for two ops: Diagonal and Diag. It both accepts tensor
```
bouthilx: Again I like the concise documentation of numpy for [diag](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.diag.html). I would first give an overview of the behavior of the function and than explain we glue two ops together which do this and this.

theano/tensor/basic.py
```diff
@@ -5972,6 +6015,21 @@ def infer_shape(self, nodes, shapes):
...
+    vector and tensor matrix. While the passed tensor variable v has
+    v.ndim>=2, it builds a Diagonal instance, and returns a vector with its
+    entries equal to v's main diagonal; otherwise if v.ndim is 1, it builds a
+    Diag instance, and returns a matrix with v at its k-th diaogonal.
+
+    Parameters
+    ----------
+    v
```
bouthilx: : symbolic tensor

theano/tensor/basic.py
```diff
@@ -5972,6 +6015,21 @@ def infer_shape(self, nodes, shapes):
...
+        theano tensor variable
+    k
```
bouthilx: : int

theano/tensor/nlinalg.py
bouthilx: @nouiz It says ExtractDiag works on the GPU but I can't find its GPU implementation.

nouiz: check the file cuda/opt.py We reuse the same op on the GPU.

On Tue, Mar 8, 2016 at 2:44 PM, Xavier Bouthillier <notifications@github.com

bouthilx: Yes I saw it and that is what confuses me. How could it run on the GPU with only CPU code?

nouiz: The perform code work on ndarray and CudaNDarray object.
Le 8 mars 2016 21:24, "Xavier Bouthillier" <notifications@github.com> a
écrit :

bouthilx: Really? :open_mouth: How could it be? Any access to x like x[0] or x[:, 0] returns a CudaNDarray? It only work if there is no arithmetic on x right?

nouiz: The opt will build a graph such the the input variable is an
CudaNdarrayVariable, so at run time, it will receive an CudaNdarray
instance. The make_node of ExtractDiag, have this at the start:

```
    def make_node(self, _x):
        if not isinstance(_x, theano.Variable):
            x = as_tensor_variable(_x)
        else:
            x = _x
```

So it don't convert a CudaNdarrayVariable to a TensorVarible.

On Wed, Mar 9, 2016 at 9:38 AM, Xavier Bouthillier <notifications@github.com

bouthilx: Yes it's not what confuses me, when I said any access to x, I meant any access to x where x is a CudaNdarray. I'm confused because normally to have any operation going on the GPU we have some c code with __global__ functions meant exclusively for the GPU (be it hidden under PyCuda or anything else).

theano/tensor/basic.py
hantek: It is not explicitly written in the code but now it is only valid to call Diag()(v) or Diag(0)(v), any other forms like Diag(2)(v) would give you the same as Diag(0)(v). But it should be returning a matrix with those values in the corresponding minor diagonals.

bouthilx: OK. The value should be tested and an error should be raised if it's not 0. You should also add this value to ```__props__```. The assert line 6034 could then be removed.

hantek:
Hi Xavier,

I pushed a new commit just now. The remaining problem is that gradients yielded by both Diag and Diagonal are not correct / not implemented when we were dealing with minor diagonals.

theano/tensor/basic.py
```diff
@@ -5910,7 +5921,9 @@ def __init__(self, offset=0, axis1=0, axis2=1):
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim >= 2
+        if x.ndim < 2:
+            raise TypeError('Diagonal needs an input with 2 or more '
+                            'dimensions', x)
         return Apply(self, [x], [tensor(dtype=x.dtype,
                                         broadcastable=[False] * (x.ndim - 1))])
```
bouthilx: You should copy the perform method from ExtractDiag so that it works on the GPU too.

theano/sandbox/cuda/opt.py
```diff
@@ -2399,6 +2399,34 @@ def local_gpu_extract_diagonal(node):
     return False
 
 
+@register_opt()
+@local_optimizer([theano.tensor.Diagonal])
```
bouthilx: You can add Diagonal to the optimization above and replace nlinalg.extract_diag by tensor.diagonal once you have modified Diagonal.perform method.

You should also add a test to make sure ExtractDiag and Diagonal behaves the same way.

bouthilx: Why don't you make a single optimization with ExtractDiag's one? Am I missing something?

hantek: I thought ExtractDiag allows offset to be non-zero, while Diagonal dosen't, so I kept it. After looking the code again I find it worng... Both of them don't support non-default offsets now. So I will merge the two opts.

theano/tensor/basic.py
```diff
@@ -5888,41 +5888,79 @@ def all(x, axis=None, keepdims=False):
...
 
     def perform(self, node, inputs, outputs):
         (x,) = inputs
         (z,) = outputs
-        z[0] = x.diagonal(self.offset, self.axis1, self.axis2)
+        if self.offset != 0 or self.axis1 != 0 or self.axis2 != 1:
+            raise ValueError('Currently Diagonal doesn\'t support non-default'
```
bouthilx: This breaks the interface for CPU. What we should do is test that if x is not CudaNdarray and any op property is not default, then return x.diagonal() as previously done.

theano/tensor/basic.py
```diff
@@ -5888,41 +5888,79 @@ def all(x, axis=None, keepdims=False):
...
+        else:
+            z[0] = rval.copy()
 
     def grad(self, inputs, gout):
-        (x,) = inputs
-        (gz,) = gout
-        return [grad_not_implemented(self, 0, x)]
+        """The current gradient is valid only for main diagonals."""
+        x = theano.tensor.zeros_like(inputs[0])
+        xdiag = diag(gout[0])
```
bouthilx: return grad_not_implemented if any of offset, axis1  or axis2 is not default

theano/sandbox/cuda/opt.py
```diff
@@ -2380,21 +2380,26 @@ def local_gpu_extract_diagonal(node):
...
     """
-    if (isinstance(node.op, nlinalg.ExtractDiag) and
-        isinstance(node.inputs[0].type,
-                   theano.tensor.TensorType)):
+    if ((isinstance(node.op, nlinalg.ExtractDiag) or
+         isinstance(node.op, theano.tensor.Diagonal)) and
+        isinstance(node.inputs[0].type, theano.tensor.TensorType)):
         inp = node.inputs[0]
+        if (isinstance(node.op, theano.tensor.Diagonal) and
+            (node.op.offset, node.op.axis1, node.op.axis2) != (0, 0, 1)):
```
bouthilx: Having such integers free in the wild is not very good. Since there is multiple places where we test if a diagonal op has default properties, I suggest we add a method Diagonal.has_default_properties() and add each default values as static attributes.  

```python
class Diagonal(Op):
    __props__ = ("offset", "axis1", "axis2")
    default_offset = 0
    default_axis1 = 0
    default_axis2 = 0

    def __init__(self, offset=default_offset, axis1=default_axis1, axis2=default_axis2):
        ...
```

theano/tensor/basic.py
```diff
@@ -5943,25 +5981,77 @@ def infer_shape(self, node, shapes):
 
 
 def diagonal(a, offset=0, axis1=0, axis2=1):
```
bouthilx: You could use Diagonal default attributes as suggested above

theano/tensor/tests/test_basic.py
```diff
@@ -7272,6 +7272,34 @@ def test_bad_number_of_shape(self):
...
 
 class TestInferShape(utt.InferShapeTester):
+    def test_Diagonal(self):
+        atens3 = tensor3()
+        atens3_val = rand(4, 5, 3)
+        atens3_diag = Diagonal()(atens3)
+
+        self._compile_and_check([atens3], [atens3_diag],
+                                [atens3_val], Diagonal)
+        # atens3_diag = Diagonal(1)(atens3)
```
bouthilx: Should uncomment those

bouthilx:
@hantek ExtractDiag and AllocDiag have their own tests. Since Diagonal and Diag should now have the same basic behavior (plus some extra with non default attributes), can you look at those tests and make sure Diagonal and Diag are tested the same way as ExtractDiag and AllocDiag. Also, add a comments like ```# ExtractDiag is deprecated``` in the tests for clarity.

theano/tensor/basic.py
```diff
@@ -5943,25 +5981,77 @@ def infer_shape(self, node, shapes):
...
-
         return Apply(self, [diag], [matrix(dtype=diag.dtype)])
 
     def perform(self, node, inputs, outputs):
         (z,) = outputs
-        z[0] = numpy.diag(inputs[0])
+        z[0] = numpy.diag(inputs[0], self.offset)
 
     def grad(self, inputs, gout):
         (gz,) = gout
```
bouthilx: You should return grad_not_implemented for non-zero offset.

theano/tensor/basic.py
hantek: Why this breaks the CPU interface? We have our own implementation of diagonal in the following lines of perform() method.

theano/tensor/basic.py
hantek: the interface of diagonal() function was like this before this PR, are we sure we need to change it? We will finally allow non-default settings for these Ops in the end, we may need to change the interface back then.

theano/tensor/basic.py
bouthilx: We talked about it in private. For the record, this breaks the CPU interface because the op used to work for non default values. Now it raises a ValueError.

theano/tensor/basic.py
bouthilx: We talked about it in private. For the record, ```offset=0``` could be replace by something like ```offset=Diagonal.default_offset``` and similarly for axis1 and axis2.

theano/tensor/basic.py
hantek: For the record, I agree with what you said.

bouthilx: :stuck_out_tongue:

theano/sandbox/cuda/opt.py
bouthilx: @hantek You forgot to use has_default_props here.

theano/sandbox/cuda/opt.py
```diff
@@ -2372,29 +2372,34 @@ def tensor_to_cuda(x):
...
+            warnings.warn("Diagonal Op does not non-default properties.")
```
bouthilx: What about "Diagonal Op has no GPU implementation for non-default properties. It will run on CPU"

theano/tensor/basic.py
```diff
@@ -5888,41 +5888,91 @@ def all(x, axis=None, keepdims=False):
...
+            self.view = True
         self.offset = offset
         self.axis1 = axis1
         self.axis2 = axis2
 
     def make_node(self, x):
         x = as_tensor_variable(x)
-        assert x.ndim >= 2
+        if x.ndim < 2:
+            raise TypeError('Diagonal needs an input with 2 or more '
```
bouthilx: I know this was in ExtractDiag, but this looks more like a ValueError than a TypeError to me.

theano/sandbox/cuda/opt.py
hantek: Yeah that's definitely better.

theano/tensor/basic.py
```diff
@@ -5888,41 +5888,91 @@ def all(x, axis=None, keepdims=False):
...
         return Apply(self, [x], [tensor(dtype=x.dtype,
                                         broadcastable=[False] * (x.ndim - 1))])
 
     def perform(self, node, inputs, outputs):
         (x,) = inputs
         (z,) = outputs
-        z[0] = x.diagonal(self.offset, self.axis1, self.axis2)
+        if not self.has_default_props():
+            raise ValueError('Currently Diagonal doesn\'t support non-default'
+                             'offset and axis values.')
```
bouthilx: @hantek Forgot to add numpy's diagonal() for non-default values?

theano/tensor/basic.py
```diff
@@ -5941,44 +5991,136 @@ def infer_shape(self, node, shapes):
         out_shape.append(diag_size)
         return [tuple(out_shape)]
 
+    def has_default_props(self):
+        if (self.offset == self.default_offset and
```
bouthilx:
```python
return (self.offset == self.default_offset and
        self.axis1 == self.default_axis1 and
        self.axis2 == self.default_axis2)
```

theano/tensor/basic.py
```diff
@@ -5941,44 +5991,136 @@ def infer_shape(self, node, shapes):
...
+            return [diagonal(gz)]
+        else:
+            return [grad_not_implemented(self, 0, inputs[0])]
 
     def infer_shape(self, nodes, shapes):
         return [(shapes[0][0],) * 2]
 
+    def has_default_props(self):
+        if self.offset == self.default_offset:
+            return True
```
bouthilx:
```python
return self.offset == self.default_offset
```

theano/tensor/basic.py
```diff
@@ -5941,44 +5991,136 @@ def infer_shape(self, node, shapes):
...
+    """
     if v.ndim == 1:
         assert k == 0, "diagonals other than main are not implemented"
         return Diag()(v)
-    elif v.ndim == 2:
+    elif v.ndim >= 2:
         return diagonal(v, k)
     else:
-        raise ValueError("Input must be 1- or 2-d.")
+        raise TypeError("Input must has v.ndim >=1.")
```
bouthilx: This also look more to ValueError than TypeError to me

hantek: I think it's a type error. If it's a vector, then the variable type is T.vector; otherwise, if it's 2-d or 3-d, it is T.matrix of T.tensor3 instead.

hantek: In the test code we can find that all confusions between vector and matrix are treated as TypeError, it looks not natural to have a ValueError in the middle of many TypeErrors.

theano/tensor/tests/test_basic.py
```diff
@@ -7271,7 +7271,284 @@ def test_bad_number_of_shape(self):
...
+
+        x = theano.tensor.tensor3()
+        y = diagonal(x)
+        assert y.owner.op.__class__ == Diagonal
+
+        # other types should raise error
+        x = theano.tensor.vector()
+        ok = False
+        try:
+            y = diagonal(x)
```
bouthilx: It would be better with ```assert_raises``` like [here](https://github.com/hantek/Theano/blob/diag/theano/tensor/tests/test_basic.py#L5509)

hantek: Corrected that.

theano/tensor/basic.py
bouthilx: Yes it is a little misleading. However, theano.tensor.vector is not a type, it is a constructor for a TensorVariable of a specific ndim. You can see that both theano.tensor.vector and theano.tensor.matrix returns objects with the same type.

```python
>>> v = theano.tensor.vector()
>>> print v
<TensorType(float32, vector)>
>>> print type(v)
<class 'theano.tensor.var.TensorVariable'>

>>> m = theano.tensor.matrix()
>>> print m
<TensorType(float32, matrix)>
>>> print type(m)
<class 'theano.tensor.var.TensorVariable'>
```

So I would stick to ValueError.

theano/tensor/tests/test_basic.py
```diff
@@ -7271,7 +7272,258 @@ def test_bad_number_of_shape(self):
...
+            floatX = config.floatX
+        self.floatX = floatX
+        self.type = type
+        super(test_diagonal, self).__init__(name)
+
+    def test_diagonal_input(self):
+        # test that it accepts the right form of input
+        x = theano.tensor.matrix()
+        y = diagonal(x)
+        assert y.owner.op.__class__ == Diagonal
```
bouthilx: Why not using ```isinstance(y.owner.op, Diagonal)```?

theano/tensor/basic.py
bouthilx: You also need to verify x is not a CudaNdarray

theano/tensor/basic.py
```diff
@@ -5888,41 +5888,89 @@ def all(x, axis=None, keepdims=False):
...
-    __props__ = ("offset", "axis1", "axis2")
+    __props__ = ("offset", "axis1", "axis2", "view")
+    default_offset = 0
+    default_axis1 = 0
+    default_axis2 = 1
 
-    def __init__(self, offset=0, axis1=0, axis2=1):
-        if numpy_diagonal_return_view:
+    def __init__(self, offset=0, axis1=0, axis2=1, view=False):
+        self.view = view
```
bouthilx: You should not set view = True if numpy_diagonal_return_view is False.

Actually, to make it more transparent I would do something like this:

```python
def __init__(self, offset=0, axis1=0, axis2=1, view=False):
    self.view = view
    if self.view and not numpy_diagonal_return_view:
        warnings.warn("View will forced to False. Diagonal property view is "
                      "set to True but numpy version %s and prior versions of "
                      "numpy.diagonal() do not return a view. Update "
                      "numpy to use Diagonal(view=True)" % 
                      numpy.version.version)
        self.view = False
    if self.view:
        self.view_map = {0: [0]}
```

theano/tensor/basic.py
bouthilx: @hantek You should harmonise this error type with Diag.make_node.

nouiz:
Should this PR be fixed?

hantek:
It is splitted into 2 tickets. One (#5247) is done, the other (#5524) is in progress. This PR may finally not make itself to be merged.

abergeron:
This was finished elsewhere

PR #4172, opened 2016-03-04
More hints for failed AbstractConv2d optimization

f0k:
With https://github.com/Theano/Theano/pull/4045, Windows users without BLAS don't get an `undefined reference to _dgemm` any more, but another error message that's not too useful either: https://groups.google.com/forum/#!topic/lasagne-users/LrP3YGzh8f0
This adds a hint about installing a BLAS library.

/disclaimer: I did this in the github web interface, hope I didn't introduce any syntax error.


lamblin:
Thanks!

PR #4176, opened 2016-03-07
Make BatchedDot don't try to use c code if there is no blas.

nouiz:
This fix crash in travis. I don't know why the new travis don't link anymore to blas. But this fix a bug in that case.

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -2134,6 +2134,10 @@ class BatchedDot(Op):
         _z, = out
         fail = sub["fail"]
 
+        if not config.blas.ldflags:
+            return super(BatchedDot, self).c_code(node, name,
+                                                  inp, out, sub)
+
         # generate contiguity condition
         def contiguous(var, ndim):
             strides = "PyArray_STRIDES(%s)" % var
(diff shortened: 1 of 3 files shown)


lamblin:
I thought we managed to link with blas on Travis.
Did something change? @abergeron

theano/tensor/blas.py
```diff
@@ -2134,6 +2134,10 @@ def c_code(self, node, name, inp, out, sub):
         _z, = out
         fail = sub["fail"]
 
+        if not config.blas.ldflags:
+            return super(BatchedDot, self).c_code(node, name,
+                                                  inp, out, sub)
```
lamblin: Is it possible for the error message to say that BLAS is needed?

lamblin:
I'm OK with skipping the tests, but we would need something to tell us when linking with BLAS on Travis is broken, so we can fix it.

theano/tensor/blas.py
nouiz: This don't raise an error. It just skip the c code for this op. It have a perform, so it won't cause crash later.

nouiz:
Agreed, but I let this for another PR to get this one merged sooner.

We can add an assert that blas.ldflags isn't empty.

abergeron:
What changed is what conda installs.  We no longer install the mkl version of numpy and we no longer install the mkl-rt pacakge.  I guess this is something that changed with a very recent update of conda itself.

PR #4186, opened 2016-03-07
Force older conda version on travis

abergeron:
This should hopefully fix the no-blas condition on travis.

fix gh-4187


abergeron:
Now this is working.

nouiz:
ok to merge when travis pass.

lamblin:
`The command "python -c 'import theano; assert(theano.config.blas.ldflags != "")'" exited with 1.`

lamblin:
The assert passed, so I'm merging.

Issue #4185, opened 2016-03-07
flake8 misc/tests/*.py


nouiz:

theaverageguy:
This one can be closed @nouiz

PR #4190, opened 2016-03-07
flake8 misc/tests/*.py

taesupkim:
Related to issue #4185.

Fixed errors related to flake8


theano/misc/tests/test_gnumpy_utils.py
```diff
@@ -44,9 +40,6 @@ def test2(shape=(3, 4, 5)):
     Make sure that the gnumpy conversion is exact from CudaNdarray to
     garray back to CudaNdarray.
     """
-    gpu = theano.sandbox.cuda.basic_ops.gpu_from_host
```
fvisin: Why did you remove these lines? Just don't assign the theano.function to `ii`.

taesupkim: The removed 3 lines, which are defining function `ii`, are not used in function `test2`.

fvisin: The test might be checking if the theano.function can be compiled. Just remove the final assignment please (i.e. call `theano.function` but don't assign it to `ii`)

theano/misc/tests/test_pycuda_theano_simple.py
```diff
@@ -14,19 +14,19 @@
 import theano.sandbox.cuda as cuda_ndarray
 import theano.misc.pycuda_init
 
+import pycuda
+import pycuda.driver as drv
+import pycuda.gpuarray
+
 if not theano.misc.pycuda_init.pycuda_available:
```
fvisin: Travis fails if you import pycuda before checking `theano.misc.pycuda_init.pycuda_available` because it tries to execute the test without GPU.

Please revert the previous import order and add `# noqa` to the lines checking `theano.misc.pycuda_init.pycuda_available`. For example here should be:
```python
import theano.sandbox.cuda as cuda_ndarray
import theano.misc.pycuda_init

if not theano.misc.pycuda_init.pycuda_available:  # noqa
    raise SkipTest('Optional theano package cuda disabled')

import pycuda
import pycuda.driver as drv
import pycuda.gpuarray
```

Please do the same for the other files too. This should fix Travis error.

taesupkim:
@fvisin Hope all done with this. Thanks!

fvisin:
Thank you @taesupkim! 
Looks good to me and Travis passes. Merged.

PR #4195, opened 2016-03-08
Skip the gpuarray tests is init_gpu_device == gpu*

abergeron:


lamblin:
LGTM. I killed the non-flake8 tests, when they pass we can merge.

Issue #4199, opened 2016-03-09
Impossible to compile a COp with more outputs than inputs


mjwillson:
This looks like it's due to a typo here:

https://github.com/Theano/Theano/blob/master/theano/gof/op.py#L1426

(should be `range(len(node.outputs))` not `range(len(node.inputs))`)

I ran into it when trying to implement a Gpu-based MaxAndArgmax op:

`('The following error happened while compiling the node', GpuMaxAndArgmaxPerRow(GpuContiguous.0), '\n', 'list index out of range')`

mjwillson:
Furthermore I didn't find any tests at all for COp (although it's indirectly covered via GpuDnnConv) -- I can make a PR for the one-line fix if you want, but not sure I want to take on the task of testing it from scratch, maybe the author has a preference on how it should be tested.

nouiz:
Good catch. A PR with the fix is welcome. I don't think testing COp
directly is possible. At least, I don't know how and the indirect test are
pretty good I think. So I wouldn't spend time on that.

nouiz:
I just saw that this is already done. Thanks.

theaverageguy:
@nouiz I would like to express my interest in Theano for GSoC 2016 with this PR. I am hunting for more bugs to fix before I introduce myself. :)

nouiz:
I would recommand that you introduce yourself now on theano-dev and that
you read https://github.com/Theano/Theano/wiki/GSoC2016

mjwillson:
Nice one, thanks!

PR #4200, opened 2016-03-09
Impossible to compile a COp

theaverageguy:
Fixes https://github.com/Theano/Theano/issues/4199


nouiz:
thanks, we can merge when travis pass.

theaverageguy:
The build passed @nouiz :) 
I am adding 3 more PRs today, and then will move to gsoc project discussions.

lamblin:
Thanks!

lamblin:
A small test for that feature would be to add a dummy Op with more outputs than inputs, in a `.c` file in `theano/gof/tests/`, and define a `COp` for instance in `test_op.py` or `test_cop.py`.
This is not a global test of `COp`, but at least a unit test for a past known issue.

Issue #4111, opened 2016-02-23
print_test_value should be mentioned in debugging tutorial


rizar:
[This feature](http://deeplearning.net/software/theano/library/config.html#config.print_test_value) should be used [here](http://deeplearning.net/software/theano/tutorial/debug_faq.html).

yash14123:
Working on it.

lamblin:
@IshmaelBelghazi You can take inspiration from #4202 if you want, taking into account the comments that have been made there. Or, if you prefer, you can start from scratch.

PR #4202, opened 2016-03-09
add print_test_value to debug tutorial

theaverageguy:
Fixes https://github.com/Theano/Theano/issues/4111

### DIFF
--- a/doc/tutorial/debug_faq.txt
+++ b/doc/tutorial/debug_faq.txt
@@ -214,6 +214,13 @@ The ``compute_test_value`` mechanism works as follows:
 * ``ignore``: Silently ignore the computation of intermediate test values, if a
   variable is missing a test value.
 
+Theano also provides the ``print_test_value`` print flag which can hold a boolean value.The default value is false.If 'True', Theano will override the ‘__str__’ method of its variables to also print the tag.test_value when this is available.
+It can be used as 
+
+.. code-block:: none
+
+    theano.config.print_test_value = 'True'
+
 .. note::
   This feature is currently incompatible with ``Scan`` and also with ops
   which do not implement a ``perform`` method.
--- a/doc/tutorial/debug_faq.txt
+++ b/doc/tutorial/debug_faq.txt
@@ -214,17 +214,12 @@ The ``compute_test_value`` mechanism works as follows:
 * ``ignore``: Silently ignore the computation of intermediate test values, if a
   variable is missing a test value.
 
-Theano also provides the ``print_test_value`` print flag which can hold a boolean value.The default value is false.If 'True', Theano will override the ‘__str__’ method of its variables to also print the tag.test_value when this is available.
+Theano also provides the ``print_test_value`` print flag which can hold a boolean value.The default value is ``false``.If ``True``, Theano will override the ‘__str__’ method of its variables to also print the tag.test_value when this is available.
 It can be used as 
 
 .. code-block:: none
 
     theano.config.print_test_value = 'True'
-
-.. note::
-  This feature is currently incompatible with ``Scan`` and also with ops
-  which do not implement a ``perform`` method.
-
 
 "How do I Print an Intermediate Value in a Function?"
 -----------------------------------------------------


doc/tutorial/debug_faq.txt
```diff
@@ -214,6 +214,13 @@ The ``compute_test_value`` mechanism works as follows:
 * ``ignore``: Silently ignore the computation of intermediate test values, if a
   variable is missing a test value.
 
+Theano also provides the ``print_test_value`` print flag which can hold a boolean value.The default value is false.If 'True', Theano will override the ‘__str__’ method of its variables to also print the tag.test_value when this is available.
```
nouiz: I think this is a little better:

Theano also provides the ``print_test_value`` flag which can take a boolean value.The default value is ``False``. If ``True``, Theano will print the test value when printing Theano variable and it have a test value.

I'm not a native english, maybe is can be made even better. Don't hesitate to change my proposition.

doc/tutorial/debug_faq.txt
```diff
@@ -214,6 +214,13 @@ The ``compute_test_value`` mechanism works as follows:
...
+It can be used as 
+
+.. code-block:: none
+
+    theano.config.print_test_value = 'True'
+
 .. note::
```
nouiz: Can you delete this note? It is outdated.

nouiz:
I'm killing travis as it is overloaded and this won't be useful as it need to run again after the change.

Did you generated the doc and looked at it? Does it look good?

theaverageguy:
I would request a confirmation about the changes. :)

theaverageguy:
I can share a snapshot if you want. You would evaluate better. :)

doc/tutorial/debug_faq.txt
```diff
@@ -214,10 +214,12 @@ The ``compute_test_value`` mechanism works as follows:
 * ``ignore``: Silently ignore the computation of intermediate test values, if a
   variable is missing a test value.
 
-.. note::
-  This feature is currently incompatible with ``Scan`` and also with ops
-  which do not implement a ``perform`` method.
+Theano also provides the ``print_test_value`` print flag which can hold a boolean value.The default value is ``false``.If ``True``, Theano will override the ‘__str__’ method of its variables to also print the tag.test_value when this is available.
+It can be used as
```
lamblin: There should be a space between sentences. Wrapping the lines at 79 characters would be better.
`__str__` should be between backticks (`)

You can refer to the section about how to set flags, since you can do that in a couple of different ways.

lamblin:
I killed Travis because some updates are required, and we are probably not going to merge it before the release, but we can restart it later.

doc/tutorial/debug_faq.txt
```diff
@@ -214,10 +214,12 @@ The ``compute_test_value`` mechanism works as follows:
...
 
-.. note::
-  This feature is currently incompatible with ``Scan`` and also with ops
-  which do not implement a ``perform`` method.
+Theano also provides the ``print_test_value`` print flag which can hold a boolean value.The default value is ``false``.If ``True``, Theano will override the ‘__str__’ method of its variables to also print the tag.test_value when this is available.
+It can be used as 
+
+.. code-block:: none
 
+    theano.config.print_test_value = 'True'
```
fvisin: I suggest to completely remove the code snipped and just refer to the config section, as this doesn't add anything to what is said in the config section and is going to clutter the documentation.

doc/tutorial/debug_faq.txt
fvisin: Please also try to be as concise as possible. What is important here is just to point out that setting the flag allows to recover the test value by accessing the variable, instead of its `tag.test_value` attribute. More details can be found in the description of the flag in the [config section](http://deeplearning.net/software/theano/library/config.html#config.print_test_value). 

While you are at it, please add the following example **to the config section**:
```python
import numpy
import theano
theano.config.compute_test_value='warn'
a = theano.tensor.matrix()
b = theano.tensor.scalar()
a.tag.test_value = numpy.array(range(2*3), dtype='float32').reshape((2,3))
b.tag.test_value = 2

# NORMAL BEHAVIOUR (print_test_value = False)
c = a * b
c 
    Elemwise{mul,no_inplace}.0 
c.tag.test_value
    array([[  0.,   2.,   4.],
           [  6.,   8.,  10.]], dtype=float32)
c.shape
    Shape.0
c.shape.tag.test_value
    array([2, 3]) 
 
# ENABLING PRINT_TEST_VALUE
theano.config.print_test_value=True
c
    Out: Elemwise{mul,no_inplace}.0
    array([[  0.,   2.,   4.],
           [  6.,   8.,  10.]], dtype=float32)
c.shape
    Shape.0
    array([2, 3]) 
```

nouiz:
@theaverageguy do you have time to finish this PR? It would be great to have this merged.

thanks

nouiz:
Super seeded by https://github.com/Theano/Theano/pull/5470

Issue #4196, opened 2016-03-08
Bump tolerance of `test_gc_never_pickles_temporaries`


lamblin:
I've seen it fail 3 times in 2 days, and each time a restart of the test "fixed" it.
We should bump the tolerance a little.

theaverageguy:
What does bump tolerance mean?

nouiz:
Check the test. It compare string length with a tolerance. In the ideal case the length would be exact, but they aren't due to technical details. You can probably run 100x the test and see the failure. Do you know how to run only one tests in Theano?

nosetests a_test_file.py:test_name.

sygi:
Hello,
I took a look at the test, but despite 500+ reruns (both cpu and gpu) couldn't reproduce the issue.
(Can I just change https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_gc.py#L95 128 -> 256, or will that be too big?)

nouiz:
ok.

PR #4203, opened 2016-03-09
(#4196) increase the tolerance of test_gc_never_pickles

sygi:
Increased the tolerance of the test, due to reported false positives.

fix gh-4196


lamblin:
Thanks!
As a good practice, we try to reference the issue number in the text of the PR, but not in the commit messages themselves, as this makes things uglier if the commits are rebased or amended.

PR #4205, opened 2016-03-10
Fix test that fails on the buildbot

abergeron:


nouiz:
On the cuda back-end we can make the ref implementation be GpuCorrMM to keep the tests. But yeh, we can merge when travis pass.

PR #4207, opened 2016-03-10
Cnmem threashold, comment, don't call dnn_version and optimization registration.

nouiz:
A few small fixes.

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -3244,10 +3244,10 @@ CudaNdarray_gpu_init(PyObject* _unused, PyObject* args)
         if (cnmem > 1)
             mem = cnmem * 1024 * 1024;
         else{
-            // Clip to 98% to let memory for the driver.
-            // 98.5% didn't worked in some cases.
-            if (cnmem > .98){
-                cnmem = .98;
+            // Clip to 95% to let memory for the driver.
+            // 98% didn't worked in some cases.
+            if (cnmem > .99){
+                cnmem = .95;
             }
             size_t free = 0, total = 0;
             cudaError_t err = cudaMemGetInfo(&free, &total);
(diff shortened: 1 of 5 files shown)


abergeron:
Ok when travis passes

theano/tensor/nnet/nnet.py
```diff
@@ -756,7 +757,8 @@ def local_logsoftmax(node):
         return [ret]
 
 
-@opt.register_specialize('stabilize', 'fast_compile')
+@opt.register_stabilize('fast_compile')
+@opt.register_specialize('fast_compile')
 @gof.local_optimizer([SoftmaxGrad])
 def local_logsoftmax_grad(node):
     """
```
lamblin: For some reason, this seems to have disrupted the giant cross-entropy optimization.

nouiz:
I updated this PR.

theano/sandbox/cuda/cuda_ndarray.cu
```diff
@@ -3244,10 +3244,10 @@ CudaNdarray_gpu_init(PyObject* _unused, PyObject* args)
         if (cnmem > 1)
             mem = cnmem * 1024 * 1024;
         else{
-            // Clip to 98% to let memory for the driver.
-            // 98.5% didn't worked in some cases.
-            if (cnmem > .98){
-                cnmem = .98;
+            // Clip to 95% to let memory for the driver.
+            // 98% didn't worked in some cases.
+            if (cnmem > .99){
```
lamblin: Wait, that means if someone asks for 100%, they will get 95%, but if they ask for 99%, they will have 99%?
That seems unintuitive to me if what we want is to cap it at 95%.

nouiz: The problem is that there isn't a good way to know the best default. It change by GPUs and other stuff. I think we should have lib.cnmem=1 always work. But we don't want to prevent people to hack there way. Do you have another idea?

I could do cnmem == 1 if you prefer.

lamblin: It can still grow below 95% if needed, right? In that case, I think capping at 95% would be OK.
And if people want to hack their way, then they can use an absolute size in MB.

nouiz: done.

PR #4208, opened 2016-03-10
Fix problems for release

abergeron:
These include:
- cudnn linking on mac
- sparse test failing on linux.

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -362,6 +362,9 @@ class DnnVersion(GpuOp):
     def c_lib_dirs(self):
         return [config.dnn.library_path]
 
+    def c_compile_args(self):
+        return ['-rpath', config.dnn.library_path]
+
     def c_support_code(self):
         return """
 #if PY_MAJOR_VERSION >= 3
--- a/theano/sandbox/gpuarray/dnn.py
+++ b/theano/sandbox/gpuarray/dnn.py
@@ -171,6 +171,9 @@ class DnnBase(COp):
     def c_lib_dirs(self):
         return [config.dnn.library_path]
 
+    def c_compile_args(self):
+        return ['-rpath', config.dnn.library_path]
+
     def c_code_cache_version(self):
         return (super(DnnBase, self).c_code_cache_version(), version())
 
@@ -189,6 +192,9 @@ class DnnVersion(Op):
 
     def c_lib_dirs(self):
         return [config.dnn.library_path]
+
+    def c_compile_args(self):
+        return ['-rpath', config.dnn.library_path]
 
     def c_support_code(self):
         return """
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -338,7 +338,7 @@ AddConfigVar('dnn.include_path',
 
 AddConfigVar('dnn.library_path',
              "Location of the cudnn header (defaults to the cuda root)",
-             StrParam(default_dnn_path('lib64')))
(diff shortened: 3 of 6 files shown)


theano/sandbox/cuda/__init__.py
```diff
@@ -362,6 +362,9 @@ def c_libraries(self):
     def c_lib_dirs(self):
         return [config.dnn.library_path]
 
+    def c_compile_args(self):
+        return ['-rpath', config.dnn.library_path]
```
nouiz: It is not sure the order is kept. Why not use ['-rpath '+config...] instead? Just for safety that we don't change the order of arguments

theano/sandbox/gpuarray/dnn.py
```diff
@@ -171,6 +171,9 @@ def c_libraries(self):
     def c_lib_dirs(self):
         return [config.dnn.library_path]
 
+    def c_compile_args(self):
+        return ['-rpath', config.dnn.library_path]
```
nouiz: idem

theano/sandbox/gpuarray/dnn.py
```diff
@@ -190,6 +193,9 @@ def c_libraries(self):
...
```
nouiz: idem

theano/sandbox/cuda/__init__.py
abergeron: Because -rpath is a long option and can't be glued to its argument.

abergeron: Also, I just saw that you propose adding a space which is also a really bad idea which complicates lots of things (such as -Xcompiler processing in nvcc_compiler that will break if we do this).

nouiz: ok

f0k:
c4d29a1 breaks compilation on Ubuntu 14.04.

```
gcc: error: unrecognized command line option ‘-rpath’

['nvcc', '-shared', '-O3', '-m64', '-Xcompiler', '-fno-math-errno,-Wno-unused-label,-Wno-unused-variable,-Wno-write-strings,-rpath,/usr/local/cuda/lib64,-DCUDA_NDARRAY_CUH=c72d035fdf91890f3b36710688069b2e,-DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION,-fPIC,-fvisibility=hidden', '-Xlinker', '-rpath,/run/shm/jan/.theano/compiledir_Linux-3.13--generic-x86_64-with-Ubuntu-14.04-trusty-x86_64-2.7.6-64/cuda_ndarray', '-I/usr/local/cuda/include', '-I/usr/local/lib/python2.7/dist-packages/numpy/core/include', '-I/usr/include/python2.7', '-I/raid/user/jan/install/Theano-git/theano/gof', '-I/raid/user/jan/install/Theano-git/theano/sandbox/cuda', '-o', '/run/shm/jan/.theano/compiledir_Linux-3.13--generic-x86_64-with-Ubuntu-14.04-trusty-x86_64-2.7.6-64/tmpPtQF69/265abc51f7c376c224983485238ff1a5.so', 'mod.cu', '-L/usr/lib', '-lcudnn', '-lpython2.7', '-lcudart']
```

Maybe it should only be passed to the linker. I'll investigate.

nouiz:
the new rpath end up after -Xcompiler and not after -Xlinker. @abergeron,
can you check?

f0k:
It actually ends up *both* after -Xlinker and after -Xcompiler. I guess it should only be used after -Xlinker. I removed it from `DnnBase` and added it to `nvcc_compiler.py:rpath_defaults` instead, but it still ends up in both places.

Issue #3029, opened 2015-06-12
Flag for deterministic GPU operations.


lamblin:
Some GPU operation use AtomicAdd for efficiency when accumulating numbers in a buffer, for summation operations for instance. The problem is that limited-precision floating-point arithmetic is not exactly associative, and AtomicAdd does not guarantee the order in which the operations will be done.
This means that the results are not exactly reproducible numerically.
Ops concerned by that would be:
- Gemm, Gemv, maybe other BLAS Ops: they do not use AtomicAdd for the moment, but could be more efficient if they did
- AdvancedIncSubtensor1_dev20: uses AtomicAdd, no way to disable it, but AdvancedIncSubtensor1 (slow) is deterministic
- CuDNN convolutions: use AtomicAdd for the gradients, no way to disable it, but other convolution implementations do not use them.

We should have a global theano flag to select between the two behaviours, and maybe a finer-grained control at the Op level itself if feasible.

nouiz:
cuBLAS have a way to get some of its implementation faster by AtomicAdd, we only need to provide that interface, we don't need to implement the GPU code ourself.

pdoongarwal:
@nouiz I would like to would like to work on it? How should i gain background for it?

nouiz:
The first step would be to make a deterministic Theano flag in theano/configdefault.py. Then use it in the opt local_gpu_advanced_incsubtensor1 and local_gpua_advanced_incsubtensor to select which implementation (point 2 above) to use.

After that, we need a way to change point 3. For this, in the cudnn op, I would check the that algo must be deterministic if that flag is true.

point 1 is a possible future case to handle if we want the extra speed from it. So nothing to do about it now.

gyom:
Created a pull request to resolve this issue.

https://github.com/Theano/Theano/pull/4209

ozancaglayan:
Just for information,

I tested a GRU based NMT code with and without that `AdvancedIncSubtensor1_dev20` optimization. Here is the mean minibatch update time for 520 updates on a Tesla K40, CUDA 6.5, without CUDNN:

```
In [19]: opt.mean()
Out[19]: 0.8737307692307692

In [20]: noopt.mean()
Out[20]: 0.88592307692307692
```

So I decided to go without the optimization since it doesn't seem to hurt the performance in my setup.

Thanks!

nouiz:
thanks for the information.

ozancaglayan:
I have another comment here.

[...]

PR #4209, opened 2016-03-10
CCW issue 3029

gyom:
This is an implementation of the flag "deterministic" for CCW issue 3029.

https://github.com/Theano/Theano/issues/3029

Tested and ran successfully, with the help of Fred (thanks !).


nouiz:
Replaced by this PR that do the other deterministic work needed: https://github.com/Theano/Theano/pull/4212

Issue #4141, opened 2016-02-29
Documentation Refactor


fvisin:
The scope of this ticket is to go through the documentation of theano and fix or report mistakes and not easily comprehensible parts of it.

JesseLivezey:
I personally think some of the ```scan``` documentation could be reorganized to make it easier to use. 

There is this page:
http://deeplearning.net/software/theano/library/scan.html
Two main ideas:
 - There are a bunch of examples before the ```scan``` function and signature are really described. I would move the ```reference``` part of the page before the ```Guide``` part of the page. Also ```reference``` could be capitalized if changes are made.
 - In the ```reference``` part, ```map```, ```reduce```, ```foldl(r)```, are all described before ```scan```, but they all reference ```scan``` in their documentation. I think it would make more sense to have ```scan``` at the top of the page, then the other functions, then the ```Guide``` below them.

Then there is this page:
http://deeplearning.net/software/theano/tutorial/loop.html
This page seems somewhat redundant with the ```Guide``` page in that it would probably be used by someone who is trying to figure out how to use ```scan``` versus someone who just needs to remember the order or meaning of arguments and just wants the ```reference```.

gokul-uf:
@fvisin is this ticket open for all or is it assigned to someone? I remember seeing in some other issue that CCWs are usually coassigned to someone. If not, let me know how I could help!

fvisin:
@JesseLivezey Thank you for your suggestions! This ticket was initially meant to address the high level organization of the documentation, but I guess we will also get into the details of some of the pages. Your comments are very helpful, should we not incorporate them in this PR I will open another more focalized issue. Please feel welcome to provide other ideas.

@gokul-uf You are right, issues tagged as `CCW` are usually assigned internally. This one is currently being addressed by #4211 (and in general documentation issues will be usually addressed internally because they require a lot of brainstorming and interaction that is much easier in person)

mgermain:
Just a note about something which is not documented.

In `.theanorc` the way I would put this flag `config.dnn.conv.algo_fwd` is:
```
[dnn]
conv.algo_fwd=small
```
But .theanorc wants max one nesting in sections flags so the right way to do it is:
```
[dnn.conv]
algo_fwd=small
```

mgermain:
Just noticed that the documentation about cuDNN is out of sync in the old and new backend and is out of date.

NEW: doc/library/sandbox/gpuarray/dnn.py
[...]

PR #4211, opened 2016-03-11
[WIP] Documentation Refactor

olimastro:
[WIP] See #4141 for a TODO list.

In a first pass to rework the documentation, we will start with the Installation section. This sections suffers from the main problem that it is very cluttered. For example, the page contains multiple "using the gpu sections" or "update sections". As a first step, the installation page should be split into multiple pages.


doc/install.txt
```diff
@@ -8,43 +8,25 @@ Installing Theano
 Requirements
 ------------
 
-In order to use Theano, the following libraries and software will need
-to be installed (MacOS and Windows users should refer to platform-specific
-instructions below for detailed installation steps):
+Theano is mainly develeopped on 64-bit Linux machines. For other architectures,
```
fvisin: Typo

doc/install.txt
```diff
@@ -8,43 +8,25 @@ Installing Theano
...
+    `NumPy <http://numpy.scipy.org/>`_ >= 1.7.1 < MAX_VERSION
 
-    `SciPy <http://scipy.org>`_ >= 0.11
-        Only currently required for sparse matrix and special functions
-        support, but *highly* recommended. SciPy >=0.8 could work,
-        but earlier versions have known bugs with sparse matrices.
+    `SciPy <http://scipy.org>`_ >= 0.11 < MAX_VERSION
 
-    A `BLAS`_ installation (with Level 3 functionality)
+     `BLAS`_ installation (with Level 3 functionality)
```
fvisin: This should not be indented.

doc/install.txt
```diff
@@ -8,43 +8,25 @@ Installing Theano
...
+Theano is mainly develeopped on 64-bit Linux machines. For other architectures,
+please refer to :ref:`install_others`
 
-    Linux, Mac OS X or Windows operating system
-        We develop mainly on 64-bit Linux machines. other architectures are
-        not well-tested.
+These libraries are common to all Linux installations and should be installed
+prior to using Theano.
 
     Python_ 2 >= 2.6 or Python_ 3 >= 3.3
```
fvisin: I would delete the second Python (just keep the version number)

doc/install.txt
```diff
@@ -69,818 +48,44 @@ The following libraries and software are optional:
     `libgpuarray`_
         Required for GPU/CPU code generation on CUDA and OpenCL devices (see: :ref:`gpuarray`.)
```
fvisin: Let's replace this line with "libgpuarray is the new backend that supports CPU/GPU code generation on CUDA and OpenCL devices. It's installation is optional (Theano can also work with the previous backend, which only supports CPU/GPU code generation on CUDA), but recommended." and delete the second note.

doc/install.txt
```diff
@@ -69,818 +48,44 @@ The following libraries and software are optional:
...
 
-    If you have problems linking with MKL, `Intel Line Advisor
-    <http://software.intel.com/en-us/articles/intel-mkl-link-line-advisor>`_
-    and the `MKL User Guide
-    <http://software.intel.com/sites/products/documentation/doclib/mkl_sa/11/mkl_userguide_lnx/index.htm>`_
-    can help you find the correct flags to use.
+If you encounter any trouble, head to the :ref:`troubleshooting` page.
 
-.. _gpu_linux:
+Finally, there is an additional section if you plan on :ref:`using_gpu`.
```
fvisin: Finally, refer to the dedicated section if you plan on :ref:`using_gpu`.

doc/install.txt
```diff
@@ -69,818 +48,44 @@ The following libraries and software are optional:
...
-3. http://blog.victorjabur.com/2011/06/05/compiling-python-2-7-modules-on-windows-32-and-64-using-msvc-2008-express/
-4. http://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio
-5. http://stackoverflow.com/questions/11182765/how-can-i-build-my-c-extensions-with-mingw-w64-in-python
-6. https://mail.python.org/pipermail/python-announce-list/2014-September/010457.html
-
-
-Generating the documentation
-----------------------------
+We recommend you look at the documentation on the website, since it
+will be more current than the documentation included with the package.
```
fvisin: I am not native, but "up to date" sounds better than "current" to me.

Apart from this, let's move the Generating the documentations section somewhere else. It doesn't belong here. I would suggest to create a new page with these few lines and link it in the [theano documentation](http://deeplearning.net/software/theano/#documentation) list.

doc/install.txt
```diff
@@ -8,43 +8,25 @@ Installing Theano
 Requirements
 ------------
 
-In order to use Theano, the following libraries and software will need
-to be installed (MacOS and Windows users should refer to platform-specific
-instructions below for detailed installation steps):
+Theano is mainly develeopped on 64-bit Linux machines. For other architectures,
+please refer to :ref:`install_others`
```
fvisin: If I recall correctly we decided to put the links to the system-specific installation instructions directly here. 
What I would do is to get rid of the `install_others` page and copy its links here. I suggest something along these lines:

    Theano is mainly developed on 64-bit Linux machines, but we do out best to provide the prerequisites installation instruction for many common OS and architectures:
    - [CentOS](whatever is the link)
    - [Ubuntu](ubuntu prereq installation link)
    ...

    In case your favorite setup is not in this list or you are curious about them, here is the list of the prerequisites of Theano: 
    - Python
    - ...

After the list, new section "Installing Theano" with the other links

fvisin: Let's keep the prerequisites list very minimal and move the "extra text" (e.g. "the CentOS installation It covers how to install Theano (for CPU-based computation only) with the distribution-packaged ATLAS, a free fast implementation of BLAS.") inside the OS-specific installation pages (e.g. in the CentOS installation page).

doc/install.txt
```diff
@@ -69,818 +48,44 @@ The following libraries and software are optional:
...
-    use this static library. Your ATLAS installation might need to be modified
-    to provide dynamically loadable libraries.  (On Linux this
-    typically means a library whose name ends with .so. On Windows this will be
-    a .dll, and on OS-X it might be either a .dylib or a .so.)
-
-    This might be just a problem with the way Theano passes compilation
-    arguments to g++, but the problem is not fixed yet.
+- :ref:`install_basic`
+- :ref:`install_bleedingedge`
+- :ref:`install_others`
```
fvisin: If I am not wrong there should not be a "Platform-specific" installation instruction.

doc/install_basic.txt
```diff
@@ -0,0 +1,47 @@
...
+===============================
+
+These are the 3 simple ways of obtaining Theano, and additional requirements
+will change depending on the way you choose.
+
+pip
+~~~
+    
+The recommended way to obtain the latest stable version of Theano is from PyPI using
+pip_. Use the [sudo] or [--user] option depending on your admin rights.
```
fvisin: This is not informative enough. Remove this line and explain [sudo] and [--user] (more in detail) in a bulleted list after the command.

doc/install_basic.txt
```diff
@@ -0,0 +1,47 @@
...
+If you don't want to bother with anything, Theano can be acquired through
+Anaconda_.
+
+
+.. _pip: http://pypi.python.org/pypi/pip
+.. _anaconda: https://docs.continuum.io/
+
+.. _testing_installation:
+
+Testing your installation
```
fvisin: This should go into another page

fvisin:
I like the way it's getting shaped!

One thing missing on this regard is the separation of prerequisites (with OS-specific instruction) and installation instructions. I don't remember what we converged on at the end of our first meeting. I vaguely remember we decided to separate prerequisites and installation and we accepted to have a small overlap between the two sections. I also seem to remember we decided to have three alternatives for the prerequisites (I guess `sudo apt-get/yum install`, `[sudo] pip install [--user]` and `conda install`) and two main installation instructions (`stable` and `bleeding-edge`). Can you check your notes and confirm this? 

As a first step, try to move the installation instructions (e.g. pip and conda) out of the OS-specific pages and keep working on the "moving around text and sections" part, so that we can discuss the overall structure before you modify the text.

doc/install.txt
olimastro: You mean Python >= 2.6 or >= 3.3 ?

doc/install.txt
olimastro: I wanted to take it off from that section because I agree it doesn't belong there. The only place that would make sense is here http://deeplearning.net/software/theano/#documentation however wouldn't a code snippet be awkward on that page?

doc/install.txt
fvisin: Yes exactly.

doc/install.txt
```diff
@@ -8,43 +8,25 @@ Installing Theano
...
     ``g++``, ``python-dev``
-        Not technically required but *highly* recommended, in order to compile
-        generated C code. Theano `can` fall back on a NumPy-based Python execution
-        model, but a C compiler allows for vastly faster execution.
-        g++ >= 4.2 (for openmp that is currently always used)
-        more recent version recommended!
 
-    `NumPy <http://numpy.scipy.org/>`_ >= 1.7.1
-        Earlier versions could work, but we don't test it.
+    `NumPy <http://numpy.scipy.org/>`_ >= 1.7.1 < MAX_VERSION
```
fvisin: Numpy's current max is 1.10. 
Please add "(tested, other versions might work)"

doc/install.txt
```diff
@@ -8,43 +8,25 @@ Installing Theano
...
 
-    `SciPy <http://scipy.org>`_ >= 0.11
-        Only currently required for sparse matrix and special functions
-        support, but *highly* recommended. SciPy >=0.8 could work,
-        but earlier versions have known bugs with sparse matrices.
+    `SciPy <http://scipy.org>`_ >= 0.11 < MAX_VERSION
```
fvisin: Numpy's current max is 0.17. 
Please add "(tested, other versions might work)"

doc/install.txt
fvisin: I meant to create a new page with the code snipped and just add to the documentation page a link to the new page.

doc/install_basic.txt
olimastro: You mean the testing and/or anaconda? I know we spoke about having a test page all by itself, but a page only for that line of code seems the extreme opposite from going where we are now with everything on one page to every details on separate pages. Extremes are usually not good! Plus the test is specific to what kind of installation you would pick.

fvisin: Yes I was talking about the testing only. Is it really installation dependent? I would have said it's not, we can maybe discuss it IRL. I agree that this is the right place to keep it if it's installation dependent, but in case it's not I'd rather merge and move all the testing instructions in a separate page or in a section in some of the existing pages (but I would be inclined toward the first option).

olimastro:
Yes we did agree to have a small overlap in prerequisites but I realized there were none if you only consider basic stable and bleeding edge (without sudo apt-get/yum). From my notes it is unclear what to do with all these very specific installations (like centOS) and I think we should keep them (the MacOS seems useful). This is where I think install_other has its utilities. Would you scrap all of them? Let's say I take out anaconda and pip out of the MacOS page because it is already in the basic page, but I keep the Mac OS page only for the other information. I feel as a MacOS user I would go straight to the MacOS page and not realize the basic applies to me as well. In this scenario I see two option, I explicitly write on the basic installation that any OS should try that and if you have trouble then try the specific pages OR we introduce redundancies and have pip and anaconda appear on both pages. 

As for the system packages, I found it nice for the first page to contain all the dependencies that are common to all installations (subjective opinion). Then I realized adding the sudo apt-get/yum would compromise this. If it's added to the basic page, it could get confusing on what is the way to obtain those packages, and there is also the case where it could interfere with your local packages and root packages. We could just repeat the packages instruction for every installations way like we discussed, but it would look cumbersome and cluttered again (also an opinion). Should we just add a separate page for someone that wants to use the system packages?

doc/install_basic.txt
olimastro: It is installation dependent because if you clone theano from git you would you use `theano-nose` and through pip you would use `python -c "import theano; theano.test()`

fvisin: We can always use the latter, so we can unify these sections.

fvisin:
I'll come to talk to you on this last point, I think it's better to discuss it in person! :)

fvisin:
As we discussed you can use Sphinx's [include directive](http://docutils.sourceforge.net/docs/ref/rst/directives.html#include) to include text from external files. Note that you can use `start-after` and `end-before` to have more than one section in the same file and include just the part you care about. This way we can create just one single file with all the text and avoid to clutter the repo with multiple files.

doc/install.txt
fvisin: Let's not forget to mention that python >=3.5 is not supported on windows.

fvisin:
It is not in the installation part, but can you remove all the references to `theano-github` from the documentation? They should only be in these two pages: `dev_start_guide` and `index.`

fvisin:
As I mentioned you last week, I found a one-liner for the installation instructions [here](https://pip.pypa.io/en/stable/reference/pip_install/#git). The idea is that we can do:
`pip install [--user] -e git+git:github.com:Theano/Theano.git [--src whatever/path/for/the/source]`

Switching from the "bleeding-edge" installation to the "developer" one becomes basically as easy as adding the `src` argument. Unless there are other differences we can then merge the two paths together.

For what concerns the colors, do they work in backticked text? That would be ideal.

olimastro:
New look to check. Few issues : 

- What is the point of the centOS? By reading it I have noticed it is almost the same exact thing as Ubuntu instead replace `apt-get` by `yum`
- The NVIDIA Jetson TX1 embedded platform instructions are very specific and problematic. They break the flow in the Ubuntu page. Should we scrap them completely or I was thinking we could do (yet another) a page for all these alternatives. In the current there is these Gentoo, Mandriva and Docker Images that are not in this version. 
- The include is great, however, what do you think doesn't follow through...colors! Maybe we should abandon this idea. If you check the Ubuntu page, I am including the instructions from the generic pages and the colors don't follow. It's quite annoying.
- I have left our proposed new line with the html hack and the old code block so we can see the differences between the two at the stable install section on the generic page. 
- Tell me what you think about the Mac page.

doc/install.txt
```diff
@@ -4,891 +4,15 @@
 Installing Theano
 =================
 
+On the following pages you will find different installation instructions depending
+on your needs. Each instruction will take care of the :ref:`requirements`.
```
fvisin: Let's put the reference to the requirements *after* the installation instruction, to make sure people don't go straight there and try to install everything. Also, add something like "if you are curious about them" or similar.

doc/install_ubuntu.txt
```diff
@@ -1,30 +1,33 @@
...
 
-Easy Installation of an Optimized Theano on Current Ubuntu
-==========================================================
 
-For NVIDIA Jetson TX1 embedded platform:
+Ubuntu Installation Instructions
+================================
 
-.. code-block:: bash
+There are two ways of getting the dependencies for Theano. You can either grab them
```
fvisin: Let's be as schematic as possible. Many users will just skip the text and look for links or lists. Let's give them both. Whenever you can use lists. I would do something like this (with links to the sections)

Installation:
- Through apt-get (system packages):
    - [bleeding edge (recommended)](internal link to the bleeding edge section)
    - [stable]()
- Through pip:
    - [bleeding edge]()
    - [stable]()

doc/install_ubuntu.txt
```diff
@@ -1,30 +1,33 @@
...
+as system packages or let pip do everything for you. 
 
-    sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libblas-dev git
-    pip install --upgrade --no-deps git+git://github.com/Theano/Theano.git --user  # Need Theano 0.8(not yet released) or more recent
+There are also two installation modes, the stable version and the bleeding-edge most recent version.
+
+Requirements through System Packages
```
fvisin: I think it's best not to split the installation in requirements and installation. Let's just use the include directive to add as last step the installation.

doc/install_ubuntu.txt
```diff
@@ -1,30 +1,33 @@
...
-    sudo pip install Theano
 
 For Ubuntu 11.04:
 
 .. code-block:: bash
 
     sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ git libatlas3gf-base libatlas-dev
-    sudo pip install Theano
+
+.. include:: install_generic.txt
```
fvisin: The text doesn't get coloured because you don't include the `<style>` instructions.
You can fix it by adding (before this line):
```
.. include:: install_generic.txt
    :end-line: 9
```

I suggest you put the style instructions in a separate file (call it css.txt) and just include it at the beginning of each txt file where you have coloured instructions.

doc/install_generic.txt
```diff
@@ -0,0 +1,110 @@
...
+
+    <sudo> pip install <--user> Theano[test] 
+
+.. role:: code
+
+.. role:: bcode
+
+.. role:: rcode
+
+:rcode:`sudo` :code:`pip install` :bcode:`--user` :code:`Theano[test]`
```
fvisin: I found several ways to get what we want with the colors, none of them is completely satisfactory.

I made a gist with a working example (you have to `include` the first 10 lines before you `include` the installation instructions in `install_ubuntu.txt`):
https://gist.github.com/fvisin/db7bbce7aed39cf1d9d6aee4fa644c1c

If writing in HTML is not too painful, I guess that the last option is the best. Otherwise we will have to disable `html_use_smartypants` in the configs, but I think we want it to be on in the rest of the documentation. Let's discuss it IRL.

fvisin:
Nice job! It is starting to have a shape, but we should speed up or we will never finish. Let's try to something complete on the install pages by the next week.

To answer to your points:
- CentOS is a common distro on servers, so it might be interesting for somebody. Also, since it's there I can imagine Fred would say "why should we get rid of it?" and I don't have an answer to that, unless it's hard to maintain. 
- Since we now have the first installation page with a list, we can probably have a separate page for the NVIDIA Jetson TX1 embedded platform.
- I am not sure I understand this sentence: "In the current there is these Gentoo, Mandriva and Docker Images that are not in this version.".
- I looked into the include + colors problem. Sphinx is not very flexible on this, but I found some workarounds. The only one that works completely is to write HTML code directly, which is not optimal but might be acceptable. The Sphinx alternative I found works on almost everything...the only problem is that the unhelpful Sphinx converts the double dash `--` into an en dash `–`.  See my example.
- For what concerns the mac page (and the ubuntu one):
    - Let's be schematic, see other comment.
    - Unless the testing section has information that are specific of one OS, I would move it into a separate page and just link it here.
    - I would move anaconda into a separate page, it should be OS independent so it can go with the links in the first page I guess.
    - The rest LGTM, we can cut some sentence but that pruning can be done at the end.

fvisin:
This is the install line for the bleeding edge:
`pip install --user -e git+https://github.com/Theano/Theano.git#egg=Theano --src ~/tmp/Theano`

For the rest:
- the stable include seems to have one line too much at the beginning
- move anaconda
- add the colors to the installation instruction for bleeding edge
- `[-e]` should be optional, and goes with `[src]`
- remove the alternative installation instruction (python setup.py develop and so on)
- Ubuntu GPU
    - remove everything but the link to nvidia. Add to download the .deb and follow the instructions
    - remove 6.5
    - Keep from "You must reboot the" onwards
    - include "Once that is done, " (maybe also $CUDA_ROOT instructions, we should ask Fred)
- Generic GPU
    - Move MacOs into MacOs only
- CentOS
    - remove we use pip because
    - move proxy into generic
    - "Test the newly installed packages" move into generic
- MacOs
    - remove anaconda
    - add include at beginning
- Windows
    - fix typo developped
    - remove anaconda
- Update
    - colors and --user
    - check if bleeding edge can be updated through pip

doc/install_generic.txt
```diff
@@ -49,34 +50,15 @@ Bleeding-Edge Installation
...
-
-Using ``develop`` mode is
-preferable to ``install`` as any modifications you make in the checkout
-directory (or changes you pull with Git) will be automatically reflected
-in the "installed" version without re-running ``python setup.py install``.
+.. raw:: html
 
-If you do not have permission to modify your ``site-packages`` directory you
-can specify an alternative installation prefix using
+    <pre><span class="red">&#60;sudo&#62;</span> pip install <span class="blue">&#60;--user&#62;</span> <span class="green">&#60;-e&#62;</span> git+https://github.com/Theano/Theano.git#egg=Theano <span class="green">&#60;--src clonedfolder&#62;</span></pre>
```
mgermain: `clonedfolder` Should be `source_folder_path` or something like that.
Also this command does not work on 2 computer I've tried.
Finally, I like the idea on just one master command but I think the `clone` + `pip -e` should be there also.

fvisin:
Regarding @mgermain comments, after the tests you two did yesterday we can safely say that pip is stupid and it is not possible to update the `--src` installation in an easy way. 

I am afraid we will have to reconsider the current template once again. Let's try to make it the last time!

Todo:

1. We will need 3 installation: stable, bleeding-edge, developer
2. Modify **stable** as follows:
    - specify that whatever is in `<...>` is optional
    - [test] --> [test, doc]
    - specify what *test* and *doc* will add
    - add an optional `no-deps` parameter (*)
    - add the proxy argument (move the information from the test section)
3. Modify **bleeding edge** as follows:
    - add (recommended)
    - keep it as is without `-e` and `--src`
    - add an optional `no-deps` parameter (*)
    - remove the `git clone` instructions
    - add the proxy argument (move the information from the test section)
4. Add a **developer** section:

    ```
      git clone git://github.com/Theano/Theano.git
      cd Theano
      <sudo> pip install <--user> <--no-deps> <-e> <sourcefolder>
      ```
    - move the current comments from bleeding edge to this section
    - add the proxy argument (move the information from the test section)

5. Split the test section in two subsections: test theano, test blas.

6. Add a Throubleshooting page in the root of the documentation of theano
    - Add a link to this page at the end of the testing part of the install instructions
    - Move here the "gfortran" error
    - From the FAQ page, move here 
        - "TypeError: object of type ‘TensorVariable’ "
        - "Out of memory... but not really" (change the name to the error message)
        - "float32 / int{32, 64} gives float64”"
        - "Theano memory/speed trade-off"
    - Add a "Not all the CPU cores are being used by BLAS instructions" with anser "Set the environment variable “OMP_NUM_THREADS=N” to specify to use N threads."

7. Move setting up the GPU before the test and add (optional)
    - Remove "You must also add the ‘lib’ subdirectory (and/or ‘lib64’ subdirectory if you have a 64-bit Linux computer) to your $LD_LIBRARY_PATH environment variable." (wait for confirmation, I should check with Fred)
    - Move the "Once that is done, [...]" part into "Using the GPU" 
    - Remove the reference to system packages to install cuda
    - Move the "To test that it was loaded correctly" into testing (add a testing the GPU (optional) subsection)
    - I think we might want to copy the "Install CUDA" section of the benmccann instructions into this section and then remove the link to his, as they should not be necessary anymore.

(*) when no-deps is specified the dependencies of Theano will not be installed through pip. This is important when they have already been installed as system packages. Consider that the system packages are usually built in a much more optimized way than the ones provided by pip, even if they will be a less updated version.

doc/updating.txt
```diff
@@ -0,0 +1,58 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To update your library to the latest revision, change directory (``cd``)
+to your ``Theano`` folder and execute the following command:
+
+.. code-block:: bash
+
+    git pull
+
+You should update frequently, bugs are fixed on a very regular basis.
```
fvisin: "and features are added even more frequently! "

Let keep it fun! ;)

olimastro:
I have implemented all your comments above except not completely point 7, some are a bit fuzzy or need confirmations. 

That new troubleshooting page seems a bit messy now doesn't it?

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+
+    <pre><span class="red">&#60;sudo&#62;</span> pip install <span class="blue">&#60;--user&#62;</span> <span class="pink">&#60;--no-deps&#62;</span> git+https://github.com/Theano/Theano.git#egg=Theano</pre>
+
+- Use :red:`sudo` for a root installation.
+
+- Use :blue:`user` for a user installation without admin rights. It will install Theano in your local site-packages.
+
+- Use :pink:`no-deps` when you don't want the dependencies of Theano to not be installed through pip. This is important when they have already been installed as system packages. 
+
+Developper Installation
```
fvisin: typo

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+   to pip. The ``[user:pass@]`` portion is optional.
+
+
+.. _stable:
+
+Stable Installation
+-------------------
+
+The recommended way to obtain the latest stable version of Theano is from PyPI using
+pip_. Any argument between <...> means it is optional.
```
fvisin: Move the "Any argument between <...>" part as the first point of the bullet list please.

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+
+- [doc] will add to the install the required packages for the documentation.
+
+.. _bleeding_edge:
+
+Bleeding-Edge Installation (recommended)
+----------------------------------------
+
+The following are general instructions that will set you up with the
+bleeding-edge version of Theano. Any argument between <...> means it is optional.
```
fvisin: Ditto

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+
+- Use :blue:`user` for a user installation without admin rights. It will install Theano in your local site-packages.
+
+- Use :pink:`no-deps` when you don't want the dependencies of Theano to not be installed through pip. This is important when they have already been installed as system packages. 
+
+Developper Installation
+-----------------------
+
+Follow these instructions if you want to setup your Theano to the developper version.
+Any argument between <...> means it is optional.
```
fvisin: Ditto

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+
+.. raw:: html
+
+    <pre><span class="red">&#60;sudo&#62;</span> pip install <span class="blue">&#60;--user&#62;</span> Theano[test, doc]</pre>
+
+- Use :red:`sudo` for a root installation.
+
+- Use :blue:`user` for a user installation without admin rights. It will install Theano in your local site-packages.
+
+- [test] will add to the install the required packages for testing.
```
fvisin: I'd say "[test] will install the requirements for testing"
Modify [doc] in the same way please

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+
+Stable Installation
+-------------------
+
+The recommended way to obtain the latest stable version of Theano is from PyPI using
+pip_. Any argument between <...> means it is optional. 
+
+.. raw:: html
+
+    <pre><span class="red">&#60;sudo&#62;</span> pip install <span class="blue">&#60;--user&#62;</span> Theano[test, doc]</pre>
```
fvisin: [test, doc] should be optional...should we put it in angle brackets <[test, doc]>?

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+- Use :red:`sudo` for a root installation.
+
+- Use :blue:`user` for a user installation without admin rights. It will install Theano in your local site-packages.
+
+- Use :pink:`no-deps` when you don't want the dependencies of Theano to not be installed through pip. This is important when they have already been installed as system packages. 
+
+Developper Installation
+-----------------------
+
+Follow these instructions if you want to setup your Theano to the developper version.
```
fvisin: Same typo

doc/install_generic.txt
```diff
@@ -0,0 +1,130 @@
...
+Any argument between <...> means it is optional.
+
+.. raw:: html
+
+    <pre>git clone git://github.com/Theano/Theano.git
+    cd Theano
```
fvisin: I will do a PR to your branch for this.

fvisin:
I modified a couple of things in a soon-to-be PR and took some notes. I will not have time to discuss them in the next few days I'm afraid, but let's find some time later this week to discuss it IRL.

In the meanwhile, 

1. can you modify the Troubleshooting section to become the usual "question and answer" setting? What I mean is that I expect it to be a list of questions/issues with the corresponding answer/solution. I'd also like to have all the questions as links (with toc should be easy?) at the beginning of the page.
2. Can you update "updating theano" to reflect the division in 3 (i.e. stable, bleeding, develop?). The bleeding edge instruction is the same as the installation, so please pick it from the generic file. The develop should probably be a `git pull --rebase` instead, but warn the user that we expect him to have a basic knowledge of git and understand what he is doing or document himself.
3. The proxy argument is missing

olimastro:
You can check if everything is good, especially the troubleshooting page, I wasn't sure if you only wanted that or to change all the section headers to the questions format as well.

doc/install.txt
```diff
@@ -4,891 +4,17 @@
 Installing Theano
 =================
 
+On the following pages you will find different installation instructions depending
```
mgermain: If the purpose is still to make all this clutter-free I think this sentence is totally unnecessary?

doc/install.txt
```diff
@@ -4,891 +4,17 @@
...
+- :ref:`install_centos6`
+- :ref:`install_macos`
+- :ref:`install_windows`
+- :ref:`install_generic`
+- :ref:`install_others`
 
-In order to use Theano, the following libraries and software will need
-to be installed (MacOS and Windows users should refer to platform-specific
-instructions below for detailed installation steps):
+If you want to update, head to the update page :ref:`updating`.
```
mgermain: To update your current installation see :ref:`updating`.

doc/install.txt
```diff
@@ -4,891 +4,17 @@
...
-
-    python ./doc/scripts/docgen.py
-
-Documentation is built into ``html/``.
-The PDF of the documentation is ``html/theano.pdf``.
-
-
-.. _theano-users: http://groups.google.com/group/theano-users?pli=1
-.. _theano-dev: http://groups.google.com/group/theano-dev?pli=1
+If you are curious about what Theano requires, you can head up
```
mgermain: Maybe this needs to be before the install options and says something like:

Before installing Theano make sure that all the dependencies are installed :ref:\`requirements\`

doc/install_centos6.txt
```diff
@@ -16,44 +16,16 @@ Easy Installation of an optimized Theano on CentOS 6
     investing money (i.e. buying a license to a proprietary BLAS
     implementation), it is possible to gain further performance.
```
mgermain: Maybe adding a line about conda free MKL blas here?

doc/install.txt
fvisin: The point was exactly preventing the user from installing the requirements by himself: pip will install all the dependencies for you in most of the cases, so the prerequisites page is more for people who for some reason want to have control on what they install, rather than for the normal user. Does it make sense?

doc/install_centos6.txt
fvisin: (It's free only for academics....which I realize is what most of Theano users are anyway.) 

We tried to avoid to refer to conda whenever possible. Maybe we should rethink this, but we tried to keep it as straighforward as possible and conda makes it very complicated because is system-agnostic (up to a certain level). The idea was to provide each user with one, unique, consecutive set of instructions. Still, it might be worth to add a reference to conda in the Generic installation instructions page and link it from here. Something along these lines might work:
"If you cannot modify the python environment in your system or prefer to have a separate environment for Theano, you can check out [conda](https://www.continuum.io/downloads), which comes with the most common python packages already installed as well as an mkl blas library which is free for academic users"

doc/install_generic.txt
slefrancois: "using pip_" generates a link when compiled, but it leads to nowhere, at least when the doc is run locally.

doc/install_others.txt
```diff
@@ -0,0 +1,49 @@
...
+
+    These instructions are not kept up to date.
+
+NVIDIA Jetson TX1 embedded platform:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: bash
+
+    sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libblas-dev git
+    pip install --upgrade --no-deps git+git://github.com/Theano/Theano.git --user  # Need Theano 0.8(not yet released) or more recent
```
slefrancois: Should we remove "(not yet released)", I guess this works with the current Theano release.

slefrancois:
The link to download stdint.h on Windows is broken. Here is the valid one:

doc/install_centos6.txt
mgermain: Aren't you confusing Enthought and Conda ? I think that contrary to Enthought, Conda is always free no ?

But ok I totally agree with the rest of what you said.

doc/install.txt
mgermain: Sure, but Theano/Numpy without MKL/OpenBlas/Etc is not crazy useful. Well, depends on what you do I guess.

doc/install_centos6.txt
fvisin: Conda itself is free, but MKL is free for 30 days unless you are an academic.

doc/install_others.txt
fvisin: Good catch, thanks!

doc/install_generic.txt
mgermain: Not sure it's worth it to specify the bracket things. If you never used a command line before you'll have plenty more problem using Theano.

Just a thought.

doc/install.txt
fvisin: That's right. But IIRC numpy/scipy are already compiled/linked against decent OpenBlas implementations in pip. I seem to recall having an interesting conversation about it with @abergeron.

doc/install.txt
```diff
@@ -4,891 +4,17 @@
...
+On the following pages you will find different installation instructions depending
+on your needs. 
 
-Requirements
-------------
+- :ref:`install_ubuntu`
+- :ref:`install_centos6`
+- :ref:`install_macos`
+- :ref:`install_windows`
+- :ref:`install_generic`
```
mgermain: I don't think we need a generic installation section. When are you installing something generically?
And see some people getting confused by "generic" and "others" installs.

doc/install.txt
mgermain: Maybe better than just removing the line you could put. "Supported platforms:"

doc/install_centos6.txt
```diff
@@ -16,44 +16,16 @@ Easy Installation of an optimized Theano on CentOS 6
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
- 1) NumPy (~30s): ``python -c "import numpy; numpy.test()"``
- 2) SciPy (~1m): ``python -c "import scipy; scipy.test()"``
- 3) Theano (~30m): ``python -c "import theano; theano.test()"``
+.. include:: install_generic.txt
+    :start-line: 11
 
 
 Speed test Theano/BLAS
```
mgermain: Shouldn't that be in generic installs ?

mgermain: Or not there at all? Maybe in a FAQ : Why is my theano so slow?

doc/install_generic.txt
mgermain: That's probably too many brackets. Maybe just adding a link. "See here to learn about python optional dependencies." somewhere.

doc/install_generic.txt
```diff
@@ -0,0 +1,96 @@
...
+-------------------
+
+The recommended way to obtain the latest stable version of Theano is from PyPI using
+pip_. Any argument between <...> means it is optional. 
+
+.. raw:: html
+
+    <pre><span class="red">&#60;sudo&#62;</span> pip install <span class="blue">&#60;--user&#62;</span> Theano[test, doc]</pre>
+
+- Use :red:`sudo` for a root installation.
```
mgermain: I think this legend should be put before you use it the first time. Also, it's repeated throughout the file. You only need it once.

doc/install_generic.txt
```diff
@@ -0,0 +1,96 @@
...
+.. _install_generic:
+
+
+Generic Installation Instructions
+#################################
+
+If you want the latest on the edge installation of Theano, you want `bleeding_edge`_.
+Otherwise go for the `stable`_.
+
+.. note::
```
mgermain: I think this might be beyond the scope and a Theano installation page.
Worst case in a FAQ "Why to I get a network error when I install theano?"

doc/install_macos.txt
```diff
@@ -0,0 +1,197 @@
+.. include:: css.txt
+
+.. _install_macos:
+
+
+Mac OS Installation Instructions
```
mgermain: Why is the mac installation way more ... exhaustive and differently formatted?

doc/install_macos.txt
```diff
@@ -0,0 +1,197 @@
...
+
+There are various ways to install Theano dependencies on a Mac. Here
+we describe the process in detail with Canopy, Anaconda, Homebrew or
+MacPorts but if you did it differently and it worked, please let us
+know the details on the `theano-users`_ mailing-list, so that we can
+add alternate instructions here.
+
+.. _theano-users: http://groups.google.com/group/theano-users?pli=1
+
+In academia: Enthought Canopy
```
mgermain: If you are to recommend something like that I would recommend conda, it's free for everyone forever and does not create a default virtualenv that can cause a ton of problems. At least last time I used it.

Actually, using one of those has nothing to fo with OS X it works everywhere O_o

doc/install_macos.txt
```diff
@@ -0,0 +1,197 @@
...
+         `gfxCardStatus <http://codykrieger.com/gfxCardStatus>`__.
+
+Installing the compiler
+-----------------------
+
+Theano officially supports only clang on OS X.  This can be installed
+by getting XCode from the App Store and running it once to install the
+command-line tools.
+
+If you still want to use g++ you can do so by setting its full path in
```
mgermain: This part is not related to installation, and you already say that  only clang is supported on OS X sooo delete of another FAQ ?

doc/install_macos.txt
```diff
@@ -0,0 +1,197 @@
...
+numpy and scipy with hombrew:
+
+.. code-block:: bash
+
+    $ brew tap homebrew/python
+    $ brew install numpy --with-openblas
+    $ brew install scipy --with-openblas
+
+
+Then install theano as usual:
```
mgermain: "As usual" does not make much sense in this context. I mean it's the installation instruction there is not much usual and the text does not add any information anyway.

doc/install_centos6.txt
mgermain: I just saw a copy of this in the Troubleshooting section.

doc/install.txt
nouiz: I also remember very active discussion in numpy mailing list about that. I think they link with resonable blas now.

doc/install_centos6.txt
nouiz: not anymore. now mkl is free for everybody.

doc/install_centos6.txt
nouiz: If so, maybe we can link here to that section and merge both version.

doc/install_generic.txt
nouiz: the new trouble section seem a good place for that.

doc/testing.txt
```diff
@@ -0,0 +1,55 @@
+.. _testing:
```
nouiz: I don't see good reference to here.

nouiz:
At the same time, we should tell in the new gpu back-end, that if we install Theano dev, we need libgpuarray dev version. If we use Theano last release, we should use the tag v-9998

https://github.com/Theano/libgpuarray/issues/183#issuecomment-224015410

thanks

doc/install_generic.txt
fvisin: I'd rather keep it where it, so that each installation instruction is self contained.

doc/install_generic.txt
fvisin: Agreed!

doc/install_macos.txt
fvisin: Mostly because we copied the old documentation, since we don't know mac enough to change it. The same goes for the Windows one. I'd like to merge this PR to give a structure to this section. Those two installation pages can then be fixed the first time one of us installs theano on mac/windows (e.g. next release?).

doc/install_macos.txt
fvisin: Agreed. I am more inclined not to have a reference to Conda at all. I think it goes beyond the scope of the installation instructions (i.e. we should provide *a* way to install Theano, not every possible way to install a python package). 

Nonetheless, I remember you suggested to have a more expanded requirement section with a reference to conda. I don't remember your proposal exactly (hopefully you wrote it somewhere here) but I remember agreeing with you. In case you didn't forget, can you refresh my memory?

doc/install_macos.txt
fvisin: I agree. If we don't support it we should not encourage the users to use it.

doc/install_centos6.txt
fvisin: Yes, this section should be removed.

doc/install_centos6.txt
fvisin: Ok I didn't know that! 
In that case I think we should remove this note and probably rethink the prerequisites section a little. I will comment there.

doc/requirements.txt
```diff
@@ -0,0 +1,61 @@
+.. _requirements:
+
+
+Requirements
+============
+
+Theano is mainly develeopped on 64-bit Linux machines.
```
fvisin: Typo

doc/requirements.txt
```diff
@@ -0,0 +1,61 @@
...
+
+
+Requirements
+============
+
+Theano is mainly develeopped on 64-bit Linux machines. 
+
+These libraries are common to all Linux installations and should be installed
+prior to using Theano.
+
```
fvisin: 1. I suggest we split them in two groups: the requirements that can be installed through pip (i.e. that the user can either install manually or let pip install for him while installing theano) and the others (python, python-dev, g++/clang, cuda, libgpuarray).

2. Move the section on how to install GPU here

3. Add "(optional)" on the side of the name of the optional requirements

fvisin: After discussion with Pascal and Arnaud, we decided to further modify it:
(I will give you more details in person)

1. Add miniconda as the default way to install prerequisites (will need to repeat the libgpuarray instructions)
2. Install MKL if miniconda and linux. Install openblas else.
3. Move here the GPU installation instructions

doc/requirements.txt
```diff
@@ -0,0 +1,61 @@
...
+        for math to show up as images.
+
+    `pydot-ng <https://github.com/pydot/pydot-ng>`_ or `pydot <https://code.google.com/p/pydot/>`_
+        To handle large picture for gif/images.
+	pydot-ng is a pydot compatible replacement that support newer Python.
+
+    `NVIDIA CUDA drivers and SDK`_
+        Required for GPU code generation/execution on NVIDIA gpus
+
+    `libgpuarray`_
```
fvisin: Should be default on dev

doc/install_others.txt
```diff
@@ -0,0 +1,49 @@
...
+Gentoo
+~~~~~~
+
+Brian Vandenberg emailed `installation instructions on Gentoo
+<http://groups.google.com/d/msg/theano-dev/-8WCMn2FMR0/bJPasoZXaqoJ>`_,
+focusing on how to install the appropriate dependencies.
+
+Nicolas Pinto provides `ebuild scripts <https://github.com/npinto/sekyfsr-gentoo-overlay/tree/master/sci-libs/Theano>`_.
+
+Mandriva 2010.2
```
fvisin: Remove this

doc/install.txt
olimastro: I think I would agree with you here, what do you think @fvisin ?

fvisin: Yes, definitely!

doc/install_macos.txt
olimastro: We always went back and forth considering that conda option. I am not sure what's the "official" opinion about that.

doc/install_macos.txt
olimastro: As @fvisin said, we didn't go into that much details here into macos/windows section. We have to change not only that "as usual" but reformat it like the others. Maybe do it after this PR is merged for structure?

doc/testing.txt
olimastro: That is true! I forgot to remove the file. We were thinking of only having the testing section in the troubleshooting page under "How do I test if everything works properly?"

olimastro:
I have made the changes from all these comments except for the windows/macos sections and the requirements (we'll discuss this one in person I assume)

fvisin:
Great thank you! I'll come talk to you tomorrow.

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+    `BLAS`_ installation (with Level 3 functionality)
+        Recommended: MKL, which is free through Conda. Otherwise, we suggest to install OpenBLAS, with the development headers (``-dev``, ``-devel``, depending on your Linux distribution). 
+
+.. _BLAS: http://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms
+.. _Python: http://www.python.org/
+
+(Optional)
+
+    ``g++`` >= 4.2
+    ``python-dev``
```
fvisin: This is rendered on the same line

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+        **Highly recommended.** Theano can fall back on a NumPy-based Python execution model, but a C compiler allows for vastly faster execution.
+
+    `nose <http://somethingaboutorange.com/mrl/projects/nose/>`_ >= 1.3.0
+        Recommended, to run Theano's test-suite.
+
+    `Sphinx <http://sphinx.pocoo.org/>`_ >= 0.5.1, `pygments <http://pygments.org/>`_
+        For building the documentation. LaTeX_ and dvipng_ are also necessary for math to show up as images.
+
+    `pydot-ng <https://github.com/pydot/pydot-ng>`
```
fvisin: The link is broken

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+        To handle large picture for gif/images.
+
+    `NVIDIA CUDA drivers and SDK`_
+        Required for GPU code generation/execution on NVIDIA gpus. See instruction below.
```
fvisin: Add highly recommended to CUDA too.

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+but we ONLY support the installation of the requirements through conda. 
+
+1. If you want fast compiled code (recommended), make sure you have g++ (Windows/Linux) or Clang (OS X) installed.
+
+2. Install Miniconda_.
+
+3. Install mandatory requirements: 
+.. code::
+
+    conda install numpy, scipy, mkl
```
fvisin: no commas

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+
+4. Install the optional requirements available through Conda:
+
+.. code::
+
+    conda install nose, Sphinx, pydot-ng
```
fvisin: no commas and lowercase, should work

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+    `libgpuarray`_
+        Required for GPU/CPU code generation on CUDA and OpenCL devices (see: :ref:`gpuarray`.)
+
+        .. note:: 
+
+            OpenCL support is still minimal for now.
+
+        .. note:: 
+        
+            libgpuarray is the new backend and is not required for the old version.
```
fvisin: Remove this note. Rewrite the first sentence as 
The *new* backend (optional, by default Theano will use the old backend), which allows to generate GPU/CPU code for CUDA and OpenCL devices.

fvisin: (Change it in the requirements list, not here.)

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+
+The first thing you'll need for Theano to use your GPU is Nvidia's
```
fvisin: Transform this verbose instruction into a **list** of operations.
I expect something like:

- Install CUDA
    1. Download CUDA
    2. ...
- Install libgpuarray
    1. Download libgpuarray
    2. ...

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
...
+2. Install Miniconda_.
+
+3. Install mandatory requirements: 
+.. code::
+
+    conda install numpy, scipy, mkl
+
+4. Install the optional requirements available through Conda:
+
+.. code::
```
fvisin: IIRC you don't need the `code` directive if you indent with empty lines before and after. Anyway, something is not parsed correctly here!

doc/requirements.txt
```diff
@@ -0,0 +1,117 @@
+.. _requirements:
+
+
+Requirements
+============
+
+These are the requirements to run Theano. See below for installation instructions.
```
fvisin: Remove this line

doc/install.txt
```diff
@@ -4,891 +4,17 @@
 Installing Theano
 =================
 
+First, have a look at the :ref:`requirements` page before installing.
```
fvisin: First, make sure to have all the *requirements*.

fvisin:
@lamblin @abergeron @mgermain @nouiz  @olimastro Re-adding the conda section for the prerequisites introduced a problem: the system-specific instructions don't make much sense anymore, as they specify how to install the prerequisites with the system package manager.

These are the options I see:

1. Do not provide linux-distribution-specific instructions.
2. Move the ubuntu/centos instructions in the prerequisites page after the conda section and add a link to the prerequisite installation for mac and windows. 
3. Change everything again (sorry @olimastro). We do not link the prerequisite from the first page (install.txt), but instead we *include* it into each system-specific installation. This way all the installation pages will begin with the prerequisites list, continue with the conda installation instructions, have the system specific (not recommended and not supported) alternative to conda, and end with the GPU instructions.

I don't like 2) because the linux instructions will be in the general page while the mac/windows have separate pages. The option I like the best is probably 3), but requires to move things around another time. Please let me know what you think is best.

doc/install_macos.txt
```diff
@@ -0,0 +1,193 @@
...
+MacPorts but if you did it differently and it worked, please let us
+know the details on the `theano-users`_ mailing-list, so that we can
+add alternate instructions here.
+
+.. _theano-users: http://groups.google.com/group/theano-users?pli=1
+
+In academia: Enthought Canopy
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are working in academia, the easiest way to install most of the
```
fvisin: Remove this part, it's the equivalent of the conda instructions. Keep the note.

doc/install_macos.txt
```diff
@@ -0,0 +1,193 @@
...
+
+.. note::
+
+   If you use version 0.6 or later of Theano, we try to automatically
+   link with the Canopy blas version.  Due to Mac OS peculiarities,
+   this requires user intervention.  We detect if the manipulation was
+   done or not and give an error message explaining what to do in case
+   it hasn't been done.
+
+.. include:: install_generic.txt
```
fvisin: For Mac this should go after the homebrew part, as that is an alternative to conda way to install the requirements.

doc/install_macos.txt
nouiz: I think it is important to keep conda in the installation instruction. That way, for people that have different linux OS, it is easy to do it. It is also a very nice way to install everything as a user that don't have root access.

I would kept enthought if the instruction still work, but if it get complicated to maintain, we can discuss to remove it. I would put conda before enthough.

doc/requirements.txt
olimastro: I am not sure I agree. Both of them are already in list where you click and install. Do you really want to clutter it with a subpoint saying download, then another subpoint saying install, repeat those subpoints for the libgpuarrya and then have a third point about the linking?

fvisin: Most of the elements of the list are links to quite useless pages (see e.g. python). In this case we are providing the download links instead, hence my suggestion to have a separate subelement (BTW, I think they changed the cuda webpage. The link should now head to https://developer.nvidia.com/cuda-downloads). 

Apart from this, the verbose instructions after the two links are for CUDA only, so the current structure doesn't make much sense. I agree we can avoid the sublist if its only element is "download", but otherwise I'd much rather have a list than a page of text.

doc/install_macos.txt
```diff
@@ -0,0 +1,193 @@
...
+
+
+Mac OS Installation Instructions
+################################
+
+There are various ways to install Theano dependencies on a Mac. Here
+we describe the process in detail with Canopy, Anaconda, Homebrew or
+MacPorts but if you did it differently and it worked, please let us
+know the details on the `theano-users`_ mailing-list, so that we can
+add alternate instructions here.
```
fvisin: alternative

doc/updating.txt
```diff
@@ -0,0 +1,81 @@
...
+
+    <pre><span class="red">&#60;sudo&#62;</span> pip install <span class="blue">&#60;--user&#62;</span> <span class="pink">&#60;--no-deps&#62;</span> git+https://github.com/Theano/Theano.git#egg=Theano</pre>
+
+- Use :red:`sudo` for a root installation.
+
+- Use :blue:`user` for a user installation without admin rights. It will install Theano in your local site-packages.
+
+- Use :pink:`no-deps` when you don't want the dependencies of Theano to not be installed through pip. This is important when they have already been installed as system packages. 
+
+Developper Installation
```
fvisin: Typo

fvisin:
@olimastro thanks for this new commit.

The inline github conversation is getting messy and some of the points got lost in the mess.
I'll try to wrap it up here so we close it once and for all.

I think the requirements should be split into two groups: the ones installable through conda and the others. This way we can have this structure:
- Requirements through anaconda (recommended)
    - Install conda ..
    - Install mandatory requirements:
        - conda install numpy scipy mkl
    - Install optional requirements:
        - *conda install nose sphinx pydot-ng*
        - *sudo apt-get install python, python-dev, g++*
- Requirement through system packages (not recommended) 
    - *system specific instructions* (e.g., old BLAS manual installation instructions)
- GPU backends
- Theano
The alternative is to move all the "Requirement through system packages" alternative instructions at the very end, in a separate section. Let's discuss it.

**TODO**

1. Implement the previous structure, using those section titles
2. g++ and python-dev are now too far away and it's not clear anymore that the following sentence refers to both. The best would be to prevent the blank line between them, but I don't know if it's possible. If not, keep them on one line one after the other with python-dev first.
3. Add "Highly recommended." to CUDA too
4. Fix the CUDA/libgpuarray lists (see inline discussion + let's talk about it IRL)
5. libgpuarray: when theano stable is installed, a specific version of libgpuarray has to be installed (tag v-9998), while for the bleeding-edge the dev libgpuarray should be used. 
6. libgpuarray should be a mandatory requirements for the *dev* version of the docs.
7. CentOS is missing the requirements page.
8. Add as first point of the bullet list after each theano installation command "Arguments between <...> are optional"
9. "The recommended way to obtain the latest stable version of Theano is from PyPI using +pip_. Any argument between <...> means it is optional." -->  "Install the latest stable version of Theano with:"
10. "The following are general instructions that will set you up with the bleeding-edge version of Theano. Any argument between <...> means it is optional." --> "Install the latest, bleeding-edge, development version of Theano with:"
11. Should we put the link to troubleshooting after each of the three sections?
12. In the mac instructions, "alternate" --> "alternative"
13. In the mac instructions, homebrew should 
    1. be a separate section (requirement through homebrew (not recommended))
    2. not include theano installation instructions
14. The same goes for macports
15. In install.txt, link to testing-theano-with-gpu instead of "using the GPU"
16. Update: fix developper typo
17. Update: keep same section names as the ones of the installation
18. Update: merge the two basic installation by making <--no deps> optional as for the bleeding-edge case.
19. Update: we should add the warning to the bleeding-edge too and possibly make it shorter.
20. Update: move the last sentence ("You should update frequently ..") as second line at the beginning of the page.
21. Update: make the last note a warning

Notes:
- Windows: will be fixed in another PR (there is an open issue already #4643)
- EPD and Canopy have been merged into a unique package. We should not have two different sets of instructions. Also, why don't we drop canopy? Let's keep it simple and stick to Anaconda everywhere.

nouiz:
sudo apt-get install python, python-dev,

Why this is under anaconda? You don't need the OS python and python-dev
when you install anaconda. You just need g++ from the OS in linux

fvisin:
You are right. I put it there as a reminder to check the non-conda installation instructions and forgot about it.

fvisin:
@olimastro please also take care of the sphinx warnings in the next commits.

fvisin:
@olimastro thanks for the new commits! 
Unfortunately there is still some work to do (and the PR could not be merged anyway because it's still missing the rebase), but I'll try to find some time this week to help you close it.

Here are my notes:

**Elements of the previous TODO that are missing or not complete**
6. libgpuarray should be a mandatory requirements for the *dev* version of the docs.
10. "Install the latest, bleeding-edge, development version of Theano with:"
13. In the mac instructions, homebrew should 
    2. not include theano installation instructions
14. The same goes for macports
15. In install.txt, link to testing-theano-with-gpu instead of "using the GPU" --> Link is broken
21. Update: make the last note a warning

**NEW TODO:**

1. Address warnings during compilation
2. Replace MAX_VERSION with the proper version of numpy and scipy. Verify that in the internal documentation, the instructions on how to do a new release include to check the max requirements for numpy and scipy.
3. Requirements through Conda (recommended) --> Requirements *installation* through Conda (recommended)
3. GPU --> GPU *backend installation*
4. In the conda section we could merge the two commands into one, adding the optional requirements after the mandatory ones, with the usual <> syntax. The more compact the better!
5. Remove "Follow the next section to install the GPU backends."
6. Remove the parenthesis from the GPU note ("(Sanity check: ..")
7. I'd rather have all the steps in each section of GPU/libgpuarray as a **list**, not subsections
8. "Any one of here is enough:" doesn't sound correct to me
9. Typo: Bleeindg --> Bleeding
10. libgpuarray instructions should be *inside* the 3 theano installation options, so that there is no need to specify which version of libgpuarray goes with which Theano
11. I would expect CentOS and company to go under Theano installation in the menu
12. Generic installation instructions should be gone, but is still in the menu
13. Every word in the troubleshooting list should be a link
14. Fix the link to download stdint.h on Windows:
https://sourceforge.net/p/mspgcc/msp430-libc/ci/master/tree/include/stdint.h
15. The git commands to install/update Theano exceed the new template width. Is there an easy way to work around this?

**Notes**
- Windows and Mac instructions should definitely be verified. They are too verbose and should be refactored as lists of instructions.
- Windows: will be fixed in another PR (there is an open issue already #4643)
- EPD and Canopy have been merged into a unique package. We should not have two different sets of instructions. Also, why don't we drop canopy? Let's keep it simple and stick to Anaconda everywhere.

doc/tutorial/using_gpu.txt
```diff
@@ -15,17 +15,35 @@ about how to carry out those computations.  One of the ways we take
...
+    * You can use the program nvida-smi to change this policy.
+    * You can choose one specific GPU by specifying 'device=gpuX', with X the
+      the corresponding GPU index (0, 1, 2, ...)
+    * By default, when ``device`` indicates preference for GPU computations,
+      Theano will fall back to the CPU if there is a problem with the GPU.
+      You can use the flag 'force_device=True' to instead raise an error when
+      Theano cannot use the GPU.
+
+You probably need to change the default version of gcc as
+`explained by Benjamin J. McCann <http://www.benmccann.com/blog/installing-cuda-and-theano/>`_ if the package you downloaded is for another Ubuntu version.
```
slefrancois: Is this still valid? The link doesn't seem to mention GCC and talks about CUDA 5.0.

Should we link to the instructions for Ubuntu 16 where it instructs how to use GCC 4.9?

fvisin: I agree, we should link to the latest instructions. We can either remove the others or move them into the page where we put dockers, etc and outdated information.

fvisin:
I am closing it, since this PR was completed by #4775 and merged. Thanks @olimastro!!!

PR #4212, opened 2016-03-11
deterministic flag

nouiz:
follow up of that also make dnn conv and block sparse raise error when it can't be deterministic and if possible, tell how to fix this by the user.

follow up of gh-4209


abergeron:
This is too big of a change for the release.  Also, it would be best to get rid of the stray merge commits with the duplicates that they bring.

nouiz:
Yesterday @lamblin was ok with this idea. I can remove the merge after review and the fixes I'll need to do. Can you still review it now?

lamblin:
That's not what I said, or at least not what I meant.
I don't think we are able to guarantee that the "deterministic" flag will behave as expected before the release, even if we try to rush a bunch of changes.
Also, I think this is an opportunity to push the new back-end, by not adding these features to the old one.
On top of that, the list available deterministic algorithms available on CuDNN will probably expand, so we will need to change that code again, so we might as well think of a cleaner way to organize that.

nouiz:
It took only 10m to port between back-end.

You are right, we can't make enough real use case before the release, but
this cover all user reported problem. They did it manually in there code
otherwise. So I still think it is good to merge. User have done it for us.
I also grepped for atomic and didn't found other cases.

Yes cudnn will change, but that is always the case. Old algo will stay with
there deterministic/non-deterministic behavior. Only new algo need to be
set in the right list. So not too much maintenance.

The list of algo and the characteristic of deterministic is per op. So we
can't make a global list of algo and tell if they are deterministic or not.
So other ways of doing this aren't clearner.

theano/configdefaults.py
```diff
@@ -1681,6 +1681,18 @@ def default_compiledir():
...
     in_c_key=False)
 
+
+# To resolve https://github.com/Theano/Theano/issues/3029.
+AddConfigVar(
+    'deterministic',
+    ("Avoids the use of the AtomicAdd operation which fails to guarantee the"
+     "deterministic properties of the addition because of the behavior of"
+     "floating-point arithmetic."
+     "This slows down execution by a large factor."),
```
lamblin: We should also say that it can trigger crashes if some non-deterministic operations are still in the graph.

abergeron:
I have no input about which backends should this work with, but I still think it is a bad idea to rush in things before a release in general.

I am wary of merging flake8 fixes even and this is a much more involved change so I still think we should wait until after the release.

lamblin:
I still think that this feature is not ready, and not sufficiently tested.

Not ready because:
- we should change the optimizations introducing cudnn convolutions to avoid introducing non-deterministic variants in the first place, not crash afterwards
- it is not clear when something will crash, or when a slower implementation would be used, and it is not possible to choose between those behaviours

Not sufficiently tested because:
- we do not have the feedback from said users that it actually solves their problem without introducing other issues
- I don't think there are any tests that the selected algorithms for CuDNN convolutions are actually deterministic
- There are no tests that setting the flag for convolution will work correctly (there could be a typo in lists of strings, etc.)

Finally, I believe users interested in such a feature, that is still experimental, are savvy enough to get the development version after the release.

nouiz:
Ok to wait for after the release. See bellow.

On Fri, Mar 11, 2016 at 2:01 PM, Pascal Lamblin <notifications@github.com>
wrote:

> I still think that this feature is not ready, and not sufficiently tested.
>
> Not ready because:
>
>    - we should change the optimizations introducing cudnn convolutions to
>    avoid introducing non-deterministic variants in the first place, not crash
>    afterwards
>
> Good idea, but this need to change the semantic of the default algo we
use. It mean that now the default cudnn algo will change depending of the
deterministic flag. I'm ok with that.


>
>    - it is not clear when something will crash, or when a slower
>    implementation would be used, and it is not possible to choose between
>    those behaviours
>
> We could document what isn't supported. I don't understand the last part
of your sentence. Do you mean that we should allow to select manually which
part of the graph is slow but deterministic, while letting other part of
the graph being fast but non-deterministic? If so, I don't see that feature
useful. But you can already do such think for cudnn, the only think not
supported is for the indexing.


>
>
> Not sufficiently tested because:
>
>    - we do not have the feedback from said users that it actually solves
>    their problem without introducing other issues
>
>
They haven't tested the PR for sure, but if they did it, they where having
reproducable results of model that was learning.


>
>    - I don't think there are any tests that the selected algorithms for
>    CuDNN convolutions are actually deterministic
>
>
We should not do that. This is like testing that gcc and nvcc generate
always the right think. This is information that is provided by NVIDIA.


>
>    - There are no tests that setting the flag for convolution will work
>    correctly (there could be a typo in lists of strings, etc.)
>
>
ok.


>
>
> Finally, I believe users interested in such a feature, that is still
> experimental, are savvy enough to get the development version after the
> release.
>

as said above, ok for after the release.

nouiz:
I cancelled travis as this PR need more work.

lamblin:
Great if we agree to delay until the release. Ok to wait for after the release. See bellow.

> >    - it is not clear when something will crash, or when a slower
> >    implementation would be used, and it is not possible to choose between
> >    those behaviours
> 
> We could document what isn't supported. I don't understand the last part
> of your sentence. Do you mean that we should allow to select manually which
> part of the graph is slow but deterministic, while letting other part of
> the graph being fast but non-deterministic? If so, I don't see that feature
> useful. But you can already do such think for cudnn, the only think not
> supported is for the indexing.

I'm not saying to have different parts of the graph behave differently, but for instance, there is no way of selecting the (slower) CPU implementation of blocksparse instead of crashing, except by completely disabling the GPU (which is unhelpful when trying to debug issues on GPU elsewhere in the graph), or disabling the specific lifter for blocksparse.

> >    - I don't think there are any tests that the selected algorithms for
> >    CuDNN convolutions are actually deterministic
> 
> We should not do that. This is like testing that gcc and nvcc generate
> always the right think. This is information that is provided by NVIDIA.

I'm not advocating putting them in unit tests, sorry if that was not clear. I mean running some scripts, after each new version of CuDNN, just to check that algos that are supposed to be deterministic are indeed.

nouiz:
So here is the list of things todo on this PR:

- Allow the flag to have 3 values, True, False, and raise. If True, we should fallback to CPU if we don't have it on GPU.
- We should change the optimizations introducing cudnn convolutions to avoid introducing non-deterministic variants in the first place.
   - If there is a GPU conv that implement that CPU conv in a determinstic way, we should change the \__init__ to force a deterministic algo if a non-deterministic algo was selected
   - If there isn't GPU conv that is deterministic, we should change the opt to not introduce it.
- There are no tests that setting the flag for convolution will work correctly (there could be a typo in lists of strings, etc.)

PR #4216, opened 2016-03-11
Add graph optimization 0/x -> 0

### LINKED ISSUE #4213: Optimization: 0/x -> 0
Discussed in https://github.com/Theano/Theano/issues/4086#issuecomment-186286921, and required for #3758. I'm adding this as an issue so it's not forgotten.

**Arguments against this optimization:**
It will hide NaNs (e.g., 0/0 or 0/nan would become 0).

**Arguments for this optimization:**
There are other optimizations that hide NaNs as well (e.g., for 0 * nan or x / x), and the equivalent 0 * inv(x) is also optimized to 0 already.

**Implementation:**
The optimization of 0*x is part of `theano.tensor.opt.Canonizer`, in its only incarnation `theano.tensor.opt.local_mul_canonizer`. Or maybe it's done by `theano.tensor.opt.local_mul_specialize`, after canonicalization? The latter clearly does it, the former has a lot of code that I don't have the time to read through.

So I guess we'd add a `local_div_of_zero` that looks similar to one of the cases of `local_mul_specialize` (https://github.com/Theano/Theano/blob/5f4eab9/theano/tensor/opt.py#L5382)? If there's consensus, I can do that.

f0k:
Resolves #4213.

This is a simple first attempt. It fixes the problems I encountered in #3758, but based on your warnings, it will probably break other things. Let's see! (I didn't run the full test suite yet.)

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5271,6 +5271,17 @@ def local_intdiv_by_one(node):
             return [node.inputs[0].astype(node.outputs[0].dtype)]
 
 
+@register_canonicalize
+@gof.local_optimizer([T.int_div, T.true_div])
+def local_zero_div(node):
+    """0 / x -> 0
+    """
+    if isinstance(node.op, T.Elemwise) and isinstance(
+            node.op.scalar_op, (theano.scalar.IntDiv, theano.scalar.TrueDiv)):
+        if local_mul_canonizer.get_constant(node.inputs[0]) == 0:
+            return [broadcast_like(0, node.outputs[0], node.fgraph)]
+
+
 @gof.local_optimizer([T.pow])
 def local_pow_specialize(node):
     # here, we are past the point of canonicalization, so we don't want
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -6123,6 +6123,27 @@ class TestIntDivByOne(unittest.TestCase):
         assert len(divs) == 0
 
 
+def test_local_zero_div():
+    """Tests 0/x -> 0"""
+    mode = theano.compile.mode.get_default_mode().including("local_zero_div")
+    for t in (T.scalar, T.ivector, T.ftensor4):
+        x = t('x')
+        for op in (T.int_div, T.true_div):
+            y = op(0, x)
+            g = optimize(FunctionGraph([x], [y]))
+            # the division should be gone
+            divs = [node for node in g.toposort()
+                    if isinstance(node.op, T.elemwise.Elemwise) and
+                    isinstance(node.op.scalar_op, type(op.scalar_op))]
+            assert len(divs) == 0
+            # the output type should match the unoptimized one
+            output = g.outputs[0]
+            assert output.ndim == y.ndim
+            assert output.type == y.type
+            # and the output should be zero
+            assert theano.tensor.get_scalar_constant_value(output) == 0
+
+
 def test_local_expm1():
     x = matrix('x')
     u = T.scalar('u')


f0k:
FWIW, the Lasagne test suite still passes with this.

abergeron:
If we want this, I'm ok with the current code, but don't merge before the release.

theano/tensor/opt.py
```diff
@@ -5271,6 +5271,16 @@ def local_intdiv_by_one(node):
             return [node.inputs[0].astype(node.outputs[0].dtype)]
 
 
+@register_canonicalize
+@gof.local_optimizer([T.int_div, T.true_div, T.floor_div])
+def local_zero_div(node):
+    """0 / x -> 0
+    """
+    if node.op in [T.int_div, T.true_div, T.floor_div]:
```
nouiz: That isn't good. There is some corner cases when the op is another instance. use:
isinstance(node.op, T.Elemwise) and isinstance(node.op.scalar_op, (theano.scalar.IntDiv, ...))

instead.

theano/tensor/opt.py
```diff
@@ -5271,6 +5271,16 @@ def local_intdiv_by_one(node):
...
+        if local_mul_canonizer.get_constant(node.inputs[0]) == 0:
```
nouiz: add the parameter `only_process_constants=True`.

theano/tensor/tests/test_opt.py
```diff
@@ -6123,6 +6123,27 @@ def test3(self):
         assert len(divs) == 0
 
 
+def test_local_zero_div():
+    """Tests 0/x -> 0"""
+    mode = theano.compile.mode.get_default_mode().including("local_zero_div")
+    for t in (T.scalar, T.ivector, T.ftensor4):
+        x = t('x')
+        for op in (T.int_div, T.true_div, T.floor_div):
```
nouiz: floor_div is int_div, so only use one everywhere
scalar/basic.py have 

floor_div = int_div

theano/tensor/tests/test_opt.py
```diff
@@ -6123,6 +6123,27 @@ def test3(self):
...
+            divs = [node for node in g.toposort()
+                    if isinstance(node.op, T.elemwise.Elemwise) and
+                    isinstance(node.op.scalar_op, op)]
+            assert len(divs) == 0
+            # the output type should match the unoptimized one
+            output = g.outputs[0]
+            assert output.ndim == y.ndim
+            assert output.type == y.type
+            # and the output should be zero
+            assert theano.tensor.get_scalar_constant_value(output) == 0
```
nouiz: That is good, but personaly I prefer to compile a full theano function in test and compile it and check its state. I also like to execute the function to make sure the full pipeline work. But all seem good. I'm ok with this test.

theano/tensor/opt.py
f0k: `local_mul_canonizer.get_constant` is a wrapper that does the correct thing already (I copied it from one of the other optimizers, not sure why they don't call `get_scalar_constant_value` directly),

theano/tensor/tests/test_opt.py
f0k: I figured that for testing graph optimizations, it should be enough to just run the optimizer and skip the compilation. This could potentially be applied to some of the other tests to make them faster.

f0k:
Addressed @nouiz's comments and rebased onto latest master so one can test locally without the `-rpath` issues. Ready to merge from my side (but feel free to wait until after the release).

lamblin:
The release is over, so I'm merging.

PR #4217, opened 2016-03-11
Pass -rpath to linker only, not to compiler

f0k:
Fixes compilation failure introduced in #4208, mentioned in https://github.com/Theano/Theano/pull/4208#issuecomment-195521755.

Warning: This does not address the changes in the new backend. Let me know if I can easily extend this PR, or feel free to create your own PR based on mine.

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -362,9 +362,6 @@ class DnnVersion(GpuOp):
     def c_lib_dirs(self):
         return [config.dnn.library_path]
 
-    def c_compile_args(self):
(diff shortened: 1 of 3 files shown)


abergeron:
If that works for the old backend, then I'm ok with it.

For the new backend we should use  `'-Wl,-rpath,' + config.dnn.library_path` instead of what is there.

nouiz:
I also have this problem, we can merge when travis pass (need only the first one, the other are useless.). I'll add an error earlier in another PR.

thanks

nouiz:
enough of travis passed, so merging.

thanks

Issue #4220, opened 2016-03-12
Possible use of uninitialized GPU memory


f0k:
A Lasagne user found a problem with nondeterministic results for a convolutional autoencoder (discussion at https://groups.google.com/d/msg/lasagne-users/bfPPuQBBzmM/TatY1IuZBAAJ).

He provided a demonstration script: https://gist.github.com/sveitser/73984476c8bf3794e3f3
It should be completely deterministic, but when run on GPU with cuDNN, it provides different results at each run, and it is influenced by allow_gc and lib.cnmem. My suspicion is that it reads uninitialized memory somewhere.

I'm reproducing a slightly simplified version of that script here -- unfortunately, it requires Lasagne (bleeding-edge, not 0.1):
```python
from lasagne import layers, objectives, updates
import numpy as np
import theano
import theano.tensor as T

np.random.seed(0)

input_var = T.tensor4('X')
target_var = T.tensor4('X2')

input = layers.InputLayer((None, 1, 5, 5), input_var)
conv = layers.Conv2DLayer(input, num_filters=64, filter_size=(3, 3), pad=1,
                          nonlinearity=None, b=None)
pool = layers.MaxPool2DLayer(conv, pool_size=(3, 3), stride=(2, 2))
inverse_pool = layers.InverseLayer(pool, pool)
network = layers.InverseLayer(inverse_pool, conv)

prediction = layers.get_output(network)
loss = objectives.squared_error(prediction, target_var).mean()

g = theano.grad(loss, input_var)
f = theano.function([input_var, target_var], {'out': prediction, 'grad': g})
X = np.random.randn(1, 1, 5, 5).astype(np.float32)

print(f(X, X))
print(f(X, X))
print(f(X, X))
```
An InverseLayer(A, B) takes its input from layer A and passes it through the gradient implementation of layer B (obtained via theano.grad).

Running this gives some very weird behaviour:
```
THEANO_FLAGS=device=gpu1,allow_gc=1,lib.cnmem=0 python grad2.py 
Using gpu device 1: Tesla K40c (CNMeM is disabled, CuDNN 4004)
{'grad': array([[[[-0.0044742 ,         nan,         nan,         nan,  0.00493809],
         [ 0.09747709,         nan,         nan,         nan,  0.02103304],
         [-0.00183207,         nan,         nan,         nan,  0.08208349],
         [ 0.02717943,         nan,         nan,         nan,  0.05635365],
[...]

PR #4222, opened 2016-03-12
Opt fix

nouiz:
fix gh-4220

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -21,6 +21,7 @@ from theano.sandbox.cuda import dnn_version as version
 from theano.sandbox.cuda.basic_ops import (as_cuda_ndarray_variable,
                                            host_from_gpu,
                                            gpu_contiguous, HostFromGpu,
+                                           gpu_alloc, GpuAlloc,
                                            gpu_alloc_empty, GpuAllocEmpty,
                                            GpuElemwise)
 from theano.sandbox.cuda.blas import (GpuConv, GpuDownsampleFactorMax,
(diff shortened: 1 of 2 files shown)


theano/sandbox/cuda/opt.py
```diff
@@ -1942,7 +1942,9 @@ def local_gpu_downsample_factor_max_grad_grad(node):
     if isinstance(node.op, pool.DownsampleFactorMaxGradGrad):
         assert node.op.__props__ == ('ds', 'ignore_border', 'st',
                                      'padding', 'mode')
-        if node.op.padding != (0, 0) or node.op.mode != 'max':
+        if (node.op.padding != (0, 0) or
+                node.op.mode != 'max' or
+                node.op.st != (1, 1)):
```
f0k: :+1:

theano/sandbox/cuda/dnn.py
```diff
@@ -2246,9 +2247,13 @@ def local_dnn_conv_inplace(node):
         inputs = list(node.inputs)
         dest = inputs[2]
         if (dest.owner and
-                isinstance(dest.owner.op, GpuAllocEmpty) and
+                type(dest.owner.op) is GpuAllocEmpty and
```
f0k: What exactly is the difference?

lamblin: `GpuAlloc` inherits from `GpuAllocEmpty`, so the previous check would have matched `GpuAlloc` nodes as well.

lamblin:
Are those fixes needed on the new back-end as well?

theano/sandbox/cuda/dnn.py
f0k: > `GpuAlloc` inherits from `GpuAllocEmpty`

Oh, I forgot. Wouldn't it be better to just swap the cases then, so we can still use `isinstance`?
```python
if ... and isinstance(..., GpuAlloc) and ...:
    ...
elif ... and isinstance(..., GpuAllocEmpty) and ...:
    ...
```
Directly checking for the type could be problematic if additional subclasses of GpuAlloc or GpuAllocEmpty are introduced.

nouiz:
They aren't needed in the new back-end as they don't inherit from each other.

theano/sandbox/cuda/dnn.py
nouiz: In the new back-end, we use type equality. We don't want subclass to get optimized by default. We prefer them to tell explicitly they want to be optimized. This is safer (we prefer to risk less optimized graph then bugged graph). Doing that here isn't a problem I think.

theano/sandbox/cuda/opt.py
lamblin: Actually, I think this should be `node.op.st != node.op.ds`.
A fix is on the way once I check that it's correct.

lamblin: See #4249.

Issue #3487, opened 2015-10-03
Support for negative indices for `addbroadcast` and `unbroadcast`


eulerreich:
This code

```
M = T.matrix()
print M.broadcastable
print T.addbroadcast(M, 1).broadcastable
print T.addbroadcast(M, -1).broadcastable
```
gives the output

```
(False, False)
(False, True)
(False, False)
```

One should expect `(False, True)` for the last line.

Similarly, this code
```
M = T.matrix()
print M.broadcastable
N = T.addbroadcast(M, 1)
print N.broadcastable
print T.unbroadcast(N, -1).broadcastable
```
gives the output
```
(False, False)
(False, True)
(False, True)
```
where one expects `(False, False)` in the last line.

nouiz:
Thanks for the suggestion. We don't have time for this in the short term. Do you want to try to implement it? It isn't complicated. The change are local to the class Rebroadcast in the file theano/compile/ops.py. You need to modify the make_node method.

khalibartan:
@eulerreich If you are not working can you give me the code to reproduce the error, I'll dig into the things .

maniteja123:
Hi devs, I am a beginner to theano and was tying the broadcast methods in theano. It was then that I realized something similar to the negative indices in the numpy is not the scenario in theano. Looking up for it, I found this issue.

If I understand this issue right, it should support ``addbroadcast`` and ``unbroadcast`` for negative indices. But something like

    M = T.matrix()
    print M.broadcastable
    >>>(False, False)
    N = T.addbroadcast(M, 0, 1)
    print N.broadcastable
    >>>(True, True)
    N = T.addbroadcast(M, 1, -1)
    print N.broadcastable

Should this just give ``(False, True)`` instead of any warning ? what is the rationale behind handling these cases ?

I am new to theano. Apologize in advance if I misunderstood something.

nouiz:
Your example is right. This issue is about implementing this. Currently
Theano support negative indexing in many places, but not this one. So until
someone implement this, use indixing >= 0 with those functions.

So mostly the rational is that we didn't taught of this when they were
implemented and later when it was dicovered to not be supported, nobody
worked on that. If you or someone else want to contribute this to Theano,
it would be great.

maniteja123:
Thanks for the reply ! I would be happy to contribute on this.

[...]

PR #4225, opened 2016-03-13
Fix for #3487.

abhishekgarg16:
Fix for [#3487](https://github.com/Theano/Theano/issues/3487)

Added support for negative indices for 'addbroadcast' and 'unbroadcast'.

When addboadcast function is called with a negative value, the corresponding positive indices is found and it is made braodcastable.

### DIFF
--- a/theano/compile/ops.py
+++ b/theano/compile/ops.py
@@ -675,6 +675,7 @@ class Rebroadcast(gof.Op):
     def make_node(self, x):
         if self.axis.keys() and (x.ndim <= max(self.axis.keys())):
             raise ValueError('Trying to rebroadcast non-existent dimension')
+        self.axis = OrderedDict([(k % x.ndim, v) if k < 0 else (k, v) for k, v in self.axis.items()])
         t = x.type.clone(
             broadcastable=[self.axis.get(i, b)
                            for i, b in enumerate(x.type.broadcastable)])


theano/compile/ops.py
```diff
@@ -675,6 +675,7 @@ def __str__(self):
     def make_node(self, x):
         if self.axis.keys() and (x.ndim <= max(self.axis.keys())):
             raise ValueError('Trying to rebroadcast non-existent dimension')
+        self.axis = OrderedDict([(k % x.ndim, v) if k < 0 else (k, v) for k, v in self.axis.items()])
```
nouiz: can you rename k, v to axis and broad as in the \__init__ method? This is more clear.

We do axis + x.ndim when we are in negative number. For consistency, can you do that here too instead of modulo?

This would also need tests. Normally, they should be in theano/compile/tests/test_ops.py, but for history and a few other reason, the current rebroadcast tests are in theano/tensor/tests/test_basic.py in the class test_broadcast. Can you add them there?

thanks for the PR.

abhishekgarg16:
Ok, i will make the required changes and submit the pull request.

abhishekgarg16:
@nouiz , I have made the changes as told by you, please review this pull request and let me know if any further changes need to be made.

theano/tensor/tests/test_basic.py
```diff
@@ -6747,6 +6747,15 @@ def test_unbroadcast_addbroadcast(self):
 
         assert unbroadcast(addbroadcast(x, 0), 0) is x
         assert addbroadcast(unbroadcast(x, 0), 0) is not x
+        
+        #test for negative indices
+        assert addbroadcast(x, -1) is not x
+        assert addbroadcast(x, -2) is not x
+        assert addbroadcast(x, -3) is x
```
nouiz: This should raise an error as this is an invalid case.

theano/compile/ops.py
```diff
@@ -675,6 +675,8 @@ def __str__(self):
     def make_node(self, x):
         if self.axis.keys() and (x.ndim <= max(self.axis.keys())):
             raise ValueError('Trying to rebroadcast non-existent dimension')
+        temp_dict = OrderedDict([(axis + x.ndim, broad) if axis < 0 else (axis, broad) for axis, broad in self.axis.items()])
+        self.axis = temp_dict
```
nouiz: Here add an assert that all axis are >=0. If not, reaise an error like on line 677.

abhishekgarg16: @nouiz , probably you are mistaken in saying that axis >= 0, because the whole point was about adding support for negative indices.

nouiz: I mean, after the correction, you can still have negative axis. For example, if you have a matrix (2d) and use axis = -3. After correction, axis will be -1, which is not valid.

sudk1896:
@nouiz: Hey, I'd like to take a stab at this. What do I need to do, to complete this issue ?

nouiz:
Start from the code in this PR and do the comments that was done on the PR.

thanks.

Issue #1064, opened 2012-11-12
Theano detect max GPU used by subprocess


nouiz:
It would be good that Theano could detect when a process import it, then call a subprocess that also import it. In most cast, this would make the 2 process lock a GPU if the GPU are in exclusive mode like on clusters.

The user could set the env variable THEANO_MAX_GPU=1, when we init the GPU, we lower it to 0. Then the sub process would inherit the 0 value and would raise an error.

This would prevent problem on Cluster

gokul-uf:
@nouiz, this can be done by using the `os.environ` dictionary. The following code snippet should do the job
```
import os
if os.environ['THEANO_MAX_GPU'] == 0:
    raise MaxGPUError
else:
    os.environ['THEANO_MAX_GPU'] = 0
```

Where should I add these lines?
Also, now that we are working towards supporting multiple GPUS, doesn't having THEANO_MAX_GPU = no_of_GPUs make more sense than setting it to 1?

gokul-uf:
@nouiz , I shall add it at the end of `__init__.py`

nouiz:
I suppose that subprocess shouldn't use the GPU by default. So if the user don't set it, the default, then subprocess shouldn't make an error.

In fact, I think we should only raise an error if we init the GPU in the sub process. This is done in the method use() in the file theano/sandbox/cuda/__init__.py and init_dev() in theano/sandbox/gpuarray/__init__.py.

abergeron:
This is a poor solution.  If you fork before importing Theano, then we can't setup the environment to notice that a gpu is already in use and if you fork after importing Theano, then you fork with a cuda context active which is most likely to lead to strange crashes and/or lockups.

So basically you would still have to be aware of your GPU use and be very careful about it.  So I don't think this is worth doing.

PR #4226, opened 2016-03-13
Prevents subprocess from importing Theano

gokul-uf:
This PR fixes issue #1064  by adding a new environment variable to keep track of whether the GPU is locked by a parent process or not

### DIFF
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -24,6 +24,16 @@ maintained at http://www.deeplearning.net/software/theano/library/
 
 __docformat__ = "restructuredtext en"
 
+# Check if theano is being imported from a child process
+# If yes, raises error, refer issue #1064
+import os
+if 'THEANO_MAX_GPU' in os.environ:
+    if os.environ['THEANO_MAX_GPU'] == '0':
+        raise ImportError("Parent process holds lock on GPU."
+                          " Need to release it first")
+else:
+    os.environ['THEANO_MAX_GPU'] = '0'
+
 # Set a default logger. It is important to do this before importing some other
 # theano code, since this code may want to log some messages.
 import logging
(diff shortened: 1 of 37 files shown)


theano/__init__.py
```diff
@@ -24,6 +24,16 @@
 
 __docformat__ = "restructuredtext en"
 
+# Check if theano is being imported from a child process
+# If yes, raises error, refer issue #1064
+import os
+if 'THEANO_MAX_GPU' in os.environ:
```
nouiz: This would need a test in theano/tests/test_init.py

In that test, make a sub process that import theano and make sure it raise an error. It should suppose theano tests was started without that env variable.

nouiz:
We should not merge this PR before the release, just to be sure there isn't unforeseen problems.

gokul-uf:
@nouiz  I'll shift the code to theano/sandbox/cuda/init.py and theano/sandbox/gpuarray/init.py.
 Forgive my ignorance, but I was not able to find test_init.py (theano/tests/test_init.py). 

Also, won't it be better to start another branch and work on it while the master branch is frozen for release?

nouiz:
We try to make the release this week. So freezing master isn't too much a
problem. If we make this take longer, we will probably switch to use branch.

My comment for the test file was before I told you to move it. I would put
the tests in sandbox/{cuda,gpuarray}/tests/test_basic*.py now.

gokul-uf:
@nouiz I have incorporated your suggestions, waiting for Travis to complete

theano/sandbox/cuda/__init__.py
```diff
@@ -19,6 +19,15 @@
 
 from theano.tensor.basic import register_transfer
 
+# Check if theano is being imported from a child process
```
nouiz: We need to be able to import the cuda and gpuarray module while not using them for computation.

So you must put them in the use() (for cuda module) and in init_dev() (for the gpuarray module) function. Not at the top level of those module.

Issue #4057, opened 2016-02-15
flake8 tensor/nnet/test*.py


nouiz:

fvisin:
This ticket is to fix all the PEP8 errors. This includes making sure that all the functions and classes have comprehensible and exhaustive docstrings.

abergeron:
The PR that fixes this is https://github.com/Theano/Theano/pull/4592

PR #4228, opened 2016-03-14
flake8 tensor/nnet/test*.py

nke001:
Modified code so that it complies with flake8 format.

fix gh-4057


nouiz:
This PR need a rebase.

lamblin:
After a quick look:
- this has to be rebased
- the remaining errors have to be fixed
- some statements that have been commented out should be restored. Usually there of the form `a = f(x)`, and flake8 complains that `a` is not used. The solution should be to just call `f(x)` without assigning the returned value, because usually *calling f* is part of the tests.

nouiz:
This PR need a rebase.

PR #4231, opened 2016-03-14
Fix the last problems identified for the release.

abergeron:
This completely fixes the -rpath problem and it works on mac and linux, with or without cudnn present.

I might add a few fixes for windows onto this later (if it isnt merged yet).


nouiz:
we can merge when a few tests from travis test pass.

abergeron:
There was a flake8 problem that I fixed and tested locally.  I'll merge so that the buildbot uses this.

Issue #4002, opened 2016-02-05
Test the CPU specific instruction set before using it


nouiz:
It happen frequently that the g++ tool chain is broken and don't accept the specific instruction set it tell -mnative use. This issue is about testing them with a few functions to be sure it work. If it don't print a warning and don't use them.

Test it with those methods that generated that problem:
- psi
- [Elemwise{Composite{(i0 + (-sqr(i1)))}}(TensorConstant{(1,) of -1..3787706641}, x)]

This is a work around as we aren't able to rapidly identify and work around all the bugged g++ version.
The error message for now can tell to try those flags:

    gcc.cxxflags='-march=core2"
    gcc.cxxflags='-march=corei7"
    gcc.cxxflags='-march=corei7-avx"

and take the last that worked.

related to gh-3780

nouiz:
I made an issue to test the flags g++ give us before using them. This should help preventing this problem:

https://github.com/Theano/Theano/issues/4002

PR #4232, opened 2016-03-14
added function to look for working march

Faruk-Ahmed:


theano/gof/cmodule.py
```diff
@@ -1787,6 +1787,28 @@ def _try_flags(flag_list, preambule="", body="",
...
+                float Nx = 5.0;
+                int Sx = 25.0;
+                double r = Nx + sqrt(Sx);
+                if ((r - 10.) > 1e-6 || (r - 10.) < -1e-6)
+                {
+                    return -1;
+                }
+                return 0;
+            }
+            """)
```
ballasn: the c test code  should correspond to one of the use case define in the ticket http://github.com/Theano/Theano/issues/4002

theano/gof/cmodule.py
```diff
@@ -1787,6 +1787,28 @@ def _try_flags(flag_list, preambule="", body="",
...
+    cflags =  flags + ['-L' + d for d in theano.gof.cmodule.std_lib_dirs()]
+    res = GCC_compiler.try_compile_tmp(
+        test_code, tmp_prefix='try_blas_',
```
ballasn: tmp_prefix should be 'try_march' instead of 'try_blas'

theano/gof/cmodule.py
```diff
@@ -2004,6 +2026,21 @@ def join_options(init_part):
         if GCC_compiler.march_flags:
             cxxflags.extend(GCC_compiler.march_flags)
 
+        # Find working march flag:
+        verbose = True
```
ballasn: Is verbose used somewhere?

theano/gof/cmodule.py
```diff
@@ -2004,6 +2026,21 @@ def join_options(init_part):
...
+        march_flags_to_try = [cxxflags[0], '-march=core2', '-march=corei7', '-march=corei7-avx']
+        for march_flag in march_flags_to_try:
+            print 'Trying march {}'.format(march_flag)
```
ballasn: you should probably remove this print

theano/gof/cmodule.py
```diff
@@ -2004,6 +2026,21 @@ def join_options(init_part):
...
+            cxxflags_test = cxxflags[:]
+            cxxflags_test[0] = march_flag
+            result = try_march_flag(cxxflags_test)
+            if result[0] and result[1]:
+                print ' -- march {} worked'.format(march_flag)
```
ballasn: same here

theano/gof/cmodule.py
```diff
@@ -2004,6 +2026,21 @@ def join_options(init_part):
         if GCC_compiler.march_flags:
             cxxflags.extend(GCC_compiler.march_flags)
 
+        # Find working march flag:
```
nouiz: We should first test the flags detected by Theano on line 2020. If they work, we use it. If not, try the alternative flags.

theano/gof/cmodule.py
```diff
@@ -2004,6 +2026,21 @@ def join_options(init_part):
...
+        verbose = True
+        march_flags_to_try = [cxxflags[0], '-march=core2', '-march=corei7', '-march=corei7-avx']
```
nouiz: Reverse the order of the strings we try. It should be done earlier as wrote before as we also want to cache the try of the flags.

Faruk-Ahmed:
PEP8ing. :/

theano/gof/cmodule.py
```diff
@@ -2022,24 +2023,25 @@ def join_options(init_part):
...
+            for march_flag in march_flags_to_try:
+                test_flags = GCC_compiler.march_flags[:]
+                test_flags[0] = '-march=' + march_flag
+                result = try_march_flag(test_flags)
+                if result[0] and result[1]:
+                    GCC_compiler.march_flags[0] = march_flag
+                    GCC_compiler.march_flags[-1] = '-mtune=' + march_flag
+                    break
+            if not result[0] or not result[1]:
+                sys.exit('Could not find a working march')
```
ballasn: This is a little bit drastic, we should raise a ValueError instead.

nouiz: If we don't find any, we should not use one. Theano need to continue to
work on other type of CPU like arm CPU.

ballasn:
seems good to me, @nouiz could you have a look to see if we should merge?

theano/gof/cmodule.py
```diff
@@ -2000,6 +2025,20 @@ def join_options(init_part):
                     _logger.info("g++ -march=native equivalent flags: %s",
                                  GCC_compiler.march_flags)
 
+            # Find working march flag:
+            default_detected_flag = GCC_compiler.march_flags[0][7:]
```
nouiz: Why the indexing [0][7:]?

I would just use "GCC_compiler.march_flags"

You don't know where march and mtune will be in that set of flags. This won't be consistent on all platform. So you need to find where they are and change them. Also changing just them could still cause the problem users had due to the other flags. So we need to test with just change of march and mtune or just the extra flags. (Or just test the extra flag)

theano/gof/cmodule.py
```diff
@@ -2000,6 +2025,20 @@ def join_options(init_part):
...
+            for march_flag in march_flags_to_try:
+                test_flags = GCC_compiler.march_flags[:]
+                test_flags[0] = '-march=' + march_flag
+                result = try_march_flag(test_flags)
+                if result[0] and result[1]:
+                    GCC_compiler.march_flags[0] = '-march=' + march_flag
+                    GCC_compiler.march_flags[-1] = '-mtune=' + march_flag
+                    break
+            if not result[0] or not result[1]:
+                raise ValueError('Could not find a working march')
```
nouiz: Don't raise an error, do: GCC_compiler.march_flags = []

We need it to still work on computer that aren't X86.

theano/gof/cmodule.py
mgermain: @nouiz  Someting like that ?
```
default_detected_flag = None
for march_flag in GCC_compiler.march_flags:
    if 'march' in march_flag:
        default_detected_flag = march_flag[march_flag.index('=')+1:]
        break

march_flags_to_try = ['corei7-avx', 'corei7', 'core2']
if default_detected_flag is not None:
    march_flags_to_try = [default_detected_flag] + march_flags_to_try
```

theano/gof/cmodule.py
```diff
@@ -1788,6 +1788,31 @@ def _try_flags(flag_list, preambule="", body="",
...
+                }
+                return 0;
+            }
+            """)
+
+    cflags = flags + ['-L' + d for d in theano.gof.cmodule.std_lib_dirs()]
+    res = GCC_compiler.try_compile_tmp(
+        test_code, tmp_prefix='try_march_',
+        flags=cflags, try_run=True)
+    return res
```
nouiz: Make this res[0] and res[1].

theano/gof/cmodule.py
nouiz: if not try_march_flag(GCC_compiler.march_flags):
   for march_flag in ['corei7-avx', 'corei7', 'core2']:
      GCC_compiler.march_flags = march_flags
      if try_march_flags(march_flag):
         break

theano/gof/cmodule.py
```diff
@@ -2025,19 +2031,31 @@ def join_options(init_part):
...
+                GCC_compiler.march_flags[mtune_ind] = '-mtune=' + march_flag
+                compilation_result, execution_result = try_march_flag(GCC_compiler.march_flags)
+                if compilation_result and execution_result:
+                    working_flag = True
                     break
-            if not result[0] or not result[1]:
-                raise ValueError('Could not find a working march')
+
+            if not working_flag:
+                GCC_compiler.march_flags = []
```
nouiz: If all failed, try:

   for i in ['corei7-avx', 'corei7', 'core2']:
      compilation_result, execution_result = try_march_flag(["-march=" + i])
      ...

it is possible that it is the others flags that pose problems.

theano/gof/cmodule.py
```diff
@@ -2025,19 +2031,31 @@ def join_options(init_part):
...
             march_flags_to_try = [default_detected_flag, 'corei7-avx', 'corei7', 'core2']
             for march_flag in march_flags_to_try:
-                test_flags = GCC_compiler.march_flags[:]
-                test_flags[0] = '-march=' + march_flag
-                result = try_march_flag(test_flags)
-                if result[0] and result[1]:
-                    GCC_compiler.march_flags[0] = '-march=' + march_flag
-                    GCC_compiler.march_flags[-1] = '-mtune=' + march_flag
+                GCC_compiler.march_flags[march_ind] = '-march=' + march_flag
+                GCC_compiler.march_flags[mtune_ind] = '-mtune=' + march_flag
```
nouiz: I'm  not convinced that in 100% of the case march and mtune will be done. Put those 2 lines with a condition for each of them like "if march_ind is not None:"

theano/gof/cmodule.py
```diff
@@ -2000,6 +2031,36 @@ def join_options(init_part):
...
+            # Find working march flag. If all fail, set march_flags to [] (for x86 archs).
+            march_ind = None
+            mtune_ind = None
+            working_flag = False
+
+            for m_ in xrange(len(GCC_compiler.march_flags)):
+                march_flag = GCC_compiler.march_flags[m_]
+                if 'march' in march_flag:
+                    march_ind = m_
+                if 'mtune' in GCC_compiler.march_flags[m_]:
```
nouiz: elif 'mtune' in march_flag:

nouiz:
This look good. Ok to merge when travis pass. Thanks.

PR #4234, opened 2016-03-15
Manually set cuda_ndarray.__file__ if it is absent

cai-lw:
Python 3 does not necessarily set `__file__` when importing (See [Python documentation](https://docs.python.org/3/reference/import.html#__file__)), and it happens that it is not set on my system (Windows 10, Python 3.4) and causes an exception.

### DIFF
--- a/theano/sandbox/cuda/type.py
+++ b/theano/sandbox/cuda/type.py
@@ -21,11 +21,16 @@ try:
     import cuda_ndarray.cuda_ndarray as cuda
     from theano.sandbox.cuda.nvcc_compiler import NVCC_compiler
     import cuda_ndarray
-    # Python 3 does not necessarily set __file__. May need manual setting
+    # Python 3 does not necessarily set __file__. May need manual setting.
+    # The problem is known to occur on Windows 10 with Python 3.4 installed by Anaconda.
     try:
         cuda_ndarray.__file__
     except AttributeError:
-        cuda_ndarray.__file__ = os.path.join(cuda_ndarray.__path__._path[0], 'cuda_ndarray.pyd')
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/type.py
```diff
@@ -21,6 +21,11 @@
     import cuda_ndarray.cuda_ndarray as cuda
     from theano.sandbox.cuda.nvcc_compiler import NVCC_compiler
     import cuda_ndarray
+    # Python 3 does not necessarily set __file__. May need manual setting
+    try:
+        cuda_ndarray.__file__
+    except AttributeError:
+        cuda_ndarray.__file__ = os.path.join(cuda_ndarray.__path__._path[0], 'cuda_ndarray.pyd')
```
nouiz: cuda_ndarray.__path__._path don't exist in Python 2. But as this is in the except clause, it is fine.

Can you add in your comment that this happen in python 3 under windows and tell which python version/installation package you used?

This would explain people that would see this code why we use the pyd extension that is only for Windows. Also, I'm currious which stack on windows have this properties, as it is working in the case we tested recently.

As I didn't saw this problem on Linux, it is probably not useful to test if the file exist.
thanks

cai-lw: Comment added.
I don't know whether the problem will occur on other platforms. You reminded me that using hardcoded `.pyd` extension is wrong, and I have changed it into `get_lib_extension()` which can determine the correct extension on different platforms.

nouiz:
@abergeron @lamblin can we merge this now or we should wait after the release? This code isn't called by default, only when we are in an environment that brake right now.

Issue #3551, opened 2015-10-24
Please help! ERROR (theano.sandbox.cuda): Failed to compile cuda_ndarray.cu: 'utf-8' codec can't decode....


liyi193328:
My environment is win10 + python3.4+ theano 0.7+cuda7.5. when I import theano from teriminal , I get the error like:
### ERROR (theano.sandbox.cuda): Failed to compile cuda_ndarray.cu: 'utf-8' codec can't decode......

I set log level as debug, get the folloing info : 
INFO (theano.gof.cmodule): Call to 'g++ -march=native' failed,not setting -march flag
DEBUG (theano.gof.cmodule): Compiling for 64 bit architecture
DEBUG (theano.gof.cmodule): Writing module C++ code to C:\Users\liyi1\AppData\Local\Theano\compiledir_Windows-8-6.2.9200-Intel64_Family_6_Model_60_Stepping_3_GenuineIntel-3.4.3-64\cutils_ext\mod.cpp
DEBUG (theano.gof.cmodule): Generating shared lib C:\Users\liyi1\AppData\Local\Theano\compiledir_Windows-8-6.2.9200-Intel64_Family_6_Model_60_Stepping_3_GenuineIntel-3.4.3-64\cutils_ext\cutils_ext.pyd
DEBUG (theano.gof.cmodule): Running cmd: C:\Program Files\mingw-w64\x86_64-5.2.0-posix-seh-rt_v4-rev0\mingw64\bin\g++.exe -shared -g -D NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -m64 -DMS_WIN64 -ID:\Anaconda3\lib\site-packages\numpy\core\include -ID:\Anaconda3\include -o C:\Users\liyi1\AppData\Local\Theano\compiledir_Windows-8-6.2.9200-Intel64_Family_6_Model_60_Stepping_3_GenuineIntel-3.4.3-64\cutils_ext\cutils_ext.pyd C:\Users\liyi1\AppData\Local\Theano\compiledir_Windows-8-6.2.9200-Intel64_Family_6_Model_60_Stepping_3_GenuineIntel-3.4.3-64\cutils_ext\mod.cpp -LD:\Anaconda3\libs -LD:\Anaconda3 -lpython34
[...]

PR #4235, opened 2016-03-15
Decoding nvcc outputs in correct encoding.

cai-lw:
`nvcc` calls `cl` (Visula Studio C/C++ compiler), and `cl` outputs are in "current ANSI console codepage" (see [Visual Studio Documentation](https://msdn.microsoft.com/en-us/library/xwy0e8f2%28v=vs.120%29.aspx)). If the encoding is not UTF8 and the output contains non-ASCII characters (very likely to happen if the VS installation is in Asian languages), the default `decode()` will result in an exception.
There had been an issue #3551 that reported the problem, but remained untreated for long.
This pull request fixed the problem by adding a function that can decode in specified encoding and compatible with both Python 2 and 3, and detecting encodings of stdout and stderr.

### DIFF
--- a/theano/compat/__init__.py
+++ b/theano/compat/__init__.py
@@ -44,6 +44,9 @@ if PY3:
     def decode_iter(itr):
         for x in itr:
             yield x.decode()
+
+    def decode_with(x, encoding):
+        return x.decode(encoding)
 else:
     from six import get_unbound_function
     from operator import div as operator_div
@@ -63,6 +66,9 @@ else:
     def decode_iter(x):
         return x
 
+    def decode_with(x, encoding):
+        return x
+        
 __all__ += ['cmp', 'operator_div', 'DictMixin', 'OrderedDict', 'decode',
             'decode_iter', 'get_unbound_function', 'imap', 'izip', 'ifilter']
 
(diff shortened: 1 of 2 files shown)


nouiz:
Thanks for the fix. I like the code, it is clean.

We try to make a release very close. So I'm not 100% sure we can merge this without too much chance to break other stuff. I'm ok to merge this, but I would like other Theano dev to agree to merge before the release. Otherwise, it should be in a few days.

cai-lw:
@nouiz The first commit is not working because `p.stdout` and `p.stderr` do not derive from `io.TextIOBase` thus don't have the `encoding` attribute. I am using `locale.getpreferredencoding()` to detect encoding now. It should work in most cases, if not all.

nouiz:
Travis have this error:

/home/travis/build/Theano/Theano/theano/compat/__init__.py:71:1: W293 blank line contains whitespace

I'm don't think we will merge this before the release. Just to be sure we don't break windows/mac. But otherwise, I think it is good. @lamblin @abergeron

nouiz:
Seem good. Just for safety, we will merge after the release that is planned Monday.

nouiz:
thanks

lamblin:
Travis now passes, and the release is over, so I'm merging.
Thanks!

maomao2991:
Thank you very much! ! It really halpful, solving the issue had troubled me for one night.

PR #4237, opened 2016-03-15
Implementation of Batch Normalization from cuDNN v.4

avostryakov:
This pull request adds one base class (GpuDnnBatchNormBase) and three classes (GpuDnnBatchNormInference, GpuDnnBatchNorm, GpuDnnBatchNormGrad). Each of them calls cuDNN API function.

GpuDnnBatchNormInference and GpuDnnBatchNormGrad are compiled and work, but I stuck with GpuDnnBatchNorm because in this function two variables are input and output variables at the same time. I tried to use view_map or destroy_map attributes but couldn’t make it works. Now GpuDnnBatchNorm is compiled, but when I attempted to execute it, I see Segmentation Fault. 

So, I need your help with it.

I used the following script to test new functions:
```
import numpy as np

import theano
import theano.tensor as T
from theano.sandbox.cuda.basic_ops import gpu_contiguous

from theano.sandbox.cuda.dnn import GpuDnnBatchNormInference, GpuDnnBatchNorm, GpuDnnBatchNormGrad


def batch_norm():
    x = np.arange(3*5, dtype=np.float32).reshape((3, 5))
    bn_scale = np.ones((1, 5), dtype=np.float32)*2
    bn_scale[0][4] = 3

    bn_bias = np.ones((1, 5), dtype=np.float32)
    running_mean = np.zeros((1, 5), dtype=np.float32)
    running_inv_variance = np.ones((1, 5), dtype=np.float32)*4

    x_var = T.fmatrix()
    ins0 = x_var.dimshuffle(0, 1, 'x', 'x')
    ins0 = gpu_contiguous(ins0)

    scale_var = T.fmatrix()
    ins1 = scale_var.dimshuffle(0, 1, 'x', 'x')
    ins1 = gpu_contiguous(ins1)

    bias_var = T.fmatrix()
    ins2 = bias_var.dimshuffle(0, 1, 'x', 'x')
    ins2 = gpu_contiguous(ins2)

    mean_var = T.fmatrix()
    ins3 = mean_var.dimshuffle(0, 1, 'x', 'x')
    ins3 = gpu_contiguous(ins3)

    inv_variance_var = T.fmatrix()
    ins4 = inv_variance_var.dimshuffle(0, 1, 'x', 'x')
    ins4 = gpu_contiguous(ins4)

    y, current_mean_var, current_variance_var, running_mean_var, running_variance_var = \
        GpuDnnBatchNorm('bc01', 'per-activation')(ins0, ins1, ins2, ins3, ins4)

    f = theano.function([x_var, scale_var, bias_var, mean_var, inv_variance_var], [y, current_mean_var,
                                                                                   current_variance_var,
                                                                                   running_mean_var,
                                                                                   running_variance_var],
                        mode='DebugMode')

    output = f(x, bn_scale, bn_bias, running_mean, running_inv_variance)

    print x
    print running_mean
    print running_inv_variance
    print np.asarray(output[0])
    print np.asarray(output[1])
    print np.asarray(output[2])

if __name__ == '__main__':
    batch_norm()
```


nouiz:
thanks for the PR. I'm checking it now. You can add the test function in cuda/tests/test_dnn.py directly.

theano/sandbox/cuda/dnn.py
```diff
@@ -2186,6 +2186,460 @@ def method(self):
...
+    ----------
+    tensor_format
+        Always set this to 'bc01'.
+    mode
+        Normalization is performed per-activation for 'per-activation' (in this mode bnBias and bnScale tensor
+        dimensions are 1xCxHxW for 4d tensor) or Normalization is performed over N+spatial dimensions for 'spatial'
+        (in this mode bnBias, bnScale tensor dimensions are 1xCx1x1).
+    exp_avg_factor
+        Factor used in the moving average computation runningMean = newMean*factor + runningMean*(1-factor). It is
+        required that exponentialAverageFactor=1 is used for the very first call of a complete training cycle.
```
nouiz: This description indicate to me that the value should change between epoch. As we don't want to recompile Theano function between epoch to change this value, I think it should be an input of the node (in the make_node).

theano/sandbox/cuda/dnn.py
```diff
@@ -2186,6 +2186,460 @@ def method(self):
...
+    __props__ = ('tensor_format', 'mode', 'exp_avg_factor', 'epsilon')
+
+    def __init__(self, tensor_format, mode, exp_avg_factor=0.1, epsilon=1e-4):
+        if tensor_format != 'bc01':
+            raise ValueError(
+                "It was discovered that since December 2014, the "
+                "tensor_format parameter was ignored and the equivalent of "
+                "'bc01' is always used.  Since your code seems to be using "
+                "another value, this might have affected previous results "
+                "ran with this code.")
```
nouiz: who ignore it? Theano or CuDNN? Theano can't have ignore it as we didn't had it before. Just specify it in the message that it is CuDNN that ignored it and tell which version.

nouiz: If you aren't sure, just tell that Theano only support that case.

theano/sandbox/cuda/dnn.py
```diff
@@ -2186,6 +2186,460 @@ def method(self):
...
+    epsilon
+        Epsilon value used in the batch normalization formula. It is 1e-4 by default. Minimum allowed value
+        for cuDNN v.4 is 1e-5.
+
+    """
+
+    direction = 'forward'
+    bn_descs = ['bn_input', 'bn_output', 'scale_bias_mean']
+    view_map = {3: [3], 4: [4]}
+    #destroy_map = {3: [3], 4: [4]}
```
nouiz: As the running* inputs value change after this call, it should be destroy_map and now view_map.

theano/sandbox/cuda/dnn.py
```diff
@@ -2186,6 +2186,460 @@ def method(self):
...
+        # Validate the input and build the input variables.
+        for input_idx, bn_input in enumerate(inputs):
+            if self.direction == 'backward':
+                if input_idx == 0:
+                    input_name = 'bn_input'
+                elif input_idx == 1:
+                    input_name = 'bn_gout'
+                else:
+                    input_name = 'scale_bias_mean'
+            else:
```
nouiz: The else here seem to be for inference and seem incomplete. You seem to set the inputs at index 1-4 to the same DNN tensor. But they should have different tensor.

You also need an elif self.direction('forward').

theano/sandbox/cuda/dnn.py
```diff
@@ -2186,6 +2186,460 @@ def method(self):
...
+    }
+    """
+        result += sub_result % subs
+
+        for output_idx, output in enumerate(outputs):
+            if self.direction == 'backward':
+                if output_idx == 0:
+                    output_name = 'bn_ginput'
+                else:
+                    output_name = 'scale_bias_mean'
```
nouiz: There is more then 2 outputs for the grad, ans all index 1-N seem to be sharing the same dnn descriptor here.

theano/sandbox/cuda/dnn.py
nouiz: I think I understand now this loop over input and the second loop over output. Both of them are to create the cudnn tensor descriptors. But as Theano have more inputs + outputs then the number of descriptor, this is probably the cause of the problems you have.

I think you should remove both of those loop and make one over self.bn_desc. I would change self.bn_decs to be something like
bn_descs = [('name', input_index), ...]
this way, you know the good mapping between which input need a descriptor and only init those one.

As all descriptor are on tensor that are in the inputs of the node, you don't need to create some for the output of this node. This is because this node don't allocate output memory, but write on some of its inputs.
I think you should have

nouiz: My last paragraph isn't right with the current version of your code. Currently, allocate all outputs of the node. The other cudnn op do not do that. In the case of batch-norm, it is probably fine do allocate the output in the node. But not all of the nodes need to allocate all the outputs. Some of them are modified views of the inputs. So you will need to only allocate those that are not views of inputs and set the output ptr to the input for view (and call Py_INCREF() on it for the good ref counting).

nouiz:
I think my comments about the handling of the inputs and outputs should help you. If I wasn't clear, tell me.

theano/sandbox/cuda/dnn.py
f0k: It's something that you can change between every batch, to turn the exponential averaging into an overall average. If we want to support that, it needs to be an input, yes.

The comment about setting it to 1 for the first call (not for the first epoch!) is generally valid, but I don't think it's relevant in practice. Usually you'd just initialize the runningMean to zero and the runningInvStd to one and then use the same factor throughout training. The running averages will only be used for validation, after a full epoch of training, so the initialization will have been mostly "forgotten" by then (e.g., only included with a factor of 0.9^num_training_batches).

theano/sandbox/cuda/dnn.py
f0k: I think this is copy/pasted from the softmax op. Shouldn't be included here.

nouiz:
@f0k do you think it is useful to have an inplace version of the grad? To my understanding, it don't update parameter, so it won't speed up computation, but I don't have a very good understanding of that.

This would be an extra optional optimization. Only after a first working version.

f0k:
> @f0k do you think it is useful to have an inplace version of the grad? To my understanding, it don't update parameter [...]

It updates the scale and bias parameters, and supports the usual alpha/beta parameters for writing to the destination tensors. So there could be an in-place version, right?

avostryakov:
Thank you, guys! I will improve the code in several days.

nouiz:
For the grad I see the utility of the inplace version. I see that the
forward have that parameter. I'm inquiring if you know a use case where we
would use a beta != 0 in the forward direction.

f0k:
> I'm inquiring if you know a use case where we would use a beta != 0 in the forward direction.

Well, the original ResNets of He et al. (http://arxiv.org/abs/1512.03385) could use it in the forward pass. The new ResNets from Wednesday (http://arxiv.org/abs/1603.05027) not.
ResNets variants

nouiz:
thanks.

avostryakov:
@nouiz Hi!

I've almost finished improving code but I don't understand the following paragraph:

"So you will need to only allocate those that are not views of inputs and set the output ptr to the input for view (and call Py_INCREF() on it for the good ref counting)."

It's situation for GpuDnnBatchNorm. I hoped that using destroy_map will be enough. Can you give me an example how to do it or explain in more details?

nouiz:
The doc for destructive operation is here:

http://deeplearning.net/software/theano/extending/inplace.html

But it don,t include c code example. Here is quick code example:

        z_in, a, x, y, b = inputs
        z_out, = outputs
        inplace = int(self.inplace)
        fail = sub['fail']

return """
...
Py_DECREF(%(z_out)s);
Py_INCREF(%(z_in)s);
%(z_out)s = %(z_in)s
...

""" % locals()

This will take care of setting the output at the right place and take care
of python ref count.

avostryakov:
I fixed all remarks and classes are compiled and executed without errors but I'm not sure that all works correctly. I'll check it later. Anyway here is list of my changes with questions:
1. exp_avg_factor was moved to GpuDnnBatchNorm.make_node. Now the make_node method has optional parameter exp_avg_factor = 0.1 by default. I converted this parameter in theano scalar variable with the following code: exp_avg_factor = as_scalar(exp_avg_factor). In 'method' I pass this parameter in cudnnBatchNormalizationForwardTraining function by name %(ins5)s. exp_avg_factor in cudnnBatchNormalizationForwardTraining should have a double type. The question is did I do it correctly?

2. tensor_format was removed.

3. I created destroy_map = {1: [3], 2: [4]} property in class GpuDnnBatchNorm. After that I had to add 'accept_inplace=True' in theano.function to compile without errors in debug mode. Otherwise I see the following error:
'Graph must not contain inplace operations'. Why I have to add accept_inplace=True?

4. I changed a logic with descriptors creation. I hope it works correctly now.

Please, check a new version of the code.

avostryakov:
I checked a work of GpuDnnBatchNorm and GpuDnnBatchNormInference ops. All works correctly including parameter exp_avg_factor. Something strange with GpuDnnBatchNormGrad ops. Every time it returns different results for dx, resultBnScaleDiff and resultBnBiasDiff. I could find a bug yet.

avostryakov:
Two more questions: 

1. Can I remove the following code with 'tensor_format' because I think it doesn't influence on cuDNN functions that I use:

cudnnTensorFormat_t format%(name)s = CUDNN_TENSOR_NCHW;
if (%(tensor_format)d == 1)
  format%(name)s = CUDNN_TENSOR_NHWC;

2. Should I execute the following code for output variables that is mapped to input variables:

   if (CudaNdarray_prep_output(&%(outsX)s, 4, CudaNdarray_HOST_DIMS(%(insX)s)) != 0)
    {
      %(fail)s
    }

theano/sandbox/cuda/dnn.py
```diff
@@ -2184,8 +2184,406 @@ def method(self):
...
+        (in this mode bnBias, bnScale tensor dimensions are 1xCx1x1).
+    epsilon
+        Epsilon value used in the batch normalization formula. It is 1e-4 by default. Minimum allowed value
+        for cuDNN v.4 is 1e-5.
+
+    """
+
+    direction = 'forward'
+    input_descs = [('bn_input', 0), ('scale_bias_mean', 1)]
+    destroy_map = {1: [3], 2: [4]}
```
nouiz: Normally, we don't do that. This is because if the user start to decide which op are inplace, it could create cycle in the computation graph without knowing. So this is better left to the optimizer that will do it and make sure to not introduce cycle. This is why by default, we don't accept inplace op in the graph build by the user.

The way we do it is by having and \__init__ method to the op. It have an inplace=False parameter and when the user build the graph, it is not inplace. Then an optimizer check if there is op like this one that aren't inplae and suggest to change it to an inplace version. Then Theano verification will reject the change if it introduce an cycle.

Here is a quick draft of the optimization needed:
```
@local_optimizer([GpuDnnBatchNorm], inplace=True)
def local_inplace_dnn_barch_norm(node):
    if isinstance(node.op, GpuDnnBatchNorm):
        return [GpuDnnBatchNorm(other_param..., inplace=True)(*node.inputs)]
```

theano/sandbox/cuda/dnn.py
```diff
@@ -2184,8 +2184,406 @@ def method(self):
...
+
+    def make_node(self, x, bn_scale, bn_bias, running_mean, running_variance, exp_avg_factor=0.1):
+        x = as_cuda_ndarray_variable(x)
+        bn_scale = as_cuda_ndarray_variable(bn_scale)
+        bn_bias = as_cuda_ndarray_variable(bn_bias)
+        running_mean = as_cuda_ndarray_variable(running_mean)
+        running_variance = as_cuda_ndarray_variable(running_variance)
+        exp_avg_factor = as_scalar(exp_avg_factor)
```
nouiz: In other op for cudnn, we use somethink like: "ensure_float(exp_avg_factor, _zero_point_one, exp_avg_factor)"

where  _zero_point_one =  constant(numpy.asarray(0.1, dtype='float32'))

use 


 def make_node(self, x, bn_scale, bn_bias, running_mean, running_variance, exp_avg_factor=None):

nouiz:
I can't do a full review now, but here is some quicks comment.

I think cudnnTensorFormat isn't needed anymore.

If output is allocated by the op, then yes, you should call *prep_output. But for the op, when it is inplace, you don't call it.

theano/sandbox/cuda/dnn.py
f0k: > Normally, we don't do that. This is because if the user start to decide which op are inplace, it could create cycle in the computation graph without knowing. So this is better left to the optimizer that will do it and make sure to not introduce cycle.

This would require the batch norm Op to be callable like:
```python
output, new_statistics = GpuDnnBatchNorm(inplace=False)(input, old_statistics)
```
However, the implementation in cuDNN does not support updating the statistics (running averages of mean and invstddev) out-of-place at all. It is always destructive.

What would you suggest then? It would be possible to create a non-destructive version by actually copying the inputs before calling the cuDNN implementation, is this the way to go?

Furthermore, if there is a single inplace parameter, how do you know if it applies to the batchnorm statistics or the actual output? Both could be inplace or out-of-place independently.

nouiz: Also, the inplace parameter influence the c code. For the not inplace case, we should copy the input that are being worked inplace by cudnn instead of calling *prep_output()

nouiz: We can simplify and support only being inplace on both or none. If you want to support all the combination, then you will need 2 inplace_N parameter.

f0k: > We can simplify and support only being inplace on both or none. If you want to support all the combination, then you will need 2 inplace_N parameter.

The statistics will almost always be updateable in-place. The input not (e.g., in residual networks, the unmodified input is still needed for later residual units). So we would need separate inplace arguments here, wouldn't we? (Disclaimer: I'm not sure if Theano performs inplace optimizations the way I think.)

nouiz: If there is use cases where the node can be inplace only on one of the
input, it make sence to make 2 inplace param and 2 optimization (each one
will enable only one of the inplace paramter)

avostryakov:
@nouiz About your last comment:
"If output is allocated by the op, then yes, you should call *prep_output. But for the op, when it is in-place, you don't call it."

Actually, I don't understand if all outputs are allocated or not :) If allocation has happened here:

return Apply(self, [x, bn_scale, bn_bias, running_mean, running_variance, exp_avg_factor],
                     [x.type(), running_mean.type(), running_variance.type(), current_mean, current_variance])

Then I allocate all outputs including mapped outputs. Should I change the code above the following way?

return Apply(self, [x, bn_scale, bn_bias, running_mean, running_variance, exp_avg_factor],
                     [x.type(), running_mean, running_variance, current_mean, current_variance])

P.S. I'm sorry but our server with GPU was broken. We'll repair it now. So, will be a pause with finishing this PR from my side. I hope not more than one-two days.

f0k:
I'm not the expert, but I think you should make the Op accept two `inplace` parameters in its constructor (one for the normalized data, one for the statistics) and then either allocate the outputs and have the destroy map be empty, or not allocate the outputs and have a destroy map.

avostryakov:
@f0k I read all comments twice and I'm lost now :) Ok. First of all, we are talking about two variables in GpuDnnBatchNorm Op: running_mean, running_variance. It's a single case where we have in-place variables. At second, in cuDNN we have two different functions for inference and for forwarding pass during training: GpuDnnBatchNormInference (for prediction) and GpuDnnBatchNorm (for training). I hope we are talking about training: GpuDnnBatchNorm. At third, in GpuDnnBatchNorm running_mean, running_variance are updated statistics and there are current_mean, current_variance which is calculated for a current mini-batch only.

  My best understanding that you suggest to not output input variables running_mean, running_variance from the Op. Now they are outputted from the Op.
  
  Now I don't understand how it works. It looks like I'm really allocating memory for running_mean, running_variance which will be outputted and copy input "running_mean, running_variance" to output "running_mean, running_variance" in C code. In this case, I don't understand why I use 
destroy_map = {1: [3], 2: [4]}

f0k:
> First of all, we are talking about two variables in GpuDnnBatchNorm Op: running_mean, running_variance. It's a single case where we have in-place variables. At second, in cuDNN we have two different functions for inference and for forwarding pass during training: GpuDnnBatchNormInference (for prediction) and GpuDnnBatchNorm (for training). I hope we are talking about training: GpuDnnBatchNorm.

Yes, I was talking about training as well (for now).

> At third, in GpuDnnBatchNorm running_mean, running_variance are updated statistics and there are current_mean, current_variance which is calculated for a current mini-batch only.

Ah yes, they're needed for the backward pass to avoid recomputing them.

> My best understanding that you suggest to not output input variables running_mean, running_variance from the Op. Now they are outputted from the Op.

No, they always need to be outputted from the Op, in the sense that the Op needs to return them. The Op would take running_* as input variables and then return new_running_* as output variables, so to say. No matter whether the Op operates in-place or out-of-place, a symbolic variable is immutable, so the Op has to return a new variable in any case. Check the Theano documentation on inplace operations for that: http://deeplearning.net/software/theano/extending/inplace.html

> Now I don't understand how it works. It looks like I'm really allocating memory for running_mean, running_variance which will be outputted and copy input "running_mean, running_variance" to output "running_mean, running_variance" in C code.

Let's call the output new_running_mean, new_running_variance to distinguish the variables before and after the update. If the Op is not operating in-place, you would indeed do the following: allocate new_running_mean and new_running_variance inside the Op, copy the running_mean and running_variance into it (in C code), then call the cuDNN function. You would *not* set the destroy map in this case, as the Op does not destroy/overwrite any of its inputs then.

Now let's consider the case for in-place operation for the statistics (running mean and variance; current mean/variance is out of question). In this case, the Op would still return new_running_mean, new_running_variance, but: The implementation would not allocate new_running_mean and new_running_variance, but make them identical to running_mean and running_variance (using @nouiz's [incref/decref/assign snippet](https://github.com/Theano/Theano/pull/4237#issuecomment-203447859)) and call the cuDNN function on them. To make Theano aware which inputs are destroyed when executing the Op, you need to set the destroy map accordingly.

As @nouiz said, you would code a single Op to be usable both for in-place and out-of-place operation. It would take an `inplace` boolean flag in its constructor (`__init__`) and change behaviour accordingly. Along with it, there would be an optimizer replacing the out-of-place Op with the inplace version, as here: https://github.com/Theano/Theano/pull/4237#discussion_r58255706. It takes care of checking whether the replacement is valid.

All right, now as a further complication, there are *two* sets of inputs in the BatchNorm training forward Op that could work in-place: 1. running mean/var, 2. input. In a standard feed-forward net (not in the new style of residual nets), the unnormalized input is not needed any more after propagating it through a batch normalization layer, so we can overwrite it in place. This means the Op would need *two* `inplace` flags in its constructor, allow inplace operation for mean/var and input independently, and have *two* optimizers.

Finally, it's often useful to allow inplace operation for the gradient as well, including the alpha/beta parameters, and we could also have an inplace version for the inference forward pass.

Sorry for the long post, hope this makes things clear! But: For this PR, it would be enough to have a first version of the three Ops that does not support any inplace operation whatsoever. This is something which can be added later without changing any user code. So I'd suggest to first try getting everything to work without any use of destroy maps.

avostryakov:
@f0k For now, can I ask just one question: Why do we want to have out-of-place Op forGpuDnnBatchNorm? From your explanation above I have in-place Op now because I use @nouiz's incref/decref/assign snippet and set the destroy map accordingly.

avostryakov:
I'm asking because for out-of-place operation we have to have additional copy operation.

nouiz:
The problem is that the user can't always know the best pattern for inplace
operation. So we ask the the graph is build without inplace operation and
it is Theano optimization that swap the out-of-place operation with inplace
operation when it can. For example, if an output of a node is used by 2
other node, only one of them can be inplace on it. There is also case where
this would introduce cycle in the graph. The user can't easily know about
this.

On Mon, Apr 4, 2016 at 5:42 AM, Anatoly Vostryakov <notifications@github.com

f0k:
> For now, can I ask just one question: Why do we want to have out-of-place Op for GpuDnnBatchNorm?

Fréd already answered, and there's more detail here: http://deeplearning.net/software/theano/extending/inplace.html

In practice, providing *only* the inplace version (for the running statistics!) is not going to do any harm here. The only way it can fail is if you do something like:
```
new_run_m, new_run_v = BatchNormOp()(input1, run_m, run_v, ...)
new_run_m2, new_run_v2 = BatchNormOp()(input2, run_m, run_v, ...)
```
Now both Ops would attempt to do an inplace operation on run_m/run_v, and this can't work. One of them would need to work out-of-place.
This will not happen in practice because nobody will attempt to use it that way, but for other Ops it's less obvious and so the general idea of "default to out-of-place, optimize to inplace if possible" makes sense to me.

> From your explanation above I have in-place Op now because I use @nouiz's incref/decref/assign snippet and set the destroy map accordingly.

If you have the in-place Op working already, how difficult is it to add the out-of-place version (with the additional copy operation) and the optimizer? The only thing you should take care of is to call the `inplace` constructor argument something like `inplace_running_stats` so it's possible to add an `inplace_data` or similarly in a follow-up PR.

f0k:
Relevant: cuDNN v5 RC is available now: https://developer.nvidia.com/cudnn
It changes the interface to compute runningVariance instead of runningInvVariance.
This means it's difficult to support v4 and v5 with the same interface. Maybe the BatchNormOp() should be coded against v5 from the start and refuse to work with v4?

avostryakov:
Damn it! if it is so, I'll change the code for v.5. Sorry for delay, don't have time to finish this PR.

lachlants:
Here's a patch for the current master for using this code with a few bug fixes. I removed the global mean and std updating since this is easily handled by the user and removes the inplace issues mention above. 

https://gist.github.com/lachlants/642fe4f88f7543b1cde71f804443c630#file-fix_cudnn_batchnorm-patch

f0k:
> I removed the global mean and std updating since this is easily handled by the user and removes the inplace issues mention above.

Ah, nice idea! That's possibly not even going to slow us down. I might have a stab at turning your patch into a PR for cuDNN v5.

nouiz:
Continued in https://github.com/Theano/Theano/pull/4582, so closing. thanks.

PR #4241, opened 2016-03-15
Fix the bilinear tests in float32.

abergeron:


nouiz:
ok to merge when travis pass.

PR #4246, opened 2016-03-16
add R_op for DisconnectedGrad

poolio:
Add missing R_op for DisconnectedGrad.


theano/gradient.py
```diff
@@ -1981,6 +1981,9 @@ class DisconnectedGrad(ViewOp):
     def grad(self, args, g_outs):
         return [disconnected_type() for g_out in g_outs]
 
+    def R_op(self, inputs, eval_points):
```
nouiz: @lamblin I let you review this.

lamblin:
Since R_op does not have the distinction between disconnected/zero/undefined, that's OK for me.

PR #4248, opened 2016-03-16
Added support for keepdims in norm function

Azrael1:


theano/tensor/var.py
```diff
@@ -592,15 +592,19 @@ def prod(self, axis=None, dtype=None, keepdims=False, acc_dtype=None):
...
 
-    def norm(self, L, axis=None):
+    def norm(self, L, axis=None, keepdims=False):
         if L == 0:
             raise NotImplementedError()
         if numpy.isinf(L):
             raise NotImplementedError()
         # optimizations will/should catch cases like L=1, L=2
-        return theano.tensor.basic.pow(
+        y = theano.tensor.basic.pow(
```
nouiz: The implementation seem good, but this need tests. They should be un tensor/tests/test_var.py probably. You can grep in the test folder about keepdims to see how it is tested for the other functions.

nouiz:
finished in https://github.com/Theano/Theano/pull/4418

PR #4249, opened 2016-03-16
Fix windows test failures

lamblin:
Hopefully these are the last ones.

### DIFF
--- a/theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
+++ b/theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
@@ -845,10 +845,15 @@ def conv_grad(mode, bs, ch, nf, rImg1, rImg2, rFlt1, rFlt2, subsample, op):
     outputs.extend([corr_op_dik, conv_op_dik,
                     corr_op_dki, conv_op_dki])
 
-    # TODO: fix when the abstractconv tests can pass debug mode.
-    mode = theano_mode
-    if theano.config.mode == 'DEBUG_MODE':
-        mode = theano.compile.mode.get_mode('FAST_RUN').including('gpu')
+    if not theano.config.blas.ldflags:
+        # Some of the operations are not transferred to the GPU,
+        # and withoug BLAS, the abstract Op will not be optimized
+        # to CorrMM either, so we have to accept the use of the
+        # slow Python convolution in that case.
+        mode = theano_mode.excluding('AbstractConvCheck')
+    else:
+        mode = theano_mode
+
     f = theano.function([i, k], outputs, mode=mode)
 
     allvals = f(npy_img, npy_kern)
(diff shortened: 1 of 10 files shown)


theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
```diff
@@ -845,10 +845,15 @@ def conv_grad(mode, bs, ch, nf, rImg1, rImg2, rFlt1, rFlt2, subsample, op):
     outputs.extend([corr_op_dik, conv_op_dik,
                     corr_op_dki, conv_op_dki])
 
-    # TODO: fix when the abstractconv tests can pass debug mode.
-    mode = theano_mode
-    if theano.config.mode == 'DEBUG_MODE':
```
nouiz: Did you time in debug mode? The buildbot is already too long. Otherwise, it is ok.

theano/scalar/basic.py
```diff
@@ -1151,7 +1151,13 @@ def c_code(self, node, name, inputs, outputs, sub):
         (z,) = outputs
         if node.inputs[0].type in complex_types:
             raise NotImplementedError()
-        return "%(z)s = isnan(%(x)s);" % locals()
+        # Windows tries to be different and sometimes return -1, but we want
+        # to be consistent with numpy (which returns True), hence the "abs".
+        return "%(z)s = abs(isnan(%(x)s));" % locals()
```
nouiz: this would cause slow down on mac and linux. Can you add the abs only on windows?

theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
lamblin: I'll time it.

theano/scalar/basic.py
lamblin: I really think the slowdown would be negligible since we are already applying some operation on `x`, and that is already the fix we have for `is_inf`.
Can that wait for after the release?

nouiz: yes

theano/sandbox/cuda/tests/test_conv_cuda_ndarray.py
abergeron: Do we have to delay the release for that?  I'd rather we change it after if it's a problem.

lamblin: Without CuDNN, the test takes:
- 200 s in default mode
- 2000 s in DEBUG_MODE

This is probably the same time it was taking before we introduced the abstract convolution, but I'm OK with not putting back debug mode before the release.

nouiz: ok for that timing for now.

PR #4251, opened 2016-03-17
(nit) tutorial typos

sygi:
This uses the same branch I already merged to master before, do you have a policy against that?


nouiz:
I killed travis as the doc part passed.

nouiz:
thanks

PR #4252, opened 2016-03-17
Add the same restriction to the perform to return the same error and …

nouiz:
…have test don't use more then 9G of RAM.


abergeron:
Ok when travis passes

PR #4253, opened 2016-03-17
Fix tests in DebugMode with DNN convolutions

lamblin:


nouiz:
ok to merge when travis pass.

lamblin:
I'll add the fix & test for the pooling descriptor here as well.

lamblin:
Apparently, we are not using the pooling descriptor at all any more in the new back-end, and trying to force the old interface results in errors in shape inference. So I'll fix it but not include a test for that, if it's OK.
I'll check the new back-end as well.

nouiz:
Don't check the new interface. Harm is removing the descriptor.

nouiz:
I merge this PR as this fix stuff for the release.

PR #4258, opened 2016-03-18
Fixes for debugmode in dnn and gpuarray

lamblin:


theano/sandbox/gpuarray/basic_ops.py
```diff
@@ -996,6 +996,8 @@ def get_params(self, node):
     def perform(self, node, axis_and_tensors, out_, ctx):
         out, = out_
         axis = int(axis_and_tensors[0])
+        if axis < 0:
+            axis += axis_and_tensors[1].ndim
```
abergeron: I'm pretty sure that this is done inside of concatenate.  It doesn't hurt to have it here though.

abergeron: Or actually it does if `axis < -ndim`.  This will cover up the error.

nouiz:
seem ok to me.

nouiz:
Just to be clear, I'm ok to merge after travis pass. @abergeron is this the same for you?

theano/sandbox/gpuarray/basic_ops.py
lamblin: I don't see the check you are talking about, and anyway, it fails because of an OverflowError even before that, because `axis` is declared as an `unsigned int`.
The existing tests check that the error is not covered up.

Issue #2472, opened 2015-02-03
Partial evaluation of a Theano function


nouiz:
This would allow to reuse the same function for test and train. With the cvm/vm linker, this is relatively easy if we compile the full function, then just ask to compute some of the outputs.

We could also allow to compute just some of the intermediate results. This would allow to unroll a big scan with the max supported lenght and compute just the number of iteration we want. Not sure how this would work for the back-ward.

nouiz:
For people wanting to know how to implement this:

- Understand how Theano graph are done. (bi-partite graph, with 2 type of node Variable and Apply node, you must understand how to navigate between each node)
- Start by the vm (python version). When this is done there, this can be ported to the cvm (c code). It is easier to work in python to develop the algorithm.
- start by understanding how the vm/cvm evaluate a Theano function. The VM is in the class VM, in the file theano/gof/vm.py. It start from the output, for each node, it check if its dependency are available, if not it try to evaluate this. This is done recursively, but with a loop implementation instead of calling recursively different function.
-Then understand how the compute_map and storage_map work. In particular, understand how the IfElse op work, as this will cause only some of the inputs to be evaluated.
- With that information, you should be able to modify the VM to start from a subset of the output nodes. Implement a user way to do that. This would be the first user available feature.

sygi:
Could I ask for (even more) clarification on that? I mean, as partial evaluation you mean:
a) putting just a subset of inputs, calculate some subset of nodes and then, if you give the rest of the inputs multiple times, you don't recompute the nodes that were saved previously OR
b) give all the inputs, but calculate only some subset of the output nodes. This seems as more expected interpretation, as I see "output nodes" in the description, but how is that different from creating separate functions for different output nodes? Isn't it the case that you have to calculate the whole function exactly once for a given input?
In particular, could you elaborate more on the "not recompiling train/test function"?
Thanks for the hints, sorry if these questions are silly.

nouiz:
Don't hesitate to ask questions. It is only by well understanding the
problem that we can find a good solution:)

It is more b). Maybe you don' t need to give all inputs as some are maybe
not needed to compute the subset of output we need to compute.

This is similar to compiling many function, but would save on the
compilation time and memory consumption (not so much a problem). As the
goal is to allow many different combination of outputs and same compilation
time, we don't want to retrain a function each time.

sygi:
I dived into the code and have few more questions about how it works:
[...]

PR #4259, opened 2016-03-18
WIP: quick-&-dirty solution to partial function evaluation

sygi:
I implemented a very simple and limited solution to the #2472 to ask you if I'm going in good direction at all.

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -375,6 +375,7 @@ class Function(object):
         self.name = None
         self.nodes_with_inner_function = []
         self.output_keys = output_keys
+        self.output_idx = None
 
         # We will be popping stuff off this `containers` object.  It is a copy.
         containers = list(self.input_storage)
@@ -751,7 +752,13 @@ class Function(object):
         f_cpy.maker.fgraph.name = name
         return f_cpy
 
-    def __call__(self, *args, **kwargs):
+    def apply(self, inputs, outputs):
+        # TODO: assert that there's a list of outputs
+        self.output_idx = outputs # indirectly passing output indices
+        # assuming for now I pass all the inputs
+        return [self.__call__(*inputs)[i] for i in outputs]
+
+    def __call__(self, *args, **kwargs): # sygi: what those kwargs mean?
         profile = self.profile
         t0 = time.time()
 
@@ -856,7 +863,7 @@ class Function(object):
         # Do the actual work
         t0_fn = time.time()
         try:
-            outputs = self.fn()
+            outputs = self.fn(self.output_idx)
         except Exception:
             if hasattr(self.fn, 'position_of_error'):
                 # this is a new vm-provided function or c linker
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -864,7 +864,7 @@ class Function(object):
         # Do the actual work
         t0_fn = time.time()
         try:
-            outputs = self.fn(self.output_idx)
+            outputs = self.fn(output_indices=self.output_idx)
         except Exception:
             if hasattr(self.fn, 'position_of_error'):
                 # this is a new vm-provided function or c linker
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -756,7 +756,8 @@ class Function(object):
         # TODO: assert that there's a list of outputs
         self.output_idx = outputs # indirectly passing output indices
         # assuming for now I pass all the inputs
-        return [self.__call__(*inputs)[i] for i in outputs]
+        result = self.__call__(*inputs)
+        return [result[i] for i in outputs]
 
     def __call__(self, *args, **kwargs): # sygi: what those kwargs mean?
         profile = self.profile
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -752,16 +752,41 @@ class Function(object):
         f_cpy.maker.fgraph.name = name
         return f_cpy
 
-    def apply(self, inputs, outputs):
-        # TODO: assert that there's a list of outputs
-        self.output_idx = outputs # indirectly passing output indices
-        # assuming for now I pass all the inputs
-        result = self.__call__(*inputs)
-        return [result[i] for i in outputs]
+    def execute(self, *inputs, **kwargs):
+        """
+        Wrapper for `self.__call__` allowing to calculate only a subset of
+        outputs.
 
-    def __call__(self, *args, **kwargs): # sygi: what those kwargs mean?
+        Parameters
+        ----------
+        inputs : list
+            List of inputs to the function. All inputs are required, even when
+            some of them are not necessary to calculate requested subset of
+            outputs.
+
+        kwargs : dict
+            Optional keyword argument `output_subset` is a list of indices of
+            the function's outputs that are requested to be calculated.
+
+        Returns
+        -------
+        List of outputs on indices `output_subset` or all of them, if
+        `output_subset` is not passed.
+        """
+
+        output_subset = kwargs.get('output_subset', range(len(self.outputs)))
+        # TODO: when passed an integer instead of a list, return just that one
+        # result
+
+        result = self.__call__(*inputs, output_subset=output_subset)
+        return [result[i] for i in output_subset]
+
+    def __call__(self, *args, **kwargs):
         profile = self.profile
         t0 = time.time()
+
+        output_subset = kwargs.pop('output_subset')\
+                if kwargs.has_key('output_subset') else None
 
         # Reinitialize each container's 'provided' counter
         if self.trust_input:
@@ -864,7 +889,7 @@ class Function(object):
         # Do the actual work
         t0_fn = time.time()
         try:
-            outputs = self.fn(output_indices=self.output_idx)
+            outputs = self.fn(output_subset=output_subset)
         except Exception:
             if hasattr(self.fn, 'position_of_error'):
                 # this is a new vm-provided function or c linker
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -752,41 +752,33 @@ class Function(object):
         f_cpy.maker.fgraph.name = name
         return f_cpy
 
-    def execute(self, *inputs, **kwargs):
+    def __call__(self, *args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


theano/compile/function_module.py
```diff
@@ -751,7 +752,13 @@ def checkSV(sv_ori, sv_rpl):
...
         return f_cpy
 
-    def __call__(self, *args, **kwargs):
+    def apply(self, inputs, outputs):
+        # TODO: assert that there's a list of outputs
+        self.output_idx = outputs # indirectly passing output indices
+        # assuming for now I pass all the inputs
+        return [self.__call__(*inputs)[i] for i in outputs]
+
+    def __call__(self, *args, **kwargs): # sygi: what those kwargs mean?
```
nouiz: **something

in a function signature, cause something to be a dict with all keyword parameter that aren't matching previous parameter.

Currently, there is mostly just one useful that exist and it exist only for the CVM, it is n_calls=N. If passed, this will do a loop in C where the function will be executed N times. I think we return only the last result. This is useful for small function to help lower the overhead when the function don't take any inputs.

theano/compile/function_module.py
```diff
@@ -751,7 +752,13 @@ def checkSV(sv_ori, sv_rpl):
         f_cpy.maker.fgraph.name = name
         return f_cpy
 
-    def __call__(self, *args, **kwargs):
+    def apply(self, inputs, outputs):
+        # TODO: assert that there's a list of outputs
+        self.output_idx = outputs # indirectly passing output indices
+        # assuming for now I pass all the inputs
+        return [self.__call__(*inputs)[i] for i in outputs]
```
nouiz: Do only 1 call to \__call__ otherwise, there will be duplicate computation of the common part. But this is a good direction.

theano/compile/function_module.py
```diff
@@ -856,7 +863,7 @@ def __call__(self, *args, **kwargs):
         # Do the actual work
         t0_fn = time.time()
         try:
-            outputs = self.fn()
+            outputs = self.fn(self.output_idx)
```
nouiz: make this an keyword argument to prevent conflitct with other keyword argument.

theano/gof/tests/test_vm.py
```diff
@@ -193,6 +193,15 @@ def time_linker(name, linker):
...
                                                        use_cloop=True))
 
+def test_partial_function():
+    def callback(node, thunk, storage_map, compute_map): # to run as Stack
+        pass
+
+    x = tensor.scalar('input')
+    y = x ** 2
+    f = theano.function([x], [y + 7, y - 9, y / 14.], mode=Mode(optimizer=None,
+        linker=vm.VM_Linker(callback=callback)))
```
nouiz: add the parameter use_cloop=False for now. Otherwise, it could call the cvm. You can remove callback and use lazy=True to force the Stack implementation.

theano/gof/vm.py
```diff
@@ -330,7 +330,7 @@ def __init__(self, nodes, thunks, pre_call_clear,
 
         self.allow_gc = allow_gc
         self.message = ""
-        self.base_apply_stack = [o.owner for o in fgraph.outputs if o.owner]
+        self.base_apply_stack = [o.owner for o in fgraph.outputs if o.owner] # sygi: why it can be False?
```
nouiz: This would happen if one of the output is a Constant. This could happen due to Theano optimization.

theano/gof/vm.py
```diff
@@ -405,8 +405,13 @@ def __call__(self):
...
             compute_map[k][0] = (k.owner is None)
 
-        # apply_stack contains nodes
-        apply_stack = list(self.base_apply_stack)
+        if output_indices is not None:
+            output_subset = [self.outputs[i] for i in output_indices]
+            apply_stack = [o.owner for o in output_subset]
+        else:
+            # apply_stack contains nodes
+            apply_stack = list(self.base_apply_stack) # sygi: why list?
```
nouiz: we call pop() on it. Tuple don't have that I think.

nouiz:
It seem a good start. I had only a few comments. You are in the right directly. I don't see much thing misssing, so think of what you could add to your proposal to have work for the full GSoC.

theano/compile/function_module.py
sygi: Thanks for the answer. I found the usage of `n_calls`, but it is on the `fn` attribute of the `Function` instance (`theano_conv.fn(n_calls=n_calls)`), and here is the `Function` instance itself. I am asking about that, because I'd prefer to pass the output_idx through the kwarg, but there's an assert later that prevents me from doing that.

theano/compile/function_module.py
sygi: Done, although as far as I see it executes [vm.__call__()](https://github.com/sygi/Theano/blob/partial-evaluation/theano/gof/vm.py#L397), which didn't have any arguments before.

theano/gof/vm.py
sygi: Right, but as far as I see the only place where you touch `base_apply_stack` is [here](https://github.com/Theano/Theano/blob/master/theano/gof/vm.py#L333), so it should be a list anyway.

nouiz:
I killed some travis run to don't have it overloaded.

sygi:
Fine, sorry for that, I didn't know it may be a problem. I'll try to limit myself with the amount of jobs I start.

sygi:
Closing to prevent travis from re-adding to queue.

theano/compile/function_module.py
```diff
@@ -752,9 +753,33 @@ def checkSV(sv_ori, sv_rpl):
...
+        Returns
+        -------
+        List of outputs on indices `output_subset` or all of them, if
+        `outpus_subset` is not passed. If there's only one output, returns just
+        the value.
+        """
         profile = self.profile
         t0 = time.time()
 
+        output_subset = kwargs.pop('output_subset') if\
```
nouiz: This would be simpler:

output_subset = kwargs.pop('output_subset', range(len(self.outputs))

theano/gof/vm.py
```diff
@@ -809,7 +819,8 @@ def make_vm(self, nodes, thunks,
         pre_call_clear = [storage_map[v] for v in self.no_recycling]
 
         if (self.callback is not None or
-                (config.profile and config.profile_memory)):
+                (config.profile and config.profile_memory) or
+                (self.allow_partial_eval and not self.use_cloop)):
```
nouiz: For now, as it isn't implemented in CVM, remove the (and not self.use_cloop). Otherwise, if use_cloop is True, then it won't allow partial evaluation.

theano/gof/vm.py
```diff
@@ -406,7 +406,12 @@ def __call__(self):
             compute_map[k][0] = (k.owner is None)
 
         # apply_stack contains nodes
-        apply_stack = list(self.base_apply_stack)
+        if output_subset is not None:
+            output_subset = [self.outputs[i] for i in output_subset]
+            apply_stack = [o.owner for o in output_subset]
```
nouiz: To lower Python overhead, this can be done with only 1 list comprehension.

theano/gof/tests/test_vm.py
```diff
@@ -194,6 +194,17 @@ def time_linker(name, linker):
...
 
+def test_partial_function():
+    x = tensor.scalar('input')
+    y = x ** 2
+    f = theano.function([x], [y + 7, y - 9, y / 14.], mode=Mode(
+        optimizer=None, linker=vm.VM_Linker(
+            lazy=True, use_cloop=False, allow_partial_eval=True)))
+
+    assert f(3, output_subset=[0, 1, 2]) == f(3)
+    assert f(4, output_subset=[0, 2]) == [f(4)[0], f(4)[2]]
```
nouiz: call f without output_subset to make sure it still work normally after.

nouiz: I mean, make a new assert with f(5).

theano/compile/function_module.py
nouiz: To lower python overhead, I would set output_subset to None if it isn't present. This would prevent object creation. But otherwise, it is good.

theano/compile/function_module.py
```diff
@@ -752,9 +753,33 @@ def checkSV(sv_ori, sv_rpl):
...
+        ----------
+        args : list
+            List of inputs to the function. All inputs are required, even when
+            some of them are not necessary to calculate requested subset of
+            outputs.
+
+        kwargs : dict
+            TODO: other kwargs?
+            Keyword argument `output_subset` is a list of indices of the
+            function's outputs that are requested to be calculated.
```
nouiz: Theano can have inputs with name and we can pass those parameter as kwargs. This isn't frequently used. We could just prevent a name as output_subset, or force people to use the apply interface for that. I'm not sure what is the best.

@lamblin @abergeron What do you think? We want to use a optional kwargs, but this would conflict with a named input of the same name that we would like to pass as a named parameter.

theano/compile/function_module.py
```diff
@@ -375,6 +375,7 @@ def __init__(self, fn, input_storage, output_storage, indices, outputs,
         self.name = None
         self.nodes_with_inner_function = []
         self.output_keys = output_keys
+        self.output_idx = None
```
nouiz: not used anymore.

theano/compile/function_module.py
nouiz: A previous version was using this interface:

a_theano_function.apply(*inputs, output_subset) or something similar. The alternatives I see to fix the issue of conflict name:

1) disallow named variable with the name output_subste
2) use a different way to call Theano function, like a_theano_function.apply(...)
3) Remove calling with named inputs. (I don't like that much, but I'm not strongly again this).

sygi: It's worth noting that the issue arose at all, because to do 2) we would still have to pass the information about the output subset to the __call__ anyway, and would have to do it indirectly (both all regular and keyword arguments are already potentially being used).

nouiz: Just to say that if we use a different way of calling the Theano function,
we could only allow the normal input to be passed as positional arguments.

I have a slight preference to not allow kwargs parameter to use the name
output_subset. It isn't frequently used, so I don't think it would cause
problems.

I would always disallow that name in kwargs, just to prevent people from
using the bad name without partial eval and then later trying partial eval.

PR #4261, opened 2016-03-18
Better error message

nouiz:
It is to help debug a corner case on Windows found by in pymc3

### DIFF
--- a/theano/gof/link.py
+++ b/theano/gof/link.py
@@ -157,7 +157,6 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
         detailed_err_msg += ("Inputs shapes: %s" % shapes +
                              "\nInputs strides: %s" % strides +
                              "\nInputs values: %s" % scalar_values)
-        detailed_err_msg += "\nInputs variables: %s" % str(node.inputs)
         if theano.config.exception_verbosity == 'high':
             detailed_err_msg += "\nInputs type_num: %s" % str(
                 [getattr(getattr(i[0], 'dtype', ''), 'num', '') for i in thunk.inputs])
--- a/theano/gof/link.py
+++ b/theano/gof/link.py
@@ -158,6 +158,8 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
                              "\nInputs strides: %s" % strides +
                              "\nInputs values: %s" % scalar_values)
         detailed_err_msg += "\nInputs variables: %s" % str(node.inputs)
+        detailed_err_msg += "\nInputs type_num: %s" % str(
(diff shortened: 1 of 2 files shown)


nouiz:
do not merge yet.

nouiz:
I think it is ready to review.

lamblin:
LGTM.
OK to merge when the pymc3 user confirms it prints the right thing for them.

theano/gof/link.py
```diff
@@ -157,6 +157,9 @@ def raise_with_op(node, thunk=None, exc_info=None, storage_map=None):
         detailed_err_msg += ("Inputs shapes: %s" % shapes +
                              "\nInputs strides: %s" % strides +
                              "\nInputs values: %s" % scalar_values)
+        if theano.config.exception_verbosity == 'high':
+            detailed_err_msg += "\nInputs type_num: %s" % str(
+                [getattr(getattr(i[0], 'dtype', ''), 'num', '') for i in thunk.inputs])
```
lamblin: Not working apparently, since it prints `''` each time.

theano/tensor/subtensor.py
```diff
@@ -1934,7 +1934,7 @@ def copy_of_x(self, x):
         # max_depth: we pass 0 to have this parameter ignored
         # requirements: here we pass NPY_ARRAY_ENSURECOPY to force a copy
         # context: this is almost always NULL, I'm not sure what it's used for
-        return """(PyArrayObject*)PyArray_FromAny(py_%(x)s, NULL, 0, 0,
+        return """(PyArrayObject*)PyArray_FromAny(py_%(x)s, PyArray_DTYPE(%(x)s), 0, 0,
```
abergeron: Maybe the comment above needs to be updated.

nouiz:
I updated this PR.

abergeron:
Ok with me, I'll let @lamblin see if his concerns are addressed.

nouiz:
It don't always works. In the user case we wanted to debug, it wasn't
working, by when I test manually it work.

So as it work some times, I would merge this.

I was not able to reproduce the not working case.

lamblin:
It makes a better error message, and sometimes it helps, so I'm OK with merging.
It would be nice to have a way of reproducing the issue, though...

theano/tensor/subtensor.py
lamblin: According to the doc, `PyArray_DTYPE` returns a borrowed reference, and `PyArray_FromAny` steals it, so we need to call `Py_INCREF` on that dtype.

nouiz: This is complicated. As this was not confirmed to help on windows, I'll revert it.

nouiz: https://github.com/Theano/Theano/pull/4606

Issue #1022, opened 2012-10-23
speed/memory opt: sqr(W.dimshuffle('x',0,1) * G.dimshuffle(0,'x',1)).sum(axis=(1,2)) -> T.dot(...)


nouiz:
The pylearn-dev mailing list email 

https://groups.google.com/d/msg/pylearn-dev/iY7swxgn-xI/N9ynRaZMFSIJ

tell that they manually added this optimization:

we want is an optimization that replaces
T.sqr( W.dimshuffle('x',0,1) * G.dimshuffle(0,'x',1) ).sum(axis=(1,2))
with
T.dot(T.sqr(G), T.sqr(W).sum(axis=0))

We should do an optimization that do it automatically. W and G are matrices. We can probably do a more general optimization that replace more cases.

hantek:
Hi 

Is there any progresses on this issue? I come across this problem again, and I am wandering how hard the optimization would be, to see if I am able to make it.

Zhouhan

nouiz:
It isn't hard to do this opt.

If you didn't do an opt, there is a tutorial in this git repo:

https://github.com/abergeron/ccw_tutorial_theano

The pdf: https://github.com/abergeron/ccw_tutorial_theano/blob/master/advanced.pdf have a section on optimization. I think this is the best doc we have for this. There is exercices for opt with the answer. Try them first, it should help you.

Having this done would be great. If you have questions, do not hesitate.

hantek:
Hi Fred,

I'd appreciate it a lot if I can do that in the next CCW ticket.

Zhouhan

nouiz:
done

PR #4263, opened 2016-03-18
sumsqr2dot added to the opt.py

shabanian:
#1022 
@adri-romsor
@nouiz

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3593,7 +3593,7 @@ def local_sumsqr2dot(node):
         if (in1.owner and isinstance(in1.owner.op, T.Elemwise) and isinstance(in1.owner.op.scalar_op, theano.scalar.basic.Sqr)):
             in_sqr = in1.owner.inputs[0]
             if (in_sqr.owner and isinstance(in_sqr.owner.op, T.Elemwise) and
-                    isinstance(in_sqr.owner.op.scalar_op, theano.scalar.basic.Mul)):
+                    isinstance(in_sqr.owner.op.scalar_op, theano.scalar.basic.Mul) and len(in_sqr.owner.inputs) == 2):
                 in_mul1, in_mul2 = in_sqr.owner.inputs
 
                 if (isinstance(in_mul1.owner.op, T.elemwise.DimShuffle) and in_mul1.owner.op.new_order == ('x', 0, 1) and
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3575,6 +3575,39 @@ def local_join_make_vector(node):
 
 
 #################
+#  speed/memory #
+#################
+
+@register_specialize
+@gof.local_optimizer([T.elemwise.Sum])
+def local_sumsqr2dot(node):
+    """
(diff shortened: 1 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -3575,6 +3575,39 @@ def local_join_make_vector(node):
 
 
 #################
+#  speed/memory #
+#################
+
+@register_specialize
```
nouiz: I would register it at canonicalize

theano/tensor/tests/test_opt.py
```diff
@@ -6123,6 +6123,24 @@ def test3(self):
         assert len(divs) == 0
 
 
+def test_local_sumsqr2dot():
+    G = matrix('G')
+    W = matrix('W')
+
+    y = T.sqr( W.dimshuffle('x',0,1) * G.dimshuffle(0,'x',1) ).sum(axis=(1,2))
+
+    f = function([W, G], y)
```
nouiz: use something similar, this should fix travis:

   MODE = theano.compile.get_default_mode().including('local_expm1')

theano/tensor/tests/test_opt.py
```diff
@@ -6173,6 +6173,27 @@ def test_local_zero_div():
...
+
+    f = function([W, G], y, mode=MODE)
+
+    w_val = numpy.random.rand(4, 3).astype(config.floatX)
+    g_val = numpy.random.rand(5, 3).astype(config.floatX)
+
+    f_val = f(w_val, g_val)
+    f_test = function([W,G], T.dot(T.sqr(G), T.sqr(W).sum(axis=0)), mode=MODE)
+
+    assert numpy.allclose(f_val, f_test(w_val, g_val))
```
nouiz: As this test exactly what the optimization do, if the optimization do a bad conversion, we won't detect it. So for the "golden" answer, compute it like the original graph, but in numpy and not Theano.

theano/tensor/opt.py
```diff
@@ -3575,6 +3575,39 @@ def local_join_make_vector(node):
...
+    if (isinstance(node.op, T.elemwise.Sum) and
+            isinstance(node.op.scalar_op, theano.scalar.basic.Add) and node.op.axis == (1, 2)):
+        in1 = node.inputs[0]
+        out = node.outputs[0]
+
+        if (in1.owner and isinstance(in1.owner.op, T.Elemwise) and isinstance(in1.owner.op.scalar_op, theano.scalar.basic.Sqr)):
+            in_sqr = in1.owner.inputs[0]
+            if (in_sqr.owner and isinstance(in_sqr.owner.op, T.Elemwise) and
+                    isinstance(in_sqr.owner.op.scalar_op, theano.scalar.basic.Mul)):
+                in_mul1, in_mul2 = in_sqr.owner.inputs
```
nouiz: Mul op can have more then 2 inputs. So this line can generate an error that will be skip during optimization, but would display an ugly warning to the user. Just add an len(in_seq_.owner.inputs) == 2 in the cond above to prevent that.

shabanian:
@adri-romsor
@nouiz

PR #4265, opened 2016-03-18
Repair the -rpath linker option in Windows System

mangwang:
Due to the '-rpath,' option is not understood by the Microsoft linker, I add an judgement to ignore this linker option when running in Windows System. Please see the code for details.

Best wishes,

Mang

### DIFF
--- a/theano/sandbox/cuda/nvcc_compiler.py
+++ b/theano/sandbox/cuda/nvcc_compiler.py
@@ -261,7 +261,7 @@ class NVCC_compiler(Compiler):
         for pa in preargs:
             if pa.startswith('-Wl,'):
                 # the -rpath option is not understood by the Microsoft linker
-                if sys.platform != 'win32':
+                if sys.platform != 'win32' or not pa.startswith('-Wl,-rpath'):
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/nvcc_compiler.py
```diff
@@ -260,8 +260,10 @@ def compile_str(
         preargs2 = []
         for pa in preargs:
             if pa.startswith('-Wl,'):
-                preargs1.append('-Xlinker')
-                preargs1.append(pa[4:])
+                # the -rpath option is not understood by the Microsoft linker
+                if sys.platform != 'win32':
```
nouiz: This condition remove all -Wl parameter. Can you be more specific like "if sys.platfrom != win32 or not ps.startswith('-Wl,-rpath')"

mangwang: Yeah, exactly. Fiexed it.

nouiz:
Can you tell us if this cause just a warning or a crash?

thanks

mangwang:
It caused a warning and the compile process would go on.

PR #4271, opened 2016-03-21
[WIP] Scan with checkpoint

carriepl:
This is work in progress. It will probably need more documentation, more (and more rigorous, the current ones only print stuff) tests, more features, better default value for the new function argument, etc. I'm only making a PR so that there is a trace of this work.

This adds a new function `theano.scan_with_checkpoints` with an interface similar to scan but it is designed to reduce memory usage as much as possible compared to the original, at the cost of additional computation. The general idea is, during the fprop, only save the states every `n` timesteps instead of every timestep. This means that, during the bprop, a partial recomputation of the fprop will be required but much less data will ever need to be in memory at the same time. 

Currently, the code has several assumptions :
    - Every sequence has the same length
    - If n_steps is specified, it has the same value as the length of any sequence
    - The value of "save_every_N" divides the number of steps the Scan will
      run without remainder
    - Only singly-recurrent and non-recurrent outputs are used.
      No multiple recurrences.
    - Only the last timestep of any output will ever be used.

### DIFF
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -73,7 +73,8 @@ from theano.misc.safe_asarray import _asarray
 
 from theano.printing import pprint, pp
 
-from theano.scan_module import scan, map, reduce, foldl, foldr, clone
+from theano.scan_module import (scan, map, reduce, foldl, foldr, clone,
+                                scan_with_checkpoints)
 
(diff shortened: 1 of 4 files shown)


nouiz:
continued in https://github.com/Theano/Theano/pull/4996

PR #4273, opened 2016-03-21
Add flag: profiling.ignore_first_call and Update How To release.

nouiz:


nouiz:
Useful if we use time_once.

PR #4279, opened 2016-03-22
Partial function evaluation

### LINKED ISSUE #2472: Partial evaluation of a Theano function

sygi:
An implementation of the partial function evaluation on the VM.

First step to fix #2472.

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -752,8 +752,31 @@ class Function(object):
         return f_cpy
 
     def __call__(self, *args, **kwargs):
+        """
+        Evaluates value of a function on given arguments.
+
+        Parameters
+        ----------
+        args : list
+            List of inputs to the function. All inputs are required, even when
+            some of them are not necessary to calculate requested subset of
+            outputs.
+
+        kwargs : dict
+            TODO: other kwargs?
+            Keyword argument `output_subset` is a list of indices of the
+            function's outputs that are requested to be calculated.
+
+        Returns
+        -------
+        List of outputs on indices `output_subset` or all of them, if
+        `outpus_subset` is not passed. If there's only one output, returns just
+        the value.
+        """
         profile = self.profile
         t0 = time.time()
+
+        output_subset = kwargs.pop('output_subset', None)
 
         # Reinitialize each container's 'provided' counter
         if self.trust_input:
@@ -856,7 +879,7 @@ class Function(object):
         # Do the actual work
         t0_fn = time.time()
         try:
-            outputs = self.fn()
+            outputs = self.fn(output_subset=output_subset)
         except Exception:
             if hasattr(self.fn, 'position_of_error'):
                 # this is a new vm-provided function or c linker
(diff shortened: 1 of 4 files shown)


theano/compile/function_module.py
```diff
@@ -963,11 +971,18 @@ def __call__(self, *args, **kwargs):
...
+                if output_subset is None:
+                    return dict(izip(self.output_keys, outputs))
+                else:
+                    return dict((self.output_keys[index], outputs[index])
+                                for index in output_subset)
 
             if output_subset is None:
                 return outputs
             else:
+                if not all(isinstance(index, int) for index in output_subset):
```
nouiz: I would not do that. It will cause extra overhead to good call. Bad call will still give error that I think is clear enough as we will see the code that is a one liner.

theano/gof/vm.py
```diff
@@ -809,13 +820,18 @@ def make_vm(self, nodes, thunks,
         pre_call_clear = [storage_map[v] for v in self.no_recycling]
 
         if (self.callback is not None or
-                (config.profile and config.profile_memory)):
+                (config.profile and config.profile_memory) or
+                self.allow_partial_eval):
```
nouiz: do getattr(self, 'allow_partial_eval', False) to work with old pickled VM_Linker instance. This make travis fail.

theano/compile/function_module.py
sygi: Done. I also removed the dict check above, as it was analogous.

theano/gof/vm.py
sygi: Thanks. I wondered what was wrong.

sygi:
Hi @nouiz, could you take a look?

theano/compile/function_module.py
```diff
@@ -752,9 +752,35 @@ def checkSV(sv_ori, sv_rpl):
...
+
+        Parameters
+        ----------
+        args : list
+            List of inputs to the function. All inputs are required, even when
+            some of them are not necessary to calculate requested subset of
+            outputs.
+
+        kwargs : dict
+            TODO: other kwargs?
```
nouiz: User can pass inputs by kwargs. The key can be the input variable name or instance. What about appending something like this to your first paragraph and removing this todo:

The function inputs can be passed as keyword argument. For this, use the name of the input or the input instance as the key.

nouiz:
I think we should add it to the doc, but i'm not sure where. The \__call__ method don't get documented via the autodoc in http://deeplearning.net/software/theano/library/compile/function.html#theano.compile.function_module.Function
(I generated it locally by calling "python doc/scripts/docgen.py --nopdf")

Do you want to look at this?

Also, there is the C version that would be great to do. But I forgot if you wanted/had time for it.

Otherwise, we can merge the PR except for my small previous comment.

sygi:
Thanks for the comment. I added `__call__` function to autodoc.
I'd like to take a look at the C code, but I'll be busy in the nearest future, so if someone'll be eager to start that part, you should help them.

nouiz:
I didn't hear anyone wanting to work on this. The last commit seem good, so I'll merge when travis pass.

Tell us when you have more time to continue this.

thanks!

PR #4285, opened 2016-03-23
[MRG] UnicodeDecodeError exceptions due to accents

glouppe:
This patch fixes some `UnicodeDecodeError` exceptions due to accents in Cuda files. 

I encountered this issue on a fresh Python 3.5 install.

### DIFF
--- a/theano/sandbox/cuda/corr3d_gemm.cu
+++ b/theano/sandbox/cuda/corr3d_gemm.cu
@@ -199,7 +199,7 @@ void col2im3d(const float* data_col, const int channels,
 
 
 // Theano op code
-// Authors: Arjun Jain, Frédéric Bastien, Jan Schlüter, Nicolas Ballas
+// Authors: Arjun Jain, Frederic Bastien, Jan Schluter, Nicolas Ballas
 // Reference code: https://github.com/BVLC/caffe/blob/master/src/caffe/layers/conv_layer.cu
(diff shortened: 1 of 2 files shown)


nouiz:
thanks. Strange that it don't cause problem before. Can you tell us more about your config? Which OS and which CUDA version?

glouppe:
It runned fine on my laptop, but failed on the NYU HPC cluster. There, the cuda version is 7.5.18. My Python distribution is a fresh Python 3.5 installed with Anaconda. Theano is the dev version.

nouiz:
What is the OS? I would be surprised that Python 3.5 is the cause of this
and we use nvcc 7.5.17 that don't have this problem.

glouppe:
```
[gl1488@login-0-1 ~]$ lsb_release -a
LSB Version:	:core-4.0-amd64:core-4.0-noarch:graphics-4.0-amd64:graphics-4.0-noarch:printing-4.0-amd64:printing-4.0-noarch
Distributor ID:	CentOS
Description:	CentOS release 6.3 (Final)
Release:	6.3
Codename:	Final
[gl1488@login-0-1 ~]$ uname -a
Linux login-0-1.local 2.6.32-431.29.2.el6.x86_64 #1 SMP Tue Sep 9 21:36:05 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
```

nouiz:
I don't understand why it wouldn't work, but it is fixed:)

PR #4287, opened 2016-03-23
Make sure the input of Neg.impl is not a numpy.bool_

lamblin:
Otherwise, -False is True, and -True is False.

This fixes an existing test in DebugMode.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -2399,9 +2399,19 @@ round_half_away_from_zero = RoundHalfAwayFromZero(same_out_float_only)
 
 
 class Neg(UnaryScalarOp):
+    # We can use numpy.negative here, because even if it gives unexpected
+    # results on Boolean arrays, it will be passed other dtypes as Theano
+    # does not have a Boolean type for tensors.
     nfunc_spec = ('negative', 1, 1)
 
     def impl(self, x):
+        # We have to make sure x is not a numpy.bool_, because
+        # `-numpy.bool_(True)` is `False` (we want 0), and
+        # `-numpy.bool_(False)` is `True` (we want 1).
+        # This happens for Composite, as the intermediate results are not
+        # casted in the dtype of the intermediate variable in general.
+        if isinstance(x, numpy.bool_):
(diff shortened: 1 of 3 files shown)


theano/scalar/basic.py
```diff
@@ -2399,9 +2399,19 @@ def c_code(self, node, name, inputs, outputs, sub):
...
     nfunc_spec = ('negative', 1, 1)
 
     def impl(self, x):
+        # We have to make sure x is not a numpy.bool_, because
+        # `-numpy.bool_(True)` is `False` (we want 0), and
+        # `-numpy.bool_(False)` is `True` (we want 1).
+        # This happens for Composite, as the intermediate results are not
+        # casted in the dtype of the intermediate variable in general.
+        if isinstance(x, numpy.bool_):
+            x = numpy.int8(x)
```
nouiz: Can you tell what this fix? Do this need a test? Why the input would be a bool anyway? It should be a valid dtype as this is during the execution of an op. What cause this bool? Maybe we need to fix that.

lamblin: There is a test that was failing in DebugMode, but we could add a new one.
As I say in the comment, when we have a composite Op, the intermediate values are not casted into the dtype of the intermediate variable. In that case, we had something like `cast{float32}(-(i0 > i1))`, so the intermediate result of `i0 > i1` was a single `numpy.bool_` scalar, and then `Neg.impl` was called on it.

nouiz: It seem I missed that part. The other option would be to cast explicitly in each op that could output a bool. I don't like that better. We can merge when travis pass. A test would be great otherwise, but I don't think it is mandatory.

Which test was failing? I don't recall it. Was it only on one specific platform?

lamblin: The test was theano/sandbox/cuda/tests/test_driver.py:test_nvcc_cast in DebugMode.
It is probably better to have a test just for the Python case, so that we can test without a GPU... I'll add that.

lamblin: Tests added.

PR #4288, opened 2016-03-23
small fix: doc, assert -> python error, printing, code clean up.

nouiz:
The assert was triggered by:

[theano-users] Assertion failed! (File: mod.cu, Line 4072)

### DIFF
--- a/doc/tutorial/faq_tutorial.txt
+++ b/doc/tutorial/faq_tutorial.txt
@@ -64,7 +64,7 @@ Defining the theano function
 
 .. code-block:: python
 
-   f = theano.function(..., updates=updates)
+   f = theano.function(..., updates=[(lookup_table, updates)])
 
 Note that you can compute the gradient of the cost function w.r.t.
 the entire lookup_table, and the gradient will have nonzero rows only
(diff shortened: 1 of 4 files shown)


doc/tutorial/faq_tutorial.txt
```diff
@@ -64,7 +64,7 @@ Defining the theano function
 
 .. code-block:: python
 
-   f = theano.function(..., updates=updates)
```
abergeron: What is this new thing about lookup tables for updates?  Since when is that in?  Why did we not mention that in the release notes?

abergeron:
Ok apart from my confusion at (apparently) new features that I had absolutely no idea where on the way/already there.

doc/tutorial/faq_tutorial.txt
nouiz: This isn't new. It is just new doc for a common case. lookup_table is a matrix shared variable defined at line 26.

abergeron: Ok, it's just really confusing because of the name of the variables and the text below.  It seem to imply that we can have a lookup table on updates to make partial updates or something of the sort, which is apparently not the case.

abergeron:
Ok when travis passes.

PR #4289, opened 2016-03-23
refactored GpuJoin c_code to not hit nested block limit of MSVC

NasenSpray:
MSVC doesn't like functions with too many gotos

Possible discussion topic: C++11 usage

### DIFF
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3502,7 +3502,6 @@ class GpuJoin(tensor.Join, GpuOp):
         PyObject *start, *stop;
         start = NULL;
         stop = NULL;
-
         """ % locals()
 
         # Test negative axis
@@ -3512,7 +3511,6 @@ class GpuJoin(tensor.Join, GpuOp):
                          "Join axis %%d out of bounds [0, %%d)", axis, nd);
             %(fail)s
         }
-
         if( axis < 0 ){
             axis = axis + nd;
         }
@@ -3521,13 +3519,21 @@ class GpuJoin(tensor.Join, GpuOp):
         # getting the shapes of all the involved tensors (input[1:])
         # + check: all input tensors have same shape as final out
         # except for "axis" dimension
-        # shape_%(cdna)s[nd] is initialized before, to prevent following
-        # error: jump to label __label_9 crosses initialization of
-        # shape_%(cdna)s[nd]
-        for i, cdna in enumerate(gof.utils.uniq(inputs[1:])):
-            str += """
-            int shape_%(cdna)s[nd];
-            """ % locals()
+        cdna_map = dict((cdna, i) for i, cdna in enumerate(gof.utils.uniq(inputs[1:])))
+        str += """
+        int shapes[%d][nd];
+        struct { 
+            CudaNdarray* const ptr;
+            int const shape_idx;
+        } const inputs[] = {
+        """ % len(cdna_map)
+        for i, cdna in enumerate(inputs[1:]):
+            cdna_idx = cdna_map[cdna]
+            if i > 0:
+                str += ",\n\t\t"
+            str += "{%(cdna)s, %(cdna_idx)d}" % locals()
+        str += "};"
+
         str += """
         if(-1 == axis && PyErr_Occurred()){
             %(fail)s;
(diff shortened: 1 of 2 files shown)


nouiz:
@lamblin @abergeron what about requesting C++11?

Do you use other C++11 code then the auto keywork? In fact, auto seem to be in C++0x and was in gcc 4.4: https://gcc.gnu.org/gcc-4.4/cxx0x_status.html

I still have a computer with GCC 4.5. I'll try it there.

NasenSpray:
> Do you use other C++11 code then the auto keywork?

`std::copy_n` and range-based for loops. Another in future potentially very useful C++11 features would be variadic templates.

> In fact, auto seem to be in C++0x

sure, C++0x _is_ C++11 after all

nouiz:
It don't compile with g++ 4.8 and cuda 7.5. If I use this diff, it compile with less error:

```
diff --git a/theano/sandbox/cuda/basic_ops.py b/theano/sandbox/cuda/basic_ops.py
index 084d4fc..f593174 100644
--- a/theano/sandbox/cuda/basic_ops.py
+++ b/theano/sandbox/cuda/basic_ops.py
@@ -3642,6 +3642,11 @@ class GpuJoin(tensor.Join, GpuOp):
         return (7,)
 
+        return ["-std=c++11"]
+
 gpu_join = GpuJoin()
 
 
diff --git a/theano/sandbox/cuda/nvcc_compiler.py b/theano/sandbox/cuda/nvcc_compiler.py
index 2b60a8d..18112b9 100644
--- a/theano/sandbox/cuda/nvcc_compiler.py
+++ b/theano/sandbox/cuda/nvcc_compiler.py
@@ -271,7 +271,7 @@ class NVCC_compiler(Compiler):
                             '--prec-div', '--prec-sqrt',  '--use_fast_math',
                             '-fmad', '-ftz', '-maxrregcount',
                             '-prec-div', '-prec-sqrt', '-use_fast_math',
-                            '--use-local-env', '--cl-version=']:
+                            '--use-local-env', '--cl-version=', '-std=']:
 
                 if pa.startswith(pattern):
                     preargs1.append(pa)
```

Error with this diff:

```
mod.cu(196): error: transfer of control bypasses initialization of:
            variable "inputs"
```

Mostly, you should define inputs before the error checking. Did it compile and work for you? I suppose so. Can you add this diff this your PR?

It don't work with g++ 4.5. It don't like the n_copy and auto. So I'm not reading correctly g++ doc about when auto was added.

NasenSpray:
> It don't work with g++ 4.5. It don't like the n_copy and auto. So I'm not reading correctly g++ doc about when auto was added.

GCC was C++11-complete at 4.8.x

abergeron:
I think that requiring GCC 4.8 is a bit too recent for now.

Also, this problem seems to be only in the old gpu backend which is scheduled to go away.

NasenSpray:
> I think that requiring GCC 4.8 is a bit too recent for now.

GCC 4.8.1 was released on May 31, 2013. Is that really too recent?

> Also, this problem seems to be only in the old gpu backend which is scheduled to go away.

Would it be possible to add this fix to a "for MSVC only" branch? The problem affects every Theano/Keras user on Windows and makes it almost impossible to use RNNs there.

nouiz:
We had a quick discussion and we are pretty sure that in some cluster, this
will cause problem. Also, it will probably raise the minimum CUDA version
as it request specific g++ version.

We don't like having a branch of the code that won't be tested on linux as
we don't have buildbot on windows. It will be brittle and have more risk of
betting bugged/broken by mistake. Could you change the code to not use
C++11?

lamblin:
Ubuntu 12.04, still supported, has GCC 4.6 only, for instance, and Anaconda provides only MinGW 4.7 for Windows.
Is that new feature actually required for the fix, though? Can we fix the issue without using `copy_n` or `for (auto it : inputs)`?

NasenSpray:
> Ubuntu 12.04, still supported, has GCC 4.6 only, for instance, and Anaconda provides only MinGW 4.7 for Windows.

Okay, makes sense then.

> Is that new feature actually required for the fix, though? Can we fix the issue without using copy_n or for (auto it : inputs)?

Not really, it's just syntax sugar. I can make a new PR w/o C++11.

lamblin:
@NasenSpray If you can make a new PR without C++11, that would be really great.
Thanks for your help!

nouiz:
I'll make a Theano 0.8.2 release next week. It would be great if we include this fix. Do you have the time to finish it? I'm pretty busy and won,t have the time to finish this.

thanks

NasenSpray:
Sorry, I almost forgot this. What's the deadline? I should have some time to work on it tomorrow.

nouiz:
I don't know when it will be done. It could be today, tomorrow or Wednesday.

thanks

nouiz:
I stumbled on this PR again. Do you have time to finish it?

nouiz:
Note, https://github.com/fchollet/keras/issues/3550 found another work around, using tile instead of concatenate when you concatenate the same tensor many times. That should be faster too.

nouiz:
The new back-end don't have this problem. So closing as we will get rid of the old back-end shortly.

Issue #4262, opened 2016-03-18
gpuarray.GpuJoin with negative axes


lamblin:
This ticket is about cleaning up the fix in #4258.
- [ ] First, we should sort out if `axis` should be changed to a signed type in `gpuarray`, in `pygpu`, or if only Theano Ops should accept negative axes.
- [ ] Then, we should make `GpuJoin.perform` raise `IndexError` instead of `OverflowError` when axis < -ndim, be it by modifying `perform` or something in `pygpu` or in `gpuarray` to add explicit checks, instead of relying on Cython's overflow checks.
- [ ] Finally, in [`test_basic.py`](https://github.com/Theano/Theano/pull/4258/files#diff-bd238af9c7f5e76b67885294b7da00aa), we should change the `assertRaises` to only accept `IndexError`, not `OverflowError`

gokul-uf:
@lamblin I'd like to work on this
I suggest that we do it in Theano Op so that it is caught early/higher up in the compilation process.

abergeron:
I think that all the points are addressed now.

PR #4301, opened 2016-03-27
GpuJoin takes negative values for axis

gokul-uf:
This PR implements the suggestions given in #4262


nouiz:
@abergeron I let you review this one.

PR #4308, opened 2016-03-29
Prevents subprocess from importing Theano

gokul-uf:
This PR fixes issue #1064 by adding a new environment variable to keep track of whether the GPU is locked by a parent process or not. Earlier sent as PR #4226

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -22,6 +22,7 @@ from theano.tensor.basic import register_transfer
 # ignore_newtrees is to speed the optimization as this is the pattern
 # we use for optimization. Otherwise, we can iterate 100s of time on
 # the graph and apply only a few optimizations each time.
+
 gpu_optimizer = EquilibriumDB(ignore_newtrees=False)
 gpu_seqopt = SequenceDB()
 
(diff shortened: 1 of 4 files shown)


nouiz:
travis is failing.

gokul-uf:
@nouiz, travis is complaining that I am importing but not using tga. 
So,instead of just importing, should I do
`import theano.sandbox.cuda as tcn`
`tcn.use()`

nouiz:
you can add '# noea' on the line to have that line not being checked for validation. I'm not sure of the PR and don't have time to review now.

gokul-uf:
This PR makes sure that the subprocesses do not import theano and lock the GPU. this is done by using a environment variable. travis is failing because I am importing but not using.

abergeron:
The stuff to add at the end of the line is `  # noqa` and it is appropriate to use that for the tests here to make flake8 happy.

As for the problem this is addressing I'm not sure.  The main problem with forking is not that subprocesses will use the GPU, it is that forking a cuda context is bad and breaks it.  If we could somehow go around that restriction, then we could use the GPU as much as we wanted in the subprocesses.

gokul-uf:
@abergeron It seems that CUDA does not allow GPUs to be shared between parent and child process and will not be doing so in the near future. I think that until that time comes it is better to have this in to warn the users about what's going on. 
One of the [forum posts](https://devtalk.nvidia.com/default/topic/382954/cuda-programming-and-performance/cuda-and-fork-/) reported that through a bug `__global__` functions can access data from both parent and child. I would suggest that we not use this though.

abergeron:
Since we can't go around the restriction on forking with a GPU active this can't move forward.

Issue #4307, opened 2016-03-29
OpenMP support on Mac OS X


kevin-keraudren:
OpenMP can be installed on a Mac with `brew install clang-omp`. The C++ compiler that needs to be used for the '-fopenmp' flag to work is then:

```python
import theano
theano.config.cxx = '/usr/local/bin/clang-omp++'
```

However, theano expects either g++ or clang++ as a custom compiler, so here:

https://github.com/Theano/Theano/blob/master/theano/gof/cmodule.py#L1828

we should add:

```python
    and 'clang-omp++' not in theano.config.cxx
```

nouiz:
thanks. Do you want to make a PR? We can do it if you prefer.

gokul-uf:
@kevin-keraudren  could you close this? Thanks

nouiz:
done, so closing.

PR #4309, opened 2016-03-29
Adding clang-omp++ as an alternative compiler to clang++ as it suppots OpenMP

kevin-keraudren:
This PR is related to the issue https://github.com/Theano/Theano/issues/4307.


nouiz:
thanks, we can merge when travis pass.

PR #4310, opened 2016-03-29
Small changes.

nouiz:


theano/misc/do_nightly_build_send
```diff
@@ -90,7 +90,7 @@ def mysend(subject, file):
             try: 
                 nb_ran+=int(token)
             except Exception, e:
```
abergeron: This is also a python3 problem, should be `except Execption as e:`

theano/tensor/elemwise.py
```diff
@@ -140,7 +140,8 @@ def __init__(self, input_broadcastable, new_order, inplace=False):
                 # numpy integers.  See
                 # <http://projects.scipy.org/numpy/ticket/2235>.
                 if not isinstance(j, (integer_types, numpy.integer)):
-                    raise TypeError("DimShuffle indices must be python ints.")
+                    raise TypeError(
+                        "DimShuffle indices must be python ints. Got: ", j)
```
lamblin: While you're at it, can you print the type of j?

theano/gof/op.py
```diff
@@ -966,7 +966,7 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
         if new_node is not None:
             node = new_node
 
-        if self._op_use_c_code:
```
lamblin: I don't think that is a good idea, we should fix the get/setstate of the Op in question instead.
Which Op was that?

nouiz: That member was set in Op. So we would need to change all subclass of Op. So I don't see how to fix it otherwise without too much code duplicate. I don't like the idea of implementing Op.\__setstate__. It would be easy to not call it from subclass that implement it.

theano/tensor/elemwise.py

theano/gof/op.py
lamblin: We may need to talk about that IRL, because I don't get what the issue is.
`self._op_use_c_code` is set in `Op.__new__`, so that, as the comment says, "all existing Ops get a _op_use_c_code attribute".
So, the only way for an unpickled Op instance to _not_ have `_op_use_c_code` would be if there was a `__getstate__` implemented for that Op subclass that would get rid of that attribute, since that is not what the Op base class does.
My point is that, in the case where `__getstate__` is implemented, it is reasonable to add a `__setstate__` for that specific subclass. Moreover, getting rid of `_op_use_c_code` is a symptom that the `__getstate__` was not well-written, and we should address that rather than mask it. That is what we did for Gemm and related classes.

Do you know which other Op was missing `_op_use_c_code` after unpickling?

lamblin: OK, so the Op in question is `RandomFunction`, and indeed its `__getstate__` and `__setstate__` methods ignore `_op_use_c_code` (https://github.com/Theano/Theano/blob/master/theano/tensor/raw_random.py#L135-L149). I think we should fix it there.

nouiz: I don't like that fix. It request all \__getstate__ \__setstate__ to know the internal representation of ops.

An alternative is to add a method to Op that would do that and call it from all Op.\__setstate__. As there isn't other attribute from Op directly, it would be less error prone to fix it there. With your suggestion or my alternative, we need to verify all \__getstate__ and \__setstate__ in Theano and outside Theano to make sure they work well. This isn't possible. Fixing it here is a general fix. I already checked the other place we use it and it is guarded by a getattr() too. So it shouldn't pop up elsewhere.

abergeron: I think we should get rid of `_op_use_c_code` and just use theano.config.cxx since otherwise you could end up reloading a graph with `_op_use_c_code` set to True and no C compiler on the machine.  I really don't see why that has to be a property of the ops themselves.

But in any case I support the change above.

nouiz:
I updated this PR.

theano/gof/op.py
```diff
@@ -965,7 +965,13 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
...
         if new_node is not None:
             node = new_node
-
+        if not hasattr(self, '_op_use_c_code'):
+            warnings.warn(
+                "The  __getstate__ method of '%s' is not implemented correctly."
+                " It should keep the attributes added by the base class."
+                " To implement it correctly, it should keep all attributes"
+                " and only remove those it does not want." % (self),
+                stacklevel=2)
```
lamblin: Should we then set `self._op_use_c_code = theano.config.cxx` here, rather than use `getattr` on the line below?

lamblin:
In any case, OK if Travis passes.

theano/tensor/raw_random.py
```diff
@@ -133,15 +133,28 @@ def __init__(self, fn, outtype, inplace=False, ndim_added=0):
...
+        return d
+
+    def __setstate__(self, dct):
+        if isinstance(dct, dict):
+            state = [dct['fn'],
+                     dct['outtype'],
+                     dct['inplace'],
+                     dct['ndim_added']]
+        else:
+            state = dct
```
abergeron: This is still not ok since it will still delete any attributes of the superclass.

lamblin: Oh, right, we should call `self.__dict__.update(...)`.

nouiz: done.

nouiz:
I updated this PR and travis pass.

PR #4323, opened 2016-03-31
Use the new GpuElemwise from libgpuarray

abergeron:
This gives us a nice speedup since we now perform dimension collapsing and call a contiguous kernel when possible.

Further improvements are possible within the libgpuarray code, but shouldn't require more changes here.

### DIFF
--- a/theano/sandbox/gpuarray/elemwise.py
+++ b/theano/sandbox/gpuarray/elemwise.py
@@ -185,7 +185,7 @@ class GpuElemwise(HideC, Elemwise):
         p = 0
         for n, o in enumerate(node.outputs):
             if n in self.inplace_pattern:
-                assert(len(node.outputs) == 1);
+                assert(len(node.outputs) == 1)
                 res += "\nargs[%(n)s].flags |= GE_WRITE;\n" % dict(n=self.inplace_pattern[n])
             else:
                 nn = len(inps) + p
--- a/theano/sandbox/gpuarray/elemwise.py
+++ b/theano/sandbox/gpuarray/elemwise.py
@@ -221,6 +221,7 @@ class GpuElemwise(HideC, Elemwise):
         # check that all inputs have valid dimensions
         emitted_inames = {}
         code = """
+        // +1 is so that MSVC is happy when nd == 0
         size_t dims[%(nd)s+1] = {%(initial_dims)s};
         void *rargs[%(nargs)s] = {0};
         """ % locals()
@@ -269,7 +270,7 @@ class GpuElemwise(HideC, Elemwise):
                 %(fail)s;
             }
         }
-        """ % locals()
+            """ % locals()
             emitted_inames[iname] = True
         # check that all outputs have valid dimensions
         p = len(node.inputs)
@@ -299,7 +300,7 @@ class GpuElemwise(HideC, Elemwise):
             }
         }
         rargs[%(p)s] = &%(oname)s->ga;
-        """ % locals()
+                """ % locals()
                 p += 1
             else:
                 input_idx = self.inplace_pattern[idx]
--- a/theano/sandbox/gpuarray/elemwise.py
+++ b/theano/sandbox/gpuarray/elemwise.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, print_function, division
 import copy
-from theano.compat import izip
 import numpy
 
 import theano
@@ -13,10 +12,9 @@ from theano.tensor.elemwise import (Elemwise, DimShuffle, CAReduceDtype)
 try:
     import pygpu
     from pygpu import gpuarray
-    from pygpu.tools import ScalarArg, ArrayArg
-    from pygpu.elemwise import ElemwiseKernel
+    from pygpu.tools import ArrayArg
     from pygpu.reduction import ReductionKernel
-    from pygpu.gpuarray import dtype_to_typecode, dtype_to_ctype
+    from pygpu.gpuarray import dtype_to_typecode
 except ImportError:
     pass
 
@@ -166,7 +164,7 @@ class GpuElemwise(HideC, Elemwise):
                 '<gpuarray/elemwise.h>']
 
     def c_support_code_struct(self, node, name):
-        return "\nGpuElemwise *ge;\n";
+        return "\nGpuElemwise *ge;\n"
 
     def c_init_code_struct(self, node, name, sub):
         inps, outs = self._get_vnames(node)
--- a/theano/sandbox/gpuarray/elemwise.py
+++ b/theano/sandbox/gpuarray/elemwise.py
@@ -185,6 +185,7 @@ class GpuElemwise(HideC, Elemwise):
         p = 0
         for n, o in enumerate(node.outputs):
             if n in self.inplace_pattern:
+                assert(len(node.outputs) == 1);
                 res += "\nargs[%(n)s].flags |= GE_WRITE;\n" % dict(n=self.inplace_pattern[n])
             else:
                 nn = len(inps) + p
--- a/theano/sandbox/gpuarray/elemwise.py
+++ b/theano/sandbox/gpuarray/elemwise.py
@@ -22,32 +22,12 @@ except ImportError:
 
 from .basic_ops import (as_gpuarray_variable, HideC, GpuKernelBase, Kernel,
                         infer_context_name)
-from .type import GpuArrayType
+from .type import GpuArrayType, gpu_context_type
 from .fp16_help import load_w, write_w
 
 
-def _is_scalar(v):
-    False
-
-
 def make_argument(v, name):
-    if _is_scalar(v):
-        return ScalarArg(numpy.dtype(v.type.dtype), name)
-    else:
-        return ArrayArg(numpy.dtype(v.type.dtype), name)
-
-
-def ensure_allocated(storage, shape, dtype, ctx):
-    odat = storage[0]
-    if odat is not None:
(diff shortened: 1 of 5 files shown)


abergeron:
Also, this requires: https://github.com/Theano/libgpuarray/pull/143

abergeron:
This is ready for review.

nouiz:

/home/travis/build/Theano/Theano/theano/sandbox/gpuarray/elemwise.py:188:47: E703 statement ends with a semicolon

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -178,76 +157,73 @@ def generate_kernel(self, node, nodename):
...
-        return '\n'.join(res)
+        return ['<numpy_compat.h>', '<gpuarray/types.h>',
+                '<gpuarray/elemwise.h>']
+
+    def c_support_code_struct(self, node, name):
+        return "\nGpuElemwise *ge;\n"
+
+    def c_init_code_struct(self, node, name, sub):
+        inps, outs = self._get_vnames(node)
+        nargs = len(inps) + len(outs)
```
nouiz: nargs should be shorter when there is inplace.

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -178,76 +157,73 @@ def generate_kernel(self, node, nodename):
...
         emitted_inames = {}
-        num_kernel_params = 1 + nd + len(inputs + outputs) * (2 + nd)
         code = """
-        size_t n_blocks = 0;
-        size_t threads_per_block = 0;
-        size_t numEls = 0;
-        const ssize_t zero = 0;
-        void *kernel_params[%(num_kernel_params)d] = {0};
-        int err;
+        size_t dims[%(nd)s+1] = {%(initial_dims)s};
```
nouiz: make a comment that the +1 is for windows to don't crash when n is 0.

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -297,9 +269,10 @@ def c_code(self, node, name, inputs, outputs, sub):
                 %(fail)s;
             }
         }
-            """ % locals()
+        """ % locals()
```
nouiz: bad indented line. It is valid python, but make reading harder.

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -325,7 +298,9 @@ def c_code(self, node, name, inputs, outputs, sub):
                 %(fail)s
             }
         }
+        rargs[%(p)s] = &%(oname)s->ga;
         """ % locals()
```
nouiz: idem

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -256,19 +232,15 @@ def c_code(self, node, name, inputs, outputs, sub):
...
-                """ % locals()
+            code += """
+            int broadcasts_%(iname)s[%(nd)s+1] = {%(broadcasts)s};
+            """ % locals()
             emitted_inames[iname] = node.inputs[idx]
 
         # check that all inputs have valid dimensions
         emitted_inames = {}
         for idx, iname in enumerate(inputs):
+            code += "rargs[%(idx)s] = &%(iname)s->ga;\n" % dict(idx=idx, iname=iname)
```
nouiz: I think it is ok. But just to be sure, the generator of elemwise don't know it when some inputs are used more then 1 as inputs.

theano/sandbox/gpuarray/elemwise.py
```diff
@@ -351,92 +326,35 @@ def c_code(self, node, name, inputs, outputs, sub):
...
-        for n, p in enumerate(param):
-            code += "kernel_params[%(n)d] = %(p)s;\n" % locals()
         code += """
-        err = GpuKernel_call(&%(kname)s, 1, &threads_per_block, &n_blocks, 0, kernel_params);
-        if (err != GA_NO_ERROR) {
-            PyErr_Format(PyExc_RuntimeError,
-                         "gpuarray error: %(kname)s: %%s.",
-                         GpuKernel_error(&%(kname)s, err));
-            %(fail)s;
+        if (GpuElemwise_call(ge, rargs, GE_BROADCAST) != GA_NO_ERROR) {
```
nouiz: Here we don't test that we respect the broadcast specified in the Theano graph. We will check it only based on the shape.

If I'm right, fix this or make an issue to fix this. We should not remove user error in the new back-end.

abergeron: There is a check that we respect the theano view before that call.  So we should raise errors if we don't respect that.

theano/sandbox/gpuarray/elemwise.py
abergeron: ok

theano/sandbox/gpuarray/elemwise.py
abergeron: ok

theano/sandbox/gpuarray/elemwise.py
abergeron: Then we should add something here to not reuse inputs that are the same, probably.  This would cut down on indexing overhead.  But it can be part of future code.

theano/sandbox/gpuarray/elemwise.py
abergeron: ok

theano/sandbox/gpuarray/elemwise.py
abergeron: _get_vnames handles that.

theano/sandbox/gpuarray/elemwise.py
nouiz: ok

theano/sandbox/gpuarray/elemwise.py
nouiz: ok

theano/sandbox/gpuarray/elemwise.py
nouiz: can you do an issue about this futur opt.

theano/tensor/tests/test_elemwise.py
```diff
@@ -189,8 +189,8 @@ def with_linker(self, linker, op, type, rand_val):
                          ((2, 3, 4, 5), (1, 3, 1, 5)),
                          ((2, 3, 4, 5), (1, 1, 1, 1)),
                          ((), ())]:
-            x = type('float64', [(entry == 1) for entry in xsh])('x')
-            y = type('float64', [(entry == 1) for entry in ysh])('y')
+            x = type('float32', [(entry == 1) for entry in xsh])('x')
+            y = type('float32', [(entry == 1) for entry in ysh])('y')
             e = op(scalar.add)(x, y)
             f = copy(linker).accept(FunctionGraph([x, y], [e])).make_function()
             xv = rand_val(xsh)
```
nouiz: this still generate float64 data, so it make this test fail.

theano/tensor/tests/test_elemwise.py
```diff
@@ -218,8 +218,8 @@ def with_linker_inplace(self, linker, op, type, rand_val):
                          ((2, 3, 4, 5), (1, 3, 1, 5)),
                          ((2, 3, 4, 5), (1, 1, 1, 1)),
                          ((), ())]:
-            x = type('float64', [(entry == 1) for entry in xsh])('x')
-            y = type('float64', [(entry == 1) for entry in ysh])('y')
+            x = type('float32', [(entry == 1) for entry in xsh])('x')
+            y = type('float32', [(entry == 1) for entry in ysh])('y')
             e = op(scalar.Add(scalar.transfer_type(0)), {0: 0})(x, y)
             f = copy(linker).accept(FunctionGraph([x, y], [e])).make_function()
             xv = rand_val(xsh)
```
nouiz: idem

theano/tensor/tests/test_elemwise.py
```diff
@@ -232,8 +232,8 @@ def with_linker_inplace(self, linker, op, type, rand_val):
             # test Elemwise.infer_shape
             # the Shape op don't implement c_code!
             if isinstance(linker, gof.PerformLinker):
-                x = type('float64', [(entry == 1) for entry in xsh])('x')
-                y = type('float64', [(entry == 1) for entry in ysh])('y')
+                x = type('float32', [(entry == 1) for entry in xsh])('x')
+                y = type('float32', [(entry == 1) for entry in ysh])('y')
                 e = op(scalar.Add(scalar.transfer_type(0)), {0: 0})(x, y)
                 f = copy(linker).accept(FunctionGraph(
                     [x, y], [e.shape])).make_function()
```
nouiz: idem 2 lines bellow.

theano/tensor/tests/test_elemwise.py
```diff
@@ -267,13 +267,15 @@ def test_c_inplace(self):
...
+                                       [self.type, self.ctype],
+                                       [self.rand_val, self.rand_cval]):
+            x = t('float32', [0, 0])('x')
+            y = t('float32', [1, 1])('y')
             e = op(scalar.Second(scalar.transfer_type(0)), {0: 0})(x, y)
             f = linker().accept(FunctionGraph([x, y], [e])).make_function()
-            xv = self.rand_cval((5, 5))
-            yv = self.rand_cval((1, 1))
+            xv = rval((5, 5))
+            yv = rval((1, 1))
```
nouiz: idem

abergeron:
Ok done.

Issue #4314, opened 2016-03-30
Move theano.sandbox.softsign to tensor/nnet/nnet.py


nouiz:
Add tests for it too and make sure it get in the doc.

yobibyte:
I would like to take it.

nouiz:
thanks.

yobibyte:
Done? #4324

nouiz:
yes. I'm closing.

PR #4324, opened 2016-03-31
Moved theano.sandbox.softsign to tensor/nnet/nnet.py. Added test. #4314

yobibyte:
* Moved the code to tensor/nnet/nnet.py (added it to the end, don't know how to do better)
* Added simple value check (found a lot of different ways of tests creating in the code. But don't know what is preferrable)
* Added documentation to the class


theano/sandbox/softsign.py
```diff
@@ -1,35 +0,0 @@
-from __future__ import absolute_import, print_function, division
```
nouiz: To help port old code, can you add a new file with the same name that will import softsign to allow old code to still work?

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1699,3 +1701,10 @@ def test_binary_crossentropy_reshape():
         fga = theano.function([a], ga, mode=mode)
         utt.assert_allclose(fga(numpy.array([[[[30.]]]], dtype=config.floatX)),
                             numpy.zeros((1, 1, 1, 1), dtype=config.floatX))
+
+SoftsignTester = makeBroadcastTester(
+    op=softsign,
+    expected=upcast_int8_nfunc(lambda inputs: check_floatX(
+        inputs, inputs/(1.0+numpy.fabs(inputs)))),
+    name='SoftsignTester',
```
nouiz: Here, this is missing which test to run. you should add at least a good parameter. You can check the AddTester example in tensor/tests/test_basic.py.

nouiz:
Thanks for the PR, I have only 2 small comments.

yobibyte:
I have a question. If I do all the changes, is it okay to do push --force within my branch? Or how what should I do in order to keep the history clean?

nouiz:
If it is just a few fixes like this, personally, I prefer that you just
create new commit. It make review easier. But if it ends up being big
changes, then yes doing a force push is fine.

But if you still make a force push, we will still review it.

yobibyte:
Done.

theano/sandbox/softsign.py
```diff
@@ -1,35 +1,4 @@
...
-        z, = out
-        if node.inputs[0].type in [theano.scalar.float32,
-                                   theano.scalar.float64]:
-            return "%(z)s = %(x)s / (1.0+fabs(%(x)s));" % locals()
-        raise NotImplementedError('only floating point x is implemented')
-
-scalar_softsign = ScalarSoftsign(theano.scalar.upgrade_to_float,
-                                 name='scalar_softsign')
-softsign = theano.tensor.Elemwise(scalar_softsign, name='softsign')
+softsign = softsign  # otherwise build fails due to unused import
```
lamblin: It is OK to silence flake8 like that, FYI another way would have been to use `# noqa`, for instance:
```
from theano.tensor.nnet.nnet import softsign  # noqa
```

Also, do you think you could add a deprecation warning in that file, telling people to import from `theano.tensor.nnet.nnet` instead?

yobibyte: Yeah, will do that.

On Wed, Apr 6, 2016 at 6:36 PM, Pascal Lamblin <notifications@github.com>
wrote:

> In theano/sandbox/softsign.py
> <https://github.com/Theano/Theano/pull/4324#discussion_r58738547>:
>
> > -            return [gz / (d * d)]
> > -        else:
> > -            return NotImplemented
> > -
> > -    def c_code(self, node, name, inp, out, sub):
> > -        x, = inp
> > -        z, = out
> > -        if node.inputs[0].type in [theano.scalar.float32,
> > -                                   theano.scalar.float64]:
> > -            return "%(z)s = %(x)s / (1.0+fabs(%(x)s));" % locals()
> > -        raise NotImplementedError('only floating point x is implemented')
> > -
> > -scalar_softsign = ScalarSoftsign(theano.scalar.upgrade_to_float,
> > -                                 name='scalar_softsign')
> > -softsign = theano.tensor.Elemwise(scalar_softsign, name='softsign')
> > +softsign = softsign  # otherwise build fails due to unused import
>
> It is OK to silence flake8 like that, FYI another way would have been to
> use # noqa, for instance:
>
> from theano.tensor.nnet.nnet import softsign  # noqa
>
> Also, do you think you could add a deprecation warning in that file,
> telling people to import from theano.tensor.nnet.nnet instead?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/Theano/Theano/pull/4324/files/e41c5e321687a4b93eed3191ade02fbc8c781ff0#r58738547>
>



-- 
С уважением,
Курин Виталий

nouiz:
thanks!

Issue #4328, opened 2016-04-02
CudaNdarray errors in nosetests


gokul-uf:
While running tests on the bleeding edge, I get the following errors

```
Ran 21758 tests in 12174.397s
FAILED (SKIP=72, errors=7)

======================================================================
ERROR: test_set_value (theano.sandbox.cuda.tests.test_basic_ops.test_shared_options)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gokul/Theano/theano/tensor/tests/test_sharedvar.py", line 256, in test_set_value
    get_x /= values_to_div
TypeError: unsupported operand type(s) for /=: 'CudaNdarray' and 'CudaNdarray'

======================================================================
ERROR: test_shared_do_alias (theano.sandbox.cuda.tests.test_basic_ops.test_shared_options)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gokul/Theano/theano/tensor/tests/test_sharedvar.py", line 303, in test_shared_do_alias
    x /= values_to_div
TypeError: unsupported operand type(s) for /=: 'CudaNdarray' and 'CudaNdarray'

======================================================================
ERROR: test_shared_dont_alias (theano.sandbox.cuda.tests.test_basic_ops.test_shared_options)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gokul/Theano/theano/tensor/tests/test_sharedvar.py", line 103, in test_shared_dont_alias
    x /= values_to_div
TypeError: unsupported operand type(s) for /=: 'CudaNdarray' and 'CudaNdarray'

======================================================================
ERROR: test_return_internal_type (theano.sandbox.cuda.tests.test_basic_ops.test_shared_options)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gokul/Theano/theano/tensor/tests/test_sharedvar.py", line 197, in test_return_internal_type
    x /= values_to_div  # supported by ndarray and CudaNdarray
TypeError: unsupported operand type(s) for /=: 'CudaNdarray' and 'CudaNdarray'

======================================================================
ERROR: test_set_value (theano.sandbox.cuda.tests.test_basic_ops.test_shared_options)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gokul/Theano/theano/tensor/tests/test_sharedvar.py", line 270, in test_set_value
    get_x /= values_to_div  # supported by ndarray and CudaNdarray
[...]

PR #4330, opened 2016-04-03
fix gh-4328 This was done for python3, but not python2.

nouiz:
fix gh-4328

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1640,8 +1640,6 @@ def local_fill_sink(node):
             if not r:
                 continue
             replacements.update(r)
-    for v in replacements.values():
-        v.tag.values_eq_approx = values_eq_approx_remove_nan
     return replacements
 
 register_canonicalize(local_fill_sink)
@@ -3851,11 +3849,9 @@ def local_merge_switch_same_cond(node):
     if not all(s.owner.inputs[0] is cond for s in node.inputs[1:]):
         return
     # pull out switch
-    ret = T.switch(cond,
-                   node.op(*[s.owner.inputs[1] for s in node.inputs]),
-                   node.op(*[s.owner.inputs[2] for s in node.inputs]))
-    ret.tag.values_eq_approx = values_eq_approx_remove_nan
-    return [ret]
+    return [T.switch(cond,
+                     node.op(*[s.owner.inputs[1] for s in node.inputs]),
+                     node.op(*[s.owner.inputs[2] for s in node.inputs]))]
 
 
 #############
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1640,6 +1640,8 @@ def local_fill_sink(node):
             if not r:
                 continue
             replacements.update(r)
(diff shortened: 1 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -1640,6 +1640,8 @@ def local_fill_sink(node):
             if not r:
                 continue
             replacements.update(r)
+    for v in replacements.values():
+        v.tag.values_eq_approx = values_eq_approx_remove_nan
```
abergeron: Why should we ignore nans in all cases here.

theano/tensor/opt.py
```diff
@@ -3849,9 +3851,11 @@ def local_merge_switch_same_cond(node):
     if not all(s.owner.inputs[0] is cond for s in node.inputs[1:]):
         return
     # pull out switch
-    return [T.switch(cond,
-                     node.op(*[s.owner.inputs[1] for s in node.inputs]),
-                     node.op(*[s.owner.inputs[2] for s in node.inputs]))]
+    ret = T.switch(cond,
+                   node.op(*[s.owner.inputs[1] for s in node.inputs]),
+                   node.op(*[s.owner.inputs[2] for s in node.inputs]))
+    ret.tag.values_eq_approx = values_eq_approx_remove_nan
```
abergeron: Again why do we ignore nans here?

theano/tensor/opt.py
```diff
@@ -5304,7 +5308,9 @@ def local_zero_div(node):
     if isinstance(node.op, T.Elemwise) and isinstance(
             node.op.scalar_op, (theano.scalar.IntDiv, theano.scalar.TrueDiv)):
         if local_mul_canonizer.get_constant(node.inputs[0]) == 0:
-            return [broadcast_like(0, node.outputs[0], node.fgraph)]
+            ret = broadcast_like(0, node.outputs[0], node.fgraph)
+            ret.tag.values_eq_approx = values_eq_approx_remove_nan
```
abergeron: Again ...

theano/tensor/opt.py
nouiz: I didn't understand well, but all those optimizations are removing nan and so without this, it make the daily buildbot fail.

lamblin: I'm pretty sure this optimization should not remove NaNs.

theano/tensor/opt.py
lamblin: Not this one either.

theano/tensor/opt.py
lamblin: I agree for that one: it is replacing `0/x` by `0`, so when evaluated at `x == 0`, a NaN is removed.
I think fixing this one should be enough to fix the test, and the ones in `local_merge_switch_same_cond`, and probably in `local_fill_sink`, should be removed.

nouiz:
I just updated this PR by removing the change @lamblin didn't agree. This still change one opt, maybe it was the root cause. No time to make sure it fix all the error I wanted to fix in the daily buildot.

I think we can merge this PR.

Issue #4085, opened 2016-02-17
Metadocumentation should request to use Numpy's docstring format


fvisin:
[This](http://deeplearning.net/software/theano/extending/extending_theano.html#documentation) should be fixed.

lamblin:
Also, go through the examples of docstrings in the documentation (in `theano/doc/`, the RST files), and update them if they do not use NumPy format.

fvisin:
@taesupkim you can probably use the email I sent you for the previous ticket as starting point for rewriting the documentation.

PR #4189, opened 2016-03-07
Check RST files in '/theano/doc/' to follow numpy's docstring

taesupkim:
Related to issue #4085.
Modified a couple of docstrings only already exist.

PR #4335, opened 2016-04-04
Check RST files in '/theano/doc/' to follow numpy's docstring + Fix elemwise

fvisin:
This PR integrates #4189 and closes issue #4085 .

### DIFF
--- a/doc/dev_start_guide.txt
+++ b/doc/dev_start_guide.txt
@@ -219,27 +219,11 @@ GitHub, to let them know that you have submitted a fix.
 Tips for Quality Contributions
 ==============================
 
-Coding Style Auto Check
+Coding Style
 -----------------------
 
-In Theano, we use the same coding style as the `Pylearn
-<http://deeplearning.net/software/pylearn/v2_planning/API_coding_style.html>`_
-project, except that we don't use the numpy docstring standard.
-The principal thing to know is that we follow the
-`PEP 8 <http://www.python.org/dev/peps/pep-0008/>`_ coding style.
-
-We use git hooks provided in the project `pygithooks
-<https://github.com/lumberlabs/pygithooks>`_ to validate that commits
-respect pep8. This happens when each user commits, not when we
-push/merge to the Theano repository. Github doesn't allow us to have
-code executed when we push to the repository. So we ask all
-contributors to use those hooks.
-
-For historic reason, we currently don't have all files respecting pep8.
-We decided to fix everything incrementally. So not all files respect it
-now. So we strongly suggest that you use the "increment" pygithooks
-config option to have a good workflow. See the pygithooks main page
-for how to set it up for Theano and how to enable this option.
+Please always respect the :ref:`Documentation`
+or your contribution will not be accepted
 
 
 Setting up your Editor for PEP8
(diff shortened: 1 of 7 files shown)


doc/extending/extending_theano.txt
```diff
@@ -876,11 +876,24 @@ Modify and execute the example to return two outputs: x + y
...
--------------
+.. _Documentation:
 
-See :ref:`metadocumentation`, for some information on how to generate
-the documentation.
+Documentation and Coding Style
+------------------------------
+
+* The code should be compatible with Python 2.6 and above, as well as Python 3.3 and above (using 2to3 for conversion
+  to Python 3.x).
```
lamblin: Actually, we are not using `2to3` any more, we need the code to be natively compatible with Python 2 and Python 3, using `six` if needed.

fvisin:
I should be done with this. If Travis passes please check if everything makes sense.

doc/dev_start_guide.txt
```diff
@@ -259,49 +244,49 @@ Detection of warnings and errors is done by the `pep8`_ script
...
 
-    Add this to your vimrc::
+    git clone https://github.com/VundleVim/Vundle.vim.git ~/.vim/bundle/Vundle.vim
 
-        call pathogen#infect()
+#. Edit ``~/.vimrc`` and add the lines::
 
-    Now any plugins you wish to install can be extracted to a subdirectory
-    under ``~/.vim/bundle``, and they will be added to the ``'runtimepath'``.
+    .. code-block:: python
```
lamblin: That's not python syntax, so maybe just use "txt"?

doc/extending/extending_theano.txt
```diff
@@ -876,11 +876,24 @@ Modify and execute the example to return two outputs: x + y
...
+* All the code should respect the `PEP8 Code Style Guide <http://www.python.org/dev/peps/pep-0008>`_.
+
+* The docstrings of all the classes and functions should respect the
+  `PEP257 <https://www.python.org/dev/peps/pep-0257/>`_ rules and follow the `Numpy docstring standard
+  <https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt>`_.
+
+* To cross-reference other objects (e.g. reference other classes or methods) in the docstrings, use the
+  `cross-referencing objects <http://www.sphinx-doc.org/en/stable/domains.html#cross-referencing-python-objects>`_ syntax. ``:py`` can be omitted, see e.g. this `stackoverflow answer <http://stackoverflow.com/a/7754189>`_.
+
+* See :ref:`metadocumentation`, for some information on how to generate the documentation.
```
lamblin: Would it make sense to add that pep8 compliance is tested by Travis on PRs, and that there are still non-compliant files in a whitelist? Or is that too much information?

doc/extending/unittest.txt
```diff
@@ -476,8 +476,10 @@ Here is an example showing how to use ``verify_grad`` on an Op instance:
 .. testcode::
 
     def test_flatten_outdimNone():
-        # Testing gradient w.r.t. all inputs of an op (in this example the op
-        # being used is Flatten(), which takes a single input).
+        """
+        Testing gradient w.r.t. all inputs of an op (in this example the op
+        being used is Flatten(), which takes a single input).
+        """
```
lamblin: This one is actually a bad idea, since it makes the test name appears as `Testing gradient w.r.t. all inputs of an op (in this example the op` rather than `test_flatten_outdimNone`, which is why we usually keep that in a comment for unit tests.

theano/tensor/elemwise.py
```diff
@@ -343,11 +353,13 @@ def statements(lst):
...
+                "if (strides[%(i)s] == 0) strides[%(i)s] = strides[%(i)s+1] * "
+                "dimensions[%(i)s+1]" % dict(i=str(i)))
 
         #
-        # PyObject* PyArray_New(PyTypeObject* subtype, int nd, npy_intp* dims, int type_num,
-        #                       npy_intp* strides, void* data, int itemsize, int flags, PyObject* obj)
+        # PyObject* PyArray_New(PyTypeObject* subtype, int nd, npy_intp* dims,
+        #                       int type_num, npy_intp* strides, void* data,
+        #                       int itemsize, int flags, PyObject* obj)
         #
```
lamblin: I think you can remove that rather than re-indenting it.

lamblin:
Otherwise, I'm OK when Travis passes.

doc/extending/unittest.txt
fvisin: This would be better IMO, 
```python
"""
test_flatten_outdimNone

Testing gradient w.r.t. all inputs of an op (in this example the op 
being used is Flatten(), which takes a single input).
"""
```
but I'll stick to the convention that has been used so far.

doc/dev_start_guide.txt
fvisin: You are right, byt some highlighting looked better than no highlighting :)

doc/extending/extending_theano.txt
fvisin: I moved things around as we discussed IRL yesterday. Everything is in the *dev_start_guide* now and the *extending_theano* section refers to it. I also changed the order of the information to prevent people from skipping it.

doc/extending/unittest.txt
lamblin: That would not give the full path, and having to put the full path each time is tedious, and hard to maintain if we move tests...

fvisin:
@lamblin travis finally passes. Can you please give it a last check and merge if ok?

PR #4336, opened 2016-04-05
Pass dims as intp to PyArray_SimpleNew

lamblin:
Should fix the bug reported [here](https://groups.google.com/d/topic/theano-users/cqdsmr6mfDg/discussion).

I'm not sure there is something to do for the GPU back-ends, I'm not sure modern versions of CUDA run on 32-bit architectures.


theano/sandbox/rng_mrg.py
```diff
@@ -441,8 +444,9 @@ def c_code(self, node, name, inp, out, sub):
 
         for (int i = 0; i < %(ndim)s; ++i)
         {
-            odims[i] = *(dtype_%(size)s *)PyArray_GETPTR1(%(size)s, i);
-            n_elements *= odims[i];
+            odims_i = *(dtype_%(size)s *)PyArray_GETPTR1(%(size)s, i);
+            odims[i] = odims_i;
```
nouiz: should we test that odims_i fix in  npy_intp? I'm not sure how useful it is to check that.

Otherwise, I'm ok to merge after travis passed if you an all mrg tests.

lamblin: If `odims_i` does not fit in an `npy_intp`, then `n_elements` will be greater than `2**31 - 1`, so there will be an error.

lamblin:
I only changed the CPU C code, so that should be tested by Travis, right?

nouiz:
travis is failing as this test it too long:

theano.sandbox.tests.test_multinomial.test_n_samples_2 ... 

I restarted and it also failed like this. I just restarted it, if it fail again, probably there is a problem.

nouiz:
Now travis passed. So I merge.

PR #4340, opened 2016-04-05
Add broadcastable dimensions in opt where needed

lamblin:
Should fix https://github.com/fchollet/keras/issues/2187


nouiz:
ok to merge when travis pass. If you can add a test, it would be ideal, but I think it would be hard.

Issue #2801, opened 2015-04-20
opt: subtensor(incsubtensor)


nouiz:
Low prio as this is currently for graph clean up.

gh-2545 optimized one case, but for the grad, the graph is ugly. It probably don't make a big speed difference in that particular case. The graph:

> > > import theano, numpy
> > > x = theano.tensor.matrix('x')
> > > splits = theano.tensor.ivector('splits')
> > > y = theano.tensor.split(x, splits, n_splits=1).sum()
> > > f = theano.function([x, splits], theano.grad(y, x))  
> > > theano.printing.debugprint(f)

output

```
Alloc [@A] <TensorType(float64, matrix)> ''   7
 |TensorConstant{(1, 1) of 1.0} [@B] <TensorType(float64, (True, True))>
 |Subtensor{int64} [@C] <TensorType(int64, scalar)> ''   6
 | |IncSubtensor{InplaceSet;int8} [@D] <TensorType(int64, vector)> ''   4
 | | |MakeVector [@E] <TensorType(int64, vector)> ''   3
 | | | |Shape_i{0} [@F] <TensorType(int64, scalar)> ''   2
 | | | | |x [@G] <TensorType(float64, matrix)>
 | | | |Shape_i{1} [@H] <TensorType(int64, scalar)> ''   1
 | | |   |x [@G] <TensorType(float64, matrix)>
 | | |Subtensor{int64} [@I] <TensorType(int32, scalar)> ''   0
 | | | |splits [@J] <TensorType(int32, vector)>
 | | | |Constant{0} [@K] <int64>
 | | |Constant{0} [@L] <int8>
 | |Constant{0} [@K] <int64>
 |Subtensor{int64} [@M] <TensorType(int64, scalar)> ''   5
   |IncSubtensor{InplaceSet;int8} [@D] <TensorType(int64, vector)> ''   4
   |Constant{1} [@N] <int64>
```

The node @ D can be removed from the graph.

lamblin:
The earlier test case does not produce that graph anymore. Here is another test:
```python
>>> x = theano.tensor.matrix('x')
>>> i = theano.tensor.iscalar('i')
>>> v = theano.tensor.vector('v')
>>> y = theano.tensor.set_subtensor(x[i], v)
>>> z = y[i]
>>> f = f = theano.function([x, i, v], z)
>>> theano.printing.debugprint(f)
Subtensor{int32} [id A] ''   2
 |IncSubtensor{Set;int32} [id B] ''   1
[...]

PR #4355, opened 2016-04-11
#2801:subtensor-incsubtensor

Saizheng:
ticket 2801, first commit, more test to be done.https://github.com/Theano/Theano/issues/2801

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1895,7 +1895,7 @@ def local_track_shape_i(node):
         return [shape_feature.shape_of[replacement][node.op.i]]
 
 
-# Subtensor(SetSubtensor(x, y, idx), idx) -> y 
+# Subtensor(SetSubtensor(x, y, idx), idx) -> y
 @register_specialize
 @register_canonicalize
 @gof.local_optimizer([Subtensor])
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1895,11 +1895,14 @@ def local_track_shape_i(node):
         return [shape_feature.shape_of[replacement][node.op.i]]
 
 
-# Subtensor(SetSubtensor(x, y, idx), idx) -> y
 @register_specialize
 @register_canonicalize
 @gof.local_optimizer([Subtensor])
 def local_subtensor_inc_subtensor(node):
+    """
+    Subtensor(SetSubtensor(x, y, idx), idx) -> y
+
+    """
     if isinstance(node.op, Subtensor):
         x = node.inputs[0]
         if not x.owner or not isinstance(x.owner.op, IncSubtensor):
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1895,7 +1895,7 @@ def local_track_shape_i(node):
         return [shape_feature.shape_of[replacement][node.op.i]]
 
 
-# TH-2801 opt: subtensor(incsubtensor)
+# Subtensor(SetSubtensor(x, y, idx), idx) -> y 
 @register_specialize
 @register_canonicalize
 @gof.local_optimizer([Subtensor])
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1894,7 +1894,8 @@ def local_track_shape_i(node):
         replacement = shape_feature.scheduled[node]
         return [shape_feature.shape_of[replacement][node.op.i]]
 
-#TH-2801 opt: subtensor(incsubtensor)
+
+# TH-2801 opt: subtensor(incsubtensor)
 @register_specialize
 @register_canonicalize
 @gof.local_optimizer([Subtensor])
@@ -1905,10 +1906,12 @@ def local_subtensor_inc_subtensor(node):
             return
         if not x.owner.op.set_instead_of_inc:
             return
-        if x.owner.inputs[2] == node.inputs[1] and tuple(x.owner.op.idx_list) == tuple(node.op.idx_list):
+
+        if x.owner.inputs[2:] == node.inputs[1:] and tuple(x.owner.op.idx_list) == tuple(node.op.idx_list):
             return [x.owner.inputs[1]]
         else:
             return
+
 
 @register_specialize
 @register_canonicalize
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1911,7 +1911,13 @@ def local_subtensor_inc_subtensor(node):
             return
 
         if x.owner.inputs[2:] == node.inputs[1:] and tuple(x.owner.op.idx_list) == tuple(node.op.idx_list):
-            return [x.owner.inputs[1]]
+            # if x[idx] and y have the same ndim (and shape), directly return y
+            if x.owner.inputs[0].ndim-(len(node.op.idx_list)-sum([isinstance(idx, slice) for idx in node.op.idx_list])) == x.owner.inputs[1].ndim:
+                return [x.owner.inputs[1]]
+            # else y is broadcastable, return alloc of broadcastable y
+            else:
+                x_subtensor = Subtensor(node.op.idx_list)(x.owner.inputs[0], *x.owner.inputs[2:])
+                return [T.alloc(x.owner.inputs[1], *x_subtensor.shape)]
         else:
             return
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1912,7 +1912,7 @@ def local_subtensor_inc_subtensor(node):
 
         if x.owner.inputs[2:] == node.inputs[1:] and tuple(x.owner.op.idx_list) == tuple(node.op.idx_list):
             # if x[idx] and y have the same ndim (and shape), directly return y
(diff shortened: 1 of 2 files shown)


ballasn:
Travis is failling

theano/tensor/tests/test_opt.py
```diff
@@ -1811,6 +1811,18 @@ def test_local_useless_subtensor():
...
+class test_subtensor_inc_subtensor(unittest.TestCase):
+    def test1(self):
+        x = tensor.matrix('x')
+        i = tensor.iscalar('i')
+        v = tensor.vector('v')
+        y = tensor.set_subtensor(x[i], v)
+        z = y[i]
+        f = theano.function([x, i, v], z)
+        prog = f.maker.fgraph.toposort()
+        assert len(prog) == 1
```
nouiz: execute the function and execute it and check the result, just to be sure it give the right answer.

Saizheng: It works when I used nosetests at local computer, no error reported. Still have not figured out the problem. @ballasn

ballasn: it seems that you have some flake8 issues, you can use theano/tests/test_flake8.py to test that locally

Saizheng: Right, the first is flake8 which is a minor fix, what I mean is the last two falied test which is an assert error.

ballasn:
It looks good to me, @nouiz  could you have another look to see if we can merge it?

theano/tensor/opt.py
```diff
@@ -1782,6 +1782,24 @@ def local_track_shape_i(node):
         return [shape_feature.shape_of[replacement][node.op.i]]
 
 
+# TH-2801 opt: subtensor(incsubtensor)
```
nouiz: remove that comment

theano/tensor/opt.py
```diff
@@ -1782,6 +1782,24 @@ def local_track_shape_i(node):
...
+@register_specialize
+@register_canonicalize
+@gof.local_optimizer([Subtensor])
+def local_subtensor_inc_subtensor(node):
```
nouiz: add a comment like
 Subtensor(SetSubtensor(x, y, idx), idx) -> y

theano/tensor/tests/test_opt.py
```diff
@@ -1838,6 +1838,53 @@ def test_local_useless_subtensor():
...
+    i4 = tensor.iscalar('i4')
+    v = tensor.tensor3('v')
+    y = tensor.set_subtensor(x[i1, :i2, i3:, ::i4], v)
+    z = y[i1, :i2, i3:, ::i4]
+    mode = theano.compile.mode.get_mode('FAST_COMPILE').including('local_subtensor_inc_subtensor')
+    f = theano.function([x, i1, i2, i3, i4, v], z, mode=mode)
+    prog = f.maker.fgraph.toposort()
+    assert len(prog) == 1
+    assert all(not isinstance(x.op, tensor.IncSubtensor) for x in prog)
+    assert (not isinstance(x.op, tensor.Subtensor) for x in prog)
```
nouiz: this assert is badly written. You want to assert that x.op is Subtensor.

theano/tensor/tests/test_opt.py
```diff
@@ -1838,6 +1838,53 @@ def test_local_useless_subtensor():
...
+    i = tensor.iscalar('i')
+    v = tensor.vector('v')
+    y = tensor.set_subtensor(x[i], v)
+    z = y[i]
+    mode = theano.compile.mode.get_default_mode().including('local_subtensor_inc_subtensor')
+    f = theano.function([x, i, v], z, mode=mode)
+    prog = f.maker.fgraph.toposort()
+    assert len(prog) == 1
+    assert all(not isinstance(x, tensor.IncSubtensor) for x in prog)
+    assert (not isinstance(x, tensor.IncSubtensor) for x in prog)
```
nouiz: assert badly written.

theano/tensor/tests/test_opt.py
```diff
@@ -1838,6 +1838,53 @@ def test_local_useless_subtensor():
...
+    z = y[i1, :i2, i3:, ::i4]
+    mode = theano.compile.mode.get_mode('FAST_COMPILE').including('local_subtensor_inc_subtensor')
+    f = theano.function([x, i1, i2, i3, i4, v], z, mode=mode)
+    prog = f.maker.fgraph.toposort()
+    assert len(prog) == 1
+    assert all(not isinstance(x.op, tensor.IncSubtensor) for x in prog)
+    assert (not isinstance(x.op, tensor.Subtensor) for x in prog)
+
+    # case not use this optimization
+    x = tensor.tensor4('x')
```
nouiz: don't recreate the variables. This isn't useful.

nouiz:
It seem good except for a few small comment about the test.

ballasn:
@Saizheng could you update the PR and address @nouiz comments,

Once this is done, we can merge this PR,

theano/tensor/opt.py
```diff
@@ -1851,6 +1851,24 @@ def local_track_shape_i(node):
         return [shape_feature.shape_of[replacement][node.op.i]]
 
 
+# Subtensor(SetSubtensor(x, y, idx), idx) -> y
```
nouiz: we put this as a docstring normally.

theano/tensor/tests/test_opt.py
```diff
@@ -1928,6 +1928,44 @@ def test_local_subtensor_remove_broadcastable_index():
...
+    # complicated test
+    x = tensor.tensor4('x')
+    i1 = tensor.iscalar('i1')
+    i2 = tensor.iscalar('i2')
+    i3 = tensor.iscalar('i3')
+    i4 = tensor.iscalar('i4')
+    v = tensor.tensor3('v')
+    y = tensor.set_subtensor(x[i1, :i2, i3:, ::i4], v)
+    z = y[i1, :i2, i3:, ::i4]
+    mode = theano.compile.mode.get_mode('FAST_COMPILE').including('local_subtensor_inc_subtensor')
```
nouiz: use "get_default_mode()" instead of "get_mode('FAST_COMPILE')". Otherwise, this would bypass DebugMode. idem bellow.

Otherwise with my other comment, it is good. thanks.

nouiz:
jenkins test this

theano/tensor/tests/test_opt.py
```diff
@@ -1941,6 +1941,44 @@ def test_local_subtensor_remove_broadcastable_index():
     f2(xn)
 
 
+def test_subtensor_inc_subtensor():
```
lamblin: Can you add calls to `f(...)` with actual numerical values, and then run these tests with `mode=DebugMode`?
I don't think the "complex case" is correct.
Moreover, we would need tests for the case where an `alloc` is returned.

theano/tensor/opt.py
```diff
@@ -1923,6 +1923,33 @@ def local_track_shape_i(node):
...
+        if not x.owner.op.set_instead_of_inc:
+            return
+
+        if x.owner.inputs[2:] == node.inputs[1:] and tuple(x.owner.op.idx_list) == tuple(node.op.idx_list):
+            # if x[idx] and y have the same ndim (and shape), directly return y
+            if x.owner.inputs[0].ndim - (len(node.op.idx_list) - sum([isinstance(idx, slice) for idx in node.op.idx_list])) == x.owner.inputs[1].ndim:
+                return [x.owner.inputs[1]]
+            # else y is broadcastable, return alloc of broadcastable y
+            else:
+                x_subtensor = Subtensor(node.op.idx_list)(x.owner.inputs[0], *x.owner.inputs[2:])
```
nouiz: Subtensor(node.op.idx_list) -> node.op ?

theano/tensor/tests/test_opt.py
lamblin: I don't remember exactly, I guess I thought `v` was a vector, not a tensor3.
In any case, running the function with values so that DebugMode can check the correctness of the optimization should be necessary.

theano/tensor/tests/test_opt.py
```diff
@@ -1997,10 +1997,10 @@ def test_subtensor_inc_subtensor():
     mode = theano.compile.mode.get_default_mode().including('local_subtensor_inc_subtensor')
     f = theano.function([x, i1, i2, v], z, mode=mode)
     prog = f.maker.fgraph.toposort()
-    assert any(isinstance(x.op, tensor.Alloc) for x in prog)
+    assert any(isinstance(x.op, (tensor.Alloc, theano.sandbox.cuda.basic_ops.GpuAlloc)) for x in prog)
```
nouiz: remove GpuAlloc. The CPU tests should not request the gpu back-end. This would also not work with the new back-end.

I think the problem is that  you had device⁼gpu* Tests suppose device=cpu. So don't try to support device!=cpu, too many tests already fail in that case. Also, currently I think this would fail when device=cpu, as theano.sandbox.cuda isn't imported.

If for some reason, you still need to keep GpuAlloc, add import theano.sandbox.cuda before using this, not at the top of the file.

theano/tensor/opt.py
nouiz: You forgot this small comment. This will make this optimization faster by building less python object. Otherwise, this PR is ready to merge. thanks

Issue #4358, opened 2016-04-11
PersistentNdarrayLoad in pkl_utils.py doesn't process same object properly


tete1030:
As mentioned in #4238 and mila-udem/blocks#1031  , I encountered the same problem when I use function __theano.misc.pkl_utils.load__ with default __persistent\_id__ parameter, __PersistentNdarrayLoad__.

If __PersistentNdarrayLoad.\_\_call\_\___ receives two same __persid__, it will return two different objects with same value instead of one single object, which causes __AssertionError__ in __\_constructor\_Function__ of __theano.compile.function\_module__.

The issues of _blocks_ mentioned above have been fixed by caching created object as https://github.com/mila-udem/blocks/pull/1032/commits/5c04f77b71a7c95182645e1cfa20299fb883f802

Besides, there is a little problem at https://github.com/Theano/Theano/blob/master/theano/compile/sharedvalue.py#L79
**container** should be **value**, **container** has already been checked.

And I found in https://github.com/Theano/Theano/blob/master/theano/misc/pkl_utils.py#L263
Why is it storing object name with key __id(obj.container.storage[0])__ but retrieving object name with key __id(obj)__ in __\_resolve\_name__ ? I'm not sure if it is correct.

Thank you

tete1030:
I was wrong about the code at Line 79 in _sharedvalue.py_. It seems like a duplicate check of **container**.

lamblin:
@tete1030 Can you confirm if #4362 fixed the issue for you?

tete1030:
I can confirm that #4362 has fixed the issue about PersistentNdarrayLoad. But I don't know if the remaining two problems is okay.

nouiz:
Which other 2 problems do you talk about?

nouiz:
I'm closing as this issue is fixed.

PR #4362, opened 2016-04-12
Fix PersistentNdarrayLoad problem restoring same object multiple times

tete1030:
Fix problems mentioned in issue #4358

### DIFF
--- a/theano/misc/pkl_utils.py
+++ b/theano/misc/pkl_utils.py
@@ -275,10 +275,14 @@ class PersistentNdarrayLoad(object):
     """
     def __init__(self, zip_file):
         self.zip_file = zip_file
+        self.cache = {}
 
     def __call__(self, persid):
         array_type, name = persid.split('.')
 
+        if name in self.cache:
+            return self.cache[name]
+        ret = None
         array = numpy.lib.format.read_array(self.zip_file.open(name))
         if array_type == 'cuda_ndarray':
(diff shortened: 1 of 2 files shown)


tete1030:
https://github.com/Theano/Theano/issues/4358#issuecomment-208849467

theano/misc/pkl_utils.py
```diff
@@ -275,25 +275,31 @@ class PersistentNdarrayLoad(object):
     """
     def __init__(self, zip_file):
         self.zip_file = zip_file
+        self.cache = {}
 
     def __call__(self, persid):
         array_type, name = persid.split('.')
 
+        if name in self.cache:
+            return self.cache[name]
```
nouiz: why not use persid as the key of the cache? I think it is more safe. I think you can have those 2 persid that would conflict with the current code:

ndarray.array_0
cuda_ndarray.array_0

Thanks for investigating and founding the root of the problem. I suppose it wasn't trivial to find.

tete1030: Because the codes are using **name** instead of **type.name** as filename to store data in **zip_file**. You can see them at Line 195 and Line 208. A **persid** with same **name** must comes from same file named by **name**. So I use **name** instead of persid in case there are two different persid with same name, though maybe it is impossible.

nouiz:
great. Seem good and tests are passing, so merging.

thanks.

PR #4374, opened 2016-04-14
Changed release doc to say new releases several times a year, instead…

slefrancois:
… of monthly

### DIFF
--- a/doc/internal/release.txt
+++ b/doc/internal/release.txt
@@ -15,12 +15,10 @@ later).
 The Theano release plan is detailed below. Comments and/or suggestions are
 welcome on the mailing list.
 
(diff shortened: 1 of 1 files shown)


doc/internal/release.txt
```diff
@@ -43,7 +41,7 @@ welcome on the mailing list.
 
 4) The release numbers will follow the X.Y.Z scheme:
 
-   1) We update Z by 1 for each lightweight release.
```
nouiz: That part is outdated.

Now we update Z for small urgent bug or support new version of dependency.
We update Y for interface change and or significant feature.

slefrancois:
Updated release number policy.

PR #4375, opened 2016-04-14
Fix crashes probably due to nvcc not handling unsigned int as we expect

nouiz:
This could cause bad value at other time. It currently crash when sA0 is negative. If you understand why this fail, it would be great to know. Maybe we should make ths work around elsewhere.

The new GPU back-end use this code and don't have this problem. But it use g++ to compile instead of nvcc.


lamblin:
It's not the first time we had to fix something similar, where we have to cast `threadIdx.x` or similar variables from `unsigned` to `signed` int, because otherwise the result of the multiplication with a negative stride would give unexpected result.

PR #4290, opened 2016-03-23
[WIP] Helper function that check stack traces

adbrebs:
#4114 @lamblin

Issue #4114, opened 2016-02-24
Helper function checking for stack traces


lamblin:
Sub-ticket of #3018.
For tests, we would need a helper function to check that some variables in an optimized graph have a stack trace. Some things we should be able to specify:
- check the outputs
- check all variables
- check all variables where the owner's op is in instance of some classes

This would be better than only checking the output.

lamblin:
Closed by #4290.

PR #4376, opened 2016-04-14
[WIP] Use check_stack_trace helper function in tensor/nnet/tests/

adbrebs:
It needs #4290 to be merged first. I will do a rebase once done.
I have only modified the asserts that were already written, I have not added any.
There are several cases I am not sure about, for which I use ops_to_check='all'.

Related ticket: #4114 @lamblin.

### DIFF
--- a/theano/tensor/nnet/sigm.py
+++ b/theano/tensor/nnet/sigm.py
@@ -602,6 +602,7 @@ def local_exp_over_1_plus_exp(node):
             else:
                 # case: 1/(1+exp(x))
                 sigmoids.append(sigmoid(-t))
+            copy_stack_trace(node.outputs[0], sigmoids[-1])
 
         if not sigmoids:  # we didn't find any.  abort
             return
@@ -615,12 +616,17 @@ def local_exp_over_1_plus_exp(node):
         if num_neg ^ denom_neg:
             new_num = -new_num
 
+        copy_stack_trace(num, new_num)
+
         if len(denom_rest) == 0:
             return [new_num]
         elif len(denom_rest) == 1:
-            return [new_num / denom_rest[0]]
+            out = new_num / denom_rest[0]
         else:
-            return [new_num / tensor.mul(*denom_rest)]
+            out = new_num / tensor.mul(*denom_rest)
+
+        copy_stack_trace(node.outputs[0], out)
+        return [out]
 
 
 def parse_mul_tree(root):
@@ -913,6 +919,7 @@ def local_sigm_times_exp(node):
     exp(x) * sigm(-x) -> sigm(x)
     exp(-x) * sigm(x) -> sigm(-x)
 
+    todo: add stack traces to the intermediate variables
     """
     # Bail early if it is not a multiplication.
     if node.op != tensor.mul:
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -134,9 +134,9 @@ class BaseTestConv2d(unittest.TestCase):
         if target_op is not None:
             assert any([isinstance(n.op, target_op) for n
                         in f.maker.fgraph.toposort()])
+            if check_trace:
+                self.assertTrue(check_stack_trace(f, ops_to_check=target_op))
 
-        if check_trace:
-            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
         utt.assert_allclose(res_ref, res)
@@ -179,13 +179,13 @@ class BaseTestConv2d(unittest.TestCase):
                     subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        if check_trace:
-            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         f_ref = theano.function([], c_ref, mode='FAST_RUN')
 
         if target_op is not None:
             assert any([isinstance(n.op, target_op) for n
                         in f.maker.fgraph.toposort()])
+            if check_trace:
+                self.assertTrue(check_stack_trace(f, ops_to_check=target_op))
 
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
@@ -230,13 +230,13 @@ class BaseTestConv2d(unittest.TestCase):
                     border_mode=border_mode, subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        if check_trace:
-            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         f_ref = theano.function([], c_ref, mode='FAST_RUN')
 
         if target_op is not None:
             assert any([isinstance(n.op, target_op) for n
                         in f.maker.fgraph.toposort()])
+            if check_trace:
+                self.assertTrue(check_stack_trace(f, ops_to_check=target_op))
 
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -99,7 +99,7 @@ class BaseTestConv2d(unittest.TestCase):
     def run_fwd(self, inputs_shape, filters_shape, ref=conv_corr,
                 subsample=(1, 1), verify_grad=True, mode=None,
                 border_mode='valid', filter_flip=True, provide_shape=False,
-                target_op=None):
+                target_op=None, check_trace=False):
         inputs_val = numpy.random.random(inputs_shape).astype('float32')
         filters_val = numpy.random.random(filters_shape).astype('float32')
 
@@ -135,7 +135,8 @@ class BaseTestConv2d(unittest.TestCase):
             assert any([isinstance(n.op, target_op) for n
                         in f.maker.fgraph.toposort()])
 
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        if check_trace:
+            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
         utt.assert_allclose(res_ref, res)
@@ -149,7 +150,7 @@ class BaseTestConv2d(unittest.TestCase):
     def run_gradweight(self, inputs_shape, filters_shape, output_shape,
                        ref=conv_corr_gw, subsample=(1, 1), filter_flip=True,
                        verify_grad=True, mode=None, border_mode='valid',
-                       provide_shape=False, target_op=None):
+                       provide_shape=False, target_op=None, check_trace=False):
 
         inputs_val = numpy.random.random(inputs_shape).astype('float32')
         output_val = numpy.random.random(output_shape).astype('float32')
@@ -178,7 +179,8 @@ class BaseTestConv2d(unittest.TestCase):
                     subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        if check_trace:
+            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         f_ref = theano.function([], c_ref, mode='FAST_RUN')
 
         if target_op is not None:
@@ -202,7 +204,7 @@ class BaseTestConv2d(unittest.TestCase):
     def run_gradinput(self, inputs_shape, filters_shape, output_shape,
                       ref=conv_corr_gi, subsample=(1, 1), filter_flip=True,
                       verify_grad=True, mode=None, border_mode='valid',
-                      provide_shape=False, target_op=None):
+                      provide_shape=False, target_op=None, check_trace=False):
 
         output_val = numpy.random.random(output_shape).astype('float32')
         filters_val = numpy.random.random(filters_shape).astype('float32')
@@ -228,7 +230,8 @@ class BaseTestConv2d(unittest.TestCase):
                     border_mode=border_mode, subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        if check_trace:
+            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         f_ref = theano.function([], c_ref, mode='FAST_RUN')
 
         if target_op is not None:
@@ -292,15 +295,18 @@ class TestCorrConv2d(BaseTestConv2d):
             raise SkipTest("Need blas to test conv2d")
         self.run_fwd(inputs_shape=i, filters_shape=f, subsample=s,
                      verify_grad=True, provide_shape=provide_shape,
-                     border_mode=b, filter_flip=flip, target_op=CorrMM)
+                     border_mode=b, filter_flip=flip, target_op=CorrMM,
+                     check_trace=True)
         self.run_gradweight(inputs_shape=i, filters_shape=f,
                             output_shape=o, subsample=s, verify_grad=True,
                             provide_shape=provide_shape, border_mode=b,
-                            filter_flip=flip, target_op=CorrMM_gradWeights)
+                            filter_flip=flip, target_op=CorrMM_gradWeights,
+                            check_trace=True)
         self.run_gradinput(inputs_shape=i, filters_shape=f,
                            output_shape=o, subsample=s, verify_grad=True,
                            provide_shape=provide_shape, border_mode=b,
-                           filter_flip=flip, target_op=CorrMM_gradInputs)
+                           filter_flip=flip, target_op=CorrMM_gradInputs,
+                           check_trace=True)
 
 
 class TestCpuConv2d(BaseTestConv2d):
@@ -344,7 +350,8 @@ class TestCpuConv2d(BaseTestConv2d):
             self.run_fwd(inputs_shape=i, filters_shape=f, subsample=s,
                          verify_grad=(gradweight_OK and gradinput_OK),
                          mode=mode, provide_shape=provide_shape,
-                         border_mode=b, filter_flip=flip, target_op=ConvOp)
+                         border_mode=b, filter_flip=flip, target_op=ConvOp,
+                         check_trace=True)
         else:
             self.assertRaises(AssertionError,
                               self.run_fwd,
@@ -355,7 +362,8 @@ class TestCpuConv2d(BaseTestConv2d):
                               mode=mode,
                               provide_shape=provide_shape,
                               border_mode=b,
-                              filter_flip=flip)
+                              filter_flip=flip,
+                              check_trace=True)
 
         if gradweight_OK:
             if not theano.config.blas.ldflags:
@@ -365,7 +373,8 @@ class TestCpuConv2d(BaseTestConv2d):
                                 verify_grad=False, mode=mode,
                                 provide_shape=provide_shape, border_mode=b,
                                 filter_flip=flip,
-                                target_op=(ConvOp, ConvGrad3D))
+                                target_op=(ConvOp, ConvGrad3D),
+                                check_trace=True)
         else:
             self.assertRaises(AssertionError,
                               self.run_gradweight,
@@ -377,7 +386,8 @@ class TestCpuConv2d(BaseTestConv2d):
                               mode=mode,
                               provide_shape=provide_shape,
                               border_mode=b,
-                              filter_flip=flip)
+                              filter_flip=flip,
+                              check_trace=True)
 
         if gradinput_OK:
             if not theano.config.blas.ldflags:
@@ -387,7 +397,8 @@ class TestCpuConv2d(BaseTestConv2d):
                                verify_grad=False, mode=mode,
                                provide_shape=provide_shape, border_mode=b,
                                filter_flip=flip,
-                               target_op=(ConvOp, ConvTransp3D))
+                               target_op=(ConvOp, ConvTransp3D),
+                               check_trace=True)
         else:
             self.assertRaises(AssertionError,
                               self.run_gradinput,
@@ -399,7 +410,8 @@ class TestCpuConv2d(BaseTestConv2d):
                               mode=mode,
                               provide_shape=provide_shape,
                               border_mode=b,
-                              filter_flip=flip)
+                              filter_flip=flip,
+                              check_trace=True)
 
 
 def test_constant_shapes():
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -7,6 +7,7 @@ from nose.tools import assert_raises
 
 import theano
 from theano import tensor
+from theano.gof.opt import check_stack_trace
 from theano.tests import unittest_tools as utt
 from theano.tensor.nnet import corr, abstract_conv as conv
 from theano.tensor.nnet.abstract_conv import get_conv_output_shape
@@ -134,7 +135,7 @@ class BaseTestConv2d(unittest.TestCase):
             assert any([isinstance(n.op, target_op) for n
                         in f.maker.fgraph.toposort()])
 
-        self.assertTrue(hasattr(f.maker.fgraph.outputs[0].tag, 'trace'))
+        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
         utt.assert_allclose(res_ref, res)
@@ -177,7 +178,7 @@ class BaseTestConv2d(unittest.TestCase):
                     subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        self.assertTrue(hasattr(f.maker.fgraph.outputs[0].tag, 'trace'))
+        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         f_ref = theano.function([], c_ref, mode='FAST_RUN')
 
         if target_op is not None:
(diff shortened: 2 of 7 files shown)


lamblin:
I merged #4290, so this needs to be rebased.
I'm having a look now.

adbrebs:
Thanks, I have just rebased.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -227,7 +228,7 @@ def run_gradinput(self, inputs_shape, filters_shape, output_shape,
                     border_mode=border_mode, subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        self.assertTrue(hasattr(f.maker.fgraph.outputs[0].tag, 'trace'))
+        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
         f_ref = theano.function([], c_ref, mode='FAST_RUN')
 
         if target_op is not None:
```
lamblin: Have you tried running the GPU tests locally? I think some of them inherit from this class, and the check would fail for them. I think they are currently passing because the output node would be HostFromGpu, not the optimized conv op.
I think it would be better to add a "check_trace" argument to `run_fwd`, `run_gradweight`, and `run_gradinput`. You can set it to `False` by default, but to `True` in `TestCorrConv2d` and `TestCpuConv2d` below.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -659,7 +661,7 @@ def test_softmax_grad_optimizations(self):
         fgraph = gof.FunctionGraph(
                 [x, one_of_n],
                 [g_x])
-        self.assertTrue(hasattr(fgraph.outputs[0].tag, 'trace'))
```
lamblin: According to the check below, I think we should check for `crossentropy_softmax_1hot_with_bias_dx` and `softmax_op`.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1295,7 +1309,7 @@ def test_argmax_pushdown():
         fgraph = gof.FunctionGraph(
                 [x],
                 [out])
-        assert hasattr(fgraph.outputs[0].tag, 'trace')
+        assert check_stack_trace(fgraph, ops_to_check='all')
```
lamblin: For that test, I think it is actually the previous `fgraph` that should be checked. We can check that one as well with 'all' since the optimization is not supposed to apply...
For the first `fgraph` above, you can check for `tensor.basic._max_and_argmax`.

lamblin: Also, I counted 9 remaining checks to be fixed below this point.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -328,14 +316,14 @@ def test_local_ultra_fast_sigmoid(self):
```
lamblin: There is a check on l.309 that should be updated or removed, depending on what actually is in the graph, or what optimizations actually get applied.

theano/tensor/nnet/tests/test_abstract_conv.py
adbrebs: No I haven't run the GPU tests, Ok I will do what you suggest.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -306,7 +306,6 @@ def test_grad_log1msigm(self):
         mode = self.get_mode()
         if not isinstance(mode, theano.compile.DebugMode):
             f = theano.function([x, lr], ux, mode=mode)
```
lamblin: You can remove that line as well.

theano/tensor/nnet/tests/test_conv3d2d.py
```diff
@@ -76,7 +76,10 @@ def pyconv3d(signals, filters):
 
 
 def check_diagonal_subtensor_view_traces(fn):
-    assert check_stack_trace(fn, [DiagonalSubtensor, IncDiagonalSubtensor])
+    def check_node_type(node):
+        return isinstance(node.op, (DiagonalSubtensor, IncDiagonalSubtensor))
+    assert check_stack_trace(fn, ops_to_check=check_node_type,
```
lamblin: Is there a reason to explicitly use `isinstance`?
I would find an explicit list/tuple passed as `ops_to_check` more readable, but I won't argue over it :)

theano/gof/opt.py
```diff
@@ -2623,16 +2623,16 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
...
         function returning a boolean and taking as input a theano.gof.Op.
         - if ops_to_check is a string, it should be either 'last' or 'all'.
           'last' will check only the last op of the graph while 'all' will
           check all the ops of the graph.
-        - if ops_to_check is an op or a tuple of ops, the function will check
-          that all the outputs of their occurrences in the graph have a stack
-          trace.
+        - if ops_to_check is an op or a tuple/list of ops, the function will
+          check that all the outputs of their occurrences in the graph have a
+          stack trace.
```
lamblin: Oh, there seems to be a confusion between `Op` classes and instances of those classes.
Ideally I think we should allow both. So the check below (the "default lambda function" if you will) would be something like:
`node.op in ops_to_check or isinstance(node.op, ops_to_check)`.

theano/tensor/nnet/tests/test_opt.py
```diff
@@ -16,12 +17,13 @@ def test_blocksparse_inplace_gemv_opt():
...
 
     f = theano.function([W, h, iIdx, b, oIdx], o)
-    assert check_stack_trace(f, ops_to_check=sparse_block_gemv_inplace)
 
     if theano.config.mode == "FAST_COMPILE":
         assert not f.maker.fgraph.toposort()[-1].op.inplace
+        assert check_stack_trace(f, ops_to_check=[sparse_block_gemv])
     else:
         assert f.maker.fgraph.toposort()[-1].op.inplace
+        assert check_stack_trace(f, ops_to_check=[sparse_block_gemv_inplace])
```
lamblin: Right, thanks.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1382,7 +1380,10 @@ def test_argmax_pushdown_bias():
     assert isinstance(fgraph.toposort()[1].op, tensor.CAReduce)
     assert isinstance(fgraph.toposort()[1].op.scalar_op, theano.scalar.Maximum)
     assert str(fgraph.toposort()[2].op) == 'OutputGuard'
-
+    assert check_stack_trace(
+        fgraph, ops_to_check=lambda node: (
+            isinstance(node, (SoftmaxWithBias, tensor.CAReduce)) or
+            isinstance(node.op.scalar_op, theano.scalar.Maximum)))
```
lamblin: I think that the second line is not needed, as it is already implied by the `isinstance(CAReduce)`. This was an additional check for node 1.

theano/gof/opt.py
adbrebs: I've corrected it, and I also check ``isinstance(node.op.scalar_op, ops_to_check)``.

theano/tensor/nnet/tests/test_sigm.py
adbrebs: which line exactly?

adbrebs:
@lamblin thanks for the review, I think that I've made all the modifications.

theano/tensor/nnet/tests/test_sigm.py
lamblin: I was confused, I thought it was another part of the code.
This is fine like it is.

theano/gof/opt.py
```diff
@@ -2669,7 +2674,9 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     # if ops_to_check is a list/tuple of ops
     elif isinstance(ops_to_check, (tuple, list)):
         apply_nodes_to_check = [node for node in fgraph.apply_nodes
-                                if node.op in ops_to_check]
+                                if node.op in ops_to_check or
+                                isinstance(node.op, ops_to_check) or
+                                isinstance(node.op.scalar_op, ops_to_check)]
```
lamblin: Wait, why does it make sense to check `node.op.scalar_op`?
Also, not all `Op` instances have a `scalar_op` attribute.

theano/gof/opt.py
```diff
@@ -2692,7 +2700,7 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
 
     for node in apply_nodes_to_check:
         for output in node.outputs:
-            if not hasattr(output.tag, 'trace'):
+            if not hasattr(output.tag, 'trace') or output.tag.trace == '':
```
lamblin: The trace can also be an empty list, apparently.

theano/tensor/nnet/tests/test_conv3d2d.py
```diff
@@ -76,10 +76,9 @@ def pyconv3d(signals, filters):
 
 
 def check_diagonal_subtensor_view_traces(fn):
-    def check_node_type(node):
-        return isinstance(node.op, (DiagonalSubtensor, IncDiagonalSubtensor))
-    assert check_stack_trace(fn, ops_to_check=check_node_type,
-                             bug_print='ignore')
+    assert check_stack_trace(
+        fn, ops_to_check=(DiagonalSubtensor, IncDiagonalSubtensor),
+        bug_print='ignore')
```
lamblin: Why 'ignore'? Is there a case where they are not there? Can you add a comment about it?

adbrebs: Yes, there is a case where they are not there. I will add a comment.

theano/gof/opt.py
adbrebs: Fixed, for cases like that https://github.com/Theano/Theano/pull/4376/commits/d4b12260e73cf7001ebae258d2e80bb10331884b#diff-dbe5c9d6bf5794a9f19f73ec4754678bR369

theano/tensor/nnet/tests/test_conv3d2d.py
```diff
@@ -153,7 +152,6 @@ def test_conv3d(mode=mode_without_gpu, shared=theano.tensor._shared):
     newconv3d = theano.function([], [],
                                 updates={s_output: out},
                                 mode=mode)
-    check_diagonal_subtensor_view_traces(newconv3d)
```
adbrebs: In these two cases, the ops (DiagonalSubtensor, IncDiagonalSubtensor) are not in the graph, do you confirm it is normal or should I investigate more?

theano/gof/opt.py
adbrebs: Ive commited the fix in this PR, I can do a separate PR if this is urgent.

theano/tensor/nnet/tests/test_conv3d2d.py
lamblin: From the comment that says "patched implementation ... with Tf=1", I guess it is normal that these Ops are not there, but I'll have a quick look.

theano/gof/opt.py
lamblin: Makes sense, thanks.

theano/gof/opt.py
```diff
@@ -2700,7 +2703,9 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
 
     for node in apply_nodes_to_check:
         for output in node.outputs:
-            if not hasattr(output.tag, 'trace') or output.tag.trace == '':
+            if (not hasattr(output.tag, 'trace') or
+                    output.tag.trace == '' or
+                    not output.tag.trace):
```
lamblin: You can just use `or not output.tag.trace`, it will match `''` as well.

theano/tensor/nnet/tests/test_conv3d2d.py
lamblin: Yes, that seems to be OK, according to https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/conv3d2d.py#L281

theano/gof/opt.py
```diff
@@ -2662,27 +2673,39 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
...
 
-    elif hasattr(ops_to_check, '__call__'):  # if ops_to_check is a function
+    # if ops_to_check is a list/tuple of ops
+    elif isinstance(ops_to_check, (tuple, list)):
         apply_nodes_to_check = [node for node in fgraph.apply_nodes
-                                if ops_to_check(node)]
+                                if node.op in ops_to_check or
+                                isinstance(node.op, ops_to_check) or
+                                (hasattr(node.op, 'scalar_op') and
+                                 isinstance(node.op.scalar_op, ops_to_check))]
```
lamblin: Apparently, you can only have classes or types in the second argument of `isinstance`, so I think you are going need to either separate instances from Ops first, or iterate over the content of `ops_to_check` explicitly for each node.

lamblin:
This broke some tests that seemed to be passing before (but should not have been).
Can you try to see if the tests need updating or if there is something wrong with the optimization?
We will extend the ticket if needed.

adbrebs:
Yes I would definitively need more ccw time to investigate the assert failures. The other errors should be fixed by my last commit.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -227,7 +230,8 @@ def run_gradinput(self, inputs_shape, filters_shape, output_shape,
                     border_mode=border_mode, subsample=subsample,
                     conv_mode=conv_mode)
         f = theano.function([], c, mode=mode)
-        self.assertTrue(hasattr(f.maker.fgraph.outputs[0].tag, 'trace'))
+        if check_trace:
+            self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```
lamblin: There are subtensors (`[:, :, ::-1, ::-1]`) inserted to flip kernels and such, and those may not have an obvious equivalent, so they have no stack trace.
I think it would make sense to put `target_op` in `ops_to_check`, and maybe put that inside the `if target_op is not None:` line below.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -755,7 +759,8 @@ def test_get_rid_of_advanced_indexing_version_of_xent(self):
         for expr in expressions:
             # Verify the optimizer worked on the expressions
             f = theano.function([x, y], expr, mode=mode)
-            self.assertTrue(hasattr(f.maker.fgraph.outputs[0].tag, 'trace'))
+            assert check_stack_trace(
+                f, ops_to_check=crossentropy_softmax_argmax_1hot_with_bias)
```
lamblin: There is a `crossentropy_softmax_argmax_1hot_with_bias` with a stack trace in the graph, but only its first output has a stack trace, the other ones have an empty one.
It kind of makes sense, since only the first output is the replacement for the original variable.

I'm not sure what the right solution would be:
- add the stack trace to all outputs of the op
- select which output to check in `check_stack_trace`

theano/tensor/nnet/tests/test_nnet.py
lamblin: So, that fgraph is not optimized yet at that point, and `local_argmax_pushdown` is not going to be applied here anyway, so you can drop this assert.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -281,7 +283,7 @@ def myfunc(x):
         # We set step to 0.1 because for big values we need a big epsilon
         utt.verify_grad(myfunc, [a], eps=0.1, mode=m)
         f = theano.function([], myfunc(a))
-        self.assertTrue(hasattr(f.maker.fgraph.outputs[0].tag, 'trace'))
+        assert check_stack_trace(f, ops_to_check='last')
```
lamblin: The problem here is that `a` is a constant, so `myfunc(a)` gets constant-folded to `DeepCopyOp(TensorConstant(...))`. And since the constant folding seems to happen before the logsoftmax gets inserted (which is another issue entirely), that constant is full of `-inf`...

I think the solution here is to compile a function from a shared variable (not a constant instead). Something like this seems to work:
```python
sa = theano.shared(a)
f = theano.function([], myfunc(sa))
self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1437,7 +1454,7 @@ def test_basic(self):
         # test that function contains softmax and no div.
         f = theano.function([c], p_y, mode=self.mode)
 
-        assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
+        assert check_stack_trace(f, ops_to_check=softmax_op)
```
lamblin: In that case, the softmax does not have a stack trace. This is probably something to fix in the optimization, or maybe in another optimization it depends on.

lamblin: Indeed, `copy_stack_trace` is not called from [`softmax_simplifier`](https://github.com/lamblin/Theano/blob/fix_stack_check_nnet/theano/tensor/nnet/nnet.py#L878-L913).

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1454,7 +1471,7 @@ def test_basic_keepdims(self):
...
```
lamblin: This one would be fixed at the same time.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -126,40 +127,35 @@ def test_exp_over_1_plus_exp(self):
 
             # tests inv_1_plus_exp
             f = theano.function([x], T.fill(x, 1.0) / (1 + T.exp(-x)), mode=m)
-            assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
+            assert check_stack_trace(f, ops_to_check=sigmoid)
```
lamblin: The stack trace is missing from the optimization [`local_exp_over_1_plus_exp`](https://github.com/lamblin/Theano/blob/fix_stack_check_nnet/theano/tensor/nnet/sigm.py#L578-L623), which is why the test is failing.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -126,40 +127,35 @@ def test_exp_over_1_plus_exp(self):
...
             f = theano.function([x], T.fill(x, 1.1) / (1 + T.exp(-x)), mode=m)
-            assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
             assert [node.op for node in f.maker.fgraph.toposort()] != [sigmoid]
             f(data)
 
             # tests inv_1_plus_exp with neg
             f = theano.function([x], T.fill(x, -1.0) / (1 + T.exp(-x)), mode=m)
-            assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
+            assert check_stack_trace(
+                f, ops_to_check=[sigmoid, theano.tensor.inplace.neg_inplace])
```
lamblin: This may be also be solved by the same fix.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -347,14 +334,14 @@ def test_local_hard_sigmoid(self):
...
-        assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
+        assert check_stack_trace(f, ops_to_check=sigmoid)
         topo = f.maker.fgraph.toposort()
         assert topo[0].op == sigmoid
         assert len(topo) == 1
 
         mode = self.get_mode().including('local_hard_sigmoid')
         f = theano.function([x], s, mode=mode)
-        assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
+        assert check_stack_trace(f, ops_to_check='all')
```
lamblin: The issue here is that the "hard_sigmoid" replacement gets fused with other elemwise operations, and elemwise fusion does not preserve the stack yet.
The way to solve that here would be to exclude "fusion".
Also, `clip` is going to be made inplace by an opt that does not preserve the stack trace yet.

In the end, I think the following should work:
```python
# For checking that local_hard_sigmoid preserves the stack trace
mode2 = mode.excluding('fusion').excluding('inplace')
f2 = theano.function([x], s, mode=mode2)
self.assertTrue(check_stack_trace(f2, ops_to_check=theano.tensor.clip)
```

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -237,27 +228,24 @@ def test_local_sigm_times_exp(self):
...
         m = self.get_mode(excluding=['local_elemwise_fusion', 'inplace'])
         x, y = tensor.vectors('x', 'y')
 
         f = theano.function([x], sigmoid(-x) * tensor.exp(x), mode=m)
-        assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
         match(f, [sigmoid])
 
         f = theano.function([x], sigmoid(x) * tensor.exp(-x), mode=m)
-        assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
         match(f, [tensor.neg, sigmoid])
```
lamblin: In this case, the `neg` does not have a trace, and it is not obvious to me how to add one given the way the optimization is structured.
For now, I would only to the stack trace check on the sigmoid.

lamblin: Actually, I think we should add a comment in [`local_sigm_times_exp`](https://github.com/lamblin/Theano/blob/fix_stack_check_nnet/theano/tensor/nnet/sigm.py#L911) or `perform_sigm_times_exp` to say we should add stack traces to the intermediate variables at some point, and open a low-priority ticket about it.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -401,7 +391,8 @@ def test_log1msigm_to_softplus(self):
         # Same test with a flatten
```
lamblin: There should be a call to `check_stack_trace` on the previous `f` first.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -430,7 +421,8 @@ def test_log1pexp_to_softplus(self):
 
         out = T.log(1 + T.exp(x))
         f = theano.function([x], out, mode=self.m)
-        assert hasattr(f.maker.fgraph.outputs[0].tag, 'trace')
+
+        assert check_stack_trace(f, ops_to_check='all')
         topo = f.maker.fgraph.toposort()
         assert len(topo) == 1
         assert isinstance(topo[0].op.scalar_op,
```
lamblin: Actually, the optimizations inserting softplus are done through PatternSub (see https://github.com/lamblin/Theano/blob/fix_stack_check_nnet/theano/tensor/nnet/sigm.py#L406), which has no way of preserving stack traces yet.

theano/tensor/nnet/tests/test_nnet.py
nouiz: If all the node in the original graph pattern have the same stack trace (on the same line) then I think it make sense to put that stack trace to all new nodes in the new pattern.

Not sure how hard it is to do or if this is worth it.

theano/tensor/nnet/tests/test_sigm.py
adbrebs: Shall I create a new ticket for that or I take care of it?

theano/tensor/nnet/tests/test_sigm.py
adbrebs: shall we open a new ticket here too?

theano/tensor/nnet/tests/test_nnet.py
lamblin: @nouiz The other nodes did not exist in the original graph. In that test case, they are not used (I think), in some real-life cases, they will replace other nodes created elsewhere, or be merged with them.

theano/tensor/nnet/tests/test_sigm.py
lamblin: Yes, can you open a ticket for that? Then just refer to that ticket in a comment here, instead of actually calling check_stack_trace.

theano/tensor/nnet/tests/test_sigm.py
lamblin: I think it should be quite easy to fix, so I would prefer if you take care of it.

adbrebs: Ok I encounter a problem though, I will come to see this morning.

theano/tensor/nnet/tests/test_nnet.py
adbrebs: I have added it.

theano/tensor/nnet/tests/test_sigm.py
adbrebs: done.

theano/tensor/nnet/sigm.py
```diff
@@ -602,6 +602,7 @@ def local_exp_over_1_plus_exp(node):
             else:
                 # case: 1/(1+exp(x))
                 sigmoids.append(sigmoid(-t))
+            copy_stack_trace(node.outputs[0], sigmoids[-1])
```
adbrebs: @lamblin Despite this change, several of the corresponding check_trace fail, the sigmoid outputs having no trace, I don't really understand why.

lamblin: The issue seems to be that an optimization during the canonicalization phase, introducing the "true_div" operation, does not copy the stack trace.
So this one probably correctly copies the trace, but it is empty in the first place... I'll investigate more.

lamblin: So, that is the `Canonizer` class in `tensor/opt.py` that is responsible for that.
It is a big piece, so for now you can just disable this test and open a ticket, or mention it in an existing ticket.

adbrebs: Ok thanks.

lamblin:
Fixes #4392

PR #4383, opened 2016-04-16
Make stack trace tests in `tensor/tests/test_opt.py` use `check_stack_trace`.

### LINKED ISSUE #4379: Make stack trace tests in `tensor/tests/test_opt.py` use `check_stack_trace`.
Sub-ticket of #3018.
Follow-up of #4114.

The existing tests in `theano/tensor/tests/test_opt.py` use the new `check_stack_trace` function, instead of whatever they are using.
They should check for the outputs of the appropriate nodes, or for `'all'` if it works.

saebrahimi:
Modified test_opt.py to use check_stack_trace and modified check_stack_trace to loop over outputs for ops_to_check=last and some refactoring

fix #4379

### DIFF
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2717,11 +2717,6 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
         for output in node.outputs:
             if (not hasattr(output.tag, 'trace') or
                 not output.tag.trace):
-                # work around some optimization inserting DeepCopyOp and Elemwise without copying the trace.
-                from theano.compile import deep_copy_op, DeepCopyOp
-                if isinstance(fgraph.toposort()[-1].op,
-                              (DeepCopyOp, theano.tensor.Elemwise)):
-                    return True
                 return False
 
     return True
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2716,7 +2716,12 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     for node in apply_nodes_to_check:
         for output in node.outputs:
             if (not hasattr(output.tag, 'trace') or
-                    not output.tag.trace):
+                not output.tag.trace):
+                # work around some optimization inserting DeepCopyOp and Elemwise without copying the trace.
+                from theano.compile import deep_copy_op, DeepCopyOp
+                if isinstance(fgraph.toposort()[-1].op,
+                              (DeepCopyOp, theano.tensor.Elemwise)):
+                    return True
                 return False
 
     return True
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2715,8 +2715,7 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
 
     for node in apply_nodes_to_check:
         for output in node.outputs:
-            if (not hasattr(output.tag, 'trace') or
-                not output.tag.trace):
+            if (not hasattr(output.tag, 'trace') or not output.tag.trace):
                 return False
 
     return True
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2667,7 +2667,7 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     # if ops_to_check is a string
     if isinstance(ops_to_check, string_types):
         if ops_to_check == 'last':
-            apply_nodes_to_check = [fgraph.outputs[0].owner]
+            apply_nodes_to_check = [fgraph.outputs[i].owner for i in range(len(fgraph.outputs))]
         elif ops_to_check == 'all':
             apply_nodes_to_check = fgraph.apply_nodes
         else:
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2667,7 +2667,8 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     # if ops_to_check is a string
     if isinstance(ops_to_check, string_types):
         if ops_to_check == 'last':
-            apply_nodes_to_check = [fgraph.outputs[i].owner for i in range(len(fgraph.outputs))]
+            apply_nodes_to_check = [fgraph.outputs[i].owner for i in range(
+                len(fgraph.outputs))]
         elif ops_to_check == 'all':
             apply_nodes_to_check = fgraph.apply_nodes
         else:
@@ -2711,6 +2712,13 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
             pass
         else:
             raise ValueError('The string bug_print is not recognised')
+
+    elif hasattr(ops_to_check, '__call__'):  # if ops_to_check is a function
+        apply_nodes_to_check = [node for node in fgraph.apply_nodes
+                                if ops_to_check(node)]
+
+    else:  # if ops_to_check is an op or a list of ops
+        raise ValueError('The value of ops_to_check is not supported')
 
     for node in apply_nodes_to_check:
         for output in node.outputs:
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2717,5 +2717,8 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
             if (not hasattr(output.tag, 'trace') or
                     not output.tag.trace):
                 return False
+            # check if trace is not empty
+            if not len(output.tag.trace) > 0:
+                return False
 
     return True
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2713,20 +2713,10 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
         else:
             raise ValueError('The string bug_print is not recognised')
 
-    elif hasattr(ops_to_check, '__call__'):  # if ops_to_check is a function
-        apply_nodes_to_check = [node for node in fgraph.apply_nodes
-                                if ops_to_check(node)]
-
-    else:  # if ops_to_check is an op or a list of ops
-        raise ValueError('The value of ops_to_check is not supported')
-
     for node in apply_nodes_to_check:
         for output in node.outputs:
             if (not hasattr(output.tag, 'trace') or
                     not output.tag.trace):
-                return False
-            # check if trace is not empty
-            if not len(output.tag.trace) > 0:
                 return False
 
     return True
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -112,6 +112,7 @@ class test_dimshuffle_lift(unittest.TestCase):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[x]")
         # Check stacktrace was copied over correctly after opt was applied
+        # Optimization may remove all apply_nodes, so we ignore these cases
         self.assertTrue(check_stack_trace(g, ops_to_check='all',
                                           bug_print='ignore'))
 
@@ -138,6 +139,7 @@ class test_dimshuffle_lift(unittest.TestCase):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[x]", str(g))
         # Check stacktrace was copied over correctly after opt was applied
+        # Optimization may remove all apply_nodes, so we ignore these cases
         self.assertTrue(check_stack_trace(g, ops_to_check='all',
                                           bug_print='ignore'))
 
@@ -2551,6 +2553,7 @@ class test_local_subtensor_merge(unittest.TestCase):
             f = theano.function([x] + input_vars, sub_x, mode=mode_opt)
 
             # Check stacktrace was copied over correctly after opt was applied
+            # Optimization may remove all Subtensors, so we ignore these cases
             self.assertTrue(check_stack_trace(f, ops_to_check=Subtensor,
                                               bug_print='ignore'))
 
@@ -5946,8 +5949,7 @@ def test_local_useless_split():
     # | |x [id D]
     # ...
 
-    #assert check_stack_trace(f_opt, ops_to_check=[
-    #    Assert], bug_print='ignore')
+    #assert check_stack_trace(f_opt, ops_to_check=[Assert])
     assert check_stack_trace(f_nonopt, ops_to_check='all')
 
 
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -2125,7 +2125,7 @@ class test_local_subtensor_merge(unittest.TestCase):
                 'local_subtensor_merge'))
 
             # Check stacktrace was copied over correctly after opt was applied
-            self.assertTrue(check_stack_trace(f, ops_to_check='last'))
+            self.assertTrue(check_stack_trace(f, ops_to_check=Subtensor))
 
             topo = f.maker.fgraph.toposort()
             assert len([t for t in topo
@@ -5929,9 +5929,7 @@ def test_local_useless_split():
 
     # Check that stacktraces have been copied over properly
     assert check_stack_trace(f_opt, ops_to_check='all')
-    assert len(f_opt.outputs[0].variable.tag.trace) > 0
     assert check_stack_trace(f_nonopt, ops_to_check='all')
-    assert len(f_nonopt.outputs[0].variable.tag.trace) > 0
 
 
 def test_local_flatten_lift():
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -2960,8 +2960,10 @@ def test_local_IncSubtensor_serialize():
 
     # Now test that the stack trace is copied over properly,
     # if we return the gradients. We need to use same mode as before.
+    # the optimizer may delete all instances of AdvancedIncSubtensor
     f = theano.function([i, j, t], dW, mode=mode)
-    assert check_stack_trace(f, ops_to_check=tensor.AdvancedIncSubtensor)
+    assert check_stack_trace(f, ops_to_check=tensor.AdvancedIncSubtensor,
+                             bug_print='ignore')
         
 def test_local_set_to_inc_subtensor():
     v = theano.tensor.fmatrix()
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -112,7 +112,8 @@ class test_dimshuffle_lift(unittest.TestCase):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[x]")
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(g, ops_to_check='all'))
+        self.assertTrue(check_stack_trace(g, ops_to_check='all',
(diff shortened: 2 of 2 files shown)


lamblin:
Thanks for the PR, I'll have a deeper look later today.
For the moment, it looks like there are lots of tests with `ops_to_test='last'`. Did you check all of them to see that the last node is indeed the one we want to check?

vmichals:
Hi Pascal, 
thanks for checking! I'm pretty sure Samira first ran the unit tests with 'all' and only set it to 'last ' for the ones where that failed.
BestVince


Sent from my Samsung device

-------- Original message --------

saebrahimi:
I set it to 'last' if 'all' failed and the previous check was on outputs[0] or on an output inside a loop over outputs. The loop cases are covered by the modification in check_stack_trace function.

lamblin:
Actually, my question is: when you set it to 'last', did you check manually that the last element of the toposort was indeed an Op of the right type? The Op to check may not be the last one.
For instance, [here](https://github.com/Theano/Theano/pull/4383/files#diff-4002cc261ad1d205365ecaa717ea330bR2104), I think we should actually check all `Subtensor` nodes. So normally we should call something like `check_stack_trace(f, ops_to_check=Subtensor)`.
The only case where `'last'` is OK would be if it was manually checked that, for instance, there is only one `Subtensor` node in the graph and it is the output.

I understand that the tests currently only look at the output, but this was not correct, which is why we want to update those tests.

lamblin:
Thanks for the update. @adbrebs will be applying that fix to `check_stack_trace` as well.
What this shows is that we are not ready for `opt_to_check='all'` in the first cases, and you will have to check the trace on the right subclass of Ops.

theano/gof/opt.py
```diff
@@ -2716,7 +2716,12 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     for node in apply_nodes_to_check:
         for output in node.outputs:
             if (not hasattr(output.tag, 'trace') or
-                    not output.tag.trace):
+                not output.tag.trace):
+                # work around some optimization inserting DeepCopyOp and Elemwise without copying the trace.
+                from theano.compile import deep_copy_op, DeepCopyOp
+                if isinstance(fgraph.toposort()[-1].op,
+                              (DeepCopyOp, theano.tensor.Elemwise)):
+                    return True
```
lamblin: No, that is not the way to go. Please do not change that function for the moment.

theano/tensor/tests/test_opt.py
```diff
@@ -111,7 +111,8 @@ def test_double_transpose(self):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[x]")
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(g, ops_to_check='all'))
+        self.assertTrue(check_stack_trace(g, ops_to_check='all',
+                                          bug_print='ignore'))
```
lamblin: Here, the optimization _removed_ some nodes, without inserting new ones. Then another optimization inserted a DeepCopyOp, but this is not what we want to check here. So you have two options:
- call `check_stack_trace(g, ops_to_check=DimShuffle, bug_print='ignore')`, since the only Op we want to check here is DimShuffle, and it will probably not be there
- not call `check_stack_trace` at all, since the graph is supposed to be empty, and leave a comment explaining that.

theano/tensor/tests/test_opt.py
```diff
@@ -137,7 +138,8 @@ def test_elim3(self):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[x]", str(g))
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(g, ops_to_check='all'))
+        self.assertTrue(check_stack_trace(g, ops_to_check='all',
+                                          bug_print='ignore'))
```
lamblin: Same here.

theano/tensor/tests/test_opt.py
```diff
@@ -2502,8 +2518,9 @@ def test_none_slice(self):
             sub_x = x[slice1][slice2]
             f = theano.function([x] + input_vars, sub_x, mode=mode_opt)
 
-            self._verify_stack_trace(f)
-
+            # Check stacktrace was copied over correctly after opt was applied
```
lamblin: Please leave a comment saying that for some cases, the optimization may remove all Subtensors, which is why you pass "bug_print='ignore'".

theano/tensor/tests/test_opt.py
```diff
@@ -123,7 +125,7 @@ def test_merge2(self):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[DimShuffle{0,1,x,x}(x)]", str(g))
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(hasattr(g.outputs[0].tag, 'trace'))
+        self.assertTrue(check_stack_trace(g, ops_to_check='all'))
```
lamblin: If this check does not pass without your change ignoring Elemwise and DeepCopy, you should probably pass `ops_to_check=DimShuffle` here.

theano/tensor/tests/test_opt.py
```diff
@@ -137,7 +138,8 @@ def test_elim3(self):
         dimshuffle_lift.optimize(g)
         self.assertTrue(str(g) == "[x]", str(g))
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(g, ops_to_check='all'))
+        self.assertTrue(check_stack_trace(g, ops_to_check='all',
+                                          bug_print='ignore'))
```
lamblin: Also here, `'all'` will not work with the correct version of `check_stack_trace`, so you would have to either pass `ops_to_check=DimShuffle`, or remove the check and add a comment.

theano/tensor/tests/test_opt.py
```diff
@@ -2514,11 +2516,11 @@ def test_none_slice(self):
...
             slice2 = slice(*slice_inputs[3:])
             sub_x = x[slice1][slice2]
-            f = theano.function([x] + input_vars, sub_x,
-                mode=mode_opt.excluding('local_useless_subtensor'))
+            f = theano.function([x] + input_vars, sub_x, mode=mode_opt)
 
             # Check stacktrace was copied over correctly after opt was applied
-            self.assertTrue(check_stack_trace(f, ops_to_check=Subtensor))
+            self.assertTrue(check_stack_trace(f, ops_to_check=Subtensor,
+                                              bug_print='ignore'))
```
lamblin: Otherwise, this is the right thing to do.

theano/tensor/tests/test_opt.py
```diff
@@ -2922,7 +2924,7 @@ def test_local_IncSubtensor_serialize():
     # Now test that the stack trace is copied over properly,
     # if we return the gradients. We need to use same mode as before.
     f = theano.function([i, j, t], dW, mode=mode)
-    assert check_stack_trace(f, ops_to_check='all')
+    assert check_stack_trace(f, ops_to_check=tensor.AdvancedIncSubtensor)
```
lamblin: We would want to include IncSubtensor and AdvancedIncSubtensor1 in there as well, I think. For instance:
`ops_to_check=[tensor.Subtensor, tensor.AvancedIncSubtensor1, tensor.AdvancedIncSubtensor]`.
If that is not working, please leave a note, as it may indicate a problem to investigate.

theano/tensor/tests/test_opt.py
```diff
@@ -2954,7 +2956,7 @@ def test_local_set_to_inc_subtensor():
 
     # Finally, test that the stack trace is copied over properly,
     # before before and after optimization.
-    assert check_stack_trace(f1, ops_to_check='all')
+    assert check_stack_trace(f1, ops_to_check=tensor.AdvancedIncSubtensor1)
```
lamblin: There is no need to do that on f1, since f1 is compiled without the optimization we want to test, so if it does not work, you can just remove it.
If it works, though, it is slightly better to keep it.

theano/tensor/tests/test_opt.py
```diff
@@ -3599,8 +3601,8 @@ def test_advanced_inc_subtensor(self):
         utt.assert_allclose(r1, r2)
         
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(f1, ops_to_check='all'))
-        self.assertTrue(check_stack_trace(f2, ops_to_check='all'))
+        self.assertTrue(check_stack_trace(f1, ops_to_check=tensor.AdvancedIncSubtensor))
```
lamblin: Same here. If it works, then it is fine.

theano/tensor/tests/test_opt.py
```diff
@@ -1870,16 +1870,17 @@ def test_stacktrace(self):
         mode = theano.compile.mode.Mode(optimizer=None).including('canonicalize_db').including("local_subtensor_make_vector")
         f = function([x, y, z], v[0], mode=mode)
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
-        #import ipdb; ipdb.set_trace()
+        # FIXME: DeepCopyOp does not have a trace
+        #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```
lamblin: Again, the issue is not that DeepCopyOp does not have a trace.
The issue is that you have to check the trace on the appropriate Op, not 'all'. Please:
- have a look that the implementation of "local_subtensor_make_vector",
- find out which Ops it is adding, and
- check the trace on these Ops.

theano/tensor/tests/test_opt.py
```diff
@@ -1870,16 +1870,17 @@ def test_stacktrace(self):
...
         
         
         # Compile function using all optimizations in fast_compile mode, 
         # including the 'local_subtensor_make_vector' optimization
         mode = theano.compile.mode.get_mode('FAST_COMPILE').including("local_subtensor_make_vector")
         f = function([x, y, z], v[0], mode=mode)
         # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # FIXME: DeepCopyOp does not have a trace
+        #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```
lamblin: Same here.

theano/tensor/tests/test_opt.py
lamblin: And if it does not add Ops and only remove one, please leave a comment saying that.

theano/tensor/tests/test_opt.py
```diff
@@ -1977,8 +1981,8 @@ def test4(self):
         y = tensor.vector('y')
         f = function([x, y], tensor.exp(x + y)[0], mode=mode_opt)
 
-        # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # opt doesn't apply, so no need for check_stack_trace
+        #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```
lamblin: The comment is correct.
You should add a space after the `#` so that pep8 is respected.

theano/tensor/tests/test_opt.py
```diff
@@ -3555,9 +3555,9 @@ def test2(self):
         # we have to change the following we should not see tensor.Alloc
         # in op_classes and we have to change the assert.
         assert tensor.Alloc in op_classes
+        # self.assertTrue(check_stack_trace(f, ops_to_check=[tensor.Alloc]))
 
         # Check stacktrace was copied over correctly after opt was applied
```
lamblin: Please update the comment, and put it before the commented-out "self.assertTrue(...)".

theano/tensor/tests/test_opt.py
```diff
@@ -3519,8 +3529,8 @@ def test0(self):
         if isinstance(mode_opt, compile.DebugMode):
             self.assertRaises(ValueError, f)
 
-        # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # No need to check_stack_trace
```
lamblin: Please add a comment explaining _why_ there is no need.

theano/tensor/tests/test_opt.py
```diff
@@ -3535,8 +3545,8 @@ def test1(self):
         op_classes = [node.op.__class__ for node in f.maker.fgraph.toposort()]
         assert tensor.Alloc not in op_classes
 
-        # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # No need to check_stack_trace
```
lamblin: Same here.

theano/tensor/tests/test_opt.py
```diff
@@ -4037,8 +4045,8 @@ def test_local_useless_tile(self):
                 f(data)
                 
                 # Check that stacktrace is copied over
-                self.assertTrue(check_stack_trace(f, ops_to_check='all'))
-                self.assertTrue(len(f.outputs[0].variable.tag.trace)>0)
+                # FIXME: DeepCopyOp has empty trace
+                #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```
lamblin: Update the comment to say that the optimization has only removed Ops, not inserted new ones, which is why there is no stack trace to check.

theano/tensor/tests/test_opt.py
```diff
@@ -2936,8 +2936,10 @@ def test_local_IncSubtensor_serialize():
 
     # Now test that the stack trace is copied over properly,
     # if we return the gradients. We need to use same mode as before.
+    # the optimizer may delete all instances of AdvancedIncSubtensor
     f = theano.function([i, j, t], dW, mode=mode)
-    assert check_stack_trace(f, ops_to_check=tensor.AdvancedIncSubtensor)
+    assert check_stack_trace(f, ops_to_check=tensor.AdvancedIncSubtensor,
```
lamblin: I think you also need to check for Subtensor and AvancedIncSubtensor1. If you do that, you may be able to remove "bug_print='ignore'".

lamblin:
Also, please run test_flake8 locally, to catch potential formatting errors.

For all the tests that discover a potential problem in the optimization itself, can you document them in a ticket? Please mention the name of the test, the optimization, what was wrong, etc. Link directly to the code if you can.

lamblin:
I'm not sure why Travis shows my last comments as "outdated", they should still be valid.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/gof/opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done. 'all' still works.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done. It still worked.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
saebrahimi: Done.

theano/tensor/tests/test_opt.py
```diff
@@ -1895,18 +1889,17 @@ def test_stacktrace(self):
...
         # Compile function using all optimizations in fast_compile mode, 
         # including the 'local_subtensor_make_vector' optimization
         mode = theano.compile.mode.get_mode('FAST_COMPILE').including("local_subtensor_make_vector")
         f = function([x, y, z], v[0], mode=mode)
-        # Check stacktrace was copied over correctly after opt was applied
-        # FIXME: DeepCopyOp does not have a trace
-        #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # The resulting graph only has a DeepCopyOp, for which the
+        # copy_stack_trace is not called. (See issue #4421)
+        # self.assertTrue(check_stack_trace(f, ops_to_check='all'))
```
lamblin: The issue here is not that the DeepCopyOp has no stack trace, it is that the two cases in this test do not check the case where `local_subtensor_make_vector` inserts a `Subtensor` node.
It is OK to just refer to #4421 for now and not fix it.

theano/tensor/tests/test_opt.py
```diff
@@ -3558,8 +3551,8 @@ def test0(self):
         if isinstance(mode_opt, compile.DebugMode):
             self.assertRaises(ValueError, f)
 
-        # No need to check_stack_trace
-        #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # No need to check_stack_trace as the graph only contains
+        # DeepCopyOp which has empty trace
```
lamblin: The issue is not with DeepCopyOp, you can just say that the optimization `local_useless_alloc` only removes nodes.

theano/tensor/tests/test_opt.py
```diff
@@ -3574,8 +3567,8 @@ def test1(self):
         op_classes = [node.op.__class__ for node in f.maker.fgraph.toposort()]
         assert tensor.Alloc not in op_classes
 
-        # No need to check_stack_trace
-        #self.assertTrue(check_stack_trace(f, ops_to_check='all'))
+        # No need to check_stack_trace as the graph only contains
+        # DeepCopyOp which has empty trace
```
lamblin: Same here.

lamblin:
I have some minor comments to clarify some comments in the tests, but otherwise it looks good.
Thanks.

lamblin:
Thanks!

Issue #4178, opened 2016-03-07
Keep stack trace: finish local opt in theano/tensor/opt.py


lamblin:

saebrahimi:
concerning local_flatten_lift: it seems flatten is wrapper around reshape and opt is not called.
should I remover test_local_flatten_lift?

lamblin:
We can ignore `local_flatten_lift` for the moment, we will remove it at some point (and the test as well).
I'll update the ticket.

saebrahimi:
In addition to the ones already mentioned in the ticket, I found these local optimizers without copy_stack_trace:

- [ ] local_merge_switch_same_cond
- [ ] Canonizer
- [ ] mul_calculate
- [ ] local_neg_to_mul
- [ ] local_sum_prod_div_dimshuffle
- [ ] local_sum_prod_all_to_none
- [ ] local_op_of_op
- [ ] local_reduce_join
- [ ] local_cut_useless_reduce
- [ ] local_reduce_broadcastable
- [ ] local_opt_alloc
- [ ] local_neg_neg
- [ ] local_neg_div_neg
- [ ] local_mul_zero
- [ ] local_div_to_inv
- [ ] local_inv_canon
- [ ] local_pow_canonicalize
- [ ] local_mul_to_sqr
- [ ] local_intdiv_by_one
- [ ] local_zero_div
- [ ] local_pow_specialize
- [ ] local_pow_specialize_device
- [ ] local_mul_specialize
- [ ] local_add_specialize
- [ ] check_for_x_over_absX
- [ ] local_abs_lift
- [ ] local_abs_merge
- [ ] local_log1p
- [ ] local_log_add
- [ ] add_calculate
- [ ] local_greedy_distributor (distribute_greedy,attempt_distribution)
- [ ] local_one_plus_erf
- [ ] local_one_minus_erf
- [ ] local_one_minus_erf2
- [ ] local_one_plus_neg_erf
- [ ] local_erf_minus_one
- [ ] local_one_minus_erfc
- [ ] local_one_minus_erfc2
- [ ] local_one_minus_erfc3
- [ ] local_one_add_neg_erfc
- [ ] local_erf_neg_minus_one
- [ ] local_erf_neg_minus_one2
- [ ] local_log_erfc
- [ ] local_grad_log_erfc_neg
- [ ] local_elemwise_fusion_op
- [ ] local_elemwise_fusion
- [ ] FusionOptimizer
- [ ] local_grad_clip
[...]

PR #4385, opened 2016-04-17
Keep stack trace: finish local opt in theano/tensor/opt.py

saebrahimi:
**local_track_shape_i** and **local_subtensor_of_alloc**: added check_stack_trace calls in unit test. They passed, so no changes to opts. 

**local_inplace_incsubtensor1**: Already had copy_stack_trace. Added a unit test and it worked.

see issue #4178


theano/tensor/tests/test_opt.py
```diff
@@ -4050,7 +4050,7 @@ def test_local_useless_tile(self):
                 f(data)
                 
                 # Check that stacktrace is copied over
-                self.assertTrue(hasattr(f.outputs[0].variable.tag, 'trace'))
+                self.assertTrue(check_stack_trace(f, ops_to_check='all'))
                 self.assertTrue(len(f.outputs[0].variable.tag.trace)>0)
```
saebrahimi: I think this should be `self.assertTrue(len(f.maker.fgraph.outputs[0].tag.trace)>0)` to check the length of the trace in **optimized** function, but that doesn't pass because the trace is empty. @lamblin

lamblin:
So there is a trace but it's empty? That is strange.
The issue seems to be that the output is a `DeepCopyOp`, can you check if that is the case? If it is, then we should find the optimization inserting it, and make it have the stack trace of the original variable.
Also, in that case, I think `check_stack_trace` should also check for the length of the stacktrace.

However, let's not worry about it for now, just keep the call to check_stack_trace.

lamblin:
Actually, could you try adding that check about `len(...) > 0` to `check_stack_trace` in the other PR, and see if the tests still pass? That may be an issue.

saebrahimi:
yes, the output is a `DeepCopyOp`.
Could it be that the function `is_same_graph_with_merge` in gof/opt.py is executed after the optimization?

I will insert the length check in previous PR as well.

theano/tensor/opt.py
```diff
@@ -3879,11 +3879,13 @@ def local_useless_tile(node):
                         # The Op don't support that case, so we can't
                         # implement the opt and test it.
                         return
+                        # TODO remove this part?
                         return [node.inputs[0]]
```
lamblin: Sure, we can do some clean-up.

theano/tensor/opt.py
```diff
@@ -3879,11 +3879,13 @@ def local_useless_tile(node):
...
                     else:
                         # The Op don't support that case, so we can't
                         # implement the opt and test it.
                         return
+                        # TODO remove this part?
```
lamblin: Same here.

theano/tensor/tests/test_opt.py
```diff
@@ -3231,6 +3233,7 @@ def test_local_subtensor_of_alloc():
             for slices in slicess:
                 z = yx.__getitem__(slices)
                 f = theano.function([x], z)
+                assert check_stack_trace(f, ops_to_check='all')
```
lamblin: If 'all' does not work, check alloc, subtensor, rebroadcast

lamblin: You can put that call inside the following "if" block if needed.

lamblin:
For `local_subtensor_of_alloc`:
- add the stack trace from node.outputs[0] (the original subtensor) to the result of `var[...]` around https://github.com/Theano/Theano/blob/master/theano/tensor/opt.py#L2965
- add the stack trace from both node.outputs[0] and `u` (the original alloc) to `rval[0]`
- add them both to the output of the rebroadcast as well when it is inserted

theano/tensor/tests/test_opt.py
```diff
@@ -3239,6 +3242,16 @@ def test_local_subtensor_of_alloc():
...
+def test_local_inplace_incsubtensor1():
+    # Test local_inplace_incsubtensor1 for stack trace
+    x = fmatrix()
+    y = x**2
+    a = lmatrix()
+    z = advanced_inc_subtensor1(y, a, [1,3])
+    f = theano.function([x, a], z)
+    assert check_stack_trace(f, ops_to_check='all')
+
+
```
lamblin: You should move that to `test_subtensor.py`.
Make sure that the adv_inc_subtensor1 in the graph is actually inplace.

theano/tensor/opt.py
saebrahimi: done

theano/tensor/opt.py
saebrahimi: done

theano/tensor/tests/test_opt.py
saebrahimi: done

theano/tensor/tests/test_opt.py
```diff
@@ -3362,25 +3362,19 @@ def test_local_subtensor_of_alloc():
             for slices in slicess:
                 z = yx.__getitem__(slices)
                 f = theano.function([x], z)
-                assert check_stack_trace(f, ops_to_check='all')
                 if theano.config.mode != 'FAST_COMPILE':
                     # Subtensor can be in the input of Alloc
                     assert not isinstance(f.maker.fgraph.toposort()[-1].op,
                                           Subtensor)
+                    if not (isinstance(f.maker.fgraph.toposort()[-1].op,
+                                       DeepCopyOp) and len(f.maker.fgraph.toposort())==1):
```
lamblin: I think you can remove that check, passing `bug_print='ignore'` should be enough.

lamblin:
Some of the tests calling `check_stack_trace(..., 'all')` are failing.
Can you investigate that, and maybe restrict `ops_to_check`?

nouiz:
@ReyhaneAskari should we close this PR?

Issue #4072, opened 2016-02-16
Replace some Alloc by DimShuffle


lamblin:
Sub-ticket of #2943.
When `Alloc` adds broadcastable dimensions (of length 1, to the left), then it can be replaced by a `DimShuffle` instead.

For instance:
- `alloc(3, 1, 1)` could be replaced by `dimshuffle{x, x}(3)` (edit: ~~`dimshuffle{x, x, 0}(3)`~~)
- `alloc([3 4], 1, 3, 2)` could be replaced by `dimshuffle{x, 0, 1}(alloc([3 4], 3, 2)`

nouiz:
I think the first example should be:

 alloc(3, 1, 1) could be replaced by dimshuffle{x, 0}(3) 

I think we can make this work not just when the added dimensions is at the left. But doing it at the left would be a good first step.

lamblin:
I meant `dimshuffle{x,x}(3)`, actually, I'll edit.
Also, alloc only adds dimensions to the left, it cannot add dimensions to the right by itself (the allocated value would have to be dimshuffled already). However, if there are other broadcastable dimensions that were already there before the alloc, then we could "lift" them, but I'm not sure if it is a good idea.

lamblin:
Closed by #4388.

PR #4388, opened 2016-04-18
Useless alloc

mohammadpz:
Regarding issue #4072 .

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1699,7 +1699,7 @@ def local_useless_alloc(node):
     # Check if alloc adds a broadcastable dimension with shape 1.
     output_shape = node.inputs[1:]
     num_dims_with_size_1_added_to_left = 0
-    for i in range(len(output_shape)):
+    for i in range(len(output_shape) - input.ndim):
         if extract_constant(output_shape[i], only_process_constants=True) == 1:
             num_dims_with_size_1_added_to_left += 1
         else:
@@ -1708,7 +1708,7 @@ def local_useless_alloc(node):
     if num_dims_with_size_1_added_to_left > 0 and len(new_output_shape) >= input.ndim:
         inner = op(*([input] + new_output_shape))
         dimshuffle_new_order = (['x'] * num_dims_with_size_1_added_to_left +
-                                range(len(new_output_shape)))
+                                list(xrange(len(new_output_shape))))
         return [DimShuffle(inner.type.broadcastable, dimshuffle_new_order)(inner)]
 
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -47,7 +47,8 @@ from theano.tensor.type import (values_eq_approx_remove_inf,
 from theano.gof.opt import (Optimizer, pre_constant_merge,
                             pre_greedy_local_optimizer)
 from theano.gof import toolbox
-from theano.tensor.basic import Alloc, get_scalar_constant_value, ShapeError, NotScalarConstantError
+from theano.tensor.basic import (Alloc, get_scalar_constant_value, ShapeError,
+                                 extract_constant, NotScalarConstantError)
 from six import StringIO
 
 _logger = logging.getLogger('theano.tensor.opt')
@@ -1699,7 +1700,7 @@ def local_useless_alloc(node):
     output_shape = node.inputs[1:]
     num_dims_with_size_1_added_to_left = 0
     for i in range(len(output_shape)):
-        if output_shape[i].get_scalar_constant_value() == 1:
+        if extract_constant(output_shape[i], only_process_constants=True) == 1:
             num_dims_with_size_1_added_to_left += 1
         else:
             break
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -47,7 +47,7 @@ from theano.tensor.type import (values_eq_approx_remove_inf,
 from theano.gof.opt import (Optimizer, pre_constant_merge,
                             pre_greedy_local_optimizer)
 from theano.gof import toolbox
-from theano.tensor.basic import get_scalar_constant_value, ShapeError, NotScalarConstantError
+from theano.tensor.basic import Alloc, get_scalar_constant_value, ShapeError, NotScalarConstantError
 from six import StringIO
 
 _logger = logging.getLogger('theano.tensor.opt')
@@ -1683,10 +1683,32 @@ def local_useless_alloc(node):
     of the input. This is not needed.
 
     """
-    if node.op == T.alloc:
-        if node.inputs[0].type == node.outputs[0].type:
-            # We don't need to copy over any stack traces here
-            return [node.inputs[0]]
+    op = node.op
+    if not isinstance(op, Alloc):
+        return False
+
+    input = node.inputs[0]
+    output = node.outputs[0]
+
+    # Check if dtype and broadcast remain the same.
+    if input.type == output.type:
+        # We don't need to copy over any stack traces here
+        return [input]
+
+    # Check if alloc adds a broadcastable dimension with shape 1.
+    output_shape = node.inputs[1:]
+    num_dims_with_size_1_added_to_left = 0
+    for i in range(len(output_shape)):
+        if output_shape[i].value == 1:
+            num_dims_with_size_1_added_to_left += 1
+        else:
+            break
+    if num_dims_with_size_1_added_to_left > 0:
+        new_output_shape = output_shape[num_dims_with_size_1_added_to_left:]
+        inner = op(*([input] + new_output_shape))
+        dimshuffle_new_order = (['x'] * num_dims_with_size_1_added_to_left +
+                                range(len(new_output_shape)))
+        return [DimShuffle(inner.type.broadcastable, dimshuffle_new_order)(inner)]
 
 
 # Don't register by default.
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1704,8 +1704,8 @@ def local_useless_alloc(node):
             num_dims_with_size_1_added_to_left += 1
         else:
             break
-    if num_dims_with_size_1_added_to_left > 0:
-        new_output_shape = output_shape[num_dims_with_size_1_added_to_left:]
+    new_output_shape = output_shape[num_dims_with_size_1_added_to_left:]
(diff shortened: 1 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -1683,10 +1683,32 @@ def local_useless_alloc(node):
...
+    # Check if dtype and broadcast remain the same.
+    if input.type == output.type:
+        # We don't need to copy over any stack traces here
+        return [input]
+
+    # Check if alloc adds a broadcastable dimension with shape 1.
+    output_shape = node.inputs[1:]
+    num_dims_with_size_1_added_to_left = 0
+    for i in range(len(output_shape)):
+        if output_shape[i].value == 1:
```
lamblin: You cannot simply access the `.value` field here, because it may not be a constant.
There is a useful function for that, `get_scalar_constant_value`, you should call it with `only_process_constants=True` if possible.

mohammadpz: I updated it. It doesn't have the argument ```only_process_constants=True``` though.

lamblin: What do you mean, it does not have that argument?
The definition is [here](https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L578-L579)

lamblin:
OK, so `get_scalar_constant_value` raises an exception if the variable is not constant.
You can use [`extract_constant`](https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L3352) instead, this may be more convenient.

mohammadpz:
Here's an example that I get error on:
Consider ```Alloc(TensorConstant{(1, 1) of 1.0}, TensorConstant{1}, Shape_i{0}.0)```.
So the input that we want to do alloc on is ```TensorConstant{(1, 1) of 1.0}``` and the output shape is ```(TensorConstant{1}, Shape_i{0}.0)```.
In code I detect the ```TensorConstant{1}``` added to the left. So I just remove it to add it using ```dimshuffle```. But the inner alloc would be ```Alloc(TensorConstant{(1, 1) of 1.0}, Shape_i{0}.0)```. So here I get this error:
```*** TypeError: ('The Alloc value to use has more dimensions than the specified dimensions', 2, 1)```.

What am I missing?

**Comment by mohammadpz** [comment]
Do you think that I can add the following condition for applying the optimization:
```if num_dims_with_size_1_added_to_left > 0 and  new_output_shape>= input.ndim```.
I'm not sure if I should also check for broadcastabality or not.

**Comment by lamblin** [comment]
Indeed, you are detecting the total number of length-1 dimension to the left of the output, not the number of *added* ones. There are no added dimensions in your case, since the input, `TensorConstant{(1, 1) of 1.0}`, has already 2 dimensions.
Your current fix does not work either, because it would fail to handle correctly something like `alloc([[1]], 1, 1, 2)`.

You have 2 options at this point:
- the simplest one is to only call dimshuffle if needed for the additional dimensions, which was the original plan. That would ignore cases like `alloc([[1]], 1, 2)`, but transform `alloc([[1]], 1, 1, 2)` into `alloc([[1]], 1, 2).dimshuffle('x', 0, 1)`.
- another option would be to try to dig into the allocated value itself, to see if we could drop dimensions before the alloc, and add it back later. That would transform `alloc([[1]], 1, 2)` into `alloc([1], 2).dimshuffle('x', 0)`, and `alloc([[1]], 1, 1, 2)` into `alloc([1], 2).dimshuffle('x', 'x', 0)`. However, the way to do that in general would be to add a dimshuffle before and after the alloc, which would make the graph more complicated.

I think the first solution is better.

**Comment by lamblin** [comment]
There is also a test failing, but that may be because the graph it is expecting is not the same when your optimization is applied.
This is something we will have to look more deeply later.

**Comment by lamblin** [comment]
I've made a sub-PR to your branch that should fix the test of the other optimization.

File: theano/tensor/opt.py [hunk 2]
```diff
@@ -1683,10 +1684,32 @@ def local_useless_alloc(node):
...
+    for i in range(len(output_shape)):
+        if extract_constant(output_shape[i], only_process_constants=True) == 1:
+            num_dims_with_size_1_added_to_left += 1
+        else:
+            break
+    new_output_shape = output_shape[num_dims_with_size_1_added_to_left:]
+    if num_dims_with_size_1_added_to_left > 0 and len(new_output_shape) >= input.ndim:
+        inner = op(*([input] + new_output_shape))
+        dimshuffle_new_order = (['x'] * num_dims_with_size_1_added_to_left +
+                                range(len(new_output_shape)))
```
lamblin: In Python 3, `range` returns a generator, not a list, so you cannot concatenate it like that.
The solution would be to use `list(xrange(...))` instead.

lamblin:
After that, I think this PR would be good.

theano/tensor/tests/test_opt.py
```diff
@@ -3547,6 +3548,34 @@ def test2(self):
...
+        alloc_lift.optimize(g)
+        self.assertTrue(str(g) == "[DimShuffle{x,0,1}"
+                                  "(Alloc(<TensorType(float64, vector)>, "
+                                  "TensorConstant{3}, "
+                                  "TensorConstant{2})), "
+                                  "DimShuffle{x,x}"
+                                  "(<TensorType(float64, scalar)>)]")
+
+        # Check stacktrace was copied over correctly after opt was applied
+        self.assertTrue(hasattr(g.outputs[0].tag, 'trace'))
```
lamblin: Actually, the way of checking for stack traces now would be to call `self._verify_stack_trace(g)`.

mohammadpz: @lamblin , ```g``` is not a function but a ```FunctionGraph```. So instead of ```self.assertTrue(hasattr(output.variable.tag, 'trace'))```, it should be ```self.assertTrue(hasattr(output.tag, 'trace'))```.
If I convert ```g``` to a function, then I don't know how to check ```str(g)```.
What's your opinion?

lamblin: Actually, `self._verify_stack_trace` has been removed since all of the other tests removed nodes, and there was no stack trace to check.
We have upgraded the way of checking for traces as well, we have a function for that, [`check_stack_trace`](https://github.com/Theano/Theano/blob/3c70348fd7c31fe169cd2038999e3a5ba106d738/theano/gof/opt.py#L2672) that accepts either functions or fgraphs.
So, you can simply call:
```python
self.assertTrue(check_stack_trace(g, ops_to_check='all'))
```
If it gives an error, please report it and we will check if the optimization has to be updated, or if the check should be less strict.

lamblin:
I killed the Travis build because two things are still missing, and we have other PRs lined up.
- Can you merge the following PR to your branch? It fixes another test: https://github.com/mohammadpz/Theano/pull/1
- Call `check_stack_trace` in the test.

lamblin:
Good, thanks!

theano/tensor/tests/test_opt.py
```diff
@@ -3034,6 +3037,10 @@ def _verify_stack_trace(self, f):
             self.assertTrue(hasattr(output.variable.tag, 'trace'))
 
     def test_remove_alloc_wo_dimshuffle(self):
+        # Exclude local_useless_alloc, since it does not introduce
+        # assert in all the same cases.
+        self.fast_run_mode = self.fast_run_mode.excluding(
+            'local_useless_alloc')
```
nouiz: I think this will change the value of self for the following tests in that class. I don't think this is a good idea. So don't affect it to self.fast_run_mode, but use it only locally.

theano/tensor/tests/test_opt.py
```diff
@@ -3547,6 +3554,57 @@ def test2(self):
...
+        self.assertTrue(str(g) == "[DimShuffle{x,0,1}"
+                                  "(Alloc(<TensorType(float64, vector)>, "
+                                  "TensorConstant{3}, "
+                                  "TensorConstant{2})), "
+
+                                  "DimShuffle{x,x}"
+                                  "(<TensorType(float64, scalar)>), "
+
+                                  "DimShuffle{x,0,1}"
+                                  "(Alloc(<TensorType(float64, matrix)>, "
```
nouiz: I don't like the idea that we have more nodes after this optimization when one of the node can do the work of both.

Why we wanted that? I understand that if we can remove the Alloc completly it is a good idea. But when we can't i would keep the original node to have a smaller graph. Why do we want to explicitly do the dimshuffle in the graph for Alloc?

theano/tensor/opt.py
```diff
@@ -1683,10 +1684,32 @@ def local_useless_alloc(node):
...
+    output_shape = node.inputs[1:]
+    num_dims_with_size_1_added_to_left = 0
+    for i in range(len(output_shape) - input.ndim):
+        if extract_constant(output_shape[i], only_process_constants=True) == 1:
+            num_dims_with_size_1_added_to_left += 1
+        else:
+            break
+    new_output_shape = output_shape[num_dims_with_size_1_added_to_left:]
+    if num_dims_with_size_1_added_to_left > 0 and len(new_output_shape) >= input.ndim:
+        inner = op(*([input] + new_output_shape))
```
nouiz: If we want to keep the case where we have a dimshuffle and an alloc, I would special case the case when we will completly remove the Alloc to speed up the compilation instead of requesting that other opt do the clean up.

theano/tensor/tests/test_opt.py
lamblin: `self.fast_run_mode` is defined in `setUp`, which is run again before each test.

theano/tensor/tests/test_opt.py
lamblin: Because it makes it possible to combine the dimshuffle with another one, or lift it further.
These optimizations were identified to help simplify the case in https://github.com/Theano/Theano/issues/2943#issuecomment-184453452, where operations like, alloc, dimshuffle, and reshape do more than one thing, which can all cancel out if they are properly separated.

It may be possible to add optimizations combining these operations again after the simplification is done, after the 'specialize' phase or so, but that is another issue.

theano/tensor/opt.py
lamblin: That is right. If `output.broadcastable[num_dims_with_size_1_added_to_left:] == input.broadcastable`, then there is no need to call `op(...)`, you can just assign `inner = input`.

mohammadpz: I created a new PR.

Issue #4073, opened 2016-02-16
Replace some Reshape by DimShuffle


lamblin:
Sub-ticket of #2943.
When Reshape only adds broadcastable dimensions, then it could be replaced by a DimShuffle instead.
However, we would still want to be able to catch shape errors, so we could decide to only make the substitution if the shapes can be inferred to be equal.

One of the ways of doing that would be to only take care of the dimensions of length 1, and have another optimization (maybe already existing) take care of removing things like `reshape(x, x.shape)`.

For instance:
- replace `reshape(x, (1, n))` with `dimshuffle{x,0}(reshape(x, (n,))`
- replace `reshape(x, (1, m, 1, n, 1, 1))` by `dimshuffle{x,0,x,1,x,x}(reshape(x, (m, n)))`

nouiz:
I like the idea to spit this this in 2 optimization. Baybe the second one
is already done.

lamblin:
Closed by #4843

PR #4389, opened 2016-04-18
reshape to 1 is replace with dimshuffle

mohammadpz:
Regarding issue #4073 .
Hi,
The PR is not complete yet. It seems that during replacing the node with the optimized node, it gets stuck in an infinite loop. It is not exactly clear to me why it is not the case for ```Alloc``` but it here with ```Reshape``` we have this problem.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3984,8 +3984,10 @@ register_canonicalize(local_reshape_chain(T.Reshape),
 @gof.local_optimizer([T.Reshape])
 def local_useless_reshape(node):
     """
-    Remove Reshape when both the input and the output have a
-    single dimension.
+    Broadcastable dimensions in reshpae are replaced with dimshuffle.
+    For example:
+        - reshape(x, (1, n)) --> dimshuffle{x,0}(reshape(x, (n,))
+        - reshape(x, (1, m, 1, n, 1, 1)) --> dimshuffle{x,0,x,1,x,x}(reshape(x, (m, n)))
 
     """
     op = node.op
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -47,7 +47,7 @@ from theano.tensor.type import (values_eq_approx_remove_inf,
 from theano.gof.opt import (Optimizer, pre_constant_merge,
                             pre_greedy_local_optimizer)
 from theano.gof import toolbox
-from theano.tensor.basic import get_scalar_constant_value, ShapeError, NotScalarConstantError
+from theano.tensor.basic import Reshape, get_scalar_constant_value, ShapeError, NotScalarConstantError
 from six import StringIO
 
 _logger = logging.getLogger('theano.tensor.opt')
@@ -3987,11 +3987,31 @@ def local_useless_reshape(node):
     single dimension.
 
     """
-    if isinstance(node.op, T.Reshape):
-        if (node.inputs[0].ndim == 1 and node.outputs[0].ndim == 1 and
-                node.inputs[0].broadcastable ==
-                node.outputs[0].broadcastable):
-            return [node.inputs[0]]
+    op = node.op
+    if not isinstance(op, Reshape):
+        return False
+
+    input = node.inputs[0]
+    output = node.outputs[0]
+    output_shape = node.inputs[1]
+
+    if (input.ndim == 1 and output.ndim == 1 and
+            input.broadcastable == output.broadcastable):
+        return [input]
+
+    dimshuffle_new_order = []
+    new_output_shape = []
+    i = 0  # index over the output of the new reshape
+    for dim in output_shape.value:
+        if dim == 1:
+            dimshuffle_new_order.append('x')
+        else:
+            dimshuffle_new_order.append(i)
+            new_output_shape.append(dim)
+            i = i + 1
+    if len(dimshuffle_new_order) > 0:
+        inner = op.__class__(len(new_output_shape))(input, new_output_shape)
+        return [DimShuffle(inner.type.broadcastable, dimshuffle_new_order)(inner)]
 
 
 @register_canonicalize
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4002,16 +4002,16 @@ def local_useless_reshape(node):
 
     dimshuffle_new_order = []
     new_output_shape = []
-    i = 0  # index over the output of the new reshape
-    import ipdb; ipdb.set_trace()
-    for dim in extract_constant(output_shape, only_process_constants=True):
+    index = 0  # index over the output of the new reshape
+    for i in xrange(output.ndim):
+        dim = extract_constant(output_shape[i], only_process_constants=False)
         if dim == 1:
             dimshuffle_new_order.append('x')
         else:
-            dimshuffle_new_order.append(i)
+            dimshuffle_new_order.append(index)
             new_output_shape.append(dim)
-            i = i + 1
-    if i != output.ndim:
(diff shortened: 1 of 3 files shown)


theano/tensor/opt.py
```diff
@@ -3987,11 +3987,31 @@ def local_useless_reshape(node):
...
+    for dim in output_shape.value:
+        if dim == 1:
+            dimshuffle_new_order.append('x')
+        else:
+            dimshuffle_new_order.append(i)
+            new_output_shape.append(dim)
+            i = i + 1
+    if len(dimshuffle_new_order) > 0:
+        inner = op.__class__(len(new_output_shape))(input, new_output_shape)
+        return [DimShuffle(inner.type.broadcastable, dimshuffle_new_order)(inner)]
```
lamblin: It looks like, for each existing reshape, you are rebuilding a new Reshape node and adding a dimshuffle, even when the Reshape could not be optimized, or even if it was already the result of this optimization.
The optimization should be skipped if they are no dimensions of length 1.

theano/tensor/opt.py
```diff
@@ -3987,11 +3987,31 @@ def local_useless_reshape(node):
...
+    output_shape = node.inputs[1]
+
+    if (input.ndim == 1 and output.ndim == 1 and
+            input.broadcastable == output.broadcastable):
+        return [input]
+
+    dimshuffle_new_order = []
+    new_output_shape = []
+    i = 0  # index over the output of the new reshape
+    for dim in output_shape.value:
```
lamblin: Again, use `get_scalar_constant_value`.

theano/tensor/opt.py
mohammadpz: line 4012
```python
if len(dimshuffle_new_order) > 0:
```
checks for this. If ```len(dimshuffle_new_order)``` is zero then it means that no reshape of 1 is found.

lamblin: That condition is not correct.
You append `i` to `dimshuffle_new_order` when `dim != 1`, so `len(dimshuffle_new_order)` is always equal to `output.ndim`.

mohammadpz:
The problem I'm facing is that when ```output_shape``` is ```TensorConstant{[1 4]}```, the function ```extract_constant(output_shape, only_process_constants=True)``` returns the same thing. Do you have any idea for the alternative?

lamblin:
Indeed, `only_process_constants` will not work in those cases.
You could use something like the following:
```
for i in xrange(output_shape).ndim:
    dim = extract_constant(output_shape[i], only_process_constants=False)
    if dim == 1:
        ...
```

```

File: theano/tensor/tests/test_opt.py [hunk 3]
```diff
@@ -5938,14 +5939,43 @@ def test_local_reshape(self):
...
+
+        reshape_lift.optimize(g)
+        self.assertTrue(str(g) == "[DimShuffle{x,0}"
+                                  "(<TensorType(float64, vector)>), "
+                                  "DimShuffle{x,0,x,1,x,x}"
+                                  "(Reshape{2}(<TensorType(float64, matrix)>, "
+                                  "TensorConstant{[5 6]}))]")
+
+        # Check stacktrace was copied over correctly after opt was applied
+        self.assertTrue(hasattr(g.outputs[0].tag, 'trace'))
```
lamblin: Here, you should now call `check_stack_trace(g, ops_to_check=[dimshuffle, reshape])` I think.

theano/tensor/opt.py
```diff
@@ -3987,11 +3988,32 @@ def local_useless_reshape(node):
...
+    for i in xrange(output.ndim):
+        dim = extract_constant(output_shape[i], only_process_constants=False)
+        if dim == 1:
+            dimshuffle_new_order.append('x')
+        else:
+            dimshuffle_new_order.append(index)
+            new_output_shape.append(dim)
+            index = index + 1
+    if index != output.ndim:
+        inner = op.__class__(len(new_output_shape))(input, new_output_shape)
```
lamblin: Copy the stack trace from `output` into both `inner` and the variable returned by `Dimshuffle(...)(inner)`.

lamblin:
Otherwise, I've opened a sub-PR to your branch that should fix the remaining failing test in FAST_COMPILE.

theano/tensor/tests/test_basic.py
```diff
@@ -5195,9 +5195,10 @@ def just_vals(v):
         # test broadcast flag for constant value of 1
         c = reshape(b, (b.shape[0], b.shape[1], 1))
         f = self.function([b], c)
+        g = theano.gof.FunctionGraph([b], [c])
         assert numpy.all(f(numpy.asarray([[0, 1, 2], [3, 4, 5]])) ==
                          numpy.asarray([[[0], [1], [2]], [[3], [4], [5]]]))
-        assert (f.maker.fgraph.toposort()[-2].outputs[0].type.broadcastable ==
+        assert (g.toposort()[-1].outputs[0].type.broadcastable ==
```
mohammadpz: Hi @lamblin ,
Does this fix seems fine to you?
I did this to handle when calling with ```mode=FAST_COMPILE```.

lamblin: You do not even need to define `g`.
You can simply replace `-2` by `-1` in the original line, like I did in https://github.com/mohammadpz/Theano/pull/2/files

theano/tensor/tests/test_opt.py
lamblin: I meant instead of that line, sorry. Can you delete the line with the `hasattr(..., 'trace')`?

theano/tensor/opt.py
```diff
@@ -3987,11 +3988,35 @@ def local_useless_reshape(node):
     single dimension.
```
nouiz: Update the docstring

theano/tensor/opt.py
nouiz: I would make the same questions here. Why do we want to make explicitly the DimShuffle when the Reshape isn't removed? When it is removed, I agree this is good, but when we keep it. I don't understand why we need to keep it.

Also, if we end up making the case where we want to have an dimshuffle and a reshape when we keep both, I would special case here the case where we remove the reshape, to not let another opt do the clean up. This will speed up optimization.

lamblin: Two reasons:
- It makes it easier to detect things like `reshape(x, x.shape)` and eliminate the reshape in another optimization, like we discussed in #4073.
- It makes it possible for the dimshuffle to be combined with another one or lifted further.

lamblin:
Remaining things to do on this ticket:
- rebase
- update the docstring of `local_useless_reshape`. You can mention the use cases in the ticket as examples
- remove the `hasattr(..., 'trace')` line.

Thanks,

mohammadpz:
That would be great if we could try to ```rebase``` it together.
Now when I call ```rebase```, I get some conflicts and then I fix them. But when I continue it says ```No changes - did you forget to use 'git add'?```.

lamblin:
I'm at the lab, we can do it IRL if you're here, or by phone or chat.

theano/tensor/opt.py
```diff
@@ -3984,8 +3984,10 @@ def f(node):
 @gof.local_optimizer([T.Reshape])
 def local_useless_reshape(node):
     """
-    Remove Reshape when both the input and the output have a
-    single dimension.
+    Broadcastable dimensions in reshpae are replaced with dimshuffle.
```
lamblin: Keep the previous description as well, we are still doing that.

nouiz:
continued in gh-4843.

PR #4401, opened 2016-04-19
Opt CpuContiguous -> GpuContiguous and add CpuContiguous.ard

nouiz:


theano/sandbox/cuda/opt.py
```diff
@@ -2280,6 +2280,17 @@ def local_gpu_contiguous_gpu_contiguous(node):
...
+@register_opt('fast_compile')
+@local_optimizer([tensor.extra_ops.CpuContiguous])
+def local_gpu_contiguous(node):
+    if isinstance(node.op, tensor.extra_ops.CpuContiguous):
+        x, = node.inputs
+        if x.owner and isinstance(x.owner.op, HostFromGpu):
+            gpu_x, = x.owner.inputs
+            return [tensor.as_tensor_variable(gpu_contiguous(gpu_x))]
+    return False
+
```
lamblin: While you're at it, can you:
- add the optimization the other way, GpuFromHost(CpuContiguous) => GpuContiguous(GpuFromHost)
- add the lifter in the new back-end as well, if it is not already there

nouiz: done.

PR #4402, opened 2016-04-19
Port multinomial to new GPU backend

harmdevries89:
WIP

### DIFF
--- a/theano/sandbox/gpuarray/tests/test_multinomial.py
+++ b/theano/sandbox/gpuarray/tests/test_multinomial.py
@@ -7,7 +7,6 @@ import numpy
 import theano
 from theano import config, function, tensor
 from ..multinomial import GPUAMultinomialFromUniform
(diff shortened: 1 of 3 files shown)


theano/sandbox/gpuarray/tests/test_multinomial.py
```diff
@@ -66,6 +67,7 @@ def body(mode, gpu):
                         for node in f.maker.fgraph.toposort()])
 
         # test that both first and second samples can be drawn
+        print(f([[1, 0], [0, 1]], [.1, .1]))
```
nouiz: Remove the print in the test. We don't want that anymore.

nouiz: it is already done in the next commit.

nouiz:
super-seeded gy gh-4413

PR #4404, opened 2016-04-19
Convert NanGuardMode to use the VM linker instead of WrapLinker.

abergeron:
This removes the check on inputs that NanGuardMode used to do, but I feel it is not a big loss since you can check your function inputs easily.

### DIFF
--- a/theano/compile/nanguardmode.py
+++ b/theano/compile/nanguardmode.py
@@ -211,7 +211,7 @@ class NanGuardMode(Mode):
         assert nan_is_error or inf_is_error or big_is_error
         compile_gpu_func(nan_is_error, inf_is_error, big_is_error)
 
-        def do_check_on(var, nd, f, is_input):
+        def do_check_on(var, nd):
             """
             Checks `var` for NaNs / Infs. If detected, raises an exception
             and / or prints information about `nd`, `f`, and `is_input` to
@@ -223,11 +223,6 @@ class NanGuardMode(Mode):
                 The value to be checked.
             nd : theano.gof.Apply
                 The Apply node being executed.
-            f : callable
-                The thunk for the apply node.
-            is_input : bool
-                If True, `var` is an input to `nd`.
-                If False, it is an output.
 
             """
             error = False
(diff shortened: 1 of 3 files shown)


theano/compile/nanguardmode.py
```diff
@@ -277,36 +264,26 @@ def do_check_on(var, nd, f, is_input):
...
+            thunk : callable
                 The thunk to execute for this Apply node.
+            storage_map : dict
+                The mapping of variables to storage cells.
+            compute_map : dict
+                The mapping to get the computed state of the nodes. (ignored)
 
             """
-            inputs = fn.inputs
-            for x, var in zip(inputs, node.inputs):
```
nouiz: Why did you remove the check on the input of the function?

abergeron: Because the callback is called after the node has run and there might be some inplace stuff that would make it confusing.

Also, this only check the inputs of the graph so I figure it would be easier to check those yourself.

We can bring that back with a bit of modification to StackVM, but I didn't have the time to do that.

nouiz: I don't like regression. So I would wait for this to be fixed before merging.

abergeron: I just pushed a fix.

nouiz:
travis fail.

theano/gof/vm.py
```diff
@@ -378,6 +379,13 @@ def run_thunk_of_node(self, node):
 
         """
         idx = self.node_idx[node]
+        if self.callback_before is not None:
+            self.callback_before(
```
nouiz: When we discussed this, I was more thinking that callback_before would be called before the function start execution, not before each node. This add extra overhead. I don't see a need for that case of callback before each node.

What do you think of that?

abergeron:
I renamed callback_before to callback_input and made be called only on the inputs of the graph.

theano/compile/nanguardmode.py
```diff
@@ -256,18 +251,14 @@ def do_check_on(var, nd, f, is_input):
...
                     error = True
             if error:
-                if not is_input:
-                    print("NanGuardMode found an error in the"
-                          " output of a node in this variable:", file=sio)
+                if nd:
+                    print("NanGuardMode found an error in the "
+                          "output of a node in this variable:", file=sio)
                     print(theano.printing.debugprint(nd, file='str'), file=sio)
+                    msg = sio.getvalue()
```
nouiz: Why this line moved? I don't think it should have moved.

nouiz:
travis fail.

nouiz:
travis fail.

abergeron:
This should be fixed.

theano/gof/vm.py
```diff
@@ -406,6 +407,8 @@ def __call__(self, output_subset=None):
 
         for k in self.storage_map:
             compute_map[k][0] = (k.owner is None)
+            if self.callback_input and compute_map[k][0]:
+                self.callback_input(k, self.storage_map[k][0])
```
nouiz: This behave as the old code. I just taught that this also check constant. I would be fine to only test the inputs of the fgraph and not constant. Otherwise, update the docstring around line 688 to tell this.

abergeron: Done.

lamblin:
We will need to update the doc as well, for instance http://deeplearning.net/software/theano/tutorial/modes.html?highlight=nanguardmode#linkers

abergeron:
I did a git grep in the doc/ directory and this appears to be the only thing that changed.

nouiz:
ok to merge when travis pass.

PR #4406, opened 2016-04-19
Hide the useless default nvcc output on Windows

nouiz:


theano/sandbox/cuda/nvcc_compiler.py
```diff
@@ -404,7 +404,8 @@ def compile_str(
         elif config.cmodule.compilation_warning and nvcc_stdout:
             print(nvcc_stdout)
 
-        if nvcc_stdout:
+        # On Windows, nvcc print useless stuff by default
+        if sys.platform != 'win32' and nvcc_stdout:
             # this doesn't happen to my knowledge
             print("DEBUG: nvcc STDOUT", nvcc_stdout, file=sys.stderr)
```
lamblin: I think we can just remove those lines, or call `_logger.info()` instead of `print`. There is no need to check the platform if the goal is for the message not to be displayed.

nouiz: This is the old back-end. I don't want to spend more time on it.

PR #4419, opened 2016-04-21
Function copy

slefrancois:


theano/compile/tests/test_function_module.py
```diff
@@ -383,6 +383,13 @@ def test_copy_delete_updates(self):
             assert cpy(1)[0] == 4
             assert cpy(1)[0] == 4
 
+        # Test if unused implicit inputs from delete_updates are ignored
+        # as intended.
+        for mode in ["FAST_RUN", "FAST_COMPILE"]:
+            ori = theano.function([x], x, mode=mode, updates={z: z * 2})
+            cpy = ori.copy(delete_updates=True)
+
```
nouiz: remove an empty line. This is what make travis fail.

nouiz: Can you add this extra test that make one explicit input useless
```
             ori = theano.function([x, y], x, mode=mode, updates={z: z + y})
             cpy = ori.copy(delete_updates=True)
```

theano/compile/function_module.py
```diff
@@ -714,7 +714,13 @@ def checkSV(sv_ori, sv_rpl):
...
         f_cpy = maker.__class__(inputs=ins, outputs=outs, fgraph=fg_cpy,
                                 mode=maker.mode, profile=profile,
-                                on_unused_input=maker.on_unused_input,
+                                # When removing updates containing variables
+                                # not used in the output function, copy
+                                # generates an unused implicit input.
+                                # We ignore the resulting errors,
+                                # but could change it to 'warn' if this might
+                                # cause problems.
+                                on_unused_input='ignore',
```
nouiz: This is good and safe. But ideally, we should remove the shared variable from inputs. For now, we can keep it like this.

theano/compile/tests/test_function_module.py
slefrancois: Does it make the first test redundant, or do they cover separate cases?

nouiz: They cover different case.

The current test make a shared variable input as non needed anymore. The
new test make an explicit input not needed anymore.

The second case make it more difficult to remove the inputs that are not
needed, as we can't get rid of that input as it would change the user
interface.

slefrancois: Ok, I pushed the test. I had to use a new variable w, with y it complained
that you can use shared variables as inputs.

Also I dit "git push" without arguments by accident, there might be two
similar commits because of that.

On Fri, Apr 22, 2016 at 10:10 AM, Frédéric Bastien <notifications@github.com

nouiz:
ok to merge when travis pass.

nouiz:
travis failed.

slefrancois:
Ok, flake8 passes for the files on my editor, do the changes break
something else? We can talk about it monday I guess.

doc/tutorial/examples.txt
```diff
@@ -285,6 +282,54 @@ expression that evaluates to a tensor of same shape and dtype.
...
+
+>>> import theano
+>>> import theano.tensor as T
+>>> state = theano.shared(0)
+>>> inc = T.iscalar('inc')
+>>> accumulator = theano.function([inc], state, updates=[(state, state+inc)])
+
+We can use it to increment the state as usual:
+
+>>> accumulator(10)
```
lamblin: So, a plugin for the documentation is treating those inline examples as unit tests to be checked.
In fact, it is running these lines, and comparing the result with the expected result that it tries to find on the next.
In that case, `accumulator(10)` returns `array(0)`, but since there is nothing on the next line, it is treated as a failure (it figures that call should return `None`).

I think this is the same for the next failures as well.

slefrancois:
I added the missing outputs, doc tests fine on my machine now.

PR #4420, opened 2016-04-22
Small stuff

nouiz:
- Make debugprint print the clients information
- doc
- error message
- simplification

### DIFF
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -575,7 +575,7 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
     if prefix_child is None:
         prefix_child = prefix
 
-    def get_id_str(obj):
+    def get_id_str(obj, get_printed=True):
         if obj in done:
             id_str = done[obj]
         elif ids == "id":
@@ -586,7 +586,8 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
             id_str = "[id %s]" % char_from_number(len(done))
         elif ids == "":
             id_str = ""
-        done[obj] = id_str
+        if get_printed:
+            done[obj] = id_str
 
         return id_str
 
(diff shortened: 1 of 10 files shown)


theano/compile/debugmode.py
```diff
@@ -512,7 +512,8 @@ def debugprint(r, prefix='', depth=-1, done=None, print_type=False,
                print_view_map=False, order=None, ids='CHAR',
                stop_on_name=False, prefix_child=None,
                scan_ops=None, profile=None,
-               scan_inner_to_outer_inputs=None, smap=None):
+               scan_inner_to_outer_inputs=None, smap=None,
+               used_ids=None):
```
lamblin: Add doc for that new parameter.

theano/tensor/opt.py
```diff
@@ -1578,7 +1578,7 @@ def dimshuffled_alloc(i):
 def local_fill_sink(node):
     """
     f(fill(a, b), fill(c, d), e) -> fill(a, fill(c, f(b, d, e)))
-
+    f(fill(a, b), fill(c, d), e) -> fill(c, fill(a, f(b, d, e)))
```
lamblin: Is the new line supposed to replace the previous one?

theano/tests/test_printing.py
```diff
@@ -270,6 +271,26 @@ def test_debugprint():
...
+    s = s.getvalue()
+    # The additional white space are needed!
+    reference = '\n'.join([
+"Elemwise{add,no_inplace} [id A] ''   0 clients:[('output', ''), ('[id C]', 1)]",
+" |A [id D]",
+" |B [id E]",
+"Elemwise{sub,no_inplace} [id C] ''   1",
+" |Elemwise{add,no_inplace} [id A] ''   0 clients:[('output', ''), ('[id C]', 1)]",
+" |D [id F]",
+    ]) + '\n'
```
lamblin: Are you changing the default appearance of `debugprint`? If so, the documentation and examples should be updated everywhere.

nouiz: Right. I don't want to change the default, so I updated this PR to make this an option.

PR #4422, opened 2016-04-22
Develop SearchsortedOp to wrap numpy's searchsorted function

tsirif:
Includes:  
* A `SearchsortedOp`, which includes `perform` and `c_code`
* A `searchsorted` function with numpy-equivalent call
* A method in tensor's basic methods which redirects to `searchsorted` with numpy-equivalent call to ndarray's method
* Unit tests for `SearchsortedOp`'s functionalities

Currently, `side` string argument is being given to SearchSortedOp as an init argument. A param version is not implemented in this patch.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -147,6 +147,8 @@ def as_tensor_variable(x, name=None, ndim=None):
         If an `Apply` with more than one output is fetched.
     AsTensorError
         If `x` cannot be converted to a TensorType Variable.
+    TypeError
+        If `x` cannot be made into a Variable with `ndim` dimensions.
 
     """
     if hasattr(x, '_as_TensorVariable'):
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -138,17 +138,15 @@ def as_tensor_variable(x, name=None, ndim=None):
         If a new `Variable` instance is created, it will be named with this
         string.
     ndim : None or integer
-        Return a Variable with this many dimensions. Raise TypeError if it's
-        not possible.
+        Return a Variable with this many dimensions.
 
     Raises
     ------
     ValueError
-        If an `Apply` with more than one output is fetched.
+        If an `Apply` with more than one output is fetched or
+        if `x` cannot be made into a Variable with `ndim` dimensions.
     AsTensorError
         If `x` cannot be converted to a TensorType Variable.
-    TypeError
-        If `x` cannot be made into a Variable with `ndim` dimensions.
 
     """
     if hasattr(x, '_as_TensorVariable'):
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -68,6 +68,132 @@ class CpuContiguous(theano.Op):
 cpu_contiguous = CpuContiguous()
 
 
+class SearchsortedOp(theano.Op):
+    """Wrapper of numpy.searchsorted.
+
+    For full documentation, see :func:`searchsorted`.
+
+    See Also
+    --------
+    searchsorted : numpy-like function to use the SearchsortedOp
+
+    """
+
+    __props__ = ("side", )
+
+    def __init__(self, side='left'):
+        self.side = side
+
+    def make_node(self, x, v, sorter=None):
+        x = basic.as_tensor(x, ndim=1)
+        v = basic.as_tensor(v)
+        out_type = v.type.clone(dtype='int64')
+        if sorter is None:
+            return theano.Apply(self, [x, v], [out_type()])
+        else:
+            sorter = basic.as_tensor(sorter, ndim=1)
+            if sorter.type not in basic.int_vector_types:
+                raise TypeError('sorter must be an integer vector',
+                                sorter.type)
+            return theano.Apply(self, [x, v, sorter], [out_type()])
+
+    def infer_shape(self, node, shapes):
+        return [shapes[1]]
+
+    def perform(self, node, inputs, output_storage):
+        x = inputs[0]
+        v = inputs[1]
+        if len(node.inputs) == 3:
+            sorter = inputs[2]
+        else:
+            sorter = None
+        z = output_storage[0]
+
+        z[0] = np.searchsorted(x, v, side=self.side, sorter=sorter)
+
+    def c_code(self, node, name, inames, onames, sub):
+        sorter = None
+        if len(node.inputs) == 3:
+            x, v, sorter = inames
+        else:
+            x, v = inames
+        if not sorter:
+            sorter = "NULL"
+        z, = onames
+        side = "NPY_SEARCHRIGHT" if self.side == 'right' else "NPY_SEARCHLEFT"
+        fail = sub['fail']
+
+        return """
+            Py_XDECREF(%(z)s);
+            %(z)s = (PyArrayObject*) PyArray_SearchSorted(%(x)s, (PyObject*) %(v)s,
+                                                          %(side)s, (PyObject*) %(sorter)s);
+            if (!%(z)s)
+                %(fail)s;
+        """ % locals()
+
+    def c_code_cache_version(self):
+        return (0, 1, 2)
+
+    def grad(self, inputs, output_gradients):
+        num_ins = len(inputs)
+        if num_ins == 3:
+            x, v, sorter = inputs
+        else:
+            x, v = inputs
+
+        x_grad = x.zeros_like()
+        if v.ndim == 1:
+            v_grad = v.zeros_like()
+        else:
+            v_grad = theano.gradient.grad_not_implemented(
+                    self, 1, v, "Grad is not implemented for inputs with "
+                                "number of dimension other than 1.")
+        if num_ins == 3:
+            sorter_grad = theano.gradient.grad_undefined(
+                    self, 2, sorter,
+                    "searchsorted is not defined for non-integer sorter so "
+                    "searchsorted(x, nb, sorter+eps), for eps > 0, "
+                    "is undefined")
+            return [x_grad, v_grad, sorter_grad]
+        else:
+            return [x_grad, v_grad]
+
+
+def searchsorted(x, v, side='left', sorter=None):
+    """Find indices where elements should be inserted to maintain order.
+
+    Wraping of numpy.searchsorted. Find the indices into a sorted array
+    `x` such that, if the corresponding elements in `v` were inserted
+    before the indices, the order of `x` would be preserved.
+
+    Parameters
+    ----------
+    x: 1-D tensor (array-like)
+        Input array. If `sorter` is None, then it must be sorted in
+        ascending order, otherwise `sorter` must be an array of indices
+        that sort it.
+    v: tensor (array-like)
+        Contains the values to be inserted into `x`.
+    side: {'left', 'right'}, optional.
+        If 'left' (default), the index of the first suitable
+        location found is given. If 'right', return the last such index. If
+        there is no suitable index, return either 0 or N (where N is the length
+        of `x`).
+    sorter: 1-D tensor of integers (array-like), optional
+        Contains indices that sort array `x` into ascending order.
+        They are typically the result of argsort.
+
+    .. versionadded:: 0.8.2
+
+    Returns
+    -------
+    indices : tensor of integers (int64)
+        Array of insertion points with the same shape as `v`.
+
+    """
+    return SearchsortedOp(side=side)(x, v, sorter)
+
+
 class CumsumOp(theano.Op):
     # See function cumsum for docstring
 
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -8,7 +8,9 @@ import theano
 from theano.tensor import basic
 from theano.tensor import nlinalg  # noqa
 from theano import gof, scalar
-from theano.gradient import DisconnectedType
+from theano.gof import Generic
+from theano import gradient
+from theano.gradient import DisconnectedType, disconnected_type
 tensor = basic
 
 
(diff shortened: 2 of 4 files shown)


theano/tensor/basic.py
```diff
@@ -147,6 +147,8 @@ def as_tensor_variable(x, name=None, ndim=None):
         If an `Apply` with more than one output is fetched.
     AsTensorError
         If `x` cannot be converted to a TensorType Variable.
+    TypeError
+        If `x` cannot be made into a Variable with `ndim` dimensions.
```
nouiz: I get a value error in that case:
```
>>> import theano
>>>v=theano.tensor.matrix()
>>> theano.tensor.as_tensor_variable(v, ndim=1)
  File "theano/tensor/basic.py", line 182, in as_tensor_variable
    % ndim, x.type
ValueError: ('TensorType could not be cast to have 1 dimensions', TensorType(float64, matrix))
```

In which case you get that error?

theano/tensor/extra_ops.py
```diff
@@ -65,6 +65,132 @@ def c_code_cache_version(self):
...
+    side: {'left', 'right'}, optional.
+        If 'left' (default), the index of the first suitable
+        location found is given. If 'right', return the last such index. If
+        there is no suitable index, return either 0 or N (where N is the length
+        of `x`).
+    sorter: 1-D tensor of integers (array-like), optional
+        Contains indices that sort array `x` into ascending order.
+        They are typically the result of argsort.
+
+    .. versionadded:: 0.8.2
```
nouiz: will be added in 0.9

theano/tensor/extra_ops.py
```diff
@@ -65,6 +65,132 @@ def c_code_cache_version(self):
...
+
+    Returns
+    -------
+    indices : tensor of integers (int64)
+        Array of insertion points with the same shape as `v`.
```
nouiz: add a link to numpy doc for the examples, or copy the examples.

nouiz: add a note that it only work on CPU.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -33,6 +35,123 @@ def test_cpu_contiguous():
...
+    def tearDown(self):
+        self.x = None
+        self.v = None
+        self.a = None
+        self.b = None
+        self.idx_sorted = None
+
+    def test_searchsortedOp_on_sorted_input(self):
+        f = theano.function([self.x, self.v], searchsorted(self.x, self.v),
+                            mode="DebugMode")
```
nouiz: don't put a mode here. In our daily buildbot, we test with many different modes.

theano/tensor/extra_ops.py
```diff
@@ -65,6 +65,132 @@ def c_code_cache_version(self):
...
+        if num_ins == 3:
+            x, v, sorter = inputs
+        else:
+            x, v = inputs
+
+        x_grad = x.zeros_like()
+        if v.ndim == 1:
+            v_grad = v.zeros_like()
+        else:
+            v_grad = theano.gradient.grad_not_implemented(
```
nouiz: why ndim > 1 is different? I would think it is the same.

@lamblin can you confirm this method is well implemented? I think so but would like a second opinion.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -33,6 +35,123 @@ def test_cpu_contiguous():
...
+
+    def test_searchsortedOp_on_none_sorter(self):
+        # Current implementation of numpy.searchsorted
+        # does not raise an error if `x` is not sorted and sorter is None.
+        sorter = T.vector('sorter', dtype="int64")
+        f = theano.function([self.x, self.v, sorter],
+                            searchsorted(self.x, self.v, sorter=sorter))
+        #  assert np.allclose(
+        #          np.searchsorted(self.a, self.b, sorter=None),
+        #          f(self.a, self.b, sorter=None))
```
nouiz: Here, you must pass a vector for the sorter input. That is why it raise an error.

nouiz: In fact, this whole test isn't useful. It is the Theano function that will parse the input and will reject the None as it expect a vector. So this don't test this new op. I would just remove this test method.

theano/tensor/basic.py
tsirif: In the params doc above, it says:
"""
ndim : None or integer
      Return a Variable with this many dimensions. Raise TypeError if it's
      not possible.
"""

I'm looking into it

theano/tensor/tests/test_extra_ops.py
```diff
@@ -33,6 +35,123 @@ def test_cpu_contiguous():
...
+
+    def test_searchsortedOp_on_right_side(self):
+        f = theano.function([self.x, self.v],
+                            searchsorted(self.x, self.v, side='right'),
+                            mode="DebugMode")
+        assert np.allclose(
+                np.searchsorted(self.a, self.b, side='right'),
+                f(self.a, self.b))
+
+    def test_use_c_code(self):
```
nouiz: We test the Python and C code in travis and in the daily buildbot. So no need for this test method.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -33,6 +35,123 @@ def test_cpu_contiguous():
...
+
+        # Test parameter ``sorter``
+        sorter = T.vector('sorter', dtype="int64")
+        self._compile_and_check([self.x, self.v, sorter],
+                                [searchsorted(self.x, self.v, sorter=sorter)],
+                                [self.a, self.b, self.idx_sorted],
+                                self.op_class)
+
+        # Test parameter ``side``
+        self.a = np.ones(10).astype(config.floatX)
```
nouiz: I'm not sure if it is garanty the tests a run in a fixed order. To be safe, don't override this instance attribute in the test.

nouiz:
This PR look pretty good. I just got a few questions/comments.

thanks

tsirif:
Thank you Fred. I am working on these comments right now. 2 questions:
1. does `side` argument need to be a dynamic op param or not?  
2. About ndim > 1 in grad method: Documentation refers that handling grads of non vector inputs is not implemented yet.

nouiz:
side don't need to be a dinamic param. But it can be. You can wrap the
string in a Generic() variable.

The other option is to use params.

Both change would allow to compile less version of this code. Otherwise, we
need to compile 2 versions of the code.

This isn't a must do to merge this PR, but if we could compile just one
version, it would be great.

2) I'm not sure what you are talking about. Here the grad implementation
would be the same as for 1d, zeros(). Where did you saw this?

tsirif:
1) Don't merge yet, I will fix what you commented and also I will make `side` a dynamic param for practice.

2) I cannot find it right now. Maybe it was a misconception of mine. I will fix this.

3) Thanks for accepting me in GSoC! :smiley:

theano/tensor/extra_ops.py
lamblin: `x_grad` should be `x.zeros_like()` only if `x` has a float-like `dtype`, and `x.zeros_like(floatX)` for int-like `dtype` (gradients have to be float). There should be a helper function doing that somewhere.
Same for `v_grad`, it should be `v.zeros_like(...)` regardless of the number of dimensions.

`sorter_grad` should be `DisconnectedType()` .

theano/tensor/extra_ops.py
```diff
@@ -65,6 +65,132 @@ def c_code_cache_version(self):
...
+                    "is undefined")
+            return [x_grad, v_grad, sorter_grad]
+        else:
+            return [x_grad, v_grad]
+
+
+def searchsorted(x, v, side='left', sorter=None):
+    """Find indices where elements should be inserted to maintain order.
+
+    Wraping of numpy.searchsorted. Find the indices into a sorted array
```
lamblin: "Wrapping"

theano/tensor/extra_ops.py
```diff
@@ -65,6 +65,132 @@ def c_code_cache_version(self):
...
+    `x` such that, if the corresponding elements in `v` were inserted
+    before the indices, the order of `x` would be preserved.
+
+    Parameters
+    ----------
+    x: 1-D tensor (array-like)
+        Input array. If `sorter` is None, then it must be sorted in
+        ascending order, otherwise `sorter` must be an array of indices
+        that sort it.
```
lamblin: "sorts"

theano/tensor/extra_ops.py
tsirif: oups.

theano/tensor/extra_ops.py
tsirif: So, would `x.zeros_like(dtype=floatX)` suffice for both cases? (Same for `v`)

tsirif:
Ok, I think I completed changes according to your comments and I also added `side` as a dynamic param.

theano/tensor/extra_ops.py
lamblin: Not really, since `x` could still be `float64` when `floatX == 'float32'`, or the other way around. In that case, we really want `x.dtype`, not `floatX`.

lamblin:
General note: could you rebase your changes on top of the master branch, rather than merging with it? Merging brought in unrelated changes in this PR.
Also, on test is failing, it looks like numerical gradient estimation went across a point where the gradient is not defined, causing it to expect a huge gradient.

tsirif:
@lamblin, what should be done on the failing test?

tsirif:
Also, you mean that I should not merge with no-ff into the branch on which I will issue a pull request, right? You do not want e.g. `Merge branch 'feature/searchsorted' into develop` commits into history tree. I commit frequently though small partial changes which may not be complete code features to be run through ci yet and I do not want to spam hooks on travis with every push. Should I not care about this? Should I use `[ci skip]` in commit messages or do you suggest something else?

lamblin:
For the failing test, I would suggest sampling fewer values, or spanning a larger range, to avoid almost-collisions in the values.
For the merge, that is right, you should avoid *merging* with `central/master` or `upstream/master` while developing. If there is a conflict or that you need to update, it is better to use `git rebase -i central/master` instead. This will "lift" your commits on top of the latest master commits and apply them there.
You can have partial local commits and push once in a while, I've never used `[ci skip]` but it may be a good idea.

tsirif:
@lamblin : So far I used to have a `feature/<something>` branch in which I was working and when then the next x commits in this branch had completed a code feature, I was merging a `develop` branch with `--no-ff` to `feature/<something>`. `develop` branch is the one that I issue for a pull request, so every time a new merging commit is pushed to the repo, a new necessary hook on travis begins. If I keep rebasing this subtree on top of most recent `theano/master` branch by the time I want to push on `develop`, would this be a desirable development scheme?

tsirif:
I rebased on top of current `theano/master` and squashed commits into a single commit for every merge I had done in my `develop` from `feature/searchsorted` branch. The build is passing successfully.

nouiz:
You can rebase on top of the central/master when needed. But you don't need to always do it. Most of the time it isn't needed.

But when a rebase/merge is needed, make a rebase as you did here. I'll check again this PR.

theano/tensor/extra_ops.py
```diff
@@ -68,6 +70,164 @@ def c_code_cache_version(self):
...
+    --------
+    searchsorted : numpy-like function to use the SearchsortedOp
+
+    """
+
+    params_type = Generic()
+    __props__ = ("side", )
+
+    def __init__(self, side='left'):
+        self.side = side
```
nouiz: Add an assert that side is "left" or "right".

theano/tensor/tests/test_extra_ops.py
```diff
@@ -37,6 +38,80 @@ def test_cpu_contiguous():
...
+        self.op = SearchsortedOp()
+
+        self.x = T.vector('x')
+        self.v = T.tensor3('v')
+
+        self.a = 30 * np.random.random(50).astype(config.floatX)
+        self.b = 30 * np.random.random((8, 10, 5)).astype(config.floatX)
+        self.idx_sorted = np.argsort(self.a)
+
+    def tearDown(self):
```
nouiz: I don't understand why this is needed. We don't do it elsewhere. Is there a particular reason you do this? I think by default all member get removed and Python gc would collect them afterward. But if you know they are needed, please explain. I never check my hypothesis.

theano/tensor/var.py
```diff
@@ -692,6 +692,9 @@ def cumsum(self, axis=None):
     def cumprod(self, axis=None):
         return theano.tensor.extra_ops.cumprod(self, axis)
 
+    def searchsorted(self, v, side='left', sorter=None):
```
nouiz: This isn't used in your test. Just change one test to call it by this way. This make sure we won't break/remove that interface by mistake.

nouiz:
I got very small/comments questions, so we should be able to merge this soon.

thanks.

theano/tensor/var.py
tsirif: :+1:

theano/tensor/tests/test_extra_ops.py
tsirif: You are right, this is redundant. I will remove it.

theano/tensor/extra_ops.py
tsirif: Maybe I should let numpy's exception to be raised. I will change the c_code implementation to allow this.

nouiz: It is better to return the error when we build the graph when possible then during execution. So I would add the check here.

tsirif:
@nouiz I pushed fixes

MarcCote:
@tsirif well done. :+1:

Issue #4360, opened 2016-04-11
GPU implementation of Diagonal


bouthilx:
When #4169 is merged, theano.tensor.basic.Diagonal will now run on the GPU for default properties: offset, axis1 and axis2. We should modify perform to support non-default values on the GPU, or write a GpuOp if it is not possible to handle non-default values with python code.

hantek:
So question number one for the non-default properties:

We need numpy.transpose() to manipulate high-dim tensors, is there such a counterpart in CudaNdarray? I find "reshape" from dir(CudaNdarray) but that is not what we need.

bouthilx:
I doubt we can do it with pure python. We will probably need to use pyCUDA or pure CUDA.

hantek:
Could you point me some example codes of pyCUDA? Haven't written that before, but I know CUDA.

abergeron:
Since it's been a while, the CudaNdarray implementation is less relevant.

We want one for the new gpuarray backend.  This has support for numpy-like transpose(), but you can't assign to the strides, so you might have to use c_code there.

abergeron:
See #5247

PR #4424, opened 2016-04-23
gpu diagonal implementation

hantek:
This PR is part of PR #4169 . It tries to solve issue #4360 .

I didn't test it on GPU, just add an outline in the first commit. So it should be super buggy for the moment.


theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
     def c_code_cache_version(self):
         return (3,)
 gpu_eye = GpuEye(dtype='float32')
+
+
+class GpuDiagonal(GpuOp):
+    __props__ = ("offset", "axis1", "axis2")
```
bouthilx: view is missing

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+    default_offset = 0
```
bouthilx: No need for this anymore

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+    default_axis1 = 0
+    default_axis2 = 1
+
+    def __init__(self, offset=0, axis1=0, axis2=1):
```
bouthilx: view is missing

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+        self.view = view
+        if self.view and not numpy_diagonal_return_view:
```
bouthilx: We don't care about numpy_diagonal_return_view inside the GpuOp as it doesn't use numpy.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+                          numpy.version.version)
+            self.view = False
+        if self.view:
+            self.view_map = {0: [0]}
+        self.offset = offset
+        self.axis1 = axis1
+        self.axis2 = axis2
+
+    def make_node(self, _x):
+        if not isinstance(_x, theano.Variable):
```
bouthilx: No need for if-else here. We always want x to be as_cuda_ndarray_variable.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+            axis_with_smaller_shape = self.axis1
+        else:
+            axis_with_bigger_shape = self.axis1
+            axis_with_smaller_shape = self.axis2
+
+        slicer = [numpy.s_[:], ] * x.ndim
+        slicer[axis_with_bigger_shape] = 0  # self.offset
+        slicer = tuple(slicer)
+        if axis_with_smaller_shape > axis_with_bigger_shape:
+            axis_with_smaller_shape -= 1
```
bouthilx: Why -1 here? Are axis1 and axis2 supposed to be consecutive? I believe not.

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+
+        new_dim_order = range(x[slicer].ndim)
```
bouthilx: A small comment here would make it easier to understand what is going on. 

> We swap axis1 and axis2 because...

theano/sandbox/cuda/basic_ops.py
```diff
@@ -4288,3 +4289,100 @@ def c_code(self, node, name, inp, out, sub):
...
+        new_dim_order[axis_with_smaller_shape], new_dim_order[-1] = \
+            new_dim_order[-1], new_dim_order[axis_with_smaller_shape]
+        rval = cuda_dimshuffle(x[slicer], new_dim_order)
+        
+        other_strides = tuple([d for i, d in enumerate(x.strides)
```
bouthilx: A small comment here too would help.

> We change the stride such that `rval` becomes a view of the diagonal.

theano/tensor/basic.py
```diff
@@ -5902,24 +5902,74 @@ class Diagonal(Op):
 
     """
     __props__ = ("offset", "axis1", "axis2")
+    default_offset = 0
```
bouthilx: No need for this anymore

theano/tensor/basic.py
```diff
@@ -5902,24 +5902,74 @@ class Diagonal(Op):
...
+    default_axis1 = 0
+    default_axis2 = 1
 
     def __init__(self, offset=0, axis1=0, axis2=1):
```
bouthilx: view is missing

theano/tensor/basic.py
```diff
@@ -5902,24 +5902,74 @@ class Diagonal(Op):
 
     """
     __props__ = ("offset", "axis1", "axis2")
```
bouthilx: view is missing

theano/tensor/basic.py
```diff
@@ -5902,24 +5902,74 @@ class Diagonal(Op):
...
         self.axis1 = axis1
         self.axis2 = axis2
 
-    def make_node(self, x):
-        x = as_tensor_variable(x)
-        assert x.ndim >= 2
-        return Apply(self, [x], [tensor(dtype=x.dtype,
-                                        broadcastable=[False] * (x.ndim - 1))])
+    def make_node(self, _x):
+        if not isinstance(_x, theano.Variable):
```
bouthilx: No need for if-else here neither. x should always be as_tensor_variable.

theano/tensor/basic.py
```diff
@@ -5902,24 +5902,74 @@ class Diagonal(Op):
...
+                             'dimensions', x)
+        return Apply(self, [x], [x.type.__class__(
+            dtype=x.dtype,
+            broadcastable=[False] * (x.ndim - 1))()])
 
     def perform(self, node, inputs, outputs):
         (x,) = inputs
         (z,) = outputs
-        z[0] = x.diagonal(self.offset, self.axis1, self.axis2)
+        # zero-dimensional matrices ...
```
bouthilx: Is there any usefulness for not using x.diagonal here? As the gpu implementation is now in GpuDiagonal, I would simply use numpy.diagonal now for simplicity's sake. Unless I am not aware of something GpuDiagonal support and numpy.diagonal doesn't?

bouthilx: Just don't forget to do rval.copy() if we use numpy.diagonal and self.view is False

theano/sandbox/cuda/basic_ops.py
hantek: It's because the dimension at `axis_with_bigger_shape` will be swiped away at line 4356 `x[slicer]`. So if `axis_with_bigger_shape` is listed in front of `axis_with_smaller_shape`, the latter one will need `-1`. 

Otherwise, if `axis_with_bigger_shape` is behind `axis_with_smaller_shape`, swiping away that dimension will have no effect on `axis_with_smaller_shape`.

theano/sandbox/cuda/basic_ops.py
hantek: should view be appearing here? Should it be a __props__?

bouthilx: Yes it is. When we compare two ```GpuDiagonal```s, if one has ```view=False``` and the other has ```view=True```, then they have different behavior and should be considered different by ```__eq__()```

theano/sandbox/cuda/basic_ops.py
bouthilx: Oh wow, I was confused. :sweat_smile: This is definitely complicated just to fetch a subarray on index 0 and swap two axis. :wink:

Do you see anything inefficient in this? 

```python
middle_axis = [axis for axis in xrange(len(a.shape))
               if axis != axis1 and axis != axis2]

new_dim_order = [large_axis] + middle_axis + [small_axis]

diag = cuda_dimshuffle(x, new_dim_order)[0]
```

It does the same thing but seems more readable (though I have a strong bias having wrote the code  :blush:).

theano/tensor/basic.py
nouiz: Historically NumPy diagonal was always returning a copy.
Le 23 avr. 2016 11:51, "Xavier Bouthillier" <notifications@github.com> a
écrit :

bouthilx: Until 1.7 and then is was a view, right? 

http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.diagonal.html

We will keep the code that verifies if numpy's diagonal is a copy or a view, and force self.view=False in the first case.

hantek: Ah yes, we should test if it returns numpy view, and do diagonal
accordingly.

On Tuesday, 26 April 2016, Xavier Bouthillier <notifications@github.com>
wrote:

> In theano/tensor/basic.py
> <https://github.com/Theano/Theano/pull/4424#discussion_r61189011>:
>
> >
> >      def perform(self, node, inputs, outputs):
> >          (x,) = inputs
> >          (z,) = outputs
> > -        z[0] = x.diagonal(self.offset, self.axis1, self.axis2)
> > +        # zero-dimensional matrices ...
>
> Until 1.7 and then is was a view, right?
>
>
> http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.diagonal.html
>
> We will keep the code that verifies if numpy's diagonal is a copy or a
> view, and force self.view=False in the first case.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/Theano/Theano/pull/4424/files/c1a864ccf486d74df2625c15a67ded78c5b85cdf#r61189011>
>


-- 
LIN, Zhouhan
Department of Computer Science and Operational Research
University of Montreal

Phone: +01-514-586-0551
Email: lin.zhouhan@gmail.com
Homepage:http://zhouhanlin.webs.com/

bouthilx: The test is done at line 5911 to set view accordingly. You just need to add 

```python
rval = x.diagonal(self.offset, self.axis1, self.axis2)
if self.view:
    z[0] = rval
else:
    z[0] = rval.copy()
```

nouiz:
travis have at least this error.
/home/travis/build/Theano/Theano/theano/tensor/basic.py:5912:63: W291 trailing whitespace

theano/sandbox/cuda/basic_ops.py
```diff
@@ -12,6 +12,7 @@
 from theano import tensor, scalar, config
 from theano.gradient import grad_undefined
 from theano.scalar import Scalar
+from theano.sandbox.cuda import dimshuffle as cuda_dimshuffle
```
nouiz: Don't do that. This cause import problem. Remove that line and when you want to use cuda_dimshuffle, just use:
cuda_ndarray.cuda_ndarray.dimshuffle(...) instead.

nouiz:
Don't forget to make a comment when adding new commit. I see them faster that way.

Travis have those errors:

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4325:1: W293 blank line contains whitespace

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4346:1: W293 blank line contains whitespace

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4351:24: E127 continuation line over-indented for visual indent

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4361:17: F821 undefined name 'grad_not_implemented'

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4372:25: F821 undefined name 'clip'

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4374:25: F821 undefined name 'clip'

/home/travis/build/Theano/Theano/theano/sandbox/cuda/basic_ops.py:4376:25: F821 undefined name 'minimum'

theano/tensor/basic.py
```diff
@@ -5901,20 +5901,32 @@ class Diagonal(Op):
...
     """
-    __props__ = ("offset", "axis1", "axis2")
-
-    def __init__(self, offset=0, axis1=0, axis2=1):
-        if numpy_diagonal_return_view:
+    __props__ = ("offset", "axis1", "axis2", "view")
+
+    def __init__(self, offset=0, axis1=0, axis2=1, view=False):
+        self.view = view
+        if self.view and not numpy_diagonal_return_view:
```
nouiz: I don't find where numpy_diagonal_return_view is defined in this PR.

nouiz:
There is a new file added that should not be there.

theano/tensor/basic.py
bouthilx: Line 5886

nouiz:
I don't have time to review it in detail, but travis is failing:

/home/travis/build/Theano/Theano/theano/sandbox/cuda/tests/test_basic_ops.py:1318:13: E129 visually indented line with same indent as next logical line

Also this PR add a file that should not be there.

theano/sandbox/cuda/tests/test_basic_ops.py
```diff
@@ -1285,6 +1285,33 @@ def check(dtype, N, M_=None, K=0):
...
+        np_x = numpy.arange(30107).reshape(7, 11, 17, 23).astype('float32')
+        for offset, axis1, axis2 in [
+            (1, 0, 1), (-1, 0, 1), (0, 1, 0), (-2, 1, 0),
+            (-3, 1, 0), (-2, 2, 0), (3, 3, 0), (-1, 3, 2),
+            (2, 2, 3), (-1, 2, 1), (1, 3, 1), (-1, 1, 3)]:
+            assert numpy.allclose(
+                B.GpuDiagonal(offset, axis1, axis2)(x).eval({x: np_x}),
+                np_x.diagonal(offset, axis1, axis2))
+
+    def test_grad(self):
```
nouiz: Don't implement that method if it isn't there.

theano/sandbox/cuda/tests/test_basic_ops.py
```diff
@@ -1285,12 +1285,31 @@ def check(dtype, N, M_=None, K=0):
...
 
-    """
-    Ensure the `size` attribute of CUDA tensors behaves as in numpy.
-    """
 
+class test_size(unittest.TestCase):
+    """Ensure the `size` attribute of CUDA tensors behaves as in numpy."""
     def test_matrix(self):
         x = cuda.fmatrix()
         y = numpy.zeros((5, 7), dtype='float32')
```
hantek: What is the right format for this line? flake 8 is complaining:

/home/travis/build/Theano/Theano/theano/sandbox/cuda/tests/test_basic_ops.py:1318:13: E129 visually indented line with same indent as next logical line

But when I was running locally the flake 8 doesn't complain about that. I don't think t is wrong.

hantek:
Hi, I think the PR is OK for now, could anyone review it?

abergeron:
There is no optimization to replace the CPU diagonal with the new op.

Also the view parameter is probably superfluous.  We can make it a view all the time.

Finally this should also be in the new backend (or maybe even only be in the new backend).

abergeron:
Was done for the new backend.

Issue #4055, opened 2016-02-15
flake8 gof


nouiz:

fvisin:
This ticket is to fix all the PEP8 errors. This includes making sure that all the functions and classes have comprehensible and exhaustive docstrings.

PR #4283, opened 2016-03-22
flake8 gof

chinnadhurai:
#4055
@adri-romsor
@nouiz

Comments:
The following functions are deprecated
- toolbox.py / NodeFinder
- utils.py / type_guard
- cc.py / get_persistent_module_cache

PR #4428, opened 2016-04-24
CCW 4283

chinnadhurai:
continuation of PR gh-4283
 
@adri-romsor
@nouiz

Comments:
fix gh-4055 
Fixed all the pep-8 errors in /gof folder except for unify.py which don't seem to be pep-8 related.


nouiz:
Just a note that travis is failing with at least:

```
    from theano.gof.op import CLinkerObject

  File "/home/travis/build/Theano/Theano/theano/gof/op.py", line 803

    """

      ^

TabError: inconsistent use of tabs and spaces in indentation
```

Also, you should rebase this PR to remove the merge commit.

theano/gof/cc.py
```diff
@@ -117,7 +115,12 @@ def failure_code(sub):
 def failure_code_init(sub):
     """
     Code for failure in the struct init.
-
+    Parameters:
+    --------------
```
nouiz: Here too much dash. Can you make a grep in this PR to fix all of them?

theano/gof/cc.py
```diff
@@ -1183,8 +1169,8 @@ def make_thunk(self, input_storage=None, output_storage=None,
             be allocated.
         storage_map: dict that map variables to storages.
             This is used when you need to customize the storage of
-            this thunk.
-
+            this thunk.keep_lock
+        WRITEME
```
nouiz: Bad indentation. Here is what you should write while at it:

keep_lock
    If True, we won't release the lock on the compiledir at the end of this function call.

theano/gof/cc.py
```diff
@@ -1850,11 +1842,12 @@ def make_all(self, profiler=None, input_storage=None, output_storage=None,
...
     """
-    WRITEME
-
     Default checker for DualLinker. This checks that the
     variables contain the same data using ==.
-
+    Parameters
+    -------------
+    x,y
+    the variables to compare data
```
nouiz: Add 4 spaces at the start of this line so that the code is aligned correctly.

theano/gof/cmodule.py
```diff
@@ -1471,10 +1474,24 @@ def _on_atexit(self):
...
+    On NFS filesystems, it is impossible to delete a directory with open
+    files in it.
+
+    So instead, some commands in this file will respond to a
+    failed rmtree() by touching a 'delete.me' file.  This file is a message
+    for a future process to try deleting the directory.
+    Paramters
+    ----------
+    parent
+    Root node to start deleting from
```
nouiz: Here too, the syntax is
```
    Parameters
    ----------
    type : Theano type
        It must be the Theano class itself and not an instance of the class.
```

mgermain:
Things were fixed in #4629

PR #4438, opened 2016-04-27
fixes the 'float64' in RNNs

deepworx:
wastes memory and breaks when warn_float64=raise

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5227,7 +5227,7 @@ def local_opt_alloc(node):
                                if i in node.op.axis]
                     if to_prod:
                         if isinstance(node.op, T.Sum):
-                            val *= T.mul(*to_prod)
+                            val *= T.mul(*to_prod).astype(str(val.dtype))
                         else:
                             val = val ** T.mul(*to_prod)
                     return [T.alloc(T.cast(val, dtype=node.outputs[0].dtype),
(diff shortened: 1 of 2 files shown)


nouiz:
Here is some test code for that:

```
import theano, numpy
s=theano.tensor.lscalar()
a=theano.tensor.alloc(numpy.asarray(5, dtype='float32'), s, s)
orig = theano.config.warn_float64
theano.config.warn_float64 = "raise"
try:
    f = theano.function([s], a.sum())
    theano.printing.debugprint(f)
    f(5)
finally:
    theano.config.warn_float64 = orig
```

I think this is good, but I would like @lamblin or someone else to comment on this. We could also remove the cast on line 5233 as it is not useful anymore.

This change the dtype in which the multiplication is computed compared to the current code. But I think it don't make us loose precision compared to the original graph.

Can someone else verify that they think it is fine?

lamblin:
It looks OK to me. It is indeed better to cast the output of `T.mul()` directly, rather than `val` at the end.
However, we should also do that at line 5232.
A test case would be similar to this one, except using `a.prod()` rather than `a.sum()`.

nouiz:
@deepworkx do you want and have time to finish this?

deepworx:
I`ll try to complete it as requested ;)

nouiz:
Thanks. I think it was quicker to do the small change then to describe them. So I did them here:

#4450

Issue #4425, opened 2016-04-23
`import theano` error workaround for "memcpy was not declared in this scope"


rh314:
Doing an `import theano` for the first time with the `THEANO_FLAGS` environment variable containing `device=gpu`, theano complains that CUDA is not available.

Inspecting the nvcc stderr output revealed that it's related to the following error:
`/usr/include/string.h: In function ‘void* __mempcpy_inline(void*, const void*, size_t)’:
/usr/include/string.h:652:42: error: ‘memcpy’ was not declared in this scope
   return (char *) memcpy (__dest, __src, __n) + __n;`

I'm not sure which systems it affects, but I'm using:
OS: `Ubuntu 16.04 LTS`.
GCC: `version 5.3.1 20160413 (Ubuntu 5.3.1-14ubuntu2)`
NVCC: `Cuda compilation tools, release 7.5, V7.5.17`

As a temporary workaround, I use the following hack:
- Add `cmd.append('-D_FORCE_INLINES')` just before `p = subprocess.Popen(` in the file `nvcc_compiler.py`

Ideas are welcome... perhaps we can close this issue and open a "less-hackish-workaround" issue that addresses the actual problem (which I couldn't figure out just yet).

chaosmail:
this fix worked for me, i am using a similar setup with `gcc 4.9.3`

Gnork:
I'm on Ubuntu 16.04 with nvidia-cuda-toolkit from official repos. I can confirm the bug and the fix is working for me with Theano 0.8.1

fsjunior:
Worked for me too. My config: Ubuntu 16.04 LTS, nvidia-cuda-toolkit 7.5.18-0ubuntu1 (from repository), nvidia-361 (from repository) and gcc 5.3.1 20160413.

phiber1:
I can confirm that since upgrading to Ubuntu 16.04 LTS, I've encountered this exact bug in torch, tensorflow, and now theano. It would appear the underlying cause to be in CUDA 7.5's interaction with the newer gcc updates, and insufficient compatibility testing on their part. I was previously using 16.04 beta (with slightly older version of gcc) with CUDA 7.5, and had no such problems.

ZiweiXU:
This workaround worked for me on gcc 4.8.5 with Ubuntu 16.04 LTS. I don't think this is a Theano issue but it's worthwhile being here.

nouiz:
[...]

PR #4439, opened 2016-04-28
Compatibility between both GPU back-end and doc.

nouiz:
fix gh-4425


abergeron:
The Ubuntu 16.04 changes are also in https://github.com/Theano/Theano/pull/4464.  Ok for the rest.

nouiz:
rebased. It removed the duplicate commit.

Issue #4353, opened 2016-04-11
relu optimization causing NaN in certain cases


dementrock:
```python
import theano.tensor as TT
import theano
import numpy as np


def test1():
    input = TT.vector('input')
    W1 = theano.shared(np.zeros((1, 1)), 'W1')

    h1 = W1.dot(input)
    a1 = TT.nnet.relu(h1)
    W2 = theano.shared(np.zeros((1, 1)), 'W2')

    output = W2.dot(a1)

    W1x = TT.matrix("W1x")
    grad = theano.grad(TT.sum(TT.square(output)), W1)
    xout = TT.grad(TT.sum(grad * W1x), W1)

    fx = theano.function([input, W1x], xout, allow_input_downcast=True)
    print fx([0.], [[0.]])

def test2():
    input = TT.vector('input')
    W1 = theano.shared(np.zeros((1, 1)), 'W1')

    h1 = W1.dot(input)
    a1 = TT.maximum(h1, 0.)
    W2 = theano.shared(np.zeros((1, 1)), 'W2')

    output = W2.dot(a1)

    W1x = TT.matrix("W1x")
    grad = theano.grad(TT.sum(TT.square(output)), W1)
    xout = TT.grad(TT.sum(grad * W1x), W1)

    fx = theano.function([input, W1x], xout, allow_input_downcast=True)
    print fx([0.], [[0.]])

test1() # => [[nan]]
test2() # => [[0.]]
```

lamblin:
That particular case can be solved by an optimization that would do CGer(0, x) -> 0, or maybe even dot(0, x) -> 0.
But that would not solve the issue of simplifying the gradients through `abs` that show up here.
Maybe the best solution would be to have a ReLU Op that has a better implementation of `grad`, and that then gets optimized away.

dementrock:
@lamblin I vote for a better implementation of grad for the RELU op. It seems like a cleaner approach.

manuels:
I am able to reproduce this bug. This code also returns NaN when using `optimizer=fast_compile` but not when using `optimizer=fast_run`:

    a =theano.tensor.fscalar("a")
    b=theano.tensor.nnet.relu(a)
    c=theano.grad(b,a)
    f=theano.function([a],[c])
    print f(0.0)

(this bug was also reported in [another project](https://github.com/fchollet/keras/issues/828#issuecomment-213889948) that uses theano)

nouiz:
[...]

PR #4443, opened 2016-04-29
Change the grad of abs to prevent relu from generating nan in fast_compile

nouiz:
fix gh-4353

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -2150,7 +2150,7 @@ class Abs(UnaryScalarOp):
             else:
                 return [x.zeros_like()]
 
-        if x.dtype.startswith('float'):
+        if x.type in float_types:
             return gz * sgn(x),
         return gz * x / abs(x),  # formula works for complex and real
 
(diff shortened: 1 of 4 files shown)


theano/scalar/basic.py
```diff
@@ -2150,6 +2150,8 @@ def grad(self, inputs, gout):
             else:
                 return [x.zeros_like()]
 
+        if x.dtype.startswith('float'):
+            return gz * sgn(x),
```
lamblin: That looks like a more stable formula, indeed.
Bud could you check for `x in float_dtypes` instead?

nouiz: done

nouiz:
Thanks. I think it was quicker to do the small change then to describe them. So I did them here:

https://github.com/Theano/Theano/pull/4450

nouiz:
Sorry, comment on the wrong branch.

nouiz:
@lamblin this PR is ready to review.

PR #4450, opened 2016-05-02
Remove upcast to float64 in the graph.

nouiz:
finish gh-4438.


theano/tensor/opt.py
```diff
@@ -5210,11 +5210,12 @@ def local_opt_alloc(node):
                     val = get_scalar_constant_value(input)
                     assert val.size == 1
                     # check which type of op
+                    casted = T.mul(*shapes).astype(str(val.dtype))
```
lamblin: Wait, `val.dtype` does not seem to be the right dtype here, I think we should use `node.outputs[0].dtype`.

theano/tensor/opt.py
```diff
@@ -5226,11 +5227,12 @@ def local_opt_alloc(node):
                     to_prod = [shapes[i] for i in xrange(len(shapes))
                                if i in node.op.axis]
                     if to_prod:
+                        casted = T.mul(*shapes).astype(str(val.dtype))
```
lamblin: Same here.

theano/tensor/opt.py
nouiz: Why? node.outputs[0].dtype should be val.dtype. Alloc output dtype is val dtype. Alloc just broadcast a value, it do not change the type.

lamblin: `get_scalar_constant_value` digs through `cast` operations.
You could use `input.dtype`, though.

nouiz: done

lamblin:

cesarsouza:
May I ask if this issue and its fix were related only to the multiplication operation? I am adding two float32 tensors together (one that comes from a quite complex set of computations) and their result is also being upcast to float64. I am using Theano from the master branch, cloned about two days ago.

lamblin:
It was related to the `sum` and `prod` reduction operations, it should not affect the elemwise addition or multiplication.

PR #4454, opened 2016-05-02
Move new GPU backend out of sandbox

slefrancois:


doc/tutorial/using_gpu.txt
```diff
@@ -470,7 +470,7 @@ the GPU object directly.  The following code is modifed to do just that.
 
   rng = numpy.random.RandomState(22)
   x = shared(numpy.asarray(rng.rand(vlen), config.floatX))
-  f = function([], sandbox.gpuarray.basic_ops.gpu_from_host(tensor.exp(x)))
+  f = function([], sandbox.gpuarray.basic_ops.GpuFromHost(None)(tensor.exp(x)))
```
nouiz: I don't like the idea to force user to always specify the context. I think we should do one of:

1) make gpu_from_host default to None.
2) make gpu_from_host be a special case where it mean any GPU context and then we will keep it on the GPU but it mean the user don't care which context. We should not make it more difficult to use 1 GPU in the new back-end for the purpose that we can support more then 1.

I made an issue about this:

https://github.com/Theano/Theano/issues/4457

slefrancois: For now, how about we add an instance "gpu_from_host = GpuFromHost(None)" to gpuarray.basic_ops?

nouiz: @abergeron was again it from memory. So I would wait for him to be back before doing this.

doc/library/gpuarray/dnn.txt
```diff
@@ -1,7 +1,7 @@
 .. _libdoc_gpuarray_dnn:
 
 ===========================================
-:mod:`theano.sandbox.gpuarray.dnn` -- cuDNN
+:mod:`gpuarray.dnn` -- cuDNN
```
abergeron: This should still be `theano.gpuarray`

doc/library/gpuarray/index.txt
```diff
@@ -2,10 +2,10 @@
 .. _libdoc_gpuarray:
 
 =======================================================
-:mod:`theano.sandbox.gpuarray` -- The (new) GPU backend
+:mod:`gpuarray` -- The (new) GPU backend
```
abergeron: Idem

doc/library/gpuarray/type.txt
```diff
@@ -1,8 +1,8 @@
 .. _libdoc_gpuarray_type:
 
 ===================================================
-:mod:`theano.sandbox.gpuarray.type` -- Type classes
+:mod:`gpuarray.type` -- Type classes
```
abergeron: Idem

theano/gpuarray/comp.py
```diff
@@ -40,7 +40,7 @@ def compile_args():
```
abergeron: This file should probably be deleted.  It isn't used anywhere.

theano/gpuarray/tests/GpuArray.pkl
```diff
@@ -1,4 +1,4 @@
-ctheano.sandbox.gpuarray.type
```
abergeron: Should we add something for backward compat?  Otherwise unfixed pickles will break.

theano/gpuarray/tests/test_pickle.py
```diff
@@ -29,7 +29,7 @@
         cuda_ndarray.use('gpu', default_to_move_computation_to_gpu=False,
                          move_shared_float32_to_gpu=False,
                          enable_cuda=False)
-        theano.sandbox.gpuarray.init_dev('cuda')
```
abergeron: This code will have to be reworked to not try to invoke demons.

theano/tests/test_flake8.py
```diff
@@ -123,7 +123,7 @@
     "sandbox/cuda/tests/test_extra_ops.py",
     "sandbox/cuda/tests/test_gemmcorr3d.py",
     "sandbox/cuda/tests/test_viewop.py",
-    "sandbox/gpuarray/tests/__init__.py",
+    "gpuarray/tests/__init__.py",
```
abergeron: You should probably move the file so that the names stay sorted.

theano/tests/test_flake8.py
```diff
@@ -134,7 +134,7 @@
     "sandbox/linalg/__init__.py",
     "sandbox/linalg/tests/__init__.py",
     "sandbox/linalg/tests/test_linalg.py",
-    "sandbox/gpuarray/__init__.py",
+    "gpuarray/__init__.py",
```
abergeron: Same here

theano/gpuarray/tests/test_pickle.py
nouiz: what do you mean? I don't see any demons here. What I'm missing?

abergeron: I mean that this should use either the normal code to get the test context or just initiailize straight away and not go through cuda_ndarray (which doesn't work anymore).

I remember that this was done this way because if you import config it raises SkipTest when pygpu is not available.  I think that catching SkipTest and noting that is a better way than what is here.

nouiz: This is seem a good code cleanup for another PR.

abergeron: I'm ok with that.

Though I suspect that this code is why we saw the testsuite use two GPUs for no reason.

nouiz: Can you make a PR to Simon PR and fix this?

doc/library/gpuarray/dnn.txt
slefrancois: All other modules omit the theano prefix in the doc, I wrote it this way to stay consistent.

theano/gpuarray/tests/GpuArray.pkl
slefrancois: I added a placeholder init in theano/sandbox/gpuarray.

doc/library/gpuarray/dnn.txt
abergeron: I know they do, but it makes the internal links not work sometimes because sphinx thinks the module is `gpuarray` not `theano.gpuarray`.

PR #4461, opened 2016-05-03
GpuSolve based on CUSOLVER instead of CULA

ddtm:
This PR is intended for replacing the old CULA backend of GpuSolve with a newer CUSOLVER-based implementation.

### DIFF
--- a/theano/sandbox/cuda/cusolver.py
+++ b/theano/sandbox/cuda/cusolver.py
@@ -1,0 +1,164 @@
+from __future__ import absolute_import, print_function, division
+import pkg_resources
+
+import theano
+from theano.sandbox.cuda.type import CudaNdarrayType
+from theano.sandbox.cuda import CudaNdarray, GpuOp
+from theano.sandbox.cuda.basic_ops import as_cuda_ndarray_variable
+
+try:
+    from theano.sandbox.cuda import cuda_ndarray
+    dimshuffle = cuda_ndarray.cuda_ndarray.dimshuffle
+except ImportError:
+    pass
+
+cusolver_available = False
+
+try:
+    from scikits.cuda import cusolver
+    cusolver_available = True
+except (ImportError, OSError, RuntimeError, pkg_resources.DistributionNotFound):
+    pass
+
+cusolver_handle = [None]
+
+class GpuSolve(GpuOp):
+    """
+    CUSOLVER GPU solver OP.
+
+    Parameters
+    ----------
+    trans
+        Whether to take the transpose of the input matrix or not.
+
+    """
+
+    __props__ = ('trans',)
+
+    def __init__(self, trans='N'):
+        self.trans = trans
+        super(GpuSolve, self).__init__()
+
+    def output_type(self, inp):
+        return CudaNdarrayType(broadcastable=[False] * inp.type.ndim)
+
+    def make_node(self, inp1, inp2):
+        inp1 = as_cuda_ndarray_variable(inp1)
+        inp2 = as_cuda_ndarray_variable(inp2)
+
+        assert inp1.ndim == 2
+        assert inp2.ndim == 2
+        return theano.Apply(self, [inp1, inp2], [self.output_type(inp1)()])
+
+    def make_thunk(self,
+                   node,
+                   storage_map, _,
+                   no_recycling=[]):
+        if not cusolver_available:
+            raise RuntimeError('CUSOLVER is not available and '
+                               'GpuSolve Op can not be constructed.')
+
+        inputs = [storage_map[v] for v in node.inputs]
+        outputs = [storage_map[v] for v in node.outputs]
+
+        def thunk():
+            # size of the matrices to invert
+            z = outputs[0]
+
+            # Matrix
+            A = inputs[0][0]
+
+            # Solution vectors
+            b = inputs[1][0]
+
+            # A is not explicitly converted between C and F order, instead we
+            # switch the "transpose" flag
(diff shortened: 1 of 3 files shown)


theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,164 @@
...
+
+cusolver_available = False
+
+try:
+    from scikits.cuda import cusolver
+    cusolver_available = True
+except (ImportError, OSError, RuntimeError, pkg_resources.DistributionNotFound):
+    pass
+
+cusolver_handle = [None]
```
nouiz: You don't need to use a list. Just use it directly.

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,164 @@
...
+
+class GpuSolve(GpuOp):
```
nouiz: As we see there could be many implementation of that, I would prefer a more precise name. GpucuSolverXXX.

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+
+    __props__ = ('trans',)
+
+    def __init__(self, trans='N'):
+        self.trans = trans
+        super(GpuSolve, self).__init__()
+
+    def output_type(self, inp):
+        return CudaNdarrayType(broadcastable=[False] * inp.type.ndim)
```
nouiz: Making a method for that isn't good. I know it was like that in the old code, but can you just inline this?

Also, what is the relation from the input shapes to the output shapes? I think they are the same, so the broadcastable can probably be better set.

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,164 @@
...
+
+            # Convert b to F-order from c-order.
+            b_cpy = dimshuffle(b, (1, 0)).reshape((b.shape[0], b.shape[1]))
+
+            # This copy forces allocation of a new C-contiguous buffer
+            # and returns it.
+            A_cpy = A.copy()
+            b_cpy = b_cpy.copy()
+
+            def cusolver_gpu_solve(A_, b_, trans='T'):
```
nouiz: I know it come from the old code, but personally, I didn't liked that it define this method and use it only once. I also think it slow down computation. So just inlining it would be great.

nouiz:
Just a few more comments,
- it is missing tests
- travis have a few flakes8 error.


Otherwise, it look good. I don't know too much how the precisson of the computation. If you know it, it would be good to add this in the docstring (as to write how you do it, some user will wan to know this)

theano/sandbox/cuda/cusolver.py
```diff
@@ -37,116 +39,109 @@ class GpuSolve(GpuOp):
...
+
+            lda = max(1, n)
+            ldb = max(1, n, l)
+
+            A_ptr = A_cpy.gpudata
+            b_ptr = b_cpy.gpudata
 
-                if thunk.dev_info is None:
-                    thunk.dev_info = CudaNdarray.zeros((1,))
+            if cusolver_handle is None:
```
nouiz: I would move the the init of the handle inside the make_thunk method, but outside the thunk() function. This will make profiling give more useful information by having it in the linker phase and not in the execute phase.

theano/sandbox/cuda/tests/test_cusolver.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Skip tests if cuda_ndarray is not available.
+from nose.plugins.skip import SkipTest
+import theano.sandbox.cuda as cuda_ndarray
+from theano.misc.pycuda_init import pycuda_available
+from theano.sandbox.cuda.cusolver import cusolver_available
+
+from theano.sandbox.cuda import cusolver
+
+if not cuda_ndarray.cuda_available:
```
nouiz: travis is failing as the tests aren't skipped. I don't understand why. I'll close/reopen travis again to make travis run with the current master of Theano.

nouiz:
close/reopen to trigger travis with the current master.

theano/sandbox/cuda/tests/test_cusolver.py
nouiz: The problem is still there, but now I understand why!

```
  File "/home/travis/build/Theano/Theano/theano/sandbox/cuda/tests/test_cusolver.py", line 13, in <module>
    from theano.sandbox.cuda.cusolver import cusolver_available
  File "/home/travis/build/Theano/Theano/theano/sandbox/cuda/cusolver.py", line 6, in <module>
    from theano.sandbox.cuda import CudaNdarray, GpuOp
ImportError: cannot import name CudaNdarray

CudaNdarray isn't availble when there is no nvcc and GPU. I'll comment at the right place for the import change needed.
```

This mostly mean we can't import the file cusolver for now.

theano/sandbox/cuda/cusolver.py
```diff
@@ -0,0 +1,159 @@
+from __future__ import absolute_import, division, print_function
+
+import pkg_resources
+import theano
+
+from theano.sandbox.cuda import CudaNdarray, GpuOp
```
nouiz: replace that by:

import theano.sandbox.cuda as cuda
from theano.sandbox.cuda import GpuOp

and use cuda.CudaNdarray elsewhere.

This will allow to import that file when cuda isn't available.

nouiz:
continued in https://github.com/Theano/Theano/pull/4917

PR #4356, opened 2016-04-11
Replace some subtensors with DimShuffle first commit

SinaHonari:
Replace some subtensors with DimShuffle, issue #4162

PR #4463, opened 2016-05-03
Scan reintroduced benchmark

nouiz:
- Add debug flags
- otherwise add an optimization that remove useless elemwise and have as a side effect to remove the extra scan introduced. So this is a work around some extra scan introduced by https://github.com/Theano/Theano/pull/4356/files#r61901454

### DIFF
--- a/theano/sandbox/cuda/dnn_fwd.c
+++ b/theano/sandbox/cuda/dnn_fwd.c
@@ -159,6 +159,36 @@ APPLY_SPECIFIC(conv_fwd)(CudaNdarray *input, CudaNdarray *kerns,
       chosen_algo = CONV_ALGO;
     }
 
+    if (0){
+      char * a;
+      switch(chosen_algo){
+      case CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM:
+	a = "implicit gemm (0)";
+	break;
+      case CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM:
+	a = "precomp gemm (1)";
+	break;
+      case CUDNN_CONVOLUTION_FWD_ALGO_GEMM:
+	a = "gemm (2)";
+	break;
+      case CUDNN_CONVOLUTION_FWD_ALGO_DIRECT:
+	a = "direct (3)";
+	break;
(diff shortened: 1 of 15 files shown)


theano/sandbox/cuda/dnn_fwd.c
```diff
@@ -159,6 +159,36 @@ APPLY_SPECIFIC(conv_fwd)(CudaNdarray *input, CudaNdarray *kerns,
...
+	break;
+#if CUDNN_VERSION > 5000
+      case CUDNN_CONVOLUTION_FWD_ALGO_WINOGRAD:
+	a = "winograd (6)";
+	break;
+#endif
+      }
+      printf("GpuDNNConv: algo %s\n", a);
+    }
+
```
abergeron: I'm not sure I like debugging stuff like this being left in.

theano/tensor/opt.py
```diff
@@ -4760,6 +4760,10 @@ def local_useless_elemwise_comparison(node):
     Elemwise[LT](add([anything that is shapes]), 0) -> Elemwise[zeros](X)
     Elemwise[GE](add([anything that is shapes]), 0) -> Elemwise[ones](X)
 
+    # Shapes are never negativ
```
abergeron: negativ -> negative

abergeron: I'm also wondering if we should make shapes into unsigned ints so that all the special casing done here would be standard.

theano/sandbox/cuda/dnn_fwd.c
nouiz: This is not useful just for debugging. We should also make this available to users. I'm not sure how to do this, I let this for later.

nouiz:
I fixed the typo and rebased.

nouiz:
I fixed the typo and rebased.

abergeron:

lamblin:
Well, Travis failed.

nouiz:
I updated this PR to fix travis by adding support for the optimization to one more cases caused by different order of optimization.

abergeron:
Travis still fails

nouiz:
I updated the PR.

abergeron:
Ok when travis passes.

abergeron:
Still ok (even if travis hasn't passed yet).

PR #4464, opened 2016-05-04
Doc Ubuntu 16.04 and clean up flag in_c_key

nouiz:
Should help lower the number of compiled files when we change Theano flags.

### DIFF
--- a/doc/install_ubuntu.txt
+++ b/doc/install_ubuntu.txt
@@ -12,6 +12,31 @@ For NVIDIA Jetson TX1 embedded platform:
     sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libblas-dev git
     pip install --upgrade --no-deps git+git://github.com/Theano/Theano.git --user  # Need Theano 0.8(not yet released) or more recent
 
+For Ubuntu 16.04 with cuda 7.5
+
+.. code-block:: bash
+
+    sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libopenblas-dev git
+    sudo pip install Theano
+
+    # cuda 7.5 don't support the default g++ version. Install an supported version and make it the default.
+    sudo apt-get install g++-4.9
+
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.9 20
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-5 10
+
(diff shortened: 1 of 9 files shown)


theano/configdefaults.py
```diff
@@ -42,6 +42,7 @@ def floatX_convert(s):
              "Note: float16 support is experimental, use at your own risk.",
              EnumStr('float64', 'float32', 'float16',
                      convert=floatX_convert,),
+             in_c_key=False
              )
```
lamblin: That flag was the whole reason for introducing `in_c_key` in the first place.
The issue was that some nodes, maybe Elemwise or CAReduce, would have an output dtype that depends not only on the input's dtypes, but also on the _current_ value of `floatX` (not the one when the node was built). It means that if C code is compiled or reloaded, and _then_ the value of `floatX` changes, then the value computed would have a different type from the output variable.

nouiz: Now CAReduce don't do that and CAReduceDtype have the output and acc dtype as parameters. I don't see any use of floatX in the file tensor/elemwise.py that would cause this.

If the output dtype of a node depend of floatX or the casting policy that we didn't finish, then I see it can cause problem, as we don't add in the key the output variable.

I find this bad that we need floatX in the in_c_key. This cause too much key when we change floatX for no good reason. I prefer that we add the node outputs type in the key. I made an issue about this: 

https://github.com/Theano/Theano/issues/4466

I just removed that change from this PR.

doc/install_ubuntu.txt
```diff
@@ -12,6 +12,31 @@ For NVIDIA Jetson TX1 embedded platform:
     sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libblas-dev git
     pip install --upgrade --no-deps git+git://github.com/Theano/Theano.git --user  # Need Theano 0.8(not yet released) or more recent
 
+For Ubuntu 16.04 with cuda 7.5
+
+.. code-block:: bash
+
+    sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libopenblas-dev git
+    sudo pip install Theano
```
kmike: For what it worths installing Theano in virtualenv is almost as easy: there is no need for python-scipy and python-numpy apt packages because scipy and numpy are installed from manylinux wheels by default if a recent pip version is used; I think they link against openblas. 16.04 has python-pip recent enough.

nouiz: If you want to make a PR to add in the doc how to use virtualenv, we can add that. But don't forget that many people use anaconda that is free and link with mkl. That is even faster.

PR #4468, opened 2016-05-04
Scan reintroduced

nouiz:
This depend on PR: https://github.com/Theano/Theano/pull/4463

### DIFF
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -330,7 +330,7 @@ class FunctionGraph(utils.object2):
             if not used_or_output:
                 if not hasattr(apply_node.tag, 'removed_by'):
                     apply_node.tag.removed_by = []
-                apply_node.tag.removed_by.append(reason)
+                apply_node.tag.removed_by.append(str(reason))
                 self.apply_nodes.remove(apply_node)
                 self.variables.difference_update(apply_node.outputs)
                 self.execute_callbacks('on_prune', apply_node, reason)
@@ -421,7 +421,7 @@ class FunctionGraph(utils.object2):
             self.apply_nodes.add(node)
             if not hasattr(node.tag, 'imported_by'):
                 node.tag.imported_by = []
-            node.tag.imported_by.append(reason)
+            node.tag.imported_by.append(str(reason))
             for output in node.outputs:
                 self.__setup_r__(output)
                 self.variables.add(output)
--- a/theano/gof/fg.py
+++ b/theano/gof/fg.py
@@ -328,6 +328,9 @@ class FunctionGraph(utils.object2):
                               if output.clients or output in self.outputs]
             # If the apply node is not used and is not an output
             if not used_or_output:
+                if not hasattr(apply_node.tag, 'removed_by'):
+                    apply_node.tag.removed_by = []
+                apply_node.tag.removed_by.append(reason)
                 self.apply_nodes.remove(apply_node)
                 self.variables.difference_update(apply_node.outputs)
                 self.execute_callbacks('on_prune', apply_node, reason)
@@ -416,6 +419,9 @@ class FunctionGraph(utils.object2):
             assert node not in self.apply_nodes
             self.__setup_node__(node)
             self.apply_nodes.add(node)
+            if not hasattr(node.tag, 'imported_by'):
+                node.tag.imported_by = []
+            node.tag.imported_by.append(reason)
             for output in node.outputs:
                 self.__setup_r__(output)
                 self.variables.add(output)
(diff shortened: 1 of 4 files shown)


rizar:
So far I tried it on CPU with fast_run. Judging by the names of the scan ops in the computation graph, it helped. It also became slightly faster, not much, probably because in this particular model it was not the bottleneck.

I will also try later on GPU.

nouiz:
@rizar, I have added a pickle related fix.

rizar:
I tried on GPU. It does fix the duplicates scan nodes, but somehow without a speedup. I can also profile it.

rizar:
Before: https://gist.github.com/rizar/0d213d05a9d1f197219c1eb674947496
After: https://gist.github.com/rizar/753bdaeebd6f16692c2218f537be831c

rizar:
Looking at the profile it seems like it really helps, but not in the bottleneck. Still, it is very nice to have it fixed.

nouiz:
What is this op:

lvsr.ops.EditDistanceOp? It take 25% globally. Where can I find this code?

abergeron:
This also looks ok, but I should look at it again after #4463 is merged.

nouiz:
rebased

nouiz:
rebased

theano/scan_module/scan_opt.py
```diff
@@ -2070,6 +2072,7 @@ def map_out(outer_i, inner_o, outer_o, seen):
             new_outer_out_mit_mot.append(outer_omm)
     na.outer_out_mit_mot = new_outer_out_mit_mot
 
+    return dict([("remove", remove)] + zip(node.outputs, na.outer_outputs))
```
nouiz: I think we need to test this a little bit more before merging. It could trigger false user error, as if they get merged, then it would be probably fine.

abergeron:
I'm ok with https://github.com/Theano/Theano/pull/4468/commits/e4dd1cf2f3aa3a4be0d59d4e69e4d6ae8f536ef2 and https://github.com/Theano/Theano/pull/4468/commits/fcbb56ed1009676f9696c7deac986bfc8c6be0f4  but I would have to spend more time to look at the remove tagging stuff.

nouiz:
Can you finish today the review?

theano/gof/fg.py
```diff
@@ -416,6 +419,9 @@ def __import__(self, apply_node, check=True, reason=None):
             assert node not in self.apply_nodes
             self.__setup_node__(node)
             self.apply_nodes.add(node)
+            if not hasattr(node.tag, 'imported_by'):
+                node.tag.imported_by = []
+            node.tag.imported_by.append(str(reason))
```
abergeron: This doesn't seem like it's used anywhere

theano/gof/fg.py
```diff
@@ -328,6 +328,9 @@ def __remove_clients__(self, r, clients_to_remove,
                               if output.clients or output in self.outputs]
             # If the apply node is not used and is not an output
             if not used_or_output:
+                if not hasattr(apply_node.tag, 'removed_by'):
+                    apply_node.tag.removed_by = []
+                apply_node.tag.removed_by.append(str(reason))
```
abergeron: And this also looks unused.

theano/scan_module/scan_opt.py
```diff
@@ -2070,6 +2072,7 @@ def map_out(outer_i, inner_o, outer_o, seen):
             new_outer_out_mit_mot.append(outer_omm)
     na.outer_out_mit_mot = new_outer_out_mit_mot
 
+    return dict([("remove", remove)] + list(zip(node.outputs, na.outer_outputs)))
```
abergeron: You should probably remove the extra return just below.

theano/scan_module/scan_opt.py
```diff
@@ -1964,8 +1964,10 @@ def scan_merge_inouts(node):
             outputs = [outputs]
 
         na = scan_args(outer_inputs, outputs, op.inputs, op.outputs, op.info)
+        remove = [node]
     else:
         na = a
+        remove = []
```
abergeron: If this is not going to remove or replace anything, it should just return None.  Otherwise we are working for no reason.

abergeron:
Ok modulo my comments.

theano/gof/fg.py
nouiz: I know, this is to help debugging.

theano/scan_module/scan_opt.py
nouiz: remove can't be none when calling replace_all_validate_remove(). It must be an iterable. The extra work is so low that I don't think it is useful to make 2 code path to not pass the remove parameter.

theano/scan_module/scan_opt.py
nouiz: done

theano/scan_module/scan_opt.py
abergeron: I meant that in this case it could just not call replace_validate_remove because it's not replacing anything.

nouiz: done

Issue #4483, opened 2016-05-09
Leky relu upcast to float64


hmendozap:
Hi,

I noticed that the leaky rectifier activation upcasts the gradient of the weights in the layer. (I mentioned on Lasagne repo)
```
theano.tensor.nnet.relu(X, np.float32(0.123))
/home/mendozah/workspace/envHDMp/bin/ipython:1: UserWarning: You are creating a TensorVariable with float64 dtype. You requested an action via the Theano flag warn_float64={ignore,warn,raise,pdb}
```
What I find weird is that if alpha is python float, there is no upcast, but if it is np.float32 there is.
So I just added this to the call:
```python
alpha =  theano.tensor.as_tensor_variable(alpha))
```

Should that be enough?

Thanks,
HM

nouiz:
What is X.type?

hmendozap:
float32

f0k:
This was reported at Lasagne/Lasagne/issues/677, and `theano.tensor.as_tensor_variable(alpha)` is a good solution (we had similar issues in other places in Lasagne). We want to allow `alpha` to be either a Python float, Theano float or numpy float without any upcasting to float64 occurring at `alpha + 1` and `alpha - 1` here: https://github.com/Theano/Theano/blob/8c7511b/theano/tensor/nnet/nnet.py#L2223-2224
This can be achieved either by wrapping `alpha` in `theano.tensor.as_tensor_variable`, or by wrapping `1` in `theano.tensor.constant`.

nouiz:
thanks:

https://github.com/Theano/Theano/pull/4484

PR #4484, opened 2016-05-09
Relu

nouiz:
fix gh-4483

### DIFF
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -676,6 +676,20 @@ Creating Tensor
 
     Returns a tensor filled with 1s that has same shape as `x`.
 
+.. function:: zeros(shape, dtype=None)
+
+    :param shape: a tuple/list of scalar with the shape information.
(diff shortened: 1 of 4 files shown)


doc/library/tensor/basic.txt
```diff
@@ -676,6 +676,20 @@ Creating Tensor
...
+
+    Returns a tensor filled with 0s of the provided shape.
+
+.. function:: ones(shape, dtype=None)
+
+    :param shape: a tuple/list of scalar with the shape information.
+    :param dtype: the dtype of the new tensor. If None, will use floatX.
+
+    Returns a tensor filled with 1s of the provided shape.
+
```
f0k: Was that meant for a different PR?

f0k:
Code and test look good to me.

doc/library/tensor/basic.txt
nouiz: I have a tendency to add small stuff in PR. Probably not the best, but as travis is slow for our big test suite, it help us.

abergeron:
Yep all good.

Issue #4485, opened 2016-05-09
crash during test_nvidia_driver---"argument  not a CudaNdarray"


braingineer:
Hi all,

I updated to cuDNN 5 and theano 0.9 and then things started to crash.  

I've pasted the crash here:

https://gist.github.com/braingineer/cabb0706dcdf334fac7bc9d59ef96a31

Thanks

braingineer:
re-installed 0.8.2 and it seems to be working.

nouiz:
Fixed in https://github.com/Theano/Theano/pull/4489

thanks

braingineer:
awesome, thanks!

PR #4489, opened 2016-05-09
Fix regression added in 1a3948cc53d2 that cause crashes.

nouiz:
Reported at: https://groups.google.com/d/topic/theano-users/DqlvVjNUK-U/discussion

fix gh-4485


nouiz:
Merging as this is an important regression and the fix is easy.

lamblin:
I checked it and it's fine for me.

PR #4491, opened 2016-05-10
Add reference to new article

lamblin:
Also update the news, and update author's aliases.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -53,6 +53,8 @@ Hani Almousli <hani.mousli@gmail.com> Hani <hani.mousli@gmail.com>
 Hani Almousli <hani.mousli@gmail.com> HaniAlmousli <hani.mousli@gmail.com>
 Ian Goodfellow <goodfellow.ian@gmail.com> Ian Goodfellow <devnull@localhost>
 Ian Goodfellow <goodfellow.ian@gmail.com> goodfeli <goodfellow.ian@gmail.com>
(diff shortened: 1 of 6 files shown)


.mailmap
```diff
@@ -159,4 +164,19 @@ Xavier Bouthillier <xavier.bouthillier@gmail.com> Xavier Bouthillier <xavier.bou
 Xavier Bouthillier <xavier.bouthillier@gmail.com> Xavier Bouthillier/ <xavier.bouthillier@gmail.com>
 Orhan Firat <orhan.firat@ceng.metu.edu.tr> orhanf <orhan.firat@ceng.metu.edu.tr>
 Søren Kaae Sønderby <skaaesonderby@gmail.com> skaae <skaaesonderby@gmail.com>
-Luke Metz <luke.metz@students.olin.edu> = <luke.metz@students.olin.edu>
\ No newline at end of file
+Luke Metz <luke.metz@students.olin.edu> = <luke.metz@students.olin.edu>
+Guillaume Alain <guillaume.alain.umontreal@gmail.com> Guillaume Alain <gyomalin@gmail.com>
```
nouiz: All that is good. But I think next time we should sort it alphabetically. Otherwise, it is starting to be hard to know if someone is in this list or not.

lamblin: Right. I thought I did it, but I apparently misplaced that change.

lamblin: Should be done in #4493

PR #4492, opened 2016-05-10
[CRASH,REGRESSION] Fix crash gpufft

nouiz:
Small doc and example change at the same time.


doc/library/compile/nanguardmode.txt
```diff
@@ -63,6 +63,12 @@ kind of abnormal values to monitor. ``nan_is_error`` and ``inf_is_error`` has
 no default values, so they need to be set explicitly, but ``big_is_error`` is
 set to be ``True`` by default.
 
+.. note::
```
mgermain: How about something a little bit less verbose?

NanGuardMode significantly slows down computations; only enable as needed.

nouiz: I like more that. I changed it.

nouiz:
@mgermain told me we can merge when travis pass. The part needed passed, so merging.

PR #4496, opened 2016-05-11
Gpu opt

nouiz:
I still need to run GPU tests.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -256,6 +256,11 @@ def inplace_elemwise_optimizer_op(OP):
                 continue
 
             if op.inplace_pattern:
+                # Maybe this isn't needed anymore, but I don't want to
+                # rish regression now. This case only happen if the
+                # original node add already some inplace patter and we
+                # still try to add more pattern.
+
                 baseline = op.inplace_pattern
                 candidate_outputs = [i for i in xrange(len(node.outputs))
                                      if i not in baseline]
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -237,6 +237,11 @@ def inplace_elemwise_optimizer_op(OP):
         else:
             update_outs = []
 
+        protected_inputs = [
+            f.protected for f in fgraph._features if
+            isinstance(f, theano.compile.function_module.Supervisor)]
+        protected_inputs = sum(protected_inputs, [])  # flatten the list
+        protected_inputs.extend(fgraph.outputs)
         for node in list(graph.io_toposort(fgraph.inputs, fgraph.outputs)):
             op = node.op
             # gpuarray GpuElemwise inherit from Elemwise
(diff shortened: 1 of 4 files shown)


abergeron:
Tentative ok, but I'm not sure about the changes in tensor/opt.py since I'm not what it is supposed to do.

theano/tensor/opt.py
```diff
@@ -155,13 +155,16 @@ def broadcast_like(value, template, fgraph, dtype=None):
     if template not in fgraph.variables:
         raise NotImplementedError('broadcast_like currently requires the '
                                   'template Variable to be in the fgraph already')
+    if dtype is None:
+        dtype = template.dtype
+    value = T.cast(value, dtype)
+    if value.type == template.type:
```
nouiz: this do like on line 152. It would prevent the creation of the alloc when it isn't needed.

theano/tensor/opt.py
```diff
@@ -234,6 +237,11 @@ def inplace_elemwise_optimizer(fgraph):
         else:
             update_outs = []
 
+        protected_inputs = [
+            f.protected for f in fgraph._features if
+            isinstance(f, theano.compile.function_module.Supervisor)]
+        protected_inputs = sum(protected_inputs, [])  # flatten the list
+        protected_inputs.extend(fgraph.outputs)
```
nouiz: This is just code moved outside the loop as it is constant during all the loop iteration.

theano/tensor/opt.py
```diff
@@ -242,27 +250,36 @@ def inplace_elemwise_optimizer(fgraph):
             # If big graph and the outputs are scalar, do not make it
             # inplace.
             if (check_each_change != 1 and
-                all([getattr(o.type, 'ndim', -1) == 0
-                     for o in node.outputs])):
+                # If multiple outputs, they must all have the same size,
+                # so only check the first.
+                    getattr(node.outputs[0].type, 'ndim', -1) == 0):
```
nouiz: This remove the list comprehension and the all. This is a speed up.

theano/tensor/opt.py
```diff
@@ -242,27 +250,36 @@ def inplace_elemwise_optimizer(fgraph):
...
-            # node inputs that are Constant, already destroyed,
-            # fgraph protected inputs and fgraph outputs can't be used as inplace
-            # target.
-            # Remove here as faster.
-            candidate_inputs = [i for i in xrange(len(node.inputs))
-                                if i not in baseline.values() and
-                                not isinstance(node.inputs[i], Constant) and
-                                not fgraph.destroyers(node.inputs[i]) and
-                                node.inputs[i] not in protected_inputs]
+            if op.inplace_pattern:
```
nouiz: This if is a speed up. Normally, baseline is empty. In the if, we have the old code. In the else, we have simplified/faster code that suppose op.inplace_pattern is empty.

theano/tensor/opt.py
```diff
@@ -2706,6 +2723,7 @@ def merge_two_slices(slice1, len1, slice2, len2):
             val = T.switch(T.lt(sl2, 0), - len1 - 1, val)
             if sl1.step:
                 val = T.switch(T.eq(sl1.step, 0), len1 + 1, val)
+            val = pre_greedy_local_optimizer(list_opt, val)
```
nouiz: This and the next one is important. It reapply what was done on line 2802 here. This make this optimization return a cleaned graph instead of making it generate ugly stuff that is cleaned up later. This is faster.

theano/tensor/opt.py
abergeron: That part is ok.

theano/tensor/opt.py
abergeron: ok

theano/tensor/opt.py
```diff
@@ -242,27 +250,36 @@ def inplace_elemwise_optimizer(fgraph):
...
+                # inplace target.
+                # Remove here as faster.
+                candidate_inputs = [i for i in xrange(len(node.inputs))
+                                    if i not in baseline.values() and
+                                    not isinstance(node.inputs[i], Constant) and
+                                    # Is next line costly?
+                                    not fgraph.destroyers(node.inputs[i]) and
+                                    node.inputs[i] not in protected_inputs]
+            else:
+                baseline = []
```
abergeron: This doesn't look like it's used anywhere.

abergeron:
I'm fine now apart from the empty baseline that looks unused.

theano/tensor/opt.py
nouiz: it is used here:

https://github.com/nouiz/Theano/blob/1328ebb3a208c400eb5ab2424cfdd8caa59d8805/theano/tensor/opt.py#L347

nouiz:
I updated this PR with @lamblin comments.

PR #4500, opened 2016-05-11
Update doc with instructions for using new gpu backend

slefrancois:
Submitting now, but it can be saved for later when the new backend is fully stable.

### DIFF
--- a/doc/install.txt
+++ b/doc/install.txt
@@ -524,8 +524,8 @@ You can also set these options in the .theanorc file's ``[global]`` section:
 Note that:
 
     * If your computer has multiple GPUs and you use 'device=cuda', the driver
-      selects the one to use (usually gpu0).
-    * You can use the program nvida-smi to change this policy.
+      selects the one to use (usually cuda0).
+    * You can use the program ``nvidia-smi`` to change this policy.
     * You can choose one specific GPU by specifying 'device=cudaX', with X the
       the corresponding GPU index (0, 1, 2, ...)
     * By default, when ``device`` indicates preference for GPU computations,
--- a/doc/install.txt
+++ b/doc/install.txt
@@ -375,7 +375,7 @@ If ``theano-nose`` is not found by your shell, you will need to add
 
     If you want GPU-related tests to run on a specific GPU device, and not
     the default one, you should use :attr:`~config.init_gpu_device`.
-    For instance: ``THEANO_FLAGS=device=cpu,init_gpu_device=gpu1``.
+    For instance: ``THEANO_FLAGS=device=cpu,init_gpu_device=cuda1``.
 
     See :ref:`libdoc_config` for more information on how to change these
     configuration options.
@@ -508,25 +508,25 @@ Any one of them is enough.
 
     :ref:`Ubuntu instructions <install_ubuntu_gpu>`.
 
-
+Next, install `libgpuarray <http://deeplearning.net/software/libgpuarray/installation.html>`_.
 
 Once that is done, the only thing left is to change the ``device`` option to name the GPU device in your
 computer, and set the default floating point computations to float32.
-For example: ``THEANO_FLAGS='cuda.root=/path/to/cuda/root,device=gpu,floatX=float32'``.
+For example: ``THEANO_FLAGS='cuda.root=/path/to/cuda/root,device=cuda,floatX=float32'``.
 You can also set these options in the .theanorc file's ``[global]`` section:
 
      .. code-block:: cfg
 
         [global]
-        device = gpu
+        device = cuda
         floatX = float32
 
 Note that:
 
-    * If your computer has multiple GPUs and you use 'device=gpu', the driver
+    * If your computer has multiple GPUs and you use 'device=cuda', the driver
       selects the one to use (usually gpu0).
     * You can use the program nvida-smi to change this policy.
-    * You can choose one specific GPU by specifying 'device=gpuX', with X the
+    * You can choose one specific GPU by specifying 'device=cudaX', with X the
       the corresponding GPU index (0, 1, 2, ...)
     * By default, when ``device`` indicates preference for GPU computations,
       Theano will fall back to the CPU if there is a problem with the GPU.
@@ -793,6 +793,8 @@ setup CUDA, but be aware of the following caveats:
    * If CUDA seems unable to find a CUDA-capable GPU, you may need to manually
      toggle your GPU on, which can be done with
      `gfxCardStatus <http://codykrieger.com/gfxCardStatus>`__.
+
+Next, install `libgpuarray <http://deeplearning.net/software/libgpuarray/installation.html>`_.
 
 Once your setup is complete, head to :ref:`using_gpu` to find how to verify
 everything is working properly.
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -103,7 +103,7 @@ import theano and print the config variable, as in:
 .. attribute:: device
 
     String value: either ``'cpu'``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
-    ``'opencl0'``, or ``'opencl1'`` ...
+    ``'opencl0:0'``, or ``'opencl0:1'`` ...
 
     Default device for computations. If ``'cuda*``, change the default to try
     to move computation to the GPU using CUDA libraries. If ``'opencl*'``,
@@ -135,7 +135,7 @@ import theano and print the config variable, as in:
 .. attribute:: init_gpu_device
 
     String value: either ``''``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
-    ``'opencl0'``, or ``'opencl1'`` ...
+    ``'opencl0:0'``, or ``'opencl0:1'`` ...
     
     Initialize the gpu device to use.
     When its value is cuda* or opencl*, the theano flag :attr:`device` must
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -103,13 +103,16 @@ import theano and print the config variable, as in:
 .. attribute:: device
 
     String value: either ``'cpu'``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
-    ``'opencl0:0'``, or ``'opencl0:1'`` ...
+    ``'opencl0:0'``, ``'opencl0:1'``, ``'gpu'``, ``'gpu0'`` ...
 
     Default device for computations. If ``'cuda*``, change the default to try
     to move computation to the GPU using CUDA libraries. If ``'opencl*'``,
     the openCL libraries will be used. To let the driver select the device,
-    use ``'cuda'`` or ``'opencl'``. If we are not able to use the GPU,
-    either we fall back on the CPU, or an error is raised, depending on the :attr:`force_device` flag.
+    use ``'cuda'`` or ``'opencl'``. If ``'gpu*'``, the old gpu backend will
+    be used, although users are encouraged to migrate to the new GpuArray 
+    backend. If we are not able to use the GPU,
+    either we fall back on the CPU, or an error is raised, depending
+    on the :attr:`force_device` flag.
 
     This flag's value cannot be modified during the program execution.
 
@@ -135,11 +138,11 @@ import theano and print the config variable, as in:
 .. attribute:: init_gpu_device
 
     String value: either ``''``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
-    ``'opencl0:0'``, or ``'opencl0:1'`` ...
+    ``'opencl0:0'``, ``'opencl0:1'``, ``'gpu'``, ``'gpu0'`` ...
     
     Initialize the gpu device to use.
-    When its value is cuda* or opencl*, the theano flag :attr:`device` must
-    be ``"cpu"``.
+    When its value is ``'cuda*'``, ``'opencl*'`` or ``'gpu*'``, the theano
+    flag :attr:`device` must be ``'cpu'``.
     Unlike :attr:`device`, setting this flag to a specific GPU will not
     try to use this device by default, in particular it will **not** move
     computations, nor shared variables, to the specified GPU.
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -51,11 +51,11 @@ Environment Variables
 
     .. code-block:: bash
 
-        THEANO_FLAGS='floatX=float32,device=gpu0,lib.cnmem=1'  python <myscript>.py
+        THEANO_FLAGS='floatX=float32,device=cuda0,lib.cnmem=1'  python <myscript>.py
 
     If a value is defined several times in ``THEANO_FLAGS``,
     the right-most definition is used. So, for instance, if
-    ``THEANO_FLAGS='device=cpu,device=gpu0'``, then gpu0 will be used.
+    ``THEANO_FLAGS='device=cpu,device=cuda0'``, then cuda0 will be used.
 
 .. envvar:: THEANORC
 
@@ -70,7 +70,7 @@ Environment Variables
 
         [global]
         floatX = float32
-        device = gpu0
+        device = cuda0
 
         [lib]
         cnmem = 1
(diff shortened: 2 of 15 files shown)


nouiz:
I think we should wait for speed regression to be fixed before merging. Otherwise seem ok.

doc/tutorial/using_gpu.txt
```diff
@@ -15,355 +15,9 @@ about how to carry out those computations.  One of the ways we take
...
-   * Circumvent the automatic cast of *int32* with *float32* to *float64*:
-
-     * Insert manual cast in your code or use *[u]int{8,16}*.
-     * Insert manual cast around the mean operator (this involves division by length, which is an *int64*).
-     * Notice that a new casting mechanism is being developed.
-
-:download:`Solution<using_gpu_solution_1.py>`
-
--------------------------------------------
+There are two ways currently to use a gpu, on that should support any OpenCL
```
abergeron: on that -> one that

doc/tutorial/using_gpu.txt
```diff
@@ -453,11 +109,11 @@ Returning a Handle to Device-Allocated Data
...
 results are returned to ensure a consistent interface with CPU code.
-This allows changing the deivce some code runs on by only replacing
+This allows changing the device some code runs on by only replacing
 the value of the ``device`` flag without touching the code.
 
 If you don't mind a loss of flexibility, you can ask theano to return
-the GPU object directly.  The following code is modifed to do just that.
+the GPU object directly.  The following code is modified to do just that.
 
 .. testcode::
```
abergeron: The example code below should be modified to use .transfer(None).

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +188,68 @@ What Can be Accelerated on the GPU
...
+* Prefer constructors like ``matrix``, ``vector`` and ``scalar`` to
+  ``dmatrix``, ``dvector`` and ``dscalar`` because the former will give
+  you *float32* variables when ``floatX=float32``.
+* Ensure that your output variables have a *float32* dtype and not *float64*.
+  The more *float32* variables are in your graph, the more work the GPU can do for
+  you.
+* Minimize transfers to the GPU device by using ``shared`` *float32* variables
+  to store frequently-accessed data (see :func:`shared()<shared.shared>`).
+  When using the GPU, *float32* tensor ``shared`` variables are stored on
+  the GPU by default to eliminate transfer time for GPU ops using those variables.
```
abergeron: This part doesn't apply to float32 only anymore

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +188,68 @@ What Can be Accelerated on the GPU
...
 
+* Consider adding ``floatX=float32`` (or the type you are using) to your
+  ``.theanorc`` file if you plan to do a lot of GPU work.
+* Use the Theano flag ``allow_gc=False``. See :ref:`gpu_async`
+* Prefer constructors like ``matrix``, ``vector`` and ``scalar`` to
+  ``dmatrix``, ``dvector`` and ``dscalar`` because the former will give
+  you *float32* variables when ``floatX=float32``.
+* Ensure that your output variables have a *float32* dtype and not *float64*.
+  The more *float32* variables are in your graph, the more work the GPU can do for
+  you.
```
abergeron: This is also not true anymore

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +188,68 @@ What Can be Accelerated on the GPU
...
 
-In general, large operations like matrix multiplication, or
-element-wise operations with large inputs, will be significatly
-faster.
+Tips for Improving Performance on GPU
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Consider adding ``floatX=float32`` (or the type you are using) to your
+  ``.theanorc`` file if you plan to do a lot of GPU work.
+* Use the Theano flag ``allow_gc=False``. See :ref:`gpu_async`
```
abergeron: This doesn't make any difference now.

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +188,68 @@ What Can be Accelerated on the GPU
...
+* Prefer constructors like ``matrix``, ``vector`` and ``scalar`` to
+  ``dmatrix``, ``dvector`` and ``dscalar`` because the former will give
+  you *float32* variables when ``floatX=float32``.
```
abergeron: It doesn't really matter anymore.

doc/tutorial/using_gpu.txt
abergeron: Although we can probably mention that computations in float64 are much slower.

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +188,68 @@ What Can be Accelerated on the GPU
...
+  Xs(X%) in transfer op*. This can tell you if not enough of your graph is
+  on the GPU or if there is too much memory transfer.
+* Use nvcc options. nvcc supports those options to speed up some computations:
+  `-ftz=true` to `flush denormals values to zeros.
+  <https://developer.nvidia.com/content/cuda-pro-tip-flush-denormals-confidence>`_,
+  `--prec-div=false` and `--prec-sqrt=false` options to speed up
+  division and square root operation by being less precise. You can
+  enable all of them with the `nvcc.flags=--use_fast_math` Theano
+  flag or you can enable them individually as in this example:
+  `nvcc.flags=-ftz=true --prec-div=false`.
```
abergeron: This doesn't apply anymore.

doc/tutorial/using_gpu_solution_1.py
```diff
@@ -101,171 +88,171 @@
...
-  17.2%    62.8%       0.492s       2.46e-05s     C    20002       4   theano.sandbox.cuda.blas.GpuGemv
-  15.1%    77.9%       0.433s       2.17e-05s     C    20001       3   theano.sandbox.cuda.basic_ops.GpuAlloc
-   8.2%    86.1%       0.234s       1.17e-05s     C    20002       4   theano.sandbox.cuda.basic_ops.HostFromGpu
-   7.2%    93.3%       0.207s       2.07e-05s     C    10000       1   theano.sandbox.cuda.basic_ops.GpuCAReduce
-   4.4%    97.7%       0.127s       1.27e-05s     C    10003       4   theano.sandbox.cuda.basic_ops.GpuFromHost
-   0.9%    98.6%       0.025s       8.23e-07s     C    30001       4   theano.sandbox.cuda.basic_ops.GpuDimShuffle
-   0.7%    99.3%       0.020s       9.88e-07s     C    20001       3   theano.tensor.elemwise.Elemwise
-   0.5%    99.8%       0.014s       7.18e-07s     C    20001       3   theano.compile.ops.Shape_i
-   0.2%   100.0%       0.006s       5.78e-07s     C    10000       1   theano.tensor.elemwise.DimShuffle
+  59.5%    59.5%       2.329s       1.16e-04s     C    20001       3   theano.sandbox.gpuarray.blas.GpuGemv
```
abergeron: Clearly something is wrong with GpuGemv here since it's 10 times slower on a faster GPU.

doc/tutorial/using_gpu.txt
slefrancois: I changed the code and it runs fine.

However, there is no more speedup from explicitely transfering the variable to GPU. On a GTX 680, both examples from the tutorial take about 2.2 sec.

abergeron: It should not be any faster than the GpuFromHost code, it's just a better interface that we want people to use.

doc/tutorial/using_gpu.txt
```diff
@@ -126,7 +126,8 @@ the GPU object directly.  The following code is modified to do just that.
 
   rng = numpy.random.RandomState(22)
   x = shared(numpy.asarray(rng.rand(vlen), config.floatX))
-  f = function([], gpuarray.basic_ops.GpuFromHost(None)(tensor.exp(x)))
+  gx = x.transfer(None)  # Transfer variable to GPU
+  f = function([], tensor.exp(gx))
```
abergeron: This is not the right change.  The .transfer should replace what GpuFromHost was doing i.e.:
```
f =  function([], tensor.exp(x).transfer(None))
```

slefrancois: Thanks, now it's faster with the explicit transfer.

nouiz:
I think we should make it clear in the doc that people should use the Dev version until Theano 0.9 is release. I don't want people using the new back-end and using theano 0.8.*.

slefrancois:
I'm adding a warning on the tutorial page.

nouiz:
We should also doc cuda* and opencl* in doc/config.py for the flags device and init_gpu_device.

slefrancois:
Done

doc/library/config.txt
```diff
@@ -102,22 +102,18 @@ import theano and print the config variable, as in:
 
 .. attribute:: device
 
-    String value: either ``'cpu'``, ``'gpu'``, ``'gpu0'``, ``'gpu1'``,
-    ``'gpu2'``, or ``'gpu3'``
+    String value: either ``'cpu'``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
+    ``'opencl0'``, or ``'opencl1'`` ...
```
nouiz: I think this isn't the good opencl syntax.

doc/library/config.txt
```diff
@@ -138,11 +134,12 @@ import theano and print the config variable, as in:
 
 .. attribute:: init_gpu_device
 
-    String value: either ``''``, ``'gpu'``, ``'gpu0'``, ``'gpu1'``, ``'gpu2'``,
-    or ``'gpu3'``
-
+    String value: either ``''``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
+    ``'opencl0'``, or ``'opencl1'`` ...
```
nouiz: idem

lamblin:
In any case, now the default doc is the one for 0.8.2, so these instructions will only show up in the documentation for the dev branch.

doc/install.txt
```diff
@@ -508,25 +508,25 @@ Any one of them is enough.
...
         [global]
-        device = gpu
+        device = cuda
         floatX = float32
 
 Note that:
 
-    * If your computer has multiple GPUs and you use 'device=gpu', the driver
+    * If your computer has multiple GPUs and you use 'device=cuda', the driver
       selects the one to use (usually gpu0).
```
lamblin: "usually cuda0"

doc/install.txt
```diff
@@ -508,25 +508,25 @@ Any one of them is enough.
...
     * You can use the program nvida-smi to change this policy.
```
lamblin: "nvidia-smi" (the typo was already there, but while we are at it...)

doc/library/config.txt
```diff
@@ -102,22 +102,18 @@ import theano and print the config variable, as in:
 
 .. attribute:: device
 
-    String value: either ``'cpu'``, ``'gpu'``, ``'gpu0'``, ``'gpu1'``,
-    ``'gpu2'``, or ``'gpu3'``
+    String value: either ``'cpu'``, ``'cuda'``, ``'cuda0'``, ``'cuda1'``,
+    ``'opencl0:0'``, or ``'opencl0:1'`` ...
```
lamblin: Should we still mention that gpu / gpuX works and maps to the old, deprecated back-end?

doc/tutorial/using_gpu.txt
```diff
@@ -470,7 +133,7 @@ the GPU object directly.  The following code is modifed to do just that.
 
   rng = numpy.random.RandomState(22)
   x = shared(numpy.asarray(rng.rand(vlen), config.floatX))
-  f = function([], gpuarray.basic_ops.GpuFromHost(None)(tensor.exp(x)))
+  f = function([], tensor.exp(x).transfer('dev0'))
```
lamblin: Should that be `.transfer(None)`? That is what the text below says, and I don't think `dev0` was introduced in any way.

doc/tutorial/using_gpu.txt
```diff
@@ -485,11 +148,9 @@ the GPU object directly.  The following code is modifed to do just that.
   else:
       print('Used the gpu')
 
-Here the :func:`theano.gpuarray.basic_ops.GpuFromHost(None)` call
-means "copy input to the GPU", with ``None`` the default GPU context when not
-explicitly given. However during the optimization phase,
-since the result will already be on the gpu, it will be removed.  It is
-used here to tell theano that we want the result on the GPU.
+Here ``tensor.exp(x).transfer('None')`` means "copy ``exp(x)`` to the GPU",
```
lamblin: No quotes around `None`.

doc/tutorial/using_gpu.txt
```diff
@@ -485,11 +148,9 @@ the GPU object directly.  The following code is modifed to do just that.
...
+with ``None`` the default GPU context when not explicitly given.
+For information on how to set GPU contexts, see :ref:`tut_using_multi_gpu`. 
 
 The output is
```
lamblin: You would have to update the output below, it still says "Using device cuda0:" instead of "Mapped name None to device cuda0: ..."

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +192,61 @@ What Can be Accelerated on the GPU
...
+  to the device pay off.
+
+The backend supports all regular theano data types (float32, float64,
+int, ...), however GPU support varies and some units can't deal with
 double (float64) or small (less than 32 bits like int16) data types.
 You will get an error at compile time or runtime if this is the case.
 
-By default all inputs will get transferred to GPU.  You can prevent an
+By default all inputs will get transferred to GPU. You can prevent an
 input from getting transferred by setting its tag.target attribute to
```
lamblin: I'd put "tag.target" in monotype.

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +192,61 @@ What Can be Accelerated on the GPU
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Consider adding ``floatX=float32`` (or the type you are using) to your
+  ``.theanorc`` file if you plan to do a lot of GPU work.
+* The GPU backend supports *float64* variables, but they are still slower
+  to compute than *float32*. The more *float32*, the better GPU performance
+  you will get. 
+* Prefer constructors like ``matrix``, ``vector`` and ``scalar`` to
+  ``dmatrix``, ``dvector`` and ``dscalar`` because the former will give
+  you *float32* variables and ignore the type given to ``floatX``.
```
lamblin: This is said in a confusing way. `matrix`, etc., *will follow* `floatX`, so they will be float32 when `floatX==float32`.
`dmatrix`, etc. *will ignore* `floatX` and use *float64* all the time.

doc/tutorial/using_gpu.txt
```diff
@@ -532,23 +192,61 @@ What Can be Accelerated on the GPU
...
+  taking more time than its share, then if you know something about GPU
+  programming, have a look at how it's implemented in theano.gpuarray.
+  Check the line similar to *Spent Xs(X%) in cpu op, Xs(X%) in gpu op and
+  Xs(X%) in transfer op*. This can tell you if not enough of your graph is
+  on the GPU or if there is too much memory transfer.
+* To investigate whether all the Ops in the computational graph are
+  running on GPU, it is possible to debug or check your code by providing
+  a value to `assert_no_cpu_op` flag, i.e. `warn`, for warning, `raise` for
+  raising an error or `pdb` for putting a breakpoint in the computational
+  graph if there is a CPU Op.
```
lamblin: Unfortunately, that does nothing in the new back-end for the moment.
I'll make a PR for that.

lamblin:
Other than that, it looks good, thanks!

doc/tutorial/using_gpu.txt
lamblin: Done in #4530.

doc/library/config.txt
nouiz: agreed.

doc/tutorial/using_gpu.txt
slefrancois: Good catch, I pasted intermediate code where I tried setting contexts, but we opted for simplicity with the default.

doc/tutorial/using_gpu.txt
slefrancois: Which part exactly does nothing?

nouiz: The assert_no_cpu_op flag. The PR fix that, so you have nothing to change here.

lamblin:
OK for me.

PR #4502, opened 2016-05-11
Add release doc switcher

slefrancois:
Make 0.8.X release backward compatible with documentation version switcher.

### DIFF
--- a/doc/.static/version_switch.js
+++ b/doc/.static/version_switch.js
@@ -1,0 +1,78 @@
+// Create version selector for documentation top bar.
+
+(function() {
+
+    var url = window.location.href;
+    var theano_dir = 'theano'; // directory containing theano doc
+    // Default theano version: release and development.
+    var versions_dir = {"release": "theano", "dev":"theano-dev"};
+    // When doc is run from server
+    if (url.startsWith('http')) {
+        var root_url = url.substring(0, url.search('/' + theano_dir)) + '/';
+        // Get theano versions from dynamically generated version file on server.
+        $.getJSON(root_url + 'theano/versions.json', function(data){
+            $.each(data, function(version, dir) {
+                    versions_dir[version] = dir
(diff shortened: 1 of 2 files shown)


abergeron:
This does work?

Because I see it filling in the versions from the server in a callback and there is absolutely no synchronization with the rest of the code that uses the version.

I suspect that if it worked in your tests, you were lucky because the sever is right next from the machine you were testing from.

kmike:
Default readthedocs.org theme has nice versioning support; are Theano docs look worse with it? Versioning should be a matter of commenting out `html_theme = 'sphinxdoc'` in docs/conf.py (and maybe making some tweaks using readthedocs interface).

nouiz:
We don't use readthedocs. We set it as a backup from our file server.

Do you mean that readthedocs have code that allow us to as it is done on it
for multiple version and allow us to have this outside readthedocs?

kmike:
Readthedocs is [open source](https://github.com/rtfd/readthedocs.org), but I'm not sure how easy is it to easy to extract just this feature. I though http://theano.readthedocs.io/en/latest/ is more "canonical", not just a backup (e.g. for doc.scrapy.org we just use readthedocs - it supports custom domains). 

By changing a theme you can get versions at least at http://theano.readthedocs.io/. It is possible to do that just for readthedocs - see https://docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs.

slefrancois:
I based myself on the python doc's version switcher. If someone found readthedoc's version switcher in their source code, I would be curious to see it. But I suspect we would have to re-implement their server-side scripts too, which knows what versions to pull and where to put them in its file system.

slefrancois:
For now I'll close the pull request, it's been pointed out it's to have one merge on the master and then backport it to the release.

PR #4503, opened 2016-05-12
Fix a test regression when we moved the new back-end

nouiz:


nouiz:
this fix a test in the gpuarray daily buildbot

theano/gpuarray/tests/test_pickle.py
```diff
@@ -19,7 +19,7 @@
 from theano.misc.pkl_utils import CompatUnpickler
 
 try:
-    from . import config  # noqa
+    from .config import mode_with_gpu  # noqa
```
abergeron: How does this makes any difference?

nouiz: it was overiding the config import above. so confix.floatX would fail...

abergeron: Then we could use `from . import config as _`

nouiz: done

nouiz:
merging as I implemented the version @abergeron suggested.

PR #4505, opened 2016-05-12
d3viz improvements for Python 3

kmike:
* Python 3 compatibility is fixed;
* Install instructions are expanded;
* added an example on how to show d3viz interactive chart inline in IPython notebook

Without a fix d3viz doesn't work in Python 3: `graph.create_dot()` returns a bytes object, and so `str(graph.create_dot())` returns string like `"b'...'"` in Python 3.

I haven't changed [index.ipynb](https://github.com/Theano/Theano/blob/master/doc/library/d3viz/index.ipynb) - should I? Are these files synced manually?


doc/library/d3viz/index.txt
```diff
@@ -130,6 +130,19 @@ may not be perfect. In this case, you can press the ``Release node``
...
 nodes to their default position, press the ``Reset nodes`` button.
 
+You can also display the interactive graph inline in
+IPython using ``IPython.display.IFrame``:
+
+.. code:: python
+
+    from IPython.display import IFrame
+    d3v.d3viz(predict, 'examples/mlp.html')
+    Iframe('examples/mlp.html', width=700, height=500)
```
nouiz: With pydotprint, you can print directly like this:

```
            import theano
            v = theano.tensor.vector()
            from IPython.display import SVG
            SVG(theano.printing.pydotprint(v*2, return_image=True,
                                           format='svg'))
```

While working on this, do you want to implement that? I don't know that part well, but I think it should be easy. We probably need just to allow the file "str" to d3viz() as in debugprint() and there is probably a way to pass the data as a string to Iframe.

nouiz:
index.ipynb should be updated I think. It isn't automatically generated.

theano/d3viz/d3viz.py
```diff
@@ -78,7 +79,10 @@ def d3viz(fct, outfile, copy_deps=True, *args, **kwargs):
     # Create DOT graph
     formatter = PyDotFormatter(*args, **kwargs)
     graph = formatter(fct)
-    dot_graph = escape_quotes(str(graph.create_dot())).replace('\n', '').replace('\r', '')
+    dot_graph_raw = graph.create_dot()
+    if not six.PY2:
+        dot_graph_raw = dot_graph_raw.decode('utf8')
+    dot_graph = escape_quotes(dot_graph_raw).replace('\n', '').replace('\r', '')
```
nouiz: Travis skiped the associated tests as pydot wasn't installed. Do you want to modify the file .travis.yml
Change this line:
  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install pydot; fi

to:
  - pip install pydot-ng

I think it also exist for python 2.

nouiz:
Thanks for the PR, I got a few small comments.

theano/d3viz/d3viz.py
kmike: Are you OK with recommending pydot-ng instead of pydot everywhere in docs?

nouiz: yes.

doc/library/d3viz/index.txt
kmike: I believe it should be possible to implement that; the tricky part is to make sure related resources are available, but it should be doable. It will likely require a custom IPython extension which users will have to `%load_ext` though (similar to https://github.com/jovyan/pythreejs). I'm interested in working on that, but it probably doesn't belong to this ticket. Theano IPython extension can do more cool things than just providing js/css fils for d3viz.

I think embedding JS and CSS in a html string won't work because IPython sanitizes HTML displayed via display.HTML. Just did a quick prototype - d3viz no longer works after embedding all scripts and styles when displayed with display.HTML.

nouiz: thanks for checking. I'll merge this PR when travis pass.

nouiz:
The package pydot-ng is only available in python 2.7 and 3. But in travis we test with python 2.6. So for python 2.6, we need to install pydot.

nouiz:
The tests in python 3.3 also failed. pydot-ng is only available for python 3.4. Can you change the version of python tested to be 3.4?

@lamblin do you agree with that change? I prefer to test more.

kmike:
I went with installing pydot-ng with conda instead of a safer `pip install pydot-ng` because there is `conda install pydot-ng` in updated docs, and I was not sure it works :) `pip install pydot-ng` (like @nouiz initially suggested) should work fine in both 2.6 and 3.3. But maybe it is better to move away from Python 2.6 (https://github.com/Theano/Theano/issues/4507) as well? What do you think?

.travis.yml
```diff
@@ -35,7 +35,7 @@ install:
   - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then conda create --yes -q -n pyenv mkl python=2.6 numpy=1.7.1 scipy=0.11 nose=1.3.0 pyparsing=1.5 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '3.3' ]]; then conda create --yes -q -n pyenv mkl python=3.3 numpy=1.9.1 scipy=0.14.0 nose=1.3.4 pyparsing=1.5 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx; fi
   - source activate pyenv
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install pydot; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install pydot; else pip install pydot-ng; fi
```
kmike: Just `pip install python-ng` should work, but by installing pydot and pydot-ng we can check theano works with both.

nouiz: Good idea.

nouiz:
thanks

PR #4506, opened 2016-05-13
Don't use march flags in cutils to help don't have the architecture m…

nouiz:
…ismatch.

There isn't many real code execution, so I suppose it won't affect performance.


lamblin:
Well, there's the C code for AdvancedIncSubtensor on CPU, but I guess it's OK to prevent crashes.

nouiz:
but that code call numpy c api and the computation is there. So this would
affect only the overhead I think.

PR #4511, opened 2016-05-13
Corrected contradictions in sparse doc

slefrancois:


nouiz:
merging as the travis doc part finished.

PR #4512, opened 2016-05-13
d3viz: fix escaping of dot graph

### LINKED ISSUE #4510: d3viz doesn't work for some graphs
Sometimes d3viz doesn't work for me - empty graph is displayed, and there is an error message in JS console. I tried to reduce it to a minimal example

```py
import os    
os.environ['THEANO_FLAGS'] = "device=gpu0,floatX=float32" #,optimizer=fast_compile"

import numpy as np
from lasagne.updates import adam
from theano import tensor as T, shared, function
import theano.d3viz as d3v

def show_d3(g):
    d3v.d3viz(g, 'example.html')
    from IPython.display import IFrame
    return IFrame('example.html', width=800, height=500)

x = T.fvector()
W = shared(np.zeros((10, 5), dtype=np.float32))
b = shared(np.zeros((10,), dtype=np.float32))
y_true = T.fvector()

y = T.nnet.sigmoid(T.dot(x, W) + b)
cost = T.sqrt(((y - y_true)**2).sum())

updates = adam(cost, [W])  # no b!
f_cost = function([x, y_true], cost, updates=updates)
show_d3(f_cost)
```

(IPython notebook: https://gist.github.com/kmike/13b0fb747dccd4f2f1e44789a9cb832c).
This is brittle: if `adam` is replaced with any other training method from Lasagne (or if a simple SGD is implemented manually) chart works; if `T.sqrt` is removed from `cost` expression chart works; if `T.nnet sigmoid` is removed from `y` expression chart works; if `b` is added to adam updates chart works; if `optimizer=fast_compile` is added to THEANO_FLAGS chart works.

kmike:
This fixes #4510 for me. The issue is hard to reproduce - the example from #4510 worked on another machine with CNMeM disabled and a different CuDNN version. But I think using JSON instead of hand-rolled escaping is better anyways. The only tricky part is `<` escaping; it is copied from https://github.com/scrapinghub/splash/pull/386.

### DIFF
--- a/theano/d3viz/d3viz.py
+++ b/theano/d3viz/d3viz.py
@@ -5,8 +5,8 @@ Author: Christof Angermueller <cangermueller@gmail.com>
 from __future__ import absolute_import, print_function, division
 
 import os
+import json
 import shutil
-import re
 import six
 from six import iteritems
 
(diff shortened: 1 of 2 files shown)


nouiz:
thanks

PR #4518, opened 2016-05-16
Fix nanguardmode with slices

nouiz:
reported at: [theano-users] The latest 0.9.0dev version seems not to support NaNGuard mode for the slice object


abergeron:
Ok when travis passes.

PR #4524, opened 2016-05-17
Opt related changes.

nouiz:
- Add opt UnShapeOpt that remove at step 10 the ShapeFeature. This can speed up compilation time.
   - It only disable opt local_elemwise_alloc in the old gpu back-end. I don't know why it was there.
- profiling of opt callbacks.
- Add EquilibriumDB the option tracks_on_change_inputs. When tested manually, speed up the move to the GPU, but the final graph is bigger.
- mark for fast_compile cudnn opt in the new back-end


nouiz:
travis is passing.

theano/gof/opt.py
```diff
@@ -870,7 +898,9 @@ def print_profile(stream, prof, level=0):
             print(blanc, "  callbacks_time", file=stream)
             for i in sorted(iteritems(callbacks_time), key=lambda a: a[1]):
                 if i[1] > 0:
-                    print(i)
+                    # We want to have the __str__ called, so we can't
+                    # just print i.
+                    print(blanc, "      ", i[0], ',', i[1], file=stream)
```
abergeron: Why do we have two spots printing almost the same thing?

theano/gpuarray/opt.py
```diff
@@ -710,18 +710,14 @@ def local_gpua_careduce(node, context_name):
                     assert reduce_mask[a] == 0
                     reduce_mask[a] = 1
 
-            shape_of = node.fgraph.shape_feature.shape_of
-
-            x_shape = shape_of[x]
-
-            new_in_shp = [x_shape[0]]
+            new_in_shp = [shape_i(x, 0)]
```
abergeron: Why drop shape_of?

theano/sandbox/cuda/__init__.py
```diff
@@ -23,6 +23,8 @@
 # we use for optimization. Otherwise, we can iterate 100s of time on
 # the graph and apply only a few optimizations each time.
 gpu_optimizer = EquilibriumDB(ignore_newtrees=False)
+#gpu_optimizer = EquilibriumDB(ignore_newtrees=True,#False,
+#                              tracks_on_change_inputs=True)#True)
```
abergeron: This looks like debugging code which should not stay.

theano/tensor/opt.py
```diff
@@ -1430,19 +1436,29 @@ def same_shape(self, x, y, dim_x=None, dim_y=None):
...
     def apply(self, fgraph):
         pass
 
+
+class UnShapeOptimizer(Optimizer):
+    """Optimizer remove ShapeFeature as an fgraph feature."""
+    def apply(self, fgraph):
+        for feature in fgraph._features:
+            if isinstance(feature, ShapeFeature):
+                fgraph.remove_feature(feature)
```
abergeron: This looks like it could invoke old ones.  I'm not sure it's a good things.

theano/gof/opt.py
nouiz: It is in 2 different optimizer class.

theano/tensor/opt.py
nouiz: I don't understand why it would invoke old one. There can be only 1 shape feature per fgraph. ShapeFeature.on_attach assert that.

theano/sandbox/cuda/__init__.py
nouiz: done

theano/tensor/opt.py
abergeron: It was a metaphoric figure of speech saying that this looks like fragile code.

I really dislike using optimizers for all sorts of things that aren't optimizations and I would rather we spend some time figuring out a solution than pushing out a quick hack.

As for the old ones, see here: https://en.wikipedia.org/wiki/Old_One_in_fiction

theano/gpuarray/opt.py
abergeron: What about this?

nouiz: shape_i will use it if shape_of is available.

theano/tensor/opt.py
nouiz: I agree, we could find a better solution. But currently this is the only way I know how we can make it work.

I don't find an alternative in a few minutes. I would go for making an issue about this problem. I don't think we should fix it here. If you have suggestion of alternative way of doing this write it.

theano/gpuarray/opt.py
abergeron: ok

abergeron:
I'm ok with this to the exception of `UnShapeOptimizer`, but apparently we don't have time for a better solution...

Issue #3376, opened 2015-09-10
Theano on Windows with Python 3.5 probably don't work


nouiz:
see: http://stevedower.id.au/blog/building-for-python-3-5/

Maybe we should switch to MSVC directly and not mingw? Or just wait to have mingw support it?

patrikhuber:
That's a very nice write-up! I haven't gone through the whole blog post, but I just wanted to add that I'm successfully (modulo a few minor issues) using Theano with [msys2-mingw64](msys2.github.io) and python3.5. I've installed their mingw64 python3.5, numpy, scipy, pip3 from pacman, and then installed Theano from pip3.

LexSong:
Does mingw64 python work with CUDA?

patrikhuber:
@LexSong: I don't know, I'm not using CUDA currently, but msys2's Python has been really great so far and it looks pretty standard. Without having any experience with it, I'd say "Probably, why not".

nouiz:
"Probably, why not" on windows with CUDA make me smile. I hope it is now as
easy is it could be, history told me it isn't as easy as we would like. But
someone need to try it to know. Here we won't have time shortly to try it,
so if someone try it, please report result. This can help other people.

patrikhuber:
Hehe yea, I can only imagine! I currently can't try with CUDA, maybe in the future. Hopefully other people can chime in :-) It would be great if it worked with msys2-mingw64, since the distribution is awesome and useful for many purposes, so you don't have to have several mingw distributions and each of them comes with their own gcc and takes up space etc. (e.g. most Python distributions for Windows come with their own mingw).

ErnstDinkelmann:
Something interesting that may just help others.
**My environment:** 
Win8.1, 64bit, Python 3.5, Anaconda4 install, MS Visual Studio 12.0, CUDA 7.5, OpenBLAS

**I also experienced the C/C++ compiler and importing issues when I had**
`cxx = C:\TDM-GCC-64\bin\g++.exe`
This is recommended in a number of installation tutorials online.

**The errors I experienced:**
File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 2016.1.2\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
ImportError: DLL load failed: A dynamic link library (DLL) initialization routine failed.

**However, when I changed to:**
`cxx = C:\MinGw64\mingw64\bin\g++.exe`
[...]

PR #4525, opened 2016-05-17
Better error related to gh-3376

nouiz:


theano/__init__.py
```diff
@@ -29,6 +29,12 @@
 # theano code, since this code may want to log some messages.
 import logging
 
+import sys
+
+if sys.platform == 'win32' and sys.version_info[0:2] == (3, 5):
+    raise RuntimeError(
+        "Theano do not support Python 3.5 on Windows. Use Python 2.7 or 3.4.")
```
abergeron: Since when?

nouiz: Since python 3.5 exist as it change all the way the c linking is done on windows. See gh-3376

PR #4526, opened 2016-05-17
Add an include of gpuarray/buffer.h for future work.

abergeron:


lamblin:
Should we merge this one before, after, or at the same time as the gpuarray change?

nouiz:
local tests pass.

nouiz:
We should merge now.

This keep it currently working and will help support futur gpuarray changes.

PR #4409, opened 2016-04-19
cudnn slowdown on new back-end

gyom:
This pull request comes at after a discussion with Fred.

I have noticed that the new back-end was considerably slower than the old back-end on a very straightforward model that I am running.

With old back-end and config.dnn.enabled = False, my code took 40 seconds.
With old back-end and config.dnn.enabled = True, my code took 112 seconds. This PR brings that time down to 40 seconds again.

With new back-end and config.dnn.enabled = True, my code took 170 seconds. This PR brings that time down to 99 seconds.

I tried all the variations of
    n.dimshuffle(0, 1, ‘x’, ‘x’)
    n.dimshuffle(0, ‘x’, 1, ‘x’)
    n.dimshuffle(0, ‘x’, ‘x’, 1)
to get the fastest.

PR #4532, opened 2016-05-18
Speed up GpuDnnSoftmaxGrad by using the best mapping.

nouiz:
replace gh-4409


lamblin:
Should we do that for the forward logsoftmax as well? Or was the gradient the only slowdown?
Otherwise, I'm fine with merging. I'll let you decide if you want to do that before the NIPS deadline or after.

nouiz:
I don't really care before after, but for extra safety, we can wait
tomorrow afternoon. Just in case.

We didn't saw this problem in the forward.

lamblin:
OK. Merging because it could have a speed impact.

PR #4530, opened 2016-05-17
Port local_assert_no_cpu_op to the new back-end

lamblin:

PR #4533, opened 2016-05-18
Fix gh-4530, we can't have 2 opt with the same tag

nouiz:


nouiz:
merging now as this is a simple regression fix.

lamblin:
Thanks!

PR #4535, opened 2016-05-18
Port FFTs to gpuarray backend

slefrancois:


theano/gpuarray/fft.py
```diff
@@ -0,0 +1,242 @@
...
+
+from theano.gpuarray import (basic_ops, GpuArrayType)
+
+try:
+    import pygpu
+    pygpu_available = True
+except ImportError:
+    pygpu_available = False
+
+import pycuda.driver
```
abergeron: This import has to be guarded.

abergeron:
Ok, modulo my comment.

slefrancois:
Done, it was also producing a Travis error.

slefrancois:
I pushed preliminary gradient code that wasn't ready, new commit fixes that.

theano/gpuarray/fft.py
```diff
@@ -121,33 +129,23 @@ def thunk():
 
         return thunk
 
-cufft = CuFFTOp()
```
nouiz: do:
curfft = CuRFFTOp()

and reuse that in the method in this file.

theano/gpuarray/fft.py
```diff
@@ -230,13 +235,34 @@ def thunk():
 
         return thunk
 
-cuifft = CuIFFTOp()
```
nouiz: idem, make curifft here

slefrancois:
Hold off on merging, there are issues with normalizations and real FFTs when it comes to gradients. We might have to change the interface.

slefrancois:
Ready to merge.

doc/library/gpuarray/fft.txt
```diff
@@ -0,0 +1,44 @@
...
+==============================================
+
+Performs Fast Fourier Transforms (FFT) on the GPU.
+
+FFT gradients are implemented as the opposite Fourier transform of the output gradients.
+
+.. warning ::
+    The real and imaginary parts of the Fourier domain arrays are stored as a pair of float32
+    array, emulating complex64. Since theano does not support complex
+    number operations, care must be taken to manually implement operators such as complex
```
nouiz: Theano support them in limited way. But we don't support the grad on it. So maybe tell that we don't support grad on complex?

theano/gpuarray/fft.py
```diff
@@ -0,0 +1,368 @@
...
+                                       batch=input_shape[0])
+
+                # Sync GPU variables before computation
+                input_pycuda.sync()
+                output_pycuda.sync()
+
+                fft.fft(input_pycuda, output_pycuda, plan[0])
+
+                # Sync results to ensure output contains completed computation
+                pycuda.driver.Context.synchronize()
```
nouiz: @abergeron is this needed? Can we add an event in the output variable? I would let this for the futur, but I would like to know. We probably need to make an issue about which op aren't async.

nouiz:
I just did a quick overview of this PR.

slefrancois:
New commit to skip test requiring GPU libraries.

slefrancois:
This is ready to review and merge.

It would be worth checking if the gpu op_lifters are as we want. I did it so that they are ignored if scikits_cuda is not present. However, I have seen other gpu ops that register the optimization, but return an error in the lifter if optional dependencies are absent.

nouiz:
travis have some failure.

theano/gpuarray/fft.py
abergeron: This might not be required anymore, now that no stream in libgpuarray is async with regards to the null stream.

PR #4540, opened 2016-05-19
Finish the regression fix in gh-4533

nouiz:
without this fix, there is a test failure in the buildbot.


nouiz:
I also added other test fix in python 32bit

abergeron:
Ok when travis passes.

PR #4543, opened 2016-05-20
Make BN graph smaller

nouiz:
This PR need speed tests.
This should speed up compilation while using Theano BN.
This also fix other problems with Elemwise with multiple outputs.


lachlants:
Hello, I ran into a few issues testing this PR with master:

1.  local_gpu_elemwise_1 doesn't like it, missing owner attribute
> ERROR (theano.gof.opt): Optimization failure due to: local_gpu_elemwise_1
> ERROR (theano.gof.opt): node: GpuFromHost(Elemwise{Composite{((i4 * i3) / i2), (-((i4 * i3) / i2)), ((-((i4 * i3) * (i0 - i1))) / (i2 * i2)), ((i4 * (i0 - i1)) / i2), identity(i4)}}.1)
> ERROR (theano.gof.opt): TRACEBACK:
> ERROR (theano.gof.opt): Traceback (most recent call last):
>   File "/home/users/lachlan.ts/src/Theano/theano/gof/opt.py", line 1779, in process_node
>     replacements = lopt.transform(node)
>   File "/home/users/lachlan.ts/src/Theano/theano/sandbox/cuda/opt.py", line 364, in local_gpu_elemwise_1
>     gpu_elemwise = split_huge_add_or_mul(gpu_elemwise.owner)
> AttributeError: 'list' object has no attribute 'owner'

2. Elemwise._c_all() has a bug for multiple aliasing outputs, needs a newline char between each "#define" / "#undef" (line 1102-1103 in elemwise.py).

theano/scalar/basic.py
```diff
@@ -3605,30 +3610,56 @@ def __init__(self, inputs, outputs):
...
             inputs, outputs = gof.graph.clone(inputs, outputs)
         else:
             # Inner Composite that we need to flatten
-            assert len(outputs) == 1
             # 1. Create a new graph from inputs up to the
             # Composite
+            out1 = []
+            for o in outputs:
+                for i in o.owner.inputs:
+                        out1.append(i)
```
kmike: the indentation looks off

nouiz:
I fixed all that.

nouiz:
I pushed a few other optimizations that could be useful.

I think we still need to udpate the user interface to don't always use the new code, but allow to use the old code too.

theano/scalar/basic.py
```diff
@@ -3605,30 +3610,56 @@ def __init__(self, inputs, outputs):
...
+                        outputs=o.owner.op.outputs,
+                        replace=repl)
+                    out = res2[1]
+                    node_done[o.owner] = out
+                new_outputs.append(out[o.owner.outputs.index(o)])
+            assert len(new_outputs) == len(outputs)
             assert len(res[0]) == len(inputs)
             assert res[0] != inputs
-            inputs, outputs = res[0], res2[1]
+            inputs, outputs = res[0], new_outputs
```
abergeron: This part is supposed to flatten Composites, but this may not work (and duplicate work).  Consider the following graph:

o1 = Composite(i1 + i2)
o2 = o1 + i3

You will replace o1 and o2 as outputs, but there will still be a copy of o1 in the inputs of o2.

theano/tensor/opt.py
```diff
@@ -6919,6 +6920,43 @@ def local_add_mul_fusion(node):
...
+
+    # Remove outputs not used.
+    idx = [i for i, o_extern in enumerate(node.outputs)
+           if o_extern.clients]
+    if len(idx) < len(node.outputs):
+        new_outputs = [comp.outputs[i] for i in idx]
+        c = scalar.Composite(inputs=comp.inputs,
+                             outputs=new_outputs)
+        e = Elemwise(scalar_op=c)(*node.inputs, return_list=True)
+        comp = e[0].owner.op.scalar_op
```
abergeron: Shouldn't we try to get rid of useless inputs too?  Unless it already does that.

abergeron:
What do you mean by "old code" here?  I would prefer that we move everything to one code path rather than try to support two.

nouiz:
by "old code" I mean the old implementation. There is probably speed difference between the 3 version and I don't want to remove those option from the user. That was requested by users in fact.

theano/scalar/basic.py
nouiz: I don't understand. I tried to implement this case as a test and it pass:

```
+        # Test a corner case
+        z = theano.scalar.float32()
+        c = theano.scalar.Composite([x, y], [x+y])
+        cc = theano.scalar.Composite([x, y, z], [c(x, y), x+z])
+        Z = tensor.matrix()
+        o = tensor.Elemwise(scalar_op=cc)(X, Y, Z)
+
+        f = theano.function([X, Y, Z], o, mode=mode)
+        topo = f.maker.fgraph.toposort()
+        assert len(topo) == 1
+        assert len(topo[0].outputs) == 2
+        assert isinstance(topo[0].op, tensor.Elemwise)
+        assert isinstance(topo[0].op.scalar_op, theano.scalar.Composite)
+        utt.assert_allclose(f([[1.]], [[3.]], [[5.]])[0], [[4.]])
+        utt.assert_allclose(f([[1.]], [[3.]], [[5.]])[1], [[6.]])
```

Is this the case that you talk about?

theano/tensor/opt.py
nouiz: done

theano/scalar/basic.py
abergeron: No, this would be the test case:

        c = theano.scalar.Composite([x, y], [x+y])
        p = c(x, y)
        cc = theano.scalar.Composite([x, y, z], [p, c(p, z)])

Or something like that.

abergeron:
Do you mean "old code" for the grad of the BN composite?  We have three versions of this?

Issue #4534, opened 2016-05-18
Shape inconsistency for strided pooling with ignore_border=False


f0k:
As reported in Lasagne/Lasagne/issues/685, there's something wrong with Theano's pooling implementation for `ignore_border=False`. Reduced test case:
```
>>> import numpy as np
>>> import theano
>>> import theano.tensor as T
>>> x = np.ones((1,1,4,4), dtype=theano.config.floatX)
>>> y = T.signal.pool.pool_2d(T.constant(x), (4,4), ignore_border=False, st=(2,2))
>>> y.shape.eval()
(1, 1, 1, 1)
>>> y.eval().shape
(1, 1, 2, 2)
```
So we pool a 4x4 image with a 4x4 window with stride 2. The result should be 1x1, regardless of whether we ignore the border or not.
The evaluated shape (`y.shape.eval()`) is correct, but the shape of the evaluated value (`y.eval().shape`) is too large.

nouiz:
I have a fix, I still need to finish it and add tests.

nouiz:
thanks the problem seem to be only when the window is bigger then the image. I have a PR:

https://github.com/Theano/Theano/pull/4544

PR #4544, opened 2016-05-20
Pool shape

nouiz:
Don't merge yet, I want to double something first. But it can be reviewed now.

fix gh-4534

### DIFF
--- a/theano/tensor/signal/pool.py
+++ b/theano/tensor/signal/pool.py
@@ -220,7 +220,7 @@ class Pool(Op):
             elif st[0] >= ds[0]:
                 nr = (r - 1) // st[0] + 1
             else:
-                nr = max(0, (r - 1 - ds[0]) // st[0] + 1) + 1
+                nr = max(0, (r - 1 - ds[0] + st[0]) // st[0]) + 1
 
             if isinstance(c, theano.Variable):
                 nc = tensor.switch(tensor.ge(st[1], ds[1]),
@@ -230,7 +230,7 @@ class Pool(Op):
             elif st[1] >= ds[1]:
                 nc = (c - 1) // st[1] + 1
             else:
-                nc = max(0, (c - 1 - ds[1]) // st[1] + 1) + 1
+                nc = max(0, (c - 1 - ds[1] + st[1]) // st[1]) + 1
 
         rval = list(imgshape[:-2]) + [nr, nc]
         return rval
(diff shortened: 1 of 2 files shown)


theano/tensor/signal/pool.py
```diff
@@ -403,7 +410,7 @@ def c_code(self, node, name, inp, out, sub):
             }
             else
             {
-                z_r = std::max(0, (r - 1 - %(ds0)s) / %(st0)s + 1) + 1;
+                z_r = std::max(0, (r - 1 - %(ds0)s + %(st0)s) / %(st0)s) + 1;
```
f0k: Looks good, that's what we have in Lasagne as well. But wouldn't this one need to be changed as well? https://github.com/nouiz/Theano/blob/ecc1b37773c284c2dc9e2b49aecc29ecb3f3ac0e/theano/tensor/signal/pool.py#L223 (Shape inference seemed to be correct, so I guess not, but why?)

nouiz: Python and C division don't behave the same. I updated them for consistency.

theano/tensor/signal/pool.py
```diff
@@ -216,7 +220,7 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
             elif st[0] >= ds[0]:
                 nr = (r - 1) // st[0] + 1
             else:
-                nr = max(0, (r - 1 - ds[0]) // st[0] + 1) + 1
+                nr = max(0, (r - 1 - ds[0] + st[0]) // st[0]) + 1
```
abergeron: Why is this better?

theano/tensor/signal/pool.py
```diff
@@ -522,7 +531,7 @@ def c_code(self, node, name, inp, out, sub):
         return ccode % locals()
 
     def c_code_cache_version(self):
-        return (0, 6, 8, 3)
+        return (0, 6, 8, 4)
```
abergeron: Could we remove some of those numbers?

theano/tensor/signal/pool.py
nouiz: Because in python and in C, division with negative number don't return the same value. It does the right thing for us in Python, but not in C. So in C we got wrong shapes. I prefer to use the same formula for consistency in Python and C. There is a test that make sure we keep it working.

theano/tensor/signal/pool.py
nouiz: I don't want to inspect the history to make sure to not create collision.

PR #4546, opened 2016-05-21
TensorInv op added, test for tensorinv added

iikulikov:
Hello, tensorinv() function from linalg added with corresponding test.
If I should add something there, please tell me.

### DIFF
--- a/theano/tensor/nlinalg.py
+++ b/theano/tensor/nlinalg.py
@@ -742,13 +742,17 @@ class TensorInv(Op):
     def make_node(self, a):
         a = as_tensor_variable(a)
         out_dtype = a.dtype
-        out = theano.tensor.matrix(dtype=out_dtype)
+        out = theano.tensor.tensor4(dtype=out_dtype)
         return Apply(self, [a], [out])
 
     def perform(self, node, inputs, outputs):
         (a,) = inputs
         (x,) = outputs
         x[0] = self._numop(a, self.ind)
+
+    def infer_shape(self, node, shapes):
+        sp = shapes[0][self.ind:] + shapes[0][:self.ind]
+        return [sp]
 
 
 def tensorinv(a, ind=2):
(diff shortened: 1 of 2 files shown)


theano/tensor/nlinalg.py
```diff
@@ -726,3 +726,59 @@ def norm(x, ord):
...
+    def __init__(self, ind=2):
+        self.ind = ind
+
+    def make_node(self, a):
+        a = as_tensor_variable(a)
+        out_dtype = a.dtype
+        out = theano.tensor.matrix(dtype=out_dtype)
+        return Apply(self, [a], [out])
+
+    def perform(self, node, inputs, outputs):
```
nouiz: Can you add the infer_shape method to that class and associated tests? Not having it can causes unexpected behavior to the user.

nouiz:
Except that small comment, all looks good. thanks.

iikulikov:
Sure, I will do that soon.

iikulikov:
Okay, I added this method and also restructured the test to the class with corresponding compile check method for infer shape.

iikulikov:
Now the input supposed to be a 4d tensor, since we already have matrix inv for 2d tensors. Is it okay? May be it is possible to use TensorType for output variable to keep the shape of the input tensor during compilation? Please give an advice.

theano/tensor/nlinalg.py
```diff
@@ -742,14 +742,18 @@ def __init__(self, ind=2):
     def make_node(self, a):
         a = as_tensor_variable(a)
         out_dtype = a.dtype
-        out = theano.tensor.matrix(dtype=out_dtype)
+        out = theano.tensor.tensor4(dtype=out_dtype)
```
nouiz: out isn't always a 4d tensor. For example:

python -c "import numpy as np; print np.linalg.tensorinv(np.random.rand(2,15,30)).shape"
(30, 2, 15)

python -c "import numpy as np; print np.linalg.tensorinv(np.random.rand(30,2,15),ind=1).shape"
(2, 15, 30)

I think out should be: `out=a.type()`

Add those 2 cases as a test.

iikulikov: okay, thanks for the hint, now it is working with a.type(), and I added tests for other tensors, infer_shape seems to works.

nouiz:
thanks

PR #4551, opened 2016-05-24
Added Global Opt to transfer Graph to GPU

Sentient07:
Added global opt to move graph to GPU


theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+    def add_requirements(self, fgraph):
+        fgraph.attach_feature(toolbox.ReplaceValidate())
+
+    def apply(self, fgraph):
+        mapping = {}
+        move_to_GPU = True
+
+        # Building a new graph
+        for i in fgraph.inputs:
+            mapping[i] = GpuFromHost(None)(i)
```
nouiz: only do that is if i is of type TensorVariable.

nouiz: otherwise, don't transfer it in the mappping.

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+
+        for node in fgraph.toposort():
+
+            # The Extra condition
+            if node.inputs is node.outputs:
```
nouiz: it should be something like:

if any([isinstance(i, CudaVariable) for i in node.inputs + node.outputs])

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+            # Oplifter's condition
+            # Will return a list of OP
+            # If None, means can't be moved.
+
+            new_ops = local_gpuaalloc(node, None)
+
+            if new_ops is None:
+                move_to_GPU = False
+
+            if not isinstance(new_ops[0], node.op):
```
nouiz: if not isinstance(new_ops, Op):

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+                move_to_GPU = False
+                continue
```
nouiz: remove the continue

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+
+            # The Extra condition
+            if node.inputs is node.outputs:
+                move_to_GPU = False
+
+            # Oplifter's condition
+            # Will return a list of OP
+            # If None, means can't be moved.
+
+            new_ops = local_gpuaalloc(node, None)
```
nouiz: to get the list of optimizer is in gpu_optimizer

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+                move_to_GPU = False
+                continue
+
+            if move_to_GPU:
+                for old_o, new_o in zip(node.outputs, new_ops):
+                    mapping[old_o] = new_o
+            else:
+                for o in node.outputs:
+                    mapping[o] = o
+
```
nouiz: for o in fgraph.outputs:
    fgraph.replace_validate(o, mapping[0])

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,52 @@ def apply(self, fgraph):
...
+            if not isinstance(new_ops[0], node.op):
+                move_to_GPU = False
+                continue
+
+            if move_to_GPU:
+                for old_o, new_o in zip(node.outputs, new_ops):
+                    mapping[old_o] = new_o
+            else:
+                for o in node.outputs:
+                    mapping[o] = o
```
nouiz: o.op(*[mapping[i] for i in node.inputs])

nouiz: mapping[o] = o.clone_with_new_inputs([mapping[i] for i in node.inputs])

theano/gpuarray/opt.py
Sentient07: @nouiz there are two types of Cuda variable (inside cuda/var.py), CudaNdarrayVariable and CudaNdarraySharedVariable. Should I check with both of them ? Also, why CudaVariable and why not GpuArrayVariable ?

Sentient07: Actually I think we should check for both of them as they both signify that it is already on the GPU ?

theano/gpuarray/opt.py
Sentient07: Shouldn't we check for Variable or list of theano Variable here ?

theano/gpuarray/opt.py
nouiz: You should check for GpuArrayVariable and GpuArraySharedVariable, the Cuda* one are for the old back-end.

theano/gpuarray/opt.py
nouiz: new_ops can be a Variable, a list of Variable or an Op instance that mush be called with the right inputs.

For now, I think you should only support opt that return Op instance.

Sentient07: Op instances in a list, right ?

nouiz: no. We are moving one node at a time. One Op instance, when called with the right inputs will generate one Variable or a list of Variable. It is those that will replace the corresponding outputs of the node we want to move to the GPU.

theano/gpuarray/opt.py
Sentient07: @nouiz I'm having a little confusion here. The inputs to a graph need not necessarily be all the inputs of the apply nodes' in graph, right ?

nouiz: a graph can have inputs that aren't used anywhere.

a node can have inputs that are inputs of the graph and/or outputs of other node.

do that answer your question?

Sentient07: That was the confirmation that I wanted. If that is the case, then in the dict `mapping`, we're creating the dictionary with inputs to the graph, and while  we're cloning the node, we're using the inputs to an apply node as the key. I don't think that would be the same ?
Also, as a workaround, can we do something like this ?
 
`
for n in fgraph.toposort():     
    mapping[n.inputs] = GpuFromHost(None)(n.inputs)`

nouiz: Here in the current code, we don't iter on the nodes on the graph, but only on the inputs of the graph.

The dic map a Theano variable in the old graph to the corresponding Theano variable in the new graph. I think the current code it good.

Also, you pseudo-code would not work as n.inputs is a list and we want the mapping to be per variable, not per node inputs list.

Sentient07: And since an input of a graph should always be an input of an apply node, this won't cause a problem while doing `replace_validate` at the end. Am I correct ?

Sentient07: > Here in the current code, we don't iter on the nodes on the graph, but only on the inputs of the graph.
The dic map a Theano variable in the old graph to the corresponding Theano variable in the new graph. I think the current code it good.

Yeah, I understand this. But, if i initialise the dict with the Variable that are input to the graph, and later iterate this dictionary with Variable that are input to an Apply node, would there be some extra Variable in the latter ?

nouiz: The mapping will contain the mapping for all Theano variable in the graph.
We won't put 2 times the same variable in the mapping (I think this is your question) for a simple reason, we only add them to the mapping when they are produced (so input of the graph and the outputs of apply node).

If we where adding them to the mapping when we would use them, then yes, there would be case where the variable would already be in the graph.

Sentient07: > The mapping will contain the mapping for all Theano variable in the graph.
My confusion is with this part. From my understandings, All the Variable in the graph would be obtained only by navigating through the nodes in the graph, and the ones obtained as inputs to the graph are subset of those obtained by former. Am I correct here ?

nouiz: It depend how you inspect the node.

A node have 2 lists of variables. One for its inputs and one for its outputs (most of the time 1, but there is a few op that have more then 1 output)

If you check all nodes in the graph and check only the inputs variables of the nodes, then yes, you will see all inputs of the graph (except if there are some that aren't used). But you won't see the output variable of the graph.
If you only check the outputs of each nodes, then you will see the outputs variable of the graph, but not the input of the graph.

Here, we want to see all variables. If we start by the input of the graph + the outputs of each nodes, this make sure we see all variables of the graph.

The code in the following for on each node, only add to the mapping the outputs of nodes, not the inputs of each nodes (they should already be there)

Sentient07: Perfect. Yeah. I will add that. Thanks.

theano/gpuarray/opt.py
Sentient07: @nouiz I wanted to ask a question here also. Don't we need to use the `query` method of the EquilibriumDB, which gives us an instance of the EquilibriumOptimizer , so that we can access those (`local_optimizers_all`, `local_optimizers_map`) class variables ?

theano/gpuarray/opt.py
Sentient07: > only do that is if i is of type TensorVariable.

Also, I think it is not correct to add only if they're of only `TensorVariable` type. From the test cases, I see `CudaNdarrayType`, `GpuArrayType` and `TensorConstant` being used.  Is it fine if I make a generic check whether it is of the type `Type` or `Variable` ?

nouiz: for i in fgraph.inputs:
   if isinstance(i.type, TensorType):
               mapping[i] = GpuFromHost(None)(i)
    else:
           mappping[i] = i.clone()

theano/gpuarray/opt.py
nouiz: I don't have a good solution for now. But for now you can use:

theano.gof.optdb.Query(include=['fast_run', 'fast_compile'])

Just make a TODO comment that this should be updated to respect the tag the user provided. I'll think of a better way later.

gcc
```diff
@@ -0,0 +1 @@
+/usr/bin/gcc
```
nouiz: Remove that file from the commited file.

theano/gpuarray/opt.py
```diff
@@ -210,6 +211,77 @@ def apply(self, fgraph):
...
+            # Oplifter's condition
+            # Will return a list of OP
+            # If None, means can't be moved.
+
+            new_ops = []
+            for lopt in (gpu_optimizer.query().local_optimizers_all +
+                         gpu_optimizer.query().local_optimizers_map.get(type(node.op), []) +
+                         gpu_optimizer.query().local_optimizers_map.get(node.op, [])):
+
+                new_ops.append(lopt.transform(node) or lopt(node))
```
nouiz: as soon as one returned something else then None or False, you can break from the loop and use that.

theano/gpuarray/opt.py
```diff
@@ -210,6 +211,77 @@ def apply(self, fgraph):
...
+
+            if all(isinstance(x, Op) for x in new_ops):
```
nouiz: with the change above, this test is not useful anymore.

theano/gpuarray/opt.py
```diff
@@ -210,6 +211,77 @@ def apply(self, fgraph):
...
+                move_to_GPU = False
+
+            if not new_ops:
+                move_to_GPU = False
+
+            for i in node.inputs:
+                if type(i) is not theano.tensor.TensorVariable:
+                    continue
+                newnode = node.clone_with_new_inputs([mapping.get(i) for i in node.inputs])
```
nouiz: The new node should be created only once. Remove the lines 263-265. Line 266 should be in the else clause at on line 273

theano/gpuarray/opt.py
```diff
@@ -210,6 +211,77 @@ def apply(self, fgraph):
...
+
+            if move_to_GPU:
```
nouiz: Here you should create a new node like:

newnode=new_ops(*[mapping.get(i) for i in node.inputs])

Sentient07:
@nouiz I guess the TensorType condition doesn't seem to work. I am getting this error while generating output Variable with `GpuFromHost` Op using the node's output. 
```
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 233, in apply
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 90, in optimize
  File "/Users/Ramana/projects/theano/Theano/theano/gpuarray/opt.py", line 237, in apply
    mapping[o] = GpuFromHost(None)(o)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/op.py", line 602, in __call__
  File "/Users/Ramana/projects/theano/Theano/theano/gpuarray/basic_ops.py", line 415, in make_node
    raise TypeError(x)
TypeError: GpuAlloc<None>.0
```

theano/gpuarray/opt.py
Sentient07: Just realised. I have forgotten to replace other nodes in the graph. I need to traverse the node's input and then make the possible replacements, right ?

nouiz: This loop over output should be after the loop in the nodes in the graph. So here, all should be in the mapping. So I don't understand the problem here.

theano/gpuarray/opt.py
```diff
@@ -210,6 +211,85 @@ def apply(self, fgraph):
...
+        for i in fgraph.inputs:
+            if isinstance(i.type, tensor.TensorType):
+                mapping[i] = GpuFromHost(None)(i)
+            else:
+                mapping[i] = i
+
+        # Iterating through output of all the nodes
+        for n in fgraph.toposort():
+            for o in n.outputs:
+                if isinstance(i.type, tensor.TensorType):
```
nouiz: i.type -> o.type

this will fix the error you pasted

Sentient07: Oh my god! This could probably be the silliest thing that I've forgotten.

Sentient07: And yes! I had pasted the output loop inside the node's iteration. Damn.
Now, there aren't any errors, but I get 11 segmentation faults. What could be the reason ?
```
(macvnev) Ramanas-MacBook-Pro:Theano Ramana$ THEANO_FLAGS='mode=FAST_RUN,device=gpu,lib.cnmem=0.5' theano-nose theano/gpuarray/tests/test_opt.py
Using gpu device 0: GeForce GT 750M (CNMeM is enabled with initial size: 50.0% of memory, cuDNN 4007)
Mapped name None to device cuda: GeForce GT 750M
Using cuDNN version 4007 on context None
.........SS......SSSSS..................
Ran 40 tests in 26.801s

OK (SKIP=7)
Segmentation fault: 11

```

theano/gpuarray/opt.py
Sentient07: I didn't see this comment till now. Sorry, I missed this out somehow. Mapping would have everything, but we're not replacing some nodes in the graphs with the new ones ? We are replacing only the outputs of the fgraph ?

theano/gpuarray/opt.py
nouiz: To clarify the message, you have only 1 segmentation fault. But this correspond to code 11. A segmentation fault always imediatly kill the process where this happen. So you can't have more then 1 per process:)  Check this if you want to know more in detail what this is:

https://en.wikipedia.org/wiki/Segmentation_fault

But in this case, we know there is a problem when we shutdown Theano with the new back-end. As this happen at the shutdown, it is not high in the list of thing to do in it. You can just ignore it for for now. We see that tests passed.

theano/gpuarray/opt.py
nouiz: We build a mapping from all the variables in the old graph to variables in the new graph.

But when we call replace, if we replace one node, it will do the replace recursively.

I'm thinking you should call replace_all_validate only once. So you need a loop over fgraph.outputs to prepare the call to replace_all_validate that you call after the loop. Otherwise, if the graph isn't a tree there could be problems and duplicate computation. This will also be faster as it will call validate only once.

Sentient07: > But when we call replace, if we replace one node, it will do the replace recursively.
 Which means, if i call the last node in the loop, rest of the nodes will also get replaced ?

nouiz: For your first question, yes and no. If you replace a node, it will replace all the previous nodes/variables to that node. We want this to be done for all the graph outputs. Then this will change the full graph. If we do it in separate call, and the the previous nodes from 2 outputs nodes are shared, then there can be problems. If 2 outputs nodes shared previous nodes and we do the change in 1 call to replace_all_validate, then all should be fine.

Sentient07: @nouiz I tried that by applying this logic, using `orderings` method , assuming that it will return all the nodes that occur prior to outputs.

```
        for o in fgraph.outputs:
            current_dependency = set(ordering[o])
            shared_node = current_dependency.interesection(prev_dependency)
            if not shared_node:
                original_node.append(o)
                new_o = mapping[o]
                if new_o.type != o.type:
                    assert isinstance(o.type, tensor.TensorType)
                    assert isinstance(new_o.type, GpuArrayType)
                    new_o = host_from_gpu(new_o)
                rep_node.append(new_o)
            prev_dependency = current_dependency
        for k,l in zip(prev_dependency, current_dependency):
            fgraph.replace_validate(k, l)
```
This is giving me a KeyError.

nouiz: Don't do that. this is too complicated for no good reason and will be slow, as you make many call to validate() (1 per call in replace_validate). Do something like:

```
new_outs = []
for o in fgraph.outputs:
   new_o = mapping[o]
   if new_o.type != o.type:
        assert isinstance(o.type, TensorType)
        assert isinstance(new_o.type, GpuArrayType)
        new_o = host_from_gpu(new_o)
    new_outs.append(new_o)
fgraph.replace_all_validate(zip(fgraph.outputs, new_out))
```

Sentient07: Yeah, i forgot to do the change replace the way you did. But how is not replacing nodes taken care of ?(The case where two output nodes share same previous nodes) Does replace `replace_validate` do that ?

nouiz: After a second tought, it will probably work with doing a loop of replace_validate(). But it will be less  efficient as it will do many call to validate and this is not needed.

Sentient07:
@nouiz I observed something while writing a new test. I was trying to test it from ipython notebook using a sigmoid activation function and binary crossentropy loss function and checking if the transfer to CrossEntropy Op on GPU is happening or not. I have observed it only while using ipython notebook, that when i re-execute the cell that compiles a new theano function(from second time), it gives me a error, 
screenshot at may 31 00-22-50

nouiz:
Did you set device=cudaN?

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,84 @@ def apply(self, fgraph):
...
+            # TODO : the tag should be updated to the one user provides
+            # currently using fast_run and fast_compile tag
+            for lopt in (gpu_optimizer.query(include=['fast_run', 'fast_compile']).local_optimizers_all +
+                         gpu_optimizer.query(include=['fast_run', 'fast_compile']).local_optimizers_map.get(type(node.op), []) +
+                         gpu_optimizer.query(include=['fast_run', 'fast_compile']).local_optimizers_map.get(node.op, [])):
+
+                new_ops = lopt.transform(node) or lopt(node)
+                break
+
+            if not new_ops:
```
nouiz: should be "if not new_ops or not isinstance(new_ops, theano.Op)

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,84 @@ def apply(self, fgraph):
...
+                move_to_GPU = False
+
+            # Oplifter's condition
+            # Will return a list of OP
+            # If None, means can't be moved.
+
+            new_ops = None
+
+            # Selecting the best optimizer
+            # TODO : the tag should be updated to the one user provides
```
nouiz: To solve this, I think we should make a new DB subinstance. I would call it GraphToGPUDB(). It will receive the tags used and query the gpu_optimizer about them. and pass those 2 lists of optimization to the GraphToGPU() instance.

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,84 @@ def apply(self, fgraph):
...
+            new_o = mapping[o]
+            if new_o.type != o.type:
+                assert isinstance(o.type, tensor.TensorType)
+                assert isinstance(new_o.type, GpuArrayType)
+                new_o = host_from_gpu(new_o)
+            new_nodes.append(new_o)
+        fgraph.replace_all_validate(zip(fgraph.outputs, new_nodes))
+
+
+gpu_seqopt.register('GraphToGPU', GraphToGPU(),
```
nouiz: Here pass the GraphToGPUDB instead of GraphToGPU().

theano/gpuarray/opt.py
```diff
@@ -210,6 +210,84 @@ def apply(self, fgraph):
...
+                    -0.5, 'fast_run', 'fast_compile', 'merge')
```
nouiz: remove 'merge'.

theano/gpuarray/opt.py
Sentient07: Wouldn't it be easier and cleaner to make it as an EquilibriumDB instance ? Something like , 

```
class GraphToGPUDB(EquilibriumDB):
    """
    Retrieves the list local optimizers based on the optimizer flag's value 
    from EquilibriumOptimizer by calling the method query.

    """

    def customs_query(self):

        #check the value of the optimizer flag
        include = [o for o in config.optimizer.split(',')]
        return super(GraphToGPUDB, self).query(include=include)
```

Sentient07:
@nouiz the tests seem to pass, but i get this error
```
ERROR (theano.gof.opt): EquilibriumOptimizer max'ed out by 'GraphToGPU'. You can safely raise the current threshold of 5.000000 with the theano flag 'optdb.max_use_ratio'.
```

theano/gpuarray/opt.py
```diff
@@ -283,9 +288,7 @@ def apply(self, fgraph):
             new_nodes.append(new_o)
         fgraph.replace_all_validate(zip(fgraph.outputs, new_nodes))
 
-
-gpu_seqopt.register('GraphToGPU', GraphToGPU(),
-                    -0.5, 'fast_run', 'fast_compile', 'merge')
+graph_optimizer.register("GraphToGPU", GraphToGPU(), 'fast_run', 'fast_compile', 'merge')
```
nouiz: remove that one registration.

theano/gof/optdb.py
```diff
@@ -410,3 +410,25 @@ def __init__(self, db):
...
+    """
+    Retrieves the list local optimizers based on the optimizer flag's value
+    from EquilibriumOptimizer by calling the method query.
+
+    """
+
+    def query(self, *tags, **kwtags):
+        _opts = super(GraphToGPUDB, self).query(*tags, **kwtags)
+        opts = [o for o in _opts]
+        return opt.EquilibriumOptimizer(
```
nouiz: Here It should return a GraphToGPU optimizer. And that class should take
```
def query(...):
    opt = gpu_optimizer.query(...)
   return GraphToGPU(opt.local_optimizer_all, opt.local_optimizer_map)
```

theano/gof/optdb.py
```diff
@@ -410,3 +410,25 @@ def __init__(self, db):
 
     def query(self, *tags, **kwtags):
         return self.db.query(*tags, **kwtags)
+
+
+class GraphToGPUDB(DB):
```
nouiz: That class should be in the file gpuarray/opt.py not here.

PR #4554, opened 2016-05-26
Make recent scan opt change deterministic.

nouiz:
Fix daily buildbot tests error in DebugMode.


abergeron:
Ok when travis passes.

PR #4556, opened 2016-05-27
Don't rebuild inplace add kernels all the time for GpuIncSubtensor.

abergeron:
This also removes the use of ElemwiseKernel which is removed by https://github.com/Theano/libgpuarray/pull/179

### DIFF
--- a/theano/gpuarray/__init__.py
+++ b/theano/gpuarray/__init__.py
@@ -50,8 +50,13 @@ def init_dev(dev, name=None):
     if v[1] < 0:
         raise RuntimeError("Wrong minor API version for gpuarray:", v[1],
                            "Please update libgpuarray/pygpu.")
-    if len(v) < 3 or v[2] < 0:
-        print("Wrong python API version for gpuarray:", v[2],
+    if len(v) < 3:
+        vpy = -1
+    else:
+        vpy = v[2]
+    vpye = 0
+    if vpy < vpye:
+        print("Wrong python API version for gpuarray:", vpy, "expected:", vpye,
               "Some python ops may not work correctly and/or crash. "
               "Consider updating pygpu.", file=sys.stderr)
     global pygpu_activated
--- a/theano/gpuarray/__init__.py
+++ b/theano/gpuarray/__init__.py
@@ -50,6 +50,10 @@ def init_dev(dev, name=None):
     if v[1] < 0:
         raise RuntimeError("Wrong minor API version for gpuarray:", v[1],
                            "Please update libgpuarray/pygpu.")
+    if len(v) < 3 or v[2] < 0:
+        print("Wrong python API version for gpuarray:", v[2],
+              "Some python ops may not work correctly and/or crash. "
+              "Consider updating pygpu.", file=sys.stderr)
     global pygpu_activated
     if dev not in init_dev.devmap:
         ctx = pygpu.init(dev,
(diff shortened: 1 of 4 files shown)


theano/gpuarray/subtensor.py
```diff
@@ -20,6 +20,22 @@
...
 
+iadd_reg = {}
+
+
+def get_iadd(a, b):
+    key = (a.type.dtype, b.type.dtype, a.type.context)
+    if key not in iadd_reg:
+        if a.dtype == 'float16' or b.dtype == 'float16':
+            raise NotImplementedError('float16 is not supported by pygpu '
+                                      'elemwise')
```
nouiz: Is this comment still true? https://github.com/Theano/Theano/issues/1471 tell that Elemwise and Subtensor are supported in float16 in the new back-end.

abergeron: The libgpuarray version doesn't support it (yet).  I could add support though.

nouiz: So this mean that for GpuIncSubtensor, this PR introduce a regression, the perform was supporting float16, but now it isn't.

This also mean that GpuAdvancedIncSubtensor1 wasn't supporting float16, while the issue told it was.

Do you agree with both statement? How long it would be to add this to remove the regression?

abergeron: Yes I agree with both statements.

I estimate about half a day to fix libgpuarray GpuElemwise to support float16.

theano/gpuarray/__init__.py
```diff
@@ -50,6 +50,10 @@ def init_dev(dev, name=None):
     if v[1] < 0:
         raise RuntimeError("Wrong minor API version for gpuarray:", v[1],
                            "Please update libgpuarray/pygpu.")
+    if len(v) < 3 or v[2] < 0:
+        print("Wrong python API version for gpuarray:", v[2],
+              "Some python ops may not work correctly and/or crash. "
+              "Consider updating pygpu.", file=sys.stderr)
```
nouiz: Tell which version you where expecting and which version you got. This is super userful for the user.

abergeron: done.

theano/gpuarray/subtensor.py
nouiz: I merged the PR in libgpuarray. So you can make the float16 in this PR.

abergeron: It will require this too: https://github.com/Theano/libgpuarray/pull/191

abergeron:
I've added support for float16 in the inplace add using https://github.com/Theano/libgpuarray/pull/191.

theano/gpuarray/subtensor.py
```diff
@@ -469,6 +466,7 @@ class GpuAdvancedIncSubtensor1(Op):
     Implement AdvancedIncSubtensor1 on the gpu.
 
     """
+    _f16_ok = True
```
nouiz: Add some basic test that it work. I think the C code don't support it. It don't pass the new flag. If this is not needed, can you add a comment when creating GpuElemwise_new() to tell this?

abergeron:
Now with more test™.

nouiz:
travis fail:
/home/travis/build/Theano/Theano/theano/gpuarray/tests/test_subtensor.py:60:1: E302 expected 2 blank lines, found 1

nouiz:
Otherwise, the test pass, we can merge when the flake8 is fixed and travis pass.

abergeron:
Done.

PR #4565, opened 2016-05-31
Fix a dependency version to fix travis error

nouiz:


lamblin:
That makes sense.

nouiz:
The doc part of travis passed, so merging.

PR #4566, opened 2016-05-31
In get_scalar_constant_value, do not return internal memory of constant

### LINKED ISSUE #4560: Incorrect pool shape
There seems to be a problem with calculating the shape in theano.tensor.signal.pool.pool_2d. The result of the symbolic shape ((5, 5, 2, 2) below) isn't the same as either the actual shape of the output or the result of `Pool.out_shape` (both (1, 1, 1, 1)).

```
import theano
import theano.tensor as T
from theano.tensor.signal.pool import pool_2d
fX = theano.config.floatX
kwargs = dict(
    ds=(2, 3),
    st=(2, 2),
    ignore_border=True,
    padding=(1, 1),
)
pooled = pool_2d(T.constant(np.random.randn(1, 1, 1, 1).astype(fX)), **kwargs)
print pooled.shape.eval() # [5, 5, 2, 2]
print pooled.eval().shape # (1, 1, 1, 1)

from theano.tensor.signal.pool import Pool
print Pool.out_shape((1, 1, 1, 1), **kwargs) # [1, 1, 1, 1]
```

lamblin:
Fix #4560.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -612,14 +612,14 @@ def get_scalar_constant_value(orig_v, elemwise=True,
             return numpy.asarray(v)
 
         if isinstance(v, numpy.ndarray):
-            return numpy_scalar(v)
+            return numpy_scalar(v).copy()
 
         if isinstance(v, Constant):
             if getattr(v.tag, 'unique_value', None) is not None:
                 data = v.tag.unique_value
             else:
                 data = v.data
-            return numpy_scalar(data)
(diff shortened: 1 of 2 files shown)


lamblin:
This was an issue when calling inplace operations, like `+=`, on the value returned by `extract_constant`, for instance in [`Pool.out_shape`](https://github.com/Theano/Theano/blob/baf5227a93b7ff1d8b8aa59010591f86ecab0d43/theano/tensor/signal/pool.py#L189-L194).

PR #4582, opened 2016-06-03
cuDNN v5 Batch Normalization

f0k:
Follow-up of #4237. Took the patch by @lachlants (https://github.com/Theano/Theano/pull/4237#issuecomment-223466380) and cleaned up / rewrote most of the code so I could understand what it was doing. I guess it was originally derived from the softmax Ops, using `if`s in the base class were inheritance would have made it much easier (so I changed it accordingly).

As per @lachlants idea, this does not cope with updating the running average statistics in the training forward pass, but just hands out the current mini-batch statistics so the user can do whatever she or he pleases. This makes it much easier for the initial version.

Still misses tests, but I've written a small test script to verify that it does the correct thing for all three Ops (will still add tests, of course). Also misses a convenience function. I'd prefer to leave inplace optimizations for future PRs, as well as extending it to support 5d tensors.

TODO:
* [x] add tests
* [x] add convenience function for inference and training (support 1d to 3d tensors as well)
* [x] add grad() method to inference Op

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -2639,6 +2639,138 @@ err%(name)s = cudnnBatchNormalizationBackward(
         return result
 
 
+def dnn_batch_normalization_train(inputs, gamma, beta, mode='per-activation',
+                                  epsilon=1e-4):
+    """
+    Performs batch normalization of the given inputs, using the mean and
+    variance of the inputs.
+
+    Parameters
+    ----------
+    mode : {'per-activation', 'spatial'}
+        Whether to normalize per activation or share normalization factors
+        across spatial dimensions (i.e., all dimensions past the second).
+    gamma : tensor
+        Learnable scale factors. Must match the dimensionality of `inputs`,
+        but have sizes of `1` for all axes normalized over (i.e., in the first
+        dimension for ``mode='per-activation'`, and additionally in all
+        dimensions past the second for ``mode='spatial'``).
+    beta : tensor
+        Learnable biases. Must match the tensor layout of `gamma`.
+    epsilon : float
+        Epsilon value used in the batch normalization formula. Minimum allowed
+        value is 1e-5 (imposed by cuDNN).
+
+    Returns
+    -------
+    out : tensor
+        Batch-normalized inputs.
+    mean : tensor
+        Means of `inputs` across the normalization axes.
+    stdinv : tensor
+        Inverse standard deviations of `inputs` across the normalization axes.
+
+    Notes
+    -----
+    For 4d tensors, returned values are equivalent to:
+
+    >>> axes = 0 if mode == 'per-activation' else (0, 2, 3)
+    >>> mean = inputs.mean(axes, keepdims=True)
+    >>> stdinv = T.inv(T.sqrt(inputs.var(axes, keepdims=True) + epsilon))
+    >>> out = (inputs - mean) * gamma * stdinv + beta
+    """
+    ndim = inputs.ndim
+    if ndim > 4:
+        raise ValueError("dnn_batch_normalization_train currently supports "
+                         "up to 4-dimensional tensors only, got %d" % ndim)
+    if gamma.ndim != ndim or beta.ndim != ndim:
+        raise ValueError("gamma and beta must be of the same dimensionality "
+                         "as inputs; got %d and %d instead of %d" %
+                         (gamma.ndim, beta.ndim, ndim))
+    if epsilon < 1e-5:
+        raise ValueError("epsilon must be at least 1e-5, got %f" % epsilon)
+
+    if ndim < 4:
+        inputs = theano.tensor.shape_padright(inputs, 4 - ndim)
+        gamma = theano.tensor.shape_padright(gamma, 4 - ndim)
+        beta = theano.tensor.shape_padright(beta, 4 - ndim)
+    batchnorm_op = GpuDnnBatchNorm(mode=mode, epsilon=epsilon)
+    result = tuple(batchnorm_op(inputs, gamma, beta))
+    if ndim < 4:
+        result = tuple(theano.tensor.flatten(r, ndim) for r in result)
+    return result
+
+
+def dnn_batch_normalization_test(inputs, gamma, beta, mean, var,
+                                 mode='per-activation', epsilon=1e-4):
+    """
+    Performs batch normalization of the given inputs, using the given mean and
+    variance.
+
+    Parameters
+    ----------
+    mode : {'per-activation', 'spatial'}
+        Whether to normalize per activation or share normalization factors
+        across spatial dimensions (i.e., all dimensions past the second).
+    gamma : tensor
+        Scale factors. Must match the dimensionality of `inputs`, but have
+        sizes of `1` for all axes normalized over (i.e., in the first dimension
+        for ``mode='per-activation'`, and additionally in all dimensions past
+        the second for ``mode='spatial'``).
+    beta : tensor
+        Biases. Must match the tensor layout of `gamma`.
+    mean : tensor
+        Means. Usually these are running averages computed during training.
+        Must match the tensor layout of `gamma`.
+    var : tensor
+        Variances. Usually these are running averages computed during training.
+        Must match the tensor layout of `gamma`.
+    epsilon : float
+        Epsilon value used in the batch normalization formula. Minimum allowed
+        value is 1e-5 (imposed by cuDNN).
+
+    Returns
+    -------
+    out : tensor
+        Batch-normalized inputs.
+
+    Notes
+    -----
+    For 4d tensors, the returned value is equivalent to:
+
+    >>> axes = (0,) if mode == 'per-activation' else (0, 2, 3)
+    >>> gamma, beta, mean, var = (T.addbroadcast(t, *axes)
+    ...                           for t in (gamma, beta, mean, var))
+    >>> out = (inputs - mean) * gamma / T.sqrt(var + epsilon) + beta
+    """
+    ndim = inputs.ndim
+    if ndim > 4:
+        raise ValueError("dnn_batch_normalization_test currently supports "
+                         "up to 4-dimensional tensors only, got %d" % ndim)
+    if gamma.ndim != ndim or beta.ndim != ndim:
+        raise ValueError("gamma and beta must be of the same dimensionality "
+                         "as inputs; got %d and %d instead of %d" %
+                         (gamma.ndim, beta.ndim, ndim))
+    if mean.ndim != ndim or var.ndim != ndim:
+        raise ValueError("mean and var must be of the same dimensionality "
+                         "as inputs; got %d and %d instead of %d" %
+                         (mean.ndim, var.ndim, ndim))
+    if epsilon < 1e-5:
+        raise ValueError("epsilon must be at least 1e-5, got %f" % epsilon)
+
+    if ndim < 4:
+        inputs = theano.tensor.shape_padright(inputs, 4 - ndim)
+        gamma = theano.tensor.shape_padright(gamma, 4 - ndim)
+        beta = theano.tensor.shape_padright(beta, 4 - ndim)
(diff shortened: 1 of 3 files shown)


nouiz:
Thanks, no time to review today. Ok to let the inplace opt and 5d tensor
support for a later PR.

nouiz:
@ballasn will review this.

f0k:
Damn it, the cuDNN v5 API is still not what you'd expect.

* `cudnnBatchNormalizationForwardInference` expects `estimatedMean` and `estimatedVariance`, where v4 expected the mean and inverse standard deviation (= sqrt(variance + epsilon)).
* According to the documentation PDF, `cudnnBatchNormalizationForwardTraining` writes `resultRunningMean`, `resultRunningInvVariance`, `resultSaveMean`, `resultSaveVariance`. According to the header, it writes `resultRunningMean`, `resultRunningVariance`, `resultSaveMean`, `resultSaveInvVariance`. And the header seems to be correct (somewhat): It returns the mean and variance in the first parameters, and the mean and inverse standard deviation in the latter parameters.
* According to the documentation PDF, `cudnnBatchNormalizationBackward` can reuse `savedMean` and `savedInvVariance`, and this matches the header and the implementation (except that `invVariance` means the inverse standard deviation).

So in order to reuse the forward pass results for the backward pass, we need the forward Op to return the mean and the inverse standard deviation of the current mini-batch (and not the variance). However, the inference implementation needs the variance, not the inverse standard deviation (this was different in v4). So the API is basically forcing us to make use of `resultRunningVariance` or to invert and square the inverse standard deviation again before calling the inference Op. And if we make use of `resultRunningVariance`, we have to make use of `resultRunningMean` as well because it's not allowed to set only one of them to a non-NULL value.

So in the end it seems the v4 API was much more sensible. Maybe in Lasagne we'll just not use the inference Op and work with inverse standard deviations throughout, that's what the current BatchNormLayer does anyway.

How shall we design the training forward Op then? We could make it useable in two ways:
* return y,  x_mean and x_invstd
* return y, x_mean, x_invstd and in-place update running_mean, running_var

As you don't want the user to introduce in-place Ops, the second way probably shouldn't be in-place by default, but it would still be one that takes two additional tensors (the old running_mean and running_var).

f0k:
All TODOs done.
@ballasn, could you have a look when you've got the time?
@nouiz, do you want to cancel the Travis build for "Add GpuDnnBatchNorm tests" (it hasn't started yet) so it can run the test for the latest commit instead?

f0k:
Some benchmark numbers for a GT 640, vanilla Theano vs cuDNN:
* training, forward pass only: 0.55 vs 0.28 (seconds for 100 calls for a (128, 20, 30, 40) tensor)
* training, forward and backward pass: 1.67 vs 0.57
* inference, forward pass only: 0.22 to 0.16

For a GTX 970:
* training, forward pass only: 0.12 vs 0.03
* training, forward and backward pass: 0.34 vs 0.07
* inference, forward pass only: 0.03 to 0.01

Can anyone trigger Travis for the last commit or shall we close/reopen the issue?

theano/sandbox/cuda/dnn.py
```diff
@@ -2270,6 +2270,375 @@ def method(self):
...
+        est_var_eps = est_var + self.epsilon
+        est_std = theano.tensor.sqrt(est_var_eps)
+        two = theano.tensor.constant(2.)
+
+        # define and return gradients
+        dx = dy * (scale / est_std)
+        dscale = (dy * (x - est_mean)).sum(axes, keepdims=True) / est_std
+        dbias = dy.sum(axes, keepdims=True)
+        dmean = -dy.sum(axes, keepdims=True) * (scale / est_std)
+        dvar = -(dy * (x - est_mean)).sum(axes, keepdims=True) * (scale / (two * est_var_eps * est_std))
```
ballasn: Why don't we use GpuDnnBatchNormGrad here as well?

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -719,6 +719,111 @@ def test_log_softmax2(self):
...
+        grads = T.grad(None, wrt=[x, scale, bias], known_grads={out: dy})
+        # reference backward pass
+        grads2 = T.grad(None, wrt=[x, scale, bias], known_grads={out2: dy})
+        # compile
+        f = theano.function([x, scale, bias, dy],
+                            [out, x_mean, x_invstd, out2, x_mean2, x_invstd2] +
+                            grads + grads2, mode=mode_with_gpu)
+        # run
+        floatX = theano.config.floatX
+        for data_shape in ((10, 20, 30, 40), (4, 3, 2, 1)):
```
ballasn: we might want to try more shape here,

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -719,6 +719,111 @@ def test_log_softmax2(self):
...
+            param_shape = tuple(1 if d in axes else s
+                                for d, s in enumerate(data_shape))
+            X = 4 + 3 * numpy.random.randn(*data_shape).astype(floatX)
+            Dy = -1 + 2 * numpy.random.randn(*data_shape).astype(floatX)
```
ballasn: why is there a 4 + 3 and -1 * line 760/761 ?

ballasn:
Travis seems to indicates some flake8 errors, and we might want to add more test cases, but, overall it is looking good!

theano/sandbox/cuda/dnn.py
f0k: Because GpuDnnBatchNormGrad assumes the mean and variance are x.mean() and x.var(). It cannot separate out the gradients wrt. x, mean, and variance.

theano/sandbox/cuda/tests/test_dnn.py
f0k: Yes, they're easy to add the way I wrote it -- any suggestions?

theano/sandbox/cuda/tests/test_dnn.py
f0k: Because randn() gives a distribution of zero mean and unit variance, and then batch normalization wouldn't have to change anything. More concretely, it wouldn't allow us to distinguish whether the cuDNN Ops compute variance, standard deviation or the inverses. So I'm adding a bias and coefficient to get a non-zero mean and a non-unit variance.

f0k:
> Travis seems to indicates some flake8 errors

My bad. I only did a pep8 check locally.

> and we might want to add more test cases

Any concrete suggestions? I'd like to do this in one go while fixing the flake8 errors so we don't need two Travis runs.

theano/sandbox/cuda/dnn.py
ballasn: I see, maybe we should just not define a gradient for the inference op, @nouiz what do you think about that?

theano/sandbox/cuda/tests/test_dnn.py
ballasn: I would try  a square shape, shape with dimension (b, c, 1, 1), shape with uneven number of spatial dimensions,

I would also try some random mean variance in the inputs

f0k: > I would try a square shape, shape with dimension (b, c, 1, 1)

I'll change `(4, 3, 2, 1)` to `(4, 3, 1, 1)` and add another shape that is square.

> shape with uneven number of spatial dimensions

This only supports 4D tensors, so we always have 2 spatial dimensions. The convenience function supports 1D, 2D and 3D tensors via shape padding.

theano/sandbox/cuda/dnn.py
f0k: > I see, maybe we should just not define a gradient for the inference op

It's required for saliency mapping by backpropagation, artistic style transfer, things like that. This will use the inference Op for the forward pass and still need the gradient to be defined.

ballasn: I see, ok then for the gradient in the inference op.

theano/sandbox/cuda/tests/test_dnn.py
ballasn: shape with uneven number of spatial dimensions
This only supports 4D tensors, so we always have 2 spatial dimensions. The convenience function supports 1D, 2D and 3D tensors via shape padding.

I meant a square with dimension size is even, one that is  uneven.  But three shape might be enough. However, we should also probably test the convenience function directly, and not only the op.

f0k:
Okay, fixed flake8 errors, changed the test shapes and added tests for the convenience functions. They ended up looking almost identical to the direct Op tests, so I removed the direct Op tests again -- they're not really needed any longer (for tensor4 inputs, the convenience functions directly call the Op anyway, for lower-dimensional inputs they do some shape padding and flattening that's worth being tested).

Good to merge from my side if Travis is happy.

@nouiz or any other admin, can you cancel https://travis-ci.org/Theano/Theano/builds/138151389? It's obsolete.

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -715,6 +715,119 @@ def test_log_softmax2(self):
...
+                Bias = numpy.random.randn(*param_shape).astype(floatX)
+                outputs = f(X, Scale, Bias, Dy)
+                # compare outputs
+                utt.assert_allclose(outputs[0], outputs[0 + 3])  # out
+                utt.assert_allclose(outputs[1], outputs[1 + 3])  # mean
+                utt.assert_allclose(outputs[2], outputs[2 + 3])  # invstd
+                # compare gradients
+                utt.assert_allclose(outputs[6], outputs[6 + 3])  # dx
+                utt.assert_allclose(outputs[7], outputs[7 + 3], rtol=3e-3)  # dscale
+                utt.assert_allclose(outputs[8], outputs[8 + 3])  # dbias
```
f0k: This high `rtol` is actually a bit troubling. Everything else is very close to what we get with vanilla Theano, but the gradient wrt. scale is a bit off. For example, we get (cuDNN vs reference):
```
[[[[ 504.36730957]]    [[[[ 504.3671875 ]]
  [[-286.85342407]]      [[-286.85232544]]
  [[-163.40234375]]      [[-163.40229797]]
  [[  21.92313004]]      [[  21.92422295]]
  [[ 226.3973999 ]]      [[ 226.39941406]]
  [[  -0.51444209]]      [[  -0.51332498]]
  [[  42.07894897]]      [[  42.08079529]]
  [[  26.67364693]]      [[  26.6755085 ]]
  [[ 154.5411377 ]]      [[ 154.54122925]]
  [[  45.03813934]]      [[  45.03807831]]
  [[  84.71562958]]      [[  84.71569061]]
  [[ -62.2019043 ]]      [[ -62.20304108]]
  [[ 259.81713867]]      [[ 259.81518555]]
  [[  22.97838593]]      [[  22.97837639]]
  [[ 141.40872192]]      [[ 141.41171265]]
  [[ 191.33883667]]      [[ 191.33883667]]
  [[  22.5411129 ]]      [[  22.5394516 ]]
  [[ 211.42269897]]      [[ 211.42268372]]
  [[-179.14613342]]      [[-179.14613342]]
  [[ 103.85586548]]]]    [[ 103.85778809]]]]
```
Some values are on spot (the second to last), some are far off in relative terms (-0.513 vs -0.514). Not sure if this is just different rounding or something wrong with cuDNN. In any case, it's probably close enough for learning.

nouiz: for the rtol, can you try different cudnn algo? The problems could be that
some aren't deterministic.

theano/sandbox/cuda/dnn.py
nouiz: I don't get why for those case they can't take the training implementation
of cudnn bn?

theano/sandbox/cuda/tests/test_dnn.py
f0k: > for the rtol, can you try different cudnn algo?

There is just one for batch normalization.

f0k:
Anything left to do? Travis is happy, I'm happy, @ballasn is happy (I assume because I addressed all comments).

PR #4584, opened 2016-06-03
Add some documentation on how to write gpu ops.

abergeron:
This explains some of the specifics about making gpu ops and also introduces a class that merges COp and GpuKernelBase if you want to use both.

I've explained all that I could think could be a problem when building new GPU ops, but if you think of something else, add a comment and I'll see what I can do about it.

### DIFF
--- a/doc/extending/extending_theano_gpu.txt
+++ b/doc/extending/extending_theano_gpu.txt
@@ -72,7 +72,10 @@ The last place where you might need the context is in the C
 initialization code.  For that you will have to use the :ref:`params
 <extending_op_params>`.  The params type should be
 :class:`theano.gpuarray.type.gpu_context_type` and the params object
-should be a context object from one of your input variables.
+should be a context object from one of your input variables::
+
+    def get_params(self, node):
+        return node.inputs[0].type.context
 
 If you don't have any input variables on the GPU you can follow the
 the example of :class:`theano.gpuarray.basic_ops.GpuFromHost` or
--- a/doc/extending/extending_theano_gpu.txt
+++ b/doc/extending/extending_theano_gpu.txt
@@ -78,3 +78,57 @@ If you don't have any input variables on the GPU you can follow the
 the example of :class:`theano.gpuarray.basic_ops.GpuFromHost` or
 :class:`theano.gpuarray.basic_ops.GpuEye`.  This is not a case that
 you should encounter often, so it will not be covered further.
+
+Defining new kernels
+====================
+
+If your op needs to do some transformation on the data, chances are
+that you will need to write a new kernel.  The best way to do this is
+to leverage GpuKernelBase (or CGpuKernelBase if you want to use the
+COp functionality).
+
+For plain GpuKernelBase, you have to define a method called
+gpu_kernels which returns a list of :class:`Kernel
+<theano.gpuarray.basic_ops.Kernel>` objects.  You can define as many
+kernels as you want for a single op.  An example would look like this:
+
+    def gpu_kernels(self, node, name):
+        code = """
+KERNEL void k(GLOBAL_MEM ga_float *a, ga_size n, ga_size m) {
+    ga_size nb = n < m ? n : m;
+    for (ga_size i = LID_0; i < nb; i += LDIM_0) {
+        a[i*m + i] = 1;
+    }
+}"""
+        return [Kernel(
+                code=code, name="k",
+                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE],
+                flags=Kernel.get_flags(self.dtype))]
+
+If you want to use COp, then you should use `CGpuKernelBase` instead.
+It add a new section to the parsed files whose tag is `kernels`.
+Inside that section you can define some kernels with `#kernel
+name:params:flags`.
+
+Here `name` is the name of the kernel function in the following code,
+`params` is a comma-separeted list of C typecode names and `flags` is
+a `|`-separeted list of C kernel flag values (can be empty).  The same kernel definition as above would look like this with `CGpuKernelBase`:
+
+    #section kernels
+
+    #kernel k : GA_BUFFER, GA_SIZE, GA_SIZE : GA_USE_CLUDA
+
+    KERNEL void k(GLOBAL_MEM ga_float *a, ga_size n, ga_size m) {
+        ga_size nb = n < m ? n : m;
+        for (ga_size i = LID_0; i < nb; i += LDIM_0) {
+        a[i*m + i] = 1;
+        }
+    }
+
+The second method is to handle the kernel compilation and cache on
+your own.  This is not recommended because there are lots of details
+to pay attention to that can cripple your performance if not done
+right, which GpuKernelBase handles for you.
+
+In any case you will need to call your compiled kernel with some data.
+This is done using the `GpuKernel_call()` method in your C code.
--- a/doc/extending/extending_theano_gpu.txt
+++ b/doc/extending/extending_theano_gpu.txt
@@ -113,14 +113,19 @@ instead.  It adds a new section to the parsed files whose tag is
 ``kernels``.  Inside that section you can define some kernels with
 ``#kernel name:params:flags``.
 
-Here ``name`` is the name of the kernel function in the following code,
-``params`` is a comma-separeted list of C typecode names and ``flags`` is
-a ``|``-separeted list of C kernel flag values (can be empty).  The same
-kernel definition as above would look like this with ``CGpuKernelBase``::
+Here ``name`` is the name of the kernel function in the following
+code, ``params`` is a comma-separeted list of numpy typecode names.
+There are three exceptions for ``size_t`` which should be noted as
+``size``, ``ssize_t`` which should be noted as ``ssize`` and a pointer
+which should be noted as ``*``.
+
+``flags`` is a ``|``-separeted list of C kernel flag values (can be
+empty).  The same kernel definition as above would look like this with
+``CGpuKernelBase``::
 
     #section kernels
 
-    #kernel k : GA_BUFFER, GA_SIZE, GA_SIZE : GA_USE_CLUDA|GA_USE_DOUBLE
+    #kernel k : *, size, size : GA_USE_DOUBLE
 
     KERNEL void k(GLOBAL_MEM ga_double *a, ga_size n, ga_size m) {
         ga_size nb = n < m ? n : m;
--- a/doc/extending/extending_theano_gpu.txt
+++ b/doc/extending/extending_theano_gpu.txt
@@ -38,13 +38,13 @@ written.  An example usage is below::
 
 In this example the Op takes three inputs, all on the GPU.  In case
 one or more of your inputs is not supposed to be on the GPU, you
-should not pass it to `infer_context_name`.
+should not pass it to :func:`infer_context_name`.
 
 Also note that :func:`theano.gpuarray.basic_ops.as_gpuarray_variable`
-takes `context_name` as a mandatory parameter.  This is because it's
+takes ``context_name`` as a mandatory parameter.  This is because it's
 not enough to know you want the value to be on the GPU, you also want
 to know which GPU to put it on.  In almost all cases, you can pass in
-the return value of `infer_context_name` there.
+the return value of :func:`infer_context_name` there.
 
 If you also need the context during runtime (for example to allocate
 the output).  You can use the context of one of your inputs to know
@@ -90,35 +90,36 @@ COp functionality).
 For plain GpuKernelBase, you have to define a method called
 gpu_kernels which returns a list of :class:`Kernel
 <theano.gpuarray.basic_ops.Kernel>` objects.  You can define as many
-kernels as you want for a single op.  An example would look like this:
+kernels as you want for a single op.  An example would look like this::
 
     def gpu_kernels(self, node, name):
         code = """
-KERNEL void k(GLOBAL_MEM ga_float *a, ga_size n, ga_size m) {
-    ga_size nb = n < m ? n : m;
-    for (ga_size i = LID_0; i < nb; i += LDIM_0) {
-        a[i*m + i] = 1;
-    }
-}"""
+    KERNEL void k(GLOBAL_MEM ga_double *a, ga_size n, ga_size m) {
+        ga_size nb = n < m ? n : m;
+        for (ga_size i = LID_0; i < nb; i += LDIM_0) {
+            a[i*m + i] = 1;
+        }
+    }"""
         return [Kernel(
                 code=code, name="k",
                 params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE],
-                flags=Kernel.get_flags(self.dtype))]
+                flags=Kernel.get_flags('float64'))]
 
-If you want to use COp, then you should use `CGpuKernelBase` instead.
-It add a new section to the parsed files whose tag is `kernels`.
-Inside that section you can define some kernels with `#kernel
-name:params:flags`.
+If you want to use COp, then you should use ``CGpuKernelBase``
+instead.  It adds a new section to the parsed files whose tag is
+``kernels``.  Inside that section you can define some kernels with
+``#kernel name:params:flags``.
 
-Here `name` is the name of the kernel function in the following code,
-`params` is a comma-separeted list of C typecode names and `flags` is
-a `|`-separeted list of C kernel flag values (can be empty).  The same kernel definition as above would look like this with `CGpuKernelBase`:
+Here ``name`` is the name of the kernel function in the following code,
+``params`` is a comma-separeted list of C typecode names and ``flags`` is
+a ``|``-separeted list of C kernel flag values (can be empty).  The same
+kernel definition as above would look like this with ``CGpuKernelBase``::
 
     #section kernels
 
-    #kernel k : GA_BUFFER, GA_SIZE, GA_SIZE : GA_USE_CLUDA
+    #kernel k : GA_BUFFER, GA_SIZE, GA_SIZE : GA_USE_CLUDA|GA_USE_DOUBLE
 
-    KERNEL void k(GLOBAL_MEM ga_float *a, ga_size n, ga_size m) {
+    KERNEL void k(GLOBAL_MEM ga_double *a, ga_size n, ga_size m) {
         ga_size nb = n < m ? n : m;
         for (ga_size i = LID_0; i < nb; i += LDIM_0) {
         a[i*m + i] = 1;
@@ -128,7 +129,25 @@ a `|`-separeted list of C kernel flag values (can be empty).  The same kernel de
 The second method is to handle the kernel compilation and cache on
 your own.  This is not recommended because there are lots of details
 to pay attention to that can cripple your performance if not done
-right, which GpuKernelBase handles for you.
+right, which GpuKernelBase handles for you.  But if you really want to
+go this way, then you can look up the C API for kernels in
+libgpuarray.
 
 In any case you will need to call your compiled kernel with some data.
-This is done using the `GpuKernel_call()` method in your C code.
+This is done using the ``GpuKernel_call()`` method in your C code.
+An example calling the above kernel would be::
+
+    size_t ls, gs;
+    size_t dims[2];
+    void *args[3];
+
+    // ...
+
+    args[0] = input->ga.data;
+    args[1] = &dims[0];
+    args[2] = &dims[1];
+    ls = 1;
+    gs = 256;
+    err = GpuKernel_call(&%(kname)s, 1, &ls, &gs, 0, args);
+
+    // ...
--- a/doc/extending/extending_theano_gpu.txt
+++ b/doc/extending/extending_theano_gpu.txt
@@ -1,0 +1,80 @@
+
+.. _extending_theano_gpu:
+
+==============================
+Extending Theano with a GPU Op
+==============================
+
+This tutorial covers how to extend Theano with an op that offers a GPU
+implementation.  It assumes you are familiar with how to write new
+Theano ops.  If that is not the case you should probably follow the
+:ref:`extending_theano` and :ref:`extending_theano_c` section before
+continuing on.
+
+Writing a new GPU op can be done in python for some simple tasks, but
+will usually done in C to access the complete API and avoid paying the
+overhead of a python function call.
+
+Dealing with the context
+========================
+
+One of the major differences with GPU ops is that they require a
+context (a.k.a. device) to execute.  Most of the time you can infer
+the context to run on from your inputs.  There is a way for the user
+to transfer things between contexts and to tag certain varaibles for
+transfer.  It might also be the case that your inputs are not all from
+the same context and you would have to choose which one to run on.
+
+In order to support all of those options and have a consistent
+interface, :func:`theano.gpuarray.basic_ops.infer_context_name` was
+written.  An example usage is below::
+
+    def make_node(self, a, b, c):
+        ctx = infer_context_name(a, b, c)
+        a = as_gpuarray_variable(a, ctx)
+        b = as_gpuarray_variable(b, ctx)
+        c = as_gpuarray_variable(c, ctx)
+        return Apply(self, [a, b, c], [a.type()])
+
+In this example the Op takes three inputs, all on the GPU.  In case
+one or more of your inputs is not supposed to be on the GPU, you
+should not pass it to `infer_context_name`.
+
+Also note that :func:`theano.gpuarray.basic_ops.as_gpuarray_variable`
+takes `context_name` as a mandatory parameter.  This is because it's
+not enough to know you want the value to be on the GPU, you also want
+to know which GPU to put it on.  In almost all cases, you can pass in
+the return value of `infer_context_name` there.
+
+If you also need the context during runtime (for example to allocate
+the output).  You can use the context of one of your inputs to know
+which one to use.  Here is another example::
+
+    def perform(self, node, inputs, output_storage):
+        A, B = inputs
+        C, = output_storage
+        C[0] = pygpu.empty([A.shape[0], B.shape[1]], dtype=A.dtype, A.context)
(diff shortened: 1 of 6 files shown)


lamblin:
I think it would be worth explaining how to call an function already wrapped in Python through pycuda or scikits.cuda, for instance, and maybe provide an example.

theano/gpuarray/basic_ops.py
```diff
@@ -313,6 +362,90 @@ def kernel_version(self, node):
...
+            split = self.kernel_re.split(code)
+            if split[0].strip() != '':
+                raise ValueError("Stray code in kernels section before the "
+                                 "first #kernel statement.")
+            def_macros, undef_macros = self._type_macros(node)
+            n = 1
+            res = []
+            while n < len(split):
+                kspec = split[n]
+                kcode = split[n+1]
```
lamblin: /home/travis/build/Theano/Theano/theano/gpuarray/basic_ops.py:431:32: E226 missing whitespace around arithmetic operator

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,80 @@
...
+
+Dealing with the context
+========================
+
+One of the major differences with GPU ops is that they require a
+context (a.k.a. device) to execute.  Most of the time you can infer
+the context to run on from your inputs.  There is a way for the user
+to transfer things between contexts and to tag certain varaibles for
+transfer.  It might also be the case that your inputs are not all from
+the same context and you would have to choose which one to run on.
```
nouiz: Tell that the most compatible way with how Theano behave is to use the context from inputs[0].

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,80 @@
...
+Note that GpuArrayType objects also have a `context_name` attribute
+which is the symbolic equivalent of `context`.  It can't be used for
+calls to pygpu or libgpuarray, but it should be used for theano
+operations and variables.
+
+The last place where you might need the context is in the C
+initialization code.  For that you will have to use the :ref:`params
+<extending_op_params>`.  The params type should be
+:class:`theano.gpuarray.type.gpu_context_type` and the params object
+should be a context object from one of your input variables.
```
nouiz: Add the corresponding code snippet.

doc/extending/extending_theano_gpu.txt
nouiz: I see that you continue on that bellow. I think we should document somewhere how we infer the context by default. But I'm not sure where. I think it should be in the use of the new back-end and we should link to it from here.

theano/gpuarray/basic_ops.py
```diff
@@ -125,6 +125,48 @@ class Kernel(object):
...
+    ----------
+    code: str
+        The source code of the kernel.
+    params: list
+        list of parameter types.
+    name: str
+        the name of the kernel function in the source.
+    flags: dict
+        dictionary of flags
+    codevar: str
```
nouiz: Tell what the default is for this one and the 2 following.

Sentient07:
There was a part in this documentation which was repeating. I remember coming across it when i was in the middle of a rebase and couldn't make the changes. 
@abergeron can i push that change to your branch ? 
cc : @nouiz

theano/gpuarray/basic_ops.py
```diff
@@ -167,6 +168,8 @@ def get_dtype(t):
 
     def _get_c_flags(self):
         res = []
+        if self.flags.get('cflags', '') != '':
```
nouiz: Why `!= ''`? I would remove that. So it they pass None, it would work.

theano/gpuarray/basic_ops.py
```diff
@@ -313,6 +320,90 @@ def kernel_version(self, node):
         return (4, self.get_params(node).bin_id)
 
 
+def fwds(name):
```
nouiz: Use a better name. I don't guess what it mean.

nouiz:
Which part?

On Tue, Jun 7, 2016 at 9:25 AM, Ramana Subramanyam <notifications@github.com

doc/extending/extending_theano_gpu.txt
```diff
@@ -78,3 +78,57 @@ If you don't have any input variables on the GPU you can follow the
...
+    for (ga_size i = LID_0; i < nb; i += LDIM_0) {
+        a[i*m + i] = 1;
+    }
+}"""
+        return [Kernel(
+                code=code, name="k",
+                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE],
+                flags=Kernel.get_flags(self.dtype))]
+
+If you want to use COp, then you should use `CGpuKernelBase` instead.
```
nouiz: Make a full example with CGpuKernelBase and put this as a test and load it in the doc. To make sure the doc is up to date and that we don't break it.

theano/gpuarray/basic_ops.py
```diff
@@ -313,6 +320,90 @@ def kernel_version(self, node):
...
+
+    kernel_re = re.compile(r'^#kernel ([a-zA-Z_].*?)$', re.MULTILINE)
+
+    c_support_code = fwds('c_support_code')
+    c_support_code_apply = fwds('c_support_code_apply')
+    c_support_code_struct = fwds('c_support_code_struct')
+    c_init_code_struct = fwds('c_init_code_struct')
+    c_cleanup_code_struct = fwds('c_cleanup_code_struct')
+
+    def _type_macros(self, node):
```
nouiz: I don't like that. This push in the old direction on compiling code for different dtypes. Do you have an idea how to allow the same thunk to support multiple-dtype? Ideally, we should have the current user interface support a possible futur extension that will compile less thunk.

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,153 @@
...
+    // ...
+
+    args[0] = input->ga.data;
+    args[1] = &dims[0];
+    args[2] = &dims[1];
+    ls = 1;
+    gs = 256;
+    err = GpuKernel_call(&%(kname)s, 1, &ls, &gs, 0, args);
+
+    // ...
```
nouiz: Add this in the full example.

doc/extending/extending_theano_gpu.txt
abergeron: I'm not sure I want to document the current behaviour since I don't think it is really good and we should rework it to make a better one.

theano/gpuarray/basic_ops.py
abergeron: If you don't want to use it, you don't set it at all.  Passing None is not supported.

theano/gpuarray/basic_ops.py
abergeron: ok

theano/gpuarray/basic_ops.py
abergeron: ok

doc/extending/extending_theano_gpu.txt
abergeron: ok

abergeron:
I've added a test for CGpuKernelBase in addition to addressing all the other comments.

nouiz:
Need to document op_lifter. We will change the interface shortly, so after that we should do it.

theano/gpuarray/basic_ops.py
```diff
@@ -377,6 +390,17 @@ def f(*args):
     return f
 
 
+def get_dtype(s):
```
nouiz: There is many different dtype. Can you have a more precise name and add a docstring?

theano/gpuarray/tests/test_cgpukernelbase.py
```diff
@@ -0,0 +1,69 @@
...
+from .config import mode_with_gpu, test_ctx_name
+
+from ..basic_ops import CGpuKernelBase
+from ..type import GpuArrayType, get_context
+
+
+from pygpu.gpuarray import dtype_to_typecode
+
+
+class TSTGpuEye(CGpuKernelBase):
```
nouiz: use TGpuEye or TestGpuEye. Don't start a new scheme for test name, we already have too many.

theano/gpuarray/tests/tstgpueye.c
```diff
@@ -0,0 +1,46 @@
...
+                   ctx, Py_None);
+  if (*z == NULL)
+    return -1;
+
+  args[0] = (*z)->ga.data;
+  args[1] = &dims[0];
+  args[2] = &dims[1];
+  ls = 1;
+  gs = 256;
+  err = GpuKernel_call(&k_eye, 1, &ls, &gs, 0, args);
```
nouiz: I find it strange that we define eye and we call k_eye. Can this be changed?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,200 @@
+
+.. _extending_theano_gpu:
+
+==============================
+Extending Theano with a GPU Op
+==============================
+
+This tutorial covers how to extend Theano with an op that offers a GPU
```
nouiz: Tell that this is for the new back-end and link to Theano doc for it.

doc/extending/extending_theano_gpu.txt
nouiz: Add a link to the test for a full example or include it (without copy/pasting the code).

theano/gpuarray/tests/test_cgpukernelbase.py
abergeron: This is not a test it's an op.

theano/gpuarray/tests/tstgpueye.c
nouiz: what about this comment?

theano/gpuarray/tests/test_cgpukernelbase.py
nouiz: right.

theano/gpuarray/tests/tstgpueye.c
abergeron: The thing is that you give a single name and it is used to generate three.  I could leave the kernel with just the bare name, but I think it's clearer that something different get prepended for each thing.

nouiz: What about user declare "eye" as the base name. But in the file, it call k_eye and declare k_eye? The only problem is that he declare eye and not k_eye. I think in the user script, he should declare and call the same name.

doc/extending/extending_theano_gpu.txt
abergeron: Done

doc/extending/extending_theano_gpu.txt
abergeron: Done.

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+==============================
+
+.. note::
+
+    This covers the :ref:`gpuarray <gpuarray>` back-end for the GPU.
+
+This tutorial covers how to extend Theano with an op that offers a GPU
+implementation.  It assumes you are familiar with how to write new
+Theano ops.  If that is not the case you should probably follow the
+:ref:`extending_theano` and :ref:`extending_theano_c` section before
```
lamblin: "sections"?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+continuing on.
+
+Writing a new GPU op can be done in python for some simple tasks, but
```
lamblin: Python

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+will usually done in C to access the complete API and avoid paying the
+overhead of a python function call.
```
lamblin: Python

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+
+Dealing With the Context
+========================
+
+One of the major differences with GPU ops is that they require a
+context (a.k.a. device) to execute.  Most of the time you can infer
+the context to run on from your inputs.  There is a way for the user
+to transfer things between contexts and to tag certain varaibles for
```
lamblin: "variables"

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+Writing a new GPU op can be done in python for some simple tasks, but
+will usually done in C to access the complete API and avoid paying the
+overhead of a python function call.
+
+Dealing With the Context
+========================
+
+One of the major differences with GPU ops is that they require a
+context (a.k.a. device) to execute.  Most of the time you can infer
+the context to run on from your inputs.  There is a way for the user
```
lamblin: How would a user do that? If there is documentation already, can we put a link?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+    def make_node(self, a, b, c):
+        ctx = infer_context_name(a, b, c)
+        a = as_gpuarray_variable(a, ctx)
+        b = as_gpuarray_variable(b, ctx)
+        c = as_gpuarray_variable(c, ctx)
+        return Apply(self, [a, b, c], [a.type()])
+
+In this example the Op takes three inputs, all on the GPU.  In case
+one or more of your inputs is not supposed to be on the GPU, you
+should not pass it to :func:`infer_context_name`.
```
lamblin: and not call `as_gpuarray_variable` on it either, I guess

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+
+Also note that :func:`theano.gpuarray.basic_ops.as_gpuarray_variable`
+takes ``context_name`` as a mandatory parameter.  This is because it's
+not enough to know you want the value to be on the GPU, you also want
+to know which GPU to put it on.  In almost all cases, you can pass in
+the return value of :func:`infer_context_name` there.
+
+If you also need the context during runtime (for example to allocate
+the output).  You can use the context of one of your inputs to know
```
lamblin: The first sentence does not have a verb. Put a comma instead?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+
+    def perform(self, node, inputs, output_storage):
+        A, B = inputs
+        C, = output_storage
+        C[0] = pygpu.empty([A.shape[0], B.shape[1]], dtype=A.dtype, A.context)
+        pygpu.blas.gemm(1, A, B, 0, C, overwrite_c=True)
+
+Finally if you require the context before perform, such as during
+make_thunk() to initialize kernels and such, you can access the
+context of your inputs through the type if the variables::
```
lamblin: "of"

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+
+    def make_thunk(self, node, storage_map, compute_map, no_recycling):
+        ctx = node.inputs[0].type.context
+
+Note that GpuArrayType objects also have a `context_name` attribute
```
lamblin: `GpuArrayType`

lamblin: Well, with double backticks so it is monotyped...

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+:class:`theano.gpuarray.basic_ops.GpuEye`.  This is not a case that
+you should encounter often, so it will not be covered further.
+
+Defining New Kernels
+====================
+
+If your op needs to do some transformation on the data, chances are
+that you will need to write a new kernel.  The best way to do this is
+to leverage GpuKernelBase (or CGpuKernelBase if you want to use the
+COp functionality).
```
lamblin: Link to autodoc reference, or monotype, for class names

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+``kernels``.  Inside that section you can define some kernels with
+``#kernel name:params:flags``.
+
+Here ``name`` is the name of the kernel function in the following
+code, ``params`` is a comma-separeted list of numpy typecode names.
+There are three exceptions for ``size_t`` which should be noted as
+``size``, ``ssize_t`` which should be noted as ``ssize`` and a pointer
+which should be noted as ``*``.
+
+``flags`` is a ``|``-separeted list of C kernel flag values (can be
```
lamblin: "separated"

lamblin: Is there a place where the existing flags are listed?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+empty).  The same kernel definition as above would look like this with
+``CGpuKernelBase``::
+
+    #section kernels
+
+    #kernel k : *, size, size : GA_USE_DOUBLE
```
lamblin: How do we know what it is a pointer to? I guess from the flag, but how can we specify different types?
Does the order of the flags matter?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+The second method is to handle the kernel compilation and cache on
+your own.  This is not recommended because there are lots of details
+to pay attention to that can cripple your performance if not done
+right, which GpuKernelBase handles for you.  But if you really want to
+go this way, then you can look up the C API for kernels in
+libgpuarray.
+
+In any case you will need to call your compiled kernel with some data.
```
lamblin: Maybe mention that we usually want to do that in the `c_code` method of the Op.

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+    void *args[3];
+
+    // ...
+
+    args[0] = input->ga.data;
+    args[1] = &dims[0];
+    args[2] = &dims[1];
+    ls = 1;
+    gs = 256;
+    err = GpuKernel_call(&%(kname)s, 1, &ls, &gs, 0, args);
```
lamblin: I have no idea what `1, &ls, &gs, 0,` are for. block size? grid size?
I guess `%(kname)s` would be `"k"` and `args` contains the arguments to pass to it.

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+For plain GpuKernelBase, you have to define a method called
+gpu_kernels which returns a list of :class:`Kernel
+<theano.gpuarray.basic_ops.Kernel>` objects.  You can define as many
+kernels as you want for a single op.  An example would look like this::
+
+    def gpu_kernels(self, node, name):
+        code = """
+    KERNEL void k(GLOBAL_MEM ga_double *a, ga_size n, ga_size m) {
+        ga_size nb = n < m ? n : m;
+        for (ga_size i = LID_0; i < nb; i += LDIM_0) {
```
lamblin: Where do `LID_0` and `LDIM_0` come from?

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,236 @@
...
+            a[i*m + i] = 1;
+        }
+    }"""
+        return [Kernel(
+                code=code, name="k",
+                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE],
+                flags=Kernel.get_flags('float64'))]
```
lamblin: The full example also has `objvar='k_eye_' + name`. Is that needed?

theano/gpuarray/basic_ops.py
```diff
@@ -124,6 +125,51 @@ class Kernel(object):
...
+    codevar: str
+        the name of the variable for the code object.
+        (defaults to `kcode_` + name)
+    binvar: str
+        the name of the variable for the binary object.
+        (defaults to `kbin_` + name)
+    objvar: str
+        the name of the variable for the kernel object.
+        (defaults to `k_` + name)
+
```
lamblin: If that is autogenerated somewhere, we should link to it in the how to.

theano/gpuarray/tests/test_cgpukernelbase.py
```diff
@@ -0,0 +1,70 @@
...
+
+from .config import mode_with_gpu, test_ctx_name
+
+from ..basic_ops import CGpuKernelBase
+from ..type import GpuArrayType, get_context
+
+
+from pygpu.gpuarray import dtype_to_typecode
+
+
```
lamblin: Can you add a comment (not in the docstring) explaining why this implementation is here?

doc/extending/extending_theano_gpu.txt
lamblin: Apparently, `%(kname)s` is actually the result of `self.gpu_kernels(node, name)[0].objvar`. I think it should be explained somewhere.

theano/gpuarray/tests/tstgpueye.c
lamblin: Yes, this is a bit confusing. If it is an automatically-generated name, I would expect something with a "macro" look, with all-caps or something, like `DTYPE_INPUT_0` or something like that.

lamblin: I'm not saying to change it right now, but at least explain what is going on.

lamblin:
I think I'm done. I stopped the nitpicking about monotype and formatting early on.

doc/extending/extending_theano_gpu.txt
abergeron: From the libgpuarray abstraction language.

doc/extending/extending_theano_gpu.txt
abergeron: There is documentation about `.transfer()`, but that is really just a sidenote here insiting on the point that it is possible that you get arguments that aren't from the same context in your make_node() and you need to handle it.

doc/extending/extending_theano_gpu.txt
abergeron: Yes of course.  I hope that is obvious.  But sometimes you need to spell out the obvious.

doc/extending/extending_theano_gpu.txt
abergeron: The libgpuarray docs (which might need some love).

doc/extending/extending_theano_gpu.txt
abergeron: We don't need to specify the type that is pointed to, it doesn't matter for the call protocol.

Also, the order of the flags doesn't matter.

doc/extending/extending_theano_gpu.txt
abergeron: I don't want to turn this into an API reference for libgpuarray so I don't want to delve into the details of the arguments of the C-API.  It's bound to get stale and wrong due to lack of updates if we document that here.

doc/extending/extending_theano_gpu.txt
abergeron: It's not strictly required no.  If you want to know these things there is a the API reference for Kernel that is linked.

theano/gpuarray/basic_ops.py
abergeron: What is autogenerated?

doc/extending/extending_theano_gpu.txt
lamblin: OK.

doc/extending/extending_theano_gpu.txt
lamblin: Is that documented somewhere?

abergeron: No.

theano/gpuarray/tests/tstgpueye.c
abergeron: It is generated, but by the GpuKernelBase code.  It is also customizable by the user.  I can add a comment that explain where it comes from.

doc/extending/extending_theano_gpu.txt
lamblin: OK.

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,249 @@
...
+    void *args[3];
+
+    // ...
+
+    args[0] = input->ga.data;
+    args[1] = &dims[0];
+    args[2] = &dims[1];
+    ls = 1;
+    gs = 256;
+    err = GpuKernel_call(&k_obj, 1, &ls, &gs, 0, args);
```
lamblin: Can you say somewhere that `k_obj` is the name automatically generated for a `GpuKernel` object produced from the declaration of `k` above?
Actually, it looks like the automatically-generated name would be `k_k`, no?

theano/gpuarray/basic_ops.py
lamblin: It's OK, there is a link to the docstring of class `Kernel` earlier

theano/gpuarray/tests/test_cgpukernelbase.py
lamblin: (Done.)

doc/extending/extending_theano_gpu.txt
lamblin: So, basically what you added in a comment to the .c file, which I had not read yet.

doc/extending/extending_theano_gpu.txt
```diff
@@ -0,0 +1,252 @@
...
+    args[1] = &dims[0];
+    args[2] = &dims[1];
+    ls = 1;
+    gs = 256;
+    err = GpuKernel_call(&k_k, 1, &ls, &gs, 0, args);
+
+    // ...
+
+The name of the kernel object depends on the name you passed to
+Kernel() when you declared it (or the name in you `#kernel`
```
lamblin: `Kernel()`, "your"

lamblin:
Otherwise, I'm ready do merge.

abergeron:
Ok, done.

lamblin:
If this fails while the previous test passed, I'll eat my hat.
Disclaimer: I'm not wearing any hat.

PR #4587, opened 2016-06-05
Implementation of 2D dilated convolution/correlation.

niasla:
Comments:
- For some reason when tensor.nnet.conv2d is called then convop is implemented with the debug slow version in abstract_conv.py. Modified already opt.py (both cpu and gpu versions) and dnn.py (both gpuarray and cuda). The idea is that whenever tensor.nnet.conv2d is called we need to check wether there is dilation and if there is the call conv_gemm implementation. 
- Following the previous point, the TestDnnConv2d in test_sandbox/cuda/tests/abstractconv.py fails. The reason is the use of cuDNN implementation of dilated architectures in tests (should we skip them or adapt with the abstract iface?)

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1393,6 +1393,9 @@ def local_abstractconv_cudnn(node):
     inp1 = node.inputs[0]
     inp2 = node.inputs[1]
 
+    if (node.op.filter_dilation != (1, 1)):
+        return None
+
     if not isinstance(inp1.type, GpuArrayType):
         return None
 
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -855,15 +855,17 @@ class BaseGpuCorrMM(GpuOp):
         or a pair of integers
     subsample
         Perform subsampling of the output (default: (1, 1)).
+    filter_dilation
+        Perform subsampling of the input, also known as dilation (default: (1, 1)).
     pad
         *deprecated*, now you should always use border_mode.
-
     """
 
     check_broadcast = False
-    __props__ = ('border_mode', 'subsample')
+    __props__ = ('border_mode', 'subsample', 'filter_dilation')
 
-    def __init__(self, border_mode="valid", subsample=(1, 1), pad=(0, 0)):
+    def __init__(self, border_mode="valid", subsample=(1, 1),
+                 filter_dilation=(1, 1), pad=(0, 0)):
         if pad != (0, 0):
             _logger.warning(
                 'do not use pad for BaseGpuCorrMM; please set padding in '
@@ -885,7 +887,10 @@ class BaseGpuCorrMM(GpuOp):
         self.border_mode = border_mode
         if len(subsample) != 2:
             raise ValueError("subsample must have two elements")
-        self.subsample = subsample
+        if len(filter_dilation) != 2:
+            raise ValueError("filter_dilation must have two elements")
+        self.subsample = tuple(subsample)
+        self.filter_dilation = tuple(filter_dilation)
 
     @property
     def pad(self):
@@ -894,10 +899,11 @@ class BaseGpuCorrMM(GpuOp):
         return (0, 0)
 
     def __str__(self):
-        return '%s{%s, %s}' % (
+        return '%s{%s, %s, %s}' % (
             self.__class__.__name__,
             self.border_mode,
-            str(self.subsample))
+            str(self.subsample),
+            str(self.filter_dilation))
 
     def flops(self, inp, outp):
         """
@@ -922,7 +928,7 @@ class BaseGpuCorrMM(GpuOp):
 
     def c_code_cache_version(self):
         # raise this whenever modifying any of the support_code_files
-        return (0, 24)
+        return (0, 26)
 
     def c_support_code_apply(self, node, nodename):
         # REMEMBER TO RAISE c_code_cache_version when changing any of
@@ -976,6 +982,7 @@ class BaseGpuCorrMM(GpuOp):
 
         """
         dH, dW = self.subsample
+        dilH, dilW = self.filter_dilation
         if self.border_mode == "half":
             padH = padW = -1
         elif self.border_mode == "full":
@@ -1022,6 +1029,8 @@ class BaseGpuCorrMM(GpuOp):
     // Optional args
     int dH = %(dH)s;
     int dW = %(dW)s;
+    int dilH = %(dilH)s;
+    int dilW = %(dilW)s;
     int padH = %(padH)s;
     int padW = %(padW)s;
 
@@ -1045,39 +1054,43 @@ class BaseGpuCorrMM(GpuOp):
         }
         else if (padH == -2) {
             // vertical full padding, we can infer the kernel height
-            kH = 2 - CudaNdarray_HOST_DIMS(bottom)[2] + (CudaNdarray_HOST_DIMS(top)[2] - 1) * dH;
+            kH = (2 - CudaNdarray_HOST_DIMS(bottom)[2] + (CudaNdarray_HOST_DIMS(top)[2] - 1)*dH - 1) / dilH + 1;
         }
         else {
             // explicit padding, we can infer the kernel height
-            kH = CudaNdarray_HOST_DIMS(bottom)[2] + 2*padH - (CudaNdarray_HOST_DIMS(top)[2] - 1) * dH;
+            kH = (CudaNdarray_HOST_DIMS(bottom)[2] + 2*padH - (CudaNdarray_HOST_DIMS(top)[2] - 1)*dH - 1) / dilH + 1 ;
         }
         if ((dW != 1) || (padW == -1)) {
             kW = %(width)s;
         }
         else if (padW == -2) {
-            kW = 2 - CudaNdarray_HOST_DIMS(bottom)[3] + (CudaNdarray_HOST_DIMS(top)[3] - 1) * dW;
+            kW = (2 - CudaNdarray_HOST_DIMS(bottom)[3] + (CudaNdarray_HOST_DIMS(top)[3] - 1) * dW - 1) / dilW + 1;
         }
         else {
-            kW = CudaNdarray_HOST_DIMS(bottom)[3] + 2*padW - (CudaNdarray_HOST_DIMS(top)[3] - 1) * dW;
+            kW = (CudaNdarray_HOST_DIMS(bottom)[3] + 2*padW - (CudaNdarray_HOST_DIMS(top)[3] - 1) * dW - 1) / dilW + 1;
         }
     }
 
+    // Implicit dilated kernel size
+    int dil_kH = (kH - 1) * dilH + 1;
+    int dil_kW = (kW - 1) * dilW + 1;
+
     // Auto-padding if requested
     if (padH == -1) {  // vertical half padding
-        padH = kH / 2;
+        padH = dil_kH / 2;
     }
     else if (padH == -2) {  // vertical full padding
-        padH = kH - 1;
+        padH = dil_kH - 1;
     }
     else if (padH < 0) {
         PyErr_SetString(PyExc_ValueError, "BaseGpuCorrMM: padH must be >= -2");
         %(fail)s
     }
     if (padW == -1) {  // horizontal half padding
-        padW = kW / 2;
+        padW = dil_kW / 2;
     }
     else if (padW == -2) {  // horizontal full padding
-        padW = kW - 1;
+        padW = dil_kW - 1;
     }
     else if (padW < 0) {
         PyErr_SetString(PyExc_ValueError, "BaseGpuCorrMM: padW must be >= -2");
@@ -1089,15 +1102,15 @@ class BaseGpuCorrMM(GpuOp):
     switch(direction) {
     case 0:  // forward pass
         // output is top: (batchsize, num_filters, height, width)
-        // height and width: top = (bottom + 2*pad - weight) / sample + 1
+        // height and width: top = (bottom + 2*pad - ((weight-1)*dil + 1)) / sample + 1
         out_dim[0] = CudaNdarray_HOST_DIMS(bottom)[0];
         out_dim[1] = CudaNdarray_HOST_DIMS(weights)[0];
-        out_dim[2] = (CudaNdarray_HOST_DIMS(bottom)[2] + 2*padH - CudaNdarray_HOST_DIMS(weights)[2]) / dH + 1;
-        out_dim[3] = (CudaNdarray_HOST_DIMS(bottom)[3] + 2*padW - CudaNdarray_HOST_DIMS(weights)[3]) / dW + 1;
+        out_dim[2] = (CudaNdarray_HOST_DIMS(bottom)[2] + 2*padH - ((CudaNdarray_HOST_DIMS(weights)[2]-1)*dilH + 1)) / dH + 1;
+        out_dim[3] = (CudaNdarray_HOST_DIMS(bottom)[3] + 2*padW - ((CudaNdarray_HOST_DIMS(weights)[3]-1)*dilW + 1)) / dW + 1;
         break;
     case 1:  // backprop wrt. weights
         // output is weights: (num_filters, num_channels, height, width)
-        // height and width: weights = bottom + 2*pad - (top - 1) * sample
+        // height and width: weights = (bottom + 2*pad - (top - 1) * sample - 1) / dil + 1
         out_dim[0] = CudaNdarray_HOST_DIMS(top)[1];
         out_dim[1] = CudaNdarray_HOST_DIMS(bottom)[1];
         out_dim[2] = kH;  // already inferred further above
@@ -1105,11 +1118,11 @@ class BaseGpuCorrMM(GpuOp):
         break;
     case 2:  // backprop wrt. inputs
         // output is bottom: (batchsize, num_channels, height, width)
-        // height and width: bottom = (top - 1) * sample + weights - 2*pad
+        // height and width: bottom = (top - 1) * sample + (weights-1)*dil + 1 - 2*pad
         out_dim[0] = CudaNdarray_HOST_DIMS(top)[0];
         out_dim[1] = CudaNdarray_HOST_DIMS(weights)[1];
-        out_dim[2] = (dH != 1) ? %(height)s : (CudaNdarray_HOST_DIMS(top)[2] - 1) * dH + CudaNdarray_HOST_DIMS(weights)[2] - 2*padH;
-        out_dim[3] = (dW != 1) ? %(width)s : (CudaNdarray_HOST_DIMS(top)[3] - 1) * dW + CudaNdarray_HOST_DIMS(weights)[3] - 2*padW;
+        out_dim[2] = (dH != 1) ? %(height)s : (CudaNdarray_HOST_DIMS(top)[2] - 1) * dH + (CudaNdarray_HOST_DIMS(weights)[2]-1)*dilH + 1 - 2*padH;
+        out_dim[3] = (dW != 1) ? %(width)s : (CudaNdarray_HOST_DIMS(top)[3] - 1) * dW + (CudaNdarray_HOST_DIMS(weights)[3]-1)*dilW + 1 - 2*padW;
         break;
     default:
         PyErr_SetString(PyExc_ValueError, "BaseGpuCorrMM: direction must be 0, 1, or 2\\n");
@@ -1137,7 +1150,7 @@ class BaseGpuCorrMM(GpuOp):
     }
 
     // Call CUDA code
-    out2 = corrMM(%(bottom)s, %(weights)s, %(top)s, direction, dH, dW, padH, padW);
+    out2 = corrMM(%(bottom)s, %(weights)s, %(top)s, direction, dH, dW, dilH, dilW, padH, padW);
     if (out2==NULL){
        %(fail)s
     }
@@ -1168,6 +1181,10 @@ class GpuCorrMM(BaseGpuCorrMM):
         `(sv, sh)` is equivalent to `GpuCorrMM(...)(...)[:,:,::sv, ::sh]`,
         but faster.
         Set to `(1, 1)` to disable subsampling.
+    filter_dilation
+        The filter dilation operation applied to each input image.
+        Should be a tuple with 2 elements.
+        Set to `(1, 1)` to disable filter dilation.
     pad
         Deprecated alias for `border_mode`.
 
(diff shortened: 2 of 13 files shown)


theano/sandbox/cuda/blas.py
```diff
@@ -855,15 +855,18 @@ class BaseGpuCorrMM(GpuOp):
...
 
     """
 
     check_broadcast = False
-    __props__ = ('border_mode', 'subsample')
+    __props__ = ('border_mode', 'subsample', 'filter_dilation')
 
-    def __init__(self, border_mode="valid", subsample=(1, 1), pad=(0, 0)):
+    def __init__(self, border_mode="valid", subsample=(1, 1),
+                 filter_dilation=(1, 1), pad=(0, 0)):
```
f0k: Even if `pad` is deprecated, you should probably add the new argument in the end to not break backwards compatibility for positional arguments. (What do the Theano devs say?)

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -167,6 +183,8 @@ CudaNdarray* corrMM(CudaNdarray *const bottom,
                     const int direction,
                     const int dH = 1,
                     const int dW = 1,
+		    const int dilH = 1,
+		    const int dilW = 1,
```
f0k: wrong indentation?

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -84,75 +85,90 @@ __global__ void im2col_kernel(const int n, const float* data_im,
...
-        val += data_col[(c_col * height_col + h_col) * width_col + w_col];
+	int h_k = h - h_col * stride_h;
+	int w_k = w - w_col * stride_w;
+	if (h_k % dilation_h == 0 && w_k % dilation_w == 0) {
+          h_k /= dilation_h;
+          w_k /= dilation_w;
+	  int c_col = c * patch_h * patch_w + h_k * patch_w
+            + w_k;
+	  val += data_col[(c_col * height_col + h_col) * width_col + w_col];
+	 }
```
f0k: Equivalent to Torch (https://github.com/torch/cunn/blob/master/lib/THCUNN/im2col.h).

Would be good to time whether this kernel is any slower for dilation_h = dilation_w = 1 than the previous version. If so, we could choose the version to use at compile time (the dilation factors are constants anyway).

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -84,75 +85,90 @@ __global__ void im2col_kernel(const int n, const float* data_im,
...
+      (c * patch_h * patch_w + h * patch_w + w) * height_col * width_col;
     int coeff_h_col = (1 - stride_h * patch_w * height_col) * width_col;
     int coeff_w_col = (1 - stride_w * height_col * width_col);
     for (int h_col = h_col_start; h_col < h_col_end; ++h_col) {
       for (int w_col = w_col_start; w_col < w_col_end; ++w_col) {
-        val += data_col[offset + h_col * coeff_h_col + w_col * coeff_w_col];
+	val += data_col[offset + h_col * coeff_h_col + w_col * coeff_w_col];
       }
-    }
+      }*/
```
f0k: Why these indentation changes? We may want to drop the "equivalent implementation" now that it's not equivalent any more.

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -286,7 +307,8 @@ CudaNdarray* corrMM(CudaNdarray *const bottom,
         for (int n = 0; n < batchSize; n++) {
             // First, im2col
             im2col(bottom->devdata + n * bottom_stride, nChannels, bottomHeight,
-                    bottomWidth, kH, kW, padH, padW, dH, dW, col->devdata);
+		   bottomWidth, kH, kW, dilH, dilW,
+		   padH, padW, dH, dW, col->devdata);
```
f0k: nitpick: I'd prefer either the previous indentation (8 spaces = a double indent) or one space more than you have now (a visual indent). Indenting by 6 spaces is not common.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -85,31 +90,37 @@ def get_conv_shape_1axis(image_shape, kernel_shape,
...
+    filter_dilation: int. It must correspond to the dilation on the
+        considered axis.
     border_mode: string or int. If it is a string, it must be
         'valid', 'half' or 'full'. If it is an integer, it must correspond to
         the padding on the considered axis.
     subsample: int. It must correspond to the subsampling on the
         considered axis.
-
+    dilation: int. It must correspond to the dilation on the
+        considered axis.
```
f0k: Now you have both dilation and filter_dilation in the docstring.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -118,6 +129,7 @@ def conv2d(input,
            filters,
            input_shape=None,
            filter_shape=None,
+           filter_dilation=(1, 1),
```
f0k: For backwards compatibility, this has to be added as the last argument.

niasla:
OK,  something weird is happening with the indentation. In my emacs configuration everything appears indented correctly. My setting is each indent =  4 spaces. It's weird in Github it appears differently . I'll modify the code following the recommendations and see how to fix indentation problem.

f0k:
> It's weird in Github it appears differently

In `corr_gemm.cu` it looks like you've used tab characters somewhere. I think the C code is not flake checked, though, so there may be funny indentation in the master branch already.

niasla:
It seems there is non compatible PEP8 code (in *.py) already in the master. I'll run test_flake8 before amending the commit and update all files.

f0k:
Okay, I'll let the Theano devs comment on this (I think some .py files are whitelisted for flake8) -- I just commented on what I noticed when reviewing! Better concentrate on the backwards-compatible interface and then we'll figure out the optimizer.

theano/sandbox/cuda/blas.py
niasla: Then we'll have to specify always pad when passing arguments by position (not a good idea for a deprecated arg). I have made the changes except for this one. Still if you guys consider it in that way then I shall make this change also.

f0k: > Then we'll have to specify always pad when passing arguments by position (not a good idea for a deprecated arg).

That's a good argument.

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -54,6 +54,7 @@ inline int GET_BLOCKS(const int N) {
 // Kernels for fast unfold + copy
 __global__ void im2col_kernel(const int n, const float* data_im,
     const int height, const int width, const int kernel_h, const int kernel_w,
+    const int dilation_h, const int dilation_w,
```
f0k: Still: What about re-copying the code from Caffe instead of making our own modifications? Might make it easier to maintain. (https://github.com/BVLC/caffe/blob/master/src/caffe/util/im2col.cu)

theano/sandbox/cuda/opt.py
```diff
@@ -1589,19 +1589,20 @@ def local_conv_gemm(node):
...
+            rval = GpuCorrMM(border_mode, subsample, filter_dilation)(
                 gpu_contiguous(img), gpu_contiguous(kern))
 
             # call GpuCorrMM_gradWeights if good
             # (the latter is faster if batchsize * kernelHeight * kernelWidth
             # is larger than inputChannels * outputHeight * outputWidth.
             # GpuConv does not always store information on the batchsize and
             # channels, though, so we only use what information we have.)
-            if ((subsample == (1, 1)) and
+            if ((subsample == (1, 1)) and filter_dilation == 1 and
```
f0k: Shouldn't this be `(1, 1)`? Is that normalized to become a 2-tuple in the constructor?

niasla: Definitely, sorry for that!

theano/sandbox/cuda/corr_gemm.cu
niasla: I've used Caffe code only in col2im_kernel in corr_gemm.cu  (the part of h_k ... ).
The rest of the code was written derived from the Theano code. But looking at Caffe code, there is high analogy between both codes. If you think we should copy exactly the code from Caffe then please confirm it to me and I shall copy it and re-submit.

niasla: Switched to Caffe implementation both im2col and col2im CUDA kernels.

f0k: > But looking at Caffe code, there is high analogy between both codes.

Yes, because I originally copied it from Caffe. Everything that has changed in between is probably because it has changed in Caffe.

> Switched to Caffe implementation both im2col and col2im CUDA kernels.

Cool! I think this is easier to maintain for the future.

niasla:
I've fixed the tests, basically the cuDNN tests and basic (not gemm) CpuConv tests are skipped when dilation is different from (1, 1). Now it passes the regression tests.
What remains is to correctly make tensor.nnet.conv2d to use gemm version when dilation != (1, 1) . When I use tensor.nnet.conv2d it uses the debug slow implementation (it skips gemm somehow..), so right now the only way to use dilation is to make CorrMM Op and apply it.
Any guidance on how to make the optimizer to choose gemm implementation when dilation is used ? I've tried in opt.py (cpu and gpu) and dnn.py and still not working.

lamblin:
Can you put a breakpoint in `theano/tensor/nnet/opt.py` in function `local_abstractconv_gemm`, and check if it gets applied (or why not)? Then the same for the gradient ops.
If your inputs are on GPU, then you should do the same in `theano/sandbox/cuda/opt.py` for `local_conv_gemm`.

The thing that is troubling for me is that you say it falls back on the slow debugging implementation of convolution in the abstract op. This should not happen by default, normally there is an optimization raising an exception if there are still an abstract op in the graph. Did you disable some optimizations manually?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -718,10 +751,17 @@ def conv2d(self, img, kern, mode="valid"):
...
                     for im0 in xrange(img.shape[1]):
-                        # some cast generates a warning here
-                        out[b, n, ...] += _convolve2d(img[b, im0, ...],
-                                                      kern[n, im0, ...],
-                                                      1, val, bval, 0)
+                        for row_offset in xrange(dilation[0]):
+                            for col_offset in xrange(dilation[1]):
+                                fragment = img[b, im0,
+                                               row_offset::dilation[0],
+                                               col_offset::dilation[1]]
```
lamblin: Why do we need two levels of nesting here?
Is it not faster to explicitly allocate a bigger kernel full of zeros, and set the non-zeros from the kernel, like we do in in the backward pass with strides, when we need to dilate the output gradient?

niasla: Yeah well didn't consider speed given it's a debug mode implementation. I've changed the implementation and will push it today. As for the optimizer, I was **wrong**  , it does use the gemm version correctly. It seems that in my script gemm was excluded (without noticing it). So i created a new clean script and indeed it used the gemm version when dilation != (1, 1). Will post the updates later today.

lamblin: No problem, it's just that we run correctness tests against that implementation, and those can take long...

lamblin:
You will need to rebase on top of the latest master (and then push --force) to resolve the conflicts.

niasla:
Done, merged and pushed :+1:

theano/sandbox/cuda/tests/test_abstractconv.py
```diff
@@ -56,28 +61,30 @@ def setUp(self):
...
+        self.run_fwd(inputs_shape=i, filters_shape=f,
+                     subsample=s, verify_grad=True, mode=mode,
                      provide_shape=provide_shape, border_mode=b,
-                     filter_flip=flip,
-                     target_op=(GpuCorrMM,
-                                GpuCorrMM_gradWeights,
-                                GpuCorrMM_gradInputs))
+                     filter_flip=flip, target_op=(GpuCorrMM,
+                                                  GpuCorrMM_gradWeights,
+                                                  GpuCorrMM_gradInputs),
```
f0k: Some unnecessary whitespace change (also the subsample=s above)... makes the history a bit less clean, but it's a nitpick.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -90,26 +95,31 @@ def get_conv_shape_1axis(image_shape, kernel_shape,
...
 
     Returns
     -------
     out_shp: int corresponding to the output image shape on the
         considered axis. None if undefined.
 
     """
-    if None in [image_shape, kernel_shape, border_mode, subsample]:
+    if None in [image_shape, kernel_shape, dilation,
+                border_mode, subsample]:
```
f0k: I'd have expected the same order as in the argument list, but that's not critical.

theano/tensor/nnet/corr.py
```diff
@@ -27,12 +27,14 @@ class BaseCorrMM(gof.Op):
         or a pair of integers
     subsample
         Perform subsampling of the output (default: (1, 1)).
-
+    filter_dilation
+        Perform dilated corrlation (default: (1,1))
```
f0k: *correlation

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -139,19 +159,24 @@ def run_fwd(self, inputs_shape, filters_shape, ref=conv_corr,
 
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
+        log.debug("CorrMM Shape: %r", res_ref.shape)
+        log.debug("Abstr. conv2d Shape: %r", res.shape)
```
f0k: Is this supposed to stay in the test? Won't `utt.assert_allclose` check the shapes anyway?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -189,11 +216,14 @@ def run_gradweight(self, inputs_shape, filters_shape, output_shape,
 
         res_ref = numpy.array(f_ref())
         res = numpy.array(f())
+        log.debug("CorrMM gradW = %r", res_ref)
+        log.debug("Abstract gradW = %r", res)
         utt.assert_allclose(res_ref, res)
```
f0k: Same here -- is the log needed?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -253,28 +285,38 @@ def abstract_conv2d_gradinputs(filters_val, output_val):
                             mode=mode, eps=1)
 
     def test_all(self):
+        log = logging.getLogger("BaseTestConv2d.test_all")
         if type(self) is BaseTestConv2d:
             raise SkipTest("base class")
         ds = [1, 1]
+        dfd = [1, 1]  # default filter dilation
```
f0k: Maybe just give `self.tcase` a default value of `(1, 1)` for the dilation?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -591,7 +642,7 @@ def test_bilinear_kernel_2D(self):
             # getting the normalized kernel
             kernel = bilinear_kernel_2D(ratio=ratio, normalize=True)
             f = theano.function([], kernel)
-            kernel_2D = kernel_2D / float(ratio**2)
+            kernel_2D = kernel_2D / float(ratio ** 2)
```
f0k: Not the business of this PR!

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -685,7 +736,7 @@ def get_upsampled_twobytwo_mat(self, two_by_two, ratio):
         if ratio % 2 == 0:
             up_2D = np.concatenate([up_2D, up_2D[:, :, -1:, :]], axis=2)
             up_2D = np.concatenate([up_2D, up_2D[:, :, :, -1:]], axis=3)
-        return up_2D / float(ratio)**2
+        return up_2D / float(ratio) ** 2
```
f0k: Same here.

f0k:
Looks good to me except for minor comments! Great job! Are there any other graph optimizers for AbstractConv2d* that need to be skipped for dilation > (1,1)?

niasla:
Updated the code as recommended and pushed. I think we've covered all AbstractConv2d optimizations (sandbox/cuda/dnn.py, theano/gpuarray/dnn.py, sandbox/cuda/opt.py and theano/tensor/nnet/opt.py). Are there any other places I should look into ?

lamblin:
I think this should cover all optimizations indeed.

niasla:
Is this PR going to get merged ?
Are there any other recommendations I should address?

f0k:
For me everything looks fine!

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -84,75 +86,90 @@ __global__ void im2col_kernel(const int n, const float* data_im,
...
+        (h_im < kernel_extent_h) ? 0 : (h_im - kernel_extent_h) / stride_h + 1;
+    const int h_col_end = min(h_im / stride_h + 1, height_col);
+    // TODO: use LCM of stride and dilation to avoid unnecessary loops
+    for (int h_col = h_col_start; h_col < h_col_end; h_col += 1) {
+      for (int w_col = w_col_start; w_col < w_col_end; w_col += 1) {
+        int h_k = (h_im - h_col * stride_h);
+        int w_k = (w_im - w_col * stride_w);
+        if (h_k % dilation_h == 0 && w_k % dilation_w == 0) {
+          h_k /= dilation_h;
+          w_k /= dilation_w;
```
nouiz: doing modulo and division in the inner loop is bad. This is pretty slow on the GPU. What about using just a kernel, but have this in it:

if (dilation_h == 0 && dilation_w == 0){
 old code
}else{new code}

this way we will be sure to not introduce slow down for the old cases.

theano/sandbox/cuda/corr_gemm.cu
```diff
@@ -54,28 +54,30 @@ inline int GET_BLOCKS(const int N) {
...
       for (int j = 0; j < kernel_w; ++j) {
-        int h = h_in + i;
-        int w = w_in + j;
-        *data_col_ptr = (h >= 0 && w >= 0 && h < height && w < width) ?
-            data_im_ptr[i * width + j] : 0;
+        int h_im = h_offset + i * dilation_h;
+        int w_im = w_offset + j * dilation_w;
+        *data_col_ptr =
+            (h_im >= 0 && w_im >= 0 && h_im < height && w_im < width) ?
+            data_im_ptr[i * dilation_h * width + j * dilation_w] : 0;
```
nouiz: Even if the code change is small here, I would also do this to be sure to not cause slowdown:

if (dilation == 1){old code}else{new code}

nouiz:
I did a quick pass. I got 2 small comments to make sure to don't have slowdown when we don't use dilation. I'll let @lamblin or @ballasn merge. They know more the AbstractConv interface, so they could spot something I missed.

niasla:
Definitely, I would go even further and have 2 CUDA kernels for the dilation case and normal case to not branch the execution flow in CUDA_KERNEL_LOOP and branch it in the host. Updated the code and pushed.

f0k:
> to not branch the execution flow in CUDA_KERNEL_LOOP

If all threads of the warp choose the same code path, the branch shouldn't hurt much. We can easily afford branching on the host, though, :+1: for going the extra mile!

niasla:
Any other recommendations for this PR ?

lamblin:
@ballasn did you want to have a look?

lamblin:
Thanks a lot, it looks good to me.

niasla:
Thanks to all of you @nouiz @lamblin @f0k .

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -266,15 +291,18 @@ def test_all(self):
...
-                    try:
-                        self.tcase(i, f, s, db, dflip, dprovide_shape)
-                    except SkipTest as e:
-                        skipped = e
+            for fd in self.filters_dilations:
+                for s in self.subsamples:
+                    for b in self.border_modes:
+                        try:
+                            self.tcase(i, f, s, db, dflip,
+                                       dprovide_shape, fd)
```
Sentient07: Hello, 
I've tried to run the test in `theano/gpuarray/tests/test_abstractconv.py` and found that tests failing because of a small mistake here. For making the tests pass, I had to remove this parameter `fd`. I don't know if this is the correct solution and would like to know from you people. The tests `TestCorrConv2d`, which uses this test as its parent class has a `tcase` method with a keyword argument fd. Would it be apt to move the `tcase` method to this class ?
cc : @lamblin @nouiz

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -24,82 +23,99 @@
...
+                                  subsample,
+                                  filter_dilation)(filters,
+                                                   topgrad,
+                                                   inputs_shape[2:])
 
 
 class TestGetConvOutShape(unittest.TestCase):
     def test_basic(self):
-        image_shape, kernel_shape = (3, 2, 8, 9), (4, 2, 5, 6)
+        image_shape, kernel_shape = (3, 2, 12, 9), (4, 2, 5, 6)
```
Sentient07: If `pad==0` , it just computes `image_shape - kernel_shape`. So, shouldn't all the assertions also be changed ? This test fails for me. For anyone else as well ? 
cc : @lamblin  @nouiz

Sentient07: Oh! There was a merge conflict and i didn't resolve it correctly before. I'm sorry about this.

niasla:
that's weird since it passed all the regression tests. What's exactly the error message? I'll install pygpu and try to run this test tomorrow to try to reproduce the issue.

niasla:
OK I've seen now your new comments, I suppose everything is OK then :) .

Sentient07:
@niasla except one, everything passes :)
https://github.com/Theano/Theano/pull/4587#discussion_r69383085 the issue i reported in this comment happens. You can test it by running 

`THEANO_FLAGS=device=cuda,lib.cnmem=0.65 theano-nose theano/gpuarray/tests/test_dnn.py:TestDnnConv2d`

theano/gpuarray/dnn.py
```diff
@@ -1393,6 +1393,9 @@ def local_abstractconv_cudnn(node):
     inp1 = node.inputs[0]
     inp2 = node.inputs[1]
 
+    if (node.op.filter_dilation != (1, 1)):
```
nouiz: Could we do here as Lasagne do and use of the grad conv from cudnn in some condition? We should time it before. Just in case someone is interrested to spend time on this.

f0k: I've got this on my todo list as well. I actually expect the cuDNN gradient to be slower (especially if we keep the gpu_contiguous on its arguments), but haven't tried yet.

f0k: Wait, an interesting side effect of how we do it in Lasagne is that the *gradient* of the dilated convolution becomes a forward pass. This is something worth considering in the optimizer as well. Maybe we should reactivate the meta-optimizer, but then we need symbolic stride inference to make it work correctly.

PR #4591, opened 2016-06-06
corr_gemm optimization to improve CNN performance

pcs-theano:
Hi experts,

This PR is to optimize current corr_gemm Ops (including corrMM, corrMM_gradWeights and corrMM_gradInputs), which uses openmp to unroll the first loop (batch size level), and it turns out 5.2x (count all these three Ops in) speedup compared than original version.

Thanks.
Ciyong

### DIFF
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -137,23 +137,24 @@ class BaseCorrMM(gof.OpenMPOp):
 
         if self.openmp:
             sub['omp_flags'] = '#pragma omp parallel for schedule(static)'
-            sub['omp_max_threads'] = 'omp_get_max_threads()'
-            sub['set_omp_threads'] = 'omp_set_num_threads'
-            sub['get_omp_threads'] = 'omp_get_thread_num()'
+            sub['omp_get_max_threads'] = 'omp_get_max_threads()'
+            sub['omp_get_thread_num'] = 'omp_get_thread_num()'
 
             if self.blas_type == 'openblas':
-                sub['set_blas_threads'] = 'openblas_set_num_threads'
-                sub['get_blas_threads'] = 'openblas_get_num_threads()'
+                sub['blas_set_num_threads'] = 'openblas_set_num_threads'
+                sub['blas_get_num_threads'] = 'openblas_get_num_threads()'
             elif self.blas_type == 'mkl':
-                sub['set_blas_threads'] = 'mkl_set_num_threads'
-                sub['get_blas_threads'] = 'mkl_get_max_threads()'
+                sub['blas_set_num_threads'] = 'mkl_set_num_threads'
+                sub['blas_get_num_threads'] = 'mkl_get_max_threads()'
+            else:
+                sub['blas_set_num_threads'] = ''
+                sub['blas_get_num_threads'] = '0'
         else:
             sub['omp_flags'] = ''
-            sub['omp_max_threads'] = '1'
-            sub['set_omp_threads'] = ''
-            sub['get_omp_threads'] = '0'
-            sub['set_blas_threads'] = ''
-            sub['get_blas_threads'] = '0'
+            sub['omp_get_max_threads'] = '1'
+            sub['omp_get_thread_num'] = '0'
+            sub['blas_set_num_threads'] = ''
+            sub['blas_get_num_threads'] = '0'
 
         files = ['corr_gemm.c']
         codes = [open(os.path.join(os.path.split(__file__)[0], f)).read()
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -9,7 +9,7 @@ from theano import Apply
 from theano import gof
 from theano.tensor import as_tensor_variable, TensorType
 from theano.tensor.nnet.abstract_conv import get_conv_output_shape
-from theano.tensor.blas_headers import blas_header_text
+from theano.tensor import blas_headers
 from theano.tensor.blas import ldflags, blas_header_version
 
 _logger = logging.getLogger(__name__)
@@ -86,7 +86,12 @@ class BaseCorrMM(gof.OpenMPOp):
             str(self.filter_dilation))
 
     def c_support_code(self):
-        return blas_header_text()
+        ccodes = blas_headers.blas_header_text()
+        if self.blas_type == 'openblas':
+            ccodes += blas_headers.openblas_threads_text()
+        elif self.blas_type == 'mkl':
+            ccodes += blas_headers.mkl_threads_text()
+        return ccodes
 
     def c_libraries(self):
         return ldflags()
@@ -105,10 +110,6 @@ class BaseCorrMM(gof.OpenMPOp):
     def c_headers(self):
         headers = ['<stdio.h>']
         headers += super(BaseCorrMM, self).c_headers()
-        if self.blas_type == 'openblas':
-            headers += ['cblas.h']
-        if self.blas_type == 'mkl':
-            headers += ['mkl.h']
         return headers
 
     def c_code_cache_version(self):
@@ -137,20 +138,22 @@ class BaseCorrMM(gof.OpenMPOp):
         if self.openmp:
             sub['omp_flags'] = '#pragma omp parallel for schedule(static)'
             sub['omp_max_threads'] = 'omp_get_max_threads()'
-            sub['omp_set_threads'] = 'omp_set_num_threads'
-            sub['omp_get_threads'] = 'omp_get_thread_num()'
+            sub['set_omp_threads'] = 'omp_set_num_threads'
+            sub['get_omp_threads'] = 'omp_get_thread_num()'
+
+            if self.blas_type == 'openblas':
+                sub['set_blas_threads'] = 'openblas_set_num_threads'
+                sub['get_blas_threads'] = 'openblas_get_num_threads()'
+            elif self.blas_type == 'mkl':
+                sub['set_blas_threads'] = 'mkl_set_num_threads'
+                sub['get_blas_threads'] = 'mkl_get_max_threads()'
         else:
             sub['omp_flags'] = ''
-            sub['omp_max_threads'] = 1
-            sub['omp_set_threads'] = ''
-            sub['omp_get_threads'] = 0
-
-        if self.blas_type == 'openblas':
-            sub['blas_flags'] = 'openblas_set_num_threads(1)'
-        elif self.blas_type == 'mkl':
-            sub['blas_flags'] = 'mkl_set_num_threads(1)'
-        else:
-            sub['blas_flags'] = ''
+            sub['omp_max_threads'] = '1'
+            sub['set_omp_threads'] = ''
+            sub['get_omp_threads'] = '0'
+            sub['set_blas_threads'] = ''
+            sub['get_blas_threads'] = '0'
 
         files = ['corr_gemm.c']
         codes = [open(os.path.join(os.path.split(__file__)[0], f)).read()
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -87,7 +87,6 @@ class BaseCorrMM(gof.OpenMPOp):
         compile_args = ldflags(libs=False, flags=True)
         compile_args += super(BaseCorrMM, self).c_compile_args()
         return compile_args
-            
 
     def c_lib_dirs(self):
         return ldflags(libs=False, libs_dir=True)
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -12,8 +12,6 @@ from theano.tensor.nnet.abstract_conv import get_conv_output_shape
 from theano.tensor.blas_headers import blas_header_text
 from theano.tensor.blas import ldflags, blas_header_version
 
-from multiprocessing import cpu_count
-
 _logger = logging.getLogger(__name__)
 
 
@@ -122,13 +120,13 @@ class BaseCorrMM(gof.OpenMPOp):
             sub['n_bytes'] = 8
             sub['c_float_type'] = 'double'
         if self.openmp:
-            sub['cores'] = self.cores
-            sub['omp_flags'] = '#pragma omp parallel for'
+            sub['omp_flags'] = '#pragma omp parallel for schedule(static)'
+            sub['omp_max_threads'] = 'omp_get_max_threads()'
             sub['omp_set_threads'] = 'omp_set_num_threads'
             sub['omp_get_threads'] = 'omp_get_thread_num()'
         else:
-            sub['cores'] = 1
             sub['omp_flags'] = ''
+            sub['omp_max_threads'] = 1
             sub['omp_set_threads'] = ''
             sub['omp_get_threads'] = 0
 
@@ -342,7 +340,7 @@ class BaseCorrMM(gof.OpenMPOp):
         else {
           typenum = PyArray_TYPE(bottom);
         }
-        %(out)s = (PyArrayObject*)PyArray_ZEROS(4,
+        %(out)s = (PyArrayObject*)PyArray_EMPTY(4,
                                           out_dim,
                                           typenum,
                                           0);
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -62,6 +62,16 @@ class BaseCorrMM(gof.OpenMPOp):
         self.subsample = tuple(subsample)
         self.filter_dilation = tuple(filter_dilation)
 
+        if not theano.config.blas.ldflags:
+            raise NotImplementedError("C code for corrMM* classes need a blas library.")
+        else:
+            if 'openblas' in theano.config.blas.ldflags:
+                self.blas_type = 'openblas'
+            elif 'mkl' in theano.config.blas.ldflags:
+                self.blas_type = 'mkl'
+            else:
+                self.blas_type = ''
+
     @property
     def pad(self):
         if self.border_mode != 'valid':
@@ -95,6 +105,10 @@ class BaseCorrMM(gof.OpenMPOp):
     def c_headers(self):
         headers = ['<stdio.h>']
         headers += super(BaseCorrMM, self).c_headers()
+        if self.blas_type == 'openblas':
+            headers += ['cblas.h']
+        if self.blas_type == 'mkl':
+            headers += ['mkl.h']
         return headers
 
     def c_code_cache_version(self):
@@ -119,6 +133,7 @@ class BaseCorrMM(gof.OpenMPOp):
             sub['float_typenum'] = 'NPY_DOUBLE'
             sub['n_bytes'] = 8
             sub['c_float_type'] = 'double'
+
         if self.openmp:
             sub['omp_flags'] = '#pragma omp parallel for schedule(static)'
             sub['omp_max_threads'] = 'omp_get_max_threads()'
@@ -129,6 +144,13 @@ class BaseCorrMM(gof.OpenMPOp):
             sub['omp_max_threads'] = 1
             sub['omp_set_threads'] = ''
             sub['omp_get_threads'] = 0
+
+        if self.blas_type == 'openblas':
+            sub['blas_flags'] = 'openblas_set_num_threads(1)'
+        elif self.blas_type == 'mkl':
+            sub['blas_flags'] = 'mkl_set_num_threads(1)'
+        else:
+            sub['blas_flags'] = ''
 
         files = ['corr_gemm.c']
         codes = [open(os.path.join(os.path.split(__file__)[0], f)).read()
@@ -173,8 +195,6 @@ class BaseCorrMM(gof.OpenMPOp):
             If self.border_mode == 'half', a variable giving the width of the
             filters for direction="backprop weights".  Ignored otherwise.
         """
-        if not theano.config.blas.ldflags:
-            raise NotImplementedError("C code for CorrMM* classes need a blas library.")
         dH, dW = self.subsample
         dilH, dilW = self.filter_dilation
         if self.border_mode == "half":
@@ -340,7 +360,8 @@ class BaseCorrMM(gof.OpenMPOp):
         else {
           typenum = PyArray_TYPE(bottom);
         }
-        %(out)s = (PyArrayObject*)PyArray_EMPTY(4,
+        //Change to PyArray_ZEROS which is faster than PyArray_EMPTY.
+        %(out)s = (PyArrayObject*)PyArray_ZEROS(4,
                                           out_dim,
                                           typenum,
                                           0);
--- a/theano/tensor/nnet/corr.py
+++ b/theano/tensor/nnet/corr.py
@@ -10,13 +10,14 @@ from theano import gof
 from theano.tensor import as_tensor_variable, TensorType
 from theano.tensor.nnet.abstract_conv import get_conv_output_shape
 from theano.tensor.blas_headers import blas_header_text
-from theano.tensor.blas import ldflags
+from theano.tensor.blas import ldflags, blas_header_version
 
+from multiprocessing import cpu_count
 
 _logger = logging.getLogger(__name__)
 
 
-class BaseCorrMM(gof.Op):
+class BaseCorrMM(gof.OpenMPOp):
     """
     Base class for `CorrMM`, `CorrMM_gradWeights` and
     `CorrMM_gradInputs`. Cannot be used directly.
@@ -34,7 +35,8 @@ class BaseCorrMM(gof.Op):
     __props__ = ('border_mode', 'subsample', 'filter_dilation')
 
     def __init__(self, border_mode="valid", subsample=(1, 1),
-                 filter_dilation=(1, 1)):
+                 filter_dilation=(1, 1), openmp=None):
+        super(BaseCorrMM, self).__init__(openmp=openmp)
         if isinstance(border_mode, integer_types):
             if border_mode < 0:
                 raise ValueError(
@@ -82,7 +84,10 @@ class BaseCorrMM(gof.Op):
         return ldflags()
 
     def c_compile_args(self):
-        return ldflags(libs=False, flags=True)
+        compile_args = ldflags(libs=False, flags=True)
+        compile_args += super(BaseCorrMM, self).c_compile_args()
+        return compile_args
+            
 
     def c_lib_dirs(self):
         return ldflags(libs=False, libs_dir=True)
@@ -91,11 +96,13 @@ class BaseCorrMM(gof.Op):
         return ldflags(libs=False, include_dir=True)
 
     def c_headers(self):
-        return ['<stdio.h>']
+        headers = ['<stdio.h>']
+        headers += super(BaseCorrMM, self).c_headers()
+        return headers
 
     def c_code_cache_version(self):
         # raise this whenever modifying any of the support_code_files
-        return (1, 2)
+        return (1, self.openmp, blas_header_version())
 
     def c_support_code_apply(self, node, nodename):
         # REMEMBER TO RAISE c_code_cache_version when changing any of
@@ -115,6 +122,17 @@ class BaseCorrMM(gof.Op):
             sub['float_typenum'] = 'NPY_DOUBLE'
             sub['n_bytes'] = 8
             sub['c_float_type'] = 'double'
+        if self.openmp:
+            sub['cores'] = self.cores
+            sub['omp_flags'] = '#pragma omp parallel for'
+            sub['omp_set_threads'] = 'omp_set_num_threads'
+            sub['omp_get_threads'] = 'omp_get_thread_num()'
+        else:
+            sub['cores'] = 1
+            sub['omp_flags'] = ''
+            sub['omp_set_threads'] = ''
+            sub['omp_get_threads'] = 0
+
         files = ['corr_gemm.c']
         codes = [open(os.path.join(os.path.split(__file__)[0], f)).read()
                  for f in files]
@@ -325,7 +343,7 @@ class BaseCorrMM(gof.Op):
         else {
           typenum = PyArray_TYPE(bottom);
         }
-        %(out)s = (PyArrayObject*)PyArray_EMPTY(4,
+        %(out)s = (PyArrayObject*)PyArray_ZEROS(4,
                                           out_dim,
                                           typenum,
                                           0);
@@ -376,9 +394,6 @@ class CorrMM(BaseCorrMM):
         Set to `(1, 1)` to disable filter dilation.
 
     """
-    def __init__(self, border_mode="valid", subsample=(1, 1),
-                 filter_dilation=(1, 1)):
-        super(CorrMM, self).__init__(border_mode, subsample, filter_dilation)
 
     def make_node(self, img, kern):
         img = as_tensor_variable(img)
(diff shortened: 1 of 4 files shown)


JesseLivezey:
This looks similar to #3689 for the corrMM and corrMM_gradInputs. Can you run the tests with gcc 4.9 or 5.x? For the other PR the tests passed with gcc 4.8, but failed with the other versions.

theano/tensor/nnet/corr.py
```diff
@@ -359,8 +377,8 @@ class CorrMM(BaseCorrMM):
         Set to `(1, 1)` to disable subsampling.
 
     """
-    def __init__(self, border_mode="valid", subsample=(1, 1)):
-        super(CorrMM, self).__init__(border_mode, subsample)
+    def __init__(self, border_mode="valid", subsample=(1, 1), openmp=None):
```
nouiz: In fact, we don't need that method for the 3 subclass. Can you just remove them?

theano/tensor/nnet/corr.py
```diff
@@ -85,11 +90,13 @@ def c_header_dirs(self):
...
     def c_headers(self):
-        return ['<stdio.h>']
+        headers = ['<stdio.h>']
+        headers += super(BaseCorrMM, self).c_headers()
+        return headers
 
     def c_code_cache_version(self):
         # raise this whenever modifying any of the support_code_files
-        return (1, 1)
+        return (1, blas_header_version())
```
nouiz: I think we should add self.openmp here as an element of the tuple.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -172,50 +171,61 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
 
     // Create temporary columns
-    npy_intp col_dim[2];
-    col_dim[0] = (npy_intp)(nChannels * kW * kH);
-    col_dim[1] = (npy_intp)(topHeight * topWidth);
-    PyArrayObject* col = (PyArrayObject*)PyArray_EMPTY(2,
-		                           col_dim,
-                                           PyArray_TYPE(top),
-					   0);
+    const int max_threads = %(cores)s < batchSize ? %(cores)s : batchSize;
```
nouiz: why not use omp_get_max_threads(); instead of core? I have the impression that using core will bypass the env variable OMP_NUM_THREADS=N. If we can keep that working, it would be good. It would also remove the dependency on multiprocessing.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -250,23 +260,58 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
+                                   weight_dim, PyArray_TYPE(weight), 0);
+
+        if (NULL == local_weight)
+        {
+            PyErr_Format(PyExc_RuntimeError,
+                    "CorrMM failed to allocate weight memory of %%ld x %%ld\n",
+                    weight_dim[0], weight_dim[1]);
+            return NULL;
+        }
+        local_weight = PyArray_GETCONTIGUOUS(local_weight);
```
nouiz: local_weight will be c contiguous by default. So no need to call that method.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -250,23 +260,58 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
     }
     else if (direction == 1) {  // backprop wrt. weights
         output = weight;
+        npy_intp weight_dim[2];
+        weight_dim[0] = (npy_intp)max_threads;
+        weight_dim[1] = (npy_intp)(M_ * K_);
+        PyArrayObject* local_weight = (PyArrayObject*)PyArray_ZEROS(2,
```
nouiz: Both col and local_weight should be deleted Py_DECREF(var) at the right place.

theano/tensor/nnet/corr.py
```diff
@@ -109,6 +116,17 @@ def c_support_code_apply(self, node, nodename):
             sub['float_typenum'] = 'NPY_DOUBLE'
             sub['n_bytes'] = 8
             sub['c_float_type'] = 'double'
+        if self.openmp:
+            sub['cores'] = self.cores
+            sub['omp_flags'] = '#pragma omp parallel for'
```
nouiz: why not force the static schedule?

theano/tensor/nnet/corr.py
```diff
@@ -312,7 +330,7 @@ def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width
         else {
           typenum = PyArray_TYPE(bottom);
         }
-        %(out)s = (PyArrayObject*)PyArray_EMPTY(4,
+        %(out)s = (PyArrayObject*)PyArray_ZEROS(4,
```
nouiz: I think we can still use EMPTY when there is only 1 thread. So it would be good to not slow that that case. Also, I think it need to be zero only for some of the direction. So try to minimize when this is set to zero. We could let it empty here and we set it to zero only when needed in the code bellow?

nouiz:
Thanks for the PR. Overall, it seem ok, I got a few small comments. As @JesseLivezey told, we need to make sure this PR work in the case the others didn't.

theano/tensor/nnet/corr_gemm.c
JesseLivezey: Agreed that this should use omp_get_max_threads() rather than just the number of cores.

theano/tensor/nnet/corr.py
pcs-theano: You're right, I will remove them in the 3 subclass.

pcs-theano:
Sure, let me try with gcc 4.9 and 5.x and update the test result.

-----------
Regards,
Ciyong

pcs-theano:
@nouiz @JesseLivezey Thanks for your comments, codes are updated according to your comments, and it works well with both gcc 4.9 and 5.3.

theano/tensor/nnet/corr.py
```diff
@@ -377,8 +374,8 @@ class CorrMM(BaseCorrMM):
         Set to `(1, 1)` to disable subsampling.
 
     """
-    def __init__(self, border_mode="valid", subsample=(1, 1), openmp=None):
-        super(CorrMM, self).__init__(border_mode, subsample, openmp=openmp)
+    def __init__(self, border_mode="valid", subsample=(1, 1)):
```
nouiz: Here, I was meaning to remove completly the \__init__ method. It need to keep the openmp parameter as before this commit. And the code end up being what the default \__init__ is when it is not implemented. So we don't need it.

JesseLivezey:
I can also try and test this on OSX this week.

nouiz:
thanks

theano/tensor/nnet/corr_gemm.c
```diff
@@ -172,50 +171,61 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
-		                           col_dim,
-                                           PyArray_TYPE(top),
-					   0);
+    const int max_threads = %(omp_max_threads)s < batchSize ? %(omp_max_threads)s : batchSize;
+
+    npy_intp col_dim[3];
+    col_dim[0] = (npy_intp)max_threads;
+    col_dim[1] = (npy_intp)(nChannels * kW * kH);
+    col_dim[2] = (npy_intp)(topHeight * topWidth);
+    PyArrayObject* col = (PyArrayObject*)PyArray_ZEROS(3,
```
nouiz: why zeros col? if this is needed only when max_threads > 1, then do it only in that case to don't slow down the single thread case.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -172,50 +171,61 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
     PyArrayObject *output;
+
+    %(omp_set_threads)s(max_threads);
+    
     if (direction == 0) {  // forward pass
         output = top;
         // valid correlation: im2col, then gemm
         // Iterate over batch
-        for (int n = 0; n < batchSize; n++) {
+        %(omp_flags)s
```
nouiz: I understand that this will parallelize the im2col that wasn't parallel. If we use a multi-thread blas, the gemm call is already parallelized. Can you explain how this case will be handled? Would it be better to use the blas paralelism for the blas call and not the parallelism here?

I guess it is better to use the parallelism here instead of in the blas library, but this is a guess, if you have a good reason why one is better, I would be interested to know.

JesseLivezey: @nouiz, I did some timing in #3689 which I'll copy here. Basically, I found that it was best to use 1 core for BLAS and parallelize the batch across cores. I was working with fairly large batches (128) on a 12 core machine.

```
Timing done on 12-core Intel Xeon X5650 with 24GB RAM.
BLAS was MKL from Anaconda Python.

Image shape, filter shape
(128, 3, 128, 128), (96, 3, 5, 5)
Imagenet-like
MKL     OpenMP  Speedup  Notes
1       12      1        Legacy, OpenMP, baseline
1       1       .097     Legacy, single thread
1       1       .75      CorrMM, MKL only
2       1       1.3      “
4       1       1.9      “
8       1       2.0      “
12      1       1.7      “
1       2       1.5      CorrMM, MKL single thread + OpenMP
1       4       2.7      “
1       8       4.5      “
1       12      5.3      “

Image shape, filter shape
(128, 85, 1, 258), (64, 85, 1, 20)
Spectrogram-like
MKL     OpenMP  Speedup  Notes
1       12      1        Legacy, OpenMP, baseline
1       1       .095     Legacy, single thread
1       1       .50      CorrMM, MKL only
2       1       .78      “
4       1       1.2      “
8       1       1.5      “
12      1       1.3      “
1       2       .94      CorrMM, MKL single thread + OpenMP
1       4       2.0      “
1       8       4.0      “
1       12      5.2      “
```

nouiz: Thanks. This confirm what is better. Can you make a last timming? What if
you only define the env variable OMP_NUM_THREADS and don't set
MKL_NUM_THREADS? I suppose you will have the best timming, but a
confirmation would be great.

JesseLivezey: Sure, I'll redo timing with the new PR and separate out the forwards and two backwards convolutions.

I'll post the timing script here since it might be useful to add for other people to use.

theano/tensor/nnet/corr.py
pcs-theano: Sorry for the late response. That makes sense, will update the codes as you suggested.

theano/tensor/nnet/corr_gemm.c
pcs-theano: PyArray_ZEROS is faster than PyArray_EMPTY as I tested in the conditinon of max_threads > 1, but I didn't test their performance with single thread case. If you're referring to PyArray_ZEROS could slow down performance than PyArray_EMPTY with single thread case, then we can distinguish those two conditions and select the faster one.

theano/tensor/nnet/corr_gemm.c
pcs-theano: Thanks @JesseLivezey for your timing data to illustrate the configuration for best performance. I was just using Theano's profiling function to compared the performance result with different XXX_NUM_THEADS, and it's really great to have such  timing script for profiling.

theano/tensor/nnet/corr_gemm.c
nouiz: I don't understand why ZEROS was faster then EMPTY. Do you remember how much difference it was doing? If it is small, we could just let it like this for single thread.

theano/tensor/nnet/corr.py
nouiz: I didn't see any new commits. Did you forgot to push?

JesseLivezey:
This is the script I've been using:
https://gist.github.com/JesseLivezey/12e0e320960a58c278138d402f724fca
to get these timings

Note that the ```baseline``` op is different than in the previous comparison. I've taken out the legacy conv op.

```
Timing done on 12-core Intel Xeon X5650 with 24GB RAM.
BLAS was MKL from Anaconda Python, gcc 4.8.5.

Op: forward
Image shape, filter shape
(128, 3, 128, 128) (96, 3, 5, 5)
Imagenet-like
MKL       OpenMP    Time(ms)  Speedup   Notes     
1         None      1668      0.45      single thread
12        None      752       1.00      BLAS only, baseline
None      12        234       3.21      only OMP set
2         1         1002      0.75      BLAS only 
4         1         680       1.10      BLAS only 
8         1         619       1.21      BLAS only 
1         2         863       0.87      OMP only  
1         4         479       1.57      OMP only  
1         8         277       2.71      OMP only  
1         12        234       3.21      OMP only  


Op: forward
Image shape, filter shape
(128, 85, 2, 258) (64, 85, 2, 20)
Spectrogram-like
MKL       OpenMP    Time(ms)  Speedup   Notes     
1         None      757       0.39      single thread
12        None      294       1.00      BLAS only, baseline
None      12        90        3.25      only OMP set
2         1         511       0.58      BLAS only 
4         1         338       0.87      BLAS only 
8         1         271       1.09      BLAS only 
1         2         397       0.74      OMP only  
1         4         209       1.40      OMP only  
1         8         117       2.50      OMP only  
1         12        87        3.36      OMP only  


Op: gradInputs
Image shape, filter shape
(128, 3, 128, 128) (96, 3, 5, 5)
Imagenet-like
MKL       OpenMP    Time(ms)  Speedup   Notes     
1         None      3256      0.27      single thread
12        None      883       1.00      BLAS only, baseline
None      12        374       2.36      only OMP set
2         1         2073      0.43      BLAS only 
4         1         1301      0.68      BLAS only 
8         1         1004      0.88      BLAS only 
1         2         1681      0.53      OMP only  
1         4         888       0.99      OMP only  
1         8         487       1.81      OMP only  
1         12        375       2.35      OMP only  


Op: gradInputs
Image shape, filter shape
(128, 85, 2, 258) (64, 85, 2, 20)
Spectrogram-like
MKL       OpenMP    Time(ms)  Speedup   Notes     
1         None      1619      0.30      single thread
12        None      491       1.00      BLAS only, baseline
None      12        217       2.26      only OMP set
2         1         1050      0.47      BLAS only 
4         1         699       0.70      BLAS only 
8         1         545       0.90      BLAS only 
1         2         834       0.59      OMP only  
1         4         440       1.11      OMP only  
1         8         271       1.81      OMP only  
1         12        214       2.29      OMP only  


Op: gradWeights
Image shape, filter shape
(128, 3, 128, 128) (96, 3, 5, 5)
Imagenet-like
MKL       OpenMP    Time(ms)  Speedup   Notes     
1         None      1661      0.45      single thread
12        None      744       1.00      BLAS only, baseline
None      12        235       3.16      only OMP set
2         1         1001      0.74      BLAS only 
4         1         667       1.11      BLAS only 
8         1         621       1.20      BLAS only 
1         2         856       0.87      OMP only  
1         4         465       1.60      OMP only  
1         8         278       2.67      OMP only  
1         12        235       3.16      OMP only  


Op: gradWeights
Image shape, filter shape
(128, 85, 2, 258) (64, 85, 2, 20)
Spectrogram-like
MKL       OpenMP    Time(ms)  Speedup   Notes     
1         None      749       0.41      single thread
12        None      307       1.00      BLAS only, baseline
None      12        89        3.44      only OMP set
2         1         512       0.60      BLAS only 
4         1         336       0.91      BLAS only 
8         1         270       1.14      BLAS only 
1         2         416       0.74      OMP only  
1         4         211       1.46      OMP only  
1         8         127       2.40      OMP only  
1         12        88        3.48      OMP only
```

JesseLivezey:
@pcs-theano, the machine I tested this on is pretty old. If you have a more modern server with lots of cores, I'd be interested to see the timings. You might have to mess around with the script a bit, but hopefully it is fairly clear.

theano/tensor/nnet/corr.py
pcs-theano: @nouiz I will push the code today.

theano/tensor/nnet/corr_gemm.c
pcs-theano: I remembered  ZEROS is around 1.7x faster than EMPTY when running AlexNet workload. Anyway, I can retry then gather and post data here.

pcs-theano:
@JesseLivezey ,thanks for your testing data and script, I can run it on my workstation which has two "Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz" sockets (36 cores in total) and 128GB RAM, will update the data when it's done.

theano/tensor/nnet/corr_gemm.c
pcs-theano: @nouiz looks like PyArray_ZEROS could run with multi-thread mode in Numpy, while PyArray_EMPTY can't. So I got the better result with PyArray_ZEROS when `max_threads > 1`.
 
Here's the snippet of profiling of AlexNet Workload when `max_threads>1`:

**Calling PyArray_ZEROS**
```
   19.7%    40.1%      23.662s       1.18e-01s     C      200      15   theano.tensor.nnet.corr.CorrMM
   10.3%    67.1%      12.288s       1.23e-01s     C      100       5   theano.tensor.nnet.corr.CorrMM_gradWeights
    9.5%    76.6%      11.422s       1.43e-01s     C       80       4   theano.tensor.nnet.corr.CorrMM_gradInputs
```

**Calling PyArray_EMPTY**
```
  23.1%    23.1%      30.808s       1.54e-01s     C      200      15   theano.tensor.nnet.corr.CorrMM
  12.2%    68.4%      16.244s       1.62e-01s     C      100       5   theano.tensor.nnet.corr.CorrMM_gradWeights
  10.7%    79.0%      14.218s       1.78e-01s     C       80       4   theano.tensor.nnet.corr.CorrMM_gradInputs
```

theano/tensor/nnet/corr_gemm.c
```diff
@@ -172,17 +172,24 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
-    PyArrayObject* col = (PyArrayObject*)PyArray_ZEROS(3,
-		                                          col_dim,
-                                        PyArray_TYPE(top),
-					                                   0);
-    if (NULL == col)
-    {
+
+    PyArrayObject* col;
+    if (max_threads > 1) {
+        col = (PyArrayObject*)PyArray_ZEROS(3,
```
nouiz: Did you do the timing with single thread to know which one was faster?

nouiz:
What do you mean by "got stuck here"? Was the process still running or it was blocked?

The number seem to tell that we are better just setting MKL to 0 threads and OpenMP to the max. Could you try mixed like MKL 2 and OpenMP 14?

From a user point of view, it is not great to set those 2 env variable separatly. Especially as other operation in the graph could need other case. Could we force mkl to user 1 threads here?

theano/tensor/nnet/corr_gemm.c
nouiz: I just saw this comment. It clearly show that ZEROS is faster, but I don't don't understand why. This won't block the PR, but if you know why, I would be interrested to know.

Did you time this with the single thread? As I don't understand why this is faster, maybe it is faster in single thread too.

pcs-theano: Hi nouiz,

When running in single thread, using PyArray_ZEROS is a little bit slow than using PyArray_EMPTY, this is what I found from the test result, but haven't figured out why these two APIs have such performance difference. I looked into their codes, the only difference between them is PyArray_ZEROS could use multi-thread (when enabled) to zero the memory.

Here's the snippet of profiling of AlexNet workload when `max_threads=1`
**Calling PyArray_ZEROS**
```
   45.0%    45.0%     321.487s       1.61e+00s     C      200      15   theano.tensor.nnet.corr.CorrMM
   20.0%    65.0%     142.646s       1.43e+00s     C      100       5   theano.tensor.nnet.corr.CorrMM_gradWeights
   19.9%    84.9%     141.937s       1.77e+00s     C       80       4   theano.tensor.nnet.corr.CorrMM_gradInputs
```
   
**Calling PyArray_EMPTY**
```
   44.9%    44.9%     319.600s       1.60e+00s     C      200      15   theano.tensor.nnet.corr.CorrMM
   20.0%    64.9%     142.701s       1.43e+00s     C      100       5   theano.tensor.nnet.corr.CorrMM_gradWeights
   19.8%    84.8%     141.364s       1.77e+00s     C       80       4   theano.tensor.nnet.corr.CorrMM_gradInputs
```

pcs-theano:
@nouiz When I ran the script, it's not going to dump the result at the place I marked "got stuck here", the process looked like still running but just not any output on the terminal. I'm checking the script...

It's a good point that we can try other combinations with BLAS thread and OMP thread.  BLAS could still run in parallel when OMP havn't taken all of the available threads. Let's see the result when it's done.

MKL BLAS will first check both MKL_NUM_THREADS  and OMP_NUM_THREADS env, but the former one has higher priority than the latter one, which means MKL_NUM_THREADS could overwrite the thread num when OMP_NUM_THREADS is also being set.

If the result shows that MKL 1 and OpenMP max_threads are the best combination, then we can force mkl to use 1 thread in this place.

JesseLivezey:
Two more concerns:
 * not every BLAS library allows the number of threads to be set dynamically, ATLAS, for instance. I'm not sure how this would behave. Maybe it's not a problem.
 * if the BLAS isn't from MKL (openblas for instance), then is it possible to have the loop use more than 1 thread and BLAS use 1 thread dynamically?

An example of the second point would be: we would want omp-loop parallelism and single thread BLAS in the conv layers of a conv net, but BLAS parallelism in the FC layers.

pcs-theano:
@nouiz @JesseLivezey ,

There's a minor issue with the line #42 of script which should be `if direction == 'gradWeights'` instead of `if direction == 'gradInputs'`. 
After clean up/kill all python process before re-run the script, script hang issue is gone.

As for ATLAS, it doesn't support changing threads dynamically, the maximum number of threads to use is determined at compile time. But I think the new codes would not slow down its performance due to parallelism in for loop has better performance than gemm in corr OP.

As for OpenBLAS, it has same way to changing threads dynamically as MKL via setting	 `OPENBLAS_NUM_THREADS` env parameter or calling API `openblas_set_num_threads()`.

I've collected the test data with both MKL and OpenBLAS, and g++(4.8.4) on a workstation with 36 cores (Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz) and 128 GB RAM. 
Besides of original combinations of `BLAS_NUM_THREADS` and `OMP_NUM_THREADS`, I also added some other combinations (`BLAS_NUM_THREADS x OMP_NUM_THREADS = max_threads`).
Test result also includes test time with calling `PyArray_ZEROS` and `PyArray_EMPTY`.

From the test result, calling `PyArray_ZEROS` gains better performance than `PyArray_EMPTY` during `forward` and `gradInputs` when `OpenMP > 1` in AlexNet-like case, and and didn't slow down with single thread mode.
So, maybe we can change `PyArray_EMPTY` to `PyArray_ZEROS` in corr.py?

`GEMM=1` and `OpenMP=max_threads` gains best and stable performance, while `GEMM=None` and `OpenMP=max_threads` sometimes could led to worse result such as `gradWeight` of ImageNet-like in file**1** and **2**, and `gradInputs` of  Spectrogram-like in file**3** and file **4**.
So, it's better to force 1 for GEMM here. Should we only support MKL and OpenBLAS API here?

Hoping these test results should make things clear.

Please check the log files as attachment:
[mkl_g++_pyarray_empty.txt](https://github.com/Theano/Theano/files/317954/mkl_g._pyarray_empty.txt)
[mkl_g++_pyarray_zeros.txt](https://github.com/Theano/Theano/files/317955/mkl_g._pyarray_zeros.txt)
[openblas_g++_pyarray_empty.txt](https://github.com/Theano/Theano/files/317956/openblas_g._pyarray_empty.txt)
[openblas_g++_pyarray_zeros.txt](https://github.com/Theano/Theano/files/317957/openblas_g._pyarray_zeros.txt)

nouiz:
Forcing gemm=1 for mkl and openblas would be great.

There is a few corner case where using MKL=2 OMP=18 would be a little bit
faster with mkl (~10%, can be much slower with OpenBLAS in the same case),
but I think it is simpler and garantie to not have the big slow down to
force GEMM=1. Probably that ~10% depend of the shape and number of core
available, so I would go as you to just ignore it.

Agreed to always use ZEROS, just make a comment that it was timed to be
faster. So later people won't try to revert that back to empty.

pcs-theano:
@nouiz, Codes have been updated to forcing gemm=1 and using PyArray_ZEROS.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -224,6 +217,8 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
                     bottomWidth, kH, kW, padH, padW, dH, dW,
                     (%(float_type)s*)PyArray_DATA(col)+ tid * col_stride);
             // Second, gemm
+            // Always forcing gemm to one thread here for best and stable performance.
+            %(blas_flags)s;
```
nouiz: How to restore to the previous value after the call to gemm? This is to allow BLAS to be parallel for other layers like dense layer that do just 1 big gemm call.

Also, I would call that before the for loop. Idem for the 3 other places.

nouiz:
Travis have some errors. It don't find the file mkl.h. Do you know how to find it automatically? The update to our code that find the good flags to pass (I suppose a -IPATH is missing) should be in the function default_blas_ldflags() in the file theano/configdefaults().

nouiz:
In fact, when using python from anaconda, we don't have access to an mkl.h file. So instead of requesting this file, can you just declare the function signature so that it know how to compile and we should link correctly?

Should we do the same for openblas?

theano/tensor/nnet/corr_gemm.c
pcs-theano: Good point, it should restore to the previous thread number after the call to gemm as you mentioned. Looks like OpenBLAS has different way as MKL did to to so, let me verify them and update later.

pcs-theano:
@nouiz 
If user want to call MKL or OpenBLAS gemm function, they should have mkl.h or cblas.h header file along with the library in their system. 

Regarding Travis build, it has such BLAS header file (mkl.h and cblas.h), right? Are you referring to adding the flag in theano/configdefaults() to let it find the header file automatically?

Regarding the scenario of using python frm anaconda, is it able to access any other header files, or it only just can't access to such BLAS header file. I'm not familiar with anaconda, but I think maybe it's not a good way to poring the function declaration from BLAS header file to corr file, if so, then we may also need to port other function declaration which might be used in future, what do you think?

nouiz:
Theano provide the blas header in the file: 

https://github.com/Theano/Theano/blob/master/theano/tensor/blas_headers.py

This help us be more rebust to different installation setup.

In the case of anaconda, it include the library, but not headers. So there is no mkl.h file. anaconda is a popular way to install python, so we can't just drop that setup.

The only work around I'm thinking is to remove the include of mkl.h and provide the declaration of that extra function ourself.

If you have other idea, tell us.

pcs-theano:
Got it, looks like adding declaration within theano is the only way to make it compatibility with different installation setup, will move forward in this way.

theano/tensor/nnet/corr.py
```diff
@@ -131,20 +132,22 @@ def c_support_code_apply(self, node, nodename):
...
-            sub['omp_get_threads'] = 'omp_get_thread_num()'
+            sub['set_omp_threads'] = 'omp_set_num_threads'
+            sub['get_omp_threads'] = 'omp_get_thread_num()'
+
+            if self.blas_type == 'openblas':
+                sub['set_blas_threads'] = 'openblas_set_num_threads'
+                sub['get_blas_threads'] = 'openblas_get_num_threads()'
+            elif self.blas_type == 'mkl':
+                sub['set_blas_threads'] = 'mkl_set_num_threads'
+                sub['get_blas_threads'] = 'mkl_get_max_threads()'
```
nouiz: Need need an else close here in case another blas is used.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -278,10 +284,13 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
         // valid convolution: im2col, then gemm
         // Iterate over batch
+        int blas_threads_saved = %(get_blas_threads)s;
+        // Always forcing gemm to one thread when OpenMP is enalbed for best and stable performance.
+        %(set_blas_threads)s(1);
         // OMP for batch-level paralization
         %(omp_flags)s
         for (int n = 0; n < batchSize; ++n) {
-        	int tid = %(omp_get_threads)s;
+        	int tid = %(get_omp_threads)s;
```
nouiz: there is a tab on that line. Can you change it to spaces?

theano/tensor/nnet/corr_gemm.c
```diff
@@ -353,12 +363,13 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
+        int blas_threads_saved = %(get_blas_threads)s;
+        // Always forcing gemm to one thread when OpenMP is enalbed for best and stable performance.
+        %(set_blas_threads)s(1);
         %(omp_flags)s
         for (int n = 0; n < batchSize; ++n) {
             // gemm into columns
-        	int tid = %(omp_get_threads)s;
-            // Always forcing gemm to one thread here for best and stable performance.
-            %(blas_flags)s;
+        	int tid = %(get_omp_threads)s;
```
nouiz: there is a tab here too.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -203,22 +206,23 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
     char Trans = 'T';
     PyArrayObject *output;
 
-    %(omp_set_threads)s(max_threads);
+    %(set_omp_threads)s(max_threads);
```
nouiz: I forgot why this line is used if we set that max blas threads. What happen if there is a computer with 20 cores and the batch size of 10 and we don't call this line? My first guess is that the 10 extra cores are just not used.

Line 322: in this commit also use %(omp_flags)s, but do  weight_dim[1] iteration and not batchSize. So if batchsize is 1, this line would like the number if threads for that section.

If it is good to keep this line, we should undo this at the end of this  op to let other node in the graph use there own logic.

theano/tensor/nnet/corr.py
```diff
@@ -131,20 +132,22 @@ def c_support_code_apply(self, node, nodename):
         if self.openmp:
             sub['omp_flags'] = '#pragma omp parallel for schedule(static)'
             sub['omp_max_threads'] = 'omp_get_max_threads()'
-            sub['omp_set_threads'] = 'omp_set_num_threads'
-            sub['omp_get_threads'] = 'omp_get_thread_num()'
+            sub['set_omp_threads'] = 'omp_set_num_threads'
+            sub['get_omp_threads'] = 'omp_get_thread_num()'
```
nouiz: The name change between the real omp fct name and the key make understanding the code a little bit harder. If you can reuse the omp name in the key, it would be good. But this is mostly a nitpik.

nouiz:
It seem pretty good, just one questions and we should be good.

theano/tensor/nnet/corr_gemm.c
pcs-theano: Regarding Line 209, it's added in the first PR, this was added to make sure the parallelism codes running with "max_threads" and the omp_get_thread_num() to return correct omp index. Since this line will force the omp threads to `batch size` when `batch size` is smaller than `max_threads`, which will impact the other region of codes for example Line322 as you pointed out, besides codes without this line (209) are working as expected, so I think it's safe to remove this line. Then we can leave line 322 as it is and no need to restore omp threads at the end of OP.
Regarding the case of the computuer has more cores number than batch size, then the extra cores are not used as you said.

pcs-theano:
@nouiz the codes were updated, any new comments for the latest commit?

nouiz:
All looks good, but a rebase is required. thanks.

Issue #4585, opened 2016-06-04
grad of T.log2 creates float64


apulver:
warn_float64 = warn

`x = T.fscalar()`
`y = T.log2(x)`
`f = theano.function(inputs=[x], outputs=T.grad(y,x))`

nouiz:
Thanks for the report, I have a fix here:

https://github.com/Theano/Theano/pull/4593

PR #4593, opened 2016-06-06
Fix gh-4585. Cast the constant to the right dtype.

nouiz:


abergeron:
Ok for me when travis passes

PR #4598, opened 2016-06-06
Do not hide errors in max and argmax

fvisin:
As found out with @lamblin and @abergeron in the magic night before NIPS's deadline, Theano does not propagate the error message correctly when something fails in max or argmax. This PR fixes it.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -1399,8 +1399,6 @@ class MaxAndArgmax(Op):
         %(axis_code)s
         %(max)s = (PyArrayObject*)PyArray_Max(%(x)s, axis, NULL);
         if(%(max)s == NULL){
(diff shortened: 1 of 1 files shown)


theano/tensor/basic.py
```diff
@@ -1399,8 +1399,6 @@ def c_code(self, node, name, inp, out, sub):
         %(axis_code)s
         %(max)s = (PyArrayObject*)PyArray_Max(%(x)s, axis, NULL);
         if(%(max)s == NULL){
-            PyErr_SetString(PyExc_ValueError,
-                         "MaxAndArgmax, max failed");
```
nouiz: should we do the same on line 1413 for argmax?

fvisin:
Updated and squashed the commits.

Issue #4597, opened 2016-06-06
cannot convert ‘PyGpuContextObject*’ to ‘PyGpuArrayObject*’ with multi gpus on new backend


cmyr:
theano version 0.9.0dev1.dev-0693ce052725a15b502068a1490b0637216feb00
Python 3.5.1

Hello,

I'm currently trying to add Theano multi-gpu support to [Keras](https://github.com/fchollet/keras). I appreciate this is very much in active development and this feature hasn't tested much. If there's anywhere that would be a better place for this please let me know.

I have `libgpuarray` working on a machine with 4 cuda devices. I have them aliased in my .theanorc as per [this](http://deeplearning.net/software/theano_versions/dev/tutorial/using_multi_gpu.html#tut-using-multi-gpu) with the line `contexts = dev0->cuda0;dev1->cuda1;dev2->cuda2;dev3->cuda3`. 

This sample code works:

```python
v01 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
                    target='dev0')
v02 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
                    target='dev0')
v11 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
                    target='dev1')
v12 = theano.shared(numpy.random.random((1024, 1024)).astype('float32'),
                    target='dev1')

f = theano.function([], [theano.tensor.dot(v01, v02),
                         theano.tensor.dot(v11, v12)])
f()
```

For real-world use, I have a little context manager that inserts the correct target into the call to `theano.shared`. My actual (keras) test code right now looks like this (this is converted into a theano graph behind the scenes):

```python
model = Sequential()
with device('$DEV1'):
    model.add(Dense(output_dim=37, input_dim=13))
with device('$DEV2'):
    model.add(Dense(5))
model.compile(loss="mse", optimizer='sgd')
```

This works if $DEV1 and $DEV2 are either the same device or if one of them is 'gpu'. If they are different gpus, however, I get this crash.

**full text of crash:**

```
===============================
===============================
.theano/compiledir_Linux-3.13--generic-x86_64-with-Ubuntu-14.04-trusty-x86_64-3.5.1-64/tmpfhdwkhbp/mod.cpp: In member function ‘int {anonymous}::__struct_compiled_op_mff375c0666634f2fb3014bf1941c639e::run()’:
[...]

PR #4600, opened 2016-06-07
Fix GpuToGpu

nouiz:
fix gh-4597

### DIFF
--- a/theano/gpuarray/tests/test_basic_ops.py
+++ b/theano/gpuarray/tests/test_basic_ops.py
@@ -15,7 +15,7 @@ from theano.tensor.tests import test_basic
 from theano.tensor.tests.test_basic import rand, safe_make_node
 from theano.tests import unittest_tools as utt
 
-from ..type import (GpuArrayType, get_context, list_contexts,
+from ..type import (GpuArrayType, get_context,
                     gpuarray_shared_constructor)
 from ..basic_ops import (
     host_from_gpu, HostFromGpu, GpuFromHost, GpuReshape, GpuToGpu,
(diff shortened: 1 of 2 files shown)


nouiz:
To have the new test pass with FAST_COMPILE and DEBUG_MODE, this other PR is needed: https://github.com/Theano/libgpuarray/pull/187

Also, note that this test don't run by default.

theano/gpuarray/tests/test_basic_ops.py
```diff
@@ -15,10 +15,10 @@
 from theano.tensor.tests.test_basic import rand, safe_make_node
 from theano.tests import unittest_tools as utt
 
-from ..type import (GpuArrayType, get_context,
+from ..type import (GpuArrayType, get_context, list_contexts,
```
abergeron: Where does `list_contexts()` come from?

theano/gpuarray/tests/test_basic_ops.py
```diff
@@ -182,6 +182,24 @@ def test_transfer_cpu_gpu():
     assert numpy.all(fv == av)
 
 
+def test_transfer_gpu_gpu():
+    ctxs = list_contexts()
+    if len(ctxs) <= 1:
+        raise SkipError("Need more then 1 context to test multi-gpu transfer")
```
abergeron: You can still do a GpuToGpu in the same context.  It should do the equivalent of a copy.

theano/gpuarray/tests/test_basic_ops.py
```diff
@@ -182,6 +182,24 @@ def test_transfer_cpu_gpu():
...
+
+    g = GpuArrayType(dtype='float32', broadcastable=(False, False),
+                          context_name=ctxs[0])()
+
+    av = numpy.asarray(rng.rand(5, 4), dtype='float32')
+    gv = gpuarray.array(av, context=get_context(ctxs[0]))
+    gv2 = gpuarray.array(av, context=get_context(ctxs[1]))
+
+    f = theano.function([g], GpuToGpu(ctxs[1])(g))
+    theano.printing.debugprint(f)
```
abergeron: Please don't leave that in the test code.

abergeron:
I'm ok with the code changes to the op, and I have a few questions about the test.

nouiz:
I updated the test

theano/gpuarray/tests/test_basic_ops.py
```diff
@@ -182,6 +182,19 @@ def test_transfer_cpu_gpu():
...
 
 
+def test_transfer_gpu_gpu():
+    g = GpuArrayType(dtype='float32', broadcastable=(False, False),
+                     context_name=test_ctx_name)()
+
+    av = numpy.asarray(rng.rand(5, 4), dtype='float32')
+    gv = gpuarray.array(av, context=get_context(test_ctx_name))
+
+    f = theano.function([g], GpuToGpu(test_ctx_name)(g))
```
abergeron: You will need to disable the `local_cut_gpu_transfers` opt because otherwise it only leaves a DeepCopyOp in the graph.  The op itself will work with the same context as source and target.

theano/gpuarray/tests/test_basic_ops.py
```diff
@@ -182,6 +182,19 @@ def test_transfer_cpu_gpu():
...
+    theano.printing.debugprint(f)
```
abergeron: Please don't leave this in.

nouiz: done

PR #4604, opened 2016-06-08
Change documentation theme to readthedocs and add source links

fvisin:
Change the theme of the documentation!
On top of being much more readable and looking nicer, I added links to the source from the function's documentation, which allow to quickly inspect the code when the documentation is not clear enough or missing.

I crawled around the documentation and everything seems to look fine.


nouiz:
@slefrancois can you test it locally? If it is good, you can merge.

slefrancois:
Looks good, here are some small things I noticed:
-The Theano logo is now blue on blue, we need to either change the logo colors or the background.
-The main text area is narrower, maybe we would want to widen it a bit, is that possible?
-The doc version switcher no longer appears, I can look into that.

fvisin:
Thanks for testing it!

> The Theano logo is now blue on blue, we need to either change the logo colors or the background.

I will look into this.

> The main text area is narrower, maybe we would want to widen it a bit, is that possible?

I think it's possible, but I would rather keep it this way: narrow text is much easier to read because your eyes don't have to move as much. If you don't agree I can look into it too though.

slefrancois:
I'm fine with keeping the width at the theme default, it's quite readable.

fvisin:
I pushed a new logo with inverted colors. I am quite satisfied by the results, but second opinions are welcome!

slefrancois:
Looks good, I'll see if I can figure out the version switcher by tomorrow.

fvisin:
Sounds great, thanks.

slefrancois:
Francesco should push the version switcher shortly.

I also changed the Google Analytics script according to:
https://developers.google.com/analytics/devguides/collection/gajs/
The RTD template started with a comment and the script wouldn't append to the head in chrome.

The version switcher has an old-school arrow in firefox on linux. It's out of place with the flat theme, but styling it right would require rewriting the switcher in javascript/css, which is a job for another day.

fvisin:
The theme itself has a version switcher (see e.g. http://lasagne.readthedocs.org/ at the bottom of the menu bar) but I cannot find a way to enable it, at least locally. Looking at the conf.py and Makefile in Lasagne it doesn't look like they set anything specific. Maybe we modify something in the template that we should not? (I didn't look into it very in detail)

fvisin:
Oh, I just got the notification of your post and I didn't notice it before. I'll update tomorrow morning.

slefrancois:
I looked into it. There is a versions.html file which gets included by the
layout.html template. The versions file has a {if READ_THE_DOCS }, which
checks if the docs is being built on the RTD website. This then pulls a
versions variable which I believe is populated by the user on the RTD web
interface.

It would be possible to write our own versions.html as a replacement, but
it's not part of the document blocks which can be extended in the template
engine. So we might have to use a modified sphinx template, which would
complicate updates and management. Maybe we can

So for now I put our regular switcher in the sidebar. I agree that the RTD
one is nicer, but finding a way to integrate it into the standard template
could take some time.

fvisin:
I just pushed your fix and my last commits. It's ready on my side, but feel free to wait until you complete the rest to merge.

PS: I am sure it's not going to be that easy, but I was wondering if you think we can "trick" the theme either by 1) setting the READ_THE_DOCS variable locally or 2) by removing the `if` in `layout.html`. In the end, we only need the version selector to be displayed on the website and locally *for us*, but if is not shown to users who build the doc locally (if any) is not a big deal. Small caveat: in `conv.py` we load the theme if READ_THE_DOCS is not True, so that too would require a small modification to work with 1).

fvisin:
Thanks @slefrancois for improving the version switcher. I like the new one, much nicer!!

@lamblin can we merge it? (@slefrancois does not have merge rights yet) 
I randomly sampled some pages and I didn't notice any problem.

slefrancois:
Thanks, I had to hammer it a bit with javascript, but it works. Eventually it may be possible to adapt RTD's versions.html and insert it instead of the combobox. But for now let's work on getting this deployed.

fvisin:
It looks great and is very close to what users familiar with the theme are used to, which makes it easy to spot for the users...and that's all we care about in the end! Let's see if everything passes the final review of Pascal :)

fvisin:
@lamblin I checked it locally on master, can you give it a last look when you have time? I also ported it back: #4660

doc/.static/version_switch.js
```diff
@@ -10,9 +9,9 @@
   // If doc is run localy
   if (url.startsWith('file')) {
     theano_dir = 'html';
-    versions_dir = {"local":"html"};
+    versions_dir = {"local":"html", "test":"test"};
```
lamblin: Should "test" here stay in the final version?

lamblin: It's not a big deal anyway.

doc/conf.py
```diff
@@ -165,6 +183,33 @@
...
+
+    if domain != 'py' or not info['module']:
+        return None
+    try:
+        filename = 'theano/%s#L%d-L%d' % find_source()
+    except Exception:
+        filename = info['module'].replace('.', '/') + '.py'
+    from subprocess import check_output
+    tag = check_output('git rev-parse HEAD', shell=True)[:-1]
+    return "https://github.com/Theano/theano/blob/%s/%s" % (tag, filename)
```
lamblin: This does not work correctly with Python 3, I have a link that looks like `https://github.com/Theano/theano/blob/b'b6a7bd28fe014655b358a66f20ae535821dfb439'/theano/compile/function.py#L76-L326` because the commit number gets printed as `b'b6a7bd28fe014655b358a66f20ae535821dfb439'`

fvisin: I pushed a fix, should work now.

doc/.static/version_switch.js
slefrancois: It only appears when the doc is run locally (url startsWith file). I use it
to make sure the version list is formatted properly when testing locally.

lamblin:
Once online, here are the issues I've seen:
- A stray green "v:" below LICENSE in the left bar
- When the bottom-left menu is expanded, the PDF link points to the released version, not the current one
- It displays "v:theano_versions/dev", where I would have expected only "dev", but maybe that is on purpose

None of these are a big deal, but if we could correct them before backporting to 0.8.X, it would be great

doc/.static/version_switch.js
```diff
@@ -28,61 +27,57 @@
     versions_dir[ver] = current_version
```
fvisin: @slefrancois I think we should add this line here:
```
current_version = ver
```
This would change `current_version` from `"theano_versions/dev"` to `"dev"`, fixing the name in the version switcher in [L36](https://github.com/Theano/Theano/pull/4604/files#diff-73310b015c0ae02b0ab8561acb3a2a59R36).

doc/.templates/layout.html
```diff
@@ -1,34 +1,45 @@
...
-    ga.setAttribute('async', 'true');
-    document.documentElement.firstChild.appendChild(ga);
-  })();
+  var span = document.createElement("SPAN")
+  var span_class = document.createAttribute("class");
+  span_class.value = "version_switcher_placeholder";
+  span.setAttributeNode(span_class);
+    
+  var div = document.createElement("DIV");
+  var text = document.createTextNode("v:");
```
fvisin: @slefrancois This, along with [L35](https://github.com/Theano/Theano/pull/4604/files#diff-6f2b65c406aa07382144f10909a074f9R35), seems to be the source of the extra `v:` instead.

fvisin:
@lamblin Thank you for the review and for checking the result once online!
@slefrancois please let me know if you think my fixes make sense (no rush!). I didn't dig into it enough to be confident and push a new version right away, and not being able to preview the changes is quite annoying!

slefrancois:
I can look into fixing this on my side. I have access to the web server and can manually push changes there.

PR #4605, opened 2016-06-08
Removed repetition

Sentient07:
There was a repetition there and I removed it. I don't think this is being covered in any of the on-going documentation work.


nouiz:
ok to merge when travis pass.

PR #4606, opened 2016-06-08
Fix crash

nouiz:
The DTYPE param wasn't increfed. As it was complicated to change that with the current code sctructure and there was no confirmation that it helped, I revert it.


abergeron:
This is ok for me.

PR #4607, opened 2016-06-08
pool.py optimization with openmp

FayW:
Hi reviewers,

    This optimization collapse the top 2 loops (batch and kernel) to 1 by hand (which can not  be 

done by omp collapse since it needs perfect nesting) to achieve a better acceleration ratio on multi-

core systems since only batch level parallelization seems less likely to feeds up the CPUs.

    The speeding up depends on the workload (b*k), compilers and libraries. For g++ plus openblas, 

running test case 'test_cnn' in DeepLearning Tutorials achieves a 2x speed up while for icpc plus mkl 

the speed up is 10x.

   The combination of compilers and libraries of g++ 4.8, 4.9, 5.x and openblas, mkl all tested.

    Hope this PR helps though it's quite simple.

    Regards.


theano/tensor/signal/pool.py
```diff
@@ -708,6 +717,10 @@ def c_code(self, node, name, inp, out, sub):
         ds0, ds1 = self.ds
         st0, st1 = self.st
         pd0, pd1 = self.padding
+        if self.openmp:
```
nouiz: You must modify the \__init__ of this class. You changed PoolGrad.\__init__, but MaxPoolGrad.\__init__ must also be changed

theano/tensor/signal/pool.py
```diff
@@ -443,13 +450,15 @@ def c_code(self, node, name, inp, out, sub):
...
         if (z_r && z_c)
         {
-            for(int b=0; b<PyArray_DIMS(%(x)s)[0]; b++){
-              for(int k=0; k<PyArray_DIMS(%(x)s)[1]; k++){
-                for(int i=0; i< z_r; i++){
+            int r_st, r_end, c_st, c_end;
+            %(omp_parallel)s
+            for(int t = 0; t < PyArray_DIMS(%(x)s)[0] * PyArray_DIMS(%(x)s)[1]; t++){
+                int b = t / PyArray_DIMS(%(x)s)[1];
+                int k = t %% PyArray_DIMS(%(x)s)[1];
```
nouiz: I would swap the / and %%. Currently, if a thread need to do the iteration i and i+1, it will read memory that is less contiguous in memory then if we swap them. As the schedule is static, this should happen frequently. So I think it would be more cache friendly to swap the b and k here.

nouiz:
I have 2 smalls comments. Otherwise, it seem good.

thanks

FayW:
Hi Nouiz,

    According to your comments, the sequence in omp changes to batch first kernel second achieve

better data locality. And the init of MaxPoolGrad has changed to be complied with PoolGrad.

    Combination of g++ (4.8, 4.9, 5), openblas, icpc, mkl all tested ok.

    Thanks.

theano/tensor/signal/pool.py
```diff
@@ -645,7 +645,7 @@ def infer_shape(self, node, in_shapes):
 
 class MaxPoolGrad(PoolGrad):
     def __init__(self, ds, ignore_border, st=None, padding=(0, 0)):
```
nouiz: I would add the openmp parameter to this method to allow the user to change it as it is for the other openmp op.

nouiz: after this, I think the PR will be ready to merge. thanks.

FayW:
Rever the wrong default setting 'openmp = true' to 'openmp=None'.

theano/tensor/signal/pool.py
```diff
@@ -629,14 +637,15 @@ def __init__(self, ds, ignore_border, st=None, padding=(0, 0), mode='max'):
...
 
     def infer_shape(self, node, in_shapes):
         return [in_shapes[0]]
 
 
 class MaxPoolGrad(PoolGrad):
-    def __init__(self, ds, ignore_border, st=None, padding=(0, 0)):
-        PoolGrad.__init__(self, ds, ignore_border, st, padding, mode='max')
+    def __init__(self, ds, ignore_border, st=None, padding=(0, 0), mode='max', openmp=None):
+        PoolGrad.__init__(self, ds, ignore_border, st, padding, mode, openmp)
```
nouiz: Use those 2 lines:

     def __init__(self, ds, ignore_border, st=None, padding=(0, 0), openmp=None):
        PoolGrad.__init__(self, ds, ignore_border, st, padding, mode, mode='max', openmp)

FayW: Hi nouiz,

 def __init__(self, ds, ignore_border, st=None, padding=(0, 0), openmp=None):
    PoolGrad.__init__(self, ds, ignore_border, st, padding, mode='max', openmp)

    will trigger error

    " SyntaxError: non-keyword arg after keyword arg " (both in python2 and
python3)

    Try

    PoolGrad.__init__(self, ds, ignore_border, st, padding, openmp,
mode='max')

trigger

" TypeError: __init__() got multiple values for keyword argument 'mode' "


    Change the init of Pool_Grad to

 def __init__(self, ds, ignore_border, st=None, padding=(0, 0),
openmp=None, mode='max'):

 may pass the compilation , however since the sequence of parameter changes
the caller also has to change the sequence of arguments.


    I think the easiest workaround is by using mode  as the keyword
argument(mode='max') of the init and calling PoolGrad init with '... mode,
openmp'  and the current code that

calling PoolGrad need no changes. And the the caller of MaxPoolGrad also
need no changes since mode, openmp are appended.

2016-06-10 23:51 GMT+08:00 Frédéric Bastien <notifications@github.com>:

nouiz: We don't want to make the mode available in MaxPoolGrad, as it most be max. This should work:
```
 def __init__(self, ds, ignore_border, st=None, padding=(0, 0), openmp=None):
    PoolGrad.__init__(self, ds, ignore_border, st, padding, mode='max', openmp=openmp)
```

I removed the double argument mode and I made openmp a named argument.

```
 def __init__(self, ds, ignore_border, st=None, padding=(0, 0), openmp=None):
    PoolGrad.__init__(self, ds, ignore_border, st, padding, 'max', openmp)
```
should also work.

FayW:
You are right about that. 

init of MaxPoolGrad changed as your suggestion.

Tested OK.

Thanks.

nouiz:

FayW:
Hi nouiz , something wrong with the master branch? the last commit of 10th June, (00cc0fb in this conversation) i.e.,

    def __init__(self, ds, ignore_border, st=None, padding=(0, 0), mode='max', openmp=None):
        PoolGrad.__init__(self, ds, ignore_border, st, padding, mode, openmp)

 passed while the third commit of today 675fe64 failed however they are same. 

currently 
    def __init__(self, ds, ignore_border, st=None, padding=(0, 0), openmp=None):
         PoolGrad.__init__(self, ds, ignore_border, st, padding, 'max', openmp)
and
     def __init__(self, ds, ignore_border, st=None, padding=(0, 0), openmp=None):
         PoolGrad.__init__(self, ds, ignore_border, st, padding, mode='max', openmp=openmp)
travis build all failed
"
test_grad_negative_axis_2d (theano.tensor.tests.test_sort.test_sort) ... 
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
"

nouiz:
Yes, there was a problem with master. I have a PR that should fix it. I'll wait for it before merging this PR, but this PR seem good. thanks.

PR #4608, opened 2016-06-09
Proofread README.txt

mockingjamie:
Made phrasing, layout and grammatical improvements to README.txt to make it more understandable and coherent.


nouiz:
thanks

PR #4611, opened 2016-06-09
Crash Fix

Sentient07:
Converting `float` type to `int` if their values are same for the shape element in `ShapeFeature`
@nouiz There is another crash fix(which we discussed), using fast_compile with AbstractConv2d. Shall I add that commit in this PR or shall I leave it to be merged with my GPUGraph PR ?

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -883,9 +883,8 @@ def local_gpua_softmaxwithbias(node, context_name):
 def local_assert(node, context_name):
 
     # Check if input nodes are already on the GPU
-    for i in node.inputs:
-        if isinstance(i.type, GpuArrayType):
-            return
+    if isinstance(node.inputs[0].type, GpuArrayType):
+        return
     return [host_from_gpu(node.op(as_gpuarray_variable(node.inputs[0],
                                                        context_name),
                                   *node.inputs[1:]))]
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -881,6 +881,11 @@ def local_gpua_softmaxwithbias(node, context_name):
 @register_opt('fast_compile')
 @op_lifter([theano.tensor.opt.Assert])
 def local_assert(node, context_name):
+
+    # Check if input nodes are already on the GPU
+    for i in node.inputs:
+        if isinstance(i.type, GpuArrayType):
+            return
     return [host_from_gpu(node.op(as_gpuarray_variable(node.inputs[0],
                                                        context_name),
                                   *node.inputs[1:]))]
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -1028,6 +1028,8 @@ class ShapeFeature(object):
             # don't make the optimizer merge a zillion ones together
             # by always returning the same object to represent 1
             return self.lscalar_one
+        if type(s_i) is float and int(s_i) == s_i:
+            s_i = int(s_i)
         if (type(s_i) in integer_types or
                 isinstance(s_i, numpy.integer) or
                 (isinstance(s_i, numpy.ndarray) and s_i.ndim == 0)):


Sentient07:
The crash fix discussed in the previous comment has been added to this PR.

theano/gpuarray/elemwise.py
```diff
@@ -75,10 +75,11 @@ def make_node(self, *inputs):
             pass
         try:
             support_code = self.scalar_op.c_support_code()
-            if (support_code.strip() != "#define THEANO_MACRO_MOD(x,y) (x % y)" and
-                    support_code.strip() != ""):
+            if "struct" in support_code:
```
abergeron: The thing is that a struct is ok if not really a C++ class.  But I guess this is ok for now.

abergeron:
I'm ok with this when travis passes.

Sentient07:
I guess this still hasn't solved the problem yet. I don't see profiling results when i use `fast_compile` and i get this error when i don't use `fast_compile`. 

```
<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.TypeError'> ('The type of the replacement must be compatible with the type of the original Variable.', Assert{msg='Theano Assert failed!'}.0, HostFromGpu(gpuarray).0, GpuArrayType<None>(int8, (False, False)), TensorType(int8, matrix), 'local_assert') local_assert
ERROR (theano.gof.opt): Optimization failure due to: local_assert
ERROR (theano.gof.opt): node: Assert{msg='Theano Assert failed!'}(GpuFromHost<None>.0, HostFromGpu(gpuarray).0)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 1853, in process_node
    remove=remove)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/toolbox.py", line 363, in replace_all_validate_remove
    chk = fgraph.replace_all_validate(replacements, reason)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/toolbox.py", line 312, in replace_all_validate
    fgraph.replace(r, new_r, reason=reason, verbose=False)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/fg.py", line 506, in replace
    str(reason))
TypeError: ('The type of the replacement must be compatible with the type of the original Variable.', Assert{msg='Theano Assert failed!'}.0, HostFromGpu(gpuarray).0, GpuArrayType<None>(int8, (False, False)), TensorType(int8, matrix), 'local_assert')
```
Meanwhile, i'll look into travis 👍

Sentient07:
I've fixed the error that I reported in the previous comment. But, I am not able to produce the travis error on my local machine.

theano/gpuarray/opt.py
```diff
@@ -881,6 +881,11 @@ def local_gpua_softmaxwithbias(node, context_name):
 @register_opt('fast_compile')
 @op_lifter([theano.tensor.opt.Assert])
 def local_assert(node, context_name):
+
+    # Check if input nodes are already on the GPU
+    for i in node.inputs:
+        if isinstance(i.type, GpuArrayType):
+            return
```
nouiz: Here we only move the 1st input to the GPU, so the code don't need a for loop:

if isinstance(node.inputs[0].type, GpuArrayType):
   return

Sentient07:
Changed that. I think it is fine now.

nouiz:
close/merge for travis

theano/gpuarray/opt.py
```diff
@@ -883,9 +883,8 @@ def local_gpua_softmaxwithbias(node, context_name):
 def local_assert(node, context_name):
 
     # Check if input nodes are already on the GPU
-    for i in node.inputs:
-        if isinstance(i.type, GpuArrayType):
-            return
+    if isinstance(node.inputs[0].type, GpuArrayType):
+        return
     return [host_from_gpu(node.op(as_gpuarray_variable(node.inputs[0],
```
nouiz: I would remove the host_from_gpu here, this will cause problem with the new opt. You can do that in your other PR to allow to merge this one faster in the master.

Sentient07: I tried doing that earlier and the test passed. I will add it here itself ? Because after this merge, that branch requires a rebase

nouiz: I would do that only in your GraphToGPU PR, not here.

Issue #4562, opened 2016-05-30
Advanced replace subtensor with dimshuffle


fvisin:
This ticket would be an extension of #4356 .

Consider a tensor `tt` with shape `(10, 1, 8, 3, 1)` and broadcastable pattern `(False, True, False, False, False)`. This is the current status of the optimization with dimshuffle:
```python
# These are optimized by #4356
tt[:, 0, :, :, :]
tt[:, -1, :, :, :]
# These are not optimized, but could be in the same fashion as #4356 combined with a 
# subtensor, i.e. these could all become tt.Dimshuffle(0, 2, 3, 4)[something]
tt[0:8, 0, :, :, :]
tt[1, 0, :, :, :]
tt[:, 0, 1, :, :]
tt[:, 0, :, :, 0]
tt[:, :, :, :, 0]
# This is not optimized, but could be optimized in an unrelated optimization
# (not the scope of this ticket)
tt[0:10, :, :, :, :]
```

fvisin:
@SinaHonari here is the issue with the example I talked you about. Your CCW ticket consists in dealing with the second block of cases, i.e. ignore the last case. Let me know if anything is not clear or you have doubts.

lamblin:
It turns out that replacing one `subtensor` node by a chain of `subtensor(dimshuffle)` causes problems for other optimizations.
In particular, it seems it would interfere with the scan `save_mem` optimization, that detects `scan()[...]` to check if only the last timesteps are actually needed.

So, it seems to be a bad idea for the moment, closing.

PR #4612, opened 2016-06-09
Advanced replace subtensor with dimshuffle, resolves issue 4562

SinaHonari:
first commit for issue #4562


lamblin:
The ticket itself looks like a bad idea now, as evidenced by the tests failing in this PR.
The PR was good work, but merging it would be an issue.

PR #4613, opened 2016-06-10
Update Docker instructions

Kaixhin:
The CUDA Docker images are moving to use the NVIDIA Docker program, so I've updated the docs appropriately.

### DIFF
--- a/doc/install.txt
+++ b/doc/install.txt
@@ -117,21 +117,24 @@ A contributor made rpm package for Mandriva_ 2010.2 of Theano 0.3.1.
 
 .. _linux_basic:
 
-Docker images
-~~~~~~~~~~~~~
+Docker
+~~~~~~
 
-Builds of Theano are available as `Docker <https://www.docker.com/whatisdocker>`_ images:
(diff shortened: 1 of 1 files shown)


nouiz:
The PR seem good. I'll wait for travis to finish before merging.

I made recently docker image with Theano that start an ipython notebook: http://github.com/nouiz/Theano-Docker

Just in case you like the idea.

Kaixhin:
Nice stuff. I'm keeping my collection of Dockerfiles as "minimum viable builds", so they can either be used in "production" or as a base image for more complex images. If you do see anything that might need changing with the [CPU](https://github.com/Kaixhin/dockerfiles/blob/master/theano/Dockerfile) or [CUDA](https://github.com/Kaixhin/dockerfiles/blob/master/cuda-theano/cuda_v7.5/Dockerfile) Dockerfiles then just let me know.

It'll take a while for the Docker hub to process all the updated images but I will start testing as soon as I can. Instructions will remain the same though.

nouiz:
thanks

Issue #4610, opened 2016-06-09
Error while trying to profile


Sentient07:
I get this error occasionally when i try profiling(at the end of profiling result). I am not able to produce this consistently. 

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 101, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1250, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 291, in print_profile
    nb_nodes, callbacks_time) = prof
ValueError: need more than 9 values to unpack
```

nouiz:
I included the fix in this PR: https://github.com/Theano/Theano/pull/4614

PR #4614, opened 2016-06-10
Split test to help work around travis timeout. They are super fast here.

nouiz:
fix https://github.com/Theano/Theano/issues/4610

### DIFF
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -345,8 +345,8 @@ class SeqOptimizer(Optimizer, list):
         Merge 2 profiles returned by this cass apply() fct.
 
         """
-        new_t = []
-        new_l = []
+        new_t = []  # the time for the optimization
+        new_l = []  # the optimization
         new_sub_profile = []
         # merge common(same object) opt
         for l in set(prof1[0]).intersection(set(prof2[0])):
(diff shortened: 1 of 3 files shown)


abergeron:
Ok when travis passes.

nouiz:
I added other fix in this PR to help travis. I canceled the old build.

theano/gof/opt.py
```diff
@@ -315,17 +315,17 @@ def print_profile(stream, prof, level=0):
...
+        for (opt, nb_n) in zip(opts, nb_nodes):
             if hasattr(opt, "__name__"):
                 name = opt.__name__
             else:
                 name = opt.name
             idx = opts.index(opt)
             ll.append((name, opt.__class__.__name__,
-                       idx))
-        lll = sorted(zip(prof, ll, nb_nodes), key=lambda a: a[0])
+                       idx) + nb_n)
```
abergeron: Is nb_n a tuple here?  Otherwise this might not work.

abergeron:
Ok modulo my comment.

theano/gof/opt.py
nouiz: yes, it is a tuple.

nouiz:
travis failed due to timeout in FAST_COMPILE in the tests that I split here. I don't know what is going on, locally, it take just a few seconds... I'll merge as the master will be better with this PR merged.

Issue #4580, opened 2016-06-02
Replace some Reshape(Dimshuffle) by Reshape


lamblin:
Sub-ticket of #2943.
When a graph contains `Reshape(Dimshuffle(a))`, and that the `Dimshuffle` only adds, remove, or moves broadcastable dimensions, while keeping non-broadcastable dimensions in the same order, then the `Dimshuffle` can be removed completely, since it does not have any influence on the value of the final `Reshape`.

Examples would be:
- `reshape(dimshuffle{'x', 0}(vec), shape) => reshape(vec, shape)`
- `reshape(dimshuffle{'x', 0, 'x', 1}(mat)) => reshape(mat, shape)` 
- `reshape(dimshuffle{0}(col), shape) => reshape(col, shape)`
- `reshape(dimshuffle{1, 'x'}(row), shape) => reshape(row, shape)`

PR #4615, opened 2016-06-10
useless dimshuffle in reshape is removed

mohammadpz:
Regarding issue #4580 .

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -570,6 +570,25 @@ def local_dimshuffle_lift(node):
 
     """
     op = node.op
+    if (isinstance(op, T.Reshape) and
+            node.inputs[0].owner is not None and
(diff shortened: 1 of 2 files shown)


lamblin:
The issue seems to be that Travis times out in one test: `test_grad_negative_axis (theano.tensor.tests.test_sort.test_sort)`, when run in `FAST_COMPILE`.
It could indicate an infinite loop, or that something is taking much longer than before.
Can you run that test locally with mode=FAST_COMPILE, and see how long it takes with that PR vs. the master?

nouiz:
The problem was in the faster of Theano. It was fixed. I will close/reopen
to retrigget travis.

lamblin:
Indeed, Travis seems to pass now. I'll have another look at the PR tomorrow.

lamblin:
It looks correct, thanks!

PR #4128, opened 2016-02-26
[ENH] faster opt by changing call to extract_constant and get_scalar_constant_value

caglar:
Started adding the only_process_constants=True argument to all extract_constant() and get_scalar_constant_value() calls.

fix gh-4004

PR #4617, opened 2016-06-11
[FIX TRAVIS] Make the new node on a graph not depend on the old node.

nouiz:
This fix the master of travis.

the problem was triggered by gh-4128. But it was already there before. It just that now we don't clean up the problem in fast_compile.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3767,7 +3767,6 @@ def local_useless_switch(node):
                     else:
                         shps.append(node.inputs[2].shape[idx])
                 out = T.alloc(out, *shps)
-                return False
             else:
                 out = out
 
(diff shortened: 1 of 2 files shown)


abergeron:
Ok for me.

theano/tensor/opt.py
```diff
@@ -3753,8 +3753,21 @@ def local_useless_switch(node):
...
+                for idx, (b1, b2), in enumerate(zip(out.type.broadcastable,
+                                                    node.outputs[0].type.broadcastable)):
+                    if b1 == b2:
+                        shps.append(out.shape[idx])
+                    elif not node.inputs[1].type.broadcastable[idx]:
+                        shps.append(node.inputs[1].shape[idx])
+                    else:
+                        shps.append(node.inputs[2].shape[idx])
+                out = T.alloc(out, *shps)
+                return False
```
lamblin: I think that `return False` is what is currently making `test_local_useless_switch` fail.

nouiz: I ran locally those tests with this change and this fix it. I updated this PR.

abergeron:
Again, I'm ok when travis passes.

PR #4618, opened 2016-06-12
Support .theanorc files in the current working directory

tillahoffmann:
Load `.theanorc` configuration file in the current working directory similar to [matplotlib](http://matplotlib.org/users/customizing.html#the-matplotlibrc-file).

### DIFF
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -87,7 +87,9 @@ Environment Variables
     with later (right-most) files taking priority over earlier files in the
     case that multiple files specify values for a common configuration option.
     For example, to override system-wide settings with personal ones,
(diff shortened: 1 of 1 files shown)


nouiz:
@abergeron @lamblin what do you think?

If we accept that, it should be documented in the file doc/library/config.txt

lamblin:
I'm not sure it is a good idea to have it on by default.
We could explicitly document that, if it is the desired behaviour, it is possible to set `THEANORC=~/.theanorc:.theanorc`

nouiz:
I like that solution. @tillahoffmann do you want to update the PR for this?

thanks for bringing this to our attention.

tillahoffmann:
Added some text to the documentation describing how to load local configuration files.

nouiz:
thanks

PR #4627, opened 2016-06-14
fix numpy 1.11 deprecationwarnings

bscellier:


theano/sparse/tests/test_basic.py
```diff
@@ -484,7 +484,7 @@ def test_adv_sub1_sparse_grad(self):
 
         # Test the sparse grad
         valm = numpy.random.rand(5, 4).astype(config.floatX)
-        valv = numpy.random.random_integers(0, 4, 10)
+        valv = numpy.random.randint(0, 4, 10)
```
lamblin: Since `randint` *excludes* the higher bound, but `random_integers` included it, I think the proper translation should be `numpy.randint(0, 5, 10)`.
It may not be a big deal, but I would prefer if we kept the same values.

lamblin: I just checked with an older version of numpy, and the same numbers are indeed generated with `random_integers(0, 4)` and `randint(0, 5)`, if the seed is the same.

theano/tensor/signal/tests/test_pool.py
```diff
@@ -35,8 +35,8 @@ def numpy_max_pool_2d(input, ds, ignore_border=False, mode='max'):
             if input.shape[-1] % ds[1]:
                 yi += 1
         out_shp = list(input.shape[:-2])
-        out_shp.append(input.shape[-2] / ds[0] + xi)
-        out_shp.append(input.shape[-1] / ds[1] + yi)
+        out_shp.append(int(input.shape[-2] / ds[0] + xi))
```
lamblin: It would be cleaner to explicitly call an integer division (`input.shape[-2] // ds[0]`) rather than do a float division and then cast.

theano/tensor/tests/test_basic.py
```diff
@@ -1650,7 +1650,7 @@ def test_py_c_match():
 _eps = 1e-10
 _good_broadcast_unary_arctanh = dict(
     normal=(rand_ranged(-1 + _eps, 1 - _eps, (2, 3)),),
-    integers=(randint_ranged(-1 + _eps, 1 - _eps, (2, 3)),),
+    integers=(randint_ranged(-1, 1, (2, 3)),),
```
lamblin: arctanh is not defined at -1 or 1, so the only integer-valued point where it makes sense is 0.
The earlier code would basically be equivalent to `random_integers(0, 0, (2, 3))` or `randint(0, 1, (2, 3))`.
Just use zeros(2, 3) I guess...

theano/tensor/tests/test_basic.py
```diff
@@ -7514,17 +7514,17 @@ def test_infer_shape(self):
...
         aiscal = lscalar()
         biscal = lscalar()
         ciscal = lscalar()
         discal = lscalar()
         adscal_val = rand()
-        aiscal_val = randint(3, 5, size=())
-        biscal_val = randint(3, 5, size=())
-        ciscal_val = randint(3, 5, size=())
-        discal_val = randint(3, 5, size=())
+        aiscal_val = randint(low=3, high=5, size=())
```
lamblin: Maybe use `high=6` to keep the same values?

theano/tensor/tests/test_basic.py
```diff
@@ -8027,7 +8027,7 @@ def test_numpy_compare_tuple(self):
         b = tensor.tensor3(dtype='float32')
         c = tensor.tensor3(dtype='float32')
 
-        A = numpy.asarray(numpy.random.random_integers(0, 1, (2, 1, 1)),
+        A = numpy.asarray(numpy.random.randint(low=0, high=1, size=(2, 1, 1)),
```
lamblin: This would only sample zeros, which is probably not what we want. `randint(0, 2, size=(...))` should be equivalent to what we had.

lamblin:
Otherwise, it looks reasonable for that particular warning.
How many others are there to fix?

nouiz:
continued in gh-4859

Issue #4622, opened 2016-06-13
libgpuarray float64 


apulver:
The following code warns about float64 when device=cuda0, but not when device=gpu0. Also, not all "cost" functions result in float64 (see comment in code). 

Incidentally, using the new backend sometimes yields very poor results when using the attached adam implementation, especially with RNNs. I haven't found a simple way to reproduce this problem, and most of the smaller tests I've tried seem to produce accurate results. 

```
import os
device = 'cuda1'
os.environ['THEANO_FLAGS'] = 'floatX=float32,device={},warn_float64=warn'.format(device)
import numpy as np
import theano
import theano.tensor as T

import adam
reload(adam)

x = T.fvector()
y = T.fscalar()
np.random.seed(123)
m = theano.shared(np.random.randn(2,2).astype(np.float32))

cost = (T.dot(x,m) - y).norm(L=2) # results in float64
#cost = x.norm(L=2) + m.norm(L=2) + y # does not result in float64
updates, _, _, _ = adam.adam_updates(cost, [m])

train = theano.function(inputs=[x,y],
                        outputs=cost,
                        updates=updates)

np.random.seed(123)
inv = np.random.randn(2).astype(np.float32)
np.random.seed(123)
ins = np.float32(np.random.randn())

z = train(inv,ins)
print inv, ins,z

```

```
# apulver
# Implementation of adam optimization

import numpy as np
import theano
import theano.tensor as T
from collections import OrderedDict

def adam_updates(cost,
                 params,
                 max_norm=None,
                 alpha=.001,
                 beta1=.9,
                 beta2=.999,
                 eps=1e-8):

    t = theano.shared(np.float64(1).astype(T.config.floatX))
    mt = [theano.shared(np.zeros_like(param.get_value(borrow=True)).astype(T.config.floatX)) for param in params]
    vt = [theano.shared(np.zeros_like(param.get_value(borrow=True)).astype(T.config.floatX)) for param in params]
    lr = theano.shared(np.float64(alpha).astype(T.config.floatX))

    updates = OrderedDict()

    grads = T.grad(cost, params, disconnected_inputs='warn')
    for m, v, param, grad in zip(mt, vt, params, grads):

        if max_norm is not None and max_norm is not False:
            grad_norm = grad.norm(L=2)
            grad = (T.minimum(max_norm, grad_norm)/ (grad_norm + eps)) * grad
        mc = beta1*m + (1.0-beta1)*grad
[...]

PR #4628, opened 2016-06-14
Single stream flag

abergeron:
This include a fix for #4622 (the float64 part).

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -243,9 +243,14 @@ AddConfigVar('gpuarray.preallocate',
              in_c_key=False)
 
 AddConfigVar('gpuarray.single_stream',
-             """Switch between single stream mode or multi-stream
-             mode.  If your computation can't benefit from multiple
-             streams, single-stream is usually faster (by about 10%)
+             """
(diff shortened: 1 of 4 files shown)


theano/configdefaults.py
```diff
@@ -242,6 +242,14 @@ def filter_nvcc_flags(s):
              FloatParam(0),
              in_c_key=False)
 
+AddConfigVar('gpuarray.single_stream',
```

theano/configdefaults.py
```diff
@@ -242,6 +242,14 @@ def filter_nvcc_flags(s):
...
+             """Switch between single stream mode or multi-stream
+             mode.  If your computation can't benefit from multiple
+             streams, single-stream is usually faster (by about 10%)
```
nouiz: Where this 10% come from? My understanding was that the old back-end was speed 1x. WIth single stream, it is 0.9x. With multiple stream, it was > 1x, like 1.1x. So the new back-end multi-stream vs single-stream is more then 10?.

nouiz:
ok to merge when travis pass.

nouiz:
Why not use float X for _one?

PR #4630, opened 2016-06-14
Unbreak travis.

abergeron:


abergeron:
This will fix the travis breakage for now, but since it could re-break soon I'll also make a PR to update everything.

PR #4631, opened 2016-06-14
Update the versions tested on travis

abergeron:
Also update our version descriptors for python versions officially supported.


abergeron:
Finally this does not update anything because we would have to jump to numpy 1.9.3 which is too recent.  This just ensures that travis won't break in the future.

PR #4634, opened 2016-06-15
Set cuda/test_dnn to run conv references in fast run

slefrancois:
cuda/test_dnn now takes up to one hour instead of days on a GTX 680/750.

### DIFF
--- a/theano/sandbox/cuda/tests/test_dnn.py
+++ b/theano/sandbox/cuda/tests/test_dnn.py
@@ -418,13 +418,6 @@ def test_pooling3d():
     if not cuda.dnn.dnn_available() or cuda.dnn.version() < (3000, 3000):
         raise SkipTest(cuda.dnn.dnn_available.msg)
 
-    # For max pooling pool3d2d explicitly pads the input with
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -425,6 +425,9 @@ def test_pooling3d():
     mode_without_gpu2 = mode_without_gpu.including()
```
nouiz: It isn't used anymore, so remove it and the comment above.

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -425,6 +425,9 @@ def test_pooling3d():
...
     mode_without_gpu2.check_isfinite = False
```
slefrancois: Does the part about not checking for infinity still apply? We could add it to the new mode.

nouiz: no, it only applied in DebugMode. But DebugMode won't be used anymore.

PR #4642, opened 2016-06-17
Verify grad speedup

slefrancois:


theano/gradient.py
```diff
@@ -1669,7 +1675,11 @@ def random_projection():
     # This sum() is defined above, it's not the builtin sum.
     cost = theano.tensor.sum(t_r * o_output)
 
-    cost_fn = function(tensor_pt, cost, name='gradient.py cost')
+    if no_debug_ref:
+        cost_fn = function(tensor_pt, cost, name='gradient.py cost',
+                           mode=mode_not_debug(mode))
+    else:
+        cost_fn = function(tensor_pt, cost, name='gradient.py cost')
```
nouiz: This version of the code is more clear:
```
mode_for_cost=mode_not_debug(mode)
cost_fn = function(tensor_pt, cost, name..., mode=mode_for_cost)
```

We see clearly that the only different is the mode.

theano/gradient.py
```diff
@@ -1676,8 +1676,9 @@ def random_projection():
     cost = theano.tensor.sum(t_r * o_output)
 
     if no_debug_ref:
+        mode_for_cost=mode_not_debug(mode)
         cost_fn = function(tensor_pt, cost, name='gradient.py cost',
-                           mode=mode_not_debug(mode))
+                           mode=mode_for_cost)
```
nouiz: I mean, the full ifelse should be:
```
mode_for_cost=mode_not_debug(mode)
cost_fn = function(tensor_pt, cost, name='gradient.py cost',
                           mode=mode_for_cost)
```

slefrancois: Isn't that what the current diff is?

PR #4653, opened 2016-06-21
Fix the gpuarray abstractconv tests.

abergeron:


abergeron:
Sorry this was too early.

PR #4654, opened 2016-06-21
yield test in test_abstract_conv for tensor.nnet, sandbox.cuda and gpuarray

slefrancois:


theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -134,9 +134,8 @@ def run_fwd(self, inputs_shape, filters_shape, ref=conv_corr,
         if target_op is not None:
             assert any([isinstance(n.op, target_op) for n
                         in f.maker.fgraph.toposort()])
-            if check_trace:
```
nouiz: why all check_trace are removed? @abergeron did you do that? @slefrancois don't recall doing that. We can't remove all the check_trace check. We need to keep them. So even if it was wanted to remove them, now it must be added back.

This PR also need a rebase.

slefrancois:
jenkins test this please

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -339,10 +329,11 @@ def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
...
-        self.mode = theano.compile.mode.get_default_mode().excluding('conv_gemm')
-        self.opt_err = theano.config.on_opt_error
+    @classmethod
+    def setup(cls):
+        BaseTestConv2d.setup_class()
+        cls.mode = theano.compile.mode.get_default_mode().excluding('conv_gemm')
+        cls.opt_err = theano.config.on_opt_error
         theano.config.on_opt_error = 'ignore'
 
     def tearDown(self):
```
nouiz: I think you should change this too to the cls version.

nouiz:
ok to merge when travis pass.

PR #4656, opened 2016-06-22
[WIP] add local_lift_reshape optimization, closes #4641, closes #4451

### LINKED ISSUE #4451: Generalize optimization like log(sigmoid) or log(1-sigmoid)  to tolerate reshape/dimshuffle
We should build helper function to we could reuse as they cause "problem" at a few different place.

One report:

https://groups.google.com/d/topic/theano-users/i1kKZmclFFo/discussion

### LINKED ISSUE #4641: Lift reshape op inside elemwise
Some optimization elemwise don't get applied when there is a reshape in the middle of the graph. For example: https://github.com/Theano/Theano/issues/4451

Why not lift reshape as we lift dimshuffle in Elemwise? This would prevent the need of the generalization of possibly many optimizations.

See the optimization local_dimshuffle_lift in tensor/opt.py. I would do the same for Reshape.

JesseLivezey:
- [ ] write tests
- [ ] write optimizations

Still need to understand how the optimization should work better. Using local_lift_dimshuffle as a reference.


JesseLivezey:
Fromm https://groups.google.com/forum/#!topic/theano-users/i1kKZmclFFo/discussion

```python
probs = output_layer.forward(h_final)

logprobs = T.log(probs).reshape(x.shape)
log1mprobs = T.log(1-probs).reshape(x.shape)
logpz = self.logpz = (logprobs * z + log1mprobs * (1-z)) * masks
```

JesseLivezey:
Graph with reshape

Graph with no reshape

theano/tensor/opt.py
```diff
@@ -4189,8 +4189,7 @@ def local_reshape_lift(node):
     """
     if (isinstance(node.op, T.Reshape) and
             node.inputs[0].owner and
-            isinstance(node.inputs[0].owner.op, T.Elemwise) and
-            len(node.inputs[0].owner.inputs) == 1):
+            isinstance(node.inputs[0].owner.op, T.Elemwise)):
         r = node.op(node.inputs[0].owner.inputs[0], node.inputs[1])
```
lamblin: If there are more than 1 input, I think all inputs may have to be reshaped, and if some of them were broadcasted, then the reshape is not going to be as simple as applying `node.op()` on them.

PR #4658, opened 2016-06-22
keep stack trace for FusionOptimizer

### LINKED ISSUE #4179: Keep stack trace: global optimizers in tensor/opt.py
Sub-ticket of #3018.

vmichals:
fixes #4179

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -6880,6 +6880,8 @@ class FusionOptimizer(Optimizer):
                                 list(zip(node.outputs, new_outputs)),
                                 reason=self.__class__.__name__)
                             did_something = True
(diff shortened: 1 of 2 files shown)


lamblin:
I don't think `check_stack_trace` needs anything in particular for functions without outputs.
I'll have a quick look.

lamblin:
I'm not finding anything obvious.
What I would suggest is to use `profile=True,profile_optimizer=True` in the flags, to try to have a list of all the optimizers that actually apply, and try to disable them one by one.
You can also check when the fusion optimizer optimization itself is applied, to see if the intermediate nodes have stack traces or of they have already been stripped.

nouiz:
@ReyhaneAskari What about this PR? Was this done in the other PR you did? Can we close this issue?

PR #4659, opened 2016-06-22
Merged previous work implementing stack trace copy over and tests for…

julianser:
… various ops. #3018. @lamblin

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4520,7 +4520,7 @@ def local_fill_cut(node):
     #         Also, I doubt this optimization is being applied anywhere.
     #         See my comment below.
 
-    #if node.op != T.Elemwise:
+    # if node.op != T.Elemwise:
     #    return False
     if (not isinstance(node.op, T.Elemwise)):
         return False
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4289,10 +4289,6 @@ def local_useless_reshape(node):
         return False
 
     input = node.inputs[0]
-
-    # Copy over stack trace
-    copy_stack_trace(node.outputs[0], input)
-
     output = node.outputs[0]
     output_shape = node.inputs[1]
 
@@ -4303,6 +4299,9 @@ def local_useless_reshape(node):
     # This could hide errors if the user provides inconsistent shapes.
     if (input.ndim == 1 and output.ndim == 1 and
             input.broadcastable == output.broadcastable):
+        # Copy over stack trace
+        copy_stack_trace(node.outputs[0], input)
+
         return [input]
 
     # Second case: all the shapes match the input shape
@@ -4310,62 +4309,9 @@ def local_useless_reshape(node):
     if output_shape.owner and isinstance(output_shape.owner.op, Shape):
         shape_input = output_shape.owner.inputs[0]
         if shape_input == input:
-            return [input]
+            # Copy over stack trace
+            copy_stack_trace(node.outputs[0], input)
 
-    # Match Reshape(x, [x.shape[0], ..., x.shape[-1]]), accounting for
-    # broadcastable and constant dimensions
-    if output_shape.owner and isinstance(output_shape.owner.op, MakeVector):
-        output_shape_is = output_shape.owner.inputs
-
-        if not hasattr(node, 'fgraph'):
-            shape_feature = None
-        else:
-            shape_feature = getattr(node.fgraph, 'shape_feature', None)
-
-        shape_match = [False] * input.ndim
-        for dim in xrange(input.ndim):
-            outshp_i = output_shape_is[dim]
-            # Match Shape_i{dim}(input)
-            if (outshp_i.owner and isinstance(outshp_i.owner.op, Shape_i) and
-                    outshp_i.owner.op.i == dim and
-                    outshp_i.owner.inputs[0] == input):
-                shape_match[dim] = True
-                continue
-
-            # Match Shape(input)[dim]
-            if (outshp_i.owner and isinstance(outshp_i.owner.op, Subtensor) and
-                    len(outshp_i.owner.inputs) == 2 and
-                    extract_constant(outshp_i.owner.inputs[1]) == dim):
-                subtensor_inp = outshp_i.owner.inputs[0]
-                if (subtensor_inp.owner and
-                        isinstance(subtensor_inp.owner.op, Shape)):
-                    shape_input_i = subtensor_inp.owner.inputs[0]
-                    if shape_input_i == input:
-                        shape_match[dim] = True
-                        continue
-
-    op = node.op
-    if not isinstance(op, Reshape):
-        return False
-
-    input = node.inputs[0]
-    output = node.outputs[0]
-    output_shape = node.inputs[1]
-
-    if input.ndim != output.ndim:
-        return False
-
-    # Simple case: both input and output have a single dimension.
-    # This could hide errors if the user provides inconsistent shapes.
-    if (input.ndim == 1 and output.ndim == 1 and
-            input.broadcastable == output.broadcastable):
-        return [input]
-
-    # Second case: all the shapes match the input shape
-    # Match Reshape(x, x.shape)
-    if output_shape.owner and isinstance(output_shape.owner.op, Shape):
-        shape_input = output_shape.owner.inputs[0]
-        if shape_input == input:
             return [input]
 
     # Match Reshape(x, [x.shape[0], ..., x.shape[-1]]), accounting for
@@ -4416,6 +4362,9 @@ def local_useless_reshape(node):
                     continue
 
         if all(shape_match):
+            # Copy over stack trace
+            copy_stack_trace(node.outputs[0], input)
+
             return [input]
 
         # TODO later: if all the shapes except one match, we may want to
@@ -4558,7 +4507,6 @@ def local_fill_cut(node):
     f(fill(a,b), c) -> f(b, c)
     If c.type == a.type.
     """
-
     # this optimization is basically for getting broadcasting to
     # replace fill.  This is always possible when using a Compound
     # Elemwise operation, but it is not always possible without one
@@ -4567,7 +4515,14 @@ def local_fill_cut(node):
     # scalars, but we can't ignore the large matrix because it gives
     # the shape of the result.
 
-    if node.op != T.Elemwise:
+    # Julian: I've fixed the if line below to check if it's an instance.
+    #         Is this correct?
+    #         Also, I doubt this optimization is being applied anywhere.
+    #         See my comment below.
+
+    #if node.op != T.Elemwise:
+    #    return False
+    if (not isinstance(node.op, T.Elemwise)):
         return False
 
     output = node.outputs[0]
@@ -4584,6 +4539,10 @@ def local_fill_cut(node):
     new_inputs = []
     new = False
     for input in node.inputs:
+        # Julian: no matter what kind of function I create,
+        #         it seems that input.owner.op == T.fill is never true.
+        #         Somehow the fill ops are replaced by other ops (e.g. Elemwise{second,no_inplace}).
+        #         If that's true, I don't think we have any tests for this opt.
         if input.owner and input.owner.op == T.fill:
             model, filling = input.owner.inputs
             if encompasses_broadcastable(reference.type.broadcastable,
@@ -5161,9 +5120,10 @@ def local_sum_prod_mul_by_scalar(node):
                 new_op_input_nb_elements = new_op_input.size
                 new_op_output = node.op(new_op_input)
 
-            # Copy over stacktrace from previous output to new mul op,
-            # for same reason as above.
-            copy_stack_trace(node.outputs, new_op_output)
+            if not len(non_scalars) == 0:
+                # Copy over stacktrace from previous output to new mul op,
+                # for same reason as above.
+                copy_stack_trace(node.outputs, new_op_output)
 
             # If node.op is a T.elemwise.Prod, then the scalars need to be
             # raised to the power of the number of elements in the input
@@ -5189,7 +5149,7 @@ def local_sum_prod_mul_by_scalar(node):
                 ret = T.mul(*mul_inputs)
                 # Copy over stacktrace from previous output to new mul op,
                 # for same reason as above.
-                copy_stack_trace(node.outputs,[ret]+mul_inputs)
+                copy_stack_trace(node.outputs, [ret] + mul_inputs)
 
                 return [ret]
 
@@ -5394,11 +5354,12 @@ def local_useless_elemwise_comparison(node):
             cst = get_scalar_constant_value(node.inputs[1],
                                             only_process_constants=True)
 
-            # Copy over stacktrace from previous output.
             res = T.zeros_like(node.inputs[0], dtype=dtype, opt=True)
-            copy_stack_trace(node.outputs, res)
 
             if cst < 0:
+                # Copy over stacktrace from previous output.
+                copy_stack_trace(node.outputs, res)
+
                 return [res]
 
         except NotScalarConstantError:
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4226,7 +4226,7 @@ def local_flatten_lift(node):
         # Copy over stacktrace from previous output node and from unary
         # elementwise output node since if there was an error, it would
         # probably have come from that operation.
-        copy_stack_trace(node.outputs + node.inputs[0], e)
+        copy_stack_trace(node.outputs + [node.inputs[0]], e)
 
         return [e]
 
@@ -4498,9 +4498,9 @@ def local_reshape_lift(node):
         if e.type != node.outputs[0].type:
             re = T.patternbroadcast(e, node.outputs[0].broadcastable)
 
-            # We assume that the broadcast op cannot fail. Thus, if the
-            # graph fails it must be due to previous UnaryElemwise op, and
-            # therefore we must copy its stacktrace over.
+            # Copy over stack trace.
+            # If the graph fails it is usually due to the fact that a dimension
+            # that should be broadcastable does not actually have length 1,
             copy_stack_trace(e, re)
         else:
             re = e
@@ -5200,7 +5200,7 @@ def local_sum_prod_mul_by_scalar(node):
             # There are never errors in the negative op, thus
             # we need only to copy over stacktrace from previous output node to
             # the two new ops.
-            copy_stack_trace(node.outputs, s+ret)
+            copy_stack_trace(node.outputs, [s, ret])
 
             return [ret]
 
@@ -5217,7 +5217,7 @@ def local_elemwise_sub_zeros(node):
             node.inputs[0] == node.inputs[1]):
         res = T.zeros_like(node.inputs[0])
         # Copy over stacktrace from previous output.
-        # Julian: Pascal, is this really necessary? Is there anyway zeros_like can ever fail?
+        # This could help for failures due to out-of-memory.
         copy_stack_trace(node.outputs, res)
         return [res]
 
@@ -5394,9 +5394,14 @@ def local_useless_elemwise_comparison(node):
         try:
             cst = get_scalar_constant_value(node.inputs[1],
                                             only_process_constants=True)
+
+            # Copy over stacktrace from previous output.
+            res = T.zeros_like(node.inputs[0], dtype=dtype, opt=True)
+            copy_stack_trace(node.outputs, res)
+
             if cst < 0:
-                return [T.zeros_like(node.inputs[0],
-                                     dtype=dtype, opt=True)]
+                return [res]
+
         except NotScalarConstantError:
             pass
     return
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4947,9 +4947,8 @@ class Canonizer(gof.LocalOptimizer):
             # theano/tensor/tests/test_opt.py:T_local_switch_sink
             new.tag.values_eq_approx = values_eq_approx_remove_inf_nan
 
-            # Julian: Pascal, maybe you can help me implement the copying of the stacktrace for this class?
-            # Because, it's so general I think we need to copy over the stacktraces of all ops being replaced
-            # to every new op?
+            # We need to implement the copy over of the stacktrace.
+            # See issue #5104.
             return [new]
         else:
             _logger.warning(' '.join(('CANONIZE FAILED: new, out = ',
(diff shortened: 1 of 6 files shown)


theano/tensor/opt.py
```diff
@@ -4050,7 +4050,18 @@ def local_flatten_lift(node):
...
+        # Copy over stacktrace from previous output node (flatten op),
+        # since this is the op which may cause an error for f.
+        copy_stack_trace(node.outputs, f)
+
         e = node.inputs[0].owner.op(f)
+
+        # Copy over stacktrace from previous output node and from unary
+        # elementwise output node since if there was an error, it would
+        # probably have come from that operation.
+        copy_stack_trace(node.outputs + node.inputs[0], e)
```
lamblin: I think you are trying to add a list an a Variable here.

theano/tensor/opt.py
```diff
@@ -4653,6 +4696,10 @@ def same(x, y):
             # This happen with test
             # theano/tensor/tests/test_opt.py:T_local_switch_sink
             new.tag.values_eq_approx = values_eq_approx_remove_inf_nan
+
+            # Julian: Pascal, maybe you can help me implement the copying of the stacktrace for this class?
+            # Because, it's so general I think we need to copy over the stacktraces of all ops being replaced
+            # to every new op?
```
lamblin: I'll have to think about it. It can indeed rewrite large graphs, and I'm not sure what would be the right thing.

theano/tensor/opt.py
```diff
@@ -4773,7 +4845,11 @@ def local_elemwise_sub_zeros(node):
             node.op.scalar_op.nin == 2 and
             node.op.scalar_op == scalar.sub and
             node.inputs[0] == node.inputs[1]):
-        return [T.zeros_like(node.inputs[0])]
+        res = T.zeros_like(node.inputs[0])
+        # Copy over stacktrace from previous output.
+        # Julian: Pascal, is this really necessary? Is there anyway zeros_like can ever fail?
```
lamblin: If it fails for out-of-memory reasons, there could be pathological cases where it could help hypothetically.
Given that it is easy to just add the stack trace, I would keep it.

theano/tensor/opt.py
```diff
@@ -4755,12 +4811,28 @@ def local_sum_prod_mul_by_scalar(node):
...
+                return [ret]
 
         if isinstance(node.op, T.Sum) and node_inps.owner and node_inps.owner.op == T.neg:
-            return [T.neg(node.op(node_inps.owner.inputs[0]))]
+            s = node.op(node_inps.owner.inputs[0])
+            ret = T.neg(s)
+            # There are never errors in the negative op, thus
+            # we need only to copy over stacktrace from previous output node to
+            # the two new ops.
+            copy_stack_trace(node.outputs, s+ret)
```
lamblin: `s` and `ret` are Variables, right? Then `s+ret` just ocmputes their sum. The stack assignment would not really do anything, then.

theano/tensor/opt.py
```diff
@@ -4755,12 +4811,28 @@ def local_sum_prod_mul_by_scalar(node):
...
+                # for same reason as above.
+                copy_stack_trace(node.outputs, mul_inputs)
+
                 return mul_inputs
             else:
-                return [T.mul(*mul_inputs)]
+                ret = T.mul(*mul_inputs)
+                # Copy over stacktrace from previous output to new mul op,
+                # for same reason as above.
+                copy_stack_trace(node.outputs, ret+mul_inputs)
```
lamblin: Variable + list should not work.

lamblin:
I had a quick look, but I need to come back later.

theano/tensor/opt.py
julianser: Ah, well spotted. This should cause an error, I will fix it and test it again.

theano/tensor/opt.py
julianser: I'll fix it.

theano/tensor/opt.py
julianser: Yes, somehow I thought I was concatenating them like this, but obviously not! I'll fix it.

theano/tensor/opt.py
julianser: Good point, I'll leave the stack trace in there.

theano/tensor/opt.py
```diff
@@ -4121,16 +4140,33 @@ def local_reshape_lift(node):
...
         # In rare case the original broadcast was (False, True), but
         # the new one is (False, False). So don't crash in that case.
         if e.type != node.outputs[0].type:
-            e = T.patternbroadcast(e, node.outputs[0].broadcastable)
-        return [e]
+            re = T.patternbroadcast(e, node.outputs[0].broadcastable)
+
+            # We assume that the broadcast op cannot fail. Thus, if the
+            # graph fails it must be due to previous UnaryElemwise op, and
+            # therefore we must copy its stacktrace over.
```
lamblin: If the broadcasting operation raises an exception, that is usually due to the fact that a dimension that should be broadcastable does not actually have length 1. In that case, the responsible Op will likely be the reshape, not the elemwise.
This is just a comment on your comment, though, since the code ends up copying both original traces, which I think is OK.

theano/tensor/opt.py
lamblin: I think we should open a separate ticket just for that one and not worry about it for now.
Actually, #5104.

theano/tensor/tests/test_opt.py
```diff
@@ -3556,6 +3557,7 @@ def test_equality_shapes(self):
             f = theano.function([x], T.eq(g, 0))
             assert f([3, 3]) == 0
             assert f([]) == 1
+            self.assertTrue(check_stack_trace(f, ops_to_check='last'))
```
lamblin: Wait, which optimization is supposed to be tested here? What does the final graph look like?

theano/tensor/tests/test_opt.py
```diff
@@ -6088,7 +6091,9 @@ def test_local_useless_reshape():
     f = theano.function([i], m, mode=mode)
     topo = f.maker.fgraph.toposort()
     assert not any(isinstance(n.op, tensor.basic.Reshape) for n in topo)
-
+    theano.printing.debugprint(f, print_type=True)
```
lamblin: You can remove that one, we try to avoid printing things during tests now (even though many tests still do).

theano/tensor/tests/test_opt.py
```diff
@@ -3567,6 +3569,7 @@ def test_equality_shapes(self):
         f = theano.function([x], T.eq(g, 0))
         assert (f([3, 3]) == 0).all()
         assert (f([]) == 1).all()
+        self.assertTrue(check_stack_trace(f, ops_to_check='last'))
```
lamblin: Same here, which optimization gets applied, and what is supposed to be the last node?

theano/tensor/tests/test_opt.py
```diff
@@ -6088,7 +6091,9 @@ def test_local_useless_reshape():
     f = theano.function([i], m, mode=mode)
     topo = f.maker.fgraph.toposort()
     assert not any(isinstance(n.op, tensor.basic.Reshape) for n in topo)
-
+    theano.printing.debugprint(f, print_type=True)
+    # Check stacktrace was copied over correctly after opt was applied
+    assert check_stack_trace(f, ops_to_check='all')
```
lamblin: OK.
Although `local_useless_reshape` itself is not supposed to insert new Ops for now, it may do so in the future, so it's good to have a test.

theano/tensor/tests/test_opt.py
```diff
@@ -6101,7 +6106,8 @@ def test_local_reshape_lift():
     topo = f.maker.fgraph.toposort()
     assert isinstance(topo[-2].op, tensor.Reshape)
     assert isinstance(topo[-1].op, tensor.Elemwise)
-
+    # Check stacktrace was copied over correctly after opt was applied
+    assert check_stack_trace(f, ops_to_check='last')
```
lamblin: I think you should check both the last one (Elemwise) and the previous one (Reshape), that was affected by the optimization as well.

theano/tensor/tests/test_opt.py
```diff
@@ -3467,6 +3467,7 @@ def assert_eqs_const(self, f, val, op=deep_copy_op):
             assert isinstance(elem.inputs[0], T.TensorConstant), elem
             assert T.extract_constant(elem.inputs[0]) == val, val
 
+
```
lamblin: I'm pretty sure this empty line goes against pep8

nouiz:
travis have those errors

```
/home/travis/build/Theano/Theano/theano/tensor/opt.py:5192:46: E231 missing whitespace after ','
/home/travis/build/Theano/Theano/theano/tensor/opt.py:5192:52: E226 missing whitespace around arithmetic operator
```

theano/tensor/opt.py
```diff
@@ -5157,10 +5162,9 @@ def local_sum_prod_mul_by_scalar(node):
                 new_op_input_nb_elements = new_op_input.size
                 new_op_output = node.op(new_op_input)
 
-            if not len(non_scalars) == 0:
-                # Copy over stacktrace from previous output to new mul op,
-                # for same reason as above.
-                copy_stack_trace(node.outputs, new_op_output)
+            # Copy over stacktrace from previous output to new mul op,
+            # for same reason as above.
+            copy_stack_trace(node.outputs, new_op_output)
```
lamblin: That change may have been introduced by the rebase, but is not correct: if `len(non_scalars)` equals 0, then `new_op_output` is the Python literal `1`, which does not have a tag or stack trace.

theano/tensor/opt.py
```diff
@@ -5186,7 +5190,7 @@ def local_sum_prod_mul_by_scalar(node):
                 ret = T.mul(*mul_inputs)
                 # Copy over stacktrace from previous output to new mul op,
                 # for same reason as above.
-                copy_stack_trace(node.outputs, [ret] + mul_inputs)
+                copy_stack_trace(node.outputs, ret+mul_inputs)
```
lamblin: Same here, `ret` is a variable, and `mul_inputs` is a list.
This actually makes some tests fail.

theano/tensor/tests/test_opt.py
```diff
@@ -6294,17 +6294,17 @@ def test_1(self):
         topo = f1.maker.fgraph.toposort()
         assert not any(isinstance(n.op, tensor.basic.Reshape) for n in topo)
 
-        # Check stacktrace was copied over correctly after opt was applied
-        assert check_stack_trace(f1, ops_to_check='all')
+        # TODO: Check that stack trace is maintained.
+        #       Currently, stack trace gets removed by some other opt.
+        #assert check_stack_trace(f1, ops_to_check='all')
+
+        m2 = m0.excluding('local_useless_reshape')
```
lamblin: Actually, `local_useless_reshape` seems to only remove nodes from the graph now, not insert new ones, so it is OK.
IIRC, the part of that optimization that inserted DimShuffle nodes has been moved to `local_reshape_to_dimshuffle`. Could you check that one (and its tests) instead?

theano/tensor/opt.py
```diff
@@ -5232,9 +5393,14 @@ def investigate(node):
         try:
             cst = get_scalar_constant_value(node.inputs[1],
                                             only_process_constants=True)
+
+            # Copy over stacktrace from previous output.
+            res = T.zeros_like(node.inputs[0], dtype=dtype, opt=True)
+            copy_stack_trace(node.outputs, res)
```
lamblin: I'd put those inside the `if`, to avoid doing unnecessary work.

theano/tensor/tests/test_opt.py
```diff
@@ -3727,6 +3725,22 @@ def test_xor(self):
...
+        x = T.vector('x', dtype=config.floatX)
+        f = theano.function([x], T.gt(x, x), mode=mode)
+        self.assertTrue(check_stack_trace(f, ops_to_check='last'))
+
+        f = theano.function([x], T.le(x, x), mode=mode)
+        self.assertTrue(check_stack_trace(f, ops_to_check='last'))
+
+        # Julian: I tried testing the stack trace for a bunch of different
+        # functions, including maximum and shapes, but other opts remove
+        # the stack traces in this case.
```
lamblin: For "minimum(x, x) => x" and "maximum(x, x) => x", the node itself is removed, so there is not much to check.
Same for "maximum(shape_i(...), 0) => shape_i(...)".
Indeed, no stack trace was copied in those cases.

However, it should work for:
"minimum(shape_i(...), 0) => 0"
"T.eq(shape_i(...), -X) => 0"

theano/tensor/tests/test_opt.py
juliangrams: Okay, I've added comment. Also tests for local_reshape_to_dimshuffle run fine (and they include testing the stack trace).

theano/tensor/opt.py
juliangrams: Another rebase error. I'll fix it.

theano/tensor/opt.py
juliangrams: I'll fix it.

theano/tensor/tests/test_opt.py
juliangrams: I tried adding this test:

    mode = theano.compile.get_default_mode().including(
            'local_useless_elemwise_comparison')
    f = theano.function([x], T.minimum(Shape_i(0)(x), 0), mode=mode)
    self.assertTrue(check_stack_trace(f, ops_to_check='all'))

However, it doesn't work. The stack trace gets removed somewhere along the way.

The unoptimized graph looks like this:

    Elemwise{minimum,no_inplace} [id A] ''   
     |Shape_i{0} [id B] ''   
     | |x [id C]
     |TensorConstant{0} [id D]

The optimized graph looks like this:

    DeepCopyOp [id A] ''   0
     |TensorConstant{0} [id B]

I'm not sure what's going on, but "local_useless_elemwise_comparison" definitely does copy over the stack trace.

lamblin: Indeed, in the case you mention, only the `TensorConstant{0}` remains in the graph, and since it is a scalar constant, it it re-used everywhere and does not have a stack trace of its creation. Then the DeepCopyOp gets inserted to avoid returning internal memory to the user.
In any case, there is no node created by the optimization in this case, so there is no stack trace to look for. I was mistaken about that in my comment above.
It may be the same for the other csae, since again we are returning a scalar 0, not an alloc(0, ...).

theano/tensor/opt.py
```diff
@@ -4584,6 +4539,10 @@ def local_fill_cut(node):
     new_inputs = []
     new = False
     for input in node.inputs:
+        # Julian: no matter what kind of function I create,
+        #         it seems that input.owner.op == T.fill is never true.
+        #         Somehow the fill ops are replaced by other ops (e.g. Elemwise{second,no_inplace}).
+        #         If that's true, I don't think we have any tests for this opt.
```
lamblin: I've looked around a bit, and I'm quite confident we can remove that optimization entirely, and that the equivalent will be performed by other ones.
Could you take care of that? There may be some clean up to do in the documentation as well.

lamblin: We have tests for `test_local_fill_useless` in `theano/tensor/tests/test_op.py`, and they still pass.

theano/tensor/tests/test_opt.py
```diff
@@ -3446,6 +3446,8 @@ def test_local_fill_useless():
     f = function([x], T.fill(x, x) * 2, mode=mode_opt)
     assert [node.op for node in f.maker.fgraph.toposort()] == [T.mul]
     f(x_)
+    # Julian: This doesn't work. See comments inside local_fill_cut.
+    # assert check_stack_trace(f, ops_to_check='all')
```
lamblin: OK. We would have to identify which other optimization is at play here, and I did not manage to do that quickly.
Let's put a placeholder comment.

theano/tensor/tests/test_opt.py
```diff
@@ -3482,6 +3482,61 @@ def test_local_fill_useless():
...
+    mode = theano.compile.get_default_mode()\
+            .excluding('canonicalize', 'uncanonicalize',\
+                       'ShapeOpt', 'local_fill_to_alloc',\
+                       'local_elemwise_alloc')\
+            .including('local_elemwise_sub_zeros')
+
+    # Test scalar minus scalar
+    f = function([scalar], scalar-scalar, mode=mode)
+    # Check optimized graph is correct
+    assert isinstance(f.maker.fgraph.toposort()[0].op, T.Elemwise)
```
lamblin: I'm confused, if there is still an Elemwise in the graph, doesn't it mean that the optimization has not been applied?

theano/tensor/tests/test_opt.py
```diff
@@ -5350,6 +5418,8 @@ def test_reduction_opt(inputs, inputs_val, reduction_op,
                            (s1_val * s2_val * v_val * m_val).prod(), 2)
 
     def test_local_sum_prod_all_to_none(self):
+        # Julian: It appears that the opt local_sum_prod_mul_by_scalar
+        #         is never used in any of these tests...
```
lamblin: I thought it was tested on the function above, `test_local_sum_prod_mul_by_scalar`.
In any case, the test below seems good, maybe just mention "local_sum_prod_mul_by_scalar" in the comment.

theano/tensor/tests/test_opt.py
```diff
@@ -5584,6 +5654,37 @@ def test_local_sum_sum_dtype(self):
         finally:
             config.on_opt_error = backup
 
+    def test_stack_trace(self):
+        """
+        Test that stack trace is copied over correctly.
+        """
```
lamblin: Actually, we now prefer to have a comment rather than a docstring, since nosetests annoyingly replaces the test name by the docstring when displaying which test is being executed.

julianser: Fixed.

theano/tensor/tests/test_opt.py
julianser: Oh, I misunderstand. Of course "test_local_sum_prod_all_to_none" is not supposed to test the local_sum_prod_mul_by_scalar opt. I've removed the comment.

theano/tensor/opt.py
```diff
@@ -4303,69 +4299,19 @@ def local_useless_reshape(node):
...
-        return False
-
-    input = node.inputs[0]
-    output = node.outputs[0]
-    output_shape = node.inputs[1]
-
-    if input.ndim != output.ndim:
-        return False
+        # Copy over stack trace
+        copy_stack_trace(node.outputs[0], input)
```
julianser: Ah, of course.

theano/tensor/opt.py
```diff
@@ -4303,69 +4299,19 @@ def local_useless_reshape(node):
...
-            input.broadcastable == output.broadcastable):
         return [input]
 
     # Second case: all the shapes match the input shape
     # Match Reshape(x, x.shape)
     if output_shape.owner and isinstance(output_shape.owner.op, Shape):
         shape_input = output_shape.owner.inputs[0]
         if shape_input == input:
+            # Copy over stack trace
+            copy_stack_trace(node.outputs[0], input)
```
julianser: Fixed.

theano/tensor/opt.py
```diff
@@ -4416,6 +4362,9 @@ def local_useless_reshape(node):
                     continue
 
         if all(shape_match):
+            # Copy over stack trace
+            copy_stack_trace(node.outputs[0], input)
```
julianser: Fixed.

theano/tensor/tests/test_opt.py
julianser: Fixed.

theano/tensor/opt.py
julianser: Alright, I'll remove local_fill_cut.

theano/tensor/tests/test_opt.py
julianser: If we remove the opt local_fill_cut, then there is nothing to test. So, I'll just remove my comments here.

theano/tensor/tests/test_opt.py
julianser: Actually, there is still an Elemwise after the optimization. It's simply a fill op.

This is the graph before optimizing:
    Elemwise{sub,no_inplace} [id A] ''   
     |<TensorType(float64, scalar)> [id B]
     |<TensorType(float64, scalar)> [id B]

This is the graph after optimizing:
    Elemwise{second,no_inplace} [id A] ''   0
     |<TensorType(float64, scalar)> [id B]
     |TensorConstant{0.0} [id C]

I'll add a test to check specifically that the Elemwise is a fill op (Elemwise{second,no_inplace}).

theano/tensor/tests/test_opt.py
julianser: Fixed.

theano/tensor/opt.py
lamblin: Can you remove that comment and replace it with a reference to ticket #5104?

theano/tensor/tests/test_opt.py
```diff
@@ -3502,6 +3502,8 @@ def test_local_elemwise_sub_zeros():
     f = function([scalar], scalar - scalar, mode=mode)
     # Check optimized graph is correct
     assert isinstance(f.maker.fgraph.toposort()[0].op, T.Elemwise)
+    assert f.maker.fgraph.toposort()[0].op.name\
+        == 'Elemwise{second,no_inplace}'
```
julianser: Good point. I guess in general it's not a good idea to compare to a string name of an op... I've fixed it now.

theano/tensor/tests/test_opt.py
```diff
@@ -3514,6 +3516,8 @@ def test_local_elemwise_sub_zeros():
     f = function([vect], vect - vect, mode=mode)
     # Check optimized graph is correct
     assert isinstance(f.maker.fgraph.toposort()[0].op, T.Elemwise)
+    assert f.maker.fgraph.toposort()[0].op.name\
+        == 'Elemwise{second,no_inplace}'
```
julianser: Fixed.

theano/tensor/tests/test_opt.py
```diff
@@ -3526,6 +3530,8 @@ def test_local_elemwise_sub_zeros():
     f = function([mat], mat - mat, mode=mode)
     # Check optimized graph is correct
     assert isinstance(f.maker.fgraph.toposort()[0].op, T.Elemwise)
+    assert f.maker.fgraph.toposort()[0].op.name\
+        == 'Elemwise{second,no_inplace}'
```
julianser: Fixed.

PR #4663, opened 2016-06-23
Updated Numpy copyright dates

SebastinSanty:
Numpy has changed copyright dates to 2005-2016. [Numpy License](https://github.com/numpy/numpy/blob/master/LICENSE.txt)


lamblin:
Since we might have copied new code from numpy since 2011, I don't see any harm in merging.

PR #4669, opened 2016-06-25
Partial evaluation in CVM

### LINKED ISSUE #4624: Partial function evaluation, implement in the cvm
Follow up of https://github.com/Theano/Theano/issues/2472. Make it work for the CVM, not just Stack.

sygi:
I started to implement partial evaluation in CVM. I didn't do the `output_keys` yet, just the partial evaluation when the `output_subset` is a list of indices.

Could you take a look?

Tries to fix #4624.

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -735,9 +735,13 @@ class Function(object):
         kwargs : dict
             The function inputs can be passed as keyword argument. For this, use
             the name of the input or the input instance as the key.
+
             Keyword argument ``output_subset`` is a list of either indices of the
             function's outputs or the keys belonging to the `output_keys` dict
-            and represent outputs that are requested to be calculated.
+            and represent outputs that are requested to be calculated. Regardless
+            of the presence of ``output_subset``, the updates are always calculated
+            and processed. To disable the updates, you should use the ``copy``
+            method with ``delete_updates=False``.
 
         Returns
         -------
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -741,7 +741,7 @@ class Function(object):
             and represent outputs that are requested to be calculated. Regardless
             of the presence of ``output_subset``, the updates are always calculated
             and processed. To disable the updates, you should use the ``copy``
-            method with ``delete_updates=False``.
+            method with ``delete_updates=True``.
 
         Returns
         -------
--- a/theano/gof/lazylinker_c.c
+++ b/theano/gof/lazylinker_c.c
@@ -865,19 +865,19 @@ CLazyLinker_call(PyObject *_self, PyObject *args, PyObject *kwds)
             }
         }
 
-        for (int i = 0; i < self->n_output_vars && (!err); ++i)
-          {
-            if (output_subset == NULL || output_subset[i] == 1)
-              {
-                err = lazy_rec_eval(self, self->output_vars[i], one, zero);
-              }
-          }
+      int first_updated = self->n_output_vars - self->n_updates;
+      for (int i = 0; i < self->n_output_vars && (!err); ++i)
+        {
+          if (i >= first_updated || output_subset == NULL || output_subset[i] == 1)
+            {
+              err = lazy_rec_eval(self, self->output_vars[i], one, zero);
+            }
+        }
 
       if (!err)
         {
           // save references to outputs prior to updating storage containers
           assert (self->n_output_vars >= self->n_updates);
-          assert (output_subset_size < 0 || output_subset_size >= self->n_updates);
           Py_DECREF(rval);
           rval = PyList_New(self->n_output_vars);
           for (int i = 0; i < (self->n_output_vars); ++i)
--- a/theano/gof/lazylinker_c.c
+++ b/theano/gof/lazylinker_c.c
@@ -801,24 +801,35 @@ CLazyLinker_call(PyObject *_self, PyObject *args, PyObject *kwds)
                                     &output_subset_ptr))
     return NULL;
 
+  int err = 0;
   // parse an output_subset list
   // it is stored as a bool list of length n_output_vars: calculate a var or not
   char *output_subset = NULL;
   int output_subset_size = -1;
   if (output_subset_ptr != NULL)
     {
-      assert (PyList_Check(output_subset_ptr));
-      output_subset_size = PyList_Size(output_subset_ptr);
-      output_subset = (char*)calloc(self->n_output_vars, sizeof(char));
-      for (int it = 0; it < output_subset_size; ++it)
+      if (! PyList_Check(output_subset_ptr))
         {
-          PyObject *elem = PyList_GetItem(output_subset_ptr, it);
-          assert (PyInt_Check(elem));
-          output_subset[PyInt_AsLong(elem)] = 1;
+          err = 1;
+          PyErr_SetString(PyExc_RuntimeError, "Output_subset is not a list");
+        }
+      else
+        {
+          output_subset_size = PyList_Size(output_subset_ptr);
+          output_subset = (char*)calloc(self->n_output_vars, sizeof(char));
+          for (int it = 0; it < output_subset_size; ++it)
+            {
+              PyObject *elem = PyList_GetItem(output_subset_ptr, it);
+              if (! PyInt_Check(elem))
+                {
+                  err = 1;
+                  PyErr_SetString(PyExc_RuntimeError, "Some elements of output_subset list are not int");
+                }
+              output_subset[PyInt_AsLong(elem)] = 1;
+            }
         }
     }
 
-  int err = 0;
   self->position_of_error = -1;
   // create constants used to fill the var_compute_cells
   PyObject * one = PyInt_FromLong(1);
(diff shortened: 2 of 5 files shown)


theano/gof/lazylinker_c.c
```diff
@@ -789,14 +789,35 @@ CLazyLinker_call(PyObject *_self, PyObject *args, PyObject *kwds)
...
+                                    &output_subset_ptr))
     return NULL;
+
+  // parse an output_subset list
+  // it is stored as a bool list of length n_output_vars: calculate a var or not
+  char *output_subset = NULL;
+  int output_subset_size = -1;
+  if (output_subset_ptr != NULL)
+    {
+      assert (PyList_Check(output_subset_ptr));
```
nouiz: It would be better to raise a Python error then doing a c assert. This would give a better error message and allow the user to catch the problem.

theano/gof/lazylinker_c.c
```diff
@@ -789,14 +789,35 @@ CLazyLinker_call(PyObject *_self, PyObject *args, PyObject *kwds)
...
+      output_subset_size = PyList_Size(output_subset_ptr);
+      output_subset = (char*)calloc(self->n_output_vars, sizeof(char));
+      for (int it = 0; it < output_subset_size; ++it)
+        {
+          PyObject *elem = PyList_GetItem(output_subset_ptr, it);
+          assert (PyInt_Check(elem));
```
nouiz: idem

theano/gof/lazylinker_c.c
```diff
@@ -833,22 +854,27 @@ CLazyLinker_call(PyObject *_self, PyObject *args, PyObject *kwds)
...
+              {
+                err = lazy_rec_eval(self, self->output_vars[i], one, zero);
+              }
+          }
 
       if (!err)
         {
           // save references to outputs prior to updating storage containers
           assert (self->n_output_vars >= self->n_updates);
+          assert (output_subset_size < 0 || output_subset_size >= self->n_updates);
```
nouiz: The second part of the condition is only needed if the user can only skip the real output of the function and not the updates. Is that the case? I find that strange, as there is no code that mark the updates as needed to be computed in that PR.

theano/gof/tests/test_vm.py
```diff
@@ -197,14 +197,19 @@ def time_linker(name, linker):
...
-        optimizer=None, linker=vm.VM_Linker(allow_partial_eval=True)))
 
-    assert f(3, output_subset=[0, 1, 2]) == f(3)
-    assert f(4, output_subset=[0, 2]) == [f(4)[0], f(4)[2]]
-    utt.assert_allclose(f(5), np.array([32., 16., 1.7857142857142858]))
+    def check_partial_function(linker_name):
+        x = tensor.scalar('input')
+        y = x ** 2
+        f = theano.function([x], [y + 7, y - 9, y / 14.], mode=Mode(
+            optimizer=None, linker=linker_name))
```
nouiz: There is no test of functions with updates. As I'm not sure how we handled that case, could you add a tests to make sure it behave as expected?

theano/gof/tests/test_vm.py
```diff
@@ -197,14 +197,19 @@ def time_linker(name, linker):
...
+
+        assert f(3, output_subset=[0, 1, 2]) == f(3)
+        assert f(4, output_subset=[0, 2]) == [f(4)[0], f(4)[2]]
+        utt.assert_allclose(f(5), np.array([32., 16., 1.7857142857142858]))
+
+    check_partial_function(vm.VM_Linker(allow_partial_eval=True))
+    check_partial_function('cvm')
```
nouiz: in the file vm.py, there is code that will force the vm and skip the CVM. Grep for allow_partial_eval in that file. Currently, both case test the vm and not the cvm.

nouiz:
thanks for continuing this. Sorry for the delay in review, I was in vacation.

theano/gof/tests/test_vm.py
sygi: Thanks, I was not sure why it seemed not to run. Should I pass a warning when I enforce usage of Stack instead of LoopGC, which doesn't support partial evaluation?

theano/gof/lazylinker_c.c
sygi: Done. It's quite improbable that it will be risen anyway, as there are also checks in python code before.

theano/gof/tests/test_vm.py
nouiz: yes, you already did that in the new commit.

nouiz:
The new commit seem good. Just don't forget my 2 other comments.

thanks

theano/gof/tests/test_vm.py
sygi: I implemented the handling of updates in partially evaluated functions. Apparently, that case was not working in either Stack nor CVM. In the code, I assumed that the outputs corresponding to the updates are after the regular outputs, but I couldn't find the piece of code that populates the outputs (it is somewhere in the `accept` function of linkers, but it seems quite convoluted). Am I right about this assumption?

nouiz: Yes, it is the right assumption.

theano/gof/tests/test_vm.py
```diff
@@ -221,6 +222,25 @@ def test_partial_function_output_keys():
...
+def test_partial_function_with_updates():
+
+    def check_updates(linker_name):
+        x = tensor.lscalar('input')
+        y = theano.shared(1, name='global')
+        f = theano.function([x], [x, x + 34], updates=[(y, x + 1)], mode=Mode(
+            optimizer=None, linker=linker_name))
+        g = theano.function([x], [x - 6], updates=[(y, y + 3)], mode=Mode(
+            optimizer=None, linker=linker_name))
+        f(3, output_subset=[])
```
nouiz: assert the output of f at the same time, to be sure it is None.

theano/gof/tests/test_vm.py
```diff
@@ -221,6 +222,25 @@ def test_partial_function_output_keys():
...
+        assert(y.get_value() == 4)
+        assert(g(30, output_subset=[0]) == [24])
+        g(40, output_subset=[])
```
nouiz: idem

theano/gof/vm.py
```diff
@@ -332,7 +332,8 @@ def recursively_evaluate(var):
 
     def __init__(self, nodes, thunks, pre_call_clear,
                  storage_map, compute_map, fgraph, allow_gc,
-                 dependencies=None, callback=None, callback_input=None):
+                 dependencies=None, callback=None, callback_input=None,
+                 n_updates=0):
```
nouiz: I don't like that n_updates is 0 by default. Can you make it None by default and if the user use output_subset, then assert it isn't None? This would prevent some error hard to diagnose by the user.

nouiz:
It seem pretty close to be merged. thanks.

theano/gof/tests/test_vm.py
sygi: Shoudn't it be [] and not None?

nouiz: I'm not sure, both seem good to me. Just assert what is already there.

theano/gof/tests/test_vm.py
sygi: done.

theano/gof/vm.py
sygi: I made it an obligatory param, as I pass it everywhere explicitly anyway.

nouiz:
thanks. I think it should be documented that if there is update they will be always evaluated. Just so that people know about this. You could also link to this function in case they want to skip the updates (they could just make a copy of the Theano function with the parameter to remove the updates:

http://deeplearning.net/software/theano/library/compile/function.html#theano.compile.function_module.Function.copy

sygi:
done.

theano/gof/tests/test_vm.py
```diff
@@ -212,14 +212,18 @@ def check_partial_function(linker_name):
...
+
+    def check_partial_function_output_keys(linker_name):
+        x = tensor.scalar('input')
+        y = 3 * x
+        f = theano.function([x], {'a': y * 5, 'b': y - 7}, mode=Mode(
+            optimizer=None, linker=linker_name))
+
+        assert f(5, output_subset=['a'])['a'] == f(5)['a']
+
+    check_partial_function_output_keys(vm.VM_Linker(allow_partial_eval=True))
```
nouiz: You should also add use_cloop=False to force the python version. Otherwise, with this PR, it will trigger the C code in both case.

theano/compile/function_module.py
```diff
@@ -735,9 +735,13 @@ def __call__(self, *args, **kwargs):
...
             The function inputs can be passed as keyword argument. For this, use
             the name of the input or the input instance as the key.
+
             Keyword argument ``output_subset`` is a list of either indices of the
             function's outputs or the keys belonging to the `output_keys` dict
-            and represent outputs that are requested to be calculated.
+            and represent outputs that are requested to be calculated. Regardless
+            of the presence of ``output_subset``, the updates are always calculated
+            and processed. To disable the updates, you should use the ``copy``
+            method with ``delete_updates=False``.
```
nouiz: delete_updates=True.

nouiz:
thanks!

PR #4570, opened 2016-06-01
New graph2gpu

Sentient07:
The continuation of https://github.com/Theano/Theano/pull/4551 after a pathetic rebase.

PR #4672, opened 2016-06-27
Clean up

Sentient07:
The cleanup of PR  #4570. Will rebase with the master once  #4570 gets merged. 
Tasks :
- [x] Correctly handle HostFromGpu nodes 
- [ ]  Remove the special case for Alloc / AllocEmpty from the new optimizer, and register these lifters in an "out2in" topooptimizer
- [x] Replace op_lifter with out2in
- [x] Factor out caching of Op instances 
- [ ] Check if all the optimizers used by op_lifter is being used by op_lifter_topo
- [ ] Implement backward pass.
- [ ] Remove op_lifter from EquilibriumOptimizer

@lamblin Am i missing anything in the to-do list?

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -70,7 +70,7 @@ def as_gpuarray_variable(x, context_name):
 
         # If we couldn't deal with transfers, then maybe it's a tensor
         if isinstance(x.type, tensor.TensorType):
-            return gpu_from_host(context_name)(x)
+            return GpuFromHost(context_name)(x)
 
     # Try _as_GpuArrayVariable if possible
     if hasattr(x, '_as_GpuArrayVariable'):
@@ -544,7 +544,7 @@ class HostFromGpu(Op):
 
     def grad(self, inputs, grads):
         gz, = grads
-        return [gpu_from_host(inputs[0].type.context_name)(gz)]
+        return [GpuFromHost(inputs[0].type.context_name)(gz)]
 
     def R_op(self, inputs, eval_points):
         ev, = eval_points
@@ -645,14 +645,6 @@ class GpuFromHost(Op):
 
     def c_code_cache_version(self):
         return (9,)
-
-
-# Caching GPUAlloc
-def gpu_from_host(ctx):
-    if ctx not in gpu_alloc.cache:
-        gpu_from_host.cache[ctx] = GpuFromHost(ctx)
-    return gpu_from_host.cache[ctx]
-gpu_from_host.cache = {}
 
 
 class GpuToGpu(Op):
@@ -878,15 +870,6 @@ class GpuAlloc(HideC, Alloc):
         return True
 
 
-# Caching GPUAlloc
-def gpu_alloc(ctx, memset_0=False):
-    key = (ctx, memset_0)
-    if key not in gpu_alloc.cache:
-        gpu_alloc.cache[key] = GpuAlloc(ctx, memset_0)
-    return gpu_alloc.cache[key]
-gpu_alloc.cache = {}
-
-
 class GpuAllocEmpty(HideC, Alloc):
     """
     Allocate uninitialized memory on the GPU.
@@ -896,7 +879,7 @@ class GpuAllocEmpty(HideC, Alloc):
     _f16_ok = True
     params_type = gpu_context_type
 
-    def __init__(self, dtype, context_name):
+    def __init__(self, dtype, context_name=None):
         self.dtype = dtype
         self.context_name = context_name
 
@@ -971,14 +954,6 @@ if (theano_prep_output(&%(zz)s, %(ndim)s, shape, %(type)s, GA_C_ORDER,
 
 def empty_like(var):
     return GpuAllocEmpty(var.type.dtype, var.type.context_name)(*var.shape)
-
-
-def gpu_alloc_empty(ctx, dtype):
-    key = (dtype, ctx)
-    if key not in gpu_alloc_empty.cache:
-        gpu_alloc_empty.cache[key] = GpuAllocEmpty(dtype, ctx)
-    return gpu_alloc_empty.cache[key]
-gpu_alloc_empty.cache = {}
 
 
 class GpuContiguous(Op):
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -588,8 +588,8 @@ class GpuFromHost(Op):
 
     def grad(self, inputs, grads):
         gz, = grads
-        return [host_from_gpu(as_gpuarray_variable(
-                gz, context_name=self.context_name))]
+        return [as_gpuarray_variable(
+                gz, context_name=self.context_name).transfer('cpu')]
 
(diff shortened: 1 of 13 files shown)


theano/gpuarray/extra_ops.py
```diff
@@ -468,4 +469,4 @@ def use_gpu_cumsumop(op, ctx_name, inputs, outputs):
         if axis is None:
             axis = 0
         assert isinstance(x.type, GpuArrayType)
-        return GpuCumsum(axis)(x)
+        return GpuCumsum(axis, ctx_name)(x)
```
Sentient07: I felt it is better to pass the context_name. Will change if necessary

lamblin: As documented [here](http://deeplearning.net/software/theano_versions/dev/extending/extending_theano_gpu.html), the best practice is to infer the context during `make_node`, not to force the context when creating the Op.
Please remove the `context_name` attribute.

Sentient07:
In line:467 of `theano/gpuarray/type.py`, if i make a call to transfer instead of `host_from_gpu`, it gives , 
`RuntimeError: maximum recursion depth exceeded`

theano/gpuarray/extra_ops.py
Sentient07: Yeah, i'll remove that commit.

Sentient07:
In this PR, i've removed `ShapeOptimizer` from `fast_compile`. There are some optimizations that need `ShapeOptimizer` and is a part of `fast_compile`. I'm removing those optimizations from `fast_compile`. Yet to open that PR of removing `ShapeOptimizer`. @lamblin I believe that PR should get merged before this ?

theano/gpuarray/opt.py
```diff
@@ -55,19 +60,43 @@
...
+gpu_seqopt.register('gpuarray_graph_optimization', GraphToGPUDB(), -0.5,
+                    'fast_compile', 'fast_run', 'gpuarray')
+
 gpu_seqopt.register('gpuarray_local_optimiziations', gpu_optimizer, 1,
                     'fast_compile', 'fast_run', 'gpuarray')
 gpu_seqopt.register('gpuarray_cut_transfers', gpu_cut_copies, 2,
                     'fast_compile', 'fast_run', 'gpuarray')
 
+gpu_seqopt.register('op_lifter_topo', TopoOptimizer(gpu_topo.query('+fast_compile'), order='out_to_in'),
+                    10, 'fast_run', 'gpuarray', 'fast_compile')
```
lamblin: Use something between -0.5 and 1 instead of 10 here.

lamblin:
It seems to make sense for a first draft of the backward pass (using TopoOptimizer and LocalGroupDB).
When you change the priorities so it is executed before the EquilibriumOptimizer, we can see in the profiling if there is something to gain, and if we need to optimize the implementation further.
For instance, maybe we would need to implement tracks in LocalGroupDB, like in EquilibriumDB.

Sentient07:
> For instance, maybe we would need to implement tracks in LocalGroupDB, like in EquilibriumDB.

Could you please elaborate on this ?

lamblin:
EquilibriumDB has "tracks", or ops to match when trying to apply its local optimizers. For instance, some local optimizers will only be tried on nodes corresponding to some Ops. It uses the Ops passed to the `local_optimizer` decorator.
For instance, if a local opt registered with `@local_optimizer([GpuFromHost, GpuToGpu, HostFromGpu])`, then the EquilibriumOptimizer will only try to apply it on nodes where `node.owner.op` is an instance of those types.
I'm not sure if the LocalGroup uses this information. If it does not, then we may need to implement that logic, like in EquilibriumOptimizer.

Sentient07:
From what I see,LocalGroup does not use that info. LocalOptimizer uses track but I don't think LocalGroup does. I'll confirm again  once.

Sentient07:
@nouiz The new optimizer `op_lifter_topo` doesn't show all the local optimizers as `gpu_optimizer`. Am I doing the query wrong ?

Sentient07:
We're registering `theano.tensor.opt.local_remove_all_assert` into `gpu_optimizer` separately. Don't we need to do the same for `gpu_optimizer2` ?

theano/gpuarray/extra_ops.py
```diff
@@ -22,8 +22,9 @@ class GpuCumsum(GpuKernelBase, Op):
     SUPPORTED_NDIMS = 3
     __props__ = ('axis',)
 
-    def __init__(self, axis):
```
nouiz: Why this change? Just remove this commit, or explain why it is needed.

Sentient07: oh! This should have gotten removed from the rebase. It existed before, but i have changed this behaviour in #4570 itself

Sentient07: Yeah, this diff doesn't exist anymore. It has been removed. Earlier, i thought it was better to pass context_name

theano/gpuarray/opt.py
```diff
@@ -56,18 +61,43 @@
...
 
+gpu_seqopt.register('gpuarray_graph_optimization', GraphToGPUDB(), -0.5,
+                    'fast_compile', 'fast_run', 'gpuarray')
+
 gpu_seqopt.register('gpuarray_local_optimiziations', gpu_optimizer, 1,
                     'fast_compile', 'fast_run', 'gpuarray')
 gpu_seqopt.register('gpuarray_cut_transfers', gpu_cut_copies, 2,
                     'fast_compile', 'fast_run', 'gpuarray')
 
+gpu_seqopt.register('op_lifter_topo', TopoOptimizer(gpu_topo.query('+fast_compile', '+fast_run', '+cudnn',
```
Sentient07: @nouiz By registering it here, i won't have all the local optimizers in the `gpu_topo` db registered into `gpu_seqopt`?

Sentient07: Also, is this is the correct way to retrieve all the local optimizers in that DB

theano/gpuarray/basic_ops.py
```diff
@@ -647,14 +647,6 @@ def c_code_cache_version(self):
         return (9,)
 
 
-# Caching GPUAlloc
```
nouiz: You removed the caching, this is good, but to be useful, you should implement the method Op.\__new__() that do the caching automatically based on \__props__. Then we should check all subclass of Op to check its props, as from memory, it will cause some problems.

Can you move that into its own PR and start by implementing the \__new__ method?

Sentient07: Yeah, i've done that. But should that be a separate PR ?

Sentient07:
There are few redundant commits. #128f257 and #3855671 (Did in #4570 but with a different name),  	#5bbbb31 did as a different PR. Shall I squash them ?

lamblin:
We will need to the the automatic cache (in Op's metaclass) first, and then go on with the cleaning up.
We will not be able to do that for 0.9, so bumping to 0.10

Sentient07:
Closing as continued in https://github.com/Theano/Theano/pull/5579

Issue #4651, opened 2016-06-19
Improvements to theano.var


dwf:
There are two things that I think would improve `theano.var`:

* [ ] Add an option to use a more stable variant, namely the "corrected two-pass" algorithm described [here](http://www.cs.yale.edu/publications/techreports/tr222.pdf). PR gh-4681
* [x] Add the `ddof` argument from `numpy.var` and `numpy.std` to `tensor.var` and `tensor.std`.

dwf:
Yes, that's pretty much it. For backwards compatibility it's probably not good to add it by default (also not sure how much of a speed hit it is).

dwf:
Also, bad @nouiz, you're on vacation 😛

nouiz:
Actually, I'm staring my vacation only tomorrow!

kumarkrishna:
Hi @nouiz 
Could you mention what you meant by embedding in the Theano graph? I could update my PR to cover the ```ddof``` parameter as well.

nouiz:
there is already someone that started doing the ddof, so to prevent duplicate work, let him do it:) By embedding in the Theano graph, I mean this can be added in the Theano graph (as you did for the corrected two-pass).

I let @ballasn review your PR and the one for ddof.

thanks

nouiz:
all is done.

PR #4681, opened 2016-07-03
Uses corrected two pass algorithm in theano.tensor.var

kumarkrishna:
Reference #4651 
Adds argument ```corrected``` to maintain backward compatibility.


slefrancois:
test this please
we're testing a jenkins bot

theano/tensor/basic.py
```diff
@@ -3167,6 +3170,12 @@ def var(input, axis=None, keepdims=False):
 
     # return the mean sqr
     v = mean((centered_input ** 2), axis, keepdims=keepdims)
+
+    # use 'corrected_two_pass' algorithm
```
ballasn: you should add a reference to the tech report http://www.cs.yale.edu/publications/techreports/tr222.pdf

theano/tensor/tests/test_basic.py
```diff
@@ -6332,6 +6332,8 @@ def test_var():
     f = function([a], var(a, axis=2))
     assert numpy.allclose(numpy.var(a_val, axis=2), f(a_val))
 
+    f = function([a], var(a, corrected=True))
+    assert numpy.allclose(numpy.var(a_val), f(a_val))
```
ballasn: I don't think numpy.var uses the corrected two pass algorithm so you probably to implement the corrected two path algorithm as well for the test.

nouiz:
@kumarkrishna Can you finish this PR? I think not much is left and it would be great to close this.

thanks.

kumarkrishna:
@nouiz Updated the PR with more tests, fixed merge conflicts. I have also updated the implementation to take into consideration the ```ddof``` parameter. I think the patch is ready for review and merge.

theano/tensor/basic.py
```diff
@@ -3179,11 +3179,16 @@ def var(input, axis=None, ddof=0, keepdims=False):
...
         the result will broadcast correctly against the original tensor.
+    corrected : bool
+        If this is set to True, the 'corrected_two_pass' algorithm is
+        used to compute the variance.
+        Refer : http://www.cs.yale.edu/publications/techreports/tr222.pdf
 
     Notes
     -----
-    It uses the two-pass algorithm for more stable results.
+    By default, uses the two-pass algorithm for more stable results.
```
nouiz: It don't use it by default. Tell that it optionally support that.

theano/tensor/var.py
```diff
@@ -643,9 +643,10 @@ def mean(self, axis=None, dtype=None, keepdims=False, acc_dtype=None):
                                         dtype=dtype, keepdims=keepdims,
                                         acc_dtype=acc_dtype)
 
-    def var(self, axis=None, keepdims=False):
+    def var(self, axis=None, keepdims=False, corrected=False):
```
nouiz: add the ddof parameter at the same time here.

theano/tensor/basic.py
kumarkrishna: > By default, uses the two-pass algorithm

This refers to the usual two-pass algorithm used default for calculating the variance, not the corrected version. I will rewrite to make it more clear.

ballasn:
pr looks good to me, @nouiz  I think we can merge thanks,

ballasn:
Merged, thanks!

PR #4690, opened 2016-07-05
Remove shape opt

Sentient07:
In this PR. I have removed ShapeOptimizer from FAST_COMPILE and have fixed the tests that has dependency over the optimization done by ShapeFeature by Skipping them when the mode is FAST_COMPILE


Sentient07:
I've fixed almost all the tests that are failing because of this optimization, except these three, 

```
FAIL: test_local_sum_prod_alloc (theano.tensor.tests.test_opt.T_local_sum_prod)
  File "/Users/Ramana/projects/theano/Theano/theano/tensor/tests/test_opt.py", line 5434, in test_local_sum_prod_alloc
    assert len(f.maker.fgraph.apply_nodes) == nb_nodes[1]
AssertionError

FAIL: test_nested_composite (theano.tensor.tests.test_opt.TestCompositeCodegen)
  File "/Users/Ramana/projects/theano/Theano/theano/tensor/tests/test_opt.py", line 1551, in test_nested_composite
    assert len(f.maker.fgraph.toposort()) == 1
AssertionError

FAIL: test_nested_gpu (theano.tensor.tests.test_opt.TestCompositeCodegen)
  File "/Users/Ramana/projects/theano/Theano/theano/tensor/tests/test_opt.py", line 1568, in test_nested_gpu
    assert len(topo) == 2
AssertionError

Ran 200 tests in 72.592s
```

Sentient07:
These are the optimizers that make use of `shape_feature`, 
```
local_elemwise_alloc_op
local_shape_to_shape_i
local_track_shape_i
local_useless_inc_subtensor
local_useless_subtensor
local_subtensor_merge
local_adv_sub1_adv_inc_sub1
local_useless_inc_subtensor_alloc
```
I've Skipped all the tests that make use of these optimizers(the OPs they try to optimize). But the above reported tests (which gives error and haven't been Skipped) don't seem to use them.

Sentient07:
The `test_nested_gpu` doesn’t pass because I need more than one GPU? or because that Elemwise Op is being used by local_elemwise_alloc_op which uses shape_feature ?

Sentient07:
IIRC, @nouiz told me this would be merged after #4570 as one of the cleanup task.

Sentient07:
The profiling result over this branch doesn't show `ShapeFeature` and it gives ~2sec of speed up.(for 11 layered net) 
The profiling result on this branchhttp://pastebin.com/C4sNtRkj

Sentient07:
From the pattern that i observe in tests, i see that all the tests that makes a call to `theano.tests.unittest_tools.py:InferShapeTester` 's  `_compile_and_check` method fails as it tests the `infer_shape`. So, could i just add these two lines at the start of the `_compile_and_check` method ?

```
        if self.is_fast_compile:
            raise SkipTest("FAST_COMPILE does not do shape optimization")
```

Sentient07:
Also, all the `tests_infer_shape` tests fail as it is dependent on Shape Optimization. For that, I think the better way is to raise an error from `test_infer_shape` method of `InferShapeTester` class (after creating one). @nouiz is this a correct way to go about ?

Sentient07:
Except the tests that i've reported to fail in the comments, all the tests are passing locally. 
@nouiz are you able to reproduce those errors ?

Sentient07:
I've resolved conflicts and travis has restarted. Since it's a been a while i've worked on this PR, I'll wait for travis build before fixing anything

PR #4691, opened 2016-07-05
CorrMM upcasts mixed input dtypes fixes #4687

### LINKED ISSUE #4687: CorrMM: upcast inputs?
When `CorrMM` receives inputs of mixed types (single and double precision), it fails at runtime:
```python
import theano
import theano.tensor as T
x = T.ftensor4()
y = T.dtensor4()
z = T.nnet.corr.CorrMM()(x, y)
fn = theano.function([x, y], z)
fn(np.random.randn(2,3,4,5).astype(np.float32), np.random.randn(1,2,3,4))
```
This fails with `ValueError: CorrMM received weight with wrong type.`, from here: https://github.com/Theano/Theano/blob/a668c6c/theano/tensor/nnet/corr_gemm.c#L130
`CorrMM`, `CorrMM_gradWeights` and `CorrMM_gradInputs` should probably upcast their input tensors to a common datatype in `make_node`, or at least check for matching types early instead of failing at runtime.

Problem reported at https://groups.google.com/d/topic/lasagne-users/I5u9RHXTHmI/discussion.

JesseLivezey:


theano/tensor/nnet/corr.py
```diff
@@ -75,6 +75,34 @@ def __str__(self):
...
+        else:
+            out1 = in1
+
+        if dtype_2 != dtype_o:
+            out2 = in2.astype(dtype_o)
+        else:
+            out2 = in2
+
+        return out1, out2
+
```
f0k: I think this could be reduced to:
```python
def cast(self, in1, in2):
    dtype = theano.scalar.upcast(in1.dtype, in2.dtype)
    return in1.astype(dtype), in2.astype(dtype)
```
`astype` calls `T.cast`, which returns the variable unchanged if it's the correct type (in contrast to numpy, where `.astype` always creates a copy).

JesseLivezey:
@f0k, that is nicer. Pushed an update.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -259,6 +260,40 @@ def test_wrong_input(self):
...
+                    if a_dtype == 'float32':
+                        a_tens = T.ftensor4()
+                    else:
+                        c_dtype = 'float64'
+                        a_tens = T.dtensor4()
+                    if b_dtype == 'float32':
+                        b_tens = T.ftensor4()
+                    else:
+                        c_dtype = 'float64'
+                        b_tens = T.dtensor4()
```
nouiz: you can simplify the code by doing:

    b_tens = T.tensor4(dtype=b_dtype)

Idem at other places.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -259,6 +260,40 @@ def test_wrong_input(self):
         self.assertRaises(Exception, self.validate, (3, 2, 8, 8), (4, 2, 5, 5),
                           'valid', input=T.dtensor3())
 
+    def dtype_upcast(self):
```
nouiz: rename to test_dtype_upcast to have it executed by nosetests.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -259,6 +260,40 @@ def test_wrong_input(self):
...
+            return r * 2 - 1
+
+        ops = [corr.CorrMM, corr.CorrMM_gradWeights, corr.CorrMM_gradInputs]
+        a_shapes = [[4, 5, 6, 3], [1, 5, 6, 3], [1, 5, 6, 3]]
+        b_shapes = [[7, 5, 3, 2], [1, 5, 3, 1], [7, 1, 3, 1]]
+        dtypes = ['float32', 'float64']
+
+        for op, a_shape, b_shape in zip(ops, a_shapes, b_shapes):
+            for a_dtype in dtypes:
+                c_dtype = 'float32'
```
nouiz: If you revert the order of the list dtypes, then this line isn't at the right place. To make this a little big more secure and more readable, can you move that inline just after the next for loop?

nouiz:
Thanks, I have just a few small comments.

theano/tensor/nnet/tests/test_corr.py
JesseLivezey: I switched this to use theano.scalar.upcast() inside of both loops.

JesseLivezey:
@nouiz, I made the requested changes.

JesseLivezey:
I can rebase into 1 commit after travis runs.

nouiz:
closing/re-opening to trigger travis with the head of Theano.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -259,6 +260,32 @@ def test_wrong_input(self):
...
+            for a_dtype in dtypes:
+                for b_dtype in dtypes:
+                    c_dtype = theano.scalar.upcast(a_dtype, b_dtype)
+                    a_tens = T.tensor4(dtype=a_dtype)
+                    b_tens = T.tensor4(dtype=b_dtype)
+                    a_tens_val = rand(a_shape, dtype=a_dtype)
+                    b_tens_val = rand(b_shape, dtype=b_dtype)
+
+                    c_tens = op()(a_tens, b_tens)
+                    f = theano.function([a_tens, b_tens], c_tens)
```
nouiz: Add the parameter mode=self.mode. Without this, travis is failing with an error when mode=FAST_COMPILE.

Otherwise, I think it is fine. I'll kill the restarted travis.

JesseLivezey:
@nouiz, I updated the mode. Should I rebase everything before running the tests again?

f0k:
Huh? Is this a scary new artificial intelligence or was somebody logged in as the bot?

Issue #4680, opened 2016-07-02
dnn.dnn_conv3d but filter_dilation is only 3D


trungnt13:
I think the new implementation of filter_dilation messed up with dnn_conv3d.
This is the full traceback (subsample is 3D but filter_dilation is only 2D):

```
Traceback (most recent call last):
  File "/Users/trungnt13/tmp/tmp6.py", line 18, in <module>
    y = f1(x)
  File "/Users/trungnt13/libs/odin/odin/nnet/base.py", line 235, in __call__
    return self.apply(*args, **kwargs)
  File "/Users/trungnt13/libs/odin/odin/nnet/base.py", line 224, in apply
    out = self._apply(*args, **kwargs)
  File "/Users/trungnt13/libs/odin/odin/nnet/conv.py", line 159, in _apply
    conved = self.convolve(x)
  File "/Users/trungnt13/libs/odin/odin/nnet/conv.py", line 545, in convolve
    filter_shape=K.get_shape(self.W))
  File "/Users/trungnt13/libs/odin/odin/backend/theano/tensor.py", line 1364, in conv3d
    conv_mode='conv')
  File "/Users/trungnt13/libs/conda/envs/ai/lib/python2.7/site-packages/theano/sandbox/cuda/dnn.py", line 1283, in dnn_conv3d
    desc_op.subsample)
  File "/Users/trungnt13/libs/conda/envs/ai/lib/python2.7/site-packages/theano/sandbox/cuda/dnn.py", line 600, in get_out_shape
    subsample)
  File "/Users/trungnt13/libs/conda/envs/ai/lib/python2.7/site-packages/theano/tensor/nnet/abstract_conv.py", line 76, in get_conv_output_shape
    subsample[i], filter_dilation[i]) for i in range(len(subsample)))
  File "/Users/trungnt13/libs/conda/envs/ai/lib/python2.7/site-packages/theano/tensor/nnet/abstract_conv.py", line 76, in <genexpr>
    subsample[i], filter_dilation[i]) for i in range(len(subsample)))
IndexError: tuple index out of range
```

Sentient07:
This issue has been addressed in #4570 . To be a little more precise, this is a part of this [commit](https://github.com/Theano/Theano/pull/4570/commits/3e71f63a1a07cc1175f8e143246644e873682a83)

trungnt13:
```
Traceback (most recent call last):
  File "/Users/trungnt13/tmp/tmp3.py", line 19, in <module>
[...]

PR #4696, opened 2016-07-06
Fixed few errors generated from Dilation PR

Sentient07:
Fixes the error reported in #4680. And, a small optimization

fix gh-4680

### DIFF
--- a/theano/gpuarray/tests/test_abstractconv.py
+++ b/theano/gpuarray/tests/test_abstractconv.py
@@ -26,7 +26,7 @@ class TestDnnConv2d(test_abstract_conv.BaseTestConv2d):
             raise SkipTest(dnn_available.msg)
         mode = mode_with_gpu
 
-        o = self.get_output_shape(i, f, s, b, (1,1))
+        o = self.get_output_shape(i, f, s, b, (1, 1))
         self.run_fwd(inputs_shape=i, filters_shape=f, subsample=s,
                      verify_grad=True, mode=mode,
                      provide_shape=provide_shape, border_mode=b,
--- a/theano/gpuarray/tests/test_abstractconv.py
+++ b/theano/gpuarray/tests/test_abstractconv.py
@@ -21,12 +21,13 @@ class TestDnnConv2d(test_abstract_conv.BaseTestConv2d):
         # provide_shape is not used by the cuDNN impementation
         self.provide_shape = [False]
 
-    def tcase(self, i, f, s, b, flip, provide_shape):
+    def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
         if not dnn_available(test_ctx_name):
             raise SkipTest(dnn_available.msg)
         mode = mode_with_gpu
-
-        o = self.get_output_shape(i, f, s, b, (1, 1))
+        if fd != (1, 1):
+            raise SkipTest("Doesn't have CUDNN implementation")
+        o = self.get_output_shape(i, f, s, b, fd)
         self.run_fwd(inputs_shape=i, filters_shape=f, subsample=s,
                      verify_grad=True, mode=mode,
                      provide_shape=provide_shape, border_mode=b,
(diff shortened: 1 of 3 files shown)


theano/gpuarray/tests/test_abstractconv.py
```diff
@@ -26,7 +26,7 @@ def tcase(self, i, f, s, b, flip, provide_shape):
             raise SkipTest(dnn_available.msg)
         mode = mode_with_gpu
 
-        o = self.get_output_shape(i, f, s, b)
+        o = self.get_output_shape(i, f, s, b, (1,1))
```
nouiz: pep8: (1, 1)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -66,6 +66,8 @@ def get_conv_output_shape(image_shape, kernel_shape,
     """
     bsize, imshp = image_shape[0], image_shape[2:]
     nkern, kshp = kernel_shape[0], kernel_shape[2:]
+    if not filter_dilation:
+        filter_dilation = numpy.ones(numpy.asarray(subsample).shape, dtype='int')
```
nouiz: why not use: numpy.ones(len(subsample), dtype='int')

theano/gpuarray/tests/test_abstractconv.py
Sentient07: Again, I forgot to run flake tests! 😞

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -296,7 +296,7 @@ def test_all(self):
                     for b in self.border_modes:
                         try:
                             self.tcase(i, f, s, db, dflip,
-                                       dprovide_shape, fd)
+                                       dprovide_shape)
```
nouiz: This is the wrong fix. It will skip all filters_dilation tests. Maybe we need to find the implementation of tcase that don't support this case and extend it.

Sentient07: Yeah I thought so. I've explained why i did that here, https://github.com/Theano/Theano/pull/4587#issuecomment-230591403

Sentient07: Sorry, here , https://github.com/Theano/Theano/pull/4587#issuecomment-230591403

nouiz: The fix is in the file gpuarray/tests/tests_abstract_conv.py. in the method tcase(), add the parameter fd. Raise skip test if "fd != (1, 1)".

slefrancois:
jenkins test this please
we're testing a new CI bot

Sentient07:
Updated this PR. @nouiz if this is fine, i'll rebase #4570

theano/tensor/nnet/abstract_conv.py
```diff
@@ -66,6 +66,8 @@ def get_conv_output_shape(image_shape, kernel_shape,
     """
     bsize, imshp = image_shape[0], image_shape[2:]
     nkern, kshp = kernel_shape[0], kernel_shape[2:]
+    if not filter_dilation:
```
nouiz: Change this condition to "if filter_dilation is not None", in case someone put 0, I think it will get converted to a tuple with 0s in them. Just to be sure to not cause different error for 0s handling.

nouiz:
ok to merge when travis pass.

nouiz:
jenkins please test this

Sentient07:
@nouiz Jenkins tests are also run on GPU ?

nouiz:
not yet, but this is our goal.

On Thu, Jul 7, 2016 at 3:35 PM, Ramana Subramanyam <notifications@github.com

nouiz:
ok to test

Sentient07:
So the bot understands the comment and replies ? 😮 That sounds brilliant!

Sentient07:
I'll also give it a try, 
jenkins test this please

niasla:
Thanks for the fix, I was out of town !

Issue #3990, opened 2016-02-03
Add tests for conv2d_grad_* methods


lamblin:
Follow-up to #3987.

lamblin:
Check for correctness, the simplest would be to check against a gradient derived by calling `theano.grad` on a graph containing a forward convolution. Test with:
- valid / full
- filters_flip: True / False (mode: conv / corr)
- with subsampling: different input shapes that correspond to the same output shape (in terms of the original, forward conv)

lamblin:
Fixed in #5554.

PR #4711, opened 2016-07-08
Added tests for conv2d_grad_* methods

shabanian:
#3990

@adri-romsor
@lamblin

### DIFF
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -7,7 +7,7 @@ from nose.tools import assert_raises, assert_true
 
 import theano
 from theano import tensor
-from theano.gof.opt import check_stack_trace
+import theano.tensor as T
 from theano.tests import unittest_tools as utt
 from theano.tensor.nnet import corr, abstract_conv as conv
 from theano.tensor.nnet.abstract_conv import get_conv_output_shape
@@ -17,6 +17,8 @@ from theano.tensor.nnet.abstract_conv import AbstractConv2d_gradWeights
 from theano.tensor.nnet.abstract_conv import bilinear_kernel_1D
 from theano.tensor.nnet.abstract_conv import bilinear_kernel_2D
 from theano.tensor.nnet.abstract_conv import bilinear_upsampling
+from theano.tensor.nnet.abstract_conv import conv2d_grad_wrt_weights
+from theano.tensor.nnet.abstract_conv import conv2d_grad_wrt_inputs
 from theano.tensor.nnet.conv import ConvOp
 from theano.tensor.nnet.corr import (CorrMM, CorrMM_gradWeights,
                                      CorrMM_gradInputs)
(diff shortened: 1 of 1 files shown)


adri-romsor:
@nouiz

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+def test_conv2d_grad_wrt_inputs():
+
+    inputs_shapes = [(8, 1, 12, 12), (8, 1, 18, 18), (2, 1, 4, 4),
+                     (6, 1, 10, 11), (2, 1, 6, 5), (1, 5, 9, 9)]
+    filters_shapes = [(5, 1, 2, 2), (4, 1, 3, 3), (2, 1, 3, 3),
+                      (1, 1, 2, 5), (4, 1, 2, 2), (4, 5, 2, 2)]
+    subsamples = [(1, 1), (2, 2)]
+    border_modes = ["valid", "full"]
+    filter_flip = [True, False]
+    x = T.cast(T.ftensor4('x'), theano.config.floatX)
```
lamblin: Why is that cast needed?
Can you just declare `x = T.tensor4('x')`, or, if you want to be explicit, `x = T.tensor4('x', dtype=theano.config.floatX)`?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
         f_2D = theano.function([], mat_2D, mode=self.compile_mode)
         utt.assert_allclose(f_1D(), f_2D(), rtol=1e-06)
+
+
+def test_conv2d_grad_wrt_inputs():
+
+    inputs_shapes = [(8, 1, 12, 12), (8, 1, 18, 18), (2, 1, 4, 4),
+                     (6, 1, 10, 11), (2, 1, 6, 5), (1, 5, 9, 9)]
+    filters_shapes = [(5, 1, 2, 2), (4, 1, 3, 3), (2, 1, 3, 3),
+                      (1, 1, 2, 5), (4, 1, 2, 2), (4, 5, 2, 2)]
```
lamblin: You do not have to test all of these shapes, but it would be good to have different (in_shape, fltr_shape) combinations that give the same output shape when ss=(2, 2).
For instance, if `fltr_shape = (1, 1, 3, 3)`, then `in_shape = (1, 1, 5, 5)`, `(1, 1, 5, 6)` and `(1, 1, 6, 6)` will all give an output shape of `(1, 1, 2, 2)` in valid mode when ss = (2, 2).

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+    subsamples = [(1, 1), (2, 2)]
+    border_modes = ["valid", "full"]
+    filter_flip = [True, False]
+    x = T.cast(T.ftensor4('x'), theano.config.floatX)
+    for (in_shape, fltr_shape) in zip(inputs_shapes, filters_shapes):
+        input_val = np.random.rand(*in_shape).astype(theano.config.floatX)
```
lamblin: Instead of using `np.random`, can you explicitly allocate a numpy RandomStream with the default test seed (`utt.fetch_seed()`)?
See http://deeplearning.net/software/theano_versions/dev/extending/unittest.html#using-random-values-in-test-cases for an example.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+
+        filter_val = np.random.rand(*fltr_shape).astype(theano.config.floatX)
+        for bm in border_modes:
+            for ss in subsamples:
+                for ff in filter_flip:
+                    cgi = conv2d_grad_wrt_inputs(output_grad=input_val,
```
lamblin: It does not look correct to specify `output_grad=input_val` and `input_shape=in_shape`.
`in_shape` is supposed to be the shape of the input of the original (forward) convolution, and `output_grad` should have the same shape as the output of that forward convolution.

One option would be to compute the shape of the output for each case (either by hand or by using `get_conv_output_shape`), and sample `output_grad_val` from something of that size.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+                                                 filters=filter_val,
+                                                 filter_shape=fltr_shape,
+                                                 input_shape=in_shape,
+                                                 border_mode=bm,
+                                                 subsample=ss,
+                                                 filter_flip=ff)
```
lamblin: It would be better to use symbolic (or shared) variables for `input` (or `output_grad`) and `filters`, so constant-folding does not get in the way during optimization.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+
+                    conv_func = T.nnet.conv.conv2d(x,
+                                                   filters=filter_val,
+                                                   border_mode=bm,
+                                                   filter_shape=fltr_shape,
+                                                   image_shape=in_shape,
+                                                   subsample=ss)
+
+                    f_prime = theano.grad(conv_func.sum(), x)
```
lamblin: Taking the gradient of the sum will be the equivalent of computing the gradient with `output_grad` containing all ones.
What would be better, and would actually allow testing with the same random values as the other function, would be to explicitly provide a tensor for the output gradient, something like:
```python
output_grad = T.tensor4()
filters = T.tensor4()
conv_out = T.nnet.conv.conv2d(x, filters=filters, ...)

grad_wrt_i = theano.grad(wrt=[x], known_grads={conv_out: output_grad})
f_prime = theano.function([x, output_grad, filters], grad_wrt_i)
```
Note that you will need to provide `output_grad` as an input to `f_prime`, because the shape of `grad_wrt_i` will depend on it.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+                    f = theano.function([x], f_prime)
+                    T.allclose(cgi, f(input_val))
```
lamblin: `T.allclose` here does not actually perform a comparison. It just builds a symbolic expression that represents the (approximate) equality between `cgi` (a symbolic variable) and `f(input_val)` (a constant).
What you want instead is another Theano function computing a value for `cgi` (given `output_grad` and `filters`), and calling `utt.assert_allclose(..., ...)` on the value returned by that function and by f (`f_prime` if you follow my code earlier).

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -781,3 +783,77 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+
+
+def test_conv2d_grad_wrt_weights():
```
lamblin: The same remarks apply for that test as well.

lamblin:
If you want to check that your tests correctly detect the problem, you can try to revert the following change, and see if your tests fail: https://github.com/Theano/Theano/pull/3987/files

nouiz:
Just a note. I merged the fix in: https://github.com/Theano/Theano/pull/4911

This mean that in travis, the tests should pass. But if you don't rebase the PR and run your tests, they should fail locally.

thanks

lamblin:
Superseded by #5554.

PR #4712, opened 2016-07-10
Decode bytes to string before split

Morgan243:
The current elemwise_openmp_speedup.py will output the following error when used with Python 3.5.1:
```
  File "lib/python3.5/site-packages/theano/misc/elemwise_openmp_speedup.py", line 35, in <module>
    (cheapTime, costlyTime) = runScript(N=options.N)
  File "lib/python3.5/site-packages/theano/misc/elemwise_openmp_speedup.py", line 26, in runScript
    return list(map(float, out.split(" ")))
TypeError: a bytes-like object is required, not 'str'
```

Quick testing shows this to be the expected exception when calling split() on a bytes object. Calling decode() on the bytes object fixes this for Python 3, and seems to still work fine in my Python 2.7.11 install.

### DIFF
--- a/theano/misc/elemwise_openmp_speedup.py
+++ b/theano/misc/elemwise_openmp_speedup.py
@@ -23,7 +23,7 @@ def runScript(N):
     if err:
         print(err)
         sys.exit()
-    return list(map(float, out.split(" ")))
+    return list(map(float, out.decode().split(" ")))
 
 if __name__ == '__main__':
     options, arguments = parser.parse_args(sys.argv)
--- a/theano/misc/elemwise_openmp_speedup.py
+++ b/theano/misc/elemwise_openmp_speedup.py
@@ -3,8 +3,12 @@ import os
 import subprocess
 import sys
 from optparse import OptionParser
+from locale import getpreferredencoding
 
 import theano
+from theano.compat import decode_with
+
+console_encoding = getpreferredencoding()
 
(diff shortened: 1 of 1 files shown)


theano/misc/elemwise_openmp_speedup.py
```diff
@@ -23,7 +23,7 @@ def runScript(N):
     if err:
         print(err)
         sys.exit()
-    return list(map(float, out.split(" ")))
+    return list(map(float, out.decode().split(" ")))
```
nouiz: thanks for looking into that. In the file theano/sandbox/cuda/nvcc_compiler.py, we do something like:

from locale import getpreferredencoding
from theano.compat import decode, decode_with
nvcc_stdout_raw, nvcc_stderr_raw = p.communicate()[:2]
console_encoding = getpreferredencoding()
nvcc_stdout = decode_with(nvcc_stdout_raw, console_encoding)
nvcc_stderr = decode_with(nvcc_stderr_raw, console_encoding)

This also allow to work with different language and we know it work with python 2.6 that we still support (but not for long). Can you update your PR to follow this more tested pattern?

thanks

Morgan243:
A quick note, the variable name 'dir' was stepping on the builtin dir() function, so I renamed it to 'path'.

nouiz:
thanks

Issue #4666, opened 2016-06-25
CudaNdarray shared variable in uncompiled graph


sisp:
Hi,

When I set ``device=gpu`` and create a shared variable using ``theano.shared``, I immediately get a ``CudaNdarray`` object. When I create a graph containing shared variables, then only the shared variable is a GPU variable plus there are ``HostFromGPU`` and vice versa ops, but all other ops and variables are default CPU ones before compiling the function. Is this desired behavior and if so, why? I feel the graph should be a raw CPU graph during construction and then the graph optimizer converts all ops to GPU ops.

If I create shared variables using ``theano.shared(..., target='cpu')`` and compile the function, then the graph is converted to contain only GPU ops and variables including the shared variable that I had initially forced to be a CPU shared variable.

I've encountered this problem of having GPU shared variables before compilation because I would like to train a model on the GPU but pickle the graph and the shared values of the trained shared variables as NumPy arrays. Later I would like to be able to load the original graph (without GPU ops at the beginning) and then load the trained values into the shared variables, but this does not work if I want to compile the graph for the **CPU** if it already contains GPU shared variables.

I appreciate your comments on this.

Thanks,
Sigurd

f0k:
> When I create a graph containing shared variables, then only the shared variable is a GPU variable plus there are `FromHostToGPU` and vice versa ops, but all other ops and variables are default CPU ones before compiling the function. Is this desired behavior and if so, why?

At least in the past, this was the desired behavior -- the GPU variables are used to seed all graph optimizations that move computation to GPU. I.e., there are many optimizations of the kind "SomeOp(HostFromGpu(x)) -> HostFromGpu(SomeGpuOp(x))". These optimizations gradually push everything to the GPU until there's only the final HostFromGpu to transfer the end result to CPU memory.

Not sure if this still applies to the new backend and the intended refactoring of how computations are moved to GPU, so far I can only speak for the Theano I grew up with.

Sentient07:
Hello, 
This is the intended behaviour in the new backend as well. When you're using the GPU, with device=cuda(for the new backend) or with device=gpu(for the old backend), shared variable will be of the type GpuArrayType. (I am not aware if there is any plan to change this). But, the symbolic variables aren't moved to the GPU.
[...]

PR #4717, opened 2016-07-11
Move shared variables of graph to specific device

f0k:
This adds `theano.gof.utils.move_shared_vars()`, a function to clone a graph while moving all contained shared variables to a particular target device or replacing them by constants, as discussed in #4666. The idea was that pickling a graph would become easier if moving everything to CPU first, so it can be unpickled regardless of GPU availability and moved to GPU afterwards.

Caveats: When moving a graph from GPU to CPU, all shared variables that were previously wrapped in a `HostFromGpu` will be moved to CPU and then wrapped in a `GpuFromHost -> HostFromGpu`. Not sure if it's worth detecting this -- it would require checking the clients of all shared variables if `target="cpu"`. The redundancy is resolved on compilation anyway.

### DIFF
--- a/theano/gof/utils.py
+++ b/theano/gof/utils.py
@@ -551,3 +551,41 @@ def hash_from_dict(d):
             second_part += [v]
     tuple_items = tuple(first_part + second_part + [d.__class__])
     return hash(tuple_items)
+
+
+def move_shared_vars(graph, target='cpu', allow_borrow=True):
+    """
+    Clones a graph, moving all contained shared variables to a particular
+    target device or replacing them by constants.
+
+    Parameters
+    ----------
+    graph : Theano expression or list of Theano expressions
+        The computational graph to clone, represented by its output expressions
+    target : str
+        The target device for shared variables, or ``"constant"`` to replace
+        by constants
+    allow_borrow : bool
+        If True, allow (but not guarantee) newly created variables to share
+        storage with the original variables if residing on the same device.
+
+    Returns
+    -------
+    Theano expression or list of Theano expressions
+        The cloned graph with all shared variables replaced by copies on the
+        given target device, or by constants
+    """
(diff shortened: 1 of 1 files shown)


sisp:
Thanks for this great feature! About ``GpuFromHost -> HostFromGpu``: Can a pickled graph with these transfer ops be unpickled on a machine without GPUs? That's important because we will need to deploy on a machine without a CUDA-capable GPU.

f0k:
> Can a pickled graph with these transfer ops be unpickled on a machine without GPUs?

That would be good to know. Can you try it?

abergeron:
It won't work with no gpu present for HostFromGpu.  So we should problably run some of the optimizations to clean up the graph and remove those patterns.

This could be done later though.

f0k:
> It won't work with no gpu present for HostFromGpu.

Thanks!

> This could be done later though.

Well, one could steer the replacement so the HostFromGpu(SharedVar) gets replaced by the new SharedVar if the target is "cpu" or "constant" (target='dev...' can never be on CPU, right?). Finding the HostFromGpu's is a little annoying, but could be integrated in the first graph traversal for finding the shared variables.
What happens if a graph is moved from dev0 to dev1? Could this introduce other redundant transfer Ops, and how can we find out in advance?

In any case, I guess this function should return a clean graph, either by being more smart or by removing redundant transfers afterwards.

sisp:
I agree , the graph should be clean after applying this function. Is there a way to reuse the optimizer that removes redundant transfer ops?

f0k:
@abergon: How would we clean up the graph afterwards? Do we need to go the full route of creating a `FunctionGraph`, instantiating a `TopoOptimizer` with the correct local optimizations and running it? Do you see any easier solution for having the function return a clean graph?

PR #4719, opened 2016-07-11
Improve tensor var

Faruk-Ahmed:
" #4651

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3099,7 +3099,7 @@ def mean(input, axis=None, dtype=None, op=False, keepdims=False,
 
 
 @constructor
-def var(input, axis=None, keepdims=False):
+def var(input, axis=None, ddof=0, keepdims=False):
     """
     Computes the variance along the given axis(es) of a tensor `input`.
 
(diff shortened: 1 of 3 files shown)


theano/tensor/basic.py
```diff
@@ -3139,7 +3141,9 @@ def var(input, axis=None, keepdims=False):
     centered_input = input - mean_input
 
     # return the mean sqr
-    v = mean((centered_input ** 2), axis, keepdims=keepdims)
+    #divisor = theano.tensor.cast(prod(input.shape[axis]) - ddof, theano.config.floatX)
```
nouiz: check if ddof is True or False. In that case, raise an error and tell that we changed the parameter order. This will help prevent user error due to the parameter order change.

theano/tensor/basic.py
```diff
@@ -3139,7 +3141,9 @@ def var(input, axis=None, keepdims=False):
...
+    divisor = prod(input.shape[axis]) - ddof
```
nouiz: if ddof is 0, don't do the add. Do something like:
divisor = prod()
if ddof:
    divisor = divisor - ddof

nouiz: Also, for consistency and prevent strange difference, we are better to handle this the same way mean is doing:

    for i in axis:
        s = true_div(s, shp[i])

Or we should change mean to use the same way of doing it. The advantage of the loop, is if there is only 1 axis, we build a smaller graph. (but I think opt will clean it up).

ballasn:
there are still some flake8 issues to fix,

nouiz:
continued in https://github.com/Theano/Theano/pull/4852

Issue #3243, opened 2015-08-04
GetItem2d.grad isn't implemented


nouiz:
This need a new op to keep the grad sparse.

vaibhav4595:
I am facing a similar problem
AttributeError: 'GetItem2d' object has no attribute 'grad'

Is there a fix for this particular issue ?

nouiz:
As the comment above tell, this request the creation of a new op to implement the grad in that case. We don't have plan to implement it as we don't use it. If you are ready to implement it, we can guide you and this could get merged in Theano. Just tell it.

vaibhav4595:
I would actually like to implement as this would benefit the work that I have been doing.
How could I get started with this ?

nouiz:
I would recommand to go in 2 steps. 1, make something that work and give
the good results. 2, optimize this.

For the first step, start by reading

http://deeplearning.net/software/theano/extending/extending_theano.html#extending-theano

Then make a new op that implement the grad of GetItem2d. Mostly, for the
dense case, this is implemented by IncSubtensor. GetItem2d is simpler.
Mostly, this is creating a 2d sparse matrix with all zeros of the same size
as the input of GetItem2d, then in add the gradient to the input subset
that was selected.

Then implement the method GetItem2d.grad and GetItem2d.connection_pattern
(this one shoudl return [[True], [False]] in that case or [True, False]
from memory)

vaibhav4595:
I'll go through with the reading first and then get back to you

vaibhav4595:
I have gone through the link and I have also looked at the Op's contract as well. I understand that each **op** would need a **grad** function in order to compute the overall gradient for the **Cost**. The points I would like some clarity are:
- What part of the gradient does the op.grad actually compute ? (I understand the chain rule procedure but I am unable to understand what exactly is passed into the arguments of op.grad and what it needs to compute and return ?)
-  Would it be feasible for you to provide a very simple example stating what are the exact values that are passed onto the op.grad function in its arguments by the gradient.grad function, and what exactly does the op.grad compute and return ? (I couldn't get a very clear picture from the Op's contract for this and the example provided in that) Maybe in terms of variables if you could state that, "okay this is what is computed in the gradient.grad and then such and such values are passed to op.grad, where it computes such and such a thing and returns".

vaibhav4595:
[...]

PR #4723, opened 2016-07-12
GetItem2d.grad implemented, fix for #3243

vaibhav4595:
Two functions have been added for the GetItem2d class

- grad
- connection_pattern

### DIFF
--- a/theano/sparse/basic.py
+++ b/theano/sparse/basic.py
@@ -1346,6 +1346,23 @@ class GetItem2d(gof.op.Op):
         assert _is_sparse(x)
         out[0] = x[start1:stop1:step1, start2:stop2:step2]
 
+    def grad(self, inputs, grads):
+        x = inputs[0]
+        rest = inputs[1:]
+        gz, = grads
+
+        assert _is_sparse_variable(x) == True
+
+        return [x.zeros_like() + gz] + [DisconnectedType()()] * len(rest)
+
+    def connection_pattern(self, node):
+        rval = [[True]]
+
+        for ipt in node.inputs[1:]:
+            rval.append([False])
+
+        return rval
+
 get_item_2d = GetItem2d()
 """
 Implement a subtensor of sparse variable, returning a sparse matrix.


theano/sparse/basic.py
```diff
@@ -1346,6 +1346,23 @@ def perform(self, node, inputs, outputs):
...
         out[0] = x[start1:stop1:step1, start2:stop2:step2]
 
+    def grad(self, inputs, grads):
+        x = inputs[0]
+        rest = inputs[1:]
+        gz, = grads
+
+        assert _is_sparse_variable(x) == True
+
+        return [x.zeros_like() + gz] + [DisconnectedType()()] * len(rest)
```
nouiz: gz and x.zeros_like() don't have the same shape. gz have the shape of the output of the forword of this op.

Doing the correct computation will request a new op that take 3 inputs:

x.zeros_like(), gz and index.

Then it will have a perform method that will do:
x[index] += gz

or

x[index] = gz

Also, check in the doc, they tell how to test the grad. It will help you know when it is right. If you have questions, don't hesitate.

vaibhav4595:
@nouiz the inputs that are passed to the grad function has x (which is the sparse variable), start1, stop1, step1 and correspondingly there is start2, stop2 and step2. The start, stop and step correspond to the indexing part of the variable.
The steps and stops can be a Constant{None} type. 
How do we handle the index when parts of it has Constant{None} type, because steps, stops and starts can be of that type ?

theano/sparse/basic.py
nouiz: I mean, make a new op like IncSubtensor, but for this Op and use it in this op grad like Subtensor.grad.

Mostly, do like Subtensor/IncSubtensor, but on sparse data. But don't forget that Subtensor is much more complicated then this op, so don't try to replicate everything from Subtensor.

nouiz:
For the None, in this Op make_node method we handle it.

We have a special Theano Variable for None and we can make Symbolic None constant. Mostly, use this to create the symbolic None constant:

 generic_None = theano.gof.Constant(theano.gof.generic, None)

davrandom:
I'd like this feature to be merged.

PR #4726, opened 2016-07-12
Fix some problems in the buildbot

abergeron:


nouiz:
ok to merge when travis pass.

PR #4728, opened 2016-07-12
Self caching

Sentient07:
This PR is about caching of different Ops, done from the Op class.

### DIFF
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -789,12 +789,15 @@ class Op(utils.object2, PureOp, CLinkerOp):
         if args:
             [key.append(i) for i in args]
         if kwargs:
-            [key.append(v) for k,v in kwargs.iteritems()]
+            [key.append(v) for k, v in kwargs.iteritems()]
 
         # not the best solution, a hack for now
         for k in key:
-            if isinstance(k, collections.Hashable):
-               key.remove(k) 
+            if not isinstance(k, collections.Hashable):
+                new_obj = object.__new__(cls)
+                print ("We Require the keys of" + cls.__name__ +
+                       "to be of Hashable type")
+                return new_obj
 
         # Since list is not hashable
         key = tuple(key)
@@ -804,8 +807,8 @@ class Op(utils.object2, PureOp, CLinkerOp):
                 for a in args:
                     Op.instances[key].a = a
             if kwargs:
-                for k,v in kwargs.iteritems():
-                    Op.instances[key].v = v
+                for k, v in kwargs.iteritems():
+                    Op.instances[key].k = v
         obj = Op.instances[key]
         if not hasattr(obj, '_op_use_c_code'):
             obj._op_use_c_code = theano.config.cxx
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -476,6 +476,36 @@ class CLinkerOp(CLinkerObject):
                                      self.__class__.__name__)
 
 
+class SingletonOp(type):
+
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        key = []
+        # This will be fully revamped with _props_dict
+        if args:
+            for a in args:
+                key.append(a)
+        if kwargs:
+            for k, v in kwargs.iteritems():
+                key.append(v)
+
+        for k in key:
+            if not isinstance(k, collections.Hashable):
+                return object.__new__(cls).__init__(*args, **kwargs)
+
+        key = tuple(key)
+
+        if key not in cls._instances:
+            cls._instances[key] = object.__new__(cls)
+
+        cls._instances[key].__init__(*args, **kwargs)
+
+        if not hasattr(cls._instances[key], '_op_use_c_code'):
+            cls._instances[key]._op_use_c_code = theano.config.cxx
+        return cls._instances[key]
+
+
 class PureOp(object):
     """
     An :term:`Op` is a type of operation.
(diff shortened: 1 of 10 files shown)


theano/gof/op.py
```diff
@@ -779,10 +780,33 @@ class Op(utils.object2, PureOp, CLinkerOp):
...
+            [key.append(v) for k,v in kwargs.iteritems()]
+
+        # not the best solution, a hack for now
+        for k in key:
+            if isinstance(k, collections.Hashable):
+               key.remove(k) 
+
+        # Since list is not hashable
+        key = tuple(key)
+        if key not in Op.instances:
```
Sentient07: All the objects in the key are not Hashable. What to do in such cases ? Do i ignore and generate a new instance of the class and return ?

Sentient07: Small edit, this comment is for line 797

theano/gof/op.py
```diff
@@ -779,10 +780,33 @@ class Op(utils.object2, PureOp, CLinkerOp):
...
+            Op.instances[key] = object.__new__(cls)
+            if args:
+                for a in args:
+                    Op.instances[key].a = a
+            if kwargs:
+                for k,v in kwargs.iteritems():
+                    Op.instances[key].v = v
```
Sentient07: this should be `Op.instances[key].v = v` . I made this change locally

theano/gof/op.py
nouiz: Yes, but print a warning. We want keys and values of props to be hashable. There was 1 excetion I think, but I forgot which one. Can you tell me which one don't hash?

theano/gof/op.py
```diff
@@ -779,10 +780,33 @@ class Op(utils.object2, PureOp, CLinkerOp):
...
-        obj = object.__new__(cls)
+        key = list()
+        if args:
+            [key.append(i) for i in args]
+        if kwargs:
+            [key.append(v) for k,v in kwargs.iteritems()]
+
+        # not the best solution, a hack for now
+        for k in key:
+            if isinstance(k, collections.Hashable):
```
Sentient07: Another silly error.  should have been `not isinstance`. It is like that in my local file though.

nouiz:
killing travis as you will make follow up commit in this PR.

theano/gof/op.py
```diff
@@ -779,10 +780,33 @@ class Op(utils.object2, PureOp, CLinkerOp):
...
+
     def __new__(cls, *args, **kwargs):
         # this function exists to silently and transparently ensure that all
         # existing Ops get a _op_use_c_code attribute
-        obj = object.__new__(cls)
+        key = list()
+        if args:
+            [key.append(i) for i in args]
+        if kwargs:
+            [key.append(v) for k,v in kwargs.iteritems()]
```
nouiz: The key, value is the props, not what it received as paremeter to \__new__

nouiz: In fact, I forgot what args and kwargs would be here. Are they what will be passed to \__init__? If so, then we can't use them. We need to get the \__props__. Can you make a break point here and check if you can access the props?

theano/gof/op.py
Sentient07: It is the `Elemwise` Op that has seem to have a dict param (`{0: 0}`). I guess it is the `scalar_op` param ?

Sentient07: Yeah, i remember reporting this previously when i was trying to hash GpuElemwise. You recollect ? And I think that param could be anything (Any Op from `theano.scalar.basic`) and not only a dict as i reported above/

theano/gof/op.py
Sentient07: All the param passed to the __init__ will be recieved here as well. So, `(*args, **kwargs)` is a more generalized way. 

> We need to get the `__props__`.

Meaning, iterating through cls.__props__ ?

nouiz:
I talked with @abergeron about this. We don't think it can be done by just implementing Op.__new__. We see 2 options:

1) using mataclass. Arnaud could do it in a short time, but this is pretty advanced stuff.
2) Using a decorator on a python function like:
class MyOp():
   ...
@cache_instance
def my_op(...):
   return MyOp(...)

2) isn't complicated, but would ask that we go over all Op in Theano (and possibly outside). This isn't great and would take much longer then 1). So I think for now we should way to see if @abergeron can implement 1) next week.

nouiz:
killing travis as this isn't the way to go.

Sentient07:
@abergeron If I understand correctly, we'd like to make the `class Op` a singleton.  I've a basic idea of how to go about this. If i'm correct with few correctable mistakes, I'd like to proceed by myself on this. 
```
class ParentOp(type):
    _instances = {}
    # This is done in generalized way of *args and **kwargs
    def __call__(cls, *args, **kwargs):
        if cls not in cls._instances:
            # Later, this check could be done using cls._props_dict() 
            cls._instances[cls] = super(ParentOp, cls).__call__(*args, **kwargs)
        return cls._instances[cls]
 
# The existing Op class in theano
class Op():
    __metaclass__ = ParentOp
```
Before i go any further, i'm confused as to whether it should be `PureOp` or `Op` class ? (I guess it should be the Op that should be having the metaclass.
Also, regarding the complexity of handling parameters as @nouiz said, i think after implementing `__props__`  consistently, that concern wouldn't be there either I guess ?

nouiz:
this work is continued in https://github.com/Theano/Theano/pull/5579, so closing.

Issue #1563, opened 2013-10-15
LogSumExp Optimisation


mgermain:
It could be nice to have a numerically more stable LogSumExp built-in.
Instead of doing T.log(T.sum(T.exp(x))) we could use something like LogSumExp(x).
```
def LogSumExp(x):
    x_max = T.max(x)
    T.log(T.sum(T.exp(x - x_max))) + x_max 
```
Or even better with axis support.
```
def LogSumExp(x, axis=None):
    x_max = T.max(x, axis=axis, keepdims=True)
    return T.log(T.sum(T.exp(x - x_max), axis=axis, keepdims=True)) + x_max
```

nouiz:
This implementation is already in Pylearn: 

http://deeplearning.net/software/pylearn2/library/expr.html?highlight=log_sum_exp#pylearn2.expr.basic.log_sum_exp

There is a PR that add test for it: https://github.com/lisa-lab/pylearn2/pull/1390

Maybe we can just port it to Theano.

mgermain:
I might be wrong but as I understand the scope of Theano, I think It would
make sense to have something basic like that in it instead of Pylearn.

abergeron:
Yes, we want an optimization for that in theano.  This is the point of ticket #2082.

nouiz:
Agreed, we just need someone to do it.
Le 16 févr. 2015 16:03, "Mathieu Germain" <notifications@github.com> a
écrit :

mgermain:
"an optimization for that in theano"

Is there an example I could look at to learn how it should be implemented ?

NeilGirdhar:
+1 This would be very useful.

lamblin:
Closed by #4736

PR #4736, opened 2016-07-13
Log sum exp optimization for numerical stability

adbrebs:
Ticket #1563, @mgermain

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -6085,6 +6085,52 @@ def local_log_add(node):
                 return [ret]
 
 
+@register_stabilize
+@register_specialize
+@gof.local_optimizer([T.log])
+def local_log_sum_exp(node):
+    # log(sum_i(exp(x_i))) = x_max + log(sum_i(exp(x_i - x_max)))
+
+    if node.op != T.log:
+        return
+
+    sum_node = node.inputs[0].owner
+    # If the sum has keepdims=True, there might be a dimshuffle
+    if sum_node and isinstance(sum_node.op, T.DimShuffle):
+        sum_node = sum_node.inputs[0].owner
+
+    if not sum_node or not isinstance(sum_node.op, T.Sum):
+        return
+
+    exp_node, axis = sum_node.inputs[0].owner, sum_node.op.axis
+    if not exp_node and exp_node.op != T.exp:
+        return
+
+    pre_exp = exp_node.inputs[0]
+    # optimisation may have already been applied
+    if (pre_exp.owner and
+            isinstance(pre_exp.owner.op, T.Elemwise) and
+            pre_exp.owner.op.scalar_op == scalar.sub):
+        max_node = pre_exp.owner.inputs[1].owner
+        if max_node and isinstance(max_node.op, T.DimShuffle):
+            max_node = max_node.inputs[0].owner
+        if not isinstance(max_node.op, T.MaxAndArgmax):
+            return
+        if max_node.inputs[0] == pre_exp.owner.inputs[0]:
+            return
+
+    max_pre_keepdims = T.max(pre_exp, axis=axis, keepdims=True)
+
+    ret = (max_pre_keepdims + T.log(T.sum(T.exp(pre_exp - max_pre_keepdims),
+                                          axis=axis, keepdims=True)))
+
+    # Restore shape and broadcastable pattern
+    ret = T.reshape(ret, node.inputs[0].shape)
+    ret = T.patternbroadcast(ret, node.inputs[0].broadcastable)
+
+    return [ret]
+
+
 def add_calculate(num, denum, aslist=False, out_type=None):
     # TODO: make sure that this function and mul_calculate are similar
     if out_type is None:
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -6095,7 +6095,10 @@ def local_log_sum_exp(node):
     sum_node = node.inputs[0].owner
     # If the sum has keepdims=True, there might be a dimshuffle
     if sum_node and isinstance(sum_node.op, T.DimShuffle):
+        dimshuffle_op = sum_node.op
         sum_node = sum_node.inputs[0].owner
+    else:
+        dimshuffle_op = None
 
     if not sum_node or not isinstance(sum_node.op, T.Sum):
         return
@@ -6107,14 +6110,15 @@ def local_log_sum_exp(node):
         return
 
     pre_exp = exp_node.inputs[0]
-    max_pre_keepdims = T.max(pre_exp, axis=axis, keepdims=True)
+    max_pre_exp = T.max(pre_exp, axis=axis)
+    max_pre_exp_keepdims = T.makeKeepDims(pre_exp, max_pre_exp, axis)
 
-    ret = (max_pre_keepdims + T.log(T.sum(T.exp(pre_exp - max_pre_keepdims),
-                                          axis=axis, keepdims=True)))
+    ret = (max_pre_exp +
+           T.log(T.sum(T.exp(pre_exp - max_pre_exp_keepdims), axis=axis)))
 
-    # Restore shape and broadcastable pattern
-    ret = T.reshape(ret, node.inputs[0].shape)
-    ret = T.patternbroadcast(ret, node.inputs[0].broadcastable)
+    # Restore the dimshuffle op, if any.
+    if dimshuffle_op:
+        ret = dimshuffle_op(ret)
 
     return [ret]
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -6085,8 +6085,6 @@ def local_log_add(node):
                 return [ret]
 
 
-@register_stabilize
-@register_specialize
 @gof.local_optimizer([T.log])
 def local_log_sum_exp(node):
     # log(sum_i(exp(x_i))) = x_max + log(sum_i(exp(x_i - x_max)))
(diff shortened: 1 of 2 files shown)


theano/tensor/tests/test_opt.py
```diff
@@ -6498,6 +6498,32 @@ def test_local_merge_alloc():
...
 
 
+def test_local_log_sum_exp1():
+    # Tests if optimization is applied
+    x = matrix('x')
+    y = T.log(T.sum(T.exp(x), axis=1))
+    MODE = theano.compile.get_default_mode().including('local_log_sum_exp')
+    f = function([x], y, mode=MODE)
+
+    assert (f.maker.fgraph.toposort()[1].op.scalar_op ==
```
mgermain: Doing a list comprehension to verify that there is at least one `maximum`might be more robust to future changes.

theano/tensor/tests/test_opt.py
```diff
@@ -6498,6 +6498,46 @@ def test_local_merge_alloc():
...
+
+    x_val = 1.0 + numpy.random.rand(4, 3).astype(config.floatX)/10.0
+
+    naive_ret = numpy.log(numpy.sum(numpy.exp(x_val), axis=1))
+    optimised_ret = f(x_val)
+
+    assert numpy.allclose(naive_ret, optimised_ret)
+
+
+def test_local_log_sum_exp3():
```
mgermain: You should just find some value that causes numerical instability and test against the result of scipy.logsumexp.
Make sure to skip the test if scipy is not available.

theano/tensor/tests/test_opt.py
```diff
@@ -6498,6 +6498,46 @@ def test_local_merge_alloc():
...
+    y = T.log(T.sum(T.exp(x), axis=1))
+    MODE = theano.compile.get_default_mode().including('local_log_sum_exp')
+    f = function([x], y, mode=MODE)
+
+    assert (theano.scalar.basic.maximum
+            in [node.op.scalar_op for node in f.maker.fgraph.toposort()])
+
+
+def test_local_log_sum_exp2():
+    # Tests if the optimization works (result is correct) around 1.0
```
mgermain: Not sure what you want to test exactly with this one.

theano/tensor/tests/test_opt.py
adbrebs: test2 checks if it still computes the right thing for standard (not extreme) values.

adbrebs: test3 checks if it still computes the right thing for an extreme value (100), for which the naive way fails.

theano/tensor/tests/test_opt.py
adbrebs: test1 just checks that the optimization has been applied (if the opt is applied, then the graph should contain a maximum node).

adbrebs:
One test fails with FAST_COMPILE flag.

adbrebs:
@mgermain @lamblin Any idea why the test with FAST_COMPILE fails?

nouiz:
I don't. Can you check what is the resulting graph? Can you verify the optimization get applied? Maybe it need other optimization to be enabled before the new optimization can work.

adbrebs:
``` python
a = T.matrix()
b = T.log(T.sum(T.exp(a), axis=1))
f = theano.function([a], b)
theano.printing.debugprint(f.maker.fgraph.outputs[0])
```

```
Reshape{1} [id A] ''
 |Elemwise{Composite{(i0 + log(i1))}}[(0, 0)] [id B] ''
 | |InplaceDimShuffle{0,x} [id C] ''
 | | |Reduce{maximum}{1} [id D] 'max'
 | |   |<TensorType(float64, matrix)> [id E]
 | |InplaceDimShuffle{0,x} [id F] ''
 |   |Sum{axis=[1], acc_dtype=float64} [id G] ''
 |     |Elemwise{Composite{exp((i0 - i1))}} [id H] ''
 |       |<TensorType(float64, matrix)> [id E]
 |       |InplaceDimShuffle{0,x} [id C] ''
 |MakeVector{dtype='int64'} [id I] ''
   |Shape_i{0} [id J] ''
     |<TensorType(float64, matrix)> [id E]
```

nouiz:
You can use the Theno flags optimizer_verbose=True on that test in
mode=FAST_RUN and mode=FAST_COMPILE and compare the sequence of
optimization being applied.

This will help understand what is going on.

Is the graph there the graph you are supposed to optimize?

adbrebs:
> Is the graph there the graph you are supposed to optimize?

yes

In mode mode=FAST_RUN, the optimization is applied as expected. In mode=FAST_COMPILE, it is not applied and verbose does not show anything.

nouiz:
Put a break point in the new optimization at the start. That way you will know if the problem is that it isn't called or if it is that the pattern it look for isn't there.

theano/tensor/opt.py
```diff
@@ -5901,6 +5901,52 @@ def local_log_add(node):
...
+@gof.local_optimizer([T.log])
+def local_log_sum_exp(node):
+    # log(sum_i(exp(x_i))) = x_max + log(sum_i(exp(x_i - x_max)))
+
+    if node.op != T.log:
+        return
+
+    sum_node = node.inputs[0].owner
+    # If the sum has keepdims=True, there might be a dimshuffle
+    if sum_node and isinstance(sum_node.op, T.DimShuffle):
```
nouiz: This could introduce a bug. Maybe the DimShuffle is there to swap axes. Add tests with keepdims and where you don't use it but swap axes.

adbrebs:
> That way you will know if the problem is that it isn't called or if it is that the pattern it look for isn't there.

I have checked and It isn't called in FAST_COMPILE.

theano/tensor/opt.py
adbrebs: Yes I agree, there might be problems due to that. I guess the best solution would be to keep the DimShuffle op and re-apply it in the new graph.

theano/tensor/opt.py
```diff
@@ -5901,6 +5901,52 @@ def local_log_add(node):
...
+        sum_node = sum_node.inputs[0].owner
+
+    if not sum_node or not isinstance(sum_node.op, T.Sum):
+        return
+
+    exp_node, axis = sum_node.inputs[0].owner, sum_node.op.axis
+    if not exp_node and exp_node.op != T.exp:
```
lamblin: I think it could be matched in a more general way, something like `isinstance(exp_node.op, Elemwise) and isinstance(exp_node.op.scalar_op, scalar.Exp)`

theano/tensor/opt.py
```diff
@@ -5901,6 +5901,52 @@ def local_log_add(node):
...
+        return
+
+    pre_exp = exp_node.inputs[0]
+    # optimisation may have already been applied
+    if (pre_exp.owner and
+            isinstance(pre_exp.owner.op, T.Elemwise) and
+            pre_exp.owner.op.scalar_op == scalar.sub):
```
lamblin: Same here, `isinstance(pre_exp.owner.op.scalar_op, scalar.Sub)`

adbrebs:
@lamblin tests fail, have I done the right thing?

adbrebs:
@lamblin Tests passed, ready for final review.

lamblin:
Jenkins, test this, please.

theano/tensor/opt.py
```diff
@@ -6085,6 +6085,45 @@ def local_log_add(node):
...
+        return
+
+    pre_exp = exp_node.inputs[0]
+    max_pre_keepdims = T.max(pre_exp, axis=axis, keepdims=True)
+
+    ret = (max_pre_keepdims + T.log(T.sum(T.exp(pre_exp - max_pre_keepdims),
+                                          axis=axis, keepdims=True)))
+
+    # Restore shape and broadcastable pattern
+    ret = T.reshape(ret, node.inputs[0].shape)
```
lamblin: This line is problematic:
- if the initial `sum_node.op` was Dimshuffle and it had actually transposed the sum, then a Reshape does not do the right thing, we should apply a Dimshuffle again instead
- depending on `node.inputs[0]` is a bad idea, because it re-introduces part of the old graph in the new one.

theano/tensor/opt.py
```diff
@@ -6085,6 +6085,45 @@ def local_log_add(node):
...
+    exp_node, axis = sum_node.inputs[0].owner, sum_node.op.axis
+    if not exp_node or not (
+            isinstance(exp_node.op, Elemwise) and
+            isinstance(exp_node.op.scalar_op, scalar.Exp)):
+        return
+
+    pre_exp = exp_node.inputs[0]
+    max_pre_keepdims = T.max(pre_exp, axis=axis, keepdims=True)
+
+    ret = (max_pre_keepdims + T.log(T.sum(T.exp(pre_exp - max_pre_keepdims),
```
lamblin: Actually, why use `max_pre_keepdims` here? If the initial sum did not have `keepdims=True`, then this add the dimension back, which is not what we want.

theano/tensor/tests/test_opt.py
```diff
@@ -6656,6 +6656,54 @@ def test_local_useless_alloc():
     assert isinstance(topo[-1].op, T.Alloc)
 
 
+def test_local_log_sum_exp1():
+    # Tests if optimization is applied
+    x = matrix('x')
+    y = T.log(T.sum(T.exp(x), axis=1))
```
lamblin: We should test with:
- axes other than the last
- multiple axes
- sum with and without `keepdims`
- more dimensions than 2, and a dimshuffle after the sum, for instance `T.log(T.sum(T.exp(t3), axis=1).T)`

And make sure the tests pass with DebugMode.

adbrebs:
Thanks @lamblin for the review, I have made corrections. Tests pass locally with DebugMode.

theano/tensor/tests/test_opt.py
```diff
@@ -6676,27 +6682,47 @@ def test_local_log_sum_exp1():
...
+
+    # If a transpose is applied to the sum
+    transpose_op = DimShuffle((False, False), (1, 0))
+    check_max_log_sum_exp(x, axis=2, dimshuffle_op=transpose_op)
+
+    # If the sum is performed with keepdims=True
+    sum_keepdims_op = x.sum(axis=(0, 1), keepdims=True).owner.op
+    check_max_log_sum_exp(x, axis=(0, 1), dimshuffle_op=sum_keepdims_op)
+
+
```
lamblin: A test for the case above would be, for instance, `x = TensorType(dtype='floatX', broadcastable=(False, True, False)('x')`.

lamblin:
Thanks for the update, sorry I could not review it earlier.
There is a last small thing to change, after that I think we are good to merge.

theano/tensor/opt.py
```diff
@@ -6085,6 +6085,49 @@ def local_log_add(node):
...
+    # Restore the dimshuffle op, if any.
+    if dimshuffle_op:
+        ret = dimshuffle_op(ret)
+
+    return [ret]
+
+
+compile.optdb.register('local_log_sum_exp',
+                       in2out(local_log_sum_exp, ignore_newtrees=True),
+                       1.6, 'fast_run')
```
nouiz: I think you should also add 'stabilize' after fast_run

PR #4737, opened 2016-07-14
OpenCL support for Image2Neibs kernels

huanzhang12:
This commit adds OpenCL support for Images2Neibs class. The original Images2Neibs 
kernel was CUDA only, and this patch modifies the kernel so that it is 
compatible with both OpenCL and CUDA with the help of libgpuarray backend.
Also the originally hard-code max threads per block (512) was replaced by
reading the actual max threads per block from the device. On modern Nvidia cards
this number is 1024, so performance should actually improve a little bit.
All modifications shouldn't adversely affect the performance of the original kernel in any ways.

During the development of this OpenCL kernel, a bug of libgpuarray was found and 
was fixed in commit f8b8e156c. So you will need the latest libgpuarray to test the code.

I have tested my code on a AMD R9 280X and a Nvidia Tesla K40c. On the Nvidia GPU
I tested both OpenCL and CUDA backends and all tests pass.

```
$ THEANO_FLAGS=mode=FAST_RUN,device=opencl0:0,floatX=float32 bin/theano-nose -s theano/gpuarray/tests/test_neighbours.py
Mapped name None to device opencl0:0: Tahiti
............
Ran 12 tests in 16.755s

OK


$ THEANO_FLAGS=mode=FAST_RUN,device=opencl0:0,floatX=float32 bin/theano-nose -s theano/gpuarray/tests/test_neighbours.py
Mapped name None to device opencl0:0: Tesla K40c
............
Ran 12 tests in 8.684s

OK


$ THEANO_FLAGS=mode=FAST_RUN,device=cuda0,floatX=float32 bin/theano-nose -s theano/gpuarray/tests/test_neighbours.py
Mapped name None to device cuda0: Tesla K40c
............
Ran 12 tests in 20.342s

OK

``` 

EDIT: deleted an incorrect paragraph in my comment on test_neighbours.py, which was actually for test_extra_ops.py. I will make a PL on that later.

### DIFF
--- a/theano/gpuarray/neighbours.py
+++ b/theano/gpuarray/neighbours.py
@@ -68,48 +68,48 @@ class GpuImages2Neibs(GpuKernelBase, Images2Neibs, Op):
         code = """
 // a version that uses less registers but doesn't work in all cases.
         KERNEL void %(kname)s(
-            const int nb_batch,
-            const int nb_stack,
-            const int height,
-            const int width,
-            const int c,
-            const int d,
-            const int step_x,
-            const int step_y,
-            const int grid_c,
-            const int grid_d,
-            const size_t stride0, const size_t stride1,
-            const size_t stride2, const size_t stride3,
-            const %(type_ten4)s * global_ten4, const size_t offset_ten4,
-            const size_t out_s0, const size_t out_s1,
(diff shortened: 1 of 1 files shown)


nouiz:
thanks.

Issue #4739, opened 2016-07-14
Theano does not use CUDA flags for cuDNN


ruslanagit:
This is an issue, similar to the 2nd issue mentioned in #2928 (fixed in #3664), when compiler-bindir was not used. 
In my setup I use flags to compile with CUDA:
[nvcc]
flags = --use-local-env  --cl-version=2010

Though cuDNN was properly installed, it was not seen by theano. After digging in, I have seen that params in dnn_available() method (in file theano\sandbox\cuda\__init__.py) does not include nvcc flags. 
Fixed manually, by appending appropriate flags to params just after the lines 
if config.nvcc.compiler_bindir:
                params.extend(['--compiler-bindir',
                               config.nvcc.compiler_bindir])
cuDNN works fine with this fix.

Please fix on main branch.

nouiz:
Can you make a PR with the fix? This way you will get the credit for this
fix.

thanks.

ruslanagit:
This is me first PR. Not sure if done correctly...

PR #4743, opened 2016-07-14
Use nvcc flags to compile with cuDNN

ruslanagit:
Fix for issue #4739


theano/sandbox/cuda/__init__.py
```diff
@@ -304,6 +304,8 @@ def dnn_available():
             if config.nvcc.compiler_bindir:
                 params.extend(['--compiler-bindir',
                                config.nvcc.compiler_bindir])
+            [params.append(flag) for flag in config.nvcc.flags.split(' ') if flag]
```
nouiz: It is working. But I think this version is more clear:

    params.extend([flag for flag in config.nvcc.flags.split(' ') if flag])

thanks.

nouiz:
thanks

Issue #4555, opened 2016-05-26
gpuarray tests should pass when floatX=float32 and add this to the daily buildbot


nouiz:

mgermain:
@nouiz 
Just to make sure, this issue is twofold right?
1. Make sure that all the gpuarray tests pass when floatX=float32
2. Add all the gpuarray tests to the daily builbot

nouiz:
1. yes.
2. just to clarify, this is in the gpuarray daily buildbot. We have 2 daily
buildbot, 3 when the new one by @lefransi is done.

Fred

Sentient07:
@nouiz  If you're having a list of opts that does not pass the test I would like to know them so that I could add few if they are missing there

nouiz:
@abergeron will finish this.

nouiz:
It is added in the gpuarray daily buildbot in float32. So I'll close this issue. It have a few errors, but I'll let this to be fixed as current buildbot errors.

PR #4744, opened 2016-07-15
Fix a test failure in GpuCumsum class

huanzhang12:
If I run the unit test for the GpuCumsum class (using the CUDA backend) I get an assertion failure:

```
$ THEANO_FLAGS=mode=FAST_RUN,device=cuda0,floatX=float32 bin/theano-nose -s theano/gpuarray/tests/test_extra_ops.py
Mapped name None to device cuda0: Tesla K40c
.........F
FAIL: test_infer_shape (theano.gpuarray.tests.test_extra_ops.TestGpuCumsum)
  File "/home/hzhang12/projects/Theano/theano/tensor/tests/test_extra_ops.py", line 157, in test_infer_shape
    self.op_class)
  File "/home/hzhang12/projects/Theano/theano/tests/unittest_tools.py", line 249, in _compile_and_check
    assert any(isinstance(t.op, cls) for t in topo_out)
AssertionError

Ran 10 tests in 30.616s

FAILED (failures=1)
```

It seems in function InferShapeTester._compile_and_check(), it is checking if the operator class is derived from CumsumOp. Currently the GpuCumsum class is derived from class Op, not CumsumOp, so it fails. I am not sure if it is intentional. To make InferShapeTester happy, I changed the base class of GpuCumsum from Op to CumsumOp, and we can pass all the tests now.

If this is not the right approach, maybe we need to fix InferShapeTester._compile_and_check()? Or maybe it is a known issue?

I also noticed that in earlier versions of extra_ops.py, GpuCumsum was inherited from CumsumOp. But it is removed by [commit d87cf8b4](https://github.com/Theano/Theano/commit/d87cf8b458b626da14e548443424c722f0e89c61). @kelvinxu I am not sure why.

### DIFF
--- a/theano/gpuarray/extra_ops.py
+++ b/theano/gpuarray/extra_ops.py
@@ -12,7 +12,7 @@ from .basic_ops import (as_gpuarray_variable, GpuKernelBase, Kernel, GpuReshape)
 from .opt import register_opt, op_lifter, register_opt2
 
 
-class GpuCumsum(GpuKernelBase, Op):
+class GpuCumsum(GpuKernelBase, CumsumOp):
     """
     Parameters
     ----------


nouiz:
This isn't the right fix. I don't have time to investigate this error more. My guess is that we move too much or not enough the op to the GPU. This is a recent failure. Could you make a git bisect to find which PR introduced it?

huanzhang12:
The Travis CI build failure is just a flake8 issue. I think I can probably fix it. But if you say the fix isn't correct because  of some deeper issues with the Op class I don't think I am able to help here.

I tried to make a bisect but I found it very difficult, because earlier commits requires early versions of libgpuarray, and earlier versions of libgpuarry have a lot of compilation errors on my system. However, I found that in [commit d87cf8b4](https://github.com/Theano/Theano/commit/d87cf8b458b626da14e548443424c722f0e89c61) and [commit 32d3ce77](https://github.com/Theano/Theano/commit/32d3ce7754d0eeb6ebea147e55eb5baf7eb868b6) the base classes of GpuCumsum changed.

It could also be an issue of InferShapeTester, because it enforces that the Op needs to be the correct class, but it probably not work for Ops on GPU.

nouiz:
Just a note that the problem is the one in that issue:

https://github.com/Theano/Theano/issues/4555

Mostly, gpuarray tests haven't bee made to pass in floatX=float32.

nouiz:
solved elsewhere, so closing.

thanks for bringing that to our attention and starting the fix.

PR #4747, opened 2016-07-15
Small mixed changes.

nouiz:
fix gh-4759

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1472,7 +1472,7 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
     return [rval]
 
 
-@register_opt('fast_compile',  'conv_dnn', 'cudnn')
+@register_opt('fast_compile', 'conv_dnn', 'cudnn')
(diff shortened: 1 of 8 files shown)


theano/gpuarray/dnn.py
```diff
@@ -1472,18 +1472,31 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
...
+    return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+
+
+@register_opt('fast_compile', 'conv_dnn', 'cudnn')
+@local_optimizer([AbstractConv2d_gradInputs])
+def local_abstractconv_gi_cudnn(node):
+    ctx = infer_context_name(*node.inputs)
+    if not isinstance(node.inputs[0].type, GpuArrayType):
+        return
+    return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
```
abergeron: Why are you copying this code three times?

nouiz: It allow to get the profiling stats per op. It don't slow down compilation.

PR #4039, opened 2016-02-11
Making DimShuffles view by default 

Sentient07:
and adding `_props_dict` to few Ops in old backend

PR #4749, opened 2016-07-16
Adding __props__ to all the Ops that has __init__

Sentient07:
Currently Raising error from the `__new__` of the Op class so that we could get to know better what classes don't have **props** implemented. I ran few tests and added the `__props__` , but for the remaining, i'll fix it from the error that travis would give.

### DIFF
--- a/theano/compile/ops.py
+++ b/theano/compile/ops.py
@@ -518,6 +518,8 @@ class FromFunctionOp(gof.Op):
 
     """
 
+    __props__ = ("fn", "itypes", "otypes", "infer_shape")
+
     def __init__(self, fn, itypes, otypes, infer_shape):
         self.__fn = fn
         self.itypes = itypes
--- a/theano/compile/ops.py
+++ b/theano/compile/ops.py
@@ -518,8 +518,6 @@ class FromFunctionOp(gof.Op):
 
     """
 
-    __props__ = ("fn", "itypes", "otypes", "infer_shape")
-
     def __init__(self, fn, itypes, otypes, infer_shape):
         self.__fn = fn
         self.itypes = itypes
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -781,7 +781,6 @@ class Op(utils.object2, PureOp, CLinkerOp):
     Convenience class to bundle `PureOp` and `CLinkerOp`.
 
     """
-
     def prepare_node(self, node, storage_map, compute_map, impl):
         """
         Make any special modifications that the Op needs before doing
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -782,7 +782,6 @@ class Op(utils.object2, PureOp, CLinkerOp):
 
     """
 
-
     def prepare_node(self, node, storage_map, compute_map, impl):
         """
         Make any special modifications that the Op needs before doing
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -782,6 +782,7 @@ class Op(utils.object2, PureOp, CLinkerOp):
 
     """
 
+
     def prepare_node(self, node, storage_map, compute_map, impl):
         """
         Make any special modifications that the Op needs before doing
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -781,6 +781,7 @@ class Op(utils.object2, PureOp, CLinkerOp):
     Convenience class to bundle `PureOp` and `CLinkerOp`.
 
     """
+
     def prepare_node(self, node, storage_map, compute_map, impl):
         """
         Make any special modifications that the Op needs before doing
--- a/theano/gof/tests/test_destroyhandler.py
+++ b/theano/gof/tests/test_destroyhandler.py
@@ -48,10 +48,6 @@ def MyConstant(data):
 
 class MyOp(Op):
 
-    __props__ = ("nin", "name", "vmap", "dmap", "nout",
-                 "destroyhandler_tolerate_same",
-                 "destroyhandler_tolerate_aliased")
-
     def __init__(self, nin, name, vmap=None, dmap=None, nout=1,
                  destroyhandler_tolerate_same=None,
                  destroyhandler_tolerate_aliased=None):
--- a/theano/gof/tests/test_destroyhandler.py
+++ b/theano/gof/tests/test_destroyhandler.py
@@ -48,6 +48,10 @@ def MyConstant(data):
 
 class MyOp(Op):
 
+    __props__ = ("nin", "name", "vmap", "dmap", "nout",
+                 "destroyhandler_tolerate_same",
+                 "destroyhandler_tolerate_aliased")
+
     def __init__(self, nin, name, vmap=None, dmap=None, nout=1,
                  destroyhandler_tolerate_same=None,
                  destroyhandler_tolerate_aliased=None):
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -852,7 +852,6 @@ class ScalarOp(Op):
 
     nin = -1
     nout = 1
-
     __props__ = ("output_types_preference", "name")
 
     def __init__(self, output_types_preference=None, name=None):
@@ -1185,6 +1184,7 @@ isinf = IsInf()
 
 class InRange(LogicalComparison):
     nin = 3
+    __props__ = ("openlow", "openhi")
 
     def __init__(self, openlow, openhi):
         self.openlow = openlow
@@ -2036,6 +2036,8 @@ identity = Identity(same_out, name='identity')
 
 # CASTING OPERATIONS
 class Cast(UnaryScalarOp):
+    __props__ = ("o_type", "name")
+
     def __init__(self, o_type, name=None):
         if not isinstance(o_type, Scalar):
             raise TypeError(o_type)
@@ -3461,7 +3463,7 @@ class Composite(ScalarOp):
     Composite depends on all the Ops in its graph having C code.
 
     """
-    init_param = ('inputs', 'outputs')
+    __props__ = ('inputs', 'outputs')
 
     def __str__(self):
         if self.name is None:
@@ -3475,7 +3477,7 @@ class Composite(ScalarOp):
         This fct allow fix patch this.
 
         """
-        d = dict([(k, getattr(self, k)) for k in self.init_param])
+        d = dict([(k, getattr(self, k)) for k in self.__props__])
         out = self.__class__(**d)
         if name:
             out.name = name
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -853,6 +853,8 @@ class ScalarOp(Op):
     nin = -1
     nout = 1
 
+    __props__ = ("output_types_preference", "name")
+
     def __init__(self, output_types_preference=None, name=None):
         self.name = name
         if output_types_preference is not None:
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -852,7 +852,6 @@ class ScalarOp(Op):
 
     nin = -1
     nout = 1
-    __props__ = ("output_types_preference", "name")
 
     def __init__(self, output_types_preference=None, name=None):
         self.name = name
@@ -861,7 +860,7 @@ class ScalarOp(Op):
                 raise TypeError(
                     "Expected a callable for the 'output_types_preference' argument to %s. (got: %s)" %
                     (self.__class__, output_types_preference))
-        self.output_types_preference = output_types_preference
+            self.output_types_preference = output_types_preference
 
     def make_node(self, *inputs):
         if self.nin >= 0:
@@ -907,6 +906,16 @@ class ScalarOp(Op):
     def grad(self, inputs, output_gradients):
         raise utils.MethodNotDefined("grad", type(self),
                                      self.__class__.__name__)
+
+    def __eq__(self, other):
+        test = (type(self) == type(other) and
+                getattr(self, 'output_types_preference', None) ==
+                getattr(other, 'output_types_preference', None))
+        return test
+
+    def __hash__(self):
+        return hash(type(self).__name__) ^ hash(
+            getattr(self, 'output_types_preference', 0))
 
     def __str__(self):
         if hasattr(self, 'name') and self.name:
@@ -1174,7 +1183,6 @@ isinf = IsInf()
 
 class InRange(LogicalComparison):
     nin = 3
-    __props__ = ("openlow", "openhi")
 
     def __init__(self, openlow, openhi):
         self.openlow = openlow
@@ -2026,8 +2034,6 @@ identity = Identity(same_out, name='identity')
 
 # CASTING OPERATIONS
 class Cast(UnaryScalarOp):
-    __props__ = ("o_type", "name")
-
     def __init__(self, o_type, name=None):
         if not isinstance(o_type, Scalar):
             raise TypeError(o_type)
@@ -3453,7 +3459,7 @@ class Composite(ScalarOp):
     Composite depends on all the Ops in its graph having C code.
 
     """
-    __props__ = ('inputs', 'outputs')
+    init_param = ('inputs', 'outputs')
 
     def __str__(self):
         if self.name is None:
@@ -3467,7 +3473,7 @@ class Composite(ScalarOp):
         This fct allow fix patch this.
 
         """
-        d = dict([(k, getattr(self, k)) for k in self.__props__])
+        d = dict([(k, getattr(self, k)) for k in self.init_param])
         out = self.__class__(**d)
         if name:
             out.name = name
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -861,7 +861,7 @@ class ScalarOp(Op):
                 raise TypeError(
                     "Expected a callable for the 'output_types_preference' argument to %s. (got: %s)" %
                     (self.__class__, output_types_preference))
-            self.output_types_preference = output_types_preference
+        self.output_types_preference = output_types_preference
 
     def make_node(self, *inputs):
         if self.nin >= 0:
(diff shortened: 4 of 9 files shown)


theano/gof/op.py
```diff
@@ -783,6 +783,9 @@ def __new__(cls, *args, **kwargs):
         # this function exists to silently and transparently ensure that all
         # existing Ops get a _op_use_c_code attribute
         obj = object.__new__(cls)
+        # Ensuring all the Ops created henceforth has __props__
+        if not hasattr(obj, '__props__'):
+            raise
```
Sentient07: I don't know if i should keep this. This raises a lot of error, regarding few Ops in old backend and tests that don't implement `__props__`. Should i keep these two lines and add the `__props__` to the required places ?
I thought this check would be useful because, if there is an Op that is being created in future in theano, this would ensure that it has `__props__`.

nouiz: I'm not sure it is the right place for this. Did you check that __props__ is already there for some ops that already have it? I think it?

In all case, I don't want to make props mandatory for now. But if all ops in Theano have it, we could raise a warning here.

Sentient07: This one is consistent in raising the error when an Op's instance has `__init__` but not `__props__`. But until we're creating `__props__` for all the Op, it is wouldn't be possible to cache using a Metaclass (the details of which I've mailed you)

theano/scalar/basic.py
```diff
@@ -851,6 +851,8 @@ class ScalarOp(Op):
     nin = -1
     nout = 1
 
+    __props__ = ("output_types_preference", "name")
```
nouiz: The class InRange override the \__init__. So it have different props. Check if there is other subclass that override init

theano/tensor/elemwise.py
```diff
@@ -134,6 +134,7 @@ class DimShuffle(Op):
 
     _f16_ok = True
     check_input = False
+    __props__ = ("input_broadcastable", "new_order", "inplace")
```
nouiz: With this change, you can delete the \__eq__ method.

theano/tensor/elemwise.py
```diff
@@ -502,6 +503,7 @@ class Elemwise(OpenMPOp):
     | Elemwise(log)(rand(3, 4, 5))
 
     """
+    __props__ = ("scalar_op", "inplace_pattern", "name", "nfunc_spec", "openmp")
```
nouiz: inplace_pattern is a dict, so it won't work with the current code. For now, I think you should concentrate on the cgt optimization, so just remove this line ( and don't add a warning for op without props)

nouiz:
I got a few comments. Don't add more now as I think you should focus on cgt, but if you could finish this PR to get it merged, it would be good.

theano/tensor/elemwise.py
Sentient07: Yes, I ignore this class for caching. So, of the places that i've added `__props__` which ones should i retain and which one should i remove ? I did not make any change to the old backend. There are a couple of Op at least that do not have `__props__` Do i add there as well ?

theano/gof/op.py
nouiz: Just to be sure, do you mean we can't cache any ops with a metaclass without having all nodes implement props or you mean we will only cache ops with props?

The second is fine.

theano/tensor/elemwise.py
nouiz: \__props__ don't support dict, (event when ignoring the caching). So don't add it here until we fix the general \__props__ mechanism to support that.

theano/gof/op.py
Sentient07: While passing parameters, it becomes confusing and there are some errors at making call to `__call__`. I don't know if this could be fixed without requirement of `__props__` I would need @abergeron 's help. But with `__props__`,  I think i can fix it.

Sentient07: Also, by using props, we can cache without building in all the parameters from `*args` and `**kwargs` and rebuilding those parameters to initialize the new class instance. (In one comment you had asked me to use `__props__` for this purpose.

theano/scalar/basic.py
Sentient07: To be clear, all the sub classes that has `__init__` should also be having `__props__` right ?

theano/tensor/elemwise.py
Sentient07: The better solution would be, Except those classes from the condition to have `__props__` and not cache them also ?

nouiz: The right fix would be to support dict in props. I tried and it wasn't easy, so I let this to later. For this, we would need frozendict and this isn't in the default python. I'm not sure it is good to add extra dependencies. So just remove it.

Sentient07:
I added a new commit where I have added that check that you have asked me to remove in the TODO section. From your previous comment, I'll open a new issue. I think that using `_props_dict` we can implement the caching by default for all the Ops.

theano/gof/tests/test_destroyhandler.py
```diff
@@ -47,6 +47,10 @@ def MyConstant(data):
 
 class MyOp(Op):
 
+    __props__ = ("nin", "name", "vmap", "dmap", "nout",
+                 "destroyhandler_tolerate_same",
+                 "destroyhandler_tolerate_aliased")
+
```
Sentient07: I have committed this a couple of months back, but haven't pushed it so I didn't notice this in the PR. Do I revert it or keep it ?

theano/scalar/basic.py
nouiz: In an ideal case, yes. But this class have subclass that need different \__props__. So you should remove this one or add the props to all the subclass that need it. InRange is such a subclass, maybe there is others.

theano/gof/tests/test_destroyhandler.py
nouiz: revert.

nouiz:
if you want to add support of dict for props, do it in another PR do not delay this one.

thanks

theano/gof/tests/test_destroyhandler.py
Sentient07: This part caused the travis build fail. The error was, 
```
ERROR: theano.gof.tests.test_destroyhandler.test_misc
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/tests/test_destroyhandler.py", line 155, in test_misc
    PatternOptimizer((transpose_view, (transpose_view, 'x')), 'x').optimize(g)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 83, in optimize
    self.add_requirements(fgraph)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 2170, in add_requirements
    fgraph.attach_feature(toolbox.NodeFinder())
  File "/Users/Ramana/projects/theano/Theano/theano/gof/fg.py", line 564, in attach_feature
    attach(self)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/toolbox.py", line 439, in on_attach
    Bookkeeper.on_attach(self, fgraph)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/toolbox.py", line 124, in on_attach
    self.on_import(fgraph, node, "on_attach")
  File "/Users/Ramana/projects/theano/Theano/theano/gof/toolbox.py", line 464, in on_import
    raise e
AttributeError: 'MyOp' object has no attribute 'vmap'
```
I didn't see it in detail then. The object was of `dict` type and that's why probably.

Sentient07:
If you're (@nouiz ) okay with changing the `openmp` parameter of Elemwise from dict to frozen dict, I'll make that change and add `__props__` there as well. 
If we have `__props__` everywhere, the Singleton [here](https://github.com/Theano/Theano/pull/4728) can use the `_props_dict` to infer the parameter instead of how i've done it now.

Sentient07:
@nouiz I didn't see this error before, was it this you were referring to as something inside theano that you'd work on? 
```
FAIL: theano.tensor.tests.test_gc.test_gc_never_pickles_temporaries
  File "/home/travis/miniconda2/envs/pyenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/Sentient07/Theano/theano/tensor/tests/test_gc.py", line 90, in test_gc_never_pickles_temporaries
    assert len_pre_f == len_post_f
AssertionError
Ran 3884 tests in 2372.584s
FAILED (SKIP=775, failures=1)
```

Sentient07:
@nouiz Ignore my previous comment. The travis passed. The one i referred to is in my travis. I'm not sure why that is occurring

Sentient07:
@nouiz just a confirmation, I'll start the work of dict -> frozen dict in a separate PR ?

nouiz:
Yes separate issue, separate PR.

doc/extending/graphstructures.txt
```diff
@@ -352,22 +352,22 @@ such that: ``var.clients[*][0].inputs[index]`` or
 >>> v = theano.tensor.vector()
 >>> f = theano.function([v], (v+1).sum())
 >>> theano.printing.debugprint(f)
-Sum{acc_dtype=float64} [id A] ''   1
+Sum{axis=None, dtype='float64', acc_dtype='float64'} [id A] ''   1
```
nouiz: We don't want that. I think I commented a few times to revert part of your changes. We don't want to remove \__str__ when it change the string representation. It will make it less readable.

Sentient07:
I haven't removed any __str__ methods in this PR. It was done at few places
in https://github.com/Theano/Theano/pull/4039 and i've reverted all of
them. I tested the requirement of this change. If you remove the __props__
from the Sum method, the ground truths change again.

ᐧ

Sentient07:
I have discussed with @nouiz regarding this. My bad i didn't look in detail of how `__hash__`, `__str__` and etc are handled by the Props. Since `CAReduceDtype` implements those methods inside the class, the ground truth isn't what is desired.
So, I see a solution here, 
Add an __str__ method to Sum class inside elemwise.py but get rid of all such methods that are implemented by Props (except __str__)  
Is this fine @nouiz ?

Sentient07:
Sorry for the delayed update on this PR. Have addressed all the comments to the best of my knowledge.

theano/compile/ops.py
```diff
@@ -518,6 +518,8 @@ class FromFunctionOp(gof.Op):
 
     """
 
+    __props__ = ("fn", "itypes", "otypes", "infer_shape")
```
nouiz: I'm not sure this is good. infer_shape is the parameter in the \__init__, but usually, we set the props with the same name in the class. We don't do that here.

For simplicity, just remove.

theano/gof/op.py
```diff
@@ -780,6 +780,10 @@ class Op(utils.object2, PureOp, CLinkerOp):
 
     """
     def __new__(cls, *args, **kwargs):
+        # TODO: Ensuring all the Ops created henceforth has __props__
+        #       This helps to ease the process of caching by making use
+        #       of the _props_dict() instead of inferring the parameters
+        #       passed to the class's constructor.
```
nouiz: Remove this comment. I don't think for now that it would be a good idea to make \__props__ mandatory. I don't see that coming in the futur.

theano/scalar/basic.py
```diff
@@ -852,6 +852,7 @@ class ScalarOp(Op):
 
     nin = -1
     nout = 1
+    __props__ = ("output_types_preference", "name")
```
nouiz: @abergeron what do you think of adding props to Scalar? I think it is fine, but as we didn't do it before, a look by someone else would be good.

theano/tensor/elemwise.py
nouiz: I think this was already in the PR that I merged about DimShuffle. Can you rebase this PR to remove this duplicate changes in both PR?

theano/tensor/elemwise.py
```diff
@@ -1835,6 +1840,7 @@ class CAReduceDtype(CAReduce):
         * for complex dtypes, we use at least complex128.
 
     """
+    __props__ = ("scalar_op", "axis", "dtype", "acc_dtype")
```
nouiz: remove the \__eq__ and \__hash__ method.

theano/tensor/elemwise.py
```diff
@@ -1993,10 +1999,24 @@ class Sum(CAReduceDtype):
...
+        name = self.__class__.__name__
+        axis = ""
+        if self.axis is not None:
+            axis = ", ".join(str(x) for x in self.axis)
+            axis = "axis=[%s], " % axis
+        return "%s{%sacc_dtype=%s}" % (
+            name,
+            axis,
+            str(self.acc_dtype)
+        )
```
nouiz: why did you add this method?

theano/tensor/elemwise.py
```diff
@@ -2041,6 +2061,7 @@ class Prod(CAReduceDtype):
     input.
 
     """
+    __props__ = ("axis", "dtype", "acc_dtype")
```
nouiz: I won't repeat it. In this file, check if you need to remove \__eq__ and \__hash__ for each class you added \__props__. Take the time to understand if you need to remove them or not and only do it if it is needed.

theano/gof/op.py
Sentient07: I wanted this earlier mainly thinking of the Ops inside elemwise so that it could be cached. Now we're using the frozen dict approach and this isn't needed anymore

theano/tensor/elemwise.py
Sentient07: I'm slightly confused here and to test i added this method. When I added `__props__` to this Op, there were errors in test/test_printing as the string that this Op returns doesn't match with the ground truth. I then realized, it is creating the `__str__` method from the `__props__` inside the utils instead of it using the parent class's `__str__`. that's why I added this method, similar to `CAReduceDtype`. 
My confusion here is, even if i add `__props__` shouldn't the parent's class `__str__` method be used instead of creating one on it's own ? Being more precise, shouldn't line 199 in theano/gof/utils.py evaluate to `False` ?

Sentient07:
I `grep`ed for ```__eq__``` methods. From that I have found that these list of Ops still have them and I'm not sure if I can add ```__props__``` to them or not. I'll write them down here. 

- [ ] ConvOp (/theano/tensor/nnet/conv.py)

- [ ] Scan(inside scan. There seems to be a lot of keys for checking equality, which aren't parameter to this Op)

- [ ] Solve(sandbox/solve.py)

- [ ]  Minimal (sandbox/solve.py)

- [ ] Many/All Ops inside the old backend (sandbox/cuda/nnet.py and sandbox/cuda/fftconv.py sandbox/cuda/blas.py)

- [ ] OpFromGraph (compile/builders)

Sentient07:
I observed that Composite Scalar checks the equality with `_c_code` class variable, which is the c-code generated. I understand that only with the inputs and outputs alone we can't check equality, but shouldn't we have something else as a comparison ? 
This creates errors when I use `MetaObject` implementation of `__eq__` for `Elemwise`. The nodes of the type .. `Elemwise{Composite{....}}` gives error with `_c_code` variable not found. I have a solution that fixes this problem, 
Add a kwarg to `Composite` class ( `_op_code`) that is defaulted to None and set the class variable if the kwarg is passed, else set it using `init_c_code` method. 
Is this good @nouiz ?

abergeron:
The __eq__ method of Composite should generate the `_c_code` attribute if it needs it for the comparison, as is the case for all methods of Composite.

Also, it's ok to leave some `__eq__` methods in place even if you do use props.  They won't be overwritten and will no longer create problems for descendant classes.  Some of our ops are special and it's ok.

nouiz:
You could change ConvOp to use props. But we will probably get rid of it in a week or 2. So I would not botter.
Don't touch scan, the old back-end or OpFromGraph. You can change Solve and Minimal.

Sentient07:
@abergeron @nouiz The nodes like `Elemwise{Composite{((i0 / i1) / i2)}}` creates problem. You can see the error in second link in travis. From what I see, the `inplace_pattern` is an immutable object inside `__init__` of `Elemwise` but not while implementing the `__eq__` from inside `utils.py`. I'm not able to understand how this happens. And from tracing back the code, I see this error arises while pickling of a graph. 
( This is related to another PR #5071 though)

Sentient07:
Travis passes. @nouiz is this good ?

theano/tensor/elemwise.py
```diff
@@ -1714,19 +1705,14 @@ class All(CAReduce):
...
     """
+    __props__ = ("axis",)
 
     def __init__(self, axis=None):
         CAReduce.__init__(self, scalar.and_, axis)
 
     def _output_dtype(self, idtype):
         return "int8"
 
-    def __str__(self):
```
nouiz: Add this back. This class (All) inherit from CAReduce that implement \__str__. We don't want CARedure \__str__.

theano/tensor/elemwise.py
```diff
@@ -1746,19 +1732,14 @@ class Any(CAReduce):
...
     """
+    __props__ = ("axis", )
 
     def __init__(self, axis=None):
         CAReduce.__init__(self, scalar.or_, axis)
 
     def _output_dtype(self, idtype):
         return "int8"
 
-    def __str__(self):
```
nouiz: Removing this change the printing from Any{0,1} to Any{axis=[0,1]}.

The long explanation is that now we use metaclass. So if a subclass implement props, but not \__str__, we won't use the parent \__str__, but the default \__str__. I think it is better to not change the \__str__ for this class, so just add this back.

theano/tensor/elemwise.py
nouiz: See the other comment that talk about metaclass. This explain why we need a new \__str__ method here.

As I think it is better that we don't change the output, keep this method.

theano/tensor/elemwise.py
```diff
@@ -2029,6 +2017,7 @@ class Prod(CAReduceDtype):
     input.
 
     """
+    __props__ = ("axis", "dtype", "acc_dtype")
```
nouiz: Add a comment that we don't want to add "no_zeros_in_input" as it don't change what this op compute. It only change what the grad change.

Sentient07:
I am getting this error message locally, but not on travis. (Cleared cache, and purged it) 
http://pastebin.com/6QwZz9vu

nouiz:
The error you have locally aren't related to Theano, but with your installation.

Do you have those failure with the master of Theano? If not, then you are fine and the problem is just that this PR is based on an older Theano that didn't support your more recent pydot. Otherwise, reinstall pydot.

nouiz:
Ok to merge when travis pass.

Sentient07:
> Do you have those failure with the master of Theano? If not, then you are fine and the problem is just that this PR is based on an older Theano that didn't support your more recent pydot. Otherwise, reinstall pydot.

Yes, I'm getting it over the current master. I tried updating using pip `pydot` package, but that didn't help

nouiz:
try "pip uninstall pydot" a few times. Then "pip install pydot-ng"

PR #4756, opened 2016-07-18
Theano variable arguments for pooling 2

### LINKED ISSUE #3349: downsample.max_pool_2d only accepts int argument
Currently downsample.max_pool_2d only accepts int argument for ds and st parameters. However it will be better if it can also accept tensor variable, for example in spatial pyramid pooling, the st and ds changes when the input image size changes.

```python
        b, c, h, w = X.shape  # tensor variable
        layer_out = []
        for i in [1,2,4]:
            out = downsample.max_pool_2d(state_below, ds=(h/i, w/i),
                                         st=(h/i, w/i),
                                         ignore_border=self.ignore_border,
                                         mode=self.mode)

            out = out.reshape((b, T.prod(out.shape)/b))
            layer_out.append(out)
```
here we get the error that max_pool_2d only takes integer value for ds and st. I think it will be better for ds and st to accept both integer and tensor variable so the max_pool_2d can change on the fly in the computation graph, if theano thinks this is useful, i can work on a pull request

Thrandis:
Fixes #3349

Previous merge failed, so restart PR here!
#### TODO:
- [x] Test GPU (2 backends)

Adapt interface in the whole lib:
- [x] tensor
- [x] gpuarray
- [x] sanbox/cuda
- [x] other


Thrandis:
@mgermain @lamblin so I need to correct the `test_old_pool_interface` that I added. It seems that I can't load a function that I pickled...

Otherwise the last step is to look for `pool` calls in Theano to adapt them to the new interface.

theano/tensor/signal/pool.py
```diff
@@ -189,9 +187,9 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
         r = tensor.extract_constant(r)
         c = tensor.extract_constant(c)
         if padding[0]:
-            r += padding[0] * 2
+            r = r + padding[0] * 2
```
mgermain: Why ?

Thrandis: If I remember correctly it was because it was raising an error otherwise. The `+=` was not working in one of the cases, because the type of variables was somehow not compatible.

mgermain: Right.

Thrandis:
TODO:
- [ ] Test GPU (2 backends)
- [ ] Adapt interface in the whole lib.

Thrandis:
@mgermain @lamblin I found what is the problem with the old backend. We have `GpuDownsampeFactorMax`, `GpuDownsampleFactorMaxGrad` and `GpuDownsampleFactorMaxGradGrad` ops in [blas.py](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/blas.py) that also need to be adapted to accept `ws` (the size of the pooling) as tensor variable.

I know I can adapt the `make_node` fuction of those ops like I did for `Pool`, but how shall I change `__eq__`, `__hash__` and `__str__` to work with tensor variables. Do I remove simply the `self.ds`?

```python
class GpuDownsampleFactorMax(GpuOp):
    """
    Implement downsample with max on the gpu.
    """
    def __init__(self, ds, ignore_border=False):
        self.ds = tuple(ds)
        self.ignore_border = ignore_border

    def __eq__(self, other):
        return (type(self) == type(other) and
                self.ds == other.ds and
                self.ignore_border == other.ignore_border)

    def __hash__(self):
        return hash(type(self)) ^ hash(self.ds) ^ hash(self.ignore_border)

    def __str__(self):
        return '%s{%s,%s}' % (self.__class__.__name__,
                              self.ds,
                              self.ignore_border)
```

Thrandis:
@mgermain Alright so I made the changes to print the warning in the old backend. I'll run the full test suit for the 2 backends and let you know when it is done!

Thrandis:
@mgermain @lamblin : Results of the test!

#### New Backend

`THEANO_FLAGS=init_gpu_device=cuda2,floatX=float32`

> FAIL: test_infer_shape (theano.gpuarray.tests.test_extra_ops.TestGpuCumsum)
>
> Ran 23925 tests in 8030.797s
>
> FAILED (SKIP=714, failures=1)

#### Old Backend

`THEANO_FLAGS=init_gpu_device=gpu2,floatX=float32`

> Ran 23196 tests in 6538.804s
> 
> OK (SKIP=490)

lamblin:
> FAIL: test_infer_shape (theano.gpuarray.tests.test_extra_ops.TestGpuCumsum)

That one is unrelated, and should be fixed in the current master.

Jenkins, test this, please.

lamblin:
Jenkins test this please

slefrancois:
jenkins test this please

slefrancois:
jenkins test this please

nouiz:
jenkins tests passed, but had a seg fault at the end. That is why it failed, but we can consider it as passed.

doc/library/tensor/signal/downsample.txt
```diff
@@ -1,17 +0,0 @@
-.. _libdoc_tensor_signal_downsample:
-
-======================================================
-:mod:`downsample` -- Down-Sampling
-======================================================
-
```
nouiz: I would keep that file but put a reference to the new version. This way old link won't completly break.

theano/tensor/signal/pool.py
```diff
@@ -234,50 +232,86 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
         if mode not in ['max', 'average_inc_pad', 'average_exc_pad', 'sum']:
             raise ValueError(
                 "Pool mode parameter only support 'max', 'sum',"
                 " 'average_inc_pad' and 'average_exc_pad'. Got %s" % mode)
         self.mode = mode
 
-    def make_node(self, x):
+    def prepare_node(self, node, storage_map, compute_map):
+        if len(node.inputs) == 1:
+            warnings.warn("Theano Pool internal changed.", stacklevel=3)
```
nouiz: Add something like "We convert it" at the end so that people don't panic with this. Or just remove it as we do it automatically? I would just remove it, but if you prefer to keep it, update the message to help prevent user questions.

theano/tensor/signal/pool.py
```diff
@@ -234,50 +232,86 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
+            # Old interface
+            self.mode = node.op.mode
+            ws = theano.tensor.constant(node.op.ds)
+            st = theano.tensor.constant(node.op.st)
+            pad = theano.tensor.constant(node.op.padding)
+            node.inputs.append(ws)
+            node.inputs.append(st)
+            node.inputs.append(pad)
+            if isinstance(ws, theano.Constant):
```
nouiz: We call the function constant() above, so they all will be constant. You can remove the conditions and keep the if clause only. If I'm wrong, add a comment to tell why this is needed.

theano/gpuarray/dnn.py
```diff
@@ -1859,13 +1859,10 @@ def local_gpua_pool_dnn_alternative(op, ctx_name, inputs, outputs):
         raise_no_cudnn()
     if not op.ignore_border:
         return
-    img, = inputs
+    img, ws, stride, pad = inputs
```
nouiz: why did you started to rename ds to ws everywhere? We didn't discuss this interface change. I'm not sure it is good. Currently we use ds and ws, and I don't think mixing them is good. So we should decide which one to keep.

This is at many places.

theano/tensor/signal/pool.py
```diff
@@ -234,50 +232,86 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
+                storage_map[pad] = [None]
+                compute_map[pad] = [False]
+
+    def make_node(self, x, ws, stride=None, pad=(0, 0)):
         # TODO: consider restricting the dtype?
         x = tensor.as_tensor_variable(x)
+        if stride is None:
+            stride = ws
+        if isinstance(pad, (tuple, list)):
+            pad = tuple(pad)
```
nouiz: this line isn't useful.

theano/tensor/signal/pool.py
```diff
@@ -234,50 +232,86 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
+            if pad != (0, 0) and not self.ignore_border:
+                raise NotImplementedError(
+                    'padding works only with ignore_border=True')
+            if isinstance(ws, (tuple, list)):
+                ws = tuple(ws)
```
nouiz: line not useful.

theano/tensor/signal/pool.py
```diff
@@ -357,18 +392,23 @@ def c_headers(self):
...
             omp_parallel = ''
         ccode = """
+        // Getting ws, stride and pad
+        int ws0, ws1, st0, st1, pd0, pd1;
+        ws0 = *((npy_intp*)PyArray_GETPTR1(%(ws)s, 0));
+        ws1 = *((npy_intp*)PyArray_GETPTR1(%(ws)s, 1));
+        st0 = *((npy_intp*)PyArray_GETPTR1(%(stride)s, 0));
+        st1 = *((npy_intp*)PyArray_GETPTR1(%(stride)s, 1));
+        pd0 = *((npy_intp*)PyArray_GETPTR1(%(pad)s, 0));
+        pd1 = *((npy_intp*)PyArray_GETPTR1(%(pad)s, 1));
```
nouiz: Check that the lenght of those vector are 2.

theano/tensor/signal/pool.py
```diff
@@ -624,57 +664,89 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
+            warnings.warn("Theano PoolGrad internal changed.", stacklevel=3)
+            # Old interface
+            self.mode = node.op.mode
+            ws = theano.tensor.constant(node.op.ds)
+            st = theano.tensor.constant(node.op.st)
+            pad = theano.tensor.constant(node.op.padding)
+            node.inputs.append(ws)
+            node.inputs.append(st)
+            node.inputs.append(pad)
+            if isinstance(ws, theano.Constant):
```
nouiz: idem for the if.

theano/tensor/signal/pool.py
```diff
@@ -624,57 +664,89 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
-        return Apply(self, [x, maxout, gz], [x.type()])
+        assert isinstance(ws, Variable) and ws.ndim == 1
+        assert isinstance(stride, Variable) and stride.ndim == 1
+        assert isinstance(pad, Variable) and pad.ndim == 1
+        return Apply(self, [x, maxout, gz, ws, stride, pad], [x.type()])
 
     def perform(self, node, inp, out):
         assert self.mode == 'max'
-        x, maxout, gz = inp
+        x, maxout, gz, ws, stride, pad = inp
```
nouiz: assert the shape of the 3 vectors.

theano/tensor/signal/pool.py
```diff
@@ -745,14 +817,22 @@ def c_code(self, node, name, inp, out, sub):
...
             %(fail)s;
         }
+        // Getting ws, stride and pad
+        int ws0, ws1, st0, st1, pd0, pd1;
+        ws0 = *((npy_intp*)PyArray_GETPTR1(%(ws)s, 0));
+        ws1 = *((npy_intp*)PyArray_GETPTR1(%(ws)s, 1));
+        st0 = *((npy_intp*)PyArray_GETPTR1(%(stride)s, 0));
+        st1 = *((npy_intp*)PyArray_GETPTR1(%(stride)s, 1));
+        pd0 = *((npy_intp*)PyArray_GETPTR1(%(pad)s, 0));
+        pd1 = *((npy_intp*)PyArray_GETPTR1(%(pad)s, 1));
```
nouiz: assert the shape. (not the number of dimensions, this is done automatically for the c code, but not for the perform.

theano/tensor/signal/pool.py
```diff
@@ -820,54 +900,59 @@ def c_code(self, node, name, inp, out, sub):
...
+        assert isinstance(ws, Variable) and ws.ndim == 1
+        assert isinstance(stride, Variable) and stride.ndim == 1
+        assert isinstance(pad, Variable) and pad.ndim == 1
+        return Apply(self, [x, gz, ws, stride, pad], [x.type()])
 
     def perform(self, node, inp, out):
-        if self.mode == 'average_exc_pad' and self.padding != (0, 0):
-            raise NotImplementedError()
-        x, gz = inp
+        x, gz, ws, stride, pad = inp
```
nouiz: assert shape, also in c code.

theano/tensor/signal/pool.py
```diff
@@ -820,54 +900,59 @@ def c_code(self, node, name, inp, out, sub):
...
+        ws = tensor.as_tensor_variable(ws)
+        stride = tensor.as_tensor_variable(stride)
+        pad = tensor.as_tensor_variable(pad)
         assert isinstance(x, Variable) and x.ndim == 4
         assert isinstance(gz, Variable) and gz.ndim == 4
-
-        return Apply(self, [x, gz], [x.type()])
+        assert isinstance(ws, Variable) and ws.ndim == 1
+        assert isinstance(stride, Variable) and stride.ndim == 1
+        assert isinstance(pad, Variable) and pad.ndim == 1
```
nouiz: you don't check the dtype.

theano/tensor/signal/pool.py
```diff
@@ -624,57 +664,89 @@ def out_shape(imgshape, ds, ignore_border=False, st=None, padding=(0, 0)):
...
+        stride = tensor.as_tensor_variable(stride)
+        pad = tensor.as_tensor_variable(pad)
         assert isinstance(x, Variable) and x.ndim == 4
         assert isinstance(maxout, Variable) and maxout.ndim == 4
         assert isinstance(gz, Variable) and gz.ndim == 4
-
-        return Apply(self, [x, maxout, gz], [x.type()])
+        assert isinstance(ws, Variable) and ws.ndim == 1
+        assert isinstance(stride, Variable) and stride.ndim == 1
+        assert isinstance(pad, Variable) and pad.ndim == 1
```
nouiz: you don't check the dtype

theano/tensor/signal/pool.py
```diff
@@ -942,11 +1013,34 @@ def make_node(self, x, maxout, gz):
...
+            raise TypeError('Pool downsample parameters must be ints.')
+        if not stride.dtype.startswith('int'):
+            raise TypeError('Stride parameters must be ints.')
+        if not pad.dtype.startswith('int'):
+            raise TypeError('Padding parameters must be ints.')
+        return Apply(self, [x, maxout, gz, ws, stride, pad], [x.type()])
 
     def perform(self, node, inp, out):
-        x, maxout, ggx = inp
+        x, maxout, ggx, ws, stride, pad = inp
```
nouiz: assert shape and in c code.

theano/tensor/signal/tests/test_pool.py
```diff
@@ -835,33 +805,74 @@ def test_infer_shape(self):
...
+                            fix_fct = theano.function([x], [y, dx])
+                            var_y, var_dx = var_fct(data, (ws, ws), (st, st),
+                                                    (pad, pad))
+                            fix_y, fix_dx = fix_fct(data)
+                            utt.assert_allclose(var_y, fix_y)
+                            utt.assert_allclose(var_dx, fix_dx)
+
+    def test_old_pool_interface(self):
+        if sys.version_info[0] != 3:
+            raise SkipTest('Skip old pool interface with python 2.x')
```
nouiz: why support this only in python 3?

nouiz:
The PR look pretty good. Thanks, just a few comments.

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: We decided to keep the code like that.

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/tests/test_pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done

theano/tensor/signal/pool.py
Thrandis: done!

theano/tensor/signal/pool.py
Thrandis: done

theano/gpuarray/dnn.py
Thrandis: It has been renamed to be consistent with the new `gpuarray` backend.

nouiz:
Travis have failures.

Thrandis:
Note: I can't reproduce the travis failures. I didn't updated the `c_code_cache`, maybe that is the problem!

nouiz:
to trigger travis.

doc/library/tensor/signal/index.txt
```diff
@@ -20,5 +20,5 @@ forms of signal processing.
     :maxdepth: 1
 
     conv
-    downsample
     pool
+    downsample
```
nouiz: why not remove it?

Thrandis: @nouiz I tought you wanted to keep it not to break the documentation. I can remove it if you want!

PR #4639, opened 2016-06-16
sample multinomial without replacement - GPU

aalmah:
This is a first working version. It's not very efficient, so we need to improve on that. This is for #4139

PR #4757, opened 2016-07-18
Multinomial Without Replacement

abergeron:
This is the finishing touches of #4639


theano/gpuarray/multinomial.py
```diff
@@ -487,16 +487,12 @@ def local_gpua_multinomial(op, context_name, inputs, outputs):
 
 @register_opt()
```
nouiz: This should have 'fast_compile' too.

PR #4764, opened 2016-07-19
Float16 convolutions

abergeron:


theano/gpuarray/dnn.py
```diff
@@ -488,7 +474,6 @@ def get_op_params(self):
         elif self.algo == 'fft':
             alg = 'CUDNN_CONVOLUTION_FWD_ALGO_FFT'
         elif self.algo == 'fft_tiling':
-            # need v4
```
nouiz: Keep that for history. It is hard to find that information otherwise.

Issue #4539, opened 2016-05-19
Allow indexing using Ellipsis


NeilGirdhar:
Sometimes, I need to update a shared variable having dynamic shape.  I do this by calling:

```
function([], [], updates=[(x, inc(x, tensor.inc_subtensor(x[index], 1.0)))])
```

When the shape of `x` is a vector like `(4,)`, then `index == slice(None)` updates the whole vector.  Using `index == ()` gives a crazy exception (that should probably be fixed).

When the shape of `x` is a scalar like `()`, then `index == ()` works fine whereas using `slice(None)` gives the reasonable error:

> ValueError: The index list is longer (size 1) than the number of dimensions of the tensor(namely 0). You are asking for a dimension of the tensor that does not exist! You might need to use dimshuffle to add extra dimension to your tensor.

In numpy, I can deal with both cases by letting `index = Ellipsis` (or `...`).  Unfortunately, theano doesn't accept it, telling me:

> /opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/theano/tensor/basic.py in as_tensor_variable(x, name, ndim)
>     206         except Exception:
>     207             str_x = repr(x)
> --> 208         raise AsTensorError("Cannot convert %s to TensorType" % str_x, type(x))
>     209
>     210 # this has a different name, because _as_tensor_variable is the
> 
> AsTensorError: ('Cannot convert Ellipsis to TensorType', <class 'ellipsis'>)

NeilGirdhar:
A related problem is given a matrix `m` with some shape, how do I prepend a newaxis?  In numpy, I would do:

    m[np.newaxis]

or, more explicitly,

    m[np.newaxis, ...]

However, Theano complains about the first that:

    ValueError: ('You cannot drop a non-broadcastable dimension.', ((False,), ('x',)))

And about the second that


    AsTensorError: ('Cannot convert Ellipsis to TensorType', <class 'ellipsis'>)

I do not know the shape of `m` a priori, so I cannot do `m[np.newaxis, :]` or similar.

NeilGirdhar:
Squashed pull request: https://github.com/Theano/Theano/pull/4840

nouiz:
merged in https://github.com/Theano/Theano/pull/4853

PR #4767, opened 2016-07-20
Allow indexing using Ellipsis as in numpy.

NeilGirdhar:
Ellipsis is automatically replaced by an appropriate number of newaxis constants up to the number of dimensions of the tensor.
Bumped numpy to 1.9.1.
Bumped scipy to 0.14.0.

### DIFF
--- a/theano/tensor/tests/test_subtensor.py
+++ b/theano/tensor/tests/test_subtensor.py
@@ -1,39 +1,40 @@
-from __future__ import absolute_import, print_function, division
+from __future__ import absolute_import, division, print_function
+
 import logging
 import sys
 import unittest
 
-from nose.plugins.skip import SkipTest
 import numpy
+from nose.plugins.skip import SkipTest
+from nose.tools import assert_equal
+from numpy.testing import assert_array_equal
 from six import StringIO
 from six.moves import xrange
 
 import theano
-from theano.compat import exc_message, izip, PY3
-from theano.compile import DeepCopyOp
-from theano import config
-from theano import gof
 import theano.scalar as scal
 import theano.tensor as tensor
-from theano.tests import unittest_tools as utt
-from theano.tensor.subtensor import (inc_subtensor, set_subtensor,
+from theano import config, gof
+from theano.compat import PY3, exc_message, izip
+from theano.compile import DeepCopyOp
+from theano.tensor import (MakeSlice, NotScalarConstantError, _shared,
+                           as_tensor_variable, cscalar, ctensor3, dmatrix,
+                           dscalar, dtensor4, dvector, fmatrix, fscalar,
+                           fvector, iscalar, lmatrix, lrow, lvector, matrix,
+                           vector)
+from theano.tensor.basic import DimShuffle
+from theano.tensor.subtensor import (AdvancedIncSubtensor,
+                                     AdvancedIncSubtensor1, AdvancedSubtensor,
+                                     AdvancedSubtensor1, IncSubtensor,
+                                     Subtensor, advanced_inc_subtensor,
                                      advanced_inc_subtensor1,
-                                     advanced_set_subtensor1,
-                                     advanced_inc_subtensor,
                                      advanced_set_subtensor,
-                                     Subtensor, IncSubtensor,
-                                     AdvancedSubtensor1, AdvancedSubtensor,
-                                     advanced_subtensor1, inplace_increment,
-                                     AdvancedIncSubtensor1,
-                                     AdvancedIncSubtensor,
-                                     get_canonical_form_slice)
-from theano.tensor import (as_tensor_variable, _shared,
-                           NotScalarConstantError,
-                           fscalar, iscalar, dscalar, cscalar,
-                           vector, dvector, fvector, lvector, lrow,
-                           fmatrix, dmatrix, lmatrix, matrix,
-                           ctensor3, dtensor4)
-from theano.tensor.tests.test_basic import rand, randint_ranged, inplace_func
+                                     advanced_set_subtensor1,
+                                     advanced_subtensor1,
+                                     get_canonical_form_slice, inc_subtensor,
+                                     inplace_increment, set_subtensor)
+from theano.tensor.tests.test_basic import inplace_func, rand, randint_ranged
+from theano.tests import unittest_tools as utt
 from theano.tests.unittest_tools import attr
 
 if PY3:
@@ -65,6 +66,7 @@ class T_subtensor(unittest.TestCase, utt.TestOptimizationMixin):
         self.adv_incsub1 = adv_incsub1
         if mode is None:
             mode = theano.compile.mode.get_default_mode()
+            mode = mode.including("local_useless_subtensor")
         self.mode = mode
         self.dtype = dtype
         self.type = type
@@ -97,18 +99,18 @@ class T_subtensor(unittest.TestCase, utt.TestOptimizationMixin):
         Subtensor.debug = False
         utt.seed_rng()
 
-    def eval_output_and_check(self, t, list=False, mode=None):
+    def eval_output_and_check(self, t, op_type=None, mode=None, length=1):
+        if op_type is None:
+            op_type = self.sub
         if mode is None:
             mode = self.mode
         f = inplace_func([], t, mode=mode)
         topo = f.maker.fgraph.toposort()
         topo_ = [node for node in topo if not isinstance(node.op,
                                                          self.ignore_topo)]
-        assert len(topo_) == 1
-        if not list:
-            assert isinstance(topo_[0].op, self.sub)
-        else:
-            assert isinstance(topo_[0].op, self.adv_sub1)
+        assert_equal(len(topo_), length)
+        if length == 1:
+            assert isinstance(topo_[0].op, op_type)
         tval = f()
         return tval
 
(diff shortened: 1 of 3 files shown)


NeilGirdhar:
Related issue: https://github.com/Theano/Theano/issues/4539

nouiz:
Thanks for looking into that. I'll way until you get more advanced before looking.

I killed many travis run, as each time you do a push it create one and even if you force push later, they will be run. This overload travis, so try to limit that pattern.

NeilGirdhar:
@nouiz Sure, sorry about that.  Didn't know that the old runs kept going.

NeilGirdhar:
@nouiz I added a test, but running the tests on my machine produces many errors (everywhere).  I used `theano-nose theano/tensor/tests/test_subtensor.py`

Now that travis has run my test, any idea on how to get it working?  Besides that, this change is ready for your review.

theano/tensor/tests/test_subtensor.py
```diff
@@ -337,6 +337,23 @@ def test_slice_symbol(self):
...
 
+    def test_ellipsis(self):
+        numpy_n = numpy.arange(24, dtype=self.dtype).reshape((2, 3, 4))
+        n = self.shared(numpy_n)
+        for slice_ in [numpy.index_exp[...],
+                       numpy.index_exp[..., 1],
+                       numpy.index_exp[1, ...],
+                       numpy.index_exp[..., 1, 2, 3],
+                       numpy.index_exp[1, ..., 2, 3],
+                       numpy.index_exp[1, 2, 3, ...],
```
nouiz: you seem to implement newaxis, add tests for it.

nouiz: Also add tests when using list of indices and not just single indices.

theano/tensor/var.py
```diff
@@ -471,6 +471,24 @@ def __getitem__(self, args):
...
+
+        # Convert an Ellipsis if provided into an appropriate number of
+        # slice(None).
+        ellipses = [i
+                    for i, index in enumerate(args)
+                    if index is Ellipsis]
+        if len(ellipses) > 1:
+            raise IndexError(
+                "an index can only have a single Ellipsis (`...`)")
+        if len(ellipses) == 1:
```
nouiz: if -> elif

theano/tensor/tests/test_subtensor.py
```diff
@@ -337,6 +337,23 @@ def test_slice_symbol(self):
...
+                       ]:
+            numpy_tval = numpy_n[slice_]
+            t = n[slice_]
+            self.assertTrue(isinstance(t.owner.op, Subtensor))
+            tval = self.eval_output_and_check(t)
```
nouiz: The error is that the optimized graph have have more then 1 node. Print it (you can modify the asser for that) It will help you understand what is going on. I have no idea just by code review.

theano/tensor/var.py
NeilGirdhar: Even with the raise?

NeilGirdhar:
@nouiz Would you mind helping me debug this?  The test passes fine on my computer, but is failing on travis.  I have no idea how to proceed:
```

~/src/Theano: bin/theano-nose -v -s theano/tensor/tests/test_subtensor.py:T_subtensor.test_ellipsis
test_ellipsis (theano.tensor.tests.test_subtensor.T_subtensor) ... ok

Ran 1 test in 3.639s

OK
~/src/Theano: git d
```

NeilGirdhar:
@nouiz ping?

NeilGirdhar:
@nouiz Can we look at this to minimize merging?

theano/tensor/tests/test_subtensor.py
```diff
@@ -97,18 +98,17 @@ def setUp(self):
         Subtensor.debug = False
         utt.seed_rng()
 
-    def eval_output_and_check(self, t, list=False, mode=None):
+    def eval_output_and_check(self, t, op_type=None, mode=None, length=1):
```
nouiz: Here op_type should be self.sub by default, not None. This probably make many tests in travis fail.

nouiz:
I  have the fix in this PR:

https://github.com/nouiz/Theano/tree/NeilGirdhar-master

Revert all the changes to the import and make sure that tests/test_flake8.py pass.

theano/tensor/tests/test_subtensor.py
NeilGirdhar: I don't understand.  Doesn't line 102 just below here set this to self.sub by default?

NeilGirdhar: Also, you can't use `self` in the definition of a default parameter because `self` isn't bound to anything yet.  Default parameters are evaluated at function definition time — not execution time.

nouiz: right. I missed that part.

NeilGirdhar:
@nouiz Thanks for your help so far.  I'm still having trouble with Python 2.6.  Are you doing something wacky with numpy array indexing in 2.6??  I'm getting a very weird error in the test as if someone is walking the indexing and converting arguments to long.

As an aside, why do you even support 2.6?  It reached end-of-life 3 years ago!

nouiz:
To investigate the python 2.6, you should install it. There is discussion to drop it, but we didn't get to the point of droping it now. It will be  probably after the next release.

Mostly, you need to find exactly which type of object is numpy_n and exactly what is in slice_. I have the feeling it is using a symbolic variable inside a slice. I'm not sure this is good.

NeilGirdhar:
@nouiz I installed it, but the test passes for me.  I'm not sure what's wrong because it's clear that one is a numpy array and the slice is a regular indexing tuple created with `numpy.index_exp`.
```

~/src/Theano: python --version
Python 2.6.9
~/src/Theano: bin/theano-nose theano/tensor/tests/test_subtensor.py:T_subtensor.test_ellipsis
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:562: DeprecationWarning: This function is deprecated. Please call randint(-5, 5 + 1) instead
  return numpy.random.random_integers(-5, 5, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(0, 1 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:578: DeprecationWarning: This function is deprecated. Please call randint(-5, 4 + 1) instead
  r = numpy.random.random_integers(-5, 4, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(-3, 3 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(-5, 5 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(1, 5 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(-1000, 1000 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(-1, 1 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(1, 1000 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(-0.9999999999, 0.9999999999 + 1) instead
  return numpy.random.random_integers(min, max, shape)
/Users/neil/src/Theano/theano/tensor/tests/test_basic.py:588: DeprecationWarning: This function is deprecated. Please call randint(1, 13 + 1) instead
  return numpy.random.random_integers(min, max, shape)
.
Ran 1 test in 0.145s

OK
```

nouiz:
Can you print the information you need just before that line? This way
travis will print it for you.

Did you try with the flag mode=FAST_COMPILE?

It could be related to an old numpy. Travis test with numpy 1.7.1 and
1.9.1. Only the one with 1.7.1 (python 2.6) 1.91 is used for python 3.3

nouiz:
If you find a work around, it is good, but if you find the mininal version
of numpy that support that, we could bump the requirement to 1.8 or 1.9 I
think.

thanks

On Sat, Aug 6, 2016 at 3:54 PM, Frédéric Bastien <frederic.bastien@gmail.com
> wrote:

> Can you print the information you need just before that line? This way
> travis will print it for you.
>
> Did you try with the flag mode=FAST_COMPILE?
>
> It could be related to an old numpy. Travis test with numpy 1.7.1 and
> 1.9.1. Only the one with 1.7.1 (python 2.6) 1.91 is used for python 3.3
>

NeilGirdhar:
I am using fast compile, but I don't see what this has to do with the error.  Reverting to numpy 1.7 causes the error.  Bumping up to numpy 1.9 fixes it.  Can we bump to 1.9?

nouiz:
1.8.1 march 2014
1.8.2 august 2014
1.9.0 7 sept 2015
1.9.3 21 sept 2015

I'm good with that, but we need to ask on theano-dev.

nouiz:
We need to check if we can update travis with python 2.6 with it. Can you
update the file .travis.yml
to install numpy 1.9 for python 2.6?

On Sat, Aug 6, 2016 at 4:50 PM, Frédéric Bastien <frederic.bastien@gmail.com
> wrote:

> 1.8.1 march 2014
> 1.8.2 august 2014
> 1.9.0 7 sept 2015
> 1.9.3 21 sept 2015
>
> I'm good with that, but we need to ask on theano-dev.
>

NeilGirdhar:
Are you sure you got your dates right?  https://github.com/numpy/numpy/blob/master/doc/release/1.9.0-notes.rst

1.9 seems to have been released in 2014?

NeilGirdhar:
FYI: a pull request can be squashed automatically now? https://github.com/blog/2141-squash-your-commits

NeilGirdhar:
Squashed in https://github.com/Theano/Theano/pull/4840

lamblin:
Finished in #4853.

PR #4771, opened 2016-07-21
Make the old backend support as_input_f32 theano.config.dnn.precision.

nouiz:


abergeron:
Ok when travis passes.

PR #4772, opened 2016-07-21
Fix batchnorm error in buildbot.

abergeron:


theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -823,7 +823,7 @@ def test_batchnorm_inference():
                 utt.assert_allclose(outputs[3], outputs[3 + 5])  # dscale
                 utt.assert_allclose(outputs[4], outputs[4 + 5])  # dbias
                 utt.assert_allclose(outputs[5], outputs[5 + 5])  # dmean
-                utt.assert_allclose(outputs[6], outputs[6 + 5])  # dvar
+                utt.assert_allclose(outputs[6], outputs[6 + 5], rtol=2e-3, atol=5e-5)  # dvar
```
nouiz: @f0k what do you think of this? It fix errors in the current buildbot.

f0k: Does the buildbot run cuDNN code now? The `rtol` seems pretty high to me, is this the minimum that works? (Just making sure. The failure message should tell the relative and absolute difference.)

nouiz: Yes, the daily buildbot have changed computer and GPUs (now GTX750). But
travis don't have GPUs.

We started to implement jenkins, it work for whitelisted people but the GPU
isn't enabled yet on job. But the computer also have 750.

I agree we should check if the rtol is at the minimim needed or not.

abergeron:
This is pretty close to the minimum.  Any lower (1e-3) and it would fail randomly.

Maybe if we use the cpu BatchNorm op as a comparison point we could get it lower but I didn't have the time to get into that.

f0k:
> This is pretty close to the minimum. Any lower (1e-3) and it would fail randomly.

Okay, thanks for confirming! Good to merge then, I'd say.

PR #4773, opened 2016-07-21
Add support for spaces in paths

ruslanagit:
Make theano support programs and packages installed on Windows under paths which
include spaces (such as 'Program Files')


theano/sandbox/cuda/__init__.py
```diff
@@ -296,11 +296,12 @@ def dnn_available():
   return 1;
 }
 """
-            params = ["-l", "cudnn", "-I" + os.path.dirname(__file__)]
+            params = ["-l", "cudnn"]
+            params.extend(['-I"%s"' % os.path.dirname(__file__)])
             if config.dnn.include_path:
-                params.append("-I" + config.dnn.include_path)
+                params.extend(['-I"%s"' % config.dnn.include_path])
```
nouiz: If you want to also support the GPU, then you should also check the file nvcc_compiler.py in that directory. There are some -L and others options that you didn't touched.

ruslanagit: OK. Thanks for pointing this out.
On my machine I did not have to fix this to make GPU work, so I was unaware of this file.

theano/configdefaults.py
```diff
@@ -1259,21 +1262,23 @@ def default_blas_ldflags():
...
+                flags = ['-L"%s"' % lib_path]
                 flags += ['-l%s' % l for l in ["mkl_core",
                                                "mkl_intel_thread",
                                                "mkl_rt"]]
                 res = try_blas_flag(flags)
                 if res:
                     return res
 
+        # to support path that includes spaces, we need to wrap it with double quotes on Windows
+        path_wrapper = "\"" if os.name == 'nt' else ""
```
nouiz: Is there a reason to do that only on nt? It would make the code easier to understand.

Also, we use sys.platform == 'win32' everywhere else. So if there isn't a reason not to use that instead of os.name, I would prefer that.

Otherwise, the PR look good. thanks.

ruslanagit: Honestly, I am not a huge specialist in platform related issues. The reasons I choose to use os.name == 'nt'condition were:
1. This is the condition used in file **theano/misc/windows.py** (line 13)
2. This is the fist solutions that pops-up with google search "python check if windows" -> 
go to http://stackoverflow.com/questions/1325581/how-do-i-check-if-im-running-on-windows-in-python (first link in search) and then the most voted solution (105 votes) suggests using os.name, and in the next solution (with sys.platform) you can find _'But os.name is probably the way to go, as others have mentioned.'_
I googled the difference between both, and as far as I understand the os.name is more general than the sys.platform. So, os.name could be used to check if I run on **any** Windows OS, and sys.platform could be used to check if I run on **some specific** Windows platform. But, once again, I am not a specialist in OS. 
I believe the paths should be treated the same across all Windows platforms (alike subprocess is treated the same in windows.py), so os.name is more appropriate here. sys.platform, may be more appropriate for compilation related checks, though.

nouiz: thanks for looking into that. It seem I was wrong when I taught we used the same way everywhere. What you tell is pretty convincing:)

nouiz:
I was going to merge, but this branch now have conflict. Can you rebase it? thanks

nouiz:
ok to merge when travis pass.

nouiz:
thanks

Issue #4601, opened 2016-06-07
Remove the name parameter to reshape op


nouiz:
I don't know any other op that have that. I think it should be on the variable and not the op. So this is clean up of old stuff. Check the GPU op and gpu opt, as they handle it.

EugenePY:
Hi! I would like to work on this. Do you mean just remove the name parameter of reshape? I can also check other op.

nouiz:
I think it is only this op that have this. So no need to check the other
ops. Yes, remove the name, but it will probably request some small
adjustment where it is used to remove it.

thanks

PR #4777, opened 2016-07-25
ccw4601

nke001:
fix gh-4601


nke001:
@nouiz

nke001:
@adri-romsor

theano/tensor/basic.py
```diff
@@ -4573,7 +4573,7 @@ def reshape(x, newshape, ndim=None, name=None):
                 "to know what the number of dimensions of the reshaped "
                 "variable will be. You can provide the 'ndim' keyword "
                 "argument to 'reshape' to avoid this problem." % newshape)
-    op = Reshape(ndim, name)
+    op = Reshape(ndim)
```
nouiz: Run all Theano tests on a computer with a GPU. There is some code in GPU that support it is there. Look at least in theano/sandox/cuda/basic_ops.py theano/sandbox/cuda/opt.py and theano/gpuarray/basic_ops.py

nke001: @nouiz Do I have to specify anything in the command for the test to be run on GPU? thanks.

nouiz: no, but a GPU must be available.

If you want to force a specific GPU, use this flag

init_gpu_device=gpu1

abergeron:
jenkins test this please

nke001:
@nouiz @adri-romsor It has passed the GPU tested on Jerkins as well. So I guess the ccw can be closed?

nouiz:
It is not normal that the test passed... So it probably mean we missed a test. I checked manually and in the file theano/gpuarray/opt.py in the function local_gpua_reshape, we use op.name.

It also mean we don't have good test for the reshape optimization in the new back-end. With that small change I described, we can merge this PR. 

thanks

nke001:
It seems like I have messed up the rebase, so I have created a new branch and pull request here.https://github.com/Theano/Theano/pull/4888

PR #4779, opened 2016-07-25
Added the broadcastable flags to all mrg functions.

### LINKED ISSUE #4159: setting broadcastable flags on output of MRG_RandomStreams methods
Should these methods not take a 'broadcastable' argument?

caglar:
Fixes #4159 

Added broadcastable flags as a tuple to all mrg functions.

### DIFF
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1293,9 +1293,10 @@ class MRG_RandomStreams(object):
             inferred from the dtype of low and high, but will be at
             least as precise as floatX.
         nstreams:
-            ? TODO
+            The number of streams to use for sampling the random numbers
+            in parallel.
         broadcastable:
-            A tuple of booleans defining the broadcastable pattern of
+            A tuple of booleans defining the broadcastable pattern of the
             output tensor.
         """
         if not (isinstance(broadcastable, tuple) or broadcastable is None):
@@ -1421,18 +1422,13 @@ class MRG_RandomStreams(object):
                              " should be a tuple.")
 
         pvals = as_tensor_variable(pvals)
-        if size is not None:
-            if any([isinstance(i, integer_types) and i <= 0 for i in size]):
-                raise ValueError(
-                    "The specified size contains a dimension with value <= 0",
-                    size)
 
         if size is not None:
-            raise ValueError(
+            raise NotImplementedError(
                 "Provided a size argument to MRG_RandomStreams.multinomial, "
                 "which does not use the size argument.")
         if ndim is not None:
-            raise ValueError(
+            raise NotImplementedError(
                 "Provided an ndim argument to MRG_RandomStreams.multinomial, "
                 "which does not use the ndim argument.")
         if pvals.ndim == 2:
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1379,12 +1379,13 @@ class MRG_RandomStreams(object):
             else:
                 x = self.uniform(size=size, nstreams=nstreams)
                 x = cast(x < p, dtype)
-                if broadcastable:
-                    if x.ndim != len(broadcastable):
-                        raise ValueError("The length of the broadcastable tuple "
-                                         "should match the number of "
-                                         "dimensions of the output tensor.")
-                    return patternbroadcast(x, broadcastable)
+
+            if broadcastable:
+                if x.ndim != len(broadcastable):
+                    raise ValueError("The length of the broadcastable tuple "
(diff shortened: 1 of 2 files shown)


caglar:
@nouiz @mgermain

caglar:
@nouiz @mgermain The tests are passing.

mgermain:
This looks good so far. Please add some unit tests.

caglar:
@mgermain the tests pass.

mgermain:
Jenkins test this

caglar:
@mgermain Jenkins test are passing as well.

theano/sandbox/rng_mrg.py
```diff
@@ -1292,8 +1292,17 @@ def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
...
+        nstreams:
+            The number of streams to use for sampling the random numbers
+            in parallel.
+        broadcastable:
+            A tuple of booleans defining the broadcastable pattern of the
+            output tensor.
         """
+        if not (isinstance(broadcastable, tuple) or broadcastable is None):
+            raise ValueError("broadcastable argument to the uniform function"
+                             " should be tuple.")
```
nouiz: add "or None"

theano/sandbox/rng_mrg.py
```diff
@@ -1351,22 +1360,39 @@ def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
...
+                                 " output tensor.")
+            return patternbroadcast(r, broadcastable)
         return r
 
     def binomial(self, size=None, n=1, p=0.5, ndim=None, dtype='int64',
-                 nstreams=None):
+                 nstreams=None, broadcastable=None):
+        if not (isinstance(broadcastable, tuple) or broadcastable is None):
+            raise ValueError("broadcastable argument to the binomial function"
+                             " should be tuple.")
```
nouiz: "or None"

theano/sandbox/rng_mrg.py
```diff
@@ -1390,33 +1416,41 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
         """
         if pvals is None:
             raise TypeError("You have to specify pvals")
+
+        if not (isinstance(broadcastable, tuple) or broadcastable is None):
+            raise ValueError("broadcastable argument to the multinomial function"
+                             " should be a tuple.")
```
nouiz: "or None" I won't repeat it, do it everywhere.

theano/sandbox/rng_mrg.py
```diff
@@ -1292,8 +1292,17 @@ def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
             The output data type. If dtype is not specified, it will be
             inferred from the dtype of low and high, but will be at
             least as precise as floatX.
-
+        nstreams:
+            The number of streams to use for sampling the random numbers
+            in parallel.
+        broadcastable:
```
nouiz: In the issue, I told I wasn't sure of the interface. Was there other discussion about that that I missed? I'm still not sure this is the right way to do it. We don't add extra op in the graph at other place when the user build the graph himself to make the broadcast pattern the one he wants.

What about changing mrg_uniform_base.make_node try to infer the broadcastable flags? If we can know that some of the shapes will be 1 (if they are constant for example). This would be more user friendly, they won't need to specify it manually themself. In the make_node, we could do:
```
for i in ndim:
   if extract_constant(size[[i]) == 1:
      br = True
   else: 
      br = False
```
There is still the case when the input size is a vector that the user provide. In that case, we could just let him add the petternbroadcast himself or keep what this PR does. @dwf it is you that talked about this. If we change the make_node as discussed, would you still benefit from the broadcastable parameter?

@lamblin @abergeron

dwf: I'm always kind of hesitant about being too "clever" about trying to infer what the user meant from incomplete information. Adding broadcastable flags based on simply the shape could introduce unexpected broadcast-related bugs. I'm much more comfortable with a solution that forces the user to specify broadcastability if that is what they want.

nouiz:
I commented in https://github.com/Theano/Theano/issues/4159 what we need to do about this issue.

theano/sandbox/rng_mrg.py
lamblin: I understand the need to have control of the broadcastable flags, however I think it is better for the user to call `.patternbroadcast()` explicitly on the result, rather than adding `, broadcastable=` to the interface all operations that may need it.

nouiz:
Closing as we will fix the problem in a different way.

PR #4412, opened 2016-04-20
Implement Rop for Pool(mode='max')

AndreasMadsen:
Creating a PR as discussed in #4387

_All credit goes to @nouiz and @rodrigob who implemented this in #3587_

/cc @nouiz

PS: I couldn't get all the test to pass on the master branch without any patches applied.

PR #4781, opened 2016-07-25
Resolve conflicts Rop implementation for Pool(mode='max')

alexlee-gk:
Following up on PR #4412 by @AndreasMadsen

Added minor changes for DownsampleFactorMaxRop to inherit from OpenMPOp, analogous to the other pooling classes.

Original implementation was done by @nouiz and @rodrigob in #3587

/cc @nouiz @AndreasMadsen

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -353,6 +353,8 @@ class GpuDot22(GpuOp):
         return hash(type(self))
 
     def make_node(self, x, y):
+        x = as_cuda_ndarray_variable(x)
+        y = as_cuda_ndarray_variable(y)
         if x.type.ndim != 2:
             raise TypeError(x)
         if y.type.ndim != 2:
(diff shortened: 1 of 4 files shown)


AndreasMadsen:
Oh, thanks for taking care of this, I completely forgot about it.

alexlee-gk:
I just noticed that I didn't modify the string returned by `DownsampleFactorMaxRop.c_code` to use `#pragma omp parallel`, so this class is not implementing the `OpenMPOp` contract at all.

We can either (1) fix that string or (2) revert the changes in `DownsampleFactorMaxRop` so that it inherits from `Op` instead of `OpenMPOp`. I can do the latter, but I don't think I'll have time to do the former. Let me know what's the best way to proceed.

nouiz:
The best would be to get this merged. I need someone else to verify the
test.

So if you don't have the time for the OpenMP version, then do the other. I
don't know anyone else who will do it.

alexlee-gk:
Alright, I updated the commit to keep `DownsampleFactorMaxRop` without OpenMP for now.

I also updated the tuples returned by `c_code_cache_version` of the other pooling classes to enforce the method's contract.

aam-at:
I need that. Is that okay if I pick up this pull request to finish it?

alexlee-gk:
It's okay with me. Thanks for taking care of this.

nouiz:
continued in https://github.com/Theano/Theano/pull/5323

Issue #4662, opened 2016-06-23
theano.tensor.roll does not handle shifts greater than axis size


hexahedria:
In numpy, if you call `roll` with a shift greater than the size of the axis, the return value is rolled by shift modulo the axis size. For example:
```python
>>> np.roll([0,1,2,3,4,5],7)
array([5, 0, 1, 2, 3, 4])
```
This makes intuitive sense, since a cyclic rotation should have a period of the axis size.

In Theano, passing a shift greater than the size of the axis results in no shift being applied at all:
```python
>>> T.roll(np.array([0,1,2,3,4,5]),7).eval()
array([0, 1, 2, 3, 4, 5])
```

Based on the [implementation of roll][rollimp], it appears that the roll is being performed using two subtensor ops and a join, with the subtensors being indexed directly by the shift parameter. When the shift is greater than the size of the axis, one of these subtensors gets the whole input, and the other subtensor is empty, causing the incorrect behavior.

[rollimp]: https://github.com/Theano/Theano/blob/d9028c7b27a3bb35c7134e6d34124af68d5f6adf/theano/tensor/basic.py#L4011 

A workaround is to make sure the shift is in the correct bounds by using the modulo operation, i.e.
```python
>>> T.roll(np.array([0,1,2,3,4,5]),7%6).eval()
array([5, 0, 1, 2, 3, 4])
```
But it would be better for the roll function to do this automatically.

nouiz:
thanks for the report. Do you want to make a PR with your suggest and a few tests? We don't have time for this now.

nouiz:
done by gh-4798

PR #4798, opened 2016-07-28
Use shift modulo axis size to handle large shifts in theano.tensor.roll

hexahedria:
This pull request addresses issue #4662 by taking the desired shift modulo the axis size. It also adds tests to verify the correct behavior.


nouiz:
thanks

PR #4750, opened 2016-07-17
Raising error for Indexing subtensor with a Boolean mask

SinaHonari:
Fixes issue #4595

PR #4803, opened 2016-07-28
[WIP] Raising error for Indexing subtensor with a Boolean mask

### LINKED ISSUE #4595: Indexing subtensor with a Boolean mask
In Numpy it is possible to index arrays with a Boolean mask. 
This allows, e.g., this kind of operation:
```python
import numpy as np
my_array = np.array([1, 42, 6])
my_array[my_array == 42] = 0  # my_array --> [1, 0, 6]
```

At the moment Theano not only does not allow this kind of indexing, but also fails silently, considering the boolean mask as 0/1 indexing.

As discussed with @lamblin, part of the problem originates from the fact that Theano has no Boolean type. I think we should either support Boolean mask indexing implicitly when the indices are `int8` (and have the same length as the tensor?) or at least raise an error with the option to suppress it setting a flag.

fvisin:
Continuation of #4750 to fix #4595 .
As we discussed, I am opening the PR to verify how many test fail and decide whether to fix the failing tests or add a boolean type.

### DIFF
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import, print_function, division
+import collections
 import copy
 import traceback as tb
 import warnings
@@ -466,6 +467,25 @@ class _tensor_py_operators(object):
 
     # SLICING/INDEXING
     def __getitem__(self, args):
+
+        def check_bool(args_el):
+            try:
+                if args_el.dtype == 'int8' or args_el.dtype == 'uint8':
+                    # isinstance(args_el, (numpy.bool_, bool)) or \
+                    raise TypeError(('TensorType does not support boolean '
+                                     'mask for indexing such as tensor[x==0]. '
+                                     'If you are indexing on purpose with an '
+                                     'int8, please cast it to int32'))
+            except AttributeError:
+                pass
+
+            if not isinstance(args_el, theano.tensor.Variable) and \
+               isinstance(args_el, collections.Iterable):
+                for el in args_el:
+                    check_bool(el)
+
+        check_bool(args)
+
         if (isinstance(args, list) and
                 any([isinstance(a, slice) for a in args])):
             pass


lamblin:
Didn't we decide some time ago that we could just use int64 for all int literals?
If so, I bet that would solve most of the issues here.
Actually, I'll try it.

lamblin:
So, using a minimum of int32 with the diff below seems to solve most of the issues, except for:
- tests that use int8 explicitly in indexing, and should stop doing so, for instance in `T_get_scalar_constant_value`
- tests that call functions that will use indexing on variables that happen to be int8, in which case those functions should upcast the indices, for instance for `TestBinCountOp`
- The doctest that shows this is an incorrect approach [here](http://deeplearning.net/software/theano_versions/dev/library/tensor/basic.html#indexing), which should be updated anyway to show the new error message

I believe setting the minimum precision to int16 or int32 should be the solution, to be discussed with @nouiz.

```diff
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -302,7 +302,8 @@ class NumpyAutocaster(object):
         # returns either an exact x_==x, or the last cast x_
         return x_
 
-autocast_int = NumpyAutocaster(('int8', 'int16', 'int32', 'int64'))
+autocast_int = NumpyAutocaster(('int32', 'int64'))
 autocast_float = NumpyAutocaster(('float16', 'float32', 'float64'))
```

SinaHonari:
@lamblin : It seems like some functions such as **scan** are also using int8, which make some of tests to fail. if you are coming early afternoon we can talk about this.

lamblin:
Are you talking about `theano.scan_module.tests.test_scan.test_compute_test_value_grad_cast`?
If so, the issue is that the `0` in `outputs_info` [here](https://github.com/fvisin/Theano/blob/f1fc38dd08d659fb2951ad4b077248fc6dec7e9a/theano/scan_module/tests/test_scan.py#L5435) was cast via `as_tensor_variable` (or something equivalent), which made it `int8`.
The diff above (or at least removing `'int8'` from the list) made that test pass when I tried it.

SinaHonari:
yes, that one and `theano.tests.test_rop.test_RopLop`.
If you have already checked, it should be fine then. 
Should we wait for Fred before applying this change (the diff that you showed above)?

lamblin:
I think you can apply it in a new commit, and then start fixing the few remaining errors.

SinaHonari:
Good.
I updated the PR [here](https://github.com/Theano/Theano/pull/4750)

PR #4806, opened 2016-07-29
Fix daily buildbot error Python32bit

nouiz:


slefrancois:
Ok to merge when Travis passes.

PR #4808, opened 2016-07-29
Add the user stack trace printing in nanguardmode

nouiz:


abergeron:
Ok when travis passes

PR #4810, opened 2016-07-29
Confusion matrix 

xoltar:
Fixes issue #3637. This is simply a rebase of #4100 plus updated example to match sklearn.confusion_matrix very closely. API signature is the same as sklearn.confusion_matrix:

`def confusion_matrix(actual, pred)`


lamblin:
Thanks!

PR #4814, opened 2016-08-01
Fix typos and other doc/comment issues, add .idea to .gitignore, and …

xoltar:
…delete apparently unused function theano.compile.mode.check_equal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ Theano.suo
 .pydevproject
 .ropeproject
 core
+.idea
--- a/theano/compile/function.py
+++ b/theano/compile/function.py
@@ -102,7 +102,8 @@ def function(inputs, outputs=None, mode=None, updates=None, givens=None,
         Variables that are neither in "updates" nor in "no_default_updates".
     accept_inplace : bool
         True iff the graph can contain inplace operations prior to the
-        optimization phase (default is False).
+        optimization phase (default is False). *Note* this parameter is unsupported,
+        and its use is not recommended.
     name : str
         An optional name for this function. The profile mode will print the time
         spent in this function.
--- a/theano/compile/function.py
+++ b/theano/compile/function.py
@@ -28,7 +28,7 @@ def function_dump(filename, inputs, outputs=None, mode=None, updates=None,
                   on_unused_input=None,
                   extra_tag_to_remove=None):
     """
-    This is helpful to make a reproducable case for problem during Theano
+    This is helpful to make a reproducible case for problems during Theano
     compilation.
 
     Ex:
@@ -36,13 +36,13 @@ def function_dump(filename, inputs, outputs=None, mode=None, updates=None,
     replace `theano.function(...)` by
     `theano.function_dump('filename.pkl', ...)`.
 
-    If you see this, you where probably asked to use this function to
+    If you see this, you were probably asked to use this function to
     help debug a particular case during the compilation of a Theano
-    function. `function_dump` allows to easily reproduce your
-    compilation without asking any code. It pickle all the objects and
+    function. `function_dump` allows you to easily reproduce your
+    compilation without generating any code. It pickles all the objects and
     parameters needed to reproduce a call to `theano.function()`. This
-    include shared variables and there values. If you do not want
-    that, you can set to replace shared variables values by zeros by
+    includes shared variables and their values. If you do not want
+    that, you can choose to replace shared variables values with zeros by
     calling set_value(...) on them before calling `function_dump`.
 
     To load such a dump and do the compilation:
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -37,4 +37,5 @@ Theano.suo
 .ipynb_checkpoints
 .pydevproject
 .ropeproject
-core
\ No newline at end of file
+core
+.idea
```
nouiz: Out or curiosity, what system generate that?

theano/compile/mode.py
```diff
@@ -18,35 +18,6 @@
 _logger = logging.getLogger('theano.compile.mode')
 
 
-def check_equal(x, y):
```
nouiz: This cause a failure in our test as now, the numpy import isn't used anymore in that file. Can you remove the numpy import from that file?

/home/travis/build/Theano/Theano/theano/compile/mode.py:8:1: F401 'numpy' imported but unused

theano/compile/mode.py
```diff
@@ -422,6 +393,7 @@ def get_mode(orig_string):
             # This might be required if the string is 'ProfileMode'
             from .profilemode import ProfileMode  # noqa
             from .profilemode import prof_mode_instance_to_print
+            # TODO: Can't we look up the name and invoke it rather than using eval here?
```
nouiz: Probably. There is only the modes ProfileMode (deprecated, we can probably remove it now) and MonitorMode that aren't the default Mode.
Mostly, if we add the elif for MonitorMode and Mode and remove ProfileMode, then we could remove that eval.

nouiz:
Thanks for the PR. I just one comment to fix before merging this PR as it make our test fail.

.gitignore
xoltar: This is generated by the Intellij IDEA Python plugin, and probably PyCharm does the same.

xoltar:
@nouiz Removed the numpy import.

theano/compile/function.py
```diff
@@ -100,6 +100,9 @@ def function(inputs, outputs=None, mode=None, updates=None, givens=None,
         If True, do not perform any automatic update on Variables. If False
         (default), perform them all. Else, perform automatic updates on all
         Variables that are neither in "updates" nor in "no_default_updates".
+    accept_inplace : bool
+        True iff the graph can contain inplace operations prior to the
+        optimization phase (default is False).
```
nouiz: Can you add that this is not recomanded to use it and is not supported? It was for that reason that it wasn't documented.

theano/compile/function.py
```diff
@@ -115,10 +118,10 @@ def function(inputs, outputs=None, mode=None, updates=None, givens=None,
         Ops in the graph, an Exception will be raised.
     allow_input_downcast: bool or None
         True means that the values passed as inputs when calling the function
-        can be silently downcasted to fit the dtype of the corresponding
+        can be silently down-casted to fit the dtype of the corresponding
```
nouiz: We use nowhere the version with a dash. Why did you change that? Maybe we are better to be consistent.

But otherwise, I don't have a strong preference for that one.

theano/compile/pfunc.py
```diff
@@ -306,6 +306,9 @@ def pfunc(params, outputs=None, mode=None, updates=None, givens=None,
         If False (default), perform them all. Else, perform automatic updates
         on all Variables that are neither in "updates" nor in
         "no_default_updates".
+    accept_inplace : bool
+        True iff the graph can contain inplace operations prior to the
+        optimization phase (default is False).
```
nouiz: idem

nouiz:
Just a few small comment. thanks.

theano/compile/function.py
xoltar: @nouiz Shall I remove the description and just say it's unsupported, or keep what's there and add the warning to the end?

theano/compile/function.py
xoltar: @nouiz Just because the spell-checker didn't like the one without the dash. I can change it back if you like.

theano/compile/function.py
lamblin: IMO, keep the description and add a warning that it is not recommended.
There are use cases where it can be useful, especially when cloning or unpickling graphs that are already optimized, but it is not the typical case.

theano/compile/function.py
nouiz: My english isn't perferct. I have found downcast in dictionary by googling. I don't have a preference. @lamblin answered for the other comments. I like is suggestion.

lamblin: Either version is fine by me, I'd say don't bother changing it back if it makes the spell-checker happy.

xoltar:
@nouiz Added the required warnings.

nouiz:
thanks

Issue #4812, opened 2016-08-01
Cuda yield CPU while waiting on GPU


lachlants:
Hello, 

I would like the option to yield the CPU when waiting on long cuda kernels, I've found this very useful for freeing CPU resources for multi threaded multi gpu code. It can be enabled by calling: cudaSetDeviceFlags() with cudaDeviceScheduleBlockingSync or cudaDeviceScheduleYield after cudaSetDevice() and before any context creation. 

I've been testing a hacked in version of the above with my own branch and it hasn't caused any issues so far. 

Cheers

nouiz:
Hi,

we are moving to a new GPU back-end called libgpuarray:

https://github.com/Theano/libgpuarray/

It is there that we should add that. There is already a mechanism for
flags, but I don't think it cover that one. @abergeron, did I miss it?

Can you give a more concrete data on the difference it does for you? Can
you give some speed up number? Can you confirm that the use case was 1
python process per GPU and inside it multiple python thread? If not, can
you describe it more? I would like to understand better the cases when that
flag help.

thanks

lachlants:
I don't have alot of concrete data but I observed the processes running kernels went from 100% to ~20% CPU utilization with cudaDeviceScheduleBlockingSync enabled. This approx. halved the time required for a set of background processes to run. You can find a bit of info online about these different flags but it basically boils down to a trade off between kernel latency and CPU utilization i.e. busy wait loop vs polling / interrupt 

My use case is: 4x GPU processes and 4-16x data loading processes (not threads). I'm using Cuda 8.0 + Ubuntu 16.04 with 4x 1080 + i7-6700K. When Cuda kernels were running they consumed all CPU cores (excluding hyperthreading) which significantly slowed down the data loading processes (which require a bit of CPU for augmentation).... making the data loading the bottleneck. This is a fairly typical use case with modern very large datasets (e.g. Imagenet, etc) which can't be completely loaded into memory.

nouiz:
thanks. I made a PR that add a Theano flags to allow this in the new
back-end:

https://github.com/Theano/Theano/pull/4821

PR #4821, opened 2016-08-03
Add the theano flags gpuarray.sched

nouiz:
close gh-4812

### DIFF
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -487,6 +487,21 @@ import theano and print the config variable, as in:
         automatically to get more memory. But this can cause
         fragmentation, see note above.
 
+.. attribute:: config.gpuarray.sched
+
+    String value: ``'default'``, ``'multi'``, ``'single'``
+
+    Default: ``'default'``
+
+    Control the stream mode of contexts.
+
+    The sched parameter passed for context creation to pygpu.  With
+    CUDA, using "multi" mean using the parameter
+    cudaDeviceScheduleYield. This is useful to lower the CPU overhead
+    when waiting for GPU. One user found that it speed up its other
+    process that was doing data augmentation.
+
+
 .. attribute:: config.gpuarray.single_stream
 
     Boolean value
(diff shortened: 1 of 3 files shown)


doc/library/config.txt
```diff
@@ -487,6 +487,21 @@ import theano and print the config variable, as in:
...
+
+    String value: ``'default'``, ``'multi'``, ``'single'``
+
+    Default: ``'default'``
+
+    Control the stream mode of contexts.
+
+    The sched parameter passed for context creation to pygpu.  With
+    CUDA, using "multi" mean using the parameter
+    cudaDeviceScheduleYield. This is useful to lower the CPU overhead
```
abergeron: SInce we use the driver api this is not exactly true but equivalent.

doc/library/config.txt
```diff
@@ -487,6 +487,21 @@ import theano and print the config variable, as in:
...
+    when waiting for GPU. One user found that it speed up its other
```
abergeron: speed up its -> speeds up his

doc/library/config.txt
```diff
@@ -487,6 +487,21 @@ import theano and print the config variable, as in:
...
+    process that was doing data augmentation.
```
abergeron: process -> processes

theano/configdefaults.py
```diff
@@ -242,6 +242,16 @@ def filter_nvcc_flags(s):
              FloatParam(0),
              in_c_key=False)
 
+AddConfigVar('gpuarray.sched',
+             """The sched parameter passed for context creation to pygpu.
+                With CUDA, using "multi" mean using the parameter
+                cudaDeviceScheduleYield. This is useful to lower the
+                CPU overhead when waiting for GPU. One user found that it
+                speed up its other process that was doing data augmentation.
+             """,
```
abergeron: Same corrections as above.

theano/configdefaults.py
```diff
@@ -242,6 +242,16 @@ def filter_nvcc_flags(s):
...
+             EnumStr("default", "multi", "single"),
+             in_c_key=True)
```
abergeron: This should not be in the C key, it doesn't change any code.

nouiz:
I updated the PR.

theano/configdefaults.py
nouiz: I updated the PR.

PR #4824, opened 2016-08-04
Full 3D convolution for conv3d2d

gvtulder:
This patch adds support for `border_mode='full'` to the 3D convolution in `tensor.nnet.conv3d2d`.

`tensor.nnet.conv2d` already provides full convolution for the H and W dimensions, so the code only has to be changed to support full convolution in the time dimension. I've implemented this by padding the stacked 2D output with zeros using `set_subtensor`. Summing over the output of `DiagonalSubtensor` then provides the result of a full convolution. This approach, unfortunately, requires an additional, temporary array for `set_subtensor`.

### DIFF
--- a/theano/tensor/nnet/conv3d2d.py
+++ b/theano/tensor/nnet/conv3d2d.py
@@ -288,7 +288,32 @@ def conv3d(signals, filters,
                 _signals_shape_5d[3] - _filters_shape_5d[3] + 1,
                 _signals_shape_5d[4] - _filters_shape_5d[4] + 1,
             ))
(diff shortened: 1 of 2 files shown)


nouiz:
Do you use an nvidia GPU? Do you know we have 3d convolution from cudnn? It is much faster then this code on GPU and don't request such extra memory.

gvtulder:
Thanks for the tip. No, I was looking for a 3D convolution on CPU. (In fact, I'd like to use this with Keras, which seems to be calling `conv3d2d.conv3d`.)

lamblin:
Thanks!

PR #4826, opened 2016-08-04
Move sandbox.cuda test to 2nd jenkins job

slefrancois:


slefrancois:
jenkins test this please

.jenkins/jenkins_test1.sh
```diff
@@ -9,6 +9,6 @@ source ~/.bashrc
 
 # Test theano core
 PARTS="theano/compat theano/compile theano/d3viz theano/gof theano/misc theano/sandbox/linalg theano/sandbox/tests theano/scalar theano/scan_module theano/sparse theano/tensor theano/tests theano/typed_list"
-THEANO_PARAM="theano --with-timer --timer-top-n 10"
+THEANO_PARAM="${PARTS} --with-timer --timer-top-n 10"
```
nouiz: I don't understand what this PR try to do. Here it don't test theano/sandbox/cuda, so it don't test the old backend.. The comment at the top of this file tell that. The init_gpu_device indicate that.

Do you want test1 to only test the CPU? and test2 to test both GPU back-end?

What this PR try to do?

slefrancois: The gpuarray tests were completing in under 20 mins, while the theano cpu + old backend took almost 2 hours.

The goal is to move the old backend to the same job as the new backend to split time more equally. I forgot to correct that comment. Of course, there are some gpu tests outside the backend folders, hence why init_gpu_device is still set.

slefrancois:
Jenkins completed in 1 hour x 2. There are a few errors in the core and new backend, could someone check if that's from changes in theano or the test scripts? I can't tell myself.

.jenkins/jenkins_test1.sh
```diff
@@ -1,13 +1,14 @@
...
+echo "===== Testing theano core"
 
 # Get environment from worker, necessary for CUDA
 source ~/.bashrc
 
-# Test theano CPU and old GPU backend sandbox.cuda
-THEANO_PARAM="theano --with-timer --timer-top-n 10"
-FLAGS="mode=FAST_RUN,init_gpu_device=gpu"
+# Test theano core
+PARTS="theano/compat theano/compile theano/d3viz theano/gof theano/misc theano/sandbox/linalg theano/sandbox/tests theano/scalar theano/scan_module theano/sparse theano/tensor theano/tests theano/typed_list"
```
nouiz: I don't like this white list. If we add one directory, it won't get executed.

Can you use the -e parameter to exclude directory? We use it in the .travis.yml file. This will prevent futur error.

slefrancois: Sounds like a smarter way to do it indeed.

abergeron:
As I've said in person, it would be really nice to have all the gpu tests in one job and all the other tests in the other.  This will require hunting down stray tests outside of sandbox/cuda that still use the GPU.

lamblin:
I think you can get a starting list of those stray tests from the nightly buildbot of theano/libgpuarray, but that can be for later.

PR #4830, opened 2016-08-05
Fix blas C code having wrong alpha and beta for float16.

abergeron:


nouiz:
I would like that we add a test for the case this fix.

nouiz:
otherwise, I think it is good.

nouiz:
merging now as this fix a bug, but I still would like to see a test for this.

mharradon:
This now warns as creating float64 with warn_float64 = 'warn' - I was under the impression that this was detrimental to performance. Is this not the case? Perhaps this should be "excepted" from warn_float64 if not?

nouiz:
Agreed. We should make this not warm with these flag.

Le 18 oct. 2016 18:05, "Michael Harradon" <notifications@github.com> a
écrit :

PR #4831, opened 2016-08-06
Add device in the compiledir_format

nouiz:


lamblin:
There seem to be many unrelated commits.
Also python3 does not seem to define `reduce` in the builtins, according to the flake8 failure.

nouiz:
I fixed the python 3 problem

nouiz:
travis pass.

PR #4835, opened 2016-08-08
Make the DLT mlp example learn in float16

abergeron:


abergeron:
The also makes the convnet example work.

PR #4840, opened 2016-08-10
Allow indexing using Ellipsis.

NeilGirdhar:
Ellipsis is automatically replaced by an appropriate number of newaxis constants up to the number of dimensions of the argument.
Bumped numpy to 1.9.1.
Bumped scipy to 0.14.0.

### DIFF
--- a/theano/tensor/tests/test_subtensor.py
+++ b/theano/tensor/tests/test_subtensor.py
@@ -1,39 +1,40 @@
-from __future__ import absolute_import, print_function, division
+from __future__ import absolute_import, division, print_function
+
 import logging
 import sys
 import unittest
 
-from nose.plugins.skip import SkipTest
 import numpy
+from nose.plugins.skip import SkipTest
+from nose.tools import assert_equal
+from numpy.testing import assert_array_equal
 from six import StringIO
 from six.moves import xrange
 
 import theano
-from theano.compat import exc_message, izip, PY3
-from theano.compile import DeepCopyOp
-from theano import config
-from theano import gof
 import theano.scalar as scal
 import theano.tensor as tensor
-from theano.tests import unittest_tools as utt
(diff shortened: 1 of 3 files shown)


theano/tensor/tests/test_subtensor.py
```diff
@@ -337,6 +339,30 @@ def test_slice_symbol(self):
...
+                (1, self.sub, numpy.index_exp[..., 1]),
+                (1, self.sub, numpy.index_exp[1, ...]),
+                (1, self.sub, numpy.index_exp[..., 1, 2, 3]),
+                (1, self.sub, numpy.index_exp[1, ..., 2, 3]),
+                (1, self.sub, numpy.index_exp[1, 2, 3, ...]),
+                (3, DimShuffle, numpy.index_exp[..., [0, 2, 3]]),
+                (1, DimShuffle,
+                 numpy.index_exp[numpy.newaxis, ...]),
+                (1, AdvancedSubtensor,
+                 numpy.index_exp[..., numpy.newaxis, [1, 2]])]:
```
lamblin: Older versions of numpy do not support `numpy.newaxis` with advanced indexing, which is why that test was failing with numpy 1.7.
An option may be to skip this test for numpy < 1.9 rather than bumping the requirements.

If we decide to change the requirements, then we should also update the doc pages and requirements.txt files that mention the older version.

NeilGirdhar: Okay, would you mind deciding what you want to do and make this change?

lamblin: I'll update the test, so we can get it merged faster.
Then we can have the discussion about bumping requirements.
Thanks!

lamblin: Done in #4853

PR #4841, opened 2016-08-10
Linting: Pep8 indentation, consistent quotes

Erotemic:
There are no functional changes. I simply fixed a few linting errors.

Several lines in this file were over or under indented. I fixed these 
without moving lines over the 80 character limit. 

There was trailing whitespace. I simply removed it.

I fixed quote consistency. There seems to be a convention that message strings use double quotes and key/flag strings use single quotes. I changed a logging message that used single quotes to use double quotes. 

The biggest change is that there were a few multiline strings of C code defined within functions. 
The multiline strings were indented all the way to the left, which breaks visual nesting of 
 python indentations. I used the builtin textwrap.dedent function to indent these 
lines of code while still preserving the original string representation. 

So, ultimately nothing has changed, the source code just looks nicer.
I'm submitting these changes as a PR before I work on code that will change behavior in this file.

### DIFF
--- a/theano/sandbox/cuda/__init__.py
+++ b/theano/sandbox/cuda/__init__.py
@@ -6,6 +6,7 @@ import os
 import shutil
 import stat
 import sys
+import textwrap
 import warnings
 
 import theano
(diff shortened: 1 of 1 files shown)


lamblin:
Thanks!
Please be aware that we are starting to deprecate the back-end in `theano/sandbox/cuda` in favor of the one in `theano/gpuarray`, so making changes to the behaviour of that file may not be the best investment now.

nouiz:
Don't hesitate to discuss the changes you are wanting to make. As we are switching to a new gpu back-end, it would be great to have what you need there and not in the old back-end. Also, maybe it is there already.

Erotemic:
Its a relatively simple change. I'ts just adding a configuration option to assume that cudnn is available (config.dnn.assume_available) when calling dnn_available(). 
The checks for cudnn significantly slow down script initialization, which is a pain for debugging. 

I've made the changes in the current version and have been using them, but I've been busy so I haven't been able to commit them or to add them to theano/gpuarray (although I checked the file and it should be failry easy to do so). 

The other reason I haven't prioritized this is because after delving in a bit more, I found the checks originate from a third party library (lasange) and not during theano startup. I'm not sure if is desirable to have this option as it essentially disables a function. It is probably more desirable to change the third party library. I bring it up because it only short circuits the expensive part of dnn_available(), and other checks are still run so I can see some value in committing it here.

PR #4843, opened 2016-08-10
Remove useless reshape

lamblin:
This is a rebase of #4389, addressing #4073, and fixing a couple of other issues so that most of #2943 is solved, and so we can update test_useless_tile.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4200,7 +4200,8 @@ def local_useless_reshape(node):
     if input.ndim != output.ndim:
         return False
 
-    # Simple case: both input and output have a single dimension
+    # Simple case: both input and output have a single dimension.
+    # This could hide errors if the user provides inconsistent shapes.
     if (input.ndim == 1 and output.ndim == 1 and
             input.broadcastable == output.broadcastable):
         return [input]
@@ -4277,7 +4278,6 @@ def local_reshape_to_dimshuffle(node):
     or be removed later on.
 
     For example:
-        - reshape(v, (m,)) --> v  # if v.ndim == 1
         - reshape(x, (1, n)) --> dimshuffle{x,0}(reshape(x, (n,))
         - reshape(x, (1, m, 1, n, 1, 1))
           --> dimshuffle{x,0,x,1,x,x}(reshape(x, (m, n)))
@@ -4295,7 +4295,11 @@ def local_reshape_to_dimshuffle(node):
     new_output_shape = []
     index = 0  # index over the output of the new reshape
     for i in xrange(output.ndim):
-        dim = extract_constant(output_shape[i], only_process_constants=False)
+        # Since output_shape is a symbolic vector, we trust extract_constant
+        # to go through however it is formed to see if its i-th element is 1.
+        # We need only_process_constants=False for that.
+        dim = extract_constant(output_shape[i], only_process_constants=False,
+                               elemwise=False)
         if dim == 1:
             dimshuffle_new_order.append('x')
         else:
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4166,8 +4166,10 @@ register_canonicalize(local_reshape_chain(T.Reshape),
 @gof.local_optimizer([T.Reshape])
 def local_useless_reshape(node):
     """
-    Remove Reshape when both the input and the output have a
-    single dimension.
+    Broadcastable dimensions in reshpae are replaced with dimshuffle.
+    For example:
+        - reshape(x, (1, n)) --> dimshuffle{x,0}(reshape(x, (n,))
+        - reshape(x, (1, m, 1, n, 1, 1)) --> dimshuffle{x,0,x,1,x,x}(reshape(x, (m, n)))
 
     """
     op = node.op
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4184,16 +4184,16 @@ def local_useless_reshape(node):
 
     dimshuffle_new_order = []
     new_output_shape = []
-    i = 0  # index over the output of the new reshape
-    import ipdb; ipdb.set_trace()
-    for dim in extract_constant(output_shape, only_process_constants=True):
+    index = 0  # index over the output of the new reshape
+    for i in xrange(output.ndim):
+        dim = extract_constant(output_shape[i], only_process_constants=False)
         if dim == 1:
             dimshuffle_new_order.append('x')
         else:
-            dimshuffle_new_order.append(i)
+            dimshuffle_new_order.append(index)
             new_output_shape.append(dim)
-            i = i + 1
-    if i != output.ndim:
+            index = index + 1
+    if index != output.ndim:
         inner = op.__class__(len(new_output_shape))(input, new_output_shape)
         return [DimShuffle(inner.type.broadcastable, dimshuffle_new_order)(inner)]
 
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4185,7 +4185,7 @@ def local_useless_reshape(node):
     dimshuffle_new_order = []
     new_output_shape = []
     i = 0  # index over the output of the new reshape
-    for dim in output_shape.value:
+    for dim in output_shape.get_scalar_constant_value():
         if dim == 1:
             dimshuffle_new_order.append('x')
         else:
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -586,9 +586,8 @@ def local_dimshuffle_lift(node):
             node.inputs[0].owner is not None and
             isinstance(node.inputs[0].owner.op, DimShuffle)):
         new_order = node.inputs[0].owner.op.new_order
-        new_order = [i for i in new_order if i != 'x']
         input = node.inputs[0].owner.inputs[0]
-        broadcastables = input.broadcastable
+        broadcastables = node.inputs[0].broadcastable
         new_order_of_nonbroadcastables = []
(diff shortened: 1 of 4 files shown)


lamblin:
jenkins test this please

lamblin:
Travis passed, and the CPU part of Jenkins as well, I believe it is ready for merge, if someone could check my changes.

theano/tensor/opt.py
```diff
@@ -4165,15 +4183,131 @@ def f(node):
...
+
+    input = node.inputs[0]
+    output = node.outputs[0]
+    output_shape = node.inputs[1]
+
+    if input.ndim != output.ndim:
+        return False
+
+    # Simple case: both input and output have a single dimension
+    if (input.ndim == 1 and output.ndim == 1 and
```
nouiz: Can you just add a comment here and in the docstring to tell that this could remove user shape errors. Just to document this.

theano/tensor/tests/test_opt.py
```diff
@@ -6164,14 +6176,65 @@ def test_local_reshape(self):
...
+
+    def test_1(self):
+        x = theano.tensor.matrix('x')
+        r = x.reshape(x.shape)
+
+        m0 = theano.compile.get_default_mode()
+        m1 = m0.including('local_useless_reshape')
+        f1 = theano.function([x], r, mode=m1)
+        topo = f1.maker.fgraph.toposort()
+        assert not any(isinstance(n.op, tensor.basic.Reshape) for n in topo)
```
nouiz: Can you explicitly test the case: Shape_i{dim}(input)

Currently it only test explicitly Shape. Maybe the shape_i is introduced by other opt, but we are playing with opt order, so maybe it won't stay like that. So testing it would be safer.

theano/tensor/tests/test_basic.py
```diff
@@ -5114,14 +5114,17 @@ def __init__(self, name, shared=tensor._shared, op=Reshape, mode=None,
...
+    def function(self, inputs, outputs, ignore_empty=False):
         f = function(inputs, outputs, mode=self.mode)
         if self.mode is not None or theano.config.mode != "FAST_COMPILE":
             topo = f.maker.fgraph.toposort()
             topo_ = [node for node in topo if not isinstance(node.op,
                                                              self.ignore_topo)]
-            assert len(topo_) == 1, topo_
-            assert type(topo_[0].op) is self.op
+            if ignore_empty:
+                assert len(topo_) <= 1, topo_
```
nouiz: I would do keep the second assert in case it is of shape 1. Otherwise, I would assert the shape is 0 and would rename ignore_empty to empty.

theano/tensor/opt.py
```diff
@@ -4165,15 +4183,131 @@ def f(node):
...
+
+    input = node.inputs[0]
+    output = node.outputs[0]
+    output_shape = node.inputs[1]
+
+    dimshuffle_new_order = []
+    new_output_shape = []
+    index = 0  # index over the output of the new reshape
+    for i in xrange(output.ndim):
+        dim = extract_constant(output_shape[i], only_process_constants=False)
```
nouiz: Why only_process_constant=False? Can you add at least elemwise=False as otherwise the elemwise cause sometimes too big investigation? Add in a comment why only_process_constant=False is needed.

theano/tensor/opt.py
```diff
@@ -4165,15 +4183,131 @@ def f(node):
...
+            dimshuffle_new_order.append('x')
+        else:
+            dimshuffle_new_order.append(index)
+            new_output_shape.append(dim)
+            index = index + 1
+    if index != output.ndim:
+        inner = op.__class__(len(new_output_shape))(input, new_output_shape)
+        copy_stack_trace(output, inner)
+        new_node = [DimShuffle(inner.type.broadcastable, dimshuffle_new_order)(inner)]
+        copy_stack_trace(output, new_node)
```
nouiz: I let it up to you, but if you want you can add tests for the copy stack trace.

theano/tensor/opt.py
```diff
@@ -4165,15 +4183,131 @@ def f(node):
...
+def local_reshape_to_dimshuffle(node):
+    """
+    Broadcastable dimensions in Reshape are replaced with dimshuffle.
+
+    The goal is to avoid using reshape to add or remove broadcastable
+    dimensions, but use dimshuffle instead, so dimshuffles can cancel out
+    or be removed later on.
+
+    For example:
+        - reshape(v, (m,)) --> v  # if v.ndim == 1
```
nouiz: remove that line. It is the other opt that does this.

theano/tensor/tests/test_opt.py
```diff
@@ -222,32 +224,34 @@ def test_dimshuffle_on_broadcastable(self):
...
-                                  "Reshape{2}(DimShuffle{0}(col), Shape(col))]")
-        dimshuffle_lift.optimize(g)
-        self.assertTrue(str(g) == "[Reshape{1}(vector, Shape(vector)), "
-                                  "Reshape{2}(mat, Shape(mat)), "
-                                  "Reshape{2}(row, Shape(row)), "
-                                  "Reshape{2}(col, Shape(col))]")
-        # Check stacktrace was copied over correctly after opt was applied
-        self.assertTrue(hasattr(g.outputs[0].tag, 'trace'))
+
+def test_useless_dimshuffle_in_reshape():
```
nouiz: rename to test_local_...

theano/tensor/opt.py
```diff
@@ -627,6 +609,42 @@ def local_dimshuffle_lift(node):
...
+    input = node.inputs[0].owner.inputs[0]
+    broadcastables = node.inputs[0].broadcastable
+    new_order_of_nonbroadcast = []
+    for i, bd in zip(new_order, broadcastables):
+        if not bd:
+            new_order_of_nonbroadcast.append(i)
+    no_change_in_order = all(
+        new_order_of_nonbroadcast[i] <= new_order_of_nonbroadcast[i + 1]
+        for i in xrange(len(new_order_of_nonbroadcast) - 1))
+    if no_change_in_order:
```
nouiz: This opt seem good. Just a comment, we could do more. We could add an else clause and keep only the order change of the dimshuffle and remove the added dimensions to canonicalize it. I let this to later if we need it.

theano/tensor/tests/test_opt.py
```diff
@@ -222,32 +224,34 @@ def test_dimshuffle_on_broadcastable(self):
...
+    vector = TensorType(broadcastable=(False,), dtype='float64')('vector')
+    mat = TensorType(broadcastable=(False, False), dtype='float64')('mat')
+    row = TensorType(broadcastable=(True, False), dtype='float64')('row')
+    col = TensorType(broadcastable=(False, True), dtype='float64')('col')
+
+    reshape_dimshuffle_vector = tensor.reshape(vector.dimshuffle('x', 0), vector.shape)
+    reshape_dimshuffle_mat = tensor.reshape(mat.dimshuffle('x', 0, 'x', 1), mat.shape)
```
nouiz: add a test where we do the transpose to verify we don't optimize it.

nouiz:
I finished my review. Just a few small comments.

theano/tensor/opt.py

theano/tensor/tests/test_opt.py
lamblin: OK.

theano/tensor/tests/test_basic.py
lamblin: OK.

theano/tensor/opt.py
lamblin: I will check that.

lamblin: Here, `output_shape` is a `vector` variable, so we would have to dig through the `MakeVector` or `Join` or whatever that comes from, or the `Shape` itself.
I think attempting to match all these cases in the opt would be too ugly, and `elemwise=False` works, so I'm going to leave it at that.
If in the future `extract_constant` gets a `max_depth` parameter, we can revisit it and add something like `max_depth=3`.

nouiz: elemwise=False is good for me.

theano/tensor/opt.py
lamblin: Thanks for the reminder. I looked and `check_stack_trace` is called in `Test_local_reshape_to_dimshuffle.test_1`.

theano/tensor/opt.py
lamblin: Right.

theano/tensor/tests/test_opt.py
lamblin: OK.

theano/tensor/opt.py
lamblin: We could do that, but I don't think this is the right place: here we would only apply it to a dimshuffle inside a reshape, whereas this could be useful elsewhere as well.
In fact, we also do something a bit similar in `local_dimshuffle_lift` when calling `is_dimshuffle_useless`, although it does not canonicalize in all cases.

theano/tensor/tests/test_opt.py
lamblin: OK.
FYI, it is also indirectly tested in T_Tile, which is how I spotted an issue in the first place.

lamblin:
Jenkins test this one more time please

slefrancois:
jenkins test this please

slefrancois:
jenkins test this please

nouiz:
ok to merge when travis pass.

lamblin:
The failure in Jenkins is fixed in another PR.

PR #4850, opened 2016-08-12
Half 3D convolution for conv3d2d

gvtulder:
This adds support for `border_mode = 'half'` to the `conv3d2d.conv3d` method. `border_mode = 'half'` is already supported by `nnet.conv2d`.

This pull request also adds a number of test cases for the method. The existing tests only evaluate odd-sized filters, so I borrowed some code from `sandbox.cuda.test_dnn` to test with even filters as well.

### DIFF
--- a/theano/tensor/nnet/tests/test_conv3d2d.py
+++ b/theano/tensor/nnet/tests/test_conv3d2d.py
@@ -55,18 +55,31 @@ def test_get_diagonal_subtensor_view(wrap=lambda a: a):
 
 
 def pyconv3d(signals, filters, border_mode='valid'):
-    if border_mode == 'full':
-        # zero-pad signals for full convolution
-        Ns, Ts, C, Hs, Ws = signals.shape
-        Nf, Tf, C, Hf, Wf = filters.shape
-        signals_padded = numpy.zeros((Ns, Ts + 2 * (Tf - 1), C,
-                                      Hs + 2 * (Hf - 1), Ws + 2 * (Wf - 1)), 'float32')
-        signals_padded[:, (Tf - 1):(Ts + Tf - 1), :, (Hf - 1):(Hs + Hf - 1),
-                       (Wf - 1):(Ws + Wf - 1)] = signals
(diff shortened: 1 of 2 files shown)


theano/tensor/nnet/tests/test_conv3d2d.py
```diff
@@ -190,3 +217,87 @@ def test_conv3d(border_mode, mode=mode_without_gpu, shared=theano.tensor._shared
     filters = numpy.random.rand(Nf, Tf, C, Hf, Wf).astype('float32')
     utt.verify_grad(lambda s, f: conv3d(s, f, border_mode=border_mode),
                     [signals, filters], eps=1e-1, mode=mode)
+
+
+def get_conv3d_test_cases(experiment='fwd'):
```
nouiz: Why did you added those 2 tests? What wasn't tested in the test above? I don't want to duplicate tests as they are already slow enough.

Otherwise, I'm ok with this PR, but it need a rebase.

thanks.

gvtulder: I don't remember exactly why I added them, except that I was probably afraid that there might be a few edge cases that are not included in the original tests, but I can't see what they are. I'll remove the new tests.

gvtulder:
Rebased to the current master. Thanks for reviewing.

nouiz:
jenkins test this

gvtulder:
The previous [Travis build](https://travis-ci.org/Theano/Theano/builds/161923558) failed because one of the Python 3 jobs took more than two hours to run. That didn't make sense to me, so I hope it will pass now.

nouiz:
thanks.

Why do you continue to work on conv3d2d, while dnn_conv3d is always faster since cudnn v5 I think?

gvtulder:
Thanks for merging. I was using conv3d2d because I don't always run things on a GPU and I didn't want to keep switching between different functions. I've now switched to the AbstractConv3d op, so conv3d2d is less urgent.

PR #4852, opened 2016-08-12
add ddof to var and std, apply Fred-fix to opt

Faruk-Ahmed:
#4651


ballasn:
Looks good to me; if @nouiz  is ok, we can merge this pr.

PR #4853, opened 2016-08-12
Allow indexing using Ellipsis.

lamblin:
Ellipsis is automatically replaced by an appropriate number of newaxis constants up to the number of dimensions of the argument.

This is a rebase of @NeilGirdhar's PR #4840, but skipping a case instead of bumping the required version of numpy.


lamblin:
Closes #4767.

PR #4855, opened 2016-08-15
gpuarray: fix typo in string

ozancaglayan:


theano/gpuarray/opt.py
```diff
@@ -81,7 +81,7 @@ def query(self, *tags, **kwtags):
 gpu_seqopt.register('gpuarray_graph_optimization', GraphToGPUDB(), -0.5,
                     'fast_compile', 'fast_run', 'gpuarray')
 
-gpu_seqopt.register('gpuarray_local_optimiziations', gpu_optimizer, 1,
+gpu_seqopt.register('gpuarray_local_optimizations', gpu_optimizer, 1,
```
nouiz: thanks for the fix. To prevent regression in case other people used that bad sentences, can you add after 'gpuarray' the old string?

Thanks

ozancaglayan:
You mean in the commit message?

theano/gpuarray/opt.py
```diff
@@ -81,7 +81,7 @@ def query(self, *tags, **kwtags):
...
                     'fast_compile', 'fast_run', 'gpuarray')
```
nouiz: I mean, replace this line by:

'fast_compile', 'fast_run', 'gpuarray', 'gpuarray_local_optimizations')

The strings parameters can be used by the user to filter which optimization get applied.

ozancaglayan:
I hope you meant like this.

nouiz:
thanks

PR #4861, opened 2016-08-15
Add tensor5, dtensor5 etc.

gvtulder:
Some Theano methods ask for 5D inputs, but there are no built-in functions to create those. Perhaps there's a good reason not to add them, but I think it would be useful.


nouiz:
I don't have an objection, but I would like  the opinion of someone else about that. Which method need that?

Also, it would need to be added in the doc under Theano/doc

lamblin:
I think 3D convolutions need them, so it makes sense to have them named, thanks.

As @nouiz said, the documentation, in particular https://github.com/Theano/Theano/blob/master/doc/tutorial/adding.txt#L176, https://github.com/Theano/Theano/blob/master/doc/library/tensor/basic.txt#L85, and the table below that I believe you can regenerate through https://github.com/Theano/Theano/blob/master/doc/generate_dtype_tensor_table.py

gvtulder:
Indeed, I use them for 3D convolutions. I've updated the documentation.

nouiz:
For me it seem good. @lamblin ? We need to wait for travis to finish to make sure it there isn't problems in the doc.

lamblin:
OK for me.

nouiz:
thanks

PR #4865, opened 2016-08-16
Pre cleanup

nouiz:


slefrancois:
jenkins test this

slefrancois:
jenkins test this please

Sentient07:
This PR is causing slow down and some errors while printing profile. (There is one more error that is there in the master apart from what this PR introduces) 
There is an error in profiling when i remove the second commit (the latest one on this PR along with an error fix)
But the optimizer timing doesn't get affected with only the first commit. I think passing the optimizers as a list to `pre_greedy_local_optimizer` is causing the error in printing.

theano/tensor/opt.py
```diff
@@ -4883,14 +4883,25 @@ def local_useless_elemwise_comparison(node):
         return
     if node.op.scalar_op.nin != 2:
         return
+
+    def zeros_like(model, dtype):
+        ret = T.zeros_like(node.inputs[0], dtype=node.outputs[0].dtype)
```
nouiz: this introduce a but. It should be model and not node.inputs[0]. I did a PR to fix it:

https://github.com/Theano/Theano/pull/4891

PR #4866, opened 2016-08-17
Fix jenkins

lamblin:
These should fix the existing issues on Jenkins.

### DIFF
--- a/theano/gradient.py
+++ b/theano/gradient.py
@@ -1369,8 +1369,10 @@ class numeric_grad(object):
     # perfectly accurate.
     type_eps = {'float64': 1e-7,
                 'float32': 3e-4,
+                'float16': 1e-3,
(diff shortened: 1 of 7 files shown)


theano/gradient.py
```diff
@@ -1606,12 +1608,13 @@ def verify_grad(fun, pt, n_tests=2, rng=None, eps=None,
     pt = [numpy.array(p) for p in pt]
 
     for i, p in enumerate(pt):
-        if p.dtype not in ('float32', 'float64'):
```
nouiz: line 1575 1592 need update in the docstring. For 1592, maybe just tell that it don't work for float16 for now?

nouiz:
I'll merge now to help fix the tests. My small comment is just about docstring update.

PR #4867, opened 2016-08-17
Fix #4809

### LINKED ISSUE #4809: Documentation says "will be supported in 0.6rc4" when it is verion 0.8.2 now
http://deeplearning.net/software/theano/library/tensor/basic.html#indexing

317070:
Fix issue #4809, it has annoyed me for a while as well...


nouiz:
Thanks. Seem good to me. We need to wait for travis to pass before merging.

Issue #4770, opened 2016-07-21
Insidious random data/memory corruption bug causing incorrect computation and training divergence


xuancong84:
It seems that sometimes by chance, Theano's (for all versions including bleeding-edge) internal memory state can get corrupted silently, with all subsequent training/testing operations produces erroneous results without throwing any exceptions/warnings. The error will accumulate until some point when the training starts to always diverge. The problem can be solved by aborting the current process, reloading the last-known good model and resuming training.

My code does iterative training, it saves the model at every iteration. After every iteration, it checks the loss function, if it increases, then load the previous iteration's model and reduce learning rate by half. The bug starts to emerge randomly somewhere in between 20-80th iteration (for my code), starting from which the loss keeps increasing after every iteration despite the fact that the learning rate is halved each time and the model keeps falling back to the previous iteration. This process continues until the learning rate becomes so small, <1e-200, and thus, there is no change in the weight or the loss function.

For example,

```
> 1310.31767385 Training acc= 0.971412992678      
> Testing Iter 50 :                               
> Test accuracy = 0.40635451505                   
> Training Iter 51 :                              
> 1213.0716554 Training acc= 0.976086617853       
> Testing Iter 51 :                               
> Test accuracy = 0.407190635452                  
> Training Iter 52 :                              
> 1248.2027879 Training acc= 0.969699330114       
> Testing Iter 52 :                               
> Test accuracy = 0.409698996656                  
> Falling back to Iter 51 ; learn_rate= 0.00125   
> Training Iter 52 :                              
> 1394.18119661 Training acc= 0.964246767409      
> Testing Iter 52 :                               
> Test accuracy = 0.418060200669                  
> Falling back to Iter 51 ; learn_rate= 0.000625  
> Training Iter 52 :                              
> 2215.10956366 Training acc= 0.937139741393      
> Testing Iter 52 :                               
> Test accuracy = 0.413879598662                  
> Falling back to Iter 51 ; learn_rate= 0.0003125 
> Training Iter 52 :                              
> 3660.30786752 Training acc= 0.897647608662      
> Testing Iter 52 :                               
> Test accuracy = 0.413879598662                  
> Falling back to Iter 51 ; learn_rate= 0.00015625
> Training Iter 52 :                              
> 5290.97243645 Training acc= 0.863919613647      
> Testing Iter 52 :                               
> Test accuracy = 0.420568561873                  
> Falling back to Iter 51 ; learn_rate= 7.8125e-05
```

[...]

PR #4871, opened 2016-08-17
scan/scan_op: Convert known_grads to OrderedDict

ozancaglayan:
This was probably causing a different order of operation
during gradient computation in scan for each run. With this
fix I'm able to finally reproduce results on my RNN system.

Can be related to #4770 as well.


nouiz:
I'm ok with this. Just waiting for travis to pass. I'll rebase my branch with the extra assert after that.

PR #4873, opened 2016-08-17
Grad assert

nouiz:
Add assert to prevent problem like https://github.com/Theano/Theano/pull/4871 fixes.

To be merged after gh-4871


nouiz:
I amended the last commit with the pep8 fix from travis failure.

lamblin:
I'm OK with that change, but I'm wondering why theano.grad cannot itself perform some deterministic sorting of the keys of `known_grads`, for instance according to the toposort order.

nouiz:
That would be another good fix

PR #4846, opened 2016-08-11
Handle filter_dilation in the abstractconv_grad* perform methods

ballasn:

PR #4874, opened 2016-08-18
Fix abstractconv_grad dilation and enable border_mode tests

gvtulder:
I think there are a few bugs in commit bbc749ab516 (e.g., lines [1116-L1119 of abstract_conv.py](https://github.com/ballasn/Theano/blob/bbc749ab516a1242d5eb359d1d34cf5fbd031167/theano/tensor/nnet/abstract_conv.py#L1116-L1119), see also #4846), which would give problems with non-symmetric filter dilations for border_modes full or half and with custom padding (which should not be multiplied with the dilation factor).

These errors would have been visible only in the debug mode, if the test had worked. It seems that the test only used the default `border_mode = 'valid'`, but the errors occur with `full` and `half`. This commit should fix that.

Finally, this commit adds `border_mode = "half"` to the tests.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -966,14 +966,15 @@ class AbstractConv2d_gradWeights(BaseAbstractConv2d):
                 '"valid", "full", "half", an integer or a pair of'
                 ' integers'.format(mode))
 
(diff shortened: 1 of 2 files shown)


lamblin:
Jenkins, test this please.

ballasn:
Thanks for the fix, looks good to me, @lamblin this PR is ok to merge

PR #4876, opened 2016-08-18
Cgt opt

Sentient07:


nouiz:
/home/travis/build/Theano/Theano/theano/gof/opt.py:2046:45: E127 continuation line over-indented for visual indent
/home/travis/build/Theano/Theano/theano/gof/opt.py:2047:45: E127 continuation line over-indented for visual indent
/home/travis/build/Theano/Theano/theano/gof/opt.py:2048:45: E127 continuation line over-indented for visual indent
/home/travis/build/Theano/Theano/theano/gof/opt.py:2049:45: E127 continuation line over-indented for visual indent

Sentient07:
Yeah, i've fixed the PEP8 errors. I'm still trying to figure out why there is slow down. I've tried reverting almost all the changes, yet i'm not able to see the timing that we got earlier

theano/compile/profiling.py
```diff
@@ -54,7 +54,7 @@ def _atexit_print_fn():
         destination_file = open(config.profiling.destination, 'w')
 
     for ps in _atexit_print_list:
-        if ps.fct_callcount >= 1 or ps.compile_time > 1:
+        if ps.fct_callcount > 1 or ps.compile_time > 1:
```
nouiz: Remove this change from the PR.

Sentient07: I've added this file to be ignored. Wonder why it keeps adding back again and again

theano/gof/opt.py
```diff
@@ -1261,10 +1289,72 @@ def tracks(self):
...
+                else:
+                    self.node_created[opt] += len(graph.ops(fgraph.variables, repl))
+                    self.applied_true[opt] += 1
+                    if not multiple_opts or not repl[0].owner:
+                        return repl
+                    assert len(repl) == 1
+                    # Ensuring not the input of graph
+                    assert repl[0].owner
+                    new_node = repl[0].owner
+                    apply_mult_opts(new_node, fgraph, True, repl)
```
nouiz: This won't change repl. So the second optimization if it apply won't be applied. Can you fix this problem in the next few hours? I would like to merge this PR soon to go to the next phase.

theano/tensor/opt.py
```diff
@@ -1755,7 +1730,7 @@ def local_fill_to_alloc(node):
                        1.52, 'fast_run')
 
 
-@register_canonicalize("fast_compile")
+@register_useless
```
nouiz: Don't remove the old register.

theano/gof/opt.py
```diff
@@ -1312,7 +1312,7 @@ def apply_mult_opts(node, fgraph, multiple_opts=False, prev_repl=None):
                     # Ensuring not the input of graph
                     assert repl[0].owner
                     new_node = repl[0].owner
-                    apply_mult_opts(new_node, fgraph, True, repl)
+                    repl = apply_mult_opts(new_node, fgraph, True, repl)
```
nouiz: That won't work. It will cause all optimization to be skiped. repl will be None. I think it should be:

new_repl = apply...
if new_repl:
    repl = new_repl

Test it. Make a break point and verify manually that this is doing what we want.

theano/tensor/opt.py
nouiz: You forgot to fix this comment.

theano/compile/profiling.py
nouiz: I think it wasn't removed. Not being added back. Ignoring a file won't revert the changes done in the PR.

theano/tensor/opt.py
Sentient07: I remember you asking me to remove canonicalize, but i wasn't able to get the comments

nouiz: I don't have an idea why I would have asked that. Maybe I wasn't clear. Anyway, just add it back.

theano/tensor/opt.py
```diff
@@ -1677,6 +1677,7 @@ def local_fill_sink(node):
 register_canonicalize(local_fill_sink)
 
 
+@register_canonicalize("fast_compile")
```
nouiz: you added it at the wrong place...

Sentient07: Ah! I'm sorry about this. The email from github showed the function before this. I'll add it at the correct place

Sentient07:
Travis is getting stopped at this `theano.tensor.signal.tests.test_pool.TestDownsampleFactorMax`. But that test passes for me locally. What could be going wrong ?

Sentient07:
I figured out the reason why there was error with the printing at the end of profiling and why there was a slow down. It was caused from either of the commits in this PR, https://github.com/Theano/Theano/pull/4865 
I will break down and see which commit caused the slowdown and which diff exactly did that.

nouiz:
I'm restarting the killed travis and the travis that was too long. We'll see how they goes.

nouiz:
jenkins test this

nouiz:
About the error in profiling, can you show me the error? The link between that PR and profiling is small. Did you find a fix for it?

Sentient07:
This is the error that is occurring, 
```
Error in atexit._run_exitfuncs:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 103, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1256, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 337, in print_profile
    level=level + 1)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 2588, in print_profile
    o.print_profile(stream, prof, level + 2)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 892, in print_profile
    callback_time, callbacks_time, nb_merged, nb_constant) = prof
ValueError: too many values to unpack
Error in sys.exitfunc:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 103, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1256, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 337, in print_profile
    level=level + 1)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 2588, in print_profile
    o.print_profile(stream, prof, level + 2)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 892, in print_profile
    callback_time, callbacks_time, nb_merged, nb_constant) = prof
ValueError: too many values to unpack
Segmentation fault: 11
```

Sentient07:
This error occurs with this branch, but there is another error that happens with master while profiling the same code. I am still to test this case, but i think it is because of passing the parameter to `pre_greedy_local_optimizer` as a list. Because, if i revert this commit, https://github.com/Theano/Theano/pull/4865/commits/99f22fb4e9424155a9d68548f642b6487e92b030

nouiz:
you didn't finish your last sentence.

Sentient07:
.... this is the error that I am getting , (after reverting the commits in the PR reported) 

``` 
rror in atexit._run_exitfuncs:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 103, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1256, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 339, in print_profile
    level=level + 1)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 2446, in print_profile
    o.print_profile(stream, prof, level + 2)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 899, in print_profile
    nb_fail, nb_merged, nb_constant), file=stream)
TypeError: %d format: a number is required, not TopoOptimizer
Error in sys.exitfunc:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 103, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1256, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 339, in print_profile
    level=level + 1)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 2446, in print_profile
    o.print_profile(stream, prof, level + 2)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 899, in print_profile
    nb_fail, nb_merged, nb_constant), file=stream)
TypeError: %d format: a number is required, not TopoOptimizer
```

Sentient07:
To verify that this isn't occurring from this PR, i profiled on master and i still get this error. So i'll open a new issue for this.

nouiz:
I restarted those 2 travis part.

theano/gof/opt.py
```diff
@@ -1261,10 +1289,78 @@ def tracks(self):
...
+                      t, a_t, count, o, n_c), file=stream)
+            print(blanc, '  %.3fs - in %d optimization that were not used (display those with runtime greater than 0)' % (
+                not_used_time, len(not_used)), file=stream)
+            not_used.sort(key=lambda nu: (nu[0], str(nu[1])))
+            for (t, o) in not_used[::-1]:
+                if t > 0:
+                    # Skip opt that have 0 times, they probably wasn't even tried.
+                    print(blanc + "  ", '  %.3fs - %s' % (t, o), file=stream)
+        else:
+            print(blanc, "--- The Optimizer wasn't successful ---", file=stream)
```
nouiz: Change the way it is printed. In your profile, it looks ugly. Add an extra level to the blanc and remove the dash.

theano/gof/opt.py
```diff
@@ -1953,12 +2047,60 @@ def print_profile(stream, prof, level=0):
         print(blanc, "  init io_toposort", io_t, file=stream)
         print(blanc, "  loop time", loop_t, file=stream)
         print(blanc, "  callback_time", callback_time, file=stream)
+        if isinstance(lopt, LocalOptGroup):
+            lopt.print_profile(stream, (lopt.time_opts,
+                                        lopt.time_nodes,
+                                        lopt.process_count,
+                                        lopt.applied_true,
+                                        lopt.node_created))
```
nouiz: Add the level parameter to that function call.
level=level+1

Otherwise, you loose the indentation.

Sentient07:
The travis is still stopping. Also, the ValueError has originated from this PR. I rebased the with master and I don't see that error in master. But would travis stop because of that error ?

Sentient07:
I debugged using pdb and I understood the error. It is because the `print_profile` of `TopoOptimizer` takes in an extra parameter (`local_opt`) for printing the details of the LocalOptGroup. I will fix this in the upcoming commit

Sentient07:
To fix it, I have applied this diff. 
```
iff --git a/theano/gof/opt.py b/theano/gof/opt.py
index 38c6f3c..7780012 100644
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -333,6 +333,10 @@ class SeqOptimizer(Optimizer, list):
                 print(blanc, '  %.6fs - %s' % (t, opt), file=stream)
 
             if sub_profs[i]:
+                if not isinstance(opts[i], TopoOptimizer):
+                    if len(sub_profs[i]) == 8:
+                        sub_profs[i].pop()
+                    assert len(sub_profs[i]) == 7
                 opts[i].print_profile(stream, sub_profs[i],
                                       level=level + 1)
         print(file=stream)
``` 
I am not sure whether you'd approve of this. Is this good enough for merge or is this a hack ?

Sentient07:
Ignore my previous commit. I had to do the check in a different way and what I have done is a hack. I can see what is happening, but I'm not able to understand or reason out. The parameters to be passed to `print_profile` of `TopoOptimizer` is being passed to that of `MergeOptimizer` . And hence due to the mismatch of number of parameters there was `ValueError` and now since the first parameter in `TopoOptimizer`'s `print_profile` is `opt` there is 
`TypeError: %d format: a number is required, not TopoOptimizer`  as reported in https://github.com/Theano/Theano/issues/4892 . 
But what I am not able to understand is why this is happening

theano/gof/opt.py
```diff
@@ -333,10 +333,6 @@ def print_profile(stream, prof, level=0):
                 print(blanc, '  %.6fs - %s' % (t, opt), file=stream)
 
             if sub_profs[i]:
-                if isinstance(opts[i], MergeOptimizer):
-                    if len(sub_profs[i]) == 8:
-                        sub_profs[i] = sub_profs[i][:-1]
-                    assert len(sub_profs[i]) == 7
```
Sentient07: I could have used the fix which Fred used in the other PR, but that didn't strike me

theano/gof/opt.py
```diff
@@ -2053,12 +2014,39 @@ def print_profile(stream, prof, level=0):
         print(blanc, "  loop time", loop_t, file=stream)
         print(blanc, "  callback_time", callback_time, file=stream)
         if isinstance(lopt, LocalOptGroup):
-            lopt.print_profile(stream, (lopt.time_opts,
```
Sentient07: From my analysis of removing commits, I suspect this call to print_profile could be the reason for slow down. Because, it all boils down to the commit where I did this. But, I am not able to figure out why that would happen. @lamblin could you please explain ?

Sentient07: Might not be the call, but could be that we're having a `staticmethod` in a LocalOptimzier

theano/gof/opt.py
```diff
@@ -1305,7 +1301,7 @@ def apply_mult_opts(node, fgraph, multiple_opts=False, prev_repl=None):
                 opt_start = time.time()
                 repl = opt.transform(node)
                 opt_finish = time.time()
-                self.time_opts[opt] += opt_start - opt_finish
+                self.time_opts[opt] += opt_finish - opt_start
```
Sentient07: By mistake, i did the subtraction the otherway. Till today, I had thought it was because of the strings that i moved around while printing

Sentient07:
Yayy! And now the Travis passes! 🎉 🎆 🍻

nouiz:
I just don't understand why the diff fix this problem. Can you add new commits that will undo your fix? Just to see if the real fix could be travis giving more memory to each jobs.

Also, I don't like this change. It is not in the spirit of the rest of the library, so it would make maintenant harder in long term. So I think we should understand why this fix it and find another work-around if it isn't travis change that fix it by the side.

Sentient07:
> I just don't understand why the diff fix this problem. Can you add new commits that will undo your fix? Just to see if the real fix could be travis giving more memory to each jobs.

I initially thought it was the circular reference that is taking place is the reason for the memory leak.  By circular reference, I meant the TopoOptimzier 's constructor takes in the local_opt variable then it's parent Class NavigatorOptimizer modifies it by `local_opt.transform()`, (here the local_opt is `LocalOptGroup`) and then we try to use the same variable from LocalOptGroup class.

> Also, I don't like this change. It is not in the spirit of the rest of the library, so it would make maintenant harder in long term. So I think we should understand why this fix it and find another work-around if it isn't travis change that fix it by the side.

Yes, the earlier approach was better, I agree. That would have enabled us to print profiling results of LocalOptGroup from which ever GlobalOptimizer, not only TopoOptimizer. If we have to do that now, we need to copy paste a lot of code which I don't think is good either.

Sentient07:
I will make two changes, 
1) Revert 4aef3e2289a9fa394e3a32df31a3f69cfb4ec4c4 and restart Travis
2) Enable the travis to restart process after doing 1 ( This is known to work)

Sentient07:
BTW, there was an awful mistake that i fixed in the most recent commit. I realized that mistake much later than applying what i thought was fix. Fingers crossed and hoping that it is the reason

theano/gof/opt.py
nouiz: I don't think having the print_profile can be the problem. But maybe keeping the results of profiling. Can you modify the method the execute the opt to return None instead of the profiling information? This will remove the information saved. If this is the extra information that raise the memory usage, then we will see the cause.

nouiz:
After the rebase, travis still pass. But the "fix" isn't a good fix I think. Also travis have this new error ```

ERROR: test_useless_alloc_with_shape_one (theano.tensor.tests.test_opt.Test_local_canonicalize_alloc)

  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_opt.py", line 3732, in test_useless_alloc_with_shape_one

    alloc_lift = out2in(local_canonicalize_alloc)

NameError: global name 'local_canonicalize_alloc' is not defined

```

Can you remove your "fix" by adding back the print_profile method. I don't think it is the extra parameter that you remove the problem. Make sure travis run with the fix removed to know if we can still repeat it. 

Then make the proposed investigation I suggested elsewhere. I think this new version of the suggestion is better: In LocalOptGroup.transform, comment all lines add add dicts like self.* the profiling information. Maybe the Opt is the same for all Theano function, then this cause the dict to contain all the nodes for all call to theano.function(). During tests, there is many of them. This could explain the problem.

jenkins test this

File: theano/gof/opt.py (hunk 9 above)
Comment (Sentient07) [inline theano/gof/opt.py]: I'm not able to see this comment as i think you had commented on the page
which was open in your browser and after I rebased, github doesn't show
this comment.
I currently don't return anything from that method. I think I have found
the error. I had forgotten to remove the time_nodes method and have
initialized the keys of the dict with opts and tried to access with the key
as node. That could be the problem. But because of the bad rebase i'm not
able to figure that out. I will do the rebase again and push that change.
That change has been done in
https://github.com/Sentient07/Theano/tree/fixed-cgt-opt.

ᐧ

**Comment by Sentient07** [comment]
As i assumed previously and as @nouiz and @lamblin said it was a dict that was initialized wrongly that caused this problem. I have now updated with the fix which i tried in the other branch of mine and travis passed, https://travis-ci.org/Sentient07/Theano/builds/156882811

**Comment by Sentient07** [comment]
The new TopoDB doesn't seem to have test. Do I add a new one ?

**Comment by nouiz** [comment]
It is tested indirectly. Theano tests is already too slow, so I don't think
it is useful to add an explicit test.

On Fri, Sep 2, 2016 at 8:15 AM, Ramana Subramanyam <notifications@github.com

**Comment by Sentient07** [comment]
There is a small problem before it gets merged. I wonder why travis doesn't test the profiling part. I'll add one more commit

File: theano/gof/opt.py [hunk 11]
```diff
@@ -335,6 +333,10 @@ def print_profile(stream, prof, level=0):
                 print(blanc, '  %.6fs - %s' % (t, opt), file=stream)
 
             if sub_profs[i]:
+                if isinstance(opts[i], MergeOptimizer):
+                    if len(sub_profs[i]) == 8:
+                        sub_profs[i] = sub_profs[i][:-1]
+                    assert len(sub_profs[i]) == 7
```
nouiz: I think this diff isn't needed anymore. I did another fix that don't require this hack.

theano/gof/opt.py
```diff
@@ -1261,10 +1294,82 @@ def tracks(self):
...
+                    continue
+                else:
+                    if self.profile:
+                        self.node_created[opt] += len(graph.ops(fgraph.variables, repl))
+                        self.applied_true[opt] += 1
+                    if not multiple_opts or not repl[0].owner:
+                        return repl
+                    assert len(repl) == 1
+                    # Ensuring not the input of graph
+                    assert repl[0].owner
```
nouiz: remove this assert. It is useless. We have the condition on line 1317 that make sure it is always true.

theano/gof/opt.py
```diff
@@ -1261,10 +1294,82 @@ def tracks(self):
...
+                    self.process_count[opt] += 1
+                if not repl:
+                    continue
+                else:
+                    if self.profile:
+                        self.node_created[opt] += len(graph.ops(fgraph.variables, repl))
+                        self.applied_true[opt] += 1
+                    if not multiple_opts or not repl[0].owner:
+                        return repl
+                    assert len(repl) == 1
```
nouiz: move this assert before the if.

theano/gof/opt.py
```diff
@@ -2448,6 +2601,10 @@ def print_profile(stream, prof, level=0):
                     print(blanc, "merge not implemented for ", o)
             for o, prof in zip(opt.cleanup_optimizers, cleanup_sub_profs[i]):
                 try:
+                    if isinstance(o, MergeOptimizer):
+                        if len(prof) == 8:
+                            prof = prof[:-1]
+                            assert len(prof) == 7
```
nouiz: Remove. I think I fixed the root problem, so this work around isn't needed anymore.

theano/gof/opt.py
```diff
@@ -201,7 +195,12 @@ def warn(exc, self, optimizer):
 
     def __init__(self, *opts, **kw):
         """
-        WRITEME
+        Parameters
+        ----------
+        opts : List
```
abergeron: List -> list

theano/gof/opt.py
```diff
@@ -201,7 +195,12 @@ def warn(exc, self, optimizer):
...
+        The List of optimizers to be applied to a node
+        kw : Dict
```
abergeron: Dict -> dict

theano/gof/opt.py
```diff
@@ -201,7 +195,12 @@ def warn(exc, self, optimizer):
 
     def __init__(self, *opts, **kw):
         """
-        WRITEME
+        Parameters
+        ----------
+        opts : List
+        The List of optimizers to be applied to a node
```
abergeron: This should be indented by 4 spaces.

theano/gof/opt.py
```diff
@@ -201,7 +195,12 @@ def warn(exc, self, optimizer):
...
+        kw : Dict
+        Dictonary containing failure callback. The only supported keyword is `failure_callback`.
```
abergeron: This too.

theano/gof/opt.py
abergeron: Actually no, this is not a list argument.  This it the list of all arguments.  Just document it as `*opts` with no type.

theano/gof/opt.py
abergeron: Here you should document the supported keyword args and mention that they are keyword-only.

theano/gof/opt.py
```diff
@@ -1233,20 +1232,50 @@ def decorator(f):
...
     """
-    WRITEME
+    Takes a list of LocalOptimizer and applies them to the nodes.
+    If apply_all_opts is False, it will return after the first optimizer applied.
+    Otherwise, it will start again with the node returned by the previous optimizer.
 
+    Parameters
+    ----------
+    optimizers : List
+    List of optimizers to be applied to a node
```
abergeron: Same as above this does not take a list, but multiple arguments.

theano/gof/opt.py
```diff
@@ -1233,20 +1232,50 @@ def decorator(f):
...
+    kwargs : Dict
+    Dictionary containing apply_all_opts parameter. When True, multiple optimizers are applied.
```
abergeron: Same here document the supported keyword arguments.

theano/gof/opt.py
```diff
@@ -1261,10 +1290,82 @@ def tracks(self):
...
+                    assert len(repl) == 1
+                    if self.profile:
+                        self.node_created[opt] += len(graph.ops(fgraph.variables, repl))
+                        self.applied_true[opt] += 1
+                    if not multiple_opts or not repl[0].owner:
+                        return repl
+                    # Ensuring not the input of graph
+                    assert repl[0].owner
+                    new_node = repl[0].owner
+                    new_repl = apply_mult_opts(new_node, fgraph, True, repl)
```
abergeron: Shouldn't we stop adding recursion?

nouiz: done as well as all other comment.

nouiz:
I checked and it still execute correctly:

   0.285784s - ('useless', 'TopoOptimizer', 3, 7787, 7664) - 0.002s

jenkins test this

PR #4877, opened 2016-08-18
N2 fast destroy

Sentient07:
WIP : Faster cyclic detection. This work was paused in May and will be resumed soon. 

To do list : 
- [ ] Make a Theano flag fast_inplace that default to False. Use the new new algo only when it is True.
- [ ]  Make sure all tests pass when the flag is False
- [ ] Make sure most tests pass when the flag is True. For failing test, investigate it this is a normal failure with the new algo. Is so, change the test to expect the failure when using the new algo and expect the old behaviour when using the old algo.
- [ ] Make an even faster version that only check the changed part of the graph by optimization, not the full graph.

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -1104,9 +1104,10 @@ AddConfigVar('cmodule.preload_cache',
              BoolParam(False, allow_override=False),
              in_c_key=False)
 
-AddConfigVar('fast_inplace',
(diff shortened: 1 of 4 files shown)


Sentient07:
Could someone stop travis on this PR for a while ? I'd resume work on this only after #4876 and the CleanUp PRs

nouiz:
this was done in another PR, so closing.

Issue #4823, opened 2016-08-03
Incompatible on nanguardmode and sandbox.rng_mrg?


lovecambi:
I try to debug with nanguardmode to find the where my code induces the NAN. However, I find this mode will always throw an NAN error if your random number generator is theano.sandbox.rng_mrg. Thus I cannot locate where the real NAN is.

Below is a simple example to show this issue. f1 and f1_nan are basically the same function except for the later one with nanguardmode, and f1 involves random numbers. f2 has no random numbers, and f2_nan is similar but with nanguardmode.

```python
import numpy as np
import theano
from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
import theano.tensor as T
from theano.compile.nanguardmode import NanGuardMode

_srng = RandomStreams(np.random.randint(1, 2147462579))

mu = T.matrix()
logvar = T.matrix()
eps = _srng.normal(mu.shape)

res1 = (mu + T.exp(0.5*logvar) * eps).mean()
f1 = theano.function([mu, logvar], res1, allow_input_downcast=True)
f1_nan = theano.function([mu, logvar], res1, allow_input_downcast=True, mode=NanGuardMode(nan_is_error=True, inf_is_error=True, big_is_error=True))

res2 = (mu + T.exp(0.5*logvar)).mean()
f2 = theano.function([mu, logvar], res2, allow_input_downcast=True)
f2_nan = theano.function([mu, logvar], res2, allow_input_downcast=True, mode=NanGuardMode(nan_is_error=True, inf_is_error=True, big_is_error=True))

v = np.random.normal(size=(100,64))
m = np.random.normal(size=(100,64))

f1(m,v)
#output: array(0.02410086616873741, dtype=float32)

f1_nan(m, v)
#output: AssertionError: NaN detected
#        NanGuardMode found an error in an input of the graph.

f2(m,v)
#output: array(1.1354453563690186, dtype=float32)

f2_nan(m, v)
#output: array(1.1354453563690186, dtype=float32)
```

In addition, I found the CudaNdarray with NAN always has the shape (92160,) regardless of the input shape, which I guess is the pre-allocated vector in GPU for further computation?

nouiz:
Can you try with Theano dev version? I think it was fixed since the last
release.

lovecambi:
I installed the Theano dev version and I am sure the program run with the new version, but the same error appears.

f0k:
[...]

PR #4881, opened 2016-08-19
Ignore more frequently the mrg rng state in the old back-end. 

nouiz:
fix gh-4823


f0k:
Looks reasonable! Cannot test this before Monday, though. @grrrr: Hooray!

Issue #4731, opened 2016-07-12
3d pooling on CPU


nouiz:
cuDNN support 3d pooling, but we don't have a CPU implementation. Some people are using a graph of Op to generate that on CPU, but it causing that we don't use the cudnn 3d pooling and so end up being not efficient.

see gh-4377

A first quick version could be python only with only python code, so I tag it with Python code Only". But ideally we should have a C version.

EugenePY:
Hi, @nouiz I'm currently work on this issue. I had read [gh-4377](https://github.com/Theano/Theano/issues/4377). We need to add 3d pooling support for theano.tesnsor.sinal.Pool and restricted to 5d tensor (which same with [cuDNN](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/dnn.py#L558). Is that right?

gvtulder:
Hi @EugenePY, have you had a look at the pull request #4884 linked above? There's code for 3D CPU pooling there.

EugenePY:
@gvtulder yeah I just saw that. Thx !

nouiz:
done by gh-4884

PR #4884, opened 2016-08-21
theano.tensor.signal.Pool with 3D support

gvtulder:
This would add a 3D CPU implementation for `theano.tensor.signal.Pool` and its gradients, which can then be optimised with a GPU implementation.

(Actually, it turns the Pool op into a more generic Nd pooling operation, which may or may not be a good idea.)

See also issue #4731.

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1877,6 +1877,7 @@ def local_gpua_pool_dnn_alternative(op, ctx_name, inputs, outputs):
         return
     img = gpu_contiguous(as_gpuarray_variable(img, ctx_name))
     mode = op.mode
+    # dnn_pool expects exactly 2 non-pooling dimensions
     if img.ndim == nd + 2:
         return dnn_pool(img, ws, stride=stride, pad=pad, mode=mode)
     else:
@@ -1903,6 +1904,7 @@ def local_gpua_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
     out_grad = gpu_contiguous(as_gpuarray_variable(out_grad, ctx_name))
     mode = op.mode
 
+    # the GPU ops expect exactly 2 non-pooling dimensions
     if inp.ndim == nd + 2:
         return GpuDnnPoolGrad(mode=mode)(inp,
                                          out,
@@ -1940,12 +1942,14 @@ def local_gpua_avg_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
     out_grad = gpu_contiguous(as_gpuarray_variable(out_grad, ctx_name))
     mode = op.mode
 
+    # the GPU ops expect exactly 2 non-pooling dimensions
     if inp.ndim == nd + 2:
         # We reuse out_grad because cuDNN does not use the value of the `out`
         # argument but still checks its shape for average pooling. This
         # has been observed in v2 and v3 as far as I know.
         return GpuDnnPoolGrad(mode=mode)(inp, out_grad, out_grad, ws, stride, pad)
     else:
+        # reshape to 4D or 5D with 2 non-pooling dimensions
         inp_padded = pad_dims(inp, 2, nd)
         out_grad_padded = pad_dims(out_grad, 2, nd)
         ret_padded = GpuDnnPoolGrad(mode=mode)(inp_padded,
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -35,7 +35,7 @@ from .nnet import GpuSoftmax
 from .opt import (gpu_seqopt, register_opt,
                   op_lifter, register_opt2)
 
-from .opt_util import alpha_merge, output_merge, inplace_allocempty
+from .opt_util import alpha_merge, output_merge, inplace_allocempty, pad_dims, unpad_dims
 
 from theano.configdefaults import SUPPORTED_DNN_CONV_ALGO_BWD_FILTER
 
@@ -1253,7 +1253,7 @@ class GpuDnnPoolGrad(DnnBase):
         return [shape[0]]
 
 
-def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
+def dnn_pool(img, ws, stride=None, mode='max', pad=None):
     """
     GPU pooling using cuDNN from NVIDIA.
 
@@ -1267,13 +1267,13 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
     img
         Images to do the pooling over.
     ws : tuple
-        Subsampling window size.
+        Subsampling window size.  Should have 2 or 3 elements.
     stride : tuple
-        Subsampling stride (default: (1, 1)).
+        Subsampling stride (default: (1, 1) or (1, 1, 1)).
     mode : {'max', 'average_inc_pad', 'average_exc_pad', 'sum'}
     pad : tuple
         (padX, padY) or (padX, padY, padZ)
-        default: (0, 0)
+        default: (0, 0) or (0, 0, 0)
 
     .. warning:: The cuDNN library only works with GPU that have a compute
         capability of 3.0 or higer.  This means that older GPU will not
@@ -1285,6 +1285,10 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
 
     """
     img = gpu_contiguous(img)
+    if stride is None:
+        stride = (1,) * len(ws)
+    if pad is None:
+        pad = (0,) * len(ws)
     if mode == "sum":
         ret = GpuDnnPool(mode="average_inc_pad")(img, ws, stride, pad)
         context_name = ret.type.context_name
@@ -1868,9 +1872,18 @@ def local_gpua_pool_dnn_alternative(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     img, ws, stride, pad = inputs
-    img = as_gpuarray_variable(img, ctx_name)
+    nd = op.ndim if op.ndim else (img.ndim - 2)
+    if nd not in (2, 3):
+        return
+    img = gpu_contiguous(as_gpuarray_variable(img, ctx_name))
     mode = op.mode
-    return dnn_pool(gpu_contiguous(img), ws, stride=stride, pad=pad, mode=mode)
+    if img.ndim == nd + 2:
+        return dnn_pool(img, ws, stride=stride, pad=pad, mode=mode)
+    else:
+        # reshape to 4D or 5D with 2 non-pooling dimensions
+        img_padded = pad_dims(img, 2, nd)
+        ret_padded = dnn_pool(img_padded, ws, stride=stride, pad=pad, mode=mode)
+        return unpad_dims(ret_padded, img, 2, nd)
 
 
 @register_opt('cudnn', 'fast_compile')
@@ -1882,17 +1895,33 @@ def local_gpua_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     inp, out, out_grad, ws, stride, pad = inputs
-    inp = as_gpuarray_variable(inp, ctx_name)
-    out = as_gpuarray_variable(out, ctx_name)
-    out_grad = as_gpuarray_variable(out_grad, ctx_name)
+    nd = op.ndim if op.ndim else (inp.ndim - 2)
+    if nd not in (2, 3):
+        return
+    inp = gpu_contiguous(as_gpuarray_variable(inp, ctx_name))
+    out = gpu_contiguous(as_gpuarray_variable(out, ctx_name))
+    out_grad = gpu_contiguous(as_gpuarray_variable(out_grad, ctx_name))
     mode = op.mode
 
-    return GpuDnnPoolGrad(mode=mode)(gpu_contiguous(inp),
-                                     gpu_contiguous(out),
-                                     gpu_contiguous(out_grad),
-                                     ws,
-                                     stride,
-                                     pad)
+    if inp.ndim == nd + 2:
+        return GpuDnnPoolGrad(mode=mode)(inp,
+                                         out,
+                                         out_grad,
+                                         ws,
+                                         stride,
+                                         pad)
+    else:
+        # reshape to 4D or 5D with 2 non-pooling dimensions
+        inp_padded = pad_dims(inp, 2, nd)
+        out_padded = pad_dims(out, 2, nd)
+        out_grad_padded = pad_dims(out_grad, 2, nd)
+        ret_padded = GpuDnnPoolGrad(mode=mode)(inp_padded,
+                                               out_padded,
+                                               out_grad_padded,
+                                               ws,
+                                               stride,
+                                               pad)
+        return unpad_dims(ret_padded, inp, 2, nd)
 
 
 @register_opt('cudnn', 'fast_compile')
@@ -1904,16 +1933,28 @@ def local_gpua_avg_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     inp, out_grad, ws, stride, pad = inputs
-    inp = as_gpuarray_variable(inp, ctx_name)
-    out_grad = as_gpuarray_variable(out_grad, ctx_name)
+    nd = op.ndim if op.ndim else (inp.ndim - 2)
+    if nd not in (2, 3):
+        return
+    inp = gpu_contiguous(as_gpuarray_variable(inp, ctx_name))
+    out_grad = gpu_contiguous(as_gpuarray_variable(out_grad, ctx_name))
     mode = op.mode
 
-    cg = gpu_contiguous(out_grad)
-
-    # We reuse cg because cuDNN does not use the value of the `out`
-    # argument but still checks its shape for average pooling. This
-    # has been observed in v2 and v3 as far as I know.
-    return GpuDnnPoolGrad(mode=mode)(gpu_contiguous(inp), cg, cg, ws, stride, pad)
+    if inp.ndim == nd + 2:
+        # We reuse out_grad because cuDNN does not use the value of the `out`
+        # argument but still checks its shape for average pooling. This
+        # has been observed in v2 and v3 as far as I know.
+        return GpuDnnPoolGrad(mode=mode)(inp, out_grad, out_grad, ws, stride, pad)
+    else:
+        inp_padded = pad_dims(inp, 2, nd)
+        out_grad_padded = pad_dims(out_grad, 2, nd)
+        ret_padded = GpuDnnPoolGrad(mode=mode)(inp_padded,
+                                               out_grad_padded,
+                                               out_grad_padded,
+                                               ws,
+                                               stride,
+                                               pad)
+        return unpad_dims(ret_padded, inp, 2, nd)
 
 
 @register_opt('cudnn', 'fast_compile')
--- a/theano/gpuarray/opt_util.py
+++ b/theano/gpuarray/opt_util.py
@@ -3,12 +3,12 @@ from functools import wraps
 
 import numpy
 
-from theano import scalar as scal, Constant
+from theano import tensor, scalar as scal, Constant
 from theano.gof import local_optimizer
 from theano.tensor import (DimShuffle, get_scalar_constant_value,
                            NotScalarConstantError)
 
-from .basic_ops import GpuFromHost, HostFromGpu, GpuAllocEmpty, gpu_alloc_empty
+from .basic_ops import GpuFromHost, HostFromGpu, GpuAllocEmpty, GpuReshape, gpu_alloc_empty
 from .elemwise import GpuDimShuffle, GpuElemwise
 
 _one = scal.constant(numpy.asarray(1.0, dtype='float32'))
@@ -329,3 +329,48 @@ def inplace_allocempty(op, idx):
             return maker(node, inputs)
         return opt
     return wrapper
+
+
+def pad_dims(input, leftdims, rightdims):
+    """Reshapes the input to a (leftdims + rightdims) tensor
+
+    This helper function is used to convert pooling inputs with arbitrary
+    non-pooling dimensions to the correct number of dimensions for the
+    GPU pooling ops.
+
+    This reduces or expands the number of dimensions of the input to
+    exactly `leftdims`, by adding extra dimensions on the left or by
+    combining some existing dimensions on the left of the input.
+    """
+    assert input.ndim >= rightdims
+
+    if input.ndim == (leftdims + rightdims):
+        return input
+
+    # extract image dimensions
+    img_shape = input.shape[-rightdims:]
+
+    # count the number of "leading" dimensions, store as dmatrix
+    batch_size = tensor.prod(input.shape[:-rightdims])
+    batch_size = tensor.shape_padright(batch_size, 1)
+
+    # store in the required shape, for example as a 4D tensor
+    # with shape: (batch_size,1,height,width)
+    new_shape = tensor.cast(tensor.join(0, batch_size,
+                                        tensor.as_tensor([1] * (leftdims - 1)),
+                                        img_shape), 'int64')
+    input_ND = GpuReshape(leftdims + rightdims)(input, new_shape)
+    return input_ND
+
+
+def unpad_dims(output, input, leftdims, rightdims):
+    """Reshapes the output after pad_dims.
+
+    This reverts the padding by `pad_dims`.
+    """
+    if output.ndim == input.ndim:
+        return output
+
+    # restore the output to the original shape
+    outshp = tensor.join(0, input.shape[:-rightdims], output.shape[-rightdims:])
+    return GpuReshape(input.ndim)(output, outshp)
--- a/theano/gpuarray/opt_util.py
+++ b/theano/gpuarray/opt_util.py
@@ -341,6 +341,24 @@ def pad_dims(input, leftdims, rightdims):
     This reduces or expands the number of dimensions of the input to
     exactly `leftdims`, by adding extra dimensions on the left or by
     combining some existing dimensions on the left of the input.
+
+    Use `unpad_dims` to reshape back to the original dimensions.
+
+    Examples
+    --------
+    Given input of shape (3, 5, 7), ``pad_dims(input, 2, 2)``
+    adds a singleton dimension and reshapes to (3, 1, 5, 7).
+    Given that output from pad_dims, ``unpad_dims(output, input, 2, 2)``
+    reshapes back to (3, 5, 7).
+
+    Given input of shape (3, 5, 7, 9), ``pad_dims(input, 2, 2)``
+    does not reshape and returns output with shape (3, 5, 7, 9).
+
+    Given input of shape (3, 5, 7, 9, 11), ``pad_dims(input, 2, 2)``
+    combines the first two dimensions and reshapes to (8, 7, 9, 11).
+
+    Given input of shape (3, 5, 7, 9), ``pad_dims(input, 2, 3)``
+    adds a singleton dimension and reshapes to (3, 1, 5, 7, 9).
     """
     assert input.ndim >= rightdims
 
@@ -350,15 +368,25 @@ def pad_dims(input, leftdims, rightdims):
     # extract image dimensions
     img_shape = input.shape[-rightdims:]
 
-    # count the number of "leading" dimensions, store as dmatrix
-    batch_size = tensor.prod(input.shape[:-rightdims])
-    batch_size = tensor.shape_padright(batch_size, 1)
+    non_pool_ndim = input.ndim - rightdims
+    if non_pool_ndim < leftdims:
+        # too few dimensions, pad on the left
+        dummy_dims = tensor.as_tensor([1] * (leftdims - non_pool_ndim))
+        new_shape = tensor.join(0, dummy_dims,
+                                input.shape[:non_pool_ndim],
+                                img_shape)
+    else:
+        # too many dimensions, combine the leading dimensions
+        batched_ndim = non_pool_ndim - leftdims + 1
+        batch_size = tensor.prod(input.shape[:batched_ndim])
+        # convert to a vector for tensor.join
+        batch_size = tensor.shape_padright(batch_size, 1)
+        new_shape = tensor.join(0, batch_size,
+                                input.shape[batched_ndim:non_pool_ndim],
+                                img_shape)
 
-    # store in the required shape, for example as a 4D tensor
-    # with shape: (batch_size,1,height,width)
-    new_shape = tensor.cast(tensor.join(0, batch_size,
-                                        tensor.as_tensor([1] * (leftdims - 1)),
-                                        img_shape), 'int64')
+    # store in the required shape
+    new_shape = tensor.cast(new_shape, 'int64')
     input_ND = GpuReshape(leftdims + rightdims)(input, new_shape)
     return input_ND
 
--- a/theano/gpuarray/opt_util.py
+++ b/theano/gpuarray/opt_util.py
@@ -347,7 +347,7 @@ def pad_dims(input, leftdims, rightdims):
     Examples
     --------
     Given input of shape (3, 5, 7), ``pad_dims(input, 2, 2)``
-    adds a singleton dimension and reshapes to (3, 1, 5, 7).
+    adds a singleton dimension and reshapes to (1, 3, 5, 7).
     Given that output from pad_dims, ``unpad_dims(output, input, 2, 2)``
     reshapes back to (3, 5, 7).
 
@@ -355,10 +355,10 @@ def pad_dims(input, leftdims, rightdims):
     does not reshape and returns output with shape (3, 5, 7, 9).
 
     Given input of shape (3, 5, 7, 9, 11), ``pad_dims(input, 2, 2)``
-    combines the first two dimensions and reshapes to (8, 7, 9, 11).
+    combines the first two dimensions and reshapes to (15, 7, 9, 11).
 
     Given input of shape (3, 5, 7, 9), ``pad_dims(input, 2, 3)``
-    adds a singleton dimension and reshapes to (3, 1, 5, 7, 9).
+    adds a singleton dimension and reshapes to (1, 3, 5, 7, 9).
     """
     assert input.ndim >= rightdims
 
--- a/theano/gpuarray/tests/config.py
+++ b/theano/gpuarray/tests/config.py
@@ -17,7 +17,7 @@ test_ctx_name = None
 
 if theano.config.mode == 'FAST_COMPILE':
     mode_with_gpu = theano.compile.mode.get_mode('FAST_RUN').including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray', 'gpu')
+    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray')
 else:
     mode_with_gpu = theano.compile.mode.get_default_mode().including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_default_mode().excluding('gpuarray', 'gpu')
+    mode_without_gpu = theano.compile.mode.get_default_mode().excluding('gpuarray')
--- a/theano/gpuarray/tests/config.py
+++ b/theano/gpuarray/tests/config.py
@@ -17,7 +17,7 @@ test_ctx_name = None
 
 if theano.config.mode == 'FAST_COMPILE':
     mode_with_gpu = theano.compile.mode.get_mode('FAST_RUN').including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray')
+    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray', 'gpu')
 else:
     mode_with_gpu = theano.compile.mode.get_default_mode().including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_default_mode().excluding('gpuarray')
+    mode_without_gpu = theano.compile.mode.get_default_mode().excluding('gpuarray', 'gpu')
--- a/theano/gpuarray/tests/test_dnn.py
+++ b/theano/gpuarray/tests/test_dnn.py
@@ -258,9 +258,7 @@ def test_pooling():
             def fn(x):
                 return pool_2d(x, (ws, ws), ignore_border=True,
                                padding=pad, mode=mode)
-            utt.verify_grad(fn, [data],
-                            cast_to_output_type=False,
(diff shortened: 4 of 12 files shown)


theano/sandbox/cuda/dnn.py
```diff
@@ -1387,19 +1387,24 @@ def c_compiler(self):
...
+            pad = (0,) * len(ws)
+
         assert len(ws) == len(stride) and len(stride) == len(pad)
         assert len(ws) in (2, 3)
         self.ws = ws
         self.stride = stride
         self.pad = pad
 
-        if (pad[0] != 0 or pad[1] != 0) and version() == -1:
+        if max(pad) > 0 and version() == -1:
```
nouiz: You can remove this if, we don't support cudnn v2 anymore.

theano/gpuarray/dnn.py
```diff
@@ -1860,6 +1860,8 @@ def local_gpua_pool_dnn_alternative(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     img, = inputs
+    if len(op.ds) not in (2, 3) or img.ndim != len(op.ds) + 2:
```
nouiz: If the second condition is false, there is a much bigger problem. The graph isn't consistent. I'm fine to keep that condition there, but we should catch that earlier. I think it is done elsewhere.

theano/sandbox/cuda/dnn.py
```diff
@@ -1972,28 +1966,28 @@ def infer_shape(self, node, shape):
...
     mode : {'max', 'average_inc_pad', 'average_exc_pad}
-    pad :
-        (pad_h, pad_w) padding information.
-        pad_h is the number of zero-valued pixels added to each of the top and
-        bottom borders.
-        pad_w is the number of zero-valued pixels added to each of the left
-        and right borders.
+    pad
+        Padding information.  A tuple with 2 or 3 integers, specifying the
+        number of zero-valued pixels to be added on both sides of the input.
```
nouiz: Can you keep the extra information from the old description? It was telling which element of the tuple correspond to which dimensions.

theano/sandbox/cuda/opt.py
```diff
@@ -1899,7 +1899,10 @@ def local_gpu_downsample_factor_max(node):
 
         assert node.op.__props__ == ('ds', 'ignore_border', 'st', 'padding',
                                      'mode')
-        if node.op.padding != (0, 0) or node.op.mode != 'max':
+        if (len(node.op.ds) not in (2, 3) or
```
nouiz: This is our old pooling. It support only 2d. So make the condition that node.op.ds == 2.

theano/sandbox/cuda/opt.py
```diff
@@ -1913,10 +1916,12 @@ def local_gpu_downsample_factor_max_grad(node):
     if (isinstance(node.op, pool.MaxPoolGrad) and node.op.ds == node.op.st):
         assert node.op.__props__ == ('ds', 'ignore_border', 'st', 'padding',
                                      'mode')
-        if (node.op.padding != (0, 0) or
+        if (len(node.op.ds) not in (2, 3) or
```
nouiz: idem, old pooling that support only 2d.

theano/sandbox/cuda/opt.py
```diff
@@ -1933,7 +1938,10 @@ def local_gpu_downsample_factor_max_grad_grad(node):
     if isinstance(node.op, pool.DownsampleFactorMaxGradGrad):
         assert node.op.__props__ == ('ds', 'ignore_border', 'st',
                                      'padding', 'mode')
-        if (node.op.padding != (0, 0) or
+        if (len(node.op.ds) not in (2, 3) or
```
nouiz: idem, only 2d supported.

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -418,26 +416,20 @@ def test_pooling3d():
     if not cuda.dnn.dnn_available() or cuda.dnn.version() < (3000, 3000):
         raise SkipTest(cuda.dnn.dnn_available.msg)
 
-    mode_without_gpu_ref = theano.compile.mode.get_mode(
```
nouiz: Using this mode was explicit. We don't want the default mode that could be DebugMode for the reference. That would make the tests too slow in DebugMode. So add this back. Was there a reason why you removed it instead of just code clean up? Maybe add a comment so that the next person understand why it is there. Something like:

# We force the FAST_RUN as we don't want the reference to run in DebugMode.

theano/tensor/signal/pool.py
```diff
@@ -113,30 +114,103 @@ def pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
...
+    ds : list or tuple of N ints
         Downsample factor over rows and column.
         ds indicates the pool region size.
     ignore_border : bool
         If ds doesn't divide imgshape, do we include an extra row/col
         of partial downsampling (False) or ignore it (True).
-    st : list or tuple of two ints or None
-        Stride size, which is the number of shifts over rows/cols to get the
+    st : list or tuple of N ints or None
+        Stride size, which is the number of shifts over rows/cols/slices to get the
```
nouiz: Are you sure it is not slices/rows/cols instead?

theano/tensor/signal/pool.py
```diff
@@ -113,30 +114,103 @@ def pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
...
+
+class Pool(OpenMPOp):
+    """
+    This Op downsamples the last N dimensions of the input by taking the max,
+    sum or average over different patches.
 
     Parameters
     ----------
-    ds : list or tuple of two ints
+    ds : list or tuple of N ints
```
nouiz: I would be more specific then N, I would tell "2 or 3 ints"

theano/tensor/signal/pool.py
```diff
@@ -262,46 +322,42 @@ def __init__(self, ds, ignore_border=False, st=None, padding=(0, 0),
...
 
     def perform(self, node, inp, out):
         x, = inp
         z, = out
-        if len(x.shape) != 4:
+        nd = len(self.ds)
+        if len(x.shape) < nd + 1:
             raise NotImplementedError(
-                'Pool requires 4D input for now')
+                'Pool requires input with {} or more dimensions'.format(nd + 1))
```
nouiz: The op only support 4 or 5 inputs. Why you write that it support more like 6 dimensions? The pool_2d/3d interface support more, but it collapse some outer dimensions and reuse the 2d/3d pool op.

theano/tensor/signal/pool.py
```diff
@@ -310,23 +366,25 @@ def perform(self, node, inp, out):
...
+        for i in xrange(nd):
+            for j in xrange(pool_out_shp[i]):
+                start = j * st[i]
+                end = builtins.min(start + ds[i], img_shp[i])
+                if not inc_pad:
+                    start = builtins.max(start, pad[i])
+                    end = builtins.min(end, img_shp[i] - pad[i])
+                region_slices[i].append(slice(start, end))
+
+        # iterate over non-pooling dimensions
```
nouiz: You seem to implement that now, The Pool op can pool over an inputs that is more then 4 or 5 dims and do that in the op instead of doing reshape in the graph. Why did you bother with that? The reshape op is free when the colapsed dimensions are contiguous in memory. Which is most of the time.

I'm fine to keep that. I'm more interrested to know your motivation for that.

theano/tensor/signal/pool.py
```diff
@@ -113,30 +114,103 @@ def pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
...
+            " False). To have consistent behavior with all Theano"
+            " version, explicitly add the parameter ignore_border=True."
+            " On the GPU, using ignore_border=True is needed to use cuDNN."
+            " When using ignore_border=False and not using cuDNN, the only"
+            " GPU combination supported is when"
+            " `ds == st and padding == (0, 0, 0) and mode == 'max'`."
+            " Otherwise, the convolution will be executed on CPU.",
+            stacklevel=2)
+        ignore_border = False
+    if input.ndim == 5:
```
nouiz: If the op support doing in the Op the extra dimensions, we could here always add a Pool op, but we will need to move that logic that add the reshapes in the optimization that move the op on the GPU to keep them working in that case. Otherwise, the op won't do the extra dimensions itself, it will be done in the graph.

theano/tensor/signal/pool.py
```diff
@@ -357,279 +415,368 @@ def c_headers(self):
...
-                "padding must be (0,0) when ignore border is False");
-              %(fail)s;
-            }
+        for (int i=0; i<%(nd)s; i++)
+        {
+            r[i] = PyArray_DIMS(%(x)s)[%(non_pool_ndim)s + i] + 2 * pd[i];
+        }
+        """
+
+        if not self.ignore_border and any(self.padding[i] > 0 for i in xrange(nd)):
```
nouiz: There is a PR that make the padding as an input of the node. To help the conflict resolution of both PR, can you keep that in the C code?

theano/tensor/signal/pool.py
```diff
@@ -146,98 +220,82 @@ class Pool(OpenMPOp):
...
-            if ds[0] == st[0]:
-                nr = r // st[0]
-            else:
-                out_r = (r - ds[0]) // st[0] + 1
-                if isinstance(r, theano.Variable):
-                    nr = tensor.maximum(out_r, 0)
+        if padding is None:
+            padding = (0,) * nd
+        patch_shape = tuple(tensor.extract_constant(imgshape[-nd + i]) +
+                            (padding[i] * 2 if padding[i] else 0)
```
nouiz: why the "if padding[i] else 0"? padding[i] is false only when it is 0. So a simpler code is the old code "padding[i] * 2" without the if/else. Did I miss something? You use that also in the PoolGrad op.

theano/tensor/signal/pool.py
```diff
@@ -310,23 +366,25 @@ def perform(self, node, inp, out):
...
-                    for c in xrange(pc):
-                        col_st = c * st1
-                        col_end = builtins.min(col_st + ds1, img_cols)
-                        if not inc_pad:
-                            col_st = builtins.max(col_st, self.padding[1])
-                            col_end = builtins.min(col_end,
-                                                   x.shape[-1] + pad_w)
-                        zz[n, k, r, c] = func(y[
-                            n, k, row_st:row_end, col_st:col_end])
+        # precompute the region boundaries for each dimension
```
nouiz: you seem to have optimized the python code. That is great. Just to let you know that it isn't needed as it is only used for debugging. We can keep the optimized code without problem.

theano/tensor/signal/pool.py
```diff
@@ -357,279 +415,368 @@ def c_headers(self):
...
+                else:
+                    return max(0, (v - 1 - downsample) // stride + 1) + 1
 
-        rval = list(imgshape[:-2]) + [nr, nc]
+        out_shape = [compute_out(patch_shape[i], ds[i], st[i]) for i in xrange(nd)]
+
+        rval = list(imgshape[:-nd]) + out_shape
         return rval
 
     def __init__(self, ds, ignore_border, st=None, padding=(0, 0), mode='max', openmp=None):
```
nouiz: Make padding=None by default.

theano/tensor/signal/pool.py
```diff
@@ -113,30 +114,103 @@ def pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
     return tensor.reshape(output, outshp, ndim=input.ndim)
 
 
-class Pool(OpenMPOp):
+def pool_3d(input, ds, ignore_border=None, st=None, padding=(0, 0, 0),
```
nouiz: make ignore_border defaul to True. It is the new interface we want to use. As this is a new user interface function, I think it would be better to use the new interface for that parameter.

theano/tensor/signal/pool.py
```diff
@@ -907,12 +1150,213 @@ def grad(self, inp, grads):
...
+            } // for loop over non-pooling dimensions
+        } // if z_prod
+        """
+        return ccode % locals()
+
+    def c_code_cache_version(self):
+        return (0, 1, self.openmp)
+
 
 class DownsampleFactorMaxGradGrad(OpenMPOp):
```
nouiz: The change to this class isn't tested. It need new tests.

theano/tensor/signal/tests/test_pool.py
```diff
@@ -708,6 +738,32 @@ def mp(input):
...
+                                                       [True, False],
+                                                       ['max', 'sum',
+                                                        'average_inc_pad',
+                                                        'average_exc_pad']):
+                # print 'maxpoolshp =', maxpoolshp
+                # print 'ignore_border =', ignore_border
+                numpy_output_val = self.numpy_max_pool_nd(imval, maxpoolshp,
+                                                          ignore_border,
+                                                          mode=mode)
+                output = pool_3d(images, maxpoolshp, ignore_border,
```
nouiz: This method will add reshape in the graph when the input isn't 5 dims. So it won't test the new code that make the op support a variable number of inputs dimensions.

If you want to keep that change, it will request more work. So why do you want that feature?

nouiz:
Thanks for this big PR. It look good. I didn't review completely all perform() and c_code() method yet. I'll try later to do so.

One problem is that we have 2 PR that change those Ops. The other is: https://github.com/Theano/Theano/pull/4756

We will need to see how we reconcile them. So mostly in which order we merge them.

theano/tensor/signal/pool.py
gvtulder: I don't have clear enough overview of the consequences, but are you really sure that is a good idea? That would give two almost identical functions, the only difference 2D/3D, with a very subtle difference in one of the default parameters.

theano/tensor/signal/pool.py
gvtulder: Mainly because it was done like that in the original 2D code as well.
Looking at it again, I think that in the original version it has the advantage of not adding anything if the padding is 0. If the imgshape is a Theano variable, I can imagine that this is slightly more efficient because it doesn't add a + 0 operation. But yes, in my revised version this no longer holds. I'll remove it.

theano/tensor/signal/pool.py
gvtulder: The reason I did that was mainly because I made the unoptimized version really, really slow, which would mess up the tests. Computing the list of `slice` objects for each region takes a lot of time, especially in 3D.

gvtulder:
Thanks for the comments so far. Two main comments:

* I don't have a very strong argument for why the pooling ops should support an arbitrary number of non-pooling dimensions. I added it, mainly, because it was relatively easy to do once you add the 2D/3D options, and because 2 is a (somewhat) arbitrary number. The reshaping feels a bit strange as well. The main reason for fixing it to 2, I think, would be that the GPU ops require that.
<br>
My preference would be to add some tests for the arbitrary dimensions and move the reshaping to exactly 2 non-pooling dimensions to the GPU optimizations, but I would be also be okay with removing it and fixing it to 2 in the CPU version.

* Is there a standard for naming and ordering the three dimensions? There seem to be a few different systems in the code. In some places it's x/y/z, in others it's height/width/depth, or depth/height/width, and I think I even saw depth/width/height, I think. That's one of the reasons I tried to avoid naming them in the comments.
<br>
Which form should I use? And what are the correct terms to use for the third dimension? So far, I used 'depth' made up of individual 'slices'. Is the padding then added to the 'front' and 'back'?

theano/gpuarray/dnn.py
gvtulder: I think it's needed here because the CPU Pool op supports an arbitrary number of non-pooling dimensions, whereas the GPU ops require exactly two dimensions. Does that make sense or did I misunderstand your comment?
This should also be replaced with a Reshape to two dimensions, or removed altogether if the Pool op only supports 2 non-pooling dimensions.

theano/sandbox/cuda/tests/test_dnn.py
gvtulder: I've added it back, including the comment you proposed. I originally removed this custom mode because it wasn't in the `test_pooling` function for 2D.

theano/tensor/signal/pool.py
nouiz: ok. Thanks for thinking about the test time. It takes already too much time.

theano/tensor/signal/pool.py
nouiz: @lamblin @abergeron, what do you think? Should we take this opportunity to change the default of pool_2d now or we need to keep the warning for longer?

nouiz:
1) I'm ok to keep it if you add tests and move the reshape to the opt. Don't forget that this will be needed for the 2 GPU back-end.
2) It is sad that we use different naming. Can you open an issue with links to the places you saw the different version? Before deciding of the name to use here, if we can find a global naming scheme we want to use, we could reuse it.
The padding is always added on both size. So for the depth dimensions, I like your name front and back. I make clear it is in both side. 

I didn't do more review, just read your comments.

theano/gpuarray/dnn.py
nouiz: Your explanation is right. I made this comment before I understand that you extended what the op does to support more dimensions.

gvtulder:
The reshape is now moved to the opts, for both gpuarray and sandbox.cuda. There are tests that should check if the reshape works. The CPU pool ops are now also tested with arbitrary numbers of pooling and non-pooling dimensions.

nouiz:
Great. I merged the other PR, so this PR will need to be rebased/adjusted.

I didn't review the new code, but if you can just reuse the CPU tests that run on multiple dimensions on the GPU, it would be the missing test.

gvtulder:
I've rebased the branch to include the variable inputs, taking into account your comments. (Except for the pool_3d default value for ignore_border, which is still the same as the existing pool_2d function.)

theano/sandbox/cuda/opt_util.py
```diff
@@ -126,3 +126,48 @@ def opt(node):
...
+def pad_dims(input, leftdims, rightdims):
+    """Reshapes the input to a (leftdims + rightdims) tensor
+
+    This helper function is used to convert pooling inputs with arbitrary
+    non-pooling dimensions to the correct number of dimensions for the
+    GPU pooling ops.
+
+    This reduces or expands the number of dimensions of the input to
+    exactly `leftdims`, by adding extra dimensions on the left or by
+    combining some existing dimensions on the left of the input.
```
nouiz: I don't understand what this does from this descriptions. I know what it does because I know the code. Can you add an example just to show what it does?

theano/sandbox/cuda/opt_util.py
```diff
@@ -126,3 +126,48 @@ def opt(node):
...
+
+    if input.ndim == (leftdims + rightdims):
+        return input
+
+    # extract image dimensions
+    img_shape = input.shape[-rightdims:]
+
+    # count the number of "leading" dimensions, store as dmatrix
+    batch_size = tensor.prod(input.shape[:-rightdims])
+    batch_size = tensor.shape_padright(batch_size, 1)
```
nouiz: why this line? I don't understand why it is needed and I think it isn't what you want.

theano/sandbox/cuda/opt_util.py
```diff
@@ -126,3 +126,48 @@ def opt(node):
...
+
+    # store in the required shape, for example as a 4D tensor
+    # with shape: (batch_size,1,height,width)
+    new_shape = tensor.cast(tensor.join(0, batch_size,
+                                        tensor.as_tensor([1] * (leftdims - 1)),
```
nouiz: This isn't the ideal way to do it. Reshape will copy sometimes. This happen when the inputs isn't contiguous in some dimensions and the new shapes don't allow to represent that.

Totally optional, merge the less number of dimensions. It could sometimes (but very infrequently) save a copy of the data on the GPU.

theano/gpuarray/dnn.py
```diff
@@ -1868,9 +1872,18 @@ def local_gpua_pool_dnn_alternative(op, ctx_name, inputs, outputs):
...
         return
     img, ws, stride, pad = inputs
-    img = as_gpuarray_variable(img, ctx_name)
+    nd = op.ndim if op.ndim else (img.ndim - 2)
+    if nd not in (2, 3):
+        return
+    img = gpu_contiguous(as_gpuarray_variable(img, ctx_name))
     mode = op.mode
-    return dnn_pool(gpu_contiguous(img), ws, stride=stride, pad=pad, mode=mode)
+    if img.ndim == nd + 2:
```
nouiz: I think dnn_pool support 3d pooling. Is that the case? If so, then you should take that branch also if img.ndim == nd+3.

theano/gpuarray/dnn.py
```diff
@@ -1868,9 +1872,18 @@ def local_gpua_pool_dnn_alternative(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     img, ws, stride, pad = inputs
-    img = as_gpuarray_variable(img, ctx_name)
+    nd = op.ndim if op.ndim else (img.ndim - 2)
+    if nd not in (2, 3):
+        return
```
nouiz: I think this is a left over from your old version that didn't suported multiple ndim. I think you should remove this if.

theano/gpuarray/dnn.py
```diff
@@ -1882,17 +1895,33 @@ def local_gpua_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     inp, out, out_grad, ws, stride, pad = inputs
-    inp = as_gpuarray_variable(inp, ctx_name)
-    out = as_gpuarray_variable(out, ctx_name)
-    out_grad = as_gpuarray_variable(out_grad, ctx_name)
+    nd = op.ndim if op.ndim else (inp.ndim - 2)
+    if nd not in (2, 3):
+        return
```
nouiz: idem, I think this if isn't needed anymore.

theano/gpuarray/dnn.py
```diff
@@ -1882,17 +1895,33 @@ def local_gpua_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
...
-                                     pad)
+    if inp.ndim == nd + 2:
+        return GpuDnnPoolGrad(mode=mode)(inp,
+                                         out,
+                                         out_grad,
+                                         ws,
+                                         stride,
+                                         pad)
+    else:
+        # reshape to 4D or 5D with 2 non-pooling dimensions
```
nouiz: this currently always pad to 4d.

theano/gpuarray/dnn.py
```diff
@@ -1904,16 +1933,28 @@ def local_gpua_avg_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
     if not op.ignore_border:
         return
     inp, out_grad, ws, stride, pad = inputs
-    inp = as_gpuarray_variable(inp, ctx_name)
-    out_grad = as_gpuarray_variable(out_grad, ctx_name)
+    nd = op.ndim if op.ndim else (inp.ndim - 2)
+    if nd not in (2, 3):
+        return
```

theano/gpuarray/dnn.py
```diff
@@ -1904,16 +1933,28 @@ def local_gpua_avg_pool_dnn_grad_stride(op, ctx_name, inputs, outputs):
...
+    out_grad = gpu_contiguous(as_gpuarray_variable(out_grad, ctx_name))
     mode = op.mode
 
-    cg = gpu_contiguous(out_grad)
-
-    # We reuse cg because cuDNN does not use the value of the `out`
-    # argument but still checks its shape for average pooling. This
-    # has been observed in v2 and v3 as far as I know.
-    return GpuDnnPoolGrad(mode=mode)(gpu_contiguous(inp), cg, cg, ws, stride, pad)
+    if inp.ndim == nd + 2:
```

theano/gpuarray/opt_util.py
```diff
@@ -329,3 +329,48 @@ def opt(node):
...
+
+    if input.ndim == (leftdims + rightdims):
+        return input
+
+    # extract image dimensions
+    img_shape = input.shape[-rightdims:]
+
+    # count the number of "leading" dimensions, store as dmatrix
+    batch_size = tensor.prod(input.shape[:-rightdims])
+    batch_size = tensor.shape_padright(batch_size, 1)
```
nouiz: I don't think this line is good.

theano/gpuarray/tests/config.py
```diff
@@ -17,7 +17,7 @@
 
 if theano.config.mode == 'FAST_COMPILE':
     mode_with_gpu = theano.compile.mode.get_mode('FAST_RUN').including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray')
+    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray', 'gpu')
 else:
     mode_with_gpu = theano.compile.mode.get_default_mode().including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_default_mode().excluding('gpuarray')
+    mode_without_gpu = theano.compile.mode.get_default_mode().excluding('gpuarray', 'gpu')
```
nouiz: I think those 2 changes aren't good. Why did you need them? This would enable the old back-end at the same time as the new back-end.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -166,6 +167,27 @@ def pool_2d_i2n(input, ds=(2, 2), strides=None,
     return pooled_output
 
 
+def pool3d2d(input, ds=(2, 2, 2), strides=None, pad=(0, 0, 0),
+             pool_function=T.max, mode='ignore_borders'):
+    if strides is None:
+        strides = ds
+
+    shape = input.shape
+
```
nouiz: assert input.ndim == 5

theano/gpuarray/tests/test_dnn.py
```diff
@@ -278,6 +300,119 @@ def fn(x):
...
+        modes = ('max', 'average_inc_pad', 'average_exc_pad')
+
+    x = T.ftensor5()
+    for mode, pad in product(modes,
+                             ((0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1),
+                              (3, 2, 2), (2, 3, 2), (2, 2, 3))):
+        if mode == 'max':
+            func = T.max
+        else:
+            func = T.mean
```
nouiz: This don't allow to test the difference between both average.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -278,6 +300,119 @@ def fn(x):
...
+            if pad[0] > stride or pad[1] > stride or pad[2] > stride:
+                # Not implemented
+                continue
+
+            # This test the CPU grad + opt + GPU implemtentation
+            def fn(x):
+                return pool_3d(x, (ws, ws, ws), ignore_border=True,
+                               padding=pad, mode=mode)
+            utt.verify_grad(fn, [data],
+                            cast_to_output_type=False,
```
nouiz: this is the default, why did you added it? Can you just remove it to make the code more clear that it don't use rare options?

theano/gpuarray/tests/test_dnn.py
```diff
@@ -278,6 +300,119 @@ def fn(x):
...
+                            mode=mode_with_gpu)
+            # Confirm that we get the good op.
+            fg = theano.function([x], theano.grad(fn(x).sum(), x),
+                                 mode=mode_with_gpu)
+            assert any([isinstance(node.op, dnn.GpuDnnPoolGrad)
+                        for node in fg.maker.fgraph.toposort()])
+            g_out = fg(data)
+
+            # Compare against the CPU result
```
nouiz: This isn't needed, the verify_grad would have caught such problems. Just remove the section after this line of the test.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -362,11 +499,86 @@ def test_pooling_opt():
...
+    # 3D pooling
+    x = T.ftensor3()
+
+    f = theano.function(
+        [x],
+        pool_3d(x, ds=(2, 2, 2), mode='average_inc_pad',
+                ignore_border=True),
+        mode=mode_with_gpu)
 
     assert any([isinstance(n.op, dnn.GpuDnnPool)
```
nouiz: keep this assert i the original test case. It got moved to the new test instead of being copied.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -362,11 +499,86 @@ def test_pooling_opt():
...
+    if dnn.version(raises=False) < 4004:
+        modes = ('max', 'average_inc_pad')
+    else:
+        modes = ('max', 'average_inc_pad', 'average_exc_pad')
+
+    for n_non_pool_dims in (0, 1, 2, 3):
+        for ws in ((2, 2), (3, 3, 3)):
+            # create input shape: non-pooling dimensions
+            # followed by 2 or 3 pooling dimensions
+            shp = (2,) * n_non_pool_dims + (5,) * len(ws)
```
nouiz: To better test it, can you take different value for each dimensions? Like: xrange(2, n_non_pool_dims+2)?

theano/sandbox/cuda/dnn.py
```diff
@@ -2972,10 +2976,21 @@ def local_pool_dnn_alternative(node):
             if not node.op.ignore_border:
                 return
             img, ws, stride, pad = node.inputs
+            nd = node.op.ndim if node.op.ndim else (img.ndim - 2)
             mode = node.op.mode
+            if nd not in (2, 3):
+                return
```
nouiz: idem as in the new back-end.

theano/sandbox/cuda/dnn.py
```diff
@@ -2972,10 +2976,21 @@ def local_pool_dnn_alternative(node):
...
             if (img.owner and isinstance(img.owner.op, HostFromGpu)):
-                ret = dnn_pool(gpu_contiguous(img.owner.inputs[0]),
-                               ws, stride=stride, pad=pad, mode=mode)
+                if img.ndim == nd + 2:
```
nouiz: idem

theano/tensor/signal/pool.py
```diff
@@ -84,156 +85,181 @@ def pool_2d(input, ds, ignore_border=None, st=None, padding=(0, 0),
...
+    ndim : int
+        The number of pooling dimensions N.
+        If this number is not specified, the default is set to the
+        (input.ndim - 2), assuming that the first two dimensions of the input
+        are non-pooling dimensions.
 
     """
 
-    __props__ = ('ignore_border', 'mode')
+    __props__ = ('ndim', 'ignore_border', 'mode')
```
nouiz: Try to keep the same order as in the \__init__, so add it at the end of the list.

theano/gpuarray/dnn.py
gvtulder: No, I don't think so. Pool on the CPU supports arbitrary ndim, but the cuDNN version only supports 2D and 3D. This optimization has to check if dnn_pool can be used.

theano/gpuarray/dnn.py
gvtulder: I'll add a comment in the code to clarify this. This checks if the number of non-pooling dimensions is exactly 2. (nd is the number of pooling dimensions, which is 2 or 3, so img should be 4D or 5D.)

theano/gpuarray/dnn.py
gvtulder: See above. :) The optimization should only swap in the GPU op if we do 2D or 3D pooling.

theano/gpuarray/dnn.py
gvtulder: No. nd is the number of pooling dimensions, so it could be 4D or 5D.
I'm beginning to think that nd is a confusing name for this variable. Would it be better to make it longer and more explicit?

theano/gpuarray/opt_util.py
gvtulder: These lines and the tensor.cast on the next lines come from the original [pool_2d (theano/tensor/signal/pool.py#L95)](https://github.com/Theano/Theano/blob/166cac0/theano/tensor/signal/pool.py#L95-L97).

It seems correct to me (and it doesn't work without this line). The purpose seems to be to compress all leading non-pooling dimensions into the first dimension and setting the second non-pooling dimension to 1.
1. Line 354 computes the product of the length all of the leading non-pooling dimensions.
2. Line 355 converts the scalar batch_size to a vector, which is required by `tensor.join`.
3. Line 359 uses tensor.join to concatenate the vectors `[batch_size], [1], img_shape`.

theano/gpuarray/tests/config.py
gvtulder: It *disables* the old backend, I think. Maybe this is only a problem on my system (I made it a separate commit).

If I don't also exclude `'gpu'`, the tests that check for a CPU reference implementation of the pooling ops fail. For example, the CPU `Pool` gets replaced by `theano.sandbox.cuda.dnn.GpuDnnPool`. (In [test_pooling](https://github.com/Theano/Theano/blob/166cac0/theano/gpuarray/tests/test_dnn.py#L271-L276), for instance.)

theano/gpuarray/tests/test_dnn.py
gvtulder: Done.

theano/gpuarray/tests/test_dnn.py
gvtulder: Copied from test_pooling. I've removed it in both.

theano/sandbox/cuda/dnn.py
gvtulder: So I think this is correct (see the other comment).

theano/sandbox/cuda/dnn.py
gvtulder: See above.

theano/tensor/signal/pool.py
gvtulder: Done.

theano/gpuarray/tests/test_dnn.py
gvtulder: Done.

theano/gpuarray/dnn.py
gvtulder: See above.

theano/gpuarray/dnn.py
gvtulder: See above.

theano/sandbox/cuda/opt_util.py
gvtulder: See my comment in theano/gpuarray/opt_util.py. (I think it's there because tensor.join requires a vector.)

theano/gpuarray/tests/test_dnn.py
gvtulder: This is the same as in test_pooling for 2D.

theano/gpuarray/tests/test_dnn.py
gvtulder: This is the same as in test_pooling for 2D. (See my comment on the pull request.)

theano/gpuarray/tests/test_dnn.py
gvtulder: Done.

gvtulder:
@nouiz: Thanks for the review so far. I've addressed your comments and pushed an update.

You're correct that the test_pooling and test_pooling_3d tests in gpuarray/cuda are not testing the two types of averaging if there is any padding. The Python reference implementations in those tests do not support average_inc_pad/average_exc_pad with padding.

However, I think that the CPU Pool ops implement all those options *and* are tested with a Python reference. Would it be an idea to remove the separate reference implementations in the gpuarray/cuda tests and use the CPU Pool ops to compute the reference results?

theano/sandbox/cuda/opt_util.py
gvtulder: Done.

theano/sandbox/cuda/opt_util.py
gvtulder: I've now changed it so that it doesn't reshape more than necessary.

You could optimise it even further by looking at which dimensions can be safely combined (left or right), but that is probably not worth the effort.

gvtulder:
As I proposed in my previous comment, I now updated the pooling tests for the GPU implementations. They now use the CPU implementations as the reference.
I've also reduced the size of the 3D examples, because they were very large and made the tests really slow.

theano/gpuarray/opt_util.py
```diff
@@ -341,6 +341,24 @@ def pad_dims(input, leftdims, rightdims):
...
+    Given input of shape (3, 5, 7), ``pad_dims(input, 2, 2)``
+    adds a singleton dimension and reshapes to (3, 1, 5, 7).
+    Given that output from pad_dims, ``unpad_dims(output, input, 2, 2)``
+    reshapes back to (3, 5, 7).
+
+    Given input of shape (3, 5, 7, 9), ``pad_dims(input, 2, 2)``
+    does not reshape and returns output with shape (3, 5, 7, 9).
+
+    Given input of shape (3, 5, 7, 9, 11), ``pad_dims(input, 2, 2)``
+    combines the first two dimensions and reshapes to (8, 7, 9, 11).
```
gvtulder: Correct.

theano/gpuarray/opt_util.py
```diff
@@ -350,15 +368,25 @@ def pad_dims(input, leftdims, rightdims):
...
-    # store in the required shape, for example as a 4D tensor
-    # with shape: (batch_size,1,height,width)
-    new_shape = tensor.cast(tensor.join(0, batch_size,
-                                        tensor.as_tensor([1] * (leftdims - 1)),
-                                        img_shape), 'int64')
+    non_pool_ndim = input.ndim - rightdims
+    if non_pool_ndim < leftdims:
+        # too few dimensions, pad on the left
+        dummy_dims = tensor.as_tensor([1] * (leftdims - non_pool_ndim))
+        new_shape = tensor.join(0, dummy_dims,
```
gvtulder: Fixed the documentation.

gvtulder:
doc/library/tensor/signal/pool.txt now also lists pool_3d.

nouiz:
jenkins test this

nouiz:
jenkins have an error: http://darjeeling.iro.umontreal.ca:8080/job/Theano%20gpu/314/testReport/junit/theano.sandbox.cuda.tests/test_blas/test_downsample/

gvtulder:
Apparently I didn't change these opts to work with the new variable padding argument (and apparently I also didn't run these tests).

I think it should be fixed now. Can you run Jenkins again?

nouiz:
jenkins test this

gvtulder:
Jenkins finished.

theano/gpuarray/tests/config.py
```diff
@@ -17,7 +17,7 @@
 
 if theano.config.mode == 'FAST_COMPILE':
     mode_with_gpu = theano.compile.mode.get_mode('FAST_RUN').including('gpuarray').excluding('gpu')
-    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray')
+    mode_without_gpu = theano.compile.mode.get_mode('FAST_RUN').excluding('gpuarray', 'gpu')
```
abergeron: Why did you need that change?

danielS91:
@nouiz is there an eta when this gets merged? Would be really nice to have it in theano!

nouiz:
What about now?

PR #4886, opened 2016-08-22
Abstractconv test with optimizer=None

ballasn:


nouiz:
Just a note that this would have discovered in normal test the bug fixed by https://github.com/Theano/Theano/pull/4846

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -331,6 +332,47 @@ def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
...
+    @classmethod
+    def setup_class(cls):
+        if theano.config.blas.ldflags == "":
+            raise SkipTest()
+        BaseTestConv2d.setup_class()
+        cls.inputs_shapes = [(8, 1, 6, 6)]
+        cls.filters_shapes = [(5, 1, 2, 2)]
+        cls.subsamples = [(1, 1), (2, 2)]
+        cls.filters_dilations = [(1, 1), (1, 2), (2, 1)]
+        cls.border_modes = ["valid", "full"]
```
nouiz: Do we want to add half here as now it is available?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -331,6 +332,47 @@ def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
...
+        cls.filter_flip = [True]
+        cls.provide_shape = [False]
+
+    def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
+        o = self.get_output_shape(i, f, s, b, fd)
+        if (not theano.config.blas.ldflags or
+                not theano.config.cxx or
+                theano.config.mode == "FAST_COMPILE"):
```
nouiz: Why the last condition? As you create your own mode, I don't think it is needed. Otherwise my 2 small comments, I think this PR is good. thanks

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1,3 +1,4 @@
+
```
lamblin: Remove that empty line.

theano/tensor/nnet/tests/test_abstract_conv.py
lamblin: I think we should remove that "if" altogether, since the un-optimized perform of AbstractConv does not actually need blas.

nouiz:
jenkins test this

PR #4891, opened 2016-08-23
fix slowdown introduced by gh-4865

nouiz:


theano/tensor/basic.py
```diff
@@ -2312,17 +2315,22 @@ def ones_like(model, dtype=None):
     """
     if dtype is None:
         dtype = model.type.dtype
-    ret = fill(model, constant(1.0, dtype=dtype))
-    return ret
+    ret = constant(1.0, dtype=dtype)
+    if opt and ret.type == model.type:
+        return ret
+    return fill(model, ret)
```
abergeron: Why should this be conditional?

theano/tensor/opt.py
```diff
@@ -2032,15 +2032,15 @@ def local_useless_elemwise(node):
 
                 # Copy stack trace from input to constant output
                 copy_stack_trace(node.outputs[0], ret)
-                return ret
+                return [ret]
```
Sentient07: This was what caused the ValueError in the profiling i think ?

Sentient07:
I can confirm that the the error reported [here](https://github.com/Theano/Theano/pull/4876#issuecomment-241769954) is fixed in this PR,

theano/tensor/opt.py
Sentient07: The ValueError is fixed 👍

Sentient07:
I profiled at the top of this PR, this doesn't speed up :(  http://pastebin.com/UNWgTAX4

nouiz:
Finally, this PR don't fix any bug, just a slodown.

theano/tensor/basic.py
nouiz: We tried and this make many tests fail.

The cause is that when the user build the graph, they suppose model is in the graph. So it cause error like inputs to theano.function aren't used in the graph.

Sentient07:
No, it fixes the ValueError bug which i reported in the [comment](https://github.com/Theano/Theano/pull/4876#issuecomment-241740656).

theano/tensor/basic.py
```diff
@@ -2298,12 +2298,15 @@ def second(a, b):
...
 
 @constructor
-def ones_like(model, dtype=None):
+def ones_like(model, dtype=None, opt=False):
     """equivalent of numpy.ones_like
     Parameters
     ----------
     model : tensor
     dtype : data-type, optional
+    opt : If True, we will return a constant instead of a graph when possible.
```
lamblin: Currently, this only applies when model is a tensor.scalar, right? If so, we should probably say that.

lamblin:

nouiz:
I added tests.

Sentient07:
@nouiz I didn't understand the last commit message ?

nouiz:
I got mixed, the tests where already there. Forget my last comment.

theano/tensor/basic.py
```diff
@@ -2312,17 +2315,22 @@ def ones_like(model, dtype=None):
     """
     if dtype is None:
         dtype = model.type.dtype
-    ret = fill(model, constant(1.0, dtype=dtype))
-    return ret
+    ret = constant(1.0, dtype=dtype)
+    if opt and ret.type == model.type:
```
Sentient07: I had profiled with this. This causes  a slight slow down. Also, at the end of the profiling there wasn't any error when i removed this change. Do you remember which error this diff fixed ?

Sentient07: With reverting the change* 
Sorry for the incomplete sentence

Sentient07: I had tried 
1) with this diff, 
2) Returning constant default
3) Returning graph default
Returning graph is the slowest whereas the returning constant is fastest. But I didn't get any error when i tried 2)

PR #4893, opened 2016-08-23
Modified Developer Start Guide VIM instructions

shawntan:
Plugins will not load without calling `vundle#end()`


nouiz:
I don't use vim, maybe @lamblin will find time to review this. I let someone else that use vim comment/merge on this PR.

thanks

lamblin:
Indeed, it is necessary.
Thanks for the fix!

shawntan:
No problem! Was trying to get it set up myself when I ran into issues.

Issue #4892, opened 2016-08-23
Error while printing profile


Sentient07:
This is the error that I'm getting at the end of printing profile over the current master.

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 103, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1256, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 339, in print_profile
    level=level + 1)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 2446, in print_profile
    o.print_profile(stream, prof, level + 2)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 899, in print_profile
    nb_fail, nb_merged, nb_constant), file=stream)
TypeError: %d format: a number is required, not TopoOptimizer
Error in sys.exitfunc:
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 103, in _atexit_print_fn
    n_apply_to_print=config.profiling.n_apply)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/compile/profiling.py", line 1256, in summary
    self.optimizer_profile[1])
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 339, in print_profile
    level=level + 1)
  File "/Users/Ramana/projects/macvnev/lib/python2.7/site-packages/theano/gof/opt.py", line 2446, in print_profile
    o.print_profile(stream, prof, level + 2)
[...]

PR #4894, opened 2016-08-24
Fix opt profile crash

nouiz:
fix gh-4892


theano/gof/opt.py
```diff
@@ -1941,10 +1941,15 @@ def pruner(node):
 
     @staticmethod
     def print_profile(stream, prof, level=0):
+        blanc = ('    ' * level)
+        if prof is None:  # Happen as merge_profile() isn't implemented
```
Sentient07: I had done this part and was about to open a PR 😅

nouiz:
I updated this PR to lower duplicate code.

Issue #4885, opened 2016-08-22
Bug in conv2d_grad_wrt_weights() method


clarken92:
I tried to use the method conv2d_grad_wrt_weights() to compute the gradients of weights. 
The arguments I passed to this function are:

"input" is the **bottom layer** of shape (n_instances, in_channels, in_width, in_height). In my example, my input shape is **(20, 3, 5, 5)**
"output_grad" is the **upper layer** gradient of shape (n_instances, out_channels, out_width, out_height). In my example, the output shape is **(20, 6, 4, 4)** 
"filter_shape" is a 4-tuple. According to the default style in Theano, it is **[6, 3, 2, 2]**.

However, I got an error when running the method. I inspected the source code and realized that the **return statement** (line 389 of file **abstract_conv.py**) of the function seem to be wrong:

`return gradWeight_op(input, output_grad, filter_shape[:-2])`

Here, I think it should be **filter_shape[-2:]** as usual rather than **filter_shape[:-2]**.
Note that if I passed **filter_shape=[6, 3, 2, 2]**, **filter_shape[-2:] = [2, 2]** while **filter_shape[:-2] = [6, 3]**. I did a small change by passing **[2, 2, 6, 3]** to **filter_shape** and the method works fine. However, as [2, 2, 6, 3] seems to be inconsistent, I think this line of code should be fixed in the future.

lamblin:
You are correct.
Someone is working on adding tests that would hopefully detect the issue, but we should fix it.
Could you submit a pull request with that fix?
Thanks,

nouiz:
There is a PR that fix this: gh-4911

nouiz:
The fix was merged. thanks for the report.

PR #4911, opened 2016-08-29
fix conv2d_grad_wrt_weights bug

ballasn:


theano/tensor/nnet/abstract_conv.py
```diff
@@ -422,7 +422,7 @@ def conv2d_grad_wrt_weights(input,
                                                filter_flip=filter_flip,
                                                filter_dilation=filter_dilation)
 
-    return gradWeight_op(input, output_grad, filter_shape[:-2])
```
nouiz: Do this fix an existing test or it need a new test?

ballasn: It needs a new test, but the test will be introduced with #4711  pull request.

It fixes the issue #4885

nouiz: thanks.

Can you finish gh-4711? Is there something left in it?

lamblin: @shabanian is working on #4711, it should be ready soon.

Issue #4257, opened 2016-03-18
Op.grad() should take node as a parameter


nouiz:
When the grad need the output of the forward, currently, we must create a new node equivalent to the forward just to get that. This ask more work from the optimizer to merge that and make the printing or not optimized graph less readable. I'm not sure if we optimize correctly all the graph for the 2nd derivative in all case where this happen.

The current interface: `def grad(self, inp, grads)`. The ideal interface I would like: `def grad(self, node, grads)`. This would allow to get the inputs and the outputs. But this would cause errors.

Should we create a grad_node() method with the new interface? Should we create grad(inp, grads, node) and if the call raise an error, call back to the old interface? I think the first is better. Other suggestion for the name?

PR #4915, opened 2016-08-30
Cudnn RNN bindings.

abergeron:
This survived some very light testing (it can run forwards and backwards without crashing).

I did not compare this implementation against a "known-good" one yet, but I plan to do that later. 

In the meantime, if anybody has feedback on the code, especially the first commits which add new functionality, that would be great.


nouiz:
travis and jenkins fail.

nouiz:
I think the travis problem is that it use python 2.6. I'll make another PR to drop it as we discussed

nouiz:
https://github.com/Theano/Theano/pull/4924 when merged, you can rebase this PR on top of it.

nouiz:
```
/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1797:13: E126 continuation line over-indented for hanging indent

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1829:27: E261 at least two spaces before inline comment

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1830:28: E261 at least two spaces before inline comment

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1834:32: E261 at least two spaces before inline comment

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1840:18: E261 at least two spaces before inline comment

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1841:19: E261 at least two spaces before inline comment

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1845:23: E261 at least two spaces before inline comment
```

nouiz:
In Python 3.3, there is a bunch of errors like:

```

ERROR: test_in_update (theano.compile.tests.test_function.TestFunctionIn)

  File "/home/travis/build/Theano/Theano/theano/compile/tests/test_function.py", line 88, in test_in_update

    mode='FAST_RUN')

  File "/home/travis/build/Theano/Theano/theano/compile/function.py", line 326, in function

    output_keys=output_keys)

  File "/home/travis/build/Theano/Theano/theano/compile/pfunc.py", line 484, in pfunc

    output_keys=output_keys)

  File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 1789, in orig_function

    defaults)

  File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 1653, in create

  File "/home/travis/build/Theano/Theano/theano/gof/link.py", line 703, in make_thunk

    storage_map=storage_map)[:3]

  File "/home/travis/build/Theano/Theano/theano/gof/vm.py", line 1120, in make_all

    self.updated_vars,

  File "/home/travis/build/Theano/Theano/theano/gof/vm.py", line 958, in make_vm

    ivar.container)

AttributeError: 'TensorVariable' object has no attribute 'container'
```

theano/gof/op.py
```diff
@@ -694,6 +694,9 @@ def __ne__(self, other):
     # Python implementation #
     #########################
 
+    def grad2(self, inputs, outputs, output_grads):
```
nouiz: As discussed a few weeks ago. I don't like the idea to have a method named grad2.

I would prefer that at the call place we have a try... except. We try with the extra args, if we get an AttributeError, then we call the old version. This would be better from a User that make Op perspective, but it have others problems.

Do you have others solution? A completly new names? I think this method don't really implement the gradient, but something close. Could this be the occasion to also please people very good in math by reconciling this other problem and not have a function ending with 2?

nouiz: This issue won't block this PR as it isn't documented for now. But if we merge it like this, we need an issue about it and hopefully we should find a good fix not too long. I would like a fix before 0.9 to don't release new interface that we could change later.

theano/gof/type.py
```diff
@@ -603,26 +633,32 @@ class CDataType(Type):
...
-        return and take a single pointer argument.
+        A function to call to free the pointer. This function must
+        have a `void` return and take a single pointer argument.
 
     """
 
-    import ctypes
-    if PY3:
-        _cdata_type = ctypes.py_object.from_address(
-            ctypes.addressof(ctypes.pythonapi.PyCapsule_Type)).value
```
nouiz: As we are droping python 2.6, I think we can clean up the code to replace PyCObject with PyCapsule usage everywhere. Can you confirm this? I make a note about this in the issue about dropping python 2.6

theano/gof/type.py
```diff
@@ -603,26 +633,32 @@ class CDataType(Type):
...
-    else:
-        _cdata_type = ctypes.py_object.from_address(
-            ctypes.addressof(ctypes.pythonapi.PyCObject_Type)).value
-    del ctypes
-
-    def __init__(self, ctype, freefunc=None):
+    def __init__(self, ctype, freefunc=None, headers=None, header_dirs=None,
```
nouiz: This probably need updates to the doc.

nouiz: I think we should use the same name as compile_str and same order to be consistent:

    def compile_str(module_name, src_code, location=None,
                    include_dirs=None, lib_dirs=None, libs=None,
                    preargs=None, py_module=True, hide_symbols=True):

theano/gof/type.py
```diff
@@ -633,10 +669,22 @@ def __hash__(self):
...
     def filter(self, data, strict=False, allow_downcast=None):
-        if data is not None and not isinstance(data, self._cdata_type):
-            raise TypeError("expected None or PyCObject/PyCapsule")
+        if data is not None and not isinstance(data, _cdata_type):
+            raise TypeError("expected None or a PyCapsule")
         return data
 
+    def _get_func(self):
+        from theano.scalar import get_scalar_type
+        import theano
```
nouiz: move at least the "import theano" at the top. I'm not sure if the other can be, but if possible do it.

theano/gof/type.py
```diff
@@ -633,10 +669,22 @@ def __hash__(self):
...
+
+        if self._fn is None:
+            v = get_scalar_type('int64')()
+            self._fn = theano.function([v], _make_cdata(self)(v))
```
nouiz: add the parameter profile=False to remove it from the user profile.

theano/gof/op.py
abergeron: I'm all for a new, better name.  I just don't have any good ideas.

theano/gof/type.py
abergeron: This is what this chunk is doing (in part).

theano/gof/type.py
abergeron: I use the same names as the methods you have to define on a type (minus the `c_` prefix).

theano/gof/type.py
abergeron: I really wanted to avoid circular imports and this is the definition of Type so I don't think we can move the other one, but for "import theano" sure.

theano/gof/type.py
abergeron: Also, yes, the docs could be improved to reflect the new arguments, I'll do that after I get the tests working.

theano/gof/type.py
abergeron: ok

abergeron:
I dropped the commit about pinned shared variables because it turns out that the final code doesn't need it and I don't want to debug it (It's the source of the "container" problem).

Apart from that, I've made flake8 fixes and addressed some small comments.

theano/gof/type.py
nouiz: ok. Sad that we use different name for different stuff.

theano/gof/op.py
nouiz: I asked @slefrancois if he have ideas.

theano/gof/type.py
```diff
@@ -710,8 +742,20 @@ def c_cleanup(self, name, sub):
...
+
+    def c_libraries(self):
+        return self.libraries
+
+    def c_lib_dirs(self):
+        return self.lib_dirs
+
     def c_code_cache_version(self):
-        return (2, self.ctype, self.freefunc)
+        return (3,)
```
nouiz: Why did you remove the other attribute and not add the new ones instead?

theano/gpuarray/type.py
```diff
@@ -20,6 +20,7 @@
     pygpu = None
 
 _context_reg = {}
+_props_map = {}
```
nouiz: add in a comment an example of props. props is an overloaded name with many meaning.

Or better, add docstring to the new user methods from this file.

theano/gpuarray/dnn.py
```diff
@@ -39,6 +40,44 @@
...
+def _dnn_lib():
+    if _dnn_lib.handle is None:
+        import ctypes.util
+
+        lib_name = ctypes.util.find_library('cudnn')
+        if lib_name is None and sys.platform == 'win32':
+            # Update these names when new versions of cudnn are supported.
+            lib_name = ctypes.util.find_library('cudnn64_5.dll')
+            if lib_name is None:
+                lib_name = ctypes.util.find_library('cudnn64_4.dll')
```
nouiz: Can you make a list and iterate on it? It will make adding new version much easier and it would remove duplicate code (1 line for now, but more later)

theano/gpuarray/dnn.py
```diff
@@ -39,6 +40,44 @@
...
+        if lib_name is None:
+            raise RuntimeError('Could not find cudnn library')
```
nouiz: We probably need a better error message, in case people use never cudnn version on windows.

theano/gpuarray/__init__.py
```diff
@@ -87,19 +87,24 @@ def init_dev(dev, name=None):
...
+                    warnings.warn("Your cuDNN version is more recent than "
+                                  "Theano. If you encounter problems, try "
+                                  "updating Theano or downgrading cuDNN to "
+                                  "version 5.1.")
+                if config.print_active_device:
+                    print("Using cuDNN version %d on context %s" %
+                          (ctx_props['cudnn_version'], name), file=sys.stderr)
+                ctx_props['cudnn_handle'] = dnn._make_handle(context)
+            except Exception:
+                pass
```
nouiz: This need at least a warning that cudnn isn't found and won't be used/available.

theano/gpuarray/dnn.py
```diff
@@ -153,20 +198,31 @@ class DnnBase(COp):
...
+    def dnn_context(self, node):
+        return node.outputs[0].type.context_name
 
     def get_params(self, node):
-        return node.outputs[0].type.context
+        try:
+            return get_prop(self.dnn_context(node), 'cudnn_handle_param')
+        except KeyError:
+            pass
+        ptr = get_prop(self.dnn_context(node), 'cudnn_handle').value
```
nouiz: Will we have raised a better error before if cudnn isn't available? If not sure, just add a try except and tell that cudnn is needed.

theano/gpuarray/dnn_base.c
```diff
@@ -103,44 +103,7 @@ c_set_filter(PyGpuArrayObject *var, cudnnFilterDescriptor_t desc) {
...
-  cuda_enter(PARAMS->ctx);
-  cudnnStatus_t err;
-  APPLY_SPECIFIC(_handle) = NULL;
-  if ((err = cudnnCreate(&APPLY_SPECIFIC(_handle))) != CUDNN_STATUS_SUCCESS) {
-    PyErr_Format(PyExc_RuntimeError, "could not create cuDNN handle: %s",
-                 cudnnGetErrorString(err));
-    cuda_exit(PARAMS->ctx);
-    FAIL;
-  }
-  if ((err = cudnnSetStream(APPLY_SPECIFIC(_handle),
```
nouiz: I don't see any new code that would call setstream. So it will probably end up in the default stream instead of the stream we want.

theano/gpuarray/dnn_base.c
```diff
@@ -103,44 +103,7 @@ c_set_filter(PyGpuArrayObject *var, cudnnFilterDescriptor_t desc) {
...
-    cuda_exit(PARAMS->ctx);
-    FAIL;
-  }
-  cuda_exit(PARAMS->ctx);
-}
-
-#section cleanup_code_struct
-
-cuda_enter(ctx->ctx);
-cudnnDestroy(APPLY_SPECIFIC(_handle));
```
nouiz: It mean anything we put in the props like the cudnnhandle is never destroyed. I don't see problems with that. Just wanted to make sure you saw this and also don't expect problems with this.

theano/gof/op.py
lamblin: Would it be possible to either:
- use a metaclass in Op so that would automatically redefine the `grad` method so it accepts (and ignore) `outputs`, or
- use `inspect` in `theano.gradient` to see what it accepts and pass that (rather than try/except)?

theano/gpuarray/dnn_base.c
```diff
@@ -88,6 +92,13 @@ c_set_filter(PyGpuArrayObject *var, cudnnFilterDescriptor_t desc) {
     dims[i] = PyGpuArray_DIM(var, i);
   }
 
+  /* Filters can't be less than 3d so we pad */
```
nouiz: Can you add in this comment and the previous like this which current version of cudnn have this limit. In case it change in the futur, it is useful.

theano/gpuarray/dnn_base.c
```diff
@@ -114,6 +126,17 @@ c_set_filter(PyGpuArrayObject *var, cudnnFilterDescriptor_t desc) {
   return 0;
 }
 
+static int c_make_filter(PyGpuArrayObject *var, cudnnFilterDescriptor_t *desc) {
```
nouiz: This can probably be used to clean up code. If that is the case, it would be great to include it. Quick to do and make a big difference in readability.

theano/gof/op.py
lamblin: Otherwise, we could use `Lop` for the name. This is less easily understandable, but technically more correct than "grad".

nouiz:
I think you removed the commit that add support for pinned shared var. Is this wanted?

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+        return False
+
+    def make_node(self, dropout, seed, context_name):
+        dropout = as_scalar(dropout).astype('float32')
+        seed = as_scalar(seed).astype('uint64')
+
+        assert context_name == self.context_name
+        # This is a dirty hack to pass the context because params is
+        # occupied by the cudnn handle
+        context = gpu_context_type.make_constant(get_context(context_name))
```
nouiz: instead of this ugly hack. Can you make a GpuArrayConstant() on the right context and pass this as the 3rd input? It would be less hacky and would like like what the other cudnn op that need the context do except that this input is only used for this. It will then allow to cache this c code. I really hate not being able to cache the c code...

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+                                   context_name=context_name)()])
+
+    def c_code_cache_version_apply(self, node):
+        # disable the cache since we can't pickle contexts
+        return None
+
+
+def _make_dropout_desc(dropout, seed, context_name):
+    desc, states = theano.function([], _DropoutDescriptor(context_name)(
+        dropout, seed, context_name))()
```
nouiz: Cache the function building and pass the parameter profile=True. Should we also pass the mode Mode(optimizer=None)?

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+
+        return Apply(self, [hidden_size, num_layers,
+                            dropoutdesc_type.make_constant(ddesc),
+                            input_mode, direction_mode, rnn_mode, dtype],
+                     [rnndesc_type()])
+
+
+def _make_rnn_desc(hidden_size, num_layers, ddesc, rnn_mode,
+                   input_mode, direction_mode, dtype, context_name):
+    desc = theano.function([], _RNNDescriptor(context_name)(
```
nouiz: idem. cache the function and pass profile=False. Maybe Mode(optimizer=None)

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+
+    def __init__(self, context_name):
+        DnnBase.__init__(self, ["dnn_rnn_paramsize.c"],
+                         "dnn_rnn_paramsize")
+        self.context_name = context_name
+
+    def dnn_context(self, node):
+        return self.context_name
+
+    def do_constant_folding(self, node):
```
nouiz: Should we make the default do_constant_folding be False if any of the inputs or outputs are CDataType?

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+        typecode = as_i32(typecode)
+        return Apply(self, [rnndesc_type.make_constant(desc), input_size,
+                            typecode],
+                     [get_scalar_type('uint64')()])
+
+
+def _get_param_size(desc, input_size, dtype, context_name):
+    typecode = gpuarray.dtype_to_typecode(dtype)
+    return theano.function([], _RNNParamSize(context_name)(
+            desc, input_size, typecode))()
```
nouiz: idem, cache, profile=False, maybe Mode(optimizer=None)

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+
+
+class GpuDnnRNNOp(DnnBase):
+    __props__ = ()
+    _cop_numi = 5
+    _cop_numo = 4
```
nouiz: Can you confirm that this is only needed as this op have a variable number of inputs and that the COp don't support that?

I don't like the name, _cop_num_inputs would be better. Why not add an extra Constant input that is NULL to the op instead of this hack? Can an CDataType instance be NULL?

It seem to work, but I don't like it. I don't have other suggestion through...

theano/gpuarray/dnn.py
```diff
@@ -1630,6 +1632,325 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+            raise ValueError('direction_mode')
+
+    def dnn_context(self, node):
+        return node.outputs[1].type.context_name
+
+    def make_node(self, desc, w, x, hx, cx=None):
+        if cx is None:
+            context_name = infer_context_name(w, x, hx, cx)
+        else:
+            context_name = infer_context_name(w, x, hx)
```
nouiz: this if seem strange. if cx is None, I would not pass it to infer_context_name. In fact, I would inverse the content of both condition. If I'm wrong, it would need a comment.

theano/gof/type.py
abergeron: Because those should not be part of the cache key.  They are attributes of the type and are checked in \__eq__ and \__hash__.

theano/gpuarray/dnn.py
abergeron: We only ever support one version back so the list shouldn't grow in size.

theano/gpuarray/__init__.py
abergeron: There wasn't any warnings in that case before and I don't see why we should start now.

theano/gpuarray/dnn_base.c
abergeron: It will end up in the NULL stream which is ok now that libgpuarray is ready for that.

abergeron:
Yes I removed the code for pinned variables since I don't need it in the end and I don't have time to debug it right now.  I've kept the commit in one of my branches somewhere for later.

theano/gpuarray/dnn.py
abergeron: That would be even more of a hack.  The proper solution is to implement a struct-like type and use it to pass multiple values in params.

theano/gpuarray/dnn.py
abergeron: No you're right and this is reversed.

theano/gpuarray/type.py
abergeron: ok

theano/gpuarray/dnn.py
abergeron: Maybe.  Not if you attempt to use cudnn ops directly, bypassing the optimizers.

theano/gpuarray/dnn_base.c
abergeron: It is destroyed at python shutdown.  And it not so much a big deal since we don't create a handle per op anymore.

theano/gpuarray/dnn_base.c
abergeron: I have no idea when this limit started.

theano/gpuarray/dnn.py
abergeron: I've added support for optional inputs/outputs for COp and this is what you need to add.  I'll rename the values.

theano/gpuarray/dnn.py
abergeron: Maybe.

theano/gpuarray/dnn.py
abergeron: also idem.

theano/gpuarray/dnn.py
abergeron: I can't cache the function since it would cache the values for dropout and seed.

Also since this is a 2-node function I don't think we will have any speed problems with it.

abergeron:
The only thing broken with the tests now should be that python 2.6 doesn't work anymore.

theano/gof/type.py
nouiz: right! It is not mandatory, but I think the new parameter should be added in \__eq__ and \__hash__. Shoud we use \__props__ for this class instead?

theano/gpuarray/__init__.py
nouiz: In the old back-end, we tell it when cudnn isn't there. There is big consequence of this. We want to lower the number of user questions on the mailing list. It will help them find the problem if we print it. Otherwise, the problems/errors won't always be super clear.

theano/gpuarray/dnn_base.c
nouiz: I mean, add that at current cudnn v5, there is this limit. So if we end up in the futur where the current and the previous don't have this limit and we still see this code, we now easily we can remove it. Also, if we drop cudnn v5, it make the clean up faster and make sure we won't forget to check that.

theano/gpuarray/dnn.py
nouiz: if you can't cache, do the others.

theano/gpuarray/dnn.py
nouiz: I think we should not merge this until we enable the caching of this. We already compile too much. Using a constant isn't very hacky and cause less problems. You use the inputs, not the value, but its context.

I agree the full solution is better, but I think it would be another PR in itself. So I would go with the constant inputs here.

theano/gof/op.py
abergeron: We already have an `Lop`  (which is the mirror of `Rop`).

abergeron: Also, it would be possible to do a number of yucky things to accept both signatures, (the simplest being a try..except in the theano.gradient method, like what was done to add the c_compiler argument), but I don't like complexifiying the call protocol like that.  It's better to have a single name with a single signature.

In the end if we really want to keep the grad() name, we could add a method called _grad() that would try both signatures on public method and returns what works.

theano/gof/type.py
abergeron: I've added an implementation of __props__ for all theano classes (at the object2 level) and used it here with the missing items.

theano/gof/op.py
slefrancois: As for a name for grad2, I don't have a rigorous mathematical name. I can suggest "grad_chain", since it's essentially the gradient of the op times the previous outputs from the chain rule.

nouiz:
This PR fix gh-4257

abergeron:
This will need https://github.com/Theano/Theano/pull/4951 and https://github.com/Theano/Theano/pull/4968 to be merged first and then a rebase.

abergeron:
I'll fix conflicts once the tests pass locally.

For now I have the forward mode passing the tests as well as the gradient with regards to inputs.

As for the gradient on weights, it returning super high values (comparing to the reference) and I'll need to find out why.

abergeron:
I finally fixed all the problems and rebased, so now this should be ready to merge.

nouiz:
some travis error:

```
/home/travis/build/Theano/Theano/theano/configparser.py:92:1: E302 expected 2 blank lines, found 1

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:1938:9: E301 expected 1 blank line, found 0

/home/travis/build/Theano/Theano/theano/gpuarray/dnn.py:2003:36: E226 missing whitespace around arithmetic operator

/home/travis/build/Theano/Theano/theano/gpuarray/tests/rnn_support.py:1:1: FI10 __future__ import "division" missing
```

nouiz:
More travis error here: https://travis-ci.org/Theano/Theano/jobs/161476634

theano/gpuarray/dnn.py
```diff
@@ -1582,6 +1640,559 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+    __props__ = ('rnn_mode',)
+
+    def __init__(self, rnn_mode):
+        DnnBase.__init__(self)
+        self.rnn_mode = rnn_mode
+
+    def make_node(self, w, desc, layer, isize, typecode):
+        w = as_gpuarray_variable(w, infer_context_name(w))
+        layer = as_scalar(layer).astype('int32')
+        isize = as_tensor_variable(isize).astype('uint64')
```
nouiz: assert ndim on isize and w.

theano/gpuarray/dnn.py
```diff
@@ -1582,6 +1640,559 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+
+    def __init__(self, rnn_mode, direction_mode):
+        DnnBase.__init__(self, ["dnn_rnn_fwd.c"], 'dnn_rnn_fwd')
+        self.rnn_mode = rnn_mode
+        if direction_mode == 'bidirectional':
+            self.num_dirs = 2
+        elif direction_mode == 'unidirectional':
+            self.num_dirs = 1
+        else:
+            raise ValueError('direction_mode')
```
nouiz: Tell that the value is invalid and show what was given.

theano/gpuarray/dnn.py
nouiz: Still not fixed.

theano/gpuarray/dnn.py
```diff
@@ -1582,6 +1640,559 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+        assert x.ndim == 3  # seqLength, minibatch, inputSize
+        assert hx.ndim == 3  # numLayers * bidi, minibatch, hiddenSize
+
+        if self.rnn_mode == 'lstm':
+            cx = as_gpuarray_variable(cx, context_name)
+            assert cx.ndim == 3  # numLayers * bidi, minibatch, hiddenSize
+            inputs.append(cx)
+
+        _3d = GpuArrayType(dtype=x.dtype, broadcastable=(False, False, False),
+                           context_name=context_name)
```
nouiz: I think we should do at some point as in the old back-end and the CPU, define fmatrix, dmatrix, matrix (that take a dtype param), instead of making new type everytimes. Can be later.

theano/gpuarray/dnn.py
```diff
@@ -1582,6 +1640,559 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+        return Apply(self, inputs, outputs)
+
+    def grad2(self, inputs, outputs, output_grads):
+        desc, w, x, hx = inputs[:4]
+        cx = inputs[4] if len(inputs) == 5 else None
+        reserve, y, hy = outputs[:3]
+        _, dy, dhy = output_grads[:3]
+        dcy = output_grads[3] if len(output_grads) == 4 else None
+        # If both dy and dhy are disconnected, then this will error
+        # out, but it is indeed an error.
```
nouiz: Can you explain why this is happening? This isn't something usual.

theano/gpuarray/dnn.py
```diff
@@ -1582,6 +1640,559 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+        outputs = [w.type()]
+        return Apply(self, inputs, outputs)
+
+
+class RNNBlock(object):
+    def __init__(self, dtype, hidden_size, num_layers, rnn_mode,
+                 input_mode='linear', direction_mode='unidirectional',
+                 dropout=0.0, dropout_seed=4242, context_name=None):
+        """
+        dtype: data type of computation
```
nouiz: This seem to be the user interface. Add some text.

Do we want to start to add that in Theano? Should this be moved to the test? Should we compare what framework interface exists to try to make them match?

@lamblin what do you think?

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1095,3 +1097,78 @@ def test_batchnorm_inference():
...
+    depth = 3
+    timesteps = 5
+
+    # test code
+    X = T.tensor3('X')
+    X.tag.test_value = numpy.zeros((timesteps, batch_size, input_dim), dtype=theano.config.floatX)
+    Y = T.tensor3('Y')
+    Y.tag.test_value = numpy.zeros((timesteps, batch_size, hidden_dim), dtype=theano.config.floatX)
+    h0 = T.tensor3('h0')
+    h0.tag.test_value = numpy.zeros((depth, batch_size, hidden_dim), dtype=theano.config.floatX)
```
nouiz: remove test_value. I don't think they help the test now.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1095,3 +1097,78 @@ def test_batchnorm_inference():
...
+    rnnb = dnn.RNNBlock(theano.config.floatX, hidden_dim, depth, 'gru')
+    psize = rnnb.get_param_size([batch_size, input_dim])
+    params_cudnn = gpuarray_shared_constructor(
+        numpy.zeros((psize,), dtype=theano.config.floatX))
+
+    model = Model()
+    last_layer = WrapperLayer(X.dimshuffle(1, 0, 2))
+    last_dim = input_dim
+    for i in range(depth):
+        gru = GRU(last_dim, hidden_dim, last_layer, s0=h0[i, :, :])
```
nouiz: It would be great to extend the tests to more parameter, event just to make sure they compile.

Not sure how priority it is.

theano/gpuarray/type.py
nouiz: still needed to do

theano/configparser.py
```diff
@@ -89,38 +89,43 @@ def config_files_from_theanorc():
...
+                return f(*args, **kwargs)
+
+    def __enter__(self):
+        self.old_vals = {}
+        for k, v in iteritems(self.confs):
+            self.old_vals[k] = v.__get__(True, None)
+        try:
+            for k, v in iteritems(self.confs):
+                v.__set__(None, self.new_vals[k])
+        except:
```
abergeron: We want to run during everything.  It shouldn't have an effect outside of the `with` statement.

theano/gof/op.py
abergeron: I don't mind renaming, I just want a consensus.

theano/gof/type.py
```diff
@@ -678,8 +673,10 @@ def _get_func(self):
         from theano.scalar import get_scalar_type
```
abergeron: ok

theano/gpuarray/dnn.py
```diff
@@ -1582,6 +1640,559 @@ def make_node(self, x, dy, scale, x_mean, x_invstd):
...
+    def __init__(self, context_name):
+        DnnBase.__init__(self, ["dnn_rnn_paramsize.c"],
+                         "dnn_rnn_paramsize")
+        self.context_name = context_name
+
+    def dnn_context(self, node):
+        return self.context_name
+
+    def do_constant_folding(self, node):
+        return False
```
abergeron: Otherwise it tries to do constant folding and fails.

theano/configparser.py
lamblin: I agree with @abergeron on that one.

theano/gof/op.py
lamblin: Since `Op.Lop` does not actually exist, and that `theano.gradient.Lop` actually calls `theano.grad`, I would favor that one.
But I have no real objection against `grad_chain`, except it is less descriptive.

theano/gpuarray/dnn.py
abergeron: Yes, indeed.

theano/gpuarray/dnn.py
lamblin: I'm not sure what the issue is. Is it the class a typical user would call?

abergeron: Yes it is the user interface.  We don't really have a choice to add something like this here since we need a central object to hold the related parameters during graph creation.

I consider this a one-off and not the start of a pattern.  In that sense I would not try to make it match anything nor would I try to ensure future endeavours will match this.  If we ever get a cudnnGetRNNDescriptor() then we could probably do away with this and should.

theano/gpuarray/dnn.py
lamblin: Agreed.
Actually, I would make a case for the default `matrix()` to instantiate a `GpuArrayType` if `device` is not `cpu`.
But that is a different issue.

theano/gpuarray/dnn.py
```diff
@@ -55,11 +55,12 @@ def _dnn_lib():
...
-            lib_name = ctypes.util.find_library('cudnn64_5.dll')
-            if lib_name is None:
-                lib_name = ctypes.util.find_library('cudnn64_4.dll')
+            for name in ['cudnn64_5.dll', 'cudnn64_4.dll']:
+                lib_name = ctypes.util.find_library(name)
+                if lib_name:
+                    break
         if lib_name is None:
-            raise RuntimeError('Could not find cudnn library')
+            raise RuntimeError('Could not find cudnn library (maybe your are using a )')
```
lamblin: a what?

abergeron:
I did another rebase and added a test for lstm mode and fixed the issues of the gradient of that.

nouiz:
there is conflict in this branch.

abergeron:
Rebased again.

theano/gpuarray/dnn.py
```diff
@@ -1586,6 +1645,597 @@ def make_node(self, x, dy, scale, x_mean, x_invstd, epsilon=1e-4):
...
+        # of them.  This leads to the situation that the other is
+        # considered "disconnected" by theano in the gradient.
+        # However we know that this isn't really the case so we fix it
+        # up here.
+
+        # If both dy and dhy are disconnected the fixup will fail, but
+        # that's ok as in that case we really are disconnected.
+        if isinstance(dy.type, DisconnectedType):
+            dy = as_gpuarray_variable(dhy[-1],
+                                      context_name=dhy.type.context_name)
```
lamblin: So, if `dy` is `DisconnectedType`, I think we should pass zeros of the right size.
If  `dhy` is disconnected as well, then I think the `Lop` method would not be called at all.

theano/gpuarray/dnn.py
```diff
@@ -1586,6 +1645,597 @@ def make_node(self, x, dy, scale, x_mean, x_invstd, epsilon=1e-4):
...
+        rnn_mode: {'rnn_relu', 'rnn_tanh', 'lstm', 'gru'}
+          See cudnn documentation for cudnnRNNMode_t.
+        input_mode: {'linear', 'skip'}
+          linear: input will be multiplied by a biased matrix
+          skip: No operation is performed on the input.  The size must match the hidden size.
+        direction_mode: {'unidirectional', 'bidirectional'}
+          unidirectional: The network operates recurrently from the
+                          first input to the last.
+
+          bidirectional: The operates from first to last then from last to first and concatenates the results at each layer.
```
lamblin: "the network operates"

theano/gpuarray/dnn.py
lamblin: There should be a warning that even though `dropout` and `dropout_seed` are parameter, only `dropout=0` is supported and the seed should not matter then.
Or maybe remove them from the interface if they are not really supported.

theano/gpuarray/dnn.py
```diff
@@ -1586,6 +1645,597 @@ def make_node(self, x, dy, scale, x_mean, x_invstd, epsilon=1e-4):
...
+                      GpuArrayType('uint8', (False,),
+                                   context_name=context_name)()])
+
+    def c_code_cache_version_apply(self, node):
+        # disable the cache since we can't pickle contexts
+        return None
+
+
+def _make_dropout_desc(dropout, seed, context_name):
+    desc, states = theano.function(
```
abergeron: Better to disable them than to get bitten later.

theano/gpuarray/dnn.py
```diff
@@ -1586,6 +1645,597 @@ def make_node(self, x, dy, scale, x_mean, x_invstd, epsilon=1e-4):
...
+  err = cudnnGetRNNLinLayerBiasParams(%(handle)s, %(desc)s, %(layer)s, xdesc, wdesc, w, %(id)s, odesc, &o);
+  if (err != CUDNN_STATUS_SUCCESS) {
+    cudnnDestroyTensorDescriptor(xdesc);
+    cudnnDestroyFilterDescriptor(wdesc);
+    cudnnDestroyFilterDescriptor(odesc);
+    PyErr_SetString(PyExc_RuntimeError, "can't fetch bias for id %(id)s");
+    %(fail)s
+  }
+  off = (intptr_t)o - (intptr_t)w;
+  assert(off >= 0 && "bias");
```
abergeron: I am (reasonably) confident it will never trigger, but if it does, then it will be much easier to debug than if it wasn't there.

Also, I disagree that asserts are hard to debug.  They print out a clear error message and are super easy to catch with gdb.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1095,3 +1097,150 @@ def test_batchnorm_inference():
...
+        fn = theano.function([X, h0], out, mode=mode_with_gpu)
+        cost = T.mean((Y - out)**2)
+        grad = T.grad(cost, [X, h0] + params)
+        grad_fn = theano.function([X, Y, h0], grad, mode=mode_with_gpu)
+        return fn, grad_fn
+
+    ref_fn, ref_grad_fn = funcs(last_layer.output(),
+                                model.get_params())
+    cudnn_fn, cudnn_grad_fn = funcs(rnnb.apply(params_cudnn, X, h0)[0],
+                                    [params_cudnn])
```
lamblin: We would also need a test where the output (that is used for the cost) depends on the hidden state at the last step (hy) only, and on a mixture of hy and y.
This is the only way of checking that `dhy` and `dy` work correctly together.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1095,3 +1097,150 @@ def test_batchnorm_inference():
...
+        fn = theano.function([X, h0, c0], out, mode=mode_with_gpu)
+        cost = T.mean((Y - out)**2)
+        grad = T.grad(cost, [X, h0, c0] + params)
+        grad_fn = theano.function([X, Y, h0, c0], grad, mode=mode_with_gpu)
+        return fn, grad_fn
+
+    ref_fn, ref_grad_fn = funcs(last_layer.output(),
+                                model.get_params())
+    cudnn_fn, cudnn_grad_fn = funcs(rnnb.apply(params_cudnn, X, h0, c0)[0],
+                                    [params_cudnn])
```
lamblin: Same here, we should test a cost that depends on `cy` as well.

theano/gpuarray/dnn.py
abergeron: * `dcy` could be connected (for LSTM), while `dy` and `dhy` are disconnected.
 * `y == hy[-1]`.  I verified this.  So it follows that `dy == dhy[-1]`.
 * We get an uninformative message even without the fix if dhy is disconnected, unless we use the special handling I am doing.
 * I can try leaving dy NULL if it's disconnected, but the documentation seemed to imply that it was a required parameter.
   * If that doesn't work I'll add a test for dhy connected and dy disconnected.

theano/gpuarray/dnn.py
lamblin: I'm not sure what would bite us the harder: not disabling them, or disabling them and preventing users from using test values in that part of the code.

theano/gpuarray/dnn.py
lamblin: - I meant, if all outputs are disconnected, `Lop` should not be called.
- If that is true even with multiple layers, it means that the shapes given in lines 2064 and following do not make sense. Either `y` should have a "numLayers" somewhere, or `hy` should have "seqLength".
  - Even then, just because the values are equal does not mean we should pass the gradient twice.

I still think the right value to pass to `dy` if disconnected is `zeros_like(y)`.

theano/gpuarray/dnn.py
abergeron: I'll remove them.

theano/gpuarray/dnn.py
abergeron: I'm ok with that.

theano/gpuarray/dnn.py
abergeron: I'll change it to zeros_like then.

abergeron:
I don't have any reference code for bidirectional rnns and I would probably waste a lot of time trying to come up with some.

Also, I'm not trying to write regression tests for cudnn, just for the bindings and the bidirectional case is basically a larger hidden_size for everything, so if the current tests pass, that should work too.

The rest of the comments should be addressed and I also rebased.

lamblin:
For bidirectional RNNs, I was mostly worried about the memory layout of the initial states, final states, and outputs.
But indeed, the reference code would be harder. We can leave that to future work.

lamblin:
I'll merge because I think it is ready, but we should make sure it is documented online.
It can be a simple `audoc` on the main class, but it should be searchable from Google and the built-in search box.

Issue #4545, opened 2016-05-21
Theano cannot find cuDNN on win7


oiale:
Hello,

I can't get cudnn working with Theano 0.9.0.dev1 on win7 x64 with gtx 660. I downloaded cudnn v5 and unpacked files over cuda installation (include > include, bin > bin and so on). However, when I check it with
python -c "from theano.sandbox.cuda.dnn import dnn_available as da; print(da() or da.msg)" 
I get the following error:
Can not compile with cuDNN. We got this error:
b''

Is there a way to fix it or at least get more useful output on the error?

nouiz:
You can set a break point by using pdb in the file theano/sandbox/cuda/__init__.py;

Add this the line "import pdb;pdb.set_trace()" at the start of the function dnn_available(). This will allow you to investigate what is going on. If you have questions, don't hesitate.

genichyar:
I had the same problem. It occured when compiler_bindir had spaces. For example, "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin".

My solution is replace
```
params.extend(['--compiler-bindir',
                               config.nvcc.compiler_bindir])
```
to
```
params.extend(['--compiler-bindir',
                               path_wrapper + config.nvcc.compiler_bindir + path_wrapper])
```
in `theano/sandbox/cuda/__init__.py`

With this fix cudnn works well.

nouiz:
Can you make a PR with this?

lamblin:
Fixed in #4923, and we're using the new back-end now anyway.

PR #4923, opened 2016-08-31
Fix cudnn when compiler-bindir specified in Windows. Fix issue #4545

genichyar:


nouiz:
ok to merge when travis pass. thanks.

nouiz:
jenkins test this

PR #4925, opened 2016-08-31
Add a wrapper function for kernels to simplify calling.

abergeron:
This comes with tests and docs changes.  This also updates GpuEye to use it so that we have a real-world example.

I did this in preparation of the CorrMM port but this is a worthwhile improvement regardless.

### DIFF
--- a/theano/gpuarray/tests/tstgpueye.c
+++ b/theano/gpuarray/tests/tstgpueye.c
@@ -37,7 +37,7 @@ int APPLY_SPECIFIC(tstgpueye)(PyArrayObject *n, PyArrayObject *m,
   ls = 1;
   gs = 256;
   /* The eye_call name comes from the kernel declaration above. */
-  err = eye_call(1, &ls, &gs, 0, (*z)->ga.data, dims[0], dims[1]);
+  err = eye_call(1, &gs, &ls, 0, (*z)->ga.data, dims[0], dims[1]);
   if (err != GA_NO_ERROR) {
     PyErr_Format(PyExc_RuntimeError,
                  "gpuarray error: kEye: %s. n%lu, m=%lu.",
--- a/theano/gpuarray/tests/tstgpueye.c
+++ b/theano/gpuarray/tests/tstgpueye.c
@@ -34,13 +34,10 @@ int APPLY_SPECIFIC(tstgpueye)(PyArrayObject *n, PyArrayObject *m,
   if (*z == NULL)
     return -1;
 
-  args[0] = (*z)->ga.data;
-  args[1] = &dims[0];
-  args[2] = &dims[1];
   ls = 1;
   gs = 256;
-  /* The k_eye name comes from the kernel declaration above. */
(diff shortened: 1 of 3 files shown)


theano/gpuarray/tests/tstgpueye.c
```diff
@@ -34,13 +34,10 @@ int APPLY_SPECIFIC(tstgpueye)(PyArrayObject *n, PyArrayObject *m,
...
 
-  args[0] = (*z)->ga.data;
-  args[1] = &dims[0];
-  args[2] = &dims[1];
   ls = 1;
   gs = 256;
-  /* The k_eye name comes from the kernel declaration above. */
-  err = GpuKernel_call(&k_eye, 1, &ls, &gs, 0, args);
+  /* The eye_call name comes from the kernel declaration above. */
+  err = eye_call(1, &ls, &gs, 0, (*z)->ga.data, dims[0], dims[1]);
```
nouiz: This user interface is already much better. What about those possible improvment:

1) eye_call(1, &ls, &gs, 0, *z, dims[0], dims[1]); 
   *z would be just z for inputs.

The in the wrapper it would do: "(*z)->ga.data" automatically for GpuArray param. The user will then only need to continue to manually pass the offset. But when he will pass the offset, then I'm not sure it is a good change.

2) why not swap ls and gs? That way a cuda call k<<<gs, ls, shared>>>(...), would be converted to k(gs, ls, shared, ...). It would help prevent some conversion mistake. If we don't do this change, we must make it clear in the doc and explain why.

nouiz:
jenkins fail:

```
  File "/home/jenkins-gpu0/workspace/Theano gpu/theano/gpuarray/basic_ops.py", line 270, in get_ctype
    return dtype.name + '_t'
AttributeError: ('The following error happened while compiling the node', forall_inplace,cpu,scan_fn}(Shape_i{0}.0, GpuSubtensor{int64:int64:int8}.0, GpuIncSubtensor{InplaceSet;:int64:}.0, Shape_i{0}.0), '\n', 'The following error happened while compiling the node', GpuCAReduceCuda{add}(<GpuArrayType<None>(float32, (False, False, False))>), '\n', "'str' object has no attribute 'name'")
```

theano/gpuarray/tests/tstgpueye.c
abergeron: 1) want to take gpudata pointers to allow people to work with low-level gpudata object directly if necessary.  Other this will force them to use the GpuKernel_call interface for that.

2) ok.

nouiz: 1) Can you give something else that they would like to pass instead of data
for inputs tagged as GpuArray? The shapes/strides will be passed with other
types. The offset must be passed with a different type as well as the
flags. I don't see what else could be passed. A kernel that support only
CUDA and the user would pass the ptr+offset in the same parameter instead
of 2 parameters?

abergeron:
The problem on jenkins is a configuration one, but all the tests passed, so this should be good to merge.

PR #4932, opened 2016-09-02
Don't move complex to the GPU in the new back-end.

nouiz:


theano/gpuarray/dnn.py
```diff
@@ -1282,6 +1282,11 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
 
     """
     img = gpu_contiguous(img)
+    if mode == "sum":
+        ret = GpuDnnPool(mode="average_inc_pad")(img, ws, stride, pad)
+        context_name = ret.type.context_name
+        window_elem = numpy.prod(ws).astype(ret.dtype)
```
lamblin: ws can be symbolic now, right? It may be the case when the test fails on Jenkins.

nouiz: Yes, but numpy.prod work on tuple of python int and Theano variables. I
have checked that before doing the commit.

lamblin: It still think it makes this test fail: http://darjeeling.iro.umontreal.ca:8080/job/Theano%20gpu/203/testReport/junit/theano.gpuarray.tests/test_dnn/test_pooling_opt/
Maybe it only works if the Theano version of `prod` supports the `out=...` keyword arg.

nouiz: I changed to use theano.tensor.prod(). It will work with constant and theano variables.

theano/gpuarray/dnn.py
```diff
@@ -1282,6 +1282,11 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
 
     """
     img = gpu_contiguous(img)
+    if mode == "sum":
```
abergeron: Update the doc to mention sum.

theano/sandbox/cuda/dnn.py
```diff
@@ -2005,6 +2005,11 @@ def dnn_pool(img, ws, stride=(1, 1), mode='max', pad=(0, 0)):
 
     """
     img = gpu_contiguous(img)
+    if mode == "sum":
+        ret = GpuDnnPool(mode="average_inc_pad")(img, ws, stride, pad)
+        window_elem = theano.tensor.prod(ws).astype(ret.dtype)
+        return as_cuda_ndarray_variable(ret * window_elem)
+
```
abergeron: Didn't we decide to not add new features in the old backend?

nouiz: Currently it crash. I didn't wanted to test a crash fix when I can just
copy/paste the code from the new back-end. There was new ops.

theano/gpuarray/dnn.py
nouiz: done

theano/gpuarray/opt.py
```diff
@@ -344,6 +346,9 @@ def apply(self, fgraph):
                          self.local_optimizers_map.get(type(c.op), []))):
                         move_to_GPU = True
             new_ops = None
+            if move_to_GPU and any(["complex" in i.dtype for i in node.inputs]):
```
abergeron: This line throws errors about Variable not having a `dtype` attribute.

nouiz: fixed.

Issue #4647, opened 2016-06-17
Uncanonicalize DimShuffle inside Alloc, Reshape and Subtensor


nouiz:
We need to explicitly make put in the graph DimShuffle for canonicalization.

But we should merge them in the Alloc, Reshape and Subtensor that do it itself. This make the graph smaller and lower the overhead.
- [x] Alloc(DimShuffle(x), ...) - > Alloc(x, ...) when the dimshuffle just add dimensions at the left.
- [x] Reshape(Dimshuffle(x), shp) -> Reshape(x, shp) when the dimshuffle just add dimsions (anywhere) and don't change the order of dimensions.
- [x]  Subtensor
  
  some_col[i].dimshuffle() => some_col[i, 0], or
  x[i:j, :, k:l].dimshuffle(0, 2) => x[i:j, 0, k:l] if x.broadcastable == (False, True, False).
  
  In general, when a dimshuffle only removes broadcastable dimensions that were kept intact by the subtensor, we can index by 0 on those dimensions instead.
  Register this in canonicalize.

olimastro:
should be in the file theano/tensor/opt_uncanonicalize.py.

If you can only partially remove the DimShuffle, don't remove it. It is useless.

lamblin:
For the subtensor case, the following could be done:
- `some_col[i].dimshuffle() => some_col[i, 0]`, or
- `x[i:j, :, k:l].dimshuffle(0, 2) => x[i:j, 0, k:l]` if `x.broadcastable == (False, True, False)`.
In general, when a dimshuffle only removes broadcastable dimensions that were kept intact by the subtensor, we can index by 0 on those dimensions instead.

nouiz:
Do we have a canonicalization opt that will add those dimshuffles? I don't
think so(and didn't find in a quick search in tensor/opt.py). I don't think
we should make one as this will probably break many other optimization that
are related to subtensors like subtensor_make_vector,
local_subtensor_merge, local_subtensor_of_alloc (maybe others, I stopped
the search).

I think the canonical form to help other optimizations to be easier to
implement would be for now what you suggest. Not an uncanolicalize form.
Doing the reverse would help merge some dimshuffles, but would make at
least a few optimizations more complicated. I'm not sure it is a good gain
to merge DimShuffle. In particular as we didn't see real gain in real life.

I would suggest your dimshuffle(subtensor()) -> subtensor opt at the
canonical phase for now. If later we see a need for the reverse, we could
[...]

PR #4935, opened 2016-09-02
[WIP] Opt Alloc(DimShuffle(...)...) -> Alloc(...) #4647

olimastro:
First attempt to fix the first point of the issue: dropping the dimshuffle of the Alloc if the dimshuffle only adds dimension to the left. This need review since I am still unfamiliar on how to play around with these classes!

gh-4647

### DIFF
--- a/theano/tensor/opt_uncanonicalize.py
+++ b/theano/tensor/opt_uncanonicalize.py
@@ -107,6 +107,8 @@ def local_alloc_dimshuffle(node):
     """
     If a dimshuffle is inside an alloc and only adds dimension to the
     left, remove it.
+
+    Alloc(DimShuffle(x), ...) - > Alloc(x, ...)
     """
     if node.op == T.alloc:
         input_ = node.inputs[0]
@@ -114,7 +116,7 @@ def local_alloc_dimshuffle(node):
             # check if it only adds dimension to the left
             new_order = input_.owner.op.new_order
             expected_new_order = ('x',) * (input_.ndim - input_.owner.inputs[0].ndim) + \
-                    tuple(range(input_.owner.inputs[0].ndim))
+                tuple(range(input_.owner.inputs[0].ndim))
             if new_order != expected_new_order:
                 return False
             return [T.alloc(input_.owner.inputs[0], *node.inputs[1:])]
@@ -127,6 +129,8 @@ def local_reshape_dimshuffle(node):
     """
     If a dimshuffle is inside a reshape and does not change the order
     of dimensions, remove it.
+
+    Reshape(Dimshuffle(x), shp) -> Reshape(x, shp)
     """
     if isinstance(node.op, T.Reshape):
         input_ = node.inputs[0]
--- a/theano/tensor/opt_uncanonicalize.py
+++ b/theano/tensor/opt_uncanonicalize.py
@@ -102,7 +102,7 @@ def local_max_to_min(node):
 
 
 @register_uncanonicalize
-@gof.local_optimizer([T.alloc])
+@gof.local_optimizer([T.Alloc])
 def local_alloc_dimshuffle(node):
     """
     If a dimshuffle is inside an alloc and only adds dimension to the
@@ -113,24 +113,16 @@ def local_alloc_dimshuffle(node):
         if input_.owner and isinstance(input_.owner.op, DimShuffle):
             # check if it only adds dimension to the left
             new_order = input_.owner.op.new_order
-            flag = False
-            for i, dim in enumerate(new_order_bool):
-                if i == 0 and dim == 'x':
-                    flag = True
-                elif dim == 'x' and flag:
-                    continue
-                elif i > 0 and flag:
-                    flag = False
-                elif i > 0 and not dim == 'x':
-                    continue
-                else:
-                    return False
+            expected_new_order = ('x',) * (input_.ndim - input_.owner.inputs[0].ndim) + \
+                    tuple(range(input_.owner.inputs.ndim))
+            if new_order != expected_new_order:
+                return False
             return input_.owner.inputs
     return False
 
 
 @register_uncanonicalize
-@gof.local_optimizer([T.reshape])
+@gof.local_optimizer([T.Reshape])
 def local_reshape_dimshuffle(node):
     """
     If a dimshuffle is inside a reshape and does not change the order
--- a/theano/tensor/opt_uncanonicalize.py
+++ b/theano/tensor/opt_uncanonicalize.py
@@ -104,20 +104,48 @@ def local_max_to_min(node):
 @register_uncanonicalize
 @gof.local_optimizer([T.alloc])
 def local_alloc_dimshuffle(node):
+    """
+    If a dimshuffle is inside an alloc and only adds dimension to the
+    left, remove it.
+    """
     if node.op == T.alloc:
         input_ = node.inputs[0]
-        if getattr(input_, 'owner', None) and isinstance(input_.owner.op, DimShuffle):
+        if input_.owner and isinstance(input_.owner.op, DimShuffle):
             # check if it only adds dimension to the left
-            pattern = input_.type.broadcastable
-            if not pattern[0]:
-                return False
-            j = 0
-            for i, bool_ in enumerate(pattern):
-                if not bool_:
-                    j = i
-                    break
-            if sum(pattern[j:]) == 0:
-                return input_.inputs
-            else :
-                return False
+            new_order = input_.owner.op.new_order
+            flag = False
+            for i, dim in enumerate(new_order_bool):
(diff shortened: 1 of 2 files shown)


theano/tensor/opt_uncanonicalize.py
```diff
@@ -98,3 +99,25 @@ def local_max_to_min(node):
...
+
+@register_uncanonicalize
+@gof.local_optimizer([T.alloc])
+def local_alloc_dimshuffle(node):
+    if node.op == T.alloc:
+        input_ = node.inputs[0]
+        if isinstance(input_.owner.op, DimShuffle):
+            # check if it only adds dimension to the left
+            pattern = input_.type.broadcastable
+            if not pattern[0] :
```
nouiz: The way to check the change that the dimshuffle do is by doing input_.owner.op.new_order. Go see the docstring of the class dimshuffle in the file tensor/elemwise.py to know how it work.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -106,14 +106,14 @@ def local_max_to_min(node):
 def local_alloc_dimshuffle(node):
     if node.op == T.alloc:
         input_ = node.inputs[0]
-        if isinstance(input_.owner.op, DimShuffle):
+        if getattr(input_, 'owner', None) and isinstance(input_.owner.op, DimShuffle):
```
nouiz: variable always have an owner field. So you can simplify the first condition as "input_.owner and ..."

nouiz: if input_.owner and isinstance(input_.owner.op, DimShuffle):

theano/tensor/opt_uncanonicalize.py
```diff
@@ -104,20 +104,48 @@ def local_max_to_min(node):
...
-                if not bool_:
-                    j = i
-                    break
-            if sum(pattern[j:]) == 0:
-                return input_.inputs
-            else :
-                return False
+            new_order = input_.owner.op.new_order
+            flag = False
+            for i, dim in enumerate(new_order_bool):
```
nouiz: new_order_bool don't exist. So the opt will crash.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -104,20 +104,48 @@ def local_max_to_min(node):
...
+                if i == 0 and dim == 'x':
+                    flag = True
+                elif dim == 'x' and flag:
+                    continue
+                elif i > 0 and flag:
+                    flag = False
+                elif i > 0 and not dim == 'x':
+                    continue
+                else:
+                    return False
```
nouiz: this condition is complicated. I think this would be simpler to understand.

expected_new_order = ('x',) * (input_ndim - input_.owner.inputs.ndim) + tuple(range(input_.owner.inputs.ndim))
if new_order != expected_new_order: 
    return

theano/tensor/opt_uncanonicalize.py
```diff
@@ -98,3 +99,53 @@ def local_max_to_min(node):
                                  max.owner.op.axis)(neg.owner.inputs[0])]
 
     return False
+
+
+@register_uncanonicalize
+@gof.local_optimizer([T.alloc])
```
nouiz: use T.Alloc. It must be a class, not an instance.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -98,3 +99,53 @@ def local_max_to_min(node):
...
+                elif i > 0 and not dim == 'x':
+                    continue
+                else:
+                    return False
+            return input_.owner.inputs
+    return False
+
+
+@register_uncanonicalize
+@gof.local_optimizer([T.reshape])
```
nouiz: idem use T.Reshape.

theano/tensor/opt_uncanonicalize.py
olimastro: Ah I made this complicated condition because I thought we could shuffle also dimensions and add more to the left. I understand now this is not something we want.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -128,16 +128,16 @@ def local_reshape_dimshuffle(node):
...
+            new_order = input_.owner.op.new_order
             offset = 0
             for i, dim in enumerate(new_order):
                 if dim == 'x':
                     offset += 1
                     continue
                 elif i != dim + offset:
                     return False
-            return input_.owner.inputs
+            return [T.reshape(input_.owner.inputs[0], tuple(node.inputs[1].owner.inputs))]
```
olimastro: I am not sure about this line. Is the inputs[1] of the reshape always a MakeVector so in order to go grab the new shape tuple you need to go into the inputs of the MakeVector node?

nouiz:
There is a few travis errors:

/home/travis/build/Theano/Theano/theano/tensor/opt_uncanonicalize.py:117:21: E127 continuation line over-indented for visual indent

  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_opt_uncanonicalize.py", line 10, in <module>
    from theano.tensor.opt import out2in
ImportError: cannot import name out2in

"get it from theano.gof.opt instead"


  File "/home/travis/build/Theano/Theano/theano/tensor/opt_uncanonicalize.py", line 142, in local_reshape_dimshuffle
    return [T.reshape(input_.owner.inputs[0], tuple(node.inputs[1].owner.inputs))]
AttributeError: 'NoneType' object has no attribute 'inputs'

theano/tensor/opt_uncanonicalize.py
```diff
@@ -98,3 +99,45 @@ def local_max_to_min(node):
...
 
     return False
+
+
+@register_uncanonicalize
+@gof.local_optimizer([T.Alloc])
+def local_alloc_dimshuffle(node):
+    """
+    If a dimshuffle is inside an alloc and only adds dimension to the
+    left, remove it.
```
nouiz: Add the example from the issue in the docstring. Idem bellow.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -98,3 +99,45 @@ def local_max_to_min(node):
...
+        input_ = node.inputs[0]
+        if input_.owner and isinstance(input_.owner.op, DimShuffle):
+            new_order = input_.owner.op.new_order
+            offset = 0
+            for i, dim in enumerate(new_order):
+                if dim == 'x':
+                    offset += 1
+                    continue
+                elif i != dim + offset:
+                    return False
```
nouiz: The entire if don't do what we want. I think it should be:
if dim == 'x':
    continue
elif i != dim + offset:
   return
else:
    offset += 1

THis will make sure we don't swap dimensions and allow 'x' anywhere.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -98,3 +99,45 @@ def local_max_to_min(node):
...
+            return [T.reshape(input_.owner.inputs[0], tuple(node.inputs[1].owner.inputs))]
```
nouiz: Here we want the inputs of the dimshuffle to be the input of the reshape, but we don't want to change the other parameter of the reshape. The output shape of the reshape should not change. We just remove a useless reshape here.

olimastro: I do not understand. To quote the issue we wanted "Reshape(Dimshuffle(x), shp) -> Reshape(x, shp) ". We don't remove any reshape, we remove a dimshuffle. So we want the inputs of the reshape to be the input of the dimshuffle. The way your comment is written, I read the opposite!

So `input_.owner.inputs[0]` refers to the inputs of the dimshuffle which we give to the new reshape. The shape now refers to `tuple(node.inputs[1].owner.inputs)` and this is might be bad, but it is related to my comment on that line. I asked if this was the correct way to retrieve the shape. I have noticed while doing the tests that the information of shp was stored in a node (makevector something).

theano/tensor/opt_uncanonicalize.py
olimastro: If you do offset += 1 only at the else it would on a new order such as ('x', 0, 1, 'x') the following :
- first pass of loop, dim == 'x' is True so continue
- second pass, i != dim + offset =­> 1 != 0 + 0 is True so return False. It should not return False on this case.
What is the thing I don't see?

olimastro:
Two of the comments have not been dealt with in this commit, please refer to my replies.

theano/tensor/opt_uncanonicalize.py
nouiz:
```
offset = 0
for dim in new_order:
    if dim == 'x':
        continue
    elif dim != offset:
        return False
    else:
        offset += 1
```

PR #4936, opened 2016-09-03
Break opt

Sentient07:
Lowering the number of iteration in Cannonicalize.


nouiz:
Closing old PR that won't be finished.

I'm not sure we still need this. If this happen, it is still in the history.

PR #4937, opened 2016-09-03
Fix d3viz multi-level/inheritance tags

gw0:
I tried to visualize a Theano computation graph build with Keras by accessing it through `model.train_function.function`. I am not exactly sure what happens, but instead of a single trace tuple a list of tuples gets returned (each one containing a path, line, module, src). This results in the following exception.

```
  File "./train.py", line 203, in <module>
    d3v.d3viz(model.train_function.function, 'd3v_model.html')
  File "/usr/local/lib/python2.7/dist-packages/theano/d3viz/d3viz.py", line 79, in d3viz
    graph = formatter(fct)
  File "/usr/local/lib/python2.7/dist-packages/theano/d3viz/formatting.py", line 199, in __call__
    vparams['tag'] = var_tag(var)
  File "/usr/local/lib/python2.7/dist-packages/theano/d3viz/formatting.py", line 311, in var_tag
    path = os.path.basename(path)
  File "/usr/lib/python2.7/posixpath.py", line 114, in basename
    i = p.rfind('/') + 1
AttributeError: 'tuple' object has no attribute 'rfind'
```

```python
(Pdb) print tag.trace
[[('./train.py', 16, '<module>', 'from keras.utils.visualize_util import plot'),
  ('/usr/local/lib/python2.7/dist-packages/keras/__init__.py', 2, '<module>', 'from . import backend'),
  ('/usr/local/lib/python2.7/dist-packages/keras/backend/__init__.py', 61, '<module>', 'from .theano_backend import *'),
  ('/usr/local/lib/python2.7/dist-packages/keras/backend/theano_backend.py', 18, '<module>', "_LEARNING_PHASE = T.scalar(dtype='uint8', name='keras_learning_phase')  # 0 = test, 1 = train")
]]
```

As a solution I am proposing this PR to use the last trace tuple for constructing the d3viz tag.

### DIFF
--- a/theano/d3viz/formatting.py
+++ b/theano/d3viz/formatting.py
@@ -292,7 +292,10 @@ def var_tag(var):
     """Parse tag attribute of variable node."""
     tag = var.tag
     if hasattr(tag, 'trace') and len(tag.trace) and len(tag.trace[0]) == 4:
-        path, line, _, src = tag.trace[0]
+        if isinstance(tag.trace[0][0], (tuple, list)):
+            path, line, _, src = tag.trace[0][-1]
+        else:
+            path, line, _, src = tag.trace[0]
         path = os.path.basename(path)
         path = path.replace('<', '')
         path = path.replace('>', '')


lamblin:
I think we changed the semantics of `tag.trace` to allow for multiple traces in cases where operations declared at different places in the source are merged together.
In that case, maybe there would be a way of displaying all of the traces?
For now, I restarted the timed out test.

nouiz:
I'll merge it now as it fix a crash, but I agree with @lamblin if we could print all of them, it would be better.

gw0:
Imho printing all of them would just mean concatenating all strings together. But it is not obvious how to make the output nice and readable in a d3viz visualization, or is it?

nouiz:
@cangermueller Can you answer that questions?

thanks

PR #4942, opened 2016-09-06
Improve the example for strict=True in the scan() doc.

waiwnf:
Change the example illustrating the effect of strict=True in the scan() doc to actually trigger the proper error. Existing example triggers a different one (non_sequences not aligned with the arguments of OneStep) if one simply omits some of the [W, bvis, bhid] from non_sequences.

### DIFF
--- a/doc/library/scan.txt
+++ b/doc/library/scan.txt
@@ -400,7 +400,6 @@ Using the original Gibbs sampling example, with ``strict=True`` added to the
         return trng.binomial(size=vsample.shape, n=1, p=vmean,
                              dtype=theano.config.floatX)
 
-
     # The new scan, adding strict=True to the original call.
     # Produces an error, because W, bvis and bhid are used by OneStep, but not
     # passed explicitly.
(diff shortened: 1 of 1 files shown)


doc/library/scan.txt
```diff
@@ -383,23 +383,33 @@ Using shared variables - the strict flag
...
+    values, updates = theano.scan(OneStep,
                                   outputs_info=sample,
-                                  non_sequences=[W, bvis, bhid],
                                   n_steps=10,
                                   strict=True)
 
-If you omit to pass ``W``, ``bvis`` or ``bhid`` as a ``non_sequence``, it will
-result in an error.
+The above will result in an error, indicating that ``OneStep`` relies on
+variables that are not passed as its explicit arguments.
```
nouiz: Can you add the error raised? This way if people google the error, they will find this doc:

MissingInputError: An input of the graph, used to compute InplaceDimShuffle{1,0}(<TensorType(float64, matrix)>), was not provided and not given a value.Use the Theano flag exception_verbosity='high',for more information on this error.

Otherwise, this PR look good. thanks!

nouiz:
@ReyhaneAskari @lamblin another quick PR that could be finished for the RC.

PR #4943, opened 2016-09-07
Remove nose-parameterized dependency from theano.sparse

kuza55:
Theano.sparse takes a transitive runtime dependency on nose-parameterized which is not bundled as a dependency when you install theano from pip

This turns out to be due to an import of a test file which imports nose-parameterized.

The util function being called was quite trivial so I inlined it. 

This is causing me trouble trying to add sparse matrix support to Keras https://github.com/fchollet/keras/pull/3695

### DIFF
--- a/theano/sparse/basic.py
+++ b/theano/sparse/basic.py
@@ -22,7 +22,6 @@ import theano
 from theano import gof, tensor, scalar, config
 from theano.gradient import DisconnectedType
 from theano.sparse.utils import hash_from_sparse
-import theano.tests.unittest_tools as utt
 from theano.gradient import grad_not_implemented, grad_undefined
 from theano.sparse.type import SparseType, _is_sparse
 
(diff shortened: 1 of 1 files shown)


nouiz:
It look goods. Ok to merge when travis pass.

jenkins test this

theano/sparse/basic.py
```diff
@@ -233,9 +233,7 @@ def conv_op(*inputs):
         out = op(*ipt)
         return oconv(out)
 
-    seed_rng()
-    rng = numpy.random
-    T.verify_grad(conv_op, dpt, 2, rng, *args, **kwargs)
```
nouiz: We need a way to keep the deterministic behavior. So we need to continue to pass a seeded rng and we need to continue to use the mechanism in utt. Instead of duplicating the code here, I see 2 possibles fix:

- In unittest_tools.py put a try..except ImportError around the nose parametrised.
- Move the sparse_verify_grad to a test file. It is only used there. I did that in this PR: https://github.com/Theano/Theano/pull/4944

@abergeron, what do you think?

kuza55: Is the seed_rng() actually doing anything though? It seems that it's fetching a seed from config.unittests.rseed, but is that actually non-None when someone calls verify_grad_sparse? I'm not clear who the expected callers for this function are.

nouiz: It is tests that is calling verify_grad and sparse_verify_grad. The
seed_rng seed the default numpy rng.

PR #4944, opened 2016-09-07
Make nose-paremetrised optional for theano.sparse

nouiz:
another try to fix the same problem as gh-4943


abergeron:
Ok when travis passes.

Issue #4931, opened 2016-09-02
cuDNN batch norm does not support 5D input


mdering:
I believe this is an issue to be opened here. Briefly, the GpuDnnBatchNorm operation has a few assertions that the input must be 4 dimensions. However, it seems like the cudnn function it calls, cudnnBatchNormalizationForwardTraining, can accept both 4D and 5D tensors as input. I'm not an experienced cudnn programmer at all but this seems like it could possibly be a quick fix. Could someone take a look at this?

nouiz:
Just to make sure that people who follow this, there is a PR here for that:

https://github.com/Theano/Theano/pull/4948

nouiz:
the PR was merged 26 day ago

PR #4948, opened 2016-09-08
GpuDnnBatchNorm with 5d inputs

gvtulder:
Here's a little patch that would add 5d support to the GPU batch normalization ops (see #4931).

The tests pass now, but I'm not really sure about the tolerances used in the tests. Some tests would fail with what seemed to be relatively small differences. Using larger input values and changing some of the atol/rtol values fixed that.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -2304,7 +2304,7 @@ class GpuDnnBatchNormBase(DnnBase):
     """
 
     __props__ = ('mode', 'epsilon')
-    tensor_4d_descs = []
+    tensor_descs = []
 
     def __init__(self, mode='per-activation', epsilon=1e-4):
         DnnBase.__init__(self)
@@ -2320,7 +2320,7 @@ class GpuDnnBatchNormBase(DnnBase):
 
     def c_support_code_struct(self, node, name):
         result = ''
-        for id in self.tensor_4d_descs:
+        for id in self.tensor_descs:
             result += c_define_tensor_desc('%s_%s' % (id, name))
         return result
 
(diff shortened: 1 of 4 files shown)


theano/sandbox/cuda/dnn.py
```diff
@@ -2395,23 +2392,22 @@ def c_code(self, node, name, inputs, outputs, sub):
...
-        result += c_set_tensor4d(scale, 'bn_params_' + name, 'err' + name, sub['fail'])
-
-        # build and prepare the output variable
+        # call cuDNN function
         result += """
-if (CudaNdarray_prep_output(&%(outp)s, 4, CudaNdarray_HOST_DIMS(%(inp)s)) != 0)
-{
+// set input tensor descriptors from input tensors
+if (c_set_tensorNd(%(inp)s, bn_input_%(name)s) != 0)
+    %(fail)s
```
nouiz: put fail between "{}". As it is hiden code, it could be more then 1 statement in the futur.

nouiz:
jenkins test this

@abergeron can you review it? I think it is good, but you know that code more then me. What do you think about the threashold problem in tests?

theano/sandbox/cuda/dnn.py
gvtulder: Done. (I copied this from elsewhere, by the way. fail without {} occurs in other places as well, so the future should be careful.)

mdering:
Currently using this implementation with keras, and its working (and is a LOT faster). Thanks a lot!

gvtulder:
The Jenkins build still failed, so I increased the rtol a little more.

On my system the tests passed with floatX=float32, but they would fail with floatX=float64 if I didn't increase the tolerance.

mdering:
Will this be merged soon?

gvtulder:
Rebased to the current master. (There was a conflict introduced by PR #5028).

lamblin:
Jenkins, test this, please.

gvtulder:
Jenkins passed.

abergeron:
The gpuarray part is ok.  I didn't review the cuda part.  If we drop that part I would be ok with merging this.  Otherwise, someone else can review it.

PR #4950, opened 2016-09-08
Mixed

abergeron:


theano/gof/op.py
```diff
@@ -1434,7 +1434,15 @@ def format_c_function_args(self, inp, out):
...
-        return ", ".join(list(inp) + ["&%s" % o for o in out])
+        inp = list(inp)
+        numi = getattr(self, '_cop_numi', len(inp))
+        while len(inp) < numi:
+            inp.append('NULL')
+        out = ["&%s" % o for o in out]
+        numo = getattr(self, '_cop_numo', len(out))
+        while len(out) < numo:
+            out.append('NULL')
+        return ", ".join(inp + out)
```
nouiz: I think we should document that somewhere.

nouiz:
Except the doc, all seem good.

theano/gpuarray/dnn_base.c
```diff
@@ -54,6 +54,18 @@ c_set_tensorNd(PyGpuArrayObject *var, cudnnTensorDescriptor_t desc) {
...
+static int c_make_tensorNd(PyGpuArrayObject *var, cudnnTensorDescriptor_t *desc) {
+  cudnnStatus_t err;
+  err = cudnnCreateTensorDescriptor(desc);
+  if (err != CUDNN_STATUS_SUCCESS) {
+    PyErr_Format(PyExc_RuntimeError,
+                 "Could not create tensor descriptor: %s",
+                 cudnnGetErrorString(err));
+    return -1;
+  }
+  return c_set_tensorNd(var, *desc);
```
nouiz: should we destroy if the set don't work? idem bellow.

theano/gof/op.py
abergeron: ok

theano/gpuarray/dnn_base.c
abergeron: ok

PR #4015, opened 2016-02-08
Add some developper documentation for Scan

carriepl:

PR #4954, opened 2016-09-09
[WIP] Scan doc internal

Thrandis:
Finish PR #4015 

@nouiz I addressed the comments of @mgermain on @carriepl 's PR.

Just one small detail: what is the correct syntax : Scan Op / scan op /Scan op ?

Otherwise I think the text is quite clear, although I havn't looked into the code of the internals of scan yet.


nouiz:
It depend of the context. The class name is ScanOp. But if you talk about the theoretical context, you could use: scan op.

nouiz:
ok to merge when travis pass.

nouiz:
thanks

PR #4956, opened 2016-09-09
Use types that exist for floats.

abergeron:


nouiz:
ok when jenkins pass.

PR #4958, opened 2016-09-09
Lower the number of iteration for local_add_mul_fusion

nouiz:


0entr0py:
This PR improves compile time from 4600+s to 700s for the my Conv + Scan [code](https://gist.github.com/0entr0py/48093cb77947b76aebdc69965694c6bd) (see discussion [here](https://groups.google.com/forum/#!topic/lasagne-users/36skUnd1ZOQ)). Will this be merged soon in the master ?

nouiz:
It need to be reviewed by someone else before being merged. I don't expect
problems. It is simple code changes of stuff we already discussed. But the
others are pretty busy, so it could take a few days.

I think it is safe for you to use this branch in the mean time.

theano/tensor/opt.py
```diff
@@ -7140,16 +7140,30 @@ def local_add_mul_fusion(node):
...
+        else:
+            new_inp.append(inp)
+
+    # We ca not compare the number of inputs as Mul and Add could have
+    # 0 or 1 inputs in some corner cases.
+    if fused:
+        output = node.op(*new_inp)
+        copy_stack_trace(node.outputs[0], output)
+
+        # Does the recursion here to help lower the number of
```
abergeron: Does -> Do

abergeron:
Ok with small change in the comment to make it not confusing to read.

nouiz:
done

PR #4966, opened 2016-09-13
Correctly check which sync to call, in case both back-end are availab…

nouiz:
You can make a pull request to Theano github repo in order to fix a bug or add new features.
Once a pull request is submitted, [Travis](https://travis-ci.org/) is launched to verify the pull request
given the current stack of unit-tests in Theano.

…le, use the one used.


abergeron:
ok

PR #4968, opened 2016-09-13
Add a metaclass implementation of __props__ and generalize it to all theano objects

abergeron:
This will remove problems with inheriting __eq__ or __hash__ that shadows our own.

### DIFF
--- a/theano/gof/utils.py
+++ b/theano/gof/utils.py
@@ -4,7 +4,7 @@ import sys
 import traceback
 
 import numpy
-from six import iteritems, integer_types, string_types
+from six import iteritems, integer_types, string_types, with_metaclass
 from six.moves import StringIO
 
 from theano import config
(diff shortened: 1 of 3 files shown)


theano/gof/utils.py
```diff
@@ -152,22 +152,73 @@ class MethodNotDefined(Exception):
...
-            if hasattr(self, '__eq__') or hasattr(self, '__cmp__'):
-                raise TypeError("unhashable object: %s" % self)
-            return id(self)
 
     def __ne__(self, other):
         return not self == other
 
 
-class scratchpad:
+class scratchpad(object):
```
nouiz: why this change? I suppose it is harmless, but I'm not 100% sure.

nouiz: IRL @abergeron told that this is to be more consistent between python 2 and 3. In Python 3, it is like that by default.

abergeron: This is to harmonize python 2 (where it switches between old-style and new-style) and python 3 (where it is a no-op).

nouiz:
ok when both CI finish.

Sentient07:
Hi @abergeron . From what I understand, we need to have `__props__` defined for all the OPs compulsorily? The Elemwise Op that takes a `dict` parameter, which is not hashable ( IIRC, `openmp`). What happens to those instances ?

nouiz:
If an op don't have props, this metaclass does nothing.

Elemwise, override the \__hash__ and \__eq__ method. This is why it work
for them, it don't use the default version.

PR #4924, opened 2016-08-31
Drop old version

nouiz:
Drop python 2.6
I will add in this PR the bump of numpy and scipy version from gh-4860

fix gh-4860. Partially do gh-4507

PR #4970, opened 2016-09-13
Test newer version of numpy scipy

nouiz:
Make Python 3 tests newer numpy and scipy version. 

Follow up from https://github.com/Theano/Theano/pull/4924


nouiz:
continued in https://github.com/Theano/Theano/pull/4976

PR #4976, opened 2016-09-15
Test min version and more recent version of NumPy and SciPy.

nouiz:
We can't do the reverse. conda don't have more recent NumPy and SciPy version for python 3.3


nouiz:
jenkins test this

nouiz:
jenkins problem was not related and is fixed in master, so merging.

PR #4979, opened 2016-09-15
Delete the tests related to fftconv optimisations.

abergeron:
They've been broken for a while now and we only noticed now since we
installed scikits-cuda on the test machine.  If people really want to
use this, they can dig though the code and call it directly.  The
functional tests are still there and they pass so that's good enough
for me.


theano/sandbox/cuda/tests/test_fftconv.py
```diff
@@ -64,106 +64,6 @@ def test_full(self):
                       filters_shape=(2, 3, 3, 3),
                       border_mode='full', pad=True)
 
-    def test_opt_valid(self):
```
nouiz: I don't like the idea of just deleting the tests. If we remove it, we should remove the optimizations too and remove it from the doc. I'm not sure we should remove it... If someone could tell what they think, it would be great. @lamblin ?

nouiz: Also, only the tests  theano.sandbox.cuda.tests.test_fftconv.TestConv2dFFT.test_opt_valid and  theano.sandbox.cuda.tests.test_fftconv.TestConv2dFFT.test_opt_full fail, but you remove 4 tests.

nouiz:
I think this PR is a better fix:

https://github.com/Theano/Theano/pull/4983

Issue #4980, opened 2016-09-16
UnusedInputError in Tutorial


rasmi:
UPDATE: This only affects the 0.8.x and not current dev or 9.x versions. The error was fixed in 0.8.2 and the documentation will be updated accordingly.

Following the tutorial section on [Copying Functions](http://deeplearning.net/software/theano/tutorial/examples.html#copying-functions), the example shown raises an error:
```
---------------------------------------------------------------------------
UnusedInputError                          Traceback (most recent call last)
<ipython-input-59-5d1acb597345> in <module>()
----> 1 null_accumulator = accumulator.copy(delete_updates=True)

/Library/Python/2.7/site-packages/theano/compile/function_module.pyc in copy(self, share_memory, swap, delete_updates, name, profile)
    719                                 # can contain inplace. DebugMode check
    720                                 # that.
--> 721                                 accept_inplace=True,
    722                                 ).create(input_storage,
    723                                          storage_map=new_storage_map)

/Library/Python/2.7/site-packages/theano/compile/function_module.pyc in __init__(self, inputs, outputs, mode, accept_inplace, function_builder, profile, on_unused_input, fgraph, output_keys)
   1413 
   1414         # Check if some input variables are unused
-> 1415         self._check_unused_inputs(inputs, outputs, on_unused_input)
   1416 
   1417         # Make a list of (SymbolicInput|SymblicInputKits, indices,

/Library/Python/2.7/site-packages/theano/compile/function_module.pyc in _check_unused_inputs(self, inputs, outputs, on_unused_input)
   1551                 elif on_unused_input == 'raise':
   1552                     raise UnusedInputError(msg % (inputs.index(i),
-> 1553                                                   i.variable, err_msg))
   1554                 else:
   1555                     raise ValueError("Invalid value for keyword "

UnusedInputError: theano.function was asked to create a function computing outputs given certain inputs, but the provided input variable at index 0 is not part of the computational graph needed to compute the outputs: inc.
To make this error into a warning, you can pass the parameter on_unused_input='warn' to theano.function. To disable it completely, use on_unused_input='ignore'.
```

The `accumulator` function must have the `on_unused_input='ignore'` option added, since the `null_accumulator` function does not use the `inc` argument.

Sentient07:
No, it doesn't happen with my Theano. Did you happen to miss the update parameter ?

Sentient07:
Oh! I now see what might have happened. 
```
[...]

PR #4981, opened 2016-09-16
Fixed UnusedInputError in Tutorial

rasmi:
Creating `null_accumulator` raises an error because it does not use `inc`. The `on_unused_input='ignore'` argument fixes the error (see issue [here](https://github.com/Theano/Theano/issues/4980).

### DIFF
--- a/doc/tutorial/examples.txt
+++ b/doc/tutorial/examples.txt
@@ -289,13 +289,13 @@ functions but with different shared variables or updates. This is done using
 the :func:`copy()<theano.compile.function_module.Function.copy>` method of ``function`` objects. The optimized graph of the original function is copied,
 so compilation only needs to be performed once.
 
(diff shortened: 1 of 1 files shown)


lamblin:
I ran the example again without your modification, and it did not raise `NotImplementedError`.
Apparently, this has been fixed shortly after 0.8.2 was released.
Since updating the documentation on the `master` branch would only update the `dev` version of the documentation, what would actually be more useful is if you could submit a PR to the `0.8.X` branch of Theano/Theano (not the `master` branch). Could you also update the message to let users know that this is not necessary any more in the development or 0.9 version?

rasmi:
@lamblin, thank you for the feedback. The new PR is [here](https://github.com/Theano/Theano/pull/4988).

PR #4982, opened 2016-09-16
Function.__call__(): correct indexing variable for group of args with…

yzotov:
… shared memory

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -837,9 +837,9 @@ class Function(object):
                 for group in args_share_memory:
                     if len(group) > 1:
                         # copy all but the first
-                        for idx in group[1:]:
-                            self.input_storage[i].storage[0] = copy.copy(
-                                self.input_storage[i].storage[0])
+                        for j in group[1:]:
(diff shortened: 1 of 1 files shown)


yzotov:
I have tried to make test case but failed. I think it is really dead code because several if-else in other places of Theano prevent entering to this point. So, I think the pull request may be closed.

nouiz:
Can you tell more on what is the dead code? If it is really dead code, it
could be great to delete it. Do you have some idea/information to share? It
was a long time since I looked at that part of the code.

thanks

abergeron:
I had a quick look and the problem is that this line: https://github.com/Theano/Theano/pull/4982/files#diff-e1ec622313fa89e96ae691b21427788fL818

Will always be an empty loop since it iterates on an empty list and therefore no checking is performed.

abergeron:
There is no dead code but the whole loop that does the copy is indented one level too high to do anything effectively.

Also, `i` will be defined to the last argument because of the loop over input_storage and so the copy loop will just copy the last argument a bunch of times without this fix.

abergeron:
Finished in https://github.com/Theano/Theano/pull/6215

PR #4983, opened 2016-09-16
Fix optimization order and jenkins

nouiz:
Better fix then:

https://github.com/Theano/Theano/pull/4979

### DIFF
--- a/theano/sandbox/cuda/opt.py
+++ b/theano/sandbox/cuda/opt.py
@@ -1656,23 +1656,23 @@ register_opt()(conv_groupopt)
 
 # FFT gets the highest priority (lowest number), but is disabled by default.
 # It can be enabled by including 'conv_fft'.
-conv_groupopt.register('conv_fft_valid', local_conv_fft_valid, 10,
-                       'conv_fft')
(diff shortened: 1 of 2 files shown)


theano/sandbox/cuda/opt.py
```diff
@@ -1656,23 +1656,23 @@ def local_conv_gemm(node):
...
                        'conv_dnn',
-                       'fast_compile', 'fast_run', 'cudnn')
+                       'fast_compile', 'fast_run', 'cudnn', position=20)
 # The GEMM-based convolution comes last to catch all remaining cases.
 # It can be disabled by excluding 'conv_gemm'.
-conv_groupopt.register('local_conv_gemm', local_conv_gemm, 30,
+conv_groupopt.register('local_conv_gemm', local_conv_gemm,
                        'conv_gemm',
-                       'fast_compile', 'fast_run')
+                       'fast_compile', 'fast_run', positin=30)
```
lamblin: Should have been "position".

nouiz:
ok, I'll do it in another PR.

PR #4996, opened 2016-09-20
Scan with Checkpoints (part 2)

### LINKED ISSUE #4124: Scan memory/speed trade-off
When you take the gradient of scan, this scan grad need the full output of the forward scan. So this take much memory. It is possible to only save part of the forward output and to recompute it during the grad to have a much lower memory peak. As the grad of scan take about 6x more time the then forward, we execpt a ~20% slowdown of this, but with a much lower memory peak.

We will make example of how to do that. It should be possible to do this with 2 scan, one inside each other.

@carriepl do you want to add something about this?

Thrandis:
Fixes #4124. This is the following of @carriepl PR #4271 

#### TODO

- [x] Write the documentation.
- [x] Finish the tests.
- [x] Write the docstring.

### DIFF
--- a/theano/scan_module/scan_checkpoint.py
+++ b/theano/scan_module/scan_checkpoint.py
@@ -18,11 +18,11 @@ def scan_with_checkpoints(fn, sequences=[], outputs_info=None,
     Current assumptions :
     - Every sequence has the same length.
     - If n_steps is specified, it has the same value as the length of any
-      sequence.
+    sequence.
     - The value of "save_every_N" divides the number of steps the Scan will
-      run without remainder.
+    run without remainder.
     - Only singly-recurrent and non-recurrent outputs are used.
-      No multiple recurrences.
+    No multiple recurrences.
     - Only the last timestep of any output will ever be used.
 
     Parameters
--- a/theano/scan_module/scan_checkpoint.py
+++ b/theano/scan_module/scan_checkpoint.py
@@ -63,6 +63,21 @@ def scan_with_checkpoints(fn, sequences=[], outputs_info=None,
         the computations of scan (ie they will have to be recomputed
         during the gradient computation).
 
+    Returns
+    -------
+    tuple
+        Tuple of the form (outputs, updates); ``outputs`` is either a
+        Theano variable or a list of Theano variables representing the
+        outputs of ``scan`` (in the same order as in ``outputs_info``).
+        ``updates`` is a subclass of dictionary specifying the update rules for
+        all shared variables used in scan.
+        This dictionary should be passed to ``theano.function`` when you compile
+        your function. The change compared to a normal dictionary is that we
+        validate that keys are SharedVariable and addition of those dictionary
+        are validated to be consistent.
+        Note that only the last time step of ``outputs`` can be used with this
+        type of scan.
+
     See Also
     --------
     scan : Looping in Theano.
@@ -75,6 +90,11 @@ def scan_with_checkpoints(fn, sequences=[], outputs_info=None,
         outputs_info = [outputs_info]
     if not isinstance(non_sequences, list):
         non_sequences = [non_sequences]
+
+    # Check that outputs_info has no taps:
+    for element in outputs_info:
+        if isinstance(element, dict) and 'taps' in element:
+            raise RuntimeError("scan_with_checkpoints doesn't work with taps.")
 
     # Determine how many steps the original scan would run
     if n_steps is None:
--- a/theano/scan_module/scan_checkpoint.py
+++ b/theano/scan_module/scan_checkpoint.py
@@ -1,0 +1,82 @@
+import theano
+import theano.tensor as T
+
+def scan_with_checkpoints(fn, sequences=[], outputs_info=None,
+                          non_sequences=[], name="checkpointscan_fn",
+                          n_steps=None, save_every_N=10):
+
+    """
+    Current assumptions : 
+    - Every sequence has the same length
+    - If n_steps is specified, it has the same value as the length of any sequence
+    - The value of "save_every_N" divides the number of steps the Scan will
+      run without remainder
+    - Only singly-recurrent and non-recurrent outputs are used.
+      No multiple recurrences.
+    - Only the last timestep of any output will ever be used.
+    """
+    
+    # Standardize the format of input arguments
+    if not isinstance(sequences, list):
+        sequences = [sequences]
+    if not isinstance(outputs_info, list):
+        outputs_info = [outputs_info]
+    if not isinstance(non_sequences, list):
+        non_sequences = [non_sequences]
+    
+    # Determine how many steps the original scan would run
+    if n_steps is None:
+        n_steps = sequences[0].shape[0]
+    else:
+        n_steps = n_steps
+
+    # Compute the number of steps of the inner and of the outer scan
+    o_n_steps = n_steps / save_every_N
+    i_n_steps = save_every_N
+
+    # Establish the input variables of the outer scan
+    o_sequences = [s.reshape([s.shape[0] / save_every_N, save_every_N] +
+                             [s.shape[i] for i in range(1, s.ndim)], s.ndim + 1) for s in sequences]
+    new_nitsots = [i for i in outputs_info if i is None]
+    new_sitsots = [i for i in outputs_info if i is not None]
+    o_nonsequences = non_sequences + [i_n_steps]
+
+    def outer_step(*args):
+        # Separate the received arguments into their respective (seq, outputs
+        # from previous iterations, nonseqs) categories
+        i_sequences = list(args[:len(o_sequences)])
+        i_prev_outputs = list(args[len(o_sequences):-len(o_nonsequences)])
+        i_non_sequences = list(args[-len(o_nonsequences):])
+        
+        # Assemble the correct outputs_info list for the inner_scan
+        i_outputs_info = []
+
+        # Call the user-provided function with the proper arguments
+        results, updates = theano.scan(fn=fn,
+                                       sequences=i_sequences,
+                                       outputs_info=i_prev_outputs + [None,] * len(new_nitsots),
+                                       non_sequences=i_non_sequences[:-1],
+                                       name=name + "_inner",
+                                       n_steps=i_non_sequences[-1])
+        if not isinstance(results, list):
+            results = [results]
+
+        # Keep only the last timestep of every output but keep all the updates
+        if not isinstance(results, list):
+            return results[-1], updates
+        else:
+            return [r[-1] for r in results], updates       
+
+    results, updates = theano.scan(fn=outer_step,
+                                  sequences=o_sequences,
+                                  outputs_info=outputs_info,
+                                  non_sequences=o_nonsequences,
+                                  name=name + "_outer",
+                                  n_steps=o_n_steps, allow_gc=True)
+                                  
+    # Keep only the last timestep of every output but keep all the updates
+    return results, updates
+    if not isinstance(results, list):
+        return results[-1:], updates
+    else:
+        return [r[-1:] for r in results], updates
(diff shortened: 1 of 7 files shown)


theano/scan_module/scan_checkpoint.py
```diff
@@ -75,7 +75,7 @@ def outer_step(*args):
                                   n_steps=o_n_steps, allow_gc=True)
                                   
     # Keep only the last timestep of every output but keep all the updates
-    return results, updates
+    return results, updates # TODO is it a bug?
```
nouiz: I think we should remove this line. @carriepl do you remember why you added it? As for this to work well, we the user to only use the last step. So to prevent problems, I would think to not use the same interface as Scan. This would help prevent user from using all the states information (and so not get any memory saving).

@carriepl what do you think?

carriepl: @nouiz : I think the line was added so that using scan_with_checkpoints would be more like using standard scan (in the tests I did, I could just swap in one or the other and everything else would remain the same). Also, it made it possible to use scan_with_checkpoints to reduce memory usage when, for instance, you want every N-th output step (ex : steps 10, 20, 30, 40 and 50 out of 50 output steps). To be fair, I don't remember if this is something a user mentioned to me as being useful for his use case or this is just something I thought of myself so I don't know how used such a feature would be

Also, even if the user uses all that is inside 'results', he would still get some memory savings. 'results' at that line, only contains the N-th output steps (because the inner_step function only returns the last output steps).

Is my response clear enough or should we talk about this in person?

nouiz:
Just to be sure, the current code return only the saved states and not ask
of them? If so, I'm good with that solution. We just need to make sure it
is well documented.

nouiz:
IRL discussion with @carriepl, he confirmed that my interpretation was good.

theano/scan_module/scan_checkpoint.py
nouiz: IRL discussion with @carriepl, he confirmed that my interpretation was good.

So remove the TODO and the old code. Update the doc if needed.

nouiz:
There is a bunch of flake8 errors: https://travis-ci.org/Theano/Theano/jobs/161370814

and this error in travis:
```

ERROR: theano.scan_module.tests.test_scan_checkpoint.test_scan_checkpoint

  File "/home/travis/miniconda2/envs/pyenv/lib/python3.3/site-packages/nose-1.3.4-py3.3.egg/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/Theano/Theano/theano/scan_module/tests/test_scan_checkpoint.py", line 91, in test_scan_checkpoint

    example1(True)

  File "/home/travis/build/Theano/Theano/theano/scan_module/tests/test_scan_checkpoint.py", line 20, in example1

    save_every_N=20)

  File "/home/travis/build/Theano/Theano/theano/scan_module/scan_checkpoint.py", line 75, in scan_with_checkpoints

    n_steps=o_n_steps, allow_gc=True)

  File "/home/travis/build/Theano/Theano/theano/scan_module/scan.py", line 403, in scan

    'is %s' % n_steps.dtype)

nose.proxy.ValueError:  n_steps must be an int. dtype provided is float64

Example 1 without checkpoints

Compile time: 0.542370080947876

Exec time: 0.008044004440307617

Output: [  0.00000000e+00   1.00000000e+00   1.26765060e+30   5.15377521e+47

   1.60693804e+60   7.88860905e+69   6.53318624e+77   3.23447651e+84

   2.03703598e+90   2.65613989e+95]
```

nouiz:
Travis have a few errors:

```
/home/travis/build/Theano/Theano/theano/scan_module/scan_checkpoint.py:docstring of theano.scan_with_checkpoints:5: ERROR: Unexpected indentation.
```

```

ERROR: theano.scan_module.tests.test_scan_checkpoint.test_scan_checkpoint

  File "/home/travis/miniconda2/envs/pyenv/lib/python3.3/site-packages/nose-1.3.4-py3.3.egg/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/Theano/Theano/theano/scan_module/tests/test_scan_checkpoint.py", line 94, in test_scan_checkpoint

    example2(True)

  File "/home/travis/build/Theano/Theano/theano/scan_module/tests/test_scan_checkpoint.py", line 66, in example2

    save_every_N=10)

  File "/home/travis/build/Theano/Theano/theano/scan_module/scan_checkpoint.py", line 39, in scan_with_checkpoints

    s.ndim + 1) for s in sequences]

  File "/home/travis/build/Theano/Theano/theano/scan_module/scan_checkpoint.py", line 39, in <listcomp>

    s.ndim + 1) for s in sequences]

  File "/home/travis/build/Theano/Theano/theano/tensor/var.py", line 329, in reshape

    return theano.tensor.basic.reshape(self, shape, ndim=ndim)

  File "/home/travis/build/Theano/Theano/theano/tensor/basic.py", line 4682, in reshape

    rval = op(x, newshape)

  File "/home/travis/build/Theano/Theano/theano/gof/op.py", line 602, in __call__

  File "/home/travis/build/Theano/Theano/theano/tensor/basic.py", line 4503, in make_node

    raise TypeError("Shape must be integers", shp, shp.dtype)

nose.proxy.TypeError: ('Shape must be integers', MakeVector{dtype='float64'}.0, 'float64')

Example 1 without checkpoints

Compile time: 0.27387452125549316

Exec time: 0.0041658878326416016

Output: [  0.00000000e+00   1.00000000e+00   1.26765060e+30   5.15377521e+47

   1.60693804e+60   7.88860905e+69   6.53318624e+77   3.23447651e+84

   2.03703598e+90   2.65613989e+95]

Example 1 with checkpoints

Compile time: 0.9054667949676514

Exec time: 0.004945516586303711

Output: [  0.00000000e+00   1.00000000e+00   1.26765060e+30   5.15377521e+47

   1.60693804e+60   7.88860905e+69   6.53318624e+77   3.23447651e+84

   2.03703598e+90   2.65613989e+95]

----

Example 2 without checkpoints

Compile time: 0.7520089149475098

Exec time: 0.004889965057373047

Output: 4950

```

Thrandis:
@nouiz I haven't computed it! Right now it is on a GTX 750. I don't want to fill up the RAM of my computer and freeze it, so we need to think of a better solution.

How about only doing this test on the GPU, and fetch the memory size of the GPU so that we can scale the size of the input correctly? Would this work? Like if we have a 750 with 2 GB of RAM, we roughly know how much memory we need to allocate to blow it up. What do you think?

nouiz:
I like the idea to make this run on the GPU. I don't know if we have an interface to know the gpu memory available. If the test pass on a GPU with 2G and one with 24G, I would accept it.

Make sure the test use the new gpu back-en.

nouiz:
@abergeron is there a way to get the total memory on the gpu in the new back-end?

Thrandis:
Alright I'll look into it. For the new backend, I'll first need to rebase, but I'll try right after.

Thrandis:
I just performed a rebase so that I can run `test_memory` with the new GPU backend.

nouiz:
travis have an error: 

https://travis-ci.org/Theano/Theano/jobs/164770433

The test should pass or be skiped when there is no GPU.

theano/scan_module/tests/test_scan_checkpoint.py
```diff
@@ -0,0 +1,76 @@
...
+        assert numpy.allclose(out, out_check)
+
+    def test_backward_pass(self):
+        """Test gradient computation of A**k."""
+        f = theano.function(inputs=[self.A, self.k],
+                            outputs=[self.grad_A, self.grad_A_check])
+        out, out_check = f(range(10), 100)
+        assert numpy.allclose(out, out_check)
+
+    @unittest.skipUnless(theano.gpuarray.type._context_reg[None],
```
nouiz: [None] fail when there is no GPU. I think this should work:

None in theano.gpuarray.type.list_contexts()

theano/scan_module/tests/test_scan_checkpoint.py
```diff
@@ -0,0 +1,76 @@
...
+                         'Requires gpuarray backend.')
+    def test_memory(self):
+        """Test that scan_checkpoint reduces memory usage."""
+        k = T.iscalar("k")
+        A = T.vector("A")
+        result, updates = theano.scan(fn=lambda prior_result, A: prior_result * A,
+                                      outputs_info=T.ones_like(A),
+                                      non_sequences=A,
+                                      n_steps=k)
```
nouiz: I think you can reuse self.k, self.A and self.result

theano/scan_module/scan_checkpoint.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Current assumptions :
+    - Every sequence has the same length
+    - If n_steps is specified, it has the same value as the length of any sequence
+    - The value of "save_every_N" divides the number of steps the Scan will
+      run without remainder
+    - Only singly-recurrent and non-recurrent outputs are used.
+      No multiple recurrences.
+    - Only the last timestep of any output will ever be used.
+
```
Thrandis: done

theano/scan_module/scan_checkpoint.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    # Standardize the format of input arguments
+    if not isinstance(sequences, list):
+        sequences = [sequences]
+    if not isinstance(outputs_info, list):
+        outputs_info = [outputs_info]
```
Thrandis: done

theano/scan_module/scan_checkpoint.py
```diff
@@ -4,16 +4,67 @@
...
+        out the amount of steps it should run given its input sequences.
+
+    save_every_N
+        ``save_every_N`` is the number of steps to go without storing
+        the computations of scan (ie they will have to be recomputed
+        during the gradient computation).
+
+    See Also
+    --------
+    scan : Looping in Theano.
```
Thrandis: done

theano/scan_module/tests/test_scan_checkpoint.py
Thrandis: done

theano/scan_module/tests/test_scan_checkpoint.py
Thrandis: done

nouiz:
travis is failing: https://travis-ci.org/Theano/Theano/jobs/167459610

We don'T install pygpu on travis. The test should be skipeed when pygpu is not available, not crash.

theano/scan_module/scan_checkpoint.py
```diff
@@ -63,6 +63,21 @@ def scan_with_checkpoints(fn, sequences=[], outputs_info=None,
         the computations of scan (ie they will have to be recomputed
         during the gradient computation).
 
+    Returns
+    -------
+    tuple
+        Tuple of the form (outputs, updates); ``outputs`` is either a
```
nouiz: What about something simpler that don't duplicate the normals can doc like. Something like:

    Tuple of the form (outputs, updates) as `scan <LINK TO SCAN DOC ABOUT THAT>`_, but with a change. It only contain the output at each save_every_N step. If only the last outputs is used, the scan will get even more memory optimized then only using this method instead of scan.

I talked with Pierre Luc. As the user only get them, it is not necessary for him to only use the last output to same memory. But using only the last step will use even less memory.

theano/scan_module/tests/test_scan_checkpoint.py
```diff
@@ -6,25 +6,25 @@
 import theano
 import theano.tensor as T
 
+from pygpu.gpuarray import GpuArrayException
```
nouiz: I think moving the import in the test (with a comment to explain) would fix the problem in travis or doing a try except ImportError: pass around that import.

nouiz:
jenkins test this.

nouiz:
jenkins test this.

nouiz:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
add to whitelist

slefrancois:
jenkins test this

slefrancois:
jenkins test this

PR #4998, opened 2016-09-20
Some fixes for problems in float16 in the DLT

abergeron:
This fixes:
 - Wrong check in shape_i() that could lead to really long compile times (15h and more)
 - Slowness in NanGuardMode for the new backend
 - Use of two gpus with NanGuardMode and the new backend

It also marks GpuDnnSoftmax* as supporting float16 because that is the case.

### DIFF
--- a/theano/compile/nanguardmode.py
+++ b/theano/compile/nanguardmode.py
@@ -213,6 +213,7 @@ def f_gpua_min(inp):
     return f(inp)
 f_gpua_min.cache = dict()
 
+
 def f_gpua_max(inp):
     dt = inp.dtype
     ctx_name = _name_for_ctx(inp.context)
--- a/theano/compile/nanguardmode.py
+++ b/theano/compile/nanguardmode.py
@@ -274,7 +274,9 @@ class NanGuardMode(Mode):
             big_is_error = config.NanGuardMode.big_is_error
 
         assert nan_is_error or inf_is_error or big_is_error
-        compile_gpu_func(nan_is_error, inf_is_error, big_is_error)
+
+        if cuda.cuda_enabled:
+            compile_gpu_func(nan_is_error, inf_is_error, big_is_error)
 
         def do_check_on(value, nd, var=None):
             """
--- a/theano/compile/nanguardmode.py
+++ b/theano/compile/nanguardmode.py
@@ -11,6 +11,14 @@ import theano.tensor as T
 import theano.sandbox.cuda as cuda
 from theano.compile import Mode
 
+try:
+    from theano.gpuarray.type import GpuArrayType, _name_for_ctx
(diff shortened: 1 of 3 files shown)


theano/compile/nanguardmode.py
```diff
@@ -187,6 +200,47 @@ def compile_gpu_func(nan_is_error, inf_is_error, big_is_error):
...
 
+def f_gpua_min(inp):
+    dt = inp.dtype
+    ctx_name = _name_for_ctx(inp.context)
+    key = (dt, ctx_name)
+    f = f_gpua_min.cache.get(key, None)
+    if f is None:
+        guard_in = GpuArrayType(str(dt), (False,), context_name=ctx_name)()
+        f = theano.function([guard_in], T.min(guard_in),
+                            mode='FAST_RUN', profile=False)
```
nouiz: I think we should force the gpu, by adding the flag 'gpuarray' idem in the old back-end(optional).

Idem in the other function.

theano/compile/nanguardmode.py
```diff
@@ -187,6 +200,47 @@ def compile_gpu_func(nan_is_error, inf_is_error, big_is_error):
             cuda_compile_failed = True
 
 
+def f_gpua_min(inp):
+    dt = inp.dtype
```
nouiz: can you use the same var name as everywhere else in the code and use dtype instead of dt

theano/compile/nanguardmode.py
```diff
@@ -187,6 +200,47 @@ def compile_gpu_func(nan_is_error, inf_is_error, big_is_error):
...
+    ctx_name = _name_for_ctx(inp.context)
+    key = (dt, ctx_name)
+    f = f_gpua_min.cache.get(key, None)
+    if f is None:
+        guard_in = GpuArrayType(str(dt), (False,), context_name=ctx_name)()
+        f = theano.function([guard_in], T.max(T.abs_(guard_in)),
+                            mode='FAST_RUN', profile=False)
+        f_gpua_min.cache[key] = f
+    return f(inp)
+f_gpua_absmax.cache = dict()
```
nouiz: You duplicate the same code 3 times... If you could refactor it to have a builder function that you just pass a lambda that does the graph to execute and include caching, it would be great, but it is optional.

theano/gpuarray/dnn.py
```diff
@@ -1357,6 +1357,7 @@ class GpuDnnSoftmax(GpuDnnSoftmaxBase):
         or per spatial location '01' per image across 'c'.
 
     """
+    _f16_ok = True
```
nouiz: Add minimal tests.

nouiz:
included in https://github.com/Theano/Theano/pull/5049

PR #5001, opened 2016-09-21
Postpone Composite name creating

nouiz:
Fix some compilation getting killed.

I does this. The error was related to the elemwise fusion. I postpone some work from Composite.__init__ to later when we use it as all the stack trace we got when doing CTRL-C during the execution was there. This should speed up compilation, but it also fixed the crashes.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -3538,18 +3538,26 @@ class Composite(ScalarOp):
         Return a list of functions that compute each output of self.
 
         """
-        # the memo is to reuse functions during this call and traverse
-        # the graph only once. We will still have the problem during
-        # execution for the next comment. But this speed up linker
-        # phase where this method is called. Not so big problem as
-        # this code isn,t when performance is important.
+        # In the case where the graph is a dag, but not a tree like:
+        # add(*1 -> mul(x, y), *1)
+
+        # We have an efficent way to build the executable (we build
+        # and traverse each node only once).
+
+        # But we don't have an efficient execution. We will execute
+        # like a tree, so nodes that have more then 1 client will be
+        # executed as many times as there number of clients. In the
+        # example aboce, it will calculate *1 twice. Doing otherwise
+        # imply making a complicated execution engine.
+
+        # We need the fast creation of the executor as we always do it
+        # even if we will use the c code. The Python implementation is
+        # already slow, so it is not as much important to have a fast
+        # execution there.
+
         memo = {}
 
         def compose_impl(r):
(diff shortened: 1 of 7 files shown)


theano/tensor/elemwise.py
```diff
@@ -849,6 +849,14 @@ def prepare_node(self, node, storage_map, compute_map):
...
             sig = char * node.nin + '->' + char * node.nout
             node.tag.sig = sig
+        node.tag.fake_node = Apply(
+            self.scalar_op,
+            [get_scalar_type(dtype=input.type.dtype).make_variable()
+             for input in node.inputs],
+            [get_scalar_type(dtype=output.type.dtype).make_variable()
+             for output in node.outputs])
+
+        self.scalar_op.prepare_node(node.tag.fake_node, [], [])
```
abergeron: At least pass None for the fake storage_map and compute_map

theano/scalar/basic.py
```diff
@@ -3538,21 +3538,38 @@ def init_py_impls(self):
...
+        # the memo is to reuse functions during this call and traverse
+        # the graph only once. We will still have the problem during
+        # execution for the next comment. But this speed up linker
+        # phase where this method is called. Not so big problem as
+        # this code isn,t when performance is important.
+        memo = {}
+
         def compose_impl(r):
             # this is not optimal at all eg in add(*1 -> mul(x, y), *1)
             # it will calculate *1 twice
```
abergeron: This isn't true anymore

nouiz: This is True for execution. I'll try to clarify.

nouiz:
I updated the PR with more optimization and the fix from code review and CI failure.

theano/tensor/elemwise.py
```diff
@@ -849,6 +849,14 @@ def prepare_node(self, node, storage_map, compute_map):
             char = numpy.sctype2char(out_dtype)
             sig = char * node.nin + '->' + char * node.nout
             node.tag.sig = sig
+        node.tag.fake_node = Apply(
+            self.scalar_op,
+            [get_scalar_type(dtype=input.type.dtype).make_variable()
+             for input in node.inputs],
+            [get_scalar_type(dtype=output.type.dtype).make_variable()
+             for output in node.outputs])
```
abergeron: This should probably use make_node()

nouiz: This is a fake node. This is faster then calling make_node as it don't do many checks. I don't want to run the risk of slowing that down compared to the old code. I took it from line 1112 in the old version.

PR #5014, opened 2016-09-23
Fix crash in Lop

ynd:
Fix Lop to comply with new requirement that theano.grad's known_grad argument be an OrderedDict.


nouiz:
Can you add a test? It would make sure we don't break it again by mistake.
thanks

nouiz:
thanks.

ynd:
thanks, I'm in a rush: can't write this test right now.

PR #5025, opened 2016-09-27
Fix Theano deprecation warning

### LINKED ISSUE #4832: Theano deprecation warning
I have started running my code with warnings as errors.  Theano trips the deprecation warning:

```
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/configparser.py", line 1209, in __init__
    DeprecationWarning, stacklevel=2
DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
```

Please use:

```
theano_cfg =  (ConfigParser.ConfigParser if six.PY3 else ConfigParser.SafeConfigParser)(…
```

medakk:
fix #4832.

### DIFF
--- a/theano/configparser.py
+++ b/theano/configparser.py
@@ -10,7 +10,7 @@ import sys
 import warnings
 from functools import wraps
 
-from six import StringIO
+from six import StringIO, PY3
 
 import theano
 from theano.compat import configparser as ConfigParser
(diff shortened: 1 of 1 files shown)


nouiz:
thanks.

lamblin:
We should either:
- put that logic in `theano.compat`, since we seem to have defined something here, or
- remove `configparser` from `theano.compat` (after all, we are just importing it from `six` there) and import the right class directly.

It does not make sense to import either `configparser` or `ConfigParser` depending on the version of Python, and to then check _again_ the version of Python to see if we should use the `ConfigParser` or `SafeConfigParser` class.

medakk:
In Python2, the class used is `ConfigParser.SafeConfigParser`.
In Python3, the class used is `configparser.ConfigParser`.

The module name as well as the class is different in both cases. To put the logic in `theano.compat`, we'll have to modify the namespace of `ConfigParser`(the module) to substitute SafeConfigParser instead of ConfigParser(the class) and leave `configparser` alone. I think this is what is being currently done for urllib.

PR #5032, opened 2016-09-27
Speed up the canonizer for big list of num/denum

nouiz:


theano/gpuarray/opt.py
```diff
@@ -298,9 +300,7 @@ def apply(self, fgraph):
         target = infer_context_name(*fgraph.inputs)
         for i in fgraph.inputs:
             # Do not move *int* scalar to the GPU.
-            if (isinstance(i.type, tensor.TensorType) and
-                    (i.ndim > 0 or 'int' not in i.dtype) and
-                    "complex" not in i.dtype):
+            if isinstance(i.type, tensor.TensorType) and move_to_gpu(i):
```
Sentient07: I remember this was one thing that I want to do it in Graph2Gpu PR. Looks much cleaner. 👍

theano/gof/fg.py
```diff
@@ -260,29 +259,27 @@ def clients(self, r):
...
+    def __add_client__(self, r, new_client):
         """
         Updates the list of clients of r with new_clients.
 
         Parameters
         ----------
         r
             Variable.
-        new_clients
-            List of (node, i) pairs such that node.inputs[i] is r.
```
abergeron: pairs -> pair

theano/gof/fg.py
```diff
@@ -296,8 +293,8 @@ def __remove_clients__(self, r, clients_to_remove,
         ----------
         r : Variable
             The clients of r will be removed.
-        clients_to_remove : List of (op, i) pairs
-            List of (op, i) pairs such that node.inputs[i] is not r anymore.
```
abergeron: pairs -> pair

theano/gpuarray/opt.py
```diff
@@ -298,9 +300,7 @@ def apply(self, fgraph):
         target = infer_context_name(*fgraph.inputs)
         for i in fgraph.inputs:
             # Do not move *int* scalar to the GPU.
```
abergeron: Remove the comment as it is not relevant anymore

theano/gpuarray/type.py
```diff
@@ -572,6 +592,8 @@ def gpuarray_shared_constructor(value, name=None, strict=False,
     See :func:`theano.shared`.
 
     """
+    if target is None and not move_to_gpu(value):
+        raise TypeError('We do not move that data by deault to the GPU')
```
abergeron: deault -> default

Also, maybe make target another unique object (like NoContext), since None is a valid context name.

theano/gpuarray/type.py
```diff
@@ -572,6 +592,8 @@ def gpuarray_shared_constructor(value, name=None, strict=False,
     See :func:`theano.shared`.
 
     """
+    if target is None and not move_to_gpu(value):
```
abergeron: Perhaps move this check until after we know it's an array.  Otherwise it can crash move_to_gpu since the value might not have a dtype or ndim attribute.

theano/tensor/opt.py
```diff
@@ -4713,13 +4713,17 @@ def get_constant(v):
...
         """
-        if isinstance(v, Variable):
-            try:
-                # As the constant folding is in the canonicalize phase,
-                # We don't need to check all the graph each time.
-                return get_scalar_constant_value(v, only_process_constants=True)
-            except NotScalarConstantError:
+        if isinstance(v, Constant):
+            if getattr(v.tag, 'unique_value', None) is not None:
+                data = v.tag.unique_value
```
abergeron: What is this unique_value business?

PR #5042, opened 2016-09-30
speed up io_toposort when there is no ordering.

nouiz:
Self note, we see that io_toposort take much time in sb_resnet repo with 65 layers. Total compile time: 

    SeqOptimizer  OPT_FAST_RUN  time 3181.337s for 83756/31926 nodes before/after optimization

gpu inplace:

    1605.887107s - ('gpua_inplace_opt', 'InplaceElemwiseOptimizer', 43, 31926, 31926) - 379.235s

In that optimization, we spend 379/3181s, so 12% of total graph optimization time.

I'll use that to confirm this PR give speed up.

### DIFF
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -947,9 +947,11 @@ def general_toposort(outputs, deps, debug_print=False,
             rlist.append(node)
             rset.add(node)
             for client in _clients.get(node, []):
+                d = [a for a in deps_cache[client] if a is not node]
+                deps_cache[client] = d
                 deps_cache[client] = [a for a in deps_cache[client]
                                       if a is not node]
-                if not deps_cache[client]:
+                if not d:
                     sources.append(client)
 
     if len(rlist) != len(reachable):
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -981,8 +981,7 @@ def io_toposort(inputs, outputs, orderings=None, clients=None):
         node->clients for each node in the subgraph that is sorted
 
     """
-
-    if not orderings:  # can be None or empty dict
+    if not orderings and clients is None:  # ordering can be None or empty dict
         # Specialized function that is faster when more then ~10 nodes
         # when no ordering.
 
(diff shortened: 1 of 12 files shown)


theano/gof/graph.py
```diff
@@ -979,55 +981,50 @@ def io_toposort(inputs, outputs, orderings=None, clients=None):
...
+                continue
+            if all([i in computed or i.owner is None for i in cur.inputs]):
+                computed.update(cur.outputs)
+                order.append(cur)
             else:
-                assert not orderings.get(obj, [])
-            return rval
+                todo.append(cur)
+                todo.extend(i.owner for i in cur.inputs if i.owner)
+        return order
```
lamblin: It looks like this is changing a depth-first search for a breadth-first search.
I think it could have negative consequences for graphs with a different topology.

theano/gof/graph.py
```diff
@@ -979,55 +981,50 @@ def io_toposort(inputs, outputs, orderings=None, clients=None):
...
+            if cur.outputs[0] in computed:
+                continue
+            if all([i in computed or i.owner is None for i in cur.inputs]):
+                computed.update(cur.outputs)
+                order.append(cur)
             else:
-                assert not orderings.get(obj, [])
-            return rval
+                todo.append(cur)
+                todo.extend(i.owner for i in cur.inputs if i.owner)
```
lamblin: Also, why schedule `cur` first, before its inputs?

nouiz: It is a list. We pop at the end of the list. So in fact, we will see all the inputs before seeing that node again.

theano/gof/graph.py
nouiz: I think you have this impression as you mixed the order of execution being used. See the comment above.

theano/gof/graph.py
lamblin: Oh, OK, for some reason I thought it was a `deque` and we popped at the front.
In that case, wouldn't it be more efficient to just peek at the last element and only remove it if needed, rather than `pop` it every time and then `append` it back?

theano/gof/graph.py
lamblin: So, all we are doing is replacing a recursive algorithm by an iterative one using a list as LIFO (rather than the call stack)?
In that case, could we even get the same execution order by reversing the order of the inputs before appending them? For instance `todo.extend(i.owner for i in cur.inputs[::-1] if i.owner`

theano/tensor/elemwise_cgen.py
```diff
@@ -100,13 +100,13 @@ def make_checks(loop_orders, dtypes, sub):
             check += """
             if (%%(lv%(j0)s)s_n%(x0)s != %%(lv%(j)s)s_n%(x)s)
             {
-                PyErr_Format(PyExc_ValueError, "Input dimension mis-match. (input[%%%%i].shape[%%%%i] = %%%%i, input[%%%%i].shape[%%%%i] = %%%%i)",
+                PyErr_Format(PyExc_ValueError, "Input dimension mis-match. (input[%%%%i].shape[%%%%i] = %%%%li, input[%%%%i].shape[%%%%i] = %%%%li)",
                    %(j0)s,
                    %(x0)s,
-                   %%(lv%(j0)s)s_n%(x0)s,
+                   (long int) %%(lv%(j0)s)s_n%(x0)s,
```
abergeron: If you want to make sure to use 64-bits values use `long long`.  `long` is still 32-bits on windows.

abergeron:
This appears to change the ordering for debugprint:

```
File "tutorial/shape_info.txt", line 63, in default
Failed example:
    theano.printing.debugprint(f) # doctest: +NORMALIZE_WHITESPACE
Expected:
    MakeVector{dtype='int64'} [id A] ''   4
     |Elemwise{Add}[(0, 0)] [id B] ''   3
     | |Shape_i{0} [id C] ''   1
     | | |x [id D]
     | |Shape_i{0} [id E] ''   2
     |   |y [id F]
     |Shape_i{1} [id G] ''   0
       |x [id D]
Got:
    MakeVector{dtype='int64'} [id A] ''   4
     |Elemwise{Add}[(0, 0)] [id B] ''   3
     | |Shape_i{0} [id C] ''   2
     | | |x [id D]
     | |Shape_i{0} [id E] ''   1
     |   |y [id F]
     |Shape_i{1} [id G] ''   0
       |x [id D]
```

I don't mind the change, but it might be a pain to fix all the tests that rely on the current ordering.  Maybe @lamblin's fix will give us the previous order and avoid that.

theano/tensor/elemwise_cgen.py
nouiz: this is to print npy_intp

lamblin: On 64-bit Windows, `npy_intp` is 64-bit, but `long int` is 32-bit, so @abergeron is correct.

nouiz: done

theano/gof/graph.py
```diff
@@ -877,7 +879,7 @@ def clone_get_equiv(inputs, outputs, copy_inputs_and_orphans=True, memo=None):
     return memo
 
 
-def general_toposort(r_out, deps, debug_print=False,
+def general_toposort(outputs, deps, debug_print=False,
                      compute_deps_cache=None, deps_cache=None,
```
nouiz: finally, we still need them when the field clients isn,t empty.

theano/gof/opt.py
```diff
@@ -642,6 +642,8 @@ def process_node(self, fgraph, node):
                 continue
             if len(node.inputs) != len(candidate.inputs):
                 continue
+            if node.op != candidate.op:
+                continue
```
ReyhaneAskari: TODO: remove that from this PR. It must be speed tested separatly. Not sure this is good to do. comparing scan op is time consuming.

nouiz:
continued in https://github.com/Theano/Theano/pull/5502

PR #5049, opened 2016-10-03
Collection of fixes to make the DLT work in float16

abergeron:
also close gh-5053

### DIFF
--- a/theano/gpuarray/elemwise.py
+++ b/theano/gpuarray/elemwise.py
@@ -167,21 +167,19 @@ class GpuElemwise(HideC, Elemwise):
             """ % dict(n=n, name='"%s"' % (name,),
                        typecode=i.type.typecode)
 
-        p = 0
+        p = len(inps)
         for n, o in enumerate(node.outputs):
             if n in self.inplace_pattern:
                 assert(len(node.outputs) == 1)
                 res += "\nargs[%(n)s].flags |= GE_WRITE;\n" % dict(n=self.inplace_pattern[n])
             else:
-                nn = len(inps) + p
-                name = outs[p]
-                p += 1
                 res += """
                 args[%(n)s].name = %(name)s;
                 args[%(n)s].typecode = %(typecode)s;
                 args[%(n)s].flags = GE_WRITE;
-                """ % dict(n=nn, name='"%s"' % (name,),
+                """ % dict(n=p, name='"%s"' % (outs[n],),
                            typecode=o.type.typecode)
+                p += 1
 
         res += """
         ge = GpuElemwise_new(%(ctx)s->ctx, %(support)s, %(kop)s, %(nargs)s, args, %(nd)s, GE_CONVERT_F16);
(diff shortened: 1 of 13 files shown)


nouiz:
Check my comment here: https://github.com/Theano/Theano/pull/4998

nouiz:
travis and jenkins fail.

abergeron:
The problem should be fixed now.

theano/gpuarray/elemwise.py
```diff
@@ -202,7 +184,7 @@ def c_init_code_struct(self, node, name, sub):
                            typecode=o.type.typecode)
```
nouiz: line 177 should probably be changed too. now outs have the same number of element then node.outputs. The variable p is probably not needed anymore.

theano/scalar/basic.py
```diff
@@ -2299,8 +2299,8 @@ def c_code___(self, node, name, inputs, outputs, sub):
         (x,) = inputs
         (z,) = outputs
         typ = node.outputs[0].type.dtype
-        if typ not in ['float32', 'float64']:
-            Exception("The output should be float32 or float64")
+        if not typ.startswith('float'):
```
abergeron: I don't want to make scalar/basic.py depend on tensor/basic.py.  Also I guess that part can be reverted since I fixed the not-float16 problems another way.

theano/tensor/nnet/sigm.py
```diff
@@ -352,7 +352,7 @@ def c_code(self, node, name, inp, out, sub):
 
         # float16 limits: -17.0, 6.0
         # We use the float32 limits for float16 for now as the
-        # computation will happend in float32 anyway.
+        # computation will happen in float32 anyway.
         if (node.inputs[0].type == scalar.float32 or
                 node.inputs[0].type == scalar.float16):
             return """%(z)s = %(x)s < -103.0f ? 0.0 : %(x)s > 14.0f ? %(x)s : log1p(exp(%(x)s));""" % locals()
```
abergeron: The GPU also uses this C code.  But it shouldn't be required after my changes in this PR and the next to convert the internal graph of Composite to float32 in GpuElemwise.

theano/gpuarray/elemwise.py
abergeron: This area does need some changes, but p is still required to keep track of which args we are at.

theano/gradient.py
```diff
@@ -1373,10 +1373,10 @@ class numeric_grad(object):
     # perfectly accurate.
     type_eps = {'float64': 1e-7,
                 'float32': 3e-4,
-                'float16': 1e-3,
+                'float16': 1e-1,
```
lamblin: `eps` is the step size, not the tolerance.
I'm OK with those values.

theano/scalar/basic.py
abergeron: This one is not strings, but Scalar() instances.

lamblin: So is `node.outputs[0].type`.

theano/gpuarray/elemwise.py
nouiz: You changed the definition of p. I'm good with the fix.

PR #5052, opened 2016-10-04
5.3x compile speedup for sb_resnet @ depth 200

xoltar:
Combination of a few small changes - fewer validations during elemwise inplace optimization (still overridable by the existing config setting, of course), a few data structure changes, etc.

Also fixed a few typos.


theano/gof/opt.py
```diff
@@ -2390,10 +2389,7 @@ def importer(node):
 
             def pruner(node):
                 if node is not current_node:
-                    try:
-                        q.remove(node)
-                    except ValueError:
-                        pass
+                    node.disavowed = True
```
nouiz: Great finding. This could explain some slowness I saw in the profile.

If some optimization re-introduce some nodes that where removed, then this won't work. I see a few possible fixes:

1) In the function importer, set node.disavowed to False or remove it.
2) Don't use a pruner. Instead do "if node not in fgraph.apply_nodes: continue" on line 2404. This would have the same algorithmic complexity. I'm not sure which one would have the lowest overhead. fgraph.apply_nodes is a set.
3) modify `FunctionGraph.__remove_client__` to do "apply_node.fgraph = None" when we remove the nodes from self.apply_nodes (around line 334 in my current checkout). Then on line 2404, we just do "if not hasattr(node, 'fgraph'): continue. This solution could be better in the long term, but it also could be not possible du to the nodes that get reintroduced sometimes. Or it could request other changes before being possible.

So the safest route is between 1 and 2. Can you time them? If they are the same, I prefer 2, as it don't let old information to the nodes after the optimizations.

theano/tensor/opt.py
```diff
@@ -189,10 +189,8 @@ def inplace_elemwise_optimizer(fgraph):
         # We execute `validate` after this number of change.
         check_each_change = config.tensor.insert_inplace_optimizer_validate_nb
         if check_each_change == -1:
-            if len(fgraph.apply_nodes) > 500:
-                check_each_change = 10
-            else:
-                check_each_change = 1
+            #By default, check 10 times during the course of the optimization.
+            check_each_change = max(int(len(fgraph.apply_nodes)/10), 1)
```
nouiz: This one need more investigation.

theano/gof/destroyhandler.py
```diff
@@ -210,13 +210,12 @@ def _build_droot_impact(destroy_handler):
             # The code here add all the variables that are views of r into
             # an OrderedSet input_impact
             input_impact = OrderedSet()
-            queue = Queue()
-            queue.put(input_root)
-            while not queue.empty():
-                v = queue.get()
+            queue = [input_root]
```
xoltar: @nouiz I'll try both and keep whichever is faster

theano/gof/opt.py
```diff
@@ -2145,15 +2144,15 @@ def importer(node):
 
         def pruner(node):
             if node is not current_node and node.op == op:
-                try:
-                    q.remove(node)
-                except ValueError:
-                    pass
+                node.disavowed = True
```
xoltar: I made the same change to all pruners, I'll try deleting this one if it turns out to be unused.

theano/gof/opt.py
xoltar: OK, will try timing 1 and 2. Hopefully 2 will work better.

xoltar:
@nouiz Thanks for your comments, they were very helpful. I'll close this for now and break it into smaller (rebased) PRs that we can merge individually after any needed discussion or changes.

PR #5054, opened 2016-10-05
Remove ProfileMode

nouiz:
There is 2 things left to port. I let this in a new issue:
https://github.com/Theano/Theano/issues/5055


doc/library/compile/profilemode.txt
```diff
@@ -1,3 +1,4 @@
+:orphan:
```
abergeron: Can we just remove the file instead?

theano/compile/tests/test_modes.py
```diff
@@ -43,20 +40,14 @@ def test1(self):
         # there should be
         # - VM_Linker
         # - OpWiseCLinker (FAST_RUN)
-        # - WrapLinker ("ProfileMode")
+        # - WrapLinker
         # - PerformLinker (FAST_COMPILE)
         # - DebugMode's Linker  (DEBUG_MODE)
```
abergeron: The comment is now confusing since there are 5 items named, but we assert the list contains 4 elements.  You should probably just remove the line about WrapLinker.

theano/compile/tests/test_modes.py
```diff
@@ -43,20 +40,14 @@ def test1(self):
...
 
 
-class T_ProfileMode_WrapLinker(unittest.TestCase):
+class T_WrapLinker(unittest.TestCase):
     def test_1(self):
-        # First, compile a function with a new ProfileMode() object
-        # No need to call that function
-        x = T.matrix()
-        mode = ProfileMode()
-        theano.function([x], x * 2, mode=mode)
```
abergeron: I suspect that this test is not useful anymore without this part.  If that is not the case, then it is at least badly misnamed since at no point does it involve WrapLinker.

theano/configdefaults.py
```diff
@@ -463,17 +463,15 @@ def warn_cxx(val):
 if rc == 0 and config.cxx != "":
     # Keep the default linker the same as the one for the mode FAST_RUN
     AddConfigVar('linker',
-                 ("Default linker used if the theano flags mode is Mode "
-                  "or ProfileMode(deprecated)"),
+                 "Default linker used if the theano flags mode is Mode ",
```
abergeron: Maybe you want to remove the final whitespace in the string.

theano/configdefaults.py
```diff
@@ -463,17 +463,15 @@ def warn_cxx(val):
...
                  EnumStr('cvm', 'c|py', 'py', 'c', 'c|py_nogc',
                          'vm', 'vm_nogc', 'cvm_nogc'),
                  in_c_key=False)
 else:
     # g++ is not present or the user disabled it,
     # linker should default to python only.
     AddConfigVar('linker',
-                 ("Default linker used if the theano flags mode is Mode "
-                  "or ProfileMode(deprecated)"),
+                 "Default linker used if the theano flags mode is Mode ",
```
abergeron: Here too.

theano/configdefaults.py
```diff
@@ -501,8 +499,7 @@ def warn_cxx(val):
 # Keep the default optimizer the same as the one for the mode FAST_RUN
 AddConfigVar(
     'optimizer',
-    ("Default optimizer. If not None, will use this linker with the Mode "
-     "object (not ProfileMode(deprecated) or DebugMode)"),
+    "Default optimizer. If not None, will use this optimizer with the Mode ",
```
abergeron: And here.

abergeron:
Overall I greatly approve!

doc/library/compile/profilemode.txt
nouiz: No, for old url...

theano/compile/tests/test_modes.py
nouiz: done

theano/compile/tests/test_modes.py
nouiz: I renamed the test

theano/configdefaults.py
nouiz: done

theano/configdefaults.py
nouiz: done

theano/configdefaults.py
nouiz: done

nouiz:
I fixed all comments.

PR #5060, opened 2016-10-06
GpuAdvancedIncSubtensor1 supports mixed dtypes.

abergeron:
The dev20 version doesn't but that's accessory.

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -999,7 +999,7 @@ def local_gpua_advanced_incsubtensor(op, context_name, inputs, outputs):
     compute_capability = int(context.bin_id[-2])
 
     if (compute_capability < 2 or x.ndim != 2 or y.ndim != 2 or
-        x.type.dtype != y.type.dtype):
+            x.type.dtype != y.type.dtype):
         return GpuAdvancedIncSubtensor1(
             set_instead_of_inc=set_instead_of_inc)
     else:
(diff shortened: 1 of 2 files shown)


nouiz:
jenkins fail.

theano/gpuarray/opt.py
```diff
@@ -994,19 +994,12 @@ def local_gpua_advanced_incsubtensor(op, context_name, inputs, outputs):
...
-            y = tensor.cast(y, dtype)
-
     set_instead_of_inc = op.set_instead_of_inc
 
     compute_capability = int(context.bin_id[-2])
 
-    if (compute_capability < 2 or x.ndim != 2 or y.ndim != 2):
+    if (compute_capability < 2 or x.ndim != 2 or y.ndim != 2 or
+        x.type.dtype != y.type.dtype):
         return GpuAdvancedIncSubtensor1(
```
nouiz: what this PR fix? The cast should have worked I think. From my understanding, this PR should fix a crash and I don't understand what it fix.

nouiz:
/home/jenkins-cpu1/workspace/Theano PR multijob/theano/gpuarray/opt.py:1002:9: E129 visually indented line with same indent as next logical line

theano/gpuarray/opt.py
abergeron: The cast works, but is unused.  The optimization just returns the op, not a full node.  I figure this was done during the move to GraphToGpu and since we don't actually need the case for the non-dev20 case, I just removed it.

nouiz: right

Issue #4317, opened 2016-03-30
Remove Op._op_use_c_code attribute


nouiz:
We should use the flag theano.config.cxx everywhere.

ryan5rdx:
@nouiz I'd like to tackle this if possible, it seems like something good for me to start off with.

nouiz:
Nobody is working on this, so don't hesitate to take it.

thanks

lamblin:
@Faruk-Ahmed Is now assigned to this tickets.

Remaining questions for @nouiz or @abergeron: what should we replace it with at the following places:
- https://github.com/Theano/Theano/blob/539ca7eb4a4e4661055c115860a60891172efe30/theano/gof/vm.py#L1046-L1051
- https://github.com/Theano/Theano/blob/cd21e2d074413e76d04e8ac996c3777d55703a2d/theano/tensor/opt.py#L6232-L6242
- (there may be more)

Faruk-Ahmed:
These are the places at which the flag seems to occur:

```
ahmedfar@eos3:~/Projects/Theano$ grep -rnw ./ -e "op._op_use_c_code"
./theano/tensor/opt.py:6236:            node.op._op_use_c_code = False
./theano/tensor/opt.py:6242:            node.op._op_use_c_code = old_value
./theano/gof/cc.py:1795:                        node.op._op_use_c_code = True
./theano/gof/cc.py:1804:                    node.op._op_use_c_code = old_value
./theano/gof/link.py:828:                node.op._op_use_c_code = False
./theano/gof/link.py:836:                node.op._op_use_c_code = old_value
./theano/gof/vm.py:1047:                    node.op._op_use_c_code = False
```

PR #4959, opened 2016-09-09
Removing _op_use_c_code attribute

Faruk-Ahmed:
reference #4317 

This is one fix in one place. Others are pending, I'm looking for feedback right now if what I'm doing looks right.

PR #5063, opened 2016-10-07
Removing _op_use_c_code attribute

nouiz:
fix gh-4317

see also PR: gh-4959

THis is a continuation of that PR, but it seem we diverged.


nouiz:
Closing, the other PR is based on the same branch as this one and have more fix.

Issue #5064, opened 2016-10-07
check_equal_numpy returns True for unequal ndarrays


medakk:
The function `check_equal_numpy` defined in [tensor/basic.py](https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L66) returns True when given unequal ndarrays. To be precise, the function (wrongly) returns True if all of the following are true:

1. Both the arguments `x` and `y` are instances of `ndarray`
2. Both the arguments `x` and `y` have equal `shape` and `dtype`
3. If **any** element in `abs(x-y)` is less than 1e-10

I believe it should be **all** elements, not **any**.

Here is a REPL session demonstrating this issue:
```
>>> import theano
>>> import numpy as np
>>> A = np.array([1,2,3,4,5])
>>> B = np.array([9,8,7,6,5])
>>> theano.tensor.check_equal_numpy(A, B)
True
```

The fix would be to change `numpy.any` to `numpy.all` in that function.

nouiz:
thanks. Do you want to make a PR to fix that? That way you will have your
nae in the history of Theano.

Otherwise, I can do it.

medakk:
I'll make a PR.

PR #5065, opened 2016-10-07
Change numpy.any to numpy.all in check_equal_numpy

medakk:
Fix gh-5064


lamblin:
Jenkins test this please.

medakk:
@lamblin This PR has passed the Jenkins CI test
Or are you asking me to write a test for this function?

lamblin:
I think the CI checks are enough for now.
Ultimately, I think this function may not be called often, and is probably redundant with other ones that we actually use more, so we may want to get rid of it in the future.
Thanks for the fix!

PR #5066, opened 2016-10-08
[WIP] Make the MergeOptimizer execution time independent of the number of clients

nouiz:
Not ready to merge. But I think it work. Need code clean up and understand why sometimes we aren't able to remove a key from self.d.

@lamblin this is what I talked about this afternoon.


nouiz:
Closing, the work around is so good that this isn't needed anymore.

PR #5067, opened 2016-10-09
Icc support

HuiLi-Intel:
We just duplicate from GCC simply to support ICC. Next we will continue to complete it.


HuiLi-Intel:
@nouiz Do you have any comments?

theano/configdefaults.py
```diff
@@ -1647,7 +1649,7 @@ def short_platform(r=None, p=None):
 # Allow to have easily one compiledir per device.
 compiledir_format_dict['device'] = config.device
 compiledir_format_keys = ", ".join(sorted(compiledir_format_dict.keys()))
-default_compiledir_format = ("compiledir_%(short_platform)s-%(processor)s-"
+default_compiledir_format = ("compiledir_%(hostname)s-%(short_platform)s-%(processor)s-"
```
nouiz: I don't think it is a good idea. This mean that by default if your home is on nfs and you use Theano on multiple computers that are exactly the same, the cache won't be shared between those computers. This is what we did up to now.

What do you wanted to do with that change?

theano/configdefaults.py
```diff
@@ -443,7 +443,7 @@ def f(suffix=suffix):
 if param and os.name == 'nt':
     param = '"%s"' % param
 AddConfigVar('cxx',
-             "The C++ compiler to use. Currently only g++ is"
+             "The C++ compiler to use. Currently only g++ and icpc is"
```
nouiz: Good change. We also support clang++ as it support the same interface as g++. Can you change it to this at the same time
"only g++ (clang++) and icpc"

theano/configdefaults.py
```diff
@@ -1116,6 +1116,18 @@ def is_valid_check_preallocated_output_param(param):
              # Added elsewhere in the c key only when needed.
              in_c_key=False)
 
+if not config.cxx == "gcc" and config.gcc.cxxflags:
```
nouiz: clang++ is also suported. I think it should be:

    if config.cxx == "icpc" and config.gcc.cxxflags:

theano/configdefaults.py
```diff
@@ -1116,6 +1116,18 @@ def is_valid_check_preallocated_output_param(param):
...
+    _logger.warning("cxx is not set to g++. gcc.cxxflags is not effective!")
+
+AddConfigVar('icc.cxxflags',
+             "Extra compiler flags for icc",
+             StrParam(""),
+             # Added elsewhere in the c key only when needed.
+             in_c_key=False)
+
+if not config.cxx == "icpc" and config.icc.cxxflags:
+    _logger.warning("cxx is not set to icpc. gcc.cxxflags is not effective!")
```
nouiz: Instead of having those 2 mutually exclusives flags, why not rename gcc.cxxflags to:

cxx.flags?

This will need some warning and automatic conversion of the old flags to the new name.

What do you think of that interface?

theano/configdefaults.py
```diff
@@ -1524,11 +1536,24 @@ def _timeout_default():
...
     # Typically means gcc cannot be found.
-    gcc_version_str = 'GCC_NOT_FOUND'
+    gcc_version_str = 'CXX_NOT_FOUND'
 
+try:
+    p_out = output_subprocess_Popen([config.cxx, '-dumpversion'])
+    icc_version_str = p_out[0].strip().decode()
+except OSError:
+    # Typically means gcc cannot be found.
+    icc_version_str = 'CXX_NOT_FOUND'
```
nouiz: This is the same code as for cxx_version_str. What about having only one cxx_version_str and remove icc_version_str and gcc_version_str and prepends to it config.cxx?

It would change the semantic to not only a version number. We could also do that without the prepend. What do you think? @lamblin @abergeron?

theano/configdefaults.py
```diff
@@ -1524,11 +1536,24 @@ def _timeout_default():
 
 try:
     p_out = output_subprocess_Popen([config.cxx, '-dumpversion'])
+    cxx_version_str = p_out[0].strip().decode()
+except OSError:
+    # Typically means gcc cannot be found.
+    cxx_version_str = 'CXX_NOT_FOUND'
+
+try:
+    p_out = output_subprocess_Popen(['g++', '-dumpversion'])
```
nouiz: if config.cxx is a full path to a different g++ version then the default one, this won't grap the right version.

theano/gof/cmodule.py
```diff
@@ -1565,6 +1565,15 @@ def get_gcc_shared_library_arg():
     else:
         return '-shared'
 
+def get_icc_shared_library_arg():
+    """
+    Return the platform-dependent GCC argument for shared libraries.
```
nouiz: As this is a copy-paste of the gcc version, why not rename the gcc version to get_shared_library_arg? Do you expect that to change in the futur? If you think it is better to keep it separated, update the docstring.

theano/gof/cmodule.py
```diff
@@ -2310,5 +2319,314 @@ def print_command_line_error():
...
+        if py_module:
+            # touch the __init__ file
+            open(os.path.join(location, "__init__.py"), 'w').close()
+            assert os.path.isfile(lib_filename)
+            return dlimport(lib_filename)
+
+if config.cxx == "icpc":
+    CXX_compiler = ICC_compiler
+else:
+    CXX_compiler = GCC_compiler
```
nouiz: no extra spaces at the end of the line and add a new line.

theano/gof/op.py
```diff
@@ -1168,11 +1168,19 @@ def test_gxx_support():
...
+                src_code=code,
+                tmp_prefix='test_omp_',
+                flags=['-qopenmp'],
+                try_run=False)
+        else:
+            default_openmp = CXX_compiler.try_compile_tmp(
+                src_code=code,
+                tmp_prefix='test_omp_',
+                flags=['-fopenmp'],
+                try_run=False)
```
nouiz: The only difference is the flags parameter. Can you just set a temp variable flags=[...] in both branch of the code and call try_compile_tmp() once after the if. This make the code more readable.

theano/gof/cmodule.py
```diff
@@ -2310,5 +2319,314 @@ def print_command_line_error():
...
+    @staticmethod
+    def version_str():
+        icc_ver = [int(n) for n in icc_version_str.split('.')[:2]]
+        if bool(icc_ver < [16, 0]):
+            _logger.warn(
+                "WARNING: Your Intel Compiler Version is %s. "
+                "For better performance, we suggest you use the latest version. " % icc_version_str)
+            _logger.info(
+                "WARNING: Your Intel Compiler Version is %s. "
+                "For better performance, we suggest you use the latest version. " % icc_version_str)
```
nouiz: don't log twice. I think the warn level is good. It should appear by default like that.

theano/gof/cmodule.py
```diff
@@ -2310,5 +2319,314 @@ def print_command_line_error():
...
+        # TODO
+        if ('icpc' not in theano.config.cxx):
+            _logger.warn(
+                "OPTIMIZATION WARNING: your Theano flag `cxx` seems not to be"
+                " the icpc compiler. So we disable the compiler optimization"
+                " specific to icpc that tell to compile for a specific CPU."
+                " At worst, this could cause slow down.\n"
+                "         You can add those parameters to the compiler yourself"
+                " via the Theano flag `gcc.cxxflags`."
+            )
```
nouiz: I think we should never call this method if cxx don't have icpc. If you agree, change this to an exception.

theano/gof/cmodule.py
```diff
@@ -2310,5 +2319,314 @@ def print_command_line_error():
...
+        # ARM (32-bit and 64-bit) architectures in order to make
+        # Theano compatible with the Raspberry Pi, Raspberry Pi 2, or
+        # other systems with ARM processors.
+        """
+        if (not any(['arm' in flag for flag in cxxflags]) and
+                not any(arch in platform.machine() for arch in ['arm', 'aarch'])):
+            n_bits = local_bitwidth()
+            cxxflags.append('-m%d' % n_bits)
+            _logger.debug("Compiling for %s bit architecture", n_bits)
+        """
```
nouiz: If icpc don't support arm, just remove all that code and the corresponding comment.

theano/gof/cmodule.py
```diff
@@ -2310,5 +2319,314 @@ def print_command_line_error():
...
+                        try_run=False, output=False):
+        return Compiler._try_compile_tmp(src_code, tmp_prefix, flags,
+                                         try_run, output,
+                                         theano.config.cxx)
+
+    @staticmethod
+    def try_flags(flag_list, preambule="", body="",
+                  try_run=False, output=False):
+        return Compiler._try_flags(flag_list, preambule, body, try_run, output,
+                                   theano.config.cxx)
```
nouiz: I don't have time now to look in detail, but why those 2 statics methods are here and not in the parent class?

I think we refactored that recently.

theano/gof/cmodule.py
```diff
@@ -2310,5 +2319,314 @@ def print_command_line_error():
...
+
+    @staticmethod
+    def compile_str(module_name, src_code, location=None,
```
nouiz: What is the difference in that method compared to the GCC version? It seem that it have much duplicated code. Could we refactor to limit that duplication?

theano/configdefaults.py
abergeron: I would be ok with a single cxx_version_str.  Otherwise, it would be kind of crazy to have a plethora of version strings for every different type of compiler.

theano/configdefaults.py
```diff
@@ -1125,6 +1125,36 @@ def is_valid_check_preallocated_output_param(param):
              # Added elsewhere in the c key only when needed.
              in_c_key=False)
 
+if config.cxx == "icpc" and config.gcc.cxxflags:
```
abergeron: Take care with these checks since `config.cxx` is the full path to the compiler not just the executable name.  It's probably better to check for `'icpc' in config.cxx`

theano/gof/cmodule.py
abergeron: I agree with that, much of the code for ICC_Compiler seems like a copy and paste from GCC_Compiler.  This is bad because it will be hard to maintain in the future and ICC_Compiler is much more likely to fall behind since there are less people using it.

There should be a refactor of the compiler interface so that multiple implementation can share as much code as possible (things like default definitions, compilation i/o, ...).

theano/configdefaults.py
HuiLi-Intel: Thank you.

theano/configdefaults.py
HuiLi-Intel: Sure

theano/configdefaults.py
HuiLi-Intel: I just want to keep the application compatibility. If I change the gcc.cxxflags to cxx.flags and user don't know the change. The original config can't work.

theano/configdefaults.py
HuiLi-Intel: Sorry for the mistake.

theano/configdefaults.py
HuiLi-Intel: Hi, @abergeron 
This seems a little crazy. 
ICC is a commercial compiler. It is a lot of different flag compared with gcc. And ICC support some features that gcc can't support good enough. Such as, vectorization, new cpu instruction and so on. And ICC also can't support arm and other CPU. These make the difference between ICC and g++ don't like the difference between g++ and clang++. And future, we will add some special optimized code to theano. These code maybe work well with gcc and other compiler. We need some more check and flag to make sure these code can work well with all compiler.
Thank you.

theano/configdefaults.py
HuiLi-Intel: @abergeron Thank you. I will remove it.

theano/gof/cmodule.py
HuiLi-Intel: If you don't think use a ICC_compiler is good idea. I can merge it to the GCC_compiler.

HuiLi-Intel:
Hi, @abergeron @nouiz 
This seems a little crazy. 
ICC is a commercial compiler. It is a lot of different flag compared with gcc. And ICC support some features that gcc can't support good enough. Such as, vectorization, new cpu instruction and so on. And ICC also can't support arm and other CPU. These make the difference between ICC and g++ don't like the difference between g++ and clang++. And future, we will add some special optimized code to theano. These code maybe work well with gcc and other compiler. We need some more check and flag to make sure these code can work well with all compiler.
This version is only init the support. And it is simple paste from GCC. But we will continue maintain it with another feature submit.
If you don't think a separate class is a good idea, I can also merge these changes to GCC_compiler.
P.S.: 
Although ICC is a commercial compiler. It is free for non-commercial use at least one year.
Thank you.

lamblin:
Hi @HuiLi-Intel,
I think having a separate ICC_compiler class is good, as icpc and g++ are indeed different in the options they support.
And starting by a copy of the GCC_compiler class is OK as a first step of a work in progress, for testing purposes and to see which problems arise.
However, before the PR is merged, we want to avoid code duplication if possible. I think this can be achieved by refactoring the existing GCC_compiler class, by having a base compiler class, say CXX_compiler, that both ICC_compiler and the new version of GCC_compiler would inherit from. That way, the shared code would belong in the base class, and only the specialized code would be in the derived classes. This may mean changing the interface a bit and cutting existing functions into smaller pieces that can be more easily re-used.
Does that make sense to you?

abergeron:
I agree about the separate class. It is a very good idea to keep it.  I was just referring to the refactor that @lamblin explained.

theano/configdefaults.py
abergeron: I think we should have a setting for cxx, and another for cxx_path so that the first one can be the compiler "class" and the other one the binary path.  So you could have `config.cxx = 'g++'` and `config.cxx_path = '/usr/bin/clang++'`, for example.

This will avoid mistakes in checks like these.

theano/configdefaults.py
nouiz: @abergeron just desribed part of the problem. Also, normally, the march flags are added in the key of each compiled module. So if the difference is only there, then it should not be a problem. There is a few exception about that. This would need refactoring for a full complete fix.

So, if you continue to get this problem, tell us more about the exact case you have. There is other fix needed than this one and we can discuss them to find the right fix.

nouiz:
Having 2 config can make thing more complicated and not clear I think. What
about the user only setting cxx_path to icpc, but not changing cxx flag? Or
changing both to not compatible combination.

To prevent user error (more frequent then change to this code) I think we
should keep only 1 flag.

nouiz:
Closing this PR as no update in a long time and it need much work.

Don't hesitate to open a new  PR when you have advanced on this.

PR #5070, opened 2016-10-09
fix softmaxgrad dnn opt error fixes #5056

### LINKED ISSUE #5056: ERROR (theano.gof.opt): Optimization failure due to: local_softmax_dnn_grad
I have a cost function that looks like:
```python
cost = T.nnet.softmax(matrix.flatten()).sum()
```
where ```matrix``` is the result of some other computation and during optimization I get
```python
ERROR (theano.gof.opt): Optimization failure due to: local_softmax_dnn_grad
ERROR (theano.gof.opt): node: SoftmaxGrad(TensorConstant{(1, 65536) of 1.0}, HostFromGpu.0)
  File "/home/jesse/Development/Theano/theano/gof/opt.py", line 1922, in process_node
    replacements = lopt.transform(node)
  File "/home/jesse/Development/Theano/theano/sandbox/cuda/dnn.py", line 3113, in local_softmax_dnn_grad
    if isinstance(n.owner.op, HostFromGpu):
AttributeError: 'NoneType' object has no attribute 'op'
```

Works fine if I switch to
```python
cost = T.sqr(matrix.flatten()).sum()
```
or
```python
matrix_f = matrix.flatten()
cost = (T.nnet.softmax(matrix_f)*matrix_f).sum()
```

JesseLivezey:
@lamblin, I checked on my machine that the new test fails with master and all tests in test_dnn.py pass with fix.

fix gh-5056

### DIFF
--- a/theano/sandbox/cuda/tests/test_dnn.py
+++ b/theano/sandbox/cuda/tests/test_dnn.py
@@ -591,7 +591,7 @@ class test_DnnSoftMax(test_nnet.test_SoftMax):
         xp = x.reshape((5, 5))
         y = T.nnet.softmax(xp.flatten()).sum()
         g = T.grad(y, x)
-        f = theano.function(inputs=[x], outputs=g, mode=self.mode)
+        theano.function(inputs=[x], outputs=g, mode=self.mode)
 
     def test_cudnn_softmax_grad_opt(self):
         # Verify that the SoftmaxGrad -> GpuDnnSoftmaxGrad optimization is
--- a/theano/sandbox/cuda/tests/test_dnn.py
+++ b/theano/sandbox/cuda/tests/test_dnn.py
@@ -582,6 +582,17 @@ class test_DnnSoftMax(test_nnet.test_SoftMax):
         if cuda.dnn.version() != (3000, 3000):
             utt.verify_grad(softmax_op, [x_val2], mode=mode_with_gpu)
 
+    def test_local_softmax_dnn_grad(self):
+        """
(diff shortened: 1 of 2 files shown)


JesseLivezey:
Based on my understanding of the current optimization, it seemed like the correct thing to do if ```n.owner``` is ```None``` is to skip the step that tries to remove a ```HostFromGpu``` and just continue to the next check.

theano/sandbox/cuda/tests/test_dnn.py
```diff
@@ -582,6 +582,17 @@ def test_dnn_softmax_grad(self):
...
+        """
+        Check for optimization error when grad of summed
+        softmax is taken over tensor with fixed shape.
+        """
+        x = T.fvector('x')
+        xp = x.reshape((5, 5))
+        y = T.nnet.softmax(xp.flatten()).sum()
+        g = T.grad(y, x)
+        theano.function(inputs=[x], outputs=g, mode=self.mode)
+
```
nouiz: Can you add an assert that the optimized function contain the good op. Something like:

assert(any(n for n in f.maker.fgraph.toposort() if isinstance(n.op, GpuDnnSoftmaxGrad))

Otherwise, the PR seem good. thanks.

nouiz:
jenkins test this

nouiz:
We can merge when jenkins pass

JesseLivezey:
jenkins passed, do I need to rebase and squash?

nouiz:
No. Merging. thanks

PR #5071, opened 2016-10-10
Frozendicts

Sentient07:
Replacing the mutable `dict` variables of `Op` with an immutable type and removing the `hash_from_dict` implementation


medakk:
In `elemwise.py`, line 487, this

```
self.inplace_pattern = frozendict(inplace_pattern)
self.destroy_map = dict((o, [i]) for o, i in frozendict(inplace_pattern).items())
```
can be simplified to
```
self.inplace_pattern = frozendict(inplace_pattern)
self.destroy_map = {o:[i] for o, i in self.inplace_pattern.items()}
```

theano/misc/frozendict.py
```diff
@@ -0,0 +1,49 @@
+import collections
```
nouiz: Include here the original license for that code.

doc/LICENSE.txt
```diff
@@ -9,9 +9,11 @@ All rights reserved.
 Contains code from NumPy, Copyright (c) 2005-2016, NumPy Developers.
 All rights reserved.
 
-Contain CnMeM under the same license with this copyright:
+Contains CnMeM under the same license with this copyright:
 Copyright (c) 2015, NVIDIA CORPORATION. All rights reserved.
 
+Contains frozendict code from slezica’s python-frozendict(https://github.com/slezica/python-frozendict/blob/master/frozendict/__init__.py), Copyright (c) 2012 Santiago Lezica. All rights reserved.
```
nouiz: Good idea to also add it here. But you need to use the license information from that code. It is not all rights reserved.

```
Contains frozendict code from slezica’s python-frozendict(https://github.com/slezica/python-frozendict/blob/master/frozendict/__init__.py), MIT License. See the top of the file theano/misc/frozendict.py for detail.
```

theano/scalar/basic.py
```diff
@@ -3659,7 +3659,7 @@ def __init__(self, inputs, outputs):
         self.nin = len(inputs)
         self.nout = len(outputs)
         self.init_fgraph()       # self.fgraph
-
+        self.init_c_code()
```
nouiz: why did you change this file? I don't see the need and this is a bad idea. I moved this to later in the code pipeline. revert.

theano/tensor/elemwise.py
```diff
@@ -477,14 +476,16 @@ class Elemwise(OpenMPOp):
 
     """
 
+    __props__ = ("scalar_op", "inplace_pattern", "name", "nfunc_spec", "openmp")
```
nouiz: Normally, name should not be a props. I think we should remove it at the same time. We removed it from another node that had it. We don't do this anymore, this is old stuff. Remove it instead from props and from \__init__.

theano/scalar/basic.py
Sentient07: The reason for that : https://github.com/Theano/Theano/pull/4749#issuecomment-252755870 
This Op compares with `_c_code` and the Op doesn't have  the attribute until we call this. When we don't have this, I get this error : 
```
ERROR (theano.gof.opt): SeqOptimizer apply <theano.tensor.opt.FusionOptimizer object at 0x105774610>
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 235, in apply
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 87, in optimize
  File "/Users/Ramana/projects/theano/Theano/theano/tensor/opt.py", line 7156, in apply
    reason=self.__class__.__name__)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/toolbox.py", line 340, in replace_all_validate
    fgraph.replace(r, new_r, reason=reason, verbose=False)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/fg.py", line 531, in replace
    self.change_input(node, i, new_r, reason=reason)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/fg.py", line 471, in change_input
    self.__import_r__(new_r, reason=reason)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/fg.py", line 372, in __import_r__
    self.__import__(variable.owner, reason=reason)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/fg.py", line 435, in __import__
    self.execute_callbacks('on_import', node, reason)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/fg.py", line 611, in execute_callbacks
    fn(self, *args, **kwargs)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 561, in on_import
    self.process_node(fgraph, node)
  File "/Users/Ramana/projects/theano/Theano/theano/gof/opt.py", line 675, in process_node
    if inputs_match and node.op == candidate.op:
  File "/Users/Ramana/projects/theano/Theano/theano/gof/utils.py", line 196, in __eq__
    tuple(getattr(other, a) for a in props))
  File "/Users/Ramana/projects/theano/Theano/theano/scalar/basic.py", line 3775, in __eq__
    return (self._c_code == other._c_code)
AttributeError: 'Composite' object has no attribute '_c_code'
```

Sentient07: I know this is a hack, but I think we shouldn't have the `_c_code` as a key to check the equality. Do you agree @nouiz ?

doc/LICENSE.txt
Sentient07: Here it is given it's rights reserved, https://github.com/slezica/python-frozendict/blob/master/LICENSE.txt . I'm very new to this. I thought since it was given there, i should include it as well

Sentient07:
Regarding the travis error, i'm fixing the instance where the dict is empty neither None nor has any elements. In the upcoming commit I'll fix that and remove the name from props and `__init__` along with addressing the comment regarding the license.

theano/tensor/elemwise.py
Sentient07: When I'm removing name from the `__init__` and `__props__`, the string will change and I have to accordingly change the ground truths in the node assertion tests. Is that fine @nouiz ?

Sentient07:
Updated this PR with a newer commit that 

- Changed the signature of Elemwise by removing the name parameter 
-  Fixed the error that was reported in travis, originating from the tests that were on unpickling of graph.
- Altered the license. 

Still, I am expecting Travis to fail because of unequal nodes in doc tests, arising because of change in `__str__` methods. Referring to the unequal nodes in travis, I'll make necessary changes.

Sentient07:
If the `inplace_pattern` doesn't have any value (or just an empty dict), should the op be `no_inplace` ?

doc/LICENSE.txt
nouiz: Write:

Contains frozendict code from slezica’s python-frozendict(https://github.com/slezica/python-frozendict/blob/master/frozendict/__init__.py). See PATH files for license. And copy the license at the top of that file.

theano/tensor/elemwise.py
```diff
@@ -589,36 +586,13 @@ def make_node(self, *inputs):
...
-            if self.inplace_pattern:
-                items = list(self.inplace_pattern.items())
-                items.sort()
-                return "Elemwise{%s}%s" % (self.scalar_op, str(items))
-            else:
-                return "Elemwise{%s}" % (self.scalar_op)
+            return "Elemwise{%s}%s" % (self.scalar_op, str(items))
         else:
-            return self.name
+            return "Elemwise{%s}" % (self.scalar_op)
```
nouiz: This is a good way. But to lower the change in what we return I would do.
name = self.scalar_op.replace('scalar_', '')

This way elemwise like xlogx won't have its name change.

lamblin:
So, here is what is happening with all the test failures like `theano.tensor.tests.test_basic.Elemwise{trunc}Tester
SyntaxError: invalid syntax`.
The issue is that, earlier, the `__str__` method returned `Elemwise{trunc,no_inplace}` by default, or the `name` attribute that is set in `theano/tensor/inplace.py` by the `@_scal_inplace` decorator. Then, `makeTester` actually performed some string replacement magic to transform that name into something more readable, and that is a valid name for a class, in that case `Trunc`.

If we really want to remove the `name` attribute, then we should fix `_scal_inplace` as well, and maybe other places using it.
Then we have to fix all the places that rely on the name having a specific form. Maybe keeping only what is after `Elemwise{` and before `}` and capitalizing it in `makeTester` would make sense.

Also, there are other tests that are failing because they were literally expecting some string, for instance `Elemwise{mul,no_inplace}`, but now get `Elemwise{mul}`.

Sentient07:
> Then, makeTester actually performed some string replacement magic to transform that name into something more readable, and that is a valid name for a class, in that case Trunc.

If the string that is returned by Elemwise is same as before why would there be a difference in the transform ?

> If we really want to remove the name attribute, then we should fix _scal_inplace as well, and maybe other places using it.

Could you please elaborate more on this?

> Also, there are other tests that are failing because they were literally expecting some string, for instance Elemwise{mul,no_inplace}, but now get Elemwise{mul}.

Regarding this, I have implemented  the `__str__` similar to `_scal_elemwise_with_nfunc`. My bad that I didn't update the PR with the new commit.

lamblin:
> If the string that is returned by Elemwise is same as before why would there be a difference in the transform ?

If it was exactly the same as before, it would not matter. However, it is different. For instance, what used to be "Elemwise{trunc,no_inplace}" is now "Elemwise{trunc}". Moreover, some code was explicitly setting the "name" attribute, which does not work any more, so that is another difference.

> Could you please elaborate more on this?

In [here](https://github.com/Theano/Theano/blob/master/theano/tensor/inplace.py#L8-L30), we call `elemwise.Elemwise(..., name=symbolname)` to explicitly set the name of the Elemwise. If we want to remove this feature, then we should update that file as well.

Sentient07:
> If it was exactly the same as before, it would not matter. However, it is different. For instance, what used to be "Elemwise{trunc,no_inplace}" is now "Elemwise{trunc}". Moreover, some code was explicitly setting the "name" attribute, which does not work any more, so that is another difference.

yes, I compared with the newer code after changing the `__str__` of Elemwise. 

> In here, we call elemwise.Elemwise(..., name=symbolname) to explicitly set the name of the Elemwise. If we want to remove this feature, then we should update that file as well.

@nouiz warned me regarding this. I've fixed this also in the commit. The test regarding no_inplace and inplace passes now. It was only the `makeTester` that was creating the problem. 
A big thank you for pointing out where it was going wrong. My bad that I scrolled through that place for two days without realizing it

Sentient07:
In this PR, I'm stuck with removing `name` parameter from the `Elemwise`. The `_scal_inplace` and `_scal_elemwise_with_nfunc` decorators has the `symbolname` variable which says if the ScalarOp is `inplace` or not. If i get rid of the `name`, I lose track of the fact if the scalar Op is an inplace or not as inside the Elemwise I can see only the `Scalar Op's` name and not the name of instance of the Ops name which are decorated. 
The only solution that I can see which makes the tests pass are, while calling the `makeBroadcastTester` , we can specify the names of the testers? Or, by doing that we're defying the purpose of why the `makeBroadcastTester` exists ?

nouiz:
If you have an elemwise Op and want to know if it is inplace or not, you can do so:

op.destroy_map

If it exist and evaluate to True, then it is inplace.

Ex:

if op.destroy_map:
   # it is an inplace op.
else:
   # it is not inplace.

theano/tensor/elemwise.py
```diff
@@ -477,15 +476,17 @@ class Elemwise(OpenMPOp):
...
         if inplace_pattern is None:
-            inplace_pattern = {}
-        self.name = name
+            inplace_pattern = frozendict({})
         self.scalar_op = scalar_op
-        self.inplace_pattern = inplace_pattern
-        self.destroy_map = dict((o, [i]) for o, i in inplace_pattern.items())
-
+        self.inplace_pattern = frozendict(inplace_pattern)
+        self.destroy_map = dict((o, [i]) for o, i in frozendict(inplace_pattern).items())
```
nouiz: don't create a new frozendict. Use self.inplace_pattern.

theano/tensor/elemwise.py
```diff
@@ -477,15 +476,17 @@ class Elemwise(OpenMPOp):
...
+        if not bool(self.inplace_pattern):
+            self.inplace = False
```
nouiz: Why add this attribute? I know we are not consisten in Theano. Many op have it, but not Elemwise. Before changing that I think we should agree on where we want to go. So if there isn't a good reason, I would not change that.

Sentient07: True. I added this before you had given me a solution to tackle `inplace` and `no_inplace` inside `__str__`

Sentient07:
The failing tests pass locally. I'm unsure of the diff created at few places and I'm commenting there to ensure if it is okay.

Sentient07:
@nouiz The doc tests are failing. I run it locally by `python doc/scripts/docgen.py` . The build is successful locally but fails on travis. I could comprehend why travis is failing but am not able to reproduce those failures locally. I'm pretty sure that there are also other places in the doc where failures are possible, but am not able to see them in the current build. I would like to check all of them locally and fix and push instead of fixing one by one on travis.

nouiz:
I have a lengthy discussion with @lamblin  about this. As the difficult part to finish this PR, is the "name" part and that is independent of the frozendict, we would prefer that you split this PR in 2. 1) The frozendict change including addining Elemwise.__props__ without "name" in that list. But keep the name parameter to Elemwise.__init__. The goal of that PR is not change change how we handle name for Elemwise.

The second PR would be to remove the name parameter in the class Elemwise. We aren't 100% sure how to fix it, but we don't change mangling of the str in Elemwise.__str__ as it is done. For example, the removing of  "scalar_" in that method should not be there. It would be better to remove the "scalar_" at the source, for example in nnet/sigm.py we should have:

scalar_sigmoid = ScalarSigmoid(scalar.upgrade_to_float, name='sigmoid') instead of the current name.

But before going there, can you make a new PR with only the change for frozendict as explained above?

thanks

Sentient07:
Commenting here what I said before for keeping better track. As @nouiz said, I've opened another PR that does only the frozendict part. Once that gets merged, I will rebase this PR and continue the work on removing `name` parameter,  Camel Casing and hoping for to work few more changes that me and @nouiz proposed

nouiz:
Just a note, part of this PR was merged in master in another PR. So this PR need to be rebased on master to finish what wasn't merged. There is still useful stuff here.

Sentient07:
Yeah, I will resume work on this after finishing ROIPooling. I will change the title of this PR.

Sentient07:
@nouiz this PR requires one change. It requires me to change the ground truth of printing of Elemwise. If you are fine with that change, I will go ahead and apply them. I just now observed before opening a new PR that I have few unmerged ones and wanted to wrap up the ones that are almost done.

Sentient07:
@nouiz the travis error message was the change in ground truth that I was talking about in my previous comment.

PR #5073, opened 2016-10-11
 Removing _op_use_c_code attribute

nouiz:
continue https://github.com/Theano/Theano/pull/4959

fix gh-4317

### DIFF
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -125,13 +125,13 @@ class OpFromGraph(gof.Op):
                          [type() for type in self.output_types])
 
     def prepare_node(self, node, storage_map, compute_map, impl):
-        if not hasattr(node.tag, "fn") and impl == 'py':
-            node.tag.fn = orig_function(self.new_inputs,
-                                        self.new_outputs,
-                                        **self.kwargs)
+        if not hasattr(self, "fn") and impl == 'py':
+            self.fn = orig_function(self.new_inputs,
+                                    self.new_outputs,
+                                    **self.kwargs)
 
     def perform(self, node, inputs, outputs):
-        variables = node.tag.fn(*inputs)
+        variables = self.fn(*inputs)
         assert len(variables) == len(outputs)
         for output, variable in zip(outputs, variables):
             # TODO: when function's output-borrowing semantics are correct,
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -124,17 +124,14 @@ class OpFromGraph(gof.Op):
                          list(inputs) + self.shared_inputs,
                          [type() for type in self.output_types])
 
-    def make_thunk(self, node, storage_map, compute_map, no_recycling):
-        ret = super(OpFromGraph, self).make_thunk(node, storage_map,
-                                                  compute_map, no_recycling)
-        if not hasattr(self, "fn"):
-            self.fn = orig_function(self.new_inputs,
-                                    self.new_outputs,
-                                    **self.kwargs)
-        return ret
+    def prepare_node(self, node, storage_map, compute_map):
+        if not hasattr(node.tag, "fn"):
+            node.tag.fn = orig_function(self.new_inputs,
+                                        self.new_outputs,
+                                        **self.kwargs)
 
     def perform(self, node, inputs, outputs):
-        variables = self.fn(*inputs)
+        variables = node.tag.fn(*inputs)
         assert len(variables) == len(outputs)
         for output, variable in zip(outputs, variables):
             # TODO: when function's output-borrowing semantics are correct,
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -124,8 +124,8 @@ class OpFromGraph(gof.Op):
                          list(inputs) + self.shared_inputs,
                          [type() for type in self.output_types])
 
-    def prepare_node(self, node, storage_map, compute_map):
-        if not hasattr(node.tag, "fn"):
+    def prepare_node(self, node, storage_map, compute_map, impl):
+        if not hasattr(node.tag, "fn") and impl == 'py':
             node.tag.fn = orig_function(self.new_inputs,
                                         self.new_outputs,
                                         **self.kwargs)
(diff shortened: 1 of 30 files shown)


theano/compile/builders.py
```diff
@@ -124,17 +124,14 @@ def make_node(self, *inputs):
...
-        if not hasattr(self, "fn"):
-            self.fn = orig_function(self.new_inputs,
-                                    self.new_outputs,
-                                    **self.kwargs)
-        return ret
+    def prepare_node(self, node, storage_map, compute_map, impl):
+        if not hasattr(node.tag, "fn") and impl == 'py':
+            node.tag.fn = orig_function(self.new_inputs,
+                                        self.new_outputs,
+                                        **self.kwargs)
```
abergeron: This could go on self, because it doesn't depend on the inputs.  It would also avoid recompiling for multiple uses of it in the same graph.

theano/gof/cc.py
```diff
@@ -1787,24 +1787,14 @@ def make_all(self, profiler=None, input_storage=None, output_storage=None,
...
-                    thunks[-1].outputs = [storage_map[v] for v in node.outputs]
-
-                finally:
-                    node.op._op_use_c_code = old_value
+                # make_thunk will try by default C code, otherwise
+                # it fall back to python.
+                thunks += [node.op.make_thunk(node,
+                                              storage_map,
+                                              compute_map,
+                                              no_recycling)]
```
abergeron: I would make the linker try both explicitly to make sure it keeps working if the behaviour of `impl=None` changes in the future.

abergeron: Also Clinker only want C code, so I would make it only try 'c'.

theano/sandbox/cuda/fftconv.py
```diff
@@ -61,7 +61,7 @@ def output_type(self, inp):
         return CudaNdarrayType(
             broadcastable=[False] * (inp.type.ndim + 1))
 
-    def make_thunk(self, node, storage_map, _, _2):
+    def make_thunk(self, node, storage_map, _, _2, impl=None):
         super(CuFFTOp, self).make_thunk(node, storage_map, _, _2)
```
abergeron: Super is bad

theano/tensor/elemwise.py
```diff
@@ -890,14 +891,15 @@ def perform(self, node, inputs, output_storage):
             # numpy the first (faster) version leads to segfaults
             if self.ufunc:
                 ufunc = self.ufunc
+            elif not hasattr(node.tag, 'ufunc'):
+                # It happen that make_thunk isn't called, like in
+                # get_scalar_constant_value
+                self.prepare_node(node, None, None, 'py')
+                if self.ufunc:
+                    ufunc = self.ufunc
```
abergeron: We already checked self.ufunc above.  Also, calling perform without prepare_node/make_thunk should not be supported.  We should fix get_scalar_constant_value instead.

theano/gof/cc.py
nouiz: I don't want to duplicate code. If the behavior change, then this is an interface changes and it is normal that this request code changes. If we do other changes, this could request change to the duplicated code. That isn't better.

theano/tensor/elemwise.py
nouiz: Fixing get_scalar_constant_value is outside this PR. I'm also not sure we want to force that. I'm not sure how many places would need to be changed to support that.

theano/scalar/basic.py
```diff
@@ -3664,10 +3664,13 @@ def __init__(self, inputs, outputs):
...
-        self.init_py_impls()  # self._impls
-        for n in theano.gof.graph.list_of_nodes(self.inputs, self.outputs):
-            n.op.prepare_node(n, None, None)
+    def prepare_node(self, node, storage_map, compute_map, impl):
+        if impl == 'py':
+            self.init_py_impls()  # self._impls
+        if not getattr(node.tag, 'graph_prepare_node_called', False):
+            for n in theano.gof.graph.list_of_nodes(self.inputs, self.outputs):
+                n.op.prepare_node(n, None, None, impl)
+            node.tag.graph_prepare_node_called = True
```
nouiz: good catch

theano/sandbox/cuda/extra_ops.py
```diff
@@ -49,28 +49,20 @@ def make_node(self, x):
...
             cuda = theano.sandbox.cuda
             device_id = cuda.use.device_number
-            if device_id is None:
-                cuda.use("gpu",
-                         force=False,
-                         default_to_move_computation_to_gpu=False,
-                         move_shared_float32_to_gpu=False,
-                         enable_cuda=False,
-                         test_driver=True)
-                device_id = cuda.use.device_number
```
nouiz: right. I wanted to convert it to prepare_node, but then I got tired to do so to many ops... I'll revert.

theano/tensor/elemwise.py
nouiz: self.prepare_node() can add to self.ufunc or node.tag.ufunc depending if it can be reused or not. So we need to check it again. I'll add a comment to make this clear.

theano/compile/builders.py
nouiz: outside the scope of this PR, but I still did it.

theano/gof/op.py
```diff
@@ -904,6 +897,9 @@ def make_thunk(self, node, storage_map, compute_map, no_recycling):
         no_recycling
             List of variables for which it is forbidden to reuse memory
             allocated by a previous call.
+        impl
+            Currently, None, 'c' or 'py'. If 'c' or 'py' we will only try
+            that version of the code.
```
f0k: Not sure if you care, but note that changing the interface like this breaks existing code such as the cuda-convnet wrappers (https://github.com/lisa-lab/pylearn2/blob/5513089/pylearn2/sandbox/cuda_convnet/filter_acts.py#L90). Just noticed that by chance. (In linker/vm.py you call make_node with impl=impl, and this breaks if make_node does not expect an impl argument at all.)

nouiz:
updates. Thanks for the info and the PR to pylearn2 to fix pylearn2.

PR #5077, opened 2016-10-11
Add a method to properly convert the graph of Composite to float32.

abergeron:


abergeron:
What is happening with this?

theano/scalar/basic.py
```diff
@@ -3807,3 +3812,79 @@ def __setstate__(self, d):
...
+            if i.dtype == 'float16':
+                mapping[i] = get_scalar_type('float32')()
+            else:
+                mapping[i] = i
+        for node in topo:
+            # Patch up for constants
+            for i in node.inputs:
+                if i not in mapping:
+                    assert type(i) is ScalarConstant
+                    mapping[i] = i
```
nouiz: What if the ScalarConstant is float16? I think in that case, it should be converted to float32.

theano/scalar/basic.py
```diff
@@ -3807,3 +3812,79 @@ def __setstate__(self, d):
...
+            if type(node.op) in self.special:
+                self.special[type(node.op)](node, mapping)
+                continue
+            # make sure we won't produce any float16.
+            assert not any(o.dtype == 'float16' for o in
```
nouiz: It sould be more efficient to do this assert on new_nodes.outputs after line 3843. This won't request the computation of output dtype multiple times.

theano/scalar/basic.py
```diff
@@ -3807,3 +3812,79 @@ def __setstate__(self, d):
...
+        new_node = node.clone_with_new_inputs([inp],
+                                              strict=False)
+        mapping[out] = new_node.outputs[0]
+
+Compositef32.special[Cast] = handle_cast
+
+
+def handle_composite(node, mapping):
+    new_op = node.op.clone_float32()
+    new_outs = new_op(*[mapping[i] for i in node.inputs], return_list=True)
```
nouiz: The call clone_float32 clone the graph, then new_op() will also clone the graph. What about making clone_float32 return op if there is no modification to the graph? I'm not sure if it is worth it as I recently did many optimization to building new Composite op.

theano/scalar/basic.py
```diff
@@ -3807,3 +3812,79 @@ def __setstate__(self, d):
...
+            # cast to f16, convert to f32
+            new_out = cast(inp, 'float32')
+            # change the node for the following if
+            node = new_out.owner
+            mapping[out] = new_out
+            node_ok = True
+    if node.inputs[0].type == float16:
+        if node.op.o_type == inp.type:
+            # cast f16 to new input type, remove
+            mapping[out] = inp
```
nouiz: I don't like that. What it inp.dtype is an int of float64? I would replace this by a cast to float32.

theano/scalar/tests/test_basic.py
```diff
@@ -66,7 +67,27 @@ def tes_mod(self):
             self.assertTrue(fn(a, b) == a%b, (a,))
 
 
+def has_f16(comp):
+    if any(i.type == float16 for i in comp.fgraph.inputs):
+        return True
+    for n in comp.fgraph.apply_nodes:
+        if any(o.type == float16 for o in n.outputs):
```
nouiz: this don't assert constant don't have float16.

What only doing a loop on comp.fgraph.variables and remove the line 71-72 at the same time.

theano/scalar/tests/test_basic.py
```diff
@@ -66,7 +67,27 @@ def tes_mod(self):
...
+        w = int8()
+        x = float16()
+        y = float32()
+        cz = Composite([x, y], [tanh(x + cast(y, 'float16'))])
+        c = Composite([w, x, y], [cz(x, y) - cz(x, y)**2 +
+                               cast(x, 'int16') + cast(x, 'float32') +
+                               cast(w, 'float16')])
+        assert has_f16(c)
+        nc = c.clone_float32()
+        assert not has_f16(nc)
```
nouiz: Make a test with constant in float16. I don't think it is well handled.

theano/scalar/basic.py
abergeron: yes

theano/scalar/basic.py
abergeron: The if above checks that we are casting what was previously a float16 input.  The second if checks if the new input type is the same as the casting output type.  In any case, this should always be correct no matter what types are involved.

In practice this should only trigger for cast of float16 to float32.  And in this case we can remove the cast since the input will already be float32.

theano/scalar/tests/test_basic.py
abergeron: I changed it to loop on variables.

theano/scalar/basic.py
```diff
@@ -3807,3 +3812,80 @@ def __setstate__(self, d):
...
```
nouiz: right.

PR #5082, opened 2016-10-12
For empty list/tuple, use floatX for the dtype instead of the numpy d…

nouiz:
…efault (float64)

### DIFF
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -7118,6 +7118,16 @@ class T_as_tensor_variable(unittest.TestCase):
         new_inp[...] = inp
         x = as_tensor_variable(new_inp)
 
+    def test_empty_dtype(self):
+        old = theano.config.floatX
+        for dtype in ['float16', 'float32', 'float64']:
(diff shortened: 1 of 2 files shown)


nouiz:
Should I bump dev number to dev4?

theano/tensor/tests/test_basic.py
```diff
@@ -7118,6 +7118,16 @@ def test_memmap(self):
         new_inp[...] = inp
         x = as_tensor_variable(new_inp)
 
+    def test_empty_dtype(self):
+        old = theano.config.floatX
+        for dtype in ['float16', 'float32', 'float64']:
+            try:
```
abergeron: What about change_flags?

abergeron:
Globally, I am ok with this.

theano/tensor/tests/test_basic.py
nouiz: It is only a decorator in Theano right now. I don't want to wait for the RNN PR to be able to use the context manager. My first version was with change_flags as a context manager as I taught it was merged.

Issue #5079, opened 2016-10-12
Do something with the old theano.tensor.nnet.conv3D


gvtulder:
When the AbstractConv3d interface (PR #4862) is merged, `theano.tensor.nnet` will have two 3D convolution functions with very similar names: the new `theano.tensor.nnet.conv3d` from AbstractConv3d and the old `theano.tensor.nnet.conv3D` imported from `theano.tensor.nnet.Conv3D`. That might be confusing.

ballasn:
I think theano.tensor.nnet.conv3D should be remove in the long term. We might want to have a warning to it for now, saying that this version is deprecated and that we should use theano.tensor.nnet.conv3d instead?

nouiz:
I would make it raise an error now (people could comment it to get it
working). This will make sure people email us if they want to keep it.
Otherwise we can remove it in a few mounts.

lamblin:
Conv3D is still sometimes inserted in the graph for gradient of strided conv2d on CPU without a BLAS to link with (which happens notably on Windows).
I would wait until #5076 is merged, so we can always rely on corrmm and corr3dmm on CPU and can get rid of the legacy conv2d and the optimization inserting it and maybe its gradient.

gvtulder:
Maybe adding a warning just for the `theano.tensor.nnet.conv3D` shortcut should be a good place to start? You can still keep the Conv3D implementation for these old cases, but without the potentially confusing shortcut.

nouiz:
I don't think we have time for the 0.9 release, so I bump that to 0.10 release.

lamblin:
So here is the plan:
- [ ] remove (or move away) the following files from `theano/tensor/nnet/`:
  - `Conv3D.py`
  - `ConvGrad3D.py`
  - `ConvTransp3D.py`
- [ ] check if all tests still pass (including GPU with old and new back-ends)
- [ ] fix tests if needed
- [ ] add back the files, with only an warning message that instructs users to use the new functions:
  -  `theano.tensor.nnet.conv3d(..., filter_flip=False)` instead of `conv3d(...)`
  - `...conv3d_grad_wrt_inputs(..., filter_flip=False)` instead of `convTransp3d(...)`
  - `...conv3d_grad_wrt_weights(..., filter_flip=False)` instead of `convGrad3d(...)`

nouiz:
done

PR #5084, opened 2016-10-12
Replace Conv3D references in tests with Corr3dMM

gvtulder:
This removes a number of unnecessary calls to the old Conv3D implementation from the tests, replacing them with calls to Corr3dMM instead. Could be part of #5079.

### DIFF
--- a/theano/gpuarray/tests/test_dnn.py
+++ b/theano/gpuarray/tests/test_dnn.py
@@ -832,7 +832,6 @@ def test_conv3d_fwd():
 
         inputs = theano.shared(inputs_val)
         filters = theano.shared(filters_val)
-        bias = theano.shared(numpy.zeros(filters_shape[0]).astype('float32'))
 
         # Compile a theano function for the cuDNN implementation
         conv = dnn.dnn_conv3d(img=inputs, kerns=filters,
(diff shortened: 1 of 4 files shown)


nouiz:
jenkins test this

nouiz:
This PR seem to slow down travis heavily. Sometimes the tests under tensor now take over 2h and cause travis to fail. Any hint on how to speed up the new tests before I dive into it?

gvtulder:
@nouiz: I've seen that problem before while testing some of the other 3D pull requests. But why do you think this particular PR would be causing it? This patch only modifies the gpuarray/sandbox.cuda tests, so they shouldn't affect travis test times at all. The only non-gpu test here is the test for catching "undefined gradient" errors, but that shouldn't take long.

The other 3D convolution/pooling updates added a lot of tests that weren't there before, so perhaps it's more likely that those take too long sometimes? The Jenkins report shows that AbstractConv tests now take 11 minutes, test_corr3d uses about the same.

nouiz:
I started to write another comment, but it seem I forgot to push it!

You are right, it is not this PR. The extra time for the new tests are causing us problems. If the total time for the new test could be halved, it would fix our problem. If it get reduced by 25%, maybe it will fix our current problem.

Do you have time to check that?

gvtulder:
See #5122.

Issue #5088, opened 2016-10-13
Blocks test broken by recent commit


dwf:
```
a16e91f7933e20b1ae77c1d43f8278af9333f2ab is the first bad commit
commit a16e91f7933e20b1ae77c1d43f8278af9333f2ab
Author: Gijs van Tulder <gvtulder@gmail.com>
Date:   Thu Sep 15 17:24:08 2016 +0200

    theano.tensor.signal.Pool with 3D support.

:040000 040000 c969fd3fe408357f15e9403f6d6bc15b240ef1c9 8855314caee1ac5adf9d234961c37628409eeaf5 M      theano
bisect run success
```

dwf:
A simplified failing test case.
```python
from theano.tensor.signal.pool import Pool


def test_wrong_shape():
    assert Pool.out_shape((9, 8, 6), (2, 2)) == [9, 4, 3]
```

This is an embarassment, and a16e91f7933e20b1ae77c1d43f8278af9333f2ab is a living argument for small, granular commits, ladies and gentlemen.

dwf:
Seems that it's specifically this assumption: https://github.com/Theano/Theano/commit/a16e91f7933e20b1ae77c1d43f8278af9333f2ab#diff-7e13e905a5451cdcfd78519fcca02c2fR210

which is a breaking change.

nouiz:
In the code before, we where doing the inverse. Mostly, suppose that we
pooling over the last 2 dimensions.

Did you try to just do that? For 3d pool, we would need to pass the ndim=3
parameter.

dwf:
See my referenced pull request. We are using (channels, height, width)
input for shape computation and the new code assuming a default of ndim-2
means only width gets computed correctly.

nouiz:
I have a PR, the CPU tests pass.

PR #5090, opened 2016-10-13
Regression FIX

nouiz:
Fix gh-5088

### DIFF
--- a/theano/tensor/signal/pool.py
+++ b/theano/tensor/signal/pool.py
@@ -206,9 +206,7 @@ class Pool(OpenMPOp):
             right margins. No padding is added if padding is None.
         ndim : int
             The number of pooling dimensions N.
-            If this number is not specified, the default is set to the
-            (input.ndim - 2), assuming that the first two dimensions of the input
-            are non-pooling dimensions.
+            If this number is not specified, the default is set 2.
 
         Returns
         -------
@@ -219,8 +217,8 @@ class Pool(OpenMPOp):
 
         """
         if ndim is None:
-            ndim = len(imgshape) - 2
-
+            ndim = 2
+        assert ndim > 0
         if len(imgshape) < ndim:
             raise TypeError('imgshape must have at least {} dimensions'.format(ndim))
 
--- a/theano/tensor/signal/pool.py
+++ b/theano/tensor/signal/pool.py
@@ -170,9 +170,7 @@ class Pool(OpenMPOp):
         'average_exc_pad' include it)
     ndim : int
         The number of pooling dimensions N.
-        If this number is not specified, the default is set to the
-        (input.ndim - 2), assuming that the first two dimensions of the input
-        are non-pooling dimensions.
+        If this number is not specified, the default is set to 2.
 
     """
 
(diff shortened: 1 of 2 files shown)


nouiz:
@gvtulder @lamblin @abergeron  what do you think of that PR?

abergeron:
I'm okay-ish.  The blocks test that was broken should not have worked with the old code that required 4d inputs (if I'm not mistaken).

Also, the assumption that there are two extra dimensions is pretty widespread in the convolution code as far as I know.  I'm not against a change of convention on this, but we would have to be careful.

nouiz:
The old code was working for 2d and more inputs. It was supposing it was doing 2d pooling, so only worked on the last 2 dims and passed unchanged all the others and didn't care the number of extra dimensions was there.

abergeron:
Ok for this change then.

dwf:
Given that it did work, there may even be some client library out there using it to compute pooling shapes from 2-tuples, so this seems like a reasonable behaviour.

nouiz:
As it already break on other software, as @dwf I think we should not change the default behavior.

dwf:
Maybe consider adding my 1-line test case from the comments on #5088? And maybe another one for length 2 inputs?

nouiz:
I added the test

abergeron:
Still ok

gvtulder:
I think part of the problem was that it worked if you used pool_2d, pool_3d, but not Pool directly.

Shouldn't you also change the default ndim in make_node?

nouiz:
I just did so.

gvtulder:
There are also 2's in perform, grad and c_code. (I think those are all the "2"s I found.) Of course, since you now set ndim = 2 in __init__, these methods will never have to use the default.

Can I ask: why didn't you set ndim=2 in the list of __init__ arguments?

gvtulder:
Sorry for posting multiple comments in a row.
There are more instances of "img.dim - 2". For example, it also appears in the optimizations for the cuda/gpuarray ops. I think that if you search the patch for instances "- 2" you catch them all.

Finally, as a sort of explanation of this mess: I've had a look at the old code and found that before this patch, the only thing that worked with inputs that were not exactly 4D was the Pool.out_shape method. All other methods would always assert img.ndim == 4. The -2 default didn't change anything for 4D cases, so that seemed acceptable. I apparently didn't realize that Pool.out_shape already implicitly defined a different default.

abergeron:
In the gpuarray ops I would leave it as-is.  I think the kernels actually count on that, or at least the C code does.

nouiz:
I think I found all the clean up. As I changed the default I'll let this PR alone so that the tests finish faster and we can merge faster. I put the clean up in another PR:

https://github.com/Theano/Theano/pull/5092

nouiz:
in gpuarray, I let alone some -2 that where needed by cudnn. But I removed
some that supported ndim=None. We don't support that now. (this is in the
other PR)

nouiz:
jenkins test this

nouiz:
jenkins error (that I restarted) also appear in other PR. So merging. The error is:

  File "/usr/local/miniconda2/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/usr/local/miniconda2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/jenkins-cpu1/workspace/Theano core/theano/tensor/tests/test_gc.py", line 92, in test_gc_never_pickles_temporaries
    assert len_pre_f == len_post_f, (len_pre_f, len_post_f)
AssertionError: (20455, 20450)

PR #5091, opened 2016-10-13
Add bool dtype in scalar and tensor.

abergeron:
This should work in gpuarray too.


nouiz:
some flake8 error: http://darjeeling.iro.umontreal.ca:8080/job/Theano%20PR%20multijob/371/testReport/junit/theano.tests/test_flake8/test_format_flake8/

theano/scalar/basic.py
```diff
@@ -161,7 +162,7 @@ class Scalar(Type):
     TODO: refactor to be named ScalarType for consistency with TensorType.
 
     """
-
+    __props__ = ('dtype',)
```
nouiz: La class Type ne supporte pas les props.

theano/scalar/basic.py
```diff
@@ -681,38 +683,35 @@ def f2(f, names):
...
-# Python 2.4.
-# Note that currently only a few functions use this mechanism, because it is
-# enough to make the test-suite pass with Python 2.4. However, it may prove
-# necessary to use this same mechanism in other places as well in the future.
-class upcast_out(object):
-    def __new__(self, *types):
-        dtype = Scalar.upcast(*types)
-        return get_scalar_type(dtype),
-
+# Using a class instead of a function makes it possible to deep-copy it.
```
nouiz: remove the comment.

theano/scalar/basic.py
```diff
@@ -728,6 +727,8 @@ def same_out_float_only(type):
 
 
 class transfer_type(gof.utils.object2):
+    __props__ = ('transfer',)
```
nouiz: gof.utils.object2 don't support props.

theano/scalar/basic.py
```diff
@@ -748,26 +749,16 @@ def __call__(self, *types):
...
 
-    def __eq__(self, other):
-        return type(self) == type(other) and self.transfer == other.transfer
-
-    def __hash__(self):
-        return hash(self.transfer)
-
 
 class specific_out(gof.utils.object2):
+    __props__ = ('spec',)
```
nouiz: idem

theano/tensor/type.py
```diff
@@ -341,15 +341,9 @@ def values_eq_approx(a, b, allow_remove_inf=False, allow_remove_nan=False,
                 return False
             if a.dtype != b.dtype:
                 return False
-            if 'int' in str(a.dtype):
+            if 'float' not in str(a.dtype):
```
nouiz: why that changes? what about complex? I think you should revert that change.

theano/tensor/basic.py
```diff
@@ -3182,6 +3182,9 @@ def mean(input, axis=None, dtype=None, op=False, keepdims=False,
     for i in axis:
         s = true_div(s, shp[i])
 
+    if s.dtype != shp.dtype and s.dtype in discrete_dtypes:
```
nouiz: Add a comment to explain that this happen with axes=()

theano/gradient.py
```diff
@@ -1334,12 +1334,11 @@ def _float_ones_like(x):
     """ Like ones_like, but forces the object to have a
     floating point dtype """
 
-    rval = tensor.ones_like(x)
+    dtype = x.type.dtype
+    if 'float' not in dtype:
```
lamblin: Why not use `tensor.float_dtypes`?

theano/scalar/basic.py
```diff
@@ -681,38 +683,35 @@ def f2(f, names):
...
 
+def upgrade_to_float(*types):
+    """
+    Upgrade any int types to float32 or float64 to avoid losing precision.
 
-class same_out(object):
-    def __new__(self, type):
-        return type,
+    """
+    conv = {int8: float32,
```
lamblin: Should we add `bool: float32` here as well? (NumPy upcasts it to float16, like int8).

theano/scalar/basic.py
```diff
@@ -1290,9 +1265,9 @@ def output_types(self, types):
 class UnaryBitOp(UnaryScalarOp):
     def output_types(self, *input_types):
         for i in input_types[0]:
-            if i not in (int8, int16, int32, int64):
-                raise TypeError('input to a BitOp must have type int8,'
-                                ' int16, int32 or int64... not %s' % i)
+            if i not in ((bool,) + discrete_types):
```
lamblin: `bool` is in `discrete_types` now, I think, so you can remove it here.

theano/scalar/basic.py
```diff
@@ -1302,10 +1277,13 @@ def grad(self, inputs, output_gradients):
 class BinaryBitOp(BinaryScalarOp):
     def output_types(self, *input_types):
         t0, t1 = input_types[0]
+        if t0 == bool and t1 == bool:
+            return [bool]
         for i in input_types[0]:
-            if i not in (int8, int16, int32, int64):
-                raise TypeError('input to a BitOp must have type int8,'
-                                ' int16, int32 or int64... not %s' % i)
+            if i not in discrete_types:
```
lamblin: If you mean to exclude mixing bool and ints here, maybe you should use `int_types + uint_types`

theano/tensor/elemwise.py
```diff
@@ -1724,7 +1716,7 @@ def __str__(self):
 
     def make_node(self, input):
         input = as_tensor_variable(input)
-        if input.dtype not in ["int8", "uint8"]:
+        if input.dtype is not "bool":
```
lamblin: I don't think a string comparison with `is` is correct.

theano/tensor/elemwise.py
```diff
@@ -1756,7 +1746,7 @@ def __str__(self):
...
```
lamblin: idem

theano/tensor/var.py
```diff
@@ -470,19 +470,17 @@ def __getitem__(self, args):
...
-                                     'Instead you can use non_zeros() such as '
-                                     'tensor[(x == 0).nonzeros()]. '
-                                     'If you are indexing on purpose with an '
-                                     'int8, please cast it to int16.'))
+                if (isinstance(args_el, (numpy.bool_, bool)) or
+                        args_el.dtype == 'bool'):
+                    raise TypeError('TensorType does not support boolean '
+                                    'mask for indexing such as tensor[x==0]. '
+                                    'Instead you can use non_zeros() such as '
+                                    'tensor[(x == 0).nonzeros()]. ')
```
lamblin: We could automatically index with `as_tensor_variable(args_el).nonzeros()` instead of raising an Exception, but it would be for another ticket.

theano/scalar/basic.py
lamblin: Maybe it could?

theano/scalar/basic.py
abergeron: Yes it does because it inherits from object2 (which not has the \__props__ implementation).

theano/scalar/basic.py
abergeron: Yes it does now.

theano/scalar/basic.py
abergeron: done

theano/gradient.py
abergeron: done

theano/scalar/basic.py
abergeron: Ok.

theano/scalar/basic.py
abergeron: Yes, it wasn't part of it for a while.

theano/tensor/elemwise.py
abergeron: Yes

theano/tensor/var.py
abergeron: Yes

theano/tensor/tests/test_elemwise.py
```diff
@@ -826,6 +826,7 @@ def test_reduce_default_acc_dtype(self):
                     uint8='uint64',
                     uint16='uint64',
                     uint32='uint64',
+                    float16='float32',
```
lamblin: I'd be inclined to use 'float64' here as well, actually.

theano/scalar/basic.py
lamblin: My comment was not at the right place, this is a unary Op...

theano/scalar/basic.py
```diff
@@ -1302,10 +1282,13 @@ def grad(self, inputs, output_gradients):
...
+                raise TypeError('input to a BitOp must have type (u)int8, '
+                                '(u)int16, (u)int32 or (u)int64 or '
+                                'be all bools not %s' % i)
```
lamblin: Here, this is where the message mentions "be all bools", which suggests missing is not supported.

abergeron: Yes, you're right.  A bitwise and between a bool and an int8 most likely doesn't do what the user wants.  True will converted to 1 and then we do the bitwise op.

theano/tensor/tests/test_elemwise.py
abergeron: This is in the tests.  And float16 does uses float32 by default for accumulation.  If we want to change that I'm ok with it, but I would make that change in another PR.

nouiz:
travis fail in FAST_COMPILE

lamblin:
I think there may be a problem in the Python implementation of eq or neq or its gradient.

lamblin:
Actually, there is a `Sum{axis=[1], acc_dtype=bool}` in the graph, which output is also a `bool`.
That may be an issue in the code creating that (Prod.grad), but it may reveal a discrepancy between C and Py implementations.

lamblin:
OK, the discrepancy indeed comes from the sum(bool): The underlying representation of the bool array output by Python code is [0, 1, 1], and C code is [0, 1, 2], which are all displayed as [True, False, False]. Then, eq(..., 1) returns [False, True, True] in Python and [False, True, False] in C (which is the desired behaviour).
I'm still working on it.

lamblin:
So, the output dtype of sum(bool) in numpy is int64.
It should be the same in Theano (with acc_dtype=int64 as well, I guess).

theano/scalar/basic.py
```diff
@@ -1513,10 +1528,13 @@ def impl(self, *inputs):
 
     def c_code(self, node, name, inputs, outputs, sub):
         (z,) = outputs
+        op = " * "
+        if z.type == bool:
```
lamblin: Here, `z` is the name of the variable in the C code, it's a string.
You probably want `node.outputs[0].type.dtype` or something like that.

theano/scalar/basic.py
```diff
@@ -2445,6 +2467,8 @@ def grad(self, inputs, gout):
     def c_code(self, node, name, inputs, outputs, sub):
         (x,) = inputs
         (z,) = outputs
+        if z.type == bool:
```
lamblin: Same here.

theano/scalar/basic.py
```diff
@@ -1474,10 +1486,13 @@ def impl(self, *inputs):
 
     def c_code(self, node, name, inputs, outputs, sub):
         (z,) = outputs
+        op = " + "
+        if z.type == bool:
```
lamblin: Same here.

theano/scalar/basic.py
```diff
@@ -2468,7 +2468,7 @@ def grad(self, inputs, gout):
     def c_code(self, node, name, inputs, outputs, sub):
         (x,) = inputs
         (z,) = outputs
-        if z.type == bool:
+        if node.outputs[0].type == bool:
             return "%(z)s = !%(x)s;" % locals()
```
lamblin: Maybe we should just refuse to do that? NumPy says it is deprecated.

theano/scalar/basic.py
```diff
@@ -1585,7 +1585,7 @@ def impl(self, x, y):
     def c_code(self, node, name, inputs, outputs, sub):
         (x, y) = inputs
         (z,) = outputs
-        if z.type == bool:
+        if node.outputs[0].type == bool:
             # xor
             return "%(z)s = (%(x)s || %(y)s) && !(%(x)s && %(y)s);" % locals()
```
lamblin: Here as well, it is deprecated in NumPy.

nouiz:
travis failed.

abergeron:
I should have fixed the last two problems.

nouiz:
I didn't look at the PR except to check if we raise the dev version. It is not yet in the PR. Make sure to do it before this PR get merged. Jenkins fail.

abergeron:
Ok, so all problems are fixed and the only thing missing are the extra tests that @lamblin wanted to do.

abergeron:
I have a PR that fixes some float16 problem which is entangled with the commits in this one.  Once this is in I can rebase and submit another PR, but in the meantime, I'm stumped.

lamblin:
OK, let's merge this.
I'll open a new ticket for additional tests.

PR #5093, opened 2016-10-13
clean following the fix

nouiz:


lamblin:
I did not check if there are more to clean up though, just that these made sense.

PR #5097, opened 2016-10-14
use floatX in gpuarray dnn tests

slefrancois:
~~Tests with float16 are successful.~~

### DIFF
--- a/theano/gpuarray/tests/test_dnn.py
+++ b/theano/gpuarray/tests/test_dnn.py
@@ -33,6 +33,13 @@ def set_precision(floatX):
     return precision
 
 
+# If using float16, cast reference input to float32
+def ref_cast(x):
+    if theano.config.floatX == 'float16':
+        x = T.cast(x, 'float32')
+    return x
+
+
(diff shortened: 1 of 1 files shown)


lamblin:
The failure may be due to non-determinism during the gradient computation, or just rounding error.
Since convolution is a linear operation, you can safely increase the step of the finite-difference estimate when calling `verify_grad`. I think we do it at other places for cuDNN convolution.

slefrancois:
It passes in float32, which was hardcoded before. I wonder if we should have a precision-dependent step size, or  find one that works for all floatX?

slefrancois:
The main conv3d tests fail in float16, apparently due to the reference C code not being implemented for float16.

lamblin:
Maybe the solution could be to have the reference code running in float32 in that case.

slefrancois:
With float32 refs and higher relative tolerances, the conv3d tests pass in float16. Other tests have deeper issues, such as CuDNN itself apparently not supporting float16 on those ops.

PR #5105, opened 2016-10-17
32bit

nouiz:


nouiz:
As travis didn't finish, I included a new commit that fix the old test failing in python 32bit.

PR #5069, opened 2016-10-09
GpuCorrMM and GpuCorr3dMM in new backend

gvtulder:
This adds the GpuCorrMM and GpuCorr3dMM ops from sandbox.cuda to the new gpuarray backend. It also applies the nnet.conv2d to dnn_conv / GpuCorrMM and nnet.conv3d to dnn_conv / GpuCorr3dMM optimizations.

This pull request uses code from the AbstractConv3d pull request #4862, so that should be merged first. (The very last patch in this long list adds the GpuCorrMM ops, all the other patches are from #4862.)

PR #5108, opened 2016-10-18
Fix a change to gpuarray conv from gh-5069. dnn conv wasn't working a…

nouiz:
…nymore in fast_compile.


nouiz:
jenkins test this

Issue #5109, opened 2016-10-18
Document cuDNN RNN binding


lamblin:
Follow-up to #4915.
We should document that we are supporting cuDNN v5's RNN bindings (without dropout) and their gradient, and make that easily searchable from the web and the built-in search box.

This could simply mean updating the docstrings of `RNNBlock` class, maybe adding an example and a few keywords, and then using `autodoc` to integrate it with Sphinx.

souravsingh:
@lamblin The issue can be closed.

nouiz:
I'm reopening as I think we need better doc. We don't have example usage. Also I think all the shapes stuff isn't well documented.

PR #5119, opened 2016-10-19
Doc update about cudnn and cudnn RNN.

nouiz:
Partially do https://github.com/Theano/Theano/issues/5109 to be able to email about that new feature.


slefrancois:
jenkins test this

doc/library/sandbox/cuda/dnn.txt
```diff
@@ -145,42 +145,68 @@ get an error when cuDNN can not be used with them, use this flag:
...
-.. automodule:: theano.sandbox.cuda.dnn
+RNN
+===
+
+New back-end only! Without dropout support.
+
+.. automodule:: theano.gpuarray.dnn
    :noindex:
-   :members: GpuDnnSoftmax, GpuDnnSoftmaxGrad
+   :members: RNNBlock
```
slefrancois: This should also be added to the CuDNN section of the theano.gpuarray.dnn documentation.

doc/library/gpuarray/dnn.txt
```diff
@@ -132,30 +132,70 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
...
-   :members: GpuDnnSoftmax, GpuDnnSoftmaxGrad
+   :members: RNNBlock
+
+.. _libdoc_gpuarray_dnn_softmax:
+
+Softmax
+=======
+
+You can manually use the op :class:`GpuDnnSoftmax
+<theano.gpuarray.dnn.GpuDnnSoftmax>` to use its extra feature.
```
nouiz: I'm not able to make this reference work. I tried with `:ref:`, `:class:`, `:func:` and it don't work. Any idea? Could it be the noindex in the autodoc?

@abergeron do you know?

theano/sandbox/cuda/dnn.py
```diff
@@ -1424,7 +1424,7 @@ class GpuDnnPoolDesc(GpuOp):
 
     Note
     ----
-    Not used anymore. Only needed to reload old pickled files.
+    Do not used anymore. Only needed to reload old pickled files.
```
slefrancois: Do not use anymore

doc/library/gpuarray/dnn.txt
```diff
@@ -132,30 +132,70 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
...
+   :members: dnn_batch_normalization_train, dnn_batch_normalization_test
 
-Softmax Ops
-===========
+.. _libdoc_gpuarray_dnn_rnn:
+
+RNN
+===
+
+New back-end only! Without dropout support.
```
abergeron: I would say that the warning about new backend only is unnecessary since this is in gpuarray/dnn.txt.

doc/library/gpuarray/dnn.txt
abergeron: It probably doesn't work because the module for the file at the top is wrong.  It's not gpuarray.dnn, but theano.gpuarray.dnn.  This matters to sphinx.

Otherwise it's because all the autodocs are marked with `:noindex:` which makes them unlinkable.

abergeron: Also the correct syntax is
```
:class:`theano.gpuarray.dnn.GpuDnnSoftmax`
```

theano/gpuarray/dnn.py
```diff
@@ -2337,25 +2337,34 @@ def make_node(self, desc, x, hx, y, reserve, w):
...
 
 class RNNBlock(object):
+    """
+    An object that allow us to use CuDNN v5 RNN implementation.
+    TODO: make an example how to use. You can check that test until we have it.
+
+
+    Parameters
+    ----------
+
```
abergeron: There shouldn't be a blank line here.

nouiz:
The problem is the noindex. Do you know how to work around that?

We need the noindex (or another work around) as otherwise, we have errors
like:

/home/nouiz/repos/Theano/doc/library/gpuarray/dnn.txt:5: SEVERE: Duplicate
ID: "module-theano.gpuarray.dnn".
/home/nouiz/repos/Theano/doc/library/gpuarray/dnn.txt:5: SEVERE: Duplicate
ID: "module-theano.gpuarray.dnn".
/home/nouiz/repos/Theano/doc/library/gpuarray/dnn.txt:5: SEVERE: Duplicate
ID: "module-theano.gpuarray.dnn".
/home/nouiz/repos/Theano/doc/library/gpuarray/dnn.txt:5: SEVERE: Duplicate
ID: "module-theano.gpuarray.dnn".


If you don't know better work around, I'll do one big autodoc with small
section with just reference to it like in the sparse doc:
http://deeplearning.net/software/theano/library/sparse/#list-of-implemented-operations

abergeron:
One big section with references is probably how you have to do it.

nouiz:
done.

nouiz:
fixed

PR #5122, opened 2016-10-19
Somewhat smaller/faster 3d convolution tests

gvtulder:
This makes the 3d convolution/correlation test examples somewhat smaller. Especially reducing the very large input for verify_gradient in the corr3d tests saves a lot of time.


gvtulder:
I don't know exactly how much time this saves. The AbstractConv3d tests are more difficult to reduce, I think, because they are already pretty limited compared with the 2d version.

nouiz:
jenkins test this

nouiz:
This PR seem to have fixed the slowdown in travis. Thanks! I'll merge now even if I started jenkins.

PR #5124, opened 2016-10-21
By default, do validation during elemwise_inplace_optimizer approx 10…

xoltar:
… times, regardless of number of apply nodes


nouiz:
We have a new mechanism that is better then this. See gh-6000

PR #5125, opened 2016-10-21
Avoid linear search for removed nodes

xoltar:


nouiz:
This was fixed in https://github.com/Theano/Theano/pull/5428. Thanks for discovering this slowdown!

PR #5127, opened 2016-10-22
Speed up GpuCorrMM in the new back-end

nouiz:
There was a slow down in the new back-end compared to the old one related to GpuCorrMM. This should fix it. I didn't benchmark it yet. I'll ask the user the reported this to benchmark it.

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -263,6 +263,9 @@ class Kernel(object):
 def get_ctype(dtype):
     if dtype is gpuarray.GpuArray:
         return "gpudata *"
+    elif isinstance(dtype, numpy.dtype):
+        # The if is needed as numpy.dtype can't be compared to gpuarray.[S]SIZE
+        return 'npy_' + dtype.name
     elif dtype == gpuarray.SIZE:
         return "size_t"
     elif dtype == gpuarray.SSIZE:
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -488,6 +488,11 @@ class BaseGpuCorrMM(CGpuKernelBase, BlasOp):
     def get_params(self, node):
         return node.inputs[0].type.context
 
+    def get_op_params(self):
+        return [("ADDR32_MAX", "4294967295"),
+                ("SADDR32_MIN", "-2147483648"),
+                ("SADDR32_MAX",  "2147483647")]
+
     def c_code_cache_version(self):
         # raise this whenever modifying any of the support_code_files
         return (0, 1)
--- a/theano/gpuarray/corr_gemm.c
+++ b/theano/gpuarray/corr_gemm.c
@@ -1,6 +1,6 @@
 #section kernels
 
-#kernel dilated_im2col_kernel : size, *, size, size, size, size, size, size, size, size, size, size, size, size, size, * : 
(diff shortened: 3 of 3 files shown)


theano/gpuarray/basic_ops.py
```diff
@@ -263,6 +263,9 @@ def m(t):
 def get_ctype(dtype):
     if dtype is gpuarray.GpuArray:
         return "gpudata *"
+    elif isinstance(dtype, numpy.dtype):
+        # The if is needed as numpy.dtype can't be compared to gpuarray.[S]SIZE
+        return 'npy_' + dtype.name
```
abergeron: Why do you say numpy.dtype can't be compared?  I never had any problems with this before.

theano/gpuarray/blas.py
```diff
@@ -488,6 +488,11 @@ def flops(self, inp, outp):
     def get_params(self, node):
         return node.inputs[0].type.context
 
+    def get_op_params(self):
+        return [("ADDR32_MAX", "4294967295"),
+                ("SADDR32_MIN", "-2147483648"),
+                ("SADDR32_MAX",  "2147483647")]
```
abergeron: These should probably be embedded in the code rather than be op_params.  I doubt they will ever change.

theano/gpuarray/corr_gemm.c
```diff
@@ -191,6 +191,165 @@ KERNEL void col2im_kernel(const ga_size n,
   }
 }
 
+#kernel dilated_im2col_int_kernel : int, *, int, int, int, int, int, int, int, int, int, int, int, int, int, * :
```
abergeron: Use uint32 instead where there was a size before.  Also use ga_uint.

theano/gpuarray/corr_gemm.c
```diff
@@ -213,24 +372,43 @@ int im2col(const size_t max_threads_dim,
   size_t n_blocks = (num_kernels + threads_per_block - 1) / threads_per_block;
   int err;
   GpuKernel *kernel;
+  // Here, we only accept non signed stride. So no need to compare to SADDR32_MIN
+  int use_int = (data_im_offset < ADDR32_MAX) &&
+    (stride_h < SADDR32_MAX) && (stride_w < SADDR32_MAX);
```
abergeron: Stride here is not the stride for an array, but the convolution stride.  Also it can't ever be negative due to being a size_t.  Finally I'm pretty sure this test isn't enough.  We have to make sure `kernel_h * kernel _w * height_col * width_col`, `stride_h * kernel_w * height_col * width_col * h_col` and  `stride_w * height_col * width_col * w_col` don't go over.  There may be other key expressions like this.

theano/gpuarray/blas.py
nouiz: It make finding them easier. I got them from libgpuarray. It is the same macro name. But if you insist to embed, I won't object.

theano/gpuarray/corr_gemm.c
nouiz: no, I don't want regression compared to the old back-end. The old back-end was using int, not uint. I can use ga_int, no problem with that. I recall some special case of uint on the GPU.

abergeron:
I just want to note that we plan to do something about this in libgpuarray instead because it would be crazy to duplicate all the kernels that touch indexing data and have a wrapper switch between 64 and 32 bits.

PR #5132, opened 2016-10-24
warn when profiling.ignore_first_call isn't used.

nouiz:
This will help user find the strange profiling information themself without asking questions.

### DIFF
--- a/theano/compile/profiling.py
+++ b/theano/compile/profiling.py
@@ -213,9 +213,9 @@ class ProfileStats(object):
         if (config.profile and
                 ((hasattr(theano, 'sandbox') and
                   hasattr(theano.sandbox, 'cuda') and
-                  theano.sandbox.cuda.cuda_enabled)) or(
+                  theano.sandbox.cuda.cuda_enabled) or (
                       hasattr(theano, 'gpuarray') and
-                      theano.gpuarray.pygpu_activated)):
+                      theano.gpuarray.pygpu_activated))):
             if os.environ.get('CUDA_LAUNCH_BLOCKING', '0') != '1':
                 raise Exception(
                     "You are running the Theano profiler with CUDA enabled."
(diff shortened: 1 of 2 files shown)


theano/compile/profiling.py
```diff
@@ -206,9 +210,12 @@ def reset(self):
...
     def __init__(self, atexit_print=True, flag_time_thunks=None, **kwargs):
-        if (hasattr(theano, 'sandbox') and
-                hasattr(theano.sandbox, 'cuda') and
-                theano.sandbox.cuda.cuda_enabled):
+        if (config.profile and
+                ((hasattr(theano, 'sandbox') and
+                  hasattr(theano.sandbox, 'cuda') and
+                  theano.sandbox.cuda.cuda_enabled)) or(
+                      hasattr(theano, 'gpuarray') and
+                      theano.gpuarray.pygpu_activated)):
```
abergeron: This or clause is outside of the config.profile one which causes the error to happen all the time.

nouiz: fixed

theano/gof/vm.py
```diff
@@ -752,9 +752,11 @@ def accept(self, fgraph, no_recycling=None, profile=None):
 
         """
         if (config.profile and
-                hasattr(theano, 'sandbox') and
-                hasattr(theano.sandbox, 'cuda') and
-                theano.sandbox.cuda.cuda_enabled):
+                ((hasattr(theano, 'sandbox') and
+                  hasattr(theano.sandbox, 'cuda') and
+                  theano.sandbox.cuda.cuda_enabled) or(
```
abergeron: The parenthesis should be on the next line.

abergeron: Or at least have a space before.

nouiz: fixed

PR #5134, opened 2016-10-24
Fix a problem where the samples buffer would not be filled.

abergeron:
Fix an issue with newer versions of libgpuarray and big samples (on the order of a few thousands) where it would not fill the samples buffer completely leaving uninitialized memory at the end.

The exact cutoff at which the problems appears varies depending on the GPU on which the code is run.


theano/sandbox/rng_mrg.py
```diff
@@ -879,7 +880,7 @@ def gpu_kernels(self, node, name):
             x22 = state_data[idx*6+4];
             x23 = state_data[idx*6+5];
 
-            for (ga_uint i = idx; i < Nsamples; i += Nstreams_used)
+            for (ga_uint i = idx; i < Nsamples; i += step)
```
nouiz: I don't think this is good. We don't re-read the stream seed if we change seed. This is what I think happen if step != Nstreams_used.

I do not think the old back-end need a fix with a rapid check. The new and old back-end don't schedule gs and ls the same way.

abergeron: Probably not.

I'm working on a test.

abergeron:
I've fixed the consistency problem and added a test to make sure it doesn't break in the future.

abergeron:
This should be ok to merge now.

PR #5139, opened 2016-10-25
remove from var.py

Faruk-Ahmed:
#5135


theano/scan_module/scan_opt.py
```diff
@@ -149,8 +149,7 @@ def remove_constants_and_unused_inputs_scan(node):
     all_ins = gof.graph.inputs(op_outs)
     for idx in xrange(op.n_seqs):
         node_inp = node.inputs[idx + 1]
-        if (isinstance(node_inp, tensor.TensorConstant) and
-                node_inp.tag.unique_value is not None):
```
nouiz: This isn't a valid change. This could explain some travis failure. I think it should be:

    if isinstance(node_inp, tensor.TensorConstant) and node_inp.ndim == 0:

ballasn: I think node_inp.tag.unique_value  is None when  node_inp.ndim == 0 so here the condition should be if isinstance(node_inp, tensor.TensorConstant) and node_inp.ndim != 0 ? @nouiz Am I missing something?

nouiz: I think it is == 0. Now we only support unique value for ndim 0.

nouiz:
There is many travis errors. But I think some of them would be fixed by a rebase. Can you do one today?

PR #5138, opened 2016-10-25
Deprecate ds, st, padding parameters in pooling

gvtulder:
This should rename the following pooling parameters for `pool_2d`, `pool_3d` and `Pool` (see #4933):
- `ds` to `ws`
- `st` to `stride`
- `padding` to `pad`

It keeps the old named parameters for backwards compatibility, but these will produce a warning.

Issue #4933, opened 2016-09-02
User interface change: pool_2d rename parameter ds to ws


nouiz:
Both GPU back-end use ws. Some internal stuff in the gpu back-ends is fixed by this PR:

https://github.com/Theano/Theano/pull/4756/

But this don't change the CPU user interface, as it is using ds for historical reason (downsample size).

I think we should be consistent in the user interface in CPU and GPU and I prefer ws to ds.

We must keep the support for the old interface with a deprecation warning and we should support unpickling old files and automatically convert them to the new interface.

lamblin:
We should change the following names in the Python interface:
- `ds` to `ws`
- `st` to `stride`
- `padding` to `pad`

There is no need to change it in the CPU Op for now, so there should be no problem regarding unpickling.

lamblin:
Now that `pool_3d` has been merged, the same change should be done there as well.
If it is done quickly enough, I think we don't need to worry about deprecation warnings and compatibility.

lamblin:
Closed by #5142

PR #5142, opened 2016-10-25
Pool 2d rename

ChihebTrabelsi:
This pull request contains modifications of the functions parameters in pool.py and test_pool.py. The modifications have been made for coherence purposes. Deprecation warnings and exceptions have been added in order to deal with all possible scenarios in pool.py. The test also contains some examples where it should print a deprecation warning.


theano/tensor/signal/pool.py
```diff
@@ -799,11 +1034,11 @@ def __init__(self, ignore_border, mode='max', ndim=2, openmp=None):
     def prepare_node(self, node, storage_map, compute_map, impl):
         if len(node.inputs) < 5:  # 5 for AveragePoolGrad, 6 for MaxPoolGrad
             # Old interface
-            self.ndim = len(node.op.ds)
+            self.ndim = len(node.op.ws)
             self.mode = node.op.mode
-            ws = theano.tensor.constant(node.op.ds)
-            st = theano.tensor.constant(node.op.st)
-            pad = theano.tensor.constant(node.op.padding)
```
lamblin: We should revert that, because the old serialized Ops will still have their attributes named 'ds', 'st' and 'padding'

gvtulder:
Perhaps link this to #4933 as well.

ChihebTrabelsi:
I took into account modification in #5138.

lamblin:
Jenkins test this please

Issue #5129, opened 2016-10-22
theano.tensor.var return float32 for float16


guoxuesong:
I add debug info to ~/Lasagne/lasagne/layers/normalization.py

> print 'input',input.dtype
> print 'var',T.var(input,self.axes).dtype

got this:

> input float16
> var float32

nouiz:
thanks for the report. Here is pure Theano code that reproduce this and give more information on the problem:

```
>>> import theano
>>> theano.tensor.vector(dtype='float16')
<TensorType(float16, vector)>
>>> v=theano.tensor.vector(dtype='float16')
>>> theano.tensor.var(v)
var
>>> theano.tensor.var(v).dtype
'float32'
>>> theano.printing.debugprint(theano.tensor.var(v), print_type=True)
Elemwise{true_div,no_inplace} [id A] <TensorType(float32, scalar)> 'var'   
 |Sum{axis=[0], acc_dtype=float64} [id B] <TensorType(float32, scalar)> ''   
 | |Elemwise{pow,no_inplace} [id C] <TensorType(float32, vector)> ''   
 |   |Elemwise{sub,no_inplace} [id D] <TensorType(float16, vector)> ''   
 |   | |<TensorType(float16, vector)> [id E] <TensorType(float16, vector)>
 |   | |Elemwise{Cast{float16}} [id F] <TensorType(float16, (True,))> 'mean'   
 |   |   |Elemwise{true_div,no_inplace} [id G] <TensorType(float32, (True,))> ''   
 |   |     |InplaceDimShuffle{x} [id H] <TensorType(float16, (True,))> ''   
 |   |     | |Sum{axis=[0], acc_dtype=float32} [id I] <TensorType(float16, scalar)> ''   
 |   |     |   |<TensorType(float16, vector)> [id E] <TensorType(float16, vector)>
 |   |     |InplaceDimShuffle{x} [id J] <TensorType(float32, (True,))> ''   
 |   |       |Subtensor{int64} [id K] <TensorType(float32, scalar)> ''   
 |   |         |Elemwise{Cast{float32}} [id L] <TensorType(float32, vector)> ''   
 |   |         | |Shape [id M] <TensorType(int64, vector)> ''   
 |   |         |   |<TensorType(float16, vector)> [id E] <TensorType(float16, vector)>
 |   |         |Constant{0} [id N] <int64>
 |   |InplaceDimShuffle{x} [id O] <TensorType(int16, (True,))> ''   
[...]

PR #5144, opened 2016-10-26
make var() return float16 when input is float16.

nouiz:
fix gh-5129.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3249,11 +3249,12 @@ def var(input, axis=None, ddof=0, keepdims=False, corrected=False):
     centered_input = input - mean_input
 
     # return the mean sqr
+    two = constant(2, dtype=centered_input.dtype)
     if ddof == 0:
-        v = mean((centered_input ** 2), axis, keepdims=keepdims)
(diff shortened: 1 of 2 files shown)


nouiz:
I think https://github.com/Theano/Theano/pull/5091 also fix this. The problem is the constant 2 that get casted to int16 that cause the upcast to float32. But gh-5091 will make the 2 be int8, so this won't upcast to float32.

abergeron:
This will fix the immediate problem, but the larger problem is that upgrade_to_float will never return something smaller than float32.  Maybe we should change it to return float16 when floatX=float16?

abergeron:
Forget what I said, pow uses upcast_out so @nouiz's analysis is right.  Also, I believe that my bool PR will fix this.

nouiz:
Do you think we should merge both? How close to be merged is your PR?

abergeron:
My PR is ready to be merged, but @lamblin wants to do some additional tests.

nouiz:
I think we should merge this PR even if both fix it. It will help prevent the optimization local_upcast_elemwise_constant_inputs to trigger.

@abergeron are you ok to merge this one?

PR #5146, opened 2016-10-27
Insert GpuSolve only when CULA is available

f0k:
The `local_gpu_solve` optimizer inserted `GpuSolve` even if `GpuSolve` was not usable due to a missing CULA library, causing an exception later on. This PR adds a check to insert `GpuSolve` only if it can be used.


nouiz:
jenkins test this

f0k:
Jenkins is happy.

PR #5153, opened 2016-10-29
Dimshuffle{0,2}(Subtensor[i:j, :, k:l]) => Subtensor[i:j, 0, k:l] #4647

olimastro:
Note: I also fixed some things I found were not good on my previous gh-4647 tickets. I mistakenly used the same name twice for a test.

This does crash into an error but I don't know how to fix it. I tried to give back a new Subtensor object then realized that this could not be done as it would lack the inputs! So I tried to modify directly the Subtensor.idx_list and then replace the dimshuffle with its Subtensor.idx_list changed input, but it throws an error...almost there!

gh-4647

### DIFF
--- a/theano/tensor/opt_uncanonicalize.py
+++ b/theano/tensor/opt_uncanonicalize.py
@@ -203,7 +203,7 @@ def local_dimshuffle_subtensor(node):
             # the arguments missing from the dimshuffles must be dims that are broadcastable
             broadcastable = input_.broadcastable
 
-            missing_dims = range(input_.ndim)
+            missing_dims = list(range(input_.ndim))
             for dim in new_order:
                 missing_dims.remove(dim)
 
--- a/theano/tensor/opt_uncanonicalize.py
+++ b/theano/tensor/opt_uncanonicalize.py
@@ -189,12 +189,14 @@ def local_dimshuffle_subtensor(node):
         if 'x' in node.op.new_order :
             return False
         new_order = node.op.new_order
-        past_dim = new_order[0]
-        for dim in new_order[1:]:
-            if not dim > past_dim:
-                return False
-            else:
-                past_dim = dim
+        # new order could be empty
+        if len(new_order) > 1:
+            past_dim = new_order[0]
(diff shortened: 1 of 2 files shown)


theano/tensor/opt_uncanonicalize.py
```diff
@@ -173,3 +173,50 @@ def local_dimshuffle_alloc(node):
...
+    scrap the dimshuffle and index the subtensor with 0
+
+    x[i:j, :, k:l].dimshuffle(0, 2) => x[i:j, 0, k:l] if x.broadcastable == (False, True, False)
+    """
+    if isinstance(node.op, DimShuffle) and node.inputs[0].owner:
+        # the dimshuffle can only drop dimensions (cannot reshape nor add 'x')
+        if 'x' in node.op.new_order :
+            return False
+        new_order = node.op.new_order
+        past_dim = new_order[0]
```
lamblin: `new_order` may be empty (for instance if all dimensions were broadcastable and we drop them all, to end up with a scalar). In that case, it can be consider "sorted", and the optimization can apply.
You can also try to see if there is a function in Python or NumPy that checks if a list/tuple is sorted.

olimastro:
This seems to work, however it assumes that we will want to insert the 0 at places in idx_list in the form of slice(None, None, None), not sure if that is correct for all the cases

lamblin:
I think it would be correct to only match slice(None), but then we would have to exit the optimization if the slice is something else.

However, since the broadcastable pattern is `True`, it means there is exactly one element on that axis, so we can always apply the optimization, regardless of what the slice was.
If the replaced slice had types in it, though, we would have to remove the corresponding symbolic outputs.

lamblin:
Closing and reopening to trigger a new merge and, hopefully, make Travis pass.

lamblin:
Jenkins test this please

nouiz:
travis have an error in python 3:

https://travis-ci.org/Theano/Theano/jobs/177796346#L3289

nouiz:
thanks

PR #5157, opened 2016-10-30
Cuda pooling with strides

aam-at:
Add strides support for cuda pooling.


theano/sandbox/cuda/blas.py
```diff
@@ -2506,11 +2502,11 @@ def c_code(self, node, nodename, inp, out, sub):
...
                 CudaNdarray_HOST_DIMS(%(x)s)[3]))
         {
+            // init with zeros
+            // for st larger than ds some pixels in the input will be skipped
             Py_XDECREF(%(gx)s);
-            %(gx)s = (CudaNdarray*)CudaNdarray_New();
-            if ((NULL == %(gx)s)
-                || CudaNdarray_alloc_contiguous(%(gx)s, 4,
-                                                CudaNdarray_HOST_DIMS(%(x)s)))
+            %(gx)s = (CudaNdarray*)CudaNdarray_ZEROS(4, (int*)CudaNdarray_HOST_DIMS(%(x)s));
```
nouiz: Can you do the call to ZEROS only when it is needed? Otherwise, it could cause speed regression compared to the current code.

theano/sandbox/cuda/blas.py
```diff
@@ -2600,76 +2597,38 @@ def c_support_code_apply(self, node, nodename):
...
-                            // gx[image_row][image_col][x_row][x_col]
-                            //   = (my_z == x[image_row][image_col][
-                            //                x_row][x_col]) ? my_gz : 0.0f;
-                            gx[i0*gxS0 + i1*gxS1 + x_row*gxS2 + x_col*gxS3]
-                               = (my_z == x[i0*xS0 + i1*xS1 + x_row*xS2 +
-                                            x_col*xS3]) ? my_gz : 0.0f;
+                            my_gz = gz[i0*gzS0 + i1*gzS1 + i2*gzS2 + i3*gzS3];
+                            // for overlapping pooling regions concurrent
+                            // writes are possible
+                            atomicAdd(&gx[i0*gxS0 + i1*gxS1 + x_row*gxS2 + x_col*gxS3], my_gz);
```
nouiz: @abergeron do you think it is useful to make a condition and do an atomicAdd only when there is overlapping?

theano/sandbox/cuda/blas.py
```diff
@@ -2729,29 +2690,29 @@ def c_code(self, node, nodename, inp, out, sub):
...
                 CudaNdarray_HOST_DIMS(%(z)s)[3]))
         {
+            // init with zeros
+            // for st larger than ds some pixels in the input will be skipped
             Py_XDECREF(%(gz)s);
-            %(gz)s = (CudaNdarray*)CudaNdarray_New();
-            if ((NULL == %(gz)s)
-                || CudaNdarray_alloc_contiguous(%(gz)s, 4,
-                                                CudaNdarray_HOST_DIMS(%(z)s)))
+            %(gz)s = (CudaNdarray*)CudaNdarray_ZEROS(4, (int*)CudaNdarray_HOST_DIMS(%(z)s));
```
nouiz: idem, call ZEROS only if needed to prevent slowdown compared to current code.

theano/sandbox/cuda/blas.py
abergeron: Since we want to pursue a deterministic mode for theano, it would be best to have an alternate mode to avoid using atomics.

But if we do want to use them, on the later generations of GPUs, there is barely any cost above normal memory access if there are no conflicts so I don't think we need to avoid them.  I think the branching cost would be greater.

abergeron:
I didn't review this in detail, but I question the wisdom of adding features to the old backend.  I get that the code is already written and all, but accepting this means that we have to port it to the new backend to get feature parity.

nouiz:
We did not port this to the new back-end. So there is no extra port. Cudnn
is the only option for pooling in the new back-end.

I have no objection to this PR. Just that I'm not sure why you can't use
cudnn for this.

theano/sandbox/cuda/blas.py
aam-at: Okay, I will add this check.

aam-at:
@nouiz I understand your concerns and agree that cudnn is much faster. I needed `PoolingGradGrad` with strides for my work and cudnn currently does not offer this functionality. As a side effect of adding strides to GradGrad, I also added strides to other pooling classes in the old backend.

In #5159 I added `PoolingGradGrad` for gpuarray backend too.

For this pull request, I will review your comments. I am sure that atomicAdd for `Grad` and `GradGrad` can be avoided by launching threads with caffe style `CUDA_KERNEL_LOOP` (in fact for new backend `atomicAdd` is unnecessary). I can add it but it will require larger changes.

aam-at:
@nouiz I borrowed kernels from caffe pooling layer for cuda backend. Based on 2d pooling, I also added 3d pooling. As new class implements MaxPooling interface completely (make node with tensorvariables including padding and strides) I renamed it to `GpuMaxPool*`. I also added tests for strides, padding and 3d pooling. In this new updated implementation no atomics and zero initialization is used.

Similarly to cuda backend I will work on adding pooling kernels for new gpuarray backend.

One question, how should I cite caffe?

nouiz:
Just a note that travis have this error:

/home/travis/build/Theano/Theano/theano/sandbox/cuda/tests/test_blas.py:425:1: E303 too many blank lines (3)

nouiz:
As this take code from caffe, can you verify which caffe verion you took and give a pointer to its license? It is better co check that now. I think it should be fine if the license didn't change.

We will need to add a note in the file LICENSE.txt about that.

nouiz:
I'll review first the version in the new back-end: https://github.com/Theano/Theano/pull/5198

The license comment still apply for it.

aam-at:
I used latest version of caffe - [LICENSE.txt](https://github.com/BVLC/caffe/blob/master/LICENSE)
It seems to me that license didn't change.

aam-at:
Following discussion for #5198, I moved pooling operation to separate file `pool.py`. Additionally, in the last commit I added `AveragePool` and `AveragePoolGrad` which were missing for cuda backend before.

Unlike new backend, there are no problem with optimization order since dnn has local optimizations to life blas operations. So all tests should pass.

nouiz:
I did a comment 9 days ago, but in the wrong PR. So I'll repeat it here:


https://github.com/nouiz/Theano/tree/aam-at-gpu_pool

Is a branch with codes to add the missing class to allow registering the opt in the right order and allow the user to select them as normal.

I still need to review more this PR. Can you add the commit to this PR or allow me to push to this branch in your fork?

nouiz:
Wrong PR again!

This PR is in the old back-end. We are deprecating it and I don't have time to review this PR. You did another PR with the equivalent in the new back-end. So I'll close this PR.

The gpuarray PR: https://github.com/Theano/Theano/pull/5198

aam-at:
Okay, I understand. I am going to keep this branch. If anyone comes across this thread, feel free to ask questions or report bug.

PR #5159, opened 2016-10-31
Gpuarray pool grad grad

aam-at:
Adds support for gpuarray pool grad grad. Currently, does not support padding and similar to cuda pool grad grad.

I placed implementation in `blas.py` which is similar to cuda backend.

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -29,6 +29,7 @@ from theano.tensor.nnet.abstract_conv import (BaseAbstractConv,
                                               AbstractConv3d,
                                               AbstractConv3d_gradWeights,
                                               AbstractConv3d_gradInputs)
+import theano.tensor.signal.pool as pool
 
 from theano.tests.breakpoint import PdbBreakpoint
 
@@ -46,7 +47,8 @@ from .blas import (gpu_dot22, GpuGemm, GpuGer, GpuGemmBatch,
                    gpugemmbatch_no_inplace,
                    gpugemv_no_inplace, gpugemv_inplace,
                    GpuCorrMM, GpuCorrMM_gradInputs, GpuCorrMM_gradWeights,
-                   GpuCorr3dMM, GpuCorr3dMM_gradInputs, GpuCorr3dMM_gradWeights)
+                   GpuCorr3dMM, GpuCorr3dMM_gradInputs, GpuCorr3dMM_gradWeights,
+                   GpuDownsampleFactorMaxGradGrad)
 from .blocksparse import (GpuSparseBlockGemv, GpuSparseBlockOuter,
                           gpu_sparse_block_outer,
                           gpu_sparse_block_outer_inplace,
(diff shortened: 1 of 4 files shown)


gvtulder:
Nice. Any chance of a 3D version? :)

aam-at:
@gvtulder I am not working on it but it shouldn't be hard. I will probably add later average pooling for 4d input which will be super-easy to add.

aam-at:
@gvtulder BTW, I was looking at source code `im2col` in new gpuarray backend (it seems you contributed it) and I saw that you are launching kernels by passing first threads and then blocks when it should be other way around. I am bringing it up as I had a problem kernel call and in last commit fix it for my code. Is that correct way?

gvtulder:
@aam-at I think you're right, that doesn't look good. I'll see if I can fix that. (It's a bit confusing that you have to pass blocks then threads, but that the `GpuKernel_call` method then swaps them again internally.)

theano/gpuarray/opt.py
```diff
@@ -1610,6 +1612,34 @@ def local_gpua_lift_abstractconv_graph(op, context_name, inputs, outputs):
     return [op(*inps)]
 
 
+@register_opt()
+@op_lifter([pool.DownsampleFactorMaxGradGrad])
+@register_opt2([pool.DownsampleFactorMaxGradGrad])
+def local_gpu_downsample_factor_max_grad_grad(op, ctx_name, inputs, outputs):
+    from theano.sandbox.cuda.opt import _check_constant_args_pool
```
abergeron: Copy the code here rather than import from the old backend.  It will make it easier when we want to delete the old one.

aam-at:
@abergeron I added padding, so check for constant args is unnecessary now.

theano/gpuarray/pool_grad_grad.c
```diff
@@ -16,9 +17,11 @@ KERNEL void pool_grad_grad_kernel(const ga_size nthreads,
     const ga_size ph = (index / pooled_width) % pooled_height;
     const ga_size c = (index / pooled_width / pooled_height) % channels;
     const ga_size n = (index / pooled_width / pooled_height / channels);
-    const ga_size hstart = ph*stride_h;
+    ga_int hstart = static_cast<ga_int>(ph*stride_h) - static_cast<ga_int>(pad_h);
```
abergeron: I don't think it's a good idea to cast to int here.  If you want a signed type, use ga_ssize.

aam-at: At the beginning `ph` will be zero and subtracting from unsigned int results in `MAX_INT - 1` value which will fail `max(result, 0)`. That's why I am using casting here.

abergeron:
jenkins test this please

aam-at:
@abergeron Build failed due to compilation error (include ext_cuda). I removed unnecessary headers. Can you rerun test server?

abergeron:
jenkins test this please

aam-at:
@abergeron Test have passed successfully but it had merging conflict which I resolved in the last commit.

abergeron:
When there is a conflict we strongly prefer a rebase to a merge.  Can you do that instead?

aam-at:
@abergeron My bad. I did rebase now.

abergeron:
jenkins test this please

PR #5160, opened 2016-10-31
Fixed typo in config doc

hsintone:
Doc had two nan_is_error attributes. One of them should be big_is_error.


nouiz:
thanks.

jenkins test this

PR #5161, opened 2016-10-31
GpuDownsampleFactorMaxGradGrad3d in cuda (+bug?)

gvtulder:
Given #5159, I thought it might be useful to have a 3D version of `GpuDownsampleFactorMaxGradGrad` as well. While trying to build that, I discovered (what I suspect is) a bug in the existing 2D code. It assumes that there is only one maximum per region (unlike the cpu version, which correctly sums all maxima in the region).

Miraculously, the random test samples do not contain multiple maxima, so the 2D test runs correctly. If you give an input of all ones the tests will fail.

The 3D function in this patch is a straightforward copy of the 2D function, so it has the same problem. I 'fixed' this by using unique values in the tests.

The new gpuarray function in #5159 seems to implement this correctly.

### DIFF
--- a/theano/sandbox/cuda/blas.py
+++ b/theano/sandbox/cuda/blas.py
@@ -2865,3 +2865,209 @@ class GpuDownsampleFactorMaxGradGrad(GpuOp):
             }
         }
         """ % locals()
+
+
+class GpuDownsampleFactorMaxGradGrad3d(GpuOp):
+    """
+    Implement the grad of downsample with max on the gpu.
+
+    """
+    __props__ = ('ds', 'ignore_border')
+
+    def __init__(self, ds, ignore_border):
+        self.ds = tuple(ds)
+        self.ignore_border = ignore_border
+
+    def make_node(self, x, z, gx):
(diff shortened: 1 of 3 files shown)


gvtulder:
The `atomicAdd` proposed in #5157 would fix the bug, if I understand the meaning of `atomicAdd` correctly.

nouiz:
https://github.com/nouiz/Theano/tree/aam-at-gpu_pool

Is a branch with codes to add the missing class to allow registering the opt in the right order and allow the user to select them as normal.

I still need to review more this PR. Can you add the commit to this PR or allow me to push to this branch in your fork?

gvtulder:
@nouiz: You should be able to push to the branch in my fork now. (I don't exactly know what you want to do, so this seems the easiest solution.)

nouiz:
My comment was for the wrong PR, it was for: https://github.com/Theano/Theano/pull/5198

nouiz:
As I think it is included in gh-5198 for the new back-end and we don't have time to review for the old back-end, I'm closing this PR.

PR #5164, opened 2016-10-31
Fix some problems in float16.

abergeron:
This fixes a number of problems observed in float16 over the Deep Learning Tutorials.

The major points are:
 * The gemm optimizations now work on float16 inputs
 * tensor and scalar share code to cast constant values to a specific type


theano/scalar/basic.py
```diff
@@ -121,33 +122,165 @@ def as_scalar(x, name=None):
...
+        elif isinstance(x, numpy.ndarray):
+            x_ = x
+        else:
+            # Here x is probably a list or a tuple. If it contains a
+            # long, we will behave like the current NumPy version: it
+            # will work if the long fits in int64 or uint64.
+            x_ = numpy.asarray(x)
+            if x_.size == 0 and not hasattr(x, 'dtype'):
+                x_ = numpy.asarray(x, dtype=config.floatX)
+    assert type(x_) in [numpy.ndarray, numpy.memmap]
```
nouiz: Here, we are in scalar and we force an ndarray as output? The docstring tell that this work with scalar. Can you clarify the docstring to tell it will convert everything to numpy ndarray, event python and numpy scalar?

theano/tensor/basic.py
```diff
@@ -3155,11 +2998,9 @@ def mean(input, axis=None, dtype=None, op=False, keepdims=False,
...
     else:
         sum_dtype = None
-
-    # float16 overflows way too fast for sum
-    if ((sum_dtype == 'float16' or input.dtype == 'float16') and
-            acc_dtype != 'float16'):
-        sum_dtype == 'float32'
+        # float16 overflows on the cast way too often
+        if input.dtype == 'float16':
+            sum_dtype = 'float32'
```
nouiz: It would be great to have a test for this. This is easy to break and it would be better not to break it.

theano/scalar/basic.py
abergeron: If you can give me a clearer formulation, I will put it in.

However, it says that it will convert to a "properly typed numpy value".  I don't see anything that could give an impression that it behaves differently for python scalars or anything.

nouiz: ok

nouiz:
jenkins fail: http://darjeeling.iro.umontreal.ca:8080/job/Theano%20PR%20multijob/467/testReport/theano.tests/test_flake8/test_format_flake8/

I kill travis.

nouiz:
jenkins test this.

PR #5165, opened 2016-10-31
Correct blocks/threads for gpuarray CorrMM and Corr3DMM

gvtulder:
@aam-at noticed that the number of threads and the number of blocks were swapped in the kernel calls for `corr3d_gemm.c` and `corr_gemm.c`. This PR corrects this. It also adds two tests with larger inputs to check the correction. (The tests don't really take a lot of time, they are just a little bit too large.)

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -490,7 +490,7 @@ class BaseGpuCorrMM(CGpuKernelBase, BlasOp):
 
     def c_code_cache_version(self):
         # raise this whenever modifying any of the support_code_files
-        return (0, 1)
+        return (0, 2)
 
     def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width=None):
         """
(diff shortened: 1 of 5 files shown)


aam-at:
@gvtulder In large input test, It seems to me that the number of kernels `channels * width * height` must be at least `max_threads_dim ^ 2`, not `channels * height > max_threads_dim` as it is divided by the number of threads.

gvtulder:
Indeed. I fixed those comments now. (The test values were already 1024 * 1024, it was just the comment.)

nouiz:
jenkins test this

nouiz:
add to whitelist

nouiz:
thanks for the PR. The *kernel*_call method use the same order as CUDA <<<>>> syntax. I want to change the order inside libgpuarray to be the same order. So yes, we don't have the same order everywhere now, but at some point, we should.

nouiz:
jenkins failure is the bool problem that is fixed. So merging.

abergeron:
The jenkins failure is not related to this PR.  It should be cleared up, but I'll run the tests again just to be sure.

jenkins test this please

Issue #1399, opened 2013-06-03
MaxAndArgmax -> GpuMaxAndArgmax and just GpuArgmax


nouiz:
A user spent 40% of his time in gpu transfer (20%) and 20% in MaxAndArgmax. He only use the argmax.
-  (see gh-5147) Make and Argmax only op, or do only 1 pass over the input in MaxAndArgmax.
- Make a Gpu version of argmax or maxandargmax.
  
  For the GPU version, we will do it only in the new back-end. The GPU implementation was merged in libgpuarray: https://github.com/Theano/libgpuarray/pull/185 and https://github.com/Theano/libgpuarray/pull/276. We need to make an op in Theano that will call this implementation. Check the CPU op MaxAndArgmax in tensor/basic.py. The make_node should be similar. No need for all the other implementation. Add perform/c_code and inherit from COp to have a working implementation.

lamblin:
This popped up a couple of times for different users in the last weeks.
Currently, the only way of having something close to argmax on the GPU is something like:
`(T.eq(x, x.argmax(axis=..., keepdims=True)) * T.arange(x.shape[...]).dimshuffle(...).sum(axis=...)`
and it only works when there are no ties for the max.

nouiz:
What is the number of dimensions of the inputs and what was the axis they where reducing on?

lamblin:
I don't remember the different cases, but I think we should do it properly and support all reduction patterns for at least 4 dimensions, and not try to only do the strict minimum that would break as soon as someone works on examples instead of minibatches, or vice-versa.

nouiz:
What was the number of input dimensions? What was the axis?

danielvarga:
@nouiz @lamblin I have this exact use case, see https://groups.google.com/forum/#!topic/theano-users/E7ProqnGUMk

In my case, the following artificial example is roughly representative, that is, I'd like to reduce a 1000x5000 2d variable to 1000 indices.

https://gist.github.com/danielvarga/d0eeacea92e65b19188c

danielvarga:
@lamblin I implemented and benchmarked your clever hack at https://gist.github.com/danielvarga/d0eeacea92e65b19188c, but it does not bring significant improvements on my task/machine:

```
bestIndices = T.argmin(s, axis=0)
  Time in Function.fn.__call__: 3.870916e-02s (99.867%)
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Class name>
[...]

PR #5168, opened 2016-11-01
This is my proposal for GpuMaxAndArgmax (issue #1399).

notoraptor:
GpuMaxAndArgmax op is implemented in a separate file
in theano/gpuarray.

A test script has also been written.

To integrate this op in Theano, an optimization has been
added to ttheano/gpuarray/opt.py.

To check if GpuMaxAndArgmax is included (or not)
in theano graphs:

Without 'device=cuda' (MaxAndArgmax should be used);
```
$ theano-cache purge && THEANO_FLAGS=floatX=float32,profile=True,profiling.ignore_first_call=True nosetests --verbose --nocapture theano/gpuarray/tests/test_GpuMaxAndArgmax.py:TestGpuMaxAndArgmax.test_none
```

With 'device=cuda' (GpuMaxAndArgmax should be used):
```
$ theano-cache purge && THEANO_FLAGS=floatX=float32,device=cuda,profile=True,profiling.ignore_first_call=True nosetests --verbose --nocapture theano/gpuarray/tests/test_GpuMaxAndArgmax.py:TestGpuMaxAndArgmax.test_none
```


theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+            }
+            """
+        ret += """
+        size_t  %(name)s_output_ndim = %(name)s_input_ndim - %(name)s_redux_len;
+        size_t* %(name)s_output_dims = NULL;
+        if(%(name)s_output_ndim == 0) {
+            /* Current backend function GpuArray_maxandargmax does not work when
+             * all axes need to be reduced. So to handle this case, we create a view
+             * of the input as a matrix with 1 row and as many columns as elements
+             * in the input, so that the 2nd dimenson of the matrix will be reduced. */
```
abergeron: Is this due to an error you encountered?  If so it should be fixed in libgpuarray instead of here.  Otherwise why do you think it is the case?

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+            size_t total_size = 1;
+            for(size_t i = 0; i < %(name)s_input_ndim; ++i) {
+                total_size *= PyGpuArray_DIM(%(X)s, i);
+            }
+            size_t newdims[2] = {1, total_size};
+            %(name)s_input = new GpuArray;
```
abergeron: No C++ for GpuArray. If this is really a temporary, allocate it statically on the stack, it's faster anyway.

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+        axes_ctype = 'int64_t'
+        ret = """
+        GpuArray* %(name)s_input = &%(X)s->ga;
+        size_t %(name)s_input_ndim = PyGpuArray_NDIM(%(X)s);
+        """
+        if NoneConst.equals(node.inputs[1]):
+            ret += """
+            unsigned  %(name)s_redux_len = %(name)s_input_ndim;
+            unsigned* %(name)s_axes_to_reduce = new unsigned[%(name)s_redux_len];
+            for(unsigned i = 0; i < %(name)s_redux_len; ++i) {
```
abergeron: Our C style includes placing a space before the parenthesis for non-functions (`for`, `if`, ...)

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+        # axes_ctype = pygpu.gpuarray.dtype_to_ctype(node.inputs[1].dtype)
+        axes_ctype = 'int64_t'
+        ret = """
+        GpuArray* %(name)s_input = &%(X)s->ga;
+        size_t %(name)s_input_ndim = PyGpuArray_NDIM(%(X)s);
+        """
+        if NoneConst.equals(node.inputs[1]):
+            ret += """
+            unsigned  %(name)s_redux_len = %(name)s_input_ndim;
+            unsigned* %(name)s_axes_to_reduce = new unsigned[%(name)s_redux_len];
```
abergeron: Please use malloc instead of new.

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+            }
+            """
+        else:
+            assert node.inputs[1].ndim == 1
+            ret += """
+            unsigned  %(name)s_redux_len = PyArray_DIM(%(axes)s, 0);
+            unsigned* %(name)s_axes_to_reduce = new unsigned[%(name)s_redux_len];
+            for(unsigned i = 0; i < %(name)s_redux_len; ++i) {
+                %(name)s_axes_to_reduce[i] =
+                    (unsigned)( ((%(axes_ctype)s*)PyArray_DATA(%(axes)s)) [i * (PyArray_STRIDES(%(axes)s)[0] / sizeof(%(axes_ctype)s))] );
```
abergeron: PyArray_GETPTR1() can help you here.

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+                }
+                for(++current_input_pos; current_input_pos < %(name)s_input_ndim; ++current_input_pos) {
+                    %(name)s_output_dims[++current_output_pos] = PyGpuArray_DIM(%(X)s, current_input_pos);
+                }
+            }
+        }
+        if(theano_prep_output(&%(max)s, %(name)s_output_ndim, %(name)s_output_dims, %(max_typecode)s, GA_C_ORDER, %(X)s->context)) {
+            %(fail)s
+        }
+        if(theano_prep_output(&%(argmax)s, %(name)s_output_ndim, %(name)s_output_dims, %(argmax_typecode)s, GA_C_ORDER, %(X)s->context)) {
```
abergeron: Since this is likely to always be `int64`, then you can use `GA_INT64`.

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+            GpuArray_sync(&%(max)s->ga);
+            GpuArray_sync(&%(argmax)s->ga);
+            """
+        return ret % {'X': input_names[0], 'axes': input_names[1], 'max': output_names[0], 'argmax': output_names[1],
+                      'axes_ctype': axes_ctype, 'max_typecode': max_typecode, 'argmax_typecode': argmax_typecode,
+                      'name': name, 'fail': sub['fail']}
+
+    def c_code_cleanup(self, node, name, inputs, outputs, sub):
+        return """
+        delete[] %(name)s_output_dims;
```
abergeron: What happens if you `delete[]` NULL? There is a path that call `%(fail)s` before you allocate this.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
+from unittest import TestCase
+from theano.tests import unittest_tools as utt
+import numpy as np
+import theano
+import theano.tensor as T
+
```
abergeron: import .config to have the gpuarray test setup (and protection against running when there is no GPU).

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+
+def randomTensor(*shapes):
+    dimlist = shapes
+    size = 1
+    for dimsize in dimlist:
+        size *= dimsize
+    return np.random.normal(size=size).astype(np.float32).reshape(dimlist)
```
abergeron: There is a function called rand_gpuarray in test_basic_ops that you should use instead.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+    reshaped_x = transposed_x.reshape(new_shape)
+    return (ref_max, np.argmax(reshaped_x, axis=-1))
+
+
+class TestGpuMaxAndArgmax(TestCase):
+    # We run all tests with 5-D tensors of 10 000 000 elements.
+    # NB: In each test, any first call of theano function should be ignored
+    # with Theano config flag profiling.ignore_first_call=True.
+    # To just check if GpuMaxAndArgmax is called:
+    # $ theano-cache purge && THEANO_FLAGS=floatX=float32,device=cuda,profile=True,profiling.ignore_first_call=True \
```
abergeron: Why do you need theano-cache purge.  It is not normal for tests to require this.

Also never run the tests with anything other than `device=cpu`.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+    # nosetests --verbose theano/gpuarray/tests/test_GpuMaxAndArgmax.py:TestGpuMaxAndArgmax.test_none
+
+    def _basic_test_tensor5(self, axis=None):
+        M = T.tensor5()
+        max_M = T.max(M, axis=axis)
+        argmax_M = T.argmax(M, axis=axis)
+        f = theano.function([M], [max_M, argmax_M])
```
abergeron: Instead of relying on `device=cuda`, set the mode of the function to `mode_with_gpu` from `.config`.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+    # NB: In each test, any first call of theano function should be ignored
+    # with Theano config flag profiling.ignore_first_call=True.
+    # To just check if GpuMaxAndArgmax is called:
+    # $ theano-cache purge && THEANO_FLAGS=floatX=float32,device=cuda,profile=True,profiling.ignore_first_call=True \
+    # nosetests --verbose theano/gpuarray/tests/test_GpuMaxAndArgmax.py:TestGpuMaxAndArgmax.test_none
+
+    def _basic_test_tensor5(self, axis=None):
+        M = T.tensor5()
+        max_M = T.max(M, axis=axis)
+        argmax_M = T.argmax(M, axis=axis)
```
abergeron: Does this really test MaxAndArgmax?  If so, you should add a check that GpuMaxAndArgmax is really in the graph.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+        f = theano.function([M], [max_M, argmax_M])
+        test_matrix = randomTensor(1000, 100, 10, 5, 2)
+        f(test_matrix)
+        theano_max, theano_argmax = f(test_matrix)
+        ref_max, ref_argmax = numpyMaxAndArgmax(test_matrix, axis=axis)
+        utt.assert_allclose(ref_max, theano_max)
+        utt.assert_allclose(ref_argmax, theano_argmax)
+
+    def _basic_test_assert_equals(self, axis1, axis2):
```
abergeron: What does this test exactly?

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+        self._basic_test_tensor5((3, 1))
+        self._basic_test_tensor5((3, 2))
+        self._basic_test_tensor5((3, 3))
+        self._basic_test_tensor5((3, 4))
+
+        self._basic_test_tensor5((4, 0))
+        self._basic_test_tensor5((4, 1))
+        self._basic_test_tensor5((4, 2))
+        self._basic_test_tensor5((4, 3))
+        self._basic_test_tensor5((4, 4))
```
abergeron: I don't believe that an exhaustive test like this is productive since it will just take a very long time to run.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,139 @@
...
+        utt.assert_allclose(theano1, theano2)
+        utt.assert_allclose(ref1, theano1)
+
+    def test_none(self):
+        self._basic_test_tensor5(None)
+
+    def test_all_axes(self):
+        self._basic_test_tensor5((0, 1, 2, 3, 4))
+
+    def test_1_axe(self):
```
abergeron: `axis` is the singular form of `axes`

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+        context_name = infer_context_name(X)
+        # Check axis and convert it to Python variable.
+        if (isinstance(axis, (integer_types, numpy.integer)) or
+           (isinstance(axis, numpy.ndarray) and axis.ndim == 0)):
+            axis = [int(axis)]
+        elif isinstance(axis, (tuple, list, numpy.ndarray)):
+            axis = [int(a) for a in axis]
+        elif isinstance(axis, Variable):
+            if NoneConst.equals(axis):
+                axis = None
```
notoraptor: Ok! I guess the wrapper method for max-and-argmax computation is ```def max_and_argmax(a, axis=None, keepdims=False)``` in theano/tensor/basic.py, isn't it ? So I'll transfer the axis parameter checking in that method.

theano/gpuarray/GpuMaxAndArgmax.py
notoraptor: Yes, I have an error when all axes need to be reduced.

Indeed I think it comes from libgpuarray:
In file https://github.com/Theano/libgpuarray/blob/master/src/gpuarray_reduction.c
In function ```static int maxandargmaxCheckargs (maxandargmax_ctx* ctx)```:
There is this code:
```
	/* Insane src or reduxLen? */
	if(!ctx->dstMax || !ctx->dstArgmax || !ctx->src || ctx->src->nd == 0 ||
	    ctx->reduxLen == 0 || ctx->reduxLen >= ctx->src->nd){
		return ctx->ret=GA_INVALID_ERROR;
	}
```
I think the problem is the test ```ctx->reduxLen >= ctx->src->nd```
That should better be ```ctx->reduxLen > ctx->src->nd``` (```>``` instead of ```>=```, as the case ```ctx->reduxLen == ctx->src->nd``` is a valid case, corresponding to the reduction of all axes).

theano/gpuarray/GpuMaxAndArgmax.py
notoraptor: In C++, deleting a NULL pointer does normally nothing ! http://en.cppreference.com/w/cpp/language/delete

But as you required, I will remove any C++ syntax from the code !

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
notoraptor: It was to check if GpuMaxAndArgmax does really compute the same results when axes are passed with different formats, for example "None" vs "(0,1,2,3,4)" (None vs a list/tuple with all axes), or "(1,0)" vs "(0,0,1)" (to check if the latter tuple is correctly sorted and simplified).

But it is indeed a little paranoid ! I'll remove that.

theano/gpuarray/GpuMaxAndArgmax.py
notoraptor: It seems there is no GA_INT64 in libgpuarray doc: http://deeplearning.net/software/libgpuarray/c_api.html

Nor in types.h : https://github.com/Theano/libgpuarray/blob/master/src/gpuarray/types.h

I think I should better let pygpu.gpuarray.dtype_to_typecode do the work, for a future compatibility !

theano/gpuarray/GpuMaxAndArgmax.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def perform(self, node, inputs, outputs):
+        X, axes = inputs
+        max, max_idx = outputs
+        if axes is None:
+            axes = tuple(range(X.ndim))
+        else:
+            axes = tuple(axes)
+            # axes = tuple(int(ax) for ax in axes)
+        max[0] = theano._asarray(numpy.max(X, axes), dtype=node.outputs[0].dtype)
```
notoraptor: Ok. I will try to rewrite this implementation using pygpu.

At a first glance, it seems I can use ```pygpu.reduction.reduce1``` function to compute max, but I still don't find a way to compute argmax.

I will look for that tomorrow. If I don't find a solution, I shall remove the "perform" implementation.

theano/gpuarray/tests/test_GpuMaxAndArgmax.py
notoraptor: Ok! As I wrote and modified C code frequently on the issues I have worked, I used to always clear the cache so that any C code (or compiled code ?) is reset before running the tests.

From now I will clear the cache only when I indeed modify a C code (or is it useless) ?

notoraptor:
New huge update, @abergeron !

theano/tensor/basic.py
```diff
@@ -1635,6 +1608,26 @@ def max_and_argmax(a, axis=None, keepdims=False):
         will broadcast correctly against the original tensor.
 
     """
+    # Check axis and convert it to a Python list of integers.
```
abergeron: This should convert to a theano vector instead of a list of ints.

theano/tensor/basic.py
```diff
@@ -1434,7 +1407,7 @@ def c_code(self, node, name, inp, out, sub):
         x, axis = inp
         max, argmax = out
         fail = sub["fail"]
-        if NoneConst.equals(node.inputs[1]):
+        if NoneConst.equals(node.inputs[1]) or len(node.inputs[1].data) == node.inputs[0].ndim:
```
abergeron: The NoneConst case should not happen anymore.  However, you can't depend on the shape of the input to generate the C code since that can change without theano recompiling the C code.  If it is really necessary to have a numeric value for axis at compile time, then make is a parameter of the op, not an input.

If that becomes too much (after all the issue was not about fixing crazy things in the CPU code), you can also choose to leave the CPU code alone and perform the necessary conversions of the axis in the optimization that moves to the GPU.

theano/gpuarray/reduction.py
```diff
@@ -110,80 +81,68 @@ def c_code(self, node, name, input_names, output_names, sub):
...
+        }
+
         size_t  %(name)s_output_ndim = %(name)s_input_ndim - %(name)s_redux_len;
         size_t* %(name)s_output_dims = NULL;
-        if(%(name)s_output_ndim == 0) {
+        if (%(name)s_output_ndim == 0) {
             /* Current backend function GpuArray_maxandargmax does not work when
              * all axes need to be reduced. So to handle this case, we create a view
              * of the input as a matrix with 1 row and as many columns as elements
              * in the input, so that the 2nd dimenson of the matrix will be reduced. */
```
notoraptor: Yes ! Please see my previous comment about it !

When all axes need to be reduced, I must reshape, or the execution fails with a runtime error:

```
RuntimeError: Unexpected error in an Op's C code. No Python exception was set.
Apply node that caused the error: GpuMaxAndArgmax(<GpuArrayType<None>(float32, (False, False, False, False, False))>, TensorConstant{[0 1 2 3 4]})
Toposort index: 0
Inputs types: [GpuArrayType<None>(float32, (False, False, False, False, False)), TensorType(int64, vector)]
Inputs shapes: [(1000, 100, 10, 5, 2), (5,)]
Inputs strides: [(40000, 400, 40, 8, 4), (8,)]
Inputs values: ['not shown', array([0, 1, 2, 3, 4])]
Outputs clients: [[HostFromGpu(gpuarray)(GpuMaxAndArgmax.0)], [HostFromGpu(gpuarray)(GpuMaxAndArgmax.1)]]

```

It seems it is related to libgpuarray.

abergeron: One of the problems here is that you call `%(fail)s` without ensuring that a python exception is set.  The second one is that it does not seem to work.

We should fix both.  I would prefer a proper fix for the issue of all-reduced rather than a workaround like this.

nouiz:
This pr will need a rebase

abergeron:
Also the code to support reduce on all axes has been merged in libgpuarray so we can remove the workaround in here.

notoraptor:
New update.

--
CPU op MaxAndArgmax rewritten so that it now takes the axes
as a parameter op. Same update for GPU op GpuMaxAndArgmax.

max_and_argmax wrapper rewritten to fully check the axis before
passing it to MaxAndArgmax.

Some other files have also been updated to ensure that the
optimization which replace CPU op by GPU op works well and
that all tests involving MaxAndArgmax still work well after
the updates.

GPU op rewritten to handle the last libgpuarray update.

test_reduction rewritten. It now tests also 0-d arrays (scalars).

I have run the following tests with success. The new
update of libgpuarray has been downloaded, compiled
and fully installed before running these tests.

```
nosetests -v theano/gpuarray/tests/test_opt.py
    # There is 1 fail here, but not related to MaxAndArgmax:
    # ERROR: theano.gpuarray.tests.test_opt.test_local_lift_abstractconv_gpu_shape
    # RuntimeError: cuDNN is required for convolution and pooling
nosetests -v theano/tensor/nnet/tests/test_nnet.py
nosetests -v theano/tensor/tests/test_opt_uncanonicalize.py
nosetests -v theano/tensor/tests/test_basic.py
CUDA_LAUNCH_BLOCKING=1 THEANO_FLAGS=floatX=float32,profile=True,profiling.n_ops=1000,profiling.n_apply=1000,profiling.ignore_first_call=True,profiling.destination=profiling.log nosetests --nocapture --verbose theano/gpuarray/tests/test_reduction.py
```

notoraptor:
But there is a weird problem now.

I had to do a rebasing before to put the update online.

Before the rebasing, the GpuMaxAndArgmax Op was faster than MaxAndArgmax:

```
Function profiling
==================
  Message: Sum of all(48) printed profiles at exit excluding Scan op profile.
  Time in 48 calls to Function.__call__: 9.868618e+00s
  Time in Function.fn.__call__: 9.867025e+00s (99.984%)
  Time in thunks: 9.866515e+00s (99.979%)
  Total compile time: 2.936071e+00s
    Number of Apply nodes: 3
    Theano Optimizer time: 4.788456e-01s
       Theano validate time: 1.185656e-03s
    Theano Linker time (includes C, CUDA code generation/compiling): 5.794275e-01s
       Import time 1.585245e-03s
       Node make_thunk time 5.678055e-01s

Time in all call to theano.grad() 0.000000e+00s
Time since theano import 32.999s
Class
---
  86.8%    86.8%       8.566s       1.78e-01s     C       48      48   theano.gpuarray.basic_ops.HostFromGpu
  13.2%   100.0%       1.298s       5.41e-02s     C Py      24      24   theano.tensor.basic.MaxAndArgmax
   0.0%   100.0%       0.002s       1.01e-04s     C       24      24   theano.gpuarray.reduction.GpuMaxAndArgmax
   ... (remaining 0 Classes account for   0.00%(0.00s) of the runtime)
```

After the rebasing, (now test_reduction.py must be run with CUDA_LAUNCH_BLOCKING=1), and then GpuMaxAndArgmax becomes again slower than MaxAndArgmax:

```
Function profiling
==================
  Message: Sum of all(48) printed profiles at exit excluding Scan op profile.
  Time in 48 calls to Function.__call__: 1.044904e+01s
  Time in Function.fn.__call__: 1.044725e+01s (99.983%)
  Time in thunks: 1.044671e+01s (99.978%)
  Total compile time: 3.299059e+00s
    Number of Apply nodes: 3
    Theano Optimizer time: 5.327797e-01s
       Theano validate time: 1.325607e-03s
    Theano Linker time (includes C, CUDA code generation/compiling): 6.472886e-01s
       Import time 1.437664e-03s
       Node make_thunk time 6.340373e-01s
           Node GpuMaxAndArgmax{axis=(0, 1)}(<GpuArrayType<None>(float32, (False, False))>) time 4.961591e-01s
           Node GpuMaxAndArgmax{axis=(2, 4)}(<GpuArrayType<None>(float32, (False, False, False, False, False))>) time 5.261087e-02s
           Node GpuMaxAndArgmax{axis=(0, 2, 4)}(<GpuArrayType<None>(float32, (False, False, False, False, False))>) time 3.413916e-03s
           Node GpuMaxAndArgmax{axis=(2,)}(<GpuArrayType<None>(float32, (False, False, False, False, False))>) time 3.237009e-03s
           Node GpuMaxAndArgmax{axis=(0, 2, 3, 4)}(<GpuArrayType<None>(float32, (False, False, False, False, False))>) time 2.639055e-03s

Time in all call to theano.grad() 0.000000e+00s
Time since theano import 38.656s
Class
---
  81.6%    81.6%       8.521s       3.55e-01s     C       24      24   theano.gpuarray.reduction.GpuMaxAndArgmax
  17.5%    99.0%       1.825s       7.61e-02s     C Py      24      24   theano.tensor.basic.MaxAndArgmax
   1.0%   100.0%       0.101s       2.09e-03s     C       48      48   theano.gpuarray.basic_ops.HostFromGpu
   ... (remaining 0 Classes account for   0.00%(0.00s) of the runtime)
```

notoraptor:
After installing Python 3 (Anaconda) and run test_reduction on it, I remarked that Python C-API has changed from Python 2 to Python 3 : integers are no longer manipulated with PyInt_* functions, but with PyLong_* functions instead. I use theses functions to access axis param op in C code.

So, I have to update the C code to make it compatible with both Python 2 and Python 3. I think I will convert axis to a numpy array in Python __init__ code and then use nump C-API (PyArray_* functions) to access axis in  C code.

notoraptor:
New update !

To ensure compatibility in Python 2 and Python 3, I finalled just use a C macro that defines PyInt_AS_LONG as PyLong_AS_LONG when Python version is >= 3.

test_reduction has also been rewritten so that every tested tensor (except 0-D) contains approximately 10 000 000 values, to observe speed on GPU and CPU op on large data.

The following tests succeed in both Python 2 and Python 3 with last libgpuarray/pygpu installed:
```
nosetests -v theano/gpuarray/tests/test_opt.py
    # 1 error ("cuDNN required")
nosetests -v theano/tensor/nnet/tests/test_nnet.py
nosetests -v theano/tensor/tests/test_opt_uncanonicalize.py
nosetests -v theano/tensor/tests/test_basic.py
```
And:
With Python 2 :
```
CUDA_LAUNCH_BLOCKING=1 THEANO_FLAGS=floatX=float32,profile=True,profiling.ignore_first_call=True,profiling.destination=profile-python2.log nosetests --verbose --nocapture theano/gpuarray/tests/test_reduction.py
```
With Python 3:
```
CUDA_LAUNCH_BLOCKING=1 THEANO_FLAGS=floatX=float32,profile=True,profiling.ignore_first_call=True,profiling.destination=profile-python3.log nosetests --verbose --nocapture theano/gpuarray/tests/test_reduction.py
```

The profile files are append below as TXT files. Things work, but now GpuMaxAndArgmax is sometimes faster than MaxAndArgmax and sometimes slower. Some days ago (and before the rebasings) GPU op was generally faster than CPU op. till now, I don't know why this changed.


[profile-python3.log.txt](https://github.com/Theano/Theano/files/579072/profile-python3.log.txt)
[profile-python2.log.txt](https://github.com/Theano/Theano/files/579073/profile-python2.log.txt)

theano/tensor/basic.py
```diff
@@ -1330,28 +1302,25 @@ def c_code_cache_version(self):
...
+            node.inputs[0].type.broadcastable) if i not in self.axis)
         return [rval, rval]
 
     def R_op(self, inputs, eval_points):
         if eval_points[0] is None:
             return [None, None]
-        if not isinstance(inputs[1], theano.Constant):
+        if len(self.axis) != 1:
             raise ValueError(('R_op supported for arg_max only for '
                               'constant axis!'))
```
abergeron: The error message should probably be updated.

theano/tensor/opt_uncanonicalize.py
```diff
@@ -41,40 +41,30 @@
...
+@gof.local_optimizer([T.MaxAndArgmax])
 def local_max_and_argmax(node):
     """
     If we don't use the argmax, change it to a max only.
     """
-    if node.op == T._max_and_argmax:
+    if isinstance(node.op, T.MaxAndArgmax):
+        axis = node.op.get_params(node)
         if len(node.outputs[1].clients) == 0:
             # MaxAndArgmax support variable axis,
```
abergeron: Remove the whole comment instead of appending to it.

abergeron:
Some minor details are left.  Hurray!

theano/gpuarray/opt.py
```diff
@@ -1775,6 +1776,14 @@ def typebuild(dtype, broadcastable, context_name=context_name):
                             context_name=context_name)
     return typebuild
 
+
+# Add optimization : maxandargmax (CPU -> GPU)
+@register_opt('fast_compile')
+@op_lifter([tensor.MaxAndArgmax])
+@register_opt2([tensor.MaxAndArgmax], 'fast_compile')
+def local_gpu_maxandargmax(op, context_name, inputs, outputs):
+    return GpuMaxAndArgmax(op.get_params(None))
```
nouiz: Should we disable the opt if axis is None? To don't have slowdown with this PR.

notoraptor:
Oh, a problem. I have just pushed a new update related to PCI bus id ( #5176  ), but as this current pull request has not yet been merged, the tests re-run ! ...

nouiz:
The extra commit problem is fixed and the PCI have its own PR.

PR #5172, opened 2016-11-03
[REG,BUG] Use _props_dict() from the right Op.

lamblin:
It was using the one from GpuFromHost instead of AdvancedIncSubtensor1.
This was caught by tests in DebugMode.

This issue was introduced on Sept. 19th by d0d006f44f3b857e160ffd60d45cce2f7cd8838e, should we add a warning?


abergeron:
I'm not against a new warning, but if you do, please remove the old one.  It's super old.

We should get rid of these warnings after a couple of years (and at least one release).

nouiz:
Merging now as the fix is good.

PR #5173, opened 2016-11-03
Fix some failing tests in debugmode

lamblin:


theano/tensor/tests/test_elemwise.py
```diff
@@ -870,6 +870,10 @@ def test_reduce_custom_dtype(self):
                         (topo, output_dtype)
                     data = numpy.random.rand(3, 4) * 10
                     data = data.astype(input_dtype)
+                    if output_dtype == 'float16' and method == 'prod':
+                        # We will likely get something infinite,
+                        # and DebugMode will complain.
+                        data = data[0:1]
```
nouiz: why not remove the "* 10" on line 871? This will make the numbers much smaller and won't need this special case. I tried it manually and removing it give very small numvers ~1e-6 or 1e-7. Using *2 make the result of the prod closer to 1. So I would go with *2

lamblin: If we remove `* 10`, then all numbers will be cast to zero if `input_dtype` is an integer dtype.
With `* 2`, still half of it would be 0, which makes the test for product pretty meaningless.
That's why I opted for fewer numbers, but more varied (especially when cast as int).

nouiz: ok. merging.

PR #5175, opened 2016-11-03
Fix grad of elemwise with boolean inputs.

### LINKED ISSUE #5171: Grad illegally returned an integer-valued variable
Hi!  Wanted to mention that a few tests broke in pymc3 when moving from 0.9.0.dev3 to 0.9.0.dev4

You can see some failures [here](https://travis-ci.org/pymc-devs/pymc3/jobs/172281533), but the gist is 

`TypeError: Elemwise{mul,no_inplace}.grad illegally  returned an integer-valued variable. (Input index 1, dtype bool)`.  

I'm looking at how to fix this in the pymc3 code -- it would be helpful to know if this is considered a bug, and if there's an easy fix!

abergeron:
Fixes #5171 and https://github.com/pymc-devs/pymc3/issues/1496

### DIFF
--- a/theano/tensor/elemwise.py
+++ b/theano/tensor/elemwise.py
@@ -668,8 +668,8 @@ second dimension
         # TODO: make sure that zeros are clearly identifiable
         # to the gradient.grad method when the outputs have
         # some integer and some floating point outputs
-        if False in [str(out.type.dtype) in theano.tensor.continuous_dtypes
-                     for out in outs]:
+        if any(out.type.dtype not in theano.tensor.continuous_dtypes
+               for out in outs):
             # For integer output, return value may
             # only be zero or undefined
             # We don't bother with trying to check
--- a/theano/tensor/elemwise.py
+++ b/theano/tensor/elemwise.py
@@ -9,6 +9,7 @@ from six.moves import xrange
 import theano
 from theano import gof
 from theano.compat import izip
+from theano.configparser import change_flags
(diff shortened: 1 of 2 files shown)


theano/tensor/elemwise.py
```diff
@@ -667,7 +668,7 @@ def grad(self, inputs, ograds):
         # TODO: make sure that zeros are clearly identifiable
         # to the gradient.grad method when the outputs have
         # some integer and some floating point outputs
-        if False in [str(out.type.dtype).find('int') == -1
+        if False in [str(out.type.dtype) in theano.tensor.continuous_dtypes
```
nouiz: This would be more clear:

    if any([out.type.dtype not in theano.tensor.continuous_dtypes for out in outs]):

otherwise a test would be great.

lamblin:
This could be a minimal test:
```python
x = theano.tensor.scalar(dtype='bool')
y = theano.tensor.bscalar()
z = x * y
dx, dy = theano.grad(z, [x, y])
```

nouiz:
bscalar() is an int8! when I tried to replicate the error, I expected it
was a bool.

lamblin:
Indeed, `bscalar` is a "byte", or `int8`, like in numpy: `np.typeDict['b']` gives `numpy.int8`.
You need an int and a bool to trigger the error, it did not happen with only bools.

ColCarroll:
Confirmed that this fixes pymc-devs/pymc3#1496 -- thanks!

PR #5179, opened 2016-11-04
Import numpy

kvmanohar22:
Changed imports of numpy to one common form of `import numpy as np` in code (only) in `Theano/benchmark/*` files

### DIFF
--- a/benchmark/convolution/conv2d.py
+++ b/benchmark/convolution/conv2d.py
@@ -17,16 +17,16 @@ if len(sys.argv)>6:
 
 setup="""
 import sys, timeit, time
-import numpy
+import numpy as np
 import theano, theano.tensor.signal.conv
 
 img_shape =  int(sys.argv[1]), int(sys.argv[2])
 ker_shape =  int(sys.argv[3]), int(sys.argv[4])
 dtype = sys.argv[5]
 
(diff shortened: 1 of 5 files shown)


benchmark/convolution/opencv.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, print_function, division
 import sys, timeit
-import numpy
+import numpy as np
```
nouiz: this will request change to the string in this file as the string contain code that will be executed.

This file isn't tested, this is why you didn't found this problem with the test.

benchmark/convolution/conv2d.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, print_function, division
 import sys, timeit, time
-import numpy
+import numpy as np
```
nouiz: idem, the numpy in string here is code that is executed. So they need to be modified. Also not tested.

nouiz:
ok to merge when travis pass.

kvmanohar22:
Should I create a new branch (after this PR is merged) for making further changes in other directories or continue with this branch?

nouiz:
I think it is better that you make a PR per directory. thanks

PR #5181, opened 2016-11-04
[REG, CRASH] fix crash related to float16 introduced in gh-5077

nouiz:
Fix gh-5077

### DIFF
--- a/theano/gpuarray/elemwise.py
+++ b/theano/gpuarray/elemwise.py
@@ -108,9 +108,8 @@ class GpuElemwise(HideC, Elemwise):
         # load in float16 and cast to float32 and do the reverse for
         # the output.
         scalar_op = self.scalar_op
-        if isinstance(scalar_op, Composite):
-            s = scalar_op.clone_float32()
-            scalar_op = s
+        if isinstance(scalar_op, (scalar.Cast, Composite)):
+            scalar_op = scalar_op.clone_float32()
         fake_node = scalar_op.make_node(*scal_v_ins)
         scal_v_out = fake_node.outputs
         assert len(scal_v_out) == len(node.outputs)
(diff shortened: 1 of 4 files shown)


theano/gpuarray/elemwise.py
```diff
@@ -108,15 +103,25 @@ def _generate_op_string(self, node):
...
         assert len(scal_v_out) == len(node.outputs)
 
         kop = fake_node.op.c_code(fake_node, 'elem_scalar',
                                   inps, outs,
                                   dict(fail='return;'))
 
-        assert 'npy_float16' not in kop
+        if isinstance(scalar_op, Composite):
+            # Other op like cast should handle float16 themself.
+            assert 'npy_float16' not in kop
```
abergeron: We need to handle float16 here, otherwise we will get bad results again.  There is no special handling for it in the scalar ops.

So if the assert triggered we can't just patch over it like this.

nouiz: So if you had just a cast op before this PR, how it was working? It didn't had special code to handle that before to my understanding. I think it still work (I'll add a test about that)

nouiz:
I updated the PR to handle the cast correctly and keep the assert to make sure no other op miss special handling.

theano/scalar/basic.py
```diff
@@ -2220,6 +2220,11 @@ def __init__(self, o_type, name=None):
     def __str__(self):
         return '%s{%s}' % (self.__class__.__name__, self.o_type.dtype)
 
+    def clone_float32(self):
+        if self.o_type == float16:
+            return identity
```
nouiz: It was working, but I still changed it and added test for cast on/to int.

Issue #850, opened 2012-08-13
Elemwise tester should test uint* dtype.


nouiz:
There was a bug recently found that those op output was not upcasted to float\* when needed!

benathi:
Could you elaborate the issue further? From what I infer, some element-wise operations are supposed to cast the output to float* but they currently don't do that so it's an issue for some uint* input? If I understand correctly, what operations are these? For example,I looked at the test_elemwise.py and test_prod_without_zeros function; I modified the input with dtype=uint32. The output elements seem to be of type float64 . (which is the desirable behavior I presume)

I'm quite a beginner for theano so appreciate any additional clarifications are appreciated.

nouiz:
The ticket is about adding tests for uint* dtype for elemwise. The tests are done in part with a test generator in the file: theano/tensor/tests/test_basic.py

For example, grep for "AddTester". If you have any questions, do not hesitate.

benathi:
So the ticket is to create more tests like 'AddTester' or for other element wise operators but for input being array of type uint*? Is this correct?

nouiz:
Yes. Just to be sure, do not create new AddTester, add other dtype test
case in the existing one.

nouiz:
In dict of tensor/tests/test_basic.py in dict like _good_broadcast_binary_normal and a new key uint that would test for uint16

PR #5182, opened 2016-11-04
uint16 added into _good_broadcast_binary_normal

mohammadpz:
Hi @nouiz ,

I added binary tests in ```_good_broadcast_binary_normal``` for uint16. The tests seem to be fine. Is there any other place that I need to add uint?

Thanks

fix gh-850

### DIFF
--- a/theano/tensor/nnet/tests/test_sigm.py
+++ b/theano/tensor/nnet/tests/test_sigm.py
@@ -44,7 +44,8 @@ UltraFastSigmoidTester = makeBroadcastTester(
     op=ultra_fast_sigmoid,
     expected=upcast_int8_nfunc(lambda inputs: check_floatX(
         inputs, 1 / (1 + numpy.exp(-inputs)))),
-    good=_good_broadcast_unary_normal_no_complex,
+    good=copymod(_good_broadcast_unary_normal_no_complex,
+                 without=['uint16']),  # numpy fucnting overflows with uint16.
     # grad=_grad_broadcast_unary_normal,
     name='UltraFastSigmoidTester',
     # This is an approx of the sigmoid. That is why we raise eps
@@ -54,7 +55,8 @@ HardSigmoidTester = makeBroadcastTester(
     op=hard_sigmoid,
     expected=upcast_int8_nfunc(lambda inputs: check_floatX(
         inputs, 1 / (1 + numpy.exp(-inputs)))),
-    good=_good_broadcast_unary_normal_no_complex,
+    good=copymod(_good_broadcast_unary_normal_no_complex,
+                 without=['uint16']),  # numpy fucnting overflows with uint16.
     # grad=_grad_broadcast_unary_normal,
     name='HardSigmoidTester',
     # This is an approx of the sigmoid. That is why we raise eps
@@ -65,7 +67,9 @@ SoftplusTester = makeBroadcastTester(
     op=softplus,
     expected=upcast_int8_nfunc(lambda inputs: check_floatX(
         inputs, numpy.log1p(numpy.exp(inputs)))),
-    good=dict(_good_broadcast_unary_normal_no_complex,
+    good=dict(copymod(_good_broadcast_unary_normal_no_complex,
+                      without=['uint8', 'uint16']),  # numpy fucnting overflows with uint16.
+              uint8=[numpy.arange(0, 89, dtype='uint8')],  # the range is different in new added uint8.
               int8=[numpy.arange(-127, 89, dtype='int8')]),
     # grad=_grad_broadcast_unary_normal,
     name='SoftplusTester',
--- a/theano/tensor/nnet/tests/test_sigm.py
+++ b/theano/tensor/nnet/tests/test_sigm.py
@@ -16,7 +16,7 @@ from theano.tensor.nnet.sigm import (
     compute_mul, is_1pexp, parse_mul_tree, perform_sigm_times_exp,
     register_local_1msigmoid, simplify_mul,
 )
-from theano.tensor.tests.test_basic import (makeBroadcastTester,
+from theano.tensor.tests.test_basic import (makeBroadcastTester, copymod,
                                             check_floatX, upcast_int8_nfunc,
                                             _good_broadcast_unary_normal_no_complex)
 
@@ -28,12 +28,14 @@ class T_sigmoid(unittest.TestCase):
     def test_elemwise(self):
         utt.verify_grad(sigmoid, [numpy.random.rand(3, 4)])
 
-
 SigmoidTester = makeBroadcastTester(
     op=sigmoid,
     expected=upcast_int8_nfunc(lambda inputs: check_floatX(
         inputs, 1 / (1 + numpy.exp(-inputs)))),
-    good=_good_broadcast_unary_normal_no_complex,
+    good=copymod(_good_broadcast_unary_normal_no_complex,
+                 without=['uint16']),  # The reason that 'uint16' is excluted is that
+                                       # theano works well but numpy overflows resulting
+                                       # in an assertion error.
     # grad=_grad_broadcast_unary_normal,
     name='SigmoidTester',
 )
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -1128,8 +1128,8 @@ _good_broadcast_unary_normal_no_complex = dict(
         normal=[numpy.asarray(rand_ranged(-5, 5, (2, 3)), dtype=floatX)],
         integers=[randint_ranged(-5, 5, (2, 3))],
         int8=[numpy.arange(-127, 128, dtype='int8')],
-        uint8=[numpy.arange(0, 255, dtype='uint8')],
-        uint16=[numpy.arange(0, 65535, dtype='uint16')],
+        uint8=[numpy.arange(0, 89, dtype='uint8')],
+        uint16=[numpy.arange(0, 89, dtype='uint16')],
         corner_case=[corner_case],
         empty=[numpy.asarray([], dtype=config.floatX)],
         )
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -381,7 +381,7 @@ def makeTester(name, op, expected, checks=None, good=None, bad_build=None,
                     expecteds = self.expected(*inputs)
                     eps = 1e-10
 
-                if any([i.dtype in ('float32', 'int8', 'uint8')
+                if any([i.dtype in ('float32', 'int8', 'uint8', 'uint16')
                         for i in inputs]):
                     eps = 1e-6
                 eps = numpy.max([eps, _eps])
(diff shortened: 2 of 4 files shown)


nouiz:
This is a good start. Can you try to add them to those other places?

thanks.

_good_broadcast_div_mod_normal_float_no_complex = dict(
_good_inv = dict(
_good_broadcast_unary_normal = dict(
_good_broadcast_unary_normal_no_complex = dict(
ExpTester
Expm1Tester
_good_broadcast_unary_wide = dict(
_good_broadcast_unary_arcsin = dict(
_good_broadcast_unary_tan = dict(
_good_broadcast_binary_arctan2 = dict(
CoshTester
SinhTester
_good_broadcast_unary_arctanh = dict(
ClipTester

maybe:
 _good_broadcast_unary_gammaln = dict(
_good_broadcast_unary_psi = dict(
_good_broadcast_unary_chi2sf = dict(

ALL_DTYPES = ('int8', 'int16', 'int32', 'int64',
              'float32', 'float64', 'complex64', 'complex128')
REAL_DTYPES = ALL_DTYPES[:-2]
COMPLEX_DTYPES = ALL_DTYPES[-2:]

mohammadpz:
Hi @nouiz ,
Travis jobs were killed once more time. I changed a few lines and commit it again to restart the tests again.

mohammadpz:
I think ```_good_broadcast_binary_arctan2``` in ```_good_broadcast_binary_arctan2```  does not seem to support ```uint```.

mohammadpz:
Hi again, 
When compiling the following function:
```python
inp = tensor.vector(dtype='uint8')
out = tensor.cast(inp, dtype='complex64')
f = function([inp], out)
```
I get the following error:
```
*** ImportError: ('The following error happened while compiling the node', 
Elemwise{Cast{complex64}}(<TensorType(uint8, vector)>), '\n',
'dlopen(/Users/mohammad/.theano/compiledir_Darwin-16.0.0-x86_64-i386-64bit-i386-2.7.12-
64/tmpxuWFCR/7ecec08e2d13b42b6890013bd3da78f8.so, 2): Symbol not found: 
__ZN16theano_complex64aSIhEERS_RKT_\n  Referenced from: 
/Users/mohammad/.theano/compiledir_Darwin-16.0.0-x86_64-i386-64bit-i386-2.7.12-
64/tmpxuWFCR/7ecec08e2d13b42b6890013bd3da78f8.so\n  Expected in: flat namespace\n in 
/Users/mohammad/.theano/compiledir_Darwin-16.0.0-x86_64-i386-64bit-i386-2.7.12-
64/tmpxuWFCR/7ecec08e2d13b42b6890013bd3da78f8.so', '[Elemwise{Cast{complex64}}
(<TensorType(uint8, vector)>)]')```

**Comment by mohammadpz** [comment]
I noticed something in ```SigmoidTester``` at ```theano/tensor/nnet/tests/test_sigm.py```.
The ```expected``` function that tests for correctness of the theano output using numpy, overflows with ```unit16``` and makes the tests fail. So it is a problem of numpy indeed. We can cast the input to float only for the numpy function. What do you think @nouiz ?

File: theano/tensor/tests/test_basic.py [hunk 1]
```diff
@@ -1577,10 +1577,6 @@ def test_py_c_match():
     integers=(randint(2, 3), randint(2, 3)),
     int8=[numpy.arange(-127, 128, dtype='int8'),
           numpy.arange(-127, 128, dtype='int8')[:, numpy.newaxis]],
-    uint8=[numpy.arange(0, 255, dtype='uint8'),
-           numpy.arange(0, 255, dtype='uint8')[:, numpy.newaxis]],
-    uint16=[numpy.arange(0, 65535, dtype='uint16'),
-            numpy.arange(0, 65535, dtype='uint16')[:, numpy.newaxis]],
```
nouiz: I tested it rapidly in numpy and it work. I think the problem is that numpy will return float16 for this case and not float32 for uint8. But why did you remove uint16 that return float32?

>>> np.arctan2(np.asarray([1], dtype='uint8'), np.asarray([1], dtype='uint8'), )
array([ 0.78515625], dtype=float16)

nouiz: About the uint8, ideally, we should fix Theano for it, but if we make uint16 work, we could make a new issue about making arctan2 work with uint8/uint8.

theano/tensor/tests/test_basic.py
```diff
@@ -1892,10 +1886,7 @@ def test_py_c_match():
     normal=(rand_ranged(1, 10, (2, 3)),
             numpy.asarray(1, dtype=config.floatX)),
     empty=(numpy.asarray([], dtype=config.floatX),
-           numpy.asarray(1, dtype=config.floatX)),
-    # The test seems to fail!!
-    int=(randint_ranged(1, 10, (2, 3)),
-            numpy.asarray(1, dtype=config.floatX)))
+           numpy.asarray(1, dtype=config.floatX)))
```
nouiz: What error did you got with this? scipy.stats.chi2.sf() work with uint8 and int8 input.

theano/tensor/tests/test_basic.py
```diff
@@ -1817,10 +1817,7 @@ def test_py_c_match():
 
 _good_broadcast_unary_gammaln = dict(
     normal=(rand_ranged(-1 + 1e-2, 10, (2, 3)),),
-    empty=(numpy.asarray([], dtype=config.floatX),),
-    int=(randint_ranged(1, 10, (2, 3)),),
-    uint8=(randint_ranged(1, 10, (2, 3)).astype('uint8'),),
-    uint16=(randint_ranged(1, 10, (2, 3)).astype('uint16'),),)
+    empty=(numpy.asarray([], dtype=config.floatX),))
```
nouiz: Why all the ints are removed? What was the error? scipy support that.

theano/tensor/tests/test_basic.py
```diff
@@ -1863,10 +1860,7 @@ def test_py_c_match():
 
 _good_broadcast_unary_psi = dict(
     normal=(rand_ranged(1, 10, (2, 3)),),
-    empty=(numpy.asarray([], dtype=config.floatX),),
-    int=(randint_ranged(1, 10, (2, 3)),),
-    uint8=(randint_ranged(1, 10, (2, 3)).astype('uint8'),),
-    uint16=(randint_ranged(1, 10, (2, 3)).astype('uint16'),),)
+    empty=(numpy.asarray([], dtype=config.floatX),))
```
nouiz: idem, why? What was the error?

nouiz:
The goal of that PR is to add uint tests. As we already tests with int8 and special case it to upcast to float32, I think for that SigmoidTester, you could just test in uint32 instead of uint16.

mohammadpz:
@nouiz , numpy overflows even in the case of ```uint32```:
```python
x = np.array([1.0], dtype='uint32')
print 1.0 / (1 + np.exp(-x))
```
```shell
>> /usr/local/bin/ipython:1: RuntimeWarning: overflow encountered in exp
array([ 0.])
```
Theano's output is actually correct (```0.73```). So what do you think we should do?

nouiz:
Can you try uint64? If that work, add a comment that THeano work with lower dtype, but numpy overflow.

mohammadpz:
@nouiz , No! It does not work with ```uint64``` either.

theano/tensor/nnet/tests/test_sigm.py
```diff
@@ -28,12 +28,14 @@ def setUp(self):
...
         utt.verify_grad(sigmoid, [numpy.random.rand(3, 4)])
 
-
 SigmoidTester = makeBroadcastTester(
     op=sigmoid,
     expected=upcast_int8_nfunc(lambda inputs: check_floatX(
         inputs, 1 / (1 + numpy.exp(-inputs)))),
-    good=_good_broadcast_unary_normal_no_complex,
+    good=copymod(_good_broadcast_unary_normal_no_complex,
+                 without=['uint16']),  # The reason that 'uint16' is excluted is that
```
mohammadpz: @nouiz , note that here I exclude ```uint16``` test.

theano/tensor/tests/test_basic.py
```diff
@@ -1683,7 +1718,7 @@ def test_py_c_match():
     expected_gamma = scipy.special.gamma
     expected_gammaln = scipy.special.gammaln
     expected_psi = scipy.special.psi
-    expected_chi2sf = lambda x, df: scipy.stats.chi2.sf(x, df).astype(x.dtype)
+    expected_chi2sf = lambda x, df: scipy.stats.chi2.sf(x, df)
```
mohammadpz: I don't know why previously the results were cast to ```x.dtype```. I removed it simply. Please check this out.

theano/tensor/tests/test_basic.py
```diff
@@ -2037,7 +2086,7 @@ def test_py_c_match():
 def _numpy_second(x, y):
     return numpy.broadcast_arrays(x, y)[1]
 
-ALL_DTYPES = ('int8', 'int16', 'int32', 'int64',
+ALL_DTYPES = ('int8', 'int16', 'int32', 'int64', 'uint8', 'uint16',
               'float32', 'float64', 'complex64', 'complex128')
 REAL_DTYPES = ALL_DTYPES[:-2]
 COMPLEX_DTYPES = ALL_DTYPES[-2:]
```
mohammadpz: However, I changed it.

mohammadpz:
There is still another error in Chi2SFTester as it returns float64 instead of float32. I will try to fix that during NIPS.

theano/tensor/tests/test_basic.py
nouiz: Add this back. This is what make travis fail. With that small changes we should be able to merge this PR.

The problem is that we don't use the same output dtype as scipy for this. They upcast the output to float64, but we force the output to be the same as the input. This is why we needed that small change.

thanks

mohammadpz:
Hi Fred,

I'm in Iran for vacation and cannot easily use git due to filtering. I'll
fix this as soon as I come back to Montreal.

Thanks,
Mohammad

On Dec 22, 2016 7:41 PM, "Frédéric Bastien" <notifications@github.com>
wrote:

*@nouiz* commented on this pull request.
------------------------------

In theano/tensor/tests/test_basic.py
<https://github.com/Theano/Theano/pull/5182>:

> @@ -1683,7 +1718,7 @@ def test_py_c_match():
     expected_gamma = scipy.special.gamma
     expected_gammaln = scipy.special.gammaln
     expected_psi = scipy.special.psi
-    expected_chi2sf = lambda x, df: scipy.stats.chi2.sf(x, df).astype(x.dtype)
+    expected_chi2sf = lambda x, df: scipy.stats.chi2.sf(x, df)

Add this back. This is what make travis fail. With that small changes we
should be able to merge this PR.

The problem is that we don't use the same output dtype as scipy for this.
They upcast the output to float64, but we force the output to be the same
as the input. This is why we needed that small change.

thanks

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Theano/Theano/pull/5182>, or mute the thread

nouiz:
thanks for the update.


On Thu, Dec 22, 2016 at 2:00 PM, Mohammad Pezeshki <notifications@github.com

theano/tensor/tests/test_basic.py
mohammadpz: Hi @nouiz , I just revert it back. Is there anything that I need to change?
Thanks!

nouiz:
I think all is good, we can merge when travis pass.

nouiz:
Finally, there was another problem in Theano related to chi2sf. scipy alreas return float64 for it, but we where using the normal upcast to float (float32 or float64) rules.

I added a commit to this PR tha should fix thix. @abergeron can you check my commit?

PR #5184, opened 2016-11-04
Add grad_scale op

YangZhang4065:
I added a new grad_scale op to scale or inverse gradient in back-propagation.

This op could be helpful if you want to
1, Implement adversarial optimization which requires gradient inverse:
https://arxiv.org/abs/1409.7495
https://arxiv.org/abs/1511.06434

2, Have relatively different learning rates across the graph.

### DIFF
--- a/theano/gradient.py
+++ b/theano/gradient.py
@@ -2067,7 +2067,7 @@ def grad_clip(x, lower_bound, upper_bound):
 class GradScale(ViewOp):
     def __init__(self,multiplier):
         self.multiplier=multiplier
-        
+
     def grad(self, args, g_outs):
         return [self.multiplier*g_out for g_out in g_outs]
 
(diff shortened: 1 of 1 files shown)


lamblin:
I'm not opposed to it, but maybe we should have a more general mechanism for defining or redefining the gradient of an Op, rather than a collection of Ops for all the things that we may want to do to gradients.
Scaling seems general enough (as well as blocking and clipping, which we already have), so I'm OK with this PR.

f0k:
> maybe we should have a more general mechanism for defining or redefining the gradient of an Op

+1

This comes up on the Lasagne mailing list every now and then. For guided backprop, I used OpFromGraph and replaced the `grad` method, but that's cumbersome (the subgraph inside OpFromGraph cannot interact with the outside graph any more in optimization, so you need to manually define it for the correct device, for example).
Maybe we can have a simpler way to define an Op from two functions (forward and backward pass). It would create a unique placeholder Op with a `grad` method and register an optimization that replaces the placeholder with the forward pass.

lamblin:
@f0k Agreed, that is indeed what we suggest in #5225.

f0k:
> Agreed, that is indeed what we suggest

Ah, great minds think alike :)

nouiz:
@YangZhang4065 Do you have time to finish this PR? It only need the flake8 fixes I comented aboce. Otherwise, it should be good.

thanks

nouiz:
continued in gh-5399

PR #5189, opened 2016-11-07
ROI Pooling Layer Op in CPU and GPU

Sentient07:
This PR consists of Yaroslav's version on the GPU and the implementation translated to C on the CPU.


aam-at:
For gpu ROI pooling you may find useful - https://github.com/ddtm/theano-roi-pooling

aam-at:
It seems that gpu op is missing in your pull request.

Sentient07:
Hi,
Yeah I still haven't completed this PR. I'm yet to add tests and GPU Op. I
became sick earlier this week and couldn't work. I'll try to finish this by
the weekend. Thank you for the suggestion :)

aam-at:
`dnn.py` for cudnn related functionality. You should move to `gpuarray/pool.py` (update to upstream). Also, consider renaming `roi_pool` to `pool_roi` as there are a bunch of files related to gpuarray pooling which starts with pool. So it will be organized nicely in the repository. Also there are no tests yet.

theano/gpuarray/roi_pool.c
```diff
@@ -0,0 +1,187 @@
...
+#include <float.h>
+
+#define max(a,b) (a>b?a:b)
+#define min(a,b) (a<b?a:b)
+
+void APPLY_SPECIFIC(ROIPoolForward)(
+    const int nchannels, const float* bottom_data,
+    const float spatial_scale, const int channels, const int height,
+    const int width, const int pooled_height, const int pooled_width,
+    const float* bottom_rois, float* top_data, float* argmax_data) {
```
aam-at: You should define kernel in `#section kernels`.

theano/gpuarray/roi_pool.c
```diff
@@ -0,0 +1,187 @@
...
+  dims[1] = channels;
+  dims[2] = POOLED_HEIGHT;
+  dims[3] = POOLED_WIDTH;
+
+  int count = batch_size * channels * POOLED_HEIGHT * POOLED_WIDTH;
+
+
+  APPLY_SPECIFIC(ROIPoolForward)(
+          count, data->devdata, SPATIAL_SCALE, channels, height, width,
+          POOLED_HEIGHT, POOLED_WIDTH, rois->devdata, (*out)->devdata);
```
aam-at: You should define kernel and launch it using '_scall`.

theano/gpuarray/roi_pool.c
```diff
@@ -0,0 +1,187 @@
...
+
+int APPLY_SPECIFIC(GPUBackward)(PyGpuArrayObject* data,
+                           PyGpuArrayObject* rois,
+                           PyGpuArrayObject* out_grad,
+                           PyGpuArrayObject** data_grad) {
+    int count = gpuarray_get_elsize(data);
+    int batch_size = PyGpuArray_DIMS(rois)[0];
+    int channels = PyGpuArray_DIMS(data)[1];
+    int height = PyGpuArray_DIMS(data)[2];
+    int width = PyGpuArray_DIMS(data)[3];
```
aam-at: Similarly to GpuForward, you should perform some checks:
- prepare and check data_grad
- check that input arrays are c-contiguous.

theano/gpuarray/dnn.py
```diff
@@ -2919,3 +2919,53 @@ def local_gpua_softmax_dnn_grad(op, ctx_name, inputs, outputs):
     out = GpuDnnSoftmaxGrad('accurate', 'instance')(
         gpu_contiguous(ins[0]), gpu_contiguous(ins[1]))
     return [out.dimshuffle(0, 2)]
+
+
+class GpuRoIPool(RoIPoolOp, COp):
```
aam-at: You should subclass from `CGpuKernelBase`, so you can use `#section kernels`.

theano/gpuarray/dnn.py
```diff
@@ -2919,3 +2919,53 @@ def local_gpua_softmax_dnn_grad(op, ctx_name, inputs, outputs):
...
+        roi_tuples = as_gpuarray_variable(roi)
+        assert feature_maps.ndim == 4
+        assert roi.ndim == 2
+        return Apply(self, [feature_maps, roi_tuples], [feature_maps.type()])
+
+    def c_code_cache_version(self):
+        return (1, 0)
+
+
+class GpuRoIPoolGradOp(RoIPoolGradOp, COp):
```
aam-at: Idem

theano/gpuarray/dnn.py
Sentient07: Yes, but I was not familiar with defining new kernels, so I did it this way after having a quick discussion with @nouiz . But I am not sure if this would work as I am facing some errors with Numpy's C- API

theano/gpuarray/roi_pool.c
Sentient07: Could you elaborate more on what you mean by launching?

Sentient07:
Hello @aam-at , thank you for your review. I still have a commit that I have which is yet to be pushed. I get some errors on the commit when I tried to use the Op to test the functionality.

Sentient07:
When I'm trying to access the `data` member of the struct `PyArrayObject`s, I get the error that `error: no member named 'data' in 'tagPyArrayObject'.`. I don't understand why `PyArrayObject` is getting converted into `tagPyArrayObject`. What could have gone wrong ?

Sentient07:
Also I'm not able to infer the context name of the input variables. 
```
import numpy as np
import theano
import theano.tensor as T

from theano.gpuarray.dnn import GpuRoIPool as ROIPoolingOp

op = ROIPoolingOp(pooled_h=2, pooled_w=2, spatial_scale=1.0)

t_data = T.ftensor4()
t_rois = T.fmatrix()

t_outs = op(t_data, t_rois)
t_c = t_outs[0].sum()

t_g_data = T.grad(t_c, t_data)[0]

f = theano.function([t_data, t_rois], t_outs + [t_g_data])
```
When I use this code for testing, since the variables don't have owners, and of course they're not on the GPU, the context name couldn't be inferred. It makes sense of the fact that the pooling layer will have inputs from other layer which would be already on the GPU and inferring context name the way I am doing wouldn't be a problem. But for writing unit tests, I am not sure how to handle this.

Sentient07:
Also, @aam-at there is no `pool` file in `gpuarray` (new backend) and `GpuDnnPool` has been implemented in `dnn.py`. It would be a good idea to move it to a new file, but i don't want to do that in this PR

theano/gpuarray/dnn.py
```diff
@@ -2919,3 +2920,72 @@ def local_gpua_softmax_dnn_grad(op, ctx_name, inputs, outputs):
...
+    def __init__(self, pooled_h, pooled_w, spatial_scale):
+        super(GpuRoIPool, self).__init__(pooled_h, pooled_w, spatial_scale)
+        self.pooled_h = pooled_h
+        self.pooled_w = pooled_w
+        self.spatial_scale = spatial_scale
+
+    def make_node(self, feature_maps, roi):
+        ctx_name = infer_context_name(feature_maps, roi)
+        feature_maps = as_gpuarray_variable(feature_maps, ctx_name)
+        roi_tuples = as_gpuarray_variable(roi, ctx_name)
```
nouiz: That is the right way to infer the context. So I don't understand the problem you have here.

nouiz:
Now there is a gpuarray/pool.py file in Theano. If you rebase your PR, you
will see it.

Sentient07:
Hi @nouiz @lamblin ! These are the errors that I am facing right now. 
I try to compile a basic function with this Op, but the function returns error with the C code, which I understand that I'm initializing a float to an integer. 
The code that I try to compile is : http://pastebin.com/pYgeNfhE 
The full traceback : http://pastebin.com/3dA9uVUx 
I have made the changes suggested in the previous review. I am not able to figure out where this problem is happening. I went through once fully, but I don't think I see anything wrong (or different from the CPU Op). Once this is fixed, I will add the tests

nouiz:
Don't wait for this to be fixed to start the test. It is useful to start the tests early even if the code isn't complete.

Sentient07:
The full traceback of the error that I reported, 
http://pastebin.com/3MMTdFnQ

nouiz:
To get a better error message try this diff:

diff --git a/theano/gof/vm.py b/theano/gof/vm.py
index 812cc34..0f5af12 100644
--- a/theano/gof/vm.py
+++ b/theano/gof/vm.py
@@ -1064,8 +1064,10 @@ class VM_Linker(link.LocalLinker):
                     # If this member isn't present, it will crash later.
                     thunks[-1].lazy = False
             except Exception as e:
+#                e.args = ("The following error happened while"
+#                          " compiling the node", node, "\n") + e.args
                 e.args = ("The following error happened while"
-                          " compiling the node", node, "\n") + e.args
+                          " compiling the node %s\n%s" % (node, e.args),)
                 raise
         t1 = time.time()



If it don't work, try this one if you can't find how to fix my first diff.

diff --git a/theano/gof/vm.py b/theano/gof/vm.py
index 812cc34..df3be4e 100644
--- a/theano/gof/vm.py
+++ b/theano/gof/vm.py
@@ -1064,8 +1064,6 @@ class VM_Linker(link.LocalLinker):
                     # If this member isn't present, it will crash later.
                     thunks[-1].lazy = False
             except Exception as e:
-                e.args = ("The following error happened while"
-                          " compiling the node", node, "\n") + e.args
                 raise
         t1 = time.time()

theano/gpuarray/roi_pool.c
```diff
@@ -0,0 +1,220 @@
...
+            // Force malformed ROIs to be 1x1
+            ga_size roi_width = fmax((DTYPE_i0) (roi_end_w - roi_start_w + 1) * 1.0, 1);
+            ga_size roi_height =  fmax((DTYPE_i0) (roi_end_h - roi_start_h + 1) * 1.0, 1);
+
+            DTYPE_i0 bin_size_h = (DTYPE_i0)roi_height
+                             / (DTYPE_i0)pooled_height;
+            DTYPE_i0 bin_size_w = (DTYPE_i0)roi_width
+                             / (DTYPE_i0)pooled_width;
+
+            ga_size phstart = floor((DTYPE_i0)(h - roi_start_h) / bin_size_h);
```
meowklaski: You need to subtract/add a small epsilon value within the floor/ceiling function to enure that boundary cases do not fail due to numerical precision.

theano/gpuarray/roi_pool.c
```diff
@@ -0,0 +1,220 @@
...
+        DTYPE_i0 bin_size_w = (DTYPE_i0)(roi_width)
+                           / (DTYPE_i0)(pooled_width);
+
+        ga_size hstart = (ga_size)(floor((DTYPE_i0)ph * bin_size_h));
+        ga_size wstart = (ga_size)(floor((DTYPE_i0)pw * bin_size_w));
+        ga_size hend = (ga_size)(ceil((DTYPE_i0)(ph + 1)
+                                         * bin_size_h));
+        ga_size wend = (ga_size)(ceil((DTYPE_i0)(pw + 1)
+                                         * bin_size_w));
+
```
meowklaski: You need to subtract/add a small epsilon value within the floor/ceiling function to ensure that boundary cases do not fail due to numerical precision.

theano/gpuarray/roi_pool.c
```diff
@@ -0,0 +1,220 @@
...
+            DTYPE_i0 bin_size_h = (DTYPE_i0)roi_height
+                             / (DTYPE_i0)pooled_height;
+            DTYPE_i0 bin_size_w = (DTYPE_i0)roi_width
+                             / (DTYPE_i0)pooled_width;
+
+            ga_size phstart = floor((DTYPE_i0)(h - roi_start_h) / bin_size_h);
+            ga_size phend = ceil((DTYPE_i0)(h - roi_start_h + 1) / bin_size_h);
+            ga_size pwstart = floor((DTYPE_i0)(w - roi_start_w) / bin_size_w);
+            ga_size pwend = ceil((DTYPE_i0)(w - roi_start_w + 1) / bin_size_w);
```
meowklaski: Subtract/add epsilon here as well

theano/tensor/signal/roi_pool.c
```diff
@@ -0,0 +1,240 @@
...
+  dims[0] = batch_size;
+  dims[1] = channels;
+  dims[2] = POOLED_HEIGHT;
+  dims[3] = POOLED_WIDTH;
+  int total_ndim = 4;
+  int count = batch_size * channels * POOLED_HEIGHT * POOLED_WIDTH;
+  int mem_nc;
+  mem_nc = 0;
+  // Checking if contiguous
+  if(!PyArray_ISCONTIGUOUS(data) || !PyArray_ISCONTIGUOUS(rois)){
```
nouiz: This if is at the wrong place. You are mixing the condition on the inputs with the code to init the output.

This are 2 orthogonals thigs. If data is contiguous, then it goes to line 114, but out isn't created, so this segfault.

Sentient07: Works! But when the input data are contiguous, don't we fill the output array with 0s which is the same process ? (I thought it was a cleaner process to use `PyArray_FILLWBYTE`  as it supports contiguous arrays)

theano/tensor/signal/tests/test_pool.py
```diff
@@ -1213,12 +1214,13 @@ def test_infer_shape(self):
...
+                             dtype='float32')
         rois = numpy.array([[0, 0, 0, 3, 3],
-                            [0, 0, 0, 7, 7]], dtype=theano.config.floatX)
+                            [0, 0, 0, 7, 7]], dtype='float32')
 
         self._compile_and_check([t_data, t_rois],
-                                [self.op(t_data, t_rois)],
+                                self.op(t_data, t_rois),
                                 [data, rois],
                                 self.op_class)
```
Sentient07: @nouiz this is a duplicate of your commit to this branch and it always happens when I pull from the origin. Is there anyway I can avoid duplicating commits ?

theano/gpuarray/ROIPoolGPUFwd.c
```diff
@@ -0,0 +1,120 @@
...
+    *out = (PyGpuArrayObject*) pygpu_zeros(4, dim, data->ga.typecode, GA_C_ORDER, ctx, Py_None);
+    *argmaxes = (PyGpuArrayObject*) pygpu_zeros(4, dim, data->ga.typecode, GA_C_ORDER, ctx, Py_None);
+    if (*out == NULL || *argmaxes == NULL) {
+      PyErr_Format(PyExc_ValueError, "Could not allocate output storage");
+      return 1;
+    }
+  }
+
+  err = ROIPoolGPUFwd_kernel_scall(1, &address, 0,
+          num_rois, data->ga.data, SPATIAL_SCALE, channels, height, width,
```
Sentient07: @nouiz I have a problem here. When i try to print SPATIAL_SCALE here, it is 1. or whatever the value I pass from the tests. But in the kernel, that is always 0.00. Could you please help me in finding the fault ?

Sentient07: Also, if I am trying to print the SPATIAL_SCALE variable after line 106,  (after preparing the output pointers), I run into segmentation fault. I don't understand why this is happening.

theano/gpuarray/ROIPoolGPUFwd.c
```diff
@@ -0,0 +1,120 @@
+#section kernels
+
+#kernel ROIPoolGPUFwd_kernel : ssize, *, ssize, ssize, ssize, ssize, ssize, ssize, *, *, * :
+KERNEL void ROIPoolGPUFwd_kernel(
+    const ga_int num_rois, GLOBAL_MEM DTYPE_i0 *bottom_data,
```
Sentient07: I have kept it as ga_int and not the traditional ga_size as the comparison (max and min function) raises an error as i'd be doing the comparison of an unsigned int with a signed integer

Sentient07:
There is one problem with this PR. 

- The variable SPATIAL_SCALE seems to be always 0.0 inside the kernel. But in section `support_code_struct`, I could see its value to be the desired one. Also, I can see the theano's autocompleted line where it defines SPATIAL_SCALE to the expected value using #define

theano/gpuarray/pool.py
```diff
@@ -411,15 +412,102 @@ def make_node(self, inp, eval_point, ws, stride=None, pad=None):
 
         return Apply(self, [inp, eval_point, ws, stride, pad], [eval_point.type()])
 
+
+class GpuRoIPoolOp(CGpuKernelBase):
+
+    __props__ = ('spatial_scale', 'pooled_h', 'pooled_w')
```
nouiz: Put the props in the same order as the parameter in \__init__. Idem for the second class.

theano/gpuarray/pool.py
```diff
@@ -411,15 +412,102 @@ def make_node(self, inp, eval_point, ws, stride=None, pad=None):
...
+    _f16_ok = True
```
nouiz: The code currently don't support float16 storage. So remove that. idem for the second class.

For now, put float16 aside. First make it work in float32/64.

theano/gpuarray/ROIPoolGPUFwd.c
```diff
@@ -1,9 +1,9 @@
 #section kernels
 
-#kernel ROIPoolGPUFwd_kernel : ssize, ssize, *, ssize, ssize, ssize, ssize, ssize, ssize, *, *, * :
+#kernel ROIPoolGPUFwd_kernel : int32, int32, *, float32, int32, int32, int32, int32, int32, *, *, * :
```
Sentient07: Thank you so much for this change. Are these options for variable type documented somewhere? I would like to add it otherwise :)

theano/gpuarray/ROIPoolGPUFwd.c
```diff
@@ -97,26 +97,38 @@ int APPLY_SPECIFIC(ROIPoolGPUFwd)(PyGpuArrayObject *data,
         return 1;
     }
 
-    if (*out != NULL || *argmaxes != NULL || (!vector_same_shape(data, *out)) || (!vector_same_shape(data, *argmaxes))){
+  if (*out == NULL || !vector_same_shape(data, *out)){
     Py_XDECREF(*out);
-    Py_XDECREF(*argmaxes);
```
Sentient07: Why this change ? Shouldn't both the outputs be decref'ed ?

nouiz: You should handle both outputs individually. Maybe only one or the other is preallocated, maybe both, maybe none.

theano/gpuarray/ROIPoolGPUFwd.c
nouiz: I don't think so, but @abergeron is/should work on the doc "shortly". So I think it is better to not duplicate work/do incompatible changes. Also this PR fix something to make this usable. Otherwise it wasn't usable.

theano/gpuarray/ROIPoolGPUFwd.c
Sentient07: Okay, will make the necessary change.

Sentient07:
I have a discrepancy in the way I am computing the region of interest. I have posted my question here. https://github.com/rbgirshick/caffe-fast-rcnn/issues/20 . The pythonic way of implementation made me realize this ambiguity.

Sentient07:
I realized the error with respect to gradient computation. I have to do either of the two, 
1) For the `grad` method of the `RoIPoolGradOp`, I should either return `DisconnectedType` for each inputs and add a `connection_pattern` method or `[tensor.zeros_like(i) for i in inputs]`. 
2) I have to implement a new `RoIPoolGradGradOp` and  have to compute the gradient for the output gradient using this class from the `grad` method of RoIPoolGradOp. 

2) seems like a more complicated way than 1) and with 1) I don't see any errors on the CPU implementation. @nouiz , with which approach should I proceed with ?  

@f0k , @benanne I will be glad for your opinions as well :)

f0k:
> I will be glad for your opinions as well :)

Hmm, why do you need a `grad` method for `RoIPoolGradOp`? Can't you omit that for now, so anybody trying to compute the gradient of the gradient of the RoIPoolOp will notice that the gradient is not implemented?

Sentient07:
Thank you for the review, Jan. Could you also please check the discrepancy
that I have mentioned in the previous comment if you have time? :)
Thanks,
Ramana

f0k:
Sorry, I currently don't have time to look at this... did you compare to the Theano ROI Pooling Op at https://github.com/ddtm/theano-roi-pooling/, both in terms of performance and results?

Sentient07:
I checked by comparing the code. That code is the same in caffe's repository. To test this functionality, and to write the python code for perform method, I conceptually translated the code. This python method gave me values that were different from the ones generated by the C code when the pooling window `(3,3)`.  Then I realized what was happening. 
In python, the maximum size of the region of interest is computed by, 
```
hend = np.ceil(hstart + col_length)
wend = np.ceil(wstart + row_length)
```
The caffe code wasn't the same as in C you need some typecasting work to ensure all are of same `dtype`. So the Caffe has it implemented like this, (also the same in the repository which you had mentioned in your comment)
```
int hend = static_cast<int>(ceil(static_cast<Dtype>(ph + 1) * bin_size_h));
int wend = static_cast<int>(ceil(static_cast<Dtype>(pw + 1) * bin_size_w));
hend = min(max(hend + roi_start_h, 0), height);
wend = min(max(wend + roi_start_w, 0), width);
```
In this implementation, I don't understand why `(ph + 1)` is done. It is mathematically correct when you expand what is inside brackets(and compare with what is given in the paper and the pythonic implementation), but that leads to an entirely different value when the pooling window dimension is `(3,3)`. 
Also since it makes sense to compute `hstart` with respect to `roi_start_h` and `hend` with respect to `hstart`, I have made the following change to the above code, 

```
int hstart = static_cast<int>(floor(static_cast<float>(ph) * bin_size_h)) + roi_start_h;
int wstart = static_cast<int>(floor(static_cast<float>(pw) * bin_size_w)) + roi_start_w;
int hend = static_cast<int>(ceil(bin_size_h)) + hstart;
int wend = static_cast<int>(ceil(bin_size_w)) + wstart;
// Clip to input boundaries
hstart = min(max(hstart, 0), height);
hend = min(max(hend, 0), height);
wstart = min(max(wstart, 0), width);
wend = min(max(wend, 0), width);
```
when the pooling dimension is (3,3), bin_size_h and bin_size_w become 2.667 respectively.  When the value of ph is 1(The value is incremented from a loop) `roi_start_h` is 0, the former method computes `hend` as  `ceil((1+1) * 2.667) ` and the latter computes `ceil(2.667) + 2`, which are 6, 5 respectively. The pythonic approach is in line with the latter, which is exact translation from the paper. Since I am not sure if my implementation is correct, I need someone to review this discrepancy :)

f0k:
> The pythonic approach is in line with the latter, which is exact translation from the paper. Since I am not sure if my implementation is correct, I need someone to review this discrepancy :)

Okay, just from your description and without looking at the paper myself, it seems like you may have discovered a discrepancy between the paper and code. The code in Caffe is by Ross Girshick (https://github.com/rbgirshick/fast-rcnn), the author of the Fast R-CNN and Faster R-CNN papers. Even if it is not matching the description in paper, it's the code that was used for the experiments. To be able to use existing pre-trained models, Theano would need the same implementation (unless the deviation is a corner case that usually doesn't occur). The only reasons to deviate from Ross' implementation would be (a) licensing issues, (b) you found a more efficient way, (c) your implementation gives better results. And even then, I guess it would be useful to have an option to use Ross' implementation if compatibility is required.
Can you benchmark your implementation against Ross', and/or try to replicate an experiment? https://github.com/ddtm/theano-roi-pooling/ looks like it's ready to use in Theano. Otherwise, the safest bet would probably be to just use Ross' implementation.

Sentient07:
Hi @f0k , 

> To be able to use existing pre-trained models, Theano would need the same implementation (unless the deviation is a corner case that usually doesn't occur). The only reasons to deviate from Ross' implementation would be (a) licensing issues, (b) you found a more efficient way, (c) your implementation gives better results. And even then, I guess it would be useful to have an option to use Ross' implementation if compatibility is required.

I understand when I can/have to  change the code from the original implementation. But, since the author's version always tends to give values greater than the one that my current implementation returns, the window over which maximum value is pooled for is bigger in their case and higher value is returned, so I doubt the result would be better. 
The tl;dr of the issue could be stated as, when `bin_size_h` is a fraction and its decimal place is more than `0.5`, the window over which maximum value is pooled for increases(Although it depends on the input image) and discrepancy occurs and hence I don't think this is a corner case. 

> https://github.com/ddtm/theano-roi-pooling/ looks like it's ready to use in Theano. Otherwise, the safest bet would probably be to just use Ross' implementation.

This is a code from the old backend. When I tried translating it into new backend and CPU, I had faced a lot of errors, so I resorted to implementing what has been done in a different way on both CPU and GPU. (The main error being, the kernel loop doesn't work properly. The variables `ph` and `pw` always seem to be 0). 

I see this is their testsuite, https://github.com/rbgirshick/caffe-fast-rcnn/blob/bcd9b4eadc7d8fbc433aeefd564e82ec63aaf69c/src/caffe/test/test_roi_pooling_layer.cpp . But I am not able to understand against what they check their results. If they have something that is hardcoded, i will refer to it and make my implementation aligned in that way.

> Can you benchmark your implementation against Ross', and/or try to replicate an experiment?

If I understand this correctly, Should I train an end to end network over the models that they have done and compare the test set accuracy(splitting according to the paper) ?

Sentient07:
Also, I found one more [implementation](https://github.com/yhenon/keras-spp), that is pythonic and this seems to do what my `perform` code does. I will fork and check the values returned and if there are any discrepancies.

Sentient07:
There is still a small error that I couldn't fix. When the image is not square shaped, there is a gradient mismatch at a particular location where the values differ by `0.2`. Apart from that, to me, everything works as expected. I spent the whole of this evening trying to debug that mismatch, but I couldn't. I'd be glad for some help with this.

Sentient07:
I'm not able to understand how the errors are happening in travis. There is one mismatch in the gradient computation, both on the CPU and on the GPU, and it happens with the penultimate shape in the list of image shapes that I'm testing, be it whatever value. I test the same image shape from outside of the loop and there is no gradient mismatch. Could this be because of caching of C code ? @nouiz when you have time could you please help me find what is going wrong here?  The way test is written is same as how it is written for Other pooling operations.

theano/gpuarray/pool.py
```diff
@@ -434,15 +435,102 @@ def make_node(self, inp, eval_point, ws, stride=None, pad=None):
...
+        argmaxes = as_gpuarray_variable(argmaxes, ctx_name)
+        out_grad = as_gpuarray_variable(out_grad, ctx_name)
+        assert data.ndim == 4
+        assert rois.ndim == 2
+        assert argmaxes.ndim == 4
+        assert out_grad.ndim == 4
+
+        return Apply(self, [data, rois, argmaxes, out_grad], [data.type()])
+
     def get_params(self, node):
```
Sentient07: OMG! That was a bad typo indeed. I will add it back. Sorry about it

theano/gpuarray/pool.py
```diff
@@ -434,15 +435,102 @@ def make_node(self, inp, eval_point, ws, stride=None, pad=None):
...
 
     def infer_shape(self, node, in_shapes):
-        ws, stride, pad = [node.inputs[2], node.inputs[3], node.inputs[4]]
-        shp = Pool.out_shape(in_shapes[0], ws, self.ignore_border, stride,
-                             pad, self.ndim)
-        return [shp]
+        return [in_shapes[0]]
+
+    def grad(self, inp, grads):
+        return [grad_undefined(self, i, inp[i]) for i in range(4)]
```
Sentient07: It is just a MaxPool Op over a specific region. If there exists gradient for `GpuMaxpoolGrad` why not for this? So I will change it to not implemented :)

Sentient07:
If I understand correctly, the problem occurs if there is a difference less than a threshold value(delta) between adjacent elements of the the maxpooled matrix ? Now it makes sense to me why there is a multiplication factor of `10.0` in MaxPooling operation. I had tried making it 10.0 earlier, but that didn't work. I had changed that multiplication factor to 100.0 and it works here. Would that be fine?

theano/gpuarray/ROIPoolGPUFwd.c
```diff
@@ -0,0 +1,140 @@
...
+bool vector_same_shape(PyGpuArrayObject* arr1, PyGpuArrayObject* arr2){
+  return (PyGpuArray_DIMS(arr1)[0] == PyGpuArray_DIMS(arr2)[0]);
+  }
+
+int APPLY_SPECIFIC(ROIPoolGPUFwd)(PyGpuArrayObject *data,
+                      PyGpuArrayObject *rois,
+                      PyGpuArrayObject **out,
+		              PyGpuArrayObject **argmaxes,
+                      PyGpuContextObject *ctx) {
+    size_t address = 1;
```
Sentient07: @nouiz here, I have made a hack here to work. I am not sure if this is the right way. I had implemented without the loop over all the dimensions of the image, that has been done in other Pooling kernels. I also have opened an issue as the documentation does not say anything about this variable, what value it has to be set to

Sentient07:
I have made the changes suggested before. Now, all the tests pass locally. Only issue is with how the kernel is called in the GPU support code. There are no errors, the implementation works as expected, but I think it is a hack. @nouiz could you please check that?

theano/gpuarray/ROIPoolGPUFwd.c
f0k: I think that's supposed to be the number of threads. Your kernel doesn't really make use of the GPU now, it's a single block with a single thread doing all of the work sequentially, in nested loops. The original implementation parallelizes over the output pixels (https://github.com/ddtm/theano-roi-pooling/blob/master/roi_pooling.cu#L200-L206). You should do the same. Check how the other caffe-to-gpuarray translations in Theano implement caffe's [CUDA_KERNEL_LOOP](https://github.com/ddtm/theano-roi-pooling/blob/master/roi_pooling.cu#L53-L58): https://github.com/Theano/Theano/blob/master/theano/gpuarray/pool.c#L13-L20

Sentient07: @f0k I first implemented similar to that. It gave the values of some variables to be 0, precisely the height and width. That is why I translated the CPU code to the GPU(with nested loops for each variable) instead of doing it like how Yaroslav did. I will give an other attempt and push it again so that you can see that problem. Also, when I set the variable to (number of threads) any value other than 1, I get mismatches between the expected and what it returns.

f0k: > I will give an other attempt and push it again so that you can see that problem.

Sorry, I think I won't have the time to try it myself -- I can only have a look at the code. Might still be worth it.

Sentient07: @f0k I understood the problem. In Caffe, I don't think they use batches, or in the output, they don't use a separate dimension to represent the ROI. The output differs per ROI, which means, my guess is that they don't take into account the batch_size. [Here](https://github.com/rbgirshick/caffe-fast-rcnn/blob/bcd9b4eadc7d8fbc433aeefd564e82ec63aaf69c/src/caffe/layers/roi_pooling_layer.cpp#L52), they initialize the batch_size, but they do not loop over it. And [here](https://github.com/rbgirshick/caffe-fast-rcnn/blob/bcd9b4eadc7d8fbc433aeefd564e82ec63aaf69c/src/caffe/layers/roi_pooling_layer.cpp#L60), there is a loop over the number of rois, and the output is assigned per each ROI. But looking at the implementation in theano that you [showed](https://github.com/ddtm/theano-roi-pooling/blob/master/roi_pooling.cu#L200), the number of rois isn't taken into account in computing the output dimension.  
So the problem is, 
1) In the version that you showed, the output is of the shape, [batch_ind, channel, pool_h, pool_w]
2) In my implementation and Caffe's C implementation, the output shape is [batch_ind, roi_ind, channel, pool_h * pool_w].

I'd be glad if you could help me with this. For now, this is the only discrepancy to my knowledge that is the reason for mismatches in implementing the kernel loop the way you suggested. 

Sorry for the delayed reply. I had to finish some work for my internship application and was busy for the past two days.

Sentient07: A correction to my previous comment. Actually the number of ROI is taken into account but not the batch size. I got confused with the nomenclature. Usually, batch_size is for batches of images, right ? Here, they have used it for number of ROIs that are used. From the Fast-RCNN paper, it is suggested that there are 128 ROI used per two images while training, from my understanding, it means batch_size is 2 and an image has 64 ROIs.

Sentient07: If we take into account the batch size, the output has to be a 5D tensor. That was why, I merged the pool_h and pool_w into a single dimension, adding batch(batch of images) as the fourth dimension. 
Now we need loop over both pooled _height and pooled_width separately to determine the starting height and width over which we do the maxpooling, similar to [here](https://github.com/ddtm/theano-roi-pooling/blob/master/roi_pooling.cu#L75-L82). After merging the two dimensions, we are left with `ph * pw` and not `ph` and `pw` separately.

Sentient07: Also, this PR needs some changes as there have been macro changes to kernels and numpy imports.

f0k: Looking at the [Caffe implementation](https://github.com/rbgirshick/caffe-fast-rcnn/blob/bcd9b4eadc7d8fbc433aeefd564e82ec63aaf69c/src/caffe/layers/roi_pooling_layer.cpp#L46), there are two inputs tensors: The image data (batchsize, channels, rows, columns) and the ROI definitions (num_rois, 5), where each ROI definition is a vector of 5 ints (batch_idx, x0, x1, y0, y1) that says which image and which region to take in the input. So there is no need to output a 5D tensor; the output is a 4D tensor of (num_rois, channels, pooled_rows, pooled_columns). You could have a 3D ROI definition of (batchsize, num_rois, 4) and a 5D output of (batchsize, num_rois, channels, pooled_rows, pooled_columns), but then you'd limit the implementation to have a fixed number of ROIs per image. The original implementation allows to have a variable number of regions per image, and a variable total number of regions -- whatever you get out of your region proposal algorithm (or network).

f0k: And a quick google search brings up Ross' presentation explaining his Caffe implementation of ROI pooling: http://tutorial.caffe.berkeleyvision.org/caffe-cvpr15-detection.pdf

Sentient07: Thank you for that tutorial link. It was very helpful when I ported the entire training to theano. (But I have done it only for fast-RCNN). From your first comment, I think I am in a confusion. Please correct my misunderstanding, 
If there is an input image of (128, 3, 224, 224) where my batch size is 128, in this case, from my understanding, shouldn't the output have same number of ROIs per image per batch? As in, if we maxpool with a specific window_size and stride, we return the pooled value per batch and per channel, and the first two dimension remains the same in output tensor as the input, shouldn't it be the same here but inclusive of num_rois, which is the number of regions of interest we need per image?

Sentient07: From your comment, I understand that the output shouldn't depend on the batch size of the input image, but rather the number of rois. But wouldn't the batch information be lost? As in, we take M regions from N images and return those M regions, our network would later carry on with those M regions and forget the information about number of images we took initially. This is where I fail to understand why. 
In the 8th slide, it says if the object proposal is N, the output has NK boxes, where I understood the K to be the batch_size.

f0k: > shouldn't the output have same number of ROIs per image per batch

It could, but it's not necessary. You could have an algorithm that proposes interesting regions for an image based on some interestingness threshold, so depending on the image, you would get more or fewer regions. But it seems that for the paper, he used 64 regions per image.

> our network would later carry on with those M regions and forget the information about number of images we took initially

Yes, that's the point. It should classify each region independently, without knowing what image it came from. That's why it makes sense to use num_rois as the new batchsize. We can still assign the classifications to the images afterwards because we know the ROI definition tensor (and in the special case of a fixed number of N regions per image, we know that we'll have N predictions for the first image, then N predictions for the second, and so on; but the network doesn't need to know that).

> In the 8th slide, it says if the object proposal is N, the output has NK boxes, where I understood the K to be the batch_size.

Possibly. It's not explained. But again, his layer implementation does not assume a fixed number of regions per image.

Sentient07: I understand now! So it proposes regions from images that are equal to num_rois * K (where K is the number of classes). Later they assign probability to how possible it could be an object of class K. I had always thought they compute this probability of N regions in K batch of images, my bad.
Now I have a confusion.  This might not be related to this PR, but from the paper in general. They propose regions around objects in the image and assign it to a class. But why do they consider N x K proposals in that case? Shouldn't it be only N x K* proposals, where K* is number of classes found in the image? Or probably could be that they're showing only the highly probable ones in the paper for simplicity ?

Sentient07: Anyway, it makes sense to me now why it is like that. I will push one big commit changing the Op and Grad Op's C code, python code and GPU code. Thank you so much for your help @f0k ! A PR on RoIPool Layer is coming soon to Lasagne :smile:

f0k: > They propose regions around objects in the image and assign it to a class. But why do they consider N x K proposals in that case?

They don't consider NxK proposals. Check Slide 9. There, they consider a single image, with N proposed regions. As I interpret it, only the *outputs* are then K probabilities for each of the N regions, and K object boxes for each of the N regions, where K is the number of classes.

> I will push one big commit changing the Op and Grad Op's C code, python code and GPU code.

I'd suggest to stick to Ross' implementation then, to be compatible. Or possibly add your different pooling region derivation as a second option? In any case, I guess you can parallelize the implementation in the same way it's done in Caffe.

Sentient07: > As I interpret it, only the outputs are then K probabilities for each of the N regions, and K object boxes for each of the N regions, where K is the number of classes.

You mean to say a box/region  is generated  in an image for a class that does not exist in the image,  with a negligible probability ? I also interpreted it that way, but my question is where in the image would it be(in the case probability is negligible). I think the way to figure that out is to plot the image with proposals before the rejection of some proposals are done based on probability.

> I'd suggest to stick to Ross' implementation then, to be compatible. 

Yes, I will do it that way for uniformity. I will keep this version that I did, alter with the code and see if the results are better. I am still not 100% convinced how this will work better (probably faster, but I'm not able to grasp why it would produce more accurate results )

f0k: > You mean to say a box/region is generated in an image for a class that does not exist in the image, with a negligible probability ?

Don't mix up boxes and regions! The network has to produce, for each class, a) the probability of an object of that class being present in the ROI the network sees, and b) a bounding box of an object of that class in the ROI. If the ground truth says that an object of that class is not actually present in the ROI, then I guess the loss for the bounding box regressor will be set to zero (note the bbox_loss_weights on slide 17). You'd have to check the paper for this.

Sentient07: > then I guess the loss for the bounding box regressor will be set to zero (note the bbox_loss_weights on slide 17)

Perfect! This explained my confusion. Now, to my knowledge, the output is still dependent on batch. In slide 17, it says the output for 64 ROIs per 2 image sample would be 128, meaning the fourth dimension should be `batch_size * num_roi` right? By that the batch data isn't lost, as I thought when writing the previous comment and neither is any information from any of the images as we take regions per each image in a batch. 
Also the output dimension should be `[num_rois * batch, channel, pool_h, pool_w]`, right?

Sentient07: Oh, wait! I made a mistake. I rushed through and didn't see the difference between `ROIDataLayer` and `RoIPoolLayer`. The former is for extracting ROIs from images, which is sent to ROI Pooling Layer. The fourth dimension of output of the RoIPoolLayer is now same as that of ROIDataLayer's output, which in our case is 128, right ?

Sentient07: One more confusion that I had got resolved. Where to pool from when K batches of images are given. The first element in the roi array tells us the `batch_id` of the image to pool from. This explains everything to me and the GPU code works. Thank you very much for your time and efforts, @f0k and sorry about my trivial questions.

f0k: > The first element in the roi array tells us the `batch_id` of the image to pool from.

That's precisely what I said in https://github.com/Theano/Theano/pull/5189#discussion_r110680241: "each ROI definition is a vector of 5 ints (batch_idx, x0, x1, y0, y1) that says which image and which region to take in the input".

> meaning the fourth dimension should be `batch_size * num_roi` right

No. In that case, `num_rois` is 128. The fact that the first 64 ROIs are from the first image and the second 64 ROIs are from the second image is of no importance to the network. It could have been 60 ROIs from the first and 68 ROIs from the second image instead. The implementation does not assume any relation of `num_rois` to `batchsize`. Don't get confused by the fact that Ross trained with a batchsize of 2 and with 64 ROIs per image, this is a special case from the perspective of the implementation.

Sentient07: >  Don't get confused by the fact that Ross trained with a batchsize of 2 and with 64 ROIs per image

Yep! I got that now, i didn't quite actually understand the significance of the first element in each ROI array and I had assumed it is the batch_id, like, 0 for the first ROI, 1 for the second and so on.. But later, when I saw how the code increments the image pointer before pooling, I realized what it actually is. :)

Sentient07: The realization came when I was writing the pythonic implementation. The way i used to slice the image array to get the region from where i have to apply a max pool was, 
` image_array[b_in, c, y1:y2, x1:x2]`
(where `b_in` is the batch id)
Later my confusion arose because I had thought it should be done like(across the region in all the images in a batch, which is why i thought it was absurd :stuck_out_tongue:  )
`image_array[:, c, y1:y2, x1:x2]`
The correct way of doing it is, (which I realized it early morning, when I made those series of comments)
```
interested_batchid = [j for j in [rois[i][0] for i in range(len(rois))]]
image_array[interested_batchid[k], cn , x1:x2, y1:y2]
```
Am i correct now ?

f0k: > Am i correct now ?

Well, yes, but `interested_batchid = [j for j in [rois[i][0] for i in range(len(rois))]]` is equivalent to `interested_batchid = rois[:, 0]`. But if you already get `x1, x2, y1, y2` for the current ROI, why don't you just get the `batchid` along with it?

Sentient07: But here, for representing that `interested_batchid` is dependent on the loop that iterates over ROIs, I wrote the comment like ` image_array[interested_batchid[k], cn , x1:x2, y1:y2] `

> is equivalent to interested_batchid = rois[:, 0]

Yeah, that is how I have altered the code now. It is something like,
```
for r_i in range(rois):
    interested_batchid = rois[r_i, 0]
    for cn in range(channels):
        .
        .
        image_array[interested_batchid, cn , x1:x2, y1:y2]
```

f0k: > for representing that `interested_batchid` is dependent on the loop that iterates over ROIs

But x1, x2, y1, y2 also depend on the loop that iterates over ROIs. You can do everything in one go.
```python
for (batchid, x1, x2, y1, y2) in rois:
    do something with image_array[batchid, :, x1:x2, y1:y2]
```

theano/tensor/signal/pool.py
```diff
@@ -2438,3 +2438,185 @@ def c_code(self, node, name, inp, out, sub):
...
+        h = self.pooled_h
+        w = self.pooled_w
+        channels = data_shape[1]
+        out_shape = [batch_size, num_rois, channels, h * w]
+        return [out_shape, out_shape]
+
+    def grad(self, inp, grads):
+        data, roi = inp
+        gz1, gz2 = grads
+        maxout, argmax = self(data, roi)
```
nouiz: Now we have a way to not request this duplicate node. Can you try to use that new feature? For this, change

    def grad(....

to

    def L_op(self, inputs, outputs, gout):

and you can do:

    maxout, argmax = outputs 

instead of rebuilding it.

Sentient07: Makes sense and nice feature indeed! I'll change it

theano/tensor/signal/pool.py
```diff
@@ -2438,3 +2438,185 @@ def c_code(self, node, name, inp, out, sub):
...
+                            mp_index = numpy.where(interest_region == bottom_index)[0]
+                            if mp_index.size == 0:
+                                continue
+
+                            # incrementing cause it was extracted from sliced array
+                            mp_index = numpy.add(mp_index, x1 + (pool_width * y1))
+                            gxxx[bottom_index] += channel_grad[mp_index].sum()
+
+    def grad(self, inp, grads):
+        disc = [tensor.zeros_like(i) for i in inp]
```
nouiz: I'm pretty sure this isn't right. Can you just remove that method instead? Currently it isn't tested too. So if you want to get that in Theano for the next release, just remove it. If needed it can be added later with tests.

Sentient07: I remember in some comment where @f0k had told me that this `grad` method is for computing the gradient of gradient. Since it's something that's not necessary, I had kept it as zeroes. Shall I change it to not implemented rather?

Issue #5187, opened 2016-11-05
Creating abstract ops for batch_normalization_train and batch_normalization_test


gvtulder:
This is a suggestion to make it easier to use the cuDNN batch normalization implementation. I'm creating an issue instead of a pull request to check if this would be a good idea, before I try to implement it.

As you can see in Keras (fchollet/keras#4162), you need quite a bit of code to properly use the cuDNN BN functions in the two gpu backends and have a fallback for when neither of those are available.

What would you think of adding abstract batch normalization ops to Theano? That is, two abstract ops that represent batch_normalization_train and batch_normalization_test, following the cuDNN semantics. Those ops can then be optimized into either one of the two gpu/cuDNN ops, or be replaced with an implementation using normal Theano ops for when cuDNN is not available.

Such functions would move much of this slightly awkward logic to Theano. For non-Keras purposes, this would also make it a lot easier to use these cuDNN BN functions, because you no longer have to write your code for a specific cpu/gpu backend.

The current `theano.tensor.nnet.bn.batch_normalization` is a bit different from the cuDNN functions, so I'd propose two new functions, for example, `theano.tensor.nnet.bn.batch_normalization_train` and `theano.tensor.nnet.bn.batch_normalization_test`, that follow exactly the cuDNN form.

nouiz:
I like the idea. I wanted to have something like this in Theano since we
wrapped cudnn bn functions.

For the interfaces, I don't like the fact that the train and test function
have different interface. So I'm not sure we want to follow it directly add
the user interface. I would prefer an interface where both functions
signature are more similar. What do you think?

f0k:
> For the interfaces, I don't like the fact that the train and test function have different interface.

Blame NVIDIA. `dnn_batch_normalization_train` returns `stdinv` because that's what cuDNN returns and requires for the gradient, and `dnn_batch_normalization_test` requires `var` instead because that's what cuDNN requires for inference.
The training Op could be extended to additionally perform the exponential moving averages, which are based on `var` and not `stdinv`. This requires inplace-versions along with an optimization, so one can do:
```
normed, mean, stdinv, new_running_mean, new_running_var = dnn_batch_normalization_train(x, beta, gamma, alpha, running_mean, running_var)
updates[running_mean] = new running_mean
updates[running_var] = new_running_var
```
[...]

PR #5190, opened 2016-11-07
Abstract Ops for batch normalization

gvtulder:
This is a project to define abstract batch normalization Ops that can get optimized to use the cuDNN Ops, while providing an alternative pure Theano implementation that can run on CPUs and on GPUs if cuDNN is not available.

I've started by copying the interface from the gpuarray backend, but this might also be an opportunity to improve some things.

See also issue #5187.

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -2005,28 +2005,3 @@ abstractconv_groupopt.register('local_abstractconv3d_gradinputs',
                                local_abstractconv3d_gradinputs_gemm, 30,
                                'conv_gemm',
                                'gpuarray', 'fast_compile', 'fast_run')
-
-
-# Register cuDNN batch normalization implementation
-abstract_batch_norm_groupopt = theano.gof.optdb.LocalGroupDB()
-abstract_batch_norm_groupopt.__name__ = "gpuarray_batchnorm_opts"
-register_opt('fast_compile')(abstract_batch_norm_groupopt)
-
-# cuDNN optimizations are only registered if cuDNN is available.
-# (we import these opts here instead of at the top of this file
-# to avoid a circular dependency problem with dnn)
-from .dnn import (local_abstract_batch_norm_train_cudnn,
-                  local_abstract_batch_norm_train_grad_cudnn,
-                  local_abstract_batch_norm_inference_cudnn)     # noqa: 402
-abstract_batch_norm_groupopt.register('local_abstract_batch_norm_train_dnn',
-                                      local_abstract_batch_norm_train_cudnn, 20,
-                                      'batchnorm_dnn',
-                                      'gpuarray', 'fast_compile', 'fast_run', 'cudnn')
-abstract_batch_norm_groupopt.register('local_abstract_batch_norm_train_grad_dnn',
-                                      local_abstract_batch_norm_train_grad_cudnn, 20,
-                                      'batchnorm_dnn',
-                                      'gpuarray', 'fast_compile', 'fast_run', 'cudnn')
-abstract_batch_norm_groupopt.register('local_abstract_batch_norm_inference_dnn',
-                                      local_abstract_batch_norm_inference_cudnn, 20,
-                                      'batchnorm_dnn',
-                                      'gpuarray', 'fast_compile', 'fast_run', 'cudnn')
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -2005,3 +2005,28 @@ abstractconv_groupopt.register('local_abstractconv3d_gradinputs',
                                local_abstractconv3d_gradinputs_gemm, 30,
                                'conv_gemm',
                                'gpuarray', 'fast_compile', 'fast_run')
+
+
+# Register cuDNN batch normalization implementation
+abstract_batch_norm_groupopt = theano.gof.optdb.LocalGroupDB()
+abstract_batch_norm_groupopt.__name__ = "gpuarray_batchnorm_opts"
+register_opt('fast_compile')(abstract_batch_norm_groupopt)
+
+# cuDNN optimizations are only registered if cuDNN is available.
+# (we import these opts here instead of at the top of this file
+# to avoid a circular dependency problem with dnn)
+from .dnn import (local_abstract_batch_norm_train_cudnn,
+                  local_abstract_batch_norm_train_grad_cudnn,
+                  local_abstract_batch_norm_inference_cudnn)     # noqa: 402
+abstract_batch_norm_groupopt.register('local_abstract_batch_norm_train_dnn',
+                                      local_abstract_batch_norm_train_cudnn, 20,
+                                      'batchnorm_dnn',
+                                      'gpuarray', 'fast_compile', 'fast_run', 'cudnn')
+abstract_batch_norm_groupopt.register('local_abstract_batch_norm_train_grad_dnn',
+                                      local_abstract_batch_norm_train_grad_cudnn, 20,
+                                      'batchnorm_dnn',
+                                      'gpuarray', 'fast_compile', 'fast_run', 'cudnn')
+abstract_batch_norm_groupopt.register('local_abstract_batch_norm_inference_dnn',
+                                      local_abstract_batch_norm_inference_cudnn, 20,
+                                      'batchnorm_dnn',
+                                      'gpuarray', 'fast_compile', 'fast_run', 'cudnn')
--- a/theano/gpuarray/tests/test_dnn.py
+++ b/theano/gpuarray/tests/test_dnn.py
@@ -1384,26 +1384,39 @@ def test_dnn_batchnorm_train():
 
     for mode in ('per-activation', 'spatial'):
         for vartype in (tensor6, T.tensor5, T.tensor4, T.tensor3, T.matrix, T.vector):
-            x, scale, bias = (vartype(n) for n in ('x', 'scale', 'bias'))
+            x, scale, bias, running_mean, running_var = (vartype(n)
+                                                         for n in ('x', 'scale', 'bias',
+                                                                   'running_mean',
+                                                                   'running_var'))
             ndim = x.ndim
             eps = 5e-3  # some non-standard value to test if it's used
+            running_average_factor = 0.3
 
             # forward pass, direct interface
             out_gpu, x_mean_gpu, x_invstd_gpu = dnn.dnn_batch_normalization_train(
                 x, scale, bias, mode, eps)
             # forward pass, abstract interface
-            out_abstract, x_mean_abstract, x_invstd_abstract = bn.batch_normalization_train(
-                x, scale, bias, mode, eps)
+            out_abstract, x_mean_abstract, x_invstd_abstract, \
+                out_running_mean_abstract, out_running_var_abstract = \
+                bn.batch_normalization_train(x, scale, bias, mode, eps,
+                                             running_average_factor,
+                                             running_mean, running_var)
             # reference forward pass
             if mode == 'per-activation':
                 axes = (0,)
             elif mode == 'spatial':
                 axes = (0,) + tuple(range(2, ndim))
             x_mean_ref = x.mean(axis=axes, keepdims=True)
-            x_invstd_ref = T.inv(T.sqrt(x.var(axis=axes, keepdims=True) + eps))
+            x_var_ref = x.var(axis=axes, keepdims=True)
+            x_invstd_ref = T.inv(T.sqrt(x_var_ref + eps))
             scale_ref = T.addbroadcast(scale, *axes)
             bias_ref = T.addbroadcast(bias, *axes)
+            m = T.cast(T.prod(x.shape) / T.prod(scale.shape), theano.config.floatX)
             out_ref = (x - x_mean_ref) * (scale_ref * x_invstd_ref) + bias_ref
+            out_running_mean_ref = running_mean * (1 - running_average_factor) + \
+                x_mean_ref * running_average_factor
+            out_running_var_ref = running_var * (1 - running_average_factor) + \
+                (m / (m - 1)) * x_var_ref * running_average_factor
             # backward pass
             dy = vartype('dy')
             grads_gpu = T.grad(None, wrt=[x, scale, bias], known_grads={out_gpu: dy})
@@ -1414,12 +1427,14 @@ def test_dnn_batchnorm_train():
             f_gpu = theano.function([x, scale, bias, dy],
                                     [out_gpu, x_mean_gpu, x_invstd_gpu] + grads_gpu,
                                     mode=mode_with_gpu)
-            f_abstract = theano.function([x, scale, bias, dy],
-                                         [out_abstract, x_mean_abstract, x_invstd_abstract] +
+            f_abstract = theano.function([x, scale, bias, running_mean, running_var, dy],
+                                         [out_abstract, x_mean_abstract, x_invstd_abstract,
+                                          out_running_mean_abstract, out_running_var_abstract] +
                                          grads_abstract,
                                          mode=mode_with_gpu)
-            f_ref = theano.function([x, scale, bias, dy],
-                                    [out_ref, x_mean_ref, x_invstd_ref] + grads_ref)
+            f_ref = theano.function([x, scale, bias, running_mean, running_var, dy],
+                                    [out_ref, x_mean_ref, x_invstd_ref,
+                                     out_running_mean_ref, out_running_var_ref] + grads_ref)
             # check if the abstract Ops have been replaced
             assert any([isinstance(n.op, dnn.GpuDnnBatchNorm) for n
                         in f_abstract.maker.fgraph.toposort()])
@@ -1438,9 +1453,11 @@ def test_dnn_batchnorm_train():
                 Dy = -1 + 2 * numpy.random.randn(*data_shape).astype(theano.config.floatX)
                 Scale = numpy.random.randn(*param_shape).astype(theano.config.floatX)
                 Bias = numpy.random.randn(*param_shape).astype(theano.config.floatX)
+                Running_mean = numpy.random.randn(*param_shape).astype(theano.config.floatX)
+                Running_var = numpy.random.randn(*param_shape).astype(theano.config.floatX)
                 outputs_gpu = f_gpu(X, Scale, Bias, Dy)
-                outputs_abstract = f_abstract(X, Scale, Bias, Dy)
-                outputs_ref = f_ref(X, Scale, Bias, Dy)
+                outputs_abstract = f_abstract(X, Scale, Bias, Running_mean, Running_var, Dy)
+                outputs_ref = f_ref(X, Scale, Bias, Running_mean, Running_var, Dy)
                 # compare outputs
                 utt.assert_allclose(outputs_gpu[0], outputs_ref[0])  # out
                 utt.assert_allclose(outputs_gpu[1], outputs_ref[1])  # mean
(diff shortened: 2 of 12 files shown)


gvtulder:
To summarize: this introduces two functions and three Ops that closely mirror the structure of the cuDNN Ops.

- `theano.tensor.bn.batch_normalization_train`
- `theano.tensor.bn.batch_normalization_test`
- `theano.tensor.bn.AbstractBatchNorm`
- `theano.tensor.bn.AbstractBatchNormInference`
- `theano.tensor.bn.AbstractBatchNormGrad`

The Ops have a Python implementation. There are optimizations that replace them with Theano Ops. Additional optimizations should introduce the cuDNN-based Ops if they are available.

Would it be useful to have a separate `AbstractBatchNormInferenceGrad` Op? At this point, the gradient of the inference Op is implemented in Theano, similar to how it is done in the cuDNN equivalent.

Using variance everywhere instead of `stdinv` is probably more consistent. I am not sure if this would introduce problems with the precision. (The batch normalization code is a bit sensitive anyway.)

Moving averages and in-place flags can also be added.

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+    out : tensor
+        Batch-normalized inputs.
+    mean : tensor
+        Means of `inputs` across the normalization axes.
+    stdinv : tensor
+        Inverse standard deviations of `inputs` across the normalization axes.
+
+    Notes
+    -----
+    Requires cuDNN 5 and Theano 0.9dev2 or more recent.
```
f0k: This sentence can be removed.

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+
+    For 5d tensors, the axes are (0, 2, 3, 4).
+    """
+    ndim = inputs.ndim
+    if gamma.ndim != ndim or beta.ndim != ndim:
+        raise ValueError("gamma and beta must be of the same dimensionality "
+                         "as inputs; got %d and %d instead of %d" %
+                         (gamma.ndim, beta.ndim, ndim))
+    if epsilon < 1e-5:
+        raise ValueError("epsilon must be at least 1e-5, got %f" % epsilon)
```
f0k: Not sure what to do about this -- this is a strict requirement for cuDNN, but not for the Theano code (although setting it too small may not work well). I'd suggest to remove this, what do others think?

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+                         "as inputs; got %d and %d instead of %d" %
+                         (mean.ndim, var.ndim, ndim))
+    if epsilon < 1e-5:
+        raise ValueError("epsilon must be at least 1e-5, got %f" % epsilon)
+
+    batchnorm_op = AbstractBatchNormInference(mode=mode)
+    return batchnorm_op(inputs, gamma, beta, mean, var, epsilon=epsilon)
+
+
+class AbstractBatchNorm(Op):
```
f0k: Maybe this should be called `AbstractBatchNormTrain` instead.

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+                              epsilon=1e-4):
+    """
+    Performs batch normalization of the given inputs, using the mean and
+    variance of the inputs.
+
+    Parameters
+    ----------
+    mode : {'per-activation', 'spatial'}
+        Whether to normalize per activation or share normalization factors
+        across spatial dimensions (i.e., all dimensions past the second).
```
f0k: I'd prefer to have an `axes` argument here, to freely specify the axes to be normalized over. Not all cases can be replaced with the cuDNN implementation then, but we shouldn't restrict the interface because of that.

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+        if self.mode == 'per-activation':
+            axes = (0,)
+        elif self.mode == 'spatial':
+            axes = (0,) + tuple(range(2, x.ndim))
+        else:
+            raise ValueError('invalid mode: %s', str(self.mode))
+
+        mean = x.mean(axes, keepdims=True)
+        stdinv = 1.0 / numpy.sqrt(x.var(axes, keepdims=True) + epsilon)
+        out = (x - mean) * scale * stdinv + bias
```
f0k: This should have `(scale * stdinv)` in brackets for improved performance (multiply the smaller tensors first).

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+        dvar = -(dy * (x - est_mean)).sum(axes, keepdims=True) * (scale / (two * est_var_eps * est_std))
+        return [dx, dscale, dbias, dmean, dvar, theano.gradient.DisconnectedType()()]
+
+    def connection_pattern(self, node):
+        # Specificy that epsilon is not connected to outputs.
+        return [[True], [True], [True], [True], [True], [False]]
+
+    def perform(self, node, inputs, output_storage):
+        x, scale, bias, estimated_mean, estimated_variance, epsilon = inputs
+        out = (x - estimated_mean) * scale / numpy.sqrt(estimated_variance + epsilon) + bias
```
f0k: Again, this should multiply the smaller tensors first.

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+        return None
+    if not isinstance(x.type, TensorType) or \
+       not isinstance(scale.type, TensorType) or \
+       not isinstance(bias.type, TensorType) or \
+       not isinstance(epsilon.type, TensorType):
+        return None
+
+    mean = x.mean(axes, keepdims=True)
+    stdinv = T.inv(T.sqrt(x.var(axes, keepdims=True) + epsilon))
+    out = (x - mean) * scale * stdinv + bias
```
f0k: `(scale * stdinv)`

theano/tensor/nnet/bn.py
```diff
@@ -73,3 +79,387 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+    if not isinstance(x.type, TensorType) or \
+       not isinstance(scale.type, TensorType) or \
+       not isinstance(bias.type, TensorType) or \
+       not isinstance(estimated_mean.type, TensorType) or \
+       not isinstance(estimated_variance.type, TensorType) or \
+       not isinstance(epsilon.type, TensorType):
+        return None
+
+    # TODO copy_stack_trace?
+    return [(x - estimated_mean) * scale / numpy.sqrt(estimated_variance + epsilon) + bias]
```
f0k: `(scale / T.sqrt(...))` (i.e., put in brackets, and use `T` instead of `numpy`)

theano/tensor/nnet/bn.py
gvtulder: Checking epsilon here makes the move to cuDNN a bit more predictable, because the optimization will never fail because of an epsilon that is too small. If epsilon is only checked in the optimization, it's more difficult for the user to find out why it doesn't use cuDNN.
If this check is removed here, the minimum value for cuDNN should be documented, or maybe even produce a warning. Is there a good reason to set epsilon lower than the cuDNN limit? If there isn't, I think it's better to check this here.

theano/tensor/nnet/bn.py
gvtulder: Yes, although it might be useful to mention the cuDNN optimization somewhere in the documentation.

theano/tensor/nnet/bn.py
gvtulder: An `axes` argument would be nice, although I do find the two modes convenient. Perhaps the `axes` argument could have special values `'per-activation'` and `'spatial'` for these common cases?

gvtulder:
I've added some tests (based on those for gpuarray).

gvtulder:
The last two commits add in-place options to the gpuarray Opts. The in-place updates for the running_mean and running_var seems to work (commit a3d8ec3).

The in-place updates for the input/output of batch_normalization_train (919f569) break the tests, for some reason. I do not know why this happens. The gradient computation no longer works if the in-place option is enabled. I have the impression that this might be because the GpuDnnBatchNormGrad depends on the original input, but that this original input is changed by GpuDnnBatchNorm before the gradient is computed.

gvtulder:
Found a temporary solution to the inplace problem by renaming the property. (See #5220.) The Jenkins GPU tests now pass.

f0k:
Sorry, I was busy. Wow, you're making some good progress there!

> The gradient computation no longer works if the in-place option is enabled. I have the impression that this might be because the GpuDnnBatchNormGrad depends on the original input, but that this original input is changed by GpuDnnBatchNorm before the gradient is computed.

Hmm, I think the inplace optimization shouldn't be applied if other clients need the input. But I'm not familiar with how the inplace optimization works in Theano, your registration seems proper. We'll need to summit some of the Theano devs for this. @nouiz?

Nice thought for `ndim > 5`, but this only works for `mode == 'spatial'`! Otherwise you're mixing up dimensions that should be normalized independently.

Generally your code looks very good. I can't do a thorough review now. Let's see what the devs have to say about the inplace problem. Good job so far!

gvtulder:
@f0k: The inplace problem is solved for the time being, but it's not really a good solution because it now works only because I reordered the optimizations. (See #5223.) Perhaps the optimization code isn't really expecting Ops to have multiple inplace optimizations that change the `destroy_map`.

> Nice thought for `ndim > 5`, but this only works for `mode == 'spatial'`! Otherwise you're mixing up dimensions that should be normalized independently.

Are you sure? I think it works for `mode == 'per-activation'` as well because that's only computing the averages across the 0 dimension. All the other activations are normalized independently, so there the number of dimensions shouldn't be a problem. I was a bit confused about this at first because cuDNN is so strict about 4D and 5D, but now I think it should work. (And I think it's tested as well.)

I'm not entirely sure how this works for non-contiguous inputs, though. That might in theory mess up the reshape and unreshape steps, but I assume the reshape Op handles non-contiguous data correctly.

f0k:
> The inplace problem is solved for the time being [...] Perhaps the optimization code isn't really expecting Ops to have multiple inplace optimizations

I just saw #5223. I guess you're right -- there were no Ops with independent inplace arguments until now, so you've uncovered a bug in the inner workings of Theano.

> Are you sure? I think it works for mode == 'per-activation' as well because that's only computing the averages across the 0 dimension.

It can be confusing, but yes: If you take a 6D numpy array `X` and compute `X.mean(axis=0)`, you get a 5D numpy array. If you reduce `X` to 5 dimensions before taking the mean across axis 0, it will average the last two dimensions together, which is not what we want. Your current code should actually fail with mismatched dimensions for 6D input, but you didn't extend the tests to check this.

gvtulder:
I'm not entirely sure what you mean with 'mismatched dimensions', but this seems to work:
```python
import numpy as np

shape = (2, 5, 7, 9)
x = np.array(range(0, np.prod(shape))).reshape(shape)
x_mean = x.mean(axis=0, keepdims=True)

x_reduced = x.reshape(shape[:2] + (-1,))
x_reduced_mean = x_reduced.mean(axis=0, keepdims=True)

x_mean_restored = x_reduced_mean.reshape((1,) + shape[1:])

print 'x_mean.shape', x_mean.shape  # (1, 5, 7, 9)
print 'x_mean_reduced.shape', x_mean_restored.shape  # (1, 5, 7, 9)
print np.all(x_mean == x_mean_restored)  # True
```
I'll have to think about this, because I might be missing something important.

f0k:
> I'll have to think about this, because I might be missing something important.

Oh, boy... you're right. With `axis=0`, it will collapse the first dimension and all the others can still be reshaped at will. I should go home and get some rest!

> I was a bit confused about this at first because cuDNN is so strict about 4D and 5D

Apparently that's because they don't want to support more strides. We can leave that up to Theano (it will copy the tensor if it can't be reshaped by just taking a different view at the memory).

> (And I think it's tested as well.)

Indeed, it is: https://github.com/Theano/Theano/pull/5190/files#diff-62bff40f5ac85ec667d250a8198c5557L1364
I thought I had seen it without tensor6. Again, it's time to go home! :)

nouiz:
We try to make an release candidate next week. I don't think we will have the times to review this well before. I think it would be great to include that in a second release candidate. This would make the release don't have GPU ops in the user graph (except for cudnn rnn). And that would be great to have.

gvtulder:
Rebased to fix a merge conflict.
The abstract helper methods `theano.tensor.nnet.bn.batch_normalization_train` and  `theano.tensor.nnet.bn.batch_normalization_test` now also accept non-broadcasted parameters. For spatial normalization, for example, you can now provide a `vector` scale instead of a `tensor4` with three 1-dimensions.

f0k:
> The abstract helper methods `theano.tensor.nnet.bn.batch_normalization_train` and `theano.tensor.nnet.bn.batch_normalization_test` now also accept non-broadcasted parameters.

I think that's a good idea. It complicates the code somewhat, but is confined to the helper functions.

Just to understand the relation to #5223: The code currently works because of the order you register the inplace optimizations? So #5223 is not strictly needed for this PR, but should be merged to prevent problems in the future?

gvtulder:
I don't know. The test works because the inplace optimizations run in a lucky order, but I don't know if this is always the case. For example, I don't know if the optimizations are always run only once. If there is a possibility that they are run more than once that would lead to problems, because the second call will succeed where the first call failed.

theano/tensor/nnet/bn.py
```diff
@@ -209,8 +209,12 @@ def batch_normalization_train(inputs, gamma, beta, axes='per-activation',
         raise ValueError("running_var must be of the same dimensionality "
                          "as gamma and beta; got %d instead of %d" %
                          (running_var.ndim, params_ndim))
+
+    # epsilon will be converted to floatX later. we need to check
+    # for rounding errors now, since numpy.float32(1e-5) < 1e-5.
```
f0k: Oh. Good catch!

gvtulder:
Jenkins failed the gpuarray tests. It looks like this is because the libgpuarray version changed from 0.1 to 1.0. Is Jenkins reusing the Theano build cache from previous runs?

slefrancois:
Indeed it was stuck with an old cache. jenkins test this

slefrancois:
jenkins test this

nouiz:
jenkins failure are not related to this PR and have been fixed in master.

theano/tensor/nnet/bn.py
```diff
@@ -75,3 +81,625 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+            running_mean = T.addbroadcast(running_mean, *axes)
+            running_var = T.addbroadcast(running_var, *axes)
+        out, mean, invstd, new_running_mean, new_running_var = batchnorm_op(
+            inputs, gamma, beta, epsilon=epsilon,
+            running_average_factor=running_average_factor,
+            running_mean=running_mean, running_var=running_var)
+        if new_running_mean.broadcastable != running_mean.broadcastable:
+            new_running_mean = T.patternbroadcast(new_running_mean, running_mean.broadcastable)
+        if new_running_var.broadcastable != running_var.broadcastable:
+            new_running_var = T.patternbroadcast(new_running_var, running_var.broadcastable)
```
nouiz: I think those 2 if aren't needed, but you can let them.

theano/tensor/nnet/tests/test_bn.py
```diff
@@ -122,21 +123,293 @@ def bn_ref(x, G, B, M, V):
...
+            # backward pass
+            dy = vartype('dy')
+            grads = T.grad(None, wrt=[x, scale, bias], known_grads={out: dy})
+            # reference backward pass
+            grads2 = T.grad(None, wrt=[x, scale, bias], known_grads={out2: dy})
+            # compile
+            f = theano.function([x, scale, bias, running_mean, running_var, dy],
+                                [out, x_mean, x_invstd, out_running_mean, out_running_var,
+                                 out2, x_mean2, x_invstd2, out_running_mean2, out_running_var2] +
+                                grads + grads2, mode='FAST_RUN')
```
nouiz: Why do you put mode here? I think it is better to let the default mode. This will allow more testing, as travis also test FAST_COMPILE and the daily buildbot run with DebugMode. It would be great to have them tested too.

theano/tensor/nnet/tests/test_bn.py
```diff
@@ -122,21 +123,293 @@ def bn_ref(x, G, B, M, V):
...
+    x, scale, bias, dy = T.tensor4('x'), T.tensor4('scale'), T.tensor4('bias'), T.tensor4('dy')
+    data_shape = (5, 10, 30, 25)
+    param_shape = (1, 10, 30, 25)
+
+    # forward pass
+    out, x_mean, x_invstd = bn.batch_normalization_train(x, scale, bias, 'per-activation')
+    # backward pass
+    grads = T.grad(None, wrt=[x, scale, bias], known_grads={out: dy})
+    # compile
+    f = theano.function([x, scale, bias, dy], [out, x_mean, x_invstd] + grads, mode='FAST_RUN')
```
nouiz: idem, remove the mode parameter.

theano/tensor/nnet/tests/test_bn.py
```diff
@@ -122,21 +123,293 @@ def bn_ref(x, G, B, M, V):
...
+                    x, scale_bc, bias_bc, running_mean_bc, running_var_bc, axes, eps)
+
+            # subtract the results of the non-broadcasted and broadcasted calls
+            results_non_bc = train_non_bc + (test_non_bc,)
+            results_bc = train_bc + (test_bc,)
+            results = [abs(r - r_bc) for (r, r_bc) in zip(results_non_bc, results_bc)]
+
+            # compile to compute all differences
+            f = theano.function([x, scale, bias, running_mean, running_var],
+                                T.sum(sum(results)), mode='FAST_RUN')
```
nouiz: idem

theano/tensor/nnet/tests/test_bn.py
```diff
@@ -122,21 +123,293 @@ def bn_ref(x, G, B, M, V):
...
+                                          for t in (scale, bias, mean, var))
+            out2 = (x - mean2) * (scale2 / T.sqrt(var2 + eps)) + bias2
+            # backward pass
+            dy = vartype('dy')
+            grads = T.grad(None, wrt=[x, scale, bias, mean, var], known_grads={out: dy})
+            # reference backward pass
+            grads2 = T.grad(None, wrt=[x, scale, bias, mean, var], known_grads={out2: dy})
+            # compile
+            f = theano.function([x, scale, bias, mean, var, dy],
+                                [out, out2] + grads + grads2, mode='FAST_RUN')
```
nouiz: idem

theano/tensor/nnet/tests/test_bn.py
```diff
@@ -122,21 +123,293 @@ def bn_ref(x, G, B, M, V):
...
+    # forward pass
+    out_train, x_mean, x_invstd = bn.batch_normalization_train(x, scale, bias, 'spatial')
+    out_test = bn.batch_normalization_test(x, scale, bias, mean, var, 'spatial')
+    # backward pass
+    grads_train = T.grad(None, wrt=[x, scale, bias], known_grads={out_train: dy})
+    grads_test = T.grad(None, wrt=[x, scale, bias], known_grads={out_test: dy})
+    # compile
+    f = theano.function([x, scale, bias, mean, var, dy],
+                        [out_train, x_mean, x_invstd, out_test] + grads_train + grads_test,
+                        mode='FAST_RUN')
```
nouiz: idem.

theano/tensor/nnet/tests/test_bn.py
gvtulder: I don't think I had a specific reason for that, it probably was copied over from somewhere else. I've removed it here and in the other places you mentioned.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1379,87 +1381,253 @@ def test_dnn_batchnorm_train():
...
                 # compare outputs
-                utt.assert_allclose(outputs[0], outputs[0 + 3])  # out
-                utt.assert_allclose(outputs[1], outputs[1 + 3])  # mean
-                utt.assert_allclose(outputs[2], outputs[2 + 3])  # invstd
+                utt.assert_allclose(outputs_gpu[0], outputs_ref[0])  # out
+                utt.assert_allclose(outputs_gpu[1], outputs_ref[1])  # mean
+                utt.assert_allclose(outputs_gpu[2], outputs_ref[2])  # invstd
+                utt.assert_allclose(outputs_gpu[3], outputs_ref[3])  # running_mean
+                utt.assert_allclose(numpy.nan_to_num(outputs_gpu[4]),
+                                    numpy.nan_to_num(outputs_ref[4]))  # running_var
```
nouiz: I didn't know about nan_to_num. Good solution. Ideally, we should find inputs range that won't give that, but I think this is a good enough solution. No need to change this.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1379,87 +1381,253 @@ def test_dnn_batchnorm_train():
...
+                                     out_running_mean_gpu, out_running_var_gpu] + grads_gpu,
+                                    mode=mode_with_gpu)
+            f_abstract = theano.function([x, scale, bias, running_mean, running_var, dy],
+                                         [out_abstract, x_mean_abstract, x_invstd_abstract,
+                                          out_running_mean_abstract, out_running_var_abstract] +
+                                         grads_abstract,
+                                         mode=mode_with_gpu)
+            f_ref = theano.function([x, scale, bias, running_mean, running_var, dy],
+                                    [out_ref, x_mean_ref, x_invstd_ref,
+                                     out_running_mean_ref, out_running_var_ref] + grads_ref)
```
nouiz: Pass `mode=mode_without_gpu`, to make sure the GPU isn't activated. It help make some tests pass when device=cuda, even if we don't officialy support that.

nouiz:
I have a local fix for the failure in FAST_COMPILE. I'll push this to this branch with the change to the opt.

nouiz:
I don't have the right to push to your branch. Can you enable it?

I have updated the optimization to the new way in the new back-end and did the small fixes I commented about:

https://github.com/nouiz/Theano/tree/gvtulder-f-batchnorm-abstract

Why did you add the automatic flatten in the gpu code?

thanks

theano/gpuarray/opt.py
```diff
@@ -2005,3 +2005,28 @@ def local_gpu_solve(op, context_name, inputs, outputs):
                                local_abstractconv3d_gradinputs_gemm, 30,
                                'conv_gemm',
                                'gpuarray', 'fast_compile', 'fast_run')
+
+
+# Register cuDNN batch normalization implementation
+abstract_batch_norm_groupopt = theano.gof.optdb.LocalGroupDB()
+abstract_batch_norm_groupopt.__name__ = "gpuarray_batchnorm_opts"
+register_opt('fast_compile')(abstract_batch_norm_groupopt)
```
gvtulder: No specific reason for the LocalGroupDB, other than that this was how it is done for the convolution ops and I'm still trying to understand the different ways to register optimizations.

gvtulder:
Thanks. I've added your commits (you should now also be able to push to this branch yourself).

I've added `__setstate__` methods to the ops with new properties. Would this need tests? (The other `__setstate__` methods in this file do not seem to have them.)

I might be misunderstanding your objections to the flattening, because I think it should work fine.

The idea behind the flattening in the GPU code is to allow you to use the cuDNN functions on inputs with more than five dimensions. There is an "unflattening" at the [end of the function](https://github.com/gvtulder/Theano/blob/b690578d54cd923323d97238352d8dfeefbc860e/theano/gpuarray/dnn.py#L2614-L2616) that restores everything back to its original shape, so I would expect it to work fine. There is no flattening and unflattening for the CPU/pure Theano version because it's not required there.

In the GPU functions, the flatten/unflatten for >5D is quite similar to the padding-to-four-dimensions that is already being done for inputs with less than four dimensions, so it seemed a logical thing to add. Does that make sense or did I completely misunderstand your objections? When could this be a problem?

Alternatively, this transformation could be moved to the optimizations. It would be a bit strange to handle <4D and >5D in different places, though.

nouiz:
About the flatten, I don't understand what the 6d dimensions mean. Can you explain? Is there only 1 interpretation?

Can you confirm that calling batch_normalization_train or  dnn_batch_normalization_train on a 6d tensor will give the same result for both mode? At first, I think it wasn't the case, now I think it is the case. If it is the case and it is tested (I think so from review), I'm good with the flatten.

nouiz:
LocalGroupDB are useful when we want a group of optimization to be executed
in a specific order inside an EquilibriumOptimizer.


Here, there is only 1 conversion possible per op as there is only 1 GPU
implementation. On the CPU, it is the same. As the CPU optimization are
already always after the GPU optimization, we don't need to force that
order.

Also, the way it is currently implemented, if we can't run it on the GPU as
if cudnn isn't there (or opencl is enabled... not ready) the CPU
optimization will get applied but we will have done this too late, so it
won't run on the GPU. So I need to revisit this and this would request an
LocalGroupDB!

I'll try to do this today.

gvtulder:
Yes, there are tests for 6D tensors in the gpu tests. They compare the results of `dnn_batch_normalization_train`, the  abstract `batch_normalization_train` and a reference implementation that doesn't do any flattening. In practice it seems to work.

It should also work on paper:
- With `per-activation`, the average is removing axis 0. With 6 dimensions, the input gets flattened to `[0, 1, 2, 3, 4 + 5]`. The mean taken over 0 will be `[x, 1, 2, 3, 4 + 5]`, which is reshaped to `[x, 1, 2, 3, 4, 5]` at the end.
- With `spatial`, the average is removing all axes except 1. With 6 dimensions, it should remove axes 0, 2, 3, 4, 5. The input is first flattened to `[0, 1, 2, 3, 4 + 5]`. The mean taken over 0, 2, 3, 4 will be `[x, 1, x, x, x + x]`, which is finally reshaped back to `[x, 1, x, x, x, x]`.

As long as you don't mix averaged and non-averaged axes, I think it works fine.

nouiz:
I think all is good, so merging.

thanks!

nouiz:
I merged this PR too quickly, here is another PR with follow up:

https://github.com/Theano/Theano/pull/5474

gvtulder:
Good. Thanks for reviewing and merging.

f0k:
:clap: to @gvtulder, very good work!

theano/tensor/nnet/bn.py
```diff
@@ -75,3 +82,631 @@ def batch_normalization(inputs, gamma, beta, mean, std,
...
+        if not isinstance(epsilon, theano.Variable):
+            epsilon = as_tensor_variable(epsilon)
+        return Apply(self, [x, dy, scale, x_mean, x_invstd, epsilon],
+                     [x.type(), scale.type(), scale.type()])
+
+    def infer_shape(self, node, shape):
+        return [shape[0], shape[2], shape[2]]
+
+    def perform(self, node, inputs, output_storage):
+        x, dy, scale, x_mean, x_invstd, epsilon = inputs
```
nouiz: epsilon isn't used in the perform nor in the opt that convert this to a graph. Is this normal? CuDNN do request it.

If it is normal that it isn't used and the only reason to add it here is that cudnn request it, then I'll add a comment about that.

f0k: It's normal. In cudnnBatchNormalizationBackward, savedMean and savedInvVariance are optional. If they're given, epsilon shouldn't be needed, but it's required by cudnnBatchNormalizationBackward anyway and they scare you by saying it must be the same than in the forward pass.
(We could make `x_mean` and `x_invstd` optional in `AbstractBatchNormTrainGrad` as well, but there's no strong use case. It would trade a little bit of memory against computation time.)

PR #5193, opened 2016-11-07
Numpy imports

kvmanohar22:
Changed imports of numpy to one common form of `import numpy as np` in `Theano/tensor/sparse/*` files


nouiz:
thanks

PR #5194, opened 2016-11-07
Update to fix python 3 error

notoraptor:
Update test function test_pycuda_memory_to_theano() to fix Python 3 error.

The refcount of a variable seems to be different between Python 2 and Python 3. So it seems better to stock an initial refcount in a variable and then do the checkings with that variable instead of directly using comparing to constant values.

Tests passed with Python 3 (Anaconda):

```
$ nosetests --verbose --nocapture theano/misc/tests/test_pycuda_theano_simple.py
Run pycuda only example to test that pycuda works. ... ok
Simple example with pycuda function and Theano CudaNdarray object. ... ok
theano.misc.tests.test_pycuda_theano_simple.test_pycuda_memory_to_theano ... 4
gpuarray ref count before creating a CudaNdarray 4
strides (20, 5, 1)
gpuarray ref count after creating a CudaNdarray 5
gpuarray ref count after deleting the CudaNdarray 4
ok

Ran 3 tests in 0.297s

OK
```

### DIFF
--- a/theano/misc/tests/test_pycuda_theano_simple.py
+++ b/theano/misc/tests/test_pycuda_theano_simple.py
@@ -86,7 +86,7 @@ def test_pycuda_memory_to_theano():
     initial_refcount = sys.getrefcount(y)
     print("gpuarray ref count before creating a CudaNdarray", end=' ')
     print(sys.getrefcount(y))
-    # assert sys.getrefcount(y) == 2
+    assert sys.getrefcount(y) == initial_refcount
     rand = numpy.random.randn(*y.shape).astype(numpy.float32)
     cuda_rand = cuda_ndarray.CudaNdarray(rand)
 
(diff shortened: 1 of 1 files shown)


theano/misc/tests/test_pycuda_theano_simple.py
```diff
@@ -83,9 +83,10 @@ def test_pycuda_memory_to_theano():
     # This increase the ref count with never pycuda. Do pycuda also
     # cache ndarray?
     # print y.get()
+    initial_refcount = sys.getrefcount(y)
     print("gpuarray ref count before creating a CudaNdarray", end=' ')
     print(sys.getrefcount(y))
-    assert sys.getrefcount(y) == 2
+    # assert sys.getrefcount(y) == 2
```
nouiz: Add this back and assert the number is the same as initial_refcount

theano/misc/tests/test_pycuda_theano_simple.py
```diff
@@ -100,16 +101,19 @@ def test_pycuda_memory_to_theano():
     y_ptr = int(y.gpudata)
     z = cuda_ndarray.from_gpu_pointer(y_ptr, y.shape, strides, y)
     print("gpuarray ref count after creating a CudaNdarray", sys.getrefcount(y))
-    assert sys.getrefcount(y) == 3
+    # assert sys.getrefcount(y) == 3
```
nouiz: remove old commented code at a few places here.

notoraptor:
Done !

Issue #3336, opened 2015-08-28
make downsample ignore_border=False, pad and use cudnn


nouiz:
This is faster then the old code and remove some CPU case.

But we also need to transition to a new default of ignore_border=True after some transition.

nouiz:
Instead of doing explicit padding, we can make 3 call to cuDNN pooling. The main that ignore the border and 2 others with changed pooling windows that will do the partial part. This would remove the need for a copy.

aam-at:
@nouiz Do you still need help with this issue for new gpuarray backend?

nouiz:
Nobody started it. If you start it, tell us.

thanks.

On Tue, Nov 1, 2016 at 2:50 PM, Alexander Matyasko <notifications@github.com

nouiz:
Done in the new back-end, so closing.

PR #5198, opened 2016-11-08
Gpuarray average and max pooling

aam-at:
Following up #3336, adds pooling with `ignore_border=False` for libgpuarray. Cudnn ignores border and when user sets ignore border to `False` in new backend cpu pooling operations are used which results in significant slowdown. Many people encountered this problem before and quite many theano libraries now issue warning to the user when ignore border is set to false.

It is slower than cudnn but necessary to reproduce some papers, such as Network in Network.

This pull request is related to #5157 which adds pooling with strides and padding to old backend. Both pull requests borrow and adopt caffe pooling kernels to theano.


aam-at:
I think similar to `abstract_conv` abstract operation for pooling should be created to have optimization database and perform optimization in correct order on graph.

Tests for new gpu pooling were passing (kernels are working) but cudnn tests are failing since new optimizations are coming before cudnn optimizations. I tried to fix by checking if cudnn is available before proceeding. Now new gpu pooling tests are failing.

Can you give me some hints on how I can fix problem with correct optimization order?

theano/gpuarray/pool.c
```diff
@@ -0,0 +1,341 @@
...
+      PyErr_SetString(PyExc_RuntimeError, "Could not fetch max_threads_dims");
+      return 1;
+    }
+    size_t threads_per_block = max_threads_dim;
+
+    if (ndims == 2) {
+      size_t num_kernels = z_dims[0] * z_dims[1] * z_dims[2] * z_dims[3];
+      size_t n_blocks = GET_BLOCKS(num_kernels, threads_per_block);
+      if (MAX_POOL) {
+        err = max_pool2d_kernel_call(1, &n_blocks, &threads_per_block, 0,
```
abergeron: I would use `_scall` here rather than the crude scheduling that mimics CAFFE.  I found that performance is often better when you don't try to maximise the number of threads, but rather strive for a reasonable balance with threads and blocks.

Anyway, the `_scall` version takes care of scheduling an appropriate number of blocks and threads as long as your kernel is ready to deal with unmatched number of blocks/threads, which is the case here. So I recommend that.

The interface is `_scall(1, &num_kernels, 0, <kernel parameters>`  The first parameter is the dimension of the launch (only 1d works for now).  The second parameter is the number of elements to compute in total.  The third one is the amount of shared memory to allocate.

abergeron:
@nouiz noticed the problem and has an idea about how to proceed.

We do need some way to order optimizations in a sensible way and we are working on a solution.  We'll update you when that is done.

We don't think AbstractPool is a good idea at the time.

theano/gpuarray/blas.py
```diff
@@ -1537,6 +1537,155 @@ def connection_pattern(self, node):
...
+        ws = as_tensor_variable(ws)
+        stride = as_tensor_variable(stride)
+        pad = as_tensor_variable(pad)
+        assert ws.type.ndim == stride.type.ndim and ws.type.ndim == pad.type.ndim
+        assert ws.type.ndim == 1
+
+        return Apply(self, [inp, ws, stride, pad], [inp.type()])
+
+    def get_params(self, node):
+        return node.inputs[0].type.context
```
abergeron: This is required for GpuKernelBase (or CGpuKernelBase).

theano/gpuarray/pool.c
```diff
@@ -0,0 +1,341 @@
...
+    ga_int wstart = static_cast<ga_int>(pw*stride_w) - static_cast<ga_int>(pad_w);
+    const ga_size wend = min(wstart + kernel_w, width);
+    dstart = max(dstart, 0);
+    hstart = max(hstart, 0);
+    wstart = max(wstart, 0);
+
+    const ga_size offset = (n*channels + c) * depth * height * width;
+    const DTYPE_i0* x_slice = x + offset;
+    // TODO: use DTYPE_o0_MAX for max value
+    DTYPE_o0 maxval = -__int_as_float(0x7f800000);  // ieee754 float max
```
abergeron: Yes that is bad and will not work for anything other than float32.  Using the first element is a good workaround.

theano/gpuarray/pool.c
aam-at: That's a nice addition as this was quite repetitive code and I also noticed some speedup when running tests.

aam-at:
I had to do rebase, as _scall was only added recently to master branch.

theano/gpuarray/blas.py
```diff
@@ -1537,6 +1537,155 @@ def connection_pattern(self, node):
...
+                'numpy_compat.h', 'float.h']
+
+    def c_header_dirs(self):
+        return [os.path.dirname(__file__), pygpu.get_include()]
+
+    def make_node(self, inp, ws, stride, pad):
+        ctx_name = infer_context_name(inp)
+        inp = as_gpuarray_variable(inp, ctx_name)
+        assert (inp.ndim in [4, 5])
+
```
aam-at: I changed assert check to use this more restrictive version.

theano/gpuarray/opt.py
```diff
@@ -1592,6 +1593,77 @@ def local_gpua_lift_abstractconv_graph(op, context_name, inputs, outputs):
...
+    inp = gpu_contiguous(as_gpuarray_variable(inp, ctx_name))
+
+    op = GpuPool(op.ignore_border, op.mode, op.ndim)
+    if inp.ndim == nd + 2:
+        return op(inp, ws, stride, pad)
+    else:
+        # reshape to 4D or 5D with 2 non-pooling dimensions
+        inp_padded = pad_dims(inp, 2, nd)
+        ret_padded = op(inp_padded, ws, stride, pad)
+        return unpad_dims(ret_padded, inp, 2, nd)
```
aam-at: Tests for 4d and 5d inputs were passing. My commit which tried to fix conflicts with cudnn optimizations broke them (cudnn optimization applied before new pooling optimizations). For now, I can remove optimizations and tests with `ignore_border = True`, so there are will be no conflict with cudnn optimizations.

aam-at:
Do I need to add somewhere pool.py?

It is failing on -
```python
from theano.gpuarray import (basic_ops, GpuArrayType)
ImportError: cannot import name basic_ops

**Comment by abergeron** [comment]
This seems to be a circular import problem.  I'll try to see where it comes from.

File: theano/gpuarray/pool.py [hunk 6]
```diff
@@ -0,0 +1,202 @@
+from __future__ import absolute_import, print_function, division
+import os.path
+
+import pygpu
```
abergeron: Add a try .. except around this like the other files.

abergeron:
The unprotected import of pygpu is what is causing the problem.

abergeron:
jenkins test this please

aam-at:
Test will fail for now as I am checking if cudnn available and `ignore_border == False`. However, in tests I am still testing when `ignore_border == True`. I can for now disable check when `ignore_border = True`, so tests will work. If this is acceptable for this pull requests (no tests for `ignore_border == True` or some other workaround), I can change it.

theano/gpuarray/tests/test_pool.py
```diff
@@ -0,0 +1,213 @@
...
+    ref_mode = copy.copy(mode_without_gpu)
+    ref_mode.check_py_code = False
+    gpu_mode = copy.copy(mode_with_gpu)
+    gpu_mode.check_py_code = False
+
+    for shp in shps:
+        for mode, ws, st in itertools.product(test_mode, test_ws, test_st):
+            if ws[0] > shp[-2] or ws[1] > shp[-1]:
+                continue
+            for ignore_border, pad in zip((True, False), [(1, 1), (0, 0)]):
```
nouiz: This make that we don't test the combination of ignore_border and padding. Is there a reason for this? If this is not supported, I think we don't raise good user error for that cases. If the code support it, add a case to test this.

theano/gpuarray/tests/test_pool.py
```diff
@@ -0,0 +1,213 @@
...
+    ref_mode = copy.copy(mode_without_gpu)
+    ref_mode.check_py_code = False
+    gpu_mode = copy.copy(mode_with_gpu)
+    gpu_mode.check_py_code = False
+
+    for shp in shps:
+        for mode, ws, st in itertools.product(test_mode, test_ws, test_st):
+            if ws[0] > shp[-3] or ws[1] > shp[-2] or ws[2] > shp[-1]:
+                continue
+            for ignore_border, pad in zip((True, False), [(1, 1, 1), (0, 0, 0)]):
```
nouiz: idem.

theano/gpuarray/pool.py
```diff
@@ -0,0 +1,207 @@
...
+class GpuAveragePoolGrad(CGpuKernelBase):
+    """
+    Implement the grad of average pooling on the gpu.
+
+    """
+    __props__ = ('ignore_border', 'mode', 'ndim')
+
+    def __init__(self, ignore_border, mode='max', ndim=2):
+        self.ndim = ndim
+        self.ignore_border = ignore_border
```
nouiz: This op and GpuMaxPoolGrad don't have a method get_op_params. This make that ignore_border isn't provided to the C code. Is that normal?

theano/gpuarray/tests/test_pool.py
```diff
@@ -0,0 +1,213 @@
...
+                    for node in gg.maker.fgraph.toposort()
+                ])
+                assert any([
+                    isinstance(node.op, DownsampleFactorMaxGradGrad)
+                    for node in gg2.maker.fgraph.toposort()
+                ])
+                assert numpy.allclose(gg(), gg2()), (shp, ws, st, pad, mode, ignore_border)
+
+
+def test_pool3d():
```
nouiz: Can you add a small test when the inputs is 6d or 7d, to trigger the code that cover that case?

theano/gpuarray/tests/test_pool.py
aam-at: This combination is not supported and the exception is raised in class for cpu pooling - https://github.com/Theano/Theano/blob/master/theano/tensor/signal/pool.py#L482

I don't think that should be duplicated as user will work with pool_2d or pool_3d interface and operations will be lifted to gpu automatically.

aam-at:
@nouiz Are you going to put those commits in a separate pull request or should I do rebase of them from your branch?

theano/gpuarray/tests/test_pool.py
nouiz: Use sometimes build the graph directly on the GPU. So can you add that check in the GPU op too?

nouiz:
You can just

nouiz:
You can just add my commits to your PR. My commits depend on your PR, so I can't make a separate PR without spliting them.

thanks

theano/gpuarray/pool.py
nouiz: thanks for the confirmation.

aam-at:
I added some checks for both c and python code and tests for 6d/7d input. As user may directly use these gpu ops, I also added `infer_shape`, `grad` and `connection_pattern`.

nouiz:
jenkins test this

I think all is good, so when jenkins pass, we can merge. Thanks.

aam-at:
@nouiz tests have passed. it is ready to be merged.

nouiz:
thanks

PR #5207, opened 2016-11-10
Conda fixes

mingwandroid:
Here are the patches we are applying to Theano at Continuum Analytics.

### DIFF
--- a/theano/scan_module/scan.py
+++ b/theano/scan_module/scan.py
@@ -265,11 +265,9 @@ def scan(fn,
         ``n_steps`` is the number of steps to iterate given as an int
         or Theano scalar. If any of the input sequences do not have
         enough elements, scan will raise an error. If the *value is 0* the
-        outputs will have *0 rows*. If the value is negative, ``scan``
-        will run backwards in time. If the ``go_backwards`` flag is already
-        set and also ``n_steps`` is negative, ``scan`` will run forward
-        in time. If n_steps is not provided, ``scan`` will figure
-        out the amount of steps it should run given its input sequences.
+        outputs will have *0 rows*. If n_steps is not provided, ``scan`` will
+        figure out the amount of steps it should run given its input
+        sequences. ``n_steps`` < 0 is not supported anymore.
 
     truncate_gradient
         ``truncate_gradient`` is the number of steps to use in truncated
--- a/theano/scan_module/scan_checkpoints.py
+++ b/theano/scan_module/scan_checkpoints.py
@@ -4,7 +4,8 @@ import theano
 
 
 def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
-                     name="checkpointscan_fn", n_steps=None, save_every_N=10):
+                     name="checkpointscan_fn", n_steps=None, save_every_N=10,
+                     padding=True):
     """Scan function that uses less memory, but is more restrictive.
 
     In :func:`~theano.scan`, if you compute the gradient of the output
@@ -52,18 +53,22 @@ def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
 
     n_steps
         ``n_steps`` is the number of steps to iterate given as an int
-        or Theano scalar. If any of the input sequences do not have
-        enough elements, scan will raise an error. If the **value is 0**
-        the outputs will have **0 rows**. If the value is negative,
-        ``scan`` will run backwards in time. If the ``go_backwards`` flag
-        is already set and also ``n_steps`` is negative, ``scan`` will run
-        forward in time. If n_steps is not provided, ``scan`` will figure
-        out the amount of steps it should run given its input sequences.
+        or Theano scalar (> 0). If any of the input sequences do not have
+        enough elements, scan will raise an error. If n_steps is not provided,
+        ``scan`` will figure out the amount of steps it should run given its
+        input sequences.
 
     save_every_N
         ``save_every_N`` is the number of steps to go without storing
         the computations of ``scan`` (ie they will have to be recomputed
         during the gradient computation).
+
+    padding
+        If the length of the sequences is not a multiple of ``save_every_N``,
+        the sequences will be zero padded to make this version of ``scan``
+        work properly, but will also result in a memory copy. It can be
+        avoided by setting ``padding`` to False, but you need to make
+        sure the length of the sequences is a multple of ``save_every_N``.
 
     Returns
     -------
@@ -96,16 +101,31 @@ def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
     if n_steps is None:
         n_steps = sequences[0].shape[0]
 
-    # Compute the number of steps of the inner and of the outer scan
-    o_n_steps = theano.tensor.cast(n_steps / save_every_N, 'int64')
-    i_n_steps = save_every_N
+    # Compute the number of steps of the outer scan
+    o_n_steps = theano.tensor.cast(theano.tensor.ceil(n_steps / save_every_N),
+                                   'int64')
+
+    # Compute the number of steps of the inner scan
+    i_n_steps = save_every_N * theano.tensor.ones((o_n_steps,), 'int64')
+    mod = n_steps % save_every_N
+    last_n_steps = theano.tensor.switch(theano.tensor.eq(mod, 0),
+                                        save_every_N, mod)
+    i_n_steps = theano.tensor.set_subtensor(i_n_steps[-1], last_n_steps)
+
+    # Pad the sequences if needed
+    if padding:
+        for i, s in enumerate(sequences):
+            n = s.shape[0] % save_every_N
+            z = theano.tensor.zeros((n, s.shape[1:]), dtype=s.dtype)
+            sequences[i] = theano.tensor.concatenate([s, z], axis=0)
 
     # Establish the input variables of the outer scan
     o_sequences = [s.reshape([s.shape[0] / save_every_N, save_every_N] +
                              [s.shape[i] for i in range(1, s.ndim)],
                              s.ndim + 1) for s in sequences]
+    o_sequences.append(i_n_steps)
     new_nitsots = [i for i in outputs_info if i is None]
-    o_nonsequences = non_sequences + [i_n_steps]
+    o_nonsequences = non_sequences
 
     def outer_step(*args):
         # Separate the received arguments into their respective (seq, outputs
@@ -117,11 +137,11 @@ def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
 
         # Call the user-provided function with the proper arguments
         results, updates = theano.scan(fn=fn,
-                                       sequences=i_sequences,
+                                       sequences=i_sequences[:-1],
                                        outputs_info=i_outputs_infos,
-                                       non_sequences=i_non_sequences[:-1],
+                                       non_sequences=i_non_sequences,
                                        name=name + "_inner",
-                                       n_steps=i_non_sequences[-1])
+                                       n_steps=i_sequences[-1])
         if not isinstance(results, list):
             results = [results]
 
--- a/theano/scan_module/tests/test_scan_checkpoints.py
+++ b/theano/scan_module/tests/test_scan_checkpoints.py
@@ -38,14 +38,14 @@ class TestScanCheckpoint(unittest.TestCase):
         """Test forward computation of A**k."""
         f = theano.function(inputs=[self.A, self.k],
                             outputs=[self.result, self.result_check])
-        out, out_check = f(range(10), 100)
+        out, out_check = f(range(10), 101)
         assert numpy.allclose(out, out_check)
 
     def test_backward_pass(self):
         """Test gradient computation of A**k."""
         f = theano.function(inputs=[self.A, self.k],
                             outputs=[self.grad_A, self.grad_A_check])
-        out, out_check = f(range(10), 100)
+        out, out_check = f(range(10), 101)
         assert numpy.allclose(out, out_check)
 
     @unittest.skipUnless(PYGPU_AVAILABLE, 'Requires pygpu.')


theano/configdefaults.py
```diff
@@ -418,6 +419,19 @@ def f(suffix=suffix):
...
     rc = 1
 
+# Anaconda on Windows has mingw-w64 packages including GCC, but they may not be on PATH.
+if rc != 0:
+    if "Continuum Analytics" in sys.version and sys.platform == "win32":
+        mingw_w64_gcc = os.path.join(os.path.dirname(sys.executable), "Library", "mingw-w64", "bin", "g++")
+        try:
+            rc = call_subprocess_Popen([mingw_w64_gcc, '-v'])
+            if rc == 0:
+                maybe_add_to_os_environ_pathlist('PATH', os.path.dirname(mingw_w64_gcc))
```
nouiz: I don't understand why you don't force to add the path. If on line 423, rc isn't 0. Then we can't just call g++ without doing something else.

Instead of adding and using the method maybe_add_to_os_environ_pathlist, did you try to replace line 429 to  "param = mingw_w64_gcc" ? param can be a full path to a compiler that support g++ interface. This would be a simpler change that I think would work.

theano/configdefaults.py
```diff
@@ -1238,21 +1252,20 @@ def default_blas_ldflags():
                                           "mk2_rt"]])
 
         # Anaconda
-        if "Anaconda" in sys.version or "Continuum" in sys.version:
-            # If the "mkl-service" conda package (available
-            # through Python package "mkl") is installed and
+        if "Continuum Analytics" in sys.version:
```
nouiz: Is this also true for older Anaconda version? I'm not opposed to only support current Anaconda, but if we do so, I would add that in the documentation so user don't get surprised.

theano/configdefaults.py
```diff
@@ -1238,21 +1252,20 @@ def default_blas_ldflags():
...
-                # This branch is executed if no exception was raised
+            # package "mkl") are actually available.  Using
+            # "conda install mkl-service" will install both,
+            # as well as optimized versions of numpy and scipy.
+             try:
+                 import mkl  # noqa
+             except ImportError as e:
+                _logger.warning('Conda mkl is not available, use `conda install mkl-service`: %s', e)
+             else:
+                 # This branch is executed if no exception was raised
```
nouiz: You seem to have added extra spaces on the lines above.

theano/configdefaults.py
```diff
@@ -1267,6 +1280,7 @@ def default_blas_ldflags():
                               blas_info.get('library_dirs', [])])
                 res = try_blas_flag(flags)
                 if res:
+                    maybe_add_to_os_environ_pathlist('PATH', lib_path)
```
nouiz: I see that we will need here the function maybe_add_to_os_environ_pathlist for here. But I think we should call this method only on windows.

theano/configdefaults.py
mingwandroid: It has been this way since at least early 2014.

theano/configdefaults.py
mingwandroid: On Windows, environment variables are limited to 8191 characters so you can cause trouble for other things by forcing it.

mingwandroid: It's not just the compilation that needs the full path, but also the execution of the kernel.

theano/configdefaults.py
mingwandroid: Oops. I will fix that.

theano/configdefaults.py
nouiz: ok. Could you just tell in the doc, the current anaconda version, so that will make clear we don't support older version and prompt people to update it if needed?

theano/configdefaults.py
nouiz: ok. Can you add a comment about that so that we don't forget the reason later?

mingwandroid: Sure thing.

theano/configdefaults.py
mingwandroid: Seems I am mistaken about `sys.version`, it can contain either `Anaconda` or `Continuum Analytics` depending on whether you installed `Anaconda` or `Miniconda`. Then there is also `conda-forge` to worry about. I am wondering whether to elide these checks altogether and just proceed to check the next parts instead: `import mkl` and `call_subprocess_Popen([mingw_w64_gcc, '-v'])`

What do you think? Otherwise the if statement looking for conda-y things will not be very pretty.

nouiz: I suppose conda-forge still use the python from anaconda and just add packages. Do you have other information? If not, I'll keep the old condition and add a comment that this is needed for miniconda and anaconda.

Another idea, why not change the condition to "if os.path.exists(...)

     mingw_w64_gcc = os.path.join(os.path.dirname(sys.executable), "Library", "mingw-w64", "bin", "g++")
     if os.path.exists(mingw_w64_gcc):

that way, if the path exist, we will try it. It will help prevent to start a new process when not needed.

nouiz:
What about Mac? I was told it would be very hard if possible to have Theano link with MKL. Did you make that work? Do Theano link with acelerate instead?

thanks

mingwandroid:
Yes, MKL is being used on Mac too (to the best of my knowledge):

```
lsof -p $(ps -A | grep [p]ython | cut -d' ' -f1) | grep mkl
python3.5 91224 rdonnelly  txt    REG    1,4  19249600 14439586 /Users/rdonnelly/r_3_3_1-x64-3.5/lib/libmkl_intel_lp64.dylib
python3.5 91224 rdonnelly  txt    REG    1,4  31395872 14439587 /Users/rdonnelly/r_3_3_1-x64-3.5/lib/libmkl_intel_thread.dylib
python3.5 91224 rdonnelly  txt    REG    1,4  26329520 14439578 /Users/rdonnelly/r_3_3_1-x64-3.5/lib/libmkl_core.dylib
```

nouiz:
This only tell if the file is open. As conda numpy link to it, it is sure it will be opened.

To know if Theano reuse it do:
python -c "import theano;print theano.config.blas.ldflags"

This will trigger the code that try to link to blas and will print what will be selected. If empty string, it mean we don't link directly to blas, so this will give slower execution.

mingwandroid:
Result is:

```
-L/Users/rdonnelly/r_3_3_1-x64-3.5/lib -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lm -lm -Wl,-rpath,/Users/rdonnelly/r_3_3_1-x64-3.5/lib
```

nouiz:
So it seem to link to MKL on mac. On which version of Mac OS did you test that? I have heard problems with the newer version, but it was working for us on older version.

@abergeron

theano/configdefaults.py
```diff
@@ -1238,21 +1252,20 @@ def default_blas_ldflags():
...
                 import mkl  # noqa
             except ImportError as e:
-                _logger.info('Conda mkl is not available: %s', e)
+                _logger.warning('Conda mkl is not available, use `conda install mkl-service`: %s', e)
             else:
                 # This branch is executed if no exception was raised
                 if sys.platform == "win32":
-                    lib_path = os.path.join(sys.prefix, 'DLLs')
+                    lib_path = os.path.join(sys.prefix, 'Library', 'bin')
                     flags = ['-L"%s"' % lib_path]
```
notoraptor: Hello !

Not sure if the problem comes exactly from here, but I have tried to install this pull request on Windows 8 an then execute theano/misc/check_blas.py to test it, but I got some errors related to os.listdir().

First, the theano version:
```
(python2) C:\Users\HPPC\mila\test>python -c "import theano; print theano.__version__"
0.9.0dev4.dev-0fc36d4db9a73e921a6b1afe15528c952de65093
```

Then, the test:

```
(python2) C:\Users\HPPC\mila\test>python -m theano.misc.check_blas > conda-pull-request-5207-theano-
python2.log 2> conda-pull-request-5207-theano-python2-stderr.log
```

And the error printed in conda-pull-request-5207-theano-python2-stderr.log:

```
ERROR (theano.gof.cmodule): [Error 123] La syntaxe du nom de fichier, de répertoire ou de volume est incorrecte: '"C:\\donnees\\programmes\\Anaconda3\\envs\\python2\\Library\\bin"/*.*'
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\misc\check_blas.py", line 300, in <module>
    order=options.order)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\misc\check_blas.py", line 67, in execute
    f = theano.function([], updates=[(c, 0.4 * c + .8 * T.dot(a, b))])
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\compile\function.py", line 326, in function
    output_keys=output_keys)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\compile\pfunc.py", line 486, in pfunc
    output_keys=output_keys)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\compile\function_module.py", line 1784, in orig_function
    defaults)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\compile\function_module.py", line 1651, in create
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\link.py", line 699, in make_thunk
    storage_map=storage_map)[:3]
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\vm.py", line 1057, in make_all
    impl=impl))
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\op.py", line 924, in make_thunk
    no_recycling)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\op.py", line 828, in make_c_thunk
    output_storage=node_output_storage)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\cc.py", line 1190, in make_thunk
    keep_lock=keep_lock)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\cc.py", line 1131, in __compile__
    keep_lock=keep_lock)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\cc.py", line 1589, in cthunk_factory
    key=key, lnk=self, keep_lock=keep_lock)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\cmodule.py", line 1155, in module_from_key
    module = lnk.compile_cmodule(location)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\cc.py", line 1490, in compile_cmodule
    lib_dirs=self.lib_dirs(),
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\cc.py", line 1085, in lib_dirs
    ret += x.c_lib_dirs()
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\tensor\blas.py", line 504, in c_lib_dirs
    return ldflags(libs=False, libs_dir=True)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\tensor\blas.py", line 386, in ldflags
    include_dir=include_dir)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\gof\utils.py", line 258, in rval
    val = f(*args, **kwargs)
  File "C:\donnees\programmes\Anaconda3\envs\python2\lib\site-packages\theano\tensor\blas.py", line 423, in _ldflags
    for f in os.listdir(d):
WindowsError: [Error 123] La syntaxe du nom de fichier, de répertoire ou de volume est incorrecte: '"C:\\donnees\\programmes\\Anaconda3\\envs\\python2\\Library\\bin"/*.*'
```

So, it seems os.listdir() called in theano/tensor/blas.py is angry, and I suspects it is because of the quotes around the path. So I run the following script:

```
from __future__ import print_function
import sys, os, platform

print('Platform:', platform.platform())
print('Python version:', sys.version_info[0:])
print('Python sys.prefix:', sys.prefix)
print()
try:
	path_content = os.listdir(sys.prefix)
	print('os.listdir() succeeds with sys.prefix.')
except Exception as e:
	print('os.listdir() fails with sys.prefix.')
	print(e)
print()
try:
	path_content = os.listdir('"%s"' % sys.prefix)
	print('os.listdir() succeeds with sys.prefix surrounded with quotes.')
except Exception as e:
	print('os.listdir() fails with sys.prefix surrounded with quotes.')
	print(e)
```

And I got this output:
```
Platform: Windows-8.1-6.3.9600
Python version: (2, 7, 12, 'final', 0)
Python sys.prefix: C:\donnees\programmes\Anaconda3\envs\python2

os.listdir() succeeds with sys.prefix.

os.listdir() fails with sys.prefix surrounded with quotes.
[Error 123] La syntaxe du nom de fichier, de rÚpertoire ou de volume est incorrecte: '"C:\\donnees\\
programmes\\Anaconda3\\envs\\python2"/*.*'
```

So it seems it is indeed because of the quotes, and I think that the path passed to os.listdir() is generated at this line of theano/configdefaults.py (```flags = ['-L"%s"' % lib_path]```).

Maybe the call to os.listdir() should be modified in theano/tensor/blas.py by replacing
```for f in os.listdir(d):```
with:
```for f in os.listdir(d.strip('"')):```

?

**Comment by nouiz** [comment]
This is recent change in Theano master to allow support of spaces in path
on windows. It transform all -Lpath to -L"path" on windows.

From my understanding, your suggested fix should work.

Ray, do you have a better suggestion/idea on how to fix this while keeping
spaces support in that?

**Comment by notoraptor** [comment]
About my problem mentioned above, I have modified 5 lines of code in my local copy of your version of `theano/tensor/blas.py` and now the test succeeds (`python -m theano.misc.check_blas`) on my computer (`Windows-8.1-6.3.9600`) with both `Python 2.7.12` and `Python 3.5.2`!

```
diff oldcode newcode
423c423
<             for f in os.listdir(d):
---
>             for f in os.listdir(d.strip('"')):
436,439c436,437
<         if t.startswith("'") or t.startswith('"'):
<             t = t[1:]
<         if t.endswith("'") or t.endswith('"'):
<             t = t[:-1]
---
>         if (t.startswith("'") and t.endswith("'")) or (t.startswith('"') and t.endswith('"')):
>             t = t[1:-1]
```

The latter modification is done because the variable `t` sometimes contains something like `-L"some\path"` (with quotes after `-L`), so the old code removes the last quote but not the first. The new code will remove quotes only if they appear both at the start and at the end of `t`, otherwise t remains unchanged.

**Comment by nouiz** [comment]
Just a note that the time out problem in travis is fixed in master. So new commit in this PR won't have that problem anymore.

**Comment by mingwandroid** [comment]
OK thanks all, I will try to get back to this PR as soon as I can. Hopefully early next week. @notoraptor your quote fix sounds fine, I guess you should make a PR and I'll rebase mine on top of that (if it gets merged first that is!).

Is it agreed that I can remove the checks for specific Continuum/conda Pythons altogether and just let the subsequent tests be performed?

**Comment by notoraptor** [comment]
@mingwandroid PR done and merged ! You can rebase from now !

File: theano/__init__.py [hunk 6]
```diff
@@ -31,10 +31,6 @@
 
 import sys
 
-if sys.platform == 'win32' and sys.version_info[0:2] == (3, 5):
-    raise RuntimeError(
-        "Theano do not support Python 3.5 on Windows. Use Python 2.7 or 3.4.")
```
nouiz: I think the Python build with MSVC (so the official python 3.5 on Windows) is still not supported. Instead of just removing it, can you skip this error when MSVC wasn't used? I don't know how to detect that, if you also don't know, you could skip that error for anaconda only.

theano/configdefaults.py
nouiz: one of your comment above tell that for anaconda and miniconda, 2 strings are used. So undo the change in this line and on line 1303.

theano/configdefaults.py
nouiz: don't forget to add a comment here.

nouiz:
I forgot, travis have some flake8 errors: https://travis-ci.org/Theano/Theano/jobs/178600756#L1010

theano/configdefaults.py
```diff
@@ -418,6 +419,19 @@ def f(suffix=suffix):
 except OSError:
     rc = 1
 
+# Anaconda on Windows has mingw-w64 packages including GCC, but they may not be on PATH.
+if rc != 0:
+    if "Continuum Analytics" in sys.version and sys.platform == "win32":
```
nouiz: I'm fine to remove the sys.version check here as discussed elsewhere.

theano/configdefaults.py
nouiz: As dicussed elsewhere, I think it is fine to remove the check here, but I think we should conditionally do the warning on line 1264. I don't think it is good that there is a warning when mkl isn't there and we don't use anaconda.

Does conda-forge have the mkl packages?

theano/__init__.py
mingwandroid: All conda Pythons are built with MSVC, and I can confirm that there are no problems here, also none when using mingw-w64 GCC.

theano/configdefaults.py
mingwandroid: Should I add the comment regarding `maybe_add_to_os_environ_pathlist` to the 2 call-sites or to the doc string of the function instead?

mingwandroid:
Ok, here's some new versions. I will work on the pkgs Arnaud sent next to get them into the Anaconda defaults channel and perform some testing with those. I still have the flake8 matters to attend to, but I wanted to get your feedback to ensure I'm moving in the right direction.

Thanks!

mingwandroid:
What version number are you aiming to release soon? I am updating the conda recipes need to fill that in.

mingwandroid:
For now, I'm hosting the conda package work at:
 
https://github.com/mingwandroid/conda-recipes

The 3 latest commits are the preparatory work for this, please feel free to make PRs. I made a few very minor modifications to Arnaud's packages.

mingwandroid:
Oh, test packages for osx-64 are at https://www.anaconda.org/rdonnelly, more to come later.

nouiz:
ok. thanks for the clarification. Ideally, we should test with the official
Python, but I don't want us to delays this PR for that. Also, I'm not sure
we have the time to mainting the support of it.

nouiz:
I would tell the docstring.

nouiz:
0.9rc1

nouiz:
0.9rc1 is the Theano release version we plan to make. I'm not sure about
libgpuarray and pygpu version numbers.

nouiz:
continued in https://github.com/Theano/Theano/pull/5290

PR #5210, opened 2016-11-10
Solve

### LINKED ISSUE #5017: Merge the Solve op in sandbox/solve.py and in tensor/nlinalg.py
See the title. Investigate what is the difference and delete sandbox/solve.py if the one in nlinalg.py does everything. Otherwise document or merge both version.

tfjgeorge:
This fixes #5017 . I removed sandbox/solve.py which was not used anywhere as linalg/ops.py already imports solve from slinalg.py. I updated the documentation for solve and its lower diagonal/upper diagonal counterparts.


nouiz:
@slefrancois

slefrancois:
jenkins test this

theano/sandbox/solve.py
```diff
@@ -1,75 +0,0 @@
-from __future__ import absolute_import, print_function, division
```
slefrancois: Instead of deleting sandbox.solve, we should keep the file as a placeholder that raises and error that tells users to import slinalg.solve.

theano/sandbox/solve.py
```diff
@@ -1,75 +0,0 @@
...
-        odtype = scalar.upcast(A_.dtype, b_.dtype)
-        otype = tensor.TensorType(broadcastable=b_.broadcastable, dtype=odtype)
-        return gof.Apply(op=self, inputs=[A_, b_], outputs=[otype()])
-
-    def perform(self, node, inp, out):
-        A, b = inp
-        output, = out
-        ret = scipy.linalg.solve(A, b)
-        if ret.dtype != node.outputs[0].dtype:
-            print("WARNING: Solve.perform() required cast.", file=sys.stderr)
```
slefrancois: @lamblin @nouiz : should we transfer the cast warning to slinalg.solve?

nouiz:
To do it well, we should test that we always use the good dtype in
make_node for the output dtype and not cast it.

Can you make sure there is tests for all inputs dtype to make sure the
output dtype is the right one? Then we won't need to port that.

good finding.

theano/sandbox/solve.py
slefrancois: Should we make the dtype test as part of this PR, or save it for the next CCW?

slefrancois:
jenkins test this

nouiz:
It should be short, so I would include it in this PR.

theano/sandbox/solve.py
```diff
@@ -1,75 +1,7 @@
-from __future__ import absolute_import, print_function, division
+import warnings
+from theano.tensor.slinalg import solve
```
slefrancois: You need to add this file to the whitelist in theano/tests/test_flake8, otherwise it generates an unused import error.

tfjgeorge:
I updated the PR with @nouiz  and @slefrancois comments, but am not sure how to tell github that I did the requested changes

theano/tensor/tests/test_slinalg.py
```diff
@@ -229,6 +229,45 @@ def test_solve_correctness(self):
         assert numpy.allclose(scipy.linalg.solve_triangular(U_val, b_val, lower=False),
                               upper_solve_func(U_val, b_val))
 
+    def test_solve_dtype(self):
```
slefrancois: We should test integer to float too, that's where the out type inference is likely to break.

theano/tensor/slinalg.py
```diff
@@ -202,10 +202,16 @@ def make_node(self, A, b):
...
-        otype = tensor.tensor(
+        if ((A.dtype == 'float32' and b.dtype == 'float32')
+                or (A.dtype in ['int8', 'int16'] and b.dtype == 'float32')
+                or (b.dtype in ['int8', 'int16'] and A.dtype == 'float32')):
+            o_dtype = 'float32'
+        else:
+            o_dtype = 'float64'
+        x = tensor.tensor(
             broadcastable=b.broadcastable,
-            dtype=(A * b).dtype)
```
slefrancois: Does this method of dtype determination fail the new tests? It would be simpler than explicit conditionals.

tfjgeorge: Yes it does, e.g. `matrix('int32') * matrix('int32') => matrix('int32')` but solve will give a float. A more compact implementation could be to actually perform scipy.linalg.solve on very small matrices (e.g. 1x1) with correct dtypes and get the dtype of the output. This would probably slow down building the graph a little bit compared to conditionals.

theano/tensor/tests/test_slinalg.py
```diff
@@ -229,6 +231,27 @@ def test_solve_correctness(self):
...
+        b_val = numpy.ones((2, 1))
+
+        for A_dtype, b_dtype in itertools.product(dtypes, dtypes):
+            A = tensor.matrix(dtype=A_dtype)
+            b = tensor.matrix(dtype=b_dtype)
+            x = solve(A, b)
+            fn = function([A, b], x)
+            x_result = fn(A_val.astype(A_dtype), b_val.astype(b_dtype))
+            print(A_dtype, b_dtype)
+            print('x', x.dtype, x_result.dtype, x.dtype == x_result.dtype)
```
slefrancois: We don't need the print() in the final code, nose_tests will print the error message from the assert if it fails.

theano/tensor/tests/test_slinalg.py
```diff
@@ -229,6 +231,27 @@ def test_solve_correctness(self):
         assert numpy.allclose(scipy.linalg.solve_triangular(U_val, b_val, lower=False),
                               upper_solve_func(U_val, b_val))
 
+    def test_solve_dtype(self):
+        if not imported_scipy:
+            raise SkipTest("Scipy needed for the Solve op.")
+
+        dtypes = ['int8', 'int16', 'int32', 'int64',
+                  'float32', 'float64']
```
slefrancois: Can you add float16 to the tests, and see if our type inference in the make_node is still correct?

theano/tensor/slinalg.py
```diff
@@ -202,10 +202,16 @@ def make_node(self, A, b):
...
         assert A.ndim == 2
         assert b.ndim in [1, 2]
-        otype = tensor.tensor(
+        if ((A.dtype == 'float32' and b.dtype == 'float32')
+                or (A.dtype in ['int8', 'int16'] and b.dtype == 'float32')
+                or (b.dtype in ['int8', 'int16'] and A.dtype == 'float32')):
+            o_dtype = 'float32'
+        else:
+            o_dtype = 'float64'
+        x = tensor.tensor(
```
nouiz: I think we should use: o_dtype=theano.scalar.upcast(A.dtype, b.dtype)

tfjgeorge: It is not really what we want to achieve, counter example : 
```
In [13]: r = scipy.linalg.solve(A.astype('int32'), b.astype('int32'))
In [14]: r.dtype
Out[14]: dtype('float64')
```
but
```
In [15]: theano.scalar.upcast('int32', 'int32')
Out[15]: 'int32'
```

theano/tensor/tests/test_slinalg.py
tfjgeorge: It fails with float16. Are we supposed to support it ? The output type of scipy's solve is not documented

slefrancois: We try to have float16 work, but if the external libraries don't support it we live with it.

In this case, is it due to scipy itself, or our type inference method? My main worry was that float16 was not accounted for at all in the if else we use.

slefrancois:
jenkins test this please

slefrancois:
jenkins test this

slefrancois:
I'm good to merge once jenkins has passed.

nouiz:
There is a few pep8/flake8 problems:

https://travis-ci.org/Theano/Theano/jobs/176453920#L998

Also, this PR will need a rebase.

thanks.

theano/sandbox/solve.py
```diff
@@ -1,75 +1,7 @@
-from __future__ import absolute_import, print_function, division
+import warnings
+from theano.tensor.slinalg import solve  # noqa
```
slefrancois: The noqa doesn't solve Travis' error on missing python 3 compatibility imports. Probable we should go back to explicite whitelisting for flake8.

nouiz:
The problem isn't with python 3.3 only. The same problem happen with python 2.7. Just add the missing import at the top of the file:

slefrancois:
jenkins test this

slefrancois:
@nouiz, can we keep the merge commit from the rebase or will it pollute the master's history?

nouiz:
We don't like merge commit in PR. Note, merge commit aren't generated by rebase, but by "git merge" or other commant that call it internally (I think git fetch do that).

If it can be removed, it would be better.

tfjgeorge:
AFAIK git pull does it, git fetch does not. Anyway I deleted the merge commit and rebased to Theano/master. Can you confirm it is the proper process ?

slefrancois:
jenkins test this

slefrancois:
Looks good to me. @nouiz, do you want to have a final look and merge once jenkins passes?

nouiz:
I'm ok to merge now. No need to wait for jenkins as this don't change gpu code.

PR #5211, opened 2016-11-10
Port opt from the old backend.

abergeron:
Since gpu_elemwise_alloc is reused from tensor, I've ported it

I've looked at local_gpu_extract_diagonal, but since we don't have extract_diag, there is little point for now.

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -756,6 +756,8 @@ inplace_gpu_elemwise_opt = tensor.opt.InplaceElemwiseOptimizer(
 optdb.register('gpua_inplace_opt', inplace_gpu_elemwise_opt, 75,
                'inplace_elemwise_optimizer', 'fast_run', 'inplace', 'gpuarray')
 
+register_opt(tensor.opt.local_useless_elemwise)
+
 
 @register_opt('fast_compile')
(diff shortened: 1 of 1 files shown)


nouiz:
jenkins and travis fail.

theano/gpuarray/opt.py
```diff
@@ -756,6 +756,11 @@ def max_inputs_to_GpuElemwise(node):
 optdb.register('gpua_inplace_opt', inplace_gpu_elemwise_opt, 75,
                'inplace_elemwise_optimizer', 'fast_run', 'inplace', 'gpuarray')
 
+gpu_elemwise_alloc = gof.local_optimizer([GpuElemwise])(
+    tensor.opt.local_elemwise_alloc_op(GpuElemwise, GpuAlloc, GpuDimShuffle)
+)
+
+tensor.opt.register_specialize_device(gpu_elemwise_alloc)
```
abergeron: As far as I know, it transforms elemwise(alloc(), other) into elemwise(dimshuffle(), other) if the shapes allow it.

It's not required per se, but I sense it might make things go faster.

abergeron:
After discussion, I've removed gpu_elemwise_alloc and registered local_useless_elemwise instead.

We can always add gpu_elemwise_alloc later if we discover it is actually useful.

PR #5223, opened 2016-11-14
DestroyHandler should not overwrite existing rval[app] values

gvtulder:
In PR #5190 I have been having problems with the order in which multiple optimizations run. In some cases, this might incorrectly optimize an Op into an inplace version, even if that that shouldn't be allowed because it destroys variables that are still needed.

I must admit that I'm not entirely sure if this is the correct solution, but this patch seemed to fix my problem. The DestroyHandler is looping through the destroyers and destroyed input to create a dictionary `rval`. It seems to assume that `rval[app]` will always be empty, so it just sets the value. This usually works fine, except in a few rare cases where `rval[app]` already contains a list of nodes. Because the code overwrites `rval[app]`, it will forget about the nodes that were already there. In just the right circumstances, this might cause the optimizer to add in-place Ops that shouldn't be run in-place.

Keeping any existing values in `rval[app]`, as this patch does,  fixes the problem in my case.

It seems quite difficult to trigger this issue. The simplest example I could make can be found [in this example script](https://gist.github.com/gvtulder/e78b4e71e58826350e7afa19845db935). This example has a DummyOp that can work in-place on both of the inputs. The first input is important and should not be modified in-place, because it is reused by the gradient Op, but the second input is not reused and can be changed. There are two separate optimizations that enable the in-place option for each of the two inputs separately. The script runs the computations twice: without the optimizations, which produce the correct answer 114, and with the optimizations, which gives the incorrect answer 123 (caused by the in-place modification of x).

A few notes:
1. I ran this on the gpu because the GpuFromHost makes a copy of the original inputs. This is necessary to get variables that can be modified.
2. I could only get this to work if I also call `gpu_contiguous` on all inputs. This seems to break some dependency chain somewhere. Without `gpu_contiguous` or with only cpu the optimizations would not be accepted.
3. This problem depends on the order of the two optimizations. The inplace_harmless must be enabled before inplace_harmful. If you rename the two optimizations to change the order (opt. 2 to 1 and 1 to 2) the problem doesn't appear.

### DIFF
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -950,6 +950,8 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
 
             # add destroyed variable clients as computational dependencies
             for app in self.destroyers:
+                # keep track of clients that should run before the current Apply
+                root_clients = OrderedSet()
                 # for each destroyed input...
                 for output_idx, input_idx_list in iteritems(app.op.destroy_map):
                     destroyed_idx = input_idx_list[0]
@@ -1015,12 +1017,14 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
 
                     # add the rule: app must be preceded by all other Apply instances that
                     # depend on destroyed_input
-                    root_clients = OrderedSet()
                     for r in root_impact:
                         assert not [a for a, c in self.clients[r].items() if not c]
                         root_clients.update([a for a, c in self.clients[r].items() if c])
-                    root_clients.remove(app)
-                    if root_clients:
-                        rval[app] = root_clients
+
+                # app itself is a client of the destroyed inputs,
+                # but should not run before itself
+                root_clients.remove(app)
+                if root_clients:
+                    rval[app] = root_clients
 
         return rval
--- a/theano/gof/tests/test_destroyhandler.py
+++ b/theano/gof/tests/test_destroyhandler.py
@@ -90,6 +90,10 @@ add_in_place_2 = MyOp(2, 'AddInPlace', dmap={0: [0]},
 add_in_place_3 = MyOp(2, 'AddInPlace', dmap={0: [0]},
                       destroyhandler_tolerate_aliased=[(0, 1)])
 dot = MyOp(2, 'Dot')
+multiple = MyOp(2, 'Multiple', nout=2)
+multiple_in_place_0 = MyOp(2, 'MultipleInPlace0', nout=2, dmap={0: [0]})
+multiple_in_place_1 = MyOp(2, 'MultipleInPlace1', nout=2, dmap={1: [1]})
+multiple_in_place_0_1 = MyOp(2, 'MultipleInPlace01', nout=2, dmap={0: [0], 1: [1]})
 
 
 def inputs():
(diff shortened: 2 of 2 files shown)


f0k:
This seems reasonable. Travis failed because it exceeded the runtime limit, but this seems to affect all recent builds (https://travis-ci.org/Theano/Theano/pull_requests). Good job tracking this down!

gvtulder:
In case it helps, I've now added a test for the DestroyHandler that demonstrates this problem.

While creating this test, I found out that the problem doesn't even depend on the order of the optimizations. The first part of the tests tries a single optimization that makes the Op run in-place on all inputs at the same time. That shouldn't be allowed, but it is.

As far as I understand now, the conditions for this problem are:
1. Op A should be able to run in-place on at least two inputs.
2. There should be an Op B that depends on the first input of Op A and on an output of Op A, but it should not depend on the second input of Op A.
3. There should be an Op C that depends on the second input of Op A but not on the output of Op A.

Then the DestroyHandler orderings loop does the following when evaluating an in-place version of Op A:
1. Looking at the first input, it records that Op B should run before Op A.
2. Looking at the second input, it records that Op C should run before Op A. At this point it forgets about Op B.
Since Op C can run before Op B, the in-place optimization is accepted.

abergeron:
This looks ok, but I'm not that familiar with DestroyHandler.

The problem with Travis is not the same as all the recent PR, but it just seems it hanged during the installation of miniconda on one of the steps.  I'll restart it.

lamblin:
Is there a way of caching the miniconda installer or its packages?
It seems like we have to restart a couple of builds for each PR.

I'll try to have a look at the code later today.

nouiz:
@notoraptor is already working on caching the download of miniconda.

Sorry for not checking this PR, too many stuff doing at the same time...

f0k:
> Is there a way of caching the miniconda installer or its packages?

Not sure if it helps, but we discarded miniconda in Lasagne/Lasagne@97c6598 and just rely on caching the pip wheels. This works both for numpy and scipy. Installing the prerequisites for Lasagne (including numpy, scipy, Theano) takes 10 to 20 seconds (https://travis-ci.org/Lasagne/Lasagne/jobs/175864559).

nouiz:
I didn't forget about this PR. The person that wrote that code is James and he isn't here anymore. I already worked in that part of the code, but never got a full understanding of all the details. As this is sensible code, I would prefer to take a few hours to understand completely all the different part.

I'll try to hide somewhere for half a day next week to do that. But for what I understand now, the PR look good. I'll probably be able to make a plan on how to do part of the work online to speed it up for big graph from that understanding.

theano/gof/destroyhandler.py
```diff
@@ -1016,6 +1016,8 @@ def orderings(self, fgraph):
                     # add the rule: app must be preceded by all other Apply instances that
                     # depend on destroyed_input
                     root_clients = OrderedSet()
+                    if app in rval:
+                        root_clients.update(rval[app])
```
nouiz: I didn't had the time to do the full review of the destroyhandler mechanism, but a local review of this diff gave us (me and @lamblin) that this PR is good to merge.

One small comment that would do a small code clean up and remove some potential slowdown when the change isn't needed:

Can you remove the 2 new lines and move line 1018 after line 952 (so between the 2 for loop). This will remove the need for the extra if, so no slow down will happen.

At the same time, can you remove one indentation level of ilne 1025 and 1026?

After that small change, we can merge the PR. If you can do it quickly, we want to make an release candidate of Theano and would like to include this fix.

thanks for all the work on this bugfix!

gvtulder:
@nouiz: Something like this?

nouiz:
Excellent. I'll merge when travis and jenkins pass.

thanks

PR #5226, opened 2016-11-14
Update TRAVIS config: tests splitted

notoraptor:
theano/tensor tests are splitted into:
- all theano/tensor tests excluding test_basic
- just test_basic

`mode=FAST_COMPILE` tests are splitted into:
- all tests excluding theano/tensor tests
- just theano/tensor tests

@nouiz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,13 +52,13 @@ matrix:
   fast_finish: true
   include:
     - python: "3.3"
-      env:
-        - PART="--exclude-dir=theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE"
-        - PART="theano/tensor" THEANO_FLAGS="mode=FAST_COMPILE"
+      env: PART="--exclude-dir=theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE"
+    - python: "3.3"
+      env: PART="theano/tensor" THEANO_FLAGS="mode=FAST_COMPILE" 
     - python: "2.7"
-      env:
-        - PART="--exclude-dir=theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE,floatX=float32"
-        - PART="theano/tensor" THEANO_FLAGS="mode=FAST_COMPILE,floatX=float32"
+      env: PART="--exclude-dir=theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE,floatX=float32"
+    - python: "2.7"
+      env: PART="theano/tensor" THEANO_FLAGS="mode=FAST_COMPILE,floatX=float32"
 
 script:
   - export THEANO_FLAGS=$THEANO_FLAGS,warn.ignore_bug_before=all,on_opt_error=raise,on_shape_error=raise,gcc.cxxflags=-pipe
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,15 +43,20 @@ install:
...
+  - PART="theano/tensor/tests/test_basic.py"
 
 matrix:
   fast_finish: true
   include:
     - python: "3.3"
-      env: PART="-e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE"
+      env:
+        - PART="-e theano/tensor test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE"
+        - PART="theano/tensor" THEANO_FLAGS="mode=FAST_COMPILE"
```
nouiz: This won't do what we want I think. I think it should be:

        - PART="-e theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE"
        - PART="theano/tensor -e test_flake8.py" THEANO_FLAGS="mode=FAST_COMPILE"

I'm not sure we can pass many -e parameter to nosetests, double check that.

Also, line 46 give the full path to test_basic.py, while here we pass only the file name. Make sure that passing the full path is valid. You can check in git log of this file, we where doing that in the path.

notoraptor:
New update @nouiz !

notoraptor:
New correction. I find a travis validator here: http://lint.travis-ci.org/

Based on it, I have modified "matrix" section in the config file.

nouiz:
I killed the "push" travis run. When the "pr" run pass, we can merge.

nouiz:
In fact, I forgot, when the "pr" travis run pass, we need to check the split run output and check that we executed the right number of tests compared to the previous run in travis. Just to be sure we didn't loose some tests. I think the PR is good, but it is better to double check that.

notoraptor:
Hello !

As some runs yet take from 40 minutes to more than 1 hour, I suggest to split the tests differently. This is my suggestion:

original: `"theano/tensor"`
split:
- `"theano/tensor --exclude-dir=theano/tensor/nnet"`
- `"theano/tensor/nnet"`

PS: `test_basic.py` juste takes 4 to 6 minutes, so it does not make such difference, then I suggest not considerer it as a split point !

For `mode=FAST_COMPILE`:
original: `"-e test_flake8.py ."`
split:
- `"--exclude-dir=theano/tensor -e test_flake8.py ."`
- `"theano/tensor --exclude-dir=theano/tensor/nnet"`
- `"theano/tensor/nnet"`

(we isolate `theano/tensor` and then we split it the same manner as above)

Also, about comparing number of tests to previous runs, it seemes that TRAVIS runs on a clone of the master branch of the official Theano repository, on which it applies the tested PR. At the top of any run log ( e.g. https://travis-ci.org/Theano/Theano/jobs/176130787 ), we have the following commands:

```
$ git clone --depth=50 https://github.com/Theano/Theano.git Theano/Theano
$ cd Theano/Theano
$ git fetch origin +refs/pull/5226/merge:
$ git checkout -qf FETCH_HEAD
```

So, to compare correctly the number of tests executed, for every travis job (i.e. every value of "PART"), it seems better to:
1. get the time the travis job started,
2. find the latest commit to theano/master that was merged just before the travis job started,
3. use this commit as reference.

But taking account that some commits do not have TRAVIS runs !



@nouiz

PR #5232, opened 2016-11-15
noticed that max_and_argmax fails, test + fix

ferrine:
noticed that max_and_argmax fails, test + fix, see this [travis message](https://travis-ci.org/pymc-devs/pymc3/jobs/175961733#L1020)


nouiz:
thanks!

twiecki:
Thanks @ferrine

Issue #790, opened 2012-07-26
enable local_sum_broadcastable opt at canonicalization


nouiz:
When enabled, this break this stability optimization:

theano/tensor/tests/test_opt.py:T_local_sum.test_local_sum_broadcast_some_0

nouiz:
done by gh-5234

PR #5234, opened 2016-11-16
Useless sum in grad removed and test added

ReyhaneAskari:
Hi @nouiz ,
Regarding the Lasagne/Lasagne#765 issue , this fix will remove a useless ```sum``` when computing the gradient. A test is also added to check the graph and test the output.


theano/tensor/tests/test_elemwise.py
```diff
@@ -1242,6 +1242,36 @@ def func(x, y, z):
                                [numpy.asarray([-1., 0.5, 2.]), 0., 1.])
 
 
+def test_grad_useless_sum():
+    """Test absence of useless sum.
+
+    When an operation (such as T.mul) is done on a broadcastable vector and
+    a matrix, the gradient in backward path is computed for the broadcasted
+    vector. So a sum reverts the broadcasted vector to a vector. In the case
+    of operations on two vectors, the sum should not be generated.
```
nouiz: on two vectors -> on two broadcasted vectors

In the case of the non broadcasted vector, we didn't had problems.

theano/tensor/tests/test_elemwise.py
```diff
@@ -1242,6 +1242,36 @@ def func(x, y, z):
...
+
+    This test checks whether there is a useless sum in the gradient
+    computations.
+    """
+    x = tensor.TensorType(theano.config.floatX, (True,))('x')
+    l = tensor.log(1.0 - tensor.nnet.sigmoid(x))[0]
+    g = tensor.grad(l, x)
+    nodes = theano.gof.graph.ops([x], [g])
+
+    f = theano.function([x], g)
```
nouiz: Our tests should pass with mode=FAST_COMPILE. I think this new test will fail in that case. Travis will confirm this. I think this extra parameter will make it work in that cases: mode=theano.compile.get_default_mode().including("canonizer")

nouiz:
As discussed, we need to fix finish the fix that this PR started to gh-790 before we can merge. Otherwise, there is chance that we cause regression in some not tested cases.

ReyhaneAskari:
@nouiz , What is "regression"? After merging this can we close Theano/Theano#790 ?

nouiz:
If all tests pass with your last commit, then we can close gh-790.

regression. I mean, having some user code that was well optimized stop
being well optimized.

PR #5237, opened 2016-11-16
Small deprecation fix

Sentient07:
fix the pydot printing error that I was able to produce locally


theano/tests/test_printing.py
```diff
@@ -83,7 +83,7 @@ def test_pydotprint_variables():
     try:
         theano.printing.pydotprint(x * 2)
         if not theano.printing.pd.__name__ == "pydot_ng":
-            theano.printing.pydotprint_variables(x * 2)
+            theano.printing.pydotprint(x * 2)
```
nouiz: This isn't the right fix. This test was there to test pydotprint_variables, so we can't do that to remove the warning.

I checked, we deprecated that function in commit 8d0e91a5d5cf6d72a9a7d2595a816b1f0def7608 in october 2014. I think we should remove the test and remove the function at the same time. It was deprecated for long enough. Can you do that?

thanks for finding this old stuff to remove!

nouiz:
I'm killing travis to help it not be overloaded.

nouiz:
ok to merge when travis pass. I restarted a sub job of travis that had a timeout during install.

PR #5254, opened 2016-11-18
Removed save_every_N sequence size limitiation.

### LINKED ISSUE #5106: Generalize Scan memory/speed trade-off
https://github.com/Theano/Theano/issues/4124 issue was about adding that.

Current implementation have this restriction that we want to lift:
- [ ] The value of `save_every_N` divides the number of steps the scan will run without remainder.

I'm not sure we should lift this limit. What do you think?
- Only singly-recurrent and non-recurrent outputs are used. No multiple recurrences.

Thrandis:
Fixes #5106

I made the changes to remove the limitation that the length of the sequence has to be a multiple of `save_every_N`. It still requires proper testing, and also I still need to check the behavior when the user pases `n_steps`.

### DIFF
--- a/theano/scan_module/scan_checkpoints.py
+++ b/theano/scan_module/scan_checkpoints.py
@@ -53,13 +53,10 @@ def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
 
     n_steps
         ``n_steps`` is the number of steps to iterate given as an int
-        or Theano scalar. If any of the input sequences do not have
-        enough elements, scan will raise an error. If the **value is 0**
-        the outputs will have **0 rows**. If the value is negative,
-        ``scan`` will run backwards in time. If the ``go_backwards`` flag
-        is already set and also ``n_steps`` is negative, ``scan`` will run
-        forward in time. If n_steps is not provided, ``scan`` will figure
-        out the amount of steps it should run given its input sequences.
+        or Theano scalar (> 0). If any of the input sequences do not have
+        enough elements, scan will raise an error. If n_steps is not provided,
+        ``scan`` will figure out the amount of steps it should run given its
+        input sequences.
 
     save_every_N
         ``save_every_N`` is the number of steps to go without storing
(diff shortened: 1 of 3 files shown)


Thrandis:
@lamblin @nouiz Is there a way to check that `n_steps > 0` when `n_steps` is symbolic? Because it may break the padding otherwise...

nouiz:
We could add an assert in the graph, but this have some inconvenience.

What happen when n_steps is negative? If it raise an error, even not very clear, it is fine for me. We don't support that for a long time. Just add a comment in the code where the reshape is added.

theano/scan_module/scan_checkpoints.py
```diff
@@ -108,8 +105,10 @@ def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
     o_n_steps = theano.tensor.cast(theano.tensor.ceil(n_steps / save_every_N),
                                    'int64')
     i_n_steps = save_every_N * theano.tensor.ones((o_n_steps,), 'int64')
-    i_n_steps = theano.tensor.inc_subtensor(i_n_steps[-1],
-                                            - n_steps % save_every_N)
+    mod = n_steps % save_every_N
```
nouiz: If you add support for neg value, test it.

Thrandis:
@nouiz I changed the `padding` flag.

For the `n_steps`, it will cause the `scan` to not behave properly, so it would be nice to make sure it is positive. But if it is the case also for the "classic" `scan`, we could do the check there.

nouiz:
I think the normal scan raise an error for negative n step. Can you double check that?

Thrandis:
@nouiz no, if `n_steps` is negative, `scan` will go backwards.

nouiz:
The execution engine don't support it anymore:

https://github.com/Theano/Theano/blob/master/theano/scan_module/scan_op.py#L1142

How did you tested that? With scan_with_checkpoints or just scan?

Thrandis:
@nouiz I only looked at the theano function `scan`, not its internals actually.

So everything is fine then. Can you relaunch travis, it seems to have failed somehow.

theano/scan_module/scan_checkpoints.py
```diff
@@ -96,16 +101,29 @@ def scan_checkpoints(fn, sequences=[], outputs_info=None, non_sequences=[],
...
 
-    # Compute the number of steps of the inner and of the outer scan
-    o_n_steps = theano.tensor.cast(n_steps / save_every_N, 'int64')
-    i_n_steps = save_every_N
+    # Compute the number of steps of the outer scan
+    o_n_steps = theano.tensor.cast(theano.tensor.ceil(n_steps / save_every_N),
+                                   'int64')
+    i_n_steps = save_every_N * theano.tensor.ones((o_n_steps,), 'int64')
+    mod = n_steps % save_every_N
+    sign = theano.tensor.sgn(mod)
```
nouiz: We do not support negative steps in scan anymore. So go back to the simplest earlisest version. Where did you saw in the doc that we support negative steps? We should remove that from the doc.

Thrandis: @nouiz So this is not to support `n_steps` negative. This is to compute how many steps the last inner scan must run.

And for the negative steps, it is in the docstring of `scan.py`

Issue #5225, opened 2016-11-14
Add an easy way of redefining gradients


lamblin:
The following feature has been requested a couple of times: being able to define a new `Op`, where the forward operation (like `perform`) would be defined as a symbolic expression, and the gradient (`grad`) would be defined explicitly (not as the result of `theano.grad` on the forward expression).
This is useful in at least two cases:
- when we have an optimized formulation of a gradient, but Theano does not optimize it that way automatically
- when we want to redefine the backpropagation of an operation to something that is not mathematically equivalent to the true gradient.

The goal would be to have an subclass of `Op` where we could define `forward` and `grad`, and with an optimization replacing it by the symbolic expression defined in `forward`.

`OpFromGraph` has some useful bits, but also things that would not apply, and that would make the optimization difficult.
The plan would be to refactor `OpFromGraph` in 3 classes:
- a base class with everything that should be shared (maybe including the definition of `perform` given the internal graph)
- the current `OpFromGraph`
- the new class

and then have an optimization in the beginning (maybe the `useless` phase) replacing the new class with its internal graph.

nouiz:
The optimization should be registered in both fast_run and fast_compile.

khaotik:
I'm planning for a implementation. Not sure if suitable for merge with master. My plans so far:

1. make a base class `CompoundOp`, subclass `CompiledCompoundOp` and `InlineCompoundOp`.

2. make a API function `compound_op(inputs, outputs, inline=False, **kwargs)`

3. `OpFromGraph=CompiledCompoundOp` for compatibility

4. add optimization

Is this OK?

nouiz:
The principe is good. I don't like Compound as this is too general. We have
many other ops that are compound op already, but done manualy for other
reason. So I would use BaseOpFromGraph, OpFromGraphOp and InlineOpFromGraph
name instead. If you have other name to suggest don't hesitate.

thanks

lamblin:
@khaotik We assigned @adbrebs to work on that ticket already.

lamblin:
As a general note, tickets tagged "CCW" are the ones we plan on having students work on.

nouiz:
This don't give any timeline. Some times are tagged CCW for a very long
time and we still didn't give them. Mostly, this mean it is better to ask
an update before starting them. If you are interrested, don't skip them if
they are tagged CCW, just ask.

khaotik:
[...]

Issue #1603, opened 2013-11-13
Finish/document OpFromGraph


nouiz:
We have a meta op called OpFromGraph() defined in theano/compile/builders.py.

This is thing that are relatively fast to do and are needed to make it usable:
- [ ] Document it so that user know how to user it.
- [x] test if it work with shared variable as input and if not make it work. Add test for that.
- [ ] Move it to the GPU. We can do it quickly for the old back-end, move all float32 inputs to the GPU. Otherwise, we need to compile the inner function, see which inputs get moved to the GPU, then create a new OpFromGraph with the corresponding input to the GPU. https://github.com/Theano/Theano/pull/2982

This is think that should be done to make it complete, but don't prevent people to use it:
- [x] Move the grad work in the grad() method. This should remove the grad_depth parameter
- [x] Make the grad use only 1  op for all outputs, otherwise, this would duplicate computation.
- [x] Opt to unfold it
- [x] implement infer_shape gh-3117
- [x] implement Rop
- [x] Allow the grad and Rop to be overrided.
- [ ] Opt to remove unused inputs/outputs (even more useful when better grad)
- [ ] implement c_code when all ops have c code.
- [ ] Implement __hash__ and __eq__
- [ ] Make the grad return DisconnectedType
- [ ] updates? givens?
- [ ] test pickle of this op
- [ ]  test random numbers with this op.

This is more an ideas that should be discussed before implementation:
- [ ] Allow the inner graph to work inplace on the inputs?
- [ ] implement optimization to remove that op to the global graph when it is small? This allow better optimization

I put it tentatively in 0.6.1.

nouiz:
I started a branch some time ago that started that, it isn't finished. I do not plan to work on it in the short term:

https://github.com/nouiz/Theano/tree/op_from_graph

nouiz:
For the next part, can you finish my PR gh-2982? I don't have time for this and I think it is one of the main missing feature.

ChienliMa:
OK, I try to understand your codes first.

ChienliMa:
I just realize that I have some problem to finish this task. 

The worst one is that I cannot reach any PC with Nvidia GPU. I don't think I can run the test without one Nvidia card.  I've apply for a GPU instance on AWS, and I will inform you if I can work on it.

Do you think I can finish this task without a PC with Nvidia GPU? 
If not, I think it's better to switch to another task.

Sentient07:
[...]

PR #5255, opened 2016-11-19
Upgraded OpFromGraph with inline support and gradient override

khaotik:
For #5225. Allowing inlined `OpFromGraph` at compile time and overriding gradient.

### Changes:
- ~~Refactored `OpFromGraph` into `OpFromGraphBase` (abstract) and subclass `OpFromGraphPrecompiled`,  `OpFromGraphInline`~~
- ~~`OpFromGraph = OpFromGraphPrecompiled` for compatibility~~
- OpFromGraph `__init__` method now have `inline` argument
- Can now override gradient/Rop for `OpFromGraph*` by supplying `grad_overrides` or `rop_overrides`  parameters respectively.
- Added an graph optimizer to expand inlined `OpFromGraph` at compile time.
- ~~new convenient function `op_from_graph(inputs, outputs, inline=False, grad_overrides=None, rop_overrides=None, **kwargs)`~~
- new tests for `compile/tests/test_builders.py`: `nested`, `grad_override` and `rop_override`
- tests in `compile/tests/test_builders.py` now test against both inline/precompiled mode.
- `name` and `__str__` for better debugging
- Docs

### Problems / Ideas:
- ~~Instead of refactoring `OpFromGraph` into 3 classes, how about adding a `inline` parameter to `OpFromGraph`, resulting simpler code?~~
- Made a commit to combine 3 classes back into one single `OpFromGraph`. This is different to the original plan on dividing `OpFromGraph`, but I think this makes the code cleaner. Is this OK?

### Related
#1603 #2982

### DIFF
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -340,11 +340,10 @@ class OpFromGraph(gof.Op):
         return self._grad_op(*(list(inputs) + list(output_grads)), return_list=True)
 
     def make_node(self, *inputs):
-        for input, type in zip(inputs, self.input_types):
-            if not type == input.type:
-                raise TypeError("Wrong type, expected %s but got %s" %
-                                (type, input.type))
-
+        num_expected_inps = len(self.local_inputs) - len(self.shared_inputs)
+        if len(inputs) != num_expected_inps:
+            raise ValueError("Expected %d inputs, got %d" % (num_expected_inps, len(inputs)))
+        inputs = [inp_t.filter_variable(inp) for inp, inp_t in izip(inputs, self.input_types)]
         apply_node = gof.Apply(
             self, list(inputs) + self.shared_inputs,
             [type() for type in self.output_types])
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -1,6 +1,6 @@
 """Define new Ops from existing Ops"""
 from __future__ import absolute_import, print_function, division
-from functools import reduce
+from functools import reduce, partial
 from collections import OrderedDict
 
 import theano
@@ -62,7 +62,7 @@ class OpFromGraph(gof.Op):
         Defaults to ``None``.
 
         ``None`` : No value, gives NullType()
-        ``0`` : zero value, gives DisconnectedType()
+        ``0`` : zero value, gives zeros_like(...)
         ``...`` : Do not override, use default R_op() result
 
         OpFromGraph instance : Override with another OpFromGraph, should
@@ -92,14 +92,15 @@ class OpFromGraph(gof.Op):
           local_outputs)
         - c_code() to remove the double overhead?
         - grad() make it support DisconnectedType and the new interface
-        - extend to lop_overrides?
+        - extend grad() to L_op
+        - add support for NullType and DisconnectedType when R_op supports them
         - check how it works with updates.
         - add test with constant as input or inside the inner graph.
         - Add support for the GPU? Probably just need an opt to remove transfer
         - Add support to pickle this Op.
         - Add support/test with random generator
         - Add optimization to removing unused inputs/outputs
-        - Add optimization to work inplace when not inline
+        - Add optimization to work inplace on inputs when not inline
 
     Notes
     -----
@@ -113,7 +114,7 @@ class OpFromGraph(gof.Op):
       ``fast_run`` mode.
     - It's recommanded to provide pure functions (no side effects like
       setting global variable) as callable(s). The callable(s) supplied
-      for overrding gradient/rop will be called only once at the first
+      for overriding gradient/rop will be called only once at the first
       call to grad/R_op, and will be converted to OpFromGraph instances.
 
     Examples
@@ -176,24 +177,34 @@ class OpFromGraph(gof.Op):
         #     grad: gradient Variable
         #     inp: the corresponding input of gradient Variable
         #
-        # Some Variable types cannot be used directly as OfG output such as
-        # NullType, or DisconnectedType.
-        #
-        # However a grad() call could return these types
+        # a grad() call could return instance of NullType() or DisconnectedType()
+        # which cannot be directly used in OfG
         #
         # Since we always use an OfG instance as self._grad_op, the current
         # workaround is to "remember" the special cases of the gradient and
         # replace them after self._grad_op is called.
         #
-        # This helper function changes invalid types into a filtered_type,
-        # and provides a overrider_type to be replaced at grad() call
+        # This helper function changes invalid types into a filtered_var,
+        # and provides a overrider_var to be replaced at grad() call
         #
         # For now, this converts NullType or DisconnectedType into zeros_like.
-        # other types are unmodified with overrider_type -> None
+        # other types are unmodified: overrider_var -> None
         if isinstance(grad.type, (NullType, DisconnectedType)):
-            return inp.zeros_like(), grad.type
+            return inp.zeros_like(), grad
         else:
             return grad, None
+
+    @staticmethod
+    def _filter_rop_var(inpJ, out):
+        # mostly similar to _filter_grad_var
+        if isinstance(inpJ.type, NullType):
+            return out.zeros_like(), inpJ
+        if isinstance(inpJ.type, DisconnectedType):
+            # since R_op does not have DisconnectedType yet, we will just
+            # make them zeros.
+            return out.zeros_like(), None
+        else:
+            return inpJ, None
 
     def __init__(
             self, inputs, outputs,
@@ -202,7 +213,7 @@ class OpFromGraph(gof.Op):
             name=None, **kwargs
     ):
         if not isinstance(outputs, list):
-            raise TypeError('outputs must be list, got %s' % outputs, outputs)
+            raise TypeError('outputs must be list, got %s' % type(outputs), outputs)
         for i in inputs + outputs:
             if not isinstance(i, gof.Variable):
                 raise TypeError(
@@ -263,21 +274,26 @@ class OpFromGraph(gof.Op):
 
         if isinstance(grad_op, OpFromGraph):
             self._grad_op_is_cached = True
-            self._grad_op_overrides_l = [None] * len(self.local_inputs)
+            self._grad_op_overrides_l = [None] * inp_len
             return
 
         output_grads = [out_t() for out_t in self.output_types]
+        fn_grad = partial(
+            theano.gradient.grad,
+            cost=None,
+            disconnected_inputs='ignore',
+            return_disconnected='Disconnected',
+            null_gradients='return',
+            known_grads=OrderedDict(izip(local_outputs, output_grads)))
+
         TYPE_ERR_MSG = 'Gradient override should be (single or list of)' \
             'OpFromGraph | Ellipsis | None | 0 | callable, got %s'
         # we need to convert _grad_op into an OfG instance
         if grad_op is Ellipsis:
-            self._grad_op_tflags = bytes(inp_len)
-            all_grads_l = theano.gradient.grad(
-                cost=None,
-                known_grads=OrderedDict(izip(local_outputs, output_grads)),
-                wrt=local_inputs,
-                disconnected_inputs='ignore')
+            gdefaults_l = fn_grad(wrt=local_inputs)
             all_grads_ov_l = [None] * inp_len
+            all_grads_l, all_grads_ov_l = izip(
+                *[OpFromGraph._filter_grad_var(grad, inp) for grad, inp in izip(gdefaults_l, local_inputs)])
         elif grad_op is None:
             all_grads_l = [inp.zeros_like() for inp in local_inputs]
             all_grads_ov_l = [self.ofg_null_t()] * inp_len
@@ -285,7 +301,7 @@ class OpFromGraph(gof.Op):
             all_grads_l = [inp.zeros_like() for inp in local_inputs]
             all_grads_ov_l = [self.ofg_discon_t()] * inp_len
         elif isinstance(grad_op, list):
-            goverrides_l = self._grad_op
+            goverrides_l = grad_op
             if len(goverrides_l) != inp_len:
                 raise ValueError(
                     'Need to override %d gradients, got %d' % (
@@ -293,18 +309,15 @@ class OpFromGraph(gof.Op):
             # compute non-overriding downsteam grads from upstreams grads
             # it's normal some input may be disconnected, thus the 'ignore'
             wrt_l = [lin for lin, gov in izip(
-                self.local_inputs, goverrides_l) if gov is Ellipsis]
-            gdefaults = iter(theano.gradient.grad(
-                cost=None,
-                known_grads=OrderedDict(izip(self.local_outputs, output_grads)),
-                wrt=wrt_l,
-                disconnected_inputs='ignore') if wrt_l else [])
+                local_inputs, goverrides_l) if gov is Ellipsis]
+            gdefaults = iter(fn_grad(wrt=wrt_l) if wrt_l else [])
             # combine overriding gradients
             all_grads_l = []
             all_grads_ov_l = []
-            for i, (inp, fn_gov) in enumerate(izip(local_inputs, goverrides_l)):
+            for inp, fn_gov in izip(local_inputs, goverrides_l):
                 if fn_gov is Ellipsis:
-                    gnext, gnext_ov = OpFromGraph._filter_grad_var(next(gdefaults), inp)
+                    gnext, gnext_ov = OpFromGraph._filter_grad_var(
+                        next(gdefaults), inp)
                     all_grads_l.append(gnext)
                     all_grads_ov_l.append(gnext_ov)
                 elif fn_gov is 0:
@@ -330,13 +343,14 @@ class OpFromGraph(gof.Op):
                     'Gradient overriding function should return a list, '
                     'got "%s"' % type(goverrides_l))
             all_grads_l, all_grads_ov_l = izip(
-                *[OpFromGraph._filter_grad_var(grad, inp) for grad, inp in izip(goverrides_l, local_inputs)])
+                *[OpFromGraph._filter_grad_var(
+                    grad, inp) for grad, inp in izip(goverrides_l, local_inputs)])
             if len(all_grads_l) != len(local_inputs):
                 raise ValueError(
                     'Gradient overriding function should return list of '
                     '%d outputs, got %d' % (inp_len, len(all_grads_l)))
-            all_grads_l = list(all_grads_l)
-            all_grads_ov_l = list(all_grads_ov_l)
+        all_grads_l = list(all_grads_l)
+        all_grads_ov_l = list(all_grads_ov_l)
         self._grad_op = type(self)(
             inputs=local_inputs + output_grads,
             outputs=all_grads_l,
@@ -347,65 +361,92 @@ class OpFromGraph(gof.Op):
         self._grad_op_is_cached = True
 
     def _recompute_rop_op(self):
+        local_inputs = self.local_inputs
+        local_outputs = self.local_outputs
+        out_len = len(local_outputs)
+        rop_op = self._rop_op
+
         if isinstance(self._rop_op, OpFromGraph):
             self._rop_op_is_cached = True
+            self._rop_op_overrides_l = [None] * out_len
             return
-        eval_points = [inp_t() for inp_t in self.input_types]
-        if self._rop_op is None:
-            self._rop_op = []
 
-        if isinstance(self._rop_op, list):
-            roverrides_l = self._rop_op
-            if len(roverrides_l) > len(self.local_outputs):
+        eval_points = [inp_t() for inp_t in self.input_types]
+        fn_rop = partial(
+            theano.gradient.Rop,
+            wrt=local_inputs,
+            eval_points=eval_points)
+        TYPE_ERR_MSG = 'R_op overrides should be (single or list of)' \
+            'OpFromGraph | Ellipsis | None | 0 | callable, got %s'
+        if rop_op is Ellipsis:
+            all_rops_l = fn_rop(f=local_outputs)
+            all_rops_ov_l = [None] * out_len
+        elif rop_op is None:
+            all_rops_l = [out.zeros_like() for out in local_outputs]
+            all_rops_ov_l = [self.ofg_null_t()] * out_len
+        elif rop_op is 0:
+            all_rops_l = [out.zeros_like() for out in local_outputs]
+            all_rops_ov_l = [None] * out_len
+        elif isinstance(rop_op, list):
+            roverrides_l = rop_op
+            if len(roverrides_l) != out_len:
                 raise ValueError(
-                    'Can override %d gradients at most, got %d' % (
-                        len(self.local_onputs), len(roverrides_l)),
-                    roverrides_l)
-            if len(roverrides_l) < len(self.local_outputs):
-                roverrides_l += [None] * (
-                    len(self.local_outputs) - len(roverrides_l))
+                    'Need to override %d Rop, got %d' % (
+                        out_len, len(roverrides_l)), roverrides_l)
             # get outputs that does not have Rop override
             odefaults_l = [
-                lo for lo, rov in izip(self.local_outputs, roverrides_l)
-                if not rov]
-            rdefaults_li = theano.gradient.Rop(
-                f=odefaults_l,
-                wrt=self.local_inputs,
-                eval_points=eval_points
-                )
-            rdefaults = iter(rdefaults_li if odefaults_l else [])
+                lo for lo, rov in izip(local_outputs, roverrides_l)
+                if rov is Ellipsis]
+            rdefaults_l = fn_rop(f=odefaults_l)
+            rdefaults = iter(rdefaults_l if odefaults_l else [])
             # combine overriding Rops
             all_rops_l = []
-            for out, rov in izip(self.local_outputs, roverrides_l):
-                if rov is None:
-                    all_rops_l.append(next(rdefaults))
-                elif rov is undef:
-                    all_rops_l.append(
-                        out.zeros_like().astype(theano.config.floatX))
+            all_rops_ov_l = []
+            for out, fn_rov in izip(local_outputs, roverrides_l):
+                if fn_rov is Ellipsis:
+                    rnext, rnext_ov = OpFromGraph._filter_rop_var(
+                        next(rdefaults), out)
+                    all_rops_l.append(rnext)
+                    all_rops_ov_l.append(rnext_ov)
+                elif fn_rov is 0:
+                    all_rops_l.append(out.zeros_like())
+                    all_rops_ov_l.append(None)
+                elif fn_rov is None:
+                    all_rops_l.append(out.zeros_like())
+                    all_rops_ov_l.append(self.ofg_null_t())
                 else:
-                    all_rops_l.append(rov(self.local_inputs, eval_points))
-        elif self._rop_op is undef:
-            all_rops_l = [
-                out.zeros_like().astype(theano.config.floatX)
-                for out in self.local_outputs]
+                    if not hasattr(fn_rov, '__call__'):
+                        raise TypeError(TYPE_ERR_MSG % fn_rov)
+                    rov, rov_ov = OpFromGraph._filter_rop_var(
+                        fn_rov(local_inputs, eval_points), out)
+                    all_rops_l.append(rov)
+                    all_rops_ov_l.append(rov_ov)
         else:
-            all_rops_l = self._rop_op(self.local_inputs, eval_points)
-            if not isinstance(all_rops_l, (tuple, list)):
-                all_rops_l = [all_rops_l]
-            if len(all_rops_l) != len(self.local_outputs):
+            if not hasattr(rop_op, '__call__'):
+                raise TypeError(TYPE_ERR_MSG % rop_op)
+            roverrides_l = rop_op(local_inputs, eval_points)
+            if not isinstance(roverrides_l, list):
+                raise TypeError(
+                    'Rop overriding function should return a list, '
+                    'got "%s"' % type(roverrides_l))
+            all_rops_l, all_rops_ov_l = izip(
+                *[OpFromGraph._filter_rop_var(
+                    rop, out) for rop, out in izip(roverrides_l, local_outputs)])
+            if len(all_rops_l) != out_len:
                 raise ValueError(
                     'Rop overriding function %s should return list of '
                     '%d outputs, got %d' % (
-                        self._rop_op,
-                        len(self.local_outputs),
-                        len(all_rops_l)),
-                    self._rop_op)
+                        self._rop_op, out_len,
+                        len(all_rops_l)), rop_op)
+            all_rops_l = list(all_rops_l)
+            all_rops_ov_l = list(all_rops_ov_l)
         self._rop_op = type(self)(
-            inputs=self.local_inputs + eval_points,
+            inputs=local_inputs + eval_points,
             outputs=all_rops_l,
             inline=self.is_inline,
             name=(None if self.name is None else self.name + '_rop'),
             on_unused_input='ignore')
+        self._rop_op_overrides_l = all_rops_ov_l
         self._rop_op_is_cached = True
 
     def get_grad_op(self):
@@ -447,10 +488,9 @@ class OpFromGraph(gof.Op):
             self._recompute_rop_op()
         ret_ofg_l = self._rop_op(
             *(list(inputs) + list(eval_points)), return_list=True)
-        ret_l = [{
-            self.TFLAG_NULL_T: self.ofg_null_t(),
-            self.TFLAG_DISCON_T: self.ofg_discon_t()
-        }[flag] if flag else ret_ofg for ret_ofg, flag in izip(ret_ofg_l, self._grad_tflags)]
+        ret_l = [
+            ret_ofg if ov is None else ov for ret_ofg, ov in izip(
+                ret_ofg_l, self._rop_op_overrides_l)]
         return ret_l
 
     def grad(self, inputs, output_grads):
@@ -459,9 +499,9 @@ class OpFromGraph(gof.Op):
         ret_ofg_l = self._grad_op(
             *(list(inputs) + list(output_grads)), return_list=True)
         ret_l = [
-            ret_ofg if ov is None else ov for ret_ofg, ov in izip(ret_ofg_l, self._grad_op_overrides_l)]
+            ret_ofg if ov is None else ov for ret_ofg, ov in izip(
+                ret_ofg_l, self._grad_op_overrides_l)]
         return ret_l
-
 
     def make_node(self, *inputs):
         num_expected_inps = len(self.local_inputs) - len(self.shared_inputs)
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -6,37 +6,17 @@ import theano
 from theano import gof
 from theano.compat import izip
 from theano.compile.function_module import orig_function
-from theano.compile import SharedVariable, rebuild_collect_shared
-from theano.compile import optdb
+from theano.compile import SharedVariable, rebuild_collect_shared, optdb
 from theano.gof import ops_with_inner_function
 from theano.gof.graph import io_connection_pattern
 
 
-class BaseOpFromGraph(gof.Op):
+class OpFromGraphBase(gof.Op):
     """
-    General syntax is similar to theano.function
-    Currently does not support 'updates' or 'givens'argument
-
-    Parameters
-    ----------
-
-    inputs: list of variables
-    outputs: list of variables
-    grad_overrides: None or function or list of [None|function]
-        Used to override default gradient routine.
-        Overriding function must take two list as inputs: original inputs
-        and upstream gradients
-        If is None, will use default gradient routine.
-        If is function, must return list of Variable.
-        If is list, each function must return a single Variable. The order
-            of the list must corresponds to inputs
-
-    Notes
-    -----
-
-    You can use regular function for grad_overrides, but the function
-    must take (list of) Variable as input and output.
+    base class for Ops with custom inner graph
     """
+    # NOTE: if you make a subclass of this, make sure add it under:
+    #     theano/compile/tests/test_builders.py
     def __init__(self, inputs, outputs, grad_overrides=None, **kwargs):
         if not isinstance(outputs, list):
             raise TypeError('outputs must be list', outputs)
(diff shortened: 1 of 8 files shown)


theano/compile/builders.py
```diff
@@ -115,40 +64,88 @@ def __hash__(self):
...
+                gs = [go if go else type(self)(
+                    grad_inps,
+                    (lambda g: g if g else (lambda *a:None))(
+                        theano.gradient.grad(
+                            cost=None,
+                            known_grads=upstream_grads,
+                            wrt=[inp],
+                            disconnected_inputs='ignore')
+                    ), on_unused_input='ignore'
+                ) for go, inp in izip(grad_ops_l, self.internal_inputs)]
```
f0k: Doing multiple `grad` calls is often slower than doing a single one wrt. multiple inputs at once. Would be good to do a single `grad` call for all the inputs you need it for, then create the `type(self)` instances.

theano/compile/builders.py
```diff
@@ -115,40 +64,88 @@ def __hash__(self):
...
+                # since OpFromGraphBase only accepts input sequence,
+                # additional filtering is needed
+                def grad_ops(inps, grds):
+                    # nonlocal gs, grad_ops_l
+                    return [(go(inps, grds) if ov else go(*(inps + grds)))
+                            for go, ov in izip(gs, grad_ops_l)]
+            else:
+                grad_ops = grad_ops_l
+            self.grad_ops = grad_ops
+        else:
```
f0k: The `else` case could be handled by the code above. You already support only providing *some* gradient overrides, just add something like this at the top:
```python
if self.grad_ops is None:
    self.grad_ops = [None] * len(self.internal_inputs)
```
Then make sure the code above does the same your `else` case does (i.e., do a single `grad` call).

theano/compile/builders.py
f0k: Actually, why should we create OpFromGraphs for the inline version? Is there an advantage over returning the graph directly?

theano/compile/builders.py
```diff
@@ -168,30 +165,175 @@ def infer_shape(self, node, shapes):
...
+            output[0] = variable.copy()
+
+
+class OpFromGraphInline(OpFromGraphBase):
+    """
+    The Op's inner graph is expanded into the outer graph at compile time
+    """
+    def perform(self, node, inputs, outputs):
+        raise RuntimeError(
+            type(self).__name__ + ' is not supposed to be executed at runtime')
```
lamblin: Yes, a perform would be helpful for test values and DebugMode.
We had to do that for abstract ops, I think it is the case here as well.

What we could have (again, like for the abstract ops) is a pseudo-optimization in a final pass that asserts there is no Inline Ops in the graph. If we are compiling without any optimization at all, then that optimization will not be triggered and the slow version will be used.

theano/compile/builders.py
lamblin: For the gradient of the inline version, we should just return the user-defined expression.
If a user wants an inline OpFromGraph as the gradient (to take second derivatives, for instance), then they can explicitly define their gradient to be an InlineOpFromGraph.

theano/compile/builders.py
nouiz: For an example, see the file theano/tensor/nnet/opt.py, search for local_abstractconv_check.

thanks.

theano/compile/builders.py
f0k: But then the base class needs to know whether it needs to create those `type(self)` instances or not, at which point it may really be easier to just add an `inline` argument (or a `precompile` argument with inverse semantics) to `OpFromGraph` instead of adding additional classes. What do you think?

khaotik: @f0k That's actually what I suggest in one of above **problems/ideas**. We should let the devs decide.

f0k: > That's actually what I suggest

Yes, I've seen that! I think that's a good idea. Let's hear what the others think.

khaotik:
@nouiz What about #2982? Should we finish that one first?

theano/compile/builders.py
```diff
@@ -168,30 +159,154 @@ def infer_shape(self, node, shapes):
...
+            [x, y, z], [e], grad_overrides=[None, rescale_dy, None])
+        e2 = op(x, y, z)
+        dx, dy, dz = grad(e2, [x, y, z])
+        fn = function([x, y, z], [dx, dy, dz])
+        # the graident wrt y is now doubled
+        fn(2., 3., 4.) # [1., 8., 3.]
+
+    """
+    return OpFromGraph(
+        inputs, outputs, inline=inline, grad_overrides=grad_overrides, **kwargs)
```
f0k: Since `op_from_graph` directly returns `OpFromGraph`, it may be better to keep the documentation in `OpFromGraph` and set `op_from_graph = OpFromGraph` as an alias in the end, if anything.

theano/compile/builders.py
```diff
@@ -115,40 +73,73 @@ def __hash__(self):
...
+            grad_op_l = self.grad_op
+            if len(grad_op_l) > len(self.local_inputs):
+                raise ValueError(
+                    'Can override %d gradients at most, got %d' % (
+                        len(self.local_inputs), len(grad_op_l)))
+            if len(grad_op_l) < len(self.local_inputs):
+                grad_op_l += [None] * (
+                    len(self.local_inputs) - len(grad_op_l))
+            wrt = [self.local_inputs[i] for i, go in
+                   enumerate(grad_op_l) if not go]
```
f0k: clearer: `if go is None`

f0k: maybe even clearer:
```python
wrt = [local_input for local_input, grad_override
       in izip(self.local_inputs, grad_op_l)
       if grad_override is None]
```
I find short variable names more difficult to follow.

theano/compile/builders.py
```diff
@@ -115,40 +73,73 @@ def __hash__(self):
...
+    def grad(self, inputs, output_grads):
+        if self.grad_op_is_cached:
+            return self.grad_op(inputs, output_grads)
+
+        if self.grad_op is None:
+            self.grad_op = []
+
+        # we need to convert a list into a single funtor
+        if isinstance(self.grad_op, list):
+            grad_op_l = self.grad_op
```
f0k: maybe `grad_overrides` would be a better name than `grad_op_l`?

theano/compile/builders.py
```diff
@@ -115,40 +73,73 @@ def __hash__(self):
...
+            wrt = [self.local_inputs[i] for i, go in
+                   enumerate(grad_op_l) if not go]
+            # compute non-overriding downsteam gradients from upstreams grads
+            # it's normal some input may be disconnected, thus the 'ignore'
+            ups_grads_d = dict(izip(self.local_outputs, output_grads))
+            nat_dns_grads = iter(theano.gradient.grad(
+                cost=None,
+                known_grads=ups_grads_d,
+                wrt=wrt,
+                disconnected_inputs='ignore'))
```
f0k: Again, not so sure about the variable names, `ups_grads_d` and `nat_dns_grads` are quite cryptic. Maybe specify `known_grads=` directly inline, and call the other variable `default_grads` or something, because they are not overridden?
Code-wise, it'd be good to wrap this in `if wrt:`, so you avoid the `grad` call if all gradients are overridden anyway. You can even leave `nat_dns_grads` undefined if `wrt` is empty, or set it to `iter([])` in the `else:` case.

theano/compile/builders.py
```diff
@@ -115,40 +73,73 @@ def __hash__(self):
...
+            # combine overriding gradients
+            dns_grads_l = [
+                go(self.local_inputs, output_grads) if go else next(nat_dns_grads) for go in grad_op_l]
```
f0k: Again, I'd prefer less cryptic variable names, but the implementation is good!

theano/compile/builders.py
khaotik: Here `<name>_l` for a list. Similar usage in `tensor/nnet/Conv3d.py` or `scan_module/scan_opt.py`.  I agree it's better to mention "overrides".

f0k:
> What about #2982? Should we finish that one first?

I'd say no. Even with the old code it's possible to create a GPU OpFromGraph if you need, and with the inline version GPU optimization will not be affected at all.

theano/compile/builders.py
khaotik: Hmm ... I didn't expect this to be so cryptic. I thought I mentioned "upstream/downstream" in comments so `ups_grads` or `dns_grads` should be clear. Yes, will make commits for more readability.

theano/compile/builders.py
```diff
@@ -115,40 +73,73 @@ def __hash__(self):
...
+            grad_ofg = type(self)(
+                inputs=self.local_inputs + output_grads,
+                outputs=dns_grads_l,
+                inline=self.is_inline, on_unused_input='ignore')
```
f0k: This will still create an `OpFromGraph` even for inline Op gradients, but on the positive side, this reduces code complexity. Otherwise we'd have to duplicate the cloning code here for the inline case. I'm not opposed to this any more.

theano/compile/builders.py
```diff
@@ -115,40 +73,73 @@ def __hash__(self):
...
+
+            def grad_op(inps, grds):
+                return grad_ofg(*(list(inps) + list(grds)))
+            self.grad_op = grad_op
```
f0k: Warning: This way an `OpFromGraph` will not be pickable once its `grad()` method was called. I'm not sure if this is a requirement? A workaround would be setting `self.grad_op` to the `OpFromGraph` instance directly, and handle the distinction in the `if cached` case. This may require rewriting this function from the current:
```python
if cached:
    return cached
cached = ...
return cached
```
to:
```python
if not cached:
    cache = ...
return cached
```
to avoid code duplication.

theano/compile/builders.py
f0k: > I thought I mentioned "upstream/downstream" in comments so `ups_grads` or `dns_grads` should be clear.

Yes, you did, and I was able to follow the code, but it still takes mental effort to read `ups_grads_d` or `go` and expand every occurrence into its actual meaning. Would be great to make it a little more readable, the `iter`/`next` in the list comprehension takes enough effort to verify :) (but it's a good solution!)

theano/compile/builders.py
khaotik: For my current implementation, there's not way cannot *ensure* `self.grad_op` is an `OpFromGraph` instance. It's probably just a single function supplied to `grad_overrides` at `__init__`.(which has different syntax to an OFG instance) I think the solution would be either always ensure `self.grad_op` a `OpFromGraph` instance or perform additional checks in `grad()`.

f0k: > For my current implementation, there's not way to *ensure* `self.grad_op` is an `OpFromGraph` instance.

Yes, I'm aware. My suggestion was to distinguish that in the case the gradient is already cached:
```python
if not self.grad_op_is_cached:
    ...
    self.grad_op_is_cached = True
if isinstance(self.grad_op, type(self)):
    return self.grad_op(*(inputs, grads))
else:
    return self.grad_op(inputs, grads)
```
To avoid having this distinction both at the top and at the bottom of the `grad()` method, I'm suggesting to change the implementation to first ensure the gradient is cached, then returning the cached gradient (instead of the current implementation of returning the cached gradient if possible, or creating the cache and returning the cached gradient otherwise.)

khaotik: @f0k I'm about to make an unified approach - make it always using an `OpFromGraph` instance. This reduces complexity and improves predictability. Assume a future utility module need to visit `grad_op`. If it can make sure `grad_op` is an OFG instance, it can perform less check and potentially cause less bugs. Also, OFG inst has internal graph which is more transparent than a lambda function.

f0k: > Assume a future utility module need to visit `grad_op`. If it can make sure `grad_op` is an OFG instance, it can perform less check and potentially cause less bugs.

But it can only assume so if the `grad` method has been called. Otherwise it's still what was given to the constructor.

> Also, OFG inst has internal graph which is more transparent than a lambda function.

Right. This would make it pickable even if the user gives a lambda function. Still, the conversion only happens after calling the `grad()` method. We don't want to do this in the constructor, because the gradient may not be needed. Maybe the conversion should be triggered on pickling as well?

All in all, I'm not sure if that's a good move or not. I think it's fine if `self.grad_op` is either `None`, a list, a callable, or an `OpFromGraph` (the first two before `grad()`, the latter two after `grad()`).

khaotik: > But it can only assume so if the grad method has been called. Otherwise it's still what was given to the constructor.

Good point. I think the future implementation should use a specialized getter method like `get_grad_op()` or `getattr` hacks, which performs check on `is_cached`. For now, maybe just rename it to `_grad_op` for privacy.

> Something fancy that gives a different gradient depending on the expressions it gets?

I think this should be the responsibility of other Ops like `IfElse`. With current design, user can use another OfG inst encapsule `IfElse` for grad overrides. After `updates` for OfG is implemented (it's on their TODO list), we should be able to implement grad functions with side effects (states). This will make jobs like implementing RTRL approximation for RNN much easier.

nouiz:
travis error is: 

/home/travis/build/Theano/Theano/theano/compile/builders.py:docstring of theano.compile.builders.OpFromGraph:6: WARNING: Literal block ends without a blank line; unexpected unindent.

theano/tests/unittest_tools.py
```diff
@@ -244,9 +244,9 @@ def _compile_and_check(self, inputs, outputs, numeric_inputs, cls,
         # Check that the Op is removed from the compiled function.
         if check_topo:
             topo_shape = shapes_function.maker.fgraph.toposort()
+            topo_out = outputs_function.maker.fgraph.toposort()
             assert not any(isinstance(t.op, cls) for t in topo_shape)
-        topo_out = outputs_function.maker.fgraph.toposort()
-        assert any(isinstance(t.op, cls) for t in topo_out)
```
nouiz: Why this change? I don't think it is the right fix, we want the old behavior most of the time. If you really need that, add a new parameter.

theano/compile/builders.py
```diff
@@ -168,30 +383,48 @@ def infer_shape(self, node, shapes):
...
+    gof.opt.in2out(inline_ofg_expansion),
+    0.5, 'fast_compile', 'fast_run')
+
+# Since OpFromGraph contains a Theano compiled function,
+# we should let DebugMode know about it
 ops_with_inner_function[OpFromGraph] = 'fn'
+
+
+# convenience function
+op_from_graph = OpFromGraph
```
nouiz: I'm not sure about that. Why do you find this useful?

I would think that having a decorator op_from_graph over a function would be a good user interface. What do you think of this? If we aren't sure what to do with that name, it would be more future proof to not add this now.

theano/compile/builders.py
```diff
@@ -168,30 +383,48 @@ def infer_shape(self, node, shapes):
...
+        return False
+    return theano.clone(
+        op.local_outputs, {
+            u: v for u, v in izip(
+                node.op.local_inputs, node.inputs)})
+
+optdb.register(
+    'inline_ofg_expansion',
+    gof.opt.in2out(inline_ofg_expansion),
+    0.5, 'fast_compile', 'fast_run')
```
nouiz: Use -0.01 instead of 0.5 and add this comment to explain a little bit the choise:

    We want to run this before the first merge optimizer and before the first scan optimizer.

theano/compile/builders.py
```diff
@@ -1,44 +1,79 @@
...
-        - opt to unfold it, work inplace on inputs
         - grad() make it support DisconnectedType and the new interface
+        - implement R_op()
         - check how it works with updates.
         - add test with constant as input or inside the inner graph.
         - Add support for the GPU? Probably just need an opt to remove transfer
         - Add support to pickle this Op.
         - Add support/test with random generator
+        - Recursion detection to prevent Op "forkbomb", either set depth
+          limit or manually check them.
```
nouiz: I don't get what the problem is. If the user make a 1000k depth op from graph, then this is the same problems as too big graph. What I'm missing?

theano/compile/builders.py
```diff
@@ -1,44 +1,79 @@
...
+        list of (None|undef|function), optional
+        Used to override default gradient routine.
+        Overriding function(s) must take two list of variable(s) as inputs,
+        the original inputs and ups gradients
+        For different `grad_overrides`:
+
+        - `None` : will use default gradient routine.
+        - theano.utils.undef : No gradient will be used (zero)
+        - OpFromGraph instance: the OfG instance should accept inputs with same
+            order and types as specified in "inputs" and "outputs" arguments
```
nouiz: outputs -> outputs grad

to be more clear.

theano/compile/builders.py
```diff
@@ -61,51 +96,74 @@ class OpFromGraph(gof.Op):
...
+            [x, y, z], [e], grad_overrides=[None, rescale_dy, None])
+        e2 = op(x, y, z)
+        dx, dy, dz = grad(e2, [x, y, z])
+        fn = function([x, y, z], [dx, dy, dz])
+        # the graident wrt y is now doubled
+        fn(2., 3., 4.) # [1., 8., 3.]
+
+    """
+    # NOTE: if you make a subclass of this, make sure add test for it under:
+    # theano/compile/tests/test_builders.py
```
nouiz: why? If the subclass is in another file, why we should do that?

theano/compile/builders.py
```diff
@@ -1,44 +1,79 @@
...
+        - theano.utils.undef : No gradient will be used (zero)
+        - OpFromGraph instance: the OfG instance should accept inputs with same
+            order and types as specified in "inputs" and "outputs" arguments
+        - function : must return list of Variable.
+        - list : each function must return a single Variable. The order
+            of the list must corresponds to inputs
+    rop_overrides: None | undef | OpFromGraph instance | function | \
+        list of (None|undef|function), optional
+        similar to grad_overrides, list order should match two list of "inputs"
+        concatenated.
```
nouiz: Add a kwargs* description here.

theano/compile/builders.py
```diff
@@ -115,40 +173,197 @@ def __hash__(self):
...
+                    all_grads_l.append(next(gdefaults))
+                elif gov is undef:
+                    all_grads_l.append(
+                        inp.zeros_like().astype(theano.config.floatX))
+                else:
+                    all_grads_l.append(gov(self.local_inputs, output_grads))
+        elif self._grad_op is undef:
+            all_grads_l = [
+                inp.zeros_like().astype(theano.config.floatX)
+                for inp in self.local_inputs]
```
nouiz: I'm not sure about that. @lamblin do you think this is good?

theano/compile/builders.py
```diff
@@ -115,40 +173,197 @@ def __hash__(self):
...
+                wrt=wrt_l,
+                disconnected_inputs='ignore') if wrt_l else [])
+            # combine overriding gradients
+            all_grads_l = []
+            for inp, gov in izip(self.local_inputs, goverrides_l):
+                if gov is None:
+                    all_grads_l.append(next(gdefaults))
+                elif gov is undef:
+                    all_grads_l.append(
+                        inp.zeros_like().astype(theano.config.floatX))
```
nouiz: @lamblin is this also good?

theano/compile/builders.py
```diff
@@ -1,44 +1,79 @@
...
+          local_outputs)
         - c_code() to remove the double overhead?
-        - opt to unfold it, work inplace on inputs
         - grad() make it support DisconnectedType and the new interface
+        - implement R_op()
         - check how it works with updates.
         - add test with constant as input or inside the inner graph.
         - Add support for the GPU? Probably just need an opt to remove transfer
         - Add support to pickle this Op.
         - Add support/test with random generator
```
nouiz: Add in the todo, to add an optimization that remove inputs not used by the outputs and another optimization that remove the outputs that don't have clients.

theano/compile/builders.py
```diff
@@ -115,40 +173,197 @@ def __hash__(self):
...
+                    self._grad_op
+                )
+        self._grad_op = type(self)(
+            inputs=self.local_inputs + output_grads,
+            outputs=all_grads_l,
+            inline=self.is_inline, on_unused_input='ignore',
+        )
+        self._grad_op_is_cached = True
+
+    def _recompute_rop_op(self):
```
nouiz: @lamblin let you review this part.

theano/compile/builders.py
```diff
@@ -115,40 +173,197 @@ def __hash__(self):
...
+        This will completely remove any previously set gradient overrides
+
+        """
+        self._grad_op = grad_overrides
+        self._grad_op_is_cached = False
+
+    def R_op(self, inputs, eval_points):
+        if not self._rop_op_is_cached:
+            self._recompute_rop_op()
+        return self._rop_op(*(list(inputs) + list(eval_points)), return_list=True)
```
nouiz: If the user pass a function for the override, then this function will get a return_list=True parameter.

Add in the doc this information and add a test for that cases.

theano/compile/builders.py
khaotik: I think there's already a decorator interface: `as_op`, which is a wrapper of `FromFunctionOp`, both in `theano/compile/ops.py`. We could add some new parameters to `as_op`.

The name does appears somewhat debatable, I will remove this line for now.

theano/compile/builders.py
khaotik: To generalize, I think a lot optimizations could apply to the inner graph *prior* to inline expansion.

theano/compile/builders.py
khaotik: `_recompute_rop_op` will convert whatever override user supplied to an OfG instance, so no need to worry that.

theano/tests/unittest_tools.py
khaotik: I don't absorb this code very well. In the docstring of that function, it says `check_topo` flags whether to check the Op was removed from the graph. So assumed those 2 lines *should* be under the if clause. Without doing so it would cause test error since OfG node was removed due to inline. Also this doesn't change default behavior since `check_topo=True` for default. Am I doing anything wrong?

theano/compile/builders.py
```diff
@@ -55,14 +61,13 @@ class OpFromGraph(gof.Op):
...
         - grad() make it support DisconnectedType and the new interface
-        - implement R_op()
         - check how it works with updates.
         - add test with constant as input or inside the inner graph.
         - Add support for the GPU? Probably just need an opt to remove transfer
         - Add support to pickle this Op.
         - Add support/test with random generator
-        - Recursion detection to prevent Op "forkbomb", either set depth
-          limit or manually check them.
+        - Add optimization prior to inilne expansion such as removing unused
```
nouiz: inilne -> inline

But I'm not sure I understand everything. What do you mean by "inline expansion"? I understand that we still miss opt that will remove useless inputs/outputs.

nouiz:
travis have this error:

/home/travis/build/Theano/Theano/theano/compile/builders.py:docstring of theano.compile.builders.OpFromGraph:36: WARNING: Inline strong start-string without end-string.

https://travis-ci.org/Theano/Theano/jobs/187985652

Otherwise, it is good for me except that we still need someone to review the Rop implementation and test. I'll check with @lamblin this week who could do that.

theano/compile/builders.py
khaotik: "inline expansion" means the opt to unpack the inner graph and merge it with the outer graph. I borrowed the `inline` keyword from C++. I'm not sure but I assume theano can already prune redundant nodes in the outer graph? If so, the removal only need to be done prior to inline expansion for some speedup or the non-inline version

nouiz:
Doing the removing of useless node before the inline won't speed up
compilation. The reason to have optimization to remove inputs/outputs not
used is when we don't inline. If we don't do that, it could cause
computation of nodes that won't be used.

theano/compile/builders.py
khaotik: Alternatively, this can be `int(0) -> zeros_like()` and `None -> None`, this preserves literal meaning of `None`. Just a interface matter, it's your call @nouiz.

lamblin:
Sorry for the delay, I'll start a review now.
I'm not convinced yet by the idea of passing a lambda function to the constructor of the base class, rather than inheriting and declaring `def grad(...)`, `def L_op(...)`, or `def R_op(...)`, but we'll see.

theano/tests/unittest_tools.py
lamblin: What the docstring meant, even though it is not clear, is that if `check_topo` is True, it will check that the original Op is not (any more) in the graph that computes its shape only.
However, when we want to compare the result of the `infer_shape` optimizations with the original Op, we _need_ the original Op in the graph to make sure we are comparing the right thing.

The way to resolve this here, in my opinion, would be to revert that change, and instead call `self._compile_and_check(..., excluding=inline_ofg_expansion)`, so the original OpFromGraph can be executed.

theano/compile/builders.py
```diff
@@ -1,44 +1,114 @@
...
+
+    .. TODO:
         - examples for a multi-layer mlp. where?
         - __hash__, __eq__ otherwise won't merge, try
-          gof.opt.is_same_graph_with_merge(op1.new_outputs, op2,
-          new_outputs)
+          gof.opt.is_same_graph_with_merge(op1.local_outputs, op2,
+          local_outputs)
         - c_code() to remove the double overhead?
-        - opt to unfold it, work inplace on inputs
```
lamblin: I think you should keep "opt to work inplace on inputs" in the list

theano/compile/builders.py
```diff
@@ -115,40 +210,203 @@ def __hash__(self):
...
+                    len(self.local_inputs) - len(goverrides_l))
+            wrt_l = [lin for lin, gov in
+                     izip(self.local_inputs, goverrides_l) if not gov]
+            # compute non-overriding downsteam grads from upstreams grads
+            # it's normal some input may be disconnected, thus the 'ignore'
+            gdefaults = iter(theano.gradient.grad(
+                cost=None,
+                known_grads=dict(izip(self.local_outputs, output_grads)),
+                wrt=wrt_l,
+                disconnected_inputs='ignore') if wrt_l else [])
```
khaotik: In fact, it's just one call. Equivalent code:
```python
if wrt_l:
    _gdefaults = theano.gradient.grad(...)
else:
    _gdefaults = []
gdefaults = iter(_gdefaults)
```

theano/compile/builders.py
khaotik: So, I propose the interface `None` -> `NullType()`, `int(0) -> DisconnectedType()`, is this okay?

theano/compile/builders.py
lamblin: Oh, OK, I thought `wrt_l` was iterated through.
It may be worth setting `return_disconnected='None'` and `null_gradients='return'` however.

khaotik:
I'm not sure about travis error and I cannot reproduce it on my machine.

nouiz:
Thanks for the update. I'm looking now at the change to this PR.

We want to make an RC Monday next week and we would like to have this PR include. So we will review/work on this PR with high priority this week.

For travis, I'm not sure what is going on, but I'll look at it this morning.

theano/compile/builders.py
nouiz: I didn't replied to this comment. Yes, pre-optimizing the graph before inline could be useful. There is many thread-off point that are sensible to use. This is for later/another PR. I'm not sure what should be the default thread-off point. Here is the one I see. Optimize the inner graph with the optimization phase up to and including:
- useless optimization (remove useless node to make it smaller)
- useless + canonicalize (convert different equivalent graph to a canonical version to help following optimization)
- useless + canonicalize + stabilize opt
- above + specialize (speed opt mostly)

Doing more opt before inline could speed up optimization, in particilar if one OpFromGraph op is reused at many places in the graph. But this could miss some optimization.

Do you have an opinion on what we could do?

theano/compile/builders.py
```diff
@@ -38,11 +40,12 @@ class OpFromGraph(gof.Op):
...
 
-    grad_overrides : single or list of {None, undef, OpFromGraph, callable}, optional
+    grad_overrides : single or list of {0, None, Ellipsis, OpFromGraph, callable}, optional
         Defaults to ``None``.
 
-        ``None`` : Do not override gradient
-        theano.utils.undef : No gradient will be used (zero)
+        ``None`` : No value, gives NullType()
+        ``0`` : zero value, gives DisconnectedType()
+        ``...`` : Do not override, use default grad() result
```
nouiz: @lamblin What do you think of this? I'm not convinced. I would think that using None for do not override made more sence. None mean do thing in subtensor semantic. But None is also close in writing to Null.

Why not ask the user to give NullType and DisconnectedType instance directly to be more explicit? We removed None in grad everywhere else I think to be more explicit. Also, using 0 for Disconected could in theory mix cases where we are able to infer the value of grad to be always 0 during theano.grad() and not disconnected. We will probably never do that, but I'm pretty sure in some corner case the graph would allow us to do so.

theano/compile/builders.py
```diff
@@ -144,25 +150,59 @@ class OpFromGraph(gof.Op):
...
+
+    def __init__(
+            self, inputs, outputs,
+            inline=False,
+            grad_overrides=Ellipsis, rop_overrides=Ellipsis,
+            name=None, **kwargs
+    ):
         if not isinstance(outputs, list):
-            raise TypeError('outputs must be list', outputs)
+            raise TypeError('outputs must be list, got %s' % outputs, outputs)
```
nouiz: remove ", outputs". This duplicate the information in the error message. Your change make it more clear which is good.

theano/compile/builders.py
```diff
@@ -216,64 +256,94 @@ def __str__(self):
...
+            all_grads_l = theano.gradient.grad(
+                cost=None,
+                known_grads=OrderedDict(izip(local_outputs, output_grads)),
+                wrt=local_inputs,
+                disconnected_inputs='ignore')
+            all_grads_ov_l = [None] * inp_len
+        elif grad_op is None:
+            all_grads_l = [inp.zeros_like() for inp in local_inputs]
+            all_grads_ov_l = [self.ofg_null_t()] * inp_len
+        elif grad_op is 0:
```
nouiz: I'm not sure that 0 is guaranted to be a singleton by all Python distribution. Unless you know why it is competly safe, I would prefer this safer version `type(grad_op) is int and grad_op==0`

theano/compile/builders.py
```diff
@@ -216,64 +256,94 @@ def __str__(self):
...
             # compute non-overriding downsteam grads from upstreams grads
             # it's normal some input may be disconnected, thus the 'ignore'
+            wrt_l = [lin for lin, gov in izip(
+                self.local_inputs, goverrides_l) if gov is Ellipsis]
             gdefaults = iter(theano.gradient.grad(
                 cost=None,
-                known_grads=dict(izip(self.local_outputs, output_grads)),
+                known_grads=OrderedDict(izip(self.local_outputs, output_grads)),
                 wrt=wrt_l,
                 disconnected_inputs='ignore') if wrt_l else [])
```
nouiz: This this work if we have a list with all/none Ellipsis? Can you make sure this corner case is tested?

theano/compile/builders.py
```diff
@@ -263,48 +274,50 @@ def _recompute_grad_op(self):
...
+            self._grad_op_overrides_l = [None] * inp_len
             return
 
         output_grads = [out_t() for out_t in self.output_types]
+        fn_grad = partial(
+            theano.gradient.grad,
+            cost=None,
+            disconnected_inputs='ignore',
+            return_disconnected='Disconnected',
+            null_gradients='return',
```
nouiz: @lamblin I think the usage of return_disconnected and null_gradients here is well done, but a double check would be great. Scan.grad use null_gradients='return and return_disconnected='None'

theano/gradient.py
```diff
@@ -1201,58 +1201,62 @@ def try_to_copy_if_needed(var):
...
+                                "verifiably zeros."
+
+                            msg %= (node.op, term, type(term), i)
+
+                        elif is_zero == 'no':
+                            msg = "%s.grad returned %s of type %s for input" \
+                                " %d. Since this input is only connected " \
+                                "to integer-valued outputs, it should " \
+                                "evaluate to zeros, but it evaluates to" \
+                                "%s."
```
nouiz: Normally, we don't use `\` anymore in the code. If you can revert this, it would be great.

theano/gradient.py
```diff
@@ -1201,58 +1201,62 @@ def try_to_copy_if_needed(var):
...
+            for i, (ipt, ig, connected) in enumerate(
+                zip(inputs, input_grads, inputs_connected)):
                 actually_connected = \
                     not isinstance(ig.type, DisconnectedType)
 
+                if isinstance(node.op, theano.OpFromGraph):
+                    ov = node.op._grad_op_overrides_l[i]
+                    if ov is not None:
+                        connected &= not isinstance(
+                            ov.type, DisconnectedType)
```
nouiz: I don't like this solution. Having Op specific code in gradient.py isn't great. I'm not sure what is the right solution, but using `isinstance(node.op, tuple(theano.compile.builders.ops_with_inner_function)) and hasattr(node.op, '...)` would definitly be better.

But is this a sympom that the method connection_pattern in OpFromGraph isn't implemented or is missing cases?

nouiz:
I have no idea about the travis test. I'm not able to reproduce it eighter.

I'll investigate more.

nouiz:
I pushed a new commit to your branch that add extra information about the travis error. I'm also not able to reproduce it here.

theano/gradient.py
khaotik: The problem is the code here tries to use `connection_pattern` for the Op *before* `grad` to check for gradient bugs. However when gradient is user overridden, the check no longer makes sense.

nouiz:
I have found the problem. It is not in this PR, but in the failing test. Change in this PR trigger this latent problem. I'll do a commit to this PR this morning with the fix.

nouiz:
I pushed the fix.

theano/compile/builders.py
nouiz: done

theano/compile/builders.py
```diff
@@ -47,7 +117,7 @@ class OpFromGraph(gof.Op):
 
     .. code-block:: python
 
-        from theano import function, OpFromGraph, tensor
+        from theano import function, tensor
```
nouiz: fixed

theano/compile/builders.py
```diff
@@ -70,42 +140,67 @@ class OpFromGraph(gof.Op):
         e2 = op(x, y, z) + op(z, y, x)
         fn = function([x, y, z], [e2])
 
-    """
+    Example 3 override gradient
+
+    .. code-block:: python
+
+        from thenao import funciton, OpFromGraph, tensor, grad
```
nouiz: done

theano/compile/builders.py
```diff
@@ -1,44 +1,121 @@
...
+        list: Each OpFromGraph/callable must return a single
+        :class:`Variable <theano.gof.Variable>`. Each list element corresponds to gradient of
+        a specific input, length of list must be equal to number of inputs.
+
+    rop_overrides : single or list of {0, None, Ellipsis, OpFromGraph, callable}, optional
+        Defaults to ``None``.
+
+        ``None`` : No value, gives NullType()
+        ``0`` : zero value, gives zeros_like(...)
+        ``...`` : Do not override, use default R_op() result
```
lamblin: This is a tough interface question, and I'm not sure how to solve it, but here's my take.
- I would like to avoid the use of `None`, since we used to have it mean an ambiguous "not defined or not connected or zero" in the past for `grad`, and it is still the case for `R_Op`.
- Following "explicit is better than implicit" from the _Zen of Python_, I would say the way of specifying `NullType()` or `DisconnectedType()` should be to pass it (maybe accept either the class or an instance, so user can define their own custom message, but do not have to). The option to return `DisconnectedType()` seems missing in the current version, btw.
- The use of `Ellipsis` is clever, and I like its compactness. I'm afraid that it may be _too_ clever, or not clear enough, or that examples look like incomplete code. However, the only alternative I have for that would be `False`, as in "do not override".

Let's discuss that and agree on a solution before implementing it, as it is going to be hard to change after the PR is merged.

theano/compile/builders.py
lamblin: lgtm

theano/compile/builders.py
```diff
@@ -115,40 +261,273 @@ def __hash__(self):
...
+        return apply_node
 
     def connection_pattern(self, node):
         """
         Return connection pattern of subfgraph defined by inputs and outputs.
 
         """
-        return io_connection_pattern(self.new_inputs, self.new_outputs)
+        return io_connection_pattern(
+            self.local_inputs, self.local_outputs)
```
lamblin: I think we should actually take the gradient overrides into account here.
It would remove the special case in theano/gradient.py, and avoid giving misleading information.

theano/compile/builders.py
khaotik: If the interface is debatable, then maybe just use explicit `NullType()` or `DisconnectedType()` for  the class constructor for now. The interface may be implemented in a future convenience function.

theano/gradient.py
lamblin: I think the `connection_pattern` should really be taken as "should gradients flow through there?", in particular "if there's an invalid output gradient, should the input gradient be invalid too?" and "if this is the only path between the cost and the param, is the user likely to have confused some variables?".
If the user explicitly overrode the gradient to be an actual expression, I say the answer to those questions would be yes, and the connection pattern should be `True`.
If the user explicitly overrode the gradient to be `DisconnectedType()`, I say the answer is no, and the connection pattern should be `False`.
If it is not overriden, then use `io_connection_pattern`.

That way, I think the current check still makes sense.

lamblin:
Jenkins test this please

theano/compile/builders.py
lamblin: OK for explicit `NullType()` and `DisconnectedType()` for those cases.
What do you think for the "no override" case? We could also just use the string `"no_override"` or something.

khaotik: For now, I think string `"default"` would suffice. In future, I think there should be global singletons like `theano.Default`, `theano.Automatic` (I lend the idea from Mathematica), and then make all theano interfaces uniform. Just a personal idea.

nouiz:
@khaotik do you have news on that? When do you think you can implement the changes?

We would like to release the RC Monday.

thanks

lamblin:
Jenkins test this please

theano/compile/builders.py
lamblin: OK.

lamblin:
Jenkins test this please

nouiz:
thanks! I think better doc could be done to make this more visible. But is
it the right time? It still don't support well the GPU, so if people don't
use inline, they can expect un efficiency at many places.

Issue #1168, opened 2013-01-17
Scan over head on GPU is 140x bigger then on CPU


nouiz:
See https://groups.google.com/forum/?fromgroups=#!topic/theano-users/i9sbLyTbBPw

There is a simple case where on the CPU the scan overhead is : 1.5e-02s, on the GPU:
2.2s.

TODO in the order what what I expect to bring the most speed up.
- [x] make GpuGemv call the new cublas dot interface
- [ ] check where scan spend 0.65s as overhead.
- [ ] Make the elemwise code faster on the CPU for sqr. Look at new code in NumPy?
- [ ] Make an [ElemwiseReduce Op](https://github.com/Theano/Theano/issues/1167)

```
import theano
import theano.tensor as tt
import numpy as np


x = tt.matrix()

def test1(x):
    r, _ = theano.scan(fn=lambda x : tt.dot(x,x), sequences=x)
    return r

def test2(x):
    return (x ** 2).sum(1)

f1 = theano.function([x], test1(x))
f2 = theano.function([x], test2(x))

xval = np.random.randn(10000,10000).astype(np.float32)

import time
t=time.time()
for i in range(3):
    f1(xval)
print time.time() - t
t=time.time()
for i in range(3):
    f2(xval)
print time.time() - t

```

wielandbrendel:
I think I experience the same issue when trying to roll the axis of a tensor independently. Here is a boiled-down toy example:

    import theano
    import theano.tensor as T

    tA = T.fmatrix('A')
    roll_direct = theano.function([tA], T.roll(tA, 10))
    roll_scan   = theano.function([tA], theano.map(lambda a: T.roll(a, 10), tA)[0])

On the GPU (K40), with A being of dimension (50.000, 1.024), roll_direct takes 100ms whereas roll_scan takes > 5s. At the same time, the CPU needs 68ms for the direct method (less then GPU because of memory transfer overhead), and 300ms for the scan version.

nouiz:
We are actively working on scan. Be sure to use the development version of
Theano, not 0.7. It have many scan updates.

Also, if you do not need scan, do not use it. Don't use a sledge hammer
when you don't need it.

c0d3rman:
+1 - How is progress on this? Really slowing down Keras RNN compile time.

nouiz:
This issue is about execution time, not compile time.

There have be very good execution and compilation time speed up in the
development version of Theano. Do you use the latest Theano release 0.7? Is
[...]

PR #5257, opened 2016-11-20
GPU gemv speedup

khaotik:
Added call to `cublasSdot` in GPU gemv code.

This is proposed solution for the 1st issue in #1168. Despite not solving all the problem, I'm getting 4x speedup on my GPU.

### DIFF
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -4468,7 +4468,7 @@ int CudaNdarray_sgemv(float alpha, const CudaNdarray * A, const CudaNdarray * B,
 
     if (CUBLAS_STATUS_SUCCESS != err)
     {
-        if (used_dot)
+        if (!used_dot)
         {
             PyErr_Format(PyExc_RuntimeError,
                          "cublasSgemv failed (%i)",
--- a/theano/sandbox/cuda/cuda_ndarray.cu
+++ b/theano/sandbox/cuda/cuda_ndarray.cu
@@ -4396,36 +4396,25 @@ int CudaNdarray_sgemv(float alpha, const CudaNdarray * A, const CudaNdarray * B,
     if (sa_0 == 0) sa_0 = 1;
     if (sa_1 == 0) sa_1 = 1;
 
+    int used_dot = 0;
+
     // This is important because we can end up not calling Sgemv at all
     cublasStatus_t err = CUBLAS_STATUS_SUCCESS;
     if (CudaNdarray_SIZE(C)) {
         // A is row vector & alpha==1 & beta==0 -> use cublasSdot
         if (CudaNdarray_HOST_DIMS(A)[0] == 1 && alpha==1.f && beta==0.f) {
-            //TODO: this is a rather temporary solution
-            // 1. better temp solution:
-            // replace this with custom inner product kernel with
-            // alpha and beta parameter
-            // 2. permanant solution:
-            // define a new "InnerProduct" Op, add an optimization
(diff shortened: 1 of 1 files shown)


theano/sandbox/cuda/cuda_ndarray.cu
```diff
@@ -4393,15 +4393,42 @@ int CudaNdarray_sgemv(float alpha, const CudaNdarray * A, const CudaNdarray * B,
...
-        if ((CudaNdarray_HOST_DIMS(A)[0] <= 1)
+        // A is row vector & alpha==1 & beta==0 -> use cublasSdot
+        if (CudaNdarray_HOST_DIMS(A)[0] == 1 && alpha==1.f && beta==0.f) {
+            //TODO: this is a rather temporary solution
+            // 1. better temp solution:
+            // replace this with custom inner product kernel with
+            // alpha and beta parameter
+            // 2. permanant solution:
+            // define a new "InnerProduct" Op, add an optimization
+            // "gemv -> inner_prod", perhaps for CPU/GPU both
```
nouiz: The new op trick won't always work. Sometimes we won't know the shapes, so having it here is good I think and is easier. So just remove this comment.

theano/sandbox/cuda/cuda_ndarray.cu
```diff
@@ -4393,15 +4393,42 @@ int CudaNdarray_sgemv(float alpha, const CudaNdarray * A, const CudaNdarray * B,
...
+            float* dev_dst = CudaNdarray_DEV_DATA(C)+1-sc_0;
+            cublasPointerMode_t pmode;
+            cublasGetPointerMode(handle, &pmode);
+            // need to store dot result on device here
+            cublasSetPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
```
nouiz: Don't set the pointermode. We don't do this anywhere, we always pass them on the CPU (the default). Also, you are passing them on the host bellow. So I don't understand why you added this.

theano/sandbox/cuda/cuda_ndarray.cu
```diff
@@ -4393,15 +4393,42 @@ int CudaNdarray_sgemv(float alpha, const CudaNdarray * A, const CudaNdarray * B,
...
+        // A is row vector & alpha==1 & beta==0 -> use cublasSdot
+        if (CudaNdarray_HOST_DIMS(A)[0] == 1 && alpha==1.f && beta==0.f) {
+            //TODO: this is a rather temporary solution
+            // 1. better temp solution:
+            // replace this with custom inner product kernel with
+            // alpha and beta parameter
+            // 2. permanant solution:
+            // define a new "InnerProduct" Op, add an optimization
+            // "gemv -> inner_prod", perhaps for CPU/GPU both
+            float* dev_dst = CudaNdarray_DEV_DATA(C)+1-sc_0;
```
nouiz: Why the "+1-sc_0"? I would completly remove that.

theano/sandbox/cuda/cuda_ndarray.cu
khaotik: If I don't add this, this will cause CUBLAS take dot result as a host pointer, causing segfault ([Example](http://stackoverflow.com/questions/28979616/segmentation-fault-while-executing-cuda-program)). I set it back to make sure other CUDA code still function correctly.

theano/sandbox/cuda/cuda_ndarray.cu
khaotik: I was misunderstanding cublas doc, removed.

nouiz:
jenkins test this

PR #5262, opened 2016-11-21
Gpu argmax

aam-at:
Following #5168, I noticed that Cpu Argmax was not lifted to gpu. This pull request adds local optimization to lift Argmax to GpuMaxAndArgmax.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -1186,7 +1186,7 @@ class MaxAndArgmax(Op):
     __props__ = ('axis',)
 
     def __init__(self, axis):
-        assert isinstance(axis, list)
+        assert isinstance(axis, (list, tuple))
         self.axis = tuple(axis)
 
     def get_params(self, node):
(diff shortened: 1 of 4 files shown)


nouiz:
GpuMaxAndArgmax sometimes is slower the the CPU version even including the transfer. So I don't think we should merge this PR now. We wanted to fix that first before using it more frequently.

I'll let @abergeron confirm that I didn't forget something. We also wanted to modify the existing code to not write the max when we only use the argmax to save memory.

abergeron:
I think we should merge the portion that simplifies axis handling in Argmax.  As for the lifting, it should use op_lifter and register properly for GraphToGpu.

aam-at:
I added local optimizer using op_lifter.
I also observed that gpu argmax is slower than cpu argmax. And it seems to be slower and faster for certain axes patterns.

nouiz:
the slow down is why we didn't added the lifter. So I think we should wait before the speed issue is fixed before merging this PR.

thanks

theano/tensor/basic.py
```diff
@@ -1394,11 +1394,14 @@ class Argmax(Op):
...
+    params_type = Generic()
+    __props__ = ('axis',)
 
-    def make_node(self, x, axis=None):
-        x = _as_tensor_variable(x)
+    def __init__(self, axis):
+        assert isinstance(axis, (list, tuple))
+        self.axis = tuple(axis)
 
+    def get_params(self, node):
```
nouiz: You need to update this method to be like MaxAndArgmax

theano/tensor/basic.py
```diff
@@ -1394,11 +1394,14 @@ class Argmax(Op):
     nin = 2  # tensor, axis
     nout = 1
     E_axis = 'invalid axis'
-    __props__ = ()
+    params_type = Generic()
```
nouiz: This interface changes will request some code to allow unpickling old Argmax. The first step is to make a graph with the old node. Then pickle that graph and make a test that unpickle it, compile a Theano function and execute it. Then we need to make that test pass.

PR #5265, opened 2016-11-21
Fix reload pickle

nouiz:
We need to find a better fix then the last commit.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -3996,6 +3996,7 @@ class Composite(ScalarOp):
     def __getstate__(self):
         rval = dict(self.__dict__)
         rval.pop('_impls', None)
+        rval.pop('prepare_node_called', None)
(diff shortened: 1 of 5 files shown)


abergeron:
I'm ok with this, the only comment is a small inefficiency, but it won't affect anything important.

theano/scalar/basic.py
```diff
@@ -3996,13 +3996,16 @@ def __hash__(self):
...
+        rval.pop('prepare_node_called', None)
         del rval['fgraph']
         return rval
 
     def __setstate__(self, d):
         self.__dict__.update(d)
         # We must call init to set fgraph and _impls again, as otherwise
         # self.perform will not work.
+        if not hasattr(self, "prepare_node_called"):
+            self.prepare_node_called = set()
```
nouiz: I updated that.

PR #5266, opened 2016-11-21
Add miniconda.sh and installation to TRAVIS cache.

notoraptor:
Add miniconda.sh and installation to TRAVIS cache.

It should works know. I have tested it on my fork,
some TRAVIS jobs can be checked here:
https://travis-ci.org/notoraptor/Theano/builds/177772594

The job for test_flake8 is already cached on my fork:
https://travis-ci.org/notoraptor/Theano/jobs/177772595
( raw log: https://api.travis-ci.org/jobs/177772595/log.txt?deansi=true )

@nouiz !


.travis/travis_before_install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Install miniconda to avoid compiling scipy
+if test -e $HOME/miniconda2/bin ; then
+    echo "miniconda already installed."
+else
+    echo "Installing miniconda."
+    rm -rf $HOME/miniconda2
+    wget -c https://repo.continuum.io/miniconda/Miniconda2-4.1.11-Linux-x86_64.sh -O $HOME/download/miniconda.sh
```
nouiz: what happen if the directory download don't exist? If the cache get deleted, I think it won't be there, so I would be safer to create it with "mkdir -p $HOME/download", just in case.

.travis/travis_install.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# In Python 3.3, we test the min version of NumPy and SciPy. In Python 2.7, we test more recent version.
+if test -e $HOME/miniconda2/envs/pyenv ; then
+    echo "pyenv already exists."
+    source activate pyenv
+    if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install mkl python=2.7 numpy=1.9.1 scipy=0.14.0 nose=1.3.0 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx mkl-service libgfortran=1; fi
```
nouiz: If it already exist, why reinstall it? I don't like the idea of having the version twice.
What about this different organisation.

If pyenv don't exist, create.

Then always install? So the package/version list will only exist once.

.travis.yml
```diff
@@ -3,26 +3,28 @@
...
 before_install:
-# Install miniconda to avoid compiling scipy
-  - mkdir -p download
-  - cd download
-  - wget -c https://repo.continuum.io/miniconda/Miniconda2-4.1.11-Linux-x86_64.sh -O miniconda.sh
-  - chmod +x miniconda.sh
-  - ./miniconda.sh -b
-  - cd ..
+  - chmod +x ./.travis/travis_before_install.sh
+  - ./.travis/travis_before_install.sh
```
lamblin: I think you can mark the file as executable from git itself, and you would not need to chmod it each time.

notoraptor: Ok ! I test it now.

notoraptor:
It seems it works now. I have applied the recommendation of @lamblin to simplify the code, and now all tests succeed.

One can check the installation cache persistence for the 9 first jobs (not for the others because I have killed the previous TRAVIS build when uploading the last commit).

Do I make a dummy commit to confirm all ?

@nouiz

Issue #3319, opened 2015-08-22
Graph contains cycles


dselsam:
Hi,

When compiling my model, I get `ValueError(graph contains cycles)` during the check in `general_toposort`. The traceback is long, but the beginning is: 

```
ERROR (theano.gof.opt): SeqOptimizer apply 
<theano.gof.opt.EquilibriumOptimizer object at 0x21e49990>
```

My model is a rather complicated recursive neural network, and unfortunately I cannot find a simple example that reproduces the error.

1. Is it correct that no model should be able to trigger this error, and thus that it is a bug in Theano itself?
2. If I sent my model and enough data to trigger the error, would a developer be willing to investigate?
3. If not, are there any potential workarounds, at the user level or involving extra checks and roll-backs during the graph optimization?

Thanks very much.

lamblin:
1. It is possible for user-defined code to explicitly that error, but it would require that the graph contains inconsistent information about which operations should be executed in place, in a destructive fashion. This is why explicitly specifying "inplace" patterns should be avoided.
2. If you provide us with the code and data, we would be interested in having a look and trying to solve the issue. However, we cannot commit on a timeline for that, as we are quite busy at the moment, and some of us are on vacation.
3. There are a couple of things you can try, that could help figuring out where the problem comes from:
  - can you compile a function with "optimizer=None" in the global flags (or passing `mode=theano.Mode(optimizer=None)` to `theano.function`)? If the error is still there, that is probably an issue in the original graph (about inconsistent inplace patterns)
  - can you disable the "inplace" optimizations? For instance `optimizer_excluding=inplace` in `THEANO_FLAGS`.
  - can you disable "ShapeOpt" the same way?
  - also, the full stack trace may be useful.

twiecki:
Related: https://github.com/pymc-devs/pymc3/issues/803

basveeling:
@dselsam I had the same issue, but after reinstalling scipy and numpy (both were already up to date) the error was gone.

fvisin:
I can confirm the same issue on one of the lab's computers, with a code that used to work and I didn't modify in the meanwhile.

fvisin:
I reverted back to https://github.com/Theano/Theano/commit/557fe6fab49262706241f0e774f3269a2ec0275b and it's working again. I wasn't able to determine exactly when it got broken though (time consuming). @lamblin: if you can provide me with the commits you think might trigger this problem, I can run more tests the next week.

twiecki:
[...]

PR #5267, opened 2016-11-21
Minor inconsistency in AbstractConv_gradInput implementations

gvtulder:
I don't know if this is a known problem, but I couldn't find an existing issue.

This [comment](https://github.com/fchollet/keras/blob/94fba3d8f05c65799e21198f845051f82b30416d/keras/layers/convolutional.py#L515-L527) in documentation for the Keras Deconvolution2D layer surprised me, because it suggests that the output shape of Theano `AbstractConv_gradInput` might depend on whether it runs on the GPU or the CPU. This is also discussed in fchollet/keras#3824.

The problem seems to be that if you set `border_mode='half'` and use an odd filter size, the result of `AbstractConv_gradInput` is not clearly defined. It works fine if you give it a gradient tensor that has the same size that you would get from the forward convolution, but if you give it a gradient tensor of a different size the results depend on the implementation. The CorrMM implementations give a different result than the cuDNN function.

These tests illustrate the problem. They fail because the cuDNN results do not match those of the reference implementation. Running them in DebugMode gives an optimization error.

I'm not sure how this problem should be solved, because changing the existing Theano implementations perhaps is not a good idea. However, it is also awkward that the results depend on the environment.


lamblin:
If this transposed convolution (or convolution gradient) does not correspond to any forward convolution, then my first reaction is that it should be invalid.
For instance, what if you take a gradient through it? Do you end up with a gradient of a different size, because the regular convolution does not support that output size?
Or at least we should try to see if there is a right answer.

nouiz:
I agree with Pascal, my first idea is that it would be invalid and we should raise an error in that cases.

Is there good use case where this would be valid to use different shape? If so, which one?
@vdumoulin, @f0k  what do you think?

nouiz:
Just a note, we try to make a release candidate, so fixing this rapidly would be ideal.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -137,6 +137,189 @@ def get_conv_shape_1axis(image_shape, kernel_shape, border_mode,
...
+        possibly depth) of the image. None where undefined.
+    border_mode: string, int (symbolic or numeric) or tuple of int (symbolic
+        or numeric). If it is a string, it must be 'valid', 'half' or 'full'.
+        If it is a tuple, its two (or three) elements respectively correspond
+        to the padding on height and width (and possibly depth) axis.
+    subsample: tuple of int (symbolic or numeric). Its two or three elements
+        respectively correspond to the subsampling on height and width (and
+        possibly depth) axis.
+    filter_dilation: tuple of int (symbolic or numeric). Its two elements
+        correspond respectively to the dilation on height and width axis.
```
f0k: Should be "two or three elements" as for the others.

f0k:
> @vdumoulin, @f0k what do you think?

Agreed, when the given transposed convolution output shape is too large or too small for the transposed convolution input (i.e., there is no possible forward pass for these settings), it should just bail out. It's interesting that neither GpuCorrMM nor cuDNN complain.

I've had a look at the proposed changes and they look good to me. It's a pain there's so much code duplication... we might want to merge the 2d and 3d cases some time.

gvtulder:
These patches change the CorrMM ops (and GPU equivalents) to make them raise a ValueError if you provide an image shape that is not possible given the output shape, filter shape and other parameters. The current behavior is to ignore the provided image shape unless there is subsampling, but now you will get an error if you specify an incorrect image shape.

I haven't looked at the cuDNN ops in sufficient detail yet. I have the impression that they always return the requested image size and ignore the extra gradient input. That should also raise a ValueError, I think, to have the same behavior as the CorrMM implementation.

Then there are the two legacy CPU ops, the ConvOp and ConvTransp3D. I haven't looked at ConvOp. ConvTransp3D computes the image size given the output size, but it uses this as a minimum size. You can ask for a larger image shape, which it will then fill with zeros. I'm not sure what to do with this. (Perhaps CorrMM and cuDNN are the most commonly used and those older ops are not that important.)

Finally, it would be nice if AbstractConv grad ops would derive the image or weight shape automatically. If there is no subsampling this should be straightforward. Would it be useful to define a default for the subsampling case as well? In those cases, the op could perhaps return the largest possible image if the image shape is not specified explicitly.

nouiz:
@lamblin told me that he also think adding the check is the good think.

I would forget ConvOp and ConvTransp3D. We should get rid of them at some point, so don't loose time with them.

Why you want AbstractConv grad ops derive the shape automatically? This would be a significant refactor I think and could break pickled graph too unless we take care to not break that. As we want to make an release candidate next week, I don't think it is a good idea.

What about providing functions to help compute the shapes without changing the graph organication? Maybe the user interface could call them when the shape isn't provided, I just don't think it is a good idea to change the graph structure.

nouiz:
Also @vdumoulin told me IRL he also think we should add the checks.

Note travis failed.

gvtulder:
Yes, Travis should fail. These are still the initial tests that only check if the shapes are the same. They don't test for the exceptions I've introduced. I have a new test for those, but I haven't added that yet.

gvtulder:
The `test_dnn_conv_merge` tests fail because, after adding the `Assert` shape check, the Asserts are not getting merged. This is because Assert-merging was disabled in PR #3344 (see also #3319).

theano/tensor/nnet/abstract_conv.py
```diff
@@ -137,6 +138,374 @@ def get_conv_shape_1axis(image_shape, kernel_shape, border_mode,
...
+    def check_dim(given, computed):
+        if given is None or computed is None:
+            return True
+        try:
+            given = get_scalar_constant_value(given)
+            computed = get_scalar_constant_value(computed)
+            return int(given) == int(computed)
+        except NotScalarConstantError:
+            # no answer possible, accept for now
+            return True
```
nouiz: There is some functions in Theano to compare the graph to know if they are the same. But it can only garanti that 2 graphs are equivalent. It can still mark some graph as not equivalent when they are. So I don't think they can be used here. I just wanted to let you know this exist in case you find them useful.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -137,6 +138,374 @@ def get_conv_shape_1axis(image_shape, kernel_shape, border_mode,
...
+    Returns a tuple similar to the given `shape`. For constant elements in `shape`,
+    the function checks the value and raises a `ValueError` if the dimension is invalid.
+    The elements that are not constant are wrapped in an `Assert` op that checks the
+    dimension at run time.
+    """
+    out_shape = []
+    for i, n in enumerate(shape):
+        try:
+            const_n = get_scalar_constant_value(n)
+            if i < 2:
```
nouiz: I think 2 should be 3 for 3d conv. Can you replace the 2 by len(shape) - 2?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -137,6 +138,374 @@ def get_conv_shape_1axis(image_shape, kernel_shape, border_mode,
...
+                if const_n < 0:
+                    raise ValueError('The convolution would produce an invalid shape (dim[%d]: %d < 0).' % (i, const_n))
+            else:
+                if const_n <= 0:
+                    raise ValueError('The convolution would produce an invalid shape (dim[%d]: %d <= 0).' % (i, const_n))
+            out_shape.append(n)
+        except NotScalarConstantError:
+            if i < 2:
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -782,7 +1151,7 @@ def conv3d_grad_wrt_weights(input,
                                                filter_flip=filter_flip,
                                                filter_dilation=filter_dilation)
 
-    return gradWeight_op(input, output_grad, filter_shape[:-3])
+    return gradWeight_op(input, output_grad, filter_shape[-3:])
```
nouiz: Can you confirm that now there is tests for this?

theano/tensor/nnet/abstract_conv.py
gvtulder: I think 2 is correct, actually. This is a test for the first two dimensions, for the batch size and number of channels. Those can be equal to zero, the convolution dimensions (i >= 2) not.

theano/tensor/nnet/abstract_conv.py
gvtulder: Yes, the [TestAssertShape](https://github.com/Theano/Theano/blob/7f1c36774f7c5163cf83790a44df6775244cf5da/theano/tensor/nnet/tests/test_abstract_conv.py#L319-L330) tests now call these `conv*_grad_wrt_*` functions. They don't do very extensive testing, but they do discover this typo.

nouiz:
Thanks!

Issue #5249, opened 2016-11-18
scan pushout output opt cause shape error


nouiz:
Reported by [[theano-users] Optimization error with scanOp_pushout_output](https://groups.google.com/d/topic/theano-users/5m5jhihDfCY/discussion)

Current work around: use Theano flag `optimizer_excluding=scanop_pushout_output `.

To reproduce, download the file in the second email of that thread and execute:

    python -c 'import theano;from six.moves import cPickle;d = cPickle.load(open("dumped_fn", "rb"))  ;f = theano.function(**d); a = cPickle.load(open("input_a.p", "rb")); b= cPickle.load(open("input_b.p", "rb"));c= cPickle.load(open("input_a.c", "rb")); f(a,b,c)'

abergeron:
I did some investigations, and the problems appears to stem from the fact that the inner dimension (K) of the dot is not the same at each iteration.  If the dot stays inside this is not a problem since the outer dims are the same.

I can't think of any sane way to not do a pushout for those cases but still do it for the rest, so my fix would be to disable the pushout of the output.

Here is a reduced test case that exposes the problem:
```
import numpy
import theano
import theano.tensor as T

def f(n, a):
    return T.dot(T.ones((100, n)), T.ones((n, 10))) + a

g, upd = theano.scan(f,
                     sequences=[theano.shared(numpy.asarray([20, 30, 40]))],
                     outputs_info=[T.zeros((100, 10))])

# Simulate gradient operation.                                                                                                                                                                          
g = g.owner.inputs[0][-1]

theano.printing.debugprint(g)
f = theano.function([], g)
theano.printing.debugprint(f)

f()
```

nouiz:
gh-5270 add a better error message that tell the user how to disable the troublesome optimization. This is the only quick fix we can do. I'll try to resume here IRL discussion with @abergeron and @lamblin about this.

We would need to detect in that optimization if the dimensions that get summer over is the same at each scan iteration.
- If we go the safe way and only do the push out when we can infer this, then we risk a speed regression when we aren't able to infer it correctly. But if we check that it cover current scan tests and some real world models, then that is probably fine.
- Detecting the reverse is much harder, so we won't do that.

Detecting if the dimensions we sum over is the same for each iteration can be done in those cases:
If that dimensions shape only depend on:
- the scan inputs, then it is constant for each iteration.
[...]

PR #5270, opened 2016-11-22
Add a better error for scan pushout failure

abergeron:
Also,  I did a flake8 pass for scan_op.py.

Better error message for gh-5249.


theano/scan_module/scan_op.py
```diff
@@ -1432,8 +1432,17 @@ def execute(self, node, args, outs):
...
+                            outs[j][0][pos[j]] = \
+                                output_storage[offset_out + j].storage[0]
+                        except ValueError as e:
+                            ne = ValueError(
+                                "An output of the scan has changed shape. "
+                                "This may be caused by a pushout optimization."
+                                " Try adding "
+                                "'optimizer_excluding=scanOp_pushout_output' "
+                                "to your Theano flags.")
+                            raise_from(ne, e)
```
nouiz: Did you manually test this? If so, I'm ok with this PR when jenkins and travis pass.

abergeron: I was missing the C version of the message, but now that's fixed.  And I tested it.

PR #5271, opened 2016-11-22
Remove the custom version check in the __init__ of pygpu.

abergeron:
We will now use the soversion like for everything else.

### DIFF
--- a/theano/gpuarray/__init__.py
+++ b/theano/gpuarray/__init__.py
@@ -41,24 +41,6 @@ register_transfer(transfer)
 
 
 def init_dev(dev, name=None):
-    v = pygpu.gpuarray.api_version()
-    expected = -9997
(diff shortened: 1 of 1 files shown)


nouiz:
@lamblin are you ok to use shared lib versioning for libgpuarray work?

What about pygpu versioning? We didn't had check related to that should we?

lamblin:
Sure, why not?
For pygpu, we can just use the version of the Python package, like the usual dependencies.

nouiz:
@abergeron, can you add a check for pygpu version here? With that I'm good
with the PR.

abergeron:
There is currently no way to check the pygpu version.

abergeron:
This needs https://github.com/Theano/libgpuarray/pull/300

nouiz:
jenkins test this

Issue #5263, opened 2016-11-21
Add optimizers for multiplication of exponentials


ajax2leet:
Theano currently doesn't seem to optimize expressions like a**(b) * a**(c) to a**(b+c). Since exponentiation is expensivem this would be a useful optimization to have by default.

nouiz:
Yes, it would be great to add this. Do you want to try to do it? If so, there is documentation and exerices in the advanced part of this tutorial:

https://github.com/abergeron/ccw_tutorial_theano

I do not think the core of Theano will work on this shortly.

PR #5272, opened 2016-11-23
Added optimization for product of logarithms with same base

ajax2leet:
Implements #5263

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -6223,6 +6223,33 @@ add_canonizer = in2out(gof.LocalOptGroup(local_add_canonizer, local_fill_cut,
 register_canonicalize(local_add_canonizer, name='local_add_canonizer')
 
 
+@register_canonicalize
+@gof.local_optimizer([T.mul])
+def local_mul_exp(node):
+    if not node.op == T.mul:
+        return False
+
+    pow_base_dict = defaultdict(list)
+    unchanged_inputs = []
+    pow_nodes_cnt = 0
+    for i in node.inputs:
+        if i.owner is not None and isinstance(i.owner.op, T.Elemwise) and\
+           i.owner.op.scalar_op in [T.pow, theano.scalar.basic.pow]:
+            assert len(i.owner.inputs) == 2
+            base = i.owner.inputs[0]
+            pow_base_dict[base].append(i.owner.inputs[1])
+            pow_nodes_cnt += 1
+        else:
+            unchanged_inputs.append(i)
+    if pow_nodes_cnt > 1:
+        mul_list = unchanged_inputs
+        for base, exp in pow_base_dict.iteritems():
+            mul_list.append(T.pow(base, T.add(*exp)))
+        rval = T.mul(*mul_list)
+        return [broadcast_like(rval, node.outputs[0], node.fgraph)]
+    else:
+        return False
+
 ##################
 # Distributivity #
 ##################


theano/tensor/opt.py
```diff
@@ -6223,6 +6223,33 @@ def add_calculate(num, denum, aslist=False, out_type=None):
...
+            assert len(i.owner.inputs) == 2
+            base = i.owner.inputs[0]
+            pow_base_dict[base].append(i.owner.inputs[1])
+            pow_nodes_cnt += 1
+        else:
+            unchanged_inputs.append(i)
+    if pow_nodes_cnt > 1:
+        mul_list = unchanged_inputs
+        for base, exp in iteritems(pow_base_dict):
+            mul_list.append(T.pow(base, T.add(*exp)))
```
nouiz: This is good, but would request extra work by the optimizer to clean up some corner cases.

If exp is a list of only 1 element, the add will be useless and will be removed by later optimization pass. I would prefer that we don't introduce it in the first place. Can you do something like:
```
if len(exp) > 1:
    tmp = T.add(*exp)
else:
    tmp = exp[0]
T.pow(base, tmp)
```

theano/tensor/opt.py
```diff
@@ -6223,6 +6223,33 @@ def add_calculate(num, denum, aslist=False, out_type=None):
...
+        rval = T.mul(*mul_list)
```
nouiz: Idem if mul_list is of size 1, don't call T.mul.

theano/tensor/opt.py
```diff
@@ -6223,6 +6223,33 @@ def add_calculate(num, denum, aslist=False, out_type=None):
 register_canonicalize(local_add_canonizer, name='local_add_canonizer')
 
 
+@register_canonicalize
```
nouiz: Where to register it is complicated. We need to register after the local_greedy_distributor optimization as otherwise, having local_mul_exp before/at the same time could remove some simplification done by local_greedy_distributor. As local_greedy_distributor is in canonicalize, this new opt can't be there.

This new opt should be done before local_pow_specialize that is is specialize, so it can't be in specialize.

Ideally, we don't want to add an extra pass on the graph between those 2 pass. We shouldn't add it in stabilize as this isn't stabilization.

So in conclusion, remove that line and use that diff to register it. This add it in an existing pass that isn't related, so they won't colide.

```
diff --git a/theano/tensor/opt.py b/theano/tensor/opt.py
index a00af70..5f57e21 100644
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5737,7 +5739,6 @@ def local_pow_canonicalize(node):
             return [broadcast_like(node.inputs[0], node.outputs[0], node.fgraph)]
     else:
         return False
-register_canonicalize(local_pow_canonicalize)
 
 
 @register_specialize
@@ -6193,7 +6194,11 @@ def local_log_sum_exp(node):
 
 
 compile.optdb.register('local_log_sum_exp',
-                       in2out(local_log_sum_exp, ignore_newtrees=True),
+                       in2out([local_log_sum_exp, local_mul_exp],
+                              ignore_newtrees=True),
+                       # between stabilize at 1.5 and specialize at 2
+                       # local_mul_exp should be after local_greedy_distributor
+                       # and before local_pow_specialize
                        1.6, 'fast_run')
 
```

ajax2leet: Isn't the registration for local_pow_canonicalize missing in your patch?

nouiz: right, it should be just the second part:
```
diff --git a/theano/tensor/opt.py b/theano/tensor/opt.py
index a00af70..5f57e21 100644
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -6193,7 +6194,11 @@ def local_log_sum_exp(node):
 
 
 compile.optdb.register('local_log_sum_exp',
-                       in2out(local_log_sum_exp, ignore_newtrees=True),
+                       in2out([local_log_sum_exp, local_mul_exp],
+                              ignore_newtrees=True),
+                       # between stabilize at 1.5 and specialize at 2
+                       # local_mul_exp should be after local_greedy_distributor
+                       # and before local_pow_specialize
                        1.6, 'fast_run')
```

theano/tensor/opt.py
```diff
@@ -6223,6 +6218,46 @@ def add_calculate(num, denum, aslist=False, out_type=None):
...
+        return False
+
+compile.optdb.register('local_log_sum_exp',
+                       in2out([local_log_sum_exp,
+                               local_mul_exp],
+                              ignore_newtrees=True),
+                       # between stabilize at 1.5 and specialize at 2
+                       # local_mul_exp should be after local_greedy_distributor
+                       # and before local_pow_specialize
+                       1.6, 'fast_run')
```
nouiz: I totally forget, this need tests. Can you add tests in theano/tensor/tests/test_opt.py for this? You can use the test test_log1p as an example.

Otherwise, it seem good to me.

ajax2leet:
Sure, I'll add a test.
I just noticed, that no further optimization is performed on the 
exponent, once the optimizer has collected all exponents into one, eg. 
x**a * x**(-a) -> x**(a + (-a)). It seems that some form of 
add_canonicalization should run after local_mul_exp, although I'm not 
quite sure how to implement this.

theano/tensor/opt.py
```diff
@@ -6251,8 +6250,8 @@ def local_mul_exp(node):
         return False
 
 compile.optdb.register('local_log_sum_exp',
-                       in2out([local_log_sum_exp,
-                               local_mul_exp],
+                       in2out(local_log_sum_exp,
+                              local_mul_exp,
                               ignore_newtrees=True),
                        # between stabilize at 1.5 and specialize at 2
                        # local_mul_exp should be after local_greedy_distributor
```
ajax2leet: Thanks for the hint. With your patch, `x**y* y**(-y)` gets optimized to 
`Elemwise{Pow}[(0, 1)] [id A] ''   
 |<TensorType(float64, vector)> [id B]
 |Elemwise{Add}[(0, 1)] [id C] ''   
   |<TensorType(float64, vector)> [id D]
   |Elemwise{neg,no_inplace} [id E] ''   
     |<TensorType(float64, vector)> [id D]
`

Something is still missing :)

Issue #5253, opened 2016-11-18
Make join work inplace on the first input


nouiz:
In some cases like in gh-5106, we will do Join of 2 elements and sometimes the second elements will be empty.

In that case, it would be useful that Join work inplace.

So modify the CPU and new gpu back-end Join op to allow to be inplace on the first inputs and make the inplace optimization needed for that.

Register that inplace optimization in the last inplace optimization to register.

This will lower memory usage by not creating the useless temporary and speed up be not doing the alloc and the copy.
- [ ] Add in Join an inplace props and make the C and Python code use it.
- [ ] Do the same for GPUAJoin in the new gpu back-end.

nouiz:
@ReyhaneAskari

PR #5276, opened 2016-11-23
Make join work inplace

ReyhaneAskari:
Hi @nouiz ,
Regarding issue #5253,  I changed the perform function to output a view of input in the case of only one non-empty input. How can we test if it is working correctly? The graph seems not to change as expected.

### DIFF
--- a/doc/tutorial/shape_info.txt
+++ b/doc/tutorial/shape_info.txt
@@ -65,7 +65,7 @@ array([8, 4])
 
 >>> f = theano.function([x,y], z)# Do not take the shape.
 >>> theano.printing.debugprint(f) # doctest: +NORMALIZE_WHITESPACE
-Join [id A] ''   0
+Join{view=-1} [id A] ''   0
  |TensorConstant{0} [id B]
  |x [id C]
  |y [id D]
--- a/doc/tutorial/shape_info.txt
+++ b/doc/tutorial/shape_info.txt
@@ -65,7 +65,7 @@ array([8, 4])
 
 >>> f = theano.function([x,y], z)# Do not take the shape.
 >>> theano.printing.debugprint(f) # doctest: +NORMALIZE_WHITESPACE
-Join{view=-1} [id A] ''   0
+Join [id A] ''   0
  |TensorConstant{0} [id B]
  |x [id C]
  |y [id D]
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -3888,10 +3888,12 @@ class Join(Op):
     check_input = False
     __props__ = ("view",)
 
-    def __init__(self, view=False):
+    def __init__(self, view=-1):
         self.view = view
-        if view:
-            self.view_map = {0: [0]}
+        if view != -1:
+            # since the first input is always the axis, the tensors
+            # start from index 1.
+            self.view_map = {0: [1 + view]}
 
     def make_node(self, *axis_and_tensors):
         """
@@ -4003,12 +4005,13 @@ class Join(Op):
 
     def perform(self, node, axis_and_tensors, out_):
         out, = out_
+        view = self.view
         axis, tensors = axis_and_tensors[0], axis_and_tensors[1:]
-        # tailing tensors are all tensors except the first one
-        tailing_tensors_are_empty = numpy.all(
-            [tensor.shape[axis] == 0 for tensor in axis_and_tensors[2:]])
-        if tailing_tensors_are_empty:
-            out[0] = tensors[0]
+        # we check these tensors for being empty.
+        if (view != -1) and numpy.all(
+                [tensor.shape[axis] == 0 for tensor in
+                 tensors[0:view] + tensors[view + 1:]]):
+            out[0] = tensors[view]
 
         else:
             ndim = tensors[0].ndim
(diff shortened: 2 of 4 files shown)


theano/tensor/basic.py
```diff
@@ -3965,18 +3970,27 @@ def _make_node_internal(self, axis, tensors,
...
+            ndim = tensors[0].ndim
+            if axis < -ndim:
+                raise IndexError("Join axis %d out of bounds [0, %d)" %
+                                 (axis, ndim))
+
+            out[0] = theano._asarray(numpy.concatenate(tensors, axis=axis),
+                                     dtype=node.outputs[0].type.dtype)
+        elif num_nonempty == 1:
+            nonempty_tensor_index = numpy.argmax(nonempty_tensors)
+            self.view_map = {0: [nonempty_tensor_index]}
```
nouiz: you can't change the view_map or destroy_map at run time. You need to check which input isn't empty and it is isn't the value of self.view_map[0], then we can't return a view.

theano/tensor/basic.py
```diff
@@ -3965,18 +3970,27 @@ def _make_node_internal(self, axis, tensors,
...
         out, = out_
         axis, tensors = axis_and_tensors[0], axis_and_tensors[1:]
-        ndim = tensors[0].ndim
-        if axis < -ndim:
-            raise IndexError("Join axis %d out of bounds [0, %d)" %
-                             (axis, ndim))
-
-        out[0] = theano._asarray(numpy.concatenate(tensors, axis=axis),
-                                 dtype=node.outputs[0].type.dtype)
+        nonempty_tensors = [0 if tensor.size == 0 else 1 for tensor in tensors]
```
nouiz: This will loose some existing error. You can't check the total size, you need to check the specific axis we concatenate. It should be something like `[tensor.shape[axis] == 0 for tensor in tensors]`

theano/tensor/basic.py
```diff
@@ -3965,18 +3970,27 @@ def _make_node_internal(self, axis, tensors,
...
+        num_nonempty = numpy.sum(nonempty_tensors)
```
nouiz: with the code change above, you can do any(num_nonempty), this would be epsilon faster.

theano/tensor/basic.py
```diff
@@ -3965,18 +3970,27 @@ def _make_node_internal(self, axis, tensors,
...
+        # checking if more than one non-empty tensors are joined.
+        if num_nonempty > 1:
```
nouiz: I would go in the reverse direction for the test.

if num_empty == 1 and the non empty one is the self.view_map (so currently the inputs 0):
   do the new logic:

else:

    do the old logic.

doc/tutorial/shape_info.txt
```diff
@@ -65,7 +65,7 @@ array([8, 4])
 
 >>> f = theano.function([x,y], z)# Do not take the shape.
 >>> theano.printing.debugprint(f) # doctest: +NORMALIZE_WHITESPACE
-Join [id A] ''   0
+Join{view=-1} [id A] ''   0
```
nouiz: I think I would add a \__str__ method to don't show that. Something like:
```
def __str__(self):
    if self.view == -1:
        return "Join"
    else:
        return super(Join, self).__str__()
```

This won't polute the printing for an option that is rarely used.

theano/tensor/basic.py
```diff
@@ -3964,49 +3976,70 @@ def _make_node_internal(self, axis, tensors,
...
+            //PyObject* PyArray_Concatenate(PyObject* obj, int axis)
+            int ndim = PyArray_NDIM(%(input_1)s);
+            if( axis < -ndim ){
+                PyErr_Format(PyExc_IndexError,
+                             "Join axis %%d out of bounds [0, %%d)", axis, ndim);
+                %(fail)s
+            }
+            Py_XDECREF(%(out)s);
+            %(out)s = (PyArrayObject *)PyArray_Concatenate(list, axis);
+             (list);
```
nouiz: This line miss Py_XDECREF that got cut. it was line 4007

doc/tutorial/shape_info.txt
ReyhaneAskari: Thanks, Done.

theano/tensor/basic.py
ReyhaneAskari: Right! I must have accidentally deleted it! Previously it was```Py_DECREF(list);```, should I change it to ```Py_XDECREF(list)```?

ReyhaneAskari:
@nouiz , I could not find concatenate in ```theano/scan_module/scan_checkpoints.py```.  Could you elaborate on what should be done and where? 

Thanks

nouiz:
rebase this PR and you will see the concatenate.

nouiz:
no, keep the old code. We know list will never be NULL, so this will be
marginally faster.

ReyhaneAskari:
@nouiz , I changed the code in ```theano/scan_module/scan_checkpoints.py``` to use join (with view) instead of concatenate as I think you would not want to change the signature of concatenate to keep it like the signature of numpy.

ReyhaneAskari:
Perfect! Thanks!

nouiz:
merging now as jenkins isn't needed to test this PR and it have problems right now.

PR #5287, opened 2016-11-25
Fix in advanced_inc_subtensor1 if index is broadcastable

### LINKED ISSUE #3575: inc_subtensor of column vector
Trying to set a column vector using inc_subtensor gives an error.  E.g.:

The following code attempts to set the first two columns in a matrix and works fine:

```
import numpy, theano 
import theano.tensor as T
mat = theano.shared(numpy.zeros((10,10)))
temp = numpy.asarray(range(20))
temp.shape = (10,2)
mat_inc = theano.shared(temp)
mat = T.inc_subtensor(mat[:,[0,1]], mat_inc)
print mat.eval()
```

The following code attempts to set the first one columns in a matrix, but returns an error:

```
import numpy, theano
import theano.tensor as T
mat = theano.shared(numpy.zeros((10,10)))
temp = numpy.asarray(range(20))
temp.shape = (10,1)
mat_inc = theano.shared(temp)
mat = T.inc_subtensor(mat[:,[0]], mat_inc)
print mat.eval()
```

The following code works as expected, but isn't as general.  I'd like to use the above approach, but if I don't know which columns I want to set beforehand, it causes a problem.  

```
import numpy, theano
import theano.tensor as T
mat = theano.shared(numpy.zeros((10,10)))
temp = numpy.asarray(range(10))
mat_inc = theano.shared(temp)
mat = T.inc_subtensor(mat[:,0], mat_inc)
print mat.eval()
```

lamblin:
Fixes #3575

### DIFF
--- a/theano/tensor/subtensor.py
+++ b/theano/tensor/subtensor.py
@@ -1157,7 +1157,11 @@ def inc_subtensor(x, y, inplace=False, set_instead_of_inc=False,
             inplace=inplace,
             set_instead_of_inc=set_instead_of_inc,
             tolerate_inplace_aliasing=tolerate_inplace_aliasing)
-        return x.owner.op(inner_incsubtensor, *x.owner.inputs[1:])
+        # The broadcastable pattern of inner_x may not be the same as
+        # the one of x, so we have to build a new dimshuffle here,
+        # instead of reusing x.owner.op().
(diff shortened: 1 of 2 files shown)


nouiz:
travis failed

lamblin:
It was just a matter of casting the input value to floatX, should be fixed now.

nouiz:
still ok to merge when travis pass

Issue #4916, opened 2016-08-30
Crash in MRG_RandomStreams with the new backend


abergeron:
I was a case where we end up in a situation where mrg_uniform gets asked for c_code with a GpuArrayType, meaning the optimization to replace it was skipped somehow.

I'll try to gather more information.

abergeron:
This only happens with compute_test_values.

We would need to fix the op and the optimization for the new backend to introduce/remove transfers.

nouiz:
A new duplicate issue about this contains test code that reproduce the problem:

https://github.com/Theano/Theano/issues/5279

PR #5288, opened 2016-11-25
fix crash in MRG_RandomStreams with the new backend

chinnadhurai:
Moved the make_node() from the base class "mrg_uniform_base" to the derived classes ( CPU, old GPU and new GPU backend) to avoid incorrect dtype settings for "rstate"

fix gh-4916

### DIFF
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -24,7 +24,8 @@ from . import multinomial
 
 import theano.sandbox.cuda
 from theano.sandbox.cuda import GpuOp
-from theano.gpuarray.basic_ops import GpuKernelBase, Kernel, infer_context_name
+from theano.sandbox.cuda.basic_ops import as_cuda_ndarray_variable
+from theano.gpuarray.basic_ops import GpuKernelBase, Kernel, infer_context_name,as_gpuarray_variable
 from theano.gpuarray.type import GpuArrayType
 from theano.gpuarray.fp16_help import write_w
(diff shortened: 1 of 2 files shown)


slefrancois:
jenkins test this

slefrancois:
There are flake8 errors to correct:
http://darjeeling.iro.umontreal.ca:8080/job/Theano%20PR%20multijob/601/testReport/junit/theano.tests/test_flake8/test_format_flake8/

theano/sandbox/rng_mrg.py
```diff
@@ -809,6 +826,20 @@ class GPUA_mrg_uniform(GpuKernelBase, mrg_uniform_base):
...
+    def make_node(self, rstate, size):
+        # error checking slightly redundant here, since
+        # this op should not be called directly.
+        #
+        # call through MRG_RandomStreams instead.
+        broad = []
+        for i in range(self.output_type.ndim):
+                broad.append(tensor.extract_constant(size[i]) == 1)
+        output_type = self.output_type.clone(broadcastable=broad)()
+        rstate = as_gpuarray_variable(rstate,infer_context_name(rstate))
```
slefrancois: @nouiz, should we add an output_type method?

slefrancois:
jenkins test this

slefrancois:
@nouiz, looks good to me, do you want to double-check and merge?

nouiz:
I'm ok with this, but it would be great to add a test for this. Just to be sure we don't bring back the error.

slefrancois:
jenkins test this

Issue #4664, opened 2016-06-23
Theano 0.8.1 ImportError in Winpython 3.5.1 64bit, no GPU


fnorlund:
Hello,
I've just installed WinPython 3.5.1.3 64bit including Theano 0.8.1 on a Windows 10 machine. The compiler is TDM-GCC-64 and I'm not using a GPU. 

After running a theano demo example I get the following error:
>>> import theano
>>> from theano import tensor
>>> a = tensor.dscalar()
>>> b = tensor.dscalar()
>>> c = a + b
>>> f = theano.function([a,b], c)

Traceback (most recent call last):
  File "<pyshell#11>", line 1, in <module>
    f = theano.function([a,b], c)
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\compile\function.py", line 320, in function
    output_keys=output_keys)
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\compile\pfunc.py", line 479, in pfunc
    output_keys=output_keys)
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\compile\function_module.py", line 1777, in orig_function
    defaults)
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\compile\function_module.py", line 1641, in create
    input_storage=input_storage_lists, storage_map=storage_map)
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\gof\link.py", line 690, in make_thunk
    storage_map=storage_map)[:3]
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\gof\vm.py", line 1003, in make_all
    no_recycling))
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\gof\op.py", line 970, in make_thunk
    no_recycling)
  File "F:\Program_Files\WinPython-64bit-3.5.1.3\python-3.5.1.amd64\lib\site-packages\theano\gof\op.py", line 879, in make_c_thunk
    output_storage=node_output_storage)
[...]

PR #5290, opened 2016-11-25
Conda fixes

mingwandroid:
Because I pushed a version of this branch which was the same as master, Github seems to have closed my old PR against it. Reopening here instead, shame we lost the conversation thread.

I made some test releases for all OSes that Conda supports on https://anaconda.org/rdonnelly from the latest code as of about 11am UTC. This PR has been rebased to the latest master branch.

If you have time to check them I'd appreciate it.

The conda-recipes are at:
https://github.com/mingwandroid/conda-recipes/tree/master/libgpuarray (this one has two patches, I can make a PR for them if you wish)
https://github.com/mingwandroid/conda-recipes/tree/master/python/pygpu (again, two patches).
https://github.com/mingwandroid/conda-recipes/tree/master/python/theano


mingwandroid:
I should say, I plan to rebuild the theano packages when rc1 is ready so people can test them.

theano/configdefaults.py
```diff
@@ -1313,13 +1310,10 @@ def default_blas_ldflags():
...
-        # This fix some case with Anaconda 2.3 on Linux.
-        # Newer Anaconda still have this problem but only have
-        # Continuum in sys.version.
-        if (("Anaconda" in sys.version or
-             "Continuum" in sys.version) and
-                "linux" in sys.platform):
+        # Add sys.prefix/lib to the runtime search path. On
+        # non-system installations of Python that use the
+        # system linker, this is generally neccesary.
+        if sys.platform in ("linux", "darwin"):
```
nouiz: @abergeron I think this is good and would help fix Mac problem. Just to make you know about this change to be merged soon.

theano/configdefaults.py
```diff
@@ -430,7 +430,7 @@ def f(suffix=suffix):
         except OSError:
             rc = 1
         if rc != 0:
-            _logger.warning("conda g++ not available, use `conda install m2w64-toolchain`")
+            _logger.warning("g++ not available, if using conda: `conda install m2w64-toolchain`")
```
nouiz: I'm not sure about this warning. I think it will be triggered too often. People not using conda could need the code after this and won't have mkl. I see 2 possibles fix:

1) Do the check for conda and keep the warning here. I prefer this option, I think it is fine to have only one place with a check for conda. This will ensure that conda user use mkl.
2) Just record here that mkl isn't available. Then if we can't find a blas in this function, raise this warning. So mostly, postpone this warning to the end of this method in case we don't find a blas.

theano/__init__.py
```diff
@@ -31,10 +31,6 @@
 
 import sys
 
-if sys.platform == 'win32' and sys.version_info[0:2] == (3, 5):
```
nouiz: Here is an error report with WinPython 3.5 on windows:

https://github.com/Theano/Theano/issues/4664#issuecomment-263134608

I don't have objection to removing the check, but I fear that there is still some python 3.5 on windows that have problems. conda and the official python are fine from what you tell. Do you have an idea why winpython have a problem?

People with python 3.5 not support will still get an error, just an error very hard to understand. If you have an idea how we could detect here if we support the current Python 3.5 on Windows, it would be great.

Otherwise, I think we will keep this and probably add a note in the installation of windows that it is not all python 3.5 that are supported.

theano/configdefaults.py
mingwandroid: OK, I will change it to look for "Continuum", or anything containing "conda" as that will get both "conda-forge" and "Anaconda".

notoraptor:
Hello ! Just to say that I have just tested this PR with the test `theano.misc.check_blas` on my computer (Windows 8.1 64 bits) and it works (for both Python 2 and 3).

These are the commands executed:

### For Python 2:
```batch
conda create --yes -n python2test python=2
activate python2test
conda install --yes numpy scipy mkl-service nose sphinx pydot-ng m2w64-gcc libpython
pip install git+https://github.com/mingwandroid/Theano.git@conda-fixes
python -m theano.misc.check_blas > log-python2.log 2>&1

python -V
rem Python 2.7.12 :: Continuum Analytics, Inc.
python -c "import platform; print (platform.platform())"
rem Windows-8.1-6.3.9600
python -c "import theano; print(theano.__version__)"
rem 0.9.0dev4.dev-c18e654db3b38c8d31b2cb5d94367e28d62f44aa

deactivate
conda remove --yes --all -n python2test
```
### For Python 3:
```batch
conda create --yes -n python3test python=3
activate python3test
conda install --yes numpy scipy mkl-service nose sphinx m2w64-gcc libpython
pip install pydot
pip install git+https://github.com/mingwandroid/Theano.git@conda-fixes
python -m theano.misc.check_blas > log-python3.log 2>&1

python -V
rem Python 3.5.2 :: Continuum Analytics, Inc.
python -c "import platform; print (platform.platform())"
rem Windows-8.1-6.3.9600-SP0
python -c "import theano; print(theano.__version__)"
rem 0.9.0dev4.dev-c18e654db3b38c8d31b2cb5d94367e28d62f44aa

deactivate
conda remove --yes --all -n python3test
```

Some remarks:
* On my computer I need to install `m2w64-gcc` package. Default conda package `mingw` does not work.
* I had installed the package `mkl-service` to prevent a warning (not an error) that theano prints when running `theano.misc.check_blas`:
```
GTXWARNING (theano.configdefaults): mkl is not available, if using conda: `conda install mkl-service`: No module named mkl
```



@nouiz

mingwandroid:
Thanks for checking.

`mingw` package is deprecated. It never supported 64-bit. We use `mingw-w64` packages from MSYS2 instead.

All those `conda` packages you mention are of course listed as dependencies for the `theano` package so end-users need only to install that.

mingwandroid:
Oh, `m2w64-toolchain` is recommended instead of `m2w64-gcc`.

mingwandroid:
I do not have winpython 3.5 installed, but if you wish, I can try to take a look?

nouiz:
Travis have some errors. Some some of them:

https://travis-ci.org/Theano/Theano/jobs/179590558#L532
https://travis-ci.org/Theano/Theano/jobs/179590558#L572

theano/configdefaults.py
```diff
@@ -1237,39 +1251,38 @@ def default_blas_ldflags():
...
+        # numpy and scipy.
+        try:
+            import mkl  # noqa
+        except ImportError as e:
+           if any([m for m in ('conda', 'Continuum') if m in sys.version]):
+               _logger.warning('install mkl with `conda install mkl-service`: %s', e)
+        else:
+            # This branch is executed if no exception was raised
+            if sys.platform == "win32":
+                lib_path = os.path.join(sys.prefix, 'Library', 'bin')
```
nouiz: On windows, lib_path is a list and otherwise, it is a list. I thin it should always be a list, so wrap this in a list here.

theano/configdefaults.py
```diff
@@ -1237,39 +1251,38 @@ def default_blas_ldflags():
...
+                                           "mkl_intel_thread",
+                                           "mkl_rt"]]
+            res = try_blas_flag(flags)
+            if res:
+                return res
+            flags.extend(['-Wl,-rpath,' + l for l in
+                          blas_info.get('library_dirs', [])])
+            res = try_blas_flag(flags)
+            if res:
+                maybe_add_to_os_environ_pathlist('PATH', lib_path)
```
nouiz: Now that lib_path will always be a list, use lib_path[0]

mingwandroid:
Is this ready to go? I believe I've addressed all the change requests? Thanks.

theano/configdefaults.py
```diff
@@ -418,6 +419,19 @@ def f(suffix=suffix):
 except OSError:
     rc = 1
 
+# Anaconda on Windows has mingw-w64 packages including GCC, but it may not be on PATH.
+if rc != 0:
+    if sys.platform == "win32":
+        mingw_w64_gcc = os.path.join(os.path.dirname(sys.executable), "Library", "mingw-w64", "bin", "g++")
```
abergeron: Is this supported for 32-bits?

theano/configdefaults.py
```diff
@@ -418,6 +419,19 @@ def f(suffix=suffix):
...
+        try:
+            rc = call_subprocess_Popen([mingw_w64_gcc, '-v'])
+            if rc == 0:
+                maybe_add_to_os_environ_pathlist('PATH', os.path.dirname(mingw_w64_gcc))
```
abergeron: Also, the value of the cxx config param is a full path, so we can set it to the full path to g++ here and not fiddle with PATH.

theano/configdefaults.py
abergeron: If it does help on mac, that's great. But from what I observed I doubt it. We should test that.

theano/configdefaults.py
mingwandroid: Yes. That supports both no problem.

mingw-w64 is just the project name to differentiate it from mingw.org.

theano/configdefaults.py
mingwandroid: I need to add to PATH I am afraid, otherwise companion tools (linker, cc1) are not found.

theano/configdefaults.py
abergeron: Ok

theano/configdefaults.py
abergeron: Ok

nouiz:
Thanks for all the work. I think all is good for this PR.

This week we should do a beta and not a release candidate. We want to make a release this week and there is probably too much stuff to finish this week, so we will go for a beta.

I lost the status of the conda packages of Theano and libgpuarray. Is there still work to do for them? thanks.

mingwandroid:
Thanks for the reviews. Conda-wise I am in pretty good shape now I think. Everything appears to be working and all my patches are upstreamed.

I don't have an Nvidia GPU nor OpenCL capable AMD so that side of things I cannot test so easily which is a shame. When you are ready for the beta I will build new packages and if you were able to test them out on such hardware/drivers then that would be great.

I did notice a good few warnings from libgpuarray on Windows and some looked like they warranted more investigation so when I rebuild them I will check into that and send logs to you both.

Issue #5261, opened 2016-11-21
Unclosed files


enfeizhan:
Got a lot of warnings saying files unclosed.

> 

Using gpu device 0: GeForce GT 650M (CNMeM is disabled, cuDNN Version is too old. Update to v5, was 2000.)
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/cmodule.py:2182: ResourceWarning: unclosed file <_io.BufferedWriter name='/dev/null'>
  p_out = output_subprocess_Popen(cmd)
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/cmodule.py:2182: ResourceWarning: unclosed file <_io.BufferedWriter name='/dev/null'>
  p_out = output_subprocess_Popen(cmd)
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/cmodule.py:2182: ResourceWarning: unclosed file <_io.BufferedWriter name='/dev/null'>
  p_out = output_subprocess_Popen(cmd)
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/cmodule.py:2182: ResourceWarning: unclosed file <_io.BufferedWriter name='/dev/null'>
  p_out = output_subprocess_Popen(cmd)
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/cmodule.py:2182: ResourceWarning: unclosed file <_io.BufferedWriter name='/dev/null'>
  p_out = output_subprocess_Popen(cmd)
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/utils.py:497: ResourceWarning: unclosed file <_io.BufferedReader name='/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/sandbox/cuda/cuda_ndarray.cuh'>
  return hash_from_code(open(file_path, 'rb').read())
/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/gof/utils.py:497: ResourceWarning: unclosed file <_io.BufferedReader name='/Users/feizhan/miniconda/envs/py3_env/lib/python3.4/site-packages/theano/sandbox/cuda/cuda_ndarray.cuh'>
  return hash_from_code(open(file_path, 'rb').read())
[...]

PR #5294, opened 2016-11-28
Close file as reading file done

enfeizhan:
Received warning while running test saying file unclosed (line 571 in old version). Close it using ``with``.

Suppose this is a tiny change so didn't run tests.  

The function ``hash_from_dict`` is from merged upstream.

fix gh-5261

### DIFF
--- a/theano/gof/utils.py
+++ b/theano/gof/utils.py
@@ -569,7 +569,9 @@ def hash_from_file(file_path):
     Return the MD5 hash of a file.
 
     """
-    return hash_from_code(open(file_path, 'rb').read())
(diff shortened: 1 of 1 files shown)


aam-at:
You should use rebase instead of merge.

nouiz:
aam-at is right. We prefer to not have merge in PR. Also, travis is failing and I think the error will be fixed a rebase. I think the error is related to the merge.

thanks for finding the cause and the PR.

nouiz:
I rebased it in https://github.com/Theano/Theano/pull/5294

thanks

Issue #5011, opened 2016-09-23
Print at exit some profile stats by default


nouiz:
Currently, we only print at exit if Theano profiler is used.

But we have all the information to print a few top level stats that would be very useful. 
- [ ] Make a new python function print_global_stats() in theano.compile.profiling.py that will print the stats above in one line.
- [ ] Make a new Theano flag profiling.print_global_stats that when True(default False) will call at exit that function. This is done in the file theano/configdefaults.py, check how force_device is done. It it a good example for this case.

Stats to print in one line.
- Time since import of Theano
- Time spend inside Theano functions.
- Time spent compiling Theano function (graph optimization and linker separate)

How to implement:
- Time since import of Theano is available in theano.compile.profiling.theano_imported_time
- Time spend inside Theano functions.
  - To have the number add a variable total_time_execution_theano_function in theano.compile.profiling. In the file theano.compile.function_module.py in Function.\__call__, search the variable dt_call. Add this to total_execution_theano_function. 
- Time spent compiling Theano function.
  - Print 2 numbers, graph optimization time and linker/c code generation. Regular Theano user don't know well the difference. Do not print the sum of those numbers.
  - To have those number add the variables total_time_graph_optimization and total_time_linker in theano.compile.profiling. In that file, in FunctionMaker.\__init__, search for opt_time. Add it to the global. Search for the variable linker_time in FunctionMaker.create. Add it to the global variable.

Look in the file compile/profiling.py and print the information in the method function _atexit_print_fn() if the flag is set.

NeilGirdhar:
Please consider putting this into a top-level method instead of just printing this at exit:

- a long-running simulation might want to print the statistics, and
- unit tests don't want to display this.

Even better would be to wrap your theano initialization code into a object.  TensorFlow calls this a "Session".  The statistics method would ideally be a method on the session.  Having a session would be beneficial when Theano is conditionally used (for example in a test).  It is a pain to work around the fact that any import of Theano constitutes initialization of a session.

nouiz:
I like the idea to make this function accesible to user.

Can you explain what is the problem related to initialization at theano
import? We already discussion doing a Session type (called interactive env
then) long before TensorFlow was even started and I don't recall that it
helped solve initialization problem. By default, if you let device=cpu, you
can manually enable the GPU afterward. Is that what you are looking for?

NeilGirdhar:
[...]

PR #5298, opened 2016-11-28
Print profile stats

Faruk-Ahmed:
reference #5011

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -126,6 +126,12 @@ AddConfigVar(
     BoolParam(False, allow_override=False),
     in_c_key=False)
 
+AddConfigVar(
+    'print_global_stats',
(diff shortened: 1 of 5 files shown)


ballasn:
PR looks good,

ballasn:
Change looks good, and works locally on my machine. However you do have some conflicts now, you probably need to do a rebase before we can merge,

theano/configdefaults.py
```diff
@@ -126,6 +126,12 @@ def __str__(self):
     BoolParam(False, allow_override=False),
     in_c_key=False)
 
+AddConfigVar(
```
nouiz: I forgot, this should go in the doc in the file doc/library/config.txt.

It would be good to also include a paragraph about it in the file doc/tutorial/profiling.txt

Can you do that too?

thanks

PR #5299, opened 2016-11-29
Improve MissingInput message

fvisin:
This PR adds info on which input of the node has not been provided as an input to the theano function.


theano/gof/fg.py
```diff
@@ -380,19 +380,20 @@ def __import__(self, apply_node, check=True, reason=None):
             for node in new_nodes:
                 if hasattr(node, 'fgraph') and node.fgraph is not self:
                     raise Exception("%s is already owned by another fgraph" % node)
-                for r in node.inputs:
+                for idx, r in enumerate(node.inputs):
```
nouiz: You can be very surprised how slow enumerate can be. As this code is used for import the original graph AND all the changes done in the graph, I would prefer that you don't do it like that.

You can remove the enumerate and call "node.inputs.index(r)" to compute the index only when needed. Also, you could call: theano.gof.utils.get_variable_trace_string(r) and append that to the error message. This will give the stack-trace when the user build that variable. 

Thanks for this good idea of error message to upgrade.

fvisin:
Thank you for your review. I implemented your suggestions, `theano.gof.utils.get_variable_trace_string(r)` is definitely a nice addition!

PR #5303, opened 2016-11-29
GPU gemv->dot speedup for new backend

khaotik:
Similar to #5257, this PR applies to the new backend.

~~**ONLY merge this after libgpuarray-294 is merged.**~~
To test this, newer `libgpuarray` with [this PR](https://github.com/Theano/libgpuarray/pull/294) merged is needed.

Results of running [this](https://gist.github.com/khaotik/9c7c72e562a919e241c9a09522fc4df5) test script on my laptop (GeForce GT 650m):

(time used in seconds, lower is better)

Before:
```
  scan with dot  |  scan with sum   |  full unrolled
 3.502193689346  |  0.411211490631  |  0.089465141296  
 1.447018623352  |  0.412187576294  |  0.086945295334  
 1.446414470673  |  0.411444425583  |  0.089369773865  
 1.447574138641  |  0.413084506989  |  0.091281890869  
 1.452363491058  |  0.412972927094  |  0.088017463684  
...
```

After:
```
  scan with dot  |  scan with sum   |  full unrolled
 2.389663457870  |  0.413246393204  |  0.089257478714  
 0.485583782196  |  0.413187980652  |  0.086747169495  
 0.486212968826  |  0.413235425949  |  0.088805198669  
 0.489252328873  |  0.413118600845  |  0.088519096375  
 0.484216928482  |  0.413124799728  |  0.087159156799  
...
```

I'm not sure this also works for OpenCL. Could use some help for testing. For OpenCL device, both Intel/AMD GPU, both clBLAS and clBLAST library, should be tested.

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -75,7 +75,6 @@ class GpuGemv(BlasOp):
         inplace = self.inplace
         if inplace and y.strides[0] < 0:
             inplace = False
-        print(alpha, beta)
         out_storage[0][0] = blas.gemv(alpha, A, x, beta, y,
                                       overwrite_y=inplace)
 
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -9,6 +9,7 @@ from theano.compile import optdb
 from theano.gof import LocalOptGroup
 from theano.tensor.basic import as_tensor_variable
 from theano.tensor.opt import in2out
+from theano.tensor.var import TensorConstant
 
 from .basic_ops import (GpuArrayType, CGpuKernelBase,
                         as_gpuarray_variable, gpu_contiguous, infer_context_name)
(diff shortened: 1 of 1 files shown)


theano/gpuarray/blas.py
```diff
@@ -52,6 +53,15 @@ def make_node(self, y, alpha, A, x, beta):
...
         beta = as_tensor_variable(beta).astype('float64')
+
+        # if alpha==1. and beta==0., we add runtime check
+        # for possible speed up using vector-vector dot as
+        # gemv tend to be slower for vector-vector dot
+        self._use_rt_dot_check = False
+        if all(map(lambda v: isinstance(v, TensorConstant), [alpha, beta])):
+            if alpha.value == 1. and beta.value == 0.:
+                self._use_rt_dot_check = True
+
```
nouiz: why this? I would remove this check and always do the case this is True. This way, the speed up will be there even if we don't know the shapes at compile time.

abergeron: This will also break Apply instances that share the op but don't have the same values for alpha and beta.  This check should be performed at runtime.

abergeron:
jenkins test this please

PR #5306, opened 2016-11-30
Try to get_scalar_constant_value for Assert nodes

gvtulder:
If the conditions of an `Assert` are constant and True, `get_scalar_constant_value` could return the value if it it also constant.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -488,6 +488,13 @@ def get_scalar_constant_value(orig_v, elemwise=True,
             elif isinstance(v.owner.op, (ScalarFromTensor, TensorFromScalar)):
                 v = v.owner.inputs[0]
                 continue
+            elif isinstance(v.owner.op, theano.tensor.opt.Assert):
+                # check if all conditions are constant and true
+                cond = [get_scalar_constant_value(c, max_recur=max_recur)
+                        for c in v.owner.inputs[1:]]
+                if builtins.all([0 == c.ndim and c != 0 for c in cond]):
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -488,6 +488,14 @@ def get_scalar_constant_value(orig_v, elemwise=True,
...
                 v = v.owner.inputs[0]
                 continue
+            elif isinstance(v.owner.op, theano.tensor.opt.Assert):
+                # check if all conditions are constant and true
+                cond = [get_scalar_constant_value(c, max_recur=max_recur)
+                        for c in v.owner.inputs[1:]]
+                if builtins.all([0 == c.ndim and c != 0 for c in cond]):
+                    v = v.owner.inputs[0]
+                # not all conditions are satisfied
+                continue
```
nouiz: The indentation of this line is not good. Add 4 spaces before continue.

Otherwise, this PR look good. Why did you need this case?

gvtulder: Changed now.

gvtulder:
I ran into this when I added an Assert in #5267. There I tried to test a shape tuple by wrapping it in an Assert, before passing it to `gpu_alloc_empty`. It turns out that `gpu_alloc_empty` uses `get_scalar_constant_value` to determine the broadcast pattern of the new array. Wrapping a constant in an Assert broke that bit.
I've now solved my original problem in a better way by testing constant values at compile time, but this would also have solved it. It might still be useful in other cases.

PR #5307, opened 2016-11-30
close file with context manager

nouiz:
rebased version of https://github.com/Theano/Theano/pull/5294


nouiz:
merging as travis passed and this don't need GPU tests.

PR #5309, opened 2016-12-01
Still merge identical Asserts while Assert merging is disabled

gvtulder:
At the moment `Assert`s are not being merged at all. This seems to be deliberate (#3344) because the fancy condition-merging logic caused problems. However, and this may be incorrect, it seems to me that merging identical Asserts shouldn't be a problem.

This patch re-enables merging of Asserts if they are identical, by temporarily disabling their special treatment. Since they're currently not being merged on a condition level, they should be treated like any other node.

Does that make sense? It seems to solve my problem in #5267.


nouiz:
thanks

PR #5315, opened 2016-12-01
Update NEWS_DEV.txt and .mailmap

nouiz:


nouiz:
jenkins failure is not related, so merging.

Issue #5197, opened 2016-11-08
theano.tensor.extra_ops.cumprod gradient handle zero incorrectly


khaotik:
I encountered this bug when trying to implement DeepMind's DNC model, here's a snippet to reproduce:

```python3
import theano as th
import theano.tensor as T

print(th.__version__)

x = T.fvector()
y = T.extra_ops.cumprod(x)
dx = th.grad(T.sum(y*y), x)
fn = th.function([x], dx)

print(fn([1.,2.])
print(fn([1.,0.])
print(dx)
```
Outputs:
```
0.9.0dev2.dev-c1dfa69ce084fc7be0fc18ccb79f3d44e9129f73
array([ 10.,   4.], dtype=float32)
array([  2.,  nan], dtype=float32)
Elemwise{true_div,no_inplace}.0
```
Note the `nan`, which is supposed to be `0.`

Turning optimization on/off, switching between GPU/CPU does not help, I  believe it's a bug. The cause is `cumprod` being converted into division op, which gives division by zero, producing `nan`.

lamblin:
Right, the gradient is doing a cumprod from the end, that includes the current index, but then divides that by the input.
I think we can fix that by shifting that cumprod by 1 to exclude the current input.
This would still fail if there are multiple zeros, though.

khaotik:
I do have an O(n) algorithm to fix this problem. (my comment on [this](https://github.com/tensorflow/tensorflow/issues/3862)) thread) I don't have a working implementation yet.

lamblin:
Great, that would indeed be a better way (I did not manage to get rid of the division either, I think there could be a way in O(l log n) for prod, not sure about cumprod yet).

PR #5317, opened 2016-12-02
Merge CumsumOp/CumprodOp into CumOp

khaotik:
Changes:
 - Merged `CumsumOp` and `CumprodOp` into `CumOp`
 - Changed `GpuCumsum` into `GpuCumOp`, now `cumprod` is available on new GPU backend

Notes:
 - `cumsum` and `cumprod` convenience functions remain syntactically same
 - Should we change the name of this Op? (If you get it)

### DIFF
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -407,6 +407,26 @@ def cumprod(x, axis=None):
     return CumOp(axis=axis, mode='mul')(x)
 
 
+# CumsumOp and CumprodOp are for compatibility with old version,
+# just in case unpickling a theano function with old Ops.
+class CumsumOp(theano.Op):
+    __props__ = ("axis",)
+
+    def __new__(typ, *args, **kwargs):
+        obj = object.__new__(CumOp, *args, **kwargs)
+        obj.mode = 'add'
+        return obj
+
+
+class CumprodOp(theano.Op):
+    __props__ = ("axis",)
+
+    def __new__(typ, *args, **kwargs):
+        obj = object.__new__(CumOp, *args, **kwargs)
+        obj.mode = 'mul'
+        return obj
+
+
 class DiffOp(theano.Op):
     # See function diff for docstring
 
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -248,6 +248,8 @@ class CumOp(theano.Op):
     __props__ = ("axis", "mode")
 
     def __init__(self, axis=None, mode='add'):
+        if mode not in ('add', 'mul'):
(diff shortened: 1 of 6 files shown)


theano/tensor/extra_ops.py
```diff
@@ -264,20 +265,37 @@ def make_node(self, x):
...
+        noimpl = NotImplementedError(
+            'CumOp: unknown gradient for mode %s' % self.mode)
 
         if self.axis is None:
-            return [cumsum(gi[::-1])[::-1].reshape(inputs[0].shape)]
+            if self.mode == 'add':
+                return [cumsum(gi[::-1])[::-1].reshape(x.shape)]
+            elif self.mode == 'mul':
+                return [cumsum(
+                    (fx * gi)[::-1])[::-1].reshape(x.shape) / x]
```
nouiz: Move the creating of fx on the line above.

theano/tensor/extra_ops.py
```diff
@@ -264,20 +265,37 @@ def make_node(self, x):
...
-        # We need to reverse the gradients along ``self.axis``,
-        #  compute cumsum, then reverse again
         reverse_slicing = [slice(None, None, None)] * gi.ndim
         reverse_slicing[self.axis] = slice(None, None, -1)
         reverse_slicing = tuple(reverse_slicing)
-        return [cumsum(gi[reverse_slicing], self.axis)[reverse_slicing]]
+        if self.mode == 'add':
+            return [cumsum(gi[reverse_slicing], self.axis)[reverse_slicing]]
+        elif self.mode == 'mul':
+            # We need to reverse the gradients along ``self.axis``,
```
nouiz: Why did you move this comment?

theano/tensor/extra_ops.py
```diff
@@ -264,20 +265,37 @@ def make_node(self, x):
...
+            #  compute cumsum, then reverse again
+            return [cumsum(
+                (fx * gi)[reverse_slicing], self.axis)[reverse_slicing] / x]
```
nouiz: Can the grad of cumprod be writen as a cumprod? If so, this could make the graph simpler. I didn't look at the math.

khaotik: The only improvement I can figure out is to introduce a "reverse" mode for CumOp:
```
cumop(vec, mode='add', reverse='True')
# same as reverse(cumop(reverse(vec), mode='add'))
```
This requires modifying `c_code`. The reverse slicing Ops can be removed this way. Still, there must be coupling cumsum and cumprod, and the buggy(#5197) division remains.

nouiz:
I think all is good. Thank!

Issue #5316, opened 2016-12-02
Optimization failure due to: local_reshape_dimshuffle


ysjakking:
**I encountered this error with the develop version of theano on Ubuntu 14：**

**Theano version** 0.9.0dev4.dev-9cb9ff4db5bf4720c47613348c8a143c3aeaa772
NumPy version 1.11.2
NumPy relaxed strides checking option: False
NumPy is installed in /usr/local/lib/python2.7/dist-packages/numpy
Python version 2.7.6 (default, Jun 22 2015, 17:58:13) [GCC 4.8.2]
nose version 1.3.7
/home/Theano-master/Theano/theano/misc/pycuda_init.py:35: UserWarning: PyCUDA import failed in theano.misc.pycuda_init
  warnings.warn("PyCUDA import failed in theano.misc.pycuda_init")

**ERROR information:**
ERROR (theano.gof.opt): Optimization failure due to: local_reshape_dimshuffle
ERROR (theano.gof.opt): node: Reshape{1}(InplaceDimShuffle{0,x}.0, <TensorType(int64, vector)>)
ERROR (theano.gof.opt): TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "/home/Theano-master/Theano/theano/gof/opt.py", line 1964, in process_node
    replacements = lopt.transform(node)
  File "/home/Theano-master/Theano/theano/tensor/opt_uncanonicalize.py", line 136, in local_reshape_dimshuffle
    return [T.reshape(input_.owner.inputs[0], node.inputs[1])]
  File "/home/Theano-master/Theano/theano/tensor/basic.py", line 4714, in reshape
    "argument to 'reshape' to avoid this problem." % newshape)
ValueError: The length of the provided shape (<TensorType(int64, vector)>) cannot be automatically determined, so Theano is not able to know what the number of dimensions of the reshaped variable will be. You can provide the 'ndim' keyword argument to 'reshape' to avoid this problem.

**My script for .theanorc:**
[global]
floatX = float32
device = gpu1
exception_verbosity=high
compute_test_value=warn
optimizer= fast_run
allow_gc=False
[nvcc]
fastmath = False
[scan]
allow_gc=False
[lib]
cnmem=0.3
[blas]
ldflags= -lblas -lgfortan

**Information on blas:**
>>> import numpy
>>> numpy.distutils.__config__.show()
lapack_opt_info:
    libraries = ['openblas', 'openblas']
[...]

PR #5318, opened 2016-12-02
Fix opt crash

nouiz:
fix gh-5316


slefrancois:
jenkins test this

lamblin:
Can you add [this code](https://github.com/Theano/Theano/issues/5316#issuecomment-264409746) as a unit test?

Issue #5191, opened 2016-11-07
Port cusolver GpuSolve to new back-end


nouiz:
Mostly, port the PR: https://github.com/Theano/Theano/pull/4917 to the new back-end.

tfjgeorge:
@slefrancois Here is a broad view of what I have started working on.
- port the PR #4917 to the gpuarray backend
  - create new `gpuarray/linalg/solve.py` with existing code from `sandbox/cuda/cusolver.py`
  - adapt code so that it works with the new backend
  - update the tests
- modify the switch that replaces the op with the gpu_op when appropriate
- further work : adapt the gpu op to implement the special matrix cases (lower triangular, ... , there are corresponding methods in cusolver also implemented in pycuda) (maybe in a future CCW ?)
- add a gradient method (maybe in a future CCW ?)

PR #5320, opened 2016-12-02
Gpu op for solve using cusolver (CCW #5191)

tfjgeorge:
@slefrancois

### DIFF
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -6,7 +6,10 @@ import theano
 from theano import Op
 from theano.gpuarray import basic_ops, GpuArrayType
 
-from pygpu import gpuarray
+try:
+    from pygpu import gpuarray
+except ImportError:
+    pass
 
 cusolver_available = False
 try:
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -10,7 +10,7 @@ from pygpu import gpuarray
 
 cusolver_available = False
 try:
-    from scikits.cuda import cusolver
+    from skcuda import cusolver
     cusolver_available = True
 except (ImportError, OSError, RuntimeError, pkg_resources.DistributionNotFound):
     pass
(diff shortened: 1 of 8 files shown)


theano/gpuarray/opt.py
```diff
@@ -1884,6 +1886,14 @@ def typebuild(dtype, broadcastable, context_name=context_name):
 def local_gpu_maxandargmax(op, context_name, inputs, outputs):
     return GpuMaxAndArgmax(op.get_params(None))
 
+
+# solve
+@register_opt('fast_compile')
```
slefrancois: Put a "if cusolver_available" (import it from gpuarray.linalg) so that the op_lifter is not applied if the optional package is missing.

@nouiz , @abergeron , do we prefer to have the op_lifter in gpuarray.opt or the op file itself?

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
There is an[ autodoc error](http://darjeeling.iro.umontreal.ca:8080/job/Theano%20PR%20multijob/675/console): can't import gpuarray.basic_ops

aam-at:
To avoid autodoc error, I believe you should use protected import for pygpu as done in other files.

theano/gpuarray/linalg.py
```diff
@@ -0,0 +1,173 @@
+from __future__ import absolute_import, division, print_function
+
+import pkg_resources
+import theano
+
+from theano import Op
+from theano.gpuarray import basic_ops, GpuArrayType
+
+try:
+    import pygpu
```
slefrancois: Not needed due to "from pygpu import...", it generates a flake8 error.

tfjgeorge:
@slefrancois just to let you know that I fixed the import for flake8 but did not add a new commit, instead I amended the previous commit

slefrancois:
jenkins test this

theano/gpuarray/opt.py
```diff
@@ -1884,6 +1886,16 @@ def typebuild(dtype, broadcastable, context_name=context_name):
...
+@op_lifter([slinalg.Solve])
+@register_opt2([theano.tensor.slinalg.Solve], 'fast_compile')
+def local_gpu_solve(op, context_name, inputs, outputs):
+    if not cusolver_available:
+        return
+    return GpuCusolverSolve()
```
nouiz: small comment, building new object in Python is costly, so this would be faster if one `gpu_cusolver_solve = GpuCusolverSolve()` is created in the file that create this op and reused here. So this will reuse always the same instance.

This do not block this PR.

slefrancois:
Jenkins failure is due to master, not this PR.

nouiz:
jenkins failure is not related and have a PR that fix it. @slefrancois agree with this and and I, so merging.

thanks

PR #5321, opened 2016-12-02
unstable commit-fixed perform on gpujoin

### LINKED ISSUE #5253: Make join work inplace on the first input

ReyhaneAskari:
Hi @nouiz ,
fix #5253 , I tried to apply the same logic for GPU. I fixed the perform function and it seems to work as expected but I can not do the test as before. The reason is that we were checking if the same object is passed to the output. Since the numerical values are converted to ```<type 'pygpu.gpuarray.GpuArray'>```, it can not be compered with a ```numpy array```.  I tried to to give the input as a gpuarray by defining it as a ```shared``` variable but its type is ```<class 'theano.gpuarray.type.GpuArraySharedVariable'>```. I tried to get the ```container``` of the shared variable but : 
```
-> x.container.type 
   GpuArrayType<None>(float32, (False,))
```

How do you think we should test it?

Thanks

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1349,23 +1349,23 @@ class GpuJoin(HideC, Join):
                     %(fail)s
                 }
             }
-        int tensors_lens_sum;
-        if(%(view)s != -1){
-        tensors_lens_sum = 0""" % locals()
 
-        for inp in tensors:
-            code += """ + PyGpuArray_DIM(%(inp)s, axis)""" % locals()
-        code += """;\n
-        tensors_lens_sum -= PyGpuArray_DIM(%(non_empty_tensor)s, axis);
-        }
-        if(%(view)s != -1 && tensors_lens_sum == 0){
-            Py_INCREF(%(non_empty_tensor)s);
-            %(out)s = %(non_empty_tensor)s;
-        }
-        else{
-            %(out)s = pygpu_concatenate(als, %(n)s, axis,
-                                        %(restype)s, (PyObject *)&PyGpuArrayType,
-                                        %(ctx)s);
+            int tensors_lens_sum;
+            if(%(view)s != -1) {
+                tensors_lens_sum = 0;
+                for(int i=0; i < %(n)s; i++){
+                    tensors_lens_sum += als[i]->dimensions[axis];
+                }
+                tensors_lens_sum -= PyGpuArray_DIM(%(non_empty_tensor)s, axis);
+            }
+
+            if(%(view)s != -1 && tensors_lens_sum == 0) {
+                Py_INCREF(%(non_empty_tensor)s);
+                %(out)s = %(non_empty_tensor)s;
+            }else{
+                %(out)s = pygpu_concatenate(als, %(n)s, axis,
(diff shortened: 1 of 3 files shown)


theano/gpuarray/basic_ops.py
```diff
@@ -1245,8 +1245,27 @@ class GpuJoin(HideC, Join):
...
+    # def __str__(self):
+    #     if self.view == -1:
+    #         return "Join"
+    #     else:
+    #         return super(Join, self).__str__()
+
+    # def __setstate__(self, d):
+    #     self.__dict__.update(d)
+    #     if not hasattr(self, "view"):
+    #         self.view = -1
```
nouiz: Remove those 2 commanted method. They are inherited. Go modify the parent \__str__ method to use self.__class__.__name__ instead of "Join". This will use the good string Join or GpuJoin depending of the class.

ReyhaneAskari: I removed ```__setstate__()``` but we can't remove  ```__str__()``` as  ```__props__```  has been defined and overrides the inherited  ```__str__()``` from "Join". Thus, I had to define it in GpuJoin as well. Also, I changed 

```python
return super(Join, self).__str__()
``` 
to 
```python
return "%s{%s}" % (
                self.__class__.__name__,
                ", ".join("%s=%r" % (p, getattr(self, p))
                          for p in self.__props__))
```

because this is what was being returned in ```__props__```.

theano/gpuarray/basic_ops.py
```diff
@@ -1245,8 +1245,25 @@ class GpuJoin(HideC, Join):
...
     params_type = gpu_context_type
 
+    def __init__(self, view=-1):
+        self.view = view
+        if view != -1:
+            # since the first input is always the axis, the tensors
+            # start from index 1.
+            self.view_map = {0: [1 + view]}
+
+    def __str__(self):
```
nouiz: Don't forget to remove this method.

theano/tensor/basic.py
```diff
@@ -3897,9 +3897,12 @@ def __init__(self, view=-1):
...
     def __str__(self):
         if self.view == -1:
-            return "Join"
+            return self.__class__.__name__
         else:
-            return super(Join, self).__str__()
+            return "%s{%s}" % (
+                self.__class__.__name__,
+                ", ".join("%s=%r" % (p, getattr(self, p))
+                          for p in self.__props__))
```
nouiz: Why calling super don't work?

ReyhaneAskari: Super returns something like this :

```<theano.tensor.basic.Join object at 0x7f523612f2d0>```

but  ```__props__``` returns ```Join{view=0}```.

PR #5322, opened 2016-12-03
Adding Copy to SparseVariable

renewang:
There's a minor problem in debug mode when checking input's viewmap ( _check_inputs when executing compiled function) where it will examine each output generated from CSMProperties and checking if the memory is shared and if not shared then issue the Optimization warning:   
   
The problem is because the type of outputs in CSMProperties Op is TensorType and its  may_share_memory (see  theano/tensor/theano/tensor/type.py@7215c905f96a9d20) will return False if both are not numpy.ndarray. However, this block of code (see theano/compile/debugmode.py:853@23e43b1b9e36e6beb) actually pass one ndarray (because they are properties of sparse array) and one sparse array. Checking may_share_memory in this way might have problem in other scenarios.   
    
Could it be possible to call SpareType.may_share_memory for preventing other potential problems in advance?   
    
Anyway, since the optimization warning does not occur in other modes, I still check in the solution as advised above with a test case added. Thanks for any suggestions in advance.

### DIFF
--- a/theano/sparse/basic.py
+++ b/theano/sparse/basic.py
@@ -309,6 +309,16 @@ class _sparse_py_operators:
             ret = get_item_2d(self, args)
         return ret
 
+    # COPYING
+    def copy(self, name=None):
+        """Return a symbolic copy and optionally assign a name.
+
+        Does not copy the tags.
+        """
+        copied_variable = CSM(self.format)(*csm_properties(self))
+        copied_variable.name = name
+        return copied_variable
+
 
 class SparseVariable(_sparse_py_operators, gof.Variable):
     dtype = property(lambda self: self.type.dtype)
--- a/theano/sparse/tests/test_basic.py
+++ b/theano/sparse/tests/test_basic.py
@@ -3185,6 +3185,30 @@ class SamplingDotTester(utt.InferShapeTester):
         verify_grad_sparse(_helper, self.a[:2])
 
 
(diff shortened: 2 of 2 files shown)


nouiz:
Thanks for the report and the PR and sorry for the long delay. SparseVariable isn't one of our current priority.

I understand the problem you explain. I think this PR do not contain a fix for this problem. Can you confirm? I see that the PR add a test. But the test pass and I don't see a fix. So I don't understand what is happening.

I think the fix could be to call theano.misc.may_share_memory.may_share_memory(out_var, in_var, type=var.type)

We need to add the new parameter type to that function. If it don't find the right type to call may_share_memory(), it should fall back to type.may_share_memory instead of raising an error.

What do you think of that?

renewang:
Hi,  execute for the late reply due to having no internet access and the need to refresh what the problem is

Yes, #5322  didn't include debugmode problem I described above.  I couldn't comment what fix should be because I'm not familiar with how theano.misc.may_share_memory.may_share_memory works neither the complicated code flow within theano structure. That's the reason I left this problem unfixed in fear of breaking other things.
 
By intuition, maybe write a special or magic method (such as __add__, __radd__ to ensure the commutativity) which will check its own properties instead of using static dispatch function relying on the type of first argument? But this is just a casual guess. As I said, I'm completely a novice for theano. Hopefully, you already find the best solution for this problem. 

Thanks for reply!

Issue #4387, opened 2016-04-17
Rop for Pool


AndreasMadsen:
I'm trying to implement a transformation invariance regularizer as described in http://yann.lecun.com/exdb/publis/pdf/simard-00.pdf section 3. Basically for a classification model with a scale transformation `s(x, alpha) = x * alpha`, it is just:

<img src="https://cloud.githubusercontent.com/assets/505333/14587583/9559395e-04b5-11e6-91b4-9c33a39ae2af.png" width="350">

I have implemented this using a scan operator:

``` python
def loop(i, x, p, t):
  p_class_t = p[i, t[i]]
  return T.dot(
    T.flatten(T.grad(p_class_t, x)[i, :, :, :]),
    T.flatten(x[i, :, :, :])
  )

jacobi_dot_x, _ = theano.scan(
  loop,
  non_sequences=[input_var, prediction, target_var],
  sequences=T.arange(self.input_var.shape[0])
)

regularizer = jacobi_dot_x.sum()
```

however it is horrifyingly slow.

As I understand it I should be able to implement this using an `Rop`:

``` python
regularizer = T.Rop(p[:, t], x, x).sum()
```

I have tested this on a simple model and the performance gain was huge. However in the full CNN model I have a max pooling layer using `theano.tensor.signal.pool.Pool` and I get the following error:

```
NotImplementedError: Pool{ds=(1, 4), ignore_border=True, st=(1, 4), padding=(0, 0), mode='max'} of class Pool does not implement R_op. If this is a theano op, write to the theano-dev mailing list for assistance. If it is your own op, implement the R_op method.
```

Is it difficult to implement the `Rop` operation for `Pool` operation? I took a look at the code, but had a very hard time understanding the general architecture (e.q. `make_nodes`) as well as the C API.

nouiz:
We have this in a PR to an older version of Theano. You can use it directly
if you want:

https://github.com/Theano/Theano/pull/3587

AndreasMadsen:
Wow that is amazing. But I think the `DownsampleFactorMaxRop` op is missing an gradient implementation. I'm getting the following error:

```
AttributeError: 'DownsampleFactorMaxRop' object has no attribute 'grad'
```

My _port_ of Theano/Theano#3587:  https://github.com/AndreasMadsen/Theano/commit/d5dfbe1d124cc155151b10f0076bd8d53a4308ff

nouiz:
What do you try to do? You build your graph with the max pool op, then you
[...]

Issue #4568, opened 2016-05-31
Incorrect subgradient of theano.tensor.max


TobyPDE:
TL;DR The calculated gradient of the max-operation is not always in the subdifferential.

**Idea**
Let **x** _= (x_1,...,x_N)_ be a real vector and consider the function 
_f(x_1,...,x_N) = max(x_1,...,x_N) = max(g_1(x_1,...,x_N),...,g_N(x_1,...,x_N))_ 
where _g_i(x_1,...,x_N) = x_i_. Basic math tells us that the subdifferential of _f_ at any fixed point **x** is the convex hull of the gradients of the active projection functions _g_i_ [1]. For _i in {1,...,N}_ the gradient of _g_i_ is the standard basis vector _**e**_i_. 

**Example**
Let _N = 2_, then the subdifferential of _f(x_1,x_2) = max(x_1, x_2)_ at a point **x** with _x_1 = x_2_ is the line segment _{a * (1, 0) + (1 - a) * (0, 1) : a in [0, 1]}_. 

**Issue**
Evaluating the gradient of _max(x_1, x_2)_ at a point **x** with _x_1 = x_2_ returns the gradient _(1, 1)_ which is not in the subdifferential. 

```python
import theano
import theano.tensor as T
import numpy as np

X = T.vector()
z = T.max(X)
Df = theano.function(
    inputs=[X],
    outputs=T.grad(z, X)
)

x = np.array([1, 1]).astype(dtype=theano.config.floatX)
gradient = Df(x)
print(gradient) # -> [1. 1.]
```

**Proposed solution**
[Here](https://github.com/Theano/Theano/blob/baf5227a93b7ff1d8b8aa59010591f86ecab0d43/theano/tensor/signal/pool.py#L798), we need to take some convex combination of the corresponding entries of `gz`. A natural choice would be to take the center of the subdifferential, i.e. divide `gz` by the number of active arguments. 

_I am not entirely sure if my reasoning is correct. So, please let's discuss._

[1] See 3.4 https://see.stanford.edu/materials/lsocoee364b/01-subgradients_notes.pdf

lamblin:
Since max is convex, I think what you say about the subgradient seems true, and even in general, it may be a good way of handling such ties.
However, for us, it did not seem to be worth the effort, potential slowdowns, and potential suprising behaviours in other cases. The gradient wrt each of the component is still consistent now with the numerical one, at least in one direction, and ties are rare, so it did not seem to be an issue in practice.

[...]

PR #5323, opened 2016-12-04
Pooling rop

aam-at:
Continuation of #4781 and all work which was done before.

I updated max pool rop implementation which now supports arbitrary ndim for pure python and cpu openmp implementation. I also dropped support for old backend and instead added support only for new backend.

Also as a sidenote, there is a large overlap between max pooling and max pooling rop. Max pooling reduces to max pooling rop if eval point is equal to the input. For now, I didn't do any refactoring and just duplicated code for max pooling with minor changes for rop.

fix gh-4387

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -49,7 +49,7 @@ from .blas import (gpu_dot22, GpuGemm, GpuGer, GpuGemmBatch,
                    gpugemv_no_inplace, gpugemv_inplace,
                    GpuCorrMM, GpuCorrMM_gradInputs, GpuCorrMM_gradWeights,
                    GpuCorr3dMM, GpuCorr3dMM_gradInputs, GpuCorr3dMM_gradWeights)
-from .pool import (GpuPool, GpuMaxPoolGrad, GpuAveragePoolGrad,
+from .pool import (GpuPool, GpuMaxPoolGrad, GpuAveragePoolGrad, GpuMaxPoolRop,
                    GpuDownsampleFactorMaxGradGrad)
 from .blocksparse import (GpuSparseBlockGemv, GpuSparseBlockOuter,
                           gpu_sparse_block_outer,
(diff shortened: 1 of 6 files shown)


aam-at:
Travis fails on pooling rop - No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself
When I run tests on locally, I don't have any problems though.

lamblin:
Jenkins test this please

lamblin:
I may be mistaken, but I think this is the same mathematical operation as `DownsampleFactorMaxGradGrad`, which also reduces to max pooling if `ggx` is equal to `x`.
Could you check if I missed something and there is a difference?

aam-at:
Correct me if I am wrong: the output shape of `DownsampleFactorMaxGradGrad` is the same as `z`. And max pool rop will be equivalent to it only if `ws` and `stride` are both 1. `GradGrad` uses maximum in output `z` and sums ups corresponding elements in `gx` (overlaps are possible). Rop uses maximum in the input and find corresponding elements in `ex`.

Relevant part from gpu kernel for pool2d_grad_grad:
``` c
if (z[index] == x_slice[h * width + w]) {
  gradient += gx_slice[h * width + w];
}
```
and pool2d_rop:
``` c
if (x_slice[(d*height + h)*width + w] > maxval) {
  maxval = x_slice[(d*height + h)*width + w];
  collector = ex_slice[(d*height + h)*width + w];
}
```

lamblin:
OK, so the difference is that rop recomputes the maximum in the input in order to select the element from `ex`, whereas gradgrad checks the positions where the input is equal to its (already precomputed) max `z`.
In gradgrad, `z`'s size is the size of `max_pool(x)`, `ggx` has the same shape as `x`, and the output of gradgrad has the same shape as `z`.
If I understand correctly, the rop's output has the same shape as `max_pool(x)` as well, right?

So the mathematical expressions should be equivalent (at least when the position of the max is not ambiguous), but the way to arrive to it are different (the max is pre-computed vs. recomputed). I really don't know which approach is the best in general. We could have both implementations, but then both would need to be available for each of the operations (MaxPoolGrad.grad and MaxPool.Rop), maybe an optimization could be used to switch between them.

aam-at:
The output shape of max pool rop is the same as `x`. In this sense rop operation is similar to `MaxPoolGrad`. And the computation itself is similar to `MaxPool`. Compare above rop snippet with the  computation for max pool kernel:
``` c
if (x_slice[h*width + w] > maxval) {
  maxval = x_slice[h*width + w];
}
```
So max pool rop selects values in `ex` which correspond to the maximum in `x`. Output will have block like structure of repeated values from `ex`.

I think we can't reuse any other operations for general case. For corner cases though some ops are equivalent but it is not straightforward to generalize.

aam-at:
I deleted my previous comment as it was incorrect. You are right that `MaxPoolRop` returns shape of size `z`. And yes, I also think that when there are no ambiguity for max value both operation should return same  result.

nouiz:
Currently, we aren't consistent how we handle ambiguous max value. We
should select only one in all cases as cudnn/caffe does.

Do you know how the new op handle the ambiguous cases?

aam-at:
This new op `MaxPoolRop` chooses only one maximum value. `MaxPoolGradGrad` accumulates them. With the exception of this, both operations are mathematically equivalent. Implementation-wise, `MaxPoolGradGrad` requires pre-computed maximum values. So if understand @lamblin correctly, we can optimization which chooses between two implementations whenever pre-computed `z` is available.

aam-at:
What else should I finish, so you can accept this pull request?

lamblin:
Sorry about the delay, I just came back from the holidays.
We decided to keep a larger refactorization (and harmonization) for later, in particular concerning the behaviour when several values tie for the maximum.

For now, the remaining issue is that `Pool.R_op()` and `MaxPoolGrad.grad()` return different Ops, even though it is the same operation and they should be merged if needed.

Would it be possible for you to make `Pool.R_op()` return an instance of `MaxPoolGradGrad`? We are still interested in your implementation that is in `MaxPoolRop`, and we are considering making it the default version in the refactorization mentioned above, so you can leave it as part of the PR.

Thanks,

aam-at:
I updated `Rop` which return `MaxPoolGradGrad` for now.

If we decide to implement case 3 from #5404, I can work on it. I am not sure what you mean by a larger refactorization but changes to c++ code for cpu and gpu should be fairly easy. For example for gpu `MaxPoolGradGrad`, we can just break from innermost cycle once the corresponding maximum is found:
```c++
bool should_stop = false;
for (ga_size h=hstart; h < hend && !should_stop; ++h) {
  for (ga_size w=wstart; w < wend && !should_stop; ++w) {
  // maximum in the region
  if (z[index] == x_slice[h * width + w]) {
    gradient = gx_slice[h * width + w];
    should_stop = true;
  }
}
```

lamblin:
Jenkins test this please

nouiz:
merging as @lamblin approved. We can continue the discussion in the issue.

thanks

nomader:
Similar analysis: https://github.com/Theano/Theano/issues/4568

PR #5326, opened 2016-12-05
Py3.5 __matmul__ operator

khaotik:
Mini syntax candy for python 3.5.

```python
x, b = theano.tensor.vectors('xb')
W = theano.tensor.matrix('W')
y = W @ x + b
```


nouiz:
Thanks. I'm not sure this is the good way to implement it. I'm not sure it is exactly the same as dot.

Also this would request tests. Can you add some tests in the file theano/tensor/tests/test_var.py? Do tests with scalar/matrix, vector/matrix, matrix/matrix, 3dtensor/matrix. I think in one of them there is a difference compared to numpy.matmul.

thanks

khaotik:
Turns out, it isn't as simple as I thought.

numpy `matmul` works like a "elemwise matrix multiplication".

```python
a = np.random.rand(10,6,1,5,4)
b = np.random.rand(1,6,20,4,3)
c = a @ b
c.shape # (10, 6, 20, 5, 3)
```

I don't see an elegant way to perform the above operation. It's possible to hack with `dimshuffle`, `tile`, `reshape`, and `batched_tensordot`, however this is not efficient.

Currently this PR is broken, because `batched_tensordot` can only use one axis as batch and it doesn't broadcast.

nouiz:
There is numpy tests for matmul: https://github.com/numpy/numpy/blob/master/numpy/core/tests/test_multiarray.py#L4938

Just in case it help. NumPy c code call einsum that we don't have in Theano when ndim !=2 for both input. Both otherwise, I think it isn't super efficient.

There is this syntax error in travis:
```
  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_basic.py", line 2775
    def test_matmul():
```

I would accept a PR that only implement a subset, if good error are raised for the not supported case if the most frequent case are supported (mostly with mix of vector/matrix).

thanks for looking into this.

emillynge:
Hi, i have a very old PR on thius exact feature [Add __matmul__ method to tensor_py_operators](https://github.com/Theano/Theano/pull/3769).

I also stumbled on the complexity of the fully featured matmul method as proposed by PEP465 and implemented in numpy.

Feel free to copy paste tests and implementation from my PR :)

khaotik:
@emillynge Thanks for the bump! Your implementation gets the job done. However I think it has performance issues. Theano's `scan` is hard coded and strictly sequential. This would cause heavy overhead if you perform matmul with shape like `[10000, 2, 3, 4]` and `[2, 4, 5]`.

PR #5331, opened 2016-12-07
Sphinx doc/conf: latex_font_size is deprecated

gvtulder:
Travis fails with this error:
```
$ if [[ $DOC == "1" ]]; then python doc/scripts/docgen.py --nopdf --check; fi
Running Sphinx v1.5

Warning, treated as error:
WARNING: latex_font_size is deprecated. Use latex_elements['pointsize'] instead.
```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -222,11 +222,16 @@ def linkcode_resolve(domain, info):
 # Options for LaTeX output
 # ------------------------
 
-# The paper size ('letter' or 'a4').
-#latex_paper_size = 'letter'
+latex_elements = {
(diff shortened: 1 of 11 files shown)


gvtulder:
For what it's worth: the Sphinx manual says that the `latex_elements` syntax has been supported [since version 0.5 (November 2008)](http://www.sphinx-doc.org/en/1.5/config.html#confval-latex_elements), so if it doesn't work it's probably time to update anyway. :)

slefrancois:
I just upgraded sphinx on the webserver from 1.4.6 to 1.5. In any case elements should work as pointed out.

PR #5332, opened 2016-12-09
Fix warning/detection on tx1. I suppose it is an update to there comp…

nouiz:
…iler, as from memory, it working in the past.

Now accept line with "-mabi=" as indicating the compiler.


nouiz:
travis failure are due to sphinx update. Another PR fixed it. As it is merged, I close and reopen to retrigger travis.

PR #5335, opened 2016-12-12
[WIP] Use __call__ instead of make_node to avoid bypassing test values

### LINKED ISSUE #5003: cuDNN batch normalization gradient breaks test values
The error only occurs when compute_test_value is enabled. Compiling a function to get the output works fine. 

Example minimal test case:

```
import theano
import theano.tensor as T
from theano.sandbox.cuda import dnn
theano.config.compute_test_value = "raise"
inputs = T.zeros((2, 3))
gamma = T.zeros((1, 3))
beta = T.zeros((1, 3))
res, mean, var = dnn.dnn_batch_normalization_train(inputs, gamma, beta)
g = T.grad(res.sum(), [inputs, gamma, beta])
```

Results in the error:

```
ValueError: Cannot compute test value: input 3 (GpuDnnBatchNorm{mode='per-activation', epsilon=0.0001}.1) of Op GpuDnnBatchNormGrad{mode='per-activation', epsilon=0.0001}(GpuContiguous.0, GpuFromHost.0, GpuContiguous.0, GpuDnnBatchNorm{mode='per-activation', epsilon=0.0001}.1, GpuDnnBatchNorm{mode='per-activation', epsilon=0.0001}.2) missing default value.  
```

lamblin:
This should fix #5003.
I'll also add some tests.

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1597,7 +1597,7 @@ class GpuDnnSoftmax(GpuDnnSoftmaxBase):
     def grad(self, inp, grads):
         x, = inp
         g_sm, = grads
-        sm = self.make_node(x).outputs[0]
+        sm = self(x)
         return [GpuDnnSoftmaxGrad(
                 self.algo,
                 self.mode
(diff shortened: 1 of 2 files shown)


nouiz:
ok to merge. I'm not convinced it need tests, so I'll merge now as I'm sure it is good. There is already existing tests. I don't think we should add tests for all op compute test value. This would be too time consuming.

Issue #4519, opened 2016-05-16
Good broadcasting tutorial


nouiz:
This page is not visualized correctly: http://deeplearning.net/software/theano/sandbox/broadcasting.html

We should not just fix the display, but make a good tutorial with picture and exercices.

TODO:
- [ ] delete the doc/sandbox/broadcasting.txt file. 
- [ ] move http://deeplearning.net/software/theano_versions/dev/library/tensor/basic.html#broadcasting-in-theano-vs-numpy to the Theano tutorial in its own page. 
- [ ] Add in this new page, as text, an example of (F, F) + (F, T).

lamblin:
There is some information and a nice figure at http://deeplearning.net/software/theano_versions/dev/library/tensor/basic.html#broadcasting-in-theano-vs-numpy
We should probably move it outside of the "library" section and into the tutorial part, with an example like @nouiz suggested. There is not much more in the sandbox/broadcasting page.

PR #5343, opened 2016-12-16
Broadcasting tutorial

ChihebTrabelsi:
This pull request is related to the issue #4519


doc/tutorial/broadcasting.txt
```diff
@@ -0,0 +1,70 @@
...
+
+============
+Broadcasting
+============
+
+Broadcasting is a mechanism which allows tensors with
+different numbers of dimensions to be added or multiplied
+together by (virtually) replicating the smaller tensor along
+the dimensions that it is lacking.
+
```
lamblin: Let's explain the general concept of broadcasting here first, before describing how it is implemented in Theano, like it was done in the original documentation (the one in basic.txt, not in sandbox/).
The one in sandbox was not that clear, that's why we wanted to remove it, but maybe we can work some of its content into the new tutorial.

For that, I think the easiest is to include the whole content of the original documentation, and after that, we can expand by adding the Theano code corresponding to the figure (with the same values if possible), then the code to add a column.

We can also add some of the information from sandbox/broadcasting.txt, but it is too verbose, and somehow redundant with the other one.

doc/sandbox/broadcasting.txt
```diff
@@ -4,26 +4,7 @@
 Broadcasting
 ============
 
-The following may go either in:
-a) numpy refresher.
-b) more details of broadcasting in the types section.
+You can find more information about Broadcasting in the :ref:`tutbroadcasting` tutorial.
```
lamblin: You can remove that file completely.

doc/tutorial/broadcasting.txt
```diff
@@ -0,0 +1,70 @@
...
+.. figure:: bcast.png
+
+Similarly, the broadcastable pattern for a column is ``(False, True)``: this means the shape must be like ``(m, 1)``, therefore adding a column to a matrix will add that column to each column of the matrix. Several Ops, such as ``DimShuffle``, can add or remove broadcastable dimensions.
+
+The following code illustrates how rows and columns are broadcasted in order to perform an addition operation with a matrix:
+
+>>> import numpy as np
+>>> import theano
+>>> import theano.tensor as T
```
lamblin: Oh OK, I understand.
`docgen.py --test` will try to run the code, and needs to know about the imports for that.

I think the way of having the imports for the tests, but not displaying them, would be to add the following at the beginning of the file:
```
.. testsetup::

   import numpy as np
   import theano
   import theano.tensor as T
```

I thought it had been defined globally for all tests, but apparently you need to do it for all new files.
Could you add that?

lamblin:
@ChihebTrabelsi Any progress on the last point?

ChihebTrabelsi:
Done.

Issue #5338, opened 2016-12-13
Performance issue with GPUAMultinomialWOReplacementFromUniform


senarvi:
I requested size argument for `multinomial()` in issue #5292 in order to efficiently get a large number of samples. I also said that `multinomial_wo_replacement()` is faster, so I'm using it currently, but there seems to be another performance issue in the new GpuArray backend.

When I'm using the new backend, processing one mini-batch takes more than 900 ms. I did some profiling and noticed that `theano.gpuarray.multinomial.GPUAMultinomialWOReplacementFromUniform` is consuming 97 % of the processing time.

Using the old backend, one mini-batch takes 90 ms. This is the top four time consumers:
```
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Class name>
  49.5%    49.5%       0.218s       2.91e-03s     C       75      75   theano.sandbox.cuda.basic_ops.GpuElemwise
  27.9%    77.4%       0.123s       6.14e-02s     Py       2       2   theano.scan_module.scan_op.Scan
  10.2%    87.5%       0.045s       2.36e-03s     C       19      19   theano.sandbox.cuda.blas.GpuDot22
   4.3%    91.8%       0.019s       1.88e-02s     C        1       1   theano.sandbox.multinomial.MultinomialWOReplacementFromUniform
```
In this case the sampling is apparently not done on GPU, but the operation `MultinomialWOReplacementFromUniform{int64}(HostFromGpu.0, HostFromGpu.0, ScalarFromTensor.0)` is using variables that are transferred to host memory from GPU. Still it's a lot faster. Or is there a reason why this operation cannot be implemented efficiently on GPU?

lamblin:
There is no reason I know that would make it slow on GPU.
Do you have an example script we could use to try and reproduce the issue?
Also, are you excluding the first call from the timing?

senarvi:
During profiling I have `profiling.ignore_first_call=True`.

That's a huge network but I'll try to make a smaller example.

senarvi:
I can reproduce this performance issue by just looping over `multinomial_wo_replacement()` calls. This is the program:

```python
from time import time
import numpy
import theano
from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
numpy.random.seed(23)
probs = numpy.random.randn(1, 100000)
probs /= probs.sum()
probs = theano.shared(probs.astype(theano.config.floatX), 'probs')
rng = RandomStreams(23)
[...]

PR #5346, opened 2016-12-16
[ENH] Multinomial without replacement

nouiz:
fix memory leak and speed up when there is few multinomials as in gh-5338.


theano/gpuarray/multinomial.py
```diff
@@ -450,12 +464,13 @@ def c_code(self, node, name, inp, outputs, sub):
         }
         if(%(sync)d)
             GpuArray_sync(&(out->ga));
+        Py_DECREF(pvals_copy);
```
abergeron: Good catch, but there are other exit points before this where we should free it too.

nouiz: I amended the last commit.

abergeron: I don't see the new locations.

Issue #5334, opened 2016-12-12
BUG IN FGRAPH.REPLACE OR A LISTENER


pcyin:
Hi,

I am using the dev version, and encountered the following error message when compiling a graph, which seems to be a bug.

```
<<!! BUG IN FGRAPH.REPLACE OR A LISTENER !!>> <type 'exceptions.TypeError'> ('The type of the replacement must be compatible with the type of the original Variable.', InplaceDimShuffle{1}.0, Subtensor{int64}.0, TensorType(float32, vector), TensorType(float32, row), 'local_dimshuffle_subtensor') local_dimshuffle_subtensor
ERROR (theano.gof.opt): Optimization failure due to: local_dimshuffle_subtensor
ERROR:theano.gof.opt:Optimization failure due to: local_dimshuffle_subtensor
ERROR (theano.gof.opt): node: InplaceDimShuffle{1}(Subtensor{int64}.0)
ERROR:theano.gof.opt:node: InplaceDimShuffle{1}(Subtensor{int64}.0)
ERROR (theano.gof.opt): TRACEBACK:
ERROR:theano.gof.opt:TRACEBACK:
ERROR (theano.gof.opt): Traceback (most recent call last):
  File "Theano/theano/gof/opt.py", line 2004, in process_node
    remove=remove)
  File "Theano/theano/gof/toolbox.py", line 391, in replace_all_validate_remove
    chk = fgraph.replace_all_validate(replacements, reason)
  File "Theano/theano/gof/toolbox.py", line 340, in replace_all_validate
    fgraph.replace(r, new_r, reason=reason, verbose=False)
  File "Theano/theano/gof/fg.py", line 481, in replace
    str(reason))
TypeError: ('The type of the replacement must be compatible with the type of the original Variable.', InplaceDimShuffle{1}.0, Subtensor{int64}.0, TensorType(float32, vector), TensorType(float32, row), 'local_dimshuffle_subtensor')

ERROR:theano.gof.opt:Traceback (most recent call last):
  File "Theano/theano/gof/opt.py", line 2004, in process_node
    remove=remove)
  File "Theano/theano/gof/toolbox.py", line 391, in replace_all_validate_remove
    chk = fgraph.replace_all_validate(replacements, reason)
  File "Theano/theano/gof/toolbox.py", line 340, in replace_all_validate
    fgraph.replace(r, new_r, reason=reason, verbose=False)
  File "Theano/theano/gof/fg.py", line 481, in replace
    str(reason))
[...]

PR #5350, opened 2016-12-19
Remove opt warning from local_dimshuffle_subtensor

nouiz:
fix gh-5334.

We where skipping an optimization with a warning when it should have applied.


nouiz:
Don't merge. There is a bug.

nouiz:
I added a test and fixed the bug I taugh about.

PR #5351, opened 2016-12-20
Get rid of redundant copy for GPU "map" style scan

khaotik:
Quick fix for a minor issue identified in #1168. For "map" style scans on GPU with vector output, this eliminates redundant device-device copying. The speedup is around 10%.


theano/scan_module/scan_perform.pyx
```diff
@@ -62,7 +62,7 @@ import copy
 
 
 def get_version():
-    return 0.295
+    return 0.296
```
nouiz: You should also bump the version expected in the file scan_perform_ext.py

theano/scan_module/scan_perform.pyx
```diff
@@ -514,7 +514,8 @@ def perform(
             if i == 0:
                 jout = j+offset_out
                 shape = (store_steps[j],) + output_storage[jout].storage[0].shape
-                if len(output_storage[jout].storage[0].shape) == 0:
+                if output_storage[jout].storage[0].ndim == 0 and \
+                    isinstance(output_storage[jout].storage[0], numpy.ndarray):
```
nouiz: One question that you can't probably answer, why do we modify vector_outs here? This is a permanent change for all futur call to this Theano function. The previous condition was on the current shape (even if constant for all the iteration of the current Theano function call, I'm pretty sure futur Theano function call could have a different shape).

Your new tests on ndim is better then the old on the shapes. But I still don't understand why it change vector_outs here and why it wasn't set correctly around line 616 in this file. Do you have an idea?

nouiz: I discussed with Pierre Luc and we think you can just delete the 2 lines in the original file 517 and 518. Can you do that in this file and in the file scan_op.py for the non cython version? If all tests pass, then it is good.

Those 2 lines dates from 2011 and since that, we re-worked how vector_outs work and those 2 lines aren't needed anymore.

Having those 2 lines don't cause wrong results, that is why it wasn't found before. You did a good finding! thanks.

khaotik: I made some changes to this PR, removed runtime checking and added compile time checking. Did some basic tests and found that only removing the 2 lines will result in errors.

nouiz:
jenkins test this.

I think all is good. I'm not sure I'll merge today as if there is problems, handing them during the holidays isn't great. So probably this will be merged in January, but if jenkins pass, I'm ok to merge this PR.

thanks

nouiz:
The jenkins error can be ignored. They are not related. So we can merge this PR in January.

thanks

nouiz:
Now in January, so merging!

thanks.

PR #5354, opened 2016-12-20
Back-port support of newer pydot to make travis work with this 0.8.X …

nouiz:
…branch

This time I make a PR to make sure travis is fixed instead of doing it on the branch itself.

### DIFF
--- a/theano/printing.py
+++ b/theano/printing.py
@@ -23,19 +23,33 @@ from theano.compile import Function, debugmode, SharedVariable
 from theano.compile.profilemode import ProfileMode
 
 pydot_imported = False
+pydot_imported_msg = ""
 try:
     # pydot-ng is a fork of pydot that is better maintained
     import pydot_ng as pd
     if pd.find_graphviz():
         pydot_imported = True
+    else:
+        pydot_imported_msg = "pydot-ng can't find graphviz. Install graphviz."
 except ImportError:
     try:
         # fall back on pydot if necessary
         import pydot as pd
-        if pd.find_graphviz():
+        if hasattr(pd, 'find_graphviz'):
+            if pd.find_graphviz():
+                pydot_imported = True
+            else:
+                pydot_imported_msg = "pydot can't find graphviz"
+        else:
+            pd.Dot.create(pd.Dot())
             pydot_imported = True
     except ImportError:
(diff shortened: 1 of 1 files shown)


nouiz:
jenkins failure can be ignored. We didn't make it work with the 0.8.X branch. I'm just tring to make travis work again with that branch.

theano/printing.py
```diff
@@ -23,19 +23,33 @@
 from theano.compile.profilemode import ProfileMode
 
 pydot_imported = False
+pydot_imported_msg = ""
 try:
     # pydot-ng is a fork of pydot that is better maintained
     import pydot_ng as pd
     if pd.find_graphviz():
```
lamblin: I think the error is raised here:
```
AttributeError: 'module' object has no attribute 'find_graphviz'
```

.travis.yml
```diff
@@ -38,7 +38,7 @@ install:
   - source activate pyenv
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install pydot; else pip install pydot-ng; fi
   - pip install . --no-deps
```
lamblin: The log says:
```
$ pip install . --no-deps

Processing /home/travis/build/Theano/Theano

  Requirement already satisfied (use --upgrade to upgrade): Theano==0.8.2 from file:///home/travis/build/Theano/Theano in /home/travis/miniconda2/envs/pyenv/lib/python2.7/site-packages
```

So I'll try with `--upgrade`

lamblin: Yep, this seems to actually be an issue with the dev branch as well...

theano/printing.py
lamblin: The issue was that it was still using an older version of Theano.

theano/printing.py
```diff
@@ -23,19 +23,33 @@
...
         pydot_imported = True
-except ImportError:
+    else:
+        pydot_imported_msg = "pydot-ng can't find graphviz. Install graphviz."
+except (ImportError, AttributeError):
```
lamblin: I don't think that's right, AttributeError should not be raised by pydot-ng, I mis-diagnosed that.

lamblin:
Continued in #5360

Issue #4476, opened 2016-05-06
BatchDot on CPU sometimes should copy the inputs


nouiz:
When one of the input don't have one dimensions of size 1.

https://groups.google.com/d/topic/theano-users/fsJhhd6Cwr4/discussion

work around, use cpu_contiguous on the inputs that cause problems.

hantek:
I assume it can be fixed by adding T.extra_ops.cpu_contiguous() in [this](https://github.com/Theano/Theano/blob/master/theano/tensor/blas.py#L2329-L2363) part, right?

The dimshuffle() in computing the grad is the culprit that fails joelthchao's code.

hantek:
Seems it requires editing the c code. Could someone give me some guidelines in how to fix it?

PR #5359, opened 2016-12-21
Fix issue #4476 related to BatchedDot.

notoraptor:
fix gh-4476. It is fixed by modifying contiguousity checking in the c_code of BatchedDot. Now the code uses `(PyArray_IS_C_CONTIGUOUS || PyArray_IS_F_CONTIGUOUS)` to do the checking.

Previously the code checked if every stride of the inputs is relevant (strictly positive and multiple of type size) and if at least one of the strides has the size of the type. I think the bug comes from the latter condition, as an array can be not-contiguous even if one stride has the size of the type (for example, a Fortran array in which the first dimension is contiguous but not the last ones).

Remarks:
* I put the test in `test_basic` even if `BatchedDot` is in `blas.py` because the `batched_dot` tests are also in `test_basic`.
* It seems some other codes check the contiguity using the previous conditions, so there is maybe some similar bugs with other ops. For example, we may have to check the Dot22 op: https://github.com/Theano/Theano/blob/master/theano/tensor/blas.py#L603

@lamblin @nouiz !

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -2149,14 +2149,8 @@ class BatchedDot(Op):
         fail = sub["fail"]
 
         # generate contiguity condition
-        def contiguous(var, ndim):
-            strides = "PyArray_STRIDES(%s)" % var
-            return " && ".join([
-                " && ".join("{strides}[{i}] > 0 && {strides}[{i}] % type_size == 0"
-                            .format(strides=strides, i=i) for i in range(ndim)),
-                "(%s)" % " || ".join("{strides}[{i}] == type_size"
-                                     .format(strides=strides, i=i) for i in range(ndim)),
-            ])
+        def contiguous(var):
+            return "(PyArray_IS_C_CONTIGUOUS(%s) || PyArray_IS_F_CONTIGUOUS(%s))" % (var, var)
 
         x_ndim, y_ndim, z_ndim = node.inputs[0].ndim, node.inputs[1].ndim, node.outputs[0].ndim
 
@@ -2171,7 +2165,7 @@ class BatchedDot(Op):
         z_shape_correct = " && ".join("PyArray_DIMS(%s)[%i] == %s"
                                       % (_z, i, dim) for i, dim in enumerate(z_dims))
         z_shape = ", ".join(z_dims)
-        z_contiguous = contiguous(_z, z_ndim)
+        z_contiguous = contiguous(_z)
         allocate = """
             if (NULL == %(_z)s || !(%(z_shape_correct)s)  || !(%(z_contiguous)s))
             {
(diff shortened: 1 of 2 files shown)


nouiz:
Squash the last commit into the first one to prevent a c code cache version problem in the history.

theano/tensor/tests/test_basic.py
```diff
@@ -2771,6 +2771,30 @@ def test_batched_dot():
...
+        return numpy.arange(size, dtype=floatX).reshape(_shape)
+
+    X = tensor3()
+    W = tensor3()
+    Z = batched_dot(X, W)
+    f = function([X, W], Z)
+
+    reversed_x_container = np_genarray(20, 40, 30)
+    x_container = reversed_x_container.T
+    x = x_container[::1, ::2, ::2]
```
nouiz: Add a test with x_container[::-1, ::2, ::2]

theano/tensor/blas.py
```diff
@@ -2149,14 +2149,8 @@ def c_code(self, node, name, inp, out, sub):
...
-        def contiguous(var, ndim):
-            strides = "PyArray_STRIDES(%s)" % var
-            return " && ".join([
-                " && ".join("{strides}[{i}] > 0 && {strides}[{i}] % type_size == 0"
-                            .format(strides=strides, i=i) for i in range(ndim)),
-                "(%s)" % " || ".join("{strides}[{i}] == type_size"
-                                     .format(strides=strides, i=i) for i in range(ndim)),
-            ])
+        def contiguous(var):
+            return "(PyArray_IS_C_CONTIGUOUS(%s) || PyArray_IS_F_CONTIGUOUS(%s))" % (var, var)
```
nouiz: This is too string. I think you could just change the range(ndim) to range(1, ndim) in the previous code.

notoraptor:
Updated. I put back the original checking and I updated it to make it more precise.

New tests are also updated.

Tests passed:

```
nosetests -vs theano/tensor/tests/test_basic.py:test_batched_dot
nosetests -vs theano/tensor/tests/test_basic.py:test_batched_dot_not_contiguous
nosetests -vs theano/tensor/tests/test_basic.py:BatchedDotTester
```

theano/tensor/blas.py
```diff
@@ -2154,8 +2154,8 @@ def contiguous(var, ndim):
             return " && ".join([
                 " && ".join("{strides}[{i}] > 0 && {strides}[{i}] % type_size == 0"
                             .format(strides=strides, i=i) for i in range(ndim)),
-                "(%s)" % " || ".join("{strides}[{i}] == type_size"
-                                     .format(strides=strides, i=i) for i in range(ndim)),
+                "(%s)" % (" || ".join("{strides}[{i}] == type_size"
+                                      .format(strides=strides, i=i) for i in range(1, ndim)) or '1'),
```
nouiz: why the `or '1'`? Also I think both range should be `range(1, ndim)`

nouiz:
jenkins failure are not related to this PR and I have another PR that fix them. So merging.

PR #5362, opened 2016-12-22
Force install of the new commit of Theano

lamblin:
Otherwise, it is just using the cached one...


nouiz:
jenkins failed, but not due to this PR, so merging.

PR #5367, opened 2016-12-26
matmul() operation

BarclayII:
After looking at PRs from https://github.com/Theano/Theano/pull/3769 and https://github.com/Theano/Theano/pull/5326 I decided to make my own implementation of matmul() operator.  It does not use `scan()` operation and it seems to pass all tests in https://github.com/Theano/Theano/pull/3769.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -6039,6 +6039,46 @@ def outer(x, y):
         y.dimshuffle('x', 0))
 
 
+def _matmul2(x, y):
+    """numpy.matmul() equivalent for 2D-or-more tensors only."""
+
+    # Make number of dimensions match, prepend broadcastable axes if needed.
+    newdims = abs(x.ndim - y.ndim)
+    if x.ndim < y.ndim:
+        x = x.dimshuffle(*(['x'] * newdims + list(range(x.ndim))))
+        dims = y.ndim
+    else:
+        y = y.dimshuffle(*(['x'] * newdims + list(range(y.ndim))))
+        dims = x.ndim
+
+    # Align the last axis of @x with the last second axis of @y
+    x = x.dimshuffle(*(list(range(dims)) + ['x']))
+    y = y.dimshuffle(*(list(range(dims - 2)) + ['x', dims - 2, dims - 1]))
+
+    # Compute the "elementwise" matrix product
+    return (x * y).sum(axis=-2)
+
+
+def matmul(x, y):
+    """Theano equivalent of numpy.matmul()."""
+
+    if (x.ndim == 0) or (y.ndim == 0):
+        raise ValueError("Scalar operands are not allowed, use '*' instead")
+
+    if (x.ndim == 1):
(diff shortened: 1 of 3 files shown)


BarclayII:
Probably not very efficient because it does the job by broadcasting axes to size `M*N*P` and then performing elementwise multiplication, given matrices with size `M*N` and `N*P`.

theano/tensor/var.py
```diff
@@ -244,6 +244,18 @@ def __rdivmod__(self, other):
     def __rpow__(self, other):
         return theano.tensor.basic.pow(other, self)
 
+    # ARITHMETIC - MATRIX MULTIPLICATION
```
emillynge: I'm curious why you place these methods here with its own section?

Is matmul a different class of aritmetic operation?

theano/tensor/basic.py
```diff
@@ -6039,6 +6039,54 @@ def outer(x, y):
...
+    else:
+        y = y.dimshuffle(*(['x'] * newdims + list(range(y.ndim))))
+        dims = x.ndim
+
+    # Align the last axis of @x with the last second axis of @y
+    x = x.dimshuffle(*(list(range(dims)) + ['x']))
+    y = y.dimshuffle(*(list(range(dims - 2)) + ['x', dims - 2, dims - 1]))
+
+    # Compute the "elementwise" matrix product
+    return (x * y).sum(axis=-2)
```
emillynge: Are we reinventing the dot operator here?

It does not seem like a good idea to do matrix multiplication using anything other than blas libraries (like numpy). You risk doing things horrificly inefficient.

theano/tensor/var.py
BarclayII: Not sure where to put them...

theano/tensor/basic.py
BarclayII: I guess you're right.  I profiled my matmul with numpy's and found that mine is more than 2 times slower:
```
In [9]: %timeit NP.matmul(RNG.randn(16, 20, 100, 50), RNG.randn(16, 20, 50, 100))
1 loop, best of 3: 277 ms per loop

In [10]: %timeit f(RNG.randn(16, 20,100, 50), RNG.randn(16, 20, 50, 100))
1 loop, best of 3: 736 ms per loop
```

I'll take a look at how `BatchedDot` works and see if I can do `matmul` more efficiently using standard BLAS.  I think there should be better ways doing this.

emillynge: I found that the `BatchedDot` implementation looks awfully complicated :/

I'm thinking maybe just using it directly for ndim <= 3 and raising a `NotImplementedError` for ndim>3.
In my PR there I use a reshape trick to get matrices down to 3 dimensions, but it most likely doesn't play nice with broadcastable dimensions.

khaotik:
I think it's okay to just call numpy's matmul with `c_code` if the version is up. I'm not smart enough to figure out how to do it efficiently in GPU with a *simple* method though.

Issue #5369, opened 2016-12-26
Failed cudnn version validation


aviveise:
Hi,
I think there is a bug in _dnn_check_version method. 
The code contains "if version >= 5200" and should contain "if v >= 5200"

Thanks

gvtulder:
Yes, that doesn't seem right. It was introduced in #5357, which also caused another problem for me. I submitted a patch to fix both issues, see #5373 .

PR #5357, opened 2016-12-21
Select the dnn convolution algorithm using actually available memory

nouiz:
New PR rebased of gh-5305 to try to get travis passing.

Issue #4978, opened 2016-09-15
gpuarray: compiler-bindir passed to g++


gvtulder:
Perhaps this is something specific to my system, perhaps not. If I have the `nvcc.compiler_bindir` config option set, the `theano.gpuarray.dnn` functions do not compile. The tests produce errors like these:
```
test_softmax (theano.gpuarray.tests.test_dnn.TestDnnInferShapes) ... SKIP: cannot compile with cuDNN. We got this error:
g++: error: unrecognized command line option ‘--compiler-bindir’
```

My suspicion is that this is caused by the code in [theano/gpuarray/dnn.py#L70-L71](https://github.com/Theano/Theano/blob/master/theano/gpuarray/dnn.py#L70-L71) that adds `--compiler-bindir` to the list of arguments:
```python
    if config.nvcc.compiler_bindir:
        params.extend(['--compiler-bindir', config.nvcc.compiler_bindir])
# ...
    avail, out, err = GCC_compiler.try_flags(
        params, preambule=preambule, body=body,
        try_run=False, output=True)
```
If I remove the lines that add the parameter, everything works.

This code seems to be similar to that in `theano/sandbox/cuda/__init__.py`, with the difference that the gpuarray code runs g++, which doesn't accept the argument, and the sandbox code calls nvcc, which does:
```python

            if config.nvcc.compiler_bindir:
                params.extend(['--compiler-bindir',
                               '%s%s%s' % (path_wrapper, config.nvcc.compiler_bindir, path_wrapper)])
# ...
            comp, out, err = nvcc_compiler.NVCC_compiler.try_flags(
                flag_list=params, preambule=preambule, body=body,
                try_run=False, output=True)
```

Is this a bug in the code or is this something specific to my setup?

lamblin:
It is an issue in Theano.
Passing `nvcc` options to `g++` looks like a bad idea in general...
@abergeron what do you think the fix should be?

gvtulder:
I'm not sure if this is the right solution, but this (#5010) fix works for me.

PR #5373, opened 2016-12-28
gpuarray: Revert part of c6ffa46

gvtulder:
These two patches revert two changes introduced by c6ffa46:
1. The cuDNN version check should use the correct variable name (reported in #5369). This removes an incorrect warning.
2. The `--compiler-bindir` option for `nvcc` should not be passed to the `g++` compiler. This problem was already fixed once in #4978 but reintroduced by the recent merging of #5357.

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -135,7 +135,7 @@ def _dnn_check_version():
     if v < 5000:
         return False, "cuDNN version is too old. Update to v5, was %d." % v
     # 5200 should not print warning with cudnn 5.1 final.
-    if version >= 5200:
+    if v >= 5200:
         warnings.warn("Your cuDNN version is more recent than "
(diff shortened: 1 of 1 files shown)


theano/gpuarray/dnn.py
```diff
@@ -116,9 +116,6 @@ def _dnn_check_compile():
         params.extend(['-I%s%s%s' % (path_wrapper, config.dnn.include_path, path_wrapper)])
     if config.dnn.library_path:
         params.extend(['-L%s%s%s' % (path_wrapper, config.dnn.library_path, path_wrapper)])
-    if config.nvcc.compiler_bindir:
```
nouiz: Can you add a comment why it should not be there? I was pretty sure it needed to be there, so it will help prevent it getting added again.

nouiz:
merging now as it fix things.

thanks

PR #5376, opened 2016-12-29
fix variable error in  _dnn_check_version()

whr94621:
version is a name of function, should be v the result of version()


nouiz:
jenkins test this

nouiz:
There is already a PR for this, so closing: https://github.com/Theano/Theano/pull/5373/files

thanks

PR #5377, opened 2016-12-30
Add regression test for nvcc.compiler_bindir (#5373)

gvtulder:
A test to keep if the issue in #5373 from coming back.


nouiz:
better then a comment!

jenkins test this.

PR #5380, opened 2017-01-03
compatible when axis < 0 in tensor.roll(x, shift, axis=None)

happygds:
When I use `T.roll(x, 1, axis=-2)` for a tensor `x` of rank `2`, it raised an error as follows:

> ValueError: The index list is longer (size 5) than the number of dimensions of the tensor(namely 3).   You are asking for a dimension of the tensor that does not exist!   You might need to use dimshuffle to add extra dimension to your tensor.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4218,7 +4218,7 @@ def roll(x, shift, axis=None):
             return roll(y, shift, axis=0).reshape(x.shape)
         else:
             axis = 0
-            
+
     if axis < 0:
         if x.ndim == 0:
             axis = 0
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4218,6 +4218,12 @@ def roll(x, shift, axis=None):
             return roll(y, shift, axis=0).reshape(x.shape)
         else:
             axis = 0
+            
+    if axis < 0:
+        if x.ndim == 0:
+            axis = 0
+        else:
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -4219,6 +4219,12 @@ def roll(x, shift, axis=None):
         else:
             axis = 0
 
+    if axis < 0:
```
nouiz: Can you add a test in the file tensor/tests/test_basic.py to make sure this don't get broken in the futur?

happygds: Yes, I have added that.

theano/tensor/basic.py
```diff
@@ -4219,6 +4219,12 @@ def roll(x, shift, axis=None):
...
+        if x.ndim == 0:
+            axis = 0
+        else:
+            axis = axis % x.ndim
```
nouiz: I find the way you handle the axis strange. Normally, we do something like:

    if axis < 0: axis += x.ndim

This have the advantage to return an error in case like axis=-1000 when ndim = 2. With your code it work. Does it work for numpy in that case? Can you add a test for that case and adjust the code if needed?

happygds: Thanks for your advice, it doesn't work for numpy in that case. I have changed it according to your suggestion.

nouiz:
thanks. merging

PR #5385, opened 2017-01-04
Bug: gradient of ifelse with gpu inputs

gvtulder:
There is a problem with the `IfElse` gradient when the inputs are on the GPU. There's a small example below. If you compute the gradient of such an `IfElse`, the [`grad` method](https://github.com/Theano/Theano/blob/master/theano/ifelse.py#L220-L227) will construct two new `IfElse` nodes. For these new nodes, one branch will get the gradient (a GPU tensor), while the other branch gets a new variable created with `tensor.zeros_like`. Because that is not a GPU tensor, the `IfElse` will complain that the types of the return values are not the same:
```
TypeError: ('IfElse requires same types for true and false return values',
    GpuFromHost<None>.0, Elemwise{second,no_inplace}.0,
    GpuArrayType<None>(float32, (False,)), TensorType(float32, vector))
```

Partial example script:
```python
x = tensor.vector('x')
y = tensor.vector('y')
c = tensor.iscalar('c')

x_gpu = basic_ops.as_gpuarray_variable(x, test_ctx_name)
y_gpu = basic_ops.as_gpuarray_variable(y, test_ctx_name)

out = theano.ifelse.ifelse(c, x_gpu, y_gpu)
theano.grad(out.sum(), [x, y])
```

This PR adds a test for this. There's also a solution: making sure that inputs for `IfElse` are always converted to normal TensorVariables. I'm not sure if that is the correct solution, but it does help.

### DIFF
--- a/theano/ifelse.py
+++ b/theano/ifelse.py
@@ -173,7 +173,9 @@ class IfElse(Op):
             # to keep them be cuda ndarrays
             nw_args = []
             for x in args:
-                if isinstance(x, theano.Variable):
+                if hasattr(x, '_as_TensorVariable'):
+                    nw_args.append(x._as_TensorVariable())
+                elif isinstance(x, theano.Variable):
                     nw_args.append(x)
                 else:
                     nw_args.append(theano.tensor.as_tensor_variable(x))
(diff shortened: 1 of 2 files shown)


nouiz:
thanks

Issue #5216, opened 2016-11-11
BadOptimization when using gpuarray backend in DebugMode


bbudescu:
When using the 'cuda' gpu backend in DebugMode function compilation fails here [https://github.com/Theano/Theano/blob/c0548210e11603141d70250871de8191e0a37037/theano/gpuarray/dnn.py#L216](url), because both expected and obtained outputs are PyCapsule objects containing NULL pointers (which, I assume, are supposed to be cudnnHandle_t's), which compare as different. BadOptimization is raised.

nouiz:
Can you give the full error message?
Do you have code to reproduce that?

@abergeron, do you have time to look at this?

Sentient07:
From what is reported (with reference to the line number of the error ) it seems that @bbudescu is having problem with `CUDNN_STATUS_INTERNAL_ERROR`. Did you try what @nouiz suggested in this issue, https://github.com/Theano/Theano/issues/5206 ?

bbudescu:
Hi, sorry for the delays (both the one until answering now, and the one until I'll be able to come up with a minimal scenario in which the error is raised). I'll get back on it as soon as I have the time. This will be another few days, I guess. The model I'm currently trying out is based on this: https://github.com/jocicmarko/ultrasound-nerve-segmentation. 

As to #5206, as far as I understand, that one happens all the time. Mine only surfaces in DebugMode. If it's of any help, I'm currently experimenting on a ubuntu machine with cuda 7.5 and cudnn 5005 (ubuntu 16.04).

nouiz:
Do you still have the full error message? Without this, we can't do something about that.

bbudescu:
Hi, as I anticipated, I'd like to apologize for the delay with which I'm writing. 

This is the error I get (ignore the "<timestamp> - WARNING" and the "(std streams)") :
...
2016-12-06 19:38:26,925 - WARNING:   File "/home/bbudescu/build/src/keras/keras/engine/training.py", line 1374, in fit_generator (std streams)
2016-12-06 19:38:26,926 - WARNING: self._make_train_function() (std streams)
2016-12-06 19:38:26,927 - WARNING:   File "/home/bbudescu/build/src/keras/keras/engine/training.py", line 726, in _make_train_function (std streams)
2016-12-06 19:38:26,927 - WARNING: **self._function_kwargs) (std streams)
[...]

PR #5387, opened 2017-01-05
Mixed fixes

abergeron:
fix gh-5216


nouiz:
there was only 1 fix, but as it is good, I merged it. If you had more in your local branch, make a new PR with them.

PR #5389, opened 2017-01-09
theano.pp crash fix and display fix

nouiz:


theano/tensor/subtensor.py
```diff
@@ -967,35 +967,43 @@ def process(self, r, pstate):
...
+                            msg2 = entry.stop
+
+                        if entry.step is None:
+                            msg3 = ""
+                        else:
+                            msg3 = ":%s" % entry.step
+
+                        sidxs.append("%s:%s%s" % (msg1, msg2, msg3))
+            finally:
+                pstate.precedence = old_precedence
```
abergeron: Is it guarenteed that there is a `precedence` attribute?  Because the changes you did seem to indicate that it's not.  So this part could still crash.

nouiz: If there isn't one, it will set a new one that will be None. In the current logic, it should be equivalent. This is already what is done inside the file theano/printing.py.

PR #5390, opened 2017-01-09
Fix the buildbot.

abergeron:


theano/gpuarray/tests/test_opt.py
```diff
@@ -501,6 +501,8 @@ def test_no_complex():
 
 
 def test_local_lift_solve():
+    if not cusolver_available:
+        raise SkipTest('No cuSolver')
```
nouiz: This is good, but I think we should update the buildbot to use cuda8 too. @slefrancois how long would this take?

slefrancois: The buildbot on tikuanyin is already on CUDA 8, same as the jenkins vms. cusolver should come standard right?

PR #5393, opened 2017-01-10
softmax opti for upstream

KevinYuk:
This code is used to do softmax and softmaxGrad opti. Can increase the theano RNN/LSTM performance.

The theano UT has all passed.
Don't affect the code style of theano.

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -384,6 +384,8 @@ class SoftmaxGrad(gof.Op):
             }
         }
 
+        int chunk = 32;
+        #pragma omp parallel for schedule(static, chunk)
         for (size_t i = 0; i < PyArray_DIMS(%(dx)s)[0]; ++i)
         {
             const dtype_%(dy)s* __restrict__ dy_i = (dtype_%(dy)s*) (PyArray_BYTES(%(dy)s) + PyArray_STRIDES(%(dy)s)[0] * i);
@@ -504,6 +506,8 @@ class Softmax(gof.Op):
         """
 
         begin_row_loop = """
+        int chunk = 32;
+        #pragma omp parallel for schedule(static, chunk)
         for (size_t i = 0; i < Nx[0]; ++i)
         {
             size_t j;


theano/tensor/nnet/nnet.py
```diff
@@ -384,6 +384,7 @@ def c_code(self, node, name, inp, out, sub):
             }
         }
 
+        #pragma omp parallel for schedule(static, 32)
```
nouiz: Thanks for the PR, but I think this isn't enought. You should make the ops inherit from theano.gof.OpenMPOp instead of theano.gof.Op. This will allow Theano to pass the right compiler parameter to enable it and will respect the flag openmp={True,False}.

Did you do some timing? I think with the current state, it should always be disabled as we don't tell g++ to enable openmp.

KevinYuk: Hi nouiz,

Thanks very much for your kindly comments.

1. Regarding to the 1st issue you point out:
Even inherit from theano.gof.Op, my test result show that the compiler still respect flag openmp={True, False}.
My .theanorc configuration is as below:
```
[global]
device = cpu
floatX = float32
cxx = g++
mode = FAST_RUN
openmp = True
openmp_elemwise_minsize = 10
profile=False
[gcc]
cxxflags = -fopenmp -O3 -opt-prefetch=2 -funroll-loops
[blas]
ldflags = -lmkl_rt
```

2. Regarding to 2nd issue:
Yes. I have done timing. (BTW, becasue our test beds are maintained, I use PC to test it now. I use BDW (Xeon E5 2699 v4) before.)
Below is test result:
```
HW: Intel(R) Core(TM) i7-5960X CPU @ 3.00GHz
workload: [dl4mt](https://github.com/nyu-dl/dl4mt-tutorial.git)
```
.theanorc configuration:
```
[global]
device = cpu
floatX = float32
cxx = g++
mode = FAST_RUN
openmp = True
openmp_elemwise_minsize = 10
profile=False
[gcc]
cxxflags = -fopenmp -O3 -opt-prefetch=2 -funroll-loops
[blas]
ldflags = -lmkl_rt
```

profile(just need focus on softmax and softmaxGrad OPs):
gcc
```
===============================================================
non-opti
-----------------------------------------------------------
Class
---
  29.4%    29.4%      82.766s       1.42e-01s     Py     584       6   theano.scan_module.scan_op.Scan
  26.9%    56.3%      75.623s       3.94e-01s     C      192       2   theano.tensor.nnet.nnet.Softmax
  23.4%    79.7%      65.695s       1.46e-02s     C     4512      46   theano.tensor.blas.Dot22
  12.1%    91.8%      34.037s       9.88e-04s     C    34440     348   theano.tensor.elemwise.Elemwise
   2.8%    94.6%       7.796s       7.80e-02s     C      100       1   theano.tensor.nnet.nnet.SoftmaxGrad
   2.7%    97.3%       7.587s       4.79e-03s     C     1584      16   theano.tensor.basic.Alloc
   1.8%    99.1%       5.176s       4.74e-03s     C     1092      11   theano.tensor.elemwise.Sum
   0.4%    99.5%       1.115s       6.28e-04s     C     1776      18   theano.tensor.subtensor.IncSubtensor
   0.2%    99.7%       0.576s       8.40e-05s     C     6856      70   theano.tensor.basic.Reshape
   0.2%    99.9%       0.543s       1.81e-03s     C      300       3   theano.tensor.subtensor.AdvancedIncSubtensor1
   0.0%    99.9%       0.123s       6.17e-04s     C      200       2   theano.tensor.blas.Gemm
   0.0%   100.0%       0.108s       1.88e-04s     C      576       6   theano.tensor.subtensor.AdvancedSubtensor1
   0.0%   100.0%       0.012s       2.07e-06s     C     5680      58   theano.tensor.opt.MakeVector
   0.0%   100.0%       0.009s       1.71e-06s     C     5296      54   theano.tensor.elemwise.DimShuffle
   0.0%   100.0%       0.009s       1.71e-06s     C     5264      54   theano.compile.ops.Shape_i
   0.0%   100.0%       0.009s       4.44e-05s     Py     192       2   theano.tensor.basic.ARange
   0.0%   100.0%       0.008s       2.25e-06s     C     3728      38   theano.tensor.subtensor.Subtensor
   0.0%   100.0%       0.004s       9.47e-07s     C     4380      45   theano.tensor.basic.ScalarFromTensor
   0.0%   100.0%       0.001s       1.97e-06s     C      384       4   theano.tensor.basic.AllocEmpty
   0.0%   100.0%       0.000s       7.93e-07s     C      584       6   theano.compile.ops.Rebroadcast

opti
-----------------------------------------------------------   
  35.9%    35.9%      85.374s       1.46e-01s     Py     584       6   theano.scan_module.scan_op.Scan
  27.1%    63.0%      64.548s       1.43e-02s     C     4512      46   theano.tensor.blas.Dot22
  17.7%    80.7%      42.142s       1.22e-03s     C    34440     348   theano.tensor.elemwise.Elemwise
  10.3%    90.9%      24.428s       1.27e-01s     C      192       2   theano.tensor.nnet.nnet.Softmax
   4.2%    95.1%      10.074s       6.36e-03s     C     1584      16   theano.tensor.basic.Alloc
   2.2%    97.4%       5.295s       4.85e-03s     C     1092      11   theano.tensor.elemwise.Sum
   1.6%    98.9%       3.735s       3.73e-02s     C      100       1   theano.tensor.nnet.nnet.SoftmaxGrad
   0.5%    99.4%       1.122s       6.32e-04s     C     1776      18   theano.tensor.subtensor.IncSubtensor
   0.2%    99.7%       0.590s       8.61e-05s     C     6856      70   theano.tensor.basic.Reshape
   0.2%    99.9%       0.533s       1.78e-03s     C      300       3   theano.tensor.subtensor.AdvancedIncSubtensor1
   0.1%    99.9%       0.122s       6.09e-04s     C      200       2   theano.tensor.blas.Gemm
   0.0%   100.0%       0.112s       1.95e-04s     C      576       6   theano.tensor.subtensor.AdvancedSubtensor1
   0.0%   100.0%       0.013s       2.33e-06s     C     5680      58   theano.tensor.opt.MakeVector
   0.0%   100.0%       0.009s       1.79e-06s     C     5296      54   theano.tensor.elemwise.DimShuffle
   0.0%   100.0%       0.009s       1.79e-06s     C     5264      54   theano.compile.ops.Shape_i
   0.0%   100.0%       0.009s       2.51e-06s     C     3728      38   theano.tensor.subtensor.Subtensor
   0.0%   100.0%       0.009s       4.85e-05s     Py     192       2   theano.tensor.basic.ARange
   0.0%   100.0%       0.004s       1.02e-06s     C     4380      45   theano.tensor.basic.ScalarFromTensor
   0.0%   100.0%       0.001s       1.83e-06s     C      384       4   theano.tensor.basic.AllocEmpty
   0.0%   100.0%       0.000s       4.39e-06s     C      100       1   theano.compile.ops.DeepCopyOp
```

Please help check it.

Thanks a lot.
Kevin Yu

nouiz: thanks for the timing. I don't understand why it should work without inheriting from OpenMPOp. Do you? Without that, we should not pass the -fopenmp parameter to g++, so it should ignore the pragma.

Which compiler do you use? On which OS?

This PR still need to inherit from OpenMPOp. Can you do that change?

thanks

KevinYuk: Hi nouiz,

Thanks for your reply.
**1. I do the timing by using gcc and icc. Both of them. 
2. My OS are Ubuntu and Centos. Both of them works.
3. I'll try inherit from OpenMPOp and update the diff of PR.**

Thanks a lot.

KevinYuk: Hi nouiz,

I have make softmax and softmaxGrad inherit from gof.OpenMPOp and commit the code.
Please help review it again.

Thanks very much.

KevinYuk: Hi nouiz,

Could you please help review it again? :) I have made softmax and softmaxGrad inherit from gof.OpenMPOp and commit the code.

Many thanks.

nouiz: thanks. I did some local benchmark. Using openmp=True, vs openmp=False make a speed difference. But here, openmp=True is slower in a small test case:

```
        x = T.matrix()
        f = theano.function([x], softmax_op(x))

        xv = numpy.random.randn(600, 12800).astype(config.floatX)
        for i in range(10):
            f(xv)
```

What shapes did you used? If the problem is that my shapes are too small, I think we should a condition to the openmp pragma to not run for too small shapes. We do this for elemwise too.

Note, we also have the ops SoftmaxWithBias that is used sometimes. When we have softmax working well, doing the same for it would be great.

KevinYuk: Hi nouiz,
Thanks for your testing.

**1. Regarding to conditional openmp for different shape: I agree with you.**
Actually, I want to do it before. But this will bring more code change. However, if you think it's OK. I am OK. I am working on it, will send out the code change later.
Just for your reference, below is my size samples which are driven by RNN/LSTM workload(I'll choose 128, 256000 as the condition to do code split. Surely, I have tested it, it works fine):
`Softmax: 544, 30000, 
Softmax: 480, 30000, 
Softmax: 448, 30000, 
Softmax: 384, 30000, 
Softmax: 320, 30000, 
Softmax: 140, 30000, 
Softmax: 2176, 30000,
Softmax: 1664, 30000,
Softmax: 1536, 30000,
Softmax: 1344, 30000,
Softmax: 1248, 30000,
Softmax: 1184, 30000,
Softmax: 1088, 30000,
Softmax: 1024, 30000,
Softmax: 960, 30000, 
Softmax: 864, 30000, 
Softmax: 800, 30000, 
Softmax: 736, 30000, 
Softmax: 672, 30000, 
Softmax: 608, 30000, 
Softmax: 576, 30000, 
Softmax: 512, 30000, 
Softmax: 416, 30000, 
Softmax: 352, 30000, 
Softmax: 225, 30000, 
Softmax: 2080, 30000,
Softmax: 1600, 30000,
Softmax: 1472, 30000,
Softmax: 1344, 30000,
Softmax: 1280, 30000,
Softmax: 1216, 30000,
Softmax: 1152, 30000,
Softmax: 1088, 30000,
Softmax: 992, 30000, 
Softmax: 960, 30000, 
Softmax: 896, 30000, 
Softmax: 800, 30000, 
Softmax: 736, 30000, 
Softmax: 672, 30000, 
Softmax: 608, 30000, 
Softmax: 544, 30000, 
Softmax: 448, 30000, 
Softmax: 384, 30000, 
Softmax: 42, 30000,
Softmax: 1920, 30000,
Softmax: 1536, 30000,
Softmax: 1376, 30000,
Softmax: 1248, 30000,
Softmax: 1120, 30000,
Softmax: 992, 30000, 
Softmax: 928, 30000, 
Softmax: 800, 30000, 
Softmax: 704, 30000, 
Softmax: 608, 30000, 
Softmax: 512, 30000, 
Softmax: 448, 30000, 
Softmax: 320, 30000, 
Softmax: 48, 30000,`

**2. Regarding to your second comment about SoftmaxWithBias:**
I have consider this before. Actually, it includes SoftmaxWithBias, LogSoftmax, etc. However, there is not the related famous workload(e.g.: Alexnet, Googlenet, dl4mt etc.) to direct our optimization work. My opinion is: don't do it now :). When there is related workload to drive it, we do it. What's your opinion?


BTW, several months before, our team focus on theano CNN optimization. At that time, softmax is not the bottleneck, because there is only about 1000 kinds  classifications result for CNN(e.g.: Alexnet and Googlent). When we try to do theano RNN/LSTM optimization, we found that softmax cost so much CPU time. Because for RNN/LSTM, there are so many kinds classifications.

KevinYuk: Hi nouiz,

I have commited the code to split the small size input for Softmax and SoftmaxGrad. Please help review the code.

Thanks a lot.

KevinYuk:
Hi nouiz,

**1. I try the "conditional openmp" now. Committed the code, please help check it.
2. I don't use script, but I use dl4mt session1 workload to check the softmax input shape.**

If you  think the threshold is not reasonable, please give your advice. I would appreciate it.
Actually, I found that the performance of **conditional openmp** is lower than **normal openmp** because it introduces the judgement overhead. Especially for small narrow matrix. The choice of threshold is a little tricky.


Thanks a lot.

KevinYuk:
In other words, if we use conditional openmp, we already failed on small narrow matrix.

nouiz:
I didn't expected conditional openmp to have significant overhead. How much
difference do you see? If there is such a difference, then it make sence to
code it manually as you did.

Finding the threashold is tricky for sure. That is why I think making a
script to test different shapes would be great. We run it with and without
openmp and it help find the threashold. I don't have time to work much on
it, but I can give you the Theano part in case you aren't familiar enough
to make that part.

import theano
import numpy
m=theano.tensor.dmatrix()
f=theano.function([m], theano.tensor.nnet.softmax(m))
for sh in [(2,3),]:
    data = numpy.random.rand(2,2)
    # TODO time the line bellow:
    f(data)
    #print sh, data

KevinYuk:
Hi nouiz,

image

KevinYuk:
Hi nouiz,

Could you please help check the reply above?

Many thanks.

nouiz:
I was sick yesterday.

The data look strange. The cut off point you show in red see way to
drastic. Why just before there was no speed up with openmp hand designed?

The data show clearly the hand-design openmp instead of conditional openmp
is better. I'm still surprised by this, but it should we should go with
this.

Also, the data only investigate one of the shapes, not both. From a
theoretical point of view, I think we could only check the total number of
elements in the inputs. This would make it more clear.

Can you give me the scripts you used for those numbers so that I can
reproduce them here?

KevinYuk:
Oh, I am sorry to hear that.

The data look strange. The cut off point you show in red see way to
drastic. Why just before there was no speed up with openmp hand designed?
>>No, I only make it work after (256, 30000). But I found that hand-designed openmp and conditional openmp perf is worse than openmp because of "judgement overhead". Only when the "judgement overhead" is less than calculation cost, we can see it speeds up.(That's why it failed on small narrow matrix) 

The data show clearly the hand-design openmp instead of conditional openmp
is better. I'm still surprised by this, but it should we should go with
this.
>>Yes, it's strange. hand-design is more faster.

Also, the data only investigate one of the shapes, not both. From a
theoretical point of view, I think we could only check the total number of
elements in the inputs. This would make it more clear.

Can you give me the scripts you used for those numbers so that I can
reproduce them here?
>>Yes. Please check below. But I don't think you can reproduce it as me, because it depends on different cpu cores and frequency. Our cpu is Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz.

`import os
import sys

import numpy

import theano
import theano.tensor as T
import time
import math

profile = False

def main():
    #1. first test case
    """x = T.matrix()
    f = theano.function([x], T.nnet.softmax(x), profile=profile)
    for i in range(1, 4):
        for j in range(200, 256):
            shapeX = (int)(128*i)
            shapeY = (int)(128*j)
            xv = numpy.random.randn(shapeX, shapeY).astype(theano.config.floatX)
            #xv = numpy.random.randn(shapeY, shapeX).astype(theano.config.floatX)

            start = time.time()
            for ii in range(10):
                f(xv)
            end = time.time()
            total = end - start
            print '{0:>8}'.format(shapeX), '{0:>8}'.format(shapeY), '{0:>24}'.format(total)
            #print '{0:>8}'.format(shapeY), '{0:>8}'.format(shapeX), '{0:>24}'.format(total)"""

    #2. second test case
    x = T.matrix()
    f = theano.function([x], T.nnet.softmax(x), profile=profile)
    for i in range(32, 1025):
        for j in range(30000, 30001):
            shapeX = i
            shapeY = j
            xv = numpy.random.randn(shapeX, shapeY).astype(theano.config.floatX)
            #xv = numpy.random.randn(shapeY, shapeX).astype(theano.config.floatX)

            start = time.time()
            for ii in range(10):
                f(xv)
            end = time.time()
            total = end - start
            print '{0:>8}'.format(shapeX), '{0:>8}'.format(shapeY), '{0:>24}'.format(total)
            #print '{0:>8}'.format(shapeY), '{0:>8}'.format(shapeX), '{0:>24}'.format(total)

    #3. third test case from nouiz
    """m=theano.tensor.dmatrix()
    f=theano.function([m], theano.tensor.nnet.softmax(m))
    for sh in [(2,3),]:
        data = numpy.random.rand(2,2)
        start = time.time()
        for ii in range(10):
            f(data)
        end = time.time()
        total = end - start
        print sh, data"""



if __name__ == '__main__':
    main()

`

KevinYuk:
Hi nouiz,

Could you please help check the reply above?

Many thanks.

KevinYuk:
Hi nouiz,

Could you please help check the issue again? It's already more than two weeks.
If you want me to do other test, please tell me. Or you can tell me your concern. Please!

Thanks very much.

PR #5395, opened 2017-01-10
Spatial pooling

Sentient07:


Sentient07:
This PR, I have implemented similar to Lasagne's implementation. @nouiz had asked me to create a new Op, but i am not sure how I should proceed with that. In SP-Pooling, the pooling size depends on the image's size. In the test, I have a fixed size image (made it to a shared variable) and have set the pooling size. But, for someone using a network, where the input to this pooling layer happens to be a TensorVariable, how can one get the size? For better consistency, is it advisable to create an Op, where the `pooling_size` is a Variable?

theano/tensor/signal/pool.py
```diff
@@ -255,6 +255,20 @@ def pool_3d(input, ws=None, ignore_border=None, stride=None, pad=(0, 0, 0),
...
+    input_size = image_shape[2:]
+    pooled_data_list = []
+    for pool_dim in out_dimension:
+        pool_size = tuple((i + pool_dim - 1) // pool_dim for i in input_size)
+        stride_size = tuple((i // pool_dim) for i in input_size)
+        pooled_part = pool_2d(image, pool_size, ignore_border=True, st=stride_size, padding=(0, 0), mode='max')
+        pooled_part = pooled_part.flatten(3)
+        pooled_data_list.append(pooled_part)
+
+    return tensor.concatenate(pooled_data_list, axis=2)
```
nouiz: I think you only need this implementation. I don't see why you need the one in dnn. The optimizer should move it completly to the GPU. If I'm missing something, tell me.

It was maybe needed in the past before we had our new pool interface, but I don't think it is needed anymore.

For the name, I'm not sure what we should use. Do you know what they use in other software like torch, tensorflow or caffe for that name?

This will also need a docstring.

I looked at the paper very rapidly, and I don't think a new op is useful for this.

I think it would be useful to put a default of None for image_shape and if it is NOne, do: `image_shape = image.shape`. This now should work, but didn't worked in the past and probably why it was in lasagne dnn interface and not normal pool interface.

Sentient07: Yeah, my bad I wasn't clear in my previous comment. Even the stride is dependent on the pooling_size (which indeed is dependent on the input layer's size). This causes the problem as stride seems to take only fixed values.

nouiz: it should also work for stride. Use the new parameter name stride instead of st.

theano/roi_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+t_data = T.ftensor4()
+t_rois = T.fmatrix()
+
+t_outs = op(t_data, t_rois)
+t_c = t_outs[0].sum()
+
+t_g_data = T.grad(t_c, t_data)[0]
+
+f = theano.function([t_data, t_rois], t_outs + [t_g_data])
```
nouiz: Remove that file. It was included by mistake.

Sentient07: I'm just noticing. This was in my `.gitignore` and i don't know how it carried away while shifting branch.

theano/tensor/signal/pool.py
```diff
@@ -540,9 +540,9 @@ def make_node(self, x, ws, stride=None, pad=None):
                 raise NotImplementedError(
                     'padding works only with ignore_border=True')
             if isinstance(ws, (tuple, list)):
-                if any(pad[i] >= ws[i] for i in range(nd)):
-                    raise NotImplementedError(
-                        'padding must be smaller than strides')
+                if not (isinstance(pad[i], theano.Variable) or isinstance(ws[i], theano.Variable) for i in range(nd)):
```
nouiz: The test isn't good. It should be some think like:
for i in range(nd):
   if not isinstance(pad[i], ...) and not isinstance(ws[i], ...):
      if pad[i] >= ws[i]:
          raise

nouiz: I edited the comment above to show the indentation.

Sentient07:
@nouiz this is ready for a final review

theano/tensor/signal/pool.py
```diff
@@ -255,6 +255,64 @@ def pool_3d(input, ws=None, ignore_border=None, stride=None, pad=(0, 0, 0),
     return output
 
 
+def spp_pooling(image, out_dimension, image_shape):
```
nouiz: Your code should work if image_shape isn't set. So here, do image_shape=None, and make that case work as we discussed.

Sentient07: Done. 👍

Sentient07:
@nouiz if it is good enough, can it be merged ? (I want to open another PR and since I have quite a few open, I'm trying to wrap up some)

nouiz:
In lasagne, they merged last week new code that give more flexibility to this code to cover more cases. Before merging this PR, we should make sure we can have a user interface that is stable and will cover both cases.

We probably won't merge this PR before the final 0.9 version.

Sentient07:
Okay, I will checkout the version that was merged in Lasgane last week

Sentient07:
I thought I had commented this here, but I had done elsewhere and couldn't recollect where. I think it is a better idea to merge the ROIPool PR, and then have  this interface over the ROI Pooling(which is a pyramid with a singular depth). According to number of pyramid levels we want, we can stack one ROIPool over the other to achieve multiple pyramid levels. Is that good @nouiz ?

PR #5398, opened 2017-01-10
Implement conv2d_transpose convenience function

### LINKED ISSUE #3989: Add `conv2d_transpose` method to `theano.tensor.nnet.abstract_conv`
... and make it available in `theano.tensor.nnet`.
The semantics and name of the parameters was discussed IRL with @vdumoulin and @fvisin .

This is a follow-up to #3681 and #3872.

vdumoulin:
Fixes #3989. @lamblin would you mind having a look?


vdumoulin:
@lamblin @slefrancois I added a unit test and would be ready for a review.

slefrancois:
The unit tests fails on fast_compile because the unoptimized Op has no perform. Do we not test conv2d_grad_wrt_inputs directly?

Also, maybe we should add the corresponding unit test in theano.gpuarray/tests/test_abstractconv.py?

vdumoulin:
@slefrancois I added a GPU test by inferring how it was done in `theano.gpuarray.tests.test_abstractconv`; tell me if this is what you had in mind.

slefrancois:
looks good, let's see what jenkins says

theano/gpuarray/tests/test_abstractconv.py
```diff
@@ -252,3 +252,12 @@ def setUp(self):
...
             numpy.zeros((3, 5, 7, 11), dtype='float32'),
             context=get_context(test_ctx_name))
+
+
+class TestConv2dTranspose(test_abstract_conv.TestConv2dTranspose):
+    mode = mode_with_gpu
+
+    def test_interface(self):
+        if not dnn_available(test_ctx_name):
+            raise SkipTest(dnn_available.msg)
```
lamblin: I think we have CorrMM in the new back-end now, so the test should still be able to run without cuDNN.
In fact, CorrMM is currently the only way of having filter_dilation and subsampling at the same time.

theano/tensor/nnet/__init__.py
```diff
@@ -151,3 +152,86 @@ def conv2d(input, filters, input_shape=None, filter_shape=None,
...
+
+    filters: symbolic 4D tensor
+        Set of filters used in CNN layer of shape
+        (input channels, output channels, filter rows, filter columns).
+        See the optional parameter ``filter_shape``. **Note: the order for
+        ``output_channels`` and ``input_channels`` is reversed with respect to
+        ``conv2d``.**
+
+    output_shape: tuple/list of len 4 of int or Constant variable
+        The shape of the output of ``conv2d_transpose``.
```
lamblin: I think we should allow the last two elements of output_shape to be symbolic variables (tensor.scalar).
This is what we do for `conv2d_grad_wrt_inputs`, and it can be useful when we deal with images of heterogeneous sizes.

theano/tensor/nnet/__init__.py
```diff
@@ -151,3 +152,86 @@ def conv2d(input, filters, input_shape=None, filter_shape=None,
...
+        Set of feature maps generated by the transposed convolution. Tensor is
+        of shape (batch size, output channels, output rows, output columns)
+
+    Notes
+    -----
+        If cuDNN is available, it will be used on the
+        GPU. Otherwise, it is the *CorrMM* convolution that will be used
+        "caffe style convolution".
+
+        This is only supported in Theano 0.8 or the development
```
lamblin: It would be 0.9, but I don't think that paragraph is necessary any more, since we now have different versions of the doc available online at the same time. The documentation for that function will only show up in the development branch, and 0.9 when it is released.

lamblin: Also, we should probably at least mention the word "deconvolution" in here, so people looking for it will find the function.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1548,3 +1549,35 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+            [[[[2, 2, 4, 4, 4, 4, 4, 4, 2, 2],
+               [2, 2, 4, 4, 4, 4, 4, 4, 2, 2],
+               [4, 4, 8, 8, 8, 8, 8, 8, 4, 4],
+               [4, 4, 8, 8, 8, 8, 8, 8, 4, 4],
+               [4, 4, 8, 8, 8, 8, 8, 8, 4, 4],
+               [4, 4, 8, 8, 8, 8, 8, 8, 4, 4],
+               [4, 4, 8, 8, 8, 8, 8, 8, 4, 4],
+               [4, 4, 8, 8, 8, 8, 8, 8, 4, 4],
+               [2, 2, 4, 4, 4, 4, 4, 4, 2, 2],
+               [2, 2, 4, 4, 4, 4, 4, 4, 2, 2]]]] * 2)
```
lamblin: That test has too many symmetries for my comfort: failure modes like ignoring filter_flip, swapping row / columns, or swapping batch_size / input_channels would not be detected.
Also, filter_dilation is not tested.
It's not really critical, given that the implementation is simple, though.

theano/tensor/nnet/__init__.py
lamblin: I read the docstring before the implementation... since this function simply calls `conv2d_grad_wrt_inputs`, then it is supported.
Could you just update the docstring?

theano/gpuarray/tests/test_abstractconv.py
```diff
@@ -252,3 +252,12 @@ def setUp(self):
...
+        super(self, TestConv2dTranspose).test_interface()
```
slefrancois: Jenkins fails because super() argument must be type, not class.

vdumoulin: Fixed via ddf0047

theano/gpuarray/tests/test_abstractconv.py
vdumoulin: Fixed via ddf0047

theano/tensor/nnet/__init__.py
vdumoulin: Fixed via 8320eac

theano/tensor/nnet/__init__.py
vdumoulin: Fixed via 35ab10c

theano/tensor/nnet/tests/test_abstract_conv.py
vdumoulin: Given that this is a very thin wrapper around `conv2d_grad_wrt_inputs`, I think it's sufficient to make sure that there aren't any surprises in the interface, e.g. with regards to the order of input/output channel axes.

lamblin: OK.

lamblin:
Sorry, I forgot about the failure in FAST_COMPILE. There are usually 3 options:
- always use FAST_RUN in that test. Usually done by
```python
        mode = theano.compile.get_mode("FAST_RUN")
    else:
        mode = None
```
- skip the test if we are in FAST_COMPILE, assuming it is going to be tested in FAST_RUN in a different run anyway.
```python
        raise SkipTest("CorrMM don't work in FAST_COMPILE")
```
- Use the slow Python implementation of the abstract Op itself, and disable the optimization inserting CorrMM. It works only when the abstract interface is used, not for tests specific to CorrMM.
```python
        compile_mode = compile_mode.excluding("conv_gemm")
        compile_mode = compile_mode.excluding('AbstractConvCheck')
```

These 3 examples are actual examples from tests inside `theano/tensor/nnet/tests`.

In this case, since it is a small computation that uses the abstract interface, I would favor the 3rd option.

vdumoulin:
@lamblin Your last comment is addressed by ab86c74.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1548,3 +1549,39 @@ def test_compare_1D_and_2D_upsampling_values(self):
         f_1D = theano.function([], mat_1D, mode=self.compile_mode)
         f_2D = theano.function([], mat_2D, mode=self.compile_mode)
         utt.assert_allclose(f_1D(), f_2D(), rtol=1e-06)
+
+
+class TestConv2dTranspose(unittest.TestCase):
+    mode = None
```
lamblin: This should probably be something like `theano.compile.get_default_mode()` or `theano.compile.get_mode(None)`.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1548,3 +1549,39 @@ def test_compare_1D_and_2D_upsampling_values(self):
...
+        """Test conv2d_transpose wrapper.
+
+        This method tests that the order of the filter's
+        axes expected by the function produces the correct
+        output shape.
+
+        """
+        mode = self.mode
+        if theano.config.mode == "FAST_COMPILE":
+            mode = mode.excluding("conv_gemm").excluding('AbstractConvCheck')
```
slefrancois: Travis fails when mode=None:
AttributeError: 'NoneType' object has no attribute 'excluding'

nouiz:
do theano.compile.get_default_mode(mode).excluding(...)

vdumoulin:
@nouiz @slefrancois @lamblin Last hiccup should be fixed via 2890f00.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1564,7 +1564,8 @@ def test_interface(self):
         """
         mode = self.mode
         if theano.config.mode == "FAST_COMPILE":
-            mode = mode.excluding("conv_gemm").excluding('AbstractConvCheck')
+            mode = theano.compile.get_default_mode(
+                mode).excluding("conv_gemm").excluding("AbstractConvCheck")
```
slefrancois: TypeError: get_default_mode() takes 0 positional arguments but 1 was given

vdumoulin:
@slefrancois @lamblin all tests are passing now.

slefrancois:
@lamblin , it looks good to me. Can you do a final review and merge?

Issue #5345, opened 2016-12-16
Fix cache loading problem when libgpuarray version changes


lamblin:
See http://darjeeling.iro.umontreal.ca:8080/job/Buildbot_gpuarray/119/ for an example of error that occurred after the transition from 0.1 to 1.0.

nouiz:
I added this to the 0.9 milestone

nouiz:
done

PR #5400, opened 2017-01-11
Change caching method of GpuarrayType to fix issue #5345

notoraptor:
This is to fix issue #5345 . The c_code_cache_version() of GpuArrayType now depends on gpuarray ABI version instead of API version, so that cached files linked with an old gpuarray shared library will be recompiled every time the gpuarray ABI version changes. See gpuarray PR https://github.com/Theano/libgpuarray/pull/323 for new ABI methods added.

The fix works on my computer.

@abergeron @nouiz

### DIFF
--- a/theano/gpuarray/type.py
+++ b/theano/gpuarray/type.py
@@ -476,10 +476,8 @@ class GpuArrayType(Type):
         return ['gpuarray']
 
     def c_code_cache_version(self):
-        ver = pygpu.gpuarray.api_version()
-        # we only use the major version since the minor revision are
-        # API-compatible.
-        return (2, ver[0])
+        ver = pygpu.gpuarray.abi_version()
+        return (2,) + ver
 
 
 class _operators(_tensor_py_operators):
--- a/theano/gpuarray/type.py
+++ b/theano/gpuarray/type.py
@@ -476,7 +476,6 @@ class GpuArrayType(Type):
         return ['gpuarray']
 
     def c_code_cache_version(self):
-        # import pdb; pdb.set_trace()
         ver = pygpu.gpuarray.abi_version()
         # we only use the major version since the minor revision are compatible.
         return (2, ver[0])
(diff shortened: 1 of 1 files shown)


theano/gpuarray/type.py
```diff
@@ -476,10 +476,8 @@ def c_libraries(self):
         return ['gpuarray']
 
     def c_code_cache_version(self):
-        ver = pygpu.gpuarray.api_version()
-        # we only use the major version since the minor revision are
-        # API-compatible.
-        return (2, ver[0])
+        ver = pygpu.gpuarray.abi_version()
+        return (2,) + ver
```
abergeron: You should still use only the major since minor revisions are compatible.

notoraptor:
Updated to use only ABI major number.

But in that case the error still occurs (tested on my computer). Codes compiled in theano cache are directly linked to the library `libgpuarray.so.ABI_MAJOR.ABI_MINOR`. If the minor version changes, the c_code_cache_version() will still return the same tuple, so the cached file will still be used, but the library will have changed to `libgpuarray.so.ABI_MAJOR.ABI_NEW_MINOR`, so the cached code won't be able to import the library it expects.

theano/gpuarray/type.py
```diff
@@ -476,9 +476,9 @@ def c_libraries(self):
         return ['gpuarray']
 
     def c_code_cache_version(self):
-        ver = pygpu.gpuarray.api_version()
-        # we only use the major version since the minor revision are
-        # API-compatible.
+        # import pdb; pdb.set_trace()
```
abergeron: Please remove the comment

abergeron:
If we still have a problem (possibly because the linux linker is dumb), then let's just use the full number.  It will only cause possibly too many recompilations.

notoraptor:
Ultimately it seems there is no more changes to do here (we still use major version), job is done in libgpuarray.

Now the fix works, either when the minor version changes (no recompilation) or when the major version changes (recompilation).

abergeron:
jenkins test this please

PR #4765, opened 2016-07-19
Reduce C code generated when using GpuDnnConvDesc

gokul-uf:
This is a continuation of PR #4638. I've made the commits a bit more clean.
TODO: to what number should I increment the c_cache_version to?

nosetests theano/gpuarray/test_dnn.py fails with the following output.
 (a part of it)

```
======================================================================
ERROR: theano.gpuarray.tests.test_dnn.test_dnn_conv_grad
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/gokul/Theano/theano/gpuarray/tests/test_dnn.py", line 763, in test_dnn_conv_grad
    utt.verify_grad(dconv, [img_val, kern_val, out_val])
  File "/home/gokul/Theano/theano/tests/unittest_tools.py", line 91, in verify_grad
    T.verify_grad(op, pt, n_tests, rng, *args, **kwargs)
  File "/home/gokul/Theano/theano/gradient.py", line 1654, in verify_grad
    o_fn = function(tensor_pt, o_output, name='gradient.py fwd')
  File "/home/gokul/Theano/theano/gradient.py", line 1635, in function
    on_unused_input='ignore', name=name)
  File "/home/gokul/Theano/theano/compile/function.py", line 322, in function
    output_keys=output_keys)
  File "/home/gokul/Theano/theano/compile/pfunc.py", line 480, in pfunc
    output_keys=output_keys)
  File "/home/gokul/Theano/theano/compile/function_module.py", line 1784, in orig_function
    defaults)
  File "/home/gokul/Theano/theano/compile/function_module.py", line 1648, in create
    input_storage=input_storage_lists, storage_map=storage_map)
  File "/home/gokul/Theano/theano/gof/link.py", line 699, in make_thunk
    storage_map=storage_map)[:3]
  File "/home/gokul/Theano/theano/gof/vm.py", line 1042, in make_all
    no_recycling))
  File "/home/gokul/Theano/theano/gof/op.py", line 975, in make_thunk
    no_recycling)
  File "/home/gokul/Theano/theano/gof/op.py", line 875, in make_c_thunk
    output_storage=node_output_storage)
  File "/home/gokul/Theano/theano/gof/cc.py", line 1189, in make_thunk
    keep_lock=keep_lock)
  File "/home/gokul/Theano/theano/gof/cc.py", line 1130, in __compile__
[...]

PR #5402, opened 2017-01-12
Reduce C code generated when using GpuDnnConvDesc 

gokul-uf:
PR continuation of #4765 (that was stale). This PR reduces the code generated while using GpuDnnConvDesc by reusing the objects created.

### DIFF
--- a/theano/gpuarray/conv_desc.c
+++ b/theano/gpuarray/conv_desc.c
@@ -1,7 +1,7 @@
 #section support_code_apply
 
-int APPLY_SPECIFIC(conv_desc)(PyArrayObject *filt_shp, PyObject *conv_mode,
-                              PyObject *precision,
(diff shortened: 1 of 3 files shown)


gokul-uf:
@nouiz @abergeron please review, thanks!

abergeron:
This has conflicts with the current master and the following issues:

 - You hardcode npy_int64/npy_int8 as the type for the additional inputs, but nothing is done from the python side to ensure that those inputs are actually of that type.
 - Using magic values to communicate between the python and the C is fragile and prone to subtle breakage with future changes
 - This changes the signature everywhere and breaks outside code that was using the Op, which we would like to avoid

In addition to that we have another approach on the way that would pass all of this information using op params by bundling the information in a struct.  This is still a little rough, but is the future direction we want to take with these things.

So for all these reasons, this PR will not be merged.

PR #5405, opened 2017-01-12
Fix shapes for cudnn bidirectional rnn since the documentation is full of lies.

abergeron:


nouiz:
Can you link to the report of this error?

Issue #5396, opened 2017-01-10
Make Theano support numpy 1.12


nouiz:
See: https://github.com/numpy/numpy/issues/8420#issuecomment-270269649

Many Theano tests are failing with numpy 1.12 rc.

In summary, there seem to be 4 types of errors:

- [ ] TypeError: 'float' object cannot be interpreted as an index errors are from using floats to index arrays, it can probably be fixed in the test by making the dimensions in out_shp integer.
- [x] TypeError: Cannot cast array data from dtype('float32') to dtype('int64') according to the rule 'safe' can be fixed by making the cast explicit, casting="unsafe".
- [ ] TypeError: invalid strides looks to be because of relaxed stride checking.
- [x] That leaves the arctanh errors and I don't know what is going on there.
- [ ] [TypeError](https://travis-ci.org/Theano/Theano/jobs/196714103#L3203): 'numpy.float64' object cannot be interpreted as an index

gokul-uf:
@nouiz I'd like to work on this if no one's assigned

nouiz:
Maybe someone was working on the first point. So start by the second and third point.

thanks.

chinnadhurai:
I have been working on this as part of the CCW. With numpy version, 1.12.0rc2, I am able to reproduce issue 2. Issue 1 and 4 are not reproducible so far. So, I shall fix the errors for issue 2.

gokul-uf:
@chinnadhurai you could send a PR for the second issue alone. We could work on this together if you like :)

chinnadhurai:
@slefrancois and I fixed issue 2 this morning. Would you be interested in working on issue 3?

slefrancois:
I can't reproduce error 1, "float object cannot be interpreted as index".
Also, the arctanh error seems to be a problem with an old glibc.
So we recommend fixing only points 2-3 for now.

gokul-uf:
@chinnadhurai, sure!

slefrancois:
invalid strides relate to issue #1586

slefrancois:
We're going to do a final test by forcing Travis to use numpy 1.12 before closing.

chinnadhurai:
numpy 1.12 requires python version 3.4 or higher (issue #4507)

rebecca-palmer:
[...]

PR #5406, opened 2017-01-13
numpy 1.12.0rc2 compatibility fix related to variable typecasting

chinnadhurai:
Fix for TypeError: Cannot cast array data from dtype('float32') to dtype('int64') according to the rule 'safe' can be fixed by making the cast explicit, casting="unsafe".

related to [5396](https://github.com/Theano/Theano/issues/5396)

### DIFF
--- a/theano/sparse/sandbox/sp2.py
+++ b/theano/sparse/sandbox/sp2.py
@@ -5,6 +5,7 @@ import theano
 import scipy.sparse
 
 from theano import gof, tensor
+from theano.tensor import discrete_dtypes
 from theano.tensor.opt import register_specialize
 from theano.sparse.basic import (
     as_sparse_variable, SparseType, add_s_s, neg,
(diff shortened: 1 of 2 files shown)


theano/sparse/sandbox/sp2.py
```diff
@@ -116,6 +117,11 @@ def make_node(self, n, p, shape):
         n = tensor.as_tensor_variable(n)
         p = tensor.as_tensor_variable(p)
         shape = tensor.as_tensor_variable(shape)
+
+        assert n.dtype in discrete_dtypes
+        assert p.dtype not in discrete_dtypes
```
slefrancois: p is a probability, shouldn't it be a float?

slefrancois:
jenkins test this

theano/sparse/sandbox/sp2.py
chinnadhurai: yeah..so, change "assert p.dtype not in discrete_dtypes" to "assert p.dtype in float_dtypes" ?

slefrancois: Let's try that, it will avoid user errors.

chinnadhurai: sure, committed the changes.. also confirmed that the nosetests still pass

slefrancois:
jenkins test this

slefrancois:
@nouiz , looks good to me once jenkins passes, can you double-check?

nouiz:
ok for me too.

PR #5407, opened 2017-01-13
Add theano.gpuarray.use similar to theano.sandbox.cuda.use() to manua…

nouiz:
…lly enable it.

### DIFF
--- a/theano/gpuarray/__init__.py
+++ b/theano/gpuarray/__init__.py
@@ -126,14 +126,6 @@ def use(device,
         import theano.compile
         theano.compile.shared_constructor(gpuarray_shared_constructor)
 
-    from .basic_ops import (GpuAlloc, GpuAllocEmpty, GpuContiguous, GpuEye,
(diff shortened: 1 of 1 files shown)


theano/gpuarray/__init__.py
```diff
@@ -110,15 +110,28 @@ def init_dev(dev, name=None):
 # This maps things like 'cuda0' to the context object on that device.
 init_dev.devmap = {}
 
+
+def use(device,
+        force=False,
+        default_to_move_computation_to_gpu=True,
+        move_shared_float32_to_gpu=True):
```
abergeron: This doesn't only move float32

theano/gpuarray/__init__.py
```diff
@@ -110,15 +110,28 @@ def init_dev(dev, name=None):
...
+    if device:
+        init_dev(device)
+    elif forece:
```
abergeron: forece -> force

nouiz:
jenkins test this

PR #5411, opened 2017-01-13
This fixes a bug in cusolver solve op when m > k

tfjgeorge:
@abergeron


nouiz:
It would be great to add a test for that case.

nouiz:
jenkins test this

slefrancois:
this is superseded by #5421

PR #5412, opened 2017-01-13
Fix theano scripts on windows

### LINKED ISSUE #5308: Make `theano-*` scripts callable on Windows
This should use the entry_points/console_scripts feature of setup.py, so the executables will be created when running pip install, for instance.

This has been done in Fuel in https://github.com/mila-udem/fuel/pull/211 for an example.

notoraptor:
Use the entry_points/console_scripts feature of `setup.py` to make theano scripts installed when running `pip install`.

These step are done to fix #5308. The bin folder containing the scripts is now importable from Python, with each script embedding its code into a `main()` function specified as the entry point of the script in `setup.py`.

We have also reported on issue #5308 a Windows-specific error related to theano-cache: the cache folder cannot be deleted because it contains some shared libraries that are always loaded at theano importation and stay opened on Windows, which forbide the deletion of these shared library files in the cache.

The workaround chosen is to unset the theano config flag `cxx` only on Windows platform and only in theano-cache script. Doing that, Theano will not try to compile any C code, and will also not try to load compiled C codes that are in the cache, allowing these files to be deleted.

To ensure this workaround does work, we also ensure that lazylinker module will be also ignored from the cache when `cxx` is unset. Previously, if cxx was unset, Theano would not compile lazylinker but would look for an already compiled version of this module in the cache.

The installation (`pip install -e .`) works on my computer (with conda, tested with Python 2.7.13) and the installed theano scripts (now with `.exe` extension) also.

Can anyone else test this PR on a windows machine ?

@nouiz @lamblin @abergeron


setup.py
```diff
@@ -173,7 +173,11 @@ def do_setup():
               'theano.misc': ['*.sh'],
               'theano.d3viz' : ['html/*','css/*','js/*']
           },
-          scripts=['bin/theano-cache', 'bin/theano-nose', 'bin/theano-test'],
+          entry_points={
+              'console_scripts': ['theano-cache = bin.theano_cache:main',
+                                  'theano-nose = bin.theano_nose:main',
+                                  'theano-test = bin.theano_test:main']
```
lamblin: Does that actually work when the current directory is not `Theano`?
I thought that `bin` here was interpreted as a package name, and that we would have to move the files inside of `Theano/theano`, so we could use `theano.<...>:main`.

notoraptor: @lamblin Yes, it works (on my computer, at least)!

It's a bit weird, but when I put the scripts into theano folder, theano is fully imported (from `theano/__init__.py`) before the code of the scripts is loaded, then THEANO_FLAGS is parsed during this importation and no more into the script code. However I need to modify THEANO_FLAGS before it is parsed, to add `cxx=` to it, specifically for `theano-cache` script. I have not be able to do it into the theano folder.

THEANO_FLAGS is modified at the top of the `theano-cache` script before the `import theano` line.

notoraptor:
Some codes in Theano look for `bin/theano-*` scripts (e.g. `bin/theano-nose` in jenkins files). I'm working to fix it. If possible, I'll just add some kinds of links or wrappers into `bin` folder to minimize the changes in the whole theano code.

notoraptor:
Updated.

notoraptor:
Updated.

theano/gof/vm.py
```diff
@@ -656,6 +656,8 @@ def __call__(self, output_subset=None):
 
 
 try:
+    if not theano.config.cxx:
+        raise theano.gof.cmodule.MissingGXX('lazylinker will not be imported if theano.config.cxx is not set.')
```
nouiz: Add a comment why you do this.

notoraptor: Done!

abergeron:
I'll have to look into all the failing tests, but it does seem unrelated.  There are so many failures to go through, it'll take me a while.

notoraptor:
@abergeron on my computer on linux I got some same errors as jenkins with this PR, but when I merge this PR with the theano/master branch in a new branch, the errors do not more occur (I have just checked 2-3 errors [e.g. `theano.gpuarray.tests.test_opt.test_reduce`], not all!).

So is it maybe because this current PR is out of date for the Gpuarray code ?

abergeron:
We can always try this: 

jenkins test this please

setup.py
lamblin: That's really nice that you found that way of bypassing the import of `theano`, I thought it would have been necessary.
Thanks!

Issue #5414, opened 2017-01-16
comment typo


cosmozhang:
A typo in the comment "Sequence is shorter then the required". It should be "Sequence is shorter than the required"

rishikksh20:
Need to change here : 
https://github.com/Theano/Theano/blob/master/theano/scan_module/scan_perform.pyx#L220

wazeerzulfikar:
Hi, I am new to Theano and really interested in making a first contribution. I would like to work on this issue.

rishikksh20:
@cosmozhang close this issue.

PR #5415, opened 2017-01-16
Typo-Fix of value error comment

wazeerzulfikar:
Fixed a typo of value error comment. Fixes issue #5414


nouiz:
thanks. @rishikksh20 the second change is a pep8 fix.

PR #5417, opened 2017-01-16
make tensor.join() return the input when there is only 1 variable to join.

### LINKED ISSUE #4897: local_useless_split considered harmful
Hi,

I noticed a slowdown in my code on the GPU between 0.7 and 0.8.  When I look at the compilation graph it turns out that some GpuFromHost nodes that should have been merged are no longer being merged.  To make a long story short (long story available on request) the reason for this is that the input to these nodes is an Assert node with a rather unhelpful message:

```
 | |   |   | | | | | |GpuFromHost [id IL] ''   254
 | |   |   | | | | | | |Assert{msg='Theano Assert failed!'} [id IM] ''   238
(a bunch of stuff...) 221
 | |   |   | | | | | |   |Elemwise{eq,no_inplace} [id IS] ''   232
 | |   |   | | | | | |     |Shape_i{1} [id IT] ''   227
 | |   |   | | | | | |     | | (my stuff)   221
 | |   |   | | | | | |     |Shape_i{1} [id IU] ''   14
 | |   |   | | | | | |       |W_uc0 [id BN]
```

As noted elsewhere, Assert nodes have the unpleasant property of blocking optimizations - and though there seems to be a local optimization that hoists them above HostFromGpu nodes, obviously we can't do that for GpuFromHost nodes.

It turs out that these Assert nodes were introduced in revision 83d458b5 from #4440 - at least, they should have a meaningful error message, but also, leaving in the "useless" Spilt nodes (which are actually the derivatives of equally "useless" Join nodes) doesn't cause a performance hit, while these Assert nodes make my code up to 40% slower!

In order to avoid creating "useless" Split nodes I have to put extra code that checks before calling theano.tensor.concatenate() that its input isn't a list of a single item.  If we still think that local_useless_split is a useful optimization, maybe there is some way to make concatenate() avoid creating useless Join/Split nodes in the first place?

ReyhaneAskari:
fix #4897 , join function is added so that we can check the number of tensors that are to concatenate. The test is also added which tests two cases. One where there are several tensors for concatenation and the other with only one tensor.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4163,6 +4163,7 @@ class Join(Op):
 
 
 join_ = Join()
+pprint.assign(Join, printing.FunctionPrinter('join'))
 
 
 def join(axis, *tensors_list):
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4170,6 +4170,10 @@ def join(axis, *tensors_list):
     """
     Convenience function to concatenate `TensorType`s along the given axis.
 
+    This function will not add the op in the graph when it is not useful.
+    For example, in the case that the list of tensors to be concatenated
+    is one, it will just return the tensor.
+
     Parameters
     ----------
     tensors : list of tensors (or list-like)
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -4162,7 +4162,11 @@ class Join(Op):
         return [tuple(out_shapes)]
 
 
-"""
+join_ = Join()
+
+
+def join(axis, *tensors_list):
+    """
     Convenience function to concatenate `TensorType`s along the given axis.
 
     Parameters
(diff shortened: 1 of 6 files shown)


theano/tensor/basic.py
```diff
@@ -4181,12 +4186,11 @@ def infer_shape(self, node, ishapes):
         former case, the axis is fixed at construction, while in the
         latter it may vary over time depending on the value of the
         `axis` variable.
```
nouiz: Add a note section, that this function won't add the op in the graph when it is not useful.

theano/tensor/tests/test_basic.py
```diff
@@ -4372,6 +4372,29 @@ def test_join_inplace():
...
+    join_0 = theano.tensor.concatenate([x_0], axis=1)
+    join_1 = theano.tensor.concatenate([x_0, x_1, theano.tensor.shape_padright(x_2)],
+                                       axis=1)
+    f = theano.gof.FunctionGraph([x_0], [join_0])
+    g = theano.gof.FunctionGraph([x_0, x_1, x_2], [join_1])
+
+    assert isinstance(g.toposort()[1].op, Join)
+    assert not f.toposort()
+    assert join_0.ndim is 2
+    assert join_1.ndim is 2
```
nouiz: You can simplify the test by not building a functiongraph.

Just do:

asset join_0 is x_0
assert join_1 is not x_0

PR #5419, opened 2017-01-16
Add a deprecation message to the old backend.

abergeron:


theano/sandbox/cuda/__init__.py
```diff
@@ -488,6 +488,8 @@ def use(device,
 
     """
     global cuda_enabled, cuda_initialization_error_message
+    _logger.warn("The cuda backend is deprecated and will be removed in "
+                  "the next release.  Please switch to the gpuarray backend.")
```
nouiz: Can you add a link to the doc about that? What about this URL: 

https://github.com/Theano/Theano/wiki/Converting-to-the-new-gpu-back-end%28gpuarray%29

Issue #5397, opened 2017-01-10
Add Cholesky decomposition for GpuSolve


nouiz:

slefrancois:
For solving linear equation systems with Hermitian matrices, we can use the Cholesky decomposition to get a 2x speedup over LU decomposition.

Add a case in gpuarray.linalg.gpu_solve to use Cholesky decomposition provided by cusolver.

tfjgeorge:
PR #5421

PR #5421, opened 2017-01-16
Cusolver using Cholesky decomposition for symmetric matrices

tfjgeorge:
@slefrancois 

I had to add a function similar to the one provided by scikit-cuda to wrap the call to cusolverDnSpotrs from libcusolver.

Maybe @abergeron should also take a look at this ?

For reference here are some timings for 2000 calls to solve with A with shape (150, 150) and b with shape (150, 5) (time in seconds)

    ('symmetric', 1.2224421501159668)
    ('general', 3.972813129425049)

### DIFF
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import pkg_resources
 import theano
+import warnings
 
 from theano import Op
 from theano.gpuarray import basic_ops, GpuArrayType
@@ -16,6 +17,7 @@ except ImportError:
 
 cusolver_available = False
 try:
+    import skcuda
     from skcuda import cusolver
     cusolver_available = True
 except (ImportError, OSError, RuntimeError, pkg_resources.DistributionNotFound):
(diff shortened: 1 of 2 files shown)


slefrancois:
jenkins test this

slefrancois:
fixes #5397

slefrancois:
Looks good at a glance, but I would like a review from @abergeron

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
@tfjgeorge, please check the jenkins errors on the custom cusolver objects.

tfjgeorge:
I am using the github version of scikit cuda and jenkins is probably using
the last release. To fix this I could add the missing method directly in
linalg.py as I did for cusolverDnSpotrs

slefrancois:
That's a good point, I actually encountered this problem, the current scikit-cuda release is missing references to cusolver in CUDA 8.0. I'll just install the dev version of scikit, we shouldn't change our code for a version incompatibility.

Maybe put a check where it warns if the old scikit is present is needed for the custom routine.

slefrancois:
jenkins test this

slefrancois:
The code looks good, but I recommend we add a version check for scikit-cuda which warns that CUDA 8 requires scikit-cuda >= 0.5.2.

tfjgeorge:
@slefrancois I just added the required warning

slefrancois:
looks good, jenkins test this

PR #5423, opened 2017-01-16
Mixed stuff

nouiz:


theano/scalar/basic.py
```diff
@@ -277,10 +277,35 @@ def convert(x, dtype=None):
...
+            x_ = theano._asarray(x, dtype=dtype)
+            if numpy.all(x == x_):
+                break
+            x_ = None
+        assert x_ is not None
+        return ScalarConstant(get_scalar_type(str(x_.dtype)), x, name=name)
+    if isinstance(x, builtin_complex):
+        # TODO: We have added the complex type, so this should be tested
+        raise NotImplementedError()
+    raise TypeError(x)
```
abergeron: Why do you rollback the move to convert?

theano/sandbox/cuda/opt.py
```diff
@@ -192,7 +192,7 @@ def apply(self, fgraph):
             # This happen frequently as we do 2 pass of the gpu optimizations
             if (len(input.clients) == 1 and
                 (input.clients[0][0] == 'output' or
-                 input.clients[0][0].op == gpu_from_host)):
+                 isinstance(input.clients[0][0].op, GpuFromHost))):
```
abergeron: Why all these changes to isinstance?

theano/scalar/basic.py
```diff
@@ -3838,30 +3868,56 @@ def __init__(self, inputs, outputs):
         # only 1 new Composite each time at the output.
         for i in inputs:
             assert i not in outputs  # This isn't supported, use identity
-        if len(outputs) > 1 or not any([isinstance(var.owner.op, Composite)
-                                        for var in outputs]):
+        if not any([isinstance(var.owner.op, Composite) for var in outputs]):
```
abergeron: While I'm all for the change to support multiple outputs, just before the release is not the right time.  Let's wait until after.

theano/tensor/basic.py
```diff
@@ -276,7 +276,9 @@ def constant(x, name=None, ndim=None, dtype=None):
     if (sig not in constant_cache and ret.data.size == 1 and
         (-10) <= ret.data <= 10 and
         (ret.dtype in int_dtypes or ret.dtype in uint_dtypes or
-         (ret.dtype in float_dtypes and int(ret.data) == ret.data))):
+         (ret.dtype in float_dtypes and
+          # Limit the size of the cache.
+          len(constant_cache) < 10000))):
```
abergeron: If we want to limit the size, a hard cutoff declaring that the first n are saved and the others aren't is a bad way to do it.  At least have some form of LRU or LFU.

theano/sandbox/cuda/opt.py
nouiz: isinstance() is faster

theano/tensor/basic.py
nouiz: I think a hard cut-off is better then the current state of no limit. Especially as the cut-off is high.

We can't make an ordinary LRU or LFU easily. We can only remove constant that aren't in the cache anymore. But I think this is overkill. This is just a cache for constant. There isn't good reason to have more then 10000 values. If that happen, then the user have very big graph and have others problems too.

theano/scalar/basic.py
nouiz: Ok.

theano/scalar/basic.py
nouiz: Good catch. I didn't do this in purpose. It come from a rebase. convert wasn't there when this was done. I'll update this.

PR #5427, opened 2017-01-18
Add user warning to prevent issue #5143

### LINKED ISSUE #5143: test_conv3d_bwd failing
I am getting 15-20 failing tests for bleeding edge w/ Ubuntu 16.04, Cuda 8.0.44, cuDNN 5105, and theano 0.9.0dev3.dev-e6a3b009bc4ce68fb5ba711d775689ff41d8a587. 

All of the failing tests are WrongValue errors in test_conv_3d_bwd. The number of failing tests varies from run to run. The mean abs differences are not small. 

Tests run with the command: `THEANO_FLAGS=device=cpu nosetests -sdv theano.sandbox.cuda.tests.test_dnn > test-errors.log 2>&1`

I have fast_math on in my theanorc but did not make a difference when I turned it off. GPU is a GTX 1080. GCC is 5.4. Happy to give you any additional information needed to diagnose.

Full text of errors attached: 

[test-errors.txt](https://github.com/Theano/Theano/files/553424/test-errors.txt)

notoraptor:
This does not fix #5143 but helps user prevent this issue.

For 3D convolutions, cuDNN currently provides 2 algorithms to compute backward filter step: algo 0 (`CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0`) and algo 3 (`CUDNN_CONVOLUTION_BWD_FILTER_ALGO_3`). In theano unit tests, when theano flag `dnn.conv.algo_bwd_filter` is set to `none`, then algo 0 is always chosen and produces the expected results. But for other values of this flag, which let cuDNN choose either the best or the fatest algorithm, then algo 3 may be selected. After investigation, it seems that algo 3 can produce values very far from the expected ones depending on the shapes of the input and the compute capability of the GPU used. For example, with compute capability of 3.0, 3.5 or 3.7, there is no error, but with compute capability of 5.2 or 6.1, the errors do occur.

The following table shows some results. Here's the theano test used:
```
THEANO_FLAGS=floatX=float32,dnn.conv.algo_bwd_filter=guess_on_shape_change nosetests -sdvx theano.sandbox.cuda.tests.test_dnn:test_conv3d_bwd
```
And here's the table:

| Compute capability | GPU Name                | Short name | Results for CUDNN 5105 |
|--------------------|-------------------------|------------|------------------------|
| 3.0                | GeForce GTX 660         | gtx660     | OK                     |
| 3.0                | GeForce GTX 680         | gtx680     | OK                     |
| 3.5                | GeForce GTX TITAN Black | titanblack | OK                     |
| 3.5                | Quadro K6000            | k6000      | OK                     |
| 3.7                | Tesla K80               | k80        | OK                     |
| 5.2                | GeForce GTX TITAN X     | titanx     | FAILED                 |
| 5.2                | Tesla M40 24GB          | m4024gb    | FAILED                 |
| 6.1                | TITAN X (Pascal)        | titanxp    | FAILED                 |

This fix just prints a warning when the user sets the theano flag `dnn.conv.algo_bwd_filter` to value other than `none` and tell him to set it to `none` if he encounters some unexpected results.

@nouiz @lamblin @abergeron


theano/gpuarray/dnn.py
```diff
@@ -1135,6 +1135,11 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1),
                                     desc_op.subsample)
     out_shp = assert_conv_shape(out_shp)
     out = gpu_alloc_empty(ctx_name, dtype=img.dtype)(*out_shp)
+    if config.dnn.conv.algo_bwd_filter != 'none':
+        warnings.warn('[new backend] cuDNN backward filter convolution computation for 3D convolutions '
```
nouiz: Remove the "[new backend]" part of the message and idem in the old back-end. It don't help user.

Can you be more specific? As the error happen only for subsample on the axis that cause problems, can you check that?

Also, the check should be moved to the make_node of the op. Otherwise, user won't get the warning then they build the graph on the CPU and let the optimizer move it to the GPU.

notoraptor: It seems it indeed depends on the subsample (when input shape, filter shape, border mode [=full] and conv mode [='conv'] are constant but subsample changes). But I can't find a subsample pattern that is always associated to an error, so it is currently difficult to tell which kind of subsamples can produce an error with algo 3.

abergeron:
Didn't we find that (1, 2, 1) as a subsample pattern would cause the bug?

theano/gpuarray/dnn.py
abergeron: Didn't we find that (1, 2, 1) as a subsample pattern would cause the bug?

notoraptor: Yes, but it is not the only one. Some other patterns also cause the bug.

abergeron: Does (1, 1, 1) causes problems?  If not, we can warn when there is subsampling.

notoraptor: No, (1,1,1) does not cause problems.

PS: As recall, I test specific values with these inputs:
```
run_conv3d_bwd(inputs_shape, filters_shape, subsample, border_mode, conv_mode)
run_conv3d_bwd((8, 1, 20, 12, 15), (5, 1, 6, 6, 4), (1, 1, 1), 'full', 'conv')
```

I will check if (1,1,1) is ok even with different values for other parameters (input shape, filter shape, border mode, conv mode).

notoraptor:
New update! Now the warning is printed only if subsample is not `(1,1,1)` for 3d conv backward filter operations. Warning is rewriten and moved into make_node() methods.



Some remarks after analysis:

While looking for subsamples I have remarked that some test_shapes are duplicated in the shapes tested by test_conv3d_bwd in both old and new backend, so I have removed the duplicates in a commit ( https://github.com/Theano/Theano/pull/5427/commits/5d72e1157df703eea202097fbcef2624beb391c4 ).

Then, when removing these duplicates I have also noticed that the amount of WrongValue errors was divided by 2. In fact, all errors in test_conv3d_bwd come from this test_shape:
```
input shape, filter shape, subsample:
[(8, 1, 20, 12, 15), (5, 1, 6, 12, 4), (3, 2, 1)]
```

I remark that it is the only one test_shape for which subsample fields are not the same (that is, either (1,1,1) or (2,2,2) or (3,3,3), etc.): https://github.com/Theano/Theano/blob/master/theano/gpuarray/tests/test_dnn.py#L972

@nouiz @abergeron @lamblin

teslaswqa:
The issue in cuDNN v5.1 has been identified now, you can expect the fix in the cudnn v6 GA release. 

Please also note that the test code is calling a deprecated API in cuDNN. That needs a change for v6.0

nouiz:
Thanks for the update. I made an issue about all cudnn v6 stuff that we need to do and added it there to test the fix and remove the warning:

https://github.com/Theano/Theano/issues/5734

PR #5434, opened 2017-01-19
mixed

nouiz:
subset of gh-5423 without the multi-output commit and the fix to the comments.


abergeron:
There are flake8 failures.

nouiz:
flake8 fixed

Issue #4734, opened 2016-07-13
EquilibriumOptimizer maxed out in some tests


abergeron:
This test causes the problem for 'local_dimshuffle_lift':
`theano.tensor.signal.tests.test_pool.py:TestDownsampleFactorMax.test_infer_shape`

This test causes the problem for 'local_upcast_elemwise_constant_inputs':
`theano.tensor.nnet.tests.test_corr.py:TestCorr2D.test_infer_shape_forward`

Sentient07:
Even `local_add_canonizer` has this problem. This is originating from this test, `theano.tensor.tests.test_opt.test_local_subtensor_merge`.

ReyhaneAskari:
`theano.tensor.signal.tests.test_pool.py:TestDownsampleFactorMax.test_infer_shape` is no longer giving the error but both `theano.tensor.nnet.tests.test_corr.py:TestCorr2D.test_infer_shape_forward` and `theano.tensor.tests.test_opt.test_local_subtensor_merge` work when:

**THEANO_FLAGS**="optdb.max_use_ratio=**6**"

nouiz:
Can you make a PR that change the default to 6 and make sure no other tests
raise error in that setup?

lamblin:
Closed by #5435.

PR #5435, opened 2017-01-20
changed optdb.max_use_ratio to 6

ReyhaneAskari:
Regarding issue #4734 , the default value for  `optdb.max_use_ratio`  is changed from 5 to 6 to prevent errors for some tests. For example this test:

```theano.tensor.tests.test_opt.test_local_subtensor_merge```

generated this error:

```ERROR (theano.gof.opt): EquilibriumOptimizer max'ed out by 'local_add_canonizer'. You can safely raise the current threshold of 5.000000 with the theano flag 'optdb.max_use_ratio'. ```

### DIFF
--- a/theano/gof/tests/test_opt.py
+++ b/theano/gof/tests/test_opt.py
@@ -572,6 +572,8 @@ class TestEquilibrium(object):
         assert str(g) == '[Op2(x, y)]'
 
     def test_low_use_ratio(self):
+        backup = theano.config.on_opt_error
+        theano.config.on_opt_error = 'warn'
         x, y, z = map(MyVariable, 'xyz')
         e = op3(op4(x, y))
         g = FunctionGraph([x, y, z], [e])
@@ -590,6 +592,7 @@ class TestEquilibrium(object):
                 max_use_ratio=1. / len(g.apply_nodes))  # each opt can only be applied once
             opt.optimize(g)
         finally:
+            theano.config.on_opt_error = backup
             _logger.setLevel(oldlevel)
         # print 'after', g
         assert str(g) == '[Op1(x, y)]'
--- a/theano/gof/tests/test_opt.py
+++ b/theano/gof/tests/test_opt.py
@@ -572,8 +572,6 @@ class TestEquilibrium(object):
         assert str(g) == '[Op2(x, y)]'
 
     def test_low_use_ratio(self):
-        backup = theano.config.on_opt_error
-        theano.config.on_opt_error = 'warn'
         x, y, z = map(MyVariable, 'xyz')
         e = op3(op4(x, y))
         g = FunctionGraph([x, y, z], [e])
@@ -583,6 +581,8 @@ class TestEquilibrium(object):
         _logger = logging.getLogger('theano.gof.opt')
         oldlevel = _logger.level
         _logger.setLevel(logging.CRITICAL)
+        backup = theano.config.on_opt_error
+        theano.config.on_opt_error = 'warn'
         try:
             opt = EquilibriumOptimizer(
                 [PatternSub((op1, 'x', 'y'), (op2, 'x', 'y')),
--- a/theano/gof/tests/test_opt.py
+++ b/theano/gof/tests/test_opt.py
@@ -571,6 +571,7 @@ class TestEquilibrium(object):
         opt.optimize(g)
         assert str(g) == '[Op2(x, y)]'
 
+    @theano.configparser.change_flags(on_opt_error='ignore')
     def test_low_use_ratio(self):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by nouiz** [comment]
This help, but don't remove all of them: http://darjeeling.iro.umontreal.ca:8080/job/Theano%20core/621/console

To find the missing one, add a commit with this diff in this PR. We will always want this new behavior, so keep it in the PR. Then run the CPU tests (Theano flag: force_device=True)
```
diff --git a/theano/gof/opt.py b/theano/gof/opt.py
index 12176e3..27163b9 100644
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2515,10 +2515,14 @@ class EquilibriumOptimizer(NavigatorOptimizer):
         end_nb_nodes = len(fgraph.apply_nodes)
 
         if max_use_abort:
-            _logger.error("EquilibriumOptimizer max'ed out by '%s'" % opt_name +
-                          ". You can safely raise the current threshold of " +
-                          "%f with the theano flag 'optdb.max_use_ratio'." %
-                          config.optdb.max_use_ratio)
+            msg = ("EquilibriumOptimizer max'ed out by '%s'" % opt_name +
+                   ". You can safely raise the current threshold of " +
+                   "%f with the theano flag 'optdb.max_use_ratio'." %
+                   config.optdb.max_use_ratio)
+            if theano.config.on_opt_err == 'raise':
+                raise AssertionError(msg)
+            else:
+                _logger.error(msg)
         fgraph.remove_feature(change_tracker)
         assert len(loop_process_count) == len(loop_timing)
         assert len(loop_process_count) == len(global_opt_timing)
```

**Comment by ReyhaneAskari** [comment]
Dear Fred (@nouiz ),
We still have these errors:
http://darjeeling.iro.umontreal.ca:8080/job/Theano%20core/627/console

What do you suggest? 

Thanks, 
Reyhane

**Comment by lamblin** [comment]
Can you add "on_opt_error=raise,on_shape_error=raise" to the Theano flags on Jenkins?
That way, Travis and Jenkins will behave the same way when there are optimization (or shape) errors.
I think you just have to update `.jenkins/jenkins_test{1,2}.sh`.

**Comment by lamblin** [comment]
For `test_low_use_ratio` [1], the test was designed to silence the warning, and check that the cycle was interrupted.
Now that it is an error (with `on_opt_error=raise`), we should also temporarily set `on_opt_error=warn` in that part of the test.

[1] https://github.com/Theano/Theano/blob/master/theano/gof/tests/test_opt.py#L574

**Comment by lamblin** [comment]
The remaining test in Travis is `test_infer_shape_forward (theano.tensor.nnet.tests.test_corr3d.TestCorr3D)`.
You can try to manually increase the ratio to 7 or 10 for that specific test, and run it manually (with `on_opt_error=raise`) to see if it passes. If it passes, we can raise it in the buildbot. If it does not, then we have to investigate if that is an actual cycle, and how to speed it up.

File: theano/gof/tests/test_opt.py [hunk 1]
```diff
@@ -572,6 +572,8 @@ def test_2(self):
         assert str(g) == '[Op2(x, y)]'
 
     def test_low_use_ratio(self):
+        backup = theano.config.on_opt_error
+        theano.config.on_opt_error = 'warn'
```
nouiz: Move that just before the Try. In case there is a crash in between.

ReyhaneAskari: Thanks. Done.

ReyhaneAskari:
For ```theano.tensor.nnet.tests.test_corr3d.TestCorr3D```,  `optdb.max_use_ratio = 8` works. Which other test do you mean?

theano/gof/tests/test_opt.py
nouiz: I mean this test. Which value of max_use_ratio is needed to don't have the error/warning?

ReyhaneAskari: It does not even pass with `max_use_ratio=100` :D

lamblin: @nouiz: This test was made explicitly so that max_use_ratio is small and the condition is triggered.
It used to test that this was not a failure (and silences the error message) and that the optimization process would move on.
Now that this is actually a failure if on_opt_error is "raise", I think we want to change the value of on_opt_error for that test, which is what I asked Reyhane to do.

@ReyhaneAskari I think "ignore" would actually be a better value for that test. There is also a decorator you can use instead of using the existing try / finally block: `@theano.configparser.change_flags(on_opt_error=ignore)`.

ReyhaneAskari: Done. Thanks.

PR #5436, opened 2017-01-20
Initialize the blas kernels when creating the contexts.

abergeron:


abergeron:
The removes the need for https://github.com/lisa-lab/DeepLearningTutorials/pull/182/files

nouiz:
ok to merge when jenkins pass

Issue #2762, opened 2015-04-14
Setting parameter with default value while missing setting obligatory parameters changes the default value


jimmycallin:
I'm finding it hard to explain, so here's an example. Notice that in 77, the output is faulty. This is because I changed the default value in 76 without setting the compulsory parameters. This is returned to normal the next time you run the code (78).

```
In [72]: a, b = T.dscalars('a', 'b')
In [73]: c = a + b

In [74]: f = theano.function([Param(a, name='first'), Param(b, default=1, name='second')], c)

In [75]: f(first=1)
Out[75]: array(2.0)

In [76]: f(second=2)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-76-091a89f42488> in <module>()
----> 1 f(second=2)

/Users/jimmy/dev/anaconda/envs/py2/lib/python2.7/site-packages/theano/compile/function_module.pyc in __call__(self, *args, **kwargs)
    581                     raise TypeError("Missing required input: %s" %
    582                                     getattr(self.inv_finder[c], 'variable',
--> 583                                             self.inv_finder[c]))
    584                 if c.provided > 1:
    585                     raise TypeError("Multiple values for input: %s" %

TypeError: Missing required input: a

In [77]: f(first=1)
Out[77]: array(3.0)

In [78]: f(first=1)
Out[78]: array(2.0)
```

nouiz:
thanks for the report.

Did you found it by change and do not get bottered by this? It is a very
unfrequent case. But we still need to fix it.

jimmycallin:
It was purely out of luck (?), although it happened five minutes into me learning Theano (going through the tutorial), so it might cause confusion among newbies like me.

nouiz:
Can you tell us which tutorial pages and which section?

jimmycallin:
Here: 
http://deeplearning.net/software/theano/tutorial/examples.html#setting-a-default-value-for-an-argument

To be clear, there is nothing wrong with the tutorial. I was just trying out different things to see what would happen when removing non-default parameters etc., which is how I discovered the bug.

affanv14:
@nouiz I'd like to take on this issue

PR #5442, opened 2017-01-23
Fix bug caused when setting default value in Param in interactive mode

affanv14:
Fixes https://github.com/Theano/Theano/issues/2762

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -810,6 +810,7 @@ class Function(object):
                             e.args = ("Bad input " + argument_name + " to " +
                                       function_name + " at index %d (0-based). %s"
                                       % (i, where),) + e.args
+                        restore_defaults()
                         raise
                 s.provided += 1
                 i += 1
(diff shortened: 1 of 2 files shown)


affanv14:
@nouiz Since theano.Param is  deprecated.Should i write a test for theano.In instead?

affanv14:
@nouiz I think this should do.Though i couldn't come up with a better commit message.

theano/compile/function_module.py
```diff
@@ -853,14 +859,17 @@ def __call__(self, *args, **kwargs):
         if not self.trust_input:
             for c in self.input_storage:
                 if c.required and not c.provided:
+                    restore_defaults()
```
nouiz: I think line 813 need a call like this before the raise. We catch and extend an exception.
I think line 898 also need a call to this. But for that one, it would be better to add the call just after the "except" as the other branch will raise the error itself and won't go up to line 898.

Can you finish that or we let this for another PR?

affanv14:
I'm on it

affanv14:
@nouiz I don't think we need a call at line 813.Since the arguments are only set from line 818 onwards.

nouiz:
I think that line 789 do the affectation. The line around 818 are only for the kwargs. If the user pass them by position and not keywords, the storage is set earlier.

affanv14:
oh..okay.Let me amend the last commit

nouiz:
thanks

Issue #5224, opened 2016-11-14
Scan: broadcast pattern bug


f0k:
This isn't very thoroughly examined, so I'm not sure about the title, but the symptoms are reproducible for quite a while (several months). They manifest when using an LSTM or GRU layer with a single unit (but not for a vanilla recurrent layer, or for multiple units).
Using bleeding-edge Lasagne:
```python
import theano
import theano.tensor as T
import lasagne
x = T.tensor3('x')
network = lasagne.layers.InputLayer((10,20,30), input_var=x)
network = lasagne.layers.GRULayer(network, num_units=1)
network = lasagne.layers.DenseLayer(network, num_units=2)
theano.grad(None, wrt=x, known_grads={lasagne.layers.get_output(network): T.matrix('y')})
```
This gives `TypeError: Cannot convert Type TensorType(float32, 3D) (of Variable Subtensor{:int64:}.0) into Type TensorType(float32, (False, False, True)). You can try to manually convert Subtensor{:int64:}.0 into a TensorType(float32, (False, False, True)).`
For a stack trace, see https://github.com/Lasagne/Lasagne/issues/730#issuecomment-249885421.

nouiz:
https://github.com/Theano/Theano/issues/5384#issuecomment-273493711 is related (same error message) I think the cause is the same, but it should be verified. In the comment I linked above, I said what I think should be the way forward. That issue have an example that don't depend on lasagne, so would be a better starting point for a test then the example in this issue.

nouiz:
After a long discussion with @lamblin it is not good in the case the user did an error it its inner function to post pone the error. This case happen frequently enough that we don't want to make this case more complicated, so we won't be more tolerant about such broadcast pattern difference.

But we should add a much better error message. Mostly, in scan, around that call to filter_variable, make a try...except. If the ndim and dtype are the same, but not the broadcastable, extend/add to the current error message.

Make 2 new error message. One if the inner function add an extra broadcast pattern, one if it remove it. Both case can happen in the case error as different dimensions could hit different case.

f0k:
> #5384 (comment) is related (same error message) I think the cause is the same, but it should be verified.

I see, thank you for the hint! So you're saying the error may be that the output_info given to scan has a different broadcast pattern than what's returned by the inner loop function. I'll investigate.

[...]

PR #5443, opened 2017-01-23
Added broadcast check in scan.

### LINKED ISSUE #5384: Theano TypeError message far from useful
Currently I have encountered a TypeError in using T.grad to compute a gradient regarding outputs of a T.scan.
Most debugging methods provided by Theano involves with compiling the function, so I'm really out of choice right now. The error is:
TypeError: Cannot convert Type TensorType(float64, 3D) (of Variable Subtensor{:int64:}.0) into Type TensorType(float64, (False, True, False)). You can try to manually convert Subtensor{:int64:}.0 into a TensorType(float64, (False, True, False)).

Let's be serious, this kind of error message is utterly junk.  The suggestion _You can try to manually convert Subtensor{:int64:}.0 into a TensorType(float64, (False, True, False))_ is obviously useless, who knows where did that Subtensor comes from? 
It's not hard to catch this TypeError in theano calling stack(say, catch it in scan-op) and provide more useful messages. 
Theano should provide more information in exception handling.

Thrandis:
Adresses #5224 and fix #5384

I added an error message (it is still unclear for the moment), and a small test with the failure case.

### DIFF
--- a/theano/scan_module/scan_op.py
+++ b/theano/scan_module/scan_op.py
@@ -422,7 +422,8 @@ class Scan(PureOp):
                    "axis %d in `output_info`. This can happen if one of the "
                    "dimension is fixed to 1 in the input, while it is still "
                    "variable in the output, or vice-verca. You have to make "
-                   "them consistent.")
+                   "them consistent, e.g. using theano.tensor."
+                   "{patternbroadcast,unbroadcast,addbroadcast}.")
             size = min(len(v1.broadcastable), len(v2.broadcastable))
             for n, (b1, b2) in enumerate(zip(v1.broadcastable[-size:],
                                              v2.broadcastable[-size:])):
--- a/theano/scan_module/scan_op.py
+++ b/theano/scan_module/scan_op.py
@@ -402,6 +402,32 @@ class Scan(PureOp):
                     'by using dimshuffle or shape_padleft. '
                     )
 
+        def _check_broadcast(v1, v2):
+            """ Checks that the broadcast pattern of v1 and v2.
+
+            Controls that the broadcast pattern of the variable provided as
+            input to `scan` matches the broadcast pattern provided in
+            `output_info`. It raises an error when they don't match. The
+            typical case is when the user provides either the input or the
+            `output_info` (but not both) with a dimension fixed to 1,
+            which may wrongly be interpreted as broadcastable.
+
+            """
+            msg = ("The broadcast pattern of the output of scan (%s) is "
+                   "inconsistent with the one provided in `output_info` "
+                   "(%s). The output on axis %d is `%r`, but it is `%r` on "
+                   "axis %d in `output_info`. This can happen if one of the "
+                   "dimension is fixed to 1 in the input, while it is still "
+                   "variable in the output, or vice-verca. You have to make "
+                   "them consistent.")
+            size = min(len(v1.broadcastable), len(v2.broadcastable))
+            for n, (b1, b2) in enumerate(zip(v1.broadcastable[-size:],
+                                             v2.broadcastable[-size:])):
+                if b1 != b2:
+                    a1 = n + size - len(v1.broadcastable) + 1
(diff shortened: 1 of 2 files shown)


Thrandis:
Ok there is some failed tests I need to look at!

theano/scan_module/scan_op.py
```diff
@@ -402,6 +402,32 @@ def make_node(self, *inputs):
                     'by using dimshuffle or shape_padleft. '
                     )
 
+        def _check_broadcast(v1, v2):
+            """ Checks that the broadcast pattern of v1 and v2.
```
nouiz: remove "that"

theano/scan_module/scan_op.py
```diff
@@ -402,6 +402,32 @@ def make_node(self, *inputs):
...
+
+            Controls that the broadcast pattern of the variable provided as
+            input to `scan` matches the broadcast pattern provided in
+            `output_info`. It raises an error when they don't match. The
+            typical case is when the user provides either the input or the
+            `output_info` (but not both) with a dimension fixed to 1,
+            which may wrongly be interpreted as broadcastable.
```
nouiz: The method format() is call for other thing then output_info. I think this method _check_broadcast should not be called inside format(), but before the format() call for the handling of output info.

theano/scan_module/scan_op.py
```diff
@@ -402,6 +402,32 @@ def make_node(self, *inputs):
...
+
+            """
+            msg = ("The broadcast pattern of the output of scan (%s) is "
+                   "inconsistent with the one provided in `output_info` "
+                   "(%s). The output on axis %d is `%r`, but it is `%r` on "
+                   "axis %d in `output_info`. This can happen if one of the "
+                   "dimension is fixed to 1 in the input, while it is still "
+                   "variable in the output, or vice-verca. You have to make "
+                   "them consistent.")
+            size = min(len(v1.broadcastable), len(v2.broadcastable))
```
nouiz: Some of the errors are due to the fact that this check was done on other stuff then the output_info, including non_sequence that can be anything. This mean v1 and v2 do not always have broadcastable. I don't remember if moving the check to only output_info will fix the problem or not. In all cases, skiping the check if the broadcastable attribute isn't there would fix this.

Thrandis: So I changed the check to be performed only if both variables have the `broadcastable` attribute, but I left it inside the `format` function.

theano/scan_module/scan_op.py
Thrandis: (see other comment)

theano/scan_module/scan_op.py
nouiz: Why you didn't move it? It could raise a bad error message in other cases.

Thrandis:
I moved the check where it needed to be.

theano/scan_module/scan_op.py
Thrandis: Done!

Thrandis:
I am wondering if my error message is clear enough! What do you think?

theano/scan_module/scan_op.py
```diff
@@ -402,6 +402,35 @@ def make_node(self, *inputs):
...
+            if (not hasattr(v1, 'broadcastable') and
+                    not hasattr(v2, 'broadcastable')):
+                return
+            msg = ("The broadcast pattern of the output of scan (%s) is "
+                   "inconsistent with the one provided in `output_info` "
+                   "(%s). The output on axis %d is `%r`, but it is `%r` on "
+                   "axis %d in `output_info`. This can happen if one of the "
+                   "dimension is fixed to 1 in the input, while it is still "
+                   "variable in the output, or vice-verca. You have to make "
+                   "them consistent.")
```
nouiz: Just add something like this to help guide the user on how to make them consistent:

    Using theano.tensor.{patternbroadcast,unbroadcast,addbroadcast} will help you do this.

PR #5445, opened 2017-01-25
Fix jenkins in 32bit

nouiz:
uint8 can't store the value of gamme for input greater then 6.


nouiz:
as this don't depend on gpu, I merge now.

PR #4973, opened 2016-09-14
Updating documentation on shared variables

317070:
You can make a pull request to Theano github repo in order to fix a bug or add new features.
Once a pull request is submitted, [Travis](https://travis-ci.org/) is launched to verify the pull request
given the current stack of unit-tests in Theano.

Updating documentation on shared variables.

PR #5447, opened 2017-01-25
Fix doc

ReyhaneAskari:
@nouiz ,

fix #4973, the old doc string was removed to allow the generation of the documentation.

### DIFF
--- a/theano/compile/sharedvalue.py
+++ b/theano/compile/sharedvalue.py
@@ -122,12 +122,7 @@ class SharedVariable(Variable):
 
         Notes
         -----
-        Prior to Theano 0.3.1, set_value did not work in-place on the GPU. This
-        meant that sometimes, GPU memory for the new value would be allocated
-        before the old memory was released. If you're running near the limits of
-        GPU memory, this could cause you to run out of GPU memory.
-
-        Beginning with Theano 0.3.1, set_value will work in-place on the GPU, if
+        Set_value will work in-place on the GPU, if
         the following conditions are met:
 
             * The destination on the GPU must be c_contiguous.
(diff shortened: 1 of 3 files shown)


theano/compile/sharedvalue.py
```diff
@@ -120,6 +120,36 @@ def set_value(self, new_value, borrow=False):
...
         this SharedVariable.
 
+        Notes
+        -----
+        Prior to Theano 0.3.1, set_value did not work in-place on the GPU. This
+        meant that sometimes, GPU memory for the new value would be allocated
+        before the old memory was released. If you're running near the limits of
+        GPU memory, this could cause you to run out of GPU memory.
+
+        Beginning with Theano 0.3.1, set_value will work in-place on the GPU, if
```
nouiz: @abergeron is this still true in the new back-end?

nouiz: As commented directly in the comment, remove all mention of theano 0.3.1. It is way too old.

abergeron: The new backend never moves data in-place when doing set_value on a shared variable.  But this could be added.

nouiz: I made an issue about this for this release. Do you have time today/tomorrow to do this?

theano/compile/sharedvalue.py
```diff
@@ -120,6 +120,31 @@ def set_value(self, new_value, borrow=False):
...
+              (which is a given for now, since only float32 is
+              supported).
+            * The old and new value must have the same shape.
+            * The old value is being completely replaced by the new
+              value (not partially modified, e.g. by replacing some
+              subtensor of it).
+            * You change the value of the shared variable via
+              set_value, not via the .value accessors. You should not
+              use the .value accessors anyway, since they will soon be
+              deprecated and removed.
```
nouiz: .value now raise an error. So remove this last bullet, thanks.

ReyhaneAskari: Done.

Issue #4455, opened 2016-05-02
Make GPUMultinomialFromUniform support multiple dtype


nouiz:
Follow up gh-4272.
- [x] make it support output dtype (odtype) != float32
- [ ] including float16
- [ ] support multiple input dtype.

PR #5451, opened 2017-01-25
multiple data type support for GPUMultinomialFromUniform output

aalmah:
This is to fix one issue in #4455, output should support float16 and input should support float16 and float64

### DIFF
--- a/theano/gpuarray/multinomial.py
+++ b/theano/gpuarray/multinomial.py
@@ -18,10 +18,12 @@ from .opt import register_opt, op_lifter, register_opt2
 from .type import GpuArrayType
 from .elemwise import GpuDimShuffle
 from theano.scalar import as_scalar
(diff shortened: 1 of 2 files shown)


nouiz:
travis fail.

theano/gpuarray/multinomial.py
```diff
@@ -60,6 +62,8 @@ def make_node(self, pvals, unis):
         return Apply(self, [pvals, unis], [out])
 
     def gpu_kernels(self, node, name):
+        out_ctype=pygpu.gpuarray.dtype_to_ctype(node.outputs[0].dtype)
+        write_out_ctype = write_w(node.outputs[0].dtype)
```
abergeron: You also need to use work_dtype for `current_out`.

abergeron:
jenkins test this please

Issue #4218, opened 2016-03-11
Numpy module imported twice in ```theano/tensor/nnet/abstract_conv.py``` and ```theano/tensor/nnet/abstract_conv.py```


Sentient07:
In the recent commit, [https://github.com/Theano/Theano/commit/5f4eab9240bfd5a23d04385312d3fa71e676817c](url) the two(above mentioned) modules have been reimported. Initially it was ```import numpy```, and this commit added a line, ```import numpy as np```. I observed this while rebasing one of my PR and am a little surprised as to how it passed travis.

nouiz:
our flake8 tests don't check for import itself. We use more numpy.* in
Theano code, but in many projects, people use np.*

I prefer np.* as this is shorter and consistent with the rest of numpy
software stack. But I'm more used to numpy.*.

Should we make an issue to convert all Theano to np.*? Is this worthwhile?

I don't like too much that one file use a mix of both...

abergeron:
https://github.com/python-rope/rope

nouiz:
We agreed to centralize on "import numpy as np". So this issue is about converting all Theano to that, as historically as used "import numpy".

kvmanohar22:
@nouiz  Is anyone working on this issue? If not I would like to fix this.

nouiz:
Nobody is working on this to my knowledge.

thanks.

Sentient07:
I told I'd work, but I have lots of tasks. So I think you might go ahead :)

PR #5452, opened 2017-01-25
import numpy (theano/compile)

bscellier:
#4218 
all files in the directory theano/compile have been changed


lamblin:
There are a couple of pep8 errors that appeared because the alignment has changed, but these are the only failing tests.

bscellier:
I've fixed these pep8 errors and nosetested flake8 before pushing the commit

PR #5454, opened 2017-01-26
Import numpy gof

bscellier:
#4218
all files in the directory theano/gof have been changed


patyork:
The change to `gof/cmodule.py` breaks the branch; `np.distutils` is not recognized whereas `numpy.distutils` was/is.

Issue #5055, opened 2016-10-05
Port ProfileMode GPU and scan extra information


nouiz:
PR https://github.com/Theano/Theano/pull/5054 removed ProfileMode.

There is still 2 things to port to the new profiler:
- (not needed, the old back-end is deprecated) ~~Port gpu stats to the new profiler (https://github.com/Theano/Theano/pull/5054/files#diff-8f504707a80fc0f2baf16a8847aae2c8). The commanted change in the file theano/sandbox/cuda/basic_ops.py in case the good section don't show up.~~
- [ ] Make a new registry like theano.compile.profilemode.register_profiler_printer in the new profiler in theano/compile/profiling.py. Add a variable "_profiler_printer=[]" and a decorator like https://github.com/Theano/Theano/pull/5054/files#diff-0b102f2db511e60cadfb19a10f8c3356L768.
- [ ] Port [gpu stats](https://github.com/Theano/Theano/pull/5054/files#diff-8f504707a80fc0f2baf16a8847aae2c8) to the new back-end. Mostly port the function profile_printer from the file theano/sandbox/cuda/basic_ops.py to theano/gpuarray/basic_ops.py that use the new register_profile_printer() method created above.
- [ ] Port scan stats (verify first it is now in the current scan stats) (https://github.com/Theano/Theano/pull/5054/files#diff-5bef5ca45d04a952bc292575d565258aR2835) Mostly, update the function profile_printer in theano/scan_module/scan_op.py to use the new profiler/register above)

Sentient07:
What GPU stats are to be ported to the new backend ?

nouiz:
Check the PR, there is 2 todo comment.

Saizheng:
https://github.com/Theano/Theano/pull/5455

PR #5455, opened 2017-01-26
5055:Port ProfileMode GPU to new back-end and profiling.py

Saizheng:
The first step is to port GPU profilemode to profiling.py and new back-end

will fix gh-5055 when finished

### DIFF
--- a/theano/compile/profiling.py
+++ b/theano/compile/profiling.py
@@ -1325,7 +1325,7 @@ class ProfileStats(object):
             print("-----------------", file=file)
             self.optimizer_profile[0].print_profile(file,
                                                     self.optimizer_profile[1])
-        self.print_extra()
+        self.print_extra(file)
         self.print_tips(file)
 
     def print_tips(self, file):
@@ -1472,11 +1472,11 @@ class ProfileStats(object):
         if not printed_tip:
             print("  Sorry, no tip for today.", file=file)
 
-    def print_extra(self):
+    def print_extra(self, file):
         params = [self.message, self.compile_time, self.fct_call_time,
                   self.apply_time, self.apply_cimpl, self.output_size]
         for f in _profiler_printers:
-            f(*params)
+            f(*params, file=file)
 
 
 class ScanProfileStats(ProfileStats):
--- a/theano/compile/profiling.py
+++ b/theano/compile/profiling.py
@@ -102,7 +102,6 @@ def _atexit_print_fn():
                         assert len(merge) == len(cum.optimizer_profile[1])
                         cum.optimizer_profile = (cum.optimizer_profile[0], merge)
                     except Exception as e:
-                        print("Got an exception while merging profile")
                         print(e)
                         cum.optimizer_profile = None
                 else:
@@ -178,7 +177,7 @@ class ProfileStats(object):
         self.apply_time = {}
         self.apply_callcount = {}
         # self.apply_cimpl = None
-        # self.messge = None
+        # self.message = None
     #
     # Note on implementation:
     # Class variables are used here so that each one can be
@@ -1472,7 +1471,8 @@ class ProfileStats(object):
             print("  Sorry, no tip for today.", file=file)
 
     def print_extra(self):
-        params = [None, None, None, None, self.apply_time, None, None, None, None]
+        params = [self.message, self.compile_time, self.fct_call_time,
+                  self.apply_time, self.apply_cimpl, self.output_size]
         for f in _profiler_printers:
             f(*params)
         
--- a/theano/compile/profiling.py
+++ b/theano/compile/profiling.py
@@ -147,10 +147,10 @@ def print_global_stats():
 
 
 _profiler_printers = []
-
 def register_profiler_printer(fct):
     _profiler_printers.append(fct)
     return fct
+
 
 class ProfileStats(object):
 
@@ -1475,7 +1475,8 @@ class ProfileStats(object):
                   self.apply_time, self.apply_cimpl, self.output_size]
         for f in _profiler_printers:
             f(*params)
-        
+
+
 class ScanProfileStats(ProfileStats):
     callcount = 0.0
     nbsteps = 0.0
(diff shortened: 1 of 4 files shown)


theano/compile/profiling.py
```diff
@@ -109,6 +109,12 @@ def _atexit_print_fn():
                     n_apply_to_print=config.profiling.n_apply)
 
 
+_profiler_printers = []
+
+def register_profiler_printer(fct):
+    profiler_printers.append(fct)
```
nouiz: There is a mistake here that make travis fail: `profiler_printers` -> `_profiler_printers`

theano/gpuarray/basic_ops.py
```diff
@@ -1387,6 +1388,58 @@ def make_node(self, x, axis, splits):
...
+                gpu += t
+            else:
+                cpu += t
+        print()
+        print("    Spent %.3fs(%.3f%%) in cpu Op, %.3fs(%.3f%%) in gpu Op and %.3fs(%.3f%%) transfert Op" % (
+            cpu, cpu / local_time * 100, gpu, gpu / local_time * 100,
+            trans, trans / local_time * 100))
+
+        print()
+        print("    Theano function input that are float64")
```
abergeron: I wonder how relevant the float64 listings are given that they are now running on the GPU as well.

nouiz: It is still relevant as most GPU have crappy float64 speed. But maybe we could make this more clear with something like this line:

    Theano function input(s) that are float64. Could cause significant slowdown on some GPUs.

Maybe we should only print them when floatX!=float64?

theano/gpuarray/basic_ops.py
```diff
@@ -1537,6 +1538,58 @@ def c_code(self, node, name, inputs, outputs, sub):
         return main_code % locals()
 
 
+@theano.compile.profiling.register_profiler_printer
+def profile_printer(fct_name, compile_time, fct_call_time, fct_call,
+                    apply_time, apply_cimpl, message, outputs_size,
+                    other_time):
```
nouiz: rename fct_name to message and remove message. Use self.**name** for most of them. Remove other_time and fct_call.
Do this for both profile_printer.

for fct_call, used in this method, you can recreate it like this:

    fct_call = set()
    for node in apply_time.keys():
       fct_call.add(node.fgraph)

nouiz:
travis still have those 2 flake8 errors:

/home/travis/build/Theano/Theano/theano/compile/profiling.py:150:1: E302 expected 2 blank lines, found 0
/home/travis/build/Theano/Theano/theano/scan_module/scan_op.py:2871:58: W291 trailing whitespace

nouiz:
I killed travis for rc3 release. we can restart it after

nouiz:
I'm restarting travis

theano/gpuarray/basic_ops.py
```diff
@@ -1549,46 +1549,45 @@ def profile_printer(message, compile_time, fct_call_time,
...
+        for fg in fgraphs:
+            for i in fg.inputs:
                 if hasattr(i.type, 'dtype') and i.type.dtype == 'float64':
-                    print('        ', fct.name, i.name, i.type, i)
+                    print('        ', fg.name, i.name, i.type, i)
 
         print()
         print("    List of apply that don't have float64 as input but have float64 in outputs")
         print("    (Useful to know if we forgot some cast when using floatX=float32 or gpu code)")
         print('    <Apply> <Apply position> <fct name> <inputs type> <outputs type>')
```
abergeron: This is less useful for the new backend.

theano/gpuarray/basic_ops.py
```diff
@@ -1563,35 +1563,37 @@ def profile_printer(message, compile_time, fct_call_time,
...
+                    print('        ', fg.name, i.name, i.type, i, file=file)
 
-        print()
-        print("    List of apply that don't have float64 as input but have float64 in outputs")
-        print("    (Useful to know if we forgot some cast when using floatX=float32 or gpu code)")
-        print('    <Apply> <Apply position> <fct name> <inputs type> <outputs type>')
+        print('', file=file)
+        print("    List of apply that don't have float64 as input but have float64 in outputs", file=file)
+        print("    (Useful to know if we forgot some cast when using floatX=float32 or gpu code)", file=file)
+        print('    <Apply> <Apply position> <fct name> <inputs type> <outputs type>', file=file)
```
abergeron: The float 64 part must be less useful here.

theano/gpuarray/basic_ops.py
```diff
@@ -1538,6 +1540,62 @@ def c_code(self, node, name, inputs, outputs, sub):
...
+        print("    <fct name> <input name> <input type> <str input>", file=file)
+        for fg in fgraphs:
+            for i in fg.inputs:
+                if hasattr(i.type, 'dtype') and i.type.dtype == 'float64':
+                    print('        ', fg.name, i.name, i.type, i, file=file)
+
+        print('', file=file)
+        print("    List of apply that don't have float64 as input but have float64 in outputs", file=file)
+        print("    (Useful to know if we forgot some cast when using floatX=float32 or gpu code)", file=file)
+        print('    <Apply> <Apply position> <fct name> <inputs type> <outputs type>', file=file)
```
abergeron: The part about ops with float64 is less useful for the new backend.

abergeron:
I'm ok with the changes, even as-is, but I question the usefulness of warning about float64.

theano/gpuarray/basic_ops.py
nouiz: yes and no. A gamer GPU is as fast as the CPU in float64 in my head. If I'm right, the problem is the same, potential slowdown by at least 10x. The new gamer GPU have a bigger speed difference between float32 and float64 then in the past, so that didn't got better.

But strictly speaking, it run on the GPU, thep problem is the speed on gamer GPU and some servers GPU. That is why I think it is still useful.

Issue #5374, opened 2016-12-28
Indexing with empty list


317070:
Unlike numpy, theano cannot index with an empty list. It seems to me that this should be something AdvancedSubtensor should be able to deal with?
```
>>> np.zeros((5,5))[:,[]]
array([], shape=(5, 0), dtype=float64)
```
versus
```
>>> T.fmatrix()[:,[]]
TypeError: index must be integers
```

Note that the following do work fine. It is only empty python lists which pose difficulty.
```
>>> T.fmatrix()[:,np.array([],dtype='int64')]
DimShuffle{1,0}.0
>>> T.fmatrix()[:,[1]]
DimShuffle{1,0}.0
```

abergeron:
This is probably because `numpy.asarray([]) == numpy.array([], dtype='float64')`.

nouiz:
If so, should we make a special case, when the idx list is an empty
constant and cast it to int?

Or detect that case and give a better user error?

I would prefer option 1, as it make us behave more like numpy.

affanv14:
@nouiz I'd like to try implementing option 1.

nouiz:
thanks. We didn't agree on the final implementation, but you can start. The
difference is small between the options. A starting point:

tensor/var.py the method __getitem__ is what is called by [] symbol in
python. This is the entry point.

lamblin:
I think the way to go is, in `__getitem__`, to check if one of the indexing arguments is an empty list before trying to convert it using `as_tensor_variable`, and instead call `as_tensor_variable(np.array([], dtype='int64'))`.
I vaguely recall something similar having been done somewhere already (or at least a ticket being open).

affanv14:
@lamblin 
Okay.If i understood you correctly. I just need to check in __getitem__ if i got an empty list in an arg and replace it with None?

317070:
Not with None, but with
as_tensor_variable(np.array([], dtype='int64'))

affanv14:
I think i did just that in the PR but in an inner call.

PR #5456, opened 2017-01-26
Implement indexing with empty list

affanv14:
closes https://github.com/Theano/Theano/issues/5374

### DIFF
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -501,6 +501,8 @@ class _tensor_py_operators(object):
             args[ellipsis_at: ellipsis_at + 1] = (
                 [slice(None)] * (self.ndim - (len(args) - 1 - new_axes)))
 
+        args = tuple([numpy.array(inp, dtype=numpy.int64)
+                      if(inp == [] or inp == ()) else inp for inp in args])
         # Convert python literals to theano constants
         args = theano.tensor.subtensor.make_constant(args)
         # Determine if advanced indexing is needed or not
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -501,6 +501,8 @@ class _tensor_py_operators(object):
             args[ellipsis_at: ellipsis_at + 1] = (
                 [slice(None)] * (self.ndim - (len(args) - 1 - new_axes)))
 
+        # Force input to be int64 datatype if input is an empty list or tuple
+        # Else leave it as is if it is a real number
         args = tuple([numpy.array(inp, dtype=numpy.int64)
                       if(inp == [] or inp == ()) else inp for inp in args])
         # Convert python literals to theano constants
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -141,6 +141,8 @@ def as_tensor_variable(x, name=None, ndim=None):
         string.
     ndim : None or integer
         Return a Variable with this many dimensions.
+    dtype : numpy dtype or None
```
nouiz: Adding this parameter here will request that we respect it at all places in the method. I think it would be simpler to not add this parameter. You can call numpy.asarray(x, dtype=...) on the input before calling as_tensor_variable in this case.

theano/tensor/subtensor.py
```diff
@@ -2304,7 +2304,10 @@ def R_op(self, inputs, eval_points):
 
 def take(a, indices, axis=None, mode='raise'):
     a = theano.tensor.as_tensor_variable(a)
-    indices = theano.tensor.as_tensor_variable(indices)
+    if(indices == []):
```
nouiz: What happen with an empty tuple? Can you test that? If numpy behave on an empty tuple like an empty list, make sure to do the same.

Also add test.

theano/tensor/basic.py
affanv14: why did'nt that occur to me!
I'm on it.

affanv14:
@nouiz 
How about this?.This seems to work.

affanv14:
Nope.I tried some tests.
This needs more work.

theano/tensor/var.py
```diff
@@ -509,6 +509,8 @@ def check_bool(args_el):
         # AdvancedIndexingError, advanced indexing
         advanced = False
         axis = None
+        args = tuple([numpy.array(inp, dtype=numpy.int64)
+                      if(inp == [] or inp == ()) else inp for inp in args])
```
nouiz: Maybe you should move that before line 505?

affanv14:
@nouiz how about this?
It does seem to work properly

theano/tensor/var.py
```diff
@@ -501,6 +501,8 @@ def check_bool(args_el):
             args[ellipsis_at: ellipsis_at + 1] = (
                 [slice(None)] * (self.ndim - (len(args) - 1 - new_axes)))
 
+        args = tuple([numpy.array(inp, dtype=numpy.int64)
```
nouiz: Just add a comment to help future reader of this code understand. Something like:

Force int64 for empty list and tuple as otherwise, then end up as real numbers.

affanv14: okay.

nouiz:
travis have this failure:

```
/home/travis/build/Theano/Theano/theano/tensor/var.py:504:9: E265 block comment should start with '# '
/home/travis/build/Theano/Theano/theano/tensor/var.py:505:9: E265 block comment should start with '# '
```

affanv14:
@nouiz It's done

nouiz:
thanks

Issue #5408, opened 2017-01-13
MRG choice implementation (replace multinomial_wo_replacement)


nouiz:
This was one option from https://github.com/Theano/Theano/issues/5292

There was an interface mistake that we need to fix. NumPy random and our wrapper of it have 2 functions: multinomial and choices. choises return the index as our multinomial_wo_replacements. So we should rename multinomial_wo_replacements to choices while respecting the choices interface. Here is what need to be done:

- [x] In sandbox/rng_mrg.py make a choice() method like the choice method in tensor/raw_random.py and have it reuse the MultinomialWOReplacementFromUniform op. The replace=True parameter (the defaul will be implemented later, so for now, by default it raise an error and when replace=False, it work).
- [x] Make multinomial_wo_replacements() in rng_mrg.py warn that it is deprecated and we will remove this.
- [x] Commit this interface change and get it merged ideally before 0.9 release.

In a second time:
- [x] Pickle a graph with the old op for later test.
- [x] Rename the op MultinomialWOReplacementFromUniform (CPU and the gpuarray version, not the old back-end version) to ChoiceFromUniform.
- [x] Make a test that reload the old pickle and make that case works (maybe just making MultinomialWOReplacementFromUniform = ChoiceFromUniform will work)

In a thrid time:
- [ ] make the replace a props and add a \__setstate__ method to still have the old pickle file unpickle correctly.
- [ ] Implement replace=True, by skipping the renormalization in the code. (the second option of gh-5292)

PR #5457, opened 2017-01-26
replace multinomial_wo_replacement by choice

adbrebs:
First part of #5408, @nouiz.

I copied the same signature as the choice method in tensor/raw_random.py , which is not the same as the choice signature of the numpy function (order of a, size reversed).

Also, should I replace multinomial_wo_replacement in the tests now?

### DIFF
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1446,55 +1446,85 @@ class MRG_RandomStreams(object):
             raise NotImplementedError(("MRG_RandomStreams.multinomial only"
                                        " implemented for pvals.ndim = 2"))
 
-    def multinomial_wo_replacement(self, size=None, n=1, pvals=None,
-                                   ndim=None, dtype='int64', nstreams=None):
-        # TODO : need description for parameter
+    def choice(self, size=None, a=2, replace=True, p=None, ndim=None,
+               dtype='int64', nstreams=None):
         """
-        Sample `n` times *WITHOUT replacement* from a multinomial distribution
-        defined by probabilities pvals, and returns the indices of the sampled
-        elements.
-        `n` needs to be in [1, m], where m is the number of elements to select
-        from, i.e. m == pvals.shape[1]. By default n = 1.
+        Sample `size` times from a multinomial distribution defined by
+        probabilities `p` and values `a`.
 
-        Example : pvals = [[.98, .01, .01], [.01, .49, .50]] and n=1 will
-        probably result in [[0],[2]]. When setting n=2, this
+        Parameters
+        ----------
+        size: integer or None (default None)
+            the number of samples to be generated. If None, a single sample is
+            generated.
+        a: integer or 1d or 2d numpy array or theano tensor
+            the values of the samples. If a is an integer, the values are
+            generated between 0 and a-1.
+        p: 1d or 2d numpy array or theano tensor
+            the probabilities of the distribution associated to each value of
+            `a`. It should have the same shape as a (if a is an array/tensor).
+        replace: bool (default True)
+            Whether the sample is with or without replacement.
+            Only replace=False is implemented for now.
+
+        Example : p = [[.98, .01, .01], [.01, .49, .50]] and size=1 will
+        probably result in [[0],[2]]. When setting size=2, this
         will probably result in [[0,1],[2,1]].
 
         Notes
         -----
-        -`size` and `ndim` are only there keep the same signature as other
+        - `size` and `ndim` are only there keep the same signature as other
         uniform, binomial, normal, etc.
-        TODO : adapt multinomial to take that into account
 
-        -Does not do any value checking on pvals, i.e. there is no
+        - Does not do any value checking on pvals, i.e. there is no
         check that the elements are non-negative, less than 1, or
         sum to 1. passing pvals = [[-2., 2.]] will result in
         sampling [[0, 0]]
 
-        """
-        if pvals is None:
-            raise TypeError("You have to specify pvals")
-        pvals = as_tensor_variable(pvals)
+        - When `a` and `p` are tensors their shape should be the same.
 
-        if size is not None:
-            raise ValueError("Provided a size argument to "
-                             "MRG_RandomStreams.multinomial_wo_replacement, "
-                             "which does not use the size argument.")
+        - Only replace=False is implemented for now.
+
+        """
+        if replace:
+            raise NotImplementedError(
+                "MRG_RandomStreams.choice only works without replacement "
+                "for now.")
+
+        if p is None:
+            raise TypeError("You have to specify p.")
+        p = as_tensor_variable(p)
+
         if ndim is not None:
-            raise ValueError("Provided an ndim argument to "
-                             "MRG_RandomStreams.multinomial_wo_replacement, "
-                             "which does not use the ndim argument.")
-        if pvals.ndim == 2:
-            # size = [pvals.shape[0], as_tensor_variable(n)]
-            size = pvals[:, 0].shape * n
-            unis = self.uniform(size=size, ndim=1, nstreams=nstreams)
-            op = multinomial.MultinomialWOReplacementFromUniform(dtype)
-            n_samples = as_tensor_variable(n)
-            return op(pvals, unis, n_samples)
-        else:
+            raise ValueError("ndim argument to "
+                             "MRG_RandomStreams.multinomial_wo_replacement "
+                             "is not used.")
+
+        if p.ndim == 1:
+            p = tensor.shape_padleft(p)
+
+        if p.ndim != 2:
             raise NotImplementedError(
                 "MRG_RandomStreams.multinomial_wo_replacement only implemented"
-                " for pvals.ndim = 2")
+                " for p.ndim = 1 or p.ndim = 2")
+
+        shape = p[:, 0].shape * size
+        unis = self.uniform(size=shape, ndim=1, nstreams=nstreams)
+        op = multinomial.MultinomialWOReplacementFromUniform(dtype)
+        sampled_indices = op(p, unis, as_tensor_variable(size))
+
+        if isinstance(a, int):
+            return sampled_indices
+        a = tensor.as_tensor_variable(a)
+        return a[sampled_indices]
+
+    def multinomial_wo_replacement(self, size=None, n=1, pvals=None,
+                                   ndim=None, dtype='int64', nstreams=None):
+        warnings.warn('MRG_RandomStreams.multinomial_wo_replacement() is '
+                      'deprecated and will be removed in the next release of '
+                      'Theano. Please use MRG_RandomStreams.choice() instead.')
+        return self.choice(size=n, replace=False, p=pvals, dtype=dtype,
+                           nstreams=nstreams, ndim=ndim)
 
     def normal(self, size, avg=0.0, std=1.0, ndim=None,
                dtype=None, nstreams=None):
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1446,7 +1446,7 @@ class MRG_RandomStreams(object):
             raise NotImplementedError(("MRG_RandomStreams.multinomial only"
                                        " implemented for pvals.ndim = 2"))
 
-    def choice(self, size=1, a=None, replace=True, p=None, ndim=None,
+    def choice(self, size=1, a=2, replace=True, p=None, ndim=None,
                dtype='int64', nstreams=None):
         """
         Sample `size` times from a multinomial distribution defined by
@@ -1456,10 +1456,13 @@ class MRG_RandomStreams(object):
         Parameters
         ----------
         size: integer or integer tensor (default 1)
-            The number of samples. It should be between 1 and `p.shape[1]-1`
-        a: None
+            The number of samples. It should be between 1 and `p.shape[1]-1`.
+        a: None (default 2)
             For now, a should be None. This function will sample
-            values between 0 and `p.shape[1]-1`.
+            values between 0 and `p.shape[1]-1`. When a != None will be
+            implemented, if `a` is a scalar, the samples are drawn from the
+            range 0,...,a-1. We default to 2 as to have the same interface as
+            RandomStream.
         replace: bool (default True)
             Whether the sample is with or without replacement.
             Only replace=False is implemented for now.
@@ -1473,7 +1476,7 @@ class MRG_RandomStreams(object):
 
         Notes
         -----
-        -`size` and `ndim` are only there keep the same signature as other
+        -`ndim` are only there keep the same signature as other
         uniform, binomial, normal, etc.
 
         -Does not do any value checking on pvals, i.e. there is no
@@ -1518,6 +1521,7 @@ class MRG_RandomStreams(object):
         warnings.warn('MRG_RandomStreams.multinomial_wo_replacement() is '
                       'deprecated and will be removed in the next release of '
                       'Theano. Please use MRG_RandomStreams.choice() instead.')
+        assert size is None
         return self.choice(size=n, a=None, replace=False, p=pvals,
                            dtype=dtype, nstreams=nstreams, ndim=ndim)
 
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1446,7 +1446,7 @@ class MRG_RandomStreams(object):
             raise NotImplementedError(("MRG_RandomStreams.multinomial only"
                                        " implemented for pvals.ndim = 2"))
 
-    def choice(self, size=1, a=2, replace=True, p=None, ndim=None,
+    def choice(self, size=1, a=None, replace=True, p=None, ndim=None,
                dtype='int64', nstreams=None):
         """
         Sample `size` times from a multinomial distribution defined by
@@ -1458,7 +1458,7 @@ class MRG_RandomStreams(object):
         ----------
         size: integer or integer tensor (default 1)
             The number of samples. It should be between 1 and `p.shape[1]-1`.
-        a: int or None (default 2)
+        a: int or None (default None)
             For now, a should be None. This function will sample
             values between 0 and `p.shape[1]-1`. When a != None will be
             implemented, if `a` is a scalar, the samples are drawn from the
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1450,14 +1450,15 @@ class MRG_RandomStreams(object):
                dtype='int64', nstreams=None):
         """
         Sample `size` times from a multinomial distribution defined by
-        probabilities `p`. Sampled values are between 0 and `p.shape[1]-1`.
+        probabilities `p`, and returns the indices of the sampled elements.
+        Sampled values are between 0 and `p.shape[1]-1`.
         Only sampling without replacement is implemented for now.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


adbrebs:
The flake8 tests fail because of
```
/home/travis/build/Theano/Theano/theano/sandbox/rng_mrg.py:docstring of theano.sandbox.rng_mrg.MRG_RandomStreams.choice:25: WARNING: Bullet list ends without a blank line; unexpected unindent.
```
Locally flake8 runs without error, so I am not sure what the problem is.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+            generated.
+        a: integer or 1d or 2d numpy array or theano tensor
+            the values of the samples. If a is an integer, the values are
+            generated between 0 and a-1.
+        p: 1d or 2d numpy array or theano tensor
+            the probabilities of the distribution associated to each value of
+            `a`. It should have the same shape as a (if a is an array/tensor).
+        replace: bool (default True)
+            Whether the sample is with or without replacement.
+            Only replace=False is implemented for now.
```
nouiz: swap p and replace.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
-        if pvals.ndim == 2:
-            # size = [pvals.shape[0], as_tensor_variable(n)]
-            size = pvals[:, 0].shape * n
-            unis = self.uniform(size=size, ndim=1, nstreams=nstreams)
-            op = multinomial.MultinomialWOReplacementFromUniform(dtype)
-            n_samples = as_tensor_variable(n)
-            return op(pvals, unis, n_samples)
-        else:
+            raise ValueError("ndim argument to "
+                             "MRG_RandomStreams.multinomial_wo_replacement "
```
nouiz: multinomial_wo_replacement -> choice

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+                             "is not used.")
+
+        if p.ndim == 1:
+            p = tensor.shape_padleft(p)
```
nouiz: Good idea to add this support, but I think this return the wrong number of dimensions in the end. To supporte it correctly, we probably need to remove the extra dimensions before returning.

Also, I think we are pretty slow in that case. So I would prefer that we don't add the support for it in this PR that is more about the user interface change. And this would request new tests.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+
+        if p.ndim != 2:
             raise NotImplementedError(
                 "MRG_RandomStreams.multinomial_wo_replacement only implemented"
-                " for pvals.ndim = 2")
+                " for p.ndim = 1 or p.ndim = 2")
+
+        shape = p[:, 0].shape * size
```
nouiz: This isn't the right way to add the support for size. If size and pvals are provided, then we need to broadcast pval to shape. I would prefer that you don't try to add the support for that feature in this PR.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        Sample `size` times from a multinomial distribution defined by
+        probabilities `p` and values `a`.
 
-        Example : pvals = [[.98, .01, .01], [.01, .49, .50]] and n=1 will
-        probably result in [[0],[2]]. When setting n=2, this
+        Parameters
+        ----------
+        size: integer or None (default None)
+            the number of samples to be generated. If None, a single sample is
+            generated.
```
adbrebs: Thanks for the review.
I don't understand this point, pval specifies the probabilities, not the size. We need an argument to the function that specifies how many samples from pval that we want, don't we?

In the original multinomial_wo_replacement function, the argument was `n`.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        if isinstance(a, int):
+            return sampled_indices
+        a = tensor.as_tensor_variable(a)
+        return a[sampled_indices]
+
+    def multinomial_wo_replacement(self, size=None, n=1, pvals=None,
+                                   ndim=None, dtype='int64', nstreams=None):
+        warnings.warn('MRG_RandomStreams.multinomial_wo_replacement() is '
+                      'deprecated and will be removed in the next release of '
+                      'Theano. Please use MRG_RandomStreams.choice() instead.')
```
nouiz: Maybe add the parameter as in the fct call? (size=n, a=!!!!, replace=False, p=pvals, dtype=dtype, ndim=ndim, nstreams=nstreams). See the other comment, I think you currently don't have the right formula for the conversion.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,85 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        return self.choice(size=n, replace=False, p=pvals, dtype=dtype,
+                           nstreams=nstreams, ndim=ndim)
```
nouiz: I'm not sure this conversion is well done. A way to test this is to make a test, that will create 2 MRG_RandomStreams() and draw from both of them one with choice and the other with multinomial_wo_replacements. And compare if we have the same output.

The code bellow is what I think is the equivalence, but compared with the numpy rng theano wrapper.

```
import theano
import theano.sandbox.rng_mrg
i=theano.tensor.iscalar()
mat=theano.tensor.matrix()
m=theano.sandbox.rng_mrg.MRG_RandomStreams()
mm=m.multinomial_wo_replacement(pvals=mat, n=i)
f=theano.function([mat,i], mm)

cm=theano.tensor.shared_randomstreams.RandomStreams()
cmm=cm.choice(a=mat.shape[1], size=[i], p=mat[0], replace=False)   # Here choice only support a vector for p, this limit the testing, but still allow enough testing I think.
cf=theano.function([mat,i], cmm)
p=[[.1]*10]
pp=[[0.01]*9+[.91,]]
for i in range(10): f(p, 2); cf(p,2)

print()
for i in range(10): f(pp, 4); cf(pp,4)
```

nouiz:
On Fri, Jan 27, 2017 at 10:27 AM, Alexandre de Brébisson <
notifications@github.com> wrote:

> *@adbrebs* commented on this pull request.
> ------------------------------
>
> In theano/sandbox/rng_mrg.py <https://github.com/Theano/Theano/pull/5457>:
>
> >
> -        Example : pvals = [[.98, .01, .01], [.01, .49, .50]] and n=1 will
> -        probably result in [[0],[2]]. When setting n=2, this
> +        Parameters
> +        ----------
> +        size: integer or None (default None)
> +            the number of samples to be generated. If None, a single sample is
> +            generated.
>
> Thanks for the review.
> I don't understand this point, pval specifies the probabilities, not the
> size. We need an argument to the function that specifies how many samples
> from pval that we want, don't we?In the original multinomial_wo_replacement
> function, the argument was n.
>

Yes, it was n. But I think the problem is that your currently, your mapping
isn't right (Or it is mine that isn't right, so causing the confusion
between us)

theano/sandbox/rng_mrg.py
adbrebs: With this version of the PR, f returns elements like [[7 2]], while cf returns elements like [7 2].

nouiz:
Yes, but this is due to a limitation that the RandomStream.choice limit to
a vector of prob and MRG_RS limit to a matrix of prob.

Other then that, as I test with a "matrix" of 1 row vs 1 vector of prob, it
should be equivalent to my knowledge.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,74 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+
+        if p is None:
+            raise TypeError("You have to specify p in "
+                            "MRG_RandomStreams.choice.")
+        p = as_tensor_variable(p)
 
-        if size is not None:
-            raise ValueError("Provided a size argument to "
-                             "MRG_RandomStreams.multinomial_wo_replacement, "
-                             "which does not use the size argument.")
```
nouiz: Add back this check.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,74 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        if p.ndim != 2:
             raise NotImplementedError(
-                "MRG_RandomStreams.multinomial_wo_replacement only implemented"
-                " for pvals.ndim = 2")
+                "MRG_RandomStreams.choice is only implemented for p.ndim = 2")
+
+        shape = p[:, 0].shape * size
+        unis = self.uniform(size=shape, ndim=1, nstreams=nstreams)
+        op = multinomial.MultinomialWOReplacementFromUniform(dtype)
+        return op(p, unis, as_tensor_variable(size))
```
nouiz: size is always None, so this should fail from my understanding.

I think on line 1505 and 1508 you want to replace size by a.

Can you make the test now before doing more change? I think it will help you do the implementation.

theano/sandbox/rng_mrg.py
adbrebs: `size` is the number of samples that we want, it corresponds to the `n` of `multinomial_wo_replacement`. So I think that it should not be None (it should be an integer or integer tensor).

theano/sandbox/rng_mrg.py
adbrebs: Ultimately `a` should either be an integer or an array of the same shape as `p`. But for now I have put its default value to `None` as you said that I should not add extra features.
The sample values are 0 to `p.shape[1]-1`.

adbrebs:
Tests have passed locally. @nouiz let me know if the new signature of the choice function is fine with you.

nouiz:
There is travis erro:

/home/travis/build/Theano/Theano/theano/sandbox/rng_mrg.py:docstring of theano.sandbox.rng_mrg.MRG_RandomStreams.choice:24: WARNING: Bullet list ends without a blank line; unexpected unindent.

Mostly, there is a problem with the docstring of choice.

To regenerate the error, you can run: `python doc/scripts/docgen.py --nopdf --check`

theano/sandbox/rng_mrg.py
```diff
@@ -1446,85 +1446,80 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
-            generated between 0 and a-1.
-        p: 1d or 2d numpy array or theano tensor
-            the probabilities of the distribution associated to each value of
-            `a`. It should have the same shape as a (if a is an array/tensor).
+        size: integer or integer tensor (default 1)
+            The number of samples. It should be between 1 and `p.shape[1]-1`
+        a: None
+            For now, a should be None. This function will sample
+            values between 0 and `p.shape[1]-1`.
         replace: bool (default True)
```
abergeron: If only replace=False is implemented, why is the default True?

lamblin:
I'll have a look tomorrow.

nouiz:
Because it is this the detail on numpy and RandomStream. We should not have
different default otherwise when it is implemented, it will be confusing
and cause users error.

theano/sandbox/rng_mrg.py
lamblin: Because the default is True in  `numpy.random.choice`, I guess.

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,84 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
-        # TODO : need description for parameter
+    def choice(self, size=1, a=2, replace=True, p=None, ndim=None,
+               dtype='int64', nstreams=None):
         """
-        Sample `n` times *WITHOUT replacement* from a multinomial distribution
-        defined by probabilities pvals, and returns the indices of the sampled
-        elements.
-        `n` needs to be in [1, m], where m is the number of elements to select
-        from, i.e. m == pvals.shape[1]. By default n = 1.
+        Sample `size` times from a multinomial distribution defined by
```
lamblin: We should keep the part about the fact that it returns sampled indices, and not a vector of counts (like multinomial() does).

theano/sandbox/rng_mrg.py
```diff
@@ -1518,6 +1521,7 @@ def multinomial_wo_replacement(self, size=None, n=1, pvals=None,
         warnings.warn('MRG_RandomStreams.multinomial_wo_replacement() is '
                       'deprecated and will be removed in the next release of '
                       'Theano. Please use MRG_RandomStreams.choice() instead.')
+        assert size is None
```
abergeron: If the default for size is 1 then we don't want to assert that the value is None.  You should change the default value or this assert.

theano/sandbox/rng_mrg.py
```diff
@@ -1456,10 +1456,13 @@ def choice(self, size=1, a=None, replace=True, p=None, ndim=None,
         Parameters
         ----------
         size: integer or integer tensor (default 1)
-            The number of samples. It should be between 1 and `p.shape[1]-1`
-        a: None
+            The number of samples. It should be between 1 and `p.shape[1]-1`.
+        a: None (default 2)
```
abergeron: The type specification doesn't make sense.  What comes after the parameter name is the type.  I think you want `int` instead of `None`.

theano/sandbox/rng_mrg.py
```diff
@@ -1473,7 +1476,7 @@ def choice(self, size=1, a=None, replace=True, p=None, ndim=None,
 
         Notes
         -----
-        -`size` and `ndim` are only there keep the same signature as other
+        -`ndim` are only there keep the same signature as other
```
abergeron: are only -> is only

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,84 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+            0 to `p.shape[1]-1`.
+
+        Example : p = [[.98, .01, .01], [.01, .49, .50]] and size=1 will
+        probably result in [[0],[2]]. When setting size=2, this
         will probably result in [[0,1],[2,1]].
 
         Notes
         -----
-        -`size` and `ndim` are only there keep the same signature as other
+        -`ndim` are only there keep the same signature as other
```
lamblin: `ndim` *is*

theano/sandbox/rng_mrg.py
```diff
@@ -1518,6 +1521,7 @@ def multinomial_wo_replacement(self, size=None, n=1, pvals=None,
         warnings.warn('MRG_RandomStreams.multinomial_wo_replacement() is '
                       'deprecated and will be removed in the next release of '
                       'Theano. Please use MRG_RandomStreams.choice() instead.')
+        assert size is None
         return self.choice(size=n, a=None, replace=False, p=pvals,
                            dtype=dtype, nstreams=nstreams, ndim=ndim)
```
abergeron: As far as I understand, the value of a is completely ignored here.  We should at least check that it's two.

lamblin:
Jenkins test this please

theano/sandbox/rng_mrg.py
```diff
@@ -1446,55 +1446,84 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
+        probabilities `p`. Sampled values are between 0 and `p.shape[1]-1`.
+        Only sampling without replacement is implemented for now.
 
-        Example : pvals = [[.98, .01, .01], [.01, .49, .50]] and n=1 will
-        probably result in [[0],[2]]. When setting n=2, this
+        Parameters
+        ----------
+        size: integer or integer tensor (default 1)
+            The number of samples. It should be between 1 and `p.shape[1]-1`.
+        a: None (default 2)
```
lamblin: Wait, the default should actually be `None`, not 2.

theano/sandbox/rng_mrg.py
nouiz: Size don't default to 1, it default to using the shape from pvals.

theano/sandbox/rng_mrg.py
nouiz: I will update to "None or int"

theano/sandbox/rng_mrg.py
nouiz: ok

theano/sandbox/rng_mrg.py
nouiz: a isn't an input of multinomial_wo_replacement. I don't see an input of multinomial_wo_replacement that is ignored except size that we have the assert on. So I don't see a problem here.

theano/sandbox/rng_mrg.py
nouiz: ok

theano/sandbox/rng_mrg.py
nouiz: No, RandomStream have a default of 2: https://github.com/Theano/Theano/blob/master/theano/tensor/raw_random.py#L977

nouiz:
I updated the docstring as per some comments and replied to all comments.

theano/sandbox/rng_mrg.py
abergeron: Currently in the function definition it does default to 1.

theano/sandbox/rng_mrg.py
abergeron: a is one of the parameters of choice() above and we don't do anything with it, hence it's ignored.

nouiz: We always pass None here to a. Inside the method choice, we assert that it is None. It is there for a futur implementation and respect the current interface of RandomStream.

theano/sandbox/rng_mrg.py
nouiz: The definition of size for multinomial_wo_replacement and choice aren't the same. size must be None for multinomial_wo_replacement. But the size of choice correspond to the n parameter of multinomial_wo_replacement.

theano/sandbox/rng_mrg.py
nouiz: done

theano/sandbox/rng_mrg.py
abergeron: This is still None

theano/sandbox/rng_mrg.py
abergeron: Still are here

theano/sandbox/rng_mrg.py
abergeron: It doesn't really matter, it is bad design to make a function that will give an AssertionError if you don't figure out what magic value to set some parameters to, which will be the case here.

theano/sandbox/rng_mrg.py
abergeron: Wait, I just realized that this function is dumb and call itself.  So in any case, we ignore everything and bust the stack.

lamblin:
Jenkins test this please

PR #5461, opened 2017-01-27
validate numpy1.12 compatibility changes

chinnadhurai:
Contains Travis.yml changes to validate the numpy1.12 compatibility changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,11 +54,11 @@ env:
 matrix:
   fast_finish: true
   include:
-    - python: "3.3"
+    - python: "3.4"
       env: PART="--exclude-dir=theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE"
-    - python: "3.3"
+    - python: "3.4"
       env: PART="theano/tensor --exclude-dir=theano/tensor/nnet" THEANO_FLAGS="mode=FAST_COMPILE"
-    - python: "3.3"
+    - python: "3.4"
       env: PART="theano/tensor/nnet" THEANO_FLAGS="mode=FAST_COMPILE"
     - python: "2.7"
       env: PART="--exclude-dir=theano/tensor -e test_flake8.py ." THEANO_FLAGS="mode=FAST_COMPILE,floatX=float32"
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,7 +81,7 @@ script:
   - echo "$PART"
   - cd -; cd Theano
   - python -c 'import theano; print(theano.config.__str__(print_doc=False))'
-  - python -c 'import theano; assert(theano.config.blas.ldflags != "")'
+  # python -c 'import theano; assert(theano.config.blas.ldflags != "")'
   - theano-nose -v $PART
   - if [[ $DOC == "1" ]]; then python doc/scripts/docgen.py --nopdf --check; fi
   - if [[ $DOC == "1" ]]; then python doc/scripts/docgen.py --test --check; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ install:
   - pip install flake8-future-import nose-parameterized==0.5.0 sphinx_rtd_theme
   # nose-exclude plugin allow us to tell nosetests to exclude folder with --exclude-dir=path/to/directory.
   - pip install nose-exclude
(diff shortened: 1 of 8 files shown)


slefrancois:
tests #5396

slefrancois:
There seems to be some legitimate errors. To make them easier to find, remove the ldblas test from travis.yml so the build doesn't fail for this :

.travis.yml
```diff
@@ -12,7 +12,7 @@ cache:
 language: python
 python:
   - "2.7"
-  - "3.3"
+  - "3.4"
```
slefrancois: You also need to change Theano/.travis/travis_install.sh to create a conda environment for python 3.4.

slefrancois:
@nouiz @lamblin ,Travis has some legitimate errors:

- https://travis-ci.org/Theano/Theano/jobs/196714103#L3203
- https://travis-ci.org/Theano/Theano/jobs/196714103#L3250
- https://travis-ci.org/Theano/Theano/jobs/196714078#L1542

The rest are packages like scipy which we haven't properly setup under python 3.4 in Travis, I think we can ignore.

nouiz:
So you seem to have been able to reproduce more of the reported errors.

So the follow up is to fix them.

thanks

On Tue, Jan 31, 2017 at 10:45 AM, Simon Lefrancois <notifications@github.com

slefrancois:
c code tests fail in Travis, we might have to flush the cache or bump the version.

jenkins test this

theano/scalar/basic.py
```diff
@@ -1547,6 +1547,9 @@ def c_code(self, node, name, inputs, outputs, sub):
         (x, y) = inputs
         (z,) = outputs
         return "%(z)s = (%(x)s & %(y)s);" % locals()
+
+    def c_code_cache_version(self):
+        return (2,)
```
slefrancois: bump the number to force cache refresh on Travis

lamblin:
The version of the scipy package does not seem correct, it indicates `np19py34_0` for Python 3 and `np19py27_0` for Python 2. Numpy and scipy versions should be compatible.
This could explain the issues when importing scipy modules in the python 3 builds.

slefrancois:
Going forward, the priority is to solve my comments above (multiple 0 shapes and c code version), plus bumping the scipy version. @chinnadhurai , please make a PR for that.

As for the jenkins failure, it seems to be because the Theano function reduction uses the numpy reduce (see [here](https://github.com/Theano/Theano/blob/master/theano/tensor/elemwise.py#L1414)). A solution might be to write our own ufunc based on a numpy copy, or just use the implemement (see [here](https://github.com/chinnadhurai/Theano/blob/bd6d8b1a184d830b2008af3ff502e0dbd79d45e3/theano/scalar/basic.py#L1541) and [here](https://github.com/chinnadhurai/Theano/blob/bd6d8b1a184d830b2008af3ff502e0dbd79d45e3/theano/tensor/tests/test_elemwise.py#L447)).

@lamblin might handle it if there is no CCW time left this week.

lamblin:
The import errors from `scipy` seem to have been solved.
I just pushed two commits, one that should solve the issue of `test_perform` on older numpy, and one for flake8.
The remaining issue seem to be `TypeError: 'numpy.float64' object cannot be interpreted as an index` in a `reshape` operation.

lamblin:
I canceled the build so I can investigate what is happening with the cache version.
The test passed on my machine with an empty cache.

chinnadhurai:
"The remaining issue seem to be TypeError: 'numpy.float64' object cannot be interpreted as an index in a reshape operation." --> I believe this is a known numpy.prod() issue. I believe we already found a solution. Do you have the error logs for this ?

lamblin:
It was in https://travis-ci.org/Theano/Theano/jobs/200141478#L3220-L3245

chinnadhurai:
thanks got it.. it was a known numpy.prod issue , @slefrancois and I discussed.I shall push the changes shortly.

lamblin:
I just pushed a new change, hopefully fixing the C cache issue.
If you need to add new commits, please pull first, so that my commits do not get lost.
Thanks!

chinnadhurai:
'numpy.float64' object cannot be interpreted --> issue seems to have been fixed. But, the cache issue still seems to persist.

lamblin:
Finished in #5527. Thanks!

Issue #5449, opened 2017-01-25
a_gpuarray_shared_var.set_value() should work inplace


nouiz:
The old back-end was doing it, so this is a regression as it increase the memory usage.

See to know the condition when this should happen. It is a subset of all cases that need to be supported: https://github.com/Theano/Theano/pull/5447

abergeron:
Done in https://github.com/Theano/Theano/pull/5462

nouiz:
done.

PR #5462, opened 2017-01-27
Add inplace filtering for GpuArrayType so that shared value replacement

abergeron:
can be done without additional memory usage.
fix gh-https://github.com/Theano/Theano/issues/5449

### DIFF
--- a/theano/gpuarray/type.py
+++ b/theano/gpuarray/type.py
@@ -197,6 +197,11 @@ class GpuArrayType(Type):
                                              self.broadcastable)
 
     def filter(self, data, strict=False, allow_downcast=None):
+        return self.filter_inplace(data, None, strict=strict,
+                                   allow_downcast=allow_downcast)
+
+    def filter_inplace(self, data, old_data, strict=False,
+                       allow_downcast=None):
         if (isinstance(data, gpuarray.GpuArray) and
                 data.typecode == self.typecode):
             # This is just to make this condition not enter the
@@ -218,9 +223,13 @@ class GpuArrayType(Type):
               (allow_downcast is None and
                type(data) == float and
                self.dtype == config.floatX)):
-            data = gpuarray.array(data, dtype=self.typecode, copy=False,
-                                  ndmin=len(self.broadcastable),
-                                  context=self.context)
+            if not isinstance(data, gpuarray.GpuArray):
+                data = numpy.array(data, dtype=self.dtype, copy=False,
+                                   ndmin=len(self.broadcastable))
+            else:
+                data = gpuarray.array(data, dtype=self.typecode, copy=False,
+                                      ndmin=len(self.broadcastable),
+                                      context=self.context)
         else:
             if not hasattr(data, 'dtype'):
                 converted_data = theano._asarray(data, self.dtype)
@@ -230,12 +239,10 @@ class GpuArrayType(Type):
                                         converted_data,
                                         force_same_dtype=False):
                     data = converted_data
-                    data = gpuarray.array(data, context=self.context)
 
             up_dtype = scalar.upcast(self.dtype, data.dtype)
             if up_dtype == self.dtype:
-                data = gpuarray.array(data, dtype=self.dtype, copy=False,
-                                      context=self.context)
+                data = numpy.array(data, dtype=self.dtype, copy=False)
             else:
                 raise TypeError("%s cannot store a value of dtype %s "
                                 "without risking loss of precision." %
@@ -250,6 +257,12 @@ class GpuArrayType(Type):
             if b and shp[i] != 1:
                 raise TypeError("Non-unit value on shape on a broadcastable"
                                 " dimension.", shp, self.broadcastable)
+        if not isinstance(data, gpuarray.GpuArray):
+            if old_data and old_data.shape == data.shape:
+                old_data.write(data)
+                data = old_data
+            else:
+                data = pygpu.array(data, context=self.context)
         return data
 
     def filter_variable(self, other, allow_convert=True):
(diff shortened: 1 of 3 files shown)


nouiz:
add a test that this work inplace.

theano/gpuarray/type.py
```diff
@@ -218,9 +223,13 @@ def filter(self, data, strict=False, allow_downcast=None):
...
-            data = gpuarray.array(data, dtype=self.typecode, copy=False,
-                                  ndmin=len(self.broadcastable),
-                                  context=self.context)
+            if not isinstance(data, gpuarray.GpuArray):
+                data = numpy.array(data, dtype=self.dtype, copy=False,
+                                   ndmin=len(self.broadcastable))
+            else:
+                data = gpuarray.array(data, dtype=self.typecode, copy=False,
+                                      ndmin=len(self.broadcastable),
+                                      context=self.context)
```
nouiz: add a test for that.

theano/gpuarray/type.py
```diff
@@ -230,12 +239,10 @@ def filter(self, data, strict=False, allow_downcast=None):
                                         converted_data,
                                         force_same_dtype=False):
                     data = converted_data
-                    data = gpuarray.array(data, context=self.context)
 
             up_dtype = scalar.upcast(self.dtype, data.dtype)
             if up_dtype == self.dtype:
-                data = gpuarray.array(data, dtype=self.dtype, copy=False,
-                                      context=self.context)
+                data = numpy.array(data, dtype=self.dtype, copy=False)
```
nouiz: If data is on the GPU, then we will transfer it back to the CPU, to move it to the GPU again.

theano/gpuarray/type.py
```diff
@@ -250,6 +257,12 @@ def filter(self, data, strict=False, allow_downcast=None):
             if b and shp[i] != 1:
                 raise TypeError("Non-unit value on shape on a broadcastable"
                                 " dimension.", shp, self.broadcastable)
+        if not isinstance(data, gpuarray.GpuArray):
+            if old_data and old_data.shape == data.shape:
+                old_data.write(data)
```
nouiz: What are the restriction about the contiguous of data and old_data? See the issue about this and the link it contain. I think we are now less restrictive then in the past. If so, we should update the doc you can find from the issue.

abergeron: Both need to be contiguous for it to work.  However if data is not contiguous it will be copied (on the CPU) to a contiguous buffer.  If old_data is not contiguous it will error out.

theano/gpuarray/type.py
abergeron: How?  Do you have an example of a test?

theano/gpuarray/type.py
abergeron: fixed

theano/gpuarray/type.py
nouiz: Just something like:

shared = theano.shared(data, target='cuda')
old_container = shared.get_value(borrow=True, return_internal_type=True)
new_data = np...
shared.set_vaue(new_data)
assert old_container is shared.get_value(borrow=True, return_internal_type=True)

theano/gpuarray/type.py
```diff
@@ -250,6 +260,12 @@ def filter(self, data, strict=False, allow_downcast=None):
             if b and shp[i] != 1:
                 raise TypeError("Non-unit value on shape on a broadcastable"
                                 " dimension.", shp, self.broadcastable)
+        if not isinstance(data, gpuarray.GpuArray):
+            if old_data is not None and old_data.shape == data.shape:
+                old_data.write(data)
+                data = old_data
+            else:
+                data = pygpu.array(data, context=self.context)
         return data
```
nouiz: The old back-end currently have the same problem. If we call set_value(a_gpu_array, borrow=False), then in GpuArraySharedVariable.set_value(), we will do a copy before calling filter_inplace. And filter_inplace won't copy it to the old memory region. So we end up doing an extra allocation for no good reason.

There isn't an easy fix. It will probably need to add a method set_value to the class Container in gof/link.py. I'll make an issue about that. This isn't for this PR.

abergeron:
This PR breaks on jenkins because of this in scan: https://github.com/Theano/Theano/blob/master/theano/scan_module/scan_op.py#L781

This ends up creating a Container with a numpy array as the value and GpuArrayType as the type.  With the new filter_inplace I have this breaks since we assume that old_data is actually a GpuArray and it isn't.  I am pretty sure this is violating something in the contract for Container() or In(), but I'm not sure how to fix it exactly.

abergeron:
Well actually I remember value_zeros() on Type, so we can just use that instead and it does fix the problem.  I hope nothing else breaks though.

theano/scan_module/tests/test_scan.py
```diff
@@ -4956,6 +4956,7 @@ def gpu_from_host(v):
         super(T_Scan_Gpuarray, self).__init__(*args, **kwargs)
 
     def setUp(self):
+        import theano.gpuarray.tests.config
```
nouiz: Can you add a comment why this is needed?

abergeron:
I rebased and fixed the numpys since that has been changed in master.

theano/gpuarray/type.py
```diff
@@ -250,6 +260,12 @@ def filter(self, data, strict=False, allow_downcast=None):
             if b and shp[i] != 1:
                 raise TypeError("Non-unit value on shape on a broadcastable"
                                 " dimension.", shp, self.broadcastable)
+        if not isinstance(data, gpuarray.GpuArray):
+            if old_data is not None and old_data.shape == data.shape:
```
abergeron: I did investigate and fix that in the commits above in scan.

theano/scan_module/tests/test_scan.py
abergeron: ok

PR #5468, opened 2017-01-30
Add exception when epsilon is too small.

abergeron:


theano/gpuarray/dnn_batchnorm.c
```diff
@@ -11,8 +11,10 @@ int dnn_batchnorm_op(PyGpuArrayObject *inp, PyGpuArrayObject *scale,
   if (c_set_tensorNd(scale, bn_params) != 0)
     return 1;
 
-  if (epsilon < 1e-5)
+  if (epsilon < 1e-5) {
+    PyErr_Format(PyExc_ValueError, "epsilon must be at least 1e-5, got %f", epsilon);
```
nouiz: I think this should be done for the 3 ops.

abergeron: done

Issue #5467, opened 2017-01-30
GpuAdvancedSubtensor1 error with libgpuarray


ozancaglayan:
With the latest Theano + libgpuarray, my code is hitting the below error and does not work with the new backend:

https://github.com/Theano/libgpuarray/issues/277

nouiz:
I was able to reproduce it here.

@abergeron, can you look at that?

abergeron:
This should fix the problem: https://github.com/Theano/Theano/pull/5469

ozancaglayan:
OK it works with the first commit. Thanks.

PR #5469, opened 2017-01-30
GpuArray_take1 requires contiguous inputs.

abergeron:
fix gh-5467

### DIFF
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -405,7 +405,7 @@ class GpuAdvancedSubtensor1(HideC, tensor.AdvancedSubtensor1):
     """
     def make_node(self, x, ilist):
         ctx_name = infer_context_name(x, ilist)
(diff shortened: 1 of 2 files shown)


theano/gpuarray/subtensor.py
```diff
@@ -405,15 +405,15 @@ class GpuAdvancedSubtensor1(HideC, tensor.AdvancedSubtensor1):
     """
     def make_node(self, x, ilist):
         ctx_name = infer_context_name(x, ilist)
-        x_ = as_gpuarray_variable(x, ctx_name)
+        x_ = gpu_contiguous(as_gpuarray_variable(x, ctx_name))
```
nouiz: This is too strict. If the outer dimensions isn't strided, it should still work.

This also need tests.

slefrancois:
jenkins test this

nouiz:
test with the index in the reverse order still missing. I'll merge as this fix a bug, but we should add a test that would run on all back-end. The way the current tests are done is easy to do this.

PR #5474, opened 2017-01-31
Follow up from abstract bn PR

nouiz:
I merged https://github.com/Theano/Theano/pull/5190 too fast. Here are some updates for corner cases.

I still need to locally make sure all tests pass and that it don't make the tests too long (also in DebugMode) before merging this PR.

The master isn't in a bad state. Just that corner case won't be as optimized as they should and some tests crash when floatX=float64.

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -28,7 +28,6 @@ from theano.tensor.nnet.abstract_conv import (AbstractConv2d,
                                               assert_conv_shape)
 from theano.tensor.signal.pool import (
     Pool, MaxPoolGrad, AveragePoolGrad)
-from theano.tensor.nnet import bn
 from . import pygpu
 from .type import (get_context, gpu_context_type, list_contexts,
                    GpuArraySharedVariable)
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3030,9 +3030,6 @@ def local_log_softmax_dnn(node):
             node.inputs[0].owner and
             isinstance(node.inputs[0].owner.op, GpuDnnSoftmax) and
             len(node.inputs[0].clients) == 1):
-        if version(raises=False) < 3000:
-            # No log-softmax before cudnn v3
-            return
         softmax_node = node.inputs[0].owner
         new_softmax = GpuDnnSoftmax('log', softmax_node.op.mode)
         return [new_softmax(softmax_node.inputs[0])]
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3100,9 +3100,6 @@ def local_gpua_softmax_dnn_grad(op, ctx_name, inputs, outputs):
     return [out.dimshuffle(0, 2)]
 
 
-@register_opt('cudnn', 'fast_compile')
-@op_lifter([bn.AbstractBatchNormTrain])
-@register_opt2([bn.AbstractBatchNormTrain], 'cudnn', 'fast_compile')
 def local_abstract_batch_norm_train_cudnn(op, ctx_name, inputs, outputs):
     x, scale, bias, epsilon, running_average_factor = inputs[:5]
     running_mean = inputs[5] if len(inputs) > 5 else None
@@ -3186,9 +3183,6 @@ def local_batch_norm_inference_inplace(node):
         return [GpuDnnBatchNormInference(mode=node.op.mode, inplace=True)(*node.inputs)]
 
 
-@register_opt('cudnn', 'fast_compile')
-@op_lifter([bn.AbstractBatchNormTrainGrad])
-@register_opt2([bn.AbstractBatchNormTrainGrad], 'cudnn', 'fast_compile')
 def local_abstract_batch_norm_train_grad_cudnn(op, ctx_name, inputs, outputs):
     x, dy, scale, x_mean, x_invstd, epsilon = inputs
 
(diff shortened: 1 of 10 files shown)


slefrancois:
jenkins test this

theano/gpuarray/dnn.py
```diff
@@ -3037,7 +3032,7 @@ def local_log_softmax_dnn(node):
             len(node.inputs[0].clients) == 1):
         if version(raises=False) < 3000:
```
abergeron: We don't support cudnn older than v4 anyway so this entire check could go away.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -26,6 +26,11 @@
 
 from theano.configdefaults import SUPPORTED_DNN_CONV_ALGO_FWD
 
+mode_with_gpu = mode_with_gpu.including()
+mode_without_gpu = mode_without_gpu.including()
+mode_with_gpu.check_py_code = False
+mode_without_gpu.check_py_code = False
```
abergeron: If you want to change mode_without_gpu globally to not check python code, that would be ok.

theano/sandbox/cuda/dnn.py
```diff
@@ -3705,6 +3705,19 @@ def local_gpu_batch_norm_inference_inplace(node):
                                          inplace=True)(*node.inputs)]
 
 
+def values_eq_approx_high_tol(a, b):
+    """
+    This fct is needed to don't have DebugMode raise useless
+    error due to ronding error.
```
abergeron: error -> errors (the first word, not the second).
ronding -> rounding

theano/sandbox/cuda/dnn.py
```diff
@@ -3705,6 +3705,19 @@ def local_gpu_batch_norm_inference_inplace(node):
...
+
+    This happen as We reduce on the two last dimensions, so this
```
abergeron: We -> we

theano/sandbox/cuda/dnn.py
```diff
@@ -3705,6 +3705,19 @@ def local_gpu_batch_norm_inference_inplace(node):
...
+    can raise the absolute error if the number of element we
```
abergeron: element -> elements

theano/tensor/nnet/bn.py
```diff
@@ -413,15 +413,27 @@ def infer_shape(self, node, shape):
     def make_node(self, x, scale, bias, epsilon=1e-4,
                   running_average_factor=0.1,
                   running_mean=None, running_var=None):
+        x = as_tensor_variable(x)
+        scale = as_tensor_variable(scale)
+        bias = as_tensor_variable(bias)
+        epsilon = as_tensor_variable(epsilon)
```
abergeron: This can cause trouble with floatX=float32 and epsilon=1e-5.  I don't know if we want to address those.

theano/tensor/nnet/tests/test_bn.py
```diff
@@ -201,7 +201,7 @@ def test_batch_normalization_train():
                                               bn.AbstractBatchNormTrainGrad))
                             for n in f.maker.fgraph.toposort()])
             # run
-            for data_shape in ((5, 10, 30, 40, 10), (4, 3, 1, 1, 1), (1, 1, 5, 5, 5)):
```
abergeron: You don't need to change the first shape here?

theano/tensor/type.py
```diff
@@ -204,8 +204,8 @@ def filter_variable(self, other, allow_convert=True):
         Convert a symbolic Variable into a TensorType, if compatible.
 
         For the moment, only a TensorType or CudaNdarrayType will be
```
abergeron: While you're there, GpuArrayType will convert here too.  As will scalars, I think.

theano/tensor/nnet/bn.py
nouiz: I think we want the normal floatX behavior here too. If epsilon don't have a dtype, we use the same algo. I prefer to not introduce special case about our floatX behavior.

theano/tensor/nnet/tests/test_bn.py
nouiz: Not for the CPU tests. I only fixed those that I need to.

theano/tensor/type.py
nouiz: Not scalar for now:

$ python -c "import theano;s=theano.scalar.float32(); ts=theano.tensor.scalar(); ts.type.filter_variable(s)"
  File "<string>", line 1, in <module>
  File "theano/tensor/type.py", line 234, in filter_variable
    self=self))
TypeError: Cannot convert Type float32 (of Variable <float32>) into Type TensorType(float64, scalar). You can try to manually convert <float32> into a TensorType(float64, scalar).


But maybe we should. Anyway, too close to an rc to change that now. I'll update about GpuArrayType.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -26,6 +26,10 @@
 
 from theano.configdefaults import SUPPORTED_DNN_CONV_ALGO_FWD
 
+mode_with_gpu = mode_with_gpu.including()
+# Globally disabled for mode_without_gpu
+mode_with_gpu.check_py_code = False
```
abergeron: What I meant by globally was in config.py directly, but this is fine too.

nouiz: I don't like doing it globally. I'm pretty sure many tests rely on that to check the correspondance of both code. But maybe doing the test in FAST_COMPILE as we do do enough, but I'm not 100% sure of that.

PR #5433, opened 2017-01-19
Update Windows installation documentation for Anaconda.

notoraptor:
Update Windows installation documentation for Anaconda.

NB: there is a conflict between `rdonnelly` package and python 3.6. It seems it's because the package is configured to depend on specific versions of Python (2.7*, 3.4* and 3.5*), so I mention it in the doc.

@nouiz

PR #5479, opened 2017-02-02
(small fix) Ensure Theano finds libgpuarray folders automatically in Windows with conda package

notoraptor:
( this is related to windows documentation PR #5433 )

Theano uses `distutils.sysconfig.get_python_inc()` to get Python headers folder (and a similar function for libraries folder), and on Windows it returns the `include` folder located inside the environment folder (e.g. `C:\donnees\programmes\Anaconda3\envs\python2\include` (PS: same value is returned when calling ` distutils.sysconfig.get_python_inc(plat_specific=True)`)).

But conda installs the `libgpuarray` package under a folder named `Library` inside the environment folder (e.g. `C:\donnees\programmes\Anaconda3\envs\python2\Library\include\gpuarray`). It seems this `Library` folder is specific to Windows, so that it is impossible to get it directly from Python.

This fix ensures that the `Library` folder on Windows is used to find libgpuarray headers and static libraries automatically, so that user has nothing more to do when installing libgpuarray through conda.

Then the windows documentation will not need anything related to "configure libgpuarray" or something like that.

@abergeron 

gpuarray

### DIFF
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -1584,7 +1584,12 @@ def std_include_dirs():
     python_inc_dirs = ([py_inc] if py_inc == py_plat_spec_inc
                        else [py_inc, py_plat_spec_inc])
     gof_inc_dir = os.path.abspath(os.path.dirname(__file__))
-    return numpy_inc_dirs + python_inc_dirs + [gof_inc_dir]
+    other_dirs = []
+    if sys.platform == 'win32':
+        alt_inc_dir = os.path.abspath(os.path.normpath(sys.exec_prefix + '\Library\include'))
+        if os.path.exists(alt_inc_dir) and os.path.isdir(alt_inc_dir):
+            other_dirs.append(alt_inc_dir)
+    return numpy_inc_dirs + python_inc_dirs + [gof_inc_dir] + other_dirs
 
 
 def std_lib_dirs_and_libs():
@@ -1602,6 +1607,9 @@ def std_lib_dirs_and_libs():
         # Also add directory containing the Python library to the library
         # directories.
         python_lib_dirs = [os.path.join(os.path.dirname(python_inc), 'libs')]
+        alt_inc_dir = os.path.abspath(os.path.normpath(sys.exec_prefix + '\Library\lib'))
+        if os.path.exists(alt_inc_dir) and os.path.isdir(alt_inc_dir):
+            python_lib_dirs.append(alt_inc_dir)
         if "Canopy" in python_lib_dirs[0]:
             # Canopy stores libpython27.a and libmsccr90.a in this directory.
             # For some reason, these files are needed when compiling Python
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -1584,12 +1584,7 @@ def std_include_dirs():
     python_inc_dirs = ([py_inc] if py_inc == py_plat_spec_inc
                        else [py_inc, py_plat_spec_inc])
     gof_inc_dir = os.path.abspath(os.path.dirname(__file__))
-    other_dirs = []
-    if sys.platform == 'win32':
-        alt_inc_dir = os.path.abspath(os.path.normpath(sys.exec_prefix + '\Library\include'))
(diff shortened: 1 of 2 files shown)


theano/gof/cmodule.py
```diff
@@ -1584,7 +1584,12 @@ def std_include_dirs():
     python_inc_dirs = ([py_inc] if py_inc == py_plat_spec_inc
                        else [py_inc, py_plat_spec_inc])
     gof_inc_dir = os.path.abspath(os.path.dirname(__file__))
-    return numpy_inc_dirs + python_inc_dirs + [gof_inc_dir]
+    other_dirs = []
+    if sys.platform == 'win32':
+        alt_inc_dir = os.path.abspath(os.path.normpath(sys.exec_prefix + '\Library\include'))
+        if os.path.exists(alt_inc_dir) and os.path.isdir(alt_inc_dir):
+            other_dirs.append(alt_inc_dir)
+    return numpy_inc_dirs + python_inc_dirs + [gof_inc_dir] + other_dirs
```
abergeron: I would rather place this hack in GpuArrayType than here.  That way we won't invoke it for things that don't require the new backend.

theano/gof/cmodule.py
```diff
@@ -1602,6 +1607,9 @@ def std_lib_dirs_and_libs():
         # Also add directory containing the Python library to the library
         # directories.
         python_lib_dirs = [os.path.join(os.path.dirname(python_inc), 'libs')]
+        alt_inc_dir = os.path.abspath(os.path.normpath(sys.exec_prefix + '\Library\lib'))
+        if os.path.exists(alt_inc_dir) and os.path.isdir(alt_inc_dir):
+            python_lib_dirs.append(alt_inc_dir)
```
abergeron: same

abergeron:
Apparently this also is a problem on other platforms so we probably should just always do it.

notoraptor:
Updated.

@mgermain do you have the same problems with your machine ? Could you check where is libgpuarray headers in your environment ?

mgermain:
~/.conda/envs/my_root/include/gpuarray

theano/gpuarray/type.py
```diff
@@ -470,7 +471,17 @@ def c_headers(self):
                 '<gpuarray_api.h>']
 
     def c_header_dirs(self):
-        return [pygpu.get_include(), numpy.get_include()]
+        other_dirs = []
+        alt_inc_dir = os.path.abspath(os.path.normpath(sys.exec_prefix + '/Library/include'))
```
lamblin: On Linux, headers seem to be in `sys.exec_prefix + '/include'` instead.
Same for `lib`, by the way. The whole `/Library` in the path does not seem to be there.

abergeron: Then we might need something system-dependant after all ...

notoraptor: @lamblin the `/Library` path seems to exist only with Conda distribution and only on Windows, so it is more a specific fix for Conda on Windows.

There are indeed `include` and `lib` paths on Linux, but I thought it was correctly included by Theano, maybe not after all ... I'll make some tests on Ubuntu tomorrow.

notoraptor:
I confirm the problem for @mgermain on Ubuntu when using conda package `pygpu`.

`distutils.sysconfig.get_python_inc()` returns

`/home/notoraptor/Programmes/anaconda3/envs/testTheano2/include/python2.7`

, but gpuarray headers are in 

`/home/notoraptor/Programmes/anaconda3/envs/testTheano2/include/gpuarray/`.

So, on both Linux and Windows, the conda package `libgpuarray is installed in some directories that Theano can't find. And directories to check are not the same on both platforms.

I think I will just add another checking for Linux also. But then, I don't know if there is a similar problem on MacOS, for example !


NB: Jenkins and Travis failures are related to flake8 errors, I will fix it.

abergeron:
I think this is normal for conda since Theano is the only package that actually builds things at runtime.  Normally the headers are used in conda-build.

abergeron:
Apparently jenkins failed because the disk is full.

abergeron:
jenkins test this please

Issue #1681, opened 2013-12-20
Faster compilation


nouiz:
This is a meta tickets with a list of possible ways to speed up Theano compilation:
- Allow faster compilation mode
  - [ ]  optimizer_excluding=slow #2470 
  - [ ] doc optimizer=fast_compile,
  - [ ] do optimizer={O0,O1,O2,O3} where O0 is fast_compile and O3 fast_run?
  - This help model development
- [ ] Generate less versions of some c code
  - [ ] In particular Elemwise, one compilation unit can do it for all dtype when all inputs dtype are the same gh-1933
  - [ ] CuDNN descriptor could generate less module.
  - [ ]  Subtensor can also generate less module.
- [ ] Cache the optimization phase. (started, but need someone to finish this: gh-2299)
- OpFromGraph gh-1603
  - This only mitigate this and remove optimization between boundary
  - Optimization to unfold it after partial optimization of the inner graph?
- Make Theano optimization phase faster
  - [ ] Make EQ optimizer loop less frequently by using on_change_input
    - [ ] Implementing the CGT algo will probalby make this not needed.
  - [x] Profile the EQ canonizer phase to know which local opt is slow when it don't do anything.
  - [ ] Find slow compiling code, profile and update the algo to make it faster. gh-2494
- [x] Check that we refresh our cache content when we take the lock in case of compilation in parallel by 2 process
- [x] Allow caching of compiled function and unpickling without recompilation. (via [a Theano flag](http://deeplearning.net/software/theano/library/config.html#config.reoptimize_unpickled_function))
- [ ] Faster cycle detection (gh-4233)
- [x] Partial evaluation of a Theano function gh-2472 (Allow to not recompile a test function after a train function)
- [ ] Move compilation phase to c, cython code?
- [ ] Prepare all modules to be compiled, then compile them in parallel (with configurable number of parallel processes, default to two times the available cores or so) gh-3340

I set this high priority as this is a current concern that affect our users.

Other idea that currently I don't think are good idea to implement:

- [ ]  Asynchronous Theano function compilation gh-3277

NOTE:
 Check those document page to have a better understanding of Theano optimization pipeline: [optimization pipeline](http://deeplearning.net/software/theano_versions/dev/extending/pipeline.html) and [more about optimizations](http://deeplearning.net/software/theano_versions/dev/extending/optimization.html)

Sentient07:
@nouiz Replacing the EquilibriumOptimizer with the CGTOptimizer is related to this issue ?

[...]

PR #5481, opened 2017-02-02
Have "theano-cache list" include compilation times

f0k:
Not sure if this is wanted, but one of the purposes of `theano-cache list` is to see how many Ops have been compiled for a particular function and identify opportunities for reducing the number of compilations (https://github.com/Theano/Theano/issues/1681#issuecomment-276746167). This PR adds the compilation time (the difference between finishing writing the `mod.cpp` or `mod.cu` file and finishing writing the `.so` file) to the list to make it even more useful for that purpose.

### DIFF
--- a/theano/gof/compiledir.py
+++ b/theano/gof/compiledir.py
@@ -122,7 +122,16 @@ def print_compiledir_content():
                 else:
                     types = list(set([x for x in flatten(keydata.keys)
                                       if isinstance(x, theano.gof.Type)]))
-                    table.append((dir, ops[0], types))
+                    compile_start = compile_end = float('nan')
+                    for fn in os.listdir(os.path.join(compiledir, dir)):
+                        if fn.startswith('mod.c'):
+                            compile_start = os.path.getmtime(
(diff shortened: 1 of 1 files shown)


nouiz:
I like the way you got this information. Nice idea. I'm good with the PR, except travis fail with this error:
```
/home/travis/build/Theano/Theano/theano/gof/compiledir.py:129:37: E126 continuation line over-indented for hanging indent
/home/travis/build/Theano/Theano/theano/gof/compiledir.py:132:37: E126 continuation line over-indented for hanging indent
```

f0k:
There was a lot of discussions that pep8's (now pycodestyle's) E126 does not follow PEP-0008. It has been disabled in 1.6: https://github.com/PyCQA/pycodestyle/blob/9f725fb0fad8919a9467834309fbd9f604280035/CHANGES.txt#L151
The problem is that a hanging indent of a single indentation level looks exactly like a control-level indentation following an `if`/`def`/`for` etc.

I'll amend the commit to fix this, but you may want to change your flake8 configuration.

PR #5485, opened 2017-02-03
update abstract_conv.py

MurugeshMarvel:
Discarded unnecessary calling of numpy object twice hence removed one and linked all others to the numpy object.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -17,7 +17,6 @@ from theano.gof import Apply, Op
 from six.moves import xrange
 
 import warnings
-import numpy
 import numpy as np
 
 try:
@@ -69,7 +68,7 @@ def get_conv_output_shape(image_shape, kernel_shape,
     nkern, kshp = kernel_shape[0], kernel_shape[2:]
 
     if filter_dilation is None:
-        filter_dilation = numpy.ones(len(subsample), dtype='int')
+        filter_dilation = np.ones(len(subsample), dtype='int')
 
     if isinstance(border_mode, tuple):
         out_shp = tuple(get_conv_shape_1axis(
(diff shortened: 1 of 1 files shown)


theano/tensor/nnet/abstract_conv.py
```diff
@@ -1613,7 +1612,7 @@ def perform(self, node, inp, out_):
         if isinstance(mode, tuple):
             pad = tuple(int(mode[i]) for i in range(self.convdim))
             mode = "valid"
-            new_img = numpy.zeros((img.shape[0], img.shape[1]) +
+            new_img = np.zeros((img.shape[0], img.shape[1]) +
                                   tuple(img.shape[i + 2] + 2 * pad[i]
```
f0k: You need to dedent all following lines up to the closing bracket by the number of characters you removed (3). Same for the other `np.zeros` instances. Otherwise PEP-0008 is violated.

MurugeshMarvel: but the indentation in ok.. why the PEP-0008 error is emerging

f0k: [No, it's not](https://www.python.org/dev/peps/pep-0008/#indentation):
> Continuation lines should align wrapped elements either vertically using Python's implicit line joining inside parentheses, brackets and braces, or using a hanging indent

This is not a hanging indent, but a visual indent (i.e., vertically aligned with the content of the opening bracket). `tuple` should start on the same column as `(img`, as both are part of the argument to `np.zeros(`. It was like that before you changed `numpy` to `np`, you need to restore this to pass the code style check!

MurugeshMarvel: I have some problems in contributing to theano, i am not good in git. could you help me out

f0k: Ahm, you should do a local clone of the repository. Then checkout your `patch-1` branch (`git checkout branch-1`), modify the file, do `git add` with the file name, then `git commit --amend --no-edit` to amend the previous commit, and `git push --force` to update this PR with the new commit. And close your second PR #5492 to avoid confusion.

nouiz:
this was done in another pr, so closing.

PR #5492, opened 2017-02-05
rectified numpy error

MurugeshMarvel:


theano/tensor/nnet/abstract_conv.py
```diff
@@ -1,5 +1,5 @@
 """
-Abstract conv interface
+Abstract conv interface updated
```
nouiz: Undo this. You didn't change the interface, so it isn't updated. You updated the file, this isn't the same.

lamblin:
Closing, let's continue on the original ticket.

PR #5499, opened 2017-02-06
Add C code for CPU and GPU for tensor.Split op.

notoraptor:
(C code only for CPU currently, C code for GPU to come later)

Method: for each split, an adapted view is created and then copied into an output.

To come up:
* Add C code for GPU.
* Check if `self.len_splits` can be defined as a param op.

@nouiz @lamblin 

close https://github.com/Theano/Theano/pull/5499

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1408,15 +1408,13 @@ class GpuSplit(HideC, Split):
             # There are no outputs, then nothing to do.
             return ''
 
+        # outputs_pointers lists the addresses of the pointers to the outputs.
+        outputs_pointers = '&' + (', &'.join(outputs))
         x, axis, splits = inputs
         fail = sub['fail']
         splits_dtype = node.inputs[2].type.dtype_specs()[1]
         axis_dtype = node.inputs[1].type.dtype_specs()[1]
         expected_splits_count = self.len_splits
-        code_for_updating_output = []  # filled later.
-        code_if_sync = []  # filled later.
-        full_code_for_updating_output = ''  # defined later.
-        full_code_if_sync = ''  # defined later.
 
         main_code = """
         int ndim = PyGpuArray_NDIM(%(x)s);
@@ -1427,7 +1425,8 @@ class GpuSplit(HideC, Split):
         size_t* split_points = NULL;
         GpuArray* split_views = NULL;
         GpuArray** split_views_pointers = NULL;
-        int i;
+        int i, j;
+        PyGpuArrayObject** outputs[] = {%(outputs_pointers)s};
 
         /* Check inputs. */
 
@@ -1493,7 +1492,31 @@ class GpuSplit(HideC, Split):
         }
 
         /* Put split views into outputs. */
-        %(full_code_for_updating_output)s
+        for (i = 0; i < splits_count; ++i) {
+            PyGpuArrayObject** output = outputs[i];
+            Py_XDECREF(*output);
+            *output = pygpu_fromgpudata(
+                split_views[i].data,
+                split_views[i].offset,
+                split_views[i].typecode,
+                split_views[i].nd,
+                split_views[i].dimensions,
+                split_views[i].strides,
+                %(x)s->context,
+                1, // output is writable
+                Py_None, Py_None
+            );
+            if (*output == NULL) {
+                PyErr_SetString(PyExc_RuntimeError, "GpuSplit: unable to update an output from a split view.");
+                for (j = 0; j < splits_count; ++j) {
+                    GpuArray_clear(split_views_pointers[j]);
+                }
+                free(split_views_pointers);
+                free(split_views);
+                free(split_points);
+                %(fail)s
+            }
+        }
 
         /* Free memory. */
         for (i = 0; i < splits_count; ++i) {
@@ -1502,43 +1525,14 @@ class GpuSplit(HideC, Split):
         free(split_views_pointers);
         free(split_views);
         free(split_points);
-
-        /* Code added if synchronization is enabled. */
-        %(full_code_if_sync)s
         """
 
-        for split_index, output in enumerate(outputs):
-
-            code_for_updating_output.append("""
-            Py_XDECREF(%(output)s);
-            %(output)s = pygpu_fromgpudata(
-                split_views[%(split_index)s].data,
-                split_views[%(split_index)s].offset,
-                split_views[%(split_index)s].typecode,
-                split_views[%(split_index)s].nd,
-                split_views[%(split_index)s].dimensions,
-                split_views[%(split_index)s].strides,
-                %(x)s->context,
-                1, // output is writable
-                Py_None, Py_None
-            );
-            if (%(output)s == NULL) {
-                PyErr_SetString(PyExc_RuntimeError, "GpuSplit: unable to update an output from a split view.");
-                for (i = 0; i < splits_count; ++i) {
-                    GpuArray_clear(split_views_pointers[i]);
-                }
-                free(split_views_pointers);
-                free(split_views);
-                free(split_points);
-                %(fail)s
-            }
-            """ % locals())
-
-            if config.gpuarray.sync:
-                code_if_sync.append("GpuArray_sync(&%(output)s->ga);" % locals())
-
-        full_code_for_updating_output = '\n'.join(code_for_updating_output)
-        full_code_if_sync = '\n'.join(code_if_sync)
+        if config.gpuarray.sync:
+            main_code += """
+        for (i = 0; i < splits_count; ++i) {
+            GpuArray_sync(&((*outputs[i])->ga));
+        }
+        """
 
         return main_code % locals()
 
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1377,6 +1377,13 @@ class GpuSplit(HideC, Split):
     Split for GPU.
 
     """
+    def __init__(self, len_splits):
+        super(GpuSplit, self).__init__(len_splits)
+        # The GPU version of Split returns splits as views of the input.
+        self.view_map = {}
+        for i in xrange(self.len_splits):
+            self.view_map[i] = [0]
+
     def make_node(self, x, axis, splits):
         node = Split.make_node(self, x, axis, splits)
         x = as_gpuarray_variable(x, infer_context_name(x))
@@ -1388,7 +1395,7 @@ class GpuSplit(HideC, Split):
     # we reuse the perform of the CPU op, which is suitable
 
     def c_code_cache_version(self):
-        return (2,)
+        return (3,)
 
     def c_headers(self):
         return ['<numpy_compat.h>', '<gpuarray_helper.h>']
@@ -1421,7 +1428,6 @@ class GpuSplit(HideC, Split):
         int splits_count = PyArray_SIZE(%(splits)s);
         size_t len_along_axis, sum_of_splits = 0, current_split_start = 0;
         %(splits_dtype)s current_split_length = 0;
-        size_t* split_dims = NULL;
         GpuArray view;
         int i;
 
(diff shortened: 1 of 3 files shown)


notoraptor:
Updated. C code for GPU added.

To come up:
* re-read the code.
* check if using splits count as a params op can be useful.
* do more profiling.

At first glance, C codes are faster than Python code (about 10 times), but current GPU code does not seem so much faster than CPU code. I test using `theano/tensor/tests/test_basic.py:T_Join_and_Split` and `nosetests -vs theano/gpuarray/tests/test_basic_ops.py:G_Join_and_Split`.

Also, about using splits count as a params op, I feel it won't be so efficient for the C code, because the current interface of split always generates as many output nodes as splits count. So it will be always necessary to generate pieces of C codes in some loops to manage all output nodes, as each output node has it own variable name ( @nouiz ).

@abergeron

notoraptor:
### Current profiling
___

I run the following script 2 times (first to compile C codes, second to profile):

```bash
# Python profiling.
THEANO_FLAGS=floatX=float32,profile=true,cxx= nosetests theano/tensor/tests/test_basic.py:T_Join_and_Split > python.log.txt 2>&1

# CPU profiling.
THEANO_FLAGS=floatX=float32,profile=true nosetests theano/tensor/tests/test_basic.py:T_Join_and_Split > cpu.log.txt 2>&1

# GPU profiling.
CUDA_LAUNCH_BLOCKING=1 THEANO_FLAGS=floatX=float32,profile=true nosetests  theano/gpuarray/tests/test_basic_ops.py:G_Join_and_Split > gpu.log.txt 2>&1
```

#### Complete profiling for Python: [python.log.txt](https://github.com/Theano/Theano/files/757786/python.log.txt)
Excerpt:
```
# ...
Ran 34 tests in 10.518s

OK
# ...
Function profiling
==================
  Message: Sum of all(84) printed profiles at exit excluding Scan op profile.
# ...
   3.3%    91.3%       0.002s       1.09e-04s     Py      22      12   theano.tensor.basic.Split
# ...
```

#### Complete profiling for CPU: [cpu.log.txt](https://github.com/Theano/Theano/files/757787/cpu.log.txt)
Excerpt:
```
# ...
Ran 34 tests in 16.327s

OK
# ...
Function profiling
==================
  Message: Sum of all(84) printed profiles at exit excluding Scan op profile.
# ...
   2.2%    91.4%       0.000s       9.94e-06s     C       22      12   theano.tensor.basic.Split
# ...
```

#### Complete profiling for GPU: [gpu.log.txt](https://github.com/Theano/Theano/files/757789/gpu.log.txt)
Excerpt:
```
Using cuDNN version 5105 on context None
# ...
Ran 35 tests in 46.887s

OK
# ...
Function profiling
==================
  Message: Sum of all(85) printed profiles at exit excluding Scan op profile.
# ...
   0.0%   100.0%       0.000s       1.69e-05s     C       22      12   theano.gpuarray.basic_ops.GpuSplit
# ...
```

So, currently, CPU C code is aproximaetely 10 times faster than Python code, and a very little bit faster than GPU C code.

notoraptor:
NB: It seems `GpuArray_split()` does not compute as the way I expect (it calls `GpuArray_index` which calls `Gpuarray_view()` on the outputs I have already allocated, so this may lead to memory leaks). I'll rewrite the GPU code carefully.

notoraptor:
Updated to avoid memory leaks on GPU. But now GPU is very slow.

Python profiling: [python.log.txt](https://github.com/Theano/Theano/files/759014/python.log.txt)
```
   3.2%    91.1%       0.002s       1.02e-04s     Py      22      12   theano.tensor.basic.Split
```

CPU profiling: [cpu.log.txt](https://github.com/Theano/Theano/files/759015/cpu.log.txt)
```
   2.2%    91.7%       0.000s       1.00e-05s     C       22      12   theano.tensor.basic.Split
```

GPU profiling: [gpu.log.txt](https://github.com/Theano/Theano/files/759016/gpu.log.txt)
```
   9.9%    99.5%       0.818s       3.72e-02s     C       22      12   theano.gpuarray.basic_ops.GpuSplit
```

GPU is even slower than Python. I think it comes from `GpuArray_move()` because it copies views that are not contiguous.

I'll re-read code again to look for solutions or other issues.

@abergeron

notoraptor:
New update. `GpuArray_move()` seems too slow when data that are copied are not contiguous (this happens frequently, as we create splits as views and then we copy it, but these views are not contiguous in memory), so I have removed the call to this function.

Instead:
* The GPU op is now created with a `view_map`, so that Theano knows the outputs of this op are views on the first input.
* The outputs are created with as few memory as possible (with `ndims=0`). This step can't be skipped, and some memory is always allocated by creation methods.
* The outputs are cleared and then reallocated as split views of the input.

This strategy generates a chain of allocation-deallocation-reallocation for each output, but it is faster than copying the split views, and I try to minimize the amount of memory allocated in the first time.

The profiling is now similar to the first I published. Python is slower than CPU (about 10 times faster), and GPU a little slower than CPU.

Python profiling: [python.log.txt](https://github.com/Theano/Theano/files/761203/python.log.txt)
```
   3.2%    90.9%       0.002s       1.08e-04s     Py      22      12   theano.tensor.basic.Split
```

CPU profiling: [cpu.log.txt](https://github.com/Theano/Theano/files/761205/cpu.log.txt)
```
   2.5%    90.9%       0.000s       1.32e-05s     C       22      12   theano.tensor.basic.Split
```

GPU profiling: [gpu.log.txt](https://github.com/Theano/Theano/files/761206/gpu.log.txt)
```
   0.0%   100.0%       0.000s       1.73e-05s     C       22      12   theano.gpuarray.basic_ops.GpuSplit
```

notoraptor:
I think the codes can be reviewed from now, @abergeron @lamblin @nouiz .

theano/gpuarray/basic_ops.py
```diff
@@ -1377,15 +1377,140 @@ class GpuSplit(HideC, Split):
...
+            }
+            """ % locals())
+
+            code_for_splitting.append("""
+            current_split_length = * (%(splits_dtype)s*) PyArray_GETPTR1(%(splits)s, %(split_index)s);
+            view.offset = PyGpuArray_STRIDE(%(x)s, axis) * current_split_start;
+            view.dimensions[axis] = current_split_length;
+            GpuArray_fix_flags(&view);
+            GpuArray_clear(&%(output)s->ga);
+            if (GpuArray_view(&%(output)s->ga, &view) != GA_NO_ERROR) {
```
abergeron: This part is almost as costly as doing Py_XDECREF() followed by pygpu_view().

It also introduces the only op in all of the new backend for which the gpudata pointer can change even if the python object didn't.  Scan relies on the new backend not doing this, so it will break in interesting ways (as in not fun to debug silent wrong data).

Also if the goal here is to reuse output memory then it fails, because this frees the output memory anyway.

So this code manages to
a) be significantly more complex,
b) not do anything inplace,
c) have some cases wrong and
d) break scan.

For all of these reasons I am vehemently against merging.

theano/gpuarray/basic_ops.py
```diff
@@ -1377,15 +1377,140 @@ class GpuSplit(HideC, Split):
...
+
+    def c_headers(self):
+        return ['<numpy_compat.h>', '<gpuarray_helper.h>']
+
+    def c_header_dirs(self):
+        return [pygpu.get_include(), os.path.dirname(__file__)]
+
+    def c_support_code(self):
+        # We want to hide the definition of this method from tensor.Split.
+        return ''
```
abergeron: HideC does that already.

theano/gpuarray/basic_ops.py
```diff
@@ -1377,15 +1377,140 @@ class GpuSplit(HideC, Split):
...
         x = as_gpuarray_variable(x, infer_context_name(x))
         outs = [GpuArrayType(dtype=o.dtype, broadcastable=o.broadcastable,
                              context_name=x.type.context_name)()
                 for o in node.outputs]
         return Apply(self, [x] + node.inputs[1:], outs)
+
     # we reuse the perform of the CPU op, which is suitable
 
+    def c_code_cache_version(self):
+        return (3,)
```
abergeron: Why start at 3?

theano/gpuarray/basic_ops.py
```diff
@@ -1377,15 +1377,140 @@ class GpuSplit(HideC, Split):
...
+        /* Code added if synchronization is enabled. */
+        %(full_code_if_sync)s
+        """
+
+        for split_index, output in enumerate(outputs):
+            # When checking output, we allocate a PyGpuArrayObject with 0 dims
+            # (that is, an object with as few memory as possible), as its GpuArray field
+            # will be cleared and re-used as a view during split operations.
+            codes_for_checking_outputs.append("""
+            if (theano_prep_output(&%(output)s, 0, NULL, %(x_typecode)s, x_order, %(x)s->context) != 0) {
```
abergeron: This function is useful if you intend on actually using the memory buffer and it can allocate new ones for you if required.  Here you just free whatever is provided as output and replace it with a view so this is wasteful.

theano/gpuarray/basic_ops.py
```diff
@@ -1377,15 +1377,140 @@ class GpuSplit(HideC, Split):
...
+                PyErr_SetString(PyExc_RuntimeError, "GpuSplit: unable to prepare an output.");
+                %(fail)s
+            }
+            """ % locals())
+
+            code_for_splitting.append("""
+            current_split_length = * (%(splits_dtype)s*) PyArray_GETPTR1(%(splits)s, %(split_index)s);
+            view.offset = PyGpuArray_STRIDE(%(x)s, axis) * current_split_start;
```
abergeron: What if x already has an offset?

theano/gpuarray/basic_ops.py
```diff
@@ -1377,15 +1377,140 @@ class GpuSplit(HideC, Split):
...
+            };
+            current_split_start += current_split_length;
+            """ % locals())
+
+            if config.gpuarray.sync:
+                code_if_sync.append("GpuArray_sync(&%(output)s->ga);" % locals())
+
+        full_code_for_checking_outputs = '\r\n'.join(codes_for_checking_outputs)
+        full_code_for_splitting = '\r\n'.join(code_for_splitting)
+        full_code_if_sync = '\r\n'.join(code_if_sync)
```
abergeron: We use unix line endings, so only '\n'.

theano/gpuarray/basic_ops.py
notoraptor: It is the iterations on the code from commit to commit. I'll put it back to 1.

theano/gpuarray/basic_ops.py
notoraptor: Indeed ...

notoraptor:
Updated, @abergeron .

Always the same profiles (Python the slowest (`1.09e-04s`), CPU the fatest (`1.16e-05s`), GPU a little slower than CPU (`1.33e-05s`)).

notoraptor:
jenkins test this

nouiz:
Do you profile with profile.ignore_first_call=True?

It could be that some compilatin time is included in the current timing.

notoraptor:
@nouiz with `profiling.ignore_first_call=true` I currently get these results:

For CPU:
```
   4.4%    97.6%       0.000s       7.96e-06s     C       10      10   theano.tensor.basic.Split
```
For GPU:
```
   0.3%    99.0%       0.000s       1.17e-05s     C       10      10   theano.gpuarray.basic_ops.GpuSplit
```

So, CPU still remains faster than GPU.

PS: For previous profiles, I ran the tests a first time to let Theano compile the C codes, then I ran the tests a second time and I got the results from this second run. So it should have been relevant even if I didn't use `ignore_first_call`, isn't it ?

PS 2: I also notice that, when using `ignore_first_call` with theano unit tests, some functions may not appear at all as some ones are called only once in unit tests. But here fortunately, `Split` is re-used many times enough.

nouiz:
libgpuarray compile code at run time. THis is why we added
ignore_first_call. Sadly, we don't compile all c code during the
compilation now.

slefrancois:
jenkins test this

theano/tensor/basic.py
```diff
@@ -3778,6 +3778,156 @@ def R_op(self, inputs, eval_points):
...
+        %(full_code_for_checking_outputs)s
+
+        /* Compute split. */
+
+        %(full_code_for_splitting)s
+
+        free(split_dims);
+        """
+
+        for split_index, output in enumerate(outputs):
```
nouiz: The code is good, but it use the old way of doing it that make it much harder to read.

A better way to code it is to make the for loops in C. THis make the code not a mix of python a C. It also help work around some compiler bug when we unroll too much that and hit limit of the C compiler. The way to not introduce new bug in Theano due to compiler limit is to do in C a vector of all the outputs and iterate on it. Here is the pseudo code to show it:

"""PyArrayObject * outputs[] = {%(outputs)s};
for(int i =0 ; i++; i< N){
   //use outputs[i]
}
""" % dict(outputs=",".join(outputs)

I'll let @abergeron review the GPU code.

notoraptor:
Done, @nouiz !

nouiz:
@abergeron, what about doing a GpuArray_split() that would do the ugly optimization internally?

I made an issue in gpuarray about that: https://github.com/Theano/libgpuarray/issues/354

We should not forget that the gpuarray code don't reuse existing output.

abergeron:
GpuArray_split already exists and is called here.  The problem is at the barrier between python and C.  I'm thinking of changing the structure of the python object to hold a pointer to a GpuArray instead of embedding the struct.  That or expose the functionality to create uninitialized python objects.

It is not really possible to reuse the output if you want to do a view.

PR #5502, opened 2017-02-06
IO_toposort

ReyhaneAskari:
This fixes a minor error generated in docs for #5042 .

### DIFF
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -652,6 +652,8 @@ class MergeFeature(object):
                 continue
             if len(node.inputs) != len(candidate.inputs):
                 continue
+            if node.op != candidate.op:
+                continue
 
             cand_has_assert = False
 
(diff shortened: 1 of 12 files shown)


theano/gof/opt.py
```diff
@@ -679,7 +681,7 @@ def process_node(self, fgraph, node):
                                in zip(node_inputs_assert_removed,
                                       cand_inputs_assert_removed))
 
-            if inputs_match and node.op == candidate.op:
+            if inputs_match:
```
nouiz: Revert both changes in this file. This should be a separate PR that will need its own timing.

nouiz:
The PR contain a merge commit. Can you do a rebase to get rid of it? We prefer to not have merge commits in PR.

ReyhaneAskari:
Apparently there is a problem with the version of gpuarray that is used by Jenkins. We can get back to this after the fix.

ReyhaneAskari:
jenkins test this

nouiz:
jenkins test this

nouiz:
Can you rebase this PR on the current master?

PR #5505, opened 2017-02-06
Mixed

abergeron:
- Test for #5469
- Disable the mkl warning on conda if we do find a working blas.

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -1170,6 +1170,7 @@ AddConfigVar('cmodule.age_thresh_use',
 
 def default_blas_ldflags():
     global numpy
+    warn_record = []
     try:
         if (hasattr(numpy.distutils, '__config__') and
                 numpy.distutils.__config__):
@@ -1284,7 +1285,7 @@ def default_blas_ldflags():
             import mkl  # noqa
         except ImportError as e:
             if any([m for m in ('conda', 'Continuum') if m in sys.version]):
-                _logger.warning('install mkl with `conda install mkl-service`: %s', e)
+                warn_record.append(('install mkl with `conda install mkl-service`: %s', e))
         else:
             # This branch is executed if no exception was raised
             if sys.platform == "win32":
@@ -1326,6 +1327,10 @@ def default_blas_ldflags():
         res = try_blas_flag(ret)
         if res:
             return res
+
+        for warn in warn_record:
+            _logger.warning(*warn)
+        del warn_record
 
         # Some environment don't have the lib dir in LD_LIBRARY_PATH.
         # So add it.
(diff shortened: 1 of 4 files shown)


theano/tensor/tests/test_subtensor.py
```diff
@@ -491,6 +491,8 @@ def test_ok_list(self):
                           # Test with TensorConstant index.
                           (rand(4, 2, 3),
                            theano.tensor.constant([3, 3, 1, 1, 2, 2, 0, 0])),
+                          # Test with non-contiguous input
+                          (rand(6, 2)[::2], [1, 0])
```
nouiz: This do not generate non-contiguous index. [1, 0] will be converted to a numpy array that will be contibuous.

I think you need to make a new test just for that.

abergeron: We don't support non-contiguous index.

nouiz: The goal of the new test is to test when the graph generate a non-contiguous index, do we work correctly? Do we add the gpu_contiguous to assert we are contiguous in the graph?

This was the bug we had. I don't think this new test detect that. The first fix added an gpu_contiguous on the input, not just the index, but we looked together at the code and it wasn't requested. So you removed that from the PR. https://github.com/Theano/Theano/pull/5469/files only add gpu_contiguous for the index, not for the input.

theano/configdefaults.py
```diff
@@ -1327,6 +1328,10 @@ def default_blas_ldflags():
         if res:
             return res
 
+        for warn in warn_record:
```
nouiz: This commit seem good as it only silence the warning when we can reuse numpy blas. By to help the maintenance, can you add this comment that explain what we want to do or similar:

    If we are using conda and can't reuse numpy blas, then doing the fallback and test -lblas could give slow computation, so warn about this. But if we can link to numpy blas like with nomkl packages of conda, don't work, it should be a good enough blas.

abergeron:
Ok, test is better now.

slefrancois:
jenkins test this

theano/sandbox/cuda/tests/test_basic_ops.py
```diff
@@ -1093,6 +1093,9 @@ def test_adv_sub1_fast(self):
             self.assertTrue(val.ndim == data.ndim)
             utt.assert_allclose(val, good)
 
+    def test_noncontiguous_idx(self):
+        raise SkipTest("test doesn't work here")
```
nouiz: THe problem was that the test always use float64.

But this is the old back-end, so ok.

nouiz:
jenkins don't test the last commit. so merging now.

PR #5508, opened 2017-02-07
Force scan n_steps on the CPU.

nouiz:
This fix crash during execution as the cython scan was expecting a CPU value. We need it on the CPU anyway, so force that value to be always on the CPU.


theano/scan_module/scan_op.py
```diff
@@ -349,7 +349,8 @@ def make_node(self, *inputs):
         assert n_outer_ins == n_inner_ins, \
             ("The number of inputs given to the inner function of scan"
              " does not match the number of inputs given to scan.")
-        new_inputs = [inputs[0]]
+        # For the inputs to be on the CPU
```
MarcCote: For -> Force?

nouiz: done

PR #5520, opened 2017-02-09
Improve the example for strict=True in the scan() doc part 2.

Thrandis:
Finishes  #4942

I addressed @nouiz's comment and added the error message in the doc.


nouiz:
travis have an error: https://travis-ci.org/Theano/Theano/jobs/200087305#L1483

PR #5521, opened 2017-02-09
Added missing softsign import

mgermain:


nouiz:
This would slow down the import of theano, but it is probably good.

@abergeron @lamblin do you agree?

mgermain:
Well, I think it is just missing. Either have all the activation functions there or none?

abergeron:
Yes I am ok.

If importing an additional symbol from an already-imported file results in any measurable differnece to the total import time of Theano, then you should stop trying to run python on a 8086.

lamblin:
I don't think it is going to slow down anything, as `sigm.py` is already parsed and compiled in order to import other symbols.

abergeron:
Or at least I would be if the softsign function was actually in the file this is trying to import it from.  softsign is in nnet, not sigm.

PR #5522, opened 2017-02-09
Speed up one scan opt

nouiz:


slefrancois:
jenkins test this

cc13ny:
In theano/scan_module/scan_opt.py, is the line 1821 the duplicate of line 1826?

nouiz:
Good catch. I'll fix that in another PR.

thanks

PR #5527, opened 2017-02-10
validate numpy1.12 compatibility changes

lamblin:
This is just a rebase of #5461 to see if the issue happens with the main cache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
 language: python
 python:
   - "2.7"
-  - "3.4"
+  - "3.3"
 
 # NB:
 # In before_install and install sections below,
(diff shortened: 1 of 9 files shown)


slefrancois:
jenkins test this

lamblin:
Splitting off `test_abstract_conv.py` in a new PART seems to have worked for Travis, and the tests pass.
I'm going to revert the changes to the numpy and scipy version so it can be merged.
@nouiz or @slefrancois could you please have a look at my commits?

nouiz:
Seem ok after the revert of the version changes and disabling of blas check.

lamblin:
I think I finally got down to the cache issue.
I'm trying once again with numpy 1.12, and if it passes, I'll revert the changes to travis again.

lamblin:
So everything works fine, as long as the non-conda packages for scipy and numpy are not installed in Travis' cache (in which case it makes linking with blas fail), which I did by temporarily removing `$HOME/miniconda2` from Travis' cache (but still keeping `$HOME/.theano` in there).

I'm confident everything is fine now, so I'm restoring `.travis.yml` and clearing the cache for that PR.
If someone can do a final review of 56a71e7, that would be great.

lamblin:
Actually, Jenkins was complaining as well.
I'll have a look at the GPU tests...

PR #5533, opened 2017-02-10
Scan compile/grad speed up

nouiz:
When the inner of scan is big, this should give good speed up.

We remove useless clone in the L_op and R_op. It also remove useless clone during the Scan.__init__() method. So this should also speed up opt that create new Scan node.

### DIFF
--- a/theano/scan_module/scan_op.py
+++ b/theano/scan_module/scan_op.py
@@ -207,6 +207,10 @@ class Scan(PureOp):
         if self.info['gpu'] or self.info['gpua']:
             self._hash_inner_graph = self.info['gpu_hash']
         else:
+            # gof don't know about SharedVariable. So add them as inputs.
+            for var in theano.gof.graph.inputs(self.outputs, self.inputs):
+                if var not in self.inputs and not isinstance(var, theano.Constant):
+                    raise theano.gof.MissingInputError("ScanOp is missing an input.")
             self._cmodule_key = gof.CLinker().cmodule_key_variables(self.inputs,
                                                                     self.outputs,
                                                                     [])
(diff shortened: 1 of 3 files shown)


nouiz:
jenkins test this

nouiz:
I rebased to have the CI run on the current master to be sure all is good. If they pass, I'll merge as it was approuved before the rebase.

theano/scan_module/scan_op.py
```diff
@@ -207,10 +207,10 @@ def tensorConstructor(broadcastable, dtype):
         if self.info['gpu'] or self.info['gpua']:
             self._hash_inner_graph = self.info['gpu_hash']
         else:
-            tmp_in, tmp_out = scan_utils.reconstruct_graph(self.inputs,
-                                                           self.outputs)
-            # This is actually required for the line just after.
-            gof.FunctionGraph(tmp_in, tmp_out, clone=False)
+            # gof don't know about SharedVariable. So add them as inputs.
```
abergeron: This comment refers to SharedVariable, but the code below deals with constants.

theano/scan_module/tests/test_scan.py
```diff
@@ -4488,8 +4488,8 @@ def _scan_loose(x):
                                n_steps=n,
                                strict=True)
 
-        f_strict = theano.function([x0_], ret_strict[0][-1])
-        result_strict = f_strict(x0)
+#        f_strict = theano.function([x0_], ret_strict[0][-1])
+#        result_strict = f_strict(x0)
```
abergeron: Remove, don't comment.

nouiz: both done.

Issue #3751, opened 2015-12-06
Compiles during import


simonzack:
I've installed the master branch of theano to get some cudnn compatibility fixes, now theano always compiles something on import which can take around 5 seconds, even when debug mode is enabled. This slows things down a bit when doing unit tests. Is this compilation necessary each time?

nouiz:
We shouldn't compile at every import. But we do it the first time we import
the GPU back-end.

We do some test for old driver bug, there is a new Theano flag to stop this:

https://github.com/Theano/Theano/pull/3718

Would this help you?

simonzack:
Thanks that works.

PR #5537, opened 2017-02-12
Allow for pooling of empty batch (like for conv in #3715).

lucasb-eyer:
Similarly to #3715 allowing size-zero batches to be sent through cuDNN conv (essentially bypassing the call to cuDNN and allocating a size-zero batch output of the correct size), this allows size-zero batches to be sent through cuDNN pooling.

Comes with unit-tests and I have used this change in a real-world application for a couple weeks already.

### DIFF
--- a/theano/sandbox/cuda/dnn.py
+++ b/theano/sandbox/cuda/dnn.py
@@ -1673,9 +1673,6 @@ if (!CudaNdarray_is_c_contiguous(%(input)s)) {
   %(fail)s
 }
 
-if (c_set_tensorNd(%(input)s, %(input_desc)s) != 0)
-  %(fail)s
-
 int win[%(nd)d];
 int pad[%(nd)d];
 int str[%(nd)d];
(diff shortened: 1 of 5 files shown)


nouiz:
jenkins test this.

Good timing, we want to do an RC tomorrow.

theano/sandbox/cuda/dnn.py
```diff
@@ -1710,6 +1707,24 @@ def c_code(self, node, name, inputs, outputs, sub):
   %(fail)s
 }
 
+// if input batch is empty, we return the empty output without calling cuDNN
+// (which will fail on zero batch size)
+if (CudaNdarray_DIMS(%(input)s)[0] == 0) {
+  cudaError_t err2 = cudaMemset((%(out)s)->devdata, 0,
+                                CudaNdarray_SIZE(%(out)s) * sizeof(real));
```
nouiz: This call is useless, as out will have a shape of (0, ....) and so the total array size is 0.

theano/sandbox/cuda/dnn.py
```diff
@@ -1731,6 +1746,8 @@ def c_code(self, node, name, inputs, outputs, sub):
                cudnnGetErrorString(err));
   %(fail)s
 }
+
+}
```
nouiz: As you didn't changed the indentation, add a comment to tell which if it is matching to help reader of the code.

theano/sandbox/cuda/dnn.py
```diff
@@ -1937,20 +1954,34 @@ def c_code(self, node, name, inputs, outputs, sub):
...
                             CudaNdarray_HOST_DIMS(%(output)s)) != 0)
 {
   %(fail)s
 }
 
+// if input batch is empty, we return the empty output without calling cuDNN
+// (which will fail on zero batch size)
+if (CudaNdarray_DIMS(%(input)s)[0] == 0) {
+  cudaError_t err2 = cudaMemset((%(output)s)->devdata, 0,
+                                CudaNdarray_SIZE(%(output)s) * sizeof(real));
```
nouiz: idem here and for the comment.

lucasb-eyer:
Haha no worries I didn't realize about the RC! No need to squeeze this in.

Good point about not needing to memset size-0 arrays. This was taken over from the conv-code linked in #3751, so I guess it's superflous there too. I changed the control flow accordingly; tests still pass on my machine. Also added the comment you requested.

About the old/new back-end, I didn't even know there was such a thing, so I have no idea. Could you point me to the code and tests? I'll see if I can do it, but I'm pretty busy this week and next, so if you think it's faster/better for you to do it, just go ahead!

nouiz:
About the new back-end, this wiki page tell how to use it:

https://github.com/Theano/Theano/wiki/Converting-to-the-new-gpu-back-end%28gpuarray%29

nouiz:
I ported the change to the new back-end.

nouiz:
jenkins test this

lucasb-eyer:
Awesome, thank you for doing the port, I wasn't sure when I'd have time to do it!

nouiz:
I rebased the PR to make flake8 tests and amended the last commit to make this work.

nouiz:
jenkins test this

nouiz:
jenkins test this

Issue #4078, opened 2016-02-17
Gradients through random sampling should be Null for MRG_RandomStreams distributions


lamblin:
The gradients wrt variables passed to methods of `MRG_RandomStreams` (for instance normal, binomial, ...) should be undefined / NaN (NullType), since this is not mathematically defined.

If users want to use some reparametrization trick, then they would have to, for instance, sample from a normal(0, 1), and then multiply by the variance and add the mean. In this case, it would become clear what gradients are expected.

The way of doing that would be to wrap the arguments in [`grad_undefined`](http://deeplearning.net/software/theano/library/gradient.html#theano.gradient.grad_undefined) before starting the chain of operations transforming the uniform sample into the sample of the required distribution.

Sentient07:
@lamblin , I'd like to take this up ? Shall I go ahead ?

lamblin:
I think we are currently lacking the time to keep up with your PRs (I think you have 4 at the moment), so maybe when we are done with some of the existing ones, you could start on this ticket?
Sorry for the slow response, but we have other duties as well...

Sentient07:
@lamblin Yes, i understand :). Of that 4, i guess 3 are done and one of them is remaining. Review at your ease :smile:

lamblin:
@SinaHonari This is your new ticket.

lamblin:
So, the plan would be:
- create a new UndefinedGrad Op, inheriting from ViewOp, in theano/gradient.py. It would be similar to ZeroGrad or DisconnectedGrad, except that the grad() method would return grad_undefined(...) for each input.
- instantiate that Op for each of the inputs (bias, variance, ...) of the random functions that are the methods of the MRG_RandomStreams class (uniform, normal, multinomial, ...)

lamblin:
Fixed by #5538.

PR #5538, opened 2017-02-13
Setting gradient for MRG_RandomStreams distribution variables to be null

SinaHonari:
fixes issue #4078

### DIFF
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -15,6 +15,7 @@ from six.moves import xrange
 import theano
 from theano import Op, Apply, shared, config, Variable
 from theano import gradient, function
+from theano.gradient import undefined_grad
 from theano import tensor
 from theano.tensor import (TensorType, as_tensor_variable, get_vector_length,
                            cast, opt, scal)
@@ -773,7 +774,9 @@ class MRG_RandomStreams(object):
 
         """
         low = as_tensor_variable(low)
+        low = undefined_grad(low)
         high = as_tensor_variable(high)
+        high = undefined_grad(high)
         if dtype is None:
             dtype = scal.upcast(config.floatX, low.dtype, high.dtype)
 
@@ -821,6 +824,7 @@ class MRG_RandomStreams(object):
                  nstreams=None):
         # TODO : need description for method, parameter and return
         if n == 1:
+            p = undefined_grad(p)
             x = self.uniform(size=size, nstreams=nstreams)
             return cast(x < p, dtype)
         else:
@@ -852,6 +856,7 @@ class MRG_RandomStreams(object):
         if pvals is None:
             raise TypeError("You have to specify pvals")
         pvals = as_tensor_variable(pvals)
+        pvals = undefined_grad(pvals)
         if size is not None:
             if any([isinstance(i, integer_types) and i <= 0 for i in size]):
                 raise ValueError(
@@ -978,7 +983,9 @@ class MRG_RandomStreams(object):
         # second half our U2's. See Wikipedia page:
         # http://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform
         avg = as_tensor_variable(avg)
+        avg = undefined_grad(avg)
         std = as_tensor_variable(std)
+        std = undefined_grad(std)
 
         if dtype is None:
             dtype = scal.upcast(config.floatX, avg.dtype, std.dtype)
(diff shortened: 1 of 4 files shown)


lamblin:
Thanks for the PR!
I think there are 3 more changes needed:
- there is a conflict with the base branch, probably due to another PR that replaced the name `multinomial_without_replacement` by `choice` (to be more consistent with numpy and the numpy rng in Theano)
- The following line shows an error in flake8:
   ```
   [grad_undefined(self, i, inp[i]) for i in xrange(args)]
   ```
   but it is not only an issue in flake8, that line would fail if we try to execute it
- Since the line above did not fail, it means i was never called in the test suite. Can you please add some test cases, to check that trying to take the gradient raise the appropriate exception (`NullTypeGradError` I think).

SinaHonari:
@lamblin 
I added some tests. They are passing for uniform, binomial and normal distributions. But for
multinomial and choice, I'm getting this error when using theano variables for probabilities:

('Cannot convert [[<TensorType(float64, scalar)>, <TensorType(float64, scalar)>]] to TensorType', <type 'list'>)

for multinomial it works without theano variable for pvals, but if I use pvals=[[theano_var1, theano_var2]], then I get the above error.

lamblin:
I just realized that the tests in `test_rng_mrg.py` are not executed on Travis, I think it is because [this line](https://github.com/SinaHonari/Theano/blob/f3f98db4dbe5e65fc1a7d4f4fc30a13c8c43a3ca/theano/sandbox/tests/test_rng_mrg.py#L21) raises `SkipTest` at import time, and not only for the tests that needs the new back-end.

Could you please:
- make that import conditional (only import it if pygpu is available)
- fix the issue with `pvals=[[theano_var1, theano_var2]]`, by using `join` or `as_tensor_variable` on the first list (inner dimension), like we discussed.

Thanks

SinaHonari:
@lamblin 
I'll do that. I have a deadline this Friday and work on it afterwards.

SinaHonari:
@lamblin I just updates the PR. Let's see if it passes the tests.

theano/sandbox/rng_mrg.py
```diff
@@ -1388,6 +1391,7 @@ def binomial(self, size=None, n=1, p=0.5, ndim=None, dtype='int64',
                  nstreams=None):
         # TODO : need description for method, parameter and return
         if n == 1:
+            p = undefined_grad(p)
```
lamblin: I think you need to call `as_tensor_variable` on `p` before that, as `undefined_grad` expect a `Variable`, but in some cases `p` is a Python `float`, so tests are failing.

theano/sandbox/tests/test_rng_mrg.py
```diff
@@ -1189,6 +1190,71 @@ def test_validate_input_types_gpuarray_backend():
         rstate = gpuarray_shared_constructor(rstate)
```
lamblin: This test should be skipped if the new back-end is not available, which is why it is failing on Travis.

SinaHonari:
@lamblin I just updated the PR. I'm not sure, I did the right thing for skipping the test.

lamblin:
Jenkins test this please

lamblin:
Tryin to re-run on Travis.

theano/sandbox/tests/test_rng_mrg.py
```diff
@@ -1186,9 +1187,75 @@ def test_validate_input_types_gpuarray_backend():
 
     with change_flags(compute_test_value="raise"):
         rstate = numpy.zeros((7, 6), dtype="int32")
-        rstate = gpuarray_shared_constructor(rstate)
+        if theano.gpuarray.pygpu:
+            rstate = gpuarray_shared_constructor(rstate)
         mrg_uniform.new(rstate, ndim=None, dtype="float32", size=(3,))
 
+
+def test_udefined_grad():
```
lamblin: "undefined"

theano/sandbox/tests/test_rng_mrg.py
lamblin: I think it would be better to import `theano.gpuarray.tests.config` in here, or just `raise SkipTest('pygpu not installed')` here, so the test does not get executed when it is not necessary.

SinaHonari:
@lamblin I updated the PR. Let's see if it's all fine.

SinaHonari:
@lamblin If I import theano.gpuarray.tests.config where you mentioned flake8 complains 

test_rng_mrg.py:1189:13: F401 'theano.gpuarray.tests.config' imported but unused

nouiz:
You can add "  # noqa" to have that line not checked by flake8

SinaHonari:
@lamblin The tests are now passing. I had to import theano inside test method to stop Travis from complaining Theano is not defined.

nouiz:
This PR need a rebase.

lamblin:
I rebased and pushed, it should be OK now.

lamblin:
Jenkins test this please

Issue #5541, opened 2017-02-14
GpuCAReduceCuda RuntimeError: 'Could not initialize elemwise support'


jc-Rosichini:
Hi ,
I'm running into a runtime error that seems related to GpuCAReduceCuda
This happens only when using cuda0 mode (it's ok in cpu and gpu modes)

(1) my config:
Windows 10
Python Anaconda 3.5
Theano 0.9.0beta1dev
Lasagne 0.2dev1
pygpy 0.6.0 and latest libgpuarray

(2) the crash:     

The code compiles ok both in cpu and/or gpu and cuda (new libgpuarray) modes

Then at runtime in cuda mode:
```
 train_fn = theano.function([net.input_var, net.target_var], [loss, I, U, acc], updates=updates
 
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\function.py", line 326, in function
    output_keys=output_keys)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\pfunc.py", line 486, in pfunc
    output_keys=output_keys)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\function_module.py", line 1795, in orig_function
    defaults)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\function_module.py", line 1661, in create
    input_storage=input_storage_lists, storage_map=storage_map)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\link.py", line 699, in make_thunk
    storage_map=storage_map)[:3]
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\vm.py", line 1063, in make_all
    impl=impl))
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\op.py", line 924, in make_thunk
    no_recycling)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\op.py", line 828, in make_c_thunk
    output_storage=node_output_storage)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\cc.py", line 1190, in make_thunk
    keep_lock=keep_lock)
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\cc.py", line 1131, in __compile__
    keep_lock=keep_lock)
[...]

Issue #5463, opened 2017-01-27
libgpuarray / cudnn issues


RyanHope:
I once had a working cuda/theano/libgpuarry setup working on windows 10 but since a reinstall I have not been able to get a python setup working with cuda and libgpuarray. Using both Anaconda's python and WinPython, when using the old backend, the common theano gpu test works.

```
python D:\test.py
WARNING (theano.sandbox.cuda): The cuda backend is deprecated and will be removed in the next release.  Please switch to the gpuarray backend. You can get more information about how to switch at this URL:
 https://github.com/Theano/Theano/wiki/Converting-to-the-new-gpu-back-end%28gpuarray%29

Using gpu device 0: GeForce GTX 1060 3GB (CNMeM is enabled with initial size: 10.0% of memory, cuDNN 5110)
[GpuElemwise{exp,no_inplace}(<CudaNdarrayType(float32, vector)>), HostFromGpu(GpuElemwise{exp,no_inplace}.0)]
Looping 1000 times took 0.402000 seconds
Result is [ 1.23178029  1.61879349  1.52278066 ...,  2.20771813  2.29967761
  1.62323296]
Used the gpu
```

But when I switch to using the libgpuarray backend cudnn can no longer be found and gpu is not used.

```
python D:\test.py
Can not use cuDNN on context None: cannot compile with cuDNN. We got this error:
c:\users\ryanhope\appdata\local\temp\try_flags_zq9bx6.c:4:19: fatal error: cudnn.h: No such file or directory
compilation terminated.

Mapped name None to device cuda: GeForce GTX 1060 3GB (0000:01:00.0)
[GpuElemwise{exp,no_inplace}(<GpuArrayType<None>(float32, (False,))>), HostFromGpu(gpuarray)(GpuElemwise{exp,no_inplace}.0)]
Looping 1000 times took 0.433000 seconds
Result is [ 1.23178029  1.61879349  1.52278066 ...,  2.20771813  2.29967761
  1.62323296]
Used the cpu
```

FWIW, I can import pygpu no problem but pygpu.test() eventually crashes.

abergeron:
If you build the packages yourself, you must make sure that you use the same compiler for the library as the one that was used for the python binary otherwise you can encounter weird crashes.

As for why the header is not found you can try setting the Theano flags `dnn.include_path` and `dnn.library_path` to the path where you installed the files.  The difference is probably that the old backend includes the cuda folders by default and the new one doesn't.

[...]

PR #5547, opened 2017-02-14
Consistently use '\n' in outputted code

valtron:
All generated C code uses `\n` line separators, except some in `gof/op.py` when on Windows. This fixes it.

Strangely, this also fixes the issue I described in [#5463](/Theano/Theano/issues/5463#issuecomment-279826487).


abergeron:
This might also fix #5541

nouiz:
I pushed flake8 fix to this pr.

PR #5548, opened 2017-02-14
Add automatic detection of log handlers

joaovictortr:
- Add has_handlers method to detect when a log handler has already been
  added (e.g. root log handler);
- Add disable_log_handler method to disable theano_logger's default log
  handler.

Signed-off-by: João Victor Tozatti Risso <joaovictor.risso@gmail.com>

### DIFF
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -31,13 +31,35 @@ import logging
 
 import sys
 
+def has_handlers(logger):
+    # copied from Logger.hasHandlers() (introduced in Python 3.2)
+    _logger = logger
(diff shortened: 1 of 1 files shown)


abergeron:
Leave it for after the release. Even if this seems perfectly all right, we don't want additional problems with the RC.

nouiz:
@abergeron do you think it look good and you agree to merge after the release?

abergeron:
Yes, absolutely.

nouiz:
thanks, now that the release is done, we can merge.

PR #4780, opened 2016-07-25
Fix NaNGuardMode to permit None variables.

davidbau:
None values in the theano tree can appear in array indexing
such as A[:, None], but NaNGuardMode currently throws an
exception when it requirest a None value.  This simple
change fixes it, and adds a relevant test.

PR #5550, opened 2017-02-15
 Fix NaNGuardMode to permit None variables

nouiz:
finish https://github.com/Theano/Theano/pull/4780

### DIFF
--- a/theano/compile/nanguardmode.py
+++ b/theano/compile/nanguardmode.py
@@ -23,6 +23,36 @@ except ImportError:
 logger = logging.getLogger("theano.compile.nanguardmode")
 
 
+def _non_numeric_value(var):
+    """
+    Checks a variable against non-numeric types such as types, slices,
+    empty arrays, and None, that need not be checked for NaN and Inf values.
+
+    Parameters
+    ----------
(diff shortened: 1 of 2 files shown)


nouiz:
jenkins test this

theano/compile/nanguardmode.py
```diff
@@ -23,6 +23,37 @@
...
+    is_non_numeric : bool
+        `True` the value is non-numeric.
+
+    """
+    if isinstance(arr, theano.gof.type._cdata_type):
+        return True
+    elif isinstance(arr, np.random.mtrand.RandomState):
+        return True
+    elif var and getattr(var.tag, 'is_rng', False):
+        return False
```
lamblin: Should be `True` here.

theano/compile/nanguardmode.py
```diff
@@ -23,6 +23,37 @@
 logger = logging.getLogger("theano.compile.nanguardmode")
 
 
+def _non_numeric_value(arr, var):
```
lamblin: I think it would be more clear not to invert the logic.

nouiz:
both comment are done.

PR #5553, opened 2017-02-15
Fix invalid pointer casts and negative stride handling in sparse

rebecca-palmer:
SamplingDotCsr is [broken on negative-stride arrays](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=855102) because it passes a first-in-index-order pointer to a BLAS interface that expects first-in-memory-order.  Fix this and add a negative stride test case.

Both it and UsmmCscDense also cast npy_intp* (usually int64*) to int32*, which is a strict aliasing violation (i.e. technically undefined behaviour: it currently appears to work, but might not in future compilers).  (It's also broken on big-endian systems, which is [how Debian discovered the problem](https://bugs.debian.org/831541), but those are now rare.)  Fix this by casting values instead of pointers.

(These have been tested by running sparse/tests/test_basic.py from the source tree, but *not* in a full build.)

### DIFF
--- a/theano/sparse/opt.py
+++ b/theano/sparse/opt.py
@@ -831,6 +831,10 @@ class UsmmCscDense(gof.Op):
             npy_intp Sptr = PyArray_STRIDES(%(x_ptr)s)[0] / PyArray_DESCR(%(x_ptr)s)->elsize;
             npy_intp Sy = PyArray_STRIDES(%(y)s)[1] / PyArray_DESCR(%(y)s)->elsize;
 
+            // blas expects ints; convert here (rather than just making N etc ints) to avoid potential overflow in the negative-stride correction
+            int N32 = N;
+            int Sy32 = Sy;
+            int Szn32 = Szn;
 
             if (!(%(inplace)s))
             {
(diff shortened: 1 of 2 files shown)


theano/sparse/opt.py
```diff
@@ -1892,15 +1896,27 @@ def c_code(self, node, name, inputs, outputs, sub):
...
                     const dtype_%(x)s* x_row = (dtype_%(x)s*)(PyArray_BYTES(%(x)s) + PyArray_STRIDES(%(x)s)[0] * m);
 
                     const dtype_%(y)s* y_col = (dtype_%(y)s*)(PyArray_BYTES(%(y)s) + PyArray_STRIDES(%(y)s)[0] * n);
+                    // dot expects pointer to the beginning of memory arrays,
+                    // so when the stride is negative, we need to get the
+                    // last element
+                    if (Sdx < 0)
+                        x_row += (K - 1) * Sdx;
+                    if (Sdy < 0)
+                        y_col += (K - 1) * Sdy;
```
nouiz: If you change the ptr, should you also change the stride? I suppose you checked that, but I want to be sure.

Otherwise, this PR look good. I don't think we will merge it for the RC or Theano 0.9. How bad is this for Debian?

rebecca-palmer: The stride stays negative because that's how BLAS knows to start from the last-in-memory element.

Debian has [already applied these fixes locally](https://anonscm.debian.org/cgit/debian-science/packages/theano.git/commit/?id=7f0a0528eb4802eb3e0cbec47bbd1198a3a44403): they have a "packages must build on as many of our supported architectures (which include big-endian ones) as reasonably possible" policy, which makes the second part a worse bug for them than for x86-only users.

I'd consider the first part the worse bug, since it produces silently wrong answers, but don't know how likely a negative-stride array (followed by memory the Theano process also owns, such as another array - if it doesn't, you'll get a segfault, which at least lets you know *something* is wrong) is in practice.

lamblin:
Jenkins test this please

nouiz:
This is only on the CPU with the sparse variables. Why start jenkins?

PR #5559, opened 2017-02-16
[CRASH] Fix crash of GpuElemwise that have too many inputs

nouiz:
Sadly, this is too strict in some case. As we don't know the shapes, we don't know if the dtype used will be int32 or int64 for the element numbers. So we must be more strict.

This is the same in the old back-end.

Fix: [theano-dev] Current State of Compile Performance / Am I Doing Something Wrong?

fix https://github.com/Theano/Theano/issues/5541


theano/gpuarray/elemwise.py
```diff
@@ -41,6 +41,27 @@ def get_scal(dt):
     return scalar.get_scalar_type(dt)
 
 
+def max_inputs_to_GpuElemwise(node_or_outputs):
+    if isinstance(node_or_outputs, gof.Apply):
+        outputs = node_or_outputs.outputs
+    else:
+        outputs = node_or_outputs
+    ptr_size = 8
+    int_size = 4
```
abergeron: Set this to 8 to be extra sure.  GpuElemwise will compile a 64-bits version.

nouiz: I did it for extra safety, but I think we start to limit too much fusion. This + the collapsing raise how much fusion we can do.

Not sure how high a priority this is...

theano/gpuarray/opt.py
```diff
@@ -752,26 +753,38 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
...
 
-    return max_nb_inputs
+    """
+    if node.op.scalar_op in (scalar.add, scalar.mul):
+        max_nb_inputs = max_inputs_to_GpuElemwise(node)
+        if max_nb_inputs <= 1 and len(node.inputs) > 1:
+            return False
+        while len(node.inputs) > max_nb_inputs:
+            inner_op = []
+            for i in xrange(0,
```
lamblin: This would fail on python 3, and flake8 complains.
Importing `xrange` from `six.moves` should work.

lamblin:
The console log for Jenkins says: "theano.gpuarray.tests.test_pool.test_pool2d ... Command terminated by signal 9"
Was it stopped on purpose?

lamblin:
The error in Jenkins was because the RAM of the VM had been busted (8 GB).
@nouiz thinks that a non-trivial amount may have been used by compiled elemwise kernels of the new test (`test_many_arg_elemwise`), and that lowering the number of cases tested could solve the issue.

lamblin:
@abergeron @nouiz could you take a look at the last changes?

nouiz:
I'm ok with your change, so merging. This will allow the daily buildbot to run with that fix.

PR #5562, opened 2017-02-17
Speed merge optimizer and allow partial optimizer profile

nouiz:
- MergeOptimizer don't call validate when merging constant
- SequenceOptimizer still print the partial profile.
- tensor.constant() don't create intermediate TensorConstant when it won't be used. (code clean up and help find where in the code constant are created)

### DIFF
--- a/theano/gof/tests/test_graph.py
+++ b/theano/gof/tests/test_graph.py
@@ -361,7 +361,7 @@ class TestAutoName:
         r3 = tensor.constant(1.6)
         # The cache still create a new object that we don't return.
         # This is why we must increase by 2 and not 1.
-        assert r3.auto_name == "auto_" + str(autoname_id + 2)
+        assert r3.auto_name == "auto_" + str(autoname_id + 1)
 
     def test_tensorvariable(self):
         # Get counter value
--- a/theano/gof/tests/test_graph.py
+++ b/theano/gof/tests/test_graph.py
@@ -359,8 +359,6 @@ class TestAutoName:
         assert r1 is r2
 
         r3 = tensor.constant(1.6)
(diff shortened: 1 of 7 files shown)


theano/gof/tests/test_graph.py
```diff
@@ -361,7 +361,7 @@ def test_constant(self):
         r3 = tensor.constant(1.6)
         # The cache still create a new object that we don't return.
         # This is why we must increase by 2 and not 1.
-        assert r3.auto_name == "auto_" + str(autoname_id + 2)
+        assert r3.auto_name == "auto_" + str(autoname_id + 1)
```
abergeron: Why did this change?  If it is required, please adjust the comment above.

nouiz: done.

Mostly, the old code was creating a tmp variable in tensor.constant() and not returining it when it is in the cache. Now we only create a variable if it is not in the cache. So now a call to constant() don't always increment the number.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1781,7 +1782,7 @@ def __init__(self,
                                                        filter_dilation=filter_dilation)
 
     # Update shape/height_width
-    def make_node(self, img, topgrad, shape):
+    def make_node(self, img, topgrad, shape, add_assert=True):
```
lamblin: Maybe call it `assert_shapes` or `add_assert_shape`, I think the name is too vague.
Also, it would need to be documented in the docstring, especially because it is a parameter of `make_node` that is not a symbolic input of the node.

nouiz:
I rebased this PR since the merge of the the PR that was part of this one.

PR #5567, opened 2017-02-17
Adding scipy.special modified Bessel function ops

matt-graham:
Adding scalar and elemwise ops for modified Bessel function of order 0 and 1 from `scipy.special`. I've tried to follow the format and naming conventions as used for the Bessel function of first kind ops added in #3883.

### DIFF
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -1,9 +1,11 @@
 from __future__ import absolute_import, print_function, division
-# definition theano.scalar op that have their python implementation taked from scipy
-# as scipy is not always available, we treat them separatly
+# Definitions of theano.scalar ops that have their python implementation taken
+# from SciPy. As SciPy is not always available, we treat them separately.
+
 import numpy
 
 import theano
+from theano.gradient import grad_not_implemented
 from theano.scalar.basic import (UnaryScalarOp, BinaryScalarOp,
                                  exp, upgrade_to_float,
                                  upgrade_to_float64,
@@ -373,9 +375,36 @@ class Chi2SF(BinaryScalarOp):
 chi2sf = Chi2SF(upgrade_to_float64, name='chi2sf')
 
 
+class Jv(BinaryScalarOp):
+    """
+    Bessel function of the first kind of order v (real).
+    """
+
+    @staticmethod
+    def st_impl(v, x):
+        return scipy.special.jv(v, x)
+
+    def impl(self, v, x):
+        if imported_scipy_special:
+            return self.st_impl(v, x)
+        else:
+            super(Jv, self).impl(v, x)
+
+    def grad(self, inputs, grads):
+        if imported_scipy_special:
+            v, x = inputs
+            gz, = grads
+            return [grad_not_implemented(self, 0, v),
+                    gz * (jv(v - 1, x) - jv(v + 1, x)) / 2.]
+        else:
+            super(Jv, self).grad(v, x)
+
+jv = Jv(upgrade_to_float, name='jv')
+
+
 class J1(UnaryScalarOp):
     """
-    Bessel function of the 1'th kind
+    Bessel function of the first kind of order 1.
     """
 
     @staticmethod
@@ -388,8 +417,13 @@ class J1(UnaryScalarOp):
         else:
             super(J1, self).impl(x)
 
-    def grad(self, inp, grads):
-        raise NotImplementedError()
+    def grad(self, inputs, grads):
+        if imported_scipy_special:
+            x, = inputs
+            gz, = grads
+            return [gz * (j0(x) - jv(2, x)) / 2.]
+        else:
+            super(J1, self).grad(v, x)
 
     def c_code(self, node, name, inp, out, sub):
         x, = inp
@@ -398,12 +432,13 @@ class J1(UnaryScalarOp):
             return """%(z)s =
                 j1(%(x)s);""" % locals()
         raise NotImplementedError('only floating point is implemented')
+
 j1 = J1(upgrade_to_float, name='j1')
 
 
 class J0(UnaryScalarOp):
     """
-    Bessel function of the 0'th kind
+    Bessel function of the first kind of order 0.
     """
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


theano/tensor/tests/test_basic.py
```diff
@@ -1735,6 +1737,8 @@ def test_py_c_match():
     expected_chi2sf = []
     expected_j0 = []
     expected_j1 = []
+    expected_i0 = []
+    expected_i1 = []
```
nouiz: You need to add the tests it self. Look in this file at J1Tester and J1InplaceTester for an example.

nouiz:
Also, this is specialized version of the some generic bessel function. Ideally, it would be great to add in Theano an optimization that would convert the generic bessel to this one. Do you know what is the generic bessel and the parameter that match those new specialized fct?

matt-graham:
The modified Bessel functions  [are related](https://en.wikipedia.org/wiki/Bessel_function#Modified_Bessel_functions:_I.CE.B1_.2C_K.CE.B1) to the already present Bessel functions but the relationship is via an imaginary argument to the Bessel functions and the current implementations don't support complex inputs. 

There are also general (non-integer) order versions of the modified and non-modified Bessel functions of the first kind - [`jv`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.jv.html#scipy.special.jv) and [`iv`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.iv.html#scipy.special.iv), which the current `j0` / `j1` and proposed `i0` / `i1` are special cases of. I could wrap the `scipy.special` implementation of these? I think this would also allow differentiation of arbitrary order Bessel functions as the derivatives are defined in terms of (modified) Bessel functions of +/- 1 order.

theano/tensor/tests/test_basic.py
```diff
@@ -1954,6 +1954,50 @@ def test_py_c_match():
...
+    expected=expected_i0,
+    good=_good_broadcast_unary_i,
+    grad=_good_broadcast_unary_i,
+    eps=2e-10,
+    mode=mode_no_scipy,
+    skip=skip_scipy)
+
+I0InplaceTester = makeBroadcastTester(
+    op=inplace.i0_inplace,
+    expected=expected_j0,
```
nouiz: typo, expected_i0

this is what make travis fail I think.

Otherwise, it look good.

matt-graham: Thanks, now corrected.

theano/tensor/tests/test_basic.py
matt-graham: Thanks, now added.

nouiz:
When you add new commits, this don't create a notification, so it takes more time for me to see it.

Anyway, with the release, I would probably not have the time to review it. I start to review your more recent changes.

theano/scalar/basic_scipy.py
```diff
@@ -388,8 +417,13 @@ def impl(self, x):
...
 
-    def grad(self, inp, grads):
-        raise NotImplementedError()
+    def grad(self, inputs, grads):
+        if imported_scipy_special:
+            x, = inputs
+            gz, = grads
+            return [gz * (j0(x) - jv(2, x)) / 2.]
+        else:
+            super(J1, self).grad(v, x)
```
nouiz: The other grad() method in this file don't use "if imported_scipy_special". Why did you use it? If there isn't a specific reason, remove the if. The other op already do the check.

This is for all the ops.

theano/tensor/tests/test_basic.py
```diff
@@ -1937,14 +1965,89 @@ def test_py_c_match():
...
+    grad=_grad_broadcast_unary_bessel,
+    eps=2e-10,
+    mode=mode_no_scipy,
+    inplace=True,
+    skip=skip_scipy)
+
+IvTester = makeBroadcastTester(
+    op=tensor.iv,
+    expected=expected_iv,
+    good=_good_broadcast_binary_bessel,
```
nouiz: You forgot to test the grad here and bellow

theano/tensor/tests/test_basic.py
```diff
@@ -1937,14 +1965,89 @@ def test_py_c_match():
...
+    grad=_grad_broadcast_unary_bessel,
+    eps=2e-10,
+    mode=mode_no_scipy,
+    inplace=True,
+    skip=skip_scipy)
+
+JvTester = makeBroadcastTester(
+    op=tensor.jv,
+    expected=expected_jv,
+    good=_good_broadcast_binary_bessel,
```
nouiz: For both Jv tester, the grad isn't tested.

matt-graham: I wasn't sure how to implement the test for the gradient such that only the gradient with respect to the second input is tested (as the gradient with respect to the first input does not have a closed form solution and there isn't a numerical implementation available in `scipy.special` I left it not implemented).

I read through [the guide in the documentation](http://deeplearning.net/software/theano/extending/unittest.html#validating-the-gradient) and tried wrapping the op call in a function which fixes the first input but then couldn't figure out how to integrate this in to the `makeBroadcastTester` interface. I could add separate `test_verify_*_grad` functions to test the gradient separately if it is acceptable to have plain functions as test cases rather than instances of `unittest.TestCase`?

theano/scalar/basic_scipy.py
matt-graham: Thanks, not entirely sure why I added these but have now removed all the unnecessary `if imported_scipy_special` checks from the `grad` methods.

theano/tensor/tests/test_basic.py
matt-graham: I have added `test_ verify_jv_grad` and `test_verify_iv_grad` functions to `test_basic.py` which wrap the ops to fix the first inputs as described in the documentation. I could not figure out how to do this correctly for the inplace versions though - when running `nosetests` on equivalent functions for the inplace ops, either an exception was raised that the (fixed) first input was non-destructible as a constant, or when tried to workaround this by making the passed first input a shared variable rather than constant, the grad numerical results have a large error. Is there a correct way to verify gradients of inplace ops with a subset of inputs fixed?

nouiz:
There isn't a big use cases to test the grad of the inplace. this is already tested by testing the forward. I'll make an issue about this, maybe we should remove all of those tests!

I'm good to merge when CI pass.

nouiz:
thanks

PR #5568, opened 2017-02-17
Clarify error message related to expected kern_shape specs (1D, int64) in GpuDnnConvDesc

notoraptor:
In new back-end, method `GpuDnnConvDesc.make_node()` expects its input `kern_shape` to be an int64 1D tensor, but on Windows machines it seems that integers are created as int32 by default. This kind of error has already been encountered, for example here: https://github.com/pymc-devs/pymc3/pull/1550 .

On Windows it make crash this test:

```
nosetests -vs theano/gpuarray/tests/test_dnn.py:TestDnnInferShapes.test_conv_gradw_full_conv
```

For the moment, I suggest to:
* modify a line in `test_dnn` so that kern_shape is always casted to `int64`,
* clarify error message in GpuDnnConvDesc.

@nouiz @abergeron

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -386,7 +386,7 @@ class GpuDnnConvDesc(COp):
 
     def make_node(self, kern_shape):
         if kern_shape.type.ndim != 1 or kern_shape.type.dtype != 'int64':
-            raise TypeError('kern must be 1D shape tensor')
+            raise TypeError('kern must be an int64 1D shape tensor')
 
         node = Apply(self, [kern_shape],
                      [CDataType("cudnnConvolutionDescriptor_t",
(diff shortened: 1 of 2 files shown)


theano/gpuarray/dnn.py
```diff
@@ -386,7 +386,7 @@ def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv',
 
     def make_node(self, kern_shape):
         if kern_shape.type.ndim != 1 or kern_shape.type.dtype != 'int64':
-            raise TypeError('kern must be 1D shape tensor')
+            raise TypeError('kern must be an int64 1D shape tensor')
```
nouiz: if kern_shape.type.ndim != 1 or kern_shape.dtype not in theano.tensor.basic.int_dtypes:
            raise TypeError('kern must be an int64 1D shape tensor')
        kern_shape = kern_shape.cast('int64')

Remove the change in the test.

We want to upcast int to int64 here to prevent user annoyance here.

PR #5569, opened 2017-02-17
Fix Runtime error with float64 precision.

### LINKED ISSUE #5555: cudnn theano gpuarray dnn.py float64 git master
Hi,
i am using 

- Ubuntu 16.04
- Theano Developer Installation (commit  fa78b21 )
- Cuda 8.0
- cudnn 5.1
- gcc 4.9 and gcc 5.4
- python 3.5.2
- nvidia 367.57 for nvidia 1080 and 
- on an other machine nvidia 367.48 for the nvidia TitanX

i am running into following RuntimeError when using floatX = float64 (the default value)
using the dnn python interface (it works with float32):
CUDNN_STATUS_NOT_SUPPORTED 

more precise:
``RuntimeError: error getting worksize: CUDNN_STATUS_NOT_SUPPORTED
Apply node that caused the error: GpuDnnConvGradI{algo='none', inplace=True}(GpuContiguous.0, GpuContiguous.0, GpuAllocEmpty{dtype='float64', context_name=None}.0, GpuDnnConvDesc{border_mode='half', subsample=(1, 1), conv_mode='conv', precision='float32'}.0, Constant{1.0}, Constant{0.0})
Toposort index: 3479
Inputs types: [GpuArrayType<None>(float64, (False, False, False, False)), GpuArrayType<None>(float64, (False, False, False, False)), GpuArrayType<None>(float64, (False, False, False, False)), <theano.gof.type.CDataType object at 0x7f309d3cfe48>, Scalar(float64), Scalar(float64)]
Inputs shapes: [(7, 64, 1, 1), (3, 7, 240, 320), (3, 64, 240, 320), 'No shapes', (), ()]
Inputs strides: [(512, 8, 8, 8), (4300800, 614400, 2560, 8), (39321600, 614400, 2560, 8), 'No strides', (), ()]
Inputs values: ['not shown', 'not shown', 'not shown', <capsule object NULL at 0x7f308b7fd480>, 1.0, 0.0]
Outputs clients: [[GpuElemwise{Composite{((i0 * i1) + (i2 * i1 * sgn(i3)))}}[(0, 1)]<gpuarray>(GpuArrayConstant{[[[[ 0.5]]]]}, GpuDnnConvGradI{algo='none', inplace=True}.0, GpuArrayConstant{[[[[ 0.5]]]]}, if{inplace,gpu}.0)]]``

in gpuarray/dnn.py the initialization and the default value in gpu_dnn_conv_desc should be float64:

```--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -351,7 +351,7 @@ class GpuDnnConvDesc(COp):
         return False
 
     def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv',
-                 precision="float32"):
+                 precision="float64"):
         COp.__init__(self, ["conv_desc.c"], "APPLY_SPECIFIC(conv_desc)")
 
         if isinstance(border_mode, integer_types):
@@ -440,7 +440,7 @@ class GpuDnnConvDesc(COp):
 
 
 def gpu_dnn_conv_desc(border_mode, subsample=(1, 1), conv_mode='conv',
-                      precision="float32"):
+                      precision="float64"):
     key = (border_mode, subsample, conv_mode, precision)
     if key not in gpu_dnn_conv_desc.cache:
         gpu_dnn_conv_desc.cache[key] = GpuDnnConvDesc(border_mode,
```

after those changes  it worked.

souravsingh:
Fixes the Runtime error when using ``floatX=float64``.
Fixes #5555

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -361,7 +361,7 @@ class GpuDnnConvDesc(COp):
         return False
 
     def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv',
-                 precision="float32", precision="float64"):
+                 precision="float64"):
         COp.__init__(self, ["conv_desc.c"], "APPLY_SPECIFIC(conv_desc)")
 
         if isinstance(border_mode, integer_types):
@@ -450,7 +450,7 @@ class GpuDnnConvDesc(COp):
 
 
 def gpu_dnn_conv_desc(border_mode, subsample=(1, 1), conv_mode='conv',
-                      precision="float32", precision="float64"):
+                      precision="float64"):
     key = (border_mode, subsample, conv_mode, precision)
     if key not in gpu_dnn_conv_desc.cache:
         gpu_dnn_conv_desc.cache[key] = GpuDnnConvDesc(border_mode,
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -361,7 +361,7 @@ class GpuDnnConvDesc(COp):
         return False
 
     def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv',
-                 precision="float32"):
+                 precision="float32", precision="float64"):
         COp.__init__(self, ["conv_desc.c"], "APPLY_SPECIFIC(conv_desc)")
 
         if isinstance(border_mode, integer_types):
@@ -450,7 +450,7 @@ class GpuDnnConvDesc(COp):
 
 
 def gpu_dnn_conv_desc(border_mode, subsample=(1, 1), conv_mode='conv',
-                      precision="float32"):
+                      precision="float32", precision="float64"):
     key = (border_mode, subsample, conv_mode, precision)
     if key not in gpu_dnn_conv_desc.cache:
         gpu_dnn_conv_desc.cache[key] = GpuDnnConvDesc(border_mode,


nouiz:
jenkins test this

theano/gpuarray/dnn.py
```diff
@@ -361,7 +361,7 @@ def do_constant_folding(self, node):
         return False
 
     def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv',
-                 precision="float32"):
+                 precision="float32", precision="float64"):
```
nouiz: This will fail. You have 2 time the same keyword precision. Can you fix that rapidly?

Also, why do you want to use float64 on GPU? Do you know that most GPU have float64 very slow? Mostly out of curiosity and to stay up to date on people usage of Theano.

lamblin: Also, I don't think this is the right place to fix it. `float32` is a reasonable default.
What we should do is check where `gpu_dnn_conv_desc` is called, and explicitly call it with the dtype corresponding to the precision of that particular convolution.

souravsingh: The Issue mentions that changing the initialization and default value worked for the user, with a patch added into the issue #5555.

lamblin:
`dnn_conv` and `dnn_conv3d` have some logic to determine the precision in which the convolution should be done.
What needs to be done is:
- if possible, factor out that logic into a separate helper function
- add that logic to `dnn_grad{weight,input}{,3d}`, as well as a "precision" keyword.
- check the other places that call `GpuDnnConvDesc` in `theano/gpuarray` to see if they need to pass the appropriate precision.

nouiz:
https://github.com/Theano/Theano/pull/5650 fix it.

PR #5571, opened 2017-02-17
Fix imports in HPCS examples

souravsingh:
Fixes NumPy imports in the examples.


nouiz:
This is the old tutorial, we prefer to don't change them. Thanks for the PR.

Sentient07:
I think the new backend follows `import numpy as np` style of import. @souravsingh to make the repository even, you could make a PR that fixes the imports in `theano/gpuarray` directory if @nouiz is fine with the proposed changes.

nouiz:
@Sentient07 if your suggestion is to do this change inside the theano/gpuarray directory, then yes.

PR #5574, opened 2017-02-18
Gpuarray elemwise

lamblin:
Copy of #5559 without the test, to help us figure out what is going on with Jenkins.

### DIFF
--- a/theano/gpuarray/elemwise.py
+++ b/theano/gpuarray/elemwise.py
@@ -42,24 +42,44 @@ def get_scal(dt):
 
 
 def max_inputs_to_GpuElemwise(node_or_outputs):
+    """
+    Compute the maximum number of inputs that fit in a kernel call.
+    """
     if isinstance(node_or_outputs, Apply):
         outputs = node_or_outputs.outputs
     else:
         outputs = node_or_outputs
+
+    n_out = len(outputs)
+    ndim = outputs[0].type.ndim
+
     ptr_size = 8
-    # We compile code that would do indexing in int64
+    # Even with call32, the interface does not change, and shapes,
+    # strides, and offset are passed as 64-bits (8 bytes)
     int_size = 8
 
     # we take the limit from CUDA for now
-    argument_limit = 232
-    ndim = outputs[0].type.ndim
-    # number of elements and shape
-    size_param_mandatory = (int_size * (ndim + 1)) + \
-        (ptr_size + int_size * ndim) * len(outputs)
+    nb_bytes_total = 4096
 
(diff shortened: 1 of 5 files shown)


lamblin:
Jenkins test this please

lamblin:
Jenkins test this please

lamblin:
Jenkins test this please

lamblin:
The issue seems unrelated to this PR, the failing test is indeed `theano/gpuarray/tests/test_pool.py:test_pool2d`, and I can reproduce it locally.
A call to `theano.function` does not seem to return, and the memory usage grows by gigabytes.
I'm trying to bisect.

lamblin:
According to `git bisect`, the offending commit is b006605057261c0429d1274018c58ea66fc1a9f2.
I'm still not sure how that makes sense, but it might trigger an infinite loop of fusion / splitting.

theano/gpuarray/elemwise.py
```diff
@@ -41,6 +41,28 @@ def get_scal(dt):
...
+    if isinstance(node_or_outputs, Apply):
+        outputs = node_or_outputs.outputs
+    else:
+        outputs = node_or_outputs
+    ptr_size = 8
+    # We compile code that would do indexing in int64
+    int_size = 8
+
+    # we take the limit from CUDA for now
+    argument_limit = 232
```
lamblin: If this is the limit, it means we cannot add 2 7D tensors in the general case, which is what happens in the failing case.

theano/gpuarray/opt.py
```diff
@@ -752,26 +753,36 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
         # cpu.
         gpu_output = res(*new_inputs)
         return [gpu_output]
+    elif op.scalar_op in (scalar.add, scalar.mul):
+        max_nb_inputs = max_inputs_to_GpuElemwise(outputs)
+        while len(inputs) > max_nb_inputs:
+            inputs = inputs[:-max_nb_inputs] + [res(*inputs[-max_nb_inputs:])]
```
lamblin: So, it gets stuck in an infinite loop here, because there is not enough memory for even 1 input...

theano/gpuarray/elemwise.py
lamblin: Why is it so low? The CUDA documentation says we have 4KB, are we really trying to support obsolete GPUs?
4KB would be enough to support > 100 dimensions, instead of choking at 7.

lamblin:
Tests have passed here, so I updated the original PR with the fixes.

PR #5576, opened 2017-02-18
Prevent pkl_utils from stomping higher sys recursion limits

mharradon:
As a stop gap for pickling large recursive structures - this should allow the user to set higher sys recursion limits outside of the library.

### DIFF
--- a/theano/misc/pkl_utils.py
+++ b/theano/misc/pkl_utils.py
@@ -40,7 +40,7 @@ __license__ = "3-clause BSD"
 
 min_recursion = 3000
 if sys.getrecursionlimit() < min_recursion:
-  sys.setrecursionlimit(min_recursion)
+    sys.setrecursionlimit(min_recursion)
 
 Pickler = pickle.Pickler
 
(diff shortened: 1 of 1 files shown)


theano/misc/pkl_utils.py
```diff
@@ -37,7 +37,11 @@
 __copyright__ = "Copyright 2013, Universite de Montreal"
 __license__ = "3-clause BSD"
 
-sys.setrecursionlimit(3000)
+
+min_recursion = 3000
+if sys.getrecursionlimit() < min_recursion:
+    sys.setrecursionlimit(min_recursion)
+
```
nouiz: Why do you want to change the default? This is dangerous.

mharradon: I just figure if the user wants to set a higher recursion limit on their system Theano shouldn't prevent them. Many systems can support much larger recursion limits, depending on the function. The current code already sets a recursion limit above the default, but it has the side effect of silently stomping higher recursion limits set by the user in calling code.

mharradon: Alternately we could rip the setrecursionlimit call out all together.

nouiz:
Right, it was already there. I agree that your PR make Theano better. We will do the RC before merging it. If we do a second RC, we will add it.

It is sad that we don't have a comment why it was added @lamblin git log tell that is is you that added it in 2013. Do you remember why? Should we keep it? To prevent Theano regression, I would keep it and merge this PR.

PR #5577, opened 2017-02-19
Use 64-bit indices in sparse.AddSD where necessary (fix #5525)

### LINKED ISSUE #5525: Segmentation Fault When training with a big dataset
I'm running a Custom MLP with Theano (using Lasagne) and I can run it over a small (input shape 10kx10k) and medium (input shape 400k_200k) but when I run it over a larger dataset (input shape 1.3Mk_100k) it gives me the segmentation fault error when it starts training. The memory starts to grow and then it crashes. I have 500GB of memory and when it gets to less than 20% of memory it throws the segmentation fault error. I'm using cpu and my .theanorc looks like this:

```
[global]
OMP_NUM_THREADS=4
openmp=True
floatX = float64
#profile_memory=True
#profile=True
device=cpu
[blas] 
ldflags=-L/usr/lib/  -lopenblas  -lblas 
```

Note that the input matrices are sparse and I can't do mini-batch and have to do **full batch**.

I used gdb to find where the error occurs and the output is:

```
Thread 17 "python" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff3d31700 (LWP 27105)]
(anonymous namespace)::__struct_compiled_op_1b42432866be1c904afdc4503a8345e1::run ()
    at /home/user1/.theano/compiledir_Linux-4.4--generic-x86_64-with-Ubuntu-16.04-xenial-x86_64-2.7.12-64/tmpB1iBsD/mod.cpp:518
518                 V1_i = V3_i + V5_i;

#0  (anonymous namespace)::__struct_compiled_op_5c38a1b25524822864d2313292bdfdc6::run ()
    at /home/arahimi/.theano/compiledir_Linux-4.4--generic-x86_64-with-Ubuntu-16.04-xenial-x86_64-2.7.12-64/tmpLIG_F8/mod.cpp:518
#1  0x00007fffce85d43e in ?? () from /usr/lib/x86_64-linux-gnu/libgomp.so.1
#2  0x00007ffff7bc16ba in start_thread (arg=0x7fffeed2f700) at pthread_create.c:333
#3  0x00007ffff78f782d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109

```
 
Theano, Scipy and Numpy versions:
```
t>>> theano.__version__
'0.9.0beta1.dev-RELEASE'
>>> scipy.__version__
'0.18.1'
>>> numpy.__version__
'1.12.0'
```

I tried with older versions and then I updated to the dev versions but the problem exists.

Theano config is attached.
Also the code that builds the model is attached.

I commented all other operations in the custom layer and left only S.dot(input, w) but the error still exists:
`activation = S.dot(input, self.W)`

where input is a csr sparse matrix 1386766 X 138640 and W is hidden layer parameters with size 138640 X 3000 where 3000 is hidden layer size.
**When I decrease the hidden layer size from 3000 to 1000 it works.**

Given that I still have enough memory, why I can't use higher hidden layer sizes?

[mlpconv_python.txt](https://github.com/Theano/Theano/files/765365/mlpconv_python.txt)

[theano_config.txt](https://github.com/Theano/Theano/files/765364/theano_config.txt)

rebecca-palmer:
Passes PYTHONPATH=. nosetests3 -v '/home/rnpalmer/Debian/theano/theano/sparse/tests' run in the source tree, but I don't have enough RAM to test the array sizes it's supposed to fix.

There is no easily-detectable speed penalty (tested on 2048x2048 with 5% and 100% nonzero), but that might be hardware-dependent.

Note that as the out-of-bounds operation is a write, this is potentially a security issue.

Speed test was done with:
import theano
import theano.sparse
import theano.tensor
import numpy as np
import scipy.sparse
import time
n=2048
p=0.05 #also tried p=1
m1=np.random.rand(n,n)
m2=scipy.sparse.csr_matrix(np.random.rand(n,n)*(np.random.rand(n,n)<p))
x=theano.shared(m1)
y=theano.sparse.csr_matrix()
f=theano.function([y],x,updates=[(x,x+y)])
t=time.time()
z1=f(m2)
print(time.time()-t)
t=time.time()
z1=f(m2)
print(time.time()-t)
print(np.max(np.abs(x.get_value()-m1-2*m2)))

### DIFF
--- a/theano/sparse/opt.py
+++ b/theano/sparse/opt.py
@@ -137,28 +137,29 @@ class AddSD_ccode(gof.op.Op):
                 }
 
                 npy_intp N =  PyArray_DIMS(%(_indptr)s)[0]-1;
-                const npy_int32 * __restrict__ indptr = (npy_int32 *)PyArray_DATA(%(_indptr)s);
-                const npy_int32 * __restrict__ indices = (npy_int32*)PyArray_DATA(%(_indices)s);
+
+                const dtype_%(_indptr)s* __restrict__ indptr = (dtype_%(_indptr)s*)PyArray_DATA(%(_indptr)s);
+                const dtype_%(_indices)s* __restrict__ indices = (dtype_%(_indices)s*)PyArray_DATA(%(_indices)s);
                 const dtype_%(_data)s* __restrict__ data = (dtype_%(_data)s*)PyArray_DATA(%(_data)s);
 
                 dtype_%(y)s* ydata = (dtype_%(y)s*)PyArray_DATA(%(y)s);
                 dtype_%(z)s* zdata = (dtype_%(z)s*)PyArray_DATA(%(z)s);
-                int Yi = PyArray_STRIDES(%(y)s)[0]/PyArray_DESCR(%(y)s)->elsize;
-                int Yj = PyArray_STRIDES(%(y)s)[1]/PyArray_DESCR(%(y)s)->elsize;
+                npy_intp Yi = PyArray_STRIDES(%(y)s)[0]/PyArray_DESCR(%(y)s)->elsize;
+                npy_intp Yj = PyArray_STRIDES(%(y)s)[1]/PyArray_DESCR(%(y)s)->elsize;
 
-                npy_int32 pos;
+                npy_intp pos;
                 if (%(format)s == 0){
-                for (npy_int32 col = 0; col < N; ++col){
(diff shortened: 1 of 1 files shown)


rebecca-palmer:
There's several more index overflow issues in other sparse ops, which my second commit fixes.  (As before, the sparse tests pass, but they're nowhere near the array size where these would matter.)

These ones probably aren't _security_ issues (reads, or wrong-index but in-bounds writes), but they can still produce crashes or nonsensical output.

PR #5579, opened 2017-02-20
Clean up

ReyhaneAskari:
Regarding #4672 , I implemented ```__call__``` method in the metaclass of ```metaObject``` as a singleton which centralizes the caching mechanism. I also resolved the conflicts as the branch was far behind the master. 
This PR is still in progress.

### DIFF
--- a/theano/gof/tests/test_utils.py
+++ b/theano/gof/tests/test_utils.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, print_function, division
 import theano
-from theano.gof.utils import CacheClass
 from theano.gof.utils import (
     give_variables_names, remove, unique)
 
@@ -62,17 +61,3 @@ def test_stack_trace():
         assert len(v.tag.trace[0]) == 2
     finally:
         theano.config.traceback.limit = orig
-
-
-class CachingClassExample(CacheClass, theano.Op):
-    __props__ = ('value',)
-    __cache_instance__ = True
-
-    def __init__(self, value):
-        self.value = value
-
-
-def test_caching():
-        obj_1 = CachingClassExample(3)
-        obj_2 = CachingClassExample(3)
-        assert obj_1 is obj_2
--- a/theano/gof/tests/test_utils.py
+++ b/theano/gof/tests/test_utils.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, print_function, division
 import theano
+from theano.gof.utils import CacheClass
 from theano.gof.utils import (
     give_variables_names, remove, unique)
 
@@ -61,3 +62,17 @@ def test_stack_trace():
         assert len(v.tag.trace[0]) == 2
     finally:
         theano.config.traceback.limit = orig
+
+
+class CachingClassExample(CacheClass, theano.Op):
+    __props__ = ('value',)
+    __cache_instance__ = True
+
+    def __init__(self, value):
+        self.value = value
+
+
+def test_caching():
+        obj_1 = CachingClassExample(3)
+        obj_2 = CachingClassExample(3)
+        assert obj_1 is obj_2
--- a/theano/gof/utils.py
+++ b/theano/gof/utils.py
@@ -9,7 +9,6 @@ from six.moves import StringIO
 
 from theano import config
 from theano.compat import PY3
-from theano.misc.frozendict import frozendict
 
 
 def simple_extract_stack(f=None, limit=None, skips=[]):
@@ -210,49 +209,6 @@ class MetaObject(type):
                 dct['__str__'] = __str__
 
         return type.__new__(cls, name, bases, dct)
-
-
-class MetaObject_caching(MetaObject):
-    _cache = {}
-
-    def __call__(cls, *args, **kwargs):
-        key = [cls]
-        for arg in args:
-            arg = make_hashable(arg)
-            key.append(arg)
-        for k, v in sorted(kwargs.items()):
-            key.append(k)
-            v = make_hashable(v)
-            key.append(v)
-        key = tuple(key)
-        if key not in cls._cache:
-            cls._cache[key] = super(MetaObject_caching, cls).__call__(*args, **kwargs)
-        return cls._cache[key]
-
-
-class CacheClass(with_metaclass(MetaObject_caching, object)):
-    pass
-
-
-def make_hashable(argument):
-    try:
-        hash(argument)
-        return argument
-    except:
-        if isinstance(argument, (list, tuple)):
-            argument = list(argument)
-            for index, instance in enumerate(argument):
-                argument[index] = make_hashable(instance)
-            argument = tuple(argument)
-        elif isinstance(argument, dict):
-            for k, v in argument.items():
-                argument[k] = make_hashable(v)
-            argument = frozendict(argument)
-        elif isinstance(argument, slice):
-            argument = argument.__reduce__()
-        elif isinstance(argument, np.ndarray):
-            argument = hash_from_ndarray(argument)
-        return argument
 
 
 class object2(with_metaclass(MetaObject, object)):
@@ -607,30 +563,6 @@ else:
         except TypeError:
             assert isinstance(msg, np.ndarray)
             return hashlib.md5(np.getbuffer(msg)).hexdigest()
-
-
-def hash_from_ndarray(data):
-                """
-                Return a hash from an ndarray.
-
-                It takes care of the data, shapes, strides and dtype.
-
-                """
-                # We need to hash the shapes and strides as hash_from_code only hashes
-                # the data buffer. Otherwise, this will cause problem with shapes like:
-                # (1, 0) and (2, 0) and problem with inplace transpose.
-                # We also need to add the dtype to make the distinction between
-                # uint32 and int32 of zeros with the same shape and strides.
-
-                # python hash are not strong, so I always use md5 in order not to have a
-                # too long hash, I call it again on the concatenation of all parts.
-                if not data.flags["C_CONTIGUOUS"]:
-                    # hash_from_code needs a C-contiguous array.
-                    data = np.ascontiguousarray(data)
-                return hash_from_code(hash_from_code(data) +
-                                      hash_from_code(str(data.shape)) +
-                                      hash_from_code(str(data.strides)) +
-                                      hash_from_code(str(data.dtype)))
 
 
 def hash_from_file(file_path):
--- a/theano/gof/utils.py
+++ b/theano/gof/utils.py
@@ -9,6 +9,7 @@ from six.moves import StringIO
 
 from theano import config
 from theano.compat import PY3
+from theano.misc.frozendict import frozendict
 
 
 def simple_extract_stack(f=None, limit=None, skips=[]):
@@ -209,6 +210,49 @@ class MetaObject(type):
                 dct['__str__'] = __str__
 
         return type.__new__(cls, name, bases, dct)
+
+
+class MetaObject_caching(MetaObject):
+    _cache = {}
+
+    def __call__(cls, *args, **kwargs):
+        key = [cls]
+        for arg in args:
+            arg = make_hashable(arg)
+            key.append(arg)
+        for k, v in sorted(kwargs.items()):
+            key.append(k)
+            v = make_hashable(v)
+            key.append(v)
+        key = tuple(key)
+        if key not in cls._cache:
+            cls._cache[key] = super(MetaObject_caching, cls).__call__(*args, **kwargs)
+        return cls._cache[key]
+
+
+class CacheClass(with_metaclass(MetaObject_caching, object)):
+    pass
+
+
+def make_hashable(argument):
+    try:
+        hash(argument)
+        return argument
+    except:
+        if isinstance(argument, (list, tuple)):
+            argument = list(argument)
+            for index, instance in enumerate(argument):
+                argument[index] = make_hashable(instance)
+            argument = tuple(argument)
+        elif isinstance(argument, dict):
+            for k, v in argument.items():
+                argument[k] = make_hashable(v)
+            argument = frozendict(argument)
+        elif isinstance(argument, slice):
+            argument = argument.__reduce__()
+        elif isinstance(argument, np.ndarray):
+            argument = hash_from_ndarray(argument)
+        return argument
 
 
 class object2(with_metaclass(MetaObject, object)):
@@ -563,6 +607,30 @@ else:
         except TypeError:
             assert isinstance(msg, np.ndarray)
             return hashlib.md5(np.getbuffer(msg)).hexdigest()
+
+
+def hash_from_ndarray(data):
+                """
+                Return a hash from an ndarray.
+
+                It takes care of the data, shapes, strides and dtype.
+
+                """
+                # We need to hash the shapes and strides as hash_from_code only hashes
+                # the data buffer. Otherwise, this will cause problem with shapes like:
+                # (1, 0) and (2, 0) and problem with inplace transpose.
+                # We also need to add the dtype to make the distinction between
+                # uint32 and int32 of zeros with the same shape and strides.
+
+                # python hash are not strong, so I always use md5 in order not to have a
+                # too long hash, I call it again on the concatenation of all parts.
+                if not data.flags["C_CONTIGUOUS"]:
+                    # hash_from_code needs a C-contiguous array.
+                    data = np.ascontiguousarray(data)
+                return hash_from_code(hash_from_code(data) +
+                                      hash_from_code(str(data.shape)) +
+                                      hash_from_code(str(data.strides)) +
+                                      hash_from_code(str(data.dtype)))
 
 
 def hash_from_file(file_path):
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -12,7 +12,7 @@ from theano.tensor.basic import (
     Alloc, AllocEmpty, alloc_validate_shape, Join, Split)
 
 from theano.gof import HideC, COp
-from theano.gof.utils import MethodNotDefined, CacheClass
+from theano.gof.utils import MethodNotDefined
 
 from collections import deque
 
@@ -629,7 +629,7 @@ class HostFromGpu(Op):
 host_from_gpu = HostFromGpu()
 
 
-class GpuFromHost(CacheClass, Op):
+class GpuFromHost(Op):
     """
     Transfer data to GPU.
 
@@ -783,7 +783,7 @@ class GpuToGpu(Op):
         return (1,)
 
 
-class GpuAlloc(CacheClass, HideC, Alloc):
+class GpuAlloc(HideC, Alloc):
     """
     Allocate initialized memory on the GPU.
 
@@ -945,7 +945,7 @@ class GpuAlloc(CacheClass, HideC, Alloc):
         return True
 
 
-class GpuAllocEmpty(CacheClass, HideC, AllocEmpty):
+class GpuAllocEmpty(HideC, AllocEmpty):
     """
     Allocate uninitialized memory on the GPU.
 
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -954,7 +954,7 @@ class GpuAllocEmpty(HideC, AllocEmpty):
     _f16_ok = True
     params_type = gpu_context_type
 
-    def __init__(self, dtype, context_name=None):
+    def __init__(self, dtype, context_name):
         self.dtype = dtype
         self.context_name = context_name
 
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -73,7 +73,7 @@ def as_gpuarray_variable(x, context_name):
 
         # If we couldn't deal with transfers, then maybe it's a tensor
         if isinstance(x.type, tensor.TensorType):
-            return gpu_from_host(context_name)(x)
+            return GpuFromHost(context_name)(x)
 
     # Try _as_GpuArrayVariable if possible
     if hasattr(x, '_as_GpuArrayVariable'):
@@ -617,7 +617,7 @@ class HostFromGpu(Op):
 
     def grad(self, inputs, grads):
         gz, = grads
-        return [gpu_from_host(inputs[0].type.context_name)(gz)]
+        return [GpuFromHost(inputs[0].type.context_name)(gz)]
 
     def R_op(self, inputs, eval_points):
         ev, = eval_points
@@ -720,14 +720,6 @@ class GpuFromHost(Op):
 
     def c_code_cache_version(self):
         return (9,)
-
-
-# Caching GPUAlloc
-def gpu_from_host(ctx):
-    if ctx not in gpu_alloc.cache:
-        gpu_from_host.cache[ctx] = GpuFromHost(ctx)
-    return gpu_from_host.cache[ctx]
-gpu_from_host.cache = {}
 
 
 class GpuToGpu(Op):
@@ -953,15 +945,6 @@ class GpuAlloc(HideC, Alloc):
         return True
 
 
-# Caching GPUAlloc
-def gpu_alloc(ctx, memset_0=False):
-    key = (ctx, memset_0)
-    if key not in gpu_alloc.cache:
-        gpu_alloc.cache[key] = GpuAlloc(ctx, memset_0)
-    return gpu_alloc.cache[key]
-gpu_alloc.cache = {}
-
-
 class GpuAllocEmpty(HideC, AllocEmpty):
     """
     Allocate uninitialized memory on the GPU.
@@ -971,7 +954,7 @@ class GpuAllocEmpty(HideC, AllocEmpty):
     _f16_ok = True
     params_type = gpu_context_type
 
-    def __init__(self, dtype, context_name):
+    def __init__(self, dtype, context_name=None):
         self.dtype = dtype
         self.context_name = context_name
 
@@ -1046,14 +1029,6 @@ if (theano_prep_output(&%(zz)s, %(ndim)s, shape, %(type)s, GA_C_ORDER,
 
 def empty_like(var):
     return GpuAllocEmpty(var.type.dtype, var.type.context_name)(*var.shape)
-
-
-def gpu_alloc_empty(ctx, dtype):
-    key = (dtype, ctx)
-    if key not in gpu_alloc_empty.cache:
-        gpu_alloc_empty.cache[key] = GpuAllocEmpty(dtype, ctx)
-    return gpu_alloc_empty.cache[key]
-gpu_alloc_empty.cache = {}
 
 
 class GpuContiguous(Op):
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -12,7 +12,7 @@ from theano.tensor.basic import (
     Alloc, AllocEmpty, alloc_validate_shape, Join, Split)
 
 from theano.gof import HideC, COp
-from theano.gof.utils import MethodNotDefined
+from theano.gof.utils import MethodNotDefined, CacheClass
 
 from collections import deque
 
@@ -629,7 +629,7 @@ class HostFromGpu(Op):
 host_from_gpu = HostFromGpu()
 
 
-class GpuFromHost(Op):
+class GpuFromHost(CacheClass, Op):
     """
     Transfer data to GPU.
 
@@ -783,7 +783,7 @@ class GpuToGpu(Op):
         return (1,)
 
 
-class GpuAlloc(HideC, Alloc):
+class GpuAlloc(CacheClass, HideC, Alloc):
     """
     Allocate initialized memory on the GPU.
 
@@ -945,7 +945,7 @@ class GpuAlloc(HideC, Alloc):
         return True
 
 
-class GpuAllocEmpty(HideC, AllocEmpty):
+class GpuAllocEmpty(CacheClass, HideC, AllocEmpty):
     """
     Allocate uninitialized memory on the GPU.
 
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -663,8 +663,8 @@ class GpuFromHost(Op):
 
     def grad(self, inputs, grads):
         gz, = grads
-        return [host_from_gpu(as_gpuarray_variable(
-                gz, context_name=self.context_name))]
+        return [as_gpuarray_variable(
+                gz, context_name=self.context_name).transfer('cpu')]
 
     def R_op(self, inputs, eval_points):
         ev, = eval_points
@@ -1132,7 +1132,7 @@ class GpuReshape(HideC, tensor.Reshape):
         ctx_name = infer_context_name(x)
         x = as_gpuarray_variable(x, context_name=ctx_name)
         shp = tensor.as_tensor_variable(shp)
-        res = host_from_gpu(x).reshape(shp, ndim=self.ndim)
+        res = x.transfer('cpu').reshape(shp, ndim=self.ndim)
         otype = GpuArrayType(dtype=res.dtype,
                              broadcastable=res.broadcastable,
                              context_name=ctx_name)
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1138,7 +1138,7 @@ def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
 
     desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample,
                           conv_mode=conv_mode, precision=precision)(kerns_shp)
-    out = GpuAllocEmpty(ctx_name, dtype=img.dtype)(*kerns_shp)
+    out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*kerns_shp)
     return GpuDnnConvGradW()(img, topgrad, out, desc)
 
 
@@ -1166,7 +1166,7 @@ def dnn_gradinput(kerns, topgrad, img_shp, border_mode='valid',
 
     desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample,
                           conv_mode=conv_mode, precision=precision)(kerns.shape)
-    out = GpuAllocEmpty(ctx_name, kerns.dtype)(*img_shp)
+    out = GpuAllocEmpty(dtype=kerns.dtype, context_name=ctx_name)(*img_shp)
     return GpuDnnConvGradI()(kerns, topgrad, out, desc)
 
 
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -13,6 +13,7 @@ from theano.scalar import as_scalar, constant, Log, get_scalar_type
 from theano.tensor import as_tensor_variable
 from theano.gradient import DisconnectedType, grad_not_implemented
 from theano.gof import Optimizer, local_optimizer, COp
+from theano.gof.utils import CacheClass
 from theano.gof.cmodule import GCC_compiler
 from theano.gof.type import CDataType, Generic
 from theano.compile import optdb
@@ -348,7 +349,7 @@ def version(raises=True):
 version.v = None
 
 
-class GpuDnnConvDesc(COp):
+class GpuDnnConvDesc(CacheClass, COp):
 
     """
     This Op builds a convolution descriptor for use in the other convolution
@@ -639,7 +640,7 @@ class GpuDnnConv(DnnBase):
         return [shape[2]]
 
 
-class GpuDnnConvGradW(DnnBase):
+class GpuDnnConvGradW(CacheClass, DnnBase):
 
     """
     The convolution gradient with respect to the weights.
@@ -770,7 +771,7 @@ class GpuDnnConvGradW(DnnBase):
         return [shape[2]]
 
 
-class GpuDnnConvGradI(DnnBase):
+class GpuDnnConvGradI(CacheClass, DnnBase):
     """
     The convolution gradient with respect to the inputs.
 
@@ -1137,9 +1138,8 @@ def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
     precision = get_precision(precision, [img, topgrad])
 
     desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample,
-                             conv_mode=conv_mode, precision=precision)(
-                                 kerns_shp)
-    out = GpuAllocEmpty(ctx_name, dtype=img.dtype)(*kerns_shp)
+                          conv_mode=conv_mode, precision=precision)(kerns_shp)
+    out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*kerns_shp)
     return GpuDnnConvGradW()(img, topgrad, out, desc)
 
 
@@ -1150,6 +1150,7 @@ def dnn_gradweight3d(img, topgrad, kerns_shp, border_mode='valid',
     """
     return dnn_gradweight(img, topgrad, kerns_shp, border_mode,
                           subsample, conv_mode, precision)
+
 
 def dnn_gradinput(kerns, topgrad, img_shp, border_mode='valid',
                   subsample=(1, 1), conv_mode='conv', precision=None):
@@ -1165,9 +1166,8 @@ def dnn_gradinput(kerns, topgrad, img_shp, border_mode='valid',
     precision = get_precision(precision, [kerns, topgrad])
 
     desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample,
-                             conv_mode=conv_mode, precision=precision)(
-                                 kerns.shape)
-    out = GpuAllocEmpty(ctx_name, kerns.dtype)(*img_shp)
+                          conv_mode=conv_mode, precision=precision)(kerns.shape)
+    out = GpuAllocEmpty(dtype=kerns.dtype, context_name=ctx_name)(*img_shp)
     return GpuDnnConvGradI()(kerns, topgrad, out, desc)
 
 
@@ -1178,6 +1178,7 @@ def dnn_gradinput3d(kerns, topgrad, img_shp, border_mode='valid',
     """
     return dnn_gradinput(kerns, topgrad, img_shp, border_mode, subsample,
                          conv_mode, precision)
+
 
 class GpuDnnPoolDesc(Op):
 
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -32,7 +32,7 @@ from . import pygpu
 from .type import (get_context, gpu_context_type, list_contexts,
                    GpuArraySharedVariable)
 from .basic_ops import (as_gpuarray_variable, infer_context_name,
-                        gpu_contiguous, gpu_alloc_empty,
+                        gpu_contiguous, GpuAllocEmpty,
                         empty_like, GpuArrayType, HostFromGpu)
 from .elemwise import GpuElemwise
 
@@ -466,18 +466,6 @@ class GpuDnnConvDesc(COp):
         return (super(GpuDnnConvDesc, self).c_code_cache_version(), version())
 
 
-def gpu_dnn_conv_desc(border_mode, subsample=(1, 1), conv_mode='conv',
-                      precision="float32"):
-    key = (border_mode, subsample, conv_mode, precision)
-    if key not in gpu_dnn_conv_desc.cache:
-        gpu_dnn_conv_desc.cache[key] = GpuDnnConvDesc(border_mode,
-                                                      subsample,
-                                                      conv_mode,
-                                                      precision)
-    return gpu_dnn_conv_desc.cache[key]
-gpu_dnn_conv_desc.cache = {}
-
-
 # scalar constants
 _zero = constant(np.asarray(0.0, dtype='float64'))
 _one = constant(np.asarray(1.0, dtype='float64'))
@@ -613,8 +601,8 @@ class GpuDnnConv(DnnBase):
 
         top = gpu_contiguous(top)
 
-        d_img = gpu_dnn_conv_gradI()(kerns, top, empty_like(img), desc)
-        d_kerns = gpu_dnn_conv_gradW()(img, top, empty_like(kerns), desc)
+        d_img = GpuDnnConvGradI()(kerns, top, empty_like(img), desc)
+        d_kerns = GpuDnnConvGradW()(img, top, empty_like(kerns), desc)
         d_alpha = grad_not_implemented(self, 4, alpha)
         d_beta = grad_not_implemented(self, 5, beta)
 
@@ -649,14 +637,6 @@ class GpuDnnConv(DnnBase):
 
     def infer_shape(self, node, shape):
         return [shape[2]]
-
-
-def gpu_dnn_conv(algo=None, inplace=False):
-    key = (algo, inplace)
-    if key not in gpu_dnn_conv.cache:
-        gpu_dnn_conv.cache[key] = GpuDnnConv(algo, inplace)
-    return gpu_dnn_conv.cache[key]
-gpu_dnn_conv.cache = {}
 
 
 class GpuDnnConvGradW(DnnBase):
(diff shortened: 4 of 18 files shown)


theano/gof/utils.py
```diff
@@ -210,6 +213,15 @@ def __str__(self):
 
         return type.__new__(cls, name, bases, dct)
 
+    _cache = {}
+
+    def __call__(cls, *args, **kwargs):
+        if not cls._has_props:
```
Sentient07: I earlier opened a PR, where I made all the `OP` instance have `__props__`. In that, I had left out `ScalarOp` and @abergeron explains here why,  https://github.com/Theano/Theano/pull/4749#issuecomment-252945665 . Because of that, I couldn't proceed with the PR that addresses a centralised caching approach. Also, from what I remember this would fail with the old backend as they don't have `__props__`.

theano/gpuarray/opt.py
```diff
@@ -200,6 +204,8 @@ def op_lifter(OP, cuda_only=False):
     """
 
     def f(maker):
+        @register_topo('fast_compile')
+        @local_optimizer(OP)
```
nouiz: Why this here? I don't think this is good. Sometimes we will want to pass extra parameter to local_optimizer and register_topo.

theano/gof/utils.py
nouiz: @Sentient07 this "if" will make this method have as the default when there is no \__props__. So it support class that don't have problems by not caching them. This is what we want.

theano/gpuarray/basic_ops.py
```diff
@@ -968,7 +960,7 @@ class GpuAllocEmpty(HideC, AllocEmpty):
     _f16_ok = True
     params_type = gpu_context_type
 
-    def __init__(self, dtype, context_name):
+    def __init__(self, dtype, context_name=None):
```
nouiz: @abergeron I'm not sure of this change. What do you think? For safety, it is better to not include it, but if it is good, we can get it in.

theano/gpuarray/dnn.py
```diff
@@ -987,10 +951,10 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1),
                    shape_i(img, 2, fgraph) - shape_i(kerns, 2, fgraph) + 1,
                    shape_i(img, 3, fgraph) - shape_i(kerns, 3, fgraph) + 1)
         out_shp = assert_conv_shape(out_shp)
-        out = gpu_alloc_empty(ctx_name, dtype=img.dtype)(*out_shp)
+        out = GpuAllocEmpty(ctx_name, dtype=img.dtype)(*out_shp)
```
nouiz: swap the 2 param. Make sure this is also done for all instance of GpuAllocEmpty introduced. gpu_alloc_empty and GpuAllocEmpty don't take the parameter in the same order.

Or better, pass both param by name. This will always work.

theano/gpuarray/elemwise.py
```diff
@@ -41,48 +41,6 @@ def get_scal(dt):
     return scalar.get_scalar_type(dt)
 
 
-def max_inputs_to_GpuElemwise(node_or_outputs):
```
nouiz: All the change to this PR is probably due to a bad rebase/merge. Make sure they don't appear in this PR.

theano/gpuarray/nerv.py
```diff
@@ -157,8 +157,8 @@ def local_gpua_dot_to_gemm16(op, ctx_name, inputs, outputs):
     B = inputs[1]
     if (A.ndim == 2 and B.ndim == 2 and
             A.dtype == 'float16' and B.dtype == 'float16'):
-        fgraph = getattr(outputs[0], 'fgraph', None)
-        C = gpu_alloc_empty(ctx_name, dtype='float16')(
+        fgraph = inputs[0].fgraph
```
nouiz: remove the change to this line. I don't think it is good. inputs don't always have an fgraph attribute I think.

@Sentient07 do you remember why this line was changed? If not, we should undo it.

theano/gpuarray/opt.py
```diff
@@ -106,6 +105,11 @@ def query(self, *tags, **kwtags):
 gpu_seqopt.register('gpuarray_cut_transfers', gpu_cut_copies, 2,
                     'fast_compile', 'fast_run', 'gpuarray')
 
+gpu_seqopt.register('op_lifter_topo', TopoOptimizer(gpu_topo.query('+fast_compile', '+fast_run', '+cudnn',
+                                                                   '+cudnn', '+conv_dnn', '+gpuarray'),
+                                                    order='out_to_in'),
+                    0.5, 'fast_run', 'gpuarray', 'fast_compile')
+
```
nouiz: remove this and the line 80. It isn't used for now as nothing is registered in gpu_topo. We don't have benchmark that show this would be useful, so I would not finish this until we see a need for it.

theano/gpuarray/opt.py
```diff
@@ -165,25 +169,36 @@ def f(local_opt):
...
 
 
+def register_topo(*tags, **kwargs):
+    def f(local_opt):
+        name = (kwargs and kwargs.pop('name')) or local_opt.__name__
+        name = name + "_graph"
+        gpu_topo.register(name, local_opt,
+                          10, 'fast_run', 'gpuarray', *tags)
+        return local_opt
+    return f
```
nouiz: This should also be removed or moved to another PR. @Sentient07 do you remember if you saw a speed up from this change? If so, on which benchmark?

theano/gpuarray/opt.py
```diff
@@ -235,10 +250,10 @@ def local_opt(node):
                     elif isinstance(new_op, (tuple, list)):
                         return [safe_to_cpu(o) for o in new_op]
                     else:  # suppose it is a variable on the GPU
-                        return [host_from_gpu(new_op)]
+                        return [new_op.transfer('cpu')]
             return False
         local_opt.__name__ = maker.__name__
-        return local_optimizer(OP)(local_opt)
+        return register_topo('fast_compile')(local_optimizer(OP)(local_opt))
```
nouiz: also to be removed/moved to another PR. This isn't a clean up, but a new feature/optimization.

theano/gpuarray/opt.py
```diff
@@ -429,7 +444,7 @@ def apply(self, fgraph):
                         new_o.owner.inputs[0].type == o.type):
                     new_o = new_o.owner.inputs[0]
                 else:
-                    new_o = safe_to_cpu(new_o)
+                    new_o = new_o.transfer('cpu')
```
nouiz: revert. safe_to_cpu() does special handling.

theano/gpuarray/opt.py
```diff
@@ -753,37 +768,26 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
...
-            inner_op = []
-            for i in range(0, len(node.inputs), max_nb_inputs):
-                inner_op.append(node.op(*node.inputs[i: i + max_nb_inputs]))
-            node = node.op(*inner_op).owner
-    return node
+    nb_bytes_avail = argument_limit - size_param_mandatory
+    nb_bytes_per_input = ptr_size + ndim * int_size
+    max_nb_inputs = nb_bytes_avail // nb_bytes_per_input
+
+    return max_nb_inputs
```
nouiz: This method max_inputs_to_GpuElemwis is the old version. There seem to be a bad rebase here.

theano/gpuarray/opt.py
```diff
@@ -1187,6 +1191,27 @@ def local_gpua_gemmbatch(op, context_name, inputs, outputs):
...
+    B = inputs[1]
+    if (A.ndim == 2 and B.ndim == 2 and
+            A.dtype == 'float16' and B.dtype == 'float16'):
+        fgraph = outputs[0].fgraph
+        C = GpuAllocEmpty('float16', context_name)(
+            shape_i(A, 0, fgraph),
+            shape_i(B, 1, fgraph))
+        return gpugemm_no_inplace(C, 1.0, A, B, 0.0)
+
+
```
nouiz: @Sentient07 why this was needed? I think it isn't needed anymore.

theano/gpuarray/dnn.py
ReyhaneAskari: added their name and also fixed the ordering. thanks.

theano/gpuarray/opt.py
ReyhaneAskari: done.

theano/gpuarray/opt.py
ReyhaneAskari: I removed this line.

theano/gpuarray/opt.py
ReyhaneAskari: I removed them.

theano/gpuarray/basic_ops.py
ReyhaneAskari: I removed this change. We can add it later if needed.

theano/gpuarray/elemwise.py
ReyhaneAskari: Yes, I fixed it. Thanks.

theano/gpuarray/nerv.py
ReyhaneAskari: I reverted this change for now. Let me know if we needed a revert.

theano/gpuarray/opt.py
ReyhaneAskari: I also removed this one.

theano/gpuarray/opt.py
ReyhaneAskari: It is removed now. Thanks.

theano/gpuarray/opt.py
ReyhaneAskari: fixed. thanks.

ReyhaneAskari:
@nouiz , I made the requested changes to first clear up the  #4672. I made a new branch and cherry picked the commits that we wanted. I have not yet added my fix. After the tests were passed I will add that part.

theano/gpuarray/opt.py
ReyhaneAskari: I removed them. Thanks Fred.

theano/gpuarray/opt.py
```diff
@@ -615,7 +613,7 @@ def local_gpua_alloc(op, context_name, inputs, outputs):
 def local_gpua_alloc_empty(op, context_name, inputs, outputs):
     # We use _props_dict() to make sure that the GPU op know all the
     # CPU op props.
-    return gpu_alloc_empty(context_name, **op._props_dict())
+    return GpuAllocEmpty(**op._props_dict())(*inputs)
```
nouiz: I think it should be:

GpuAllocEmpty(context_name=ontext_name, **op._props_dict())(*inputs)

You can test it locally with: 

nosetests theano/scan_module/tests/test_scan.py:T_Scan_Gpuarray.test_one_sequence_one_output_weights_gpu1

this is one of the jenkins failing test.

ReyhaneAskari: I see. Thanks.

theano/gpuarray/opt.py
Sentient07: This was done to lift the left over nodes to the CPU in the backward pass. The main motivation for starting a fresh clean up PR was implementing a backward pass. I will post more details on this in the following comment

Sentient07: During July, @lamblin proposed this, "I'm proposing to have a second pass in the other direction: from outputs towards the inputs, where we could lift the remaining Ops by their output.
Actually, both passes could also become part of the GraphToGPUOptimizer. This remain to be decided. "
Now it makes sense for me why I registered a topo optimizer. It doesn't give any change (or shouldn't) in profiling immediately. But when we implement backward pass, there should be. But earlier we stopped from implementing a backward pass, not knowing the exact details of number of nodes that weren't moved to the GPU in the primary pass and whether it is worth implementing one.

Sentient07: I remember `Alloc` was a special case and the nodes containing this OP were not move to the GPU. That is why I had done a special handling of GPUAlloc.

theano/gpuarray/opt.py
Sentient07: I agree this is aimed at introducing a new feature, the backward pass. Please let me know if the plan of having one is still on.

theano/gpuarray/opt.py
Sentient07: This is a special case of handling GPUAlloc and GPUAlloc empty. They were not moved to the GPU in the first pass by the GraphToGPU and hence these changes(hack) were made till the time we implement a backward pass that lifts these as well. These made some tests fail (Elemwise, I think) at that time.

Sentient07: The to-do list that @lamblin gave me when you (@nouiz) were on vacation, 

> " Remove the special case for Alloc / AllocEmpty from the new optimizer, and register these lifters in an "out2in" topooptimizer that will act as a second phase. Indeed, it makes sense to lift these operations from their output (GpuFromHost(op(x)) => gpu_op(GpuFromHost(x))."

ReyhaneAskari:
@Sentient07 , thanks for your comments. 
@nouiz , please let me know if I need to change anything referred to @Sentient07 's comments. I added the centralized caching mechanism. In my local tests, I saw many cases where the key : 
```
key = (cls, args, tuple(kwargs.values()))
````
was not valid as the ```args``` or the ```tuple(kwargs.values())``` contained dictionaries or ```numpy.ndarray```. I think converting them to hashable types is not wise nor easy. So, in this commit, I implemented the case where the ```key``` is a valid key (hashable). There are still cases where ```args``` or ```kwargs``` or both are empty and we are caching them. Please let me know if this is what you are looking for.

File: theano/gof/utils.py [hunk 14]
```diff
@@ -210,6 +210,19 @@ def __str__(self):
 
         return type.__new__(cls, name, bases, dct)
 
+    _cache = {}
+
+    def __call__(cls, *args, **kwargs):
+        props = cls.__dict__.get('__props__', None)
+        if props is None:
+            return super(MetaObject, cls).__call__(*args, **kwargs)
+        key = (cls, args, tuple(kwargs.values()))
```
nouiz: For the problem with dict as parameter, find where we pass a dict and modify that code to pass a frozendict instead.

Maybe around tensor/opt.py line 349, you need something like:
inplace_patern = theano.misc.fronzendict.frozendict(inplace_pattern)

nouiz: For the ndarray not hashable problem, can you convert them to python int/float at a few places before that method get called?

theano/gpuarray/opt.py
nouiz: Before merging it in Theano, I would prefer to see a profile that show this make a speed difference. Adding code for theoretical case isn't super useful.

I think we can let this where it is and take it again if we see a profile that show this is useful.

Sentient07: The profiling will show a difference only if we implement a backward pass. I am not sure how to estimate how efficient the backward pass will be. Since this is a part of implementing backward pass, I agree this shouldn't be in this PR

slefrancois:
jenkins got stuck on the gpuarray tests:
http://darjeeling.iro.umontreal.ca:8080/job/Theano_gpu/853/console

You can put the tests in verbose [here](https://github.com/Theano/Theano/blob/master/.jenkins/jenkins_test2.sh#L80) if it gets stuck again.

theano/gpuarray/opt.py
nouiz: Currently, as we don't have a backward pass, if we need it, we will see a big time inside the gpu_optimizer Equilibrium that will do an inefficient backward pass. The backward pass is just to speed up the compilation as it is currently done inefficiently. I didn't see a profile that show this is taking significant time.

Sentient07: I'm not sure I got you. Do you mean in the current setting, Equilibrium optimizer will perform an inefficient backward pass ? That was why @lamblin wanted the backward pass to be registered as an out to in Topological Optimizer, that will go from output to the input, to lift off these operations, 
`GpuFromHost(op(x)) => gpu_op(GpuFromHost(x))`

Sentient07: Also, the resnet example's code builds a graph which doesn't have a node(anything but the output node) whose output isn't on the GPU, as far as I remember. Hence nodes after that in the graph does not get blocked, moving them to the GPU would have shown significance profiling improvement. This was the principal cause for building a backward pass.

Sentient07: One more thing that slipped out of my mind. We wanted to be aggressive(not while merging the PR, but a little later) with the moving of OPs to the GPU, do you remember @nouiz ? It was by setting the `move_to_gpu` to True by default, instead of checking if the inputs to the nodes are of GpuArrayType.

ReyhaneAskari:
@nouiz, I was able to reproduce the error which is generated by running these two tests in the same run:
```
theano/gof/tests/test_fg.py:TFunctionGraph.test_node_outputs_not_used
theano/scan_module/tests/test_scan_opt.py:TestGaussNewton.test_nobatch 
```
It was really weird as running only the second test (the one giving error in travis) does not give any errors.  The combination of both of them gives the error as the first test is using a pickled test and creates an object which is then mistakenly used by the second test. They both use an ```Elemwise``` class with the same arguments. But the elemwise gives outputs with different ```dtype```s, thus we had the error. I added another part for the key of caching where we use the ```dtype``` of ```args``` ```kwargs```and their output as key.

ReyhaneAskari:
My guess is that the tests are being killed due to not enough memory. I think we are caching too many objects or some tests have objects that are too large.

theano/gof/utils.py
```diff
@@ -217,14 +217,21 @@ def __call__(cls, *args, **kwargs):
...
+            arg_dtype = get_arg_dtype(arg)
+            if arg_dtype:
+                key.extend(arg_dtype)
         for k, v in kwargs.items():
             key.append(k)
             v = make_hashable(v)
             key.append(v)
+            v_dtype = get_arg_dtype(v)
+            if v_dtype:
+                key.extend(v_dtype)
```
Sentient07: I'm not sure if it is a wise idea to have `dtype` as a class's key for hashing. But @nouiz told me to remove `name` from the `Elemwise` as it is not supposed to be cached. That was the reason why I started this [PR](https://github.com/Theano/Theano/pull/5071) as a clean up to the `frozendict` PR.

theano/gpuarray/opt.py
nouiz: For now, I think we should not change the logic of move_to_gpu. Maybe it could be a Theano flags, but I would not change the default for now. There is cases with subtensor where we want to keep them on the CPU.

Can you point me to the example that would show a speed up from a fast back-ward pass? If I see a profile that show it is useful, we can spin that to another PR.

theano/gof/utils.py
```diff
@@ -216,13 +217,37 @@ def __call__(cls, *args, **kwargs):
...
+    except:
+        if isinstance(argument, list) or isinstance(argument, tuple):
+            argument = list(argument)
+            for index, instance in enumerate(argument):
+                argument[index] = make_hashable(instance)
+            argument = tuple(argument)
+        elif isinstance(argument, dict):
+            argument = frozendict(argument)
+            for k, v in argument.iteritems():
+                v = make_hashable(v)
```
nouiz: This won't change the values inside the frozendict. Do like the list/tuple. Make a copy of the dict, then modifies its values, the after make it a frozendict.

Sentient07:
Yes, I remember some subtensor tests failed. Now, over the current master,
I ran the tests with the aggressive mode and it passed.
Also, I'll write an example for which backward pass is necessary

theano/gof/utils.py
```diff
@@ -221,7 +221,7 @@ def __call__(cls, *args, **kwargs):
         for arg in args:
             arg = make_hashable(arg)
             key.append(arg)
-        for k, v in kwargs.iteritems():
+        for k, v in kwargs.items():
```
nouiz: Why? I think it is not good for python2/3 support. Check what is done elsewhere to be sure.

theano/gof/utils.py
ReyhaneAskari: Yes, I thought it would generate lots of errors! So I decided to first try this and debug this part and then change the values after the tests were passed.

theano/gof/utils.py
ReyhaneAskari: We had an error here in python 3. So I changed it to ```items()``` which is supported in both.

theano/gof/utils.py
nouiz: @ReyhaneAskari Why this? This remove the value from the key. The value is important.
@Sentient07 your comment with the one from @ReyhaneAskari about caching too much stuff make me think of something. If one of the kwargs key isn't a prop, we could skip the cache. This with the elemwise name parameter that isn't a props, would allow the python gc to do much of its work.

@ReyhaneAskari can you try that? If that isn't enough, we could try to just skip the cache for all elemwise.

theano/gof/utils.py
```diff
@@ -210,6 +211,66 @@ def __str__(self):
...
+            return super(MetaObject, cls).__call__(*args, **kwargs)
+
+        key = [cls]
+        for arg in args:
+            arg = make_hashable(arg)
+            key.append(arg)
+            arg_dtype = get_arg_dtype(arg)
+            if arg_dtype:
+                key.extend(arg_dtype)
+        for k, v in kwargs.items():
```
abergeron: You probably want to sort this so it stays stable.

abergeron:
It's weird that both on travis and on jenkins we ate a SIGKILL during the tests.

There may be a memory problem which causes the OOM to trigger.  It might be worth it to implement some sort of limit on the cache.

theano/gof/utils.py
```diff
@@ -243,10 +243,12 @@ def make_hashable(argument):
             argument = tuple(argument)
         elif isinstance(argument, dict):
             argument = frozendict(argument)
-            for k, v in argument.iteritems():
+            for k, v in argument.items():
                 v = make_hashable(v)
         elif isinstance(argument, slice):
             argument = argument.__reduce__()
+        elif isinstance(argument, np.ndarray):
+            argument = np.asscalar(argument)
```
ReyhaneAskari: Thanks. I did.

theano/gof/utils.py
ReyhaneAskari: I did not fully understand what you mean here.  I still have the value in the key for caching. The reason that I added the ```dtype``` was explained in my yesterday's comment. We had an error where an object gave outputs with different ```dtypes``` in two cases.``` dtype``` was added to make sure the same object is not used in these two cases. 

Yes I will test both cases.

theano/gof/utils.py
ReyhaneAskari: You are right. Thanks.

theano/gof/utils.py
Sentient07: >  If one of the kwargs key isn't a prop, we could skip the cache.

This was exactly my point for implementation. But from what I understand, @ReyhaneAskari in some cases, is having an Elemwise with the elements of props being the same. I have an idea, I will comment at the appropriate place.

ReyhaneAskari:
@nouiz , @abergeron  I made the requested changes.  These are the rules that I followed for caching:

- when the props is None, we never cache
- when the props is (), we always cache
- when the props is defined but not empty, the args and kwargs should only contain the porps

Travis has some problems today, but I guess the reason that Jenkins is failing is due to not enough memory. We can also remove caching when props is empty. Do you have other suggestions?

theano/gof/tests/test_utils.py
```diff
@@ -61,3 +61,44 @@ def test_stack_trace():
...
+        obj_2 = self.CachingClassExample(3)
+        assert obj_1 is obj_2
+
+        obj_1 = self.CachingClassExample_2(3)
+        obj_2 = self.CachingClassExample_2(3)
+        assert obj_1 is not obj_2
+
+        obj_1 = self.CachingClassExample_3(3)
+        obj_2 = self.CachingClassExample_3(3)
+        assert obj_1 is obj_2
```
nouiz: I think this should not be cached. I think this is the cause that the memory error still continue.

theano/gof/utils.py
```diff
@@ -210,6 +212,66 @@ def __str__(self):
...
+        #       gives an error since cls.__init__ is not a python function.
+
+        props = cls.__dict__.get('__props__', None)
+
+        # The cases are being separated for code readability.
+        to_cache = False
+        if props is None:
+            pass
+        elif props == ():
+            to_cache = True
```
nouiz: I think this `elif props == ()` should be removed. The next one should cover that cases. If there is no props, there must be no parameter to the \__init__ to enable the cache. If there is parameter to the init and there is an empty props, then we should not cache.

ReyhaneAskari: The thing is that ```inspect.getargspec(cls.__init__)```  is not always defined. It is only valid when ```__init__``` was defined. Otherwise it's a slot wrapper.  I changed the caching policy as you suggested. We will see if that is suffices the memory issue.

ReyhaneAskari:
I ran a banchmarking on sb_resnet to see how much the caching was actually used. So here it is in short:

number of items not used:3919
number of items used once:3
number of items used twice:3
Size of Cache: 196888 bytes

These are the objects that were used more than 3 times and the number of times they were used: 

[<theano.gpuarray.elemwise.GpuCAReduceCuda object at 0x7f3ad66b6dd0>, 9]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3ad9b0ca50>, 157]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3adb545190>, **769**]
[<theano.tensor.elemwise.Sum object at 0x7f3ad9b71b50>, 55]
[<theano.gpuarray.basic_ops.GpuAllocEmpty object at 0x7f3ad627f190>, 153]
[<theano.tensor.elemwise.Sum object at 0x7f3ad9db7350>, 8]
[<theano.tensor.opt.Assert object at 0x7f3ad9a0a5d0>, 85]
[<theano.tensor.elemwise.Sum object at 0x7f3adac6f650>, 5]
[<theano.gpuarray.dnn.GpuDnnConvDesc object at 0x7f3ad6236c90>, 37]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3ad9b91fd0>, 31]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3adb60b090>, **478**]
[<theano.gpuarray.dnn.GpuDnnConvDesc object at 0x7f3ad5d1ecd0>, 24]
[<theano.gpuarray.dnn.GpuDnnConvDesc object at 0x7f3ad179b210>, 20]
[<theano.gpuarray.dnn.GpuDnnConvDesc object at 0x7f3ad16e2a10>, 17]
[<theano.gpuarray.elemwise.GpuCAReduceCuda object at 0x7f3ad661a050>, 5]
[<theano.tensor.opt.Assert object at 0x7f3ad17653d0>, 74]
[<theano.gpuarray.basic_ops.GpuFromHost object at 0x7f3adb33ed90>, **2833**]
[<theano.tensor.elemwise.Sum object at 0x7f3ad9cd6490>, 101]
[<theano.tensor.opt.Assert object at 0x7f3ad17ccf10>, 74]
[<theano.tensor.elemwise.Sum object at 0x7f3adb4a7b50>, 270]
[<theano.tensor.opt.Assert object at 0x7f3adb4b6c90>, 145]
[<theano.compile.ops.Shape_i object at 0x7f3adc37eed0>, 834]
[<theano.gpuarray.elemwise.GpuCAReduceCuda object at 0x7f3ad65ab110>, 191]
[<theano.gpuarray.dnn.GpuDnnConv object at 0x7f3ad1760790>, 74]
[<theano.gpuarray.dnn.GpuDnnConvDesc object at 0x7f3ad17d3390>, 35]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3adb612310>, 517]
[<theano.tensor.opt.Assert object at 0x7f3ad99f5a90>, 78]
[<theano.tensor.opt.MakeVector object at 0x7f3adb539a10>, 122]
[<theano.compile.ops.Shape_i object at 0x7f3adc4be3d0>, **511**]
[Scalar(bool), 60]
[<theano.gof.type.CDataType object at 0x7f3ad627f5d0>, 153]
[<theano.tensor.elemwise.Sum object at 0x7f3ada2c2950>, 9]
[<theano.gpuarray.elemwise.GpuCAReduceCuda object at 0x7f3ad36bf290>, 8]
[<theano.gpuarray.dnn.GpuDnnConvDesc object at 0x7f3ad1c60d50>, 15]
[Scalar(float32), 55]
[<theano.tensor.elemwise.Sum object at 0x7f3ad9ead750>, 237]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3ad9dc1a50>, 8]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3adb5fdcd0>, **1075**]
[<theano.tensor.opt.Assert object at 0x7f3adb66da50>, 145]
[<theano.tensor.opt.Assert object at 0x7f3ad1765a90>, 74]
[<theano.compile.ops.Shape_i object at 0x7f3adc4be510>, 251]
[<theano.gpuarray.dnn.GpuDnnConvGradW object at 0x7f3ad4b8b110>, 42]
[Scalar(float32), 7414]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3ad9bde950>, 83]
[<theano.gpuarray.dnn.GpuDnnConvGradI object at 0x7f3ad4cb8d10>, 35]
[<theano.gpuarray.basic_ops.GpuAlloc object at 0x7f3ad658f790>, 13]
[<theano.compile.ops.Shape_i object at 0x7f3adc4be650>, 251]
[<theano.tensor.elemwise.DimShuffle object at 0x7f3ad9a34690>, 20]
[<theano.tensor.opt.Assert object at 0x7f3ad1765fd0>, 74]
[<theano.tensor.elemwise.Sum object at 0x7f3ad9b71390>, **196**]

ReyhaneAskari:
Here is a timing benchmark for caching. Unfortunately, there no improvement. The restnet has 29 layers.
[out_with_caching_29.txt](https://github.com/Theano/Theano/files/815151/out_29.txt)
[out_master_29.txt](https://github.com/Theano/Theano/files/815152/out_master_29.txt)

In short:

**Function1 profiling:**
  Total compile time: 2.414667e+02s
    Theano Optimizer time: 7.742537e+01s
       Theano validate time: 3.758423e-01s
    Theano Linker time (includes C, CUDA code generation/compiling): 1.505997e+02s
       Import time 4.270077e-02s
       Node make_thunk time 8.662519e+01s

**Function1 profiling-with caching:**
 Total compile time: 2.514404e+02s
   Theano Optimizer time: 8.401098e+01s
      Theano validate time: 3.725383e-01s
   Theano Linker time (includes C, CUDA code generation/compiling): 1.530572e+02s
      Import time 4.084539e-02s
      Node make_thunk time 9.101742e+01s


**Function2 profiling:**
  Total compile time: 4.739103e+00s
    Theano Optimizer time: 1.822968e+00s
       Theano validate time: 9.664536e-03s
    Theano Linker time (includes C, CUDA code generation/compiling): 2.670952e+00s
       Import time 1.080036e-03s
       Node make_thunk time 2.267937e+00s

**Function2 profiling-with caching:**
 Total compile time: 5.426061e+00s
   Theano Optimizer time: 2.152175e+00s
      Theano validate time: 9.719849e-03s
   Theano Linker time (includes C, CUDA code generation/compiling): 2.989004e+00s
      Import time 1.043558e-03s
      Node make_thunk time 2.578031e+00s


**Function3 profiling:**
  Total compile time: 3.999801e+00s
    Theano Optimizer time: 1.815703e+00s
       Theano validate time: 9.712696e-03s
    Theano Linker time (includes C, CUDA code generation/compiling): 1.938766e+00s
       Import time 0.000000e+00s
       Node make_thunk time 1.530776e+00s

**Function3 profiling-with caching:**
 Total compile time: 4.686823e+00s
   Theano Optimizer time: 2.155384e+00s
      Theano validate time: 9.829044e-03s
   Theano Linker time (includes C, CUDA code generation/compiling): 2.238642e+00s
      Import time 0.000000e+00s
      Node make_thunk time 1.828856e+00s

abergeron:
Not only is there no improvement, but it's actually slower.  This may have to do with the large amount of items that are never used.

Maybe we can change the approach and make a decorator to enable caching on specific Ops?

ReyhaneAskari:
yes, I agree, it has slowed down. We can also manage that here. We can apply the caching using the names of the classes. This way it will be more centralized. If we use decorators, each function will have its own cache and I think it won't be really different from what it was before. I can implement both and we can see which one is faster.

abergeron:
I mean decorators on classes.  That way we can still cache at the op level.

nouiz:
Why not an attribute on the OP? It would be easy to implement with the
current PR. I don't like decorator on op. People are less used to that then
adding attribute to classes.

abergeron:
On the contrary, people (well python programmers) are much more used to having a class decorator changing things in how the class works than have an attribute magically enabling functionality.

We've had to go this way before because of sweeping changes that affected all the ops, but if we want to add caching to specific ops without touching the others, then a decorator is a better option.

ReyhaneAskari:
I implemented the caching as a decorator function but I had some weird errors for class types and class methods. Suppose we have the ```class GpuAlloc(HideC, Alloc)``` with a simple decorator : 
```
def cachable(cls):
    instances = {}

    def get_cache(*args, **kwargs):
        if (str(args)) not in instances:
            instances[(str(args))] = cls(*args, **kwargs)
        return instances[(str(args))]

    return get_cache
```
While objects created using GpuAlloc() would cache perfectly, GpuAlloc itself is not a class anymore. It's a function:
```
-> type(GpuAlloc)
   <type 'function'>
```
Thus you cannot call class methods from it. Thus I implemented it as a class.

nouiz:
I'm killing travis as we need it urgently for another PR.

nouiz:
I'm kililng jenkins for rc3

nouiz:
jenkins test this

slefrancois:
jenkins test this

ReyhaneAskari:
@nouiz , @abergeron , my implementation of a class decorator did not work. The reason is that we had an error with ```GpuDnnConvDesc``` class which was using caching. ```GpuDnnConvDesc``` calls up the ```__init__``` of its super inside its ```__init__``` which gives the error: 
```python
@cachable
class GpuDnnConvDesc(COp):
....
    def __init__(self, border_mode, subsample=(1, 1), conv_mode='conv',precision="float32"):
               COp.__init__(self, ["conv_desc.c"], "APPLY_SPECIFIC(conv_desc)")
```
 I can't think of any of other implementation of the decorator that would work in this case.

In the last commit as Fred had suggested, I added an attribute for caching. It works perfectly fine. Here is a profiling of sb_resnet with 11 layers. I think the previous comparison of profilings was not fair as on the master branch we were only caching 6 Ops but in the new commit we were caching many more Ops. 
So I figured the only fair way to compare them is to have caching only on those 6 classes in the new commit. The profiling shows there is not much difference in terms of timing but in terms of memory the new commit should be better as we were not making the keys hashable in master so there were cases that we skipped the caching:

**Master:**
------------------
**Function profiling 1**:
  Message: sb/sb_resnet.py:332
  Time in 352 calls to Function.__call__: **1.423117e+02s**
  Time in Function.fn.__call__: 1.420824e+02s (99.839%)
  Time in thunks: 1.309948e+02s (92.048%)
  Total compile time: **2.646939e+02s**

**Function profiling 2**:
  Message: sb/sb_resnet.py:354
  Time in 395 calls to Function.__call__: **9.556025e+00s**
  Time in Function.fn.__call__: 9.487131e+00s (99.279%)
  Time in thunks: 9.165212e+00s (95.910%)
  Total compile time: **2.343059e+00s**

**Function profiling 3**:
Message: sb/sb_resnet.py:346
  Time in 120 calls to Function.__call__: **3.268025e+00s**
  Time in Function.fn.__call__: 3.246984e+00s (99.356%)
  Time in thunks: 3.151774e+00s (96.443%)
  Total compile time: **6.705444e+00s**

**Caching using attribute(new commit):**
------------------
**Function profiling 1**:
Message: sb/sb_resnet.py:332
  Time in 351 calls to Function.__call__: **1.418564e+02s**
  Time in Function.fn.__call__: 1.416228e+02s (99.835%)
  Time in thunks: 1.302248e+02s (91.800%)
  Total compile time: **2.668669e+02s**

**Function profiling 2**:
Message: sb/sb_resnet.py:354
  Time in 394 calls to Function.__call__: **9.562855e+00s**
  Time in Function.fn.__call__: 9.494436e+00s (99.285%)
  Time in thunks: 9.173488e+00s (95.928%)
  Total compile time: **1.553995e+00s**

**Function profiling 3**:
Message: sb/sb_resnet.py:346
  Time in 119 calls to Function.__call__: **2.851708e+00s**
  Time in Function.fn.__call__: 2.831021e+00s (99.275%)
  Time in thunks: 2.738053e+00s (96.014%)
  Total compile time: **7.805245e+00s**

ReyhaneAskari:
There are so many unused commits here. I will rebase it tomorrow if we agree on this.

theano/gof/tests/test_utils.py
```diff
@@ -61,3 +61,17 @@ def test_stack_trace():
         assert len(v.tag.trace[0]) == 2
     finally:
         theano.config.traceback.limit = orig
+
+
+class TestCaching():
+    class CachingClassExample(theano.Op):
+        __props__ = ('value',)
+        _set_cache = True
```
nouiz: I don't like the name. What about: __cache_instance = True

theano/gof/utils.py
```diff
@@ -210,6 +211,47 @@ def __str__(self):
...
+    def __call__(cls, *args, **kwargs):
+        _set_cache = cls.__dict__.get('_set_cache', None)
+        if _set_cache:
+            key = [cls]
+            for arg in args:
+                arg = make_hashable(arg)
+                key.append(arg)
+            for k, v in sorted(kwargs.items()):
+                key.append(k)
+                v = make_hashable(v)
```
nouiz: I don't like this idea of make_hashable, but I'm not sure we should remove it.

@lamblin @abergeron what do you think?

theano/gof/tests/test_utils.py
```diff
@@ -79,8 +79,8 @@ def __init__(self, var):
     class CachingClassExample_3(theano.Op):
         __props__ = ()
 
-        def __init__(self, var):
-            self.var = var
+        def __init__(self):
+            pass
```
ReyhaneAskari: I did implement this test at the time you asked. But these tests are not valid anymore as we have changed the implementation.

theano/gof/tests/test_utils.py
ReyhaneAskari: Sure, I'll change the name.

theano/gof/utils.py
```diff
@@ -217,22 +217,24 @@ def __str__(self):
...
         # Note: when the props is empty inspect.getargspec(cls.__init__)
         #       gives an error since cls.__init__ is not a python function.
 
         props = cls.__dict__.get('__props__', None)
-
-        # The cases are being separated for code readability.
+        try:
+            init_args = sorted(inspect.getargspec(cls.__init__)[0][1:])
+        except:
+            init_args = None
```
ReyhaneAskari: If the __init__ was defined in the super, inspect can find it. I had the exceptions when the init was a slot wrapper. Thus it was not a function and inspect could not find it.

theano/gof/utils.py
ReyhaneAskari: My very first implementation was only caching the ops with valid keys. The benchmarking does not show any significant change in the timing so at least for now it is not causing any problems.

Suppose a simple case that user passes a list instead of a tuple. It's very easy to convert it to a hashable key. Maybe we can remove the ```frozendict ``` or ```hash_from_ndarray``` if you think they might be bottle necks.

abergeron:
I still don't like caching enabled with an attribute.  If we make the decorator slightly more complex we can avoid all the issues:

```python
import inspect
import functools

def cached(cache_keys):
    def nindex(l, v, default):
        try:
            return l.index(v)
        except ValueError:
            return default

    def wrap(cls):
        # We don't want to touch 'self' here                                    
        regular_args = inspect.getargspec(cls.__init__).args[1:]
        cache_map = [nindex(regular_args, k, k) for k in cache_keys]
        def get_val(mi, args, kwargs):
            if isinstance(mi, int):
                return args[mi]
            else:
                return kwargs[mi]
        class Cache(cls):
            __cache = {}
            def __new__(cls2, *args, **kwargs):
                cache_key = tuple(get_val(mi, args, kwargs) for mi in cache_map\
)
                val = cls2.__cache.get(cache_key, None)
                if val is None:
                    val = object.__new__(Cache)
                    cls.__init__(val, *args, **kwargs)
                    cls2.__cache[cache_key] = val
                return val

            def __init__(self, *args, **kwargs):
                pass

        functools.update_wrapper(Cache, cls, updated=())
        return Cache
    return wrap
```

theano/gof/utils.py
```diff
@@ -565,6 +607,30 @@ def hash_from_code(msg):
             return hashlib.md5(np.getbuffer(msg)).hexdigest()
 
 
+def hash_from_ndarray(data):
```
abergeron: This function has weird intentation

nouiz:
@abergeron, we already use attribute for many related thinks like props.
Why do you want developer to use 2 mechanims? This don't seem good. It put
the special code not all at the same places. I don't think this is more
Pythonic.

theano/gof/utils.py
ReyhaneAskari: @nouiz , @abergeron , right now we are only caching the 6 ops that had the caching implementation in master. In these 6 ops, ```make_hashable()``` is never used. I can safely remove it. Since at the start we were caching all the ops with ```__props___```, we had so many weird ```args``` and ```kwargs``` so we implemented the ```make_hashable()```.  We can comment it for now and reuse it if necessary in the future. What to you think?

ReyhaneAskari:
@abergeron , thanks for your comments. I am not against the decorators and can continue working on it, but I think using an attribute is simpler. As @nouiz  also mentioned we are already using attributes for other tasks like props so using the attributes is more known to the users. Other than that, I believe using the meta class for caching is more intuitive. Instead of having a complex function as decorator, we can have a much simpler and straightforward implementation with meta class.

ReyhaneAskari:
@lamblin 
I ran the ```sb_resnet``` with 11 layers on both master and a version of theano without caching to see the difference. There is a small improvement. It might be more significant in other cases. 

 **Function profiling - no chaching**
  
  Message: sb/sb_resnet.py:332
  Time in 352 calls to Function.__call__: **1.232502e+02s**
  Time in Function.fn.__call__: 1.230624e+02s (99.848%)
  Time in thunks: 1.150734e+02s (93.366%)
  Total compile time: 2.375120e+02s
  Number of Apply nodes: 9305

**Function profiling - master**
  
  Message: sb/sb_resnet.py:332
  Time in 352 calls to Function.__call__: **1.217683e+02s**
  Time in Function.fn.__call__: 1.215773e+02s (99.843%)
  Time in thunks: 1.135268e+02s (93.232%)
  Total compile time: 2.928643e+02s
    Number of Apply nodes: 9305

abergeron:
I just want to note that your timings seem to indicate that the current caching is a major slowdown.

We go from 292s (current master) to 237s (no caching) compile time.  Either something is wrong or the current cache has way too much overhead.

Can you also do timings with the new cache approach to see if the overhead is similar?

ReyhaneAskari:
@abergeron @lamblin @nouiz 

Here is comprehensive benchmark of the first function in sb_resnet with 11 layers of _master_ vs the new commit (_clean up_) vs having _no caching_ at all. So what do you think?

|                             | calls to Function.__call__     | total compile time | Optimizer time | Time since theano import |
|-----------------------------|---------------|--------------------|----------------|--------------------------|
| Clean cache, run master 1   | 8.203986e+01s | 1.338113e+02s      | 1.441458e+01s  | 252.919s                 |
| run master 2                | 8.244323e+01s | 5.724723e+01s      | 9.412006e+00s  | 166.474s                 |
| run master 3                | 8.298326e+01s | 5.721275e+01s      | 9.346654e+00s  | 167.419s                 |
| **Avg. master**                 | **8.248878e+01s** | **8.2757093e+01s**     | **1.105774e+01s**  | **195.604s**                 |
| Clean cache, run cleanup 1  | 8.482095e+01s | 1.380990e+02s      | 1.401109e+01s  | 258.559s                 |
| run cleanup 2               | 8.303796e+01s | 5.764808e+01s      | 9.565826e+00s  | 168.739s                 |
| run cleanup 3               | 8.210065e+01s | 5.750025e+01s      | 9.542340e+00s  | 167.883s                 |
| **Avg. Cleanup**                | **8.331985e+01s** | **8.4415776e+01s**     | **1.1039e+01s**    | **198.3936s**                |
| Clean cache, run no cache 1 | 8.239032e+01s | 1.344595e+02s      | 1.473346e+01s  | 255.532s                 |
| run no cache 2              | 8.212105e+01s | 5.756044e+01s      | 9.738374e+00s  | 167.802s                 |
| run no cache 3              | 8.318885e+01s | 5.813005e+01s      | 9.820622e+00s  | 169.673s                 |
| **Avg. no cache**               | **8.256674e+01s** | **8.33833e+01s**       | **1.143081e+01s**  | **197.669s**                 |

ReyhaneAskari:
Hi all,

I tried to increase the size of _sb_resnet_ but I had memory problems. In the smaller size the linker time was still smaller than the optimizer time. So I decided to use another benchmark with 225 layers of dilated convolution with residual connections and batch norm. In this benchmark the linker time is less than half of optimizer time. Here is the results. The improvement is not really significant. 

|                             | total compile time | Optimizer time | Linker time | Time since theano import | GraphToGPU |
|-----------------------------|--------------------|----------------|-------------|--------------------------|-----------------------------|
| Clean cache, run master 1   | 635.4937s          | 420.8941es     | 181.2463s   | 661.546s                 | 39.129510s                  |
| run master 2                | 564.7791s          | 417.5182s      | 113.9806s   | 587.208s                 | 39.351908s                  |
| run master 3                | 564.2408s          | 416.9501s      | 113.7625s   | 586.810s                 | 39.299309s                  |
| **Avg. run 2, 3 master**        | **564.5099s**          | **417.2341s**      | **113.8715s**   | **587.009s**                 | **39.325608s**                  |
| Clean cache, run cleanup 1  | 633.4908s          | 419.0257s      | 181.0189s   | 658.936s                 | 39.121614s                  |
| run cleanup 2               | 563.7316s          | 416.4784s      | 113.6587s   | 586.170s                 | 39.342534s                  |
| run cleanup 3               | 563.3795s          | 416.8252s      | 113.2907s   | 586.222s                 | 39.181378s                  |
| **Avg. run 2, 3 Cleanup**       | **563.5555s**          | **416.6518s**      | **113.4747s**   | **586.196s**                 | **39.261956s**                  |
| Clean cache, run no cache 1 | 638.4234s          | 423.1799s      | 181.9027s   | 664.099s                 | 39.364141s                  |
| run no cache 2              | 567.8251s          | 420.8730s      | 113.7029s   | 590.605s                 | 39.433293s                  |
| run no cache 3              | 563.7638s          | 417.0186s      | 113.4645s   | 586.387s                 | 39.374739s                  |
| **Avg. run 2, 3 no cache**      | **565.7944s**          | **418.9458s**      | **113.5837s**   | **588.496s**                 | **39.404016s**                  |

lamblin:
I just thought about it: at some point, we prevented the scalar operations (for shape computations for instance) from being lifted to the GPU. Could it be that it reduced the number of lifted elements so much that the cache is no longer necessary?
A way of testing that would be to get the old line profiles back and see how many HostFromGPU nodes (for instance) were created, and how many are created now, for the same exact model.
Another way would be to temporarily re-activate the lifting of scalars, and see if it makes a difference in the timing.

abergeron:
I'm not sure it warrants that much exploration.  In the current state of things the best the cache seems to buy us is a 0.3-0.5% improvement in optimization time.

Maybe it did make more sense in the past, but we're not going to roll the rest of the code back to a previous state to justify the cache code.  I think we should just remove it.

ReyhaneAskari:
I do agree that the caching does not help with the timing and we are using the memory at the same time. So the trade off is not reasonable. The only thing that I can think of is to add the caching to some ```ops``` that are usually frequently used but are not being cached right now and see if that would make any significant difference. I had [this benchmark ](https://github.com/Theano/Theano/pull/5579#issuecomment-283775041)previously that counted the number times the cache was used for each ```op```.  Maybe adding it to ```DimShuffle``` or any other ```op``` that you think might need to be cached would be helpful.

abergeron:
The benchmark with the number of uses you did indicate that GpuFromHost was reused 2833 times and DimShuffles are reused at most 1075 times and are more complex to compare.  The current code caches GpuFromHost and it barely makes a difference, so I'm not expecting much from caching DimShuffle.

If we have timings that demonstrate an appreciable speedup from caching certain ops, then by all means, we should keep it.  However in the absence of such evidence, I err on the side of less code.

ReyhaneAskari:
I talked with Fred and he also thinks that we should remove all the caching. He told me to keep the ```meta_class``` implementation in the history. So there will be a commit for that in case we needed it in the future.

lamblin:
I'm OK with that plan.

Issue #5586, opened 2017-02-20
Test fail deterministically in DebugMode theano/scan_module/tests/test_scan_opt.py:TestGaussNewton.test_nobatch


nouiz:
This test always fail in DebugMode.

THEANO_FLAGS=DebugMode.patience=1,mode=DebugMode theano-nose theano/scan_module/tests/test_scan_opt.py:TestGaussNewton.test_nobatch -s --pdb --pdb-failure

It is the scanOp_save_mem opt that seem to shift all result when it should not do so. Here is the error:
```
======================================================================
ERROR: test_nobatch (theano.scan_module.tests.test_scan_opt.TestGaussNewton)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/u/bastienf/repos/Theano/theano/scan_module/tests/test_scan_opt.py", line 119, in test_nobatch
    class GaussNewtonMatrix(object):
  File "/u/bastienf/repos/Theano/theano/scan_module/tests/test_scan_opt.py", line 106, in _run
    f()
  File "/u/bastienf/repos/Theano/theano/compile/function_module.py", line 884, in __call__
    self.fn() if output_subset is None else\
  File "/u/bastienf/repos/Theano/theano/compile/debugmode.py", line 2305, in deco
    return f()
  File "/u/bastienf/repos/Theano/theano/compile/debugmode.py", line 2230, in f
    r_vals)
  File "/u/bastienf/repos/Theano/theano/compile/debugmode.py", line 1078, in _find_bad_optimizations0
    new_graph=new_graph_str)
BadOptimization: BadOptimization Error 
  Variable: id 140677454264592 Subtensor{::int64}.0
  Op Subtensor{::int64}(Subtensor{int64:int64:int8}.0, Constant{1})
  Value Type: <type 'numpy.ndarray'>
  Old Value shape, dtype, strides: (10, 10) float64 (80, 8)

  Old Value:  [[ 0.45176911  0.436364    0.44078463  0.43683308  0.4447106   0.42295613
   0.50847826  0.47602453  0.45637602  0.47293546]
 [ 0.51382751  0.51165349  0.47543479  0.53047943  0.50142264  0.48407497
   0.51621107  0.51573665  0.50520342  0.53558663]
 [ 0.55589683  0.49987857  0.49289679  0.57590271  0.51975115  0.51570383
   0.55758665  0.54078606  0.54665063  0.54495092]
[...]

PR #5587, opened 2017-02-20
Scan fix

nouiz:
fix gh-5586


theano/scan_module/scan_opt.py
```diff
@@ -1516,15 +1516,10 @@ def sanitize(x):
...
-            # twice since bad things usually happen if I do that
-            # TODO: why not check if save mem was done on any of merged nodes?
-            #       That way, if none of them had save mem applied, it would
-            #       be applied later.
-            info['_scan_savemem_visited'] = True
-
             # TODO: currently we don't support scan with 0 step. So
             # don't create one.
+            # For test, mark that savemem have optimized this node
+            info['_scan_savemem_visited'] = True
```
abergeron: Is this for a test of just because it's required?

nouiz:
For a specific test that exist and fail without that. I didn't found
quickly a few to do an equivalent test directly on the optimized graph to
find that information.

PR #5597, opened 2017-02-21
Fix various GPU unit tests on Windows

notoraptor:
This PR fixes some unit tests that fail on GPU on Windows.
___
### GPU test_reduction makes NVIDIA driver crash.

image

I resolved it by reducing the size of test values from 10 000 000 to 10 000. We initially use a big size to check if GPU is faster than CPU on big data, but I think it is not really useful for simple unit testing.

PS: But it is quite strange, as this test does not fail on my Ubuntu partition.

___

### test_noncontiguous_idx fails for subtensor on GPU.
This test;
```
nosetests -vs theano/gpuarray/tests/test_subtensor.py:G_subtensor.test_noncontiguous_idx
```
fails with this output:
```
Using cuDNN version 5110 on context None
test_noncontiguous_idx (theano.gpuarray.tests.test_subtensor.G_subtensor) ... FAIL

FAIL: test_noncontiguous_idx (theano.gpuarray.tests.test_subtensor.G_subtensor)
  File "C:\Users\HPPC\mila\dev\git\theano\theano\tensor\tests\test_subtensor.py", line 538, in test_noncontiguous_idx
    val = self.eval_output_and_check(t, op_type=self.adv_sub1, length=2)
  File "C:\Users\HPPC\mila\dev\git\theano\theano\tensor\tests\test_subtensor.py", line 114, in eval_output_and_check
    assert_equal(len(topo_), length)
AssertionError: 3 != 2

Ran 1 test in 3.889s

FAILED (failures=1)
```
This test checks that the length of the function graph is exactly 2 when ignoring a list of operations. On GPU the ignored operations are `HostFromGpu, GpuFromHost, DeepCopyOp, GpuContiguous` ( https://github.com/Theano/Theano/blob/master/theano/gpuarray/tests/test_subtensor.py#L39 ). But the function graph (defined here: https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_subtensor.py#L110 ) looks like this on my computer:
```
HostFromGpu(gpuarray) [id A] ''   4
 |GpuAdvancedSubtensor1 [id B] ''   3
   |<GpuArrayType<None>(float64, (False, False, False))> [id C]
   |GpuContiguous [id D] ''   2
     |GpuElemwise{Cast{int64}}[]<gpuarray> [id E] ''   1
       |GpuSubtensor{::int64} [id F] ''   0
         |<GpuArrayType<None>(int32, (False,))> [id G]
         |Constant{2} [id H]
```
So there are 3 operations counted when we let the ignored ones: `GpuAdvancedSubtensor1`, `GpuElemwise{Cast{int64}}` and `GpuSubtensor{::int64}`. Then, it is the cast{int64} op that seems extra. Indeed, it seems it is created here: https://github.com/Theano/Theano/blob/master/theano/gpuarray/subtensor.py#L414 :
```
        if ilist__.type.dtype != 'int64':
            ilist__ = tensor.cast(ilist__, 'int64')
```
So it is certainly (again) because index list is `int32` on Windows.

To fix it, I have just updated the creation of index list in the test, by casting it to `int64` before it is passed to `make_node()`, as it is the test only that checks the number of ops (it should not be a real use-case).

___

I'm looking for some last other bugs, so I will update this PR tomorrow (unit tests take much time to run on my computer!).

@abergeron @nouiz @lamblin

### DIFF
--- a/theano/gpuarray/tests/test_reduction.py
+++ b/theano/gpuarray/tests/test_reduction.py
@@ -14,7 +14,7 @@ from .. import GpuArrayType
 import math
 
 # Number of values to be used in test tensors (except with 0-D tensors!).
-test_size = 10000000
+test_size = 10000
 
 # NB: This order of "unsorted axes" is arbitrary and is here
 # just to have the same informations on profile output
--- a/theano/gpuarray/tests/test_reduction.py
+++ b/theano/gpuarray/tests/test_reduction.py
@@ -14,7 +14,7 @@ from .. import GpuArrayType
 import math
 
 # Number of values to be used in test tensors (except with 0-D tensors!).
-test_size = 10000
+test_size = 10000000
 
 # NB: This order of "unsorted axes" is arbitrary and is here
 # just to have the same informations on profile output
--- a/theano/tensor/tests/test_subtensor.py
+++ b/theano/tensor/tests/test_subtensor.py
@@ -533,7 +533,7 @@ class T_subtensor(unittest.TestCase, utt.TestOptimizationMixin):
         data = rand(4, 2, 3)
         idx = [2, 2, 0, 0, 1, 1]
         n = self.shared(data)
-        t = n[self.shared(numpy.asarray(idx))[::2]]
+        t = n[self.shared(numpy.asarray(idx).astype('int64'))[::2]]
         self.assertTrue(isinstance(t.owner.op, tensor.AdvancedSubtensor1))
         val = self.eval_output_and_check(t, op_type=self.adv_sub1, length=2)
         utt.assert_allclose(data[idx[::2]], val)
--- a/theano/gpuarray/reduction.py
+++ b/theano/gpuarray/reduction.py
@@ -141,4 +141,4 @@ class GpuMaxAndArgmax(Op):
         """ % {'name': name, 'X': inputs[0]}
 
     def c_code_cache_version(self):
-        return (1,1)
+        return (1, 1)
--- a/theano/gpuarray/reduction.py
+++ b/theano/gpuarray/reduction.py
@@ -61,6 +61,8 @@ class GpuMaxAndArgmax(Op):
             #endif
         #endif
 
+        int err = 0;
+
         unsigned  %(name)s_redux_len = PyTuple_GET_SIZE(%(axes)s);
         unsigned* %(name)s_axes_to_reduce = (unsigned*)malloc(%(name)s_redux_len * sizeof(unsigned));
         for (unsigned i = 0; i < %(name)s_redux_len; ++i) {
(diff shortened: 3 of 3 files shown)


notoraptor:
No more bugs detected after running all `test_*.py` files into `theano/gpuarray/tests`. Some tests are skipped because they required `scikits.cuda` or `nervanagpu`, but I don't find versions of these packages that are ready for Windows, so I let these skipped tests.

A this point I have finished to test libgpuarray and theano new backend on Windows (8.1 64 bits, Python 2.7.13).

This PR can be reviewed from now!

theano/tensor/tests/test_subtensor.py
```diff
@@ -533,7 +533,7 @@ def test_noncontiguous_idx(self):
         data = rand(4, 2, 3)
         idx = [2, 2, 0, 0, 1, 1]
         n = self.shared(data)
-        t = n[self.shared(numpy.asarray(idx))[::2]]
+        t = n[self.shared(numpy.asarray(idx).astype('int64'))[::2]]
```
nouiz: ok for this change.

theano/gpuarray/tests/test_reduction.py
```diff
@@ -14,7 +14,7 @@
 import math
 
 # Number of values to be used in test tensors (except with 0-D tensors!).
-test_size = 10000000
+test_size = 10000
```
nouiz: We need to find the good fix. This just hide the error.

notoraptor:
Just an update to back to big size for tested values into `test_reduction.py` and make error message more explicit.

As suggested by @nouiz I was looking for the real origin of this error. This is the test I run on Windows:

```
set CUDA_LAUNCH_BLOCKING=1

nosetests -xvs theano\gpuarray\tests\test_reduction.py:TestColumn.test_none > out.log 2>&1
```

The column is a matrix with shape `(10 000 000, 1)`. `test_none` will test the reduction on all axes (thus, a scalar should be returned). But it makes my driver crash, and this is the content of `out.log`:

```
Using cuDNN version 5110 on context None
test_none (theano.gpuarray.tests.test_reduction.TestColumn) ... ERROR

ERROR: test_none (theano.gpuarray.tests.test_reduction.TestColumn)
  File "C:\Users\HPPC\mila\dev\git\theano\theano\gpuarray\tests\test_reduction.py", line 143, in test_none
    self.compute(None)
  File "C:\Users\HPPC\mila\dev\git\theano\theano\gpuarray\tests\test_reduction.py", line 131, in compute
    self.compute_gpu(test_gpu_tensor, test_host_tensor, axis)
  File "C:\Users\HPPC\mila\dev\git\theano\theano\gpuarray\tests\test_reduction.py", line 120, in compute_gpu
    f(test_gpu_tensor)
  File "C:\Users\HPPC\mila\dev\git\theano\theano\compile\function_module.py", line 898, in __call__
  File "C:\Users\HPPC\mila\dev\git\theano\theano\gof\link.py", line 325, in raise_with_op
  File "C:\Users\HPPC\mila\dev\git\theano\theano\compile\function_module.py", line 884, in __call__
RuntimeError: GpuMaxAndArgmax: unable to compute gpuarray maxandargmax: error 3: Unknown device error (unspecified launch failure).
Apply node that caused the error: GpuMaxAndArgmax{axis=(0, 1)}(<GpuArrayType<None>(float64, (False, False))>)
Toposort index: 0
Inputs types: [GpuArrayType<None>(float64, (False, False))]
Inputs shapes: [(10000000, 1)]
Inputs strides: [(8, 8)]
Inputs values: ['not shown']
Outputs clients: [[HostFromGpu(gpuarray)(GpuMaxAndArgmax{axis=(0, 1)}.0)], [HostFromGpu(gpuarray)(GpuMaxAndArgmax{axis=(0, 1)}.1)]]

Ran 1 test in 3.776s

FAILED (errors=1)

```

So, the error is raised when calling `GpuArray_maxandargmax()`, and the error message returned by GpuArray_error() is `unspecified launch failure`.

Note that there is no error when I try to reduce on juste one axis (either axis 0 or axis 1). The bug seems to happen only when we want to reduce all axes.

@abergeron

abergeron:
That last error is probably just because the kernel is long-running since the all-axes case is not optimized yet.

Let's switch the size to 10000 since it will work on all platforms and it doesn't really reduce test coverage.

theano/gpuarray/tests/test_reduction.py
nouiz: After discussion with @abergeron this is the right fix. It don't lower the quality of the test. The problem is that our reduction with axis =1 use only 1 thread on the GPU (so it isn't used by default yet. We still work to optimize that case). So this is super slow and it bust the 5s limit on windows per kernel call.

Issue #5606, opened 2017-02-23
remove the benchmark/ folder


lamblin:
Most of its content seems non-functional, and it has not been used in years.

PR #5609, opened 2017-02-23
Remove the old benchmark directory.

nouiz:
They use the module system that was removed years ago. Now the benchmark are outside Theano.

Someone tried to run them recently.

fix gh-5606


Sentient07:
Also, it'd good if someone could do the benchmarks, and update the results in Soumith's convnet benchmark repository. I'd like to do, but I don't have powerful GPUs.

nouiz:
This is not Soumith's benchmark. This is much older benchmark that don't
run with current Theano version. I don't think there is anything good here.

We already use the deep learning tutorial as a speed benchmark in our daily
buildbot. This include models like the one I remove.

On Thu, Feb 23, 2017 at 8:21 AM Ramana Subramanyam <notifications@github.com>
wrote:

Also, it'd good if someone could do the benchmarks, and update the results
in Soumith's convnet benchmark repository. I'd like to do, but I don't have
powerful GPUs.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Theano/Theano/pull/5609#issuecomment-281989788>, or mute
the thread

nouiz:
travis passed, this don't affect the GPU so merging.

Sentient07:
I meant it'd be good if someone could run those convnet models and update with the current results there. That repository has a great number of watchers, and it doesn't look good to be slowest (which, currently isn't / wouldn't be true).  :)

nouiz:
I don't understand what you mean.

PR #5612, opened 2017-02-23
Implement a struct generator for wrapping op params in Python and C codes.

notoraptor:
According to @abergeron guidelines, this PR proposes an implementation of a struct generator that allows to pack many Theano variables into one object and make them accessible from both `perform()` and `c_code()` methods of ops. The generator is intended to be used to create param ops.

I have much work to do again (_re-read the code, check if it can be simplified or if I have forgotten some things, check if I can write hash methods differently, make more tests and extend the current unit testing_), but I think the current code can already be reviewed to check if the global design and the global ideas are suitable ! Also, I don't really know where to put the new class, so I have created a new module called `theano.common` (as this generator could be used in all other modules), but I can move it elsewhere if needed.

There is currently a working unit test, and a small "tutorial" at the top of the generator implementation.

@abergeron @lamblin @nouiz !


slefrancois:
jenkins test this

theano/common/tests/test_wrapper.py
```diff
@@ -0,0 +1,204 @@
...
+from theano.gof import Op, Apply
+from theano import Generic
+from theano.tensor import TensorType
+from theano.common import Wrapper, Wrap
+from theano import config
+from theano import tensor
+from theano.tests import unittest_tools as utt
+
+dtype = config.floatX
+ScalarType = TensorType(dtype, tuple())
```
nouiz: As this is an instance, use scalar_type, not a camel case name.

theano/common/tests/test_wrapper.py
```diff
@@ -0,0 +1,204 @@
...
+        assert w.values_eq_approx(o, o2)
+
+        # Check value_eq_approx.
+        o3 = Wrap(a1=numpy.asarray([[1, 2.0, 3.000, 4, 5.0, 6], [7, 8, 9, 10, 11, 12]]).astype('int32'),
+                  a2=random_tensor.astype('float32'),
+                  a3=2000.0)
+
+        assert w.values_eq_approx(o, o3)
+
+    def test_wrapper(self):
```
nouiz: rename to test_op_params

theano/common/tests/test_wrapper.py
```diff
@@ -0,0 +1,204 @@
+from __future__ import absolute_import, print_function, division
```
nouiz: I think you should not create a new module, but put this inside theano/gof module.

theano/common/wrapper.py
```diff
@@ -0,0 +1,448 @@
...
+        >>> d = dict(a=1, b=2, c='test')
+        >>> w2 = Wrap(**d)
+        >>> print(w2.a, w2.b, w2.c)
+
+    """
+
+    def __init__(self, **kwargs):
+        if len(kwargs) == 0:
+            raise TypeError('Wrap: cannot wrap empty data.')
+        super(Wrap, self).__setattr__('data', kwargs)
```
nouiz: Add why you don't put them directly in self.

theano/common/wrapper.py
```diff
@@ -0,0 +1,448 @@
...
+                    attributes += (numpy.asscalar(self.data[k]),)
+                else:
+                    attributes += tuple(self.data[k])
+            else:
+                try:
+                    iter(self.data[k])
+                except TypeError:
+                    attributes += (self.data[k],)
+                else:
+                    attributes += tuple(self.data[k])
```
nouiz: This is to cast list to tuple? Why do we want to support list here?

theano/common/tests/test_wrapper.py
```diff
@@ -0,0 +1,204 @@
...
+        y[0] = coefficients.a * (x**2) + coefficients.b * x + coefficients.c
+
+    def c_code_cache_version(self):
+        return (1, 1)
+
+    def c_support_code_apply(self, node, name):
+        float_type = node.inputs[0].type.dtype_specs()[1]
+        return """
+        /* Computes: x = a*x*x + b*x + c for x in matrix. */
+        int quadratic_%(float_type)s(PyArrayObject* matrix, %(float_type)s a, %(float_type)s b, %(float_type)s c) {
```
nouiz: you support more then matrix, so rename it to something like tensor or input.

theano/common/tests/test_wrapper.py
```diff
@@ -0,0 +1,204 @@
...
+    def c_code(self, node, name, inputs, outputs, sub):
+        X = inputs[0]
+        Y = outputs[0]
+        coeff = sub['params']
+        fail = sub['fail']
+        float_type = node.inputs[0].type.dtype_specs()[1]
+        float_typenum = numpy.dtype(node.inputs[0].type.dtype).num
+        coeff_type = 'npy_' + numpy.dtype(dtype).name
+        return """
+        PyArrayObject* o_a = %(coeff)s.a;
```
nouiz: It is good to test it with Tensor input object. But can you make one of the inputs a theano.scalar.Scalar()?

I think more people will use that, as it should give simpler code to access that data.

theano/common/wrapper.py
notoraptor: I need to remember the keys and values in kwargs because I use them to hash the object, so I don't want to confound them with builtin object attributes. I'll add a comment about that.

theano/common/wrapper.py
notoraptor: I assume that the values of op params can be lists or tuples, that could be converted later to ndarrays when passed to `theano_type.filter(data)` when `strict==False`, for example. So I try to support any iterable object here.

But indeed, I don't need to cast to tuple if the object is already iterable. I can directly write `attributes += self.data[x]`.

nouiz:
There is `from theano.tensor.utils import hash_from_ndarray` that would
help you.

I think we should have the filter called on the data. This would simplify
the logic for hash and eq. You won't need to handle the case where the
param is a list/tuple that will be casted to an ndarray for example.

For this, I propose that in the class COp, we add/modify the method
get_param, such that if the param_type is an instance of Wrapper and all
the keys are in \__props__, it return an instance of Wrap, where he would
have called filter on the value.

This would request less user code too.

notoraptor:
Rebased (flake8 commits removed), and @nouiz comments applied.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+        c_init_list = []
+        c_cleanup_list = []
+        c_extract_list = []
+        for attribute_name, type_instance in zip(self.fields, self.types):
+            type_name = type_instance.__class__.__name__
+
+            try:
+                c_declare_list.append(type_instance.c_declare(attribute_name, sub))
+            except MethodNotDefined:
+                raise RuntimeError('Wrapper: class "%s" should implement method Type.c_declare().' % type_name)
```
abergeron: I don't this we need customized error messages at this point.  Also catching and re-raising exceptions is super annoying when trying to debug with pdb since we don't get the original error location.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+            }
+            for (int i = 0; i < %(length)s; ++i) {
+                if (!PyObject_HasAttrString(py_%(name)s, fields[i])) {
+                    PyErr_Format(PyExc_TypeError, "Wrapper: missing expected attribute %%s in object.", fields[i]);
+                    %(fail)s
+                }
+            }
+        }
+        for (int i = 0; i < %(length)s; ++i) {
+            PyObject* o = PyObject_GetAttrString(py_%(name)s, fields[i]);
```
abergeron: You have to check if the return value is NULL and %(fail)s in that case.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+            %(name)s.extract(o, i);
+            if (%(name)s.errorOccurred()) {
+                PyErr_Format(PyExc_ValueError, "Wrapper: error when extracting value for attribute \\"%%s\\".", fields[i]);
```
abergeron: There should already be a python exception here (from the extract code of the inner type) that you are masking with this.  I don't know which message would be more useful though.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+            if (py_%(name)s == Py_None) {
+                PyErr_SetString(PyExc_ValueError, "Wrapper: expected an object, not None.");
+                %(fail)s
+            }
+            for (int i = 0; i < %(length)s; ++i) {
+                if (!PyObject_HasAttrString(py_%(name)s, fields[i])) {
+                    PyErr_Format(PyExc_TypeError, "Wrapper: missing expected attribute %%s in object.", fields[i]);
+                    %(fail)s
+                }
+            }
```
abergeron: No need to do that check since GetAttrString will fail below when the attribute is not present anyway.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+        length = self.length
+        fields_list = '"%s"' % '", "'.join(self.fields)
+        check = 1 if check_input else 0
+        return """
+        const char* fields[] = {%(fields_list)s};
+        if (%(check)s) {
+            if (py_%(name)s == Py_None) {
+                PyErr_SetString(PyExc_ValueError, "Wrapper: expected an object, not None.");
+                %(fail)s
+            }
```
abergeron: Since this check is really fast anyway, you could always do it.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+
+    """
+
+    def __init__(self, **kwargs):
+        if len(kwargs) == 0:
+            raise TypeError('Wrap: cannot wrap empty data.')
+        # We want to use only the params provided in kwargs to hash the object,
+        # so I prefer to put them into a separate attribute (self.data) instead
+        # of directly in self.__dict__, to avoid confusion with builtin fields.
+        super(Wrap, self).__setattr__('data', kwargs)
```
abergeron: In this case you could consider making Wrap inherit from dict so that we have a separate dict, but also so that access to the values from C is faster and doesn't go through python.

But this is not a blocker.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def check_that_values_are_compatible(self, data, strict, allow_downcast):
+        wrap_instance = dict()
+        for i in range(self.length):
+            wrap_instance[self.fields[i]] = self.types[i].filter(getattr(data, self.fields[i]), strict, allow_downcast)
+        return Wrap(**wrap_instance)
+
+    # Returns a wrapped object with expected attributes or (in strict mode) checks that data has expected attributes.
+    def filter(self, data, strict=False, allow_downcast=None):
+        if strict:
```
notoraptor: So, I will have to convert the dict into a Wrap. Is it allowed in strict mode (I thought the data should be kept as is in strict mode) ?

theano/gof/wrapper.py
```diff
@@ -0,0 +1,446 @@
...
+        return True
+
+    def c_compile_args(self):
+        c_compile_args_list = []
+        for _type in self.types:
+            try:
+                try:
+                    c_compile_args_list.extend(_type.c_compile_args())
+                except TypeError:
+                    c_compile_args_list.extend(_type.c_compile_args(compiler))
```
notoraptor: I use GCC_compiler (imported at line 49: `from theano.gof.cmodule import GCC_compiler as compiler`).

notoraptor: NB: Understood, I should get the compiler from the Wrapper.c_compile_args()` args (`def c_compile_args(self, compiler)`) and pass it to underlying types when it's necessary. Same for following methods.

theano/gof/wrapper.py
abergeron: In that case, probably not.  However, please try to avoid three times what is essentially the same code.

You could have a if not strict section first that converts the dict to a Wrap and then check the properties.

theano/gof/wrapper.py
notoraptor: I think that, even if Wrap inherits from dict, I will still have to call `PyObject_GetAttrString()` (or `PyDict_GetItemString()`) (to transfer values from Wrap to C struct), isn't it ?

notoraptor:
I have done a final reading, then I think the code is ready and can be completely reviewed from now!

NB: I don't know if `Wrapper.value_eq` and `Wrapper.value_eq_approx` have the right behavior from now. `value_eq` is strongly strict, whereas `value_eq_approx` is very tolerant (not strict, downcasting allowed).

theano/gof/wrapper.py
abergeron: You can use PyDict_GetItemString(), which won't use your __getattr__ python implementation.

notoraptor: @abergeron I was just thinking: the current implementation of Wrap allows to write something like `param.attribute` in `perform()` method. Then this syntax is currently allowed in both Python and C codes.

If I implement Wrap from a dict (without overriding `__getattr__`), that won't be possible: the user will have to write something like `param['attribute']` (in `perform()` code) which seems less "user-friendly", I think.

abergeron: You can both inherit from dict and use your custom `__getattr__`.  Then the C code for extraction can use `PyDict_GetItemString()` and the python code can use `param.attribute`.

notoraptor: Done.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,419 @@
...
+        for k in keys:
+            types += (type(self[k]),)
+            if isinstance(self[k], numpy.ndarray):
+                # Note: hash_from_ndarray returns a string, so the hash is not yet complete
+                # (__hash__ must return an integer).
+                attributes += (hash_from_ndarray(self[k]),)
+            else:
+                # No checking, data should be hashable.
+                attributes += (self[k],)
+        return hash((type(self),) + tuple(keys) + tuple(types) + tuple(attributes))
```
nouiz: As you inherit from dict and implement \__hash__, can you  override the methods that will make sure people won't modify the content of the dict? I mean, don't add new keys, don't remove keys, don't change the values of keys.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,419 @@
...
+        if type(self) != type(other) or len(self) != len(other):
+            return False
+        for k in self:
+            if k not in other or not (isinstance(self[k], type(other[k])) and isinstance(other[k], type(self[k]))):
+                return False
+            if isinstance(self[k], numpy.ndarray):
+                if not numpy.allclose(self[k], other[k]):
+                    return False
+            elif self[k] != other[k]:
+                return False
```
nouiz: I don't like implementing special cases for some type here. Why not merge the class Wrap and Wrapper into 1 class? If we keep 2 classes, I think the Wrap class need the type from the Wrapper class, not just the value. It then can use type.values_eq() to compare them. This won't request new special cases for each type of variables we have in Theano. For the hash, we could create a Theano Constant of each values and use the method signature() from it in the hash.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,419 @@
...
+        if len(kwargs) == 0:
+            raise ValueError('Cannot create Wrapper from empty data.')
+
+        for attribute_name in kwargs:
+            if re.match('^[A-Za-z_][A-Za-z0-9_]*$', attribute_name) is None:
+                raise SyntaxError('Wrapper: attribute "%s" should be a valid identifier.' % attribute_name)
+            type_instance = kwargs[attribute_name]
+            type_name = type_instance.__class__.__name__
+            if not isinstance(type_instance, Type):
+                raise TypeError('Wrapper: attribute "%s" should inherit from theano Type, got "%s".'
```
nouiz: theano -> Theano

theano/gof/wrapper.py
```diff
@@ -0,0 +1,419 @@
...
+        definition to check if a similar Wrapper has already been created
+        (see c_support_code() below).
+        """
+        fields_string = ','.join(self.fields).encode('utf-8')
+        types_string = ','.join(str(t) for t in self.types).encode('utf-8')
+        fields_hex = hashlib.md5(fields_string).hexdigest()
+        types_hex = hashlib.md5(types_string).hexdigest()
+        return '_wrapper_%s_%s' % (fields_hex, types_hex)
+
+    def check_that_values_are_compatible(self, data, strict, allow_downcast):
```
nouiz: rename the method. check_ indicate that it will return a bool. It will raise an exception or wrap the data.

theano/gof/tests/test_wrapper.py
```diff
@@ -0,0 +1,205 @@
...
+from theano.tests import unittest_tools as utt
+
+dtype = config.floatX
+tensor_type_0d = TensorType(dtype, tuple())
+scalar_type = Scalar(dtype)
+generic_type = Generic()
+
+
+# A test op to compute `y = a*x^2 + bx + c` for any tensor x, with a, b, c as op params.
+class QuadraticFunction(Op):
```
nouiz: Make another class that use the COp way of implementing an Op. It can do exactly the same thing as here.

theano/gof/wrapper.py
notoraptor: Ok ! I think it's better to keep both classes (to let Wrapper manage the types and Wrap manage the values), but I will modify Wrap so that the constructor takes as argument the Wrapper instance its depends on, and I'll use it in Wrap.__eq__ and Wrap.__hash__ as you suggest.

nouiz:
Good idea to pass the object completly.

notoraptor:
New update,. All @nouiz 's comments have been taken account.

NB: I have modified a little `apply_policy` in `theano/gof/cc.py` because I remarked it expects a list (or tuple) of functions, but sometimes it receives directly a function object, and does not execute it in that case. Now it handles both a function and a list of functions.

Wrapper doc has been added, you can see a screen capture of the rendering here:
[wrapper](https://cloud.githubusercontent.com/assets/3467054/23523936/b1388a52-ff56-11e6-9b71-3efe55d34461.png)

notoraptor:
NB: Other change: now in C code, params are available through a pointer, then we must write `param->attr1` and no more `params.attr1` in C.

doc/library/gof/wrapper.txt
```diff
@@ -0,0 +1,15 @@
+.. _libdoc_gof_wrapper:
+
+=============================================
+:mod:`wrapper` -- Wrapper class for op params
```
abergeron: The module is theano.gof.wrapper

theano/gof/cc.py
```diff
@@ -476,11 +476,12 @@ def apply_policy(policy, r, name, sub):
...
-        for sub_policy in policy:
-            ret += sub_policy(r, name, sub)
-        return ret
+    if not isinstance(policy, (list, tuple)):
+        policy = [policy]
+    ret = ""
+    for sub_policy in policy:
+        ret += sub_policy(r, name, sub)
+    return ret
     return policy(r, name, sub)
```
abergeron: Remove the extra return since it's now dead code.

theano/gof/op.py
```diff
@@ -784,6 +784,21 @@ class Op(utils.object2, PureOp, CLinkerOp):
     Convenience class to bundle `PureOp` and `CLinkerOp`.
 
     """
+
+    # We add a default get_params() implementation which will try to detect params from the op
+    # if params_type is set to a Wrapper. If not, we raise a MethodNotDefined exception.
+    def get_params(self, node):
+        if hasattr(self, 'params_type') and isinstance(self.params_type, theano.gof.wrapper.Wrapper):
+            wrapper = self.params_type
+            if hasattr(self, '__props__') and all(field in self.__props__ for field in wrapper.fields):
```
abergeron: Why do you require that all params be in \_\_props__ here?

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+    In a C code, attributes created to represent an instance of the type associated to ``key`` will be available via:
+
+    .. code-block:: c
+
+        structObject->key;
+        structObject->dtype_key; // e.g. from TensorType C code.
+        structObject->other_attribute_named_from_key;
+        /* etc. */
+
+    **NB**: This Type is not a complete type and should never be used for regular graph operations.
```
abergeron: There is a `.. note::` formatting for those things.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+
+        structObject.key
+
+    In a C code, attributes created to represent an instance of the type associated to ``key`` will be available via:
+
+    .. code-block:: c
+
+        structObject->key;
+        structObject->dtype_key; // e.g. from TensorType C code.
+        structObject->other_attribute_named_from_key;
```
abergeron: This seems to suggest that you can have multiple attributes from a single param.  I don't think this is the case.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+    **NB**: This Type is not a complete type and should never be used for regular graph operations.
+    """
+
+    def __init__(self, **kwargs):
+        if len(kwargs) == 0:
+            raise ValueError('Cannot create Wrapper from empty data.')
+
+        for attribute_name in kwargs:
+            if re.match('^[A-Za-z_][A-Za-z0-9_]*$', attribute_name) is None:
+                raise SyntaxError('Wrapper: attribute "%s" should be a valid identifier.' % attribute_name)
```
abergeron: I don't think SyntaxError is appropriate here.  Probably ValueError instead.  Or a custom exception.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+        self.length = len(kwargs)
+        self.fields = tuple(sorted(kwargs.keys()))
+        self.types = tuple(kwargs[field] for field in self.fields)
+        self.name = self.generate_struct_name()
+
+    def __repr__(self):
+        return 'Wrapper<%s>' % ', '.join([('%s:%s' % (self.fields[i], self.types[i])) for i in range(self.length)])
+
+    def __eq__(self, other):
+        # To be checked.
```
abergeron: This is checked and looks ok :)

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+    # Returns a wrapped object with expected attributes or (in strict mode) checks that data has expected attributes.
+    def filter(self, data, strict=False, allow_downcast=None):
+        if strict and not isinstance(data, Wrap):
+            raise TypeError('%s: strict mode: data should be an instance of Wrap.' % self)
+        return self.wrap_data(data, strict, allow_downcast)
+
+    def values_eq(self, a, b):
+        # We check that a and b have expected attributes and strict values.
+        a = self.filter(a, strict=True)
+        b = self.filter(b, strict=True)
```
abergeron: I'm pretty sure that this is the only type to filter in values_eq.  We probably don't need to do that.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+        # Then we compare.
+        return all(self.types[i].values_eq(getattr(a, self.fields[i]), getattr(b, self.fields[i]))
+                   for i in range(self.length))
+
+    def values_eq_approx(self, a, b):
+        # We check, wrap and round a and b if necessary.
+        a = self.filter(a, strict=False, allow_downcast=True)
+        b = self.filter(b, strict=False, allow_downcast=True)
```
abergeron: Same here, don't filter.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+    def c_compile_args(self, c_compiler):
+        c_compile_args_list = []
+        for _type in self.types:
+            try:
+                try:
+                    c_compile_args_list.extend(_type.c_compile_args(c_compiler))
+                except TypeError:
+                    c_compile_args_list.extend(_type.c_compile_args())
+            except MethodNotDefined:
+                pass
```
abergeron: This code makes me want to find a better way to do this.  But there are currently none.

doc/library/gof/wrapper.txt
notoraptor: I have copied the syntax from `doc/library/gof.type.txt` and from the other files in this folder, in which `:mod:` is always associated to a short name, not a complete name (e.g. from `type.txt`: `:mod:`type` -- Interface for types of variables`).

Should I write `theano.gof.wrapper` anyway ?

theano/gof/cc.py
notoraptor: I was wrong here, the old code did the job, I was confused just because I expected `Wrapper.c_init()` to be run, but there are some cases where `c_init()` is skipped and only `c_extract()` is run, e.g. : https://github.com/Theano/Theano/blob/master/theano/gof/cc.py#L686 .

I don't yet understand all the rules here, but it seems that some C codes to init a Theano type should sometimes better be at top of `c_extract()`, instead of in `c_init()`. That's what I have done for Wrapper class,

I'll put back the old code of `apply_policy()`.

theano/gof/op.py
notoraptor: Based on @nouiz 's comments above, here: https://github.com/Theano/Theano/pull/5612#discussion_r103782532 , and here: https://github.com/Theano/Theano/pull/5612#issuecomment-282417090 .

theano/gof/wrapper.py
notoraptor: I'll use `AttributeError`.

theano/gof/tests/test_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+from theano.gof import Op, COp, Apply
+from theano import Generic
+from theano.scalar import Scalar
+from theano.tensor import TensorType
+from theano.gof.wrapper import Wrapper, Wrap
+from theano import config
+from theano import tensor
+from theano.tests import unittest_tools as utt
+
+dtype = config.floatX
```
nouiz: rename it to floatX to be more clear that this is the dtype for the full test.

theano/gof/tests/test_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+        """ % {'float_type': float_type}
+
+    def c_code(self, node, name, inputs, outputs, sub):
+        X = inputs[0]
+        Y = outputs[0]
+        coeff = sub['params']
+        fail = sub['fail']
+        float_type = node.inputs[0].type.dtype_specs()[1]
+        float_typenum = numpy.dtype(node.inputs[0].type.dtype).num
+        coeff_type = 'npy_' + numpy.dtype(dtype).name
```
nouiz: This is a bad example. You should find the dtype without access to a global variable. You should look in node.params_type

theano/gof/tests/test_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+# Same op as above, but implemented as a COp (with C code in an external file).
+class QuadraticCOpFunc(COp):
+    __props__ = ('a', 'b', 'c')
+    params_type = Wrapper(a=tensor_type_0d,
+                          b=scalar_type,
+                          c=generic_type)
+
+    def get_op_params(self):
+        return [('QUADRATIC_WRAPPER', self.params_type.name),
+                ('COEFF_TYPE', 'npy_' + numpy.dtype(dtype).name)]
```
nouiz: Idem, here you should not use a global variable to get the dtype. Use self.params_type

theano/gof/tests/test_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+scalar_type = Scalar(dtype)
+generic_type = Generic()
+
+
+# A test op to compute `y = a*x^2 + bx + c` for any tensor x, with a, b, c as op params.
+class QuadraticOpFunc(Op):
+    __props__ = ('a', 'b', 'c')
+    params_type = Wrapper(a=tensor_type_0d,
+                          b=scalar_type,
+                          c=generic_type)
```
nouiz: There is something I don't like in this logic. First the dtype is fixed for the class instance. We will have cases where the same op instance will be used in many apply nodes that have different inputs dtypes. We need to cover that cases.

I see those options:

- Have the params_type be created in \__init__. I don't like it much, we it will force the dtype to be know before the init.
- Have the params_type allow to be a callable/class method that will be called with the node and when called, should return a Wrapper instance.

Where do we need to have this wrapper? If we need to create it at only one place, a class method seem good. But if we need to call the method at many places, we will need to include that in the Apply node I think (or accept that it get called many times, not my preference)

theano/gof/tests/test_wrapper.py
```diff
@@ -0,0 +1,245 @@
...
+
+# Same op as above, but implemented as a COp (with C code in an external file).
+class QuadraticCOpFunc(COp):
+    __props__ = ('a', 'b', 'c')
+    params_type = Wrapper(a=tensor_type_0d,
+                          b=scalar_type,
+                          c=generic_type)
+
+    def get_op_params(self):
+        return [('QUADRATIC_WRAPPER', self.params_type.name),
```
notoraptor: About macros, can I suggest to extend the abstract class `Type` (precisely `CLinkerType`) ? I think about a new method:
 * `c_dtype()` which should return the primitive C type of items into a variable handled by the type. For example, with `TensorType(dtype='int64' ...)`, it should return `npy_int64`. For `GpuArrayType(dtype='int32', ...)`, it should return `ga_long`.

Which such a contract, if `c_dtype()` is implemented for tthe type of a param named `key` into Wrapper, I could automatically add a default local macro like `DTYPE_PARAM_key`, which could help when we have static conversions to do into C files for COps. If `c_dtype()` is not implemented (e.g. for `Generic()`), then the macro is not defined (developper will have to deal with types dynamically).

@nouiz @abergeron

doc/library/gof/wrapper.txt
abergeron: Yes.  I know a lot of the doc doesn't do that yet, but it is a problem for linking inside the doc because those statements have meaning for sphinx.

theano/gof/op.py
```diff
@@ -784,6 +784,28 @@ class Op(utils.object2, PureOp, CLinkerOp):
...
+        if hasattr(self, 'params_type'):
+            # If params_type is a Wrapper, we try to extract params from the op.
+            if isinstance(self.params_type, theano.gof.wrapper.Wrapper):
+                wrapper = self.params_type
+                op_has_wrap_attributes = True
+                for field in wrapper.fields:
+                    if not hasattr(self, field):
+                        op_has_wrap_attributes = False
+                        break
+                if op_has_wrap_attributes:
```
abergeron: I don't think so.  Just to give a quick example: context.  It doesn't appear in the props yet it is a param.  Although we can always explicitly define get_params() I guess.

theano/gof/tests/test_wrapper.py
abergeron: I agree with you somewhat, but this is a rather contrived example of params usage.  Usually the parameters should have well-defined types known in advance.

In this case here, they could all be float64 (since they are all functionally scalars), but for a real op those would most likely be inputs since they don't need to be know before execution time and they won't change from one execution of the script to another.

So in essence I wouldn't change the implementation to accommodate the this use case as it is not one we are likely to encounter for real.

theano/gof/tests/test_wrapper.py
abergeron: No, this should access a global since this is the type of the params and those should be fixed.

theano/gof/tests/test_wrapper.py
abergeron: In fact, the test should probably be changed to always use 'float64' for a, b, c.  This would make more sense and be a better example.

theano/gof/tests/test_wrapper.py
abergeron: I approve of this extension.

abergeron: But perhaps with a more explicit name like `c_element_type()`.

theano/gof/cc.py
abergeron: The changed code is functionally equivalent to the old code, so it won't break anything.  However having two return statements one after the other is superfluous.

You can either get rid of the second one or revert the code I guess.

notoraptor: I have reverted the code !

theano/gof/op.py
notoraptor: As `get_params()` has to do the job automatically when Wrapper is used, I think I should better look for params directly into the object (with `hasattr(self, field)`).

So I'll update the code here.

theano/gof/tests/test_wrapper.py
notoraptor: Ok, so I will use `'float64'` instead of `floatX` for this test.

notoraptor:
PR updated.

The extension of `CLinkerType` interface with `c_element_type()` is in the commit https://github.com/Theano/Theano/pull/5612/commits/9f17ce3be85aebe811cd4211dffd4c527e2bc810 . `Scalar`, `TensorType` and `GpuArrayType` have also been updated in this commit to implement the new method. I think it could be good to put it into a separate pull request, but I don't yet know how to do it properly (while allowing the current PR to work properly also!).

Macros `APPLY_SPECIFIC_WRAPPER` and `DTYPE_PARAM_key` are now implemented. Tests are updated (COp example uses these macros). Now, Wrapper attributes are searched into op, not only into props.

@nouiz @abergeron

theano/gof/op.py
```diff
@@ -1381,7 +1381,7 @@ def get_op_params(self):
         The names must be strings that are not a C keyword and the
         values must be strings of literal C representations.
 
-        If op uses a :class:`theano.gof.wrapper.Wrapper` as ``params_type``,
+        If op uses a :class:`theano.gof.Wrapper` as ``params_type``,
```
abergeron: The class name for documentation purposes is still theano.gof.wrapper.Wrapper so don't change that here.

notoraptor: Done !

notoraptor:
Two small updates:
 * In Wrapper C code, I remove some useless calls to `cleanup()` method as it is always called into  wrapper destructor when wrapper is deleted.
 * Wrapper C code cache version now depends on C code cache versions of wrapper Theano types.

@nouiz @abergeron

theano/gof/type.py
```diff
@@ -35,6 +35,19 @@ class CLinkerType(CLinkerObject):
...
+        Optional: Return the name of the primitive C type of items into variables
+        handled by this type.
+
+        e.g:
+
+         - For ``TensorType(dtype='int64', ...)``: should return ``"npy_int64"``.
+         - For ``GpuArrayType(dtype='int32', ...)``: should return ``"ga_int"``.
+
+        """
+        return ''
```
nouiz: why return an empty string and not MethodNotDefined? I think by default, it should be an error is we look for the element type and it isn't provided.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,419 @@
+"""
```
nouiz: I think this doc should be moved into the web doc that explain how to make an op with c code.

notoraptor:
@nouiz About your first comment, I have updated the doc: Wrapper module is still in API doc, and I have added a reference to it in `Using Op params` tutorial. You can see this tutorial rendering here:

http://notoraptor.net/theano/html-make-structs/extending/using_params.html#registering-the-params-with-your-op

Doc rendering for the Wrapper module itself is available here: http://notoraptor.net/theano/html-make-structs/library/gof/wrapper.html

notoraptor:
Doc added for `c_element_type()` in `doc/extending/ctype.txt`: http://notoraptor.net/theano/html-make-structs/extending/ctype.html#CLinkerType.c_element_type .

Doc for this method already appeared in Type module: http://notoraptor.net/theano/html-make-structs/library/gof/type.html#theano.gof.type.CLinkerType.c_element_type .

theano/gof/graph.py
```diff
@@ -125,8 +125,12 @@ def run_params(self):
         Returns the params for the node, or NoParams if no params is set.
 
         """
-        if hasattr(self.op, 'get_params'):
-            return self.op.get_params(self)
+        if hasattr(self.op, 'params_type'):
```
notoraptor: Ok, I didn't know that there was an old interface where get_params() is the only one expected attribute.

I have changed it because I add a default implementation for get_params() (which checks if op has a Wrapper as params_type), so that all ops have now a get_params by default.

I will use this code instead:
```python
def run_params(self):
    try:
        return self.op.get_params(self)
    except MethodNotDefined:
        # the default implementation of get_params() raises a 
        # MethodNotDefined exception if it does not find a Wrapper
        return NoParams
```

notoraptor: Done.

theano/gof/type.py
notoraptor: That was just to avoid using `try-catch` block when looking for `c_element_type()` in an op ^^ ! If it isn't overrided, then the returned string is empty and then evaluated to False.

Is it better to raise an exception ?

notoraptor:
@nouiz :
```
Remove the old interface support and make the c_compiler parameter mandatory? 
I would accept that as we just did a release.
```

I think this will need another PR, as we should also look for all ops that use the old interface, for all methods related (at least: `c_compile_args`, `c_no_compile_args`, `c_headers`, `c_libraries`).

```
should we fix this problem instead and have c_init called on param? 
If people don't use Wrapper, then they will have problems.
```
Maybe a separate PR here also ? More concretely, after first investigations, the problem seems to come from here: https://github.com/Theano/Theano/blob/master/theano/gof/cc.py#L682 .

theano/gof/wrapper.py
```diff
@@ -0,0 +1,439 @@
...
+    def c_declare(self, name, sub, check_input=True):
+        struct_name = self.name
+        return """
+        %(struct_name)s* %(name)s;
+        """ % locals()
+
+    def c_init(self, name, sub):
+        # NB: It seems c_init() is not called for an op param.
+        # So the real initialization is done at top of c_extract.
+        return """
```
abergeron: We don't need to call c_init() since we never want "empty" values.  params are only ever used for extra "inputs" so we call c_extract() instead of c_init().

theano/gof/graph.py
abergeron: We never supported defining only get_params.  If you had get_params defined, the rest of the code assumed that you also had params_type and this was documented.  Please don't introduce support for that.

theano/gof/type.py
abergeron: I think the exception is safer.  It is more annoying to use for us developers, but it is still probably a better idea.

abergeron:
I'm not sure about making c_compiler mandatory.  Maybe we just want to remove it since the old backend is going away and the NVCC_compiler with it.

However if we still plan on adding support for other compilers (like MSVC, icc), then we should keep it and in that case making it mandatory is a good idea.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,449 @@
...
+        %(nams)s = NULL;
```
abergeron: %(nams)s -> %(name)s

theano/gof/wrapper.py
abergeron: One solution to all this would be to declare the struct directly rather than as a pointer.  Then it is always default-initialized on declaration and we don't need c_init().

notoraptor: Yes, that was my first approach (declare struct on stack) ! But I changed it to pointers because of 2 considerations:
1) If struct is an object rather than a pointer, then it is copied when passed to APPLY_SPECIFIC main function for COps. We should then expect that wrapper objects are not too big so that they are copied faster enough. But I think it is not so problematic, finally.
2) The original wrapper and the copy passed to APPLY_SPECIFIC main function in COps will share pointers values (like PyObject* pointers). Then, at the end of APPLY_SPECIFIC function, the destructor will be called on the wrapper copy, that will call the cleanup function, that will clean the PyObjects and other memory data. Then, we must be sure that the original wrapper will not be reused after the main function is called, as its pointers should have been cleaned (e.g. PyObject pointers into wrapper would have been decremented and set to NULL, etc.). For this point, are we sure ? If yes, I will go back to structs on stack (which is indeed better, in my opinion !).

theano/gof/graph.py
notoraptor: Ok. So, now that I have added a default get_params() implementation, I should better check here if the op also has params_type, right ?

notoraptor:
Updated. `c_element_type` now raises a MethodNotDefined exception. Typos are fixed in `wrapper.py` and `test_wrapper.py`. I have performed a rebase to check if it can fix a strange TRAVIS error which does not seem to be related to this PR.

I wait for @abergeron precisions about c_init and run_params to add additiional updates (cf. my comments above: https://github.com/Theano/Theano/pull/5612#discussion_r108419664 and https://github.com/Theano/Theano/pull/5612#discussion_r108420729 ).

notoraptor:
Still a strange error for this PR in TRAVIS with `theano.tensor.tests.test_gc.test_gc_never_pickles_temporaries`: https://travis-ci.org/Theano/Theano/jobs/215952499 :

```
FAIL: theano.tensor.tests.test_gc.test_gc_never_pickles_temporaries
  File "/home/travis/miniconda2/envs/pyenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTes
    self.test(*self.arg)
  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_gc.py", line 92, in test_gc_never_pickles_temporaries
    assert len_pre_f == len_post_f, (len_pre_f, len_post_f)
AssertionError: (22766, 22761)
```

abergeron:
This *may* be transient, so I restarted the affected travis test.  If it fails again, we'll have to investigate.

notoraptor:
Error still occurs. I'm investigating, but it is very strange. For example, my other PR for GpuDnnSoftmax ( PR https://github.com/Theano/Theano/pull/5755 ,  which use Wrapper from this current PR) passes all checks (both jenkins and travis).

abergeron:
I am also investigating, but I can't reproduce locally on any machine.

notoraptor:
Same here (on Windows), `theano/tensor/tests/test_gc.py` does not fail.

notoraptor:
Error has disappeared, still strange.

I'm going to remove the printing in test_gc and let tests run for a last time.

notoraptor:
No more error. Incomprehensible, but fantastic.

notoraptor:
Debug commits removed. I have re-launched travis.

abergeron:
@nouiz You still have requested changes.  Have they been addressed with the new commits?  On my end I'm ready to merge.

theano/gof/wrapper.py
```diff
@@ -0,0 +1,443 @@
...
+        if not isinstance(wrapper, Wrapper):
+            raise TypeError('Wrap: 1st constructor argument should be a Wrapper.')
+        for field in wrapper.fields:
+            if field not in kwargs:
+                raise TypeError('Wrap: Wrapper attribute "%s" not in Wrap args.' % field)
+        super(Wrap, self).__init__(**kwargs)
+        self.__dict__.update(wrapper=wrapper)
+
+    def __repr__(self):
+        return 'Wrap(%s)' % ', '.join([('%s:%s' % (k, type(self[k]))) for k in sorted(self.keys())])
```
notoraptor: Ah ok ! I wanted to avoid a too long printing when a Wrap contain big numpy ndarrays, for example. Also, I would have to print the associated Wrapper, too.

Can I use `__str__()` instead, and just remove `__repr__()` implementation ?

theano/gof/wrapper.py
```diff
@@ -0,0 +1,443 @@
...
+            # NB: Wrapped data should have been already filtered.
+            self.wrapper.types[i].values_eq(self[self.wrapper.fields[i]], other[self.wrapper.fields[i]])
+            for i in range(self.wrapper.length)
+        ))
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
+class Wrapper(Type):
```
notoraptor: About this, I was thinking to `ParamsType` (with `s` at end of `Params`) as a new name, as it should become the main class to use to provide multiple parameters for an Op !

What do you prefer ?

notoraptor:
Updated. @nouiz I have made 2 comments above.

theano/gof/wrapper.py
```diff
@@ -101,7 +101,7 @@ def __init__(self, wrapper, **kwargs):
             if field not in kwargs:
                 raise TypeError('Wrap: Wrapper attribute "%s" not in Wrap args.' % field)
         super(Wrap, self).__init__(**kwargs)
-        self.__dict__.update(wrapper=wrapper)
+        self.__dict__.update(wrapper=wrapper, signatures=None)
```
nouiz: Use \__signatures__ to lower the probability of having that name used by the user.

if there is a colision, there will be an error already.

theano/gof/wrapper.py
nouiz: even better!

theano/gof/wrapper.py
nouiz: big numpy ndarray should not be used frequently. So I don't think this is a problem. Also numpy already cap the output size (probably too high limit then what we want, but as it should not happen, it should be good enough)

notoraptor:
@nouiz Updated with important renaming. Wrap class method `__repr__()` updated and private fields more protected.

**Renaming**:
 * class Wrap -> Params
 * class Wrapper -> ParamsType
 * file wrapper.py -> params_type.py
 * file test_wrapper.py --> test_params_type.py
 * macro definition for Wrapper class name APPLY_SPECIFIC_WRAPPER -> PARAMS_TYPE

Renaming also done in documentation. Doc rendering available here: http://notoraptor.net/theano/html-make-structs/extending/using_params.html#registering-the-params-with-your-op

doc/library/gof/params_type.txt
```diff
@@ -0,0 +1,15 @@
+.. _libdoc_gof_wrapper:
```
nouiz: updates

notoraptor:
@nouiz Updated.

PR #5558, opened 2017-02-16
Fix compilation crash show by THEANO_FLAGS='floatX=float32,dnn.conv.a…

nouiz:
…lgo_bwd_data=time_once,dnn.conv.algo_bwd_filter=time_once' theano-nose theano/sandbox/cuda/tests/test_dnn.py:test_dnn_conv_grad. fix gh-5517

PR #5614, opened 2017-02-23
Fix dnn conv grad issues

lamblin:
follow-up on #5558.


lamblin:
Just running the following on a Titan Black triggered the issue:
`THEANO_FLAGS=floatX=float32,dnn.conv.algo_fwd=time_once,dnn.conv.algo_bwd_data=time_once,dnn.conv.algo_bwd_filter=time_once,init_gpu_device=cuda theano-nose theano/gpuarray/tests/test_dnn.py`
I did not check which algo was selected by cuDNN, but it was not one of the ones we expose in Theano.

PR #5616, opened 2017-02-24
Explicitly specify the output ndim in reshape

### LINKED ISSUE #5613: Problem with neibs2images
Hi,
I am trying to use the images2neibs and neibs2images functions, but the neibs2images seems not to be working. My theano is updated. Here is the simplest example I've tryied

```python
import numpy as np
import theano

imagesP = T.tensor4('imagesP')
patches = T.nnet.neighbours.images2neibs(imagesP, [16,16])
extractPatches = theano.function([imagesP], patches,on_unused_input='ignore')

patsRecovery = T.matrix('patsRecovery')
original_size  = T.vector('original_size')
reconstructed_imgs = T.nnet.neighbours.neibs2images(patsRecovery, (16,16),original_size)
recoverImg = theano.function([patsRecovery,original_size],reconstructed_imgs,on_unused_input='ignore')

im_val = np.ones((1,3, 320, 320),dtype=np.float32)
neibs = extractPatches(im_val)    #shape = (1200,256)
recovered_img = recoverImg(neibs ,im_val.shape)

```

While the "images2neibs" works fine, the other one keep getting the following error "ValueError: ('You cannot drop a non-broadcastable dimension.', ((False, False, False, False), ('x', 'x', 0, 1)))"
Can someone help me?

lamblin:
The whole code expects the shape to be of length 4, and the output to be
4D already.

Fixes #5613.

### DIFF
--- a/theano/tensor/nnet/neighbours.py
+++ b/theano/tensor/nnet/neighbours.py
@@ -633,9 +633,16 @@ def neibs2images(neibs, neib_shape, original_shape, mode='valid'):
                              new_neib_shape, mode=mode)
 
     if mode == 'ignore_borders':
-        valid_shape = list(original_shape)
-        valid_shape[2] = (valid_shape[2] // neib_shape[0]) * neib_shape[0]
-        valid_shape[3] = (valid_shape[3] // neib_shape[1]) * neib_shape[1]
+        # We use set_subtensor to accept original_shape we can't infer
+        # the shape and still raise error when it don't have the right
+        # shape.
+        valid_shape = original_shape
+        valid_shape = T.set_subtensor(
+            valid_shape[2],
+            (valid_shape[2] // neib_shape[0]) * neib_shape[0])
+        valid_shape = T.set_subtensor(
+            valid_shape[3],
+            (valid_shape[3] // neib_shape[1]) * neib_shape[1])
         output_4d = output_2d.reshape(valid_shape, ndim=4)
         # padding the borders with zeros
         for d in [2, 3]:
--- a/theano/tensor/nnet/neighbours.py
+++ b/theano/tensor/nnet/neighbours.py
@@ -620,7 +620,7 @@ def neibs2images(neibs, neib_shape, original_shape, mode='valid'):
         valid_shape = list(original_shape)
         valid_shape[2] = (valid_shape[2] // neib_shape[0]) * neib_shape[0]
         valid_shape[3] = (valid_shape[3] // neib_shape[1]) * neib_shape[1]
-        output_4d = output_2d.reshape(valid_shape)
+        output_4d = output_2d.reshape(valid_shape, ndim=4)
         # padding the borders with zeros
         for d in [2, 3]:
             pad_shape = list(output_4d.shape)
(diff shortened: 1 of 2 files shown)


theano/tensor/nnet/neighbours.py
```diff
@@ -620,7 +620,7 @@ def neibs2images(neibs, neib_shape, original_shape, mode='valid'):
         valid_shape = list(original_shape)
         valid_shape[2] = (valid_shape[2] // neib_shape[0]) * neib_shape[0]
         valid_shape[3] = (valid_shape[3] // neib_shape[1]) * neib_shape[1]
-        output_4d = output_2d.reshape(valid_shape)
+        output_4d = output_2d.reshape(valid_shape, ndim=4)
```
nouiz: list(original_shape) will fail if it can't determine the shape length.

theano/tensor/nnet/neighbours.py
```diff
@@ -629,7 +629,7 @@ def neibs2images(neibs, neib_shape, original_shape, mode='valid'):
     elif mode == 'valid':
         # TODO: we do not implement all mode with this code.
         # Add a check for the good cases.
-        output_4d = output_2d.reshape(original_shape)
+        output_4d = output_2d.reshape(original_shape, ndim=4)
```
nouiz: ok. If the shape is't 4, there should be an error at run time. This is fine.

theano/tensor/nnet/neighbours.py
lamblin: Right. I'll fix that.

theano/tensor/nnet/neighbours.py
```diff
@@ -317,8 +317,24 @@ def c_code(self, node, name, inp, out, sub):
...
         // (step_x,step_y) = neib_step
-        const npy_intp step_x = (npy_intp) *(dtype_%(neib_step)s*) PyArray_GETPTR1(%(neib_step)s, 0);
-        const npy_intp step_y = (npy_intp) *(dtype_%(neib_step)s*) PyArray_GETPTR1(%(neib_step)s, 1);
+        const dtype_%(neib_step)s step_x = *(dtype_%(neib_step)s*) PyArray_GETPTR1(%(neib_step)s, 0);
+        const dtype_%(neib_step)s step_y = *(dtype_%(neib_step)s*) PyArray_GETPTR1(%(neib_step)s, 1);
+
+        if (step_x <=0 || step_y <=0)
+        {
+            PyErr_Format(PyExc_ValueError,
+                         "neib_step wrong step ; values <= 0. Got %%d %%d.",
```
abergeron: Don't use %d for unknown data type, use %lld and cast to long long.

theano/tensor/nnet/neighbours.py
```diff
@@ -317,8 +317,24 @@ def c_code(self, node, name, inp, out, sub):
...
+                         step_x, step_y);
+            %(fail)s;
+        }
+
+        if (c <=0 || d <=0)
+        {
+            PyErr_Format(PyExc_ValueError,
+                         "neib_shape values <= 0. Got %%d %%d.",
```
abergeron: Don't use %d for npy_intp.  Either use the platform-appropriate size_t modifier or %lld and cast to long long.

nouiz:
I pushed fixes to this PR that fix that.

abergeron:
They aren't on github

nouiz:
I did the fixes

PR #5629, opened 2017-02-27
Prepare release 0.9.0rc2

notoraptor:
@nouiz @lamblin

### DIFF
--- a/NEWS_DEV.txt
+++ b/NEWS_DEV.txt
@@ -105,7 +105,7 @@ Others:
 
 
 Other more detailed changes:
- - Use of 64-bit indices in sparse.AddSD where necessary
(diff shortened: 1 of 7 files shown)


NEWS_DEV.txt
```diff
@@ -104,6 +105,7 @@ Others:
 
 
 Other more detailed changes:
+ - Use of 64-bit indices in sparse.AddSD where necessary
```
nouiz: This is the title of the PR, but there was more stuff in it. I think this would better describe it.
- Use of 64-bit indexing in sparse ops to allow matrix with more then 2**31-1 elements.

notoraptor: Done

PR #5631, opened 2017-02-28
Faster compilation and import with cuda backend

f0k:
This partly solves the problem identified in https://github.com/Theano/Theano/issues/1681#issuecomment-283022286 of GPU Ops taking much longer to compile than CPU Ops.

The problem is that a `mod.cu` file goes through `cudafe` and `cudafe++`, which takes a whopping 1.5 seconds of compile time even for an empty file. Many GPU Ops don't actually need that: If an Op doesn't require compiling any kernels, there's no point in calling `cudafe`. The simple solution is to rename such files to `mod.cpp` instead, and manually add the `cuda.h` and `cuda_runtime_api.h` headers.

This PR is a proof of concept. I'm not sure the distinction code catches all cases, but it works for me.
On my machine, `import theano` takes 2.5s before this PR, and 0.7s after (with an existing compile directory, not from scratch).
Compiling the Lasagne MNIST example with the CNN architecture from an empty compile directory takes 2m45s before this PR, and 1m38s after. Before this PR, it produces 34 `mod.cpp` files and 70 `mod.cu` files. After this PR, it produces 71 `mod.cpp` files and 33 `mod.cu` files.

Now what do we do with this? I know the backend is meant to be dropped any time soon (before or after releasing 0.9?), so it's probably more of a "why didn't we do this 5 years ago" suggestion, but maybe there is a chance to backport this to Theano 0.8.2 and have a final maintenance release with it, for people who need to use the old backend for some reason? Finally, is any of this relevant for the new backend? On my machine, it takes even longer to import than the old backend, but I don't see any nvcc process involved.

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -239,6 +239,16 @@ AddConfigVar('nvcc.fastmath',
              # if theano.sandbox.cuda is loaded or not.
              in_c_key=False)
 
+AddConfigVar('nvcc.cudafe',
+             "If 'always' (the default), cudafe will be called for every GPU"
+             " Op compilation. If 'heuristic', it will only be called if the"
+             " source code appears to contain CUDA code. This can speed up"
+             " compilation and importing theano, but might fail to compile"
+             " some custom GPU Ops.",
+             EnumStr('always', 'heuristic'),
+             # Not needed in c key, does not affect the compilation result.
+             in_c_key=False)
+
 AddConfigVar('gpuarray.sync',
              """If True, every op will make sure its work is done before
                 returning.  Setting this to True will slow down execution,
--- a/theano/sandbox/cuda/nvcc_compiler.py
+++ b/theano/sandbox/cuda/nvcc_compiler.py
@@ -230,7 +230,19 @@ class NVCC_compiler(Compiler):
             if python_lib not in lib_dirs:
                 lib_dirs.append(python_lib)
 
-        cppfilename = os.path.join(location, 'mod.cu')
+        if (config.nvcc.cudafe == 'heuristic' and not
+            any(marker in src_code for marker in ("__global__", "__device__",
+                                                  "__host__", "<<<",
+                                                  "nvmatrix.cuh"))):
+            # only calls existing CUDA functions, can compile much faster
+            cppfilename = os.path.join(location, 'mod.cpp')
+            src_code = ("#include <cuda.h>\n"
+                        "#include <cuda_runtime_api.h>\n" +
+                        src_code)
+        else:
+            # contains CUDA host code or device functions, needs .cu extension
+            cppfilename = os.path.join(location, 'mod.cu')
+
         with open(cppfilename, 'w') as cppfile:
 
             _logger.debug('Writing module C++ code to %s', cppfilename)


nouiz:
I like this PR, but it is not relevent for the new back-end. We won't back port it to 0.8. Maybe we could include it for the rc3 we plan to do next Monday? I'm not sure I want to do this, in case it break things. I don't want to spend time debugging problems in the old back-end. But if @abergeron is pretty sure this is good, I'm not opposed to it.

the new back-end compile the GPU code with the nvidia driver itself. So it should be faster. The current regression in speed is that we don't cache the compilation of the nvidia driver, only of g++. There is a PR in libgpuarray about that, but it will be done after the 0.9 release.

abergeron:
If we weren't about to delete the old backend I would say yes to this.  However I really don't want things to break and have to chase down why, even is this seems really safe.

nouiz:
I put the 0.9 milestone to talk of this with @lamblin when he isn't sick anymore, but it isn't sure we will merge it.

f0k:
> However I really don't want things to break and have to chase down why

Yes, I completely understand. It's not 100% sure this will cover each and every custom Op. It's a pity nobody investigated this earlier...

Would it be an acceptable compromise to add this as a configuration option that will disappear together with the old backend? E.g, `theano.config.nvcc.cudafe` with a default of `'always'` and the other option being `'heuristic'`, with a note that this is experimental? Not sure how long the old backend will still be used by people, but it'd just be more convenient to tell colleagues/users to turn on a particular option for faster compilation than to tell them to apply a particular patch, or to install from a branch in my Theano clone.
Let me know if you could live with that and I'll update the PR!

f0k:
> On my machine, `import theano` takes 2.5s before this PR, and 0.7s after (with an existing compile directory, not from scratch).
Compiling the Lasagne MNIST example with the CNN architecture from an empty compile directory takes 2m45s before this PR, and 1m38s after.

These timings used CUDA 7.5. With CUDA 8.0, `import theano` takes 1.4s without, and 0.8s with this PR. Compiling the CNN example takes 1m26s without, and 1m1s with this PR (still 30% faster).

lamblin:
With the flag, I'd say OK.
0.9 is going to be around for a while for people using the old back-end, even if we don't actively maintain it, so it could be useful.

f0k:
> 0.9 is going to be around for a while for people using the old back-end, even if we don't actively maintain it, so it could be useful.

Unfortunately I found a problem in the meantime -- the heuristics won't work for pylearn2's cuda-convnet wrapper the way they are. The Ops for that do not include any of the markers in their source code, but they `#include` a `.cuh` file that does. If we advertise this as a heuristic and not enable it by default, I can just add the name of the file to the list of markers and it will work. I can update the PR accordingly tomorrow.
Taking this more seriously would require parsing the source code for `#include` directives and reading any `.cuh` files to see if they contain CUDA code (except for the `cuda_ndarray.cuh` file because we know it doesn't). This seems like a lot of effort considering the backend will be phased out.

f0k:
> I can update the PR accordingly tomorrow.

Did it right away instead.

nouiz:
jenkins test this

PR #5634, opened 2017-02-28
Warn if taking softmax over broadcastable dimension

### LINKED ISSUE #5501: Warn if taking softmax over broadcastable dimension
At least one user inadvertently had a softmax applied to a network layer with a single output unit. The softmax Op could issue a warning if the axis to be softmaxed is broadcastable (i.e., guaranteed to only have a single element), because usually you wouldn't want this. (Otherwise, Theano could also use an optimizer replacing softmax over a broadcastable dimension with `T.ones_like()` ;) ).

bordesf:
fix #5501

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -802,6 +802,9 @@ def softmax_graph(c):
 
 
 def softmax(c):
+    c = as_tensor_variable(c)
+    if c.broadcastable[-1]:
+        warnings.warn("The softmax is appllied on a dimension of shape 1, which doesn't have a semantic meaning.")
     return softmax_op(c)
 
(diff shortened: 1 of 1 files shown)


theano/tensor/nnet/nnet.py
```diff
@@ -802,6 +803,9 @@ def softmax_graph(c):
 
 
 def softmax(c):
+    c = as_tensor_variable(c)
+    if c.broadcastable[-1]:
+        warnings.warn("The softmax is appllied on a dimension of shape 1, which doesn't have a semantic meaning.")
```
lamblin: "applied" with one "l"
"does not" is better than "doesn't" in error messages in general

PR #5635, opened 2017-02-28
[ENH] for rc3, Graph clean up, faster compilation, conv.assert_shape Theano flags, add unsafe in useless opt

nouiz:
I think we should include this in our rc3

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -127,6 +127,13 @@ AddConfigVar(
     in_c_key=False)
 
 AddConfigVar(
+    'conv.assert_shape',
+    "If False, AbstractConv* ops won't add assert that verify that"
+    " the user provided shapes are also the one at run time",
+    BoolParam(True),
+    in_c_key=False)
+
+AddConfigVar(
     'print_global_stats',
     "Print some global statistics (time spent) at the end",
     BoolParam(False),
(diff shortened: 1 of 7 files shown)


theano/configdefaults.py
```diff
@@ -127,6 +127,13 @@ def __str__(self):
     in_c_key=False)
 
 AddConfigVar(
+    'conv.assert_shape',
+    "If False, AbstractConv* ops won't add assert that verify that"
+    " the user provided shapes are also the one at run time",
+    BoolParam(True),
```
abergeron: I would default this to False, since it seems like a debugging option.

nouiz:
I agree with @abergeron. If @lamblin is OK with that, I'll include that in
rc3.

nouiz:
closing and reopening to trigger travis.

theano/gof/opt.py
```diff
@@ -607,11 +607,20 @@ def process_node(self, fgraph, node):
         # properly.
         # The clients should at least contain `node` itself!
         if node.inputs:
-            assert len(node.inputs[0].clients) > 0
-            assert (node, 0) in node.inputs[0].clients
+            # Take the smallest clients list. Some ops like elemwise
+            # have optimization that put constant as the first inputs.
+            # As constant have in general more clients then other type of nodes
```
Sentient07: Sorry for a late comment, a small correction, here it should be "than" and not "then".

nouiz: done thanks.

PR #5636, opened 2017-02-28
Fix overflow in pooling

abergeron:


aam-at:
[MaxPoolRop](https://github.com/Theano/Theano/blob/master/theano/gpuarray/pool_max_rop.c#L109) should be updated also.

theano/gpuarray/tests/test_pool.py
```diff
@@ -47,6 +47,19 @@ def test_pool_c_interface(self):
             f = theano.function([], ds_op(inp, [2, 2], pad=pad), mode=gpu_mode)
             f()
 
+    def test_pool_big_ws(self):
+        gpu_mode = copy.copy(mode_with_gpu).excluding("cudnn")
```
nouiz: why copy.copy()? Normlally just doing mode_with_gpu.excluding('cudnn') should do the work.

abergeron: I copy pasted the test above (and its mistakes apparently).

aam-at: I also copied (from cuda tests) and I thought that changing `check_py_code` has something to do with `copy`.

Issue #5426, opened 2017-01-18
Change grad() method to L_op in ops that need the outputs for gradient


nouiz:
Do as https://github.com/Theano/Theano/pull/5420/commits/0f5a06d7f7bd8480d52d1d2b754bffd7abd14f21 for other Ops that need it.

At least, those ops need to be modified for this:
- [ ] tensor/signal/pool.py
- [ ]  Elemwise
- [ ] All scalar op under scalar/basic.py,
   - [ ] scalar/basic_scipy.py
   - [ ] theano/tensor/nnet/nnet.py
   - [ ] theano/tensor/nnet/sigm.py
   - [ ] theano/tensor/xlogx.py

Is there other ops? OpFromGraph? Should we look at all ops in Theano?

wazeerzulfikar:
Hi @nouiz , I would like to work on this issue. Do I need to make a similar change as the one in 0f5a06d , in the ops listed below?

nouiz:
That is exact. Start at the top of the file. They are easier to do.

wazeerzulfikar:
@nouiz , Should all the grad() functions be changed to L_op in these ops, or only the ones which are similar to the one in scan_op.py?

nouiz:
For now, only the grad when we recreate the outputs like in the commit.

We didn't discuss changing all of them. I see a good reason for this, but
before doing such change, we should have more then just me think it is
good. @abergeron @lamblin, what do you think?

abergeron:
Eventually I think we should change all of them.  But for now the biggest gain is the ones where we recreate the outputs.

abergeron:
Not all the changes were made.

affanv14:
Is anyone working on this?
or can i work on it?

nouiz:
There isn't news from other people, so I think you can work on this. thanks.

affanv14:
@nouiz in scalar/basic.py ,should i just change grad to L_op for classes which 
inherit from BinaryScalarOp only?

because if i change for classes which inherit from ScalarOp or UnaryScalarOp i get errors like this:
MethodNotDefined: ('grad', <class 'theano.scalar.basic.Abs'>, 'Abs')

nouiz:
The file scalar/basic.py is more complicated. It request changes at
different files and I can't name all of them.

You should modify the Op ScalarOp and add a method grad like:

    def L_op(self, inputs, outputs, output_grads):
        return self.grad(inputs, output_grads)

[...]

PR #5641, opened 2017-03-02
L_op for OpFromGraph

khaotik:
#5426 

- Interface is backward compatible
- added a test case

### DIFF
--- a/theano/compile/builders.py
+++ b/theano/compile/builders.py
@@ -47,7 +47,8 @@ class OpFromGraph(gof.Op):
 
         OpFromGraph instance : Override with another OpFromGraph, should
         accept inputs as the same order and types of "inputs" and "output_grads"
-        arguments as one would specify in grad() method.
+        arguments as one would specify in grad() method. This argument is mutually
+        exclusive to lop_overrides.
 
         callable : similar to OpFromGraph instance, must return list of
         :class:`Variable <theano.gof.Variable>`.
@@ -59,6 +60,13 @@ class OpFromGraph(gof.Op):
         list: Each OpFromGraph/callable must return a single
         :class:`Variable <theano.gof.Variable>`. Each list element corresponds to gradient of
         a specific input, length of list must be equal to number of inputs.
+
+    lop_overrides : single or list of {'default', OpFromGraph, callable, Variable with special type}, optional
+        Defaults to ``'default'``.
+
+        Similar to ``grad_overrides``, except callables should accept inputs as the
+        same order and types of "inputs", "outputs", and "output_grads" in L_op. This
+        argument is mutually exclusive with grad_overrides.
 
     rop_overrides : single or list of {'default', OpFromGraph, callable, Variable with special type}, optional
         Defaults to ``default``.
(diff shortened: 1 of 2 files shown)


theano/compile/builders.py
```diff
@@ -251,7 +284,18 @@ def __init__(
...
+            raise ValueError('lop_overrides and grad_overrides are mutually exclusive')
+        elif cond == 1:
+            self.set_lop_overrides(lop_overrides)
+            self._lop_type = 'lop'
+        elif cond == 2:
+            self.set_lop_overrides(grad_overrides)
+            self._lop_type = 'grad'
+        else:
+            self.set_lop_overrides('default')
+            self._lop_type = 'lop'
```
abergeron: This trickery is confusing to read (even if correct).  Please use normal if/else for this.

theano/compile/builders.py
```diff
@@ -271,21 +315,40 @@ def __str__(self):
...
+            if self._lop_op_is_cached:
+                return
+            if self._lop_type == 'grad':
+                needed_ninps = inp_len + len(local_outputs)
+                ninps = len(lop_op.local_inputs)
+                if needed_ninps != ninps:
+                    raise ValueError(
+                        self.OV_INP_LEN_ERR_MSG % (needed_ninps, ninps))
+                # make a wrapper callable
+                lop_op = lambda inps, grads: self._lop_op(*(inps + grads))  # flake8: noqa
```
abergeron: You can use
```
def lop_op(inps, grads):
    return self._lop_op(*(inps  + grads))
```
 here instead of disabling flake8.

theano/compile/builders.py
```diff
@@ -271,21 +315,40 @@ def __str__(self):
...
+                ninps = len(lop_op.local_inputs)
+                if needed_ninps != ninps:
+                    raise ValueError(
+                        self.OV_INP_LEN_ERR_MSG % (needed_ninps, ninps))
+                self._lop_op_is_cached = True
+                self._lop_op_stypes_l = [None] * inp_len
+                self._lop_op.kwargs['on_unused_input'] = 'ignore'
+                return
+            else:
+                raise ValueError(self.LOP_TYPE_ERR_MSG % self._lop_type)
```
abergeron: This should be an assert, not an exception since `_lop_type` is not user-settable and therefore should never be wrong.

theano/compile/builders.py
```diff
@@ -296,26 +359,28 @@ def _recompute_grad_op(self):
...
-        STYPE_ERR_MSG = ('Overriding Variable instance can only have type'
-                         ' of DisconnectedType or NullType, got %s')
-        # we need to convert _grad_op into an OfG instance
-        if grad_op == 'default':
+        if self._lop_type == 'lop':
+            callable_args = (local_inputs, local_outputs, output_grads)
+        elif self._lop_type == 'grad':
+            callable_args = (local_inputs, output_grads)
+        else:
+            raise ValueError(self.LOP_TYPE_ERR_MSG % self._lop_type)
```
abergeron: Assertion here too.

khaotik:
@nouiz Just to note this PR is ready for review.

nouiz:
Quick note, no time for a full review now, but this PR will need a release. Someone in gh-6378 would need this.

khaotik:
rebased

abergeron:
jenkins test this please

Issue #5642, opened 2017-03-02
Change the default of Theano flag conv.assert_shape to False


nouiz:
It cause significant slow down. In one case, compilation time go from 20m to around 1m!

Should we create an AssertShape op to have this in a very small graph? I think until we get this don't slow down compilation as much, we should remove it by default. We could add it back when we have a fast implementation.

nouiz:
done

PR #5643, opened 2017-03-02
Change the default of conv.assert_shape flags to speed up compilation time

nouiz:
This happen in corner cases with scan from 20m to 1m.

@f0k How bad you think that we don't assert user shape by default? @abergeron point of view is that this is more for debugging purpose. We won't generate bad results. It would just put back the user shape as compiler hint as it was done in some of the ops (while others was enforcing it). We will be consistent between all ops.

see gh-5642

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -130,7 +130,7 @@ AddConfigVar(
     'conv.assert_shape',
     "If False, AbstractConv* ops won't add assert that verify that"
     " the user provided shapes are also the one at run time",
-    BoolParam(True),
+    BoolParam(False),
     in_c_key=False)
 
 AddConfigVar(
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -128,8 +128,9 @@ AddConfigVar(
 
 AddConfigVar(
     'conv.assert_shape',
-    "If False, AbstractConv* ops won't add assert that verify that"
-    " the user provided shapes are also the one at run time",
+    "If True, AbstractConv* ops will verify that user-provided"
+    " shapes match the runtime shapes (debugging option,"
+    " may slow down compilation)",
     BoolParam(False),
     in_c_key=False)
 
--- a/theano/configparser.py
+++ b/theano/configparser.py
@@ -98,15 +98,17 @@ class change_flags(object):
 
     Useful during tests.
     """
-    def __init__(self, **kwargs):
+    def __init__(self, *args, **kwargs):
         confs = dict()
-        for k in kwargs:
+        args = dict(*args)
+        args.update(kwargs)
+        for k in args:
             l = [v for v in theano.configparser._config_var_list
                  if v.fullname == k]
             assert len(l) == 1
             confs[k] = l[0]
         self.confs = confs
-        self.new_vals = kwargs
+        self.new_vals = args
 
     def __call__(self, f):
         @wraps(f)
--- a/theano/configparser.py
+++ b/theano/configparser.py
@@ -98,9 +98,9 @@ class change_flags(object):
 
     Useful during tests.
     """
-    def __init__(self, *args, **kwargs):
+    def __init__(self, args=[], **kwargs):
         confs = dict()
-        args = dict(*args)
+        args = dict(args)
         args.update(kwargs)
         for k in args:
             l = [v for v in theano.configparser._config_var_list
--- a/theano/configparser.py
+++ b/theano/configparser.py
@@ -98,7 +98,7 @@ class change_flags(object):
 
     Useful during tests.
     """
-    def __init__(self, args=[], **kwargs):
+    def __init__(self, args=(), **kwargs):
         confs = dict()
         args = dict(args)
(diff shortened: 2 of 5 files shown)


nouiz:
I should add, that my personal point of view is that as it cause big inconvenience, even in "corner" cases, we should not enable it by default.

Why does it cause such big slow down with scan? It is due in big part to the order of optimization. We push out nodes from the scan on the not optimized graph, that contain many duplicated nodes that aren't yet merged. We could pre-optimize the inner graph before the push out, but this is much more complicated and we won't do that so close to a release.

See also: https://github.com/Theano/Theano/pull/5635/commits/c52b93961e4143af356800f1ade0ca3efb408a00 and https://github.com/Theano/Theano/pull/5635/commits/5c2d16704ab31e5ba71e0f8a2db7427c4519a379

f0k:
> How bad you think that we don't assert user shape by default?

I don't think this is important. Most convolution implementations don't require knowing the shape in advance anyway. +1 for reducing compilation time (not only for scan, also for regular convnets).

IIRC, @gvtulder also added assertions for invalid shape combinations (e.g., shapes that would lead to negative output sizes). Are those still there? Or were they implemented as part of the convolution implementations, not as assertions?

theano/configdefaults.py
```diff
@@ -130,7 +130,7 @@ def __str__(self):
     'conv.assert_shape',
     "If False, AbstractConv* ops won't add assert that verify that"
     " the user provided shapes are also the one at run time",
-    BoolParam(True),
+    BoolParam(False),
```
f0k: Maybe also update the documentation to something like "If True, AbstractConv* ops will verify that user-provided compile time shapes match the runtime shapes (debugging option, may slow down compilation)".

nouiz:
I updated the doc. I removed "compile time" as it isn't at compilation
time, but when building the graph. I don't think making the disctinction in
which phase is useful, so simpler seem better.

There is other assert for negative shapes. They are still there.

f0k:
> I removed "compile time" as it isn't at compilation time, but when building the graph.

Well yes, but what you give on building the graph is what's there at compile time, and what will end up in the generated code. So I usually just distinguish compile-time from runtime. Anyway, simpler is better!

theano/configparser.py
```diff
@@ -98,15 +98,17 @@ class change_flags(object):
 
     Useful during tests.
     """
-    def __init__(self, **kwargs):
+    def __init__(self, *args, **kwargs):
         confs = dict()
-        for k in kwargs:
+        args = dict(*args)
```
abergeron: It might be better if we accepted a single dictionary argument in addition to the kwargs.

nouiz: I'm not sure this is clear enough for me. Do you mean keeping *args as a parameter, then make sure it is only 1 element and it is a dict?

If so, this would be more restrictive then accepting anything that get converted to a dict. I don't have a strong preference. I just don't understand what it would bring to be more restrictive here except that the code would be more similar, but would make it harder for people to remember and probably look up more frequently what is accepted.

lamblin: I find it confusing to have `*args` here, but to require it to only be 1 argument at most (otherwise the call to `dict` fails).
Why not `def __init__(self, d=None, **kwargs):` and then `if not d: d = {}; else d = dict(d)`?
I won't hold the release for that, though.

nouiz: I think you are miss reading. Here there isn't any constrain on how many args there is. The only constrain is that dict accept it, so that we can convert it to a dict.

The line 108 with the assert of len(l)==1, is there to make sure the key only correspond to 1 Theano flag. In fact, if this is false, the problem isn't with the parameter args, but with Theano itself. We should never have 2 Theano flags with the same name.

theano/configparser.py
```diff
@@ -98,15 +98,17 @@ class change_flags(object):
 
     Useful during tests.
     """
-    def __init__(self, **kwargs):
+    def __init__(self, args=[], **kwargs):
```
abergeron: Don't use a list as a default argument.  I know it doesn't really matter here, but it's a bad habit and a bad example.

lamblin: Maybe an empty tuple?

bin/theano-test
```diff
@@ -1,3 +0,0 @@
-#!/usr/bin/env python
-import theano_test
-theano_test.main()
```
lamblin: If you remove those files, you have to remove references to it from `setup.py` as well.

lamblin: And from `MANIFEST.in`.

nouiz:
I removed the extra commit from this PR. It was already in another PR: https://github.com/Theano/Theano/pull/5651

I'll do the comment related to that in its own PR.

f0k:
> It would just put back the user shape as compiler hint as it was done in some of the ops

That's an important point, actually. For things like meta-optimization, we may need a way of specifying an example shape to optimize for, without asserting that this will be the only possible shape at runtime. +2 for not using the assertion by default!

PR #5646, opened 2017-03-02
Fix problems in daily buildbot (debug mode)

abergeron:
Some failures are not addressed:
- ModTester.test_grad will fail if the two inputs are really close to one another because then you wrap around and numeric_grad is super large.  This happens pretty rarely.
- test_dnn_batchnorm_train in the old backend failed once.  It seems to be a random fluke and we are about to delete the backend so I didn't want to investigate further.

### DIFF
--- a/theano/tensor/tests/test_nlinalg.py
+++ b/theano/tensor/tests/test_nlinalg.py
@@ -404,7 +404,7 @@ class test_Eig(utt.InferShapeTester):
         self.rng = numpy.random.RandomState(utt.fetch_seed())
         self.A = theano.tensor.matrix(dtype=self.dtype)
         self.X = numpy.asarray(self.rng.rand(5, 5),
-                          dtype=self.dtype)
+                               dtype=self.dtype)
         self.S = self.X.dot(self.X.T)
 
     def test_infer_shape(self):
@@ -440,6 +440,8 @@ class test_Eigh(test_Eig):
 
     def test_grad(self):
         X = self.X
+        # We need to do the dot inside the graph because Eigh needs a
+        # matrix that is hermitian
         utt.verify_grad(lambda x: self.op(x.dot(x.T))[0], [X], rng=self.rng)
         utt.verify_grad(lambda x: self.op(x.dot(x.T))[1], [X], rng=self.rng)
         utt.verify_grad(lambda x: self.op(x.dot(x.T), 'U')[0], [X], rng=self.rng)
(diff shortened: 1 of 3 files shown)


theano/tensor/tests/test_nlinalg.py
```diff
@@ -403,9 +403,9 @@ def setUp(self):
         super(test_Eig, self).setUp()
         self.rng = numpy.random.RandomState(utt.fetch_seed())
         self.A = theano.tensor.matrix(dtype=self.dtype)
-        X = numpy.asarray(self.rng.rand(5, 5),
+        self.X = numpy.asarray(self.rng.rand(5, 5),
                           dtype=self.dtype)
```
nouiz: flake8!

theano/tensor/tests/test_nlinalg.py
```diff
@@ -439,21 +439,19 @@ def test_uplo(self):
                                   vl * numpy.sign(vl[0, :]))
 
     def test_grad(self):
-        S = self.S
-        utt.verify_grad(lambda x: self.op(x)[0], [S], rng=self.rng)
-        utt.verify_grad(lambda x: self.op(x)[1], [S], rng=self.rng, eps=1e-4)
-        utt.verify_grad(lambda x: self.op(x, 'U')[0], [S], rng=self.rng)
-        utt.verify_grad(lambda x: self.op(x, 'U')[1], [S], rng=self.rng, eps=1e-4)
+        X = self.X
+        utt.verify_grad(lambda x: self.op(x.dot(x.T))[0], [X], rng=self.rng)
```
nouiz: Add a comment why you add dot here.

theano/tensor/tests/test_slinalg.py
```diff
@@ -80,19 +73,16 @@ def test_cholesky_grad():
...
-        eps = 2e-8
+    r = rng.randn(5, 5).astype(config.floatX)
     # Check the default.
-    yield (lambda: utt.verify_grad(cholesky, [pd], 3, rng, eps=eps))
+    yield (lambda: utt.verify_grad(lambda r: cholesky(r.dot(r.T)),
+                                   [r], 3, rng))
     # Explicit lower-triangular.
-    yield (lambda: utt.verify_grad(Cholesky(lower=True), [pd], 3,
-                                   rng, eps=eps))
+    yield (lambda: utt.verify_grad(lambda r: Cholesky(lower=True)(r.dot(r.T)),
```
nouiz: idem, doc why you put dot here.

Issue #5649, opened 2017-03-03
*** RuntimeError: error getting worksize: CUDNN_STATUS_NOT_SUPPORTED ** 0.9.0rc2


jc-Rosichini:
**the problem:**
changing floatX = float32 to floatX = float64 triggers a runtime error 'CUDNN_STATUS_NOT_SUPPORTED'

**my config:**
theano:  0.9.0rc2.dev-674f8d5148cb8dcea446e7243e629d8f83647da5
numpy:   1.11.0
pygpu:   0.6.1
lasagne:  0.2.dev1
Using cuDNN version 5105 
GeForce GTX 980M
Windows 10 x64

**theanorc:**

[global]
#floatX = float32         # works ok
floatX = float64
base_compiledir=E:\theano_compiledirxx

#device=gpu
device=cuda0
optimizer=fast_run
#optimizer_including=cudnn
allow_gc=True

[lib]
cnmem=0.815

[cuda]
root=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v8.0

[nvcc]
compiler_bindir=C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin
optimizer_including=cudnn
flags=-LC:\Users\PCGP1\Anaconda3\libs 

[dnn]
library_path = C:\CUDA\v8.0\lib\x64
enabled = True

**Dump with floatX = float64 **

Compilation starts at 2017-03-03 12:26:22
optimizer: rmsprop, initial learning rate:  0.001
train compilation took 391.546 seconds
Compilation starts at 2017-03-03 12:32:58
valid compilation took 132.803 seconds
Training starts at 2017-03-03 12:35:17

Traeback (most recent call last):
  File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\function_module.py", line 884, in __call__
    self.fn() if output_subset is None else\
RuntimeError: error getting worksize: CUDNN_STATUS_NOT_SUPPORTED

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "train.py", line 601, in <module>
    initiate_training(cf)
  File "train.py", line 567, in initiate_training
    train(cf)
  File "train.py", line 442, in train
    history, nan_occured = batch_loop(train_iter, train_fn, epoch, 'train', history, limit_randomize)
  File "train.py", line 97, in batch_loop
    loss, logT, logonmT, ypred, I, U, acc = f(X, y) # avec f  train_function or val_function
[...]

PR #5650, opened 2017-03-03
Fix precision in cudnn

nouiz:
Fix gh-5649. Replace https://github.com/Theano/Theano/pull/5569

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1179,21 +1179,10 @@ def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
 def dnn_gradweight3d(img, topgrad, kerns_shp, border_mode='valid',
                      subsample=(1, 1, 1), conv_mode='conv', precision=None):
     """
-    TODO: document this
+    3d version of dnn_gradweight
     """
-    ctx_name = infer_context_name(img, topgrad)
-    img = as_gpuarray_variable(img, ctx_name)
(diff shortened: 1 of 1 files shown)


theano/gpuarray/dnn.py
```diff
@@ -1169,31 +1167,26 @@ def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
...
-    topgrad = as_gpuarray_variable(topgrad, ctx_name)
-    img = gpu_contiguous(img)
-    topgrad = gpu_contiguous(topgrad)
-    kerns_shp = as_tensor_variable(kerns_shp)
-    desc = gpu_dnn_conv_desc(border_mode=border_mode, subsample=subsample,
-                             conv_mode=conv_mode)(kerns_shp)
-    out = gpu_alloc_empty(ctx_name, dtype=img.dtype)(*kerns_shp)
-    return gpu_dnn_conv_gradW()(img, topgrad, out, desc)
+    return dnn_gradweight3d(img, topgrad, kerns_shp, border_mode,
+                            subsample, conv_mode, precision)
```
abergeron: This is now an infinite recursion loop.  You probably meant dnn_gradweight()

nouiz: J'ai amendé le commit.

lamblin: Wait, what? Is it just calling itself?

PR #5657, opened 2017-03-05
Fixing the precision of GpuDnnConvDesc in #5555

gurunathpr:
Made changes in class GpuDnnConvDesc and gpu_dnn_conv_desc to enable handling of float64 values by using get_precision function

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -354,7 +354,9 @@ class GpuDnnConvDesc(COp):
     See the doc of :func:`dnn_conv` for a description of the parameters
 
     """
-
+    
+    prec_GpuDnn = get_precision(precision, [img, kerns])
+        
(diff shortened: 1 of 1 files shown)


theano/gpuarray/dnn.py
```diff
@@ -354,7 +354,9 @@ class GpuDnnConvDesc(COp):
     See the doc of :func:`dnn_conv` for a description of the parameters
 
     """
-
+    
+    prec_GpuDnn = get_precision(precision, [img, kerns])
```
nouiz: I don't know how this commit can work. Where img and kerns come from?

Anyway, I fixed Friday: https://github.com/Theano/Theano/pull/5650 and the fix is merged.

thanks for trying to fix this. I wanted to have this fixed quickly for the rc3.

gurunathpr: @nouiz , will replacing with the default precision (in below line) work?

`precision = `theano.config.dnn.conv.precision``

PR #5665, opened 2017-03-06
Optimization to remove upcast in local_cast_cast

### LINKED ISSUE #949: Make an optimization to remove cast from graph
import theano
v=theano.tensor.bvector()
f=theano.function([v], v.astype('int32').astype('int8'))
f.maker.fgraph.toposort()
[Elemwise{Composite{[Cast{int8}(Cast{int32}(i0))]}}(<TensorType(int8, vector)>)]

in the opt local_cast_cast in file tensor/opt.py

vikramnitin9:
fix #949
I've made a list of pairs of data types that would constitute an upcast, in **opt.py** in the function _upcast_
Function _local_cast_cast_ has been updated accordingly.

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -2225,7 +2225,7 @@ def local_cast_cast(node):
         # We don't need to copy over any stack traces here
         return [x]
 
-    if(upcast(base.dtype, type1.dtype)):
+    if(is_an_upcast(base.dtype, type1.dtype)):
         # Checking for further redundancy. Eg: int8 -> int32 -> int8
         if(type2.dtype == base.dtype):
             return x.owner.inputs
@@ -2237,17 +2237,23 @@ def local_cast_cast(node):
             return [v]
 
 
-def upcast(type1, type2):
+def is_an_upcast(type1, type2):
     """Given two data types (as strings), check if converting to
     type2 from type1 constitutes an upcast.
+    Differs from theano.scalar.upcast
 
     """
     category = {
-        # Pair of numbers : the 'super-index' and 'sub-index'
-        'uint8': (0, 0), 'uint16': (0, 1), 'uint32': (0, 2), 'uint64': (0, 3),
-        'int8': (1, 0), 'int16': (1, 1), 'int32': (1, 2), 'int64': (1, 3),
-        'float16': (2, 0.5), 'float32': (2, 1.5), 'float64': (2, 2.5),
-        'complex64': (3, 2), 'complex128': (3, 3)
+        # The first number in the pair is the dtype (bool, uint, int, float,
+        # complex). Conversion from higher to lower is never an upcast.
+        # The second number roughly indicates the precision. Again, conversion
+        # from higher to lower is never an upcast.
+
+        'bool': (0, 0),
+        'uint8': (1, 1), 'uint16': (1, 2), 'uint32': (1, 3), 'uint64': (1, 4),
+        'int8': (2, 1), 'int16': (2, 2), 'int32': (2, 3), 'int64': (2, 4),
+        'float16': (3, 1.5), 'float32': (3, 2.5), 'float64': (3, 3.5),
+        'complex64': (4, 3), 'complex128': (4, 4)
     }
 
     cat1 = category[type1]
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -2204,7 +2204,7 @@ def local_cast_cast(node):
 
     when those contrain:
     dtype1 == dtype2
-    TODO: the base dtype is the same (int, uint, float, complex)
+    OR the base dtype is the same (int, uint, float, complex)
           and the first cast cause an upcast.
 
     """
@@ -2216,10 +2216,51 @@ def local_cast_cast(node):
             not isinstance(x.owner.op, T.Elemwise) or
             not isinstance(x.owner.op.scalar_op, scalar.Cast)):
         return
-    if node.op.scalar_op.o_type == x.owner.op.scalar_op.o_type:
+
+    type1 = x.owner.op.scalar_op.o_type
+    type2 = node.op.scalar_op.o_type
+    base = x.owner.inputs[0]
+
+    if type1 == type2:
         # We don't need to copy over any stack traces here
         return [x]
 
+    if(upcast(base.dtype, type1.dtype)):
+        # Checking for further redundancy. Eg: int8 -> int32 -> int8
+        if(type2.dtype == base.dtype):
+            return x.owner.inputs
+        else:
+            # Apply the second cast only
+            v = node.op(base)
+            return [v]
+
+def upcast(type1, type2):
+    '''
+    Given two data types (as strings), check if converting to 
+    type2 from type1 constitutes an upcast.
+    '''
+    upcast_pairs = (
+        ('int8','int16'),('int8','int32'),('int8','int64'),('int16','int32'),('int16','int64'),('int32','int64'),
+        ('uint8','uint16'),('uint8','uint32'),('uint8','uint64'),('uint16','uint32'),('uint16','uint64'),('uint32','uint64'),
+        ('float16','float32'),('float16','float32'),('float16','float64'),('float32','float64'),
+        ('complex64','complex128'),
+
+        ('uint8','int16'),('uint8','int32'),('uint8','int64'),('uint16','int32'),('uint16','int64'),('uint32','int64'),
+        ('int8','float16'),('int8','float32'),('int8','float64'),('int16','float32'),('int16','float64'),('int32','float64'),
+        ('uint8','float16'),('uint8','float32'),('uint8','float64'),('uint16','float32'),('uint16','float64'),('uint32','float64'),
+
+        ('int8','complex64'),('int16','complex64'),
+        ('uint8','complex64'),('uint16','complex64'),
+        ('float32','complex64'),
+        ('int8','complex128'),('int16','complex128'),('int32','complex128'),
+        ('uint8','complex128'),('uint16','complex128'),('uint32','complex128'),
+        ('float32','complex128'),('float64','complex128')
+    )
+
+    for pair in upcast_pairs:
+        if(type1 == pair[0] and type2 == pair[1]):
+            return True
+    return False
 
 @register_canonicalize
 @register_specialize
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -2232,6 +2232,8 @@ def local_cast_cast(node):
         else:
             # Apply the second cast only
             v = node.op(base)
+            # Copy stack trace from the output of the original cast
+            copy_stack_trace(node.outputs[0], v)
             return [v]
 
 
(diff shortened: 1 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -2216,10 +2216,64 @@ def local_cast_cast(node):
...
+        ('int8', 'complex128'), ('int16', 'complex128'), ('int32', 'complex128'),
+        ('uint8', 'complex128'), ('uint16', 'complex128'), ('uint32', 'complex128'),
+        ('float32', 'complex128'), ('float64', 'complex128')
+    )
+
+    for pair in upcast_pairs:
+        if(type1 == pair[0] and type2 == pair[1]):
+            return True
+    return False
+
```
nouiz: This will need tests. They would go as a new method in the class T_cast_cast in the file tensor/tests/test_opt.py

Make sure to test case of mixed dtype and case where we should not optimize.

theano/tensor/opt.py
```diff
@@ -2216,10 +2216,64 @@ def local_cast_cast(node):
...
         return [x]
 
+    if(upcast(base.dtype, type1.dtype)):
+        # Checking for further redundancy. Eg: int8 -> int32 -> int8
+        if(type2.dtype == base.dtype):
+            return x.owner.inputs
+        else:
+            # Apply the second cast only
+            v = node.op(base)
+            return [v]
```
nouiz: Add this line just before the return:

    v = copy_stack_trace(node.outputs[0], v)

this help Theano give better error message. It allow us to tell which node was created in which file/line by the user. So they can find the mapping between the error and their code

theano/tensor/opt.py
```diff
@@ -2204,7 +2204,7 @@ def local_cast_cast(node):
 
     when those contrain:
     dtype1 == dtype2
-    TODO: the base dtype is the same (int, uint, float, complex)
+    OR the base dtype is the same (int, uint, float, complex)
           and the first cast cause an upcast.
 
     """
```
nouiz: Add this or similar:

Also remove single cast that cast to the original dtype. like:
x{fvector}.cast('float32') -> x

this was already done and missing from the comment.

theano/tensor/opt.py
```diff
@@ -2216,10 +2216,64 @@ def local_cast_cast(node):
...
+        ('uint8', 'complex64'), ('uint16', 'complex64'),
+        ('float32', 'complex64'),
+        ('int8', 'complex128'), ('int16', 'complex128'), ('int32', 'complex128'),
+        ('uint8', 'complex128'), ('uint16', 'complex128'), ('uint32', 'complex128'),
+        ('float32', 'complex128'), ('float64', 'complex128')
+    )
+
+    for pair in upcast_pairs:
+        if(type1 == pair[0] and type2 == pair[1]):
+            return True
```
nouiz: I don't like this big list, looping on it could be slow. Can you do a dict instead, where the key is type1 and the value is a list of the allowed upcast:

{'int8':['int16', 'int32', 'int64', uint..., float*..., complex,...], ....}

This will do less looping and comparison, so should be faster.

theano/tensor/opt.py
vikramnitin9: Is this handled by this local_cast_cast function? I think it's eliminated at an earlier stage of the optimization. This function doesn't seem to be called at all when something like "x{fvector}.cast('float32') -> x" is done. Although I can't trace where exactly.

Also, a related question : is the type1 == type2 comparison also redundant? I think it's eliminated at the same previous stage, ie, x.cast(dtype1).cast(dtype1) gets reduced to x.cast(dtype1) without the aid of the local_cast_cast function.

theano/tensor/opt.py
vikramnitin9: Adding v = copy_stack_trace(node.outputs[0], v) gives "ValueError: A local optimizer tried to remove a Variable that is used".
v = copy_stack_trace([node.outputs[0], node.inputs[0]], v) doesn't seem to solve the problem either.
Any ideas?

theano/tensor/opt.py
nouiz: You are right, this isn't done, so don't do that!

theano/tensor/opt.py
nouiz: Can you give the full error message?

vikramnitin9: I solved it by making it just "copy_stack_trace(node.outputs[0], v)" instead of "v = copy_stack_trace(node.outputs[0], v)"

vikramnitin9:
I can't seem to fix this... my own test fails. In the _test_upcast_ function of _T_cast_cast class_ in _theano/tensor/tests/opt.py_, 
the last line `assert isinstance(topo[0].op.scalar_op, scal.basic.Composite)` fails.

It works when I run it as a stand-alone file (without the _mode_ parameter set).
Any idea what I'm doing wrong? Thanks.

vikramnitin9:
The problem occurs when I set mode='FAST_COMPILE'. The optimization doesn't seem to be performed at all.

nouiz:
In fast_compile, not all optimization are enabled.

So if we want to force an optimization to be included, we must do it. The
other tests in the class does it. In setUp(), it create a special mode that
force to include an optimization.

Then when compiling, they do: f=theano.function(...mode=self.mode)

That should fix the problem you describe.

On another point, we don't like to have merge in PR. this make the history
ugly and make the review of PR more complicated. Can you remove them? You
can do a rebase and they should be removed. If you need help, I could find
you some doc that explain that.

vikramnitin9:
I'm sorry for the merge mess, I'm working on it. The code should be fine now though.

nouiz:
To help you about the too many commits, sometimes it is easier to start a new branch and use the commit "git cherry-pick commit" on the commit you want to keep. Read the doc of that command. Maybe it can help you.

vikramnitin9:
Commit history is clean now, although the last commit shows that I've modified some 15 files. Actually, I've modified only one, theano/tensor/tests/test_opt.py.
I suppose a rebase will be needed now?

vikramnitin9:
Should I do a rebase now?

nouiz:
no need for a rebase.

vikramnitin9:
Thanks for the help with sorting out the merge mess.
Is the code itself fine now?

theano/tensor/opt.py
```diff
@@ -2216,10 +2216,48 @@ def local_cast_cast(node):
...
+        'complex64': (3, 2), 'complex128': (3, 3)
+    }
+
+    cat1 = category[type1]
+    cat2 = category[type2]
+
+    if(cat2[0] >= cat1[0] and cat2[1] > cat1[1]):
+        return True
+    else:
+        return False
```
abergeron: I'm not sure I like having another definition of what is an upcast and what is not.  This should use `scalar.upcast()` like the rest of the code.

Also, this will crash with `'bool'`.

nouiz:
I'm not sure we can use theano.scalar.upcast(), when the dtypes are mixed.
Vikram, can you double check that using upcast() won't give the same
results? If it does give the same results, then we can use it. Otherwise,
it will need a fix for bool.

vikramnitin9:
The results are different for the following pairs:
uint64 float64
uint64 complex128
int64 float64
int64 complex128

My function says False, whereas the scalar.upcast says True.

Shouldn't it be not an upcast? float64 has a 52 bit mantessa and a 11 bit exponent. How can it represent a 64 bit integer without loss of precision?

vikramnitin9:
The results are different for the following pairs:
uint64 float64
uint64 complex128
int64 float64
int64 complex128

My function says False, whereas the scalar.upcast function says True.

Shouldn't it be False? float64 has a 52 bit mantessa and a 11 bit exponent. How can it represent a 64 bit number without loss of precision?

theano/tensor/opt.py
```diff
@@ -2240,39 +2242,21 @@ def upcast(type1, type2):
...
-        ('uint8', 'complex128'), ('uint16', 'complex128'), ('uint32', 'complex128'),
-        ('float32', 'complex128'), ('float64', 'complex128')
-    )
+    category = {
+        # Pair of numbers : the 'super-index' and 'sub-index'
+        'uint8': (0, 0), 'uint16': (0, 1), 'uint32': (0, 2), 'uint64': (0, 3),
+        'int8': (1, 0), 'int16': (1, 1), 'int32': (1, 2), 'int64': (1, 3),
+        'float16': (2, 0.5), 'float32': (2, 1.5), 'float64': (2, 2.5),
+        'complex64': (3, 2), 'complex128': (3, 3)
+    }
```
nouiz: You need to add bool here. Otherwise there will be some crashes

theano/tensor/opt.py
```diff
@@ -2240,39 +2242,21 @@ def upcast(type1, type2):
...
-
-        ('int8', 'complex64'), ('int16', 'complex64'),
-        ('uint8', 'complex64'), ('uint16', 'complex64'),
-        ('float32', 'complex64'),
-        ('int8', 'complex128'), ('int16', 'complex128'), ('int32', 'complex128'),
-        ('uint8', 'complex128'), ('uint16', 'complex128'), ('uint32', 'complex128'),
-        ('float32', 'complex128'), ('float64', 'complex128')
-    )
+    category = {
+        # Pair of numbers : the 'super-index' and 'sub-index'
```
nouiz: The logic seem good, but it was hard to understand. Can you add more comments like:

super-index is an dtype (uint, int, ...) number that will only allow legal upcast of the base dtype.
sub-index is an number that represent the bits used for the exact precision numbers, to be sure we upcast the precision.

theano/tensor/opt.py
```diff
@@ -2216,10 +2216,48 @@ def local_cast_cast(node):
...
+            return x.owner.inputs
+        else:
+            # Apply the second cast only
+            v = node.op(base)
+            # Copy stack trace from the output of the original cast
+            copy_stack_trace(node.outputs[0], v)
+            return [v]
+
+
+def upcast(type1, type2):
```
nouiz: this isn't a real upcast method. So I would rename it to something like:

is_an_upcast()

Also, in the docstring, tell that it isn't exactly equivalent to theano.scalar.upcast.

vikramnitin9: Why does `theano.scalar.upcast('int64', 'float64')` return `float64`? I think that int64 cannot be cast to float64 without loss of precision.
I'll rename the function and add the docstring changes, but I was just wondering if I could be wrong about int64 to float64.

theano/tensor/tests/test_opt.py
```diff
@@ -4658,7 +4658,7 @@ def test_consecutive(self):
         f(dx)
         topo = f.maker.fgraph.toposort()
         assert len(topo) == 1
-        assert isinstance(topo[0].op.scalar_op, scal.basic.Cast)
+        assert isinstance(topo[0].op, T.Elemwise)
```
vikramnitin9: `assert isinstance(topo[0].op, T.Elemwise)` is the current line in the Theano repo. I thought it needed changing, and then put it back when my own tests started failing. I'll change it to `scal.basic.Cast` again.

nouiz:
Le sam. 18 mars 2017 03:39, Vikram Nitin <notifications@github.com> a
écrit :

> *@vikramnitin9* commented on this pull request.
> ------------------------------
>
> In theano/tensor/opt.py
> <https://github.com/Theano/Theano/pull/5665#discussion_r106775739>:
>
> >          # We don't need to copy over any stack traces here
>          return [x]
>
> +    if(upcast(base.dtype, type1.dtype)):
> +        # Checking for further redundancy. Eg: int8 -> int32 -> int8
> +        if(type2.dtype == base.dtype):
> +            return x.owner.inputs
> +        else:
> +            # Apply the second cast only
> +            v = node.op(base)
> +            # Copy stack trace from the output of the original cast
> +            copy_stack_trace(node.outputs[0], v)
> +            return [v]
> +
> +
> +def upcast(type1, type2):
>
> Why does theano.scalar.upcast('int64', 'float64') return float64? I think
> that int64 cannot be cast to float64 without loss of precision.
>

This is what numpy and c upcast does. It is better to lose precision then
to lose range. Ideally, it would return a float with more bits, but that
don't exist.

I'll rename the function and add the docstring changes, but I was just

nouiz:
thanks!

Issue #5478, opened 2017-02-02
Confusing error when not using non_sequences arg to scan.


petered:
I had this bug that confused me for a bit.  I think the error message could be improved to help point to the problem (or the "problem" should not be a problem at all...).

Here is some code that causes the error:
```
import theano
import theano.tensor as tt

c = theano.shared(0.)
inc = tt.scalar()

def count_up():
    return tt.zeros(()), {c: c+inc}

_, updates = theano.scan(count_up, n_steps=20)
func = theano.function(inputs=[inc], outputs=[], updates = updates)
func(2)
assert c.get_value()==40.
```
The error is:
```
theano.gof.fg.MissingInputError: A variable that is an input to the graph was neither provided as an input to the function nor given a value. A chain of variables leading from this input to an output is [<TensorType(float64, scalar)>, Elemwise{add,no_inplace}.0]. This chain may not be unique
Backtrace when the variable is created:
  File "/Users/peter/projects/spiking-experiments/scratch/demo_scan_fault.py", line 7, in <module>
    inc = tt.scalar()
```

It turns out that what should actually be done is to define inc as a non-sequence.  The correct code is

```
import theano
import theano.tensor as tt

c = theano.shared(0.)
inc = tt.scalar()

def count_up(inc_):
    return tt.zeros(()), {c: c+inc_}

_, updates = theano.scan(count_up, n_steps=20, non_sequences=[inc])
func = theano.function(inputs=[inc], outputs=[], updates = updates)
func(2)
assert c.get_value()==40.
```

It seems that either
A) The first code should not be considered a bug, and the use of "inc" should be detected and silently added as a non_sequence input to the scan loop.
B) There error should actually points to the fact that you should have added all variables used in your scanned function as non_sequences arguments when calling scan.

nouiz:
I agree, a better error would be great. Thanks for the report.

lamblin:
Closed by #5667.

PR #5667, opened 2017-03-06
Added meaningful message when missing inputs to scan.

Thrandis:
Solves #5478


nouiz:
I tested it myself and it was working good, except the stack trace was printed 2 times. I added a commit that remove one of the printing. I'm ok to merge, but I'll try to get another review.

PR #5673, opened 2017-03-07
docs: Fix openmp config section

mdraw:
- openmp has been disabled by default since https://github.com/Theano/Theano/commit/f1f06bb960f63e098a7ce32de50c8547d2fb7713.
- grammar improvements

### DIFF
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -253,13 +253,12 @@ import theano and print the config variable, as in:
 
     Bool value: either ``True`` or ``False``
 
-    Default: ``True`` if the environment variable OMP_NUM_THREADS!=1 or
-             if we detect more than 1 CPU core. Otherwise ``False``.
+    Default: ``False``
 
-    Enable or not parallel computation on the CPU with OpenMP.
-    It is the default value used when creating an Op that support it.
-    The best is to define it via Theano configuration
-    file or with the environment variable THEANO_FLAGS.
+    Enable or disable parallel computation on the CPU with OpenMP.
+    It is the default value used when creating an Op that supports it.
+    It is best to define it in .theanorc
+    or in the environment variable THEANO_FLAGS.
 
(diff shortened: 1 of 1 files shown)


mdraw:
On a sidenote, I just compared the training speed of a small CNN on an Intel X5670 (8 cores, 16 threads) with openmp off and on. Training with openmp was about 264% faster. I don't know how representative this experiment is, but for me there was a significant speedup.

JesseLivezey:
@mdraw you can find additional timing info in the PR that added openmp to the corrmm op here #4591. It does look like having openmp=True as default might make more sense. For cpus with more than 2 cores, it's faster to have it enabled.

nouiz:
@abergeron @lamblin should/could we enable it by default now? The problem was with the old convolution that too frequently was slower with multiple threads. This was fixed.

So it would make sense to enable it by default again.

lamblin:
I'm OK with updating the doc to be more precise, of course :)
I would not change the default before the release, but after that, why not. We should be careful that everything works on platforms where g++ may not support openMP by default (MacOS with clang++ and Windows with some versions of gcc).

nouiz:
The multi OS problem is already taken care of. Maybe it need to be retested, but it should be fine.

I think we can change in the master now the default as now the stuff for 0.9 final is only in the branch. We agreed to merge in master stuff that won't go in 0.9.

@mdraw what is your OS?

mdraw:
@JesseLivezey thanks for the link!
@nouiz I ran it on CentOS 6 x86_64, using Theano 0.8.2 from conda-forge (packages linked to OpenBLAS).
To be clear, I intended this PR to go into 0.9 if it is not too late for that. It just fixes the inconsistency between docs and actual code. I agree enabling openmp by default would probably be too late for 0.9.

nouiz:
Ok. I merge now and made an issue to change the default.

I'll back-port this to the 0.9.X in a few minutes.

PR #5682, opened 2017-03-09
dnn.enabled=no_check

nouiz:
To help speed the import, in particular when multiple process start at the same time.

### DIFF
--- a/doc/install_ubuntu.txt
+++ b/doc/install_ubuntu.txt
@@ -30,7 +30,7 @@ For Ubuntu 16.04 with cuda 7.5
 
 .. code-block:: bash
 
-    sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libopenblas-dev git
+    sudo apt-get install python-numpy python-scipy python-dev python-pip python-nose g++ libopenblas-dev git graphviz
     sudo pip install Theano
 
     # cuda 7.5 don't support the default g++ version. Install an supported version and make it the default.
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -693,7 +693,7 @@ import theano and print the config variable, as in:
 
     If ``'False'``, do not use cuDNN or check if it is available.
 
-    If ``'no_check'``, assume present and the version between heasers and linked match (so less compilation at context init)",
+    If ``'no_check'``, assume present and the version between header and library match (so less compilation at context init)",
 
 .. attribute:: config.conv.assert_shape
 
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -693,6 +693,8 @@ import theano and print the config variable, as in:
 
     If ``'False'``, do not use cuDNN or check if it is available.
 
+    If ``'no_check'``, assume present and the version between heasers and linked match (so less compilation at context init)",
+
 .. attribute:: config.conv.assert_shape
 
     If True, AbstractConv* ops will verify that user-provided
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -670,7 +670,7 @@ class Function(object):
             if name:
                 message = name
             else:
-                message = str(maker.profile.message) + " copy"
+                message = str(profile.message) + " copy"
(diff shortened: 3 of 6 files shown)


nouiz:
See thread: [theano-dev] quick way to simultaneously initialize multiple GPUs in multiple processes?

lamblin:
I'm OK with that change, but I'm not convinced it should go in 0.9.

abergeron:
I would like to try it on a machine with no cudnn and/or broken cudnn to make sure it would be ok.

Also, the gain in compilation time is less than 1 second at startup.  If it is a problem that multiple processes are fighting for the lock then we can address that with a better cache protocol (which I have ideas for).

In any case this is not something I believe we should change in the rcs.

nouiz:
Agree that this can wait after 0.9. But I don't want to wait for a proper cache fixes. This is in the waiting time for a long time already and this quick work around allow people to wait more easily until it get done. I think we should still keep it when done. interactive use is important and for people that care about that, they would have the options.

nouiz:
On a computer when cudnn isn't available, with dnn.enabled=auto, then it just disable it as it should.
with dnn.enabled=True, it raise an error, with dnn.enabled=False, it just don't use it and don't test it.

With dnn.enabled=no_check, it fail later with an ugly compilation error with the dump of the c code. This is normal, this happen when it try to get the version, as it compile c code and it don't find cudnn.h.

With a mixed environment, then nothing is garanteed from our point of view. It depend of the mix and how cudnn handle that. I don't think we should test that cases.

abergeron:
You fixed the doc, but not the flag help string.

astooke:
Hi, thanks for working on this!  But it is not complete.  Even with this flag, I still get the slow start-up and compile-lock messages.  I think it comes from inside the `version()` function in `gpuarray/dnn.py`, where it has to compile a Theano function the first time...seems like `version()` is called in a lot of places.  Even when I comment out the call to `_dnn_check_version()` immediately after the if-statement for the `no_check`, this still happens.  It loads fully, but I never see the message `"presence check disabled"`.

Perhaps there is a way to get the version from `cudnn.h` without compiling a Theano function??

Loading up 8 GPUs with `dnn.enabled=False,gpuarray.preallocate=1` happens almost instantaneously, which is very nice!

nouiz:
Can you try this diff?

diff --git a/theano/gof/cmodule.py b/theano/gof/cmodule.py
index ffac6be..a95fadb 100644
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -1126,7 +1126,10 @@ class ModuleCache(object):
         if module is not None:
             return module

-        with compilelock.lock_ctx(keep_lock=keep_lock):
+        # If not version, it will be recompiled for each process, so
+        # no need to take a lock.
+        version, _ = key
+        with compilelock.lock_ctx(keep_lock=keep_lock, take_lock=version):
             # 1) Maybe somebody else compiled it for us while we
             #    where waiting for the lock. Try to load it again.
             # 2) If other repo that import Theano have Theano ops defined,
diff --git a/theano/gof/compilelock.py b/theano/gof/compilelock.py
index d5e2d82..2a1e6b7 100644
--- a/theano/gof/compilelock.py
+++ b/theano/gof/compilelock.py
@@ -35,10 +35,11 @@ def force_unlock():


 @contextmanager
-def lock_ctx(lock_dir=None, keep_lock=False, **kw):
-    get_lock(lock_dir=lock_dir, **kw)
+def lock_ctx(lock_dir=None, keep_lock=False, take_lock=True, **kw):
+    if take_lock:
+        get_lock(lock_dir=lock_dir, **kw)
     yield
-    if not keep_lock:
+    if not keep_lock and take_lock:
         release_lock()



Mostly, if we compile for each process a file (so bypass the cache) we do
not need to take the lock. This diff do this change.

astooke:
OK I'm starting vacation for a week but will try it when I get back.

In a related note, when we have multiple worker processes simultaneously unpickling functions from the master, there are compile lock warnings all over the place and it slows down.  Is there a way to make the workers do less work and just trust what they are unpickling or use what the master has already compiled?

Issue #5669, opened 2017-03-07
cuda launch blocking back in hints & documentation


astooke:
It seems like profiling GPU use still requires setting `CUDA_LAUNCH_BLOCKING=1` to get meaningful results?  But this suggestion no longer appears when profiling a function, and it seems to have disappeared from readthedocs (it was in the CUDA backend section)....tripped me up for a bit before I remembered!

lamblin:
I just tried with `device=cuda,profiling=True`, on the latest development version, and got the following error:
```
Exception: You are running the Theano profiler with CUDA enabled. Theano GPU ops execution is asynchronous by default. So by default, the profile is useless. You must set the environment variable CUDA_LAUNCH_BLOCKING to 1 to tell the CUDA driver to synchronize the execution to get a meaningful profile.
```

Can you elaborate on what you did to profile a GPU function without this message getting printed? Which Theano version are you using, Theano flags, etc.?

astooke:
Yes, sorry I didn't already.  I passed `profile=True` to a function and used `f.profile.summary()`, as described on [this page](http://deeplearning.net/software/theano_versions/0.9.X/tutorial/profiling.html) of the docs.  Neither that page of the docs nor the function produced the hint.

lamblin:
Indeed, we only raise the exception if `config.profile` is `True`, not if `profile=True` was passed manually. This does not look like a recent change.
We could raise the exception in this case as well, or just make it a warning (like for `ignore_first_call`).

nouiz:
https://github.com/Theano/Theano/pull/5686

does a warning in that case.

PR #5686, opened 2017-03-09
If the user do theano.function(profile=True), do a warning about CUDA…

nouiz:
…_LAUNCH_BLOCKING not being used.

fix gh-5669

### DIFF
--- a/theano/compile/profiling.py
+++ b/theano/compile/profiling.py
@@ -18,6 +18,7 @@ import sys
 import time
 from collections import defaultdict
 from six import iteritems
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


theano/gof/fg.py
```diff
@@ -16,7 +15,6 @@
 from theano import config
 
 from six import iteritems, itervalues
-from six.moves import StringIO
```
lamblin: These changes actually make flake8 fail. This should probably go in a different PR.

theano/compile/profiling.py
```diff
@@ -253,20 +254,25 @@ def reset(self):
...
+                " Theano GPU ops execution is asynchronous by default."
+                " So by default, the profile is useless."
+                " You must set the environment variable"
+                " CUDA_LAUNCH_BLOCKING to 1 to tell the CUDA driver to"
+                " synchronize the execution to get a meaningful profile.")
+            if config.profile:
+                raise Exception(msg)
+            else:
+                warnings.warn(msg)
+
```
lamblin: OK with that part.

nouiz:
I remobed the commit that was in the wrong PR. travis passed, only jenkins is missing.

PR #5687, opened 2017-03-09
Fix some DebugMode problems

abergeron:


theano/compile/debugmode.py
```diff
@@ -1026,7 +1026,7 @@ def _lessbroken_deepcopy(a):
     # This logic is also in link.py
     from theano.gof.type import _cdata_type
     if type(a) in (np.ndarray, np.memmap):
-        rval = a.copy()
```
lamblin: OK.

theano/sparse/tests/test_basic.py
```diff
@@ -2965,7 +2965,8 @@ def structured_function(*args):
     sparse.trunc,
     np.trunc,
     test_dtypes=[m for m in sparse.all_dtypes
-                 if not m in sparse.complex_dtypes])
+                 if not m in sparse.complex_dtypes],
+    grad_test=False)
```
lamblin: I don't think this is correct, we should actually check it gives 0 most of the time.
I have a fix in #5693 that avoid sampling close to the places it would be evaluated to non-zero.

abergeron: We can add another test that just checks that grad is always zero (because that is how it's defined).

But trying to use verify_grad and fiddling with the sampling is the wrong way to go I think.

nouiz: gh-5693 don't fix this particular test. Personaly, I think that messing with the input value for the tests is an easy and good fix.

@abergeron What else do you think we could do?

I put in https://github.com/Theano/Theano/pull/5702 the commit that we agree on for rc4

abergeron: We can't easily mess with the input values for sparse.

Also after some discussion, I realized that we could remove most of the tests for the sparse elemwise ops and just keep one or two to test the dense wrapper but there is no need to re-test all the ops themselves since they are tested in tensor.

lamblin: I did not realize this was in `sparse`, so I'm OK with it.

Issue #4115, opened 2016-02-24
Keep stack trace in opt: theano/gpuarray


lamblin:
##Sub-ticket of #3018.
Use copy_stack_trace for optimizations defined in `theano/gpuarray/`, and add checks that the stack trace is still there during tests.

nouiz:
We should not spend time in the old back-end for this. Instead, do this in theano/gpuarrray. There is at least the file opt.py and dnn.py.

lamblin:
We should start with `theano/gpuarray/opt.py`, most optimizations can be updated by simply updating `op_lifter`. Then, the optimizations that are not decorated with `@op_lifter(...)` should be updated.

PR #5688, opened 2017-03-10
gpuarray: keep stack trace

cooijmanstim:
for #4115
@lamblin 

- [ ] Follow up: Document: Variable.notify_construction_observers

### DIFF
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -6,6 +6,7 @@ amount of useful generic optimization tools.
 from __future__ import absolute_import, print_function, division
 
 from collections import deque, defaultdict, OrderedDict
+import contextlib
 import copy
 import inspect
 import logging
@@ -2975,6 +2976,24 @@ def with_stack_trace(from_var, to_var):
     """
     copy_stack_trace(from_var, to_var)
     return to_var
+
+
+@contextlib.contextmanager
+def inherit_stack_trace(from_var):
+    """
+    Contextmanager that copies the stack trace from one or more tensor variables to all tensor
+    variables constructed in the body.
+
+    Parameters
+    ----------
+    from_var
+        Tensor variable or list of tensor variables to copy stack traces from.
+
+    """
+    with graph.nodes_constructed() as new_nodes:
+        yield
+    copy_stack_trace(from_var, new_nodes)
+
 
 def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     """
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2953,13 +2953,14 @@ def copy_stack_trace(from_var, to_var):
 @contextlib.contextmanager
 def inherit_stack_trace(from_var):
     """
-    Contextmanager that copies the stack trace from one or more tensor variables to all tensor
-    variables constructed in the body.
+    Contextmanager that copies the stack trace from one or more variable nodes to all
+    variable nodes constructed in the body. new_nodes is the list of all the newly created
+    variable nodes inside an optimization that is managed by graph.nodes_constructed().
 
     Parameters
     ----------
     from_var
-        Tensor variable or list of tensor variables to copy stack traces from.
+        Variable node or a list of variable nodes to copy stack traces from.
 
     """
     with graph.nodes_constructed() as new_nodes:
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2903,7 +2903,7 @@ def pre_greedy_local_optimizer(list_optimizations, out):
 def copy_stack_trace(from_var, to_var):
     """
     Copies the stack trace from one or more tensor variables to
-    one or more tensor variables.
+    one or more tensor variables and returns the destination variables.
 
     Parameters
     ----------
@@ -2947,34 +2947,6 @@ def copy_stack_trace(from_var, to_var):
         # Copy over stack traces from from_var to each variable to
         # to_var, including the stack_trace of the to_var before
         to_var.tag.trace = getattr(to_var.tag, 'trace', []) + tr
-
-
-def with_stack_trace(from_var, to_var):
-    """
-    Copies the stack trace from one or more tensor variables to
-    one or more tensor variables and returns the destination variables.
-
-    Parameters
-    ----------
-    from_var
-        Tensor variable or list of tensor variables to copy stack traces from.
-    to_var
-        Tensor variable or list of tensor variables to copy stack traces to.
-
-    Returns
-    -------
-    tensor variable or list of tensor variables
-        `to_var`, augmented with the stack traces from `from_var`.
-
-    Notes
-    -----
-    The stacktrace is assumed to be of the form of a list of lists
-    of tuples. Each tuple contains the filename, line number, function name
-    and so on. Each list of tuples contains the truples belonging to a
-    particular variable.
-
-    """
-    copy_stack_trace(from_var, to_var)
     return to_var
 
 
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -2948,6 +2948,34 @@ def copy_stack_trace(from_var, to_var):
         to_var.tag.trace = getattr(to_var.tag, 'trace', []) + tr
 
 
+def with_stack_trace(from_var, to_var):
+    """
+    Copies the stack trace from one or more tensor variables to
+    one or more tensor variables and returns the destination variables.
+
+    Parameters
+    ----------
+    from_var
+        Tensor variable or list of tensor variables to copy stack traces from.
+    to_var
+        Tensor variable or list of tensor variables to copy stack traces to.
+
+    Returns
+    -------
+    tensor variable or list of tensor variables
+        `to_var`, augmented with the stack traces from `from_var`.
+
+    Notes
+    -----
+    The stacktrace is assumed to be of the form of a list of lists
+    of tuples. Each tuple contains the filename, line number, function name
+    and so on. Each list of tuples contains the truples belonging to a
+    particular variable.
+
+    """
+    copy_stack_trace(from_var, to_var)
+    return to_var
+
 def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
     """
     This function checks if the outputs of specific ops of a compiled graph
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -7,6 +7,7 @@ from theano import Apply, Op
 from theano.compile import optdb
 from theano.gof import LocalOptGroup, ParamsType
 from theano.scalar import bool as bool_t
+from theano.gof.opt import inherit_stack_trace
 from theano.tensor.basic import as_tensor_variable
 from theano.tensor.opt import in2out
 
@@ -1830,17 +1831,20 @@ class GpuCorr3dMM_gradInputs(BaseGpuCorr3dMM):
 
 @inplace_allocempty(GpuGemv, 0)
 def local_inplace_gpuagemv(node, inputs):
-    return [gpugemv_inplace(*inputs)]
+    with inherit_stack_trace(node.outputs):
+        return [gpugemv_inplace(*inputs)]
 
 
 @inplace_allocempty(GpuGemm, 0)
 def local_inplace_gpuagemm(node, inputs):
-    return [gpugemm_inplace(*inputs)]
+    with inherit_stack_trace(node.outputs):
+        return [gpugemm_inplace(*inputs)]
 
 
 @inplace_allocempty(GpuGer, 0)
 def local_inplace_gpuager(node, inputs):
-    return [gpuger_inplace(*inputs)]
+    with inherit_stack_trace(node.outputs):
+        return [gpuger_inplace(*inputs)]
 
 
 @inplace_allocempty(GpuGemmBatch, 0)
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -1831,20 +1831,17 @@ class GpuCorr3dMM_gradInputs(BaseGpuCorr3dMM):
 
 @inplace_allocempty(GpuGemv, 0)
 def local_inplace_gpuagemv(node, inputs):
-    with inherit_stack_trace(node.outputs):
-        return [gpugemv_inplace(*inputs)]
+    return [gpugemv_inplace(*inputs)]
 
 
 @inplace_allocempty(GpuGemm, 0)
 def local_inplace_gpuagemm(node, inputs):
-    with inherit_stack_trace(node.outputs):
-        return [gpugemm_inplace(*inputs)]
+    return [gpugemm_inplace(*inputs)]
 
 
 @inplace_allocempty(GpuGer, 0)
 def local_inplace_gpuager(node, inputs):
-    with inherit_stack_trace(node.outputs):
-        return [gpuger_inplace(*inputs)]
+    return [gpuger_inplace(*inputs)]
 
 
 @inplace_allocempty(GpuGemmBatch, 0)
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -7,7 +7,6 @@ from theano import Apply, Op
 from theano.compile import optdb
 from theano.gof import LocalOptGroup, ParamsType
 from theano.scalar import bool as bool_t
-from theano.gof.opt import inherit_stack_trace
 from theano.tensor.basic import as_tensor_variable
 from theano.tensor.opt import in2out
 
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -18,6 +18,7 @@ from theano.gradient import DisconnectedType, grad_not_implemented
 from theano.gof import Optimizer, local_optimizer, COp, ParamsType, EnumList
 from theano.gof.cmodule import GCC_compiler
 from theano.gof.type import CDataType, Generic
+from theano.gof.opt import inherit_stack_trace
 from theano.compile import optdb
 from theano.compile.ops import shape_i, shape_i_op
 from theano.tensor.nnet import LogSoftmax, SoftmaxGrad
@@ -3132,12 +3133,13 @@ def local_abstractconv_cudnn(node):
     ctx = infer_context_name(*node.inputs)
     if not isinstance(node.inputs[0].type, GpuArrayType):
         return
-    if node.op.unshared:
-        return None
-    if isinstance(node.op, AbstractConv2d):
-        return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
-    elif isinstance(node.op, AbstractConv3d):
-        return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+    with inherit_stack_trace(node.outputs):
+        if node.op.unshared:
+            return None
+        if isinstance(node.op, AbstractConv2d):
+            return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+        elif isinstance(node.op, AbstractConv3d):
+            return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
 
 
 @local_optimizer([AbstractConv2d, AbstractConv2d_gradWeights, AbstractConv2d_gradInputs])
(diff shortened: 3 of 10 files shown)


theano/gpuarray/opt.py
```diff
@@ -1141,7 +1156,6 @@ def local_gpua_careduce(op, context_name, inputs, outputs):
 
             reshaped_x = x.reshape(tensor.stack(new_in_shp))
             gpu_reshaped_x = as_gpuarray_variable(reshaped_x, context_name)
-            gvar = greduce(gpu_reshaped_x)
             # We need to have the make node called, otherwise the mask can
             # be None
```
abergeron: I don't understand it either, maybe @nouiz would know.

nouiz: The comment is correct, we need to keep that line. So add it back.

The problem is that the call supports_c_code need the attribute reduce_mask to be not none. And currently only a call to make_node do this. It is ugly and need refactoring, but removing this line without the refactoring isn't a good think.

cooijmanstim:
I can forge ahead and make the tests pass, but I'm not sure littering the code with unnecessary temporary variables and `copy_stack_trace`s is the right way to go about this. For instance, https://github.com/cooijmanstim/Theano/blob/661be5b763fc19d09f791ab66fa2c72d1e1c7fa9/theano/gpuarray/opt.py#L1157-L1174 would become:

```python
    stacked_in_shp = tensor.stack(new_in_shp)
    copy_stack_trace(outputs, stacked_in_shp)
    reshaped_x = x.reshape(stacked_in_shp)
    copy_stack_trace(outputs, reshaped_x)
    gpu_reshaped_x = as_gpuarray_variable(reshaped_x, context_name)
    copy_stack_trace(outputs, gpu_reshaped_x)
    # We need to have the make node called, otherwise the mask can
    # be None
    reshaped_gpu_inputs = [gpu_reshaped_x]
    if greduce.supports_c_code(reshaped_gpu_inputs):
        reduce_reshaped_x = greduce(gpu_reshaped_x)
        copy_stack_trace(outputs, reduce_reshaped_x)
    
        if reduce_reshaped_x.ndim != outputs[0].ndim:
            out_shp = []
            for i in range(x.ndim):
                if i not in op.axis:
                    out_shp.append(shape_i(x, i))
            stacked_out_shp = tensor.stack(out_shp)
            copy_stack_trace(outputs, stacked_out_shp)
            unreshaped_reduce = GpuReshape(len(out_shp))(reduce_reshaped_x,
                                                         stacked_out_shp)
        else:
            unreshaped_reduce = reduce_reshaped_x
        return [unreshaped_reduce]
```

I would find this unacceptable if I ever had to write a nontrivial optimization. One option that @lamblin and I briefly talked about is to use a contextmanager to track all nodes created inside the optimization. It's easy enough to do but it would introduce global state into Theano so I want to make sure this is okay before I write it up. Any thoughts?

theano/gpuarray/opt.py
cooijmanstim: Good catch, will undo.

theano/gpuarray/opt.py
```diff
@@ -656,6 +669,8 @@ def local_gpu_contiguous_gpu_contiguous(node):
     if isinstance(node.op, GpuContiguous):
         inp = node.inputs[0]
         if inp.owner and isinstance(inp.owner.op, GpuContiguous):
+            if not getattr(inp.tag, 'trace', None):
+                copy_stack_trace(node.outputs[0], inp)
```
cooijmanstim: Hmm. Maybe I don't need to copy the trace at all here, since the optimization doesn't add any nodes.

lamblin: Oh, right. Since `inp` already existed, no need to copy the trace.

lamblin:
So, I don't think that the variables created by `tensor.stack` (shapes) should have the stack trace of the creation  of the original CAReduce node in that case. They do not really correspond to the same variable, or even something that holds the same values.
For me, this would have two consequences:
- it lowers the number of calls to `copy_stack_trace` in the code
- it makes it harder to select which nodes created inside a context should be tagged.

The first point, combined to the fact that `copy_stack_trace` can be called once to add a trace to multiple nodes, would make the code above simpler, with only two calls to `copy_stack_trace`.
```python
    reshaped_x = x.reshape(tensor.stack(new_in_shp))
    gpu_reshaped_x = as_gpuarray_variable(reshaped_x, context_name)
    # We need to have the make node called, otherwise the mask can
    # be None
    reshaped_gpu_inputs = [gpu_reshaped_x]
    if greduce.supports_c_code(reshaped_gpu_inputs):
        reduce_reshaped_x = greduce(gpu_reshaped_x)
        copy_stack_trace(outputs, [reshaped_x, gpu_reshaped_x, reduce_reshaped_x])
    
        if reduce_reshaped_x.ndim != outputs[0].ndim:
            out_shp = []
            for i in range(x.ndim):
                if i not in op.axis:
                    out_shp.append(shape_i(x, i))
            unreshaped_reduce = GpuReshape(len(out_shp))(reduce_reshaped_x,
                                                         tensor.stack(out_shp))
             copy_stack_trace(outputs, unreshaped_reduce)
        else:
            unreshaped_reduce = reduce_reshaped_x
        return [unreshaped_reduce]
```

Also, in non-trivial optimizations, newly-created variables can correspond to different variables in the original graph.

I'm not opposed to having a context manager for cases where it would help having cleaner code, but I'm not sure this is one of these cases.

Another option might be to add another special keyword to `Op.__call__`, so that it uses some explicit stack trace. For instance, this could make us able to call `GpuReshape(len(out_shp))(reduce_reshaped_x, tensor.stack(out_shp), _stacktrace_from=outputs)` without having to assign it to a temporary variable.

nouiz:
I like the last option.

Le ven. 17 mars 2017 21:32, Pascal Lamblin <notifications@github.com> a
écrit :

cooijmanstim:
The idea of having a keyword argument on `Op.__call__` didn't cut it as nodes are rarely created by directly calling an op. Instead I defined a function `with_stack_trace(from_var, to_var)` that calls `copy_stack_trace` and returns `to_var`. This still isn't really satisfying as sometimes `make_node` creates intermediate nodes that may also need stack traces; I think having a context manager that is explicitly used by the optimization (rather than by a general decorator like `op_lifter`) should solve this.

The next problem is that although the trace seems to be copied over correctly, often the original nodes didn't have a trace and so my tests fail. In one or two cases I drilled down to put make sure the original nodes have traces, and some other cases I just skip the test, but there are still failing tests.

nouiz:
this PR need a rebase.

Do you suggest using a context manager inside op lifter that will mark all created variable with a given stack trace? If so, such similar feature would be reused elsewhere too.

I didn't follow this PR, but in case you don't know, not only op_lifter lift node to the GPU. There is also GraphToGPU. It would need the same "context manager" then op_lifter.

theano/gpuarray/tests/test_opt.py
```diff
@@ -23,6 +23,16 @@
...
 
 
+def _check_stack_trace(thing):
+    def _ops_to_check(op):
+        if not isinstance(op, theano.gof.Op):
+            op = op.op # assume node
+        return not isinstance(op, (theano.compile.ops.Shape_i,
+                                   theano.ifelse.IfElse,
+                                   GpuFromHost, HostFromGpu,
+                                   GpuElemwise))
```
lamblin: Why exclude GpuElemwise? Is is just temporary?

cooijmanstim: Definitely temporary. Probably the reason was that GpuElemwise creates a bunch of nodes in its make_node that it's not obvious how to copy the stack traces to. Also I believe the original Elemwise that is replaced by the optimization does not have a stack trace to begin with.

cooijmanstim:
I've rebased and implemented the context manager I was talking about. I'm tempted to just use it everywhere instead of selectively copying traces over. (But the nice thing about being selective is that I can copy traces over one for one, as e.g. in https://github.com/Theano/Theano/pull/5688/files#diff-ba9b2bcf5eab82177b89e680c96a55c3R244.)

In order to make the tests pass I've also gone and tracked down places outside of theano.gpuarray where stack traces were getting lost. There's still quite a few things that don't have traces that really should, like some instances of GpuElemwise. Not sure how far I should go with this.

theano/tensor/blas.py
```diff
@@ -1618,19 +1619,20 @@ def local_dot_to_dot22(node):
...
+            if x.ndim == 2 and y.ndim == 1:
+                # print "local_dot_to_dot22: MV"
+                return [_dot22(x, y.dimshuffle(0, 'x')).dimshuffle(0)]
+            if x.ndim == 1 and y.ndim == 2:
+                # print "local_dot_to_dot22: VM"
+                return [_dot22(x.dimshuffle('x', 0), y).dimshuffle(1)]
+            if x.ndim == 1 and y.ndim == 1:
+                # print "local_dot_to_dot22: VV"
+                return [_dot22(x.dimshuffle('x', 0),
+                               y.dimshuffle(0, 'x')).dimshuffle()]
```
lamblin: While you're at it, can you remove the commented-out `print`s?

theano/tensor/blas.py
```diff
@@ -1639,32 +1641,36 @@ def local_dot_to_dot22(node):
...
         z, a, x, y, b = node.inputs
-        if z.broadcastable == x.broadcastable == (True, False):
-            r = gemv_no_inplace(z.dimshuffle(1), a, y.T, x.dimshuffle(1), b)
-            return [r.dimshuffle('x', 0)]
-        if z.broadcastable == y.broadcastable == (False, True):
-            r = gemv_no_inplace(z.dimshuffle(0), a, x, y.dimshuffle(0), b)
-            return [r.dimshuffle(0, 'x')]
+        with inherit_stack_trace(node.outputs):
+            if z.broadcastable == x.broadcastable == (True, False):
+                r = gemv_no_inplace(z.dimshuffle(1), a, y.T, x.dimshuffle(1), b)
```
lamblin: It may not be optimal, tagging things like y.T and the other dimshuffles with the stack of gemm, but I guess it's OK as they're unlikely to fail.

lamblin: Also, we could argue that the transpose operations are implicitly added when calling `dot()` in the original code.
So I'm OK with it.

theano/gpuarray/opt.py
```diff
@@ -398,6 +409,10 @@ def apply(self, fgraph):
             elif isinstance(new_ops, theano.Variable):
                 outputs = [new_ops]
 
+            for old_output, new_output in zip(node.outputs, outputs):
+                copy_stack_trace(old_output, new_output)
+                new_output.tag.tracefrom = old_output
+
```
cooijmanstim: Whoops, that's debugging code I didn't mean to commit. It allows my to track back where the lack of trace comes from.

nouiz:
travis failed

theano/gpuarray/opt.py
ReyhaneAskari: Done.

theano/tensor/blas.py
ReyhaneAskari: Done.

theano/gpuarray/opt.py
```diff
@@ -231,12 +232,22 @@ def local_opt(node):
...
-                        return [new_op.transfer('cpu')]
+                        new_outputs = [new_op]
+                        to_cpu_fn = lambda x: x.transfer('cpu')
+                    # copy stack traces onto gpu outputs
+                    for old_output, new_output in zip(node.outputs, new_outputs):
+                        copy_stack_trace(old_output, new_output)
+                    new_outputs = [to_cpu_fn(o) for o in new_outputs]
+                    # also copy the stack traces onto HostFromGpu outputs
+                    for old_output, new_output in zip(node.outputs, new_outputs):
+                        copy_stack_trace(old_output, new_output)
```
ReyhaneAskari: Done.

theano/gpuarray/dnn.py
```diff
@@ -2838,47 +2839,53 @@ def local_abstractconv_cudnn(node):
...
-    elif isinstance(node.op, AbstractConv3d_gradInputs):
-        return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+    with inherit_stack_trace(node.outputs):
+        if isinstance(node.op, AbstractConv2d_gradInputs):
+            return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+        elif isinstance(node.op, AbstractConv3d_gradInputs):
+            return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
 
 
 @inplace_allocempty(GpuDnnConv, 2)
```
ReyhaneAskari: Done.

theano/gpuarray/dnn.py
```diff
@@ -2891,40 +2898,46 @@ def local_dnn_convgi_inplace(node, inputs):
 @register_opt('cudnn')
 @alpha_merge(GpuDnnConv, alpha_in=4, beta_in=5)
 def local_dnn_conv_alpha_merge(node, *inputs):
-    return [GpuDnnConv(algo=node.op.algo)(*inputs)]
+    with inherit_stack_trace(node.outputs):
```
ReyhaneAskari: Done.

ReyhaneAskari:
After the tests passed, I will run the tests by setting the ``check_stack_trace `` to `raise` to see which GPU optimizations still have empty stack traces.

ReyhaneAskari:
Jenkins test this.

nouiz:
jenkins test this

nouiz:
jenkins test this

ReyhaneAskari:
jenkins test this

nouiz:
There was WIP in the title of this PR, jenkins skip PR in that case.

ReyhaneAskari:
Thanks, I just realized jenkins is too smart! Simon told me to removed the other WIP in the first comment then it started working.

ReyhaneAskari:
I ran the tests with setting the flag ``check_stack_trace `` to `raise` but there was no change in the results. [Here is a list of](https://github.com/Theano/Theano/files/1252280/list_out_25_aug.txt) optimizations that have empty stack traces that I got by parsing [this file](https://github.com/Theano/Theano/files/1252269/out_25_aug.txt).

ReyhaneAskari:
jenkins test this.

ReyhaneAskari:
Did a rebase due to the conflicts.

ReyhaneAskari:
jenkins test this.

nouiz:
jenkins have failure.

nouiz:
jenkins test this

theano/gpuarray/blas.py
```diff
@@ -1723,17 +1724,20 @@ def connection_pattern(self, node):
 
 @inplace_allocempty(GpuGemv, 0)
 def local_inplace_gpuagemv(node, inputs):
-    return [gpugemv_inplace(*inputs)]
+    with inherit_stack_trace(node.outputs):
```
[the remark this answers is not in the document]
nouiz: done

theano/gpuarray/opt.py
```diff
@@ -252,12 +253,22 @@ def local_opt(node):
...
+                        new_outputs = [new_op]
+
+                        def to_cpu_fn(x):
+                            return x.transfer('cpu')
+                    # copy stack traces onto gpu outputs
+                    # also copy the stack traces onto HostFromGpu outputs
+                    for old_output, new_output in zip(node.outputs, new_outputs):
+                        copy_stack_trace(old_output, new_output)
+                        copy_stack_trace(old_output, to_cpu_fn(new_output))
+                    return new_outputs
```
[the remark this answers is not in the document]
nouiz: done

theano/gpuarray/opt.py
```diff
@@ -1305,57 +1322,62 @@ def local_gpua_gemm(op, context_name, inputs, outputs):
...
+        if len(output_dims) != 3:
+            out = GpuDimShuffle(out.broadcastable, output_dims)(out)
+        return out
 
 
 @register_opt()
 @alpha_merge(GpuGemm, alpha_in=1, beta_in=4)
 def local_gpua_gemm_alpha_merge(node, *inputs):
-    return [gpugemm_no_inplace(*inputs)]
+    with inherit_stack_trace(node.outputs):
```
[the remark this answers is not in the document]
nouiz: done

theano/gof/opt.py
```diff
@@ -2946,6 +2947,24 @@ def copy_stack_trace(from_var, to_var):
...
+    Contextmanager that copies the stack trace from one or more tensor variables to all tensor
+    variables constructed in the body.
+
+    Parameters
+    ----------
+    from_var
+        Tensor variable or list of tensor variables to copy stack traces from.
+
+    """
+    with graph.nodes_constructed() as new_nodes:
```
nouiz: Having a context manager inside a context manager make it more complicated to understand how this all work. We can keep it like this, but I think it should be explained somewhere in a follow up PR.

theano/gpuarray/dnn.py
```diff
@@ -3132,12 +3133,13 @@ def local_abstractconv_cudnn(node):
...
         return
-    if node.op.unshared:
-        return None
-    if isinstance(node.op, AbstractConv2d):
-        return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
-    elif isinstance(node.op, AbstractConv3d):
-        return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+    with inherit_stack_trace(node.outputs):
+        if node.op.unshared:
+            return None
```
nouiz: Ideally moce this if outside the with statement.

theano/gpuarray/dnn.py
```diff
@@ -3356,25 +3358,27 @@ def local_abstractconv_gw_cudnn(node):
...
         return
-    if node.op.unshared:
-        return None
-    if isinstance(node.op, AbstractConv2d_gradWeights):
-        return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
-    elif isinstance(node.op, AbstractConv3d_gradWeights):
-        return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+    with inherit_stack_trace(node.outputs):
+        if node.op.unshared:
+            return None
```

theano/gpuarray/dnn.py
```diff
@@ -3356,25 +3358,27 @@ def local_abstractconv_gw_cudnn(node):
...
         return
-    if node.op.unshared:
-        return None
-    if isinstance(node.op, AbstractConv2d_gradInputs):
-        return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
-    elif isinstance(node.op, AbstractConv3d_gradInputs):
-        return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
+    with inherit_stack_trace(node.outputs):
+        if node.op.unshared:
+            return None
```

theano/gpuarray/opt.py
```diff
@@ -252,12 +253,25 @@ def local_opt(node):
                 # This is needed as sometimes new_op inherits from OP.
                 if new_op and new_op != node.op:
                     if isinstance(new_op, theano.Op):
```
nouiz: Should the call to maker on line 251 should be in a "with inherit_stack_trace", as sometimes it will build intermediate nodes. For example see the local_gpua_careduce optimization

I don't know if we should do that only when there is only 1 outputs (or all outputs have the same stack trace) or we should always do that and use the first stack trace? Currently this PR modify the local_gpua_careduce opt to manually add the stack trace. So this is more for a follow up PR.

theano/gof/opt.py
ReyhaneAskari: Yes, exactly. I would try to write a more readable form of this.

theano/gof/opt.py
```diff
@@ -2948,6 +2949,52 @@ def copy_stack_trace(from_var, to_var):
...
+        yield
+    copy_stack_trace(from_var, new_nodes)
```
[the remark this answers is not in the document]
ReyhaneAskari: Do you mean that new_nodes is defined after the with statement?

theano/gpuarray/dnn.py
ReyhaneAskari: That's true. Sure.

ReyhaneAskari: This is actually a `if .. else` block:
```python
with inherit_stack_trace(node.outputs):
        if node.op.unshared:
            return None
        if isinstance(node.op, AbstractConv2d):
            return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
        elif isinstance(node.op, AbstractConv3d):
            return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
```

So if I want to move the if outside the with, I will have to do something like this:

```python
if node.op.unshared:
	with inherit_stack_trace(node.outputs):
                return None
if isinstance(node.op, AbstractConv2d):
	with inherit_stack_trace(node.outputs):
               return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
elif isinstance(node.op, AbstractConv3d):
	with inherit_stack_trace(node.outputs):
                return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
```

nouiz: What about
```
if node.op.unshared:
        return None
if isinstance(node.op, AbstractConv2d):
	with inherit_stack_trace(node.outputs):
               return local_abstractconv_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
elif isinstance(node.op, AbstractConv3d):
	with inherit_stack_trace(node.outputs):
                return local_abstractconv3d_cudnn_graph(node.op, ctx, node.inputs, node.outputs)
```

As this is an optimization, when it return none, it mean it do nothing. So no new node build, do no new stack trace needed.

ReyhaneAskari: I see! Thanks.

theano/gof/opt.py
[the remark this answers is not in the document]
nouiz: Yes, this is what make this code work. Otherwise it would raise an error that new_nodes isn't defined.

Python 3 changed the scope of list comprehensions. I didn't know what the scope was for with statement or if it changed between Python2 and 3. I checked all is fine. It didn't changed and the var defined on the line of the with statement still exist after the with statement.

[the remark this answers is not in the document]
ReyhaneAskari: I see. Thanks for the explanations.

nouiz:
jenkins test this

nouiz:
jenkins test this

Issue #5094, opened 2016-10-14
Opt: IncSubtensor(zeros, x, ...) -> SetSubtensor(zeros, x,  ...)


nouiz:
See https://github.com/Theano/Theano/pull/5086/commits/1649d3eac9742347254d6de512954f3767233e56

I suppose this can stay in the final graph if there is no add.

lamblin:
I'm not convinced there is anything to gain from doing that.
Apart maybe in the case where we have both inc_subtensor and set_subtensor with the same indices, then they could be merged.

nouiz:
I think you didn't understand what I want by this optimization.
The goal is to transforme  IncSubtensor(zeros, x , slices/ints) to
SetSubtensor(zeros, x slices/ints)

There is no merge needed. This is a local opt on only 1 node with the
condition that the source is zeros.

This will save execution time as we won't need to read the zeros, we will
only overwrite them. I didn't saw this as a bottleneck, but I saw this in
some graph. I don't remember which graph, it was maybe tests, not real
model. This is low priority.

lamblin:
OK, let me clarify.
I don't think that reading the zeros will make a measurable difference in practice. In the inplace case, we will have to write to that memory anyway, in the no-inplace case the rest of the array will need to be copied.
However, if for some reason there is both `set_subtensor(zeros[idx], x)` and `inc_subtensor(zeros[idx], x)` in the same graph, then having a standard form may help merge them (and potentially further computation as well). I have no idea if it happens in real life, though.

PR #5690, opened 2017-03-10
Opt: Inc(zeros, x) --> Set(zeros, x)

Faruk-Ahmed:
#5094

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -3411,6 +3411,24 @@ def local_incsubtensor_of_zeros(node):
             return
 
 
+@register_canonicalize
+@register_specialize
+@gof.local_optimizer([IncSubtensor])
+def incsubtensor_of_zeros_to_setsubtensor(node):
+    """
+    IncSubtensor(zeros, x, ...) -> SetSubtensor(zeros, x, ...)
+    """
+    if (isinstance(node.op, (IncSubtensor)) and not node.op.set_instead_of_inc):
+        x = node.inputs[0]
+
+        if isinstance(x, T.Constant) and not numpy.any(x.data):
+            return [IncSubtensor(node.op.idx_list,
+                                 node.op.inplace,
+                                 set_instead_of_inc=True,
+                                 destroyhandler_tolerate_aliased=node.op.destroyhandler_tolerate_aliased,
+                                 )(*node.inputs)]
+
+
 @register_canonicalize('local_setsubtensor_of_allocs')
 @register_stabilize('local_setsubtensor_of_allocs')
 @gof.local_optimizer([IncSubtensor])
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -3001,6 +3001,37 @@ class Test_alloc_zero(unittest.TestCase):
         assert np.all([not isinstance(n.op, tensor.IncSubtensor)
                        for n in f.maker.fgraph.toposort()])
 
+    def test_incsubtensor_x_zeros(self):
+        x = tensor.constant(np.asarray(np.zeros((4, 4)),
+                                       dtype=config.floatX))
+        y = tensor.matrix()
+        z = tensor.inc_subtensor(x[:4], y)
+        f = theano.function([y], z)
+        inc_nodes = [n for n in f.maker.fgraph.toposort()
+                     if isinstance(n.op, tensor.IncSubtensor)]
+
+        assert(len(inc_nodes) == 1)
+        node_is_set_instead_of_inc = inc_nodes[0].op.set_instead_of_inc
+        mode = theano.config.mode
+        assert((mode != "FAST_COMPILE" and node_is_set_instead_of_inc) or
+               (mode == "FAST_COMPILE" and not node_is_set_instead_of_inc))
+        test_X = np.random.random((4, 4)).astype(config.floatX)
+        utt.assert_allclose(f(test_X), test_X)
+
+        # also check the flag doesn't get set if first input is not zeros:
+        not_all_zeros = np.zeros((4, 4))
+        not_all_zeros[1, 0] = 0.001
+        x = tensor.constant(np.asarray(not_all_zeros, dtype=config.floatX))
+        y = tensor.matrix()
+        z = tensor.inc_subtensor(x[:4], y)
+        f = theano.function([y], z)
+        inc_nodes = [n for n in f.maker.fgraph.toposort()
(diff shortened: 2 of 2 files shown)


theano/tensor/opt.py
```diff
@@ -3360,6 +3360,23 @@ def local_incsubtensor_of_zeros(node):
...
+def incsubtensor_of_zeros_to_setsubtensor(node):
+    """
+    IncSubtensor(zeros, x, ...) -> SetSubtensor(zeros, x, ...)
+
+    """
+    if (isinstance(node.op, (IncSubtensor)) and not node.op.set_instead_of_inc):
+        x = node.inputs[0]
+        y = node.inputs[1]
+
+        if isinstance(x, T.zeros):
```
khaotik: `T.zeros` is not a class, just a convenience function

theano/tensor/opt.py
```diff
@@ -3360,6 +3360,23 @@ def local_incsubtensor_of_zeros(node):
...
+            node.op.set_instead_of_inc = True
```
nouiz: You should not modify an op instance. You should create a new one with the new parameter values.

theano/tensor/opt.py
```diff
@@ -3360,6 +3360,23 @@ def local_incsubtensor_of_zeros(node):
...
+        return node
```
nouiz: You should return a Theanno variable, not a new Apply node.

It should be something like:

    return IncSubtensor(node.op.idx_list, set_instead_of_inc=True, node.op.idx_list)(*node.inputs)

nouiz:
Both fixed idea are good. Just select the one you prefer to implement.

Le dim. 19 mars 2017 23:36, Xavier Bouthillier <notifications@github.com> a
écrit :

theano/tensor/tests/test_opt.py
```diff
@@ -2995,6 +2995,30 @@ def test_incsubtensor_allocs1(self):
...
+    def test_incsubtensor_x_zeros(self):
+        x = tensor.constant(numpy.asarray(numpy.zeros((4, 4)),
+                                           dtype=config.floatX))
+        y = tensor.matrix()
+        z = tensor.inc_subtensor(x[:4], y)
+        f = theano.function([y], z, mode=self.mode)
+        inc_nodes = [n for n in f.maker.fgraph.toposort() 
+                     if isinstance(n.op, tensor.IncSubtensor)]
+        assert(len(inc_nodes) == 1)
+        assert(inc_nodes[0].op.set_instead_of_inc is True)
```
bouthilx: > Both fixed idea are good. Just select the one you prefer to implement.

That should do it then.

```python
node_is_set_instead_of_inc = inc_nodes[0].op.set_instead_of_inc
assert((mode != "FAST_COMPILE" and node_is_set_instead_of_inc) or
       (mode == "FAST_COMPILE" and not node_is_set_instead_of_inc))
```

theano/tensor/opt.py
```diff
@@ -3360,6 +3360,25 @@ def local_incsubtensor_of_zeros(node):
             return
 
 
+@register_canonicalize
+@register_stabilize
```
nouiz: This isn't a stabilization opt. Is should be in register_specialize instead of register_stabilize

theano/tensor/tests/test_opt.py
```diff
@@ -2995,6 +2995,33 @@ def test_incsubtensor_allocs1(self):
...
+        not_all_zeros = numpy.zeros((4, 4))
+        not_all_zeros[2, 3] = 0.001
+        x = tensor.constant(numpy.asarray(not_all_zeros, dtype=config.floatX))
+        y = tensor.matrix()
+        z = tensor.inc_subtensor(x[:4], y)
+        f = theano.function([y], z)
+        inc_nodes = [n for n in f.maker.fgraph.toposort()
+                     if isinstance(n.op, tensor.IncSubtensor)]
+        assert(len(inc_nodes) == 1)
+        assert(inc_nodes[0].op.set_instead_of_inc is False)
```
nouiz: Execute both test function and check the result. Just in case, if we compile executing them won't add much tests time.

theano/tensor/tests/test_opt.py
```diff
@@ -3021,6 +3023,8 @@ def test_incsubtensor_x_zeros(self):
                      if isinstance(n.op, tensor.IncSubtensor)]
         assert(len(inc_nodes) == 1)
         assert(inc_nodes[0].op.set_instead_of_inc is False)
+        test_X = numpy.random.random((4, 4)).astype(theano.config.floatX)
+        assert(numpy.any(f(test_X) == test_X))
```
bouthilx: @Faruk-Ahmed  numpy.any? A little slip? 

Anyway it would be better to use utt.assert_allclose than numpy.all.

theano/tensor/tests/test_opt.py
```diff
@@ -3023,8 +3023,8 @@ def test_incsubtensor_x_zeros(self):
                      if isinstance(n.op, tensor.IncSubtensor)]
         assert(len(inc_nodes) == 1)
         assert(inc_nodes[0].op.set_instead_of_inc is False)
-        test_X = numpy.random.random((4, 4)).astype(theano.config.floatX)
-        assert(numpy.any(f(test_X) == test_X))
+        test_X = numpy.random.random((4, 4)).astype(config.floatX)
+        utt.assert_allclose(f(test_X), test_X+not_all_zeros)
```
bouthilx: pep8

bouthilx:
@nouiz Do you have any Idea why flake8 raises ```F821 undefined name 'numpy'```? I can clearly see ```import numpy``` at the beggining of the file.

lamblin:
There have been changes to change `import numpy` to `import numpy as np` and use `np` throughout the files. I guess that one such PR has been merged recently.
The solution would be a rebase, and then using `np`.

bouthilx:
@lamblin Thanks Pascal! I failed to notice this, only looking at the files from Faruk's branch. I'm suprised ```ìmport numpy as np``` was introduced everywhere. Aren't we trying to avoid shortcuts like ```np``` and ```T``` and rather use ```numpy``` and ```tensor```? I might mix up with blocks coding style.

lamblin:
That was a seriously strange rebase...
Did you rebase the current master on top of your PR (and not the other way around)?

Faruk-Ahmed:
switching to #5785

lamblin:
@bouthilx Since numpy themselves use `np` everywhere, we decided to switch that way.
For `T`, I agree that it should be `tensor` instead, but we did not make any move towards that yet.

PR #5693, opened 2017-03-10
Various small fixes.

lamblin:
Should we back-port 3c0b51f to 0.9?

### DIFF
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -79,29 +79,6 @@ utt.seed_rng()
 test_rng = numpy.random.RandomState(seed=utt.fetch_seed())
 
 
-# A helpful class to check random values close to the boundaries
-# when designing new tests
-class MockRandomState:
-    def __init__(self, val):
-        self.val = val
-
-    def rand(self, *shape):
-        return numpy.zeros(shape, dtype='float64') + self.val
-
-    def randint(self, minval, maxval=None, size=1):
-        if maxval is None:
-            minval, maxval = 0, minval
-        out = numpy.zeros(size, dtype='int64')
-        if self.val == 0:
-            return out + minval
-        else:
-            return out + maxval - 1
-# Examples of use:
-# test_rng = MockRandomState(0)
-# test_rng = MockRandomState(0.99999982)
-# test_rng = MockRandomState(1)
-
-
 if PY3:
     def L(i):
         return i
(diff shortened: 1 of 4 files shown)


theano/tensor/tests/test_basic.py
```diff
@@ -73,8 +73,34 @@
...
-### seed random number generator so that unittests are deterministic ###
+
+# Use a seeded random number generator so that unittests are deterministic
 utt.seed_rng()
+test_rng = numpy.random.RandomState(seed=utt.fetch_seed())
+
+
+# A helpful class to check random values close to the boundaries
+# when designing new tests
+class MockRandomState:
```
abergeron: This isn't used anywhere so maybe we shouldn't add it yet.  Also, if you do add it make it inherit from object.

lamblin: It helped me detect and fix issues, but it is not a replacement that should be used in general (there are issues with minimum, maximum, for instance, because all values are the same, which would not happen with actual random samples).
So I figured it would be useful to have that code here. Do you have a better idea of where to put it, so it is ready to use when adding new tests?

theano/tensor/tests/test_basic.py
```diff
@@ -1629,7 +1658,7 @@ def test_py_c_match():
     uint8=[numpy.arange(1, 256, dtype='uint8')],
     complex=(randc128_ranged(1, 1000, (2, 3)),),
     empty=(numpy.asarray([], dtype=config.floatX),),)
-_grad_broadcast_unary_arccosh = dict(normal=(rand_ranged(1, 1000, (2, 3)),),)
+_grad_broadcast_unary_arccosh = dict(normal=(rand_ranged(1 + _eps, 1000, (2, 3)),),)
```
nouiz: I don't think this one is useful, but it won't hurt the test quality. So I'll merge it.

theano/tensor/tests/test_basic.py
nouiz: I'm pushing in this PR to allow merging this rapidly for rc4. I'm moving this to theano/tests/unittest_tools.py

If CI pass, I'll merge tomorrow morning.

theano/tensor/tests/test_basic.py
lamblin: Actually, if the value sampled was too close to 1, `verify_grad` would fail. It may happen rarely in practice, but it would eventually.

PR #5704, opened 2017-03-13
Add pathed version of theano_logo_allblue.svg

akx:
The original theano_logo_allblue.svg image uses the DejaVu Mono font, so rendering is broken on systems which don't have that font. This version has the text flattened into a path.


abergeron:
I'm not against adding a pathed version, but since we don't have anything using it right now, I fail to see the benefit.

lamblin:
I guess if someone wants to credit Theano in a poster made on Windows, or something, it might be useful.

Issue #5696, opened 2017-03-10
Questions about Theano C code syntax


notoraptor:
As it is planned to reduce variants of C codes generated by Theano, I suggest to use this ticket to centralize all questions about that ! @abergeron @nouiz @lamblin 

---

### (1) Some C codes use a typedef `dtype_%(var)s`, but it is not guaranteed to be defined. 

For example, here: https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L2863

Code assumes there is a type called `dtype_%(shp_i)s` based on the name of an input `shp_i`. Indeed, this type is defined via a typedef in `TensorType.c_declare()`: https://github.com/Theano/Theano/blob/master/theano/tensor/type.py#L385

But the typedef is declared only if `c_declare()` is called with `check_input=True.`

I suggest to create this typedef in any case, whatever be the value of `check_input`.

abergeron:
You should notice that check_input=True is the default.  So most ops will not have a problem with that.

If an op decides to use check_input=False then the defines will not be present but that is part of the option.  This will also allow a reduction in the variants of C code generated for ops that do not specifically care about the type of their inputs.

So in all I am against what you propose.

notoraptor:
@abergeron Ok!

I have another question:
---
### (2) Some literal strings are directly compared in some C codes.

Example here, in `theano.gpuarray.neighbours.GpuImages2Neibs` : https://github.com/Theano/Theano/blob/master/theano/gpuarray/neighbours.py#L316

A literal string is directly compared to another literal string. It should not work, as this does not compare strings, but instead it compares memory addresses. It does not seem to create any error in tests, and I have read that some compilers are smart enough to store a literal string in one address and reuse it everywhere this literal is used, so maybe that's what g++ does here.

Should we be more strict here, to prevent any future problem  ? Or isn't it so problematic ?

abergeron:
Yes, this is bad and should be avoided.  This should be replaced with an int or a set of defines.

notoraptor:
### (3) How to manage vectors or arrays props into C code ?

Example with `theano.gpuarray.dnn.GpuDnnPoolDesc`: https://github.com/Theano/Theano/blob/master/theano/gpuarray/dnn.py#L1217

[...]

PR #5710, opened 2017-03-14
Implement new op param type `EnumType`

notoraptor:
This PR suggests a new abstract attribute `enum` added into COp to define enumerations of constants that will be automatically converted into C macros by `COp.get_op_params()` method.

This should help to reduce C code of some C-ops, for example with GpuDnnSoftmax, as we could then define a fixed-list of constants (that won't change from an op instance to another), and then use an op param that takes a runtime value within this list of constants.

The PR also adds two functions to help create enums quickly, and a test for this new feature.

@nouiz @abergeron  @lamblin

### DIFF
--- a/theano/gof/tests/test_types.py
+++ b/theano/gof/tests/test_types.py
@@ -2,9 +2,9 @@ from __future__ import absolute_import, print_function, division
 import numpy as np
 
 import theano
-from theano import Op, Apply
+from theano import Op, Apply, scalar
 from theano.tensor import TensorType
-from theano.gof.type import CDataType
+from theano.gof.type import CDataType, EnumType, EnumList
 
 from nose.plugins.skip import SkipTest
 
@@ -76,3 +76,118 @@ def test_cdata():
 
     v2 = f(v)
     assert (v2 == v).all()
+
+
+class TestOpEnumList(Op):
+    __props__ = ('op_chosen',)
+    params_type = EnumList('ADD', 'SUB', 'MULTIPLY', 'DIVIDE')
+
+    def __init__(self, choose_op):
+        assert self.params_type.ADD == 0
+        assert self.params_type.SUB == 1
+        assert self.params_type.MULTIPLY == 2
+        assert self.params_type.DIVIDE == 3
+        op_to_const = {'+': self.params_type.ADD,
+                       '-': self.params_type.SUB,
+                       '*': self.params_type.MULTIPLY,
+                       '/': self.params_type.DIVIDE}
+        self.op_chosen = op_to_const[choose_op]
+
+    def get_params(self, node):
+        return self.op_chosen
+
+    def make_node(self, a, b):
+        return Apply(self, [scalar.as_scalar(a), scalar.as_scalar(b)], [scalar.float64()])
+
+    def perform(self, node, inputs, outputs, op):
+        a, b = inputs
+        o, = outputs
+        if op == self.params_type.ADD:
+            o[0] = a + b
+        elif op == self.params_type.SUB:
+            o[0] = a - b
+        elif op == self.params_type.MULTIPLY:
+            o[0] = a * b
+        elif op == self.params_type.DIVIDE:
+            o[0] = a / b
+        else:
+            raise NotImplementedError('Unknown op id ' + str(op))
+
+    def c_code_cache_version(self):
+        return (1,)
+
+    def c_code(self, node, name, inputs, outputs, sub):
+        a, b = inputs
+        o, = outputs
+        fail = sub['fail']
+        op = sub['params']
+        return """
+        switch((int)%(op)s) {
+            case ADD:
+                %(o)s = %(a)s + %(b)s;
+                break;
+            case SUB:
+                %(o)s = %(a)s - %(b)s;
+                break;
+            case MULTIPLY:
+                %(o)s = %(a)s * %(b)s;
+                break;
+            case DIVIDE:
+                %(o)s = %(a)s / %(b)s;
+                break;
+            default:
+                {%(fail)s}
+                break;
+        }
+        """ % locals()
+
+
+def test_enum():
+    # Check that invalid enum name raises exception.
+    for invalid_name in ('a', '_A', '0'):
+        try:
+            EnumList(invalid_name)
+        except AttributeError:
+            pass
+        else:
+            raise Exception('EnumList with invalid name should faild.')
+
+        try:
+            EnumType(**{invalid_name: 0})
+        except AttributeError:
+            pass
+        else:
+            raise Exception('EnumType with invalid name should fail.')
+
+    # Check that invalid enum value raises exception.
+    try:
+        EnumType(INVALID_VALUE='string is not allowe.')
+    except ValueError:
+        pass
+    else:
+        raise Exception('EnumType with invalid value should fail.')
+
+    # Check EnumType.
+    e1 = EnumType(C1=True, C2=12, C3=True, C4=-1, C5=False, C6=0.0)
+    e2 = EnumType(C1=1, C2=12, C3=1, C4=-1.0, C5=0.0, C6=0)
+    assert e1 == e2
+    assert not (e1 != e2)
+    assert hash(e1) == hash(e2)
+
+    # Test an op with EnumList.
+    a = scalar.int32()
+    b = scalar.int32()
+    c_add = TestOpEnumList('+')(a, b)
+    c_sub = TestOpEnumList('-')(a, b)
+    c_multiply = TestOpEnumList('*')(a, b)
+    c_divide = TestOpEnumList('/')(a, b)
+    f = theano.function([a, b], [c_add, c_sub, c_multiply, c_divide])
+    va = 12
+    vb = 15
+    ref_add = va + vb
+    ref_sub = va - vb
+    ref_multiply = va * vb
+    ref_divide = va // vb
+    ref = [ref_add, ref_sub, ref_multiply, ref_divide]
+    out = f(va, vb)
+    assert ref == out, (ref, out)
--- a/theano/gof/tests/test_types.py
+++ b/theano/gof/tests/test_types.py
@@ -117,7 +117,7 @@ class TestOpEnumList(Op):
             raise NotImplementedError('Unknown op id ' + str(op))
 
     def c_code_cache_version(self):
-        return None
+        return (1,)
 
     def c_code(self, node, name, inputs, outputs, sub):
         a, b = inputs
@@ -125,7 +125,7 @@ class TestOpEnumList(Op):
         fail = sub['fail']
         op = sub['params']
         return """
-        switch((int)%(op)s) {
+        switch(%(op)s) {
             case ADD:
                 %(o)s = %(a)s + %(b)s;
                 break;
@@ -175,14 +175,14 @@ class TestOpCEnumType(Op):
         return Apply(self, [], [scalar.uint32()])
 
     def c_code_cache_version(self):
-        return None
+        return (1,)
 
     def c_code(self, node, name, inputs, outputs, sub):
         o, =  outputs
-        ctype_index = sub['params']
+        # params in C code will already contains expected C constant value.
+        sizeof_ctype = sub['params']
         return """
-        /* ctype_index already contains expected C constant value. */
-        %(o)s = %(ctype_index)s;
+        %(o)s = %(sizeof_ctype)s;
         """ % locals()
 
 
@@ -205,7 +205,7 @@ def test_enum_class():
 
     # Check that invalid enum value raises exception.
     try:
-        EnumType(INVALID_VALUE='string is not allowe.')
+        EnumType(INVALID_VALUE='string is not allowed.')
     except ValueError:
         pass
     else:
@@ -220,7 +220,6 @@ def test_enum_class():
 
 
 def test_op_with_enumlist():
-    # Test an op with EnumList.
     a = scalar.int32()
     b = scalar.int32()
     c_add = TestOpEnumList('+')(a, b)
@@ -241,4 +240,4 @@ def test_op_with_cenumtype():
     sizeof_long_long = TestOpCEnumType('long long')()
     f = theano.function([], [sizeof_int, sizeof_float, sizeof_long_long])
     out = f()
-    print('(sizeof(int): ', out[0], ', sizeof(float): ', out[1], ', sizeof(long long): ', out[2], ') ', sep='', end='')
+    print('(sizeof(int): ', out[0], ', sizeof(float): ', out[1], ', sizeof(long long): ', out[2], ') ', sep='')
--- a/theano/gof/tests/test_types.py
+++ b/theano/gof/tests/test_types.py
@@ -4,7 +4,7 @@ import numpy as np
 import theano
 from theano import Op, Apply, scalar
 from theano.tensor import TensorType
-from theano.gof.type import CDataType, EnumType, EnumList
+from theano.gof.type import CDataType, EnumType, EnumList, CEnumType
 
 from nose.plugins.skip import SkipTest
 
@@ -80,7 +80,7 @@ def test_cdata():
 
 class TestOpEnumList(Op):
     __props__ = ('op_chosen',)
-    params_type = EnumList('ADD', 'SUB', 'MULTIPLY', 'DIVIDE')
+    params_type = EnumList('ADD', 'SUB', 'MULTIPLY', 'DIVIDE', ctype='unsigned long long')
 
     def __init__(self, choose_op):
         assert self.params_type.ADD == 0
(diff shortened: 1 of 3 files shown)


theano/gof/op.py
```diff
@@ -1377,7 +1436,12 @@ def get_op_params(self):
         The names must be strings that are not a C keyword and the
         values must be strings of literal C representations.
 
+        If your COp defines an attribute `enum` created from ``enum_list()``
+        or ``enum_type()``, then this method will try to deduce macros from
+        the enum fields.
         """
+        if hasattr(self, 'enum') and hasattr(self.enum, '__dict__'):
+            return [(k, str(v)) for (k, v) in self.enum.__dict__.items()]
```
abergeron: This will break in horrible and interesting ways since a \_\_dict__ can contain all sorts of things and we don't check the type of the field.

Also, this will also break for any op that overrides get_op_params() like they are supposed to do.

Finally get_op_params() is a horrible name and is/was a temporary stopgap measure before we could pass multiple things with the params argument.  Now that we can, we should phase it out.

theano/gof/op.py
```diff
@@ -1236,6 +1237,64 @@ def f(self, node, name):
...
+    and attributes values are ranks of these attributes into args
+    (starting from 0).
+
+    Example::
+
+        my_enum = enum_list('CONSTANT_NAME_1', 'CONSTANT_NAME_2', 'CONSTANT_NAME_N')
+
+    :param args: list of constants names (strings).
+    :return: a namedtuple which attributes names are those listed into args and
+    value of i-th attribute into args is integer i-1 (values from 0 to len(args)-1).
```
abergeron: Use numpydoc format for this.

theano/gof/op.py
notoraptor: Sorry, I had forgotten that using `__dict__` works correctly only if enum is a `namedtuple` (in that case, a_named_tuple.__dict__ contains only attributes we define). Indeed, this code is unsafe.

nouiz:
I don't understand what this PR tries to help. I think it is saver to put the enum string then the value in the c code. It also make the c code more readable.

The get_op_params() of GpuDnnSoftmaxBase could be simplified as:

    mode = "CUDNN_SOFTMAX_MODE_" + self.mode.upper()
    algo = "CUDNN_SOFTMAX_" + self.algo.upper()
    return [("SOFTMAX_MODE", mode), ("SOFTMAX_ALGO", algo)]

But I agree with @abergeron we should stop using get_op_param. Using it make us generate more c code version. So we should convert them to params.

notoraptor:
@nouiz This PR was based on the discussion here: https://github.com/Theano/Theano/issues/5696#issuecomment-286232264

What I understand is that we want to reduce variants of C code, particularly for GpuDnnSoftmaxBase.

With current code, as C constants `SOFTMAX_MODE` and `SOFTMAX_ALGO` change every time `mode` or `algo` change, then there is at least 6 possible versions of C code (2 values for mode, 3 values for algo, so 2x3 cases).

What I suggest to have only 1 C code is to add a way to define a list of constants C macros just once (e.g. 2 macros for mode and 3 macros for algo), and then use mode and algo as op params that take the right macro values at runtime. To do that, we need a way to create sort of enum type so that constant values will be available with same values in both Python and C code. That's what I try to do here.

At the end, the C code will be something like this in all cases:

```C
#define ALGO_FAST 0
#define ALGO_ACCURATE 1
#define ALGO_LOG 2
#define MODE_INSTANCE 1
#define MODE_CHANNEL 2

/* ... */

// In main function:
int APPLY_SPECIFIC(softmax)(PyGpuArrayObject *x, PyGpuArrayObject **out, 
                            /* We will use wrapper to wrap all params 
                               (mode, algo and cudnn handle). */
                            APPLY_SPECIFIC_WRAPPER* wrapper) {
    int mode = wrapper->mode;
    int algo = wrapper->algo;
    switch(mode) {
        case MODE_INSTANCE: mode = CUDNN_SOFTMAX_MODE_INSTANCE; break;
        case MODE_CHANNEL: mode = CUDNN_SOFTMAX_MODE_CHANNEL; break;
        default:break;
    }
    switch(algo) {
        case ALGO_FAST: algo = CUDNN_SOFTMAX_FAST; break;
        case ALGO_ACCURATE: algo = CUDNN_SOFTMAX_ACCURATE; break;
        case ALGO_LOG: algo = CUDNN_SOFTMAX_LOG; break;
        default: break;
    }
    /* algo and mode are now well configured, we can work with it. */
    /* ... */
}
```

Then, this C code will no more vary depending on mode and algo.

Is it exactly what we want ?

abergeron:
I wouldn't say "exactly what we want", but something like that, yes.

theano/gof/tests/test_types.py
```diff
@@ -76,3 +76,118 @@ def test_cdata():
...
+    f = theano.function([a, b], [c_add, c_sub, c_multiply, c_divide])
+    va = 12
+    vb = 15
+    ref_add = va + vb
+    ref_sub = va - vb
+    ref_multiply = va * vb
+    ref_divide = va // vb
+    ref = [ref_add, ref_sub, ref_multiply, ref_divide]
+    out = f(va, vb)
+    assert ref == out, (ref, out)
```
nouiz: Test this with Wrapper too. We will need that cases.

theano/gof/type.py
```diff
@@ -795,3 +796,137 @@ def signature(self):
         return (self.type,)
 
 CDataType.Constant = CDataTypeConstant
+
+
+class EnumType(Type, dict):
```
nouiz: The name is EnumType, but this don't create or represent an enum, but macros. I'm not sure what to do about that, so at least, this should be in the docstring.

I'm not sure if we could reuse the enum from cudnn. Could CData type help us here? Or create something like that for enum? This way we will reuse the original enum and that remove the conversion in C. Right now, I think this PR move the conversion from Python to C. With the master of Theano, we can move the get_op_param to params if we hardcode the enum values that correspond to the cuda enum.

abergeron: The problem this is facing is that we want enum values in C and some sort of value in python to indicate what we want.

We can't just use the C values in python since we don't know what they are.  We can't really use the define name as a python value because this will be super painful to interact with from the C side, so the compromise is this.  A type that will define a bunch of values in python and C and can translate between each other.

Maybe we could have something that uses ints or whatever in python and the actual define values in C to have the best of both worlds.  In this case the python values wouldn't have any relation to the actual C values, but the type itself would have a mapping and use it to translate.  However I haven't done the work myself, so I'm not sure if it makes sense.

theano/gof/tests/test_types.py
notoraptor: PR for Wrapper is not yet merged, so Wrapper is not available in this branch.  I should wait for a PR to be merged, either this one or the PR for Wrapper, to make additional changes.


---

In any case, I will have to update Wrapper interface to handle constants, as the constant type is a bit special. Currently Wrapper interface is:
```python
wrapper = Wrapper(a=Scalar('int64'), 
                  b=Scalar('float32'),
                  c=Scalar('float64'))
```

I plan to extend this interface by allowing also the following one:
```python
wrapper = Wrapper(EnumList('CONSTANT_1', 'CONSTANT_2', 'CONSTANT_N'),
                  a=Scalar('int64'), 
                  b=Scalar('float32'),
                  c=Scalar('float64'))
```

ie., if the EnumType is declared in `*args` (not in `**kwargs`), then it will be treated as a special case by Wrapper, to that we could have:
```python
wrapper.CONSTANT_1
wrapper.CONSTANT_2
wrapper.CONSTANT_N
```
to make constants easy to access in Python.

theano/gof/type.py
notoraptor: There may be some very eccentric solutions if we allow using C++ syntax.

For example, we could use a special constant type which handle only C-defined constants:

```python
# we will need a type (for C code) and a name (to use as class name).
params_type = CEnumType('CUDNN_SOFTMAX_SOFT', 'CUDNN_SOFTMAX_ACCURATE',
              dtype='int32', name='MyAlgo')
```

Then in Python, we will use specific values, different from real C values, but that may be used for mapping steps (e.g array index values):

```python
params_type.CUDNN_SOFTMAX_SOFT # == 0
params_type.CUDNN_SOFTMAX_ACCURATE # == 1
```

We allow `CEnumType` to have values, like the other classic Theano types. Here it will be the `algo` value:
```python
# in __init__
self.algo = self.params_type.CUDNN_SOFTMAX_ACCURATE # == 1

# in get_params() (automatically handled by Wrapper is CEnumType is defined into a Wrapper)
return self.algo
```

But when converted to C++ code, we could create a specific class to represent the value passed in runtime, in which we overload the comparison operators:

```C++
/* PS: I have forgotten C++ syntax a little, but I could recover it quickly ! */

struct MyAlgo {
    static dtype map[2];
    int value;
    MyAlgo(int runtime_value): value(runtime_value) {}
    bool operator==(dtype other_value) {
        // The trick is here.
        return MyAlgo::map[this->value] == other_value;
    }
}
MyAlgo::map = {CUDNN_SOFTMAX_SOFT, CUDNN_SOFTMAX_ACCURATE};
```

Then, in main C function, we could write:
```C++
/* Algo would be of type MyAlgo */
if (algo == CUDNN_SOFTMAX_ACCURATE); // ...
```

A little hard, but if we allow C++ syntax, I think this is possible, although I will have to recover C++ practice. What do you think ?

abergeron: What about

```python
params_type  = CEnumType('cudnnSoftmaxAlgorithm_t', ('CUDNN_SOFTMAX_SOFT', 'CUDNN_SOFTMAX_ACCURATE'))
```

Then in python we have the simple ints that you propose.

And in the C++ code in c_extract() for CEnumType it does something like this:

```C
switch (PyInt_AsLong(py_val)) {
 case 0:
  val = CUDNN_SOFTMAX_SOFT;
  break;
 case 1:
  val = CUDNN_SOFTMAX_ACCURATE;
  break;
 default:
  %(fail)s
}
```

nouiz:
That is what I had in mind with more detail when I refered to an EnumType
like CDataType. Thanks for the detail and it also mean Iike this idea:)

@notoraptor, what do you think of that? You are the one that is the most in
the code.

notoraptor:
@abergeron @nouiz OK ! I will try the solution suggested by @abergeron !

notoraptor:
@abergeron @nouiz Updated. To summarize interface:
 * **`EnumType`**: base class, to explicitely define constant values. E.g.: `EnumType(CONSTANT_A=1, CONSTANT_B=2.5, CONSTANT_C=False)`. It allows integers, floating values and booleans (False becomes 0, True becomes 1). Now it can take another arg to define C type for op param: `EnumType(CONSTANT_A=1, CONSTANT_B=2.5, ctype='unsigned long')`. Default C type is `double`.
 * **`EnumList`**: extends `EnumType`, to define constants automatically. C type is configurable, default is `int`. E.g: `EnnumList('CONSTANT_A', 'CONSTANT_B', 'CONSTANT_C')`, constants will be respectively 0, 1 and 2, in the order they are specified.
 * **`CEnumType`**: extends `EnumList`. Constant names must exist in C code. Implemented as @abergeron suggested. C type is configurable, default is `int`. E.g: `CEnumType('CUDNN_SOFTMAX_SOFT', 'CUDNN_SOFTMAX_ACCURATE')`. Integer C type may be sufficient, but developer is free to specify another one:  `CEnumType('CUDNN_SOFTMAX_SOFT', 'CUDNN_SOFTMAX_ACCURATE', ctype='cudnnSoftmaxAlgorithm_t')`.

**NB**: Currently, I impose that constant names use capital letters instead of lowercase letters, as a convention, even for `CEnumType`.

notoraptor:
Small updates (doc updated, classes safer and tests clearer).

theano/gof/type.py
```diff
@@ -795,3 +796,241 @@ def signature(self):
...
+        super(EnumType, self).__init__(**kwargs)
+
+    def __repr__(self):
+        return '%s(%s)' % (type(self).__name__, ', '.join('%s:%s' % (k, self[k]) for k in sorted(self.keys())))
+
+    def __getattr__(self, key):
+        if key in self:
+            return self[key]
+        if key == 'ctype':
+            return self.ctype
```
abergeron: This will probably cause infinite recursion.  Just let it fall through the normal case below since that is a regular attribute.

theano/gof/type.py
```diff
@@ -795,3 +796,241 @@ def signature(self):
...
+
+    # EnumType should be used to create constants available in both Python and C code.
+    # However, for convenience, we make sure EnumType can have a value, like other common types,
+    # such that it could be used as-is as an op param.
+    # C type of value is defined in self.ctype.
+
+    def filter(self, data, strict=False, allow_downcast=None):
+        if not strict and isinstance(data, bool):
+            data = int(data)
+        assert isinstance(data, (int, float))
```
abergeron: You could even go further and make sure that the value we receive is actually one of the values defined for this type.

theano/gof/type.py
```diff
@@ -795,3 +796,241 @@ def signature(self):
...
+        return """
+        #if PY_MAJOR_VERSION >= 3
+            #ifndef PyInt_Check
+                #define PyInt_Check PyLong_Check
+            #endif
+            #ifndef PyInt_AsLong
+                #define PyInt_AsLong PyLong_AsLong
+            #endif
+        #endif
+        """
```
abergeron: Why do you have an extra method here?

theano/gof/type.py
```diff
@@ -795,3 +796,241 @@ def signature(self):
...
+
+    def c_extract(self, name, sub, check_input=True):
+        swapped_dict = dict((v, k) for (k, v) in self.items())
+        # swapped_dict's keys are integers.
+
+        return """
+        switch(PyInt_AsLong(py_%(name)s)) {
+            %(cases)s
+            default:
+                {%(fail)s}
```
abergeron: This fail will not have a python exception set when hit.

theano/gof/type.py
```diff
@@ -795,3 +796,241 @@ def signature(self):
...
+        Be sure C constants are available in your C code. If they come from a C header, consider implementing
+        ``c_headers()`` and ``c_header_dirs()`` in the Op class where you use CEnumType as op parameter type.
+
+    """
+
+    def c_support_code(self):
+        return self.c_support_macro_code()
+
+    def c_extract(self, name, sub, check_input=True):
+        swapped_dict = dict((v, k) for (k, v) in self.items())
```
abergeron: Should this be sorted to make sure that the C code remains the same?

theano/gof/type.py
```diff
@@ -795,3 +796,241 @@ def signature(self):
...
+            ''.join("""
+            #define %s %s
+            """ % (k, str(self[k])) for k in sorted(self.keys()))
+        )
+
+    def c_declare(self, name, sub, check_input=True):
+        return """%(ctype)s %(name)s;""" % dict(ctype=self.ctype, name=name)
+
+    def c_init(self, name, sub):
+        return "%(name)s = 0;" % locals()
```
abergeron: Please avoid locals() in new code.

theano/gof/type.py
notoraptor: It is sorted later just when building the returned string.

theano/gof/type.py
notoraptor: To reuse it specifically in CEnumType later, instead of copying this string again.

notoraptor:
Updated. I have made 2 comments above.

theano/gof/type.py
abergeron: Ok then perhaps you should name it differently so that it doesn't look like a new portion of the C code interface.

theano/gof/type.py
abergeron: Yes, ok

theano/gof/type.py
notoraptor: Renamed.

nouiz:
Since https://github.com/Theano/Theano/pull/5612 is now merged, can you rebase this PR on top of master and adjust it for the name changes?

notoraptor:
Rebased with theano master branch, @abergeron @nouiz .

I will check if there are some adjustements in GpuDnnSoftmax PR that I could transfer here.

notoraptor:
Ultimately, I think we could approve this PR as is, and then switch to GpuDnnSoftmax PR for the next works. It will help us see ParamsType and EnumTypes in action, and then have a better look for code integration needs.

abergeron:
I'm ok to merge right now, what about you @nouiz?

theano/gof/type.py
```diff
@@ -808,3 +809,244 @@ def signature(self):
...
+        if not strict and isinstance(data, bool):
+            data = int(data)
+        assert data in self.values()
+        return data
+
+    def values_eq(self, a, b):
+        return a == b
+
+    def values_eq_approx(self, a, b):
+        return float(a) == float(b)
```
nouiz: Put a comment like:

    For an enum, it don't have a meaning to be approx equal.

And I would compute it as values_eq()

notoraptor:
@nouiz Updated.

nouiz:
@abergeron do you agree with the last change I asked?
jenkins test this

PR #5715, opened 2017-03-15
Remove useless warning when the profile was manually disabled

nouiz:


nouiz:
merging, only jenkins gpu tests missing and this is not related.

PR #5716, opened 2017-03-16
Better error message when an opt make an invalide replace.

nouiz:


abergeron:
I'm afraid this will lead to more obscure crashes where some catches an exception to append some stuff, but assumes that the \_\_init__ only takes a single argument.

Would it be possible to make the \_\_init__ of BadOptimization have a single string argument?

theano/gof/toolbox.py
```diff
@@ -100,8 +103,15 @@ def __init__(self, old_r, new_r, old_r_val, new_r_val, reason,
         self.reason = reason
         self.old_graph = old_graph
         self.new_graph = new_graph
+        self.full_err = None
+        if isinstance(old_r, str):
+            assert (new_r is None and old_r_val is None and new_r_val is None and
+                    reason is None and old_graph is None and new_graph is None)
+            self.full_err = old_r
```
lamblin: I think it's quite confusing to have essentially two modes for the same Exception.
At least we should have comments explaining why we are doing that here.

nouiz: There is one such comment on line 43 of this file.

abergeron: It might be confusing, but otherwise we end up with even more confusing errors like this:

```
TypeError: __init__ has 7 arguments but got only 1
```

Instead of the real error which just triggered a BadOptimization.

lamblin: I'm not opposed to the change in the code, I just think that there should be a comment in the code of `__init__` and `__str__` explaining how the different paths taken relate to these two modes.
If you think this is already clear enough, I won't block the PR, and blame my cold for why it took me a while to understand.

lamblin:
@nouiz We started updating the comments together, but I don't think you updated the PR yet.

nouiz:
I rebased and clarified the code and doc and added tests.

Ready to be reviewed.

PR #5717, opened 2017-03-16
crash fix of careduce with bool

nouiz:


theano/tensor/elemwise.py
```diff
@@ -1539,15 +1539,18 @@ def _c_all(self, node, name, inames, onames, sub):
...
+                    # numpy does not define NPY_MIN_UINT* and NPY_MIN_BOOL
                     identity = "0"
                 else:
                     identity = "NPY_MIN_" + str(input.type.dtype).upper()
             if self.scalar_op == scalar.minimum:
                 scal_name = 'minimum'
                 if input.type.dtype in ["float32", "float64"]:
                     identity = "__builtin_inf()"
+                elif input.type.dtype == 'bool':
+                    # numpy does not define NPY_MIN_UINT* and NPY_MAX_BOOL
```
lamblin: Update this comment, this case only concerns `NPY_MAX_BOOL`

nouiz:
As the fix is CPU only and travis passed, I'll merge. jenkins is delayed as it want to reboot...

PR #5718, opened 2017-03-16
Prepare final release 0.9.0.

notoraptor:
Few remarks:
 * I have moved old content of `NEWS.txt` into `HISTORY.txt`. New content of `NEWS.txt` is the old from `NEWS_DEV.txt` with corrections and updates from last PRs since 0.9.0rc4.
 * Last version for flag `warn.ignore_bug_before` is `0.9`, but I use `0.9.0` to name this new release. Is it a problem ?
 * In `LICENSE.txt`, Theano copyright year was already updated in previous PRs, but copyright years for NumPy and NVIDIA are still  previous years (2015, 2016). Should I update them also ?
* There was a line in section "new features" of NEWS_DEV: `conv3d2d support full and half mode (REMOVE?)`. I have moved it to section "convolution updates" and rewritten it : `Support of full and half modes for 2D and 3D convolutions`.

Rendering of `NEWS.txt` for this PR is available here: http://notoraptor.net/theano/html/NEWS.html

@nouiz @lamblin @abergeron


nouiz:
For the license, I think it is ok to let the other copywith with the old years. We didn't copy more recent codes.

For the move of the convolution line, it is good, but not the rewrite. conv3d2d is a specific implementation. But I think we now support it for all implementation. So what about:

Support of full and half modes for 2D and 3D convolutions including in conv3d2d

NEWS_DEV.txt
```diff
@@ -15,115 +15,147 @@ git shortlog -sn rel-0.8.0..
 
 TODO: better Theano conv doc
 
+# NB: Following notes are related to final release 0.9.0.
```
nouiz: All what is bellow seem a duplicate of what is in NEWS.txt. Also, what should be in NEWS_DEV.txt is what is in the master since the last release. So I would just remove all that.

NEWS.txt
```diff
@@ -3,334 +3,253 @@ Release Notes
...
+ - Use of 64-bit indexing in sparse ops to allow matrix with more then 2\ :sup:`31`\ -1 elements
+ - Allowed more then one output to be an destructive inplace
  - More support of negative axis
  - Added the keepdims parameter to the norm function
- - Crash fixes
  - Make scan gradient more deterministic
- - Add support for space in path on Windows
- - remove ProfileMode (use Theano flag profile=True instead)
 
+Commiters since 0.9.0rc4:
```
nouiz: I think this list is made on the current master. Theano 0.9 will be the same version as 0.9.0rc4. Not again the master. We have stuff in the master that should not go in 0.9.

I think this will also request some updates of the NEWS.txt file.

notoraptor:
@nouiz Updated. NEWS.twt only takes account of modifications from 0.8.0 to 0.9.0rc4.

Notes in NEWS_DEV.txt are now only related to modifications since 0.9.0rc4.

notoraptor:
Rendering updated: http://notoraptor.net/theano/html/NEWS.html

notoraptor:
I was wondering: If we merge this PR into master branch and tag the merge commit with `0.9.0`, then the release archive will include commits pushed between tag 0.9.0rc4 and tag 0.9.0. But we don't want that, as release 0.9.0 must contain exactly the sames changes as rc4, isn't it ?

So, I think this PR should better be merged and tagged directly and only into 0.9.X branch. I can re-do a PR on 0.9.X branch and close this one. What do you think, @nouiz ?

nouiz:
I think we need this PR on both branch (master and 0.9.X) For the tag, it should be in the 0.9.X branch. This way, the both branch have all what they need and we don't have the problem you tell.

I'll merge this PR now. You can cherry-pick the commits from this PR to the 0.9.X branch, do the tag and push them directly to that branch.

Issue #5712, opened 2017-03-14
Softmax improvement


bordesf:
I will write some improvement for the softmax implementation as CCW, here is my plan:

- [ ] Allow to have n-dimensional tensor as input where the softmax operation is applied on the last dimension. (Clean some C code and try to parallelize it eventually) 
- [ ] Correct some issues with vector as input (Currently the vector operation works only with the cpu optimization and it returns a matrix instead of a vector)
- [ ] Add a flag to use a spatial softmax in case of 4d tensor (cf https://github.com/Theano/Theano/issues/4269). I think that I will try to match the new implementation with CUDNN where we can choose between the 'instance' mode that corresponds to apply the softmax on the last dimension or the 'channel' mode that allows to perform the softmax on the first dimension.
- [] ...

f0k:
It would be great if one could directly specify the axes to softmax over (accepting an int, a tuple of int, or the string shortcuts 'instance' and 'channel'), and have an optimization that uses the cuDNN implementation whenever possible. But I haven't checked if this is feasible to implement (there's already a jungle of Ops and optimizers just for the simple 2D case). Note that the categorical crossentropy needs to be extended along with the softmax.

JesseLivezey:
FYI, PR #5393 exists and seems to be pretty close to finished although still needs review

nouiz:
That PR just make the current softmax parallel on the CPU. It don't change
anything about what is here.

I should finish it at some point, I want to investigate a few things and do
more detailled profile before merging it.

astooke:
Bump...any updates?

nouiz:
There was much work around, but I tihnk it didn,t got finished/reviewed due
to the of our work on theano.

PR #5719, opened 2017-03-16
Softmax improvement

bordesf:
Ref to #5712 

- Some modifications to Softmax and GradSoftmax to allow a N-dimensional array as input and apply the softmax operation on the last dimension.

For the next commits, I will need to add and change some tests in test_nnet.py. With the new implementation if the input is a vector, the output will be a vector. Currently the tests consider the fact that the softmax give a matrix if the input is a vector, so the test will crash in those cases.

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -447,7 +447,7 @@ class Softmax(gof.Op):
         # is the same as the grad
         if None in eval_points:
             return [None]
-        return self.grad(inputs, eval_points)
+        return self.L_op(inputs, [self(*inputs)], eval_points)
 
     def infer_shape(self, node, shape):
         return shape
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -292,7 +292,6 @@ softmax_with_bias = SoftmaxWithBias()
 class SoftmaxGrad(gof.Op):
     """
     Gradient wrt x of the Softmax Op.
-
     """
 
     nin = 2
@@ -302,39 +301,29 @@ class SoftmaxGrad(gof.Op):
     def make_node(self, dy, sm):
         dy = tensor.as_tensor_variable(dy)
         sm = tensor.as_tensor_variable(sm)
-        if dy.type.ndim not in (1, 2) \
-                or dy.type.dtype not in tensor.float_dtypes:
-            raise ValueError('dy must be 1-d or 2-d tensor of floats. Got ',
-                             dy.type)
-        if dy.ndim == 1:
-            dy = tensor.shape_padleft(dy, n_ones=1)
-        if sm.ndim == 1:
-            sm = tensor.shape_padleft(sm, n_ones=1)
         return Apply(self, [dy, sm], [sm.type()])
 
     def perform(self, node, input_storage, output_storage):
         dy, sm = input_storage
+        if (dy.shape != sm.shape):
+            raise ValueError('dy and the softmax output should have the same shape.')
         dx = numpy.zeros_like(sm)
-        # dx[i,j] = - (\sum_k dy[i,k] sm[i,k]) sm[i,j] + dy[i,j] sm[i,j]
-        for i in xrange(sm.shape[0]):
-            dy_times_sm_i = dy[i] * sm[i]
-            dx[i] = dy_times_sm_i - sum(dy_times_sm_i) * sm[i]
+        dx_tmp = numpy.zeros_like(sm)
+        dy_times_sm = dy * sm
+        dx = dy_times_sm - (numpy.sum(dy_times_sm, axis=-1, keepdims=True) * sm)
         output_storage[0][0] = dx
 
     def grad(self, inp, grads):
         dy, sm = inp
         g, = grads
-
-        tmp = g + tensor.neg(tensor.sum(g * sm, axis=1).dimshuffle((0, 'x')))
+        tmp = g + tensor.neg(tensor.sum(g * sm, axis=-1, keepdims=True))
         g_dy = tmp * sm
-
-        tmp2 = tensor.sum(dy * sm, axis=1).dimshuffle((0, 'x'))
+        tmp2 = tensor.sum(dy * sm, axis=-1, keepdims=True)
         g_sm = tmp * dy - g * tmp2
-
         return g_dy, g_sm
 
     def infer_shape(self, node, shape):
-        return [shape[1]]
+        return [shape[-1]]
 
     def c_code_cache_version(self):
         return (3,)
@@ -343,68 +332,80 @@ class SoftmaxGrad(gof.Op):
         dy, sm = inp
         dx, = out
         return '''
-        if ((PyArray_TYPE(%(dy)s) != NPY_DOUBLE) &&
-            (PyArray_TYPE(%(dy)s) != NPY_FLOAT))
-        {
-            PyErr_SetString(PyExc_TypeError,
-                 "types should be float or float64");
-            %(fail)s;
-        }
-        if ((PyArray_TYPE(%(sm)s) != NPY_DOUBLE) &&
-            (PyArray_TYPE(%(sm)s) != NPY_FLOAT))
-        {
-            PyErr_SetString(PyExc_TypeError,
-                 "types should be float or float64");
-            %(fail)s;
-        }
-        if ((PyArray_NDIM(%(dy)s) != 2)
-            || (PyArray_NDIM(%(sm)s) != 2))
-        {
-            PyErr_SetString(PyExc_ValueError, "rank error");
-            %(fail)s;
-        }
-        if (PyArray_DIMS(%(dy)s)[0] != PyArray_DIMS(%(sm)s)[0])
-        {
-            PyErr_SetString(PyExc_ValueError, "dy.shape[0] != sm.shape[0]");
-            %(fail)s;
-        }
-        if ((NULL == %(dx)s)
-            || (PyArray_DIMS(%(dx)s)[0] != PyArray_DIMS(%(sm)s)[0])
-            || (PyArray_DIMS(%(dx)s)[1] != PyArray_DIMS(%(sm)s)[1]))
-        {
-            Py_XDECREF(%(dx)s);
-            %(dx)s = (PyArrayObject*) PyArray_SimpleNew(2,
-                                                        PyArray_DIMS(%(sm)s),
-                                                        PyArray_TYPE(%(sm)s));
-            if (!%(dx)s)
-            {
-                PyErr_SetString(PyExc_MemoryError,
-                     "failed to alloc dx output");
-                %(fail)s;
-            }
-        }
+            npy_intp* input_shape = PyArray_DIMS(%(sm)s);
+            npy_int ndim = PyArray_NDIM(%(sm)s);
+            if ((PyArray_TYPE(%(dy)s) != NPY_DOUBLE) &&
+                    (PyArray_TYPE(%(dy)s) != NPY_FLOAT))
+                {
+                    PyErr_SetString(PyExc_TypeError,
+                        "types should be float or float64");
+                    %(fail)s;
+                }
 
-        for (size_t i = 0; i < PyArray_DIMS(%(dx)s)[0]; ++i)
-        {
-            const dtype_%(dy)s* __restrict__ dy_i = (dtype_%(dy)s*) (PyArray_BYTES(%(dy)s) + PyArray_STRIDES(%(dy)s)[0] * i);
-            npy_intp Sdy = PyArray_STRIDES(%(dy)s)[1]/sizeof(dtype_%(dy)s);
-            const dtype_%(sm)s* __restrict__ sm_i = (dtype_%(sm)s*) (PyArray_BYTES(%(sm)s) + PyArray_STRIDES(%(sm)s)[0] * i);
-            npy_intp Ssm = PyArray_STRIDES(%(sm)s)[1]/sizeof(dtype_%(sm)s);
-            dtype_%(dx) s* __restrict__ dx_i = (dtype_%(dx)s*) (PyArray_BYTES(%(dx)s) + PyArray_STRIDES(%(dx)s)[0] * i);
-            npy_intp Sdx = PyArray_STRIDES(%(dx)s)[1]/sizeof(dtype_%(dx)s);
+            // Check input type
+            if ((PyArray_TYPE(%(sm)s) != NPY_DOUBLE) &&
+                    (PyArray_TYPE(%(sm)s) != NPY_FLOAT))
+            {
+                    PyErr_SetString(PyExc_TypeError,
+                        "Types should be float or float64");
+                    %(fail)s;
+                    }
 
-            double sum_dy_times_sm = 0.;
-            for (size_t j = 0; j < PyArray_DIMS(%(dx)s)[1]; ++j)
+            // Check the number of dimension
+            if (PyArray_NDIM(%(dy)s) != PyArray_NDIM(%(sm)s))
+                {
+                    PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same number of dimension.");
+                    %(fail)s;
+                }
+
+            // Check the shapes
+            if (PyArray_DIMS(%(dy)s) == PyArray_DIMS(%(sm)s))
+                {
+                    PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same shape.");
+                    %(fail)s;
+                }
+
+            // Alloc memory for the output
+            if ((NULL == %(dx)s) || (PyArray_DIMS(%(dx)s) != PyArray_DIMS(%(sm)s)))
+                {
+                    Py_XDECREF(%(dx)s);
+                    %(dx)s = (PyArrayObject*) PyArray_SimpleNew(PyArray_NDIM(%(sm)s),
+                                                                PyArray_DIMS(%(sm)s),
+                                                                PyArray_TYPE(%(sm)s));
+                    if (!%(dx)s)
+                    {
+                        PyErr_SetString(PyExc_MemoryError,
+                            "failed to alloc dx output");
+                        %(fail)s;
+                    }
+                }
+
+            // Compute array size except on the last dimension
+            npy_int input_size;
+            input_size = PyArray_SIZE(%(sm)s) / input_shape[ndim - 1];
+
+            // Compute stride softmax
+            npy_int stride;
+            stride = PyArray_DIM(%(sm)s, ndim - 1) * PyArray_ITEMSIZE(%(sm)s);
+
+            for (size_t i = 0; i < input_size; ++i)
             {
-                dx_i[j * Sdx] = dy_i[j * Sdy] * sm_i[j * Ssm];
-                sum_dy_times_sm += dx_i[j * Sdx];
-            }
-            for (size_t j = 0; j < PyArray_DIMS(%(dx)s)[1]; ++j)
-            {
-                dx_i[j * Sdx] -= sum_dy_times_sm * sm_i[j * Ssm];
-            }
-        }
-        ''' % dict(locals(), **sub)
+                const dtype_%(dy)s* __restrict__ dy_i = (dtype_%(dy)s*) (PyArray_BYTES(%(dy)s) + stride * i);
+                const dtype_%(sm)s* __restrict__ sm_i = (dtype_%(sm)s*) (PyArray_BYTES(%(sm)s) + stride * i);
+                dtype_%(dx) s* __restrict__ dx_i = (dtype_%(dx)s*) (PyArray_BYTES(%(dx)s) + stride * i);
+                double sum_dy_times_sm = 0.;
+                for (size_t j = 0; j < input_shape[ndim - 1]; ++j)
+                {
+                    dx_i[j] = dy_i[j] * sm_i[j];
+                    sum_dy_times_sm += dx_i[j];
+                }
+                for (size_t j = 0; j < input_shape[ndim - 1]; ++j)
+                {
+                    dx_i[j] -= sum_dy_times_sm * sm_i[j];
+                }
+                }
+            ''' % dict(locals(), **sub)
+
 softmax_grad = SoftmaxGrad()
 
 
@@ -423,20 +424,17 @@ class Softmax(gof.Op):
     __props__ = ()
 
     def make_node(self, x):
-        x = tensor.as_tensor_variable(x)
-        if x.type.ndim not in (1, 2) \
-                or x.type.dtype not in tensor.float_dtypes:
-            raise ValueError('x must be 1-d or 2-d tensor of floats. Got %s' %
-                             x.type)
+        x = theano.tensor.as_tensor_variable(x)
+        # TODO : Delete this and modify the test accordly
         if x.ndim == 1:
             x = tensor.shape_padleft(x, n_ones=1)
-
         return Apply(self, [x], [x.type()])
 
     def perform(self, node, input_storage, output_storage):
         x, = input_storage
-        e_x = numpy.exp(x - x.max(axis=1)[:, None])
-        sm = e_x / e_x.sum(axis=1)[:, None]
+        # Apply softmax on the last dimension
+        e_x = numpy.exp(x - x.max(axis=-1, keepdims=True))
+        sm = e_x / e_x.sum(axis=-1, keepdims=True)
         output_storage[0][0] = sm
 
     def L_op(self, inp, outputs, grads):
@@ -449,7 +447,7 @@ class Softmax(gof.Op):
         # is the same as the grad
         if None in eval_points:
             return [None]
-        return self.L_op(inputs, [self(*inputs)], eval_points)
+        return self.grad(inputs, eval_points)
 
     def infer_shape(self, node, shape):
         return shape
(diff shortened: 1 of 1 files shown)


theano/tensor/nnet/nnet.py
```diff
@@ -423,20 +423,17 @@ class Softmax(gof.Op):
     __props__ = ()
 
     def make_node(self, x):
-        x = tensor.as_tensor_variable(x)
-        if x.type.ndim not in (1, 2) \
-                or x.type.dtype not in tensor.float_dtypes:
-            raise ValueError('x must be 1-d or 2-d tensor of floats. Got %s' %
```
ReyhaneAskari: You should still raise an error if the dtype is not in ```float_dtypes ```.

theano/tensor/nnet/nnet.py
```diff
@@ -423,20 +423,17 @@ class Softmax(gof.Op):
...
-                             x.type)
+        x = theano.tensor.as_tensor_variable(x)
```
ReyhaneAskari: ```tensor``` is already imported. So change this line to ```x = tensor.as_tensor_variable(x)```

theano/tensor/nnet/nnet.py
```diff
@@ -302,39 +301,28 @@ class SoftmaxGrad(gof.Op):
     def make_node(self, dy, sm):
         dy = tensor.as_tensor_variable(dy)
         sm = tensor.as_tensor_variable(sm)
-        if dy.type.ndim not in (1, 2) \
-                or dy.type.dtype not in tensor.float_dtypes:
```
ReyhaneAskari: You should still raise an error if the dtype is not in ```float_dtypes ```.

theano/tensor/nnet/nnet.py
```diff
@@ -302,39 +301,28 @@ class SoftmaxGrad(gof.Op):
...
-            raise ValueError('dy must be 1-d or 2-d tensor of floats. Got ',
-                             dy.type)
-        if dy.ndim == 1:
-            dy = tensor.shape_padleft(dy, n_ones=1)
-        if sm.ndim == 1:
-            sm = tensor.shape_padleft(sm, n_ones=1)
```
ReyhaneAskari: Keep lines 309-312. We still need to handle vectors.

theano/tensor/nnet/nnet.py
```diff
@@ -302,39 +301,28 @@ class SoftmaxGrad(gof.Op):
...
         return Apply(self, [dy, sm], [sm.type()])
 
     def perform(self, node, input_storage, output_storage):
         dy, sm = input_storage
+        if (dy.shape != sm.shape):
+            raise ValueError('dy and the softmax output should have the same shape.')
```
ReyhaneAskari: Isn't this always guaranteed? @lamblin  What do you think?

theano/tensor/nnet/nnet.py
bordesf: The goal is to handle vector without the need to pad them. (We don't really need to transform them into a matrix).

theano/tensor/nnet/nnet.py
```diff
@@ -465,119 +462,114 @@ def c_code_template(dtype):
...
-        {
-            Py_XDECREF(%(sm)s);
-            %(sm)s = (PyArrayObject*)PyArray_SimpleNew(2, PyArray_DIMS(%(x)s),
-                                                       PyArray_TYPE(%(x)s));
-            if(!%(sm)s) {
-                PyErr_SetString(PyExc_MemoryError,
-                     "failed to alloc sm output");
-                %(fail)s
+            // Allocate memory for the softmax output
+            if (NULL == %(sm)s || (PyArray_DIMS(%(x)s) != PyArray_DIMS(%(sm)s)))
```
ReyhaneAskari: @lamblin. Do we need to check each dimension separately or we can check it like that. I can't think of a case where ```x.shape != sm.shape```. Could you give an example? Then we can also add it to the tests.

theano/tensor/nnet/nnet.py
```diff
@@ -465,119 +462,114 @@ def c_code_template(dtype):
...
         """
 
         inside_row_loop = """
-            for (j = 0; j < Nx[1]; ++j)
+            // Substract the max and take the exponential
+            for (j = 0; j < input_shape[ndim - 1]; ++j)
             {
-                dtype_%(sm)s row_ij = x_i[j * Sx1] ;
-                //std::cout << "2 " << j << " " << row_ij << " " << row_max << "\\n";
+                dtype_%(sm)s row_ij = x_i[j] ;
```
ReyhaneAskari: You need to use the stride here to move in a row: 

```
dtype_%(sm)s row_ij = x_i[j * stride_of_the_last_dimentsion]
```

theano/tensor/nnet/nnet.py
```diff
@@ -465,119 +462,114 @@ def c_code_template(dtype):
...
                 dtype_%(sm)s sm_ij = exp(row_ij - row_max);
-                //std::cout << "3 " << j << " " << sm_ij << "\\n";
                 sum += sm_ij;
-                sm_i[j * Ssm1] = sm_ij;
+                sm_i[j] = sm_ij;
```
ReyhaneAskari: You need to use the stride here to move in a row: 

```
sm_i[ j  * stride_of_the_last_dimentsion ] = sm_ij;
```

theano/tensor/nnet/nnet.py
```diff
@@ -465,119 +462,114 @@ def c_code_template(dtype):
...
-        for (size_t i = 0; i < Nx[0]; ++i)
-        {
-            size_t j;
-            double sum = 0.0;
-
-            const dtype_%(x)s* __restrict__ x_i = (dtype_%(x)s*)(PyArray_BYTES(%(x)s) + PyArray_STRIDES(%(x)s)[0] * i);
-            dtype_%(sm) s* __restrict__ sm_i = (dtype_%(sm)s*)(PyArray_BYTES(%(sm)s) + PyArray_STRIDES(%(sm)s)[0] * i);
+            // Compute stride
+            npy_int stride;
+            stride = PyArray_DIM(%(x)s, ndim - 1) * PyArray_ITEMSIZE(%(x)s);
```
ReyhaneAskari: Have you tested this? c_code is sometimes loaded from the cache and older versions were used. 

@lamblin for the ```c_code ``` function we needed to clean the cache or disable the c_code_cache. How can Florian do it in this  ```c_code_template``` function?

bordesf: When I was testing this, I just disable the c_code_cache. I just put it back when I committed.

theano/tensor/nnet/nnet.py
bordesf: In this case, we are on the last dimension. We don't really need to have a stride here since it's value will be always 1. But I agree in other cases, we should compute it.

lamblin:
I still have not gotten around to reviewing this, I'll try to do that tomorrow.

theano/tensor/nnet/nnet.py
lamblin: Some tests would assume that the softmax of a vector is a row, so that `softmax(v)[arange(1), y]` still inserts the `cross_entropy_softmax...` Op.
It is OK to change it in this PR, but we will have to handle the transition.

theano/tensor/nnet/nnet.py
lamblin: Yes, the shape of a variable and of the gradient wrt that variable should have the same shape.

theano/tensor/nnet/nnet.py
```diff
@@ -302,39 +301,28 @@ class SoftmaxGrad(gof.Op):
...
         dy, sm = input_storage
+        if (dy.shape != sm.shape):
+            raise ValueError('dy and the softmax output should have the same shape.')
         dx = numpy.zeros_like(sm)
-        # dx[i,j] = - (\sum_k dy[i,k] sm[i,k]) sm[i,j] + dy[i,j] sm[i,j]
-        for i in xrange(sm.shape[0]):
-            dy_times_sm_i = dy[i] * sm[i]
-            dx[i] = dy_times_sm_i - sum(dy_times_sm_i) * sm[i]
+        dy_times_sm = dy * sm
+        dx = dy_times_sm - (numpy.sum(dy_times_sm, axis=-1, keepdims=True) * sm)
```
lamblin: OK. I'd still keep the comment with the math (or update it).

theano/tensor/nnet/nnet.py
```diff
@@ -302,39 +301,28 @@ class SoftmaxGrad(gof.Op):
...
-
-        tmp2 = tensor.sum(dy * sm, axis=1).dimshuffle((0, 'x'))
+        tmp2 = tensor.sum(dy * sm, axis=-1, keepdims=True)
         g_sm = tmp * dy - g * tmp2
-
         return g_dy, g_sm
 
     def infer_shape(self, node, shape):
-        return [shape[1]]
+        return [shape[-1]]
```
lamblin: `shape[1]` here means the shape of the symbolic input number 1 (here, `sm` as passed to `make_node`).
it works with `[-1]` as well, but its less clear.

theano/tensor/nnet/nnet.py
```diff
@@ -302,39 +301,28 @@ class SoftmaxGrad(gof.Op):
...
 
     def c_code_cache_version(self):
         return (3,)
```
lamblin: Update the version number if you update the C code

theano/tensor/nnet/nnet.py
```diff
@@ -343,68 +331,80 @@ def c_code(self, node, name, inp, out, sub):
...
+            npy_int ndim = PyArray_NDIM(%(sm)s);
+            if ((PyArray_TYPE(%(dy)s) != NPY_DOUBLE) &&
+                    (PyArray_TYPE(%(dy)s) != NPY_FLOAT))
+                {
+                    PyErr_SetString(PyExc_TypeError,
+                        "types should be float or float64");
+                    %(fail)s;
+                }
+
+            // Check input type
```
lamblin: Maybe put that comment before the first check.

theano/tensor/nnet/nnet.py
```diff
@@ -343,68 +331,80 @@ def c_code(self, node, name, inp, out, sub):
...
+
+            // Check the number of dimension
+            if (PyArray_NDIM(%(dy)s) != PyArray_NDIM(%(sm)s))
+                {
+                    PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same number of dimension.");
+                    %(fail)s;
+                }
+
+            // Check the shapes
+            if (PyArray_DIMS(%(dy)s) == PyArray_DIMS(%(sm)s))
```
lamblin: PyArray_DIMS just returns a pointer to the array of shapes, the comparison will fail.
You will need to loop and compare each `PyArray_DIMS(...)[i]`.

theano/tensor/nnet/nnet.py
```diff
@@ -343,68 +331,80 @@ def c_code(self, node, name, inp, out, sub):
...
+                {
+                    PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same shape.");
+                    %(fail)s;
+                }
+
+            // Alloc memory for the output
+            if ((NULL == %(dx)s) || (PyArray_DIMS(%(dx)s) != PyArray_DIMS(%(sm)s)))
```
lamblin: Idem here, you will have to loop to compare the shapes.

theano/tensor/nnet/nnet.py
```diff
@@ -343,68 +331,80 @@ def c_code(self, node, name, inp, out, sub):
...
+                    {
+                        PyErr_SetString(PyExc_MemoryError,
+                            "failed to alloc dx output");
+                        %(fail)s;
+                    }
+                }
+
+            // Compute array size except on the last dimension
+            npy_int input_size;
+            input_size = PyArray_SIZE(%(sm)s) / input_shape[ndim - 1];
```
lamblin: This could crash the program if `input_shape[ndim - 1]` is equal to 0.
Could you detect that case and raise a Python exception instead? It makes things much easier to debug.

theano/tensor/nnet/nnet.py
```diff
@@ -343,68 +331,80 @@ def c_code(self, node, name, inp, out, sub):
...
+
+            // Compute stride softmax
+            npy_int stride;
+            stride = PyArray_DIM(%(sm)s, ndim - 1) * PyArray_ITEMSIZE(%(sm)s);
```
lamblin: It's more complex than that if the arrays are not contiguous. The stride can also be different for the different arrays.
That being said, I believe numpy gives iterators that can be easier to use than trying to compute the strides explicitly.

theano/tensor/nnet/nnet.py
```diff
@@ -343,68 +331,80 @@ def c_code(self, node, name, inp, out, sub):
...
-            }
-        }
-        ''' % dict(locals(), **sub)
+                const dtype_%(dy)s* __restrict__ dy_i = (dtype_%(dy)s*) (PyArray_BYTES(%(dy)s) + stride * i);
+                const dtype_%(sm)s* __restrict__ sm_i = (dtype_%(sm)s*) (PyArray_BYTES(%(sm)s) + stride * i);
+                dtype_%(dx) s* __restrict__ dx_i = (dtype_%(dx)s*) (PyArray_BYTES(%(dx)s) + stride * i);
+                double sum_dy_times_sm = 0.;
+                for (size_t j = 0; j < input_shape[ndim - 1]; ++j)
+                {
+                    dx_i[j] = dy_i[j] * sm_i[j];
```
lamblin: `dx_i`, `dy_i`, and `sm_i` can all have different strides as well...

theano/tensor/nnet/nnet.py
lamblin: We need to check each dimension separately (I had not seen your comment here before commenting earlier about that).
It is possible that output memory `%(sm)s` is kept around between different calls to the compiled theano function, but that the shape of the arguments changed. We re-use it only if it has the right shape.
(We could try to be more clever, and re-use it if it is too big, but not too small, but it brings additional issues.)
DebugMode with `check_preallocated_out` checks this sort of thing, but I don't think it has been run in a while (it is really slow).

theano/tensor/nnet/nnet.py
lamblin: It will not always be 1. The arrays are not guaranteed to be contiguous, even though they often are.

bordesf:
Should I use the vectorized version of exp in logSoftmax if is available ? (Like in Softmax op)

theano/tensor/nnet/nnet.py
ReyhaneAskari: So, please remove lines 310 and 311.

theano/tensor/nnet/nnet.py
```diff
@@ -302,109 +301,151 @@ class SoftmaxGrad(gof.Op):
...
-        {
-            Py_XDECREF(%(dx)s);
-            %(dx)s = (PyArrayObject*) PyArray_SimpleNew(2,
-                                                        PyArray_DIMS(%(sm)s),
-                                                        PyArray_TYPE(%(sm)s));
-            if (!%(dx)s)
+            npy_intp* input_shape = PyArray_DIMS(%(sm)s);
+            npy_int ndim_sm = PyArray_NDIM(%(sm)s);
+            npy_int ndim_dy = PyArray_NDIM(%(dy)s);
+            npy_intp* shape_sm = PyArray_DIMS(%(sm)s);
```
ReyhaneAskari: Lines 337 and 340 are the same. Remove line 337 and replace ```input_shape``` with ```shape_sm```, in the rest of the code.

theano/tensor/nnet/nnet.py
```diff
@@ -302,109 +301,151 @@ class SoftmaxGrad(gof.Op):
...
+                }
+
+            // Check the number of dimension
+            if (ndim_dy != ndim_sm)
+                {
+                    PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same number of dimension.");
+                    %(fail)s;
+                }
+
+            // Check the shapes of sm and dy
```
ReyhaneAskari: @pascal, I presume that shape of ```dy``` and ```sm``` are always the same. Right? So I don't understand why such a check has been here

ReyhaneAskari: @Florian, if needed, you also need to check if ```dx``` has the same shape as ```sm```.

theano/tensor/nnet/nnet.py
```diff
@@ -645,79 +701,98 @@ def c_headers(self):
...
+            last_axis = ndim_x - 1;
+            // Get numpy iterator
+            PyArrayIterObject *it_x, *it_sm;
+            it_x = (PyArrayIterObject *) PyArray_IterAllButAxis((PyObject *) %(x)s, &last_axis);
+            it_sm = (PyArrayIterObject *) PyArray_IterAllButAxis((PyObject *) %(sm)s, &last_axis);
+            // Compute the stride on the selected dimension
+            stride_x = PyArray_STRIDE(%(x)s, last_axis)/sizeof(dtype_%(x)s);;
+            stride_sm = PyArray_STRIDE(%(sm)s, last_axis)/sizeof(dtype_%(sm)s);
+            // Get the shape on the specified dimension
+            num_data = shape_x[last_axis];
```
ReyhaneAskari: Use a better name instead of ```num_data```. Maybe, ```num_sm_classes```

theano/tensor/nnet/nnet.py
```diff
@@ -465,119 +503,138 @@ def c_code_template(dtype):
...
-                sm_i[j * Ssm1] *= sum_inv;
-            }
-
+                sm_i[j * stride_sm] *= sum_inv;
+                }
             """ % locals()
 
             inside_row_loop = """
-            if(Ssm1 == 1){
                 %(inside_row_loop_contig)s
```
ReyhaneAskari: What if it's not contiguous? Shouldn't you keep the else here? Or I'm missing where are you using
```inside_row_loop```?

File: theano/tensor/nnet/nnet.py (hunk 20 above)
Comment (bordesf) [inline theano/tensor/nnet/nnet.py]: I already check that few lines below (Between 384 and 393).

File: theano/tensor/nnet/nnet.py (hunk 22 above)
Comment (bordesf) [inline theano/tensor/nnet/nnet.py]: I corrected that in a new version that I didn't committed yet.

File: theano/tensor/nnet/nnet.py (hunk 21 above)
Comment (bordesf) [inline theano/tensor/nnet/nnet.py]: Yeah, it will be more meaningful, thanks

File: theano/tensor/nnet/nnet.py (hunk 5 above)
Comment (bordesf) [inline theano/tensor/nnet/nnet.py]: In the old version of the code (The current one of Theano), they check if dy and sm has the same shape in the C code. So, to be consistent with the C code, I though it will be needed to do the same things in the python code.

**Comment by bordesf** [comment]
I modified the code in order to select the axis on where we want to compute the softmax. I added an input node "axis" and I was wondering which method should I use about the gradient of this new input. For now, I used the DisconnectedType but maybe disconnected_grad or consider_constant would be better. What do you think ?

**Comment by ReyhaneAskari** [comment]
I don’t know why you have chosen to have a node for the axis and not simply an integer. I presume we rarely need to change it on the run time. @lamblin what do you think?

File: theano/tensor/nnet/nnet.py [hunk 23]
```diff
@@ -54,35 +54,30 @@ class SoftmaxWithBias(gof.Op):
...
     __props__ = ()
 
-    def make_node(self, x, b):
+    def make_node(self, x, b, axis=-1):
         x = tensor.as_tensor_variable(x)
         b = tensor.as_tensor_variable(b)
-        if x.type.ndim != 2 \
-                or x.type.dtype not in tensor.float_dtypes:
-            raise ValueError('x must be 2-d tensor of floats')
+        axis = tensor.as_tensor_variable(axis)
```
lamblin: if `axis` is supposed to be a scalar, you should also check that its number of dimensions is 0.

theano/tensor/nnet/nnet.py
```diff
@@ -130,281 +127,372 @@ def c_code_template(dtype):
...
+            // Check input types
+            if ((PyArray_TYPE(%(x)s) != NPY_DOUBLE) &&
+                    (PyArray_TYPE(%(x)s) != NPY_FLOAT))
+            {
+                    PyErr_SetString(PyExc_TypeError, "Not a float");
+                    %(fail)s;
+                    }
+
+            // Check if the axis is valid
+            if(axis > ndim_x - 1 || axis < -1){
```
lamblin: A negative axis means to use `ndim_x + axis` instead. It is not just a special case for `-1`, it works for all negative axes in numpy.
So the checks should be something like:
```C
if (axis < 0)
    axis += ndim_x;
if (axis > ndim_x - 1 || axis < 0)
    PyErr_SetString(...);
```

theano/tensor/nnet/nnet.py
f0k: The question is whether this should be symbolic at all. `T.sum()`, `T.max()` etc. have nonsymbolic `axis` arguments.

lamblin: Non-symbolic arguments would be fine by me too.
It could make some things much easier, for instance shape inference.

f0k: > It could make some things much easier, for instance shape inference.

... and optimization (we will want to use cuDNN if `axis==1` and `x.ndim in (4,5)`), and replacing it with an expression graph that uses `T.sum()` and `T.max()` (if ever the compiled expression graph becomes as fast as the manual C code).

f0k:
Hmm, why do we need an abstract softmax? Can anybody enlighten me? Is it that implementing a Softmax Op supporting arbitrary axes is too complicated, and you'd rather add an optimizer inserting a dimshuffle->reshape->softmax->reshape->dimshuffle graph as needed?

ReyhaneAskari:
@f0k , well I think he has used the abstract softmax for the general case where we can have a list of axes to apply the softmax on. In that case, he uses the old softmax when the list has only one element. The purpose is to not lose the optimizations that already exist for the old softmax. 

Implementing an op that would  _dimshuffle->reshape->softmax->reshape->dimshuffle_ would have overhead. We now have the implementations in C code which is much faster.

f0k:
> I think he has used the abstract softmax for the general case where we can have a list of axes to apply the softmax on. In that case, he uses the old softmax when the list has only one element.
> [...]
> Implementing an op that would dimshuffle->reshape->softmax->reshape->dimshuffle would have overhead. We now have the implementations in C code which is much faster.

If the plan is to eventually implement the general multi-axis softmax in C, then I don't see the reason for the abstract/placeholder softmax. What is the advantage over extending the old softmax to support multiple axes?

ReyhaneAskari:
Implementing a softmax that would cover all the cases is very hard at this point. We would rather continue this approach and add the functionality to theano. @bordesf can probably explain better in more details.

f0k:
> Implementing a softmax that would cover all the cases is very hard at this point.

I agree. This will be complicated in C.

> We would rather continue this approach and add the functionality to theano.

So I'm back at assuming there will be an optimizer replacing the abstract softmax with a dimshuffle->reshape->(old)softmax->reshape->dimshuffle whenever the abstract softmax has multiple axes (and they don't coincide with what's supported by cuDNN softmax, which would be another optimizer)?

lamblin:
@f0k: the first option is correct: optimize to cuDNN if possible, otherwise dimshuffle / reshape if multiple axes, otherwise the regular Op if only one axis. We may add other specially-optimized cases later if needed.

theano/tensor/nnet/nnet.py
```diff
@@ -130,284 +131,574 @@ def c_code_template(dtype):
...
 
-    def make_node(self, dy, sm):
+    def make_node(self, dy, sm, axis=-1):
         dy = tensor.as_tensor_variable(dy)
         sm = tensor.as_tensor_variable(sm)
-        if dy.type.ndim not in (1, 2) \
-                or dy.type.dtype not in tensor.float_dtypes:
-            raise ValueError('dy must be 1-d or 2-d tensor of floats. Got ',
-                             dy.type)
+        axis = tensor.as_tensor_variable(axis)
```
bordesf: If we want to keep the axis as a node in the graph, we still need to define it as a tensor variable.
Yeah, I will do some cleaning in the code.

ReyhaneAskari: Why do we want to have the axis as a node?

theano/tensor/nnet/nnet.py
```diff
@@ -130,329 +132,496 @@ def c_code_template(dtype):
...
+            x = tensor.shape_padleft(x, n_ones=1)
+        # Check if axis is an integer
+        if axis.type.dtype not in tensor.int_dtypes:
+            raise ValueError('axis must be an integer. Got ', axis.type)
+        return Apply(self, [x, axis], [x.type()])
+
+    def L_op(self, inp, outputs, grads):
+        x, axes = inp
+        # To investigate
+        if len(grads) == 2:
```
bordesf: During the rop test, len of grads can be two and I don't really get why it happens, it should be one. Do you have an idea ? @ReyhaneAskari

ReyhaneAskari: Well I don't really know but I presume that in the case of Softmax with Bias, ```grads[1]``` is wrt bias term . I think you can simply ignore it. Also, you don't need to define an ```if .. else``` here.  ```g_sm, = grads``` will always take the first element. @lamblin am I right?

f0k: > `g_sm, = grads` will always take the first element.

No, it will complain when grads is not an iterable of size 1. `g_sm = grads[0]` would work for both cases. Or `g_sm, *_ = grads` in Python 3.

bordesf: @f0k Yes, I will use `g_sm = grads[0]`, thanks.
@ReyhaneAskari I don't think it corresponds to a bias term because in the current Theano code,  we have `g_sm, = grads` and it means that grads is a list of length one (Otherwise as @f0k said, it will not work). I think, I have done something inside the optimization that add this term (Maybe it corresponds to an axis, when I run the rop.test_softmax tests, the second element of grads is `[Elemwise{second,no_inplace}.0]` that corresponds to `[0]`. I would like to understand what's going on. `g_sm, = grads[0]` works but I am wondering if this second element in grads can have any kind of importance.

bordesf:
I add a PR #5980 where I introduce a function that normalize the axes to avoid some code redundancy.

theano/tensor/nnet/nnet.py
f0k: > Maybe it corresponds to an axis

@bordesf: I haven't fully followed the code, but based on your comment, is the axis still registered as an input node for the softmax op? Wouldn't it be easier to make it a nonsymbolic attribute? We don't want the axis/axes to be changed at runtime anyway.

bordesf: I spoke with @lamblin and as consequences of having axis as a node, we get the grad of the axis in eval_points, since this grad isn't defined we don't need to pass it through the L_op. So I put as argument eval_point[0] and it resolves the issue.
Yes, we keep the axis as input node. It simplifies the access of the axis value during the optimization. I think @lamblin can give other reasons about this choice.

lamblin: Honestly, both could work. Having it an input will generate fewer versions of the C code, but that could be mitigated by making it a Param instead. I'll try to see with @notoraptor if that would be a better solution.

theano/tensor/nnet/nnet.py
```diff
@@ -93,23 +93,25 @@ def perform(self, node, input_storage, output_storage):
         if x.dtype == 'float16':
             x = x.astype('float32')
 
-        x_plus_b = x + b[None, :]
-        e_x = np.exp(x_plus_b - x_plus_b.max(axis=1)[:, None])
-        e_x *= 1.0 / e_x.sum(axis=1)[:, None]
+        # We add bias on the last dimension
+        x_plus_b = x + b
```
ReyhaneAskari: Hi @bordesf ,

Here, you cannot sum x and b as they don't have the same shape. You need to make it broadcastable first. Also you need to change the C code accordingly.
A test case:
```python
a = T.tensor3('a')
b = T.vector('b')
res =  softmax_with_bias(a, b, 0)
f = theano.function([a, b], res)
a_ = np.random.rand(3, 4, 5)
b_ = np.random.rand([3, 4, 5][axis])
res_ = f(a_, b_)
```
```
ValueError: operands could not be broadcast together with shapes (3,4,5) (3,) 
Apply node that caused the error: SoftmaxWithBias(a, b, TensorConstant{0})
Toposort index: 0
Inputs types: [TensorType(float64, 3D), TensorType(float64, vector), TensorType(int8, scalar)]
Inputs shapes: [(3, 4, 5), (3,), ()]
Inputs strides: [(160, 40, 8), (8,), ()]
Inputs values: ['not shown', array([ 0.66083873,  0.16459271,  0.09435837]), array(0, dtype=int8)]
Outputs clients: [['output']]```

File: theano/tensor/nnet/nnet.py [hunk 28]
```diff
@@ -800,15 +1143,94 @@ def softmax_graph(c):
...
+        axis = [int(axis)]
+    elif isinstance(axis, (tuple, list, np.ndarray)):
+        axis = [int(a) for a in axis]
+        if axis == list(range(c.type.ndim)):
+            axis = None
+    elif isinstance(axis, Variable):
+        # If axis is None, we select the last axis
+        if NoneConst.equals(axis):
+            axis = -1
+        elif not isinstance(axis, tensor.var.TensorConstant):
```
ReyhaneAskari: There is no ```tensor.var.TensorConstant```. You probably meant ```tensor.TensorConstant```?

theano/tensor/nnet/nnet.py
```diff
@@ -800,15 +1143,94 @@ def softmax_graph(c):
...
+            raise TypeError("Softmax needs a constant axis. Got %s" % axis)
+        else:
+            assert axis.dtype in integer_dtypes
+            if isinstance(axis.data, (integer_types, np.integer)) or \
+                    (isinstance(axis.data, np.ndarray) and axis.data.ndim == 0):
+                        axis = [int(axis.data)]
+            elif isinstance(axis.data, (list, np.ndarray)):
+                axis = [int(i) for i in axis.data]
```
ReyhaneAskari: Why not simply ```axis = axis.data```? Why the casting is needed?

theano/tensor/nnet/nnet.py
```diff
@@ -800,15 +1143,94 @@ def softmax_graph(c):
...
+        axis = [int(axis)]
+    elif isinstance(axis, (tuple, list, np.ndarray)):
+        axis = [int(a) for a in axis]
+        if axis == list(range(c.type.ndim)):
+            axis = None
+    elif isinstance(axis, Variable):
+        # If axis is None, we select the last axis
+        if NoneConst.equals(axis):
+            axis = -1
+        elif not isinstance(axis, tensor.var.TensorConstant):
```
ReyhaneAskari: The same issue here.

theano/tensor/nnet/nnet.py
```diff
@@ -130,329 +132,486 @@ def c_code_template(dtype):
...
+        inputs = [x, axis]
+        return Apply(self, inputs, [x.type()])
 
     def perform(self, node, input_storage, output_storage):
-        x, = input_storage
-        e_x = np.exp(x - x.max(axis=1)[:, None])
-        sm = e_x / e_x.sum(axis=1)[:, None]
+        x, axes = input_storage
+        # Apply softmax on the specified dimension
+        e_x = np.exp(x - x.max(axis=axes, keepdims=True))
```
ReyhaneAskari: Try this example:
```python
a = T.tensor3('a')
axis = T.constant([0, 1, -1])
res = softmax(a, axis=axis)
f = theano.function([a], res)
a_ = np.random.rand(3, 4, 5)
res_ = f(a_)
print res_.shape
```
When we reach here, ```axes``` is ```array([0, 1, 2])``` which leads to an error. A fix that comes to my mind is to cast it to tuple like ```x.max(axis=tuple(axes), keepdims=True)```.

theano/tensor/nnet/nnet.py
```diff
@@ -130,329 +132,486 @@ def c_code_template(dtype):
...
+        sm = e_x / e_x.sum(axis=axes, keepdims=True)
```
ReyhaneAskari: The same here.

theano/tensor/nnet/nnet.py
```diff
@@ -577,57 +773,57 @@ def c_code_template(dtype):
...
+        return Apply(self, [x, axis], [x.type()])
 
     def perform(self, node, input_storage, output_storage):
-        x, = input_storage
-        xdev = x - x.max(axis=1)[:, None]
-        lsm = xdev - np.log(np.sum(np.exp(xdev), axis=1,
-                            keepdims=True))
+        x, axes = input_storage
+        xdev = x - x.max(axis=axes, keepdims=True)
+        lsm = xdev - np.log(np.sum(np.exp(xdev), axis=axes, keepdims=True))
```
ReyhaneAskari: The problem with tuple here as well.

bordesf:
@ReyhaneAskari Thanks for your review, I push some commits to correct what you mentioned. About the axes, I used the function of PR #5980. A todo for the next days, is to raise an error in case of the len of axis is superior to one when using the old softmax.

theano/tensor/nnet/nnet.py
```diff
@@ -401,6 +396,11 @@ def make_node(self, dy, sm, axis=-1):
             dy = tensor.shape_padleft(dy, n_ones=1)
         if sm.ndim == 1:
             sm = tensor.shape_padleft(sm, n_ones=1)
+        # Check if axis is a scalar
+        try:
+            opt.get_scalar_constant_value(axis)
+        except tensor.NotScalarConstantError:
+            raise ValueError('axis must be a scalar. Use abstract_softmax instead Got ', axis.type)
         return Apply(self, [dy, sm, axis], [sm.type()])
```
f0k: Sorry, but I still don't get why `axis` is a symbolic input node at all. `theano.tensor.elemwise.Sum` has a nonsymbolic tuple of axes:
```
>>> import theano.tensor as T
>>> x = T.matrix('x')
>>> y = x.sum(axis=-1)
>>> y.owner.op.axis
(1,)
>>> y.owner.inputs
[x]
```
What's the advantage of making it symbolic?

bordesf: If there are no advantages of making it symbolic, it makes sens that it should become nonsymbolic. I will wait the answer of @lamblin and @notoraptor before making new commits.

bordesf: After discussion, we agreed to use a params type: http://deeplearning.net/software/theano_versions/dev/library/gof/params_type.html instead of the symbolic axis.

theano/tensor/basic.py
```diff
@@ -968,6 +968,53 @@ def _pack(x):
...
+    Returns
+    -------
+    axis: list of integers
+    """
+    x = as_tensor_variable(x)
+    if axis is None:
+        axis = []
+    elif (isinstance(axis, (integer_types, np.integer)) or
+            (isinstance(axis, np.ndarray) and axis.ndim == 0)):
+                axis = [int(axis)]
```
ReyhaneAskari: I know this is merged but you need to fix a corner case. If the ```axis=np.array(1.5)``` it will be floored to 1 while we should raise an error.

theano/tensor/basic.py
```diff
@@ -968,6 +968,53 @@ def _pack(x):
...
+    elif isinstance(axis, (tuple, list, np.ndarray)):
```
ReyhaneAskari: Check the type here as well to be int.

theano/tensor/nnet/nnet.py
```diff
@@ -130,329 +130,496 @@ def c_code_template(dtype):
...
+    where :math:`K` is the total number of neurons in the layer. This
+    activation function gets applied on the selected axis (Default axis is the last).
+    This function get applied when axis is a scalar and uses a fast C optimization.
+    """
+
+    nin = 2  # tensor, axis
+    nout = 1
+    __props__ = ()
+
+    def make_node(self, x, axis=-1):
```
ReyhaneAskari: At this point we already processed the axis and it should be a list of valid axes. So no default value is needed.

theano/tensor/nnet/nnet.py
```diff
@@ -130,329 +130,496 @@ def c_code_template(dtype):
...
+        x = tensor.as_tensor_variable(x)
+        axis = tensor.as_tensor_variable(axis)
+        if x.ndim == 1:
+            x = tensor.shape_padleft(x, n_ones=1)
+        # Check if axis is an integer
+        if axis.type.dtype not in tensor.int_dtypes:
```
ReyhaneAskari: This check is also already done in check_and_normalize_axis. Remove this.

theano/tensor/nnet/nnet.py
```diff
@@ -130,329 +130,496 @@ def c_code_template(dtype):
...
+            raise ValueError('axis must be an integer. Got ', axis.type)
+        # Check if axis is a scalar
+        try:
+            opt.get_scalar_constant_value(axis)
+        except tensor.NotScalarConstantError:
```
ReyhaneAskari: I feel like this check is unnecessary. Again the reason is that axis is already processed in check_and_normalize_axis. Am I missing something? Could you come up with a case that check_and_normalize_axis passes but here the axis variable is not scalar? If so, I presume this check must be moved to check_and_normalize_axis.

I appreciate it if you reply to my comments whenever you fix them by a new commit. I want to make sure that you've seen these.

theano/tensor/nnet/nnet.py
ReyhaneAskari: @bordesf , any updates? This is still failing.

theano/tensor/nnet/nnet.py
ReyhaneAskari: Still failing.

theano/tensor/nnet/nnet.py
```diff
@@ -2445,7 +2815,7 @@ def elu(x, alpha=1):
         "Fast and Accurate Deep Network Learning by
         Exponential Linear Units (ELUs)" <http://arxiv.org/abs/1511.07289>`.
     """
-    return tensor.switch(x > 0, x, alpha * tensor.expm1(x))
+    return tensor.switch(x > 0, x, alpha * (tensor.exp(x) - 1))
```
ReyhaneAskari: For small values of x, tensor.expm1(x) provides higher precision. So it's better not to remove this.

theano/tensor/nnet/nnet.py
```diff
@@ -1435,8 +1804,10 @@ def make_node(self, coding_dist, true_one_of_n):
         """
         _coding_dist = tensor.as_tensor_variable(coding_dist)
         _true_one_of_n = tensor.as_tensor_variable(true_one_of_n)
+        if _coding_dist.type.ndim == 1:
+            _coding_dist = tensor.shape_padleft(_coding_dist, n_ones=1)
         if _coding_dist.type.ndim != 2:
-            raise TypeError('matrix required for argument: coding_dist')
+            raise TypeError('iatrix required for argument: coding_dist')
```
ReyhaneAskari: Typo.

theano/tensor/nnet/nnet.py
```diff
@@ -800,15 +1155,20 @@ def softmax_graph(c):
...
-    return softmax_op(c)
+    axis = tensor.check_and_normalize_axes(c, axis)
+    return softmax_op(c, axis)
 
 
-def logsoftmax(c):
-    return logsoftmax_op(c)
+def logsoftmax(c, axis=(-1,)):
+    c = tensor.as_tensor_variable(c)
+    if c.broadcastable[-1]:
```
ReyhaneAskari: The same here.

theano/tensor/nnet/nnet.py
```diff
@@ -637,108 +841,258 @@ def R_op(self, inputs, eval_points):
...
+    activation function gets applied on the selected axis (Default axis is the last).
+    This function get applied when axis is a scalar and uses a fast C optimization.
+    """
+    __props__ = ()
+
+    def make_node(self, x, axis=-1):
+        x = tensor.as_tensor_variable(x)
+        axis = tensor.as_tensor_variable(axis)
+        # Check if axis is an integer
+        if axis.type.dtype not in tensor.int_dtypes:
```
ReyhaneAskari: We checked these already. See my other comment.

theano/tensor/nnet/nnet.py
bordesf: check_and_normalize_axis return a list of sorted integers. This Softmax op with the C code optimization corresponds to the old op that manages only the case where we specify **only one axis**. So, if check_and_normalize_axis return (2,3) and this op is called instead of the Abstract_Softmax, it will fail because the C code doesn't manage more than one axis. I think we should really keep it because in case where there are some issue in the optimization it will be much easier to detect the case where a wrong op is called.

ReyhaneAskari: Thanks for the explanation. Maybe we could add a comment here that explains what you expect.

ReyhaneAskari:
@bordesf I did the minor fixes so that you can focus on replacing the symbolic nodes by param types and @shawntan  can start working on the tests.

bordesf:
Since this pull request implies too much modifications of the code, after discussion with @lamblin and @ReyhaneAskari, in order to improve the readability and to make the review easier, I will create different pull requests with few features for each.

ReyhaneAskari:
thanks @bordesf. Please tag me in your new PRs.

PR #5721, opened 2017-03-17
Index overflow fixes

nouiz:


rebecca-palmer:
See latest comment on #5525 before merging this.

Issue #5709, opened 2017-03-14
Remove from *InplaceTester the grad tests


nouiz:
This is already tested. It don't test the inplace of the grad, so I don't see any utility of this except duplicating existing tests.

Before doing it, I would like a confirmation by  @abergeron or @lamblin that I don't forget something.

abergeron:
I agree.

PR #5732, opened 2017-03-19
skip_verify_grad added to makeTester for all InplaceTester

shubhvachher:
For issue #5709.

All InplaceTests (that use inplace=True) have the skip_verify_grad flag enabled.

Added skip_verify_grad flag to the makeTester function in case verify_grad needs to be skipped for other cases as well.

PS: Sorry I did not inform I was working on this, came across the issue and solved while trying to go through codebase

### DIFF
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -242,7 +242,7 @@ def upcast_int8_nfunc(fn):
 def makeTester(name, op, expected, checks=None, good=None, bad_build=None,
                bad_runtime=None, grad=None, mode=None, grad_rtol=None,
                eps=1e-10, skip=False, test_memmap=True, check_name=True,
-               grad_eps=None):
+               grad_eps=None, skip_verify_grad=False):
     # :param check_name:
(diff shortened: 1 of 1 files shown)


nouiz:
This PR need a rebase, but I think it isn't the right fix. The fight fix would be to just remove the grad parameter to the *InplaceTester. Otherwise, having the param and just ignoring it could lead to confusion I think.

Can you do this alternative fix?

thanks

shubhvachher:
Yes, sorry. rebasing!

Also, @nouiz Could you please explain if there was any use to line del kwargs['inplace'] in makeBroadcastTester. If I do not del that argument then that can be used as the flag to make grad empty {}

Shubh

nouiz:
The del is there as the makeTester() fct call on the line bellow the del don't accept an inplace parameter. It don't need it and don't take it as a parameter, so it would crash if you remove that line.

shubhvachher:
Uh, @nouiz. Please check pr #5746. I don't know why that didn't show up here as another commit?

Issue #5731, opened 2017-03-18
ability to change libgpuarray preallocate value after importing theano to CPU


astooke:
Hi,

  It would be very helpful to be able to change the `gpuarray.preallocate` setting in python, after `import theano` (to CPU only), and before using `theano.gpuarray.use()`.  

Not sure whether preferable to make that config value editable or to allow `gpuarray.use()` to take an optional argument, I think either would work equally.

Thanks,
Adam

astooke:
OK that was really simple, just change to `allow_override=True` for `gpuarray.preallocate` in `theano.configdefaults`.   I suggest making this standard unless there is some reason not to?

lamblin:
The main reason I have not to do that is that it would build the expectation that the allocated memory size can be changed _at any point_ during the execution.
Moreover, it is only useful when using `device=cpu` but manually calling `gpuarray.use()`, and we do not have a way of allowing changes in the configuration options only until `gpuarray.use()` is called.

I would actually support `preallocate` being an argument to `use` and `init_dev`, and defaulting to `config.gpuarray.preallocate` if it is not provided. It would make it easier to pre-allocate different amounts of memory on different devices as well.

Finally, most of what `gpuarray.preallocate` is doing is calling `pygpu.empty()` to literally pre-allocate a block of the right size, which is something you could do just after calling `gpuarray.use()` as well.

lamblin:
#5735 makes it possible to specify a value for `preallocate` when calling `gpuarray.use()`.

astooke:
Great! Thank you!

Issue #4817, opened 2016-08-02
Slow compilation with device=cuda0


juancamilog:
I have a script that builds a small multilayer perceptron using lasagne 0.2.  The script is available here [https://drive.google.com/open?id=0B-j0Gj1S1Cb9UU5LNXFEeUJtQ3M](https://drive.google.com/open?id=0B-j0Gj1S1Cb9UU5LNXFEeUJtQ3M)

If you run the script as follows:

THEANO_FLAGS='device=gpu,floatX=float32' python testNN.py

compilation takes about 2 seconds on my laptop (i7 6700HQ, GTX 960M). When running as,

THEANO_FLAGS='device=cuda0,floatX=float32' python testNN.py

compilation times increase considerably ( to about 30 seconds ). For more complex models (like feeding the output to the network to its input inside a scan loop), the slowdown is even bigger.

Note that there is no slow down at run time. In my case, using cuda0 is actually a bit faster at run time.

Inside the zip file you'll find the python script to reproduce the error, along with the outputs of pydotprint and debugprint for all of the compiled functions. These correspond to a two layer network with 4 inputs, 256 hidden units in each layer, and 2 outputs.
It looks like the graphs in the cuda0 case are actually smaller (which looks like a good thing)

nouiz:
I started to look. I have found dtype problem:

NN.py(235)predict_symbolic()
-> C = theano.tensor.slinalg.solve_lower_triangular(Lx,C)
(Pdb) l
230              # empirical covariance
231              S = theano.tensor.diag(theano.tensor.exp(self.logsn2)) +
y.T.dot(y)/n - theano.tensor.outer(M,M)
232              # Sx^-1 times empirical input output covariance
233              if Sx is not None:
234                  C = x.T.dot(y)/n - theano.tensor.outer(mx,M)
235  ->                C =
theano.tensor.slinalg.solve_lower_triangular(Lx,C)

Here Lx is float64 and not float32. So the solve isn't moved to the GPU and
print warnings.

nouiz:
also, that op isn't ported to the new back-end. So if you fix the dtype,
then it could be faster to execute in the old back-end until it get ported.

nouiz:
I got some result. The compilation is done in 2 phases. The extra time
consuming is in the second phases called Linker.

We currently cache a less think in the new back-end. It is on our todo to
fix that. Not sure when.

juancamilog:
[...]

PR #5735, opened 2017-03-20
Disk cache option for gpuarray

abergeron:
This goes with https://github.com/Theano/libgpuarray/pull/335

also fix gh-5731
fix https://github.com/Theano/Theano/issues/4817

### DIFF
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -431,6 +431,7 @@ import theano and print the config variable, as in:
         automatically to get more memory. But this can cause
         fragmentation, see note above.
 
+
 .. attribute:: config.gpuarray.sched
 
     String value: ``'default'``, ``'multi'``, ``'single'``
@@ -460,6 +461,13 @@ import theano and print the config variable, as in:
     In the future when true multi-stream is enabled in libgpuarray,
     this may change.  If you want to make sure to have optimal
     performance, check both options.
+
+
+.. attribute:: config.gpuarray.cache_path
+
+   String value
+
+   Directory to cache pre-compiled kernels for the gpuarray backend.
 
 
 .. attribute:: linker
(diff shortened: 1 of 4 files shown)


theano/configdefaults.py
```diff
@@ -1792,7 +1792,7 @@ def default_compiledir():
     ConfigParam(
         lambda: os.path.join(config.compiledir, 'gpuarray_kernels'),
         filter=filter_base_compiledir,
-        allow_override=False)
+        allow_override=False),
```

nouiz:
The code seem good, still the doc update is missing. I'll test it later.

doc/library/config.txt
```diff
@@ -528,6 +528,13 @@ import theano and print the config variable, as in:
     performance, check both options.
 
 
+.. attribute:: config.gpuarray.cache_path
+
+   String value
+
+   Directory to cache pre-compiled kernels for the gpuarray backend.
```
nouiz: Tell that it default to the subdirectory gpuarray_kernels inside theano compile dir.

abergeron: done

nouiz:
When I run Theano tests, it don't speed them up and there is nothing in the
cache.

abergeron:
I stopped refresh from deleting the gpuarray cache directory so that we can use the cache.

slefrancois:
jenkins test this

nouiz:
I tested it and it speed up the test of gpuarray/tests/test_basic.py when Theano cache is full compared to the master.

Ok to merge after the memory leak fix is merged and the old back-end PR.

lamblin:
Jenkins test this again please

nouiz:
This PR need a rebase

lamblin:
I tried the new "Resolve conflicts" editor on GitHub, it may not have been a good idea.
I'll rebase instead if needed.

abergeron:
The automatic change resurrected the lib.cnmem documentation so I did a manual rebase instead.

PR #5746, opened 2017-03-21
skip_verify_grad added to makeTester for all InplaceTester

### LINKED ISSUE #5709: Remove from *InplaceTester the grad tests

shubhvachher:
close #5709 

Rebased. Thanks for the help @nouiz. 
Grad is being passed as None for all *InplaceTester.


theano/tensor/tests/test_basic.py
```diff
@@ -2159,6 +2164,7 @@ def fixed_first_input_iv(x):
     op=inplace.conj_inplace,
     expected=numpy.conj,
     good=_good_broadcast_unary_normal,
+    grad=None,
```
Sentient07: This `grad` parameter that you pass to `makeBroadcastTester` is a `kwarg` which is simply passed on to `makeTester` that sets the default value of `grad` to `None`. Instead, it is better you delete all these instances. If that works, that is a better fix than passing `None`.

shubhvachher: The `<makeTester>` function takes any `<grad=None>` and sets `<grad={}>`. Thus `<utt.verify_grad>` is never called in the function `<test_grad>`.

Although makeTester gives grad=None by default with *InplaceTester I would like to be explicit about the grad parameter so that it is understood that verify_grad tests are not done. 
I can remove all these and let makeTester default grad to None if necessary. Please review.

nouiz:
There was a conflict. I tested github online conflict resolution tool. Not super good. It introduce a merge commit and is so, next time I'll prefer personnaly to do it on the command line.

PR #5747, opened 2017-03-21
Add support for atomic{Exch,Add} on long longs.

abergeron:
This would break at compilation because cuda did not support it.

### DIFF
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -871,7 +871,7 @@ class GpuAdvancedIncSubtensor1_dev20(GpuKernelBase, HideC,
         return super(GpuAdvancedIncSubtensor1_dev20, self).perform(node, inp, out)
 
     def c_code_cache_version(self):
-        return (13,)
+        return (14,)
 
     def c_headers(self):
         return ['<numpy_compat.h>', '<gpuarray_helper.h>',
@@ -924,9 +924,8 @@ if (GpuArray_vector_add_fast(%(out)s, %(y)s, %(ind)s, %(params)s->set_instead_of
  */
 #if __CUDA_ARCH__ < 600
 __device__ ga_double atomicAdd(ga_double* address, ga_double val) {
-    unsigned long long int* address_as_ull =
-                                          (unsigned long long int*)address;
-    unsigned long long int old = *address_as_ull, assumed;
+    ga_ulong *address_as_ull = (ga_ulong *)address;
+    ga_ulong old = *address_as_ull, assumed;
     do {
         assumed = old;
         old = atomicCAS(address_as_ull, assumed,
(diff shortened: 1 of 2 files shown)


theano/gpuarray/subtensor.py
```diff
@@ -800,7 +800,7 @@ def make_node(self, x, y, ilist):
         """
         ctx_name = infer_context_name(x, y, ilist)
         x_ = as_gpuarray_variable(x, ctx_name)
-        y_ = as_gpuarray_variable(y, ctx_name)
+        y_ = as_gpuarray_variable(y.astype(x.dtype), ctx_name)
```
nouiz: Why not cast it inside the c code before calling atomicAdd or atomicExch around line 1000 in this file? It would  prevent a copy in memory and seem easy to do.

abergeron: I would prefer not to multiply the possible C code b y including a cartesian product of all the types.

Also, if we do it this way it gives the opportunity to merge the cast with another elemwise operation.

Finally since we have float16 support, there would need to be special-case handling of casting to and from that type which would complicate the kernel code.

nouiz:
jenkins test have an error.

lamblin:
Is this PR still needed?

nouiz:
This PR have conflict and need a rebase if it is still needed.

abergeron:
I think is it still needed.  I did a rebase.

nouiz:
jenkins have failure

lamblin:
The error is:
```
[secondBranch] ======================================================================
[secondBranch] ERROR: theano.gpuarray.tests.test_subtensor.test_advinc_subtensor1_dtype
[secondBranch] ----------------------------------------------------------------------
[secondBranch] Traceback (most recent call last):
[secondBranch]   File "/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
[secondBranch]     self.test(*self.arg)
[secondBranch]   File "/home/jenkins/workspace/Theano_PR/theano/gpuarray/tests/test_subtensor.py", line 121, in test_advinc_subtensor1_dtype
[secondBranch]     rep[[0, 2]] += yval
[secondBranch] TypeError: Cannot cast ufunc add output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'
```
It looks like NumPy is actually complaining from the mismatch for ('uint64', 'float64'). (I tested with 1.11 and 1.12.)

lamblin:
And I have no idea where the `float64` is coming from, since `yval` is in fact 'int8'.
Actually, NumPy seems to upcast 'int64' + 'uint8' to 'float64'.
`np.add.at(rep, [0, 2], yval)` seems to work though.

theano/gpuarray/tests/test_subtensor.py
```diff
@@ -93,14 +93,15 @@ def test_advinc_subtensor1():
                     for node in f.maker.fgraph.toposort()]) == 1
         rval = f(yval)
         rep = xval.copy()
-        rep[[0, 2]] += yval
+        np.add.at(rep, [0, 2], yval)
```
lamblin: That change should be done at line 121 instead...

nouiz:
close/reopen to trigger travis with the new master.

abergeron:
Rebased.

nouiz:
I killed travis to help fix travis timeout on the master branch faster

Issue #5742, opened 2017-03-21
flake8 sandbox/linalg/*


nouiz:
Only:

    "sandbox/linalg/ops.py",
    "sandbox/linalg/tests/test_linalg.py",

nouiz:
see https://github.com/Theano/Theano/issues/5744

lamblin:
Closed by #5756

PR #5750, opened 2017-03-22
#5742 Removed theano/sandbox/linalg/ops.py and theano/sandbox/linalg/tests/test_linalg.py from flake8 whitelist

shawntan:
Removed the following files from flake8  whitelist:
- `sandbox/linalg/ops.py`
- `sandbox/linalg/tests/test_linalg.py`

@ReyhaneAskari #5742


theano/sandbox/linalg/ops.py
```diff
@@ -280,8 +244,6 @@ def transinv_to_invtrans(node):
 @register_stabilize
 @local_optimizer([Dot, Dot22])
 def inv_as_solve(node):
-    if not imported_scipy:
```
shawntan: Not quite sure if allowing the execution to go through here is the right thing to do, since `scipy.linalg` is not imported.

But since the library is never used, I suppose the check is not needed anyway, and an alternative has been found.

nouiz:
just a note that travis have some failures.

theano/sandbox/linalg/ops.py
lamblin: The call to `solve(...)` should fail if `scipy` is not present, and the `Solve` node that it builds actually needs `scipy`, see [here](https://github.com/shawntan/Theano/blob/f1f1cc087ff1a10896240ee566b4e4b155353ef7/theano/tensor/slinalg.py#L199-L200).
Did that show up as a flake8 violation?

lamblin:
Also, it looks like some imports in `sandbox/linalg/ops.py` from `theano.tensor.slinalg` or `nlinalg` because they used to be defined in sandbox, but have been moved. Having imports there means that we can still import the names from `sandbox.linalg.ops`, which some tests do.

Since this is sandbox, I think we do not care that much about compatibility, and the solution would be to update the imports in `sandbox/linalg/tests/test_linalg.py`.

For the failure in pickle, I'm not really sure what is going on. Are you able to import `theano.sandbox.linalg.ops` manually?

ReyhaneAskari:
@shawntan will create another pull request.

theano/sandbox/linalg/ops.py
shawntan: Yes it did.

I propose importing `imported_scipy` from `theano.tensor.slinalg` as a fix?

nouiz:
that seem good.

PR #5751, opened 2017-03-22
MultinomialWOReplacementFromUniform := ChoiceFromUniform

laurent-dinh:
Following instructions by @nouiz from issue #5408.

### DIFF
--- a/theano/sandbox/multinomial.py
+++ b/theano/sandbox/multinomial.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, print_function, division
 import numpy
+import warnings
 
 import theano
 from theano import Op, Apply
@@ -211,7 +212,7 @@ class MultinomialFromUniform(Op):
                 z[0][n, numpy.searchsorted(cumsum, unis_n)] += 1
 
 
-class MultinomialWOReplacementFromUniform(MultinomialFromUniform):
+class ChoiceFromUniform(MultinomialFromUniform):
     """
     Converts samples from a uniform into sample (without replacement) from a
     multinomial.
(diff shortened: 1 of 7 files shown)


nouiz:
All that seem goods. It just miss the test to reload an old pickle that contain the old op to make sure we don't break that.

thanks

jenkins test this

theano/sandbox/tests/test_multinomial_wo_replacement.py
```diff
@@ -43,13 +46,13 @@ def test_select_distinct(self):
...
         """
-        Tests that MultinomialWOReplacementFromUniform fails when asked to sample more
+        Tests that ChoiceFromUniform fails when asked to sample more
         elements than the actual number of elements
         """
         p = tensor.fmatrix()
         u = tensor.fvector()
         n = tensor.iscalar()
-        m = multinomial.MultinomialWOReplacementFromUniform('auto')(p, u, n)
+        m = multinomial.ChoiceFromUniform('auto')(p, u, n)
```
nouiz: Similar rename is probably needed in the file theano/sandbox/rng_mrg.py. To make sure we didn't miss such changes, can you run Theano tests while commenting the line "MultinomialWOReplacementFromUniform = ChoiceFromUniform" and the equivalent GPU one? If there is rename forgotten, then this will causes errors.

theano/sandbox/multinomial.py
```diff
@@ -626,3 +626,6 @@ def local_gpu_multinomial(node):
             # The dimshuffle is on the cpu, but will be moved to the
             # gpu by an opt.
             return [gpu_from_host(ret)]
+
+
+MultinomialWOReplacementFromUniform = ChoiceFromUniform
```
nouiz: I think we should make this raise a warning to help people know about the deprecation. I think you can follow this example, but change the method to just print a warning:

```
+# This is for compatibility with pickled things.  It should go away at
+# some point.
+class DownsampleFactorMaxGrad(object):
+    def __new__(self, ds, ignore_border, st=None, padding=(0, 0), mode='max'):
+        if mode == 'max':
+                return MaxPoolGrad(ds=ds, ignore_border=ignore_border, st=st,
+                                   padding=padding, mode='max')
+        else:
+            return AveragePoolGrad(ds=ds, ignore_border=ignore_border, st=st,
+                                   padding=padding, mode=mode)
```

theano/sandbox/tests/test_multinomial_wo_replacement.py
laurent-dinh: Done.

theano/sandbox/multinomial.py
laurent-dinh: Done.

PR #5755, opened 2017-03-23
Use Op params for GpuDnnSoftmax

notoraptor:
Reduce C code by using op params for GpuDnnSoftmax.

This PR depends on:
 * PR for Wrapper that allows to make structs: #5612
 * PR for EnumType that allows to create constants: #5710 

Here I update these PRs with some fixes, improvements and adjustments, and I finally use op params for `GpuDnnSoftmaxBase`, `GpuDnnSoftmax` and `GpuDnnSoftmaxGrad`. To help see differences with the other PRs, all these changes are in one commit (the current last): https://github.com/notoraptor/Theano/commit/6b204315c85ae41fb3abbdeb8668b3560864dad2 .

*Summary of changes added by this PR*:

 1) Fix conversion from integers to C enum in C code for Theano enum types.

2) Update Wrapper with new convenient methods to ease wrapping dicts and objects.

3) Update Wrapper interface to handle Theano enum types. Enum constants from enum types wrapped into a Wrapper are directly available as wrapper attributes. Add a convenient method ``Wrapper.get_enum(enum_name)`` to get enum constants by name.

4) Add a test to check Wrapper+enum interface.

5) Update Wrapper documentation.

6) Update default ``Op.get_params()`` implementation to use new Wrapper convenient methods.

7) **Add Op param to GpuDnnSoftmaxBase**

8) **Update C code for ``GpuDnnSoftmax`` and ``GpuDnnSoftmaxGrad``**.

*Changes tested with success with*:
```
nosetests -xvs theano/gpuarray/tests/test_dnn.py:test_SoftMax
```

@nouiz @abergeron @lamblin

### DIFF
--- a/theano/gof/params_type.py
+++ b/theano/gof/params_type.py
@@ -238,7 +238,7 @@ class ParamsType(Type):
         self.length = len(kwargs)
         self.fields = tuple(sorted(kwargs.keys()))
         self.types = tuple(kwargs[field] for field in self.fields)
-        self.name = self.__generate_struct_name()
+        self.name = self.generate_struct_name()
 
         self.__const_to_enum = {}
         self.__alias_to_enum = {}
@@ -246,10 +246,15 @@ class ParamsType(Type):
         if enum_types:
             # We don't want same enum names in different enum types.
             if sum(len(t) for t in enum_types) != len(set(k for t in enum_types for k in t)):
-                raise AttributeError('ParamsType: found different enum types with common constant names.')
+                raise AttributeError('ParamsType: found different enum types with common constants names.')
             # We don't want same aliases in different enum types.
             if sum(len(t.aliases) for t in enum_types) != len(set(alias for t in enum_types for alias in t.aliases)):
-                raise AttributeError('ParamsType: found different enum types with common constant aliases.')
+                raise AttributeError('ParamsType: found different enum types with common constants aliases.')
+            # We don't want aliases that have same names as some constants.
+            all_enums = {e for t in enum_types for e in t}
+            all_aliases = {a for t in enum_types for a in t.aliases}
+            if [a for a in all_aliases if a in all_enums]:
+                raise AttributeError('ParamsType: found aliases that have same names as constants.')
             # We map each enum name to the enum type in which it is defined.
             # We will then use this dict to find enum value when looking for enum name in Wrapper object directly.
             self.__const_to_enum = {enum_name: enum_type for enum_type in enum_types for enum_name in enum_type}
@@ -270,7 +275,7 @@ class ParamsType(Type):
     def __hash__(self):
         return hash((type(self),) + self.fields + self.types)
 
-    def __generate_struct_name(self):
+    def generate_struct_name(self):
         # This method tries to generate an unique name for the current instance.
         # This name is intended to be used as struct name in C code and as constant
         # definition to check if a similar ParamsType has already been created
@@ -396,7 +401,7 @@ class ParamsType(Type):
             value_for_c = False
 
             # Value for c can't be retrieved from o, so we add a value for that field in kwargs.
-            params1 = params_type.get_params(o, c=value_for_c)
+            params = params_type.get_params(o, c=value_for_c)
             # params.a contains 10
             # params.b contains [[1, 2, 3], [4, 5, 6]]
             # params.c contains value_for_c
@@ -519,7 +524,6 @@ class ParamsType(Type):
         struct_name = self.name
         struct_name_defined = struct_name.upper()
         c_support_code_set = set()
-        c_support_code_list = []
         c_declare_list = []
         c_init_list = []
         c_cleanup_list = []
@@ -527,10 +531,7 @@ class ParamsType(Type):
         for attribute_name, type_instance in zip(self.fields, self.types):
 
             try:
-                c_support_code_current = type_instance.c_support_code()
-                if c_support_code_current not in c_support_code_set:
-                    c_support_code_list.append(c_support_code_current)
-                    c_support_code_set.add(c_support_code_current)
+                c_support_code_set.add(type_instance.c_support_code())
             except MethodNotDefined:
                 pass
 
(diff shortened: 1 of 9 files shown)


notoraptor:
@nouiz Updated.

*NB*:
1) I still have to update documentations to take account of new interfaces added to Enum types and Wrapper.
2) I have added some utility methods to Wrapper. Maybe I can merge 2 or 3 of them to reduce number of Wrapper methods.

theano/gof/type.py
```diff
@@ -854,27 +854,44 @@ class EnumType(Type, dict):
...
+                                    'constant alias followed by constant value.' % type(self).__name__)
+                nickname, value = kwargs[k]
+                if not isinstance(nickname, str):
+                    raise TypeError('%s: constant alias should be a string, got "%s".'
+                                    % (type(self).__name__, nickname))
+                if nickname in self.aliases:
+                    raise TypeError('%s: consant alias "%s" already used.' % (type(self).__name__, nickname))
+                self.aliases[nickname] = k
+                kwargs[k] = value
             if isinstance(kwargs[k], bool):
```
nouiz: elif

theano/gpuarray/dnn.py
```diff
@@ -234,6 +234,9 @@ def get_params(self, node):
             ptr = ctx.cudnn_handle.value
             res = handle_type.make_value(ptr)
             ctx.cudnn_handle_param = res
+        if isinstance(self.params_type, Wrapper) and self.params_type.has_type(handle_type):
```
nouiz: If it is a Wrapper and it don't have handle_type, you should raise an error. Currently params_type will just be ignored.

theano/gof/type.py
```diff
@@ -997,10 +1019,26 @@ def __init__(self, *args, **kwargs):
             type(self).__name__ + ': expected 0 or only 1 extra parameter "ctype".'
         ctype = kwargs.pop('ctype', 'int')
```
nouiz: I saw elsewhere a pop ctype that default to double. why not always the same dtype as default?

theano/gof/type.py
notoraptor: In fact, I want to pass through this IF in any case: if kwargs[k] is a tuple above, the constant alias is extracted, and then the real constant value is put to kwargs[k], then it should be checked again here.

theano/gof/type.py
notoraptor: I use double as default ctype for base class EnumType (to allow constants to have double values), but I use int as default ctype for class EnumList, as here the constants are integers in any case.

theano/gpuarray/dnn.py
notoraptor: Done.

notoraptor:
A first rebase to take into account last updates in other PRs.

notoraptor:
New rebasing to take into account updates in master branch and renaming of params type.

I have dispatched the files changes in this specific PR into 5 different commits starting from https://github.com/Theano/Theano/pull/5755/commits/5612c844b63ff5069fc58bd45882398881a47ad2 .

notoraptor:
Important update: I had forgotten to take into account c_support_code() for types wrapped into ParamsType.

I have discovered the problem because enum types (e.g. CEnumType used for GpuDnnSoftmax) need support code in Python 3 for PyInt_AsLong/PyLong_AsLong.

It seems jenkins test GPU with Python 2, but not with Python 3 !

notoraptor:
Rebased with master branch ! Now there are just 6 commits to review !

theano/gof/params_type.py
```diff
@@ -191,7 +238,28 @@ def __init__(self, **kwargs):
...
+        self.__const_to_enum = {}
+        self.__alias_to_enum = {}
+        enum_types = [t for t in self.types if isinstance(t, EnumType)]
+        if enum_types:
+            # We don't want same enum names in different enum types.
+            if sum(len(t) for t in enum_types) != len(set(k for t in enum_types for k in t)):
+                raise AttributeError('Wrapper: found different enum types with common constant names.')
+            # We don't want same aliases in different enum types.
+            if sum(len(t.aliases) for t in enum_types) != len(set(alias for t in enum_types for alias in t.aliases)):
+                raise AttributeError('Wrapper: found different enum types with common constant aliases.')
```
nouiz: I think we should also check that we don't have alias that have the same name as enums name

theano/gof/params_type.py
```diff
@@ -202,7 +270,7 @@ def __eq__(self, other):
     def __hash__(self):
         return hash((type(self),) + self.fields + self.types)
 
-    def generate_struct_name(self):
+    def __generate_struct_name(self):
```
notoraptor: I was thinking this method should be private, as it should not be called outside this class (it is used to generate the C struct name, and this name is available via `self.name` after class is instantiated).

But, indeed, it may not be so useful.

theano/gof/params_type.py
```diff
@@ -309,12 +518,22 @@ def c_support_code(self):
...
         c_init_list = []
         c_cleanup_list = []
         c_extract_list = []
         for attribute_name, type_instance in zip(self.fields, self.types):
 
+            try:
+                c_support_code_current = type_instance.c_support_code()
+                if c_support_code_current not in c_support_code_set:
+                    c_support_code_list.append(c_support_code_current)
+                    c_support_code_set.add(c_support_code_current)
```
notoraptor: I find an ordered dict, but not an ordered set, into python collections.

I will use a set, and I will convert it to a sorted list just when necessary.

notoraptor:
Updated, @nouiz .

theano/gof/params_type.py
```diff
@@ -519,18 +524,14 @@ def c_support_code(self):
         struct_name = self.name
         struct_name_defined = struct_name.upper()
         c_support_code_set = set()
```
nouiz: Make this an orderedset. Otherwise, the code generated could change due to non related causes.

nouiz: ok, not needed, you sorted it instead which make it deterministic.

PR #5756, opened 2017-03-23
#5742 Removed theano/sandbox/linalg/ops.py and theano/sandbox/linalg/tests/test_linalg.py from flake8 whitelist

shawntan:
Removed the following files from flake8 whitelist:
- `sandbox/linalg/ops.py`
- `sandbox/linalg/tests/test_linalg.py`
@ReyhaneAskari #5742


theano/sandbox/linalg/ops.py
```diff
@@ -24,7 +24,9 @@
 from theano.tensor.slinalg import (Cholesky,
                                    cholesky,
                                    Solve,
-                                   solve)
+                                   solve,
+                                   imported_scipy)
```
shawntan: included the check from the original file so the check isn't done twice.

theano/sandbox/linalg/ops.py
```diff
@@ -1,66 +1,34 @@
 from __future__ import absolute_import, print_function, division
 import logging
-
-logger = logging.getLogger(__name__)
```
ReyhaneAskari: @lamblin , I know the ```logger``` in here generates flake8 errors, but wouldn't moving it down miss logging the imports?

shawntan: Ah right. I didn't consider that.

lamblin: The other theano sub-modules will have their own logger, so that is fine.
The logger is only used explicitly, for instance the calls to `logger.debug(...)` later in the same file.

ReyhaneAskari:
Good job on this PR Shawn!

PR #5758, opened 2017-03-23
Add sigmoid_binary_crossentropy function

jeffdonahue:
This adds a function `sigmoid_binary_crossentropy(pred, target)` which does the same computation (mathematically) as `binary_crossentropy(sigmoid(pred), target)`, but in a stabler and hopefully slightly more efficient way.  The gradient with respect to `pred` can be computed in a particularly elegant and simple way relative to how Theano currently computes the gradients (even with graph optimizations). See [this gist](https://gist.github.com/jeffdonahue/b0641bafc2df22d7bd46d953dd89ef0e) for a comparison of the computation graphs each one produces, and [this gist](https://gist.github.com/jeffdonahue/84eab3b74db5309da871a5abb78bdde6) for a comparison of numerical errors: relative to `binary_crossentropy`, in `FAST_RUN` mode, this version reduces numerical error (MSE) on standard Gaussian inputs by ~18% in the cross-entropy computation, ~43% for the gradient w.r.t. `pred`, and >99% for the gradient w.r.t. `target`.  (In `FAST_COMPILE` mode the error reductions are bigger.)

One almost always has access to the real-valued predictions before they are passed through `sigmoid` to produce probabilties in `(0, 1)`.  In those cases,`sigmoid_binary_crossentropy(p, t)` should be preferred over `binary_crossentropy(sigmoid(p), t)`.

This might be possible to do solely via graph optimization, and if so that would be preferable to having to make people update their code to use this new function (with an admittedly less clean interface), but I spent a while trying to make that work through pattern replacement and couldn't get there.  (I was very happy when I discovered the `inline` and `grad_overrides` kwargs for `OpFromGraph`, which is what's used here instead of graph optimization -- it was exactly what I'd been looking for to implement this and a couple other things.  Thanks @khaotik and reviewers of #5255!)

If this is potentially worth merging, I'm open to any suggested changes (including to the name of the function, which I'm not too happy with).  Otherwise if nothing else it could be a good example of using `OpFromGraph` with `grad_overrides`.


nouiz:
I think it is good like this with doc in the file doc/library/tensor/nnet/nnet.txt

@lamblin can you confirm it is good with you?

For the grad vs the target, to optimize it, it would need a opt that would match the exact pattern I think. Not too hard to do.

For the grad vs the pred, we would need to move that opt during the canonizer phase or reapply the canonizer phase after to have the optimization related to the distributivity. This is probably the one you had problem doing. Or we would need to reapply the full canonizer after this opt. This isn't trial and I'm not sure we want that.

So I think your solution is great!

jeffdonahue:
Thanks for checking this @nouiz! I added documentation to nnet.txt that mirrors the `binary_crossentropy` doc. (Let me know if I should squash to a single commit.)

nouiz:
I'm good with that. We need to wait @lamblin comments.

The more I think about this and other related problem, I would be willing
to explore the idea to add inside the stabilize optimization step more of
the opt from the canonicalize if this help get better optimized graph. I
recall other cases similar when this could give a better graph.

jeffdonahue:
Yes, I agree there are likely many similar cases that would benefit from this kind of stabilization -- e.g., softmax + categorical cross-entropy could be stabilized analogously as it's a generalization of sigmoid + binary cross-entropy (though it's possible that Theano already does the stabilization for that case somehow; didn't look into it).  I'd be happy to see this function deprecated later on if this kind of stabilization can be handled just as well via graph optimization.  (Or if the nicer graph optimization solution could happen soon -- not sure how much effort it would be -- maybe this PR should just be skipped altogether to avoid adding bloat to the API that will quickly get deprecated.)

nouiz:
@lamblin what do you think of this PR?

jeffdonahue:
Thanks for the review and merge @lamblin! (I noticed the docs weren't quite consistent -- unfortunately I don't think I'll get a chance to clean them up anytime soon.)

Issue #5753, opened 2017-03-23
AssertionError, BUG IN FGRAPH.REPLACE OR A LISTENER (theano.gof.fg.CachedConstantError)


fvisin:
When I run my code with `optimizer=fast_compile` I get this `AssertionError` and a bunch of `BUG IN FGRAPH.REPLACE OR A LISTENER` messages. This is really annoying as I am trying to locate a shape error in a huge and messy code :)

Here is a gist with the full log:
https://gist.github.com/fvisin/beff6c95918b18713f03a17f58c8353b

lamblin:
I think we would need the code to reproduce that... I cannot think of a particular reason why this would happen.
Are you using `compute_test_values`? Do you have any scan or OpFromGraph?

fvisin:
Thank you for your answer!

I am not using `compute_test_values`, I definitely have multiple `scan` and I am afraid I have no idea if I have `OpFromGraph`! The code is the one @nouiz was looking into last week. 

I will give you access to the repo as well and send you a parameter file that you can use with the latest code on master.

nouiz:
You have blas.ldflags 2 times. This isn't useful. You can/should remove now
optimizer_excluding=local_softmax_dnn_grad

This is fixed.

nouiz:
Quick reply, I'm able to reproduce it and I think I got a fix.

nouiz:
https://github.com/Theano/Theano/pull/5760

PR #5760, opened 2017-03-24
Fix opt crash fix and  more information when verify_grad detect an error.

nouiz:
fix problem introduced in 80d3b56afe15e9ac802957ad31d3d0
fix gh-5753


fvisin:
I confirm this PR fixes gh-5753.
Thank you!

PR #5763, opened 2017-03-25
Updated numpy as np #4218

Amrithasuresh:
@nouiz 

>In theano/tensor/tests/mlp_test.py:
>In theano/tensor/tests/test_nlinalg.py:

As per suggestion I did that np. to numpy. replacements in docstring.

>Theano/theano/tensor/blas.py

If I change np to numpy in the file (except numpy.distutils. parts). At line :136:5: flake gives error F401 'numpy' imported but unused. What should I do here? The below line shows flake error. So, I didn't the blas.py file. 

try:
    import numpy.distutils.__config__
except ImportError:
    pass

Commited changes in few files for this issue https://github.com/Theano/Theano/issues/4218.

Thank you.

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -1221,7 +1221,7 @@ class MaxAndArgmax(Op):
         # Numpy does not support multiple axes for argmax
         # Work around
         keep_axes = np.array([i for i in range(x.ndim) if i not in axes],
-                                dtype='int64')
+                             dtype='int64')
         # Not-reduced axes in front
         transposed_x = np.transpose(x, np.concatenate((keep_axes, axes)))
         kept_shape = transposed_x.shape[:len(keep_axes)]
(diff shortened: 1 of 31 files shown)


nouiz:
To help prevent rebase of this PR, I'll merge it now. we can check later why tensor/blas.py isn'T working when you do your changes.

nouiz:
For the blas.py file, this diff should remove the error:

diff --git a/theano/tensor/blas.py b/theano/tensor/blas.py
index 8b19798..bb38268 100644
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -133,7 +133,7 @@ import time
 import numpy as np
 import numpy.distutils
 try:
-    import numpy.distutils.__config__
+    import numpy.distutils.__config__  # noqa
 except ImportError:
     pass
 
Can you make a new PR with this diff and the numpy to np changes?

Issue #5580, opened 2017-02-20
A new engine with Intel MKL for Theano


ciyongch:
We are from Intel corporation Ltd., and working on Theano optimization for CPU especially Intel® Xeon® and Intel® Xeon Phi™ processor.

To leverage the performance improvements for all Theano user, we plan to merge our optimization into Theano’s master branch 
and would like to hear your suggestions in advance. 
 
Please check the following details:

# 1 Motivation
Recently, Intel released a new version of MKL® library which provides the high performance primitives to accelerate deep neural network (dnn) on Xeon/Xeon Phi processors.
We have created a new engine under theano/theano/sandbox/mkl as same as cuda for MKL-accelerated Ops. Now, we are doing further verification and performance test for upstreaming in March 2017. 

So we would like to hear from you for this new engine, any comments or suggestion for the PR are highly appreciated.

# 2	Current Status/Perf
We added some new Theano Ops implementation with Intel MKL, including convolution (support bias and group conv), pooling (max pooling and average pooling), relu, normalization(batch norm and local response norm), elemwise sum Ops. 
Also, we added several graph optimizers to automatically replace the original Ops for better performance on Xeon/Xeon Phi processors.
   The portal to the optimized Theano is [here](https://github.com/intel/theano)
    
Currently, we have finished the first version of these Ops. Compared Stock Theano by running [convnet-benchmark](https://github.com/soumith/convnet-benchmarks) on Xeon processor (Dual E5-2699 v4 @ 2.20GHz), the new version shows strongly performance improvement as listed in below table:

| FWD: sec/batch| Stock Theano/Xeon | Optimized Theano/Xeon | speed up |
|---------------|-------------------|-------------------|-----------------------|
| AlexNet       | 1.045             | 0.077             | 13.6                  |
| GoogLeNet     | 2.228             | 0.655             | 3.4                   |
| VGG           | 5.089             | 0.836             | 6.1                   |
| OverFeat      | 6.105             | 0.265             | 23.0                  |

--

| FWD+BWD: sec/batch: sec/batch| Stock Theano/Xeon | Optimized Theano/Xeon | speed up |
|---------------|-------------------|-------------------|-----------------------|
| AlexNet       | 2.333             | 0.236             | 9.9                   |
| GoogLeNet     | 5.866             | 1.535             | 3.8                   |
| VGG           | 12.783            | 2.721             | 4.7                   |
| OverFeat      | 13.202            | 0.830             | 15.9                  |

# 3 Details
[...]

PR #5770, opened 2017-03-27
Add MKL ndarray and MKL library check for the new MKL engine

pcs-theano:
As previously mentioned in issue #5580, we would like to merge our optimization with MKL engine into the master branch. Since there're a lot of code change, so we separate our implementations into several different PRs for the upstream. 

This is our first PR for the optimization, we created a new folder `mkl` under sandbox for all the new codes. Also we added a new tensor type named `MKLNdarrayType` for the data with MKL layout and buffer. The new data type will be used for all the MKL related Ops.

The rest of the code changes will be separated into different PRs based on each different Op and its optimizer in the following days after this one.

Any comments are welcome.

### DIFF
--- a/theano/__init__.py
+++ b/theano/__init__.py
@@ -126,6 +126,12 @@ else:
         raise ImportError("The nose module is not installed."
                           " It is needed for Theano tests.")
 
+if config.device.startswith('cpu'):
+    try:
+        import theano.sandbox.mkl
+    except Exception as e:
+        pass
+
 if config.device.startswith('gpu') or config.init_gpu_device.startswith('gpu'):
     import theano.sandbox.cuda
     # We can't test the driver during import of theano.sandbox.cuda as
--- a/theano/sandbox/mkl/__init__.py
+++ b/theano/sandbox/mkl/__init__.py
@@ -1,0 +1,248 @@
+from __future__ import absolute_import, print_function, division
+import os
+import sys
+import logging
+import textwrap
+import stat
+import theano
+from theano import config, gof
+from six import integer_types
+from theano.gof.cmodule import Compiler, get_lib_extension, GCC_compiler
+from theano.sandbox.mkl.mkl_helper import header_text
+from theano.gof.compilelock import get_lock, release_lock
+
+from theano.gof import EquilibriumDB, SequenceDB
+
+from theano.tensor.blas import ldflags
+
+_logger_name = 'theano.sandbox.mkl'
+_logger = logging.getLogger(_logger_name)
+
+
+mkl_optimizer = EquilibriumDB(ignore_newtrees=False)
+mkl_seqopt = SequenceDB()
+
+
+def register_opt(*tags, **kwargs):
+    if any([not isinstance(t, str) for t in tags]):
+        raise RuntimeError("Bad call to register_opt."
+                           " All tags must be strings.", tags)
+
+    def f(local_opt):
+        name = (kwargs and kwargs.pop('name')) or local_opt.__name__
+        mkl_optimizer.register(name, local_opt, 'fast_run', 'fast_compile',
+                               'mkl', *tags, **kwargs)
+        return local_opt
+    return f
+
+
+mkl_path = os.path.abspath(os.path.split(__file__)[0])
+mkl_ndarray_loc = os.path.join(config.compiledir, 'mkl_ndarray')
+mkl_ndarray_so = os.path.join(
+    mkl_ndarray_loc, 'mkl_ndarray.' + get_lib_extension())
+libmkl_ndarray_so = os.path.join(
+    mkl_ndarray_loc, 'libmkl_ndarray.' + get_lib_extension())
+
+
+def try_import_mkl_ndarray():
+    mkl_files = ('mkl_ndarray.c', 'mkl_ndarray.h')
+
+    stat_times = [os.stat(os.path.join(mkl_path, mkl_file))[stat.ST_MTIME]
+                  for mkl_file in mkl_files]
+    date = max(stat_times)
+    if os.path.exists(mkl_ndarray_so):
+        if date >= os.stat(mkl_ndarray_so)[stat.ST_MTIME]:
+            return False
+
+    try:
+        sys.path[0:0] = [config.compiledir]
+        import mkl_ndarray.mkl_ndarray  # noqa
+        del sys.path[0]
+    except ImportError:
+        return False
+    return True
+
+
+compile_mkl_ndarray = not try_import_mkl_ndarray()
+
+if compile_mkl_ndarray:
+    get_lock()
+    try:
+        if not try_import_mkl_ndarray():
+            try:
+                code = open(os.path.join(mkl_path, 'mkl_ndarray.c')).read()
+                if not os.path.exists(mkl_ndarray_loc):
+                    os.makedirs(mkl_ndarray_loc)
+                compiler = GCC_compiler()
+                preargs = compiler.compile_args()
+                compiler.compile_str('mkl_ndarray',
+                                     code,
+                                     location=mkl_ndarray_loc,
+                                     include_dirs=[mkl_path],
+                                     libs=['mklml_intel' if 'mklml_intel' in config.blas.ldflags else 'mkl_rt'],
+                                     preargs=preargs,)
+
+                from mkl_ndarray.mkl_ndarray import *  # noqa
+            except Exception as e:
+                _logger.error('Failed to compile mkl_ndarray.c: %s', str(e))
+    finally:
+        release_lock()
+
+
+def ok():
+    try:
+        open(libmkl_ndarray_so).close()
+        return True
+    except IOError:
+        return False
+
+
+if not ok():
+    if sys.platform == "win32":
+        shutil.copyfile(mkl_ndarray_so, libmkl_ndarray_so)
+    else:
+        try:
+            os.symlink(mkl_ndarray_so, libmkl_ndarray_so)
+        except OSError as e:
+            if getattr(e, 'errno', None) != errno.EEXIST or not ok():
+                raise
+
+
+class MKLVersion(gof.Op):
+    def c_headers(self):
+        return super(MKLVersion, self).c_headers()
+
+    def c_header_dirs(self):
+        return [config.dnn.include_path]
+
+    def c_libraries(self):
+        return ldflags()
+
+    def c_lib_dirs(self):
+        return ldflags(libs=False, libs_dir=True)
+
+    def make_node(self):
+        return gof.Apply(self, [], [gof.Generic()()])
+
+    def c_support_code(self):
+        ccode = header_text()
+        ccode += """
+        #if PY_MAJOR_VERSION >= 3
+            #define PyInt_FromLong PyLong_FromLong
+        #endif
+        """
+        return ccode
+
+    def c_code(self, node, name, inputs, outputs, sub):
+        o = outputs[0]
+        return textwrap.dedent(
+            """
+            MKLVersion v;
+            mkl_get_version(&v);
+            %(o)s = PyInt_FromLong(atoi(v.Build));
+            """) % locals()
+
+    def c_code_cache_version(self):
+        # Not needed, but make it clear that we do not want to cache this.
+        return None
+
+
+def mkl_version():
+    """
+    Return the current mkl version (e.g., 20160701) we compile with.
+    """
+    if not mkl_available():
+        raise Exception(
+            "We can't determine the mkl version as it is not available",
+            mkl_available.msg)
+
+    if mkl_version.v is None:
+        f = theano.function([], MKLVersion()(),
+                            theano.Mode(optimizer=None),
+                            profile=False)
+        mkl_version.v = f()
+    return mkl_version.v
+
+
+mkl_version.v = None
+
+
+def mkl_available():
+    if config.device != "cpu":
+        mkl_available.avail = False
+        mkl_available.msg = "MKL is disabled since device is not CPU. " \
+                            "Set the device to 'CPU' if you want to use MKL."
+        return mkl_available.avail
+
+    if config.mkl.lib != "mkl":
+        raise NotImplementedError("MKL lib only supports 'mkl', got %s." % config.mkl.lib)
+
+    if config.mkl.nn.enabled == "False":
+        mkl_available.avail = False
+        mkl_available.msg = "MKL is disabled by the 'mkl.nn.enabled' setting."
+        return mkl_available.avail
+    elif mkl_available.avail is not True:
+        preambule = """
+            #include <stdio.h>
+         """
+        preambule += textwrap.dedent(header_text())
+
+        body = textwrap.dedent(
+            """
+            dnnError_t err;
+            dnnLayout_t usr_layout = NULL;
+            size_t size[1] = {256};
+            size_t stride[1] = {1};
+
+            if ((err = dnnLayoutCreate_F32(&usr_layout, 1, size, stride)) != E_SUCCESS) {
+                fprintf(stderr, "Failed to create user layout with mkl: %s", err);
+                return (-1);
+            }
+            """)
+        if 'mklml_intel' in config.blas.ldflags:
+            params = ['-l', 'mklml_intel']
+        else:
+            params = ['-l', 'mkl_rt']
+
+        comp, out, err = Compiler._try_flags(
+            flag_list=params, preambule=preambule, body=body,
+            try_run=False, output=True, compiler=theano.config.cxx, comp_args=False)
+
+        mkl_available.avail = comp
+        if mkl_available.avail is False:
+            mkl_available.msg = (
+                "Can not compile with MKL. We got this error: " +
+                str(err))
+        else:
+            # If we can compile, check that we can import and run.
+            v = theano.function([], MKLVersion()(),
+                                theano.Mode(optimizer=None),
+                                profile=False)()
+            if not isinstance(v, integer_types):
+                mkl_available.avail = False
+                mkl_available.msg = ("Got incorrect mkl version format")
+                raise RuntimeError(mkl_available.msg)
+            if v == -1 or v < 20160802:
+                mkl_available.avail = False
+                mkl_available.msg = "Version(%d) is too old, please use the version of %d or newer one." % (v, int(20160802))
+                raise RuntimeError(mkl_available.msg)
+            else:
+                mkl_available.avail = comp
+
+    if config.mkl.nn.enabled == "True":
+        if not mkl_available.avail:
+            raise RuntimeError(
+                "You enabled MKL, but we aren't able to use it, %s" % mkl_available.msg)
+    return mkl_available.avail
+
+
+mkl_available.avail = None
+mkl_available.msg = None
+
+
+# register name of 'mkl_opt' in opt.py and then add tags for it.
+try:
+    from . import opt
+    opt.optdb.add_tags('mkl_opt', 'fast_compile', 'fast_run')
+except:
+    pass
--- a/theano/sandbox/mkl/mkl_ndarray.c
+++ b/theano/sandbox/mkl/mkl_ndarray.c
@@ -1,0 +1,1251 @@
+#include <Python.h>
+#include <structmember.h>
+#include "theano_mod_helper.h"
+#include <numpy/arrayobject.h>
+#include <stdlib.h>
+#include "mkl_ndarray.h"
+
+
+int MKLNdarray_Check(const PyObject *ob);
+PyObject* MKLNdarray_New(int nd, int typenum);
+int MKLNdarray_CopyFromArray(MKLNdarray *self, PyArrayObject *obj);
+
+
+/*
+ * This function is called in MKLNdarray_dealloc.
+ *
+ * Release all allocated buffer and layout in self.
+ */
+static int
+MKLNdarray_uninit(MKLNdarray *self) {
+    int rval = 0;
+    if (self->dtype == MKL_FLOAT64) {  // for float64
+        if (self->private_data) {
+            rval = dnnReleaseBuffer_F64(self->private_data);
+
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release data: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_data = NULL;
+        }
+
+        if (self->private_layout) {
+            rval = dnnLayoutDelete_F64(self->private_layout);
+
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release layout: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_layout = NULL;
+        }
+
+        if (self->private_workspace) {
+            rval = dnnReleaseBuffer_F64(self->private_workspace);
+
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release workspace: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_workspace = NULL;
+        }
+    } else {  // for float32
+        if (self->private_data) {
+            rval = dnnReleaseBuffer_F32(self->private_data);
+
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release data: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_data = NULL;
+        }
+
+        if (self->private_layout) {
+            rval = dnnLayoutDelete_F32(self->private_layout);
+
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release layout: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_layout = NULL;
+        }
+
+        if (self->private_workspace) {
+            rval = dnnReleaseBuffer_F32(self->private_workspace);
+
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release workspace: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_workspace = NULL;
+        }
+    }
+
+    self->data_size = 0;
+    self->nd = -1;
+    self->dtype = -1;
+    Py_XDECREF(self->base);
+    self->base = NULL;
+    return rval;
+}
+
+
+/*
+ * type: tp_dealloc
+ *
+ * This function will be called by Py_DECREF when object's reference count is reduced to zero.
+ * DON'T call this function directly.
+ *
+ */
+static void
+MKLNdarray_dealloc(MKLNdarray *self) {
+    if (Py_REFCNT(self) > 1) {
+        printf("WARNING: MKLNdarray_dealloc called when there is still active reference to it.\n");
+    }
+    if (self != NULL) {
+        MKLNdarray_uninit(self);
+        Py_TYPE(self)->tp_free((PyObject*)self);
+    }
+}
+
+
+/*
+ * type:tp_new
+ *
+ * This function is used to create an instance of object.
+ * Be first called when do a = MKLNdarray() in python code.
+ *
+ */
+static PyObject*
+MKLNdarray_new(PyTypeObject *type, PyObject *args, PyObject *kwds) {
+    MKLNdarray* self = NULL;
+    self = (MKLNdarray*)(type->tp_alloc(type, 0));
+
+    if (self != NULL) {
+        self->base              = NULL;
+        self->nd                = -1;
+        self->dtype             = -1;
+        self->private_workspace = NULL;
+        self->private_data      = NULL;
+        self->private_layout    = NULL;
+        self->data_size         = 0;
+
+        memset((void*)(self->user_structure), 0, 2 * MAX_NDIM * sizeof (size_t));
+    } else {
+        PyErr_SetString(PyExc_RuntimeError, "MKLNdarray_new: fail to create a new instance \n");
+        return NULL;
+    }
+    return (PyObject*)self;
+}
+
+
+/*
+ * type:tp_init
+ *
+ * This function is called after MKLNdarray_new.
+ *
+ * Initialize an instance. like __init__() in python code.
+ *
+ * args: need input a PyArrayObject here.
+ *
+ */
+static int
+MKLNdarray_init(MKLNdarray *self, PyObject *args, PyObject *kwds) {
+    PyObject* arr = NULL;
+
+    if (!PyArg_ParseTuple(args, "O", &arr))
+        return -1;
+
+    if (!PyArray_Check(arr)) {
+        PyErr_SetString(PyExc_TypeError, "MKLNdarray_init: PyArrayObject arg required");
+        return -1;
+    }
+
+    // do type conversion here. PyArrayObject -> MKLNdarray
+    int rval = -1;
+    if (self != NULL) {
+        rval = MKLNdarray_CopyFromArray(self, (PyArrayObject*)arr);
+        return rval;
+    } else {
+        PyErr_SetString(PyExc_RuntimeError, "MKLNdarray_init: input MKLNdarray* self is NULL");
+        return -1;
+    }
+}
+
+
+/*
+ * type:tp_repr
+ *
+ * Return a string or a unicode object. like repr() in python code.
+ *
+ */
+PyObject* MKLNdarray_repr(PyObject *self) {
+    if (self == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "MKLNdarray_repr: input PyObject* self is NULL");
+        return NULL;
+    }
+    MKLNdarray* object = (MKLNdarray*)self;
+    char cstr[64]; // 64 chars is enough for a string.
+    sprintf(cstr, "ndim=%d, dtype=%s", object->nd, MKL_TYPE[object->dtype]);
+    PyObject* out = PyString_FromFormat("%s%s%s", "MKLNdarray(", cstr, ")");
+#if PY_MAJOR_VERSION >= 3
+    PyObject* out2 = PyObject_Str(out);
+    Py_DECREF(out);
+    return out2;
+#else
+    return out;
+#endif
+}
+
+
+/*
+ * Get dims in user_structure.
+ *
+ * A pointer is returned.
+ *
+ */
+const size_t*
+MKLNdarray_DIMS(const MKLNdarray *self) {
+    if (self != NULL) {
+        return self->user_structure;
+    } else {
+        return NULL;
+    }
+}
+
+
+/*
+ * Get strides in user_structure.
+ *
+ * A pointer is returned. stride has a self->nd offset n user_structure
+ *
+ */
+const size_t*
+MKLNdarray_STRIDES(const MKLNdarray *self) {
+    if (self != NULL) {
+        return self->user_structure + self->nd;
+    } else {
+        return NULL;
+    }
+}
+
+
+/*
+ * Get ndim.
+ *
+ * An integer is returned.
+ *
+ */
+int MKLNdarray_NDIM(const MKLNdarray *self) {
+    if (self != NULL) {
+        return self->nd;
+    } else {
+        return -1;
+    }
+}
+
+
+/*
+ * Get dtype.
+ *
+ * An integer is returned.
+ *
+ */
+int MKLNdarray_TYPE(const MKLNdarray *self) {
+    if (self != NULL) {
+        return self->dtype;
+    } else {
+        return -1;
+    }
+}
+
+
+/*
+ * Get address of private_data.
+ *
+ * An void* pointer is returned.
+ *
+ */
+void*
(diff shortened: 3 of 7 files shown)


pcs-theano:
Currently, there're two kinds of mkl library which supports dnn functionality.
1) mkl library after version `2017.0.0` (build date: `20160801)`, which could be installed with parallel studio 2017 in common way.
2) mklml which is a small version of mkl, and it's free and easy to get and install. You can find the latest mklml from [here](https://github.com/01org/mkl-dnn/releases/download/v0.7/mklml_lnx_2017.0.2.20170209.tgz)

In order to make it more convenient to use, we're planning to submit another PR to add a script for downloading and setting up environment for mklml library in theano, which could be run independently or via setup.py as below:
	`python setup.py install --mkl`
	
Any comments or suggestions?

theano/__init__.py
```diff
@@ -126,6 +126,12 @@ def test():
         raise ImportError("The nose module is not installed."
                           " It is needed for Theano tests.")
 
+if config.device.startswith('cpu'):
+    try:
+        import theano.sandbox.mkl
+    except Exception as e:
```
nouiz: This is a very broad catch and could hide syntax error and other such things. Can you be more strict in what you catch?

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
+from __future__ import absolute_import, print_function, division
+import os
+import errno
+import sys
+import logging
+import textwrap
+import stat
+import shutil
+import theano
```
nouiz: nitpik, add an empty line before the import of Theano to separate the different import group. (we aren't always consistent in Theano, but we try to go in that direction, at one point, we would probably do it in all files)

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+    Return the current mkl version (e.g., 20160701) we compile with.
+    While `0` indicate fail to get mkl version.
+    """
+    if mkl_version.v is None:
+        try:
+            f = theano.function([], MKLVersion()(),
+                                theano.Mode(optimizer=None),
+                                profile=False)
+            mkl_version.v = f()
+        except:
```
nouiz: Why this try except? If there is an error here, is that a sign of a problem we want to raise to the user?

If this need to stay, you should always catch some exception, otherwise, it will also catch interruption like ctrl-C and we don't want that. Be the most precise in which exception you catch (Exception is better then not setting one. But ValueError is better then Exception).

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+    return _dnn_is_able_to_link.status
+
+_dnn_is_able_to_link.status = None
+
+
+def _dnn_is_enabled_by_user():
+    if config.device != "cpu":
+        mkl_available.msg = "MKL is disabled since device is not CPU. " \
+                            "Set the device to 'CPU' if you want to use MKL."
+        return False
```
nouiz: If config.mkl.nn.enabled is False, you should raise an error.

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+            params = ['-l', 'mkl_rt']
+        else:
+            params = []
+
+        compilation_ok, run_ok, out, err = Compiler._try_flags(
+            flag_list=params, preambule=preambule, body=body,
+            try_run=True, output=True, compiler=theano.config.cxx,
+            comp_args=False)
+
+        mkl_available.msg = (
```
nouiz: I think you should set this message only if compilation_ok isn't True or run_ok isn't True

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+                    compiler.compile_str('mkl_ndarray',
+                                         code,
+                                         location=mkl_ndarray_loc,
+                                         include_dirs=[mkl_path],
+                                         libs=[mkl_lib],
+                                         preargs=preargs,)
+
+                    from mkl_ndarray.mkl_ndarray import *  # noqa
+                except Exception as e:
+                    _logger.error('Failed to compile mkl_ndarray.c: %s', str(e))
```
nouiz: All that remind me that we want to refactor/ease compiling other lib with Theano. It probably won't be done shortly, so don't change the PR, but we should do it at some point.

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+                os.symlink(mkl_ndarray_so, libmkl_ndarray_so)
+            except OSError as e:
+                if getattr(e, 'errno', None) != errno.EEXIST or not ok():
+                    raise
+
+
+# register name of 'mkl_opt' in opt.py and then add tags for it.
+try:
+    from . import opt
+    opt.optdb.add_tags('mkl_opt', 'fast_compile', 'fast_run')
```
nouiz: Do that call only if we want to enable mkl part. that should not be done if mkl_available() return False.

I think you can just increate the indentation of this try block by one tab

theano/sandbox/mkl/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+except:
```
nouiz: I think you can remove the try except here comptly. This will put the add_tags inside "if mkl_available()" and at that point, it should not fail.

theano/sandbox/mkl/mkl_ndarray.h
```diff
@@ -0,0 +1,86 @@
...
+
+    /* Type-specific fields go here. */
+    int         nd;                             // the number of dimensions of the tensor, maximum is 16 (MAX_NDIM).
+    int         dtype;                          // an integer type number is given here.
+    size_t      data_size;                      // the number of bytes allocated for mkl_data
+    size_t      user_structure[2 * MAX_NDIM];   // user layout: [size0, size1, ..., stride0, stride1, ..., 0, 0].
+    dnnLayout_t private_layout;                 // layout instance create by MKL APIs
+    void*       private_data;                   // data buffer
+    void*       private_workspace;              // computation workspace for forward and backward
+}MKLNdarray;
```
nouiz: Why not reuse numpy ndarray obejct?

theano/sandbox/mkl/mkl_ndarray.h
```diff
@@ -0,0 +1,86 @@
...
+__attribute__((visibility ("default"))) int MKLNdarray_set_structure(MKLNdarray* self, int nd, const size_t* dims);
+__attribute__((visibility ("default"))) PyObject* MKLNdarray_CreateArrayObj(MKLNdarray* self);
+
+__attribute__((visibility ("default"))) void* MKLNdarray_DATA(const MKLNdarray* self);
+__attribute__((visibility ("default"))) void* MKLNdarray_WORKSPACE(const MKLNdarray* self);
+__attribute__((visibility ("default"))) dnnLayout_t MKLNdarray_LAYOUT(const MKLNdarray* self);
+__attribute__((visibility ("default"))) const size_t* MKLNdarray_DIMS(const MKLNdarray* self);
+__attribute__((visibility ("default"))) const size_t* MKLNdarray_STRIDES(const MKLNdarray* self);
+__attribute__((visibility ("default"))) int MKLNdarray_NDIM(const MKLNdarray* self);
+__attribute__((visibility ("default"))) int MKLNdarray_TYPE(const MKLNdarray* self);
```
nouiz: Use a macro to define the visibility as is done with the MOD_PUBLIC macro. Maybe you can reuse that one.

theano/sandbox/mkl/mkl_type.py
```diff
@@ -0,0 +1,270 @@
...
+    dtype = property(lambda s: s.type.dtype)
+    broadcastable = property(lambda s: s.type.broadcastable)
+    ndim = property(lambda s: s.type.ndim)
+
+
+class MKLNdarrayVariable(_operators, Variable):
+    pass
+
+
+class MKLNdarrayType(Type):
```
nouiz: Before doing a full review of that, why do you want a new C object and not reuse numpy ndarray object or our libgpuarray object. libgpuarray object isn't just for GPU.

If there is a good reason to have a new C object defined in Theano, then I think we need to refactor that, as there is a big quantity of copied code from TensorType. You could make a GenericTensorType() object that both of them inherit from it, and maybe GpuArrayType (but we could do this in another PR).

theano/__init__.py
pcs-theano: Sure, we can handle with the known exceptions here, while leave others as a warning instead of a simple "pass" to be more meaningful.

theano/sandbox/mkl/__init__.py
pcs-theano: No problem, I will change it in this way to catch that point.

theano/sandbox/mkl/__init__.py
pcs-theano: Sure, I will add "Exception" here for the case of failing to compile or link with MKL library.

theano/sandbox/mkl/__init__.py
pcs-theano: OK, I will add the condition here.

theano/sandbox/mkl/__init__.py
pcs-theano: config.mkl.nn.enabled=False allows user not to use MKL even if mkl is available. If raising an error here, there's only one way to go. 
Not flexible?

theano/sandbox/mkl/__init__.py
pcs-theano: Currently, mkl_ndarray is developed and validated only with Theano, so we follow the similar way as cuda. Maybe we can build such package as a third party lib to Theano in future.

theano/sandbox/mkl/__init__.py
pcs-theano: I will remove the try and except and move the other codes within `if mkl_available()`

theano/sandbox/mkl/mkl_ndarray.h
pcs-theano: Will change the code in that way.

theano/sandbox/mkl/mkl_type.py
pcs-theano: Let me introduce some basic background about Intel MKL DNN primitives firstly.
 
Intel MKL DNN primitives work with special opaque data types such as operation primitives (such as convolution forward/backward primitive, etc.,) and layouts of processed data (specifies relative location of elements of processed arrays in memory). Each DNN operation requires input and output arrays have certain data layouts.
 
More detail information can be found [here](https://software.intel.com/en-us/node/684759)
 
 
The reasons why we added such a new C object are: 
 
1.	Numpy ndarray and GPU ndarray doesn’t support storing extra variable such as layout structure.
The memory view described by Numpy ndarray and GPU ndarray is via dims and strides, while it’s described by an opaque data types (layouts) in MKL.
Besides, layouts needs to be passed from one Op to the next one for the computation. We need an C object (TensorType) which allows us to pass the layout.

2.	The data buffer of Numpy ndarray and GPU ndarray can’t be reused by MKL due to different size.
The size of data buffer of Numpy ndarray is exact same as described by dims and strides. While the size of data buffer of mkl op is described by its layout, and it might have different size compared with numpy (might include padding)
 
3.	Same as item 1.
For some Ops, backward pass needs the intermediate result which calculated in forward pass. So we need another data buffer named "workspace" in the variable.
 
Actually, in the beginning, we're thinking about leveraging Numpy ndarray or Cuda/GPU ndarray for this purpose, but none of them allows us to do so. So that's why we create a simplified but sufficient structure for mkl Ops. 

We're glad to see If there's any better suggestions.

theano/sandbox/mkl/__init__.py
nouiz: Sorry, my first comment should have been:

If config.mkl.nn.enabled is True, you should raise an error.

(True instead of False)

theano/sandbox/mkl/__init__.py
nouiz: I'm not sure my comment was clear. I mean, I don't have problems to have that inside Theano. The problem we have is that it is hard to do so in Theano, and we do that for a few things already (the old gpu back-end, cutils, scan cython version, in pylearn2 for cuda convnet (the old alex net cuda code)). My only point is that we should make this easier for Theano to compile lib that aren't Theano op.

theano/sandbox/mkl/mkl_type.py
nouiz: I'm not sure I get everything. By layout, what do you mean? I see 2 interpretations that could be used here that are close to each others and used in Theano.
1) The memory layout. specificed at 100% by shapes and strides. It don't give any semantic to each dimensions, just how to read the data in the right order.
2) The "semantic layout". Is the layout for the inputs images (batch, in channel, image height, image width) or the other semantic layout as tensorflow, cuda convnet and nervana use different layout.

In one, do you mean that your C object have the "semantic layout" inside it?

I checked rapidly the URL you gave. A small comments. I saw many fct that ends with F32 or F64. I suppose it mean float32 or float64. Can you confirm?

theano/sandbox/mkl/__init__.py
pcs-theano: Sure, will raise an error when device is not CPU but mkl.nn.enabled is True, otherwise will return False.

theano/sandbox/mkl/__init__.py
pcs-theano: Got it, sounds great, we can keep the current code as it is, and align to the new design when it's available.

theano/sandbox/mkl/mkl_type.py
pcs-theano: I mean the first one, layout specifies relative location of elements in memory. 
Different MKL DNN operation (for example, convolution, pool, etc.,) might use different layout , and even if the same operation, it could have different layout in different layers or topologies according to their parameters (such as kernel/windows, stride, padding).

You're correct about F32/F64, there're two sets of functions for float32 and float64 currently, and more precision will be supported in future.

theano/sandbox/mkl/mkl_type.py
```diff
@@ -0,0 +1,270 @@
...
+    context_num = 'mkl'
+    Variable = MKLNdarrayVariable
+    Constant = None
+    SharedVariable = None
+    ndim = None
+    broadcastable = None
+
+    def __init__(self, broadcastable, name=None, dtype=None):
```
nouiz: This is the order from our old gpu-back-end. It had a mistake in the order compared to the CPU. Please, use the order 

    def __init__(self, dtype, broadcastable, context_name=None, name=None):

If we end up needing a new Type.

theano/sandbox/mkl/mkl_type.py
```diff
@@ -0,0 +1,270 @@
...
+        self.dtype = dtype
+        self.typenum = 11 if dtype is 'float32' else 12
+        self.broadcastable = tuple(bool(b) for b in broadcastable)
+        self.name = name
+        self.dtype_specs()  # error checking is done there
+
+    def clone(self, dtype=None, broadcastable=None):
+        if broadcastable is None:
+            broadcastable = self.broadcastable
+        return self.__class__(broadcastable, name=self.name, dtype=dtype)
```
nouiz:
```
if dtype is None:
    dtype = self.dtype
```

theano/sandbox/mkl/mkl_type.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def filter(self, data, strict=False, allow_downcast=None):
+        # return self.filter_inplace(data, None, strict=strict, allow_downcast=allow_downcast)
+        return data
```
nouiz: The goal of filter is to cast data into a data represented by this type. Currently it don't do that.

theano/sandbox/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1278 @@
...
+        self->nd                = -1;
+        self->dtype             = -1;
+        self->private_workspace = NULL;
+        self->private_data      = NULL;
+        self->private_layout    = NULL;
+        self->data_size         = 0;
+
+        memset((void*)(self->user_structure), 0, 2 * MAX_NDIM * sizeof (size_t));
+    } else {
+        PyErr_SetString(PyExc_RuntimeError, "MKLNdarray_new: fail to create a new instance \n");
```
nouiz: Should this be an PyExc_MemoryError?

theano/sandbox/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1278 @@
...
+    // prepare user layout and mkl buffer
+    err = MKLNdarray_create_buffer_from_structure(self);
+    if (err < 0) {
+        return err;
+    }
+
+    // copy data to mkl buffer
+    size_t element_size = (size_t)PyArray_ITEMSIZE(obj);
+    // assert (user_size * element_size <= self->data_size);
+    memcpy((void*)self->private_data, (void*)PyArray_DATA(obj), user_size * element_size);
```
nouiz: What if the ndarray is strided? Currently it don't raise an error nor copy correctly in that case.

theano/sandbox/mkl/mkl_type.py
nouiz: I understand your last explanation. But I don't understand why this is a reason for a new C object. shapes and strides in the data object are enough to specify the memory layout of your point 1) 2 comments above.

pcs-theano:
@nouiz 

The MKL layout concept is essential just memory format description, it not only contains shapes, strides but some other stuff which related with hardware specifications. So we can't leverage the current Numpy ndarrays's structure to represent it.

All MKL DNN primitives use their own layouts, all these layouts are optimized based on the tensors' logical size and what the current hardware is (for example, Xeon, Xeon Phi have different layout for the same Numpy ndarray), in order to show good efficiency for DNN Ops. It's much complicated than NCHW or NHWC, etc., which are used by current DL FWs.

And the buffer allocated based on MKL layout includes rearranged data, and might includes some extra information. So the memory size might be larger than Numpy ndarray's data buffer.

Since the layout structure needs to be passed from one Op to the next, and a bigger buffer is required to store the optimal rearrange data, that's why we need a special C object to support it. 

The good thing is all this stuff are no matter for end users, since the MKL Ops handles all of them. 

Currently, there's no any page for MKL layout. Hope the above description could help you better understand it.

nouiz:
I think I got part of why a new object is needed. If I combine the information from a few places, as you sometimes use a 6d inputs tensor where some of the normal dimensions are split into 2 non concecutive dimensions. In that case, you need to track not just how to read the data from the RAM, but also the meaning of each dimensions. And that request something. Using a new C object for that make sence.

@lamblin suggested to create a new directory called "contrib" and put your code there instead of sandbox, what do you think of that?

As we understand the why of the new object, you can do the first batch of comments I did. After that I'll have another look at it.

thanks

pcs-theano:
@nouiz 

Thanks for your time to review and discuss about the new C object, glad to see that it's accepted, it will make the other MKL Ops easy since they're all built based on this new object.

We'll update the PR according to your previous comments, and move all our codes from sandbox to "contrib".

BTW, regarding the script which supports automatically downloading the small MKL library as I mentioned in the very beginning of this issue page, any comments for that?

nouiz:
I like the idea to help user install mkl. So if you have a script or can add that to setup.py, it would be great. I don't have a preference. Using setup.py could make this easier to install on windows then scripts.

theano/contrib/mkl/mkl_ndarray.h
```diff
@@ -0,0 +1,91 @@
...
+
+#ifndef Py_TYPE
+#define Py_TYPE(o) ((o)->ob_type)
+#endif
+
+#define MAX_NDIM     (16)
+#define MKL_FLOAT32  (11)
+#define MKL_FLOAT64  (12)
+
+char* MKL_TYPE[] = {"", "", "", "int16", "", "int32", "", "int64",
```
nouiz: does some ints are supported?

theano/contrib/mkl/mkl_type.py
```diff
@@ -0,0 +1,350 @@
...
+        if self.dtype not in ('float32', 'float64'):
+            raise TypeError('%s only supports float32/float64 for now.'
+                            'Tried using dtype %s for variable %s' %
+                            (self.__class__.__name__, dtype, name))
+
+        self.typenum = 11 if dtype is 'float32' else 12
+        self.broadcastable = tuple(bool(b) for b in broadcastable)
+        self.name = name
+        self.dtype_specs()
+        if not (context_name is None):
```
nouiz: Why a context_name? Is it always "mkl"?

Is this for Xeon Phi?

theano/contrib/mkl/mkl_type.py
```diff
@@ -0,0 +1,350 @@
...
+            self.context_name = context_name
+
+    def clone(self, dtype=None, broadcastable=None):
```
nouiz: If we need to keep the context_name, I would add it here too for completness. But if you know it isn't useful, don't add it.

theano/contrib/mkl/mkl_type.py
```diff
@@ -0,0 +1,350 @@
...
+                                ' dimension.', shp, self.broadcastable)
+
+        return data
+
+    def filter_variable(self, other, allow_convert=True):
+        raise NotImplementedError('MKLNdarrayType filter_variable')
+
+    @staticmethod
+    def values_eq(a, b):
+        raise NotImplementedError('MKLNdarray values_eq')
```
nouiz: This probably need to be implemented. values_eq_approx() must be implemented to be able to work in DebugMode. At worst, convert both object to ndarray and call TensorType.values_eq_approx() on the converted version.

theano/contrib/mkl/mkl_type.py
```diff
@@ -0,0 +1,350 @@
...
+                self.context_name == other.context_name)
+
+    def __hash__(self):
+        """
+        Hash equal for same kinds of MKLNdarrayType.
+
+        """
+        return hash((type(self), self.dtype, self.broadcastable, self.context_name))
+
+    ndim = property(lambda self: len(self.broadcastable), doc='number of dimensions')
```
nouiz: Move that at the top of this class to replace your ndim = None there.

theano/contrib/mkl/mkl_type.py
```diff
@@ -0,0 +1,350 @@
...
+        %(oname)s = (PyArrayObject*)PyArray_SimpleNew(1, shape, NPY_INT64);
+    }
+
+    for (int i=0; i<shape[0]; i++) {
+        ((npy_int64*)PyArray_GETPTR1(%(oname)s, i))[0] = MKLNdarray_DIMS(%(iname)s)[i];
+    }
+    """,
+    version=(1,))
+
+theano.compile.register_shape_i_c_code(
```
nouiz: ideally, theano.compile.register_deep_copy_op_c_code() and 
theano.compile.register_specify_shape_c_code() should have c code as in tensor/type.py, but this isn't mandatory.

pcs-theano:
@nouiz 
Sorry for the late response on this issue, we've updated the codes according to your comments, please help to have a check on the new changes.

theano/contrib/mkl/__init__.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def register_opt(*tags, **kwargs):
+    if any([not isinstance(t, str) for t in tags]):
+        raise RuntimeError("Bad call to register_opt."
+                           " All tags must be strings.", tags)
+
+    def f(local_opt):
+        name = (kwargs and kwargs.pop('name')) or local_opt.__name__
+        mkl_optimizer.register(name, local_opt, 'fast_run', 'fast_compile',
```
pcs-theano: Sure, will keep it for now and align to the new change in future.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+    int status = -1;
+    dnnPrimitive_t primitive = NULL;
+
+    size_t mkl_size[MAX_NDIM] = {0};
+    size_t mkl_stride[MAX_NDIM] = {0};
+
+    // nchw -> whcn
+    for (int i = 0; i < self->nd; i++) {
+        mkl_size[i] = (MKLNdarray_DIMS(self))[self->nd - i - 1];
+        mkl_stride[i] = (MKLNdarray_STRIDES(self))[self->nd - i -1];
```
pcs-theano: MKL object has its own description for the layout, and it's enough to MKL object.
The reason to keep the original information (strides and dims) is that, when we convert MKL object to Numpy Ndarray we need to know the layout information of target object (Numpy Ndarray).

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+ * ml_flags: indicate how the call should be constructed
+ * ml_doc: docstring
+ *
+ */
+static PyMethodDef MKLNdarray_methods[] = {
+    {"__array__",
+        (PyCFunction)MKLNdarray_CreateArrayObj, METH_VARARGS,
+        "Copy from MKL to a numpy ndarray."},
+
+    {"zeros",
```
pcs-theano: Both of them were added in the new codes.

theano/contrib/mkl/mkl_type.py
pcs-theano: Both of them were added in the new codes.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+            if (rval != E_SUCCESS) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_uninit: fail to release layout: %d, line: %d",
+                             rval, __LINE__);
+            }
+            self->private_layout = NULL;
+        }
+
+        if (self->private_workspace) {
+            rval = dnnReleaseBuffer_F32(self->private_workspace);
```
pcs-theano: Regarding the dnn function names, it was a legacy design in MKL that including dtype in it. 

So with such fixed APIs, we can't avoid duplicated codes, no matter we write them in this way or create a new header file which combines the two functions into a wrapper by using a parameter "dtype" to distinguish different data types.

So I suggest to keep the code as it was for now, and align to MKL when it evolves.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+    if (self == NULL || obj == NULL) {
+        PyErr_SetString(PyExc_RuntimeError,
+                        "MKLNdarray_CopyFromArray: input self or obj is NULL");
+        return -1;
+    }
+    int ndim = PyArray_NDIM(obj);
+    npy_intp* d = PyArray_DIMS(obj);
+    int typenum = PyArray_TYPE(obj);
+
+    if (typenum != MKL_FLOAT32 && typenum != MKL_FLOAT64) {
```
pcs-theano: Code was updated according to your comments.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+    size_t user_size = 1;
+
+    for (int i = 0; i < ndim; i++) {
+        dims[i] = (size_t)d[i];
+        user_size *= dims[i];
+    }
+
+    int err = MKLNdarray_set_structure(self, ndim, dims);
+    if (err < 0) {
+        return err;
```
pcs-theano: Code was updated according to your comments.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+    }
+
+    // prepare user layout and mkl buffer
+    err = MKLNdarray_create_buffer_from_structure(self);
+    if (err < 0) {
+        return err;
```
pcs-theano: Code was updated according to your comments.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+        return NULL;
+    }
+
+    if (MKLNdarray_create_buffer_from_structure(rval)) {
+        PyErr_SetString(PyExc_RuntimeError, "MKLNdarrya_create_with_zeros: create buffer from structure failed.");
+        Py_DECREF(rval);
+        return NULL;
+    }
+    // Fill with zeros
+    memset(rval->private_data, 0, total_size);
```
pcs-theano: Yes, it's a potential issue, already fixed in the new codes.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+        PyErr_SetString(PyExc_RuntimeError,
+                        "MKLNdarray_get_size: input MKLNdarray* self is NULL");
+        return NULL;
+    }
+
+    if (self->nd <= 0) {
+        total_element = 0;
+    } else {
+        for (int i = 0; i < self->nd; i++) {
+            total_element *= self->user_structure[i];
```
pcs-theano: Code was updated according to your comments.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+        }
+
+        if (!dnnLayoutCompare_F64(self->private_layout, layout_user)) {
+            status = dnnConversionCreate_F64(&primitive, self->private_layout, layout_user);
+            if (E_SUCCESS != status || NULL == primitive) {
+                PyErr_Format(PyExc_RuntimeError,
+                             "MKLNdarray_CreateArrayObj: dnnConversionCreate_F64 failed: %d, line: %d",
+                             status, __LINE__);
+                Py_DECREF(rval);
+                return NULL;
```
pcs-theano: Code was updated according to your comments.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+    for (int i = 0; i < self->nd; i++) {
+        mkl_size[i] = (MKLNdarray_DIMS(self))[self->nd - i - 1];
+        mkl_stride[i] = (MKLNdarray_STRIDES(self))[self->nd - i -1];
+    }
+
+    if (self->dtype == MKL_FLOAT64) { // float64
+        status = dnnLayoutCreate_F64(&layout_user,
+                                     self->nd,
+                                     mkl_size,
+                                     mkl_stride);
```
pcs-theano: Yes, you're right about the conversion. But we can't create the conversion primitive for two same layout, so we can't call the convert function (which requires conversion primitive as one of its parameters) to do the simple copy work.

No matter we convert private layout to user layout or vice verse, private layout is always an internal description for mkl object, but user layout is built from Numpy Ndarray actually.

For example, wehen convert private layout to user layout, so the destination layout is layout_user. Although the information we used to build the layout_user is `mkl_size` and `mkl_stride` (and they're stored in user_structure of mkl object), they're acquired from Numpy Ndarray (`dims` and `strides` respectively), the only difference between `mkl_size` and `dims` is element order, `mkl_size` use `WHCN` order, while `dims` use `NCHW` order, same for strides.

So mkl_strides should be ok in this case.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+PyObject* MKLNdarray_Zeros(PyObject *_unused, PyObject *args) {
+    if (!args) {
+        PyErr_SetString(PyExc_TypeError, "MKLNdarray_Zeros: function takes at least 1 argument");
+        return NULL;
+    }
+
+    PyObject* shape = NULL;
+    int typenum = -1;
+
+    if (!PyArg_ParseTuple(args, "O|i", &shape, &typenum)) {
```
pcs-theano: Code was updated according to your comments.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -0,0 +1,1361 @@
...
+        return NULL;
+    }
+    self->base              = NULL;
+    self->nd                = nd;
+    self->dtype             = typenum;
+    self->private_data      = NULL;
+    self->private_workspace = NULL;
+    self->data_size         = 0;
+    self->private_layout    = NULL;
+    memset((void*)(self->user_structure), 0, 2 * MAX_NDIM * sizeof (size_t));
```
pcs-theano: Since we need strides information to build layout_user, so we need to keep this.

theano/contrib/mkl/mkl_ndarray.h
pcs-theano: Not yet, but they should be supported in future.

theano/contrib/mkl/mkl_type.py
pcs-theano: No difference for Xeon or Xeon Phi, so we remove it.

theano/contrib/mkl/mkl_type.py
pcs-theano: Ok, we just remove it.

theano/contrib/mkl/mkl_type.py
pcs-theano: Both functions were added in the new code.

theano/contrib/mkl/mkl_type.py
pcs-theano: Done.

pcs-theano:
Hi @nouiz, any updates for this PR?
Currently, we're preparing other PRs such as convolution/pooling/relu, etc., but all these Ops will depend on this one since they're built based on mkl ndarray data type. If the current design of mkl ndarray is ok, then we can move forward with other Ops smoothly, or we need to make relevant changes according to new comments on mkl ndarray.

theano/contrib/mkl/mkl_ndarray.h
```diff
@@ -54,16 +59,18 @@ char* MKL_TYPE[] = {"", "", "", "int16", "", "int32", "", "int64",
...
-    void*       private_data;                   // data buffer
-    void*       private_workspace;              // computation workspace for forward and backward
+    int         nd;                                 // the number of dimensions of the tensor, maximum is 16 (MNDA_MAX_NDIM).
+    int         dtype;                              // an integer type number is given here.
+    size_t      data_size;                          // the number of bytes allocated for private_data
+    size_t      workspace_size;                     // the number of bytes allocated for private_workspace
+    size_t      user_structure[2 * MNDA_MAX_NDIM];  // user layout: [size0, size1, ..., stride0, stride1, ..., 0, 0].
+    dnnLayout_t private_layout;                     // layout instance create by MKL APIs
+    void*       private_data;                       // data buffer
+    dnnLayout_t private_layout_ws;                  // layout for workspace
```
nouiz: Can you confirmt this extra layout is needed as sometimes the data layout and the workspace layout are different?

Currently, theano handle different workspace, but the way you do it is hidded from Theano. So it is fine. Is it possible that different node in the graph share the same workspace to lower the memory usage? I don't think it is mandatory, but if it is possible/already done, it would be great.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -6,95 +6,129 @@
...
+        self->data_size = 0;
+        self->workspace_size = 0;
+        self->nd = -1;
+        self->dtype = -1;
+        Py_DECREF(self->base);
+        self->base = NULL;
+        rval = 0;
+
+    } else {
+        if (MNDA_FLOAT64 == self->dtype) {  // for float64
```
pcs-theano: MNDA stands for MKL NDArray to avoid naming conflict with MKL library.

theano/contrib/mkl/mkl_ndarray.c
```diff
@@ -1139,7 +1540,23 @@ static PyMethodDef MKLNdarray_methods[] = {
 
     {"zeros",
         (PyCFunction)MKLNdarray_Zeros, METH_STATIC | METH_VARARGS,
-        "Create a new MklNdarray with specified shape, filled with zeros."},
+        "Create a new MklNdarray with specified shape, filled ith zeros."},
+
+    {"__copy__",
+        (PyCFunction)MKLNdarray_View, METH_NOARGS,
```
pcs-theano: Sure, will align to latest codes.

theano/contrib/mkl/mkl_ndarray.h
pcs-theano: `private_layout_ws` is different than `private_layout`, workspace layout/buffer is needed to store the intermediate result by some MKL Ops such as pool/lrn. For MKL Op, only backward and forward node share the same workspace (for example, pool and poolGrad), not for different node.
So it's necessary to keep them.

nouiz:
Ok. Put that in documentation/comments somewhere.

But for the dtype, why not reuse MKL dtype? You don't need to change it, I
was curious why you redefined them instead of using there definition in
some MKL headers.

pcs-theano:
The reason we changed to `MNDA_FLOATx` is: `MNDA_FLOATx` stands for type 
number (11, 12) to align with Numpy, which has different meaning compared to `MKL_INTx`
used by MKL library. For example `MKL_INT8` and `MKL_INT64` stands for
`int` or `long long int` types respectively, besides, there's no `MKL_FLOATx` in MKL.
Since MKLNdarray is not a part of MKL library, so we changed it to avoid namespace conflict
with MKL.
> But for the dtype, why not reuse MKL dtype? You don't need to change it, I
was curious why you redefined them instead of using there definition in
some MKL headers.

theano/contrib/mkl/mkl_ndarray.c
pcs-theano: Before submit the new code, did you mean both `__copy__` and `__deepcopy__` should do the same deepcopy, and keep the current `copy` and `view` as it is?

nouiz: Change MKLNdarray_View to MKLNdarray_DeepCopy here.

Issue #5765, opened 2017-03-25
ERROR (theano.gof.opt): Optimization failure due to: constant_folding (python 3.6 issue ?)


jc-Rosichini:
Hi I just installed a new windows 10 configuration with exactly the same theano/lasagne code than runs properly on another similar machine. Both using the latest versions of theano/lasagne and gpuarray.

On the new one at compile time, I'm getting a bunch or errors:
ERROR (theano.gof.opt): Optimization failure due to: constant_folding
....
TypeError: ("a bytes-like object is required, not 'str'", '[GpuFromHost(TensorConstant{(1,) of 1e-06})]')

(see below and attached dump)

which doesn't occur on the other configuration.

As far as I can see the only difference is that the new one uses Anaconda Python 3.6 versus Anaconda Python 3.5.

Here is part of the dump:

Python 3.6.0 |Anaconda 4.3.1 (64-bit)| (default, Dec 23 2016, 11:57:41) [MSC v.1900 64 bit (AMD64)] on win32
Using cuDNN version 5105 on context None
Mapped name None to device cuda0: GeForce GTX 1080 (0000:01:00.0)
theano:          0.9.0.dev-b3d19ce8e8d7cdf87c3f84599be81647803aa1f6
numpy:           1.11.3
pygpu:           0.6.2
lasagne:         0.2.dev1

> 
> Compilation starts at 2017-03-25 19:57:39
> optimizer: rmsprop, initial learning rate:  0.001
> ERROR (theano.gof.opt): Optimization failure due to: constant_folding
> ERROR (theano.gof.opt): node: GpuFromHost<None>(TensorConstant{(1,) of 1e-06})
> ERROR (theano.gof.opt): TRACEBACK:
> ERROR (theano.gof.opt): Traceback (most recent call last):
>   File "C:\Users\MachineLearning\AppData\Local\Continuum\Anaconda3\lib\site-packages\theano\gof\opt.py", line 1996, in process_node
>     replacements = lopt.transform(node)
>   File "C:\Users\MachineLearning\AppData\Local\Continuum\Anaconda3\lib\site-packages\theano\tensor\opt.py", line 6497, in constant_folding
>     no_recycling=[], impl=impl)
>   File "C:\Users\MachineLearning\AppData\Local\Continuum\Anaconda3\lib\site-packages\theano\gof\op.py", line 935, in make_thunk
>     no_recycling)
[...]

PR #5773, opened 2017-03-27
Fix error during error in Python 3

nouiz:
Found in https://github.com/Theano/Theano/issues/5765

it don't fix this PR, as this only hide the original error.


lamblin:
Wouldn't #5769 be a better fix?

abergeron:
If that other PR fixes the problem, it's a much better solution.

nouiz:
I merged the other PR.

Issue #5764, opened 2017-03-25
float64 warning


apulver:
Warns about creating a float64 when lr1 is used as the learning rate, even though lr1's value was given as a float32. 
```
import theano
import theano.tensor as T
import numpy as np
from collections import OrderedDict


def test():
    X = T.matrix()
    x = np.random.randn(10,32).astype(np.float32)
    m1 = theano.shared(np.random.randn(32,8).astype(np.float32))
    m2 = theano.shared(np.random.randn(8,32).astype(np.float32))

    loss = (X - T.dot(T.dot(X,m1),m2)).norm(L=2)
    params = [m1,m2]
    updates = OrderedDict()
    lr0 = .001
    lr1 = theano.shared(np.float32(.001))
    print(lr1.get_value().dtype)
    grads = T.grad(loss, params)
    for g,p in zip(grads, params):
        updates[p] = p - lr1*g
    train = theano.function(inputs=[X], outputs=loss, updates=updates)
    for i in range(100):
        print(train(x))

test()
```

nouiz:
I'm able to reproduce it. It is a harmless float64. Sadly this happen from time to time on the new back-end, as it now support float64 and sometimes, scalar end up in float64 for some reasons.

You can ignore it (so use warn_float64=warn instead of raise for now).

I'll try to fix that. thanks for the report.

nouiz:
@lamblin @abergeron what do you think of the option at the end?
While investigating this I found that we moved too much elemwise of scalar to the GPU. This is fixed by this PR:

https://github.com/Theano/Theano/pull/5774

That do not fix the original problem. With this PR, the problem is that the elemwise fusion opt will fuse a cast to float64 introduced by GpuGemm make_node.

It is getting hard to make all those extra float64 not trigger warn_float64. I see 3 options:

- Continue as now and patch manually each cases.
- Disable warn_float64 during compilation by default by moving that to a new flag compile_warn_float64
- Stop using float64 for some ops in the graph when not needed.

I would go with option 2.

abergeron:
I would go for option 2 or option 4 which is: remove the warn_float64 flag completely since the problem it tries to address is no longer present in the new backend.

nouiz:
I got discussion with @abergeron live and we found that we didn't need the float64 in the graph. It was there to make the code faster to implement. So we dicided to go with option 3.

nouiz:
This PR should fix this: https://github.com/Theano/Theano/pull/5774

PR #5774, opened 2017-03-27
Don't move scalar float* elemwise unless the result is needed on the GPU.

nouiz:
was found while investigating:

https://github.com/Theano/Theano/issues/5764

- also stop introducing useless float64 cast in the graph. fix gh-5764

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -50,16 +50,15 @@ class GpuGemv(BlasOp):
         A = as_gpuarray_variable(A, ctx_name)
         x = as_gpuarray_variable(x, ctx_name)
         y = as_gpuarray_variable(y, ctx_name)
-        with theano.configparser.change_flags(warn_float64='ignore'):
-            alpha = as_tensor_variable(alpha).astype('float64')
-            beta = as_tensor_variable(beta).astype('float64')
+        alpha = as_tensor_variable(alpha)
+        beta = as_tensor_variable(beta)
 
         assert alpha.ndim == 0
         assert beta.ndim == 0
         assert A.ndim == 2
         assert x.ndim == 1
         assert y.ndim == 1
-        assert A.dtype == x.dtype == y.dtype
+        assert A.dtype == x.dtype == y.dtype == alpha.dtype == beta.dtype
         return Apply(self, [y, alpha, A, x, beta], [y.type()])
 
     def perform(self, node, inputs, out_storage):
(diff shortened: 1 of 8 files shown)


abergeron:
I would prefer a unified policy regarding operations on scalars rather than patching op-by-op in the lifters.

There is already a central point (op_lifter) where decisions about whether to lift or not are made and I would like to keep it that way.  We can declare that float scalars should stay on the CPU unless requested on the GPU.

nouiz:
I did that and other changes we talked live. But I still have problems with float16.

theano/gpuarray/blas.py
```diff
@@ -244,8 +256,11 @@ def make_node(self, A, alpha, x, y):
         A = as_gpuarray_variable(A, ctx_name)
         x = as_gpuarray_variable(x, ctx_name)
         y = as_gpuarray_variable(y, ctx_name)
-        with theano.configparser.change_flags(warn_float64='ignore'):
-            alpha = as_tensor_variable(alpha).astype('float64')
+        alpha = as_tensor_variable(alpha)
+        if len(set([A.dtype, alpha.dtype, x.dtype, y.dtype])) != 1:
```
abergeron: plz no

theano/gpuarray/type.py
```diff
@@ -24,22 +24,36 @@
 _context_reg = {}
 
 
+def do_gpu_support(data):
```
abergeron: I would name this gpu_supported, since do_gpu_support is an action, not a question.

theano/gpuarray/tests/test_opt.py
```diff
@@ -493,6 +493,27 @@ def test_many_arg_elemwise():
                 utt.assert_allclose(results_gpu, results_cpu)
 
 
+def test_not_useless_scalar_gpuelemwise():
+    # We don't want to move elemwise on scalar on the GPU when the
+    # result will be used on the GPU!
```
abergeron: will be used -> will not be used

theano/gpuarray/tests/test_blas.py
```diff
@@ -15,16 +15,24 @@
...
+                    gpugemm_inplace, gpugemm_no_inplace,
+                    gpugemmbatch_no_inplace,
                     gpuger_inplace, gpuger_no_inplace,
                     GpuGer, gpu_dot22)
 
 
 GpuGemvTester = makeTester(
     'GpuGemvTester',
     op=gemv_inplace, gpu_op=gpugemv_inplace,
+    # It don't support float16
```
abergeron: It don't -> It doesn't

theano/gpuarray/blas.py
nouiz: why? This seem much clear to my point of view? This is copied code from tensor/blas.py anyway...

theano/gpuarray/blas.py
```diff
@@ -257,15 +257,14 @@ def make_node(self, A, alpha, x, y):
         x = as_gpuarray_variable(x, ctx_name)
         y = as_gpuarray_variable(y, ctx_name)
         alpha = as_tensor_variable(alpha)
-        if len(set([A.dtype, alpha.dtype, x.dtype, y.dtype])) != 1:
+        if not(A.dtype == x.dtype == y.dtype == alpha.dtype)
```
abergeron: Missing `:` at the end.

nouiz: I amended the last commit.

theano/gpuarray/opt.py
```diff
@@ -1175,6 +1175,8 @@ def local_gpua_gemv(op, context_name, inputs, outputs):
 @op_lifter([tensor.blas.Gemm])
 @register_opt2([tensor.blas.Gemm], 'fast_compile')
 def local_gpua_gemm(op, context_name, inputs, outputs):
+    if inputs[0].dtype not in ['float32', 'float64']:
+        return
```
abergeron: Why do you disable float16 gemm?

PR #5775, opened 2017-03-27
Correct gradient of scan until.

### LINKED ISSUE #5439: Is there a bug when computing gradient of scan with until?
Hi, 

I want to compute the gradient from the scan's output w.r.t the "sequences" input.
I used a "scan_module.until" to early-stop the iteration.
However, when I ran the following lines, the results seem weird:

```
import theano
import theano.tensor as tensor
import theano.scan_module as ts
import numpy

print theano.__version__

x1 = tensor.vector(dtype='float32')
r,_= theano.scan(lambda i, x: (x[i]*x[i], ts.until(tensor.eq(x[i], 5))),
                sequences=tensor.arange(x1.shape[0]),
                non_sequences=[x1])

g = theano.grad(r.sum(), x1)
f = theano.function([x1], [r, g])

x0 = numpy.arange(20).astype('float32')

print x0
print f(x0)
```

The results are:
```
[  0.   1.   2.   3.   4.   5.   6.   7.   8.   9.  10.  11.  12.  13.  14.  15.  16.  17.  18.  19.]
[array([  0.,   1.,   4.,   9.,  16.,  25.], dtype=float32),
 array([  0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0., 0.,   0.,   0.,  28.,  30.,  32.,  34.,  36.,  38.], dtype=float32)]
```

In my view, since the iterations only ran 5 steps before it stops, the gradient should only be computed for the first 5 elements. However,  it did the last 5 elements of the sequence.

Thank you very much!

Thrandis:
Fixes #5439

Still work in progress. I made a few changes. There is one test failing: `test_grad_multiple_outs_some_truncate`.
I will investigate it.

### DIFF
--- a/theano/scan_module/scan_op.py
+++ b/theano/scan_module/scan_op.py
@@ -2198,7 +2198,7 @@ class Scan(PureOp):
                 dC_dinps_t[dx + self.n_seqs] += dC_dXtm1
         # Construct scan op
         # Seqs
-        outer_inp_seqs = [x[::-1] for x in inputs[1:1 + self.n_seqs]]
+        outer_inp_seqs = [x[:grad_steps][::-1] for x in inputs[1:1 + self.n_seqs]]
         for idx in xrange(self.n_mit_mot + self.n_mit_sot):
             mintap = np.min(self.tap_array[idx])
             if idx < self.n_mit_mot:
@@ -2216,7 +2216,7 @@ class Scan(PureOp):
             x[:-1][::-1] for x in self.outer_sitsot_outs(outs)]
         for x in self.outer_nitsot_outs(dC_douts):
             if not isinstance(x.type, DisconnectedType):
-                outer_inp_seqs.append(x[::-1])
+                outer_inp_seqs.append(x[:grad_steps][::-1])
 
         if hasattr(inputs[0].tag, 'test_value'):
             # Here we tests that the new scan input sequence all have
@@ -2229,15 +2229,15 @@ class Scan(PureOp):
                 mintap = np.min(taps)
                 if hasattr(x[::-1][:mintap], 'test_value'):
                     assert (x[::-1][:mintap].tag.test_value.shape[0] ==
-                            inputs[0].tag.test_value)
+                            grad_steps.tag.test_value)
             for x in self.outer_sitsot_outs(outs):
                 if hasattr(x[::-1][:-1].tag, 'test_value'):
                     assert (x[::-1][:-1].tag.test_value.shape[0] ==
-                            inputs[0].tag.test_value)
+                            grad_steps.tag.test_value)
             for x in self.outer_nitsot_outs(outs):
                 if hasattr(x[::-1].tag, 'test_value'):
-                    assert (x[::-1].tag.test_value.shape[0] ==
-                            inputs[0].tag.test_value)
+                    assert (x[:grad_steps][::-1].tag.test_value.shape[0] ==
+                            grad_steps.tag.test_value)
         outer_inp_seqs += [x[::-1][:np.min(taps)]
                            for taps, x in zip(self.mitsot_taps(),
                                               self.outer_mitsot_outs(outs))]
@@ -2536,7 +2536,7 @@ class Scan(PureOp):
                         outer_inp_seqs +
                         outer_inp_mitmot +
                         outer_inp_sitsot +
-                        [inputs[0] for _ in xrange(n_nit_sot)] +
+                        [grad_steps for _ in xrange(n_nit_sot)] +
                         self.outer_shared(inputs) +
                         self.outer_non_seqs(inputs))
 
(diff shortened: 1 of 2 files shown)


theano/scan_module/scan_op.py
```diff
@@ -2198,7 +2198,7 @@ def compute_all_gradients(known_grads):
                 dC_dinps_t[dx + self.n_seqs] += dC_dXtm1
         # Construct scan op
         # Seqs
-        outer_inp_seqs = [x[::-1] for x in inputs[1:1 + self.n_seqs]]
+        outer_inp_seqs = [x[:grad_steps][::-1] for x in inputs[1:1 + self.n_seqs]]
```
nouiz: Instead of doing 2 subtensor and asking Theano optimization to do the clean up, can you combine both? I think could do it:

    x[grad_steps-1::-1]
```

>>> range(10)[4-1::-1]
[3, 2, 1, 0]
>>> range(10)[:4][::-1]
[3, 2, 1, 0]

```

lamblin: It would not work if `grad_steps <= 0`, though. Are we sure it is never going to be 0?
```python
>>> range(10)[0-1::-1]
[9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
>>> range(10)[:0][::-1]
[]
```

Thrandis:
```python 
i = 0
cond = True 
# ############# THE MAIN LOOP ##############
while (i < n_steps) and cond:
    # Scan loop
```
And `n_steps` can't be `0` (It is checked before in the code). So it seems that we are good with what @nouiz proposes.

Thrandis:
Now I will check the behavior when both `as_while` and `truncate_gradient`.

Thrandis:
I am indeed breaking the gradient when using `truncate_gradient`.

nouiz:
this need a rebase

Thrandis:
I did the rebase and fixed what I broke in `truncate_gradient`.

Now scan works exactly as before when not in an `as_while` mode, and I cut the input sequences and pad the output sequences when in an `as_while` mode. `as_while` also works with `truncate_gradients`.

I should maybe add some more tests to check if everything behaves properly (for example with multiple sequences, taps,...)

theano/scan_module/scan_op.py
```diff
@@ -2198,7 +2200,10 @@ def compute_all_gradients(known_grads):
                 dC_dinps_t[dx + self.n_seqs] += dC_dXtm1
         # Construct scan op
         # Seqs
-        outer_inp_seqs = [x[::-1] for x in inputs[1:1 + self.n_seqs]]
+        if self.as_while:
+            outer_inp_seqs = [x[n_steps - 1::-1] for x in inputs[1:1 + self.n_seqs]]
```
nouiz: Add a comment to what is the equivalent, more readable version of x[n_steps - 1::-1].

theano/scan_module/scan_op.py
```diff
@@ -2229,15 +2237,15 @@ def compute_all_gradients(known_grads):
...
             for x in self.outer_sitsot_outs(outs):
                 if hasattr(x[::-1][:-1].tag, 'test_value'):
                     assert (x[::-1][:-1].tag.test_value.shape[0] ==
-                            inputs[0].tag.test_value)
+                            grad_steps.tag.test_value)
             for x in self.outer_nitsot_outs(outs):
                 if hasattr(x[::-1].tag, 'test_value'):
-                    assert (x[::-1].tag.test_value.shape[0] ==
-                            inputs[0].tag.test_value)
+                    assert (x[grad_steps - 1::-1].tag.test_value.shape[0] ==
```
nouiz: should there be a condition here with as_while?

theano/scan_module/scan_op.py
```diff
@@ -2564,7 +2573,19 @@ def compute_all_gradients(known_grads):
...
                 type_outs[offset:offset + self.n_seqs])):
             if t == 'connected':
-                gradients.append(x[::-1])
+                # If the forward scan is in as_while mode, we need to pad
+                # the gradients, so that they match the size of the input
+                # sequences.
+                if self.as_while:
+                    n_zeros = inputs[0] - n_steps
+                    shp = (n_zeros,)
+                    if x.ndim > 1:
```
nouiz: The if is useless. If it is of ndim 0 or 1, you would concatenate an empty tuple to shp. So the answer would be good. No harm to let it here.

theano/scan_module/tests/test_scan.py
```diff
@@ -5529,3 +5529,41 @@ def count_up():
 
         _, updates = theano.scan(count_up, n_steps=20)
         func = theano.function(inputs=[inc], outputs=[], updates=updates)
+
+
+class TestGradUntil(unittest.TestCase):
+
+    def setUp(self):
+        self.x = tensor.vector(name='x')
+        self.until = tensor.scalar(name='until', dtype='int64')
```
nouiz: I don't like the name until here. I don't have a much better name. The only thing that come to mind is cond, but it isn't the full condition. What about threshold?

theano/scan_module/scan_op.py
Thrandis: done

theano/scan_module/scan_op.py
Thrandis: Done

theano/scan_module/tests/test_scan.py
Thrandis: Done

theano/scan_module/scan_op.py
Thrandis: done

PR #5778, opened 2017-03-28
Add magma gpu svd and matrix inverse

aam-at:
Adds magma gpu for computing svd and matrix inverse.

Benchmark results of magma library against lapack on 6 core cpu and Titan X. Results were taken from magma testing suite. Similar performance should be expected for theano.

<a id="orgfc86fbd"></a>

## SVD

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-right" />

<col  class="org-right" />

<col  class="org-right" />
</colgroup>
<tbody>
<tr>
<td class="org-right">M</td>
<td class="org-right">CPU time (sec)</td>
<td class="org-right">GPU time (sec)</td>
</tr>

<tr>
<td class="org-right">1088</td>
<td class="org-right">0.1155</td>
<td class="org-right">0.2345</td>
</tr>

<tr>
<td class="org-right">2112</td>
<td class="org-right">0.5797</td>
<td class="org-right">0.5252</td>
</tr>

<tr>
<td class="org-right">3136</td>
<td class="org-right">2.5200</td>
<td class="org-right">1.1590</td>
</tr>

<tr>
<td class="org-right">4160</td>
<td class="org-right">6.1262</td>
<td class="org-right">2.1075</td>
</tr>

<tr>
<td class="org-right">5184</td>
<td class="org-right">15.5205</td>
<td class="org-right">4.0243</td>
</tr>
</tbody>
</table>

<a id="org50a95e3"></a>

## Matrix inverse

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-right" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<tbody>
<tr>
<td class="org-right">N</td>
<td class="org-left">CPU Gflop/s (sec)</td>
<td class="org-left">GPU Gflop/s (sec)</td>
</tr>

<tr>
<td class="org-right">1088</td>
<td class="org-left">78.01 (0.02)</td>
<td class="org-left">118.18 (0.01)</td>
</tr>

<tr>
<td class="org-right">2112</td>
<td class="org-left">132.95 (0.09)</td>
<td class="org-left">272.37 (0.05)</td>
</tr>

<tr>
<td class="org-right">3136</td>
<td class="org-left">135.28 (0.30)</td>
<td class="org-left">571.12 (0.07)</td>
</tr>

<tr>
<td class="org-right">4160</td>
<td class="org-left">139.91 (0.69)</td>
<td class="org-left">909.70 (0.11)</td>
</tr>

<tr>
<td class="org-right">5184</td>
<td class="org-left">141.60 (1.31)</td>
<td class="org-left">1358.48 (0.14)</td>
</tr>
</tbody>
</table>

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -361,6 +361,23 @@ AddConfigVar('dnn.enabled',
              EnumStr("auto", "True", "False", "no_check"),
              in_c_key=False)
 
+AddConfigVar('magma.include_path',
+             "Location of the magma header",
+             StrParam(None),
+             in_c_key=False)
+
+AddConfigVar('magma.library_path',
+             "Location of the magma library",
+             StrParam(None),
+             in_c_key=False)
+
+AddConfigVar('magma.enabled',
+             " If True, use magma for matrix computation."
+             " If False, disable magma"
+             " If no_check, assume present and the version between header and library match (so less compilation at context init)",
+             BoolParam(False),
+             in_c_key=False)
+
 # This flag determines whether or not to raise error/warning message if
 # there is a CPU Op in the computational graph.
 AddConfigVar(
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -363,12 +363,12 @@ AddConfigVar('dnn.enabled',
 
 AddConfigVar('magma.include_path',
              "Location of the magma header",
-             StrParam(None),
+             StrParam(''),
              in_c_key=False)
 
 AddConfigVar('magma.library_path',
              "Location of the magma library",
-             StrParam(None),
+             StrParam(''),
              in_c_key=False)
 
 AddConfigVar('magma.enabled',
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -373,8 +373,7 @@ AddConfigVar('magma.library_path',
 
 AddConfigVar('magma.enabled',
              " If True, use magma for matrix computation."
-             " If False, disable magma"
-             " If no_check, assume present and the version between header and library match (so less compilation at context init)",
+             " If False, disable magma",
              BoolParam(False),
              in_c_key=False)
 
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -357,8 +357,7 @@ class GpuMagmaSVD(COp):
     def __init__(self, full_matrices=True, compute_uv=True):
         self.full_matrices = full_matrices
         self.compute_uv = compute_uv
-        COp.__init__(self, ['magma_svd.c'],
-                     'APPLY_SPECIFIC(magma_svd)')
+        COp.__init__(self, ['magma_svd.c'], 'APPLY_SPECIFIC(magma_svd)')
 
     def c_headers(self):
         return ['gpuarray/types.h', 'gpuarray/array.h', 'gpuarray/ext_cuda.h',
@@ -429,9 +428,10 @@ class GpuMagmaMatrixInverse(COp):
     params_type = gpu_context_type
 
     def __init__(self, inplace=False):
-        COp.__init__(self, ['magma_inv.c'],
-                     'APPLY_SPECIFIC(magma_inv)')
+        COp.__init__(self, ['magma_inv.c'], 'APPLY_SPECIFIC(magma_inv)')
         self.inplace = inplace
+        if self.inplace:
+            self.destroy_map = {0: [0]}
 
     def c_headers(self):
         return ['gpuarray/types.h', 'gpuarray/array.h', 'gpuarray/ext_cuda.h',
@@ -450,6 +450,9 @@ class GpuMagmaMatrixInverse(COp):
         if config.magma.library_path:
             return [config.magma.library_path]
         return []
+
+    def clone_inplace(self):
+        return self.__class__(inplace=True)
 
     def make_node(self, x):
         ctx_name = infer_context_name(x)
@@ -471,4 +474,18 @@ class GpuMagmaMatrixInverse(COp):
         return shapes
 
 
-gpu_matrix_inverse = GpuMagmaMatrixInverse()
+def gpu_matrix_inverse(a, inplace=False):
+    """
+    This function performs the matrix inverse on GPU.
+
+    Parameters
+    ----------
+    inplace : bool, optional
+        Whether or not to compute matrix inverse inplace.
+
+    Returns
+    -------
+    a_inv: matrix
+
+    """
+    return GpuMagmaMatrixInverse(inplace=inplace)(a)
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -355,12 +355,13 @@ class GpuMagmaMatrixInverse(COp):
     params_type = gpu_context_type
 
     def __init__(self, inplace=False):
-        COp.__init__(self, ['magma_linalg.c'],
-                     'APPLY_SPECIFIC(magma_matrix_inv)')
+        COp.__init__(self, ['magma_inv.c'],
+                     'APPLY_SPECIFIC(magma_inv)')
         self.inplace = inplace
 
     def c_headers(self):
-        return ['gpuarray/array.h', 'gpuarray/blas.h', 'gpuarray_helper.h', 'magma.h']
+        return ['gpuarray/types.h', 'gpuarray/array.h', 'gpuarray/ext_cuda.h',
+                'gpuarray_helper.h', 'magma.h']
 
     def c_header_dirs(self):
         return [os.path.dirname(__file__), pygpu.get_include()]
@@ -371,8 +372,8 @@ class GpuMagmaMatrixInverse(COp):
     def make_node(self, x):
         if x.ndim != 2:
             raise LinAlgError("Matrix rank error")
-        context_name = infer_context_name(x)
-        x = as_gpuarray_variable(x, context_name)
+        ctx_name = infer_context_name(x)
+        x = as_gpuarray_variable(x, ctx_name)
         return theano.Apply(self, [x], [x.type()])
 
     def get_params(self, node):
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -1,14 +1,19 @@
 from __future__ import absolute_import, division, print_function
 
-import pkg_resources
-import theano
+import os
 import warnings
 
-from theano import Op
-from theano.gpuarray import basic_ops, GpuArrayType
-
 import numpy as np
+import pkg_resources
 from numpy.linalg.linalg import LinAlgError
+
+import theano
+from theano import Op
+from theano.gof import COp
+from theano.gpuarray import GpuArrayType
+
+from .basic_ops import as_gpuarray_variable, gpu_contiguous, infer_context_name
+from .type import gpu_context_type
 
 try:
     import pygpu
@@ -94,13 +99,13 @@ class GpuCusolverSolve(Op):
                                'GpuCusolverSolve Op can not be constructed.')
         if skcuda.__version__ <= '0.5.1':
             warnings.warn('The GpuSolve op requires scikit-cuda > 0.5.1 to work with CUDA 8')
-        context_name = basic_ops.infer_context_name(inp1, inp2)
+        context_name = infer_context_name(inp1, inp2)
 
-        inp1 = basic_ops.as_gpuarray_variable(inp1, context_name)
-        inp2 = basic_ops.as_gpuarray_variable(inp2, context_name)
+        inp1 = as_gpuarray_variable(inp1, context_name)
+        inp2 = as_gpuarray_variable(inp2, context_name)
 
-        inp1 = basic_ops.gpu_contiguous(inp1)
-        inp2 = basic_ops.gpu_contiguous(inp2)
+        inp1 = gpu_contiguous(inp1)
+        inp2 = gpu_contiguous(inp2)
 
         # this op can only operate on float32 matrices
         assert inp1.ndim == 2
@@ -260,11 +265,11 @@ class GpuCholesky(Op):
         if not pygpu_available:
             raise RuntimeError('Missing pygpu or triu/tril functions.'
                                'Install or update libgpuarray.')
-        context_name = basic_ops.infer_context_name(inp)
+        context_name = infer_context_name(inp)
 
-        inp = basic_ops.as_gpuarray_variable(inp, context_name)
+        inp = as_gpuarray_variable(inp, context_name)
 
-        inp = basic_ops.gpu_contiguous(inp)
+        inp = gpu_contiguous(inp)
 
         # this op can only operate on float32 matrices
         # because of current implementation of triu/tril.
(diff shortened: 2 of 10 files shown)


theano/gpuarray/linalg.py
```diff
@@ -212,3 +217,129 @@ def perform(self, node, inputs, outputs):
...
+    def c_lib_dirs(self):
+        if config.magma.library_path:
+            return [config.magma.library_path]
+        return []
+
+    def make_node(self, A):
+        if A.ndim != 2:
+            raise LinAlgError("Matrix rank error")
+        ctx_name = infer_context_name(A)
+        A = as_gpuarray_variable(A, ctx_name)
```
nouiz: Move that line before the if. In case the user pass a list of list, ndim won't exist in that case.

theano/gpuarray/linalg.py
```diff
@@ -212,3 +217,129 @@ def perform(self, node, inputs, outputs):
...
+    def c_lib_dirs(self):
+        if config.magma.library_path:
+            return [config.magma.library_path]
+        return []
+
+    def make_node(self, x):
+        if x.ndim != 2:
+            raise LinAlgError("Matrix rank error")
+        ctx_name = infer_context_name(x)
+        x = as_gpuarray_variable(x, ctx_name)
```
nouiz: idem

theano/gpuarray/magma_inv.c
```diff
@@ -0,0 +1,101 @@
...
+  dims = PyGpuArray_DIMS(A);
+  if (dims[0] != dims[1]) {
+    PyErr_SetString(PyExc_ValueError,
+                    "GpuMagmaMatrixInverse: matrix is not square");
+    goto fail;
+  }
+#ifdef INPLACE
+  Py_XDECREF(out);
+  A_inv = A;
+  Py_INCREF(out);
```
nouiz: out isn't defined. If the op get inplace, the compilation will crash.

theano/gpuarray/tests/test_linalg.py
```diff
@@ -112,3 +115,75 @@ def test_linalgerr_solve(self):
 
         fn = theano.function([A, b], [solver], mode=mode_with_gpu)
         self.assertRaises(LinAlgError, fn, A_val, x_val)
+
+
+class TestMagma(unittest.TestCase):
+    def test_gpu_matrix_inverse(self):
```
nouiz: Make the tests skipped if magma isn't there.

Add a test for the inplace matrix_inverse

theano/tensor/nlinalg.py
```diff
@@ -625,9 +625,12 @@ def perform(self, node, inputs, outputs):
...
         assert x.ndim == 2, "The input of svd function should be a matrix."
-        w[0], u[0], v[0] = self._numop(x,
-                                       self.full_matrices,
-                                       self.compute_uv)
+        if self.compute_uv:
+            w[0], u[0], v[0] = self._numop(x,
+                                           self.full_matrices,
+                                           self.compute_uv)
+        else:
+            u[0] = self._numop(x, self.full_matrices, self.compute_uv)
```
nouiz: This change show that we need to test compute_uv. It is probably not tested, as the old code should crash.

Also, if we compute only 1 output, the make_node should have only 1 output.

theano/gpuarray/linalg.py
```diff
@@ -212,3 +217,129 @@ def perform(self, node, inputs, outputs):
...
+class GpuMagmaMatrixInverse(COp):
+    """Computes the inverse of a matrix :math:`A` using magma library.
+    """
+    __props__ = ('inplace', )
+    params_type = gpu_context_type
+
+    def __init__(self, inplace=False):
+        COp.__init__(self, ['magma_inv.c'],
+                     'APPLY_SPECIFIC(magma_inv)')
+        self.inplace = inplace
```
nouiz: the inplace isn't working. Do you want to implement it? If so, you should add here:
if inplace:
    self.destry_map = {0:[0]}

And you should create an optimization to convert the non inplace version with the inplace version see the opt local_advincsub1_gpua_inplace in gpuarray/opt.py as an example.

The c code for inplace also need a update, I have another comment about that.

nouiz:
Did you saw timing between cusolve svd and magma one?

theano/tensor/nlinalg.py
aam-at: I have numpy 1.12 and if `compute_uv = True` it returns only singular values. Which version of numpy does theano use for testing?

aam-at:
@nouiz Thanks for the comments.

I used scikit-cuda demos to profile cusolver and magma. For `1000x1000` matrices, cusolver is about x2 times slower than magma. I didn't profile other matrix sizes.

Before starting this, I implemented cusolver svd with scikit-cuda in other branch [gpu_svd_op](https://github.com/aam-at/theano/tree/gpu_svd_op). `cusolver` interface is limited (only supports `full_matrices = True`; works only with wide matrices) and performance was unsatisfying. `magma` for matrix computations seems more promising and it is also used in other libraries for gpu matrix computations (e.g. pytorch).

theano/tensor/nlinalg.py
nouiz: I don't know which version of numpy is installed, but I think your fix is good and don't depend of numpy version. What trouble me is that why tests didn't catch that crash? Do you have a Theano test that crash without that fix? I suppose not, so this would mean this case isn't tested.

nouiz:
thanks for the speed information. Can you confirm that when you timed cusolve, was this with a python interface as skcuda or it was compared with c code? Just the inteface itself could cause this difference.

It is good to know other framework use the same, it point that it is better for our cases. The sad part is that it make installation more complicated, so if using cusolve wasn't slower, for that reason it could be a better library to use. But that isn't the case.

aam-at:
@nouiz Both tested scikit-demos use python interface using ctypes.

Yeah, and unfortunately `magma` is not yet distributed via pip or ubuntu packages. I guess I also need to edit travis config to add installation and compilation for magma.

nouiz:
travis don't test the gpu, but our jenkins setup does. You won't have access to it, so it is to us to add it :)

theano/gpuarray/linalg.py
aam-at: @nouiz I added inplace optimization.

theano/configdefaults.py
```diff
@@ -419,6 +419,23 @@ def f(suffix=suffix):
...
+
+AddConfigVar('magma.library_path',
+             "Location of the magma library",
+             StrParam(None),
+             in_c_key=False)
+
+AddConfigVar('magma.enabled',
+             " If True, use magma for matrix computation."
+             " If False, disable magma"
+             " If no_check, assume present and the version between header and library match (so less compilation at context init)",
```
aam-at: Removed.

theano/gpuarray/magma_inv.c
aam-at: Fixed.

theano/gpuarray/tests/test_linalg.py
aam-at: Fixed.

theano/tensor/nlinalg.py
aam-at: @nouiz yeah, old tests do not check when `compute_uv = False`. How can I return only one output from `Apply` node without duplicating C code?

theano/gpuarray/linalg.py
aam-at: Fixed.

aam-at:
@nouiz I have addressed your comments. Is anything else necessary for this pull request?

theano/configdefaults.py
```diff
@@ -361,6 +361,22 @@ def default_dnn_library_path():
              EnumStr("auto", "True", "False", "no_check"),
              in_c_key=False)
 
+AddConfigVar('magma.include_path',
+             "Location of the magma header",
+             StrParam(None),
```
nouiz: I think None should be repalced by "". Otherwise I think it end up as 'None' as default value.

Also, they would need to be documented in doc/library/config.txt

theano/gpuarray/linalg.py
```diff
@@ -335,3 +340,146 @@ def perform(self, node, inputs, outputs):
...
+                             A.type()])
+
+    def get_params(self, node):
+        return node.inputs[0].type.context
+
+    def get_op_params(self):
+        compute_uv = int(self.compute_uv)
+        full_matrices = int(self.full_matrices)
+        return [('COMPUTE_UV', compute_uv),
+                ('FULL_MATRICES', full_matrices)]
```
nouiz: I think you can easily add the infer_shape() method.

theano/tensor/nlinalg.py
nouiz: For the CPU op, as there is no C code, this isn't a problem.

For the GPU op, as it is a props, you can do it. I think this is an old questions that you found the answer.

Can you add a CPU test for that and fix the CPU make_node?

nouiz:
jenkins test this

theano/gpuarray/magma_inv.c
```diff
@@ -0,0 +1,101 @@
+#section init_code
+
+setup_ext_cuda();
+
+#section support_code_struct
+
+int APPLY_SPECIFIC(magma_inv)(PyGpuArrayObject *A, PyGpuArrayObject **_A_inv,
+                              PyGpuContextObject *c) {
+  PyGpuArrayObject *A_inv = *_A_inv;
```
notoraptor: I think you should always use `*_A_inv` instead of a temporar value `A_inv`. If a copy of `A_inv` is done (at line 47) but an error occurs after that and before setting `*_A_inv = A_inv`, then the memory allocated for the copy of `A_inv` will not be freed (as it is not checked at all at `fail` label) and will not be seen/handled by Theano (as `*_A_inv` will have not been updated).

aam-at:
@nouiz I added `infer_shape` to both cpu and gpu interface. I also updated `make_node` for both cpu and gpu versions.
@notoraptor I fixed that problem.

Jenkins currently skips `magma` tests.

theano/gpuarray/linalg.py
```diff
@@ -335,3 +340,165 @@ def perform(self, node, inputs, outputs):
...
+        if self.inplace:
+            return [('INPLACE', '1')]
+        else:
+            return []
+
+    def infer_shape(self, node, shapes):
+        return shapes
+
+
+def gpu_matrix_inverse(a, inplace=False):
```
notoraptor: I think you should remove the argument `inplace`, as it is intended to be used internally by Theano for optimizations, not by users. Then you could just write:

`gpu_matrix_inverse = GpuMatrixInverse()`

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,100 @@ def test_invalid_input_fail_negative_definite(self):
...
+class TestMagma(unittest.TestCase):
+
+    def setUp(self):
+        if not config.magma.enabled:
+            self.skipTest('Magma is not enabled, skipping test')
+
+    def test_gpu_matrix_inverse(self):
+        A = theano.tensor.fmatrix("A")
+
+        fn = theano.function([A], gpu_matrix_inverse(A), mode=mode_with_gpu.including('magma'))
```
notoraptor: For all your tests, could you test if it works without `.including('magma')` (ie. just with `mode=mode_with_gpu`) ?

theano/tensor/tests/test_nlinalg.py
```diff
@@ -128,18 +129,47 @@ def test_qr_modes():
...
+        super(test_SVD, self).setUp()
+        self.rng = np.random.RandomState(utt.fetch_seed())
+        self.A = theano.tensor.matrix(dtype=self.dtype)
+        self.op = svd
 
-    assert _allclose(n_u, t_u)
-    assert _allclose(n_v, t_v)
-    assert _allclose(n_t, t_t)
+    def test_svd(self):
+        rng = np.random.RandomState(utt.fetch_seed())
```
notoraptor: You could use `self.rng` instead, isn't it ?

theano/gpuarray/tests/test_linalg.py
aam-at: I removed that. My original idea was that the user can enable magma by enabling its optimizer. But approach with config `magma.enable` seems better.

theano/gpuarray/linalg.py
aam-at: Fixed.

theano/tensor/tests/test_nlinalg.py
aam-at: Fixed.

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,97 @@ def test_invalid_input_fail_negative_definite(self):
...
+    def setUp(self):
+        if not config.magma.enabled:
+            self.skipTest('Magma is not enabled, skipping test')
+
+    def test_gpu_matrix_inverse(self):
+        A = theano.tensor.fmatrix("A")
+
+        fn = theano.function([A], gpu_matrix_inverse(A), mode=mode_with_gpu)
+        N = 1000
+        A_val = rand(N, N)
```
notoraptor: As your code only accepts float32 (`fmatrix` in Python, `GA_FLOAT` checking in C), always use float32 inputs, so that tests can pass even on 64 bits machines (where floatX is float64 by default). Here:
```python
A_val = rand(N, N).astype('float32')
```

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,97 @@ def test_invalid_input_fail_negative_definite(self):
...
+        A_val_inv = fn(A_val)
+        utt.assert_allclose(np.dot(A_val_inv, A_val), np.eye(N), atol=1e-3)
+
+    def test_gpu_matrix_inverse_inplace(self):
+        N = 1000
+        A_val_gpu = gpuarray_shared_constructor(rand(N, N))
```
notoraptor: Same here (`rand(N, N).astype('float32')`).

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,97 @@ def test_invalid_input_fail_negative_definite(self):
...
+        utt.assert_allclose(np.dot(Ot.T, Ot), np.eye(Ot.shape[1]))
+
+    def check_svd(self, A, U, S, VT, rtol=None, atol=None):
+        S_m = np.zeros_like(A)
+        np.fill_diagonal(S_m, S)
+        utt.assert_allclose(
+            np.dot(np.dot(U, S_m), VT), A, rtol=rtol, atol=atol)
+
+    def test_gpu_svd_wide(self):
+        A = rand(100, 50)
```
notoraptor: Same here.

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,97 @@ def test_invalid_input_fail_negative_definite(self):
...
+        self.check_svd(A, U, S, VT)
+
+        U, S, VT = self.run_gpu_svd(A, full_matrices=False)
+        self.assertEqual(U.shape[1], min(M, N))
+        self.assert_column_orthonormal(U)
+        self.assertEqual(VT.shape[0], min(M, N))
+        self.assert_column_orthonormal(VT.T)
+
+    def test_gpu_svd_tall(self):
+        A = rand(50, 100)
```
notoraptor: Same here.

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,97 @@ def test_invalid_input_fail_negative_definite(self):
...
+
+    def test_gpu_singular_values(self):
+        A = theano.tensor.fmatrix("A")
+        f_cpu = theano.function(
+            [A], theano.tensor.nlinalg.svd(A, compute_uv=False),
+            mode=mode_without_gpu)
+        f_gpu = theano.function(
+            [A], gpu_svd(A, compute_uv=False), mode=mode_with_gpu)
+
+        A_val = rand(50, 100)
```
notoraptor: Same here.

theano/gpuarray/tests/test_linalg.py
```diff
@@ -191,3 +199,97 @@ def test_invalid_input_fail_negative_definite(self):
...
+        utt.assert_allclose(f_cpu(A_val), f_gpu(A_val))
+
+        A_val = rand(100, 50)
```
notoraptor: Same here.

theano/gpuarray/tests/test_linalg.py
aam-at: Fixed.

notoraptor:
NB: There is a single conflict because of a PR recently merged that modifies an import statement in `gpuarray/opt.py`. Could you rebase your PR ?

aam-at:
@notoraptor I did rebase.

nouiz:
thanks!

theano/gpuarray/magma_inv.c
```diff
@@ -0,0 +1,99 @@
...
+
+  ldwork = N * magma_get_sgetri_nb(N);
+  dwork = gpudata_alloc(c->ctx, ldwork * sizeof(float), NULL, 0, NULL);
+  if (dwork == NULL) {
+    PyErr_SetString(PyExc_RuntimeError,
+                    "GpuMagmaMatrixInverse: failed to allocate working memory");
+    goto fail;
+  }
+
+  if (magma_imalloc_cpu(&piv, N)) {
```
botev: Out of curiosity, why is the pivot array always allocated on the CPU?

aam-at: `piv` is allocated on `cpu` in magma testing suite.

PR #5779, opened 2017-03-28
Update css.inc

Naereen:
Faster for the DOM parser to have one <style> tag with 5 lines than 5 <style> tags with 1 line.


Naereen:
:+1:

PR #5782, opened 2017-03-29
Updated numpy as np #4218

Amrithasuresh:
@nouiz 

As per the suggestion https://github.com/Theano/Theano/pull/5763#issuecomment-289902759 , I have updated the file https://github.com/Theano/Theano/issues/4218.

Thank you.


nouiz:
thanks

PR #5787, opened 2017-03-29
Fix out-of-date link to sympy

ageitgey:
I just noticed that the current link to `sympy` in the docs is pointing to a dead Google Code site. This just updates the link to the current webpage.

Feel free to close this if you have some other way you prefer to make documentation updates.


nouiz:
thanks

nouiz:
note, I included that in the 0.9.X branch so the default doc show it too.

Issue #2763, opened 2015-04-14
Implement GpuEye with k!=0


lamblin:
`GpuEye.make_node` currently contains:
```python
        # k != 0 isn't implemented on the GPU yet.
        assert tensor.get_scalar_constant_value(k) == 0
```
We should implement the missing cases.

nouiz:
Should we make the optimization not introduce it to remove the crash until that time?

lamblin:
I'm not against it, but the proper fix should not be that long to implement.

trungnt13:
Have the case implemented in newest Theano?

nouiz:
Nobody stared to implement it. Do you want to try?

ReyhaneAskari:
@shawntan 
This ticket is about completing the implementation of [```GpuEye```](https://github.com/Theano/Theano/blob/master/theano/gpuarray/basic_ops.py#L1574) in the new backend (gpuarray). Here is a simple code that uses the ```GpuEye``` (run it with ```THEANO_FLAGS='device=cuda,floatX=flaot32'```):

```python
import theano
import numpy as np
from theano import tensor as T

n = T.iscalar('n')
m = T.iscalar('m')
k = np.asarray(0)

e = T.eye(n, m, k) + 1
f = theano.function([n, m], e)
print f(4, 4)
```
 The ```GpuEye``` doesn't have the implementation of _diagonal offset_ (```k```) and always uses the main diagonal (```k = 0```). So in the above code if you change  ```k``` to 1, you will get an assertion error.  Your task is to add this functionality in C. There are three functions that need changes;

- [ ] [gpu_kernels](https://github.com/Theano/Theano/blob/master/theano/gpuarray/basic_ops.py#L1614) 
- [ ] [c_code](https://github.com/Theano/Theano/blob/master/theano/gpuarray/basic_ops.py#L1629)
- [ ] [c_code_cache_version](https://github.com/Theano/Theano/blob/master/theano/gpuarray/basic_ops.py#L1671)
- [ ] And at the end you will need to write a test that tests this new functionality. 

Note that before starting you implementation, disable the cache by writing a ```return``` in the ```c_code_cache_version```. Otherwise theano will use cache and you will not see your changes. 
When your implementation was done, you should remove the  ```return``` and increment the cache version to 7. 

When you were ready to test it, write your test in _theano/gpuarray/tests/test_basic_ops.py._
[...]

PR #5789, opened 2017-03-30
Changes to GpuEye to enable `k` offset parameter.

shawntan:
The tests will be coming up, thought I'd do a pull request to get some feedback before I carry on.

@ReyhaneAskari #2763

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1611,9 +1611,9 @@ class GpuEye(GpuKernelBase, Op):
 
     def gpu_kernels(self, node, name):
         code = """
-KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_long k) {
-    ga_long col_off = max(k, (ga_long) 0);
-    ga_long row_off = -min(k, (ga_long) 0);
+KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_ssize k) {
+    ga_ssize col_off = max(k, (ga_ssize) 0);
+    ga_ssize row_off = -min(k, (ga_ssize) 0);
     if (row_off < n && col_off < m) {
         ga_size nb = (ga_size) min(n - row_off, m - col_off);
         for (ga_size i = LID_0; i < nb; i += LDIM_0) {
@@ -1624,7 +1624,7 @@ KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_long k) {
             name=name, write_a=write_w(self.dtype))
         return [Kernel(
                 code=code, name="eye",
-                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE, 'int64'],
+                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE, gpuarray.SSIZE],
                 flags=Kernel.get_flags(self.dtype),
                 objvar='k_eye_' + name)]
 
@@ -1639,7 +1639,7 @@ KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_long k) {
         s = """
         size_t dims[2] = {0, 0};
         size_t ls, gs;
-        long k;
+        ssize_t k;
         int err;
 
         dims[0] = ((dtype_%(n)s*)PyArray_DATA(%(n)s))[0];
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1613,19 +1613,20 @@ class GpuEye(GpuKernelBase, Op):
 
     def gpu_kernels(self, node, name):
         code = """
-KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_int k) {
-    ga_int col_off = max(k, (ga_int) 0);
-    ga_int row_off = -min(k, (ga_int) 0);
-    ga_size nb = (ga_size) min(n - row_off, m - col_off);
-    nb = max(nb, (ga_size) 0);
-    for (ga_size i = LID_0; i < nb; i += LDIM_0) {
-        a[(i + row_off)*m + i + col_off] = %(write_a)s(1);
+KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_long k) {
+    ga_long col_off = max(k, (ga_long) 0);
+    ga_long row_off = -min(k, (ga_long) 0);
+    if (row_off < n && col_off < m) {
+        ga_size nb = (ga_size) min(n - row_off, m - col_off);
+        for (ga_size i = LID_0; i < nb; i += LDIM_0) {
+            a[(i + row_off)*m + i + col_off] = %(write_a)s(1);
+        }
     }
 }""" % dict(ctype=pygpu.gpuarray.dtype_to_ctype(self.dtype),
             name=name, write_a=write_w(self.dtype))
         return [Kernel(
                 code=code, name="eye",
-                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE, np.int32],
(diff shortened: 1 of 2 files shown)


shawntan:
- `k` is currently of type `int32`. Should it be `int64`?
- The code currently allows `|k| > max(m, n)`, the result will just be zeroes. Let me know if this is not the desired behaviour
- I've been testing it with the following cases:

```python
import theano
import theano.tensor as T
import numpy as np
n = T.iscalar('n')
m = T.iscalar('m')
k = theano.shared(np.int32(0))

e = T.eye(n, m, k) + 1
f_ = theano.function([n, m], e)


def f(a, b):
    return f_(a, b) - 1

print f(4, 4)
k.set_value(np.int32(1))
print f(4, 4)
k.set_value(np.int32(2))
print f(4, 4)
k.set_value(np.int32(3))
print f(4, 4)
print f(8, 4)
print f(4, 8)
k.set_value(np.int32(-3))
print f(4, 4)
print f(8, 4)
print f(4, 8)
k.set_value(np.int32(-4))
print f(4, 4)
print f(8, 4)
print f(4, 8)
k.set_value(np.int32(4))
print f(4, 4)
print f(8, 4)
print f(4, 8)
```

theano/gpuarray/basic_ops.py
```diff
@@ -1613,21 +1613,24 @@ def grad(self, inp, grads):
...
     for (ga_size i = LID_0; i < nb; i += LDIM_0) {
-        a[i*m + i] = %(write_a)s(1);
+        a[(i + row_off)*m + i + col_off] = %(write_a)s(1);
     }
 }""" % dict(ctype=pygpu.gpuarray.dtype_to_ctype(self.dtype),
             name=name, write_a=write_w(self.dtype))
         return [Kernel(
                 code=code, name="eye",
-                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE],
+                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE, np.int32],
```
ReyhaneAskari: You need to indicate the type of ```k``` as ```'int32'```. Take a look at [this example](https://github.com/Theano/Theano/blob/99e32f41ec92fe2d123c7770074b99674d9d5073/theano/gpuarray/elemwise.py#L2626).

nouiz:
I don't think k need to be int64. Don't forget, this is approx the max
number of lines/columns. So an k that don't fit in int32 would be very thin
or tall. But I don't think int64 would hurt.

shawntan:
I've added the same test cases I found for the [cpu version](https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_basic.py) for `T.eye`, but I think in both there should be further test cases where `k > M` or `-k > N` depending on whether `k` is negative.

I've removed the `stack` call, not quite sure why that was there. If there is a reason for it, let me know.

theano/gpuarray/basic_ops.py
```diff
@@ -1613,21 +1611,25 @@ def grad(self, inp, grads):
 
     def gpu_kernels(self, node, name):
         code = """
-KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m) {
-    ga_size nb = n < m ? n : m;
-    for (ga_size i = LID_0; i < nb; i += LDIM_0) {
-        a[i*m + i] = %(write_a)s(1);
+KERNEL void eye(GLOBAL_MEM %(ctype)s *a, ga_size n, ga_size m, ga_long k) {
```
abergeron: Use ga_ssize for k to be consistent with the other two args.

theano/gpuarray/basic_ops.py
```diff
@@ -1613,21 +1611,25 @@ def grad(self, inp, grads):
...
+    ga_long col_off = max(k, (ga_long) 0);
+    ga_long row_off = -min(k, (ga_long) 0);
+    if (row_off < n && col_off < m) {
```
abergeron: Instead of doing the check in the kernel, it would probably be a better idea to do this check outside and skip the kernel call completely when the output would be all-zeros.

abergeron: I'm not 100% sure that would work in all cases though.

theano/gpuarray/basic_ops.py
shawntan: k needs to be negative, ssize is an unsigned long right?

theano/gpuarray/basic_ops.py
shawntan: Right.

Do you mean doing the check wouldn't work in all cases, or that the current code as it is wouldn't work in all cases?

theano/gpuarray/basic_ops.py
abergeron: ssize is signed.  size is unsigned.

theano/gpuarray/basic_ops.py
abergeron: I mean that moving the check outside might not always work.

theano/gpuarray/basic_ops.py
shawntan: Okay thanks!

theano/gpuarray/basic_ops.py
```diff
@@ -1674,4 +1679,5 @@ def c_code(self, node, name, inp, out, sub):
         return s
 
     def c_code_cache_version(self):
-        return (7,)
+        # return (7,)
+        return
```
ReyhaneAskari: I guess you forgot to remove the return in here.

theano/gpuarray/basic_ops.py
```diff
@@ -1658,7 +1658,12 @@ def c_code(self, node, name, inp, out, sub):
...
         ls = 1;
         gs = 256;
-        err = eye_call(1, &gs, &ls, 0, %(z)s->ga.data, dims[0], dims[1], k);
+        col_off = (size_t) (k > 0?k:0);
+        row_off = (size_t) (k < 0?-k:0);
+        if (row_off < dims[0] && col_off < dims[1]) {
+            err = eye_call(1, &gs, &ls, 0, %(z)s->ga.data, dims[0], dims[1], k);
+        }
+
         if (err != GA_NO_ERROR) {
```
ReyhaneAskari: I think you will get an error here. You have defined ```err``` but have not initialized it. It will be indeterminate.

shawntan: No error there, because err has already been initialised, probably to the same value as `GA_NO_ERROR`. But you're right it doesn't make sense, I should move it in.

theano/gpuarray/basic_ops.py
shawntan: Argh I always forget. Thanks!

shawntan:
Not sure where the flake8 errors are occuring. Running the command locally doesn't reproduce it.

Seems to be happening on other builds too.

abergeron:
The flake8 problem is a bug in sphinx on Travis, your code is fine.

nouiz:
jenkins test this

theano/gpuarray/basic_ops.py
```diff
@@ -1613,21 +1611,23 @@ def grad(self, inp, grads):
...
         return [Kernel(
                 code=code, name="eye",
-                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE],
+                params=[gpuarray.GpuArray, gpuarray.SIZE, gpuarray.SIZE, gpuarray.SSIZE],
                 flags=Kernel.get_flags(self.dtype),
                 objvar='k_eye_' + name)]
 
     def c_code(self, node, name, inp, out, sub):
-        n, m = inp
+        n, m, k = inp
```
nouiz: Here you should check if inp have 2 or 3 inputs. If only 2, then suppose k is 0 and hardcode zero.

This is to support old pickled files.

lamblin:
Jenkins test this please

shawntan:
I'm a bit clueless as to how to deal with the errors here.
I've pulled the latest from master and testing them locally on GPU seems to be fine.

ReyhaneAskari:
There seems to be a problem either with jenkins or with theano's master that needs to be fixed. Your ticket seems to be complete. Just wait for an update.
Thanks.

nouiz:
jenkins test this

nouiz:
closed and reopening to trigger the CI with the new base.

PR #5793, opened 2017-03-31
OpFromGraph on GPU when not inline

khaotik:
This relies on optimizer's ability to remove redundant `<gpu> -> <cpu> -> <gpu>` transfers so a bit hacky.


khaotik:
@nouiz 

Yes, adding `device` parameter is definitely a good idea. However I still don't understand the complicated version. What does "select each input individually" mean? Does it mean "do transfer based on whether the storage is on CPU/GPU"? (I think `var.transfer()` is already doing that?) What would be an example corner case for that?

Another question, is there a way to force a part of main graph (not inside `op.fn`) being executed on a specific device *with* all opt turned on? I think it's a good idea to make both inline/compiled OpFromGraph being executed on the same device, so more consistent results.

nouiz:
By default, if all the inputs of the OpFromGraph is one one device, all the compuation inside the OpFromGraph should be on the same device. We don't support well multiple device in one graph, so I would not butter with that too much. But you can force that when creating the GpuFromHost in the inner graph. Put the device you want the user of them will use.

khaotik:
@nouiz Sorry for the delay, I was kinda occupied.

I changed the implementation, by copying the scan opt code. For forcing `device=cpu` on GPU, this is done via adding custom compilation mode. However, still not covering `device=gpu` when global config has `device=cpu`. I'm not sure how to cover that corner case, as it requires "just-in-time" loading gpuarray.

Issue #5472, opened 2017-01-31
Finish Gpu Cholesky decomposition op


lamblin:
Based on #4006.
- It should be based on cuSolver
- In the new back-end only
- with lifter optimizations

nouiz:
The good commit from that PR are at the end. The other one are from a bad
rebase/merge. Ignore them.

tfjgeorge:
I am working on this, but just a remark: there is a Cholesky version for the GpuSolve op in the new backend that deals with symmetric matrices. Do we really need a standalone Cholesky decomposition ? What would be a use case for just using a Cholesky decomposed matrix in Theano, without solving an equation of the form Ax = b ?

For reference, I put the direct link to the correct commits in the PR by matt graham: https://github.com/matt-graham/Theano/commits/cholesky_grad

matt-graham:
@tfjgeorge 

> I am working on this, but just a remark: there is a Cholesky version for the GpuSolve op in the new backend that deals with symmetric matrices. Do we really need a standalone Cholesky decomposition ? What would be a use case for just using a Cholesky decomposed matrix in Theano, without solving an equation of the form Ax = b ?

The advantage of having a standalone Cholesky decomposition op is that often you may want to cache the Cholesky decomposition of a matrix to do multiple solve operations. It also allows easy calculation of the determinant of the matrix the decomposition was calculated of by taking the product of the diagonal entries. 

Both of these come up when for example working with the probability density function of a Gaussian where we will usually want to pre-compute the Cholesky of the D×D covariance matrix as a one-off O(D³) cost to allow for evaluations of the (log) density at a O(D²) cost.

Did you manage to make any progress with this? I have some time over the next couple of weeks to help on this if it's any use.

nouiz:
done

PR #5799, opened 2017-03-31
Gpu cholesky

tfjgeorge:
this solves #5472 @notoraptor


theano/gpuarray/tests/test_linalg.py
```diff
@@ -112,3 +112,105 @@ def test_linalgerr_solve(self):
...
+        # A = M.dot(M) will be positive definite for all non-singular M
+        A_val = M_val.dot(M_val.T)
+        self.compare_gpu_cholesky_to_numpy(A_val, lower=False, inplace=False)
+
+    def test_diag_chol_inplace(self):
+        """ Diagonal matrix input inplace Cholesky test. """
+        # make sure all diagonal elements are positive so positive-definite
+        A_val = numpy.diag(numpy.random.uniform(size=5).astype("float32") + 1)
+        self.compare_gpu_cholesky_to_numpy(A_val, lower=True, inplace=True)
+
```
notoraptor: It miss some `test_diag_chol_*()` for `lower=False, inplace=True` and `lower=False, inplace=False` !

Also, consider factorizing the code for valid test cases (diagonal matrices and dense matrices). Maybe you could write one function for each type of matrix (e.g. `run_test_diag(inplace, lower)` and `run_test_dense(inplace, lower)`), and then use it in every combination of lower and inplace.

notoraptor:
NB: I test with this code:

```python
import numpy as np
import theano
import pygpu
from theano.gpuarray.linalg import GpuCholesky
from theano.gpuarray.type import get_context
from theano.gpuarray.tests.config import mode_with_gpu, test_ctx_name

lower = True
inplace = True
A = theano.gpuarray.GpuArrayType(name='A', dtype='float32', broadcastable=[False, False])()
chol_A = GpuCholesky(lower=lower, inplace=inplace)(A)
f = theano.function([A], chol_A, mode=mode_with_gpu, accept_inplace=inplace)

M_val = np.absolute(np.random.normal(size=(3, 3)).astype("float32"))
A_val = M_val.dot(M_val.T)
A_gpu = pygpu.asarray(A_val, context=get_context(test_ctx_name))

res = f(A_gpu)

ref = np.linalg.cholesky(A_val)
if not lower:
    ref = ref.T

assert np.allclose(ref, np.array(res))
theano.printing.debugprint(f)
print('ID input: ', id(A_gpu))
print('ID output:', id(res))
print('[OK]')
```

And I got this output:
```
Using cuDNN version 5105 on context None
DeepCopyOp [id A] ''   2
 |GpuCholesky{lower=True, inplace=True} [id B] ''   1
   |GpuContiguous [id C] ''   0
     |<A> [id D]
('ID input: ', 139637045185232)
('ID output:', 139637045185744)
[OK]
```

So, there is a DeepCopy inserted. What do you think, @abergeron ?

abergeron:
It is working inplace (as indicated by `inplace=True`).

The DeepCopyOp is there so that Theano does not return internal memory.  This is normal.

abergeron:
Oh I see, what you are trying to do.  It's very hard (and not recommended at all) to get a Theano function that actually works inplace on its inputs.  Inplace is reserved for the internal memory because we don't want to alias with the user memory.

See: http://deeplearning.net/software/theano/tutorial/aliasing.html

notoraptor:
Ok. So, should we remove the `inplace` option for `GpuCholesky` op ?

Or let it as is, and add an optimization that replaces CPU Cholesky with either GpuCholesky inplace or GpuCholesky not inplace ? (I had totally forgotten we usually did that)

abergeron:
It's still useful to have an inplace mode since for bigger graphs it will actually work inplace and help reduce the required working memory.

theano/gpuarray/tests/test_linalg.py
```diff
@@ -112,3 +112,83 @@ def test_linalgerr_solve(self):
...
+    def test_invalid_input_fail_tensor3(self):
+        # Invalid Cholesky input test with 3D tensor as input. 
+        def invalid_input_func():
+            A = theano.tensor.tensor3("A", dtype="float32")
+            GpuCholesky(lower=True, inplace=False)(A)
+        self.assertRaises(AssertionError, invalid_input_func)
+
+    def test_diag_chol(self):
+        # Diagonal matrix input Cholesky test. 
+        # make sure all diagonal elements are positive so positive-definite
```
notoraptor: Move this comment line inside the loop (juste before `A_val = ...`).

theano/gpuarray/opt.py
```diff
@@ -1967,6 +1967,16 @@ def local_gpu_solve(op, context_name, inputs, outputs):
...
 
+
+# Cholesky decomposition
+@register_opt('fast_compile')
+@op_lifter([slinalg.Cholesky])
+@register_opt2([theano.tensor.slinalg.Cholesky], 'fast_compile')
+def local_gpu_cholesky(op, context_name, inputs, outputs):
+    if not cusolver_available:
+        return
+    return GpuCholesky()
```
notoraptor: `slinalg.Cholesky` contains two attributes `lower` and `destructive` (which can be interpreted as `inplace`), so transfer them also into `GpuCholesky` here:
```
return GpuCholesky(lower=op.lower, inplace=op.destructive)
```

theano/gpuarray/tests/test_opt.py
```diff
@@ -584,6 +584,23 @@ def test_local_lift_solve():
     utt.assert_allclose(f_cpu(A_val, b_val), f_gpu(A_val, b_val))
 
 
+def test_local_lift_cholesky():
+    if not cusolver_available:
+        raise SkipTest('No cuSolver')
+    A = tensor.fmatrix()
+    o = slinalg.cholesky(A)
+    f_cpu = theano.function([A], o)
```
notoraptor: Make sure f_cpu is on CPU only, by adding `mode=mode_without_gpu`.

theano/gpuarray/linalg.py
```diff
@@ -5,15 +5,21 @@
 import warnings
 
 from theano import Op
+
 from theano.gpuarray import basic_ops, GpuArrayType
 
+from pygpu.gpuarray import GpuKernel, GpuArray
+from string import Template
+
```
notoraptor: Remove these 3 lines as GpuKernel, GpuArray and Template are not used anymore in this file (it currently makes flake8 fail).

theano/gpuarray/tests/test_linalg.py
```diff
@@ -112,3 +112,83 @@ def test_linalgerr_solve(self):
...
         self.assertRaises(LinAlgError, fn, A_val, x_val)
+
+
+class TestGpuCholesky(unittest.TestCase):
+
+    def setUp(self):
+        utt.seed_rng()
+
+    def get_gpu_cholesky_func(self, lower=True, inplace=False):
+        # Helper function to compile function from GPU Cholesky op.
```
notoraptor: There is a trailing white space at the end of the comment, that makes flake8 fail. Make sure there are no trailing spaces at the end of comments you have added (lines 123, 130, 140, 146, 153, 160, 168, 188 and 194, at least!).

theano/gpuarray/linalg.py
```diff
@@ -212,3 +219,122 @@ def perform(self, node, inputs, outputs):
...
+        # this op can only operate on float32 matrices
+        assert inp.ndim == 2
+        assert inp.dtype == 'float32'
+
+        return theano.Apply(
+            self, [inp],
+            [GpuArrayType('float32',
+                          broadcastable=inp.broadcastable,
+                          context_name=context_name)()])
```
nouiz: The creation of the output var can be simplied to 

inp.type()

as it is exactly the same type as the input variable.

theano/gpuarray/linalg.py
```diff
@@ -212,3 +219,122 @@ def perform(self, node, inputs, outputs):
...
+                               'Try updating libgpuarray?')
+        context_name = basic_ops.infer_context_name(inp)
+
+        inp = basic_ops.as_gpuarray_variable(inp, context_name)
+
+        inp = basic_ops.gpu_contiguous(inp)
+
+        # this op can only operate on float32 matrices
+        assert inp.ndim == 2
+        assert inp.dtype == 'float32'
```
nouiz: Where this limit come from? cusolver? triu/tril? Just not implemented? If just not implemented add it as a comment on that line to make clear. Otherwise, tell where the limitation come from.

theano/gpuarray/opt.py
```diff
@@ -1967,6 +1967,16 @@ def local_gpu_solve(op, context_name, inputs, outputs):
         return
     return GpuCusolverSolve()
 
+
+# Cholesky decomposition
+@register_opt('fast_compile')
+@op_lifter([slinalg.Cholesky])
+@register_opt2([theano.tensor.slinalg.Cholesky], 'fast_compile')
+def local_gpu_cholesky(op, context_name, inputs, outputs):
```
nouiz: Currently there is no opt that make this inplace. 

```
@register_inplace()
@local_optimizer([GpuCholesky], inplace=True)
def local_inplace_cholesky(node):
    if isinstance(node.op, GpuCholesky) and not node.op.inplace:
        return [GpuCholesky(node.op.lower, inplace=True)(*node.inputs)]
```
this also need tests.

theano/gpuarray/linalg.py
notoraptor: In fact, triu/tril currently works only with 32-bits floating values: https://github.com/Theano/libgpuarray/blob/master/pygpu/basic.py#L7

We could improve it by adding support for float64 values in triu/tril.

There may have some improvements here too for both GpuCholesky and GpuCusolverSolve: these ops currently use single-precision functions from cusolver (e.g.: `cusolverDnSpotrf_bufferSize`, `cusolverDnSpotrf`), so we could add double-precision support (cusolverDn**D**potrf_bufferSize, cusolverDn**D**potrf).

I suggest to add a TODO comment and then investigate improvements in another PR.

notoraptor:
As it is small changes (globally), I have already addressed it.

notoraptor:
I am OK with this PR. Maybe a final review needed ? @abergeron @nouiz .

PR #5806, opened 2017-04-04
make work Gammaln on the new gpu back-end

nouiz:
fix https://github.com/pymc-devs/pymc3/issues/1939

also add the offset information to the memory profile

### DIFF
--- a/theano/gpuarray/elemwise.py
+++ b/theano/gpuarray/elemwise.py
@@ -8,7 +8,7 @@ from six.moves import StringIO, xrange
 from theano.gof.utils import MethodNotDefined
 from theano.scalar import Scalar, Composite
 from theano.tensor.elemwise import (Elemwise, DimShuffle, CAReduceDtype)
-from theano.scalar.basic_scipy import Erfinv, Erfcinv, GammaLn
+from theano.scalar.basic_scipy import Erfinv, Erfcinv
 from theano.scalar.basic import upgrade_to_float_no_complex, complex_types
 
 try:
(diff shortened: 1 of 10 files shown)


theano/gpuarray/elemwise.py
```diff
@@ -2493,6 +2493,13 @@ def gpu_kernels(self, node, nodename):
         return kernels
 
 
+class GpuGammaLn(GammaLn):
+
+    def c_headers(self):
+        return ['math_functions.h']
```
abergeron: I don't think this helps anything, because the header will be for the op code, not the kernel.

Also, we can't add headers to kernels since nvrtc needs special parameters for them.

theano/gpuarray/opt.py
```diff
@@ -711,18 +711,16 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
...
+               Erfcinv: gpu_erfcinv,
+               GammaLn: gpu_gammaln}
+
+    if scal_op.__class__ in convert:
+        scal_op = convert[scal_op.__class__]
         if have_opencl:
-            _logger.warning('Function "%s" is not supported with OpenCL. Use "device=cuda" instead.' % opname)
+            _logger.warning(
+                'Function "%s" is not supported with OpenCL. Use "device=cuda" instead.' %
+                scal_op)
```
abergeron: lgamma is available and working on OpenCL, so we don't need this limit.

nouiz:
the travis part that faile due to sphinx passed, so I can merge.

PR #5808, opened 2017-04-04
replace as ChoiceFromUniform prop

laurent-dinh:
fix https://github.com/Theano/Theano/issues/5408

### DIFF
--- a/theano/gpuarray/multinomial.py
+++ b/theano/gpuarray/multinomial.py
@@ -242,11 +242,17 @@ class GPUAChoiceFromUniform(GpuKernelBase, Op):
 
     """
 
-    __props__ = ("odtype",)
+    __props__ = ("odtype", "replace")
 
-    def __init__(self, odtype):
+    def __init__(self, odtype, replace=False):
         Op.__init__(self)
         self.odtype = odtype
+        self.replace = replace
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        if "replace" not in state:
+            self.replace = False
 
     def get_params(self, node):
         return node.outputs[0].type.context
@@ -282,6 +288,7 @@ class GPUAChoiceFromUniform(GpuKernelBase, Op):
         return Apply(self, [pvals, unis, as_scalar(n)], [out])
 
     def gpu_kernels(self, node, name):
+        replace = int(self.replace)
         code = """
 KERNEL void k_multi_warp_multinomial_wor(
     const ga_size nb_multi,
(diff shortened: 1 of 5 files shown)


theano/sandbox/multinomial.py
```diff
@@ -219,6 +219,17 @@ class ChoiceFromUniform(MultinomialFromUniform):
...
 
+    __props__ = ("replace",)
+
+    def __init__(self, replace=False, *args, **kwargs):
+        self.replace = replace
+        super(ChoiceFromUniform, self).__init__(*args, **kwargs)
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        if "replace" not in state.keys():
```
nouiz: Can be simplified to "if "replace" not in state". This will be faster too.

theano/sandbox/multinomial.py
```diff
@@ -239,6 +250,7 @@ def c_code_cache_version(self):
     def c_code(self, node, name, ins, outs, sub):
         (pvals, unis, n) = ins
         (z,) = outs
+        replace = ("1" if self.replace else "0")
```
nouiz: as self.replace is a bool, you can simplify to "int(self.replace)" and use "%(replace)d" in the c code.

nouiz:
jenkins test this

theano/sandbox/multinomial.py
```diff
@@ -219,6 +219,17 @@ class ChoiceFromUniform(MultinomialFromUniform):
 
     """
 
+    __props__ = ("replace",)
+
+    def __init__(self, replace=False, *args, **kwargs):
```
nouiz: It don't have the same param order as the GPU version. I would go the same as the GPU version:

def __init__(self, odtype, replace=False):

then when you call super, you just pass odtype, the only one needed to pass.

theano/gpuarray/multinomial.py
```diff
@@ -318,23 +325,29 @@ def gpu_kernels(self, node, name):
...
             // No need to renormalize after the last samples.
             if (c == (n_samples - 1))
                 break;
-            // parallel renormalize the multinomial
-            for (ga_int k = LID_1; k < nb_outcomes; k+=LDIM_1)
+            if (! %(replace)s )
             {
-                global_pvals_copy[k * pvals_col_stride + n * pvals_row_stride] /= cummul;
+                // parallel renormalize the multinomial
+                for (ga_int k = LID_1; k < nb_outcomes; k+=LDIM_1)
```
nouiz: On line 473 change this line:

         if (nb_threads * nb_blocks < 2048)

to

         if (nb_threads * nb_blocks < 2048 && %(replace)d)

This will optimize the GPU usage as this part is skipped

laurent-dinh: Should it be `%(replace)d` or `! %(replace)d` ?

nouiz:
jenkins test this

nouiz:
thanks.

PR #5814, opened 2017-04-05
Remove the old backend and all of its tentacles

abergeron:


nouiz:
CI fail

theano/tensor/nnet/conv3d2d.py
```diff
@@ -198,8 +197,7 @@ def conv3d(signals, filters,
     Another way to define signals: (batch,  time, in channel, row, column)
     Another way to define filters: (out channel,time,in channel, row, column)
 
-    For the GPU, you can use this implementation or
-    :func:`conv3d_fft <theano.sandbox.cuda.fftconv.conv3d_fft>`.
+    For the GPU, use nnet.conv3d.
```
nouiz: Exact, but I read user cases where conv3d2d was faster then the GPU version as the cudnn conv2d was faster in some cases then conv3d.

Just to keep in mind that porting the missing ops to the new back-end would be useful.

theano/tensor/tests/test_opt.py
```diff
@@ -2,8 +2,6 @@
```
nouiz: Why do you remove the file theano/tensor/tests/shape_opt_cycle.pkl? You also removed another pickled file.

doc/index.txt
```diff
@@ -21,6 +21,9 @@ learning/machine learning <https://mila.umontreal.ca/en/cours/>`_ classes).
 News
 ====
 
+* Removed support for the old (device=gpu) backend.  Use the new
+  backend (device=cuda) for gpu computing.
+
```
nouiz: Tell that it is removed from the dev version and link to https://github.com/Theano/Theano/wiki/Converting-to-the-new-gpu-back-end%28gpuarray%29

doc/library/tensor/nnet/conv.txt
```diff
@@ -70,61 +70,27 @@ TODO: Give examples on how to use these things! They are pretty complicated.
...
+      i.e., it flips the kernel.
       Most of the more efficient GPU implementations listed below can be
       inserted automatically as a replacement for nnet.conv.conv2d via graph
       optimizations. Some of these graph optimizations are enabled by default,
       others can be enabled via Theano flags.
-      Since November 24th, 2014, you can also use a meta-optimizer to
-      automatically choose the fastest implementation for each specific
-      convolution in your graph using the old interface. For each instance,
-      it will compile and benchmark each applicable implementation of the ones
+      You can also use a meta-optimizer to automatically choose the
```
nouiz: The optimizer isn't yet ported... Could be done during GSoC.

theano/gpuarray/tests/test_rng_mrg.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, print_function, division
 import functools
 
-import numpy
+import numpy as np
```
nouiz: Don't mix that in other PR! Don't loose time to remove it, but don't do it again...

theano/gpuarray/tests/test_type.py
```diff
@@ -79,8 +79,9 @@ def test_unpickle_gpuarray_as_numpy_ndarray_flag0():
     finally:
         config.experimental.unpickle_gpu_on_cpu = oldflag
 
-
-test_shared_options = makeSharedTester(
+# These tests are disabled because they expect the impossible
```
nouiz: I don't agree with that. Can you disable just one or some part?

theano/sandbox/rng_mrg.py
```diff
@@ -1151,7 +887,7 @@ def updates(self):
         # TODO : need description for method and return
         return list(self.state_updates)
 
-    def __init__(self, seed=12345, use_cuda=None):
+    def __init__(self, seed=12345):
```
abergeron: I would prefer to get rid of it.  We will get an error anyway if something tries to use it.

theano/compile/profiling.py
```diff
@@ -1245,16 +1243,6 @@ def print_stats(stats1, stats2):
 
             print("---", file=file)
 
-        if (hasattr(theano, 'sandbox') and
```
abergeron: There is no way to do it currently, but it could be added.  See https://github.com/Theano/libgpuarray/issues/400

theano/tensor/nnet/conv3d2d.py
abergeron: Here nnet.conv3d is AbstractConv3d.  We can always add alternative implementations though.

theano/tensor/tests/test_opt.py
abergeron: That file contains pickled CudaNdArrayType.  So it doesn't work when you remove the old backend.

abergeron: The other one had a similar problem.

doc/index.txt
abergeron: This is the doc for the dev version.  We are not on the release branch.  I'll add the link.

doc/library/tensor/nnet/conv.txt
abergeron: I was sure it was.  I guess I should update this text to remove mentions of it?

theano/gpuarray/tests/test_type.py
```diff
@@ -94,8 +94,8 @@ def test_unpickle_gpuarray_as_numpy_ndarray_flag0():
     test_internal_type_=lambda a: isinstance(a, pygpu.gpuarray.GpuArray),
     theano_fct_=theano.tensor.exp,
     ref_fct_=np.exp,
-    cast_value_=lambda v: pygpu.asarray(v, context=get_context(test_ctx_name),
-                                        cls=pygpu._array.ndgpuarray),
+    cast_value_=lambda v: pygpu.array(v, context=get_context(test_ctx_name),
```
abergeron: It doesn't support the cls argument.

theano/gpuarray/tests/test_type.py
abergeron: The expect get_value() to return the internal type without passing return_internal_type=True.

They also expect that if you pass a ndarray and then call get_value() you get the same object.

There are a couple of options you can play with to tweak certain expectations, but those two seem hardcoded.

Overall this is not a reduction in testing since those tests were never done for gpuarray before.  This can motivate someone to actually make it work, but I don't have any more time to spend on this (and I already spent too much).

theano/gpuarray/fft.py
```diff
@@ -28,9 +28,9 @@
 try:
     import skcuda
     from skcuda import fft
-    scikits_cuda_available = True
+    skcuda_available = True
```
abergeron: If it is indeed public api, I'll change it back, but I really hope not.

abergeron:
I should have fixed the problem in python 2.

abergeron:
Last few commits should have addressed all the remaining failures.

theano/sandbox/rng_mrg.py
nouiz: ok

theano/tensor/tests/test_opt.py
nouiz: k

doc/index.txt
nouiz: It would make porting the doc harder. What about making it clear on the line that this is in the dev version so we can keep the same file?

doc/library/tensor/nnet/conv.txt
nouiz: I would just update the doc that we have that and we plan to port it instead of removing? No strong feeling what is better, just a suggestion.

theano/gpuarray/tests/test_type.py
nouiz: Was this tested in the old back-end? I think you took that from there. So this is a regression compared to the old back-end.

I don't want to delay this PR for this, can you make an issue about this and tag it for 0.10 and CCW so it don't get delayd too much? Maybe Reyhane could do that too.

nouiz:
still failure in travis.

nouiz:
jenkins have a failure.

slefrancois:
jenkins test this

lamblin:
Jenkins test this again please

Issue #5804, opened 2017-04-03
Missing optimization: Remove needless flattening


f0k:
Theano currently does not identify and remove a needless flattening operation:
```python
import theano
import theano.tensor as T
x = T.matrix('x')
y = x.flatten(2)  # useless reshape
theano.printing.debugprint(theano.function([x], y))
```
Output:
```
DeepCopyOp [id A] ''   3
 |Reshape{2} [id B] ''   2
   |x [id C]
   |MakeVector{dtype='int64'} [id D] ''   1
     |Shape_i{0} [id E] ''   0
     | |x [id C]
     |TensorConstant{-1} [id F]
```
This needlessly computes `x.reshape((x.shape[0], -1))` although `x` only has two dimensions anyway.

----

This has a real impact when trying to implement spatial softmax with spatial cross-entropy. With the current 2D softmax and 2D categorical crossentropy implementations, a spatial softmax requires some dimshuffle and reshape operations before and after the softmax. This hinders optimization:
```python
import theano
import theano.tensor as T
x = T.tensor4('x')
t = T.itensor3('t')  # an integer target per image and pixel
# simulate a softmax across the channel axis:
sm = T.nnet.softmax(x.dimshuffle(0,2,3,1).flatten(2))
sm = sm.reshape(x.dimshuffle(0,2,3,1).shape).dimshuffle(0,3,1,2)
# compute categorical cross-entropy across the channel axis:
ce = T.nnet.categorical_crossentropy(sm.dimshuffle(0,2,3,1).flatten(2), t.flatten())
# compile and print optimized graph
theano.printing.debugprint(theano.function([x, t], ce))
```
Output:
```
CrossentropyCategorical1Hot [id A] ''   7
 |Reshape{2} [id B] ''   6
 | |Softmax [id C] ''   5
 | | |Reshape{2} [id D] ''   4
 | |   |InplaceDimShuffle{0,2,3,1} [id E] ''   2
 | |   | |x [id F]
 | |   |MakeVector{dtype='int64'} [id G] ''   3
 | |     |Shape_i{0} [id H] ''   1
 | |     | |x [id F]
 | |     |TensorConstant{-1} [id I]
 | |MakeVector{dtype='int64'} [id G] ''   3
 |Reshape{1} [id J] ''   0
   |t [id K]
   |TensorConstant{(1,) of -1} [id L]
```
[...]

PR #5815, opened 2017-04-05
Make opt useless_reshape handle -1

nouiz:
This optimize more cases.

fix https://github.com/Theano/Theano/issues/5804


theano/tensor/basic.py
```diff
@@ -4745,6 +4745,8 @@ class Reshape(Op):
 
     def __init__(self, ndim, name=None):
         self.ndim = ndim
+        if ndim < 0:
+             raise ValueError("The output dimensions after reshape must be 0 or greater")
```
f0k: Indented by 5 spaces, should be 4.

PR #5817, opened 2017-04-06
Add covariance matrix function theano.tensor.cov

ieee8023:
I added a simple function theano.tensor.cov using existing theano functions so that it may run on a GPU. This is is modeled after np.cov but matches only one of it's function signatures.

These guys helped me figure this out!
Olexa Bilaniuk @obilaniu
Zhouhan Lin @hantek

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -2169,9 +2169,26 @@ square = sqr
 
 
 def cov(a):
-    X -= X.mean(axis=1, keepdims=1)
-    c = X.dot(X.T)
-    return c/(X.shape[1]-1)
+    """Calculate the covariance matrix.
+    Covariance indicates the level to which two variables vary together.
+    If we examine N-dimensional samples, :math:`X = [x_1, x_2, ... x_N]^T`,
+    then the covariance matrix element :math:`C_{ij}` is the covariance of
+    :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
+    of :math:`x_i`.
+    ----------
+    a : array_like
+        A 2-D array containing multiple variables and observations.
+        Each row of `a` represents a variable, and each column is 
+        observations of all those variables. 
+        
+    Returns
+    -------
+    out : The covariance matrix of the variables.
+    """
+        
+    a -= a.mean(axis=1, keepdims=1)
+    c = a.dot(a.T)
+    return c/(a.shape[1]-1)
 
 @_scal_elemwise
 def sqrt(a):
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -2168,27 +2168,66 @@ def sqr(a):
 square = sqr
 
 
-def cov(a):
+def cov(X, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=None):
     """Calculate the covariance matrix.
     Covariance indicates the level to which two variables vary together.
     If we examine N-dimensional samples, :math:`X = [x_1, x_2, ... x_N]^T`,
     then the covariance matrix element :math:`C_{ij}` is the covariance of
     :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
-    of :math:`x_i`.
+    of :math:`x_i`. Code and docstring ported from numpy.
     ----------
     a : array_like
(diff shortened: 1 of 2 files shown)


JesseLivezey:
It's probably be a good idea to make the interface more similar to the ```numpy.cov``` function
https://docs.scipy.org/doc/numpy/reference/generated/numpy.cov.html

```
cov(m, rowvar=True, bias=False, ddof=None)
```
would probably be a reasonable minimal interface.

ieee8023:
@JesseLivezey I agree it will be more useful to have this functionality. I can get to this next week! Thanks for the feedback.

lamblin:
Thanks! I agree that mimicking the full functionality of NumPy would be better.
Also, the automatic flake8 test currently [fails with these errors](https://travis-ci.org/Theano/Theano/jobs/219138467#L893), which would probably be easy to fix.

JesseLivezey:
The full ```numpy.cov``` signature is a little more complex than what I posted above
```
numpy.cov(m, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=None)
```
but they list the full computation on this page
https://docs.scipy.org/doc/numpy/reference/generated/numpy.cov.html
and you can look at the numpy code here
https://github.com/numpy/numpy/blob/561f1accf861ad8606ea2dd723d2be2b09a2dffa/numpy/lib/function_base.py#L2725
so it might not be that much more work to write something a little more general.

nouiz:
Just a comment. There is no problem to implement only a subset of np.cov, but having everything is better as other said :)

But if you don't implement everything, make sure to have a fct signature compatible with np.cov. For example, if you don't implement y, you still need to keep it in the signature and raise not implemented if someone try to use it. As if we add support for it later, we don't want to break the old interface by changing the order of parameter.

thanks for the contribution.

ieee8023:
I implemented all the arguments mentioned initially by @JesseLivezey. The weight arguments were a bit too complicated so I made the function raise a NotImplementedError if they are set as @nouiz said.

I added a bunch of tests. If I missed a use case let me know and I can work more on it.

theano/tensor/basic.py
```diff
@@ -2168,6 +2168,67 @@ def sqr(a):
...
 
+def cov(X, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=None):
+    """Calculate the covariance matrix.
+    Covariance indicates the level to which two variables vary together.
+    If we examine N-dimensional samples, :math:`X = [x_1, x_2, ... x_N]^T`,
+    then the covariance matrix element :math:`C_{ij}` is the covariance of
+    :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
+    of :math:`x_i`. Code and docstring ported from numpy.
+    ----------
+    a : array_like
```
JesseLivezey: I would change ```X``` above and ```a``` here to ```m``` to be consistent with numpy.cov.

ieee8023: Maybe it should all be `a` to match the other theano code?

theano/tensor/tests/test_basic.py
```diff
@@ -8172,6 +8172,107 @@ def test_norm():
...
 
+class test_cov(unittest.TestCase):
+
+    def test_core(self):
+        x = theano.tensor.matrix('x')
+        c = theano.tensor.cov(x)
+        f = theano.function([x], c)
+
+        # basic cov function
+        data = np.asarray(np.random.rand(3, 5), dtype=config.floatX)
```
JesseLivezey: This series of test could be written more concisely as a loop over shapes. Same for other tests.

theano/tensor/basic.py
abergeron: We try to mimic the numpy API in general so going with what numpy does is not a bad idea.  In any case, this absolutely needs to have the same name in the function definition and the doc.

nouiz:
We try to make Theano 0.10. It would be great that this is finished and included. We don't have time to finish it ourself, so if someone have time to finish it this week, it would be great!

JesseLivezey:
@ieee8023 I think the only thing you need to do is change the variable names in the code and documentation to be consistent with numpy.

ieee8023:
Ok, I can get this done before the end of the week

PR #5819, opened 2017-04-06
add GLOBAL_MEM

fo40225:
1. add GLOBAL_MEM

2. remove static_cast<> from OpenCL kernel code


nouiz:
jenkins test this.

I'm good with this PR, but I would prefer the review of @abergeron to confirm the static_cast stuff.

Issue #5776, opened 2017-03-27
libgpu backend with CuDNN won't build (Windows)


jtc42:
Fresh install of Anaconda with Python 3.6, on Windows 10 64-bit. libpython and m2w64-toolchain have both been installed using conda install.

Upon importing Theano using the new libgpu backend, I get the following error:

```
Can not use cuDNN on context None: cannot compile with cuDNN. We got this error:

In file included from C:\\CUDA\\v8.0\\include/driver_types.h:53:0,      
from C:\\CUDA\\v8.0\\include/cudnn.h:63,
from C:\\Users\\jtc92\\AppData\\Local\\Temp\\try_flags_733_4ik6.c:4:
C:\\CUDA\\v8.0\\include/host_defines.h:84:0: warning: "__cdecl" redefined
#define __cdecl <built-in>: 
note: this is the location of the previous definition

C:/Anaconda3/Library/mingw-w64/bin/../lib/gcc/x86_64-w64-mingw32/5.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: 
cannot find -lcudnn\r\ncollect2.exe: error: ld returned 1 exit status

Mapped name None to device cuda: GeForce GTX 1060 3GB (0000:01:00.0)
```

My .theanorc is set up as:
```
[global]
floatX = float32
device = cuda

compiler_bindir=C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin

[cuda]
root = C:\CUDA\v8.0

[dnn]
library_path = C:\CUDA\v8.0\lib
include_path = C:\CUDA\v8.0\include
```
CuDNN is installed to the default CUDA directory, as above.

Any advice would be great. Weirdly, using the old CUDA backend works perfectly, but this is obviously not an ideal solution.

Thanks!

jc-Rosichini:
[dnn]
library_path = C:\CUDA\v8.0\lib\**x64**

jtc42:
Well, It works perfectly now. I can't believe I missed that. Thank you so much!

cehkongfu:
dear jtc42:
     could you tell me how did you dissove the question about "In file included from C:\\CUDA\\v8.0\\include/driver_types.h:53:0,      
from C:\\CUDA\\v8.0\\include/cudnn.h:63,
from C:\\Users\\jtc92\\AppData\\Local\\Temp\\try_flags_733_4ik6.c:4:
[...]

PR #5821, opened 2017-04-07
change default dnn path on windows

fo40225:
change default dnn path on windows

include path: CUDA_ROOT\include

library path: CUDA_ROOT\lib\x64

#5776

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -323,23 +323,32 @@ def get_cuda_root():
     return ''
 
 
-def default_dnn_path(suffix):
-    def f(suffix=suffix):
-        cuda_root = get_cuda_root()
-        if cuda_root == '':
-            return ''
-        return os.path.join(cuda_root, suffix)
-    return f
+def default_dnn_include_path():
+    cuda_root = get_cuda_root()
+    if cuda_root == '':
+        return ''
+    return os.path.join(cuda_root, 'include')
+
+
+def default_dnn_library_path():
+    cuda_root = get_cuda_root()
+    if cuda_root == '':
+        return ''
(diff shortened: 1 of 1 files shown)


theano/configdefaults.py
```diff
@@ -390,23 +390,29 @@ def safe_no_dnn_algo_bwd(algo):
...
-    def f(suffix=suffix):
-        if theano.config.cuda.root == '':
-            return ''
-        return os.path.join(theano.config.cuda.root, suffix)
-    return f
+def default_dnn_library_path():
+    if theano.config.cuda.root == '':
+        return ''
+    if sys.platform == 'darwin':
+        return os.path.join(theano.config.cuda.root, 'lib64')
```
nouiz: The old code was the reverse. on darwin, it is lib and by default, it is lib643

nouiz:
@notoraptor @abergeron what do you think of that?

notoraptor:
It looks good for Windows.

abergeron:
I'm ok with the change, but it will conflict with the PR to remove the old backend.  Can we delay this until then?

theano/configdefaults.py
fo40225: Do you mean "darwin ->lib, linux ->lib64"?

nouiz:
Can you rebase this PR now that the other PR got merged?

thanks

nouiz:
travis have a few failures. Can you fix them? See

https://travis-ci.org/Theano/Theano/jobs/221345146#L873

nouiz:
jenkins test this

nouiz:
thanks

PR #5834, opened 2017-04-11
Use expm1 for elu

f0k:
Just a minor optimization that @danstowell found for Lasagne: Replace `tensor.exp(x) - 1` by `tensor.expm1(x)` in the elu definition so Theano doesn't have to. (Theano has an optimization that does the same.)


nouiz:
thanks

Issue #5816, opened 2017-04-05
BNComposite has no attribute _c_code


mpslxz:
After upgrading to v0.9 from rc, I cannot load pickles if the model has batch normalization calls. 

> theano/scalar/basic.py", line 4032, in __setstate__
    assert self._c_code
AttributeError: 'BNComposite' object has no attribute '_c_code'

lamblin:
I'm pretty sure that line should be removed, now we do not always generate the C code (5930ecd4).
Could you try it, and submit a PR?

mpslxz:
That is right. C code is not always generated. I'm not sure where c_code() function is called, though. Would you elaborate on what you'd like me to try for PR, please?

nouiz:
Can you try this diff:
```
index 9fe81b1..628d263 100644
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -4025,7 +4025,6 @@ class Composite(ScalarOp):
         self.prepare_node_called = set()
         self.init_fgraph()
         self.init_py_impls()
-        assert self._c_code
 
 
 class Compositef32(object):
```

thanks

nouiz:
I have this diff in a branch if you prefer: https://github.com/nouiz/Theano/tree/unsigned_char

thanks

mpslxz:
Sure. I'll do it.

Issue #5803, opened 2017-04-02
error compiling gof (lazylinker_ext.pyd missing)?


dattienhoang:
Hello,

I'm on a 64-bit Windows 10 machine using Python 2.7 (Anaconda2). I don't receive any errors when installing Theano with conda, but when I try to import in Python, I get the following message:

```
===============================
g++.exe: error: Tien: No such file or directory
g++.exe: error: Hoang\AppData\Local\Theano\compiledir_Windows-10-10.0.14393-Intel64_Family_6_Model_78_Stepping_3_GenuineIntel-2.7.13-64\lazylinker_ext\lazylinker_ext.pyd: No such file or directory
g++.exe: error: C:\Users\Dat: No such file or directory
g++.exe: error: Tien: No such file or directory
g++.exe: error: Hoang\AppData\Local\Theano\compiledir_Windows-10-10.0.14393-Intel64_Family_6_Model_78_Stepping_3_GenuineIntel-2.7.13-64\lazylinker_ext\mod.cpp: No such file or directory


```

At first, I thought it was maybe because I have a space in my username, but saw a previous issue had fixed this error already [link](https://github.com/Theano/Theano/issues/5183). This compilation error happens regardless of whether I use the standard install by conda, or following the bleeding-edge installation.

From the errors, it looks like I'm missing a couple files (lazylinker_ext.pyd and mod.cpp), which are not present when I navigate to the specified folder in the error using Windows Explorer. Is there a way to get these files?

What's crazy is that I have previously installed Theano on a different computer successfully using the normal installation method using conda, and can't figure why I'm having problems on a very similar computer!

Any help appreciated,
Dat

nouiz:
The error tell me the problem is still related to spaces in path.

Can you can in pdb at the place of the error and investigate a little bit?
For do one reason the previous fix didn't worked in all cases or this git
broken again. Check how we build the command we execute to compile.

Thanks for your help.

Le dim. 2 avr. 2017 16:13, Dat Tien Hoang <notifications@github.com> a
écrit :

dattienhoang:
hopefully this is what you wanted (not a python expert):

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Anaconda2\lib\site-packages\theano\__init__.py", line 88, in <module>
    from theano.compile import (
  File "C:\Anaconda2\lib\site-packages\theano\compile\__init__.py", line 10, in <module>
    from theano.compile.function_module import *
[...]

PR #5842, opened 2017-04-12
Unsigned char:  fix: theano/scalar/basic.py

mpslxz:
Crash issue with theano/scalar/basic.py [gh-5816](https://github.com/Theano/Theano/issues/5816#issuecomment-293435358)
The diff does not crash when unpickling the batch normalization op.
Confirming what Pascal @lamblin suggested, it should be removed.

fix https://github.com/Theano/Theano/issues/5803
fix gh-5816
fix https://github.com/Theano/libgpuarray/issues/403

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -319,7 +319,9 @@ class GpuKernelBase(object):
     def _generate_kernel_code(self, k):
         code = '\\n'.join(l for l in k.code.split('\n'))
         code = code.replace('"', '\\"')
-        return ("""static const char *%(cname)s = "%(code)s";""" %
+        return ("""static const char *%(cname)s_unsigned = "%(code)s";
+                static const char *%(cname)s = (char *)%(cname)s_unsigned;
+                """ %
                 dict(cname=k.codevar, code=code))
 
     def _generate_kernel_vars(self, k):
--- a/theano/gpuarray/tests/test_elemwise.py
+++ b/theano/gpuarray/tests/test_elemwise.py
@@ -13,7 +13,7 @@ from .config import mode_with_gpu, mode_without_gpu, test_ctx_name
 from .test_basic_ops import rand_gpuarray
 from ..elemwise import (GpuElemwise, GpuDimShuffle,
                         GpuCAReduceCuda, GpuCAReduceCPY, GpuErfinv, GpuErfcinv)
-from ..type import GpuArrayType, get_context
+from ..type import GpuArrayType, get_context, gpuarray_shared_constructor
 
 from pygpu import ndgpuarray as gpuarray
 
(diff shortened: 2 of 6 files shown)


nouiz:
jenkins test this

lamblin:
I killed Jenkins so ​the tests for #5735 can go first.
We should close / reopen to force a new merge afterwards, when the speed issue with Jenkins is fixed.

nouiz:
jenkins test this

theano/gpuarray/tests/test_elemwise.py
```diff
@@ -40,16 +40,21 @@ def test_elemwise_pow():
...
-            base = theano.tensor.vector(dtype=dtype_base)
-            exp = theano.tensor.vector(dtype=dtype_exp)
-            output = base ** exp
-            f = theano.function([base, exp], output)
-
             base_val = np.random.randint(0, 5, size=10).astype(dtype_base)
             exp_val = np.random.randint(0, 3, size=10).astype(dtype_exp)
 
+            base = theano.tensor.vector(dtype=dtype_base)
+            exp = gpuarray_shared_constructor(exp_val)
```
lamblin: I would add a check here to make sure that `exp.dtype == dtype_exp`, to make sure we are actually testing the right thing.

theano/gpuarray/tests/test_elemwise.py
```diff
@@ -40,16 +40,21 @@ def test_elemwise_pow():
...
+            output = base ** exp
+            f = theano.function([base], output, mode=mode_with_gpu)
+            theano.printing.debugprint(f)
+            # We don't transfer to the GPU when the output dtype is int*
+            n = len([n for n in f.maker.fgraph.apply_nodes
+                     if isinstance(n.op, GpuElemwise)])
+            assert n == int("float" in output.dtype)
```
lamblin: `output.dtype in float_dtypes` would be better IMO.

theano/gpuarray/basic_ops.py
```diff
@@ -319,7 +319,9 @@ def c_header_dirs(self):
     def _generate_kernel_code(self, k):
         code = '\\n'.join(l for l in k.code.split('\n'))
         code = code.replace('"', '\\"')
-        return ("""static const char *%(cname)s = "%(code)s";""" %
+        return ("""static const char *%(cname)s_unsigned = "%(code)s";
+                static const char *%(cname)s = (char *)%(cname)s_unsigned;
```
nouiz: This fix a compilation error with newer g++ that are more strict and don't like that some of the values are more then 128. So this is still needed. the newer g++ is probably still not supported everywhere.

lamblin: I understand.
My point is that `%(code)s` here is unlikely to have values > 128, but `%(bcode)s` that was defined in `_generate_kernel_bin` definitely had them (but it was removed in #5847).
Since it could still happen in `%(code)s`, I'm OK with the change.

lamblin:
Jenkins test this please

PR #5847, opened 2017-04-13
Remove call to GpuKernel_init with binary

lamblin:
As it is trying over and over to compile it as source.
Also, start removing generation of binaries from Theano.


lamblin:
Jenkins test this with a full cache, please.

lamblin:
Jenkins test this with a full cache, please.

nouiz:
jenkins gpu took 47m and passed. So I merge and the slower compilation problem is fixed!

good finding.

PR #5850, opened 2017-04-14
Disable compute_test_value during Theano optimization phase. This giv…

nouiz:
…e a better error in pymc3 issue 1939

https://github.com/pymc-devs/pymc3/issues/1939

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -1800,12 +1800,12 @@ def orig_function(inputs, outputs, mode=None, accept_inplace=False,
     try:
         Maker = getattr(mode, 'function_maker', FunctionMaker)
         m = Maker(inputs,
-                       outputs,
-                       mode,
-                       accept_inplace=accept_inplace,
-                       profile=profile,
-                       on_unused_input=on_unused_input,
-                       output_keys=output_keys)
+                  outputs,
+                  mode,
+                  accept_inplace=accept_inplace,
+                  profile=profile,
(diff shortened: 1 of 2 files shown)


theano/compile/function_module.py
```diff
@@ -1475,7 +1475,8 @@ def __init__(self, inputs, outputs,
                     optimizer_profile = self.optimize_graph_with_cache(
                         optimizer, inputs, outputs)
                 else:
-                    optimizer_profile = optimizer(fgraph)
+                    with theano.configparser.change_flags(compute_test_value="off"):
```
lamblin: There is already the option `theano.config.compute_test_value_opt` that governs this behaviour, and its default is `"off"`. Why override an option here that has been explicitly set?

nouiz:
I needed to also disable it during the linker. I forgot the details. Currently it is not working in some cases. Personally, I would completly disable test value during optimization and linker and don't give that option. I don't want to spend time supporting this. It isn't used.

Others options that I don't like  would be to make a new flag to disable it during linker or reuse the current flag to also disable it during linker, or make it work. I don't think it is worthwhile.

lamblin:
Having the option for test values during optimization can be an alternative to DebugMode in some cases. I don't think removing the option would be worth it.
But I'm OK with always setting it to "off" for the linker.

nouiz:
If we want to do it only for the linker, I'll update this PR to do that.

nouiz:
Updated to only disable it during linker

Issue #5572, opened 2017-02-17
Split Elemwise Add/Mul that have too many inputs as in the old back-end


nouiz:
In case the user build them or other opt not yet found could do it.

https://github.com/Theano/Theano/pull/5559 only prevent the creation

This should be done in the lifter of Elemwise. The code to split is already in gpuarray/opt.py in the function split_huge_add_or_mul

nouiz:
done by https://github.com/Theano/Theano/pull/5852

PR #5852, opened 2017-04-14
Split elemwise addmul

Faruk-Ahmed:
fix for #5572

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -760,8 +760,27 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
 
 
 def split_inputs(inputs, max_nb_inputs, op):
+    """
+    For some ops like add and mul, a large number of inputs can make nvcc fail
+    compilation of our current code. We don't want node in the graph that can't
+    execute as this break DebugMode.      
+        
+    This should not happen for other GpuElemwise as their is only the fusion
+    that can generate op with too much input and it check for that.
+
+    Parameters
+    ----------
+    inputs: List of inputs to node.
+
+    max_nb_inputs: Maximum nummber of inputs the node can handle without
+                   compilation fail.
+    
+    op : Theano operator instance.
+         Operator that should be used to rebuild the computation graph with smaller
+         number of inputs per node.
+    """
     if max_nb_inputs <= 1 and len(inputs) > 1:
-        ValueError("Can not split nodes because inputs' dimensionality and/or \
+        raise ValueError("Can not split nodes because inputs' dimensionality and/or \
                     number of outputs is too large")
 
     while len(inputs) > max_nb_inputs:
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -770,11 +770,11 @@ def split_inputs(inputs, max_nb_inputs, op):
 
     Parameters
     ----------
-    inputs: List of inputs to node.
-
-    max_nb_inputs: Maximum nummber of inputs the node can handle without
+    inputs: List of theano variables.
+            List of inputs to node.
+    max_nb_inputs: int
+                   Maximum nummber of inputs the node can handle without
                    compilation fail.
-
     op : Theano operator instance.
          Operator that should be used to rebuild the computation graph with smaller
          number of inputs per node.
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -763,8 +763,8 @@ def split_inputs(inputs, max_nb_inputs, op):
     """
     For some ops like add and mul, a large number of inputs can make nvcc fail
     compilation of our current code. We don't want node in the graph that can't
-    execute as this break DebugMode.      
-        
+    execute as this break DebugMode.
+
     This should not happen for other GpuElemwise as their is only the fusion
     that can generate op with too much input and it check for that.
 
@@ -774,7 +774,7 @@ def split_inputs(inputs, max_nb_inputs, op):
 
     max_nb_inputs: Maximum nummber of inputs the node can handle without
                    compilation fail.
-    
+
     op : Theano operator instance.
          Operator that should be used to rebuild the computation graph with smaller
          number of inputs per node.
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -773,15 +773,15 @@ def split_inputs(inputs, max_nb_inputs, op):
     inputs: List of theano variables.
             List of inputs to node.
     max_nb_inputs: int
-                   Maximum nummber of inputs the node can handle without
+                   Maximum number of inputs the node can handle without
                    compilation fail.
     op : Theano operator instance.
          Operator that should be used to rebuild the computation graph with smaller
          number of inputs per node.
     """
     if max_nb_inputs <= 1 and len(inputs) > 1:
-        raise ValueError("Can not split nodes because inputs' dimensionality and/or \
-                    number of outputs is too large")
+        raise ValueError("Can not split nodes because inputs' dimensionality and/or"
+                         " number of outputs is too large")
 
     while len(inputs) > max_nb_inputs:
         inner_ops = []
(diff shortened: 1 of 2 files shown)


theano/gpuarray/opt.py
```diff
@@ -771,16 +767,19 @@ def split_huge_add_or_mul(node):
     that can generate op with too much input and it check for that.
 
     """
+    if op is None:
+        op = node.op
     if node.op.scalar_op in (scalar.add, scalar.mul):
         max_nb_inputs = max_inputs_to_GpuElemwise(node)
         if max_nb_inputs <= 1 and len(node.inputs) > 1:
             return False
```
nouiz: I think this should return op(inputs).owner (the new parameter suggested in the other comment)

theano/gpuarray/opt.py
```diff
@@ -751,16 +751,12 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
         gpu_output = res(*new_inputs)
         return [gpu_output]
     elif op.scalar_op in (scalar.add, scalar.mul):
-        max_nb_inputs = max_inputs_to_GpuElemwise(outputs)
-        if max_nb_inputs > 1:
-            while len(inputs) > max_nb_inputs:
-                inputs = inputs[:-max_nb_inputs] + [res(*inputs[-max_nb_inputs:])]
-        return res(*inputs)
+        return split_huge_add_or_mul(outputs[0].owner, res).outputs
```
nouiz: This is an invalid change. This will use node.inputs instead of inputs (as in the old code).

I think you can fix that by addint a new parameter inputs=None to split_huge_add_or_mul and if None, use node.inputs. Like you do with op.

theano/gpuarray/tests/test_opt.py
```diff
@@ -467,30 +468,20 @@ def test_many_arg_elemwise():
...
-                    if mode is mode_with_gpu:
-                        assert any([isinstance(node.op, GpuElemwise)
-                                    for node in f.maker.fgraph.apply_nodes])
 
-                    # test the optijmization local_gpu_elemwise_1
-                    f = theano.function(
-                        symb_args,
-                        GpuFromHost(test_ctx_name)(op_to_test(*symb_args)),
-                        mode=mode.excluding("local_gpu_elemwise_0"))
-                    out = f(*args)
```
nouiz: Why all those changes to the test? It lower the test quality. You don't call the compiled function anymore. You don't test each optimization individually as before.

theano/gpuarray/tests/test_opt.py
```diff
@@ -450,14 +451,15 @@ def test_local_gpu_elemwise():
...
 
-    for num_args in [75]:
+    rng = np.random.RandomState([1, 2, 3])
+    nb_of_inputs_overflows = []
+    for num_args in [64]:
         for op_to_test in [theano.tensor.add, theano.tensor.mul]:
-            for nb_dim in [2, 3, 4, 5, 7]:
-                shapes = [rng.randint(1, 5) for i in range(nb_dim)]
+            for nb_dim in [2, 8]:
+                shapes = [rng.randint(1, int(32 / nb_dim)) for i in range(nb_dim)]
```
bouthilx: You should leave this to 5

theano/gpuarray/opt.py
nouiz: This is how we handle the optimizaton when we move to the GPU in the new back-end. This make the optimization faster. We move the full model at once instead of each node at a time. So much less intermediate work that don't stay in the final graph.

outputs.owner.inputs is the original (on the CPU inputs). Inputs is the new inputs we want to use (probably moved to the GPU). This is why we didn't give node, we don't want node.inputs to be used.

theano/gpuarray/opt.py
nouiz: you use this line somewhere: split_huge_add_or_mul(outputs[0].owner, res).outputs

So here, you should not return False. Of you must modify the place that call split_huge_add_or_mul() to don't always call .outputs.

theano/gpuarray/tests/test_opt.py
nouiz: right, the new back-end have only 1 lifter, so only 1 test is still needed.

theano/gpuarray/opt.py
bouthilx: Yes that's what I meant with the reference to line 754 and the code snippet.

theano/gpuarray/opt.py
bouthilx: I see, thanks for the clarification!

Let's decouple a few things then. I propose this for the optimization.
```python
try:
    return [split_inputs(inputs, max_inputs_to_GpuElemwise(outputs), res)]
except ValueError:
    return False
```

And a bit more general function to split the input.
```python
def split_inputs(inputs, max_nb_inputs, op):
    if max_nb_inputs <= 1 and len(node.inputs) > 1:
        ValueError("Can not split nodes because inputs' dimensionality and/or number of outputs is too large")

    while len(inputs) > max_nb_inputs:
        inner_ops = []
        for i in range(0, len(node.inputs), max_nb_inputs):
            inner_ops.append(op(*inputs[i: i + max_nb_inputs]))
        inputs = inner_ops

    return op(*inputs)
```

The try-except clause combined with raise ValueError will adress this [comment](#r112282662).

theano/gpuarray/opt.py
```diff
@@ -751,38 +751,27 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
...
         return res
 
 
-def split_huge_add_or_mul(node, op=None):
-    """
-    For add and mul, it can happen that we have too much input
-    That will make nvcc fail compilation of our current code.
-    We don't want node in the graph that can't execute
-    as this break DebugMode.
+def split_inputs(inputs, max_nb_inputs, op):
```
bouthilx: Do not drop the docstring, adapt it and add definitions of parameters, namely for ```op```
```
Parameters
----------
op : Theano operator instance
     Operator that should be used to rebuild the computation graph with smaller number of inputs per node.
```

theano/gpuarray/opt.py
```diff
@@ -751,36 +751,46 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
...
-            return False
-        while len(node.inputs) > max_nb_inputs:
-            inner_op = []
-            for i in range(0, len(node.inputs), max_nb_inputs):
-                inner_op.append(node.op(*node.inputs[i: i + max_nb_inputs]))
-            node = node.op(*inner_op).owner
-    return node
+    if max_nb_inputs <= 1 and len(inputs) > 1:
+        raise ValueError("Can not split nodes because inputs' dimensionality and/or \
+                    number of outputs is too large")
```
bouthilx: I'm surprised this is pep8 compliant. 

```python
raise ValueError("Can not split nodes because inputs' dimensionality and/or"
                 "number of outputs is too large")
```

Seems better to me.

theano/gpuarray/opt.py
```diff
@@ -751,36 +751,46 @@ def local_gpua_elemwise(op, context_name, inputs, outputs):
...
 
     This should not happen for other GpuElemwise as their is only the fusion
     that can generate op with too much input and it check for that.
 
+    Parameters
+    ----------
+    inputs: List of theano variables.
+            List of inputs to node.
+    max_nb_inputs: int
+                   Maximum nummber of inputs the node can handle without
```
bouthilx: *nummber

bouthilx:
Looks fine to me. I'll run the GPU tests on my side before merging. @nouiz Do you have any other requests?

bouthilx:
Jenkins test this please

slefrancois:
jenkins test this

theano/gpuarray/tests/test_opt.py
```diff
@@ -467,32 +469,30 @@ def test_many_arg_elemwise():
 
                 outputs = []
                 for mode in [mode_with_gpu, mode_without_gpu]:
-                    # test the optijmization local_gpu_elemwise_0
-                    f = theano.function(
-                        symb_args, op_to_test(*symb_args),
-                        mode=mode.excluding("local_gpu_elemwise_1"))
+                    # test the optimization local_gpua_elemwise
+                    output = op_to_test(*symb_args)
+                    f = theano.function(symb_args, output)
```
bouthilx: Oops, mode=mode should still be passed otherwise we can't force mode_with_gpu and mode_without_gpu.

PR #5854, opened 2017-04-15
Fix hard crash when dividing by 0

lamblin:
This should fix the crash (floating point exception, core dumped) in the buildbot in DebugMode.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -1928,14 +1928,16 @@ class IntDiv(BinaryScalarOp):
 
         t = node.inputs[0].type.upcast(*[i.type for i in node.inputs[1:]])
         if t in imap(str, discrete_types):
+            # If we are in a gpuarray kernel, %(fail)s exits the kernel,
+            # and we do not have any error report, and we cannot set
+            # Python error messages either, so for now we just call the
+            # cuda function, which return a binary pattern of all 1s.
             check = dedent('''
-                if (%(y)s == 0) {
-                // do not set Python error message in gpuarray kernel for now
-                #ifndef LID_0
+                #ifndef KERNEL
                     PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero");
-                #endif
                     %(fail)s
-                }''') % locals()
+                #endif
+                ''') % locals()
             x_div_y_pp = '(%(x)s / %(y)s)' % locals()
             x_div_y_mp = '((-%(x)s) / %(y)s)' % locals()
             x_mod_y_mp = 'THEANO_MACRO_MOD((-%(x)s), %(y)s)' % locals()
@@ -1967,16 +1969,18 @@ class IntDiv(BinaryScalarOp):
             raise NotImplementedError('type not supported', t)
 
         return dedent("""
-            %(check)s
-            if (%(x)s < 0) {
-                if (%(y)s < 0) {
+            if (%(y)s == 0) {
+                %(check)s
+                %(z)s = %(x_div_y_pp)s;
+            } else if (%(y)s < 0) {
+                if (%(x)s < 0) {
                     %(z)s = %(x_div_y_mm)s;
                 } else {
-                    %(z)s = - %(x_div_y_mp)s - ((%(x_mod_y_mp)s == 0) ? 0 : 1);
+                    %(z)s = - %(x_div_y_pm)s - ((%(x_mod_y_pm)s == 0) ? 0 : 1);
                 }
             } else {
-                if (%(y)s < 0) {
-                    %(z)s = - %(x_div_y_pm)s - ((%(x_mod_y_pm)s == 0) ? 0 : 1);
+                if (%(x)s < 0) {
+                    %(z)s = - %(x_div_y_mp)s - ((%(x_mod_y_mp)s == 0) ? 0 : 1);
                 } else {
                     %(z)s = %(x_div_y_pp)s;
                 }
@@ -1984,7 +1988,7 @@ class IntDiv(BinaryScalarOp):
             """) % locals()
 
     def c_code_cache_version(self):
-        return (4,)
+        return (5,)
 
     def grad(self, inputs, g_output):
         return [inp.zeros_like(dtype=theano.config.floatX)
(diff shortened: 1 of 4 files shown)


lamblin:
On GPU, the new kernel does not compile. For instance, in `theano.gpuarray.tests.test_neighbours.T_GpuImages2Neibs.test_grad_valid` this line:
```C
0063    if (i1 == 0) {PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero"); return;}
```
gives the following error:
```
NVRTC compile log::
default_program(16): error: identifier "PyExc_ZeroDivisionError" is undefined

default_program(16): error: identifier "PyErr_SetString" is undefined

2 errors detected in the compilation of "default_program".
```

I understand that `Py...` macros may not work inside the kernel (even if included), but for now I have no idea how to handle the error in dividing by zero here.

lamblin:
Trying again with a hack to go back to detect that we are running in a GPU kernel, and disable that error checking.
We will need to find a better solution with @abergeron.

theano/tensor/basic.py
```diff
@@ -4822,33 +4822,45 @@ def infer_shape(self, node, ishapes):
...
             requ = list(requ.data)
             requ_part = [ele for ele in requ if ele != -1]
             crit = len(requ) - len(requ_part)
             if crit == 1 and len(requ_part) > 0:
-                missing = mul(*ishapes[0]) // mul(*requ_part)
+                # If there are both 0 and -1 in requ_size, it is impossible
+                # to determine a right output, but we can at least prevent
+                # a division by 0.
+                requ_size = mul(*requ_part)
+                missing = input_size // (1 if requ_size == 0 else requ_size)
```
nouiz: I think if requ_size is 0, we should not do any replacement. Otherwise, currently we replace with the total size of the input. I don't think this is good.

theano/tensor/basic.py
```diff
@@ -4822,33 +4822,45 @@ def infer_shape(self, node, ishapes):
...
-                rest_size = (mul(*ishapes[0]) // -mul(*new_dims))
-            return [tuple([switch(eq(new_dims[i], -1),
+                requ_size = -mul(*requ)
+                # If there are both 0 and -1 in requ_size, it is impossible
+                # to determine a right output, but we can at least prevent
+                # a division by 0.
+                rest_size = (input_size //
+                             switch(eq(requ_size, 0),
+                                    1,
+                                    requ_size))
```
nouiz: There is an old problem here. If self.ndim is 0, we don't define rest_size, but we always use it! I let you decide if you want to fix this in this PR.

theano/tensor/basic.py
```diff
@@ -4822,33 +4822,45 @@ def infer_shape(self, node, ishapes):
...
             # for the case when there are two or more '-1' values.
             if self.ndim:
-                rest_size = (mul(*ishapes[0]) // -mul(*new_dims))
-            return [tuple([switch(eq(new_dims[i], -1),
+                requ_size = -mul(*requ)
+                # If there are both 0 and -1 in requ_size, it is impossible
+                # to determine a right output, but we can at least prevent
+                # a division by 0.
+                rest_size = (input_size //
+                             switch(eq(requ_size, 0),
```
nouiz: I don't see a better fix for now, but this have the consequence to make the graph bigger. To help aleviate this, maybe a new optimization would be needed.

theano/scalar/basic.py
```diff
@@ -1928,7 +1928,14 @@ def c_code(self, node, name, inputs, outputs, sub):
 
         t = node.inputs[0].type.upcast(*[i.type for i in node.inputs[1:]])
         if t in imap(str, discrete_types):
-            check = 'if (%(y)s == 0) {PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero"); %(fail)s}' % locals()
+            check = dedent('''
+                if (%(y)s == 0) {
+                // do not set Python error message in gpuarray kernel for now
+                #ifndef LID_0
```
nouiz: I think there is a better macro to check for then this one. But I'm not sure if there is one for CUDA and a different one for OpenCL. @abergeron should know that.

theano/scalar/basic.py
```diff
@@ -1924,8 +1924,18 @@ def c_support_code(self):
...
         (x, y) = inputs
         (z,) = outputs
+        fail = sub['fail']
+
         t = node.inputs[0].type.upcast(*[i.type for i in node.inputs[1:]])
         if t in imap(str, discrete_types):
+            check = dedent('''
+                if (%(y)s == 0) {
+                // do not set Python error message in gpuarray kernel for now
+                #ifndef LID_0
```
abergeron: Use `KERNEL` to check if you are compiling for GPU.

theano/scalar/basic.py
```diff
@@ -2029,6 +2042,14 @@ def c_code(self, node, name, inputs, outputs, sub):
             # keep them out of safety, and verify they are useless with an
             # assert.
             assert str(t) in imap(str, discrete_types)
+            check = dedent('''
+                if (%(y)s == 0) {
+                // do not set Python error message in gpuarray kernel for now
+                #ifndef LID_0
```
abergeron: Use `KERNEL`

theano/tensor/basic.py
lamblin: Right, if we know that in advance, we can do that instead of mimicking the symbolic case.
However, a shape of -1 is not valid, and it could lead to weird failures, like alloc(..., -1), or min(-1, 0) getting replaced by 0.

theano/tensor/basic.py
lamblin: If `self.ndim` is 0, then `xrange(self.ndim)` will be empty, and we will not need `rest_size`.

theano/scalar/basic.py
lamblin: OK.

lamblin:
So, integer division by 0 in a cuda kernel seems to return -1, and nothing indicates a launch failure.
Btw, in numpy, it returns 0 and prints a warning the first time.

theano/tensor/basic.py
lamblin: Would the following be better?
`rest_size = input_size // minimum(requ_size, 1)`

nouiz: yes, it would be the same number of nodes as before.

theano/tensor/basic.py
nouiz: right

theano/tensor/basic.py
nouiz: right. Maybe extend the comment to talk about that case.

PR #5855, opened 2017-04-15
Consistent dtype

Sentient07:
In `binary_crossentropy`, we have been using a python float(which is `float64` by default). When we use the `binary_crossentropy` with two `float32`s  the resulting output is a `float64`, which is resulting mismatch in dtype error. The similar change, I would like to apply to other places where we use pythonic floats.
Edit : I will post the detailed error in a short while(As it is on the computer in my lab)

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -2014,7 +2014,8 @@ def binary_crossentropy(output, target):
     TODO : Rewrite as a scalar, and then broadcast to tensor.
 
     """
-    return -(target * tensor.log(output) + (1.0 - target) * tensor.log(1.0 - output))
+    one = np.asarray(1.0).astype(theano.config.floatX)
+    return -(target * tensor.log(output) + (one - target) * tensor.log(one - output))
 
 
 def categorical_crossentropy(coding_dist, true_dist):


Sentient07:
The error that I got was  `GpuArray Exception: 'Out of memory'` while I was training a stacked GAN for an image size of 256 * 256 on a Titan K20m GPU(with 4Gb memory). When I changed the implementation of binary_crossentropy to what I proposed in this PR, I was able to train for a batch size of 10(which initially was 128, and it works with the diff in this PR and without the diff in this PR). I am not able to conclude whether the training starts (with batch size 10) because of reducing the batch_size or due to conversion of the loss function's output to float32. I tried, but wasn't able to find a value of batch size where the training does not happen without applying the diff in this PR but the training starts with the diff applied in this PR. In general, is there an optimization done when we compile the graph which converts all the floatY variables -> floatX ? 
P.S : I used old backend for this training as the new backend is not adapting to a low memory environment. This problem also happened with a segnet esque model (but over VGG16).

theano/tensor/nnet/nnet.py
```diff
@@ -2014,7 +2014,8 @@ def binary_crossentropy(output, target):
     TODO : Rewrite as a scalar, and then broadcast to tensor.
 
     """
-    return -(target * tensor.log(output) + (1.0 - target) * tensor.log(1.0 - output))
+    one = np.asarray(1.0).astype(theano.config.floatX)
```
nouiz: This is what should happen automatically. So I don't see why this is needed? Maybe if floatX is float64 and target and output are float32, there is a problem here. I'm not sure about that. But if output, target and floatX are float32, I don't see a need for this fix.

What is your exact case? dtypes of target and output?

Sentient07: Sorry about the delayed reply. I later recognised there was a problem with my code, that it passed scalars to this method instead of vectors. The problem is when I pass a float32 instead of an arrays/vectors of float32 to this method, it gives me a float64 return. This isn't a trivial case and hence I don't think this difference would matter. I think this is more of a numpy issue, am I correct?

Sentient07:
Closing as it was a mistake in my code, apologies.

theano/tensor/nnet/nnet.py
nouiz: I don't think this is a numpy issue. I won't investigate more if it is fixed on your side.

Issue #5849, opened 2017-04-13
Drastic performance drop encountered when upgrading to latest git version on Windows 10


jc-Rosichini:
**# SYMPTOM:**
Upgrading to pygpu:0.6.4 (from 0.6.2) and to latest theano git, 
the run-time performances in training mode **drop by a factor of about 8** !!!
The GPU load drops from an average of 88% to merely 11%...
![gpu perf](https://cloud.githubusercontent.com/assets/25772796/25025571/181855c8-20a3-11e7-9b14-cd49230b6c44.gif)

**# REFERENCE CONFIG (ie the one that was running ok)**
windows 10
GPU GTX 1080
nvidia driver: 378.92-desktop-win10-64bit-international-whql.exe
cuDNN version 5105 - cudnn-8.0-windows10-x64-v5.1
theano:          0.9.0.dev-6c3d8b4302eab84c7ad8ad375c0d4061f675a11e
numpy:           1.11.3
pygpu:           0.6.2
lasagne:         0.2.dev1

**# AFTER UPGRADE**
**(1) new config 1:**
windows 10
GPU GTX 1080
nvidia driver: 378.92-desktop-win10-64bit-international-whql.exe
cuDNN version 5110 - cudnn-8.0-windows10-x64-v5.1
theano:          0.9.0.dev-6c3d8b4302eab84c7ad8ad375c0d4061f675a11e
numpy:           1.11.3
pygpu:           0.6.4
lasagne:         0.2.dev1

**(2) then tried to use cuDNN v6, roughly the same perfs**
new config 2:
windows 10
GPU GTX 1080
nvidia driver: 378.92-desktop-win10-64bit-international-whql.exe
cudnn-8.0-windows10-x64-v6.0
theano:          0.9.0.dev-6c3d8b4302eab84c7ad8ad375c0d4061f675a11e
numpy:           1.11.3
pygpu:           0.6.4
lasagne:         0.2.dev1

**(3) as Microsoft is moving Windows 10 to Windows Creator, I upgraded both
the OS and the driver... same bad performances :(**
new config 3:
windows 10 - 
GPU GTX 1080
nvidia driver: 381.65-desktop-win10-64bit-international-whql.exe
cudnn-8.0-windows10-x64
theano:          0.9.0.dev-6c3d8b4302eab84c7ad8ad375c0d4061f675a11e
numpy:           1.11.3
pygpu:           0.6.4
lasagne:         0.2.dev1

**# BOTTOM LINE**
[...]

PR #5859, opened 2017-04-18
Always import gpuarray.rng_mrg to have its optimizer imported.

nouiz:
should fix https://github.com/Theano/Theano/issues/5849


nouiz:
jenkins passed, so merging.

Issue #5864, opened 2017-04-18
an issue with tensor.ifelse


rizar:
I have the optimizer swearing on me that I give inputs of different types to `IfElse` op. But in fact they are the same: float32 scalars. Somehow during the optimization the optimizer moves one of the inputs to the gpu without moving the other, and after that, indeed, the types become different, as the error message below says. This only happens with `optimizer=fast_run`.

```
ERROR:theano.gof.opt:Traceback (most recent call last):
  File "/u/bahdanau/Dist/theano/theano/gof/opt.py", line 241, in apply
    sub_prof = optimizer.optimize(fgraph)
  File "/u/bahdanau/Dist/theano/theano/gof/opt.py", line 87, in optimize
    ret = self.apply(fgraph, *args, **kwargs)
  File "/u/bahdanau/Dist/theano/theano/gpuarray/opt.py", line 380, in apply
    node.outputs)
  File "/u/bahdanau/Dist/theano/theano/gpuarray/opt.py", line 939, in local_gpua_lazy_ifelse
    return IfElse(op.n_outs, gpu=True)(c, *inps, return_list=True)
  File "/u/bahdanau/Dist/theano/theano/gof/op.py", line 615, in __call__
    node = self.make_node(*inputs, **kwargs)
  File "/u/bahdanau/Dist/theano/theano/ifelse.py", line 188, in make_node
    'false return values'), t, f, t.type, f.type)
TypeError: ('IfElse requires same types for true and false return values', shared_mbcloss, TensorConstant{0.0}, GpuArrayType<None>(float32, scalar), TensorType(float32, scalar))
```

rizar:
Correction: this also happens with `device=cuda00,optimizer=fast_compile`. Not on CPU though.

nouiz:
Do you have code to reproduce this?

nouiz:
```
import numpy
import theano
from theano import tensor

x = tensor.lscalar('x')
y = theano.shared(numpy.asarray(1, dtype='float32'), name='y')
z = tensor.constant(2.)

a = theano.ifelse.ifelse(x, y, z)
f = theano.function([x], [a])
```
@rizar give me that code to reproduce it.

nouiz:
I have a fix here:
https://github.com/Theano/Theano/pull/5865

thanks

PR #5865, opened 2017-04-18
fix opt warning in the ifelse lifter

nouiz:
fix https://github.com/Theano/Theano/issues/5864

### DIFF
--- a/theano/gpuarray/tests/test_basic_ops.py
+++ b/theano/gpuarray/tests/test_basic_ops.py
@@ -358,9 +358,12 @@ class G_Join_and_Split(test_basic.T_Join_and_Split):
         self.shared = shared
 
     def test_gpusplit_opt(self):
+        # Test that we move the node to the GPU
+        # Also test float16 computation at the same time.
         rng = np.random.RandomState(seed=utt.fetch_seed())
-        m = self.shared(rng.rand(4, 6).astype(self.floatX))
+        m = self.shared(rng.rand(4, 6).astype('float16'))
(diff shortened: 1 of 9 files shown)


theano/gpuarray/tests/test_type.py
```diff
@@ -60,6 +60,20 @@ def test_filter_float():
...
 
 
+def test_gpuarray_shared_scalar():
+    # By default, we don't put scalar as shared variable on the GPU
+    try:
+        gpuarray_shared_constructor(np.asarray(1, dtype='float32'))
+    except TypeError:
+        pass
+    else:
+        assert False, "Expected an error"
```
lamblin: Use nose.tools.assert_raises, rather than trying to reimplement it

nouiz:
done

theano/gpuarray/tests/test_basic_ops.py
```diff
@@ -351,12 +351,19 @@ def setUp(self):
...
+            return gpuarray_shared_constructor(x, target=test_ctx_name,
+                                               **kwargs)
+        self.shared = shared
 
     def test_gpusplit_opt(self):
+        # Test that we move the node to the GPU
+        # Also test float16 computation at the same time.
         rng = np.random.RandomState(seed=utt.fetch_seed())
-        m = self.shared(rng.rand(4, 6).astype(self.floatX))
+        m = self.shared(rng.rand(4, 6).astype('float16'))
```
lamblin: I'm not sure I like that part, as it slightly decreases coverage, but that's a minor point.

PR #5866, opened 2017-04-18
Use ParamsType for other ops

notoraptor:
This PR updates many Ops so that they now use ParamsType to bundle many parameters into one Python/C object.

I was mainly focused on new backend, but I have also worked on other Ops outside gpuarray.

I may have to re-read the code again, but we can now observe ParamsType in some different use cases.

@abergeron @lamblin @nouiz

### DIFF
--- a/theano/gpuarray/extra_ops.py
+++ b/theano/gpuarray/extra_ops.py
@@ -24,7 +24,8 @@ class GpuCumOp(GpuKernelBase, Op):
     """
     SUPPORTED_NDIMS = 3
     __props__ = ('axis', 'mode')
-    params_type = ParamsType(axis=scalar.int32, context=gpu_context_type, device_is_not_cuda=scalar.bool)
+    params_type = ParamsType(axis=scalar.int32,
+                             context=gpu_context_type)
 
     def __init__(self, axis, mode='add'):
         self.axis = int(axis) if axis is not None else 0
@@ -39,7 +40,7 @@ class GpuCumOp(GpuKernelBase, Op):
         return hash(self.axis) ^ hash(self.mode)
 
     def c_code_cache_version(self):
-        return (4,)
+        return (5,)
 
     def c_headers(self):
         return ['<numpy_compat.h>', '<gpuarray/types.h>', '<gpuarray_helper.h>']
@@ -48,9 +49,8 @@ class GpuCumOp(GpuKernelBase, Op):
         return [os.path.dirname(__file__)]
 
     def get_params(self, node):
-        context = GpuKernelBase.get_params(self, node)
-        device_is_not_cuda = (context.kind != b'cuda')
-        return self.params_type.get_params(axis=self.axis, context=context, device_is_not_cuda=device_is_not_cuda)
+        context = node.inputs[0].type.context
+        return self.params_type.get_params(axis=self.axis, context=context)
 
     def make_node(self, x):
         assert x.type.dtype == 'float32', "Only float32 supported for GpuCumOp"
@@ -235,12 +235,9 @@ class GpuCumOp(GpuKernelBase, Op):
         return kernels
 
     def c_code(self, node, nodename, inp, out, sub):
+        if node.inputs[0].type.context.kind != b'cuda':
+            raise NotImplementedError("cuda only")
         return """
-        if (%(params)s->device_is_not_cuda) {
-            PyErr_SetString(PyExc_NotImplementedError, "cuda only");
-            %(fail)s
-        }
-        {
             const size_t* shape = PyGpuArray_DIMS(%(x)s);
             bool needAllocation = !%(z)s || PyGpuArray_NDIM(%(x)s) != PyGpuArray_NDIM(%(z)s);
 
@@ -279,7 +276,6 @@ class GpuCumOp(GpuKernelBase, Op):
                     %(fail)s;
                 }
             }
-        }
         """ % dict(x=inp[0], z=out[0], nodename=nodename, fail=sub['fail'], params=sub['params'])
 
     def c_support_code_struct(self, node, nodename):
--- a/theano/gpuarray/extra_ops.py
+++ b/theano/gpuarray/extra_ops.py
@@ -10,6 +10,9 @@ except ImportError:
 
 from .basic_ops import (as_gpuarray_variable, GpuKernelBase, Kernel, GpuReshape, infer_context_name)
 from .opt import register_opt, op_lifter, register_opt2
+from .type import gpu_context_type
+from theano.gof import ParamsType
+import theano.scalar as scalar
 
 
 class GpuCumOp(GpuKernelBase, Op):
@@ -21,9 +24,10 @@ class GpuCumOp(GpuKernelBase, Op):
     """
     SUPPORTED_NDIMS = 3
     __props__ = ('axis', 'mode')
+    params_type = ParamsType(axis=scalar.int32, context=gpu_context_type, device_is_not_cuda=scalar.bool)
 
     def __init__(self, axis, mode='add'):
-        self.axis = axis if axis else 0
+        self.axis = int(axis) if axis is not None else 0
         self.mode = mode
 
     def __eq__(self, other):
@@ -35,13 +39,18 @@ class GpuCumOp(GpuKernelBase, Op):
         return hash(self.axis) ^ hash(self.mode)
 
     def c_code_cache_version(self):
-        return (3,)
+        return (4,)
 
     def c_headers(self):
         return ['<numpy_compat.h>', '<gpuarray/types.h>', '<gpuarray_helper.h>']
 
     def c_header_dirs(self):
         return [os.path.dirname(__file__)]
+
+    def get_params(self, node):
+        context = GpuKernelBase.get_params(self, node)
+        device_is_not_cuda = (context.kind != b'cuda')
+        return self.params_type.get_params(axis=self.axis, context=context, device_is_not_cuda=device_is_not_cuda)
 
     def make_node(self, x):
         assert x.type.dtype == 'float32', "Only float32 supported for GpuCumOp"
@@ -226,26 +235,23 @@ class GpuCumOp(GpuKernelBase, Op):
         return kernels
 
     def c_code(self, node, nodename, inp, out, sub):
-        if node.inputs[0].type.context.kind != b'cuda':
-            raise NotImplementedError("cuda only")
-        x, = inp
-        z, = out
-        axis = self.axis if self.axis is not None else 0
-        fail = sub['fail']
-        ctx = sub['params']
-
-        code = """
-
+        return """
+        if (%(params)s->device_is_not_cuda) {
+            PyErr_SetString(PyExc_NotImplementedError, "cuda only");
+            %(fail)s
+        }
+        {
             const size_t* shape = PyGpuArray_DIMS(%(x)s);
             bool needAllocation = !%(z)s || PyGpuArray_NDIM(%(x)s) != PyGpuArray_NDIM(%(z)s);
 
-            int axis = %(axis)s;
+            int axis = %(params)s->axis;
             if (axis < 0) {
                 // Convert negative axis to positive axis.
                 axis += PyGpuArray_NDIM(%(x)s);
             }
 
-            if (theano_prep_output(&%(z)s, PyGpuArray_NDIM(%(x)s), PyGpuArray_DIMS(%(x)s), %(x)s->ga.typecode, GA_C_ORDER, %(ctx)s) != 0){
+            if (theano_prep_output(&%(z)s, PyGpuArray_NDIM(%(x)s), PyGpuArray_DIMS(%(x)s),
+                                   %(x)s->ga.typecode, GA_C_ORDER, %(params)s->context) != 0) {
                 %(fail)s;
             }
 
@@ -254,17 +260,17 @@ class GpuCumOp(GpuKernelBase, Op):
                 size_t max_grid_size1;
                 size_t max_grid_size2;
                 int err;
-                err = gpucontext_property(%(ctx)s->ctx, GA_CTX_PROP_MAXLSIZE0, &max_threads_dim0);
+                err = gpucontext_property(%(params)s->context->ctx, GA_CTX_PROP_MAXLSIZE0, &max_threads_dim0);
                 if (err != GA_NO_ERROR){
                     PyErr_SetString(PyExc_RuntimeError, "Could not fetch max_threads_dims0");
                     %(fail)s;
                 }
-                err = gpucontext_property(%(ctx)s->ctx, GA_CTX_PROP_MAXGSIZE1, &max_grid_size1);
+                err = gpucontext_property(%(params)s->context->ctx, GA_CTX_PROP_MAXGSIZE1, &max_grid_size1);
                 if (err != GA_NO_ERROR){
                     PyErr_SetString(PyExc_RuntimeError, "Could not fetch max_grid_size1");
                     %(fail)s;
                 }
-                err = gpucontext_property(%(ctx)s->ctx, GA_CTX_PROP_MAXGSIZE2, &max_grid_size2);
+                err = gpucontext_property(%(params)s->context->ctx, GA_CTX_PROP_MAXGSIZE2, &max_grid_size2);
                 if (err != GA_NO_ERROR){
                     PyErr_SetString(PyExc_RuntimeError, "Could not fetch max_grid_size2");
                     %(fail)s;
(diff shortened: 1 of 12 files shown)


theano/gpuarray/basic_ops.py
```diff
@@ -293,6 +293,12 @@ class GpuKernelBase(object):
     """
     params_type = gpu_context_type
 
+    def get_params(self, node):
+        # Default implementation, suitable for most sub-classes.
+        # To be necessarly overridden if a subclass uses a ParamsType
+        # (see GpuCumOp, GpuAdvancedIncSubtensor1_dev20 for examples).
+        return node.inputs[0].type.context
```
nouiz: Check if the op have a ParamsType and raise an error if so. In case the user forget to add it.

theano/gpuarray/extra_ops.py
```diff
@@ -229,26 +235,23 @@ def gpu_kernels(self, node, nodename):
...
-        x, = inp
-        z, = out
-        axis = self.axis if self.axis is not None else 0
-        fail = sub['fail']
-        ctx = sub['params']
-
-        code = """
-
+        return """
+        if (%(params)s->device_is_not_cuda) {
```
nouiz: That if should be in python and we should raise an error in Python. This will allow Theano to fall back to the Python code. This is the old way of doing think. Why did you changed it?

theano/sandbox/rng_mrg.py
```diff
@@ -425,19 +493,28 @@ def c_code(self, node, name, inp, out, sub):
         const npy_int32 MASK2 = 65535;      //2^16 - 1
         const npy_int32 MULT2 = 21069;
 
+        // The +1 is to avoid odims[0] which fails on windows
+        // We have to read size[i] as an int64, but odims has to be intp*
+        // for NumPy on 32-bit platforms.
```
nouiz: As you now do a malloc, you can remove the +1 and the first line of this comment.

theano/sandbox/rng_mrg.py
```diff
@@ -285,11 +286,17 @@ def mrg_next_value(rstate, new_rstate):
 class mrg_uniform_base(Op):
     # TODO : need description for class, parameter
     __props__ = ("output_type", "inplace")
+    params_type = ParamsType(inplace=bool_t, ndim=int_t, otypenum=int_t, otype_is_float32=bool_t)
```
nouiz: ndim don't need to be a param. It can be taken from the input at run time.

PyArray_NDIM(...)

Similar for otype_is_float32

theano/gpuarray/rng_mrg.py
```diff
@@ -166,40 +167,35 @@ def gpu_kernels(self, node, name):
...
         size_t n_elements = 1;
         unsigned int n_streams;
         int must_alloc_sample = ((NULL == %(o_sample)s)
                 || !pygpu_GpuArray_Check((PyObject*)%(o_sample)s)
                 || !(%(o_sample)s->ga.flags & GA_C_CONTIGUOUS)
-                || (PyGpuArray_NDIM(%(o_sample)s) != %(ndim)s));
+                || (PyGpuArray_NDIM(%(o_sample)s) != %(params)s->ndim));
+
+        // The +1 is to avoid odims[0] which fails on windows
+        size_t* odims = (size_t*)malloc((%(params)s->ndim+1) * sizeof(size_t));
```
nouiz: remove the +1 and the comment

nouiz: We used sometimes in the old back-end alloca() instead. This don't request the call to free(). Can you check if it work with the new back-end especially on windows? This would make the code more readable/simpler.

theano/gpuarray/subtensor.py
```diff
@@ -582,7 +584,14 @@ class GpuAdvancedIncSubtensor1(Op):
...
     __props__ = ('inplace', 'set_instead_of_inc')
-    params_type = gpu_context_type
+    params_type = ParamsType(inplace=bool_t,
+                             set_instead_of_inc=bool_t,
+                             context=gpu_context_type,
+                             # following params are for c_init_code_struct.
+                             ndim_input_0=size_t,
+                             ndim_input_1=size_t,
+                             typecode_input_0=int_t,
+                             typecode_input_1=int_t)
```
nouiz: I think we should see the ParamsType as being the props + context and nothing more. I think all what can be computed at run time should be there at run time. This make the code simpler/more readable I think. If you think otherwise, tell it.

notoraptor: There are some C codes when runtime inputs are not available but we have access to params. For example, in c_init_code_struct() ( https://github.com/Theano/Theano/blob/master/theano/gpuarray/subtensor.py#L361 ). So, in such cases, using params may help avoid hardcoding input properties.

But I will be more strict. Indeed, the general rule should be that params are Op props + gpu_context for gpuarray ( + handle_type for cuDNN ), and should not bundle any property that is reachable at runtime.

theano/sandbox/rng_mrg.py
notoraptor: In this case, `ndim`, `otypenum` and `otype_is_float32` are got from `self.output_type` which is indeed used to create an output type of the Op. But we can not be sure than output storages are allocated or even have right properties when we are in C code (an output pointer may point to NULL).

So I think I should better take these properties here. I will add comments at this line to explain that ?

theano/gpuarray/rng_mrg.py
notoraptor: It seems `alloca()` is machine- and compiler-dependent, and not a standard C function: http://man7.org/linux/man-pages/man3/alloca.3.html . Also, there is no error indication if the required memory cannot be allocated.

I could try it, but isn't it to risky ?

notoraptor:
Updated and rebased to resolve a merge conflicts.

theano/tensor/signal/pool.py
```diff
@@ -614,6 +613,9 @@ def c_headers(self):
         headers += super(Pool, self).c_headers()
         return headers
 
+    def c_code_cache_version(self):
+        return (COp.c_code_cache_version(self), self.openmp)
```
nouiz: good finding. Can you check the other Pool* ops don't have the same problem?

theano/gpuarray/basic_ops.py
```diff
@@ -297,6 +297,9 @@ def get_params(self, node):
         # Default implementation, suitable for most sub-classes.
         # To be necessarly overridden if a subclass uses a ParamsType
         # (see GpuCumOp, GpuAdvancedIncSubtensor1_dev20 for examples).
+        assert (self.params_type is gpu_context_type and
+                node.inputs and
+                isinstance(node.inputs[0].type, GpuArrayType))
```
nouiz: why this condition? Maybe one GPU op would have many inputs and not all of them on the GPU. This would force the first inputs to be on the GPU. Why add this constrain?

theano/gpuarray/extra_ops.py
```diff
@@ -48,9 +49,8 @@ def c_header_dirs(self):
         return [os.path.dirname(__file__)]
 
     def get_params(self, node):
-        context = GpuKernelBase.get_params(self, node)
-        device_is_not_cuda = (context.kind != b'cuda')
-        return self.params_type.get_params(axis=self.axis, context=context, device_is_not_cuda=device_is_not_cuda)
+        context = node.inputs[0].type.context
```
nouiz: I see that the inputs[0] restriction is needed here. Why not keep the old line 51 and remove that restriction?

theano/sandbox/rng_mrg.py
```diff
@@ -287,21 +287,31 @@ def mrg_next_value(rstate, new_rstate):
...
-        self.otype_is_float32 = (self.output_type.dtype == 'float32')
-        # end params
         if inplace:
             self.destroy_map = {0: [0]}
         self.warned_numpy_version = False
 
+    # These attributes (used as params) are created as properties
+    # to avoid some crashes in FAST_COMPILE mode. It seems __init__()
+    # method is not called in that mode, so that some attributes may
+    # not have been initialized.
```
nouiz: I think the problem isn't FAST_COMPILE, but old pickled graph? Adding new attribute request that we implement \__setstate__ to unpickle correctly. Using properties is another fix for that and I like it. If you agree with what I said, just update the comment.

theano/tensor/signal/tests/test_pool.py
```diff
@@ -1146,6 +1146,11 @@ def test_pooling_with_tensor_vars_deprecated_interface(self):
                 utt.assert_allclose(var_dx, fix_dx)
 
     def test_old_pool_interface(self):
+        # Temporarly desactivated because I got some errors
+        # even with Python 3 since I converted signal.Pool to a COp.
+        # Should we just drop old interface testing ? Is it used again ?
+        if issubclass(Pool, theano.gof.COp):
```
nouiz: Enable again the test and use something similar in the class you changed

-    def __setstate__(self, d):
-        self.__dict__.update(d)
-        # Work around to reload old pickle.
-        # We need to find the new file name and reload c code.
-        self.load_c_code(["dnn_base.c", "dnn_conv_base.c", "dnn_fwd.c"])

theano/tensor/signal/pool.py
notoraptor: Checked, all OpenMPOp sub-classes take into account `self.openmp`, and for the moment I have not modified the other CPU Pool* ops.

theano/gpuarray/basic_ops.py
notoraptor: In fact, every GpuKernelBase sub-class that does not get context from first input currently overrides `get_params()`, and I have also overridden it in the Ops that I have modified.

**NB:**
My initial goal was to factorize get_params() because:
 * All GpuKernelBase sub-classes previously defined get_params()
 * About half of all these sub-classes defined get_params() with the same manner: `return node.inputs[0].type.context`.
So, adding this default implementation to GpuKernelBase helped remove many lines of codes in whole code of sub-classes, while keeping get_params() overridden in sub-classes that get context differently. See commit https://github.com/Theano/Theano/pull/5866/commits/c9efa82453dbe28a541fed2791aac92324744054 for the factorization.

I could revert it if necessary !

theano/sandbox/rng_mrg.py
notoraptor: Ok !

theano/tensor/signal/tests/test_pool.py
notoraptor: Done !

theano/gpuarray/extra_ops.py
notoraptor: Because I have added a type checking into parent method GpuKernelBase.get_params(self, node), that will fail if op.params_type is not a gpu_context. As here it's a ParamsType, I must manually get context.

notoraptor:
Updated. Rewritten 4 more Ops in gpuarray DNN and fix an error in op params for GpuDnnConv.

New ops rewritten:
 - GpuDnnConvGradW
 - GpuDnnConvGradI
 - GpuDnnBatchNormInference
 - GpuDnnBatchNormGrad

I have added `check_input = False` for all these ops (including GpuDnnConv).

I wait for testings on Linux.

theano/gpuarray/subtensor.py
```diff
@@ -802,12 +803,12 @@ def c_code(self, node, name, inputs, outputs, sub):
...
                         free(step);
-                        %s
+                        %(fail)s
                    }
-                   """ % sub['fail'])
+                   """ % dict(fail=sub['fail']))
 
     def c_code_cache_version(self):
-        return (3,)
+        return (3, 1)
```
nouiz: Next time, only use 1 number. Otherwise we loose what all the number means.

notoraptor:
New update:
 1. Simplify printing for big sets of ops with `theano-cache list`, to avoid printing too long lines.
 2. Wrap Op params for `Shape_i`.
 3. Allow Theano to support cuDNN v6.

I have tested cuDNN v6 on Windows 8.1 64 bits with Python 2.7 and alll unit tests for DNN module (`theano/gpuarray/tests/test_dnn.py`), and it works. I was thinking we'll have to update some C codes, but strangely, all tests passed.

Single example:

```
(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> nosetests -xvs theano\gpuarray\tests\test_dnn.py:test_version
theano.gpuarray.tests.test_dnn.test_version ... ok

Ran 1 test in 0.004s

OK
```

**NB**: Not yet tested on Linux.

notoraptor:
New updates:
 - Wrap op params for theano.tensor.subtensor.AdvancedIncSubtensor1:
    - inplace (bool)
    - set_instead_of_inc (bool, converted to int)
 - Set `check_input=False` for theano.tensor.subtensor.AdvancedSubtensor1.
 - Wrap op params for theano.tensor.nnet.corr3d.BaseCorr3dMM:
    - direction (enum list)
    - dH, dW, dD: int64
    - dilH, dilW, dilD: int64
    - padH, padW, padD: int64
 - This influences all sub-classes:
    - theano.tensor.nnet.corr3d.Corr3dMM
    - theano.tensor.nnet.corr3d.Corr3dMM_gradWeights
    - theano.tensor.nnet.corr3d.Corr3dMM_gradInputs

notoraptor:
Just to do a checkpoint here: @lamblin I have allowed Theano to support cuDNN v6 (this PR currently allows both cuDNN v5 and v6). I have also tested cuDNN v6 on my computer (Windows 8.1 64 bits, Python 2.7) by running entirely `gpuarray/tests/test_dnn.py`, and all tests succeed. Now I have to test it on Linux (it seems Jenkins currently runs with cuDNN v5 only). Related commits: https://github.com/Theano/Theano/pull/5866/commits/272988f54815aafaecb42404dc8dc06e44ac6e16 and https://github.com/Theano/Theano/pull/5866/commits/1d6949e27f1080d2050539d45e22d07e988a6c05 .

Also, in DNN module:
- It seems GpuDnnPoolDesc is not used at all in whole Theano code ?
- It remains 2 classes to rewrite:
  - theano.gpuarray.dnn._RNNSplitParams
  - theano.gpuarray.dnn.GpuDnnBatchNorm (a little hard, because some macro params are used to include or exclude main function parameters (parts of inputs and outputs), so I can't convert these macros as op params for the moment).

nouiz:
stop adding more stuff to this PR. It is already too big.

notoraptor:
@nouiz Can I rebase and (squash the small correction commits) to reduce number of commits ? I may have to add other ones about cuDNN v6 integration.

nouiz:
yes.

nouiz:
I did reviews of the new commits only. I'll do a global review when you
have rebased/merged commits.

notoraptor:
Branch rebased and commits re-organized to help facilitate reviewing. Gone from 28 to 21 commits.

Removed `__setstate__()` from ParamsType. Let's see if everithing is OK with Travis.

Fixed supported cuDNN version (>= v5 and < v6.1).

I have run some tests about Corr3DMM on my computer: clear theano cache, run test_basic() with master branch, and then run same test when switching to this current branch to check if there are some problems with unpickling: no problems detected on my computer. Let's see what Travis will say.

Apart for potential corrections, I prepare a commit for a better integration of cuDNN v6 (especially, handling new pooling algorithm CUDNN_POOLING_MAX_DETERMINISTIC only when v6 is used).

notoraptor:
NB: Github does not print commits in the right new order ...

notoraptor:
@nouiz was right about corr3d pickling problem. Added a commit that should solve it.

notoraptor:
New update (and last major !) for cuDNN v6 integration:
 - Support MAX DETERMINISTIC algorithm for GpuDnnPool with cuDNN v6.
 - Update pooling tests for DNN module so that it use the right available pooling algorithms depending on runtime cuDNN version.
 - Update CPU Pool and PoolGrad so that they can use same algorithms as GPU counterparts (ie. MAX_DETERMINISTIC when using cuDNN v6).
 - Encapsulate cuDNN constants used in DNN module, so that we can now choose the right cuDNN definitions depending on the runtime cuDNN version. Currently supported cuDNN versions: v5*, v6*.

I think I won't add more major updates from now (as @nouiz said, this PR is big !). Next commits will just be bug fixes, small corrections, or revert commits (if needed).

notoraptor:
jenkins test this again

notoraptor:
Jenkins error fixed.

notoraptor:
PR rebased to resolve conflicts, and commits related to DNN were removed (as this work is now handled by another PR).

This branch becomes a quite "experimental"!

notoraptor:
Definitively closed. Work done here will be transferrred/rewritten into other future PRs.

nouiz:
What is left to port? Porting them would help us for some travis delay problems. Can you list what need to be ported from this PR?

notoraptor:
@nouiz I will have to review the code, since Theano master branch has much changed. But from this PR we could port:
- GpuCumOp
- GPUA_mrg_uniform (CPU op already parameterized)
- AdvancedIncSubtensor1
- GpuAdvancedIncSubtensor1
- Images2Neibs
- GpuImages2Neibs
- compile.ops.Shape_i

I had also converted CPU Pool to a COp, but as the C code seems to have changed, a careful re-reading will be required.

PR #5873, opened 2017-04-19
[Interface change] be more consistent, use ndim instead of outdim

nouiz:


theano/tensor/basic.py
```diff
@@ -5098,31 +5098,41 @@ def flatten(x, outdim=1):
         x : theano.tensor.var.TensorVariable
             the variable that should be reshaped.
 
-        outdim : int
+        ndim : int
             the number of dimensions of the returned variable
-
+        outdim : int
+            DEPRECATED synonyme for ndim
```
lamblin: "synonym"

theano/tensor/basic.py
```diff
@@ -5098,31 +5098,41 @@ def flatten(x, outdim=1):
         x : theano.tensor.var.TensorVariable
             the variable that should be reshaped.
 
-        outdim : int
+        ndim : int
             the number of dimensions of the returned variable
```
lamblin: Maybe add that the default is equivalent to 1.
It was showing in the default value of the function before, but now it shows "None" instead.

theano/tensor/basic.py
```diff
@@ -5098,31 +5098,41 @@ def flatten(x, outdim=1):
...
     -------
     theano.tensor.var.TensorVariable
         the flattend variable with dimensionality of outdim
     """
-    # Any input variable can be flattened to have outdim of 1,
-    # even if it's a scalar. Otherwise, outdim must be positive
+    if outdim is None and ndim is None:
+        ndim = 1
+    elif outdim is not None and ndim is not None:
+        raise ValueError("You should only specify outdim or ndim")
```
lamblin: I would rather say to only use "ndim" as outdim is deprecated.

theano/tensor/basic.py
```diff
@@ -5098,31 +5098,41 @@ def flatten(x, outdim=1):
...
+    elif outdim is not None:
+        warnings.warn(
+            "flatten outdim parmaeter is deprecated, use ndim instead.")
```
lamblin: "parameter"

nouiz:
done

pixelou:
Remaining usages to fix (according to pycharm's code analyser):
- theano/scan_module/scan_opt.py:752, PushOutScanOutput::process_node 
    outer_dot_inputs[0].dimshuffle(1, 0, 2), outdim=2)
- theano/tensor/signal/conv.py:108,110, conv2d
        output = tensor.flatten(output.T, outdim=2).T
        output = tensor.flatten(output.T, outdim=3).T

They trigger the deprecation warnings.

nouiz:
Do you want to make a PR for this? We can do it, but if you do you will be in the git history :)

nouiz:
I found other places, so I'll do it. thanks for the heads up.

nouiz:
https://github.com/Theano/Theano/pull/5974

Issue #5875, opened 2017-04-20
integer division regression


bbudescu:
After commit https://github.com/Theano/Theano/commit/318dd6af2187e5afe8a4d3fe3d84385cb29cac42 which adds debug info for division by zero errors, the compilation of modules using integer division fails when using openmp backend.

Reason: 
c++ [code](https://github.com/Theano/Theano/blob/80a1e8e04f31631ae29936811f3e42d47025542b/theano/gof/cc.py#L112) chunk in theano.gof.cc.failure_code uses `goto` statement, which breaks compilation when passing `-fopenmp` flag to g++ (`theano.elemwise.Elemwise` enables this flag by default), being derived from OpenMPOp.

Code to reproduces issue:
```python
import theano
arr_shape = theano.shared(numpy.zeros((3, 3))).shape
t = arr_shape // numpy.array([2, 2], dtype=numpy.int64)
t.eval()
```

Note that the openmp code is called on array's shape, even if gpuarray backend is available, i.e., if
```python
arr_shape = K.variable(numpy.array([3, 3], dtype=numpy.int64), dtype=numpy.int64)
```
is used, instead, the crash is not reproduced.

gcc error message:
error: jump to label '__label_7' from here exits OpenMP structured block; invalid branch to/from OpenMP structured block

If more info required, please let me know.

nouiz:
thanks for the report, here is the fix:

https://github.com/Theano/Theano/pull/5876

bbudescu:
Thanks for the fix. It seems to work on my side.
Btw, is it expected for openmp to be used on shape tensors when gpuarray is available (and selected)?

nouiz:
I don't understand your last sentence. Do you mean we enable openmp on scalar? or vector?

It is normal that some computation stay on the CPU when when the GPU is activated. Or you mean that we activated openmp for GPU elemwise? I would be very surprised of that.

bbudescu:
No no, I am using the GPU, and was just a bit surprised to see that there are OpenMP Ops in the graph. I was, actually, wondering whether I managed to do something wrong with my model and instead of doing everything on GPU, some stuff remains on the CPU, and, as such, incurs a performance penalty.

nouiz:
We try to keep on the CPU what is faster on the CPU. Elemwise on scalar
shapes is faster on the CPU so we keep them there.

Normally we should move vector elemwise to the GPU, as we only activate
openmp for tensor where ndim >= 1, it mean some such case stay on the CPU.
[...]

PR #5876, opened 2017-04-20
Crash fix with openmp and intdiv and modulo

nouiz:
fix gh-5875

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -1934,8 +1934,10 @@ class IntDiv(BinaryScalarOp):
             # cuda function, which return a binary pattern of all 1s.
             check = dedent('''
                 #ifndef KERNEL
+                #ifndef _OPENMP
                     PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero");
                     %(fail)s
+                #endif
                 #endif
                 ''') % locals()
             x_div_y_pp = '(%(x)s / %(y)s)' % locals()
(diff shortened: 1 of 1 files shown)


lamblin:
Have you tried dividing by zero on CPU with OpenMP?
Because if it brings back the floating point exception / core dump, I'm not sure it is a good idea.

lamblin:
An option may be to return -1 in that case, like with CUDA, but I do not like that either.
A better option IMO would be to redefine `%(fail)s;` in Elemwise, but I'm not sure it is feasible.

nouiz:
I think we should first fix the regression in master. It is better to bring back the segfault as it is more rare then having all openmp code not working. the segfault is due to a problem in the user code, the openmp is due to our problem, so the user can't fix it.

When someone have the time, I think we should revisit the fail in elemwise to allow to give a python error also with openmp. I think the fail in the scalar should just be something like my_error_code = N, and the scalar set it to its own error code. Then there should be another method in the scalar op that will convert that number to a python error and that error should be executed after the loop.

On the GPU, with the unified memory, we could allocate the variable my_error_code on the CPU and still have the kernel write to it. Yes, the error will be delayed if we don't sync as all error reporting on the GPU. This one will be delayed to the next call of that thunk (of it could be one error code for the cmodule, not just one per thunk).

As this is more then just a quick fix, I would push that to its own issue to fix segfaul for openmp and give python error for integer division by 0 on the GPU.

I think we should fix the regression now.

nouiz:
I let @lamblin fix this. It does more then what I understood at first.

PR #5877, opened 2017-04-21
Remove goto in OpenMP Elemwise fail code

### LINKED ISSUE #5875: integer division regression

lamblin:
This makes the code compile, but crash gracefully when intdiv / mod by zero, when using openmp in elemwise.

Fixes #5875.
Counter-proposal for #5876.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -1928,22 +1928,24 @@ class IntDiv(BinaryScalarOp):
 
         t = node.inputs[0].type.upcast(*[i.type for i in node.inputs[1:]])
         if t in imap(str, discrete_types):
-            # If we are in a gpuarray kernel, %(fail)s exits the kernel,
-            # and we do not have any error report, and we cannot set
-            # Python error messages either, so for now we just call the
-            # cuda function, which return a binary pattern of all 1s.
-            check = dedent('''
-                #ifndef KERNEL
-                    PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero");
-                    %(fail)s
-                #endif
-                ''') % locals()
             x_div_y_pp = '(%(x)s / %(y)s)' % locals()
             x_div_y_mp = '((-%(x)s) / %(y)s)' % locals()
             x_mod_y_mp = 'THEANO_MACRO_MOD((-%(x)s), %(y)s)' % locals()
             x_div_y_pm = '(%(x)s / (-%(y)s))' % locals()
             x_mod_y_pm = 'THEANO_MACRO_MOD(%(x)s, (-%(y)s))' % locals()
             x_div_y_mm = '((-%(x)s) / (-%(y)s))' % locals()
+            # If we are in a gpuarray kernel, %(fail)s exits the kernel,
+            # and we do not have any error report, and we cannot set
+            # Python error messages either, so for now we just call the
+            # cuda function, which return a binary pattern of all 1s.
+            div_zero = dedent('''
+                #ifdef KERNEL
+                    %(z)s = %(x_div_y_pp)s;
+                #else
+                    PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero");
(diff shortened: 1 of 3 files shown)


theano/scalar/basic.py
```diff
@@ -1928,22 +1928,24 @@ def c_code(self, node, name, inputs, outputs, sub):
...
             # Python error messages either, so for now we just call the
             # cuda function, which return a binary pattern of all 1s.
-            check = dedent('''
-                #ifndef KERNEL
+            div_zero = dedent('''
+                #ifdef KERNEL
+                    %(z)s = %(x_div_y_pp)s;
+                #else
                     PyErr_SetString(PyExc_ZeroDivisionError, "integer division by zero");
                     %(fail)s
```
nouiz: Did you test that this remove the openmp bug? This don't change the code generated with openmp, the goto is still in the middle of the loop

theano/scalar/basic.py
```diff
@@ -2046,41 +2047,42 @@ def c_code(self, node, name, inputs, outputs, sub):
...
             # Python error messages either, so for now we just call the
             # cuda function, returning a binary pattern depending on dtype
-            check = dedent('''
-                #ifndef KERNEL
+            mod_zero = dedent('''
+                #ifdef KERNEL
+                    %(z)s = %(x_mod_y)s;
+                #else
                     PyErr_SetString(PyExc_ZeroDivisionError, "integer modulo by zero");
                     %(fail)s
```
nouiz: idem

theano/scalar/basic.py
lamblin: Yes, I did. I changed the definition of `%(fail)s` in `Elemwise` to remove the `goto` when `self.openmp` is True.
This fixes the compilation error, as well as the floating point exception.

theano/tensor/elemwise.py
```diff
@@ -1037,12 +1037,26 @@ def _c_all(self, node, nodename, inames, onames, sub):
...
+            # statement in sub['fail']. For now we recreate it here.
+            fail = '''
+                {
+                    %(failure_var)s = %(id)s;
+                    if (!PyErr_Occurred()) {
+                        PyErr_SetString(PyExc_RuntimeError,
+                            "Unexpected error in an Op's C code. "
+                            "No Python exception was set.");
+                    }
+                }''' % sub
```
nouiz: I missed that part. I think it will fix the openmp part. Instead of duplicating this code and we will forget it when we will update the master version of it, what about calling gof.cc.failure_code(sub, use_goto=False)? Add to it a use_goto=True param that allow to remove the goto.

With that small changes, this PR does in part what I suggested.

lamblin: OK.

PR #5878, opened 2017-04-21
avoid creating float64 variables

bbudescu:
If theano config.warn64 is set on 'raise' or 'warn', this generates an error/warning. This should help avoid that.


nouiz:
I don't get how that will remove upcast to float64. Can you explain? Did you saw in real code that this remove the float64 warning?

thanks

bbudescu:
I'm using keras for training a net, and use config.warn_float64='raise'. I got the exception when initializing some weights. This is actually triggered when building the UndefinedGrad op _before_ doing the cast (maybe explicitly casting it to dtype as well would be an alternative fix...).

PR #5883, opened 2017-04-23
Add on_error option for Cholesky

aseyboldt:
The cholesky Op always raises an exception if the matrix is not positive definite. In some cases it can help if it returned nans instead, so that that case can be handled within the theano graph.
This was first implemented in pymc-devs/pymc3#2028
CC @nouiz

### DIFF
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -36,16 +36,28 @@ class Cholesky(Op):
 
     L = cholesky(X, lower=True) implies dot(L, L.T) == X.
 
+    Parameters
+    ----------
+    lower : bool, default=True
+        Whether to return the lower or upper cholesky factor
+    on_error : ['raise', 'nan']
+        If on_error is set to 'raise', this Op will raise a
+        `scipy.linalg.LinAlgError` if the matrix is not positive definite.
+        If on_error is set to 'nan', it will return a matrix containing
+        nans instead.
     """
     # TODO: inplace
     # TODO: for specific dtypes
     # TODO: LAPACK wrapper with in-place behavior, for solve also
 
-    __props__ = ('lower', 'destructive')
+    __props__ = ('lower', 'destructive', 'on_error')
 
-    def __init__(self, lower=True):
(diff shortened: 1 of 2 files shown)


theano/tensor/slinalg.py
```diff
@@ -79,6 +97,13 @@ def grad(self, inputs, gradients):
         dz = gradients[0]
         chol_x = self(x)
 
+        # Replace the cholesky decomposition with 1 if there are nans
+        # or solve_upper_triangular will throw a ValueError.
+        if self.on_error == 'nan':
+            ok = tensor.all(tensor.nlinalg.diag(chol_x) > 0)
```
nouiz: ok = tensor.isnan(tensor.nlinalg.diag(cho_x))

why why check only the diag? the perform set the full output to nan.

theano/tensor/slinalg.py
```diff
@@ -97,9 +122,14 @@ def conjugate_solve_triangular(outer, inner):
             chol_x, tril_and_halve_diagonal(chol_x.T.dot(dz)))
 
         if self.lower:
-            return [tensor.tril(s + s.T) - tensor.diag(tensor.diagonal(s))]
+            grad = tensor.tril(s + s.T) - tensor.diag(tensor.diagonal(s))
+        else:
+            grad = tensor.triu(s + s.T) - tensor.diag(tensor.diagonal(s))
+
+        if self.on_error == 'nan':
+            return [tensor.switch(ok, grad, np.nan)]
```
nouiz: if ok is false, this won't return data in the right shape. I think you want to replace np.nan to:

theano.tensor.fill(grad, np.nan)

lamblin: `tensor.switch` is an Elemwise, so it will broadcast its 3 inputs to the same shape. So I think this code is correct.

nouiz: right

theano/tensor/slinalg.py
aseyboldt: done

Issue #5797, opened 2017-03-31
Redefine the R_Op of ZeroGrad


lamblin:
See #5792, `theano.gradient.ZeroGrad.R_Op` should return zeros (None should work as well for now, but would would be less clean).
Also add tests.

vikramnitin9:
@lamblin : I've submitted a pull request for the above issue at #5891.
Thanks.

PR #5891, opened 2017-04-25
R_op for ZeroGrad + tests

vikramnitin9:
Related to issue #5797 .

I've modified theano/gradient.py and theano/tests/test_gradient.py


nouiz:
I'll let @lamblin review this. We are pretty busy right now, so it could take some times. Thanks for the PR.

Issue #3233, opened 2015-08-01
Calling shape.eval() on SpecifyShape raises an exception


apolotsk:
x_with_const_shape = theano.tensor.specify_shape(theano.tensor.matrix(), (2, 3))
    print(x_with_const_shape.shape.eval())

The code raises an "An input [...] was not provided and not given a value [...]" exception.
Theano waits for input, even though only (a specified) shape is requested.

nouiz:
This is more complicated that it look.

The not optimized graph, need the input it request. We do this check at the
start of compilation.

But yes, the optimized graph won't request it.

I'm not sure this is useful to change this in Theano. What are you tring to
do? Could you do it in another way that won't request that semantic change
in Theano?

apolotsk:
> What are you tring to do?

I am trying automatically (rather than manually) compute "constant" shape through the net, so that if some dimension changes (e.g. input image or filter shape) is changed, all other dependent shaped are changed automatically.

This is needed because some functions expect "constant" input shape for optimization. For example, `theano.tensor.nnet.conv.conv2d`, as in this tutorial [here](http://deeplearning.net/tutorial/lenet.html#lenet):

    # Construct the first convolutional pooling layer:
    # filtering reduces the image size to (28-5+1 , 28-5+1) = (24, 24)
    # maxpooling reduces this further to (24/2, 24/2) = (12, 12)
    # 4D output tensor is thus of shape (batch_size, nkerns[0], 12, 12)
    layer0_input_shape = (batch_size, 1, 28, 28)
    layer0 = LeNetConvPoolLayer(
      rng,
      input=layer0_input,
      image_shape=layer0_input_shape,
      filter_shape=(nkerns[0], 1, 5, 5),
      poolsize=(2, 2)
    )
    ...
    layer1_input_shape = (batch_size, nkerns[0], 12, 12)
    layer1 = LeNetConvPoolLayer(
      rng,
      input=layer0.output,
      image_shape=layer1_input_shape,
      filter_shape=(nkerns[1], nkerns[0], 5, 5),
      poolsize=(2, 2)
    )

The comments explain why we used "12, 12" in `layer1_input_shape`, but if we change layer0's input or filter shape, we have to manually recompute `layer1_input_shape`.

I was hoping to specify shape for `layer0_input` and then define `layer1_input_shape` as `layer0.output.shape.eval()`.

> Could you do it in another way that won't request that semantic change in Theano?

[...]

Issue #5424, opened 2017-01-17
Move Keras Theano infer_shape to Theano and use it when we build the graph


nouiz:
We want for a long time to infer the shape when we build the graph. For this, we can reuse Theano infer_shape mechanism that is used during compilation.

Keras already implemented that:

https://github.com/fchollet/keras/pull/2336/files#diff-ff634bb5c5441d7052449f20018872b8

We can port this to Theano and modify it for our setup and hook it in Theano.

I think we could implement it in Theano like this (quick draft, should be finished later!):

- Move this keras function in gof/utils.py
- Have in each variable an attribute rshape for "real shape" (better name welcome).
- Modify Constant constructor to always set rshape.
- Modify shared variable to have a constant_shape param. If True, use the current shape as the rshape.
- Make propagation of shapes works:
   - Hook in gof/op.py in ``PureOp.__call__`` (close to test_value handling), and call that
- ???Modify the keras function to use infer_shape directly? Faster to execute, but could infer less constant.

f0k:
> I think we could implement it in Theano like this

Not sure how feasible this is, but couldn't `Variable.shape` become a mixed tuple of constants (for constant dimensions) and symbolic variables (for unknown dimensions)? The symbolic dimensions would still be placeholders like now (i.e., a `Subtensor` op acting on a `Shape` op acting on the variable), and only be expanded to a graph on compilation if this avoids computing the variable (i.e., the variable has no other clients than `Shape` ops).

nouiz:
We need to think about it, but if it work, it would give a better user
interface.

Some note to think about. Now var.shape is the output of Shape()(var) that
is a vector. We could make var.shape an attribute that return a tuple of
scalar Theano variables. If needed, it would be converted back to a vector
by a call to as_tensor_variable(). So we would probably not break code (or
only few code).

With that changes, we could have Theano constant in that tuple. Could we
put Python int? Yes, but this would create even uglier code when doing
as_tensor_variable(var.shape). To not request too much change to Theano opt
at first, we could do an opt pass that convert those ugly graph to the
current graph in some cases.

f0k:
> With that changes, we could have Theano constant in that tuple. Could we put Python int? Yes, but [...]

If it still has Theano constants, the user interface is worse than having an extra `rshape` or `cshape` attribute.

> this would create even uglier code when doing as_tensor_variable(var.shape)

[...]

PR #5892, opened 2017-04-25
WIP: Allow missing inputs when they aren't needed after optimization.

nouiz:
fix gh-3233

Do we want that behavior by default? Did I miss something in the implementation?

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -1419,24 +1419,18 @@ class FunctionMaker(object):
         if not isinstance(outputs, (list, tuple)):
             unpack_single = True
             outputs = [outputs]
-        if not isinstance(inputs, (list, tuple)):
-            inputs = [inputs]
+
+        # Find missing inputs and delay the error to after optimization
+        # in case we are able to remove there need during optimization
+        all_inputs = theano.gof.graph.inputs([o.variable for o in outputs])
+        missing_inputs = [i for i in all_inputs if i not in inputs]
 
         # Wrap them in In or Out instances if needed.
-        inputs = [self.wrap_in(i) for i in inputs]
+        inputs = [self.wrap_in(i) for i in inputs + missing_inputs]
         outputs = [self.wrap_out(o) for o in outputs]
-        _inputs = gof.graph.inputs([o.variable for o in outputs] +
-                                   [i.update for i in inputs
-                                    if getattr(i, 'update', False)])
 
         # Check if some input variables are unused
         self._check_unused_inputs(inputs, outputs, on_unused_input)
-
-        # Make a list of (SymbolicInput|SymblicInputKits, indices,
-        # [SymbolicInput,...]), one tuple for each input. (See
-        # Function.indices for more details)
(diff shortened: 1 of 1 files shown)


nouiz:
This need test and we still need to determine if we want that the default behavior or not.

lamblin:
To summarize the discussion we had IRL:
- It changes the behavior of theano.function in ways that can be non-obvious or surprising for users (i.e., DebugMode can start complaining about missing inputs, but that does not mean a missing input is what was wrong in the graph in the first place), and contradict the current documentation and material (i.e., "you need to provide values for all variables in the ancestors of your outputs, except shared variables, because the function is created before the optimization process").
- We would need at least a switch to control that behaviour
- If a useless input of a function is actually an intermediate variable of the full graph, this new behaviour would look for inputs all the way to the graph inputs, and add that part of the graph to the FunctionGraph, which could potentially make optimization longer, or even interfere with shape inference.
- I suggest we add a new argument to `theano.function`, the semantigs being "inputs we would have to provide, but should not need actual values for". They would be used:
   - in `blockers`, to stop looking for input variables earlier
   - in `missing_inputs`, so that we check that they do not have clients at the end of the optimization process
- We still need to come up with a good name and interface for this list of variables, and a way of easily specifying "all the inputs variables" or "only these variables I provide", for instance.

f0k:
> To summarize the discussion we had IRL

Thank you! :+1:

> If a useless input of a function is actually an intermediate variable of the full graph, this new behaviour would look for inputs all the way to the graph inputs, and add that part of the graph to the FunctionGraph, which could potentially make optimization longer, or even interfere with shape inference.

Ah, good point.

> We still need to come up with a good name and interface for this list of variables, and a way of easily specifying "all the inputs variables" or "only these variables I provide", for instance.

Well, we could still call it `defer_check_missing`, allowing `False` (the default, and the current behavior), `True` (implicitly use all input variables not given), or a list of variables.

Note that the only use case this came up for was calling `x.shape.eval()` on a variable that should be able to infer its shape (due to a `specify_shape` somewhere in the graph). If we implement #5424, this use calls falls away. Are there other use cases?

lamblin:
@f0k I had not thought of other cases, but I just thought of one.
One may want to break their computation graph into partial functions, for instance to separate gradient computation from updates, or call some Python code (or repeatedly call another Python function to optimize some value) inbetween parts of the graph. Some times, one may know that some variables will not be needed to compute gradients for instance, but we still have to declare them as inputs and provide dummy data.
In that case, having the option to add `defer_check_missing` on these variables would be useful.

f0k:
@lamblin: Okay, fair enough! I was just worried that this feature is unneeded once we have nonsymbolic shape inference.

lamblin:
It was my concern too :)

nouiz:
I did what we discussed. Ready for another round of review. Do you have a better name for the parameter?

lamblin:
I like the name proposed by @f0k, `defer_check_missing`.
The new test should probably be skipped if the optimizations are not present, and in DebugMode.

Issue #5754, opened 2017-03-23
Porting cuSOLVER linear system solver - help needed. 


botev:
I'm trying to put the cuSOLVER linear system solver routine under a theano op #3332 . However, although there are a few examples of how to do custom ops, the documentation is significantly sparse on what many things are and where my mistake is coming from. As such I'm posting below my first attempt and list all my questions to which I did not find answers online. The code is based on the `fft` wrappers around `skcuda` as the implementation uses `skcuda` as well.
The code for this is in this [gist](https://gist.github.com/botev/156eb765b497c1120f11b25dfe973846)

Now, at the moment, however, the solver provides wrong results. Note that the pure code in `skcuda_cuSOLVER.py` works so there is something wrong of managing things inside the `Theano` op. Possible reasons why this could be happening are:

1. The `skcuda` require transposing the numpy arrays and transposing back because numpy stores row major, and cuda uses column major. It is unclear what is the situation with Theano and how to make that work. 

2. The fact that the `solve` routine is inplace, means that I had to copy the values of `A` and `b` such that they are not modified. However, I'm not sure if the copying is done correctly, as I do not understand how that should be done in `Theano`.

Some extra questions:

1. Why is the syncing of the inputs done? When should it be done for other situations?

2. If we want to perform the op with `overwrite_a` or `overwrite_b` how can we tell `Theano` that the operations are going to be in_place? There was no documentation of what is the `storage_map` and how it works, what are its purposes. 

3. If we want the operator to be automatically promoted for matrix inversion for when the device selected is a GPU, what is the correct way of doing that? Any pointers?

In general, any help of making the op works are most welcome and I'm happy to experiment with various things, but at the stage where it executes, but produces wrong results I do need a few questions answered. 

Another issue I'm having is that there is 0 docs on `pygpu`. How do we create a `pygpu` context or anything else so that we can test things in isolation.?

nouiz:
Some notes, your inputs won't be numpy ndarray, but a gpu version of it. It isn't 100% the same but very close.

You don't know in advance if the inputs will be c order, fortran order or have an arbitrary strides. Probably skcuda expect inputs in fortran order. This add complication to the code for sure, there is example of this in the file gpuarray/linalg.py from memory.

Before continuing, I checked your gist rapidly and found that in gpuarray/linalg.py, there is the class GpuCusolverSolve. Is it what you want? Otherwise, it could be a better starting point the one that wrap cufft, as it seem pretty close to what you want to do.

[...]

PR #5895, opened 2017-04-26
Added an optimization for inplace gpu_cusolver_solve and tests

tfjgeorge:
This addresses some comments in #5754 :
 - fixed the destroy map
 - added an optimization for inplace (for A) in gpuarray/opt.py
 - added A_structure and trans in the op lifter

For the op lifter, as only 2 cases are implemented in gpu_solve (general using LU decomposition and symmetric using cholesky) I considered that if the user requires a solve op for another case, it would be more efficient to keep it on the cpu, than to move it to the gpu using the general LU case.
@notoraptor

### DIFF
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -60,6 +60,11 @@ def attach_cusolver_handle_to_context(ctx):
         with ctx:
             ctx.cusolver_handle = cusolver.cusolverDnCreate()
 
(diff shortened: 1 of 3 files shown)


notoraptor:
jenkins test this

nouiz:
jenkins test this

notoraptor:
CI passed, merged.

PR #5900, opened 2017-04-27
Update libgpuarray version in the doc

nouiz:
as Theano 0.9 work with it and it contain a bug fix.

https://github.com/fchollet/keras/issues/6400

when merged, it should be ported to 0.9 branch.

### DIFF
--- a/doc/install_generic.inc
+++ b/doc/install_generic.inc
@@ -56,12 +56,12 @@ libgpuarray
 :::::::::::
 
 For the stable version of Theano you need a specific version of libgpuarray,
-that has been tagged ``v0.6.2``.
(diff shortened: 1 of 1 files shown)


lamblin:
I'm OK with the change for the moment, but why not write that we require the 0.6 major version, and we recommend getting the latest minor, for instance 0.6.5 at the moment?
This way, if we get new bugfixing releases in 0.6, we will not have to update the doc each time.

nouiz:
It isn't clear if all 0.6* version will be supported by Theano 0.9. So I'll merge for now and when @abergeron is back, we can discuss this together.

nouiz:
I pushed this to the 0.9.X branch

PR #5902, opened 2017-04-28
[BUG, CRASH] Fixes in DebugMode for GPU

lamblin:


theano/gpuarray/blas.py
```diff
@@ -119,11 +119,11 @@ def c_code(self, node, name, inp, out, sub):
             if (%(A)s->ga.flags & GA_C_CONTIGUOUS) {
                 ssize_t a_stride0 = %(A)s->ga.strides[0];
                 %(A)s->ga.strides[0] = %(A)s->ga.strides[1];
-                if (pygpu_blas_rdot(%(x)s, %(A)s, %(y)s, 0) == -1) {
+                if (pygpu_blas_rdot(%(x)s, %(A)s, %(out)s, 0) == -1) {
```
nouiz: I think this fix wrong results. a pretty bad error! Good catch.

theano/gpuarray/multinomial.py
```diff
@@ -327,6 +327,7 @@ def gpu_kernels(self, node, name):
 
                     if (! %(replace)s )
                     {
+                        __syncthreads();
```
nouiz: I'm not sure this is good. Are you sure then when this is executed 100% of the threads in a given blocks execute this line?

Otherwise, there will be cases where we never finish.

theano/tensor/subtensor.py
```diff
@@ -2001,7 +2001,7 @@ def perform(self, node, inp, out_):
         if self.set_instead_of_inc:
             x[idx] = y
         else:
-            if config.cxx:
+            if config.cxx and node.inputs[0].dtype != 'float16':
```
nouiz: I think we should do an issue about that. It could be a but in numpy. Maybe in PyArray_BroadcastToShape. It could also depend of the numpy version.

which numpy version caused you this problem?

.jenkins/jenkins_buildbot_python2_debug.sh
```diff
@@ -9,6 +9,40 @@ export PATH=/usr/local/cuda/bin:$PATH
 export LD_LIBRARY_PATH=/usr/local/cuda/lib64:$LD_LIBRARY_PATH
 export LIBRARY_PATH=/usr/local/cuda/lib64:$LIBRARY_PATH
 
+GPUARRAY_CONFIG="Release"
```
nouiz: @slefrancois is right?

theano/gpuarray/blas.py
lamblin: This would always have crashed here, since `y` had the wrong number of dimensions if it was not aliased to `out`.

theano/gpuarray/multinomial.py
lamblin: I'm not completely sure.
Each thread is supposed to sample something at each iteration, otherwise it means we have un-initialized memory in `global_outs`.
However, if there are completely inactive threads due to scheduling, I'm not sure.

theano/tensor/subtensor.py
lamblin: I don't think it really is an issue with numpy, more that `float16` is not a type in C, just an alias for `int16`.
So the "increment" calls "+" on int16, and then we interpret the result as float16, which does not make sense.

A better way of implementing that from Theano's end could be to use `ufunc.at`, rather than loading a precompiled module. I just opened #5904.

.jenkins/jenkins_buildbot_python2_debug.sh
lamblin: This is copy-pasted from the Theano GPU buildbot.
I think compiling libgpuarray in release is what we want, as we do not need slowness and debugging symbols.

theano/gpuarray/multinomial.py
lamblin: A safer way in the short term may be to remove the parallel normalization, which would make sure only one thread reads / writes from each global location.

slefrancois:
We should add the libgpuarray loading to the regular python2 and python3 buildbots as well.

.jenkins/jenkins_buildbot_python2_debug.sh
slefrancois: It's ok, we always use Release for buildbots. The alternatives are for development of libgpuarray itself.

lamblin:
If I remember correctly:
- for the regular theano 2 buildbot, it would duplicate what is currently in the buildbot for libgpuarray, so maybe not needed
- for python3 and 32-bit python2, I think we only run in FAST_COMPILE, so I think almost all of the GPU tests are skipped (or should be)

However, this is something that we would want to fix / extend in the future, in particular:
- testing with floatX=float32 and float64
- testing everything with python2 and python3 (maybe combine both options)
- refactoring the bash scripts

slefrancois:
My thought is to pull the theano.gpuarray tests out of the libgpuarray buildbot. I think they belong with the rest of the Theano suite, and it will make tracking changes easier.

More precisely, I would leave the theano tests in the gpuarray_buildbot, but have it only trigger when a change is pushed to libgpuarray. That way we still test if changes in libgpuarray break theano.

For FAST_COMPILE, my understanding is that GPU transfers are now down with it:
http://deeplearning.net/software/theano/optimizations.html

Python2_32bit indeed doesn't test GPU, since we don't want to bother withdriver and CUDA in 32-bit.

theano/gpuarray/multinomial.py
nouiz: I let decide which one you prefer. I don't have the head to evaludate the code completly today.

lamblin: In the end, I think scaling the uniform sample rather than the distribution would remove the need for LID_1 > 1, so we can avoid the conflicts. I'll do that soon.

lamblin:
@slefrancois Transfers to GPU are done with `optimizer='fast_compile'`. Technically, they are done with `mode='FAST_COMPILE'` as well, but the latter disables C code generation, so only GPU Ops with a `perform` method would actually run, so in practice it does not make sense, and we do not test the GPU back-end in that mode (maybe setting `force_device=True` or something).

theano/gpuarray/multinomial.py
lamblin: Fixed and rebased.

slefrancois:
@lamblin , I see. Indeed, the python2_32bit and python3 buildbots are in fast_compile.

I would still recommend pulling back the theano.gpuarray into the python2 and python2_debug buildbots.

lamblin:
@slefrancois I'm fine with that.
@nouiz Can you check one last time? The PR is not approved yet.

nouiz:
I added some commits that fix other problem I found while reviewing.

lamblin:
I'm OK with your new changes.

lamblin:
(Sorry, I mis-clicked, blaming cat hair under my mouse.)

PR #5903, opened 2017-04-28
[BUG,CRASH] Fix crash in GPU advanced indexing with no arrays

### LINKED ISSUE #5898: slicetype breaks with GpuArray
```
import theano
var = theano.shared(np.zeros([3, 3], dtype=theano.config.floatX))
slc = theano.tensor.slicetype()
f = theano.function([slc], var[slc])
s = slice(1, 3)
f(s)
```

Works on CPU, but on GPU returns:

```
  File "/home/adam/GitRepos/Theano/theano/compile/function_module.py", line 884, in __call__
  File "/home/adam/GitRepos/Theano/theano/gof/op.py", line 888, in rval
Process Process-3:
    r = p(n, [x[0] for x in i], o)
  File "/home/adam/GitRepos/Theano/theano/gpuarray/subtensor.py", line 559, in perform
    take_idx = sum((i * s for i, s in zip(nidx, strides)))
  File "/home/adam/GitRepos/Theano/theano/gpuarray/subtensor.py", line 559, in <genexpr>
  File "/home/adam/GitRepos/Theano/theano/compile/function_module.py", line 884, in __call__
  File "/home/adam/GitRepos/Theano/theano/gof/op.py", line 888, in rval
    r = p(n, [x[0] for x in i], o)
  File "/home/adam/GitRepos/Theano/theano/gpuarray/subtensor.py", line 559, in perform
    take_idx = sum((i * s for i, s in zip(nidx, strides)))
  File "/home/adam/GitRepos/Theano/theano/gpuarray/subtensor.py", line 559, in <genexpr>
    take_idx = sum((i * s for i, s in zip(nidx, strides)))
TypeError: unsupported operand type(s) for *: 'slice' and 'int'
```

and 

```
TypeError: unsupported operand type(s) for *: 'slice' and 'int'
Apply node that caused the error: GpuAdvancedSubtensor(<GpuArrayType<None>(float32, matrix)>, <slice>)
Toposort index: 0
Inputs types: [GpuArrayType<None>(float32, matrix), <theano.tensor.type_other.SliceType object at 0x7fb98e45b898>]
Inputs shapes: [(50, 2), 'No shapes']
Inputs strides: [(8, 4), 'No strides']
Inputs values: ['not shown', slice(0, 16, None)]
Outputs clients: [[HostFromGpu(gpuarray)(GpuAdvancedSubtensor.0)]]
```

Using var[start:stop] works fine.

lamblin:
Fixes #5898.

### DIFF
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -547,7 +547,8 @@ class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
         if p == 0:
             # The only indexing was through slices and indices.
             # This can happen with symbolic slices for instance.
-            out[0] = x
+            # Since no view_map is set, we need to copy the returned value
+            out[0] = x.copy()
             return
 
         # flatten the array-indexed dimensions
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -544,6 +544,12 @@ class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
         idx_ = ([slice(None)] * p + nidx[p:])
         x = x.__getitem__(idx_)
 
+        if p == 0:
+            # The only indexing was through slices and indices.
+            # This can happen with symbolic slices for instance.
+            out[0] = x
+            return
+
         # flatten the array-indexed dimensions
         shape = ((np.prod(x.shape[0: p]),) +
                  x.shape[p:])
--- a/theano/tensor/subtensor.py
+++ b/theano/tensor/subtensor.py
@@ -2169,9 +2169,14 @@ class AdvancedSubtensor(Op):
 
     def perform(self, node, inputs, out_):
         out, = out_
-        # TODO: in general, we need to re-pack the inputs into a valid
-        # index, just like subtensor
-        out[0] = inputs[0].__getitem__(inputs[1:])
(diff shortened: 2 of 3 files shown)


theano/gpuarray/subtensor.py
```diff
@@ -544,6 +544,12 @@ def perform(self, node, inputs, out_):
         idx_ = ([slice(None)] * p + nidx[p:])
         x = x.__getitem__(idx_)
 
+        if p == 0:
+            # The only indexing was through slices and indices.
+            # This can happen with symbolic slices for instance.
+            out[0] = x
```
nouiz: Are we guaranteed that this won't cause a view of the inputs?

lamblin: Good point, that would actually be in place.
Since this case can be detected in advance, we could set the view_map, but I think for now I'll just force a copy.

lamblin: Actually, it looks like it would also be an issue on the CPU Op.

lamblin:
DebugMode on the new test case confirmed aliasing was an issue, both on CPU and GPU.
The last commit should fix that.

nouiz:
When both CI pass, we can merge.

This PR and the other one related to gpuarray and debugmode make me think we should push hard to make a 0.10 version after NIPS deadline...

Issue #5907, opened 2017-05-01
Grad of OpFromGraph doesn't set a test value


aseyboldt:
Taking derivatives of `OpFromGraph` instances fails, if `config.compute_test_value` is set to `raise`.

```python
import theano
theano.config.compute_test_value='raise'
import theano.tensor as tt
import numpy as np
x = tt.scalar('x')
x.tag.test_value = np.array(1.)
op = theano.OpFromGraph([x], [x ** 3])
y = tt.scalar('y')
y.tag.test_value = np.array(1.)
f = op(y)
grad_f = tt.grad(f, y)
```
```pytb
...
  File /home/adr/git/Theano/theano/compile/builders.py, line 488, in get_grad_op
    self._recompute_grad_op()
  File /home/adr/git/Theano/theano/compile/builders.py, line 306, in _recompute_grad_op
    gdefaults_l = fn_grad(wrt=local_inputs)
  File /home/adr/git/Theano/theano/gradient.py, line 554, in grad
    grad_dict, wrt, cost_name)
  File /home/adr/git/Theano/theano/gradient.py, line 1316, in _populate_grad_dict
    rval = [access_grad_cache(elem) for elem in wrt]
  File /home/adr/git/Theano/theano/gradient.py, line 1316, in <listcomp>
    rval = [access_grad_cache(elem) for elem in wrt]
  File /home/adr/git/Theano/theano/gradient.py, line 1271, in access_grad_cache
    term = access_term_cache(node)[idx]
  File /home/adr/git/Theano/theano/gradient.py, line 1097, in access_term_cache
    *packed):
  File /home/adr/git/Theano/theano/gof/op.py, line 1105, in get_debug_values
     has no test value)
  File /home/adr/git/Theano/theano/gof/op.py, line 1024, in missing_test_message
    raise AttributeError(msg)
AttributeError: Argument 1 has no test value
```
I believe the problem is in [builder.py#L290](https://github.com/Theano/Theano/tree/master/theano/compile#L290), where new variables are created without setting a test value.

nouiz:
thanks for the report fixed in: https://github.com/Theano/Theano/pull/5908

PR #5908, opened 2017-05-01
crash fix and opt warning fix(make opt apply)

nouiz:
fix gh-5907

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -5718,6 +5718,11 @@ def local_opt_alloc(node):
                     assert val.size == 1
                     # check which type of op
                     size = T.mul(*shapes)
+                    if input.dtype == "float32":
+                        # shapes are ints and normally int64.
+                        # We don't want to have a float64 upcast here
+                        # if input is a float32.
+                        size = size.astype(input.dtype)
                     if isinstance(node.op, T.Sum):
                         val = val.reshape(1)[0] * size
                     else:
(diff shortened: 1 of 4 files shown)


aseyboldt:
Thanks for the fix. I can confirm that it works.

theano/tensor/opt.py
```diff
@@ -5696,11 +5696,24 @@ def local_opt_alloc(node):
                                                     only_process_constants=True)
                     assert val.size == 1
                     # check which type of op
-                    casted = T.mul(*shapes).astype(str(input.dtype))
+                    size = T.mul(*shapes)
+                    if input.dtype == "float32":
+                        # shapes are ints and normally int64.
+                        # We don't want to have a float64 upcast here
+                        # if input is a float32.
+                        size = size.astype(input.dtype)
```
lamblin: Why to this cast here when we reduced on all axes, but not below when we reduce on only some of them?

nouiz: I don't understand your comment.

lamblin: The "if" on line 5692 means we are in a case where the reduction was on all axes. The "else" block after line 5721 is the case where the reduction is only done on some axes.
The code in both blocks is similar, but the the cast `size = size.astype(input.dtype)` only happens in the first block, here, and not in the other block (for instance, just after line 5730).
Why that difference?

nouiz: Good catch. I refactored the code to have that part comment to both branch.

PR #5915, opened 2017-05-04
add magma to buildbots

slefrancois:


.jenkins/jenkins_buildbot_python2.sh
```diff
@@ -6,6 +6,7 @@ export THEANO_FLAGS=init_gpu_device=cuda
 
 # CUDA
 export PATH=/usr/local/cuda/bin:$PATH
+export CPATH=/usr/local/cuda/include/$CPATH
```
notoraptor: It should be a colon before $CPATH (`:$CPATH` instead of `/$CPATH`) !

Same in codes added below.

aam-at:
`Magma` fails during compilation `undefined symbol: cublasSetMatrixAsync` and `undefined symbol: cudaMemcpyAsync` which is defined in `cublas_api.h`. @slefrancois Do you know what can be a reason?

slefrancois:
The build system had the conda version of magma leftover, which is at 2.1. I removed it so it uses the one we compiled.

jenkins test this

notoraptor:
@aam-at Another reason seems to be that Theano wants to work with static libraries (.a files) and shared libraries (.so files), but conda package of magma only provide static libraries (and maybe these files are not compiled in a way compatible with Theano). So you must compile magma yourself (on your computer) with shared libraries enabled.

The tricky step will be to configure `make` correctly. In magma folder, there are examples of configuration into sub-folder `make.inc-examples`. You could copy the `make.inc.openblas` example from example folder to magma folder, rename it to `make.inc`, and modify it to correctly set environment variables LIB, OPENBLASDIR and CUDADIR. Even if you don't have openblas, you can use blas and lapack instead (LIB = -lblas -llapack, OPENBLASDIR=/usr, CUDADIR=/usr/local/cuda).

Tell us if you try and succeed or have problems.

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
Can be merged, jenkins earlgrey failure was a test.

nouiz:
merging. But I think there is much copy/paste in all the jenkins related files. Maybe it could be refactorized to have one files that install libgpuarray that all the jobs reuse. Idem for other common sub-part. It could be done later when we updated those part.

Issue #5757, opened 2017-03-23
Hardcoded md5 causes issue with FIPS enabled systems


dareneiri:
I would like to know if there's any discussion of removing the hardcoded md5 hashing in the library, and allowing flexibility for FIPS-enabled systems which do not allow md5. 

I am using RHEL 7.3 with FIPS enabled and installed the following libraries: 
- Theano (0.9.0)
- Lasagne (0.2.dev1)
- deepsurv (0.1.0)

But I cannot import lasagne:
```
$: python
Python 2.7.5 (default, Aug  2 2016, 04:20:16)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import lasagne
ERROR (theano.sandbox.cuda): Failed to compile cuda_ndarray.cu: error:060800A3:digital envelope routines:EVP_DigestInit_ex:disabled for fips
```

If I am incorrect in theano not working on FIPS enabled systems, I'd be happy to know how I can resolve this issue. Thank you!!

lamblin:
To my knowledge, you are the first to report this issue.
I think using a different hashing would not be a problem for that functionality of Theano.
Contributions would be welcome.

dareneiri:
Is there any reason not to switch the hardcoded md5 to sha256? I forked this repo and switched from `hashlib.md5()` to `hashlib.sha256()` and on my FIPS enabled system this seemed to work without issue (I ran an MNIST.py example script on the GPU). I want to know if there could be any negative consequences from this change.

nouiz:
The only downside is a super small slowdown. I don't understand why this would help.

If this really fix your problem, you can make a PR with your fix and link to this issue.

PR #5916, opened 2017-05-04
Change hashlib.md5 to hashlib.sha256 for FIPS enabled systems

dareneiri:
Some enterprise systems are required to be FIPS 140-2 compliant, which blocks MD5 hashing. MD5 is hardcoded in some parts of the code, which I change to sha256 so it is FIPS compliant. I have reported this as [issue 5757](https://github.com/Theano/Theano/issues/5757).

fix gh-5757

I came across this issue on RHEL 7 with Python 2.7 and FIPS enabled. 
```
$ python
Python 2.7.5 (default, Aug 2 2016, 04:20:16)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import lasagne
ERROR (theano.sandbox.cuda): Failed to compile cuda_ndarray.cu: error:060800A3:digital envelope routines:EVP_DigestInit_ex:disabled for fips
```

After the modification I can use the GPU and [run an MNIST example](https://github.com/Lasagne/Lasagne/blob/master/examples/mnist.py) on a FIPS enabled system: 

```
$ : cat /proc/sys/crypto/fips_enabled
1
```

```
(theano-fips) : python
Python 2.7.13 |Continuum Analytics, Inc.| (default, Dec 20 2016, 23:09:15)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import lasagne
/home/spark/anaconda2/envs/theano-fips/lib/python2.7/site-packages/theano/gpuarray/dnn.py:135: UserWarning: Your cuDNN version is more recent than Theano. If you encounter problems, try updating Theano or downgrading cuDNN to version 5.1.
  warnings.warn("Your cuDNN version is more recent than "
Mapped name None to device cuda: Tesla P100-PCIE-16GB (0000:82:00.0)
```
The MNIST example runs without issue.

I have also tested this with python 3.6
```
$ : python
Python 3.6.1 |Continuum Analytics, Inc.| (default, Mar 22 2017, 19:54:23)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import lasagne
/home/spark/anaconda3/envs/theano-fips/lib/python3.6/site-packages/theano/gpuarray/dnn.py:135: UserWarning: Your cuDNN version is more recent than Theano. If you encounter problems, try updating Theano or downgrading cuDNN to version 5.1.
  warnings.warn("Your cuDNN version is more recent than "
Mapped name None to device cuda: Tesla P100-PCIE-16GB (0000:82:00.0)

```
I would appreciate discussion on this modification and whether it is suitable for merging. Thanks!


nouiz:
jenkins test this.

I'm good with this change. It request us to use a stronger hash then needed for our use. Our code is FIPS complient I think as we don't use it for security purpose. But this is trivial change, so I don't mind. I restarted a travis part that had a strange timeout, but still ran to completion...

lamblin:
This change invalidates all the cache for previously-compiled modules, which is probably why Jenkins timed out. I restarted the offending one, hopefully with a cache pre-populated by the other parts.

nouiz:
I also restarted it again.

travis was passing some times ago with an empty cache.

I don't think this invalidate the full cache. Only the part when the version number isn't in the cache. That bypass the code generation completly, so it won't affect that part.

lamblin:
Well, it changes the value returned by the hash of the C code, so I assumed it would recompile everything, but maybe not.
I restarted it again. Not sure why it times out on python3 but the equivalent python2 run works.

nouiz:
This is pretty strange and the fact that it timeout with repetition...
Maybe we should clear the cache to have less element in it? If it fail
again, we should try that I think.

nouiz:
closing/reopen to trigger travis with the new master. hopefully, this will fix the timeout.

lamblin:
Deleted the cache and restarted.

lamblin:
No idea what is happening, but even after deleting the cache, once it timed out because a test did not finish in 10 minutes, and then it reached the global limit again, but only on Python 3.
I think we will need to profile it manually a bit more.

lamblin:
Jenkins  test this please.

nouiz:
@dareneiri do you have any information about the speed difference of the 2 hash? How much slower the new one is compared to the older?

dareneiri:
I haven't noticed any difference in speed when I tested the code prior to any change and when I switched from md5 to sha256. I provide the time per epoch for the mnist.py example below. 

I've been running theano with sha256 and FIPS enabled in combination with lasagne/keras and it seems to run fine -- it doesn't seem abnormally slow.  

If there's anything you would like me to test on my FIPS enabled system, please let me know and I will report back my findings. Although since my machine is administered through an outside systems administrator I cannot easily switch back and forth between FIPS and non-FIPS. 

**sha256**
```
$ : python mnist.py
Mapped name None to device cuda: Tesla P100-PCIE-16GB (0000:82:00.0)
Loading data...
Building model and compiling functions...
Starting training...
Epoch 1 of 500 took 0.245s
training loss: 1.212316
validation loss: 0.401524
validation accuracy: 88.64 %
Epoch 2 of 500 took 0.189s
training loss: 0.564747
validation loss: 0.305702
validation accuracy: 91.03 %
Epoch 3 of 500 took 0.190s
```

**md5**
```
$ : python mnist.py
Mapped name None to device cuda: Tesla P100-PCIE-16GB (0000:82:00.0)
Loading data...
Building model and compiling functions...
Starting training...
Epoch 1 of 500 took 0.243s
training loss: 1.215956
validation loss: 0.410079
validation accuracy: 88.51 %
Epoch 2 of 500 took 0.192s
training loss: 0.568685
validation loss: 0.302084
validation accuracy: 91.29 %
Epoch 3 of 500 took 0.190s
```

dareneiri:
Let me know if you need additional information. 

I set up a virtual environment with python 3.6, my changes in Theano, and FIPS enabled. Then I ran the [mnist.py example](https://github.com/Lasagne/Lasagne/blob/master/examples/mnist.py) and performance seems similar: 

```
$ : python mnist.py
Mapped name None to device cuda: Tesla P100-PCIE-16GB (0000:82:00.0)
Loading data...
Building model and compiling functions...
Starting training...
Epoch 1 of 500 took 0.201s
  training loss:                1.226219
  validation loss:              0.408171
  validation accuracy:          88.82 %
Epoch 2 of 500 took 0.190s
  training loss:                0.567809
  validation loss:              0.304786
  validation accuracy:          91.21 %
Epoch 3 of 500 took 0.189s
  training loss:                0.466900
  validation loss:              0.261007
  validation accuracy:          92.48 %

```

lamblin:
The slowdown may be only for the special workload of the unit tests, where we generate lots of code for different operations, much more than we would in one typical run of an experiment. In other words, the slowdown could be only in terms of code generation or compilation, not execution.

dareneiri:
Is this still under consideration for pushing to master?

lamblin:
It is, once we figure out what is going on with the excessive time for tests, and have a solution.
This branch is very probably OK to use, but merging it as it is would probably make CI fail on the master and the subsequent builds.

lamblin:
If you want to help us figure things out, you could try to run the same script as the failing test, maybe with the timer plugin for nosetests as well, and compare the runtime with the previous version.
Maybe Python 3.3 is to blame for some reason, and we could consider bumping the requirement to python 3.4. Maybe it is a specific package version in the environment the tests use.
You could try to enable Travis on your fork of Theano if you have not done so already, rather than replicating the testing set up elsewhere. Then you can modify .travis.yml to run nosetests with nose-timer, for instance.

dareneiri:
I ended up updating my branch up to master, and then made an additional change of switching numpy to np to get Travis to work. My build passed: https://travis-ci.org/dareneiri/Theano

I did not yet set up nose-timer.

nouiz:
Can you rebase your PR on top of Theano? Having too many merge commit in the history make reading the history complicated. If you need help, tell me.

dareneiri:
My build passed again after rebasing on top of Theano. Let me know what else I can do to help. Thanks!

lamblin:
It failed again the first time, I deleted the cache (thinking that may be the difference between your job and this) and relaunched, but it timed out again.

dareneiri:
I added nose-timer. I don't notice any consistency in specific tests running longer than usual. The last build (#13095.9) looks like was just about to finish completely but simply hit the 50 minute timeout mark. Could the test be split up to avoid getting timed out? 

From https://travis-ci.org/dareneiri/Theano/jobs/239674812:
Ran 1247 tests in 2194.534s

From https://travis-ci.org/Theano/Theano/jobs/239674843
Ran 1247 tests in 2836.003s

lamblin:
`theano.tensor.tests.test_elemwise.test_CAReduce.test_c` went from 273.8209s to 353.7594s. In fact, it seems that the tests involving reductions are the ones to take much longer now, compared to the others.

Maybe the reason is the old cache, and lot of time is spent trying to re-use code from there.
Something you could try is change the .travis.yml in this PR to not use the cached Theano compiledir. If it works, then we could remove Travis's master cache, and subsequent builds should run fine.

I'm still curious about what cause those specific tests to be slower, an if there is a way of preventing that. Maybe by bumping a global version number in all the keys, to force ignoring earlier modules.

dareneiri:
@lamblin Sorry, but I'm not 100% sure how I would change the .travis.yml to not use the cached Theano compiledir. Would I just comment out `$HOME/.theano` in

```
cache:
  directories:
    - $HOME/.cache/pip
    - $HOME/.theano
```

dareneiri:
I ended up adding `rm -rf ~/.theano` in .travis.yml. My build initially failed but I restarted and they all passed. I'm not sure if restarting the build on travis for this PR would help.

lamblin:
I also realized there are a couple of references to "md5" or "MD5" in the docstrings and comments. I think those should be updated.
Also, part of the cache key explicitly mentions "md5", and some checks look for the string "md5" inside the key. For backward-compatibility maybe we should keep it for now, but at least have a warning in comments, as this is largely misleading.

nouiz:
@lamblin how did you delete travis cache? Did you delete it for the PR AND for the master? I think we need to delete both to do what we want.

dareneiri:
@lamblin Correct. I only changed instances of md5 where hashlib was being used, thus the function names are misleading since they still refer to md5 and not sha256: 

In configparser.py, `get_config_md5()` returns `hash_from_code`.

In gof/utils.py, `hash_from_code()` returns `hashlib.md5` 

Finally, in gof/cc.py, is where sig.append specifies md5:
 ```python
       # IMPORTANT: The 'md5' prefix is used to isolate the compilation
        # parameters from the rest of the key. If you want to add more key
        # elements, they should be before this md5 hash if and only if they
        # can lead to a different compiled file with the same source code.
        if insert_config_md5:
            sig.append('md5:' + theano.configparser.get_config_md5())
```

I can comment here to denote that `get_config_md5` is no longer using md5, but sha256, and refer to this PR until the string "md5" is changed, along with the names of some functions that have "md5". 

I will rebase, resolve conflicts, and then update the comments.

dareneiri:
I've rebased but my branch is not completely up to date to theano's master, causing the build to fail. For example, `SUPPORTED_DNN_CONV_ALGO_RUNTIME ` is not defined for me:
https://github.com/dareneiri/Theano/blob/master/theano/configdefaults.py

but is on theano's master: 
https://github.com/Theano/Theano/blob/master/theano/configdefaults.py

nouiz:
Continued in https://github.com/Theano/Theano/pull/6271.

We fixed the travis problem, so that new PR should pass. thanks.

PR #5918, opened 2017-05-05
Allow c_support_code() to support list of strings and update ParamsType

notoraptor:
This PR adds a small part of work done in #5866 , to help reduce the size of that (big) PR, and prevent some future problems:

 - Allow c_support_code() implementations to return a string or a list of strings, to help avoid duplicated support codes into final module code.
 - Make ParamsType.c_support_code() return a list of strings instead of a string.
 - Add utility functions and methods implementations into ParamsType.

@nouiz @lamblin

### DIFF
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -152,7 +152,7 @@ class CLinkerObject(object):
 
     def c_support_code(self):
         """
-        Optional: Return utility code (a string, or a LIST of strings) for use by a `Variable` or `Op` to be
+        Optional: Return utility code (a string, or a list of strings) for use by a `Variable` or `Op` to be
         included at global scope prior to the rest of the code for this class.
 
         QUESTION: How many times will this support code be emitted for a graph
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -152,7 +152,7 @@ class CLinkerObject(object):
 
     def c_support_code(self):
         """
-        Optional: Return utility code for use by a `Variable` or `Op` to be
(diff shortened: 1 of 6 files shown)


theano/gof/params_type.py
```diff
@@ -611,11 +647,13 @@ def c_support_code(self):
...
-                   struct_name_defined=struct_name_defined, struct_name=struct_name, struct_declare=struct_declare,
+        /** End ParamsType %(struct_name)s **/
+        """ % dict(struct_name_defined=struct_name_defined, struct_name=struct_name, struct_declare=struct_declare,
                    struct_init=struct_init, struct_cleanup=struct_cleanup, struct_extract=struct_extract,
                    struct_extract_method=struct_extract_method)
 
+        return list(sorted(list(c_support_code_set))) + [final_struct_code]
+
     def c_code_cache_version(self):
         return ((1, 8), tuple(t.c_code_cache_version() for t in self.types))
```
lamblin: You probably have to update the cache version number here, since the definition of `c_support_code` has changed.

theano/gof/op.py
```diff
@@ -152,7 +152,7 @@ def c_lib_dirs(self):
 
     def c_support_code(self):
         """
-        Optional: Return utility code for use by a `Variable` or `Op` to be
+        Optional: Return utility code (a string, or a LIST of strings) for use by a `Variable` or `Op` to be
```
lamblin: Can you check in the documentation, in particular `doc/extending/cop.txt` and `doc/extending/extending_theano_c.txt`, to see if references to `c_support_code` need to be updated?

notoraptor:
@lamblin fixed.

PR #5923, opened 2017-05-07
don't move op that don't support float16 on the GPU

nouiz:


theano/gpuarray/opt.py
```diff
@@ -1060,7 +1060,7 @@ def local_gpua_advanced_incsubtensor(op, context_name, inputs, outputs):
     set_instead_of_inc = op.set_instead_of_inc
 
     compute_capability = int(context.bin_id[-2])
-    if compute_capability >= 2 and x.ndim == 1 and y.ndim == 0:
+    if compute_capability >= 2 and x.ndim == 1 and y.ndim == 0 and inputs[0].dtype != 'float16':
```
abergeron: AdvancedIncSubtensor1_dev20 has been audited for float16 support so it should be ok.

abergeron:
After some analysis the failing tests on the GPU ended up testing GpuAdvancedIncSubtensor on the CPU and failing because inplace_increment doesn't work with float16.  I replaced it with np.add.at() in the python code and now it works.

I also took the time to remove inplace_increment from cutils_ext so that the code is versioned.

nouiz:
Some tests are failing.

The following branch is something I forgot to push. It is a
rebased/re-ordered/merged of fixes for some of the failure:

https://github.com/nouiz/Theano/commits/float16_v2

In particular this commit is an amended commit that fix one of the failing
test:

https://github.com/nouiz/Theano/commit/2f970e01189bc9e3eaf10f620b40f4d9b015f057

nouiz:
There is a test that expect inplace_increment to be present and some that fail with wrong values:

  File "/home/jenkins/workspace/Theano_PR_core/theano/tensor/tests/test_opt.py", line 2939, in test_local_IncSubtensor_serialize
    if theano.tensor.subtensor.inplace_increment:
AttributeError: 'module' object has no attribute 'inplace_increment'

theano/tensor/subtensor.py
```diff
@@ -1943,8 +1941,7 @@ def copy_of_x(self, x):
                 NPY_ARRAY_ENSURECOPY, NULL)""" % locals()
 
     def c_support_code(self):
-        from theano.gof.cutils import compile_cutils_code
-        return compile_cutils_code()
+        return inc_code()
```
nouiz: I'm not sure of this. One of the goal of the old code was to not always recompile all that code. That is why it was inside cutils. But it seem it was inside cutils and in the op! Do you agree with that interpretation? If so, keeping only 1 of the copy is good.

But currently, we recompile many ops for each input dtype, even if we include all the dtype in part of the code.

Can you add the attribute `check_input = False` to the op? Maybe it is enough to remove the duplicate thunk compiled as the support code already support many c code.

abergeron: The code is compiled exactly as often as before for the op.

I just removed the copy in cutils_ext.

PR #5926, opened 2017-05-08
Avoid AttributeError in `pp`.

### LINKED ISSUE #5925: `pp` should not raise `AttributeError` if input has no 'owner`
[This line](https://github.com/Theano/Theano/blob/7179282715fb04052e0f4e3c14f242eaac1cfe1e/theano/printing.py#L509) causes an `AttributeError` when the input has no `owner` attribute.
Given that the subsequent code already tries to handle the case when `owner is None`, a simple fix would be to use 
```python
node = getattr(output, 'owner', None)
```
instead. In that case, `pp` simply falls back to `print` (or rather `repr`), as one would probably expect when pretty-printing is not possible.
I'll propose a PR to fix this.

kohr-h:
Easy fix. 

Question: now `pp` behaves like `repr` in the cases it failed before - is that ok? Or should it behave like `print`?

fix gh-5925

### DIFF
--- a/theano/printing.py
+++ b/theano/printing.py
@@ -506,7 +506,7 @@ class DefaultPrinter:
         if output in pstate.memo:
             return pstate.memo[output]
         pprinter = pstate.pprinter
-        node = output.owner
+        node = getattr(output, 'owner', None)
         if node is None:
             return leaf_printer.process(output, pstate)
         new_precedence = -1000


nouiz:
This isn't the fix that was discussed in gh-5925 in this comment: https://github.com/Theano/Theano/issues/5925#issuecomment-301162674

This isn't the right fix, so I'll close this PR. If someone want to make a new PR, that would be great.

PR #5927, opened 2017-05-08
Partially factorize GpuKernelBase.get_params() and configure it to use ParamsType.

notoraptor:
This PR adds a part of work done in #5866 , to help reduce the size of that (big) PR, prevent some future problems with ParamsType, and reduce size of Theano code:

 - Add default `GpuKernelBase.get_params()` that retrieves GPU context from first node input type, and then factorize `get_params()` for all sub-classes that retrieve context by this way. Method `get_params()` must be overridden for any subclass that gets context by different way. This proposal helps remove some lines of Theno Python code.
 - Extend GpuKernelBase interface to allow sub-classes to use parameters types other than `gpu_context_type` (especially `ParamsType`). This helps prevent problems when some subclasses of GpuKernelBase will be rewritten with a ParamsType as params type (done in #5866 ).

@lamblin @nouiz

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -295,8 +295,7 @@ class GpuKernelBase(object):
 
     def get_params(self, node):
         # Default implementation, suitable for most sub-classes.
-        # To be necessarly overridden if a subclass uses a ParamsType
-        # (see GpuCumOp, GpuAdvancedIncSubtensor1_dev20 for examples).
+        # To be necessarly overridden in a subclass that uses a ParamsType.
         assert (self.params_type is gpu_context_type and
                 node.inputs and
                 isinstance(node.inputs[0].type, GpuArrayType))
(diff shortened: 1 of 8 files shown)


theano/gpuarray/basic_ops.py
```diff
@@ -293,6 +293,37 @@ class GpuKernelBase(object):
     """
     params_type = gpu_context_type
 
+    def get_params(self, node):
+        # Default implementation, suitable for most sub-classes.
+        # To be necessarly overridden if a subclass uses a ParamsType
+        # (see GpuCumOp, GpuAdvancedIncSubtensor1_dev20 for examples).
```
lamblin: Just to make sure, I suppose this is going to be added in another split of the original PR?
For now, this one removes the definition of `get_params` in `GpuCumOp`.

lamblin: and `GpuAdvancedIncSubtensor1_dev20.get_params` is still the original implementation.

notoraptor: Yes, it will be added to another splits of the original big PR. Maybe should I remove this comment line for the moment ?

lamblin: Since we are focusing on the convolution Ops first, and we may not merge the changes to `GpuCumOp` and `GpuAdvancedIncSubtensor1_dev20` yet, it would make sense to update the comment for the moment.

notoraptor:
@lamblin updated.

PR #5931, opened 2017-05-09
add gpuarray to python2 buildbot

slefrancois:
The FAST_COMPILE buildbot exposes a few ops that don't have perform methods:
http://darjeeling.iro.umontreal.ca:8080/job/Theano_buildbot_python2/168/testReport/

### DIFF
--- a/.jenkins/jenkins_buildbot_python2_debug.sh
+++ b/.jenkins/jenkins_buildbot_python2_debug.sh
@@ -10,7 +10,7 @@ export LD_LIBRARY_PATH=/usr/local/cuda/lib64:$LD_LIBRARY_PATH
 export LIBRARY_PATH=/usr/local/cuda/lib64:$LIBRARY_PATH
 
 GPUARRAY_CONFIG="Release"
-DEVICE=cuda0
+DEVICE=cuda
 LIBDIR=${WORKSPACE}/local
 
 # Make fresh clones of libgpuarray (with no history since we don't need it)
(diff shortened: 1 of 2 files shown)


slefrancois:
jenkins test this

nouiz:
All tests passed. In the message, you talk as if you expected some tests failure. Is that the case? I'm ok to merge, but I'm not sure if that is what you wanted.

slefrancois:
I think it's good to merge. But we need to look at the new errors it exposes. The old libgpuarray buildbot never tested fast_compile.

slefrancois:
jenkins test this on earlgrey?

slefrancois:
jenkins test this on earlgrey

slefrancois:
jenkins test earlgrey

slefrancois:
jenkins test earlgrey

slefrancois:
jenkins test earlgrey

nouiz:
I'm ok to merge if you think it is good. I think we can ignore the earlgrey failures, but just in case, I let you merge.

slefrancois:
They are unrelated tests. I'll merge.

.jenkins/jenkins_buildbot_python2_debug.sh
```diff
@@ -10,7 +10,7 @@ export LD_LIBRARY_PATH=/usr/local/cuda/lib64:$LD_LIBRARY_PATH
 export LIBRARY_PATH=/usr/local/cuda/lib64:$LIBRARY_PATH
 
 GPUARRAY_CONFIG="Release"
-DEVICE=cuda0
```
lamblin: `DEVICE` is only read by libgpuarray / pygpu, which do not understand `"cuda"` anyway, they need the full name.
I think it was needed for running tests on the right device, but I do not think it is necessary for Theano's tests.
If using `cuda` worked, then I think we can delete the whole line.

slefrancois: It was used by the libgpuarray buildbot to do the cmake libgpurray and pygpu tests, which we don't do here. We can remove the line in another PR.

PR #5932, opened 2017-05-09
Add a Python script to help run more exhaustive cuDNN algorithms tests

notoraptor:
This PR adds a script that test all cuDNN algorithms supported by Theano. Current version of script is written as a bash script, but, depending of what we want, it may be rewritten (e.g. as Python script, for more control).

@lamblin


theano/configdefaults.py
```diff
@@ -264,17 +264,15 @@ def safe_no_dnn_algo_bwd(algo):
...
 # The tests will reference those lists.
-SUPPORTED_DNN_CONV_ALGO_FWD = ('small', 'none', 'large', 'fft', 'fft_tiling',
-                               'winograd', 'guess_once', 'guess_on_shape_change',
-                               'time_once', 'time_on_shape_change')
+SUPPORTED_DNN_CONV_ALGO_RUNTIME = ('guess_once', 'guess_on_shape_change', 'time_once', 'time_on_shape_change')
 
-SUPPORTED_DNN_CONV_ALGO_BWD_DATA = ('none', 'deterministic', 'fft', 'fft_tiling',
-                                    'winograd', 'guess_once', 'guess_on_shape_change',
-                                    'time_once', 'time_on_shape_change')
+SUPPORTED_DNN_CONV_ALGO_FWD = ('small', 'none', 'large', 'fft', 'fft_tiling', 'winograd')  # PS: And 'direct' ?
```
lamblin: According to the documentation, "direct" has no implementation at the moment.
However, "winograd_nonfused" has been added.

theano/configdefaults.py
```diff
@@ -264,17 +264,15 @@ def safe_no_dnn_algo_bwd(algo):
...
 
-SUPPORTED_DNN_CONV_ALGO_BWD_FILTER = ('none', 'deterministic', 'fft', 'small',
-                                      'guess_once', 'guess_on_shape_change',
-                                      'time_once', 'time_on_shape_change')
+SUPPORTED_DNN_CONV_ALGO_BWD_DATA = ('none', 'deterministic', 'fft', 'fft_tiling', 'winograd')
+
+SUPPORTED_DNN_CONV_ALGO_BWD_FILTER = ('none', 'deterministic', 'fft', 'small')
```
lamblin: "winograd_nonfused" and "fft_tiling" have been added.

theano/configdefaults.py
```diff
@@ -264,17 +264,15 @@ def safe_no_dnn_algo_bwd(algo):
...
 
-SUPPORTED_DNN_CONV_ALGO_BWD_DATA = ('none', 'deterministic', 'fft', 'fft_tiling',
-                                    'winograd', 'guess_once', 'guess_on_shape_change',
-                                    'time_once', 'time_on_shape_change')
+SUPPORTED_DNN_CONV_ALGO_FWD = ('small', 'none', 'large', 'fft', 'fft_tiling', 'winograd')  # PS: And 'direct' ?
 
-SUPPORTED_DNN_CONV_ALGO_BWD_FILTER = ('none', 'deterministic', 'fft', 'small',
-                                      'guess_once', 'guess_on_shape_change',
-                                      'time_once', 'time_on_shape_change')
+SUPPORTED_DNN_CONV_ALGO_BWD_DATA = ('none', 'deterministic', 'fft', 'fft_tiling', 'winograd')
```
lamblin: "winograd_nonfused" has been added.

theano/gpuarray/tests/check_dnn.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
lamblin: I think in the end, we may want a Python script, as it would be able to interact more easily with the "normal" test cases, for instance inheriting from them, etc.

theano/gpuarray/tests/check_dnn.sh
```diff
@@ -0,0 +1,74 @@
...
+## for every combination of DNN algorithms.
+
+declare -a special_tests=('theano.gpuarray.tests.test_dnn.test_conv3d_bwd'
+                          'theano.gpuarray.tests.test_dnn.test_conv3d_fwd'
+                          'theano.gpuarray.tests.test_dnn.test_dnn_conv_alpha_output_merge'
+                          'theano.gpuarray.tests.test_dnn.test_dnn_conv_border_mode'
+                          'theano.gpuarray.tests.test_dnn.test_dnn_conv_grad'
+                          'theano.gpuarray.tests.test_dnn.test_dnn_conv_inplace'
+                          'theano.gpuarray.tests.test_dnn.test_dnn_conv_merge'
+                          'theano.gpuarray.tests.test_dnn.TestDnnInferShapes')
```
lamblin: There are ones in theano/gpuarray/tests/test_abstractconv.py as well (TestDnnConv2d, TestDnnConv3d).

theano/gpuarray/tests/check_dnn.sh
```diff
@@ -0,0 +1,74 @@
...
+## Run all special Theano DNN tests for every combination of DNN algorithms.
+# This may take a very (very) long time.
+
+echo Running DNN tests depending on algorithms parameters.
+
+for fwd in $SUPPORTED_DNN_CONV_ALGO_FWD
+do
+    for bwd_filter in $SUPPORTED_DNN_CONV_ALGO_BWD_FILTER
+    do
+        for bwd_data in $SUPPORTED_DNN_CONV_ALGO_BWD_DATA
```
lamblin: I don't think we need to test all combinations of algos, they should be pretty independent.

theano/gpuarray/tests/check_dnn.sh
```diff
@@ -0,0 +1,74 @@
...
+            for precision in $SUPPORTED_DNN_CONV_PRECISION
+            do
+                # THEANO_FLAGS is set on multiple lines to avoid a declaration with one too long line.
+                export THEANO_FLAGS=warn.dnn_recent=False
+                export THEANO_FLAGS=$THEANO_FLAGS,dnn.conv.algo_fwd=$fwd
+                export THEANO_FLAGS=$THEANO_FLAGS,dnn.conv.algo_bwd_data=$bwd_data
+                export THEANO_FLAGS=$THEANO_FLAGS,dnn.conv.algo_bwd_filter=$bwd_filter
+                export THEANO_FLAGS=$THEANO_FLAGS,dnn.conv.precision=$precision
+                echo "Running with: $THEANO_FLAGS"
+                eval "nosetests $NOSEARGS $SPECIAL_NOSETESTS"
```
lamblin: We need to expand the existing tests, though, in particular so that:
- we test different precisions with tensors that actually make sense with these precisions
- we check that algorithms that are supposed to be deterministic actually are
- more combinations are tested (for instance, alpha / beta in combination with strides and dilation, etc.)

notoraptor:
New update @lamblin . The exhaustive tester is now a Python script.

As we want to mainly check DNN algorithms, the tester just focuses on the 3 classes that use them (`GpuDnnConv`, `GpuDnnConvGradW` and `GpuDnnConvGradI`).

Next step will be to choose more accurately tests cases. Current cases are just copied from test_dnn for 3D convolutions, and from TestDnnConv2d (theano/gpuarray/tests/test_abstractconv.py) for 2D convolutions.

Also, the number of generated tests is very big (many tens of thousands). Maybe we should have to choose less but more accurate cases to reduce this number.

notoraptor:
New update @lamblin . Tests rewritten to fix some issues, simplify them and allow to change some parameters, including alpha and beta.

Now I'll work on test cases and inputs checkings.

notoraptor:
New huge update. 
- CEnumType updated to make it complete.
- New ops added into check_dnn to get chosen algorithms at runtime. These ops are intended to be used in tests when algo is one of `(time_once, time_on_shape_change, guess_once, guess_on_shape_change)`, as in such cases, we only want to check algo choice, without actually computing convolution. This should help save time. Remark that any algo that may be chosen here would have been tested by the other exhaustive tests.
- Added a class to help generate exhaustive test cases.
- Added base code for test cases checking based on cuDNN documentation info. The goal is to check if a test case actually fails when it does not respect conditions listed into cuDNN documentation. For the moment I have implemented these checking only for FWD algorithms with cuDNN V5.1. And from primary results, some tests that should fail actually pass well. So, I don't know if either cuDNN doc is out of date or if I missed some things, but I think we will have to really run all tests, check which tests pass, which ones fail, and why they fail.
- Added an utility function to get the expected number of tests, given current conditions (floatX, instance of cases generattor, etc.),

There are about 19 000 tests when floatX is float32 or float64, and about 39 000 tests when floatX is float16. This number is limited in many ways, including the cuDNN data type configurations currently supported by Theano: TRUE_HALF_CONFIG, PSEUDO_HALF_CONFIG, FLOAT_CONFIG and DOUBLE_CONFIG.

Next steps:
- Decide how to predict failing tests. It seems following cuDNN conditions is not such a reliable idea (at least for FWD algorithms).
- Add some specific test cases to help cover any corner case.
- Run `check_dnn.py` to see it in action. It may take time!

I think I should temporarly stop here and let @lamblin @abergeron @nouiz make a new review of that code.

theano/gpuarray/dnn_fwd.c
```diff
@@ -163,51 +163,59 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
...
-                   cudnnGetErrorString(err));
-      cuda_exit(c->ctx);
-      return 1;
-    }
+  /*
+  NB: For next algorithm checkings, I think we should just keep the given algorithm
+  if the user did not explicitely want to choose the algorithm at runtime.
+  User should be responsible for having chosen a bad algorithm, and he should
+  have been warned in Python code.
+  */
```
abergeron: If we want to move in that direction, we should probably remove most of the checking done below or at least re-validate it with the supported version.

These checks were added for cudnn v1 and v2 (and maybe v3)

theano/gpuarray/tests/check_dnn_doc.py
```diff
@@ -0,0 +1,284 @@
+"""
```
abergeron: Why is this file called check_dnn_doc if it is a collection of helper functions?

theano/gpuarray/tests/check_dnn.py
```diff
@@ -0,0 +1,637 @@
...
+
+class BaseGpuDnnConvChooseAlgo(DnnBase):
+    """
+    This class and its subclasses allow to retrieve a cuDNN algorithm
+    at runtime without any computation, given the user choose option
+    (time_once, time_on_shape_change, guess_once or guess_on_shape_change).
+    To help reduce whole test time, I suggest we use these classes when
+    algo is one of choose options, as any chosen algorithm would have
+    been tested by the other exhaustive tests.
+    """
```
abergeron: I don't get why we need this.  We shouldn't be testing our own bindings in this script only that the cudnn functions produce the expected results in all cases.

theano/gpuarray/tests/check_dnn_doc.py
notoraptor: I'll rename it. These helper functions are intended to be used to check if specifications listed into cuDNN documentation about algorithms are correct (ie. if doc says FWD/FFT algo should fail in certain conditions, check that it indeed fail), that's why I had named this file "check dnn doc".

theano/gpuarray/dnn_fwd.c
notoraptor: Do we still (want to) support cuDNN < v5 ?

theano/gpuarray/tests/check_dnn.py
notoraptor: The goal was to reduce whole execution time for this script.

As the script will test all algorithms one by one, then an algo chosen at runtime (when using `time_once`, `guess_once`, etc.) will be also tested by the script, in any case. So I considered that we don't need to explicitely run convolutions when we choose algo at runtime (otherwise, we will have 2 runs for a same algo with the same test case): we just want to check if algo is "well chosen".

But indeed, to check if an algo is "well chosen", we may need to run it. I thought we could also get it from other tests, but it would imply to force an order of execution for the tests.

As it is quite experimental, I can remove it.

theano/gpuarray/tests/check_dnn_doc.py
abergeron: I don't think we want to be that exhaustive, but I'm not opposed to it.

We mostly want to make sure that what is supposed to work, works.

theano/gpuarray/tests/check_dnn.py
abergeron: Since this is supposed to be a script that we only run for new cudnn releases, it doesn't matter if it takes 2 days to run (but it may be a pain in the ass for development).

notoraptor: Ok. It seems to take about 4h with current code (tested on a lab machine), so we might stay under 8-10h with runtime algorithms fully tested (4 more algorithms for each of fwd, bwd_data and bwd_filter, for each of 2D and 3D conv) !

notoraptor:
New update. Code is simplified, and a lot of useless code was removed.

I have also removed any changes in C code: currrent PR adds pure Python changes only.

Next steps could be:
- Run check_dnn to see it in action (I could launch it this afternoon).
- Check exhaustive tests cases: maybe add corner cases and extend generation of test cases.
- Check what can be improved in C side, mainly for checking runtime algorithms and other additional checkings that cannot be done on symbolic variables.

A review can be done at this point, @abergeron @lamblin @nouiz .

notoraptor:
Updated to make script more usable, It can now be used to run a single specific test via command line arguments. This should help to check some corner cases quickly and then decide (or not) to add it into test cases. Number of test cases are also reduced by default. 

Could you review the code at this point, @abergeron ?

---
**NB**:
Last week I ran tests, but with cuDNN V5 (I had forgotten that we want to test v6 mainly), and I have sadly overwritten my result log file ... So I will relaunch the script today.

Another remark about one case: some tests have failed on Titan X Pascal for forward 2D convolution with algorithm winograd_non_fused and TRUE_HALF_CONFIG (input float16 and precision float16) when filter dimensions were not (3,3) or (5,5). This seems to follow what is said in cuDNN doc, but based on this doc, tests should fail as soon as filter dimensions are not (3,3) or (5,5). In practice, it fails with TRUE_HALF_CONFIG but pass with other configs (PSEUDO_HALF_CONFIG, FLOAT_CONFIG).

So, if we add some checking in C code to reject any case that is said to be unsupported in cuDNN doc, there may be many situations where case can work, even if doc said it should fail. But we can't know it in advance (it mays depend on GPU also). How to handle it ? Should we simply print some kinds of warning message when a potential failing case is detected ? Or should we just say "theano only supports what is said in cuDNN doc" and reject any potential error case ? @lamblin @abergeron @nouiz .

notoraptor:
Updated, we can now quickly add special test cases.

Added a special test case representing the following code (that fails):
```python
import theano
import numpy as np


x = theano.shared(np.ones((1, 1, 541211, 10), 'f'))
y = theano.shared(np.ones((1, 50, 541211, 1), 'f'))
z = theano.tensor.nnet.abstract_conv.conv2d_grad_wrt_weights(x, y, 
                                                             filter_shape=(50, 1, 3, 10), 
                                                             border_mode=(1, 0), 
                                                             filter_flip=False)
f = theano.function([], z)
f()
```

@lamblin @nouiz @abergeron

notoraptor:
Tests ultimately ended (cuDNN 6020, TITAN X Pascal): `Ran 40817 tests in 127062.725s` (about 35 hours). I have some errors (`FAILED (errors=12715)`), so I will inspect them to adjust the script. Now we may need to go into the C code.

From a first check, I have some errors with runtime chosen algorithms like `time_once`, but I can not know which algorithm have been chosen. This does not help debugging.

Do you think we can introduce some options to tell C code to display chosen algorithm when algo is selected at runtime ? Maybe a config flag then converted to a DEBUG macro in C code ? @abergeron @lamblin @nouiz .

lamblin:
Yes, having access to the algo that was chosen would be really helpful in different cases.
For now, I think a debug flag can make sense, or maybe even a Param (I'm not sure we want to expose that in the interface, though).

In fact, @olimastro has some timings where `time_once` was slower than the default option (even when excluding the first call, and with inputs of all exactly the same size). So it could help with debugging that as well.

notoraptor:
Updated. Current script should be convenient enough. Anyway, I have tried to skip tests that are intended to fail according to cuDNN doc, by specifying ranges of tests for each algorithm (that is ultimately not too complicated, as certain algos have same conditions across different directions [e.g. fft_tiling or winograd_non_fused]).

The main test script is `check_dnn_conv.py`. I have also added another script, `run_dnn_conv.py`, that should help to quickly test a specific case without having to manually write a script test every time.

I have also relaunched the test script (current status is PENDING), I am wathing it.

I think a review can be done. @abergeron @lamblin

notoraptor:
Rebased and updated for 2 specific test cases, see https://github.com/Theano/Theano/pull/5932/commits/4492cc8df47965a7a42bc3cb003ad08f40ed6384 .

Test script finished to run, 5884 seconds. I saved the log, it take 73,2 MBytes on disk.

Now that PR is rebased, I will relaunch it with cmodule.debug set to True to take into account last added error messages.

notoraptor:
Ok, my last log file for script test is available here (titan x pascal, cudnn v6020, cmodule.debug=True): https://drive.google.com/file/d/0B0gVKJadp0n1eEdKUzUtS2ZURUU/view?usp=sharing
```
Ran 31974 tests in 6399.565s

FAILED (errors=7036)
```
I am analyzing it.

notoraptor:
Rebased to take last changes into account.

I will launch the script on cuDNN 7 tonight.

notoraptor:
Finally run script with cuDNN V7 after a rebasing to integrate all last changes, including #6354 (fix about data type configuration).

This is the script output: 
[tests-check-dnn-conv-cudnn-v7-2017-08-24-19h-36min-23sec.log.txt](https://github.com/Theano/Theano/files/1252067/tests-check-dnn-conv-cudnn-v7-2017-08-24-19h-36min-23sec.log.txt)

Output head:
```
Using cuDNN version 7001 on context None
Mapped name None to device cuda: TITAN X (Pascal) (0000:02:00.0)
```

Output end:
```
Ran 21510 tests in 4889.058s

FAILED (errors=704)
```

I have successfully tried to eliminate any CUDNN_STATUS_BAD_PARAM error, and also any CUDNN_STATUS_NOT_SUPPORTED error according to cuDNN documentation (v6 and v7 doc specify same conditions, except for grouped convolutions introduced in v7, and a new restriction for `CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0 (none)`: `Not supported if output is of type CUDNN_DATA_HALF and the number of elements in dw is odd.`).

So, currently, all 704 errors are `WrongValue` errors.

For a quick checking, this is the list of the 21510 generated cases: 
[test-cases.txt](https://github.com/Theano/Theano/files/1252135/test-cases.txt)

Is it exhaustive enough ? If so, I think this PR is correct enough, now. It would just need a quick final review.

notoraptor:
I have rerun the tests after rebasing to integrate fix in #6354 .

Output: 
[tests-check-dnn-conv-cudnn-v7-2017-08-30-10h-48min-34sec.txt](https://github.com/Theano/Theano/files/1271125/tests-check-dnn-conv-cudnn-v7-2017-08-30-10h-48min-34sec.txt)

Output end:
```
Ran 21510 tests in 4117.695s

FAILED (errors=538)
```

We have now 538 errors. Again, these are all WrongValue errors. I tried to make a little categorization of all these errors, and this is the report below.

I think some errors may be ignored by using some less strict absolute tolerance (e.g. there are WrongValue errors with max abs diff about `3e-5`, `5e-5`, etc. where atol is `1e-5`).

There are also errors that seems to happen when conv result contains too big values (ex. `99` or `100`), and small atol (`1e-5`) don't seem to be adequate for such values. These big values may be produced with FWD computation, even if inputs and filters are lowered (`inputs /= 10; filters /= 10`) before computation.

And ultimately, there are reccurrent WrongValue errors with GRADINPUT computation with algo deterministic and TRUE_HALF_CONFIG (float16, float16).

Now, maybe should I investigate more accurately the patterns of errors (inputs shape, filters shape, subsample, dilation, borders, etc.), but it will take more time to do that.

# 538 WrongValue errors

## 2 errors when testing successive calls to functions with runtime-chosen algorithms

Error with guess_on_shape_change:
```
(using CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM , ws:0, hash:FWD|GPU#0000:06:00.0 -g 1 -dim 1,1,5,100,200,100000,100000,20000,200,1 -filt 1,1,4,50,200 -mode conv -pad 0,0,0 -subsample 1,1,1 -dilation 1,1,1 -ff [unaligned])
```
```
Expected : (1, 1, 2, 51, 1) float32 (408, 408, 204, 4, 4) 99.3735 100.258 0 0
Value    : (1, 1, 2, 51, 1) float32 (408, 408, 204, 4, 4) 99.3743 100.258 0 0
```
```
Max Abs Diff:  0.00124359
Mean Abs Diff:  0.000338984
Median Abs Diff:  0.000259399
Std Abs Diff:  0.000279424
Max Rel Diff:  1.24618e-05
Mean Rel Diff:  3.3958e-06
Median Rel Diff:  2.59959e-06
Std Rel Diff:  2.79932e-06

rtol, atol: 1e-05 1e-05
```

Error with time_on_shape_change:
```
(using CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM (timed), ws:0, hash:FWD|GPU#0000:06:00.0 -t -g 1 -dim 1,1,5,100,200,100000,100000,20000,200,1 -filt 1,1,4,50,200 -mode conv -pad 0,0,0 -subsample 1,1,1 -dilation 1,1,1 -ff [unaligned])
```
```
Expected : (1, 1, 2, 51, 1) float32 (408, 408, 204, 4, 4) 99.3735 100.258 0 0
Value    : (1, 1, 2, 51, 1) float32 (408, 408, 204, 4, 4) 99.3743 100.258 0 0
```
```
Max Abs Diff:  0.00124359
Mean Abs Diff:  0.000338984
Median Abs Diff:  0.000259399
Std Abs Diff:  0.000279424
Max Rel Diff:  1.24618e-05
Mean Rel Diff:  3.3958e-06
Median Rel Diff:  2.59959e-06
Std Rel Diff:  2.79932e-06

rtol, atol: 1e-05 1e-05
```

It seems it's the same error: it occurs after 4 calls to conv function, and it seems it's caused by the random values generated after these 4 calls.

I guess computed values are too high (between 99 and 100 here), so that absolute tolerance (1e-5) is too strict.

# 536 other WrongValue errors

## 38 fwd, algo winograd_non_fused, FLOAT_CONFIG (float32, float32)

Max abs diff:
```
min: Max Abs Diff:  1.46553e-05
max: Max Abs Diff:  0.000153534 # 1.53534e-04
```

## 39 gradweight, algo winograd_non_fused, FLOAT_CONFIG (float32, float32)

Max abs diff:
```
min: Max Abs Diff:  1.21184e-05
max: Max Abs Diff:  0.00123596 # 1.23596e-03
```

## 459 gradinput

### With forced algorithms:

#### 39 algo winograd_non_fused, FLOAT_CONFIG (float32, float32)

Max abs diff:
```
min: Max Abs Diff:  1.20923e-05
max: Max Abs Diff:  0.000154063 # 1.54063e-04
```

#### 84 algo deterministic, TRUE_HALF_CONFIG (float16, float16)

Max abs diff:

```
min: Max Abs Diff:  0.023972
max: Max Abs Diff:  4.64614
```

### With runtime algorithms:

All fails with selected algo CUDNN_CONVOLUTION_BWD_DATA_ALGO_1 (deterministic), TRUE_HALF_CONFIG (float16, float16)
Max abs diff:
```
min: Max Abs Diff:  0.0131748 # 1.31748e-02
max: Max Abs Diff:  3.60378
```
Details:
- 84 algo guess_once
- 84 algo guess_on_shape_change
- 84 algo time_once
- 84 algo time_on_shape_change

notoraptor:
Just a note to say that I have finally compared tests script output for cuDNN V6020 and cuDNN V7001, and both versions produce same results. Outputs vary only where cuDNN V7 uses a TENSOR_OP mathtype, or when algorithm is chosen at runtime and timed (which is expectable). So, I can focus only on cudnn V7 output to analyze the 538 WrongValue errors.

Now I will run the script with latest cudnn version 7.0.2, and I will compare output with cudnn 7.0.1.

notoraptor:
One note: ran tests with cuDNN 7.0.2, still same failing tests with WrongValue errors (538). Differences only related to timed runtime chosen algorithms (as expectable), cached vs non-cached and aligned vs non-aligned.

notoraptor:
Some updates here:

#### a) Trying to reduce the number of WrongValue errors

I added 2 commits to help reduce the number of WrongValue errors. These commits introduce a scaling function applied to GPU and reference outputs just before comparing them:
1. Make sure that all compared values have absolute values in [0, 1). Given arrays gpu_arr and cpu_arr, if the max value in both arrays has an integral part (e.g. `12.345`), then divide both arrays with the same ten power to "push" integral part into the mantissa (e.g. here we will divide by 100 so that `12.345` becomes `0.12345`). Doing that, final compared values have more digits into mantissa, and then `assert_allclose` has more chance to pass. See commit c53d879 .
2. Scale down the compared values by dividing them with alpha (`gpu_arr /= alpha`, `cpu_ar /= alpha`) if alpha is not 1. I do that because it seems the error level is proportional with alpha. So, for example, if a comparison has max abs diff to 1e-5 with alpha == 1 (thus, this test passes), it can raise to 3e-5 with alpha = 3 (and then it will fail). I think it's because the distance between compared values is proportional to alpha, as both values have form `alpha * something + beta * something_else`. See commit https://github.com/Theano/Theano/pull/5932/commits/5be0a3c8fa04b6d0bda61edc2cf7c0e75522878e .

With these changes, I have now 502 WrongValue errors (vs 538 before, -36). Not so much, but little less to analyse.

#### b) Analysis of remaining WrongValue errors for runtime-chosen algorithms

As mentionned above ( https://github.com/Theano/Theano/pull/5932/#issuecomment-326653778 ), there are 336 errors related to runtime algorithms that happen only with gradinput tests. From analysis, all these test cases are in TRUE_HALF_CONFIG (float16, float16) with algorithm CUDNN_CONVOLUTION_BWD_DATA_ALGO_1, which we call `deterministic` algorithm. And all these 336 cases are in fact 84 cases repeated for each runtime-algorithm.

It seems that any runtime-choice can select only `deterministic` for these cases because:
- Only 2 algorithms supports TRUE_HALF_CONFIG for gradinput: deterministic and winograd_non_fused
- winograd_non_fused works with TRUE_HALF_CONFIG only if filter width and height equal 3, which is not the case for these 84 cases.

So, these errors are not related to runtime-choice processing, but they seems exclusively related to dereministic algorithm for gradinput.

-----------

Based on that, it seems there are 2 main cases that generates WrongValue errors:
- algorithm `winograd_non_fused` with `FLOAT_CONFIG` in three computations: `fwd`, `gradweight` and `gradinput`
- algorithm `deterministic` with `TRUE_HALF_CONFIG` only in `gradinput` computation.

Now I have to look for a particular pattern into error-cases related to these algorithms.

**NB**: For errors related to winograd_non_fused, some errors are reported with a max abs diff greated than 1e-5 (which is atol) but less than 1e-4 (e.g. `1.5e-5`, `3e-5`, sometimes `7e-5`). Should we increase atol or rtol to skip these cases, or should it be still considered as errors ?

nouiz:
winograd and float16 computation have both cause loose of precision. So raising a little the tolerance for them seem the good thing to do.

Also, in previous discussion, someone told that there precision lost when combined could combine. So even more lost when together.

notoraptor:
With last commit (adjust atol and rtol for float16 and for winograd_non_fused) I get rid of WrongValue errors related to winograd_non_fused: https://github.com/Theano/Theano/pull/5932/commits/cec49e4c602f4d1b192fe9c54ea5c6c0c3cf8d20

But I still have 418 errors, all related to test_gradinput, for deterministic algo and for true half config (float16, float16). So, it seems adjustement for float16 is insufficient (atol to 1e-2, rtol to 5e-2), but some cases have too high absolute difference. I will check some of them.

notoraptor:
Updated. Now I run tests with the following tolerances:
- if dtype is `float16`, `atol = rtol = 5e-2`
- if dtype is `float32` and algo is `winograd_non_fused`, `atol = rtol = 1e-4`
- otherwise, use the default values (`atol = rtol = None`)

With this, I now get **259** WrongValue errors. Still all related to gradinput with algorithm deterministic (CUDNN_CONVOLUTION_BWD_DATA_ALGO_1) in TRUE_HALF_CONFIG (float16, float16).

Note that, in fact, there are many errors which seems repeated across 5 algorithms (deterministic, time_once, time_on_shape_change, guess_once, guess_on_shape_change). It seems to have 71 unique failing remaining cases:
**`dtype  precision  input_shape  filter_shape  subsample  dilation  border  conv  alpha  beta`**
```
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 1), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 1), 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 1), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 2), 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 2), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 2), 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 2), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'valid', 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'valid', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'valid', 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'valid', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 1), 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 1), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 1), 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 1), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 2), 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 2), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 2), 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), (1, 2), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'half', 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'half', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'valid', 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'valid', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'valid', 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'valid', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'cross', 2.0, 0)
float16, float16, ((2, 3, 300, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'half', 'conv', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'half', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'half', 'cross', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 1), (1, 1), 'half', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 40, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'conv', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'cross', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 1), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'conv', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'cross', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), (1, 2), 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'full', 'cross', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, -3.0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'cross', 2.0, 0)
float16, float16, ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'cross', 2.0, -3.0)
```

This is an example of remaining errors. Max abs diff is `0.0672612`, but min and max values for expected and computed tensors are quite different:
```
theano.gpuarray.tests.check_dnn_conv.TestDnnConv2D.test_gradinput('deterministic', 'float16', 'float16', ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, -3.0)) ... ERROR
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
ERROR: theano.gpuarray.tests.check_dnn_conv.TestDnnConv2D.test_gradinput('deterministic', 'float16', 'float16', ((2, 3, 5, 5), (2, 3, 4, 4), (1, 2), (1, 1), 'half', 'conv', 2.0, -3.0))
File "/home/notoraptor/Programmes/anaconda3/envs/python2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
self.test(*self.arg)
File "/mnt/00447B2C447B2396/Users/notoraptor/mila/dev/git/theano/theano/gpuarray/tests/check_dnn_conv.py", line 718, in run_conv_gradinput
utt.assert_allclose(cpu_res, res, rtol=rtol, atol=atol)
File "/mnt/00447B2C447B2396/Users/notoraptor/mila/dev/git/theano/theano/tests/unittest_tools.py", line 358, in assert_allclose
raise WrongValue(expected, value, rtol, atol)
WrongValue: WrongValue
: shape, dtype, strides, min, max, n_inf, n_nan:
Expected : (2, 3, 5, 5) float32 (300, 100, 20, 4) -0.127534 0.048236 0 0
Value    : (2, 3, 5, 5) float16 (150, 50, 10, 2) -0.14917 -0.00093126 0 0
expected    : [[[[ 0.01383977 -0.0599084  -0.02431579 -0.088958   -0.09511721]
[ 0.01279269  0.0092953  -0.05513935 -0.08952527 -0.09402135]
[ 0.00476633 -0.02141697 -0.04480135 -0.05996003 -0.02167717]
[-0.03603354 -0.03427452  0.04823595 -0.072328   -0.09841608]
[-0.01363448 -0.05430799  0.0283673  -0.02096004 -0.1157776 ]]

[[-0.05822901 -0.03308655 -0.08274499 -0.01744862 -0.06213428]
[-0.08868347 -0.02545207 -0.07361984  0.01923475 -0.07867073]
[-0.06562344  0.00594902 -0.08684886 -0.02103244 -0.10678245]
[ 0.03032453  0.00993267  0.03749088 -0.06398831 -0.06275602]
[-0.04535627  0.02032717 -0.04618645 -0.02061405 -0.05832857]]

[[ 0.00572303 -0.06971376 -0.05741205  0.02229025 -0.08218125]
[-0.10079245 -0.09033392 -0.10633874 -0.10963057 -0.09621751]
[-0.00290192 -0.06193265 -0.03449099  0.00181314 -0.03492211]
[ 0.02783681 -0.03674164 -0.11624435  0.01349288  0.00210702]
[-0.08118074 -0.05791102 -0.04231723  0.00699312 -0.01659183]]]


[[[-0.09110334 -0.02693149 -0.03363961  0.03264691 -0.02187998]
[ 0.00170941 -0.02827735 -0.07152166 -0.03649241 -0.07471522]
[ 0.03111568 -0.06222881 -0.06922023 -0.04530059 -0.03484683]
[ 0.0368811  -0.02710874 -0.01658145 -0.08992258 -0.04052602]
[-0.03772923 -0.04390669 -0.08738317  0.02629213 -0.08191486]]

[[-0.07935514 -0.07528234 -0.07779968 -0.05657539 -0.12753391]
[ 0.0291242   0.03731368 -0.03455332  0.03354887 -0.11604305]
[ 0.00092547  0.01501106 -0.01848198 -0.09410138 -0.10381958]
[ 0.00663791 -0.01353155  0.02028484 -0.09701575 -0.08752244]
[-0.06972301 -0.10312776 -0.07674409 -0.05615232 -0.02712751]]

[[ 0.00687717 -0.02907195 -0.07300282  0.0128695   0.00579197]
[-0.00840496 -0.04976337 -0.09022745 -0.01079083 -0.00401966]
[ 0.01778711 -0.11682665  0.01214049 -0.00911287 -0.0687777 ]
[ 0.03675405 -0.10030204 -0.10492836 -0.05332955  0.00979016]
[-0.02100684 -0.08008748  0.00223133 -0.0547365  -0.02444182]]]]
value    : [[[[-0.02871704 -0.09326172 -0.06567383 -0.11773682 -0.11700439]
[-0.04089355 -0.04144287 -0.12030029 -0.14379883 -0.13134766]
[-0.05371094 -0.07519531 -0.10253906 -0.10693359 -0.05557251]
[-0.08410645 -0.07543945 -0.00206566 -0.11590576 -0.13232422]
[-0.05474854 -0.09228516 -0.01130676 -0.05535889 -0.14001465]]

[[-0.09765625 -0.05960083 -0.11828613 -0.04751587 -0.08520508]
[-0.13037109 -0.06542969 -0.12030029 -0.02157593 -0.10565186]
[-0.10565186 -0.03283691 -0.13867188 -0.06634521 -0.13647461]
[-0.00897217 -0.02764893 -0.00710678 -0.10125732 -0.0892334 ]
[-0.07995605 -0.00650024 -0.08422852 -0.04943848 -0.07543945]]

[[-0.01678467 -0.09106445 -0.08483887 -0.00101471 -0.09265137]
[-0.13671875 -0.11853027 -0.14892578 -0.14379883 -0.11883545]
[-0.04278564 -0.09375    -0.07165527 -0.02934265 -0.05737305]
[-0.00807953 -0.06774902 -0.14733887 -0.01860046 -0.01791382]
[-0.11077881 -0.08813477 -0.07080078 -0.0160675  -0.03436279]]]


[[[-0.13500977 -0.0625     -0.08032227 -0.00093126 -0.04510498]
[-0.06555176 -0.09179688 -0.13769531 -0.09393311 -0.10595703]
[-0.02246094 -0.11187744 -0.12469482 -0.09503174 -0.06573486]
[-0.02288818 -0.08520508 -0.07922363 -0.1427002  -0.07202148]
[-0.07531738 -0.08056641 -0.12286377 -0.00857544 -0.10046387]]

[[-0.1151123  -0.10620117 -0.11956787 -0.08361816 -0.14477539]
[-0.02209473 -0.00444794 -0.08911133 -0.0171051  -0.14257812]
[-0.04882812 -0.02905273 -0.06866455 -0.13806152 -0.13183594]
[-0.03790283 -0.05218506 -0.02740479 -0.13525391 -0.10601807]
[-0.10894775 -0.13500977 -0.11694336 -0.08984375 -0.04364014]]

[[-0.02270508 -0.05026245 -0.09869385 -0.01100159 -0.008255  ]
[-0.04846191 -0.08850098 -0.12805176 -0.04309082 -0.0259552 ]
[-0.02011108 -0.14916992 -0.0269165  -0.04766846 -0.08520508]
[-0.0014019  -0.13513184 -0.14660645 -0.08361816 -0.01271057]
[-0.04992676 -0.10931396 -0.0213623  -0.08288574 -0.04095459]]]]
Max Abs Diff:  0.0672612
Mean Abs Diff:  0.0366234
Median Abs Diff:  0.0359213
Std Abs Diff:  0.0118088
Max Rel Diff:  53.7603
Mean Rel Diff:  2.28853
Median Rel Diff:  0.839438
Std Rel Diff:  5.75738

rtol, atol: 0.05 0.05
```

notoraptor:
Updated. For testing, I skip float16 precision for gradinput deterministic algorithm, as that case generates the remaining strange WrongValue errors and we can not make distinction between errors related to NVIDIA and errors related to tolerance (note that tolerance has been increased for float16 precision, with atol = rtol = 5e-2).

I think this PR can be fully reviewed now.

theano/gpuarray/tests/run_dnn_conv.py
```diff
@@ -0,0 +1,153 @@
...
+class BorderAction(TupleAction):
+    # Border extractor for command line args parser.
+    def __call__(self, parser, namespace, values, option_string=None):
+        if values not in ('valid', 'full', 'half'):
+            super(BorderAction, self).__call__(parser, namespace, values, option_string)
+        else:
+            setattr(namespace, self.dest, values)
+
+
+if __name__ != '__main__':
```
nouiz: Move this before the class and add this information in the file docstring.

theano/gpuarray/tests/run_dnn_conv.py
```diff
@@ -0,0 +1,153 @@
...
+
+test = args.computation
+if len(args.input_shape) != len(args.filter_shape):
+    raise ValueError('Expected same length for input shape and filter shape')
+if len(args.input_shape) not in (4, 5):
+    raise ValueError('Expected length 4 or 5 for input shape')
+ndim = len(args.input_shape) - 2
+if ndim == 2:
+    tests = TestDnnConv2D()
+if ndim == 3:
```
nouiz: use "elif" when it make sence instead of "if" in the file. It make it more readable my making it clear there are condition that won't happen together.

theano/gpuarray/tests/check_dnn_conv.py
```diff
@@ -0,0 +1,1136 @@
...
+from theano.gpuarray.dnn import GpuDnnConvDesc, GpuDnnConv, GpuDnnConvGradW, GpuDnnConvGradI, version, get_precision
+from theano.gpuarray.tests.config import mode_with_gpu, ref_cast
+from theano.tensor.nnet.abstract_conv import get_conv_output_shape, assert_conv_shape
+from theano.tensor.nnet.corr import CorrMM, CorrMM_gradInputs, CorrMM_gradWeights
+from theano.tensor.nnet.corr3d import Corr3dMM, Corr3dMM_gradInputs, Corr3dMM_gradWeights
+from theano.tensor.opt import Assert
+
+
+# We provide a special implementation of dnn_conv, dnn_gradweight and dnn_gradinput
+# that support alpha, beta and out as parameters.
```
nouiz: Do we want to move that to the main functions instead? It would be less test code and will help make this more maintained.

theano/gpuarray/tests/run_dnn_conv.py
notoraptor: Fixed

theano/gpuarray/tests/run_dnn_conv.py
notoraptor: Fixed

theano/gpuarray/tests/check_dnn_conv.py
notoraptor: That depends if we want to allow user to specify `alpha`, `beta` and `out`.

But these 3 parameters will be available only if user (or developer) directly calls cuDNN functions (dnn_conv, dnn_gradweight, dnn_gradinput, etc.). They are not currently available in CPU convolutions interface.

notoraptor:
@nouiz updated. I will run script once to check if everything is ok.

notoraptor:
Script run, no errors.

theano/gpuarray/tests/check_dnn_conv.py
nouiz: Personally, I would do the change I propose and I would also do this on the CPU if we where continuing Theano. But to lower the maintenance, I would do it only on the GPU for now.

As this extend the user interface, I would prefer that we ask @abergeron or @lamblin comments on this.

lamblin: I don't think it is worth extending the interface for the abstract convolution and other implementations (nor adding `gemm` and `gemv` functions to provide explicit accumulation and scaling for `dot`).
If you think it would help maintenance, then we can either:
- do what you propose and expand `dnn_conv`, `dnn_gradweight` and `dnn_gradinput`, or
- move the definition with new names into `dnn.py`, and have the current function keep their interface, and call the new ones.

I would favour the second one, because:
- exposing the new options would create an expectation that all convolutions do the same
- exposing the new options means that we have to document them, have meaningful error checking and error messages, and tests that it behaves correctly in all cases (not only the ones in the test script)
- we do not have to worry about changing the interface and what the right order of parameters should be, for instance

theano/gpuarray/cudnn_defs.py
```diff
@@ -123,6 +253,8 @@ class CuDNNV6(CuDNNV51):
                                                 ('CUDNN_CONVOLUTION_BWD_FILTER_ALGO_FFT_TILING', 'fft_tiling'),
                                                 ctype='cudnnConvolutionBwdFilterAlgo_t')
 
+    deterministic_bwd_filter_algorithms = CuDNNV51.deterministic_bwd_filter_algorithms + ('fft_tiling',)
```
nouiz: You didn't reimplement that for cudnn v7. Can you confirm this isn't a forget and it don't have more deterministic algo?

theano/gpuarray/tests/check_dnn_conv.py
nouiz: I was proposing to extend dnn_conv, ... with the new parameters at the end to don't break the interface.

I do not agree this would add the expectation that all convolution do the same. We had and still have some special cases under gpuarray.

But if you insist that this file isn't enough tests, then we should just keep this PR as is. I don't think the options you propose are worth the time of extra testing as we stop investing time in Theano.

theano/gpuarray/cudnn_defs.py
notoraptor: Yes I confirm (checked now), no new deterministic algorithms.

PR #5935, opened 2017-05-10
Wrap Op params for theano.sandbox.rng_mrg.mrg_uniform

notoraptor:
This PR adds a part of work done in #5866 , to help reduce the size of that (big) PR.

Wrap Op params for **`theano.sandbox.rng_mrg.mrg_uniform`**:
 - inplace (bool scalar)
 - ndim (integer scalar)
 - otypenum (integer scalar)
 - otype_is_float32 (bool scalar)

@nouiz @lamblin


theano/sandbox/rng_mrg.py
```diff
@@ -371,45 +388,105 @@ def perform(self, node, inp, out):
...
-        fail = sub['fail']
-        if self.output_type.dtype == 'float32':
-            otype = 'float'
-            NORM = '4.6566126e-10f'  # np.float32(1.0/(2**31+65))
-            # this was determined by finding the biggest number such that
-            # np.float32(number * M1) < 1.0
-        else:
-            otype = 'double'
-            NORM = '4.656612873077392578125e-10'
+        # NB: first node should always be a TensorType, as it is created with as_tensor_variable() into make_node.
```
lamblin: The original comment is about classes inheriting from this one.
For instance, `GPUA_mrg_uniform` inherits from `mrg_uniform_base`, but redefines `make_node` so it does not call `as_tensor_variable()` (since its inputs are supposed to be on GPU).

notoraptor: Ah, ok ! But as `GPUA_mrg_uniform` does have now its own make_node() and c_code(), I think we don't need to check here if first node is a tensor variable, isn't it ?

abergeron: Keeping the assert is not a bad thing to make sure we don't generate code that will lead to a segfault later.

notoraptor:
Assert is back.

Issue #5734, opened 2017-03-20
cudnn v6 support


nouiz:
Since issue https://github.com/Theano/Theano/issues/5679, we make Theano compile with it. We should do:

- [ ] Make a todo list of all the new features we need to wrap, to help do the follow up.
- [ ] Test that our warning in https://github.com/Theano/Theano/pull/5427 isn't useful anymore and remove it.

Dapid:
The new features are:

- [ ] Dilated convolutions
- [ ] Convolution, bias, and activation fused in a single kernel. (Only ReLU and per channel bias).
- [ ] Inference on int8 dtype.
- [ ] 2 new implementations of RNN (Pascal only)
- [ ] 1D FFT convolution
- [ ] `cudnnReduceTensor`, supporting 8 reduce operations
- [ ] ELU activation
- [ ] Deterministic max pooling

nouiz:
Thanks, I put that in the description.

Dapid:
I am doing a lot of 1D convolutions, so I'd would like to help with that. Any pointers on where to start?

Also, regarding the ELU, it will probably not make a big difference, but it seems easy enough.

HapeMask:
> Convolution, bias, and activation fused in a single kernel. (Only ReLU and per channel bias).
>> Is it worth it?

I've seen at least one paper mention eliminating biases to avoid extra kernel calls where possible: https://arxiv.org/pdf/1606.02147.pdf (but they don't say how much performance is gained by reducing kernel calls).

I took a quick look to see if it could be easily made into a local optimization, but since ReLU doesn't currently have its own Op I'm not sure how to reliably detect where this optimization could be applied.

nouiz:
I think the speed up for big model is small. Also, that optimization maybe
could be applied only for the inference depending how it was implemented.
Any idea is we can use it for gradient?

It won't request just an optimizer, but also c code changes.

HapeMask:
> maybe could be applied only for the inference

You're right, I just looked at the cudnn docs and v6 only adds ```cudnnConvolutionBiasActivationForward``` so it probably wouldn't help training performance much.

astooke:
Word to the wise...when upgrading CuDNN version, make sure to call `theano-cache clear`.

[...]

PR #5936, opened 2017-05-10
Add support for cudnn v6 dilated convolution.

HapeMask:
Addresses the first part of #5734.

The old "upscale" parameter to ```cudnnSetConvolutionNdDescriptor``` is now the dilation parameter, but coincidentally everything worked fine from v5->v6 since the default value passed was {1,1,1}. I just added some code to properly set it from the python side. Only the IMPLICIT_GEMM algorithm supports dilation but the existing code already falls back to this if the user's algo choice isn't valid.

As a small anecdote, I patched Lasagne to use Theano's AbstractConv2D interface for dilated conv as well and compared the performance. For a network with Input -> DilatedConv2D, the speedup was small, ~1.33x but with more dilated layers, the improvement is significant. With 3 dilated layers, there is a >5x speedup.


nouiz:
@notoraptor I let you check this PR.

theano/gpuarray/dnn.py
```diff
@@ -1009,13 +1027,14 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1),
...
-                                    desc_op.subsample)
+                                    desc_op.subsample,
+                                    filter_dilation=dilation)
     out_shp = assert_conv_shape(out_shp)
     out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*out_shp)
     return GpuDnnConv(algo=algo)(img, kerns, out, desc)
 
 
-def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1),
+def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1, 1, 1),
```
notoraptor: Same here for dilation entry in docstring.

theano/gpuarray/dnn.py
```diff
@@ -1141,7 +1161,7 @@ def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
     kerns_shp = as_tensor_variable(kerns_shp)
     precision = get_precision(precision, [img, topgrad])
 
-    desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample,
+    desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample, dilation=dilation,
                           conv_mode=conv_mode, precision=precision)(kerns_shp)
     out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*kerns_shp)
     return GpuDnnConvGradW()(img, topgrad, out, desc)
```
notoraptor: You should also add dilation parameter for `dnn_gradweight3d()` below, isn't it ?

theano/gpuarray/dnn.py
```diff
@@ -1169,7 +1189,7 @@ def dnn_gradinput(kerns, topgrad, img_shp, border_mode='valid',
     img_shp = as_tensor_variable(img_shp)
     precision = get_precision(precision, [kerns, topgrad])
 
-    desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample,
+    desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample, dilation=dilation,
                           conv_mode=conv_mode, precision=precision)(kerns.shape)
     out = GpuAllocEmpty(dtype=kerns.dtype, context_name=ctx_name)(*img_shp)
     return GpuDnnConvGradI()(kerns, topgrad, out, desc)
```
notoraptor: You should also add dilation parameter for `dnn_gradinput3d()` below, isn't it ?

HapeMask:
Thanks for looking over it! I think I covered all the requested changes w/the last commit.

notoraptor:
jenkins test this

notoraptor:
jenkins test this

notoraptor:
ok to test

notoraptor:
jenkins test this

slefrancois:
jenkins test this

HapeMask:
Oh I see I need to update the DNN tests to take into account dilation as well.

It looks like the tests try two different subsample patterns, 1x and 2x, so I'll add a similar set of dilation patterns.

HapeMask:
I updated the tests to always handle dilation. Though the cudnn v6 release notes / docs don't say anything about 3D convs not supporting dilation, I couldn't get it to work. Trying to allocate the workspace always returned CUDNN_STATUS_NOT_SUPPORTED whenever a 3D conv had dilation. Maybe I'm missing something, but for now the tests don't cover that case.

I see that Jenkins is currently testing w/cudnnv5 so it won't actually run the new dilation tests. I made sure to run it on my machine w/cudnnv6 though and they all passed.

slefrancois:
BTW, we're setting up a new buildbot server and it should have CuDNN 6. It will probably go online by the month.

theano/gpuarray/tests/test_dnn.py
```diff
@@ -656,55 +666,62 @@ def test_conv3d_none(self, border_mode, conv_mode):
                         border_mode,
                         conv_mode,
                         [(1, 1, 1), (2, 2, 2)],
+                        [(1, 1, 1)],
```
notoraptor: About dilation problem for 3D conv, I am not familiar with concepts and purposes around dilation, but have you tried to either increase the size of image shape or decrease the size of kernel shape while applying a dilation other than (1, 1, 1) ? It seems it may works with dilation == (2, 2, 2) if you change shape of image value (at line 664), e.g: if `np.random.rand(10, 2, 6, 4, 11)` becomes `np.random.rand(10, 2, 7, 5, 11)`.

Could you try it ?

HapeMask: Ah, I had tried increasing it before by a small amount as you suggested above and that didn't work, but I went and computed a size that should work (10,2,15,16,17) and it works fine.

theano/gpuarray/dnn.py
```diff
@@ -1104,7 +1136,7 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1),
                    shape_i(img, 4, fgraph) + shape_i(kerns, 4, fgraph) - 1)
         out_shp = assert_conv_shape(out_shp)
         out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*out_shp)
-        desc = GpuDnnConvDesc(border_mode='valid', subsample=(1, 1, 1),
+        desc = GpuDnnConvDesc(border_mode='valid', subsample=(1, 1, 1), dilation=(1, 1, 1),
```
HapeMask: I did, thanks.

lamblin:
According to the doc, `IMPLICIT_GEMM` ("none") for the forward, and `ALGO_0` ("none") for both backward should support dilation > 0 with 3D.
The issue could be in Theano if, for some reason, we fall back to an different, unsupported algo.
It is OK if 3d convolutions do not support dilation for now, but it is something to investigate.

notoraptor:
@lamblin For dilation in conv 3D, it seems it is a question about image shape vs kernel shape. When we increase image shape, then dilation > 0 works. It is discussed here: https://github.com/Theano/Theano/pull/5936#discussion_r116270284 .

But indeed, we might need some additional tests.

HapeMask:
Yes b2f47d6 fixed the issues with the input shape, the failure makes sense in retrospect (the dilated kernel was larger than the input and cudnn only performs valid convolutions).

theano/gpuarray/tests/test_dnn.py
```diff
@@ -636,75 +639,87 @@ def _test_conv(self, img, kerns, out, img_val, kern_vals, border_mode, conv_mode
                                         [conv_modes[0]])),
                           testcase_func_name=utt.custom_name_func)
     def test_conv(self, algo, border_mode, conv_mode):
+        # Currently only CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM (algo 'none')
+        # supports dilation > 1.
+        dilations = [(1, 1), (2, 2)] if (algo == "none" and dnn.version() >= 6000) else [(1, 1)]
```
lamblin: I think we should also allow "time_*" and "guess_*" here (which should always select `IMPLICIT_GEMM`, as you say).

lamblin:
OK, it makes sense.

lamblin:
Jenkins test this please

lamblin:
Tests are failing with `raise RuntimeError("Dilation > 1 not supported for cuDNN version < 6.").`
I think some cases should raise SkipTest depending on the version of cuDNN.

HapeMask:
Ah yes I just noticed and was changing the tests to not run if version() < 6000 but that makes more sense, I'll make them all raise SkipTest when appropriate.

notoraptor:
@HapeMask  Tests like `theano.gpuarray.tests.test_dnn.TestDnnInferShapes.test_conv3d_none*` and `theano.gpuarray.tests.test_dnn.test_conv3d_fwd` work like Python generators, so I think raising SkipTest in a test case will skip only this case and not all.

HapeMask:
@notoraptor I checked and if, for example in test_conv3d_none_valid_conv, you raise a SkipTest for the only first iteration in the _test_conv loop, all other loops are skipped.

notoraptor:
@HapeMask Indeed, you're right. The problem is that some tests are still in for-loops.

Need to restart jenkins for the last commit, @lamblin @nouiz !

notoraptor:
@HapeMask There are flake8 format errors.

```
theano/gpuarray/tests/test_dnn.py:1052:16: E126 continuation line over-indented for hanging indent

theano/gpuarray/tests/test_dnn.py:1055:21: E126 continuation line over-indented for hanging indent
```

lamblin:
Jenkins test this please

f0k:
> As a small anecdote, I patched Lasagne to use Theano's AbstractConv2D interface for dilated conv as well

@HapeMask: A PR would be welcome! Comes with a few hurdles, though. We need to change the weight layout in a way that will make it easy to adapt existing code (e.g., by providing a keyword argument switching to transposed weights) and for now we can only use the abstract conv2d interface if Theano is recent enough (0.9 is not yet a requirement for Lasagne). See https://github.com/Lasagne/Lasagne/issues/716.

HapeMask:
Sure, I can do that. I was planning to wait and make a PR after 0.9 was required since I actually removed DilatedConv2DLayer entirely and just added "dilation" as a parameter to the regular Conv2DLayer. I'll look into doing what you suggested instead when I have a chance.

f0k:
> I was planning to wait and make a PR after 0.9 was required

I see. In any case, I'd like to switch to the standard weight layout before publishing a Lasagne release. Any help with this is welcome! But let's continue the discussion over at the Lasagne issue.

> I actually removed DilatedConv2DLayer entirely and just added "dilation" as a parameter to the regular Conv2DLayer

This seems reasonable in the long term, since implementations support combining dilation with subsampling now. We could also add dilation to the TransposedConv2DLayer then, although I haven't seen any paper doing this.

Sorry for hijacking this thread -- let's continue in the Lasagne issue tracker!

PR #5947, opened 2017-05-16
Update nnet.txt

lrast:
Fixes mismatch between the formula in implemented function for binary_crossentropy in the docs


nouiz:
thanks

PR #5950, opened 2017-05-17
Test for fix in libgpuarray, plus fix in batched_dot opt

### LINKED ISSUE #5730: batched_dot fails to compile with the new gpuarray backend when the second tensor doesn't have at least 3 dimensions
For example, the following runs without errors when `device=gpu`. However, the following prints an optimization error when `device=cuda`. The program doesn't break though since the error is caught and a non-gpuarray backend is probably used. I'm using the release candidate `rel-0.9.0rc4`.
```
import theano
import theano.tensor as T
A_var = T.tensor3('A')
b_var = T.matrix('b')
c_var = T.batched_dot(A_var, b_var)
c_fn = theano.function([A_var, b_var], c_var)  # optimization error is printed when THEANO_FLAGS=device=cuda
d_var = T.batched_dot(b_var, b_var)
d_fn = theano.function([b_var], d_var)  # optimization error is printed when THEANO_FLAGS=device=cuda
```
The issues is caused because batched_dot ends up using [`local_gpua_gemmbatch`](https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1186-L1189), which assumes the input `b` to have at least 3 dimensions (since `b.shape[2]` is called).

lamblin:
This should be merged only after https://github.com/Theano/libgpuarray/pull/436
It should fix #5730 completely.

### DIFF
--- a/theano/gpuarray/tests/test_opt.py
+++ b/theano/gpuarray/tests/test_opt.py
@@ -656,3 +656,27 @@ def test_local_gpua_advanced_incsubtensor():
     w = tensor.set_subtensor(w[tensor.eq(y, 1.0).nonzero()], 100)
     w = tensor.set_subtensor(w[tensor.eq(y, -1.0).nonzero()], 0)
     theano.function([target], w)
+
+
+def test_batched_dot_lifter():
+    # The CPU Op accepts 2D and 3D inputs, as well as mixed dtypes.
+    # Make sure the lifter adds the appropriate dimshuffles and casts
+    rng = np.random.RandomState(utt.fetch_seed())
+
+    def randX(*args):
+        return rng.rand(*args).astype(theano.config.floatX)
+
+    cases = [
+        (randX(3, 5, 7), randX(3, 7)),
+        (randX(3, 5), randX(3, 5, 7)),
+        (randX(3, 5), randX(3, 5)),
+        (rng.rand(3, 5, 7).astype('float32'), randX(3, 7, 9)),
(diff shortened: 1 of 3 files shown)


nouiz:
close/reopen to trigger with the updated libgpuarray

theano/gpuarray/tests/test_opt.py
```diff
@@ -656,3 +656,27 @@ def test_local_gpua_advanced_incsubtensor():
...
+        return rng.rand(*args).astype(theano.config.floatX)
+
+    cases = [
+        (randX(3, 5, 7), randX(3, 7)),
+        (randX(3, 5), randX(3, 5, 7)),
+        (randX(3, 5), randX(3, 5)),
+        (rng.rand(3, 5, 7).astype('float32'), randX(3, 7, 9)),
+        (rng.rand(3, 5, 7).astype('float64'), randX(3, 7, 9))]
+    for x_val, y_val in cases:
+        x = tensor.TensorType(broadcastable=[s == 1 for s in x_val.shape],
```
nouiz: the broadcastable never have shape of 1. I think it is fine, you can let the code like this. Just in case you wanted to add tests.

nouiz:
I merged this PR before the PR in libgpuarray as jenkins and travis passed. This indicate that this PR don't test well the fix in libgpuarray. It is still good to merge this PR now, but if you have time, checking your test of the libgpuarray fix would be great.

slefrancois:
jenkins test this

slefrancois:
The new jenkins PR tester didn't run the GPU tests. We'll see if the daily buildbot passes.

Issue #5608, opened 2017-02-23
Feature request: top_k


khaotik:
Tensorflow have `tf.nn.top_k`

A possible workaround can be using `argsort` then take subtensor using adv. indexing, however this would introduce some overhead against a fixed and small `k` with optimized kernel.

Use cases:
 - naive ML models like KNN or K-means clustering
 - k-sparse autoencoder

nouiz:
Agreed. If you have a python function that build a Theano graph that does that, it would be useful to past it here. It would be a good starting point for test.

Do you have a profile that show how much speed up it would give in a real use case? To help access the priority. I'm not convinced that this is a problem on the CPU, but as we don't have yet a good argmax on the GPU, it would hit much more the GPU. Someone is working on that.

What values do you use and saw in the literature used?

mustgoplay:
I was recently trying to use this to get the top active neibs (simply looking at a sum of values) in B/W pictures.  Not sure if the following will help but had to play a lot in numpy to get the top_k values and then translate that into theano.  Here's a numpy example:

```
import numpy as np

neibs = np.array([[  1.7117279052734375000000e+01,   1.7255817413330078125000e+01,
    1.7325582504272460937500e+01,   1.7325582504272460937500e+01,
    1.7255817413330078125000e+01,   2.2046510696411132812500e+01,
    2.2232553482055664062500e+01,   2.2325582504272460937500e+01,
    2.2325582504272460937500e+01,   2.2232553482055664062500e+01,
    2.4651163101196289062500e+01,   2.4883720397949218750000e+01,
    2.5000000000000000000000e+01,   2.5000000000000000000000e+01,
    2.4883720397949218750000e+01,   2.4651163101196289062500e+01,
    2.4883720397949218750000e+01,   2.5000000000000000000000e+01,
    2.5000000000000000000000e+01,   2.4883720397949218750000e+01,
    2.4651163101196289062500e+01,   2.4883720397949218750000e+01,
    2.5000000000000000000000e+01,   2.5000000000000000000000e+01,
    2.4883720397949218750000e+01],
 [  1.8334333419799804687500e+01,   1.8333333969116210937500e+01,
    1.8333333969116210937500e+01,   1.6000000000000000000000e+01,
    1.2333333015441894531250e+01,   2.3333333969116210937500e+01,
    2.3333333969116210937500e+01,   2.3333333969116210937500e+01,
    2.0333333969116210937500e+01,   1.5666666984558105468750e+01,
[...]

PR #5959, opened 2017-05-18
TopKOp implementation

khaotik:
#5608; Including `topk`, `argtopk`, also with corresponding GPU kernels modified from pytorch.


khaotik:
There are close to 1k newly added test cases due to using Cartesian product of parameter combination. These tests took around 5 mins to finish on my 750 Ti, 2 mins to finish on a 4 core CPU. I don't know if this is acceptable.

nouiz:
jenkins test this

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+        assert np.allclose(xval, yvval)
+        assert yvval.dtype == xval.dtype
+        assert yival.dtype == np.dtype(idx_dtype)
+
+    @utt.parameterized.expand(chain(
+        product(
+            (16, 61, 257),
+            (1, -1, 10, -10, 'n//2', 'n-1', '-n', '1-n'),
+            ('float64', 'float16', 'int16', 'int8')),
+        ((2049, 1337, 'float64'),)))
```
nouiz: why this extra case?

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+        yvval, yival = fn(xval)
+        assert yival == np.asarray([0], dtype=idx_dtype)
+        assert np.allclose(xval, yvval)
+        assert yvval.dtype == xval.dtype
+        assert yival.dtype == np.dtype(idx_dtype)
+
+    @utt.parameterized.expand(chain(
+        product(
+            (16, 61, 257),
+            (1, -1, 10, -10, 'n//2', 'n-1', '-n', '1-n'),
```
nouiz: To lower the test time a little, I think you can remove 10 from this list.

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+        y = topk(x, k)
+        fn = theano.function([x], y)
+        # generate a all-unique array
+        xval = gen_unique_vector(size, dtype)
+        yval = fn(xval)
+        idx = slice(-k, None) if k > 0 else slice(-k)
+        goal = np.sort(xval)[idx]
+
+        print(np.sort(yval))
+        print(goal)
```
nouiz: remove print from tests.

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+        y = argtopk(x, k, idx_dtype=idx_dtype)
+        fn = theano.function([x], y)
+        # generate a all-unique array
+        xval = gen_unique_vector(size, dtype)
+        yval = fn(xval)
+        idx = slice(-k, None) if k > 0 else slice(-k)
+        goal = np.argsort(xval)[idx].astype(idx_dtype)
+
+        # due to uniqueness, we expect indices same
+        assert np.all(xval[np.sort(yval)] == xval[np.sort(goal)])
```
nouiz: numpy garantie that if 2 values are the same, we will give the first indices. Do you keep that feature? Also on GPU? It should be clear in the doc.

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+        y = argtopk(x, k, idx_dtype='int32')
+        fn = theano.function([x], y)
+        xval = np.repeat(np.random.uniform(-100., 100., size=size // 2).astype(dtype), 2)
+        xval = xval[np.random.permutation(size)]
+        yval = fn(xval)
+        idx = slice(-k, None) if k > 0 else slice(-k)
+        goal = np.argsort(xval)[idx].astype('int32')
+        print(goal)
+        print(np.argsort(xval))
+        assert np.allclose(np.sort(xval[yval]), np.sort(xval[goal]))
```
nouiz: For allclose, it is better to use:

utt.asser_allclose()

This give a better error message in case of error. This does the assert and the comparison.

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+
+    @utt.parameterized.expand(product(
+        ((1, 1), (2, 3), (17, 15), (15, 17), (11, 7, 5), (2, 3, 5, 7, 11), (2017, 5, 3)),
```
nouiz: You can remove (15, 17) from this list.

theano/tensor/tests/test_sort.py
```diff
@@ -219,3 +235,247 @@ def test_argsort_grad():
...
+            print('goal:')
+            print(np.sort(goal, axis=axis))
+            # print(np.argsort(xval))
+            assert np.all(np.sort(yval, axis=axis) == np.sort(goal, axis=axis))
+
+
+class TopKInferShapeTester(utt.InferShapeTester):
+    @utt.parameterized.expand(product(
+        ((2, 3), (15, 17), (11, 7, 5), (2, 3, 5, 7, 11), (2, 4, 3, 1)),
+        (1, -1, '(1+n)//2', 'n-1', '-n', '1-n')))
```
nouiz: You can test less cases here. In the first row, keep just 3 shapes with different ndim. On the second row, keep just 3 values.

idem bellow

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,308 @@
...
+    import pygpu
+    import pygpu.gpuarray as ga
+except ImportError as e:
+    # To make sure theano is importable
+    pass
+
+# TODO sort / argsort
+
+# TODO add runtime opt, if k==1, use max/min reduce
+#      also if k is axis size, just copy input tensor
```
nouiz: Do you know if in real use case, will k be constant most of the time? If so, we could use a Theano optimizer that when k is constant in the graph, do the optimization at the graph level. That would be much simpler then at run time.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,308 @@
...
+        if node.inputs[0].type.context.kind != b'cuda':
+            raise NotImplementedError(
+                '%s: We only have CUDA '
+                'implementation so far.' % self.__class__.__name__)
+        x, k = inps
+        inp_dtc = ga.dtype_to_typecode(node.inputs[0].dtype)
+        if not self.return_indices:
+            yv, = outs
+        else:
+            if self.return_values:
```
nouiz: you can do an elif here.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,308 @@
...
+                broadcastable=bcast,
+                context_name=ctx_name)())
+        return Apply(self, [inp, k], outs)
+
+    def get_params(self, node):
+        return node.inputs[0].type.context
+
+
+@register_opt('fast_compile')
+@op_lifter([TopKOp])
```
nouiz: On this line, add the cuda_only=True parameter.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,308 @@
...
+except ImportError as e:
+    # To make sure theano is importable
+    pass
+
+# TODO sort / argsort
+
+# TODO add runtime opt, if k==1, use max/min reduce
+#      also if k is axis size, just copy input tensor
+# TODO add opt to merge argtopk / topk, or split topk_and_argtopk when only
+#      one result is needed
```
nouiz: Note, those optimization should be in tensor/opt.py.

theano/tensor/sort.py
```diff
@@ -206,3 +218,275 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+    '''
+    sorted: bool
+        Defaults to ``False``
+
+        If True, the result array would be incremental-sorted.
+
+    only_top_kth: bool
+        Defaults to ``False``
+
+        If ``True``, will only find the exact top k-th element. The Op behaves
```
khaotik: Yes, there are use cases where only the top-k-th values are needed. An 1D array would become 0D scalar. (Hence "like" a reduction but not really is). Due to the algorithm used, I expect the GPU kernel would gain a big speed up if only the top-k-th value is needed.

theano/tensor/sort.py
```diff
@@ -206,3 +218,275 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+    # TODO c_code
+
+    __props__ = ('axis', 'return_values', 'return_indices', 'idx_dtype')
+
+    def __init__(
+            self,
+            axis=-1,
+            return_indices=False,
+            return_values=True,
+            idx_dtype='int64'):
```
khaotik: I think there are cases where user want to save GPU memory and can ensure smaller types hold it well. There are also [complaints](https://github.com/numpy/numpy/issues/2989) about numpy only using `int64` as arg*() function output dtype.

theano/tensor/sort.py
```diff
@@ -206,3 +218,275 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+        self.return_values = return_values
+        self.idx_dtype = idx_dtype
+
+    def __str__(self):
+        return '%(op)s{axis=%(axis)d}' % dict(
+            op=self.__class__.__name__, axis=self.axis)
+
+    def make_node(self, inp, k):
+        # numpy always uses float64 as output dtype for arg*() routines
+        # however, we add this option as memory is more precious on gpu
```
khaotik: I meant `int64`. That was a typo :P

theano/tensor/sort.py
```diff
@@ -206,3 +218,275 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+        if not -ndim <= self.axis < ndim:
+            raise IndexError(
+                '"axis" parameter out of range,'
+                ' expected integer within [%d, %d]' % (-ndim, ndim - 1))
+        shp[self.axis] = np.abs(node.inputs[1])
+        shp = tuple(shp)
+        return [shp for i in [self.return_values, self.return_indices] if i]
+
+
+def topk(x, k, axis=-1):
```
khaotik: Yes it is.

theano/tensor/tests/test_sort.py
khaotik: In GPU, max size for thread block is 1024. There are two kernels on GPU, a small version for length <= 1024 and a larger version. This forces GPU to test the "large" kernel.

theano/tensor/tests/test_sort.py
khaotik: I put the notice in `tensor/sort.py` around line 307.

khaotik:
Hmm ... the gradient expression is questionable. When some inputs are equal, the gradient behaves like relu at zero ... undefined.

theano/gpuarray/sort.py
khaotik: I'd say for most simple models, k is constant. Yes, it's often good enough to do graph opt. We can always runtime opt later anyway.

theano/tensor/sort.py
```diff
@@ -206,3 +218,275 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+            zv = np.partition(x, -k, axis=axis)[idx]
+            return zv
+        elif op.return_values:
+            zi = np.argpartition(x, -k, axis=axis)[idx]
+            idx2 = tuple(
+                np.arange(s).reshape(
+                    (s,) + (1,) * (ndim - i - 1)
+                    ) if i != axis else zi for i, s in enumerate(x.shape))
+            zv = x[idx2]
+            return zv, zi.astype(idx_dtype)
```
khaotik: I just did the timing, although not extensive, for a `4096 x 4096` `float64` array, finding the first 20 largest values and indices. Taking indices is almost 2x fast than sorting twice. The timing includes theano functions call overhead.

Taking indices : **0.165s**
Sorting twice : **0.318s**

The result is averaged from 10 calls, excluding the 1st call.

nouiz: thanks.

theano/tensor/sort.py
nouiz: I would think that an op line Top-k would always return only the top-k element. What do you do when it is False? You sort them? What is the difference with sort in that case.

theano/tensor/sort.py
```diff
@@ -350,14 +337,11 @@ class TopKOp(theano.Op):
     def __init__(
             self,
             axis=-1,
-            return_indices=False,
```
nouiz: You need to keep them in the \__init__ method, but you don't need to keep them in the topk function. We should never modify the \__props__ of an instance of an Op.

theano/tensor/sort.py
```diff
@@ -450,7 +459,7 @@ def topk(x, k, axis=-1, idx_dtype='int64'):
 
     Notes
     -----
-    - The returned values may not be sorted.
+    - The returned values are not sorted.
```
nouiz: Do you know what other framework do in that case? Are they given sorted?

ozancaglayan:
Hello. I quickly tested this out of curiosity to compare with `np.argpartition`. Setup is: Python 3.6, Tesla K40, CUDA-8.

I tried with small tensors as well cause I thought that maybe the reason of the speed difference is because of the data transfer between cpu-gpu at each call but its still slower than numpy. Is this expected?

```
In [27]: tx = theano.tensor.matrix('x', dtype='float32')
In [28]: idxs = theano.tensor.argtopk(tx, -12, axis=None)
In [29]: f = theano.function([tx], idxs)

In [30]: x = np.random.rand(200,200).astype('float32')

In [31]: %timeit -n5 f(x)
2.2 ms ± 54.2 µs per loop (mean ± std. dev. of 7 runs, 5 loops each)

In [32]: %timeit -n5 x.flatten().argpartition(11)[:12]
941 µs ± 85.3 µs per loop (mean ± std. dev. of 7 runs, 5 loops each)

In [33]: x = np.random.rand(200,2000).astype('float32')

In [35]: %timeit -n5 f(x)
21.4 ms ± 153 µs per loop (mean ± std. dev. of 7 runs, 5 loops each)

In [34]: %timeit -n5 x.flatten().argpartition(11)[:12]
4.86 ms ± 1.3 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)

In [36]: x = np.random.rand(2000,2000).astype('float32')

In [37]: %timeit -n5 f(x)
235 ms ± 966 µs per loop (mean ± std. dev. of 7 runs, 5 loops each)

In [38]: %timeit -n5 x.flatten().argpartition(11)[:12]
47.9 ms ± 6.64 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```

khaotik:
Hi, @ozancaglayan 

There are a lot other overhead running on GPU, data transfer, kernel compilation (1st call), data allocation etc. The code above also includes a reshape operation from flattening matrix to vector. To get pure kernel execution overhead, you can use `nvprof`:

`env CUDA_LAUNCH_BLOCKING=1 nvprof python test_script.py`

I don't have time to do a extensive timing right now. I expect GPU being faster with larger batched array. I might do a detailed timing in a few days. There could be some problems in kernel because I almost rewrite the original kernel in PyTorch.

theano/tensor/sort.py
khaotik: I took a look at PyTorch and TF, both gives sorted result by default. Didn't check out other frameworks.
Could be a good idea to finish GPU sort/argsort first.

theano/tensor/sort.py
khaotik: Example:

```python
>>> arr = np.asarray([0,2,1,3])
>>> topk(arr, 2)
array([3, 2])
>>>  # could also give array([2,3])
>>> topk(arr, 2, sorted=True)  # deterministic result
array([2, 3])
>>> topk(arr, 2, only_top_kth=True)  # sorting does not matter here
array(2)
```

nouiz:
There is travis errors: https://travis-ci.org/Theano/Theano/jobs/237928562#L896

And gradients errors: https://travis-ci.org/Theano/Theano/jobs/237928586#L7667

For the gradient, take care of the inputs values. They must be spaced far enough that the delta added for the gradient estimation don't change the output of the forward. There is code related to that in the tests for max gradient.

khaotik:
Grad test should be fixed by now. Also managed to get some speedup for the GPU kernel. The GPU kernel is only useful when array is very large. Compared to PyTorch, this kernel is faster on small arrays but falls behind on larger ones:

| shape |PyTorch CUDA | Theano CUDA | `np.argpartition` |
| --- | --- | --- | --- |
(256,) | 13.5us | 13.0us | **5.42us**
(1024,) | 21.6us | 22.4us | **7.86us**
(4096,) | 46us | 68us | **17.4us**
(65536,) | 0.621ms | 1.09ms | **0.485ms**
(256, 256) | 0.218ms | **0.182ms** | 0.69ms
(1024,1024) | 4.55ms | **4.51ms** | 10.2ms
(4096, 4096) | **37.2ms** | 42.9ms | 159ms

I'm re-implementing PyTorch kernels because I'd like to learn more about CUDA programming. However these GPU kernels clearly still have ground for optimization. Any tips?

(The tests are done with `float32` dtype and `k=20`, on a 750 Ti with CUDA 8.0, CPU is 4-core i5-2400)

nouiz:
Someone arrived yesterday at the lab and will work on gpu sort. If the
other framework does sort by default, then I think we should make that our
default too, but raise an error. This way, when the sort is implemented, it
will work and we won't break the user interface.

Does your GPU timing include the transfer time? If so, can you do timing
without transfer? Most of the time, the data will already be on the GPU, so
there won't be transfer.

nouiz:
For idea to speed up, you could check how we do cumsum on the GPU. Maybe it
would give you speed up idea.

khaotik:
The timing do not contain transfer, it's kernel execution time from `nvprof`

theano/tensor/sort.py
```diff
@@ -206,3 +219,311 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+            pzv = output_storage[0]
+            pzi = output_storage[1]
+            pzv[0], pzi[0] = _topk_py_impl(
+                self, x, k, axis, node.outputs[1].dtype)
+        else:
+            pzi = output_storage[0]
+            pzi[0] = _topk_py_impl(self, x, k, axis, node.outputs[0].dtype)
+
+    def infer_shape(self, node, inp_shapes):
+        _check_tensor_is_scalar(node.inputs[1])
```
nouiz: Move that check in the make_node()

theano/tensor/sort.py
```diff
@@ -206,3 +219,311 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+        shp = list(inp_shapes[0])
+        shp[self.axis] = np.abs(node.inputs[1])
+        shp = tuple(shp)
+        return [shp for i in [self.return_values, self.return_indices] if i]
+
+    def L_op(self, inputs, outputs, out_grads):
+        x, k = inputs
+        k_grad = grad_undefined(self, 1, k, 'topk: k is not differentiable')
+        if not (self.return_indices, self.return_values):
```
nouiz: this if will always be False. not (non empty tuple) will evaluate like not True, as non-empty tuple always eval to True. You must do:

if not self.return... and not self.return...:

theano/tensor/sort.py
```diff
@@ -206,3 +219,311 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+            x_shp = theano.tensor.shape(x)
+            z_grad = out_grads[0]
+            ndim = x.ndim
+            axis = self.axis % ndim
+            grad_indices = [
+                arange(x_shp[i]).dimshuffle([0] + ['x'] * (ndim - i - 1))
+                if i != axis else outputs[-1] for i in range(ndim)]
+            x_grad = x.zeros_like(dtype=z_grad.dtype)
+            x_grad = theano.tensor.advanced_set_subtensor(
+                x_grad, z_grad, *grad_indices)
```
nouiz: I think this case can be simplified and can give more optimized code like this:

x_grad = x.zeros_like(dtype=z_grad.dtype)
grad_indices = [slice(None) if i != axis else outputs[-1] for i in range(ndim)]
x_grad = set_subtensor(x_grad[*grad_indices], z_grad, )

Why this would be more optimized? Theano will sometimes reuse advancedincsubtensor1 in such case as it is faster. Your version force the advancedincsubtensor that isn't as much optimized.

theano/tensor/sort.py
```diff
@@ -206,3 +219,311 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+    Tensor variable with same dtype as `x`.
+
+    Notes
+    -----
+    - The returned values are not sorted.
+
+    """
+    if axis is None:
+        x = theano.tensor.flatten(x)
+        axis = -1
```
nouiz: I would set the axis to 0 in that case. It would make the modulo not needed and this is more clear when the op get printed to the user I think. Idem for the others.

theano/tensor/sort.py
```diff
@@ -206,3 +219,311 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+        Defaults to ``False``
+
+        If ``True``, will only find one exact top k-th element on given axis.
+
+    '''
+
+    # TODO c_code
+    # TODO add opt, if k==1, use max/min reduce
+    #      also if k is axis size, just copy input tensor
+    # TODO add opt to merge argtopk / topk, or split topk_and_argtopk when only
```
nouiz: As by default we need the forward node to always return both result, we need in this PR the opt that remove the output not used. Here is code that should be close to working that does that. To put in tensor/opt.py

```
@register_useless('fast_compile')
@gof.local_optimizer([TopK])
def local_useless_topk(node):
    """ Remove output of TopK that aren't used.
    """
    if not isinstance(node.op, TopK):
        return
    ret_val = False
    ret_ind = False
    if node.op.return_values and node.outputs[0].clients:
       ret_val = True
    if node.op.return_indices and node.outputs[-1].clients:
       ret_ind = True
    out = TopK(dtype=..., return...)(node.inputs[0], return_list=True)
    ret = []
    if node.op.return_values and node.outputs[0].clients:
        ret.append(out.pop(0))
    elif node.op.return_values:
        ret.append(None)
    if node.op.return_indices and node.outputs[-1].clients:
        ret.append(out.pop(0))
    elif node.op.return_indices:
        ret.append(None)
    return ret
```

nouiz:
Ok. I understand that you want to do GPU code. What about this, make a
first version that we merge with the pytorch GPU code, then if you get
something better we merge this in another PR? This way we will be sure to
be at least as fast as pytorch.

theano/tensor/sort.py
```diff
@@ -206,3 +219,311 @@ def argsort(a, axis=-1, kind='quicksort', order=None):
...
+
+    Notes
+    -----
+    - By default, this Op give two outputs: values and indices. However optimizer may
+      remove a certain output if not needed.
+
+    - Computing gradient is only possible when both values and indices are computed in
+      forward pass.
+
+    - If the top-k-th value is not unique, we cannot guarantee the output indices being
```
khaotik: numpy only mentions `np.[arg]partition` is not *stable*, so I guess the disclaimer applies for both device.

theano/tensor/sort.py
khaotik: I did a few tests. For a local opt, replacement must be same type as old output, and can't be `None`.

What's the right fix for this? Should I just plug a orphan variable?

theano/tensor/sort.py
nouiz: thanks for checking.

nouiz:
I pushed a fix to this PR for the optimizer.

khaotik:
Sorry for the delay, was kinda occupied and had to deal with some bizarre bugs.

For large arrays, I fully ported the PyTorch kernel. The small kernel seems faster so I kept it there.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+            yv, yi = outs
+        else:
+            yi, = outs
+        out_dtype_s = self.idx_dtype
+        out_dtc = ga.dtype_to_typecode(out_dtype_s)
+        fail = sub['fail']
+        ctx = sub['params']
+        k_dtype = node.inputs[1].type.dtype_specs()[1]
+        MAX_TPB = 1024  # max threads per block
+        WARP_SIZE = 32
```
abergeron: Don't hardcode these things.  Query them from the context in the C code.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+        return Apply(self, [inp, kth], outs)
+
+    def get_params(self, node):
+        return node.inputs[0].type.context
+
+
+@register_opt('fast_compile')
+@op_lifter([TopKOp], cuda_only=True)
+@register_opt2([TopKOp], 'fast_compile')
+def local_gpua_topkop(op, ctx_name, inputs, outputs):
```
abergeron: This optimization should avoid the replacement if not on a cuda context.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+        ndim = node.inputs[0].ndim
+        reordered_axes = list(range(ndim))
+        axis = self.axis % ndim
+        del(reordered_axes[axis])
+        reordered_axes = [axis] + reordered_axes
+        dims = ''.join('(void*)(dims+%d), ' % i for i in reordered_axes[1:])
+        prep_output = ''
+        if self.return_values:
+            def_dvstrides = 'const ssize_t *dvstrides = PyGpuArray_STRIDES(%s)' % yv
+            params_dv = '(void*)(%s->ga.data),\n' % yv
```
abergeron: You also need to handle array offset for inputs and outputs.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+        &%(yv)s, %(ndim)d, odims,
+        %(inp_dtc)s, GA_C_ORDER, %(ctx)s)) {
+        %(fail)s;
+    }\n''' % locals()
+        else:
+            def_dvstrides = params_dv = ''
+
+        if self.return_indices:
+            def_distrides = 'const ssize_t *distrides = PyGpuArray_STRIDES(%s)' % yi
+            params_di = '(void*)(%s->ga.data),\n' % yi
```
abergeron: Array offset here too

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+
+    %(def_dvstrides)s;
+    %(def_distrides)s;
+    const ssize_t *sstrides = PyGpuArray_STRIDES(%(x)s);
+    void* args[] = {
+        %(dims)s
+        %(params_dv)s
+        %(params_di)s
+        (void*)(&k_),
+        (void*)(%(x)s->ga.data),
```
abergeron: Array offset.

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+        %(sstrides)s,
+        (void*)(dims+%(axis)d),
+    };
+
+    int err;
+    if (dims[%(axis)d] > PY_SSIZE_T_MAX) {
+        PyErr_SetString(
+            PyExc_ValueError,
+            "topk: array size on specified axis is too large, should be less than PY_SSIZE_T_MAX.");
+        %(fail)s;
```
abergeron: This situation is not an error for GpuArrays (although I really doubt you will encounter any array that big).

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+    } else if (dims[%(axis)d] < odims[%(axis)d]) {
+        PyErr_SetString(
+            PyExc_ValueError,
+            "topk: kth cannot be larger than the size of specified axis %(axis)d");
+        %(fail)s;
+    }
+    %(prep_output)s
+
+    size_t blk[6];
+    size_t *grd = blk+3;
```
abergeron: This is just obscuring intent.  Please declare two arrays.

Also you never use anything other than the first element so you can declare them as size 1 (or as scalars).

theano/gpuarray/sort.py
```diff
@@ -0,0 +1,314 @@
...
+    if (dims[%(axis)d] > PY_SSIZE_T_MAX) {
+        PyErr_SetString(
+            PyExc_ValueError,
+            "topk: array size on specified axis is too large, should be less than PY_SSIZE_T_MAX.");
+        %(fail)s;
+    } else if (dims[%(axis)d] > (1u << 31)) {
+        blk[0] = %(MAX_TPB)d;
+        err = GpuKernel_call(
+            &k_topk_dense_xlarge%(nodename)s, 3,
+            grd, blk, 0, args);
```
abergeron: Since all the values above 0 are 1 in grd and blk you can just say the calls are 1d instead of 3d.

abergeron:
I know that the optimisation change was made by @nouiz, but I strongly disagree with it and it should be removed from this PR.

khaotik:
Thanks for the very detailed review @abergeron. Current I'm kinda occupied so there would be a delay before I can check them.

guolinke:
@khaotik any updates ? looking forward to this feature .

theano/gof/opt.py
```diff
@@ -1353,21 +1353,27 @@ def transform(self, node):
...
                     continue
-                else:
-                    if self.profile:
-                        self.node_created[opt] += len(graph.ops(fgraph.variables, new_repl))
-                        self.applied_true[opt] += 1
-                    break  # break from the for loop over optimization.
+                if isinstance(new_repl, (tuple, list)):
+                    new_vars = new_repl
+                else:  # It must be a dict
+                    new_vars = list(new_repl.values())
```
nouiz: We don't care of the order of the dict. This is just to count the new variable introduced in the graph. We don't use it to change the graph. Line 1371 return new_repl

theano/gof/opt.py
```diff
@@ -1353,21 +1353,27 @@ def transform(self, node):
...
+                break  # break from the for loop over optimization.
             if not new_repl:  # No optimization applied in the last iteration
                 return repl
-            # only 1 iteration or we are at the start of the graph.
-            if not self.apply_all_opts or not new_repl[0].owner:
+            # only 1 iteration
+            if not self.apply_all_opts:
+                return new_repl
+            if not new_vars[0].owner:
+                # We are at the start of the graph.
```
nouiz: This LocalOptGroup optimization apply a list of optimization on one node. When one opt is applied, it optionaly, can try to optimize that new node imediatly. This remove sequential change to the FunctionGraph. It does mostly an equilibrium on that node.

If one optimization return as variable an input of the graph, we can't get the node that generated it. So we can't continue the "Equilibrium". This if is for that, stop when we can't continue and don't crash.

theano/tensor/opt.py
```diff
@@ -7510,3 +7511,35 @@ def local_merge_alloc(node):
...
+        # both false -> let pruner handle
+        return False
+
+    old_output = node.outputs[ret_idx]
+    new_output = TopKOp(
+        axis=op.axis,
+        idx_dtype=op.idx_dtype,
+        return_values=ret_val,
+        return_indices=ret_idx)(x, k)
+    return {old_output: new_output}
```
nouiz: Using dict make the code much cleaner. I tried to do it otherwise and it was getting much longer.

nouiz:
Note, we are doing an Theano 0.10 release. It would be great to finish this. I don't know if we can finish it ourself, so if someone can do it, it would be great!

theano/gof/opt.py
abergeron: I just learned that optimizer have support for handling dict replacements from LocalOptimizer by reading the code.  I don't believe I ever saw a single optimization in Theano use that.  So this won't replace random things.

However, with the non-dict code we always try again with the owner of the first output.  With the dict code, it may be one or the other, depending in which order they end up in the list.

Also the commit messages made it sound like this was necessary to remove nodes from the graph, but it's not since optimizers can include `None` in the list to remove certain outputs.

khaotik:
@nouiz What's ETA for 0.10? I can probably reboot this in 2~4 weeks.

theano/gpuarray/sort.py
khaotik: I would still hardcode `WARP_SIZE`, reasons:

1. The kernel code relies on the fact that warp size is constant 32. Making this value dynamic may drop performance and adding complexity.

2. This value is constant across all CUDA GPUs. AFAIK many code just assume it's constant 32. NVIDIA probably won't change it in short term.

theano/gpuarray/sort.py
khaotik: I assume that's the purpose of `cuda_only=True`?

theano/gpuarray/sort.py
```diff
@@ -168,7 +168,7 @@ def c_code(self, node, nodename, inps, outs, sub):
         prep_output = ''
         if self.return_values:
             def_dvstrides = 'const ssize_t *dvstrides = PyGpuArray_STRIDES(%s)' % yv
-            params_dv = '(void*)(%s->ga.data),\n' % yv
+            params_dv = '(void*)((char*)(%s->ga.data) + (%s->ga.offset)),\n' % (yv, yv)
```
abergeron: PyGpuArray_DATA() is cleaner

theano/gpuarray/sort.py
abergeron: Yes, you're right.

abergeron:
You still have a flake8 error in sort.py

nouiz:
jenkins test this

PR #5960, opened 2017-05-18
Fix the opt crash regression introduced by gh-5950.

nouiz:


lamblin:
Failures on Jenkins seem unrelated: 'libcudnn.so.5: cannot open shared object file: No such file or directory'

nouiz:
I updated this PR to have the cudnn version in the key more frequently.

If I got all missing places, all tests should pass. I think this is better
then deleting the cache.

PR #5963, opened 2017-05-19
added polygamma

ferrine:
I've tried to add polygamma, hope it's all right. It adds m'th derivative of gammaln function

### DIFF
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -301,7 +301,7 @@ class Psi(UnaryScalarOp):
     def grad(self, inp, grads):
         x, = inp
         gz, = grads
-        return [gz * polygamma(3, x)]
+        return [gz * polygamma(1, x)]
 
     def c_support_code(self):
         return (
@@ -552,11 +552,14 @@ class Polygamma(ScalarOp):
     """
     nin = 2
 
+    @staticmethod
+    def st_impl(m, x):
+        scipy.special.polygamma(m, x)
+
     def impl(self, m, x):
         if imported_scipy_special:
-            return scipy.special.polygamma(m, x)
+            return self.st_impl(m, x)
         else:
-            # raises an error
             super(Polygamma, self).impl(x)
 
     def grad(self, inp, grads):
@@ -565,4 +568,4 @@ class Polygamma(ScalarOp):
         # derivative is again polygamma
         return [grad_not_implemented(self, 0, m), gz * polygamma(m+1, x)]
 
-polygamma = Polygamma(upgrade_to_float, name='trigamma')
+polygamma = Polygamma(upgrade_to_float, name='polygamma')
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -553,6 +553,8 @@ class Polygamma(UnaryScalarOp):
     __props__ = ('m', )
 
     def __init__(self, m, *args, **kwargs):
+        if not isinstance(m, int) and m >= 0:
+            raise ValueError('m should be an int >= 0')
         super(Polygamma, self).__init__(*args, **kwargs)
         self.m = m
 
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -9,7 +9,7 @@ from theano.gradient import grad_not_implemented
 from theano.scalar.basic import (UnaryScalarOp, BinaryScalarOp,
                                  exp, upgrade_to_float,
                                  upgrade_to_float64,
-                                 float_types)
+                                 float_types, ScalarOp)
 from theano.scalar.basic import (upgrade_to_float_no_complex,
                                  complex_types, discrete_types,
                                  upcast)
@@ -301,7 +301,7 @@ class Psi(UnaryScalarOp):
     def grad(self, inp, grads):
         x, = inp
         gz, = grads
-        return [gz * trigamma(x)]
+        return [gz * polygamma(3, x)]
 
     def c_support_code(self):
         return (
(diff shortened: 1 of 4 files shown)


theano/scalar/basic_scipy.py
```diff
@@ -542,3 +544,30 @@ def grad(self, inp, grads):
...
+            return scipy.special.polygamma(self.m, x)
+        else:
+            # raises an error
+            super(Polygamma, self).impl(x)
+
+    def grad(self, inp, grads):
+        x, = inp
+        gz, = grads
+        # derivative is again polygamma
+        poly = Polygamma(self.m+1, upgrade_to_float, name='_%dgamma' % (self.m+1))
```
ferrine: Is it the best way to provide derivative?

ferrine:
```
ERROR (theano.gof.opt): SeqOptimizer apply <theano.tensor.opt.InplaceElemwiseOptimizer object at 0x7f5abc884208>
  File "/home/ferres/dev/theano/theano/gof/opt.py", line 241, in apply
  File "/home/ferres/dev/theano/theano/gof/opt.py", line 87, in optimize
  File "/home/ferres/dev/theano/theano/tensor/opt.py", line 360, in apply
    *node.inputs, **dict(return_list=True))
  File "/home/ferres/dev/theano/theano/gof/op.py", line 615, in __call__
  File "/home/ferres/dev/theano/theano/tensor/elemwise.py", line 562, in make_node
    DimShuffle, *inputs)
  File "/home/ferres/dev/theano/theano/tensor/elemwise.py", line 504, in get_output_info
    for i in inputs])
  File "/home/ferres/dev/theano/theano/scalar/basic.py", line 1044, in make_node
    for input in inputs])]
  File "/home/ferres/dev/theano/theano/scalar/basic.py", line 1063, in output_types
    "Cannot calculate the output types for %s" % self)
NotImplementedError: Cannot calculate the output types for Polygamma{m=<theano.scalar.basic.transfer_type object at 0x7f5a4e7606a0>}
```
I have the following error, what can I do with it?

theano/scalar/basic_scipy.py
nouiz: I'm not sure. If we want to compute the polygamma of x and the derivative. Can you reuse some of the forward computation? Is there a numpy/scipy function that does that?

The way you do it looks fine. What would you think could be made better?

theano/tensor/tests/test_basic.py
```diff
@@ -1847,6 +1849,24 @@ def test_py_c_match():
     inplace=True,
     skip=skip_scipy)
 
+TrigammaTester = makeBroadcastTester(
+    op=tensor.trigamma,
+    expected=expected_trigamma,
+    good=_good_broadcast_unary_normal,
+    grad=_grad_broadcast_unary_normal,
+    eps=2e-10,
```
nouiz: why such a low eps?

theano/scalar/basic_scipy.py
```diff
@@ -298,8 +298,10 @@ def impl(self, x):
         else:
             super(Psi, self).impl(x)
 
-    def grad(self, inputs, outputs_gradients):
-        raise NotImplementedError()
+    def grad(self, inp, grads):
```
nouiz: This will request that you expect Psi test to test the gradient.

theano/scalar/basic_scipy.py
nouiz: Why the underscore at the start of the name?

theano/tensor/basic.py
```diff
@@ -2338,6 +2338,12 @@ def conj(z):
 def complex_from_polar(abs, angle):
     """Return complex-valued tensor from polar coordinate specification."""
 
+
+@_scal_elemwise
+def trigamma(x):
```
nouiz: why not also expose polygamma directly?

theano/scalar/basic_scipy.py
ferrine: I think I do not correctly understand what forward computations I can reuse. [Wikipedia](https://en.wikipedia.org/wiki/Polygamma_function) says that I can reuse the same polygamma with m+1 order

theano/tensor/tests/test_basic.py
ferrine: Just copy paste from gamma. BTW
```
%env THEANO_FLAGS=device=cpu
from theano import theano, tensor as tt
import numpy as np
import scipy.special as sp
x = tt.matrix()
xx = tt.trigamma(x)
ttp = theano.function([x], xx)
M = np.random.randn(10, 10).astype('float32')
np.testing.assert_allclose(ttp(M), sp.polygamma(3, M))
```
Generats
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-48-ddec2d7ec72b> in <module>()
      7 ttp = theano.function([x], xx)
      8 M = np.random.randn(10, 10).astype('float32')
----> 9 np.testing.assert_allclose(ttp(M), sp.polygamma(3, M))

/home/ferres/.pyenv/versions/3.5.0/lib/python3.5/site-packages/numpy/testing/utils.py in assert_allclose(actual, desired, rtol, atol, equal_nan, err_msg, verbose)
   1409     header = 'Not equal to tolerance rtol=%g, atol=%g' % (rtol, atol)
   1410     assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
-> 1411                          verbose=verbose, header=header, equal_nan=equal_nan)
   1412 
   1413 

/home/ferres/.pyenv/versions/3.5.0/lib/python3.5/site-packages/numpy/testing/utils.py in assert_array_compare(comparison, x, y, err_msg, verbose, header, precision, equal_nan)
    794                                 names=('x', 'y'), precision=precision)
    795             if not cond:
--> 796                 raise AssertionError(msg)
    797     except ValueError:
    798         import traceback

AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

(mismatch 4.0%)
 x: array([[  2.187731e+02,   1.666113e+04,   1.999472e+02,   1.203099e+01,
          4.913559e+04,   1.569319e+01,   4.008031e+03,   2.261310e+00,
          2.162573e+04,   1.934545e+02],...
 y: array([[  2.187731e+02,   1.666112e+04,   1.999472e+02,   1.203099e+01,
          4.913559e+04,   1.569319e+01,   4.008031e+03,   2.261310e+00,
          2.162573e+04,   1.934545e+02],...
```

theano/scalar/basic_scipy.py
ferrine: I don't know how to name a function correctly

theano/tensor/basic.py
ferrine: What is the right way to create op with parameters?

theano/tensor/tests/test_basic.py
ferrine: 4% is a mystery for me

ferrine:
I got stuck with gradient in one of my models in pymc3. This error is from one of my notebooks

ferrine:
@nouiz TrigammaTester works now(locally). Do I need something else?

theano/tensor/basic.py
nouiz: If you make m an input instead of a parameter, it would work. I'm not sure what the error with wrong value come from. It could be because you create the first scalar op with a props. Maybe it isn't supported somewhere.

Is there a reason why m was make a props instead of an inputs?

nouiz:
travis must pass to be able to merge. There is this easy error to fix:

https://travis-ci.org/Theano/Theano/jobs/236324460#L900

and there is this more hard to fix:

https://travis-ci.org/Theano/Theano/jobs/236324457#L2004

ferrine:
That's done. I've also added test to Psi derivative

Issue #5621, opened 2017-02-24
grouped convolution


nouiz:
We don't have it. The corrmm implementation in Theano have reference to it from the code we moved to Theano. So there is probably small change to Theano to support it.

f0k:
The [TensorFlow implementation](https://github.com/tensorflow/tensorflow/blob/d42facc3cc9611f0c9722c81551a7404a0bd3f6b/tensorflow/core/kernels/depthwise_conv_op_gpu.cu.cc) is relatively direct and relies on the read-only cache (via __ldg) to be fast enough. The [Torch implementation](https://github.com/soumith/cudnn.torch/blob/f746b79/SpatialConvolution.lua#L193) just calls cuDNN in a loop, as does the [Caffe cuDNN implementation](https://github.com/BVLC/caffe/blob/80f44100e19fd371ff55beb3ec2ad5919fb6ac43/src/caffe/layers/cudnn_conv_layer.cu#L18). The [Caffe gemm implementation](https://github.com/BVLC/caffe/blob/bbc4e57/src/caffe/layers/base_conv_layer.cpp#L335) also does multiple convolutions in a loop.

So I guess this would need to be added to the abstract convolution interface, and then most (if not all) existing implementations will have to support it by adding a loop and some indexing tricks somewhere. Not a small project, but there are many use cases.

nouiz:
What about the inputs/outputs number of dimensions? Are they the same or it
request an extra dimensions?

https://github.com/Theano/Theano/issues/5580

tell that they have one for CPU. I did a quick discussion of this with
@lamblin and he was thinking a new abstract group conv op would be better
due to the different handling of the dimensions.

f0k:
> What about the inputs/outputs number of dimensions? Are they the same or it request an extra dimensions?

They are the same in all implementations I've seen. The number of input channels and output channels has to be divisible by the number of groups. This could be implemented as an extra dimension instead, but it seems nobody did it that way.

> he was thinking a new abstract group conv op would be better due to the different handling of the dimensions.

This would lead to a lot of code duplication. Grouped convolution is a generalization of what we have now, so we'd only need an additional parameter `num_groups` for the existing Ops and functions that defaults to `1`.

lamblin:
It was natural to me to think of the number of groups as an additional dimension, that's why I was leaning to another Abstract Op (like for Conv3d).
[...]

PR #5973, opened 2017-05-23
Grouped Convolution

affanv14:

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -512,7 +512,7 @@ def conv2d(input,
            border_mode='valid',
            subsample=(1, 1),
            filter_flip=True,
-           filter_dilation=(1, 1),num_groups=1):
+           filter_dilation=(1, 1), num_groups=1):
     """This function will build the symbolic graph for convolving a mini-batch of a
     stack of 2D inputs with a set of 2D filters. The implementation is modelled
     after Convolutional Neural Networks (CNN).
@@ -527,7 +527,7 @@ def conv2d(input,
                              border_mode=border_mode,
                              subsample=subsample,
                              filter_flip=filter_flip,
-                             filter_dilation=filter_dilation,num_groups=num_groups)
+                             filter_dilation=filter_dilation, num_groups=num_groups)
     return conv_op(input, filters)
 
 
@@ -1396,7 +1396,7 @@ class BaseAbstractConv(Op):
 
     def __init__(self, convdim,
                  imshp=None, kshp=None, border_mode="valid",
-                 subsample=None, filter_flip=True, filter_dilation=None,num_groups=1):
+                 subsample=None, filter_flip=True, filter_dilation=None, num_groups=2):
 
         self.convdim = convdim
         if convdim not in (2, 3):
@@ -1462,7 +1462,7 @@ class BaseAbstractConv(Op):
             raise ValueError("num_groups must have value greater than zero")
         elif num_groups > 1 and convdim == 3:
             raise ValueError("grouped convolution not supported for 3D convolutions")
-        self.num_groups=num_groups
+        self.num_groups = num_groups
 
     def do_constant_folding(self, node):
         # Disable constant folding since there is no implementation.
@@ -1489,7 +1489,7 @@ class BaseAbstractConv(Op):
             raise NotImplementedError(
                 'flops not implemented for convdim={}', self.convdim)
 
-    def conv(self, img, kern, mode="valid", dilation=1,num_groups=1):
+    def conv(self, img, kern, mode="valid", dilation=1, num_groups=1):
         """
         Basic slow Python 2D or 3D convolution for DebugMode
         """
@@ -1510,7 +1510,7 @@ class BaseAbstractConv(Op):
                                                                  self.convdim))
 
         out_shape = get_conv_output_shape(img.shape, kern.shape,
-                                          mode, [1] * self.convdim, dilation,num_groups)
+                                          mode, [1] * self.convdim, dilation, num_groups)
 
         out = np.zeros(out_shape, dtype=img.dtype)
         dil_kern_shp = kern.shape[:-self.convdim] + tuple(
@@ -1524,8 +1524,8 @@ class BaseAbstractConv(Op):
         if self.convdim == 2:
             val = _valfrommode(mode)
             bval = _bvalfromboundary('fill')
-            input_channel_offset=img.shape[1]//self.num_groups
-            output_channel_offset=kern.shape[0]
+            input_channel_offset = img.shape[1] // self.num_groups
+            output_channel_offset = kern.shape[0]
 
             with warnings.catch_warnings():
                 warnings.simplefilter('ignore', np.ComplexWarning)
@@ -1534,9 +1534,9 @@ class BaseAbstractConv(Op):
                         for n in xrange(output_channel_offset):
                             for im0 in xrange(input_channel_offset):
                                 # some cast generates a warning here
-                                out[b, g * output_channel_offset + n, ...] += _convolve2d(img[b, g *input_channel_offset + im0, ...],
-                                                              dilated_kern[n, g * input_channel_offset + im0, ...],
-                                                              1, val, bval, 0)
+                                out[b, g * output_channel_offset + n, ...] += _convolve2d(img[b, g * input_channel_offset + im0, ...],
+                                                                                        dilated_kern[n, g * input_channel_offset + im0, ...],
+                                                                                        1, val, bval, 0)
         elif self.convdim == 3:
             for b in xrange(img.shape[0]):
                 for n in xrange(kern.shape[0]):
@@ -1562,13 +1562,13 @@ class AbstractConv(BaseAbstractConv):
                  border_mode="valid",
                  subsample=None,
                  filter_flip=True,
-                 filter_dilation=None,num_groups=1):
+                 filter_dilation=None, num_groups=1):
         super(AbstractConv, self).__init__(convdim=convdim,
                                            imshp=imshp, kshp=kshp,
                                            border_mode=border_mode,
                                            subsample=subsample,
                                            filter_flip=filter_flip,
-                                           filter_dilation=filter_dilation,num_groups=num_groups)
+                                           filter_dilation=filter_dilation, num_groups=num_groups)
 
     def make_node(self, img, kern):
         # Make sure both inputs are Variables with the same Type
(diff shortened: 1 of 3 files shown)


affanv14:
Is the implementation alright?

Also I am not able to replicate the error in the test on my computer. It's passing for me.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1396,7 +1397,7 @@ class BaseAbstractConv(Op):
 
     def __init__(self, convdim,
                  imshp=None, kshp=None, border_mode="valid",
-                 subsample=None, filter_flip=True, filter_dilation=None):
+                 subsample=None, filter_flip=True, filter_dilation=None, num_groups=2):
```
nouiz: Here, the default value is different then at other places. This isn't good. The default should be the same everywhere expect in rare condition. It should also default to the old behavior, like without groups

theano/tensor/nnet/abstract_conv.py
```diff
@@ -33,7 +33,7 @@
 
 def get_conv_output_shape(image_shape, kernel_shape,
                           border_mode, subsample,
-                          filter_dilation=None):
+                          filter_dilation=None, num_groups=1):
```
nouiz: Add num_groups in the docstring parameters.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -66,7 +66,8 @@ def get_conv_output_shape(image_shape, kernel_shape,
     """
     bsize, imshp = image_shape[0], image_shape[2:]
     nkern, kshp = kernel_shape[0], kernel_shape[2:]
-
```
nouiz: I would be more strict here in the condition to make the graph simpler when nkern is symbolic. I think this condition would do it: "if nkern is not None and num_groups > 1".

This will probably fix the error in travis.

Some explanation of the travis error. Some part of Theano are not garanted to finish. So we added some check so if it iterate too much, it stop iterating instead of potentially never finishing. The threshold where we do this detection can be sensitive in some corner case. You hit such a case. This extra multiplication caused one optimization to apply more frequently and hit the threshold.

theano/tensor/nnet/abstract_conv.py
affanv14: Sorry, that way a Typo. I fixed it now.

affanv14:
I think i have finished the initial python implementation. Though tests still need to be implemented.

The `num_input_channels` is split into `num_groups` distinct groups, the forward pass is simple enough but the backward passes require the matrices to be transformed (through some reshapes and transposes) which lead to decrease in efficiency. This may be avoided by specifically passing the offsets to the `conv` function defined `BaseAbstractConv` for each of the three passes, but this function is also used by `Abstractconv3D` as well, so the number of parameters passed to this function was kept to a minimum. 
Any suggestions for this?

The gemm implementations have more finer control while handling convolutions, so there should not be much of a problem there.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -512,7 +515,7 @@ def conv2d(input,
            border_mode='valid',
            subsample=(1, 1),
            filter_flip=True,
-           filter_dilation=(1, 1)):
+           filter_dilation=(1, 1), num_groups=1):
```
nouiz: put num_groups on its own line to follow the code around.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -527,7 +530,7 @@ def conv2d(input,
                              border_mode=border_mode,
                              subsample=subsample,
                              filter_flip=filter_flip,
-                             filter_dilation=filter_dilation)
+                             filter_dilation=filter_dilation, num_groups=num_groups)
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1458,6 +1461,11 @@ def __init__(self, convdim,
...
             raise ValueError("filter_dilation must have {} elements".format(convdim))
         self.filter_dilation = tuple(filter_dilation)
+        if num_groups < 1:
+            raise ValueError("num_groups must have value greater than zero")
+        elif num_groups > 1 and convdim == 3:
+            raise ValueError("grouped convolution not supported for 3D convolutions")
+        self.num_groups = num_groups
 
     def do_constant_folding(self, node):
         # Disable constant folding since there is no implementation.
```
nouiz: note that the flops method bellow will need to be adjusted for groups. But do this later, when the rest work. For now, just make it raise an error when groups isn't 1.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1622,7 +1635,7 @@ def perform(self, node, inp, out_):
             img = new_img
         if not self.filter_flip:
             kern = kern[(slice(None), slice(None)) + (slice(None, None, -1),) * self.convdim]
-        conv_out = self.conv(img, kern, mode="valid", dilation=self.filter_dilation)
+        conv_out = self.conv(img, kern, mode="valid", dilation=self.filter_dilation, num_groups=self.num_groups)
         conv_out = conv_out[(slice(None), slice(None)) +
                             tuple(slice(None, None, self.subsample[i])
                                   for i in range(self.convdim))]
```
nouiz: The method R_op bellow will need update. Do that later when the the rest work.

But make it raise an error now if groups isn't 1.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1517,18 +1525,23 @@ def conv(self, img, kern, mode="valid", dilation=1):
                      ] = kern
 
         if self.convdim == 2:
+            if img.shape[1] % self.num_groups != 0:
+                raise ValueError("Number of input channels must be divisible by num_groups")
             val = _valfrommode(mode)
             bval = _bvalfromboundary('fill')
+            input_channel_offset = img.shape[1] // self.num_groups
+            output_channel_offset = kern.shape[0]
```
nouiz: The value don't make sense with how you use it. You use it on line 1539 and 1542. The change for line 1539 don't change the behavior.

Where did you took that implementation? I don't think it does the right think.

affanv14: I did informally test this implementation by taking some input and kernel and splitting them into separate groups and then using normal convolutions. Then i concatenated the outputs and checked them against the outputs i got in grouped convolutions by giving the same input and kernel.  The difference was 1e-15 or less.
I checked the same for the backward passes as well.

yes,  line 1539 doesnt change behavior. I thought since the same value was being fetched, substituting them in both places would make them more readable. I was mistaken there.

affanv14: Adding to this, I think there were two ways for this to be implemented.

The first method takes input of shape `(batches, in_channels, ih, iw )` and kernels of shape `(num_filters, in_channels, kh,  kw)` and `num_groups` and splits `in_channels` into `num_groups` different groups, giving inputs of each group shape 
`(batches, in_channels // num_groups , ih, iw)` and kernel of shape `(num_filters, in_channels // num_groups , kh, kw)`. The final output of shape will be` (batches, num_filters * num_groups , oh, ow)` .
This method would require the user be aware of the fact that total number of output channels is `num_filters * num_groups`.  This is similar to the tensorflow depthwise_convolution implementation (there num_groups = in_channels).

The other implementation would be to take inputs of shape `(batches, in_channels, ih, iw)` and kernel of shape `(num_filters, in_channels // num_groups , kh, kw)` then after splitting, the inputs would be of shape `(batches, in_channels // num_groups, ih, iw)` and kernel would be of shape `(num_filters // num_groups , in_channels // num_groups , kh , kw)` the final output would be of shape `(batches , out_channels, oh , ow)` .This method would require the user to compute the shape of the kernel taking `num_groups` into account i.e the 2nd dimension of the kernel to be exactly `in_channels // num_groups` .Also `num_filters` must be divisible by `num_groups`.

Currently, the first method is implemented. But it should not take me more than a day to shift it to implement the second method instead if that is preferred

affanv14: After working out the finer details, I think the second method is better to implement for CorrMM and GPUCorrMM as we can easily divide the weight tensor into separate groups . 

Since this PR is a mess, should i close this one and open a new one  or make changes in this one itself?

affanv14:
The method I implemented here would be quite inefficient for CorrMM and GPUCorrMM.https://github.com/Theano/Theano/pull/5973#discussion_r118651963 and https://github.com/Theano/Theano/pull/5973#discussion_r118930124 describe the details for the change in depth. Since this PR has become too messy, I have shifted work to https://github.com/Theano/Theano/pull/5991

theano/tensor/nnet/abstract_conv.py
nouiz: I don't understand the second option. The part that I don't understand is "The other implementation would be to take inputs of shape (batches, in_channels, ih, iw) and kernel of shape (num_filters, in_channels // num_groups , kh, kw)"

There, how the inputs can be of size "in_channels // num_groups"? I don't get that.

affanv14: For grouped convolution, the input would need to be divided into `num_groups` different groups. So for each group the input would be of shape (batches, in_channels // num_groups, ih, iw) .

 The kernel shape for each group would be (num_filters // num_groups, in_channels // num_groups , kh, kw). Now, for the entire kernel, the individual  kernels from each group would need to be combined either in the first or the second dimension. Concatenating using the second dimension would result in a kernel of shape (numfilters // num_groups , in_channels, kh, kw) which is done in the first method. 

Combining using the first dimension would result in a kernel of shape (num_filters, in_channels // num_groups , kh, kw) which is the second method.

affanv14: Adding to this.
@f0k has explained it quite well here https://github.com/Theano/Theano/issues/5621#issuecomment-282803917, here he switched the dimensions of inputs and outputs though

f0k: > switched the dimensions of inputs and outputs though

Oops, yes. :+1: for working on this. Regarding the choice of memory layout / shape, we might want to check other frameworks' implementations again (linked at https://github.com/Theano/Theano/issues/5621#issuecomment-282344400). If there's no clear argument for either layout, we might opt for compatibility (to make it easier to port models between frameworks).

affanv14: > we might want to check other frameworks' implementations again (linked at #5621 (comment)). If there's no clear argument for either layout, we might opt for compatibility (to make it easier to port models between frameworks).

Regarding this, [torch](https://github.com/soumith/cudnn.torch/blob/f746b79/SpatialConvolution.lua#L22) and [caffe](https://github.com/BVLC/caffe/blob/bbc4e57/src/caffe/layers/base_conv_layer.cpp#L137) have used the same memory layout for the weights, I have used the same layout in the new PR as well.
Tensorflow's layout is quite different [here ](https://www.tensorflow.org/versions/r0.11/api_docs/python/nn/convolution#depthwise_conv2d),we probably cannot port to tensorflow
So, shall I continue with the method being implemented in the new PR?

f0k: Thanks for checking! So TF actually only implements the special case of `num_groups == num_input_channels`. For this, they use a tensor of `[filter_height, filter_width, in_channels, channel_multiplier]`. Transposing to our layout, this is `[channel_multiplier, in_channels, filter_height, filter_width]`, which corresponds to `[out_channels // num_groups, in_channels, filter_height, filter_width]`. So yes, it's incompatible.

> So, shall I continue with the method being implemented in the new PR?

Seems like a good idea to me, yes!

nouiz:
I agree, I would use the existing interface that support all the feature.

Issue #5976, opened 2017-05-24
Merge Supervisor Feature inside the DestroyHandler


ReyhaneAskari:
After the PR on faster toposort, we can continue to make it faster. In the benchmark, the gpu_elemwise_inplace optimization was still the slowest optimization. This is due to the Supervisor that trigger the building of the complex datastructure needed for toposort and cycle detection. But we don't need that strucuture just for the Supervisor.

- [ ] merge the Supervisor class inside the DestroyHandler. Mostly, add the Supervisor __init__ parameter to DestroyHandler. Then merge the Supervisor.validate() inside the DH.validate. Then make sure all works.
- [ ] In the case of faster topo and when the structure isn't build, we can speed up the computation by not building the structure. So in that case, replace the check by this new check that is a subset of building the structure. First, don't test the fgraph.outputs, as they will always be True with the new check.
- [ ] Then, remove the need for building the complex data structure. The new check will start from the protected variables. Check there clients and all there views. If there is one clients of those variables that is a destroy, the check should fail.

ReyhaneAskari:
New plan : 
1. start over from the faster topo branch
2. Modify destroyhandler to add a new fgraph.has_destroyers(list) (as fgraph.destroyers) that will return True is any of the variable in the list have a destroyers. (the old Superviser.validate check without the raise)
3. Make that new fct do the Superviser.validate check when self.algo = cycle.
4. Make sure all tests still pass
5. Modify the new method to be faster (don't build the big data structure) when the algo is fast.

nouiz:
The changes needed where done in gh-6000

PR #5977, opened 2017-05-24
Merge supervisor

ReyhaneAskari:
This is the first for issue #5976. The supervisor class is removed and now the DestroyHandler should do the task.


ReyhaneAskari:
I will do a rebase after the merge of faster_topo PR.

ReyhaneAskari:
@nouiz , after the discussions with @lamblin last week, we found out that [here](https://github.com/Theano/Theano/blob/master/theano/compile/function_module.py#L185), the destroy handler should be given the protected variables. Since there were too many instantiations of DH we decided to have a minimal change to see if the tests would pass. So I created a protected attribute for fgraph and replaced the ```DH.protected``` with ```fgraph.protected```. We will see how the tests will pass.

ReyhaneAskari:
jenkins test this.

ReyhaneAskari:
Some of the tests were failing because it seems that the output of the graph was protected by the output guard. However, now we check the output of the graph inside the destroy handler as protected variables and we do not allow the inplace to happen on an output of the graph. So the output guard is not added as a node in the graph.

nouiz:
I'm not sure setting protected as a FunctionGraph attribute/parameter is a good fix to the link you provided: https://github.com/Theano/Theano/blob/master/theano/compile/function_module.py#L185

I think we can just remove this loop, as now it is always done just bellow. (the Supervisor was converted to a DestroyHandler and is now always introduced).

theano/compile/function_module.py
```diff
@@ -121,26 +121,6 @@ def fgraph_updated_vars(fgraph, expanded_inputs):
     return updated_vars
 
 
-class Supervisor:
```
ReyhaneAskari: Yes this is an option. I did see the pickle test failing and I was working on it yesterday. I thought maybe a cleaner option would be to unpickle the old test and remove the Supervisor from there and pickle it again. What do you think?

theano/compile/function_module.py
```diff
@@ -121,26 +121,6 @@ def fgraph_updated_vars(fgraph, expanded_inputs):
...
-    Listener for FunctionGraph events which makes sure that no
-    operation overwrites the contents of protected Variables. The
-    outputs of the FunctionGraph are protected by default.
-
-    """
-
-    def __init__(self, protected):
-        self.protected = list(protected)
-
-    def validate(self, fgraph):
```
nouiz: New way of doing it:
- start over from the faster topo branch
- Modify destroyhandler to add a new fgraph.has_destroyers(list) (as fgraph.destroyers) that will return True is any of the variable in the list have a destroyers. (the old Superviser.validate check without the raise)
- Make that new fct do the Superviser.validate check when self.algo = cycle.
- Make sure all tests still pass
- Modify the new method to be faster (don't build the big data structure) when the algo is fast.

PR #5980, opened 2017-05-25
Function that check the axes

### LINKED ISSUE #5979: Add a function that check the axes and convert them to a Python list of integers
Since we sometimes have an axis as argument of an op (max_and_argmax, Argmax or the new softmax), we could avoid some code redundancy by adding a function (maybe in tensor.utils) that take as input an integer/list/tuple/whatever with tensor x and return an ordered python list of integers if the axes are valid or return an error if not. We could call it: check_and_normalize_axes(x, axes).

bordesf:
fix #5979

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -18,6 +18,7 @@ from theano.gof import Apply, Constant, Op, Variable
 from theano.gof.type import Generic
 
 from theano.tensor import elemwise
+from theano.tensor.utils import check_and_normalize_axes
 from theano.tensor.var import (AsTensorError, TensorVariable,
                                TensorConstant, TensorConstantSignature,
                                _tensor_py_operators)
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -968,6 +968,53 @@ def _pack(x):
...
+    Parameters
+    ----------
+    x: Tensor variable
+    axis = Integer, tuple or list of integers
+
+    Returns
+    -------
+    axis: list of integers
+    """
+    x = as_tensor_variable(x)
```
ReyhaneAskari: If x is already a Tensor variable (as you have mentioned in the Parameters L#978), why do we need this conversion? Also you do the conversions at L#1600 and L#1431.

theano/tensor/basic.py
```diff
@@ -968,6 +968,53 @@ def _pack(x):
...
+            raise TypeError("Computation needs a constant axis. Got %s" % axis)
+        else:
+            assert axis.dtype in integer_dtypes
+            if (isinstance(axis.data, (integer_types, np.integer)) or
+                    (isinstance(axis.data, np.ndarray) and axis.data.ndim == 0)):
+                        axis = [int(axis.data)]
+            elif isinstance(axis.data, (list, np.ndarray)):
+                axis = [int(i) for i in axis.data]
+    else:
+        axis = []
```
ReyhaneAskari: What will this else cover? Don't we want to raise an error in such cases?

theano/tensor/basic.py
nouiz: It is standard to call as_tensor_variable on exising Theano variable. This will be a no op and just return the input. It prevent making if at many places in Theano. So this is good.

theano/tensor/basic.py
bordesf: I corrected this in the last commit, thanks.

ReyhaneAskari:
Thanks @bordesf .

PR #5982, opened 2017-05-26
Add flag and implementation of build_infer_shape.

nouiz:
TODO:
- make ShapeFeature reuse the constant shape from shared variable
- don't change the autoname due to that PR.
- Make basic safety check that sharedvariable shapes don't change, like in set_value. But don't try to cover the hard cases.

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -800,6 +800,15 @@ AddConfigVar(
 
 
 AddConfigVar(
+    'build_infer_shape',
+    ("If 'True', Theano will try to infer the shape while building the graph."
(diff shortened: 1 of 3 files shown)


lamblin:
Here is an experimental feature that uses the optimization and replacement mechanism, but could replace Variables with ones of different shape #3314.
There could be more in external code for instance, so having a way of detecting changes of inputs, or at least force the recomputation of shapes in a graph or subgraph would be good.

nouiz:
I updated the description with what I think we need to do to get a first version.

slefrancois:
jenkins test win, to check buildbot

Issue #5978, opened 2017-05-25
Wrong arguments order in register_specify_shape_c_code of tensor/type.py


TaoLv:
In code of `register_specify_shape_c_code` of tensor/type.py, try to check the input dimension and shape length: 
```
if (PyArray_NDIM(%(iname)s) != PyArray_DIMS(%(shape)s)[0]) {
            PyErr_Format(PyExc_AssertionError,
                         "SpecifyShape: vector of shape has %%d elements,"
                         " but the input has %%d dimensions.",
                         PyArray_NDIM(%(iname)s),
                         PyArray_DIMS(%(shape)s)[0]);
            %(fail)s;
}
```

Here in the error message, two formatting inputs may be in an opposite order?

nouiz:
Right! Do you want to make a PR and be in the history?

TaoLv:
Sure, I will try..

PR #5983, opened 2017-05-27
Fix arguments order in register_specify_shape_c_code of tensor/type.py

TaoLv:
[Issue #5978](https://github.com/Theano/Theano/issues/5978)

fix gh-5978


nouiz:
thanks

PR #5987, opened 2017-05-29
Do not forget offset in GpuCumOp

lamblin:
This should work for CUDA, not for OpenCL but I'm not even sure this Op supported OpenCL in the first place.

### DIFF
--- a/theano/gpuarray/extra_ops.py
+++ b/theano/gpuarray/extra_ops.py
@@ -35,7 +35,7 @@ class GpuCumOp(GpuKernelBase, Op):
         return hash(self.axis) ^ hash(self.mode)
 
     def c_code_cache_version(self):
-        return (3,)
+        return (4,)
 
     def c_headers(self):
         return ['<numpy_compat.h>', '<gpuarray/types.h>', '<gpuarray_helper.h>']
@@ -380,8 +380,8 @@ class GpuCumOp(GpuKernelBase, Op):
                     size_t dimGrid[3] = {dimGridX, localDimGridY, localDimGridZ};
                     size_t dimBlock[3] = {dimBlockX, 1, 1};  // One cum op per block.
                     size_t sharedBytes = (2*dimBlockX) * sizeof(float);
-                    void* kernel_params[] = {(void*) input->ga.data,
-                                             (void*) output->ga.data,
+                    void* kernel_params[] = {(void*) ((char *)(input->ga.data) + input->ga.offset),
+                                             (void*) ((char *)(output->ga.data) + output->ga.offset),
                                              (void*) &nbElementsPerCumOp,
                                              (void*) &inputStrides_x,
                                              (void*) &inputStrides_y,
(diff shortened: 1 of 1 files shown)


abergeron:
The current code will not work for OpenCL, nor will the solution.  When we port this code to OpenCL we should also just add offset parameters to the kernels to handle that cleanly.

theano/gpuarray/extra_ops.py
```diff
@@ -430,8 +430,8 @@ def c_support_code_struct(self, node, nodename):
                         size_t dimBlock[3] = {1, 1, 1};
                         size_t tmp0 = shape[axis]-2;
                         size_t tmp1 = shape[axis]-1;
-                        void* kernel_params[] = {(void*) input->ga.data,
-                                                 (void*) output->ga.data,
+                        void* kernel_params[] = {(void*) ((char *)(input->ga.data) + input->ga.offset),
```
nouiz: From memory, there was a macro that does that automatically. I think we should try to always use it. It will help prevent such error.

obilaniu: @lamblin Related to this: https://github.com/mila-udem/support/issues/65, where the problem is a `gpudata` with an invalid `ctx`.

Why is `(void*) ((char *)(input->ga.data) + input->ga.offset` at all valid?

From my understanding of the code, `input` is `PyGpuArrayObject*`, `input->ga` is `GpuArray*` and `input->ga.data` is therefore `gpudata*`. So why on earth would one add a byte offset to it?

lamblin: I think `input->ga.data` is the address of the beginning of the memory segment, not the address of the first element, and we need to add the offset to get to the first element. Am I missing something?

obilaniu: @lamblin Well, yes:  `input->ga.data` is `gpudata*`. It's _*not*_ a `void*` of some variety.

obilaniu: @lamblin Basically, I think you had in mind `((partial_gpudata*)input->ga.data)->devptr`, as 

```
typedef struct _partial_gpudata {
  void *devptr;
  gpucontext *ctx;
} partial_gpudata;
```

seems to suggest.

lamblin: Oh, OK, I did not realize that call_kernel did some magic.
I'll fix that.

obilaniu: @lamblin I don't know exactly what `cuda_callkernel()` does, but it certainly doesn't seem to substitute `data` with `data->devptr + offset` automatically.

At any rate, the point is that we must be careful here to match the declared type of the kernel with the provided argument. If we give `GA_BUFFER`, then we provide a `gpudata*`; If we give `GA_POINTER`, then we provide `devptr+offset`.

abergeron: Please do not use GA_POINTER.  It isn't functional and I'll get rid of it soon.  But otherwise, yes, you're right.

I must have been too addled when I merged this to miss that problem.

PR #5991, opened 2017-05-30
Implement Grouped Convolutions

affanv14:
Continues work from https://github.com/Theano/Theano/pull/5973

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -499,6 +499,9 @@ class GpuDnnConv(DnnBase):
     algo : {'small', 'none', 'large', 'fft', 'fft_tiling', 'winograd', 'guess_once',
             'guess_on_shape_change', 'time_once', 'time_on_shape_change'}
         Default is the value of :attr:`config.dnn.conv.algo_fwd`.
+    num_groups :
+        Divides the image, kernel and output tensors into num_groups
+        separate groups. Each which carry out convolutions separately
 
     """
     _f16_ok = True
@@ -639,6 +642,9 @@ class GpuDnnConvGradW(DnnBase):
     algo : {'none', 'deterministic', 'fft', 'small', 'guess_once',
             'guess_on_shape_change', 'time_once', 'time_on_shape_change'}
         Default is the value of :attr:`config.dnn.conv.algo_bwd_filter`.
+    num_groups :
+        Divides the image, kernel and output tensors into num_groups
+        separate groups. Each which carry out convolutions separately
 
     """
     _f16_ok = True
@@ -772,6 +778,9 @@ class GpuDnnConvGradI(DnnBase):
     algo : {'none', 'deterministic', 'fft', 'fft_tiling', 'winograd', 'guess_once',
             'guess_on_shape_change', 'time_once', 'time_on_shape_change'}
         Default is the value of :attr:`config.dnn.conv.algo_bwd_data`.
+    num_groups :
+        Divides the image, kernel and output tensors into num_groups
+        separate groups. Each which carry out convolutions separately
 
     """
     _f16_ok = True
@@ -912,6 +921,9 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1), dilation=(1, 1),
         should be done. Possible values are 'as_input', 'float16', 'float32'
         and 'float64'. Default is the value of
         :attr:`config.dnn.conv.precision`.
+    num_groups :
+        Divides the image, kernel and output tensors into num_groups
+        separate groups. Each which carry out convolutions separately
 
 
     .. warning:: The cuDNN library only works with GPUs that have a compute
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -868,8 +868,8 @@ class GpuDnnConvGradI(DnnBase):
 
 
 def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1), dilation=(1, 1),
-             num_groups=1, conv_mode='conv', direction_hint=None, workmem=None,
-             algo=None, precision=None):
+             conv_mode='conv', direction_hint=None, workmem=None,
+             algo=None, precision=None, num_groups=1):
     """
     GPU convolution using cuDNN from NVIDIA.
 
@@ -1111,7 +1111,8 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1
 
 
 def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
-                   subsample=(1, 1), dilation=(1, 1), num_groups=1, conv_mode='conv', precision=None):
+                   subsample=(1, 1), dilation=(1, 1), conv_mode='conv',
+                   precision=None, algo=None, num_groups=1):
     """
     TODO: document this
     """
@@ -1126,7 +1127,7 @@ def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
     desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample, dilation=dilation,
                           conv_mode=conv_mode, precision=precision)(kerns_shp)
     out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*kerns_shp)
-    return GpuDnnConvGradW(num_groups=num_groups)(img, topgrad, out, desc)
+    return GpuDnnConvGradW(algo=algo, num_groups=num_groups)(img, topgrad, out, desc)
 
 
 def dnn_gradweight3d(img, topgrad, kerns_shp, border_mode='valid',
@@ -1139,7 +1140,8 @@ def dnn_gradweight3d(img, topgrad, kerns_shp, border_mode='valid',
 
 
 def dnn_gradinput(kerns, topgrad, img_shp, border_mode='valid',
-                  subsample=(1, 1), dilation=(1, 1), num_groups=1, conv_mode='conv', precision=None):
+                  subsample=(1, 1), dilation=(1, 1), conv_mode='conv',
+                  precision=None, algo=None, num_groups=1):
     """
     TODO: document this
     """
@@ -1154,7 +1156,7 @@ def dnn_gradinput(kerns, topgrad, img_shp, border_mode='valid',
     desc = GpuDnnConvDesc(border_mode=border_mode, subsample=subsample, dilation=dilation,
                           conv_mode=conv_mode, precision=precision)(kerns.shape)
     out = GpuAllocEmpty(dtype=kerns.dtype, context_name=ctx_name)(*img_shp)
-    return GpuDnnConvGradI(num_groups=num_groups)(kerns, topgrad, out, desc)
+    return GpuDnnConvGradI(algo=algo, num_groups=num_groups)(kerns, topgrad, out, desc)
 
 
 def dnn_gradinput3d(kerns, topgrad, img_shp, border_mode='valid',
@@ -2679,9 +2681,9 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
                         border_mode=op.border_mode,
                         subsample=op.subsample,
                         dilation=op.filter_dilation,
-                        num_groups=op.num_groups,
                         direction_hint='forward!',
-                        conv_mode=conv_mode)
+                        conv_mode=conv_mode,
+                        num_groups=op.num_groups)
     elif isinstance(op, AbstractConv2d_gradWeights):
         shape = (inp2.shape[1], inp1.shape[1],
                  inputs[2][0], inputs[2][1])
@@ -2689,8 +2691,8 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
                               border_mode=op.border_mode,
                               subsample=op.subsample,
                               dilation=op.filter_dilation,
-                              num_groups=op.num_groups,
-                              conv_mode=conv_mode)
+                              conv_mode=conv_mode,
+                              num_groups=op.num_groups)
     elif isinstance(op, AbstractConv2d_gradInputs):
         shape = (inp2.shape[0], inp1.shape[1],
                  inputs[2][0], inputs[2][1])
@@ -2698,8 +2700,8 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
                              border_mode=op.border_mode,
                              subsample=op.subsample,
                              dilation=op.filter_dilation,
-                             num_groups=op.num_groups,
-                             conv_mode=conv_mode)
+                             conv_mode=conv_mode,
+                             num_groups=op.num_groups)
     return [rval]
 
 
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -508,9 +508,10 @@ class GpuDnnConv(DnnBase):
     params_type = ParamsType(conv_algo=cudnn.cudnnConvolutionFwdAlgo_t,
                              choose_algo=bool_t, choose_once=bool_t, choose_time=bool_t,
                              inplace=bool_t,
-                             handle=handle_type)
+                             handle=handle_type,
+                             num_groups=int_t)
 
-    def __init__(self, algo=None, inplace=False):
+    def __init__(self, algo=None, inplace=False, num_groups=1):
         DnnBase.__init__(self, ["dnn_conv_base.c", "dnn_fwd.c"],
                          "APPLY_SPECIFIC(conv_fwd)")
 
@@ -530,6 +531,7 @@ class GpuDnnConv(DnnBase):
         self.choose_algo = self.algo in SUPPORTED_DNN_CONV_ALGO_RUNTIME
         self.choose_once = self.algo in DNN_CONV_ALGO_CHOOSE_ONCE
         self.choose_time = self.algo in DNN_CONV_ALGO_CHOOSE_TIME
+        self.num_groups = num_groups
 
     def __setstate__(self, d):
         self.__dict__.update(d)
@@ -642,9 +644,10 @@ class GpuDnnConvGradW(DnnBase):
     params_type = ParamsType(conv_algo=cudnn.cudnnConvolutionBwdFilterAlgo_t,
                              choose_algo=bool_t, choose_once=bool_t, choose_time=bool_t,
                              inplace=bool_t,
-                             handle=handle_type)
+                             handle=handle_type,
+                             num_groups=int_t)
 
-    def __init__(self, inplace=False, algo=None):
+    def __init__(self, inplace=False, algo=None, num_groups=1):
         DnnBase.__init__(self, ["dnn_conv_base.c", "dnn_gw.c"],
                          "APPLY_SPECIFIC(conv_gw)")
         self.inplace = bool(inplace)
@@ -662,6 +665,7 @@ class GpuDnnConvGradW(DnnBase):
         self.choose_algo = self.algo in SUPPORTED_DNN_CONV_ALGO_RUNTIME
         self.choose_once = self.algo in DNN_CONV_ALGO_CHOOSE_ONCE
         self.choose_time = self.algo in DNN_CONV_ALGO_CHOOSE_TIME
+        self.num_groups = num_groups
 
     def __setstate__(self, d):
         self.__dict__.update(d)
@@ -771,9 +775,10 @@ class GpuDnnConvGradI(DnnBase):
     params_type = ParamsType(conv_algo=cudnn.cudnnConvolutionBwdDataAlgo_t,
                              choose_algo=bool_t, choose_once=bool_t, choose_time=bool_t,
                              inplace=bool_t,
-                             handle=handle_type)
+                             handle=handle_type,
+                             num_groups=int_t)
 
-    def __init__(self, inplace=False, algo=None):
+    def __init__(self, inplace=False, algo=None, num_groups=1):
         DnnBase.__init__(self, ["dnn_conv_base.c", "dnn_gi.c"],
                          "APPLY_SPECIFIC(conv_gi)")
         self.inplace = bool(inplace)
@@ -791,6 +796,7 @@ class GpuDnnConvGradI(DnnBase):
         self.choose_algo = self.algo in SUPPORTED_DNN_CONV_ALGO_RUNTIME
         self.choose_once = self.algo in DNN_CONV_ALGO_CHOOSE_ONCE
         self.choose_time = self.algo in DNN_CONV_ALGO_CHOOSE_TIME
+        self.num_groups = num_groups
 
     def __setstate__(self, d):
         self.__dict__.update(d)
@@ -854,7 +860,7 @@ class GpuDnnConvGradI(DnnBase):
 
 
 def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1), dilation=(1, 1),
-             conv_mode='conv', direction_hint=None, workmem=None,
+             num_groups=1, conv_mode='conv', direction_hint=None, workmem=None,
              algo=None, precision=None):
     """
     GPU convolution using cuDNN from NVIDIA.
@@ -973,7 +979,7 @@ def dnn_conv(img, kerns, border_mode='valid', subsample=(1, 1), dilation=(1, 1),
                                     filter_dilation=dilation)
     out_shp = assert_conv_shape(out_shp)
     out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*out_shp)
-    return GpuDnnConv(algo=algo)(img, kerns, out, desc)
+    return GpuDnnConv(algo=algo, num_groups=num_groups)(img, kerns, out, desc)
 
 
 def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1, 1, 1),
@@ -1097,7 +1103,7 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1
 
 
 def dnn_gradweight(img, topgrad, kerns_shp, border_mode='valid',
-                   subsample=(1, 1), dilation=(1, 1), conv_mode='conv', precision=None):
+                   subsample=(1, 1), dilation=(1, 1), num_groups=1, conv_mode='conv', precision=None):
     """
     TODO: document this
     """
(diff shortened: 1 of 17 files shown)


theano/tensor/nnet/abstract_conv.py
```diff
@@ -181,6 +183,8 @@ def get_conv_gradweights_shape(image_shape, top_shape,
 
     if filter_dilation is None:
         filter_dilation = np.ones(len(subsample), dtype='int')
+    if num_groups > 1 and len(subsample) == 2:
```
nouiz: Here, I would do something like:
if num_groups > 1:
    assert len(subsample) == 2
    nchan = nchan // num_groups

to get an error in the not supported case.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -286,6 +293,8 @@ def get_conv_gradinputs_shape(kernel_shape, top_shape,
...
+        nkern = nkern * num_groups
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1396,7 +1417,7 @@ class BaseAbstractConv(Op):
 
     def __init__(self, convdim,
```
nouiz: In the \__props__ attribute above, you should add num_groups. See http://deeplearning.net/software/theano/extending/op.html#__props__ for an explanation. If that isn't clear, tell me!

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2097,10 +2168,20 @@ def perform(self, node, inp, out_):
         axes_order = (1, 0) + tuple(range(2, self.convdim + 2))
         flip_filters = ((slice(None), slice(None)) +
                         (slice(None, None, -1),) * self.convdim)
+
+        def correct_for_groups(mat):
```
nouiz: The use of correct_for_groups is fine for the slow python version.
But as you wrote yourself, the transpose/reshape isn't optimal as it will cause a copy.
In the c code, we can just call the convolution many times and don't you have done that here? Anyway, it is slow python code, so I don't mind which version is used it. I just want to be sure the C code won't have that and using the same/more similar algo could be useful to help you make the change to the C code.

affanv14:
>  It mostly need at least a C version in this PR before we merge it.

Yes, I'm  done with all three passes of CorrMM implementation, I just need to write tests for it. I will push the commits by today or tomorrow.

theano/tensor/nnet/abstract_conv.py
affanv14: I haven't used this anywhere in the c code. The c code gives  more freedom than the python implementation.There was no need reshapes and  transposes .

affanv14: I will look into how I can change the python implementation as well.

theano/tensor/nnet/abstract_conv.py
affanv14: Got it, the docs explained it pretty well.

affanv14:
I will do that.

Also, do I need to add more test cases?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1699,3 +1698,130 @@ def test_conv2d_grad_wrt_weights(self):
...
                         utt.assert_allclose(f_new(input_val, out_grad_val), f_old(input_val, filter_val, out_grad_val))
+
+
+class Grouped_conv_noOptim(unittest.TestCase):
+    def setUp(self):
+        self.num_groups = 3
+        self.border_mode = 'valid'
+        self.subsample = (1, 1)
+        self.img_shape = (5, 6, 5, 5)
+        self.kern_shape = (6, 2, 3, 3)
```
nouiz: Can you make img_shape and kern_shape a list of shapes and test some non-square image and kernel? Then in each test, loop over those shapes. Just 3-5 different shapes should be good I think.

I would also also like a test that compare the group convolution to a Theano graph of many convolution that do the equivalent. Mostly, do in the graph what is supported to be done inside the op when group convolution is used. I think it could be interesting to put that in the doc to help people understand what it does.

affanv14: > Can you make img_shape and kern_shape a list of shapes and test some non-square image and kernel? Then in each test, loop over those shapes

I was trying to test  values for num_groups == num_channels. But I ran into some trouble with verify_grad with this error 'Non-unit value on shape on a broadcastable dimension.'

affanv14:
Sorry it took so long for the corrmm implementation. My first implementation had im2col  inside the both loops.

affanv14:
These tests pass on my system. Could it be that cache wasn't cleared on travis? or is it something else?

nouiz:
The error is related to the shapes, not to the dtype. The dtype shouldn't affect the shape normally.

f0k:
Weird, the flake8 test failed, but it ran for over 6 hours and there's no failure in its log. Maybe you can restart just the flake8 build, @nouiz?

affanv14:
could [this](https://software.intel.com/en-us/mkl-developer-reference-fortran-gemm-batch) be used for faster computations?

affanv14:
I still have the following left to do which i should finish by today.

- change the python code again as specified [here](https://github.com/Theano/Theano/pull/5991#discussion_r119603007)
- Write another test as specified [here](https://github.com/Theano/Theano/pull/5991#discussion_r119887385)  in the second paragraph

I also may need to make a few small changes along with the docs.

f0k:
> could [this](https://software.intel.com/en-us/mkl-developer-reference-fortran-gemm-batch) be used for faster computations?

See https://github.com/Theano/Theano/issues/5621#issuecomment-290178569 for my answer on this.

affanv14:
for the Abstractconv changes, memory has to be allocated for the output in the perform function of each of the three passes and the in each iteration of the for loop a part of this memory will be filled, wouldn't this lead to more memory use?

affanv14:
Also for the GPUcorrmm changes, Could you explain [this](https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1473-L1525) part, i did not understand it.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -430,14 +430,14 @@ class TestGroupCorr2d(Grouped_conv_noOptim):
     def test_graph(self):
         # define common values  first
         groups = 3
-        bottom = np.random.rand(3, 6, 5, 5)
-        kern = np.random.rand(9, 2, 3, 3)
+        bottom = np.random.rand(3, 6, 5, 5).astype('float32')
+        kern = np.random.rand(9, 2, 3, 3).astype('float32')
```
f0k: Shouldn't this be `.astype(theano.config.floatX)`?

affanv14: the old tests in test_abstract_conv.py  used float32, so I used it in all my tests as well. But yeah, I think theano.config.floatX is more appropriate. I'll make those changes.

f0k:
> Could you explain [this](https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1473-L1525) part, i did not understand it.

This is actually a remnant from when Theano did not have the abstract conv2d interface yet. In those dark times, everything was a conv2d -- the forward pass, and the two backward passes. This code figures out which of the three corrmm implementations -- the forward pass, the backward pass wrt. weights, or the backward pass wrt. inputs -- would probably be the most performant for the given combination of shapes and padding. This code was kept because it can also improve performance in other cases. E.g., when you define a network with full convolutions in the forward pass, it still makes sense to use the backward pass wrt. inputs for carrying out the full convolution. See benanne.github.io/2014/12/09/theano-metaopt.html for some more information on how convolution implementations can be swapped (but ignore the part on the meta-optimizer, this is something else not relevant here).

It's not immediately clear to me if the last part of the heuristics should take the groups into account. The general idea of swapping implementations should not be affected by the groups.

nouiz:
For the old part that Jan told, just don't forget to pass the group
parameter in the if part. This would need test if you do that. But I would
also accept that you don't take the if part if the number of group is
greater then one. I think the meta-optimizer should take care of that, not
the if part. So don't spend too much time on that.

affanv14:
@nouiz I think for now i'll skip the if part if groups is greater than 1 and finish the implementation and later come back and make changes to the if part if required.

>  In those dark times

@f0k interesting choice of words!

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1701,126 +1702,137 @@ def test_conv2d_grad_wrt_weights(self):
...
-                        mode=self.mode)
+        for imshp, kshp, groups in zip(self.img_shape, self.kern_shape, self.num_groups):
+            img = np.random.random(imshp).astype('float32')
+            kern = np.random.random(kshp).astype('float32')
+            split_imgs = np.split(img, groups, axis=1)
+            split_kern = np.split(kern, groups, axis=0)
+            img_sym = theano.tensor.tensor4('img')
+            kern_sym = theano.tensor.tensor4('kern')
+
+            grouped_abstractconv_func = self.conv2d(border_mode=self.border_mode,
```
nouiz: This isn't a func, but an op. You can change the variable name to just op. with a lower case, we use it elsewhere to mean an op instance.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1701,126 +1702,137 @@ def test_conv2d_grad_wrt_weights(self):
...
-
-        def abstract_conv_gradweight(inputs_val, output_val):
-            return grouped_abstractconvgrad_func(inputs_val, output_val, self.kern_shape[-2:])
+        for imshp, kshp, tshp, groups in zip(self.img_shape, self.kern_shape, self.top_shape, self.num_groups):
+            img = np.random.random(imshp).astype('float32')
+            top = np.random.random(tshp).astype('float32')
+            split_imgs = np.split(img, groups, axis=1)
+            split_top = np.split(top, groups, axis=1)
+            img_sym = theano.tensor.tensor4('img')
+            top_sym = theano.tensor.tensor4('top')
```
nouiz: You can move img_sym and top_sym out of the loop.

nouiz:
For the if part, add this diff. It will make sure it don't go in code that don't support groups. Then that all for that. The meta optimizer should take care later of that part:
```
diff --git a/theano/tensor/nnet/opt.py b/theano/tensor/nnet/opt.py
index 6e8b3db..0f37b97 100644
--- a/theano/tensor/nnet/opt.py
+++ b/theano/tensor/nnet/opt.py
@@ -449,7 +449,7 @@ def local_conv2d_gradinputs_cpu(node):
         return None
 
     # Conv 3d implementation, needed when subsample > 2
-    if node.op.border_mode == 'valid' and node.op.subsample != (1, 1):
+    if node.op.border_mode == 'valid' and node.op.subsample != (1, 1) and node.op.num_groups==1:
         kern = kern[:, :, ::-1, ::-1]
         shuffled_kern = kern.dimshuffle(0, 2, 3, 'x', 1)
         shuffled_topgrad = topgrad.dimshuffle(0, 2, 3, 'x', 1)
```

nouiz:
For MKL, intel is also working on optimizing Theano on CPU, but with another approch. So for now, I would prefer you don't spend time on that. There is other things on your todo list and this would be in big part duplicate work.

nouiz:
I forgot to ask, do you have a GPU for the gpu implementation?

theano/tensor/nnet/tests/test_abstract_conv.py
affanv14: oh, my bad. I'll get on it.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1878,7 +1945,7 @@ def infer_shape(self, node, input_shapes):
         imshp = input_shapes[0]
         topshp = input_shapes[1]
         kshp = self.kshp[:] if self.kshp is not None else [None] * (2 + self.convdim)
-        fallback_kshp = ([topshp[1], imshp[1]] +
+        fallback_kshp = ([topshp[1], imshp[1] // self.num_groups] +
```
nouiz: Here add the // only if self.num_groups isn't 1.

affanv14:
> I forgot to ask, do you have a GPU for the gpu implementation?

I have an amd apu on my laptop. But since i need an Nvidia graphics card, I planned to work on it using aws since i still have over 140$ worth of aws free credits. That should last me about 250+ hours on a gpu spot instance.
So i'll just write the code locally and then test it on aws before pushing any code.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -66,7 +66,6 @@ def get_conv_output_shape(image_shape, kernel_shape,
     """
```
nouiz: In the description, I would add a note that the num_group don't influence the shape. So people don't think it was forgotten.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -286,6 +294,9 @@ def get_conv_gradinputs_shape(kernel_shape, top_shape,
 
     if filter_dilation is None:
         filter_dilation = np.ones(len(subsample), dtype='int')
+    if num_groups > 1:
+        assert len(subsample) == 2
+        nkern = nkern * num_groups
```
nouiz: I don't get this. If the forward op don't have the output shape changed, why the grads ops would compute different shapes?

nouiz:
ok. But Theano should have a significant part working with OpenCL including
the GpuCorrMM part. So you could use AWS only for the CUDNN part.

theano/tensor/nnet/abstract_conv.py
affanv14: The only way the shapes are different from normal comvolutions are that 
kshp[1] = ishp[1] // num_groups as opposed to kshp[1] = ishp[1] in normal convolutions.

In the forward pass, The top shape isn't influenced by kshp[1], instead its shape would be 
(ishp[0], khsp[0], oh, ow).This is the same of normal convolutions as well, so it does not change.

In the backward pass with respect to weights, in normal convolutions  it would have been
( tshp[1], ishp[1], kh, kw) but for grouped convolution kshp[1] would be ishp[1] // numgroups.

It is similar wrt input as well only that ishp[1] would be kshp[1]  multiplied by num_groups

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1392,11 +1415,11 @@ class BaseAbstractConv(Op):
     """
     check_broadcast = False
     __props__ = ('convdim', 'border_mode', 'subsample', 'filter_flip',
-                 'imshp', 'kshp', 'filter_dilation')
+                 'imshp', 'kshp', 'filter_dilation', 'num_groups')
 
     def __init__(self, convdim,
                  imshp=None, kshp=None, border_mode="valid",
-                 subsample=None, filter_flip=True, filter_dilation=None):
+                 subsample=None, filter_flip=True, filter_dilation=None, num_groups=1):
```
affanv14: I did add a condition in init function in BaseAbstractconv to check that.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -416,6 +417,45 @@ def test_non_contiguous(self):
...
+    if theano.config.mode == "FAST_COMPILE":
+        mode = theano.compile.get_mode("FAST_RUN")
+    else:
+        mode = None
+    conv2d = staticmethod(corr.CorrMM)
+    conv2d_gradw = staticmethod(corr.CorrMM_gradWeights)
+    conv2d_gradi = staticmethod(corr.CorrMM_gradInputs)
+    flip_filter = True
+
+    def test_graph(self):
```
affanv14: This test was implemented in accordance with this comment https://github.com/Theano/Theano/pull/5991#discussion_r119887385 in the second paragraph.

theano/tensor/nnet/abstract_conv.py
nouiz: ok, I didn't found it when I looked for this check.

affanv14:
I commited that from an aws ec2 instance( for some reason opencl had some unknown error), That is why it shows the user as ubuntu .

theano/tensor/nnet/tests/test_abstract_conv.py
affanv14: I still haven't done this. I'll finish this by today

affanv14:
I need to make a small change in the tests for gpu, It should be done by today. I also need check a few conditions .

affanv14:
I noticed that the make_node functions in the backward passes of GpuCorrmm require the shape argument to be a tensorvariable whereas the shape argument in CorrMM or AbstractConv is a  list. Wouldn't  it be better if we make all three of them the same?

f0k:
> Wouldn't it be better if we make all three of them the same?

They serve different purposes. The shape arguments in the `CorrMM` constructor or the `abstract_conv` functions are provided by the user, to give optional hard-coded shape information for the inputs or filters that Theano can use for selecting the optimal implementation. The `make_node` method gets symbolic shapes that are needed to compute a properly-sized gradient in case it cannot be unambiguously inferred from the output size. These symbolic shapes are treated as extra inputs of the Op, rather than attributes of the Op, that's why they appear in `make_node`.

nouiz:
This PR need a rebase.

affanv14:
> This PR need a rebase.
Ok  ill get on it

nouiz:
I reviewed the new code and it look good. After the rebase, we will be able to have jenkins test the GPU code.

affanv14:
I somehow managed to solve the rebase issue in this PR itself. I can continue with this.

nouiz:
You don't need to crate a new PR when you rebase one. You just need to force push to the same branch. Note travis have failures.

nouiz:
travis failure is a timeout that i restarted.

jenkins test this

affanv14:
There is some duplication of code in tests for cudnn implementation due to the way dnn_conv is called.

I still need add documentation and some other small changes.

theano/gpuarray/dnn.py
```diff
@@ -530,6 +531,7 @@ def __init__(self, algo=None, inplace=False):
         self.choose_algo = self.algo in SUPPORTED_DNN_CONV_ALGO_RUNTIME
         self.choose_once = self.algo in DNN_CONV_ALGO_CHOOSE_ONCE
         self.choose_time = self.algo in DNN_CONV_ALGO_CHOOSE_TIME
+        self.num_groups = num_groups
 
     def __setstate__(self, d):
```
notoraptor: Just a suggestion: I think you should also update `__setstate__()`, as you add a new class attribute (`num_groups`) that may not exist in old pickled instances of this Op. That may causes some errors, e.g. in FAST_COMPILE mode.

Same for every other Op where you add a new attribute.

theano/tensor/nnet/corr.py
```diff
@@ -274,6 +281,7 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
     int dilW = %(params)s->dilW;
     int padH = %(params)s->padH;
     int padW = %(params)s->padW;
+    int numgroups = %(params)s->num_groups;
```
nouiz: As this file have c code in the Op method and not in separate c file, you need to bump the version number returned by the method c_code_cache_version() of each op you changed the c code.

Op that have its c code in separate file, don't implement that method, so you are fine.

nouiz: Idem in the file gpuarray/blas.py

nouiz:
Just a quick note that tomorrow is an holiday here. So I won't be checking my emails/github.

affanv14:
@nouiz does the setstate method look alright?

affanv14:
I made the same changes to the other setstate function just for uniformity

affanv14:
Do I need to make changes for Rop? I think it works as is. Or am I wrong?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1622,14 +1665,17 @@ def perform(self, node, inp, out_):
...
             kern = kern[(slice(None), slice(None)) + (slice(None, None, -1),) * self.convdim]
-        conv_out = self.conv(img, kern, mode="valid", dilation=self.filter_dilation)
+        conv_out = self.conv(img, kern, mode="valid", dilation=self.filter_dilation, num_groups=self.num_groups)
         conv_out = conv_out[(slice(None), slice(None)) +
                             tuple(slice(None, None, self.subsample[i])
                                   for i in range(self.convdim))]
 
         o[0] = node.outputs[0].type.filter(conv_out)
 
     def R_op(self, inputs, eval_points):
```
nouiz: This should be done for the other ops too. At least two ops in this file.

affanv14: So I need to implement Rop for grads as well?
Do I need to implement it for all implementations or only abstractconv?

nouiz:
There is new information about Rop. Don't implement it. It is probably
useless and we will maybe remove the R_op method...

Just make sure for now that R_op return an error if num_groups  != 1 for
the 3 ops.

affanv14:
Ok. I checked. The other three ops have not implemented R_op

theano/gpuarray/dnn_base.c
```diff
@@ -42,7 +42,8 @@ c_set_tensorNd(PyGpuArrayObject *var, cudnnTensorDescriptor_t desc) {
     strs[i] = 1;
     dims[i] = 1;
   }
-
+  //only for grouped convolution i.e when groups > 1
+  dims[1] = dims[1] / groups;
```
abergeron: Please use the same code for this in c_set_tensorNd() and c_set_filter()

theano/gpuarray/dnn_base.c
```diff
@@ -1,7 +1,7 @@
 #section support_code
 
 static int
-c_set_tensorNd(PyGpuArrayObject *var, cudnnTensorDescriptor_t desc) {
+c_set_tensor_for_conv(PyGpuArrayObject *var, cudnnTensorDescriptor_t desc, int groups) {
```
abergeron: Use a size_t or unsigned int for groups.  It cannot be negative that way.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1699,3 +1699,157 @@ def test_conv2d_grad_wrt_weights(self):
...
                         utt.assert_allclose(f_new(input_val, out_grad_val), f_old(input_val, filter_val, out_grad_val))
+
+
+class Grouped_conv_noOptim(unittest.TestCase):
+    conv2d = staticmethod(theano.tensor.nnet.abstract_conv.AbstractConv2d)
+    conv2d_gradw = staticmethod(theano.tensor.nnet.abstract_conv.AbstractConv2d_gradWeights)
+    conv2d_gradi = staticmethod(theano.tensor.nnet.abstract_conv.AbstractConv2d_gradInputs)
+    conv2d_op = staticmethod(theano.tensor.nnet.abstract_conv.AbstractConv2d)
+    conv2d_gradw_op = staticmethod(theano.tensor.nnet.abstract_conv.AbstractConv2d_gradWeights)
+    conv2d_gradi_op = staticmethod(theano.tensor.nnet.abstract_conv.AbstractConv2d_gradInputs)
```
abergeron: Why are the _op versions of attribute the exact same thing as the non _op variants?

Why do you need to wrap ops in `staticmethod`?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1699,3 +1699,157 @@ def test_conv2d_grad_wrt_weights(self):
...
+        self.num_groups = [3, 2, 4, 4]
+        self.border_mode = 'valid'
+        self.subsample = (1, 1)
+        self.img_shape = [(5, 6, 5, 5), (4, 4, 7, 5), (3, 8, 5, 3), (2, 4, 7, 7)]
+        self.kern_shape = [(6, 2, 3, 3), (6, 2, 5, 3), (4, 2, 3, 3), (4, 1, 3, 5)]
+        self.top_shape = [(5, 6, 3, 3), (4, 6, 3, 3), (3, 4, 3, 1), (2, 4, 5, 3)]
+        self.filter_dilation = (1, 1)
+        self.ref_mode = 'FAST_RUN'
+        if theano.config.cxx == "":
+            raise SkipTest("CorrMM needs cxx and blas")
```
abergeron: Note that this doesn't check for blas presence.

abergeron:
jenkins test this please

theano/gpuarray/dnn_base.c
affanv14: I did not understand what you meant here, can you elaborate?

affanv14:
I forgot to change dtype of groups in c_set_tensor_for_conv.

affanv14:
I just checked, staticmethod call is needed only for cudnn

affanv14:
I ran the tests for all 4 implementations of the latest commit and all of them are passing for me

theano/gpuarray/tests/test_dnn.py
```diff
@@ -2158,11 +2158,11 @@ def dconvi(kern, topgrad, imshp):
...
     mode = mode_with_gpu
-    conv2d = staticmethod(dconv2d)
-    conv2d_gradw = staticmethod(dconv2dw)
-    conv2d_gradi = staticmethod(dconv2di)
-    conv2d_op = staticmethod(dnn.GpuDnnConv)
-    conv2d_gradw_op = staticmethod(dnn.GpuDnnConvGradW)
-    conv2d_gradi_op = staticmethod(dnn.GpuDnnConvGradI)
+    conv2d = dconv2d
+    conv2d_gradw = dconv2dw
+    conv2d_gradi = dconv2di
```
abergeron: These functions probably need the staticmethod otherwise, it will attempt to convert them to class methods.  But the classes don't need them.

abergeron:
jenkins test this please

affanv14:
I only haven't done this https://github.com/Theano/Theano/pull/5991#discussion_r127012126 can you elaborate on that

theano/gpuarray/dnn_base.c
abergeron: I mean that you use `dims[1] =  dims[1] / groups` here and `dims[1] /= groups` in the other.

abergeron:
jenkins test this please

slefrancois:
jenkins test this

slefrancois:
jenkins test this

lamblin:
The default implementation of some convolution gradients is not deterministic. This means that when computing the numerical gradient of these operations, there can be some noise in the estimate (the cost changes for other reasons that one given input having changed by epsilon).
Since convolutions are linear operations, the solution is to use a large step (eps) when computing the numerical gradient. This is what is done in the existing tests.

affanv14:
Got it , thanks!

abergeron:
jenkins test this please

abergeron:
Once this passes all the tests, anybody can merge.

PR #5794, opened 2017-03-31
Faster topo

ReyhaneAskari:
Regarding issue #4233, this PR adds a theano flag disbale_cycle_detection.

PR #6000, opened 2017-06-01
New destroy handler

ReyhaneAskari:
Issue #5976. Supervisor class is kept but we added a has_destroyers attribute to fgraph, but some tests regarding dumping the pickle failed because a new attribute was added to the fgraph. I changed it like this : 
```python
fgraph.destroyers = [get_destroyers_of, has_destroyers]
```

@nouiz  what do you suggest?

fix https://github.com/Theano/Theano/issues/6230

### DIFF
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -195,7 +195,7 @@ def std_fgraph(input_specs, output_specs, accept_inplace=False):
                    for spec, input in zip(input_specs, fgraph.inputs)
                    if not (spec.mutable or
                            (hasattr(fgraph, 'destroyers') and
-                            fgraph.destroyers(input)))))
+                            fgraph.has_destroyers([input])))))
 
     # If named nodes are replaced, keep the name
     for feature in std_fgraph.features:
@@ -1090,7 +1090,7 @@ def insert_deepcopy(fgraph, wrapped_inputs, wrapped_outputs):
 
     # We can't use fgraph.inputs as this don't include Constant Value.
     all_graph_inputs = gof.graph.inputs(fgraph.outputs)
-    has_get_destroyers = hasattr(fgraph, 'get_destroyers_of')
+    has_destroyers_attr = hasattr(fgraph, 'has_destroyers')
 
     for i in xrange(len(fgraph.outputs)):
         views_of_output_i = set()
@@ -1121,7 +1121,7 @@ def insert_deepcopy(fgraph, wrapped_inputs, wrapped_outputs):
                 #    being updated
                 if input_j in updated_fgraph_inputs:
                     continue
-                if input_j in views_of_output_i and not (has_get_destroyers and fgraph.get_destroyers_of(input_j)):
+                if input_j in views_of_output_i and not (has_destroyers_attr and fgraph.has_destroyers([input_j])):
                     # We don't put deep_copy_op if the input and the
                     # output have borrow==True
                     if input_j in fgraph.inputs:
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -132,6 +132,11 @@ class Supervisor:
         self.protected = list(protected)
 
     def validate(self, fgraph):
+        if config.cycle_detection == 'fast' and hasattr(fgraph, 'fast_destroyers_check'):
+            if fgraph.fast_destroyers_check(self.protected):
+                raise gof.InconsistencyError("Trying to destroy a protected"
+                                             "Variable.")
+
         if not hasattr(fgraph, 'destroyers'):
             return True
         for r in self.protected + list(fgraph.outputs):
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -136,8 +136,7 @@ class Supervisor:
             if fgraph.has_destroyers(self.protected):
                 raise gof.InconsistencyError("Trying to destroy a protected"
                                              "Variable.")
-            else:
-                return True
+            return True
         if not hasattr(fgraph, 'destroyers'):
             return True
         for r in self.protected + list(fgraph.outputs):
--- a/theano/compile/function_module.py
+++ b/theano/compile/function_module.py
@@ -132,11 +132,12 @@ class Supervisor:
         self.protected = list(protected)
 
     def validate(self, fgraph):
-        if config.cycle_detection == 'fast' and hasattr(fgraph, 'fast_destroyers_check'):
-            if fgraph.fast_destroyers_check(self.protected):
+        if config.cycle_detection == 'fast' and hasattr(fgraph, 'has_destroyers'):
+            if fgraph.has_destroyers(self.protected):
                 raise gof.InconsistencyError("Trying to destroy a protected"
                                              "Variable.")
-
+            else:
+                return True
         if not hasattr(fgraph, 'destroyers'):
             return True
         for r in self.protected + list(fgraph.outputs):
@@ -1090,7 +1091,7 @@ def insert_deepcopy(fgraph, wrapped_inputs, wrapped_outputs):
 
     # We can't use fgraph.inputs as this don't include Constant Value.
     all_graph_inputs = gof.graph.inputs(fgraph.outputs)
-    has_destroyers = hasattr(fgraph, 'get_destroyers_of')
+    has_get_destroyers = hasattr(fgraph, 'get_destroyers_of')
 
     for i in xrange(len(fgraph.outputs)):
         views_of_output_i = set()
@@ -1121,7 +1122,7 @@ def insert_deepcopy(fgraph, wrapped_inputs, wrapped_outputs):
                 #    being updated
                 if input_j in updated_fgraph_inputs:
                     continue
-                if input_j in views_of_output_i and not (has_destroyers and fgraph.get_destroyers_of(input_j)):
+                if input_j in views_of_output_i and not (has_get_destroyers and fgraph.get_destroyers_of(input_j)):
                     # We don't put deep_copy_op if the input and the
                     # output have borrow==True
                     if input_j in fgraph.inputs:
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -395,6 +395,16 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         fgraph.destroyers = get_destroyers_of
 
         def has_destroyers(protected_list):
+            if not self.stale_droot:
+                droot, _, root_destroyer = self.refresh_droot_impact()
+                for protected_var in protected_list:
+                    try:
+                        root_destroyer[droot[protected_var]]
+                    except AttributeError:
+                        pass
+                    else:
+                        return True
+                return False
 
             def recursive_destroys_finder(protected_var):
                 # protected_var is the idx'th input of app.
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -395,15 +395,10 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         fgraph.destroyers = get_destroyers_of
 
         def has_destroyers(protected_list):
-            visited_app_set = set()
 
             def recursive_destroys_finder(protected_var):
                 # protected_var is the idx'th input of app.
                 for (app, idx) in protected_var.clients:
-                    if app in visited_app_set:
-                        continue
-                    else:
-                        visited_app_set.add(app)
                     if app == 'output':
                         continue
                     destroy_maps = getattr(app.op, 'destroy_map', {}).values()
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -297,7 +297,7 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         <unknown>
 
     """
-    pickle_rm_attr = ["destroyers"]
+    pickle_rm_attr = ["destroyers", "fast_destroyers_check"]
 
     def __init__(self, do_imports_on_attach=True, algo=None):
         self.fgraph = None
@@ -394,6 +394,26 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
                 return []
         fgraph.destroyers = get_destroyers_of
 
+        def recursive_destroys_finder(clients_list):
+            for client in clients_list:
+                # client is a tuple (I don't know if its size is always one)
+                for item in client:
+                    if item.op.destroy_map:
+                        return True
+                    if len(item.outputs) == 0:
+                        return False
+                    for output in item.outputs:
+                        if recursive_destroys_finder(output.clients):
+                            return True
+            return False
+
+        def fast_destroyers_check(protected_list):
+            for protected_var in protected_list:
+                if recursive_destroys_finder(protected_var.clients):
+                    return True
+
+        fgraph.fast_destroyers_check = fast_destroyers_check
+
     def refresh_droot_impact(self):
         """
         Makes sure self.droot, self.impact, and self.root_destroyer are up to
@@ -416,6 +436,7 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         del self.stale_droot
         assert self.fgraph.destroyer_handler is self
         delattr(self.fgraph, 'destroyers')
+        delattr(self.fgraph, 'fast_destroyers_check')
         delattr(self.fgraph, 'destroy_handler')
         self.fgraph = None
 
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -395,7 +395,7 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         fgraph.destroyers = get_destroyers_of
 
         def has_destroyers(protected_list):
-            if not self.stale_droot:
+            if self.algo != 'fast':
                 droot, _, root_destroyer = self.refresh_droot_impact()
                 for protected_var in protected_list:
                     try:
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -479,11 +479,11 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
                     if len(v) > 0:
                         self.fail_validate[app] = theano.gof.InconsistencyError(
                             "Destroyed variable has view_map. " + str(reason))
-                    elif d:
-                        d = d.get(inp_idx2, [])
-                        if len(d) > 0:
-                            self.fail_validate[app] = theano.gof.InconsistencyError(
-                                "Destroyed variable has destroy_map. " + str(reason))
+                elif d:
+                    d = d.get(inp_idx2, [])
+                    if len(d) > 0:
+                        self.fail_validate[app] = theano.gof.InconsistencyError(
+                            "Destroyed variable has destroy_map. " + str(reason))
 
                 # These 2 assertions are commented since this function is called so many times
                 # but they should be true.
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -400,7 +400,7 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
                 for protected_var in protected_list:
                     try:
                         root_destroyer[droot[protected_var]]
-                    except AttributeError:
+                    except KeyError:
                         pass
                     else:
                         return True
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -400,10 +400,9 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
                 for protected_var in protected_list:
                     try:
                         root_destroyer[droot[protected_var]]
+                        return True
                     except KeyError:
                         pass
-                    else:
-                        return True
                 return False
 
             def recursive_destroys_finder(protected_var):
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -250,7 +250,7 @@ def fast_inplace_check(inputs):
 
     inputs = [i for i in inputs if
               not isinstance(i, graph.Constant) and
-              not fgraph.destroyers(i) and
+              not fgraph.has_destroyers([i]) and
               i not in protected_inputs]
     return inputs
 
--- a/theano/gof/destroyhandler.py
+++ b/theano/gof/destroyhandler.py
@@ -297,7 +297,7 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         <unknown>
 
     """
-    pickle_rm_attr = ["destroyers", "fast_destroyers_check"]
+    pickle_rm_attr = ["destroyers", "has_destroyers"]
 
     def __init__(self, do_imports_on_attach=True, algo=None):
         self.fgraph = None
@@ -395,24 +395,23 @@ class DestroyHandler(toolbox.Bookkeeper):  # noqa
         fgraph.destroyers = get_destroyers_of
 
         def recursive_destroys_finder(clients_list):
-            for client in clients_list:
-                # client is a tuple (I don't know if its size is always one)
-                for item in client:
-                    if item.op.destroy_map:
+            for (app, idx) in clients_list:
+                if app == 'output':
+                    continue
+                destroy_maps = getattr(app.op, 'destroy_map', {}).values()
+                if idx in [dmap for sublist in destroy_maps for dmap in sublist]:
+                    return True
+                for var in getattr(app.op, 'view_map', {}).keys():
+                    if recursive_destroys_finder(app.outputs[var].clients):
                         return True
-                    if len(item.outputs) == 0:
-                        return False
-                    for output in item.outputs:
-                        if recursive_destroys_finder(output.clients):
-                            return True
             return False
 
-        def fast_destroyers_check(protected_list):
+        def has_destroyers(protected_list):
             for protected_var in protected_list:
                 if recursive_destroys_finder(protected_var.clients):
                     return True
 
-        fgraph.fast_destroyers_check = fast_destroyers_check
+        fgraph.has_destroyers = has_destroyers
 
     def refresh_droot_impact(self):
         """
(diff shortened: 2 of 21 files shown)


lamblin:
Now that #5794 has been merged, you will probably need to rebase on the master. That will make the diff easier to check.

theano/gof/destroyhandler.py
```diff
@@ -394,6 +394,17 @@ def get_destroyers_of(r):
                 return []
         fgraph.destroyers = get_destroyers_of
 
+        def has_destroyers(l):
+            for item in l:
+                droot, _, root_destroyer = self.refresh_droot_impact()
+                try:
+                    [root_destroyer[droot[item]]]
```
nouiz: I think you wanted this line to be:

```
if root_destroyer[droot[item]]:
     return True
```

theano/gof/destroyhandler.py
```diff
@@ -394,6 +394,17 @@ def get_destroyers_of(r):
...
+                    return True
+                except Exception:
```
nouiz: I know you copied the Exception from above, but this is a bad practice. We should try to catch a much more precise exception.

Anyway, this isn't the optimized algorythm anyway. Both otherwise, you seem to have done the first part well. If with my modif in the other comment and tests pass, the structural change would be done. It would only miss the optimized implementation here.

theano/gof/destroyhandler.py
ReyhaneAskari: yeah! thanks.

theano/gof/destroyhandler.py
ReyhaneAskari: You are right! Thanks I changed it to ```KeyError```

theano/compile/function_module.py
```diff
@@ -135,10 +135,15 @@ def __init__(self, protected):
     def validate(self, fgraph):
         if not hasattr(fgraph, 'destroyers'):
             return True
-        for r in self.protected + list(fgraph.outputs):
-            if fgraph.destroyers(r):
+        if config.cycle_detection == 'fast':
+            if fgraph.destroyers[1](self.protected + list(fgraph.outputs)):
```
ReyhaneAskari: @nouiz , do we need to pass ```fgraph.outputs``` here ?

nouiz: I think so.

theano/gof/destroyhandler.py
```diff
@@ -394,21 +394,25 @@ def get_destroyers_of(r):
...
+                    outputs_clients_set.update(app.outputs[var].clients)
+                if recursive_destroys_finder(outputs_clients_set):
+                    return True
             return False
 
         def has_destroyers(protected_list):
+            clients_set = set()
             for protected_var in protected_list:
-                if recursive_destroys_finder(protected_var.clients):
+                clients_set.update(protected_var.clients)
```
nouiz: the clients list of tuple (app, idx) is already a set.

nouiz: So mostly, that commit don't help. There is some saving possible if one node reuse the same variable multiple time as input, but it happen so infrequently that I don't think the extra complexity is useful.

ReyhaneAskari:
I removed that commit. Now we need to check the failing tests and make sure they are the same tests that are failing in the master with ```THEANO_FLAGS=cycle_detetion='fast'```.  @lamblin suggested that I mark the failing tests in master as nosetests' known failures.

nouiz:
Just to be sure, make them as known failure only when cycle='fast'

theano/gof/destroyhandler.py
```diff
@@ -394,6 +394,31 @@ def get_destroyers_of(r):
                 return []
         fgraph.destroyers = get_destroyers_of
 
+        def has_destroyers(protected_list):
+            visited_app_set = set()
+
+            def recursive_destroys_finder(clients_list):
```
nouiz: I would change the interface to have this function take a variable as input. Then it would loop over var.clients itself. I think it is a nicer interface and will help the conversion to the recursion free version.

theano/gof/destroyhandler.py
```diff
@@ -394,6 +394,31 @@ def get_destroyers_of(r):
...
+                    if app in visited_app_set:
+                        continue
+                    else:
+                        visited_app_set.add(app)
+                    if app == 'output':
+                        continue
+                    destroy_maps = getattr(app.op, 'destroy_map', {}).values()
+                    if idx in [dmap for sublist in destroy_maps for dmap in sublist]:
+                        return True
+                    for var in getattr(app.op, 'view_map', {}).keys():
```
nouiz: Here, this is too much restrictif. We are asking that all view of app are not destroyed. But we only care the outputs that are a view of app.inputs[idx].

theano/gof/destroyhandler.py
ReyhaneAskari: Sure. Thanks.

theano/gof/destroyhandler.py
ReyhaneAskari: I see. Thanks. Is this check good? 
``` python
for var in getattr(app.op, 'view_map', {}).keys():
    if idx in app.op.view_map[var] and recursive_destroys_finder(app.outputs[var]):
          return True

```

ReyhaneAskari:
@lamblin, @nouiz I could only find ```unittest.expectedFailure``` as a decorator. To make sure it only adds when ```cycle_detection='fast'```, I wrote another decorator like this because there were no builtin options:

```python
import unittest
from theano import config

def expectedFailure_fast():
    return unittest.expectedFailure if config.cycle_detection == 'fast' else lambda x: x

@expectedFailure_fast()
def test_usage_loop_through_views_2():
```
Do you have any suggestions on where to add this code snippet? I was thinking maybe ```theano.gof.utils```?

theano/gof/destroyhandler.py
nouiz: that seem good. To help read it, I would rename var to var_idx, to tell it is an index, not a variable.

nouiz:
theano.gof.utils seem the right place for that.

theano/gof/destroyhandler.py
```diff
@@ -409,12 +409,12 @@ def recursive_destroys_finder(clients_list):
                     if idx in [dmap for sublist in destroy_maps for dmap in sublist]:
                         return True
                     for var in getattr(app.op, 'view_map', {}).keys():
-                        if recursive_destroys_finder(app.outputs[var].clients):
-                            return True
+                        if idx in app.op.view_map[var] and recursive_destroys_finder(app.outputs[var]):
+                                return True
```
nouiz: you added an extra indentation that isn't useful.

nouiz:
theano.tests.unittests_tools would be better then theano.gof.utils I think.

theano/gof/destroyhandler.py
ReyhaneAskari: Sure. Thanks.

lamblin:
Most issues seem to be:
```
 Error Details

OutputGuard.0 is a view/destroyed version of more then one inputs. Currently, we only support the case where an output is a view or a destroyed version of one input.

 Stack Trace

  File "/miniconda/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/jenkins/workspace/Theano_PR/theano/typed_list/tests/test_basic.py", line 257, in test_inplace
    accept_inplace=True)
  File "/home/jenkins/workspace/Theano_PR/theano/compile/function.py", line 326, in function
    output_keys=output_keys)
  File "/home/jenkins/workspace/Theano_PR/theano/compile/pfunc.py", line 486, in pfunc
    output_keys=output_keys)
  File "/home/jenkins/workspace/Theano_PR/theano/compile/function_module.py", line 1814, in orig_function
    output_keys=output_keys)
  File "/home/jenkins/workspace/Theano_PR/theano/compile/function_module.py", line 1491, in __init__
    insert_deepcopy(fgraph, inputs, outputs + additional_outputs)
  File "/home/jenkins/workspace/Theano_PR/theano/compile/function_module.py", line 1099, in insert_deepcopy
    view_tree_set(alias_root(fgraph.outputs[i]), views_of_output_i)
  File "/home/jenkins/workspace/Theano_PR/theano/compile/function_module.py", line 57, in alias_root
    str(v) + " is a view/destroyed version of more then one inputs. "
NotImplementedError: OutputGuard.0 is a view/destroyed version of more then one inputs. Currently, we only support the case where an output is a view or a destroyed version of one input.
```

theano/tensor/tests/test_sharedvar.py
```diff
@@ -27,6 +27,7 @@ def makeSharedTester(shared_constructor_,
                      theano_fct_,
                      ref_fct_,
                      cast_value_=np.asarray,
+                     need_decorator=True,
```
nouiz: Don't forget @lamblin comment in gh-6014.

ReyhaneAskari: Yes, I'm on it.

ReyhaneAskari:
There was a small conflict here with jenkins and travis wouldn't run with out resolving it. I resolved the conflict online committed it. The name was automatically generated as ```Merge branch 'master' into new_destroy_handler```.  I will have to rebase in the end and I will clean this up.

ReyhaneAskari:
@lamblin @nouiz 
There were so many tests that were failing due to the insertion of ```OutputGurad```. I checked several tests and I figured that we should raise an error when [AdddestoryHandler tries to insert the ```OutputGurad``` ](https://github.com/Theano/Theano/blob/be022807cc9e3f8115a01ad2343c57f819f13ad9/theano/compile/mode.py#L116). So we need to check the ```graph.outputs``` as well as the ```protected_var```.  I also changed the ```has_destroyer``` such that in the case when the ```protected_var``` is an outputGuard, we do the check. 

I was not sure if there are cases where ```app == 'output'``` but the ```protected_var``` is not an OutputGuard. There are still some failing tests that I'm looking into.

ReyhaneAskari:
There were many failures in the tests because we were inplace in some cases where we shouldn't have been. I checked these tests and found out that we are not raising the error in the ```Supervisor``` class in some cases that the inplace was wrong. The problem lied in the case where we had a graph in which one node was cached in the ```visited_app_set``` and never checked again. 

The mechanism of ```visited_app_set``` is wrong. Consider the following case when an apply node is the client of serval variable nodes and has a destroy map of {0 : [1]}. In the first visit to this node since the destroy map is on the second input, we do not raise any error and we mark the apply node as visited while it should be checked for the second variable node as well where it is destroying one of its inputs. 

photo_2017-06-12_13-55-52

I guess it will be too complicated to support these cases. For now I removed the ```visited_app_set``` to see if I all the tests pass. Then in the benchmarking we can figure where is the bottleneck.

lamblin:
Conclusion of IRL discussion:
- remove view_map from OutputGuard
- In AddDestroyHandler, do not try to introduce OutputGuard when fast cycle detection is on
- undo the new check (full=True).

ReyhaneAskari:
We won't remove the view_map for the output Guard for now since it's causing many errors. We will handle that in another PR where we'll completely remove the output Guard.

ReyhaneAskari:
Is there a way to add the flag to the examples inside the docs? Maybe the easier way to change the cycle_detection flag was to change the default value to ```fast``` for the testing purposes and revert it back in the end.

nouiz:
Doing your suggestion would have done it.

ReyhaneAskari:
I cleaned up some commits and also removed the changes to jenkins and travis and changed the cycle_detection flag to have ```fast``` as the default value.

ReyhaneAskari:
Just to make sure we don't have any errors on the master.

ReyhaneAskari:
Here is the results of profiling:

|               |   Function.call    | total compile time | Opt time    | Linker time   | Time since theano import |
|---------------|--------------------|--------------------|-------------|---------------|--------------------------|
| [master_2](https://github.com/Theano/Theano/files/1076079/master_2.txt)                  | 2317.635           | 586.9048           | 553.4296      | 22.69159                 | 3075.815      |
| [master_3](https://github.com/Theano/Theano/files/1076080/master_3.txt)                  | 2320.0             | 578.2175           | 545.9906      | 21.54995                 | 3058.587      |
| **avg master_2 master_3** | **2318.8175**      | **582.56115**      | **549.7101**  | **22.12077**             | **3067.201**  |
| [faster_2](https://github.com/Theano/Theano/files/1076078/faster_2.txt)                  | 2332.769           | 300.8603           | 269.1528      | 20.46626                 | 2796.888      |
| [faster_3](https://github.com/Theano/Theano/files/1076077/faster_3.txt)                  | 2334.987           | 298.6716           | 266.9097      | 20.59228                 | 2802.241      |
| **avg faster_2 faster_3** | **2333.878**       | **299.76595**      | **268.03125** | **20.52927**             | **2799.5645** |

Memory profiling:

||master|faster|
|----|--|--|
| peak memory cpu                      |       3.0 |       2.0 |
| peak memory gpu                      | 2296853.0 | 2186307.0 |
| peak memory cpu + gpu                | 2296855.0 | 2186309.0 |
| flag = optimizer_excluding cpu       |       3.0 |       2.0 |
| flag = optimizer_excluding gpu       | 2313898.0 | 2199204.0 |
| flag = optimizer_excluding cpu + gpu | 2313901.0 | 2199206.0 |
| allow_gc=False cpu                   |      47.0 |      47.0 |
| allow_gc=False gpu                   | 8456750.0 | 8693082.0 |
| allow_gc=False cpu + gpu             | 8456797.0 | 8693129.0 |

I also did the python profiling you can check them in the attached files. ```get_destroyers_of``` has the largest cumulative time in faster. Can we remove all the calls to it?

theano/tensor/opt.py
```diff
@@ -265,8 +265,8 @@ def apply(self, fgraph):
                 candidate_inputs = [i for i in xrange(len(node.inputs))
                                     if i not in baseline.values() and
                                     not isinstance(node.inputs[i], Constant) and
-                                    # Is next line costly?
-                                    not fgraph.destroyers(node.inputs[i]) and
+                                    # Is next line costly? (used to be fgraph.get_destroyers)
```
nouiz: I would update the comment to:

the next line should not be costly most of the time.

theano/gof/destroyhandler.py
```diff
@@ -394,6 +394,32 @@ def get_destroyers_of(r):
                 return []
         fgraph.destroyers = get_destroyers_of
 
+        def has_destroyers(protected_list):
```
nouiz: I think I would do something like this at the start of this method:

```
if not self.stale_droot:
    for var in protected_var:
        if self.destroyers(var):
            raise ...
```

This do 2 things:
1) We should always call has_destroyers what ever the value of cycle_detection is.
2) This could speed up or slow down the case when cycle_detection is the default value. So we need to speed test that. If is don't speed up computation, we should change the if to be `if self.cycle_detection != 'fast':`

What do you think of that?

theano/compile/function_module.py
```diff
@@ -133,6 +133,11 @@ def __init__(self, protected):
...
 
     def validate(self, fgraph):
+        if config.cycle_detection == 'fast' and hasattr(fgraph, 'has_destroyers'):
+            if fgraph.has_destroyers(self.protected):
+                raise gof.InconsistencyError("Trying to destroy a protected"
+                                             "Variable.")
+            return True
         if not hasattr(fgraph, 'destroyers'):
             return True
         for r in self.protected + list(fgraph.outputs):
```
nouiz: With the other change proposed, we could replace the full implementation of this method to do the same thing for all values of cycle_detection.

nouiz:
In the file compile/debugmode.py, the function _optcheck_fgraph need to be updated.

I checked this function and in fact, it should be the same as std_fgraph() + the addition of the EquivalenceTracker. It is currently different by mistake, as it wasn't updated as it should!

So, can you update this function to call std_fgraph and then add the EquivalentTracker to the fgraph. This will prevent futur error of not updating both function.

nouiz:
I think jenkins error isn't related to this PR and will be fixed at the next run.

theano/tensor/opt.py
ReyhaneAskari: Sure. Thanks.

theano/gof/destroyhandler.py
ReyhaneAskari: Are you sure that ```self.stale_droot``` is a good condition? I quickly implemented this : 

```python
Supervisor:
def validate(self, fgraph):
        if not hasattr(fgraph, 'has_destroyers'):
            return True
        if fgraph.has_destroyers(self.protected):
            raise gof.InconsistencyError("Trying to destroy a protected"
                                         "Variable.")
DestroyHandler:
def unpickle(self, fgraph):
        def get_destroyers_of(r):
            droot, _, root_destroyer = self.refresh_droot_impact()
            try:
                return [root_destroyer[droot[r]]]
            except Exception:
                return []
        fgraph.destroyers = get_destroyers_of

        def recursive_destroys_finder(protected_var):
            # protected_var is the idx'th input of app.
            for (app, idx) in protected_var.clients:
                if app == 'output':
                    continue
                destroy_maps = getattr(app.op, 'destroy_map', {}).values()
                # If True means that the apply node, destroys the protected_var.
                if idx in [dmap for sublist in destroy_maps for dmap in sublist]:
                    return True
                for var_idx in getattr(app.op, 'view_map', {}).keys():
                    if idx in app.op.view_map[var_idx]:
                        # We need to recursivly check the destroy_map of all the
                        # outputs that we have a view_map on.
                        if recursive_destroys_finder(app.outputs[var_idx]):
                            return True
            return False

        def has_destroyers(protected_list):
            if not self.stale_droot:
                for var in protected_list + list(fgraph.outputs):
                    if get_destroyers_of(var):
                        return True
                return False

            for protected_var in protected_list:
                if recursive_destroys_finder(protected_var):
                    return True
            return False

        fgraph.has_destroyers = has_destroyers
```

But I had many errors with the OutputGuard:
```
NotImplementedError: ('OutputGuard.0 is a view/destroyed version of more then one inputs. Currently, we only support the case where an output is a view or a destroyed version of one input.', 'Test <function <lambda> at 0x7fac379dbb18>::x2: Error occurred while computing the gradient on the following inputs: [array([[-0.07363263, -0.66992872,  0.980925  ]])]')
```

 So I guess the condition is not correct. I think the could would be nicer with your change but why do you think this will be a speedup?

ReyhaneAskari:
jenkins test this.

ReyhaneAskari:
I'm going to create another PR for this:
> In the file compile/debugmode.py, the function _optcheck_fgraph need to be updated...

ReyhaneAskari:
The test suit is running with ```cycle_detection = regular```. I will change it to ```fast``` when they all passed.

ReyhaneAskari:
jenkins test this.

ReyhaneAskari:
I didn't get the error for jenkins locally.

nouiz:
test passed. Can you redo the Theano opt profile to see if this remove all the call that are slow?

ReyhaneAskari:
Sure. I'll update you when the results were available.

ReyhaneAskari:
I added the last commit to make sure the tests will pass when the default of ```cycle_detection```  is set to ```fast```.

ReyhaneAskari:
I don't get jenkins' errors locally. jenkins test this.

ReyhaneAskari:
Here is the results of profiling. I also did the python profiling you can check them in the attached files. ```get_destroyers_of``` is no longer the appearing on the top of the list with faster. 

|               |   Function.call    | total compile time | Opt time    | Linker time   | Time since theano import |
|---------------|--------------------|--------------------|-------------|---------------|--------------------------|
| [master_2](https://github.com/Theano/Theano/files/1091779/master_2.txt)                  | 2304.337           | 469.5146           | 437.3199      | 21.41075                 | 2936.39      |
| [master_3](https://github.com/Theano/Theano/files/1091778/master_3.txt)                  | 2305.517           | 470.0153           | 437.6318      | 21.52863                 | 2937.23      |
| **avg master_2 master_3** | **2304.927**       | **469.76495**      | **437.47585** | **21.46969**             | **2936.81**  |
| [faster_2](https://github.com/Theano/Theano/files/1091776/faster_2.txt)                  | 2305.755           | 242.6038           | 211.492       | 20.39481                 | 2719.367     |
| [faster_3](https://github.com/Theano/Theano/files/1091777/faster_3.txt)                  | 2309.326           | 245.9866           | 214.8715      | 20.31912                 | 2727.227     |
| **avg faster_2 faster_3** | **2307.5405**      | **244.2952**       | **213.18175** | **20.356965**            | **2723.297** |


Memory profiling: 

||master|faster|
|----|--|--|
| peak memory cpu                      |       3.0 |       2.0 |
| peak memory gpu                      | 2293322.0 | 2186307.0 |
| peak memory cpu + gpu                | 2293324.0 | 2186309.0 |
| flag = optimizer_excluding cpu       |       3.0 |       2.0 |
| flag = optimizer_excluding gpu       | 2306834.0 | 2199204.0 |
| flag = optimizer_excluding cpu + gpu | 2306837.0 | 2199206.0 |
| allow_gc=False cpu                   |      47.0 |      47.0 |
| allow_gc=False gpu                   | 8392359.0 | 8693082.0 |
| allow_gc=False cpu + gpu             | 8392406.0 | 8693129.0 |

slefrancois:
jenkins test this

nouiz:
jenkins test this

ReyhaneAskari:
jenkins test this.

ReyhaneAskari:
jenkins test this.

theano/compile/mode.py
```diff
@@ -111,6 +111,8 @@ class AddDestroyHandler(gof.Optimizer):
 
     """
     def apply(self, fgraph):
+        if theano.config.cycle_detection == 'fast':
```
nouiz: Why this change? I didn't expected this PR to change the tests that fail with cycle_detection=False. I would not be surprised that this change isn't needed and removing it "fix" the failing tests.

ReyhaneAskari: This is what we had discussed previously with Pascal. When we have the ```fast```, we don't want to have the check for the ```outputguard```.  There is no failing test on this PR. All the tests have passed.

nouiz:
This PR skip many new test. I don't think this is good.

If the destroy handler isn't there,  then from memory, I'm in the subway,
has_destroyers won't be there, as it is the destroy handler that add it.

Le jeu. 29 juin 2017 17:55, Reyhane Askari <notifications@github.com> a
écrit :

ReyhaneAskari:
Which new tests do you mean?

nouiz:
This PR create a decorator to mess with many tests. I don't think this is
good. We probably need to discuss this live later. This PR was just about
removing call to build the internal structure. You can't just not add the
destroy handler when cycle_detection is fast, for the simple reason that
"fast" is implemented by the destroy handler itself. So mostly, you remove
all the checks in fast, even the fast check.

ReyhaneAskari:
After a review with @nouiz , he found that there was a minor indentation problem inside ```fast_destroy``` function (my last commit).  This change added 3 new failures which are probably due to the new check for ```destroy_handler```. I think they should be added to the ```expectedFailure_fast```.

nouiz:
Agreed to take the 5 new failure with expected failure in fast cyclde_detection

ReyhaneAskari:
Fixed the problem and rebased due to the conflicts that were introduced after the merge of the branch which removed the outputguard.

lamblin:
I think the issue in DebugMode is that we should skip trying to insert OutputGuard if cycle_detection is "fast". See https://github.com/Theano/Theano/pull/6044#pullrequestreview-50136155

lamblin:
The remaining issues seem to come from a change in the behaviour of `expectedFailure` between python 2 and 3.
For python 2, `expectedFailure` catches failures as well as Exception being raised.
For python 3, since Exceptions cause ERROR instead of FAIL, the whole thing is considered as an issue.

In that case, I think we would need to write our own decorator, consistent in python 2 and 3 instead.

ReyhaneAskari:
Thanks Pascal. I checked and there is no decorator in ```unittest``` to mark a test as expected error! 

So I'd suggest that we skip these tests for now:

```python

    if theano.config.cycle_detection == 'fast':
        return unittest.skip("cycle_detection is set to fast."
                             "Many inplaces are not allowed any more.")(f)
    else:
        return f
```

theano/tests/unittest_tools.py
```diff
@@ -447,11 +447,16 @@ def attempt_multiple_times(*args, **kwargs):
...
     THEANO_FLAGS =cycle_detection='fast'.
     """
     if theano.config.cycle_detection == 'fast':
-        return unittest.expectedFailure(f)
+        class TestAssertion(unittest.TestCase):
+            def runTest(self, *args, **kwargs):
+                with self.assertRaises(Exception):
+                    f(*args, **kwargs)
+        test_assertion = TestAssertion()
+        return test_assertion
```
lamblin: I think you can simplify this a bit by importing
```
from nose.tools import assert_raises
```
and then
```
def assertFailure_fast(f):
    if theano.config.cycle_detection == 'fast':
        def new_test(*args, **args):
            assert_raises(f, *args, **kwargs)
        return new_test
    else:
        return f
```
rather than defining and instanciating a whole new class

lamblin:
Also, I'm not sure why the tests for cholesky and matrix inverse with magma are failing, rather than just rejecting the optimization and failing later.

ReyhaneAskari:
jenkins test this.

theano/tests/unittest_tools.py
ReyhaneAskari: Thanks, I didn't know about this one!

ReyhaneAskari:
I did a rebase locally and the tests passed but since there were no changes I couldn't push. So I made a dummy commit. Hopefully the tests will pass now.

ReyhaneAskari:
I can't reproduce jenkins' errors. Do you have any ideas why this is happening? @lamblin

ReyhaneAskari:
I guess this is finally good to merge. Do we want the ```cycle_detection``` flag's default be ```fast``` or `regular`?

theano/gof/destroyhandler.py
nouiz: I implemented it in 5e0cc2b7c89da35da27beb3302034d6befa9135d

PR #6005, opened 2017-06-03
Fix min function to work with uint* and bool

### LINKED ISSUE #5791: Implement tensor.min() for bool and uint*
Currently, it raises NotImplementedError.
It could use tensor.max(), like for float and int, or maybe in that case actually use a "minimum" reduction.

See [here](https://github.com/Theano/Theano/blob/372ae95ef9aae402b7f426d159585d26556eb0a1/theano/tensor/basic.py#L1734)

naitonium:
fix #5791

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -1715,6 +1715,11 @@ def min(x, axis=None, keepdims=False):
     str_x_type = str(x.dtype)
     if str_x_type.startswith('float') or str_x_type in int_dtypes:
         return -max(-x, axis=axis, keepdims=keepdims)
+    elif str_x_type in uint_dtypes:
+        itype = np.iinfo(x.dtype)
+        return itype.max - max(itype.max - x, axis=axis, keepdims=keepdims)
+    elif str_x_type == 'bool':
+        return ~max(~x, axis=axis, keepdims=keepdims)
     else:
         # Be careful about unsigned integers, complex
         raise NotImplementedError()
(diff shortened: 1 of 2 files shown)


nouiz:
The implementation look good, but this would request tests. You can extend the current test inside the file: theano/tensor/tests/test_basic.py
Check the classes: T_argmin_argmax and T_min_max

thanks

nouiz:
We should do a Theano 0.10.beta3 Monday. I don't have time to add the tests, but if someone can do that, we could include it in that release.

nouiz:
@gvtulder thanks for the tests. Do you understand why we need to handle bool/uint different for the min function? We don't do that for the max function. So I don't understand the reason for the fix in the original issue. I'll comment in the code what I think would be the issue.

nouiz:
Now I get the original issue. We build a graph with max in the graph for min! I forgot that "detail" :)

theano/tensor/basic.py
```diff
@@ -1715,6 +1715,11 @@ def min(x, axis=None, keepdims=False):
     str_x_type = str(x.dtype)
     if str_x_type.startswith('float') or str_x_type in int_dtypes:
         return -max(-x, axis=axis, keepdims=keepdims)
+    elif str_x_type in uint_dtypes:
+        itype = np.iinfo(x.dtype)
+        return itype.max - max(itype.max - x, axis=axis, keepdims=keepdims)
```
nouiz: The only problem I have with that is that it upcast the result to an higher dtype then the original dtype.

I think this can be easily fixed by this change:

    itype.max -> np.asarray(itype.max, dtype=x.dtype)

Can you try that and add in the test an assert that the output dtype is the same as the input dtype?

gvtulder: Good point, fixed.

gvtulder:
Travis passed. CircleCI fails because this PR is old enough to miss a CircleCI configuration file.

Issue #5620, opened 2017-02-24
Add unshared (Locally-connected) convolution


nouiz:
There was such op in Pylearn2:

https://github.com/lisa-lab/pylearn2/tree/master/pylearn2/packaged_dependencies/theano_linear/unshared_conv

But before working on this, check if there is a better implementation somewhere else.

It will need to be ported to the new back-end.

f0k:
> But before working on this, check if there is a better implementation somewhere else.

A possible alternative would be a im2col-based implementation. We have a pure Theano implementation in Lasagne to compare performance against.

affanv14:
Wont there be a huge memory overhead with im2col and unshared convolutions combined?

f0k:
> Wont there be a huge amount of memory used with im2col

Memory use for the `im2col` operation would be the same. But the way `im2col` works, we can't use gemm for the reduction. We'd have to check what's the best option. Not sure if batched gemm would be a good idea.

nouiz:
Batched gemm in cublas was optimized for "small" matrix, up to 128 or 256
on each dimensions. I don't know how efficient it is for bigger matrix.
This could be tested with a small Theano graph that just do a batched gemm
with the right shape to see how efficient it is at those sizes.

Do you know the sizes on which it would be called?

f0k:
> Do you know the sizes on which it would be called?

In the forward pass, `im2col` is called for each item in a batch. Each item has a shape of `(channels, h, w)` and will be transformed to `(regions, channels*kh*kw)` (where `regions` is the number of positions the filter kernel will be applied).
In normal convolution, the weight tensor would be reshaped to `(filters, channels*kh*kw)`, so you can apply `gemm` with the second argument transposed, and obtain a `(filters, regions)` output.
In unshared convolution, the weight tensor would be reshaped to `(filters, regions, channels*kh*kw)`. With these shapes, what we'd want to compute would be `(im2col(input) * weights).sum(axis=-1)`, so we get an output of `(filters, regions)` as in normal convolution. With a different layout of shapes, this could be done as `regions` many matrix products of a `(1, channels*kh*kw)` and a `(channels*kh*kw, filters)` matrix. So typically, the matrices would be up to maybe 256*9 x 256, and the number of matrices depends on the image size and stride. And actually a batched gemv would suffice, but that doesn't exist.

Before we benchmark a batched gemm, we'd need to think through if we can actually provide the weight tensor in the correct memory layout for this to work for the forward and two backward passes.

[...]

PR #6007, opened 2017-06-04
Unshared convolution

vikramnitin9:
Related to #5620 .

Task list :

- [x] C code for grad inputs and tests

- [x] All tests need to pass.

- [x] Change weight tensor dimension ordering instead of dimshuffle

- [x] Integrate unshared tests with AbstractConv

- [x] GPU code

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -1058,8 +1058,12 @@ class GpuCorrMM_gradWeights(BaseGpuCorrMM):
             assert shape[0].ndim == 0
             assert shape[1].ndim == 0
 
-        broadcastable = [topgrad.type.broadcastable[1], img.type.broadcastable[1],
-                         False, False]
+        if self.unshared:
+            broadcastable = [topgrad.type.broadcastable[0], False, False,
+                             img.type.broadcastable[1], False, False]
+        else:
+            broadcastable = [topgrad.type.broadcastable[1], img.type.broadcastable[1],
+                             False, False]
         return Apply(self, [img, topgrad] + height_width, [GpuArrayType(dtype=img.dtype,
                                                                         context_name=ctx_name,
                                                                         broadcastable=broadcastable)()])
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -690,8 +690,8 @@ class BaseGpuCorrMM(CGpuKernelBase):
     size_t kH, kW, dil_kH, dil_kW;
     if (direction != 1) {
         // weight is an input variable, we can just read its shape
-        kH = PyGpuArray_DIMS(weights)[2];
-        kW = PyGpuArray_DIMS(weights)[3];
+        kH = PyGpuArray_DIMS(weights)[wdim-2];
+        kW = PyGpuArray_DIMS(weights)[wdim-1];
     }
     else {
         if (%(height)s != -1) {
@@ -909,7 +909,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
         if (odim == 6) {
             PyErr_Format(PyExc_RuntimeError,
                     "BaseGpuCorrMM: Failed to allocate output of %%lld x %%lld x %%lld x %%lld %%lld %%lld",
-                    out_dim[0], out_dim[1], out_dim[2], out_dim[3]);
+                    out_dim[0], out_dim[1], out_dim[2], out_dim[3], out_dim[4], out_dim[5]);
         }
         %(fail)s
     }
@@ -1011,11 +1011,13 @@ class GpuCorrMM(BaseGpuCorrMM):
         top = gpu_contiguous(top)
         d_bottom = GpuCorrMM_gradInputs(self.border_mode,
                                         self.subsample,
-                                        self.filter_dilation)(
+                                        self.filter_dilation,
+                                        self.unshared)(
             weights, top, bottom.shape[-2:])
         d_weights = GpuCorrMM_gradWeights(self.border_mode,
                                           self.subsample,
-                                          self.filter_dilation)(
+                                          self.filter_dilation,
+                                          self.unshared)(
             bottom, top, weights.shape[-2:])
         return d_bottom, d_weights
 
@@ -1081,11 +1083,12 @@ class GpuCorrMM_gradWeights(BaseGpuCorrMM):
         weights = gpu_contiguous(weights)
         d_bottom = GpuCorrMM_gradInputs(self.border_mode,
                                         self.subsample,
-                                        self.filter_dilation)(weights,
-                                                              top,
-                                                              bottom.shape[-2:])
+                                        self.filter_dilation,
+                                        self.unshared)(weights,
+                                                       top,
+                                                       bottom.shape[-2:])
         d_top = GpuCorrMM(
-            self.border_mode, self.subsample, self.filter_dilation)(bottom, weights)
+            self.border_mode, self.subsample, self.filter_dilation, self.unshared)(bottom, weights)
         d_height_width = (
             theano.gradient.DisconnectedType()(),
             ) * 2 if len(inp) == 4 else ()
@@ -1137,7 +1140,7 @@ class GpuCorrMM_gradInputs(BaseGpuCorrMM):
             assert shape[0].ndim == 0
             assert shape[1].ndim == 0
 
-        broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[1],
+        broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[-3],
                          False, False]
         return Apply(self, [kern, topgrad] + height_width, [GpuArrayType(dtype=topgrad.dtype,
                                                                          context_name=ctx_name,
@@ -1156,12 +1159,14 @@ class GpuCorrMM_gradInputs(BaseGpuCorrMM):
         bottom = gpu_contiguous(bottom)
         d_weights = GpuCorrMM_gradWeights(self.border_mode,
                                           self.subsample,
-                                          self.filter_dilation)(bottom,
-                                                                top,
-                                                                weights.shape[-2:])
+                                          self.filter_dilation,
+                                          self.unshared)(bottom,
+                                                         top,
+                                                         weights.shape[-2:])
         d_top = GpuCorrMM(self.border_mode,
                           self.subsample,
-                          self.filter_dilation)(bottom, weights)
+                          self.filter_dilation,
+                          self.unshared)(bottom, weights)
         d_height_width = (
             theano.gradient.DisconnectedType()(),
             ) * 2 if len(inp) == 4 else ()
@@ -1783,7 +1788,7 @@ class GpuCorr3dMM_gradInputs(BaseGpuCorr3dMM):
             assert shape[1].ndim == 0
             assert shape[2].ndim == 0
 
-        broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[1],
+        broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[-3],
                          False, False, False]
         return Apply(self, [kern, topgrad] + height_width_depth,
                      [GpuArrayType(dtype=topgrad.dtype,
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -767,7 +767,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
                 PyErr_Format(PyExc_ValueError,
                              "GpuCorrMM: impossible output shape\\n"
                              "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
-                             "  weights shape: %%ld x %%ld x %%ld x %%ld %%ld %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld x %%ld x %%ld\\n"
                              "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
                              PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
                              PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
@@ -812,7 +812,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
                 PyErr_Format(PyExc_ValueError,
                              "GpuCorrMM backprop wrt. weights: impossible output shape\\n"
                              "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
-                             "  weights shape: %%ld x %%ld x %%ld x %%ld %%ld %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld x %%ld x %%ld\\n"
                              "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
                              PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
                              PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
@@ -855,7 +855,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
                 PyErr_Format(PyExc_ValueError,
                              "GpuCorrMM backprop wrt. inputs: impossible output shape\\n"
                              "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
-                             "  weight shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weight shape: %%ld x %%ld x %%ld x %%ld x %%ld x %%ld\\n"
                              "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
                              out_dim[0], out_dim[1], out_dim[2], out_dim[3],
                              PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -798,8 +798,8 @@ class BaseGpuCorrMM(CGpuKernelBase):
         out_dim[0] = PyGpuArray_DIMS(top)[1];
         if (unshared){
             odim = 6;
-            out_dim[1] = (npy_intp)PyGpuArray_DIMS(top)[2];
-            out_dim[2] = (npy_intp)PyGpuArray_DIMS(top)[3];
+            out_dim[1] = PyGpuArray_DIMS(top)[2];
+            out_dim[2] = PyGpuArray_DIMS(top)[3];
         }
         out_dim[wdim-3] = PyGpuArray_DIMS(bottom)[1];
         out_dim[wdim-2] = kH;  // already inferred further above
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -496,13 +496,15 @@ class BaseGpuCorrMM(CGpuKernelBase):
         Perform subsampling of the output (default: (1, 1)).
     filter_dilation
         Perform subsampling of the input, also known as dilation (default: (1, 1)).
+    unshared
+        Perform unshared correlation (default: False)
     """
     check_broadcast = False
-    __props__ = ('border_mode', 'subsample', 'filter_dilation')
+    __props__ = ('border_mode', 'subsample', 'filter_dilation', 'unshared')
     _f16_ok = True
 
     def __init__(self, border_mode="valid", subsample=(1, 1),
-                 filter_dilation=(1, 1)):
+                 filter_dilation=(1, 1), unshared=False):
         if isinstance(border_mode, integer_types):
             border_mode = (border_mode, border_mode)
         if isinstance(border_mode, tuple):
@@ -521,6 +523,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
             raise ValueError("filter_dilation must have two elements")
         self.subsample = tuple(subsample)
         self.filter_dilation = tuple(filter_dilation)
+        self.unshared = unshared
         CGpuKernelBase.__init__(self, ['corr_gemm.c'])
 
     @property
@@ -530,11 +533,12 @@ class BaseGpuCorrMM(CGpuKernelBase):
         return (0, 0)
 
     def __str__(self):
-        return '%s{%s, %s, %s}' % (
+        return '%s{%s, %s, %s, %s}' % (
             self.__class__.__name__,
             self.border_mode,
             str(self.subsample),
-            str(self.filter_dilation))
+            str(self.filter_dilation),
+            str(self.unshared))
 
     def flops(self, inp, outp):
         """
@@ -609,6 +613,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
         """
         dH, dW = self.subsample
         dilH, dilW = self.filter_dilation
+        unshared = int(self.unshared)
         if self.border_mode == "half":
             padH = padW = -1
         elif self.border_mode == "full":
@@ -669,11 +674,16 @@ class BaseGpuCorrMM(CGpuKernelBase):
     size_t dilW = %(dilW)s;
     int padH = %(padH)s;
     int padW = %(padW)s;
+    int unshared = %(unshared)s;
 
     PyGpuArrayObject * bottom = %(bottom)s;
     PyGpuArrayObject * weights = %(weights)s;
     PyGpuArrayObject * top = %(top)s;
     PyGpuArrayObject * out2 = NULL;
+
+    int wdim, odim;
+    wdim = unshared ? 6 : 4;
+    odim = 4; //Can be set to 6 later for unshared backprop wrt weights
 
     // Obtain or infer kernel width and height
     // (we need to know it early to be able to handle auto-padding)
@@ -735,8 +745,10 @@ class BaseGpuCorrMM(CGpuKernelBase):
 
     // Infer output shape and type
     // The inferred shape can be negative.
-    long long out_dim[4];
-    size_t out_dim_size[4];
+    long long out_dim[6];
+    size_t out_dim_size[6];
+    out_dim[4] = out_dim[5] = 0; //Only used for unshared backprop wrt weights
+    out_dim_size[4] = out_dim_size[5] = 0; //Same
     int out_typecode;
     PyGpuContextObject *out_context;
     switch(direction) {
@@ -745,71 +757,130 @@ class BaseGpuCorrMM(CGpuKernelBase):
         // height and width: top = (bottom + 2*pad - ((weight-1)*dil + 1)) / sample + 1
         out_dim[0] = PyGpuArray_DIMS(bottom)[0];
         out_dim[1] = PyGpuArray_DIMS(weights)[0];
-        out_dim[2] = (PyGpuArray_DIMS(bottom)[2] + 2*padH - ((PyGpuArray_DIMS(weights)[2]-1)*dilH + 1)) / dH + 1;
-        out_dim[3] = (PyGpuArray_DIMS(bottom)[3] + 2*padW - ((PyGpuArray_DIMS(weights)[3]-1)*dilW + 1)) / dW + 1;
+        out_dim[2] = (PyGpuArray_DIMS(bottom)[2] + 2*padH - ((PyGpuArray_DIMS(weights)[wdim-2]-1)*dilH + 1)) / dH + 1;
+        out_dim[3] = (PyGpuArray_DIMS(bottom)[3] + 2*padW - ((PyGpuArray_DIMS(weights)[wdim-1]-1)*dilW + 1)) / dW + 1;
         out_typecode = bottom->ga.typecode;
         out_context = bottom->context;
         if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
         {
-            PyErr_Format(PyExc_ValueError,
-                         "GpuCorrMM: impossible output shape\\n"
-                         "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
-                         "  weights shape: %%ld x %%ld x %%ld x %%ld\\n"
-                         "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
-                         PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
-                         PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
-                         PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
-                         PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
-                         out_dim[0], out_dim[1], out_dim[2], out_dim[3]);
-            %(fail)s
+            if (unshared) {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld %%ld %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
+                             PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
+                             PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
+                             PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
+                             PyGpuArray_DIMS(weights)[4], PyGpuArray_DIMS(weights)[5],
+                             out_dim[0], out_dim[1], out_dim[2], out_dim[3]);
+                %(fail)s
+            }
+            else {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
+                             PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
+                             PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
+                             PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
+                             out_dim[0], out_dim[1], out_dim[2], out_dim[3]);
+                %(fail)s
+            }
         }
         break;
     case 1:  // backprop wrt. weights
         // output is weights: (num_filters, num_channels, height, width)
         // height and width: weights = (bottom + 2*pad - (top - 1) * sample - 1) / dil + 1
         out_dim[0] = PyGpuArray_DIMS(top)[1];
-        out_dim[1] = PyGpuArray_DIMS(bottom)[1];
-        out_dim[2] = kH;  // already inferred further above
-        out_dim[3] = kW;  // how convenient
+        if (unshared){
+            odim = 6;
+            out_dim[1] = (npy_intp)PyGpuArray_DIMS(top)[2];
+            out_dim[2] = (npy_intp)PyGpuArray_DIMS(top)[3];
+        }
+        out_dim[wdim-3] = PyGpuArray_DIMS(bottom)[1];
+        out_dim[wdim-2] = kH;  // already inferred further above
+        out_dim[wdim-1] = kW;  // how convenient
         out_typecode = top->ga.typecode;
         out_context = top->context;
-        if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
-        {
-            PyErr_Format(PyExc_ValueError,
-                         "GpuCorrMM backprop wrt. weights: impossible output shape\\n"
-                         "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
-                         "  weights shape: %%ld x %%ld x %%ld x %%ld\\n"
-                         "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
-                         PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
-                         PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
-                         out_dim[0], out_dim[1], out_dim[2], out_dim[3],
-                         PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
-                         PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
-            %(fail)s
+        if (unshared) {
+            if (out_dim[0] < 0 || out_dim[1] <= 0 || out_dim[2] <= 0 || out_dim[3] < 0
+                    || out_dim[4] <= 0 || out_dim[5] <= 0){
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM backprop wrt. weights: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld %%ld %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
+                             PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
+                             out_dim[0], out_dim[1], out_dim[2], out_dim[3],
+                             out_dim[4], out_dim[5],
+                             PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
+                             PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
+                %(fail)s
+            }
+        }
+        else {
+             if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
+            {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM backprop wrt. weights: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             PyGpuArray_DIMS(bottom)[0], PyGpuArray_DIMS(bottom)[1],
+                             PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
+                             out_dim[0], out_dim[1], out_dim[2], out_dim[3],
+                             PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
+                             PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
+                %(fail)s
+            }
         }
         break;
     case 2:  // backprop wrt. inputs
         // output is bottom: (batchsize, num_channels, height, width)
         // height and width: bottom = (top - 1) * sample + (weights-1)*dil + 1 - 2*pad
         out_dim[0] = PyGpuArray_DIMS(top)[0];
-        out_dim[1] = PyGpuArray_DIMS(weights)[1];
-        out_dim[2] = (%(height)s != -1) ? %(height)s : (PyGpuArray_DIMS(top)[2] - 1) * dH + (PyGpuArray_DIMS(weights)[2]-1)*dilH + 1 - 2*padH;
-        out_dim[3] = (%(width)s != -1) ? %(width)s : (PyGpuArray_DIMS(top)[3] - 1) * dW + (PyGpuArray_DIMS(weights)[3]-1)*dilW + 1 - 2*padW;
+        out_dim[1] = PyGpuArray_DIMS(weights)[wdim-3];
+        out_dim[2] = (%(height)s != -1) ? %(height)s : (PyGpuArray_DIMS(top)[2] - 1) * dH + (PyGpuArray_DIMS(weights)[wdim-2]-1)*dilH + 1 - 2*padH;
+        out_dim[3] = (%(width)s != -1) ? %(width)s : (PyGpuArray_DIMS(top)[3] - 1) * dW + (PyGpuArray_DIMS(weights)[wdim-1]-1)*dilW + 1 - 2*padW;
         out_typecode = top->ga.typecode;
         out_context = top->context;
-        if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
-        {
-            PyErr_Format(PyExc_ValueError,
-                         "GpuCorrMM backprop wrt. inputs: impossible output shape\\n"
-                         "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
-                         "  weight shape: %%ld x %%ld x %%ld x %%ld\\n"
-                         "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
-                         out_dim[0], out_dim[1], out_dim[2], out_dim[3],
-                         PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
-                         PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
-                         PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
-                         PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
-            %(fail)s
+        if (unshared) {
+            if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
+            {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM backprop wrt. inputs: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weight shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             out_dim[0], out_dim[1], out_dim[2], out_dim[3],
+                             PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
+                             PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
+                             PyGpuArray_DIMS(weights)[4], PyGpuArray_DIMS(weights)[5],
+                             PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
+                             PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
+                %(fail)s
+            }
+        }
+        else {
+            if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
+            {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM backprop wrt. inputs: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weight shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             out_dim[0], out_dim[1], out_dim[2], out_dim[3],
+                             PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
+                             PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
+                             PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
+                             PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
+                %(fail)s
+            }
         }
         break;
     default:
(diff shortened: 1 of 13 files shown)


theano/tensor/nnet/abstract_conv.py
```diff
@@ -1507,6 +1526,13 @@ def conv(self, img, kern, mode="valid", dilation=1):
         out_shape = get_conv_output_shape(img.shape, kern.shape,
                                           mode, [1] * self.convdim, dilation)
 
+        out_rows = out_shape[self.convdim]
+        out_cols = out_shape[self.convdim+1]
```
gvtulder: This only works for 2D, so make sure you check that somewhere if you don't want to implement it now (the rest of the function does 3D as well).

I also don't think `self.convdim` is the best thing to use here, it doesn't really make sense. It's just a coincidence that batch and output channels use exactly `self.convdim` dimensions, so I'd just use 2.

Even better: you could make this whole first part work in a convdim-agnostic way, by checking if `kern.shape[2] == np.prod(out_shape[2:])`.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1507,6 +1526,13 @@ def conv(self, img, kern, mode="valid", dilation=1):
...
+
+        if unshared == True:
```
gvtulder: Do you have a specific reason for explicitly checking for `== True` here? (See also `unshared == False` below.)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1526,9 +1552,19 @@ def conv(self, img, kern, mode="valid", dilation=1):
...
-                                                          dilated_kern[n, im0, ...],
-                                                          1, val, bval, 0)
+                            if unshared == False:
+                                out[b, n, ...] += _convolve2d(img[b, im0, ...],
+                                                              dilated_kern[n, im0, ...],
+                                                              1, val, bval, 0)
+                            else:
+                                for reg in xrange(kern.shape[1]):
+                                    row = reg // out_cols
+                                    col = reg - row*out_cols
```
gvtulder: Perhaps `col = reg % out_cols` would be nicer?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1574,8 +1612,15 @@ def make_node(self, img, kern):
...
         if img.type.ndim != 2 + self.convdim:
             raise TypeError('img must be %dD tensor' % (2 + self.convdim))
-        if kern.type.ndim != 2 + self.convdim:
-            raise TypeError('kern must be %dD tensor' % (2 + self.convdim))
+
+        if unshared == True: 
+            if self.convdim == 3:
+                raise NotImplementedError("Unshared 3D convolution not implemented")
+            elif kern.type.ndim != 3 + self.convdim:
+                raise TypeError('kern must be %dD tensor' % (3 + self.convdim))
```
gvtulder: I would swap this: explicitly check that `self.convdim == 2` and raise and error if it is not. The AbstractConv class works for any convdim, even though it's only currently used for 2D and 3D.

theano/tensor/nnet/__init__.py
```diff
@@ -103,6 +103,11 @@ def conv2d(input, filters, input_shape=None, filter_shape=None,
         Factor by which to subsample (stride) the input.
         Also called dilation elsewhere.
 
+    unshared: bool
+        If true, then unshared or 'locally connected' convolution will be
+        performed. A different kernel will be used for each region of the
+        input.
+
```
gvtulder: I think it would be useful to change the documentation for `filters` as well, to explain how you should pass the unshared weights.

gvtulder:
What will the shape of the filter parameter look like for unshared convolution? From your Python implementation, it seems that you combine the `output_rows * output_cols` filters in one extra dimension, i.e., `(output channels, input channels, output rows * output cols, filter rows, filter columns)`. Is this the shape we want?

For comparison: Lasagne uses two additional dimensions, i.e., `(output channels, input channels, output rows, output cols, filter rows, filter columns)`. I think I would find that easier to work with.

vikramnitin9:
The method I've used to compute the gradients wrt inputs seems very messy. In order to express it as a convolution, the weight matrix had to be rearranged quite a bit. Is there a better way?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1565,6 +1565,7 @@ def conv(self, img, kern, mode="valid", dilation=1, unshared=False):
                             if unshared is True:
                                 for row in xrange(kern.shape[2]):
                                     for col in xrange(kern.shape[3]):
+                                        val = _valfrommode('valid')
                                         out[b, n, row, col] += _convolve2d(img[b, im0, row:row + kern.shape[4],
                                                                            col:col + kern.shape[5]],
                                                                            dilated_kern[n, im0, row, col, ...],
```
gvtulder: Is convolve2d necessary here? Since the image part you're extracting has the same size as the kernel, isn't this equal to an elementwise multiplication and summation? (Perhaps with a flip somewhere.)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2212,18 +2213,23 @@ def perform(self, node, inp, out_):
...
+            new_kern = np.zeros((new_kshp), dtype=topgrad.dtype)
+            for i in range(0, new_kshp[2]):
+                for j in range(0, new_kshp[3]):
+                    for m in range(0, new_kshp[4]):
+                        for n in range(0, new_kshp[5]):
+                            if i + m >= kern.shape[2] or j + n >= kern.shape[3]:
+                                new_kern[:, :, i, j, m, n] = 0
                             else:
-                                kern[:, :, i, j, m, m] = kern[:, :, i + m, j + n, m, n]
+                                new_kern[:, :, i, j, m, n] = kern[:, :, i + m, j + n, m, n]
```
gvtulder: Does this work? It's somewhat complex, so I might have made a mistake in my manual derivations, but I think this is missing the first part of the kernel and shifting everything else too far to the left.

In a 1D case, with a 5-pixel input and 3-pixel kernel, I think you want the kernel to be changed like this:
```
kern = [[a, b, c], [d, e, f], [g, h, i]]
new_kern = [[_, _, a], [_, b, d], [c, e, g], [f, h, _], [i, _, _]]
```
but this implementation would change it to this, I think:
```
new_kern = [[a, b, c], [d, e, _], [g, _, _], [_, _, _], [_, _, _]]
```

gvtulder:
> The method I've used to compute the gradients wrt inputs seems very messy. In order to express it as a convolution, the weight matrix had to be rearranged quite a bit. Is there a better way?

Maybe it's better to implement the gradient computation directly in Python, without using `convolve2d`?

I guess the reason to use `convolve2d` is that it can speed things up in the shared case, because it's really doing a series of convolutions. In the unshared case you have to call it separately for every pixel, so `convolve2d` might not be much faster than an elementwise multiplication. For the gradient wrt inputs you have to do the same, but you would need a tricky reorganisation of the kernel matrix first. At that point, `convolve2d` may be more trouble than it's worth.

I would suggest making a new helper function for unshared grad wrt inputs in `BaseAbstractConv`, reusing a little bit of code from `conv`, and using that in the gradInputs `perform`. I think that would be simpler than trying to reorganise the kernel so you can do a forward convolution.

What do you think?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1505,7 +1535,13 @@ def conv(self, img, kern, mode="valid", dilation=1):
...
+                                          mode, [1] * self.convdim, dilation, unshared)
+
+        if unshared is True:
+            if (kern.shape[2], kern.shape[3]) != (out_shape[2], out_shape[3]):
+                raise ValueError('Kernel shape ({},{}) does not match '
+                                 'output size ({},{})'.format(kern.shape[2], kern.shape[3],
+                                                              out_shape[2], out_shape[3]))
 
         out = np.zeros(out_shape, dtype=img.dtype)
         dil_kern_shp = kern.shape[:-self.convdim] + tuple(
```
gvtulder: A few lines below, I think you might need to change this assignment of `dilated_kern`:
```python
dilated_kern[(slice(None), slice(None)) +
             tuple(slice(None, None, dilation[i]) for i in range(self.convdim))
             ] = kern
```
because it doesn't include the row/column dimensions in an unshared convolution.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1022,7 +1044,8 @@ def conv2d_grad_wrt_weights(input,
                                                border_mode=border_mode,
                                                subsample=subsample,
                                                filter_flip=filter_flip,
-                                               filter_dilation=filter_dilation)
+                                               filter_dilation=filter_dilation,
+                                               unshared=False)
```
gvtulder: I think you want `unshared=unshared` here.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1699,3 +1699,49 @@ def test_conv2d_grad_wrt_weights(self):
...
+        inputs = theano.tensor.tensor4()
+        topgrad = theano.tensor.tensor4()
+
+        inputs_val = np.random.random(self.imshp).astype(theano.config.floatX)
+        topgrad_val = np.random.random(self.topgrad_shape).astype(theano.config.floatX)
+
+        grad_weights = conv.conv2d_grad_wrt_weights(inputs, topgrad,
+                                                    self.kshp, unshared=True)
+        grad_func = theano.function([inputs, topgrad], grad_weights, mode=self.mode)
+        grad_val = grad_func(inputs_val, topgrad_val)
```
gvtulder: Can I suggest a small extension of this test that also checks the result?
```python
    def test_gradweight(self):                                                                                                                                               
        inputs = theano.tensor.tensor4()                                                                                                                                       
        topgrad = theano.tensor.tensor4()                                                                                                                                  
                                                                                                                                                                                         
        inputs_val = np.random.random(self.imshp).astype(theano.config.floatX)                                                                       
        topgrad_val = np.random.random(self.topgrad_shape).astype(theano.config.floatX)                                                             
                                                                                                                                                           
        # compute gradient for unshared convolution
        grad_weights_unshared = conv.conv2d_grad_wrt_weights(inputs, topgrad,                                                                            
                                                             self.kshp, unshared=True)                                                                     
        grad_func_unshared = theano.function([inputs, topgrad], grad_weights_unshared, mode=self.mode)                                                         
        grad_val_unshared = grad_func_unshared(inputs_val, topgrad_val)                                                                                                  
                                                                                                                                                                                   
        # compute gradient for shared convolution as a reference
        grad_weights_shared = conv.conv2d_grad_wrt_weights(inputs, topgrad,                                                                                                        
                                                           self.kshp, unshared=False)                                                                      
        grad_func_shared = theano.function([inputs, topgrad], grad_weights_shared, mode=self.mode)                                                                                 
        grad_val_shared = grad_func_shared(inputs_val, topgrad_val)                                                                          
                                                                                                                                                                              
        # summing over all unshared gradients should give the shared gradient
        utt.assert_allclose(grad_val_shared, np.sum(grad_val_unshared, axis=(2, 3)))                                                                               
```

vikramnitin9: I'd actually added this locally and forgotten to commit it. Fixed now, thanks.

theano/tensor/nnet/abstract_conv.py
vikramnitin9: > I'd actually added this locally and forgotten to commit it. Fixed now, thanks.
Ditto here :)

theano/tensor/nnet/abstract_conv.py
vikramnitin9: I'd retained the convolve2d because it would eliminate two more loops, and because then I would need to flip the filter manually. But I've made it element-wise now.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2094,15 +2223,44 @@ def perform(self, node, inp, out_):
...
+        else:
+            flip_filters = ((slice(None), slice(None)) +
+                            (slice(None, None, -1),) * self.convdim)
+            if self.filter_flip:
+                topgrad = topgrad[flip_filters]
+            axes_order = (1, 0) + tuple(range(2, self.convdim + 2))
+            kern = kern.transpose(axes_order)
+            img = self.conv(topgrad, kern, mode="full", dilation=self.filter_dilation)
+            if self.filter_flip:
+                img = img[flip_filters]
```
vikramnitin9: I have a doubt here. Why do we flip the topgrad, convolve and then again flip the output? Wouldn't it be easier to just flip (or not flip) the kern instead?

vikramnitin9:
I have a doubt [here](https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/abstract_conv.py#L2101-L2105). Why do we flip the topgrad, convolve and then again flip the output? Isn't this equivalent to flipping (or not flipping) the kern and convolving?

vikramnitin9:
> I would suggest making a new helper function for unshared grad wrt inputs in BaseAbstractConv, reusing a little bit of code from conv, and using that in the gradInputs perform. I think that would be simpler than trying to reorganise the kernel so you can do a forward convolution.

I've just packed it all into the gradInputs perform function for now. Do you think it looks fine or should I move it into a separate function?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2247,25 +2242,24 @@ def perform(self, node, inp, out_):
...
-                                        reg_col = (img.shape[3] - 1) - col + \
-                                                  (kern.shape[5] - 1) - k_col
-                                        reg_row = min(reg_row, kern.shape[2] - 1)
-                                        reg_row = max(reg_row, 0)
-                                        reg_col = min(reg_col, kern.shape[3] - 1)
-                                        reg_col = max(reg_col, 0)
+                                        reg_row = row - ((kern.shape[4] - 1) - k_row)
+                                        reg_col = col - ((kern.shape[45] - 1) - k_col)
+                                        reg_row = max(0, min(reg_row, kern.shape[2] - 1))
+                                        reg_col = max(0, min(reg_col, kern.shape[3] - 1))
```
gvtulder: I'm surprised to see max/min here, are you sure this is correct? If a value falls outside the available range, wouldn't you want to ignore it, rather than add it to the nearest row/col?

Isn't this the same as the loops and indices for the [forward convolution](https://github.com/vikramnitin9/Theano/blob/d803e2784a457f71a6eb31c88271c4ca15631d32/theano/tensor/nnet/abstract_conv.py#L1568-L1579), except that the source and target variables are swapped?

gvtulder:
> I have a doubt [here](https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/abstract_conv.py#L2101-L2105). Why do we flip the topgrad, convolve and then again flip the output? Isn't this equivalent to flipping (or not flipping) the kern and convolving?

Possibly, yes. Maybe there is a combination of dilation, subsampling etc. when it doesn't? (I tried to come up with an example that wouldn't work, but I failed.)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1928,35 +1941,43 @@ def perform(self, node, inp, out_):
...
+            kern = kern.transpose(axes_order)
 
-        if any(self.filter_dilation[i] > 1 for i in range(self.convdim)):
-            kern = kern[(slice(None), slice(None)) +
-                        tuple(slice(None, None, self.filter_dilation[i])
-                              for i in range(self.convdim))]
+            if any(self.filter_dilation[i] > 1 for i in range(self.convdim)):
+                kern = kern[(slice(None), slice(None)) +
+                            tuple(slice(None, None, self.filter_dilation[i])
+                                  for i in range(self.convdim))]
```
vikramnitin9: But the kernel has a different number of dimensions, so don't we have to account for that?

vikramnitin9: Sorry, never mind. Just saw your earlier comment

theano/tensor/nnet/abstract_conv.py
vikramnitin9: > I'm surprised to see max/min here, are you sure this is correct? If a value falls outside the available range, wouldn't you want to ignore it, rather than add it to the nearest row/col?

I did this since the topgrad for these entries will anyway be zero so nothing will be added. But I've made a new version now.

> Isn't this the same as the loops and indices for the forward convolution, except that the source and target variables are swapped?

The difference is that the `row` and `col` of the kern vary within the inner two loops also.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1690,9 +1690,9 @@ def perform(self, node, inp, out_):
             img = new_img
         if not self.filter_flip:
             if self.unshared is True:
```
vikramnitin9: I've now added a check in the `init` function of `BaseAbstractConv`. Will that do?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1565,11 +1572,11 @@ def conv(self, img, kern, mode="valid", dilation=1, unshared=False):
...
-                                        val = _valfrommode('valid')
-                                        out[b, n, row, col] += _convolve2d(img[b, im0, row:row + kern.shape[4],
-                                                                           col:col + kern.shape[5]],
-                                                                           dilated_kern[n, im0, row, col, ...],
-                                                                           1, val, bval, 0)
+                                        for k_row in xrange(kern.shape[4]):
+                                            for k_col in xrange(kern.shape[5]):
+                                                out[b, n, row, col] += img[b, im0, row + k_row,
+                                                                           col + k_col] * \
+                                                    dilated_kern[n, im0, row, col, k_row, k_col]
```
vikramnitin9: I was actually thinking of doing an `np.sum(np.multiply(..., ...))` to eliminate the two inner loops and not unnecessarily call `_convolve2d`. Do you think I should?

vikramnitin9:
Regarding moving code into helper functions, I'd probably have to have different functions for each of the three operations (forward, gradWeights, gradInput), since the computation performed in the loop is different in each of these cases (unlike for a normal computation, where the same conv function can be used).

Should I go ahead and create separate functions just the same?

theano/tensor/nnet/abstract_conv.py
gvtulder: It's certainly good to have it there, and hopefully it's somewhat temporary anyway. However, I do think I'd like to have an explicit check in `perform` as well. It's good to have the check and warning close to where the limitation actually is, in my opinion. It's useful as documentation and also makes it future-proof: if at some point the check is removed from BaseAbstractConv, it's better for `perform` to crash than to silently run a shared convolution.

Compare this with the 2D/3D limitation of the convolution: that's also explicitly checked in [`BaseAbstractConv.conv`](https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/abstract_conv.py#L1540).

theano/tensor/nnet/abstract_conv.py
gvtulder: Yes, that could work.

gvtulder:
> Regarding moving code into helper functions, I'd probably have to have different functions for each of the three operations (forward, gradWeights, gradInput), since the computation performed in the loop is different in each of these cases (unlike for a normal computation, where the same conv function can be used).
>
> Should I go ahead and create separate functions just the same?

Yes, I still think that would be useful. But maybe that's a matter of taste. Ideally, I'd like to see a `perform` method that includes only a few unshared-specific lines.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1633,10 +1633,7 @@ def make_node(self, img, kern):
             raise TypeError('img must be %dD tensor' % (2 + self.convdim))
 
         if self.unshared is True:
-            if self.convdim != 2:
-                raise NotImplementedError('Unshared convolution not implemented for %dD'
-                                          % self.convdim)
-            elif kern.type.ndim != 6:
+            if kern.type.ndim != 6:
```
gvtulder: `!= 2 + 2 * self.convdim`

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2247,11 +2228,12 @@ def perform(self, node, inp, out_):
                                         # Deciding which kernel to select
                                         reg_row = row - ((kern.shape[4] - 1) - k_row)
                                         reg_col = col - ((kern.shape[5] - 1) - k_col)
-                                        reg_row = max(0, min(reg_row, kern.shape[2] - 1))
-                                        reg_col = max(0, min(reg_col, kern.shape[3] - 1))
+                                        if reg_row not in range(kern.shape[2]) or \
+                                                reg_col not in range(kern.shape[3]):
+                                            continue
```
gvtulder: I don't think this is very efficient: it's creating these `range` objects for every pixel. Normal comparisons are probably faster. (And I'd invert the comparison, so you don't need the `continue`.)

gvtulder:
> Regarding moving code into helper functions, I'd probably have to have different functions for each of the three operations (forward, gradWeights, gradInput), since the computation performed in the loop is different in each of these cases (unlike for a normal computation, where the same conv function can be used).

Actually, I think you can do with only one helper function for unshared convolution, if you use a `direction` argument that indicates if it's a forward, gradinputs or gradweights convolution (similar to the parameter in the corrMM implementation). The loops should be exactly the same for all three cases, I think. It's just the computation inside the loop that read and write from different variables.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -71,10 +70,8 @@ def get_conv_output_shape(image_shape, kernel_shape,
     """
     bsize, imshp = image_shape[0], image_shape[2:]
     nkern = kernel_shape[0]
-    if unshared is True:
-        kshp = kernel_shape[4:]
-    else:
-        kshp = kernel_shape[2:]
+
+    kshp = kernel_shape[-2:]
```
gvtulder: It's nice to get rid of the `unshared` argument if you don't need it, but I'm not sure this is correct. I think it assumes that you do a 2D convolution?

Perhaps you can use the length of `image_shape` to compute the value of `convdim`, then use that instead of 2?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1520,7 +1517,7 @@ def flops(self, inp, outp):
             raise NotImplementedError(
                 'flops not implemented for convdim={}', self.convdim)
 
-    def conv(self, img, kern, mode="valid", dilation=1, unshared=False):
+    def conv(self, img, kern, mode="valid", dilation=1, unshared=False, direction=0):
```
gvtulder: Could you change the `direction` to a string? (E.g., "forward", "backprop weights", "backprop inputs" as in BaseCorrMM.)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1592,6 +1586,47 @@ def conv(self, img, kern, mode="valid", dilation=1, unshared=False):
...
+                else:  # direction = 2
+                    for k_row in xrange(kern.shape[2]):
+                        for k_col in xrange(kern.shape[3]):
+                            # Deciding which kernel to select
+                            reg_row = row - ((kern.shape[2] - 1) - k_row)
+                            reg_col = col - ((kern.shape[3] - 1) - k_col)
+                            if (reg_row >= 0 and reg_row < kern.shape[0]) and \
+                                    (reg_col >= 0 and reg_col < kern.shape[1]):
+                                out[row, col] += kern[reg_row, reg_col, k_row, k_col] * \
+                                    inp[reg_row, reg_col]
```
gvtulder: I think this is more complicated than it needs to be. If you loop over the correct range, I don't think you need these `if`s checking whether you're inside the range.

Reversing your [forward convolution](https://github.com/vikramnitin9/Theano/blob/d803e2784a457f71a6eb31c88271c4ca15631d32/theano/tensor/nnet/abstract_conv.py#L1568-L1579), these loops give the same result for the grad wrt inputs:
```python
for row in xrange(kern.shape[0]):
    for col in xrange(kern.shape[1]):
        for k_row in xrange(kern.shape[2]):
            for k_col in xrange(kern.shape[3]):
                out[row + k_row, col + k_col] += inp[row, col] * \
                    kern[row, col, -(k_col + 1), -(k_row + 1)]
```

vikramnitin9: This certainly looks a lot simpler!

But for the common `unshared2d` function, the outer two loops are common for all three directions (they go from `0` to `out_shape[0]` and `0` to `out_shape[1]`). We need them to go from `0` to `kern.shape[0]` and `0` to `kern.shape[1]`.

I think the easiest way to fix this is to have a condition to check whether `row` and `col` are less than `kern.shape[0]` and `kern.shape[1]` respectively, rather than modifying the loops for `k_row` and `k_col`.

The conditions will look much simpler and cleaner also.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1589,45 +1587,37 @@ def conv(self, img, kern, mode="valid", dilation=1, unshared=False, direction=0)
...
-                                out[row, col] += kern[reg_row, reg_col, k_row, k_col] * \
-                                    inp[reg_row, reg_col]
+                elif direction == "backprop inputs":
+                    if row < kern.shape[0] and col < kern.shape[1]:
+                        for k_row in xrange(kern.shape[2]):
+                            for k_col in xrange(kern.shape[3]):
+                                out[row + k_row, col + k_col] += inp[row, col] * \
+                                    kern[row, col, -(k_row + 1), -(k_col + 1)]
+                else:
+                    raise ValueError("unshared2d: invalid value for 'direction'")
```
gvtulder: It might be helpful to include the value in the error message.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1540,6 +1587,39 @@ def conv(self, img, kern, mode="valid", dilation=1):
...
+            for col in xrange(out_shape[1]):
+                if direction == "forward":
+                    out[row, col] += np.sum(np.multiply(inp[row:row + kern.shape[2],
+                                                        col:col + kern.shape[3]],
+                                            kern[row, col, ::-1, ::-1]))
+                elif direction == "backprop weights":
+                    out[row, col, ...] += kern[row, col] * \
+                        inp[row:row + out_shape[2], col:col + out_shape[3]]
+                elif direction == "backprop inputs":
+                    if row < kern.shape[0] and col < kern.shape[1]:
```
gvtulder: I appreciate your effort not to duplicate code, but I still think this solution is less than ideal. It feels a bit silly to have two loops that go on for too long, and then correct this with an if that checks the indices.

There may also be a conceptual problem: your for loops suggest that you're iterating over the rows and columns of `out`, while in fact for the grad wrt inputs you're iterating over the rows and columns of `kern` (which just happens to be slightly smaller than `out`).

So, I would propose to move the `for row` and `for col` loops inside the ifs, so that they're specific to each direction. This adds three lines of code, but is easier to understand and does not have loops that are longer than necessary. (As an added bonus it tests the value of `direction` only once.)

vikramnitin9:
Right now, the weights are of the form `[out_filters, in_filters, out_rows, out_cols, kH, kW]`. I think it actually make more logical sense to have them as `[out_filters, out_rows, out_cols, in_filters, kH, kW]`.

Also, it makes things easier in the C code.

vikramnitin9:
@gvtulder Do you think I should go ahead and change this?

gvtulder:
@vikramnitin9 I'm not yet convinced that it's a good idea.

Of the current order, I like that the first two dimensions always have the same meaning for shared and unshared. There's now a clear relation `img.shape[1] == kern.shape[1]` and `kern.shape[0] == out.shape[1]`, which you can depend on without knowing if it's shared or unshared convolution. So while it may make some things simpler, I think it would add complexity in a lot of other places.

Lasagne uses [`[out_filters, in_filters, out_rows, out_cols, kH, kW]`](https://lasagne.readthedocs.io/en/latest/modules/layers/local.html), which may be another reason to use that approach. (Keras puts `in_filters` at the end.)

Would it make the C or GPU code a lot more efficient if the `out_cols, out_rows` came before the `in_filters`? (There was some [open-ended discussion](https://github.com/Theano/Theano/issues/5620) in the issue you linked to before.)

vikramnitin9:
im2col reshapes the image as `[regions, in_filters*kH*kW]`. Normally the weight matrix would be implicitly reshaped to `[out_filters, in_filters*kH*kW]`.

For unshared, it could be reshaped to `[out_filters, regions, in_filters*kH*kW]` and `out_filters * regions` number of vector-matrix products will be performed. This will be easy if `in_filters`, `kH` and `kW` are together in the weights array so we don't actually need to reshape.

So that's why I said the C code will become easier.

Lasagne uses `[out_filters, in_filters, kH, kW, out_rows, out_cols]`. That can be implicitly reshaped to `[out_filters, in_filters*kH*kW, regions]`. Only thing is, I find it hard to imagine the weight matrix logically organised like this. 

Either way, the three dimensions being together makes the reshaping easy.

gvtulder:
@vikramnitin9 You're right, I misread the order of the Lasagne weights.

If it's more efficient, as you say, to have `[out_filters, out_rows, out_cols, in_filters, kH, kW]`, then I think it's fine to use that order. Would it be a good idea to finish the C implementation first, using that new shape, before you go back and change the Python code you already wrote? You can do a dimshuffle for now to convert the input to the new shape.

vikramnitin9:
Sure. I'm currently working on the C code under the assumption that weights are in the required format. Still need to do grad wrt inputs and add tests.

vikramnitin9:
I have a question [here](https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/corr_gemm.c#L263-L266) : why do we use `tid` in one case and `n` in the other? Doesn't that mean that the result of one sample (inp[tid]) may go to a different location in `top` (`top[n]`)?

gvtulder:
I think you have to read that together with the previous call to `im2col`. That `im2col` copies input image `n` to the `tid`th row of the temporary `col` array. Next, the call to `gemm` computes the results for this image (the `n`th image is the `tid`th element in the temporary array) and writes the result to the `n`th row of `top`. So in the end, the output at `n` in `top` is computed from the input at `n`, using the temporary storage at `tid` in `col`.

gvtulder:
For your task list:
* The unshared convolution should eventually be included in the AbstractConv tests. Those tests will automatically run all CPU/GPU implementations, so it's better to do most of the testing there. (The tests you currently have are useful while you're developing, though.) 
* For testing gradients (I'm not sure what you mean by "Setting all regions' weights equal isn't a good way to verify grad inputs. Idea : subsampling = filter size?"), have a look at the `verify_grad` function used elsewhere.

vikramnitin9:
Everything seems fine, except that the order of regions is inverted for gradWeights ([here](https://github.com/Theano/Theano/pull/6007/files#diff-86424aa0121840d0b00c62c2e2b3c6d7R471)). I'm not yet sure why.

gvtulder:
Good. No, I think Travis is explicitly reusing the compilation cache to make the tests run faster. I don't know if that's the previous version the master branch or if it's the cache of the previous run on this PR.

But in any case: if you change the C code, make sure you also change the `c_code_cache_version` for that class. That will ensure that the cached previous version isn't used, which is also important for users who upgrade Theano.

vikramnitin9:
Does the C code look okay? Can I go ahead and move this to the GPU?

theano/tensor/nnet/corr.py
```diff
@@ -297,13 +303,17 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
             break;
     }
 
+    int wdim, odim;
+    wdim = unshared ? 6 : 4;
+    odim = 4; //Can be set to 6 later if required
```
gvtulder: This comment is bit vague.

theano/tensor/nnet/corr.py
```diff
@@ -357,15 +367,16 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
     }
 
     // Infer output shape
-    npy_intp out_dim[4];
+    npy_intp out_dim[6];
+    out_dim[4] = out_dim[5] = 0; //Will be changed if required
```
gvtulder: This is a bit vague. I'd prefer something like: "only used for unshared backprop wrt. weights".

theano/tensor/nnet/corr.py
```diff
@@ -580,8 +611,12 @@ def make_node(self, img, topgrad, shape=None):
         else:
             height_width = [as_tensor_variable(shape[0]).astype('int64'), as_tensor_variable(shape[1]).astype('int64')]
 
-        broadcastable = [topgrad.type.broadcastable[1], img.type.broadcastable[1],
-                         False, False]
+        if self.unshared is True:
+            broadcastable = [topgrad.type.broadcastable[0], img.type.broadcastable[1]] +\
```
gvtulder: I think you need a space before the `\`, but that may be a flake thing.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -119,7 +120,7 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
         return NULL;
     }
 
-    if (PyArray_NDIM(weight) != 4)
+    if (PyArray_NDIM(weight) != (unshared ? 6 : 4))
     {
         PyErr_SetString(PyExc_ValueError, "CorrMM requires weight of 4D");
```
gvtulder: Change the error message as well.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -173,6 +175,23 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
+                "  bottom shape: %%d %%d %%d %%d\n"
+                "  weight shape: %%d %%ld %%ld %%d %%d %%d"
+                " (expected %%d %%d %%d %%d %%d %%d)\n"
+                "  top shape(calculated): %%d %%d %%d %%d\n",
+                batchSize, nChannels, bottomHeight, bottomWidth,
+                nFilters, PyArray_DIMS(weight)[1],
+                PyArray_DIMS(weight)[2], nChannels, kH, kW,
+                nFilters, topHeight, topWidth, nChannels, kH, kW,
+                batchSize, nFilters, topHeight, topWidth);
+        return NULL;
```
gvtulder: Indentation.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -173,6 +175,23 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
...
+        }
+    }
     if (batchSize != PyArray_DIMS(top)[0] ||
             nFilters != PyArray_DIMS(top)[1] ||
             topHeight != PyArray_DIMS(top)[2] ||
```
gvtulder: I think this needs to be special-cased for unshared.

theano/tensor/nnet/corr_gemm.c
```diff
@@ -220,8 +239,10 @@ PyArrayObject* corrMM(PyArrayObject* bottom,
     const int N_ = col_dim[2];
     const int col_stride = (K_ * N_);
     const int M_ = nFilters;
+    const int one_int = 1;
     const %(c_float_type)s one = 1.0;
     const %(c_float_type)s zero = 0.0;
+    const int ldw = (K_ * N_);
```
gvtulder: Since this is equal to `col_stride`, do you need a separate variable?

gvtulder:
Do you want to implement it on the GPU first, or would it be better to extend the AbstractConv tests before you do that? So far you're only testing the simple case, but you can get a lot more rigorous testing almost for free by including the `unshared` parameter in the general AbstractConv tests. It might be worthwhile to do that first, but that's up to you.

theano/tensor/nnet/corr_gemm.c
vikramnitin9: I thought it wouldn't make sense to use `col_stride` to indicate the leading dimension of weights, so a separate variable would be needed for clarity.

vikramnitin9:
> It would be interesting to see if it also passes the tests.

The Travis errors disappeared after I changed the cache version.

I'll integrate the tests with the general AbstractConv tests before moving on to GPU.

vikramnitin9:
Update : I've integrated the tests but I've discovered problems with subsampling and dilation. Working on fixing them.

gvtulder:
Good. That means the tests are useful then!

vikramnitin9:
I've fixed a lot of errors, and the `TestConvNoOptim` tests work fine. But the `TestCorrConv2d` (with optimiser enabled) still gives me allclose errors, and I have to figure out why.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1723,12 +1730,30 @@ def perform(self, node, inp, out_):
...
             kern = kern[(slice(None),) * (kern.ndim - self.convdim) + (slice(None, None, -1),) * self.convdim]
 
-        # from (nFilters, out_rows, out_cols, nChannels, kH, kW)
-        # to (nFilters, nChannels, out_rows, out_cols, kH, kW)
         if self.unshared:
+            out_shape = get_conv_output_shape(img.shape, kern.shape,
+                                              mode, self.subsample, self.filter_dilation)
+            if kern.shape[1:1 + self.convdim] != out_shape[2:2 + self.convdim]:
+                raise ValueError('Kernel shape {} does not match '
+                                 'output size {}'.format(kern.shape[1:1 + self.convdim],
```
gvtulder: Maybe make that computed output size, so it's clear that it's not an user-defined input that has the wrong shape?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -564,13 +597,21 @@ def test_all(self):
...
-                for fd in self.filters_dilations:
-                    for s in self.subsamples:
-                        for b in self.border_modes:
-                            yield (self.tcase, i, f, s, b, dflip,
-                                   dprovide_shape, fd)
-                for flip in self.filter_flip:
-                    yield (self.tcase, i, f, ds, db, flip, dprovide_shape)
+                for u in self.unshared:
+                    if u and len(i) != 4:
+                        continue
```
gvtulder: Can you add a comment saying that this is because it's not yet implemented in 3D?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -564,13 +597,21 @@ def test_all(self):
...
+                    for fd in self.filters_dilations:
+                        for s in self.subsamples:
+                            for b in self.border_modes:
+                                if u:
+                                    out_shp = get_conv_output_shape(i, f, b, s, fd)
+                                    fmod = f[:1] + out_shp[2:] + f[1:]
```
gvtulder: A comment would help here, I think.

gvtulder:
The allclose errors, are they large? It might be useful to run them with float64 to see if it's because of rounding errors. If they work in float64 but not in float32, you'd have to increase the tolerance a little bit.

theano/tensor/nnet/opt.py
```diff
@@ -200,6 +200,7 @@ def local_abstractconv_gradinputs_gemm(node):
                              subsample=node.op.subsample,
                              filter_dilation=node.op.filter_dilation,
                              unshared=node.op.unshared)(kern, topgrad, shape)
+
```
gvtulder: I don't know if these newlines are things that you have introduced yourself in a previous commit, but in general it's probably better not to have these changes that don't really do anything.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -480,18 +480,21 @@ def test_gradweight(self):
         utt.assert_allclose(ref_val, unshared_val)
 
     def test_gradinput(self):
-        filters_val = np.random.random(self.kshp).astype(theano.config.floatX)
-        topgrad_val = np.random.random(self.topgrad_shape).astype(theano.config.floatX)
+        filters_val = np.random.random(self.kshp).astype('float32')
+        topgrad_val = np.random.random(self.topgrad_shape).astype('float32')
```
gvtulder: Why do you need this?

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -480,18 +480,21 @@ def test_gradweight(self):
...
 
         conv_unshared = corr.CorrMM_gradInputs(unshared=True, subsample=self.sub,
                                                filter_dilation=self.dil)(self.filters, self.topgrad, self.imshp[-2:])
-        unshared_func = theano.function([self.filters, self.topgrad], conv_unshared)
+        unshared_func = theano.function([self.filters, self.topgrad], conv_unshared, mode="FAST_RUN")
```
gvtulder: Isn't that what `self.mode` is for?

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -480,18 +480,21 @@ def test_gradweight(self):
...
-        conv_ref = theano.tensor.nnet.abstract_conv.conv2d_grad_wrt_inputs(self.topgrad, self.filters, self.imshp,
-                                                                           filter_flip=False, unshared=True,
-                                                                           subsample=self.sub, filter_dilation=self.dil)
-        ref_func = theano.function([self.topgrad, self.filters], conv_ref)
+        conv_ref = theano.tensor.nnet.abstract_conv.AbstractConv2d_gradInputs(self.imshp, self.kshp, border_mode="valid",
+                                                                              filter_flip=False, unshared=True,
+                                                                              subsample=self.sub,
+                                                                              filter_dilation=self.dil)(self.filters,
+                                                                                                        self.topgrad,
+                                                                                                        self.imshp[-2:])
```
gvtulder: Shouldn't it also work with `conv2d_grad_wrt_inputs`?

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -480,18 +480,21 @@ def test_gradweight(self):
...
+        ref_func = theano.function([self.topgrad, self.filters], conv_ref, mode=None)
```
gvtulder: Here as well: does `self.mode` work?

vikramnitin9:
Sorry for the confusion. All four of the changes you mentioned were caused by my trying to reproduce an error I was getting in the nose tests. I'll remove them.

I'll probably have to do something better for the `TestUnsharedCorr2d` function, right? Rather than have it depend on AbstractConv. What do you suggest?

vikramnitin9:
Another issue now is that because of the extra unshared loop, the abstract conv tests are taking double the time (Travis says maximum time limit for job exceeded).

gvtulder:
For the AbstractConv tests, perhaps something similar to the grouped convolution tests could be a solution. In that PR the grouped tests are kept separate from the main tests. But perhaps it's better to do the GPU first, so we can test that with the extensive tests first? The Jenkins CI doesn't have a time limit, I think. Then we can ask nouiz for his opinion on how to reduce the tests later.

For the CPU corrMM tests, perhaps you could do something similar to the main tests? But again: GPU first, I'd say.

vikramnitin9:
I've done the GPU code, but it gives
`RuntimeError: GpuCorrMM forward encountered an error running gemm: 11`

Is it just something to do with my installation of blas or is there an error in the code?

Also, the corr_gemm.c in gpuarray has been modified since I cloned the repo (the three calls to sgemm, dgemm and hgemm for each direction have been moved into one dgemm function). So there are conflicts. Should I do a rebase?

gvtulder:
Yes, a rebase would be good. I haven't looked at your new code in detail yet, but error 11 sounds mysterious. The [cublas document](http://docs.nvidia.com/cuda/cublas/index.html#cublasstatus_t) could find wasn't too helpful. If you do a rebase, we can see if it works on the Jenkins system.

gvtulder:
So if I understand the error message correctly, this mapping error means that cublas couldn't copy some data. Are you sure that all the sizes and indices are correct?

It would also be good to know in which direction the error occurs. If I interpreted your code correctly, all three directions will print "forward" in their error message.

A somewhat random idea: I am not 100% sure if you can just run these blas calls in a loop, or if that causes problems if they're using the same temporary variables. It probably won't help, but it would be good to see if it also crashes if you run only one iteration of your new loop.

lamblin:
If I remember correctly, error 11 can occur when strides or ld* variables are considered incorrect by cublas.
One such instance is for row or vector matrices, where ld* has to be at least as large as the number of elements, but numpy-style strides would use 1.
I have no idea if that is the case here, but printing the dimensions and strides could help.

vikramnitin9:
I used gemm instead of gemv and the error disappeared. No idea why. I'm using OpenCL with CLBlast.

I'll do the rebase now. I'm thinking of opening a new PR since this one is already very messy.

vikramnitin9:
Rebase done. Should be all fine now.

gvtulder:
Jenkins test this

gvtulder:
jenkins test this

gvtulder:
It looks like the magic phrase doesn't work for me (although Jenkins does run if I create a PR myself, but that might be a different switch). It was worth trying.

@nouiz or @lamblin, could you trigger the GPU tests please?

vikramnitin9:
The Travis tests passed except for the abstractconv, which timed out as usual.

For the Jenkins test, should I add unshared tests to `theano/gpuarray/test_abstractconv.py` or will the current set do?

gvtulder:
Hmm, yes, there need to be some GPU tests. I thought I had seen them, also because you had these errors, but apparently I was wrong.

I think there need to be tests in `test_gemmcorr.py` and in `test_abstractconv.py` in the `theano.gpuarray` directory. (In fact, I suspect the abstractconv tests might even fail right now, because you've introduced this new `u` parameter that the GPU tests don't know about.) The abstractconv tests should be fairly easy to expand, the other test can be based on the existing GPU and CPU tests.

vikramnitin9:
Sorry for the delay. Here are the GPU tests.

theano/gpuarray/tests/test_abstractconv.py
```diff
@@ -26,7 +26,7 @@ def setup_class(cls):
         # provide_shape is not used by the cuDNN impementation
         cls.provide_shape = [False]
 
-    def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
+    def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1), u=False):
```
gvtulder: It would be good to raise a SkipTest error if unshared is true. (Like for `fd`.) See also the other instances below.

theano/gpuarray/tests/test_abstractconv.py
```diff
@@ -196,7 +196,7 @@ def setup_class(cls):
         cls.shared = staticmethod(gpuarray_shared_constructor)
         cls.mode = mode_with_gpu.excluding('cudnn')
 
-    def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1, 1)):
+    def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1, 1), u=False):
```
gvtulder: SkipTest for those cases as well. (You shouldn't just ignore values you don't like, since that would suggest that the test is run and passes.)

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -425,8 +425,7 @@ def test_non_contiguous(self):
 
 class TestUnsharedCorr2D(utt.InferShapeTester):
     if theano.config.mode == "FAST_COMPILE":
-        # mode = theano.compile.get_mode("FAST_RUN").excluding("conv_dnn")
-        mode = theano.compile.get_default_mode().including('gpuarray').excluding('gpu').excluding('cudnn')
+        mode = theano.compile.get_mode("FAST_RUN").excluding("conv_dnn")
```
gvtulder: Doyou need to exclude conv_dnn here? Since the shared convolution tests don't do this.

theano/tensor/nnet/tests/test_corr.py
```diff
@@ -453,13 +452,13 @@ def test_fwd(self):
 
         conv_unshared = corr.CorrMM(unshared=True, subsample=self.sub,
                                     filter_dilation=self.dil)(self.input, self.filters)
-        unshared_func = theano.function([self.input, self.filters], conv_unshared)
+        unshared_func = theano.function([self.input, self.filters], conv_unshared, mode=self.mode)
```
gvtulder: This looks a bit weird, but it is because you're using the Python implementation as your reference, right? I'm thinking if it wouldn't be better to have a separate implementation for the tests, like the one that's used for the normal convolution.

gvtulder:
@nouiz or @lamblin, there at GPU tests now, so can you give this a go on Jenkins?

lamblin:
Jenkins test this please

lamblin:
Also, tests are timing out on Travis because some parts exceed 50 minutes.
I think we should find a way of testing fewer cases or smaller sizes, at least in the slower (python) modes.

theano/gpuarray/blas.py
```diff
@@ -745,71 +757,130 @@ def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width
...
-                         PyGpuArray_DIMS(bottom)[2], PyGpuArray_DIMS(bottom)[3],
-                         PyGpuArray_DIMS(weights)[0], PyGpuArray_DIMS(weights)[1],
-                         PyGpuArray_DIMS(weights)[2], PyGpuArray_DIMS(weights)[3],
-                         out_dim[0], out_dim[1], out_dim[2], out_dim[3]);
-            %(fail)s
+            if (unshared) {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld %%ld %%ld\\n"
```
gvtulder: There are some `x` missing near the end.

theano/gpuarray/blas.py
```diff
@@ -745,71 +757,130 @@ def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width
...
-                         PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
-                         PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
-            %(fail)s
+        if (unshared) {
+            if (out_dim[0] < 0 || out_dim[1] <= 0 || out_dim[2] <= 0 || out_dim[3] < 0
+                    || out_dim[4] <= 0 || out_dim[5] <= 0){
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM backprop wrt. weights: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld %%ld %%ld\\n"
```
gvtulder: There are some `x` missing near the end.

theano/gpuarray/blas.py
```diff
@@ -745,71 +757,130 @@ def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width
...
-                         PyGpuArray_DIMS(top)[0], PyGpuArray_DIMS(top)[1],
-                         PyGpuArray_DIMS(top)[2], PyGpuArray_DIMS(top)[3]);
-            %(fail)s
+        if (unshared) {
+            if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
+            {
+                PyErr_Format(PyExc_ValueError,
+                             "GpuCorrMM backprop wrt. inputs: impossible output shape\\n"
+                             "  bottom shape: %%ld x %%ld x %%ld x %%ld\\n"
+                             "  weight shape: %%ld x %%ld x %%ld x %%ld\\n"
```
gvtulder: This is missing the extra two dimensions.

theano/gpuarray/corr_gemm.c
```diff
@@ -367,21 +368,35 @@ PyGpuArrayObject* corrMM(PyGpuArrayObject *const bottom,
...
-                "but strides are: %ld %ld %ld %ld\n",
-                PyGpuArray_STRIDES(weight)[0],
-                PyGpuArray_STRIDES(weight)[1],
-                PyGpuArray_STRIDES(weight)[2],
-                PyGpuArray_STRIDES(weight)[3]);
-        return NULL;
+        if (unshared) {
+            PyErr_Format(PyExc_ValueError,
+                    "GpuCorrMM requires weight to be C-contiguous, "
+                    "but strides are: %ld %ld %ld %ld\n",
```
gvtulder: This is missing two dimensions.

theano/gpuarray/corr_gemm.c
```diff
@@ -429,21 +444,56 @@ PyGpuArrayObject* corrMM(PyGpuArrayObject *const bottom,
...
-        return NULL;
+    if (unshared) {
+        if (topHeight != PyGpuArray_DIMS(weight)[1] ||
+                topWidth != PyGpuArray_DIMS(weight)[2]) {
+            PyErr_Format(PyExc_ValueError,
+                    "GpuCorrMM regions in kernel must match output regions:\n"
+                    "  bottom shape: %d %d %d %d\n"
+                    "  weight shape: %d %ld %ld %d %d %d"
+                    " (expected %d %d %d %d %d %d)\n"
+                    "  top shape(calculated): %d %d %d %d\n",
```
gvtulder: Since the shared version uses `%ld` everywhere, it's probably good to do that here as well. (My compiler gave some warnings here.)

theano/gpuarray/corr_gemm.c
```diff
@@ -429,21 +444,56 @@ PyGpuArrayObject* corrMM(PyGpuArrayObject *const bottom,
...
+        }
+        if (batchSize != PyGpuArray_DIMS(top)[0] ||
+                nFilters != PyGpuArray_DIMS(top)[1] ||
+                topHeight != PyGpuArray_DIMS(top)[2] ||
+                topWidth != PyGpuArray_DIMS(top)[3]) {
+            PyErr_Format(PyExc_ValueError,
+                    "GpuCorrMM shape inconsistency:\n"
+                    "  bottom shape: %d %d %d %d\n"
+                    "  weight shape: %d %d %d %d %d %d\n"
+                    "  top shape: %ld %ld %ld %ld (expected %d %d %d %d)\n",
```
gvtulder: Idem.

gvtulder:
For what it's worth: I tried to use `gemv` in the forward convolution [`gpuarray/corr_gemm.c`](https://github.com/vikramnitin9/Theano/blob/5b6e1e62dcfef552f2a9eaa798ab60484ff988ab/theano/gpuarray/corr_gemm.c#L557-L562) on my system. With `device=cuda`, it works without problems. `opencl` on my GPU worked as well. Only `opencl` on CPU with the Intel OpenCL library had a segfault.

It may be interesting to change the GPU version to use the batched `gemm` functions, instead of calling `gemm` multiple times. (But it's more important to get the tests to work first, I think.)

vikramnitin9:
Interestingly, I had a segfault with opencl on the Intel Integrated HD Graphics device, but it ran fine on my i5 CPU.

gvtulder:
When you start working on the rebase, be careful that you keep things backwards compatible: put the new parameters, props etc. for unshared convolution after the parameters that have already been added for the grouped convolution.

theano/gpuarray/tests/test_abstractconv.py
vikramnitin9: I've changed this, thanks

theano/gpuarray/tests/test_abstractconv.py
vikramnitin9: For the 3D case, I've put a [check](https://github.com/Theano/Theano/pull/6007/files#diff-3aa5dca5e3dfdb813cd63e62f174bf06R383) in `run_fwd` to SkipTest if it is not a 2D convolution.

vikramnitin9: Actually, that would require passing the parameter here, which I haven't done. Rather than add it here, I'll just put a SkipTest right here.

vikramnitin9:
Update : I'm working on the integration with grouped convolution. Should be done tomorrow or day after.

vikramnitin9:
I've squashed the commits, rebased and started a new PR since this one is already very messy. That is at #6286. I'm closing this PR.

PR #6009, opened 2017-06-05
[doc/requirements.inc] Specify which 'lib' or 'lib64' should be added…

nalepae:
… to LD_LIBRARY_PATH

When reading this documentation, I first added '/lib' and '/lib64' in LD_LIBRARY_PATH.
In order to avoid other users to do this mistake, I propose to update the
documentation like this.


nouiz:
Good idea. I'll merge when travis pass. thanks.

nalepae:
Thanks for your amazing framework by the way!

nouiz:
Welcome :)

PR #6012, opened 2017-06-05
Fix offset problems in the new backend.

abergeron:


theano/gpuarray/corr3d_gemm.c
```diff
@@ -35,14 +35,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // GPU kernel for the case of dilation
 KERNEL void dilated_im3d2col_kernel(const ga_size n,
     GLOBAL_MEM const DTYPE_INPUT_0 * data_im,
+    const ga_size offset_im,
     const ga_size data_im_offset,
```
lamblin: What's the difference between `offset_im` and `data_im_offset`?

theano/gpuarray/corr3d_gemm.c
```diff
@@ -80,16 +84,20 @@ KERNEL void dilated_im3d2col_kernel(const ga_size n,
   }
 }
 
-#kernel im3d2col_kernel : size, *, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, * : 
+#kernel im3d2col_kernel : size, *, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, *, size :
 KERNEL void im3d2col_kernel(const ga_size n,
     GLOBAL_MEM const DTYPE_INPUT_0 * data_im,
+    const ga_size offset_im,
     const ga_size data_im_offset,
```
lamblin: Same here

theano/gpuarray/corr3d_gemm.c
```diff
@@ -137,7 +146,10 @@ KERNEL void dilated_col2im3d_kernel(const ga_size n,
     const ga_size stride_h, const ga_size stride_w, const ga_size stride_d,
     const ga_size height_col, const ga_size width_col, const ga_size depth_col,
     GLOBAL_MEM DTYPE_INPUT_0 * data_im,
+    const ga_size offset_im,
     const ga_size data_im_offset) {
```
lamblin: Same here.

theano/gpuarray/corr3d_gemm.c
lamblin: As discussed IRL, it would be nice to have a comment, for instance something like
"`offset_im` corresponds to the data pointer offset of the GpuArray object, whereas `data_im_offset` is an offset of elements inside the array".

theano/gpuarray/corr3d_gemm.c
```diff
@@ -201,6 +207,7 @@ KERNEL void dilated_col2im3d_kernel(const ga_size n,
 }
 
 #kernel col2im3d_kernel : size, *, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, size, *, size, size :
+
```
nouiz: remove

theano/gpuarray/corr_gemm.c
```diff
@@ -75,10 +77,13 @@ KERNEL void dilated_im2col_kernel(const ga_size n,
 }
 
 #kernel im2col_kernel : size, *, size, size, size, size, size, size, size, size, size, size, size, size, *, size :
+
```
nouiz: remove. It isn't like this everywhere. Why do you add that?

theano/gpuarray/corr_gemm.c
```diff
@@ -165,6 +172,7 @@ KERNEL void dilated_col2im_kernel(const ga_size n,
 }
 
 #kernel col2im_kernel : size, *, size, size, size, size, size, size, size, size, size, size, size, size, *, size, size :
+
```
nouiz: idem

theano/gpuarray/corr3d_gemm.c
abergeron: I feel it's much easier to read when there is a space between the "header" and the function.  Why do you want to remove it?

nouiz: just for consistency with the rest of the code. But I agree, it look clearner. So i'm good to keep it. But we should ideally do this everywhere at some point. I'm merging.

PR #6016, opened 2017-06-06
Add script to download and prepare mkl environment

TaoLv:
As mentioned in issue #5580  and PR #5770 , in order to make MKL OPs more convenient to use, we would like to provide a script to download the small MKL library and set up environment for Theano with MKL automatically.
To align with MKL code  in #5770 , this script is located in theano/contrib/mkl/.
 For Theano users, they just need source this script to set up environment:
`source prepare_mkl.sh`


nouiz:
As we stop developing Theano, I'll close this issue. We don't want to support more case and we already support MKL from conda.

PR #6022, opened 2017-06-07
Fix doc typos and spelling errors. Make spelling of "optimize" consistent.

arkeet:
"Optimise" (and its derived words) only occurred twice, while "optimize" was overwhelmingly popular. I also saw a fairly even mixture of "behavior" and "behaviour", so I hesitate to change these for lack of guidance on adopting U.S./Canadian/British spelling.


lamblin:
I don't think we have real guidelines for spelling, given that Canadian spelling is somewhat between UK and US, and the main developers speak mostly French anyway :)
I think most of the time US spelling is used (even though `theano.tensor.nnet.neighbours`), so if you want to harmoni~s~ze to "behavior" I don't think we would object, unless it is in a class or variable name.

Issue #5701, opened 2017-03-12
while using NanGuardMode(nan_is_error=True ) -->  basic.py  in min raise NotImplementedError() NotImplementedError


jc-Rosichini:
Just went though the following issue when trying to catch a Nan using mode=NanGuardMode(nan_is_error=True ) 

**my config:**
theano: 0.9.0rc2.dev-674f8d5148cb8dcea446e7243e629d8f83647da5
numpy: 1.11.0
pygpu: 0.6.1
lasagne: 0.2.dev1
Using cuDNN version 5105
GeForce GTX 980M
Windows 10 x64

**compiled training function**
train_fn = theano.function([net.input_var, net.target_var], [loss, I, U, acc], updates=updates
 , mode=NanGuardMode(nan_is_error=True )                               

> Compilation starts at 2017-03-12 16:12:29
optimizer: rmsprop, initial learning rate:  0.001
train compilation took 169.327 seconds
Compilation starts at 2017-03-12 16:15:22
valid compilation took 116.101 seconds
Validation score before pursuing training 0.0539679
Training restarts at 2017-03-12 16:17:24

> Traceback (most recent call last):
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\vm.py", line 482, in __call__
>     _, dt = self.run_thunk_of_node(current_apply)
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\gof\vm.py", line 402, in run_thunk_of_node
>     compute_map=self.compute_map,
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\nanguardmode.py", line 344, in nan_check
>     do_check_on(storage_map[var][0], node)
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\nanguardmode.py", line 296, in do_check_on
>     if contains_nan(value, nd, var):
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\nanguardmode.py", line 121, in contains_nan
>     return np.isnan(f_gpua_min(arr.reshape(arr.size)))
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\compile\nanguardmode.py", line 227, in result
>     f = theano.function([guard_in], op(guard_in),
>   File "C:\Users\PCGP1\Anaconda3\lib\site-packages\theano\tensor\basic.py", line 1742, in min
>     raise NotImplementedError()
[...]

PR #6025, opened 2017-06-08
[ENH] Speed up nanguardmode by not checking *int* dtype

nouiz:
suggested in gh-5701

### DIFF
--- a/theano/compile/nanguardmode.py
+++ b/theano/compile/nanguardmode.py
@@ -106,7 +106,7 @@ def contains_nan(arr, node=None, var=None):
     """
     if not _is_numeric_value(arr, var):
         return False
-    elif "int" in str(getattr(arr, 'dtype', '')):
+    elif getattr(arr, 'dtype', '') in T.discrete_dtypes:
         return False
     elif pygpu_available and isinstance(arr, GpuArray):
         return np.isnan(f_gpua_min(arr.reshape(arr.size)))
@@ -141,7 +141,7 @@ def contains_inf(arr, node=None, var=None):
     """
     if not _is_numeric_value(arr, var):
         return False
-    elif "int" in str(getattr(arr, 'dtype', '')):
+    elif getattr(arr, 'dtype', '') in T.discrete_dtypes:
         return False
     elif pygpu_available and isinstance(arr, GpuArray):
         return (np.isinf(f_gpua_min(arr.reshape(arr.size))) or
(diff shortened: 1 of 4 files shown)


theano/compile/nanguardmode.py
```diff
@@ -106,6 +106,8 @@ def contains_nan(arr, node=None, var=None):
     """
     if not _is_numeric_value(arr, var):
         return False
+    elif "int" in str(getattr(arr, 'dtype', '')):
```
lamblin: Use `discrete_dtypes`, so it will work with Booleans as well

theano/compile/nanguardmode.py
```diff
@@ -139,6 +141,8 @@ def contains_inf(arr, node=None, var=None):
     """
     if not _is_numeric_value(arr, var):
         return False
+    elif "int" in getattr(arr, 'dtype', ''):
```
lamblin: Same here.

theano/compile/tests/test_nanguardmode.py
```diff
@@ -69,3 +69,14 @@ def test_NanGuardMode():
...
+
+    # Test with uint, we should not crash
+    x = T.matrix(dtype='uint8')
+    w = theano.shared(np.random.randn(5, 7).astype(theano.config.floatX))
+    y = T.dot(x, w)
+    a = (np.random.randn(3, 5).astype(theano.config.floatX) * 10).astype('uint8')
+    fun = theano.function(
+        [x], y,
+        mode=NanGuardMode(nan_is_error=True, inf_is_error=True)
+    )
```
lamblin: This would not have reproduced the crash in #5701 anyway, as we do not compile a Theano function using `T.min` unless we are executing on GPU.

theano/compile/nanguardmode.py
lamblin: It is crashing anyway because you get the `dtype` object here, not a string.

theano/compile/tests/test_nanguardmode.py
nouiz: I tested it manually with device=cuda and there was no more crash.

lamblin: Did you have the crash before your changes?

lamblin: I just ran the test with the master of Theano, and even with `device=cuda` (which is not supposed to happen when running tests), it still passes.

theano/compile/nanguardmode.py
lamblin: You're still missing Booleans. Why don't you want to use `getattr(arr, 'dtype', '') in T.discrete_dtypes`?

nouiz:
done

abergeron:
flake8

PR #6029, opened 2017-06-10
Fix issue in optimizations with bitwise operations

### LINKED ISSUE #4633: Bitwise "and" and "or" broken
I've found that the bitwise "and" and "or" operators do not function correctly on Theano 0.8.0 and 0.8.2:

```
>>> import theano, theano.tensor as T
>>> theano.function([], T.and_(5,6))()
array(6, dtype=int8)    # Should be 4
>>> theano.function([], T.or_(5,6))()
array(1, dtype=int8)    # Should be 7
```

I've tried several other pairs of values. It appears that "T.or_" is performing a logical instead of bitwise operation, outputting 0 if both operands are zero and 1 otherwise. "T.and_" outputs 0 if either operand is zero, and the second operand otherwise.

By the way, "and_" and "or_" are indeed intended as bitwise (not logical) operators, because they have the aliases "bitwise_and" and "bitwise_or".

However, the following is correct:

```
>>> theano.function([], theano.scalar.and_(5,6))()
4
>>> theano.function([], theano.scalar.or_(5,6))()
7
```

The bug did not exist in Theano 0.7.0.

lamblin:
Fix #4633

### DIFF
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -3493,8 +3493,12 @@ class Test_local_useless_elemwise_comparison(unittest.TestCase):
         topo = f.maker.fgraph.toposort()
         assert len(topo) == 1
         assert topo[0].op == deep_copy_op
-        x_val = 10
-        assert f(x_val) == x_val
+        if f.outputs[0].variable.dtype == 'bool':
+            x_vals = [0, 1]
+        else:
+            x_vals = [0, 1, 10]
+        for x_val in x_vals:
+            assert f(x_val) == x_val
 
     def test_inequality_with_self(self):
         x = T.scalar('x', dtype=config.floatX)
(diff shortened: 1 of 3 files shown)


theano/tensor/tests/test_opt.py
```diff
@@ -3601,50 +3605,69 @@ def test_equality_shapes(self):
...
-        for zero, one in [(np.int8(0), np.int8(1)), (0, 1)]:
             f = theano.function([x], T.and_(x, zero), mode=mode)
             self.assert_eqs_const(f, 0)
 
             f = theano.function([x], T.and_(zero, x), mode=mode)
             self.assert_eqs_const(f, 0)
 
             f = theano.function([x], T.and_(x, one), mode=mode)
-            if f.outputs[0].variable.dtype == x.dtype:
+            if dtype == 'bool':
```
nouiz: This condition is useless. Idem bellow. You already check for bool dtype inside assert_identity

lamblin: But it should _not_ be the identity if the dtype is not bool.
For instance, `T.and_(2, one)` should give `0`, and `T.and_(3, one)` should give `1`.

PR #6030, opened 2017-06-10
Add lifter for CrossentropyCategorical1Hot and grad

### LINKED ISSUE #5036: CrossentropyCategorical1Hot() not on GPU
I have see a case where CrossentropyCategorical1Hot was not on the GPU. We don't have it implemented, just CrossentropyCategorical1Hot with the softmax was implemented on the GPU.

It happen that there is something between the softmax and the crossentropy. This currently make this not working on the GPU.

As we implemented logsoftmax on the CPU and GPU, I think we can just implement CrossentropyCategorical1Hot on the GPU and we will support those cases while staying numericaly stable?

lamblin:
Fix #5036.

Unfortunately, there is no easy way of introducing the GPU Ops for subtensor and inc_subtensor directly, so they have to be cleaned up in a later pass.

I did not see GpuAdvancedSubtensor or even AdvancedIncSubtensor to be an issue in practice, it did not even register in the profile. It could be different on other models.

### DIFF
--- a/theano/gpuarray/tests/test_opt.py
+++ b/theano/gpuarray/tests/test_opt.py
@@ -689,7 +689,6 @@ def test_crossentropycategorical1hot_lifter():
     z = tensor.nnet.crossentropy_categorical_1hot(x, y)
     gx = theano.grad(z.mean(), x)
     f = theano.function([x, y], [z, gx], mode=mode_with_gpu)
-    theano.printing.debugprint(f, print_type=1)
     assert not any(isinstance(n.op, (tensor.nnet.CrossentropyCategorical1Hot,
                                      tensor.nnet.CrossentropyCategorical1HotGrad))
                    for n in f.maker.fgraph.apply_nodes)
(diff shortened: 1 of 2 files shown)


theano/gpuarray/tests/test_opt.py
```diff
@@ -680,3 +680,18 @@ def randX(*args):
...
+
+
+def test_crossentropycategorical1hot_lifter():
+    rng = np.random.RandomState(utt.fetch_seed())
+    x = tensor.matrix()
+    y = tensor.lvector()
+    z = tensor.nnet.crossentropy_categorical_1hot(x, y)
+    gx = theano.grad(z.mean(), x)
+    f = theano.function([x, y], [z, gx], mode=mode_with_gpu)
+    theano.printing.debugprint(f, print_type=1)
```
nouiz: remove this line. Otherwise I'm ok to merge. I ran that test in DebugMode and it passed.

lamblin: Oops.

lamblin:
Updated.

Issue #6035, opened 2017-06-12
Rop is now Obsolete for Graph based Autodiff


botev:
So I waited for one of my colleges, who first discovered this, to finalise this blog before I put it here. 
I think it is very important since for Theano it makes the R-op totally unscary and its maintenance pointless. 
It also means that as long you have a Lop implemented you can already have an R-op.
The whole blog post is here:
https://j-towns.github.io/2017/06/12/A-new-trick.html
The TL;DR
```
def Rop(f, x, u):
    v = f.type()
    g = T.Lop(f, x, v)
    return T.Lop(g, v, u)
```
This is always true and valid for anything that the Lop aka the gradient is implemented. 
You can as well use it to actually re-test the R-operators.

lamblin:
I will have a look soon.
If I understand correctly, we would need additional graph optimizations in order to generate graphs as clean as what one call to Rop would, but thankfully that is something Theano can do.
It may have been known before, but not by us, although I started to observe a pattern that 2nd order grad and R_op could be implemented with the same new Ops (for MaxPooling for instance).

nouiz:
Note, I would prefer that we keep generating a clean graph when possible instead of relying on optimization when possible. But this could be the default implementation and each Op could optionally implement a more optimal implementation.

botev:
@nouiz  Please note that the **only** optimisation needed is dead code elimination. There is nothing else there to optimise. The graphs constructed via the double Lop is exactly identical to the one with the Rop, just with an extra branch, which however can be eliminated using dead code elimination. There is no other optimisation needed. 
Just to illustrate let's got rough GEMM:
```
dW, dx - directional vectors
dy - auxiliary variable for y
y = Wx
Rop(y, [W, x], [dW, dx]) = dW x + W dx
Lop(y, [W, x], dy) = [dy x^T, W^T dy]
Lop(Lop(y, [W, x], dy), dy, [dW, dx]) = Lop([dy x^T, W^T dy], dy, [dW, dx]) 
= Lop(dy x^T, dy, dW) + Lop(W^T dy, dy, dW) = dW x + W dx
```
E.g. for Theano all of the ops in the first Lop which has `dy` will be dad code an will be eliminated from the computation. Nothing else is needed.

j-towns:
[...]

PR #6037, opened 2017-06-13
use twice Lop to get Rop by default

khaotik:
Quick PR for #6035

### DIFF
--- a/theano/gof/op.py
+++ b/theano/gof/op.py
@@ -708,7 +708,13 @@ class PureOp(object):
     #########################
 
     def L_op(self, inputs, outputs, output_grads):
-        return self.grad(inputs, output_grads)
+        if hasattr(self, 'grad'):
+            return self.grad(inputs, output_grads)
+        else:
+            raise NotImplementedError(
+                "%s of class %s do not have grad or L_op implementation."
+                "If this is theano Op, mail to theano-dev mailing list for assistance."
+                "If this is your own Op, implement them." % (self, self.__class__.__name__))
 
     def R_op(self, inputs, eval_points):
         """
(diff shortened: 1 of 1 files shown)


theano/gof/op.py
```diff
@@ -708,7 +708,13 @@ def __ne__(self, other):
     #########################
 
     def L_op(self, inputs, outputs, output_grads):
-        return self.grad(inputs, output_grads)
+        if hasattr(self, 'grad'):
+            return self.grad(inputs, output_grads)
+        else:
+            raise NotImplementedError(
+                "%s of class %s do not have grad or L_op implementation."
+                "If this is theano Op, mail to theano-dev mailing list for assistance."
```
nouiz: theano -> Theano

theano/gof/op.py
```diff
@@ -733,13 +739,17 @@ def R_op(self, inputs, eval_points):
...
-        raise NotImplementedError(
-            "%s of class %s does not "
-            "implement R_op. If this is a theano op, write to the "
-            "theano-dev mailing list for assistance. If it is your "
-            "own op, implement the R_op method." %
-            (self, self.__class__.__name__))
+        outputs = self.make_node(inputs).outputs
+        lop_inputs = [o.type() for o in outputs]
+        lop_outputs = self.L_op(inputs, outputs, lop_inputs)
+        return theano.Lop(lop_outputs, lop_inputs, eval_points)
```
nouiz: This would need some tests to be sure we don't break it.

It could be a simple test that create a new op with Lop implemented, but not Rop and make sure we can take the Rop. This new Op could inherit from an existing op and delete the Rop method.

botev:
I would actually suggest to possibly add a test, where you go loop to all implemented ops and if they implement Rop you verify that this gives the same result and the standard Rop. 
Additionally, it might be interesting to benchmark and see if there are any Ops which have faster-implemented Rops or not.

lamblin:
Agreed with @botev.
The place for such a test would be `theano/tests/test_rop.py`, and the existing Ops with R_op defined should all be mentioned there already.

PR #6038, opened 2017-06-13
(small fix) Add doc for `dnn.include_path` and `dnn.library_path`.

notoraptor:
It seems these two Theano flags are not in the documentation: http://deeplearning.net/software/theano_versions/dev/library/config.html

### DIFF
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -638,13 +638,13 @@ import theano and print the config variable, as in:
 
 .. attribute:: config.dnn.include_path
 
-    Default: ``include`` sub-folder in CUDA root directory.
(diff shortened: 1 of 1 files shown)


doc/library/config.txt
```diff
@@ -636,6 +636,18 @@ import theano and print the config variable, as in:
 
     If ``'no_check'``, assume present and the version between header and library match (so less compilation at context init)",
 
+.. attribute:: config.dnn.include_path
+
+    Default: ``''``
```
nouiz: I would put the phrase in parenthesis here. Something like:

Default: cuda_root/include or ''

notoraptor:
Updated.

notoraptor:
Updated.

PR #6040, opened 2017-06-13
md5 to sha256

nouiz:
Rebased https://github.com/Theano/Theano/pull/5916, as its last commit had no timeout. Also, I merged today some tests change could removed some 10m timeout. So mostly, testing it again in case it would now work.

The original PR had problems with the merge that make it not mergeable. Maybe the call to theano-cache purge would be what fixed his timing. We will see.


slefrancois:
jenkins test this

nouiz:
Will start again from: https://github.com/Theano/Theano/pull/5916

PR #6042, opened 2017-06-14
Extend DiffOp so it allows to take gradients of multidimensional inputs

ASpeiser:
Edited DiffOp so you can take gradients for inputs with more than 1 dimension

### DIFF
--- a/theano/tensor/extra_ops.py
+++ b/theano/tensor/extra_ops.py
@@ -453,15 +453,11 @@ class DiffOp(theano.Op):
     def grad(self, inputs, outputs_gradients):
         inputs = inputs[0]
 
-        if inputs.ndim != 1:
-            raise NotImplementedError("Grad is not implemented for inputs with"
-                                      "number of dimension other than 1.")
-
         z = outputs_gradients[0]
 
         def _grad_helper(z):
-            pre = basic.concatenate([[0.], z])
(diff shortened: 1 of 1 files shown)


theano/tensor/extra_ops.py
```diff
@@ -453,15 +453,11 @@ def perform(self, node, inputs, output_storage):
...
-        if inputs.ndim != 1:
-            raise NotImplementedError("Grad is not implemented for inputs with"
-                                      "number of dimension other than 1.")
-
         z = outputs_gradients[0]
 
         def _grad_helper(z):
-            pre = basic.concatenate([[0.], z])
-            app = basic.concatenate([z, [0.]])
+            pre = basic.concatenate([T.zeros(T.sum(z,axis=self.axis,keepdims=True).shape), z],axis=self.axis)
```
nouiz: thanks for the PR. The test theano/tests/test_flake8.py will fail. You need a space after the comma. I'll stop travis as it won't help build this PR and yesterday it was overloaded.

This PR also need test in the file tensor/tests/test_extra_ops.py to test the new fonctionality.

nouiz:
We will do a Theano 0.10beta3 Monday. So if someone can add the missing tests, it could get included. I don't have time to do this.

Issue #6033, opened 2017-06-12
Clean up OutputGuard


lamblin:
The check for inserting OutputGuard in AddDestroyHandler should not be performed each time we are optimizing a graph, we should guarantee that such a case does not happen in practice, and keep those kinds of check only in DebugMode (as an actual check that fails if there is an issue).

- [ ] Add warning (or even an error) when AddDestroyHandler is inserted and there is no Supervisor.
It indicates that the Supervisor will not be able to check for protected variables, and at that point it is too late to know which variables will be protected.
This case can happen when code calls `some_optimizer.optimize(fgrapfh)`, not going through `theano.function(...)`.
- [ ] Stop trying to insert OutputGuard in AddDestroyHandler
- [ ] Transfer that check over to DebugMode, but make it raise an Exception if the OutputGuard gets inserted successfully.

lamblin:
Done in #6044.

PR #6044, opened 2017-06-15
Remove output guard

ReyhaneAskari:
Regarding https://github.com/Theano/Theano/issues/6033.  I didn't completely understand the 3rd point. I don't need to change the place of the OutputGuard class, right? Just change the mechanism somehow that we would see a failure when outputguard got inserted successfully?

Do we want to benchmark this?

### DIFF
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -2312,6 +2312,13 @@ class _Maker(FunctionMaker):  # inheritance buys a few helper functions
                               "of", len(li), "events was stable.",
                               file=sys.stderr)
         self.fgraph = fgraph
+        destroy_handler_added = False
+        for feature in fgraph._features:
+            if isinstance(feature, gof.DestroyHandler):
+                destroy_handler_added = True
+                break
+        if not destroy_handler_added:
+            fgraph.attach_feature(gof.DestroyHandler())
         for o in fgraph.outputs:
             try:
                 fgraph.replace_validate(o, _output_guard(o), reason='output_guard')
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -27,7 +27,8 @@ from theano.compile.function_module import (
     FunctionMaker, Function, infer_reuse_pattern,
     SymbolicOutput, Supervisor, std_fgraph)
 from theano.compile.mode import Mode, register_mode
-from theano.compile.ops import OutputGuard
+from theano.compile.ops import OutputGuard, _output_guard
+
 
 __docformat__ = "restructuredtext en"
 _logger = logging.getLogger("theano.compile.debugmode")
@@ -2311,6 +2312,17 @@ class _Maker(FunctionMaker):  # inheritance buys a few helper functions
                               "of", len(li), "events was stable.",
                               file=sys.stderr)
         self.fgraph = fgraph
+        for o in fgraph.outputs:
+            try:
+                fgraph.replace_validate(o, _output_guard(o), reason='output_guard')
+                raise Exception("Output variable %s required output_guard, "
+                                "how was this output left unprotected against "
+                                "destructive operations?" % o)
+
+            except gof.InconsistencyError:
+                # This output is already impossible to destroy.
+                # No guard necessary
+                pass
 
         linker = _Linker(self)
 
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -4,6 +4,7 @@ WRITEME
 """
 from __future__ import absolute_import, print_function, division
 import logging
+import warnings
 
 import theano
 from theano import gof
@@ -11,6 +12,7 @@ import theano.gof.vm
 from theano.configparser import config
 from theano.compile.ops import _output_guard
 from six import string_types
+from theano.compile.function_module import Supervisor
 
 
 _logger = logging.getLogger('theano.compile.mode')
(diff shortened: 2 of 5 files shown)


abergeron:
The point number 3 would be to add a check in DebugMode whether outputs are destroyed or not.

You can probably move the OutputGuard class into the DebugMode code and add a phase that tries to insert OutputGuards, which should succeed.  There might also be another way to check for that, but I'm not sure.

theano/compile/mode.py
```diff
@@ -111,6 +113,16 @@ class AddDestroyHandler(gof.Optimizer):
 
     """
     def apply(self, fgraph):
+        supervisor_added = False
+        for feature in fgraph._features:
+            if isinstance(feature, Supervisor):
+                supervisor_added = True
+                return
```
nouiz: return -> break

You want to break the loop, not return from the method.

ReyhaneAskari: Thanks, I will fix it but it wouldn't make any difference since the ```if``` would always be false when we returned.

nouiz:
I would not move the OutputGuard class, as some old pickle could still have it. Or if it get moved, make sure to import it at the old place to don't break stuff.

theano/compile/mode.py
```diff
@@ -111,6 +113,16 @@ class AddDestroyHandler(gof.Optimizer):
...
+        if not supervisor_added:
+            warnings.warn("WARNING: Supervisor is not added. Please do not"
+                          "use some_op.optimize(fgraph). Instead use theano.function"
+                          "to add the optimiztions.",
```
nouiz: The message isn't good. I think what we need to tell is something like:

WARNING: Supervisor is not added. Please build a FunctionGraph via theano.compile.function_module.std_graph() or add the Supervisor class manually.

theano/tensor/basic.py
```diff
@@ -447,7 +447,6 @@ def get_scalar_constant_value(orig_v, elemwise=True,
                 max_recur > 0):
             max_recur -= 1
             if isinstance(v.owner.op, (Alloc, DimShuffle, Rebroadcast,
-                                       compile.ops.OutputGuard,
```
nouiz: I would keep OutputGuard here for now.

theano/compile/mode.py
ReyhaneAskari: Thanks. Done.

theano/tensor/basic.py
ReyhaneAskari: Why do we want to keep it here?

nouiz: In case of old pickled stuff. I'm not 100% it is needed, but I prefer to be safe. It cost nothing to keep it there as we already keep the class. Maybe add a comment like:

We keep it here just in case some old pickled file need that.

nouiz:
some tests fail. Maybe it is just the error type that was changed?

theano/tensor/basic.py
ReyhaneAskari: I see. Thanks. I'll revert the change and add a comment.

ReyhaneAskari:
Thanks Fred. There are 2 failing tests that appear in both jenkins and travis: ```test_debugmode.py:test_badoptimization``` and ```test_debugmode.py:test_badoptimization_opt_err```. The exception that we added for outputGuard is raised with these tests. In both cases ```mode``` is set to ```debugmode.DebugMode(optimizer=opt)``` (these are the only tests with this mode in test_debugmode.py). Before the check for the outputguard I check the ```fgraph._features```. The ```Supervisor``` was added but not the ```destroy_handler```. So ```fgraph.destroyers``` was never called in ```on_attach``` and ```Supervisor.validate``` could never do anything due to this check:

```python 
            return True
```

I think there is something wrong with either the  ```debugmode.DebugMode``` or the opts that these tests used. What do you suggest?

abergeron:
The tests replace all the optimizers with a single local optimizer to test broken optimizations/ops.  This also removes all the optimizers that would insert destroy_handler.

I belive that it is DebugMode we need to fix here, but there are two ways we can do it and I'm not sure which one to take.  The first is to condition the check in DebugMode to the presence of the destroy handler.  The second is to have DebugMode add the destroy handler to be able to perform the check.

What do you think?

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1338,9 +1332,8 @@ def test_argmax_pushdown():
         # print 'AFTER'
         # for node in fgraph.toposort():
         # print node.op
-        assert len(fgraph.toposort()) == 2  # an output_guard is second
+        assert len(fgraph.toposort()) == 1  # an output_guard is second
```
ReyhaneAskari: Thanks. Done.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1364,12 +1357,11 @@ def test_argmax_pushdown():
         # print 'AFTER'
         # for node in fgraph.toposort():
             # print node.op
-        assert len(fgraph.toposort()) == 4  # an output_guard is second
+        assert len(fgraph.toposort()) == 3  # an output_guard is second
```
ReyhaneAskari: Done.

ReyhaneAskari:
I guess both ways would work and it narrows down to whether destroy handler is actually required to be there or not.  From your explanations my understanding is that the destroy handler is removed unintentionally. So maybe the better way is to add the destroy handler if it is not already attached. Do you agree?

abergeron:
I think this is the safer choice, yes.

ReyhaneAskari:
@abergeron I guess we are good to merge.

theano/compile/debugmode.py
```diff
@@ -2311,6 +2312,24 @@ def __init__(self, inputs, outputs, mode,
                               "of", len(li), "events was stable.",
                               file=sys.stderr)
         self.fgraph = fgraph
+        destroy_handler_added = False
```
lamblin: In fact, I think all of that block should be skipped if `cycle_detection=='fast'`, as trying to insert `_output_guard` that way does not actually insert a new cycle, and is not detected as a cycle by the 'fast' algo.
To be more precise: the old cycle detection prevents working inplace on outputs that existed _before_ the optimization was applied. The "fast" one considers the outputs _after_ the substitution, so there's no issue.

PR #6045, opened 2017-06-15
Work around erroneous dnn_conv results with algo small an large batches

### LINKED ISSUE #5985: erroneous dnn_conv result with algo 'small' an large batches
Hi,

I have a call to a forward pass through a convnet using `theano.gpuarray.dnn.dnn_conv()` as the convolution for a Lasagne `Conv2DLayer`.  It gives erroneous results from forward pass for `dnn.conv.algo_fwd=small` and for large batch sizes.  Has anyone had this problem?  Might be a cuDNN thing?

I found that if I compute some values using the entire batch in one function call, they are not the same as when calling the function multiple times using only a subset of the data each time.  The effect kicks in at batch size about 70,000 in my case, corresponding to about a couple of GBs.  Instead of the comparison result returning `0.0`, it will be at least `1e-5`.  Another strange thing is that as my algorithm progresses, checking the batch vs multi-call results in values at each iteration, the comparison gradually diverges to values much greater than one.  

Calling the GPU function on data subsets is stable and results in overall comparison `1e-9` or less against values computed on the CPU.  When computing everything on the CPU, it works fine.

I pulled latest Theano and built latest libgpuarray / pygpu today.

CuDNN 6020, CUDA 8.  GTX 1080 (8GB)

Any ideas?  

Thanks,
Adam

### LINKED ISSUE #6020: Make sure we check for overflows before calling cuDNN
The error in #5985 could be an issue in cuDNN, or in the wrapper code if we have overflow/wraparound in shapes (or strides).
We should go through the code and check that we correctly test for overflows.

Maybe also add that to the cuDNN extensive test list.

notoraptor:
A PR to help fix #5985 and fix #6020 .

Currently, I have added two tests to reproduce the problem: one with 10 000 as batch size, and another with 70 000 as batch size (all other parameters are identical). Then the test with 70 000 inputs fails (see console output below).

Currently, I don't think that the problem is about strides or dimensions, as all these values for current tested inputs may be handled by a system integer (`int`). Still investigating.

@abergeron @lamblin @nouiz 

```
$ nosetests -xvs theano/gpuarray/tests/test_dnn.py:test_batched_conv_success
Mapped name None to device cuda: GeForce GTX TITAN X (0000:02:00.0)
theano.gpuarray.tests.test_dnn.test_batched_conv_success((10000, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3)) ... Input size: 0.152587890625 Gb
Computing
Output shapes: (10000, 1, 6, 6) (1, 1, 6, 6)
ok

Ran 1 test in 26.128s

OK
$ nosetests -xvs theano/gpuarray/tests/test_dnn.py:test_batched_conv_fail
Mapped name None to device cuda: GeForce GTX TITAN X (0000:02:00.0)
theano.gpuarray.tests.test_dnn.test_batched_conv_fail((70000, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3)) ... Input size: 1.06811523438 Gb
Computing
Output shapes: (70000, 1, 6, 6) (1, 1, 6, 6)
ERROR

ERROR: theano.gpuarray.tests.test_dnn.test_batched_conv_fail((70000, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3))
  File "/u/boccoset/.conda/envs/boccosetEnv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/boccoset/mila/dev/git/theano/theano/gpuarray/tests/test_dnn.py", line 1104, in run_conv_batched_vs_multicall
    utt.assert_allclose(res_batch[i], res_all[size - batch_sub + i], atol=1e-6, rtol=1e-6)
  File "/u/boccoset/mila/dev/git/theano/theano/tests/unittest_tools.py", line 357, in assert_allclose
    raise WrongValue(val1, val2, rtol, atol)
WrongValue: WrongValue
  expected    : [[[[ 262.81149292  257.28344727  257.68167114  263.76553345  253.07557678
     254.87948608]
   [ 255.46412659  260.05529785  258.79153442  265.90890503  256.80264282
     249.12014771]
   [ 255.78485107  260.87399292  254.5199585   256.53036499  253.32684326
     252.63471985]
   [ 253.66004944  252.29753113  260.47439575  261.97869873  253.92384338
     259.56750488]
   [ 262.41348267  259.84350586  261.61276245  260.75463867  256.12188721
     253.10206604]
   [ 259.37319946  258.40789795  261.78692627  259.37680054  262.13986206
     255.63438416]]]]
  value    : [[[ 245.11114502  244.91969299  243.5846405   253.72325134  254.38612366
    255.17477417]
  [ 253.94763184  244.61291504  254.25749207  252.80860901  259.48983765
    255.08586121]
  [ 249.11782837  248.26765442  248.66650391  253.08294678  257.47140503
    253.18663025]
  [ 247.57498169  251.08808899  248.44218445  251.96794128  255.5085144
    252.05444336]
  [ 250.04606628  249.66490173  249.98672485  252.85968018  253.4756012
    247.5184021 ]
  [ 249.64077759  250.12475586  249.40283203  250.69918823  257.88140869
    251.75737   ]]]
  Max Abs Diff:  17.7003
  Mean Abs Diff:  7.45197
  Median Abs Diff:  7.09004
  Std Abs Diff:  4.62775
  Max Rel Diff:  0.06735
  Mean Rel Diff:  0.0287206
  Median Rel Diff:  0.0275048
  Std Rel Diff:  0.0176318

  rtol, atol: 1e-06 1e-06


Ran 1 test in 45.255s

FAILED (errors=1)
```

### DIFF
--- a/theano/gpuarray/dnn_fwd.c
+++ b/theano/gpuarray/dnn_fwd.c
@@ -171,8 +171,8 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
     algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
 
   // Algo `small` does not work for a batch size > 2^16, with cuDNN >= V5.1.
-  // Issue should be resolved for cuDNN > V6.0.20.
-  if (cudnnGetVersion() <= 6020 &&
+  // Issue should be resolved for cuDNN > V6.0.
+  if (cudnnGetVersion() < 6100 &&
       algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM &&
       PyGpuArray_DIM(input, 0) > 65536)
       algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
--- a/theano/gpuarray/dnn_fwd.c
+++ b/theano/gpuarray/dnn_fwd.c
@@ -163,10 +163,15 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
     }
   }
 
-  /* These two algos are not supported for 3d conv */
+  /* Only these algos are supported for 3d conv with cuDNN >= V5.1. */
   if (PyGpuArray_NDIM(input) == 5 &&
-      (algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM ||
-       algo == CUDNN_CONVOLUTION_FWD_ALGO_GEMM))
+      !(algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM ||
+        algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM ||
+        algo == CUDNN_CONVOLUTION_FWD_ALGO_FFT_TILING))
+    algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
+
+  /* Algo `small` seems to not work for a batch size > 2^16, with cuDNN >= V5.1. */
+  if (algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM && PyGpuArray_DIM(input, 0) > 65536)
     algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
 
   // The FFT implementation does not support strides, 1x1 filters or inputs
--- a/theano/gpuarray/dnn_fwd.c
+++ b/theano/gpuarray/dnn_fwd.c
@@ -170,9 +170,12 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
         algo == CUDNN_CONVOLUTION_FWD_ALGO_FFT_TILING))
     algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
 
-  /* Algo `small` seems to not work for a batch size > 2^16, with cuDNN >= V5.1. */
-  if (algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM && PyGpuArray_DIM(input, 0) > 65536)
-    algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
+  // Algo `small` does not work for a batch size > 2^16, with cuDNN >= V5.1.
+  // Issue should have been resolved for cuDNN >= V6.0.20.
+  if (cudnnGetVersion() <= 6020 &&
+      algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM &&
+      PyGpuArray_DIM(input, 0) > 65536)
+      algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
 
   // The FFT implementation does not support strides, 1x1 filters or inputs
   // with a spatial dimension larger than 1024. The tiled-FFT implementation
--- a/theano/gpuarray/dnn_fwd.c
+++ b/theano/gpuarray/dnn_fwd.c
@@ -171,7 +171,7 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
     algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
 
   // Algo `small` does not work for a batch size > 2^16, with cuDNN >= V5.1.
-  // Issue should have been resolved for cuDNN >= V6.0.20.
+  // Issue should be resolved for cuDNN > V6.0.20.
   if (cudnnGetVersion() <= 6020 &&
       algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM &&
       PyGpuArray_DIM(input, 0) > 65536)
--- a/theano/gpuarray/tests/test_dnn.py
+++ b/theano/gpuarray/tests/test_dnn.py
@@ -1067,15 +1067,19 @@ def get_conv3d_test_cases():
     return itt
 
 
-def run_conv_small_batched_vs_multicall(inputs_shape, filters_shape, batch_sub, subsample):
+def run_conv_small_batched_vs_multicall(inputs_shape, filters_shape, batch_sub):
     # Function to check issue #5985 (see tests below): https://github.com/Theano/Theano/issues/5985
 
+    # Error occurs with algorithm `small` (CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM)
     algo = 'small'
+
     batch_size = inputs_shape[0]
 
     utt.seed_rng()
     inputs_val = np.random.random(inputs_shape).astype('float32')
     filters_val = np.random.random(filters_shape).astype('float32')
+    # Scale down the input values to prevent very large absolute errors
+    # due to float rounding
     inputs_val /= 10
     filters_val /= 10
     inputs = theano.shared(inputs_val)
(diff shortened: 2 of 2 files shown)


slefrancois:
jenkins test this

notoraptor:
Update to make tests more precise. It seems tests start failing when input size it greater than 1 gigabyte. Current tests run with these shapes (fwd, algo small, dtype `float32`; subsample is set to (3, 3) just to reduce output size):
- input shape (65535, 4, 32, 32),  filter shape (1, 4, 16, 16), subsample (3, 3): PASS
- input shape (65536, 4, 32, 32),  filter shape (1, 4, 16, 16), subsample (3, 3): PASS
- input shape (65537, 4, 32, 32),  filter shape (1, 4, 16, 16), subsample (3, 3): **FAIL**

All strides and dimensions are still under int32 limits. Maybe a problem with any internal pointer that iterates over the whole input memory block (which is indeed big) ?

Still investigating.

Tested on TITAN X with cuDNN V6:
```
(boccosetPython2) [SLURM] boccoset@leto15:~/mila/dev/other/Theano$ nosetests -xvs theano/gpuarray/tests/test_dnn.py:test_batched_conv_small
Mapped name None to device cuda: GeForce GTX TITAN X (0000:06:00.0)
theano.gpuarray.tests.test_dnn.test_batched_conv_small((65535, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3)) ... (input size 0.999984741211 Gb) ok
theano.gpuarray.tests.test_dnn.test_batched_conv_small((65536, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3)) ... (input size 1.0 Gb) ok
theano.gpuarray.tests.test_dnn.test_batched_conv_small((65537, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3)) ... (input size 1.00001525879 Gb) ERROR

ERROR: theano.gpuarray.tests.test_dnn.test_batched_conv_small((65537, 4, 32, 32), (1, 4, 16, 16), 25, (3, 3))
  File "/u/boccoset/.conda/envs/boccosetPython2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/u/boccoset/mila/dev/other/Theano/theano/gpuarray/tests/test_dnn.py", line 1102, in run_conv_small_batched_vs_multicall
    utt.assert_allclose(res_batch[i], res_all[size - batch_sub + i])
  File "/u/boccoset/mila/dev/other/Theano/theano/tests/unittest_tools.py", line 357, in assert_allclose
    raise WrongValue(val1, val2, rtol, atol)
WrongValue: WrongValue
  expected    : [[[[ 252.8598175   249.0565033   253.22346497  250.5092926   251.46174622
     257.7545166 ]
   [ 254.42881775  253.95082092  248.78294373  246.41479492  253.34909058
     254.14724731]
   [ 248.40374756  249.57064819  250.97595215  252.41111755  249.61836243
     255.28413391]
   [ 243.89065552  244.64581299  246.30297852  245.46357727  245.64276123
     259.18984985]
   [ 244.61361694  248.20509338  243.40785217  242.1103363   245.04740906
     247.53959656]
   [ 249.13037109  246.27787781  249.16470337  244.68641663  247.45684814
     253.18463135]]]]
  value    : [[[ 252.38020325  257.27377319  255.3911438   260.16137695  265.61608887
    269.98699951]
  [ 253.45674133  254.73764038  255.47238159  263.4364624   262.04730225
    265.51553345]
  [ 250.00076294  248.17248535  249.62051392  250.40525818  254.76145935
    255.33004761]
  [ 248.78088379  251.12950134  251.70974731  248.04762268  251.18678284
    250.54277039]
  [ 251.86959839  250.79647827  250.22761536  254.01750183  247.85797119
    248.0612793 ]
  [ 248.8586731   248.77662659  249.46583557  246.38783264  248.00254822
    252.84272766]]]
  Max Abs Diff:  17.0217
  Mean Abs Diff:  4.86412
  Median Abs Diff:  2.70097
  Std Abs Diff:  4.4551
  Max Rel Diff:  0.0690773
  Mean Rel Diff:  0.0195176
  Median Rel Diff:  0.0109983
  Std Rel Diff:  0.017837

  rtol, atol: 1e-05 1e-05


Ran 3 tests in 142.416s

FAILED (errors=1)
```

@lamblin @abergeron @nouiz

notoraptor:
Update. I have added a test `theano.gpuarray.tests.test_dnn.test_batched_conv_small` to reproduce the problem. From the results, the error occurs when inputs batch size becomes greater than 655536 (2**16) with algorithm `small` and forward convolution, even with small inputs (e.g. with shape (65537, 2, 2, 2)). It seems computation is not correctly done anymore after the 65536th image (in fact, the conv result at position 65536 is equal to the conv result at position 0, as if a modulo loop has been done).

Looking into `dnn_fwd.c` and `dnn_base.c`, I don't find any code that might manage badly the input shape. As we use `int` integers everywhere, a size of `2**16` should not be a problem. So, currently, I don't know how to resolve this issue. @lamblin @nouiz @abergeron .

notoraptor:
jenkins test this

abergeron:
This is most likely the fault of cuDNN, so we can't fix this, only work around it (and probably report the issue).

notoraptor:
Updated. Tests added for conv 3D (fails too), and fix added (algo small is replaced with none if such case in encountered, see `dnn_fwd.c`).

I'll make the bug report. Maybe should we wait for the bug to be checked by NVIDIA before merging this PR into master branch.

notoraptor:
Bug submitted: https://developer.nvidia.com/nvidia_bug/1950985 .

notoraptor:
NVIDIA responded on bug thread, saying that it is a known issue that has been fixed for the next cuDNN release: https://developer.nvidia.com/nvidia_bug/1950985 .

So, I'll just update this PR to apply the workaround only for cuDNN releases <= 6020, and I will also add this test to the test script of PR #5932 , so that we could automatically check it for next cuDNN releases.

notoraptor:
Updated. This PR can be rewiewed (and merged if good) @abergeron @nouiz @lamblin .

theano/gpuarray/dnn_fwd.c
```diff
@@ -163,12 +163,20 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
...
   if (PyGpuArray_NDIM(input) == 5 &&
-      (algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM ||
-       algo == CUDNN_CONVOLUTION_FWD_ALGO_GEMM))
+      !(algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM ||
+        algo == CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM ||
+        algo == CUDNN_CONVOLUTION_FWD_ALGO_FFT_TILING))
     algo = CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM;
 
+  // Algo `small` does not work for a batch size > 2^16, with cuDNN >= V5.1.
+  // Issue should have been resolved for cuDNN >= V6.0.20.
```
nouiz: >= should be >

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1067,6 +1067,63 @@ def get_conv3d_test_cases():
...
+        if p >= checked_limit:
+            # It seems results are repeated in the entire conv.
+            # It should not happen.
+            if np.allclose(res_all[p % checked_limit], res_all[p]):
+                print('\nconv[%d] == conv[%d] == %s' % (p % checked_limit, p, res_all[p]))
+        utt.assert_allclose(res_batch_bottom[i], res_all[p])
+
+
+def test_batched_conv_small():
+    yield (run_conv_small_batched_vs_multicall, (65534, 2, 2, 2), (1, 2, 2, 2), 5)  # OK
```
nouiz: this is in each PR tests run. To help keep it fast, only test the size 65536 and 65537 in both test.

notoraptor:
Updated.

notoraptor:
Still fail with cuDNN version 6021 (> 6020) on Jenkins. Does `next release` mean V6.1 (or V7) ?

nouiz:
Suppose that it will work for 6.1. We don't know if such version will
exist, but as it is a bug fix, it will with high probability included in
the next release, what ever the number is.

PR #6049, opened 2017-06-16
removed part of _optcheck_fgraph and called std_fgraph instead

ReyhaneAskari:
@nouiz found a problem in ```_optcheck_fgraph``` when reviewing https://github.com/Theano/Theano/pull/6000 . This function should be the same as ```std_fgraph``` but the code was just copied and not updated properly. So this PR removes the duplicate code and just calls the ```std_fgraph``` directly.

### DIFF
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -613,11 +613,10 @@ def _optcheck_fgraph(input_specs, output_specs, accept_inplace=False):
         instances already installed.
 
     """
-    updates = [spec.update for spec in input_specs if spec.update]
     equivalence_tracker = _VariableEquivalenceTracker()
     fgraph, updates = std_fgraph(input_specs, output_specs, accept_inplace)
     fgraph.attach_feature(equivalence_tracker)
-    return fgraph, list(map(SymbolicOutput, updates)), equivalence_tracker
+    return fgraph, updates, equivalence_tracker
 
 
 class DataDestroyed():
--- a/theano/compile/debugmode.py
+++ b/theano/compile/debugmode.py
@@ -25,7 +25,7 @@ from theano.gof import (graph, utils, link, ops_with_inner_function)
 from theano.gof.link import raise_with_op
 from theano.compile.function_module import (
     FunctionMaker, Function, infer_reuse_pattern,
-    SymbolicOutput, std_fgraph)
+    std_fgraph)
 from theano.compile.mode import Mode, register_mode
 from theano.compile.ops import OutputGuard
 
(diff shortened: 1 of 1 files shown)


slefrancois:
jenkins test this

slefrancois:
jenkins test this

lamblin:
There seems to be a remaining issue with `OutputGuard` in the code tested from the documentation.

theano/compile/debugmode.py
```diff
@@ -620,37 +620,9 @@ def _optcheck_fgraph(input_specs, output_specs, accept_inplace=False):
     equivalence_tracker = _VariableEquivalenceTracker()
     fgraph = gof.fg.FunctionGraph(orig_inputs, orig_outputs,
                                   features=[equivalence_tracker])
```
nouiz: I think the way to do it is simpler. Restart from master. Then the full code for this method is:
```
def _optcheck_fgraph(...):
    fgraph, updates = std_fgraph(input_specs, output_specs, accept_inplace)
    fgraph.attach_feature(_VariableEquivalenceTracker())
    return fgraph, updates
```

ReyhaneAskari: Thanks. Changed as you asked. I thought this wouldn't work because ```st_fgraph``` had a slightly different implementation for creating the fgraph.

theano/compile/debugmode.py
```diff
@@ -613,44 +613,10 @@ def _optcheck_fgraph(input_specs, output_specs, accept_inplace=False):
         instances already installed.
 
     """
-    orig_inputs = [spec.variable for spec in input_specs]
     updates = [spec.update for spec in input_specs if spec.update]
```
nouiz: delete this line and return from this function like this:

fgraph, updates, equivalence_tracker

ReyhaneAskari: I can't remove the ```orig_inputs``` because it's used here:

```python
fgraph = gof.fg.FunctionGraph(orig_inputs, orig_outputs,
                                  features=[equivalence_tracker])

```

do you want to remove the fgraph too?

nouiz:
/home/jenkins/workspace/Theano_PR/theano/compile/debugmode.py:26:1: F401 'theano.compile.function_module.SymbolicOutput' imported but unused

ReyhaneAskari:
Thanks Fred.

ReyhaneAskari:
jenkins test this.

ReyhaneAskari:
jenkins test this.

nouiz:
thanks travis error was just a timeout that is a random time out, so I merged.

Issue #6048, opened 2017-06-16
Random failures on Travis related to convolutions


abergeron:
I did a little digging into why there seems to be a lot of new failures on travis and apparently we now skip less cases in FAST_COMPILE than before, which most probably take more time:

Before:
```
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 1, 'valid', 1) ... SKIP: No dilation implementation for basic cpu Conv3D.
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 1, 'valid', 2) ... SKIP: No dilation implementation for basic cpu Conv3D.
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 2, 'valid', 1) ... SKIP: No dilation implementation for basic cpu Conv3D.
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 2, 'valid', 2) ... SKIP: No dilation implementation for basic cpu Conv3D.
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 3, 'valid', 1) ... SKIP: No dilation implementation for basic cpu Conv3D.
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 1, 'valid', 3) ... SKIP: No dilation implementation for basic cpu Conv3D.
theano.tensor.nnet.tests.test_abstract_conv.TestCpuConv3d.test_gradinput_impossible_output_shapes((1, 1, 1, 1, 1), (1, 1, 1, 1, 1), 1, 'half', 1) ... SKIP: No dilation implementation for basic cpu Conv3D.
[...]

PR #6050, opened 2017-06-18
Scaled Exponential Linear Unit (SELU) activation

vipulraheja:
[Self-Normalizing Neural Networks](https://arxiv.org/abs/1706.02515)
 - [x] A new activation function called SELU
 - [x] Unit test

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -2463,7 +2463,7 @@ def selu(x, alpha=1):
 
     References
     ----------
-    .. [1] Klambauer, Gunter, et al.
+    .. [1] Klambauer G, Unterthiner T, Mayr A, Hochreiter S.
         "Self-Normalizing Neural Networks" <https://arxiv.org/abs/1706.02515>
     """
     alpha = 1.6732632423543772848170429916717
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -2448,7 +2448,7 @@ def elu(x, alpha=1):
     return tensor.switch(x > 0, x, alpha * tensor.expm1(x))
 
 
-def selu(x, alpha=1):
+def selu(x):
     """Compute the element-wise Scaled Exponential Linear unit.
 
     Parameters
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -2423,7 +2423,7 @@ def h_softmax(x, batch_size, n_outputs, n_classes, n_outputs_per_class,
 
 def elu(x, alpha=1):
     """
-    Compute the element-wise exponential linear activation function.
+    Compute the element-wise exponential linear activation function [2]_.
 
     .. versionadded:: 0.8.0
 
@@ -2441,7 +2441,7 @@ def elu(x, alpha=1):
 
     References
     -----
-    .. [1] Djork-Arne Clevert,  Thomas Unterthiner, Sepp Hochreiter
+    .. [2] Djork-Arne Clevert,  Thomas Unterthiner, Sepp Hochreiter
         "Fast and Accurate Deep Network Learning by
         Exponential Linear Units (ELUs)" <http://arxiv.org/abs/1511.07289>`.
     """
@@ -2449,7 +2449,7 @@ def elu(x, alpha=1):
 
 
 def selu(x):
-    """Compute the element-wise Scaled Exponential Linear unit.
+    """Compute the element-wise Scaled Exponential Linear unit [3]_.
 
     .. versionadded:: 0.9.0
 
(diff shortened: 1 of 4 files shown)


vipulraheja:
The build failures seem to be unrelated to the change, since the respective test [passed](https://travis-ci.org/Theano/Theano/jobs/244119050) in TravisCI. Seems most likely related to https://github.com/Theano/Theano/issues/6048

vipulraheja:
Looks like the changes in #6062 fixed it.

vipulraheja:
@nouiz @lamblin Could you please take a look whenever you get a chance? Would love some feedback :)

theano/tensor/nnet/nnet.py
```diff
@@ -2448,6 +2448,29 @@ def elu(x, alpha=1):
...
+    -------
+    symbolic tensor
+        Element-wise scaled exponential linear activation function applied to `x`.
+
+    References
+    ----------
+    .. [1] Klambauer G, Unterthiner T, Mayr A, Hochreiter S.
+        "Self-Normalizing Neural Networks" <https://arxiv.org/abs/1706.02515>
+    """
+    alpha = 1.6732632423543772848170429916717
```
nouiz: Here you override the alpha from the function signature. If you want this value to be the default, set it in the fct signature.

A quick look at the other framework interface make me think we should just remove the alpha parameter from the fct signature.

vipulraheja: You are absolutely right. It was indeed a residual argument. Removed it.

theano/tensor/nnet/nnet.py
```diff
@@ -2448,6 +2448,29 @@ def elu(x, alpha=1):
     return tensor.switch(x > 0, x, alpha * tensor.expm1(x))
 
 
+def selu(x):
```
nouiz: Add that in the doc in the file doc/library/tensor/nnet/nnet.txt. Add elu at the same time.

After that, I think the PR will be ready to merge.

thanks

vipulraheja: Done and done.

theano/tensor/nnet/nnet.py
```diff
@@ -2448,6 +2448,31 @@ def elu(x, alpha=1):
...
+        Tensor to compute the activation function for.
+
+    Returns
+    -------
+    symbolic tensor
+        Element-wise scaled exponential linear activation function applied to `x`.
+
+    References
+    ----------
+    .. [1] Klambauer G, Unterthiner T, Mayr A, Hochreiter S.
```
nouiz: This line see to raise an error in the doc gen:

/home/travis/build/Theano/Theano/theano/tensor/nnet/nnet.py:docstring of theano.tensor.nnet.selu:13: WARNING: Duplicate explicit target name: "1".

Maybe you can just remove the [1]?

vipulraheja: Seems like citation numbers are [global](http://www.sphinx-doc.org/en/stable/rest.html#citations) in Sphinx. Fixed them all here.

nouiz:
travis failed in python 3.3. I updated sphinx to the same version as python 2.7. Hopefully this will fix travis.

vipulraheja:
@nouiz I was able to generate the documentation properly on a new environment, (both with Sphinx v1.3.1 and v1.5.1). Hopefully this should pass. :crossed_fingers: 
screen shot 2017-07-01 at 11 24 22 pm

vipulraheja:
@nouiz Looks like a lot of libraries (particularly sphinx) don't like Python3.3. I tried it with Python2.7 and it works just fine (the issues with references also were fixed, with both versions of Sphinx - 1.3.1 and 1.5.1).

nouiz:
We are having others problems with python 3.3. We are seeing if bumping the min requirement to 3.4 would fix them. If so, this would probably also fix the problems in this PR. So I would just wait for the result of that and probably just rebase when this is done.

vipulraheja:
Sounds good. Thank you!

nouiz:
Can you rebase this PR on top of the master of Theano? This will remove the merge commit (so history is cleaner) and it should fix problem you have.

thanks

vipulraheja:
@nouiz Ran into some issues with sourcetree due to merge conflicts and ended up with duplicate commits. Could you please suggest a solution? One thing I can think of is a squashed merge.

abergeron:
Can you do a proper rebase to avoid the merge commits?  Also after a rebase, you need to do `git push -f` Even if git advises you to do a merge, don't listen to it.

abergeron:
I did a proper rebase.  Please check that all the changes you want are there.

vipulraheja:
@abergeron Sorry was sick and didn't get a chance to follow up. Thanks for taking this up. Looks good! 👍

vipulraheja:
@abergeron Thanks!

PR #6052, opened 2017-06-18
description for jacobian usage

MurugeshMarvel:
The documentation is not flexible for newbie like me,. so added some more description


lamblin:
Should be clarified (and displayed better) in #6165

PR #6057, opened 2017-06-19
Fail if output memory not allocated

lamblin:
instead of segfaulting like a barbarian.
Also add a couple of exceptions before failures.

### DIFF
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -844,7 +844,7 @@ class GpuAdvancedIncSubtensor1_dev20(GpuKernelBase, HideC,
         return super(GpuAdvancedIncSubtensor1_dev20, self).perform(node, inp, out)
 
     def c_code_cache_version(self):
-        return (11,)
+        return (12,)
 
     def c_headers(self):
         return ['<numpy_compat.h>', '<gpuarray_helper.h>',
@@ -906,8 +906,9 @@ if (GpuArray_vector_add_fast(%(out)s, %(y)s, %(ind)s, %(set_instead_of_inc)s)) {
         code = """
 /*
  * This is an atomicAdd that works for doubles since that is not provided
- * natively by cuda.
+ * natively by cuda before arch 6.0.
  */
+#if __CUDA_ARCH__ < 600
 __device__ ga_double atomicAdd(ga_double* address, ga_double val) {
     unsigned long long int* address_as_ull =
                                           (unsigned long long int*)address;
@@ -920,6 +921,7 @@ __device__ ga_double atomicAdd(ga_double* address, ga_double val) {
     } while (assumed != old);
     return __longlong_as_double(old);
 }
+#endif
 
 __device__ ga_double atomicExch(ga_double *address, ga_double val) {
     return atomicExch((unsigned long long int *)address,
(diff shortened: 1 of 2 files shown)


theano/gpuarray/subtensor.py
```diff
@@ -107,7 +107,10 @@ def c_code(self, node, name, inputs, outputs, sub):
             return """
         Py_XDECREF(%(out)s);
         %(out)s = pygpu_copy(%(inp)s, GA_ANY_ORDER);
-        if (!%(out)s) { %(fail)s }
+        if (!%(out)s) {
+            PyErr_SetString(PyExc_MemoryError, "Could not allocate copy of input.");
```
abergeron: There should already be a proper exception with a suitable message here.

theano/gpuarray/subtensor.py
```diff
@@ -733,6 +736,7 @@ def c_code(self, node, name, inputs, outputs, sub):
         if (!%(inplace)s) {
           %(out)s = theano_try_copy(%(out)s, %(x)s);
           if (%(out)s == NULL)
+            PyErr_SetString(PyExc_MemoryError, "Could not allocate copy of input.");
```
abergeron: There is already an exception.

theano/gpuarray/subtensor.py
```diff
@@ -874,6 +878,7 @@ def c_code(self, node, name, inputs, outputs, sub):
   %(out)s = theano_try_copy(%(out)s, %(x)s);
 }
 if (!%(out)s) {
+  PyErr_SetString(PyExc_MemoryError, "Could not allocate copy of input.");
```
abergeron: There is already an exception.

theano/tensor/subtensor.py
```diff
@@ -1408,6 +1408,10 @@ def c_code(self, node, name, inputs, outputs, sub):
         {
             Py_XDECREF(%(z)s);
             %(z)s = %(copy_of_x)s;
+            if (!%(z)s) {
+                PyErr_SetString(PyExc_MemoryError, "Could not allocate copy of input.");
```
abergeron: I am reasonably certain that the CPU code already sets an exception and 100% certain that the GPU code does.

theano/tensor/subtensor.py
```diff
@@ -1972,6 +1978,10 @@ def c_code(self, node, name, input_names, output_names, sub):
         {
             Py_XDECREF(%(out)s);
             %(out)s = %(copy_of_x)s;
+            if (!%(out)s) {
+                PyErr_SetString(PyExc_MemoryError, "Could not allocate copy of input.");
```
abergeron: I am reasonably certain that the CPU code already sets an exception and 100% certain that the GPU code does.

abergeron:
Weirdly, with all the messages you added, jenkins complains that there is an exception missing.

theano/gpuarray/subtensor.py
lamblin: OK.

theano/gpuarray/subtensor.py
lamblin: Oh, I'm missing curly braces there, which would explain why things get weird.
I'll fix that later.

lamblin:
Updated and squashed.

lamblin:
While I'm playing in that code, I added a fix for 6.0 arch.

Issue #5837, opened 2017-04-11
Update tests for multi-dimensional softmax


lamblin:
Update tests so that they check the work done in #5719.
We would have to:
- update existing tests to check more dimensions
- write new tests if needed

This is a separate ticket from #5712 so that development and testing are done by different people.

shawntan:
I guess most of the changes are to be done here: `theano/tensor/nnet[master] $ vim tests/test_nnet.py`?

Increasing the dimensions of the matrices passed to `utt.verify_grad`.

ReyhaneAskari:
hey @shawntan ,

> I guess most of the changes are to be done here: theano/tensor/nnet/tests/test_nnet.py?

yes, so basically you need to do 2 things. 1. Change the existing tests so that they would all pass.
 2. Write new tests to cover the new functionalities. 

Laurent has also worked on this but he was unable to finish it. You can start from his branch. But do a rebase with florian because Laurent's branch is behind.

shawntan:
`bordesf:master` is also behind, I'm trying to rebase anyway, but there's a lot of conflicts to resolve.

ReyhaneAskari:
Are you sure? There isn't any conflicts shown on Florian's branch here: https://github.com/Theano/Theano/pull/5719

Maybe it's easier to get Florian's branch and cherry-pick Laurent's commit into that. He has only made one commit.

PR #6058, opened 2017-06-20
Tests for the softmax modifications  #5837

shawntan:
Issue https://github.com/Theano/Theano/issues/5837. Not sure if this is the way to do this, but I started from @bordesf branch and rebased master. Had to resolve quite a lot of conflicts. @ReyhaneAskari


ReyhaneAskari:
It's really weird. I actually did it my self and saw the conflicts. I don't know why on github it said able to merge. 
Thanks anyway to do the job. I guess the next step would be to work on these tests and make them pass. There will be cases that the problem might be from Florian's side. So it's better to communicate often on github to resolve these issues. Also @bordesf needs to make some changes to his branch. You can wait for his changes and than start. 

@bordesf when will you be able to finish the changes?

shawntan:
Got it. Will wait.

bordesf:
@ReyhaneAskari Since, I have to replace the symbolic nodes by param types, it will requires a lot of modifications (Especially the way that softmax_op is managed). I will try to do my best this week but with the summer school coming next week, I will maybe not be able to finalize this before mid-july.

shawntan:
@ReyhaneAskari  I think I understand it now. A merge will work, but a rebase won't. The intermediate commits have conflicts with the master branch, which is why `pull --rebase` won't work. I'm not sure if the maintainers are okay with doing a merge instead of a rebase to bring in these changes though, but I'll work on this for now.

shawntan:
The other alternative is to squash florian's changes into one commit, that should avoid the rebase conflicts as well.

shawntan:
The current capability of softmax doesn't seem to be able to have optional axis. Should I be testing for this?

bordesf:
@ReyhaneAskari @shawntan I think we should have a meeting next week to speak about how we can synchronise the tests writing. As I said, I am currently trying to figure out how to include the Param_type in the best way and I am doing some profiling to see if we can get ride of some optimization or class like Softmax_with_bias. This will implies a lot of change and I am not sure that writing right now tests corresponding to my last implementation that use symbolic axes make sense since the code will change a lot in the next days. We should avoid writing test that will become useless if the main code change too much. So, I suggest to wait that I am done with reorganizing the Softmax and then we could see how to test this new implementation.

ReyhaneAskari:
@shawntan Florian, Pascal and I had a chat today and we think that the best approach is that Florian would create a new clean branch and cherry-pick the commits that are ready to merge and you add/modify tests if necessary. It's better that you push your commits to that branch too so that handling the changes would be easier.

After the merge of that branch we will still work on it to improve it.

shawntan:
@ReyhaneAskari Sounds good. @bordesf Could you ping me or reference the branch here once the branch is ready? Thanks!

nouiz:
Which PR is the continuation of this one? To have a link to it from here.

thanks

ReyhaneAskari:
Since we wanted to develop the softmax and the tests both at the same time, there is this PR : https://github.com/Theano/Theano/pull/6132 where the tests and the softmax are both in the same PR.

PR #6059, opened 2017-06-20
separate copying inputs and copying orphans

rizar:
This PR has two effects:
- `theano.gof.graph.clone_get_equiv` now has better interface with separate options for `copy_inputs` and `copy_orphans`
- `theano.gof.graph.clone` now always copies orphans, its `copy_inputs` flag will only affect the inputs, not the orphans

This makes it possible to clone a graph including the constant but preserving the inputs.

### DIFF
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -822,7 +822,7 @@ def clone(i, o, copy_inputs=True):
     return [equiv[input] for input in i], [equiv[output] for output in o]
 
 
-def clone_get_equiv(inputs, outputs, copy_inputs_and_orphans=True, memo=None):
+def clone_get_equiv(inputs, outputs, copy_inputs=True, copy_orphans=True, memo=None):
     """
     Return a dictionary that maps from Variable and Apply nodes in the
     original graph to a new node (a clone) in a new graph.
@@ -850,7 +850,7 @@ def clone_get_equiv(inputs, outputs, copy_inputs_and_orphans=True, memo=None):
 
     # clone the inputs if necessary
     for input in inputs:
-        if copy_inputs_and_orphans:
+        if copy_inputs:
             cpy = input.clone()
             cpy.owner = None
             cpy.index = None
@@ -862,7 +862,7 @@ def clone_get_equiv(inputs, outputs, copy_inputs_and_orphans=True, memo=None):
     for apply in io_toposort(inputs, outputs):
         for input in apply.inputs:
             if input not in memo:
-                if copy_inputs_and_orphans:
+                if copy_orphans:
                     cpy = input.clone()
                     memo[input] = cpy
                 else:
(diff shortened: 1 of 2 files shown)


rizar:
Ping!

rizar:
Ping!

theano/gof/graph.py
```diff
@@ -822,7 +822,7 @@ def clone(i, o, copy_inputs=True):
     return [equiv[input] for input in i], [equiv[output] for output in o]
 
 
-def clone_get_equiv(inputs, outputs, copy_inputs_and_orphans=True, memo=None):
+def clone_get_equiv(inputs, outputs, copy_inputs=True, copy_orphans=True, memo=None):
```
nouiz: In this file, there is a function that call clone_get_equiv while passing a parameter. This commit will change current Theano behavior. I don't think we want that. So I would add this diff to this PR:
```
diff --git a/theano/gof/graph.py b/theano/gof/graph.py
index d0c7652..9340c29 100644
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -810,7 +810,7 @@ def clone(i, o, copy_inputs=True):
     o : list
         Output Variables.
     copy_inputs : bool
-        If True, the inputs will be copied (defaults to True).
+        If True, the inputs and orphans will be copied (defaults to True).
 
     Returns
     -------
@@ -818,7 +818,7 @@ def clone(i, o, copy_inputs=True):
         The inputs and outputs of that copy.
 
     """
-    equiv = clone_get_equiv(i, o, copy_inputs)
+    equiv = clone_get_equiv(i, o, copy_inputs, copy_inputs)
     return [equiv[input] for input in i], [equiv[output] for output in o]
 ```

You will need to update the docstring bellow.

rizar: Haha, but changing the behaviour of `clone` was exactly what I wanted to do! It's behaviour was very weird in a sense that `copy_inputs=False` also prevented constants from being copied. Are you sure that you want to keep this behaviour? The tests pass even with this change.

rizar:
Ping!

Should I look for you at the office so that we discuss this PR offline?

abergeron:
jenkins test this please

theano/gof/graph.py
abergeron: I get what you want but it's probably safer to make a new function (maybe clone2) that does what you want and call that.

Because otherwise we may end up chasing a weird bug in six months due to that change.

rizar:
Are you asking me? How do I do it?

abergeron:
You mean the tests?  No I was asking jenkins.  It reads the comments on here.  However it won't start when there is a WIP in the title which is why this did nothing.

nouiz:
@rizar can you update this PR with the diff I gave you? This way, it will contain the function to do what you need without changing the default interface.

We can discuss the clone() interface change after the release and be sure it don't break stuff. This way for software will be able to work with the last release of Theano.

nouiz:
I updated this PR with what we discussed. @abergeron can you review?

PR #6062, opened 2017-06-20
Use FAST_RUN for reference version.

lamblin:
This should speed up the FAST_COMPILE buildbot

### DIFF
--- a/theano/tensor/nnet/tests/test_neighbours.py
+++ b/theano/tensor/nnet/tests/test_neighbours.py
@@ -258,9 +258,9 @@ class T_Images2Neibs(unittest_tools.InferShapeTester):
                     padded_x = T.set_subtensor(padded_x[:, :, extra[0]:-extra[0], extra[1]:-extra[1]], x)
                     x_using_valid = images2neibs(padded_x, neib_shape, neib_step, mode="valid")
                     x_using_half = images2neibs(x, neib_shape, neib_step, mode="half")
-                    close = T.allclose(x_using_valid, x_using_half)
-                    f = theano.function([], close, mode=self.mode)
-                    assert f()
+                    f_valid = theano.function([], x_using_valid, mode='FAST_RUN')
+                    f_half = theano.function([], x_using_half, mode=self.mode)
+                    unittest_tools.assert_allclose(f_valid(), f_half())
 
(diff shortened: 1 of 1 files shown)


abergeron:
I don't know.  It doesn't seem like either is the reference version, but more like they are testing each other.

lamblin:
"full" and "half" have been added recently, and it looks like the tests are building a more complex graph (explicit padding) to be able to use "valid".
@botev can you confirm?

botev:
Yes, the tests for "full" and "half" compare against the explicitly padded input and then running "valid". This was so that the testing is less prone to my own design/choice of tests and relies on already well-tested implementation.

lamblin:
OK, thanks. So using the C version of "valid" when testing the Python version of "full" and "half" would be reasonable in that case.

theano/tensor/nnet/tests/test_neighbours.py
```diff
@@ -283,9 +283,9 @@ def test_neibs_full_step_by_valid(self):
                     padded_x = T.set_subtensor(padded_x[:, :, extra[0]:-extra[0], extra[1]:-extra[1]], x)
                     x_using_valid = images2neibs(padded_x, neib_shape, neib_step, mode="valid")
                     x_using_full = images2neibs(x, neib_shape, neib_step, mode="full")
-                    close = T.allclose(x_using_valid, x_using_full)
-                    f = theano.function([], close, mode=self.mode)
-                    assert f()
+                    f_valid = theano.function([], x_using_valid, mode='FAST_RUN')
+                    f_full = theano.function([], x_using_full, mode=self.mode)
+                    unittest_tools.assert_allclose(f_valid(), f_full())
```
nouiz: Did you time that this make a difference in the test time? This lower a little bit the test coverage as DebugMode isn't used. But I think think this is a problem. If you see test time difference on your computer, I'm ok to merge this PR.

lamblin: The tests in question goes from around 8 minutes to around 4, I think.

lamblin: There are already tests for "valid" that are executed in DebugMode.

PR #6069, opened 2017-06-22
Add reduction based on cudnn.

abergeron:
The op also supports argmin/argmax, but I haven't done the optimization for those yet.


theano/gpuarray/dnn.py
```diff
@@ -2985,6 +3056,73 @@ def local_gpua_logsoftmax_to_dnn(op, ctx_name, inputs, outputs):
...
+    if node.inputs[0].dtype not in ['float16', 'float32', 'float64']:
+        return
+
+    if (node.inputs[0].dtype == 'float64' and
+            node.op.acc_dtype != 'float64'):
+        return
+
+    if (node.inputs[0].dtype == 'float32' and
+            node.op.acc_dtype != 'float32'):
+        return
```
lamblin: I think it means this will almost never be used.
Currently, the default `acc_dtype` for `float32` inputs on CPU is `float64`. In the old back-end, the accumulation dtype was always `float32`, because the reduction structure made it more precise.
However, now the `acc_dtype` for `GpuCAReduceCuda` set in the lifter is the CPU one, which is `float64` by default.

If it is indeed a limitation of cuDNN, we need to do one of the following:
- force `acc_dtype=float32` in `GpuCAReduceCuda` (which would be consistent with the old back-end)
- force `acc_dtype=float32` in `GpuDnnReduction`

theano/gpuarray/tests/test_elemwise.py
```diff
@@ -346,7 +348,9 @@ def setUp(self):
 
 class T_gpureduce_dtype(test_elemwise.T_reduce_dtype):
     mode = mode_with_gpu.excluding('local_cut_useless_reduce')
-    op = GpuCAReduceCuda
+
+    # GpuDnnReduction doesn't cover all cases, but should cover some
+    op = (GpuCAReduceCuda, GpuDnnReduction)
```
lamblin: I think there should be a couple of tests to check that `GpuDnnReduction` is actually inserted at least at a couple of places.

abergeron: This is done in `test_dnn.py:test_dnn_reduction_opt`

theano/gpuarray/dnn.py
abergeron: Addressed in new commit.

theano/gpuarray/opt.py
```diff
@@ -1112,10 +1112,20 @@ def local_gpua_careduce(op, context_name, inputs, outputs):
         else:
             return False
         x, = inputs
+        idtype = x.dtype
```
nouiz: You should also need an opt for the classes Argmax and MaxAndArgmax. Otherwise, it won't be used on the GPU in the end.

abergeron: Yes, I know.  This is noted in the description of this PR.

theano/gpuarray/dnn.py
lamblin: :+1:

theano/gpuarray/dnn_redux.c
```diff
@@ -0,0 +1,209 @@
...
+  if (!GpuArray_IS_C_CONTIGUOUS(&input->ga)) {
+    PyErr_SetString(PyExc_ValueError, "Only contiguous inputs are supported.");
+    return 1;
+  }
+
+  if (c_set_tensorNd(input, APPLY_SPECIFIC(input)) != 0)
+    return 1;
+
+  p = 0;
+  rsz = 1;
```
lamblin: `error: 'rsz' was not declared in this scope.`

theano/gpuarray/dnn.py
```diff
@@ -1557,6 +1561,71 @@ def make_node(self, dy, sm):
...
+            assert self.acc_dtype == 'float64'
+        if inp.dtype == 'float32':
+            assert self.acc_dtype == 'float32'
+        if inp.dtype == 'float16':
+            assert self.acc_dtype != 'float64'
+
+        bcast = []
+        for i in range(inp.ndim):
+            if not (self.axis & (1 << i)):
+                bcast.append(inp.broadcastable[i])
```
abergeron: I'll do that.

nouiz:
jenkins fail

abergeron:
Jenkins has problems, because most of these tests pass locally on my machine (I haven't tried everything).

nouiz:
flake8 errors: 
/home/jenkins/workspace/Theano_PR/theano/gpuarray/tests/test_opt.py:135:24: F821 undefined name 'GpuDnnReduction'
/home/jenkins/workspace/Theano_PR/theano/gpuarray/tests/test_opt.py:139:21: F821 undefined name 'GpuDnnReduction'

I'm killing travis as it is overloaded and you will update this PR.

abergeron:
This doesn't have any errors in my local tests anymore.

theano/gof/type.py
```diff
@@ -634,7 +634,7 @@ def c_code(self, node, name, inputs, outputs, sub):
         """ % dict(ctype=self.rtype.ctype, out=outputs[0], inp=inputs[0])
 
     def c_code_cache_version(self):
-        return (0,)
+        return (0, self.rtype.version)
```
nouiz: if self.rtype.version is None (the default), do we want to enable the cache? I think so, but I'm not 100% sure. We normally do the opposite, but I think we should still enable it. If you agree, add a comment that tell we want that here.

theano/gpuarray/dnn.py
```diff
@@ -1557,6 +1561,75 @@ def make_node(self, dy, sm):
...
+    _cop_num_outputs = 2
+
+    __props__ = ('red_op', 'axis', 'arg', 'acc_dtype', 'dtype')
+
+    params_type = ParamsType(red_op=cudnn.cudnnReduceTensorOp_t,
+                             acc_dtype=cudnn.cudnnDataType_t,
+                             c_axis=uint32_t,
+                             handle=handle_type)
+
+    def __init__(self, red_op, axis, acc_dtype, dtype, arg):
```
nouiz: Keep the order the same as props. Also document them. In particular the name arg isn't clear what this does. I would rename it return_indices. The same as the topk PR. It is more clear what it does.

theano/gpuarray/dnn.py
```diff
@@ -2985,6 +3060,73 @@ def local_gpua_logsoftmax_to_dnn(op, ctx_name, inputs, outputs):
...
+    return [out.dimshuffle(0, 2)]
+
+
+@register_opt('cudnn')
+@local_optimizer([GpuCAReduceCuda])
+def local_dnn_reduction(node):
+    if not isinstance(node.op, GpuCAReduceCuda):
+        return
+
+    if not dnn_available(node.inputs[0].type.context_name):
```
nouiz: You don't check for the number of input ndim. Currently max is 8. This isn't tested too.

theano/gof/type.py
abergeron: I would way that yes we want to.  But we also need to make it part of the key for cases it isn't None (like for cudnn).

theano/gpuarray/dnn.py
abergeron: ok

theano/gpuarray/dnn.py
abergeron: ok

theano/gpuarray/dnn.py
```diff
@@ -3090,6 +3090,9 @@ def local_dnn_reduction(node):
     if version(raises=False) < 6000:
         return
 
+    if node.inputs[0].ndim > 8:
```
nouiz: Add a test.

nouiz:
jenkins have failure. I'm killing travis.

theano/gof/type.py
nouiz: In most ops that have such inner version, we do something like:

if inner_version is None:
    return None
return (N, inner_version)

I don't think we want that. Do you agree? If so, add a comment to talk about this.

theano/gpuarray/dnn_redux.c
```diff
@@ -0,0 +1,210 @@
...
+    if (workspace == NULL) {
+      PyErr_Format(PyExc_RuntimeError, "gpudata_alloc: %s",
+                   gpucontext_error(c->ctx, e));
+      return 1;
+    }
+  }
+
+  err = cudnnReduceTensor(params->handle, APPLY_SPECIFIC(red),
+                          indices ? PyGpuArray_DEV_DATA(*indices) : NULL, indsize,
+                          worksize ? *((void **)workspace) : NULL, worksize,
```
nouiz: The condition isn't needed I think. workspace will be NULL if worksize is 0.

abergeron: Yes it will be NULL and `*((void **)workspace)` will segfault.

theano/gpuarray/dnn_redux.c
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  if (indices != NULL) {
+    if (theano_prep_output(indices, p, dims, GA_UINT, GA_C_ORDER, c) != 0)
+      return 1;
+    indsize = PyGpuArray_SIZE(*indices);
+  }
+
+  if (p == input->ga.nd || rsz == 1) {
+    *output = pygpu_reshape(input, p, dims, GA_C_ORDER, 0, -1);
```
abergeron: ok

nouiz:
I pushed a commit in this PR that fix problems with cudnn 5.1.

I agree to merge this PR when both CI pass. Do you agree with my changes?

For the argmax, you can finish that in a follow up PR.

abergeron:
I apparently can't approve a PR I started, but I'm ok with the changes.

Issue #5811, opened 2017-04-05
AbstractBatchNormTrainGrad misses `grad()`


f0k:
The `AbstractBatchNormTrainGrad` Op misses a `grad()` method. This is required for the gradient penalty in "Improved Training of Wasserstein GANs" (https://arxiv.org/abs/1704.00028).

PR #6070, opened 2017-06-23
Implement grad for AbstractBatchNormTrainGrad

gvtulder:
This adds a `grad` method for `AbstractBatchNormTrainGrad` (see #5811).


nouiz:
jenkins test this

Issue #5851, opened 2017-04-14
MatrixPinv


Entonytang:
The tensor.nlinalg.matrix_inverse(inv operation) has Grad function while MatrixPinv(pinv operation)  Object has no attribute 'grad' . So, how can we use pinv in Neural Network Training?

PR #6072, opened 2017-06-23
Gradient for matrix pseudoinverse

gvtulder:
This provides a gradient for the matrix pseudoinverse MatrixPinv (see #5851).


abergeron:
Your code looks good but the CIs are not having a good time right now and this is why they failed.

We'll try to fix them this week or next week and the tests should pass after a rerun.

nouiz:
jenkins test this

PR #6075, opened 2017-06-25
wrapper for calling non-simbolic functions

noskill:


noskill:
One of usecases - using an external simulator, which could be useful for reinforcement learning.

lamblin:
Maybe I'm missing something, but what is the advantage of this approach, as opposed to using something like `theano.as_op(func, idtype=dtype, odtype=dtype, infer_shape=(lambda...: shape))`?

Can `func` change at runtime?

noskill:
Sorry, I'll better ask next time first.

PR #6076, opened 2017-06-26
Allow fallback on GpuCorrMM if cuDNN is not there

lamblin:
Otherwise, the RuntimeError will skip the other optimizations, including the one inserting GpuCorrMM


lamblin:
Jenkins error unrelated:
```
UnboundLocalError: ('The following error happened while compiling the node', DnnVersion(), '\n', "local variable 'md5' referenced before assignment")
```

PR #6077, opened 2017-06-26
Add a DEBUG option for C code

notoraptor:
Discussed here: https://github.com/Theano/Theano/pull/5932#issuecomment-310209926

This PR adds a boolean Theano flag `cmodule.debug` (default `false`) used to activate a DEBUG mode for C code by adding a `DEBUG` macro at the top of any generated C code. The PR also uses this new option in an example to print runtime chosen algorithm for cuDNN backward filter convolution.

@lamblin @abergeron  @nouiz .

### DIFF
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -910,6 +910,12 @@ class CLinker(link.Linker):
         The support code from Variables is added before the support code from Ops.This might contain duplicates.
         """
         ret = []
+        if config.cmodule.debug:
+            ret.append("""
+            #ifndef DEBUG
+            #define DEBUG
+            #endif
+            """)
         # generic support code
         for x in [y.type for y in self.variables] + [
                 y.op for y in self.node_order]:
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -1058,6 +1058,12 @@ import theano and print the config variable, as in:
     reused by Theano. Automatic deletion of those c module 7 days
     after that time.
 
+.. attribute:: config.cmodule.debug
+
+    Bool value, default: ``False``
+
+    If True, define a DEBUG macro (if not exists) for any compiled C code.
+
 .. attribute:: config.traceback.limit
 
     Int value, default: 8
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -1139,6 +1139,11 @@ AddConfigVar('cmodule.age_thresh_use',
              IntParam(60 * 60 * 24 * 24, allow_override=False),
              in_c_key=False)
 
+AddConfigVar('cmodule.debug',
+             "If True, define a DEBUG macro (if not exists) for any compiled C code.",
+             BoolParam(False),
+             in_c_key=True)
+
 
 def default_blas_ldflags():
     global numpy
--- a/theano/gof/type.py
+++ b/theano/gof/type.py
@@ -850,19 +850,27 @@ class EnumType(Type, dict):
         int constant_3 = CONSTANT_3; // constant_3 == 0
         int constant_4 = CONSTANT_4; // constant_4 == 1
 
-    You can also specify a C type for the op param if you want to pass one of these constant values at runtime.
-    Default C type is ``double``.
+    You can also specify a C type for the op param. Default C type is ``double``.
 
     .. code-block:: python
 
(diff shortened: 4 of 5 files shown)


notoraptor:
PS: Example in action: running this script:

File `test_error_cudnn.py`
```
import theano
import numpy as np

x = theano.shared(np.ones((1, 1, 541211, 10), 'f'))
y = theano.shared(np.ones((1, 50, 541211, 1), 'f'))
z = theano.tensor.nnet.abstract_conv.conv2d_grad_wrt_weights(x, y, filter_shape=(50, 1, 3, 10), border_mode=(1, 0), filter_flip=False)
f = theano.function([], z)
f()
print('... OK')
```

Results:
```
(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> set THEANO_FLAGS=device=cuda0,floatX=float32,dnn.conv.algo_bwd_filter=time_once

(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> python ..\test_error_cudnn.py
Mapped name None to device cuda0: GeForce 840M (0000:0A:00.0)
... OK

(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> set THEANO_FLAGS=device=cuda0,floatX=float32,dnn.conv.algo_bwd_filter=time_once,cmodule.debug=true

(python2) HPPC@NOTORAPTOR-HP C:\Users\HPPC\mila\dev\git\theano
> python ..\test_error_cudnn.py
Mapped name None to device cuda0: GeForce 840M (0000:0A:00.0)
(using CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0) ... OK
```

theano/gof/cc.py
```diff
@@ -910,6 +910,12 @@ def support_code(self):
         The support code from Variables is added before the support code from Ops.This might contain duplicates.
         """
         ret = []
+        if config.cmodule.debug:
```
nouiz: I don't like having only one global state for this. But I don't see a clean/easy work around.

slefrancois:
jenkins test this

notoraptor:
Updated to fix jenkins error.

@nouiz what do you mean by "global" state ? Would you prefer a local debug option (like `config.gpuarray.debug`) ? Or a more specific macro name (liike `THEANO_DEBUG`) ?

nouiz:
the theano flag is global. You can't set it per node in the graph. But as I
told, I don't see a clean solution.

notoraptor:
@nouiz By `per node`, do you mean per each individual node, or just per Op ?

If it is per Op, I don't know how difficult it could be to implement, but maybe we could instead use `cmodule.debug` to specify an Op name to debug, and then ensure that DEBUG macro is defined only around this Op code ?

abergeron:
I think what @nouiz means is per node in the graph (or per Apply node, if you wish).

However it's not a bad thing to have a global default like this and it could be a good first step.

abergeron:
Travis was struck by the 50 minutes curse again.

notoraptor:
Yes, I just restarted erroneous tests, but it will probably fail again.

abergeron:
I am willing to ignore travis since the problems are not linked to this PR.

nouiz:
I think by merging this now you just caused all the master commit to now fail travis this. If this is the case, this is pretty bad, as now all PR will probably fail...

Why this happen? From memory, this PR cause all the c code to be recompiled. So I think we needed to fix this before merging this PR... this PR try to fix the last problem: https://github.com/Theano/Theano/pull/6089

abergeron:
This PR passed almost all the travis tests expect one so I don't think it is that bad.

nouiz:
If that run deterministacally fail, it is bad. It isn't the timeout bug. It
fail at a different place... The failure look more like the other PR that
cause all c code to be recompiled. So I currently suspect this is the
failure of that case. So the master and all new PR will start to fail that
run of travis until it is fixed in the master (unless we revert).

As we areclose tofix it in the master, we can wait a few hours before
reverting.

nouiz:
I mean, it isn't the random timeout bug. it fail at a different place. The
random timeout bug was always failing at the same place.

PR #6079, opened 2017-06-27
[Intel MKL] implement high performance convolution OP

pengzhao-intel:
This PR is for the implementations of Intel MKL-based Convolution OP, regarding our previous discussion in [#5580](https://github.com/Theano/Theano/issues/5580) and [PR5770](https://github.com/Theano/Theano/pull/5770) and [PR6016](https://github.com/Theano/Theano/pull/6016).

In this PR, both data layout conversion and convolution computation OPs are included.
BTW, the MKLNdarray code is duplicated so we can continuously track it in 5770 separately.

**Data layout conversion OPs (in file basic_ops.py):**
- BaseConvertOp(MKLOp): base layer for all data conversion OP
- U2IConv(BaseConvertOp): change the user layout (numpy array) to MKLNdarray for convolution layers
- I2U(BaseConvertOp): change the MKLNdarray to user layout (numpy array) and this OP can work for any MKL outputs (such as convolution, relu, pool)  
- U2IGrad(BaseConvertOp): it's for backward computation of U2I and actually it's equal with I2U in functionality
- I2UGrad(BaseConvertOp): MKLNdarry to numpy layout translation for backward computation

**Convolution OPs (in file mkl_conv.py):**
- MKLConvBase(gof.Op):  base layer for convolution including common variables definitions
- Conv2D(MKLConvBase): forward path of convolution with MKL API
- ConvGradInputs(MKLConvBase): gradient input path of convolution with MKL API
- ConvGradWeights(MKLConvBase): gradient weight path of convolution with MKL API
- AbstractConvGroup(gof.Op): grouped convolution and there may be conflict with [PR5991](https://github.com/Theano/Theano/pull/5991) and we will take look for details
- AbstractConvGroupGrad(gof.Op): gradient computation for grouped convolution

**Tests (in file tests/test_conv.py)**
- test_conv_U2I(self):
- test_conv_no_bias(self):
- test_conv_with_bias(self):
- test_no_shape(self):
- test_conv_no_bias(self):
- test_conv_with_bias(self):

This is only a high-level description about our new Ops and feels free to raise any questions and suggestions for us.

The related link:
https://software.intel.com/en-us/mkl-developer-reference-c-dnn-operations

### DIFF
--- a/theano/contrib/mkl/basic_ops.py
+++ b/theano/contrib/mkl/basic_ops.py
@@ -161,7 +161,7 @@ class U2IGrad(BaseConvertOp):
                                                      z_size,
                                                      z_stride), err);
 
-            if (!dnnLayoutCompare_%(precision)s(MKLNdarray_LAYOUT(%(gz)s),layout_user) {
+            if (!dnnLayoutCompare_%(precision)s(MKLNdarray_LAYOUT(%(gz)s), layout_user) {
                 CHECK_ERR( dnnConversionCreate_%(precision)s(&from_internal,
                                                              MKLNdarray_LAYOUT(%(gz)s),
                                                              layout_user), err);
@@ -256,10 +256,11 @@ class I2UGrad(BaseConvertOp):
             gz_stride[3] = gz_size[0] * gz_size[1] * gz_size[2];
 
             CHECK_ERR( dnnLayoutCreate_%(precision)s(&layout_user,
-                                                     DIMENSION, gz_size,
+                                                     DIMENSION,
+                                                     gz_size,
                                                      gz_stride), err);
 
-            if (!dnnLayoutCompare_%(precision)s(MKLNdarray_LAYOUT(%(z)s),layout_user)) {
+            if (!dnnLayoutCompare_%(precision)s(MKLNdarray_LAYOUT(%(z)s), layout_user)) {
                 CHECK_ERR( dnnConversionCreate_%(precision)s(&to_internal,
                                                              layout_user,
                                                              MKLNdarray_LAYOUT(%(z)s)), err);
@@ -273,7 +274,7 @@ class I2UGrad(BaseConvertOp):
                                                           PyArray_DATA(%(gz)s),
                                                           MKLNdarray_DATA(%(z)s)), err);
         } else {
-            memcpy((void*)MKLNdarray_DATA(%(z)s),(void*)PyArray_DATA(%(gz)s),%(z)s->data_size);
+            memcpy((void*)MKLNdarray_DATA(%(z)s),(void*)PyArray_DATA(%(gz)s), %(z)s->data_size);
         }
         first_run = 0;
         """ % sub
--- a/theano/contrib/mkl/basic_ops.py
+++ b/theano/contrib/mkl/basic_ops.py
@@ -1,0 +1,470 @@
+import theano.tensor as T
+from theano.gof import Apply, Op
+from theano.tensor.blas import ldflags
+from theano.tensor.nnet.abstract_conv import get_conv_output_shape
+from theano.contrib.mkl.mkl_helper import header_text
+from theano.contrib.mkl.mkl_type import MKLNdarrayType
+
+
+class MKLOp(Op):
+    def c_lib_dirs(self):
+        return ldflags(libs=False, libs_dir=True)
+
+    def c_libraries(self):
+        return ldflags()
+
+    def c_compile_args(self):
+        compile_args = ldflags(libs=False, flags=True)
+        compile_args += super(MKLOp, self).c_compile_args()
+        return compile_args
+
+    def c_support_code(self):
+        ccode = header_text()
+        ccode += """
+        #define DIMENSION  4
+
+        #define CHECK_ERR(f, err) \\
+            do { \\
+                (err) = (f); \\
+                if ((err) != E_SUCCESS) { \\
+                    printf("Error in file [%s:%d], err code (%d)", \\
+                           __FILE__, __LINE__, err); \\
+                    exit(1); \\
+                } \\
+            } while(0)
+        """
+
+        return ccode
+
+
+class BaseConvertOp(MKLOp):
+    def c_support_code_struct(self, node, name):
+        ccode = """
+        dnnError_t err;
+        int first_run;
+        void* internal_buf;
+        void* user_buf;
+        dnnLayout_t layout_internal;
+        dnnLayout_t layout_user;
+        dnnPrimitive_t to_internal;
+        dnnPrimitive_t from_internal;
+        dnnPrimitive_t primitive;
+        void *convert_resources[dnnResourceNumber];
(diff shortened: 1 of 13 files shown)


theano/contrib/mkl/basic_ops.py
```diff
@@ -148,38 +144,40 @@ def c_code(self, node, name, inp, out, sub):
...
 
-        ndim = (int)MKLNdarray_NDIM(%(gz)s);
-        assert (ndim == DIMENSION);
-
-        for(int i=0; i<DIMENSION; i++) {
-            z_size[i] = (size_t)PyArray_DIMS(%(z)s)[ndim-i-1];
-            z_stride[i] = (size_t)PyArray_STRIDES(%(z)s)[ndim-i-1] / %(x_item_size)s;
-        }
-
         //create usr layerout
```
nouiz: typo: usr ->user

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+class MKLOp(Op):
+    def c_lib_dirs(self):
+        return ldflags(libs=False, libs_dir=True)
+
+    def c_libraries(self):
+        return ldflags()
+
+    def c_compile_args(self):
+        compile_args = ldflags(libs=False, flags=True)
+        compile_args += super(MKLOp, self).c_compile_args()
```
nouiz: This line will cause an error to be raised an that will cause this method to be ignored. I think you should just delete this line.

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+        ccode += """
+        #define DIMENSION  4
+
+        #define CHECK_ERR(f, err) \\
+            do { \\
+                (err) = (f); \\
+                if ((err) != E_SUCCESS) { \\
+                    printf("Error in file [%s:%d], err code (%d)", \\
+                           __FILE__, __LINE__, err); \\
+                    exit(1); \\
```
nouiz: We should not call exit(). This will cause the python process to stop.

You should set a python error (PyErr_Format, or PyErr_SetString), then follow python C convention to have that error raised in the end.

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+        from_internal = NULL;
+        primitive = NULL;
+        """
+        return ccode
+
+    def c_code_cache_version(self):
+        return (1, 0)
+
+
+class I2U(BaseConvertOp):
```
nouiz: The name isn't clear. Can you find/suggest a better class name? MklToNdarray?

Also add a class docstring.

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def make_node(self, x):
+        assert isinstance(x.type, MKLNdarrayType)
+        return Apply(self, [x], [T.TensorType(broadcastable=x.broadcastable, dtype=x.dtype)()])
+
+    def grad(self, inp, grads):
+        x, = inp
+        gz, = grads
+
+        return [I2UGrad()(x, gz)]
```
nouiz: It is a bad practice to keep x just for its shape. This will force it to be kept for the gradient instead of being freed. This will cause higher memory peak usage.

I think you need to create only 2 class: U2I and I2U. Each are the grad of the other.

Can you infer the shape from an MKLNdarray of the corresponding Ndarray? If not, you could make one of the op take 2 inputs, the data and the shape. This will only kept the shape in the graph and the data itself.

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+                                                          MKLNdarray_DATA(%(z)s)), err);
+        } else {
+            memcpy((void*)MKLNdarray_DATA(%(z)s),(void*)PyArray_DATA(%(gz)s), %(z)s->data_size);
+        }
+        first_run = 0;
+        """ % sub
+        return ccode
+
+
+class U2IConv(BaseConvertOp):
```
nouiz: I don't understand what this op does. Can you explain?

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+        first_run = 0;
+        """ % sub
+        return ccode
+
+
+class I2UGrad(BaseConvertOp):
+    __props__ = ()
+
+    def make_node(self, x, gz):
+        out = x.type()
```
nouiz: In all the make_node() call `x=as_tensor_variable(x)` for all variable that are numpy ndarray. Do the equivalent for your own variable.

theano/contrib/mkl/basic_ops.py
```diff
@@ -0,0 +1,472 @@
...
+                                                          PyArray_DATA(%(z)s)), err);
+        } else {
+            memcpy ((void*)PyArray_DATA(%(z)s), MKLNdarray_DATA(%(gz)s), PyArray_SIZE(%(z)s)*PyArray_ITEMSIZE(%(z)s));
+        }
+        first_run = 0;
+        """ % sub
+        return ccode
+
+
+class I2UGrad(BaseConvertOp):
```
nouiz: In the PR comments, you told it is equivalent to I2U. So why do need this one? What is the difference? Can you merge the 2 ops?

Here, gz should have the same size as x. So you should be able to have only 1 inputs to this op.

theano/contrib/mkl/mkl_conv.py
```diff
@@ -0,0 +1,1613 @@
...
+
+class Conv2D(MKLConvBase):
+    __props__ = ('imshp', 'kshp', 'border_mode', 'subsample', 'filter_flip', 'filter_dilation')
+
+    def __init__(self, imshp=None, kshp=None, border_mode='valid', subsample=(1, 1), filter_flip=False, filter_dilation=(1, 1)):
+        super(Conv2D, self).__init__(imshp=imshp, kshp=kshp, border_mode=border_mode, subsample=subsample)
+        self.filter_flip = filter_flip
+        self.filter_dilation = filter_dilation
+
+    def c_cleanup_code_struct(self, node, name):
```
nouiz: should be in the parent class MKLConvBase I think.

pengzhao-intel:
@nouiz Thanks for your comments. I will provide more details and our design mindset based on your questions. 
We'd like to align with the principle of  Theano Ops and apply these rules into the other MKL Ops (Relu, Pool, LRN, etc).

theano/contrib/mkl/basic_ops.py
pengzhao-intel: Good suggestion, I will delete this line to avoid the error.

theano/contrib/mkl/basic_ops.py
pengzhao-intel: Good point. Will change.

PR #6081, opened 2017-06-27
Params for Gemm, CGer, CGemv

nouiz:
They seem to slow down tests with empty cache.

### DIFF
--- a/theano/tensor/signal/pool.py
+++ b/theano/tensor/signal/pool.py
@@ -1092,6 +1092,8 @@ class PoolGrad(OpenMPOp):
 
 
 class MaxPoolGrad(PoolGrad):
+    # params_type ignore_border don't change c code
+
     def __init__(self, ignore_border, ndim=2, openmp=None):
         PoolGrad.__init__(self, ignore_border, mode='max', ndim=ndim, openmp=openmp)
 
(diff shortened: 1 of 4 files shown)


nouiz:
I did not change Ger or Gemv, but CGer and CGemv, the subclasses. So I don't see why I would need to change the parent classes.

theano/tensor/signal/pool.py
```diff
@@ -2369,13 +2378,13 @@ def c_code(self, node, name, inp, out, sub):
...
                   r_end[%(i)s] -= pd[%(i)s];
                   // handle the case where no padding, ignore border is True
-                  if (%(ignore_border)s)
+                  if (%(params)s->ignore_border)
                   {
                     r_end[%(i)s] = r_end[%(i)s] > r[%(i)s] ? r[%(i)s] : r_end[%(i)s];
                   }
                   // use the index to find the correct position in the output
                   o_idx[%(non_pool_ndim)s + %(i)s] = r_idx[%(i)s];
-            """ % dict(i=i, ignore_border=ignore_border, non_pool_ndim=non_pool_ndim)
```
notoraptor: Missing `params=sub['params']` for this dict, as `%(params)s->ignore_border` is used in the formatted string.

nouiz: thanks. done.

PR #6082, opened 2017-06-27
Buildbot testsuite names

slefrancois:
Change buildbot scripts with more detailed test suite names, to distinguish in jenkins pipelines.

[New pipeline buildbot here](http://earlgrey.iro.umontreal.ca:8080/job/Theano_buildbot/)


nouiz:
travis is a random 10m test timeout. That test take only 8s with full cache here or 15s with empty cache. So merging.

PR #4198, opened 2016-03-09
Added the polygamma function, implemented digamma gradient

enalisnick:
Added the polygamma function to basic_scipy.py, which allowed the
gradient of Psi() (the digamma function) to be implemented.

Tested with theano/misc/do_nightly_build:

Executing tests with mode=FAST_COMPILE
THEANO_FLAGS=on_shape_error=raise,on_opt_error=raise,,warn.argmax_pushdo
wn_bug=False,warn.gpusum_01_011_0111_bug=False,warn.sum_sum_bug=False,wa
rn.sum_div_dimshuffle_bug=False,warn.subtensor_merge_bug=False,,device=c
pu,floatX=float64,mode=FAST_COMPILE
/Users/enalisnick/GitHub/Theano/theano/../bin/theano-nose
/Users/enalisnick/GitHub/Theano/theano/misc/pycuda_init.py:34:
UserWarning: PyCUDA import failed in theano.misc.pycuda_init
warnings.warn("PyCUDA import failed in theano.misc.pycuda_init")
SS..S...SSS
----------------------------------------------------------------------
Ran 11 tests in 0.050s

OK (SKIP=6)
Number of elements in the compiledir:
139

Issue #6080, opened 2017-06-27
Digamma function always executed in double precision on GPU?


botev:
I'm not too competent on the topic, so was wondering if someone can tell me that my assumption is correct, given that in the `c_code` of the function in Theano it is defined as:

`DEVICE double _psi(ga_double x)`

How it is possible to make the code generic and potentially run for `float32`. 
Also, is there a good reason not to do that etc...

abergeron:
It is true that the code will always run in double the way it's written.  We might want to make the function a template so that it can use float32 for faster speed also.  However the constants in the code are computed for float64 and I don't how portable that would be to float32.

botev:
so at this stage, first I'll ask how exactly is the function C code inserted when compiling a composite Elementwise operation or a CUDA kernel? Is it as a separate function, which can get the `dtype` as a template or is it something that needs `%(dtype)s`as per some of other code? I just have no idea how this conversion goes to have some better perspective on the problem. 

As from the code of the two operators, it seems that the most unstable computations are of the form `1/x`, `1/(x+1)` ... and `1/x^2`, `1/(x+1)^2`... where these are turned off when `x` is less than a predefined value. As far as numerical stability is concerned the only issue is when the `x` is really small, thus why the computation is truncated by default. One might need to define potentially a slightly higher truncation threshold for lower precision types, but other than that I don't see why the rest of the computation can not be done in other types.

PR #6083, opened 2017-06-28
Added the tri-gamma function.

botev:
I've added the Tri-gamma function, which can be run on the GPU as well in comparison to #5963.

Note that I'm not too sure how you guys want to keep the name as the digamma function in Theano is named psi. 
As a note I'm aware of some efforts of doing the more general poly-gamma function, however, this requires a lot more effort and debugging of CUDA kernels as in #4198 which I can not do.
The implementation C code was taken from http://people.sc.fsu.edu/~jburkardt/cpp_src/asa121/asa121.html

As a side note if the `polygamma` makes it into to Theano you might want for `k=0` and `k=1` the Theano interface to return the `Psi` and `TriGamma` Ops, this might simplify the interface significantly.

### DIFF
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -298,8 +298,18 @@ class Psi(UnaryScalarOp):
         else:
             super(Psi, self).impl(x)
 
-    def grad(self, inputs, outputs_gradients):
-        raise NotImplementedError()
+    def L_op(self, inputs, outputs, grads):
+        x, = inputs
+        gz, = grads
+        if x.type in complex_types:
+            raise NotImplementedError()
+        if outputs[0].type in discrete_types:
+            if x.type in discrete_types:
+                return [x.zeros_like(dtype=theano.config.floatX)]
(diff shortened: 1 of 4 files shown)


abergeron:
For the naming issues, we try to follow what numpy/scipy is doing.  If the names differ then we might want to rename and deprecate the old name.

botev:
Hmm, so numpy/scipy do not have explicitely tri-gamma function, but only the polygamma. 
On the other hand for the digamma there is both `scipy.special.digamma` and `scipy.special.psi`. 
I think that the tri-gamma is a good enough name.
I actually had a question here - #6080 - is that all of the implementation as they borrow code from different C++ implementations they use double precision variables and my guess is that they do the same on the GPU. I don't know if this is an issue or should the function somehow be able to deal with "generic" types?

abergeron:
You're right about always computing in float64.  But since all the other ops with a specialized implementation do that for now, I'll merge as-is.  Follow-up on that can be done on #6080

PR #6085, opened 2017-06-29
Add deterministic={default,more} flag

nouiz:


lamblin:
I'm mixed on this one.
I'm afraid it would build expectations for the users, that we cannot meet at the moment (for instance, we sometimes use non-deterministic cuDNN algorithms as a fall-back, even if the requested one was deterministic).
Also, I'm not sure how the interface for the flag would work if we add more deterministic flags later. 'default', 'more', 'even more'?

On the other hand, it _is_ easier to type `deterministic=more` than trying to disable insertion of _dev20 by hand.

nouiz:
I think we absolutly need an easy way to allow to remove _dev20 from the
graph.

If you don't find other idea, I would merge this PR now. It is waiting for
too long.

For the futur, I would continue to use more for GPU ops that have 2
implementations. The other option we could set is to "full" or "cpu" that
would prefer the CPU ops when there is no GPU ops deterministic.

For cudnn fallback, I think we should fix the fall-back to fall-back to a
deterministic algo if the user specified a deterministic algo. I think it
would be easy to do. If no deterministic algo, I would print a warning and
select a non-deterministic option.

PR #6086, opened 2017-06-29
Doc that people can use the dev version of Theano.

nouiz:


nouiz:
This need a full rewrite. We have also our conda channel that will help with this for dev version.

PR #6087, opened 2017-06-29
add windows buildbot

slefrancois:
Pending [test run on jenkins](http://earlgrey.iro.umontreal.ca:8080/job/Theano_buildbot_win/50/).

### DIFF
--- a/.jenkins/jenkins_buildbot_windows.bat
+++ b/.jenkins/jenkins_buildbot_windows.bat
@@ -52,5 +52,5 @@ echo "Directory of stdout/stderr %BUILDBOT_DIR%"
 REM Fast run and float32
 set FILE=%BUILDBOT_DIR%\theano_python2_fastrun_f32_tests.xml
 set NAME=fastrun_f32
-set THEANO_FLAGS=%THEANO_FLAGS%,compiledir=%COMPILEDIR%,mode=FAST_RUN,warn.ignore_bug_before=all,on_opt_error=raise,on_shape_error=raise,magma.enabled=true,floatX=float32,dnn.include_path=C:\\lib\\cuda\\include,dnn.library_path=C:\\lib\\cuda\\lib\\x64,gcc.cxxflags='-I"C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v8.0\\include" -I"C:\\lib\\libgpuarray\\src" -L"C:\\Program Files\\NVIDIA GPU Computing Toolkit\CUDA\\v8.0\\lib\\x64" -LC:\\lib\\libgpuarray\\lib'
+set THEANO_FLAGS=%THEANO_FLAGS%,compiledir=%COMPILEDIR%,mode=FAST_RUN,warn.ignore_bug_before=all,on_opt_error=raise,on_shape_error=raise,floatX=float32,dnn.include_path=C:\\lib\\cuda\\include,dnn.library_path=C:\\lib\\cuda\\lib\\x64,gcc.cxxflags='-I"C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v8.0\\include" -I"C:\\lib\\libgpuarray\\src" -L"C:\\Program Files\\NVIDIA GPU Computing Toolkit\CUDA\\v8.0\\lib\\x64" -LC:\\lib\\libgpuarray\\lib'
 python bin\theano-nose %THEANO_PARAM% %XUNIT%%FILE% %SUITE%%NAME%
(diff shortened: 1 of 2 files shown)


slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
Ready to merge

.jenkins/jenkins_buildbot_windows.bat
```diff
@@ -0,0 +1,56 @@
...
+
+REM Make fresh clones of libgpuarray (with no history since we dont need it)
+rmdir libgpuarray /s/q
+git clone --depth 1 "https://github.com/Theano/libgpuarray.git"
+
+REM Clean up previous installs (to make sure no old files are left)
+rmdir %LIBDIR% /s/q
+mkdir %LIBDIR%
+
+REM Build libgpuarray
```
lamblin: Wait, does that work?
According to @abergeron it was hell to make work, unless we have access to the unholy magic at the disposal of the Conda people.

abergeron: The windows compiler for python 2.7 is a million times better now and this is actually rather easy now.

The only thing would be to validate that it does actually build with the right compiler.

abergeron:
The jenkins tests don't look like they built on windows.  Is there a magic command?

abergeron:
Nevermind in found the windows logs.  However it looks like the latest build failed because magma is missing and and unpickle failed to find `type`.

slefrancois:
I used miniconda + m2w64-toolchain for the Theano C compiler and the Microsoft Visual C++ Compiler for Python 2.7 to compile libgpuarray. With that it worked without too much trouble, besides having to manage library and header paths manually.

Since magma is an optional dependancy and the pickle type error is minor, I suggest we merge and solve those on the go.

.jenkins/jenkins_buildbot_windows.bat
notoraptor: About that, @abergeron , I had never found this new version.

Visual C++ 2017 comes with many Python-related packages, but I didn't find any clear documentation about how to use it to compile Python modules. On my computer I still use the old stuffs for Python 2.7.

nouiz:
Having buildbot fail on each PR is bad. We should make sure it pass before
enabling by default. We could fix the problem or skip those tests.

lamblin:
Missing magma should raise SkipTest.
Some pickle tests are already skipped on Windows.

slefrancois:
Having magma.enabled=true raises errors if magma is not installed. I can remove it for now, or keep it as a reminder to correct.

slefrancois:
For the missing types, any idea?

slefrancois:
I disable magma test, @notoraptor mentioned it only compile well under linux.

notoraptor:
PS: Sorry, to be more precise: I am just not able to compile it for Windows currently, but maybe it is possible, under some conditions/customizations ^^ !

I will investigate it later, but indeed, we should deactivate magma testing on Windows for the moment.

.jenkins/jenkins_buildbot_windows.bat
abergeron: What I mean by the new python 2.7 compiler is the "Visual C++ for Python 2.7" package that is now available from the Microsoft website.

They updated it so that it's not just a badly maintained portion of VC++ 2008 anymore and it doesn't have any of the problems that I encountered before.

VC++ 2017 will not work for python 2.7 modules (but it should for python 3.5).

slefrancois: I installed [Microsoft Visual C++ Compiler for Python 2.7
](https://www.microsoft.com/en-ca/download/details.aspx?id=44266)

notoraptor: Ok ! That's what I called "old stuff" ^^ ! But indeed, it works fine (after 2-3 tweaks also, from memory).

slefrancois:
jenkins test this

nouiz:
I killed travis

slefrancois:
gpuarray/test_type was loading a text pickle in binary mode, Windows didn't like it.

lamblin:
Right, depending on the protocol used when pickling, the output is either text or binary, and windows cares about the difference (I think python 3 cares on unix as well).

Is there anything else, or is this PR ready to merge^

.jenkins/jenkins_buildbot_windows.bat
abergeron: If it's the right package, it shouldn't require any tweaks besides setting up the environment with vcvarsall.bat.

PR #6088, opened 2017-06-29
Fix opt crash in float16 and enable C code for MaxAndArgmax and Argmax

nouiz:


theano/tensor/basic.py
```diff
@@ -1427,6 +1428,7 @@ class Argmax(Op):
     nout = 1
     E_axis = 'invalid axis'
     __props__ = ()
+    _f16_ok = True
```
lamblin: Does that enable C code for float16 on the CPU?
If so, is it expected to work?

nouiz: yes, we just call numpy for the max and argmax. I added tests and numpy handle it well.

theano/tensor/tests/test_basic.py
```diff
@@ -3088,6 +3091,20 @@ def test2(self):
             v_shape = eval_outputs(max_and_argmax(n, axis)[0].shape)
             assert tuple(v_shape) == np.max(data, np_axis).shape
 
+    def test2_float16(self):
+        data = rand(2, 3).astype("float16")
```
lamblin: We should at least test for negative numbers, and a couple of large ones, so make sure we are not just taking the argmax of the int16 representation of the float16 number.

nouiz: done

lamblin: Does this check fail if we view the array as `int16`?

theano/tensor/basic.py
lamblin: You mean, the C code of numpy, right?

nouiz:
Yes

Le jeu. 29 juin 2017 17:56, Pascal Lamblin <notifications@github.com> a
écrit :

Issue #4507, opened 2016-05-13
drop Python 2.6 and 3.3 support


kmike:
What do you think about dropping Python 2.6 support in Theano, and starting to run tests in Python 2.7 by default? Numpy will drop Python 2.6 support soon (https://github.com/numpy/numpy/blob/master/doc/release/1.12.0-notes.rst), and Python 2.6 is not supported by Python core team for years - no even security support.

nouiz:
Personally, I think we should do it.

But we should ask on theano-user and theano-dev about this. Do you want to
do this?

abergeron:
Just to note that numpy 1.12 also drops support for python 3.3 which we
support.

2016-05-13 8:30 GMT-04:00 Frédéric Bastien <notifications@github.com>:

nouiz:
It seem nobody objected. I would wait to make Theano 0.9 still support them, but remove the support just after 0.9 release. We will try to make it this summer. It will officially deprecate the old back-end. So we don't want to wait to long for it.

nouiz:
See another reason to drop python 2.6 support:

https://github.com/Theano/Theano/issues/4685 one package got updated and that broke travis. I'm fixing this for pydot by fixing the version in https://github.com/Theano/Theano/pull/4701, so if pydot drop support for it, travis won't break again.

johnyf:
Upstream `pydot` fixed https://github.com/erocarrera/pydot/issues/123 and is again compatible with python 2.6. However, support for python 2.6 will not be actively maintained.

nouiz:
Python 2.6 was dropped.

nouiz:
NumPy 1.12 that was just released dropped Python 3.3. We should do it not for this release 0.9, but the next one 0.10

PR #6089, opened 2017-06-29
Time all test in travis.

nouiz:
I think we should always keep that. I think it don't really change the tests time and will help to speed up the tests to always have it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
 language: python
 python:
   - "2.7"
-  - "3.3"
+  - "3.4"
 
 # NB:
 # In before_install and install sections below,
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -82,7 +82,7 @@ script:
   - cd -; cd Theano
   - python -c 'import theano; print(theano.config.__str__(print_doc=False))'
   - python -c 'import theano; assert(theano.config.blas.ldflags != "")'
-  - theano-nose -v $PART
+  - theano-nose -v --with-timer --timer-top-n 10 $PART
```
abergeron: Why not keep all the timings?

abergeron:
I'm ok with always using nose-timer.  I'm not so sure about the cProfile or theano profiling part.  Mostly because they add output and I really dislike chatty tests.

nouiz:
I'll remove the theano profiler and cprofiler I added. They are in this
branch now to help identify the problem.

Having a too long output isn't great. I'm ok to changing the size from 10
to something else, but I don't like having it too big like 100.

Do you think 20 would be better then 10? 10 have been enough to help me
debug the slowness. We can increase it later if needed.

nouiz:
issue https://github.com/Theano/Theano/issues/4507 is about droping python 3.3

PR #6091, opened 2017-06-30
Params for other ops

notoraptor:
Continues #5866 (cannot reopen that PR). Rebased and re-read. Adding op params for:
- GpuCumOp
- GPUA_mrg_uniform
- AdvancedIncSubtensor1 (+ set `check_input=False` for AdvancedSubtensor1)
- GpuAdvancedIncSubtensor1
- GpuAdvancedIncSubtensor1_dev20
- Images2Neibs
- GpuImages2Neibs
- compile.ops.Shape_i

I think it will not speed up TRAVIS so much, as changes are mainly in gpuarray module, but now we can end with this big old PR!

@nouiz @lamblin @abergeron

### DIFF
--- a/theano/compile/ops.py
+++ b/theano/compile/ops.py
@@ -346,6 +346,18 @@ class Shape_i(gof.Op):
         i = int(i)
         self.i = i
 
+    # NB:
+    # 1) params_type is defined as a property to avoid
+    #    loop in Python import caused by importing theano.scalar below
+    #    when params_type is defined directly in class code.
+    # 2) We wrap scalar into ParamsType (instead of directly using scalar as op param)
+    #    to avoid Theano converting scalar param to constant that would be later
+    #    hardcoded as litteral in C code, making us loose all the advantages of
+    #    using params.
+    @property
+    def params_type(self):
+        return gof.ParamsType(i=theano.scalar.basic.int64)
+
     def __str__(self):
         return '%s{%i}' % (self.__class__.__name__, self.i)
 
@@ -360,7 +372,7 @@ class Shape_i(gof.Op):
                             (x, self.i))
         return theano.Apply(self, [x], [theano.tensor.lscalar()])
 
-    def perform(self, node, inp, out_):
+    def perform(self, node, inp, out_, params):
         x, = inp
         out, = out_
         if out[0] is None:
@@ -383,7 +395,7 @@ class Shape_i(gof.Op):
             version.append((str(t), v))
 
         if version:
-            version.append(1)
+            version.append(2)
 
         return tuple(version)
 
@@ -391,7 +403,8 @@ class Shape_i(gof.Op):
         iname, = inames
         oname, = onames
         fail = sub['fail']
-        i = self.i
+        # i is then 'params->i', not just 'params'.
+        i = sub['params'] + '->i'
 
         itype = node.inputs[0].type.__class__
         if itype in self.c_code_and_version:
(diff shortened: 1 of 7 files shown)


theano/compile/ops.py
```diff
@@ -346,6 +346,18 @@ def __init__(self, i):
...
         self.i = i
 
+    # NB:
+    # 1) params_type is defined as a property to avoid
+    #    loop in Python import caused by importing theano.scalar below
+    #    when params_type is defined directly in class code.
+    # 2) We wrap scalar into ParamsType (instead of directly using scalar as op param)
+    #    to avoid Theano converting scalar param to constant that would be later
+    #    hardcoded as litteral in C code, making us loose all the advantages of
+    #    using params.
```
nouiz: Why it get converted to constant? Maybe that should be fixed. Not mandatory for this PR.

theano/tensor/subtensor.py
```diff
@@ -1872,10 +1873,13 @@ class AdvancedIncSubtensor1(Op):
...
     __props__ = ('inplace', 'set_instead_of_inc')
+    check_input = False
+    params_type = ParamsType(inplace=scal.bool,
+                             set_instead_of_inc=scal.int8)
 
     def __init__(self, inplace=False, set_instead_of_inc=False):
-        self.inplace = inplace
-        self.set_instead_of_inc = set_instead_of_inc
+        self.inplace = bool(inplace)
+        self.set_instead_of_inc = int(bool(set_instead_of_inc))
```
notoraptor: Fixed.

theano/gpuarray/rng_mrg.py
```diff
@@ -168,40 +173,34 @@ def gpu_kernels(self, node, name):
...
-        size_t odims[%(ndim)s+1];
         size_t n_elements = 1;
         unsigned int n_streams;
         int must_alloc_sample = ((NULL == %(o_sample)s)
                 || !pygpu_GpuArray_Check((PyObject*)%(o_sample)s)
                 || !(%(o_sample)s->ga.flags & GA_C_CONTIGUOUS)
-                || (PyGpuArray_NDIM(%(o_sample)s) != %(ndim)s));
+                || (PyGpuArray_NDIM(%(o_sample)s) != %(params)s->ndim));
+
+        size_t* odims = (size_t*)malloc(%(params)s->ndim * sizeof(size_t));
```
notoraptor: Ok, I will test on Windows. From memory, the use of alloca() seems discouraged as there may be some compatibility issue (function is not POSIX, and is machine and compiler-dependent): http://man7.org/linux/man-pages/man3/alloca.3.html#NOTES .

theano/gpuarray/extra_ops.py
```diff
@@ -21,9 +24,11 @@ class GpuCumOp(GpuKernelBase, Op):
     """
     SUPPORTED_NDIMS = 3
     __props__ = ('axis', 'mode')
+    params_type = ParamsType(axis=scalar.int32,
+                             context=gpu_context_type)
 
     def __init__(self, axis, mode='add'):
-        self.axis = axis if axis else 0
+        self.axis = int(axis) if axis is not None else 0
```
notoraptor: Done.

theano/gpuarray/rng_mrg.py
nouiz: So I'm not sure what to do. Maybe we should keep that for another PR to don't delay this PR. We should discuss that with @lamblin and @abergeron

theano/compile/ops.py
notoraptor: Ok. For info, the related code seems to be here: https://github.com/Theano/Theano/blob/master/theano/gof/cc.py#L685 .

theano/gpuarray/rng_mrg.py
nouiz: I made an issue about that: https://github.com/Theano/Theano/issues/6094

PR #6097, opened 2017-07-01
Make sure MRG uniform in float16 do not return 0

### LINKED ISSUE #6096: mrg uniform include the lower bound, should exclude it.
see: [lasagne-users] float16 issues

Mostly the uniform generator include the lower bound 0, while it should exclude it.

This cause the normal generator to do log(0), so it return an inf.

lamblin:
Also update Python code to be consistent.

Fix #6096.


lamblin:
I am not sure what test should be performed.

theano/sandbox/rng_mrg.py
```diff
@@ -330,6 +330,7 @@ def R_op(self, inputs, eval_points):
 
 class mrg_uniform(mrg_uniform_base):
     # CPU VERSION
+    _f16_ok = True
```
lamblin: I'm not actually sure it is correct for the case where the perform should accept float16, but the c_code is explicitly disabled for float16.

nouiz:
For the test, generating a big output of size 1000x1000 and making an assert that all values are >0 allow me to reproduce the problem. My test need to be re-written.

f0k:
The fix for float16 seems legit! But aren't the float32 and float64 versions also at risk of producing zeros (and therefore NaNs after the Box-Muller transform for Gaussian samples)? To avoid breaking backwards compatibility, for the Box-Muller transform, only 0 could be mapped to 1. This would modify behaviour only in the case where it would otherwise fail otherwise (but it would add some overhead).

> For the test, generating a big output of size 1000x1000 and making an assert that all values are >0 allow me to reproduce the problem.

Probably the test should have a fixed seed to be sure (maybe you do that anyway, haven't checked).

f0k:
Ah, wait, your comment (https://github.com/Theano/Theano/issues/6096#issuecomment-312451322) indicates that the float32 and float64 versions do not produce zeros. Nevermind!

slefrancois:
jenkins test this

theano/sandbox/rng_mrg.py
abergeron: _f16_ok is only for C code,  perform is always allowed.

PR #6101, opened 2017-07-03
Bump min Python version to 3.4 (fix slowness travis in Python 3.3)

nouiz:
Also enable nosetests profile of tests

fix https://github.com/Theano/Theano/issues/4507

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,6 @@ Operating System :: MacOS
 Programming Language :: Python :: 2
 Programming Language :: Python :: 2.7
 Programming Language :: Python :: 3
-Programming Language :: Python :: 3.3
 Programming Language :: Python :: 3.4
 Programming Language :: Python :: 3.5
 """
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
 language: python
 python:
   - "2.7"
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -33,7 +33,6 @@
 Programming Language :: Python :: 2
 Programming Language :: Python :: 2.7
 Programming Language :: Python :: 3
-Programming Language :: Python :: 3.3
 Programming Language :: Python :: 3.4
 Programming Language :: Python :: 3.5
 """
```
abergeron: You might want to add 3.6 to the list.

nouiz: I'm not sure we support it well. As I think the master have travis failing, I prefer not to do that in this PR.

There was many report of problems of Python 3.6, so until I got someone report that it work on the 3 platforms, I prefer we don't do it. It could be just 1 days of testing...

abergeron: ok

lamblin: I tested it on Python 3.6, and it worked fine.

nouiz: On which OS?

nouiz:
Merging now even if travis fail, as it fix the python 3.3 timeout on the master:

https://travis-ci.org/Theano/Theano/builds/249743533

PR #6102, opened 2017-07-04
Try to fix travis. Don't recompile du to new flag when not needed.

nouiz:


lamblin:
I don't understand. From the code, it looks like the generated C code will be different depending on the value of this flag.
If it is not part of the key, and not part of any version(), how is the new C code supposed to be generated?

nouiz:
If this flag is false, we don't modify the c code. As before this flag. Bug
adding this flag in the c key change the key, so it force the recompilation
of everything. even if the c code wasn't changed. Which made travis bust
the 50m timeout.

I tried to remove the need of the recompilation with this change. But it
didn't repair travis sadly.

lamblin:
My problem is that if the old code is in the cache, changing the value of the flag will not work.
Worse, if someone compiles with that flag first and then removes it, the DEBUG code will still be executed.
I agree that the global key may not be the best place to put it. What about always having the code in `#ifdef DEBUG` and add `-DDEBUG` to the compile flags?

nouiz:
I don't understand the problem you raise.

We modify the key and the c code at the same time if the flag is there. So
if the flag change, the key always change and this will trigger a new thunk
to be used/compiled. this is part of the "hard" part of the key, not the
"soft" part.

Do you mean we should make that flag as allow_override=False? I don't see
this need, as it is used only during theano.function() and there is no way
a user can change its value between when we generate the hash of the flags
and when we use the value in the c code generation.

notoraptor:
I just noticed that Theano flag `cxxflags` does have `in_c_key` set to `False`, because "Added elsewhere in the c key only when needed".

A little surprising, but maybe should I use `cmodule.debug` the same manner as `cxxflags` ? Ie. find a strategy to add it in the Op key only when necessary. E.g.:
```python
c_code = op.c_code()
if cmodule.debug and ("#ifdef DEBUG" in c_code or "defined(DEBUG)" in c_code):
    sig.append('debug')
    # And then add `-DDEBUG` in compile args, instead of hardcoding `#define DEBUG`.
```

or something like that ?

nouiz:
Reverted in https://github.com/Theano/Theano/pull/6271 as it force recompilation and the revert make the code base simpler.

PR #6105, opened 2017-07-04
Speed up some tests

nouiz:


nouiz:
trigger with new master.

theano/tensor/extra_ops.py
```diff
@@ -311,57 +320,51 @@ def c_code(self, node, name, inames, onames, sub):
...
 
-        if self.axis is None or (self.axis == 0 and node.inputs[0].ndim == 1):
-            code = """
+        code = """
+                int axis = %(params)s->c_axis;
+                if (axis == 0 && PyArray_NDIM(%(x)s) == 1)
+                    axis = NPY_MAXDIMS;
                 npy_intp shape[1] = { PyArray_SIZE(%(x)s) };
-                if(!(%(z)s && PyArray_DIMS(%(z)s)[0] == shape[0]))
+                if(axis == NPY_MAXDIMS && !(%(z)s && PyArray_DIMS(%(z)s)[0] == shape[0]))
```
notoraptor: I think this does not check that %(z)s is a 1-D tensor (`PyArray_NDIM(%(z)s) == 1`), it only checks that z[0] has the expected shape.

theano/tensor/extra_ops.py
```diff
@@ -311,57 +320,51 @@ def c_code(self, node, name, inames, onames, sub):
...
-                       %(fail)s;
-                    }
-                    // Because PyArray_%(func)s returns a newly created reference on t.
-                    Py_XDECREF(t);
-                }
-            """ % locals()
-        else:
-            code = """
-                if(!(%(z)s && PyArray_CompareLists(PyArray_DIMS(%(z)s), PyArray_DIMS(%(x)s), PyArray_NDIM(%(x)s))))
+                else if(axis != NPY_MAXDIMS && !(%(z)s && PyArray_CompareLists(PyArray_DIMS(%(z)s), PyArray_DIMS(%(x)s), PyArray_NDIM(%(x)s))))
```
notoraptor: Here also, if z and x don't have the same number of dimensions, then some extra-memory cases may be unfortunately read by `PyArray_CompareLists()`.

**NB**: There are some functions like `theano_size_check()` and `theano_prep_output()` into new GPU backend. Should we think about implementing equivalent functions for CPU codes ?

theano/tensor/extra_ops.py
```diff
@@ -311,57 +320,51 @@ def c_code(self, node, name, inames, onames, sub):
...
                 {
 
-                    PyObject * t = PyArray_%(func)s(
-                        %(x)s, %(axis)s,
-                        PyArray_TYPE((PyArrayObject*) py_%(x)s), %(z)s);
+                    PyObject * t = NULL;
+                    if(%(params)s->mode == MODE_ADD)
+                        t = PyArray_CumSum(
+                            %(x)s, axis,
+                            PyArray_TYPE((PyArrayObject*) py_%(x)s), %(z)s);
```
notoraptor: `PyArray_TYPE((PyArrayObject*) py_%(x)s)` -> `PyArray_TYPE((x)s)` (as you do above!)

theano/tensor/extra_ops.py
```diff
@@ -311,57 +320,51 @@ def c_code(self, node, name, inames, onames, sub):
...
+                    else if(%(params)s->mode == MODE_MUL)
+                        t = PyArray_CumProd(
+                            %(x)s, axis,
+                            PyArray_TYPE((PyArrayObject*) py_%(x)s), %(z)s);
```
notoraptor: `PyArray_TYPE((PyArrayObject*) py_%(x)s)` -> `PyArray_TYPE((x)s)`

theano/tensor/extra_ops.py
nouiz: Above, like 327, I check cof the ndim==1. If it is this case, I set axis=NPY_MAXDIMS. So it should perform as before.

theano/tensor/extra_ops.py
nouiz: preallocated output should have the same ndim as what was declared in the make_node. If axis != NPY_MAXDIMS, the output ndim will be the same as the inputs ndim. So we are safe.

theano/tensor/extra_ops.py
nouiz: done

theano/tensor/extra_ops.py
nouiz: done

theano/tensor/extra_ops.py
notoraptor: Axis can be NPY_MAXDIMS if `self.axis` was None in Python code, and in such case I think tensor may have more than 1 dimension, isn't it ?

theano/tensor/extra_ops.py
notoraptor: OK!

theano/tensor/extra_ops.py
notoraptor: **NB**: OK, seen `make_node()`, everything is right!

PR #6108, opened 2017-07-04
Make the CAReduce tests faster

abergeron:
This also fixes:
 - minor memory leak in the elemwise alloc code
 -  compile of ParamsType with CLinker of more than one op.

### DIFF
--- a/theano/gof/params_type.py
+++ b/theano/gof/params_type.py
@@ -655,7 +655,7 @@ class ParamsType(Type):
         return list(sorted(list(c_support_code_set))) + [final_struct_code]
 
     def c_code_cache_version(self):
-        return ((2,), tuple(t.c_code_cache_version() for t in self.types))
+        return ((3,), tuple(t.c_code_cache_version() for t in self.types))
 
     # As this struct has constructor and destructor, it could be instanciated on stack,
     # but current implementations of C ops will then pass the instance by value at functions,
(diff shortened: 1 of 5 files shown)


theano/gof/params_type.py
```diff
@@ -684,6 +684,7 @@ def c_extract(self, name, sub, check_input=True):
         /* Seems c_init() is not called for a op param. So I call `new` here. */
         %(name)s = new %(struct_name)s;
 
+        { // This need a separate namespace for Clinker
```
nouiz: What is the consequence of this? a crash?

abergeron: A failure to compile.

nouiz:
It would be great to have a test for that case.

abergeron:
The optimizations include stuff that eliminates useless cases and merges some similar cases so we end up compiling less C code.

My timings indicate that test_c goes from ~200s to ~100s.

abergeron:
What case?

nouiz:
the failure to compile.

Can you confirm yout timing was always done with an empty cache?

nouiz:
I approuve this PR. You can merge if CI passes (except for the timeout).

It would be great if you can add a test case for the crash fix in commit 3465a40

abergeron:
Yes I made sure to delete the cache between each run.

abergeron:
The new form of the test for CAReduce do test the CLinker case.  That's how I found the problem.

nouiz:
jenkins fail.

nouiz:
flake8: https://travis-ci.org/Theano/Theano/jobs/250419945#L907

nouiz:
stoping travis.

abergeron:
Someone killed almost all of the travis jobs for the last commit, why?

PR #6111, opened 2017-07-05
Fix alerts found with lgtm

jhelie:
This PR fixes some alerts flagged by lgtm.com code queries. Most are minor improvements suggestions but a few some should fix unintentional code behaviour (e.g. correct call to `copy()`). 

The standard lgtm engineering analytics highlights dependencies, vulnerabilities and tracks introduced and fixed alerts as the code changes (and shows that [Theano keeps getting better](https://lgtm.com/projects/g/Theano/Theano/)). There are currently [295 alerts](https://lgtm.com/projects/g/Theano/Theano/alerts/) flagged, probably better investigated by someone more familiar with the code base than me, but you can also investigate further by writing [your own custom queries](https://lgtm.com/docs/ql/about-ql) to explore your code base.

I personally find lgtm alerts most useful when using [PR integration](https://lgtm.com/docs/lgtm/config/pr-integration) as it allows to check and deal with potential issues before they find their way into master (see for instance [12ff9ab](https://lgtm.com/projects/g/Theano/Theano/rev/12ff9ab5867d83fc9b92c926fd9d774c55c4047a)).

Full disclosure: I work on lgtm.com, hence my interest! Hope that helps, any question please ask.

### DIFF
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -864,6 +864,9 @@ class TensorConstantSignature(tuple):
             # here).
             return (self.sum == other.sum) and np.all(d0 == d1)
 
+    def __ne__(self, other):
+        return not self == other
+
     def __hash__(self):
         t, d = self
         return hashtype(self) ^ hash(t) ^ hash(d.shape) ^ hash(self.sum)
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -762,13 +762,13 @@ class _tensor_py_operators(object):
         """Fill inputted tensor with the assigned value."""
         return theano.tensor.basic.fill(self, value)
 
-    def choose(self, a, choices, out=None, mode='raise'):
+    def choose(self, choices, out=None, mode='raise'):
         """
         Construct an array from an index array and a set of arrays to choose
         from.
 
         """
-        return theano.tensor.basic.choose(a, choices, out=None, mode='raise')
+        return theano.tensor.basic.choose(self, choices, out=None, mode='raise')
 
     def squeeze(self):
         """
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -768,8 +768,7 @@ class _tensor_py_operators(object):
         from.
 
         """
-        return theano.tensor.basic.choose(self, a, choices, out=None,
-                                          mode='raise')
+        return theano.tensor.basic.choose(a, choices, out=None, mode='raise')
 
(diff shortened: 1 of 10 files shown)


nouiz:
I'm stopping travis as there is an urgent PR to merge. We will restart it later. Thanks for the PR.

nouiz:
I got a quick look at this PR. It look good, but some of the errors mean we where missing some tests.

jhelie:
@nouiz merci :) if you point me to the right place I'm happy to address the test issue?

theano/tensor/var.py
```diff
@@ -768,8 +768,7 @@ def choose(self, a, choices, out=None, mode='raise'):
         from.
 
         """
-        return theano.tensor.basic.choose(self, a, choices, out=None,
-                                          mode='raise')
+        return theano.tensor.basic.choose(a, choices, out=None, mode='raise')
```
lamblin: First, I think the signature of the original function is not correct. We should really have:
```python
def choose(self, choices, out=None, mode='raise'):
    return theano.tensor.basic.choose(self, choices, out=out, mode=mode)
```

Second, that means there is no test in the test suite that actually follows this path, otherwise it would have failed.

jhelie: yes those signatures were confusing, this makes much more sense. Updating.

As for testing all I can see is `class T_Choose(utt.InferShapeTester)` in `/tensor/tests/test_basic` seems like it should have covered this code.

jhelie:
@lamblin I've updated the signatures but someone more familiar with the code might know why `T_Choose` did not trigger/fail

lamblin:
Thanks!
From the code of `T_Choose`, the test simply always called `tensor.choose()`, and never the method `choose` of `TensorVariable`.
I'll add a test.

lamblin:
Test added.

jhelie:
great, thanks

jhelie:
@lamblin it seems Travis got stuck, I'm restarting it.

jhelie:
and now it's happy :)

jhelie:
@lamblin I rebased against the most recent master and also fixed 2 more trivial crash issues

lamblin:
Thanks a lot!

jhelie:
you're welcome - and if you think these were useful you can give a go at [PR integration]((https://lgtm.com/docs/lgtm/config/pr-integration) :)

PR #6112, opened 2017-07-05
Wrap Op params for theano.gpuarray.dnn.GpuDnnBatchNorm

notoraptor:
When working on DNN module I had not parameterized GpuDnnBatchNorm because its inputs and outputs lengths depended on a macro param. But this op has a maximum inputs length and outputs length, and then `_cop_num_inputs` and `_cop_num_outputs` allow to handle this case for COps.

This PR adds params type to GpuDnnBatchNorm. Now DNN module does not contain `get_op_params()` implementations anymore.

@abergeron @nouiz @lamblin 

**NB**: I think the last op that may be rewritten into this module is `theano.gpuarray.dnn._RNNSplitParams`. This op does not have parameters, but it also has a variable outputs length, C code here is a little more complicated, and it is not currently a real COp (`c_code()` is redefined).

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -1666,13 +1666,20 @@ class GpuDnnBatchNorm(DnnBase):
     __props__ = ('mode', 'running_averages', 'inplace_running_mean',
                  'inplace_running_var', 'inplace_output')
 
+    check_input = False
+    params_type = ParamsType(mode=cudnn.cudnnBatchNormMode_t,
+                             inplace_output=bool_t,
+                             inplace_running_mean=bool_t,
+                             inplace_running_var=bool_t,
+                             handle=handle_type)
+
     def __init__(self, mode='per-activation', running_averages=False,
                  inplace_running_mean=False, inplace_running_var=False,
                  inplace_output=False):
         DnnBase.__init__(self, ['dnn_batchnorm_base.c', 'dnn_batchnorm.c'],
                          'dnn_batchnorm_op')
 
-        assert (mode in ('per-activation', 'spatial'))
+        assert cudnn.cudnnBatchNormMode_t.has_alias(mode)
         self.mode = mode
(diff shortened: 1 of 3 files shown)


slefrancois:
jenkins test this

abergeron:
Also, if this is the last use of get_op_params(), then we should delete the support for it.

notoraptor:
@abergeron It was the last use of `get_op_params()` in DNN module, but it is still implemented at other places in Theano code!

From a quick search:
```
theano\gof
	op.py
		def get_op_params(self): # COp
theano\gpuarray
	blocksparse.py
		def get_op_params(self):
		def get_op_params(self):
	linalg.py
		def get_op_params(self):
		def get_op_params(self):
	nerv.py
		def get_op_params(self):
	pool.py
		def get_op_params(self):
		def get_op_params(self):
		def get_op_params(self):
theano\gpuarray\tests
	test_cgpukernelbase.py
		def get_op_params(self):
```

abergeron:
Then we still have work to do.

theano/gpuarray/dnn.py
```diff
@@ -1700,24 +1707,12 @@ def __setstate__(self, d):
...
-                params.append(('INPLACE_RUNNING_VAR', '1'))
-        params.append(('MODE', ("CUDNN_BATCHNORM_SPATIAL"
-                                if self.mode == "spatial"
-                                else "CUDNN_BATCHNORM_PER_ACTIVATION")))
-        return params
-
     def infer_shape(self, node, shape):
         return [shape[0]] + [shape[1]] * (len(node.outputs) - 1)
 
+    _cop_num_inputs = 7
```
nouiz: it would be better to have the attribute at the top of the class, next to props.

notoraptor: Done.

notoraptor:
1 timeout error with TRAVIS, not related to this PR.

PR #6116, opened 2017-07-05
Use GpuDnnReduction to replace GpuMaxAndArgmax when possible.

abergeron:


abergeron:
This might be missing tests for the Argmax -> GpuDnnReduction optimization path.

theano/tensor/basic.py
```diff
@@ -1427,21 +1427,31 @@ class Argmax(Op):
     nin = 2  # tensor, axis
     nout = 1
     E_axis = 'invalid axis'
-    __props__ = ()
+    __props__ = ('axis',)
```
nouiz: why this change? This break compatibility. It will be hard to fix, as we can't fix it with just a set_state method as we need to change all apply nodes that reuse that op, not just the op instance.

I would prefer that we don't change that. You can easily do get_scalar_constant_value(node.inputs[1]) to get the constant axis.

abergeron: The axis can be
 - NoneConst
 - A theano 1d tensor
so get_scalar_constant_value() will not help.

Most of the code of the op does accesses `node.inputs[1].data`, fully expecting it to be a constant, but never checking therefore it is a constant and trying to treat it like an input is dumb.

abergeron:
I've added a fix for the problem reported in #6118

theano/tensor/basic.py
abergeron: Also, I don't think we should try to maintain pickle compatibility in this case.  It won't matter for unoptimized graphs since this op is not in them, and we can ask people to re-optimize the graph for those that need it.

nouiz: It can be in the non-optimized graph... But the probability is pretty low. I'll let @lamblin choose.

theano/gpuarray/dnn_redux.c
```diff
@@ -61,11 +61,6 @@ int APPLY_SPECIFIC(dnn_redux)(PyGpuArrayObject *input,
   static float fbeta = 0.0f;
   static double dbeta = 0.0;  
 
-  if (!GpuArray_IS_C_CONTIGUOUS(&input->ga)) {
```
nouiz: This would need tests, as it is not part of the cudnn doc. So even if it work in one cases, it don't mean it will work with all cases. Also futur version can change this.

Until we get an official confirmation that this is safe, we should add back this restriction and add a gpu_contiguous in the opt that move the code to the GPU.

theano/gpuarray/dnn.py
```diff
@@ -3130,6 +3132,66 @@ def local_dnn_reduction(node):
...
+    if node.inputs[0].ndim > 8:
+        return
+
+    if node.inputs[0].dtype != node.outputs[0].dtype:
+        return
+
+    if node.inputs[0].dtype not in ['float16', 'float32', 'float64']:
+        return
+
+    # order of the axes influences the output indices
```
nouiz: On the CPU, I think we don't have this behavior. The order isn't important. Could we just sort them and so be able to always use cudnn reduction?

theano/gpuarray/dnn.py
```diff
@@ -3130,6 +3132,66 @@ def local_dnn_reduction(node):
...
+    if (node.op.axis is not None and
+            tuple(sorted(node.op.axis)) != node.op.axis):
+        return
+
+    max, arg = GpuDnnReduction('maximum', node.op.axis, node.outputs[0].dtype,
+                               node.outputs[0].dtype, True)(node.inputs[0])
+
+    # cudnn can only return int32 indices
```
nouiz: This isn't super efficient. I think we should make an output dtype param to the CPU op. I make an issue about this. If the user casted the argmax output to int32, do we have the cast optimization run that would remove those extra optimization?

theano/gpuarray/dnn.py
```diff
@@ -3130,6 +3132,66 @@ def local_dnn_reduction(node):
...
+        return
+
+    if inputs[0].ndim > 8:
+        return
+
+    if inputs[0].dtype not in ['float16', 'float32', 'float64']:
+        return
+
+    # order of the axes influences the output indices
+    if op.axis is not None and tuple(sorted(op.axis)) != op.axis:
```
nouiz: You don't have the same restriction as the GpuMaxAndArgmax optimizer. You don't have the input dtype == output dtype restrict.

theano/gpuarray/tests/test_reduction.py
```diff
@@ -54,13 +56,13 @@ def numpy_maxandargmax(X, axis=None):
 
 
 def check_if_gpu_maxandargmax_in_graph(theano_function):
```
nouiz: Change the name. Now you check that GpuDnnReduction is in the graph.

theano/tensor/basic.py
nouiz: I thinked of another option. You could just implement the prepare_node() method and make it raise a good error if we try to compile an old node. I discussed this IRL with @lamblin and he also agreed with this. What do you think?

slefrancois:
jenkins test this (testing master cache)

theano/gpuarray/dnn.py
abergeron: I am pretty sure that axis is sorted on the CPU so this check should always pass.  But I would like to keep this to make sure we don't return bad results.

theano/gpuarray/dnn.py
abergeron: Yes there is an optimization that removes consecutive casts.  However most people will want int64 indices.

cuDNN might get support for 64 bit indices in a future version and we can switch that at that time.

theano/gpuarray/dnn.py
abergeron: Of course.  This is irrelevant for argmax.

theano/gpuarray/tests/test_reduction.py
abergeron: ok

theano/gpuarray/dnn_redux.c
abergeron: I'll add some tests for that.

lamblin: Indeed, I checked the doc as well, and nothing indicates that negative strides, or even overlapping strides (!) would be an issue, as those restrictions are indicated for other algos (LCN for instance).
I'd still feel safer if we have a test that the library indeed behaves according to the doc...

lamblin:
Flake8 is angry because:
```
/home/jenkins/workspace/Theano_PR/theano/gpuarray/tests/test_reduction.py:110:9: F821 undefined name 'check_if_gpu_maxandargmax_not_in_graph'
/home/jenkins/workspace/Theano_PR/theano/gpuarray/tests/test_reduction.py:121:9: F821 undefined name 'check_if_gpu_maxandargmax_in_graph'
/home/jenkins/workspace/Theano_PR/theano/gpuarray/tests/test_dnn.py:1526:57: E901 SyntaxError: invalid syntax
```

abergeron:
Apparently `...` is python3-only.

lamblin:
Some errors because:
`"TypeError: ('The following error happened while compiling the node', Argmax{axis=(0,)}(x), '\n', "prepare_node() got an unexpected keyword argument 'storage_map'")"`

Also, as we discussed IRL, please check that the values for strided reductions are correct.

abergeron:
Should be fixed now, and I compare the results with numpy for strange strides.

abergeron:
Why did I have to pick the one example in the source tree that got it wrong...

Anyways both spot are fixed.

PR #6117, opened 2017-07-05
Remove duplicate tests. No need to retest them all in 2d and 4d. Fusi…

nouiz:
…on don't depend too much about ndim

### DIFF
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -1178,7 +1178,7 @@ class test_fusion(unittest.TestCase):
         mode._optimizer = mode._optimizer.including(
             'local_elemwise_fusion', 'composite_elemwise_fusion',
             'canonicalize')
-        self.do(mode, self._shared, shp)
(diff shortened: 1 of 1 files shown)


theano/tensor/tests/test_opt.py
```diff
@@ -1178,7 +1178,19 @@ def test_elemwise_fusion_4d(self):
...
+
+        def my_init(shp, dtype='float64', num=0):
+            ret = np.zeros(shp, dtype=dtype) + num
+            return ret
+        fxv = my_init(shp, 'float32', 2)
+        fyv = my_init(shp, 'float32', 3)
+        fzv = my_init(shp, 'float32', 4)
+        f = theano.function([fx, fy, fz], fx + fy + fz, mode=mode)
+        out = f(fxv, fyv, fzv)
+        utt.assert_allclose(out, fxv + fyv + fzv)
```
lamblin: How exactly does that test verify that the optimization has applied?

nouiz:
I updated this PR

PR #6119, opened 2017-07-06
Get rid of get_op_params()

### LINKED ISSUE #6114: Should nervanagpu be still supported by Theano ?
A Python module called `nervanagpu` is used in `theano/gpuarray/nerv.py`.

Got from this page: https://github.com/NervanaSystems/nervanagpu :

> ## Deprecated
> 
> NervanaGPU is now maintained as a backend inside of the [neon](https://github.com/NervanaSystems/neon) project. This repository remains for reference only and is no longer updated.

And it seems the PyPI package has not been updated since 2015: https://pypi.python.org/pypi/nervanagpu/0.3.1

Do we want to still keep the `nerv` module into Theano ? Or should it be removed / marked as `deprecated` too / rewritten to use the new module `neon` ?

@nouiz @lamblin @abergeron

notoraptor:
Discussed here https://github.com/Theano/Theano/pull/6112#issuecomment-313126999 . This PR removes `COp.get_op_params()` implementations everywhere it is used in current Theano master branch, so that this function becomes unsupported.

This PR also remove `nerv` module in gpuarray backend, and then fix #6114 .

**NB** (reminder): there may be some currently open PRs that still use `get_op_params()`. So, either this PR will have to be updated if some of that other PRs are merged before, or all these other PRs will need to be rebased (and then replace `get_op_params()` with use of `ParamsType`) if this PR is merged first.

@abergeron @nouiz @lamblin


theano/gpuarray/nerv.py
```diff
@@ -1,193 +1,7 @@
...
-            isinstance(C.owner.op, GpuAllocEmpty) and
-            len(C.clients) > 1):
-        inputs[0] = C.owner.op(*C.owner.inputs)
-    return [Gemm16(relu=node.op.relu, inplace=True)(*inputs)]
-
-optdb.register('local_gemm16_inplace',
-               tensor.opt.in2out(local_gemm16_inplace,
-                                 name='local_gemm16_inplace'),
-               70.0, 'fast_run', 'inplace', 'gpuarray')
+raise SkipTest("You are importing theano.gpuarray.nerv. "
```
abergeron: Maybe not SkipTest?

notoraptor: I copied the style in `theano.sandbox.cuda`.

`ImportError` instead ? Should I change it into sandbox cuda also ?

abergeron: I apparently didn't catch that sandbox.cuda uses SkipTest, but yeah, ImportError is better.

If that makes some tests fail, then we can fix them.

notoraptor: Done.

notoraptor:
A surprise: now there is a Jenkins error because something tries to import sandbox cuda: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/331/console . It is a little difficult to get the import point from Jenkins output. I'm investigating.

notoraptor:
NB: I think it's because Jenkins run nosetests on the entire `theano` folder. Then, nosetests looks for tests in all subdirectories, and fails when it looks into folder `theano/sandbox/cuda`. That may be why we originally raise a SkipTest into `theano/sandbox/cuda/__init__.py`.

Maybe a solution could be to convert module folder `theano/sandbox/cuda` to a module file `theano/sandbox/cuda.py`, so that nosetests automatically skips `cuda.py` as this file won't start with `test_` prefix ?

@abergeron

nouiz:
We can't raise importerror when a file is loaded. Otherwise nosetests will
complain. This is why I put skiptest instead.

notoraptor:
It seems to be the only one failing case (nosetests import step), meaning that there is no more place into Theano where sandbox cuda is still unfortunately imported. So I think I can reraise a SkipTest instead of an ImportError for sandbox cuda (while keeping ImportError for nerv). @abergeron @nouiz .

abergeron:
Fine let's go with SkipTest.  But this might warrant a comment that explains why it's that exception and not ImportError.

notoraptor:
@abergeron Done!

theano/gpuarray/linalg.py
```diff
@@ -385,25 +396,19 @@ def make_node(self, A):
         assert A.dtype == 'float32'
         if self.compute_uv:
             return theano.Apply(self, [A],
-                                [A.type(),
-                                GpuArrayType(A.dtype, broadcastable=[False],
-                                             context_name=ctx_name)(),
-                                A.type()])
+                                # return S, U, VT
+                                [GpuArrayType(A.dtype, broadcastable=[False],
+                                              context_name=ctx_name)(),
```
nouiz: this change the interface. We should add a detection somewhre that if an old pickled file reload the old order, then it raise an error when w try to make a thunk of this. I think you should  implement the check in th emethod prepare_node()

notoraptor: Done. Is it right (see last commit) ?

theano/gpuarray/linalg.py
```diff
@@ -407,6 +407,15 @@ def make_node(self, A):
                                 [GpuArrayType(A.dtype, broadcastable=[False],
                                               context_name=ctx_name)()])
 
+    def prepare_node(self, node, storage_map, compute_map, impl):
```
nouiz: Did you manually tested it? Just to be sure it does what we want.

notoraptor: Yes, I tested in (running nosetests on `test_linalg.py`), assertion is evaluated and succeeds.

PR #6120, opened 2017-07-06
Various improvements for Travis

lamblin:
- stages so we stop if pep8 or doc fails
- speed up or split some tests
- limit to 1 thread to avoid aberrant timings for corr3d
- pep8

It is not enough to make it pass on an empty cache yet.


.travis.yml
```diff
@@ -42,32 +40,61 @@ install:
...
+      env: FAST_COMPILE=1 PART="theano/tensor --exclude-dir=theano/tensor/nnet"
+    - <<: *normaltest
+      env: FAST_COMPILE=1 PART="theano/tensor/nnet"
+    - <<: *normaltest
+      python: "3.4"
+      env: FAST_COMPILE=1 PART="theano/tensor/nnet"
 
 script:
+  - if [[ $FAST_COMPILE == "1" ]]; then export THEANO_FLAGS=$THEANO_FLAGS,mode=FAST_COMPILE; fi
+  - if [[ $FAST_COMPILE == "1" && $TRAVIS_PYTHON_VERSION == '2.7' ]]; then export THEANO_FLAGS=$THEANO_FLAGS,floatX=float32; fi
```
nouiz: I don't like this. Why not set directly the THEANO_FLAGS in the "env:" line? Yes, you prevent to duplicate that flag, but it make the code more complicated and hide the config difference about floatX=float32. Can you remove line 97 and set floatX=float32 in the "env:" line? Or make an FLOATX env variable and do the same equivalent to FAST_COMPILE env var? That would not hide that flag change.

lamblin: Ultimately, I wanted to also use float32 for python 3 and FAST_RUN, to have more combinations with the same number of jobs. I thought it would be more clear to do that automatically rather than copy/paste it all over the place.
I'm not sure why it is so important to show floatX, but I can revert that,

nouiz:
jenkins test this

jenkins had one failure due to small numerical error:

http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/338/testReport/theano.gpuarray.tests.test_linalg/TestMagma/test_gpu_matrix_inverse/

PR #6060, opened 2017-06-20
(small fix) Increase error tolerance for magma matrix_inverse tests.

notoraptor:
About this error: http://earlgrey.iro.umontreal.ca:8080/job/Theano_buildbot_python3/203/testReport/theano.gpuarray.tests.test_linalg/TestMagma/test_gpu_matrix_inverse_2/

I can not reproduce the problem (with Python 3), maybe because test values are randomly chosen, but I think increasing the rtol to 1e-3 should help tests to pass, as results compare `matrix * matrix inverse` with `np.eye()`.

@slefrancois @lamblin @nouiz .

PR #6123, opened 2017-07-07
Reorganize Travis

lamblin:
This should guarantee that even with the launch / sync overhead (3-6 minutes), and even on slower workers, on an empty Theano cache, we are under the 50 minute limits.

There is one more split, so 2 more jobs in total (python 2 and 3).


nouiz:
jenkins failure can be ignored. It will be fixed by this PR: 

https://github.com/Theano/Theano/pull/6060

Issue #6039, opened 2017-06-13
Feature Request : Gumbel Softmax


Sentient07:
The equation (1) in this [paper](https://arxiv.org/pdf/1611.01144.pdf) gives the equation of a sampler that samples from a categorical distributions. The advantage of this discrete sampler over the others is that it is differentiable and could be trained with backprop in an environment of discrete data

nouiz:
Can you put the equation in the issue to help make it easier to follow? I
checked and Theano have all the ops needed to implement it. Mostly, log,
elemwise add, argmax and one_hot. one_hot can be done with sub alloc(0,
shape) and set_subtensor.

Sentient07:
> Can you put the equation in the issue to help make it easier to follow?

I tried to add in the comment field, but the code of github doesn't support latex, so I added a screenshot with the three necessary equations (sampling, argmax approximation, gumbel distribution's density)

>  one_hot can be done with sub alloc(0, shape) and set_subtensor.

I thought this could be done with the [Op](http://deeplearning.net/software/theano/library/tensor/extra_ops.html#theano.tensor.extra_ops.to_one_hot) , but why the longer approach?

nouiz:
The longer to implement approch is using an op. The faster to implement is
using existing ops and build a graph. This can be hidden inside a python
function.

So when do we need a new op? If we want more speed, sometimes, it could be
faster. But it isn't always the case. But this would request you to make a
CPU and GPU version. To implement the grad on CPU an GPU. So this is much
more time consuming. Start with a graph version and only try to make an op
version if this part take a significant execution time.

Sentient07:
> The faster to implement is using existing ops and build a graph.

Ah, I get it! 

> If we want more speed, sometimes, it could be faster. But it isn't always the case.

Basically, if I build a graph to do this computation with the existing Ops and it happens to be costly, then it is worth making a new Op. Am I correct ?

> Start with a graph version and only try to make an op version if this part take a significant execution time.

Makes sense

Sentient07:
This is how the graphs look (first is the that of the variable `z` in equation (1) and second is variable `y` in equation (2)). The crux here is we can use (1) in the forward pass, and differentiate (2) at the backward pass. 

P. S : From what I understand, this approximation of using (1) for forward and (2) in backward holds good at low temperature, but the training they have described in the paper is to start from high temperature and anneal it. 

[...]

PR #6124, opened 2017-07-07
Gumbel softmax

Sentient07:
Implements Gumbel Softmax, addresses #6039


Sentient07:
Travis is failing, I am able to produce it locally. The same code worked on my ipython-notebook, but with non-seeded random numbers. I will look into it

PR #6125, opened 2017-07-07
Execution order

nouiz:
This should Theano function to return async more frequently.

So if combined with the flag gpuarray.single_stream=False, would allow more transfer/compute overlap between mini-batch.


botev:
Could you provide an example of using this, the test does not really shed too much light on what is going on.

nouiz:
2 other people else timed it and it give just a small speed up. So this need more investigation. We where hoping to get ~20% speed, but got ~2%. So this need more investigation.

slefrancois:
jenkins test this (https enabled)

PR #6132, opened 2017-07-10
Add a parameter axis as Param Type for Softmax

bordesf:
Ref to #5712 and #5719.

This pull request add the needed code improvements to be able to choose as parameter one arbitrary axis for the Softmax, GradSoftmax and LogSoftmax op.

### DIFF
--- a/theano/tensor/nnet/nnet.py
+++ b/theano/tensor/nnet/nnet.py
@@ -27,12 +27,12 @@ from theano.tensor import basic as tensor, subtensor, opt, elemwise
 from theano.tensor.type import (values_eq_approx_remove_inf,
                                 values_eq_approx_remove_nan)
 from theano.compile import optdb
-from theano.gof import Apply
-
+from theano.gof import Apply, ParamsType
 from theano.tensor.nnet.sigm import sigmoid, softplus
 from theano.gradient import DisconnectedType
 from theano.gradient import grad_not_implemented
 from theano.tensor.nnet.blocksparse import sparse_block_dot
+from theano.scalar import Scalar
 
 ############
 #
@@ -292,119 +292,182 @@ softmax_with_bias = SoftmaxWithBias()
 class SoftmaxGrad(gof.Op):
     """
     Gradient wrt x of the Softmax Op.
-
     """
 
     nin = 2
     nout = 1
     __props__ = ()
+    params_type = ParamsType(axis=Scalar('int8'))
+
+    def __init__(self, axis=-1):
+        self.axis = axis
 
     def make_node(self, dy, sm):
         dy = tensor.as_tensor_variable(dy)
         sm = tensor.as_tensor_variable(sm)
-        if dy.type.ndim not in (1, 2) \
-                or dy.type.dtype not in tensor.float_dtypes:
-            raise ValueError('dy must be 1-d or 2-d tensor of floats. Got ',
-                             dy.type)
+        if dy.type.dtype not in tensor.float_dtypes:
+            raise ValueError('dy must be tensor of floats. Got ', dy.type)
         if dy.ndim == 1:
             dy = tensor.shape_padleft(dy, n_ones=1)
         if sm.ndim == 1:
             sm = tensor.shape_padleft(sm, n_ones=1)
         return Apply(self, [dy, sm], [sm.type()])
 
-    def perform(self, node, input_storage, output_storage):
+    def perform(self, node, input_storage, output_storage, param):
         dy, sm = input_storage
+        axes = param.axis
+        if (dy.shape != sm.shape):
+            raise ValueError('dy and the softmax output should have the same shape.')
         dx = np.zeros_like(sm)
         # dx[i,j] = - (\sum_k dy[i,k] sm[i,k]) sm[i,j] + dy[i,j] sm[i,j]
-        for i in xrange(sm.shape[0]):
-            dy_times_sm_i = dy[i] * sm[i]
-            dx[i] = dy_times_sm_i - sum(dy_times_sm_i) * sm[i]
+        dy_times_sm = dy * sm
+        dx = dy_times_sm - (np.sum(dy_times_sm, axis=axes, keepdims=True) * sm)
         output_storage[0][0] = dx
 
     def grad(self, inp, grads):
         dy, sm = inp
         g, = grads
-
-        tmp = g + tensor.neg(tensor.sum(g * sm, axis=1).dimshuffle((0, 'x')))
+        axes = self.axis
+        tmp = g + tensor.neg(tensor.sum(g * sm, axis=axes, keepdims=True))
         g_dy = tmp * sm
-
-        tmp2 = tensor.sum(dy * sm, axis=1).dimshuffle((0, 'x'))
+        tmp2 = tensor.sum(dy * sm, axis=axes, keepdims=True)
         g_sm = tmp * dy - g * tmp2
-
         return g_dy, g_sm
 
     def infer_shape(self, node, shape):
         return [shape[1]]
 
-    def c_code_cache_version(self):
-        return (3,)
-
     def c_code(self, node, name, inp, out, sub):
         dy, sm = inp
+        param = sub['params']
         dx, = out
         return '''
-        if ((PyArray_TYPE(%(dy)s) != NPY_DOUBLE) &&
-            (PyArray_TYPE(%(dy)s) != NPY_FLOAT))
-        {
-            PyErr_SetString(PyExc_TypeError,
-                 "types should be float or float64");
-            %(fail)s;
-        }
-        if ((PyArray_TYPE(%(sm)s) != NPY_DOUBLE) &&
-            (PyArray_TYPE(%(sm)s) != NPY_FLOAT))
-        {
-            PyErr_SetString(PyExc_TypeError,
-                 "types should be float or float64");
-            %(fail)s;
-        }
-        if ((PyArray_NDIM(%(dy)s) != 2)
-            || (PyArray_NDIM(%(sm)s) != 2))
-        {
-            PyErr_SetString(PyExc_ValueError, "rank error");
-            %(fail)s;
-        }
-        if (PyArray_DIMS(%(dy)s)[0] != PyArray_DIMS(%(sm)s)[0])
-        {
-            PyErr_SetString(PyExc_ValueError, "dy.shape[0] != sm.shape[0]");
-            %(fail)s;
-        }
-        if ((NULL == %(dx)s)
-            || (PyArray_DIMS(%(dx)s)[0] != PyArray_DIMS(%(sm)s)[0])
-            || (PyArray_DIMS(%(dx)s)[1] != PyArray_DIMS(%(sm)s)[1]))
-        {
-            Py_XDECREF(%(dx)s);
-            %(dx)s = (PyArrayObject*) PyArray_SimpleNew(2,
-                                                        PyArray_DIMS(%(sm)s),
-                                                        PyArray_TYPE(%(sm)s));
-            if (!%(dx)s)
+            npy_int ndim_sm = PyArray_NDIM(%(sm)s);
+            npy_int ndim_dy = PyArray_NDIM(%(dy)s);
+            npy_intp* shape_sm = PyArray_DIMS(%(sm)s);
+            npy_intp* shape_dy = PyArray_DIMS(%(dy)s);
+            npy_int axis = %(param)s->axis;
+
+            // Check input type
+            if ((PyArray_TYPE(%(dy)s) != NPY_DOUBLE) &&
+                    (PyArray_TYPE(%(dy)s) != NPY_FLOAT))
+                {
+                    PyErr_SetString(PyExc_TypeError,
+                        "types should be float or float64");
+                    %(fail)s;
+                }
+
+            if ((PyArray_TYPE(%(sm)s) != NPY_DOUBLE) &&
+                (PyArray_TYPE(%(sm)s) != NPY_FLOAT))
+                {
+                    PyErr_SetString(PyExc_TypeError,
+                        "Types should be float or float64");
+                    %(fail)s;
+                }
+
+            // Check the number of dimension
+            if (ndim_dy != ndim_sm)
+                {
+                    PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same number of dimension.");
+                    %(fail)s;
+                }
+
+            // Check if the axis is valid
+            if (axis < 0)
             {
-                PyErr_SetString(PyExc_MemoryError,
-                     "failed to alloc dx output");
+                axis += ndim_sm;
+            }
+            if(axis > ndim_sm - 1 || axis < 0){
+                PyErr_SetString(PyExc_ValueError,
+                "GradSoftmax, bad axis argument");
+                %(fail)s
+            }
+
+            // Check the shapes of sm and dy
+            npy_int shape_not_equal;
+            shape_not_equal = 0;
+            for (size_t i = 0; i < ndim_dy && !shape_not_equal; i++)
+            {
+                shape_not_equal = (shape_sm[i] != shape_dy[i]);
+            }
+
+            if (shape_not_equal)
+            {
+                PyErr_SetString(PyExc_ValueError, "Sm and dy should have the same shape.");
                 %(fail)s;
             }
-        }
 
-        for (size_t i = 0; i < PyArray_DIMS(%(dx)s)[0]; ++i)
-        {
-            const dtype_%(dy)s* __restrict__ dy_i = (dtype_%(dy)s*) (PyArray_BYTES(%(dy)s) + PyArray_STRIDES(%(dy)s)[0] * i);
-            npy_intp Sdy = PyArray_STRIDES(%(dy)s)[1]/sizeof(dtype_%(dy)s);
-            const dtype_%(sm)s* __restrict__ sm_i = (dtype_%(sm)s*) (PyArray_BYTES(%(sm)s) + PyArray_STRIDES(%(sm)s)[0] * i);
-            npy_intp Ssm = PyArray_STRIDES(%(sm)s)[1]/sizeof(dtype_%(sm)s);
-            dtype_%(dx) s* __restrict__ dx_i = (dtype_%(dx)s*) (PyArray_BYTES(%(dx)s) + PyArray_STRIDES(%(dx)s)[0] * i);
-            npy_intp Sdx = PyArray_STRIDES(%(dx)s)[1]/sizeof(dtype_%(dx)s);
+            //If the memory is already allowed
+            if(NULL != %(dx)s)
+            {
+                npy_int ndim_dx = PyArray_NDIM(%(dx)s);
+                npy_intp* shape_dx = PyArray_DIMS(%(dx)s);
+                //Check if dims are equal
+                shape_not_equal = (NULL == shape_dx || ndim_sm != ndim_dx);
+                // Check if the shape are identical
+                for (size_t i = 0; i < ndim_sm && !shape_not_equal; i++)
+                {
+                    shape_not_equal = (shape_sm[i] != shape_dx[i]);
+                }
+            }
+            // Alloc memory for the output
+            if ((NULL == %(dx)s) || shape_not_equal)
+                {
+                    Py_XDECREF(%(dx)s);
+                    %(dx)s = (PyArrayObject*) PyArray_SimpleNew(PyArray_NDIM(%(sm)s),
+                                                                PyArray_DIMS(%(sm)s),
+                                                                PyArray_TYPE(%(sm)s));
+                    if (!%(dx)s)
+                    {
+                        PyErr_SetString(PyExc_MemoryError,
+                            "failed to alloc dx output");
+                        %(fail)s;
+                    }
+                }
 
-            double sum_dy_times_sm = 0.;
-            for (size_t j = 0; j < PyArray_DIMS(%(dx)s)[1]; ++j)
+            // Use numpy iterator
+            npy_int num_sm_classes, stride_dx, stride_sm, stride_dy;
+            // Get numpy iterator
+            PyArrayIterObject *it_dx, *it_sm, *it_dy;
+            it_dx = (PyArrayIterObject *) PyArray_IterAllButAxis((PyObject *) %(dx)s, &axis);
+            it_sm = (PyArrayIterObject *) PyArray_IterAllButAxis((PyObject *) %(sm)s, &axis);
+            it_dy = (PyArrayIterObject *) PyArray_IterAllButAxis((PyObject *) %(dy)s, &axis);
+            // Compute the stride on the selected dimension
+            stride_dx = PyArray_STRIDE(%(dx)s, axis)/sizeof(dtype_%(dx)s);;
+            stride_sm = PyArray_STRIDE(%(sm)s, axis)/sizeof(dtype_%(sm)s);
+            stride_dy = PyArray_STRIDE(%(dy)s, axis)/sizeof(dtype_%(dy)s);
+            // Get the shape on the specified dimension
+            num_sm_classes = shape_sm[axis];
+            // Go through the array
+            dtype_%(dx)s *dx_i;
+            dtype_%(sm)s *sm_i;
+            dtype_%(dy)s *dy_i;
+            while(PyArray_ITER_NOTDONE(it_dx))
             {
-                dx_i[j * Sdx] = dy_i[j * Sdy] * sm_i[j * Ssm];
-                sum_dy_times_sm += dx_i[j * Sdx];
+                dx_i = (dtype_%(dx)s *) PyArray_ITER_DATA(it_dx);
+                sm_i = (dtype_%(sm)s *) PyArray_ITER_DATA(it_sm);
+                dy_i = (dtype_%(dy)s *) PyArray_ITER_DATA(it_dy);
+
+                double sum_dy_times_sm = 0.;
+                for (size_t j = 0; j < num_sm_classes; ++j)
+                {
+                    dx_i[j] = dy_i[j] * sm_i[j];
+                    sum_dy_times_sm += dx_i[j];
+                }
+                for (size_t j = 0; j < num_sm_classes; ++j)
+                {
+                    dx_i[j] -= sum_dy_times_sm * sm_i[j];
+                }
+                PyArray_ITER_NEXT(it_dx);
+                PyArray_ITER_NEXT(it_sm);
+                PyArray_ITER_NEXT(it_dy);
             }
-            for (size_t j = 0; j < PyArray_DIMS(%(dx)s)[1]; ++j)
-            {
-                dx_i[j * Sdx] -= sum_dy_times_sm * sm_i[j * Ssm];
-            }
-        }
-        ''' % dict(locals(), **sub)
+            ''' % dict(locals(), **sub)
+
+    # @staticmethod
+    #def c_code_cache_version(self):
+    #    return (4,)
+
 softmax_grad = SoftmaxGrad()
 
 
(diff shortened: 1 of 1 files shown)


bordesf:
@ReyhaneAskari @lamblin @shawntan  I think we can start from this PR to modify the tests in order to include the vector case.

theano/tensor/nnet/nnet.py
```diff
@@ -414,40 +477,40 @@ class Softmax(gof.Op):
...
+    This function get applied when axis is a scalar and uses a fast C optimization.
     """
 
-    nin = 1
+    nin = 2  # tensor, axis
     nout = 1
     __props__ = ()
+    params_type = ParamsType(axis=Scalar('int8'))
+
+    def __init__(self, axis=-1):
```
bordesf: Yeah, but for this PR, we start by considering only the last axis. It will be easier to write the test for this specific case (Like testing vector, matrix, tensor*d). Then, we can have a new PR that will add this functionality for the user. (Since it will requires to modify the optimization, and that it will generate more cases to test, I prefer to have it in another PR).

shawntan: OK! @ReyhaneAskari explained to me earlier. Thanks!

shawntan:
I rebased `master` and added the tests.

shawntan:
So I've attempted changing some of the tests so that they don't assume the extra padding is needed, but I'm running into errors like this:
```
FAIL: test_optimize_xent_vector (theano.tensor.nnet.tests.test_nnet.T_CrossentropyCategorical1Hot)
  File "/u/tanjings/Theano/theano/tensor/nnet/tests/test_nnet.py", line 1027, in test_optimize_xent_vector
    assert crossentropy_softmax_argmax_1hot_with_bias in ops
AssertionError:
Elemwise{Neg}[(0, 0)] [id A] ''   4
 |Sum{acc_dtype=float64} [id B] ''   3 
   |Elemwise{Log}[(0, 0)] [id C] ''   2
     |AdvancedSubtensor1 [id D] ''   1 
       |Softmax [id E] ''   0
       | |x [id F]
       |y [id G]

```
Is there an optimisation that is supposed to trigger that is not happening here?

bordesf:
@shawntan I think so. You can push your changed tests and I will manage the optimization things.

shawntan:
There is also another error for `coding_dist` which requires `.ndim == 2`.

theano/tensor/nnet/nnet.py
```diff
@@ -292,119 +292,178 @@ def c_code_cache_version():
...
         g_sm = tmp * dy - g * tmp2
-
         return g_dy, g_sm
 
     def infer_shape(self, node, shape):
         return [shape[1]]
 
-    def c_code_cache_version(self):
-        return (3,)
-
```
ReyhaneAskari: Why have you removed the whole ```c_code_cache_version``` function?

theano/tensor/nnet/nnet.py
```diff
@@ -414,39 +473,37 @@ class Softmax(gof.Op):
...
+        # Apply softmax on the specified dimension
+        e_x = np.exp(x - x.max(axis=axes, keepdims=True))
+        sm = e_x / e_x.sum(axis=axes, keepdims=True)
         output_storage[0][0] = sm
 
     def L_op(self, inp, outputs, grads):
         x, = inp
         g_sm, = grads
-        return [softmax_grad(g_sm, outputs[0])]
+        return [SoftmaxGrad(self.axis)(g_sm, outputs[0])]
```
ReyhaneAskari: This way, every time that perform is called a new ```SoftmaxGrad``` object is created. Doesn't it increase the memory usage and time?

theano/tensor/nnet/nnet.py
```diff
@@ -600,34 +693,36 @@ class LogSoftmax(gof.Op):
...
+        axes = param.axis
+        xdev = x - x.max(axis=axes, keepdims=True)
+        lsm = xdev - np.log(np.sum(np.exp(xdev), axis=axes, keepdims=True))
         output_storage[0][0] = lsm
 
     def grad(self, inp, grads):
         x, = inp
-        sm = softmax_op(x)
-        return [grads[0] - tensor.sum(grads[0], axis=1, keepdims=True) * sm]
+        sm = Softmax(self.axis)(x)
```
ReyhaneAskari: Same in here.

bordesf: Hmm yes. Maybe I can move the softmax = Softmax(self.axis) to the __init__. So, I will only need to call self.softmax(x) in the grad function. Same for above.

theano/tensor/nnet/nnet.py
bordesf: Oups

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -88,12 +88,20 @@ def test_infer_shape(self):
         self._compile_and_check([admat], [Softmax()(admat)],
                                 [admat_val], Softmax)
 
-    def test_vector(self):
-        x = T.vector()
-        f = theano.function([x], softmax_op(x))
-
-        xv = np.random.randn(6).astype(config.floatX)
-        assert np.allclose(f(xv), np.exp(xv) / np.exp(xv).sum())
+    def test_multi_axis(self):
```
ReyhaneAskari: Please change ```axis``` to ```axes```.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -88,12 +88,20 @@ def test_infer_shape(self):
...
+        dims = 4
+        shape = (5,) * dims
+        xv = np.random.randn(*shape).astype(config.floatX)
+        for axes in xrange(1, dims + 1):
```
ReyhaneAskari: The name ```axes``` is not good for here. It is actually a dimension. So please change the name.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -88,12 +88,20 @@ def test_infer_shape(self):
...
+            x = T.TensorType(dtype=config.floatX,
+                             broadcastable=(False,) * axes)()
+            outputs = softmax_op(x)
+            test_val = xv[((0,) * (dims - axes))]
```
ReyhaneAskari: This is not quite readable. So either make it more readable or add a comment.

theano/tensor/nnet/nnet.py
ReyhaneAskari: Yeah, thanks. I didn't see it.

theano/tensor/nnet/nnet.py
ReyhaneAskari: Yes, that makes more sense. Although it depends on how you want to access the ```axis```. If it's going to be through ```params```, it's better to remove the whole ```init``` and initialize without having the axis as variable. So like before, you would create one object outside the class and use it several times.

lamblin:
Summary of IRL discussion:
- it does not really matter if the big CrossEntropy Op is inserted in the 1D case. We should check numerical stability though, with saturated softmax. If the logsoftmax optimization applies, it may be enough.
- we should check that the log(softmax) -> logsoftmax optimization applies for ndim != 2
- we should check that the lifter optimizations for softmax and logsoftmax do not try to lift the Op when ndim != 2, at least until this is supported on GPU.

theano/tensor/nnet/nnet.py
bordesf: I don't get it. For my knowledge, the only way to use `params` is through the init of the class. If I create an object outside the class like before, I will need to specify the axis in the make_node, in this case I will not be able to use params.

shawntan:
These changes remove tests for all the other softmax optimisations.

theano/tensor/nnet/nnet.py
```diff
@@ -484,6 +484,7 @@ class Softmax(gof.Op):
 
     def __init__(self, axis=-1):
         self.axis = axis
+        self.grad_op = SoftmaxGrad(self.axis)
```
ReyhaneAskari: Just for the sake of consistency, please change the name to ```softmax_grad_op```.

theano/tensor/nnet/nnet.py
```diff
@@ -704,6 +705,7 @@ class LogSoftmax(gof.Op):
 
     def __init__(self, axis=-1):
         self.axis = axis
+        self.sm_op = Softmax(self.axis)
```
ReyhaneAskari: This one here too. ```sm_op``` to ```softmax_op```

bordesf:
@ReyhaneAskari @lamblin May I have a review of local_indexing_logsoftmax_grad ?

ReyhaneAskari:
I'm currently reviewing. I need to finish and submit the review.

theano/tensor/nnet/nnet.py
```diff
@@ -889,6 +890,39 @@ def local_logsoftmax(node):
...
+                axis = node.inputs[0].owner.op.axis
+                ret = LogSoftmax(axis)(softmax_input)
+                ret.tag.values_eq_approx = values_eq_approx_remove_inf
+                copy_stack_trace([node.inputs[0], node.outputs[0]], ret)
+                return [ret]
+
+            elif isinstance(node.inputs[0].owner.op, list_classes):
+                subtensor_op = node.inputs[0].owner.op
+                subtensor_input = node.inputs[0].owner.inputs[0]
+                subtensor_inputs = node.inputs[0].owner.inputs[1]
```
ReyhaneAskari: @bordesf , please rename ```subtensor_inputs``` to ```subtensor_idx_list``` as it has ambiguity now.

theano/tensor/nnet/nnet.py
```diff
@@ -889,6 +890,39 @@ def local_logsoftmax(node):
...
+                if isinstance(subtensor_input.owner.op, Softmax):
+                    softmax_op = subtensor_input.owner.op
+                    softmax_input = subtensor_input.owner.inputs[0]
+                    axis = softmax_op.axis
+                    ret = subtensor_op(LogSoftmax(axis)(softmax_input), subtensor_inputs)
+                    ret.tag.values_eq_approx = values_eq_approx_remove_inf
+                    copy_stack_trace([node.inputs[0], node.outputs[0]], ret)
+                    return [ret]
+
+            else:
```
ReyhaneAskari: At the end of such a function it will return any way so no ```else return``` is needed. Its existence does not hurt any logics but let's remove it for the sake of consistency.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1013,33 +934,34 @@ def test_optimize_xent_vector(self):
         mode = theano.compile.mode.get_default_mode()
         if mode == theano.compile.mode.get_mode('FAST_COMPILE'):
             mode = 'FAST_RUN'
-        rng = np.random.RandomState(utt.fetch_seed())
-        x_val = rng.randn(5).astype(config.floatX)
+#        rng = np.random.RandomState(utt.fetch_seed())
```
ReyhaneAskari: Please remove the commented line.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -1048,10 +970,10 @@ def test_optimize_xent_vector(self):
...
             try:
                 ops = [node.op for node in g.maker.fgraph.toposort()]
-                assert len(ops) == 4
-                assert crossentropy_softmax_1hot_with_bias_dx in ops
-                assert softmax_op in ops
-                assert softmax_grad not in ops
+                # assert len(ops) == 4
+                # assert crossentropy_softmax_1hot_with_bias_dx in ops
+                # assert softmax_op in ops
+                # assert softmax_grad not in ops
```
ReyhaneAskari: Shouldn't we remove these?

theano/tensor/nnet/nnet.py
```diff
@@ -869,14 +869,13 @@ def c_code_cache_version():
 
 # This is not registered in stabilize, as it cause some crossentropy
 # optimization to not be inserted.
+"""
```
ReyhaneAskari: Please remove this instead of commenting.

theano/tensor/nnet/tests/test_nnet.py
shawntan: Ah yes. Sorry about that.

ReyhaneAskari:
@bordesf there is a flake8 error:
 ```theano/tensor/nnet/nnet.py:36:1: F401 'AdvancedIncSubtensor' imported but unused```

Travis doesn't continue to do the rest of the tests when there is flake8 errors. If you could fix this, we could see the results of the tests.

bordesf:
There are some issues in debugMode:
```

ERROR: test_optimize_xent_vector (theano.tensor.nnet.tests.test_nnet.T_CrossentropyCategorical1Hot)
  File "/u/bordesfl/CCW/Theano/theano/tensor/nnet/tests/test_nnet.py", line 964, in test_optimize_xent_vector
    assert np.allclose(f(x_val, y_val), true_val)
  File "/u/bordesfl/CCW/Theano/theano/compile/function_module.py", line 883, in __call__
  File "/u/bordesfl/CCW/Theano/theano/compile/debugmode.py", line 2127, in deco
    return f()
  File "/u/bordesfl/CCW/Theano/theano/compile/debugmode.py", line 1867, in f
    specific_hint=hint2)
InvalidValueError: InvalidValueError
        type(variable) = TensorType(float64, vector)
        variable       = Elemwise{log,no_inplace}.0
        type(value)    = <type 'numpy.ndarray'>
        dtype(value)   = float64
        shape(value)   = (4,)
        value          = [-inf -inf   0. -inf]
        min(value)     = -inf
        max(value)     = 0.0
        isfinite       = False
        client_node    = None
        hint           = perform output
        specific_hint  = non-finite elements not allowed
        context        = ...
  Elemwise{log,no_inplace} [id A] ''   
   |Softmax [id B] ''   
     |x [id C]

        
Elemwise{Neg}[(0, 0)] [id A] ''   3
 |Subtensor{int64} [id B] ''   2
   |LogSoftmax [id C] ''   1
   | |x [id D]
   |ScalarFromTensor [id E] ''   0
     |y [id F]

```
I think it comes from this expression l.952 in test_nnet:
`matrix_expr = -T.sum(T.log(softmax(x[None, :])[[0], [y]]))`

bordesf:
I am currently thinking at a list of the things to do next for this PR. Feel free to complete it.

About the ops:
- Make sure that local_logsoftmax and local_indexing_logsoftmax_grad work for tensor.ndim > 2
- Improve readability of the optimizations

About the tests:
Softmax op:
- Test gradients with multi dimensional tensors (and vector)
LogSoftmax:
- Test log softmax with multi dimensional tensors (and vector) + test gradients
- Test optimization Log(softmax(x)) into logsoftmax_op(x) for tensor.ndim > 2
- Test optimization Log(softmax(x)[y]) into logsoftmax_op(x) for tensor.ndim > 2 and len(y) >= 2
- Move test_optimize_xent_vector to Test_logsoftmax class.
- Test that Log(Softmax(x))'s grad is replace with the local_logsoftmax_grad for tensor.ndim > 2
- Test that Log(Softmax(x)[y])'s grad is replace with the local_indexing_logsoftmax_grad optimization for tensor.ndim >2 and len(y) > 2 in case of AdvancedIndexing

About the useness of Crossentropy in case of matrix:
-Compare performances of Crossentropy op vs log_softmax optimization and indexing in case of matrices.

theano/tensor/nnet/nnet.py
```diff
@@ -292,119 +293,178 @@ def c_code_cache_version():
...
-        if sm.ndim == 1:
-            sm = tensor.shape_padleft(sm, n_ones=1)
+        if dy.type.dtype not in tensor.float_dtypes:
+            raise ValueError('dy must be tensor of floats. Got ', dy.type)
         return Apply(self, [dy, sm], [sm.type()])
 
-    def perform(self, node, input_storage, output_storage):
+    def perform(self, node, input_storage, output_storage, param):
         dy, sm = input_storage
+        axes = param.axis
```
lamblin: There's only one axis, so maybe `axis` instead of `axes`

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
         ret.tag.values_eq_approx = values_eq_approx_remove_nan
         copy_stack_trace(node.outputs[0], ret)
         return [ret]
 
 
+@opt.register_specialize('fast_compile')
+@gof.local_optimizer([SoftmaxGrad])
+def local_indexing_logsoftmax_grad(node):
+    # One case is supported:
+    # 1. AdvancedIncSubtensor(
```
lamblin: You can maybe remove the 1., and just give the graph we are trying to match

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+    #           zeros_like(softmax(x)),
+    #           -out_grad / AdvancedSubtensor(softmax(x), arange(y.shape[0]), y),
```
lamblin: actually, it does not have to be an `AdvancedSubtensor`, nor `arange(y.shape[0])` as first argument.
I think you can just represent it as `SubtensorOp(softmax(x), *idx)`

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+    #           arange(y.shape[0]),
+    #           y)
+    #   which arises from the gradient of log(softmax(x)[arange(y.shape[0]), y])
```
lamblin: Maybe "...gradient of `log(softmax(x)[*idx])`

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+    if not (isinstance(node.op, SoftmaxGrad)):
+        return
+
+    sm = None
+    try:
+        d_sm, sm = node.inputs
+    except Exception:
+        return
+
+    if not((sm is not None) and sm.owner and isinstance(sm.owner.op, Softmax)):
```
lamblin: flake8 should complain about the lack of space after `not`

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+        return
+
+    if d_sm.owner and isinstance(d_sm.owner.op, list_classes_IncSubtensor):
+        try:
+            incr = d_sm.owner.inputs[1]
+            subtensor_inputs, = d_sm.owner.inputs[2:]
+        except:
+            return
```
lamblin: Can you remove the try/except? I don't think it is right to silence that error without understanding the possible issues.

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+
+        out_grad = 1.
+        if incr.owner and incr.owner.op == tensor.neg:
+            incr = incr.owner.inputs[0]
+            out_grad = - out_grad
+
+        if incr.owner and incr.owner.op == tensor.true_div:
+            num, denom = incr.owner.inputs
+
+            if num.ndim == 1 or np.all(num.broadcastable):
```
lamblin: That condition does not look right to me in general.
It may be OK for now if the only things we test are dimensions 1 and 2, but it would need to be upgraded later for more indexing types, I think.

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+                return
+
+            if not denom.owner:
+                return
+            if isinstance(denom.owner.op, list_classes_Subtensor):
+                # Base case
+                adv_subtensor = denom
+                # out_grad /= 1.
+            elif denom.owner.op == tensor.mul:
+                # Try to find the AdvancedSubtensor node mentionned above,
```
lamblin: Not necessarily an AdvancedSubtensor

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+                # and the output gradient
+                for i, input in enumerate(denom.owner.inputs):
+                    if input.owner and isinstance(input.owner.op, list_classes_Subtensor):
+                        other_inputs = [in_ for (j, in_) in enumerate(denom.owner.inputs) if j != i]
+                        if len(other_inputs) == 1:
+                            rest = other_inputs[0]
+                        else:
+                            rest = tensor.mul(*[other_inputs])
+                        # Check that rest is a vector or a scalar
+                        if rest.ndim == 1 or np.all(rest.broadcastable):
```
lamblin: same here

theano/tensor/nnet/nnet.py
```diff
@@ -786,16 +934,102 @@ def local_logsoftmax_grad(node):
...
+                            adv_subtensor = input
+                            out_grad /= rest
+                            break
+            else:
+                return
+
+            if adv_subtensor is not None:
+                try:
+                    maybe_sm = adv_subtensor.owner.inputs[0]
+                    maybes_subtensor_inputs, = adv_subtensor.owner.inputs[1:]
```
lamblin: Actually, that only works for a subtensor with only one index. We want to make that work whatever the indices are.
Also, I'd name the variable `maybe_subtensor_idx` or something like that.

theano/tensor/nnet/nnet.py
```diff
@@ -1450,9 +1689,7 @@ def make_node(self, coding_dist, true_one_of_n):
     def perform(self, node, inp, out):
         coding, one_of_n = inp
         y_out, = out
-        y = np.zeros_like(coding[:, 0])
-        for i in xrange(len(y)):
-            y[i] = -np.log(coding[i, one_of_n[i]])
+        y = -np.log(coding[range(coding.shape[0]), one_of_n])
```
lamblin: I think we should use `np.arange` explicitly here.

lamblin:
Also, the optimizations converting to GPU Ops should only be applied if ndim==2 and axis in (-1, 1).

theano/tensor/nnet/nnet.py
```diff
@@ -600,34 +695,37 @@ class LogSoftmax(gof.Op):
...
+        xdev = x - x.max(axis=axes, keepdims=True)
+        lsm = xdev - np.log(np.sum(np.exp(xdev), axis=axes, keepdims=True))
         output_storage[0][0] = lsm
 
     def grad(self, inp, grads):
         x, = inp
-        sm = softmax_op(x)
-        return [grads[0] - tensor.sum(grads[0], axis=1, keepdims=True) * sm]
+        sm = self.softmax_op(x)
+        return [grads[0] - tensor.sum(grads[0], axis=self.axis, keepdims=True) * sm]
```
bordesf: @ReyhaneAskari said I shouldn't do that because for each call at this function, we'll create a new object. So this I why, I was calling the Softmax into the init.

theano/tensor/nnet/nnet.py
```diff
@@ -746,17 +877,34 @@ def local_logsoftmax(node):
...
+            elif isinstance(node.inputs[0].owner.op, list_classes):
+                subtensor_op = node.inputs[0].owner.op
+                subtensor_input = node.inputs[0].owner.inputs[0]
+                subtensor_inputs = node.inputs[0].owner.inputs[1:]
+                if subtensor_input.owner is not None and isinstance(subtensor_input.owner.op, Softmax):
+                    softmax_op = subtensor_input.owner.op
+                    softmax_input = subtensor_input.owner.inputs[0]
+                    # If we have a matrix, we still use the
+                    # CrossentropySoftmaxArgmax1HotWithBias
+                    if softmax_input.type.ndim != 2:
```
bordesf: Yeah, if I don't specify this, the Cross Entropy optimization isn't applied. I will investigate.

shawntan:
Alright, so @ReyhaneAskari and I spoke, I think my current task is to simply write tests for `T_LogSoftmax` that deal with dimensions that are != 2 now that @bordesf has extended softmax to work with more than just matrices.

shawntan:
The logsoftmax doesn't seem to be working for bigger values, not sure why.

nouiz:
note, travis have error from this PR.

bordesf:
Yes, I am currently correcting it.

bordesf:
I think this PR is done.
However, there are two others things that should be done:
- After speaking with @nouiz  we will add a lift of softmax i.e. replace log(sofmax(x).dimshuffle) by log(softmax(x.dimshuffle)) same for reshape. I prefer leave this for another PR. 
- Maybe we could detect the case where we have reshape(log(softmax(x.reshape))) where the last reshape shape the new tensor to the original shape of x (This was something done when softmax was only supporting 2D cases). I am not sure that this optimization is really needed since it will be essentially used in case of old theano code.

I think that it could be great if tests could be reviewed by different peoples and if we could test this PR with DL models for classification and compare the results and performance with the current theano master. Note that this PR shouldn't change anything since the 2D case should still be used and optimized by ops that are unchanged in this PR.

nouiz:
I don't have time to review this big PR before we finish the release of
Theano.

Note, the missing optimization is:

dimshuffle(softmax(x)) -> softmax(dimshuffle(x)) at the canonicalization
step. We should always do it, even if there is no log.

bordesf:
Yes, I will add this optimization in the next todo list PR (Where we can choose the axis). In fact, it will not even make sense to add this optimization right now because, if we have a dimshuffle(softmax(x)) -> softmax(dimshuffle(x)), it means that we have to change the axis on which is applied the softmax. This current PR doesn't allows that (Since it considers only the case where axis == -1).
I will try to resume the next things to do and start the new PR in the next days.

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -255,14 +293,109 @@ def test_local_softmax_optimization(self):
...
+            g = T.grad(logsm, x)
+            fgraph = gof.FunctionGraph([x], [g])
+            theano.compile.mode.optdb.query(theano.compile.mode.OPT_FAST_RUN).optimize(fgraph)
+            assert softmax_grad not in [n.op for n in fgraph.toposort()]
+            utt.verify_grad(f, [np.random.rand(*shape)])
+
+            # Case 1: Log(Subtensor(softmax(x)))
+
+            def f(a):
+                return tensor.log(tensor.nnet.softmax(a)[3])
```
bordesf: I found an issue in this PR, when we have tensor.log(tensor.nnet.softmax(a)[:, 3]), i.e we get a subtensor on the column instead of the row, we got wrong gradients when we check them with the finite difference methods.
I printed the differences for an example and with the finite differences method, I get:
```
[array([[-0.23060371, -0.15813434, -0.2098594 ,  0.97880861, -0.38021126],
       [-0.3059443 , -0.30146484, -0.21215589,  1.04132029, -0.22175533],
       [-0.1581472 , -0.31918843, -0.27057473,  0.91833941, -0.17042915],
       [-0.27672931, -0.18285466, -0.22751856,  0.92796999, -0.24086756],
       [-0.27860693, -0.27760008, -0.18710686,  1.18787938, -0.44456558]])]
```
And with my optimization, I have:
```
[array([[ 0.        ,  0.        ,  0.        ,  0.01876505,  0.        ],
       [ 0.        ,  0.        ,  0.        , -0.09793607,  0.        ],
       [ 0.        ,  0.        ,  0.        , -0.22800455,  0.        ],
       [ 0.        ,  0.        ,  0.        , -0.33536798,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.28916284,  0.        ]])]
```

I think, I will have to check the maths to see if the gradients is effectively defined for all the matrix (even if intuitively it should be defined only on the subtensor and be 0 for the rest).

ReyhaneAskari: I just check the math. I think there is a problem with the optimization. It should result in non zero gradients for the other columns. The reason is that for example, softmax(a)[0,3] is a function of a[0,:]. I'm going to check the code too.

theano/tensor/nnet/nnet.py
```diff
@@ -750,21 +899,53 @@ def c_code_cache_version():
...
+            if isinstance(node.inputs[0].owner.op, Softmax):
+                softmax_input = node.inputs[0].owner.inputs[0]
+                axis = node.inputs[0].owner.op.axis
+                ret = LogSoftmax(axis)(softmax_input)
+                ret.tag.values_eq_approx = values_eq_approx_remove_inf
+                copy_stack_trace([node.inputs[0], node.outputs[0]], ret)
+                return [ret]
+
+            if isinstance(node.inputs[0].owner.op, tensor.elemwise.DimShuffle) and isinstance(node.inputs[0].owner.inputs[0].owner.op, list_classes):
+                node = node.inputs[0].owner
```
ReyhaneAskari: I'm wondering, aren't you completely removing the DimShuffle node here and on line 929?

theano/tensor/nnet/nnet.py
lamblin: We realized when timing similar Ops that caching them did not really make a difference.

ReyhaneAskari: @bordesf could you also add tests for that?

theano/tensor/nnet/nnet.py
```diff
@@ -774,34 +955,144 @@ def local_logsoftmax(node):
...
+            # Compute grad
+            ret = grads - tensor.sum(grads, axis=axis, keepdims=True) * sm
+            ret.tag.values_eq_approx = values_eq_approx_remove_nan
+            copy_stack_trace(node.outputs[0], ret)
+            return [ret]
+
+        if isinstance(d_sm.owner.op, list_classes):
+            for i, input in enumerate(d_sm.owner.inputs):
+                if input.owner and isinstance(input.owner.op, list_classes_IncSubtensor):
+                    d_sm = input
```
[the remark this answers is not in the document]
ReyhaneAskari: @bordesf could also add tests for these cases?

theano/tensor/nnet/nnet.py
bordesf: I did this because those dimshuffle are canceling each other. They happens in order to be managed by the subtensor op which can be used only on a specific dimension. It looks like dimshuffle(subtensor(dimshuffle(x)). So I though that, since I am creating a new subtensor, it will apply back the dimshuffle himself.

bordesf:
I think, I will simply drop cases where Dimshuffle appeared in the graph due to special subtensors. (As I did for the reshape one). I would rather like focus on the gpu optimization. What do you think ? @lamblin @ReyhaneAskari

ReyhaneAskari:
Yeah I agree. It's not worth the effort. Thanks.

lamblin:
I agree as well.

ReyhaneAskari:
@nouiz is there a problem with circleci ?

theano/tensor/nnet/tests/test_nnet.py
```diff
@@ -166,103 +194,240 @@ def test_infer_shape(self):
...
-        x, y = tensor.matrices('xy')
-        sm = tensor.nnet.softmax(x)
+            # Case 2: Log(Advanced_Subtensor1(softmax(x)))
+            x = T.TensorType(dtype=config.floatX, broadcastable=(False,) * d)('x')
+            sm = tensor.nnet.softmax(x)[:, np.random.randint(10, size=3)]
+            logsm = tensor.log(sm)
+            f = theano.function([x], logsm)
+            printing.debugprint(f)
+            assert isinstance(f.maker.fgraph.outputs[0].owner.inputs[0].owner.op, theano.tensor.nnet.nnet.LogSoftmax)
+
```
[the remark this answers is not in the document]
ReyhaneAskari: @bordesf could you also add the cases that Pascal has mentioned?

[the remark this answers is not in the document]
bordesf: We said that we don't manage the case of Dimshuffle and reshape. So, there will be not optimization to LogSoftmax in those case. I don't think there is a need to test that but maybe I am wrong

PR #6138, opened 2017-07-11
Repack the pickle with protocol 2 and open in binary mode.

abergeron:
This should make the test work across python 2 and 3 on windows and linux.


lamblin:
Failure in travis is unrelated.

Issue #2072, opened 2014-09-02
Meta Optimizers


f0k:
Theano now provides access to several different convolution implementations that perform very differently depending on image and kernel shapes, and depending on the direction when used in training a CNN (forward pass, gradient wrt. weights, gradient wrt. inputs), but also have different restrictions (filter shapes, memory usage).
A lot could be gained by picking the best-performing implementation for each individual convolution in a CNN: https://github.com/soumith/convnet-benchmarks
Because the relative performance of different implementations heavily depends on the GPU used, it is impossible to select the optimal implementation at compile time via some clever heuristic.

So I would like to start a discussion on adding "meta optimizers" to the Theano graph optimizers. A meta optimizer would collect a set of possible replacements for an op, compile each of them, time their performance and then select the best-performing one.

This might be implemented in the form of a regular graph optimizer, but it should have the following features:
- It should operate on a list of op replacement functions that each return a replacement for a given op or a falseish value to indicate that they are not applicable -- similar to what the current graph optimizers do anyway.
- It should be possible to register additional op replacement functions with a given meta optimizer. For example, user code may want to register a wrapper for cuda-convnet, cuda-convnet2 or some other Theano-external implementation with Theano's conv2d meta optimizer.
- It should be able to cache its decisions, so that it does not need to re-run any benchmarks if it is repeatedly asked to find a replacement for a given op and device with a given set of registered replacement functions. (This cache should be as permanent as the compilation cache.)

I'm currently interested in GPU-based conv2d only. In this case the meta optimizer could spring into action for any `GpuConv` op with fully specified shape on compile time, by compiling and running a set of proposed replacements on random dummy data. Other use cases for meta-optimization may require more realistic data -- they could hijack the test value computation, for example.

Sooo...
1. What do you think of the idea in general?
2. What would it need to implement the features listed above for a GpuConv meta optimizer?
3. Should the implementation plan for possible future meta optimizers or not care about generality/reusability for now?

/Update Nov 24: A first implementation was done in #2266. TODOs left:
- [ ] The meta-optimizer currently tries all available convolutions, it should only try the ones selected via `optimizer_including` and `optimizer_excluding`
- [ ] For gemm, it should try both GpuCorrMM and GpuCorrMM_gradWeights if both are applicable, not rely on the heuristic in local_conv_gemm
- [ ] Timings should be cached on disk to speed up repeated invocations.
[...]

Issue #5713, opened 2017-03-14
3D convolution implementation in 0.9rc


mpslxz:
I was testing 3D convolution in Theano as well as Keras (with Theano backend) in a very simple setup. Surprisingly, the Keras version of the same architecture was way faster than Theano. I had a conversation with @nouiz  ([here](https://groups.google.com/forum/#!topic/theano-users/Lnbets1YlqY)) and he suggested profiling the code. I figured out that Keras is using CorrMM while Theano uses Corr3dMM when tensor.nnet.conv3d is called. Fred is suggesting that Corr3dMM is inefficient, so I am opening up this issue.
Attached is the code which reproduces my results. 
* My Keras version is **1.0.4.**
* This is how my .theanorc looks like:
`[global]
floatX = float32
assert_no_cpu_op = warn
device = cpu
openmp = True
[cuda]
root = /usr/local/cuda-7.5
`

[theano_profile.txt](https://github.com/Theano/Theano/files/842798/theano_profile.txt)
[keras_profile.txt](https://github.com/Theano/Theano/files/842799/keras_profile.txt)
[issue_code.txt](https://github.com/Theano/Theano/files/842803/issue_code.txt)

nouiz:
Thanks for the information. I see the problems. If you use a more recent keras, I think you will get the slower version.

The old keras use the conv3d2d() function in Theano. You can use it too. It end up being faster. There is no optimization you can enable to have it used automatically, but if you use it manually, it will work.

To be sure to always have the fastest implementation between corr3d and conv3d2d, it would request to revive the MetaOptimizer: https://github.com/Theano/Theano/issues/2072. 

So for now, use conv3d2d manually, but if the shapes changes, it isn't sure to be faster.

With intel Theano branch, it could be even faster if CPU speed is important for you: https://github.com/intel/Theano

We don't support it, but it will probably be merged back in Theano at some point.

mpslxz:
Thanks for your reply, Fred. I looked at the source of conv3d2d and it seems that there is no convolution along the time axis. Because if there were, there should be a reshape along time and conv2d again to take into account the dependencies along time axis. I believe with this implementation, the receptive field does not extend into time. I guess theoretically, conv3d2d is faster in all cases regardless of the input shape(what do you think?)
I like the idea of MetaOptimizer. But the problem with that would be these subtle differences in algorithms which potentially could yield different results.

nouiz:
conv3d2d cause extra computation that won't be used. So in theory, it
[...]

PR #6143, opened 2017-07-12
Implement Meta-optimizer on new backend

affanv14:
related to https://github.com/Theano/Theano/issues/2072
Task list:
- [x] Port Meta-optimizer on new backend 
- [x] change verbose config flag to int
- [x] add alternative optimizers for GPUCorrMM for conv2d
- [x] add alternative optimizers for cudnn for conv2d  
- [x] add conv3d2d optimizer 
- [x] add alternative optimizers for GPUCorrMM for conv3d
- [x] add alternative optimizers for cudnn for conv3d
- [x] add tests to demonstrate proper working for all optimisers 
- [x] modify implementation to only select implementations using optimizer_including and optimizer_excluding

### DIFF
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1185,9 +1185,9 @@ class LocalMetaOptimizer(LocalOptimizer):
         # compile the resulting subgraphs and time their execution
         if self.verbose > 1:
             print(("%s meta-optimizing %s (%d choices):" %
-                   (self.__class__.__name__, node, len(self.optimizers))))
+                   (self.__class__.__name__, node, len(self.track_dict[type(node.op)]))))
         timings = []
-        for opt in (self.track_dict[type(node.op)] + self.track_dict[node.op]):
+        for opt in self.track_dict[type(node.op)]:
             outputs = opt.transform(node)
             if outputs:
                 try:
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1131,13 +1131,20 @@ class LocalMetaOptimizer(LocalOptimizer):
 
     """
 
-    def __init__(self, tracks=None, optimizers=()):
-        self._tracks = tracks
+    def __init__(self, optimizers=()):
+        self._tracks = [x for o in optimizers for x in o.tracks()]
         self.optimizers = list(optimizers)
         self.verbose = config.metaopt.verbose
+        self.track_dict = defaultdict(lambda: [])
+
+        for o in optimizers:
+            for c in o.tracks():
+                self.track_dict[c].append(o)
 
     def register(self, optimizer):
         self.optimizers.append(optimizer)
+        for c in optimizer.tracks():
+            self.track_dict[c].append(optimizer)
 
     def tracks(self):
         return self._tracks
@@ -1178,7 +1185,7 @@ class LocalMetaOptimizer(LocalOptimizer):
             print(("%s meta-optimizing %s (%d choices):" %
                    (self.__class__.__name__, node, len(self.optimizers))))
         timings = []
-        for opt in self.optimizers:
+        for opt in (self.track_dict[type(node.op)] + self.track_dict[node.op]):
             outputs = opt.transform(node)
             if outputs:
                 try:
@@ -2313,7 +2320,6 @@ class EquilibriumOptimizer(NavigatorOptimizer):
         self.final_optimizers = []
         self.cleanup_optimizers = []
         self.tracks_on_change_inputs = tracks_on_change_inputs
-
         for opt in optimizers:
             if isinstance(opt, LocalOptimizer):
                 if opt.tracks() is None:
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1131,12 +1131,12 @@ class LocalMetaOptimizer(LocalOptimizer):
 
     """
 
-    def __init__(self, optimizers=()):
-        self._tracks = [x for o in optimizers for x in o.tracks()]
-        self.optimizers = list(optimizers)
+    def __init__(self):
         self.verbose = config.metaopt.verbose
         self.track_dict = defaultdict(lambda: [])
         self.tag_dict = defaultdict(lambda: [])
+        self._tracks = []
+        self.optimizers = []
 
     def register(self, optimizer, tag_list):
         self.optimizers.append(optimizer)
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1134,17 +1134,19 @@ class LocalMetaOptimizer(LocalOptimizer):
     def __init__(self, optimizers=()):
         self._tracks = [x for o in optimizers for x in o.tracks()]
         self.optimizers = list(optimizers)
-        self.verbose = config.metaopt.verbose
+        self.verbose = True
         self.track_dict = defaultdict(lambda: [])
 
         for o in optimizers:
             for c in o.tracks():
                 self.track_dict[c].append(o)
 
-    def register(self, optimizer):
-        self.optimizers.append(optimizer)
-        for c in optimizer.tracks():
-            self.track_dict[c].append(optimizer)
+    def register(self, optimizers):
+        self.optimizers.extend(optimizers)
+        for o in optimizers:
+            for c in o.tracks():
+                self.track_dict[c].append(o)
+                self._tracks.append(c)
 
     def tracks(self):
         return self._tracks
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1191,7 +1191,8 @@ class LocalMetaOptimizer(LocalOptimizer):
                 try:
                     fn = theano.function([], outputs, givens=givens,
                                          on_unused_input='ignore')
-                    timing = min(self.time_call(fn) for _ in range(3))
+                    fn.trust_input = True
+                    timing = min(self.time_call(fn) for _ in range(2))
                 except Exception as e:
                     if self.verbose > 0:
                         print("* %s: exception" % opt, e)
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1137,10 +1137,6 @@ class LocalMetaOptimizer(LocalOptimizer):
         self.verbose = config.metaopt.verbose
         self.track_dict = defaultdict(lambda: [])
 
-        for o in optimizers:
-            for c in o.tracks():
-                self.track_dict[c].append(o)
-
     def register(self, optimizers):
         self.optimizers.extend(optimizers)
         for o in optimizers:
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1134,7 +1134,7 @@ class LocalMetaOptimizer(LocalOptimizer):
     def __init__(self, optimizers=()):
         self._tracks = [x for o in optimizers for x in o.tracks()]
         self.optimizers = list(optimizers)
-        self.verbose = True
+        self.verbose = config.metaopt.verbose
         self.track_dict = defaultdict(lambda: [])
 
         for o in optimizers:
@@ -1176,14 +1176,14 @@ class LocalMetaOptimizer(LocalOptimizer):
             missing.difference_update(givens.keys())
         # ensure we have data for all input variables that need it
         if missing:
-            if self.verbose:
+            if self.verbose > 0:
                 print(("%s cannot meta-optimize %s, "
                        "%d of %d input shapes unknown" %
                        (self.__class__.__name__, node, len(missing), node.nin)))
             return
         # now we can apply the different optimizations in turn,
         # compile the resulting subgraphs and time their execution
-        if self.verbose:
+        if self.verbose > 1:
             print(("%s meta-optimizing %s (%d choices):" %
                    (self.__class__.__name__, node, len(self.optimizers))))
         timings = []
@@ -1195,20 +1195,20 @@ class LocalMetaOptimizer(LocalOptimizer):
                                          on_unused_input='ignore')
                     timing = min(self.time_call(fn) for _ in range(3))
                 except Exception as e:
-                    if self.verbose:
+                    if self.verbose > 0:
                         print("* %s: exception" % opt, e)
                     continue
                 else:
-                    if self.verbose:
+                    if self.verbose > 1:
                         print("* %s: %.5g sec" % (opt, timing))
                     timings.append((timing, outputs, opt))
             else:
-                if self.verbose:
+                if self.verbose > 0:
                     print("* %s: not applicable" % opt)
         # finally, we choose the fastest one
         if timings:
             timings.sort()
-            if self.verbose:
+            if self.verbose > 1:
                 print("= %s" % timings[0][2])
             return timings[0][1]
         return
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1136,13 +1136,15 @@ class LocalMetaOptimizer(LocalOptimizer):
         self.optimizers = list(optimizers)
         self.verbose = config.metaopt.verbose
         self.track_dict = defaultdict(lambda: [])
+        self.tag_dict = defaultdict(lambda: [])
 
-    def register(self, optimizers):
-        self.optimizers.extend(optimizers)
-        for o in optimizers:
-            for c in o.tracks():
-                self.track_dict[c].append(o)
-                self._tracks.append(c)
+    def register(self, optimizer, tag_list):
+        self.optimizers.append(optimizer)
+        for c in optimizer.tracks():
+            self.track_dict[c].append(optimizer)
+            self._tracks.append(c)
+        for tag in tag_list:
+            self.tag_dict[tag].append(optimizer)
 
     def tracks(self):
         return self._tracks
@@ -1181,9 +1183,9 @@ class LocalMetaOptimizer(LocalOptimizer):
         # compile the resulting subgraphs and time their execution
         if self.verbose > 1:
             print(("%s meta-optimizing %s (%d choices):" %
-                   (self.__class__.__name__, node, len(self.track_dict[type(node.op)]))))
+                   (self.__class__.__name__, node, len(self.get_opts(node)))))
         timings = []
-        for opt in self.track_dict[type(node.op)]:
+        for opt in self.get_opts(node):
             outputs = opt.transform(node)
             if outputs:
                 try:
@@ -1217,6 +1219,12 @@ class LocalMetaOptimizer(LocalOptimizer):
 
         """
         raise NotImplementedError()
+
+    def get_opts(self, node):
+        """
+        Can be overrided to change the way opts are selected
+        """
+        return self.track_dict[type(node.op)]
 
     def time_call(self, fn):
         start = time.time()
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -2915,9 +2915,11 @@ def local_abstractconv_cudnn_alternative(node):
         conv_mode = 'cross'
 
     if isinstance(op, AbstractConv2d):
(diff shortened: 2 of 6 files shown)


affanv14:
Currently only the forward pass is done

theano/gpuarray/opt.py
```diff
@@ -1726,6 +1728,66 @@ def local_abstractconv3d_gradinputs_gemm(node):
...
+
+        shapes = (node.op.imshp, node.op.kshp)
+        if(node.op.imshp is None or node.op.kshp is None or
+                any([s is None for shape in shapes for s in shape])):
+            return result
+
+        if type(node.op) in [AbstractConv2d, AbstractConv3d]:
+            img, kern = node.inputs
+            for(var, shape) in zip((img, kern), shapes):
+                result[var] = theano.shared(np.random.random(shape).astype(theano.config.floatX),
```
f0k: Should be `astype(var.dtype)`, I suppose.

theano/gpuarray/opt.py
```diff
@@ -1726,6 +1728,66 @@ def local_abstractconv3d_gradinputs_gemm(node):
...
+            tshp = get_conv_output_shape(node.op.imshp,
+                                         node.op.kshp,
+                                         node.op.border_mode,
+                                         node.op.subsample,
+                                         node.op.filter_dilation)
+
+            result[kshape] = theano.tensor.as_tensor_variable(node.op.kshp[-2:])
+
+            for(var, shape) in zip((img, top), (node.op.imshp, tshp)):
+                result[var] = theano.shared(np.random.random(shape).astype(theano.config.floatX),
```
f0k: Same here.

theano/gpuarray/opt.py
```diff
@@ -1726,6 +1728,66 @@ def local_abstractconv3d_gradinputs_gemm(node):
...
+            tshp = get_conv_output_shape(node.op.imshp,
+                                         node.op.kshp,
+                                         node.op.border_mode,
+                                         node.op.subsample,
+                                         node.op.filter_dilation)
+
+            result[ishape] = theano.tensor.as_tensor_variable(node.op.imshp[-2:])
+
+            for(var, shape) in zip((kern, top), (node.op.kshp, tshp)):
+                result[var] = theano.shared(np.random.random(shape).astype(theano.config.floatX),
```
f0k: Same here.

theano/gpuarray/opt.py
```diff
@@ -2219,7 +2286,7 @@ def local_gpu_svd(op, context_name, inputs, outputs):
                                'conv_gemm',
                                'gpuarray', 'fast_compile', 'fast_run')
 
-
+conv_metaopt.register(abstractconv_groupopt.query(*['+' + name for name in abstractconv_groupopt._names]).opts)
```
f0k: Won't this register the meta-optimizer as an optimizer for itself? I think I'd move the `abstractconv_groupopt.register('conv_metaopt', ...)` to after this line. See my original code: https://github.com/Theano/Theano/blob/0.8.X/theano/sandbox/cuda/opt.py#L1728-L1735

theano/gpuarray/opt.py
```diff
@@ -1726,6 +1728,66 @@ def local_abstractconv3d_gradinputs_gemm(node):
...
+        shapes = (node.op.imshp, node.op.kshp)
+        if(node.op.imshp is None or node.op.kshp is None or
+                any([s is None for shape in shapes for s in shape])):
+            return result
+
+        if type(node.op) in [AbstractConv2d, AbstractConv3d]:
+            img, kern = node.inputs
+            for(var, shape) in zip((img, kern), shapes):
+                result[var] = theano.shared(np.random.random(shape).astype(theano.config.floatX),
+                                            var.name, borrow=True)
```
f0k: And you're missing the broadcast pattern. See https://github.com/Theano/Theano/blob/0.8.X/theano/sandbox/cuda/opt.py#L1715-L1724 for my original code, I guess you can copy this verbatim.

theano/gof/opt.py
```diff
@@ -1132,13 +1132,22 @@ class LocalMetaOptimizer(LocalOptimizer):
 
     """
 
-    def __init__(self, tracks=None, optimizers=()):
-        self._tracks = tracks
+    def __init__(self, optimizers=()):
+        self._tracks = [x for o in optimizers for x in o.tracks()]
         self.optimizers = list(optimizers)
-        self.verbose = config.metaopt.verbose
+        self.verbose = True
```
f0k: The verbosity should be configurable, though.

theano/gpuarray/opt.py
affanv14: Yes, I was planning to remove it later when I  modify it to support optimizer_including and optimizer_excluding.

theano/gpuarray/opt.py
```diff
@@ -2182,6 +2247,11 @@ def local_gpu_svd(op, context_name, inputs, outputs):
 # to avoid a circular dependency problem with dnn
 from .dnn import (local_abstractconv_cudnn, local_abstractconv_gw_cudnn,
                   local_abstractconv_gi_cudnn)     # noqa: 402
+
+conv_metaopt = ConvMetaOptimizer()
+abstractconv_groupopt.register('conv_metaopt', conv_metaopt, 0, 'conv_meta')
```
affanv14: small question, the number argument in register is used for the position right?
because while I was going through the code in the method `register` in `LocalGroupdb` I found that the number argument is being taken as a tag.

affanv14: If I register the metaoptimizer before the registering others, it works. But If I register at the end with the integer argument as 0 or even 10. It does not work. 
But if i explicitly add position=10(or anything lesser) to the register function after registering everything else, it works

affanv14:
I dont think this is the way to go for adding support to optimizer_including and optimizer_excluding. I have a better idea in mind which may work. I'll have that out in a couple of days

theano/gpuarray/opt.py
f0k: Interesting... in Theano 0.8, `LocalGroupDB` is a subclass of `SequenceDB`, which has [`def register(self, name, obj, position, *tags)`](https://github.com/Theano/Theano/blob/0.8.X/theano/gof/optdb.py#L311).
In Theano 0.9, `LocalGroupDB` overrides this method as [`def register(self, name, obj, *tags, **kwargs):`](https://github.com/Theano/Theano/blob/0.9.X/theano/gof/optdb.py#L417). While it still passes the positional arguments to `SequenceDB.register()` (including the position), it also overrides the `query` method in a way that only `position=...` keyword arguments to `register` are considered.
I don't know if the interface was broken on purpose, but yes, it means you will have to use `position=0` for it to work.

theano/gof/opt.py
```diff
@@ -1131,13 +1131,22 @@ class LocalMetaOptimizer(LocalOptimizer):
 
     """
 
-    def __init__(self, tracks=None, optimizers=()):
-        self._tracks = tracks
+    def __init__(self, optimizers=()):
+        self._tracks = [x for o in optimizers for x in o.tracks()]
         self.optimizers = list(optimizers)
         self.verbose = config.metaopt.verbose
+        self.track_dict = defaultdict(lambda: [])
```
abergeron: Any optimizers specified in `__init__` will not be added to `track_dict`.  Perhaps you should just setup minimal things and call register() on the passed in optimizers.

abergeron:
Apart from the small issue in `__init__`, the rest looks good.

theano/gof/opt.py
f0k: Yes, they are in the following lines. But initializing everything to be empty and then calling `self.register(optimizers)` would reduce code duplication / possible errors.

abergeron: Ah you're right, I was confused by the register function and the way the diff is presented.

But yeah it would help reduce duplication to do it that way.

affanv14:
I would also need to write tests to demonstrate that all optimizers are working properly. right?

abergeron:
Yes, that would be best.

abergeron:
Also regarding the task list I don't think you need to do anything special for conv3d2d, since it will just be a bunch of 2d convolutions with reshapes, so the regular 2d path can handle it.

affanv14:
well , I just tested the output and experienced a convolution inception.
```Using cuDNN version 5110 on context None
Mapped name None to device cuda0: GRID K520 (0000:00:03.0)
ConvMetaOptimizer meta-optimizing AbstractConv3d{convdim=3, border_mode='valid', subsample=(1, 1, 1), filter_flip=True, imshp=(10, 12, 5, 5, 5), kshp=(11, 12, 3, 3, 3), filter_dilation=(1, 1, 1), num_groups=1}(GpuFromHost<None>.0, GpuFromHost<None>.0) (3 choices):
* local_abstractconv_cudnn: 7.2956e-05 sec
* local_abstractconv3d_gemm: 0.00053787 sec
ConvMetaOptimizer cannot meta-optimize AbstractConv2d{convdim=2, border_mode='valid', subsample=(1, 1), filter_flip=True, imshp=(None, None, None, None), kshp=(None, None, None, None), filter_dilation=(1, 1), num_groups=1}(GpuReshape{4}.0, GpuReshape{4}.0), 2 of 2 input shapes unknown
ConvMetaOptimizer cannot meta-optimize AbstractConv2d{convdim=2, border_mode='valid', subsample=(1, 1), filter_flip=True, imshp=(None, None, None, None), kshp=(None, None, None, None), filter_dilation=(1, 1), num_groups=1}(GpuReshape{4}.0, GpuReshape{4}.0), 2 of 2 input shapes unknown
ConvMetaOptimizer cannot meta-optimize AbstractConv2d{convdim=2, border_mode='valid', subsample=(1, 1), filter_flip=True, imshp=(None, None, None, None), kshp=(None, None, None, None), filter_dilation=(1, 1), num_groups=1}(GpuReshape{4}.0, GpuReshape{4}.0), 2 of 2 input shapes unknown
ConvMetaOptimizer cannot meta-optimize AbstractConv2d{convdim=2, border_mode='valid', subsample=(1, 1), filter_flip=True, imshp=(None, None, None, None), kshp=(None, None, None, None), filter_dilation=(1, 1), num_groups=1}(GpuReshape{4}.0, GpuReshape{4}.0), 2 of 2 input shapes unknown
* local_abstractconv3d2d: 0.00028896 sec
= local_abstractconv_cudnn
```

affanv14:
well I got the shapes working as well, does this look alright?
the output  becomes a little bit muddled though because it contains metaoptimizations for both 3d and 2d(due to conv3d2d)

```
ConvMetaOptimizer meta-optimizing AbstractConv3d{convdim=3, border_mode='valid', subsample=(1, 1, 1), filter_flip=True, imshp=(10, 12, 5, 5, 5), kshp=(11, 12, 3, 3, 3), filter_dilation=(1, 1, 1), num_groups=1}(GpuFromHost<None>.0, GpuFromHost<None>.0) (3 choices):
* local_abstractconv_cudnn: 8.5115e-05 sec
* local_abstractconv3d_gemm: 0.00052404 sec
ConvMetaOptimizer meta-optimizing AbstractConv2d{convdim=2, border_mode='valid', subsample=(1, 1), filter_flip=True, imshp=(50, 12, 5, 5), kshp=(33, 12, 3, 3), filter_dilation=(1, 1), num_groups=1}(GpuReshape{4}.0, GpuReshape{4}.0) (4 choices):
* local_abstractconv_cudnn: 7.9155e-05 sec
* local_abstractconv_gemm: 0.00033808 sec
* local_abstractconv_gemm_alternative: 0.00033402 sec
* local_abstractconv_cudnn_alternative: 0.00010204 sec
= local_abstractconv_cudnn
* local_abstractconv3d2d: 0.00027609 sec
= local_abstractconv_cudnn

```

abergeron:
I don't think it's worth it to try to convert the 3d convolutions to conv3d2d in the meta-optimizer by default.  It is mostly a relic from the time when we did not have any good 3d convolution implementations.

However it might be a good idea to leave the option open for people that want to try it, if you already have the code.

affanv14:
Oh, i did it because it was given in the task list here https://github.com/Theano/Theano/issues/2072#issue-41718396
So while implementing optimizer_including and optimizer_excluding functionality again, i will make it so that it will be off by default and will only run  if it is included in optimizer_including. Will that be alright?

affanv14:
I had to explicitly specify tags for optimizer_excluding and optimizer_including to work

since all convolutions work in both fastcompile and fastrun, I used the tag 'default'. Or would it be better to use fastcompile and fastrun itself.

the tag 'alternative' can be used to exclude all alternative implementations
for conv3d2d, it is excluded by default, and can be included if we add the 'conv3d2d' tag

f0k:
> experienced a convolution inception for conv3d2d

That's fun :) Nice you got it working, although cuDNN is way ahead of conv3d2d in your example and might always be.

> Oh, i did it because it was given in the task list here

This references #5713, in which conv3d2d is faster than corrMM on the CPU. So it seems conv3d2d is still useful. I think conv3d2d should not be enabled by default since it unnecessarily slows down meta-optimization for GPU. But wherever we recommend using the meta-optimizer, we can tell that for 3D convolutions on CPU, it may be worth enabling conv3d2d as well (since users will have to change their configuration to enable the meta-optimizer anyway).

Good job so far!

affanv14:
@abergeron, I have finished making all changes required. Anything else to add?

theano/gof/opt.py
```diff
@@ -1131,13 +1131,20 @@ class LocalMetaOptimizer(LocalOptimizer):
...
+        self.tag_dict = defaultdict(lambda: [])
```
abergeron: Now this code will not fill in tag_dict and track_dict for initial optimizers.

Since it appears that this is not used anywhere, can you just get rid of the `optimizers` argument in `__init__`.

theano/gof/opt.py
```diff
@@ -1131,13 +1131,18 @@ class LocalMetaOptimizer(LocalOptimizer):
 
     """
 
-    def __init__(self, tracks=None, optimizers=()):
-        self._tracks = tracks
-        self.optimizers = list(optimizers)
+    def __init__(self):
         self.verbose = config.metaopt.verbose
```
abergeron: You still need to create an empty `_tracks` list otherwise the append in register() will crash.

You also need an empty `optimizers` list for the same reason.

affanv14: sorry, I was being hasty there. I should have double checked.

affanv14: I amended the commit and made all the necessary changes

abergeron:
jenkins test this please

affanv14:
I had forgotten to do the rebase to incorporate changes in the other PR. But that doesn't solve the reason why it is failing in jenkins. These tests do pass for me, its probably a graph traversal problem.

affanv14:
This seems to work as well.

abergeron:
We'll see: jenkins test this please

f0k:
Seems that `optimizer.transform(conv_op)` returns `None` here: https://github.com/Theano/Theano/pull/6143/commits/32720deac2ca73dd96671b8a2f36726bd9f842ba#diff-a92f2d52074a47c1cbf443c40653fedeR775

affanv14:
I'm unable to replicate the error though. I did try to use as many of the same theano flags as i could as in jenkins but still couldn't replicate.
There are 3 reasons optimizer.transform is returning None. 
1. It isn't an abstractconv2d instance 
2. it's inputs are not of gpuarray type 
3. the params(border_mode etc.) dont meet the conditions.

it can't be 1 since the method `find_abstractconv` returns the abstractconv instance. It can't even be 3 since it is passing for my system. So it should be 2 but `local_conv_gpu_conv` is used for that sole purpose. Any idea what I could be missing?

affanv14:
I made the selection of optimizers more flexible. this allows us to select the individual implementations we want. For example, if we want only the alternative passes, we can do 
metaopt.optimizer_including='alternative'
metaopt.optimizer_excluding='default'
this was not possible is the previous method. I also did create two more config variables though. Is this alright?

abergeron:
Yes, adding config variables is fine.  They also have reasonable names, so that is good.

abergeron:
Jenkins test this please

abergeron:
All right no capitals: jenkins test this please

affanv14:
I was finally able to replicate the error, it was due to my theanorc file. The init_gpu_device requires device=cpu but in my theanorc file I had device=cuda0. That is why tests were passing for me.
By adding the shape info and the modes, those errors got corrected now.

abergeron:
For the last time I hope: jenkins test this please

affanv14:
I hope so too

theano/gpuarray/dnn.py
```diff
@@ -2888,6 +2888,215 @@ def local_abstractconv_cudnn(node):
...
+
+        rval = dnn_conv(inp1, inp2,
+                        border_mode=border_mode,
+                        subsample=subsample,
+                        dilation=filter_dilation,
+                        direction_hint=direction_hint,
+                        conv_mode=conv_mode,
+                        num_groups=num_groups)
+
+    if isinstance(op, AbstractConv2d_gradWeights):
```
nouiz: make an elif to make it easier to read.

theano/gpuarray/dnn.py
```diff
@@ -2888,6 +2888,215 @@ def local_abstractconv_cudnn(node):
...
+
+            conv = GpuDnnConv(algo=None, num_groups=num_groups)(img, topgrad, out, desc)
+            if conv_mode == 'conv':
+                conv = conv[:, :, ::-1, ::-1]
+
+            rval = as_gpuarray_variable(conv.dimshuffle(1, 0, 2, 3), ctx_name)
+        else:
+            return None
+
+    if isinstance(op, AbstractConv2d_gradInputs):
```
nouiz: elif

theano/gpuarray/opt.py
```diff
@@ -1694,6 +1740,74 @@ def local_abstractconv3d_gemm(node):
     return [rval]
 
 
+@local_optimizer([AbstractConv3d])
+def local_abstractconv3d_alt(node):
```
nouiz: use the same name everywhere alt or alternative

theano/gpuarray/opt.py
```diff
@@ -1716,6 +1830,70 @@ def local_abstractconv_gradweights_gemm(node):
     return [rval]
 
 
+@local_optimizer([AbstractConv2d_gradWeights])
+def local_abstractconv_gemm_gradweights_alt(node):
```
nouiz: there seem to be much code duplicate between the 2d and 3d case and cudnn and gemm alternative.

Do you think it is easy to refactor?

theano/gof/opt.py
```diff
@@ -1167,39 +1174,39 @@ def transform(self, node):
...
+                   (self.__class__.__name__, node, len(self.get_opts(node)))))
         timings = []
-        for opt in self.optimizers:
+        for opt in self.get_opts(node):
             outputs = opt.transform(node)
             if outputs:
                 try:
                     fn = theano.function([], outputs, givens=givens,
                                          on_unused_input='ignore')
                     timing = min(self.time_call(fn) for _ in range(3))
```
nouiz: In the new back-end, and with cudnn conv "time_*" mode, the first call will be much slower. As there is 3 call, I'm not sure it is worth to make another one before, but in reality, there will probably be only the 2 second that will give sensible timing.

Also, to help distinguish more the difference, you could some of Theano function overhead by doing:
fn.trust_input = True
before using it.

theano/gpuarray/opt.py
affanv14: About this, it should be easy to do but it might become a little harder to read since there will be more if else conditions and also subsamples, filter_dilations, filter_flips and different ops  need to be handled for both 2d and 3d. Just wanted to check if it is alright before I proceed

abergeron:
I consider the travis failure to be irrelevent.

jenkins test this please

PR #6146, opened 2017-07-12
add windows pr tests

slefrancois:

PR #6147, opened 2017-07-12
Fix dnn loading/usage on mac

abergeron:


abergeron:
This could break things on windows, so we'll have to watch out for that.

slefrancois:
jenkins test this

slefrancois:
It might break things on linux too, see jenkins.

abergeron:
Apparently it does, I'll fallback to a mac-specific fix then.

abergeron:
This is now less likely to break other platforms.

abergeron:
Also is there a way to ask jenkins to test a specific PR on mac and windows?

slefrancois:
not yet, but [windows pr are coming](https://github.com/Theano/Theano/pull/6146)

abergeron:
Nice!

abergeron:
This has now passed.

slefrancois:
We can test once #6149 is merged.

lamblin:
jenkins test mac

f0k:
Just noticed that @275bd60 breaks cuDNN support for me, I get:
```
Can not use cuDNN on context None: cannot compile with cuDNN. We got this error:
/usr/bin/ld: cannot find -lcudnn
collect2: error: ld returned 1 exit status
```
`libcudnn.so` is installed in `/usr/local/cuda/lib64`, which is included in `/etc/ld.conf.so.d/cudatoolkit` so `ldconfig` can find it. What is the rationale behind not defaulting `dnn.library_path` to the CUDA library path any longer?

abergeron:
The rationale was that the packages provided by NVIDIA install in /usr/local.

However it seems that lots of people have cudnn in the cuda directory so that has been restored as part of #6146

f0k:
Hmm, wait, #6146 sets `theano.config.dnn.library_path` to `/usr/local/cuda/lib`, not `/usr/local/cuda/lib64` as before 275bd60. So it still doesn't work for me without manually overriding `dnn.library_path`. Should it default to `lib64` on unix platforms?

abergeron:
Yes, you're right.  Does this work on your system: https://github.com/Theano/Theano/pull/6256?

f0k:
> Does this work on your system:

Yes, it does!

Issue #6133, opened 2017-07-10
Keep a trace of which optimizations don't have stack trace


ReyhaneAskari:
Regarding https://github.com/Theano/Theano/issues/3018, we need a way to keep track of the optimizations that do not have the stack trace. Also, this could be helpful to notify the developers writing new optimization to keep the stack trace. 

- [x] Add a flag for checking the stack trace during the optimization process. Off by default.
- [x] Adding a check that verifies if the stack trace of each optimization is empty or not, each time optimizations are applied, or each time a variable is added to the graph. (This check is only done when the flag is set. The flag should be set in at least a large proportion of unit tests in the buildbots.)
- [x] Update all optimizations so that they at least include a dummy value in the stack trace, for instance something like "No traceback for variable created in optimization %s".
- [x] Update the flag to 3 options: default (Not checking), check_all (checks if the stack trace is added), check_and_skip (check all the stack trace of optimizations but skip the ones with the dummy value mentioned above.)

PR #6150, opened 2017-07-13
Fix stack trace

ReyhaneAskari:
Regarding https://github.com/Theano/Theano/issues/6133, this PR checks which optimizations have empty stack traces. The last step for this ticket is to add a dummy value for the empty stack traces.

### DIFF
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -226,6 +226,15 @@ optdb.register('add_destroy_handler', AddDestroyHandler(),
 optdb.register('merge3', gof.MergeOptimizer(),
                100, 'fast_run', 'merge')
 
+if theano.config.check_stack_trace in ['check_all', 'check_and_skip']:
+    tags = ('fast_run', 'fast_compile')
+
+if theano.config.check_stack_trace == 'not_checking':
+    tags = ()
+
+optdb.register('CheckStackTrace',
+               gof.CheckStackTraceOptimization(), -1, *tags)
+
 
 class Mode(object):
     """
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -226,10 +226,10 @@ optdb.register('add_destroy_handler', AddDestroyHandler(),
 optdb.register('merge3', gof.MergeOptimizer(),
                100, 'fast_run', 'merge')
 
-if theano.config.check_stack_trace in ['check_all', 'check_and_skip']:
+if theano.config.check_stack_trace in ['raise', 'warn', 'log']:
     tags = ('fast_run', 'fast_compile')
 
-if theano.config.check_stack_trace == 'not_checking':
+if theano.config.check_stack_trace == 'off':
     tags = ()
 
 optdb.register('CheckStackTrace',
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -227,13 +227,14 @@ optdb.register('merge3', gof.MergeOptimizer(),
                100, 'fast_run', 'merge')
 
 if theano.config.check_stack_trace in ['raise', 'warn', 'log']:
-    tags = ('fast_run', 'fast_compile')
+    _tags = ('fast_run', 'fast_compile')
 
 if theano.config.check_stack_trace == 'off':
-    tags = ()
+    _tags = ()
 
 optdb.register('CheckStackTrace',
-               gof.CheckStackTraceOptimization(), -1, *tags)
+               gof.CheckStackTraceOptimization(), -1, *_tags)
+del _tags
 
 
 class Mode(object):
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -1511,7 +1511,7 @@ AddConfigVar('check_stack_trace',
              " optimization. check_all checks if the stack trace is added"
              " and check_and_skip checks all the stack trace of optimizations"
              " but skips the ones that have a dummy value ",
-             EnumStr('check_and_skip', 'check_all', 'not_checking'),
+             EnumStr('not_checking', 'check_and_skip', 'check_all'),
              in_c_key=False)
 
 
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -1511,7 +1511,7 @@ AddConfigVar('check_stack_trace',
              " optimization. check_all checks if the stack trace is added"
              " and check_and_skip checks all the stack trace of optimizations"
              " but skips the ones that have a dummy value ",
-             EnumStr('check_all', 'not_checking', 'check_and_skip'),
+             EnumStr('check_and_skip', 'check_all', 'not_checking'),
              in_c_key=False)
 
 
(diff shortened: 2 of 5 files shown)


ReyhaneAskari:
- [ ] Find the list of the optimizations that fail with check_all.
- [ ] For each of them either fix it by adding a stack trace or add the dummy stack trace explicitly inside the optimization. 
- [ ]  Add the flag to the test suit. 
- [ ] Remove the explicit code that checks the stack trace.

theano/gof/opt.py
```diff
@@ -3039,3 +3039,31 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
...
     return True
+
+
+class CheckStrackTraceFeature(object):
+    def on_import(self, fgraph, node, reason):
+        if theano.config.check_stack_trace == 'check_all':
+            if not check_stack_trace(fgraph, 'all'):
+                raise NotImplementedError(
+                    "Empty stack trace! The optimization that whose"
+                    " stacktrace is empty is" + str(reason))
```
lamblin: I would rather say `"Empty stack trace! The optimization that inserted that variable is " + str(reason)`.

theano/gof/opt.py
```diff
@@ -3039,3 +3039,31 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
...
+        if theano.config.check_stack_trace == 'check_and_skip':
+            if not check_stack_trace(fgraph, 'all'):
+                apply_nodes_to_check = fgraph.apply_nodes
+                for node in apply_nodes_to_check:
+                    for output in node.outputs:
+                        if not hasattr(output.tag, 'trace') or not output.tag.trace:
+                            output.tag.trace = [[('/dummy/path', 0, 'This is a dummy value! The stack trace of ' +
+                                                 str(reason) + ' was empty!', 'dummy(value)')]]
```
lamblin: Actually, I would re-use the same message as above (it is the stack trace of the creation of the variable, not of the optimization).
Is it possible to have empty strings for the dummy values?

theano/configdefaults.py
```diff
@@ -1505,6 +1505,15 @@ def filter_vm_lazy(val):
              EnumStr('regular', 'fast'),
              in_c_key=False)
 
+AddConfigVar('check_stack_trace',
+             "A flag for checking the stack trace during the optimization"
+             " process. The default is not checking the stack trace of any"
+             " optimization. check_all checks if the stack trace is added"
+             " and check_and_skip checks all the stack trace of optimizations"
```
lamblin: Actually, "check_and_skip" inserts the dummy value and does not raises any warnings or exceptions.

theano/configdefaults.py
```diff
@@ -1505,6 +1505,15 @@ def filter_vm_lazy(val):
              EnumStr('regular', 'fast'),
              in_c_key=False)
 
+AddConfigVar('check_stack_trace',
+             "A flag for checking the stack trace during the optimization"
+             " process. The default is not checking the stack trace of any"
+             " optimization. check_all checks if the stack trace is added"
```
lamblin: Mention that "check_all" raises an exception if a stack trace is missing.

theano/configdefaults.py
```diff
@@ -1505,6 +1505,15 @@ def filter_vm_lazy(val):
...
+             " and check_and_skip checks all the stack trace of optimizations"
+             " but skips the ones that have a dummy value ",
+             EnumStr('not_checking', 'check_and_skip', 'check_all'),
```
lamblin: Maybe we should rename the options a little bit. I would suggest:
- `"off"`: do not perform any check (default)
- `"log"` (or maybe `"insert_dummy"? not sure about how to name it): inserts a dummy stack trace that identifies the optimization that inserted that variable
- `"warn"`: prints a warning if a stack trace is missing
- `"raise"`: raises an exception if a stack trace is missing

theano/gof/opt.py
ReyhaneAskari: I forgot to change it. Thanks.

theano/gof/opt.py
ReyhaneAskari: I'm gonna try empty strings but the zero is a line number so I'm gonna leave it as is.

theano/configdefaults.py
ReyhaneAskari: Thanks! Better naming! So for the ```log``` don't we want to add a warning too?

ReyhaneAskari:
I'm going to remove the last commit after the tests passed.

theano/configdefaults.py
lamblin: I think it should be the other way around: for "warn", we should also add the dummy stack trace (so that we only have the warning for the right optimization, and not the next ones).

ReyhaneAskari: Done. Thanks.

theano/gof/opt.py
```diff
@@ -3039,3 +3039,34 @@ def check_stack_trace(f_or_fgraph, ops_to_check='last', bug_print='raise'):
...
+
+
+class CheckStrackTraceFeature(object):
+    def on_import(self, fgraph, node, reason):
+        # In optdb we only register the CheckStackTraceOptimization when
+        # theano.config.check_stack_trace is not off but we also double check here.
+        if theano.config.check_stack_trace != 'off' and not check_stack_trace(fgraph, 'all'):
+            if theano.config.check_stack_trace == 'raise':
+                    raise NotImplementedError(
+                        'Empty stack trace! The optimization that inserted this variable is' + str(reason))
```
lamblin: Raising something else than `AssertionError` will, by default, just cause the optimization to be skipped (that is, unless `config.on_opt_error='raise'`).
So I would prefer to use `AssertionError` rather than `NotImplementedError` here.
Also don't forget a space between 'is' and 'reason'.

theano/configdefaults.py
```diff
@@ -1505,6 +1505,16 @@ def filter_vm_lazy(val):
...
 
+AddConfigVar('check_stack_trace',
+             "A flag for checking the stack trace during the optimization process. "
+             "default (off): does not check the stack trace of any optimization "
+             "log: inserts a dummy stack trace that identifies the optimization"
+             "that inserted that variable."
+             "warn: prints a warning if a stack trace is missing"
+             "raise: raises an exception if a stack trace is missing",
+             EnumStr('off', 'log', 'warn', 'raise'),
+             in_c_key=False)
```
lamblin: Can you add documentation for that in `doc/lib/config.txt`?

theano/compile/mode.py
```diff
@@ -226,6 +226,15 @@ def apply(self, fgraph):
 optdb.register('merge3', gof.MergeOptimizer(),
                100, 'fast_run', 'merge')
 
+if theano.config.check_stack_trace in ['raise', 'warn', 'log']:
+    tags = ('fast_run', 'fast_compile')
```
lamblin: Maybe name that temporary variable something that looks more private like `_tags`, or just `del tags` after calling `optdb.register`.
That code would make it usable as `theano.compile.mode.tags`, but that would not really make sense for anything else than `CheckStackTrace`.

ReyhaneAskari: Sure. Thanks.

theano/gof/opt.py
ReyhaneAskari: Oh, I didn't know that. Done.

theano/configdefaults.py
ReyhaneAskari: Sure. Done.

PR #6156, opened 2017-07-15
Added options for `disconnected_outputs` to Rop.

botev:
Could someone explain me where does this ... is coming from:
`/home/travis/build/Theano/Theano/theano/gradient.py:docstring of theano.gradient.Rop:21: WARNING: Field list ends without a blank line; unexpected unindent.`

### DIFF
--- a/theano/gradient.py
+++ b/theano/gradient.py
@@ -160,7 +160,8 @@ disconnected_type = DisconnectedType()
 ########################
 
 
-def Rop(f, wrt, eval_points):
+def Rop(f, wrt, eval_points, disconnected_outputs="raise",
+        return_disconnected="zero"):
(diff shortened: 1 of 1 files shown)


lamblin:
It comes from the `napoleon` plugin for sphinx, used in autodoc, to generate this for instance:
http://deeplearning.net/software/theano_versions/dev/library/gradient.html#theano.gradient.Rop

botev:
@lamblin But from the error message, I don't understand where exactly is the error coming from? Could you point me out so that I fix it?

lamblin:
Thanks!
I started converted some docstrings to numpydoc and making sure they display correctly, I'll submit that as a different PR.

PR #6159, opened 2017-07-17
make variable class new style

jhelie:
In #6111 I introduced the user of `super()` for class initialisation without realising that the parent class `Variable` was old-style.

This PR makes it new-style (alternatively the changes in #6111 could be reverted to use `Variable.__init__` in child classes but I believe using `super()` is more proper - and in Python 3 all classes are new-style).


lamblin:
Jenkins test this please

PR #6168, opened 2017-07-18
PR cache correct

slefrancois:
In progress

### DIFF
--- a/.jenkins/jenkins_pr_build_cache.sh
+++ b/.jenkins/jenkins_pr_build_cache.sh
@@ -42,7 +42,18 @@ export PYTHONPATH=${PYTHONPATH}:$LIBDIR/lib/python
 # Then install
 (cd libgpuarray && python setup.py install --home=$LIBDIR)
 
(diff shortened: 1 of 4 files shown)


slefrancois:
jenkins test win

slefrancois:
jenkins test this

abergeron:
It doesn't seem to have sped up anything.  Also, this version of the cache code will reuse the cache between PRs between rebuilds of the cache.

You could try to use rsync --delete to remove the differing files from the destination so you don't have to copy from scratch all the time.

slefrancois:
I put and rm -rf first to avoid reusing the previous PR cache. Also I don't think rsync --delete would change anything since before I was copying onto the docker filesystem, which is reset everytime.

slefrancois:
jenkins test this

slefrancois:
jenkins test this

slefrancois:
jenkins test mac

nouiz:
Closing for now. We need to first make Theano better support moving/coping its cache.

PR #6169, opened 2017-07-18
Fix strides when using dot inside gemv

### LINKED ISSUE #6142: Slicing within tensordot
Hi,

I noticed a weird result using slicing within tensordot. The following code shows how slicing in a call to theano.tensor.tensordot produces a different result from slicing in a call to a function mapping inputs to the result of theano.tensor.tensortdot.

The second value seems to be right since numpy.tensordot produces the same result.

I checked the `t[..., 0]` syntax (t being a theano tensor) which seems to work properly. 

```
import numpy as np
import theano
import theano.tensor as T

floatX = theano.config.floatX

a1 = T.matrix(dtype=floatX)
b1 = T.tensor3(dtype=floatX)
c1 = T.tensordot(a1, b1[..., 0])

testFunc1 = theano.function([a1, b1], c1)

a2 = T.matrix(dtype=floatX)
b2 = T.matrix(dtype=floatX)
c2 = T.tensordot(a2, b2)

testFunc2 = theano.function([a2, b2], c2)

mat = np.random.random((3, 3)).astype(floatX)
tensor = np.random.random((3, 3, 2)).astype(floatX)

print(testFunc1(mat, tensor)) # slicing in call to theano.tensor.tensordot
print(testFunc2(mat, tensor[..., 0])) # slicing in call to theano function
print(np.tensordot(mat, tensor[..., 0])) # slicing in call to numpy.tensordot
print(np.tensordot(mat, tensor)[0]) # no slicing with numpy.tensordot
```

(This result was obtained using theano 0.9)

lamblin:
Fix #6142

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -120,16 +120,12 @@ class GpuGemv(BlasOp):
             %(out)s->ga.nd = 0;
             %(A)s->ga.nd = 1;
             %(A)s->ga.dimensions[0] = %(A)s->ga.dimensions[1];
-            if (%(A)s->ga.flags & GA_C_CONTIGUOUS) {
-                ssize_t a_stride0 = %(A)s->ga.strides[0];
-                %(A)s->ga.strides[0] = %(A)s->ga.strides[1];
-                if (pygpu_blas_rdot(%(x)s, %(A)s, %(out)s, 0) == -1) {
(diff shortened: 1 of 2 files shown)


lamblin:
I'll add a test.

lamblin:
Test added.

PR #6170, opened 2017-07-19
Fix some problems with the GPU

abergeron:
- Skip tests when the GPU could not be initialized
 - Stop using arithmetic on void pointers


slefrancois:
jenkins test mac

slefrancois:
Convolution issues on mac seem ok now.

PR #6173, opened 2017-07-19
Fix tests that always expect a C compiler

abergeron:


theano/gof/tests/test_vm.py
```diff
@@ -441,6 +441,8 @@ def check_storage(storage_map):
 
 
 def test_no_recycling():
+    if theano.config.cxx == '':
+        raise SkipTest('need c++')
     x = theano.tensor.vector()
     for lnk in [vm.VM_Linker(use_cloop=True),
```
lamblin: I think we should only skip that linker, not all of them

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -770,6 +770,10 @@ def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
                            mode=mode)
 
     def tcase_gi(self, i, f, o, s, b, flip, provide_shape, fd=(1, 1), expect_error=False):
+
+        if not theano.config.cxx:
+            raise SkipTest("Need cxx to test conv2d")
```
lamblin: This one is supposed to pass, as it is testing the Python perform in AbstractConv2d.
It is possible that some calls to verify_grad are failing because _they_ insert nodes that require a c++ compiler.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -904,6 +908,9 @@ def tcase(self, i, f, s, b, flip, provide_shape, fd=(1, 1)):
                           filter_dilation=fd)
 
     def tcase_gi(self, i, f, o, s, b, flip, provide_shape, fd=(1, 1), expect_error=False):
+        if not theano.config.cxx:
+            raise SkipTest("Need cxx to test conv2d")
```
lamblin: This one as well, as Conv2d has Python code.
We may decide it would be too long to execute, but it would at least deserve a comment, and it was tested before.

theano/tensor/nnet/tests/test_abstract_conv.py
abergeron: I think this fails because it compares against CorrMM and that doesn't have python code.

theano/tensor/nnet/tests/test_abstract_conv.py
abergeron: Same here

theano/gof/tests/test_vm.py
abergeron: This is a test that the generated C code is fine with the no_recycling change that @nouiz did.

I don't see the point of testing this with no C++ compiler.

lamblin: Oh, right, we access `thunk.module` later. OK.

theano/tensor/nnet/tests/test_abstract_conv.py
lamblin: OK.

Issue #6094, opened 2017-07-01
use alloca instead of malloc/free?


nouiz:
This would simplify the code and could give a small speed up.

See comment from @notoraptor about this: https://github.com/Theano/Theano/pull/6091#discussion_r125052740

It is not sure to be safe... We should ask more about that.

PR #6174, opened 2017-07-20
Wrap op params for [Gpu]DimShuffle.

notoraptor:
As a quick benchmark, running:
```
nosetests -xvs
 theano\tensor\tests\test_basic.py:test_batched_dot
 theano\tensor\tests\test_basic.py:test_batched_dot_not_contiguous
 theano\tensor\tests\test_basic.py:BatchedDotTester
```
With `master` branch (after `theano-cache purge`), 70.428s and 11 compilations for CPU DimShuffle.

With this branch (after `theano-cache purge`), 62.231s and 1 compilation for CPU DimShuffle.

@abergeron @lamblin

### DIFF
--- a/theano/gpuarray/c_code/dimshuffle.c
+++ b/theano/gpuarray/c_code/dimshuffle.c
@@ -1,0 +1,1 @@
+#section support_code
--- a/theano/gpuarray/c_code/dimshuffle.c
+++ b/theano/gpuarray/c_code/dimshuffle.c
@@ -1,1 +1,81 @@
-#section support_code
+#section support_code_apply
+
+int gpu_dimshuffle(PyGpuArrayObject* input, PyGpuArrayObject** out, PARAMS_TYPE* params) {
+    PyGpuArrayObject *tmp = NULL;
+    npy_intp nd_in = PyArray_SIZE(params->input_broadcastable);
+    npy_intp nd_out = PyArray_SIZE(params->_new_order);
+    npy_int64* new_order = NULL;
+    unsigned int* transposition = NULL;
+    size_t* sh = NULL;
+    int e;
+
+    if (input->ga.nd != nd_in) {
+        PyErr_SetString(PyExc_TypeError, "input nd");
+        return 1;
+    }
+    if (!PyArray_IS_C_CONTIGUOUS(params->_new_order)) {
+        PyErr_SetString(PyExc_RuntimeError, "DimShuffle: param _new_order must be C-contiguous.");
+        return 1;
+    }
+    if (!PyArray_IS_C_CONTIGUOUS(params->transposition)) {
+        PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: param transposition must be C-contiguous.");
+        return 1;
+    }
+
+    Py_XDECREF(*out);
+
+    /** Do shuffle. **/
+
+    new_order = (npy_int64*) PyArray_DATA(params->_new_order);
+    transposition = (unsigned int*) malloc(nd_in * sizeof(unsigned int));
+    sh = (size_t*) malloc(nd_out * sizeof(size_t));
+    if (transposition == NULL || sh == NULL) {
+        PyErr_NoMemory();
+        free(transposition);
+        free(sh);
+        return 1;
+    }
+    for (npy_intp i = 0; i < nd_in; ++i) {
+        transposition[i] = ((npy_int64*) PyArray_DATA(params->transposition))[i];
+    }
+    tmp = pygpu_transpose(input, transposition);
+    if (!tmp) {
+        PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to transpose input.");
+        free(transposition);
+        free(sh);
+        return 1;
+    }
+    e = 0;
+    for (npy_intp i = 0; i < nd_out; ++i) {
+        if (new_order[i] == -1) {
+            sh[i] = 1;
+        } else {
+            sh[i] = tmp->ga.dimensions[e];
+            ++e;
+        }
+    }
+    *out = pygpu_reshape(tmp, nd_out, sh, GA_ANY_ORDER, 1, -1);
+    Py_DECREF(tmp);
+    free(transposition);
+    free(sh);
+
+    /** End shuffle. **/
+
+    if (*out == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to reshape output.");
+        return 1;
+    }
+
+    if (!params->inplace) {
+        tmp = pygpu_copy(*out, GA_ANY_ORDER);
+        Py_DECREF(*out);
+        if (!tmp) {
+            PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to copy output.");
+            *out = NULL;
+            return 1;
+        }
+        *out = tmp;
+    }
+
+    return 0;
+}
--- a/theano/gpuarray/c_code/dimshuffle.c
+++ b/theano/gpuarray/c_code/dimshuffle.c
@@ -27,21 +27,16 @@ int gpu_dimshuffle(PyGpuArrayObject* input, PyGpuArrayObject** out, PARAMS_TYPE*
     /** Do shuffle. **/
 
     new_order = (npy_int64*) PyArray_DATA(params->_new_order);
-    transposition = (unsigned int*) malloc(nd_in * sizeof(unsigned int));
+    /* Type of params->transposition (npy_uint32) should be an alias of unsigned int
+     * on platforms supported by Theano. */
+    transposition = (unsigned int*) PyArray_DATA(params->transposition);
     sh = (size_t*) malloc(nd_out * sizeof(size_t));
-    if (transposition == NULL || sh == NULL) {
+    if (sh == NULL) {
         PyErr_NoMemory();
-        free(transposition);
-        free(sh);
         return 1;
-    }
-    for (npy_intp i = 0; i < nd_in; ++i) {
-        transposition[i] = ((npy_int64*) PyArray_DATA(params->transposition))[i];
     }
     tmp = pygpu_transpose(input, transposition);
     if (!tmp) {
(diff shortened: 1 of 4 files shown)


notoraptor:
I forgot to update GpuDimShuffle.

notoraptor:
Updated. I think GpuDimShuffle should be rewritten in a separate PR, as it is not trivial !

theano/tensor/elemwise.py
```diff
@@ -265,98 +278,113 @@ def c_code(self, node, name, inp, out, sub):
...
+        npy_intp nd_out;
+        PyArrayObject* %(basename)s;
+        npy_intp* dimensions;
+        npy_intp* strides;
+
+        if (!PyArray_IS_C_CONTIGUOUS(%(params)s->input_broadcastable)) {
+            PyErr_SetString(PyExc_RuntimeError, "DimShuffle: param input_broadcastable must be C-contiguous.");
+            %(just_fail)s
+        }
+        if (!PyArray_IS_F_CONTIGUOUS(%(params)s->_new_order)) {
```
abergeron: You check for F_CONTIGUOUS here.

abergeron:
With the scope of the changes to DimShuffle, you might even consider moving to COp.

theano/tensor/elemwise.py
notoraptor: Oh, what a mistake. Sorry.

notoraptor:
@abergeron Done. Some remarks:
- as GpuDimShuffle inherits from DimShuffle which is now a COp, I added an empty C file for GpuDimShuffle to prevent file loading error. I have not rewritten the GPU op in that PR.
- from memory, there may be some pickling issues with old objects when we convert an Op to a COp, so I added a `__setstate__()` implementation to prevent it.

theano/gpuarray/c_code/dimshuffle.c
```diff
@@ -0,0 +1 @@
+#section support_code
```
lamblin: I thought it would not be necessary, since `GpuDimShuffle` inherits from `HideC`.
@abergeron did I miss something?
I don't really like the idea of an almost-empty file.

notoraptor: @lamblin In fact, GpuDimShuffle inherits from DimShuffle which is now a COp. But DimShuffle has an `__init__()` in which there is a call to `COp.__init__(self, [filenames], function_name)`. Then this `__init__()` is inherited into GpuDimShuffle, even if C functions are hidden.

I don't know how to bypass that. Maybe should I just rewrite GpuDimShuffle too in that PR as soon as possible ? (Monday, I think !)

abergeron: GpuDimShuffle inherits from HideC and DimShuffle.  HideC will ensure that we don't try to reuse the C code from DimShuffle, so you could probably just add an empty `__init__` to GpuDimShuffle.

If you intend on rewriting GpuDimShuffle soon, even if not in this PR, I would say the current fix is fine.

notoraptor: In fact, some important attributes like `self.suffle`, `self.drop`, `self.new_order` are created and checked  into DimShuffle `__init__()`, so this method is necessary for GpuDimShuffle too.

notoraptor: I may rewrite GpuDimShuffle today.



NB: small remark: One of the problems when we rewrite ops is that we now need to allocate small arrays (`int* shape = (int*)malloc(params->size * sizeof(int))`), where we previously just hardcoded it (`int shape[%(size)s]`). It's quite regrettable, but I don't find a way to do it otherwise.

abergeron: We could also go with a "max case" array of like 32 and check that the input array doesn't have more dimensions than that.

notoraptor: Indeed, it's generally about array of strides or dimensions. But are we sure that users always work with at most 32 dimensions ^^ ? I think we can't do that assumption, unfortunately.

abergeron: NumPy has that limit.  libgpuarray doesn't currently, but I'm thinking about adding it in, since it would simplify some things.

notoraptor:
Updated (with malloc()). GpuDimShuffle rewritten.

notoraptor:
Jenkins fails, but I don't know why: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/461/#showFailuresLink

jenkins test this

notoraptor:
Jenkins succeeded.

theano/gpuarray/c_code/dimshuffle.c
lamblin: What about `alloca`?

theano/gpuarray/elemwise.py
```diff
@@ -563,7 +504,7 @@ def __init__(self, scalar_op, axis=None,
             reduce_mask = tuple(reduce_mask)
         self.reduce_mask = reduce_mask
 
-        # used to make sure that calls to scalar op
+        # used to make sure that callfs to scalar op
```
lamblin: Probably a typo here

theano/gpuarray/c_code/dimshuffle.c
notoraptor: Not sure that alloca is portable across systems and compilers:
- http://man7.org/linux/man-pages/man3/alloca.3.html#NOTES
- https://github.com/Theano/Theano/issues/6094

theano/gpuarray/elemwise.py
notoraptor: Indeed, I'll fix it!

notoraptor:
Typo fixed @lamblin .

theano/gpuarray/c_code/dimshuffle.c
```diff
@@ -0,0 +1,81 @@
...
+        free(transposition);
+        free(sh);
+        return 1;
+    }
+    for (npy_intp i = 0; i < nd_in; ++i) {
+        transposition[i] = ((npy_int64*) PyArray_DATA(params->transposition))[i];
+    }
+    tmp = pygpu_transpose(input, transposition);
+    if (!tmp) {
+        PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to transpose input.");
```
abergeron: There is already a python exception here.

theano/gpuarray/c_code/dimshuffle.c
```diff
@@ -0,0 +1,81 @@
...
+    if (*out == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to reshape output.");
+        return 1;
+    }
+
+    if (!params->inplace) {
+        tmp = pygpu_copy(*out, GA_ANY_ORDER);
+        Py_DECREF(*out);
+        if (!tmp) {
+            PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to copy output.");
```
abergeron: Masking exception

theano/gpuarray/c_code/dimshuffle.c
```diff
@@ -0,0 +1,81 @@
...
+    }
+    *out = pygpu_reshape(tmp, nd_out, sh, GA_ANY_ORDER, 1, -1);
+    Py_DECREF(tmp);
+    free(transposition);
+    free(sh);
+
+    /** End shuffle. **/
+
+    if (*out == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "GpuDimShuffle: unable to reshape output.");
```
abergeron: Masking exception.

theano/gpuarray/c_code/dimshuffle.c
```diff
@@ -0,0 +1,81 @@
...
+    transposition = (unsigned int*) malloc(nd_in * sizeof(unsigned int));
+    sh = (size_t*) malloc(nd_out * sizeof(size_t));
+    if (transposition == NULL || sh == NULL) {
+        PyErr_NoMemory();
+        free(transposition);
+        free(sh);
+        return 1;
+    }
+    for (npy_intp i = 0; i < nd_in; ++i) {
+        transposition[i] = ((npy_int64*) PyArray_DATA(params->transposition))[i];
```
abergeron: You could set the type of transposition to uintt32 and use it directly to avoid a malloc and copy.

notoraptor: `pygpu_transpose` expects transposition to be an array of `unsigned int`, but we can't be sure that `int` it always 32 bits in all machines and computers.

NB: I will make a commit to suggest a solution in next hours.

theano/gpuarray/c_code/dimshuffle.c
notoraptor: Ah ok, I see.

theano/gpuarray/c_code/dimshuffle.c
abergeron: We know that this is the case on all platforms that we care about and we can always add an assert so that we get a failure rather than wrong results on exotic platforms.

notoraptor:
@abergeron I was working on a way to get sizeof a machine-dependent C type, so I added the work here. What do you think ? If it's unsafe or too complicated, I can remove it and just put an assert as you suggest.

notoraptor:
Code ultimately updated and simplified (transposition is simply an array of `npy_uint32`) @abergeron .

theano/tensor/elemwise.py
```diff
@@ -185,6 +206,13 @@ def __init__(self, input_broadcastable, new_order, inplace=True):
         if self.inplace:
             self.view_map = {0: [0]}
```
abergeron: I just thought about this, but don't we enable views all the time.  Usually inplace is for destroy_map.

@lamblin ?

notoraptor: Should I change `view_map` with `destroy_map` ?

abergeron: No, because this doesn't actually destroy anything.  I am saying that we can probably remove the inplace property and act as if it was always True.

PR #6176, opened 2017-07-20
Lower and upper triangular matrix solvers on the GPU

botev:
As pointed out by @juancamilog  there exist triangular solvers on the GPU, but they are part of cuBLAS rather than cuSOLVE. Here I have added the operator for handling these inside the `GpuSolve`. However, there is a question of whether there should not be a separate operator for this or this is fine. I'm fine with writing and a new operator but that is all op to you guys.

### DIFF
--- a/theano/gpuarray/linalg.py
+++ b/theano/gpuarray/linalg.py
@@ -25,10 +25,16 @@ except ImportError:
     pygpu_available = False
 
 cusolver_available = False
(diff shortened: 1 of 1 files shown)


botev:
I think #6182  is the better alternative so if that get's merged this should be closed.

abergeron:
#6182 was merged

PR #6178, opened 2017-07-20
Repair tests that fail with compute_test_value.

abergeron:


theano/tensor/nnet/abstract_conv.py
```diff
@@ -2300,8 +2300,8 @@ def correct_for_groups(mat):
             mshp0 = mat.shape[0] // self.num_groups
             mshp1 = mat.shape[1] * self.num_groups
             mat = mat.reshape((self.num_groups, mshp0) + mat.shape[1:])
-            mat = mat.transpose((1, 0, 2, 3, 4))
-            mat = mat.reshape((mshp0, mshp1) + mat.shape[-2:])
+            mat = mat.transpose((1, 0, 2) + tuple(range(3, 3+self.convdim)))
```
lamblin: E226 missing whitespace around arithmetic operator

theano/tensor/basic.py
```diff
@@ -5862,53 +5862,6 @@ def R_op(self, inputs, eval_points):
...
-
-            for i in xrange(2):
-                if eval_point_values[i] is not None and \
-                   input_values[i].shape != eval_point_values[i].shape:
-                    raise ValueError(
-                        'input ' + str(i) + ' and eval_point ' + str(i) +
-                        ' to Dot.R_op should have the same shape, but '
-                        'their shapes are %s and %s, respectively' % (
-                            str(input_values[i].shape),
-                            str(eval_point_values[i].shape)))
```
abergeron: It gave an error about undefined variables and I question its usefulness.

This is the only op to have a weird debug behavior in reaction to compute_test_values, which should anyway catch most of the problem that this code was attempting to report.

lamblin: OK.

PR #6182, opened 2017-07-21
Added GpuCublasTriangularSolve and the associated lifter optimizations

juancamilog:
This is an update of #4007 to use the gpuarray backend. This is an alternative to #6176 that separates the triangular solve into a new op, and deals with the case when the right hand side is a vector (as in #4007).


abergeron:
jenkins test this please

PR #6186, opened 2017-07-21
Exclude CI files from sdist

ghisvail:


lamblin:
Jenkins test this please

lamblin:
This makes sense, I just want to make sure that it does not break our current Jenkins setup.

PR #6194, opened 2017-07-24
Support more debuggers for PdbBreakpoint

aam-at:
`PdbBreakpoint` by default supports only `pdb` debugger which has limited functionality. This pull request adds support for `pudb` or `ipdb` if available.

### DIFF
--- a/doc/tutorial/debug_faq.txt
+++ b/doc/tutorial/debug_faq.txt
@@ -608,3 +608,6 @@ Breakpoint during Theano function execution
 
 You can set a breakpoint during the execution of a Theano function with
 :class:`PdbBreakpoint <theano.tests.breakpoint.PdbBreakpoint>`.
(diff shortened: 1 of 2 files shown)


lamblin:
Jenkins test this please

lamblin:
I approve in principle, but could you update the documentation (inline and in `doc/` if needed) to reflect the change of behaviour?
Thanks!

aam-at:
@lamblin I updated docstring. Tests were failing on `theano.sparse.tests.test_basic.test_structureddot.test_csr_correct_output_faster_than_scipy` and not related to this pull request.

lamblin:
Jenkins test this please

nouiz:
Why not make this a parameter that default to the current test? I'm pretty sure we will have at some point someone have pudb installed by default and don't know about it and have it being used.

Anyway, the PR make Theano better. If such problems appear, we can make it a parameter later.

PR #6199, opened 2017-07-24
Fix some problems in the daily buildbot in DEBUG_MODE.

abergeron:


theano/compile/debugmode.py
```diff
@@ -2286,7 +2287,8 @@ def __init__(self, inputs, outputs, mode,
             fgraph.attach_feature(gof.DestroyHandler())
         for o in fgraph.outputs:
             try:
-                fgraph.replace_validate(o, _output_guard(o), reason='output_guard')
+                with change_flags(compute_test_value='off'):
```
lamblin: Actually, I think we should set it ideally to `theano.config.compute_test_value_opt`, since `o` should have a test value in that case.

abergeron: done

PR #6204, opened 2017-07-26
using op params for Reshape ops

### LINKED ISSUE #6137: Use params for Reshape ops
Some props for these ops may be converted as op params to reduce C code.

- theano.tensor.basic.Reshape
  - ndim
- theano.gpuarray.basic_ops.GpuReshape
  - ndim

anirudh9119:
Fix #6137

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1157,12 +1157,11 @@ class GpuReshape(HideC, tensor.Reshape):
         out[0] = x.reshape(tuple(shp))
 
     def c_code_cache_version(self):
-        return (1,)
+        return (2,)
 
     def c_code(self, node, name, inputs, outputs, sub):
         x, shape = inputs
         output, = outputs
-        new_ndim = self.ndim
         sdtype = node.inputs[1].type.dtype_specs()[1]
         fail = sub['fail']
         params = sub['params']
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1129,7 +1129,7 @@ class GpuReshape(HideC, tensor.Reshape):
                              context_name=ctx_name)
         return Apply(self, [x, shp], [otype()])
 
-    def perform(self, node, inp, out_):
+    def perform(self, node, inp, out_, params):
         x, shp = inp
         out, = out_
         if (len(shp) != self.ndim):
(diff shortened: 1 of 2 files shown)


theano/tensor/basic.py
```diff
@@ -4871,11 +4876,11 @@ def c_code(self, node, name, inputs, outputs, sub):
             fail = sub['fail']
             return """
             assert (PyArray_NDIM(%(shp)s) == 1);
-            npy_intp new_dims[%(new_ndim)s];
+            npy_intp new_dims[%(params)s->ndim];
```
notoraptor: The string here is intended to be formatted, using the Python old-style formatting: every occurence of `%(varname)s` will be replace with value of a field `varname` in a dict passed as formatting argument (`string_to_format % dict_with_fields`). Here the dict passed for formatting is `locals()` that contains any local variable defined into current function.

So, as you add `%(params)s` in the string, there should be a variable named `params` in the code of current function, but you have not added that variable. An error should occur if you test your code.

`params` must contain the name of the C object representing the op param. It is located into `sub`:
```python
params = sub['params']
```

theano/tensor/basic.py
```diff
@@ -4708,29 +4709,33 @@ def vertical_stack(*args):
...
+        self.ndim = int(ndim)
         if ndim < 0:
             raise ValueError("The output dimensions after reshape must be 0 or greater")
         assert name is None, 'name attribute for Reshape has been deprecated'
 
     def __str__(self):
         return '%s{%s}' % (self.__class__.__name__, self.ndim)
 
+
+    def get_params(self, node):
```
notoraptor: When we use `ParamsType`, and if all params defined into `ParamsType` have corresponding op attributes, then we don't need to define `get_params()`: params values will be retrieved automatically.

Here, we define `ParamsType` with a param `ndim`, and this Op has an attribute named `ndim`, so we don't need to define `get_params()`.

anirudh9119: okay!

theano/tensor/basic.py
anirudh9119: Okay, Resolved!

theano/tensor/basic.py
```diff
@@ -4869,13 +4871,15 @@ def c_code(self, node, name, inputs, outputs, sub):
             new_ndim = self.ndim
```
notoraptor: Also, remove this variable (new_ndim), as it should not be used anymore.

notoraptor:
Hello @anirudh9119 , increment number in `c_code_cache_version()` for both ops, to tell Theano that C code has changed.

theano/gpuarray/basic_ops.py
```diff
@@ -1165,25 +1165,26 @@ def c_code(self, node, name, inputs, outputs, sub):
         new_ndim = self.ndim
```
notoraptor: Remose new_ndim here also.

anirudh9119:
you meant, change it to 8 ? return (8,) ?




- - Anirudh

notoraptor:
@anirudh9119 Yes, 7 ->8 for CPU op and 1 -> 2 for GPU op.

notoraptor:
jenkins test this

notoraptor:
jenkins test this

theano/tensor/basic.py
```diff
@@ -29,6 +29,7 @@
 from theano.printing import pprint, min_informative_str
 # For history
 from theano.compile import Rebroadcast, Shape, shape
+from theano.scalar import Scalar
```
notoraptor: Instead of `Scalar`, import `int32` here, which is a predefined Scalar instance for 32-bits integers.

```python
from theano.scalar import int32
```

theano/tensor/basic.py
```diff
@@ -4710,20 +4711,19 @@ def vertical_stack(*args):
...
     The number of dimensions to which to reshape to (ndim) must be
     known at graph build time.
-
     """
     view_map = {0: [0]}  # output 0 is potentially aliased to inputs [0]
     _f16_ok = True
 
     check_input = False
     __props__ = ("ndim",)
+    params_type = ParamsType(ndim=Scalar('int32'))
```
notoraptor: Then use imported `int32` here:

```python
params_type = ParamsType(ndim=int32)
```

notoraptor:
jenkins test this

Issue #5192, opened 2016-11-07
Add support of bool indexing in Theano


nouiz:
Now that we support bool dtype in Theano, we should add the support for indexing with bool.

khaotik:
Should this be implemented as an elemwise Op? (internally an if switch)

nouiz:
If we just call numpy implementation, then it won't be hard to implement on
CPU.

There is 2 difficulties:
1) we need to determine the output size for this we need to do a first pass
on the bool mask
2) Then doing serially the copy isn't hard, but doing it in parallel is
more complicated. So the GPU version will be hard to do. Do to so and 1),
we should probably do a prefix scan first, this could give us the
destination index.

lamblin:
The easiest immediate way of making that work would be to return `x[mask.nonzeros()]`, rather than implementing a new op.

nouiz:
The corresponding keras issue: https://github.com/fchollet/keras/issues/4976

nouiz:
Note, @lamblin suggestion would make this work on CPU and GPU.

But on CPU, we can just use the class AdvancedSubtensor. Its perform should work for that cases. Maybe small adjustment would be needed. It would be faster. This would need adjustment to the GPU opt of AdvancedSubtensor to introduce the nonzero() as I think the GPU version don't support it.

nouiz:
Done in gh-6207, so I'll close this issue.

PR #6207, opened 2017-07-27
Add support for boolean indexing

gvtulder:
This would be the easy way of adding support for boolean indexing, as suggested in #5192. It passes the simple tests, but it's not unlikely that there are edge cases that I didn't consider.

### DIFF
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -70,9 +70,11 @@ from .elemwise import (GpuElemwise, GpuDimShuffle, GpuCAReduceCuda,
 from .subtensor import (GpuIncSubtensor, GpuSubtensor,
                         GpuAdvancedSubtensor,
                         GpuAdvancedSubtensor1,
+                        GpuAdvancedBooleanSubtensor,
                         GpuAdvancedIncSubtensor,
                         GpuAdvancedIncSubtensor1,
                         GpuAdvancedIncSubtensor1_dev20,
+                        GpuAdvancedBooleanIncSubtensor,
                         GpuAllocDiag, GpuExtractDiag)
 from .opt_util import alpha_merge, output_merge, pad_dims, unpad_dims
 from .reduction import GpuMaxAndArgmax
@@ -1079,7 +1081,7 @@ def local_gpua_advanced_subtensor(op, context_name, inputs, outputs):
 @op_lifter([tensor.AdvancedBooleanSubtensor])
 @register_opt2([tensor.AdvancedBooleanSubtensor], 'fast_compile')
 def local_gpua_advanced_boolean_subtensor(op, context_name, inputs, outputs):
-    return GpuAdvancedSubtensor()
+    return GpuAdvancedBooleanSubtensor()
 
 
 @register_opt('fast_compile')
@@ -1134,7 +1136,7 @@ def local_gpua_advanced_boolean_incsubtensor(op, context_name, inputs, outputs):
     # GpuAdvancedIncSubtensor only works with a single boolean mask,
     # but not with fancy combinations.
     if not op.set_instead_of_inc and len(inputs) == 3:
-        return GpuAdvancedIncSubtensor()
+        return GpuAdvancedBooleanIncSubtensor()
     else:
         return False
 
--- a/theano/gpuarray/opt.py
+++ b/theano/gpuarray/opt.py
@@ -1076,6 +1076,13 @@ def local_gpua_advanced_subtensor(op, context_name, inputs, outputs):
 
 
 @register_opt('fast_compile')
+@op_lifter([tensor.AdvancedBooleanSubtensor])
+@register_opt2([tensor.AdvancedBooleanSubtensor], 'fast_compile')
+def local_gpua_advanced_boolean_subtensor(op, context_name, inputs, outputs):
+    return GpuAdvancedSubtensor()
+
+
+@register_opt('fast_compile')
 @op_lifter([tensor.AdvancedIncSubtensor1])
 @register_opt2([tensor.AdvancedIncSubtensor1], 'fast_compile')
 def local_gpua_advanced_incsubtensor1(op, context_name, inputs, outputs):
@@ -1113,6 +1120,20 @@ def local_gpua_advanced_incsubtensor1(op, context_name, inputs, outputs):
 # @register_opt2([tensor.AdvancedIncSubtensor], 'fast_compile')
 def local_gpua_advanced_incsubtensor(op, context_name, inputs, outputs):
     if not op.set_instead_of_inc:
+        return GpuAdvancedIncSubtensor()
+    else:
+        return False
+
+
+# Do not register this optimization for now, as it slows down the
+# execution by a lot in important cases.
+# @register_opt('fast_compile')
+# @op_lifter([tensor.AdvancedBooleanIncSubtensor])
+# @register_opt2([tensor.AdvancedBooleanIncSubtensor], 'fast_compile')
+def local_gpua_advanced_boolean_incsubtensor(op, context_name, inputs, outputs):
+    # GpuAdvancedIncSubtensor only works with a single boolean mask,
+    # but not with fancy combinations.
+    if not op.set_instead_of_inc and len(inputs) == 3:
         return GpuAdvancedIncSubtensor()
     else:
         return False
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -628,6 +628,9 @@ class GpuAdvancedSubtensor(HideC, BaseGpuAdvancedSubtensor, tensor.AdvancedSubte
     """
     def make_node(self, x, *inputs):
         ctx_name = infer_context_name(x)
+        # This method relies on AdvancedSubtensor.make_node to
+        # call tensor.subtensor.check_and_reject_bool(inputs),
+        # which raises an IndexError if there are any boolean indices.
         rval = tensor.AdvancedSubtensor.make_node(self, x, *inputs)
         otype = GpuArrayType(dtype=rval.outputs[0].type.dtype,
                              broadcastable=rval.outputs[0].type.broadcastable,
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -512,13 +512,13 @@ def check_and_convert_boolean_masks(input, idx_list):
     return out_idx_list
 
 
-class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
+class GpuAdvancedSubtensor(HideC, tensor.AdvancedBooleanSubtensor):
     """
-    AdvancedSubtensor On the GPU.
+    AdvancedBooleanSubtensor On the GPU.
     """
     def make_node(self, x, *inputs):
         ctx_name = infer_context_name(x)
-        rval = tensor.AdvancedSubtensor.make_node(self, x, *inputs)
+        rval = tensor.AdvancedBooleanSubtensor.make_node(self, x, *inputs)
         otype = GpuArrayType(dtype=rval.outputs[0].type.dtype,
                              broadcastable=rval.outputs[0].type.broadcastable,
                              context_name=ctx_name)
@@ -634,14 +634,14 @@ class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
         out[0] = o
 
 
-class GpuAdvancedIncSubtensor(HideC, tensor.AdvancedIncSubtensor):
+class GpuAdvancedIncSubtensor(HideC, tensor.AdvancedBooleanIncSubtensor):
     """
-    Implement AdvancedIncSubtensor on the gpu.
+    Implement AdvancedBooleanIncSubtensor on the gpu.
 
     """
     def make_node(self, x, y, *inputs):
         ctx_name = infer_context_name(x, y)
-        rval = tensor.AdvancedIncSubtensor.make_node(self, x, y, *inputs)
+        rval = tensor.AdvancedBooleanIncSubtensor.make_node(self, x, y, *inputs)
         otype = GpuArrayType(dtype=rval.outputs[0].type.dtype,
                              broadcastable=rval.outputs[0].type.broadcastable,
                              context_name=ctx_name)
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -4,7 +4,7 @@ import os
 
 import numpy as np
 from six import integer_types
-from six.moves import StringIO
+from six.moves import StringIO, xrange
 
 from theano import tensor, gof, Op
 from theano.gof import ParamsType
@@ -481,6 +481,37 @@ if (err != GA_NO_ERROR) {
         return (0,)
 
 
+def check_and_convert_boolean_masks(input, idx_list):
+    """
+    This function checks if the boolean mask arrays in the index have
+    the right shape and converts them to index arrays by calling nonzero.
+    For each boolean mask, we check if the mask has the
+    same shape as the input. This is enforced in NumPy 0.13.0 and
+    newer, but not by earlier versions. If the size is not the same,
+    this method raises an IndexError.
+    """
+    dim_seen = 0
+    out_idx_list = []
+    for index in idx_list:
+        if index is np.newaxis:
+            # skip, does not count as an input dimension
+            out_idx_list.append(index)
+        elif isinstance(index, np.ndarray) and index.dtype == 'bool':
+            for i in xrange(index.ndim):
+                if index.shape[i] != input.shape[dim_seen + i]:
+                    raise IndexError('boolean index did not match indexed array '
+                                     'along dimension %d; dimension is %d but '
+                                     'corresponding boolean dimension is %d' %
+                                     (dim_seen + i, input.shape[dim_seen + i],
+                                      index.shape[i]))
+            dim_seen += index.ndim
+            out_idx_list += index.nonzero()
+        else:
+            dim_seen += 1
+            out_idx_list.append(index)
+    return out_idx_list
+
+
 class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
     """
     AdvancedSubtensor On the GPU.
@@ -498,6 +529,9 @@ class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
         out, = out_
         x = inputs[0]
         idx = inputs[1:]
+
+        # convert boolean masks to index arrays
+        idx = check_and_convert_boolean_masks(x, idx)
 
         # detect and transpose array indices
         nidx = []
@@ -630,6 +664,9 @@ class GpuAdvancedIncSubtensor(HideC, tensor.AdvancedIncSubtensor):
         for i in range(len(idx)):
             if isinstance(idx[i], gpuarray.GpuArray):
                 idx[i] = np.asarray(idx[i])
+
+        # convert boolean masks to index arrays
+        idx = check_and_convert_boolean_masks(x, idx)
 
         # Insert axes for None indexing
         nidx = []
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -512,19 +512,7 @@ def check_and_convert_boolean_masks(input, idx_list):
     return out_idx_list
 
 
-class GpuAdvancedSubtensor(HideC, tensor.AdvancedBooleanSubtensor):
-    """
-    AdvancedBooleanSubtensor On the GPU.
-    """
-    def make_node(self, x, *inputs):
-        ctx_name = infer_context_name(x)
-        rval = tensor.AdvancedBooleanSubtensor.make_node(self, x, *inputs)
-        otype = GpuArrayType(dtype=rval.outputs[0].type.dtype,
-                             broadcastable=rval.outputs[0].type.broadcastable,
-                             context_name=ctx_name)
-        x = as_gpuarray_variable(x, ctx_name)
-        return gof.Apply(self, [x] + rval.inputs[1:], [otype()])
-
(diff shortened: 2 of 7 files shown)


theano/tensor/var.py
```diff
@@ -460,30 +460,48 @@ def astype(self, dtype):
...
             args = args,
 
+        # Convert boolean arrays to calls to mask.nonzero()
+        tmp_args = []
+        for arg in args:
+            # NumPy arrays or tensors of type bool can be converted to
+            # normal integer indices.
+            if (isinstance(arg, (np.ndarray, theano.tensor.Variable)) and
+                    hasattr(arg, 'dtype') and hasattr(arg, 'nonzero') and
+                    arg.dtype == 'bool'):
```
nouiz: you can simplify this condition:

    if isinstance(arg, (np.ndarray, theano.tensor.TensorVariable) and arg.dtype == 'bool':

theano/tensor/tests/test_subtensor.py
```diff
@@ -362,6 +362,51 @@ def test_ellipsis(self):
...
+
+        # indexing with a mask for some dimensions
+        mask = np.array([True, False])
+        assert_array_equal(numpy_n[mask], n[mask].eval())
+
+        # indexing with a mask for the second dimension
+        mask = np.array([True, False, True])
+        assert_array_equal(numpy_n[0, mask], n[0, mask].eval())
+        assert_array_equal(numpy_n[:, mask], n[:, mask].eval())
+        assert_array_equal(numpy_n[:, mask], n[:, self.shared(mask)].eval())
```
nouiz: Can you test with a partial slice too? Just in case.

theano/tensor/tests/test_subtensor.py
```diff
@@ -362,6 +362,51 @@ def test_ellipsis(self):
...
+
+        # indexing with a boolean array
```
nouiz: array -> ndarray

theano/tensor/tests/test_subtensor.py
```diff
@@ -362,6 +362,51 @@ def test_ellipsis(self):
...
+        mask = np.array([[True, False, True], [False, False, True]])
+        assert_array_equal(numpy_n[mask], n[mask].eval())
+        assert_array_equal(numpy_n[mask], n[self.shared(mask)].eval())
+
+        # for the GpuArray tests, self.shared will create GpuArray variables,
+        # but we also need to test the combination of GPU and normal TensorVariables
+        n_cpu = theano.shared(numpy_n)
+        assert_array_equal(numpy_n[mask], n[theano.shared(mask)].eval())
```
nouiz: this won't test on the GPU. The GPU tests don't enable the GPU by default. They need to manually pass the mode. You can use self.mode as the mode to use to enable the GPU optimization to make the test do what you want.

But I don't think we need to test that. I would just remove that test. This is the normal workflow and should work without problems.

theano/tensor/tests/test_subtensor.py
```diff
@@ -362,6 +362,51 @@ def test_ellipsis(self):
...
+        assert_array_equal(numpy_n4[numpy_n > 2, ...], n4[n > 2, ...].eval())
+        assert_array_equal(numpy_n4[numpy_n > 2, ..., 1], n4[n > 2, ..., 1].eval())
+
+        # special cases: Python bools and bools nested in Python arrays are not supported
+        self.assertRaises(TypeError, n.__getitem__, (True,))
+        self.assertRaises(TypeError, n.__getitem__, (False,))
+        self.assertRaises(TypeError, n.__getitem__, (True, False))
+        self.assertRaises(TypeError, n.__getitem__, ([True, False]))
+        self.assertRaises(TypeError, n.__getitem__, ([0, 1], [0, False]))
+        self.assertRaises(TypeError, n.__getitem__, ([0, 1], [0, theano.shared(True)]))
```
nouiz: Also add a test where the mask is too big of too small. Just to make sure we raise the same error as numpy on CPU and GPU.

nouiz:
See my comment in the issue this address. Just in case you wan to do that version of the implementation. I'm good with both of them.

theano/tensor/var.py
gvtulder: That's what I tried first, but that fails on this test:
```python
    assert_array_equal(numpy_n[:, mask], n[:, self.shared(mask)].eval())
```
because `theano.tensor.sharedvar.TensorSharedVariable` is not a `TensorVariable`. There is a similar problem `TensorConstant`, `GpuArraySharedVariable` et cetera. This was the least ugly solution I could think of that covers all those types.

nouiz: ok. You could remove the hasattr(arg, 'dtype'), as currently all the var that have nonzero have dtype. But if you prefer to keep it to safeguard again possible futur that will probably not happen, you can keep it.

gvtulder:
Thanks for the comments. I'll have a look to see how difficult it is to move this to the GPU implementation of AdvancedSubtensor, so that the CPU version doesn't use `nonzero`.

The issue of checking the size of the mask is tricky. Right now, it doesn't do that: you can use a boolean mask that is larger than the tensor you're indexing, as long as there are no `True`s outside the range of your tensor. You can also pass a smaller mask. I agree that it's better to check the size, which would most likely require an `Assert`. The problem is that checking the dimensions gets a bit complicated when you allow for an ellipsis.

I don't fully understand how the ellipsis is handled at the moment: if I do this
```python
a = np.array(...) # some 2D array
x = T.tensor4()
x[a, ..., 0]
```
then it looks like the `__getitem__` method expands the ellipsis to `[a, slice(None), slice(None), 0]`, which is one dimension too many. Yet it still seems to work fine, so I have to look at this in a bit more detail.

nouiz:
For the size, I just checked and numpy currently just do a deprecation
warning for smaller mask and accept longer mask without warning if they are
only False. If True it raise an error. As you wrote.

So, can you just add test that it work? So we don't break it by accident in
the futur. If we break it volontarily, we can easily change the test to
request it to raise an error.

For elipsis. I don't understand why you expected one less dimensions. There
can be only 1 ellipsis. It get replaced by as many full slices as needed to
have the same number of argument then the inputs ndim. So 4 in your
example. Then the normal subtensor logic apply.

gvtulder:
Different sizes will [raise an error in numpy 0.13.0](https://docs.scipy.org/doc/numpy-1.13.0/release.html#boolean-indexing-changes). I think I'd prefer Theano to raise an error right now, instead of allowing it now and removing it again in the near future.

> For elipsis. I don't understand why you expected one less dimensions.

Sorry, that might have been my misunderstanding of how the indexing works. I expected that if `a` is 2D boolean ndarray and `x` is a 4D ndarray, then `x[a]` would be the same as `x[a, ...]`, but apparently it isn't. Now that I know that, I think it makes sense.

nouiz:
Honestly, I always forget what happen when you index with a tensor of indices/boolean. I don't understand why this is useful too. It probably just remove a reshape/dimshuffle. I never use it myself, I find it too confusing.

To make it raise an error for smaller size, it is easy if all the indexing is done in one op, like the AdvancedSubtensor class on the CPU. For the GPU, doing it in the graph would be complicated. We now have a GpuAdvancedIndexing op that only have python code. I think the perform could be modified to check the shapes and call nonzeros() on the index at the start and let the rest of the method unchanged. What do you think of that?

lamblin:
@nouiz It corresponds to selecting elements with a mask, rather than indices. See https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#boolean-array-indexing

gvtulder:
Something like this? The shape checks are now done in the `perform()` methods and `nonzero` is only used for the GPU version.

There's a little issue with the type of the error that is raised when the index dimension is larger than the dimension of the indexed array. In those cases, NumPy raises an `IndexError: too many indices for array`. Theano, currently, has two options: for simple cases, Subtensor [raises a `ValueError`](https://github.com/Theano/Theano/blob/61cb8c4/theano/tensor/subtensor.py#L474-L479), whereas for complex cases AdvancedSubtensor relies on NumPy to throw an `IndexError`. The new code would always raise an `IndexError`, but that breaks [these Subtensor tests](https://github.com/Theano/Theano/blob/61cb8c4/theano/tensor/tests/test_subtensor.py#L188-L195). What's the best way to solve this?

nouiz:
I think it make sence to move in the direction of being more like NumPy. I don't like the idea of having different error type for different indexing error. They are all indexing errors. So I would update the tests to expect IndexError.

theano/tensor/var.py
```diff
@@ -499,25 +503,22 @@ def includes_bool(args_el):
...
                 "an index can only have a single Ellipsis (`...`)")
         elif len(ellipses) == 1:
-            new_axes = sum(1
-                           for index in args
-                           if index is np.newaxis)  # numpy.newaxis is None
             ellipsis_at = ellipses[0]
             args = list(args)
             args[ellipsis_at: ellipsis_at + 1] = (
-                [slice(None)] * (self.ndim - (len(args) - 1 - new_axes)))
+                [slice(None)] * (self.ndim - index_dim_count))
```
nouiz: This change the current logic how we handle indices array that have more then 1 dimensions. As that what you want? What was the problem in how we handle them?

theano/tensor/var.py
```diff
@@ -533,7 +534,12 @@ def includes_bool(args_el):
         axis = None
         for i, arg in enumerate(args):
             try:
-                if arg is not np.newaxis:
+                if (isinstance(arg, (np.ndarray, theano.tensor.Variable)) and
```
nouiz: Why not modify Subtensor.convert() to also raise AdvancedIndexingError? This would make sure that elsewhere where we call convert(), if we forgot to update the code, we will end up using advanced indexing.

theano/gpuarray/subtensor.py
```diff
@@ -481,6 +481,37 @@ def c_code_cache_version(self):
...
+    this method raises an IndexError.
+    """
+    dim_seen = 0
+    out_idx_list = []
+    for index in idx_list:
+        if index is np.newaxis:
+            # skip, does not count as an input dimension
+            out_idx_list.append(index)
+        elif isinstance(index, np.ndarray) and index.dtype == 'bool':
+            for i in xrange(index.ndim):
```
gvtulder: I find it confusing as well, but I think what numpy does is this:
- for multi-dimensional integer arrays, it uses the values as indices for a single dimension (and it reshapes the output in that dimension to have the same shape as the index array);
- for boolean arrays, it uses them as a mask, so that if you pass a boolean matrix as the first indexing element, this will be applied to the first two dimensions of the indexed array.

Example of how I think it works:
```python
x = np.arange(16).reshape(2, 2, 2, 2)

# integer index array
a = np.zeros((2, 2), int)
x[a, 0, 0, 0]  # returns [[0, 0], [0, 0]]
# the indices in matrix a applies only to the first dimension of x

# bool mask array
b = np.array([[True, False], [False, True]])
x[b, 0, 0]  # returns [0, 12]
# the first dimension of mask b applies to the first dimension of x,
# the second dimension of b to the second dimension of x,
# so only the third and fourth remain to be indexed
```
For integer arrays, the shape of the index array doesn't matter, because they're used as indices for a single dimension. For boolean arrays, the shape must be the same as that in the corresponding dimensions of the input.

theano/tensor/var.py
gvtulder: Yes, that's on purpose, because the boolean arrays are treated differently from integer arrays (see my other comment). An integer array counts for 1 dimension no matter what shape it is, but a boolean array counts for ndim dimensions. The loop a few lines before this does that counting.

theano/tensor/tests/test_subtensor.py
```diff
@@ -382,41 +391,66 @@ def test_boolean(self):
         assert_array_equal(numpy_n[:, mask], n[:, self.shared(mask)].eval())
         assert_array_equal(numpy_n[1:, mask], n[1:, mask].eval())
         assert_array_equal(numpy_n[:1, mask], n[:1, mask].eval())
+        assert_array_equal(numpy_n[1:, mask, np.newaxis], n[1:, mask, np.newaxis].eval())
+        assert_array_equal(numpy_n[np.newaxis, 1:, mask], n[np.newaxis, 1:, mask].eval())
+        assert_array_equal(numpy_inc_subtensor(numpy_n, [0, mask], 1),
+                           inc_subtensor(n[0, mask], 1).eval())
+        assert_array_equal(numpy_inc_subtensor(numpy_n, [Ellipsis, mask], 1),
+                           inc_subtensor(n[:, mask], 1).eval())
```
gvtulder: Good catch, that's supposed to be a `slice(None)`.

theano/tensor/var.py
gvtulder: I moved the test to Subtensor.convert() now.

It still requires a test here, though: if there is a boolean mask, it should be given to AdvancedSubtensor and not AdvancedSubtensor1, even if there is only one mask.

nouiz:
I got a lenghy IRL discussion with @lamblin Friday. Here is his comment that I agree with.

- If you remember, we go some bugs due to some optimization not checking the set_instead_of_inc props of IncSubtensor. Reusing the same old op for boolean indeixing  and advanced indexing risk that. We don't like that. So we would prefer to have a new AdvancedBooleanIndexing op. There could be a base class that both AdvancedBooleanIndexing and AdvancedIndexing to share much of the code.
- I don't understand why all tests pass when you didn't update the make_node and infer_shape of  AdvancedIndexing. Check your own example, one return 2d and one return 1d. Is there something I'm missing of tests are missing?

@lamblin don't care if we don't return now shapes errors as the new numpy 1.13. He don't see problem of using nonzeros in the graph for now. Theano already have some cases where we don't returns errors.

We try to make a Theano beta this week and rc next week. So we should merge new feature ideally this week. If you don't have time to finish this PR, can you make a new PR with the nonzero to have this in Theano 0.10?

We won't have time to finish this ourself as a few people are in vacation.

gvtulder:
> If you remember, we go some bugs due to some optimization not checking the set_instead_of_inc props of IncSubtensor. Reusing the same old op for boolean indeixing and advanced indexing risk that. We don't like that. So we would prefer to have a new AdvancedBooleanIndexing op.

I don't fully understand the reason, but I've now added a new `AdvancedBooleanSubtensor` op that shares a `BaseAdvancedSubtensor` class with `AdvancedSubtensor`. Is this more or less what you meant? (I don't know if it's necessary to have two classes for the gpu version as well?)

> I don't understand why all tests pass when you didn't update the make_node and infer_shape of AdvancedIndexing. Check your own example, one return 2d and one return 1d. Is there something I'm missing of tests are missing?

- For `infer_shape`, you're right: that wasn't tested. The `infer_shape` of `AdvancedSubtensor` doesn't do much anyway, but the special case that is implemented did give a problem with boolean indexing.

- `make_node` doesn't need to be changed, because it gets its shape information from `adv_index_broadcastable_pattern`, and [I did have to change that](https://github.com/Theano/Theano/pull/6207/files#diff-b8180324f78f23b3f0df24954e3d1e54R2104). That should solve the 2D/1D problem.

theano/tensor/subtensor.py
```diff
@@ -40,6 +41,16 @@ def __init__(self, *args):
         TypeError.__init__(self, *args)
 
 
+class AdvancedBooleanIndexingError(TypeError):
+    """
+    Raised when Subtensor is asked to perform advanced indexing with boolean masks.
+
+    """
+
+    def __init__(self, *args):
```
nouiz: You don't need this init method. Just use "pass" as the content of this class for python syntax.

theano/tensor/subtensor.py
```diff
@@ -2170,11 +2215,71 @@ def grad(self, inputs, grads):
         return [advanced_inc_subtensor(theano.tensor.zeros_like(x), gz,
```
nouiz: remove grad from the base class.

theano/tensor/var.py
nouiz: right

gvtulder:
Fixed.

theano/gpuarray/opt.py
```diff
@@ -1076,6 +1076,13 @@ def local_gpua_advanced_subtensor(op, context_name, inputs, outputs):
 
 
 @register_opt('fast_compile')
+@op_lifter([tensor.AdvancedBooleanSubtensor])
+@register_opt2([tensor.AdvancedBooleanSubtensor], 'fast_compile')
+def local_gpua_advanced_boolean_subtensor(op, context_name, inputs, outputs):
+    return GpuAdvancedSubtensor()
```
nouiz: All the PR look good, but I have one thing that make me hesitant. On the GPU, you have just one op that also does the boolean indexing. This is different then what we do. I'm not opposed to it, but I would prefer that someone else is also ok about this before merging.

Also, having just one class named GpuAdvancedBooleanSubtensor would be more consistent to what the CPU does. What do you think about this?

gvtulder:
I've now split the GpuAdvancedSubtensor into a boolean and non-boolean class. I have mixed feelings about this: it's nice that it mirrors the CPU structure more closely, but it's also a bit weird to have two classes that do exactly the same thing. The same holds for the CPU implementation, where the non-boolean op is just a restricted case of the boolean op. Since the CPU version has two classes, I guess the neatest solution is to map it to a similar structure on the GPU.

Note that I've also added the GpuAdvancedBooleanIncSubtensor op and a disabled optimization, but that op still has bugs. The tests fail if you enable the optimization. I think it's useful to have the optimization so that it won't be forgotten in the IncSubtensor repairs.

nouiz:
I'm also not sure what to do for the same reason. I would like to discuss
this with @abergeron or lamblin before merging for this reason.

I don't like the idea of having 2 ops that does exactly the same thing on
the GPU. If we go in that direction the nonbool op should need to raise an
error like on the CPU if bool are used with it.

What I think is a good middleground would be to just implement the bool
version on the GPU. So that keep the corresponding with the GPU and it make
it clear in the tests and to the developer that it should also support the
bool case. (then we don't implement the nonbool op on the GPU). For the non
bool optimization, the would continue to be done on the CPU as before.

What do you think of that?

gvtulder:
I see that a new error has appeared in Jenkins. Something to do with 16-bit floats. I'll have to look into that.

> I don't like the idea of having 2 ops that does exactly the same thing on
the GPU. If we go in that direction the nonbool op should need to raise an
error like on the CPU if bool are used with it.

Yes, that's what I had done now, both on the CPU and on the GPU. The nonbool versions of the ops are exactly the same as the bool versions, except that the nonbool ops run a `check_and_reject_bool(index)` to throw an error if there are any boolean indices.

> What I think is a good middleground would be to just implement the bool
version on the GPU. So that keep the corresponding with the GPU and it make
it clear in the tests and to the developer that it should also support the
bool case. (then we don't implement the nonbool op on the GPU). For the non
bool optimization, the would continue to be done on the CPU as before.

I'm not sure what you mean by this. Do you mean:
* CPU: keep `AdvancedSubtensor`, `AdvancedBooleanSubtensor`, `AdvancedIncSubtensor` and `AdvancedBooleanIncSubtensor`
* GPU: rename `GpuAdvancedSubtensor` and `GpuAdvancedIncSubtensor` to `GpuAdvancedBooleanSubtensor` and `GpuAdvancedBooleanIncSubtensor`

In that case I think I'd prefer it to be the way it is now: each CPU op maps to a `Gpu*` equivalent. Wouldn't removing/renaming existing ops cause problems?

theano/gpuarray/subtensor.py
```diff
@@ -634,21 +622,35 @@ def perform(self, node, inputs, out_):
...
         otype = GpuArrayType(dtype=rval.outputs[0].type.dtype,
                              broadcastable=rval.outputs[0].type.broadcastable,
                              context_name=ctx_name)
         x = as_gpuarray_variable(x, ctx_name)
-        y = as_gpuarray_variable(y, ctx_name)
-        return gof.Apply(self, [x, y] + rval.inputs[2:], [otype()])
+        return gof.Apply(self, [x] + rval.inputs[1:], [otype()])
+
 
+class BaseGpuAdvancedIncSubtensor(object):
```
nouiz: This look great. Can you make the non bool op raise an error if it receive a bool as input as on the CPU? Maybe just an attribute support_bool=True/False in each sub class and have the perform reuse that?

It is not great to have the CPU and GPU op not behave the same. It caused subtle user problem in the past (code that worked only on GPU, but not on CPU)

After that we should be ok to merge.

theano/gpuarray/subtensor.py
```diff
@@ -481,24 +481,46 @@ def c_code_cache_version(self):
         return (0,)
 
 
-class GpuAdvancedSubtensor(HideC, tensor.AdvancedSubtensor):
+def check_and_convert_boolean_masks(input, idx_list):
```
nouiz: Just an idea, if you make an attribute to the op as said in the other comment, mostly, you could have only one implementation of this function in the CPU code that have a parameter to raise or convert.

theano/gpuarray/subtensor.py
gvtulder: This is already checked, but it is done in `AdvancedSubtensor.make_node`. The `GpuAdvancedSubtensor.make_node` calls that function, so it's automatically checked and already raises `IndexError`s for boolean indices.

But it is a bit hidden and implicit. Two options to improve that:
* Repeat the check in `GpuAdvancedSubtensor.make_node` (by calling `tensor.subtensor.check_and_reject_bool`). Then the inputs are checked twice.
* Add a comment to `GpuAdvancedSubtensor.make_node` explaining that the checks are done in the CPU op's `make_node`.

theano/gpuarray/subtensor.py
gvtulder: The checks are done at `make_node` time, not at runtime in `perform`, so there is only one implementation. I thought it wouldn't do much harm to have the conversion in the `perform` for the non-bool op as well, because it just doesn't change anything and having a one-size-fits-all method keeps things simple.

nouiz: ok. thanks merging.

PR #6210, opened 2017-07-30
Add Sort Function ( Partition )

aeloyq:
modified:   theano/tensor/__init__.py
	modified:   theano/tensor/sort.py
	modified:   theano/tensor/var.py

### DIFF
--- a/doc/library/tensor/basic.txt
+++ b/doc/library/tensor/basic.txt
@@ -528,6 +528,8 @@ TensorVariable
     .. method:: nonzero_values(self)
     .. method:: sort(self, axis=-1, kind='quicksort', order=None)
     .. method:: argsort(self, axis=-1, kind='quicksort', order=None)
+    .. method:: partition(self, kth, axis=-1, kind='introselect', order=None)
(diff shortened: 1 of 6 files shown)


aeloyq:
Hi , 
I've add function partition(argpartition) to sort.py
Function partition is a wrapper of np.partition
This function is useful when you only need to get top kth value(or its index) along a given axes.
For example it takes 10% time and same effect when you need top 10 value of a tensor with shape (16L,300000L)
First:
`a=numpy.random.randint(0,10000,[16,300000])
`x=theano.shared()
For past you should:
`y=Tensor.sort(a)[:,:10]
Now you can:
`y=Tensor.partitial(a,10-1)[:,:10]

khaotik:
I have an old PR #5959 also utilizing `np.argpartition`. However there are a number of interface differences between the two PRs.

aeloyq:
Thank you @khaotik  for concern i'm reading it ~
BR

nouiz:
This PR seem a subset of https://github.com/Theano/Theano/pull/5959. Can you just use that PR until we merge it?

Otherwise, what advantage your PR have over the more complete one above?

PR #6213, opened 2017-07-31
Changed grad to L_op for cholesky and solve ops

juancamilog:
Addresses #5426 for the slinalg.py ops.

### DIFF
--- a/theano/tensor/slinalg.py
+++ b/theano/tensor/slinalg.py
@@ -93,7 +93,6 @@ class Cholesky(Op):
 
         """
 
-        x = inputs[0]
         dz = gradients[0]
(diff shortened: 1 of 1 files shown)


theano/tensor/slinalg.py
```diff
@@ -95,7 +95,7 @@ def grad(self, inputs, gradients):
 
         x = inputs[0]
         dz = gradients[0]
-        chol_x = self(x)
+        chol_x = outputs[0]
```
lamblin: With that change, `x` is not used any more:
```
F841 local variable 'x' is assigned to but never used
```

nouiz:
thanks

Issue #6160, opened 2017-07-17
Add synch to theano function


ReyhaneAskari:
Create a method that would synchronize the output of a theano function so that all the outputs would be available at the synch time.

ReyhaneAskari:
The original ticket is here: https://github.com/Theano/Theano/issues/6160

PR #6214, opened 2017-07-31
sync attribute added to theano function

ReyhaneAskari:
Regarding ticket https://github.com/Theano/Theano/issues/6160, I added a sync attribute to theano function. I still need to figure out how to sync the outputs in addition to the shared variables.

### DIFF
--- a/theano/compile/tests/test_function_module.py
+++ b/theano/compile/tests/test_function_module.py
@@ -909,6 +909,12 @@ def test_empty_givens_updates():
 
 
 def test_sync_update():
+    # This test if sync_update work. This can only be tested when
+    # there is a GPU.  To test if we really sync, we compare a case we
+    # can run in parallel GPU and CPU computation. Then we sync to
+    # disable that parallel computation. Then we assert the time is
+    # higher.
+
     import theano.gpuarray.tests.config
     if theano.gpuarray.pygpu_activated:
         sizes = [100, 500, 1000, 2000, 5000, 10000, 20000, 40000]
--- a/theano/compile/tests/test_function_module.py
+++ b/theano/compile/tests/test_function_module.py
@@ -924,14 +924,13 @@ def test_sync():
         f(x_)
         t_0 = time.time()
         for i in range(1000):
+            f.sync_shared()
             f(x_)
         t_1 = time.time()
-        f.sync_shared()
-        t_2 = time.time()
         for i in range(1000):
             f(x_)
-        t_3 = time.time()
-        assert (t_1 - t_0) > (t_3 - t_2)
+        t_2 = time.time()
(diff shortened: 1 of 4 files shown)


theano/compile/tests/test_function_module.py
```diff
@@ -907,6 +908,34 @@ def test_empty_givens_updates():
...
+
+        updates = [(w, w + T.dot(w, x) + T.dot(w, w))]
+
+        f = theano.function([x], y, updates=updates)
+        f.sync_shared()
+        g = theano.function([x], y, updates=updates)
+        x_ = np.random.rand(2000, 2000).astype('float32')
+        f(x_)
+        g(x_)
+        t_0 = time.time()
```
nouiz: I don't understand this timing. I would do something like:

t0 = ..
for i in ...
    f()
    f.sync_update()
t1 = ...
for i in range ...
   f()
t2 = ...
assert t1-t0 is bigger then t2-t1.

the sync_update should slow down computation. I would use a small range then 1000. Maybe 10 or 50.

theano/compile/tests/test_function_module.py
```diff
@@ -907,6 +908,34 @@ def test_empty_givens_updates():
     function([theano.In(x)], y, updates={})
 
 
+def test_sync():
+    if theano.config.device == 'cuda' and theano.gpuarray.pygpu_activated:
```
ReyhaneAskari: Thanks. Done.

theano/compile/tests/test_function_module.py
ReyhaneAskari: The difference is not really significant, that's why I was using a large range. It's like 33.7922520638 and 33.9593539238. I think it's not being synced at all in this way. I'm thinking maybe we need to do this in the call and when we have access to the values. What to you think?

nouiz: You need to call sync_shared() inside one of the loop. This is what will make a difference between both loop.

ReyhaneAskari:
Thanks @nouiz ,with that change, the timing is 12.9660871029 and 13.467031002. Do you think that's significant enough?

nouiz:
No. I'm actively working on it.

nouiz:
I updated this PR with to have the test be meaningful. There was many problem with it. The function was only on CPU. The shared var was also on CPU. The function had an output, as the sync don't sync it, to easy the test I remove it. I made sure there is only 1 op that is inplace to don't have GPU memory allocation during the test.

I also modified how the timing is done to have the sync at the right place.

I also updated the timing to have CPU and GPU compute in parallel vs sequential in the other loop. This is the only way to get significant speed difference I found.

Also, I select the size dynamically to have the test more resistant to GPU changes. I tested it on a 750 and P100 SMX2.

Issue #5736, opened 2017-03-21
unpickled function slowed by excessive calls to fromnumeric.py (any)


astooke:
Hi,
I've encountered a significant slowdown in a worker process when calling a Theano function pickled from a different process, and it comes from an excessive number of calls to `any` from `fromnumeric.py`.  (In the cProfile report this is one level below `function_module.__call__`, same level where `filter` shows up.)

In the master process, `any` is called roughly once per function call, but in the process with the unpickled function, it is called about 20 times per function call, and this becomes almost 40% of the total function time.

The function in question is the training function from the MNIST example off the Lasagne repo, run with the `cnn` option.  And I've just pulled the latest Theano and it's still happening.  Happy to provide any more details I can, but so far this is all I have.  You can run the example [here](https://github.com/astooke/synkhronos/tree/master/demos), if you have at least 2 GPUs.  Right now it's set to make a cProfile in the worker, and if you also cProfile the main call to `lasagne_mnist.py`, you'll see the difference.

p.s. This was sneaky!  I only found it because I noticed that the example code was running much slower when using NCCL to reduce the training error output (a scalar!) compared to only keeping the value from the master process (the example calls a training function with parameter updates in a loop and accumulates training error as output).  NCCL has some overhead but using CPU-based communication had the same result!  This despite there being a call to NCCL `all_reduce()` on all the training parameters at every loop iteration, after a barrier at the end of function execution.  When not collecting the result from the worker but forcing `np.asarray()` on its output, which is a GPU variable, the slowdown is apparent.   I guess when not bringing the worker result to the CPU, the worker's actual GPU computation time was hidden.  

This makes me worry that a call to `all_reduce` on the network parameters right after the training function call might lead to race conditions on the parameters?  Such as if some update rule within the Theano function hasn't started yet, before the functions return, pass the barrier, and make the NCCL call?   I haven't noticed any degraded performance but I really have not run very many cases.

lamblin:
Do you still measure a large amount of time spent there if you set `CUDA_LAUNCH_BLOCKING=1` in the environment first?
Converting a gpuarray into a numpy ndarray forces a wait on all the scheduled GPU kernels, so a profiler would show a large amount of time spent there, but it actually is waiting for the previous operations to actually be executed.
I think `all_reduce` will also wait for the previous computations to be finished.

astooke:
[...]

PR #6215, opened 2017-07-31
Fix the copy for aliased function arguments

abergeron:
Follow up for https://github.com/Theano/Theano/pull/4982


lamblin:
This may actually be the reason too many comparisons happened in #5736.

PR #6218, opened 2017-07-31
Alloc ops params

### LINKED ISSUE #6135: Use params for [Gpu]Alloc[Empty] ops
Some props for these ops may be converted as op params to reduce C code.

- theano.gpuarray.basic_ops.GpuAlloc
  - `memset_0`
- theano.gpuarray.basic_ops.GpuAllocEmpty
  - `dtype` (if possible, use typecode instead)
- theano.tensor.basic.AllocEmpty
  - `dtype` (use typecode instead)

tfjgeorge:
fix #6135 @notoraptor

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -970,7 +970,10 @@ class GpuAllocEmpty(HideC, AllocEmpty):
     def __init__(self, dtype, context_name):
         self.dtype = dtype
         self.context_name = context_name
-        self.typecode = gpuarray.dtype_to_typecode(self.dtype)
+
+    @property
+    def typecode(self):
+        return gpuarray.dtype_to_typecode(self.dtype)
 
     def get_params(self, node):
         return self.params_type.get_params(context=get_context(self.context_name),
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -8,6 +8,7 @@ import theano
 from theano import Op, Apply, Type, Variable
 from theano import tensor, config
 from theano.gradient import grad_undefined
+from theano.scalar import bool as bool_t
 from theano.tensor.basic import (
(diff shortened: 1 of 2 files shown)


tfjgeorge:
Whoops currently it breaks the op lifter :

```
File "/u/georgeth/lib/Theano/theano/gpuarray/opt.py", line 630, in local_gpua_alloc_empty
    return GpuAllocEmpty(context_name=context_name, **op._props_dict())(*inputs)
TypeError: __init__() got an unexpected keyword argument 'typecode'
```

I will look into this

theano/gpuarray/basic_ops.py
```diff
@@ -968,16 +970,22 @@ class GpuAllocEmpty(HideC, AllocEmpty):
     Allocate uninitialized memory on the GPU.
 
     """
-    __props__ = ('dtype', 'context_name')
+    __props__ = ('dtype', 'context_name', 'typecode')
```
notoraptor: @tfjgeorge It may be because of this line. As `typecode` is just a class property (Python class property), we don't need to set it as a Theano op property. So remove `'typecode'` from `__props__`.

theano/tensor/basic.py
```diff
@@ -6632,13 +6633,18 @@ def perform(self, node, inputs, outputs):
 class AllocEmpty(gof.Op):
     """Implement Alloc on the cpu, but without initializing memory."""
 
-    __props__ = ("dtype",)
+    __props__ = ("dtype", "typecode")
```
notoraptor: Here too.

notoraptor:
ok to test

lamblin:
You will also need to rebase on the current master for Travis to pass.

nouiz:
This introduce a problem in DebugMode. I fixed it in the first commit of this PR:

https://github.com/Theano/Theano/pull/6289

Issue #6144, opened 2017-07-12
Compile Failure and Performance Issue w/#5881


HapeMask:
After #5881 was merged (https://github.com/Theano/Theano/commit/4f3a52a55a31ce82e737d44f3025a26da247958a), I noticed a significant (10x) performance decrease while training a particular CNN.

I narrowed things down to the following small example: https://gist.github.com/HapeMask/01c36d97ea7c9afd4c4383ff014a9a18
The key is that the loss comes from an adv. subtensor, otherwise the bug isn't present.

The code is running on the gpu (profile mode shows 99.9% GPU time), but nvidia-smi shows a GPU utilization of only 8%. A quick check suggests that the time is being spent here: https://github.com/Theano/Theano/blob/master/theano/gpuarray/subtensor.py#L691

As a separate note, if you take the code in the above gist and remove all use of ```valid``` in ```nll()```, the sample fails to compile with the following error:

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/Theano-0.10.0.dev1-py3.6.egg/theano/compile/function_module.py", line 883, in __call__
    self.fn() if output_subset is None else\
  File "/usr/lib/python3.6/site-packages/Theano-0.10.0.dev1-py3.6.egg/theano/gof/op.py", line 892, in rval
    r = p(n, [x[0] for x in i], o)
  File "/usr/lib/python3.6/site-packages/Theano-0.10.0.dev1-py3.6.egg/theano/gpuarray/subtensor.py", line 696, in perform
    x_flat.__setitem__(i, tmp)
  File "pygpu/gpuarray.pyx", line 2096, in pygpu.gpuarray.GpuArray.__setitem__ (pygpu/gpuarray.c:26904)
  File "pygpu/gpuarray.pyx", line 346, in pygpu.gpuarray.array_setarray (pygpu/gpuarray.c:6287)
ValueError: b'Dimension error. a->nd = 0, v->nd = 1'
```

This is not the case at https://github.com/Theano/Theano/commit/86c195d221ffdb3bc024cd99f073fd357481562e, where the code compiles and runs as expected.

aam-at:
My code is using [multiclass_hinge_loss](https://github.com/Lasagne/Lasagne/blob/master/lasagne/objectives.py#L302). After merging #5881, the code is 40x times slower. Gpu utilization is less than 10 percent. At 86c195d221ffdb3bc024cd99f073fd357481562e, no performance degradation.

[...]

PR #6220, opened 2017-07-31
Updates in GPU indexing

lamblin:
- disable lifter for AdvancedIncSubtensor (see #6144)
- update tests to execute more on GPU


lamblin:
I removed the change that exposed an issue in GpuIncSubtensor.
See #6221 for that one.

PR #6221, opened 2017-07-31
Use E_PADSHAPE flag in GPU Incsubtensor

lamblin:
This needs E_PADSHAPE to be present in libgpuarray.

### DIFF
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -387,7 +387,7 @@ int sub_setarray(GpuArray *dst, GpuArray *src) {
           void *args[2];
           args[0] = &zview->ga;
           args[1] = &%(x)s->ga;
-          if (GpuElemwise_call(iadd, args, GE_BROADCAST) != GA_NO_ERROR) {
+          if (GpuElemwise_call(iadd, args, GE_BROADCAST | GE_PADSHAPE) != GA_NO_ERROR) {
             PyErr_SetString(PyExc_RuntimeError, "Error doing inplace add");
             Py_DECREF(zview);
             %(fail)s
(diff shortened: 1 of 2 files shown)


theano/gpuarray/opt.py
```diff
@@ -1095,9 +1095,11 @@ def local_gpua_advanced_incsubtensor1(op, context_name, inputs, outputs):
             set_instead_of_inc=set_instead_of_inc)
 
 
-@register_opt('fast_compile')
-@op_lifter([tensor.AdvancedIncSubtensor])
-@register_opt2([tensor.AdvancedIncSubtensor], 'fast_compile')
+# Do not register this optimization for now, as it slows down the
+# execution by a lot in most cases.
```
nouiz: I don't agree with the work "most", but don't loose time to change it. I would more tell "in important case".

Can you confirm that the timing you did is after the optimization you did in the perform?

theano/gpuarray/subtensor.py
```diff
@@ -389,7 +389,7 @@ def add_to_zview(self, nodename, x, fail):
           void *args[2];
           args[0] = &zview->ga;
           args[1] = &%(x)s->ga;
-          if (GpuElemwise_call(iadd, args, GE_BROADCAST) != GA_NO_ERROR) {
+          if (GpuElemwise_call(iadd, args, GE_BROADCAST | GE_PADSHAPE) != GA_NO_ERROR) {
```
nouiz: Note, @abergeron added the python interface for this. So we can now update the perform() with the same fix in the 2 ops.

Also, if we want to merge this earlier, we could do the padshape in the graph or in the perform/c_code. I don't suggest doing that, just in case you find the idea useful or if we start to not be able to wait for the 0.6.10 release.

abergeron: The perform() doesn't need to be fixed since the python interface will set this flag to True by default.

theano/gpuarray/opt.py
lamblin: Yes, I timed it afterwards.

nouiz:
I rebased this PR.

jenkins test this.

Issue #6047, opened 2017-06-16
Move C codes into a sub-folder `c_code` for each module


notoraptor:
As suggested by @nouiz , to help not have too many files in directories, C code files could be moved into separate sub-folder `c_code` in each module. For example, `theano/gpuarray/c_code` for all C code files in `theano/gpuarray` module (as done here: https://github.com/Theano/Theano/commit/5b6d3ac7bb74ef02e86d9091c80ca00e1f5c7e6d ).

What do you think, @abergeron @lamblin ?

abergeron:
I am ok with moving to a sub-folder.  One thing to watch out for is to make sure that those sub-folders are included with installs.

PR #6223, opened 2017-08-01
Add "c_code" folders to Theano installations

### LINKED ISSUE #6222: `.../c_code/dimshuffle.c` missing from setup.py?
A Lasagne user just reported that
```bash
pip install --upgrade --no-deps https://github.com/Theano/Theano/archive/master.zip
```
lead to the following error (I assume after installation, when trying to import it):
```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/XXXX/anaconda/lib/python3.5/site-packages/theano/tensor/c_code/dimshuffle.c
```
The file was added in #6174 and is included in the `master.zip` snapshot, but maybe it's missing in `setup.py` and thus not copied over? You wouldn't notice when doing a development/editable install. The user solved it by downloading `tensor/c_code/dimshuffle.c` and `gpuarray/c_code/dimshuffle.c` manually from github.

Disclaimer: I haven't verified this behavior myself.

notoraptor:
To do that, I converted current `c_code` folders to python modules (I don't find other solution, currently). This should fix #6222 . I had forgotten that @abergeron warned about it in #6047.

@nouiz @lamblin


f0k:
> To do that, I converted current `c_code` folders to python modules (I don't find other solution, currently). 

But that will introduce `theano.gpuarray.c_code` as an empty module. Can't you just adapt the `find_packages` function in `setup.py` (https://github.com/Theano/Theano/blob/master/setup.py#L62) so it not only looks for `__init__.py`, but also for `c_code`?

f0k:
Thank you for updating the PR, but it seems this doesn't work, I'm just testing it. I've found another way. Will send a PR.

notoraptor:
@f0k  It works for me with new changes that you suggested (ie. using `find_packages()`):

```
pip install --upgrade --no-deps https://github.com/notoraptor/Theano/archive/fix-setup-for-c-code-folders.zip
```

f0k:
Funny. Yes, I can confirm it works. I tried it locally before and it didn't work, but now it does, not sure what was wrong. I assumed distutils still expects a `__init__.py`, as mentioned in the documentation: https://docs.python.org/2/distutils/setupscript.html#listing-whole-packages. But apparently it's not so strict, or it's different in setuptools.

In any case, the officially supported way is to use `package_data` instead. That would be:
```diff
diff --git a/setup.py b/setup.py
index bf26983..26b65ed 100755
--- a/setup.py
+++ b/setup.py
@@ -167,12 +167,12 @@ def do_setup():
               'test': ['nose>=1.3.0', 'nose-parameterized>=0.5.0', 'flake8<3'],
               'doc': ['Sphinx>=0.5.1', 'pygments']
           },
-          package_data={
-              '': ['*.txt', '*.rst', '*.cu', '*.cuh', '*.c', '*.sh', '*.pkl',
-                   '*.h', '*.cpp', 'ChangeLog'],
-              'theano.misc': ['*.sh'],
-              'theano.d3viz' : ['html/*','css/*','js/*']
-          },
+          package_data=dict([
+              ('', ['*.txt', '*.rst', '*.cu', '*.cuh', '*.c', '*.sh', '*.pkl',
+                    '*.h', '*.cpp', 'ChangeLog']),
+              ('theano.misc', ['*.sh']),
+              ('theano.d3viz', ['html/*','css/*','js/*']),
+          ] + [(package, ['c_code/*.c']) for package in find_packages()]),
           entry_points={
               'console_scripts': ['theano-cache = bin.theano_cache:main',
                                   'theano-nose = bin.theano_nose:main']
```

notoraptor:
@f0k Does your patch work ? If so, you can send your PR and I will close that current one.

f0k:
It works, yes, but I'm not sure if it's nicer than yours :)

notoraptor:
Ok!

@abergeron @nouiz @lamblin what do you think ?

abergeron:
I think using package_data is a better solution in theory, but I've had a bunch of problem with libgpuarray where package_data would end up in a strange folder somewhere else in the filesystem.

I'm just worried that some version of setuptools and/or distutils is going to check for `__init__.py` and complain or refuse to work when it's not there.

notoraptor:
@abergeron So, which solution should we use ? Convert any `c_code` folder to a module by adding an `__init__.py` file ?

f0k:
I think I found another way, wait a minute...

f0k:
Yes, it works. PR underway.

abergeron:
I think the package_data solution from @f0k is better that trying to fake python packages.

We would just need to test it on a wide variety of machines to make sure it doesn't break on some weird version of setuptools/disutils.

notoraptor:
@abergeron Ok, so PR #6225 will do the work.

PR #6225, opened 2017-08-01
Install c_code/* files for every submodule

### LINKED ISSUE #6222: `.../c_code/dimshuffle.c` missing from setup.py?

f0k:
Fixes #6222. Closes #6223.

Reading the source code of distutils (https://github.com/python/cpython/blob/master/Lib/distutils/command/build_py.py#L97-L132), I found that a much simpler option is to make use of the `''` entry in `package_data` -- it specifies globbing patterns that are applied to every package specified in `packages`. We can just add `c_code/*` there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -169,7 +169,7 @@ def do_setup():
           },
           package_data={
               '': ['*.txt', '*.rst', '*.cu', '*.cuh', '*.c', '*.sh', '*.pkl',
-                   '*.h', '*.cpp', 'ChangeLog'],
+                   '*.h', '*.cpp', 'ChangeLog', 'c_code/*'],
               'theano.misc': ['*.sh'],
               'theano.d3viz' : ['html/*','css/*','js/*']
           },


notoraptor:
I confirm that it works.

nouiz:
jenkins test windows mac

lamblin:
Can you rebase? It looks like Jenkins and Travis are using their config files from this branch, and not the merge with the master, so they are not installing the package (`parameterized`) that the code expects.

f0k:
Rebased! But at least Travis should be testing the merge with upstream: https://docs.travis-ci.com/user/pull-requests/#How-Pull-Requests-are-Tested

notoraptor:
ok to test

f0k:
Tests passed. Does somebody have to say "jenkins test windows mac" again?

notoraptor:
Not sure if jenkins tests theano installation itself.

But: jenkins test windows mac

lamblin:
Apparently, the code tested by Travis comes from the merge head, but the .travis.yml file it uses is the one in the branch, for some reason.

f0k:
> but the .travis.yml file it uses is the one in the branch, for some reason.

Aha... maybe that's a bug in Travis. Anyway, all tests have passed, should be good to merge now so the bleeding-edge Theano can be installed properly again.

Issue #6200, opened 2017-07-24
Error during compilation


Rabia-Noureen:
Sir thank you for providing us such a great system. I was trying to run the my code in Anaconda Prompt and cmd but i got the following error on Windows 7 32 bit-python 2.7 with numpy and theano installed.
00878 {
00879 // save references to outputs prior to updating storage contai
ners
00880 assert (self->n_output_vars >= self->n_updates);
00881 Py_DECREF(rval);
00882 rval = PyList_New(self->n_output_vars);
00883 for (int i = 0; i < (self->n_output_vars); ++i)
00884 {
00885 Py_ssize_t src = self->output_vars[i];
00886 PyObject * item = PyList_GetItem(self->var_value_cells[src
], 0);
00887 if ((output_subset == NULL || output_subset[i]) &&
00888 self->var_computed[src] != 1)
00889 {
00890 err = 1;
00891 PyErr_Format(PyExc_AssertionError,
00892 "The compute map of output %d should cont
ain "
00893 "1 at the end of execution, not %d.",
00894 i, self->var_computed[src]);
00895 break;
00896 }
00897 Py_INCREF(item);
00898 PyList_SetItem(rval, i, item);
00899 }
00900 }
00901
00902 if (!err)
00903 {
00904 // Update the inputs that have an update rule
00905 for (int i = 0; i < self->n_updates; ++i)
00906 {
00907 PyObject* tmp = PyList_GetItem(rval, self->n_output_vars -
self->n_updates + i);
00908 Py_INCREF(tmp);
00909 Py_ssize_t dst = self->update_storage[i];
00910 PyList_SetItem(self->var_value_cells[dst], 0, tmp);
00911 }
00912 }
00913 }
00914
00915 /*
00916 Clear everything that is left and not an output. This is needed
00917 for lazy evaluation since the current GC algo is too conservative
00918 with lazy graphs.
00919 /
00920 if (self->allow_gc && !err)
00921 {
00922 for (Py_ssize_t i = 0; i < self->n_vars; ++i)
00923 {
00924 int do_cleanup = 1;
00925 if (!self->var_has_owner[i] || !self->var_computed[i])
00926 continue;
00927 for (int j = 0; j < self->n_output_vars; ++j)
00928 {
00929 if (i == self->output_vars[j])
00930 {
00931 do_cleanup = 0;
00932 break;
00933 }
00934 }
00935 if (!do_cleanup)
00936 continue;
00937 Py_INCREF(Py_None);
[...]

PR #6251, opened 2017-08-02
Update cmodule.py

utimatu:
This is to fix Issue #6200.  

This is a hack, and it might not be very precise, but it works on my system.


nouiz:
We won't make a 0.9.X release. We are doing an 0.10 release. So there is no need to merge to the 0.9.X branch. Do you have the problem with the master of Theano?

If not, there is nothing to change. If you still have problem with the master of Theano, moving this PR to it would be useful.

nouiz:
I'm convinced this is now supported in the dev version of Theano. So I'll close this PR. We already released Theano 0.10beta1.

If you still have problem with that release, open a new issue a new PR to the master of Theano.

thanks.

PR #6252, opened 2017-08-02
phase out outdated conv3d

Faruk-Ahmed:
#5079


Faruk-Ahmed:
The old conv3d appears to be still used in master here:
https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/opt.py#L281

What should be done about it?

nouiz:
Delete that method and detele the code that use it elsewhere in that file.

nouiz:
After that and the travis passed (so we know nothing else use it), you can do an updated version of the last step in this comment:

https://github.com/Theano/Theano/issues/5079#issuecomment-274965547

- [ ] add back the files, with only an ERROR message that instructs users to use the new functions:

    theano.tensor.nnet.conv3d(..., filter_flip=False) instead of conv3d(...)
    theano.tensor.nnet.conv3d_grad_wrt_inputs(..., filter_flip=False) instead of convTransp3d(...)
    theano.tensor.nnet.conv3d_grad_wrt_weights(..., filter_flip=False) instead of convGrad3d(...)

theano/tensor/nnet/opt.py
```diff
@@ -254,315 +251,6 @@ def local_conv2d_cpu(node):
...
-        return None
-    if node.op.border_mode not in ['full', 'valid']:
-        return None
-    if not node.op.filter_flip:
-        # Not tested yet
-        return
-    if node.op.num_groups > 1:
-        return None
-
-    if node.op.border_mode == 'valid' and \
```
nouiz: By removing this optimization, you also remove the usage of an old op called Conv (from the file nnet/conv.py). You could have must changed what is inside this if to always return False.

But maybe we should also remove that old conv? @abergeron what do you think? This old conv in rare case faster on CPU... But I don't it isn't work to keep it. Do you agree?

If so, @Faruk-Ahmed, also remove the file conv.py.

nouiz: This will also request that some tests are removed. There is at least the file nnet/tests/test_conv3d.py that need to be removed.

nouiz:
Test failed due to the documentation:

https://travis-ci.org/Theano/Theano/jobs/263148438#L1247

nouiz:
jenkins test this

nouiz:
A rebase was needed. I squashed all commits to help in the rebase. Small fixup at the same time.

jenkins test this

PR #6253, opened 2017-08-03
Fix missing scipy and warn that Theano will have a c++ compiler mandatory

nouiz:


theano/configdefaults.py
```diff
@@ -533,6 +533,10 @@ def warn_cxx(val):
              in_c_key=False)
 del param
 
+if not config.cxx:
+    warnings.warn("DeprecationWarning: there is no c++ compiler."
+                  "This is deprecated and Theano 0.11 a c++ compiler will be mandatory")
```
abergeron: and Theano -> and with Theano

PR #6256, opened 2017-08-03
Tweak the dnn config again to cover more cases.

abergeron:


f0k:
I can't comment on all of the changes, but https://github.com/Theano/Theano/pull/6256/files#diff-39f974ded3d488392ace6d4d74baeeefR398 fixes the problem I reported in https://github.com/Theano/Theano/pull/6147#issuecomment-319983031.

PR #6264, opened 2017-08-04
Remove config.gpuarray.sync from Theano code.

### LINKED ISSUE #6249: How to manage `config.gpuarray.sync` ?
In some Ops, it is used inside C code in string formatting:
```python
"""
if (%(sync)d) {...}
""" % {'sync': config.gpuarray.sync}
```

But in other Ops it is used as a condition to append C code:
```python
if config.gpuarray.sync:
    code += "..."
```

What is the best design to use ? Always use it in string formatting ? Find a way to always use it as an op param ? Always use it as a Python condition to append C code ? Or isn't it a problem if it is used in different ways across Theano code ?

@nouiz @lamblin @abergeron

notoraptor:
This PR is intended to fix #6249 . I think these changes won't be included in the next release, but we can start working on it!

Let's see if all Jenkins tests pass.

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -192,14 +192,6 @@ AddConfigVar(
     BoolParam(True, allow_override=False),
     in_c_key=False)
 
-
-AddConfigVar('gpuarray.sync',
-             """If True, every op will make sure its work is done before
-                returning.  Setting this to True will slow down execution,
-                but give much more accurate results in profiling.""",
-             BoolParam(False),
-             in_c_key=True)
-
 AddConfigVar('gpuarray.preallocate',
              """If negative it disables the allocation cache. If
              between 0 and 1 it enables the allocation cache and
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -192,6 +192,17 @@ AddConfigVar(
     BoolParam(True, allow_override=False),
     in_c_key=False)
 
+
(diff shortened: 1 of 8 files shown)


theano/configdefaults.py
```diff
@@ -192,14 +192,6 @@ def filter(val):
     BoolParam(True, allow_override=False),
     in_c_key=False)
 
-
-AddConfigVar('gpuarray.sync',
```
nouiz: You should make it raise an error if used. Keep it. Update the docstring to tell it is removed. Give a lambda as a parameter to the boolparam that raise an error.

theano/configdefaults.py
```diff
@@ -193,11 +193,14 @@ def filter(val):
...
+    return False
+
 AddConfigVar('gpuarray.sync',
-             """If True, every op will make sure its work is done before
-                returning.  Setting this to True will slow down execution,
-                but give much more accurate results in profiling.""",
-             BoolParam(False),
+             """This flag is deprecated and will be removed in next Theano release.""",
+             ConfigParam(False, allow_override=False, filter=deprecated_gpuarray_sync),
              in_c_key=True)
```
abergeron: Perhaps this flag can be completely removed or at least `in_c_key=False`

notoraptor: Done here: https://github.com/Theano/Theano/pull/6281

Issue #5047, opened 2016-10-03
Update/clean up conv doc


nouiz:
I would remove reference to our own fft convolution and pylearn2 convolution in the convolution doc:

http://deeplearning.net/software/theano/library/tensor/nnet/conv.html

They are slower and don't offer better feature then cudnn. So this make the doc more complicated for no good reason.

Do this after/based on this PR to help prevent conflict.

https://github.com/Theano/Theano/pull/4862/files

nouiz:
If people don't agree to remove them, I would move them to a deprecated section at the end of this file.

PR #6267, opened 2017-08-05
3D Grouped Convolutions

affanv14:
Tasks:

- [x] Implement AbstractConv version
- [x] Implement CorrMM version
- [x] Implement GPUCorrMM version

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -516,13 +516,13 @@ class BaseGpuCorrMM(CGpuKernelBase):
         # flops for any direction, sampling, padding, and border mode
         inputs, filters = inp
         outputs, = outp
-        assert inputs[1] == filters[1]
+        assert inputs[1] == (filters[1] * self.num_groups)
         # nb mul and add by output pixel
         flops = filters[2] * filters[3] * 2
         # nb flops by output image
         flops *= outputs[2] * outputs[3]
         # nb patch multiplied
-        flops *= inputs[1] * filters[0] * inputs[0]
+        flops *= inputs[1] * filters[0] * inputs[0] / self.num_groups
         return flops
 
     def c_headers(self):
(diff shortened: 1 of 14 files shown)


nouiz:
Note cudnn v7 added support for group convolution for 2d and 3d convolution. I think it would be better to don't spend time making conv3d support group convolution for cudnn in this PR. After this PR, you or someone else could update the conv to compile with  cudnn v7 and use their group convolution.

nouiz:
I did the review up to now and it look good.

nouiz:
I checked cudnn v7 group conv doc. It is implemented for 3d and 2d conv for forward and both grad. So the person that will add support for should not loop for num_group and should pass to it the right parameter.

affanv14:
I just checked out the cudnn docs, I should be able to handle that. I can start working on it in a new PR.

nouiz:
Someone else should make tomorrow a PR to make Theano compile with cudnn v7. I would prefer that you wait for this PR to be made. Otherwise it would duplicate the work.

If you have nothing to do, could you do this documentation issue in the mean time? https://github.com/Theano/Theano/issues/5047

Do you think you have finished this PR or there is more change you need to do?

affanv14:
> Someone else should make tomorrow a PR to make Theano compile with cudnn v7. I would prefer that you wait for this PR to be made. Otherwise it would duplicate the work.

Ok, no problem.
> If you have nothing to do, could you do this documentation issue in the mean time?

I still have separable conv3d to do, which shouldn't take more than two days, I can tackle the documentation after that. Then I can pick up cudnn v7 if it is has been merged by then.

> Do you think you have finished this PR or there is more change you need to do?

A few more changes, like skipping old cpu convolution and also skipping cudnn if num_groups is greater than 1

affanv14:
I finished making changes and also did the rebase

nouiz:
jenkins test that

nouiz:
jenkins test this

theano/gpuarray/blas.py
```diff
@@ -1102,11 +1105,12 @@ def pad(self):
...
-        return '%s{%s, %s, %s}' % (
+        return '%s{%s, %s, %s, %s}' % (
             self.__class__.__name__,
             self.border_mode,
             str(self.subsample),
-            str(self.filter_dilation))
+            str(self.filter_dilation),
+            str(self.num_groups))
 
     def flops(self, inp, outp):
```
nouiz: I forgot, do this method need an update?

affanv14: I seem to have missed that. I will update that soon for both 2d and 3d.

nouiz:
jenkins test this

PR #6068, opened 2017-06-22
Make the Theano c code cache keys forward compatible

nouiz:
In case we remove the md5 from the key and in case An ops attribute change and the c code cache version isn't changed.

PR #6271, opened 2017-08-07
Use sha256 for config hash to be FIPS compliant and make Theano cache more forward compatible

nouiz:
The first part (sha256) if the rebase/finish up from: https://github.com/Theano/Theano/pull/5916

As this force the recompilation of all c code, I also include in this PR gh-6068 to have only 1 PR that force the recompilation. This commit help the cache don't collide if we add props to an op but forget to bump the version number.

### DIFF
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -377,7 +377,7 @@ def is_same_entry(entry_1, entry_2):
 
 def get_module_hash(src_code, key):
     """
-    Return a SHA256 hash that uniquely identifies a module.
+    Return a SHA256 (for FIPS compatibility) hash that uniquely identifies a module.
 
     This hash takes into account:
         1. The C source code of the module (`src_code`).
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -416,10 +416,10 @@ def get_module_hash(src_code, key):
             to_hash += list(key_element)
         elif isinstance(key_element, string_types):
             if key_element.startswith('md5:'):
-                # This is actually a sha256 hash of the config options. 
-				# Ref PR#5916. String and function names will be updated in 
-				# future release.
-				# We can stop here.
+                # This is actually a sha256 hash of the config options.
+                # Ref PR#5916. String and function names will be updated in
+                # future release.
+                # We can stop here.
                 break
             elif (key_element.startswith('NPY_ABI_VERSION=0x') or
                   key_element.startswith('c_compiler_str=')):
@@ -449,7 +449,7 @@ def get_safe_part(key):
     assert version
 
     # Find the hash part, which is using sha256, not md5.
-	# Instances of md5 will be replaced in future release.
+    # Instances of md5 will be replaced in future release.
     c_link_key = key[1]
     # In case in the future, we don't have an md5 part and we have
     # such stuff in the cache.  In that case, we can set None, and the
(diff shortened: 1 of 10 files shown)


theano/gof/cmodule.py
```diff
@@ -377,7 +377,7 @@ def is_same_entry(entry_1, entry_2):
 
 def get_module_hash(src_code, key):
     """
-    Return an MD5 hash that uniquely identifies a module.
+    Return a SHA256 (for FIPS compatibility) hash that uniquely identifies a module.
```
abergeron: I don't think that we care about FIPS which is a security standard for encryption in the US government.

I would avoid referencing it.

theano/gof/cmodule.py
```diff
@@ -415,9 +415,13 @@ def get_module_hash(src_code, key):
...
-            if key_element.startswith('md5:'):
-                # This is the md5 hash of the config options. We can stop
-                # here.
+            if (key_element.startswith('md5:') or
+                    key_element.startswith('sha256:') or
+                    key_element.startswith('hash:')):
+                # This is actually a sha256 hash of the config options.
+                # Currently, we still keep md5 to don't break old Theano.
+                # We add 'sha256:' and 'hash:' so that when we change it in
+                # the futur, it won't break this version of Theano.
```
abergeron: I will admit that I don't understand why we need to rifle through the key to find the hash.  Isn't it at a defined position where we could just grab it?

Also, I think we should just add support for `hash:` instead of listing the algorithm used.  That way we can swap out in the future and not have to update this code.

theano/gof/utils.py
```diff
@@ -546,28 +546,28 @@ def remove(predicate, coll):
...
 
 else:
     import hashlib
 
     def hash_from_code(msg):
         try:
-            return hashlib.md5(msg).hexdigest()
+            return hashlib.sha256(msg).hexdigest()
         except TypeError:
             assert isinstance(msg, np.ndarray)
```
abergeron: Under which circumstance would a code string be a numpy array?  This doesn't make sense.  It also appear that this can't be the case under python 3 for some reason.

I would unify the two functions since I'm sure that python2 will work with the python 3 code if we swap `str` for `unicode`

theano/gof/cmodule.py
nouiz: We only care about it because it allow to use Theano on computer that are configured to be FIPS compatible from what I understand.

I'm ok to not reference it as we don't make any garanty about it in Theano.

theano/gof/cmodule.py
nouiz: The hash split the key in 2 part, the "soft" part and the "hard" part. I don't want to change that logic now. This would go in the refactoring of the cache.

I'm not to remove from the code the possibility to do "sha256:" in the futur. If we want to add the name, we could do "hash:NAME:" and it will always work as we only check for "hash:"

theano/gof/utils.py
nouiz: I don't want to mess with that in this PR. Too close to a release and I'm not confident enough about that change myself.

nouiz: I think that some combination of numpy/python don't like that we call directly hashlib.sha256(msg) on ndarray. The Python 3 version don't have that. So when we drop python 2.7, this will be removed. I don't want to chance which version needed that to know if it is safe to remove.

nouiz:
I updated the code

theano/gof/utils.py
abergeron: ok

Issue #6268, opened 2017-08-05
ZeroDivisionError in profile


ferrine:
I have weird bug in profiling. Is it a king of corner case? Should I care about it?
<details>

```
100%|##########| 300/300 [00:10<00:00, 29.21it/s]
Function profiling
==================
  Message: /Users/ferres/dev/pymc3/pymc3/variational/opvi.py:297
  Time in 300 calls to Function.__call__: 1.019666e+01s
  Time in Function.fn.__call__: 1.018261e+01s (99.862%)
  Time in thunks: 1.013567e+01s (99.402%)
  Total compile time: 3.605265e+00s
    Number of Apply nodes: 87
    Theano Optimizer time: 2.497554e+00s
       Theano validate time: 3.463984e-02s
    Theano Linker time (includes C, CUDA code generation/compiling): 1.033863e+00s
       Import time 2.375710e-01s
       Node make_thunk time 4.780002e-01s
           Node for{cpu,grad_of_scan_fn}(Shape_i{0}.0, Alloc.0, Alloc.0, Subtensor{int64:int64:int64}.0, Shape_i{0}.0) time 1.219413e-01s
           Node Subtensor{int64:int64:int64}(histogram, ScalarFromTensor.0, ScalarFromTensor.0, Constant{-1}) time 2.799010e-02s
[...]

PR #6273, opened 2017-08-07
[CRASH] Crash fix when we don't do scan(..., profile=True)

nouiz:
Fix gh-6268. 

but we do theano.function(..., profile=True)


ferrine:
Thanks

PR #6274, opened 2017-08-08
Allow to specify the target for mrg random.

nouiz:
WIP, what do you think of that user interface?

This would need test and be added to all random distribution.


abergeron:
I'm ok with adding target like this, but it would probably be better to have it as a keyword-only argument (through `**kwargs`) so that we don't need a notset hack.

nouiz:
finished in https://github.com/Theano/Theano/pull/6355

PR #6280, opened 2017-08-08
Move C files into c_code sub-folder in every module.

### LINKED ISSUE #6047: Move C codes into a sub-folder `c_code` for each module

notoraptor:
Closes #6047 . All `.c` and `.h` files should have been moved to sub-folder `c_code` in every module. Let's see if Jenkins and Travis accept these changes.

@abergeron @nouiz 

NB: As these are sensitive changes, I think it won't be merged for next beta, but I prepare the work.


abergeron:
I'm mostly ok with that change.  We just have to make sure that the C files are installed correctly for `pip install` and `python setup.py install` with distutils and setuptools.

notoraptor:
@abergeron @nouiz I have just uploaded Theano beta version 0.10.0b1 to PyPI, and current `c_code` sub-folders (in `theano.gpuarray` and `theano.tensor`) are correctly included ! Check here: https://pypi.python.org/pypi/Theano/0.10.0b1

To test it, you can run DimShuffle tests, as they are now COps with C files in `c_code`:
```
nosetests -xvs theano.gpuarray.tests.test_elemwise:test_GpuDimShuffle
nosetests -xvs theano.tensor.tests.test_elemwise:test_DimShuffle
```

So C files seems to be installed correctly.

abergeron:
Perfect then.  Do we want this for 0.10?

nouiz:
Yes! It seem ready, so why wait. merged

PR #6283, opened 2017-08-09
Squashed commit of NVIDIA changes for CUDA9/Volta

borisfom:
This may not be a PR per se - more like work in progress.

- I have implemented tensor_op setting for Volta which seems to work. This should be extended to try fp16 HMMA - in the next cycle.

A larger chunk of changes is about algo cache. I am happy with the refactoring part - it works. Actual caching is disabled as It still needs work, I am looking for your help:
1. Shape string generation needs to be fixed
2. Alignment part of the hash key should be added (from C code?)
3. device id needs to be added to the hash key, as well as fwd/bwd_filter/bwd_data selector.


theano/configdefaults.py
```diff
@@ -1580,7 +1580,7 @@ def filter_vm_lazy(val):
 
 
 def _timeout_default():
-    return theano.config.compile.wait * 24
+    return theano.config.compile.wait * 48
```
nouiz: why change this? The current number is 5*24, so 120s. What prompted you to change this?

theano/gpuarray/__init__.py
```diff
@@ -123,7 +123,8 @@ def init_dev(dev, name=None, preallocate=None):
               (name, dev, context.devname, pcibusid),
               file=sys.stderr)
     pygpu_activated = True
-
+    cuda_available = True
```
nouiz: We don't use that variable anymore. This was the name in the old back-end. User should start to use pygpu_activated instead.

If this is difficult for user, can you explain why? To have this change work well, it would request other change to this file.

theano/gpuarray/cudnn_defs.py
```diff
@@ -145,7 +151,7 @@ def get_definitions(cudnn_version=None):
     if None, return definitions for the  most recent supported cuDNN version.
 
     """
-    if cudnn_version is not None and cudnn_version // 1000 == 5:
-        return CuDNNV51()
+    if cudnn_version is not None and cudnn_version // 1000 == 6:
```
nouiz: You seem to keep cudnn v5 support. I'm fine to keep it or remove it now. If it is easy to keep, why not.

If we keep it, you would need to add this:

elif cudnn_version is not None and cudnn_version // 1000 == 6:
    return CuDNNV51()

theano/gpuarray/dnn.py
```diff
@@ -58,7 +58,7 @@
     pass
 
 # Update these names when new versions of cudnn are supported.
-WIN32_CUDNN_NAMES = ['cudnn64_6.dll', 'cudnn64_5.dll']
+WIN32_CUDNN_NAMES = ['cudnn64_7.dll', 'cudnn64_6.dll', 'cudnn64_5.dll']
```
nouiz: If we remove v5 support, remove the last element of this list.

theano/gpuarray/dnn.py
```diff
@@ -164,13 +164,13 @@ def _dnn_check_compile():
 
 def _dnn_check_version():
     v = version()
-    if v < 5000:
-        return False, "cuDNN version is too old. Update to v5* or higher, was %d." % v
-    if v >= 6100:
+    if v < 6000:
+        return False, "cuDNN version is too old. Update to v6* or higher, was %d." % v
+    if v >= 7200:
```
nouiz: why 7200? I would keep doing as before and use 7100. This way user will know we didn't tested that version of Theano with a new release.

theano/gpuarray/dnn_fwd.c
```diff
@@ -142,6 +143,9 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
           return 1;
         }
         algo = choice.algo;
+        params->conv_algo = algo;
+        params->conv_ws_size = choice.memory;
+        params->conv_tensor_op = choice.mathType;
```
nouiz: I don't think this is a good use of params. Can you just create new variable like algo (around line 86) instead for those 3 extra information?

Why did you needed to modify params?

theano/gpuarray/dnn_fwd.c
```diff
@@ -172,7 +176,18 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
     } else {
       algo = prev_algo;
     }
-
+    
+    params->conv_algo = algo;
+    // CUDNN7: need to set math type
+    err = cudnnSetConvolutionMathType(desc, params->conv_tensor_op);
```
nouiz: This need a "#ifdef CUDNN>=7" or similar I think. In this file, it is not like in libgpuarray, we include cudnn.h and I think this function don't exist in cudnn v6

theano/gpuarray/dnn_gw.c
```diff
@@ -154,8 +154,8 @@ APPLY_SPECIFIC(conv_gw)(PyGpuArrayObject *input, PyGpuArrayObject *output,
         return 1;
       }
 
-      // Guess 4Mb if the info is not available
-      if (free == 0) free = 4 * 1024 * 1024;
+      // Guess 1Gb if the info is not available
+      if (free == 0) free = 1 * 1024 * 1024 * 1024;
```
nouiz: Why this change? I think it is risky to use an high value here.

theano/gpuarray/subtensor.py
```diff
@@ -1108,15 +1107,19 @@ def gpu_kernels(self, node, nodename):
...
+      __half2float(old_perm));
     new_ = __byte_perm(old, sum, ((ga_size)addr & 2) ? 0x5410 : 0x3254);
     old = atomicCAS(base, assumed, new_);
   } while (assumed != old);
-  return (ga_half)__byte_perm(old, 0,
-                                  ((ga_size)addr & 2) ? 0x4432 : 0x4410);
+  ga_half ret;
+  __HALF_TO_US(ret) = __byte_perm(old, 0,
+        ((ga_size)addr & 2) ? 0x4432 : 0x4410);
+  return ret;
```
nouiz: @abergeron I let you review this part.

theano/scalar/basic.py
```diff
@@ -2658,6 +2658,13 @@ class Neg(UnaryScalarOp):
     nfunc_spec = ('negative', 1, 1)
 
     def impl(self, x):
+        # We have to make sure x is not a numpy.bool_, because
+        # `-numpy.bool_(True)` is `False` (we want 0), and
```
nouiz: numpy.bool_ is python bool. numpy.bool is a numpy scalar of type bool. 

So mostly, the problem is with python bool, not numpy bool.

It would be great to add a test for that. Do you still have the problematic case? If not, I'll try to make one.

theano/gpuarray/dnn_fwd.c
borisfom: This is because I want to access new values from the caller in Python.

theano/gpuarray/dnn_gw.c
borisfom: 4MB is very low for common nets - I have seen larger than 1G workspaces required.

theano/gpuarray/subtensor.py
borisfom: Yes this is previously done CUDA9 fix.

theano/scalar/basic.py
borisfom: Please do!

theano/gpuarray/dnn.py
borisfom: People were talking about 7.1 coming soon ... 7.1 is fine.

theano/gpuarray/cudnn_defs.py
borisfom: I meant to remove it - just did not see the point of modifying existing structs.

theano/gpuarray/__init__.py
borisfom: This is an artifact, please remove

theano/configdefaults.py
borisfom: I had a timeout once - we can revert it

requirement-rtd.txt
```diff
@@ -1,6 +1,12 @@
 sphinx>=1.3.0
 pygments
 nose>=1.3.0
+numpy
+gnumpy
+pydot
+pydot2
+Cython
```
borisfom: It sort of worked in our script - for Docker at least.

doc/index.txt
```diff
@@ -20,6 +20,9 @@ learning/machine learning <https://mila.umontreal.ca/en/cours/>`_ classes).
 
 News
 ====
+* 2017/03/07: Release of Theano 0.9.0rc3, with crash fixes, bug fixes and improvements.
+
+* 2017/03/20: Release of Theano 0.9.0. Everybody is encouraged to update.
```
borisfom: Artifact

clean.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
borisfom: Please remove - this is stuff I added for NV release

theano/gpuarray/dnn.py
nouiz: k

theano/gpuarray/subtensor.py
abergeron: I want to move those definitions to libgpuarray so that we don't need to provide them manually everywhere.

theano/gpuarray/dnn_fwd.c
abergeron: Params doesn't allow you to do that.  It's one-way only.

theano/gpuarray/dnn_gw.c
abergeron: This is fallback code in case we can't get the real amount of free memory.  We don't want to assume that there is a very large amount available and then fail to allocate it.

borisfom:
Will reopen on a new branch.

PR #6285, opened 2017-08-09
(small fix) use `os.remove` instead of system `rm` command

notoraptor:
Reported in #6284 .

@nouiz @abergeron


abergeron:
Travis barfed over itself, but I don't see that as a reason to not merge.

nouiz:
Another example of why we should not merge in the master without the CI passing. There was a bug in this PR that make all tests CI run fails. See the fix:

https://github.com/Theano/Theano/pull/6299

PR #6286, opened 2017-08-09
Implement Unshared Convolution

vikramnitin9:
Continues work from #6007. Related to #5620.

Old PR was rebased since the grouped convolution was merged first. I thought this would be neater to continue discussion, seeing as the old one was already rather messy.

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -1787,7 +1787,7 @@ class GpuCorr3dMM_gradInputs(BaseGpuCorr3dMM):
             broadcastable = [topgrad.type.broadcastable[0], False,
                              False, False, False]
         else:
-            broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[-3],
+            broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[-4],
                              False, False, False]
         return Apply(self, [kern, topgrad] + height_width_depth,
                      [GpuArrayType(dtype=topgrad.dtype,
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -924,6 +924,8 @@ class GpuCorrMM(BaseGpuCorrMM):
         divided into.
         should be an int
         set to 1 to disable grouped convolution
+    unshared
+        Perform unshared correlation (default: False)
 
     Notes
     -----
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -458,13 +458,15 @@ class BaseGpuCorrMM(CGpuKernelBase):
     num_groups :
         Divides the image, kernel and output tensors into num_groups
         separate groups. Each which carry out convolutions separately (default : 1).
+    unshared
+        Perform unshared correlation (default: False)
     """
     check_broadcast = False
-    __props__ = ('border_mode', 'subsample', 'filter_dilation', 'num_groups')
+    __props__ = ('border_mode', 'subsample', 'filter_dilation', 'num_groups', 'unshared')
     _f16_ok = True
 
     def __init__(self, border_mode="valid", subsample=(1, 1),
-                 filter_dilation=(1, 1), num_groups=1):
+                 filter_dilation=(1, 1), num_groups=1, unshared=False):
         if isinstance(border_mode, integer_types):
             border_mode = (border_mode, border_mode)
         if isinstance(border_mode, tuple):
@@ -487,6 +489,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
             raise ValueError("Number of groups should be greater than 0")
         self.num_groups = num_groups
         CGpuKernelBase.__init__(self, ['c_code/corr_gemm.c'])
+        self.unshared = unshared
 
     @property
     def pad(self):
@@ -495,12 +498,13 @@ class BaseGpuCorrMM(CGpuKernelBase):
         return (0, 0)
 
     def __str__(self):
-        return '%s{%s, %s, %s, %s}' % (
+        return '%s{%s, %s, %s, %s, %s}' % (
             self.__class__.__name__,
             self.border_mode,
             str(self.subsample),
             str(self.filter_dilation),
-            str(self.num_groups))
+            str(self.num_groups),
+            str(self.unshared))
 
     def __setstate__(self, d):
         self.__dict__.update(d)
@@ -533,7 +537,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
 
     def c_code_cache_version(self):
         # Raise this whenever modifying the C code (including the file).
-        return (10,)
+        return (11,)
 
     def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width=None):
         """
@@ -581,6 +585,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
         dH, dW = self.subsample
         dilH, dilW = self.filter_dilation
         numgroups = self.num_groups
+        unshared = int(self.unshared)
         if self.border_mode == "half":
             padH = padW = -1
         elif self.border_mode == "full":
@@ -633,19 +638,24 @@ class BaseGpuCorrMM(CGpuKernelBase):
     int padH = %(padH)s;
     int padW = %(padW)s;
     int numgroups = %(numgroups)s;
+    int unshared = %(unshared)s;
 
     PyGpuArrayObject * bottom = %(bottom)s;
     PyGpuArrayObject * weights = %(weights)s;
     PyGpuArrayObject * top = %(top)s;
     PyGpuArrayObject * out2 = NULL;
 
+    int wdim, odim;
+    wdim = unshared ? 6 : 4;
+    odim = 4; //Can be set to 6 later for unshared backprop wrt weights
+
     // Obtain or infer kernel width and height
     // (we need to know it early to be able to handle auto-padding)
     size_t kH, kW, dil_kH, dil_kW;
     if (direction != 1) {
         // weight is an input variable, we can just read its shape
-        kH = PyGpuArray_DIMS(weights)[2];
-        kW = PyGpuArray_DIMS(weights)[3];
+        kH = PyGpuArray_DIMS(weights)[wdim-2];
+        kW = PyGpuArray_DIMS(weights)[wdim-1];
     }
     else {
         if (%(height)s != -1) {
@@ -699,8 +709,10 @@ class BaseGpuCorrMM(CGpuKernelBase):
 
     // Infer output shape and type
     // The inferred shape can be negative.
-    long long out_dim[4];
-    size_t out_dim_size[4];
+    long long out_dim[6];
+    size_t out_dim_size[6];
+    out_dim[4] = out_dim[5] = 0; //Only used for unshared backprop wrt weights
+    out_dim_size[4] = out_dim_size[5] = 0; //Same
     int out_typecode;
     PyGpuContextObject *out_context;
     switch(direction) {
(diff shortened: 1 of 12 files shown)


vikramnitin9:
Any suggestions about what I can do to make the tests run within the time limit? Maybe I should shift the unshared tests to a separate class, like the Separable and Grouped tests.

gvtulder:
Yes, I think moving the tests to a separate class would be a good way to limit the number of tests you're running.

vikramnitin9:
I've fixed some small errors. From some rudimentary tests of my own, I think the code is all right, but the tests need to be rewritten to move unshared stuff into a separate class.

vikramnitin9:
I've moved the tests to a separate class.

There's just one thing I'm worried about. The methods I'm using to test `gradweight` and `gradinput` are not foolproof. For `gradweight` I'm summing over all the regions, and for `gradinput` I'm applying the same filter on all regions. Both are necessary but not sufficient conditions.

I know that `utt.verify_grad` will cross-check, so do I need to even do a manual test like these? I tried to follow the same pattern as the Grouped conv tests.

theano/tensor/nnet/corr.py
```diff
@@ -130,12 +133,13 @@ def pad(self):
...
     def __str__(self):
-        return '%s{%s, %s, %s, %s}' % (
+        return '%s{%s, %s, %s, %s %s}' % (
             self.__class__.__name__,
             self.border_mode,
             str(self.subsample),
             str(self.filter_dilation),
-            str(self.num_groups))
+            str(self.num_groups),
+            str(self.unshared))
```
gvtulder: Increment `c_code_cache_version` for this class.

theano/tensor/nnet/corr.py
```diff
@@ -464,21 +541,29 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
...
-                    "BaseCorrMM: Failed to allocate output of %%lld x %%lld x %%lld x %%lld",
-                    (long long)out_dim[0], (long long)out_dim[1], (long long)out_dim[2], (long long)out_dim[3]);
+            if (odim == 4) {
+                PyErr_Format(PyExc_RuntimeError,
+                        "BaseCorrMM: Failed to allocate output of %%lld x %%lld x %%lld x %%lld",
+                        (long long)out_dim[0], (long long)out_dim[1], (long long)out_dim[2], (long long)out_dim[3]);
+            }
+            if (odim == 6) {
+                PyErr_Format(PyExc_RuntimeError,
+                        "BaseCorrMM: Failed to allocate output of %%lld x %%lld x %%lld x %%lld",
```
gvtulder: This needs two extra `%%lld`.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1931,3 +1931,129 @@ def test_interface(self):
...
+        for imshp, kshp, mode, sub, groups in zip(self.img_shape, self.kern_shape, self.border_mode,
+                                                  self.subsample, self.num_groups):
+            img = np.random.random(imshp).astype(theano.config.floatX)
+            kern = np.random.random(kshp).astype(theano.config.floatX)
+
+            unshared_conv_op = self.conv2d(border_mode=mode, subsample=sub,
+                                           filter_dilation=self.filter_dilation,
+                                           num_groups=groups, unshared=True)
+            unshared_out_sym = unshared_conv_op(img_sym, kern_sym)
+            unshared_func = theano.function([img_sym, kern_sym], unshared_out_sym, mode=self.mode)
```
gvtulder: Here you need to check that you're using the right op:
```python
            assert any([isinstance(node.op, self.conv2d_op)
                       for node in unshared_func.maker.fgraph.toposort()])
```
and something similar for `test_gradweight` and `test_gradinput`.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1931,3 +1931,129 @@ def test_interface(self):
...
+    conv2d_op = theano.tensor.nnet.abstract_conv.AbstractConv2d
+    conv2d_gradw_op = theano.tensor.nnet.abstract_conv.AbstractConv2d_gradWeights
+    conv2d_gradi_op = theano.tensor.nnet.abstract_conv.AbstractConv2d_gradInputs
+
+    def setUp(self):
+        self.mode = theano.compile.mode.Mode(optimizer='None')
+
+        self.img_shape = [(2, 6, 5, 5), (1, 4, 7, 5), (1, 8, 5, 3), (1, 4, 7, 7)]
+        self.kern_shape = [(2, 3, 3, 6, 3, 3), (2, 6, 4, 4, 5, 3), (4, 2, 1, 2, 3, 3), (4, 4, 5, 2, 5, 3)]
+        self.topgrad_shape = [(2, 2, 3, 3), (1, 2, 6, 4), (1, 4, 2, 1), (1, 4, 4, 5)]
```
gvtulder: Can you make these any smaller? (Or remove a few?) The `verify_grad` on these large inputs is quite slow (see the [Travis log](https://travis-ci.org/Theano/Theano/jobs/264669314#L3204-L3206)), so reducing the size of the inputs would help.

Alternative: run `verify_grad` for a subset of test cases (for example, pass a list of booleans in your `zip` loop to say when you want to run it). The rest of the test isn't too slow.

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1931,3 +1931,129 @@ def test_interface(self):
...
+            single_kshp = kshp[:1] + kshp[3:]
+
+            ref_conv_op = self.conv2d_gradw(border_mode=mode, subsample=sub,
+                                            filter_dilation=self.filter_dilation,
+                                            num_groups=groups, unshared=False)
+            ref_out_sym = ref_conv_op(img_sym, top_sym, single_kshp[-2:])
+            ref_func = theano.function([img_sym, top_sym], ref_out_sym, mode=self.mode)
+            ref_output = ref_func(img, top)
+
+            utt.assert_allclose(unshared_output.sum(axis=(1, 2)), ref_output)
```
gvtulder: Summing is nice, but perhaps you can also do a loop like the one you do for `test_fwd`? Loop over every pixel in the output, set all other topgrad pixels to zero, compute and check the gradient for that unshared kernel?

theano/tensor/nnet/tests/test_abstract_conv.py
```diff
@@ -1931,3 +1931,129 @@ def test_interface(self):
...
+        kern_sym = tensor6('kern')
+        top_sym = theano.tensor.tensor4('top')
+        ref_kern_sym = theano.tensor.tensor4('ref_kern')
+
+        for imshp, kshp, topshp, mode, sub, groups in zip(self.img_shape, self.kern_shape, self.topgrad_shape,
+                                                          self.border_mode, self.subsample, self.num_groups):
+            single_kshp = kshp[:1] + kshp[3:]
+            single_kern = np.random.random(single_kshp).astype(theano.config.floatX)
+            kern = single_kern.reshape((kshp[:1] + (1, 1) + kshp[3:]))
+            kern = np.tile(kern, (1, kshp[1], kshp[2], 1, 1, 1))
```
gvtulder: Similar here: can you do a loop to compute the individual outputs? I have the impression that the implementation is quite fast and this wouldn't take too long.

theano/tensor/nnet/tests/test_abstract_conv.py
vikramnitin9: This works, thanks

theano/tensor/nnet/tests/test_abstract_conv.py
vikramnitin9: I'm not sure exactly how to do this. Would I set all but a few pixels in the outgrad to zero?

vikramnitin9:
There's an error on the GPU tests that happens some of the time (I tried passing [True, True, True, True] as the verify_flags also, just in case that was the source of the randomness).
```
ValueError: GpuCorrMM shape inconsistency:
  bottom shape: 1 2 4 2
  weight shape: 1 4 2 2 4 2
  top shape: 1 2 4 2 (expected 1 1 4 2)
```
I'll keep trying to fix this.

vikramnitin9:
The above error occurs only when the batch size is 1. When it is 1, the second dimension (num_filters) also becomes 1 somehow. And it occasionally disappears and the tests run fine.

Apart from the above error, I also get the following error while running `test_corr.TestUnsharedCorr2d`:
`RuntimeError: Unexpected error in an Op's C code. No Python exception was set.`

The normal `test_abstract_conv.TestUnsharedConv` tests work fine.

I've been trying to figure out what's going on for the last two days without success.

theano/tensor/nnet/tests/test_abstract_conv.py
gvtulder: Yes, set all but a single pixel in the outgrad to zero, then convolve this single pixel with the output kernel for that position to compute the gradient for the input. Repeat for all pixels in the outgrad and sum to get the total input gradient.

Like this: https://gist.github.com/gvtulder/16b250549f20ccc0a51efba80560fe4b

theano/tensor/nnet/corr.py
```diff
@@ -370,75 +381,136 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
...
+                             "  weights shape: %%ld x %%ld x %%ld x %%ld x %%ld x %%ld\\n"
+                             "  top shape: %%ld x %%ld x %%ld x %%ld\\n",
+                             (long int)PyArray_DIMS(bottom)[0], (long int)PyArray_DIMS(bottom)[1],
+                             (long int)PyArray_DIMS(bottom)[2], (long int)PyArray_DIMS(bottom)[3],
+                             (long int)out_dim[0], (long int)out_dim[1], (long int)out_dim[2],
+                             (long int)out_dim[3], (long int)out_dim[4], (long int)out_dim[5],
+                             (long int)PyArray_DIMS(top)[0], (long int)PyArray_DIMS(top)[1],
+                             (long int)PyArray_DIMS(top)[2], (long int)PyArray_DIMS(top)[3]);
+            }
             %(fail)s
```
gvtulder: This is the cause of your `RuntimeError: Unexpected error in an Op's C code. No Python exception was set.` error.

I didn't know this, but apparently `%(fail)s` should always go together with setting a Python exception. In this case, you're calling the PyErr_Format only if the shape isn't correct, but you're always calling `%(fail)s`, even if you didn't set an error because everything is fine. If you move the `%(fail)s` inside the if it should work.

vikramnitin9: Thank you so much! That error had been bothering me for a while now!

theano/tensor/nnet/tests/test_abstract_conv.py
vikramnitin9: Done, thanks

gvtulder:
> There's an error on the GPU tests that happens some of the time (I tried passing [True, True, True, True] as the verify_flags also, just in case that was the source of the randomness).
>
> ```
> ValueError: GpuCorrMM shape inconsistency:
>   bottom shape: 1 2 4 2
>   weight shape: 1 4 2 2 4 2
>   top shape: 1 2 4 2 (expected 1 1 4 2)
> ```

I think this problem is caused by the batch size in those tests being 1. If I increase it to anything higher than 1 it works for me. Try increasing it to 9, for instance: that's not a number that appears anywhere else, so it's not that you've accidentally swapped some dimensions, but it does make the problem go away.

This is not a problem in your code. I've tried to find out why this happens, but it may have something to do with the way the broadcastable patterns are converted between GPU and CPU. If I remember correctly, the GPU tensors don't have a broadcastable pattern, so when an array is converted from GPU to CPU, any dimensions with length 1 are automatically made broadcastable. It looks like this causes some problems in the `verify_grad` function in this case.

vikramnitin9:
> The above error occurs only when the batch size is 1. When it is 1, the second dimension (num_filters) also becomes 1 somehow. And the error occasionally disappears and the tests run fine.

Exactly! My initial thinking was that it might be an optimiser error, ie, an optimiser choosing an alternative implementation if the batch size was 1. But I couldn't find any such instance.

Why only for unshared though? Shouldn't it always happen, if the broadcasting is a problem?

gvtulder:
It looks like most of the other tests don't test with a batch size of 1, or only do that on the CPU.

It's also possible that this problem occurs only because you call the GPU ops directly, instead of relying on an optimization to move it to the GPU. In that case, perhaps a large part of the `verify_grad` logic is done with CPU tensors where this is not a problem.

Related to that, I'm wondering if this does what we want it to do:
https://github.com/vikramnitin9/Theano/blob/4b0f48fcffff71174c288a7bd720380987e80627/theano/gpuarray/tests/test_gemmcorr.py#L275-L280
I've seen it's like this for the grouped convolution tests, but is this really what we want? If you directly call the GPU ops, you're not testing whether the optimization replaces the abstract op with the GPU version. In that case checks like these don't really make sense, because they'll always pass:
https://github.com/vikramnitin9/Theano/blob/4b0f48fcffff71174c288a7bd720380987e80627/theano/tensor/nnet/tests/test_abstract_conv.py#L1766-L1767
I would expect that you'd want to call the `conv2d = AbstractConv2d` op and then check if it has been replaced by `conv2d_op = GpuCorrMM` after compilation. Am I missing something?

vikramnitin9:
Wow, this fixed it! Thanks!

I'm guessing I should modify the same for grouped tests as well?

vikramnitin9:
Travis is not clearing the compile cache. So do I increment the c_cache_code_version again to force it to do so? I have already incremented it from 7 to 8 in the PR.

gvtulder:
I think it's fine to increment it again.

gvtulder:
> I'm guessing I should modify the same for grouped tests as well?

In a separate PR, perhaps. But maybe it's good to check first if this indeed the best way to do it. There may be a good reason to do it the way it's done now. I've opened a new issue to ask.

vikramnitin9:
Anything else I need to do? Something related to the MetaOptimizer maybe?

gvtulder:
I think it's nearly there. Can you run this on Jenkins, please?

I'm currently running some of the GPU tests locally. The GpuCorrMM tests pass, so that's good. I do get some optimization errors for 3D, which is weird because that shouldn't be affected. (Something to do with broadcasting patterns.)

theano/gpuarray/opt.py
```diff
@@ -1639,7 +1646,8 @@ def local_abstractconv_gemm(node):
             if prod1 > prod2:
                 rval = GpuCorrMM_gradWeights(border_mode,
                                              subsample,
-                                             filter_dilation)(
+                                             filter_dilation,
+                                             unshared)(
```
vikramnitin9: Actually, `unshared` is redundant because the outer thing has a check for `not unshared`. I've removed it.

vikramnitin9:
Can you tell me how to reproduce the 3D errors? `TestCorr3dMM` passes for me.

theano/gpuarray/blas.py
```diff
@@ -1682,7 +1787,7 @@ def make_node(self, kern, topgrad, shape=None):
             broadcastable = [topgrad.type.broadcastable[0], False,
                              False, False, False]
         else:
-            broadcastable = [topgrad.type.broadcastable[0], kern.type.broadcastable[1],
```
gvtulder: This should be `kern.type.broadcastable[-4]` (breaks a test otherwise).

gvtulder: (The failing test is `theano.gpuarray.tests.test_abstractconv.TestCorrMMConv3d.test_all`.)

vikramnitin9: Thank you.

nouiz:
jenkins test this

theano/gpuarray/blas.py
```diff
@@ -859,18 +943,22 @@ class GpuCorrMM(BaseGpuCorrMM):
     """
     def __init__(self, border_mode="valid",
                  subsample=(1, 1),
-                 filter_dilation=(1, 1), num_groups=1):
+                 filter_dilation=(1, 1), num_groups=1, unshared=False):
```
gvtulder: This new parameter needs to be added to the documentation.

nouiz:
jenkins test this

theano/gpuarray/blas.py
```diff
@@ -495,12 +498,13 @@ def pad(self):
...
+        return '%s{%s, %s, %s, %s, %s}' % (
             self.__class__.__name__,
             self.border_mode,
             str(self.subsample),
             str(self.filter_dilation),
-            str(self.num_groups))
+            str(self.num_groups),
+            str(self.unshared))
 
     def __setstate__(self, d):
```
nouiz: Does the method flops() need an update to take into accound the new parameter?

theano/tensor/nnet/__init__.py
```diff
@@ -105,6 +109,11 @@ def conv2d(input, filters, input_shape=None, filter_shape=None,
         Divides the image, kernel and output tensors into num_groups
         separate groups. Each which carry out convolutions separately
 
+    unshared: bool
+        If true, then unshared or 'locally connected' convolution will be
+        performed. A different kernel will be used for each region of the
```
nouiz: kernel -> filter

theano/tensor/nnet/abstract_conv.py
```diff
@@ -44,9 +44,13 @@ def get_conv_output_shape(image_shape, kernel_shape,
...
-        kernel shape. Its four (or five) elements must correspond respectively
-        to: number of output channels, number of input channels, height and
-        width (and possibly depth) of the kernel. None where undefined.
+        kernel shape. For a normal convolution, its four (or five) elements
+        must correspond respectively to : number of output channels, number of
+        input channels, height and width (and possibly depth) of the kernel.
+        For an unshared convolution, its six channels must correspond to :
+        number of output channels, height and width
+        of the output, number of input channels, height and width of the kernel.
+        None where undefined.
```
nouiz: So this mean unshared convolution is only supported for 2d convolution. Can you make that more clear? What about this:

For a normal convolution, its four (2d convolution) or five (3d convolution) ...
For an unshared 2d convolution, ...

theano/tensor/nnet/abstract_conv.py
```diff
@@ -888,6 +911,10 @@ def conv2d_grad_wrt_inputs(output_grad,
         for dim in [0, 1, 2, 3]:
             assert isinstance(filter_shape[dim], (theano.tensor.TensorConstant,
                                                   integer_types, type(None)))
+        if unshared:
+            for dim in [4, 5]:
+                assert isinstance(filter_shape[dim], (theano.tensor.TensorConstant,
+                                                      integer_types, type(None)))
```
nouiz: what about this small refactorisation that will detect more problems.
```
if filter_shape is not None:
    expected = 4
    if unshared:
        expected = 6
    assert len(dim) == expected
    for dim in range(expected):
        assert ...
```

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1663,18 +1716,27 @@ def conv(self, img, kern, mode="valid", dilation=1, num_groups=1):
...
                 'invalid dilation {}, expected {} values'.format(dilation,
                                                                  self.convdim))
+        if unshared and direction == "backprop weights":
+            if mode != "valid":
+                raise ValueError('conv mode for unshared backprop wrt weights must be "valid"')
+            # Do a transpose later to bring it to required shape
+            out_shape = (img.shape[0], kern.shape[0],
+                         kern.shape[2], kern.shape[3],
+                         img.shape[2] - kern.shape[2] + 1,
+                         img.shape[3] - kern.shape[3] + 1)
```
nouiz: why not have this logic inside the get_conv_output_shape while adding it an extra parameter unshared? This would allow to reuse it elsewhere.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1699,11 +1761,19 @@ def conv(self, img, kern, mode="valid", dilation=1, num_groups=1):
...
+                                                                                                  out_shape[2:], direction)
+                                else:
+                                    # some cast generates a warning here
+                                    out[b, g * output_channel_offset + n, ...] += _convolve2d(img[b, g * input_channel_offset + im0, ...],
+                                                                                              dilated_kern[g * output_channel_offset + n, im0, ...],
+                                                                                              1, val, bval, 0)
+
         elif self.convdim == 3:
+            if unshared:
+                raise NotImplementedError('Unshared 3D convolution is not implemented')
```
nouiz: This should be asserted in the make_node to give an error when we build the graph instead of just when we execute the graph.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1849,14 +1992,16 @@ def __init__(self,
                  subsample=(1, 1),
                  filter_flip=True,
                  filter_dilation=(1, 1),
-                 num_groups=1):
+                 num_groups=1,
+                 unshared=False):
```
nouiz: It seem it is supported to combine num_groups and unshared. Make that clear in the doc.

theano/gpuarray/blas.py
vikramnitin9: The number of flops remains the same since we're just multiplying different weights. I had initially raised a NotImplemented error before @gvtulder pointed this out.

theano/tensor/nnet/abstract_conv.py
vikramnitin9: The thing is, this is not the actual output shape of gradWeights. This is `(batch_size, num_filters, out_rows, out_cols, kH, kW)`, but the actual output shape is `(batch_size, out_rows, out_cols, num_filters, kH, kW)`. A transpose is done in the perform function to bring it into the required shape.
I could call `get_conv_gradweights_shape` and then shuffle around the values. Do you think I should do this?

nouiz: I think it is interreting to have all get_conv_*shape() function work with unshared. So I would have a tendency to make this. The code will be more consistent and have will have only 1 way to compute the output shape for all type of convolution.

theano/tensor/nnet/abstract_conv.py
vikramnitin9: I've put a check [here](https://github.com/Theano/Theano/pull/6286/files#diff-683982cb7c0d94726ca578a15a769a3cR1671). Do I need to put it in make_node also?

theano/tensor/nnet/abstract_conv.py
vikramnitin9: Okay, I've changed that.

vikramnitin9: There are some problems while doing that. The first two dimensions of img and topgrad are flipped before being passed to `conv`. So I can't use `get_conv_gradweights_shape`. And I can't use `get_conv_output_shape` since this is a special case (4d img and 4d kern -> 6d output).

Actually, `get_conv_output_shape` doesn't need an unshared parameter since the shape doesn't change. So it is complete as is.

theano/tensor/nnet/abstract_conv.py
vikramnitin9: @nouiz - Does `BaseAbstractConv.__init__` get called while building the graph? If so there is a check there. I'll certainly put another in `make_node` if you want, but I thought I'd check with you first.

vikramnitin9:
Also, I have added documentation for `num_groups` in `BaseAbstractConv` and `CorrMM`. I hope that's okay.

theano/gpuarray/opt.py
```diff
@@ -2069,7 +2085,7 @@ def provide_inputs(self, node, inputs):
                                          node.op.subsample,
                                          node.op.filter_dilation)
 
-            result[kshape] = theano.tensor.as_tensor_variable(node.op.kshp[2:])
+            result[kshape] = theano.tensor.as_tensor_variable(node.op.kshp[-2:])
```
affanv14: Hey @vikramnitin9, just a small suggestion. This will probably cause an error for 3d convolutions.

theano/tensor/nnet/abstract_conv.py
nouiz: Ok to keep it like this if you think it is better. For get_conv_output_shape(), if the output shape is the same for unshared or not, can you add a note in its docstring about that? Otherwise people will think it is just not implemented instead of not needed.

@gvtulder I let you confirm that this is true. I didn't follow all what is related to the shapes computation.

theano/tensor/nnet/abstract_conv.py
nouiz: If it is tested in `BaseAbstractConv.__init__`, then this is good.

nouiz:
jenkins test this

theano/gpuarray/opt.py
gvtulder: @affanv14 Do you know if this is tested anywhere? It doesn't seem to crash the `theano.gpuarray.tests.test_opt.Conv_opt_test`.

theano/tensor/nnet/abstract_conv.py
gvtulder: @nouiz Yes, that's correct. `get_conv_output_shape()`, `get_conv_gradinputs_shape()` and `get_conv_gradweights_shape()` methods all support unshared convolution. Only `get_conv_gradweights_shape()` needs an additional `unshared` parameter for this, because the type of convolution is irrelevant for the other two. The shape of those is only influenced by the kernel width and height, but it's the same for shared and unshared convolutions.

The shape that's computed here is very much a special case, which is only relevant when it's combined with the other trickery below. It's not something that you would want to compute in practice, so I think computing it directly is fine.

theano/gpuarray/opt.py
affanv14: `theano.gpuarray.tests.test_opt.Conv_opt_test`  tests the new optimisers I wrote, it does not test the default optimisers which are already there. The alternative optimisers for gradweight pass do not require the kshape argument, that is why these tests pass. The default gradweights optimisers ie `local_abstractconv_gw_cudnn` and `local_abstractconv3d_gradweights_gemm` will cause an error since they require the kshape argument

affanv14: Also the meta-optimiser was written in such a way that if one optimiser causes an error, it will select another one

affanv14: but I think it would be a good idea to test the default optimisers as well using the meta-optimizer to prevent it from happening again.
@nouiz @gvtulder what do you think?

theano/tensor/nnet/abstract_conv.py
nouiz: @vikramnitin9 so just add a note in the docstring of get_conv_output_shape() and get_conv_gradinputs_shape() to tell that the output shape is independent if the computation is for unshared or shared convolution.

theano/gpuarray/opt.py
nouiz: I don't understand. What is not currently tested?

I don't understand what would be covered by testing the meta optimizer with the default optimizers. Do you mean the "default optimizer not in meta optimizer"? Or do you mean the default optimizer inside the meta optimizer?

If the second, then I think this is good to add if they aren't too long.

gvtulder: I didn't really follow the metaoptimizer work, so I don't know how difficult it would be to test this. But if the metaopimizer can select those optimizers, then I think it would be good to have a test. (Although probably not as part of this PR.)

affanv14: @nouiz , yes I meant the second one.

@gvtulder yes, this is really easy, I'll have one up soon

nouiz:
I have done a high level review and I'm good with it.

@gvtulder I'll let you merge it when both CI pass if you are also good with this PR.

jenkins test this.

thanks to both of you.

vikramnitin9:
Jenkins failed because it tried to apply `GpuDnnConvGradW` and `GpuDnnConvGradI`. I've added .exluding('cudnn') now, so hopefully it should pass.
But I'm puzzled how the dnn optimiser wasn't disabled. I've added checks in dnn.py. Is there anywhere else that I'm missing?
I can't test this on my computer because I'm using OpenCL.
Travis failed with the following message.
```
File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_gc.py", line 125, in test_merge_opt_runtime
    assert dt < 5.0, dt
AssertionError: 8.634841442108154
```
Again, I'm unable to reproduce this on my computer. Any idea what it is?

vikramnitin9:
Also, an earlier Jenkins [build](http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/691/) for commit baef882 passed fine, so I don't get what has changed in the meantime.

vikramnitin9:
And [this](https://travis-ci.org/Theano/Theano/builds/266994470) Travis build also passed, a couple of commits ago.

gvtulder:
The Travis failure with "AssertionError: 8.634841442108154" is probably just a random glitch. It's from this test that times something completely unrelated to your changes: https://github.com/Theano/Theano/blob/a5588fce364c3d6d067affc645dd9f9f31103f6e/theano/tensor/tests/test_gc.py#L124-L125 It must have been a temporary Travis slowdown that caused this bit to take just a few seconds too long. This test will probably pass in the next Travis run.

gvtulder:
Can you remove the last commit with `.exluding('cudnn')`? That's just hiding the problem. The optimizations shouldn't move unshared convolution to the cudnn ops, so there might be a bug somewhere.

I'm trying to pin down what causes this problem. It doesn't seem to occur when I merge your PR with the current master. However, merging baef882 into 7c5a427165f51dbfc6271338985135cff2a27ec5 gives me these GpuDnnConvGradI errors. Perhaps something has changed since then.

vikramnitin9:
That means that something changed between 308a9f1 and a5588fc that fixed the problem, right?

gvtulder:
It seems to be related to the `num_groups` parameter. I think this might have been the bug in the `num_groups` optimization that affanv14 fixed yesterday. The tests pass with the current master.

There's still a little problem in when your GPU tests are run in DebugMode. That's another of those broadcastable issues again, so I think it's best to just avoid 1s in your patterns. These shapes do pass the DebugMode tests:
```
--- a/theano/tensor/nnet/tests/test_abstract_conv.py
+++ b/theano/tensor/nnet/tests/test_abstract_conv.py
@@ -1757,9 +1757,9 @@ class TestUnsharedConv(unittest.TestCase):
     mode = theano.compile.mode.Mode(optimizer='None')
 
     def setUp(self):
-        self.img_shape = [(2, 1, 4, 4), (1, 2, 4, 2), (1, 3, 5, 3), (1, 4, 4, 4)]
-        self.kern_shape = [(2, 2, 2, 1, 3, 3), (2, 4, 2, 2, 4, 2), (3, 2, 1, 1, 3, 3), (4, 3, 3, 2, 4, 2)]
-        self.topgrad_shape = [(2, 2, 2, 2), (1, 2, 4, 2), (1, 3, 2, 1), (1, 4, 3, 3)]
+        self.img_shape = [(2, 2, 4, 4), (3, 2, 4, 2), (3, 3, 5, 3), (3, 4, 4, 4)]
+        self.kern_shape = [(2, 2, 2, 2, 3, 3), (2, 4, 2, 2, 4, 2), (3, 2, 1, 1, 3, 3), (4, 3, 3, 2, 4, 2)]
+        self.topgrad_shape = [(2, 2, 2, 2), (3, 2, 4, 2), (3, 3, 2, 1), (3, 4, 3, 3)]
         self.border_mode = ['valid', 'full', 'valid', 'full']
         self.subsample = [(1, 1), (2, 2), (2, 1), (3, 2)]
         self.filter_dilation = (1, 1)
```

vikramnitin9:
Thanks for spotting the DebugMode problem. I've changed the shapes to the ones you suggested. I wish there was a better solution to this though.

gvtulder:
@nouiz I'm satisfied with this PR if the tests pass (but they should by now). I'm not allowed to merge anything, so I'll leave that to you.

Can you run Jenkins again?

nouiz:
jenkins test this

affanv14:
@vikramnitin9 since a few more optimizers got merged for the meta-optimizer https://github.com/Theano/Theano/pull/6336, could you include unshared in [this](https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1658) optimizer as well?

vikramnitin9:
@nouiz - do we need to run Jenkins again?

gvtulder:
The change looks good to me.

I do think it would be good to have a test for this, but since these meta-optimizations also aren't tested for grouped convolution (as far as I can see) we should maybe open a new issue for that.

nouiz:
jenkins test this

gvtulder:
Jenkins passed. Good to go!

Issue #6287, opened 2017-08-09
[BUG, CRASH] C code compilation for code that was working


nouiz:
This issue is for what is reported in [[theano-users] Error while compiling two theano functions](https://groups.google.com/d/topic/theano-users/2D1zj_cs-kQ/discussion)

To reproduce:
- install https://github.com/kundan2510/pixelCNN
- run: python cifar10_gen.py

I'll try to find a Theano version where this work

Current work around: THEANO_FLAGS=optimizer_excluding=fusion

nouiz:
It also err on Theano 0.9 on CPU and in the old and new GPU back-end.

PR #6289, opened 2017-08-10
[BUG] Fix gh-6287 and a regression introduced in gh-6218

nouiz:
fix https://github.com/Theano/Theano/issues/6287

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -6663,8 +6663,8 @@ class AllocEmpty(gof.Op):
         output.tag.nan_guard_mode_check = False
         return Apply(self, shape, [output])
 
-    def debug_perform(self, node, inputs, out_):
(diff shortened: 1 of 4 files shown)


theano/tensor/basic.py
```diff
@@ -6663,8 +6663,8 @@ def make_node(self, *shape):
         output.tag.nan_guard_mode_check = False
         return Apply(self, shape, [output])
 
-    def debug_perform(self, node, inputs, out_):
-        self.perform(node, inputs, out_)
+    def debug_perform(self, node, inputs, out_, params):
```
notoraptor: There seems to be also a `debug_perform()` in a child class (`theano.gpuarray.basic_ops.GpuAllocEmpty#debug_perform`). I think it should be fixed too.

nouiz: done

PR #6292, opened 2017-08-10
Fix CTC tests in FAST_COMPILE

nouiz:


notoraptor:
A rebase is needed here. After that, it will be ok for me if both CI tests pass!

nouiz:
rebased

PR #6295, opened 2017-08-10
Update conv doc.

nouiz:
fix https://github.com/Theano/Theano/issues/5047

this also include a fix to the optimization tag.

### DIFF
--- a/doc/library/tensor/nnet/conv.txt
+++ b/doc/library/tensor/nnet/conv.txt
@@ -22,78 +22,62 @@
 .. moduleauthor:: LISA
 
 
+The recomanded user interface are:
+
+- :func:`theano.tensor.nnet.conv2d` for 2d convolution
+- :func:`theano.tensor.nnet.conv3d` for 3d convolution
+
+With those new interface, Theano will automatically use the fastest
+implementation in many cases. On the CPU, the implementation is a GEMM
+based one. On the GPU, there is a GEMM based and :ref:`cuDNN
+<libdoc_gpuarray_dnn>` version.
+
+By default on the GPU, if cuDNN is available, it will be used,
+otherwise we will fall back to using gemm based version (slower than
+cuDNN in most cases and uses more memory). To get an error if cuDNN
+can not be used, you can supply the Theano flag ``dnn.enable=True``.
+
+Either cuDNN and the gemm version can be disabled using the Theano flags
+``optimizer_excluding=conv_dnn`` and ``optimizer_excluding=conv_gemm``,
+respectively. If both are disabled, it will raise an error.
+
+
+For the cuDNN version, there is different algorythm with different
+memory/speed trade-off. Manual selection of the right one is very
+difficult as it depend of the shapes and hardware. So it can change
+for each layer. An auto-tuning mode exist and can be activated by
+those flag: ``dnn.conv.algo_fwd=time_once``,
+``dnn.conv.algo_bwd_data=time_once`` and
+``dnn.conv.algo_bwd_filter=time_once``.
+
+This auto-tuning have the inconvenience that the first call is much
+slower as it try and time each implementation it have. So if you
+benchmark, it is important that you remove the first call from your
+timing.
+
+
 .. note::
 
-    As of December 2015, a new conv2d interface has been introduced.
-    :func:`nnet.conv2d <theano.tensor.nnet.conv2d>` defines an
-    abstract theano graph convolution operation
-    (:func:`nnet.abstract_conv.AbstractConv2d <theano.tensor.nnet.abstract_conv.AbstractConv2d>`)
-    that will be replaced by an actual convolution implementation during
-    the optimization phase.
-
-    As of October 2016 (version 0.9.0dev3), there is also a conv3d interface that provides
-    a similar operation for 3D convolution. :func:`nnet.conv3d <theano.tensor.nnet.conv3d>`
-    defines the abstract theano graph convolution operation
-    :func:`nnet.abstract_conv.AbstractConv3d <theano.tensor.nnet.abstract_conv.AbstractConv3d>`.
-
-    Since the abstract Op does not have any implementation, it will prevent
-    computations in the un-optimized graph, and cause problems with DebugMode,
-    test values, and when compiling with optimizer=None.
-
-    By default, if :ref:`cuDNN <libdoc_gpuarray_dnn>`
-    is available, we will use it, otherwise we will fall back to using the
-    gemm version (slower than cuDNN in most cases and uses more memory).
-
-    Either cuDNN and the gemm version can be disabled using the Theano flags
-    ``optimizer_excluding=conv_dnn`` and ``optimizer_excluding=conv_gemm``,
-    respectively. In this case, we will fall back to using the legacy
-    convolution code, which is slower, but does not require extra memory.
-    To verify that cuDNN is used, you can supply the Theano flag
-    ``optimizer_including=cudnn``. This will raise an error if cuDNN is
-    unavailable.
-
-    It is not advised to ever disable cuDNN, as this is usually the fastest
-    option. Disabling the gemm version is only useful if cuDNN is unavailable
-    and you run out of GPU memory.
-
-    There are two other implementations of 2D convolution: An FFT-based
-    convolution integrated into Theano, and an implementation by Alex Krizhevsky
-    available via Pylearn2. See the documentation below on how to use them.
-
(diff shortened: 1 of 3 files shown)


doc/library/tensor/nnet/conv.txt
```diff
@@ -22,78 +22,62 @@
...
-    To verify that cuDNN is used, you can supply the Theano flag
-    ``optimizer_including=cudnn``. This will raise an error if cuDNN is
-    unavailable.
+For the cuDNN version, there is different algorythm with different
+memory/speed trade-off. Manual selection of the right one is very
+difficult as it depend of the shapes and hardware. So it can change
+for each layer. An auto-tuning mode exist and can be activated by
+those flag: ``dnn.conv.algo_fwd=time_once``,
+``dnn.conv.algo_bwd_data=time_once`` and
+``dnn.conv.algo_bwd_filter=time_once``.
```
notoraptor: For theses flages, should we recommand `time_once`, or `time_on_shape_change` instead (even if it is currently not perfect) ?

In fact, `time_once` is good if shapes don't change. But it may cause errors if shapes change and the new shape is not supported by the algorithm selected in first call.

doc/library/tensor/nnet/conv.txt
```diff
@@ -22,78 +22,62 @@
...
-    respectively. In this case, we will fall back to using the legacy
-    convolution code, which is slower, but does not require extra memory.
-    To verify that cuDNN is used, you can supply the Theano flag
-    ``optimizer_including=cudnn``. This will raise an error if cuDNN is
-    unavailable.
+For the cuDNN version, there is different algorythm with different
+memory/speed trade-off. Manual selection of the right one is very
+difficult as it depend of the shapes and hardware. So it can change
+for each layer. An auto-tuning mode exist and can be activated by
+those flag: ``dnn.conv.algo_fwd=time_once``,
```
notoraptor: ->
For the cuDNN version, there ``are`` different `algorithms` with different
memory/speed trade-`offs`. Manual selection of the right one is very
difficult as it `depends on` the shapes and hardware. So it can change
for each layer. An auto-tuning mode `exists` and can be activated by
those `flags`:

doc/library/tensor/nnet/conv.txt
```diff
@@ -22,78 +22,62 @@
...
+``dnn.conv.algo_bwd_data=time_once`` and
+``dnn.conv.algo_bwd_filter=time_once``.
 
-    It is not advised to ever disable cuDNN, as this is usually the fastest
-    option. Disabling the gemm version is only useful if cuDNN is unavailable
-    and you run out of GPU memory.
+This auto-tuning have the inconvenience that the first call is much
+slower as it try and time each implementation it have. So if you
```
notoraptor: This auto-tuning `has` the inconvenience that the first call is much
slower as it `tries and times` each implementation it `has`.

doc/library/tensor/nnet/conv.txt
```diff
@@ -22,78 +22,62 @@
...
+benchmark, it is important that you remove the first call from your
+timing.
```
notoraptor: Should we indicate how to do that here (wich `profile=True,profiling.ignore_first_call=True`) ?

doc/library/tensor/nnet/conv.txt
```diff
@@ -22,78 +22,62 @@
...
 
-    There are two other implementations of 2D convolution: An FFT-based
-    convolution integrated into Theano, and an implementation by Alex Krizhevsky
-    available via Pylearn2. See the documentation below on how to use them.
 
-    Old conv2d interface is still accessible through :func:`nnet.conv.conv2d <theano.tensor.nnet.conv.conv2d>`.
+.. note::
 
+    Theano had older user interface like
+    theano.tensor.nnet.conv.conv2d. Do not user them anymore. They
```
notoraptor: -> Do not `use` them anymore

PS: Are this old interface deprecated ? (if not, maybe should we deprecate it).

doc/library/tensor/nnet/conv.txt
```diff
@@ -22,78 +22,62 @@
...
+    will give you slower code and won't allow easy switch between CPU
+    and GPU computation. They also support less type of convolution.
 
 
-TODO: Give examples on how to use these things! They are pretty complicated.
+Implementation Details
+======================
 
+This section give more implementation detail. Most of the time you do
```
notoraptor: -> This section `gives`

doc/library/tensor/nnet/conv.txt
nouiz: It is deprecated, but there isn't any warning. I'll add one.

doc/library/tensor/nnet/conv.txt
nouiz: I was think about user doing there global profile with time.time(). Not Theano profiler. I don't know how to make this more clear. Theano profiler will print a warning if the user don't add that flag. So I think we are fine.

But if you have an idea to make that better tell me. There was discussion to make a how to profile page, when done, it could be linked to from here.

doc/library/tensor/nnet/conv.txt
nouiz: done. This is a way to make the comment very clear. Good idea

doc/library/tensor/nnet/conv.txt
nouiz: done

doc/library/tensor/nnet/conv.txt
nouiz: done

doc/library/tensor/nnet/conv.txt
nouiz: done

doc/library/tensor/nnet/conv.txt
notoraptor: Ok ! I have no better idea, but indeed, theano profiler prints a warning in such cases, so it is ok.

doc/library/tensor/nnet/conv.txt
```diff
@@ -102,68 +87,37 @@ TODO: Give examples on how to use these things! They are pretty complicated.
...
     - :func:`Corr3dMM <theano.tensor.nnet.corr3d.Corr3dMM>`
       This is a CPU-only 3d correlation implementation based on
       the 2d version (:func:`CorrMM <theano.tensor.nnet.corr.CorrMM>`).
       It does not flip the kernel. As it provides a gradient, you can use it as a
       replacement for nnet.conv3d. For convolutions done on CPU,
-      nnet.conv3d will be replaced by Corr3dMM. To explicitly disable it, set
-      ``THEANO_FLAGS=optimizer_excluding=conv_gemm`` in your environment.
+      nnet.conv3d will be replaced by Corr3dMM.
 
     - :func:`dnn_conv <theano.gpuarray.dnn.dnn_conv>` GPU-only
```
notoraptor: At last suggestion: I think we should reference dnn_conv3d here (as we are in conv3d section):
```
    - :func:`dnn_conv3d <theano.gpuarray.dnn.dnn_conv3d>` GPU-only
```

nouiz: done

notoraptor: I was thinking more about something like this:
```patch
Index: doc/library/tensor/nnet/conv.txt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- doc/library/tensor/nnet/conv.txt	(revision 853d04a6771b345c298b976cd0c880bac8dc6194)
+++ doc/library/tensor/nnet/conv.txt	(revision )
@@ -109,13 +109,11 @@
       replacement for nnet.conv3d. For convolutions done on CPU,
       nnet.conv3d will be replaced by Corr3dMM.
 
-    - :func:`dnn_conv <theano.gpuarray.dnn.dnn_conv>` GPU-only
+    - :func:`dnn_conv3d <theano.gpuarray.dnn.dnn_conv3d>` GPU-only
       convolution using NVIDIA's cuDNN library.
 
       If cuDNN is available, by default, Theano will replace all nnet.conv3d
-      operations with dnn_conv.
-
-    - :func:`dnn_conv3d <theano.gpuarray.dnn.dnn_conv3d>` as dnn_conv() but for 3d.
+      operations with dnn_conv3d.
 
     - :func:`conv3d2d <theano.tensor.nnet.conv3d2d.conv3d>`
       Another conv3d implementation that uses the conv2d with data reshaping.
```

nouiz:
travis error aren't related and are fixed in the master, so merging.

PR #6296, opened 2017-08-10
(small fix) fix requirement-rtd

notoraptor:
Pick changes from #6283 related to `requirement-rtd.txt`.

@nouiz


nouiz:
I'll merge now. Hopefully it will work.

PR #6061, opened 2017-06-20
Spatial Transformer Network Ops using cuDNN

joaovictortr:
Goal: implement [Spatial Transformer Networks'](https://arxiv.org/abs/1506.02025) Ops using cuDNN

Issue: #5622 

Tasks:

- [x] Add Op to create and manage a spatial transformer network descriptor (cudnnSpatialTransformerDescriptor_t)
- [x] Implement Grid Generator Op
- [x] Implement Grid Sampler Op
- [x] Add gradients to Grid Generator Op (backpropagation)
- [x] Add gradients to Grid Sampler Op (backpropagation)
- [x] Document functions and classes in Theano's documentation

PR #6298, opened 2017-08-11
[WIP] CPU Spatial Transformer

joaovictortr:
Tasks:
* [ ] Abstract spatial transformer Op
* [ ] Optimization to lift spatial transformer Op to CPU implementation
* [ ] Optimization to lift spatial transformer Op to GPU implementation
* [ ] Unit test the spatial transformer Op
* [ ] Compute gradients of inputs in the CPU
* [ ] Compute gradients of affine transformation in the CPU
* [ ] Unit test the gradient implmentations


joaovictortr:
Hello @notoraptor , I've added this PR to keep track of the work on the CPU spatial transformer. There are still commits from the GPU implementation, since it is not yet merged. I think it can be solved with a rebase later on.

theano/tensor/nnet/spatialtf.py
```diff
@@ -0,0 +1,186 @@
...
+from theano.tensor import as_tensor_variable
+from theano.gof import Apply
+from theano.gof.opt import local_optimizer
+from theano.tensor.opt import register_canonicalize
+
+class AbstractSpatialTransformerOp(Op):
+    def __init__(self, border_mode):
+        self.border_mode = border_mode
+
+    def make_node(self, inp, theta, scale_height, scale_width):
```
notoraptor: I suggest that Abstract op here receives real width and height instead of scale_width and scale_height, as we do on GPU. This will give more options (e.g. by allowing to instanciate Abstract op with already known dimensions without having to use scales).

As on GPU, `spatialtf()` below should be the only one responsible for inferring real width and height before passing it to abstract op.

notoraptor:
Oh, I should have not approve this PR as it is still a WIP. Mistake here.

joaovictortr:
Hello @notoraptor , I've addresed your comment, and layed out the abstract and concrete implementations of the spatial transformer into separate files.

In the case of ``perform`` for ``DebugMode``, it should also be implemented for the gradient Ops, correct?

notoraptor:
@joaovictortr yes, ideally, as gradient ops may be also tested in debug mode.

notoraptor:
Hello @joaovictortr ! PR #6061 was merged. Can you rebase this PR ?

joaovictortr:
Hello @notoraptor , I've implemented an initial version of the gradients (not yet optimized/vectorized with NumPy). I'm currently looking for problems in the computations, because the tests fail in ``verify_grad``.

I'm using the following [code to test the implementation](https://gist.github.com/joaovictortr/9e6a662a3361ab96f3787a42a320ef1a)

notoraptor:
Hello @joaovictortr , just two quick remarks:
- To test gradient, also consider comparing gradient computed by your CPU implementation with gradient computed by cuDNN. Currently, the best reference we have is cuDNN implementation, so let's use it.
- Can you rename ops by removing the `Op` suffix ? E.g. `AbstractSpatialTransformerOp` -> `AbstractSpatialTransformer` ?

joaovictortr:
Hello @notoraptor ,  I'll change the names of the ops, as requested. As for your first point, there is also a reference implementation I have found [here](https://github.com/qassemoquab/stnbhwd). You're absolutely correct that the cuDNN should be the one we are comparing to, and by doing the comparison, I've found that the only problem remaining with the gradient computations is the gradient of the affine transformation.

Fixing that problem, I can optimize the implementation to use only NumPy.

joaovictortr:
Hello @notoraptor , the implementation of the spatial transformer is almost complete. There are two things left to do: add docstrings for the Op classes, and fix an issue with verify_grad in the gradients of the transformation.

I have changed the initial design to use concrete classes, since I had implemented the ``debug_perform`` methods using NumPy, and I was having issues with the optimizations to replace the abstract Ops when lifting them to the GPU.

Please let me know, if there is anything else.

theano/tensor/nnet/spatialtf.py
```diff
@@ -0,0 +1,413 @@
...
+        grad_grid = theano.tensor.tensor(dtype=_inp.type.dtype,
+                                         broadcastable=_inp.broadcastable)
+
+        return Apply(self, [_inp, _grid, _grad_outputs], [grad_inp, grad_grid])
+
+    def perform(self, node, inputs, output_storage):
+        inp, grid, grad_outputs = inputs
+        assert len(inp.shape) == 4
+        assert len(grid.shape) == 4
+        assert len(grad_outputs.shape) == 4
```
notoraptor:
```python
assert inp.ndim == 4
assert grid.ndim == 4
assert grad_outputs.ndim == 4
```
But, as it is already checked in make_node(), you don't need to check it here again. Consider removing these checkings here.

theano/tensor/nnet/spatialtf.py
```diff
@@ -0,0 +1,413 @@
...
+        grad_gradients[0, :] = np.sum(xf, axis=1)  # * (width_f - 1) / 2
+        grad_gradients[1, :] = np.sum(yf, axis=1)  # * (height_f - 1) / 2
+
+        grad_grid_out[0] = np.reshape(grad_gradients, grid.shape)
+
+
+class TransformerGradT(Op):
+    def make_node(self, theta, grad_grid):
+        _theta = as_tensor_variable(theta)
+        _grad_grid = as_tensor_variable(grad_grid)
```
notoraptor: Don't forget to check ndims here for theta and grad_grid.

theano/tensor/nnet/tests/test_spatialtf.py
```diff
@@ -0,0 +1,226 @@
...
+
+        cpu_out_op = spatialtf(t_inp, t_theta, t_scale_width, t_scale_height)
+        cpu_out_fn = theano.function([t_inp, t_theta, t_scale_width, t_scale_height],
+                                     cpu_out_op, mode=mode_without_gpu)
+        cpu_out = cpu_out_fn(inp, theta, scale_width, scale_height)
+
+        # Check if results are approx. equal
+        utt.assert_allclose(symb_out, cpu_out)
+
+    def test_gpu(self):
```
notoraptor: We should better let GPU tests in GPU module and not put it in CPU module.

As there is already a GPU test in GPU module, remove this test from here. Or, if it is a particular test, transfert it to GPU test module.

theano/tensor/nnet/tests/test_spatialtf.py
```diff
@@ -0,0 +1,226 @@
...
+                                      symb_out_op, mode=mode_without_gpu)
+        symb_out = symb_out_fn(inp, theta, scale_width, scale_height)
+
+        cpu_out_op = spatialtf(t_inp, t_theta, t_scale_width, t_scale_height)
+        cpu_out_fn = theano.function([t_inp, t_theta, t_scale_width, t_scale_height],
+                                     cpu_out_op, mode=mode_without_gpu)
+        cpu_out = cpu_out_fn(inp, theta, scale_width, scale_height)
+
+        # Check if results are approx. equal
+        utt.assert_allclose(symb_out, cpu_out)
```
notoraptor: I don't think the purpose of symbolic implementation should be to compare its results with your numpy-based implementation, as it is quite useless. To test your numpy-based implementation, you should better compare results with expected hardcoded concrete values, just as you did to test CPU CTC implementation.

However, about symbolic implementation, it may be useful to compare it to numpy-based implementation for benchmarking, to check which version is the fastest (in practice, symbolic implementation might be fastest depending on Theano implementations). But a benchmark is not a unit test. So I suggest to keep the symbolic implementation, remove this test (`test_symb`), replace it with a real test (which compares your values with expected hardcoded values), and manually benchmark both implementstions yourself (especially with big inputs).

theano/gpuarray/dnn.py
```diff
@@ -3717,6 +3721,89 @@ def local_dnn_argmax(op, ctx_name, inputs, outputs):
...
+    inp, _grid = inputs
+    grid = _grid.dimshuffle(1, 2, 3, 0)
+
+    if inp.ndim != 4:
+        return
+
+    if grid.ndim != 4:
+        return
+
+    sampler = GpuDnnTransformerSampler()(inp, grid)
```
notoraptor: CPU transformer sampler has a border_mode attribute, but you don't check it here. Can all CPU border modes be transferred to GPU ? I don't think so, isn'it ?

theano/tensor/nnet/spatialtf.py
```diff
@@ -0,0 +1,413 @@
...
+        grid_out = output_storage[0]
+
+        out_height, out_width = out_dims[2:]
+        grid = sampling_grid(out_height, out_width)
+        # Generate transformed grid with shape (num_batch, 2, out_height * out_width)
+        transformed_grid = np.dot(theta, grid)
+        # Dimshuffle grid into (2, num_batch, out_height * out_width)
+        transposed_grid = np.transpose(transformed_grid, axes=(1, 0, 2))
+        # Reshape into (2, num_batch, out_height, out_width)
+        grid_out[0] = np.reshape(transposed_grid, (2, num_batch, out_height, out_width)).astype(theta.dtype)
```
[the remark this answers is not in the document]
joaovictortr: It eases the CPU computations and (I assumed it) makes the grid data contiguous, but it can easily reverted to the same format as in the GPU. Otherwise, a dimshuffle is required when lifting to the GPU, as you've stated. I'll change it to use same format as the GPU.

theano/tensor/nnet/tests/test_spatialtf.py
joaovictortr: You are correct, I forgot to change this test to the dnn test. It should also check if the CPU Op is properly lifted to the GPU (e.g. by checking the apply nodes).

theano/gpuarray/dnn.py
joaovictortr: I don't think they can be transparently transferred to the GPU, and cuDNN does not provide such a parameter, nor does it is documented how the borders are treated in the implementation. Comparing the implementation on the CPU using the nearest mode, computes the same outputs, I will test with other modes and report back.

theano/tensor/nnet/opt.py
```diff
@@ -42,6 +42,8 @@ def local_inplace_sparse_block_gemv(node):
         copy_stack_trace(node.outputs[0], new_node)
         return [new_node]
     return False
+
```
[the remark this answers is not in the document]
joaovictortr: I've added these lines, because of flake8 warnings (might be some misconfiguration on my side), I thought the spaces were removed by doing the rebase.

theano/gpuarray/dnn.py
notoraptor: Ok. If nearest mode is the only one supported in cuDNN, then you should check it here and optimize only if op.border_mode == 'nearest'.

joaovictortr: I've checked with every mode, comparing the output with GPU, and they all pass here. Can you also check it?

theano/tensor/nnet/tests/test_spatialtf.py
```diff
@@ -0,0 +1,222 @@
...
+        for border_mode in ('nearest', 'mirror', 'wrap'):
+            # Setup spatial transformer on the CPU
+            op_cpu = spatialtf(t_inp, t_theta, t_scale_width, t_scale_height, border_mode)
+            fn_cpu = theano.function([t_inp, t_theta, t_scale_width, t_scale_height],
+                                     op_cpu, mode=mode_without_gpu)
+            # Evaluate CPU function
+            out_cpu = fn_cpu(inp, theta, scale_width, scale_height)
+
+            # Check results
+            utt.assert_allclose(out_gpu, out_cpu)
```
notoraptor: `out_gpu` -> `np.asarray(out_gpu)`

theano/tensor/nnet/tests/test_spatialtf.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def getInputs(self):
+        inp_shape = (np.random.randint(1, 11), 3, np.random.randint(10, 101),
+                     np.random.randint(10, 101))
+        num_images, num_channels, height, width = inp_shape
+
+        inp = np.random.random(inp_shape).astype(config.floatX)
+
+        transform = [[-1, 0, 0],
+                     [0, -1, 0]]
```
notoraptor: When I change transform value, I got severe WrongValue errors.

Extend your tests to test with various values of transform (including floating values).

joaovictortr: I'll add more tests, and use random transformation matrices, to see where the problems may be arising.

theano/tensor/nnet/tests/test_spatialtf.py
joaovictortr: Addressed in commit 88efff8440cc492522799d2205e0bc9537780a2d

theano/tensor/nnet/tests/test_spatialtf.py
joaovictortr: I agree, but I think, however, the highest priority is getting a complete functional spatial transformer. As soon as I fix the gradient of the transformation and improve the tests, translating NumPy code into Theano symbolic operations is straightforward.

notoraptor: Indeed, let's finish and fix the gradient implementation, and then we will look for ultimate corrections.

theano/tensor/nnet/tests/test_spatialtf.py
notoraptor: Hello @joaovictortr don't forget this !

notoraptor:
Updated. Not yet finished: CPU code seems ok, but still problems with gradients (let's see jenkins and travis).

jenkins test this.

@nouiz @lamblin

notoraptor:
Updated. After a week of work, I am still unable to write correct gradient implementations for grid and theta, but op implementation itself and gradient implementation for input seems to work.

Reminder: `spatial_transformation` op depends on 3 main variables: `theta`, `grid` (computed from theta and out dimensions) and `input` (images to transform). `spatial_transformation` and `grad input` seems well implemented, but tests for `grad grid` and `grad theta` still fails.

About gradients, I guess I miss just one point in the logic, but it is surprisingly hard to find it ...

jenkins test this

@abergeron @nouiz @lamblin

notoraptor:
ok to test
ready to test

notoraptor:
Just a note to tell that it seems part of problem about gradient is coming from values: e.g., for gradient wrt/ theta, if theta contains only integer values (ie. even integer values passed as floating values, like `[[1.0, 0.0, 0.0], [0.0, 1.0. 0.0]]`), then gradient fails. But if I just change it by adding just `0.000000001` to each value (=> `[[1.000000001, 0.000000001, 0.000000001], [0.000000001, 1.000000001. 0.000000001]]`), then gradient succeeds.

It's really hard to understand why such changes occurs. But I am still investigating. It seems problem is more about handling integer values.

@abergeron @nouiz @lamblin

notoraptor:
I will rewrite the code tomorrow to make it clearer, simplify it and add supplementary tests.

notoraptor:
A final work for this PR. Failing gradients tests are skipped.

@nouiz @abergeron

nouiz:
We aren't able to fix the gradient on the CPU. So we won't include that for the next release.

PR #6299, opened 2017-08-11
Fix test broken in gh-6285

nouiz:


theano/gof/tests/test_graph_opt_caching.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import absolute_import, print_function, division
-import os
+import os.path
```
notoraptor: We use `os.path.*` and `os.remove.*`, so we should just `import os`.

PR #6302, opened 2017-08-11
Tensor op, cache

borisfom:
Here, a less intrusive take.


theano/gpuarray/c_code/dnn_conv_find.cc
```diff
@@ -0,0 +1,133 @@
+#section support_code
```
notoraptor: This file does not seem to be used, so it should be removed.

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -142,6 +170,12 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
           return 1;
         }
         algo = choice.algo;
+        prev_algo.algo = (int)algo;
+        prev_algo.wsSize = worksize = choice.memory;
+        prev_algo.mathType = mathtype = choice.mathType;
```
notoraptor: This line will make compilation fail with cuDNN < V7. I think it should be surrounded with:
```C
        #if CUDNN_MAJOR >= 7
        prev_algo.mathType = mathtype = choice.mathType;
        #endif
```

theano/gpuarray/c_code/dnn_gi.c
```diff
@@ -182,6 +204,12 @@ APPLY_SPECIFIC(conv_gi)(PyGpuArrayObject *kerns, PyGpuArrayObject *output,
       }
 
       algo = choice.algo;
+        prev_algo.algo = (int)algo;
+        prev_algo.wsSize = worksize = choice.memory;
+        prev_algo.mathType = mathtype = choice.mathType;
```
notoraptor: Same here.
```C
        #if CUDNN_MAJOR >= 7
        prev_algo.mathType = mathtype = choice.mathType;
        #endif
```

theano/gpuarray/c_code/dnn_gi.c
```diff
@@ -293,7 +303,47 @@ APPLY_SPECIFIC(conv_gi)(PyGpuArrayObject *kerns, PyGpuArrayObject *output,
...
+  // save worksize for next time/cache
+  prev_algo.wsSize = worksize;
+  
+  // Add to the cache
+  if (params->choose_algo)
+      dnn_conv_update_cache(hashkey, prev_algo);
+    }  
+
+    #ifdef DEBUG
+    char algorithm_name[128];
```
notoraptor: `algorithm_name` is already defined earlier with `#idef DEBUG`, so this line should be removed.

theano/gpuarray/c_code/dnn_gw.c
```diff
@@ -1,14 +1,17 @@
...
 reuse_algo = 0;
-prev_algo = PARAMS->conv_algo;
 memset(prev_img_dims, 0, sizeof(prev_img_dims));
 memset(prev_top_dims, 0, sizeof(prev_top_dims));
 
 #section support_code_struct
-
-int reuse_algo;
-cudnnConvolutionBwdFilterAlgo_t prev_algo;
+#line 12 "dnn_gw.c"
```
notoraptor: I think it should be `11` instead of `12` here!

theano/gpuarray/c_code/dnn_gw.c
```diff
@@ -184,6 +210,12 @@ APPLY_SPECIFIC(conv_gw)(PyGpuArrayObject *input, PyGpuArrayObject *output,
         }
 
         algo = choice.algo;
+        prev_algo.algo = (int)algo;
+        prev_algo.wsSize = worksize = choice.memory;
+        prev_algo.mathType = mathtype = choice.mathType;
```
notoraptor: Same comment about mathtype:
```C
        #if CUDNN_MAJOR >= 7
        prev_algo.mathType = mathtype = choice.mathType;
        #endif
```

theano/gpuarray/subtensor.py
```diff
@@ -1106,15 +1106,19 @@ def gpu_kernels(self, node, nodename):
   old = *base;
   do {
     assumed = old;
-    sum = __float2half_rn(
+    ga_half old_perm;
+    __HALF_TO_US(old_perm)  = __byte_perm(old, 0,
```
notoraptor: According to @abergeron comment in your old PR ( here: https://github.com/Theano/Theano/pull/6283/files#r132486589 ), those definitions should be moved to libgpuarray in the future. So, maybe the `theano.gpuarray.subtensor` module should not be modified in this PR for the moment ?

theano/gpuarray/c_code/dnn_fwd.c
borisfom: No, why, I am defining this type for CUDNN < 7

theano/gpuarray/c_code/dnn_gw.c
borisfom: Let me know if my typedef did not work for you.

theano/gpuarray/subtensor.py
borisfom: Well, tests that use atomics, fail (compile errors) if I don't make this change.

borisfom:
I have added group convolutions, also groups should have been part of the cache - fixed.

theano/gpuarray/c_code/dnn_fwd.c
notoraptor: @borisfom In fact, you use `choice.mathType` here, and `choice` is a `cudnnConvolutionFwdAlgoPerf_t`, but this structure has field `mathType` on cuDNN 7 only, not on cuDNN 6 and earlier.

So, when compiling with cuDNN < 7, this line generates an error.

On cuDNN 7:
```C
typedef struct {
    cudnnConvolutionFwdAlgo_t   algo;
    cudnnStatus_t               status;
    float                       time;
    size_t                      memory;
    cudnnDeterminism_t          determinism;
    cudnnMathType_t             mathType; // Field mathType is defined.
    int                         reserved[3];
} cudnnConvolutionFwdAlgoPerf_t;
```

On cuDNN 6:
```C
typedef struct {
    cudnnConvolutionFwdAlgo_t   algo;
    cudnnStatus_t               status;
    float                       time;
    size_t                      memory;
    cudnnDeterminism_t          determinism;
    // No mathType defined.
    int                         reserved[4];
} cudnnConvolutionFwdAlgoPerf_t;
```

Same for equivalent structures in `dnn_gw.c` and `dnn_gi.c`.

theano/gpuarray/c_code/dnn_conv_base.c
```diff
@@ -3,6 +3,23 @@ cudnnTensorDescriptor_t APPLY_SPECIFIC(input);
 cudnnTensorDescriptor_t APPLY_SPECIFIC(output);
 cudnnFilterDescriptor_t APPLY_SPECIFIC(kerns);
 
+static int c_set_groups_for_conv(cudnnConvolutionDescriptor_t desc, int groups) {
+#if CUDNN_MAJOR >= 7
+  if (groups > 1) {
+    cudnnStatus_t err = cudnnSetConvolutionGroupCount(desc, groups);
```
affanv14: Ok, I can make those changes

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -167,31 +204,17 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
             cuda_exit(c->ctx);
             return 1;
           }
+          prev_algo.algo = algo;
+          // no tensor_op returned from Get()
```
nouiz: prev_algo.wsSize is not set. Is that normal?

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -167,31 +204,17 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
...
-    if (params->choose_once) {
-      reuse_algo = 1;
-    } else {
-      for (unsigned int i = 0; i < PyGpuArray_NDIM(input); i++) {
-        prev_img_dims[i] = PyGpuArray_DIM(input, i);
-        prev_kern_dims[i] = PyGpuArray_DIM(kerns, i);
-      }
-    }
-  }
+  } else { /* choose_algo */
```
nouiz: I think this else should be removed. Sometimes there is bugs in cudnn and the code bellow allow us to override cudnn algo selected. We do not want to get rid of that.

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -100,10 +112,25 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
                       PyGpuArray_DIM(kerns, i) == prev_kern_dims[i]);
       }
     }
-
+    
+    
     if (!reuse_algo) {
```
nouiz: // once or shapes changed

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -72,24 +76,32 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
...
+  std::string hashkey;
+  
   #ifdef DEBUG
   char algorithm_name[128];
   #endif
 
   cuda_enter(c->ctx);
-
+  
   if (params->choose_algo) {
```
nouiz: IRL discussion with @notoraptor. We think we should get rid of the prev_img_dims and prev_kern_dims (and this whole if).

This seem to allow to bypass your cache if the input and kernel shapes don't change. In that case, it set reuse_algo to 1. This won't look at the other characteristic your cache look at.

Do you think we should care the the speed to compute the hash and query the cache? I have the impression we could also query it before each convolution, but being sure of that would be great.

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -167,31 +204,17 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
...
           }
+          prev_algo.algo = algo;
+          // no tensor_op returned from Get()
+          prev_algo.mathType = mathtype = CUDNN_DEFAULT_MATH;
+	  // fprintf(stderr, "(cudnnGetConvolutionForwardAlgorithm: (err:%d), algo: %d\n", err, algo);
       }
-      prev_algo = algo;
-    } else {
-      algo = prev_algo;
+    } else {
```
nouiz: Can you add this comment to help read the code

// if (!(reuse_algo || use_cached)), so for once algo after the first call as reuse_algo will be True.

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -309,7 +333,52 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
       cuda_exit(c->ctx);
       return 1;
     }
+    // save worksize for next time/cache
+    prev_algo.wsSize = worksize;
 
+    // Add to the cache
+    if (params->choose_algo)
```
nouiz: if (params->choose_algo && !params->choose_once)

notoraptor:
ok to test

notoraptor:
I activate Jenkins, so that we can also check if everything compiles with cuDNN < 7 !

But there are still flake8 errors to fix first.
```
/home/travis/build/Theano/Theano/theano/gpuarray/subtensor.py:1133:15: W291 trailing whitespace
```

nouiz:
I merged changed in the master for group convolution. Can you rebase this PR on the master? This would allow @affanv14 to add the tests and do the modification I wrote about.

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Right, sorry!

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Right - my assumption was that the rest of the parameters don't change (and as you pointed out, even convolution descriptor may have changed in Theano model). I don't have data for small images - but I am sure for practical cases, querying cache should be fine.

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Yes it is set here: 
  // if FindEx was used (choose_time), workspace size is set.                                                                                                                                                                                                   
  if (!(reuse_algo || use_cached || params->choose_time))

theano/gpuarray/c_code/dnn_fwd.c
borisfom: I would disagree. Sections like that may create false impression for the user that the algo selected actually works. If there is a CUDNN bug/limitation, Theano should get CUDNN_NOT_SUPPORTED and report it to the user (with additional info how to work around) so that the workaround ends up in user's code. 
Can we put this code in a fallback section, like:
 if (status==CUDNN_NOT_SUPPORTED || ! params->choose_algo) ?

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Why, we have successfully used cache hits across many (repeatedly created) modules with same parameters and choose_once.

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Actually: all those cases will be covered by the fallback after first cudnnGetConvolutionForwardWorkspaceSize() call - including Get() cases - so the safest course will be to simply remove them.

theano/gpuarray/c_code/dnn_fwd.c
borisfom: That said - _once should be kept as is, to avoid using cache repeatedly when it's not needed.

theano/gpuarray/c_code/dnn_fwd.c
nouiz: This cover 2 types of "problems":

1) Work around bug in cudnn.
1) Not implemented cases for some specific algo in cudnn.

We absolutly need to keep 1). We need workaround/bugfix in Theano for cudnn bug then what the cudnn release cycle give us. We can't wait for cudnn release.

For 2), specifiing manually which algo to use for each layer is pretty not fun for user. So user specify one algo for all layers most of the time. So if one algo don't support some cases, it would just crash if we don't fall back to another algo in that case.

Also, Theano hardcode a specific algo by default, we need to make sure it work in all cases, which mean also fall back to other algo if it don't support those cases. This is a sub case of 2).

This design choice could be changed, but we won't do that just before a release.

So for both reason, it should be reverted from this PR. for 2), it could be discussed after the release and we could change it if enough people agree with that. But I think in practice, people use time_once or the Theano default and after your PR, they could use the cache (possibly the new default after the release. Also to discuss).

theano/gpuarray/c_code/dnn_fwd.c
nouiz: If the user specify time_once of guess_once, it mean the cache isn't used. As normally, the algo is specified globally, what I wrote made sence.

But you raised a good point, in case people specify the algo per layer, the updating the cache make sense. To help us not forget that case, can you add a comment for this? Something like:

//update the cache, even if this node use *_once algo in case the user specify the algo per layer and not globally.

nouiz:
@borisfom We want to finish this before the RC that we want to make this week. To help speed that up, @notoraptor will push commits to this PR about the comment we have. This will make the loop between comments/commit faster.

Can you also check the commits he does to make sure we don't forget cases?

theano/gpuarray/c_code/dnn_conv_find.c
```diff
@@ -13,7 +13,7 @@ typedef std::unordered_map<std::string, AlgoRec> AlgoCache;
 #endif
 #include "pthread.h"
 
-#line 10 "dnn_conv_find.cc"
+#line 10 "dnn_conv_find.c"
```
notoraptor: From what I remarked, this helps to find the right line from where a compilation issue occurs when compilation fails, so it's useful for debugging purposes.

Previously, when a compilation error occurs, g++ returned line numbers based on `mod.cpp`, and then we must infer corresponding line in Theano C code files. Now (with `#line`), g++ directly returns a line number based on Theano C code file (here `dnn_conv_find.c`).

theano/gpuarray/c_code/dnn_fwd.c
borisfom: @nouiz: actually, this is not about 'algo-per-layer' : I don't think we should cache user-specified algos as they may not be optimal. However, if it's guess_once - locally or globally - we do want to cache it so that next time we create the module with same parameters (real life use case) we don't need to run benchmarks.

borisfom:
I have integrated PR#1 and fixed flake8 errors.
Should be ready to merge. I do have some rnn test errors on Volta though - should not be related to this PR.

theano/gpuarray/c_code/dnn_fwd.c
borisfom: @nouiz : no convolution tests failed for me after I have put all of falbacks after cudnnGetConvolutionForwardWorkspaceSize.

notoraptor: There are tests that fail on Jenkins: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/666/

For example, this test is normally managed by fallbacks, but now it fails: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/666/testReport/junit/theano.gpuarray.tests/test_dnn/test_batched_conv_small__65537__2__2__2____1__2__2__2___5_/

That's because, for this test, no CUDNN_STATUS_NOT_SUPPORTED is raised, everything seems to goes well but there is a bug in cuDNN that makes it return wrong values.

So, I also think we really need to maintain fallbacks.

notoraptor:
Hello @borisfom @nouiz ! I pushed 2 commits with many small typo fixes and other small fixes to clear code, simplify it, and secure it (e.g. to better manage errors when generating hash key).

I have not yet re-added the fallbacks, I will do it in a next commit. Fallbacks are still needed (see my comment above: https://github.com/Theano/Theano/pull/6302#discussion_r133702579 ).

I have also removed modifications to `theano/gpuarray/subtensor.py` because I think this PR should focus mainly on cuDNN only, and also because @abergeron started to remove atomic definitions from theano side (see his PR here: https://github.com/Theano/Theano/pull/6317 ).

notoraptor:
**NB**:

Also, about current code logic in `dnn_fwd.c`, when getting worksize for no-timed algorithm, if a CUDNN_STATUS_NOT_SUPPORTED is raised, then an algorithm is chosen as fallback (currently CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM) and a new worksize is retrieved for this fallback algorithm. Later, the new retrieved worksize is added to cached algorithm (`prev_algo.wsSize = worksize;`), but the cached algorithm is not updated with the fallback algorithm. That means cache may contain an algorithm with a worksize that doesn't belong to it. I think it's an issue, isn't it ?

nouiz:
From what you wrote, yes.

notoraptor:
From tests, I remarked that caches (an `reuse_algo` value, too) can be shared across functions that were not compiled with same algorithm (e.g. a function with `time_once` and a function with `guess_once` may share the same cache). I think it's because an op is compiled once and then loaded once in memory and reused every time we need it (without being reinitialized). So, as long as the program runs (ie. if we are in memory while switching functions), a cache filled in one function may be used by next function.

So, I think we should add a bit to the hash to make distinction between a timed and a guessed algorithm. Otherwise, if a function compiled with guess_once is called first, then a guessed algorithm will be put in the cache (and `reuse_algo` will be set to 1). Then, if a function compiled with time_once is called after, it will take the guessed algorithm saved in memory (as `reuse_algo` is 1), but that is not what we want.

@nouiz @borisfom

theano/gpuarray/c_code/dnn_conv_base.c
```diff
@@ -33,3 +56,205 @@ if (APPLY_SPECIFIC(output) != NULL)
...
+                     " but received gradient with shape %dx%dx% dx%d",
+                     expected_output_dims[0], expected_output_dims[1] / groups,
+                     expected_output_dims[2], expected_output_dims[3],
+                     PyGpuArray_DIMS(output)[0], PyGpuArray_DIMS(output)[1],
+                     PyGpuArray_DIMS(output)[2], PyGpuArray_DIMS(output)[3]);
+        return "";
+      }
+    } else if (PyGpuArray_NDIM(filter) == 5) {
+      if ((PyGpuArray_DIMS(output)[0] != expected_output_dims[0]) ||
+          (PyGpuArray_DIMS(output)[1] != expected_output_dims[1]) ||
```
affanv14: @borisfom regarding https://github.com/borisfom/Theano/pull/1#issuecomment-322869458 this line should be updated for 3d grouped convolutions. Something similar must be done to the two backward passes as  well.

borisfom: There used to be  the same code for forwards and back passes before (all using forward function to check) - I assume, the shape relation is the same.

borisfom: Also, (for CUDNN7) the groups are already included here - we set groups on the descriptor before this check.

theano/gpuarray/c_code/dnn_fwd.c
borisfom: I see - I ignored that one failure as it was a test that expected failure. If the values are not good then override (not the fallback) is needed indeed. But this is only one case - how about the rest of overrides? Do you still need them ?

borisfom: For the case 2) - we could be able to stay with fallback if we cached it - in its own section, as proposed earlier.

notoraptor: Currently there are fallbacks for fft/fft_tiling and a fallback for 3D supported conv algos in dnn_fwd.
1. We need to decide if we put a fallback for 3D supported conv algos in all computations (ie. in dnn_gw and dnn_gi too)
2. Checkings for fft/fft_tiling seems quite old, so we may need to update them. Or, as related cases should raise a CUDNN_STATUS_NOT_SUPPORTED, we may drop these checkings and let worksize fallback do the job.

I think a discussion is needed for that, @nouiz @abergeron . But in any case, we need an entry for fallback checkings!

I am preparing new commits about that. I can push them in next minutes, so that we can discuss about it tonight and tomorrow.

notoraptor:
I added a commit with my last suggestions. To summarize:
-  Removed `AlgoRec.dataType` field as it is used nowhere.
- Added some tests, to check at least if everything runs well.
- Tried to standardize all three files (dnn_fwd, dnn_gi), to have same code logic and organization across these three codes.
- Added fallback entries as fallback functions.
- Added `"-t"` to hash for timed algorithms.

There are still Jenkins errors that will certainly occur again ( http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/666/ ) as I didn't try to fix them yet, I will see tomorrow.

What do you think ? @borisfom @nouiz

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Actually, I think it may be better idea to not do override at all, but add those cases to dnn_conv_shape   checks. At least the one testcase that is falling can be addressed this way as there are no meaningful override there.

borisfom:
@notoraptor : commit looks good to me - except I would probably call the functions 'override', not fallback.  Also, I will check if we do need dataType - me may - if currently we are always set F32 for F16 data, we may need to try F16 to tale full advantage of HMMA/Tensor Op. No problem - I will add it back if we do.

theano/gpuarray/c_code/dnn_conv_base.c
```diff
@@ -33,3 +56,205 @@ if (APPLY_SPECIFIC(output) != NULL)
...
+
+static std::string dnn_conv_shape(cudnnTensorDescriptor_t inputDesc, PyGpuArrayObject* input,
+				  cudnnFilterDescriptor_t filterDesc, PyGpuArrayObject* filter,
+				  cudnnConvolutionDescriptor_t convDesc,
+				  PyGpuArrayObject* output, int groups)
+{
+    cudnnDataType_t  dType;
+    std::ostringstream s;
+    int expected_output_dims[5] = {0};
+    cudnnStatus_t err = cudnnGetConvolutionNdForwardOutputDim(convDesc, inputDesc, filterDesc,
```
notoraptor: There are 2 Jenkins tests that will still fail:
- http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/672/testReport/junit/theano.gpuarray.tests.test_abstractconv/TestDnnConv2d/test_gradinput_impossible_output_shapes__1__1__1__1____1__1__2__2___2___valid___1_/
- http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/672/testReport/junit/theano.gpuarray.tests.test_abstractconv/TestDnnConv2d/test_gradinput_impossible_output_shapes__1__1__1__1____1__1__2__2___3___valid___1_/

These tests expect cuDNN to fail for inputs/filters shapes that should produce output shapes assumed to be bad. They currently fail with default algorithms (none for dnn_gi and dnn_gw, small for dnn_fwd), but they succeed when I test with runtime-chosen algorithms (e.g. time_once). I think it's because output checking that occurs here is now executed only for runtime-chosen algorithms, as `dnn_conv_shape()` is called only in such cases.

I will put the output checking to an independent function and ensure it is always called.

notoraptor:
Rebased to resolve conflicts, and updated to fix last failing tests.

Let's see if Jenskins/Travis fail again.

@nouiz @borisfom

borisfom:
@notoraptor : Tests passed, looks good to me.

theano/gpuarray/c_code/conv_desc.c
```diff
@@ -1,5 +1,18 @@
...
+#if CUDNN_MAJOR >= 7
+  cudnnStatus_t err = cudnnSetConvolutionGroupCount(desc, groups);
+  if (err != CUDNN_STATUS_SUCCESS) {
+    PyErr_Format(PyExc_RuntimeError,
+		   "error setting groups for convolution : %s",
+		   cudnnGetErrorString(err));
+    return -1;
+  }
+#endif
+  return 1;
```
nouiz: return 0 on success?

theano/gpuarray/c_code/dnn_conv_base.c
```diff
@@ -3,6 +3,43 @@ cudnnTensorDescriptor_t APPLY_SPECIFIC(input);
 cudnnTensorDescriptor_t APPLY_SPECIFIC(output);
 cudnnFilterDescriptor_t APPLY_SPECIFIC(kerns);
 
+static int c_check_groups_for_conv(cudnnConvolutionDescriptor_t desc, int groups) {
```
nouiz: I don't like the name. "Check" seem to indicate it would return True/False to me. What about renaming it to "get_groups_for_conv".

theano/gpuarray/c_code/conv_desc.c
notoraptor: Indeed. Done.

theano/gpuarray/c_code/dnn_conv_base.c
```diff
@@ -33,3 +70,220 @@ if (APPLY_SPECIFIC(output) != NULL)
...
+#if __cplusplus < 201103L
+#include <tr1/unordered_map>
+typedef std::tr1::unordered_map<std::string, AlgoRec> AlgoCache;
+#else
+#include <unordered_map>
+typedef std::unordered_map<std::string, AlgoRec> AlgoCache;
+#endif
+#include "pthread.h"
+
+#line 73 "dnn_conv_base.c"
```
nouiz: good line number?

notoraptor:
@nouiz Done.

I suggest to wait at least for Jenkins tests to pass.

nouiz:
thanks

PR #6306, opened 2017-08-14
Cudnn v7 grouped convolutions

affanv14:
Adds the grouped convolutions feature included in the recently released cudnn v7
tasks:

- [x] Implement cudnn v7 grouped 2d convolutions 
- [ ] Implement cudnn v7 grouped 3d convolutions

### DIFF
--- a/theano/gpuarray/c_code/dnn_base.c
+++ b/theano/gpuarray/c_code/dnn_base.c
@@ -1,7 +1,7 @@
 #section support_code
 
 static int
-c_set_tensor_for_conv(PyGpuArrayObject *var, cudnnTensorDescriptor_t desc, size_t groups) {
+c_set_tensorNd(PyGpuArrayObject *var, cudnnTensorDescriptor_t desc) {
   cudnnDataType_t dt;
   size_t ds;
   switch (var->ga.typecode) {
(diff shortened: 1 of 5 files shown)


notoraptor:
Hello ! Just to notice that your new code here won't work on cuDNN V6 (because `cudnnSetConvolutionGroupCount()` does not exist for versions prior to v7) !

Should we definitively drop support for all cuDNN versions prior to v7, @nouiz @abergeron ?

affanv14:
> Hello ! Just to notice that your new code here won't work on cuDNN V6 

yes, How did I forget

borisfom:
I think the group CUDNN7 feature can be implemented in much less invasive fashion. I don't think Theano will drop CUDNN6 right now. I will submit another PR.

borisfom:
See https://github.com/Theano/Theano/pull/6302

affanv14:
Well, it was discussed it would be best for the looped version of cudnn grouped convolutions be removed and use the GPUCorrMM version for cudnn < 6 . This was because by the time cudnn v8 comes out the looped would have to be taken out anyway. Also to be noted is the fact that cudnn versions have  short release cycles so the looped version would be removed very soon.

nouiz:
@affanv14 you did what I wrote, I had forgotten that we where looping for group conv 2d already. So I di not think what I wrote would remove some feature that is currently in the master. Even if it isn't there for a long time, it is better to keep it. Especially if @borisfom did it.

As this affect the cache @borisfom is working on, I'll close this PR and let this continue in gh-6302.

@affanv14 can you make a PR to gh-6302 PR that will add the tests for group conv 3d?

affanv14:
@nouiz Alright. Shall I make the optimizer changes for cudnn and also the tests in https://github.com/Theano/Theano/pull/6267  itself? so that once that it merged, @borisfom can then rebase.

nouiz:
I think it will be better that you make a PR to this PR. (so mostly a PR to
boris Theano fork on the branch that this PR is based on).

Mixing them in different PR isn't great. I'll try to review/merge gh-6267
today.

PR #6312, opened 2017-08-15
Specifying culprit node in `MissingInputError`

shawntan:
The error message here is unclear about the node causing problem. Added the `repr(var)` to be a bit more specific.

### DIFF
--- a/theano/scan_module/scan_op.py
+++ b/theano/scan_module/scan_op.py
@@ -208,7 +208,7 @@ class Scan(PureOp):
             # Do the missing inputs check here to have the error early.
             for var in theano.gof.graph.inputs(self.outputs, self.inputs):
                 if var not in self.inputs and not isinstance(var, theano.Constant):
-                    raise theano.gof.MissingInputError("ScanOp is missing an input.")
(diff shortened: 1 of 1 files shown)


theano/scan_module/scan_op.py
```diff
@@ -208,7 +208,7 @@ def tensorConstructor(broadcastable, dtype):
             # Do the missing inputs check here to have the error early.
             for var in theano.gof.graph.inputs(self.outputs, self.inputs):
                 if var not in self.inputs and not isinstance(var, theano.Constant):
-                    raise theano.gof.MissingInputError("ScanOp is missing an input.")
+                    raise theano.gof.MissingInputError("ScanOp is missing an input: %s" % repr(var))
```
nouiz: This is good. Before merging, I have an idea to make this better, you could call theano.gof.utils.get_variable_trace_string(var) and add the output of that call to the string. It should give even more time.

I don't have time to do it, so if you don't just tell it and I'll merge as is this PR.

nouiz:
As no news and close to a release, I'll merge now.

thanks

shawntan:
Sorry I just saw this. I'll look further into this, thanks!

PR #6315, opened 2017-08-16
Adding tensor6 and tensor7 constructors

gvtulder:
The upcoming unshared convolution requires 6D tensors, so I think it would be useful to add those to the list of available tensor constructors. (Is it possible to have too many of those? The list is getting quite long.)

Since 3D unshared convolution will need 7D tensors, I've added those as well.


doc/generate_dtype_tensor_table.py
```diff
@@ -18,17 +18,19 @@
...
 print(hdr)
-print('Constructor  dtype       ndim shape        broadcastable')
+print('Constructor  dtype       ndim shape            broadcastable')
 print(hdr)
 for letter in letters:
     for shape in shapes:
         suff = ',)' if len(shape[1])==1 else ')'
         s = '(' + ','.join('1' if b else '?' for b in shape[1]) + suff
-        print('%s%-10s  %-10s  %-4s %-11s  %-20s' %(
+        print('%s%-10s  %-10s  %-4s %-15s  %-20s' %(
```
nouiz: I didn't look at the output, but I suport it print something like `(False, ....)`. It it isn't too hard, changing it to `(False, ) * 6` would make this more clear I think. Anyway, just a suggestion.

gvtulder:
Second try. I've summarised the long lists of `(False, False, ...)` to `(False,) * 6` and `(False,) * 7` now.

PR #6316, opened 2017-08-16
Add unravel_index and ravel_multi_index

gvtulder:
These two ops would make the NumPy functions `unravel_index` and `ravel_multi_index` available for Theano arrays.

It's a bit annoying that the `unravel_index` function needs an extra `ndim` parameter, because the length of a `Shape` cannot be determined. Is there some clever trick to get the shape of a shape? The `ndim` of the variable that the shape belongs to is known at compile time, after all, so in theory you shouldn't have to specify that manually.

Most of the documentation is borrowed from the NumPy docs. Docs for other Theano functions gave me the idea that this is acceptable.


nouiz:
We use tensor.basic.get_vector_length to try to guess the shape of a vector in other ops that need that. It cover most cases, but not all of them. So ops define ndim=None and try to guess is None. So the user can add the information in case we can't infer it.

theano/tensor/basic.py
```diff
@@ -2588,6 +2588,175 @@ def nonzero_values(a):
     return a.flatten()[flatnonzero(a)]
 
 
+class UnravelIndex(gof.Op):
```
nouiz: basic.py is getting pretty long. I would put the new op in tensor/extra_ops.py to help split it. extra_ops is ops that isn't used very frequently.

theano/tensor/basic.py
```diff
@@ -2588,6 +2588,175 @@ def nonzero_values(a):
...
+    __props__ = ('order',)
+
+    def __init__(self, order='C'):
+        self.order = order
+
+    def make_node(self, indices, dims, ndim):
```
nouiz: This isn't super. We expect at some places in Theano that we can re-create a new node just from the inputs. To respect that, you would need to also add ndim to the inputs of this node just to be sure to don't have problems elsewhere. As it will be a constant, it should not be a problem.

I would do ndim=None and call get_vector_lenght() here if ndim is None.

theano/tensor/basic.py
```diff
@@ -2588,6 +2588,175 @@ def nonzero_values(a):
     return a.flatten()[flatnonzero(a)]
 
 
+class UnravelIndex(gof.Op):
+    __props__ = ('order',)
+
+    def __init__(self, order='C'):
+        self.order = order
```
nouiz: assert that order is C or F.

theano/tensor/basic.py
```diff
@@ -2588,6 +2588,175 @@ def nonzero_values(a):
...
+            if i.dtype not in int_dtypes:
+                raise TypeError("'%s' object cannot be interpreted as an index" % str(i.dtype))
+        if dims.dtype not in int_dtypes:
+            raise TypeError("'%s' object cannot be interpreted as an index" % str(dims.dtype))
+        if dims.ndim != 1:
+            raise TypeError("dims must be a 1D array")
+
+        return gof.Apply(
+            self, multi_index + [dims],
+            [TensorType(dtype=multi_index[0].dtype,
```
nouiz: from a quick test of numpy, the output dtype seem to be always int64 on my computer. It don't seem based on the input dtype.

theano/tensor/basic.py
```diff
@@ -2588,6 +2588,175 @@ def nonzero_values(a):
...
+    else:
+        return tuple(res)
+
+
+class RavelMultiIndex(gof.Op):
+    __props__ = ('mode', 'order')
+
+    def __init__(self, mode='raise', order='C'):
+        self.mode = mode
+        self.order = order
```
nouiz: assert mode and order are valid values.

theano/tensor/tests/test_basic.py
```diff
@@ -2744,6 +2745,130 @@ def check(m):
...
+
+            # reference result
+            ref = np.ravel_multi_index(multi_index, shape, mode, order)
+
+            def fn(mi, s):
+                return function([], ravel_multi_index(mi, s, mode, order))
+
+            # shape given as a tuple
+            f_array_tuple = fn(multi_index, shape)
+            f_symb_tuple = fn(multi_index_symb, shape)
```
nouiz: I don't think it is very useful to compile with constant and shared variable. What about doing only one of them here and the other around line 2841? This would help speed up our tests that is already too long.

idem for the other test classe.

theano/tensor/basic.py
gvtulder: I moved `ndim` to `__init__` instead, because that seemed the easiest solution. It's now similar to how it is done in `Reshape`.

gvtulder:
Thanks, nice to learn about `get_vector_length`. I've made the changes you asked for.

nouiz:
thanks

PR #6317, opened 2017-08-16
Work to integrate libgpuarray 0.7 changes.

abergeron:
Don't merge before https://github.com/Theano/libgpuarray/pull/499 (merged)

fix https://github.com/Theano/Theano/issues/6363


theano/gpuarray/opt.py
```diff
@@ -1091,25 +1091,19 @@ def local_gpua_advanced_boolean_subtensor(op, context_name, inputs, outputs):
...
-    # This is disabled on non-cuda contexts
-    if context.kind != b'cuda':
-        return None
-
     x, y, ilist = inputs
 
     set_instead_of_inc = op.set_instead_of_inc
 
-    compute_capability = int(context.bin_id[-2])
-    if (compute_capability >= 2 and x.ndim == 1 and y.ndim == 0 and
```
nouiz: Why remove the compute_capability for CUDA? Is that tested elsewhere?

theano/gpuarray/basic_ops.py
```diff
@@ -701,39 +701,35 @@ def c_code(self, node, name, inputs, outputs, sub):
         if (%(name)s_tmp == NULL)
           %(fail)s
 
-        if (%(out)s != NULL && GpuArray_IS_C_CONTIGUOUS(&%(out)s->ga) &&
+        if (%(out)s == NULL || GpuArray_IS_C_CONTIGUOUS(&%(out)s->ga) ||
```
nouiz: ! contiguous

theano/gpuarray/__init__.py
```diff
@@ -49,21 +49,25 @@ def init_dev(dev, name=None, preallocate=None):
...
+        args = dict()
         if config.gpuarray.cache_path != '':
-            os.environ['GPUARRAY_CACHE_PATH'] = config.gpuarray.cache_path
+            args['kernel_cache_path'] = config.gpuarray.cache_path
         if preallocate is None:
             preallocate = config.gpuarray.preallocate
+        if preallocate < 0:
+            args['max_cache_size'] = 0
+        else:
+            args['initial_cache_size'] = preallocate
```
nouiz: Something don't work or I don't understand. If you don't define max_cache_size, it is by default 0, but you always call gpucontext_props_alloc_cache() in the c code and that method assert that the the max size is bigger then the initial. Which end up being false here.

theano/gpuarray/opt.py
abergeron: CUDA 7.0 is the minimum version we support and it doesn't work with devices that have compute capability below 2 so this check is redundant.

If there are other tests for cc 2.0+ we should also remove them.  Things that check for higher compute capabilities would need to stay.

theano/gpuarray/__init__.py
abergeron: You're right.  I'll change de default value for max_cache_size to basically infinite.

theano/gpuarray/basic_ops.py
abergeron: fixed

abergeron:
I did a rebase to use the new required tag feature for jenkins to be able to test properly.

nouiz:
jenkins test this

nouiz:
jenkins failed with this errors: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/746/testReport/junit/theano.gpuarray.tests.test_abstractconv/TestDnnConv2d/test_all__8__1__6__6____5__1__2__2____1__1____half___True__True___1__1__/

I resterted it.

nouiz:
I fixed some jenkins errors.

nouiz:
rebased and bumped jenkins to v0.7.1 of libgpuarray. When CI pass, we could merge. but it would be good to update the installation doc about this versoin:

- [ ] update installation doc about the conda channel.

abergeron:
I just want to note that I'm ok with the latest changes, but since I'm the one that opened this PR I can't approve it.

Issue #2655, opened 2015-03-23
Add a flag to disable pickling test values


abergeron:

lamblin:
Also, a warning should probably be raised by default if there are test values at pickling time, and the flag was not explicitly set to pickle them.
The problem is that it can take tens of gigabytes when pickling, and it is not obvious where that comes from.

abergeron:
Then maybe we should just not pickle them by default.

nouiz:
Agreed

abergeron:
Do we really need a config flag then?  Is there any use case where it would make sense to pickle test values?

nouiz:
To show error that get triggered by test value? To give test value that can
be used with theano.function_dump?

If you want to just remove them without a flag, I do not have an objection

lamblin:
Does deep_copy use the pickling mechanism?
Test values of intermediate node do not matter much, but it would be surprising to have the test values for leaf variable disappear without warning, and without a way of getting them back.

Another way would be to only save the test values of variables with no owner, and recompute them when unpickling.

nouiz:
deep_copy use pickling

lamblin:
Then I would definitely expect my deep_copied variable to still have their test values.

abergeron:
We can always implement \__copy__ and \__deepcopy__ on Variables to not strip the test_value

abhishekgarg16:
Hi @nouiz , can you please explain in little more detail what exactly needs to be done?

nouiz:
In the file theano/configdefaults.py, add a new Theano flag. I would guess the name "pickle_test_value" that default to True, but can be set to False.

Then in the class Variable(), you can add a \__getstate__() method that will don't cause the pickle of tag.test_value(). There is some example elsewhere of \__getstate__(). grep for it in Theano.

thanks

abhishekgarg16:
Ok, i will be working on it and submit a pull request soon.

SadhanaSrinivasan:
Hi @nouiz 

Should I be modifying the __getstate__ function in class Variable(node) (the one in theano/gof/graph.py) to throw an exception instead of returning the dictionary when the flag is set to false? Or should I be writing a new __getstate__ for class Variable (the one in theano/gof/unify.py) ?

In the latter case, how does the __getstate__ stop the pickling?

nouiz:
It is for the class Variable in the file gof/graph.py. not in unify.py. 

[...]

PR #6319, opened 2017-08-17
Flag to disable pickling test values

yikangshen:
A Flag is added to disable pickling test values. By default, it's False. A warning will raised by default if there are test values at pickling time.

fix gh-2655

### DIFF
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -55,6 +55,13 @@ AddConfigVar('warn_float64',
              in_c_key=False,
              )
 
+AddConfigVar('pickle_test_value',
+             "Dump test values while pickling model. "
+             "If True, test values will be dumped with model.",
+             BoolParam(False),
+             in_c_key=False,
+             )
+
 AddConfigVar('cast_policy',
              'Rules for implicit type casting',
              EnumStr('custom', 'numpy+floatX',
--- a/theano/configdefaults.py
+++ b/theano/configdefaults.py
@@ -58,7 +58,7 @@ AddConfigVar('warn_float64',
 AddConfigVar('pickle_test_value',
              "Dump test values while pickling model. "
              "If True, test values will be dumped with model.",
-             BoolParam(False),
+             BoolParam(True),
              in_c_key=False,
              )
 
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -7,6 +7,8 @@ from collections import deque
 from copy import copy
 from itertools import count
 
+import warnings
+
 import theano
 from theano import config
 from theano.gof import utils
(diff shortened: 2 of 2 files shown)


theano/configdefaults.py
```diff
@@ -55,6 +55,13 @@ def floatX_convert(s):
              in_c_key=False,
              )
 
+AddConfigVar('pickle_test_value',
+             "Dump test values while pickling model. "
+             "If True, test values will be dumped with model.",
+             BoolParam(False),
```
abergeron: After some discussion with @nouiz, we now believe that `True` would be a better default for the release, in order to avoid breaking anything.

We plan to change it to `False` after the release.

theano/gof/graph.py
```diff
@@ -524,6 +526,11 @@ def eval(self, inputs_to_values=None):
     def __getstate__(self):
         d = self.__dict__.copy()
         d.pop("_fn_cache", None)
+        if (not config.pickle_test_value) and (hasattr(self.tag, 'test_value')):
+            warnings.warn("Test value of variable %s(%s) will not be dumped." % (d['auto_name'], d['name']))
```
abergeron: The code for the warning should check if the flag has the default value with the following ugly code:

`type(config).pickle_test_value.is_default`

theano/gof/graph.py
```diff
@@ -524,6 +526,11 @@ def eval(self, inputs_to_values=None):
...
+            t = d["tag"]
+            del t.test_value
+            d["tag"] = t
```
abergeron: This should make a copy of the tag for the pickle since otherwise we remove thew test_value from the object as well as the pickle.

yikangshen: Hi, I have submitted a fix.

abergeron:
jenkins test this please

theano/gof/graph.py
```diff
@@ -524,6 +526,14 @@ def eval(self, inputs_to_values=None):
     def __getstate__(self):
         d = self.__dict__.copy()
         d.pop("_fn_cache", None)
+        if (not config.pickle_test_value) \
+           and (hasattr(self.tag, 'test_value')) \
+           and (not type(config).pickle_test_value.is_default):
```
nouiz: This is not good. There should be 2 if. If the flag is True, but not the default, we should not print the warning. So if should be something like:
```
if (not config.pickle_test_value and
          hasattr(self.tag, 'test_value')):
    if type(config).pickle_test_value.is_default:
        warnings...
    t = ...
```

yikangshen: Do I need to add another pull request to fix this?

abergeron: This might be a good thing, yes

yikangshen: It's fixed in #6366

PR #6323, opened 2017-08-18
Fix test in the daily buildbot and make sure to don't loose the current linker.

nouiz:
Also allow the flag mode to be an instance of Mode.

### DIFF
--- a/theano/gpuarray/tests/test_opt.py
+++ b/theano/gpuarray/tests/test_opt.py
@@ -1,5 +1,4 @@
 from __future__ import absolute_import, print_function, division
-from nose.plugins.skip import SkipTest
 from nose.tools import assert_raises
 import numpy as np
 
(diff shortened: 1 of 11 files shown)


nouiz:
Note, all the change in this PR except the test split is needed to fix tests in the daily buildbot.

theano/gpuarray/tests/test_opt.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, print_function, division
-import numpy as np
+from nose.plugins.skip import SkipTest
```
abergeron: SkipTest is already imported below.

nouiz:
fixed

nouiz:
I removed the commits that convert the old Pow fix to the new as it was causing a jenkins failure:

http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/682/testReport/junit/theano.gpuarray.tests/test_elemwise/test_elemwise_pow/

PR #6324, opened 2017-08-18
Attempt to lock PR tests to the specific version of libgpuarray that they require.

abergeron:


.jenkins/gpuarray-branch
```diff
@@ -0,0 +1 @@
+v0.6.9
```
nouiz: This force the test to a specific tag. Do we always want that? Or do we want to test with the master of libgpuarray?

abergeron: I think we should lock to a specific tag or even release. Because, in the end Theano is not meant to work with any possible state of the master of libgpuarray, and we have had too much breakage in the past because the master introduced incompatible changes.

The buildbots still test master against master, even though I think they shouldn't either.

nouiz: The master of Theano should work with the master of libgpuarray. An most of the time, the master of Theano should work with the last release of libgpuarray.

So having the daily buildbot run with the master of libgpuarray and the pr buildbot run with a fixed version is great.

merging.

PR #6325, opened 2017-08-18
Create axis as Param Type for Softmax considering the case where axis is an arbitrary scalar

bordesf:
This PR generalize PR #6132 to arbitrary scalar axis.

Here is a todo list associate to this PR (Will be updated later):
- [ ] Test gradients of softmax and logsoftmax given arbitrary axis
- [ ] Test optimization Log(softmax(x)) into logsoftmax_op(x) given arbitrary axis
- [ ] Test optimization Log(softmax(x)[y]) into logsoftmax_op(x)[y] given arbitrary axis
- [ ] Test that Log(Softmax(x))'s grad is replace with the local_logsoftmax_grad given arbitrary axis
- [ ] Test that Log(Softmax(x)[y])'s grad is replace with the local_indexing_logsoftmax_grad optimization given arbitrary axis
- [ ] For all those test, verify the gradients calculation with finite difference methods.
- [ ] Add lift of dimshuffle(softmax(x)) -> softmax(dimshuffle(x)) at the canonicalization step (Idem for reshape). (Need to compute on which axis to compute the softmax after the dimshuffle)

@nouiz @lamblin @ReyhaneAskari


bordesf:
I will correct the flake8 on Monday or Tuesday. I think this PR should be ready soon. I only need to add the lift log SoftmaxGrad to CudnnSoftmaxGrad with the log mode. (I will maybe need some help for that).
We'll still have to discuss what to do with the Abstract softmax. If we drop it, I think it could still be useful to be possible to use the instance mode of cudnn directly (this mode apply the softmax over c01). (Maybe in addition to the axis parameters, we can have a second parameters in softmax function called mode that may the link with the cudnn mode)

PR #6326, opened 2017-08-18
Add some documentation to RNNBlock.

abergeron:
For #5109

### DIFF
--- a/doc/library/gpuarray/dnn.txt
+++ b/doc/library/gpuarray/dnn.txt
@@ -149,6 +149,94 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
 - Spatial Transformer:
     - :func:`theano.gpuarray.dnn.dnn_spatialtf`.
 
+
+cuDNN RNN Example
+=================
+
+This is a code example of using the cuDNN RNN functionality.  We
+present the code with some commentary in between to explain some
+peculiarities.
+
+The terminology here assumes that you are familiar with RNN structure.
+
+.. code-block:: python
+
+    dtype = 'float32'
+    input_dim = 32
+    hidden_dim = 16
+    batch_size = 2
+    depth = 3
+    timesteps = 5
+
+To clarify the rest of the code we define some variables to hold sizes.
+
+.. code-block:: python
+
+    X = T.tensor3('X')
+    Y = T.tensor3('Y')
+    h0 = T.tensor3('h0')
+
+We also define some Theano variables to work with.  Here `X` is input,
+`Y` is output (as in expected output) and `h0` is the initial state
+for the recurrent inputs.
+
+.. code-block:: python
+
+    rnnb = dnn.RNNBlock(dtype, hidden_dim, depth, 'gru')
+
+This defines an RNNBlock.  This is a departure from usual Theano
+operations in that it has the structure of a layer more than a
+separate operation.  This is constrained by the underlying API.
+
+.. code-block:: python
+
+    psize = rnnb.get_param_size([batch_size, input_dim])
+    params_cudnn = gpuarray_shared_constructor(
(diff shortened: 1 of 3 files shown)


theano/gpuarray/dnn.py
```diff
@@ -2501,7 +2501,7 @@ def make_node(self, desc, x, hx, y, reserve, w):
 
 class RNNBlock(object):
     """
-    An object that allow us to use CuDNN v5 RNN implementation.
+    An object that allow us to use CuDNN RNN implementation.
     TODO: make an example how to use. You can check Theano tests
```
nouiz: Can you make an example?

abergeron: Not in the docstring.  It would be about 100 lines long.  I still think that the tests are a sort of good reference.

nouiz:
We need the example to be in the doc. If you mostly copy and adapt the
tests and put it in doc/library/gpuarray/dnn.py, I and many people will be
happy.

doc/library/gpuarray/dnn.txt
```diff
@@ -149,6 +149,94 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
...
+
+Here we allocate space for the trainable parameters of the RNN.  The
+first function tells us how many elements we will need to store the
+parameters.  This space if for all the parameters of all the layers
+inside the RNN and the layout is opaque.
+
+.. code-block:: python
+
+   layer = 0
+    = rnnb.split_params(params_cudnn, layer,
```
nouiz: Here is miss the var name. The next code explain it. But I think it would be more clear if you create a var named something like: splited_params

doc/library/gpuarray/dnn.txt
```diff
@@ -149,6 +149,94 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
...
+                                  [batch_size, input_dim])
+
+If you need to access the parameters individually, you can call
+split_params on your shared variable to get all the parameters for a
+single layer. The order and number of returned items depends on the
+type of RNN.
+
+rnn_relu, rnn_tanh
+  input, recurrent
```
nouiz: and here and bellow, do something like:

input, recurrent = splitted_params.

doc/library/gpuarray/dnn.txt
```diff
@@ -149,6 +149,94 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
...
+
+gru
+  input reset, input update, input newmem, recurrent reset, recurrent
```
nouiz: don't forget to convert some spaces to underscore to make the code work and add missing comma at the end of the line.

doc/library/gpuarray/dnn.txt
```diff
@@ -149,6 +149,94 @@ To get an error if Theano can not use cuDNN, use this Theano flag:
...
+All of these elements are composed of a weights and bias (matrix and
+vector).
+
+.. code-block:: python
+
+    y, hy = rnnb.apply(params_cudnn, X, h0)
+
+This is more akin to an op in Theano in that it will apply the RNN
+operation to a set of symbolic inputs and return symbolic outputs.
+`y` is the output, `hy` is the final state for the recurrent inputs.
```
nouiz: Can we get all the states and not just the final? Add a comment about that case. It happen frequently enough that people want that, so we are better to document it.

abergeron: You can't get more than what is returned.  Although I don't remember if hy contains the intermediate states or not.

nouiz:
cudnn v6 doc tell for y:

Data pointer to GPU memory associated with the output tensor
descriptor yDesc. The data are expected to be packed contiguously
with the first element of iteration n+1 following directly from the
last element of iteration n.

So it is all the step. Can you add that do the doc?

Issue #5515, opened 2017-02-08
Dilated Causal Convolution


nouiz:
This keras issue talk about adding dilated causal convolution. I think this will need new op or a flag in existing op.

https://github.com/fchollet/keras/issues/5297

It would be great to add this to Theano.

basveeling:
Suggestion: perhaps this can be added as a border_mode flag, see: https://github.com/fchollet/keras/pull/5489#issuecomment-282255974

affanv14:
So,for this i need to add a flag for causal and i also need to implement temporal padding along  with a function to compute its gradients?

vikramnitin9:
Is there a logical extension of this concept in 2D? The original WaveNet paper only talks about 1D causal convolutions.
I'm not sure if there is one single way to model sequences in 2D. So would that mean that there is no way to extend causal convolutions to higher dimensions?

benanne:
In 2D it is trickier. The PixelRNN/CNN papers ( https://arxiv.org/abs/1601.06759 https://arxiv.org/abs/1606.05328 ) describe a few possible strategies, but there are many more. In general they will also require masking the filters, rather than changing the padding method, which is less efficient.

PR #6331, opened 2017-08-21
Dilated causal convolution

vikramnitin9:
Related to #5515. I've tried to use a similar approach as in [Keras](https://github.com/fchollet/keras/pull/5489).

This also adds asymmetric padding.

- [x] ~~Dilated Causal wrapper function~~

- [x] ~~Make multiple test cases~~

- [x] ~~Docstrings everywhere~~

- [x] ~~3D case?~~ - NotImplementedError

- [ ] Use cuDNN if the tuple has two same integers. Maybe use cuDNN even otherwise and pre/postprocess the result to simulate asymmetric padding

### DIFF
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -480,11 +480,11 @@ class BaseGpuCorrMM(CGpuKernelBase):
                     'tuple of length 2'.format(border_mode))
             border = ()
             for mode in border_mode:
-                if isinstance(mode, integer_types) and mode >= 0:
-                    border += ((mode, mode),)
-                elif isinstance(mode, tuple) and len(mode) == 2 and \
+                if isinstance(mode, tuple) and len(mode) == 2 and \
                         min(mode) >= 0:
                     border += ((int(mode[0]), int(mode[1])),)
+                elif mode >= 0:
+                    border += ((int(mode), int(mode)),)
                 else:
                     raise ValueError(
                         'invalid border mode {}. The tuple can only contain '
@@ -630,13 +630,13 @@ class BaseGpuCorrMM(CGpuKernelBase):
         if height:
             height = '(*(npy_int*)(PyArray_DATA(%s)))' % height
         else:
-            if ((direction != 0) and (dH != 1)) or ((direction == 1) and (padH_l == -1)):
+            if ((direction != 0) and (dH != 1)) or ((direction == 1) and (padH_l == -1 or padH_r == -1)):
                 raise ValueError("height must be given for backprop with vertical sampling or pad='half'")
             height = '-1'
         if width:
             width = '(*(npy_int*)(PyArray_DATA(%s)))' % width
         else:
-            if ((direction != 0) and (dW != 1)) or ((direction == 1) and (padW_l == -1)):
+            if ((direction != 0) and (dW != 1)) or ((direction == 1) and (padW_l == -1 or padW_r == -1)):
                 raise ValueError("width must be given for backprop with horizontal sampling or pad='half'")
             width = '-1'
 
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -449,8 +449,8 @@ class BaseGpuCorrMM(CGpuKernelBase):
     Parameters
     ----------
     border_mode : {'valid', 'full', 'half'}
-        Additionally, the padding size could be directly specified by an integer
-        or a pair of integers
+        Additionally, the padding size could be directly specified by an integer,
+        a pair of integers, or two pairs of integers.
     subsample
         Perform subsampling of the output (default: (1, 1)).
     filter_dilation
--- a/theano/gpuarray/blas.py
+++ b/theano/gpuarray/blas.py
@@ -468,16 +468,33 @@ class BaseGpuCorrMM(CGpuKernelBase):
     def __init__(self, border_mode="valid", subsample=(1, 1),
                  filter_dilation=(1, 1), num_groups=1, unshared=False):
         if isinstance(border_mode, integer_types):
-            border_mode = (border_mode, border_mode)
-        if isinstance(border_mode, tuple):
-            pad_h, pad_w = map(int, border_mode)
-            border_mode = (pad_h, pad_w)
-        if not ((isinstance(border_mode, tuple) and min(border_mode) >= 0) or
-                border_mode in ('valid', 'full', 'half')):
+            if border_mode < 0:
+                raise ValueError(
+                    'invalid border_mode {}, which must be a '
+                    'non-negative integer'.format(border_mode))
+            border_mode = ((border_mode, border_mode),) * 2
+        elif isinstance(border_mode, tuple):
+            if len(border_mode) != 2:
+                raise ValueError(
+                    'invalid border_mode {} which must be a '
+                    'tuple of length 2'.format(border_mode))
+            border = ()
+            for mode in border_mode:
+                if isinstance(mode, integer_types) and mode >= 0:
+                    border += ((mode, mode),)
+                elif isinstance(mode, tuple) and len(mode) == 2 and \
+                        min(mode) >= 0:
+                    border += ((int(mode[0]), int(mode[1])),)
+                else:
+                    raise ValueError(
+                        'invalid border mode {}. The tuple can only contain '
+                        'integers or tuples of length 2'.format(border_mode))
+            border_mode = border
+        elif border_mode not in ('valid', 'full', 'half'):
             raise ValueError(
                 'invalid border_mode {}, which must be either '
-                '"valid", "full", "half", an integer or a pair of'
-                ' integers'.format(border_mode))
+                '"valid", "full", "half", an integer or a tuple '
+                'of length 2'.format(border_mode))
         self.border_mode = border_mode
         if len(subsample) != 2:
             raise ValueError("subsample must have two elements")
@@ -495,7 +512,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
     def pad(self):
         if self.border_mode != 'valid':
             return self.border_mode
-        return (0, 0)
+        return ((0, 0),) * 2
 
     def __str__(self):
         return '%s{%s, %s, %s, %s, %s}' % (
@@ -537,7 +554,7 @@ class BaseGpuCorrMM(CGpuKernelBase):
 
     def c_code_cache_version(self):
         # Raise this whenever modifying the C code (including the file).
-        return (11,)
+        return (12,)
 
     def c_code_helper(self, bottom, weights, top, direction, sub, height=None, width=None):
         """
@@ -587,14 +604,14 @@ class BaseGpuCorrMM(CGpuKernelBase):
         numgroups = self.num_groups
         unshared = int(self.unshared)
         if self.border_mode == "half":
-            padH = padW = -1
+            padH_l = padH_r = padW_l = padW_r = -1
         elif self.border_mode == "full":
-            padH = padW = -2
+            padH_l = padH_r = padW_l = padW_r = -2
         elif isinstance(self.border_mode, tuple):
-            padH, padW = self.border_mode
+            (padH_l, padH_r), (padW_l, padW_r) = self.border_mode
         else:
             assert self.border_mode == "valid"
-            padH = padW = 0
+            padH_l = padH_r = padW_l = padW_r = 0
         if direction == "forward":
             direction = 0
             out = top
@@ -613,13 +630,13 @@ class BaseGpuCorrMM(CGpuKernelBase):
         if height:
             height = '(*(npy_int*)(PyArray_DATA(%s)))' % height
         else:
-            if ((direction != 0) and (dH != 1)) or ((direction == 1) and (padH == -1)):
+            if ((direction != 0) and (dH != 1)) or ((direction == 1) and (padH_l == -1)):
                 raise ValueError("height must be given for backprop with vertical sampling or pad='half'")
             height = '-1'
         if width:
             width = '(*(npy_int*)(PyArray_DATA(%s)))' % width
         else:
-            if ((direction != 0) and (dW != 1)) or ((direction == 1) and (padW == -1)):
+            if ((direction != 0) and (dW != 1)) or ((direction == 1) and (padW_l == -1)):
                 raise ValueError("width must be given for backprop with horizontal sampling or pad='half'")
             width = '-1'
 
@@ -635,8 +652,10 @@ class BaseGpuCorrMM(CGpuKernelBase):
     size_t dW = %(dW)s;
     size_t dilH = %(dilH)s;
     size_t dilW = %(dilW)s;
-    int padH = %(padH)s;
-    int padW = %(padW)s;
+    int padH_l = %(padH_l)s;
+    int padH_r = %(padH_r)s;
+    int padW_l = %(padW_l)s;
+    int padW_r = %(padW_r)s;
     int numgroups = %(numgroups)s;
     int unshared = %(unshared)s;
 
@@ -662,22 +681,22 @@ class BaseGpuCorrMM(CGpuKernelBase):
             // kernel height is specified (perhaps vertical subsampling or half padding)
             kH = %(height)s;
         }
-        else if (padH == -2) {
+        else if (padH_l == -2 || padH_r == -2) {
             // vertical full padding, we can infer the kernel height
             kH = (2 - PyGpuArray_DIMS(bottom)[2] + (PyGpuArray_DIMS(top)[2] - 1) * dH - 1) / dilH + 1;
         }
         else {
             // explicit padding, we can infer the kernel height
-            kH = (PyGpuArray_DIMS(bottom)[2] + 2*padH - (PyGpuArray_DIMS(top)[2] - 1) * dH - 1) / dilH + 1 ;
+            kH = (PyGpuArray_DIMS(bottom)[2] + padH_l + padH_r - (PyGpuArray_DIMS(top)[2] - 1) * dH - 1) / dilH + 1 ;
         }
         if (%(width)s != -1) {
             kW = %(width)s;
         }
-        else if (padW == -2) {
+        else if (padW_l == -2 || padW_r == -2) {
             kW = (2 - PyGpuArray_DIMS(bottom)[3] + (PyGpuArray_DIMS(top)[3] - 1) * dW - 1) / dilW + 1;
         }
         else {
-            kW = (PyGpuArray_DIMS(bottom)[3] + 2*padW - (PyGpuArray_DIMS(top)[3] - 1) * dW - 1) / dilW + 1;
+            kW = (PyGpuArray_DIMS(bottom)[3] + padW_l + padW_r - (PyGpuArray_DIMS(top)[3] - 1) * dW - 1) / dilW + 1;
         }
     }
 
@@ -686,23 +705,23 @@ class BaseGpuCorrMM(CGpuKernelBase):
     dil_kW = (kW - 1) * dilW + 1;
 
     // Auto-padding if requested
-    if (padH == -1) {  // vertical half padding
-        padH = dil_kH / 2;
+    if (padH_l == -1 || padH_r == -1) {  // vertical half padding
+        padH_l = padH_r = dil_kH / 2;
     }
-    else if (padH == -2) {  // vertical full padding
-        padH = dil_kH - 1;
+    else if (padH_l == -2 || padH_r == -2) {  // vertical full padding
+        padH_l = padH_r = dil_kH - 1;
     }
-    else if (padH < 0) {
+    else if (padH_l < 0 || padH_r < 0) {
         PyErr_SetString(PyExc_ValueError, "BaseGpuCorrMM: padH must be >= -2");
         %(fail)s
     }
-    if (padW == -1) {  // horizontal half padding
-        padW = dil_kW / 2;
+    if (padW_l == -1 || padW_r == -1) {  // horizontal half padding
+        padW_l = padW_r = dil_kW / 2;
     }
-    else if (padW == -2) {  // horizontal full padding
-        padW = dil_kW - 1;
+    else if (padW_l == -2 || padW_r == -2) {  // horizontal full padding
+        padW_l = padW_r = dil_kW - 1;
     }
-    else if (padW < 0) {
+    else if (padW_l < 0 || padW_r < 0) {
         PyErr_SetString(PyExc_ValueError, "BaseGpuCorrMM: padW must be >= -2");
         %(fail)s
     }
@@ -718,11 +737,11 @@ class BaseGpuCorrMM(CGpuKernelBase):
     switch(direction) {
     case 0:  // forward pass
         // output is top: (batchsize, num_filters, height, width)
-        // height and width: top = (bottom + 2*pad - ((weight-1)*dil + 1)) / sample + 1
+        // height and width: top = (bottom + pad_l + pad_r - ((weight-1)*dil + 1)) / sample + 1
         out_dim[0] = PyGpuArray_DIMS(bottom)[0];
         out_dim[1] = PyGpuArray_DIMS(weights)[0];
-        out_dim[2] = (PyGpuArray_DIMS(bottom)[2] + 2*padH - ((PyGpuArray_DIMS(weights)[wdim-2]-1)*dilH + 1)) / dH + 1;
-        out_dim[3] = (PyGpuArray_DIMS(bottom)[3] + 2*padW - ((PyGpuArray_DIMS(weights)[wdim-1]-1)*dilW + 1)) / dW + 1;
+        out_dim[2] = (PyGpuArray_DIMS(bottom)[2] + padH_l + padH_r - ((PyGpuArray_DIMS(weights)[wdim-2]-1)*dilH + 1)) / dH + 1;
+        out_dim[3] = (PyGpuArray_DIMS(bottom)[3] + padW_l + padW_r - ((PyGpuArray_DIMS(weights)[wdim-1]-1)*dilW + 1)) / dW + 1;
         out_typecode = bottom->ga.typecode;
         out_context = bottom->context;
         if (out_dim[0] < 0 || out_dim[1] < 0 || out_dim[2] <= 0 || out_dim[3] <= 0)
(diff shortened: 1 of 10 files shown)


vikramnitin9:
I think the Python code is okay-ish now. @gvtulder

nouiz:
closing/reopening to trigger travis

vikramnitin9:
The Travis build will fail right now because of errors carried over from unshared convolution (this branched off early, before the later commits that fixed the errors). I'll need to rebase, but I was planning to do that once the unshared convolution was merged.

gvtulder:
Hi @vikramnitin9 (and @nouiz and others if you'd like to give your views). I haven't looked at all of the code in detail yet, but one thing that caught my eye: forcing 1D inputs into a 2D convolution op gets a bit awkward. Things like this don't look good, I think:
```python
if any(i != 1 for i in img.shape[1 - self.convdim:]):
    raise ValueError('All but the first spatial dimension of the '
                     'input must be 1 for dilated causal convolution.')
```
Implementing causal convolution for only one dimension seems fine and useful, but I think it would be nice to do that in a way that's a bit more elegant than this. I'm not sure how we could do that, though:

* Having 1D convolution ops would be a good way to have 1D-only causal convolution (even if they are abstract 1D ops only that map to 2D ops for the implementation). 1D conv ops would be a project on itself, so probably not something we want to try here.
* I don't know if there is a straightforward way to extend causal convolution to 2D. In your project proposal you wrote that there isn't one obvious way, so perhaps we shouldn't aim to do that and stick to 1D-only causal convolution.
* Could we maybe let users specify which dimension they want for the causality? So that instead of `border_mode='causal'` you would have something like `border_mode=('causal', 'valid')`?

I think it's fine to implement dilated causal convolution on one dimension of 2D convolution for now, but I'd prefer a nicer interface. Rather than having `border_mode='causal'` on 2D convolution mean 1D causal dilated convolution on the first dimension, I would prefer something that would make sense in a 2D context.

What are your thoughts on this? Am I making things more complicated than necessary?

nouiz:
Why not make a function conv1d_causal or something like that?

vikramnitin9:
According to the [Wavenet paper](https://arxiv.org/pdf/1609.03499.pdf) (bottom of Page 2), he mentions that the way to extend causal convolution to images would be [Pixel RNN](https://arxiv.org/pdf/1601.06759.pdf). But there's also something like [this](https://arxiv.org/pdf/1606.05328.pdf), and there may be even more. Also, these will force us to use some sort of masking instead of simple padding.

@gvtulder, I agree. I was going to do a wrapper function that would build the `AbstractConv` Op with the correct combination of parameters. But an even simpler method might be to have _that_ function pad with zeros and create an `AbstractConv` with `border_mode=valid`.

gvtulder:
@vikramnitin9 How are you planning to implement this in the C ops? It looks like it's mostly just asymmetric padding. I think it could be useful to implement it mostly in that way, rather than with temporal-padding specific code.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2314,29 +2554,62 @@ def perform(self, node, inp, out_):
...
+                img = img[flip_filters]
+
         if any(p > 0 for p in pad):
             img = img[(slice(None), slice(None)) +
                       tuple(slice(pad[i], img.shape[i + 2] - pad[i])
                             for i in range(self.convdim))]
+        if mode == 'causal':
+            left_pad = self.filter_dilation[0] * (kern.shape[-self.convdim] - 1)
+            img = img[(slice(None), slice(None),
+                      slice(left_pad, left_pad + img.shape[2]))]
```
gvtulder: For example, these lines could be merged with the normal padding logic if you split `pad` into `pad_left` and `pad_right`, which would be equal for normal padding and different for causal. (My argument for this would be that that it doesn't make it harder to implement causal convolution now, but would make it easier to add support for asymmetric padding at some point.)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2026,26 +2104,42 @@ def perform(self, node, inp, out_):
...
 
         if mode == "full":
-            mode = tuple(dil_kernshp[i] - 1 for i in range(self.convdim))
+            mode = tuple((dil_kernshp[i] - 1,) * 2 for i in range(self.convdim))
         elif mode == "half":
-            mode = tuple(dil_kernshp[i] // 2 for i in range(self.convdim))
+            mode = tuple((dil_kernshp[i] // 2,) * 2 for i in range(self.convdim))
         if isinstance(mode, tuple):
-            pad = tuple(int(mode[i]) for i in range(self.convdim))
+            pad = mode
```
gvtulder: Maybe start filling `pad` in the lines above?

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2026,26 +2104,42 @@ def perform(self, node, inp, out_):
...
-                ' integers'.format(mode))
+                '"valid", "full", "half", an integer or a tuple '
+                'of length {}'.format(mode, self.convdim))
 
         if mode == "full":
-            mode = tuple(dil_kernshp[i] - 1 for i in range(self.convdim))
+            mode = tuple((dil_kernshp[i] - 1,) * 2 for i in range(self.convdim))
         elif mode == "half":
-            mode = tuple(dil_kernshp[i] // 2 for i in range(self.convdim))
+            mode = tuple((dil_kernshp[i] // 2,) * 2 for i in range(self.convdim))
```
gvtulder: I think you can combine this with the previous block of ifs. Before your change there was a clear separation: first a check, then a conversion. Now there is some conversion, then a check, then some more conversion.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2311,19 +2421,18 @@ def perform(self, node, inp, out_):
                           for i in range(self.convdim))
 
         if mode == "full":
-            mode = tuple(dil_shape[i] - 1 for i in range(self.convdim))
+            mode = tuple((dil_shape[i] - 1,) * 2 for i in range(self.convdim))
         elif mode == "half":
-            mode = tuple(dil_shape[i] // 2 for i in range(self.convdim))
+            mode = tuple((dil_shape[i] // 2,) * 2 for i in range(self.convdim))
```
gvtulder: Idem: I think this can be combined with the previous if. (And maybe extracted to a shared helper function.)

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2612,12 +2721,28 @@ def perform(self, node, inp, out_):
...
+                        'invalid border mode {}. The tuple can only contain '
+                        'integers or tuples of length 2'.format(mode))
+            mode = border
+        elif mode not in ('valid', 'full', 'half'):
             raise ValueError(
                 'invalid border_mode {}, which must be either '
-                '"valid", "full", "half", an integer or a tuple of'
-                ' integers'.format(mode))
+                '"valid", "full", "half", an integer or a tuple '
+                'of length {}'.format(mode, self.convdim))
```
gvtulder: This if can be moved a few lines down, and then it looks very similar to that of the other two `perform` functions, so it can be extracted into a shared helper function that computes `pad` given `border_mode`.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2311,19 +2421,18 @@ def perform(self, node, inp, out_):
                           for i in range(self.convdim))
 
         if mode == "full":
-            mode = tuple(dil_shape[i] - 1 for i in range(self.convdim))
+            mode = tuple((dil_shape[i] - 1,) * 2 for i in range(self.convdim))
         elif mode == "half":
-            mode = tuple(dil_shape[i] // 2 for i in range(self.convdim))
+            mode = tuple((dil_shape[i] // 2,) * 2 for i in range(self.convdim))
         if isinstance(mode, tuple):
```
gvtulder: The check in the `perform` for gradInputs is nicer. This just checks if mode might be equal to `valid`, whereas the check you do in that other `perform` below also checks if `mode` is a tuple of all zeros.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2705,9 +2830,9 @@ def correct_for_groups(mat):
             if self.filter_flip:
                 img = img[flip_filters]
 
-        if any(p > 0 for p in pad):
+        if any(p != (0, 0) or p != 0 for p in pad):
```
gvtulder: This is a nice check that you could also do in the other `perform`s (instead of `if isinstance(pad, tuple)`).

theano/tensor/nnet/corr.py
```diff
@@ -66,20 +67,29 @@ def __init__(self, border_mode="valid", subsample=(1, 1),
...
-                border_mode in ('valid', 'full', 'half')):
+                    'invalid border_mode {} which must be a '
+                    'tuple of length 2'.format(border_mode))
+            border = ()
+            for mode in border_mode:
+                if isinstance(mode, integer_types) and mode >= 0:
+                    border += ((mode, mode),)
+                elif isinstance(mode, tuple) and len(mode) == 2 and \
+                        min(mode) >= 0:
+                    border += ((int(mode[0]), int(mode[1])),)
```
gvtulder: I'm not really sure why those casts to `int` were there in the original version (`map(int, border_mode)`). But if you want to keep it backwards compatible, your new version is changing the way that non-integers are handled. It would be better to swap the `if`
```python
if isinstance(mode, tuple):
    # ...
else:
    # mode must be a number, but has to be casted to an int
```

theano/tensor/nnet/corr.py
```diff
@@ -271,13 +283,13 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
         if height:
             height = '(*(npy_int64 *)(PyArray_DATA(%s)))' % height
         else:
-            if ((self.direction != 0) and (self.dH != 1)) or ((self.direction == 1) and (self.padH == -1)):
+            if ((self.direction != 0) and (self.dH != 1)) or ((self.direction == 1) and (self.padH_l == -1)):
```
gvtulder: Maybe check for `self.padH_r == -1` as well (you do that in all the instances below).

theano/tensor/nnet/corr.py
```diff
@@ -271,13 +283,13 @@ def c_code_helper(self, bottom, weights, top, sub, height=None, width=None):
...
                 raise ValueError("height must be given for backprop with vertical sampling or border_mode='half'")
             height = '-1'
         if width:
             width = '(*(npy_int64 *)(PyArray_DATA(%s)))' % width
         else:
-            if ((self.direction != 0) and (self.dW != 1)) or ((self.direction == 1) and (self.padW == -1)):
+            if ((self.direction != 0) and (self.dW != 1)) or ((self.direction == 1) and (self.padW_l == -1)):
```
gvtulder: Idem.

theano/tensor/nnet/corr.py
```diff
@@ -706,36 +721,36 @@ def make_node(self, img, topgrad, shape=None):
...
     def infer_shape(self, node, input_shape):
         if self.border_mode == "half":
-            padH = padW = -1
+            padH_l = padH_r = padW_l = padW_r = -1
         elif self.border_mode == "full":
-            padH = padW = -2
+            padH_l = padH_r = padW_l = padW_r = -2
         elif isinstance(self.border_mode, tuple):
-            padH, padW = self.border_mode
+            (padH_l, padH_r), (padW_l, padW_r) = self.border_mode
```
gvtulder: This breaks backward compatibility, since `self.border_mode` for older, pickled ops might still be `(int, int)`.

gvtulder: And why isn't this using `get_conv_gradweights_shape` anyway?

theano/tensor/nnet/corr.py
```diff
@@ -820,42 +835,42 @@ def make_node(self, kern, topgrad, shape=None):
...
```
gvtulder: As for gradweights: this isn't backwards compatible.

theano/tensor/nnet/corr.py
```diff
@@ -820,42 +835,42 @@ def make_node(self, kern, topgrad, shape=None):
 
     def infer_shape(self, node, input_shape):
         if self.border_mode == "half":
```
gvtulder: Do you have any idea why this isn't using `get_conv_gradinputs_shape`? `get_conv_gradinputs_shape` doesn't work with `subsample`, but then why can we do it here? (Sort of a rhetoric question for another time, I suppose.)

gvtulder:
There's a minor drawback to implementing causal convolutions as asymmetric padding that I hadn't realised before: the convolution ops don't accept symbolic padding, so you're forced to pass a constant `filter_shape` to the dilated causal convolution function. That's a slightly unfriendly interface, so perhaps it would be nice to see if we can fix that in the future (but not in this PR).

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1484,6 +1505,54 @@ def conv3d_grad_wrt_weights(input,
...
+    filter_dilation = (filter_dilation, 1)
+
+    conv_op = AbstractConv2d(imshp=input_shape,
+                             kshp=filter_shape,
+                             border_mode=((left_pad, 0), 0),
+                             subsample=subsample,
+                             filter_flip=filter_flip,
+                             filter_dilation=filter_dilation,
+                             num_groups=num_groups,
+                             unshared=False)
```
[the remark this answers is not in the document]
vikramnitin9: Hmm yes. I'll change that

vikramnitin9:
I did a rebase followed by cherry-picking the old commits in order. Hopefully the commit history is neat now.

I have yet to implement about half the suggestions.

vikramnitin9:
What do we do about the 3D case? Should I raise a NotImplementedError?

gvtulder:
A NotImplementedError for 3D would be fine.

theano/tensor/nnet/abstract_conv.py
vikramnitin9: I'm not sure about this - won't that involve multiple assignments in each case instead of just one line (pad = mode)?

theano/tensor/nnet/corr.py
vikramnitin9: Can't it also be something else in that case, say a string or a list?

vikramnitin9: But I see that the original code doesn't account for that either. Okay then

theano/tensor/nnet/corr.py
vikramnitin9: I thought this is fine since we normalize the border_mode in the BaseCorrMM.__init__ function. But I'll make the change.

vikramnitin9:
I've implemented the suggestions. I've used a helper function mode_to_pad

vikramnitin9:
Ready for another review.

vikramnitin9:
Travis stalled for some reason. https://travis-ci.org/Theano/Theano/builds/268787751?utm_source=github_status&utm_medium=notification

theano/tensor/nnet/corr.py
gvtulder: That's true, `BaseCorrMM__init__` covers most of the cases, but not the case where a graph that was pickled with the old version is unpickled in the new version. Unpickling doesn't run __init__ again, it directly restores the internal state of the object.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -2629,27 +2815,16 @@ def perform(self, node, inp, out_):
                  for i in range(2 + self.convdim)]
         expected_topgrad_shape = get_conv_output_shape(
             imshp, kern.shape,
-            self.border_mode, self.subsample, self.filter_dilation)
+            mode, self.subsample, self.filter_dilation)
```
gvtulder: I think you can just use `self.border_mode` here and a few lines above this.

theano/tensor/nnet/corr.py
```diff
@@ -66,20 +67,29 @@ def __init__(self, border_mode="valid", subsample=(1, 1),
...
+                        'invalid border mode {}. The tuple can only contain '
+                        'integers or tuples of length 2'.format(border_mode))
+            border_mode = border
+        elif border_mode not in ('valid', 'full', 'half'):
             raise ValueError(
                 'invalid border_mode {}, which must be either '
-                '"valid", "full", "half", an integer or a pair of'
-                ' integers'.format(border_mode))
+                '"valid", "full", "half", an integer or a tuple '
+                'of length 2'.format(border_mode))
```
gvtulder: a tuple of two integers or pairs of integers

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1484,6 +1584,102 @@ def conv3d_grad_wrt_weights(input,
     return gradWeight_op(input, output_grad, filter_shape[-3:])
 
 
+def causal_conv(input,
```
gvtulder: I think it's better to rename this to `causal_conv1d`, for consistency with the other methods.

vikramnitin9:
Are the causal tests fine? Also I realised that I had added `.excluding('cudnn')`. So I removed it. Can't test on my laptop though because OpenCL.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -52,11 +52,11 @@ def get_conv_output_shape(image_shape, kernel_shape,
         number of output channels, height and width of the output, number of
         input channels, height and width of the kernel.
         None where undefined.
-    border_mode: string, int (symbolic or numeric) or tuple of int (symbolic
-        or numeric). If it is a string, it must be 'valid', 'half' or 'full'.
-        If it is a tuple, its two (or three) elements respectively correspond
-        to the padding (possibly left and right) on height and width
-        (and possibly depth) axis.
+     border_mode: string, int (symbolic or numeric) or tuple of int (symbolic
```
gvtulder: This space doesn't line up.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -173,11 +174,11 @@ def get_conv_gradweights_shape(image_shape, top_shape,
         image shape. Its four (or five) element must correspond respectively
         to: batch size, number of output channels, height and width (and
         possibly depth) of the image. None where undefined.
-    border_mode: string, int (symbolic or numeric) or tuple of int (symbolic
-        or numeric). If it is a string, it must be 'valid', 'half' or 'full'.
-        If it is a tuple, its two (or three) elements respectively correspond
-        to the padding (possibly left and right) on height and width
-        (and possibly depth) axis.
+     border_mode: string, int (symbolic or numeric) or tuple of int (symbolic
```
gvtulder: Idem.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -296,11 +298,11 @@ def get_conv_gradinputs_shape(kernel_shape, top_shape,
...
```
gvtulder: Idem.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -423,11 +426,11 @@ def check_conv_gradinputs_shape(image_shape, kernel_shape, output_shape,
         output shape. Its four (or five) elements must correspond respectively
         to: batch size, number of output channels, height and width
         (and possibly depth) of the output. None where undefined.
-    border_mode: string, int (symbolic or numeric) or tuple where each element
-        is either an int or a tuple of length 2 (symbolic or numeric).
-        If it is a string, it must be 'valid', 'half' or 'full'.
-        If it is a tuple, its two (or three) elements respectively correspond
-        to the padding on height and width (and possibly depth) axis.
+     border_mode: string, int (symbolic or numeric) or tuple of int (symbolic
```
gvtulder: Idem.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1885,8 +1893,7 @@ class BaseAbstractConv(Op):
         element is not known at compile time.
         kshp is defined w.r.t the forward conv.
 
-     border_mode: str, int or tuple of ``convdim`` elements where each element
-        is an integer or a tuple of length 2.
+    border_mode: str, int or a tuple of two ints or pairs of ints
```
gvtulder: Idem.

gvtulder:
I think the point of the causal tests is mostly to check if the right op is constructed, since the asymmetric padding is already tested elsewhere. So this test seems sufficient. Is there something that you think is missing and should be tested as well?

Eventually, if/when asymmetric padding is supported in the cudnn ops, the `.excluding('cudnn')` in `test_gemcorr.py` should be added back. I guess it's fine to leave it out for now.

gvtulder:
I just re-ran `theano/gpuarray/tests/test_gemmcorr.py`, without `excluding('cudnn')` in the asymmetric tests, and the tests still pass.

vikramnitin9:
> Is there something that you think is missing and should be tested as well?

I was just wondering if we should test using the definition of causality, ie, whether varying future inputs affects current output. But I guess it's fine as it stands as well.

gvtulder:
@nouiz Could you run Jenkins on this please?

nouiz:
jenkins test this

nouiz:
@gvtulder is this PR ready for you? Should I do the final review?

theano/gpuarray/blas.py
```diff
@@ -449,8 +449,8 @@ class BaseGpuCorrMM(CGpuKernelBase):
     Parameters
     ----------
     border_mode : {'valid', 'full', 'half'}
-        Additionally, the padding size could be directly specified by an integer
-        or a pair of integers
+        Additionally, the padding size could be directly specified by an integer,
+        a pair of integers, or two pairs of integers.
```
gvtulder: The GpuCorrMM doc should probably also mention this.

theano/tensor/nnet/corr.py
```diff
@@ -34,8 +34,8 @@ class BaseCorrMM(gof.OpenMPOp):
     Parameters
     ----------
     border_mode : {'valid', 'full', 'half'}
-        Additionally, the padding size could be directly specified by an integer
-        or a pair of integers
+        Additionally, the padding size could be directly specified by an integer,
+        a pair of integers, or two pairs of integers.
```
gvtulder: Also make a change in the `border_mode` doc for CorrMM.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1484,6 +1505,54 @@ def conv3d_grad_wrt_weights(input,
...
+    filters = filters.dimshuffle(0, 1, 2, 'x')
+
+    if input_shape is not None:
+        assert(len(input_shape) == 3)
+        input_shape = tuple(input_shape)
+        input_shape += (1,)
+
+    assert(len(filter_shape) == 3)
+    filter_shape = tuple(filter_shape)
+    filter_shape += (1,)
```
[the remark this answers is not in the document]
nouiz: I think it would be easy to add support for filter_shape to be simbolic. Just something like:
if isinstance(filter_shape, tuple):
   filter_shape = ...
else:
   filter_shape = theano.tensor.concatenate([[1,], v.shape]).type

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1751,21 +1954,31 @@ def __init__(self, convdim,
...
+        elif border_mode not in ('valid', 'full', 'half'):
             raise ValueError(
                 'invalid border_mode {}, which must be either '
-                '"valid", "full", "half", an integer or a tuple of {}'
-                ' integers'.format(border_mode, convdim))
+                '"valid", "full", "half", an integer or a tuple '
+                'of length {}'.format(border_mode, convdim))
+        if isinstance(border_mode, tuple) and \
+                all(mode == (0, 0) or mode == 0 for mode in border_mode):
+            border_mode = 'valid'
```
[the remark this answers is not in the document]
gvtulder: I think this is done in line 1975 to 1977.

theano/tensor/nnet/abstract_conv.py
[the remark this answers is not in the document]
gvtulder: But that wouldn't work with any of the convolution ops, right? The padding/border mode is passed to `__init__` and if I understood things correctly you can't use Theano variables in that way? I thought they had to be inputs for `make_node` to be able to do that.

My impression was that adding support for symbolic padding would require a fair number of changes. The padding would need to be converted from an `__init__` parameter to an input for each of the convolution ops. But then border mode strings such as `valid`, `full` etc. wouldn't fit inside a Theano variable and those would have to stay in `__init__`, so perhaps you'd need a new `border_mode="custom"` to indicate when the symbolic padding should be used.

theano/tensor/nnet/abstract_conv.py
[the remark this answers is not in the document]
nouiz: right

theano/tensor/nnet/abstract_conv.py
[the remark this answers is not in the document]
nouiz: right. It is more complicated then my simple line.

theano/gpuarray/dnn.py
```diff
@@ -3038,6 +3038,9 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
     if op.unshared:
         return None
 
+    if isinstance(op.border_mode, tuple) and any(isinstance(p, tuple) for p in op.border_mode):
```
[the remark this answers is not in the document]
vikramnitin9: Sorry, I didn't understand this. Could you explain in a little more detail? Which manual padding?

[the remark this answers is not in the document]
gvtulder: I think that refers to the second part of my comment, where I suggested to use cuDNN with asymmetric padding. Depending on the padding values, you'd have to manually pad the input before you give it to the cuDNN op or manually crop the result of the output. That would work, but it might indeed use more memory and might also be slower than GpuCorrMM with custom padding.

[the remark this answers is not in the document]
vikramnitin9: Got it, thanks

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1751,21 +1961,34 @@ def __init__(self, convdim,
...
+            for mode in border_mode:
+                if isinstance(mode, tuple) and convdim != 2:
+                    raise NotImplementedError(
+                        'Asymmetric padding not implemented for {}D'.format(convdim))
+                if not((isinstance(mode, integer_types) and mode >= 0) or
+                        (isinstance(mode, tuple) and len(mode) == 2 and min(mode) >= 0 and
+                         all(isinstance(m, integer_types) for m in mode))):
+                    raise ValueError(
+                        'invalid border mode {}. The tuple can only contain integers '
+                        ' or pairs of integers'.format(border_mode))
```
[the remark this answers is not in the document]
vikramnitin9: Since this is a type of optimization, how about I do this in dnn.py, in the places where we currently return None?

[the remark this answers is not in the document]
nouiz: If you do it in dnn.py, you will need to do it in many optimization. This would duplicate much code. You could do it in dnn.py, but in the \__init__ method. This would work and won't duplicate code. But I prefer to do it at the earliest. I see this as a simplication. Maybe later other places would need that. So doing it very early would make sure they work directly without needing to handle that case.

But if you do it in the \__init__.py in dnn.py, I would also accept that. Just not as much future proof.

gvtulder:
jenkins test this

nouiz:
I'm ok with the PR. But I prefer to wait to merge it after the beta2 is out. There is enough stuff going on to stabilize that I prefer to post-pone it. beta2 should have been out 1 week ago already. I don't want to risk postponing it again.

nouiz:
we will do a beta3, so I can merge this PR now.

thanks.

PR #6332, opened 2017-08-21
test optimization in grouped convolution

affanv14:
There seems to be an optimization error for cudnn. Thanks @gvtulder  for pointing that out

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3054,7 +3054,7 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
                         conv_mode=conv_mode,
                         num_groups=op.num_groups)
     elif isinstance(op, AbstractConv2d_gradWeights):
-        shape = (inp2.shape[1], inp1.shape[1],
+        shape = (inp2.shape[1], inp1.shape[1] // op.num_groups,
                  inputs[2][0], inputs[2][1])
         rval = dnn_gradweight(inp1, inp2, shape,
                               border_mode=op.border_mode,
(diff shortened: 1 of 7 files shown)


nouiz:
jenkins test this

nouiz:
jenkins test this

nouiz:
jenkins failed:

http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/695/testReport/junit/theano.tests/test_flake8/test_format_flake8/

affanv14:
I finished making all the changes

nouiz:
jenkins test this

theano/gpuarray/dnn.py
```diff
@@ -3054,7 +3054,7 @@ def local_abstractconv_cudnn_graph(op, context_name, inputs, outputs):
                         conv_mode=conv_mode,
                         num_groups=op.num_groups)
     elif isinstance(op, AbstractConv2d_gradWeights):
-        shape = (inp2.shape[1], inp1.shape[1],
+        shape = (inp2.shape[1], inp1.shape[1] // op.num_groups,
```
nouiz: Can you confirm that the consequence of those bug where crashes?

affanv14: yes, this leads to a crash in the cudnn version

affanv14: Sorry I missed this

nouiz: It happen to everybody. I like to make the distinction between a few catétory of bug depending of how it affect user. Mostly, is it a user visible bug or not.

If it is a crash, it is not the worst bug, as user see it clearly, it is not too bad.

The worst bug are if we return wrong result for some reasons.

PR #6333, opened 2017-08-21
Restrict cuDNN reduction to contiguous inputs

abergeron:
There is a problem that was reported with cuDNN reduction where it would give different results than before it was integrated.

This *may* be because there are corner cases where non-contiguous inputs are not supported, so we are tentatively reversing this.  More testing is required to determine the exact cause.


abergeron:
jenkins test this please

nouiz:
jenkins have one test case failure. The test need to be updated.

nouiz:
jenkins failed.

nouiz:
I canceled the end of travis.

PR #6336, opened 2017-08-22
Add more tests for meta-optimizer

affanv14:


affanv14:
for GPUCorrMM , I need to add subsample or  [this](https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1625-L1652) code will take effect and change it to GPUCorrMM_gradWeights.

affanv14:
would it be a good idea to remove https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1625-L1652
and https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1599-L1608 and leave it to the meta-optimizer which performs the same? @nouiz

nouiz:
This optimization seem to do the same logic as dnn_conv2d related to the direction_hint parameter. But it do not allow us to enable it or not.

When the meta optimizer is used and try those 3 implementation, I think we should make that optimizer not force this selection. One simple way to fix that would to implement a new optimization that don't does that and have the meta optimizer user that optimization instead of the current one.

nouiz:
jenkins test this

PR #6341, opened 2017-08-22
Try to fix segmentation fault related to last updates in cuDNN conv codes.

notoraptor:
@nouiz @abergeron

### DIFF
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -235,7 +235,6 @@ def struct_gen(args, struct_builders, blocks, sub):
 
     # declares the storage
     storage_decl = "\n".join(["PyObject* %s;" % arg for arg in args])
-    storage_init = "\n".join(["this->%s = NULL;" % arg for arg in args])
     # in the constructor, sets the storage to the arguments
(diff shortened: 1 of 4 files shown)


theano/gof/cc.py
```diff
@@ -235,6 +235,7 @@ def struct_gen(args, struct_builders, blocks, sub):
 
     # declares the storage
     storage_decl = "\n".join(["PyObject* %s;" % arg for arg in args])
+    storage_init = "\n".join(["this->%s = NULL;" % arg for arg in args])
```
abergeron: We don't need this, there is no problem with the storage.

nouiz: Does this remove the need for the memset? I don't think so it enough to remove the memset, as it won't fix other case we cover, but it still greatly help. Op that don't have there own struct_decl (they just use params) won't need the memset.

nouiz: Forget my comments, it is for the storage, not for the param. See @abergeron comments. I'll delete mine.

notoraptor: Indeed, removed.

nouiz:
2 person told me it fix there problem, so we can merge when jenkins CI pass. We don't need to wait for travis CI.

PR #6349, opened 2017-08-23
Fix complilation on mac.  It claims c++99 but has unordered_map, so...

abergeron:


abergeron:
I did test it locally, but we should probably try this too:

jenkins test mac

nouiz:
http://earlgrey.iro.umontreal.ca:8080/job/Theano_pr_mac/20/

jenkins test mac

nouiz:
jenkins test mac

Issue #6337, opened 2017-08-22
MRG distribution, add a target parameter.


nouiz:
see https://github.com/Theano/Theano/pull/6274 that start that.

nouiz:
done by https://github.com/Theano/Theano/pull/6355

PR #6355, opened 2017-08-24
MRG distribution, add a target parameter

yikangshen:
Allow manual work around #6337

### DIFF
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -866,8 +866,10 @@ class MRG_RandomStreams(object):
         rstates = self.get_substream_rstates(nstreams, dtype)
 
         d = {}
-        if kwargs.has_key('target'):
+        if 'target' in kwargs:
             d = dict(target=kwargs.pop('target'))
+        if len(kwargs) > 0:
+            raise TypeError("uniform() got unexpected keyword arguements %s" % (str(kwargs.keys())))
         node_rstate = shared(rstates, **d)
         u = self.pretty_return(node_rstate,
                                *mrg_uniform.new(node_rstate,
@@ -883,8 +885,6 @@ class MRG_RandomStreams(object):
                 '`low` and `high` arguments')
 
         assert r.dtype == dtype
-        if len(kwargs) > 0:
-            raise TypeError("uniform() got unexpected keyword arguements %s" % (str(kwargs.keys())))
         return r
 
     def binomial(self, size=None, n=1, p=0.5, ndim=None, dtype='int64',
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -805,7 +805,7 @@ class MRG_RandomStreams(object):
         return sample
 
     def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
-                nstreams=None):
+                nstreams=None, **kwargs):
         # TODO : need description for parameter 'size', 'ndim', 'nstreams'
         """
         Sample a tensor of given size whose element from a uniform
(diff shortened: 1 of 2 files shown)


nouiz:
travis failure: https://travis-ci.org/Theano/Theano/builds/268041569?utm_source=github_status&utm_medium=notification

probably flake8

theano/sandbox/rng_mrg.py
```diff
@@ -865,7 +865,10 @@ def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
             nstreams = self.n_streams(size)
         rstates = self.get_substream_rstates(nstreams, dtype)
 
-        node_rstate = shared(rstates)
+        d = {}
```
nouiz: This would need some tests. Easy tests would be to just use the random generator with target=cpu for all distribution. This would at least make sure it don't crash.

theano/sandbox/rng_mrg.py
```diff
@@ -865,7 +865,10 @@ def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
...
+        if kwargs.has_key('target'):
+            d = dict(target=kwargs.pop('target'))
```
nouiz: THis can be simplified by:

    d = kwargs.pop('target', {})

the last argument in the default value if target isn't in the dic.t

theano/sandbox/rng_mrg.py
yikangshen: I tested with following code:
```
print rng.uniform((3,2), target='cpu')
print rng.binomial((3,2), target='cpu')
print rng.multinomial(pvals=numpy.array([[.98, .01, .01], [.01, .49, .50]]).astype('float32'), target='cpu')
print rng.choice(p=numpy.array([[.98, .01, .01], [.01, .49, .50]]).astype('float32'), replace=False, target='cpu')
print rng.multinomial_wo_replacement(pvals=numpy.array([[.98, .01, .01], [.01, .49, .50]]).astype('float32'), target='cpu')
```
no crash happened. What is a flake8 error?

theano/sandbox/rng_mrg.py
yikangshen: If kwargs has 'target', pop() will only return the value. But d should be a dict like {'targat':value}.

theano/sandbox/rng_mrg.py
nouiz: The tests should go in the one test file. For this file, the tests are in theano/sandbox/tests/tests_rng_mrg.py

Don't use print in the tests. Add a comment that you just test if it don't crash.

flake8 is a python code syntax standard. Theano enforce most of it in some tests to help make the code more readable and more consistent.

On each PR, Theano have CI (Continuous Integration) system that will run our tests automatically. In the PR, next to the commit, there if an red "x" that tell some of our tests failed. Click on it. You will see the errors.

theano/sandbox/rng_mrg.py
nouiz: Right. What I had in mind won't work, so there is just the small changes from my other comment.

thanks

theano/sandbox/rng_mrg.py
```diff
@@ -880,20 +883,22 @@ def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
                 '`low` and `high` arguments')
 
         assert r.dtype == dtype
+        if len(kwargs) > 0:
```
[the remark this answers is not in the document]
nouiz: Don't forget to move that error check earlier.

theano/sandbox/rng_mrg.py
yikangshen: A fix is submitted

PR #6356, opened 2017-08-24
Automatically add dnn.bin_path to PATH at runtime on Windows

### LINKED ISSUE #6330: Could it be acceptable to extend PATH at runtime with required folders ?
For example, on Windows, we always need to add cuDNN bin path to system `PATH` environment variable so that `cudnn*.dll` can be found to run cuDNN-related Theano ops.

What about directly add the required folder (here `theano.config.dnn.bin_path`, if defined) at runtime to `PATH` ? Doing that, PATH will be automatically correctly set by Theano and only inside Theano (changes are discarded as soon as Python closes). I made a branch about that: https://github.com/Theano/Theano/compare/master...notoraptor:allow-modify-path-at-runtime

Is it risky/unsafe, @nouiz @abergeron @lamblin ?

notoraptor:
Fix #6330 

Add `dnn.bin_path` to system's `PATH` environment variable on Windows only at runtime, so that user does not need to do it himself anymore. This works well on my computer (Windows 10 64 bits).

@abergeron @nouiz

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -49,6 +49,7 @@ from .opt import (gpu_seqopt, register_opt, pool_db, pool_db2,
 from .opt_util import alpha_merge, output_merge, inplace_allocempty, pad_dims, unpad_dims
 
 from theano.configdefaults import SUPPORTED_DNN_CONV_ALGO_RUNTIME
+import theano.pathparse
 
 DNN_CONV_ALGO_CHOOSE_ONCE = ['guess_once', 'time_once']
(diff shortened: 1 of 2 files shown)


notoraptor:
Sonething has stopped Jenkins, I don't know what and why.

abergeron:
The jenkins machine had some problems but it should be fine now:

jenkins test this please
jenkins test windows

nouiz:
jenkins test windows

nouiz:
jenkins test window

notoraptor:
jenkins test window

notoraptor:
jenkins test windows

PR #6359, opened 2017-08-25
Fixing math_type for F32 and cache update

borisfom:
Added conditions to exclude TENSOR_OP (HMMA) convo algos when operands  are 32-bit - since they lose precision. 
Also, streamlined saving old algo and cache updates - I think there was a bug in cache update condition there, too.

### DIFF
--- a/theano/gpuarray/c_code/dnn_fwd.c
+++ b/theano/gpuarray/c_code/dnn_fwd.c
@@ -360,7 +360,7 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
         dnn_conv_update_cache(hashkey, prev_algo);
 
       if (params->choose_once)
-      reuse_algo = 1;
+        reuse_algo = 1;
     }
 
   } // params->choose_algo
--- a/theano/gpuarray/c_code/dnn_fwd.c
+++ b/theano/gpuarray/c_code/dnn_fwd.c
@@ -231,7 +231,8 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
           return -1;
         }
         // set the 'tensor math ok' flag
-        c_set_math_type_for_conv(desc, CUDNN_TENSOR_OP_MATH);
+        if (input->ga.typecode == GA_HALF)
+          c_set_math_type_for_conv(desc, CUDNN_TENSOR_OP_MATH);
 
         // We don't sync the buffer as we don't care about the values.
         err = cudnnFindConvolutionForwardAlgorithmEx(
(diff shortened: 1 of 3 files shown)


nouiz:
jenkins test this

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -231,7 +231,8 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
           return -1;
         }
         // set the 'tensor math ok' flag
-        c_set_math_type_for_conv(desc, CUDNN_TENSOR_OP_MATH);
+        if (input->ga.typecode == GA_HALF)
```
nouiz: I understand that enabling that when input are float32 reduce the precision as you wrote in the description. But if it is enabled, does it speed up computation?

One option would be to cast the input to half in the graph before doing computation in float16, vs letting cudnn read in float32 and compute in float16. Is there one better?

At GTC, I recall that they kept the weights in float32, but casted it to float16 for computation. But I don't recall them telling if they would do a cast and call cudnn on the float16 result or let cudnn read float32 and convert it itself to float16.

theano/gpuarray/c_code/dnn_fwd.c
```diff
@@ -334,18 +331,17 @@ APPLY_SPECIFIC(conv_fwd)(PyGpuArrayObject *input, PyGpuArrayObject *kerns,
...
-    // Add to the cache if we choose on shape change, or first time if we choose once.
-    dnn_conv_update_cache(hashkey, prev_algo);
-  }
+    prev_algo.mathType = mathtype;
+    // Add to the cache if we choose on shape change, or first time if
+    // we choose once.
+    if (!use_cached)
+      dnn_conv_update_cache(hashkey, prev_algo);
 
 #ifdef DEBUG
```
notoraptor: With these changes, this debug code is not printed anymore when an algorithm is reused (for time_once or guess_once), as this code is now in a block with condition `if params->choose_algo and not reused`. So, it will not help us debug time_once and guess_once algorithms.

Same remark for dnn_gw and dnn_gi.

notoraptor:
Hello, @borisfom , I made a small commit just to reactivate debug printing in any case when params->choose_algo !

Everything should be fine, now @nouiz !

notoraptor:
jenkins test this

nouiz:
jenkins test this

nouiz:
jenkins test this

theano/gpuarray/c_code/dnn_fwd.c
borisfom: Thanks for the input, @nouiz !
I have raised the question with CUDNN team - it's understood that loss of precision should not be allowed by default - however, it may be a useful option to test HMMA speed without changing any code.

PR #6361, opened 2017-08-25
Fix rpath for cuDNN module.

notoraptor:
This PR fixes rpath for cuDNN ops by removing any quotes around paths. We should not need quotes on Windows anymore when #6356 will be merged.

NB: We should check if this PR does not break anything in buildbot. We also need Jenkins tests to pass, probably on all platforms ?

@abergeron @nouiz

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -227,7 +227,7 @@ dnn_available.msg = None
 
 def CUDNNDataType(name, freefunc=None):
     cargs = []
-    if config.dnn.bin_path:
+    if config.dnn.bin_path and sys.platform != 'win32':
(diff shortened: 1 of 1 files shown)


abergeron:
We probably want to avoid passing the rpath on windows since it could break the command line when there are spaces in the path.

notoraptor:
Ok ! I'll update it.

notoraptor:
Updated @abergeron .

nouiz:
jenkins test mac window

nouiz:
jenkins test window

nouiz:
jenkins test windows

nouiz:
I don't think this need Mac or windows tests. As the normal CI passed, I'll merge.

nouiz:
Forget last comment. I got mixed in PR.

PR #6372, opened 2017-08-31
Raise an error when device or init_gpu_device is set and we can't use it for some reason.

abergeron:


theano/gpuarray/__init__.py
```diff
@@ -173,26 +174,22 @@ def use(device,
...
-              config.init_gpu_device.startswith('opencl')):
-            if config.device != 'cpu':
-                raise ValueError(
-                    'you must set device=cpu to use init_gpu_device.')
-            if config.contexts != '':
-                print("Using contexts will make init_gpu_device act like device and move all computations by default, which might not be what you want.")
-            init_dev(config.init_gpu_device)
+    if (config.device.startswith('cuda') or
+            config.device.startswith('opencl')):
+        use(config.device)
```
nouiz: This one still need to fall back. There is the flag force_device=True that make this raise an error.

Personaly, I'm fine of changing this: No possible fallback and remove force_device. But I would like @lamblin agreement before doing that.

abergeron: I agree with you, we shouldn't (almost) silently ignore what the user asked for.

abergeron:
jenkins test this please

nouiz:
We don't want to change the interface now.

PR #6376, opened 2017-09-01
[FIX] Broadcast of corrmm

nouiz:


gvtulder:
There's another one here: https://github.com/Theano/Theano/blob/d66cefe72eaaf269d454eedb6320fddc372d0497/theano/gpuarray/blas.py#L1035-L1040

nouiz:
good finding. Fixed too.

PR #6377, opened 2017-09-01
fix flake8 whitelist on windows

slefrancois:


slefrancois:
jenkins test win

slefrancois:
jenkins test this

slefrancois:
jenkins test windows

nouiz:
I don't get why this is needed. It was working in the past. Also, in python, the directory separator is the same on all OS I think.

slefrancois:
Since we moved the windows buildbot to a new machine, the python os module has been generating paths with \ as the separator. The whitelist hardcodes / as the separator, hence the comparison was failing.

slefrancois:
flake8 wasn't installed before so it was skipped:
http://earlgrey.iro.umontreal.ca:8080/job/Theano_pr_win/32/testReport/theano.tests/test_flake8/test_format_flake8/

PR #6384, opened 2017-09-05
Fix pygpu version detection.

notoraptor:
Replace https://github.com/Theano/libgpuarray/pull/515 .

@nouiz


nouiz:
Jenkins GPU tests started. So all is good. I'll merge now to have the daily tests also run with the fix.

Issue #5633, opened 2017-02-28
Make AllocDiag support more than 2d output.


abergeron:
This is a follow-up for https://github.com/Theano/Theano/pull/5524 to make sure that we have the gradient for all cases.

ReyhaneAskari:
@abergeron , could you give some more explanation about this issue?

nouiz:
Mostly, we need to extend this class to be able to reprecent the gradient of GpuExtractDiag: https://github.com/Theano/Theano/pull/5524/files#diff-79c99d9fc3cc04cdd7b3d4121bd28c8cL1082

Mostly, it take as input a n-dim tensor instead of only 2d tensor.

Maybe we should make ExtractDiag support ndim as GpuExtractDiag first? Pick the order you prefer.

shawntan:
How should the n-dimensional GpuExtractDiag work though?

It is also 2 dimensional on the CPU version, so no clues there:
https://github.com/Theano/Theano/blob/e76d05d6ddf2384a15d42f627e8ade3e5411ae2b/theano/tensor/nlinalg.py#L225

The current implementation of GpuExtractDiag also has a line which I do not understand:
https://github.com/hantek/Theano/blob/9ae243d382eb80591cb17bc43daea91e6f73cb60/theano/gpuarray/subtensor.py#L1100

What is axis1 and axis2? Are these supplied by the user to define which axes define the matrix we are extracting the diagonal from?

If so, is the extension to make it so that we can extract diagonals from tensors as well? Like:
```X[[1,2,3,4,..], [1,2,3,4,...], [1,2,3,4,...]....]```

nouiz:
GpuExtractDiag is an equivalent implementation of np.diagonal from: https://github.com/hantek/Theano/blob/9ae243d382eb80591cb17bc43daea91e6f73cb60/theano/gpuarray/tests/test_subtensor.py#L189

So mostly, the first step is to make ExtractDiag support that case (but could just call np.diagonal in the perform instead of the complicated implementation in the GPU op. Make sure to document what it does (link to np doc) so user can find this information rapidly.

Then [Gpu]AllocDiag should be the gradient of that operation when the input have more then 2 dimensions.

shawntan:
I'm assuming you mean specify the offset, axis1, axis2  as part of perform instead of in __init__ ?
[...]

PR #6386, opened 2017-09-06
ExtractDiag cleanup #5633

shawntan:
Pull request for issue #5633 @nouiz @ReyhaneAskari

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -6413,6 +6413,26 @@ class ExtractDiag(Op):
         out_shape.append(diag_size)
         return [tuple(out_shape)]
 
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        if self.view and not numpy_diagonal_return_view:
+            warnings.warn("View will forced to False. ExtractDiag property view is "
+                          "set to True but numpy version %s and prior versions of "
+                          "numpy.diagonal() do not return a view. Update "
+                          "numpy to use ExtractDiag(view=True)" %
+                          np.version.version)
+            self.view = False
+
+        if self.view:
+            self.view_map = {0: [0]}
+
+        if "offset" not in state:
+            self.offset = 0
+        if "axis1" not in state:
+            self.axis1 = 0
+        if "axis2" not in state:
+            self.axis2 = 1
+
 
 def diagonal(a, offset=0, axis1=0, axis2=1):
     """
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -6365,10 +6365,8 @@ class ExtractDiag(Op):
     Returns
     -------
     array_of_diagonals:
-        If x is 2-D and not a matrix, a 1-D array of the
-        same type as a containing the diagonal is returned.
-        If x is a matrix, a 1-D array containing the diagonal
-        is returned in order to maintain backward compatibility.
+        If x is 2-D, a 1-D array of the same type as a
+        containing the diagonal is returned.
         If the dimension of x is greater than two, then an
         array of diagonals is returned, "packed" from left-most
         dimension to right-most (e.g., if x is 3-D, then the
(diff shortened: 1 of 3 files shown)


shawntan:
Starting the PR off.

shawntan:
Not sure how the `__setstate__` should be written, but I basically checked the attributes that would be set by nlinalg and set them to their default if the nlinalg version doesn't set them

theano/tensor/nlinalg.py
```diff
@@ -193,75 +193,6 @@ def infer_shape(self, node, shapes):
...
-            x = _x
-
-        if x.type.ndim != 2:
-            raise TypeError('ExtractDiag only works on matrices', _x)
-        y = x.type.clone(broadcastable=(False,))()
-        return Apply(self, [x], [y])
-
-    def perform(self, node, ins, outs):
-        """ For some reason numpy.diag(x) is really slow, so we
-        implemented our own. """
```
ReyhaneAskari: @nouiz , the perform function in here is different from the one in tensor/basic.  It says numpy's diagonal is slow and has its own implementation. Do we want to remove this code?

theano/tensor/tests/test_nlinalg.py
```diff
@@ -384,6 +375,8 @@ def test_extract_diag(self):
             extract_diag(xx)
         except TypeError:
             ok = True
+        except ValueError:
```
ReyhaneAskari: @shawntan , is this for the case where `x.ndim < 2`?

theano/tensor/tests/test_nlinalg.py
```diff
@@ -429,6 +422,9 @@ def test_trace():
         trace(xx)
     except TypeError:
         ok = True
+    except ValueError:
```
ReyhaneAskari: same here.

ReyhaneAskari:
Thanks @shawntan . I think your `__setstat__` method is correct. Could you also add some explanation for what `offset`, `axis1` and `axis2` are? You can find the equivalent explanations here: https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.diagonal.html

Could you also remove the parts that `theano.tensor.nlinalg.ExtractDiag` are mentioned? 
https://github.com/shawntan/Theano/blob/55ccab9aa44d59195145b8ba3f3dd0350e23edc8/theano/tensor/basic.py#L6388

I don't have access to see the results of appoveyor. @nouiz could you give me access?

theano/tensor/tests/test_nlinalg.py
shawntan: Yes, because the `basic.py` implementation throws a `ValueError` instead of a type error. 
The fix would either be to:
- Add the exception for `ValueError`
- Change the exception to `TypeError`

Whichever you guys think is more appropriate.

ReyhaneAskari: I think the `ValueError` is correct. The old code where the `TypeError` was raised is going to be removed so I think the `TypeError` should be removed. Is it raised any other place than line 225 of the old code?

theano/tensor/nlinalg.py
nouiz: This is for an old version of NumPy. I think we can get rid of this code now.

theano/tensor/basic.py
```diff
@@ -6333,18 +6333,59 @@ def all(x, axis=None, keepdims=False):
...
+        Defaults to second axis (1).
+
+
 
     Returns
     -------
-    vector
-        A vector representing the diagonal elements.
+    array_of_diagonals:
+        If x is 2-D and not a matrix, a 1-D array of the
```
nouiz: remove "and not a matrix". (Theano don't have numpy matrix.  a numpy matrix isn't a 2d array, it is a different object).

theano/tensor/basic.py
```diff
@@ -6333,18 +6333,59 @@ def all(x, axis=None, keepdims=False):
...
+        same type as a containing the diagonal is returned.
+        If x is a matrix, a 1-D array containing the diagonal
+        is returned in order to maintain backward compatibility.
```
nouiz: Remove that sentence about matrix.

Issue #6360, opened 2017-08-25
Convolution metaoptimizer needs tests for grouped and unshared convolution


gvtulder:
The [ConvMetaOptimizer](https://github.com/Theano/Theano/blob/c32b0db8070271a5ae0bf08120bab29c04bb00ae/theano/gpuarray/opt.py#L2086) works with several optimizations that handle grouped convolution and, when #6286 is merged, unshared convolution. It looks like most of these optimisations are only tested for normal convolutions. It would be good to add some tests to the [Conv_opt_test](https://github.com/Theano/Theano/blob/c32b0db8070271a5ae0bf08120bab29c04bb00ae/theano/gpuarray/tests/test_opt.py#L709) that check if the grouped and unshared parameters are passed correctly.

affanv14:
Hi @gvtulder, all alternative optimizers return None for grouped convolutions (I presume it is the same for unshared convolutions as well?). The default optimizers, on the other hand,  have been tested in the grouped convolution tests and unshared convolutions test. So the only two Optimizers which need to be tested are `local_abstractconv_gemm_def` and `local_abstractconv3d_gemm_def` which I added the other day.

While doing the meta-optimizer tests, I did think of adding a check to see that the alternative optimizers are returning None when it has attributes that dont apply(like subsample >(1, 1) or numgroups >1 etc.). But then I left the Idea because it would bloat the tests. would it be a good idea to pursue this test?

gvtulder:
Hi @affanv14. This came about because I was wondering where to add a test for the change you suggested in the unshared convolution PR #6286. That was one of the opts that didn't understand the `unshared` flag and just ignored it. Failing to rejecting an invalid optimization is also an error, so I think it would be safer to have a test to make sure that that happens.

Aside: I'm not too familiar with these new optimizations (and perhaps this is already fixed in a PR somewhere), but shouldn't this optimization be doing something with `grouped` as well?
https://github.com/Theano/Theano/blob/c32b0db8070271a5ae0bf08120bab29c04bb00ae/theano/gpuarray/opt.py#L2061-L2083

affanv14:
yeah, I just realised that 5 mins back and edited the message above your message. I'm already working on a fix and I will add the tests soon.

affanv14:
This error probably happened because I was working on both of them concurrently.

gvtulder:
[...]

PR #6389, opened 2017-09-06
Improve tests for meta-optimizer

affanv14:
Related to https://github.com/Theano/Theano/issues/6360


affanv14:
the problem I had in https://github.com/Theano/Theano/pull/6362#issuecomment-326071182 was due to calling conv2d instead of calling `AbstractConv2d` directly. Specifically, the  `as_tensor_variable` function in conv2d(and other backward passes) was causing this.

nouiz:
jenkins test this

nouiz:
thanks.

Issue #6387, opened 2017-09-06
Implementation of zoom operator


Overdrivr:
I have implemented for a private project a Theano operator for `zoom` function from `scipy.ndimage.interpolation`, and I'd like to bring this out in the open. Would you be interested in a PR ?

It is a pure python implementation, gradient is implemented and I wrote a few tests as well.
The only downside is that `verify_gradient` tests fail if the scaling factor is different than 1.0. I'm not sure how to fix it or if there is something to fix given that optimisation seems to perform ok in my application, but maybe with some guidance I can make this work.

nouiz:
I think it would be great to include it. Can you make a PR that include it
in a new file theano/tensor/scipy_ndimage.py and the tests in
theano/tensor/tests/test_scipy_ndimage.py?

With the code, we can maybe spot the problem with a different scaling
factor.

thanks

PR #6390, opened 2017-09-07
Implement scipy.ndimage.interpolation:zoom operator

Overdrivr:
Implementation of `scipy.ndimage.interpolation:zoom` function as a Theano operator.

This was discussed briefly in #6387 

IMO there are three things to fix:

- test `test_Zoom_gradient_single_channel_factor_2`  fails, most likely the way gradient is computed is wrong 
- The operator accepts two tensors, and will zoom the first tensor so that it has the same shape than the second. It would be better if the operator accepts a tensor and a resize factor/resize shape. I implemented it in the first way because it was the easiest given my limited knowledge of Theano, but we should be able to fix that pretty easily.
- interpolation order is hardcoded to 0, expose it as parameter

### DIFF
--- a/theano/tensor/scipy_ndimage.py
+++ b/theano/tensor/scipy_ndimage.py
@@ -1,0 +1,47 @@
+import theano
+from theano.gradient import DisconnectedType
+from scipy.ndimage.interpolation import zoom
+import numpy as np
+
+class Zoom(theano.Op):
+    # Properties attribute
+    __props__ = ()
+
+    def __init__(self):
+        super(Zoom, self).__init__()
+
+    def make_node(self, x, ref):
+        x = theano.tensor.as_tensor_variable(x)
+        ref = theano.tensor.as_tensor_variable(ref)
+        return theano.Apply(self, [x, ref], [x.type()])
+
+    # Python implementation:
+    def perform(self, node, inputs, outputs_storage):
+        x, ref = inputs
+        z = outputs_storage[0]
+        assert x.ndim == ref.ndim
+        assert x.shape[0] == 1
+        #
+        resize_factors = [
+            ref.shape[2] / x.shape[2],
+            ref.shape[3] / x.shape[3]
+            ]
+        #
+        final_shape = (1, x.shape[1], ref.shape[2], ref.shape[3])
+        z[0] = np.empty(final_shape, dtype=np.float32)
+
+        # Loop over channels
+        for i in range(x.shape[1]):
+            z[0][0,i] = zoom(x[0,i], resize_factors, order=0)
+
+    def infer_shape(self, node, i0_shapes):
+        return i0_shapes[1],
+
+    def connection_pattern(self, node):
+        return [[True], [False]]
+
+    def grad(self, inputs, output_grads):
+        op = Zoom()
+        x, ref = inputs
+        grads = output_grads[0]
+        return op(output_grads[0], x), DisconnectedType()()
--- a/theano/tensor/scipy_ndimage.py
+++ b/theano/tensor/scipy_ndimage.py
@@ -1,7 +1,11 @@
+from __future__ import absolute_import, print_function, division
+
+import numpy as np
+from scipy.ndimage.interpolation import zoom
+
 import theano
 from theano.gradient import DisconnectedType
-from scipy.ndimage.interpolation import zoom
-import numpy as np
+
 
 class Zoom(theano.Op):
     # Properties attribute
@@ -32,7 +36,7 @@ class Zoom(theano.Op):
 
         # Loop over channels
         for i in range(x.shape[1]):
-            z[0][0,i] = zoom(x[0,i], resize_factors, order=0)
+            z[0][0, i] = zoom(x[0, i], resize_factors, order=0)
 
     def infer_shape(self, node, i0_shapes):
         return i0_shapes[1],
@@ -43,5 +47,4 @@ class Zoom(theano.Op):
     def grad(self, inputs, output_grads):
         op = Zoom()
         x, ref = inputs
-        grads = output_grads[0]
         return op(output_grads[0], x), DisconnectedType()()
--- a/theano/tensor/tests/test_scipy_ndimage.py
+++ b/theano/tensor/tests/test_scipy_ndimage.py
@@ -1,25 +1,35 @@
+from __future__ import absolute_import, print_function, division
+
 import unittest
+from scipy.ndimage.interpolation import zoom
+import numpy as np
+from numpy.testing import assert_array_equal
+
+import theano as T
 from theano.tensor.scipy_ndimage import Zoom
 from theano.tests.unittest_tools import verify_grad
-import theano as T
-from scipy.ndimage.interpolation import zoom
-from numpy.testing import assert_array_equal
-import numpy as np
+
 
 def getFn():
-    x   = T.tensor.tensor4()
+    x = T.tensor.tensor4()
     ref = T.tensor.tensor4()
     op = Zoom()
     return T.function([x, ref], op(x, ref))
+
 
 class T_SharedRandomStreams(unittest.TestCase):
     def test_Zoom_single_channel_factor_2(self):
         NCHANS = 1
         ZFACTOR = 2
-        insize  = (1, NCHANS, 32, 32)
-        outsize = (1, NCHANS, int(insize[2] * ZFACTOR), int(insize[3] * ZFACTOR))
+        insize = (1, NCHANS, 32, 32)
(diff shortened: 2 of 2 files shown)


theano/tensor/scipy_ndimage.py
```diff
@@ -0,0 +1,47 @@
...
+class Zoom(theano.Op):
+    # Properties attribute
+    __props__ = ()
+
+    def __init__(self):
+        super(Zoom, self).__init__()
+
+    def make_node(self, x, ref):
+        x = theano.tensor.as_tensor_variable(x)
+        ref = theano.tensor.as_tensor_variable(ref)
```
gvtulder: Wouldn't a shape work here as well? You don't really use `ref` except to get its shape.

gvtulder: Wouldn't a shape work here as well? You don't really use `ref` except to get its shape.

theano/tensor/scipy_ndimage.py
```diff
@@ -0,0 +1,47 @@
...
+    # Python implementation:
+    def perform(self, node, inputs, outputs_storage):
+        x, ref = inputs
+        z = outputs_storage[0]
+        assert x.ndim == ref.ndim
+        assert x.shape[0] == 1
+        #
+        resize_factors = [
+            ref.shape[2] / x.shape[2],
+            ref.shape[3] / x.shape[3]
```
gvtulder: I think this might be the cause of some of your gradient problems. Your shapes are integers, so your `resize_factors` will be computed as integers as well. That doesn't work for the gradient of a `zoom(2)` upsampling, because you'd have to downsample the gradient with `zoom(0.5)`.

gvtulder:
I played a bit more with the `zoom` function, and I think it might be tricky to express an exact gradient in terms of the `zoom` function itself. With `order=0`, it skips every second row. So this:
```python
import numpy as np
import scipy.ndimage
topgrad = np.zeros((6, 6))
topgrad[1, 1] = 1
print(scipy.ndimage.zoom(topgrad, 0.5, order=0))
```
produces a result with only zeros, which isn't correct:
```
[[ 0.  0.  0.]
 [ 0.  0.  0.]
 [ 0.  0.  0.]]
```
If you place a single `1` at `topgrad[0, 0]`, the result at `[0, 0]` would be `1`.

For your specific 2D test, a strided convolution would give the correct gradient:
```python
    def grad(self, inputs, output_grads):                                                                 
        x, ref = inputs
        grads = output_grads[0]
        return (T.nnet.conv2d(grads, T.ones((1, 1, 2, 2)), subsample=(2, 2)),
                DisconnectedType()())
```
but that probably wouldn't work for more difficult cases.

theano/tensor/scipy_ndimage.py
Overdrivr: I've added the `from __future__ import division` line which will ensure that using the `/` operator will return a floating point number. Tests confirm this behavior so this should be ok.

Overdrivr:
I've linted the files, test seem to be running fine now.

With order 0, no interpolation is performed so the behavior with a .5 factor is indeed to skip every 2nd line. With a higher order (default is 3) the output is much more what you will be expecting:
```
>>> import numpy as np
>>> import scipy.ndimage
>>> topgrad = np.zeros((6, 6))
>>> topgrad[1, 1] = 1
>>> print(scipy.ndimage.zoom(topgrad, 0.5, order=3))
[[  0.00000000e+00   1.95156391e-17  -3.46944695e-18]
 [  2.10335221e-17   1.85950413e-02  -4.38763067e-19]
 [  1.73472348e-18  -7.03037346e-20  -2.71050543e-20]]
```
Depending on the use case, different orders could be interesting.

In Theano, do all operators pass the `verify_grads` test ?

gvtulder:
True, the problem might be less severe for higher orders, but it still seems to be quite far off. I've tried to run the test with `order=3`, and then `verify_grad` fails with `abs. error = 2.905477` and `rel. error = 0.811812`. I'm not sure if that's still useful.

You could increase the tolerance of `verify_grad`, but I think it's good to check if the gradient is at least somewhat close to what it should be. I haven't seen any ops with a gradient that doesn't pass `verify_grad` (save for a few rounding errors here and there perhaps).

Would it be an option to reuse and adapt the [SciPy zoom code](https://github.com/scipy/scipy/blob/0c42ad1b50c1d6c738c964e2aa60351e34ac2812/scipy/ndimage/src/ni_interpolation.c#L686-L688) so that it produces a more exact gradient?

Overdrivr:
At the moment, I don't have the time resources for modifying scipy zoom code.

Do you think the gradient is wrong by a constant gain/offset or something more elaborate ?
Is there a way to obtain both gradients obtained by `verify_grads` function in order to plot them ?

gvtulder:
My apologies for hijacking your project, but since you don't have time I hope this is somewhat okay.

I've taken the scipy C code and modified it to compute gradients (see the changes in [this branch](https://github.com/gvtulder/Theano/tree/f-scipy-zoom)) to see what would happen. It needs a lot of polishing, but it can compute the gradients by reversing the mapping of the scipy ZoomShift function. It seems to pass an initial `verify_grad` test ([messy example script](https://github.com/gvtulder/Theano/blob/f-scipy-zoom/test_scipy_zoomshift.py)).

To see what happens, I computed a few gradients using scipy.zoom (order=3) with inverse scaling factors and the C gradient. For a gradient of `T.sum(Zoom(... 2x upscaling ...))` they give these results:
```
C gradient:
[[ 0.44444444  0.24622222  0.35911111  0.27644444  0.00711111]
 [ 0.33333333  0.18466667  0.26933333  0.20733333  0.00533333]
 [ 0.44444444  0.24622222  0.35911111  0.27644444  0.00711111]
 [ 0.33333333  0.18466667  0.26933333  0.20733333  0.00533333]
 [ 0.44444444  0.24622222  0.35911111  0.27644444  0.00711111]]

inverted scipy.zoom:
[[ 1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.]
 [ 1.  1.  1.  1.  1.]]
```
I think this makes sense, because the top gradient coming from `T.sum` is a matrix of all ones, which then gets downsampled to all ones as well.

These are the results with `T.grad(..., known_grads={...})` with a random known gradient:
```
C gradient:
[[ 0.29333606  0.14946435  0.04184127  0.0322095   0.00082854]
 [ 0.16314582  0.08438488  0.04207543  0.03238975  0.00083318]
 [ 0.06591116  0.03861082  0.08461917  0.06514001  0.00167563]
 [ 0.22838382  0.1186777   0.06712076  0.0516697   0.00132912]
 [ 0.42381206  0.21805     0.09193194  0.07076939  0.00182043]]

inverted scipy.zoom:
[[ 0.37836519  0.25961676  0.44215691  0.62892908  0.53266454]
 [ 0.31825665  0.2986151   0.53101039  0.71017307  0.66028428]
 [ 0.57483345  0.40763474  0.56329811  0.52056539  0.53055316]
 [ 0.43326476  0.43874088  0.66140777  0.5397504   0.67137665]
 [ 0.34846017  0.47797978  0.69656616  0.42527339  0.57287496]]
```
So I don't think it's a simple scaling.

Overdrivr:
@gvtulder No problem, I'm glad you did actually :) That's great stuff.

If I understand correctly, you implemented a way to exactly reverse the Zoom operation, and apply that to the gradients. Is this correct ?

Do you want/have time to take over this ticket ?

gvtulder:
@nouiz The gradient of `scipy.ndimage.zoom` cannot be implemented in Python. To do it in C you would have to copy [these three](https://github.com/gvtulder/Theano/tree/f-scipy-zoom/theano/tensor/c_code/scipy_ndimage) (slightly modified) C files from the SciPy source to Theano. Is that acceptable?

@Overdrivr Sure, I can clean up the code.
The implementation in that branch is indeed an exact reverse of the ZoomShift function in SciPy. It uses the same loop and coefficient computations as the forward function, but instead of reading values from the input image and writing them to the output image, it reads from the output gradient and writes to the input gradient.

nouiz:
@gvtulder Using code from numpy/scipy is fine. We have the same license. Just make sure it is written clearly the change you did and where the original file is. Ideally, you should have a commit with the original file and one with the changes. This make the change very clear.

I'm sure this can be done in a op with a perform. It would be probably very slow. It isn't requested to add the python implementation for this to be merged.

We decided to make a 0.10.beta3 I think we will do it Monday. So it it get finished today or during the week-end, we could include it.

PR #6392, opened 2017-09-07
Fix test error in FAST_COMPILE (show in daily buildbot)

nouiz:


theano/gpuarray/tests/test_dnn.py
```diff
@@ -1032,6 +1034,8 @@ def dconvw(img, kern, out):
 
 
 def get_conv3d_test_cases():
+    if not dnn.dnn_available(test_ctx_name):
+        raise SkipTest(dnn.dnn_available.msg)
```
notoraptor: This checking can be removed here (`get_conv3d_test_cases()` is not executed, it is instead called by other test functions [`test_conv3d_fwd` and `test_conv3d_bwd`, which are already skipped if cuDNN is not available]).

nouiz: For some unknow reason, when I do:

nosetests theano/gpuarray/tests/test_dnn.py

this function get executed as a test. Even if the setup.cfg tell otherwise. As I don't understand why it try to execute it, I just make it sure it will work as a test fct. To speed up making the next rc.

notoraptor: Ok !

PR #6396, opened 2017-09-08
Scipy nfunc

nouiz:
See [[theano-users] Overhead of Elemwise{Erfinv} on CPU](https://groups.google.com/d/topic/theano-users/gdDKrM2gBs8/discussion) for user error report.


nouiz:
Before merging, we need to verify that Theano still work when scipy isn't there.

juancamilog:
It looks like that fixes the issue for the example I posted. However, the issue persists if you change the example to produce a composite elementwise operation.

For example:

import numpy as np
import theano
import theano.tensor as tt
from scipy.special import erfinv

z = tt.vector()
y = tt.erfinv(z**2)

f = theano.function([z], y)
f.trust_input = True

x = np.linspace(-1,1,1000000)

%timeit f(x)

%timeit erfinv(x**2)

In this case the theano version is about 150x slower.

nouiz:
For Composite, they should not be fused when we don't have C code. So I don't have this problem here as that op don't have C code. I suppose the problem is that locally, you added C code as you wrote on the mailing list and that this C code is slower then scipy c code.

Just remove it. The time I have with your new code here is:
10 loops, best of 3: 35.3 ms per loop
10 loops, best of 3: 34.5 ms per loop

juancamilog:
Ok, I've tested the latest version of this pull request and it works in my examples and other use cases. Thanks!

Issue #6381, opened 2017-09-05
Request for truncated normal distribution


mrTsjolder:
In neural networks, it has become common to initialise the weight by means of truncated normals. Tensorflow uses a simplified version of the truncated normal that simply cuts off the tails after two standard deviations ([tensorflow docs](https://www.tensorflow.org/api_docs/python/tf/truncated_normal)). An implementation of a truncated normal still seems to be missing in Theano, however. I only discovered this issue through the [poor man's implementation in keras](https://github.com/fchollet/keras/blob/master/keras/backend/theano_backend.py#L2238-L2246), which got me on the wrong leg in my analysis.

After taking a look at the code of `MRG_RandomStreams`, I found out that sampling from the [normal distribution](https://github.com/Theano/Theano/blob/master/theano/sandbox/rng_mrg.py#L1032-L1118) is done by means of the Box-Muller transform. While reading the referenced Wikipedia page, I found [this section on tails truncation](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform#Tails_truncation), which suggests that a truncated normal distribution with borders at `2 * std` should not be too hard to implement. 

Concretely, it would be nice to have something like

    def _normal_from_uniform(self, size, avg=0.0, std=1.0, low=0.0
                             ndim=None, dtype=None, nstreams=None):
        # the previous normal implementation with extra parameter
        # from: value in range [0, 1[ for cutting off tails
        avg = undefined_grad(as_tensor_variable(avg))
        std = undefined_grad(as_tensor_variable(std))

        if dtype is None:
            dtype = scal.upcast(config.floatX, avg.dtype, std.dtype)

        avg = cast(avg, dtype)
        std = cast(std, dtype)

        evened = False
        constant = False
        if (isinstance(size, tuple) and
                all(isinstance(i, (np.integer, integer_types)) for i in size)):
            constant = True
            n_samples = np.prod(size, dtype='int64')

            if n_samples % 2 == 1:
                n_samples += 1
                evened = True
        else:
            n_samples = prod(size) + (prod(size) % 2)

        # Use from to limit the number of samples (only necessary change)
        flattened = self.uniform(size=(n_samples,), low=from,
                                 dtype=dtype, nstreams=nstreams)

        if constant:
            U1 = flattened[:n_samples // 2]
            U2 = flattened[n_samples // 2:]
        else:
            U1 = flattened[:prod(flattened.shape) // 2]
            U2 = flattened[prod(flattened.shape) // 2:]

[...]

PR #6401, opened 2017-09-10
Implement truncated normal with box-muller

mrTsjolder:
This is an implementation of the box-muller transform for a normal distribution truncated at two standard deviations (as requested in issue #6381).

I did not quite get why there would be need to distinguish between integer and tensor sizes, given that theano optimises the graph and probably includes constant folding. Therefore, I have introduced my own interpretation of the normal code (which I personally find more readable, if I'm allowed to be smug for a second). If one of two approaches turn out to be better/faster/more readable/more whatever, quite a bit of code can be reused and a refactoring might be advisable.

### DIFF
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -1070,7 +1070,11 @@ class MRG_RandomStreams(object):
         std = tensor.cast(std, dtype=dtype)
 
         # generate even number of uniform samples
-        n_odd_samples = tensor.prod(size, dtype='int64')
+        # Do manual constant folding to lower optiimizer work.
+        if isinstance(size, theano.Constant):
+            n_odd_samples = size.prod(dtype='int64')
+        else:
+            n_odd_samples = tensor.prod(size, dtype='int64')
         n_even_samples = n_odd_samples + n_odd_samples % 2
         uniform = self.uniform((n_even_samples, ), low=0., high=1.,
                                ndim=1, dtype=dtype, nstreams=nstreams, **kwargs)
@@ -1106,8 +1110,12 @@ class MRG_RandomStreams(object):
             norm_samples = tensor.join(0, z0_valid, z0_fixed, z1_valid, z1_fixed)
         else:
             norm_samples = tensor.join(0, z0, z1)
-
-        samples = norm_samples[:n_odd_samples]
+        if isinstance(n_odd_samples, theano.Variable):
+            samples = norm_samples[:n_odd_samples]
+        elif n_odd_samples % 2 == 1:
+            samples = norm_samples[:-1]
+        else:
+            samples = nrom_samples
         samples = tensor.reshape(samples, newshape=size, ndim=ndim)
         samples *= std
         samples += avg
--- a/theano/sandbox/rng_mrg.py
+++ b/theano/sandbox/rng_mrg.py
@@ -19,7 +19,6 @@ from theano.gradient import undefined_grad
 from theano import tensor
 from theano.tensor import (TensorType, as_tensor_variable, get_vector_length,
                            cast, opt, scal)
-from theano.tensor import sqrt, log, sin, cos, join, prod
 from theano.compile import optdb
 from theano.gof import local_optimizer, ParamsType
 from theano.scalar import bool as bool_t, int32 as int_t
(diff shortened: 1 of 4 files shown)


nouiz:
The reason why we generate a simpler graph for constant is that it make reading the not optimized graph easier and make the compilation faster. It isn't stricly requested, but is nice to do.

This would need test in the file theano/sandbox/tests/test_rng_mrg.py to verify is the generated distribution is well clipped.

theano/sandbox/rng_mrg.py
```diff
@@ -1117,6 +1118,79 @@ def normal(self, size, avg=0.0, std=1.0, ndim=None,
...
+        to_fix0 = bitwise_or(z0 < -2., z0 > 2.)
+        to_fix1 = bitwise_or(z1 < -2., z1 > 2.)
+        z0_valid = z0[nonzero(bitwise_not(to_fix0))]
+        z1_valid = z1[nonzero(bitwise_not(to_fix1))]
+
+        # re-sample invalid samples
+        to_fix0 = nonzero(to_fix0)[0]
+        to_fix1 = nonzero(to_fix1)[0]
+        n_fix_samples = to_fix0.size + to_fix1.size
+        u_fix = self.uniform(size=(n_fix_samples, ), low=0.13585, high=1.,
```
nouiz: From the wikipedia page, I think the "low" number would be dtype dependent.

I would be interrested to know how the other framework does it. Did you looked at that? I checked tensorflow and it a variable number of time until it accept a number. The way you do it seem interresting and could prevent that need.

I would like the input from @lamblin and @abergeron on the way you implement it. resample only once with with a different uniform to force a max of 2 stddev.

mrTsjolder:
I would argue that it is more valuable to have readable code than to have readable graphs.  If there are really problems with readability of optimised graphs, it should be solved where the origin of these problems lie. Note that this is just a personal opinion of mine.

As to the requested changes: What am I supposed to change? Is it just the test case or are other changes necessary (e.g. refactoring)? Strictly speaking, it is probably not completely correct, but I think it does well for the 5% of data that is "out of bounds".

mrTsjolder:
I noticed that there is a `numpy`-based random generator as well. `numpy` does not offer a truncated normal distribution (yet). I proposed a `numpy` implementation of the box-muller transform in #6381, which could be used. Another option, would be to use random number generators from `scipy.stats`, but I am not sure whether that is an option.

theano/sandbox/rng_mrg.py
mrTsjolder: In what sense would `low` depend on the `dtype`? If `low = 0.`, the `dtype` will indeed cause tail-truncation (that is where my idea for the resampling came from).

Which other framework? Keras simply samples from a normal and clips all values to the edges, leaving a distribution that is very non-Gaussian and severely messes with the variance. Tensorflow does indeed some iterative resampling and I have no idea where the CNTK implementation is hidden.

nouiz:
I don't think it is useful to implement it in the numpy based random generator. It is slower and don't work on GPU.

theano/sandbox/rng_mrg.py
```diff
@@ -1117,6 +1118,81 @@ def normal(self, size, avg=0.0, std=1.0, ndim=None,
         assert final_samples.dtype == dtype
         return final_samples
 
+    def truncated_normal(self, size, avg=0.0, std=1.0, ndim=None,
+                         dtype=None, nstreams=None):
```
abergeron: This would need to add support for the new `target` parameter: https://github.com/Theano/Theano/pull/6355

mrTsjolder:
In my enthusiasm, I changed quite a bit of code and added an (inefficient) implementation for binomial distributions for n > 1. I have also made it possible to get 0-dimensional sizes (which happens to be convenient for the normal distribution). I also tried to improve the documentation and overall consistency. I hope this is no issue

The tensorflow implementation has been made available through the `truncate`-flag in `normal`. `truncated_normal` returns samples from a truncated normal with the provided standard deviation.

As to the failing circleci-test: could it be that this is not due to code that I implemented?

abergeron:
You can ignore CircleCI for this PR, the base branch did not have the proper config.

theano/sandbox/rng_mrg.py
```diff
@@ -1029,96 +1048,177 @@ def multinomial_wo_replacement(self, size=None, n=1, pvals=None,
...
+        Returns
+        -------
+        samples : TensorVariable
+            A Theano tensor of samples randomly drawn from a truncated normal distribution.
+
+        See Also
+        --------
+        normal
+        """
+        std = std / tensor.constant(.87962566103423978)
```
notoraptor: Is there a mathematical explanation/expression for that number (`.87962566103423978`) ? If so, could you use it instead of putting directly the number value ? As you do in line 1118 with `lower = tensor.constant(1. / np.e**2, dtype=dtype)`.

theano/sandbox/rng_mrg.py
```diff
@@ -804,101 +803,121 @@ def pretty_return(self, node_rstate, new_rstate, sample, size, nstreams):
...
 
-        if u.type.broadcastable != r.type.broadcastable:
-            raise NotImplementedError(
-                'Increase the size to match the broadcasting pattern of '
-                '`low` and `high` arguments')
+        return samples
 
-        assert r.dtype == dtype
-        return r
+    def binomial(self, size, n=1, p=0.5,
```
notoraptor: There are currently no tests for `binomial()` when `n > 1`, in `theano/sandbox/tests/test_rng_mrg.py`. Could you add a test for that case ?

theano/sandbox/rng_mrg.py
```diff
@@ -804,101 +803,121 @@ def pretty_return(self, node_rstate, new_rstate, sample, size, nstreams):
...
+        kwargs
+            Other keyword arguments for random number generation (see uniform).
+
+        Returns
+        -------
+        samples : TensorVariable
+            A Theano tensor of samples randomly drawn from a binomial distribution.
+
+        """
+        if isinstance(n, (np.integer, integer_types)):
```
notoraptor: What if `n` is a floating python or numpy value ?

theano/sandbox/rng_mrg.py
mrTsjolder: It is the result of `scipy.stats.truncnorm.std(a=-2, b=2, loc=0., scale=1.)` (see comment in `test_normal_truncation`). I was not sure whether this is worth a scipy import or whether Theano already depends on scipy. Should I replace the constant by the above mentioned line?

lamblin: Theano has no hard dependency on scipy, so it's better not to import it here.
If you could just copy the comment here as well, I think that should be enough.

notoraptor: I agree with @lamblin , just adding a comment to tell how it is calculated should be sufficient.

notoraptor: PS: Ok, comment is in `test_normal_truncation`. Could you add it here too, in `truncated_normal()` function ?

theano/sandbox/rng_mrg.py
```diff
@@ -887,11 +887,12 @@ def binomial(self, size, n=1, p=0.5,
             A Theano tensor of samples randomly drawn from a binomial distribution.
 
         """
-        if isinstance(n, (np.integer, integer_types)):
+        # TODO: use ABCs for these checks?
+        if not isinstance(p, theano.Variable):
```
notoraptor: I think it should be `n` instead of `p` here, isn't it ?

Also, maybe add after that line something like:
```python
assert n == int(n), "n must be an integer"
```

theano/sandbox/tests/test_rng_mrg.py
```diff
@@ -321,60 +321,62 @@ def test_broadcastable():
...
+        ((3,), (3,), [], [], 3, .7, default_rtol, 0.02)
+    ]
 
-def t_binomial(mean, size, const_size, var_input, input, steps, rtol):
-    R = MRG_RandomStreams(234)
-    u = R.binomial(size=size, p=mean)
-    f = theano.function(var_input, u)
-    f(*input)
+    for size, const_size, var_input, input, n, p, rtol, std_tol in test_cases:
+        print(const_size, n, p, rtol, std_tol)
```
notoraptor: This print should be removed for final code.

theano/sandbox/rng_mrg.py
```diff
@@ -887,15 +887,18 @@ def binomial(self, size, n=1, p=0.5,
...
             elif n > 10:
-                warnings.warn("Current implementation binomial distribution is memory-inefficient for large n (%d)" % n)
+                warnings.warn("Current implementation binomial distribution leads to "
+                              "memory consumption linear in n (%d)" % n)
 
-        if not isinstance(p, theano.Variable) and (p < 0. or p > 1.):
-            raise ValueError("Probability of success must be between 0 and 1 for a binomial distribution", p)
+        if not isinstance(p, theano.Variable):
+            assert p == float(p), "p must be a float"
+            if not 0. < p < 1.:
```
notoraptor: p can be 0 or 1, so it should be `if not 0. <= p <= 1.` instead !

mrTsjolder: I would like to apologise for my idiocy...

theano/sandbox/rng_mrg.py
```diff
@@ -804,101 +803,125 @@ def pretty_return(self, node_rstate, new_rstate, sample, size, nstreams):
         node_rstate.default_update = new_rstate
         return sample
 
-    def uniform(self, size, low=0.0, high=1.0, ndim=None, dtype=None,
-                nstreams=None, **kwargs):
-        # TODO : need description for parameter 'size', 'ndim', 'nstreams'
+    def uniform(self, size, low=0.0, high=1.0,
+                ndim=None, dtype=None, nstreams=None, target=None):
```
nouiz: Why the change about target? Now it is different then the other method and I would prefer to not change this. So I would like to know why you changed it.

theano/sandbox/rng_mrg.py
```diff
@@ -804,101 +803,125 @@ def pretty_return(self, node_rstate, new_rstate, sample, size, nstreams):
...
+        low = undefined_grad(as_tensor_variable(low))
+        high = undefined_grad(as_tensor_variable(high))
 
         if dtype is None:
-            dtype = scal.upcast(config.floatX, low.dtype, high.dtype)
+            dtype = scal.upcast(floatX, low.dtype, high.dtype)
 
-        low = cast(low, dtype=dtype)
-        high = cast(high, dtype=dtype)
+        low = tensor.cast(low, dtype=dtype)
```
nouiz: This change the graph. Now the undefined_grad is swapped comparted to cast in the graph. I don't think this is a problem, but I prefer to raise it in case someone think it is a problem.

theano/sandbox/rng_mrg.py
```diff
@@ -13,17 +13,15 @@
...
-from theano.compile import optdb
-from theano.gof import local_optimizer, ParamsType
+from theano.tensor import (TensorType, as_tensor_variable, scal)
+from theano.gradient import grad_undefined, undefined_grad
 from theano.scalar import bool as bool_t, int32 as int_t
-from . import multinomial
+from theano.sandbox import multinomial
+
+floatX = theano.config.floatX
+intX = floatX.replace('float', 'int')
```
nouiz: I don't like this change. This change the user interface just before a release and isn't good. If you want this PR to be included in the next release, can you keep in it just the changes about the new random generator?

Also, did you see the announce: https://groups.google.com/d/msg/theano-users/7Poq8BZutbY/rNCIfvAEAwAJ

This mostly tell we won't change the interface, so moving this interface change to another PR isn't useful.

mrTsjolder: Would it be okay to use `_floatX` and `_intX` instead to make them *private*. That would leave the interface unchanged (assuming that private variables are not part of the interface).

Note: as you might notice, I am not a full-blood python developer...

theano/sandbox/rng_mrg.py
mrTsjolder: Abergon requested support for the `target` parameter.

theano/sandbox/rng_mrg.py
mrTsjolder: @nouiz This is the review of abergeron I referred to.

theano/sandbox/rng_mrg.py
mrTsjolder: This change was introduced to improve the consistency of the code. I used the original implementation of `MRG_RandomStreams.normal` as a reference. Just to provide some more background on where this comes from.

theano/sandbox/rng_mrg.py
nouiz: This isn't the interface change that cause problem. The problem is that when you set floatX=float32, then the dtype change compared to before. In the past, but default it was always int64 and didn't looked at floatX. It should stay like that.

theano/sandbox/rng_mrg.py
nouiz: ok, but he was referencing to adding its support to truncated_normal, not change the interface here. Before you changed this code, it was an example of how to add that support.

mrTsjolder: @nouiz In that case: this used to be a `**kwargs` that could only get the keyword argument `'target'` and was not passed to any other function. I considered that, in this case, it would be cleaner to make `target` a direct keyword argument.

Should I revert this change?

theano/sandbox/rng_mrg.py
mrTsjolder: @nouiz I see, but wouldn't it make more sense to have `int32` as default when `floatX` is `float32` and `int64` when `floatX` is `float64`? E.g. for people working on 32-bit hardware...

nouiz: float32 isn't a questions of 32 bit hardware or not. You can want to use float32, but need int64 for indexing. There was a discussion to make an intX in the past. I forgot why we didn't do it, probably not enough time to get to it. But now isn't a good time to implement that.

theano/sandbox/rng_mrg.py
nouiz: It is better to have all method implemented the same. Also, the fact of passing `target=None` or not passing target don't do exactly the same thing. This is why it was done like this. We should not change the behavior here.

theano/sandbox/rng_mrg.py
nouiz: So this confirm it should not be a problem. thanks.

theano/sandbox/rng_mrg.py
```diff
@@ -940,16 +969,16 @@ def multinomial(self, size=None, n=1, pvals=None, ndim=None, dtype='int64',
...
-            unis = self.uniform(size=size, ndim=1, nstreams=nstreams, **kwargs)
+            unis = self.uniform(size=size, ndim=1, **kwargs)
             op = multinomial.MultinomialFromUniform(dtype)
             n_samples = as_tensor_variable(n)
             return op(pvals, unis, n_samples)
         else:
             raise NotImplementedError(("MRG_RandomStreams.multinomial only"
                                        " implemented for pvals.ndim = 2"))
 
-    def choice(self, size=1, a=None, replace=True, p=None, ndim=None,
```
nouiz: You changed many thing unrelated to your PR. Can you remove those commits instead of reverting all the changes? Hre, nstreams must stay and the order of parameter must not change. Those 2 changes change the user interface.

mrTsjolder: @nouiz I'm sorry

Does this mean I should also delete the binomial implementation for `n>1`?

`nstreams` is still there in some sense. It will just be passed through `kwargs`. I personally find it cleaner to use `kwargs` to simply pass keyword arguments to other functions. It also allows for easier extensibility (just add another `kwarg` to uniform and it will work). 

This being said, I will just do as you ask.

nouiz:
I removed the mostly duplicated test. We can merge when CI passes

nouiz:
jenkins test this

nouiz:
I ran the CPU random tests with the gpu enabled. This discovered one corner case in random (0 sized output) problem. I included the fix in this PR.

jenkins test this.

abergeron:
BTW you can ignore CircleCI failures for this PR.

nouiz:
jenkins test this

mrTsjolder:
@nouiz The test you deleted was not completely a duplicate. It was mainly there to assure that the standard deviation of the truncated normal is actually the required standard deviation, whereas the normal-truncation-test rather tests that the normal is truncated. I must admit that both tests are a little bit overkill.

It's not that serious, but I considered it would be better to avoid surprises.

nouiz:
jenkins test this

nouiz:
I did small modification and re-added  the test.

jenkins test this.

I think it can be merged when CI pass. thanks.

nouiz:
jenkins test this

PR #6405, opened 2017-09-12
test MRG random stream with device=cuda and target=cpu

### LINKED ISSUE #6382: test MRG random stream with device=cuda and target=cpu, shared variable + computation in a real model stay on CPU.
test MRG random stream with device=cuda and target=cpu, shared variable + computation in a real model stay on CPU.

Other stuff like the back-ward optimization pass could move the computation to the GPU. Maybe it should be used in combination to a `optimizer_excluding=...` flag to do exactly what we want.

Test it manually and after we should also make a real theano tests under gpuarray.

yikangshen:
test MRG random stream with device=cuda and target=cpu, shared variable + computation in a real model stay on CPU. fix #6382

### DIFF
--- a/theano/gpuarray/tests/config.py
+++ b/theano/gpuarray/tests/config.py
@@ -18,7 +18,7 @@ if (not theano.gpuarray.pygpu_activated and
 
 if not theano.gpuarray.pygpu_activated:
     if init_error:
-        raise SkipTest(e)
+        raise SkipTest(init_error)
     else:
         raise SkipTest("pygpu disabled")
 
--- a/theano/gpuarray/tests/config.py
+++ b/theano/gpuarray/tests/config.py
@@ -7,15 +7,20 @@ import theano.gpuarray
 if theano.gpuarray.pygpu is None:
     raise SkipTest("pygpu not installed")
 
+
+init_error = None
 if (not theano.gpuarray.pygpu_activated and
         not theano.config.force_device):
     try:
         theano.gpuarray.init_dev('cuda')
-    except Exception:
-        pass
+    except Exception as e:
+        init_error = e
 
 if not theano.gpuarray.pygpu_activated:
-    raise SkipTest("pygpu disabled")
+    if init_error:
+        raise SkipTest(e)
+    else:
+        raise SkipTest("pygpu disabled")
 
 test_ctx_name = None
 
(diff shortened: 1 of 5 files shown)


nouiz:
travis is failing: https://travis-ci.org/Theano/Theano/jobs/274457172#L1263

nouiz:
I want to include this in the RC we plan to do this week, so I pushed commit to finish this. There was stuff hard to explain how to do.

jenkins test this

thanks.

theano/gpuarray/tests/config.py
```diff
@@ -7,15 +7,20 @@
...
         theano.gpuarray.init_dev('cuda')
-    except Exception:
-        pass
+    except Exception as e:
+        init_error = e
 
 if not theano.gpuarray.pygpu_activated:
-    raise SkipTest("pygpu disabled")
+    if init_error:
+        raise SkipTest(e)
```
abergeron: This should probably be `SkipTest(init_error)`.  It happens to work because it's all in the same namespace, but that's an accident.

theano/gpuarray/tests/test_rng_mrg.py
```diff
@@ -165,4 +165,25 @@ def test_validate_input_types_gpuarray_backend():
...
+    x = gpuarray_shared_constructor(s, name='x')
+    y = srng.uniform(x.shape, target='cpu')
+    y.name = 'y'
+    z = (x * y).sum()
+    z.name = 'z'
+
+    fz = theano.function([], z, mode=mode)
+
+    nodes = fz.maker.fgraph.toposort()
+    assert not any([isinstance(node.op, GPUA_mrg_uniform) for node in nodes])
```
abergeron: This is a poor test since it doesn't run with `device=cuda` anyway.

nouiz: I force the shared variable x to be on the GPU and I passe the mode to compile with the GPU opt. So it cover our need. All GPUs tests are done like that.

But I see, for this particular tests, mimiking more device=cuda by having a chance that the state is on the GPU would make a better test. I'll update it like that. But this test was about the optimization, so it was covering the need.

I just pushed an update to the test for this.

theano/gpuarray/tests/config.py
nouiz: done

nouiz:
jenkins test this

PR #6410, opened 2017-09-13
Fix unit tests for CEnumType.

notoraptor:
@nouiz


theano/gof/tests/c_code/test_cenum.h
```diff
@@ -0,0 +1,17 @@
+#ifndef THEANO_TEST_CENUM
+#define THEANO_TEST_CENUM
+
+#define SIZE_INT 0
```
nouiz: Can you use different value then the python value?

Issue #6162, opened 2017-07-18
Memory leak in GPU backend?


f0k:
Lasagne user @guoxuesong discovered what seems to be a memory leak in the new GPU backend: Lasagne/Lasagne#860
Updating Theano and libgpuarray to bleeding-edge does not help. I have been able to reproduce the leak as follows (requires Lasagne, any version will do):
1. download the mnist example: `wget https://raw.githubusercontent.com/Lasagne/Lasagne/master/examples/mnist.py`
2. patch it with the following command (run in the directory you put mnist.py):
```bash
patch << EOF
--- mnist.py
+++ mnist.py
@@ -18,4 +18,6 @@
 import os
 import time
+import gc
+import psutil
 
 import numpy as np
@@ -295,5 +297,5 @@
         train_batches = 0
         start_time = time.time()
-        for batch in iterate_minibatches(X_train, y_train, 500, shuffle=True):
+        for batch in iterate_minibatches(X_train, y_train, 32, shuffle=True):
             inputs, targets = batch
             train_err += train_fn(inputs, targets)
@@ -318,4 +320,9 @@
         print("  validation accuracy:\t\t{:.2f} %".format(
             val_acc / val_batches * 100))
+        while gc.collect() > 0:
+            pass
+        print('  memory use:\t\t\t{:.3f} MiB'.format(
+            psutil.Process().get_memory_info()[0] / float(1024**2)))
+
 
     # After training, we compute and print the test error:
EOF
```
3. Run it with `THEANO_FLAGS=mode=FAST_RUN,device=cuda,floatX=float32,allow_gc=1,gpuarray.preallocate=0.25 python mnist_leak.py`

It will train an MLP on MNIST and print the CPU memory use after every epoch (after calling `gc.collect()` in a loop until it cannot free anything more). For a long time, it stays constant:
```
Epoch 60 of 500 took 1.519s
  training loss:		0.035076
  validation loss:		0.054378
  validation accuracy:		98.61 %
  memory use:			742.488 MiB

...

Epoch 257 of 500 took 1.518s
  training loss:		0.019314
  validation loss:		0.056173
  validation accuracy:		98.72 %
  memory use:			742.488 MiB
```
But then it raises every epoch:
```
Epoch 270 of 500 took 1.519s
  training loss:		0.017606
  validation loss:		0.062359
  validation accuracy:		98.53 %
  memory use:			743.090 MiB
Epoch 271 of 500 took 1.519s
  training loss:		0.017495
[...]

PR #6412, opened 2017-09-13
Fix a memory leak in theano

notoraptor:
Related to #6162 .

I can't yet say that every memory leak is fixed, but some of them are fixed here by calling `GpuElemwise_free()` in `c_cleanup_code_struct`.

@nouiz @abergeron

### DIFF
--- a/theano/gpuarray/basic_ops.py
+++ b/theano/gpuarray/basic_ops.py
@@ -1151,17 +1151,21 @@ class GpuReshape(HideC, tensor.Reshape):
         out[0] = x.reshape(tuple(shp))
 
     def c_code_cache_version(self):
-        return (2,)
+        return (3,)
 
     def c_code(self, node, name, inputs, outputs, sub):
         x, shape = inputs
         output, = outputs
         sdtype = node.inputs[1].type.dtype_specs()[1]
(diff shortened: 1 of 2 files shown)


theano/gpuarray/basic_ops.py
```diff
@@ -1171,7 +1175,13 @@ def c_code(self, node, name, inputs, outputs, sub):
                          "GpuReshape: given shape is of incorrect "
                          "length (%%d should be %%d).",
                          PyArray_DIM(%(shape)s, 0), %(params)s->ndim);
-            %(fail)s;
+            %(just_fail)s;
+        }
+
+        new_dims = (size_t*) malloc(sizeof(size_t) * %(params)s->ndim);
```
nouiz: Why this change? This would slow down computation as it convert a stack alloc to a malloc. What was the memory leak here?

notoraptor: Not a memory leak here, but the previous code uses syntax `size_t array[variable];` (creating an array on stack with a dynamic size). It compiles with GCC, but it is not standard. I assume we try to have a valid C/C++ code.

But I can remove these changes.

PR #6415, opened 2017-09-14
Add the setup to build conda packages

abergeron:
This also makes Theano use versioneer for its version, so that tag pushes will work.


conda/meta.yaml
```diff
@@ -5,24 +5,20 @@ package:
...
-    - m2-filesystem    [win]
-    - m2-git           [win]
-    - git              [not win]
     - python
     - setuptools
     - six >=1.9.0
     - numpy >=1.9.1
     - scipy >=0.14.0
-    - pygpu >=0.6,<0.7
+    - pygpu >=0.7
```
nouiz: should we put <0.8?

notoraptor:
@abergeron With these changes, do we need to update `How to make a release` at this point (about editing `setup.py`) : https://github.com/Theano/Theano/blob/master/doc/internal/how_to_release.txt#L28 ?

Maybe also about `theano/doc/conf.py` ? https://github.com/Theano/Theano/blob/master/doc/conf.py

conda/meta.yaml
abergeron: Actually we could remove this since we don't need anything for the build (except setuptools and python).

abergeron:
Yes we should probably update the doc.  I can also modify conf.py so that it uses versioneer to automatically pick up the right version.

abergeron:
close/reopen to trigger circleci

abergeron:
I consider this ready.

nouiz:
travis failed: https://travis-ci.org/Theano/Theano/jobs/275659391#L1263

theano/tests/test_flake8.py
```diff
@@ -34,8 +34,8 @@
           'FI50', 'FI51', 'FI53')
 
 whitelist_flake8 = [
-    "compat/six.py",  # This is bundled code that will be deleted, don't fix it
     "__init__.py",
+    "_version.py",  # This is generated by versioneer
```
nouiz: I don't get it. This file is included in the repo, so I don't understand why it would be generated by versioneer

abergeron: verisoneer is a script that you run on your project and it generates a couple of files and installs them.  After that if you want to update you re-run the script with the new version and it will regenerate the files.

If we do local modifications, we will loose those modifications every update, which is a pain in the ass.

PR #6416, opened 2017-09-14
Gpu stack follow up

ReyhaneAskari:
Follow up for https://github.com/Theano/Theano/pull/5688.

### DIFF
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -97,7 +97,6 @@ class Apply(Node):
         self.op = op
         self.inputs = []
         self.tag = utils.scratchpad()
-        utils.add_tag_trace(self)
 
         if not isinstance(inputs, (list, tuple)):
             raise TypeError("The inputs of an Apply must be a list or tuple")
@@ -118,7 +117,8 @@ class Apply(Node):
                 if output.owner is None:
                     output.owner = self
                     output.index = i
-                    output.tag = utils.scratchpad()
+                    if not hasattr(output, "tag"):
+                        output.tag = utils.scratchpad()
                     utils.add_tag_trace(output)
                 elif output.owner is not self or output.index != i:
                     raise ValueError("All output variables passed to Apply must belong to it.")
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -117,9 +117,6 @@ class Apply(Node):
                 if output.owner is None:
                     output.owner = self
                     output.index = i
-                    if not hasattr(output, "tag"):
-                        output.tag = utils.scratchpad()
-                    utils.add_tag_trace(output)
                 elif output.owner is not self or output.index != i:
                     raise ValueError("All output variables passed to Apply must belong to it.")
                 self.outputs.append(output)
--- a/theano/gof/graph.py
+++ b/theano/gof/graph.py
@@ -97,6 +97,7 @@ class Apply(Node):
         self.op = op
         self.inputs = []
         self.tag = utils.scratchpad()
+        utils.add_tag_trace(self)
 
         if not isinstance(inputs, (list, tuple)):
             raise TypeError("The inputs of an Apply must be a list or tuple")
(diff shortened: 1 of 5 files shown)


theano/gpuarray/tests/test_opt.py
```diff
@@ -338,7 +337,9 @@ def test_print_op():
     assert isinstance(topo[1].op, theano.printing.Print)
     assert isinstance(topo[2].op, GpuElemwise)
     assert topo[3].op == host_from_gpu
-    assert _check_stack_trace(f)
+    # gpu print_op copies the stack trace
+    # but _print_fn has an empty stack.
+    # assert _check_stack_trace(f)
```
nouiz: Why do you remove it? The test was passing and your commit just make us keep more stack trace. So it should not start failing.

ReyhaneAskari: So I fixed `local_gpua_print_op` to copy the stack trace but since it's previous node is a node with an empty stack trace it copies nothing and `_check_stack_trace` still fails. I can keep the assertion but I will also need to bring `theano.printing.Print` to the `_ops_to_check` list.

nouiz: Can you try to make sure the previous node is also fixed in this PR to have the pipeline work? This way you can re-add this assert and still check this op.

ReyhaneAskari: Yeah sure, I was trying to first focus on the gpu ops. But I can do this.

theano/gof/graph.py
```diff
@@ -97,6 +97,7 @@ def __init__(self, op, inputs, outputs):
         self.op = op
         self.inputs = []
         self.tag = utils.scratchpad()
+        utils.add_tag_trace(self)
```
nouiz: I don't think we should put stack trace on the Apply node and Variables. This cause duplicate stack traces. Why did you added it here?

ReyhaneAskari: I was trying to check multiple places for the print op. Now I removed the other places.

theano/gof/graph.py
```diff
@@ -118,7 +117,8 @@ def __init__(self, op, inputs, outputs):
                 if output.owner is None:
                     output.owner = self
                     output.index = i
-                    output.tag = utils.scratchpad()
+                    if not hasattr(output, "tag"):
+                        output.tag = utils.scratchpad()
```
nouiz: Why not remove those 2 lines? Why they are needed?

I'm ok with the change, but I don't understand why it was like this first.

theano/gof/graph.py
```diff
@@ -117,6 +117,9 @@ def __init__(self, op, inputs, outputs):
                 if output.owner is None:
                     output.owner = self
                     output.index = i
+                    if not hasattr(output, "tag"):
+                        output.tag = utils.scratchpad()
```
nouiz: While thinking about this, I think Theano variable should always have a tag attribute. I would like to understand when they don't have it and fix the root cause if possible.

Do you know more about when this case happen? Unpickle of old graph?

theano/gof/graph.py
```diff
@@ -117,6 +117,9 @@ def __init__(self, op, inputs, outputs):
...
+                    utils.add_tag_trace(output)
```
nouiz: Why this line is needed? We already do this at many other places. It should be done for all Constant I think. It not, tell me. For shared variable, I found a problem in our logic. Currently, we set it only when we call theano.shared(), but not when we call directly the type specific shared variable constructor like gpuarray_shared_variable as done in many tests.

Is that the case you found? If so, I don't think this is the good fix. This will add the stack-trace as of when we use the shared variable, not when we create it. I think the fix should be to call add_tag_trace in SharedVariable.__init__() and remove it from theano/compile/sharedvalue.py:shared()

ReyhaneAskari:
I thought Theano variable should always have a tag attribute too and what I was doing was mostly copying the stack trace. But in some cases it seemed like the original node did not have the stack trace. It's not an unpickle or anything. It was a simple print node that originally had that problem.

 I followed it's creation and there were many places that the nodes were missing the stack trace and they weren't the output of any other node (to copy the stack trace from that). So in my 3rd commit, I  added the stack trace to all those places but you told me that we might have duplications so I removed all those places except the one that was fixing the print node. 

I'm going yo check for the shared variables.

nouiz:
@abergeron can we ignore CircleCI error? Is that normal?

abergeron:
We can ignore Circle CI since it runs without the config from master.

Issue #6423, opened 2017-09-17
UndefinedGrad wrapping of pvals in mrg multinomial results in GPU->CPU->GPU transfer


amurashov:
Hello all.

With special regards to this post: https://github.com/Theano/Theano/issues/6408

Wrapping mentioned there (and in topic of this issue) results in following graph in the function - which is, probably, no what was intended:
![gpucpugpu](https://user-images.githubusercontent.com/32039151/30523274-fe8d13a2-9bde-11e7-848e-c6d712683868.JPG)

nouiz:
This code generate the problem with device=cuda

```
import numpy as np
import theano
import theano.tensor as T
from theano.gradient import zero_grad
from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
random = RandomStreams(np.random.randint(1, 2147462579))
pvals = theano.shared(np.random.rand(10, 20).astype(theano.config.floatX))
pvals = pvals / pvals.sum(axis=1)
pvals = zero_grad(pvals)
samples = random.multinomial(pvals=pvals, n=1)
samples = T.cast(samples, pvals.dtype)
samples = zero_grad(samples)
cost = T.sum(samples + pvals)
grad = T.grad(cost, samples)
f=theano.function([], grad)
theano.printing.debugprint(f)
```

nouiz:
Fix here:

https://github.com/Theano/Theano/pull/6425

We will remove that node from the graph completely.

PR #6425, opened 2017-09-18
Remove ViewOp subclass during optimization

nouiz:
fix gh-6423

### DIFF
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1545,11 +1545,9 @@ class OpRemove(LocalOptimizer):
         return [self.op]
 
     def transform(self, node):
-        if inspect.isclass(self.op):
-            if isinstance(node.op, self.op):
-                return node.inputs
-        elif node.op == self.op:
-            return node.inputs
+        if node.op != self.op:
+            return False
+        return node.inputs
 
     def __str__(self):
         return "%s(x) -> x" % (self.op)
--- a/theano/gof/opt.py
+++ b/theano/gof/opt.py
@@ -1333,9 +1333,12 @@ class LocalOptGroup(LocalOptimizer):
                 self.process_count.setdefault(o, 0)
                 self.applied_true.setdefault(o, 0)
                 self.node_created.setdefault(o, 0)
-
-            for c in o.tracks():
-                self.track_map[c].append(o)
+            tracks = o.tracks()
+            if tracks is None:
+                self.track_map[None].append(o)
+            else:
+                for c in tracks:
+                    self.track_map[c].append(o)
 
     def __str__(self):
         return getattr(self, '__name__',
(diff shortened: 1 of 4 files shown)


theano/gof/opt.py
```diff
@@ -1542,9 +1542,11 @@ def tracks(self):
...
 
     def transform(self, node):
-        if node.op != self.op:
-            return False
-        return node.inputs
+        if inspect.isclass(self.op):
+            if isinstance(node.op, self.op):
+                return node.inputs
+        elif node.op == self.op:
+            return node.inputs
```
notoraptor: Maybe should this method return `False` by default (if these both checkings failed) ?

nouiz: By default, all python method return "None" if there is no return statement. Theano will confider this equivalent to False. So it is not requested to return explicitly False. This is done in many optimization already, so is a well tested behavior.

notoraptor: Ok !

theano/tensor/opt.py
```diff
@@ -7490,23 +7490,20 @@ def local_useless_composite(node):
...
+                (gof.OpRemove(theano.gradient.disconnected_grad_),
+                 'remove_disconnected_grad'),
+                (gof.OpRemove(theano.gradient.undefined_grad_),
+                 'remove_undefined_grad'),
+                (gof.OpRemove(theano.gradient.GradClip),
+                 'remove_grad_clip'),
+                (gof.OpRemove(theano.gradient.GradScale),
+                 'remove_grad_scale')]:
+    register_canonicalize(o, 'fast_compile', 'fast_run', name=name)
+    register_useless(o, 'fast_compile', 'fast_run', name=name)
```
abergeron: Isn't the point of this to avoid this listing and just have one opt working on ViewOp?

theano/gof/opt.py
abergeron: Could this affect things other than ViewOp?  I'm not comfortable making this change close to the release.  However it can go in after.

abergeron:
I don't think we should merge this right now.

nouiz:
IRL summary:
- Make 1 opt that check with isinstance of ViewOp
- Remove change to OpRemove

nouiz:
Done

theano/tensor/opt.py
```diff
@@ -7487,26 +7487,14 @@ def local_useless_composite(node):
...
-@register_canonicalize
-@gof.local_optimizer([theano.gradient.GradClip])
-def local_grad_clip(node):
-    if isinstance(node.op, theano.gradient.GradClip):
-        return node.inputs
+@register_canonicalize('fast_compile')
+@register_useless('fast_compile')
+@gof.local_optimizer(None)
+def local_view_op(node):
+    if isinstance(node.op, theano.compile.ops.ViewOp):
```
abergeron: This is currently a syntax error.

abergeron:
jenkins failed.

nouiz:
jenkins failure was in master and is fixed in another PR, so merging.

Issue #6426, opened 2017-09-18
Conda package does not install Theano scripts on Windows


notoraptor:
Conda package `theano` from `mila-udem` does not install working theano scripts like `theano-cache` on Windows:

```
(testConda2) C:\Users\notoraptor>conda install -c mila-udem/label/pre theano
Fetching package metadata ...............
Solving package specifications: .

Package plan for installation in environment C:\Users\notoraptor\data\programs\Anaconda3\envs\testConda2:

The following NEW packages will be INSTALLED:

    theano: 0.10.0beta2-pyha8067b5_0 mila-udem/label/pre

Proceed ([y]/n)?


(testConda2) C:\Users\notoraptor>python -c "import theano; print(theano.__version__)"
0.10.0beta2.dev-f6f45e57a0151fa91b04de60b27f01eceaa0c17a

(testConda2) C:\Users\notoraptor>theano-cache
'theano-cache' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.

(testConda2) C:\Users\notoraptor>where theano-cache
C:\Users\notoraptor\data\programs\Anaconda3\envs\testConda2\Scripts\theano-cache
```

Instead, it installs a file `theano-cache` which is just a bash script test, that cannot run on Windows.

Everything is ok when I install from local repository (`pip install -e .`), so I guess the problem comes from configuration for conda package generation.

@nouiz @abergeron

notoraptor:
PS: On Windows, we expect an executable file to be generated for each script (ex. `theano-cache.exe` for `theano-cache`).

abergeron:
Fixed by #6427

PR #6427, opened 2017-09-18
Fix conda package builds

abergeron:
noarch: python is broken according to continuum so do it the hard way with multiple packages.


abergeron:
Since none of the actual code changed, this can go in.

nouiz:
Does this fix https://github.com/Theano/Theano/issues/6426?

abergeron:
Yes

PR #6431, opened 2017-09-19
Generalised AllocDiag for more than 2D input

shawntan:
#5633 @ReyhaneAskari 

Starting off the pull request.

### DIFF
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -1394,22 +1394,18 @@ class GpuAllocDiag(AllocDiag):
                                        context=x.context)
 
         # Slice out a view of the diagonals
-        row_size = abs(offset) + x.shape[-1]
-        if offset <= 0:  # diag in the lower triangle
-            diag_view = result_buffer[:, abs(offset), :x.shape[-1]]
-            diag_view.strides = (diag_view.strides[0],
-                                 (row_size + 1) * x.dtype.itemsize)
+        if offset < 0:  # diag in the lower triangle
+            diag_view = result_buffer[:, abs(offset):, 0]
         else:  # diag in the upper triangle
             diag_view = result_buffer[:, :x.shape[-1], abs(offset)]
-            diag_view.strides = (diag_view.strides[0],
-                                 diag_view.strides[1] + x.dtype.itemsize)
+        diag_view.strides = (diag_view.strides[0],
+                             diag_view.strides[1] + x.dtype.itemsize)
 
         # Fill view with flattened array of diagonals
         diag_view[:] = x.reshape(diag_view.shape)[:]
 
         # Unflatten buffer into output size
         result = result_buffer.reshape(result_shape)
-        print(result)
 
         if len(x.shape) > 1:
             # Re-order axes so they correspond to diagonals at axis1, axis2
--- a/theano/gpuarray/subtensor.py
+++ b/theano/gpuarray/subtensor.py
@@ -1357,34 +1357,71 @@ class GpuExtractDiag(Op):
 
 
 class GpuAllocDiag(Op):
-    __props__ = ("offset",)
+    __props__ = ("offset", "axis1", "axis2")
 
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, axis1=0, axis2=1):
         self.offset = offset
+        self.axis1 = axis1
+        self.axis2 = axis2
 
-    def make_node(self, _x):
-        ctx_name = infer_context_name(_x)
-        x = as_gpuarray_variable(_x, ctx_name)
-
-        if x.ndim != 1:
-            raise ValueError('AllocDiag argument must be a vector!', x)
-
-        return gof.Apply(self, [x], [x.type.clone(broadcastable=(False, False))()])
+    def make_node(self, diag):
+        ctx_name = infer_context_name(diag)
+        diag = as_gpuarray_variable(diag, ctx_name)
+        if diag.type.ndim < 1:
+            raise ValueError('AllocDiag needs an input with 1 or more '
+                             'dimensions', diag.type)
+        return gof.Apply(
+            self, [diag],
+            [diag.type.__class__(
+                dtype=diag.dtype,
+                broadcastable=[False] * (diag.ndim + 1))()]
+        )
 
     def perform(self, node, inputs, outputs):
         (x,) = inputs
         (z,) = outputs
+        axis1 = np.minimum(self.axis1, self.axis2)
+        axis2 = np.maximum(self.axis1, self.axis2)
+        offset = self.offset
 
-        dim = x.shape[0] + abs(self.offset)
-        z[0] = gpuarray.zeros((dim, dim), dtype=x.dtype, context=x.context)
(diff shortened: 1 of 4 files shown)


shawntan:
Right now there is `theano/tensor/tests/test_basic.py:test_diag` which tests the diag function, but doesn't explicitly test `AllocDiag`. I'm gonna write a separate test for `AllocDiag`, and we can move that later if need be.

theano/tensor/basic.py
```diff
@@ -6520,30 +6520,74 @@ class AllocDiag(Op):
 
     """
 
-    __props__ = ("offset", )
-    default_offset = 0
+    __props__ = ("offset", "axis1", "axis2")
 
-    def __init__(self, offset=0):
-        if numpy_diagonal_return_view:
```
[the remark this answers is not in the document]
nouiz: It depend how old. We still support numpy 1.9.1. I prefer to don't drop version now. Can you just keep it? If you know 1.9.1 is recent enough, you can remove support of older numpy.

theano/tensor/basic.py
```diff
@@ -6520,30 +6520,74 @@ class AllocDiag(Op):
...
+        (x,) = inputs
         (z,) = outputs
-        z[0] = np.diag(inputs[0], self.offset)
+
+        axis1 = np.minimum(self.axis1, self.axis2)
+        axis2 = np.maximum(self.axis1, self.axis2)
+        offset = self.offset
+
+        # Create array with one extra dimension for resulting matrix
+        result_shape = x.shape[:-1] + (x.shape[-1] + offset,) * 2
```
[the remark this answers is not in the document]
shawntan: Oh yeah, stupid mistake.

ReyhaneAskari:
@shawntan could you also add a test for `infer_shape`?

theano/tensor/basic.py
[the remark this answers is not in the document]
shawntan: That would mean checking if the input is a vector and using `np.diag` right?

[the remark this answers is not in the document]
ReyhaneAskari: Check this line: https://github.com/shawntan/Theano/blob/e67873bbd4ce74055ab453bda142c8c023df7d63/theano/tensor/basic.py#L6342. 

I guess you would just need to keep the `if numpy_diagonal_return_view:`

nouiz:
travis have an error.

ReyhaneAskari:
@shawntan the failing test is this one:  THEANO_FLAGS=mode=FAST_COMPILE nosetests -s theano/tensor/tests/test_basic.py:test_alloc_diag

shawntan:
How do I force a `fast_run`? My test relies on the optimisation compiling.

theano/gpuarray/subtensor.py
```diff
@@ -1350,34 +1350,71 @@ def infer_shape(self, node, shapes):
...
+                end_flattened_offset = row_size * abs(offset)
+            else:
+                start_flattened_offset = row_size * abs(offset)
+                end_flattened_offset = abs(offset)
+
+            diag_view = result_buffer[:, start_flattened_offset:-end_flattened_offset:row_size + 1]
+            # print("offset", offset)
+            # print("buffer shape:", result_buffer.shape)
+            # print("result_buffer[%d:%d:%d]" % (start_flattened_offset, -end_flattened_offset, row_size + 1), diag_view.shape)
+            # print("input_shape:", x.shape)
```
ReyhaneAskari: Don't forget to remove the prints here.

theano/gpuarray/subtensor.py
```diff
@@ -1350,34 +1350,71 @@ def infer_shape(self, node, shapes):
...
+        else:
+            diag_view = result_buffer[:, ::x.shape[-1] + 1]
+
+        diag_view[:] = x.reshape(diag_view.shape)[:]
+
+        result = result_buffer.reshape(result_shape)
+        # print(result)
```
ReyhaneAskari: same.

theano/gpuarray/subtensor.py
```diff
@@ -1349,40 +1350,66 @@ def infer_shape(self, node, shapes):
...
-        diag_z.strides = (sum(z[0].strides),)
+            diag_view = result_buffer[:, :x.shape[-1], abs(offset)]
+        diag_view.strides = (diag_view.strides[0],
+                             diag_view.strides[1] + x.dtype.itemsize)
+
+        # Fill view with flattened array of diagonals
+        diag_view[:] = x.reshape(diag_view.shape)[:]
+
+        # Unflatten buffer into output size
+        result = result_buffer.reshape(result_shape)
```
ReyhaneAskari: The result will still be all zero, you are never using `x` to fill it.

theano/tensor/tests/test_basic.py
```diff
@@ -7561,6 +7561,88 @@ def test_diag_grad(self):
         tensor.verify_grad(diag, [x], rng=rng)
 
 
+class test_alloc_diag(unittest.TestCase):
+    def __init__(self, name, alloc_diag=AllocDiag, mode=None):
+        self.alloc_diag = AllocDiag
```
ReyhaneAskari: These tests are always running the cpu version because here you should set it like this:

```python
self.alloc_diag = alloc_diag

```

ReyhaneAskari: Please rename the class to capitals: `TestAllocDiag`

theano/gpuarray/subtensor.py
shawntan: `diag_view` is a view into `result_buffer`, so filling `diag_view` fills the relevant elements in `result_buffer`

ReyhaneAskari: Oh! that's not the case in numpy, so I was confused.

ReyhaneAskari:
Thanks @shawntan . I think it's good to merge.

theano/gpuarray/subtensor.py
```diff
@@ -1353,40 +1354,66 @@ def infer_shape(self, node, shapes):
         return [tuple(out_shape)]
 
 
-class GpuAllocDiag(Op):
-    __props__ = ("offset",)
+class GpuAllocDiag(AllocDiag):
+    __props__ = ("offset", "axis1", "axis2")
 
-    def __init__(self, offset=0):
+    def __init__(self, offset=0, axis1=0, axis2=1):
```
nouiz: Remove, as this is the same as the parent class

theano/gpuarray/subtensor.py
```diff
@@ -1353,40 +1354,66 @@ def infer_shape(self, node, shapes):
...
+        result = result_buffer.reshape(result_shape)
 
-        diag_z[:] = x[:]
+        if len(x.shape) > 1:
+            # Re-order axes so they correspond to diagonals at axis1, axis2
+            axes = list(range(len(x.shape[:-1])))
+            last_idx = axes[-1]
+            axes = axes[:axis1] + [last_idx + 1] + axes[axis1:]
+            axes = axes[:axis2] + [last_idx + 2] + axes[axis2:]
+            result = result.transpose(axes)
```
nouiz: Just a comment, so This mean this is probably the only op where the output is allocated and isn't c contiguous. No objection, we should support that well. But this isn't frequent.

PR #6440, opened 2017-09-23
Fix memory leak in fallback sdot

notoraptor:
I forgot to add an extra call to `Py_XDECREF()` after calling `PyArray_MatrixProduct2()`.

@nouiz @abergeron

### DIFF
--- a/theano/tensor/c_code/alt_blas_template.c
+++ b/theano/tensor/c_code/alt_blas_template.c
@@ -307,6 +307,8 @@ void %(precision)sgemv_(
     PyArray_MatrixProduct2(vector_x, vector_y, dot_product);
     if (PyErr_Occurred())
         alt_fatal_error("NumPy %(precision)sdot_ implementation: unable to compute dot.");
+    // PyArray_MatrixProduct2 adds an extra reference to the output array.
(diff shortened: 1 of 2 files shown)


theano/tensor/c_code/alt_blas_template.c
```diff
@@ -307,6 +307,8 @@ void %(precision)sgemv_(
     PyArray_MatrixProduct2(vector_x, vector_y, dot_product);
     if (PyErr_Occurred())
         alt_fatal_error("NumPy %(precision)sdot_ implementation: unable to compute dot.");
+    // PyArray_MatrixProduct2 adds an extra reference to the output array.
+    Py_XDECREF(dot_product);
```
abergeron: It's there like two lines below this.

notoraptor:
@abergeron I made the code less confusing.

`PyArray_MatrixProduct2` increases ref count of output, so we need to decrement output ref count at least once after calling this function, whatever we do later.

nouiz:
@notoraptor I think numpy already incref what it return. I'm pretty sure this is because PyArray_MatrixProduct2 return a ptr of what you decref.

PR #6444, opened 2017-09-25
Fix PushOutputScan when using trucate_gradient.

### LINKED ISSUE #4652: Scan: truncated gradient bug
A Lasagne user found a bug related to the gradient of theano.scan: https://groups.google.com/d/msg/lasagne-users/Dzkp4szn0Vk/3S-pYDIsBAAJ

I've simplified his example to use pure Theano:
```python
import numpy as np
import theano
import theano.tensor as T
floatX = theano.config.floatX

num_seqs, len_seqs, num_in = (4, 10, 2)
num_hid = 5

nonlinearity = lambda X: X  # causes the error
#nonlinearity = lambda X: 1*X  # does not cause the error
gsteps = 3  # causes the error
#gsteps = -1  # does not cause the error


# input layer
x = theano.tensor.tensor3('x')

# recurrent layer
w_inhid = theano.shared(np.random.randn(num_in, num_hid).astype(floatX))
w_hidhid = theano.shared(np.random.randn(num_hid, num_hid).astype(floatX))
hid_init = theano.shared(np.zeros((num_seqs, num_hid), floatX))
y, _ = theano.scan(
        lambda i, h: nonlinearity(i + T.dot(h, w_hidhid)),
        [T.dot(x, w_inhid)],  # causes the error
        #[x],  # does not cause the error (with T.dot(i, w_inhid) inside scan)
        outputs_info=hid_init,
        truncate_gradient=gsteps)

# gradient of output wrt recurrent weights
grad = theano.grad(y.sum(), w_hidhid)  # causes the error
#grad = theano.grad(y.sum(), w_inhid)  # does not cause the error

# trigger the error
fn = theano.function([x], grad)
fn(np.random.randn(len_seqs, num_seqs, num_in).astype(floatX))
```

When running this (no matter if on CPU or GPU), I get:
```
  File "bug2.py", line 35, in <module>
    fn(np.random.randn(len_seqs, num_seqs, num_in).astype(floatX))
  File "/raid/user/jan/install/Theano-git/theano/compile/function_module.py", line 875, in __call__
  File "/raid/user/jan/install/Theano-git/theano/gof/link.py", line 317, in raise_with_op
  File "/raid/user/jan/install/Theano-git/theano/compile/function_module.py", line 862, in __call__
ValueError: Shape mismatch: x has 12 cols (and 5 rows) but y has 40 rows (and 5 cols)
Apply node that caused the error: Dot22(Reshape{2}.0, Reshape{2}.0)
Toposort index: 80
Inputs types: [TensorType(float32, matrix), TensorType(float32, matrix)]
Inputs shapes: [(5, 12), (40, 5)]
Inputs strides: [(48, 4), (20, 4)]
Inputs values: ['not shown', 'not shown']
Outputs clients: [[Assert{msg='Theano Assert failed!'}(Dot22.0, Elemwise{eq,no_inplace}.0, Elemwise{eq,no_inplace}.0)]]
```

The error disappears with either of the following changes (indicated in the code above):
* Adding an output nonlinearity, even if it is just 1 * X
* Setting `truncate_gradient=-1`
* Moving the dot product of input and `w_inhid` into the scan loop instead of performing it outside
* Computing the gradient wrt. `w_inhid` instead of `w_hidhid` (well, this is kind of obvious, since it's outside scan)

Thrandis:
Fixes #4652

@nouiz is it ok like that?

### DIFF
--- a/theano/scan_module/scan_opt.py
+++ b/theano/scan_module/scan_opt.py
@@ -744,6 +744,18 @@ class PushOutScanOutput(gof.Optimizer):
                             self.push_out_inner_vars(fgraph, inner_dot_inputs,
                                                      node, args)
 
+                        # If truncate_gradient is used, then some sequences
+                        # don't have the same length as args.n_steps. Thus,
+                        # it causes a shape mismatch in the dot product
+                        # pushed outside of scan. Simply cutting the sequences
+                        # breaks scanOp_save_mem. Since saving memory is the
+                        # main reason for using truncate_gradient, we don't
+                        # apply the push when truncate_gradient is used.
+                        if outer_dot_inputs[0].shape[0] != args.n_steps:
+                            return None
+                        if outer_dot_inputs[1].shape[0] != args.n_steps:
+                            return None
+
                         # Collapse some of the dimensions of the tensors
                         # so that they become matrices. This is because a
                         # dot is usually faster on two large matrices than
--- a/theano/scan_module/tests/test_scan_opt.py
+++ b/theano/scan_module/tests/test_scan_opt.py
@@ -418,3 +418,30 @@ class TestPushOutSumOfDot():
         output_no_opt = f_no_opt(input1_value, input2_value, input3_value)
 
         utt.assert_allclose(output_opt, output_no_opt)
+
+
+class TestPushOutputTruncatedGradient(unittest.TestCase):
+
+    def test_truncated_gradient(self):
+        theano.config.compute_test_value = 'warn'
+        nin = 2
+        nhu = 5
+
+        x = theano.tensor.tensor3('x')
+        x.tag.test_value = np.random.randn(10, 4, nin)
+        Wxh = theano.shared(np.random.randn(nin, nhu).astype(config.floatX),
+                            name='Wxh')
+        Whh = theano.shared(np.random.randn(nhu, nhu).astype(config.floatX),
+                            name='Whh')
+        h0 = theano.shared(np.zeros((4, nhu), config.floatX), name='h0')
+        seq = T.dot(x, Wxh)
+
+        y, _ = theano.scan(fn=lambda i, h: i + T.dot(h, Whh),
+                           sequences=[seq],
+                           outputs_info=h0,
+                           truncate_gradient=3)
+        grad = theano.grad(y.sum(), Whh)
+        fn = theano.function([x], grad)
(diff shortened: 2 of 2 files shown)


theano/scan_module/scan_opt.py
```diff
@@ -744,6 +744,18 @@ def process_node(self, fgraph, node):
...
                                                      node, args)
 
+                        # If truncate_gradient is used, then some sequences
+                        # don't have the same length as args.n_steps. Thus,
+                        # it causes a shape mismatch in the dot product
+                        # pushed outside of scan. Simply cutting the sequences
+                        # breaks scanOp_save_mem. Since saving memory is the
+                        # main reason for using truncate_gradient, we don't
+                        # apply the push when truncate_gradient is used.
+                        if outer_dot_inputs[0].shape[0] != args.n_steps:
```
nouiz: This won't work .shape[0] will always return new Theano variable. So you end up always disabling this optimization.

What about just doing something like:
```
if node.op.truncate_gradient != -1:
    return
```

Thrandis: @nouiz: The thing is that, if I'm not mistaken, when you set `truncate_gradient`, you set it for both scans (the scan of the forward computation, and the scan for the gradient computation). I was thinking that during the forward computation, we won't have the shape problem, so we can apply this optimization. It is only during the gradient computation that we might have some issues.

So it would be nice to detect if the scan is computing the forward computation (in which case we apply the opt) or in the gradient computation (in which case we don't apply it). Is there a way to know which one is which?

nouiz: Can check your if there is mitmot? If so, it is sure this is a grad. Otherwise, it isn't sure what it is. If that is enough, then that would be great.

Thrandis:
@nouiz Can you take a look to see what is the new problem?

nouiz:
I think that now I understand the root of the problem.

The runtime, expect all inputs to have the same lenght. It wasn't the case in the past, but this caused user to have code that didn't execute what it expected and made the code complicated. So we decided to force each each scan input that we iterate over to have the same lenght.

So this probably mean in Scan.grad (or Scan.L_op), we create the grad scan but don't add the subtensor on the inputs shared by the forward and back-ward scan when truncate_gradient is used. So we would need to first add this. Then maybe with only this fix, the rest of the current PR won't be needed. But I'm not sure of that.

PR #6445, opened 2017-09-26
buildbot release test without scipy

slefrancois:
Dockers have been modified to add a no_scipy conda env.


nouiz:
How we activate those tests?

nouiz:
merging now as this only change CI config.

PR #6446, opened 2017-09-26
Fix opt crash and crash/tests when SciPy is not there.

nouiz:


theano/gpuarray/opt_util.py
```diff
@@ -273,7 +271,9 @@ def opt(node):
                     return None
                 inputs = list(targ.inputs)
                 inputs[out_in] = W
-                inputs[beta_in] = _one.clone()
+                dtype = inputs[beta_in].dtype
```
notoraptor: Does `inputs[beta_in]` always exist ?

nouiz: Yes

notoraptor: Ok.

theano/tensor/tests/test_mpi.py
```diff
@@ -45,15 +46,21 @@ def test_mpi_roundtrip():
     if not mpi_enabled:
         raise SkipTest('MPI not enabled')
     theano_root = theano.__file__.split('__init__')[0]
+    d = {}
+    if PY3:
+        # Is some not understood cases, the subprocess never finish.
```
notoraptor: As you have now added `p.communicate()` below, does the subprocess still not finish ?

nouiz: I don't know. We only get that in the daily buildbot. I prefer to fix all the cases we want.

Also, jenkins got broken due to system update... So I'm not sure if futur case where it pass would have this fixed, as I'm not sure the new install will reproduce the same error.

notoraptor: ok

PR #6447, opened 2017-09-26
correct gradient for ARange

### LINKED ISSUE #5790: Return correct gradient in ARange with float dtype (and floatX-typed gradient when dtype is discrete)
Currently, the [implementation](https://github.com/Theano/Theano/blob/372ae95ef9aae402b7f426d159585d26556eb0a1/theano/tensor/basic.py#L5395-L5397) re-uses the dtype of the inptus, which is not correct.

I think there is a helper function to get the right dtype (floatX if dtype in discrete_dtypes, else dtype) somewhere.

Faruk-Ahmed:
fix #5790

### DIFF
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5475,6 +5475,7 @@ class ARange(Op):
 
         inputs = [start, stop, step]
         outputs = [tensor(self.dtype, (False,))]
+
         return Apply(self, inputs, outputs)
 
     @theano.configparser.change_flags(warn_float64='ignore')
@@ -5534,9 +5535,16 @@ class ARange(Op):
         # no gradient through them
         # stop does not affect the output values,
         # just the output shape, so it is disconnected
-        return [start.zeros_like(),
-                DisconnectedType()(),
-                step.zeros_like()]
+
+        if (self.dtype in discrete_dtypes) and (step > 0):
+            return [start.zeros_like(),
+                    DisconnectedType()(),
+                    step.zeros_like()]
+        else:
+            num_steps_taken = (stop-start)/step
+            return [gz.sum(dtype=config.floatX),
+                    stop.zeros_like(dtype=config.floatX),
+                    (gz*arange(num_steps_taken+1)).sum(dtype=config.floatX)]
 
     def R_op(self, inputs, eval_points):
         return [None]
--- a/theano/tensor/basic.py
+++ b/theano/tensor/basic.py
@@ -5527,7 +5527,7 @@ class ARange(Op):
 
         return [[True], [False], [True]]
 
-    def grad(self, inputs, grads):
+    def L_op(self, inputs, outputs, grads):
         start, stop, step = inputs
         gz, = grads
         # start and step affect the output values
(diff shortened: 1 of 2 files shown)


nouiz:
Some travis failure due to flake8:

https://travis-ci.org/Theano/Theano/jobs/280177905#L1265

nouiz:
While investigating the error you told me in your tests, I ended up finishing it. The time consuming part was the investigation, not the code change.

But this change what was said in the original issue, so it will need review by @abergeron or @lamblin

theano/tensor/basic.py
```diff
@@ -5526,17 +5527,25 @@ def connection_pattern(self, node):
...
-        return [start.zeros_like(),
-                DisconnectedType()(),
-                step.zeros_like()]
+
+        if (self.dtype in discrete_dtypes) and (step > 0):
+            return [start.zeros_like(),
+                    DisconnectedType()(),
+                    step.zeros_like()]
+        else:
+            num_steps_taken = outputs[0].shape[0]  # (stop-start)/step
```
lamblin: I realize that "number of steps" is ambiguous in that case. In my mind, if the shape was 2 for instance, it meant there was 1 step taken from the starting point.
The comment is not really accurate either: if for instance `step > 0` but `start > stop`, then `outputs[0].shape[0]` will be `1` instead.

theano/tensor/basic.py
```diff
@@ -5526,17 +5527,25 @@ def connection_pattern(self, node):
...
+            return [gz.sum(dtype=config.floatX),
```
lamblin: Not `floatX`, if `start.dtype` is float64, for instance, we should keep `gz`'s dtype.

theano/tensor/basic.py
```diff
@@ -5526,17 +5527,25 @@ def connection_pattern(self, node):
...
+                    DisconnectedType()(),
```
lamblin: OK for `DisconnectedType` rather than 0.

theano/tensor/basic.py
```diff
@@ -5526,17 +5527,25 @@ def connection_pattern(self, node):
...
+                    (gz * arange(num_steps_taken, dtype=self.dtype)).sum(
+                        dtype=config.floatX)]
```
lamblin: Same thing for the dtype here, we should use `floatX` only if `stop.dtype` is discrete.

theano/tensor/tests/test_basic.py
```diff
@@ -5911,6 +5911,20 @@ def test_Op_integers(self):
         assert np.all(f(10, 2, 2) == np.arange(10, 2, 2))
         assert np.all(f(0, 0, 1) == np.arange(0, 0, 1))
 
+    def test_grads(self):
+        def f(start, stop, step):
+            return ARange(start.type.dtype)(start, stop, step)
+
+        rng = np.random.RandomState(utt.fetch_seed())
+        # Due to the random projection, we should not use the exact
+        # point that shage the shape of the output.
```
lamblin: "change"?

theano/tensor/basic.py
```diff
@@ -5526,17 +5527,25 @@ def connection_pattern(self, node):
 
         return [[True], [False], [True]]
 
-    def grad(self, inputs, grads):
+    def L_op(self, inputs, outputs, grads):
         start, stop, step = inputs
         gz, = grads
         # start and step affect the output values
         # but the outputs are integers so there's
         # no gradient through them
```
lamblin: Update the comment, the point of that PR is that it is not always the case.

theano/tensor/tests/test_basic.py
```diff
@@ -5911,6 +5911,20 @@ def test_Op_integers(self):
...
+        for start, stop, step in [(0, 4.9, 1),
+                                  (1, 5.1, 0.5)]:
+            utt.verify_grad(f, [np.float32(start),
+                                np.float32(stop),
+                                np.float32(step)],
```
lamblin: Use floatX in the test, not float32. This would have caught the issue with the dtype.

Faruk-Ahmed:
continuing on in #6452

Issue #6399, opened 2017-09-09
CTC failes with omp_get_num_threads’ was not declared in this scope.


botev:
The following code:
```
import theano
import theano.tensor as T
from theano.tensor.nnet.ctc import ctc
a = T.ftensor3()
b = T.imatrix()
c = T.ivector()
s = ctc(a,b,c)
f = theano.function([a,b,c], s)
```
Fails with the following error:
```
You can find the C code in this temporary file: /tmp/theano_compilation_error_nfqys_ne
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/media/hdd/work/python/Theano/theano/compile/function.py", line 317, in function
    output_keys=output_keys)
  File "/media/hdd/work/python/Theano/theano/compile/pfunc.py", line 486, in pfunc
    output_keys=output_keys)
  File "/media/hdd/work/python/Theano/theano/compile/function_module.py", line 1842, in orig_function
    fn = m.create(defaults)
  File "/media/hdd/work/python/Theano/theano/compile/function_module.py", line 1716, in create
    input_storage=input_storage_lists, storage_map=storage_map)
  File "/media/hdd/work/python/Theano/theano/gof/link.py", line 699, in make_thunk
    storage_map=storage_map)[:3]
  File "/media/hdd/work/python/Theano/theano/gof/vm.py", line 1084, in make_all
    impl=impl))
  File "/media/hdd/work/python/Theano/theano/gof/op.py", line 955, in make_thunk
    no_recycling)
  File "/media/hdd/work/python/Theano/theano/gof/op.py", line 858, in make_c_thunk
    output_storage=node_output_storage)
  File "/media/hdd/work/python/Theano/theano/gof/cc.py", line 1217, in make_thunk
    keep_lock=keep_lock)
  File "/media/hdd/work/python/Theano/theano/gof/cc.py", line 1157, in __compile__
    keep_lock=keep_lock)
  File "/media/hdd/work/python/Theano/theano/gof/cc.py", line 1620, in cthunk_factory
    key=key, lnk=self, keep_lock=keep_lock)
  File "/media/hdd/work/python/Theano/theano/gof/cmodule.py", line 1174, in module_from_key
    module = lnk.compile_cmodule(location)
  File "/media/hdd/work/python/Theano/theano/gof/cc.py", line 1523, in compile_cmodule
    preargs=preargs)
  File "/media/hdd/work/python/Theano/theano/gof/cmodule.py", line 2353, in compile_str
[...]

PR #6455, opened 2017-10-02
Fix omp_get_num_threads scope error in CPU CTC wrapper

joaovictortr:
Fix omp_get_num_threads scope error in CPU CTC wrapper. The error was caused due to a call to 'omp_get_num_threads' outside of a parallel region.

Issue: #6399

### DIFF
--- a/theano/tensor/nnet/ctc.py
+++ b/theano/tensor/nnet/ctc.py
@@ -4,6 +4,7 @@ import sys
 import theano.tensor as T
 from theano import config
 from theano import gof
+from theano.scalar import as_scalar
 from theano.gof import local_optimizer
 from theano.gof.cmodule import GCC_compiler
 from theano.tensor.opt import register_canonicalize
@@ -102,7 +103,7 @@ class ConnectionistTemporalClassification(gof.COp, gof.OpenMPOp):
     """
     __props__ = ('compute_grad',)
 
-    _cop_num_inputs = 3
+    _cop_num_inputs = 4
     _cop_num_outputs = 2
 
     func_file = os.path.join('c_code', 'ctc_wrapper.c')
@@ -175,13 +176,16 @@ class ConnectionistTemporalClassification(gof.COp, gof.OpenMPOp):
         if t_input_lengths.ndim != 1:
             raise ValueError('input_lengths must have 1 dimension.')
 
+        openmp_enabled = 1 if gof.OpenMPOp(self).openmp is not None else 0
+        openmp_enabled = as_scalar(openmp_enabled).astype('int32')
+
         costs = T.fvector(name="ctc_cost")
         outputs = [costs]
         if self.compute_grad:
             gradients = T.ftensor3(name="ctc_grad")
             outputs += [gradients]
 
-        return gof.Apply(self, inputs=[t_activations, t_labels, t_input_lengths],
+        return gof.Apply(self, inputs=[t_activations, t_labels, t_input_lengths, openmp_enabled],
                          outputs=outputs)
 
     def L_op(self, inputs, outputs, output_grads):
@@ -193,7 +197,8 @@ class ConnectionistTemporalClassification(gof.COp, gof.OpenMPOp):
         total_grad = T.basic.batched_dot(grad_op, gradients.dimshuffle(1, 0, 2)).dimshuffle(1, 0, 2)
         return [total_grad,
                 grad_undefined(self, 1, inputs[1]),
-                grad_undefined(self, 2, inputs[2])]
+                grad_undefined(self, 2, inputs[2]),
+                grad_undefined(self, 3, inputs[3])]
 
 
 def ctc(activations, labels, input_lengths):
@@ -239,5 +244,6 @@ def local_ctc_no_grad(node):
     if isinstance(node.op, ConnectionistTemporalClassification):
         if len(node.outputs) > 1:
             if len(node.outputs[1].clients) == 0:   # gradient is not used
-                return [ConnectionistTemporalClassification(compute_grad=False)(*node.inputs), None]
+                activations, labels, input_lengths, _ = node.inputs
+                return [ConnectionistTemporalClassification(compute_grad=False)(activations, labels, input_lengths), None]
     return False
--- a/theano/tensor/nnet/ctc.py
+++ b/theano/tensor/nnet/ctc.py
@@ -1,11 +1,11 @@
 from __future__ import (division, absolute_import, print_function)
 import os
 import sys
+import theano
 import theano.tensor as T
 from theano import config
 from theano import gof
-from theano.scalar import as_scalar
-from theano.gof import local_optimizer
+from theano.gof import (local_optimizer, ParamsType)
 from theano.gof.cmodule import GCC_compiler
 from theano.tensor.opt import register_canonicalize
 from theano.tensor.extra_ops import cpu_contiguous
@@ -102,8 +102,9 @@ class ConnectionistTemporalClassification(gof.COp, gof.OpenMPOp):
         If set to True, enables the computation of gradients of the CTC loss function.
     """
     __props__ = ('compute_grad',)
+    params_type = ParamsType(openmp=theano.scalar.bool)
 
-    _cop_num_inputs = 4
+    _cop_num_inputs = 3
     _cop_num_outputs = 2
 
     func_file = os.path.join('c_code', 'ctc_wrapper.c')
@@ -119,6 +120,7 @@ class ConnectionistTemporalClassification(gof.COp, gof.OpenMPOp):
         gof.OpenMPOp.__init__(self, openmp=openmp)
 
         self.compute_grad = compute_grad
+        self.openmp = gof.OpenMPOp(self).openmp
         # Return only the cost. Gradient will be returned by grad()
         self.default_output = 0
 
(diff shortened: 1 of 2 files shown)


joaovictortr:
@botev , could you please test it in your system?

theano/tensor/nnet/ctc.py
```diff
@@ -175,13 +176,16 @@ def make_node(self, activations, labels, input_lengths):
         if t_input_lengths.ndim != 1:
             raise ValueError('input_lengths must have 1 dimension.')
 
+        openmp_enabled = 1 if gof.OpenMPOp(self).openmp is not None else 0
```
nouiz: I think it should be this line:

    openmp_enabled = 1 if self.openmp is not None else 0

self.openmp should exist and should be a bool.

theano/tensor/nnet/ctc.py
```diff
@@ -175,13 +176,16 @@ def make_node(self, activations, labels, input_lengths):
...
+        openmp_enabled = as_scalar(openmp_enabled).astype('int32')
+
         costs = T.fvector(name="ctc_cost")
         outputs = [costs]
         if self.compute_grad:
             gradients = T.ftensor3(name="ctc_grad")
             outputs += [gradients]
 
-        return gof.Apply(self, inputs=[t_activations, t_labels, t_input_lengths],
+        return gof.Apply(self, inputs=[t_activations, t_labels, t_input_lengths, openmp_enabled],
```
nouiz: I don't see a problem, but I'm not sure this is good.

@abergeron what do you think of this?

Should this be a param?

Ideally, we should allow this to change when we pickle/unpickle, but we didn't supported that up to now. So this isn't a requirement.

theano/tensor/nnet/ctc.py
```diff
@@ -239,5 +244,6 @@ def local_ctc_no_grad(node):
     if isinstance(node.op, ConnectionistTemporalClassification):
         if len(node.outputs) > 1:
             if len(node.outputs[1].clients) == 0:   # gradient is not used
-                return [ConnectionistTemporalClassification(compute_grad=False)(*node.inputs), None]
+                activations, labels, input_lengths, _ = node.inputs
+                return [ConnectionistTemporalClassification(compute_grad=False)(activations, labels, input_lengths), None]
```
nouiz: This would be a good reason why the current make_node isn't good. We need the old code to work. Otherwise, the cloning system won't work.

A quick work around would be to add an openmp_enabled parameter to make_node() and have it default to None. When none, it should use `self.openmp`. But do we really support having the openmp setting change for the same node in the same Theano function? I don't think we should support that.

So that point more that we should use params for that.

notoraptor:
I agree that we should better use params here. Maybe by converting directly openmp to a param, as it is alreay a member of the op ?

Something like this:

```python
# import
from theano.gof import ParamsType

# in Op class code (e.g. after `__props__` declaration)
params_type = ParamsType(openmp=theano.scalar.bool)
```
```C
// In C code.
 int APPLY_SPECIFIC(ctc_cost_cpu)(PyArrayObject *  in_activations,
                                  PyArrayObject *  in_labels,
                                  PyArrayObject *  in_labels,
                                  PyArrayObject *  in_input_lengths,
                                  PyArrayObject *  in_input_lengths,
                                  PyArrayObject ** out_costs,
                                  PyArrayObject ** out_costs,
                                  PyArrayObject ** out_gradients)
                                  PyArrayObject ** out_gradients,
                                  // params received as last parameter of apply function
                                  PARAMS_TYPE* params) {

int openmp_enabled = params->openmp;
// ...
```

theano/tensor/nnet/ctc.py
joaovictortr: Yes, I was just not sure about the conversion and/or support for bool in C compilers, that is why I've used this expression with an int. Will fix that.

theano/tensor/nnet/ctc.py
joaovictortr: Indeed, I don't think this kind of change should be supported too, I'll also fix it.

joaovictortr:
Using @notoraptor 's suggestion, I've changed the CTC Op to use ParamType instead.

theano/tensor/nnet/ctc.py
```diff
@@ -118,6 +120,7 @@ def __init__(self, compute_grad=True, openmp=None):
         gof.OpenMPOp.__init__(self, openmp=openmp)
 
         self.compute_grad = compute_grad
+        self.openmp = gof.OpenMPOp(self).openmp
```
notoraptor: Is this line still useful ? As op inherits from `OpenMPOp`, I thought it should already have an attribute nammed `openmp`. Could you check if everything works when you remove that line ?

botev:
@joaovictortr I still get an error:
```
Exception: ('The following error happened while compiling the node', 
ConnectionistTemporalClassification{compute_grad=False}(CpuContiguous.0, <TensorType(int32, 
matrix)>, <TensorType(int32, vector)>), '\n', 'Compilation failed (return status=1): /home/alex/.theano
/compiledir_Linux-4.10--generic-x86_64-with-debian-stretch-sid-x86_64-3.6.1-64/tmp9xg_s622
/mod.cpp: In function ‘void ctc_context_init(ctc_context_t*, int)’:. /home/alex/.theano/compiledir_Linux-
4.10--generic-x86_64-with-debian-stretch-sid-x86_64-3.6.1-64/tmp9xg_s622/mod.cpp:104:52: error: 
‘omp_get_max_threads’ was not declared in this scope.          options->num_threads = 
omp_get_max_threads();.                                                     ^. ', 
'[ConnectionistTemporalClassification{compute_grad=False}(<TensorType(float32, 3D)>, 
<TensorType(int32, matrix)>, <TensorType(int32, vector)>)]')
```

nouiz:
@botev what is your value of "import theano; print(theano.config.openmp)"? To help @joaovictortr reproduce the problem.

Which version of c++ compiler do you have? What OS?

botev:
So that gives `False`, which indicates that `openmp` is not present (or at least not detected by Theano). The system is `Ubuntu 16.04` with `4.10.8` Kernel. gcc and g++ are version `5.4.0`.
Cmake in the `warp-ctc` library gives the following output:
```
-- The C compiler identification is GNU 5.4.0
-- The CXX compiler identification is GNU 5.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CUDA_TOOLKIT_ROOT_DIR not found or specified
-- Could NOT find CUDA (missing:  CUDA_TOOLKIT_ROOT_DIR CUDA_NVCC_EXECUTABLE CUDA_INCLUDE_DIRS CUDA_CUDART_LIBRARY) (Required is at least version "6.5")
-- cuda found FALSE
CMake Warning at CMakeLists.txt:48 (FIND_PACKAGE):
  By not providing "FindTorch.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Torch", but
  CMake did not find one.

  Could not find a package configuration file provided by "Torch" with any of
  the following names:

    TorchConfig.cmake
    torch-config.cmake

  Add the installation prefix of "Torch" to CMAKE_PREFIX_PATH or set
  "Torch_DIR" to a directory containing one of the above files.  If "Torch"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Torch found Torch_DIR-NOTFOUND
-- Building shared library with no GPU support
-- Configuring done
-- Generating done
```

joaovictortr:
OpenMP is not present, however the compiler parses the block containing the ```omp_get_max_threads``` function. In such case, ```_OPENMP``` might be defined or the macro is being removed before the compilation, depending on how Theano handles ```#section``` macros.

abergeron:
`#section` has no influence on externally defined macros like `_OPENMP`

botev:
Hmm, ok I found the error. From very old times I still had this line in my `.theanorc`:
```
[gcc]
cxxflags = -O3 -fopenmp -ffast-math -ftree-loop-distribution -funroll-loops -ftracer
```
Removing the `-fopenmp` flag from there resolved the issue. I'm not too sure why that is the case and how are the `cxxflags` handled, but that seems to have been the issue all along.

nouiz:
Can you try with this Theano flag and no gcc.cxxflags with the master and
with this PR?

openmp=True

This is the right way to enable openmp for Theano. I would like to be sure
if you use the right interface it also work well.

botev:
@nouiz It does run, but how can I check if it actually uses openmp?

nouiz:
Make a small graph with just a bit CTC and check with top in another shell?

botev:
Seems to be using a single core.

nouiz:
@joaovictortr Did you see it use more then 1 core when developing this?

joaovictortr:
@nouiz yes, using htop to see it, it used the 4 cores on my machine. Again, if @botev can provide an example, I'll investigate it promptly.

botev:
I think the issue is with Theano detecting openmp. The example is literally 
```
import theano
import theano.tensor as T
from theano.tensor.nnet.ctc import ctc
a = T.ftensor3()
b = T.imatrix()
c = T.ivector()
s = ctc(a,b,c)
f = theano.function([a,b,c], s)
```
and than running `f`. Where in Theano is the openmp check/flag so I can just print it out by edditing to check if it finds it correctly?

nouiz:
Theano by default disable OpenMP. To enable it, you must set the Theano flag openmp=True.

We did that in the past as it was slowing down by default some cases. The part of the code that caused that was disabled. So we could check if it still slow down computation by default or not.

I'll close this PR. The problem was your flag. But can you try with the flag openmp=True and see it if work for you?

nouiz:
This PR give a better user error in that case: https://github.com/Theano/Theano/pull/6472

Issue #6435, opened 2017-09-20
WrongValue error after 2 consecutives cudnn conv gradinput on same data with TRUE_HALF_CONFIG


notoraptor:
While debugging errors related to #5932 , I think I get a strange behaviour with cudnn conv gradinput, algo deterministic and float16 precision. When I compute gradinput with same arrays twice, I get expected results twice (compared to CPU) with float32 and float64, but I get a WrongValue error in the second call with float16.

This is a script to reproduce the problem (it contains code extracted from #5932 to ease tests). Script can receive the dtype as argument (float16, float32 and float64). To quick check used arrays, it also prints a hash of tensors involved in computation.

Script: https://gist.github.com/notoraptor/f6f8cd0471c66a9de227ba85a1c47d21

Parameters used:
- inputs shape (2, 3, 300, 5)
- filters shape (2, 3, 40, 4)
- borders (1, 1)
- alpha 2
- beta -3

Output with float64:
```
$ python ../test_dnn_float16.py float64
Using cuDNN version 7002 on context None
Mapped name None to device cuda: GeForce GTX 1050 (0000:01:00.0)
deterministic float64 float64 ((2, 3, 300, 5), (2, 3, 40, 4), (1, 1), (1, 1), (1, 1), 'conv', 2, -3)
Hash inputs_val            : eec5f0862f89822d1cbc830a89d49c72a318e98d5d4deedf6837dcd6864b674a
Hash filters_val           : 73ebccff5f4a23da52130e436000461c7f90929532557d6f14560c9b469ad89e
Hash topgrad_val           : b6ea724a38cdf6e2e80fb7617c7d7e6140a1d30a14c67cba47a3bbd420dee375
Hash CPU res before scaling: 6724456cb9ddc84e5696d25225a60f24ddb30589b532b3b17b970566c5d1f5c3
Hash     res before scaling: a3a871976ac9df9d5be88d62232637b0b1d3b86b857653769ee8006e3fc8dab5
Hash CPU res after  scaling: c7b3b0aa1108102193a059477647088c17c96e0e26cd8fa2271673e368394c1b
Hash     res after  scaling: 2295eda58b24fbbfcf23fc58a5ded70aba22861e70e4d0ed98d48a2fb7d81233
CPU
[-0.007 -0.003  0.001 -0.003 -0.007] [-0.007 -0.003  0.001 -0.003 -0.007]
res
[...]

PR #6457, opened 2017-10-03
Disallow float16 precision for convolution gradients functions.

notoraptor:
@nouiz @abergeron @lamblin 

fix https://github.com/Theano/Theano/issues/6435

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -329,7 +329,7 @@ handle_type = CUDNNDataType('cudnnHandle_t', 'cudnnDestroy')
 cudnn = cudnn_defs.get_definitions(version(raises=False))
 
 
-def get_precision(precision, inputs):
+def get_precision(precision, inputs, for_grad=False):
     if precision is None:
         precision = theano.config.dnn.conv.precision
     if precision == 'as_input' or precision == 'as_input_f32':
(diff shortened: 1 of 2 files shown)


theano/gpuarray/dnn.py
```diff
@@ -1129,6 +1135,8 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1
                    shape_i(img, 4, fgraph) - shape_i(kerns, 4, fgraph) + 1)
         out_shp = assert_conv_shape(out_shp)
         out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*out_shp)
+        if precision == 'float16':
```
nouiz: Currently, if precission is float16 and we cann dnn_conv() and dnn_conv3d, it will silently upcast it to float32 in some cases. This isn't tested.

I don't like that too much. Why not raise an error in that case?

notoraptor:
@nouiz Updated with your suggestions.

abergeron:
I'm not sure an error is the right choice.  The point here is to ask cudnn to make the accumulation in float32, but the inputs and outputs still stay as float16.

What was recommended to us, by the nvidia folks, was to disable support for accumulation in float16 completely, including forward, even if that currently works because it is unmaintained and risks breaking in future versions of cudnn.

nouiz:
The last recommendation is to disable it just for the grad. This is why it
was done like this.

PR #6430, opened 2017-09-19
Upsamp2

erakra:
handling fractional bilinear upsampling

PR #6460, opened 2017-10-05
small simplification

erakra:

PR #6461, opened 2017-10-05
Simplifies `frac_bilinear_upsampling`

### LINKED ISSUE #5995: Generalize bilinear interpolation to rational factor
Currently, bilinear interpolation only handles upsampling by an integer factor (#3928, #3951).
It would be nice to handle any rational factor `p / q`. In theory, it can be done by upsampling by `p` using the current method, and then simply downsampling by `q`.

Fortunately, not that Theano supports both kernel dilation and efficient subsampling for convolutions, there is a way of doing that efficiently, with a _forward_ convolution:
- as _input_, the constant pyramidal filter for upsampling by `p` (of shape `2p-1`),
- as _filters_, the image (with duplicated border), and a _filters dilation_ of `p`
- a _subsampling_ of `q`
- as _border_mode_, the appropriate padding to get an output of the right size

If we want an output of `(inputsize * p) // q`, then `border_mode = inputsize * p - (p + q) // 2 + 1`.
However, I think that we would want the output to be `ceil(inputsize * p / q)` instead, at least by default.

There are still a couple of things to figure out.

What should the interface be to provide that ratio?
- A pair of integers? That could be confused with different upsampling integer ratios for rows and columns.
- Providing a float would not be nice because it makes it hard to recover both ints.
- We could ask for the output size instead and figure out the ratio (divide sizes by their GCD). But it would not make possible to upsample something of shape 5 by 3/2 and get something of shape 8 (which is slightly padded, but less blurry than an upsampling factor of 8/5).
- Maybe change the interface and have other arguments for that?

Should we round up or down when there is an uneven ratio, or even forbid it? I think rounding up would be better, but maybe the other behaviours could be options.

Make sure the PR does not interfere with #5954.

pqrth:
* As [suggested](https://github.com/Theano/Theano/pull/6430#discussion_r140538184) in PR #6430, this PR removes `ratio` parameter of the `frac_bilinear_upsampling` as `frac_ratio=(ratio, 1)` can be used instead.
* Closes #5995

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -1761,8 +1761,7 @@ def bilinear_kernel_1D(ratio, normalize=True):
 
 
 def frac_bilinear_upsampling(input,
-                             ratio=None,
-                             frac_ratio=None):
+                             frac_ratio):
     """Compute bilinear upsampling
     This function will build the symbolic graph for upsampling
     a tensor by the given ratio using bilinear interpolation.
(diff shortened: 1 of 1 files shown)


pqrth:
Already merged in a separate PR #6460.

PR #1957, opened 2014-07-11
Pad

Tanjay94:
Added new pad instruction to match Numpy 1.7.

PR #6463, opened 2017-10-05
Supports the NumPy pad function

### LINKED ISSUE #1216: Support the NumPy pad fct.
NumPy 1.7 support a new pad instruction. We should support it. Some mode are easy to support with current ops, but some of then could need new opt or more complicated graph. In the cases were we don't need a new complicated graph, it is better to don't use a new op. But may to support all mode, making a new op "Pad" that wrap this fct is needed.
## Ability to pad rank-n arrays

A pad module containing functions for padding n-dimensional arrays has been
added. The various private padding functions are exposed as options to a
public 'pad' function.  Example::

```
pad(a, 5, mode='mean')
```

Current modes are `constant`, `edge`, `linear_ramp`, `maximum`,
`mean`, `median`, `minimum`, `reflect`, `symmetric`, `wrap`, and
`<function>`.

pqrth:
* Rebases #1957
* Resolves #1216


lamblin:
flake8 is failing unfortunately

PR #6466, opened 2017-10-06
[BUG,DOC] Doc MILA status about Theano and don't review a view.

nouiz:


lamblin:
Changes to `doc/index.txt` should be back-ported to 0.9.X I think.

nouiz:
I pushed both doc change (README.rst and doc/index.txt) to the 0.9.X branch.

PR #6467, opened 2017-10-06
Improve alternative BLAS code and add tests for alternative gemm function.

### LINKED ISSUE #6464: Intermittent error in TestCGemvNoFlags.test_cgemv
It looks like `theano.tensor.tests.test_blas_c.TestCGemvNoFlags.test_cgemv` fail in a non-deterministic way, see [here](https://travis-ci.org/Theano/Theano/jobs/283891616#L3503).

It looks like the output memory is not always zeroed out when `beta == 0`, so `nan` can result if there were non-finite elements in that memory, whereas the reference (and expected) result is `0`.

notoraptor:
This should fix #6464 . Now if a tensor must be elemwise-multiplied by zero, then it is explicitly set to 0 to overwrite any previous value.

I also added unit tests for alt gemm, so that all current alternative BLAS functions are tested.

@lamblin @nouiz @abergeron

### DIFF
--- a/theano/tensor/c_code/alt_blas_template.c
+++ b/theano/tensor/c_code/alt_blas_template.c
@@ -1,17 +1,17 @@
 /** Alternative template NumPy-based implementation of BLAS functions used in Theano. **/
 
-/* Compute tensor[i] = scalar for every position i in tensor. */
-void alt_numpy_memset_inplace_%(float_type)s(PyArrayObject* tensor, const %(float_type)s* scalar) {
-    if (PyArray_IS_C_CONTIGUOUS(tensor) && *scalar == (char)(*scalar)) {
+/* Compute matrix[i][j] = scalar for every position (i, j) in matrix. */
+void alt_numpy_memset_inplace_%(float_type)s(PyArrayObject* matrix, const %(float_type)s* scalar) {
+    if (PyArray_IS_C_CONTIGUOUS(matrix) && *scalar == (char)(*scalar)) {
         // This will use memset.
-        PyArray_FILLWBYTE(tensor, (char)(*scalar));
+        PyArray_FILLWBYTE(matrix, (char)(*scalar));
         return;
     }
-    NpyIter* iterator = NpyIter_New(tensor,
+    NpyIter* iterator = NpyIter_New(matrix,
         NPY_ITER_READWRITE | NPY_ITER_EXTERNAL_LOOP | NPY_ITER_REFS_OK,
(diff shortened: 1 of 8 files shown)


theano/tensor/c_code/alt_blas_template.c
```diff
@@ -1,8 +1,43 @@
 /** Alternative template NumPy-based implementation of BLAS functions used in Theano. **/
 
+/* Compute tensor[i] = scalar for every position i in tensor. */
+void alt_numpy_memset_inplace_%(float_type)s(PyArrayObject* tensor, const %(float_type)s* scalar) {
+    if (PyArray_IS_C_CONTIGUOUS(tensor) && *scalar == (char)(*scalar)) {
+        // This will use memset.
+        PyArray_FILLWBYTE(tensor, (char)(*scalar));
+        return;
+    }
+    NpyIter* iterator = NpyIter_New(tensor,
```
nouiz: This seem a duplicate of alt_numpy_scale_matrix_inplace_%(float_type)s. Did you wanted to change line 24 too to do a affectation only?

theano/tensor/c_code/alt_blas_template.c
```diff
@@ -1,8 +1,43 @@
 /** Alternative template NumPy-based implementation of BLAS functions used in Theano. **/
 
+/* Compute tensor[i] = scalar for every position i in tensor. */
```
nouiz: I think the code here only work well for matrix. So rename tensor to matrix in this diff.

theano/tensor/tests/test_blas.py
```diff
@@ -373,6 +373,98 @@ def t(z, x, y, a=1.0, b=0.0, l='c|py', dt='float64'):
             1, 0, 2)), dt='float32')
 
 
+class t_gemm_no_flags(object):
```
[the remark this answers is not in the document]
notoraptor: Fixed. Note that this test is in `test_blas`, while other tests (for gemv and dot) are in `test_blas_c`.

theano/tensor/c_code/alt_blas_template.c
notoraptor: Yes, both functions have pretty similar code (because of NumPy iteration code). I have indeed changed line 24 because here I just want to do an affectation.

It is needed in some cases, e.g. when beta is 0. In such case output tensor may contain arbitrary values, so we must explicitly set them to 0, otherwise a nan might be multiplied by 0, resulting to still a nan in final output.

PR #6468, opened 2017-10-06
Don't ignore CDataType version.

abergeron:


abergeron:
The main consequence is that files don't get recompiled.  I'm not 100% sure about wrong values, but I don't think so.

This is particularly annoying when you are switching between cudnn versions since it will complain about libcunn.so not being found because it doesn't find the old version.

I discovered it while working on the V100 hang.

PR #6469, opened 2017-10-06
Skip test when no c++ compiler

nouiz:


abergeron:
There are flake8 problems

Issue #6470, opened 2017-10-06
Incorrect gradient when arguments to minimum() and maximum() are equal


astooke:
Hi,
 
EDITED: previously had written about `theano.tensor.clip()` in here, but that was not the problem.

I'm experiencing incorrect gradients when the loss function includes `theano.tensor.minimum()` and the two arguments to `minimum` are equal.  It returns either the sum, or one of the values multiplied by 2.  

See the example below.  Comparing all the pieces against the same in Tensorflow, only the last assertion--the final gradient--fails for me when `np_ratios1` is used for both inputs.

It also happens in `theano.tensor.maximum()`.

Please Help!!!

Happens in Theano 0.9 and bleeding edge 0.10.

```
import numpy as np
import theano
import theano.tensor as T
import tensorflow as tf

SIZE = 10

# NUMPY

np_ratios1 = 0.4 * np.random.randn(SIZE).astype('float32') + 1
np_ratios1 = np.clip(np_ratios1, 0.1, 100)
np_ratios2 = 0.4 * np.random.randn(SIZE).astype('float32') + 1
np_ratios2 = np.clip(np_ratios2, 0.1, 100)
np_advs = np.random.randn(SIZE).astype('float32')
np_surr1 = np_ratios1 * np_advs
np_surr2 = np_ratios2 * np_advs
np_surr = np.minimum(np_surr1, np_surr2)
np_loss = - np.sum(np_surr)

print("np_advs:\n", np_advs)
print("np_ratios1:\n", np_ratios1)
print("np_ratios2:\n", np_ratios2)
print("np_surr:\n", np_surr)


# THEANO

th_ratios1 = T.vector('ratios1')
th_ratios2 = T.vector('ratios2')
th_advs = T.vector('th_advs')
th_surr1 = th_ratios1 * th_advs
th_loss1 = - T.sum(th_surr1)
th_grad1 = T.grad(th_loss1, wrt=th_advs)
th_surr2 = th_ratios2 * th_advs
th_loss2 = - T.sum(th_surr2)
th_grad2 = T.grad(th_loss2, wrt=th_advs)
th_surr = T.minimum(th_surr1, th_surr2)
th_loss = - T.sum(th_surr)
th_grad = T.grad(th_loss, wrt=th_advs)

th_f = theano.function(
    inputs=[th_ratios1, th_ratios2, th_advs],
    outputs=[th_surr1, th_surr2, th_surr, th_grad1, th_grad2, th_grad]
)

[...]

PR #6477, opened 2017-10-12
Minimum maximum gradient fix

nouiz:
Fix gh-6470. But also enable more optimization by using a real cast instead of mimiking one.


abergeron:
flake8 problems

PR #6479, opened 2017-10-13
Don't initialize the GPU in MPI subprocesses.

abergeron:


abergeron:
This may help with the buildbot freezes encountered in this test.

theano/tensor/tests/test_mpi.py
```diff
@@ -46,17 +46,18 @@ def test_mpi_roundtrip():
     if not mpi_enabled:
         raise SkipTest('MPI not enabled')
     theano_root = theano.__file__.split('__init__')[0]
-    d = {}
-    if PY3:
-        # Is some not understood cases, the subprocess never finish.
-        d = dict(timeout=5*60)
+    env = os.environ.copy()
+    flags = env['THEANO_FLAGS']
+    keep_flags = ','.join((f for f in flags.split(',') if not f.startswith('init_gpu_device')))
```
nouiz: Also filter out device?

abergeron: For the tests device is supposed to be 'cpu' so that's why I didn't filter it out.

abergeron:
Initializing cuda + forking + reinitializing is not a good idea.

nouiz:
Right.

nouiz:
Can you revert this PR to have it tested in the daily buildbot? https://github.com/Theano/Theano/pull/6459

abergeron:
Had to rebase to get the commits to revert in the history.

nouiz:
Note, this remind me of an old issue that would make Theano raise an error when this happen. It didn't got finished, but we see it would have helped.

Mostly, if when we init a device, we can set an env variable that would be detected by subprocess and have them raise an error in that case. This would give a much better user error.

PR #6489, opened 2017-10-18
Change some texts to prepare theano 1.0.

### LINKED ISSUE #6487: Rename 0.10 to 1.0 in the doc/tags/website.

notoraptor:
Should fix #6487 

@nouiz 

NB: I don't find other texts to update. Other places when `"0.10"` appears should be updated only when we are ready to do the release.


doc/library/config.txt
```diff
@@ -571,7 +571,7 @@ import theano and print the config variable, as in:
 .. attribute:: config.warn.ignore_bug_before
 
     String value: ``'None'``, ``'all'``, ``'0.3'``, ``'0.4'``, ``'0.4.1'``, ``'0.5'``,
-    ``'0.6'``, ``'0.7'``, ``'0.8'``, ``'0.8.1'``, ``'0.8.2'``, ``'0.9'``, ``'0.10'``
+    ``'0.6'``, ``'0.7'``, ``'0.8'``, ``'0.8.1'``, ``'0.8.2'``, ``'0.9'``, ``'0.10'``, ``'1.0'``
```
nouiz: I think you should also update the fallback version. Check the doc how to release, the recent version in the master, it should tell you where this is.

notoraptor: Fixed (`1.0.0dev+unknown`).

PR #6491, opened 2017-10-19
Optimize SUM(x^2), SUM(ABS(X)) and MAX(ABS(X)) to cuDNN reduction.

notoraptor:
@nouiz @abergeron

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3755,13 +3755,18 @@ def local_dnn_reduction(node):
     scal = node.op.scalar_op.name
     post = _identity
 
-    if (isinstance(node.op.pre_scalar_op, theano.scalar.basic.Sqr) and
-            isinstance(node.op.scalar_op, theano.scalar.basic.Add)):
-        scal = 'norm2'
-        post = _square
-    elif node.op.pre_scalar_op is not None:
-        # Might want to handle absmax, avg, norm1, and other cases of norm2 here
-        return
+    if node.op.pre_scalar_op is not None:
+        # Might want to handle absmax, avg, and other cases for (norm1, norm2) here
+        if isinstance(node.op.scalar_op, theano.scalar.basic.Add):
+            if isinstance(node.op.pre_scalar_op, theano.scalar.basic.Sqr):
+                scal = 'norm2'
+                post = _square
+            elif isinstance(node.op.pre_scalar_op, theano.scalar.basic.Abs):
+                scal = 'norm1'
+            else:
+                return
+        else:
+            return
 
     if not cudnn.cudnnReduceTensorOp_t.has_alias(scal):
         return
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3746,19 +3746,33 @@ def local_dnn_reduction(node):
             node.op.acc_dtype == 'float64'):
         return
 
-    if node.op.pre_scalar_op is not None:
-        # Might want to handle absmax, avg, norm1, norm2 here
+    def _identity(a):
+        return a
+
+    def _square(a):
+        GpuElemwise(theano.scalar.basic.sqr)(a)
+
+    scal = node.op.scalar_op.name
+    post = _identity
+
+    if (isinstance(node.op.pre_scalar_op, theano.scalar.basic.Sqr) and
+            isinstance(node.op.scalar_op, theano.scalar.basic.Add)):
+        scal = 'norm2'
+        post = _square
+    elif node.op.pre_scalar_op is not None:
+        # Might want to handle absmax, avg, norm1, and other cases of norm2 here
         return
 
-    if not cudnn.cudnnReduceTensorOp_t.has_alias(node.op.scalar_op.name):
+    if not cudnn.cudnnReduceTensorOp_t.has_alias(scal):
         return
 
     with inherit_stack_trace(node.outputs):
-        return (GpuDnnReduction(node.op.scalar_op.name,
-                                node.op.axis,
-                                node.op.acc_dtype,
-                                node.op.dtype,
-                                False)(node.inputs[0]),)
+        ret = GpuDnnReduction(scal,
+                              node.op.axis,
+                              node.op.acc_dtype,
+                              node.op.dtype,
+                              False)(node.inputs[0])
+        return [post(ret)]
 
 
 @register_opt('cudnn')
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3765,6 +3765,9 @@ def local_dnn_reduction(node):
                 scal = 'norm1'
             else:
                 return
+        elif (isinstance(node.op.scalar_op, theano.scalar.basic.Maximum) and
+                isinstance(node.op.pre_scalar_op, theano.scalar.basic.Abs)):
+            scal = 'absmax'
         else:
             return
 
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3750,7 +3750,7 @@ def local_dnn_reduction(node):
         return a
 
     def _square(a):
-        GpuElemwise(theano.scalar.basic.sqr)(a)
+        return GpuElemwise(theano.scalar.basic.sqr)(a)
 
     scal = node.op.scalar_op.name
     post = _identity
--- a/theano/gpuarray/elemwise.py
+++ b/theano/gpuarray/elemwise.py
@@ -932,7 +932,7 @@ class GpuCAReduceCuda(GpuKernelBase, HideC, CAReduceDtype):
                 Z = (%(out_type)s *)(((char *)Z)+offset_Z);
         """ % locals()
 
-    def _assign_init(self, first_item, dtype='float32'):
+    def _assign_init(self, first_item, dtype):
         """
         This return the initial value for myresult.
         If the scalar op have an identity value, return it.
@@ -1707,7 +1707,7 @@ class GpuCAReduceCuda(GpuKernelBase, HideC, CAReduceDtype):
         """ % locals(), file=sio)
 
     def c_code_cache_version_apply(self, node):
-        version = [22]  # the version corresponding to the c code in this Op
+        version = [23]  # the version corresponding to the c code in this Op
 
         # now we insert versions for the ops on which we depend...
         scalar_node = Apply(
@@ -1728,6 +1728,7 @@ class GpuCAReduceCuda(GpuKernelBase, HideC, CAReduceDtype):
         in_dtype = node.inputs[0].dtype
         out_dtype = node.outputs[0].dtype
         acc_dtype = self._acc_dtype(node.inputs[0].dtype)
+        assign_dtype = in_dtype
         flags = Kernel.get_flags(in_dtype, acc_dtype, out_dtype)
         in_type = gpuarray.dtype_to_ctype(in_dtype)
         out_type = gpuarray.dtype_to_ctype(out_dtype)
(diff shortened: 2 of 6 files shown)


abergeron:
I'm wondering why we don't just disable the elemwise fusion.

nouiz:
It will raise the peak memory usage and will be much slower then with this PR.

old code:
- V100: 0.136s
- TITAN Xp: 0.127s

new code:
v100: 0.000235s
Titan Xp: 0.00035s

speed up:
- V100 578x
- Titan Xp 361x

abergeron:
Ok for this then.

notoraptor:
One Jenkins failure: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/961/testReport/junit/theano.gpuarray.tests/test_opt/test_local_gpu_elemwise_careduce/

Fixed: 4f2862a

theano/gpuarray/tests/test_opt.py
```diff
@@ -359,24 +359,19 @@ def test_pdbbreakpoint_op():
...
-    data = np.random.rand(3, 4).astype(theano.config.floatX)
-    utt.assert_allclose(f(data), (data * data).sum())
-
-    o = (x * x).sum(axis=1)
-    f = theano.function([x], o, mode=mode_with_gpu)
-    topo = f.maker.fgraph.toposort()
-    assert len(topo) == 3
-    assert topo[1].op.pre_scalar_op == theano.scalar.sqr
-    assert _check_stack_trace(f)
-    utt.assert_allclose(f(data), (data * data).sum(axis=1))
```
lamblin: Maybe we should keep that test in addition to the new one, but disable dnn optimizations in the mode.
That way we make sure the old optimization still works when there is no cudnn.

notoraptor:
@lamblin Fixed !

theano/gpuarray/dnn.py
```diff
@@ -3746,19 +3746,33 @@ def local_dnn_reduction(node):
...
+        return GpuElemwise(theano.scalar.basic.sqr)(a)
+
+    scal = node.op.scalar_op.name
+    post = _identity
+
+    if (isinstance(node.op.pre_scalar_op, theano.scalar.basic.Sqr) and
+            isinstance(node.op.scalar_op, theano.scalar.basic.Add)):
+        scal = 'norm2'
+        post = _square
+    elif node.op.pre_scalar_op is not None:
```
nouiz: While at it, can you add the case for norm1? This way, we shouldn't need to make such fix in the futur. If you can't do that today, I think we should merge this PR, but if it can be done quickly, it can be added to this PR.

notoraptor:
@nouiz updated.

notoraptor:
@nouiz Updated, I added optimization for `MAX(ABS(X))`.

I cannot currently add optimization for average (`MEAN(X)`) because theano `mean()` function produces a graph with a sum divided iteratively by axes (something like `x.sum() / shape[0] / shape[1] / ...`). So I don't know if it is a good idea to parse such case.

Mean function is here: https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L3272

theano/gpuarray/tests/test_opt.py
```diff
@@ -372,12 +372,13 @@ def fn_sum_abs(x, axis):
...
-        for fn, pre_scalar_op in ((fn_sum_square, theano.scalar.sqr),
-                                  (fn_sum_abs, theano.scalar.abs_),
-                                  (fn_max_abs, theano.scalar.abs_)):
+    for fn, pre_scalar_op in ((fn_sum_square, theano.scalar.sqr),
+                              (fn_sum_abs, theano.scalar.abs_),
+                              (fn_max_abs, theano.scalar.abs_)):
+        for axis in (None, 0, 1):
             o = fn(x, axis)
             f = theano.function([x], o, mode=mode_with_gpu_no_cudnn)
+            theano.printing.debugprint(f)
```
nouiz: remove from the test

nouiz:
jenkins a eu une erreur

notoraptor:
I am not able to reproduce the problem (tested on: cuda 8 + cudnn 7002/7003, then cuda 9 + cudnn 7003).

nouiz:
jenkins failure: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/968/

jenkins test this

Maybe it was a random error, so I start the test again, just in case.

slefrancois:
jenkins test this

slefrancois:
Jenkins tests were stuck so I killed and restarted manually.

notoraptor:
In fact, error can be reproduced with `floatX=float16`: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/972/console
```
[Theano_GPU]         // rest of function is handled by one warp
[Theano_GPU]         if (threadNum < warpSize) {
[Theano_GPU]             //round up all the partial sums into the first `warpSize` elements
[Theano_GPU]             for (int i = threadNum + warpSize; i < threadCount; i += warpSize)
[Theano_GPU]             {
[Theano_GPU]                 myresult = ((buf[i])>(myresult)? (buf[i]): ((myresult)>=(buf[i])? (myresult): nan("")));
[Theano_GPU]             }
[Theano_GPU]             buf[threadNum] = myresult;
[Theano_GPU]         }
[Theano_GPU]         __syncthreads();
[Theano_GPU]         for (unsigned int _n = warpSize / 2; _n > 0; _n /= 2) {
[Theano_GPU]             if (threadNum < _n && threadNum + _n < threadCount)
[Theano_GPU]             buf[threadNum] = ((buf[threadNum+_n])>(buf[threadNum])? (buf[threadNum+_n]): ((buf[threadNum])>=(buf[threadNum+_n])? (buf[threadNum]): nan("")));
[Theano_GPU]             __syncthreads();
[Theano_GPU]         }
[Theano_GPU]         if (threadNum == 0) {
[Theano_GPU]           Z[0] = (buf[0]);
[Theano_GPU]         }
[Theano_GPU]         
[Theano_GPU]             }
[Theano_GPU]             
[Theano_GPU] NVRTC compile log::
[Theano_GPU] default_program(11): error: more than one instance of overloaded function "fabs" matches the argument list:
[Theano_GPU]             function "fabs(double)"
[Theano_GPU]             function "fabs(float)"
[Theano_GPU]             argument types are: (const short)
```

It seems a call to `fabs()` receives a `short` but can not then choose between `fabs(double)` and `fabs(float)`. It seems related to how the kernel code is generated for pre scalar op. My fix in last commit seems to resolve the problem: https://github.com/Theano/Theano/pull/6491/commits/5e80e5fa46a0e47159c59052ae1e84579b9e155e

@nouiz @abergeron @lamblin

nouiz:
travis fail with wrong result errors.

notoraptor:
Yes, I'm working on it.

notoraptor:
Let's see if jenkins still have errors.

notoraptor:
Still jenkins errors: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/977/

I retry with libgpuarray compiled in `Release` mode.

notoraptor:
Oh, that's strange: no more error when libgpuarray is compiled in Release mode: http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/980/

Is it something expectable ? @abergeron @nouiz @lamblin 

I guess no.

nouiz:
I do not expect changing from Debug compilation to Release compilation of libgpuarray to change the results. I'll review in detail this PR. Can you make a new PR that set again the Debug compilation of libgpuarray?

notoraptor:
@nouiz Done: #6496 .

theano/gpuarray/elemwise.py
```diff
@@ -932,7 +932,7 @@ def _k_init(self, node, nodename):
                 Z = (%(out_type)s *)(((char *)Z)+offset_Z);
         """ % locals()
 
-    def _assign_init(self, first_item):
+    def _assign_init(self, first_item, dtype='float32'):
```
nouiz: remove the default value for dtype.

theano/gpuarray/elemwise.py
```diff
@@ -1744,7 +1743,7 @@ def gpu_kernels(self, node, nodename):
             reduce_fct = self._assign_reduce(node, nodename, "myresult",
                                              load_in + "(A[i0])",
                                              {}, True)
-            reduce_init = self._assign_init(load_in + "(A[0])")
+            reduce_init = self._assign_init(load_in + "(A[0])", acc_dtype)
```
nouiz: I think you should use the input dtype, not the acc_dtype in all call to assign_init()

theano/gpuarray/opt.py
```diff
@@ -1207,7 +1207,7 @@ def local_gpua_careduce(op, context_name, inputs, outputs):
             return False
         x, = inputs
         idtype = x.dtype
-        adtype = getattr(op, 'acc_dtype', None)
+        adtype = getattr(op, 'acc_dtype', idtype)
```
nouiz: Why this change? I don't understand it.

notoraptor: It's because of this line in optimization that converts gpucareduce to gpudnnreduction: https://github.com/Theano/Theano/blob/master/theano/gpuarray/dnn.py#L3737

The `acc dtype` in CPU careduce may be undefined or set to None, but it is currently updated by the optimization that converts careduce to gpucareducecuda (1) for float32 case, so that everything goes well in float32. But when floatX=float64, `acc dtype` will still be None after optimization from CPU to GPU (1), and then optimization from GPU to cuDNN will fail at line pointed above.

(1): https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1210

I detected this problem with `theano.gpuarray.tests.test_dnn.test_dnn_reduction_absmax`, which failed with floatX=float64 on my computer.

With the introduced change, I make sure `acc dtype` falls back to input dtype if it is not defined by CPU careduce.

theano/gpuarray/elemwise.py
notoraptor: Fixed

theano/gpuarray/elemwise.py
notoraptor: OK

notoraptor:
Updated.

theano/gpuarray/opt.py
nouiz: From tensor/elemwise.py line 1340, I think we should use the output dtype as the acc_dtype when there is no acc_dtype.

notoraptor: This line (tensor/elemwise.py line 1340) is in `perform()` method, so it does not change the op attributes (and it will not be called when optimizing to GPU).

notoraptor: PS: Maybe can I just fix it for float64 as it is done for float32 ?  https://github.com/Theano/Theano/blob/master/theano/gpuarray/opt.py#L1216

With float64:
```python
        if idtype == 'float64' and odtype == 'float64':
            adtype = 'float64'
```

nouiz: This won't be enough. Your change to assing_init are good. We should not always use float32 inside it. I'm still reviewing your new commit.

lamblin:
When libgpuarray is compiled in Debug, it also adds an offset to each memory allocation.
If incorrect results get produced, it could mean that the offset is not correctly passed. Since the offset is zero for most arrays, compiling in Release could work most of the time, but still be incorrect in some cases.

nouiz:
@notoraptor can you try to reproduce those error locally by building libgpuarray in Debug?

if you can reproduce them, try with Theano mode=DebugMode in addtion to the above. This could help identify which op have the problem.

notoraptor:
@nouiz Yes I am able to reproduce it on Ubuntu (by compiling libgpuarray in Debug mode). I will try theano debug mode tomorrow.

Note that DNN errors are fixed by @abergeron 's commit : http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/977/

notoraptor:
(bad jenkins link above, this is the last one in the other PR): http://earlgrey.iro.umontreal.ca:8080/job/Theano_PR/984/

notoraptor:
This PR can be closed, as it continues with #6496 .

PR #6495, opened 2017-10-23
Make sure MKL uses GNU OpenMP.

abergeron:


abergeron:
So this doesn't work for MKL 2017 so we need an alternate solution that handles this.

Unfortunately all means of checking the current version of MKL already load the library and prevent this fix from working.

Also I just realized that this is pointless if the user has already loaded numpy before theano since MKL is already loaded so it's too late.

abergeron:
I've amended the commit to be more extensive.  Also instead of attempting to fix the problem, we just raise an error if it is detected.

There is an undocumented flag to disable the check if required.

abergeron:
Just to check if the mac and windows checkers need special config.

jenkins test mac windows

abergeron:
jenkins test mac windows

theano/configdefaults.py
```diff
@@ -1238,6 +1238,29 @@ def is_valid_check_preallocated_output_param(param):
...
+    import os
+    if ('MKL_THREADING_LAYER' in os.environ and
+            os.environ['MKL_THREADING_LAYER'] == 'GNU'):
+        return
+    try:
+        import mkl
+        if '2018' in mkl.get_version_string():
+            raise RuntimeError('To use MKL 2018 with Theano you MUST set "MKL_THREADING_LAYER=GNU" in your environement.')
+    except ImportError:
+        raise RuntimeError("""
```
nouiz: I just want to clarity, with this PR, now the mkl package is mandatory with conda. If the user don't use conda, and the string "mkl" isn't in the flag, then he don't need to change his environment. Can you confirm that?

abergeron: If the string mkl isn't in the flag, then nothing changes.

However if you use mkl and are not in conda then you either need to obtain the mkl-service package somehow, set MKL_THREADING_LAYER=GNU or disable the check with blas.check_openmp=False

PR #6498, opened 2017-10-27
Prepare release candidate 1.0.0rc1

notoraptor:
@nouiz @lamblin

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -12,6 +12,11 @@ We recommend that every developer updates to this version.
 
 Highlights:
  - Make sure MKL uses GNU OpenMP
+
+   - **NB**: Matrix dot product (``gemm``) with numpy ``1.13`` and ``mkl`` from conda
+     could return wrong results in some cases. We have reported the problem upstream
+     and we have a work around that raises an error with information about how to fix it.
+
  - Optimized ``SUM(x^2)``, ``SUM(ABS(X))`` and ``MAX(ABS(X))`` operations with cuDNN reductions
  - Added Python scripts to help test cuDNN convolutions
  - Fixed invalid casts and index overflows in ``theano.tensor.signal.pool``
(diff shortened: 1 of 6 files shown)


NEWS_DEV.txt
```diff
@@ -30,6 +30,7 @@ Highlights:
  - Theano now internally uses ``sha256`` instead of ``md5`` to work on systems that forbide ``md5`` for security reason
  - Removed old GPU backend ``theano.sandbox.cuda``. New backend ``theano.gpuarray`` is now the official GPU backend
  - Support more debuggers for ``PdbBreakpoint``
+ - Make sure MKL uses GNU OpenMP
```
nouiz: I would add as a subpoint more details:

matrix dot product (gemm) with numpy 1.13 and mkl from conda could is some cases return wrong results. We have reported the problem upstream and have a work around to have an error raised with information how to fix it to the user.

NEWS.txt
```diff
@@ -12,6 +12,11 @@ We recommend that every developer updates to this version.
 
 Highlights:
  - Make sure MKL uses GNU OpenMP
+
+   - **NB**: Matrix dot product (``gemm``) with numpy ``1.13`` and ``mkl`` from conda
```
nouiz: I would do:
Matrix dot product (``gemm``) with numpy ``1.13`` and ``mkl`` from conda
->
Matrix dot product (``gemm``) with and ``mkl 2018`` from conda

I think this is independent of numpy and depend only on the mkl version. I just tried to install something and it tried to install numpy 1.13 with mkl 2017 in conda. 

@abergeron does that sound good to you?

notoraptor:
@nouiz Updated.

NEWS.txt
Gam3356: Willie Parish IPHONE 12 Pro max

PR #6506, opened 2017-11-06
Attempt to prevent re-initialization of the GPU in a child process

abergeron:


theano/gpuarray/__init__.py
```diff
@@ -58,6 +58,8 @@ def pygpu_parse_version(version_string):
 
 def init_dev(dev, name=None, preallocate=None):
     global pygpu_activated
+    if os.environ.get('THEANO_GPU_IS_ALREADY_ACTIVE', '') == 'Yes':
```
nouiz: I forgot, why we can import theano, but not the GPU in the subprocess?

Otherwise, I'm good with this PR. I restarted a travis run that failed due to a 10m timeout.

f0k:
Hmm, this means we can't have a Python process that imports Theano with `device=cpu`, fork it, and then use cuda0 in one and cuda1 in another subprocess (via `theano.gpuarray.use()`). Not sure if that's useful anyway, but is that intended? Which case do you need to prevent?

nouiz:
I think your use case is supported with this PR. This PR try to give a better error message for the case when we import theano, init the GPU and then fork. Forking before initializing the GPU should be allowed.

What make you think this case is blocked by this PR?

abergeron:
@f0k This use case would work since we only export the environment tag when the GPU is initialized.

What we are trying to prevent in reuse of the GPU stuff when the parent was already using it.  And the main reason we do that is that if fails in interesting ways due to it not being supported by the toolkit.

f0k:
> What make you think this case is blocked by this PR?

Nevermind, I mistakenly thought `os.environ` had some magic powers...

> What we are trying to prevent in reuse of the GPU stuff when the parent was already using it.

Right, the toolkit doesn't support forking after initializing CUDA (`cuInit()` in the driver API, which is implicitly called by most runtime API calls such as `cudaGetDeviceCount()`). Unfortunately it doesn't seem like there's a way for a process to check whether this has happened, or for a subprocess to undo it.

Thanks for clearing this up! But why do you use `os.environ` instead of some ordinary variable like `theano.gpuarray.is_active`, or the existing `theano.gpuarray.pygpu_activated`? Is there anything that can go wrong with the latter? Do you ever need to read this state outside of `theano/gpuarray/__init__.py`?

abergeron:
Because those get lost if you actually exec something.  However environment variables stay (most of the time).

One example where we had that problem was in one of the tests that would fork and execute some script for MPI.  This would re-initialize the GPU and hang on some machines.

PR #6508, opened 2017-11-09
Don't complain about MKL threading on mac since openmp doesn't work.

abergeron:


theano/configdefaults.py
```diff
@@ -1241,7 +1241,9 @@ def is_valid_check_preallocated_output_param(param):
 def check_mkl_openmp():
     if not theano.config.blas.check_openmp:
         return
-    import os
+    import os, sys
```
nouiz: why the import was here and why you extend it? Both of those import are already at the top of this file.

abergeron: I don't remember why it was there, but I extended it to be sure.

theano/configdefaults.py
```diff
@@ -1241,7 +1241,8 @@ def is_valid_check_preallocated_output_param(param):
 def check_mkl_openmp():
     if not theano.config.blas.check_openmp:
         return
-    import os
+    if sys.platform == 'darwin':
```
nouiz: Could we also check if this is a conda python or not? I don't think we have this logic

abergeron: We don't need that.

clang doesn't support OpenMP so even if you're not in conda you still won't have problems.

Or did you have something else in mind?

PR #6509, opened 2017-11-14
Err msg et small opt in python code

nouiz:


theano/gpuarray/dnn.py
```diff
@@ -97,12 +97,18 @@ def _dnn_lib():
                     if lib_name:
                         break
             if lib_name is None:
-                raise RuntimeError('Could not find cudnn library (looked for v5* to v7*)')
+                raise RuntimeError(
+                    'Could not find cudnn library (looked for v5* to v7*).'
+                    ' Check your cudnn installation. Maybe using the Theano'
+                    ' flag dnn.bin_path can help you. Current value "%s"' %
+                    config.dnn.bin_path)
```
abergeron: I'd prefer if dnn.base_path was referenced.  bin_path is for special cases.

theano/gpuarray/dnn.py
```diff
@@ -97,12 +97,18 @@ def _dnn_lib():
...
             else:
                 dnn_handle = ctypes.cdll.LoadLibrary(lib_name)
-
         if dnn_handle is None:
-            raise RuntimeError('Could not load cudnn library')
+            raise RuntimeError('Could not load cudnn library. Check your cudnn'
+                               ' installation. Maybe using the Theano'
+                               ' flag dnn.bin_path can help you. Current value "%s"' %
+                               config.dnn.bin_path)
```
abergeron: Same here

PR #6517, opened 2017-11-20
Suggest a fix (and tests) for gpudnnreduction when axes to reduce have size 1.

### LINKED ISSUE #6514: tensor.abs_() returning negative values in Theano 1.0.0
When I upgraded from Theano 0.9.0 to 1.0.0 my code just broke. I narrowed down the issue to this minimal working example:

```
import theano
import theano.tensor as T
import numpy as np

x = T.dmatrix('x')

f = theano.function([x], T.mean(T.abs_(x), axis=-1))

a = np.array([[-1.], [0.], [1.]])

print f(a)
```

In Theano 0.9.0, the result is: [1.  0.  1.]
In Theano 1.0.0, the result is: [-1.  0.  1.]

Curiously, Theano 1.0.0 gives [-1. 0. 1.] on the GPU, but [1.  0.  1.] on the CPU. Could this be an issue with libgpuarray?

notoraptor:
This should help fix #6514 .

@abergeron @nouiz @lamblin


theano/gpuarray/c_code/dnn_redux.c
```diff
@@ -97,6 +97,49 @@ int APPLY_SPECIFIC(dnn_redux)(PyGpuArrayObject *input,
...
+            arg.typecode = (*output)->ga.typecode;
+            arg.flags = GE_READ | GE_WRITE;
+            GpuElemwise* elemwise = GpuElemwise_new(c->ctx, "", "out = (out < 0 ? -out : out)", 1, &arg, p, GE_CONVERT_F16);
+            if (!elemwise) {
+                PyErr_SetString(PyExc_RuntimeError, "Unable to create GpuElemwise for output.");
+                return 1;
+            }
+            void* args[1] = { (void*)&(*output)->ga };
+            int err = GpuElemwise_call(elemwise, args, 0);
+            GpuElemwise_free(elemwise);
```
nouiz: I think creating one each time is costly. SO I think we should keep the GpuElemwise instance in the node instance.

theano/gpuarray/c_code/dnn_redux.c
```diff
@@ -97,6 +97,49 @@ int APPLY_SPECIFIC(dnn_redux)(PyGpuArrayObject *input,
       PyErr_Format(PyExc_RuntimeError, "GpuArray_reshape_inplace: %s", GpuArray_error(&(*output)->ga, err));
       return 1;
     }
+
+    if (rsz == 1 && cudnnGetVersion() <= 7004) {
```
nouiz: I think this is dangerous. We don't know when this support will be added in cudnn. So for new version of cudnn, we should do the work around I think. We should not keep the old behavior of returning wrong result. Only when we tested a new version that don't request this, we can remove the work around. (Or return an error for newer cudnn version, but I don't like this option)

theano/gpuarray/tests/test_dnn.py
```diff
@@ -1611,6 +1611,39 @@ def test_dnn_reduction_absmax():
         utt.assert_allclose(np.max(np.abs(M_val), axis=axis), f(M_val))
 
 
+def test_dnn_reduction_axis_size_one():
+    if not dnn.dnn_available(test_ctx_name) or dnn.version(raises=False) < 6000:
+        raise SkipTest(dnn.dnn_available.msg)
+    for dtype in ('float16', 'float32', 'float64'):
+        for shape, axis in [[(1, 2, 3), 0], [(2, 1, 3), 1], [(2, 3, 1), 2]]:
```
nouiz: Can you test with multiple reduction axis? like shape (1,2,3,4) and axis 0,2 and shape (1,2,3,1) and axis 0 and 3.

theano/gpuarray/c_code/dnn_redux.c
notoraptor: To create the GpuElemewise, I need the args (which typecode may vary across op calls) and the context. These infos would be hard to get from init_code_struct section, I think.

theano/gpuarray/tests/test_dnn.py
notoraptor: Updated.

**NB**: When trying to test multiple axes, I got some optimizations issues. Sometimes acc_dtype is None again. Sometimes, when reduced axes are consecutive, then there is an intermediate optimization for GpuCaReduceCuda, so that GpuDnnReduction is not introduced as expected. And sometimes the reduction pattern (e.g. `1010`) is not supported by GpuCaReduceCuda. I finally choosed patterns that are supported (`101`, `0101`).

nouiz: So it mean we don't always introduce it. The good fix would be to change how this is optimized, but we won't do it. The only problem is a bigger graph then needed.

theano/gpuarray/c_code/dnn_redux.c
nouiz: You can init it in the c_code() and just reuse it between Theano function call for each node.

notoraptor: ok

notoraptor: PS: Even if I init the GpuElemwise a first time, it depends on the output typecode, which can change from one call to another (as now we used op params and we don't explicitely check the dtypes (`check_input` is False for GpuDnnReduction)):
```C++
gpuelemwise_arg arg;
arg.name = "out";
arg.typecode = (*output)->ga.typecode; // <- HERE !
        // At next call, output typecode may be different, I think.
        //  But we need it to create the elemwise.
arg.flags = GE_READ | GE_WRITE;
GpuElemwise* elemwise = GpuElemwise_new(c->ctx, "", "out = (out < 0 ? -out : out)", 1, &arg, p, GE_CONVERT_F16);
```
Isn't it a problem ?

nouiz: The dtype is fixed for a node instance. Don't use one ptr per class, but per instance and it should work well. Don't forget to free it when we class is destroyed

PR #6530, opened 2017-12-01
Fix GPU initialization checking on buildbot.

notoraptor:
In the Windows buildbot, we test GPU initialization with `"import theano.gpuarray; theano.gpuarray.use('%DEVICE%')"`, But we set `THEANO_FLAGS=init_gpu_device=cuda` before (see top of file), thus `import theano.gpuarray` will automatically initialize the GPU, then the second instruction (`theano.gpuarray.use(%DEVICE%)`) will force a second initialization, that is forbidden by the @abergeron 's PR (first initialization will set OS env var `THEANO_GPU_IS_ALREADY_ACTIVE`, and second initialization will fail because that OS env var will be already set).

We must either: find a better way to distinguish mutiple initalization accross different processes vs multiiple initialization inside the same process, or update jenkins scripts to prevent multiple initialization inside the same process (I guess that case is not desired, neither). I choose to update the jenskins script for Windows buildbot.

PS: I guess some other Jenkins scripts should fail too, as they used the same strategy, but I did not yet change them.

@nouiz @abergeron @slefrancois @lamblin

### DIFF
--- a/.jenkins/jenkins_buildbot_windows.bat
+++ b/.jenkins/jenkins_buildbot_windows.bat
@@ -54,7 +54,9 @@ mkdir %BUILDBOT_DIR%
 echo "Directory of stdout/stderr %BUILDBOT_DIR%"
 
 REM Exit if theano.gpuarray import fails
-python -c "import theano.gpuarray; theano.gpuarray.use('%DEVICE%')" || exit /b
+REM NB: As we define `set THEANO_FLAGS=init_gpu_device=cuda` above,
+REM     importing theano.gpuarray will automatically initialize GPU.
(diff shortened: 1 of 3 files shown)


nouiz:
I think we should allow multiple init in the same process. Can you try to do that? Mostly, when the env var is created, keep a python variable that tell we created it. And if we created it, we ignore if the env var is present or not.

notoraptor:
@nouiz Done.

nouiz:
jenkins test windows mac

notoraptor:
jenkins test windows

notoraptor:
jenkins test windows

PR #6540, opened 2017-12-19
Fix tests in DebugMode and make them less verboase

nouiz:


theano/tensor/sort.py
```diff
@@ -309,14 +308,19 @@ class TopKOp(theano.Op):
 
     Notes
     -----
-    - By default, this Op give two outputs: values and indices. However optimizer may
-      remove a certain output if not needed.
-
-    - Computing gradient is only possible when both values and indices are computed in
+    - The CPU and GPU op have a different order of the elements. This is
```
notoraptor: I think I would rewrite this sentence with something like this: `CPU and GPU ops don't produce same output order. This is expected`.

theano/tensor/tests/test_opt.py
```diff
@@ -1163,7 +1163,8 @@ def my_init(shp, dtype='float64', num=0):
                 fail4.append((id, out_dtype, out.dtype))
 
         failed = len(fail1 + fail2 + fail3 + fail4)
-        print("Executed", len(cases), "cases", "failed", failed)
+        if failed > 0:
```
notoraptor: There is an identical `if` statement below, which raises an exception if failed > 0. I guess this `print` is not needed anymore ?

theano/gpuarray/sort.py
```diff
@@ -334,12 +360,18 @@ def local_gpua_topkop(op, ctx_name, inputs, outputs):
...
         idx_dtype=op.idx_dtype,
         return_values=rv,
         return_indices=ri)
-    rets = gpu_op(x, k)
+    rets = gpu_op(x, k, return_list=True)
+    c = ValuesEqApproxNoOrder(axis)
+    if rv or ri:
+        rets[0].tag.values_eq_approx = c
+    if rv and ri:
+        rets[1].tag.values_eq_approx = c
```
notoraptor: PS: maybe a suggestion here:
```python
if rv:
    rets[0].tag.values_eq_approx = c
if ri:
    rets[1].tag.values_eq_approx = c
```

nouiz: Both are indenpendent. I'll loop over rets to make this more readable.

theano/tensor/tests/test_opt.py
nouiz: right.

theano/tensor/sort.py
nouiz: ok.

theano/gof/utils.py
```diff
@@ -129,7 +129,9 @@ def get_variable_trace_string(v):
         else:
             # Print separate message for each element in the list of
             # batcktraces
```
ReyhaneAskari: "batcktraces" - > backtraces

theano/gpuarray/opt.py
```diff
@@ -183,7 +183,11 @@ def f(local_opt):
 # in order to avoid introducin new CPU Ops, or useless ones.
 def safe_to_gpu(x, ctx_name):
     if isinstance(x.type, tensor.TensorType):
-        return GpuFromHost(ctx_name)(x)
+        ret = GpuFromHost(ctx_name)(x)
+        values_eq_approx = getattr(x.tag, 'values_eq_approx', None)
+        if values_eq_approx:
+            ret.tag.values_eq_approx = values_eq_approx
```
ReyhaneAskari: I think this is good, but maybe we might want to do the tag transfer in a more general level. Like in the transfer functions such as `GpuFromHost` or `transfer`.

theano/gpuarray/opt.py
```diff
@@ -194,7 +194,11 @@ def safe_to_gpu(x, ctx_name):
 
 def safe_to_cpu(x):
     if isinstance(x.type, GpuArrayType):
-        return x.transfer('cpu')
+        ret = x.transfer('cpu')
```
ReyhaneAskari: same here.

theano/gof/utils.py
nouiz: ok

theano/gpuarray/opt.py
nouiz: I did it. Good idea.

Issue #6518, opened 2017-11-20
multiple tests are failing due to "error: no matching function for call to 'batch_gemm<float>(<unresolved overloaded function type>, int&, PyArrayObject*&, PyArrayObject*&, PyArrayObject*&)'"


ignatenkobrain:
```python
======================================================================
ERROR: test_bad_runtime (theano.tensor.tests.test_basic.BatchedDotTester)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/configparser.py", line 117, in res
    return f(*args, **kwargs)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/tensor/tests/test_basic.py", line 462, in test_bad_runtime
    f = inplace_func([], node.outputs, mode=mode, name="test_bad_runtime")
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/tensor/tests/test_basic.py", line 106, in inplace_func
    name=name)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/compile/function.py", line 317, in function
    output_keys=output_keys)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/compile/pfunc.py", line 486, in pfunc
    output_keys=output_keys)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/compile/function_module.py", line 1841, in orig_function
    fn = m.create(defaults)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/compile/function_module.py", line 1715, in create
    input_storage=input_storage_lists, storage_map=storage_map)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/gof/link.py", line 699, in make_thunk
    storage_map=storage_map)[:3]
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/gof/vm.py", line 1084, in make_all
    impl=impl))
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/gof/op.py", line 955, in make_thunk
    no_recycling)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/gof/op.py", line 858, in make_c_thunk
    output_storage=node_output_storage)
  File "/builddir/build/BUILD/Theano-rel-1.0.0/theano/gof/cc.py", line 1217, in make_thunk
    keep_lock=keep_lock)
[...]

PR #6549, opened 2018-01-10
Remove a template argument from batch_gemm

abergeron:
This might in some case where the compiler got confused and could not
infer the right type for the gemm function.

Might help for #6518

### DIFF
--- a/theano/tensor/blas.py
+++ b/theano/tensor/blas.py
@@ -2054,9 +2054,10 @@ class BatchedDot(Op):
 
     def c_support_code(self):
         batch_gemm_defn = """
-        template<typename dtype, typename function>
-        bool batch_gemm(function gemm, int type_size,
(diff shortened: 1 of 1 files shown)


nouiz:
jenkins test mac windows

nouiz:
@abergeron Mac and Windows CI failed.

abergeron:
This looks like a build machine failure, not anything wrong with this PR.

jenkins test mac windows

nouiz:
jenkins test windows

slefrancois:
jenkins test win

abergeron:
The windows machine seems to be having problems right now.  I think we can ignore those results.

In any case I don't expect any additional fallout from windows so since it passed on linux I think we're good to go.

slefrancois:
jenkins test win

nouiz:
jenkins test windows

slefrancois:
jenkins test windows

slefrancois:
jenkins test windows

PR #6557, opened 2018-01-30
Switched lambda to partial in EighGrad

jordan-melendez:
Partial is required to make pickling possible. This is required in PyMC3 and [this](https://github.com/pymc-devs/pymc3/pull/2731/) PR in particular.


nouiz:
thanks

PR #6558, opened 2018-02-01
Add a script to help test GPU power consumption.

nouiz:


theano/misc/burn_gpu.py
```diff
@@ -0,0 +1,46 @@
...
+
+import theano
+import theano.tensor as T
+from theano.gpuarray import dnn
+from theano.tensor.nnet.abstract_conv import get_conv_output_shape
+
+
+def burn():
+    # This test that we have inplace work correctly even when
+    # GpuAllocEmpty get merged together.
```
abergeron: This comment might not be relevant here.

abergeron:
Apart from the comment, I'm ok with this.  If you don't think it's an issue, you can merge as-is.

PR #6569, opened 2018-03-02
Add a map of values_approx functions from cpu to gpu.

abergeron:
This should fix the problems in DEBUG_MODE.

### DIFF
--- a/theano/gpuarray/type.py
+++ b/theano/gpuarray/type.py
@@ -360,33 +360,8 @@ class GpuArrayType(Type):
     def values_eq_approx(a, b,
                          allow_remove_inf=False, allow_remove_nan=False,
                          rtol=None, atol=None):
-        if a.shape != b.shape or a.dtype != b.dtype:
-            return False
-        if str(a.dtype) in theano.tensor.discrete_dtypes:
-            return GpuArrayType.values_eq(a, b)
-        else:
-            if allow_remove_inf or allow_remove_nan:
-                raise NotImplementedError(
-                    "GpuArrayType.values_eq_approx() don't implemented the"
-                    " allow_remove_inf and allow_remove_nan parameter")
-            atol_, rtol_ = theano.tensor.basic._get_atol_rtol(a, b)
-            if rtol is not None:
-                rtol_ = rtol
-            if atol is not None:
-                atol_ = atol
-            res = elemwise2(a, '', b, a, odtype=np.dtype('bool'),
-                            op_tmpl="res = (fabs(a - b) <"
-                            "(%(atol_)s + %(rtol_)s * fabs(b)))" %
-                            locals())
-            ret = np.asarray(res).all()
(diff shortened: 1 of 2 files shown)


lamblin:
Jenkins seems to be full of the following message:
```
[Theano_GPU] ERROR (theano.gof.opt): Optimization failure due to: local_abstractconv_check
[Theano_GPU] ERROR (theano.gof.opt): node: AbstractConv2d{convdim=2, border_mode='valid', subsample=(1, 1), filter_flip=True, imshp=(10, 3, 5, 5), kshp=(12, 3, 3, 3), filter_dilation=(1, 1), num_groups=1, unshared=False}(GpuReshape{4}.0, GpuReshape{4}.0)
[Theano_GPU] ERROR (theano.gof.opt): TRACEBACK:
[Theano_GPU] ERROR (theano.gof.opt): Traceback (most recent call last):
[Theano_GPU]   File "/home/jenkins/workspace/Theano_PR/theano/gof/opt.py", line 2034, in process_node
[Theano_GPU]     replacements = lopt.transform(node)
[Theano_GPU]   File "/home/jenkins/workspace/Theano_PR/theano/tensor/nnet/opt.py", line 500, in local_abstractconv_check
[Theano_GPU]     node.op.__class__.__name__)
[Theano_GPU] LocalMetaOptimizerSkipAssertionError: AbstractConv2d Theano optimization failed: there is no implementation available supporting the requested options. Did you exclude both "conv_dnn" and "conv_gemm" from the optimizer? If on GPU, is cuDNN available and does the GPU support it? If on CPU, do you have a BLAS library installed Theano can link against? On the CPU we do not support float16.
```
I'm not sure where that's coming from.

nouiz:
I'm ok with this PR. @abergeron, can you check if what @lamblin found was already there before this PR? If so, we could merge this PR now.

abergeron:
That logspam from the tests is there since at least 0.9, probably even before.  It's not actually an error in the tests themselves.

theano/gpuarray/type.py
```diff
@@ -542,6 +517,65 @@ def c_code_cache_version(self):
...
+                     rtol=None, atol=None):
+    if a.shape != b.shape or a.dtype != b.dtype:
+        return False
+    if str(a.dtype) in theano.tensor.discrete_dtypes:
+        return GpuArrayType.values_eq(a, b)
+    else:
+        if allow_remove_inf or allow_remove_nan:
+            raise NotImplementedError(
+                "GpuArrayType.values_eq_approx() don't implemented the"
+                " allow_remove_inf and allow_remove_nan parameter")
```
lamblin: Maybe we only have to remove that error, and jump directly to the section after `# maybe the trouble is that there are NaNs`?

abergeron: Yes, I think that could work.

PR #6571, opened 2018-03-06
Fix non-cudnn reductions that were all subtly broken by 91bc16c3.

abergeron:
This was particularly visible for float16 since they would just stop working, but it also caused a reduction in precision for all other data types which could have lead to wrong results.

This was only avoided by the fact that we use cudnn all the time so only DEBUG_MODE was able to sniff this out since it runs the intermediate graph before we switch to cudnn reductions.

91bc16c3

### DIFF
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3783,7 +3783,7 @@ def local_dnn_reduction(node):
     with inherit_stack_trace(node.outputs):
         ret = GpuDnnReduction(scal,
                               node.op.axis,
-                              node.op.acc_dtype,
+                              node.op._acc_dtype(node.inputs[0].dtype),
                               node.op.dtype,
                               False)(node.inputs[0])
         return [post(ret)]
--- a/theano/gpuarray/dnn.py
+++ b/theano/gpuarray/dnn.py
@@ -3729,27 +3729,26 @@ def local_dnn_reduction(node):
     if node.inputs[0].ndim > 8:
         return
 
+    acc_dtype = node.op._acc_dtype(node.inputs[0].dtype)
+
     if node.inputs[0].dtype != node.outputs[0].dtype:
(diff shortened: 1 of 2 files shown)


abergeron:
This fixes the DEBUG_MODE problem for me.

theano/gpuarray/dnn.py
```diff
@@ -3783,7 +3783,7 @@ def _square(a):
     with inherit_stack_trace(node.outputs):
         ret = GpuDnnReduction(scal,
                               node.op.axis,
-                              node.op.acc_dtype,
+                              node.op._acc_dtype(node.inputs[0].dtype),
```
nouiz: I'm not convinced of this change. What if the user specified a different acc_dtype?

theano/gpuarray/dnn.py
```diff
@@ -3729,27 +3729,26 @@ def local_dnn_reduction(node):
     if node.inputs[0].ndim > 8:
         return
 
+    acc_dtype = node.op._acc_dtype(node.inputs[0].dtype)
```
nouiz: Ignoring the original acc_dtype is not good I think. If there isn,t one provided, it is good.

theano/gpuarray/dnn.py
abergeron: This is superseded by the following commit.

theano/gpuarray/dnn.py
abergeron: We don't ignore the original.  `_acc_dtype` takes it into account : https://github.com/Theano/Theano/blob/42275cba95851ecca4d5580af9fe8bcbd2a3f481/theano/tensor/elemwise.py#L1747

Anyway, this is also what the non-cudnn code does so I don't see what's wrong.

nouiz: thanks for the explanation. Merging.

PR #6574, opened 2018-03-10
Make Theano work under pypy (WIP)

GSam:
In trying to find an ML library to work on pypy, I figured out some hacks to get it working (at least superficially with runtime compiled cpyext modules). 

There seems to be a few problems that need to be resolved before it would work out of the box:

1. A single usage of the ctypes.pythonapi.PyCapsule_Type

2. A single usage of sys.getrefcount (to check that there isn't leaks)

3. Missing build information to runtime load the compiled .so python module
- The correct binary .so filename looks to be able to be constructed using distutils.sysconfig.get_config_var("SO") (or EXT_SUFFIX in python3)
- Expects libpython.so (or equivalent) in the distutils.sysconfig.get_config_var("LIBDIR") location which does not exist
- distutils.sysconfig.get_config_var("LDLIBRARY") does not resolve in pypy currently -- but should in theory point to libpypy-c

### DIFF
--- a/theano/gof/cmodule.py
+++ b/theano/gof/cmodule.py
@@ -269,16 +269,23 @@ def dlimport(fullpath, suffix=None):
     if not os.path.isabs(fullpath):
         raise ValueError('`fullpath` must be an absolute path', fullpath)
     if suffix is None:
-        if fullpath.endswith('.so'):
-            suffix = '.so'
-        elif fullpath.endswith('.pyd'):
-            suffix = '.pyd'
-        elif fullpath.endswith('.dll'):
-            suffix = '.dll'
-        elif fullpath.endswith('.py'):
-            suffix = '.py'
-        else:
-            suffix = ''
+        suffix = ''
+
+        dist_suffix = distutils.sysconfig.get_config_var("SO")
+        if dist_suffix is not None and dist_suffix != '':
+            if fullpath.endswith(dist_suffix):
+                suffix = dist_suffix
+
+        if suffix == '':
+            if fullpath.endswith('.so'):
+                suffix = '.so'
+            elif fullpath.endswith('.pyd'):
+                suffix = '.pyd'
+            elif fullpath.endswith('.dll'):
+                suffix = '.dll'
+            elif fullpath.endswith('.py'):
+                suffix = '.py'
+
     rval = None
     if fullpath.endswith(suffix):
         module_name = '.'.join(fullpath.split(os.path.sep)[-2:])[:-len(suffix)]
@@ -1675,22 +1682,30 @@ def std_lib_dirs_and_libs():
     elif sys.platform == 'darwin':
         std_lib_dirs_and_libs.data = [], []
     else:
-        # assume Linux
-        # Typical include directory: /usr/include/python2.6
+        if platform.python_implementation() == 'PyPy':
+            # Assume Linux (note: Ubuntu doesn't ship this .so)
+            libname = "pypy-c"
+            # Unfortunately the only convention of this .so is that it appears
+            # next to the location of the interpreter binary.
+            libdir = os.path.dirname(os.path.realpath(sys.executable))
+        else:
+            # Assume Linux
+            # Typical include directory: /usr/include/python2.6
 
-        # get the name of the python library (shared object)
-        libname = distutils.sysconfig.get_config_var("LDLIBRARY")
+            # get the name of the python library (shared object)
 
-        if libname.startswith("lib"):
-            libname = libname[3:]
+            libname = distutils.sysconfig.get_config_var("LDLIBRARY")
 
-        # remove extension if present
-        if libname.endswith(".so"):
-            libname = libname[:-3]
-        elif libname.endswith(".a"):
-            libname = libname[:-2]
+            if libname.startswith("lib"):
+                libname = libname[3:]
 
-        libdir = distutils.sysconfig.get_config_var("LIBDIR")
+            # remove extension if present
+            if libname.endswith(".so"):
+                libname = libname[:-3]
+            elif libname.endswith(".a"):
+                libname = libname[:-2]
+
+            libdir = distutils.sysconfig.get_config_var("LIBDIR")
 
         std_lib_dirs_and_libs.data = [libname], [libdir]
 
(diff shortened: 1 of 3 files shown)


GSam:
It seems that the sys.getrefcount could be made optional using platform.python_implementation() -- returns 'CPython' or 'PyPy'.

GSam:
libpypy-c.so also appears to reside by default in the bin directory next to the pypy binary. While I can notice distros seem to also put it here, Ubuntu noticeably doesn't appear to build with --shared to have the library.

nouiz:
I like your idea for getrefcount.

For libpypy-c.so, if the fix work not in all cases, this can still be merged.  Can you make the modification you suggested?

theano/gof/cmodule.py
```diff
@@ -269,7 +269,9 @@ def dlimport(fullpath, suffix=None):
     if not os.path.isabs(fullpath):
         raise ValueError('`fullpath` must be an absolute path', fullpath)
     if suffix is None:
-        if fullpath.endswith('.so'):
+        if fullpath.endswith('.pypy-41.so'):
```
nouiz: For this, why not use Python regex and if fullpath end with ".pypy*.so", we reuse that as the suffix? This won't be version dependant that way.

theano/gof/cmodule.py
```diff
@@ -1679,15 +1681,19 @@ def std_lib_dirs_and_libs():
...
-            libname = libname[:-2]
+            # remove extension if present
+            if libname.endswith(".so"):
+                libname = libname[:-3]
+            elif libname.endswith(".a"):
+                libname = libname[:-2]
+        else:
+            libname = "pypy-c"
 
+        # Must create a copy of pypy-c in prefix/lib
```
nouiz: I do not understand this comment. Can you explain more? Is that a modification the user must do himself or this is already done automatically?

theano/gof/cmodule.py
```diff
@@ -1679,15 +1681,19 @@ def std_lib_dirs_and_libs():
...
-        if libname.endswith(".so"):
-            libname = libname[:-3]
-        elif libname.endswith(".a"):
-            libname = libname[:-2]
+            # remove extension if present
+            if libname.endswith(".so"):
+                libname = libname[:-3]
+            elif libname.endswith(".a"):
+                libname = libname[:-2]
+        else:
```
nouiz: Here, to be safer, I would go with something like:
elif ((this is a pypy interpretor, you described how to do it.):
   libname = "pypy-c"

theano/gof/cmodule.py
```diff
@@ -2277,7 +2283,7 @@ def compile_str(module_name, src_code, location=None,
 
         lib_filename = os.path.join(
             location,
-            '%s.%s' % (module_name, get_lib_extension()))
+            '%s.%s.%s' % (module_name, "pypy-41", get_lib_extension()))
```
nouiz: I would do something like this:
pypy_name = ""
if pypy:
    pypy_name=get_the_pypy_name with regex as talked above + "."
lib_filename = os.path.join(
             location,
            '%s.%s%s' % (module_name, pypy_name, get_lib_extension()))

theano/gof/type.py
```diff
@@ -606,8 +606,8 @@ def __str__(self):
 
 generic = Generic()
 
-_cdata_type = ctypes.py_object.from_address(
-    ctypes.addressof(ctypes.pythonapi.PyCapsule_Type)).value
+#_cdata_type = ctypes.py_object.from_address(
```
nouiz: What about:
_cdata_type = None
if not pypy:
   _cdata_type = ...

This would remove the check, but should make it work.

theano/gof/cmodule.py
GSam: To originally minimize the changes, I just symlinked the .so into a manually created lib folder in the right location. From looking around, it seems like the only out-of-the-box compatible solution at the moment would be to search for the .so in the sys.executable (bin) directory if we detect we are running PyPy.

GSam:
I've tested this with the official tarballs of PyPy 2 and 3. I expect it should work with at least one of the major distros (in terms of locating the libpypy-c.so binary). I've run it with some examples from http://deeplearning.net/software/theano/tutorial/examples.html and so far so good. Presumably this is all run on the CPU. Would be interesting to figure out if the GPU stuff works too.

theano/gof/type.py
```diff
@@ -606,8 +607,11 @@ def __str__(self):
 
 generic = Generic()
 
-_cdata_type = ctypes.py_object.from_address(
-    ctypes.addressof(ctypes.pythonapi.PyCapsule_Type)).value
+_cdata_type = None
+
+if platform.python_implementation() != 'PyPy':
+    _cdata_type = ctypes.py_object.from_address(
+        ctypes.addressof(ctypes.pythonapi.PyCapsule_Type)).value
```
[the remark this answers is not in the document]
GSam: I'll see if I can run it on a GPU and see exactly what happens.

GSam:
Surprisingly it also works on the GPU too (using the example in http://deeplearning.net/software/theano/tutorial/using_gpu.html).

$ LD_LIBRARY_PATH=/usr/local/cuda-9.1/lib64/ THEANO_FLAGS='mode=FAST_RUN,device=cuda,floatX=float32' python /tmp/check.py
[GpuElemwise{exp,no_inplace}(<GpuArrayType<None>(float32, vector)>), HostFromGpu(gpuarray)(GpuElemwise{exp,no_inplace}.0)]
Looping 1000 times took 0.735148 seconds
Result is [1.2317803 1.6187935 1.5227807 ... 2.2077181 2.2996776 1.623233 ]
Used the gpu

$ LD_LIBRARY_PATH=/usr/local/cuda-9.1/lib64/ THEANO_FLAGS='mode=FAST_RUN,device=cpu,floatX=float32' python /tmp/check.py
[Elemwise{exp,no_inplace}(<TensorType(float32, vector)>)]
Looping 1000 times took 4.794002 seconds
Result is [1.2317803 1.6187934 1.5227807 ... 2.2077181 2.2996776 1.6232328]
Used the cpu

It's complaining about cudnn.h which I haven't installed, but everything seems to be working just fine (with a 7x speed boost no less). I was initially afraid that libgpuarray and pygpu wouldn't install ok, but apparently Cython seems to work just fine in PyPy.

abergeron:
All right then, I'm good.

PR #6583, opened 2018-03-17
Support more recent flake8 version

nouiz:
Note, I'M not sure about the change to setup.py


setup.py
```diff
@@ -101,7 +101,7 @@ def do_setup():
           install_requires=['numpy>=1.9.1', 'scipy>=0.14', 'six>=1.9.0'],
           # pygments is a dependency for Sphinx code highlight
           extras_require={
-              'test': ['nose>=1.3.0', 'parameterized', 'flake8<3'],
+              'test': ['nose>=1.3.0', 'parameterized', 'flake8<3,>3.3'],
```
VargMon: Maybe try `'flake8<3,>=3'`
B^)

nouiz: thanks for the suggestion. "<3,>=3.3" would have been better, as I tested that version and it worked.

I just taught, why not check the older version documentation. Version 3.0.0 should work based on the documentation, so I moved the version restriction.

Issue #6580, opened 2018-03-16
Remove MKL_THREADING_LAYER=GNU not present error with newer NumPy version


nouiz:
Our report about the bug needing MKL_THREADING_LAYER=GNU this work around got followed up by a fix in https://github.com/AnacondaRecipes/numpy-feedstock/pull/5, we do not need that error anymore with mkl 2018. Those version of numpy include the fix:

numpy 1.13.3, build 3 or higher
numpy 1.14.1, build 2 or above
numpy 1.14.2, all versions

I was told that we could detect if the fix is present like this:

"Detecting it based on version won't really be possible, because numpy itself has no idea about build numbers.  You should instead be able to test for the existence of the _mklinit module in numpy.  Please see Sasha's PR at https://github.com/AnacondaRecipes/numpy-feedstock/pull/5 for more information."

Also:

"Also, import numpy with MKL_VERBOSE set to 1 should print some diagnostics, like so:

(idp_py3_dev) [09:03:34 linvm native]$ MKL_VERBOSE=1 python -c "import numpy"
Numpy + Intel(R) MKL: THREADING LAYER: (null)
Numpy + Intel(R) MKL: setting Intel(R) MKL to use INTEL OpenMP runtime
Numpy + Intel(R) MKL: preloading libiomp5.so runtime
"

Removing this error would ease the installation of Theano, especially for people on Windows that are less used to set variable environment.

twiecki:
@nouiz Could you give a few more hints as to what needs to change where?

nouiz:
What come to mind is to mostly grep for  MKL_THREADING_LAYER=GNU in Theano and find where the error is raised. Then at that place, change the condition such that newer version of numpy do not trigger the error. This should be tested with "You should instead be able to test for the existence of the _mklinit module in numpy."

Maybe also update the error message to tell they can update NumPy in conda to remove the need of the env variable.

abergeron:
This is done

PR #6589, opened 2018-03-29
Don't ask for MKL_THREADING_LAYER=GNU if NumPy is fixed

abergeron:


abergeron:
Should fix #6580

theano/configdefaults.py
```diff
@@ -1246,6 +1246,11 @@ def check_mkl_openmp():
...
             os.environ['MKL_THREADING_LAYER'] == 'GNU'):
         return
+    try:
+        import numpy._mklinit  # noqa
+        return
+    except ImportError:
+        pass
     try:
         import mkl
         if '2018' in mkl.get_version_string():
```
lamblin: I would update the error message below to mention updating the numpy conda package as the first option, rather than only setting MLK_THREADING, since it would fix the problem as well.

abergeron: done

theano/configdefaults.py
```diff
@@ -1254,11 +1254,16 @@ def check_mkl_openmp():
...
+To use MKL 2018 with Theano either update the numpy conda packages to
+their latest build or set "MKL_THREADING_LAYER=GNU" in your
+environment.
+""")
     except ImportError:
         raise RuntimeError("""
-Could not import 'mkl'.  Either install mkl-service with conda or set
-MKL_THREADING_LAYER=GNU in your environment for MKL 2018.
+Could not import 'mkl'.  If you are using conda, update the numpy
+packages to the latest build otherwise, set MKL_THREADING_LAYER=GNU in
```
lamblin: I'm not sure updating numpy is the right solution here, especially if they have the non-mkl version.

abergeron: We only call this function if we attempt to link with MKL.  So the non-mkl case is not an issue.

abergeron: Also, if they have the mkl version and update the package to a version with the _mklinit module, then we don't care if they have mkl-service installed or not (at least from this point of view).

nouiz: I agree with @abergeron answer, so I'll approve this PR. But I'll wait for @lamblin to see this before merging.

lamblin:
OK, I trust you :)

PR #6597, opened 2018-04-20
Fix MissingInputError when scan use hidden variable to calculate condition

### LINKED ISSUE #6595: MissingInputError when scan use hidden variable to calculate condition
When I reading code in scan, I found if variable is used by condition and not be declared in 'non_sequences'， a MissingInputError will be raised.
So I write a test case:

```python
def test_condition_hidden_inp():

    max_value = theano.tensor.scalar("max_value")
    n_steps = theano.tensor.iscalar("n_steps")

    def accum(prev_value, step):
        new_value = prev_value + step
        new_step = step + 1
        condition = theano.scan_module.until(new_value > max_value)
        return [new_value, new_step], condition

    rs, updates = theano.scan(
        fn=accum,
        outputs_info=[0, 0],
        n_steps=n_steps)

    f = theano.function(
        inputs=[max_value, n_steps],
        outputs=rs)

    _sum, total_steps = f(100, 100)
```

I found it's easy to fix.
https://github.com/Theano/Theano/blob/f2a2df153871470ce7d33082202e805af8f508a7/theano/scan_module/scan.py#L820-L840

Just remove line 839-840, and add 
```python
    if condition is not None:
        outputs.append(condition)
```
at line 823.

I'm not sure if it's written this way on purpose or a bug.

Glexin:
Fix #6595

### DIFF
--- a/theano/scan_module/scan.py
+++ b/theano/scan_module/scan.py
@@ -820,6 +820,8 @@ def scan(fn,
 
     # extract still missing inputs (there still might be so) and add them
     # as non sequences at the end of our args
+    if condition is not None:
+        outputs.append(condition)
     fake_nonseqs = [x.type() for x in non_seqs]
     fake_outputs = scan_utils.clone(outputs,
                                     replace=OrderedDict(izip(non_seqs,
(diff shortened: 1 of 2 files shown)


lamblin:
Jenkins test this please.

PR #6617, opened 2018-06-26
Don't use the dnn precision as the dtype of inputs

abergeron:


abergeron:
This has been confirmed to fix the tensorcore case.

theano/gpuarray/dnn.py
```diff
@@ -343,17 +343,20 @@ def version(raises=True):
 
 
 def get_precision(precision, inputs, for_grad=False):
+    common_dtype = theano.scalar.upcast(*[i.dtype for i in inputs])
+    if not common_dtype.startswith('float'):
+        raise TypeError("cuDNN convolution only works on floats")
```
nouiz: "on floats" -> "on real numbers". Float in C specify float32.

abergeron: ok

Issue #6616, opened 2018-06-24
OpFromGraph too restrictive on gradient types


nlgranger:
When overriding the gradient in OpFromGraph, Theano expects to get a gradient for all outputs even if those outputs are integers or disconnected.

Minimal working example:

```
import theano
import theano.tensor as T
import numpy as np


def f1(x):
    return x + 1, T.cast(x + 1, "int32")

def f1_back(inputs, output_gradients):
    return [output_gradients[0]]


def f2(x1, x2):
    return x1


def f2_back(input, output_gradients):
    return [
        output_gradients[0],
        theano.gradient.disconnected_type()
        # even faking some value won't work:
        # T.ones_like(output_gradients[0])
        # T.cast(T.ones_like(output_gradients[0]), "int32")
    ]


a = T.vector()

op1 = theano.OpFromGraph(
    inputs=[a],
    outputs=list(f1(a)),
    grad_overrides=f1_back,
    inline=True, name="op1")

a = T.vector()
b = T.ivector()

op2 = theano.OpFromGraph(
    inputs=[a, b],
    outputs=[f2(a, b)],
    grad_overrides=f2_back,
    inline=True, name="op2")


x = T.vector()
y1, y2 = op1(x)
z = op2(y1, y2)

theano.grad(z.sum(), [x]).eval({x: np.ones((5,), dtype=np.float32)})
```

To give a use case, imagine side results from the forward are needed to compute the gradient, one way to do this is to make two Op:

- op1 that computes the output and side results
- op2 that forwards the result and has the backpropagation computation

op1 just takes the computed gradient and back-propagate it to its input.

nlgranger:
To get an idea of the problem on a more concrete example, you can see a hack [here](https://github.com/nlgranger/Recipes/blob/connectionist_temporal_classification/papers/connectionist_temporal_classification/ctc.py#L260) where I need to trick Theano into thinking that all inputs are actually used (add x.max() * 0) and provide a dummy gradient values for those continous inputs that do not actually need a gradient.

abergeron:
I don't see how we could do anything but require the a gradient for all outputs.  What would you want to see here?

You have the right approach in using disconnected_type and/or zeros_like to ignore gradients for outputs that don't have one.

As for tricking Theano into thinking inputs are used, I fail to see how this is necessary.  I tried a simple example with unused inputs and there are no problems.  If you have a reproducible case that causes problems, I can look at it.

nlgranger:
[...]

PR #6618, opened 2018-06-26
Add option to override connection_pattern for OpFromGraph

abergeron:


nouiz:
I added from the comment https://github.com/Theano/Theano/issues/6616#issuecomment-400430281 as a test. If all CI pass, I'll merge.

PR #6624, opened 2018-07-18
Make sure to always pass num_groups

abergeron:


theano/gpuarray/dnn.py
```diff
@@ -1213,8 +1215,9 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1
         out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*out_shp)
         precision, _ = get_precision(precision, [img, kerns], for_grad=True)
         desc = GpuDnnConvDesc(border_mode='valid', subsample=(1, 1, 1), dilation=(1, 1, 1),
+                              num_groups=num_groups,
```
nouiz: This is useless, line 1200 assert that num_groups is 1

theano/gpuarray/dnn.py
```diff
@@ -1213,8 +1215,9 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1
...
                               conv_mode='cross', precision=precision)(out.shape)
-        conv = GpuDnnConvGradW()(img, kerns, out, desc)
+        conv = GpuDnnConvGradW(num_groups=num_groups)(img, kerns, out, desc)
```

theano/gpuarray/dnn.py
```diff
@@ -1234,8 +1237,9 @@ def dnn_conv3d(img, kerns, border_mode='valid', subsample=(1, 1, 1), dilation=(1
         out = GpuAllocEmpty(dtype=img.dtype, context_name=ctx_name)(*out_shp)
         precision, _ = get_precision(precision, [img, kerns], for_grad=True)
         desc = GpuDnnConvDesc(border_mode='valid', subsample=(1, 1, 1), dilation=dilation,
+                              num_groups=num_groups,
                               conv_mode=conv_mode, precision=precision)(kerns.shape)
-        return GpuDnnConvGradI()(kerns, img, out, desc)
+        return GpuDnnConvGradI(num_groups=num_groups)(kerns, img, out, desc)
```
nouiz: idem but line 1224

theano/gpuarray/dnn.py
```diff
@@ -3435,9 +3439,11 @@ def local_abstractconv3d_cudnn_alt(node):
                                   subsample=subsample,
                                   dilation=filter_dilation,
                                   conv_mode='cross',
+                                  num_groups=num_groups,
                                   precision=precision)(out.shape)
 
-            conv = GpuDnnConv(algo=None)(img, topgrad, out, desc)
+            conv = GpuDnnConv(algo=None, num_groups=num_groups)(
```

theano/gpuarray/dnn.py
```diff
@@ -3455,6 +3461,7 @@ def local_abstractconv3d_cudnn_alt(node):
                                   subsample=subsample,
                                   dilation=filter_dilation,
                                   conv_mode=conv_mode,
+                                  num_groups=num_groups,
```

theano/gpuarray/dnn.py
abergeron: Yes, but for future inspection and making sure that num_groups is passed correctly it's much easier to always pass it than to depend on a condition some place above that makes sure it is 1.

PR #6629, opened 2018-07-26
Don't return int from subtensor grads

abergeron:


nouiz:
Can you check for other ops if we tested this or not? If so, can you copy the test? Otherwise we can merge.

abergeron:
It doesn't look like there are any tests for that for any of the ops.

Issue #6626, opened 2018-07-18
Scan broken with python 3.7


abergeron:
Some of the internal C structures have changed names with python 3.7 and the scan C code no longer compiles.

This might be fixed by regenerating the scan code with a very recent version of cython.

In the meantime, stick with python 3.6.

PR #6640, opened 2018-09-13
Python 3.7 compatibility: Regenerate scan_perform.c with more recent cython.

twiecki:
Attempt to fix https://github.com/Theano/Theano/issues/6626.

I couldn't really test this as the unittests passed fine even on Python 3.7.


ColCarroll:
The PyMC3 test suite appears to fail on Python 3.7 (I haven't confirmed this). Perhaps test there?

twiecki:
Turns out I ran the test suite wrong and now confirmed that theano tests do not pass in python 3.7 without this, but do pass with this fix.

abergeron:
This looks good.

I'm curious, did you have to apply the numpy api patch or did cython finally fixed their binding?

nouiz:
When you ran the tests, was it only on CPU or it would have also tested GPU?

twiecki:
Only on CPU.

nouiz:
Ok. So not enough to support it officially.

Le ven. 14 sept. 2018 03:48, Thomas Wiecki <notifications@github.com> a
écrit :

abergeron:
From my limited tests before, that was the only thing missing.  But we could also add it to the buildbots to see if it works.

twiecki:
@abergeron That'd be great.

springcoil:
@twiecki How did you figure out all of this C stuff?

twiecki:
This is just a cython recompilation of the .pyx file.

springcoil:
Ahh - I thought you did it manually :)

PR #6644, opened 2018-09-18
Added support for the axis parameter of numpy.unique

lucianopaz:
Since `numpy` version 1.13.0, the `unique` function also takes the input parameter `axis`. This PR adds support for the `axis` parameter for the extra op `Unique`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,15 +63,9 @@ jobs:
       python: "2.7"
       env: NUMPY_VERSION=1.9.1 PART="theano/compat theano/compile theano/d3viz theano/gof theano/misc theano/sandbox theano/scalar theano/scan_module theano/tests -e test_flake8.py theano/typed_list"
     - <<: *normaltest
-      python: "2.7"
-      env: NUMPY_VERSION=1.13.1 PART="theano/compat theano/compile theano/d3viz theano/gof theano/misc theano/sandbox theano/scalar theano/scan_module theano/tests -e test_flake8.py theano/typed_list"
-    - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 PART="theano/compat theano/compile theano/d3viz theano/gof theano/misc theano/sandbox theano/scalar theano/scan_module theano/tests -e test_flake8.py theano/typed_list"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 PART="theano/compat theano/compile theano/d3viz theano/gof theano/misc theano/sandbox theano/scalar theano/scan_module theano/tests -e test_flake8.py theano/typed_list"
-    - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 PART="theano/sparse theano/tensor --exclude-test=theano.tensor.tests.test_basic --exclude-test=theano.tensor.tests.test_elemwise --exclude-test=theano.tensor.tests.test_opt --exclude-dir=theano/tensor/nnet"
     - <<: *normaltest
       env: NUMPY_VERSION=1.13.1 PART="theano/sparse theano/tensor --exclude-test=theano.tensor.tests.test_basic --exclude-test=theano.tensor.tests.test_elemwise --exclude-test=theano.tensor.tests.test_opt --exclude-dir=theano/tensor/nnet"
@@ -84,73 +78,39 @@ jobs:
     - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/tests/test_basic.py"
     - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/tests/test_basic.py"
-    - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/tests/test_basic.py"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/tests/test_basic.py"
-    - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/tests/test_elemwise.py theano/tensor/tests/test_opt.py"
-    - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/tests/test_elemwise.py theano/tensor/tests/test_opt.py"
     - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/tests/test_elemwise.py theano/tensor/tests/test_opt.py"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/tests/test_elemwise.py theano/tensor/tests/test_opt.py"
-    - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/nnet -e test_abstract_conv.py"
-    - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/nnet -e test_abstract_conv.py"
     - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/nnet -e test_abstract_conv.py"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/nnet -e test_abstract_conv.py"
-    - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/nnet/tests/test_abstract_conv.py"
-    - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/nnet/tests/test_abstract_conv.py"
     - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 PART="theano/tensor/nnet/tests/test_abstract_conv.py"
-    - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 PART="theano/tensor/nnet/tests/test_abstract_conv.py"
     - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 FAST_COMPILE=1 FLOAT32=1 PART="theano -e test_flake8.py --exclude-dir=theano/tensor/nnet --exclude-dir=theano/tensor/signal"
-    - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 FAST_COMPILE=1 FLOAT32=1 PART="theano -e test_flake8.py --exclude-dir=theano/tensor/nnet --exclude-dir=theano/tensor/signal"
     - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 FAST_COMPILE=1 PART="theano -e test_flake8.py --exclude-dir=theano/tensor/nnet --exclude-dir=theano/tensor/signal"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 FAST_COMPILE=1 PART="theano -e test_flake8.py --exclude-dir=theano/tensor/nnet --exclude-dir=theano/tensor/signal"
-    - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 FAST_COMPILE=1 FLOAT32=1 PART="theano/tensor/nnet"
-    - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 FAST_COMPILE=1 FLOAT32=1 PART="theano/tensor/nnet"
     - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 FAST_COMPILE=1 PART="theano/tensor/nnet"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 FAST_COMPILE=1 PART="theano/tensor/nnet"
-    - <<: *normaltest
       env: NUMPY_VERSION=1.9.1 FAST_COMPILE=1 FLOAT32=1 PART="theano/tensor/signal"
-    - <<: *normaltest
-      env: NUMPY_VERSION=1.13.1 FAST_COMPILE=1 FLOAT32=1 PART="theano/tensor/signal"
     - <<: *normaltest
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 FAST_COMPILE=1 PART="theano/tensor/signal"
     - <<: *normaltest
-      python: "3.6"
-      env: NUMPY_VERSION=1.13.1 FAST_COMPILE=1 PART="theano/tensor/signal"
 
 script:
   - if [[ $FAST_COMPILE == "1" ]]; then export THEANO_FLAGS=$THEANO_FLAGS,mode=FAST_COMPILE; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ jobs:
       python: "3.4"
       env: NUMPY_VERSION=1.9.1 DOC=1 PART="theano/tests/test_flake8.py"
     - <<: *doctest
-      python: "3.7"
+      python: "3.6"
       env: NUMPY_VERSION=1.13.0 DOC=1 PART="theano/tests/test_flake8.py"
     - &normaltest
       stage: test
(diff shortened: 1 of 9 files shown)


twiecki:
Thanks, is this a required argument now?

lucianopaz:
@twiecki, it's not a required argument (at least it shouldn't be). By default it's assumed to be `None`, and will only raise a `RuntimeError` if the `numpy` version is below 1.13.0 and is supplied with `axis` not None.

lucianopaz:
@twiecki, it looks like I'm failing to make travis install numpy 1.13.0 instead of 1.9.1. I'm not really sure why. With numpy > 1.13 I can run all tests locally without errors.

twiecki:
Maybe we need to update the miniconda version here: https://github.com/Theano/Theano/blob/master/.travis/travis_before_install.sh#L10

twiecki:
https://github.com/Theano/Theano/blob/master/.travis.yml#L42 could also pull in an older version of numpy

twiecki:
We need to update scipy also: 
image

lucianopaz:
I updated the scipy version where I found it, but I'm not sure how to update miniconda in the `travis_before_install.sh`.

twiecki:
You can just replace it with `https://repo.continuum.io/miniconda/Miniconda3-4.5.11-Linux-x86.sh`.

twiecki:
Seems like we also should go to Python 3.6 (or directly 3.7?), there's no reason to use 3.4 anymore.

twiecki:
The other error looks legit, however, and it seems like a more recent numpy is being installed:
```

Warning, treated as error:
WARNING: **********************************************************************
File "library/tensor/extra_ops.txt", line 23, in default
Failed example:
    h1 = theano.function([y], Unique(True, True, True, 0)(z))
Exception raised:
      File "/home/travis/miniconda2/envs/pyenv/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest default[1]>", line 1, in <module>
        h1 = theano.function([y], Unique(True, True, True, 0)(z))
      File "/home/travis/build/Theano/Theano/theano/compile/function.py", line 317, in function
        output_keys=output_keys)
      File "/home/travis/build/Theano/Theano/theano/compile/pfunc.py", line 486, in pfunc
        output_keys=output_keys)
      File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 1839, in orig_function
        name=name)
      File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 1474, in __init__
        self._check_unused_inputs(inputs, outputs, on_unused_input)
      File "/home/travis/build/Theano/Theano/theano/compile/function_module.py", line 1627, in _check_unused_inputs
        i.variable, err_msg))
    UnusedInputError: theano.function was asked to create a function computing outputs given certain inputs, but the provided input variable at index 0 is not part of the computational graph needed to compute the outputs: <TensorType(float64, matrix)>.
    To make this error into a warning, you can pass the parameter on_unused_input='warn' to theano.function. To disable it completely, use on_unused_input='ignore'.
```

twiecki:
```
FAIL: theano.tests.test_flake8.test_format_flake8
  File "/home/travis/miniconda2/envs/pyenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/Theano/Theano/theano/tests/test_flake8.py", line 157, in test_format_flake8
/home/travis/build/Theano/Theano/theano/tensor/extra_ops.py:1170:25: E126 continuation line over-indented for hanging indent
/home/travis/build/Theano/Theano/theano/tensor/extra_ops.py:1216:25: E126 continuation line over-indented for hanging indent
```

abergeron:
Just so you know, those versions are used in travis because they are the minimum supported versions to make sure all new code works with them.

We are hesitant to change the minimum version now that Theano is in maintenance mode, especially since your addition doesn't actually require NumPy 1.13.  You should probably just skip the tests when the required version is not present.

.travis/travis_before_install.sh
```diff
@@ -7,7 +7,7 @@ else
     rm -rf $HOME/miniconda2
     mkdir -p $HOME/download
     if [[ -d $HOME/download/miniconda.sh ]] ; then rm -rf $HOME/download/miniconda.sh ; fi
-    wget -c https://repo.continuum.io/miniconda/Miniconda2-4.1.11-Linux-x86_64.sh -O $HOME/download/miniconda.sh
+    wget -c https://repo.continuum.io/miniconda/Miniconda3-4.5.11-Linux-x86.sh -O $HOME/download/miniconda.sh
```
abergeron: You are now using the 32 bit version of miniconda, and switching to python 3.  Are you sure this is what you want?

.travis/travis_install.sh
```diff
@@ -9,6 +9,6 @@ else
 fi
 
 source activate pyenv
-if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install --yes -q mkl numpy=1.9.1 scipy=0.14.0 nose=1.3.0 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx=1.5.1 mkl-service libgfortran=1 graphviz; fi
-if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]]; then conda install --yes -q mkl numpy=1.9.1 scipy=0.14.0 nose=1.3.4 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx=1.5.1 mkl-service libgfortran=1 graphviz; fi
+if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install --yes -q mkl numpy=1.13.0 scipy=0.19.1 nose=1.3.0 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx=1.5.1 mkl-service libgfortran=1 graphviz; fi
+if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]]; then conda install --yes -q mkl numpy=1.13.0 scipy=0.19.1 nose=1.3.4 pip flake8=2.3 six=1.9.0 pep8=1.6.2 pyflakes=0.8.1 sphinx=1.5.1 mkl-service libgfortran=1 graphviz; fi
```
abergeron: Scipy 0.19.1 and numpy 1.13 have been out for little more than a year.  This not a lot of time for clusters and other slow-moving environments to update.  I would be hesitant to bump up the minimum versions like this.

theano/tensor/extra_ops.py
```diff
@@ -1116,34 +1116,64 @@ class Unique(theano.Op):
     >>> import theano
 
     >>> x = theano.tensor.vector()
-    >>> f = theano.function([x], Unique(True, True, False)(x))
+    >>> f = theano.function([x], Unique(True, True, False, None)(x))
```
abergeron: The last `None` is not a mandatory argument, so you don't need to include it in old examples.

theano/tensor/extra_ops.py
```diff
@@ -1116,34 +1116,64 @@ class Unique(theano.Op):
...
         self.return_index = return_index
         self.return_inverse = return_inverse
         self.return_counts = return_counts
+        self.axis = axis
         numpy_ver = [int(n) for n in np.__version__.split('.')[:2]]
         if self.return_counts and bool(numpy_ver < [1, 9]):
             raise RuntimeError(
                 "Numpy version = " + np.__version__ +
                 ". Option 'return_counts=True' works starting"
                 " from version 1.9.0.")
```
abergeron: This check can probably be removed since we already require NumPy 1.9+

theano/tensor/extra_ops.py
```diff
@@ -1116,34 +1116,64 @@ class Unique(theano.Op):
...
+
+    >>> h2 = theano.function([y], Unique(True, True, True, None)(z))
+    >>> h2([[[0, 1, 1], [0, 1, 1]], [[0, 1, 1], [0, 1, 1]], [[0, 0, 0], [0, 1, 1]], [[3, 2, 0], [0, 1, 1]]])
+    [array([0., 1., 2., 3.]), array([ 0,  1, 19, 18]), array([0, 1, 1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 0, 0, 0, 0, 1, 1, 3, 2, 0, 0, 1, 1]), array([10, 12,  1,  1])]
+
     """
 
-    __props__ = ("return_index", "return_inverse", "return_counts")
+    __props__ = ("return_index", "return_inverse", "return_counts",
+                 "axis")
```
abergeron: You will need to add support for loading old pickles that didn't have this property.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -677,7 +677,23 @@ def setUp(self):
...
+                    Unique(False, True, True, axis=0),
+                    Unique(True, True, True, axis=0),
+                    Unique(axis=-1),
+                    Unique(True, axis=-1),
+                    Unique(False, True, axis=-1),
+                    Unique(True, True, axis=-1),
+                    Unique(False, False, True, axis=-1),
+                    Unique(True, False, True, axis=-1),
+                    Unique(False, True, True, axis=-1),
+                    Unique(True, True, True, axis=-1)]
```
abergeron: You might make the instances with axis into a second list that can be skipped when the required NumPy version is not present.

.travis/travis_install.sh
lucianopaz: You're right, this is a really minor change to Theano and it shouldn't imply a bump up to a new required version of numpy. However, I think it would be nice to have the CI run the tests with different versions of numpy. Based on the errors I was getting when I tried to make travis install numpy 1.13.0, it seems that I will have to configure the tests for python3.6 also, or maybe 3.7 as @twiecki .

theano/tensor/extra_ops.py
lucianopaz: I'll remove it

theano/tensor/extra_ops.py
lucianopaz: I'll remove it

theano/tensor/tests/test_extra_ops.py
lucianopaz: You're right. I'll move these onto a different class that will be expected to fail if the numpy version is below 1.13

theano/tensor/extra_ops.py
lucianopaz: You're totally right! I had not considered the old pickles at all. I'll add a `__setstate__` which sets the class instance's `axis` attribute to `None` by default.

.travis/travis_install.sh
twiecki: Agreed, we should add a Python 3.7 + new numpy + new scipy configuration to travis in addition to the old ones, rather than replacing. That way we make sure minimum requirements as well as the most recent versions are supported.

.travis.yml
```diff
@@ -39,65 +39,118 @@ install:
...
+      env: NUMPY_VERSION=1.9.1 DOC=1 PART="theano/tests/test_flake8.py"
     # re-use prototype, changing the Python version
+    - <<: *doctest
+      python: "2.7"
+      env: NUMPY_VERSION=1.13.0 DOC=1 PART="theano/tests/test_flake8.py"
     - <<: *doctest
       python: "3.4"
+      env: NUMPY_VERSION=1.9.1 DOC=1 PART="theano/tests/test_flake8.py"
+    - <<: *doctest
+      python: "3.6"
```
twiecki: I would do 3.7, if we support 3.4 and 3.7 it's fairly unlikely we wouldn't support 3.6 as well.

.travis/travis_before_install.sh
```diff
@@ -7,7 +7,7 @@ else
     rm -rf $HOME/miniconda2
     mkdir -p $HOME/download
     if [[ -d $HOME/download/miniconda.sh ]] ; then rm -rf $HOME/download/miniconda.sh ; fi
-    wget -c https://repo.continuum.io/miniconda/Miniconda2-4.1.11-Linux-x86_64.sh -O $HOME/download/miniconda.sh
+    wget -c https://repo.continuum.io/miniconda/Miniconda3-4.5.11-Linux-x86_64.sh -O $HOME/download/miniconda.sh
```
twiecki: my bad, this should still be `Miniconda2`.

twiecki:
```
Setting APT mirror in /etc/apt/sources.list: http://archive.ubuntu.com/ubuntu/
3.7 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.7.tar.bz2
```

Seems like that doesn't exist for 14.04. It's a bit odd as we don't need it as we get it through miniconda. To keep things simple maybe just try with 3.6 then (or try to figure out how not to have travis interpret this as the required system python).

twiecki:
sigh, seems like Miniconda changed the MKL installation. Maybe go back to the original miniconda version.

lucianopaz:
@twiecki, I was about to commit with `export MKL_THREADING_LAYER=GNU` in the travis script. Which of the two do think is better?

twiecki:
@lucianopaz I like your approach better.

lucianopaz:
@twiecki, I'm stumped with the docgen that is failing in the python3.6 environment. I don't understand what's wrong, `theano.compile.ops.DeepCopyOp.c_code` doesn't even have a docstring but Sphinx is warning that it does not know the "py" target name, and that warning is being treated as an error.

twiecki:
I think the problem is the underscore, maybe try escaping that?

twiecki:
```

ERROR: test_op (theano.tensor.tests.test_extra_ops.test_Unique_axis)
  File "/home/travis/build/Theano/Theano/theano/tensor/tests/test_extra_ops.py", line 795, in test_op
    self.assertRaises(RuntimeError, self.op_class(*args, **kwargs))
  File "/home/travis/build/Theano/Theano/theano/tensor/extra_ops.py", line 1144, in __init__
    " from version 1.13.0.".format(axis))
RuntimeError: Numpy version = 1.9.1. Option 'axis=0' works starting from version 1.13.0.
``
`
I think the tests need to know which numpy version is used.

**Comment by abergeron** [comment]
The tests already take roughly 2-3 hours to complete on travis.  Doubling the number of tests will push that up to 4-6 hours which is painfully long.

This is why we only test the minimum version.  I'm not opposed to this change per se, but it will make it more painful every time there is a PR.

**Comment by twiecki** [comment]
Are they not parallelized?

File: theano/tensor/tests/test_extra_ops.py [hunk 9]
```diff
@@ -705,7 +705,7 @@ def test_basic_matrix(self):
         # Done by using the op and checking that it returns the right answer.
 
         x = theano.tensor.matrix()
-        inp = np.asarray([[2, 1], [3, 2], [2, 3]], dtype=config.floatX)
+        inp = np.asarray([[2, 1], [3, 2], [2, 1]], dtype=config.floatX)
```
twiecki: Why was this required?

twiecki:
LGTM!

theano/tensor/tests/test_extra_ops.py
lucianopaz: Just to have a duplicate row. Numpy returns a 2x2 array when the `axis=0` parameter value is supplied, and a 3x2 array when `axis=-1`. I wanted to test that the `Op` did this too.

lucianopaz:
I just noticed that the tests in `test_Unique_axis` should raise a `SkipTest` exception instead of just silently passing. I'll push that small change and then I think that this PR will finally be ready.

.travis/travis_install.sh
```diff
@@ -4,11 +4,15 @@ if test -e $HOME/miniconda2/envs/pyenv; then
     echo "pyenv already exists."
 else
     echo "Creating pyenv."
-    if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda create --yes -q -n pyenv python=2.7 ; fi
-    if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]]; then conda create --yes -q -n pyenv python=3.4 ; fi
+    if [[ $TRAVIS_PYTHON_VERSION == '2.7' && $NUMPY_VERSION == '1.9.1' ]]; then conda create --yes -q -n pyenv python=2.7 ; fi
+    if [[ $TRAVIS_PYTHON_VERSION == '2.7' && $NUMPY_VERSION == '1.13.1' ]]; then conda create --yes -q -n pyenv python=2.7 ; fi
+    if [[ $TRAVIS_PYTHON_VERSION == '3.4' && $NUMPY_VERSION == '1.9.1' ]]; then conda create --yes -q -n pyenv python=3.4 ; fi
+    if [[ $TRAVIS_PYTHON_VERSION == '3.6' && $NUMPY_VERSION == '1.13.1' ]]; then conda create --yes -q -n pyenv python=3.6 ; fi
```
nouiz: Testing for the Numpy version is useless here. I would remove the numpy version from the condition and remove the duplicate python=2.7 line.

theano/tensor/tests/test_extra_ops.py
```diff
@@ -758,6 +758,137 @@ def test_infer_shape_matrix(self):
...
+                self._compile_and_check([x],
+                                        [f],
+                                        [np.asarray(np.array([2, 1, 3, 2]),
+                                                    dtype=config.floatX)],
+                                        self.op_class)
+        else:
+            raise utt.SkipTest('Requires numpy >= 1.13')
+
+    def test_infer_shape_matrix(self):
+        if self.expect_success:
```
nouiz: Personal preference. Feel free to ignore.

I perfer this construction. It limit the indentation level of the code. So it make it more clear. I also find that more clear to have the small else case first. Here is what I prefer:

if not self.expect_success:
   raise ...
...

The long clause is one less indentation and we see more clearly the small clause at the beginning.

lucianopaz: I like your suggestion. I'll implement it that way.

twiecki:
Thanks @lucianopaz!

PR #6656, opened 2018-11-06
Fix syntax error in format string in gpuarray/multinomial.py

bfredl:
I got this error on master when using a multinomial op:

 ```
~/.local/lib/python3.7/site-packages/Theano-1.0.3+21.g813faf6a9.dirty-py3.7.egg/theano/gpuarray/multinomial.py in c_code(self, node, name, inp, outputs, sub)
    226 
    227     } // END NESTED SCOPE
--> 228         """ % locals()
    229 
    230         return s

TypeError: ('The following error happened while compiling the node', GPUAMultinomialFromUniform{odtype='int64'}(GpuAdvancedSubtensor.0, GPUA_mrg_uniform{GpuArrayType<None>(float32, vector),inplace}.1), '\n', 'not enough arguments for format string')
```

The change fixes it for me.

### DIFF
--- a/theano/gpuarray/multinomial.py
+++ b/theano/gpuarray/multinomial.py
@@ -183,7 +183,7 @@ KERNEL void k_multi_warp_multinomial(
         do
         {
             nb_threads*=2;
-            if (nb_multi % %nb_threads == 0)
+            if (nb_multi %% nb_threads == 0)
                 nb_blocks = nb_multi/nb_threads;
             else
                 nb_blocks = (int)((float)nb_multi/(float)nb_threads + 1.);


abergeron:
The fix looks good.

Could you add a test to make sure that the problem (or a similar one) doesn't reappear?

bfredl:
I could try. What would be good location?

nouiz:
tests would go in theano/gpuarray/tests/test_multinomial.py 
Thanks

abergeron:
It's better to have the fix with no tests than nothing, so I'm merging.

nouiz:
Which version of Python do you use?

I tested it with Python 2.7, and we already have test for this. But for Python 2.7, the extra space do not cause problem. It get interpreted "correctly" (in the way we wanted). So I'm curious in which case the error happen. Maybe newer python version are more strict.

nouiz:
I tried with Python 3.6 and it also work. Do you use Python 3.7?

bfredl:
Yes, python 3.7. Sorry about the test, I forgot about it..

nouiz:
So no need for tests. It was already tested, it is just Python 3.7 that is more strict.
Thanks.

PR #6659, opened 2018-11-16
Update extension suffix config var name

dfm:
This should remove the annoying deprecation warning on Python 3 where the configuration variable has been renamed to `EXT_SUFFIX` from `SO`. Feel free to ignore if this is wrong or not a good fit.


nouiz:
It seem good to me. @abergeron what do you think?

abergeron:
This looks good.

PR #6665, opened 2018-12-10
fixed bilinear_kernel_1D to return a float32 tensor

justanhduc:
The current implementation returns a float64 tensor which raises a warning/error when the warn_float64 flag is set. I fixed it by casting `ratio` to float32 before the division.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -1766,7 +1766,7 @@ def bilinear_kernel_1D(ratio, normalize=True):
     kern = T.concatenate([half_kern, half_kern[-2::-1]])
 
     if normalize:
-        kern /= ratio
+        kern /= np.cast[theano.config.floatX](ratio)
     return kern
 
 


twiecki:
Thanks @justanhduc, LGTM.

theano/tensor/nnet/abstract_conv.py
```diff
@@ -1766,7 +1766,7 @@ def bilinear_kernel_1D(ratio, normalize=True):
     kern = T.concatenate([half_kern, half_kern[-2::-1]])
 
     if normalize:
-        kern /= ratio
+        kern /= np.cast[theano.config.floatX](ratio)
```
abergeron: Use `T.cast` here, ratio can be symbolic.

abergeron:
You have a bunch of failing tests.  This may be because of the symbolic issue I pointed out.

justanhduc:
> 
> 
> You have a bunch of failing tests. This may be because of the symbolic issue I pointed out.

you are right. the tests failed because the ratio is defined as a theano scalar. i have made a new commit.

twiecki:
separate would be great, thanks!

PR #6671, opened 2019-01-03
Be compatible with Numpy 1.16

rebecca-palmer:
Since Debian unstable moved to Numpy 1.16, all compiles have been failing with "module 'numpy.core.multiarray' has no attribute '_get_ndarray_c_version'":
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=918090
https://ci.debian.net/data/autopkgtest/unstable/amd64/t/theano/1576998/log.gz

Probably because that function was moved by this Numpy commit:
https://github.com/numpy/numpy/commit/31df302656b88df2937c5ac338c96b1a0ec6d687

I do **not** yet know whether this is the only incompatibility, as I have not yet tried the full set of tests with Numpy 1.16; I will report any others I find.

### DIFF
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -1375,7 +1375,7 @@ class CLinker(link.Linker):
 
         # We must always add the numpy ABI version here as
         # DynamicModule always add the include <numpy/arrayobject.h>
-        if np.lib.NumpyVersion(np.__version__)<'1.16.0a':
+        if np.lib.NumpyVersion(np.__version__) < '1.16.0a':
             ndarray_c_version = np.core.multiarray._get_ndarray_c_version()
         else:
             ndarray_c_version = np.core._multiarray_umath._get_ndarray_c_version()
--- a/theano/gof/cc.py
+++ b/theano/gof/cc.py
@@ -1375,8 +1375,12 @@ class CLinker(link.Linker):
 
         # We must always add the numpy ABI version here as
         # DynamicModule always add the include <numpy/arrayobject.h>
+        if np.lib.NumpyVersion(np.__version__)<'1.16.0a':
+            ndarray_c_version = np.core.multiarray._get_ndarray_c_version()
+        else:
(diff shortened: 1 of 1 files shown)


twiecki:
@nouiz is the CI information public?

nouiz:
our own buildbot with GPU and Windows test is down due to our move. We need to rebuild it too (it was broken before it was shutdown for the move). But travis passed and this isn't related to GPU or Windows. So we can merge this PR.

Hopefully, in a few weeks, we can repair the buildbot.

asmeurer:
As you can see from the numerous cross-referenced issues, this is causing problems for a lot of downstream projects. Are there plans to do a bugfix release with this change soon?

nouiz:
Was there report that using the master was enough to support numpy 1.16? We don't have time to test it, but probably we can make the release. In particular, does someone tested it with GPU? I wouldn't like to make many point release for the support of numpy 1.16.

rebecca-palmer:
The Debian package (1.0.3 + this patch, numpy 1.16) [passed testing](https://buildd.debian.org/status/fetch.php?pkg=theano&arch=amd64&ver=1.0.3%2Bdfsg-1&stamp=1547180271&raw=0), but this was without a GPU.

nouiz:
@abergeron did a release. So at least, it will help on the CPU.

asmeurer:
Thank you for doing a release. The number of people pinning to numpy 1.15 because of this was a bit concerning (the problem with version pins is that they tend to stay there longer than they are needed, and you don't realize that you aren't testing against the latest numpy). Hopefully everyone who pinned to 1.15 will remember to remove it now.

nouiz:
I did a comment on the issue that linked here.
I did a quick test on the GPU and it seam to be working. Not all tests
where run. Some optional dependency wasn't installed.

eric-wieser:
Numpy did not intend to remove this function, but no one told us that we did.

It will be coming back in 1.17 and 1.16.5

twiecki:
Great - thanks.

PR #6673, opened 2019-01-09
Fixed shape error of frac bilinear upsample

justanhduc:
This is my promised fix for [this](https://github.com/Theano/Theano/pull/6665#issuecomment-447538668).


nouiz:
It is possible to add a test that cover the case that was broken? This would ensure it do not get broken again later. The test would go in the file:
 theano/tensor/nnet/tests/test_abstract_conv.py

Issue #6669, opened 2018-12-24
Memory issue with the current frac_bilinear_upsampling


justanhduc:
The following code raises a memory error with frac_bilinear_upsampling
```
    X = T.tensor4()
    Y = T.nnet.abstract_conv.bilinear_upsampling(X, 2)
    Z = T.nnet.abstract_conv.frac_bilinear_upsampling(X, (226, 224))
    f1 = nn.function([X], Y)
    f2 = nn.function([X], Z)

    x = np.random.rand(8, 3, 224, 224).astype('float32')
    print(f1(x).shape)
    print(f2(x).shape)
```
It is due to the concatenation of many replications of the input. I have made a quick hack using a modification of `dnn_spatialtf` as follows

```
def frac_bilinear_upsampling(x, frac_ratio):
    from math import gcd
    if not isinstance(frac_ratio, tuple):
        raise ValueError("frac_ratio must be a tuple")
    else:
        if isinstance(frac_ratio[0], tuple):
            f_r = []
            for i, fr in enumerate(frac_ratio):
                p, q = fr
                div = gcd(p, q)
                f_r.append(tuple(np.array(fr) // div))
            frac_ratio = tuple(f_r)
            ratio = (frac_ratio[0][0], frac_ratio[1][0])
            subsample = (frac_ratio[0][1], frac_ratio[1][1])
        else:
            p, q = frac_ratio
            div = gcd(p, q)
            frac_ratio = tuple(np.array(frac_ratio) // div)
            ratio = (frac_ratio[0], frac_ratio[0])
            subsample = (frac_ratio[1], frac_ratio[1])

    theta = np.array([[1, 0, 0],
                      [0, 1, 0]])[None]
    theta = T.tile(T.constant(theta, dtype=theano.config.floatX), (x.shape[0], 1, 1))

    def _dnn_spatialtf(img):
        from theano.scalar import as_scalar
        from theano.gpuarray.dnn import (GpuDnnTransformerGrid, GpuDnnTransformerSampler)
        h = T.ceil(
            img.shape[2].astype(theano.config.floatX) / subsample[0].astype(theano.config.floatX) * ratio[0].astype(
                theano.config.floatX))
        w = T.ceil(
            img.shape[3].astype(theano.config.floatX) / subsample[1].astype(theano.config.floatX) * ratio[1].astype(
                theano.config.floatX))
        out_dims = (img.shape[0], img.shape[1], h, w)
[...]

PR #6674, opened 2019-01-09
Implemented frac bilinear upsample using spatial transformer

justanhduc:
Hi. As I mentioned earlier in [this post](https://github.com/Theano/Theano/issues/6669), there is a memory issue with the current implementation of fraction bilinear upsampling. I re-implemented the function using the dnn spatial transformer op without modifying its interface.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -1803,61 +1803,37 @@ def frac_bilinear_upsampling(input,
     """
 
     T = theano.tensor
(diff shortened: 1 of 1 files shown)


nouiz:
I think it is better to continue the discussion here.
Theano is in maintenance mode only. So personal I'm hesitant to merge new features. If other maintainer want to review this PR, I'm fine to have it merged. But I suppose few people would use it other then you.

This PR need a rebase.

I do not have the time to review this PR correctly. Does someone else have it?

PR #6680, opened 2019-01-17
Update the documentation on gpuarray.sched

### LINKED ISSUE #6677: Update `config.gpuarray.sched` docs with CU_CTX_SCHED_BLOCKING_SYNC
AFAIU, libgpuarray is using `cudaDeviceScheduleBlockingSync` instead of `CudaDeviceScheduleYield` [for a few releases already](https://github.com/Theano/libgpuarray/pull/448#event-1113418241) (see also: #5706). But docs still [say](https://github.com/Theano/Theano/blob/b37f333bd15d404dbb5eb4617eb176d61f63dbd1/doc/library/config.txt#L484-L485) `using "multi" mean using the parameter cudaDeviceScheduleYield`.  Maybe, those last single word should be replaced?

i3v:
Docs update that closes #6677.

### DIFF
--- a/doc/library/config.txt
+++ b/doc/library/config.txt
@@ -482,7 +482,7 @@ import theano and print the config variable, as in:
 
     The sched parameter passed for context creation to pygpu.  With
     CUDA, using "multi" mean using the parameter
-    cudaDeviceScheduleYield. This is useful to lower the CPU overhead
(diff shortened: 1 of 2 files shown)


i3v:
Thanks for merging!

PR #6683, opened 2019-01-25
Fix accidental lowercase

domenzain:
This accidental lowercase makes compilation fail for the gammaincc function.


nouiz:
This mean there is no tests for that. Do you have time to add a basic test for it?

Issue #6685, opened 2019-01-26
`local_add/mul_canonizer` doesn't always reduce to variadic/vector form


brandonwillard:
`Canonizer` doesn't always reduce expressions to their equivalent collapsed variadic/vector forms: e.g. for symbolic tensors `x`, `y`, and `z`, `tt.add(x, tt.add(y, z))` does not reduce to `tt.add(x, y, z)`.

This does occur when some of the inputs are constants, but not when they're all purely symbolic.

Is this the expected behavior, or should it always collapse the graph?

## Example
```python
import theano.tensor as tt
from theano.printing import debugprint as tt_dprint

x, y, z = tt.dvectors('xyz')

def tt_canonicalize(x):
    res = tt.gof.fg.FunctionGraph(tt.gof.graph.inputs([x]), [x])
    opt = tt.gof.opt.EquilibriumOptimizer([tt.opt.local_add_canonizer,
                                           tt.opt.local_mul_canonizer],
                                          max_use_ratio=10)
    opt_res = opt.optimize(res)
    return res.outputs[0], opt_res
```
With only symbolic variables:
```python
>>> test_expr = x + y + z
>>> tt_dprint(test_expr)
Elemwise{add,no_inplace} [id A] ''   
 |Elemwise{add,no_inplace} [id B] ''   
 | |x [id C]
 | |y [id D]
 |z [id E]

>>> tt_dprint(tt_canonicalize(test_expr)[0])
Elemwise{add,no_inplace} [id A] ''   
 |Elemwise{add,no_inplace} [id B] ''   
 | |x [id C]
 | |y [id D]
 |z [id E]

```
With a constant:
```python
>>> test_expr = x + y + z + 0
>>> tt_dprint(test_expr)
Elemwise{add,no_inplace} [id A] ''   
 |Elemwise{add,no_inplace} [id B] ''   
 | |Elemwise{add,no_inplace} [id C] ''   
 | | |x [id D]
 | | |y [id E]
 | |z [id F]
 |InplaceDimShuffle{x} [id G] ''   
   |TensorConstant{0} [id H]

>>> tt_dprint(tt_canonicalize(test_expr)[0])
Elemwise{add,no_inplace} [id A] ''   
 |x [id B]
 |y [id C]
 |z [id D]

```

PR #6686, opened 2019-01-26
Make Canonizer always collapse nested ops

brandonwillard:
Fixes https://github.com/Theano/Theano/issues/6685

### DIFF
--- a/theano/tensor/opt.py
+++ b/theano/tensor/opt.py
@@ -4992,8 +4992,18 @@ class Canonizer(gof.LocalOptimizer):
             return len(x) == len(y) and all(np.all(xe == ye) for xe, ye in
                                             zip(x, y))
 
-        if same(orig_num, num) and same(orig_denum, denum):
-            # We return False if there are no changes
+        if (same(orig_num, num) and same(orig_denum, denum) and
+            # Check to see if we've collapsed some nested ops.
+            not (len(orig_denum) == 0 and
+                 # Make sure this change would increase the number of vector
+                 # arguments--decreasing the number of unnecessary `self.main`
+                 # nodes.
+                 len(node.inputs) < len(orig_num)) and
+            # Do a similar check for the reciprocal op.
+            not (self.use_reciprocal and
+                 node.op == self.reciprocal and
+                 len(orig_num) == 0 and node.inputs[0].owner and
+                 len(node.inputs[0].owner.inputs) < len(orig_denum))):
             return False
 
         new = self.merge_num_denum(num, denum)
--- a/theano/tensor/tests/test_opt.py
+++ b/theano/tensor/tests/test_opt.py
@@ -367,9 +367,9 @@ class test_canonize(unittest.TestCase):
         cases = [
             (fx + fy, (fx, fy), (fxv, fyv), 1, 'float32'),
             (fx * fy, (fx, fy), (fxv, fyv), 1, 'float32'),
-            # (fx+fy+fz,(fx,fy,fz),(fxv,fyv,fzv),1,'float32'),
+            (fx + fy + fz, (fx, fy, fz), (fxv, fyv, fzv), 1, 'float32'),
             # (dx+dy+dz,(dx,dy,dz),(dxv,dyv,dzv),1,'float64'),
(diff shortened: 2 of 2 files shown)


twiecki:
LGTM. I think a more specific test like https://github.com/Theano/Theano/issues/6685 would be good.

brandonwillard:
@twiecki, the uncommented lines in https://github.com/Theano/Theano/blob/5b06db27d6fa9fcb681a5a457871d8e1b6797df3/theano/tensor/tests/test_opt.py#L370 and https://github.com/Theano/Theano/blob/5b06db27d6fa9fcb681a5a457871d8e1b6797df3/theano/tensor/tests/test_opt.py#L372 are effectively the same&mdash;previously&mdash;non-collapsing `x + y + z` setup as my example in #6685.  The `assert` in https://github.com/Theano/Theano/blob/5b06db27d6fa9fcb681a5a457871d8e1b6797df3/theano/tensor/tests/test_opt.py#L427 checks that there's only one `Apply` node corresponding to the single, collapsed `add`/`mul`, which is close to what I would've done within the context of #6685.

Otherwise, I agree with the idea of being more direct by performing the opts exclusively under `tt.opt.local_add_canonizer`/`tt.opt.local_mul_canonizer` and without the other implicit opts or `function` machinery.  Such a set of tests would&mdash;for instance&mdash;likely require considerably less runtime and avoid unwanted interactions with other opts and `function` compilation processes.  However, without changing *all* the relevant tests to work in such a way, I doubt those improvements would really be realized.

My only hesitation involves those interactions and the general question "Besides the clearly relevant `Apply` nodes check, what exactly are those tests targeting?"  Does it include the non-opt aspects of `function` compilation?  The compiled graph is being used for a single numerical calculation, but all that's checked is the resulting dtype; why not optimize the `FunctionGraph` and simply check `fgraph.outputs[0].dtype`?  Alternatively, wouldn't a test value accomplish the same thing?

If the consensus is that those opt interactions shouldn't be there (or at least tested there) and that we don't need a `function`-derived numerical dtype check, then I'll simply refactor those tests to more directly target the relevant opts and graph changes alongside&mdash;if necessary&mdash;some simple test value checks.

brandonwillard:
By the way, with these changes, the currently disabled `test_canonize.test_elemwise_multiple_inputs_optimisation2` succeeds (after fixing a dtype misspecification, e.g. by adding `allow_input_downcast=True`).  There's considerable overlap between
that set of tests and the currently active `test_canonize.test_elemwise_multiple_inputs_optimisation`, but the former does add `DimShuffle`ed vector cases.

twiecki:
I'll wait for @nouiz to take a look also.

nouiz:
I'm reluctant to change that in the current status of Theano (dead).
You are right that you possibly fix some long standing issue in Theano.
But as this is a long standing issue in Theano, some optimization could rely on this behavior.
So there could be other optimization that need to be adapted to this behavior.

If Theano was still actively developed, I would tell we should do it.
But introducing potential degradation inside Theano when there isn't good support to help diagnose and fix possibly consequence that Ididn't foresaw isn't something that I find interesting.

Now, if this change was optional and not enabled by default, I would be happy to review the code in detail and merge it when ready.

To make it optional, you can check in the file theano/configdefaults.py and add a Theano flags for it.

brandonwillard:
As long as the project isn't archived and there are people willing to submit, review and merge PRs, it's still alive!  That said, I can understand if you're saying that you personally don't want to review PRs; that's perfectly fine.  Otherwise, why not archive the repo?

Likewise, I'm not clear on the stated concerns: is the current test suite insufficient for validating these changes?  Is the concern for users and external libraries that rely on `master`?

In light of the aforementioned concerns, adding more code and potential points of failure in lieu of fixing an existing&mdash;albeit broken&mdash;functionality contract, doesn't sound good.

nouiz:
I'm fine doing review. This is what I'm doing. But there isn't significant work. Very minimal work is being done in Theano. So I review code with that point of view.

The current test could find some possible bad consequence of this PR. But it certainly can't guaranty there isn't. We do not have a speed benchmark for example.

What do you think of making this change optional?

PR #6689, opened 2019-02-02
added test for output shape of frac bilinear upsampling

justanhduc:
I added a test to check the output shape error [here](https://github.com/Theano/Theano/pull/6673#issuecomment-453591187). Also, fractions.gcd is deprecated so I changed to math.gcd as its suggestion.

### DIFF
--- a/theano/tensor/nnet/abstract_conv.py
+++ b/theano/tensor/nnet/abstract_conv.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import, print_function, division
 import logging
 from six import reraise, integer_types
 import sys
-from math import gcd
+from fractions import gcd
 
 import theano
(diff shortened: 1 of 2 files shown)


justanhduc:
Sorry for being missing. I'm having some deadline so i will come back with this afterwards.

nouiz:
Thanks for the PR. I do not see anything related to gcd in this PR or in Theano. So I do not understand that part of the description of this issue.

justanhduc:
Hi. Actually in py37 there is a deprecation warning for fraction gcd so I changed to math gcd but in earlier python versions there is no gcd in math so that’s why I had to roll back the fraction gcd.

Sent from Mail for Windows 10

PR #6709, opened 2019-06-01
 Implement gradient in DiffOp for ndim > 1

dsfulf:
The elementwise difference in an array is useful for regularization of noise in a timeseries, e.g. based enforcing smoothness based upon _curvature_: https://www.continuummechanics.org/beambending.html. Adding gradient for ndim > 1 helps when working with multiple time series simultaneously, such as during signal analysis.


twiecki:
Thanks - needs a test.

PR #6716, opened 2019-08-19
Fix deprecation warning.

rpgoldman:
Change the import of `Iterable` from `collections.Iterable` to `collections.abc.Iterable` to future-proof against Python 3.7.8.

Caught this warning when running the PyMC3 tests.

### DIFF
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -465,7 +465,7 @@ class _tensor_py_operators(object):
                     (hasattr(args_el, 'dtype') and args_el.dtype == 'bool')):
                 return True
             if (not isinstance(args_el, theano.tensor.Variable) and
-                    isinstance(args_el, collections.Iterable)):
+                    isinstance(args_el, collections.abc.Iterable)):
                 for el in args_el:
                     if includes_bool(el):
                         return True
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, print_function, division
 try:
     from collections.abc import Iterable
-except AttributeError:
+except (ImportError, AttributeError):
     from collections import Iterable
 import copy
 import traceback as tb
--- a/theano/tensor/var.py
+++ b/theano/tensor/var.py
@@ -1,5 +1,8 @@
 from __future__ import absolute_import, print_function, division
-import collections
+try:
+    from collections.abc import Iterable
+except AttributeError:
+    from collections import Iterable
 import copy
 import traceback as tb
 import warnings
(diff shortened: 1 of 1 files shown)


nouiz:
Thanks for the PR, but this break our build bot. As Theano ins't developed anymore, we won't change the requirement. Can you make this work with the older Python version that we still support? Python 2.7, 3.4-3.6

rpgoldman:
@nouiz OK, working on it. I'm afraid I don't have a full setup for Theano development here, so I'm relying on your build bot to test.  If this works, you can squash merge it, otherwise I will try again until I get something that passes tests.

rpgoldman:
@nouiz I'm confused by the test results: when I look at https://travis-ci.org/Theano/Theano/jobs/573893290 I see that the test failed, but AFAICT it's because the code isn't formatted to comply with `flake8`, and that problem isn't limited to my file -- it's uniform across the code base. So is this a real failure, or should this test be revised or dropped? Alternatively, is there a chance that `flake8` has changed, and that's why this is happening?
Also, the Python 3.4 doc tests seem to fail because of something going on in Pygments, that is unrelated to my little fix -- it's also seen in Adrian's pull request.
Maybe it's time to force a rebuild of master and see if the tests themselves have issues since last run 3 months ago?

nouiz:
we fix the pep8 and flake8 version for tests. So this shouldn't be a problem. But there is one:)
I started a build on the master of Theano while having deleted travis cache: 
https://travis-ci.org/Theano/Theano/builds/574872469

nouiz:
master failed. I do not understand what changes could have caused this.
Here is a PR that try to fix master: https://github.com/Theano/Theano/pull/6717
Mostly I just ignore the error that we didn't enforced before.

rpgoldman:
@nouiz 
> master failed. I do not understand what changes could have caused this.
> Here is a PR that try to fix master: #6717
> Mostly I just ignore the error that we didn't enforced before.

There's also something bad going on [testing with python 3.4](https://travis-ci.org/Theano/Theano/jobs/574872475):

> Pygments requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*' but the running Python is 3.4.5
> The command "pip install flake8-future-import parameterized sphinx_rtd_theme" failed and exited with 1 during .

We should probably move this discussion to a separate issue where it's easier to find.

Note that I have been discussing this with @twiecki and he also tried travis, which now failed completely on `master`.

nouiz:
I fixed the master. Can you rebase this PR?

rpgoldman:
@nouiz OK, rebased and force-pushed. Note that the history of this PR is uninteresting, so if it looks good, you should just squash-merge it...

nouiz:
Thanks. Merging.

PR #6717, opened 2019-08-21
Ignore new error that we didn't enforced before.

nouiz:


rpgoldman:
Dropping support for python 3.4 may be necessary..,

nouiz:
As Theano isn’t well maintained, I do not like much that option.
I would prefer to have a if in the import code that does the right thing for different Python version.

PR #6718, opened 2019-09-06
use string_types instead of str in get_substream_rstates dtype assertion

timodonnell:
Using Theano 1.0.4 via Keras 2.2.5 under Python 2.7.12, calling `keras.model.model_from_json()` for model deserialization fails at this assertion in theano at `get_substream_rstates` because `json.loads()` as called by Keras is returning a dict with unicode objects as values. This PR fixes the problem.


nouiz:
Thanks.

PR #6723, opened 2019-10-01
Theano it is...

Fortschritt69:


nouiz:
I'm curious, why do you want to add that text there? On the page it is rendered, there is a description of Theano above the files: https://github.com/Theano/Theano

If Theano was still being developed, I would agree with your 'let's learn more' at the end, but Theano is barely maintained, so I do not want to encourage people to use it. If you remove that part, I'll merge this PR.

thanks.

PR #6727, opened 2019-10-28
Remove restrictive type assertion

brandonwillard:
An overly restrictive `assert` in [`theano/scalar/basic.py`](https://github.com/theano/Theano/blob/master/theano/scalar/basic.py#L284) prevents the use of `ndarray` subclasses.  This PR simply updates the `assert` so that it considers subclasses.

### DIFF
--- a/theano/scalar/basic.py
+++ b/theano/scalar/basic.py
@@ -166,8 +166,7 @@ class NumpyAutocaster(object):
 
     def __call__(self, x):
         # Make sure we only deal with scalars.
-        assert (isinstance(x, six.integer_types) or
-                isinstance(x, builtin_float) or
+        assert (isinstance(x, (six.integer_types, builtin_float)) or
                 (isinstance(x, np.ndarray) and x.ndim == 0))
 
         if config.cast_policy == 'numpy':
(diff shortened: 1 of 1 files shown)


nouiz:
Why do you need that?

Télécharger Outlook pour Android<https://aka.ms/ghei36>

________________________________

brandonwillard:
It's needed in order to let `theano.tensor.as_tensor_variable` handle `ndarray` subclasses (e.g. when using [`numpy.ndarray.view`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.view.html)).

twiecki:
This looks like a good change to me.

nouiz:
view should return the same class as the input expect if you asked him to change the type. So this is just not a good example for the use case.

Mixing matrix and ndarray isn't a good idea in general. Theano currently only support ndarray from memory.

This change make Theano function return matrix classes that aren't supported in Theano. So I'm hesitant to do such small changes, as I do not have a big picture of where matrix would work and not.

Can you get rid of the matrix class in your code? I prefer to have Theano have a clear we do not support upfront then having something partially supported and not tested at all. What about other ndarray subclass? How Theano would work with them? We are tightly tied to the real ndarray class at many point in Theano.

twiecki:
@noiuz I don't see matrix being mentioned anywhere and that's not what this PR is about. The idea is that a user might subclass ndarray in a specific way and use that in theano which is a standard OOP idiom which gets broken by theano here.

PR #6664, opened 2018-11-28
MAINT: collections import is deprecated

andyfaff:
In Py37 importing from `collections` is a DeprecationWarning and will become an Exception in Py38.

```
/Users/anz/miniconda3/envs/dev3/lib/python3.7/site-packages/theano/compat/__init__.py:79: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  not isinstance(default_factory, collections.Callable)):
```

PR #6741, opened 2020-02-04
Various Python 3.9 related import changes

hroncok:
See https://github.com/Theano/Theano/pull/6664#issuecomment-581845517

### DIFF
--- a/theano/compat/__init__.py
+++ b/theano/compat/__init__.py
@@ -6,15 +6,14 @@ from __future__ import absolute_import, print_function, division
 from six import PY3, b, BytesIO, next
 from six.moves import configparser
 from six.moves import reload_module as reload
+from collections import OrderedDict
 try:
-    from collections.abc import (OrderedDict, MutableMapping as DictMixin,
-                                 Callable)
+    from collections.abc import MutableMapping as DictMixin, Callable
 except ImportError:
     # this raises an DeprecationWarning on py37 and will become
-    # and Exception in py38. Importing from collections.abc
+    # and Exception in py39. Importing from collections.abc
     # won't work on py27
-    from collections import (OrderedDict, MutableMapping as DictMixin,
-                             Callable)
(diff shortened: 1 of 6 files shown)


nouiz:
This looks good to my. I won't test it manually. I'll wait a few days to let a chance to other people to comments too.

hroncok:
There are some more things importe from collections directly, will update this.

theano/compat/__init__.py
```diff
@@ -6,15 +6,16 @@
 from six import PY3, b, BytesIO, next
 from six.moves import configparser
 from six.moves import reload_module as reload
+from collections import OrderedDict
 try:
-    from collections.abc import (OrderedDict, MutableMapping as DictMixin,
-                                 Callable)
+    from collections.abc import Callable, Iterable, Mapping, ValuesView
+    from collections.abc import MutableMapping as DictMixin
```
twiecki: Why change the style syntax? I would just leave it as is with the parentheses.

hroncok: I can do that. The code was changed twice in the row (in the first commit, the import was shorter, in the second it turned out to be longer again, so I just did what I consider more readable).

hroncok: Rebased.

nouiz:
It looks good to me and CI passed. So at least, it didn't cause regression in the Python version in our CI. @twiecki are you also good to merge that?

@hroncok did you ran Theano tests in Python 3.9? I would prefer to not have many PR that fix stuff in a new version in Python if possible.
Thanks

hroncok:
> did you ran Theano tests in Python 3.9?

Yes.

[builder-live.log.gz](https://github.com/Theano/Theano/files/4159952/builder-live.log.gz)


You can add Python 3.9 to Travis CI.

nouiz:
Thanks

PR #6746, opened 2020-03-31
DOC: Allow building with Sphinx >= 2.0

rebecca-palmer:
Needed as build_main is no longer top-level:
https://www.sphinx-doc.org/en/master/extdev/deprecated.html
https://bugs.debian.org/955065

### DIFF
--- a/doc/scripts/docgen.py
+++ b/doc/scripts/docgen.py
@@ -60,6 +60,7 @@ if __name__ == '__main__':
 
     def call_sphinx(builder, workdir):
         import sphinx
+        import sphinx.cmd.build
         if options['--check']:
             extraopts = ['-W']
         else:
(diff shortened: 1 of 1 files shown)


nouiz:
Does this works with old sphinx?

rebecca-palmer:
It's been tested with:
- Sphinx 1.8.5 and Debian's theano (1.0.4 with some patches)
- Sphinx 2.4.3 and theano master (with both this and #6747)

nouiz:
Some CI failure: https://travis-ci.org/github/Theano/Theano/builds/669431703

rebecca-palmer:
Looks like it doesn't work before Sphinx 1.7, and you're using 1.5.1.  Will need a conditional...

rebecca-palmer:
Probably fixed, but only tested with Sphinx 2.4 so wait for the CI.

rebecca-palmer:
Travis failed, but this appears to be because it merged this with master before testing and master is currently failing (see #6747/#6748).

nouiz:
Trigger CI with the new master.

rebecca-palmer:
Travis passed, except the usual .3 uninstallability.  (Maybe .3 should be disabled, or changed to something recent?)

PR #6747, opened 2020-03-31
MAINT: Avoid SyntaxWarnings on import in Python 3.8

rebecca-palmer:
get_io_macros is still broken after this (it doesn't return its results),
but as nobody else seems to have noticed it's probably unused

The "is vs ==" ones don't seem to be a problem in practice,
but let's not scare the user unnecessarily

https://bugs.debian.org/950463

### DIFF
--- a/theano/compile/mode.py
+++ b/theano/compile/mode.py
@@ -261,7 +261,7 @@ class Mode(object):
     def __init__(self, linker=None, optimizer='default'):
         if linker is None:
             linker = config.linker
-        if optimizer is 'default':
+        if type(optimizer)==str and optimizer == 'default':
             optimizer = config.optimizer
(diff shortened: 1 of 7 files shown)


nouiz:
I merged too fast. It break the travis CI. It still run, but it doesn't link to github. Can you look at the failures: https://travis-ci.org/github/Theano/Theano/builds/669432283 and make another PR that fix them? Otherwise, I'll need to revert this PR.

rebecca-palmer:
This appears to be two issues:
- Code style violation (3 instances of == without spaces).  Fixed.  (You may not see this because already-closed pull requests don't update.)
- In variant 3: Jinja2 is no longer installable on Python 3.4.  Nothing to do with Theano.

PR #6748, opened 2020-04-01
flake8 code style fixes

rebecca-palmer:
To #6747


nouiz:
It fixed some failures, but there is still this one: https://travis-ci.org/github/Theano/Theano/jobs/669808036

rebecca-palmer:
That's the "can't install Jinja2 in Python 3.4" one, which I expect would fail on pre-#6747 theano if you ran it now.

Issue #6719, opened 2019-09-12
several nosetest fail


mslacken:
When packaging theano for openSUSE Tumleweed, follwoing nosetets (functions or classes fail):
  * theano/gof/tests/test_compute_test_value.py:
   class TestComputeTestValue
  * theano/sandbox/tests/test_rng_mrg.py:
   def test_uniform() (endless loop)
  * theano/sparse/tests/test_basic.py:
    class SparseInferShapeTester
    def test_csm
    class test_csm
    class Remove0Tester
  * theano/tensor/signal/tests/test_pool.py:
    class TestDownsampleFactorMax
    ClipTester
  * theano/tensor/tests/test_raw_random.py
    def test_vector_arguments
  * theano/tensor/tests/test_shared_randomstreams.py
    class T_SharedRandomStreams
  * tests/test_var.py b/theano/tensor/tests/test_var.py
    def test_numpy_method

nouiz:
Theano isn't actively maintained now. So do not expect fast answer. We need the full error message to be able to help. ALso make sure to use the master branch and not the last release. It contain a few fixes for more recent version of some dependencies. This could be the kind of error you see.

nouiz:
Theano isn't actively maintained now. So do not expect fast answer. We need the full error message to be able to help. ALso make sure to use the master branch and not the last release. It contain a few fixes for more recent version of some dependencies. This could be the kind of error you see.

rebecca-palmer:
Debian has [most of these](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=955471) but not the test_uniform() hang.  Does the fix in #6749 help?

nouiz:
#6749 is merged. So I'll suppose your issues is fixed and will close this issues. If not, do not hesitate to reopen.

PR #6749, opened 2020-04-02
MAINT: Be compatible with numpy 1.17 and scipy 1.3

### LINKED ISSUE #6715: Numpy 1.17.0 issues
I maintain the Theano package for the Fedora Linux distribution.  We recently built all packages for the upcoming Fedora 31, and the Theano package, version 1.0.4, failed its tests.  There was 1 ERROR and 4 FAILs.  I tracked the issue down to the introduction of numpy 1.17.0 in Fedora.  See https://bugzilla.redhat.com/show_bug.cgi?id=1737011 for more information.  Here is what I have figured out so far.

The ERROR is because of changes to numpy's ceil, floor, and trunc ufuncs.  To fix it, class `_tensor_py_operators` in `theano/tensor/var.py` needs `__ceil__`, `__floor__`, and `__trunc__` methods that do exactly what the current `ceil`, `floor`, and `trunc` methods do.

The FAIL in `theano/gof/tests/test_compute_test_value.py` is because there is now one extra frame on the stack.  Line 289 of `theano/gof/tests/test_compute_test_value.py` reads:

```
                frame_info = traceback.extract_tb(tb)[-5]
```

but the -5 has to be a -6 for numpy 1.17.0.

The FAIL in `theano/tensor/tests/test_basic.py` is because the behavior of numpy's clip function changed.  With numpy 1.16.4:

```
>>> np.clip(3, 5, 2)
5
```

but with numpy 1.17.0:

```
>>> np.clip(3, 5, 2)
2
```

I think the order of comparing x to min and max in `Clip.impl` (`theano/scalar/basic.py`) has to be reversed for numpy 1.17.0, if theano's clip is to behave like numpy's, as this test indicates.

I do not know what is causing the last two FAILs.  Two random number tests used to raise ValueError and now do not raise any error at all.

rebecca-palmer:
Fixes #6715 and most of #6719 (= [Debian 955471](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=955471)).  Includes and supersedes #6721.

Tested (affected tests, not a full run) in:
- Debian unstable (Python 3.8, numpy 1.17, scipy 1.3)
- Debian stable (Python 3.7, numpy 1.16, scipy 1.1)

ClipTester: As noted in #6715, numpy changed its behaviour when min > max, but theano didn't.  Removed the test of this case, but not warning for it because clip is a scalar op and we don't want a warning for every entry of a large matrix.

random test_vector_args (both instances): numpy, and hence theano, now accepts size=(1,) even if low/high aren't scalars.  Removed test cases that assume it doesn't.

sparse (both instances): Only a problem with the test setup - the method it used to use for producing a non-index-sorted test matrix now produces an index-sorted matrix.  Switched to a different setup method.

TestComputeTestValue: As previously noted, test was looking at the wrong stack level.  Switched to checking all stack levels to be compatible with both versions.

sort (both instances): numpy now enforces that sort axes are of integer type, not just integer value.  Added a check for integer value then cast to int.

test_numpy_method: Fix taken from #6721.

### DIFF
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -2835,7 +2835,7 @@ ClipTester = makeTester(
               correct6=(randint(5, 5).astype('int64'),
                         np.array(-1, dtype='int64'),
                         np.array(1, dtype='int64')),
-              # min > max case removed as numpy has changed
+              # min > max case moved below as numpy has changed
               correct8=(randint(0, 5).astype('uint8'),
                         np.array(2, dtype='uint8'),
                         np.array(4, dtype='uint8')),
@@ -2843,6 +2843,18 @@ ClipTester = makeTester(
                         np.array(2, dtype='uint16'),
                         np.array(4, dtype='uint16')),)
     # I can't think of any way to make this fail at runtime
+    )
+
+
+# min > max case - numpy.clip has changed but we haven't
+# https://github.com/Theano/Theano/issues/6715
+BackwardsClipTester = makeTester(
+    name='BackwardsClipTester',
+    op=clip,
+    expected=lambda x, y, z: np.where(x < y, y, np.minimum(x, z)),
+    good=dict(correct7=((5 * rand(5, 5)).astype('float64'),
+                        np.array(1, dtype='float64'),
+                        np.array(-1, dtype='float64')),)
     )
 
 
--- a/theano/tensor/tests/test_basic.py
+++ b/theano/tensor/tests/test_basic.py
@@ -2835,11 +2835,7 @@ ClipTester = makeTester(
               correct6=(randint(5, 5).astype('int64'),
                         np.array(-1, dtype='int64'),
                         np.array(1, dtype='int64')),
-              # min > max. messed up behaviour, but
-              # should be same as NumPy's
-              correct7=((5 * rand(5, 5)).astype('float64'),
-                        np.array(1, dtype='float64'),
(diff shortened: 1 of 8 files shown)


nouiz:
Restarting travis.

rebecca-palmer:
Travis green.

theano/tensor/tests/test_basic.py
```diff
@@ -2835,11 +2835,7 @@ def test_cast_from_complex_to_real_raises_error(self):
               correct6=(randint(5, 5).astype('int64'),
                         np.array(-1, dtype='int64'),
                         np.array(1, dtype='int64')),
-              # min > max. messed up behaviour, but
-              # should be same as NumPy's
-              correct7=((5 * rand(5, 5)).astype('float64'),
-                        np.array(1, dtype='float64'),
-                        np.array(-1, dtype='float64')),
+              # min > max case removed as numpy has changed
```
nouiz, line 2838: A better fix could be to modify the expected lambda on line 2819.
It could be something like:
`lambda x, y z: old_expected_behavior if x > y else np.clip(x, y, z)`.

If it is too complicated for you to make that change, I can still merge this PR as is I think.

Thanks for the recent PRs.

rebecca-palmer: Done, and also fixed the documentation that previously claimed it matched numpy.

nouiz:
Thanks.

PR #6750, opened 2020-04-02
Use conda-forge for Python 3.4 to fix travis

nouiz:
Hopefully, this will fix travis.


nouiz:
This fix travis by removing the doctest in python 3.4. It is still tested in python 3.6

PR #6751, opened 2020-04-06
Fixing typo in GammaIncC Op

dfm:
There was a typo in the `GammaIncC` where the injected C code referenced an undefined `gammaQ` function. The function should be called `GammaQ`.

### DIFF
--- a/theano/scalar/basic_scipy.py
+++ b/theano/scalar/basic_scipy.py
@@ -596,7 +596,7 @@ class GammaIncC(BinaryScalarOp):
         if node.inputs[0].type in float_types:
             dtype = 'npy_' + node.outputs[0].dtype
             return """%(z)s =
-                (%(dtype)s) gammaQ(%(k)s, %(x)s);""" % locals()
+                (%(dtype)s) GammaQ(%(k)s, %(x)s);""" % locals()
         raise NotImplementedError('only floatingpoint is implemented')
 
     def __eq__(self, other):


theano/scalar/basic_scipy.py
```diff
@@ -596,7 +596,7 @@ def c_code(self, node, name, inp, out, sub):
         if node.inputs[0].type in float_types:
             dtype = 'npy_' + node.outputs[0].dtype
             return """%(z)s =
-                (%(dtype)s) gammaQ(%(k)s, %(x)s);""" % locals()
+                (%(dtype)s) GammaQ(%(k)s, %(x)s);""" % locals()
```
nouiz, line 599: This mean there was no tests. Normally, I would ask for a tests. But I think it isn't absolutly mandatory. But do you want to add one?

dfm: Sure thing - I'll take a look!

dfm: I should note that I *think* this only ever gets called if `scipy.special` doesn't get imported properly (`imported_scipy_special` is `False`). Is that right? If that's true then I think none of the non-scipy implementations are tested currently: https://github.com/Theano/Theano/blob/master/theano/tensor/tests/test_basic.py#L1717

I'm a little confused because I found this when adding a new method to PyMC3 and running the tests. This op fails to compile when running the tests even though I can import `scipy.special` just fine. I'll keep digging, but I'd love to hear thoughts about why PyMC3 is trying to compile the C code instead of just using scipy!

dfm: The test is failing on Travis too so it's not just my environment that is the problem: https://travis-ci.com/github/dfm/pymc3/jobs/315508171#L3122

ColCarroll: driveby comment, and haven't looked at any of this, but could it be that `scipy` finally moved `logsumexp` (or some other function) from `scipy.special` to `scipy.misc`?

dfm: I haven't figured out why PyMC3 seems to be doing this, but I can confirm that the following steps reproduce the problem:

```bash
conda create --prefix=./env python=3.7 numpy theano
conda activate ./env
python -m pip uninstall scipy
python -c "import theano; theano.scalar.basic_scipy.gammaincc(3.0, 1.0).eval()"
```

I think that testing this on CI would require adding a separate travis job without scipy installed.

nouiz: Thanks for looking at this. As Theano is just in maintenance mode, I do not think it is very useful to add extra supported condition.
If you disagree, I could try to add one rapidly. I'll merge this PR as is.

Thanks.

dfm: Sounds good - thanks for this!

domenzain:
@dfm the reason to use the C code instead of just plain scipy was for some performance gain in GPU as you can see in  #6648 . Originally, the code was available separately from Theano proper at [Theano/Theano_lgpl](https://github.com/Theano/Theano_lgpl).
The underlying C [code](https://borgelt.net/software.html) and [contributions](https://github.com/Theano/Theano_lgpl/issues/1) there were relicensed by the respective authors.

All of this was intended for the creation of Gamma distribution family logCDFs in the context of a [greater push for logCDFs](https://github.com/pymc-devs/pymc3/pull/2688).

Issue #6755, opened 2020-05-15
FutureWarnings all over the place


sammosummo:
I see warnings such as the following in numerous places:

```
/Users/samuelrobertmathias/miniconda3/envs/sammosummo.github.io/lib/python3.8/site-packages/theano/tensor/subtensor.py:2197: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  rval = inputs[0].__getitem__(inputs[1:])
```

I realize Theano is no longer developed. Googling suggests this warning is something packages using numpy under the hood need fix themselves, so posting in case anyone sees.

nouiz:
Make sure to use the github version of Theano. It contains fixes for newer NumPy from memory.

1fish2:
No doubt you've all seen these warnings like from pytest running the PyPI release `Theano==1.0.4`:

```
.../site-packages/theano/compat/__init__.py:16: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import (OrderedDict, MutableMapping as DictMixin,

.../site-packages/theano/misc/frozendict.py:9: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    class frozendict(collections.Mapping):
```

Is there a plan to release an update to PyPI for compatibility with Python 3.9, even if it needs a new PyPI name?

twiecki:
@1fish2 Want to do a PR with those import fixes? I can release a new version on pypi for that.

1fish2:
@twiecki Yes! I'll give it a go.

twiecki:
@1fish2 Any progress?

1fish2:
Yes. Progress in the sense that I can start on it today after crossing off top priority tasks. [Getting to Python 3 is harder than it seems, partly due to problems in installing libraries.] From the docs this looks straightforward...

twiecki:
Closed by #6759.

PR #6759, opened 2020-07-26
Fix collections warnings

1fish2:
Fixes the scary deprecation warnings that clients see, as mentioned in #6755, and some of the other warnings in the code.

* The `compat/__init__.py` DeprecationWarning about 'collections' was already fixed in 1b9804b65ecd008cfb424752e65aced92b63ff92, not yet on PyPI.
* The `misc/frozendict.py` DeprecationWarning about 'collections' just needed to add `'Callable'` to `__all__` in that module.

For a speedy way to test this fix, make a pyenv virtualenv, use this `requirements.txt` file (my best guess on it) to install pips, and run a limited nose test:

```bash
pip install --upgrade pip setuptools virtualenv virtualenvwrapper virtualenv-clone wheel
pip install -r requirements.txt && pyenv rehash
bin/theano-nose theano/tests/test_config.py
```

I ran the limited and the full `bin/theano-nose` in Python 2.7.18, 3.8.5, and 3.9.0b5. (I tried to run it in Python 3.10-dev but `pip install numpy` failed.) The full test can take 1.5 - 3 hours.

I'm unsure how far to go in this PR on fixing other warnings, but Python 3.9 is stricter about regex patterns so I fixed some problems there.

### DIFF
--- a/bin/theano_nose.py
+++ b/bin/theano_nose.py
@@ -23,6 +23,7 @@ import os
 import nose
 import textwrap
 import sys
+import warnings
 from nose.plugins import Plugin
 
 def main_function():
@@ -59,7 +60,7 @@ def main_function():
     batch_size = None
     if len(batch_args):
         if len(batch_args) > 1:
-            _logger.warn(
+            _logger.warning(
                 'Multiple --batch arguments detected, using the last one '
                 'and ignoring the first ones.')
 
@@ -103,7 +104,7 @@ def main_function():
             from numpy.testing.noseclasses import KnownFailure
             addplugins.append(KnownFailure())
         except ImportError:
-            _logger.warn(
+            _logger.warning(
                 'KnownFailure plugin from NumPy could not be imported. '
                 'Use --without-knownfailure to disable this warning.')
     else:
@@ -125,7 +126,7 @@ def main_function():
     except TypeError as e:
         if "got an unexpected keyword argument 'addplugins'" in e.message:
             # This means nose is too old and does not support plugins
-            _logger.warn(
+            _logger.warning(
                 'KnownFailure plugin from NumPy can\'t'
                 ' be used as nosetests is too old. '
                 'Use --without-knownfailure to disable this warning.')
@@ -204,6 +205,8 @@ def main():
     if '--help' in sys.argv or '-h' in sys.argv:
         help()
     else:
+        warnings.simplefilter("default")  # Enable warnings before importing theano
+        os.environ["PYTHONWARNINGS"] = "default"  # Also affect subprocesses
         result = main_function()
         sys.exit(result)
 
--- a/theano/gpuarray/tests/test_subtensor.py
+++ b/theano/gpuarray/tests/test_subtensor.py
@@ -321,7 +321,7 @@ class test_gpuextractdiag(unittest.TestCase):
 
 class TestGpuAllocDiag(test_basic.TestAllocDiag):
     def __init__(self, name):
-        return test_basic.TestAllocDiag.__init__(
+        test_basic.TestAllocDiag.__init__(
             self, name,
             alloc_diag=GpuAllocDiag,
             mode=mode_with_gpu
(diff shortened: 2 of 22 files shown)


theano/gpuarray/tests/test_subtensor.py
```diff
@@ -321,7 +321,7 @@ def test_tensor_float16(self):
 
 class TestGpuAllocDiag(test_basic.TestAllocDiag):
     def __init__(self, name):
-        return test_basic.TestAllocDiag.__init__(
+        test_basic.TestAllocDiag.__init__(
```
twiecki, line 324: I guess this doesn't return anything?

twiecki:
Thanks @1fish2! This looks good. For the `collections` import, could we just drop Python <= 3.6 and get rid of it?

bin/theano_nose.py
```diff
@@ -204,6 +205,8 @@ def main():
     if '--help' in sys.argv or '-h' in sys.argv:
         help()
     else:
+        warnings.simplefilter("default")  # Enable warnings before importing theano
+        os.environ["PYTHONWARNINGS"] = "default"  # Also affect subprocesses
```
nouiz, line 209: I think this should not overwrite that variable if it is already defined by the user.

twiecki: Fixed in https://github.com/Theano/Theano/pull/6760.

theano/gpuarray/tests/test_subtensor.py
1fish2: Conceptually, an `__init__()` method is not allowed to return anything, which technically means it's required to return `None`, which is what happens by default without an explicit `return` value. https://docs.python.org/3/reference/datamodel.html#object.__init__

So this change clarifies the situation and removes a PyCharm inspection warning.

1fish2:
Thank you, @twiecki !

> For the `collections` import, could we just drop Python <= 3.6 and get rid of it?

TL;DR: Can't drop Python 2.7.

It's OK to drop Python 3.0 to 3.3. Some of those earlier versions of Python 3 are hard to even install now. The [PyPI download stats for Theano](https://pypistats.org/packages/theano) shows -- for the last week of data, July 13-19 -- no downloads for Python 3.3 and it doesn't even track 3.0 - 3.2.

But Python 2.7 accounts for about 1/3 of Theano downloads last week. It turns out to be a really big job to move a project from Python 2.7 to 3, so a lot of organizations are still not there. I just spent 2+ months doing it for one project, and that was after a lot of background steps.

The key point: The only practical way to move an application to Python 3 is to incrementally change things until all its parts run in both Python 2.7 and (say) 3.8 then finish testing & debugging. Then they can drop Python 2.7. The long transition requires Python libraries that work in both Python 2.7 and 3.8 and ideally produce identical results.

Even at that, the team has to change the behavior on Python 2.7 to match changes forced by Python 3. To take one example, Python 3 adds "salt" to every string hash as a security measure, but that makes the values differ on every run and always differ from Python 2.7, so it's no longer useful for a pseudo-random number seed.

twiecki:
Thanks for the fixes and explanations @1fish2. Theano 1.0.5 is now released.

bin/theano_nose.py
1fish2: This is a good idea. Perhaps the ideal would be to bump `simplefilter()` and `PYTHONWARNINGS` up to `default` if they're weaker, thus preserving stronger filters, but there doesn't seem to be an API to find out what the warnings filter contains.

So it could be worth moving the `warnings.simplefilter("default")` call into the same `if` statement, that is, set both only if the environment variable was unset.

nouiz:
About Python 2.7, if you release a Theano 1.1, it would be fine to drop it. Most other lib dropped it already. They can still use previous Theano version on 2.7. They just won't get the updates. Just my personal opinion. If NumPy drop it, then taking the few times spend on Theano should be spend elsewhere then supporting Python 2.7.

PR #6775, opened 2021-04-29
Require setuptools theano/gpuarray/linalg.py imports pkg_resources

hrnciar:


twiecki:
This package is dead, see here for a fork: https://github.com/pymc-devs/aesara

PR #6785, opened 2021-12-01
Added logging statements where ever data loading was detected

LoggingMan:
These logging statements will log whenever any data is loaded by the code.


nouiz:
Theano isn't maintained anymore. It was forked at: https://github.com/aesara-devs/aesara

